Issue #622, opened 2018-10-22
Fix PHP Errors


belcherj:
FILE: /Users/jonathanbelcher/Local Sites/woodash/app/public/wp-content/plugins/wc-admin/includes/class-wc-admin-reports-orders-stats-query.php
FOUND 1 ERROR AFFECTING 1 LINE
- [x] 52 | ERROR | Static class properties and methods, as well as class constants, could not be accessed using a dynamic (variable) classname in PHP 5.2 or earlier.

FILE: /Users/jonathanbelcher/Local Sites/woodash/app/public/wp-content/plugins/wc-admin/includes/class-wc-admin-reports-products-query.php
FOUND 1 ERROR AFFECTING 1 LINE
- [x] 44 | ERROR | Static class properties and methods, as well as class constants, could not be accessed using a dynamic (variable) classname in PHP 5.2 or earlier.

FILE: /Users/jonathanbelcher/Local Sites/woodash/app/public/wp-content/plugins/wc-admin/includes/class-wc-admin-reports-products-stats-query.php
FOUND 1 ERROR AFFECTING 1 LINE
- [x] 44 | ERROR | Static class properties and methods, as well as class constants, could not be accessed using a dynamic (variable) classname in PHP 5.2 or earlier.

FILE: /Users/jonathanbelcher/Local Sites/woodash/app/public/wp-content/plugins/wc-admin/includes/class-wc-admin-reports-revenue-query.php
FOUND 1 ERROR AFFECTING 1 LINE
- [x] 60 | ERROR | Static class properties and methods, as well as class constants, could not be accessed using a dynamic (variable) classname in PHP 5.2 or earlier.

FILE: /Users/jonathanbelcher/Local Sites/woodash/app/public/wp-content/plugins/wc-admin/includes/data-stores/class-wc-admin-reports-data-store.php
FOUND 10 ERRORS AFFECTING 8 LINES
- [x] 141 | ERROR | Empty ELSEIF statement detected
- [x] 162 | ERROR | Missing parameter comment
- [x] 163 | ERROR | Missing parameter comment
- [x] 164 | ERROR | Missing parameter comment
- [x] 182 | ERROR | Doc comment for parameter "$query_args" missing
- [x] 182 | ERROR | Doc comment for parameter "$db_interval_count" missing
- [x] 182 | ERROR | Doc comment for parameter "$expected_interval_count" missing
- [x] 189 | ERROR | Doc comment for parameter $db_records does not match actual variable name $query_args
[...]

PR #641, opened 2018-10-23
Fix errors in lib/common.php

belcherj:
#622

Fixes linting errors in the PHP for lib/common.php

 - Not sure how to test.  Not even sure the one function is still even used.

### DIFF
--- a/lib/common.php
+++ b/lib/common.php
@@ -1,5 +1,11 @@
 <?php
 /**
+ * Common lib functions
+ *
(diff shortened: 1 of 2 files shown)


ryelle (approved):
Looks good. Clicked around pages & scripts are still embedded on the right pages. `wc_admin_get_embed_breadcrumbs` adds the correct items to `wcSettings.embedBreadcrumbs`.

PR #642, opened 2018-10-23
Fix linting errors in tests/test-sample.php

belcherj:
#622

Fixes linting errors in tests/test-sample.php

You shouldn't need to test if all tests pass


claudiosanches (approved):
LGTM

PR #644, opened 2018-10-23
Fixes lints in lib/admin.php

belcherj:
#622

Fixes lints in lib/admin.php.  There is one remaining error but I do not have the knowledge to fix it.

 - Not sure.

### DIFF
--- a/lib/admin.php
+++ b/lib/admin.php
@@ -226,7 +226,9 @@ function wc_admin_admin_after_notices() {
 add_action( 'admin_notices', 'wc_admin_admin_after_notices', PHP_INT_MAX );
 
 /**
- * TODO Can we do some URL rewriting so we can figure out which page they are on server side?
+ * Edits Admin title based on section of wc-admin.
+ *
+ * @TODO Can we do some URL rewriting so we can figure out which page they are on server side?
  *
  * @param string $admin_title Modifies admin title.
  */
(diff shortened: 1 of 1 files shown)


lib/admin.php
```diff
@@ -155,6 +161,7 @@ function wc_admin_link_structure() {
 	array_unshift( $submenu['woocommerce'], $menu );
 
 	// Rename "Analytics" to Overview (otherwise this reads Analytics > Analytics).
+	// phpcs:ignore
```
claudiosanches: This should ignore everything, since we only want to allow overide of a WP global, better use `// WPCS: override ok.` instead.

claudiosanches: Like: 

```php
$submenu['wc-admin#/analytics'][0][0] = __( 'Overview', 'wc-admin' ); // WPCS: override ok.
```

lib/admin.php
```diff
@@ -209,7 +226,11 @@ function wc_admin_admin_after_notices() {
 }
 add_action( 'admin_notices', 'wc_admin_admin_after_notices', PHP_INT_MAX );
 
-// TODO Can we do some URL rewriting so we can figure out which page they are on server side?
+/**
+ * TODO Can we do some URL rewriting so we can figure out which page they are on server side?
```
claudiosanches: Better include a properly docblock here, and then use `@todo`, we use this on core and will make easy to find TODOs later.

claudiosanches (changes requested):
Looks good, but needs some small changes before we merge.

belcherj:
All Fixed

claudiosanches (approved):
LGTM

Issue #621, opened 2018-10-22
Tables CSV download includes hidden columns


Aljullu:
**Describe the bug**
Downloading table data as a CSV doesn't honor hidden columns.

**To Reproduce**
1. Go to any page with a table and hide some of the columns.
2. Download the table.

_Expected behavior:_
Hidden columns should not be included in the CSV.

_Actual behaviour:_
Hidden columns are included in the CSV.

**Screenshots**
![screenshot from 2018-10-22 17-49-12](https://user-images.githubusercontent.com/3616980/47303762-163c8880-d625-11e8-894d-090411edce91.png)
Notice that _Status_, _Customer_, _Products_, etc. are hidden in the UI but included in the CSV.

PR #652, opened 2018-10-23
Move Download table logic inside the TableCard component

Aljullu:
While working on #621, I noticed we were repeating the same code to download tables as CSV's in every component that contains a downloadable table:
```ES6
downloadCSVFile(
	generateCSVFileName( 'revenue', query ),
	generateCSVDataFromTable( headers, rows )
);
```

Given that the process to download a table as a CSV is the same no matter what the table is, I think it makes sense to move this logic inside the `<TableCard>` component.

This PR does that and adds a `downloadable` prop to `TableCard` that will display the download button for the table. I didn't remove the `onClickDownload` prop, beucase we might want to use it in the future when implementing big data sets download via an API request.

### Screenshots

_(this PR shouldn't produce any visible change)_

 - Go to any page with a table that can be downloaded (`/wp-admin/admin.php?page=wc-admin#/analytics/revenue` and `/wp-admin/admin.php?page=wc-admin#/analytics/orders`).
- Download the table and verify the process works properly.


timmyc (approved):
Nice abstraction here Albert! Tested it out and working well for me.

PR #655, opened 2018-10-23
Fixes PHP 5.2 compatibility issues

belcherj:
#622

Fixes lint issues around a 5.2 PHP Compatibility issues.

Detailed test instructions:
- Do reports load?
- run `npm run lint:php`

You should not see: 
```
FILE: /Users/jonathanbelcher/Local Sites/woodash/app/public/wp-content/plugins/wc-admin/includes/data-stores/class-wc-admin-reports-data-store.php
292 | ERROR | Static class properties and methods, as well as class constants, could not be accessed using a dynamic (variable) classname in PHP 5.2 or earlier.
```

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-data-store.php
@@ -290,7 +290,7 @@ class WC_Admin_Reports_Data_Store {
 	 */
 	protected function get_cache_key( $params ) {
 		// TODO: this is not working in PHP 5.2 (but revenue class has static methods, so it cannot use object property).
-		return 'woocommerce_' . $this::TABLE_NAME . '_' . md5( wp_json_encode( $params ) );
+		return 'woocommerce_' . self::TABLE_NAME . '_' . md5( wp_json_encode( $params ) );
 	}
 
 	/**


allendav:
`WC_Admin_Reports_Data_Store TABLE_NAME` is an empty string. Are you sure this is what you want?

allendav (commented):
`WC_Admin_Reports_Data_Store TABLE_NAME` is an empty string. Are you sure this is what you want?

belcherj:
I have no idea what this code is supposed to do but this change fixes the PHP compatibility issue without changing the functionality.

ryelle:
@peterfabian committed this file, so perhaps he knows what [the `TABLE_NAME`](https://github.com/woocommerce/wc-admin/blob/master/includes/data-stores/class-wc-admin-reports-data-store.php#L36) should be?

peterfabian:
The idea was to use `TABLE_NAME` from descendants of class `WC_Admin_Reports_Data_Store` to construct the identifier for the db cache (together with parameters for the query). That is why `$this` was used. However, as mentioned in the comment, this does not work in PHP 5.2 and `TABLE_NAME` can't be switched to a member variable as some descendants have static methods.

Changing `$this` to `self` would break the cache id. 

Just from the top of my head, one option would be to move the cache id function to each of the descendants. Another option would be to split off descendants with static methods into multiple classes. Perhaps there are other, better options?

belcherj:
Closing this as the solution will break things.  I'll leave it to the PHP experts to figure this one out @peterfabian @claudiosanches

Issue #624, opened 2018-10-22
Clean up analytics menu


LevinMedia:
A few house keeping items to clean up the top level wc-admin analytics menu:

- [ ] Remove "overview"  from sub menu - should default to revenue report when clicking on analytics
- [ ] The order of submenu items should be as follows as the reports become available: Revenue, Orders, Products, Categories, Coupons, Taxes, Downloads, Stock, Customers

PR #656, opened 2018-10-23
Updates wc-admin menu order and entries #624

belcherj:
#624

Chnages order and contents of the Analytics menu.

### Screenshots

screen shot 2018-10-23 at 4 35 36 pm

 - Open wp-admin
 - Click on analytics
 - Observe order matches that in #624

### DIFF
--- a/lib/admin.php
+++ b/lib/admin.php
@@ -95,14 +95,6 @@ function wc_admin_register_pages() {
 
 	wc_admin_register_page(
 		array(
-			'title'  => __( 'Products', 'wc-admin' ),
-			'parent' => '/analytics',
-			'path'   => '/analytics/products',
-		)
-	);
-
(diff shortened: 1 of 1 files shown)


LevinMedia:
Per a slack convo with @belcherj 

Analytics will be removed as a sub menu item, clicking the primary analytics menu will default to the revenue report. Likewise, clicking analytics in the breadcrumbs will also default a user to the revenue report for the time being. 

We'll hide the reports until we begin working on each one - for now we'll only show revenue, orders, and products

ryelle (approved):
Submenu is correctly reordered 👍 

Will "Clicking "analytics" in the breadcrumbs will also default a user to the revenue report" happen in a following PR, or was that meant to be here, too?

belcherj:
There is a follow up PR for that.

PR #666, opened 2018-10-24
Fix resolvers tests giving false positives

### LINKED ISSUE #654: Store resolvers tests give false positives
As @psealock pointed out here: https://github.com/woocommerce/wc-admin/pull/619#discussion_r227182044, resolvers tests are giving false positives, for example, when the path in the tests is not correct.

Aljullu:
Fixes #654.

@psealock noticed resolvers tests were giving false positives, this PR should fix that.

 - Run `npm test` and verify tests pass.
 - Modify resolvers tests to make them fail and verify there are no false positives.

### Example modifications to make tests fail
Modify [this line](https://github.com/woocommerce/wc-admin/pull/666/files#diff-7568b97f47327e3efc4f74e97aee1c3cR50) in ` client/store/notes/test/resolvers.js` from:
```ES6
expect( dispatch().setNotes ).toHaveBeenCalledWith( NOTES_2, { page: 2 } );
```
to:
```ES6
expect( dispatch().setNotes ).toHaveBeenCalledWith( NOTES_1, { page: 2 } );
```
----
Modify [this line](https://github.com/woocommerce/wc-admin/pull/666/files#diff-323232275f2eca0c97da4c2adb46d760R49) in `client/store/products/test/resolvers.js` from:
```ES6
if ( options.path === '/wc/v3/reports/products' ) {
```
to:
```ES6
if ( options.path === '/wc/v3/products' ) {
```
Tests should fail in both cases (they weren't failing before).

### DIFF
--- a/client/store/notes/test/resolvers.js
+++ b/client/store/notes/test/resolvers.js
@@ -6,6 +6,7 @@
  * External dependencies
  */
 import apiFetch from '@wordpress/api-fetch';
+import { dispatch } from '@wordpress/data';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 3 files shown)


belcherj (approved):
Tested and works.

PR #667, opened 2018-10-24
Don't include hidden columns when downloading a table as CSV

### LINKED ISSUE #621: Tables CSV download includes hidden columns

Aljullu:
Fixes #621.

Downloading tables while a column was hidden, was still including that column in the generated CSV file. This PR fixes that.

### Screenshots
image

- Go to a page with a table (eg: `/wp-admin/admin.php?page=wc-admin#/analytics/products`).
- Hide some columns with the top right button.
- Download the table and verify the hidden columns are not included in the CSV.

### DIFF
--- a/client/components/table/__mocks__/table-mock-csv.js
+++ b/client/components/table/__mocks__/table-mock-csv.js
@@ -1,5 +1,5 @@
 /** @format */
 
-export default `Date,Orders,Gross Revenue,Refunds,Coupons,Taxes,Shipping,Net Revenue
-2018-10-01 00:00:00,1,411,0,0,0,0,411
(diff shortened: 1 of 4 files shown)


belcherj (approved):
Nice! Love how small that change is to make!

PR #619, opened 2018-10-22
Products table

Aljullu:
Fixes #535.

Adds a table connected to the API to the _Products_ report.

As explained in #535, there is some data which is still not returned by the API, for now it's replaced with empty defaults and is marked with `// @TODO` comments in the code.

### Screenshots
image

### Detailed test instructions:

* Go to _Analytics_ > _Products_.
* Verify the table with products is there and works as expected.
* Try sorting rows, downloading its contents, clicking on the links, etc.

PR #668, opened 2018-10-24
Use downloadable prop in Products report table

Aljullu:
After merging #619 and #652 in parallel, downloading the _Products_ table as a CSV was downloading two files. This PR should fix that.

 - Go to the _Products_ page (`/wp-admin/admin.php?page=wc-admin#/analytics/products`).
 - Download the table and verify only one file is downloaded.


belcherj (approved):
Looks great, tested, works.

PR #677, opened 2018-10-24
Build: Update babel config

ryelle:
This PR moves `.babelrc` to `babel.config.js`, matching gutenberg's babel config. This also adds the missing plugin `@babel/transform-react-jsx`. This previously fixed an issue with Jest + lerna, and while that doesn't seem to be the case as of today 🙃 it looks like `babel.config.js` is the recommended format for config.

**To test**

- Build the project: `npm run build`
- It should work
- Watch for changes: `npm start`
- It should work


psealock (approved):
Testing as expected, LGTM 🚢

PR #678, opened 2018-10-24
Try: Add Lerna to split out utility packages

### LINKED ISSUE #508: Add Utils to Exports/Global Namespace
Currently we are making our wc-admin components available via the `@woocommerce/components` [and `wc.components`](https://github.com/woocommerce/wc-admin/blob/master/webpack.config.js#L7) which allows developers a route in which to consume and use them within their extensions.

While working on an example Report extension - I hit a few scenarios where having access to utility functions like [`formatCurrency`](https://github.com/woocommerce/wc-admin/blob/master/client/lib/currency/index.js#L15) [#](https://github.com/timmyc/wcs-example-report/blob/master/js/src/report.js#L142-L145)  and [`getCurrentDates`](https://github.com/woocommerce/wc-admin/blob/master/client/lib/date/index.js#L261-L284) [#](https://github.com/timmyc/wcs-example-report/blob/master/js/src/report.js#L63-L79) would have made my development a bit easier.

The pattern established in Gutenberg is to create new packages for this sort of functionality ( i.e. [`@wordpress/date`](https://github.com/WordPress/gutenberg/tree/master/packages/date) ) So I was wondering if we should follow that pattern, or consider having a more catch-all / generic `@woocommerce/utils` - either way being able to access these types of utils via a global or an import would really help extension developers work more efficiently against our project.

/cc @ryelle for your thoughts on this since you have worked quite a bit with the extensibility in the project thus far.

ryelle:
Fixes #508 – This exposes new libraries, `wc.currency` and `wc.date` for third party developer use. It also sets up these packages in a lerna framework, so we can publish to npm `@woocommerce/currency` and `@woocommerce/date`.

In this PR:

- Moved date & currency to new `packages` directory, can now be included as `@woocommerce/date`, `@woocommerce/currency`
- Added build process for building packages, generates build & build-module like gutenberg packages. Will also generate CSS, but neither package currently has styles.
- NPM scripts:
  - Added `clean` command to wipe all built files from the directory
  - `build` command has been renamed to `build:core`, while `build` itself runs both `build:core` and `build:packages`.
  - `start` command has also been updated to watch packages along with `client` code changes.
  - If you want to watch just the packages for changes, you can run `dev:packages` directly.
  - `publish:check`, `publish:dev`, `publish:prod`: these are lerna commands to build and publish the packages to npm
- Each package now has a `package.json`, `.npmrc`, and `README.md`. See the [Managing Packages workflow from gutenberg](https://github.com/WordPress/gutenberg/blob/master/CONTRIBUTING.md#managing-packages) for details.
- The Jest tests have been updated to point to these packages when `@woocommerce/date` etc are found. Further packages will need to be added to `setup-globals.js` as they're created.
- The Webpack config has also been updated to handle `@woocommerce/` packages, and will also need updates as more packages are added.

I've published alpha versions of [date](https://www.npmjs.com/package/@woocommerce/date) & [currency](https://www.npmjs.com/package/@woocommerce/currency) to npm, using `lerna publish --canary`

**To test**

- Build the app, make sure everything succeeds
- There should be no change in functionality
- Check that `wc.currency` and `wc.date` are available on the global object by trying it in the console.
- Check that tests pass (run yourself, not just relying on travis)
- Try out the various npm scripts, though I would avoid publishing packages just yet 🙂


psealock (approved):
Wow, nice work. Every test item works (did not try to publish 😸) and scripts behave as they should. 

Is the idea that `/components` eventually be moved to packages _a la_ Gutenberg?

bin/packages/post-css-config.js
```diff
@@ -0,0 +1,22 @@
...
+	require( './node_modules/@wordpress/postcss-themes' )( {
+		defaults: {
+			primary: '#0085ba',
+			secondary: '#11a0d2',
+			toggle: '#11a0d2',
+			button: '#0085ba',
+			outlines: '#007cba',
+		},
+		themes: {
+			'woocommerce-page': {
```
timmyc: So this means we could potentially add another theme to target the calypsoify env to make things match colors there?

ryelle: I think so – I think it's based on a body class. I'm not entirely sure how this part works, I copied this from @psealock's work on the postcss theme, but since nothing here relies on CSS (yet) I haven't been able to test it.

ryelle:
> Is the idea that /components eventually be moved to packages a la Gutenberg?

Yep, along with most of the things in `client/lib`, all in follow-up PRs though 🙂

bin/packages/post-css-config.js
psealock: >So this means we could potentially add another theme to target the calypsoify

Yes, exactly.

>but since nothing here relies on CSS (yet) I haven't been able to test it.

The GB buttons are themed using this post-css function and config. I tested this out by adding a new theme 

```js
'testing-add-lerna': {
    primary: '#4ade1d',
    secondary: '#4ade1d',
    toggle: '#4ade1d',
    button: '#4ade1d',
    outlines: '#4ade1d',
},
```
I can see the correctly constructed CSS rules in `/dist/css/index.css` and `/dist/css/embedded.css`:

```
body.testing-add-lerna .components-button.is-primary { ... }
```

For some reason those rules aren't being added to `dist/app/index.css` or `dist/components/index.css` and so I can't see the theme expressed in wc-admin.

Also, the old `postcss.config.js` can be removed.

ryelle: > For some reason those rules aren't being added to dist/app/index.css or dist/components/index.css and so I can't see the theme expressed in wc-admin.

Which file did you add the new theme to?

> Also, the old postcss.config.js can be removed.

Once the components are built from packages, sure – right now they shouldn't be using this build process though, which is what I meant by not being able to test it.

psealock: Oh I see now, thanks. So if I were to add a new theme now, it would have to be to both postcss configs. In any case, that means the new `bin/packages`/post-css-config.js` is working as expected.

timmyc:
Since this touches so many files, lets get it in so we don't have to deal with conflicts too much. This is amazing work Ryelle, thank you!

PR #685, opened 2018-10-25
refactoring x-axis ticks

LevinMedia:
@greenafrican this is the PR I mentioned in slack, for reference: 

This gets the x-axis most of the way to where I want it, per what I was describing at p6riRB-3C9-p2 

I added an if / else to week intervals, similar to what you already had for day intervals. However, I ran into two issues when the interval was set to week and it's using the `else` logic:

1.) I couldn't figure out how to remove duplicate instances of the month from x1

screen shot 2018-10-25 at 4 56 01 am

2.) If we're able to remove the duplicate months from x1, I'd just like to ensure that over longer time periods, we're able to place a tick on the first week of the month. The way it's set up now, some times the first label of a month won't happen until the second week, such as this example with feb:

screen shot 2018-10-25 at 5 13 41 am

### DIFF
--- a/client/lib/date/index.js
+++ b/client/lib/date/index.js
@@ -442,13 +442,8 @@ export function getDateFormatsForInterval( interval, ticks = 0 ) {
 			}
 			break;
 		case 'week':
-			if ( ticks < weekTicksThreshold ) {
-				xFormat = '%d';
-				xFormat = '%b %Y';
-			} else {
-				xFormat = '%d %b';
-				x2Format = '%Y';
-			}
+			xFormat = '%d';
+			x2Format = '%b %Y';
 			tooltipFormat = __( 'Week of %B %d %Y', 'wc-admin' );
 			break;
 		case 'quarter':


greenafrican:
@LevinMedia 
I've rebased against `master` and this seems to address 1) above but not 2):

screenshot 2018-10-26 10 26 00

Due to the specificness of weekly intervals, should we ensure the day is also included? I'll work on something that I think makes sense and update this PR for your review.

greenafrican:
@LevinMedia I've updated to include the day in weekly intervals, now that we're eliminating duplicate date components from the x2 axis, this feels better. Let me know what you think:
screenshot 2018-10-26 15 02 53

Here's what it looks like over a longer period:
screenshot 2018-10-26 15 09 52

LevinMedia:
@greenafrican I'd actually prefer the x-axis ticks for weekly intervals the way they were in my first commit - and here's my logic behind that:

My goal is to keep the x-axis ticks as consistent as possible across time scales, and different intervals. 

Regardless of whether or not the interval is day or week, if the time period is more than two months (63 days / 8 weeks)  - I'd like  to render x1 as month, and x2 as year. 

>Due to the specificness of weekly intervals, should we ensure the day is also included?  

Yes, but only if the time frame is two months or less. In that case x1 would be day, and x2 would be month / year. (same as your example above)

Time periods of more than two months with weekly intervals are more of a trend analysis, and the specific date a week starts or end becomes less meaningful. If there is a question about a specific date a week starts, the user is able to hover or tap the chart to get that information out of the tool tip.

timmyc:
Are we good to merge this one in @LevinMedia ?

LevinMedia:
Sure, lets merge it, we can continue it in another PR

PR #686, opened 2018-10-25
Create QUERY_DEFAULTS object with default values for queries

### LINKED ISSUE #634: Create a QUERY_DEFAULTS object with the default values for all queries
Coming from https://github.com/woocommerce/wc-admin/pull/619#discussion_r227086689.

In several places of the codebase we have hardcoded query defaults, like the tables page size (`25`). It would be better having all of them stored in a global so it can be easily modified in the future. That would include date presets as well.

Aljullu:
Fixes #634.

Stores the default page size and default period and compare values in a QUERY_DEFAULTS global.

- Go to _Analytics_ pages (_Revenue_, _Orders_ and _Products_) and verify:
  - the default date range is still 'Month to Date' and the comparison period is previous year
  - the default table size is still 25


timmyc (approved):
Excellent! working well in my testing too

psealock (approved):
Cool, thanks for this

PR #689, opened 2018-10-25
Adds Section Header component

### LINKED ISSUE #660: Component: Dashboard Section Header
Designs: [inVision](https://automattic.invisionapp.com/d/main#/projects/prototypes/15060831)
Discussion: p6riRB-3qG-p2

wc-admin dashboard analytics preview mode - invision 2018-10-23 12-44-52

Not certain that is the best name for the component, so feel free to come up with one that tickles your fancy! But essentially the component needs to provide a section heading, and action area for the dashboard sections.

Possible Props:
- Title
- Ellipsis Menu
- Other Buttons ( like chart switcher )

belcherj:
Closes #660

Adds a Section Header component that abstracts out Card's header and allows for stand alone use with different formatting.

### Screenshots
screen shot 2018-10-25 at 3 15 21 pm
screen shot 2018-10-25 at 3 16 47 pm
screen shot 2018-10-25 at 3 16 59 pm

 - Open WooCommerce -> Dashboard
 - Does section header render correctly
 - Open each avail report
 - Do chart headers render correctly?


LevinMedia:
Awesome to see this starting to take shape! Just want to point this out, there's a dashboard post here: p6riRB-3qG-p2 it details some specific behavior of the section header. Essentially, people may add as many of these as they want, so we also want to give them the option to re-name each section as they are added: 

yw2egf5_z6

That may change the way you abstract the existing card headers.

belcherj:
I think this PR is good to be reviewed as that functionality can be added later when needed.

ryelle:
@belcherj I'm not sure I agree with the approach that Card headers and Section headers should be the same component, I think they have distinct-enough functionality to keep them separate - Section headers won't have `actions`, and Card headers don't need to be renamed, for example.

belcherj:
screen shot 2018-10-26 at 1 29 58 pm

I'm happy to separate them out.  The section header will have actions.  The only thing currently different between the two is the css which is very minor.

Given that Section Headers do have actions do you still feel the same?  @ryelle

ryelle:
> Given that Section Headers do have actions do you still feel the same? 

I do still _feel_ like they're separate components, but given that I have no other reasoning, merging them should be fine 😁

belcherj:
@ryelle restored card header and made title optional to not show the header.

client/components/card/style.scss
```diff
@@ -1,9 +1,9 @@
 /** @format */
 
 .woocommerce-card {
-	margin-bottom: $gap-large;
 	background: white;
 	border: 1px solid $core-grey-light-700;
+	margin-bottom: $gap-large;
```
ryelle: Out of curiosity, why move this? We don't have a set CSS property order for the project, would you want to add one?

client/components/section-header/index.js
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Internal dependencies
+ */
+import './style.scss';
+import EllipsisMenu from 'components/ellipsis-menu';
+import { H } from 'components/section';
+import { validateComponent } from 'lib/proptype-validator';
+
+/**
+ * A basic card component with a header. The header can contain a title, an action, and an `EllipsisMenu` menu.
```
ryelle: This needs to be updated, this is copied from `Card` currently.

client/components/section-header/index.js
```diff
@@ -0,0 +1,66 @@
...
+	 */
+	menu: validateComponent( EllipsisMenu ),
+	/**
+	 * The title to use for this card.
+	 */
+	title: PropTypes.oneOfType( [ PropTypes.string, PropTypes.node ] ).isRequired,
+	/**
+	 * Is this a standalone header
+	 */
+	standAlone: PropTypes.bool,
```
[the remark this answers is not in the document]
ryelle: Actually, do you think there will be a case for a `SectionHeader` that isn't standalone, since Card is keep its header?

client/components/section-header/test/index.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * External dependencies
+ *
+ * @format
+ */
```
ryelle: I usually leave out the `External dependencies` comment if there are none.

client/components/section-header/test/index.js
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Internal dependencies
+ */
+import SectionHeader from '../';
+
+describe( 'SectionHeader', () => {
```
ryelle: This is probably a good place [to use snapshot tests](https://jestjs.io/docs/en/snapshot-testing).

ryelle (commented):
Most of my comments are nitpicks, sorry - but overall this is working well 🙂

client/components/section-header/example.md
```diff
@@ -0,0 +1,9 @@
+/\*_ @format _/
+
+/\*_ @format _/
```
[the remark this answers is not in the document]
belcherj: I think prettier is adding these.  Probably need to remove prettier from running on markdown

client/components/section-header/test/index.js
belcherj: 6:1  error  Missing external, internal dependencies docblocks  wpcalypso/import-docblock

client/components/section-header/example.md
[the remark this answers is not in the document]
belcherj: Removed

client/components/section-header/style.scss
```diff
@@ -0,0 +1,63 @@
...
+	@include breakpoint( '<782px' ) {
+		margin-left: -16px;
+		margin-right: -16px;
+		margin-bottom: $gap-small;
+		border-left: none;
+		border-right: none;
+		width: auto;
+	}
+
+	&.stand-alone {
```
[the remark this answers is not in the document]
belcherj: Removed

client/components/section-header/index.js
```diff
@@ -0,0 +1,66 @@
...
+		const { children, menu, title, standAlone } = this.props;
+		const className = classnames( 'woocommerce-section-header', this.props.className, {
+			'stand-alone': !! standAlone,
+		} );
+		return (
+			<div className={ className }>
+				<H className="woocommerce-section-header__title woocommerce-section-header__header-item">
+					{ title }
+				</H>
+				{ standAlone && <hr /> }
```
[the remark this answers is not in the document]
belcherj: Added role since I would have to create a div or span to do it via css anyways.

client/components/section-header/index.js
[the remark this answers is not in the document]
belcherj: Removed

client/components/card/style.scss
belcherj: Moved it back.  Just habit.

client/components/section-header/index.js
belcherj: Updated

client/components/section-header/test/index.js
belcherj: Added external dependencies so it doesn't matter either way.

client/components/section-header/test/index.js
belcherj: Added a snapshot!!

belcherj:
Ready for review again.

ryelle:
@belcherj Are all your commits pushed? I still don't see the `role="presentation"` or the updated snapshot tests.

belcherj:
Sorry just pushed the last two commits.

client/dashboard/store-performance/index.js
```diff
@@ -72,41 +73,44 @@ class StorePerformance extends Component {
...
-							delta={ 0 }
-						/>
-					) }
-				</SummaryList>
-			</Card>
+			<Fragment>
+				<SectionHeader
+					title={ __( 'Store Performance', 'wc-admin' ) }
+					menu={ this.renderMenu() }
+					standAlone
```
ryelle: This isn't needed anymore.

ryelle (approved):
One last thing, once that's fixed this is good to merge 👍🏻

ryelle:
Actually, it looks like you also need to update the test with the `role="presentation"` change as well.

belcherj:
https://jestjs.io/docs/en/snapshot-testing#updating-snapshots

Issue #631, opened 2018-10-23
Hover / sorting style bugs: tabular data


LevinMedia:
This bug relates to  #571

**the First column "sorted" background color is not being overriden by hover background color.** 

In both the revenue, and orders report, the first column is sorted by default. When hovering over the rows in the chart, the hover state isn't overriding the "sorted" background color of the first column.

Example: 
https://www.dropbox.com/s/xiaxj69hon0u29x/hoverbug.mov?dl=0

Some additional info, for what it's worth: the temporary table in the product report doesn't have any sorting by default. If you manually sort the first column, the hover is applied correctly. 

Example:
https://www.dropbox.com/s/5zk543wyug79hhx/ProductsExample.mov?dl=0

**First column sorting in conjunction with comparison check boxes**

When the first column is sorted, and the tabular data features comparison checkboxes, we should update the background color behind the checkboxes to match the sorted style. 

<img width="722" alt="screen shot 2018-10-22 at 7 06 13 pm" src="https://user-images.githubusercontent.com/4500952/47330742-315fc600-d62e-11e8-993e-991a329a8507.png">

CC: @joshuatf 

Let me know if you have any issues viewing the dropbox links - there wasn't enough contrast in the styles for the gif to render correctly.

belcherj:
On master Product title does not appear to be sortable.

belcherj:
Looks like there is an API reason Product Name is not sortable. Will update when I have more info!

belcherj:
https://app.swaggerhub.com/apis/peterfabian/wc-v3-api/1.0.0#/default/get_reports_products

The API does not allow for sorting by name.  

I think we can close this one once #706 is merged and open another issue to request API sorting by name.

timmyc:
Opened #740 to track adding `orderBy` support for product name, closing this out.

PR #706, opened 2018-10-26
Fixes hover background on first column in tables

belcherj:
#631 

When hovering also change background of first column which is a th no a td

### Screenshots

screen shot 2018-10-26 at 10 14 31 am

- Goto revenue report
- Have date column be sorted
- Hover over first row
- Does hover background overtake first column


Aljullu (approved):
Tested and it works as expected.

LGTM :rocket:

PR #707, opened 2018-10-26
Move withSelect() inside table components

Aljullu:
Fixes #703.

Move API calls inside table components instead of having them in the report index file.

This PR should have not produced any visible change, it's just a code refactoring.

### Detailed test instructions:

 - Go to the _Orders_, _Products_ and _Revenue_ reports.
 - Interact with the tables (pagination, sorting, changing time intervals, etc.) and make sure there are no regressions.

PR #713, opened 2018-10-28
Add isLoading prop to TableCard

Aljullu:
Depends on #707.

This PR adds an `isLoading` prop to `<TableCard>` which will render `<TablePlaceholder>` instead of `<Table>` when set to true.

In parallel, now it was easier to make `Table` and `TablePlaceholder` components to have the same props. So `TablePlaceholder` will be a much better approximation of what `Table` will look like when it's loaded: number of rows, actions and menus, summary, etc.

### Screenshots
Before:
imatge

After:
imatge

- Go to a page with a table (eg: `/wp-admin/admin.php?page=wc-admin#/analytics/products`).
- Verify the table placeholder appears when it's loading (including when changing pages or sorting) and it has the same actions, menus, number of rows, etc. than the resulting table.

### DIFF
--- a/client/components/table/index.js
+++ b/client/components/table/index.js
@@ -24,6 +24,7 @@ import MenuTitle from 'components/ellipsis-menu/menu-title';
 import Pagination from 'components/pagination';
 import Search from 'components/search';
 import Table from './table';
+import TablePlaceholder from './placeholder';
 import TableSummary from './summary';
 
 /**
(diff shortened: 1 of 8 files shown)


client/components/table/index.js
```diff
@@ -244,6 +246,7 @@ class TableCard extends Component {
 						<IconButton
 							key="download"
 							className="woocommerce-table__download-button"
+							disabled={ isLoading }
```
psealock: 👍

psealock (approved):
Nice improvement, it even works with removed table columns

PR #714, opened 2018-10-29
Filters: Add Variations extra filters

psealock:
Fixes https://github.com/woocommerce/wc-admin/issues/681

Add Variations filter for Product Detail
**Designs:** p6riRB-3dP-p2

screen shot 2018-10-30 at 4 59 43 pm

Changes here include a change to the filter configs because more information is required, eg url parameter "filter" v. "filter-variations" or a label "Show:" v. no label. As a result, I needed to do some renaming to avoid things like `filter.filters`.

## Testing

1. `/wp-admin/admin.php?page=wc-admin#/analytics/products`.
2. Filter to Single Product and select one from the autocomplete.
3. See the Variations filter appear.
4. Try all the filters including "Compare".
5. Make sure the url query parameters update accordingly.
6. Make sure results appear in Network requests for data.
7. Poke around for any possible regressions on other reports.

### DIFF
--- a/client/components/filters/filter/index.js
+++ b/client/components/filters/filter/index.js
@@ -44,8 +44,8 @@ class FilterPicker extends Component {
 
 		if ( selectedFilter.settings && selectedFilter.settings.getLabels ) {
 			const { query } = this.props;
-			const { param, getLabels } = selectedFilter.settings;
-			getLabels( query[ param ] ).then( this.updateSelectedTag );
+			const { param: filterParam, getLabels } = selectedFilter.settings;
+			getLabels( query[ filterParam ] ).then( this.updateSelectedTag );
 		}
 	}
 
@@ -53,10 +53,10 @@ class FilterPicker extends Component {
 		this.setState( { selectedTag: tags[ 0 ] } );
 	}
 
-	getFilter( value = false ) {
-		const { filters, query } = this.props;
-		const allFilters = flatenFilters( filters );
-		value = value || query.filter || DEFAULT_FILTER;
+	getFilter( value ) {
+		const { config, query } = this.props;
+		const allFilters = flatenFilters( config.filters );
+		value = value || query[ config.param ] || DEFAULT_FILTER;
 		return find( allFilters, { value } ) || {};
 	}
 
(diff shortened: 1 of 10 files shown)


client/components/filters/filter/index.js
```diff
@@ -207,31 +211,52 @@ FilterPicker.propTypes = {
...
+		showFilters: PropTypes.func.isRequired,
+		/**
+		 * An array of filter a user can select.
+		 */
+		filters: PropTypes.arrayOf(
+			PropTypes.shape( {
+				/**
+				 * A custom component used instead of a button, might have special handling for filtering. TBD, not yet implemented.
+				 */
+				component: PropTypes.string,
```
timmyc: If this would be a component, the proptype wouldn't be string correct? Or is this just a temporary since it has yet to be developed?

client/components/filters/filter/index.js
```diff
@@ -207,31 +211,52 @@ FilterPicker.propTypes = {
...
-			 * This set of filters is shown if the parent filter is clicked.
-			 */
-			subFilters: PropTypes.array,
-			/**
-			 * The value for this filter, used to set the `filter` query param when clicked, if there are no `subFilters`.
-			 */
-			value: PropTypes.string.isRequired,
-		} )
-	).isRequired,
+	config: PropTypes.shape( {
```
timmyc: Excellent docs, thank you.

client/store/reports/test/utils.js
```diff
@@ -343,12 +343,19 @@ describe( 'getFilterQuery', () => {
...
 	const filters = [
-		{ value: 'top_meal', query: { lunch: 'burritos' } },
-		{ value: 'top_dessert', query: { dinner: 'ice_cream' } },
-		{ value: 'compare-cuisines', settings: { param: 'region' } },
 		{
-			value: 'food_destination',
-			subFilters: [ { value: 'choose_a_european_city', settings: { param: 'european_cities' } } ],
+			param: 'filter',
+			filters: [
+				{ value: 'top_meal', query: { lunch: 'burritos' } },
```
timmyc: 🌯

timmyc (approved):
Exciting stuff here Paul! This was testing out well for me, and in my playing around with filters in other reports I couldn't find any regressions. LGTM.

Issue #663, opened 2018-10-24
Changing chart interval on orders, or product report always switches back to revenue report


LevinMedia:
to reproduce: 

On master, navigate to the orders report
Change the chart interval from day to week
Note the report reloads as the revenue report, with the selected chart interval applied.

![screen recording 2018-10-23 at 06 49 pm](https://user-images.githubusercontent.com/4500952/47401151-fd9ba380-d6f4-11e8-91a9-d2e334298cb1.gif)

belcherj:
Dupe:  https://github.com/woocommerce/wc-admin/issues/704

belcherj:
I am not able to replicate.

Aljullu:
@belcherj

> I am not able to replicate.

Right, the fix from #717 was needed to replicate this issue.

I will assign this issue to myself, if that's ok for you! :slightly_smiling_face:

PR #568, opened 2018-10-18
Fix/date params persistence

psealock:
Fixes https://github.com/woocommerce/wc-admin/issues/455
Fixes https://github.com/woocommerce/wc-admin/issues/554

* Changed the default period to `month`
* Changed the default compare to `previous_year`
* Changed links in the wp-admin sidebar such that current time-related query parameters persist when switching reports.

### Test

1. Go to any report and see the page load Month to date and Previous Year for date selection.
2. Choose a new date preset.
3. See the links in the sidebar now persist only the associated query parameters.
4. Test on Dashboard and do a whole bunch of navigation.
5. Make sure the current screen is highlighted on the menubar

PR #717, opened 2018-10-29
Fix: date params were no longer persisted when navigating pages

Aljullu:
When trying to reproduce #663, I noticed date params persistence was broken in master. This PR should fix that (some changes to the selector that highlights the links was required too).

- Go to any page of `wc-admin`.
- Change the time range and verify the links are still correctly highlighted.
- Navigate to another page and verify the time range persisted and links are correctly highlighted too.

cc @belcherj you did some modifications to that part of the code recently, so you might want to take a look and make sure I'm not breaking anything new :slightly_smiling_face:

### DIFF
--- a/client/layout/controller.js
+++ b/client/layout/controller.js
@@ -21,31 +21,31 @@ const getPages = () => {
 			container: Dashboard,
 			path: '/',
 			wpOpenMenu: 'toplevel_page_woocommerce',
-			wpClosedMenu: 'toplevel_page_wc-admin--analytics',
+			wpClosedMenu: 'toplevel_page_wc-admin--analytics-revenue',
 		},
 		{
 			container: Analytics,
 			path: '/analytics',
-			wpOpenMenu: 'toplevel_page_wc-admin--analytics',
+			wpOpenMenu: 'toplevel_page_wc-admin--analytics-revenue',
 			wpClosedMenu: 'toplevel_page_woocommerce',
 		},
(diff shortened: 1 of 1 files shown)


ryelle:
I'm not sure if I'm reading the problem correctly– we want the selected date range to persist between different reports? So if I'm on the Revenue report, and I pick "last year", then go to the Products report, it should stay "last year"?

Aljullu:
@ryelle Exactly, that was implemented in #568, but it got broken recently when we did some changes in the menu.

Getting in some more detail, I found that [this selector](https://github.com/woocommerce/wc-admin/pull/717/files#diff-5c8a66f908738b94586c976c499edfe5R73) which updates the navigation menu URLs stopped working for links under _Analytics_ because their `wpOpenMenu` id was `toplevel_page_wc-admin--analytics` but the id of the menu element was `toplevel_page_wc-admin--analytics-revenue`, so they were never selected and therefore never updated with the time parameters.

allendav:
I see .current correctly moving from item to item under the Analytics menu as I use the browser's forward and back buttons, but I don't understand how this PR fixes the date params. Can you explain?

allendav (approved):
Tests well for me!

Aljullu:
> Nevermind - I see it now - needed to expand the li to see the a to see the href change in dev tools.

Exactly! It works changing the `href` of the links.

PR #720, opened 2018-10-29
Fix automattically fixable php lint warnings

belcherj:
Ran the PHPCBF autofixing PHP Lint script.  Just some minor spacing fixes.

 - Just spacing.  I do not think there is a need to test.


Aljullu (approved):
Looks good to me! :+1:

PR #729, opened 2018-10-29
Stylelint autofixes

belcherj:
This PR used Stylelint auto-fixing command to clean up all the lints possible automatically

 - Hard to test other than looking at the UI for css issues.

### DIFF
--- a/client/components/calendar/style.scss
+++ b/client/components/calendar/style.scss
@@ -117,10 +117,10 @@
 
 		.woocommerce-calendar__input-text {
 			border: 1px solid $error-red;
-			box-shadow: inset 0px 0px 8px $error-red;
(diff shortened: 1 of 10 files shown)


Aljullu:
When I run `npm run lint:css-fix` I get some other fixes (eg: `'` replaced with `"`), shouldn't those be included in this PR too?

belcherj:
They should be but prettier changes them back when committing.  I think some rules need to be modified.

Aljullu:
> They should be but prettier changes them back when committing. I think some rules need to be modified.

Oh, ok. Makes sense.

Good to go, then! :rocket:

Aljullu (approved):

Issue #704, opened 2018-10-26
Bug: switching intervals switches between analytics pages


greenafrican:
**Describe the bug**
When changing intervals in the chart, I noticed that the entire page changed from the revenue report to the product report.

**To Reproduce**

1. Go to the revenue report, change the interval to 'by week'
2. Go to products report
3. Select product comparison and 3 products
4. Go back to revenue report and change interval to 'by week'
5. You get taken to the product comparison page

**Expected behavior**
Stay on the revenue report and change the interval to 'by week'

**Screenshots**
![change](https://user-images.githubusercontent.com/1160732/47566387-df6bb880-d92b-11e8-8cd3-4d454d10e6cf.gif)

**Desktop (please complete the following information):**

* OS: MacOS 10.13.6 
* Browser: Chrome 
* Version 69.0.3497.100

**Additional context**

belcherj:
Dupe?  https://github.com/woocommerce/wc-admin/issues/663

belcherj:
I am not able to replicate.

Aljullu:
Closing since it should have been fixed by #731.

PR #731, opened 2018-10-30
Fix charts redirecting to previous path when changing interval value

### LINKED ISSUE #663: Changing chart interval on orders, or product report always switches back to revenue report

Aljullu:
Fixes #663 and #704.

I have to say I don't understand yet where that bug came from. The problem was that calling `getPath()` (which is called by `updateQueryString()`) inside the `<Chart>` component returns the previous location instead of the current one.

The solution I found is passing down the `path` prop and using that one to generate the new route when changing the interval. We are also passing down the path for `<ReportFilters>` and React Router [docs](https://reacttraining.com/react-router/web/api/history/history-is-mutable) also seem to recommend using the `path` prop instead of relying on the history object to get the current path, so I think the proposed solution is good.

But if somebody has a better understanding of React Router/this specific problem or has any idea of a better solution, feel free to propose it. :slightly_smiling_face: 

- On master, go to the _Revenue_ report.
- Navigate to the _Orders_ report.
- Change the chart interval from _day_ to _week_.
- Note the report **no longer** reloads as the _Revenue_ report.

### DIFF
--- a/client/analytics/components/report-chart/index.js
+++ b/client/analytics/components/report-chart/index.js
@@ -79,6 +79,7 @@ class ReportChart extends Component {
 		return (
 			<Chart
 				path={ path }
+				query={ query }
 				data={ chartData }
 				title={ selectedChart.label }
 				interval={ currentInterval }
--- a/client/analytics/components/report-chart/index.js
+++ b/client/analytics/components/report-chart/index.js
@@ -29,7 +29,7 @@ import ReportError from 'analytics/components/report-error';
 
 class ReportChart extends Component {
 	render() {
-		const { primaryData, secondaryData, selectedChart, query } = this.props;
+		const { path, primaryData, secondaryData, selectedChart, query } = this.props;
 
 		if ( primaryData.isError || secondaryData.isError ) {
 			return <ReportError isError />;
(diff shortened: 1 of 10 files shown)


LevinMedia:
@Aljullu thanks for jumping on this one

I've set my time period to year to date. Switching to the orders report, the orders summary number is selected by default., Switching the interval to month behaves as expected. We're staying on the orders report now. 🎉 

So far so good... however... the plot thickens. 

if I change the selected summary number to anything else: net revenue, aov, average items per order, then change the interval, the summary number jumps back to orders:

screen recording 2018-10-30 at 09 56 am

Something similar happens when you switch to a new summary number, then activate the advanced filters... it switches back to the default summary number: orders. 

screen recording 2018-10-30 at 09 59 am

LevinMedia:
Thanks @Aljullu looking good!

timmyc (approved):
Looks good and testing well for me.

PR #733, opened 2018-10-30
Makes stylelint config match prettier settings

belcherj:
Prettier transforms double quotes to single quotes and stylelint fix was transforming single quotes to double quotes.   Stylelint is easier to configure and prettier format is the current standard so I changed Stylelint.

 - Make a single quote a double.
 - Run `npm run lint:css`
 - Observe error
 - Run `npm run lint:css-fix` to fix error

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -14,6 +14,7 @@
 		"rule-empty-line-before": null,
 		"selector-class-pattern": null,
 		"string-quotes": "single",
-		"value-keyword-case": null
+		"value-keyword-case": null,
+		"value-list-comma-newline-after": null
 	}
 }
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -5,6 +5,7 @@
 		"at-rule-no-unknown": null,
 		"comment-empty-line-before": null,
 		"declaration-block-no-duplicate-properties": null,
+		"declaration-colon-newline-after": null,
 		"declaration-property-unit-whitelist": null,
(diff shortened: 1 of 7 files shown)


Aljullu (approved):
Looks good, I also prefer single quotes in CSS. :slightly_smiling_face: 

When I run `npm run lint:css` I still get some errors, is that to be expected?
imatge

ryelle:
Since @belcherj is AFK, I've taken over the PR & pushed fixes to those remaining linting errors:

- `declaration-colon-newline-after` is another conflict with prettier, so I've disabled it
- The other issue in `split-button` needed to be fixed
- I've disabled `block-closing-brace-newline-after` in `_breakpoints.scss` since obeying it made the code look strangely-spaced -- it's a rule for real CSS, this file is more like a huge conditional.

ryelle:
🙃 Prettier reformatted the `split-button` fix, so that's great.

ryelle:
I've disabled `value-list-comma-newline-after` now. It's a good rule in theory, but it happens that prettier decides to wrap based on line-length, while stylelint wants a new line after every comma. Since this would probably keep happening, I've disabled the rule and we'll follow prettier's lead.

Note also that the `fix` command still fixes comment-ignored files (there's an issue about it https://github.com/stylelint/stylelint/issues/2643 ), so that keeps trying to fix the breakpoints file.

Aljullu:
@ryelle thanks for picking this up! It looks good to me. :+1:

PR #734, opened 2018-10-30
Fix/all css lints

belcherj:
Fixes all CSS lints except one that gets ignored because stylelint does not understand how to handle it.

Only way to test is to look at the site and make sure css renders correctly.


timmyc (approved):
Thanks JB! LGTM

PR #735, opened 2018-10-30
Adds css linting to the main lint command to be run in ci

belcherj:
Adds Stylelint to main linting command.  This will cause commit hooks and ci to fail if you are not fully linted!


Aljullu:
I see travis is also complaining about these issues: https://github.com/woocommerce/wc-admin/pull/733#pullrequestreview-170226253

I think they are added by the CSS prettier as well.

Would it be easy to fix those ones (or changing the config either in the prettier or the CSS Lint config) before merging this PR to master? Otherwise all other PRs will get travis warnings due to these CSS lines.

Aljullu (approved):
Now that #733 has been merged, it should be safe to merge this one too.

Issue #682, opened 2018-10-24
Compare Filter: UI Improvements


psealock:
When comparing products, the compare button could use improvements:

1. When clicking "Compare" query parameters from the previous filter should be cleared out
2. The "Compare" button should be disabled when the current selection is already represented in the url.

![screen shot 2018-10-25 at 12 44 54 pm](https://user-images.githubusercontent.com/1922453/47467666-11a9e700-d854-11e8-8279-5b2ce7acea0c.png)

ryelle:
> The "Compare" button should be disabled when the current selection is already represented in the url.

Why? We say "select at least two products to compare", and then show the currently selected items in the filter. If the button is disabled now (ex, in your screenshot), that text is confusing (two items _are_ selected). Does it matter that the button doesn't do anything when clicked?

psealock:
>Does it matter that the button doesn't do anything when clicked?

When using the UI, it felt odd to me that clicking an action button did nothing and I had to think about why that might be. The Compare UI can be in the exact state shown in my screen shot, but the report can be in two different states (filtered, not filtered). I thought it would be a nice way to indicate the comparison has already been applied.

I see what you're saying about the text, however. We could change the text to match the context. @LevinMedia what do you think?

LevinMedia:
I think the biggest affordance to whether or not the comparison has been applied or not hasn't actually been implemented into this report yet. That is... once the comparison is made (a user clicks on the comparison button), the graph will switch to the comparison version with the compared data points in in the legend (per https://github.com/woocommerce/wc-admin/issues/695 ) Additionally, once a comparison is made, the tabular data and summary numbers will update to only include what's being compared. 

One more thing we could do to clarify things: The "Clear All" link is always visible - I think it would be good to hide that until a comparison is made. Clicking "clear all" would return the report back to it's original state, and the clear all link would disappear. How's that sound?

Aljullu:
@psealock I'm not sure to understand this point:
> When clicking "Compare" query parameters from the previous filter should be cleared out

Isn't that already happening?

psealock:
>Isn't that already happening?

Hmm, I can't reproduce what I was referring to. So, yes.

Thanks for the explanation @LevinMedia. https://github.com/woocommerce/wc-admin/pull/736 fixes this issue then.

PR #736, opened 2018-10-30
Hide Clear filter button when no filters are selected

Aljullu:
Part of #682.

This PR adds several fixes:
- Hides the _Clear all_ button in _Compare Products_ card (Products report) when there are no products selected.
- Hides the _Clear all filters_ button in _Advanced filters_ card (Orders report) when there are no filters selected.
- Updates the _Clear all_ button in _Compare Products_ card (Products report) so it clears the selected products even when the comparison was not done yet (aka the user pressed _Clear all_ before pressing _Compare_).

### Screenshots
`/wp-admin/admin.php?page=wc-admin#/analytics/products?filter=compare-product`
imatge

`/wp-admin/admin.php?page=wc-admin#/analytics/orders?filter=advanced`
imatge

- Go to `/wp-admin/admin.php?page=wc-admin#/analytics/products?filter=compare-product` and verify there isn't a _Clear all_ button visible.
- Add a product using the search box.
- Verify the _Clear all_ button appears.
- Click on it.
- Verify the selected product and the _Clear all_ button disappeared.

---

- Go to `/wp-admin/admin.php?page=wc-admin#/analytics/orders?filter=advanced` and verify there isn't a _Clear all filters_ button visible.
- Add a filter using the _Add a filter_ button.
- Verify the _Clear all filters_ button appears.
- Click on it.
- Verify the filter and the _Clear all filters_ button disappeared.

### DIFF
--- a/client/components/filters/advanced/index.js
+++ b/client/components/filters/advanced/index.js
@@ -229,9 +229,11 @@ class AdvancedFilters extends Component {
 							{ __( 'Filter', 'wc-admin' ) }
 						</Link>
 					) }
-					<Link type="wc-admin" href={ this.getUpdateHref( [] ) } onClick={ this.clearFilters }>
-						{ __( 'Clear all filters', 'wc-admin' ) }
-					</Link>
+					{ activeFilters.length > 0 && (
+						<Link type="wc-admin" href={ this.getUpdateHref( [] ) } onClick={ this.clearFilters }>
+							{ __( 'Clear all filters', 'wc-admin' ) }
+						</Link>
+					) }
 				</div>
 			</Card>
 		);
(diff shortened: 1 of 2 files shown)


belcherj:
Getting some history state errors on the filters: 
screen shot 2018-10-30 at 1 48 09 pm

I thinks it is okay to merge with these as that console.warn should probably be removed.

belcherj (approved):
Change looks good, tested and works

LevinMedia:
Looking good, Thanks @Aljullu !

Aljullu:
> I think it is okay to merge with these as that console.warn should probably be removed.

Agree. We could add a comparison in `updateQueryString()` to avoid pushing the same path, but I think hiding the warning would also be ok.

PR #746, opened 2018-10-31
Update summary number focus styles

LevinMedia:
Per a design review with @josemarques we've to decided to make the focus styles on the summary numbers less intense. The Weight has been reduced from 2px to 1px and the color from `$black` to `$core-grey-dark-500`

Additionally, when a summary number is selected, the highlight at the top of the container becomes 5px rather than 4px to offset the weight of the inset focus.

### Screenshots

**Before**

screen recording 2018-10-31 at 11 19 am

**After**

screen recording 2018-10-31 at 11 13 am

### DIFF
--- a/client/components/summary/style.scss
+++ b/client/components/summary/style.scss
@@ -340,7 +340,7 @@ $inner-border: $core-grey-light-500;
 
 	&:focus {
 		// !important to override button styles
-		box-shadow: inset -2px -2px 0 $black, inset 2px 2px 0 $black !important;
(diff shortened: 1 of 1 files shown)


timmyc (approved):
LGTM

josemarques:
@LevinMedia any reason why the hover state has a thinner top border? I think it makes sense to keep it the same width as active state.

LevinMedia:
After chatting with @josemarques this am, I'm going to close this PR and open another from a different branch.

PR #747, opened 2018-10-31
Run dbdelta on init, at least as long as this is a separate plugin

### LINKED ISSUE #531: Notes: Tables Not Setup on Branch Switch / dbdelta / database
While testing #507 - the new database tables for housing all the new inbox/notfications data did not automagically get setup for me. I had to deactivate/reactivate the plugin for the db changes to be applied. The following PHP error was being thrown otherwise when accessing the new `/wp-json/wc/v3/admin/notes` endpoint:

```
[15-Oct-2018 18:44:57 UTC] WordPress database error Table 'example.wp_woocommerce_admin_notes' doesn't exist for query 
            SELECT note_id, title, content
            FROM wp_woocommerce_admin_notes
            ORDER BY note_id DESC
         LIMIT 0, 10 made by require('wp-blog-header.php'), wp, WP->main, WP->parse_request, do_action_ref_array, WP_Hook->do_action, WP_Hook->apply_filters, rest_api_loaded, WP_REST_Server->serve_request, WP_REST_Server->dispatch, WC_Admin_REST_Admin_Notes_Controller->get_items, WC_Admin_Notes::get_notes, WC_Data_Store->__call, WC_Admin_Notes_Data_Store->get_notes
[15-Oct-2018 18:44:57 UTC] WordPress database error Table 'example.wp_woocommerce_admin_notes' doesn't exist for query 
            SELECT COUNT(*)
            FROM wp_woocommerce_admin_notes
         made by require('wp-blog-header.php'), wp, WP->main, WP->parse_request, do_action_ref_array, WP_Hook->do_action, WP_Hook->apply_filters, rest_api_loaded, WP_REST_Server->serve_request, WP_REST_Server->dispatch, WC_Admin_REST_Admin_Notes_Controller->get_items, WC_Admin_Notes::get_notes_count, WC_Data_Store->__call, WC_Admin_Notes_Data_Store->get_notes_count
```

__To Recreate__
- Checkout the branch in #510, on a test site where the plugin is already activated
- Follow the steps in #510 to test out the new endpoint

allendav:
Fixes #531 

WooCommerce has a robust updater that we will be leveraging when this codebase becomes part of it. For now, let's run dbdelta on init.

 - Check out the master branch.
 - Modify the database schema in class-wc-admin-api-init.php in some way, e.g. change the name column in woocommerce_admin_note_actions from varchar(255) to varchar(200)
- Deactivate and activate the wc-admin plugin to get dbdelta to run
- Launch mysql and describe the woocommerce_admin_note_actions table. Make sure woocommerce_admin_note_actions is now varchar(200)
- Check out this branch
- Navigate to anywhere on the site
- Describe the woocommerce_admin_note_actions table again. Make sure the name column is back to  varchar(255)

```
mysql> use wordpress1;
Reading table information for completion of table and column names
You can turn off this feature to get a quicker startup with -A

Database changed
mysql> describe wca_woocommerce_admin_notes;
+---------------+---------------------+------+-----+---------------------+----------------+
| Field         | Type                | Null | Key | Default             | Extra          |
+---------------+---------------------+------+-----+---------------------+----------------+
| note_id       | bigint(20) unsigned | NO   | PRI | NULL                | auto_increment |
| name          | varchar(200)        | NO   |     | NULL                |                |
| type          | varchar(20)         | NO   |     | NULL                |                |
| locale        | varchar(20)         | NO   |     | NULL                |                |
| title         | longtext            | NO   |     | NULL                |                |
| content       | longtext            | NO   |     | NULL                |                |
| icon          | varchar(200)        | NO   |     | NULL                |                |
| content_data  | longtext            | YES  |     | NULL                |                |
| status        | varchar(200)        | NO   |     | NULL                |                |
| source        | varchar(200)        | NO   |     | NULL                |                |
| date_created  | datetime            | NO   |     | 0000-00-00 00:00:00 |                |
| date_reminder | datetime            | YES  |     | NULL                |                |
+---------------+---------------------+------+-----+---------------------+----------------+
12 rows in set (0.00 sec)

```

### DIFF
--- a/wc-admin.php
+++ b/wc-admin.php
@@ -93,6 +93,16 @@ function deactivate_wc_admin_plugin() {
 register_deactivation_hook( WC_ADMIN_PLUGIN_FILE, 'deactivate_wc_admin_plugin' );
 
 /**
+ * Update the database tables if needed. This hooked function does NOT need to
+ * be ported to WooCommerce's code base - WC_Install will do this on plugin
+ * update automatically.
+ */
+function wc_admin_init() {
+	WC_Admin_Api_Init::create_db_tables();
+}
+add_action( 'init', 'wc_admin_init' );
+
+/**
  * Set up the plugin, only if we can detect both Gutenberg and WooCommerce
  */
 function wc_admin_plugins_loaded() {
--- a/wc-admin.php
+++ b/wc-admin.php
@@ -98,7 +98,10 @@ register_deactivation_hook( WC_ADMIN_PLUGIN_FILE, 'deactivate_wc_admin_plugin' )
  * update automatically.
  */
 function wc_admin_init() {
-	WC_Admin_Api_Init::create_db_tables();
+	// Only create/update tables on init if WP_DEBUG is true.
+	if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {
+		WC_Admin_Api_Init::create_db_tables();
+	}
 }
 add_action( 'init', 'wc_admin_init' );
 


wc-admin.php
```diff
@@ -92,6 +92,16 @@ function deactivate_wc_admin_plugin() {
 }
 register_deactivation_hook( WC_ADMIN_PLUGIN_FILE, 'deactivate_wc_admin_plugin' );
 
+/**
+ * Update the database tables if needed. This hooked function does NOT need to
+ * be ported to WooCommerce's code base - WC_Install will do this on plugin
+ * update automatically.
+ */
+function wc_admin_init() {
+	WC_Admin_Api_Init::create_db_tables();
```
timmyc: Is there any issue with running this on every `init`? Or is this done in woo core too?

allendav: It's not a long term solution. We won't want to merge this bit to core (hence the comment). I can open an issue to specifically remove this then.

FWIW, core has the same problem - dbdelta is only run on version updates.

I just had another idea - what if we test for some define (e.g. WC_ADMIN_DEV) and run it on init only when that is present - i.e. just in case this accidentally makes it into the core merge.

timmyc: Yeah I like that idea of adding a sanity check - maybe just do WP_DEBUG check. I figure most beta testers will never hit this issue - just those of us that are swapping out git branches often.

allendav: Fixed in https://github.com/woocommerce/wc-admin/pull/747/commits/83f9a1859cdc2721f063a0c58c216780525d7848

PR #748, opened 2018-10-31
Add markdown docs with details on contributing

allendav:
Patterned after what we did for WooCommerce Services.


CONTRIBUTING.md
```diff
@@ -0,0 +1,26 @@
+# Contributing to WooCommerce Admin
+
+Hi! Thank you for your interest in contributing to WooCommerce Admin. We appreciate it.
```
psealock: Are we sticking with `wc-admin` as a name? If not, maybe best to change this to `wc-admin` so when it gets changed this isn't overlooked.

CONTRIBUTING.md
```diff
@@ -0,0 +1,26 @@
...
+
+## License
+
+WooCommerce Admin is licensed under [GNU General Public License v2 (or later)](/LICENSE.md).
+
+All materials contributed should be compatible with the GPLv2. This means that if you own the material, you agree to license it under the GPLv2 license. If you are contributing code that is not your own, such as adding a component from another Open Source project, or adding an `npm` package, you need to make sure you follow these steps:
+
+1. Check that the code has a license. If you can't find one, you can try to contact the original author and get permission to use, or ask them to release under a compatible Open Source license.
+2. Check the license is compatible with [GPLv2](http://www.gnu.org/licenses/license-list.en.html#GPLCompatibleLicenses), note that the Apache 2.0 license is *not* compatible.
+3. Add the code source URL (e.g. a GitHub URL), the files where it's used in the WooCommerce Services and the full license terms to [`CREDITS.md`](/CREDITS.md)
```
psealock: WooCommerce Services -> wc-admin (or WooCommerce Admin)

psealock (approved):
LGTM

CONTRIBUTING.md
psealock: Nevermind... the plugin is titled WooCommerce Admin anyways 😄

CONTRIBUTING.md
```diff
@@ -0,0 +1,26 @@
+# Contributing to WooCommerce Admin
+
+Hi! Thank you for your interest in contributing to WooCommerce Admin. We appreciate it.
+
+There are many ways to contribute – reporting bugs, feature suggestions, and fixing bugs.
+
+## Reporting Bugs, Asking Questions, Sending Suggestions
+
+Open [a GitHub issue](https://github.com/woocommerce/woocommerce-admin/issues/), that's all. If you want to prefix the title with a “Question:”, “Bug:”, or the general area of the application, that would be helpful but is no means mandatory. If you have write access, add the appropriate labels.
```
timmyc: Incorrect link here, should be `https://github.com/woocommerce/wc-admin/issues/new/choose`

timmyc: Also you can remove the prefix bit since @belcherj made some great templates for bugs vs feature requests.

timmyc (commented):
Couple of small tweaks, otherwise looking good. Thanks for getting this up Allen.

CONTRIBUTING.md
allendav: Fixed in https://github.com/woocommerce/wc-admin/pull/748/commits/8d827876244a21487add21a0e78ef8c879bcd300

CONTRIBUTING.md
allendav: Fixed in https://github.com/woocommerce/wc-admin/pull/748/commits/8d827876244a21487add21a0e78ef8c879bcd300

CONTRIBUTING.md
allendav: Ultimately, a lot of this sort of documentation will go away or be absorbed into WooCommerce's repo when we merge this into that plugin, so the name "WooCommerce Admin" won't really be "stuck" with long term - but for now if we say "WooCommerce Admin" - we match the rest of this repo.

justinshreve: > Are we sticking with wc-admin as a name? 

The admin classes in Woo Core are already WC_Admin_* -- so we picked a future compatible name for code/classes :).

CREDITS.md
```diff
@@ -0,0 +1,52 @@
+Credits
+=======
+
+This project makes use of Open Source components. Below is a list of these components included in this project's source code, and their license information. This project also uses js packages released by NPM, see [package.json](/package.json). Source code and license information for each of these packages is available at https://npmjs.org. Many thanks to all of the original authors!
+
+### https://github.com/facebook/react
```
ryelle: We shouldn’t be including either of these libraries in the built files, so I don’t think we need this

allendav: Oh? Is that because we're relying on Gutenberg to bring them over?

allendav: Removed credits for now, until we actually need them, see https://github.com/woocommerce/wc-admin/pull/748/commits/ecc43a60ec78db2cf0d87c003be7a6674408f3dc

ryelle: Yeah, anything react should be provided by `@wordpress/element` so we're not loading multiple Reacts. It looks like we're not using `react-redux` at all, either (just `redux`).

PR #755, opened 2018-11-01
Build: Switch react imports to use core packages

ryelle:
We don't need to import from external `react`, `react-dom`, or `redux`, since these are exposed through `@wordpress/*` packages. 

The import changes are:

- remove importing React for JSX support. The [`babel-plugin-import-jsx-pragma`](https://github.com/WordPress/gutenberg/tree/master/packages/babel-plugin-import-jsx-pragma) handles this for us, we don't need to import React to use JSX.
- replace imports from `react` or `react-dom` with the equivalent from `@wordpress/element`.
- replace imports from `redux` with the equivalent from `@wordpress/data`.

I've also split out `package.json`'s dependencies and devDependencies - we had a lot of devDependencies that are required for the app, not just for building. This will make it easier to pull out dependencies for lerna packages (if needed).

**To test**

- Check that the project builds, and that no app functionality has changed.


allendav (approved):
Works fine. npm run build, npm start, npm test all good to go. I also get this in the browser console, but I also get it on master, so :shipit: 

```
Jed localization error: 
Error: Domain `wc-admin` was not found.
```

PR #758, opened 2018-11-02
PHPUnit testing improvements

### LINKED ISSUE #751: Fix PHPUnit version in composer
PHP unit tests don't run with PHPUnit v7, so updating composer to run with v6, similarly to WooCommerce.

peterfabian:
Fixes #751 

Composer was configured incorrectly to use PHPUnit v7, fixed that detail.
Additionally, moved wc-admin install to PHPUnit bootstrap so that it's easier and less confusing to run unit tests locally and in Travis.

One of the problems encountered is that composer install would only work with PHP 7.1+ (because of dependencies on WooCommerce). Unless WC is fixed, I don't think it makes sense to hack around in wc-admin.

Check if wikipage instructions work: https://github.com/woocommerce/wc-admin/wiki/Running-PHP-unit-tests


allendav:
Could we make the instructions on the wiki page a setup script?

allendav:
FWIW, I checked `/Users/allendav/.composer/vendor/bin` to see if I had accidentally installed phpunit globally and did not see it there either, so 🤷‍♂️

allendav:
Noteworthy: The line

`export DB_NAME=wc-admin_tests`

is no bueno for me - the script was unable to drop/create the database successfully on repeated runs of the script.  Changing it to the following (underscore, not hyphen) worked:

`export DB_NAME=wc_admin_tests`

peterfabian:
Thanks for testing Allen. Updated the wiki page, please check.

PR #759, opened 2018-11-02
Persist time queries when clicking on breadcrumbs links

### LINKED ISSUE #718: Dates are not persisted when using the breadcrumbs
**To Reproduce**

1. Go to `/wp-admin/admin.php?page=wc-admin#/analytics/orders`.
2. Change the time range.
3. Click on _Analytics_ in the breadcrumb.

Expected: time range is persisted.

Actual: time range is lost.

**Screenshots**
imatge

Aljullu:
Fixes #718.

### Screenshots
_(this PR shouldn't have produced any visible change)_
imatge

- Go to `/wp-admin/admin.php?page=wc-admin#/analytics/products` and change the time period.
- Click on _Analytics_ in the breadcrumbs of the top bar.
- Verify the time period is still the same.

### DIFF
--- a/client/lib/nav-utils/index.js
+++ b/client/lib/nav-utils/index.js
@@ -7,6 +7,11 @@ import { parse, stringify } from 'qs';
 import { isEmpty, pick, uniq } from 'lodash';
 
 /**
+ * Internal dependencies
+ */
+import * as navUtils from './index';
+
+/**
  * Returns a string with the site's wp-admin URL appended. JS version of `admin_url`.
  *
  * @param {String} path Relative path.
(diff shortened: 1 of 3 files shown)


client/lib/nav-utils/index.js
```diff
@@ -6,6 +6,11 @@ import history from 'lib/history';
 import { parse, stringify } from 'qs';
 import { isEmpty, pick, uniq } from 'lodash';
 
+/**
+ * Internal dependencies
+ */
+import * as navUtils from './index';
```
psealock: Below we are using `getQuery` directly. I didn't know you could do this. I think its easier to see if used directly (relies on hoisting).

psealock: Oh is this for `jest.mock` to work. Maybe add a comment?

psealock (approved):
Testing well!

client/lib/nav-utils/index.js
Aljullu: Yes, exactly, this seems to be the only way to mock functions from the same module.

Added a comment in 6dd3c7f9c452cd29f7815020bb3be2bdb2f23d7f.

PR #761, opened 2018-11-02
Switch to using `@wordpress/viewport` for responsive screen layouts

ryelle:
This PR removes our `lib/ui` functions in favor of the core viewport library. This package listens and responds to screen size changes using JS + media queries ([`MediaQueryList`](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList)). It exposes this in a store called `core/viewport`, along with an HOC called `withViewportMatch`, so that we can pass booleans based on screen size as props. This also updates when the screen size is changed, without having to hook into the `resize` event (and poorly affecting performance).

For example, in DateRange, we can say

```js
export default withViewportMatch( {
	isViewportMobile: '< medium', // smaller than 782px
	isViewportSmall: '< small', // smaller than 600px
} )( DateRange );
```

And now we have the props `isViewportMobile` & `isViewportSmall` to use in the component.

-------

**Screenshots**

We used `1100px` and `1365px` as breakpoints originally, but these are not core breakpoints, so they're not supported by `@wordpress/viewport`. I've had to update where these breakpoints are used. `1365px` was changed to `1440px`. `1100px` was usually changed to `960px`, but in some cases that was too tight, so it's `1280px` there (devdocs & filters).

The SummaryNumbers switch to "mobile view" at 960px now:

960-numbers

The following screenshots are at `961px` wide, so this is the smallest screen before the `960` switch (it's unlikely anyone will see the app at this size, 1024 or 960 are more likely than anything in between). 

The filters are broken onto 1-per-line at 1280, but SummaryNumbers are not collapsed until below 960 (above). The dashboard also stays two cols until 960 - with just one item it seems like we'd want to make it full-width, but I imagine other items will end up here. But maybe we want to make that breakpoint 1280?

961-filters
961-dashboard

This applies to embedded screens, since the header shrunk down a little at 1100px before. Now it shrinks at 960, so 961 looks like:

screen shot 2018-11-02 at 11 18 45 am

The switch up to a larger number of SummaryNumbers is now 1440, so at 1441px wide we flip to this layout:

1441-numbers

-------

**To test**

- Check that screens look OK at different screen sizes.


client/components/filters/style.scss
```diff
@@ -10,7 +10,7 @@
 	display: flex;
 	margin-bottom: $gap-large;
 
-	@include breakpoint( '<1100px' ) {
+	@include breakpoint( '<1280px' ) {
```
Aljullu: Not for this PR, but would it make sense at some point using `<wide` instead of `<1280px` in the CSS as we do in the JS? It looks weird to me that in the JS we use names for the breakpoints but in the CSS we use pixel values.

Aljullu (approved):
Nice improvement. :+1:  Tested and looking good for me.

I noticed in the function `calculateMaxXTicks()` ([#](https://github.com/woocommerce/wc-admin/blob/master/client/components/chart/utils.js#L217)) inside `client/components/chart/utils.js` we are using the breakpoint `1365` and some other breakpoints which are not used anywhere else. Would it make sense to modify those ones too or should we leave them like they are right now?

ryelle:
> I noticed in the function calculateMaxXTicks() (#) inside client/components/chart/utils.js we are using the breakpoint 1365 and some other breakpoints which are not used anywhere else. 

Yeah... I'm not familiar enough with the chart code to feel comfortable editing that, and since it's not directly layout related I thought it would be OK to leave those as "special cases".

client/components/filters/style.scss
ryelle: That's possible, we can update `breakpoint` to use the names from the viewport, but I'm not sure it's really necessary. Designers would likely tell us "break at 1280px", then we'd have to go look up which name that is, and vice-versa. We could revisit it in the future, maybe.

LevinMedia (approved):
Hi @ryelle ! this is Looking good. Things might get a little tight right before the 960 breakpoint if stores have summary numbers with large values. In the screen shot below if there was one extra character in either the summary or the percentage change, there would be a collision. 

screen shot 2018-11-02 at 10 04 52 am


**BUT** to your point, I think that's okay, because most folks will either see 960 or 1024, so I'm good to go on this one.

psealock (approved):
Testing well, this is nice improvement and alignment with Gutenberg.

>The dashboard also stays two cols until 960 - with just one item it seems like we'd want to make it full-width, but I imagine other items will end up here. But maybe we want to make that breakpoint 1280?

Yeah, as the dashboard blocks get written, I think we'll find 1280 a better breakpoint.

PR #763, opened 2018-11-02
Build: Move `lib/nav-utils` and `lib/history` to `@woocommerce/navigation`

ryelle:
Move the navigation utilities to a new package managed by lerna. This will now be `@woocommerce/navigation`. This combines `nav-utils` and `history`, since `history` is used in nav-utils, and should be provided from the package so that we're only using one history object.

There should be no functionality change. 

I tried something new with creating [the readme](https://github.com/woocommerce/wc-admin/blob/38662ad1c4d25cd85ca3826b6ee6f9ad3b27690d/packages/navigation/README.md), generating the docs using [`jsdoc-to-markdown`](https://github.com/jsdoc2md/jsdoc-to-markdown/) and then manually finessing it. It generates a nice API reference - once more packages are moved over, we might want to pick up an automated docs process like the component docs.

To test:

- Build the project, make sure it succeeds.
- Navigation utils are used all over, so check that nothing looks obviously broken
- For a more specific test, check that reports still work- view a few reports, update date ranges, manipulate the table sorting, etc.


belcherj (approved):
Tested, looks good.  Reviewed code, looks good.  Looks to be some conflicts.  Merge away when fixed.

ryelle:
Thanks! will merge when the tests pass.

PR #764, opened 2018-11-02
Product Detail: Add Variations Table

psealock:
Fixes https://github.com/woocommerce/wc-admin/issues/762

**Designs:** p6riRB-3dP-p2

Adds a table for displaying variations of a Product in Product Detail.

**Note:** The endpoint for variations is not yet available. Updates/discussion in p7bje6-1c8-op2

### Test

1. `/wp-admin/admin.php?page=wc-admin#/analytics/products`
2. Filter by single product. Any product will do because the endpoint (SwaggerHub) returns example data.
3. See the Variations table appear with a single row of example data.
4. Check that the links are correct.

### DIFF
--- a/client/analytics/report/products/table-variations.js
+++ b/client/analytics/report/products/table-variations.js
@@ -6,14 +6,14 @@ import { __ } from '@wordpress/i18n';
 import { Component } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import { withSelect } from '@wordpress/data';
-import { get, map, orderBy } from 'lodash';
+import { map, orderBy } from 'lodash';
 
 /**
  * WooCommerce dependencies
  */
-import { appendTimestamp, getCurrentDates } from '@woocommerce/date';
 import { Link, TableCard } from '@woocommerce/components';
 import { formatCurrency, getCurrencyFormatDecimal } from '@woocommerce/currency';
+import { appendTimestamp, getCurrentDates } from '@woocommerce/date';
 import { getNewPath, getPersistedQuery, onQueryChange } from '@woocommerce/navigation';
 
 /**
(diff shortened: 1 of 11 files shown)


client/store/variations/resolvers.js
```diff
@@ -0,0 +1,27 @@
...
+export default {
+	async getVariations( state, query ) {
+		try {
+			const params = isEmpty( query ) ? '' : '?' + stringify( query );
+			// @TODO: Use /reports/variations when it becomes available
+			// const variations = await apiFetch( {
+			// 	path: '/wc/v3/reports/variations' + params,
+			// } );
+			const variations = await fetch(
+				'https://virtserver.swaggerhub.com/peterfabian/wc-v3-api/1.0.0/reports/variations' + params
```
belcherj: Not sure how I feel about using the swagger apis

psealock: How come?

belcherj: They give us a very static version of the data.  When we merge UI that rely on the swagger apis it is non-functional.  Additionally there is no pressure to get those API's working because the UI "working"

belcherj: I'm okay with approving, just wanted to see what others thought.  I have three PR's ready to roll that I could use this method for.

belcherj:
Fixed Merge Conflict for ya, @psealock

client/analytics/report/products/table-variations.js
```diff
@@ -0,0 +1,198 @@
...
+		const { query } = this.props;
+		const timeRelatedQuery = getTimeRelatedQuery( query );
+
+		return map( data, row => {
+			const {
+				items_sold,
+				gross_revenue,
+				orders_count,
+				stock_status = 'outofstock',
+				stock_quantity = '0',
```
timmyc: Should this be a `0` / number?

client/analytics/report/products/table-variations.js
```diff
@@ -0,0 +1,198 @@
...
+				isSortable: true,
+				isNumeric: true,
+			},
+			{
+				label: __( 'Status', 'wc-admin' ),
+				key: 'stock_status',
+			},
+			{
+				label: __( 'Stock', 'wc-admin' ),
+				key: 'stock',
```
timmyc: should this be `isNumeric` so the value gets localized?

client/store/variations/resolvers.js
timmyc: Per our discussion in the hangout yesterday, this is fine for now, we just need to call it out in release notes as not yet complete.

timmyc (approved):
Couple of minor comments, but overall this is working well for me, and code is looking good. Thanks for including tests!

client/analytics/report/products/table-variations.js
psealock: Most likely, yes. This value will eventually be returned by the API. Other table queries use a string, so I'll keep it this way for consistency

PR #773, opened 2018-11-04
Do not attempt to run table setup without dependency check first

### LINKED ISSUE #771: Fatal error: Class 'WC_Admin_Api_Init' not found when activating plugin
**Describe the bug**

If the user attempts to activate the wc-admin plugin when the Gutenberg plugin is not active, or if the user deactivates the Gutenberg plugin while the wc-admin plugin is active, the following fatal error occurs on both front side and admin side, requiring editing the plugin (or cli) to recover:

```
Fatal error: Class 'WC_Admin_Api_Init' not found in /home/allsno1/allensnook.com/wp-content/plugins/wc-admin/wc-admin.php on line 103
```

**To Reproduce**

1. Go to wp-admin > Plugins > Add New
2. Install and attempt to activate the wc-admin plugin
3. See error

**Expected behavior**

I expected the plugin to not fatal error

allendav:
Fixes #771 

We should run the dependency check in the init hook since it is possible that the plugin is in the process of activating, but will fail to complete activation due to a missing dependency (and then when we attempt to use the api init class, it is not present, and then boom.)

 - Install this branch on a site with the Gutenberg plugin deactivated
 - Activate this plugin
 - Make sure activation does not result in a Fatal Error

### DIFF
--- a/wc-admin.php
+++ b/wc-admin.php
@@ -98,6 +98,10 @@ register_deactivation_hook( WC_ADMIN_PLUGIN_FILE, 'deactivate_wc_admin_plugin' )
  * update automatically.
  */
 function wc_admin_init() {
+	if ( ! dependencies_satisfied() ) {
+		return;
+	}
+
 	// Only create/update tables on init if WP_DEBUG is true.
 	if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {
 		WC_Admin_Api_Init::create_db_tables();


wc-admin.php
```diff
@@ -98,6 +98,10 @@ function deactivate_wc_admin_plugin() {
  * update automatically.
  */
 function wc_admin_init() {
+	if ( ! dependencies_satisfied() ) {
```
timmyc: maybe we should rename this method `wc_admin_dependencies_satisfied` at some point too.

timmyc (approved):
Looks good to me, we can change that method name now or some other time.

allendav:
> Looks good to me, we can change that method name now or some other time.

Yeah, lets change that later

PR #774, opened 2018-11-05
Allow WC-Admin to activate on WP 5 without Gutenberg plugin active

### LINKED ISSUE #772: WC-ADMIN should not require Gutenberg plugin to be active for WordPress 5.0 and higher
**Describe the bug**

Attempting to activate the wc-admin plugin on a site running WordPress 5 ( 5.0-beta2-43859 )and WooCommerce (3.5.1) fails because even though Gutenberg is present in WordPress 5, our logic doesn't recognize it.

**To Reproduce**

1. Install a WordPress 5 beta (e.g. 5.0-beta2-43859 )
2. The Gutenberg plugin should be automatically disabled as a result.
3. Attempt to activate wc-admin plugin
4. See admin notice that Gutenberg is required

**Expected behavior**

wc-admin should activate on WordPress 5.0 without requiring Gutenberg

**Screenshots**

screen shot 2018-11-04 at 5 45 33 pm

allendav:
Fixes #772 

- Don't rely on GUTENBERG_DEVELOPMENT_MODE or GUTENBERG_VERSION - since those are not present in WordPress 5.
- Don't call `gutenberg_get_jed_locale_data` on WordPress 5 - use `wp_get_jed_locale_data` instead (see https://github.com/WordPress/gutenberg-examples/pull/51 )

- Install on WordPress 5
- Make sure you have `define('SCRIPT_DEBUG', true);` - there is a separate issue with the minified scripts on 5.0 that is causing a problem ( #796 )
- Do NOT have the Gutenberg plugin active
- Make sure you can activate this (the WC-Admin) plugin 1) without a fatal error and 2) that the wp-admin > WooCommerce > Dashboard loads properly (including no console errors)

- Install on WordPress 4.9.8
- Do NOT have the Gutenberg plugin active
- Make sure that you cannot activate this (the WC-Admin) plugin
- Install and activate the Gutenberg plugin
- Make sure you can activate this plugin 1) without a fatal error and 2) that the wp-admin > WooCommerce > Dashboard loads properly (including no console errors)

### DIFF
--- a/wc-admin.php
+++ b/wc-admin.php
@@ -43,7 +43,7 @@ function wc_admin_plugins_notice() {
  * @return bool
  */
 function dependencies_satisfied() {
-	$woocommerce_minimum_met      = class_exists( 'WooCommerce' ) && version_compare( WC_VERSION, '3.5', '>' );
+	$woocommerce_minimum_met = class_exists( 'WooCommerce' ) && version_compare( WC_VERSION, '3.5', '>' );
 	if ( ! $woocommerce_minimum_met ) {
 		return false;
 	}
--- a/wc-admin.php
+++ b/wc-admin.php
@@ -43,8 +43,16 @@ function wc_admin_plugins_notice() {
  * @return bool
  */
 function dependencies_satisfied() {
-	return ( defined( 'GUTENBERG_DEVELOPMENT_MODE' ) || defined( 'GUTENBERG_VERSION' ) )
-			&& class_exists( 'WooCommerce' ) && version_compare( WC_VERSION, '3.5', '>' );
+	$woocommerce_minimum_met      = class_exists( 'WooCommerce' ) && version_compare( WC_VERSION, '3.5', '>' );
+	if ( ! $woocommerce_minimum_met ) {
+		return false;
+	}
+
+	$wordpress_version            = get_bloginfo( 'version' );
+	$wordpress_includes_gutenberg = version_compare( $wordpress_version, '4.9.9', '>' );
+	$gutenberg_plugin_active      = defined( 'GUTENBERG_DEVELOPMENT_MODE' ) || defined( 'GUTENBERG_VERSION' );
+
+	return $wordpress_includes_gutenberg || $gutenberg_plugin_active;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


allendav:
Work in progress - almost there - but I am getting a few `Uncaught ReferenceError: regeneratorRuntime is not defined` errors in my console on WP 5 when the Gutenburg plugin is not active - it is breaking dashboard display.

cc @ryelle since my babel fu is limited

wc-admin.php
```diff
@@ -43,8 +43,16 @@ function wc_admin_plugins_notice() {
...
 function dependencies_satisfied() {
-	return ( defined( 'GUTENBERG_DEVELOPMENT_MODE' ) || defined( 'GUTENBERG_VERSION' ) )
-			&& class_exists( 'WooCommerce' ) && version_compare( WC_VERSION, '3.5', '>' );
+	$woocommerce_minimum_met = class_exists( 'WooCommerce' ) && version_compare( WC_VERSION, '3.5', '>' );
+	if ( ! $woocommerce_minimum_met ) {
+		return false;
+	}
+
+	$wordpress_version            = get_bloginfo( 'version' );
+	$wordpress_includes_gutenberg = version_compare( $wordpress_version, '4.9.9', '>' );
```
timmyc: Since we have the known issue of `SCRIPT_DEBUG` needing to be on, should we add that as a required condition to this line? That would allow us to ensure `wc-admin` will function in 5.0 until the other issue is resolved - and we could cut a release too!

```
$wordpress_includes_gutenberg = version_compare( $wordpress_version, '4.9.9', '>' ) && is_defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG;
```

timmyc (approved):
Tested out all use cases in 4.9.8 with and without Gutenberg active, as well as on 5.0-beta3-43876. Working as expected in both. I did have one comment/idea on adding in a check for SCRIPT_DEBUG so we can ensure all will work in 5.0 for now.

wc-admin.php
allendav: Changed in https://github.com/woocommerce/wc-admin/pull/774/commits/ca754100cd004a12361eb7e05a3760cf6122a4ef

timmyc (approved):
Changes look great, thanks for the updates and for fixing this up 🙌

PR #782, opened 2018-11-05
i18n: Inject translations when the i18n lib is included

### LINKED ISSUE #775: Jed localization error: Error: Domain `wc-admin` was not found.
**Describe the bug**

The following is logged to the console when navigating to wp-admin > WooCommerce > Dashboard

```
Jed localization error: 
Error: Domain `wc-admin` was not found.
```

ryelle:
Fixes #775 – wc-date uses i18n, but the locale isn't loaded until wc-components is enqueued. Now we set up the locale after including `wp-i18n`, to ensure our locale is always there.

**To test**

- Load up any wc-admin page
- There should not be any console error about jed/ domain `wc-admin`.


coderkevin (approved):
Tests good for me, no more warnings in the dev console! 🎉

PR #784, opened 2018-11-05
Try: Extend SummaryNumber

psealock:
# Don't Merge

# Extend Summary Numbers in Orders Report

Use apply filters to get a new chart object

screen shot 2018-11-05 at 4 52 07 pm

### DIFF
--- a/client/analytics/report/orders/chart.js
+++ b/client/analytics/report/orders/chart.js
@@ -6,6 +6,7 @@ import { __ } from '@wordpress/i18n';
 import { Component, Fragment } from '@wordpress/element';
 import { find } from 'lodash';
 import PropTypes from 'prop-types';
+import { applyFilters } from '@wordpress/hooks';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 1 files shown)


client/analytics/report/orders/chart.js
```diff
@@ -6,16 +6,19 @@ import { __ } from '@wordpress/i18n';
...
 import PropTypes from 'prop-types';
+import { applyFilters } from '@wordpress/hooks';
 
 /**
  * Internal dependencies
  */
 import ReportChart from 'analytics/components/report-chart';
 import ReportSummary from 'analytics/components/report-summary';
 
+const CHARTS_FILTER = 'woocommerce-charts-list';
```
claudiosanches: We should use `_` instead of `-` for hooks, to follow the same as WooCommerce core.

ryelle: Since these are JS hooks, not PHP hooks, I don't think the naming standards are the same. AFAIK there isn't anything written yet, but [the examples](https://wordpress.org/gutenberg/handbook/extensibility/extending-blocks/) [tend to use](https://wordpress.org/gutenberg/handbook/extensibility/extending-editor/) `-` or camelCase.

Probably this should actually be `woocommerce.getCharts`, looking at some of the existing gutenberg hooks.

psealock:
closing this because its so outdated and incomplete. Here is an issue to track extending summary numbers, https://github.com/woocommerce/woocommerce-admin/issues/2110

PR #785, opened 2018-11-06
Component - chart: more than 9 weeks of ticks change xFormat to months & x2Format to yrs

### LINKED ISSUE #632: Chart: x-axis refactor master thread
Master P2 Thread: p6riRB-3C9

## Sub-tasks
- [x] Don’t include the last date of the selected period if it produces uneven primary x-axis ticks - #635 
- [x] Display the full year - #638
- [x] Only include the year on first secondary x-axis tick of that year - #639
- [x] When the x-axis includes more than 62 days, switch the primary x-axis ticks to months, and only show the year on the secondary axis  #685 + #785

greenafrican:
Fixes #632 

Continuation of the work done in #685 and @LevinMedia 's feedback here: https://github.com/woocommerce/wc-admin/pull/685#issuecomment-434059061 We've changed the `xFormat` to months and the `x2Format` to years when there are more than 9 weeks (63 days) of x1 ticks.

### Screenshots
ticks > 9 weeks (63 days)
screenshot 2018-11-06 16 01 23

ticks < 9 weeks (63 days)
screenshot 2018-11-06 16 14 50

 - Go to Analytics >> Revenue
 - Select 'Last Quarter' compared to 'Previous Quarter'
 - Change the chart interval to 'By Week'
 - Ensure months are shown in the x1 axis and years in the x2 axis and that there aren't duplicates in the x1 axis

### DIFF
--- a/client/components/chart/utils.js
+++ b/client/components/chart/utils.js
@@ -374,6 +374,14 @@ export const compareStrings = ( s1, s2, splitChar = ' ' ) => {
 
 export const drawAxis = ( node, params ) => {
 	const xScale = params.type === 'line' ? params.xLineScale : params.xScale;
+	const removeDuplicateDates = ( d, i, ticks, formatter ) => {
+		const monthDate = d instanceof Date ? d : new Date( d );
+		let prevMonth = i !== 0 ? ticks[ i - 1 ] : ticks[ i ];
+		prevMonth = prevMonth instanceof Date ? prevMonth : new Date( prevMonth );
+		return i === 0
+			? formatter( monthDate )
+			: compareStrings( formatter( prevMonth ), formatter( monthDate ) ).join( ' ' );
+	};
 
 	const yGrids = [];
 	for ( let i = 0; i < 4; i++ ) {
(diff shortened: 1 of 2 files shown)


Aljullu:
Good improvement! :+1: 

I still found one case where the chart looks a like bit weird to me. When selecting _Year to date_ and the interval _By date_ this is how it looks in my site:
year-to-date
I think months' ticks should be rendered on the first week of each month. This way we would also avoid having an empty tick between January and February. Like this:
image

What do you think @LevinMedia?

@greenafrican 

Modifying [this line](https://github.com/woocommerce/wc-admin/blob/master/client/components/chart/utils.js#L306) from:
```ES6
	if ( uniqueDates.length >= dayTicksThreshold && interval === 'day' ) {
```
to:
```ES6
	if ( uniqueDates.length >= dayTicksThreshold && interval === 'day' ||
			uniqueDates.length >= weekTicksThreshold && interval === 'week' ) {
```
should be enough. :slightly_smiling_face:

greenafrican:
Thanks, @Aljullu that is a good catch and a thank you for proposing a quick fix, which works well!

PR #794, opened 2018-11-07
Creates a helper function to get selected chart

belcherj:
Adds a getSelectedChart helper function so that it can be used by all reports.

 - Open orders chart and click through summary numbers

### DIFF
--- a/client/lib/get-selected-chart/index.js
+++ b/client/lib/get-selected-chart/index.js
@@ -12,7 +12,7 @@ import { find } from 'lodash';
  * @param {array} charts - list of charts for a particular report
  * @returns {object} - chart configuration object
  */
-export default function getSelectedChart( chartName, charts ) {
+export default function getSelectedChart( chartName, charts = [] ) {
 	const chart = find( charts, { key: chartName } );
 	if ( chart ) {
 		return chart;
--- a/client/lib/get-selected-chart/index.js
+++ b/client/lib/get-selected-chart/index.js
@@ -6,7 +6,8 @@
 import { find } from 'lodash';
 
 /**
- * Takes a chart name returns the configuration for that chart from and array of charts
+ * Takes a chart name returns the configuration for that chart from and array
+ * of charts. If the chart is not found it will return the first chart.
  *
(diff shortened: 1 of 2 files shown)


client/lib/get-selected-chart/index.js
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Takes a chart name returns the configuration for that chart from and array of charts
+ *
+ * @param {string} chartName - the name of the chart to get configuration for
+ * @param {array} charts - list of charts for a particular report
+ * @returns {object} - chart configuration object
+ */
+export default function getSelectedChart( chartName, charts ) {
+	const chart = find( charts, { key: chartName } );
+	console.log( chartName, charts, chart );
```
coderkevin: You'll need to remove this before merging.

client/lib/get-selected-chart/index.js
```diff
@@ -0,0 +1,22 @@
...
+ * External dependencies
+ */
+import { find } from 'lodash';
+
+/**
+ * Takes a chart name returns the configuration for that chart from and array of charts
+ *
+ * @param {string} chartName - the name of the chart to get configuration for
+ * @param {array} charts - list of charts for a particular report
+ * @returns {object} - chart configuration object
```
coderkevin: May want to mention that if the chart name isn't found that it returns the first in the array.

client/lib/get-selected-chart/index.js
```diff
@@ -0,0 +1,22 @@
...
+ */
+export default function getSelectedChart( chartName, charts ) {
+	const chart = find( charts, { key: chartName } );
+	console.log( chartName, charts, chart );
+	if ( chart ) {
+		return chart;
+	}
+	return charts[ 0 ];
```
coderkevin: Can we be sure that charts will never be empty?

client/lib/get-selected-chart/index.js
belcherj: Added to the function description

client/lib/get-selected-chart/index.js
belcherj: Removed

PR #803, opened 2018-11-07
Build: Move filter code to the navigation package

ryelle:
- Moves filter-related code to navigation, so it can be shared between components and the app
- Fixes spelling of “flatten”
- Update README docs with new functions

Navigation seemed like a logical place for this since they're somewhat query-related, but I could also see adding it elsewhere. It needs to be external to the component directory because we use these functions in the data store. 

**To test**

- Check that no functionality has changed, especially on reports pages


timmyc (approved):
+1 on navigation being a good home for this, since it is all GET query based. One minor comment in the code, but overall in all my testing of filters, date picker etc in the reports things are working as expected.

Thanks for getting this into the exports as they will definitely be of use to all woo devs 🌟

psealock (approved):
Working as expected

ryelle:
I've added tests for `getDefaultOptionValue`, including a case where options is undefined -- which we now catch with `get`. Since the only change here is tests, I'm going to merge when the test  ✅

PR #804, opened 2018-11-07
Build: Move components to packages dir

ryelle:
This PR moves the components to a new package, `@woocommerce/components`, which will be published to NPM for 3rd party use. The components are also currently available on the `wc` global, ex: `wc.components.Card`.

The changes here are:
- The CSS for each component is now included in [a separate scss file](https://github.com/woocommerce/wc-admin/blob/update/packages-components/packages/components/src/style.scss), so that the build script picks up the CSS – the package build process doesn't handle extracting CSS from JS files, so we build these separately.
- File paths to other components are now using relative URLs, rather than relying on the components alias
- Chart components are left alone in `client/components`

**To test**

- Check that there is no functionality change in the app
- Test devdocs: `/wp-admin/admin.php?page=wc-admin#/devdocs`
- Note: there is an error in the filters example because the config objects passed in `packages/components/src/filters/example.md` are not up-to-date, that can be fixed in a following PR.


belcherj:
Tested and works.

belcherj:
Dev docs:

screen shot 2018-11-12 at 12 40 24 pm

belcherj:
Dev docs `npm start`:
screen shot 2018-11-12 at 12 41 18 pm

ryelle:
@belcherj yep, that's the filters issue I mentioned – the page still renders correctly, right? it's just the `Filters` example that's broken.

belcherj:
Ahhh, okay, will continue testing in a bit.

packages/components/src/style.scss
```diff
@@ -0,0 +1,30 @@
+/**
+ * Internal Dependencies
+ */
+@import 'animation-slider/style.scss';
```
belcherj: Is there a reason to do this?  I like having the css directly included via the component.

belcherj (dismissed):
Tested and works as long as we take care of the docs issue is a subsequent PR. Let me know if you want me to take that on.

packages/components/src/style.scss
ryelle: The package build process doesn't extract CSS from JS (it doesn't use webpack, just babel), so we [build the CSS separately and compile it with node-sass](https://github.com/woocommerce/wc-admin/blob/update/packages-components/bin/packages/build.js#L118).

ryelle: Thanks for mentioning it, I haven't seen that before. Right now we don't have a need to keep the css imports in the component files (since we're using this extra scss "entry point" file), but it might be useful in the future 👍

coderkevin (approved):
LGTM, just had a couple questions. But don't let those hold you up from merging. 👍

client/devdocs/example.js
```diff
@@ -16,7 +16,8 @@ import PropTypes from 'prop-types';
 /**
  * Internal dependencies
  */
-import * as components from '@woocommerce/components';
+import * as components from 'components';
+import * as pkgComponents from '@woocommerce/components';
```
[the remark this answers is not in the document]
ryelle: Eventually the chart will be its own package too, and this might contain any other app-specific components. They are both "permanent".

PR #808, opened 2018-11-09
Remove the SCRIPT_DEBUG:true requirement

### LINKED ISSUE #796: WP 5: regeneratorRuntime is not defined error when SCRIPT_DEBUG false
**Describe the bug**

The WooCommerce dashboard fails to load unless you have `define('SCRIPT_DEBUG', true);` in wp-config.php. Possible WordPress 5 problem.

**To Reproduce**

1. Install WordPress 5 (e.g. 5.0-beta3-43876)
2. Set SCRIPT_DEBUG to false (in wp-config.php)
3. Navigate to wp-admin > WooCommerce > Dashboard
4. Note that the dashboard fails to load - the following is logged in the console: 

```
Uncaught ReferenceError: regeneratorRuntime is not defined
    at api-fetch.min.js?ver=5.0-beta3-43876:1
```

allendav:
Fixes #796 

As of WP 5.0-beta3-43878 SCRIPT_DEBUG is no longer required to avoid the regeneratorRuntime Javascript asplosion.

- Make sure SCRIPT_DEBUG is not true/defined
- Deactivate and activate this plugin, and ensure you are able to do so.
- Deactivate WooCommerce itself and ensure the alert doesn't mention SCRIPT_DEBUG anymore
- Make sure the README no longer mentions SCRIPT_DEBUG as a WP 5 requirement

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,10 +14,6 @@ This is a feature plugin for a modern, javascript-driven WooCommerce Admin exper
 
 For better debugging, it's also recommended you add `define( 'SCRIPT_DEBUG', true );` to your wp-config. This will load the unminified version of all libraries, and specifically the development build of React.
 
-### WordPress 5
-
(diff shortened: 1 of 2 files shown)


belcherj (approved):
Awesome Sauce!!

Issue #695, opened 2018-10-26
Component: Comparison Chart


LevinMedia:
**Blocked: ** Until the API can be segmented, all results are lumped together. When that happens, `ReportChart` [can be modified](https://github.com/woocommerce/wc-admin/blob/b945875fae27f44e7a49d9b12347ab007cf2ca4c/client/analytics/components/report-chart/index.js#L66-L76) to deliver the segments to the chart:

The comparison chart differs from a standard chart (IE, what you'll see in the revenue report) in that it allows a user to compare (surprise!) multiple series of data. it's detailed at: p6riRB-3u4-p2

**The core reports that will feature the comparison chart in addition to the standard chart include:**
Product / Product Detail Report
Category Report
Coupon Report
Tax Report

**Comparison chart in the context of the product detail report:**

<img width="1172" alt="screen shot 2018-10-25 at 7 07 58 pm" src="https://user-images.githubusercontent.com/4500952/47540302-5b60f480-d889-11e8-8f15-9ad4f0e1c142.png">

- Data points selected for comparison appear in the legend to the left of the chart
- A maximum of (X) data points can be plotted in the graph at any given time, however, an unlimited number of data points can be added to the comparison (The final number of data points to plot at any give time are still TBD, and depend on two factors: load performance, and visual design)
- As data points are added to the comparison, the legend will eventually overflow its container and become scrollable vertically. 
- To provide context, the number of data points being compared are displayed at the bottom of the legend.  

**A variation on the comparison chart: Filtering a report by top products, top coupons etc** 

A select menu is introduced to the chart header which allows a user to pick the number of data points to compare. Further detail at p6riRB-3jq-p2

<img width="711" alt="screen shot 2018-10-25 at 7 32 55 pm" src="https://user-images.githubusercontent.com/4500952/47541159-1e96fc80-d88d-11e8-91e9-51518a4d8852.png">

In all cases where a comparison chart is being used, the summary numbers and tabular data update to reflect only what's being compared.

psealock:
Information on the progress of segmenting the API response in p7bje6-1bY-p2

justinshreve:
> In all cases where a comparison chart is being used, the summary numbers and tabular data update to reflect only what's being compared.

I wanted to double check about this @LevinMedia.

So with all variations selected it would look like this:

[...]

Issue #582, opened 2018-10-18
Chart with empty data set


joshuatf:
If dates are selected on the Product Reports chart with no order history, an emtpy chart is shown.  This may be confusing or look broken to users.

<img width="1113" alt="screen shot 2018-10-18 at 4 49 57 pm" src="https://user-images.githubusercontent.com/10561050/47183808-00625700-d2f7-11e8-860f-85a4c720c30d.png">

#### Options

Should we:

* Show a message or something here?  E.g., "There is no data for your selected range."
* Hide the chart
* Something else?

#### Additional context
Depends on empty product data returned from API - #581

LevinMedia:
Hi Josh, this is part of a larger "empty states" issue I need to address.  

For the time being I'll repurpose an SVG image file we can include to indicate that there is no data for the selected time period.

timmyc:
@LevinMedia do you have time at all next week to get a design for the empty state on the charts? I'd like to get this into sprint 11

LevinMedia:
@joshuatf @timmyc  Did some fiddling with empty states this afternoon... we're not that far off. Here's what I was thinking: 

<img width="1380" alt="screen shot 2019-01-30 at 8 05 39 pm" src="https://user-images.githubusercontent.com/4500952/52029986-91d28f00-24ca-11e9-9a81-538cd0029ecc.png">

- Always show every series in the legend (both standard charts and comparison charts) with their values being 0, or $0 depending on if the value is a number or a currency. 
- Remove the values from the Y-axis ticks except for the baseline. 
- The baseline should read 0, or $0 depending on if the value is a number or currency. 
- The x-axis should always show the ticks according to the date range selected. 
- If the chart is in line chart mode, it should flatline across the bottom of the chart.
- If the chart is in bar chart mode, it should be blank
- in the middle of the chart, display the message "No data for the selected date range" 
- The text should wrap to two lines and maintain it's position (centered vertically) on smaller viewports
`font-size:18px;`
`line-height:27px` 
`font-weight: bold;`
`color: light-grey-700`

timmyc:
Thanks @LevinMedia !

Issue #564, opened 2018-10-17
Product Detail Report


timmyc:
Since this is some distinct functionality that is an extension of the standard chart in the products report, it seemed wise to break it out into its own issue [see David's comment for context](https://github.com/woocommerce/wc-admin/issues/534#issuecomment-430759198).

<img width="1361" alt="screen shot 2018-10-24 at 12 16 40 pm" src="https://user-images.githubusercontent.com/4500952/47455580-0dfc5e80-d787-11e8-9016-bed4373fd243.png">

Full detail at the second in the second half of the Product Report post here: p6riRB-3dP-p2

When the product report is filtered to look at a single product, (either by using the filter at the top of the report, or clicking a product title in the tabular data) the variation filter should appear when applicable and the chart should respond accordingly.

Related - for the product detail screen, we'll also need to build out / include the comparison chart detailed at: p6riRB-3u4-p2

<img width="748" alt="screen shot 2018-10-17 at 12 31 59 pm" src="https://user-images.githubusercontent.com/4500952/47111552-b307bc00-d208-11e8-8bcb-e84cdd1e9f5c.png">

@LevinMedia what is the anticipated changes to the tabular data when in the detail view? Raw sales data for the product/variation(s) selected over the time period? 

### Child Tasks
- [x] Add Variations filter https://github.com/woocommerce/wc-admin/issues/681
- [x] Add Variations autocompleter https://github.com/woocommerce/wc-admin/pull/742
- [x] Variations Table https://github.com/woocommerce/wc-admin/issues/762
- [x] Segmentation of API response. Trackable in p7bje6-1bY-p2
- [x] VariationImage for autocompleter #832
- [x] Comparison Chart: design #816
- [x] Comparison Chart: API #695

PR #816, opened 2018-11-12
Update comparison chart design and show it on the Product report

Aljullu:
Part of #695.

This PR does several things:
- Renders the charts with the comparison layout in the Product report.
- Updates the comparison layout to the new designs.
- Splits the CSS of the chart component in two different files (created a new one for the legend).

### Screenshots
image

Given that the API endpoints are not ready yet, we must do a couple of tweaks in the code so we generate a long legend for the chart.

Go to `client/components/chart/legend.js` and replace [this line](https://github.com/woocommerce/wc-admin/compare/add/product-detail-report-chart?expand=1#diff-979f811d595d5bbcc243fc535a1ba3f0R83):
```ES6
const showTotalLabel = legendDirection === 'column' && data.length > 5 && itemsLabel;
```
with:
```ES6
const showTotalLabel = legendDirection === 'column' && data.length > 0 && itemsLabel;
```
and [this other line](https://github.com/woocommerce/wc-admin/compare/add/product-detail-report-chart?expand=1#diff-979f811d595d5bbcc243fc535a1ba3f0R102):
```ES6
{ data.map( row => ( 
```
with something like this:
```ES6
{ data.concat( data ).concat( data ).concat( data ).concat( data ).concat( data ).concat( data ).map( row => (
```
This way we will have many items in the legend and will be ready to test the PR.
- Go to the _Products_ report (`/wp-admin/admin.php?page=wc-admin#/analytics/products`).
- Verify the chart is displayed with the comparison layout (screenshot above):
  - The legend is on the left on wide screens and moves down to the bottom on mobile.
  - A shadow is displayed over the legend to make it clear it's scrollable and it disappears when you reach the end of the list.
  - A label is displayed at the bottom with the words _2 products_.
- Activate the Single Product filter (see screenshot).
image
  - Verify the label at the bottom of the legend changed to _2 variations_.

### To do
- The chart still displays all products data together instead of split by product/variation. I will leave #695 open to track that once the API allows querying the data that way.
- The chart height and the legend height don't match (as can be seen in the screenshot above). I will work on that in a follow-up PR, given that this one is already quite big.


belcherj:
Getting this error on products screen:  `Error: <g> attribute transform: Trailing garbage, "translate(0,NaN)".`

belcherj:
Should this be merged w/o apis being ready?

client/analytics/components/report-chart/index.js
```diff
@@ -29,7 +29,15 @@ import ReportError from 'analytics/components/report-error';
 
 class ReportChart extends Component {
 	render() {
-		const { path, primaryData, secondaryData, selectedChart, query } = this.props;
+		const {
+			comparisonChart,
+			itemsLabel,
+			query,
```
belcherj: not alpha sorted

client/analytics/report/products/index.js
```diff
@@ -23,6 +24,11 @@ export default class ProductsReport extends Component {
 	render() {
 		const { path, query } = this.props;
 
+		const itemsLabel =
+			'single_product' === query.filter && !! query.products
+				? __( '%s variations', 'wc-admin' )
```
belcherj: What is the convention on multi-line ternaries?

client/components/chart/index.js
```diff
@@ -211,12 +212,12 @@ class Chart extends Component {
 		chartHeight = width <= 783 ? 180 : chartHeight;
 		const legend = (
 			<Legend
-				className={ 'woocommerce-chart__legend' }
```
belcherj: Was this deliberate?  Does this break other chart legends?

client/components/chart/legend.js
```diff
@@ -14,7 +15,7 @@ import { formatCurrency } from '@woocommerce/currency';
 /**
  * Internal dependencies
  */
-import './style.scss';
+import './legend.scss';
```
belcherj: Why the change to sass file name?

client/components/chart/legend.js
```diff
@@ -36,65 +37,108 @@ function getFormatedTotal( total, valueType ) {
...
+		this.updateListScroll();
+		window.addEventListener( 'resize', this.updateListScroll );
+	}
+
+	componentWillUnmount() {
+		window.removeEventListener( 'resize', this.updateListScroll );
+	}
+
+	updateListScroll = () => {
+		const list = this.listRef.current;
```
belcherj: does react allow for request animation frame?

client/components/chart/index.js
Aljullu: Yes, that class is not used anywhere.

client/components/chart/legend.js
Aljullu: `style.scss` was getting quite big so I decided to divide it in two files.

`style.scss` is still called from `charts.js` so it's not a problem removing it here.

Aljullu:
> Getting this error on products screen: `Error: <g> attribute transform: Trailing garbage, "translate(0,NaN)".`

This seems to be related to the chart `<svg>` and it also occurs in master. I would say it's part of #582 and not related with this PR.

belcherj (approved):
LGTM

client/analytics/report/products/index.js
Aljullu: Per https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/#multi-line-statements I understand it's ok using them, right?

client/components/chart/legend.js
Aljullu: I found something in the docs that might be useful in this particular case:
https://reactjs.org/docs/faq-functions.html#requestanimationframe-throttling

I spent some time playing with it but I don't have anything ready yet. I created an issue so we can take a look at it in the future: #822.

Aljullu:
> Should this be merged w/o apis being ready?

We have been merging everything in #564 so I think it makes sense merging this one too.

PR #817, opened 2018-11-12
Update table gradient code to be more similar to chart legend gradient code

Aljullu:
While working on #816 I added a gradient at the bottom of the chart legend to make it clear it's scrollable and I based my code on the same gradient we have in the tables. But made a couple of small code improvements that I'm backporting now to the table gradient code.

### Screenshots
_(this PR shouldn't have created any visual change other than adding a transition)_
image

- Go to a page with a table (eg: `/wp-admin/admin.php?page=wc-admin#/analytics/revenue`).
- Resize the viewport so the gradient appears.
- Make sure there are no regressions and it's still displayed as usual and hidden when you reach the end of the table.

### DIFF
--- a/client/components/table/style.scss
+++ b/client/components/table/style.scss
@@ -144,11 +144,13 @@
 		width: 41px;
 		height: 100%;
 		background: linear-gradient(90deg, rgba(0, 0, 0, 0), rgba(0, 0, 0, 0.2));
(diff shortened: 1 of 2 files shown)


belcherj (approved):
Tested, looks good.  Code, looks good!

PR #818, opened 2018-11-12
Try: chart component as a package part 1

greenafrican:
There have been a few projects using D3 for charts and each project has created new components from scratch. It was our intention to make the chart component from this project a separate package that can be used in other projects and code. 

I've decided to break this down into a few PRs so that it doesn't constantly stray too far from other chart and component related work and PRs.

In this PR, I've attempted to:

- split out the `d3chart` (base, `chart.js`, `utils.js`) to be moved into a separate component and then `chart` (`index.js`) component to wrap the `d3chart` for `wc-admin`
- remove internal dependencies from `d3chart`
- simplify the tooltip value and label formatting
- merge `mode` and `layout` props and uses
- split out styling
- allow formatting functions, like`formatCurrency()` to be passed as the format prop, this works for `xFormat`, `x2Format`, `yFormat`, `tooltipLabelFormat`, `tooltipValueFormat`.

Testing:
- ensure bar and line chart works as expected in all pages, intervals, date ranges
- you can also write your own formatting function for any of the format props, for example:
```
const toTitleCase = ( str ) => {
	str = str.toLowerCase().split(' ');
	for ( let i = 0; i < str.length; i++ ) {
		str[ i ] = str[ i ].charAt( 0 ).toUpperCase() + str[ i ].slice( 1 );
	}
	return str.join( ' ' );
};
...
<D3Chart
    ...
    tooltipLabelFormat={ toTitleCase }
    ...
/>
```


client/analytics/components/report-chart/index.js
```diff
@@ -60,6 +61,24 @@ class ReportChart extends Component {
 		const { primary, secondary } = getCurrentDates( query );
 		const primaryKey = `${ primary.label } (${ primary.range })`;
 		const secondaryKey = `${ secondary.label } (${ secondary.range })`;
+		let tooltipValueFormat;
```
Aljullu: Not a big deal, but maybe we can extract the `switch` logic into a method to make the render function easier to read?
```suggestion
		const tooltipValueFormat = this.getTooltipValueFormat( selectedChart.type );
```

client/components/d3chart/d3-base/index.js
```diff
@@ -5,7 +5,7 @@
  */
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
-import { Component, createRef } from '@wordpress/element';
+import { Component, createRef } from 'react';
```
Aljullu: Other components are using `@wordpress/element` instead of `react` for these dependencies:
https://github.com/woocommerce/wc-admin/blob/master/packages/components/src/search/index.js#L6
I think it would make sense keeping it consistent and use `@wordpress/element`.

client/components/d3chart/example-hour.js
```diff
@@ -13,7 +13,7 @@ import { Card } from '@woocommerce/components';
 /**
  * Internal dependencies
  */
-import Chart from './index';
+import Chart from '../chart/index';
```
Aljullu: Are these examples still used somewhere?

client/components/d3chart/index.js
```diff
@@ -4,7 +4,7 @@
  * External dependencies
  */
 import { isEqual } from 'lodash';
-import { Component, createRef } from '@wordpress/element';
+import { Component, createRef } from 'react';
```
Aljullu: Same as before. :slightly_smiling_face:

client/components/d3chart/index.js
```diff
@@ -115,7 +115,7 @@ class D3Chart extends Component {
 			valueType,
 		} = this.props;
 		const { width } = this.state;
-		const calculatedWidth = width || node.offsetWidth;
+		const calculatedWidth = width && width > 0 ? width : node.offsetWidth;
```
Aljullu: Is this change required? A `width` of `0` would evaluate to `false`, isn't it?

client/components/d3chart/utils.js
```diff
@@ -295,16 +293,16 @@ const calculateXTicksIncrementFactor = ( uniqueDates, maxTicks ) => {
...
  */
-export const getXTicks = ( uniqueDates, width, layout, interval ) => {
-	const maxTicks = calculateMaxXTicks( width, layout );
+export const getXTicks = ( uniqueDates, width, mode, interval ) => {
+	const maxTicks = calculateMaxXTicks( width, mode );
 
 	if (
-		( uniqueDates.length >= dayTicksThreshold && interval === 'day' ) ||
-		( uniqueDates.length >= weekTicksThreshold && interval === 'week' )
+		( uniqueDates.length >= 63 && interval === 'day' ) ||
```
Aljullu: Even if these variables are only used here, could we avoid magic numbers?
```ES6
 	const dayTicksThreshold = 63;
 	const weekTicksThreshold = 9;
 	if ( 
 		( uniqueDates.length >= dayTicksThreshold && interval === 'day' ) || 
 		( uniqueDates.length >= weekTicksThreshold && interval === 'week' ) 
 	) { ... }
```

client/components/d3chart/utils.js
```diff
@@ -458,22 +456,12 @@ export const drawAxis = ( node, params ) => {
...
-			break;
-		case 'currency':
-			rowTotal = formatCurrency( total );
-			break;
-		case 'number':
-			break;
-	}
-	return rowTotal;
+const getTooltipRowLabel = ( d, row, params ) => {
+	const isDate =
```
Aljullu: I would expect a `isDate` variable to be a bool instead of a date. Could we rename this one to something else? Maybe just `date` or `labelDate`.

client/components/d3chart/utils.js
```diff
@@ -458,22 +456,12 @@ export const drawAxis = ( node, params ) => {
...
+		d[ row.key ].labelDate instanceof Date
+			? d[ row.key ].labelDate
+			: new Date( d[ row.key ].labelDate );
+	return d[ row.key ].labelDate ? params.tooltipLabelFormat( isDate ) : row.key;
```
Aljullu: I'm not sure to understand how this function works. Couldn't we check if `labelDate` exists before doing everything else?
```ES6
const getTooltipRowLabel = ( d, row, params ) => {
	if ( d[ row.key ].labelDate ) {
		const isDate = d[ row.key ].labelDate instanceof Date
			? d[ row.key ].labelDate
			: new Date( d[ row.key ].labelDate );
		return params.tooltipLabelFormat( isDate );
	}
	return row.key;
};
```

client/components/d3chart/d3-base/index.js
greenafrican: I think we should move away from `@wordpress` dependencies in `d3chart` so that it can be used in other non-WordPress related projects?

client/components/d3chart/example-hour.js
greenafrican: No, good point. I'll remove them :)

client/components/d3chart/index.js
greenafrican: It turns out that in certain circumstances the width comes through as negative, for example `width=-320`, likely from here https://github.com/woocommerce/wc-admin/blob/master/client/components/chart/index.js#L313 and although there is now a catch here https://github.com/woocommerce/wc-admin/blob/master/client/components/chart/index.js#L298 we can't rely on that in other implementations.

client/components/d3chart/d3-base/index.js
ryelle: We import [`@wordpress/element` in calypso,](https://github.com/Automattic/wp-calypso/blob/master/package.json#L50) and a long term goal AFAIK is to converge onto gutenberg– so I would stick with the `wordpress` packages. These are all NPM packages, so they _can_ be used in non-WP projects.

client/components/d3chart/index.js
Aljullu: Makes sense. :+1:

PR #820, opened 2018-11-12
Remove unneeded leading whitespace

belcherj:
This has made it difficult to quick create test instructions since I have to add a leading space before every step.

 - No testing needed


Aljullu (approved):
Love this change :heart_eyes: I was always removing the leading whitespaces manually.

PR #823, opened 2018-11-12
Apply onQueryChange to more than just 'filter'

### LINKED ISSUE #814: Table Compare: Move away from hard coded 'filter'
@woocommerce/navigation function `onQueryChange` uses the `filter` parameter to update. The Variations filter and Table operates on a different query parameter, however.

https://github.com/woocommerce/wc-admin/blob/4bdfddcb20481dc36376bb3d7baa03ec473ac46e/packages/navigation/src/index.js#L108-L109

psealock:
Fixes https://github.com/woocommerce/wc-admin/issues/814

Addition of the Variations table in https://github.com/woocommerce/wc-admin/pull/764 along with the "Compare Variations" filter mean that usage of `onQueryChange` in the table be able to amend multiple url parameters, not just `filter`. For example, `filter-variations` for comparing variations.

Also, some config parameters weren't updated in a recent change, ie `product` => `products`, and caused breakage when using a table's Compare function. So thats fixed here as well.

## Test

No changes, filters should work as they have been.

### DIFF
--- a/client/analytics/report/products/config.js
+++ b/client/analytics/report/products/config.js
@@ -60,7 +60,7 @@ const filterConfig = {
 		},
 		{
 			label: __( 'Product Comparison', 'wc-admin' ),
-			value: 'compare-product',
+			value: 'compare-products',
 			settings: {
 				type: 'products',
 				param: 'products',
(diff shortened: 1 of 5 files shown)


Aljullu (approved):
LGTM, fixes several issues with the filters and will enable comparisons in the Variations table. :tada:

belcherj:
@psealock I fixed the merge conflict for you.

PR #838, opened 2018-11-13
Update chart breakpoints

Aljullu:
Follow up of #816.

* Use `@wordpress/viewport` for chart breakpoints.

* Fix the issue with chart height not matching legend height.

### Screenshots
Before:
image

After:
image

Go to `client/components/chart/legend.js` and replace [this line](https://github.com/woocommerce/wc-admin/blob/master/client/components/chart/legend.js#L83):
```ES6
const showTotalLabel = legendDirection === 'column' && data.length > 5 && itemsLabel;
```
with:
```ES6
const showTotalLabel = legendDirection === 'column' && data.length > 0 && itemsLabel;
```
- Go to the _Products_ report (`/wp-admin/admin.php?page=wc-admin#/analytics/products`).
- Verify the _X products_ label border and the X axis of the legend are rendered at the same distance to the bottom.
- Resize the chart, navigate to other charts, hover the points, switch to the bar chart, etc. and verify nothing broke and there are no regressions.

### DIFF
--- a/client/components/chart/charts.js
+++ b/client/components/chart/charts.js
@@ -118,7 +118,6 @@ class D3Chart extends Component {
 		const { width } = this.state;
 		const calculatedWidth = width || node.offsetWidth;
 		const calculatedHeight = height || node.offsetHeight;
-		const scale = width / node.offsetWidth;
 		const adjHeight = calculatedHeight - margin.top - margin.bottom;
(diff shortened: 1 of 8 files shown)


belcherj:
Merged in master to make CI pass. @Aljullu

PR #840, opened 2018-11-13
Add a note about using SwaggerHub in PR template

psealock:
Adds a section in the pull request template to include use of SwaggerHub api's in the release notes.


joshuatf (approved):
Good idea to add this in if we're going to use SwaggerHub for the time being.  LGTM!

PR #841, opened 2018-11-14
Chart: Persist interval and chart type

psealock:
Partial fix of https://github.com/woocommerce/wc-admin/issues/674

Chart settings Interval and Type should persist between reports and when changing a report, such as filtering or changing dates.

### Changes
* `getTimeRelatedQuery` is now `getPersistedQuery` to reflect the inclusion of chart `type`.
* Remove `type` from `<Chart />` state in favor of props passed down, similar to how `interval` is used.

### Test

1. Go to any report.
2. Change chart interval and chart type.
3. Change the time period.
4. Navigate around the app (via sidebar and breadcrumbs) and use filters.
5. Confirm the chart settings don't change.

### DIFF
--- a/client/analytics/components/report-chart/index.js
+++ b/client/analytics/components/report-chart/index.js
@@ -17,6 +17,7 @@ import {
 	getCurrentDates,
 	getDateFormatsForInterval,
 	getIntervalForQuery,
+	getChartTypeForQuery,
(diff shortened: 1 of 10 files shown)


Aljullu (approved):
Works well and code looks good.

LGTM :shipit:

Issue #845, opened 2018-11-14
REST API: Implement `reports/coupons` and `reports/coupons/stats` endpoints


timmyc:
Currently the REST API reporting endpoints for coupons are using `WC_Reports_Orders_Stats_Query` with a `todo` to swap out the correct class name. In order to fully implement the reports, we both of the coupons endpoints fully developed.

Planning Post on Proton: p7bje6-15o-p2

PR #867, opened 2018-11-15
Add Coupons report table

### LINKED ISSUE #835: Coupons Report: Table
Child task of #171.

This task involves creating the table from API data for the Coupons Report.

Aljullu:
Fixes #835.

This PR creates a table for the Coupons Report.

API calls are made to swaggerhub for now and are marked with `@TODO` comments (see #845).

### Screenshots
image

- Go to the _Coupons_ report.
- Verify there is a table and interact with it.

### Pending tasks
The search autocompleter is not working because it also depends on the API.

### DIFF
--- a/client/analytics/report/coupons/table.js
+++ b/client/analytics/report/coupons/table.js
@@ -1,0 +1,227 @@
+/** @format */
+/**
+ * External dependencies
+ */
+import { __, _n } from '@wordpress/i18n';
+import { Component } from '@wordpress/element';
+import { format as formatDate } from '@wordpress/date';
+import { compose } from '@wordpress/compose';
+import { withSelect } from '@wordpress/data';
+import { get, map, orderBy } from 'lodash';
+
+/**
+ * WooCommerce dependencies
+ */
+import {
+	appendTimestamp,
+	getCurrentDates,
+	getIntervalForQuery,
+	getDateFormatsForInterval,
+} from '@woocommerce/date';
+import { Link, TableCard } from '@woocommerce/components';
+import { formatCurrency, getCurrencyFormatDecimal } from '@woocommerce/currency';
+import { onQueryChange } from '@woocommerce/navigation';
+
+/**
+ * Internal dependencies
+ */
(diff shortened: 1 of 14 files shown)


client/analytics/report/coupons/table.js
```diff
@@ -0,0 +1,227 @@
...
+/**
+ * Internal dependencies
+ */
+import ReportError from 'analytics/components/report-error';
+import { QUERY_DEFAULTS } from 'store/constants';
+import { getReportChartData, getFilterQuery } from 'store/reports/utils';
+
+class CouponsReportTable extends Component {
+	constructor( props ) {
+		super( props );
```
timmyc: Since we aren't binding or setting any default state, do we need the constructor?

client/analytics/report/coupons/table.js
```diff
@@ -0,0 +1,227 @@
...
+			// @TODO must link to the coupon detail report
+			const couponLink = (
+				<Link href="" type="wc-admin">
+					{ coupon_id }
+				</Link>
+			);
+
+			const ordersLink = (
+				<Link
+					href={ '/analytics/orders?filter=advanced&code_includes=' + coupon_id }
```
timmyc: Nice!

timmyc (approved):
A new report is born! Awesome!

So one minor comment in the code, but this is testing well for me. I did get a propType warning on this branch:

image

Other than that, this is looking great, so pre-approving.

client/analytics/report/coupons/table.js
Aljullu: Good catch! I removed it also from the Orders table, which I used as an example to build the Coupons table (eb435e150717918d025bacd9ca7685457dac1d37).

Aljullu:
> So one minor comment in the code, but this is testing well for me. I did get a propType warning on this branch:

Oh, right. Forgot to add `SCRIPT_DEBUG` to `true` after changing my laptop. :sweat_smile: 

I fixed the propType warning in 496a52812b11d4a4b352e243b299a10e9d039621.

PR #868, opened 2018-11-15
Minor fixes to stores

Aljullu:
Small PR with some minor fixes to several stores.

Fixes are very small and they shouldn't have any impact in the way wc-admin works, but an easy way to test would be going to every report page and making sure everything (summary numbers, chart and table data) still loads.

### DIFF
--- a/client/store/orders/selectors.js
+++ b/client/store/orders/selectors.js
@@ -16,7 +16,7 @@ import { ERROR } from 'store/constants';
  * Returns orders for a specific query.
  *
  * @param  {Object} state     Current state
- * @param  {Object} query     Report query paremters
+ * @param  {Object} query     Report query paremeters
  * @return {Array}            Report details
  */
 function getOrders( state, query = {} ) {
@@ -27,7 +27,7 @@ export default {
 	getOrders,
 
 	/**
-	 * Returns true if a query is pending.
+	 * Returns true if a getOrders request is pending.
 	 *
 	 * @param  {Object} state   Current state
 	 * @return {Boolean}        True if the `getOrders` request is pending, false otherwise
@@ -37,7 +37,7 @@ export default {
 	},
 
 	/**
-	 * Returns true if a get orders request has returned an error.
+	 * Returns true if a getOrders request has returned an error.
 	 *
 	 * @param  {Object} state     Current state
 	 * @param  {Object} query     Query parameters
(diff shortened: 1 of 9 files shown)


client/store/orders/selectors.js
```diff
@@ -16,7 +16,7 @@ import { ERROR } from 'store/constants';
  * Returns orders for a specific query.
  *
  * @param  {Object} state     Current state
- * @param  {Object} query     Report query paremters
+ * @param  {Object} query     Report query paremeters
```
allendav: parameters

client/store/products/selectors.js
```diff
@@ -16,7 +16,7 @@ import { getJsonString } from 'store/utils';
  * Returns products report details for a specific report query.
  *
  * @param  {Object} state  Current state
- * @param  {Object} query  Report query paremters
+ * @param  {Object} query  Report query paremeters
```
allendav: parameters

client/store/reports/stats/selectors.js
```diff
@@ -17,7 +17,7 @@ import { getJsonString } from 'store/utils';
  *
  * @param  {Object} state     Current state
  * @param  {String} endpoint  Stats endpoint
- * @param  {Object} query     Report query paremters
+ * @param  {Object} query     Report query paremeters
```
allendav: parameters

allendav (changes requested):
A few small changes needed

client/store/orders/selectors.js
Aljullu: Ouch. :upside_down_face: 

Should be fixed everywhere, now. 614d905fa0921845362b4ad0987188e834253713

client/store/products/resolvers.js
```diff
@@ -15,7 +20,7 @@ export default {
 
 		try {
 			const params = query ? '?' + stringify( query ) : '';
-			const products = await apiFetch( { path: '/wc/v3/reports/products' + params } );
+			const products = await apiFetch( { path: NAMESPACE + 'reports/products' + params } );
```
timmyc: Nice trash pickup here!

timmyc (approved):
This is looking good to me and did not notice any regressions in testing. Thanks for cleaning these items up @Aljullu

PR #876, opened 2018-11-16
Packages: Remove imports from file that doesn’t exist in packages

### LINKED ISSUE #873: Import from `store/constants` in published components is causing error
I encountered this error when trying to use `@woocommerce/components` in a new project:

```
ERROR in ./node_modules/@woocommerce/components/build-module/search/autocompleters/product-cat.js
    Module not found: Error: Can't resolve 'store/constants' in 'woocommerce-gutenberg-products-block/node_modules/@woocommerce/components/build-module/search/autocompleters'
```

Looks like we import the `NAMESPACE` constant, for publishing these we should just switch this to a local variable.

ryelle:
Fixes #873 – Trying to use packages in an external project triggers an error, because we try to import from `store` files that don't exist outside of wc-admin. This PR removes those imports. For `NAMESPACE` we just use `/wc/v3/` directly, and for `DEFAULT_QUERY` we define a local constant.

**To test**

Not sure how best to test this without a local project, but I tested it by running `npm link` in each package, then `npm link @woocommerce/components` & `npm link @woocommerce/date` in the woo blocks plugin, which then let me run webpack to successfully build the project.

Perhaps just a code-review 👀  would be enough.

### DIFF
--- a/packages/components/src/search/autocompleters/coupons.js
+++ b/packages/components/src/search/autocompleters/coupons.js
@@ -13,7 +13,6 @@ import { stringifyQuery } from '@woocommerce/navigation';
  * Internal dependencies
  */
 import { computeSuggestionMatch } from './utils';
-import { NAMESPACE } from 'store/constants';
 
 /**
  * A coupon completer.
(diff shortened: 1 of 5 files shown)


timmyc (approved):
Changes look good to me 👍

PR #877, opened 2018-11-16
Tag component snapshots

belcherj:
Adds snapshot tests for the Tag component.

- `npm test`

What else would you test for this component?  I feel like the snapshots cover all the bases.


packages/components/src/tag/test/index.js
```diff
@@ -0,0 +1,41 @@
...
+		const tree = renderer.create( <Tag label="foo" /> ).toJSON();
+		expect( tree ).toMatchSnapshot();
+	} );
+
+	test( 'it renders correctly with id', () => {
+		const tree = renderer.create( <Tag label="foo" id={ 1 } /> ).toJSON();
+		expect( tree ).toMatchSnapshot();
+	} );
+
+	test( 'it renders correctly with remove', () => {
```
ryelle: this could be `renders the tag and a remove button`

packages/components/src/tag/test/index.js
```diff
@@ -0,0 +1,41 @@
...
+	} );
+
+	test( 'it renders correctly with popover', () => {
+		const tree = renderer
+			.create( <Tag label="foo" popoverContents={ <p>This is a popover</p> } /> )
+			.toJSON();
+		expect( tree ).toMatchSnapshot();
+	} );
+
+	test( 'it renders correctly with popover', () => {
```
ryelle: this name doesn't describe this test correctly.

ryelle (commented):
A few comments about test names. I wonder if you can test for visible/invisible popovers?

packages/components/src/tag/test/index.js
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Internal dependencies
+ */
+import Tag from '../';
+
+const noop = () => {};
+
+describe( 'Tag', () => {
+	test( 'it renders correctly', () => {
```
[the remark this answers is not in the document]
belcherj: Updated

packages/components/src/tag/test/index.js
```diff
@@ -0,0 +1,41 @@
...
+		const tree = renderer.create( <Tag label="foo" /> ).toJSON();
+		expect( tree ).toMatchSnapshot();
+	} );
+
+	test( 'it renders correctly with id', () => {
```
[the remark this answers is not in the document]
belcherj: Removed

packages/components/src/tag/test/index.js
belcherj: Done

packages/components/src/tag/test/index.js
belcherj: Fixed

ryelle (approved):
I'm not sure about adding the JSX into the test name, but the descriptions are better and the tests pass 👍

PR #884, opened 2018-11-18
Add chartMode property to filters configuration

### LINKED ISSUE #879: Product report: Incorrect default chart
**Describe the bug**
The product report is currently using the comparison chart when no products are being compared

**To Reproduce**

On the latest version of master, view the products report, and note the chart layout. 

screen shot 2018-11-16 at 11 51 14 am

**Expected behavior**

The chart should use the standard layout until products are selected and compared. For reference, the screen shot below is the standard layout chart on the revenue report. ( The legend should be in the chart header )

screen shot 2018-11-16 at 11 52 17 am

Aljullu:
Fixes #879.

Adds a `chartMode` property to filters configuration so each filter can specify which kind of chart must be shown when it's activated.

### Screenshots
Notice _All Products_ and _Top Products by Items Sold_ have different chart layouts:
image
image

- Try all filters that appear in the dropdown under _Show:_.
- Make sure only the _All Products_ filter displays the time comparison chart layout (legend on top) while all the others have the item comparison chart layout (legend on the left).

 @psealock I'm marking you as a reviewer since you worked a lot with filters and I would like to know your opinion about this PR.

### DIFF
--- a/client/analytics/components/report-chart/index.js
+++ b/client/analytics/components/report-chart/index.js
@@ -8,10 +8,12 @@ import { compose } from '@wordpress/compose';
 import { format as formatDate } from '@wordpress/date';
 import { withSelect } from '@wordpress/data';
 import PropTypes from 'prop-types';
+import { find, get } from 'lodash';
 
 /**
  * WooCommerce dependencies
  */
+import { flattenFilters } from '@woocommerce/navigation';
 import {
 	getAllowedIntervalsForQuery,
 	getCurrentDates,
@@ -28,17 +30,32 @@ import { Chart, ChartPlaceholder } from 'components';
 import { getReportChartData } from 'store/reports/utils';
 import ReportError from 'analytics/components/report-error';
 
+export const DEFAULT_FILTER = 'all';
+
 class ReportChart extends Component {
+	getAppliedFilter( filterConfig, query ) {
+		const allFilters = flattenFilters( filterConfig.filters );
+		const value = query[ filterConfig.param ] || DEFAULT_FILTER;
+
+		return find( allFilters, { value } );
+	}
+
+	getChartMode() {
+		const { filters, query } = this.props;
+		if ( ! filters ) {
+			return null;
+		}
+		const filterConfig = filters[ 0 ];
+		const appliedFilter = this.getAppliedFilter( filterConfig, query );
+		const appliedFilterParam = get( appliedFilter, [ 'settings', 'param' ] );
+
+		return ! appliedFilterParam || Object.keys( query ).includes( appliedFilterParam )
+			? appliedFilter.chartMode
+			: filterConfig.chartMode;
+	}
+
 	render() {
-		const {
-			comparisonChart,
-			query,
-			itemsLabel,
-			path,
-			primaryData,
-			secondaryData,
-			selectedChart,
-		} = this.props;
+		const { query, itemsLabel, path, primaryData, secondaryData, selectedChart } = this.props;
 
 		if ( primaryData.isError || secondaryData.isError ) {
 			return <ReportError isError />;
@@ -84,6 +101,8 @@ class ReportChart extends Component {
 				},
 			};
 		} );
+		const mode = this.getChartMode();
+		const layout = mode === 'item-comparison' ? 'comparison' : 'standard';
 
 		return (
 			<Chart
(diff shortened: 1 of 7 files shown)


LevinMedia (approved):
Looking great - it's also using the correct chart when comparing products and viewing a single product report. 🎉

client/analytics/components/report-chart/index.js
```diff
@@ -27,17 +29,43 @@ import { Chart, ChartPlaceholder } from 'components';
...
+			const value = query[ filterConfig.param ] || DEFAULT_FILTER;
+			const selectedFilter = find( allFilters, { value } );
+			const selectedFilterParam = get( selectedFilter, [ 'settings', 'param' ] );
+
+			if ( ! selectedFilterParam || Object.keys( query ).includes( selectedFilterParam ) ) {
+				return selectedFilter;
+			}
+		}
+
+		return this.getSelectedFilter( filters, query );
```
psealock: This code will never reach the second item in the `filters` array, ie `variationsConfig`: 

https://github.com/woocommerce/wc-admin/blob/c1e5f4c19cb72242df73182b0cec7d53bbc361f5/client/analytics/report/products/config.js#L116

The `value: 'all'` filter will be selected by default.


You could simplify by importing the first item directly from config to avoid recursively working through the array

client/analytics/report/products/config.js
```diff
@@ -38,10 +38,12 @@ const filterConfig = {
 		{
 			label: __( 'Single Product', 'wc-admin' ),
 			value: 'select_product',
+			chartMode: 'item-comparison',
```
psealock: `chartMode` will need to be added to the propTypes definition in `/components/filter/index.js` for documentation:

https://github.com/woocommerce/wc-admin/blob/fe46fe1955b75fa8923af2e80d54b792c38a994b/packages/components/src/filters/filter/index.js#L237-L260

client/analytics/components/report-chart/index.js
```diff
@@ -27,17 +29,43 @@ import { Chart, ChartPlaceholder } from 'components';
...
+
+	getChartMode() {
+		const { filters, query } = this.props;
+		if ( ! filters ) {
+			return null;
+		}
+		const clonedFilters = filters.slice( 0 );
+		const selectedFilter = this.getSelectedFilter( clonedFilters, query );
+
+		return get( selectedFilter, [ 'chartMode' ] );
```
psealock: default to `time-comparison` ?

psealock (commented):
Looks good, testing well. I left a couple comments.

client/analytics/components/report-chart/index.js
Aljullu: We are starting by the end of the array (https://github.com/woocommerce/wc-admin/pull/884/files/c1e5f4c19cb72242df73182b0cec7d53bbc361f5#diff-e634ee55f27a62796c86de6d587b69b1R40). We test if that filter is visible and if it's applied (https://github.com/woocommerce/wc-admin/pull/884/files/c1e5f4c19cb72242df73182b0cec7d53bbc361f5#diff-e634ee55f27a62796c86de6d587b69b1R42), otherwise we test the previous filters in the array. So it should always find the most specific filter applied.

In the Product report, for example, it will first find if a _Variations_ filter is applied and if not, will find if a _Products_ filter is.

> You could simplify by importing the first item directly from config to avoid recursively working through the array

But that wouldn't handle the case where a subfilter has a different display mode than the parent filter, right?

client/analytics/components/report-chart/index.js
Aljullu: I think I prefer passing `null` to the `Chart` component so it can get the value from the `defaultProps`..

This way, if some time we want to change the default display mode of all charts, we will only have to do that in the component `defaultProps`. How does that sound?

In #818 we are going to merge `layout` and `mode` props into one single prop, so the code will be simplified a bit as well.

client/analytics/report/products/config.js
Aljullu: Fixed!

https://github.com/woocommerce/wc-admin/pull/884/commits/5f81c33a3fb3e31bf59b7ab76844c767f89114c8

Aljullu:
@psealock I have answered some of your comments and rebased the branch. Can you take another look and let me know what you think?

client/analytics/components/report-chart/index.js
psealock: The links you supplied aren't around anymore, but I think I know what you're referring to. 

>In the Product report, for example, it will first find if a Variations filter is applied and if not, will find if a Products filter is.

Gotcha, I was thinking the Variations filter was always going to be the `standard` mode for some reason. Makes sense

client/analytics/components/report-chart/index.js
psealock: Makes sense to me, thanks.

psealock (approved):
LGTM 🚢

client/analytics/components/report-chart/index.js
Aljullu: > The links you supplied aren't around anymore, but I think I know what you're referring to.

My bad, I did a `git push -f` after posting the links. :sweat_smile:

PR #889, opened 2018-11-19
Fix wrong variable name in Legend tests

Aljullu:
Legend tests were using a wrong variable name.

- Verify `npm test` still works.


belcherj (approved):
Great change

PR #893, opened 2018-11-19
Activity Panel: Inbox: Filter update and error notes out of the inbox

### LINKED ISSUE #722: Activity Panel: Inbox: Filter out update and error notices
Right now, the activity panel inbox displays notes of all types. The inbox should NOT be displaying notes of type update or error. Those belong in the separate yet-to-be-built priority notices component instead.

See also #176 (Activity Panel: Inbox) and #115 (Store Alerts aka Priority Notices)

allendav:
Fixes #722

Filters the inbox so that error and update notes do NOT appear. (They belong in the as-yet-not-implemented Store Alerts component.)

- Use the https://github.com/Automattic/wc-admin-notes-test-bench plugin to add notes of all four types.
- Navigate to wp-admin > WooCommerce > Dashboard > Inbox
- Make sure the error and update notes do NOT appear
- Make sure the info and warning notes DO appear
- Run the phpunit tests and make sure they pass
- You can run `WC_Tests_API_Admin_Notes::test_get_warning_notes` separately if so desired.
- You can also play with this in Postman:

screen shot 2018-11-19 at 3 56 29 pm

### DIFF
--- a/includes/api/class-wc-admin-rest-admin-notes-controller.php
+++ b/includes/api/class-wc-admin-rest-admin-notes-controller.php
@@ -109,6 +109,12 @@ class WC_Admin_REST_Admin_Notes_Controller extends WC_REST_CRUD_Controller {
 			'page'     => $page,
 		);
 
+		$type = isset( $request['type'] ) ? $request['type'] : '';
+		$type = sanitize_text_field( trim( $type ) );
+		if ( ! empty( $type ) ) {
+			$args['type'] = $type;
+		}
+
 		$notes = WC_Admin_Notes::get_notes( 'edit', $args );
 
 		$data = array();
--- a/includes/api/class-wc-admin-rest-admin-notes-controller.php
+++ b/includes/api/class-wc-admin-rest-admin-notes-controller.php
@@ -110,7 +110,7 @@ class WC_Admin_REST_Admin_Notes_Controller extends WC_REST_CRUD_Controller {
 		);
 
 		$type = isset( $request['type'] ) ? $request['type'] : '';
-		$type = sanitize_text_field( trim( $type ) );
+		$type = sanitize_text_field( $type );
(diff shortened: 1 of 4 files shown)


psealock:
I'm not sure if I have the notes test bench plugin working correctly. Should I be seeing at least one warning message?

screen shot 2018-11-21 at 11 26 02 am

I do see the `type: info,warning` query added to the request, however

includes/api/class-wc-admin-rest-admin-notes-controller.php
```diff
@@ -109,6 +109,12 @@ public function get_items( $request ) {
 			'page'     => $page,
 		);
 
+		$type = isset( $request['type'] ) ? $request['type'] : '';
+		$type = sanitize_text_field( trim( $type ) );
```
claudiosanches: `sanitize_text_field()` already trim empty spaces.

```bash
$ wp shell
wp> sanitize_text_field( '  text   ' )
=> string(4) "text"
wp> 
```

joshuatf (changes requested):
@allendav This is testing well for me in the dashboard and only info/warnings are showing up 👍 

The phpunit tests are failing though:
```
1) WC_Tests_API_Admin_Notes::test_get_warning_notes
Failed asserting that 2 matches expected 1.
```
It doesn't look like the types are passed to `get_notes()` so `$escaped_where_types` is empty.  Do we want to set these by default to only show info/warning in the API or change the tests to assert a count of 2?

allendav:
@joshuatf wrote:

> It doesn't look like the types are passed to get_notes() so $escaped_where_types is empty

Thanks for catching that - will fix asap

jeffstieler:
This is ready for another review.

timmyc:
@joshuatf could you give this one another 👀 ?

joshuatf (approved):

PR #896, opened 2018-11-20
Merge table stores

### LINKED ISSUE #880: Unify the stores for coupons, products, taxes
The stores for coupons, products and taxes are very similar just changing some variable names. It would make sense to unify them as we did for the stats stores.

That would cleanup the code and might make it easier to add new reports.

Aljullu:
Fixes #880.

Merge table stores for categories, coupons, products and variations.

- Go to the _Categories_, _Coupons_ and _Products_ reports and interact with them as much as possible and verify data is always loaded, there are no JS errors or warnings in the console and everything works fine without regressions. Keep in mind, categories, coupons and variations data is still loaded from SwaggerHub so no real data is displayed.


client/store/reports/utils.js
```diff
@@ -15,12 +15,13 @@ import { formatCurrency } from '@woocommerce/currency';
...
  * Internal dependencies
  */
-import { MAX_PER_PAGE } from 'store/constants';
+import { MAX_PER_PAGE, QUERY_DEFAULTS } from 'store/constants';
 import * as categoriesConfig from 'analytics/report/categories/config';
 import * as couponsConfig from 'analytics/report/coupons/config';
 import * as ordersConfig from 'analytics/report/orders/config';
 import * as productsConfig from 'analytics/report/products/config';
 import * as taxesConfig from 'analytics/report/taxes/config';
+import * as reportsUtils from './utils';
```
psealock: Whats the benefit of doing it this way?

psealock (approved):
screen shot 2018-11-23 at 11 07 09 am
🎉 

Nice abstraction here, thanks for handling it.

Testing well and I don't see any regressions.

PR #898, opened 2018-11-20
Update uses of Gutenberg popover for content sizing

psealock:
Gutenberg 4.5 will remove use of `refresh` on its dropdown component in favour of auto refreshing every 200ms. More on that change in https://github.com/WordPress/gutenberg/pull/11257.

Use of `<Autocomplete />` inside the popover (such as in filters) stopped displaying results which are absolutely positioned. So this PR adds a prop to control the position of the results so the popover can recognise them as occupying space.

screen shot 2018-11-21 at 12 14 59 pm

## Test

1. Use Gutenberg 4.3 or higher
2. Products Report > Filter Single Product
3. Search for a product and make sure the results appear

### DIFF
--- a/packages/components/src/filters/date/content.js
+++ b/packages/components/src/filters/date/content.js
@@ -24,7 +24,7 @@ class DatePickerContent extends Component {
 		this.onTabSelect = this.onTabSelect.bind( this );
 	}
 	onTabSelect( tab ) {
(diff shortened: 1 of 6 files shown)


Aljullu (approved):
I noticed this bug some days ago but didn't have to dig into it. Thanks for working on this @psealock! :clap: 

LGTM

PR #901, opened 2018-11-21
Chart: Persist orderedKeys between state changes

psealock:
Partially solves https://github.com/woocommerce/wc-admin/issues/674

When switching a chart's intervals, the selected or un-selected series should persist.

**_NOTE_: I'm using data that is already in state for this PR**
While data is fetching, the placeholder replaces the chart component and a new `<Chart />` is instantiated when data arrives. Unfortunately, this causes state to be re-constructed. This can be solved by moving the rendering of the placeholder (or perhaps something else). I'd like to tackle this aspect separately.

## Test

1. Choose a Report with a period big enough to allow multiple intervals. "Last Month", for instance.
2. After the chart loads, choose another interval to pre-load the data (for reasons explained in the note above).
3. De-select a time series so it doesn't appear on the chart.
4. Switch back to to original interval.
5. Confirm the series previously unselected remains unselected.

### DIFF
--- a/client/components/chart/index.js
+++ b/client/components/chart/index.js
@@ -7,7 +7,7 @@ import classNames from 'classnames';
 import { Component, createRef } from '@wordpress/element';
 import { decodeEntities } from '@wordpress/html-entities';
 import { formatDefaultLocale as d3FormatDefaultLocale } from 'd3-format';
-import { get, isEqual, partial } from 'lodash';
+import { get, isEqual, partial, find } from 'lodash';
 import Gridicon from 'gridicons';
 import { IconButton, NavigableMenu, SelectControl } from '@wordpress/components';
 import { interpolateViridis as d3InterpolateViridis } from 'd3-scale-chromatic';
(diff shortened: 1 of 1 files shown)


client/components/chart/index.js
```diff
@@ -76,11 +79,11 @@ class Chart extends Component {
 
 	componentDidUpdate( prevProps ) {
 		const { data } = this.props;
-		const orderedKeys = getOrderedKeys( this.props );
 		if ( ! isEqual( [ ...data ].sort(), [ ...prevProps.data ].sort() ) ) {
+			const orderedKeys = getOrderedKeys( this.props, this.state.orderedKeys );
 			/* eslint-disable react/no-did-update-set-state */
 			this.setState( {
-				orderedKeys: orderedKeys,
+				orderedKeys,
```
Aljullu: :+1:

client/components/chart/index.js
```diff
@@ -33,20 +33,23 @@ d3FormatDefaultLocale( {
...
 			}, [] )
 		),
-	].map( key => ( {
-		key,
-		total: props.data.reduce( ( a, c ) => a + c[ key ].value, 0 ),
-		visible: true,
-		focus: true,
-	} ) );
+	].map( key => {
+		const previousKey = find( previousOrderedKeys, item => key === item.key );
```
Aljullu: When there is a ES6 method equivalent to a lodash method, I think using ES6 makes it easy for contributors who might not be used to lodash to understand the code. What do you think?
```suggestion
		const previousKey = previousOrderedKeys.find( item => key === item.key );
```

Aljullu (approved):
Works as you described and code looks good (just added an opinion comment :slightly_smiling_face: ).

Looks good to me. :rocket:

PR #902, opened 2018-11-21
Chart: persist orderedKeys when requesting data

### LINKED ISSUE #674: Remember chart setting when switching between reports
For each session, remember the chart settings while switching between reports

Settings include

Chart interval (hour, day, week, month, quarter, year)

psealock:
Fixes #674 
Follows #901 

Make selection/un-selection of chart time series persistent when data is being fetched.

### Test

1. Go to a report with a long enough period to allow more than one interval (day, week).
2. De-select a time series
3. Select a different interval whose data is not already in state. 
4. When data arrives and the chart populates, confirm the de-selected series is still deselected.

**Note:** It seems that the selector `isReportStatsRequesting` isn't so reliable as a detector of an in-flight request. It often returns false several moments before data actually arrives, resulting in a chart looking like this:

screen shot 2018-11-22 at 1 46 04 pm

Has anyone else experienced this?

### DIFF
--- a/client/components/chart/index.js
+++ b/client/components/chart/index.js
@@ -4,10 +4,10 @@
  */
 import { __ } from '@wordpress/i18n';
 import classNames from 'classnames';
-import { Component, createRef } from '@wordpress/element';
+import { Component, createRef, Fragment } from '@wordpress/element';
 import { decodeEntities } from '@wordpress/html-entities';
 import { formatDefaultLocale as d3FormatDefaultLocale } from 'd3-format';
-import { get, isEqual, partial, find } from 'lodash';
+import { get, isEqual, partial } from 'lodash';
 import Gridicon from 'gridicons';
 import { IconButton, NavigableMenu, SelectControl } from '@wordpress/components';
 import { interpolateViridis as d3InterpolateViridis } from 'd3-scale-chromatic';
@@ -26,6 +26,7 @@ import { H, Section } from '@woocommerce/components';
 import './style.scss';
 import D3Chart from 'components/d3chart';
 import Legend from 'components/d3chart/legend';
+import ChartPlaceholder from './placeholder';
 
 d3FormatDefaultLocale( {
 	decimal: '.',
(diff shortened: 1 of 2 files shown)


client/components/chart/index.js
```diff
@@ -81,7 +82,13 @@ class Chart extends Component {
 	componentDidUpdate( prevProps ) {
 		const { data } = this.props;
 		if ( ! isEqual( [ ...data ].sort(), [ ...prevProps.data ].sort() ) ) {
-			const orderedKeys = getOrderedKeys( this.props, this.state.orderedKeys );
+			/**
+			 * Only update the orderedKeys when data is present so that
+			 * selection may persist while requesting new data.
+			 */
+			const orderedKeys = data.length
```
greenafrican: Might be better to move the `data.length` test into the `if` statement above?

client/components/chart/index.js
```diff
@@ -399,6 +416,10 @@ Chart.propTypes = {
 	 * What type of data is to be displayed? Number, Average, String?
 	 */
 	valueType: PropTypes.string,
+	/**
+	 * Render a chart placeholder to signify an in-flight data request.
+	 */
+	isRequesting: PropTypes.string,
```
greenafrican: Should this be a `boolean`? I'm getting a warning:

screenshot 2018-11-22 09 55 58

greenafrican (approved):
One small code suggestion but this works as expected and once the warning is cleared it is good to merge.

client/components/chart/index.js
psealock: I tried that, but I still want `visibleData` to be updated when `data = []` so the displayed data series are removed. If `isRequesting` were truly accurate (see NOTE in original comment) this wouldn't be an issue because the placeholder would be shown instead. 

I'm thinking I'll leave it like this for now

Aljullu:
> Has anyone else experienced this?

Yes, I faced the same issue in #908. I was trying to figure out what was going on without luck yet.

PR #905, opened 2018-11-21
Default charts to 'time-comparison'

Aljullu:
Charts were defaulting to `item-comparison`, but they should default to `time-comparison`.

### Screenshots
Before:
image

After:
image

- Go to the _Revenue_ report.
- Make sure the chart is shown correctly (like the second screenshot).

### DIFF
--- a/client/analytics/components/report-chart/index.js
+++ b/client/analytics/components/report-chart/index.js
@@ -57,7 +57,7 @@ export class ReportChart extends Component {
 	getChartMode() {
 		const { filters, query } = this.props;
 		if ( ! filters ) {
-			return null;
+			return;
 		}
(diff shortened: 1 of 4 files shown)


client/analytics/components/report-chart/index.js
```diff
@@ -57,7 +57,7 @@ export class ReportChart extends Component {
 	getChartMode() {
 		const { filters, query } = this.props;
 		if ( ! filters ) {
-			return null;
+			return;
```
psealock: It looks like you'll need to fix a unit test with `undefined` v. `null`.

Aljullu: Completely missed that one. Thank you!

Issue #854, opened 2018-11-15
Charts: Error: <g> attribute transform: Trailing garbage, "translate(0,NaN)"


allendav:
**Describe the bug**

Occurs when loading the wp-admin > Revenue > Analytics page. The following appears in the console:

```

  | (anonymous) | @ | attr.js:31
-- | -- | -- | --
  | wc.components../node_modules/d3-selection/src/selection/each.js.__webpack_exports__.default | @ | each.js:5
  | wc.components../node_modules/d3-selection/src/selection/attr.js.__webpack_exports__.default | @ | attr.js:53
  | axis | @ | axis.js:102
  | wc.components../node_modules/d3-selection/src/selection/call.js.__webpack_exports__.default | @ | call.js:4
  | drawAxis | @ | utils.js:438
  | drawChart | @ | charts.js:89
  | drawChart


```

**To Reproduce**

1. Go to the wp-admin > Revenue > Analytics page
2. See error (8x)

**Expected behavior**

No error

**Screenshots**

<img width="1433" alt="screen shot 2018-11-14 at 4 37 20 pm" src="https://user-images.githubusercontent.com/1595739/48521971-a4262080-e82b-11e8-8030-f4b237a3ddce.png">

**Versions**

WP 5 w/o Gutenberg plugin

Aljullu:
This error appears when we are trying to render the chart but there is no data (#582). What's weird here and I couldn't reproduce yet is that we are trying to render the chart before the data is loaded.

timmyc:
Closing in favor of #582 as we should not render the chart without data.

PR #908, opened 2018-11-22
Make ChartPlaceholder have the same height than the resulting Chart

### LINKED ISSUE #869: Chart placeholder doesn't have the same height than final chart
This produces a scroll jump if you are below the chart.

image

image

This issue affects both, the chart with the legend at the bottom and when it's on the side/top.

Aljullu:
Fixes #869.

### Screenshots
Before:
image

After:
image

There seems to be a bug that makes stores return `false` for `isReportStatsLoading()` before the data is loaded which I think is related to #854.

So, for now, we will have to force `isRequesting` to be true modifying [this line](https://github.com/woocommerce/wc-admin/blob/master/client/analytics/components/report-chart/index.js#L122) from:
```ES6
isRequesting={ primaryData.isRequesting || secondaryData.isRequesting }
```
to:
```ES6
isRequesting={ true }
```
- Open the _Products_ report in a tab.
- Make the code change described above.
- Open the _Products_ report in another tab (so one will be displaying the chart and the other one the placeholder).
- Verify they have the same height.
- You can repeat the same with different window sizes and comparison charts (setting the filter to _Top Products by Items Sold_ in the _Products_ report, for example).

### DIFF
--- a/client/components/chart/index.js
+++ b/client/components/chart/index.js
@@ -312,7 +312,7 @@ class Chart extends Component {
 								<span className="screen-reader-text">
 									{ __( 'Your requested data is loading', 'wc-admin' ) }
 								</span>
-								<ChartPlaceholder />
+								<ChartPlaceholder height={ chartHeight } />
 							</Fragment>
 						) }
 						{ ! isRequesting &&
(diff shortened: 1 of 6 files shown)


greenafrican (approved):
Thanks for updating bits of the chart README's and this works well and looks much better, to the point where I can now see that the height of the summary item placeholders needs a little work so that we get a smooth transition :)

:shipit:

PR #910, opened 2018-11-23
Bug: noticed a new NaN error caused by attempt to draw chart with empty data

greenafrican:
I noticed an error while reviewing #908 where the chart is being called with an empty data array.

### Screenshots
screenshot 2018-11-23 10 12 26

- Go to _Products Report_ 
- Change to an interval that isn't cached
- Ensure there aren't any `NaN` errors in the console

### DIFF
--- a/client/components/chart/index.js
+++ b/client/components/chart/index.js
@@ -7,7 +7,7 @@ import classNames from 'classnames';
 import { Component, createRef, Fragment } from '@wordpress/element';
 import { decodeEntities } from '@wordpress/html-entities';
 import { formatDefaultLocale as d3FormatDefaultLocale } from 'd3-format';
-import { get, isEqual, partial } from 'lodash';
+import { get, isEmpty, isEqual, partial } from 'lodash';
 import Gridicon from 'gridicons';
 import { IconButton, NavigableMenu, SelectControl } from '@wordpress/components';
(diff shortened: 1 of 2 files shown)


psealock:
>In that case, probably a different UI stating there is no data would be better than the loading placeholder. What do you think?

I agree with this. @LevinMedia maybe this could use some design love?

greenafrican:
> Chart data being empty doesn't always mean it's loading: for example, when I go to the Products report and select Today as the date range, I get an empty array of data, even after loading finished, just because there is no data.

If you select a date and there isn't any data you should still get an array, for example for today (with no orders):
```
[{
	"date": "2018-11-26T00:00:00",
	"Today (Nov 26, 2018)": {
		"labelDate": "2018-11-26 00:00:00",
		"value": 0
	},
	"Previous Period (Nov 25, 2018)": {
		"labelDate": "2018-11-25 00:00:00",
		"value": 0
	}
}]
```

>probably a different UI stating there is no data

I think these are 2 different issues, perhaps this PR has a better fix to properly implement `isRequesting` and agree with @psealock that we could use some design assistance to gracefully handle the loading and no data statuses.

greenafrican:
@psealock @Aljullu 

Rethinking this a bit and I feel that #911 should resolve this for this project. I'm going to use this PR to put in a check inside the `d3chart` component too as this may resurface in other implementations of the component.

PR #913, opened 2018-11-23
Use numberFormat() to format all non-currency numbers in tables

Aljullu:
Fixes https://github.com/woocommerce/wc-admin/pull/904#discussion_r236018943.

Tables were displaying many numbers in rows and summaries which could be formatted with `numberFormat()`.

### Screenshots
Before:
image

After:
image

- Go to every report page and verify table numbers are correctly formatted.

Note that numbers <=999 look the same when they are formatted and they are not, so it might make sense to modify some of them in the code.


psealock (approved):
LGTM

PR #928, opened 2018-11-25
Merge report tables

Aljullu:
Follow-up of #896.

As we did with `<ReportChart>`, I merged most report tables into a single component `<ReportTable>`. Tables are a little bit different than charts because every report needs a different method to render table headers, rows and summary, but I still think this will make creating future report tables easier.

### Screenshots
_(there shouldn't be any visible change)_

- Go to the _Categories_, _Coupons_, _Products_ (including the filter combination that displays the _Variations_ table). and _Taxes_ reports.
- Interact with tables: sorting, using pagination, downloading the CSV, hidding/showing columns, etc.
  _Notice:_ from those tables, only the _Products_ one has real data, all the others use SwaggerHub to render placeholder entries.

### DIFF
--- a/client/analytics/components/report-table/index.js
+++ b/client/analytics/components/report-table/index.js
@@ -31,10 +31,10 @@ class ReportTable extends Component {
 			tableData,
 			totalsCountField,
 			// These two props are not used in the render function, but are destructured
-			// so they are not included in the `props` variable.
+			// so they are not included in the `tableProps` variable.
 			endpoint,
 			tableQuery,
-			...props
+			...tableProps
 		} = this.props;
 
 		const { items, query } = tableData;
@@ -66,7 +66,7 @@ class ReportTable extends Component {
 				rowsPerPage={ query.per_page }
 				summary={ summary }
 				totalRows={ totalRows }
-				{ ...props }
+				{ ...tableProps }
 			/>
 		);
 	}
--- a/client/analytics/components/report-table/index.js
+++ b/client/analytics/components/report-table/index.js
@@ -1,0 +1,138 @@
+/** @format */
+/**
+ * External dependencies
+ */
+import { Component } from '@wordpress/element';
+import { compose } from '@wordpress/compose';
+import { withSelect } from '@wordpress/data';
+import { get, orderBy } from 'lodash';
+import PropTypes from 'prop-types';
+
+/**
+ * WooCommerce dependencies
+ */
+import { TableCard } from '@woocommerce/components';
+import { onQueryChange } from '@woocommerce/navigation';
+
+/**
+ * Internal dependencies
+ */
+import ReportError from 'analytics/components/report-error';
+import { getReportChartData, getReportTableData } from 'store/reports/utils';
+
+class ReportTable extends Component {
+	render() {
+		const {
+			getHeadersContent,
+			getRowsContent,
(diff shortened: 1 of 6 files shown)


client/analytics/components/report-table/index.js
```diff
@@ -0,0 +1,138 @@
...
+		const {
+			getHeadersContent,
+			getRowsContent,
+			getSummary,
+			itemIdField,
+			primaryData,
+			tableData,
+			totalsCountField,
+			// These two props are not used in the render function, but are destructured
+			// so they are not included in the `props` variable.
```
psealock: Does it matter?

client/analytics/components/report-table/index.js
```diff
@@ -0,0 +1,138 @@
...
+			endpoint,
+			tableQuery,
+			...props
```
psealock: `tableProps` ?

client/analytics/components/report-table/index.js
```diff
@@ -0,0 +1,138 @@
...
+	tableQuery: {},
+	totalsCountField: '',
+};
+
+export default compose(
+	withSelect( ( select, props ) => {
+		const { endpoint, query, tableQuery } = props;
+		// @TODO allow loading the primary data for the variations table once #926 is fixed.
+		const primaryData =
+			endpoint === 'variations' ? {} : getReportChartData( endpoint, 'primary', query, select );
```
psealock: Good point. When looking at a single product the chart will display data from `/reports/products/stats` segmented by variations. Would it be possible to construct `primaryData.data.totals` for the table summary numbers without having to query a new endpoint?

Just curious here, even though I think simplicity of a new endpoint may be better.

psealock (approved):
Nice abstraction here (again). I just have some questions/thoughts.

client/analytics/components/report-table/index.js
Aljullu: I wanted to add the comment to make sure in the future we remember we are defining those variables in order to exclude them from the `tableProps` object. Otherwise, doing a cleanup of the code we might remove `endpoint` and `tableQuery` from the destructuring thinking defining them was a mistake.

That would not cause an error, so we would not notice it, but I think it's better not to pass unnecessary props down to `<TableCard>`.

But I don't have a strong opinion here, so I would be happy removing the comment if you think it's clear without it.

client/analytics/components/report-table/index.js
Aljullu: I didn't think about that. Would it be possible to update SwaggerHub or post an example here of how the endpoint response would look like to make sure we are in the same page before implementing it?

client/analytics/components/report-table/index.js
psealock: >Otherwise, doing a cleanup of the code we might remove endpoint and tableQuery from the destructuring thinking defining them was a mistake.

Thats a good idea. Thanks for the clarification

client/analytics/components/report-table/index.js
psealock: >Would it be possible to update SwaggerHub or post an example here of how the endpoint response would look like

I think thats a good idea because we'll also run into the same issue on comparisons. I need to think about this some more. Lets tackle this separately.

PR #931, opened 2018-11-26
Updated PHPCS to follow core standards

claudiosanches:
Now [WooCommerce Sniffs](https://packagist.org/packages/woocommerce/woocommerce-sniffs) loads all PHPCS dependencies, also includes https://packagist.org/packages/phpcompatibility/phpcompatibility-wp that do a better job looking for PHP compatibility when using WP.

Note that most of the rules excluded from `phpcs.xml.dist` is due WPCS 1.0, and some of the common rules now are part of WooCommerce Sniffs: 

https://github.com/woocommerce/woocommerce-sniffs/blob/master/src/WooCommerce-Core/ruleset.xml

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,12 +10,8 @@
     "composer/installers": "~1.6"
   },
   "require-dev": {
-    "squizlabs/php_codesniffer": "*",
-    "wp-coding-standards/wpcs": "1.2.0",
     "phpunit/phpunit": "6.5.13",
-    "woocommerce/woocommerce-sniffs": "*",
-    "wimg/php-compatibility": "9.0.0",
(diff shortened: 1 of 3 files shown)


claudiosanches:
@ryelle yes, it's intentional, but we can go fixing while changing each file, or before move into core, still nice too have it here and make sure that we enforce the same standards in both places.

psealock:
To test this, I `composer install` and ran phpcs on files such as `/includes` and only got a handful of warnings. Did I miss a step?

claudiosanches:
@psealock no that's it, should throw warnings, we can fix as we go for each file or before get into core, but good if we can fix things while developing.

PR #939, opened 2018-11-27
Components: Remove import from local files

ryelle:
Looks like more local imports got merged with #881 – Since these are exported packages, they can’t rely on any local files. We should be using just the string `/wc/v3/` , instead of the undefined NAMESPACE constant. See #876 for the previous fix.

**To test**

You need an external project, but if you have one, you can run `npm link` from the components dir, and then `npm link @woocommerce/components` from the external project. That project should have errors with master, but be fixed here.


timmyc (approved):
Thanks for fixing this up @ryelle and sorry this got merged in. Would it be possible for us to setup a linter to prevent this from happening in the `packages` directory?

timmyc:
Merging in prior to doing a build.

PR #944, opened 2018-11-27
Component: add external d3chart package

greenafrican:
This PR attempts to move the chart component and d3chart component into the @woocommerce/components package. It incorporates the update from #961 

### Screenshots
screenshot 2018-11-28 15 07 04

- navigate to `WooCommerce > Analytics > Products` report
- select a date range with known orders
- toggle the chart between intervals and type (bar or line)
- hover over legend items and select/deselect items
- hover over the chart to ensure the tooltip renders as expected
- tab through the page to ensure the chart accessibility features respond as it expected


packages/components/src/chart/d3chart/style.scss
```diff
@@ -1,5 +1,140 @@
 /** @format */
 
+.d3-chart__container {
```
Aljullu: Is there any reason to merge `legend.scss` and `style.scss` in the same file? I think both files are quite big already so it's easier to maintain them if they are split.

greenafrican:
@Aljullu I can't seem to reproduce the height issue above. Any ideas on reproducing it?

Aljullu:
My steps to reproduce:
- Go to the _Orders_ report.
- Make the window `1000px` wide.
- Notice the legend moves below the table, but it's to high.

Let me know if that doesn't work for you!

greenafrican:
height

@Aljullu I can't seem to reproduce.

PR #946, opened 2018-11-27
Roll version to 0.2.0

allendav:
In preparation for release

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "wc-admin",
-  "version": "0.1.0",
+  "version": "0.2.0",
   "main": "js/index.js",
   "author": "Automattic",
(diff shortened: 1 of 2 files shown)


ryelle (approved):
Looks good, plugin version is correct in wp-admin 👍

timmyc:
Did this need to be merged prior to uploading the .zip file @allendav - well the version number shows correctly from the .zip file that I downloaded.

allendav:
> Did this need to be merged prior to uploading the .zip file @allendav - well the version number shows correctly from the .zip file that I downloaded.

The ZIP was cut from the branch. That's how we used to do it at hydra so we could do x.y.N releases if needed. Merging this now.

PR #947, opened 2018-11-27
Add unit tests to /reports/variations

psealock:
Fixes https://github.com/woocommerce/wc-admin/issues/934
Continues from https://github.com/woocommerce/wc-admin/pull/930

Adds a test suite similar to the one for `/reports/products`.

## Test

Run `phpunit`

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-variations-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-variations-data-store.php
@@ -129,7 +129,7 @@ class WC_Admin_Reports_Variations_Data_Store extends WC_Admin_Reports_Data_Store
 				$extended_attributes = apply_filters( 'woocommerce_rest_reports_products_extended_attributes', $this->extended_attributes, $product_data );
 				foreach ( $extended_attributes as $extended_attribute ) {
 					$function = 'get_' . $extended_attribute;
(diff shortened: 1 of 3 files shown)


psealock:
Thanks for the review @claudiosanches !

I realised I was missing some tests around `extended_info`, so I added those in https://github.com/woocommerce/wc-admin/pull/947/commits/76a36baa2cdb3a5db68600a9905e3682bc944728

This also uncovered a bug where `extended_info => price` was referencing the parent product's price and not the variation. I fixed that in this PR.

includes/data-stores/class-wc-admin-reports-variations-data-store.php
```diff
@@ -129,7 +129,7 @@ protected function include_extended_info( &$products_data, $query_args ) {
 				$extended_attributes = apply_filters( 'woocommerce_rest_reports_products_extended_attributes', $this->extended_attributes, $product_data );
 				foreach ( $extended_attributes as $extended_attribute ) {
 					$function = 'get_' . $extended_attribute;
-					if ( is_callable( array( $product, $function ) ) ) {
+					if ( is_callable( array( $product, $function ) ) && 'get_price' != $function ) {
```
claudiosanches: This will trigger warnings on PHPCS, need to use `!==`.

PR #943, opened 2018-11-27
Adding support for advanced filters in orders reports.

peterfabian:
Fixes #749 

Adding support for advanced filters in orders reports.
Plus includes a few fixes for issues discovered along the way in testing.

The PHPCS problem will be sorted out separately.

### Screenshots
N/A

### Detailed test instructions:

- Please see #749 .

PR #951, opened 2018-11-28
Remove order statuses "wc-" prefix

psealock:
#943 assumes order statuses without the `wc-` prefix. Elsewhere in the app we use statuses without the prefix as well. 

This PR changes `wcSettings.orderStatuses` such that object keys are stripped of `wc-`.

## Test

Ensure no regressions are experienced with stuff related to order statuses.

1. Go to _Products Report_ and see the order statuses correctly displayed in the table.
2. In the console, check `wcSettings.orderStatuses` to see if it looks OK.


Aljullu (approved):
Looks good to me. :+1: 

> Go to _Products Report_ and see the order statuses correctly displayed in the table.

You mean the _Orders Report_?

Issue #900, opened 2018-11-21
REST API: /reports/categories should return 'category_name' for every category and /reports/categories/stats should return 'categories_count'


Aljullu:
According to the designs (p6riRB-3dS-p2) some changes must be done to the `categories` endpoints which are not included in SwaggerHub:
- `/reports/categories` should return `category_name` for every category.
- `/reports/categories/stats` should return `categories_count` or there should be some other way to know the total number of categories.

peterfabian:
- as mentioned in p7bje6-151-p2, I assumed the name of category would be fetched from already existing `/products/categories/{id}`, as `/reports/categories` response contains `id` to fetch further info about categories from existing endpoint. If needed, we can add name also to `/reports/categories`, but we would prefer to not duplicate data in different endpoints

- should the `categories_count` show number of categories of products sold in the selected time period or the total of all categories in the shop, irrespective if any products have been sold from the category? Not sure if available in JS, but there is a `X-WP-Total` header returned from `/wp-json/wc/v3/products/categories` that shows total number of categories.

Aljullu:
Thank you for your replies, @peterfabian! I will be answering them in the following days in parallel as I work on them.

> should the categories_count show number of categories of products sold in the selected time period or the total of all categories in the shop, irrespective if any products have been sold from the category? Not sure if available in JS, but there is a X-WP-Total header returned from /wp-json/wc/v3/products/categories that shows total number of all categories.

Thanks for pointing me to that! I saw `/reports/products` also includes the `X-WP-Total` header. Will other endpoints (like `/reports/categories`, `/reports/coupons`, etc.) include it too? In that case, I can use that one which is exactly what I was looking for.

peterfabian:
Yes, `X-WP-Total` should be included in all endpoints.

PR #952, opened 2018-11-28
Use X-WP-Total header to set the total number of items in tables

Aljullu:
From discussion in #900, we will use `X-WP-Total` response header to set the total number of items in tables.

This will affect all tables that use the tables store, but for now only the _Products_ table uses real data from the API.

- Make sure the number of total items shown in the summary (_X products sold_) matches the number of rows in the table.
- Verify also that pagination works as expected.

### DIFF
--- a/client/analytics/components/report-table/index.js
+++ b/client/analytics/components/report-table/index.js
@@ -29,7 +29,6 @@ class ReportTable extends Component {
 			itemIdField,
 			primaryData,
 			tableData,
-			totalsCountField,
 			// These two props are not used in the render function, but are destructured
 			// so they are not included in the `tableProps` variable.
 			endpoint,
@@ -38,6 +37,7 @@ class ReportTable extends Component {
 		} = this.props;
 
 		const { items, query } = tableData;
+		const { data: itemsData, totalResults } = items;
 
 		const isError = tableData.isError || primaryData.isError;
 
@@ -48,12 +48,11 @@ class ReportTable extends Component {
 		const isRequesting = tableData.isRequesting || primaryData.isRequesting;
 
 		const headers = getHeadersContent();
-		const orderedItems = orderBy( items, query.orderby, query.order );
+		const orderedItems = orderBy( itemsData, query.orderby, query.order );
 		const ids = orderedItems.map( item => item[ itemIdField ] );
 		const rows = getRowsContent( orderedItems );
 		const totals = get( primaryData, [ 'data', 'totals' ], null );
 		const summary = getSummary( totals );
-		const totalRows = get( totals, [ totalsCountField ], items.length );
 
 		return (
 			<TableCard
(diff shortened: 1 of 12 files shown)


client/store/reports/items/selectors.js
```diff
@@ -21,7 +21,7 @@ import { getJsonString } from 'store/utils';
  * @return {Object}           Report details
  */
 function getReportItems( state, endpoint, query = {} ) {
-	return get( state, [ 'reports', 'items', endpoint, getJsonString( query ) ], [] );
+	return get( state, [ 'reports', 'items', endpoint, getJsonString( query ) ], {} );
```
psealock: Do you think it would be worthwhile to make the default object `{ data: [] }` ?

client/store/reports/items/test/selectors.js
```diff
@@ -26,23 +26,28 @@ const queryKey = getJsonString( query );
 const endpoint = 'coupons';
 
 describe( 'getReportItems()', () => {
-	it( 'returns an empty array when no items are available', () => {
+	it( 'returns an empty object when no items are available', () => {
 		const state = deepFreeze( {} );
-		expect( getReportItems( state, endpoint, query ) ).toEqual( [] );
+		expect( getReportItems( state, endpoint, query ) ).toEqual( {} );
```
psealock: This would also need to change if the default is changed

psealock (approved):
Testing well. Unit tests pass. I just left a small comment

client/analytics/components/report-table/index.js
```diff
@@ -109,25 +107,17 @@ ReportTable.propTypes = {
...
 };
 
 export default compose(
 	withSelect( ( select, props ) => {
-		const { endpoint, query, tableQuery } = props;
-		// @TODO allow loading the primary data for the variations table once #926 is fixed.
-		const primaryData =
-			endpoint === 'variations' ? {} : getReportChartData( endpoint, 'primary', query, select );
+		const { endpoint, getSummary, query, tableQuery } = props;
+		const primaryData = getSummary ? getReportChartData( endpoint, 'primary', query, select ) : {};
```
psealock: `getSummary` will always be truthy because there is a defaultProp set

client/store/reports/items/selectors.js
Aljullu: Sounds good! :+1:

client/analytics/components/report-table/index.js
Aljullu: I thought like you but I tested it and it seems `defaultProps` are assigned after `withSelect` is executed. So here, `getSummary` might be false.

Anyway, since that's a behavior that might mislead us in the future, I refactored the code to make it easier to understand: 5c314c75d71c58f505ee6266e2b37add2787d0a0.

PR #955, opened 2018-11-28
Fix false positives in Reports stats resolvers tests

Aljullu:
There were false positive tests in `client/store/reports/stats/test/resolvers.js`. This PR fixes that.

- Run `npm test` and verify tests pass.
- Switch `REPORT_2` with `REPORT_1` in `client/store/reports/stats/test/resolvers.js` and verify tests fail (they were passing, before)


psealock (approved):
LGTM 🚢

PR #959, opened 2018-11-28
Don't encode search text in autocompleters before stringifying it

### LINKED ISSUE #827: Search suggestions don't work for multi-word entries
If I have a product named `sit nihil iusto id sed` and I search for `sit n` in the _Single product_ filter of the _Product_ report, I would expect it to appear in the suggestions, but it currently doesn't.

image

image

Aljullu:
Fixes #827.

We were encoding the search parameters twice, that broke searching for item names with special characters like spaces.

The fact is that `stringifyQuery()` calls `qs.stringify()` ([#](https://github.com/ljharb/qs#stringifying)) which by default already encodes the output, so there is no need to encode it before.

### Screenshots
image

- Go to the _Products_ report and select the _Single Product_ filter.
- Start writing the name of a product which contains spaces.
- Verify the product appears even when you write the space and the following letter in the search box.
- In the same search input, try breaking the search, for example, writing `?orderby=name` and verify via the Chrome devtools that's encoded to `%3Forderby%3Dname
`. So users' input is encoded before it's send to the server.

### DIFF
--- a/packages/components/src/search/autocompleters/coupons.js
+++ b/packages/components/src/search/autocompleters/coupons.js
@@ -27,7 +27,7 @@ export default {
 		let payload = '';
 		if ( search ) {
 			const query = {
-				search: encodeURIComponent( search ),
+				search,
 				per_page: 10,
 			};
 			payload = stringifyQuery( query );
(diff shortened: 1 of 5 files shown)


psealock (approved):
🚢

PR #930, opened 2018-11-26
REST API: Add /reports/variations

psealock:
## Methodology

* Add `variation_id` column to the `wp_wc_order_product_lookup` table.
* Group results of querying this table by `variation_id`.
* Add attribute data as part of `extended_product_info = true` query arg.

A sample response:

```js
[
    {
        "product_id": 12,
        "variation_id": 28,
        "items_sold": 2,
        "gross_revenue": 84,
        "orders_count": 1,
        "extended_info": {
            "name": "Hoodie",
            "price": 42,
            "image": "<img width=\"324\" height=\"324\" src=\"http://tangaroa.test/wp-content/uploads/2018/11/hoodie-2-324x324.jpg\" class=\"attachment-woocommerce_thumbnail size-woocommerce_thumbnail\" alt=\"\" srcset=\"http://tangaroa.test/wp-content/uploads/2018/11/hoodie-2-324x324.jpg 324w, http://tangaroa.test/wp-content/uploads/2018/11/hoodie-2-150x150.jpg 150w, http://tangaroa.test/wp-content/uploads/2018/11/hoodie-2-300x300.jpg 300w, http://tangaroa.test/wp-content/uploads/2018/11/hoodie-2-768x768.jpg 768w, http://tangaroa.test/wp-content/uploads/2018/11/hoodie-2-416x416.jpg 416w, http://tangaroa.test/wp-content/uploads/2018/11/hoodie-2-100x100.jpg 100w, http://tangaroa.test/wp-content/uploads/2018/11/hoodie-2.jpg 801w\" sizes=\"(max-width: 324px) 100vw, 324px\" />",
            "permalink": "http://tangaroa.test/product/hoodie/",
            "attributes": [
                {
                    "id": 1,
                    "name": "color",
                    "option": "Red"
                },
                {
                    "id": 0,
                    "name": "logo",
                    "option": "No"
                }
            ]
        },
        "_links": {
            "product": [
                {
                    "href": "http://tangaroa.test/wp-json/wc/v3/products/12"
                }
            ],
            "variation": [
                {
                    "href": "http://tangaroa.test/wp-json/wc/v3/products/12/variation/28"
                }
            ]
        }
    },
]
```

## Test

Using [Postman](https://www.getpostman.com/) with an API key generated by a test site, check the endpoint for correct responses.

1. Create a product(s) with variations.
1. Create completed orders for those variations.
1. Or, if you have already created these, you may need to `Rebuild reports` (under WC > Status > Tools).
[...]

PR #960, opened 2018-11-29
Variations: Add table data

psealock:
Continues from #930 

Hook up the Variations table in Product Detail report to the new `/reports/variations` endpoint.

screen shot 2018-11-29 at 12 45 39 pm

## Caveats

* Summary numbers at the bottom of the table will show product data. When segmentation by variation comes, we can fix this.
* The chart reflects all data for the product, not just variations. Again, we're waiting on segmentation.

## Questions

* Variation Title: Since we don’t filter products by single variation, where does this link to?
* Orders Count: Since we don’t advanced filter the orders report by variation,  where does this link to?

cc @LevinMedia 

## Test

1. Create some orders for a product with variations (Hoodie works if you have sample data)
1. _Products Report_
2. Single Product > Hoodie
3. See the Orders of variations appear
4. Sort columns, use Variations filters

### DIFF
--- a/client/analytics/components/report-table/index.js
+++ b/client/analytics/components/report-table/index.js
@@ -116,7 +116,8 @@ ReportTable.defaultProps = {
 export default compose(
 	withSelect( ( select, props ) => {
 		const { endpoint, getSummary, query, tableQuery } = props;
-		const primaryData = getSummary ? getReportChartData( endpoint, 'primary', query, select ) : {};
+		const chartEndpoint = 'variations' === endpoint ? 'products' : endpoint;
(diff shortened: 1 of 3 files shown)


coderkevin (approved):
Looks good to me, just one optional suggestion.

LevinMedia:
>Variation Title: Since we don’t filter products by single variation, where does this link to?

Can we link this level to the actual product page? 

> Orders Count: Since we don’t advanced filter the orders report by variation, where does this link to?

Would it be possible to link to the orders report filtered by the parent product?

psealock:
>Can we link this level to the actual product edit page?

yes

>Would it be possible to link to the orders report filtered by the parent product?

yes


Thanks @LevinMedia

PR #961, opened 2018-11-29
Make sure D3Base always updates when width changes

### LINKED ISSUE #907: Switching between chart directions doesn't update the chart width
- Change the filter to _Top Products by Items Sold_.
- Notice the chart overflows.
- Reload the page.
- Change the filter to _All Products_.
- Notice the chart doesn't cover the available space.
image

Aljullu:
Fixes #907.

In some circumstances, `D3Base` was not redrawing the chart even thought the width of the chart changed.

This PR fixes that, removes the state from `D3Base` and moves the `propTypes` at the bottom of the component, as we have in most of the other components.

### Screenshots
image

- Change the filter to _Top Products by Items Sold_.
- Notice the chart doesn't overflow.
- Reload the page.
- Change the filter to _All Products_.
- Notice the chart covers the available space.

### DIFF
--- a/client/components/d3chart/d3-base/index.js
+++ b/client/components/d3chart/d3-base/index.js
@@ -29,12 +29,11 @@ export default class D3Base extends Component {
 	constructor( props ) {
 		super( props );
 
-		this.getUpdatedParams = this.getUpdatedParams.bind( this );
 		this.chartRef = createRef();
 	}
 
 	componentDidMount() {
-		window.addEventListener( 'resize', this.getUpdatedParams );
+		window.addEventListener( 'resize', this.drawChart );
 
 		this.drawUpdatedChart();
 	}
@@ -43,6 +42,7 @@ export default class D3Base extends Component {
 		return (
 			this.props.className !== nextProps.className ||
 			! isEqual( this.props.data, nextProps.data ) ||
+			this.props.height !== nextProps.height ||
 			this.props.type !== nextProps.type ||
 			this.props.width !== nextProps.width
 		);
@@ -53,7 +53,7 @@ export default class D3Base extends Component {
 	}
 
 	componentWillUnmount() {
-		window.removeEventListener( 'resize', this.getUpdatedParams );
+		window.removeEventListener( 'resize', this.drawChart );
 
 		this.deleteChart();
 	}
@@ -69,15 +69,12 @@ export default class D3Base extends Component {
 	 */
 	drawUpdatedChart() {
 		const { drawChart } = this.props;
-		const params = this.getUpdatedParams();
-
-		const svg = this.getContainer( params );
-		drawChart( svg, params, this.props.width );
+		const svg = this.getContainer();
+		drawChart( svg );
 	}
 
-	getContainer( params ) {
-		const { className } = this.props;
-		const { height, width } = params;
+	getContainer() {
+		const { className, height, width } = this.props;
 
 		this.deleteChart();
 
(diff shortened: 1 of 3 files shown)


client/components/d3chart/d3-base/index.js
```diff
@@ -26,70 +26,34 @@ import './style.scss';
...
+	shouldComponentUpdate( nextProps ) {
 		return (
-			( nextState.params !== null && ! isEqual( this.state.params, nextState.params ) ) ||
-			! isEqual( this.state.data, nextState.data ) ||
-			this.state.type !== nextState.type
+			this.props.className !== nextProps.className ||
+			! isEqual( this.props.data, nextProps.data ) ||
+			this.props.height !== nextProps.height ||
+			this.props.type !== nextProps.type ||
+			this.props.width !== nextProps.width
```
psealock: A check to see if the `drawChart` function has changed is a good idea.

client/components/d3chart/index.js
```diff
@@ -74,27 +70,27 @@ class D3Chart extends Component {
...
 	}
 
-	drawChart( node, params ) {
-		const { data, margin, type } = this.props;
+	drawChart( node ) {
+		const { data, height, margin, type, width } = this.props;
+		const params = this.getParams();
+		const adjParams = Object.assign( {}, params, {
+			height: height - margin.top - margin.bottom,
+			width: width - margin.left - margin.right,
```
psealock: Lets move this width and height calculation to `getParams` function so the removal of the margins is done in one place.

client/components/d3chart/d3-base/index.js
```diff
@@ -26,70 +26,34 @@ import './style.scss';
...
-			return { ...state, params: null };
-		}
-
-		return null;
+		this.chartRef = createRef();
 	}
 
 	componentDidMount() {
-		window.addEventListener( 'resize', this.updateParams );
+		window.addEventListener( 'resize', this.drawChart );
```
psealock: `this.drawChart` is not defined, should be `this.props.drawChart`?

On second thought. The whole point of this listener was to change the dimensions on resize, but now with those values explicitly defined in props, it appears to no longer be needed.

That responsibility has been shifted up one component:

https://github.com/woocommerce/wc-admin/blob/78fb70a233261b721599e6700dba4719bf4127ac/client/components/chart/index.js#L103

Is that right?

client/components/d3chart/d3-base/index.js
Aljullu: Done!

https://github.com/woocommerce/wc-admin/pull/961/commits/b58d725771d29ddc96b30076b70cb4adcd2b5db7

client/components/d3chart/index.js
Aljullu: Should be fixed here: https://github.com/woocommerce/wc-admin/pull/961/commits/5a447c852f1a61d15c11fd941cdc24f4225c4239.

client/components/d3chart/d3-base/index.js
Aljullu: You're absolutely right! I like that we can remove those listeners from there. :)

https://github.com/woocommerce/wc-admin/pull/961/commits/925e6ed7a03bb8a11204f3024514a63cc8473c06

Aljullu:
@psealock want to take a second look?

client/components/d3chart/index.js
```diff
@@ -130,9 +124,11 @@ class D3Chart extends Component {
 		const xScale = getXScale( uniqueDates, adjWidth );
 		const xTicks = getXTicks( uniqueDates, adjWidth, mode, interval );
 		return {
+			adjHeight,
+			adjWidth,
 			colorScheme,
 			dateSpaces: getDateSpaces( data, uniqueDates, adjWidth, xLineScale ),
-			height: calculatedHeight,
+			height,
```
psealock: this is unused.

Aljullu: Removed! :+1: 

16ec7bdf67893b849d554fd5afe4060f8bad205f

PR #964, opened 2018-11-29
Add screen reader labels to table headers

### LINKED ISSUE #859: Allow table headers to have different labels for screen readers
In several tables we have some columns were we cut the text for spacing reasons but it would make sense to read the entire text for screen readers.

For example, in the _Coupons_ report:

image

The screen reader text for that column should be _Gross discounted_.

Currently, the `Table` component doesn't allow that but it would be a good addition for accessibility.

Aljullu:
Fixes #859.

In some table headers we are displaying a short form of the label which might be problematic for screen readers (for example _G. Revenue_ instead of _Gross Revenue_. This PR allows the table component to receive a property named `screenReaderLabel` for each header, which will be used for screen readers.

- Go to one of the reports modified, for example, the _Products_ report.
- Navigate the page with the `Tab` button and the screen reader on.
- Verify it reads _Gross Revenue_ instead of _G Revenue_ when focusing that header.

### DIFF
--- a/client/analytics/report/categories/table.js
+++ b/client/analytics/report/categories/table.js
@@ -38,6 +38,7 @@ export default class CategoriesReportTable extends Component {
 			},
 			{
 				label: __( 'G. Revenue', 'wc-admin' ),
+				screenReaderLabel: __( 'Gross Revenue', 'wc-admin' ),
 				key: 'gross_revenue',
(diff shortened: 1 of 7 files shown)


timmyc (approved):
This is testing out well for me, nice one!

PR #967, opened 2018-11-30
Bump @woocommerce/components to version 1.2.0

ryelle:
This PR updates the changelog with all changes since v1.1.0. I've already run `publish`, so Lerna has updated the package.json version. Finally this includes some documentation around publishing packages, since I had the developer edit the version in package.json, but that caused a conflict in publishing, so it should only be touched by Lerna.

**To test**

- Make sure my documentation makes sense 👍


timmyc (approved):
Makes sense to me, thanks @ryelle

coderkevin (approved):
Looks good to me too.

PR #974, opened 2018-12-03
[REST API] Stock endpoint

### LINKED ISSUE #919: REST API: Create New Stock Report Endpoint
**Design Post**: p6riRB-3dV-p2

Per the designs, the stock report should be filterable by "Out of Stock", "Low Stock", and "In Stock". The current [products endpoint](http://woocommerce.github.io/woocommerce-rest-api-docs/#list-all-products) supports a `stock_status` query arg, that accepts the values of `instock`, `outofstock` and `backorder`.

For this issue I'm hoping we can explore adding in support to `/wp-json/wc/v3/products`'s `instock` query arg so you can also filter by "low stock" products.

claudiosanches:
This PR introduces a new endpoint to check product stock, listing products and variations at the same time.

It's possible control the type of report using the `type` filters with any of the follow values: `all`, `lowstock`, `outofstock` or `onbackorder`.

Example:

```
GET /wp-json/wc/v3/reports/stock?type=lowstock
```

Fixes #919

### DIFF
--- a/includes/api/class-wc-admin-rest-reports-stock-controller.php
+++ b/includes/api/class-wc-admin-rest-reports-stock-controller.php
@@ -166,6 +166,7 @@ class WC_Admin_REST_Reports_Stock_Controller extends WC_REST_Reports_Controller 
 		$data = array(
 			'id'             => $product->get_id(),
 			'parent_id'      => $product->get_parent_id(),
+			'name'           => $product->get_name(),
 			'sku'            => $product->get_sku(),
 			'stock_status'   => $product->get_stock_status(),
 			'stock_quantity' => (float) $product->get_stock_quantity(),
@@ -227,6 +228,12 @@ class WC_Admin_REST_Reports_Stock_Controller extends WC_REST_Reports_Controller 
 				'parent_id'      => array(
 					'description' => __( 'Product parent ID.', 'wc-admin' ),
 					'type'        => 'integer',
+					'context'     => array( 'view', 'edit' ),
+					'readonly'    => true,
+				),
+				'name'           => array(
+					'description' => __( 'Product name.', 'wc-admin' ),
+					'type'        => 'string',
 					'context'     => array( 'view', 'edit' ),
 					'readonly'    => true,
 				),
--- a/includes/api/class-wc-admin-rest-reports-stock-controller.php
+++ b/includes/api/class-wc-admin-rest-reports-stock-controller.php
@@ -54,12 +54,12 @@ class WC_Admin_REST_Reports_Stock_Controller extends WC_REST_Reports_Controller 
 		} elseif ( 'stock_quantity' === $args['orderby'] ) {
 			$args['meta_key'] = '_stock'; // WPCS: slow query ok.
 			$args['orderby']  = 'meta_value_num';
-		} elseif ( 'include' === $query_args['orderby'] ) {
-			$query_args['orderby'] = 'post__in';
-		} elseif ( 'id' === $query_args['orderby'] ) {
-			$query_args['orderby'] = 'ID'; // ID must be capitalized.
-		} elseif ( 'slug' === $query_args['orderby'] ) {
-			$query_args['orderby'] = 'name';
+		} elseif ( 'include' === $args['orderby'] ) {
+			$args['orderby'] = 'post__in';
+		} elseif ( 'id' === $args['orderby'] ) {
+			$args['orderby'] = 'ID'; // ID must be capitalized.
+		} elseif ( 'slug' === $args['orderby'] ) {
+			$args['orderby'] = 'name';
 		}
 
 		$args['post_type'] = array( 'product', 'product_variation' );
--- a/includes/api/class-wc-admin-rest-reports-stock-controller.php
+++ b/includes/api/class-wc-admin-rest-reports-stock-controller.php
@@ -1,0 +1,354 @@
+<?php
+/**
+ * REST API Reports stock controller
+ *
+ * Handles requests to the /reports/stock endpoint.
+ *
+ * @package WooCommerce Admin/API
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * REST API Reports stock controller class.
+ *
+ * @package WooCommerce/API
+ * @extends WC_REST_Reports_Controller
+ */
+class WC_Admin_REST_Reports_Stock_Controller extends WC_REST_Reports_Controller {
+
+	/**
+	 * Endpoint namespace.
+	 *
+	 * @var string
+	 */
+	protected $namespace = 'wc/v3';
+
+	/**
+	 * Route base.
+	 *
+	 * @var string
+	 */
+	protected $rest_base = 'reports/stock';
+
+	/**
+	 * Maps query arguments from the REST request.
+	 *
+	 * @param  array $request Request array.
+	 * @return array
+	 */
+	protected function prepare_reports_query( $request ) {
+		$args                        = array();
+		$args['offset']              = $request['offset'];
+		$args['order']               = $request['order'];
+		$args['orderby']             = $request['orderby'];
+		$args['paged']               = $request['page'];
+		$args['post__in']            = $request['include'];
+		$args['post__not_in']        = $request['exclude'];
+		$args['posts_per_page']      = $request['per_page'];
+		$args['post_parent__in']     = $request['parent'];
+		$args['post_parent__not_in'] = $request['parent_exclude'];
+
+		if ( 'date' === $args['orderby'] ) {
+			$args['orderby'] = 'date ID';
+		}
+
+		$args['post_type'] = array( 'product', 'product_variation' );
+
+		if ( 'lowstock' === $request['type'] ) {
+			$low_stock = absint( max( get_option( 'woocommerce_notify_low_stock_amount' ), 1 ) );
+			$no_stock  = absint( max( get_option( 'woocommerce_notify_no_stock_amount' ), 0 ) );
+
+			$args['meta_query'] = array( // WPCS: slow query ok.
+				array(
+					'key'   => '_manage_stock',
+					'value' => 'yes',
+				),
+				array(
+					'key'     => '_stock',
+					'value'   => array( $no_stock, $low_stock ),
+					'compare' => 'BETWEEN',
+					'type'    => 'NUMERIC',
+				),
+				array(
+					'key'   => '_stock_status',
+					'value' => 'instock',
+				),
+			);
+		} elseif ( in_array( $request['type'], array_keys( wc_get_product_stock_status_options() ), true ) ) {
+			$args['meta_query'] = array( // WPCS: slow query ok.
+				array(
+					'key'   => '_stock_status',
+					'value' => $request['type'],
+				),
+			);
+		}
+
+		return $args;
+	}
+
+	/**
+	 * Query products.
+	 *
+	 * @param  array $query_args Query args.
+	 * @return array
+	 */
+	protected function get_products( $query_args ) {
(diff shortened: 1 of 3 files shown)


Aljullu:
Thanks for working on this @claudiosanches, it's testing well! I'm just missing being able to sort by `stock_status` and `stock_quantity`, would that be possible?

claudiosanches:
@Aljullu about `stock_status` you can filter using `type`, and `stock_quantity` I think I missed this in the design post.

claudiosanches:
After my support rotation I'll check it.

timmyc:
@Aljullu and @claudiosanches what do you think about merging this in and coming back and adding in any more remaining items?

includes/api/class-wc-admin-rest-reports-stock-controller.php
```diff
@@ -0,0 +1,354 @@
...
+	 *
+	 * @param  WC_Product      $product  Report data.
+	 * @param  WP_REST_Request $request Request object.
+	 * @return WP_REST_Response
+	 */
+	public function prepare_item_for_response( $product, $request ) {
+		$data = array(
+			'id'             => $product->get_id(),
+			'parent_id'      => $product->get_parent_id(),
+			'sku'            => $product->get_sku(),
```
timmyc: Since we are returning `sku` in the response, would it be okay to return the product name in the default object too? That would cut down on an additional query for us. I know this is kind of a debated topic but figured since we are already adding in the sku here, seems like the product name wouldn't be much more.

includes/api/class-wc-admin-rest-reports-stock-controller.php
```diff
@@ -0,0 +1,354 @@
...
+			'type'              => 'string',
+			'default'           => 'desc',
+			'enum'              => array( 'asc', 'desc' ),
+			'validate_callback' => 'rest_validate_request_arg',
+		);
+		$params['orderby']        = array(
+			'description'       => __( 'Sort collection by object attribute.', 'wc-admin' ),
+			'type'              => 'string',
+			'default'           => 'date',
+			'enum'              => array(
```
timmyc: These look like the might have been copied from the products endpoint and won't work here?

includes/class-wc-admin-api-init.php
```diff
@@ -103,6 +104,7 @@ public function rest_api_init() {
 			'WC_Admin_REST_Reports_Taxes_Stats_Controller',
 			'WC_Admin_REST_Reports_Coupons_Controller',
 			'WC_Admin_REST_Reports_Coupons_Stats_Controller',
+			'WC_Admin_REST_Reports_Stock_Controller',
```
timmyc: If a store is set to no track inventory, should we still load this endpoint?

includes/api/class-wc-admin-rest-reports-stock-controller.php
```diff
@@ -0,0 +1,354 @@
...
+	 *
+	 * @param  WC_Product $product Object data.
+	 * @return array
+	 */
+	protected function prepare_links( $product ) {
+		$links = array(
+			'product' => array(
+				'href' => rest_url( sprintf( '/%s/products/%d', $this->namespace, $product->get_parent_id() ? $product->get_parent_id() : $product->get_id() ) ),
+			),
+		);
```
peterfabian: Just thinking out loud here, but should this maybe return `https://local.wordpress.test/wp-json/wc/v3/products/XYZ/variations/ABC` for variations?

includes/api/class-wc-admin-rest-reports-stock-controller.php
```diff
@@ -0,0 +1,354 @@
...
+			'description'       => __( 'Sort collection by object attribute.', 'wc-admin' ),
+			'type'              => 'string',
+			'default'           => 'date',
+			'enum'              => array(
+				'date',
+				'id',
+				'include',
+				'title',
+				'slug',
+			),
```
peterfabian: I think ordering by title would work only partially, as variations with more than 3 attributes have the same title as the parent product, IIRC...

includes/api/class-wc-admin-rest-reports-stock-controller.php
```diff
@@ -0,0 +1,354 @@
...
+	 * @param  WP_REST_Request $request Request object.
+	 * @return WP_REST_Response
+	 */
+	public function prepare_item_for_response( $product, $request ) {
+		$data = array(
+			'id'             => $product->get_id(),
+			'parent_id'      => $product->get_parent_id(),
+			'sku'            => $product->get_sku(),
+			'stock_status'   => $product->get_stock_status(),
+			'stock_quantity' => (float) $product->get_stock_quantity(),
```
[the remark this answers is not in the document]
claudiosanches: Some plugins may filter inventory and change it, that's why we never can force into `integer`.

includes/api/class-wc-admin-rest-reports-stock-controller.php
claudiosanches: I see no problem to include more fields here, for now I just put those in the design post. I'll include product's name.

includes/api/class-wc-admin-rest-reports-stock-controller.php
claudiosanches: This works here without any trouble, still valid to query products.

includes/api/class-wc-admin-rest-reports-stock-controller.php
claudiosanches: Still should group all at the same time, but not a problem, we can maybe remove at some point.
The default should per stock quantity anyway.

claudiosanches:
@Aljullu @timmyc @peterfabian fixed all issues and implement code to query by `stock_quantity`.
Now will order by stock quantity in ascending order.

includes/api/class-wc-admin-rest-reports-stock-controller.php
claudiosanches: @Aljullu For variations will already include the attributes like in the design post.

Sample result using the WooCommerce CSV (I just updated a few quantities to test):

```json
[
  {
    "id": 34,
    "parent_id": 11,
    "name": "Hoodie - Blue, Yes",
    "sku": "woo-hoodie-blue-logo",
    "stock_status": "instock",
    "stock_quantity": 2,
    "_links": {
      "product": [
        {
          "href": "https:\/\/woo.local\/wp-json\/wc\/v3\/products\/11\/variations\/34"
        }
      ],
      "parent": [
        {
          "href": "https:\/\/woo.local\/wp-json\/wc\/v3\/products\/11"
        }
      ]
    }
  },
  {
    "id": 10,
    "parent_id": 0,
    "name": "V-Neck T-Shirt",
    "sku": "woo-vneck-tee",
    "stock_status": "instock",
    "stock_quantity": 2,
    "_links": {
      "product": [
        {
          "href": "https:\/\/woo.local\/wp-json\/wc\/v3\/products\/10"
        }
      ]
    }
  }
]
```

So those names should replace this IDs on your screenshot.

Note that `Hoodie - Blue, Yes` is a variable product `Hoodie - Blue, Yes`, since we already include the attributes, not sure if we need the parent column, also grouped products does not have parents anymore.

I noticed that there is a few links there, they are for editing? If is just for editing should be fine, it's possible to generate with the ID, but if is on frontend, we probably can include the product's permalink.

Or do you think that we should include more fields there.

Aljullu: > Note that `Hoodie - Blue, Yes` is a variable product `Hoodie - Blue, Yes`, since we already include the attributes, not sure if we need the parent column, also grouped products does not have parents anymore.

@claudiosanches thanks for explaining that again! :slightly_smiling_face: 

There is still something I don't understand here. With the WooCommerce CSV products, it works as you described, but with products generated with `wc-smooth-generator`, it's not. I would say the difference is that the first ones use _[global attributes](https://docs.woocommerce.com/document/variable-product/#section-3)_ while the later ones use _custom attributes specific to product_.

As a prove of that: I have a product `perferendis unde atque qui qui` with variation `nesciunt`, `dolor` and `sed`:
image
And this is what I get from the server:
image

As I said, with the product `Hoodie` it works as you described. The only difference I could see between them is that _Hoodie_ uses global attributes while `perferendis unde atque qui qui` uses a custom attribute.

> I noticed that there is a few links there, they are for editing? If is just for editing should be fine, it's possible to generate with the ID, but if is on frontend, we probably can include the product's permalink.
> 
> Or do you think that we should include more fields there.

That's a good point. I think we are not using any of those links neither in this report nor in any other, we usually generate them in the frontend. So I would say we can remove them, if nobody has anything against that.

claudiosanches: About custom attributes this is the way it works, on our current report system should display the same way, but now there we also display the SKU first.

> That's a good point. I think we are not using any of those links neither in this report nor in any other, we usually generate them in the frontend. So I would say we can remove them, if nobody has anything against that.

If you are talking about `_links`, this is just for the REST API, just providing a route if need to do more request, just a standard.

Aljullu: > About custom attributes this is the way it works, on our current report system should display the same way, but now there we also display the SKU first.

That's unfortunate because we need some way to let the user distinguish between different variations. So is there no way to include the custom attributes in this endpoint? Otherwise we would need to make another query to populate the table.

> If you are talking about `_links`, this is just for the REST API, just providing a route if need to do more request, just a standard.

Sorry, I think I misunderstood your message before! The links we are displaying in the _Stock_ report all go to edit pages, so there is no need for a permalink.

peterfabian: Based on my investigations for https://github.com/woocommerce/woocommerce/pull/22083, I found that if there are too many attributes (too many = more than 2) or attribute name has 2+ words, the variation name (aka post title) does not include any identification for the variation, just the parent's name.
 
Don't want to go directly into solution mode, but maybe if https://github.com/woocommerce/woocommerce/pull/22083 gets accepted and merged, we could use the string from the excerpt and expose it via API to identify the variation? (and it can be truncated on the front end if too long)

claudiosanches: This is a bigger conversation, something that we should not try to solve it here now, and maybe need to be patched on master.
Also including links to the product and SKU, is not complete required to list all attributes there, at least is what we have been doing on core.

includes/api/class-wc-admin-rest-reports-stock-controller.php
```diff
@@ -0,0 +1,386 @@
...
+			'description'       => __( 'Sort collection by object attribute.', 'wc-admin' ),
+			'type'              => 'string',
+			'default'           => 'stock_quantity',
+			'enum'              => array(
+				'stock_quantity',
+				'date',
+				'id',
+				'include',
+				'title',
+				'slug',
```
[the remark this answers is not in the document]
claudiosanches: Fixed now.

includes/api/class-wc-admin-rest-reports-stock-controller.php
claudiosanches: This information on variations title is something new, so maybe we can improve, probably Mike have better ideas about it too.

timmyc:
Looks like once the CI issues are cleared up this is good to merge

timmyc:
@claudiosanches do you have time to look at fixing the tests and getting this merged in?

claudiosanches:
@timmyc all fixed, ready to review.

timmyc:
Thanks @claudiosanches! Rebased and re-ran phpunit locally and all was looking good so merged this one in to unblock the report PRs.

PR #975, opened 2018-12-03
REST API: Group extended_info for /reports/products

psealock:
Fixes https://github.com/woocommerce/wc-admin/issues/933

When passing the `extended_info` flag as a REST query parameter, the response includes additional information in the `extended_info` property instead of scattering several properties in the top level of the object.

* Change query parameter `extended_product_info` to `extended_info`.
* Return `extended_info: {}` when the query parameter is not passed.
* Make corresponding client side changes.

## Test

1. _Products Report_.
2. Ensure no errors happen when viewing the table.

### DIFF
--- a/tests/reports/class-wc-tests-reports-products.php
+++ b/tests/reports/class-wc-tests-reports-products.php
@@ -52,6 +52,7 @@ class WC_Tests_Reports_Products extends WC_Unit_Test_Case {
 					'items_sold'    => 4,
 					'gross_revenue' => 100.0, // $25 * 4.
 					'orders_count'  => 1,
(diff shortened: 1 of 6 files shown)


tests/reports/class-wc-tests-reports-products.php
```diff
@@ -167,4 +172,56 @@ public function test_order_by_product_name() {
 		$this->assertEquals( $expected_data, $query->get_data() );
 	}
 
+	/**
+	 * Test the extended ifo.
```
timmyc: info

timmyc (approved):
One small typo noted, otherwise this is looking good and testing out well for me.

PR #980, opened 2018-12-03
Order data: fix `isGetOrdersRequesting()` for the first request.

### LINKED ISSUE #911: isReportStatsRequesting returns false before the data arrives
Described by @psealock in #902:

> **Note:** It seems that the selector `isReportStatsRequesting` isn't so reliable as a detector of an in-flight request. It often returns false several moments before data actually arrives, resulting in a chart looking like this:

I also found this issue in #908.

jeffstieler:
Ensure a boolean return from `isGetOrdersRequesting()` and handle cases where requests haven’t taken off or responses haven’t been received yet.

Fixes #911 (just for Orders)

The previous logic of `isGetOrdersRequesting()` didn't properly handle the state before the initial Orders request takes flight (where `lastRequested ` and `lastReceived` are both `undefined`)

* Create some orders if you don't already have them
* Open WC admin dashboard
* Click Orders
* Verify that the placeholder order is shown during request (use dev tools to slow down your network requests if need be)
* Verify that orders display - note that only "processing" status orders are queried

### DIFF
--- a/client/wc-api/orders/selectors.js
+++ b/client/wc-api/orders/selectors.js
@@ -1,4 +1,10 @@
 /** @format */
+
+/**
+ * External dependencies
+ */
+import { isNil } from 'lodash';
+
 /**
  * Internal dependencies
  */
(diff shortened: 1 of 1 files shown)


psealock (approved):
LGTM and testing well

PR #982, opened 2018-12-03
 Use ReportTable in Revenue report

Aljullu:
Similar to #968.

This PR makes the _Revenue_ report to use the `ReportTable` component to render the table. This way, all reports will be using the same component to render the table.

### Screenshots
image

- Go to the _Revenue report_.
- Verify everything related to the table is still working fine (pagination, sorting, etc.).


timmyc (approved):
🎉 looks great and testing out well for me.

PR #988, opened 2018-12-04
Reduce order table net revenue by refund total

### LINKED ISSUE #739: Refund data not populating in revenue report
**Describe the bug**
When a refund has been issued in your selected date range, the data is not appearing in the revenue report.

**To Reproduce**

Start by refunding an order that occurred within the last week. (When I attempted to refund an order, after clicking the refund button, I got stuck on the preloaded indefinitely. Eventually I refreshed the page, and it noted the order refunded. @psealock noted the same behavior) 

1. Visit the legacy reports, and note the refund data showing up there
2. Navigate to the revenue report, and ensure your date range includes the last week.
3. Note the refunded revenue is not appearing in the summary number, or in the tabular data on the date of the refund.

**Expected behavior**
Refund data should appear in both the summary numbers and tabular data

joshuatf:
Fixes #739 

Updates the net revenue used for order table data and summary data to reflect refund amounts.

#### Before
screen shot 2018-12-04 at 12 21 17 pm

#### After
screen shot 2018-12-04 at 12 20 45 pm

1. Create an order with any amount of refunds.
2.  You may also need to manually click "Update" once more on this refund if you're processing from the dashboard.
3.  Visit the orders page `wp-admin/admin.php?page=wc-admin#/analytics/orders`.
4.  Check that the correct amount is reflected in both the table and the summary beneath the table.
5.  Run phpunit tests and make sure all pass.

### DIFF
--- a/client/analytics/report/orders/utils.js
+++ b/client/analytics/report/orders/utils.js
@@ -22,7 +22,9 @@ export function formatTableOrders( orders ) {
 			total_tax,
 			shipping_total,
 			discount_total,
+			refunds,
 		} = order;
+		const refundTotal = refunds.reduce( ( acc, refund ) => acc - parseInt( refund.total ), 0 );
 
 		return {
 			date: date_created,
@@ -33,7 +35,9 @@ export function formatTableOrders( orders ) {
 			items_sold: line_items.reduce( ( acc, item ) => item.quantity + acc, 0 ),
 			coupon_lines,
 			currency,
-			net_revenue: getCurrencyFormatDecimal( total - total_tax - shipping_total - discount_total ),
+			net_revenue: getCurrencyFormatDecimal(
+				total - total_tax - shipping_total - discount_total - refundTotal
+			),
 		};
 	} );
 }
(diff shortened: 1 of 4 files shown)


client/analytics/report/orders/utils.js
```diff
@@ -22,7 +22,9 @@ export function formatTableOrders( orders ) {
 			total_tax,
 			shipping_total,
 			discount_total,
+			refunds,
 		} = order;
+		const refundTotal = refunds.reduce( ( acc, refund ) => acc - parseInt( refund.total ), 0 );
```
greenafrican: `parseInt` rounds a decimal down here. Is that intentional? If we're displaying decimals I'd expect numbers not be rounded down (or up)

greenafrican (approved):
One small comment, otherwise this looks good.

client/analytics/report/orders/utils.js
joshuatf: Great catch @greenafrican - definitely did not want to do that.  While fixing this I stumbled across `getOrderRefundTotal()` which already does this the correct way with `parseFloat()` so I dropped that in 👍

PR #990, opened 2018-12-04
Trigger sync_order on order stats table when refunding

### LINKED ISSUE #989: Refund does not trigger update for order stats table
**Describe the bug**
Clicking "Refund" from the dashboard does not update the order stats table with the refund amount.  The correct amount is updated *only* after clicking "Update" since this triggers the hook on `save_post`.

**To Reproduce**
1.  Go to any order in the dashboard.
2.  Make a refund on that order.
3.  Note that the `wc_order_stats` table `refund_total` column remains unchanged.
4.  Click "Update" on that same order and note that the `refund_total` is now the correct amount.

**Expected behavior**
The refund total should be added to the order stats when the refund is processed.

joshuatf:
Fixes #989 

Automatically updates the `wc_order_stats` table when a refund occurs since this bypasses the original `save_post` hook.

### Screenshots
screen shot 2018-12-04 at 12 34 45 pm

1. Visit any order in the dashboard.
2. Refund any amount in that order.
3. Check that the `refund_total` column is updated to reflect the refunded amount.


emdashcodes (approved):
This is testing nicely for me. It just needs a rebase, so pre-approving. A similar line was added for syncing things on `clean_post_cache` -- but that doesn't seem to handle refunds so it looks like your change is still needed.

PR #994, opened 2018-12-04
Add table summary to Revenue report

Aljullu:
Fixes #985.

### Screenshots
image

### Detailed test instructions:
- Go to the _Revenue_ report.
- Verify a summary appears below the table with correct values.

PR #995, opened 2018-12-04
Fix 'categories' value in Categories report table summary

Aljullu:
Follow-up of #994.

Fixes the _Categories_ report table summary count number being undefined.

### Screenshots
_Before:_
image

_After:_
image

- Verify a number appears before _categories_. Currently it's going to be _0_ because it relies in the `X-WP-Total` header, which is missing in SwaggerHub endpoints, but should work when we use our own endpoints.
- Go to the _Revenue_ report.
- Verify the number before _days_ in the table summary still works.


timmyc (approved):
Looks good and testing well for me.

PR #996, opened 2018-12-04
Categories table: Hook up to endpoint

psealock:
Fixes https://github.com/woocommerce/wc-admin/issues/900

Collect Category data for the table from `/reports/categories`. In order to supply the category name without making an extra request, I introduced the `extended_info` query parameter similar to other report endpoints.

## Question

The design (p6riRB-3dR-p2) calls for links in the table to Products and Orders Reports which require filters that don't exist:

>Number of products in the category (clicks to product report filtered by category)

Product reports are not filterable by a single category, only by comparison of multiple categories. I could add a "Single Category" filter similar to "Single Product".

>Number of orders that included a product in the category (links to the order report filtered by orders that include the products in the category)

There is no Advanced Filter for categories, but it could be added with some API work such that `category_includes` and `category_excludes` are accepted parameters.

thoughts, @LevinMedia ?

## Test

1. _Categories Report_
2. See the table populate with real data.

### DIFF
--- a/client/store/reports/items/resolvers.js
+++ b/client/store/reports/items/resolvers.js
@@ -20,7 +20,7 @@ export default {
 	async getReportItems( ...args ) {
 		const [ endpoint, query ] = args.slice( -2 );
 
-		const swaggerEndpoints = [ 'categories', 'coupons' ];
+		const swaggerEndpoints = [ 'coupons' ];
 		if ( swaggerEndpoints.indexOf( endpoint ) >= 0 ) {
 			try {
 				const response = await fetch(
--- a/includes/api/class-wc-admin-rest-reports-categories-controller.php
+++ b/includes/api/class-wc-admin-rest-reports-categories-controller.php
@@ -38,15 +38,16 @@ class WC_Admin_REST_Reports_Categories_Controller extends WC_REST_Reports_Contro
 	 * @return array
 	 */
 	protected function prepare_reports_query( $request ) {
-		$args               = array();
-		$args['before']     = $request['before'];
-		$args['after']      = $request['after'];
-		$args['interval']   = $request['interval'];
-		$args['page']       = $request['page'];
-		$args['per_page']   = $request['per_page'];
-		$args['orderby']    = $request['orderby'];
-		$args['order']      = $request['order'];
-		$args['categories'] = (array) $request['categories'];
+		$args                  = array();
+		$args['before']        = $request['before'];
+		$args['after']         = $request['after'];
+		$args['interval']      = $request['interval'];
+		$args['page']          = $request['page'];
+		$args['per_page']      = $request['per_page'];
+		$args['orderby']       = $request['orderby'];
+		$args['order']         = $request['order'];
+		$args['extended_info'] = $request['extended_info'];
+		$args['categories']    = (array) $request['categories'];
 		$args['status_is']     = (array) $request['status_is'];
 		$args['status_is_not'] = (array) $request['status_is_not'];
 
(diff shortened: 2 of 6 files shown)


LevinMedia:
Hey @psealock that's a good catch, and an oversight in the design. For now, I think it's okay to set up this table without the links to products and orders, but lets add a ticket to the backlog to address this after we get through some of the other API stuff that's currently in flight. 

cc @timmyc @peterfabian

client/store/reports/items/resolvers.js
```diff
@@ -20,7 +20,7 @@ export default {
 	async getReportItems( ...args ) {
 		const [ endpoint, query ] = args.slice( -2 );
 
-		const swaggerEndpoints = [ 'categories', 'coupons', 'taxes' ];
```
timmyc: 🎉 yasss!

includes/api/class-wc-admin-rest-reports-categories-controller.php
```diff
@@ -190,6 +191,14 @@ public function get_item_schema() {
 					'context'     => array( 'view', 'edit' ),
 					'readonly'    => true,
 				),
+				'extended_info' => array(
+					'name'       => array(
+						'type'        => 'string',
+						'readonly'    => true,
+						'context'     => array( 'view', 'edit' ),
+						'description' => __( 'Product name.', 'wc-admin' ),
```
timmyc: Should this be Category name?

includes/data-stores/class-wc-admin-reports-categories-data-store.php
```diff
@@ -148,6 +148,22 @@ protected function page_records( $data, $page_no, $items_per_page ) {
...
+	 * Enriches the category data.
+	 *
+	 * @param array $categories_data Categories data.
+	 * @param array $query_args  Query parameters.
+	 */
+	protected function include_extended_info( &$categories_data, $query_args ) {
+		foreach ( $categories_data as $key => $category_data ) {
+			$extended_info = new ArrayObject();
+			if ( $query_args['extended_info'] ) {
+				$extended_info['name'] = get_the_category_by_ID( $category_data['category_id'] );
```
timmyc: I _think_ this is okay from a performance perspective but would want to get @peterfabian / @claudiosanches input on it. First inclination for me was that we should grab a mapping of all category ID=> names and reference that mapping while looping `$categories_data`

psealock: A mapping would be useful an API driven approach for nested categories (https://github.com/woocommerce/wc-admin/issues/1045), but we'd need to do that for the Products tabular data as well.

`get_the_category_by_ID` uses [get_term](https://core.trac.wordpress.org/browser/tags/5.0/src/wp-includes/taxonomy.php#L754), but I'm not sure either of the performance implications.

psealock:
>lets add a ticket to the backlog to address this after we get through some of the other API stuff that's currently in flight.

Cool, #1046 and #1047

timmyc:
@claudiosanches or @peterfabian could you give this one a review?

includes/data-stores/class-wc-admin-reports-categories-data-store.php
psealock: ping @peterfabian @claudiosanches

timmyc:
Also tagging @joshuatf for one more set of 👀 on this one.

psealock:
Thanks for the review, suggestions and new issues @joshuatf !

Issue #844, opened 2018-11-14
REST API: Implement `reports/taxes` and `reports/taxes/stats` endpoints


timmyc:
Currently the REST API reporting endpoints for taxes are using `WC_Reports_Orders_Stats_Query` with a `todo` to swap out the correct class name. In order to fully implement the reports, we both of the taxes endpoints fully developed.

Planning Post on Proton: p7bje6-15p-p2

Aljullu:
According to the design post (p6riRB-3dT-p2), the swaggerhub examples should be updated:
- `reports/taxes` should include the the _rate_ for each tax code.
- `reports/taxes/stats` should include the _total number of tax codes_.

peterfabian:
Hi @Aljullu, the rate can be fetched from existing endpoint, as described here: p7bje6-15p-p2. If we decide to include it in `/reports/taxes`, then we can update.
Should the total number of tax codes include all tax codes or only those used in the selected time period/subset of orders?

joshuatf:
@peterfabian Is it possible to add in a query param for `tax_rate_id` to both the `/reports/taxes/stats` and `/reports/taxes` routes?  We'll probably need that for filtering unless we opt to do that client-side.

joshuatf:
Closed via #1019 and #1000.

PR #999, opened 2018-12-05
Add tax report comparison filters

### LINKED ISSUE #924: Analytics: Tax Reports Filters
**Design Post**: p6riRB-3dT-p2
**Mock API**: [https://href.li/?https://app.swaggerhub.com/apis/peterfabian/wc-v3-api/1.0.0#/default/get_reports_taxes_stats](get_reports_taxes_stats)
**REST API Post**: p7bje6-15p-p2

The tax report needs to support the ability to filter by Tax Codes. Akin to being able to filter the orders report by one or many products - a similar advanced filter needs to be created and implemented on the Tax Report.

joshuatf:
Fixes #924 

Adds in the comparison filter for taxes per the designs p6riRB-3dT-p2

### Screenshots
screen shot 2018-12-05 at 2 37 05 pm

1.  Make sure you've added at least 2 tax rates under `wp-admin/admin.php?page=wc-settings&tab=tax&section=standard`
2.  Open `wp-admin/admin.php?page=wc-admin#/analytics/taxes`
3.  Change the filter method to "Comparison" under "Show:"
4.  Select a couple tax rates based on the the tax code (follows the WC format of {country}-{state}-{name}).
5.  Click "Compare."
6.  Note the change in filter params in the URL.

The query is passed to the report table, but will not affect data until #844 is implemented.

### DIFF
--- a/client/analytics/report/taxes/utils.js
+++ b/client/analytics/report/taxes/utils.js
@@ -1,0 +1,8 @@
+/** @format */
+/**
+ * External dependencies
+ */
+
+export function getTaxCode( tax ) {
+	return [ tax.country, tax.state, tax.name ].join( '-' );
+}
--- a/client/analytics/report/taxes/utils.js
+++ b/client/analytics/report/taxes/utils.js
@@ -2,7 +2,16 @@
 /**
  * External dependencies
  */
+import { __ } from '@wordpress/i18n';
 
 export function getTaxCode( tax ) {
-	return [ tax.country, tax.state, tax.name ].join( '-' );
+	return [ tax.country, tax.state, tax.name || __( 'TAX', 'wc-admin' ), tax.priority ]
+		.map( item =>
+			item
+				.toString()
+				.toUpperCase()
+				.trim()
+		)
+		.filter( Boolean )
+		.join( '-' );
 }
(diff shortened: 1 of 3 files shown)


client/analytics/report/taxes/utils.js
```diff
@@ -0,0 +1,8 @@
+/** @format */
+/**
+ * External dependencies
+ */
+
+export function getTaxCode( tax ) {
+	return [ tax.country, tax.state, tax.name ].join( '-' );
```
Aljullu: When `tax.country` or `tax.state` are blank this shows several dashes in a row:

image

How does it sound filtering the strings before joining them?
```suggestion
	return [ tax.country, tax.state, tax.name ].filter( Boolean ).join( '-' );
```

Aljullu (commented):
This is testing well and code looks good. Just a couple of comments.

I'm not sure if it's related to this PR or we should open a separate issue, but I noticed apostrophes appear with a backslash before them:
image

client/analytics/report/taxes/utils.js
joshuatf: Thanks!  Seems like a smart move.  Also trimmed here and modified to match the tax codes in WC, adding 'TAX' as the name if none is defined and appending the priority to the end of the tax code.

joshuatf:
Thanks for reviewing @Aljullu !  Made the changes to the tax code mentioned.

As for the escaping issue, opened an issue in WC as this appears to be an existing issue on current tax reports as well.  https://github.com/woocommerce/woocommerce/issues/22149

Aljullu (approved):
It works as expected and code looks good to me! :rocket:

PR #1000, opened 2018-12-05
Add taxes endpoint and data store

joshuatf:
Fixes (partially) #844 

Adds in the endpoint for the taxes report table.

### Screenshots
screen shot 2018-12-07 at 8 56 45 am

1.  Add at least a couple of tax rates under `/wp-admin/admin.php?page=wc-settings&tab=tax&section=standard`
2.  Add several rows of data into the `wp_wc_order_tax_lookup` table.  This can be any data, but match the tax rate ID to an existing tax rate from step 1 for more thorough testing.
3.  Make a post request to `wp/wp-json/wc/v3/reports/taxes` with WC Rest API keys.
4.  Make sure you receive accurate data from the request for these fields:
    * tax_rate_id
    * name
    * country
    * state
    * priority
    * total_tax
    * order_tax
    * shipping_tax
    * orders_count
    * _links
5.  Test with these query params as well
    * before
    * after
    * page
    * per_page
    * orderby
    * order
    * taxes (comma separated tax_rate_ids from the `woocommerce_tax_rates` table
6. Run phpunit tests to make sure they pass.

### Todo
This PR is already very large, so it's being broken down into smaller pieces.
* Tax report hookup
* Tax report stats
* Tax report stats hookup
* `wp_wc_order_tax_lookup` syncing

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-taxes-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-taxes-data-store.php
@@ -80,6 +80,27 @@ class WC_Admin_Reports_Taxes_Data_Store extends WC_Admin_Reports_Data_Store impl
 	}
 
 	/**
+	 * Fills ORDER BY clause of SQL request based on user supplied parameters.
+	 *
+	 * @param array $query_args Parameters supplied by the user.
+	 * @return array
+	 */
+	protected function get_order_by_sql_params( $query_args ) {
+		$sql_query['order_by_clause'] = '';
+		if ( isset( $query_args['orderby'] ) ) {
+			$sql_query['order_by_clause'] = $this->normalize_order_by( $query_args['orderby'] );
+		}
+
+		if ( isset( $query_args['order'] ) ) {
+			$sql_query['order_by_clause'] .= ' ' . $query_args['order'];
+		} else {
+			$sql_query['order_by_clause'] .= ' DESC';
+		}
+
+		return $sql_query;
+	}
+
+	/**
 	 * Returns the report data based on parameters supplied by the user.
 	 *
 	 * @param array $query_args  Query parameters.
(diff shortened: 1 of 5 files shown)


includes/data-stores/class-wc-admin-reports-taxes-data-store.php
```diff
@@ -0,0 +1,236 @@
...
+
+			$tax_data = $wpdb->get_results(
+				"SELECT
+						{$selections}
+					FROM
+						{$table_name}
+						{$sql_query_params['from_clause']}
+					WHERE
+						1=1
+						{$sql_query_params['where_clause']}
```
jeffstieler: Looks like this is missing `where_time_clause`. (Found testing `before`/`after` query params).

includes/data-stores/class-wc-admin-reports-taxes-data-store.php
```diff
@@ -0,0 +1,236 @@
...
+		'orders_count' => 'COUNT(DISTINCT order_id) as orders_count',
+	);
+
+	/**
+	 * Constructor
+	 */
+	public function __construct() {
+		global $wpdb;
+		$table_name                          = $wpdb->prefix . self::TABLE_NAME;
+		$this->report_columns['tax_rate_id'] = $table_name . '.' . $this->report_columns['tax_rate_id'];
```
jeffstieler: Curious - why is this needed?

psealock (commented):
I just came across a bug which may be outside the scope of this PR but renders the `total_tax` and `order_tax` portion of the response incorrectly.

I created two orders that would involve a tax rate, but it appears the `wp_wc_order_tax_lookup` table was populated by many duplicate rows:

screen shot 2018-12-10 at 1 10 22 pm

When I hit the endpoint, the response sums up the duplications

screen shot 2018-12-10 at 1 10 34 pm

It should be

```
"total_tax": 91.6,
"order_tax": 91.6,
```

joshuatf:
@psealock I also had issues with the `wc_order_tax_lookup` table not populating correctly, but in my case no information gets added to the table at all.

Opened a separate issue here - https://github.com/woocommerce/wc-admin/issues/1009

includes/data-stores/class-wc-admin-reports-taxes-data-store.php
joshuatf: Nice, catch @jeffikus !  Added this and it's testing well for me.  Can you give it another shot?

jeffstieler (approved):

PR #1006, opened 2018-12-05
Use `wc-api` for Reports.

### LINKED ISSUE #856: Analytics: Revenue: Last Year: Uncaught (in promise) TypeError: Cannot read property 'data' of null
**Describe the bug**

When paging is required to concatenate pages of intervals (e.g. when per_page is 100 and interval is day and totalResults is 365), _data is sometimes null here in client/store/reports/utils.js, causing the error (since _data is null, it cannot have a data property):

```
forEach( pagedData, function( _data ) {
			intervals = intervals.concat( _data.data.intervals );
		} );
```

**To Reproduce**

1. Go to wp-admin > Analytics > Revenue
2. Set the Date Range to Last Year
3. See error in console

**Expected behavior**

Concatenation without error

**Screenshots**

screen shot 2018-11-14 at 6 53 47 pm

WP 4.9.8
Gutenberg 4.3.0

cc @LevinMedia  - this is what we were debugging with you Wednesday night

jeffstieler:
Fixes #856.

Introduces `wc-api`/`fresh-data` for Reports in order to solve the `TypeError` issue when fetching multiple pages of report data.

* Go to wp-admin > Analytics > Revenue
* Set the Date Range to Last Year
* Verify no console error (cannot read property 'data' of null)
* Visit all other report types and verify function

### DIFF
--- a/client/wc-api/reports/items/operations.js
+++ b/client/wc-api/reports/items/operations.js
@@ -26,6 +26,7 @@ const typeEndpointMap = {
 	'report-items-query-categories': 'categories',
 	'report-items-query-coupons': 'coupons',
 	'report-items-query-taxes': 'taxes',
+	'report-items-query-variations': 'variations',
 };
 
 function read( resourceNames, fetch = apiFetch ) {
--- a/client/wc-api/reports/items/operations.js
+++ b/client/wc-api/reports/items/operations.js
@@ -53,7 +53,6 @@ function read( resourceNames, fetch = apiFetch ) {
 		try {
 			const response = await fetch( fetchArgs );
 			const report = await response.json();
-			// TODO: exclude these if using swagger?
 			const totalResults = parseInt( response.headers.get( 'x-wp-total' ) );
 			const totalPages = parseInt( response.headers.get( 'x-wp-totalpages' ) );
 
--- a/client/wc-api/reports/items/operations.js
+++ b/client/wc-api/reports/items/operations.js
@@ -39,18 +39,18 @@ function read( resourceNames, fetch = apiFetch ) {
 		const endpoint = typeEndpointMap[ prefix ];
 		const query = getResourceIdentifier( resourceName );
 
-		let apiPath = NAMESPACE + '/reports/' + endpoint + stringifyQuery( query );
+		const fetchArgs = {
+			parse: false,
+		};
 
 		if ( swaggerEndpoints.indexOf( endpoint ) >= 0 ) {
-			apiPath = SWAGGERNAMESPACE + 'reports/' + endpoint + stringifyQuery( query );
+			fetchArgs.url = SWAGGERNAMESPACE + 'reports/' + endpoint + stringifyQuery( query );
+		} else {
+			fetchArgs.path = NAMESPACE + '/reports/' + endpoint + stringifyQuery( query );
 		}
 
 		try {
-			const response = await fetch( {
-				parse: false,
-				path: apiPath,
-			} );
-
+			const response = await fetch( fetchArgs );
 			const report = await response.json();
 			// TODO: exclude these if using swagger?
 			const totalResults = parseInt( response.headers.get( 'x-wp-total' ) );
--- a/client/wc-api/reports/items/operations.js
+++ b/client/wc-api/reports/items/operations.js
@@ -1,0 +1,74 @@
+/** @format */
+/**
+ * External dependencies
+ */
+import apiFetch from '@wordpress/api-fetch';
+
+/**
+ * WooCommerce dependencies
+ */
+import { stringifyQuery } from '@woocommerce/navigation';
+
+/**
+ * Internal dependencies
+ */
+import { getResourceIdentifier, getResourcePrefix } from '../../utils';
+import { NAMESPACE } from '../../constants';
+import { SWAGGERNAMESPACE } from 'store/constants';
+
+// TODO: Remove once swagger endpoints are phased out.
+const swaggerEndpoints = [ 'categories', 'coupons', 'taxes' ];
+
+const typeEndpointMap = {
+	'report-items-query-orders': 'orders',
+	'report-items-query-revenue': 'revenue',
+	'report-items-query-products': 'products',
+	'report-items-query-categories': 'categories',
+	'report-items-query-coupons': 'coupons',
+	'report-items-query-taxes': 'taxes',
+};
+
+function read( resourceNames, fetch = apiFetch ) {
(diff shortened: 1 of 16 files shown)


client/wc-api/reports/items/operations.js
```diff
@@ -0,0 +1,74 @@
...
+ * Internal dependencies
+ */
+import { getResourceIdentifier, getResourcePrefix } from '../../utils';
+import { NAMESPACE } from '../../constants';
+import { SWAGGERNAMESPACE } from 'store/constants';
+
+// TODO: Remove once swagger endpoints are phased out.
+const swaggerEndpoints = [ 'categories', 'coupons', 'taxes' ];
+
+const typeEndpointMap = {
```
Aljullu: Given that we are using this for items and stats, could it be extracted to a `/utils.js` file so we avoid repeating it?

client/wc-api/reports/items/selectors.js
```diff
@@ -0,0 +1,45 @@
...
+
+const getReportItems = ( getResource, requireResource ) => (
+	type,
+	query = {},
+	requirement = DEFAULT_REQUIREMENT
+) => {
+	const resourceName = getResourceName( `report-items-query-${ type }`, query );
+	const data = requireResource( requirement, resourceName ) || {};
+
+	return data;
```
Aljullu: Small suggestion: we can get rid of the `data` assignment and return its value directly.
```ES6
	return requireResource( requirement, resourceName ) || {};
```

client/wc-api/reports/items/selectors.js
```diff
@@ -0,0 +1,45 @@
...
+};
+
+const isGetReportItemsRequesting = getResource => ( type, query = {} ) => {
```
Aljullu: I noticed we use `isGetReportItemsRequesting` here while in `client/wc-api/reports/stats/selectors.js` we use `isReportStatsRequesting`, notice one includes `Get` in the name while the other doesn't. I think it would probably be a good idea to unify that and either remove `Get` from `isGetReportItemsRequesting` or add it to `isReportStatsRequesting`.

client/wc-api/reports/stats/selectors.js
```diff
@@ -0,0 +1,45 @@
...
+
+const getReportStats = ( getResource, requireResource ) => (
+	type,
+	query = {},
+	requirement = DEFAULT_REQUIREMENT
+) => {
+	const resourceName = getResourceName( `report-stats-query-${ type }`, query );
+	const data = requireResource( requirement, resourceName ) || {};
+
+	return data;
```
Aljullu: Same as before. :slightly_smiling_face:

Aljullu (commented):
This look really good, I'm very happy to see this going forward! :heart_eyes:

I noticed the variations table doesn't show any data and the placeholder stays forever:

image

Same for the _Coupons_ table. Can you reproduce that as well?

client/wc-api/reports/items/selectors.js
jeffstieler: I'll remove "get" from the error and requesting selectors. In the future this will all be DRYed up. 👍

client/wc-api/reports/items/operations.js
Aljullu: Oh right, didn't notice that. :sweat_smile:

client/wc-api/reports/stats/selectors.js
jeffstieler: Done in 7f934aa.

client/wc-api/reports/items/selectors.js
jeffstieler: Done in 7f934aa.

client/wc-api/reports/items/selectors.js
jeffstieler: Done in 7f934aa.

jeffstieler:
> I noticed the variations table doesn't show any data and the placeholder stays forever

Hmm, I'm seeing that as well. It looks like the variations API request is never made.. looking into that now.

jeffstieler:
This is ready for a retest now that https://github.com/woocommerce/wc-admin/pull/1037 landed.

client/wc-api/reports/items/operations.js
```diff
@@ -0,0 +1,75 @@
...
+		if ( swaggerEndpoints.indexOf( endpoint ) >= 0 ) {
+			fetchArgs.url = SWAGGERNAMESPACE + 'reports/' + endpoint + stringifyQuery( query );
+		} else {
+			fetchArgs.path = NAMESPACE + '/reports/' + endpoint + stringifyQuery( query );
+		}
+
+		try {
+			const response = await fetch( fetchArgs );
+			const report = await response.json();
+			// TODO: exclude these if using swagger?
```
Aljullu: Given that it's working fine like it is now (`totalResults` and `totalPages` end up being `NaN`, which doesn't break the components), maybe we can just remove the comment?

client/wc-api/reports/stats/operations.js
```diff
@@ -0,0 +1,77 @@
...
+		} else if ( statEndpoints.indexOf( endpoint ) >= 0 ) {
+			fetchArgs.path = NAMESPACE + '/reports/' + endpoint + '/stats' + stringifyQuery( query );
+		} else {
+			fetchArgs.path = endpoint + stringifyQuery( query );
+		}
+
+		try {
+			const response = await fetch( fetchArgs );
+			const report = await response.json();
+			// TODO: exclude these if using swagger?
```
Aljullu: Same as before. :slightly_smiling_face:

Aljullu (approved):
Everything seems to work fine and code looks great!

I just added a couple of small comments, but feel free to merge as is. :+1:

client/wc-api/reports/stats/operations.js
```diff
@@ -0,0 +1,77 @@
...
+	return filteredNames.map( async resourceName => {
+		const prefix = getResourcePrefix( resourceName );
+		const endpoint = typeEndpointMap[ prefix ];
+		const query = getResourceIdentifier( resourceName );
+
+		const fetchArgs = {
+			parse: false,
+		};
+
+		if ( swaggerEndpoints.indexOf( endpoint ) >= 0 ) {
```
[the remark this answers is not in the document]
jeffstieler: This was directly copied over from the old resolver file - since it's bound to be deleted, I'll choose to leave it. 👍

LevinMedia:
So much ✨Thanks for getting this one sorted out!

PR #1008, opened 2018-12-06
Update/taxes endpoint hookup/941

### LINKED ISSUE #941: Analytics: Hook REST Data up to Taxes Report
Child task of #172, blocked by #844 

When #844 lands, this task involves hooking up the REST endpoint to the Taxes Report

joshuatf:
Fixes #941 

Depends on #1000 and #1019 - will be rebased pending those PRs.

Hooks up the tax reports to the new endpoint created in #1000 

### Screenshots
screen shot 2018-12-06 at 4 14 35 pm

1.  Add at least a couple of tax rates under /wp-admin/admin.php?page=wc-settings&tab=tax&section=standard
2.  Add several rows of data into the wp_wc_order_tax_lookup table. This can be any data, but match the tax rate ID to an existing tax rate from step 1 for more thorough testing.
3.  Visit `wp-admin/admin.php?page=wc-admin#/analytics/taxes`.
4.  Check that data for each tax rate is shown properly based on info added from step 2.
5.  Check that tax stat totals are correct.

### DIFF
--- a/client/analytics/report/taxes/table.js
+++ b/client/analytics/report/taxes/table.js
@@ -133,7 +133,7 @@ export default class TaxesReportTable extends Component {
 			},
 			{
 				label: _n( 'order', 'orders', totals.orders, 'wc-admin' ),
-				value: numberFormat( totals.orders ),
+				value: numberFormat( totals.orders_count ),
 			},
 		];
 	}
--- a/client/analytics/report/taxes/table.js
+++ b/client/analytics/report/taxes/table.js
@@ -87,7 +87,6 @@ export default class TaxesReportTable extends Component {
 					value: tax_rate_id,
 				},
 				{
-					// @TODO add `rate` once it's returned by the API
 					display: tax_rate.toFixed( 2 ) + '%',
 					value: tax_rate,
 				},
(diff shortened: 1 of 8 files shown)


packages/components/src/search/autocompleters/taxes.js
```diff
@@ -13,6 +13,7 @@ import { stringifyQuery } from '@woocommerce/navigation';
  * Internal dependencies
  */
 import { computeSuggestionMatch } from './utils';
+import { getTaxCode } from 'analytics/report/taxes/utils';
```
ryelle: Can you pull this function to a new package, or copy it into `packages/components`? This file won't exist in other apps using the components.

joshuatf: Thanks for catching this, @ryelle!  I think @timmyc copied this into packages to fix up the build - https://github.com/woocommerce/wc-admin/blob/2985730ca5a541fbf05a2deb58781b4ddc0455b8/packages/components/src/search/autocompleters/utils.js#L28

timmyc (approved):
This is testing out well for me, though might have hit a bit of a bug on the lookup table side of things as I'm seeing the following generated for one order in the table:

image

Guessing there should only be one row per order per tax rate correct? I can do a follow-up issue on that

PR #1016, opened 2018-12-06
Update/php sniffs

timmyc:
Re-do of #931


emdashcodes (approved):
Change still looks good.

Tested the same command I did on the original PR.

👍

PR #1017, opened 2018-12-06
Add/tax report filters/924

timmyc:
Re-do of #999


emdashcodes (approved):
This is testing well for me now. 👍

Issue #1025, opened 2018-12-07
Summary numbers display 'N/A' instead of '0' for the previous period


Aljullu:
**Steps to reproduce:**
- Go to the _Orders_ report and select a period so you have no orders in the previous period. For me, selecting 'Today' in the date range worked because I didn't have any order yesterday.

**Actual result:**

- The summary numbers for the previous period will display _N/A_.

![image](https://user-images.githubusercontent.com/3616980/49663638-e995d200-fa14-11e8-8104-57e70ade4cbb.png)

**Expected result:**

- It should display _0_ and _N/A_ should only be used when we have no data for that period.

timmyc:
This appears to be all fixed up after #1026

PR #1026, opened 2018-12-07
Fix ReportNumber sometimes receiving NaN values

### LINKED ISSUE #1022: Error in Orders report: Received NaN for the children attribute
When going to the _Orders_ report, this error appears in the console:
`Warning: Received NaN for the children attribute. If this is expected, cast the value to a string.`
image

Aljullu:
Fixes #1022 and #1025.

This started as what I thought would be an easy fix for #1022 but I ended up making a refactor of `<ReportSummary>` and adding tests for that component. Hopefully I didn't break anything and it keeps working as expected. :slightly_smiling_face: 

### Screenshots
_Before:_
image

_After:_
image

- Go to the _Orders_ report and verify there is no error in the console and it shows _0_ values instead of _N/A_ when actually the value is 0 instead of missing.
- Go to the other reports and verify the Summary number on top of the chart render the correct values.

### DIFF
--- a/client/analytics/components/report-summary/index.js
+++ b/client/analytics/components/report-summary/index.js
@@ -25,26 +25,27 @@ import { calculateDelta, formatValue } from './utils';
 export class ReportSummary extends Component {
 	render() {
 		const { charts, query, selectedChart, summaryData } = this.props;
+		const { totals, isError, isRequesting } = summaryData;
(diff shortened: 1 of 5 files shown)


jeffstieler (approved):
Looks good! :ship: it.

PR #1042, opened 2018-12-08
Fix wrong default data structure in 'getReportTableData'

Aljullu:
Found this while reviewing #1006.

By default, `getReportTableData` was returning `items` as an array, but it's actually an object with a property `data`, which contains the array of items.

`items` can also contain `totalCount` and `totalPages`, but I think it's ok leaving those undefined because they are integers.

- Go to any report and verify table is still working and there are no regressions.

### DIFF
--- a/client/analytics/components/report-table/index.js
+++ b/client/analytics/components/report-table/index.js
@@ -118,7 +118,7 @@ ReportTable.propTypes = {
 };
 
 ReportTable.defaultProps = {
-	tableData: null,
+	tableData: {},
 	tableQuery: {},
 };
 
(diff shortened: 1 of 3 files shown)


psealock (approved):
Nice catch on this. I was just looking into the implications and came across a default prop that should be changed in light of whats returned in `getReportTableData`.

https://github.com/woocommerce/wc-admin/blob/aaa738fc30765aa85dc625ed36e3b27f9bd88dc4/client/analytics/components/report-table/index.js#L121

`tableData` also gets destructured in the render function.

Aljullu:
> Nice catch on this. I was just looking into the implications and came across a default prop that should be changed in light of whats returned in `getReportTableData`.

Fixed in 33a81859055eed438b6399c466625e5a69187600.

PR #1050, opened 2018-12-10
Updated report regeneration to use Action Scheduler

### LINKED ISSUE #429: Lookup tables: Use Action Scheduler
Action Scheduler v2 has been released and should be a robust, well-supported background processing solution. We should integrate all of the lookup table syncing with action scheduler instead of doing it all on `save_post`, as it will give better performance. The hooks in `save_post` should just schedule an action scheduler job. We don't want placing an order to take a long time because the lookup tables are updating in the same request. 

The update routine for building the lookup tables when updated to 3.5 and the button in System Status Tools to rebuild all of the order lookup tables should also kick off action scheduler jobs. Both of those should be able to use the same function, so it should be fairly straightforward.

Examples of how to integrate with the new library are [in this PR](https://github.com/woocommerce/woocommerce/pull/20030).

### LINKED ISSUE #860: Analytics: Enhancements to Rebuild Reports Tool
Currently the process to Rebuild, or populate the reporting tables for the first time, is a manually triggered process accessed via the "Tools" tab within the Status page. When that tool is launched, the [following logic](https://github.com/woocommerce/wc-admin/blob/501244778606bc7ec18433a26128efd4481ba966/includes/api/class-wc-admin-rest-system-status-tools-controller.php#L56-L60) calls [queue_order_stats_repopulate_database](https://github.com/woocommerce/wc-admin/blob/09d15f6a264561117c89aa9343ffc11c2e0007f6/includes/data-stores/class-wc-admin-reports-orders-data-store.php#L394) - which as it states in a TODO, needs to be reworked since it currently grabs *all* orders to process at once.

So I'm creating this issue to track that TODO and a few other potential enhancements:

- [ ] Update `queue_order_stats_repopulate_database` to work in a batch mode, perhaps using 100 orders per batch?
- [ ] Add ability to re-process orders based upon a date range. Thinking this could be useful if a specific time period of orders needs to be reprocessed
- [ ] Consider adding a link to the Scheduled Actions tab ( /wp-admin/admin.php?page=wc-status&tab=action-scheduler#/ ) after a job has been queued, so users can view progress of the job
- [ ] Run performance tests of the repopulate job on some limited hosting environments.

peterfabian:
Fixes #860.
Fixes #429 

This is the first attempt to switch regenerating reports to Action Scheduler and make it scalable. 
This PR makes orders being processed in batches of 10 (default, configurable via a filter) and each batch performs 2 actions:
- plans the next batch at a future datetime (e.g. 5 seconds from now)
- plans updating of 10 orders at a future datetime (e.g. 5 seconds from now)

Next batch will plan the following one and so on, until all orders are processed.
Orders are processed from the oldest to the newest. 

Potential problems/TODOs: 
- [ ] handle refunds
- if you get orders faster than they're processed, reports will never finish being generated
- size of batch needs to be tested on different hosts and needs to differ based on each shop's specifics
- if one of the batches fail before planning the next one for some reason, the regeneration will stop--how will the user know
- if one of the orders becomes invalid, orders following it won't be processed, despite using `try` - `catch`... investigate
- lots of garbage in wp_post after this finishes: 2 posts per order batch (~10 orders, by deafult at the moment)

Question marks:
- should we `TRUNCATE` the report tables when a user clicks `Rebuild reports` button?

### Screenshots
screenshot 2018-12-10 at 11 11 10


timmyc:
> should we TRUNCATE the report tables when a user clicks Rebuild reports button?

I think the TRUNCATE should be a distinct tool, or maybe empty & rebuild. I also mentioned the notion of adding a date range to the rebuild if a store owner wants to only rebuild a set of dates that might have become corrupted for some reason.

peterfabian:
This PR has been superseded by #1291 .

PR #1052, opened 2018-12-10
Add screen reader labels to table headers which are not buttons

Aljullu:
Same as #964, but that one only affected table headers which are buttons while this PR will affect all table headers.

Testing is blocked by  #1061.
- Go to the _Customers_ report with the screen reader on.
- Navigate the page with your screen reader shortcuts/commands until you reach the _AOV_ cell.
  - Orca (GNU/Linux) commands can be found here: https://help.gnome.org/users/orca/stable/commands_structural_navigation.html.en#tables
  - I don't know the ones for Voice Over (Mac OS), maybe this page is useful: https://help.apple.com/voiceover/mac/10.14/#/vo27958
- Verify it reads _Average Order Value_ instead of _AOV_ when that header is selected.


Aljullu:
Thanks for testing! I will merge this PR for now and we can open further issues if we see there is any special issue with Chrome in OS X.

Issue #1023, opened 2018-12-07
Lodash issues on none wc-admin routes break WordPress.com notifications


justinshreve:
My previous PR #1004 re-enabled the activity panel tabs on none wc-admin routes. There was an attempt to fix some JS errors that prevent the WordPress.com notifications bar from working because of lodash version mismatches.

There still seems to be some issues with this.

With WordPress 5.0 and WooCommerce 3.6-dev I'm seeing

> notes-common-v2.js?ver=6.9-alpha-201849:25 Uncaught TypeError: _.contains is not a function

@jeffstieler with WC version is 3.6.0-dev at commit f39b006264cf9330e8e2a7f4b19d7ae0b0699281

is seeing `Uncaught TypeError: Cannot read property 'noConflict' of undefined`.

**To Reproduce**

1. Go to to any none wc-admin route (orders, etc).
2. Open console and see error.
3. Click WordPress.com notifications.
4. See more errors.

**Expected behavior**

No errors.

PR #1053, opened 2018-12-10
Remove lodash noConflict

emdashcodes:
After reading https://make.wordpress.org/core/2018/12/06/javascript-packages-and-interoperability-in-5-0-and-beyond/ and doing some more testing, it looks like we don't need the noConflict call after all, in conjunction with the webpack change I added previously.

To Test:
* Test with WordPress 5.0+ now that it is out
* Close any currently running `npm start` process and re-run `npm install` and `npm start`
* Clear cache
* Open the orders page, hopefully see no JS warnings
* Open the WordPress.com notification dropdown

See #1023.

### DIFF
--- a/lib/client-assets.php
+++ b/lib/client-assets.php
@@ -95,15 +95,6 @@ function wc_admin_register_script() {
 		wp_set_script_translations( WC_ADMIN_APP, 'wc-admin' );
 	}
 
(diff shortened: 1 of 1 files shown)


jeffstieler (approved):
Tested on WP 4.9.8 and 5.0.0. :shipit:

PR #1054, opened 2018-12-10
Customers: add one advanced filter

psealock:
Hook up one Advanced Filter so the wiring in correct for the development of Customers Report filters.
 
I added the _name_ filter with a reference to the products autocompleter. The filter with autocompleter can be fully fleshed out when dealing with https://github.com/woocommerce/wc-admin/issues/1029.

### Test

1. _Customers Report_
2. Show > Advanced Filters
3. Make sure the filter is working and updating the url

### DIFF
--- a/client/analytics/report/customers/config.js
+++ b/client/analytics/report/customers/config.js
@@ -2,9 +2,9 @@
 /**
  * External dependencies
  */
-import { __ } from '@wordpress/i18n';
-
-export const showDatePicker = false;
+import { __, _x } from '@wordpress/i18n';
+import { getRequestByIdString } from '../../../lib/async-requests';
+import { NAMESPACE } from '../../../store/constants';
 
 export const filters = [
 	{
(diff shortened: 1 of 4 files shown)


client/analytics/report/customers/config.js
```diff
@@ -2,16 +2,63 @@
 /**
  * External dependencies
  */
-import { __ } from '@wordpress/i18n';
-
-export const showDatePicker = false;
+import { __, _x } from '@wordpress/i18n';
+import { getRequestByIdString } from '../../../lib/async-requests';
```
joshuatf: Nitpick, but I think these 2 imports can be put under internal dependencies and we can drop the `../../../` for readability.

```js
/**
 * Internal dependencies
 */
import { getRequestByIdString } from 'lib/async-requests';
import { NAMESPACE } from 'store/constants';
```

joshuatf (approved):
Code looks good and tests well!  Pre-approving with tiny import change.

client/analytics/report/customers/config.js
psealock: whoops, I merged before getting to this. I'll address in the next one, thanks

PR #1056, opened 2018-12-10
Table: Add onColumnsChange

coderkevin:
This adds an `onColumnsChange` property which is called when the column
visibility is toggled on the table. It also cleans up a bit of the
showCols code and fixes an incorrect update in the component update.

Test via #1057


joshuatf (approved):
Testing well with #1057 - LGTM! :shipit:

PR #1057, opened 2018-12-10
wc-api: Add user meta for UI preferences

### LINKED ISSUE #892: Dashboard: Persist Store Performance Selections
Prior discussion in #661

Currently when one toggles the display summary numbers shown in the Store Performance section of the dashboard, the selections are lost upon navigating away from the dashboard because the selections are currently stored in component state.

For this task, we should persist which summary numbers a user has displayed as a user preference so each subsequent visit to the dashboard, the user only sees the summary numbers they previously configured to visible.

@justinshreve approached this problem in #255 which was never merged, so perhaps that could be used as a starting point for this functionality. Persisting preferences like this for `wc-admin` will also likely be used in the general display/configuration of blocks on the dashboard, and potentially config of which table columns are shown on individual reports

coderkevin:
Fixes #892 

This adds wc-api code to store UI preferences and provides one example
of updating the revenue table to save column preferences.

1. Go to Analytics -> Revenue.
2. Toggle some columns off.
3. Close browser tab.
4. Open another browser tab and go to Analytics -> Revenue (or use another browser, etc)
5. Observe that the same columns are toggled off.

For more detailed testing, open dev console network tab and look for "me?context=edit" requests and "me" POST requests.

### DIFF
--- a/lib/admin.php
+++ b/lib/admin.php
@@ -338,5 +338,73 @@ function woocommerce_embed_page_header() {
 	</div>
 	<?php
 }
-
 add_action( 'in_admin_header', 'woocommerce_embed_page_header' );
+
+/**
+ * Registers WooCommerce specific user data to the Wordpress user API.
+ */
+function wc_admin_register_user_data() {
+	register_rest_field(
+		'user',
+		'woocommerce_meta',
+		array(
+			'get_callback'    => 'wc_admin_get_user_data_values',
+			'update_callback' => 'wc_admin_update_user_data_values',
+			'schema' => null,
+		)
+	);
+}
+ add_action( 'rest_api_init', 'wc_admin_register_user_data' );
+
+ /**
+  * We store some WooCommerce specific user meta attached to users endpoint,
+  * so that we can track certain preferences or values such as the inbox activity panel last open time.
+  * Additional fields can be added in the function below, and then used via wc-admin's currentUser data.
+  *
+  * @return array Fields to expose over the WP user endpoint.
+  */
+function wc_admin_get_user_data_fields() {
+	$user_data_fields = array(
+		'revenue_report_columns',
+	);
+
+	return apply_filters( 'wc_admin_get_user_data_fields', $user_data_fields );
+}
+
+ /**
+  * For all the registered user data fields (  wc_admin_get_user_data_fields ), fetch the data
+  * for returning via the REST API.
+  *
+  * @param WP_User $user Current user.
+  */
+function wc_admin_get_user_data_values( $user ) {
+	$values = array();
+	foreach ( wc_admin_get_user_data_fields() as $field ) {
+		$values[ $field ] = get_user_meta( $user['id'], 'woocommerce_meta_' . $field, true );
+	}
+	return $values;
+}
+
+ /**
+  * For all the registered user data fields (  wc_admin_get_user_data_fields ), update the data
+  * for the REST API.
+  *
+  * @param array   $values   The new values for the meta.
+  * @param WP_User $user     The current user.
+  * @param string  $field_id The field id for the user meta.
+  */
+function wc_admin_update_user_data_values( $values, $user, $field_id ) {
+	if ( empty( $values ) || ! is_array( $values ) || 'woocommerce_meta' !== $field_id ) {
+		return;
+	}
+	$fields  = wc_admin_get_user_data_fields();
+	$updates = array();
+	foreach ( $values as $field => $value ) {
+		if ( in_array( $field, $fields ) ) {
+			$updates[ $field ] = $value;
+			update_user_meta( $user->ID, 'woocommerce_meta_' . $field, $value );
+		}
+	}
+
+	return $updates;
+}
--- a/lib/admin.php
+++ b/lib/admin.php
@@ -341,7 +341,7 @@ function woocommerce_embed_page_header() {
 add_action( 'in_admin_header', 'woocommerce_embed_page_header' );
 
 /**
- * Registers WooCommerce specific user data to the Wordpress user API.
+ * Registers WooCommerce specific user data to the WordPress user API.
  */
 function wc_admin_register_user_data() {
 	register_rest_field(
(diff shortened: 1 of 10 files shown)


lib/admin.php
```diff
@@ -338,5 +338,73 @@ function woocommerce_embed_page_header() {
...
+ /**
+  * For all the registered user data fields (  wc_admin_get_user_data_fields ), fetch the data
+  * for returning via the REST API.
+  *
+  * @param WP_User $user Current user.
+  */
+function wc_admin_get_user_data_values( $user ) {
+	$values = array();
+	foreach ( wc_admin_get_user_data_fields() as $field ) {
+		$values[ $field ] = get_user_meta( $user['id'], 'woocommerce_meta_' . $field, true );
```
joshuatf: Could you see anyone using this to get user meta not prefixed with `woocommerce_meta`?

Was thinking of using this as a 3rd party developer and it might be nice if we could add in other user meta fields:

```php
	$user_data_fields = array(
		'revenue_report_columns' => 'woocommerce_meta_revenue_report_columns',
         'my_custom_user_field' => 'other_plugin_prefix_meta_field',
	);
```

joshuatf (commented):
Code looks good and tests well across different browsers/tabs!  Nice work @coderkevin 

Couple comments about the `columnPrefsKey` prop and extensibility with the user meta fields.

lib/admin.php
```diff
@@ -338,5 +338,73 @@ function woocommerce_embed_page_header() {
...
+
+/**
+ * Registers WooCommerce specific user data to the Wordpress user API.
+ */
+function wc_admin_register_user_data() {
+	register_rest_field(
+		'user',
+		'woocommerce_meta',
+		array(
+			'get_callback'    => 'wc_admin_get_user_data_values',
```
justinshreve: These array items should be aligned per the coding standards. https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#indentation

lib/admin.php
```diff
@@ -338,5 +338,73 @@ function woocommerce_embed_page_header() {
 	</div>
 	<?php
 }
-
 add_action( 'in_admin_header', 'woocommerce_embed_page_header' );
+
+/**
+ * Registers WooCommerce specific user data to the Wordpress user API.
```
justinshreve: [capital_P_dangit](https://codex.wordpress.org/Function_Reference/capital_P_dangit)() 😄

lib/admin.php
```diff
@@ -338,5 +338,73 @@ function woocommerce_embed_page_header() {
...
+		array(
+			'get_callback'    => 'wc_admin_get_user_data_values',
+			'update_callback' => 'wc_admin_update_user_data_values',
+			'schema' => null,
+		)
+	);
+}
+ add_action( 'rest_api_init', 'wc_admin_register_user_data' );
+
+ /**
```
justinshreve: There are some alignment issues with the docblocks for the next three functions. Looks like extra spaces.

If you run `composer install` and then `./vendor/bin/phpcs lib/admin.php` there will be some warnings about the spacing:

`Found precision alignment of 1 spaces.`

lib/admin.php
```diff
@@ -338,5 +338,73 @@ function woocommerce_embed_page_header() {
...
+  * @param string  $field_id The field id for the user meta.
+  */
+function wc_admin_update_user_data_values( $values, $user, $field_id ) {
+	if ( empty( $values ) || ! is_array( $values ) || 'woocommerce_meta' !== $field_id ) {
+		return;
+	}
+	$fields  = wc_admin_get_user_data_fields();
+	$updates = array();
+	foreach ( $values as $field => $value ) {
+		if ( in_array( $field, $fields ) ) {
```
justinshreve: `403 | WARNING | [ ] Not using strict comparison for in_array; supply true for third argument.`

emdashcodes (commented):
A few comments about formatting in the PHP files.

lib/admin.php
coderkevin: Haha, I'm guilty of copypasta here, honest!

lib/admin.php
coderkevin: Fixed in 8ba7438

lib/admin.php
coderkevin: Fixed in 8ba7438

lib/admin.php
coderkevin: Fixed in 8ba7438

lib/admin.php
coderkevin: Fixed in 8ba7438

lib/admin.php
coderkevin: Maybe we should change the prefix to `wc_admin` instead? That describes the usage and still namespaces it a bit. 3rd parties can still extend via the filter.

lib/admin.php
```diff
@@ -338,5 +338,73 @@ function woocommerce_embed_page_header() {
...
+  * Additional fields can be added in the function below, and then used via wc-admin's currentUser data.
+  *
+  * @return array Fields to expose over the WP user endpoint.
+  */
+function wc_admin_get_user_data_fields() {
+	$user_data_fields = array(
+		'revenue_report_columns',
+	);
+
+	return apply_filters( 'wc_admin_get_user_data_fields', $user_data_fields );
```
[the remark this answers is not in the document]
coderkevin: I can't take credit for this, it's based off [Justin's original PR.](https://github.com/woocommerce/wc-admin/pull/255/files#diff-1469878d77268810c7d104694590899dR280)

lib/admin.php
joshuatf: @coderkevin Sounds like a good idea to me!

lib/admin.php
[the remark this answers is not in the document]
joshuatf: Nice work @justinshreve !

lib/admin.php
coderkevin: In d40ed87, I changed the prefix to `wc_admin_`, but kept the user meta as `woocommerce_meta`. Now there's room for other stuff there, but anything that we use in wc_admin will be prefixed appropriately.

coderkevin:
This PR is updated, rebased, and ready for review once more.

joshuatf (approved):
LGTM! 👍

PR #1059, opened 2018-12-11
Fix item refunds report data

### LINKED ISSUE #1024: Regenerate Report Data fails on Refunded Orders
**Describe the bug**

The report table generation fails when refunded orders are present.

```
Fatal error: Uncaught Error: Call to undefined method WC_Order_Refund::get_customer_id() in /srv/www/woocommerce/public_html/wp-content/plugins/wc-admin/includes/class-wc-admin-api-init.php:222 Stack trace: #0 /srv/www/woocommerce/public_html/wp-content/plugins/wc-admin/includes/class-wc-admin-api-init.php(157): WC_Admin_Api_Init::order_product_lookup_store_init() #1 /srv/www/woocommerce/public_html/wp-content/plugins/woocommerce/includes/api/v2/class-wc-rest-system-status-tools-v2-controller.php(560): WC_Admin_Api_Init::regenrate_report_data() #2 /srv/www/woocommerce/public_html/wp-content/plugins/woocommerce/includes/admin/class-wc-admin-status.php(41): WC_REST_System_Status_Tools_V2_Controller->execute_tool('rebuild_stats') #3 /srv/www/woocommerce/public_html/wp-content/plugins/woocommerce/includes/admin/views/html-admin-page-status.php(34): WC_Admin_Status::status_tools() #4 /srv/www/woocommerce/public_html/wp-content/plugins/woocommerce/includes/admin/class-wc-admin-status.php(20): include_once('/srv/www/woocom...') in /srv/www/woocommerce/public_html/wp-content/plugins/wc-admin/includes/class-wc-admin-api-init.php on line 222
```

**To Reproduce**

1. Ensure you have at least one refunded order that had items specified in the refund (click "refund" on an order and enter a quantity for one or more products in the order during the refund)
2. Go to WooCommerce > Status > Tools
3. Click "Rebuild Reports"
4. See PHP Fatal error

**Expected behavior**

No PHP Fatal 😄 

**Screenshots**

screen shot 2018-12-07 at 9 52 58 am

joshuatf:
Fixes #1024

This PR:
* Fixes the bug where refunding items throws an error and modifies.
* Modifies`wc_order_product_lookup` stats based on refunds given.  Detailed discussion on how refunds should be issued - p7bje6-1dV-p2
* Syncs the `wc_order_product_lookup` table when refunds are processed.

### Screenshots
screen shot 2018-12-11 at 7 04 59 pm

1.  Visit any order in your dashboard.
2.  Try the following with various orders:
    * Refund a line item with a qty of 1 by inputting a quantity next to it in the field.  This item should be removed from the Product report.
    * Refund a partial qty from a line item that has more than 1 qty.  This item should still show in the Product report with a reduced quantity.
    * Refund an amount without specifying a quantity.  No adjustment should be made to the product reports revenue.
    * Set an order status to "Refunded."  All products in this order should be removed from the Product report.

### DIFF
--- a/includes/wc-admin-order-functions.php
+++ b/includes/wc-admin-order-functions.php
@@ -16,7 +16,7 @@ function wc_admin_order_product_lookup_entry( $order_id ) {
 	global $wpdb;
 
 	$order = wc_get_order( $order_id );
-	if ( ! $order ) {
+	if ( ! $order || 'shop_order_refund' === $order->get_type() ) {
 		return;
 	}
 
--- a/includes/wc-admin-order-functions.php
+++ b/includes/wc-admin-order-functions.php
@@ -16,6 +16,8 @@ function wc_admin_order_product_lookup_entry( $order_id ) {
 	global $wpdb;
 
 	$order = wc_get_order( $order_id );
+
+	// This hook gets called on refunds as well, so return early to avoid errors.
 	if ( ! $order || 'shop_order_refund' === $order->get_type() ) {
 		return;
 	}
@@ -23,7 +25,8 @@ function wc_admin_order_product_lookup_entry( $order_id ) {
 	if ( 'refunded' === $order->get_status() ) {
 		$wpdb->delete(
 			$wpdb->prefix . 'wc_order_product_lookup',
-			array( 'order_id' => $order->get_id() )
+			array( 'order_id' => $order->get_id() ),
+			array( '%d' )
 		);
 		return;
 	}
@@ -31,18 +34,19 @@ function wc_admin_order_product_lookup_entry( $order_id ) {
 	$refunds = wc_admin_get_order_refund_items( $order );
 
 	foreach ( $order->get_items() as $order_item ) {
-		$quantity_refunded = isset( $refunds[ $order_item->get_id() ] ) ? $refunds[ $order_item->get_id() ] : 0;
+		$order_item_id     = $order_item->get_id();
+		$quantity_refunded = isset( $refunds[ $order_item_id ] ) ? $refunds[ $order_item_id ] : 0;
 		if ( $quantity_refunded >= $order_item->get_quantity( 'edit' ) ) {
 			$wpdb->delete(
 				$wpdb->prefix . 'wc_order_product_lookup',
-				array( 'order_item_id' => $order_item->get_id() ),
+				array( 'order_item_id' => $order_item_id ),
 				array( '%d' )
 			);
 		} else {
 			$wpdb->replace(
 				$wpdb->prefix . 'wc_order_product_lookup',
 				array(
-					'order_item_id'         => $order_item->get_id(),
+					'order_item_id'         => $order_item_id,
 					'order_id'              => $order->get_id(),
 					'product_id'            => $order_item->get_product_id( 'edit' ),
 					'variation_id'          => $order_item->get_variation_id( 'edit' ),
@@ -66,9 +70,9 @@ function wc_admin_order_product_lookup_entry( $order_id ) {
 	}
 }
 // TODO: maybe replace these with woocommerce_create_order, woocommerce_update_order, woocommerce_trash_order, woocommerce_delete_order, as clean_post_cache might be called in other circumstances and trigger too many updates?
-add_action( 'save_post', 'wc_admin_order_product_lookup_entry', 10, 1 );
+add_action( 'save_post', 'wc_admin_order_product_lookup_entry' );
 add_action( 'woocommerce_order_refunded', 'wc_admin_order_product_lookup_entry' );
-add_action( 'clean_post_cache', 'wc_admin_order_product_lookup_entry', 10, 1 );
+add_action( 'clean_post_cache', 'wc_admin_order_product_lookup_entry' );
 
 /**
  * Get total refund amount and line items refunded.
@@ -79,7 +83,6 @@ add_action( 'clean_post_cache', 'wc_admin_order_product_lookup_entry', 10, 1 );
 function wc_admin_get_order_refund_items( $order ) {
 	$refunds             = $order->get_refunds();
 	$refunded_line_items = array();
-	$single_item_order   = ( 1 === count( $order->get_items() ) );
 	foreach ( $refunds as $refund ) {
 		foreach ( $refund->get_items() as $refunded_item ) {
 			$line_item_id                          = wc_get_order_item_meta( $refunded_item->get_id(), '_refunded_item_id', true );
--- a/includes/wc-admin-order-functions.php
+++ b/includes/wc-admin-order-functions.php
@@ -36,7 +36,7 @@ function wc_admin_order_product_lookup_entry( $order_id ) {
 	foreach ( $order->get_items() as $order_item ) {
 		$order_item_id     = $order_item->get_id();
 		$quantity_refunded = isset( $refunds[ $order_item_id ] ) ? $refunds[ $order_item_id ]['quantity'] : 0;
-		$amount_refunded   = isset( $refunds[ $order_item_id ] ) ? $refunds[ $order_item_id ]['amount'] : 0;
+		$amount_refunded   = isset( $refunds[ $order_item_id ] ) ? $refunds[ $order_item_id ]['subtotal'] : 0;
 
 		if ( $quantity_refunded >= $order_item->get_quantity( 'edit' ) ) {
 			$wpdb->delete(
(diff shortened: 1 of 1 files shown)


includes/wc-admin-order-functions.php
```diff
@@ -16,40 +16,80 @@ function wc_admin_order_product_lookup_entry( $order_id ) {
...
-			)
+			array( 'order_id' => $order->get_id() )
 		);
+		return;
+	}
+
+	$refunds = wc_admin_get_order_refund_items( $order );
+
+	foreach ( $order->get_items() as $order_item ) {
+		$quantity_refunded = isset( $refunds[ $order_item->get_id() ] ) ? $refunds[ $order_item->get_id() ] : 0;
```
timmyc: Should we assign `$order_item->get_id()` to a variable in this loop since it is used numerous times throughout the logic?

includes/wc-admin-order-functions.php
```diff
@@ -16,40 +16,80 @@ function wc_admin_order_product_lookup_entry( $order_id ) {
...
-				'%d',
-				'%d',
-				'%d',
-				'%d',
-				'%d',
-				'%d',
-				'%f',
-				'%s',
-			)
+			array( 'order_id' => $order->get_id() )
```
timmyc: Do you want `array( '%d' )` as a third arg here?

includes/wc-admin-order-functions.php
```diff
@@ -16,40 +16,80 @@ function wc_admin_order_product_lookup_entry( $order_id ) {
...
+/**
+ * Get total refund amount and line items refunded.
+ *
+ * @param object $order WC_Order.
+ * @return array Refunded line items with line item ID as key.
+ */
+function wc_admin_get_order_refund_items( $order ) {
+	$refunds             = $order->get_refunds();
+	$refunded_line_items = array();
+	$single_item_order   = ( 1 === count( $order->get_items() ) );
```
timmyc: Doesn't look like this is used?

includes/wc-admin-order-functions.php
```diff
@@ -16,40 +16,80 @@ function wc_admin_order_product_lookup_entry( $order_id ) {
...
+					'%f',
+					'%s',
+				)
+			);
+		}
 	}
 }
 // TODO: maybe replace these with woocommerce_create_order, woocommerce_update_order, woocommerce_trash_order, woocommerce_delete_order, as clean_post_cache might be called in other circumstances and trigger too many updates?
 add_action( 'save_post', 'wc_admin_order_product_lookup_entry', 10, 1 );
+add_action( 'woocommerce_order_refunded', 'wc_admin_order_product_lookup_entry' );
```
timmyc: Side note, I know there is another issue that talks about using Action Scheduler to queue up jobs for lookup tables. #429 - not sure if we want to start that trend here or do it in a follow-up. Doesn't seem like there is too much heavy lifting going on here, but something worth considering.

timmyc (commented):
Some quick comments on the code, I will do some runs through test scenarios to verify the use cases.

includes/wc-admin-order-functions.php
joshuatf: Added!

includes/wc-admin-order-functions.php
joshuatf: Done!

includes/wc-admin-order-functions.php
```diff
@@ -16,40 +16,80 @@ function wc_admin_order_product_lookup_entry( $order_id ) {
 	global $wpdb;
 
 	$order = wc_get_order( $order_id );
-	if ( ! $order ) {
+	if ( ! $order || 'shop_order_refund' === $order->get_type() ) {
```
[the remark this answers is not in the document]
joshuatf: Good idea.  Added a comment here.

includes/wc-admin-order-functions.php
joshuatf: I tried to implement this but ran into an infinite loop with the scheduler.  Pinged @claudiosanches to see if he had any insight into the issue.

peterfabian: My attempt is at https://github.com/woocommerce/wc-admin/pull/1050, if you want to have a look.

peterfabian:
I have an example order that contains one item and a couple of partial amount refunds (but no qty decrease), but the revenue update ends up as full item, not reduced. Is that intentional?
screenshot 2018-12-14 at 15 07 37
```

REPLACE INTO `wp_wc_order_product_lookup` (`order_item_id`, `order_id`, `product_id`, `variation_id`, `customer_id`, `product_qty`, `product_gross_revenue`, `date_created`) 
VALUES (40, 106, 27, 0, NULL, 1, 90.000000, '2018-06-29 13:37:15')

```

includes/wc-admin-order-functions.php
```diff
@@ -16,39 +16,82 @@ function wc_admin_order_product_lookup_entry( $order_id ) {
...
+			$wpdb->replace(
+				$wpdb->prefix . 'wc_order_product_lookup',
+				array(
+					'order_item_id'         => $order_item_id,
+					'order_id'              => $order->get_id(),
+					'product_id'            => $order_item->get_product_id( 'edit' ),
+					'variation_id'          => $order_item->get_variation_id( 'edit' ),
+					'customer_id'           => ( 0 < $order->get_customer_id( 'edit' ) ) ? $order->get_customer_id( 'edit' ) : null,
+					'product_qty'           => $order_item->get_quantity( 'edit' ) - $quantity_refunded,
+					'product_gross_revenue' => $order_item->get_subtotal( 'edit' ),
```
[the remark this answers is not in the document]
joshuatf: You're right it does.  I don't think we touched directly on this in the refund discussion but everything else we agreed on points to this being the correct way to handle things if:
* Any quantity is left over after the refund
* AND the user entered a refund amount under the "Total" (actually Subtotal but misnamed in WC) column for a specific line item.

I added this change in https://github.com/woocommerce/wc-admin/pull/1059/commits/3b3113a0ac5d91856e5fae020582ac74f6bffc2b - can you let me know if that tests okay for you now @peterfabian?

includes/wc-admin-order-functions.php
joshuatf: I'm happy to get this merged into master if the below changes to the refund amount look okay if you want to handle the action scheduler changes in #1050.

peterfabian (changes requested):

includes/wc-admin-order-functions.php
```diff
@@ -16,39 +16,88 @@ function wc_admin_order_product_lookup_entry( $order_id ) {
...
+	$refunded_line_items = array();
+	foreach ( $refunds as $refund ) {
+		foreach ( $refund->get_items() as $refunded_item ) {
+			$line_item_id                          = wc_get_order_item_meta( $refunded_item->get_id(), '_refunded_item_id', true );
+			if ( ! isset( $refunded_line_items[ $line_item_id ] ) ) {
+				$refunded_line_items[ $line_item_id ]['quantity'] = 0;
+				$refunded_line_items[ $line_item_id ]['subtotal'] = 0;
+			}
+			$refunded_line_items[ $line_item_id ]['quantity'] += absint( $refunded_item['quantity'] );
+			$refunded_line_items[ $line_item_id ]['subtotal'] += absint( $refunded_item['subtotal'] );
```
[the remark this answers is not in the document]
joshuatf: You're absolutely (:trollface:) right- good catch!

includes/wc-admin-order-functions.php
```diff
@@ -16,39 +16,88 @@ function wc_admin_order_product_lookup_entry( $order_id ) {
...
 		);
+		return;
+	}
+
+	$refunds = wc_admin_get_order_refund_items( $order );
+
+	foreach ( $order->get_items() as $order_item ) {
+		$order_item_id     = $order_item->get_id();
+		$quantity_refunded = isset( $refunds[ $order_item_id ] ) ? $refunds[ $order_item_id ]['quantity'] : 0;
+		$amount_refunded   = isset( $refunds[ $order_item_id ] ) ? $refunds[ $order_item_id ]['amount'] : 0;
```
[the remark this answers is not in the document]
joshuatf: Fixed!

joshuatf:
@peterfabian Made those changes.  Can you take another look when you get a chance?

peterfabian (approved):
Thanks, works great now. 

Not sure if net_amount should take into account discounts. If it should, maybe coupons should be applied as well, but we can leave that to another PR, I need to revisit this update routine for segmenting anyway...

PR #1018, opened 2018-12-06
Add Customers Report page

Aljullu:
Part of #918.

This PR creates the _Customers_ report page with a table connected to the SwaggerHub endpoint.

### Screenshots
image

### Detailed test instructions:
- Go to the _Customers_ report.
- Verify there is a table with the correct columns.
- Verify the date range filter is hidden.

### Follow-up work
- Load more data for each _Customer_ from the ` /wc/v3/customers` endpoint.
- Table summary.

PR #1061, opened 2018-12-11
Add customers endpoint to wc-api operations

Aljullu:
We merged #1006 (using `wc-api` for endpoints) and #1018 (adding _Customers_ report) in parallel, and I didn't think that would break the _Customers_ report —it's not loading table data in master. :man_facepalming: 

This PR fixes that adding the customers endpoint to `wc-api` operations list.

### Screenshots
_Before:_
image

_After:_
image

- Verify table data loads.


timmyc (approved):
The Customers table now loads for me 👍

joshuatf (approved):
Also looking good to me ✅  - Merging in so we can unblock https://github.com/woocommerce/wc-admin/pull/1052

PR #1063, opened 2018-12-11
Fix zero filling when ordered by not-date ASC

### LINKED ISSUE #1012: Ordering with 0-filling does not work correctly
**Describe the bug**
e.g. `http://local.wordpress.test/wp-json/wc/v3/reports/coupons/stats?orderby=amount&order=asc&after=2018-05-14T04:21:35&interval=week&page=2`

So... e.g. the first page is full of auto-generated intervals, second page contains 3 db-based records, they should be at the bottom when using `ORDER BY {value} ASC`. However, the auto-generated values are not correct, sometimes they appear both on page 1 and page 2, sometimes it misbehaves slightly differently. Needs to be investigated.

**Expected behavior**
It should work like a charm in all circumstances.

peterfabian:
Fixes #1012 

Removing extra intervals after zero-filling was not using correct offset in case of ordering by something other than date in ascending order.

Make sure zero-filling works correctly using different ordering:
- by date desc
- by date asc
- by [sth else] desc
- by [sth else] asc

In all cases test multiple scenarios, e.g. when the db records are first/last on page 1 of 1, when db records are first/last on page 1/2 of 2, etc.

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-data-store.php
@@ -172,12 +172,17 @@ class WC_Admin_Reports_Data_Store {
 	 * @param int      $db_interval_count Database interval count.
 	 * @param int      $expected_interval_count Expected interval count on the output.
 	 * @param string   $order_by Order by field.
+	 * @param string   $order    Ordering: ASC or DESC.
 	 */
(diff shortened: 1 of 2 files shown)


peterfabian:
Replaced by #1383

PR #1064, opened 2018-12-11
Add autolabeler to automatically label PRs

ryelle:
This PR configures [autolabeler](https://github.com/probot/autolabeler) to automatically label PRs with edits to matched files. This sets up the following rules:

- Any changes to `/packages` adds the **Packages** label
- Any changes to `/packages/components` adds the **Components** label
- Any changes to config or build process files adds the **Build** label

We could probably automate a few others, like **REST API**, **Documentation**, etc– but the **Components** & **Packages** labels are most important so we can track changes between release versions.


emdashcodes (approved):
Good idea! Starting with these labels makes sense.

PR #1066, opened 2018-12-11
Update CONTRIBUTING.md with PR instructions

coderkevin:
When I was trying to commit on a fresh clone, I discovered we have to `composer install` before we can commit. So I updated our documentation to reflect that and gave it a home in some other instructions.

No testing needed, just review the .md file.


timmyc (approved):
Thanks for adding this Kevin - I was also thinking we need to flag some issues as "Good First Issue" to encourage folks to contribute.

Issue #1027, opened 2018-12-07
Advanced Filter: Date


timmyc:
For Customers Report ( #175 ) we need a new advanced filter to be developed to handle the filtering of different fields ( Registration Date, Last Active ) by date.

Designs: p6riRB-3mb-p2

![image](https://user-images.githubusercontent.com/22080/49666328-9750b300-fa0c-11e8-8845-673cb4006aa9.png)

The filter needs to handle the following options:

- Before ( single date picker )
- After ( single date picker )
- Between ( date range picker )

PR #1069, opened 2018-12-12
Add Single Date Picker

psealock:
Working towards #1027 
Depends on https://github.com/WordPress/gutenberg/pull/12855

screen shot 2018-12-14 at 1 37 53 pm

screen shot 2018-12-14 at 1 38 37 pm

## Test

1. Checkout `psealock:add/focusOnMount-to-dropdown` [branch](https://github.com/psealock/gutenberg/tree/add/focusOnMount-to-dropdown) of Gutenberg or just make the two line change in https://github.com/WordPress/gutenberg/pull/12855.
1. View Calendar in DevDocs `/wp-admin/admin.php?page=wc-admin#/devdocs/calendar`
2. Play around with the single date picker
    * Mobile sizes
    * Keyboard interactions
    * VoiceOver
3. If you see a propType warning for `block`, the fix for that is here https://github.com/WordPress/gutenberg/pull/12933

## Key Decisions

* **Make the react-dates datepicker unreachable by keyboard**. There is no way to simultaneously keep focus on the input so keyboard users can type inputs and allow focus on the dropdown for selection via keyboard. One solution is to keep focus on the input, then allow focus on the dropdown via tabbing. I don't like this because it misrepresents the sequential flow of a form and causes confusion, for me at least. React-dates [implementation](https://airbnb.io/react-dates/?selectedKind=SingleDatePicker%20%28SDP%29&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel) chooses similar behaviour.
* **Don't allow full screen expansion of dropdown on mobile screens**. Otherwise, it forces focus away from the input field, effectively removing the ability to use the keyboard.


timmyc:
This is looking really good @psealock - so just wondering if we should consider using the [WP core date picker](https://github.com/WordPress/gutenberg/blob/master/packages/components/src/date-time/index.js#L18) here - apologies for not mentioning that as a suggestion on the original issue:

image

While the design is a bit different, it has passed the a11y review from core. @LevinMedia thoughts from the design side of things?

LevinMedia:
@timmyc @psealock pre the question about the core component... could we re-style so it looks more like what we have now, and behaves like what I describe at #1097 in regards to using number input, and automatically advancing through each part of the input... while maintaining the accessibility?

psealock:
Here is core's DatePicker with some basic Woo styling:

screen shot 2018-12-17 at 9 13 13 am

We should go in this direction. It will involve some functionality upstream pull requests (invalid days, and firstDayOfWeek support) and some stylistic changes (day sizing and box v. circle).

wc-admin's date range picker is heavily tied to its input format (single text input v. GB's multiple inputs) so we can't reuse Core's datepicker there, but we could match styles.

> could we re-style so it looks more like the input we have now, and behaves like what I describe at #1097 in regards to using number input, and automatically advancing through each part of the input... while maintaining the accessibility?

I really like the suggestion by @LevinMedia  in #1097. Happily, Core has separated the inputs from the picker for us, so we can deal with it separately. I think there can be some a11y gains to be made with that approach. I'll explore those thoughts further in that issue.

@LevinMedia how do you feel about the styles in the screen shot and moving towards this direction with the date-range picker for reports?

psealock:
https://github.com/woocommerce/wc-admin/pull/1069/commits/07be9258f9c8b93a163ea36aebf91f42cdc0d6a5 switches to use Core's Datepicker. 

Here are upstream Gutenberg changes needed to get to parity:
* Add `firstDayOfWeek` support.
* Add `invalidDays` support.
* Allow `null` date selection. Current implementation forces a selection of current day when none is provided, which means the calendar shows `today` as selected when mounted.

These seem to be reasonable requests and I think I'll be able to make them happen.

psealock:
Confirmed with @LevinMedia that this is the direction we'll want to pursue. I'm going to mark this as blocked for now until the above^^^ Gutenberg issues can be resolved.

psealock:
@timmyc how do you feel about getting this merged ahead of pending Gutenberg fixes? The only one that affects usability is https://github.com/WordPress/gutenberg/pull/12855 which deals with focus and makes the experience a bit painful for keyboard users.

* https://github.com/WordPress/gutenberg/pull/12855 
* https://github.com/WordPress/gutenberg/pull/12933
* https://github.com/WordPress/gutenberg/pull/12962
* https://github.com/WordPress/gutenberg/pull/12963

Doing so would be on the assumption that these get merged, but allow the continuation of work on filters. In the worst case, we can swap out GB components retrospectively.

packages/components/src/calendar/example.md
```diff
@@ -1,30 +1,60 @@
...
+			<Section component={ false }>
+				<DatePicker
+					date={ after }
+					text={ afterText }
+					error={ afterError } 
+					onUpdate={ onDatePickerUpdate }
+					dateFormat={ dateFormat }
+					invalidDays="none"
+					onUpdate={ onDatePickerUpdate }
+					invalidDays="future"
```
psealock: note to self: `invalidDays` and `onUpdate` are duplicated

timmyc (approved):
One minor comment in the code, but testing out nicely. Verified all seems well with current datepicker usage. I'm fine with landing this before the gberg PRs are merged myself.

Issue #918, opened 2018-11-24
Analytics: Customers Report Table


timmyc:
**Design Post**: p6riRB-3dW-p2
**Mock API**: [https://href.li/?https://app.swaggerhub.com/apis/peterfabian/wc-v3-api/1.0.0#/default/get_reports_customers](get_reports_customers)
**REST API Post**: p7bje6-16h-p2

This issue involves creation of the Customers Report Table. Currently the REST API endpoint has yet to be developed ( #916 ), so we will utilize the Swagger Hub mock endpoint for now. There is a desire to not include extended object information in the base reporting endpoint responses ( i.e. only `customer_id` will be returned ). While it will add query complexity to the client side, this report being a simple one seems like one where we could explore requesting the additional data via the current [`/wc/v3/customers`](http://woocommerce.github.io/woocommerce-rest-api-docs/#retrieve-a-customer) endpoint.

jeffstieler:
Reopening - table summary still needs to be implemented.

PR #1075, opened 2018-12-12
Extend customers table data with data from WC endpoint

Aljullu:
Part of #918.

This PR adds the ability to the _Customers_ report table to request extended info for each customer using a WC endpoint. It adds a prop to the `<ReportTable>` component named `extendItemsMethodNames`. This prop allows passing a method name that will be used to load more data for each table item. Then, `<ReportTable>` takes care of merging the data from both endpoints.

### Open Questions
The designs display columns for _Country_, _City_ and _Postal Code_, but those values might be different between the Shipping and the Billing addresses. In this PR I'm using the values from the Billing address, thoughts @LevinMedia?

### Screenshots
image

Given that we are loading some data from the SwaggerHub endpoint and some other from a WC endpoint, we will have to make one id from both data sets match. That can be done adding this line:
```ES6
itemsData[ 0 ].id = 18;
```
to `client/analytics/components/report-table/utils.js` [line 24](https://github.com/woocommerce/wc-admin/pull/1075/files#diff-7f79b9a09cecd84c214c6ecd1e88433cR24). You will have to modify `18` with a real user id you can get from editing a profile in `/wp-admin/users.php`.

- Go to the _Customers_ report and verify all columns are filled.
- Verify there are no regressions in other reports.

### DIFF
--- a/client/analytics/components/report-table/index.js
+++ b/client/analytics/components/report-table/index.js
@@ -85,8 +85,9 @@ ReportTable.propTypes = {
 	 */
 	endpoint: PropTypes.string,
 	/**
-	 * Method names used to load more data for table items. If omitted, no call will
-	 * be made and only the data returned by the reports endpoint will be used.
+	 * Name of the methods available via `select( 'wc-api' )` that will be used to
+	 * load more data for table items. If omitted, no call will be made and only
+	 * the data returned by the reports endpoint will be used.
 	 */
 	extendItemsMethodNames: PropTypes.shape( {
 		isError: PropTypes.string,
--- a/client/analytics/components/report-table/index.js
+++ b/client/analytics/components/report-table/index.js
@@ -5,7 +5,7 @@
 import { applyFilters } from '@wordpress/hooks';
 import { Component } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
-import { first, get, orderBy } from 'lodash';
+import { get, orderBy } from 'lodash';
 import PropTypes from 'prop-types';
 
 /**
@@ -20,6 +20,7 @@ import { onQueryChange } from '@woocommerce/navigation';
 import ReportError from 'analytics/components/report-error';
 import { getReportChartData, getReportTableData } from 'store/reports/utils';
 import withSelect from 'wc-api/with-select';
+import { extendTableData } from './utils';
 
 const TABLE_FILTER = 'woocommerce_admin_report_table';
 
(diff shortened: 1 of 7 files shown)


LevinMedia:
@Aljullu billing sounds good to me.

client/analytics/components/report-table/utils.js
```diff
@@ -0,0 +1,54 @@
+/** @format */
+/**
+ * External dependencies
+ */
+import { first } from 'lodash';
+
+export function extendTableData( select, props, queriedTableData ) {
+	const { extendItemsMethodNames, itemIdField } = props;
```
psealock: How about a shortcut `if( ! extendItemsMethodNames ) { return queriedTableData; }` ?

psealock (commented):
This is testing well @Aljullu ! I see you've provided a way to make subsequent data requests based on results from a previous request, very nice!

With other endpoints, we've been using query param `extended_info` to add extra information to the `/reports/*` response, such as name, email, country for a customer.  Sorry if its been answered elsewhere, but is there a reason we're not doing it here?

client/analytics/components/report-table/utils.js
psealock: whoops, sorry!

client/analytics/components/report-table/index.js
```diff
@@ -83,6 +84,15 @@ ReportTable.propTypes = {
 	 * The endpoint to use in API calls.
 	 */
 	endpoint: PropTypes.string,
+	/**
+	 * Method names used to load more data for table items. If omitted, no call will
```
justinshreve: Should this specify that these method names need to be available via `select( 'wc-api' )`?

Aljullu: Makes sense! :+1: 

Done: 3d21724494c18af52e75729308e74af61458c96e.

client/wc-api/customers/selectors.js
```diff
@@ -0,0 +1,43 @@
...
+	if ( isNil( lastRequested ) || isNil( lastReceived ) ) {
+		return true;
+	}
+
+	return lastRequested > lastReceived;
+};
+
+const isGetCustomersError = getResource => ( query = {} ) => {
+	const resourceName = getResourceName( 'customers-query', query );
+	return getResource( resourceName ).error;
```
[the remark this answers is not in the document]
Aljullu: Oh, right, that's a good point, I was convinced `error` was a boolean instead of an object.

I fixed it here for now (09d7eafc5ac172b03e05a2f45fa0b5173c214e8f and 12d2550f4ce40613df06bca6fd3d5fc3799a2527) and created a PR with fixes for other selectors that were also affected by this: #1101.

client/analytics/components/report-table/index.js
```diff
@@ -83,6 +84,16 @@ ReportTable.propTypes = {
 	 * The endpoint to use in API calls.
 	 */
 	endpoint: PropTypes.string,
+	/**
+	 * Name of the methods available via `select( 'wc-api' )` that will be used to
+	 * load more data for table items. If omitted, no call will be made and only
+	 * the data returned by the reports endpoint will be used.
+	 */
+	extendItemsMethodNames: PropTypes.shape( {
```
timmyc: Very cool implementation here @Aljullu

timmyc (approved):
This is testing out quite nicely for me - and this approach implemented here I can see being a great way for extension developers to layer in additional data on reports via their own custom REST endpoints - super awesome.

The other fun thing I did was to modify the customer record in another window after loading the report. Checked back to the report tab a few minutes later, and the new data was already there. ✨ via `fresh-data` and @coderkevin - rad!

PR #1078, opened 2018-12-12
Remove trailing slash on base_dir variable.

### LINKED ISSUE #852: Double slashes in wc_test_includes paths in tests/bootstrap.php
**Describe the bug**

In wc_test_includes, the trailing slash in `$wc_tests_framework_base_dir` is appended to strings beginning with slashes - resulting in a double slash in the path.  Doesn't seem to break anything, on Mojave at least, but we should fix it.

timmyc:
Fixes #852 

Verify phpunit tests pass either via CI or locally.

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -60,7 +60,7 @@ function wc_admin_install() {
  * Adds WooCommerce testing framework classes.
  */
 function wc_test_includes() {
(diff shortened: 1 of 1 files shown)


emdashcodes (approved):
👍.
All tests are passing for me locally.

PR #1079, opened 2018-12-12
Add util to packages to fix build.

timmyc:
This is a follow-up to #1017, which was a re-do of #999 😅 and here I'm adding the `getTaxCode` util method to the utils that are part of the built version of the components to fix the lerna builds ( sorry @ryelle! )

To Test:
Follow the instructions in #999

### DIFF
--- a/packages/components/src/search/autocompleters/taxes.js
+++ b/packages/components/src/search/autocompleters/taxes.js
@@ -12,8 +12,7 @@ import { stringifyQuery } from '@woocommerce/navigation';
 /**
  * Internal dependencies
  */
-import { computeSuggestionMatch } from './utils';
-import { getTaxCode } from 'analytics/report/taxes/utils';
+import { computeSuggestionMatch, getTaxCode } from './utils';
 
 /**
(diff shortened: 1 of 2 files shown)


packages/components/src/search/autocompleters/taxes.js
```diff
@@ -12,8 +12,7 @@ import { stringifyQuery } from '@woocommerce/navigation';
 /**
  * Internal dependencies
  */
-import { computeSuggestionMatch } from './utils';
-import { getTaxCode } from 'analytics/report/taxes/utils';
```
psealock: You can delete this file now -> `analytics/report/taxes/utils`

timmyc: It is still used here:

```
analytics/report/taxes/config.js
11:import { getTaxCode } from './utils';
54:						label: getTaxCode( tax ),
```

psealock: Oh I see 🚢

PR #1080, opened 2018-12-12
Add a country filter to the customers report

### LINKED ISSUE #1032: Customers: Country Advanced Filter
Child Task of #175 

image

Create an Advanced Filter config and associated [autocompleter](https://github.com/woocommerce/wc-admin/tree/master/packages/components/src/search/autocompleters) for customer countries

@peterfabian - for the country REST API arg - is that the country code that should be passed in?

@LevinMedia when interacting with this advanced filter - should this be a type-ahead autocompleter or should we just use a select filter populated with country names?

emdashcodes:
Closes #1032.

This PR adds a country filter to the customers report. Instead of making multiple requests while typing, I've added support for preloading WooCommerce's /data endpoints (they return things like countries, currencies, etc) which is useful here, and in future work like providing country drop-downs.

### Screenshots

screen shot 2018-12-12 at 2 05 23 pm

* Open the customers report
* Add a customer filter
* Test the autocompleter, make sure countries show up correctly, and that you can select countries
* Hit filter, and make sure the query string updates
* Refresh, and make sure the previously selected countries show up

### DIFF
--- a/client/analytics/report/customers/config.js
+++ b/client/analytics/report/customers/config.js
@@ -3,6 +3,7 @@
  * External dependencies
  */
 import { __, _x } from '@wordpress/i18n';
+import { decodeEntities } from '@wordpress/html-entities';
 import { getRequestByIdString } from '../../../lib/async-requests';
 import { NAMESPACE } from '../../../store/constants';
 
@@ -30,7 +31,7 @@ export const advancedFilters = {
 		name: {
 			labels: {
 				add: __( 'Name', 'wc-admin' ),
-				placeholder: __( 'Search customer name', 'wc-admin' ),
+				placeholder: __( 'Search', 'wc-admin' ),
 				remove: __( 'Remove customer name filter', 'wc-admin' ),
 				rule: __( 'Select a customer name filter match', 'wc-admin' ),
 				/* translators: A sentence describing a Product filter. See screen shot for context: https://cloudup.com/cCsm3GeXJbE */
@@ -57,6 +58,41 @@ export const advancedFilters = {
 					id: product.id,
 					label: product.name,
 				} ) ),
+			},
+		},
+		country: {
+			labels: {
+				add: __( 'Country', 'wc-admin' ),
+				placeholder: __( 'Search', 'wc-admin' ),
+				remove: __( 'Remove country filter', 'wc-admin' ),
+				rule: __( 'Select a country filter match', 'wc-admin' ),
+				/* translators: A sentence describing a Product filter. See screen shot for context: https://cloudup.com/cCsm3GeXJbE */
+				title: __( 'Country {{rule /}} {{filter /}}', 'wc-admin' ),
+				filter: __( 'Select country', 'wc-admin' ),
+			},
+			rules: [
+				{
+					value: 'includes',
+					/* translators: Sentence fragment, logical, "Includes" refers to countries including a given country or countries. Screenshot for context: https://cloudup.com/cCsm3GeXJbE */
+					label: _x( 'Includes', 'countries', 'wc-admin' ),
+				},
+				{
+					value: 'excludes',
+					/* translators: Sentence fragment, logical, "Excludes" refers to countries excluding a given country or countries. Screenshot for context: https://cloudup.com/cCsm3GeXJbE */
+					label: _x( 'Excludes', 'countriess', 'wc-admin' ),
+				},
+			],
+			input: {
+				component: 'Search',
+				type: 'countries',
+				getLabels:
+					( wcSettings.dataEndpoints &&
+						wcSettings.dataEndpoints.countries &&
+						wcSettings.dataEndpoints.countries.map( country => ( {
+							id: country.code,
+							label: decodeEntities( country.name ),
+						} ) ) ) ||
+					[],
 			},
 		},
 	},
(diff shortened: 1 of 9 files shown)


packages/components/src/search/index.js
```diff
@@ -225,7 +228,10 @@ Search.propTypes = {
 	 */
 	selected: PropTypes.arrayOf(
 		PropTypes.shape( {
-			id: PropTypes.number.isRequired,
+			id: PropTypes.oneOf( [
+				PropTypes.number,
+				PropTypes.string,
+			] ).isRequired,
```
psealock: I'm getting propType warnings for `Tag` as well as `Search`.

packages/components/src/search/autocompleters/countries.js
```diff
@@ -0,0 +1,50 @@
...
+		const name = decodeEntities( country.name );
+		const match = computeSuggestionMatch( name, query ) || {};
+		return [
+			<span key="name" className="woocommerce-search__result-name" aria-label={ name }>
+				{ match.suggestionBeforeMatch }
+				<strong className="components-form-token-field__suggestion-match">
+					{ match.suggestionMatch }
+				</strong>
+				{ match.suggestionAfterMatch }
+			</span>,
```
psealock: I don't think its in the design, but it might be cool to include the `<Flag />` component in a suggestion similar to how we show products autocompleter suggestions:

screen shot 2018-12-13 at 11 10 15 am

packages/components/src/filters/advanced/search-filter.js
```diff
@@ -25,7 +25,14 @@ class SearchFilter extends Component {
...
 
 		if ( filter.value.length ) {
-			config.input.getLabels( filter.value, query ).then( this.updateLabels );
+			if ( typeof config.input.getLabels === 'function' ) {
+				config.input.getLabels( filter.value, query ).then( this.updateLabels );
+			} else {
+				const filters = filter.value.split( ',' );
+				this.state.selected = config.input.getLabels.filter( ( label ) => {
+					return filters.includes( label.id );
+				} );
```
psealock: In writing the `getLabels` functionality I was going for this component being as agnostic as possible to the logic required to create the `this.state.selected`.

I think by putting this into a `getLabels` function, we can keep this component clean. What do you think?

lib/client-assets.php
```diff
@@ -142,6 +142,22 @@ function wc_admin_print_script_settings() {
...
+	if ( function_exists( 'gutenberg_preload_api_request' ) ) {
+		$preload_function = 'gutenberg_preload_api_request';
+	} else {
+		$preload_function = 'rest_preload_api_request';
+	}
+
+	$preload_data = array_reduce(
+		array_values( $preload_data_endpoints ),
+		$preload_function
+	);
```
psealock: Cool, did not know about this.

psealock (commented):
Nice job here! Just a few comments and an issue with `getLabels` and keeping `SearchFilter` as free from specific logic as possible.

packages/components/src/search/index.js
justinshreve: Thanks - even though I had `SCRIPT_DEBUG` enabled, I wasn't getting un-minified files / seeing the warnings but I got them showing now. Fix here and in `Tag`.

packages/components/src/filters/advanced/search-filter.js
justinshreve: Cleaned this up a bit, and moved back to a function in the config. Thanks for the feedback here!

client/analytics/report/customers/config.js
```diff
@@ -59,6 +60,41 @@ export const advancedFilters = {
...
+			rules: [
+				{
+					value: 'includes',
+					/* translators: Sentence fragment, logical, "Includes" refers to countries including a given country or countries. Screenshot for context: https://cloudup.com/cCsm3GeXJbE */
+					label: _x( 'Includes', 'countries', 'wc-admin' ),
+				},
+				{
+					value: 'excludes',
+					/* translators: Sentence fragment, logical, "Excludes" refers to countries excluding a given country or countries. Screenshot for context: https://cloudup.com/cCsm3GeXJbE */
+					label: _x( 'Excludes', 'countriess', 'wc-admin' ),
```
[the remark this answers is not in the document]
justinshreve: Fixed!

packages/components/src/search/autocompleters/countries.js
justinshreve: Great suggestion! 

screen shot 2018-12-13 at 11 35 36 am

justinshreve:
@psealock @Aljullu Thank you for the feedback! I've addressed everything above.

packages/components/src/filters/advanced/README.md
```diff
@@ -84,7 +84,7 @@ input: {
 ```
 
 `type`: A string Autocompleter type used by the [Search Component](https://github.com/woocommerce/wc-admin/tree/master/packages/components/src/search).
-`getLabels`: A function returning a Promise resolving to an array of objects with `id` and `label` properties.
+`getLabels`: Either an array of objects with `id` and `label` properties, or a function returning a Promise resolving to an array of objects with `id` and `label` properties.
```
psealock: This can be reverted now. I didn't know you could return an `async` function as you did 👍 . I thought it would need to be `return Promise.resolve( [] );`, so maybe that can be described here?

psealock (approved):
Looks great, thanks for the changes. Just a comment on a README.

packages/components/src/search/autocompleters/countries.js
timmyc: Nice addition, and great to see our old friend the Flag component get some love!

timmyc:
Only addition that I'd like to see prior to merge is to add some entries to the changelog in the components directory that talk about the new additions here. Or you can add them to #1088 too. Either works and I can wait to run the build till after this is merged.

justinshreve:
Done in this PR @timmyc.

PR #1081, opened 2018-12-12
Add `TextControlWithAffixes` component

jeffstieler:
Yeah, this branch name isn't great. 😅 

This PR seeks to add a `<TextControlWithAffixes />` component that can be used to implement the currency-prefixed controls in https://github.com/woocommerce/wc-admin/issues/1028. See below:

The component is essentially a rewrite of [`TextControl`](https://github.com/WordPress/gutenberg/tree/master/packages/components/src/text-control) that incorporates the affixes from [`FormTextInputWithAffixes`](https://github.com/Automattic/wp-calypso/tree/master/client/components/forms/form-text-input-with-affixes).

### Screenshots

screen shot 2018-12-12 at 1 29 56 pm

- Open devdocs and scroll to `TextControlWithAffixes`
- Click the component name and review full documentation
- Check that component renders well in multiple viewport sizes and devices
- Modify the example code and try out different props?

_Note: specifically requesting an accessibility review from @ryelle as discussed in Slack_

### DIFF
--- a/docs/components/text-control-with-affixes.md
+++ b/docs/components/text-control-with-affixes.md
@@ -1,0 +1,71 @@
+Text Control With Affixes
+============================
+
+This component is essentially a wrapper (really a reimplementation) around the TextControl component that adds support for affixes, i.e. the ability to display a fixed part either at the beginning or at the end of the text input.
+
+
+
+Props
+-----
+
+The set of props accepted by the component will be specified below.
+Props not included in this set will be applied to the input element.
+
+### `label`
+
+If this property is added, a label will be generated using label property as the content.
+
+- Type: `String`
+
+### `help`
+
+If this property is added, a help text will be generated using help property as the content.
+
+- Type: `String`
+
+### `type`
+
+Type of the input element to render. Defaults to "text".
+
+- Type: `String`
+- Default: "text"
+
+### `value`
+
+The current value of the input.
+
+- **Required**
+- Type: `String`
+
+### `className`
+
+The class that will be added with "components-base-control" to the classes of the wrapper div.
+If no className is passed only components-base-control is used.
+
+- Type: `String`
+
+### `onChange`
+
+A function that receives the value of the input.
+
+- **Required**
+- Type: `function`
+
+### `prefix`
+
+Markup to be inserted at the beginning of the input.
+
+- Type: ReactNode
+
+### `suffix`
+
+Markup to be appended at the end of the input.
+
+- Type: ReactNode
+
+### `noWrap`
+
+A flag that prevents the prefix and suffix from wrapping when the component is displayed on small viewports. This basically disables the corresponding breakpoint.
+
+- Type: Boolean
+- Default: null
--- a/docs/components/text-control-with-affixes.md
+++ b/docs/components/text-control-with-affixes.md
@@ -1,64 +1,69 @@
-Text Control With Affixes
-============================
+`TextControlWithAffixes` (component)
+====================================
 
-This component is essentially a wrapper (really a reimplementation) around the TextControl component that adds support for affixes, i.e. the ability to display a fixed part either at the beginning or at the end of the text input.
-
-
+This component is essentially a wrapper (really a reimplementation) around the
+TextControl component that adds support for affixes, i.e. the ability to display
+a fixed part either at the beginning or at the end of the text input.
 
 Props
 -----
 
-The set of props accepted by the component will be specified below.
-Props not included in this set will be applied to the input element.
-
 ### `label`
+
+- Type: String
+- Default: null
 
 If this property is added, a label will be generated using label property as the content.
 
-- Type: `String`
-
 ### `help`
+
+- Type: String
+- Default: null
 
 If this property is added, a help text will be generated using help property as the content.
 
-- Type: `String`
-
 ### `type`
+
+- Type: String
+- Default: `'text'`
 
 Type of the input element to render. Defaults to "text".
 
-- Type: `String`
-- Default: "text"
-
 ### `value`
+
+- **Required**
+- Type: String
+- Default: null
 
 The current value of the input.
 
-- **Required**
-- Type: `String`
-
 ### `className`
+
+- Type: String
+- Default: null
 
 The class that will be added with "components-base-control" to the classes of the wrapper div.
 If no className is passed only components-base-control is used.
 
-- Type: `String`
-
 ### `onChange`
+
+- **Required**
+- Type: Function
+- Default: null
 
 A function that receives the value of the input.
 
-- **Required**
-- Type: `function`
-
 ### `prefix`
+
+- Type: ReactNode
+- Default: null
 
 Markup to be inserted at the beginning of the input.
 
-- Type: ReactNode
-
 ### `suffix`
+
+- Type: ReactNode
+- Default: null
 
 Markup to be appended at the end of the input.
 
-- Type: ReactNode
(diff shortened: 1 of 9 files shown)


packages/components/src/text-control-with-affixes/example.md
```diff
@@ -0,0 +1,21 @@
...
+        />
+        <TextControlWithAffixes
+            prefix="Prefix"
+            suffix="Suffix"
+            placeholder="Placeholder text..."
+        />
+        <TextControlWithAffixes
+            suffix="Suffix"
+            placeholder="Placeholder text..."
+        />
```
psealock: Can you add an example with the `help` prop set?

psealock (commented):
Nice! Looking good. I made a comment about using the `help` prop because I went through it with VoiceOver and it's just an input with a placeholder and there is no mention of prefix/suffix. `help` can provide information to VoiceOver.

While you're there, if you add state and an `onChange` method, the component can be fully tested.

packages/components/src/text-control-with-affixes/style.scss
```diff
@@ -0,0 +1,135 @@
...
+}
+
+@mixin no-suffix-wrap() {
+	border-left: none;
+}
+
+.text-control-with-affixes__prefix,
+.text-control-with-affixes__suffix {
+	position: relative;
+	background: $core-grey-light-100;
```
LevinMedia: Can you make this background white? that's the same color as the background it will be sitting on top of.

packages/components/src/text-control-with-affixes/style.scss
```diff
@@ -0,0 +1,135 @@
...
+	border: 1px solid $core-grey-light-500;
+	color: $gray-text;
+	padding: 7px 14px;
+	white-space: nowrap;
+	flex: 1 0 auto;
+	font-size: 14px;
+	line-height: 1.5;
+}
+
+.text-control-with-affixes__prefix {
```
LevinMedia: Can you prevent this from wrapping to two lines when the viewport is less than 400px please?

LevinMedia (commented):
heya, looking good, just a few small notes on styles!

packages/components/src/text-control-with-affixes/style.scss
LevinMedia: just clarifying that last comment... 

 At the smallest viewport sizes, I noticed the entry field wraps under the the prefix, and decreases in height. 

screen shot 2018-12-12 at 8 17 08 pm

Can you keep it all on the same line, even at the smallest sizes, please?

jeffstieler: Do we never want it to wrap? I can remove the `noWrap` stuff (just ported from Calypso) as well then.

jeffstieler:
> Nice! Looking good. I made a comment about using the help prop because I went through it with VoiceOver and it's just an input with a placeholder and there is no mention of prefix/suffix. help can provide information to VoiceOver.

Do you have any ideas on how to properly surface prefixes and suffixes in VO help text? I checked the Calypso devdocs ([#](https://wpcalypso.wordpress.com/devdocs/design/form-fields)) for an example and the affixes aren't read in VO.

> While you're there, if you add state and an onChange method, the component can be fully tested.

Done in https://github.com/woocommerce/wc-admin/pull/1081/commits/1cc6d68b8bacc96e04871b1aae7cc8da9ee618e5

packages/components/src/text-control-with-affixes/style.scss
LevinMedia: Yeah, I'd just prefer it to never wrap, so that'd be great. Thanks!

packages/components/src/text-control-with-affixes/style.scss
jeffstieler: Done in e3a45da.

packages/components/src/text-control-with-affixes/style.scss
jeffstieler: Done in e3a45da.

docs/components/text-control-with-affixes.md
```diff
@@ -0,0 +1,71 @@
+Text Control With Affixes
```
ryelle: These docs files are actually generated from inline docs in the component, not hand-written – I've [added actual instructions for this,](https://github.com/woocommerce/wc-admin/blob/edf68545802824cf1cf9bc4d3e803d5cfedf3a54/docs/documentation.md) can you update the `TextControlWithAffixes` file with this as inline docs?

packages/components/src/text-control-with-affixes/example.md
```diff
@@ -0,0 +1,31 @@
+```jsx
+import { TextControlWithAffixes } from '@woocommerce/components';
+
+const MyTextControlWithAffixes = withState( {
+	first: '',
+    second: '',
```
ryelle: super nitpick– mix of tabs and spaces here. AFAIK we don't have a standard in these md files, but it looks like we use tabs in the others.

packages/components/src/text-control-with-affixes/style.scss
```diff
@@ -0,0 +1,66 @@
+.text-control-with-affixes {
+	display: inline-flex;
+	flex-direction: row;
+    width: 100%;
```
ryelle: This file is also a mix of tabs & spaces – stylelint is picking it up & flagging it, but the tests are still considered passing. If you run `npm run lint:css` you'll see the issues.

docs/components/text-control-with-affixes.md
jeffstieler: Done in 5e2d473.

packages/components/src/text-control-with-affixes/index.js
```diff
@@ -0,0 +1,54 @@
+/** @format */
```
ryelle: This file is using only spaces, not tabs – our JS files should be using tab indentation.

ryelle (commented):
leaving my style comments for now, but i'm still going to run through a11y testing

packages/components/src/text-control-with-affixes/example.md
jeffstieler: Fixed in cc92f19.

packages/components/src/text-control-with-affixes/style.scss
jeffstieler: Fixed in cc92f19.

packages/components/src/text-control-with-affixes/index.js
jeffstieler: Fixed in cc92f19.

ryelle:
The commits I just pushed update the examples a little for better testing, since I wasn't sure how characters like `$` would be read by a screen reader (with my settings, it's read as "dollar").

screen shot 2018-12-13 at 12 05 31 pm

Since `aria-describedby` can take a list of IDs, we can connect the affixes by IDs along with the help text. This reads off in order, `[label] [field type] [help] [prefix] [suffix]`– for example, "Text field with prefix and help text. Edit text. This is some help text. $", or "Text field with both affixes. Edit text. Prefix Suffix"

The exact order and verbosity will be different on other screen readers, and can depend on user settings.

If we're only ever using prefix/suffix to indicate field type ('$', '%', etc) this method makes sense. If we were to use prefix/suffix to make a sentence, that won't work here and would need a different solution for screen reader users.

psealock (approved):
Looking good and testing well. Given this

>The exact order and verbosity will be different on other screen readers, and can depend on user settings.

I think its good to merge because the help text and label are enough to gather what's required from the user.

PR #1083, opened 2018-12-13
 Add documentation for how to write documentation

ryelle:
We have an automated method for creating the documentation files, and this was not clear from existing project docs. This PR adds a step-by-step guide to adding the inline docs in the format expected, building the markdown file, and creating the devdocs example.

[View the new docs guide](https://github.com/woocommerce/wc-admin/blob/edf68545802824cf1cf9bc4d3e803d5cfedf3a54/docs/documentation.md)

This also updates docsify, and adds prismjs as a peer depdendency to fix an issue where docsify `Cannot find module 'prismjs/components/prism-markup-templating'`.

**To test**

- Read the guide, and make sure it makes sense


package.json
```diff
@@ -134,6 +134,7 @@
     "interpolate-components": "1.1.1",
     "lodash": "^4.17.11",
     "marked": "0.5.2",
+    "prismjs": "^1.15.0",
```
jeffstieler: Did we need to add this to our dependencies? Isn't it a dependency of `docsify-cli` > `docsify`?

jeffstieler (commented):
I used these instructions to build documentation for the `TextControlWithAffixes` component. 👍 

One small question about the added `prismjs` dependency, otherwise this looks good.

package.json
ryelle: It is, but I was still getting the `Cannot find module 'prismjs/components/prism-markup-templating'` issue when trying to run `npx docsify serve docs`, until I added this. It's probably an issue upstream, but I don't have the time to track it down right now 😕

psealock (approved):
The guide is great and makes sense.

PR #1085, opened 2018-12-13
Customers report: Name Advanced Filter

### LINKED ISSUE #1029: Customers: Name Advanced Filter
Child Task of #175 

image

Create an Advanced Filter config and associated [autocompleter](https://github.com/woocommerce/wc-admin/tree/master/packages/components/src/search/autocompleters) for customer names

### LINKED ISSUE #1084: Add search box to tables without comparison
Currently, tables only display the search box when they get a `compareBy` prop, but in the _Customers_ report, we want to display a search box even though customers can't be compared.

image

Aljullu:
Fixes #1029.
Fixes #1084.

Adds the Name Advanced Filter to the _Customers_ report and a search box to the _Customers_ table.

The `<TableCard>` component was extended with two new props: `searchBy` and `searchParam`.

### Screenshots
image

There are several things to test here.

**Test 1**
- Go to the _Customers_ report without any filter applied `wc-admin#/analytics/customers`.
- Select _Advanced filters_ under the filter dropdown and _Add filter_ of type _Name_.
- Type any letter in the text box and verify suggestions appear.

**Test 2**
- Go to the _Customers_ report without any filter applied `wc-admin#/analytics/customers`.
- Search for a customer name in the table search box.
- Verify suggestions appear and, when selecting one, it's added to the filter.

**Test 3**
- Go to the _Customers_ report without any filter applied `wc-admin#/analytics/customers`.
- Select _Advanced filters_ under the filter dropdown and _Add filter_ of type _Name_.
- Write any name and click _Filter_.
- Search for a customer name in the table search box
- Verify suggestions appear but the current selected customer in the filters is not in the list. (In the screenshot above, notice _Aditya Jones_ doesn't appear as a suggestion).
- Verify that, when selecting another customer, it's added to the filter without deleting the previous one.

**Test 4**
- Test there are no regressions in other filters/reports.

### Future work
- The table search box will also have to search by email. But since that would have increased this PR complexity, I left it for now.


packages/components/src/search/autocompleters/customers.js
```diff
@@ -13,6 +13,9 @@ import { stringifyQuery } from '@woocommerce/navigation';
  * Internal dependencies
  */
 import { computeSuggestionMatch } from './utils';
+import { SWAGGERNAMESPACE } from 'store/constants';
```
ryelle: I know you're still working on this, but this file won't be available outside the project- can you define `SWAGGERNAMESPACE` in this file instead? Or not use it, we ended up removing the NAMESPACE const in the other autocompleters.

packages/components/src/search/index.js
```diff
@@ -221,12 +232,14 @@ Search.propTypes = {
 	 */
 	placeholder: PropTypes.string,
 	/**
-	 * An array of objects describing selected values.
+	 * An array of objects describing selected values. If the label of the selected
+	 * value is omitted, the Tag of that value will not be rendered inside the
+	 * search box.
```
psealock: Is this so that search terms found in things like email are not rendered?

psealock (commented):
Looking good @Aljullu 

Just the comment about labels on tags. If I understand correctly, if the label is undefined, then a Tag is created, but it won't get rendered. 

Does `customer.first_name + ' ' + customer.last_name` not exist for some values?

packages/components/src/table/index.js
```diff
@@ -268,12 +276,13 @@ class TableCard extends Component {
 							{ labels.compareButton || __( 'Compare', 'wc-admin' ) }
 						</CompareButton>
 					),
-					compareBy && (
+					( compareBy || searchBy ) && (
 						<Search
 							key="search"
 							placeholder={ labels.placeholder || __( 'Search by item name', 'wc-admin' ) }
```
joshuatf: Should we pass in props above to change this to `Search by customer name`?

joshuatf (commented):
Testing well!  Couple minor comments, but overall this looks great 👍

packages/components/src/search/index.js
Aljullu: For the table search box, we need to pass a `selected` prop so the selected elements are excluded from the autocompleter suggestions, but we don't want them to be displayed in `<Tag>`s because they are already shown in the filters search box above. That's why I added the possibility of passing `selected` items that will be hidden.

> Just the comment about labels on tags. If I understand correctly, if the label is undefined, then a Tag is created, but it won't get rendered.

Not really, the `Tag` for selected items without label should not be created at all. See:

https://github.com/woocommerce/wc-admin/pull/1085/files#diff-9d5e585ab84ca2db27f702f4836fe65cR97

client/analytics/report/customers/config.js
```diff
@@ -50,12 +50,11 @@ export const advancedFilters = {
...
-				// Use products autocompleter for now, see https://github.com/woocommerce/wc-admin/issues/1029 for progress
 				component: 'Search',
-				type: 'products',
-				getLabels: getRequestByIdString( NAMESPACE + 'products', product => ( {
-					id: product.id,
-					label: product.name,
+				type: 'customers',
+				getLabels: getRequestByIdString( NAMESPACE + 'customers', customer => ( {
+					id: customer.id,
+					label: customer.first_name + ' ' + customer.last_name,
```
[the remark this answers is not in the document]
Aljullu: Good catch! I implemented it here: b03206038617776a2f1080d5d38793cace577d6e.

packages/components/src/table/index.js
Aljullu: Absolutely! Fixed: 416bba29ea3cfa633caefac55b41ca351157caa0

Aljullu:
@psealock @joshuatf want to take another look? :slightly_smiling_face:

psealock (approved):
LGTM

PR #1089, opened 2018-12-14
Add Advanced Numerical Value Filter

### LINKED ISSUE #1028: Advanced Filter: Numerical Value
Design Post: p6riRB-3mb-p2
For the Customers Report ( #175 ) we need to create a new advanced for numerical values:

image

If possible it would be nice for this advanced filter to support both numerical and currency based values - perhaps toggled via a config prop, because both filters are needed to add the following comparisons, with the only design difference being the prefixing of the currency symbol and input formatting ( side note, some currencies display the symbol on the right of the currency, should we support that in the filter @LevinMedia ? )

- More Than
- Less Than
- Beteween

jeffstieler:
Fixes #1028.

This PR adds a `'Number'` component type to advanced filtering. It has an optional `type` parameter that can be `'currency'` to create an input affixed with the store's currency symbol.

**Note: this PR removes the rounded corners from `TextControlWithAffixes`.**

### Screenshots

screen shot 2018-12-14 at 10 00 31 am

- Go to WooCommerce > Devdocs
- Scroll to Advanced Filters
- Add both "item quantity" and "subtotal" filters
- Note that subtotal inputs have the store's currency symbol (and position respects the store's currency symbol position option (WooCommerce > Settings > General > Currency Position))
- Verify that the query string is sensibly generated based on the input values (note: the "between" mode will comma delimit the two input values)
- Verify that the legend text generated is sensible (check while interacting)

### DIFF
--- a/packages/components/src/filters/advanced/index.js
+++ b/packages/components/src/filters/advanced/index.js
@@ -27,6 +27,7 @@ import Card from '../../card';
 import Link from '../../link';
 import SelectFilter from './select-filter';
 import SearchFilter from './search-filter';
+import NumberFilter from './number-filter';
 
 const matches = [
 	{ value: 'all', label: __( 'All', 'wc-admin' ) },
@@ -171,6 +172,15 @@ class AdvancedFilters extends Component {
 								) }
 								{ 'Search' === input.component && (
 									<SearchFilter
+										filter={ filter }
+										config={ config.filters[ key ] }
+										onFilterChange={ this.onFilterChange }
+										isEnglish={ isEnglish }
+										query={ query }
+									/>
+								) }
+								{ 'Number' === input.component && (
+									<NumberFilter
 										filter={ filter }
 										config={ config.filters[ key ] }
 										onFilterChange={ this.onFilterChange }
--- a/packages/components/src/filters/advanced/number-filter.js
+++ b/packages/components/src/filters/advanced/number-filter.js
@@ -10,9 +10,27 @@ import classnames from 'classnames';
 import { _x } from '@wordpress/i18n';
 
 class NumberFilter extends Component {
+	getBetweenString() {
+		return _x(
+			'{{moreThan /}} and {{lessThan /}}',
+			'Numerical range inputs arranged on a single line',
+			'wc-admin'
+		);
+	}
 	getLegend( filter, config ) {
 		const rule = find( config.rules, { value: filter.rule } ) || {};
-		const filterStr = '';
+		let filterStr = filter.value || '';
+
+		if ( 'between' === rule.value ) {
+			const [ moreThan, lessThan ] = filterStr.split( ',' );
+			filterStr = interpolateComponents( {
+				mixedString: this.getBetweenString(),
+				components: {
+					moreThan: <Fragment>{ moreThan }</Fragment>,
+					lessThan: <Fragment>{ lessThan }</Fragment>,
+				},
+			} );
+		}
 
 		return interpolateComponents( {
 			mixedString: config.labels.title,
@@ -57,11 +75,7 @@ class NumberFilter extends Component {
 		};
 
 		return interpolateComponents( {
-			mixedString: _x(
-				'{{moreThan /}} and {{lessThan /}}',
-				'Numerical range inputs arranged on a single line',
-				'wc-admin'
-			),
+			mixedString: this.getBetweenString(),
 			components: {
 				lessThan: (
 					<TextControl
@@ -91,17 +105,17 @@ class NumberFilter extends Component {
 		const children = interpolateComponents( {
 			mixedString: labels.title,
 			components: {
+				rule: (
+					<SelectControl
+						className="woocommerce-filters-advanced__rule"
+						options={ rules }
+						value={ rule }
+						onChange={ partial( onFilterChange, key, 'rule' ) }
+						aria-label={ labels.rule }
+					/>
+				),
 				filter: (
-					<Fragment>
-						<SelectControl
-							className="woocommerce-filters-advanced__rule"
-							options={ rules }
-							value={ rule }
-							onChange={ partial( onFilterChange, key, 'rule' ) }
-							aria-label={ labels.rule }
-						/>
-						<div>{ this.getFilterInput() }</div>
-					</Fragment>
+					<div>{ this.getFilterInput() }</div>
 				),
 			},
 		} );
--- a/packages/components/src/filters/advanced/number-filter.js
+++ b/packages/components/src/filters/advanced/number-filter.js
@@ -10,29 +10,6 @@ import classnames from 'classnames';
 import { _x } from '@wordpress/i18n';
 
 class NumberFilter extends Component {
-	constructor( { config } ) {
-		super( ...arguments );
-
-		this.rules = config.rules || [
-			{
-				value: 'lessthan',
-				/* translators: Sentence fragment, logical "Less Than" (x < 10). Screenshot for context: https://cloudup.com/cZaDYBpzqsN */
-				label: _x( 'Less Than', 'number', 'wc-admin' ),
-			},
-			{
-				value: 'morethan',
-				/* translators: Sentence fragment, logical "More Than" (x > 10). Screenshot for context: https://cloudup.com/cZaDYBpzqsN */
-				label: _x( 'More Than', 'number', 'wc-admin' ),
-			},
-			{
-				value: 'between',
-				/* eslint-disable-next-line max-len */
-				/* translators: Sentence fragment, logical "Between" (1 < x < 10). Screenshot for context: https://cloudup.com/cZaDYBpzqsN */
-				label: _x( 'Between', 'number', 'wc-admin' ),
-			},
-		];
-	}
-
 	getLegend( filter, config ) {
 		const rule = find( config.rules, { value: filter.rule } ) || {};
 		const filterStr = '';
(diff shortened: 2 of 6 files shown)


packages/components/src/filters/advanced/index.js
```diff
@@ -178,6 +179,15 @@ class AdvancedFilters extends Component {
 										query={ query }
 									/>
 								) }
+								{ 'Number' === input.component && (
+									<NumberFilter
+										filter={ filter }
+										config={ config.filters[ key ] }
+										onFilterChange={ this.onFilterChange }
+										isEnglish={ isEnglish }
```
joshuatf: Not specific to this PR, but I'm curious what this does.  Looks like it sets specific grid widths for English.  Is the plan to set column widths for all languages or is there something specific to English that we need to do this for? /cc @psealock

packages/components/src/filters/advanced/number-filter.js
```diff
@@ -0,0 +1,184 @@
...
+			}
+
+			filterStr = interpolateComponents( {
+				mixedString: this.getBetweenString(),
+				components: {
+					moreThan: <Fragment>{ moreThan }</Fragment>,
+					lessThan: <Fragment>{ lessThan }</Fragment>,
+				},
+			} );
+		} else if ( 'currency' === inputType ) { // need to format a single input value as currency
```
joshuatf: Do you think we could run `.split( ',' )` on all rules (not just between) and map any changes like `formatCurrency()` so we don't have to repeat them?  Or does that add more overhead overall?

packages/components/src/filters/advanced/number-filter.js
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	getFormControl( type, value, onChange ) {
+		if ( 'currency' === type ) {
+			const currencySymbol = get( wcSettings, [ 'currency', 'symbol' ] );
+			const symbolPosition = get( wcSettings, [ 'currency', 'position' ] );
+
+			return (
+				0 === symbolPosition.indexOf( 'right' )
+				? <TextControlWithAffixes
```
joshuatf: Do you think it's worth adding an `.is-error` class if moreThan > lessThan?

joshuatf (commented):
Awesome component, @jeffstieler 🎉 

Added a few comments, but overall this code looks good and tests well.

This may not be an issue outside of devdocs, but should we add some styling to change the input widths?  I get a bit of overlapping use the 'between' rule:

screen shot 2018-12-14 at 11 29 28 am
screen shot 2018-12-14 at 11 30 12 am

packages/components/src/filters/advanced/number-filter.js
jeffstieler: I don't think it adds any overhead, but I'm not sure it simplifies things either. If we always `.split()`, we have to figure out how to handle cases where we switch from a "between" filter that only has the first input populated (the "morethan" side) to a "less than" filter. If we split that string in the same fashion as above, `lessThan` is `undefined`.. do we use that value, or take the known good `moreThan` in its place?

packages/components/src/filters/advanced/number-filter.js
```diff
@@ -0,0 +1,184 @@
...
+import { get, find, partial } from 'lodash';
+import interpolateComponents from 'interpolate-components';
+import classnames from 'classnames';
+import { _x } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import TextControlWithAffixes from '../../text-control-with-affixes';
+import { formatCurrency } from '../../../../currency/src';
```
ryelle: `formatCurrency` is not in this package, so you should use `@woocommerce/currency` here. The reason for relative imports is to not create circular dependencies inside a single package– so files in components should be imported using relative paths, but other packages (anything top-level in the `packages` dir) should still use the `@woocommerce/<name>` syntax.

Also, if you've added any new package imports, those need to be added to the local package.json (ex `packages/components/package.json`). I don't think you have, but for future work :)

packages/components/src/filters/advanced/number-filter.js
```diff
@@ -0,0 +1,184 @@
...
+			if ( 'currency' === inputType ) {
+				moreThan = formatCurrency( moreThan );
+				lessThan = formatCurrency( lessThan );
+			}
+
+			filterStr = interpolateComponents( {
+				mixedString: this.getBetweenString(),
+				components: {
+					moreThan: <Fragment>{ moreThan }</Fragment>,
+					lessThan: <Fragment>{ lessThan }</Fragment>,
```
ryelle: Do these need the `Fragment` wrapper?

packages/components/src/filters/advanced/number-filter.js
jeffstieler: 🤔 

As long as they aren't equal, we can form a range out of the values - so long as "_ is between 10 and 1" isn't confusing. Allowing that would still need an error case for the values being equal. If we do allow the greater value in the first input, that could simplify the `.split()` logic when changing filter rules (just always take the first value when changing from "between").

I looked and didn't see any project-wide `.is-error` styling, just some one-off uses in the calendar, pagination, and summary. Can we add generic component error styling or does it all need to be done on a per-component basis?

packages/components/src/filters/advanced/number-filter.js
```diff
@@ -0,0 +1,184 @@
...
+			} );
+		} else if ( 'currency' === inputType ) { // need to format a single input value as currency
+			filterStr = formatCurrency( filterStr );
+		}
+
+		return interpolateComponents( {
+			mixedString: config.labels.title,
+			components: {
+				filter: <span>{ filterStr }</span>,
+				rule: <span>{ rule.label }</span>,
```
ryelle: When `filterStr` is empty, this leads to a strange screen reader label for the fieldset– "Item Quantity is Between and", for example. Can you add a new label for the legend so that it reads "Item quantity is between two numbers" or even just always use "Item quantity limits"?

packages/components/src/filters/advanced/number-filter.js
```diff
@@ -0,0 +1,184 @@
...
+			return (
+				0 === symbolPosition.indexOf( 'right' )
+				? <TextControlWithAffixes
+					suffix={ <span dangerouslySetInnerHTML={ { __html: currencySymbol } } /> }
+					className="woocommerce-filters-advanced__input-numeric-range"
+					type="number"
+					value={ value }
+					onChange={ onChange }
+				/>
+				: <TextControlWithAffixes
```
ryelle: All 3 of these text controls need labels– You should be able to use the `aria-label` attribute to pass a label directly to the input, which won't display to sighted users.

jeffstieler:
> This may not be an issue outside of devdocs, but should we add some styling to change the input widths? I get a bit of overlapping use the 'between' rule:

Fixed in a3d8e1b.

packages/components/src/filters/advanced/number-filter.js
jeffstieler: I believe so - without passing a node, the `mixedString` isn't interpolated. See: https://github.com/Automattic/interpolate-components/blob/master/src/index.es6#L52

packages/components/src/filters/advanced/number-filter.js
ryelle: So I actually see that we're doing this in all the filters, automatically updating the legend with the completed filter, which means all of them sound strange when the filter is empty.

I think the best solution would be to update all filters (in another PR), so that each legend is just the title, and there can be a hidden screen reader string that displays the "built" sentence– so the output HTML would look something like this:

```html
<fieldset>
	<legend>Item Quantity</legend>
	<div className="woocommerce-filters-advanced__fieldset">
		<span aria-hidden="true">Item Quantity is</span>
		<select aria-label="Item quantity filter match" />
		<input aria-label="Item quantity minimum amount" />
		<span aria-hidden="true">and</span>
		<input aria-label="Item quantity maximum amount" />
	</div>
	<span class="screen reader text">Item quantity is between 0 and 10</span>
</fieldset>
```

And that last span would only exist if the filter values are set.

packages/components/src/filters/advanced/number-filter.js
jeffstieler: Logic got cleaned up a bit in fa6474f.

packages/components/src/filters/advanced/number-filter.js
jeffstieler: Discussed this with @timmyc and we decided to allow non-positional range values for the "between" rule. (E.g. "between 200 and 100")

Punting on the error handling for now (which would only be checking that range start != range end).

timmyc (approved):
Looks like almost all of the items brought up in prior review were covered, or will be addressed in follow-up PRs. This is testing out quite nicely for me, so I say we get this one in.

packages/components/src/filters/advanced/number-filter.js
jeffstieler: Opened an in-progress PR to solicit feedback on my approach: https://github.com/woocommerce/wc-admin/pull/1099

I just modified the `<legend>` because the `<span>` in addition to the `<legend>` sounded confusing in testing (though I'm still getting used to how VoiceOver reads things).

packages/components/src/filters/advanced/index.js
psealock: @jeffstieler @joshuatf Good question.

It has to do with column widths. The designs include a specified width for columns 1, 2 and 4 (the X to remove filter). Column 3, in the case below takes up the remaining space:

screen shot 2018-09-25 at 4 29 02 pm

When dealing with non-english sentences, order can't be guaranteed. So when `isEnglish === false`, the layout changes such that Columns 1 and 2 have no defined width:

screen shot 2018-09-25 at 4 27 28 pm

More info here https://github.com/woocommerce/wc-admin/pull/457

PR #1092, opened 2018-12-14
Add customer email filter

### LINKED ISSUE #1031: Customers: Email Advanced Filter
Child task of #175 

image

Create an Advanced Filter config and associated [autocompleter](https://github.com/woocommerce/wc-admin/tree/master/packages/components/src/search/autocompleters) for customer emails

emdashcodes:
Fixes #1031.

This PR adds a customer email filter to the customer report. There is an API change to allow the `?email=` query string option to accept partial matches instead of a strict email.

Also note that it uses the current `/v3/customers` endpoint, so this will only be registered customer accounts. Once we have an endpoint that supports guests, which should supersede this one -- see p7bje6-1dM-p2, this should support all customers.

### Screenshots

screen shot 2018-12-14 at 10 28 11 am

* Open the customers report
* Add an email filter
* Test the autocompleter, make sure emails show up correctly, and that you can select emails.
* Hit filter, and make sure the query string updates
* Refresh, and make sure the previously selected emails show up

### DIFF
--- a/includes/api/class-wc-admin-rest-customers-controller.php
+++ b/includes/api/class-wc-admin-rest-customers-controller.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * REST API Customers Controller
+ *
+ * Handles requests to /customers/*
+ *
+ * @package WooCommerce Admin/API
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Customers controller.
+ *
+ * @package WooCommerce Admin/API
+ * @extends WC_REST_Customers_Controller
+ */
+class WC_Admin_REST_Customers_Controller extends WC_REST_Customers_Controller {
+
+	// TODO Add support for guests here. See https://wp.me/p7bje6-1dM.
+
+	/**
+	 * Searches emails by partial search instead of a strict match.
+	 * See "search parameters" under https://codex.wordpress.org/Class_Reference/WP_User_Query.
+	 *
+	 * @param array $prepared_args Prepared search filter args from the customer endpoint.
+	 * @param array $request Request/query arguments.
+	 * @return array
+	 */
+	public static function update_search_filters( $prepared_args, $request ) {
+		if ( ! empty( $prepared_args['search_columns'] ) && in_array( 'user_email', $prepared_args['search_columns'] ) && empty( $request['name'] ) ) {
+			$prepared_args['search'] = '*' . $prepared_args['search'] . '*';
+		}
+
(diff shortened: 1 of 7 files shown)


includes/api/class-wc-admin-rest-customers-controller.php
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	/**
+	 * Get the query params for collections.
+	 *
+	 * @return array
+	 */
+	public function get_collection_params() {
+		$params = parent::get_collection_params();
+		// Allow partial email matches.
```
jeffstieler: Optional: expand on this comment to explain that the partial match is being allowed by way of removing the email validation from the parameter.

packages/components/src/search/autocompleters/emails.js
```diff
@@ -0,0 +1,62 @@
...
+			</span>,
+		];
+	},
+	// This is slightly different than gutenberg/Autocomplete, we don't support different methods
+	// of replace/insertion, so we can just return the value.
+	getOptionCompletion( customer ) {
+		const value = {
+			id: customer.id,
+			label: customer.email,
+		};
```
jeffstieler: Nitpick: superfluous variable.

jeffstieler (approved):
Works as advertised - just a few small optional nitpicks and a question. Ready to 🛳 as is though.

includes/api/class-wc-admin-rest-customers-controller.php
```diff
@@ -0,0 +1,51 @@
...
+	/**
+	 * Searches emails by partial search instead of a strict match.
+	 * See "search parameters" under https://codex.wordpress.org/Class_Reference/WP_User_Query.
+	 *
+	 * @param array $prepared_args Prepared search filter args from the customer endpoint.
+	 * @param array $request Request/query arguments.
+	 * @return array
+	 */
+	public static function update_search_filters( $prepared_args, $request ) {
+		if ( ! empty( $prepared_args['search_columns'] ) && in_array( 'user_email', $prepared_args['search_columns'] ) && empty( $request['name'] ) ) {
```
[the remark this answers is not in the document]
justinshreve: Left over from some testing I was doing with the name filter -- thank you for catching it. And yeah, we could look at `email` instead. I'll update that.

includes/api/class-wc-admin-rest-customers-controller.php
justinshreve: Done!

PR #1093, opened 2018-12-14
Rename 'totalCount' to 'totalResults'

Aljullu:
Fixes pagination in _Products_ table.

We were using inconsistently `totalCount` and `totalResults` in different parts of the codebase. This PR tries to unify that to `totalResults`.

- Verify the table has pagination.
- Verify it stills works in the other reports.


jeffstieler (approved):
:shipit:

PR #1095, opened 2018-12-14
Fix dataEndpoints notice

emdashcodes:
See p1544812123280800-wc-admin-slack.

- Check debug/error log and see no notice.
- Optionally spot check the country dropdown on the customer report.

### DIFF
--- a/lib/client-assets.php
+++ b/lib/client-assets.php
@@ -178,6 +178,7 @@ function wc_admin_print_script_settings() {
 		'stockStatuses'    => wc_get_product_stock_status_options(),
 		'siteTitle'        => get_bloginfo( 'name' ),
 		'trackingEnabled'  => $tracking_enabled,
+		'dataEndpoints'    => array(),
 	);
 
 	foreach ( $preload_data_endpoints as $key => $endpoint ) {


timmyc:
I'm still seeing the warning output with this branch:

`<b>Notice</b>:  Undefined index: /wc/v3/data/countries in <b>/srv/www/example/public_html/wp-content/plugins/wc-admin/lib/client-assets.php</b> on line <b>185</b><br />`

but adding in a check for the key like so clears it up:

```
foreach ( $preload_data_endpoints as $key => $endpoint ) {
		if ( ! isset($settings['dataEndpoints'][ $key ] ) ) {
			continue;
		}
		$settings['dataEndpoints'][ $key ] = $preload_data[ $endpoint ]['body'];
	}
```

jeffstieler:
I can't seem to get the warning to show up in the first place 🤷‍♂️.

> but adding in a check for the key like so clears it up

That looks like it will prevent the `dataEndpoints` array from being populated.

justinshreve:
@timmyc Does the country autocompleter make suggestions for you? Seems like `$preload_data` is not getting set.

* What version(s)/branches of WP and WC are you using? (And Gutenberg if applicable).
* Can you do a `error_log( print_r( $preload_data, 1 ) );` after the `array_reduce` on line 156? Does this give you an error instead of API data?

timmyc (approved):
Updated my Woo and G-berg versions and all is well now.

PR #1099, opened 2018-12-15
Rework advanced filters screen reader text generation

jeffstieler:
Based on conversation here: https://github.com/woocommerce/wc-admin/pull/1089#discussion_r241843856

**This PR seeks to:**
* Only describe full filter if all values are set
* Describe each input more precisely

**Notes:**

There is an opportunity to DRY up the markup in the render of `NumberFilter`, `SearchFilter`, and `SelectFilter`, but I've left that undone as it might be a premature optimization. I can certainly turn that into another component if we decide that's cleanup worth doing.

Also, this PR adds a utility method called `textContent()`. It's essentially [`react-addons-text-content`](https://github.com/rwu823/react-addons-text-content/blob/master/src/index.js) - I made the decision to just copy pasta the code into our project since the dependency is rather small and hasn't been updated in 3 years.

**To test:**
* Navigate to devdocs, filters component
* Add a filter of each type (order status, products, item quantity)
* Verify that the `<legend>` for each doesn't change when interacting the the component
* Verify that a `<div>` with class `screen-reader-text` is rendered only when all of the filter inputs have values

### DIFF
--- a/packages/components/src/filters/advanced/number-filter.js
+++ b/packages/components/src/filters/advanced/number-filter.js
@@ -13,8 +13,12 @@ import { sprintf, __, _x } from '@wordpress/i18n';
  * Internal dependencies
  */
 import TextControlWithAffixes from '../../text-control-with-affixes';
-import { formatCurrency } from '@woocommerce/currency';
 import { textContent } from './utils';
+
+/**
+ * WooCommerce dependencies
+ */
+import { formatCurrency } from '@woocommerce/currency';
 
 class NumberFilter extends Component {
 	getBetweenString() {
--- a/packages/components/src/filters/advanced/number-filter.js
+++ b/packages/components/src/filters/advanced/number-filter.js
@@ -7,7 +7,7 @@ import { SelectControl, TextControl } from '@wordpress/components';
 import { get, find, partial } from 'lodash';
 import interpolateComponents from 'interpolate-components';
 import classnames from 'classnames';
-import { _x } from '@wordpress/i18n';
+import { sprintf, __, _x } from '@wordpress/i18n';
 
 /**
  * Internal dependencies
@@ -28,6 +28,14 @@ class NumberFilter extends Component {
 		const inputType = get( config, [ 'input', 'type' ], 'number' );
 		const rule = find( config.rules, { value: filter.rule } ) || {};
 		let [ rangeStart, rangeEnd ] = ( filter.value || '' ).split( ',' );
+
+		// Return the filter label if we're missing input(s)
+		if (
+			! rangeStart ||
+			( 'between' === rule.value && ! rangeEnd )
+		) {
+			return get( config, [ 'labels', 'add' ] );
+		}
 
 		if ( 'currency' === inputType ) {
 			rangeStart = formatCurrency( rangeStart );
@@ -50,13 +58,14 @@ class NumberFilter extends Component {
 		return interpolateComponents( {
 			mixedString: config.labels.title,
 			components: {
+				ariaHide: <span />,
 				filter: <span>{ filterStr }</span>,
 				rule: <span>{ rule.label }</span>,
 			},
 		} );
 	}
 
-	getFormControl( type, value, onChange ) {
+	getFormControl( { type, value, label, onChange } ) {
 		if ( 'currency' === type ) {
 			const currencySymbol = get( wcSettings, [ 'currency', 'symbol' ] );
 			const symbolPosition = get( wcSettings, [ 'currency', 'position' ] );
@@ -68,6 +77,7 @@ class NumberFilter extends Component {
 					className="woocommerce-filters-advanced__input"
 					type="number"
 					value={ value || '' }
+					aria-label={ label }
 					onChange={ onChange }
 				/>
 				: <TextControlWithAffixes
@@ -75,6 +85,7 @@ class NumberFilter extends Component {
 					className="woocommerce-filters-advanced__input"
 					type="number"
 					value={ value || '' }
+					aria-label={ label }
 					onChange={ onChange }
 				/>
 			);
@@ -85,6 +96,7 @@ class NumberFilter extends Component {
 				className="woocommerce-filters-advanced__input"
 				type="number"
 				value={ value || '' }
+				aria-label={ label }
 				onChange={ onChange }
 			/>
 		);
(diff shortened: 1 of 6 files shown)


packages/components/src/filters/advanced/test/utils.js
```diff
@@ -0,0 +1,83 @@
+/**
+ * Internal dependencies
+ */
+import { textContent } from '../utils';
+
+describe( 'textContent()', () => {
+	test( 'should be got text `Hello World`', () => {
```
psealock: `should be got text` => `should get text`

packages/components/src/filters/advanced/number-filter.js
```diff
@@ -68,13 +81,15 @@ class NumberFilter extends Component {
...
+					aria-label={ label }
 					onChange={ onChange }
 				/>
 				: <TextControlWithAffixes
 					prefix={ <span dangerouslySetInnerHTML={ { __html: currencySymbol } } /> }
 					className="woocommerce-filters-advanced__input"
 					type="number"
 					value={ value }
+					aria-label={ label }
 					onChange={ onChange }
```
psealock: Looks like you can add a prop `min="0"` to prevent negative numbers.

packages/components/src/filters/advanced/number-filter.js
```diff
@@ -178,6 +230,11 @@ class NumberFilter extends Component {
 				>
 					{ children }
 				</div>
+				{ screenReaderText && (
+					<span className="screen-reader-text">
+						{ screenReaderText }
+					</span>
+				) }
```
psealock: Using Chrome and VoiceOver, this `screenReaderText` never gets read. I see it appropriately added to the DOM when the sentence is complete, however.

Sorry if this has been previously discussed, but would it not make sense to add `screenReaderText` to the legend?

psealock (commented):
Looking good @jeffstieler !

I noted a couple small things and a comment on unread `screenReaderText`.

packages/components/src/filters/advanced/number-filter.js
ryelle: Is it not read even when you move to it?

It's not meant to be automatically read, as that interrupts the content entry - having it in legend is too noisy, and why @jeffstieler is doing all this work 😁

ryelle (commented):
Tested with VoiceOver + Safari, and it's working as expected. I'm not exactly sure why we need `textContent`, though, can you explain that a little more?

packages/components/src/filters/advanced/number-filter.js
jeffstieler: It can't be tabbed to, but if you use the VO controls you'll navigate to it after the last filter input, before the remove button.

jeffstieler:
> I'm not exactly sure why we need textContent, though, can you explain that a little more?

I initially introduced that so we'd be passing a string to `wp.a11y.speak()`. Let me double check to see if that's still needed..

jeffstieler:
> > I'm not exactly sure why we need textContent, though, can you explain that a little more?
> 
> I initially introduced that so we'd be passing a string to wp.a11y.speak(). Let me double check to see if that's still needed..

In testing, removing `textContent()` results in the following markup:

```html
<span class="screen-reader-text">Item Quantity is <span>Less Than</span> <span>1</span></span>
```

Which VoiceOver requires several interactions to completely read - 3 in total for this example. Using `textContent()` to extract only the text content results in VO reading the full sentence ("Item Quantity is Less Than 1") once the component is navigated to.

packages/components/src/filters/advanced/test/utils.js
jeffstieler: Heh, copy pasta'd from https://github.com/rwu823/react-addons-text-content/blob/master/test/react-addons-text-content.spec.js#L5.

I'll fix it.

packages/components/src/filters/advanced/number-filter.js
jeffstieler: Let's handle this in a separate PR - the `NumberFilter` component doesn't have any sort of input validation yet.

packages/components/src/filters/advanced/test/utils.js
jeffstieler: Fixed in 9ce4d9e.

packages/components/src/filters/advanced/number-filter.js
psealock: >but if you use the VO controls you'll navigate to it after the last filter input

via tabbing? Or is there another method?

psealock: Ok, I figured out how to navigate to the text using VO controls. I'm not a VO power user (yet), so  forgive my novice comment, but I'm not experiencing the input being too noisy on master -- the legend only gets read when the fieldset has focus. 

I assumed VO users would tab their way through a form, instead of using the arrow controls to read text in a span, and by doing so skip right over the text being added in this PR. Is this assumption wrong?

jeffstieler: > I assumed VO users would tab their way through a form, instead of using the arrow controls to read text in a span, and by doing so skip right over the text being added in this PR. Is this assumption wrong?

I'm not knowledgeable enough to say - I'm new to a11y 🙂. I defer to @ryelle.

ryelle: > Is this assumption wrong?

It's over-simplifying, not exactly wrong – screen reader users will tab to move through interactive components (forms, buttons, etc), but they'll use the arrow nav to move through the page's full contents (otherwise they'd never read non-interactive page content). They might also use other navigation methods (skipping through headings, or a list of links, or a list of buttons, etc).

ryelle: That said, I don't know what a real user would do in this scenario - but this approach is better than interrupting the user while they're typing.

jeffstieler:
Bumping for a resolution to the `textContent()` question ([#](https://github.com/woocommerce/wc-admin/pull/1099#issuecomment-448404382)). _cc: @ryelle_

packages/components/src/filters/advanced/number-filter.js
psealock: Cool, thank you for the clarification.

jeffstieler:
@ryelle it seems that VO (even in Safari) doesn't cope well with line breaks in the HTML. In my testing (on this PR) using an outer `<div>` with inner `<span>`s still takes multiple key presses to read the full phrase of "Item Quantity is Less Than 1".

Here's a screen capture: https://cloudup.com/c_Ne8_uyoj3 (you can see the selected VO element, the VO overlay screen, and the markup of the `screen-reader-text` div all in frame)

I'll look in/around the `interpolateComponents()` to see if there's a culprit for the line breaks.

jeffstieler:
Perhaps then I'm not navigating correctly? I'm using `Control + Option + Right arrow` to move between readable elements.

Pushed up my changes - 2532539.

ryelle:
I figured it out – the wp core CSS for `.screen-reader-text` also puts styles on `.screen-reader-text span`, and that is trigger VO to read the spans as individual items 🙃 In my testing, I was just creating plain HTML pages with no CSS, to test the markup… not thinking that the screen-reader-text class did that. Okay, to fix, you can use the class `ui-helper-hidden-accessible` instead of `screen-reader-text`, which doesn't do the child span thing.

jeffstieler:
@ryelle _(Following up on a Slack convo we had)_

Changing to the `ui-helper-hidden-accessible` didn't solve the problem - apparently WP's `line-height` ([#](https://github.com/WordPress/WordPress/blob/master/wp-admin/css/common.css#L216)) is causing the problem.

Even while resetting the `line-height` to either `initial` or `normal` seems to fix the Safari VO issue, the text of the `span`s are all concatenated, ignoring padding whitespace. This results in the screen reader reading `Item Quantity is Between$4.00and$6.00` which causes it to read "between dollar sign four point zero zero ..." instead of "between four dollars and zero zero cents ...".

I don't believe any of these issues were present with the `textContent()` utility function I introduced. I'm going to go back to that logic and test these scenarios. If they are no longer problems, I think we should move forward with that solution.

ryelle (approved):
This is reading correctly now, no pauses/broken sentences. Thanks for all your work on this!

PR #1101, opened 2018-12-15
Rename 'isError' methods to 'getError' when they return an object instead of a boolean

Aljullu:
Follow-up of the discussion in https://github.com/woocommerce/wc-admin/pull/1075#discussion_r241866256.

We had many selectors named `isGetOrdersError`, `isReportItemsError`, etc. that were returning an error object instead of a boolean. This PR renames them to `getOrdersError`, `getReportItemsError`...

A follow-up discussion would be if we want to display those errors in the UI in some way or make the UI react to different kind of errors in different ways. But for now, this PR keeps the same behavior as before and only uses the error objects to check if an error occurred.

This PR modifies all error selectors so the best way to test would be faking an error in each one of them.

For example, modifying:
```ES6
const url = `${ NAMESPACE }/orders${ stringifyQuery( query ) }`;
```
to:
```ES6
const url = `${ NAMESPACE }/lorem-ipsum${ stringifyQuery( query ) }`;
```
in `client/wc-api/orders/operations.js` and then verifying the error is correctly shown in the _Orders_ report.

Also, verifying that when the URL is the correct one, no error is displayed in the UI.

### DIFF
--- a/client/analytics/report/orders/table.js
+++ b/client/analytics/report/orders/table.js
@@ -262,7 +262,7 @@ export default compose(
 		const datesFromQuery = getCurrentDates( query );
 		const filterQuery = getFilterQuery( 'orders', query );
 
-		const { getOrders, getOrdersTotalCount, isGetOrdersError, isGetOrdersRequesting } = select(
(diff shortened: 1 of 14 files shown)


psealock (approved):
Testing well, I can force an error and unit tests passing.

PR #1105, opened 2018-12-17
Change product_gross_revenue to product_net_revenue

### LINKED ISSUE #1072: Column product_gross_revenue should be named product_net_revenue
**Describe the bug**
As @LevinMedia pointed out, the `product_gross_revenue` field in the `wc_order_product_lookup` table should be renamed to `product_net_revenue` since we're saving the revenue without shipping and tax included.

**To Reproduce**
Check the database table `wc_order_product_lookup` and note the `product_gross_revenue` column name.

**Expected behavior**
Expected the column to be named `product_net_revenue`.

joshuatf:
Fixes #1072 

Updates the products table and column/property naming to reflect actual values that are net and not gross.

### Screenshots
screen shot 2018-12-17 at 6 11 02 pm

1. You may need to delete your `wc_order_product_lookup` table or manually modify the `product_gross_revenue` column to be `product_net_revenue` to test this.
2. Open the Product report page.
3.  Check that "Gross Revenue" has been replaced by "Net Revenue."
4.  Check that the Revenue report page has not been affected since it contains both gross and net revenue reporting.


psealock (approved):
Changes look good to me works as expected. Unit tests pass locally, although one failed on CI. 

Potentially silly question: The design (p6riRB-3dP-p2) still has "Gross Revenue". Is renaming `gross` -> `net` the correct change or should the calculation just be changed to so the column excludes shipping and tax?

joshuatf:
>Potentially silly question: The design (p6riRB-3dP-p2) still has "Gross Revenue". Is renaming gross -> net the correct change or should the calculation just be changed to so the column excludes shipping and tax?

Not a silly question at all.  I spoke with @LevinMedia about this and he was the one that actually caught the issue and recommended changing to `net`.  I think that `net` probably makes more sense in the context of this table since tax and shipping probably wouldn't be factors in the business decisions for individual products (not to mention trying to determine how to split shipping across products opens another can of worms entirely).

LevinMedia:
@psealock @joshuatf  I concur... I just haven't gotten around to updating the designs on P2 yet. 

Noting that we'll  to make this same update on the product detail report, and category report.

joshuatf:
@psealock I updated this PR to include changes in the category report as well.  The Net Revenue won't appear correctly on the category page until we get #996 merged in though.

Could you give it another quick once over?

psealock (approved):
Cool, thanks for the confirmation. #996 is merged

PR #1109, opened 2018-12-17
Add fix for npm install https issue

coderkevin:
We've been running afoul of a known npm issue that has existed for over
6 months, where our package-lock.json file switches between http: and
https: on several dependencies:

https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285/17

Since npm itself always uses https as noted in the ticket above, this
change forces all dependencies in the package-lock.json file to be
https, therefore eliminating the churn on our package-lock.json files
each time we run `npm install`.

- `npm install`
- Observe that no http/https changes have occurred in `package-lock.json`
- `npm run build`, `npm start`, and ensure all dependencies are met and loading correctly.


Aljullu (approved):
This is testing well for me. Hopefully we will get much less diffs in `package-lock.json` now. Thanks for working on this!

PR #1114, opened 2018-12-18
Use <Link> instead of <a> for Orders table link

Aljullu:
Tiny PR that replaces an `<a>` link with `<Link>`.

### Screenshots
image

- Go to the _Orders_ report and click on an Order ID link.
- Verify the link still redirects to something like `/wp-admin/post.php?post=461&action=edit`.


emdashcodes (approved):
LGTM. 🚢 it.

Issue #1115, opened 2018-12-18
Number filter values should be set in state


joshuatf:
**Describe the bug**
Values for inputs should be set in state.  Currently this error is shown when a number filter is used:
`A component is changing an uncontrolled input of type number to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component.`

**To Reproduce**
1.  Visit the devdocs.
2. Add the subtotal filter under the "Advanced Filters."
3. Add and remove a value in the input field.
4. Note the error in the console.

**Expected behavior**
Expected no error to occur.

**Screenshots**
<img width="623" alt="screen shot 2018-12-18 at 6 09 10 pm" src="https://user-images.githubusercontent.com/10561050/50147156-63b93700-02f0-11e9-8521-14f95f2c55ed.png">

jeffstieler:
I can't seem to reproduce this - I've tried Chrome, Safari, and Firefox. Is there any additional information you can provide, or perhaps can someone else reproduce this?

PR #1116, opened 2018-12-18
Add customer numeric filters

### LINKED ISSUE #1034: Customers: Add Numeric Advanced Filters
Child task of #175 
Blocked by #1028 

When the Numeric Advanced filter is available, this task involves adding the number specific advanced filter configs to the Customers report:

image

REST API arg names:

image

joshuatf:
Fixes #1034 

Adds in the AOV, Total Spend, and Order Count filters to the customers report.

Testing is blocked by #1115

### Screenshots
screen shot 2018-12-18 at 6 04 42 pm

1. Visit `wp-admin/admin.php?page=wc-admin#/analytics/customers`.
2. Add the 3 filters (AOV/Total Spend/No. of Orders).
3. Add amounts for less than, more than, and between.
4.  Check that the respective params are added to the URL on filter.

### DIFF
--- a/client/analytics/report/customers/config.js
+++ b/client/analytics/report/customers/config.js
@@ -166,6 +166,81 @@ export const advancedFilters = {
 				} ) ),
 			},
 		},
+		order_count: {
+			labels: {
+				add: 'No. of Orders',
+				remove: 'Remove order  filter',
+				rule: 'Select an order count filter match',
+				title: 'No. of Orders {{rule /}} {{filter /}}',
+			},
+			rules: [
+				{
+					value: 'max',
+					label: 'Less Than',
+				},
+				{
+					value: 'min',
+					label: 'More Than',
+				},
+				{
+					value: 'between',
+					label: 'Between',
+				},
+			],
+			input: {
+				component: 'Number',
+			},
+		},
+		total_spend: {
+			labels: {
+				add: 'Total Spend',
+				remove: 'Remove total spend filter',
+				rule: 'Select a total spend filter match',
+				title: 'Total Spend {{rule /}} {{filter /}}',
+			},
+			rules: [
+				{
+					value: 'max',
+					label: 'Less Than',
+				},
+				{
+					value: 'min',
+					label: 'More Than',
+				},
+				{
+					value: 'between',
+					label: 'Between',
+				},
+			],
+			input: {
+				component: 'Number',
+			},
+		},
+		avg_order_value: {
+			labels: {
+				add: 'AOV',
+				remove: 'Remove average older value filter',
+				rule: 'Select an average order value filter match',
+				title: 'AOV {{rule /}} {{filter /}}',
+			},
+			rules: [
+				{
+					value: 'max',
+					label: 'Less Than',
+				},
+				{
+					value: 'min',
+					label: 'More Than',
+				},
+				{
+					value: 'between',
+					label: 'Between',
+				},
+			],
+			input: {
+				component: 'Number',
+			},
+		},
 	},
 };
 /*eslint-enable max-len*/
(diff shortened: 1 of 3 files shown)


client/analytics/report/customers/config.js
```diff
@@ -136,6 +136,81 @@ export const advancedFilters = {
...
+		order_count: {
+			labels: {
+				add: 'No. of Orders',
+				remove: 'Remove order  filter',
+				rule: 'Select an order count filter match',
+				title: 'No. of Orders {{rule /}} {{filter /}}',
+			},
+			rules: [
+				{
+					value: 'max',
```
jeffstieler: The `NumberFilter` component expects one of: `lessthan`, `morethan`, `between`.

joshuatf: Thanks for testing @jeffstieler!  I changed these to `min` and `max` to try and match the existing controller params as closely as possible.  Is there something that limits to those values?  

In testing, it seems to work okay with different values for `lessthan` and `morethan` with the exception of `between` that's needed to split those fields into 2.

client/analytics/report/customers/config.js
```diff
@@ -167,6 +167,81 @@ export const advancedFilters = {
 				} ) ),
 			},
 		},
+		order_count: {
+			labels: {
+				add: 'No. of Orders',
```
psealock: You'll need translations for all these strings

client/analytics/report/customers/config.js
```diff
@@ -167,6 +167,81 @@ export const advancedFilters = {
...
+					label: 'Between',
+				},
+			],
+			input: {
+				component: 'Number',
+			},
+		},
+		avg_order_value: {
+			labels: {
+				add: 'AOV',
```
psealock: I didn't check the designs, but I thought we were moving away from "AOV" in favor of something more readable.

client/analytics/report/customers/config.js
```diff
@@ -167,6 +167,81 @@ export const advancedFilters = {
...
+					value: 'min',
+					label: 'More Than',
+				},
+				{
+					value: 'between',
+					label: 'Between',
+				},
+			],
+			input: {
+				component: 'Number',
```
psealock: This will point to `<NumberFilter />`, which already has currency built in if the `type` prop is set to `currency`. Maybe we can support an input type, `component: 'Currency'`, [here](https://github.com/woocommerce/wc-admin/blob/2a729477b81758106d288a1e7352e3236e6ef72d/packages/components/src/filters/advanced/index.js#L190) which uses `<NumberFilter />` with a currency prop.

psealock (commented):
Working as advertised. Just a couple comments about strings and using our new currency input capabilities.

client/analytics/report/customers/config.js
joshuatf: 🤦‍♂️ Thanks for catching!  Added.

client/analytics/report/customers/config.js
joshuatf: It looks like "AOV" is being used in designs unless I'm looking at outdated ones.  Do you have a link to any discussion on this?  /cc @LevinMedia

client/analytics/report/customers/config.js
joshuatf: Good idea.  Added this in https://github.com/woocommerce/wc-admin/pull/1116/commits/caed8340592e3152cc15184d5498a2a9e1d47faa

coderkevin:
Is this ready to review again?

joshuatf:
@coderkevin Ready for review!

psealock (approved):
Changes LGTM, testing well

PR #1120, opened 2018-12-18
Update npm versions and build script

timmyc:
Per some feedback from [Ben in slack](https://a8c.slack.com/archives/C9K5T0XSP/p1544822558289800) - this changes our build script to use `lerna publish from-package` to get around the issues of pushing tags to `master`. On this branch I ran `lerna version` which did the tag creation and the package.json updates below.

I still need to run `publish:prod` but guessing that should be done from `master` once this branch is merged.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "test:update-snapshots": "jest --updateSnapshot --config tests/js/jest.config.json",
     "docs": "node ./bin/generate-docs",
     "publish:check": "npm run build:packages && lerna updated",
-    "publish:dev": "npm run build:packages && lerna publish --npm-tag next",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -42,7 +42,7 @@
     "docs": "node ./bin/generate-docs",
     "publish:check": "npm run build:packages && lerna updated",
     "publish:dev": "npm run build:packages && lerna publish --npm-tag next",
-    "publish:prod": "npm run build:packages && lerna publish"
+    "publish:prod": "npm run build:packages && lerna publish from-package"
```
ryelle: Do we need to add this to `publish:dev` too?

timmyc: Updated, thanks!

Issue #428, opened 2018-09-18
REST API: Downloads reports endpoint


claudiosanches:
Schema:

- https://app.swaggerhub.com/apis/peterfabian/wc-v3-api/1.0.0#/default/get_reports_downloads
- https://app.swaggerhub.com/apis/peterfabian/wc-v3-api/1.0.0#/default/get_reports_downloads_files
- https://app.swaggerhub.com/apis/peterfabian/wc-v3-api/1.0.0#/default/get_reports_downloads_stats

timmyc:
Few more details:

P2: p7bje6-15q-p2
swagger: [https://app.swaggerhub.com/apis/peterfabian/wc-v3-api/1.0.0#/default/get_reports_downloads_stats](get_reports_downloads_stats)

The [current download stats endpoint](https://github.com/woocommerce/wc-admin/blob/master/includes/api/class-wc-admin-rest-reports-downloads-stats-controller.php) is mostly placeholder logic.

This new endpoint is what will be queried to build out the chart on the Downloads Report page:

![image](https://user-images.githubusercontent.com/22080/48962482-631dc280-ef35-11e8-9108-08e93ade4a0b.png)

It is unclear to me if this endpoint requires a new reporting table to be built, or if it can use the existing tables that have already been implemented.

timmyc:
Also as I mentioned in #915 - the table design is calling for us to display the product name attached to the downloaded file, along with the username that downloaded the file. The current API response is return the associated object ID's for those two pieces of data. Would there be an option for us to request extended product data as we built in other endpoints so we don't need to query 3 endpoints to build out one row of data in the table?

PR #1122, opened 2018-12-18
Add `reports/downloads` REST API endpoint.

emdashcodes:
See #428. This implements one of the required endpoints needed for the downloads report. It implements https://app.swaggerhub.com/apis/peterfabian/wc-v3-api/1.0.0#/default/get_reports_downloads, in addition to the required filter options that were missing from the schema.

- Run `phpunit` and make sure all tests are passing.

Extra steps:
- Create two different test downloadable products. Make sure a file is attached. https://docs.woocommerce.com/document/digital-downloadable-product-handling/
- Place some orders, and purchase access to the download.
- Under your stores front end, go to "My Account", and download each file at least once.
- Use Postman or another tool to make some test requests to `/wc/v3/reports/downloads`. You can use various filters with your product, order, or user IDs to test the filters.
- Spot check the responses, spot check the headers for pagination info.

### DIFF
--- a/includes/api/class-wc-admin-rest-reports-downloads-controller.php
+++ b/includes/api/class-wc-admin-rest-reports-downloads-controller.php
@@ -291,6 +291,7 @@ class WC_Admin_REST_Reports_Downloads_Controller extends WC_REST_Reports_Control
 			),
 			'default'           => array(),
 			'sanitize_callback' => 'wp_parse_id_list',
+			'validate_callback' => 'rest_validate_request_arg',
 
 		);
 		$params['product_excludes'] = array(
@@ -300,6 +301,7 @@ class WC_Admin_REST_Reports_Downloads_Controller extends WC_REST_Reports_Control
 				'type' => 'integer',
 			),
 			'default'           => array(),
+			'validate_callback' => 'rest_validate_request_arg',
 			'sanitize_callback' => 'wp_parse_id_list',
 		);
 		$params['order_includes'] = array(
--- a/includes/api/class-wc-admin-rest-reports-downloads-controller.php
+++ b/includes/api/class-wc-admin-rest-reports-downloads-controller.php
@@ -30,4 +30,332 @@ class WC_Admin_REST_Reports_Downloads_Controller extends WC_REST_Reports_Control
 	 * @var string
 	 */
 	protected $rest_base = 'reports/downloads';
+
+	/**
+	 * Get items.
+	 *
+	 * @param WP_REST_Request $request Request data.
+	 *
+	 * @return array|WP_Error
+	 */
+	public function get_items( $request ) {
+		$args       = array();
+		$registered = array_keys( $this->get_collection_params() );
+		foreach ( $registered as $param_name ) {
+			if ( isset( $request[ $param_name ] ) ) {
+				$args[ $param_name ] = $request[ $param_name ];
+			}
+		}
+
+		$reports        = new WC_Admin_Reports_Downloads_Query( $args );
+		$downloads_data = $reports->get_data();
+
+		$data = array();
+
+		foreach ( $downloads_data->data as $download_data ) {
+			$item   = $this->prepare_item_for_response( $download_data, $request );
+			$data[] = $this->prepare_response_for_collection( $item );
+		}
+
+		$response = rest_ensure_response( $data );
+
+		$response->header( 'X-WP-Total', (int) $downloads_data->total );
+		$response->header( 'X-WP-TotalPages', (int) $downloads_data->pages );
+
+		$page      = $downloads_data->page_no;
+		$max_pages = $downloads_data->pages;
+		$base      = add_query_arg( $request->get_query_params(), rest_url( sprintf( '/%s/%s', $this->namespace, $this->rest_base ) ) );
+		if ( $page > 1 ) {
+			$prev_page = $page - 1;
+			if ( $prev_page > $max_pages ) {
+				$prev_page = $max_pages;
+			}
+			$prev_link = add_query_arg( 'page', $prev_page, $base );
+			$response->link_header( 'prev', $prev_link );
+		}
+		if ( $max_pages > $page ) {
+			$next_page = $page + 1;
+			$next_link = add_query_arg( 'page', $next_page, $base );
+			$response->link_header( 'next', $next_link );
+		}
+
+		return $response;
+	}
+
+	/**
+	 * Prepare a report object for serialization.
+	 *
+	 * @param Array           $report  Report data.
+	 * @param WP_REST_Request $request Request object.
+	 * @return WP_REST_Response
+	 */
+	public function prepare_item_for_response( $report, $request ) {
+		$data = $report;
+
+		$context = ! empty( $request['context'] ) ? $request['context'] : 'view';
+		$data    = $this->add_additional_fields_to_object( $data, $request );
+		$data    = $this->filter_response_by_context( $data, $context );
+
+		// Wrap the data in a response object.
+		$response = rest_ensure_response( $data );
+		$response->add_links( $this->prepare_links( $report ) );
+
+		$response->data['date'] = get_date_from_gmt( $data['date_gmt'], 'Y-m-d H:i:s' );
+
+		// Figure out file name.
+		// Matches https://github.com/woocommerce/woocommerce/blob/4be0018c092e617c5d2b8c46b800eb71ece9ddef/includes/class-wc-download-handler.php#L197.
+		$product_id                  = intval( $data['product_id'] );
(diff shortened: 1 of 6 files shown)


justinshreve:
> The current API response is return the associated object ID's for those two pieces of data. Would there be an option for us to request extended product data as we built in other endpoints so we don't need to query 3 endpoints to build out one row of data in the table?

@timmyc to Answer your question, I've used WP-APIs `_embed` system to allow optional inclusion of product/user info: https://developer.wordpress.org/rest-api/using-the-rest-api/global-parameters/#_embed

includes/api/class-wc-admin-rest-reports-downloads-controller.php
```diff
@@ -30,4 +30,332 @@ class WC_Admin_REST_Reports_Downloads_Controller extends WC_REST_Reports_Control
...
+			'validate_callback' => 'rest_validate_request_arg',
+		);
+		$params['product_includes'] = array(
+			'description'       => __( 'Limit result set to items that have the specified product(s) assigned.', 'wc-admin' ),
+			'type'              => 'array',
+			'items'             => array(
+				'type' => 'integer',
+			),
+			'default'           => array(),
+			'sanitize_callback' => 'wp_parse_id_list',
```
timmyc: I see `'validate_callback' => 'rest_validate_request_arg',` used on some of the other similar arguments like `order_includes`, should we have it here too?

includes/api/class-wc-admin-rest-reports-downloads-controller.php
```diff
@@ -30,4 +30,332 @@ class WC_Admin_REST_Reports_Downloads_Controller extends WC_REST_Reports_Control
...
+
+		);
+		$params['product_excludes'] = array(
+			'description'       => __( 'Limit result set to items that don\'t have the specified product(s) assigned.', 'wc-admin' ),
+			'type'              => 'array',
+			'items'             => array(
+				'type' => 'integer',
+			),
+			'default'           => array(),
+			'sanitize_callback' => 'wp_parse_id_list',
```
timmyc: Likewise here on the `validate_callback`?

includes/class-wc-admin-reports-downloads-query.php
```diff
@@ -0,0 +1,47 @@
...
+ * WC_Admin_Reports_Downloads_Query
+ */
+class WC_Admin_Reports_Downloads_Query extends WC_Admin_Reports_Query {
+
+	/**
+	 * Valid fields for downloads report.
+	 *
+	 * @return array
+	 */
+	protected function get_default_query_vars() {
```
timmyc: Is this method needed outside of being used in the filter in `get_data()`?

includes/class-wc-admin-reports-downloads-query.php
```diff
@@ -0,0 +1,47 @@
...
+		return array();
+	}
+
+	/**
+	 * Get downloads  data based on the current query vars.
```
timmyc: extra space between `downloads  data`

includes/data-stores/class-wc-admin-reports-data-store.php
```diff
@@ -634,6 +634,65 @@ protected function get_excluded_coupons( $query_args ) {
...
+	 * Returns comma separated ids of included orders, based on query arguments from the user.
+	 *
+	 * @param array $query_args Parameters supplied by the user.
+	 * @return string
+	 */
+	protected function get_included_orders( $query_args ) {
+		$included_orders_str = '';
+
+		if ( isset( $query_args['order_includes'] ) && is_array( $query_args['order_includes'] ) && count( $query_args['order_includes'] ) > 0 ) {
+			$included_orders_str = implode( ',', $query_args['order_includes'] );
```
timmyc: do we need to cast/ensure each value is an int here?

timmyc: Same comment on all of the below `get_` methods?

includes/data-stores/class-wc-admin-reports-downloads-data-store.php
```diff
@@ -0,0 +1,375 @@
...
+		global $wpdb;
+	}
+
+	/**
+	 * Updates the database query with parameters used for downloads report.
+	 *
+	 * @param array $query_args Query arguments supplied by the user.
+	 * @return array            Array of parameters used for SQL query.
+	 */
+	protected function get_sql_query_params( $query_args ) {
```
timmyc: Whew some pretty intensive query setup here, very legible and easy to follow though, nice work.

includes/data-stores/class-wc-admin-reports-downloads-data-store.php
```diff
@@ -0,0 +1,375 @@
...
+	 * Returns comma separated ids of included ip address, based on query arguments from the user.
+	 *
+	 * @param array $query_args Parameters supplied by the user.
+	 * @return string
+	 */
+	protected function get_included_ip_addresses( $query_args ) {
+		$included_ips_str = '';
+
+		if ( isset( $query_args['ip_address_includes'] ) && is_array( $query_args['ip_address_includes'] ) && count( $query_args['ip_address_includes'] ) > 0 ) {
+			$included_ips_str = implode( "','", $query_args['ip_address_includes'] );
```
timmyc: Do we need to sanitize the input here at all or are we assuming that is all handled in `rest_validate_request_arg`?

tests/api/reports-downloads.php
```diff
@@ -0,0 +1,401 @@
...
+/**
+ * Reports Downloads REST API Test
+ *
+ * @package WooCommerce Admin\Tests\API.
+ */
+
+/**
+ * WC_Tests_API_Reports_Downloads
+ */
+class WC_Tests_API_Reports_Downloads extends WC_REST_Unit_Test_Case {
```
timmyc: Excellent and thorough test cases here.

timmyc (commented):
Couple of quick questions - I tried to dig a bit into how we are sanitizing args on the REST API, but figured I'd ask too :)

includes/data-stores/class-wc-admin-reports-downloads-data-store.php
```diff
@@ -0,0 +1,375 @@
...
+	 * Mapping columns to data type to return correct response types.
+	 *
+	 * @var array
+	 */
+	protected $column_types = array(
+		'id'          => 'intval',
+		'date'        => 'strval',
+		'date_gmt'    => 'strval',
+		'download_id' => 'strval', // String because this can sometimes be a hash.
+		'file_name'   => 'strval',
```
joshuatf: Should some of these fall into the `extended_info` area?

joshuatf (commented):
Code looks really good and is testing well!  Awesome work @justinshreve 👍 
 
Small question on whether or not we plan to use `extended_info` for some of the response variables.

includes/api/class-wc-admin-rest-reports-downloads-controller.php
justinshreve: Yup, good catch. Resolved in https://github.com/woocommerce/wc-admin/pull/1122/commits/64306a404db16669ceaefea5ceb8f77f1051c5df.

includes/class-wc-admin-reports-downloads-query.php
justinshreve: I double checked, and yes it is called elsewhere in the chain it looks like and I get some failing tests trying to remove it.

includes/class-wc-admin-reports-downloads-query.php
justinshreve: Fixed.

includes/data-stores/class-wc-admin-reports-data-store.php
justinshreve: Correct - the ID ones get ran through `absint` from it: https://developer.wordpress.org/reference/functions/wp_parse_id_list/

includes/data-stores/class-wc-admin-reports-downloads-data-store.php
justinshreve: I've added escaping here to make sure, since that just validates the arg but does not sanitation.

includes/data-stores/class-wc-admin-reports-downloads-data-store.php
justinshreve: I don't think so. All of these fields are columns in the report table, so I think we will always want them present in the request.

justinshreve:
Thanks for the reviews @timmyc and @joshuatf! I've pushed a few changes and answered your questions above.

joshuatf (approved):
Changes look good- LGTM!

PR #1131, opened 2018-12-19
Add product categories wc-api and breadcrumbs

### LINKED ISSUE #1045: Categories and Products Report: Nested Categories
Categories and Products Report should include information about nested categories in tables.

**Design:** p6riRB-3dP-p2 (updated Dec 6)

Ideas on how to proceed from @timmyc from a comment in the design:

>Since Product Categories are mostly static, likely won’t change often – we could look at doing a client-side solution to this where we load in all category mappings ( ID => Name ) in our initial page load, then we can render items as you have designed.

>This solution would be quite a bit like what @ryelle33 implemented in the category block – or to fully make it like hers, we do a single request to the categories when the report is first loaded and request them all. Ideally then we could maybe share the logic she has developed in the block to do the exact same display in the report here ( and in the Categories report too ).

Category block references work done in https://href.li/?https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/189

joshuatf:
Fixes #1045

Adds in an API to retrieve a full list of product categories to then refer to and build breadcrumbs on the front-end.

### Screenshots
screen shot 2018-12-19 at 3 28 40 pm

1.  Make sure you've added categories and nested categories (at least 4 levels to test this PR).
2.  Visit the categories report `/wp-admin/admin.php?page=wc-admin#/analytics/categories`.
3.  Check that categories and nested categories correctly generate breadcrumbs.
4.  Visit the products report `/wp-admin/admin.php?page=wc-admin#/analytics/products`.
5.  Check that product category names are listed and that if more than 1 category exists, a tab is displayed that shows the remaining category breadcrumbs when clicked.

### DIFF
--- a/client/wc-api/categories/operations.js
+++ b/client/wc-api/categories/operations.js
@@ -1,0 +1,54 @@
+/** @format */
+
+/**
+ * External dependencies
+ */
+import apiFetch from '@wordpress/api-fetch';
+
+/**
+ * WooCommerce dependencies
+ */
+import { stringifyQuery } from '@woocommerce/navigation';
+
+/**
+ * Internal dependencies
+ */
+import { isResourcePrefix, getResourceIdentifier, getResourceName } from '../utils';
+
+function read( resourceNames, fetch = apiFetch ) {
+	const filteredNames = resourceNames.filter( name => isResourcePrefix( name, 'category-query' ) );
+
+	return filteredNames.map( async resourceName => {
+		const query = getResourceIdentifier( resourceName );
+		const url = `/wp/v2/product_cat${ stringifyQuery( query ) }`;
+
+		try {
+			const response = await fetch( {
+				parse: false,
+				path: url,
+			} );
+
+			const categories = await response.json();
+			const totalCount = parseInt( response.headers.get( 'x-wp-total' ) );
+			const ids = categories.map( category => category.id );
+			const categoryResources = categories.reduce( ( resources, category ) => {
+				resources[ getResourceName( 'category', category.id ) ] = { data: category };
+				return resources;
+			}, {} );
+
+			return {
+				[ resourceName ]: {
+					data: ids,
+					totalCount,
+				},
+				...categoryResources,
+			};
+		} catch ( error ) {
+			return { [ resourceName ]: { error } };
+		}
+	} );
+}
+
+export default {
+	read,
+};
--- a/client/wc-api/categories/operations.js
+++ b/client/wc-api/categories/operations.js
@@ -20,7 +20,7 @@ function read( resourceNames, fetch = apiFetch ) {
 
 	return filteredNames.map( async resourceName => {
 		const query = getResourceIdentifier( resourceName );
-		const url = `/wp/v2/product_cat${ stringifyQuery( query ) }`;
+		const url = `/wc/v3/products/categories${ stringifyQuery( query ) }`;
 
 		try {
 			const response = await fetch( {
--- a/client/wc-api/categories/selectors.js
+++ b/client/wc-api/categories/selectors.js
@@ -1,0 +1,59 @@
+/** @format */
+
+/**
+ * External dependencies
+ */
+import { isNil } from 'lodash';
+
+/**
+ * Internal dependencies
+ */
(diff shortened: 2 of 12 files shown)


client/wc-api/categories/operations.js
```diff
@@ -0,0 +1,54 @@
...
+ * Internal dependencies
+ */
+import { isResourcePrefix, getResourceIdentifier, getResourceName } from '../utils';
+
+function read( resourceNames, fetch = apiFetch ) {
+	const filteredNames = resourceNames.filter( name => isResourcePrefix( name, 'category-query' ) );
+
+	return filteredNames.map( async resourceName => {
+		const query = getResourceIdentifier( resourceName );
+		const url = `/wp/v2/product_cat${ stringifyQuery( query ) }`;
```
ryelle: Can you use [the `wc/v3/product/categories` endpoint](https://woocommerce.github.io/woocommerce-rest-api-docs/#product-categories) here, or does the core taxonomy endpoint give you something else that you need?

psealock:
@joshuatf I went ahead and deleted `extended_info` functions from `/reports/categories` endpoints and pushed onto this branch in https://github.com/woocommerce/wc-admin/pull/1131/commits/3b4e03ede33193c4965785ca8ebc290e0474bae2. Hope you don't mind.

This solves the first part of https://github.com/woocommerce/wc-admin/issues/1110 by removing `get_the_category_by_ID()`

client/wc-api/categories/operations.js
joshuatf: Good question @ryelle.  I initially used the core taxonomy endpoint because it allows us to override the collection param maximums so we can request more than 100 in the `per_page` param.

We can also do this with the WC API.  However, this doesn't remove the limitation and instead forces the `per_page` (currently set to `PHP_INT_MAX`) on every request to `wc/v3/product/categories`.

I implemented this in https://github.com/woocommerce/wc-admin/pull/1131/commits/32e5325f1bda9c620385fdd9882557896501a04d so you can take a look.  Which one do you think is a better approach?

joshuatf:
@psealock Perfect- thanks for the update!

This PR is ready for a review.  The category and tags follow the design where it appears the first category is displayed outside of the popover and also along with any remaining categories inside the popover.  @LevinMedia Does this look correct?

client/wc-api/categories/operations.js
ryelle: Are you trying to get all categories here? If you're using `apiFetch`, you can actually pass `per_page=-1` and it will work– behind the scenes [it fetches all items in batches of `per_page=100`,](https://github.com/WordPress/gutenberg/blob/master/packages/api-fetch/src/middlewares/fetch-all-middleware.js) and only returns to you when it's pulled all results. this is generally better on the server too, less chance of the API timing out.

LevinMedia:
Looking good @joshuatf, thanks!

client/wc-api/categories/selectors.js
```diff
@@ -0,0 +1,59 @@
...
+	);
+	return categories;
+};
+
+const getCategoriesTotalCount = ( getResource, requireResource ) => (
+	query = {},
+	requirement = DEFAULT_REQUIREMENT
+) => {
+	const resourceName = getResourceName( 'category-query', query );
+	return requireResource( requirement, resourceName ).totalCount || 0;
```
coderkevin: Just to check, this will trigger a fetch if it's not available. Is that what you want here? If not, maybe switch this to `getResource` instead.

joshuatf: Gotcha- thanks!  Updated to use `getResource`.

joshuatf:
Thanks for the review, @coderkevin!

For branch naming, do you have a preferred style?  I've been going with the Calypso style `{type}/{name}/{issue}` naming scheme, but maybe it's better to drop the issue number from the end?

client/wc-api/categories/operations.js
joshuatf: That's awesome!  I didn't know we had this.  Thanks @ryelle .

I prematurely merged this before I saw your comment, but opened an issue to handle this - https://github.com/woocommerce/wc-admin/issues/1148

PR #1132, opened 2018-12-19
Fix Tag misalignment inside Search in Firefox

Aljullu:
Fixes a visual glitch that was affecting the Search box in Firefox (see screenshots below).

### Screenshots
_Before:_
before

_After:_
after

- Go to the _Customers_ report with Firefox.
- Try applying an advanced filter.
- Verify the search box doesn't change its size when adding a `Tag` and they are vertically aligned.
- Verify it didn't cause regressions with other browsers.


psealock (approved):
Works as expected. Tested with Chrome, FF, Safari

PR #1133, opened 2018-12-19
Add Accessibility section to the PR template

ryelle:
This PR adds a new section to the PR template for accessibility, adding some checkboxes to remind everyone to test the a11y of new features. These can be deleted if they don't apply to the PR.

See an example on https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/266 – I don't need all the items because this PR has no animations or new colors, but I can verify that I tested with a screen reader & keyboard.


psealock (approved):
Nice, LGTM

PR #1134, opened 2018-12-19
Variations Table: add product, variations query params

psealock:
Add `variations` and `products` query parameters back to Variations Table query. They've somehow been removed.

### Test

1. Find a product with multiple variations. `Hoodie` comes with some if you're using sample data.
2. Go to Product Report and Show > Single Product > Hoodie.
3. See the table populate with those variations, not all variations across all products.


timmyc (approved):
Changes LGTM

PR #1144, opened 2018-12-20
Add `/reports/customers` endpoint

### LINKED ISSUE #916: REST API: Implement /reports/customers endpoint
p2: p7bje6-16h-p2
swagger: [https://app.swaggerhub.com/apis/peterfabian/wc-v3-api/1.0.0#/default/get_reports_customers](https://app.swaggerhub.com/apis/peterfabian/wc-v3-api/1.0.0#/default/get_reports_customers)

Currently the `/reports/customers` endpoint has yet to be fully implemented. It appears the new endpoint will require another reporting table to be built, so for this issue we will need the required table to be built, along with the endpoint logic.

jeffstieler:
Fixes #916.

This PR seeks to implement the `/reports/customers` endpoint.

It introduces a new lookup table, `wc_customer_lookup`. Schema discussion: p7bje6-1dM-p2.

- Run `phpunit` and make sure all tests are passing.
- Ensure that your test store has some registered customers
- Deactivate/activate `wc-admin` to kick off customer lookup table creation
- Verify that all registered customers have a row in `wc_customer_lookup`
- Verify that checking out as a guest results in a new row in `wc_customer_lookup` without a `user_id`
- Verify that registering at checkout results in a new row in `wc_customer_lookup` with a `user_id`
- Use Postman or another tool to make some test requests to `/wc/v3/reports/customers`
Available params:
`page`
`per_page`
`order` (`asc`, `desc`)
`orderby` (`username`, `name`, `country`, `city`, `postcode`, `date_registered`, `date_last_active`, `orders_count`, `total_spend`, `avg_order_value`)
`match` (`any`, `all`)
`name`
`username`
`email`
`country`
`last_active_before`
`last_active_after`
`registered_before`
`registered_after`
`orders_count_min`
`orders_count_max`
`total_spend_min`
`total_spend_max`
`avg_order_value_min`
`avg_order_value_max`
`last_order_before`
`last_order_after`

Note that all "range" type filters will be `AND`ed regardless of `match` value. E.g. `orders_count_min = 5, orders_count_max = 10` always results in `(orders_count >= 5 AND orders_count <= 10)`.


includes/api/class-wc-admin-rest-reports-customers-controller.php
```diff
@@ -235,6 +279,41 @@ public function get_collection_params() {
...
+			'enum'              => array( 'asc', 'desc' ),
+			'validate_callback' => 'rest_validate_request_arg',
+		);
+		$params['orderby']             = array(
+			'description'       => __( 'Sort collection by object attribute.', 'wc-admin' ),
+			'type'              => 'string',
+			'default'           => 'date_registered',
+			'enum'              => array(
+				'username',
+				'name',
```
timmyc: How do we orderby name since it is a combined field?

includes/class-wc-admin-reports-customers-query.php
```diff
@@ -0,0 +1,54 @@
...
+class WC_Admin_Reports_Customers_Query extends WC_Admin_Reports_Query {
+
+	/**
+	 * Valid fields for Customers report.
+	 *
+	 * @return array
+	 */
+	protected function get_default_query_vars() {
+		return array(
+			'per_page' => get_option( 'posts_per_page' ), // not sure if this should be the default.
```
timmyc: Seems okay to me - I think be default we are doing, seems like most of the query classes omit this default query.

includes/data-stores/class-wc-admin-reports-customers-data-store.php
```diff
@@ -0,0 +1,399 @@
...
+	/**
+	 * SQL columns to select in the db query and their mapping to SQL code.
+	 *
+	 * @var array
+	 */
+	protected $report_columns = array(
+		'customer_id'      => 'customer_id',
+		'user_id'          => 'user_id',
+		'username'         => 'username',
+		'name'             => "CONCAT_WS( ' ', first_name, last_name ) as name", // TODO: what does this mean for RTL?
```
timmyc: Aye didn't think about the RTL impact here. Guessing as long as we are following suit with how WC handles first/last names in other areas ( order listing, order single ) we are fine.

includes/data-stores/class-wc-admin-reports-customers-data-store.php
```diff
@@ -0,0 +1,399 @@
...
+
+	/**
+	 * Maps ordering specified by the user to columns in the database/fields in the data.
+	 *
+	 * @param string $order_by Sorting criterion.
+	 * @return string
+	 */
+	protected function normalize_order_by( $order_by ) {
+		if ( 'name' === $order_by ) {
+			return "CONCAT_WS( ' ', first_name, last_name )";
```
timmyc: Disregard my question above about ordering. The only thing I can see this causing problems is some people might anticipate ordering by name to be last name driven.

includes/data-stores/class-wc-admin-reports-customers-data-store.php
```diff
@@ -0,0 +1,399 @@
...
+			if ( ! empty( $query_args[ $before_arg ] ) ) {
+				$datetime     = new DateTime( $query_args[ $before_arg ] );
+				$datetime_str = $datetime->format( WC_Admin_Reports_Interval::$sql_datetime_format );
+				$subclauses[] = "{$table_name}.{$column_name} <= '$datetime_str'";
+			}
+
+			if ( ! empty( $query_args[ $after_arg ] ) ) {
+				$datetime     = new DateTime( $query_args[ $after_arg ] );
+				$datetime_str = $datetime->format( WC_Admin_Reports_Interval::$sql_datetime_format );
+				$subclauses[] = "{$table_name}.{$column_name} <= '$datetime_str'";
```
timmyc: Shouldn't this be `"{$table_name}.{$column_name} >= '$datetime_str'";` since this is the after arg clause?

includes/data-stores/class-wc-admin-reports-customers-data-store.php
```diff
@@ -0,0 +1,399 @@
...
+					WHERE
+						1=1
+						{$sql_query_params['where_time_clause']}
+						{$sql_query_params['where_clause']}
+				"
+			); // WPCS: cache ok, DB call ok, unprepared SQL ok.
+
+			$total_pages = (int) ceil( $db_records_count / $sql_query_params['per_page'] );
+			if ( $query_args['page'] < 1 || $query_args['page'] > $total_pages ) {
+				return $data;
```
timmyc: When we return early, do we still want to set the cache like we do in line 307?

includes/data-stores/class-wc-admin-reports-customers-data-store.php
```diff
@@ -0,0 +1,399 @@
...
+				'username'         => $customer->get_username( 'edit' ),
+				'first_name'       => $customer->get_first_name( 'edit' ),
+				'last_name'        => $customer->get_last_name( 'edit' ),
+				'email'            => $customer->get_email( 'edit' ),
+				'city'             => $customer->get_billing_city( 'edit' ),
+				'postcode'         => $customer->get_billing_postcode( 'edit' ),
+				'country'          => $customer->get_billing_country( 'edit' ),
+				'orders_count'     => $order_count,
+				'total_spend'      => $total_spend,
+				'avg_order_value'  => $order_count ? ( $total_spend / $order_count ) : 0,
```
timmyc: should we also verify that $total_spend is > 0? Perhaps a customer has an order, but the order had zero value ( sounds strange, but i have "customers" like this on a site )

tests/api/reports-customers.php
```diff
@@ -0,0 +1,155 @@
...
+		$this->assertEquals( 200, $response->get_status() );
+		$this->assertEquals( 5, count( $reports ) );
+		$this->assertArrayHasKey( 'X-WP-Total', $headers );
+		$this->assertEquals( 10, $headers['X-WP-Total'] );
+		$this->assertArrayHasKey( 'X-WP-TotalPages', $headers );
+		$this->assertEquals( 2, $headers['X-WP-TotalPages'] );
+		$this->assertEquals( $test_customers[0]->get_id(), $reports[0]['user_id'] );
+		$this->assertEquals( 1, $reports[0]['orders_count'] );
+		$this->assertEquals( 100, $reports[0]['total_spend'] );
+		$this->assert_report_item_schema( $reports[0] );
```
timmyc: Might be nice to have a few more test cases here around some of the REST API query options like a date based one, maybe one on the name filter too, just because that is quite unique to this endpoint.

timmyc (commented):
:boom: solid work here Jeff! Some comments for now, and I'll take it for a spin shortly.

includes/data-stores/class-wc-admin-reports-customers-data-store.php
jeffstieler: Core uses a translatable format to stitch together the first and last names: https://github.com/woocommerce/woocommerce/blob/master/includes/admin/list-tables/class-wc-admin-list-table-orders.php#L168

includes/data-stores/class-wc-admin-reports-customers-data-store.php
jeffstieler: Should we make a change to that?

includes/data-stores/class-wc-admin-reports-customers-data-store.php
jeffstieler: Whoops! Copy pasta. 🍝

jeffstieler: Fixed in 72a7cea.

includes/data-stores/class-wc-admin-reports-customers-data-store.php
jeffstieler: I don't think so - in this case, a page that is outside the possible result set is being requested (page `0` or a page beyond the total number of pages). We haven't actually queried for data yet, just a count, so there's really nothing to cache.

includes/data-stores/class-wc-admin-reports-customers-data-store.php
jeffstieler: I don't think we need to. If `$total_spend` is `0`, then the division operation results in `0`. If `$order_count` is `0` and we don't prevent the division, we get PHP warnings.

tests/api/reports-customers.php
jeffstieler: Great thinking, this teased out a bug in the `name` parameter! Fixed bug in 72a7cea, added tests in c07474e.

includes/class-wc-admin-api-init.php
```diff
@@ -445,6 +508,26 @@ private static function get_schema() {
...
+				username varchar(60) DEFAULT '' NOT NULL,
+				first_name varchar(255) NOT NULL,
+				last_name varchar(255) NOT NULL,
+				email varchar(100) NOT NULL,
+				orders_count BIGINT UNSIGNED NOT NULL,
+				total_spend double DEFAULT 0 NOT NULL,
+				avg_order_value double DEFAULT 0 NOT NULL,
+				date_last_active timestamp DEFAULT '0000-00-00 00:00:00' NOT NULL,
+				date_last_order timestamp DEFAULT '0000-00-00 00:00:00' NOT NULL,
+				date_registered timestamp DEFAULT '0000-00-00 00:00:00' NOT NULL,
```
jeffstieler: @timmyc

I was just following the pattern of the other report tables here - why are we using `timestamp` instead of `datetime`?

Also, I think I'm going to allow these columns to be null. They won't always have a value, and the all-zero default is causing stuff like `-0001-11-30T00:00:00` when processed through `wc_rest_prepare_date_response()`.

includes/data-stores/class-wc-admin-reports-customers-data-store.php
timmyc: TIL you can divide zero in PHP. :trollface:

tests/api/reports-customers.php
timmyc: 💯

includes/class-wc-admin-api-init.php
timmyc: > I was just following the pattern of the other report tables here - why are we using timestamp instead of datetime

I'm not entirely sure the background on that decision, but am guessing it might be detailed in some old posts on +protonpower - but perhaps @claudiosanches  or @peterfabian might know why.

includes/class-wc-admin-api-init.php
```diff
@@ -315,6 +321,101 @@ public static function order_product_lookup_store_init( $updater = null ) {
...
+				return true;
+			}
+		}
+
+		// Backfill customer lookup table with guests.
+		$guest_order_ids = get_transient( 'wc_update_350_all_guest_orders' );
+
+		if ( false === $guest_order_ids ) {
+			$guest_order_ids = wc_get_orders(
+				// TODO: restrict to certain order status?
```
joshuatf: I think for now it's safe to store all of them as these will still provide insight into user interaction with the site and won't alter the actual order stats.

If we're connecting the order lookup table here, filtering by number of orders >=1 will only show customers with accepted order statuses.

includes/api/class-wc-admin-rest-reports-customers-controller.php
```diff
@@ -39,25 +39,30 @@ class WC_Admin_REST_Reports_Customers_Controller extends WC_REST_Reports_Control
...
 		$args['email']                   = $request['email'];
 		$args['country']                 = $request['country'];
 		$args['last_active_before']      = $request['last_active_before'];
 		$args['last_active_after']       = $request['last_active_after'];
-		$args['order_count_min']         = $request['order_count_min'];
-		$args['order_count_max']         = $request['order_count_max'];
-		$args['order_count_between']     = $request['order_count_between'];
+		$args['orders_count_min']        = $request['orders_count_min'];
+		$args['orders_count_max']        = $request['orders_count_max'];
+		$args['orders_count_between']    = $request['orders_count_between'];
```
joshuatf: Do we want to normalize this to set `orders_count_min` and `orders_count_max` if it's present?

joshuatf (commented):
Looks great and tests well, @jeffstieler!

Left a comment as the only thing I had trouble with was the between params, but maybe I'm using them wrong.

includes/data-stores/class-wc-admin-reports-customers-data-store.php
```diff
@@ -0,0 +1,523 @@
...
+
+		if ( $existing_guest ) {
+			$order_timestamp  = date( 'Y-m-d H:i:s', $order->get_date_created( 'edit' )->getTimestamp() );
+			$new_orders_count = $existing_guest['orders_count'] + 1;
+			$new_total_spend  = $existing_guest['total_spend'] + (float) $order->get_total( 'edit' );
+
+			return $wpdb->update(
+				$wpdb->prefix . self::TABLE_NAME,
+				array(
+					'orders_count'     => $new_orders_count,
```
joshuatf: Since we need to add the `customer_id` to the order stats table, I'm wondering if we should just perform a JOIN here on the order stats to get the current information.

We can then grab `orders_count`, `total_spend`, `avg_order_value`, and `date_last_order` without needing to update them here regularly.

I think this would result in less syncing issues and less overhead here.  Thoughts?

includes/data-stores/class-wc-admin-reports-customers-data-store.php
```diff
@@ -0,0 +1,523 @@
...
+		if ( empty( $email ) ) {
+			return true;
+		}
+
+		$existing_guest = $this->get_guest_by_email( $email );
+
+		if ( $existing_guest ) {
+			$order_timestamp  = date( 'Y-m-d H:i:s', $order->get_date_created( 'edit' )->getTimestamp() );
+			$new_orders_count = $existing_guest['orders_count'] + 1;
+			$new_total_spend  = $existing_guest['total_spend'] + (float) $order->get_total( 'edit' );
```
joshuatf: Since this takes into account all orders (including cancelled, refunded, etc), `total_spend` might be off.

includes/data-stores/class-wc-admin-reports-customers-data-store.php
```diff
@@ -0,0 +1,523 @@
...
+
+		$email = $order->get_billing_email( 'edit' );
+
+		if ( empty( $email ) ) {
+			return true;
+		}
+
+		$existing_guest = $this->get_guest_by_email( $email );
+
+		if ( $existing_guest ) {
```
joshuatf: If we make changes to customer stats below, we might be able to clean up this logic a bit by using `$wpdb->replace()` and appending the customer ID.

[Example of this](https://github.com/woocommerce/wc-admin/blob/20fad383849bd33094f5893f3238f1bd46f5fb6d/includes/data-stores/class-wc-admin-reports-customers-data-store.php#L479).

joshuatf (commented):
Sorry for adding another review here.  I ran into a couple issues when working on https://github.com/woocommerce/wc-admin/issues/558 that made me rethink this a bit.

If you think these are better handled in a follow-up PR, I think that's okay too.

includes/api/class-wc-admin-rest-reports-customers-controller.php
jeffstieler: Definitely - this was missed and is why the `_between` params didn't work in your testing.

includes/data-stores/class-wc-admin-reports-customers-data-store.php
jeffstieler: _Restating to make sure I've understood:_

1. At the time of guest order creation (and therefore update to the `wc_order_stats` table) we'd create a customer in the `wc_customer_lookup` table if none exists for the billing email
2. When hitting the `reports/customers` endpoint, the values for `orders_count`, `total_spend`, `avg_order_value`, and `date_last_order` would come from a `JOIN` on the `wc_order_stats` table, and all syncing code for `wc_customer_lookup` can be removed

includes/data-stores/class-wc-admin-reports-customers-data-store.php
jeffstieler: Yep, we could just rework [`update_registered_customer()`](https://github.com/woocommerce/wc-admin/pull/1144/files/1399528579267b959af4bbefec04e57a200c4f20#diff-061c71b3c8238141695a290a5becc8baR466) instead of having multiple functions.

includes/class-wc-admin-api-init.php
jeffstieler: Allowed the `date_registered` column to be null in 91c8495.

includes/api/class-wc-admin-rest-reports-customers-controller.php
jeffstieler: On second thought.. perhaps we should remove that parameter and handle this normalization on the JS side?

jeffstieler: Nevermind on the JS-side normalization, I think that's perhaps the wrong place to perform this operation.

In an effort to land this PR, and to properly handle `_between` params for all reports that use the advanced number filter, I've opened https://github.com/woocommerce/wc-admin/issues/1239.

For this PR, the `_between` param will continue to be inert.

timmyc:
@jeffstieler is this ready for another review?

jeffstieler:
@timmyc need a resolution to this: https://github.com/woocommerce/wc-admin/pull/1144#discussion_r243981590

includes/data-stores/class-wc-admin-reports-customers-data-store.php
timmyc: The solution you outlined above @jeffstieler seems reasonable to me, and will ensure the most current data about a customer for the given query to the REST API. Also seems like it should be performant.

joshuatf: Looks good to me, @jeffstieler 👍

jeffstieler:
This is ready to be retested @timmyc @joshuatf - a full retest/review is required since so much changed in the refactor to JOIN on the order stats table.

includes/api/class-wc-admin-rest-reports-customers-controller.php
joshuatf: Seems like a good way of handling this 👍

timmyc (approved):
First off this is some seriously awesome stuff here Jeff - to me this is one of the bangers of all the new reports, one that Woo users will utilize quite a bit, so thanks for tackling this beefy task. phpunit tests are all green, and looking good, I've left some comments in the code, but nothing major.

I noodled around a bit using the "smooth generated" customer data, and all was functioning well, and then I spent some more time creating some guest orders with the same email address, and registering a new user and doing some odd things with it too. Generally trying to find things that didn't work as expected :)

The first oddness I encountered was checking out with two distinct orders using the same email address. The customer response from the API looked great wrt number of orders, total spend and average. The only thing I could get to not work properly was I switched the postal code on the last order, and the latest postal code was not shown in the REST API response - the one used in the first order was. Super minor, I think this is fine as a follow-up task or really not sure if it is a bug we need to concern ourselves with at all.

The next issue I hit, as I mentioned in slack - was `null` timestamps being returned as `-0001-11-30T00:00:00` from the API. Again I think we can handle this in a follow-up PR, unless you have already implemented a fix.

The last flow I tested out was to place orders as a logged-in, registered user. All worked as expected 🌟 - and then I logged out of the account, and created a "guest" order using the same email address. This resulted in a new Customer entry being created:

image

Not sure if this is a big deal either, because really the order was not attached to the account, but figured I'd mention it.

All in all - I'm thinking we should land this PR, continue to test on it, and discuss little things like this in follow-up issues. As such I'm going to give this one the ✅ to merge unless you wanted to fix up some things I noted above first.

includes/data-stores/class-wc-admin-reports-customers-data-store.php
```diff
@@ -0,0 +1,574 @@
...
+	 */
+	protected function get_time_period_sql_params( $query_args, $table_name ) {
+		global $wpdb;
+
+		$sql_query           = array(
+			'where_time_clause' => '',
+			'where_clause'      => '',
+			'having_clause'     => '',
+		);
+		$date_param_mapping  = array(
```
joshuatf: Love how all these dates are handled. ❤️

joshuatf (approved):
Fantastic work on this, Jeff!

Everything is testing really well for me.  Only issue I had is some of my orders getting a customer ID of `0` on rebuild.  Wondering if this is an issue with my own install or a bug, but either way I think we can follow up in another PR.

>and then I logged out of the account, and created a "guest" order using the same email address. This resulted in a new Customer entry being created:

@timmyc I would have expected this behavior since the customer chose not to sign in, but I suppose the email should be unique enough to associate with an existing user.  Would separating into guest and user be more useful for reporting or combining their data?

includes/data-stores/class-wc-admin-reports-customers-data-store.php
```diff
@@ -0,0 +1,574 @@
...
+	 *
+	 * @param int $user_id WP User ID to update customer data for.
+	 * @return int|bool|null Number or rows modified or false on failure.
+	 */
+	public static function update_registered_customer( $user_id ) {
+		global $wpdb;
+
+		$customer = new WC_Customer( $user_id );
+
+		if ( $customer->get_id() != $user_id ) {
```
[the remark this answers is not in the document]
jeffstieler: Quite easily in fact - if you give the `WC_Customer` constructor a bad user id, the subsequent `get_id()` call returns `0`.

PR #1150, opened 2018-12-21
Make user column preferences refer to hidden columns

### LINKED ISSUE #1136: New columns are hidden by default in reports table
**Describe the bug**
#1082 added in persistent column settings based on user preferences 🎉 

A minor issue is if a new column is added by us or 3rd party devs, they will be hidden by default.  We might consider saving the user preferences as the hidden columns instead of the shown columns in https://github.com/woocommerce/wc-admin/blob/9c0c6faba85937c020a9f2564bead90546c05a66/client/analytics/components/report-table/index.js#L40

**To Reproduce**

1. Go to any report table.
2. Add a new column (via code editor).
3. Note on refresh that the column is hidden by default.

**Expected behavior**
Expected new columns to be shown by default.

**Screenshots**
screen shot 2018-12-20 at 9 48 27 am

joshuatf:
Fixes #1136 

Changes column preferences to store hidden columns instead of shown so that new columns are shown by default.  Also prevents required columns from being saved to this array.

### Screenshots
screen shot 2018-12-21 at 3 01 00 pm

1.  Open any report page.
2.  Turn on/off various columns.
3.  Check that columns continue to persist across browers/sessions.
4.  Note that new columns (e.g., Products-> Net Revenue) are now shown by default.

### DIFF
--- a/client/analytics/components/report-table/index.js
+++ b/client/analytics/components/report-table/index.js
@@ -26,24 +26,26 @@ import { extendTableData } from './utils';
 const TABLE_FILTER = 'woocommerce_admin_report_table';
 
 class ReportTable extends Component {
-	onColumnsChange = columns => {
+	onColumnsChange = shownColumns => {
 		const { columnPrefsKey } = this.props;
+		const columns = this.props.getHeadersContent().map( header => header.key );
+		const hiddenColumns = columns.filter( column => ! shownColumns.includes( column ) );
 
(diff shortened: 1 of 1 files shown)


timmyc (approved):
Works as described on the tin! Also I think this is a good approach to ensure any new column that gets added is shown by default - also might reduce the size of the persisted value by only storing hidden columns.

PR #1153, opened 2018-12-22
Chart: Fix/current of undefined

timmyc:
While testing PRs lately, I would periodically see the following error:

dashboard woo test woocommerce 2018-12-21 16-44-23

I could never track down exact steps to reproduce the bug, but it seems to be thrown potentially when a chart component is re-rendering or unmounting. I figured adding in a simple undefined check for the ref would be easy enough to add, so here it is.

__To Test__
Verify charts and chart legends are still working properly.


coderkevin (approved):
Good catch! LGTM

PR #1162, opened 2018-12-24
Clear filter query on removing all items

### LINKED ISSUE #998: Removing all filters does not reset query
**Describe the bug**
Removing the filters from a comparison by using the "x" does not clear the attributes from the query.

**To Reproduce**

1. Go to `wp-admin/admin.php?page=wc-admin#/analytics/products?filter=compare-products`.
2. Add multiple products to the comparison and click "Compare".
3. Remove those products using the "x" next to each.
4. Note that the URL still contains the the filter query `products?filter=compare-products&products=xxxxxxx`

**Expected behavior**
Removing all products (or maybe anything < 2 in the case of the comparison filter) should remove this filter.  To remove these filters, you'll either need to switch to another filter or add the products back and click "Clear all."

**Screenshots**
screen shot 2018-12-05 at 2 24 53 pm

joshuatf:
Fixes #998 

Clears the filter query from the URL when removing all tags/items from a filter.

### Screenshots
screen shot 2018-12-24 at 12 35 32 pm

1.  Visit any reports page (e.g. Products report).
2.  Add a comparison filter.
3.  Add items by typing/searching for items and click on the item.
4.  After adding at least 2 items, click "Compare."
5.  Note the query params for this filter in the URL.
6.  Use the "x" next to the tag to remove each item.
7.  On the last item removal, note the removal in the query params from the URL.

### DIFF
--- a/packages/components/src/filters/compare/index.js
+++ b/packages/components/src/filters/compare/index.js
@@ -39,13 +39,11 @@ class CompareFilter extends Component {
 		}
 	}
 
-	componentDidUpdate( { param: prevParam, query: prevQuery } ) {
-		const { getLabels, param, path, query } = this.props;
-		if ( prevParam !== param ) {
-			/* eslint-disable react/no-did-update-set-state */
-			this.setState( { selected: [] } );
-			/* eslint-enable react/no-did-update-set-state */
-			updateQueryString( { [ param ]: undefined }, path, query );
+	componentDidUpdate( { param: prevParam, query: prevQuery }, { selected: prevSelected } ) {
+		const { getLabels, param, query } = this.props;
+		const { selected } = this.state;
+		if ( prevParam !== param || ( prevSelected.length > 0 && selected.length === 0 ) ) {
+			this.clearQuery();
 			return;
 		}
 


packages/components/src/filters/compare/index.js
```diff
@@ -39,13 +39,11 @@ class CompareFilter extends Component {
...
-		if ( prevParam !== param ) {
-			/* eslint-disable react/no-did-update-set-state */
-			this.setState( { selected: [] } );
-			/* eslint-enable react/no-did-update-set-state */
-			updateQueryString( { [ param ]: undefined }, path, query );
+	componentDidUpdate( { param: prevParam, query: prevQuery }, { selected: prevSelected } ) {
+		const { getLabels, param, query } = this.props;
+		const { selected } = this.state;
+		if ( prevParam !== param || ( prevSelected.length > 0 && selected.length === 0 ) ) {
+			this.clearQuery();
```
Aljullu: Good catch we could use `this.clearQuery()` here. :clap:

Aljullu (approved):
Working great.

 :shipit:

PR #1164, opened 2018-12-24
Check for product categories before loading value

### LINKED ISSUE #1163: Categories don't load in time for product report
**Describe the bug**
On the products report, with many categories added (100s), sometimes the categories don't load in time, causing the category column to return an error.
 
**To Reproduce**

1. Add a lot of categories (maybe around 2-300).
2. Visit the products report page.
3. Note the failure to load and console error.

**Expected behavior**
Expected page to load without issue.

**Screenshots**
screen shot 2018-12-24 at 12 33 16 pm

**Additional context**
Probably just need to check if categories is defined yet or not.

joshuatf:
Fixes #1163 

Checks if product categories exist before attempting to map them to column value.

1.  Add a lot of product categories (~200).
2.  Open the products report.
3.  Note that the page loads as expected and no console warnings exist.

### DIFF
--- a/client/analytics/report/products/table.js
+++ b/client/analytics/report/products/table.js
@@ -166,7 +166,8 @@ class ProductsReportTable extends Component {
 							) }
 						</div>
 					),
-					value: productCategories.map( category => category.name ).join( ', ' ),
+					value:
+						productCategories && productCategories.map( category => category.name ).join( ', ' ),
 				},
 				{
 					display: numberFormat( variations.length ),
--- a/client/analytics/report/products/table.js
+++ b/client/analytics/report/products/table.js
@@ -112,7 +112,9 @@ class ProductsReportTable extends Component {
 				products: product_id,
 			} );
 			const categories = this.props.categories;
-			const productCategories = category_ids.map( category_id => categories[ category_id ] );
+			const productCategories = category_ids
+				.map( category_id => categories[ category_id ] )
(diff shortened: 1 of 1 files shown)


Aljullu (commented):
Not sure if it's the same bug or another one, but in some cases I get the list `productCategories` full of undefined values, and that still produces JS errors in some other parts.

That's probably because we would need to filter the array after doing this:
```ES6
const productCategories = category_ids.map( category_id => categories[ category_id ] );
```

joshuatf:
@Aljullu Good catch and you're right.  I have another PR open that addresses that ( https://github.com/woocommerce/wc-admin/pull/1149/files ) but I'll add this to this PR for now as it's a blocker for working on products.

Or we can also merge that PR in and close this one if things test well for you there.

client/analytics/report/products/table.js
```diff
@@ -166,7 +168,8 @@ class ProductsReportTable extends Component {
 							) }
 						</div>
 					),
-					value: productCategories.map( category => category.name ).join( ', ' ),
+					value:
+						productCategories && productCategories.map( category => category.name ).join( ', ' ),
```
Aljullu: So, now this change is no longer necessary, right?

Aljullu (approved):
Looking good. :rocket: 

Just a small comment.

PR #1166, opened 2018-12-24
Set minimum content height to prevent cutoff of absolute content

### LINKED ISSUE #1107: Autocompleter: Results get cut off with short table
**Describe the bug**
When a short table exists, the search autocompleter results get cut off.

**To Reproduce**

1. Go to Customers Report
2. Ensure the table only has one row
3. Search for a string with lots of results
4. See the results getting cut off

**Screenshots**

cutoff

joshuatf:
Fixes #1107 

Sets a minimum height on the content body so that long search results do not get cut off.

### Before
screen shot 2018-12-24 at 2 27 39 pm

### After
screen shot 2018-12-24 at 2 04 58 pm

1.  Open any report with a search.
2.  Make sure only 1 row is visible to limit content height.
3.  Search using a term that will return at least 10 items.
4.  Note that all items are visible and not cut off.

### DIFF
--- a/client/header/style.scss
+++ b/client/header/style.scss
@@ -11,7 +11,7 @@
 	height: 80px;
 	position: fixed;
 	width: 100%;
-	top: 32px;
+	top: $adminbar-height;
 	z-index: 99991; /* higher than component-popover (99990), lower than notification dropdown at 99999 */
 
 	&.is-scrolled {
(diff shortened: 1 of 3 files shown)


Aljullu (approved):
That works as expected. :clap: 

The only small thing I was wondering is whether it would make sense to move the `32px` and `46px` values into variables, maybe inside `client/stylesheets/abstracts/_variables.scss`. That would be useful to self-document these declarations and would make it easier to reuse the values if at some point we need to. What do you think?

joshuatf:
@Aljullu Fantastic idea.  Added those and replaced where I could find references that were used to manipulate distance due to admin bar (e.g., the main header navbar).

Issue #1159, opened 2018-12-23
`reports/categories` returns false positives when filtering by category if a product is in several categories


Aljullu:
Having a product P1 that belongs to category C1 and C2. If you query `reports/categories` with `categories=<C1_id>`, it will return data for C1 and C2.

I guess, the issue is we are using `get_included_products` ([#](https://github.com/woocommerce/wc-admin/blob/master/includes/data-stores/class-wc-admin-reports-categories-data-store.php#L82), [#](https://github.com/woocommerce/wc-admin/blob/master/includes/data-stores/class-wc-admin-reports-data-store.php#L570-L573)) to filter categories, which might not be reliable if products belong to several categories.

----

Not sure if it helps, but here is an example from my system with some screenshots.

Query:
![image](https://user-images.githubusercontent.com/3616980/50387970-6f9a6400-0709-11e9-83ce-375ac1445c6c.png)

Response:
![image](https://user-images.githubusercontent.com/3616980/50387973-7b862600-0709-11e9-843c-e9063fef80ef.png)

Notice the query has two ID's in the `categories` param (206 and 207) but three categories are returned (156, 206 and 207).

Products of category 156:
![image](https://user-images.githubusercontent.com/3616980/50387987-e2a3da80-0709-11e9-971c-f265bee2df08.png)

Products of category 206:
![image](https://user-images.githubusercontent.com/3616980/50387992-f9e2c800-0709-11e9-85a1-85342c10eab8.png)

Products of category 207:
![image](https://user-images.githubusercontent.com/3616980/50387993-05ce8a00-070a-11e9-8cf3-bd6bfa2f1733.png)

PR #1167, opened 2018-12-24
Add Categories report filters

### LINKED ISSUE #1157: Analytics: Exception in Categories Report when Using Advanced Filters
When selecting "Advanced Filters" for the categories report, an exception is thrown:

image

__Steps To Reproduce__
- Open the categories report
- Select "Advanced Filters" from the "Show" drop down in the filter area

Aljullu:
Fixes #1157.

Until now, we were showing an _Advanced filters_ option, but according to the designs the filters should be _Comparison_, _Top Categories by Items Sold_ and _Top Categories by Net Revenue_.

### Screenshots
image

- Test all filters and verify they work as expected and show the `item-comparison` chart (the legend appears on the left).
- Verify the error described in #1157 doesn't appear anymore.

_Note:_ The _Comparison_ seems to have a bug which doesn't correctly filter the categories (#1159).

### DIFF
--- a/client/analytics/report/categories/config.js
+++ b/client/analytics/report/categories/config.js
@@ -4,6 +4,12 @@
  */
 import { __ } from '@wordpress/i18n';
 
+/**
+ * Internal dependencies
+ */
+import { getRequestByIdString } from 'lib/async-requests';
+import { NAMESPACE } from 'store/constants';
+
 export const charts = [
 	{
 		key: 'items_sold',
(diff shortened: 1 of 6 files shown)


joshuatf (approved):
Looking good and testing well!  LGTM 🚢 

Only thing I noticed was that the "Comparison" filter doesn't seem to work properly, but the client-side filter is setting the correct params and it looks like the issue is on the API side.  Will open a separate issue for that.

PR #1168, opened 2018-12-24
Charts: fix weird Y-axis ticks when max Y value is a small number

### LINKED ISSUE #1155: Chart: Y-Axis Tick Issues
When viewing a chart that is plotting quantity 1 ( i.e. one item sold in a time period ) - three y-axis ticks are displayed for "2 - 1 - 1"

products analytics woo test woocommerce 2018-12-20 12-42-57

Also from @LevinMedia 

When the only data to graph has a value of 1, the y-axis ticks count: 1, 1, 2

screen shot 2018-12-14 at 11 16 37 pm

they should count 1, 2, 3, and the bar / line should only extend to the first y-axis tick.

Aljullu:
Fixes #1155.

When the max Y value was a small number, Y-axis ticks could be a decimal number, which was later rounded. That made that some times different Y-axis ticks displayed the same number (see screenshot below: there are two `1` ticks).

This PR forces the max Y value to be an integer and all Y-axis ticks to be integers too and not to be repeated.

### Screenshots
_Before:_
image

_After:_
image

- Go to a report and select the filters/date range so the maximum Y value in the chart is 1.
- Verify Y-axis ticks are not repeated.

### DIFF
--- a/packages/components/src/chart/d3chart/utils.js
+++ b/packages/components/src/chart/d3chart/utils.js
@@ -176,14 +176,14 @@ export const getXLineScale = ( uniqueDates, width ) =>
 		.rangeRound( [ 0, width ] );
 
 /**
- * Describes and rounds the maximum y value to the nearest thousadn, ten-thousand, million etc.
+ * Describes and rounds the maximum y value to the nearest thousand, ten-thousand, million etc. In case it is a decimal number, ceils it.
  * @param {array} lineData - from `getLineData`
  * @returns {number} the maximum value in the timeseries multiplied by 4/3
  */
 export const getYMax = lineData => {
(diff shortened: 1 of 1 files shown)


packages/components/src/chart/d3chart/utils.js
```diff
@@ -176,14 +176,14 @@ export const getXLineScale = ( uniqueDates, width ) =>
 		.rangeRound( [ 0, width ] );
 
 /**
- * Describes and rounds the maximum y value to the nearest thousadn, ten-thousand, million etc.
+ * Describes and rounds the maximum y value to the nearest thousand, ten-thousand, million etc. In case it is a decimal number, ceils it.
```
joshuatf: Nice typo catch!

joshuatf (approved):
Testing well!  :shipit:

PR #1170, opened 2018-12-25
Fix ie11 blank screen by babel transforms

### LINKED ISSUE #1062: Blank screen in IE11
Last versions of `wc-admin` seem to be broken in IE11:

image

`<div id="root">`, which should contain the pages markup, is empty. There are no JS errors in the Console nor Network failures.

(Tested with BrowserStack)

joshuatf:
Fixes #1062 

Adds babel transforms to specific vendor packages to transform ES6.  Also removes Buble since it seems it's not being used as direct dependency.

### Before
49806220-70211b00-fd57-11e8-976f-6281aded2228

### After
screen shot 2018-12-25 at 6 07 31 pm

1.  Open any wc admin page in IE11.
2.  Check that the page loads and no console errors appear.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -74,6 +74,26 @@ const webpackConfig = {
 				loader: 'babel-loader',
 				exclude: /node_modules/,
 			},
+			{
+				test: /\.js?$/,
+				use: {
+					loader: 'babel-loader',
+					options: {
+						presets: [
+							[ '@babel/preset-env', { loose: true, modules: 'commonjs' } ],
+						],
+						plugins: [ 'transform-es2015-template-literals' ],
+					},
+				},
+				include: new RegExp( '/node_modules\/(' +
+					'buble' +
+					'|d3-array' +
+					'|debug' +
+					'|regexpu-core' +
+					'|unicode-match-property-ecmascript' +
+					'|unicode-match-property-value-ecmascript)/'
+				),
+			},
 			{ test: /\.md$/, use: 'raw-loader' },
 			{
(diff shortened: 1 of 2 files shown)


webpack.config.js
```diff
@@ -74,6 +74,25 @@ const webpackConfig = {
...
+						],
+						plugins: [ 'transform-es2015-template-literals' ],
+					},
+				},
+				include: new RegExp( '/node_modules\/(' +
+					'|d3-array' +
+					'|debug' +
+					'|regexpu-core' +
+					'|unicode-match-property-ecmascript' +
+					'|unicode-match-property-value-ecmascript)/'
```
Aljullu: Does this list mean that every time we install a new package in the future we will have to check whether it works with IE11 and if not, add it here?

Aljullu (approved):
Happy to see `wc-admin` working in IE11 again! :clap:

webpack.config.js
joshuatf: Yes, unfortunately.  I'm not super happy about this either, but I'm not really sure if there are better ways.  The other options I looked at came with some pretty big downfalls.

I opened an issue [here](https://github.com/woocommerce/wc-admin/issues/1176) so we can discuss further and change this later if someone else knows a better way around this.

timmyc:
If I'm reading the Calypso PR correctly, it seems they have gone with a similar approach implemented in #1176 correct @joshuatf @Aljullu - if so I say we close this out. Not certain how often we will be adding dependencies to wc-admin to make the current solution a nuisance or not.

PR #1179, opened 2018-12-26
Filter by categories query param in categories data store

### LINKED ISSUE #1159: `reports/categories` returns false positives when filtering by category if a product is in several categories

joshuatf:
Fixes #1159 

Fixes the `categories` param in API requests to the category data store. 

### Before
screen shot 2018-12-26 at 3 56 54 pm

### After
screen shot 2018-12-26 at 3 57 03 pm

1.  Make a request to `/wp-json/wc/v3/reports/categories`.
2.  Add the categories param with a category ID or list of comma separated category IDs.
3.  Note the correct categories returned in the response.


Aljullu (approved):
Tests well and code looks good. :rocket:

PR #1180, opened 2018-12-26
Update returning customer data to use customer ID

### LINKED ISSUE #558: REST API: Add New/Returning Customer to Order Stats Endpoint
Follow-up issue of #418

As @Aljullu pointed out in the original issue:

> We are currently displaying the customer ID instead of the customer type (new/returning) because that's not implemented in the API yet.

So tagging @peterfabian and @claudiosanches on this one. Is it feasible to add this new customer field to the current API response, or are we going to have to pull data in from another endpoint.

joshuatf:
Fixes #558 

Updates the order stats table to use customer ID and provides accurate data for returning/new customers.  This PR includes changes that:
* Uses the new customer ID to associate both users and guests with an order
* Return the new/returning **customers** and not new/returning **orders** in order stats endpoint.

Some notes on how new/returning customers are calculated:
* Only orders with a `customer_id` field will be calculated.  You may need to rebuild your reports to get accurate information.
* Customers are only counted once no matter how many orders they have placed.
* If you request a period where a customer placed their first order and subsequent orders, this person is tracked as new since they became a new customer during this period.

### Before
screen shot 2018-12-27 at 5 29 05 pm

### After
screen shot 2018-12-27 at 5 28 46 pm

1.  Make sure you have some orders with the `customer_id` field and some orders from the same user. (note that you may need to deactivate and reactivate this plugin as well as run the rebuild reports tool to get customer IDs filled in).
2.  Open the order reports page. `/wp-admin/admin.php?page=wc-admin#/analytics/orders`
3.  Check that returning and new customer stats are correct and reflect customer numbers instead of returning/new orders.
4.  Check that new + returning customers is equal to total customers during that time period.
5.  Check that a customer with multiple orders, but having their initial order in a given time period is counted as a new customer.


timmyc:
When I went to re-run the rebuild report data tool - I had an exception thrown:

```
2018-12-22T01:35:34+00:00 CRITICAL Uncaught Error: Call to undefined method WC_Order_Refund::get_customer_id() in /srv/www/example/public_html/wp-content/plugins/wc-admin/includes/wc-admin-order-functions.php:31
Stack trace:
#0 /srv/www/example/public_html/wp-includes/class-wp-hook.php(288): wc_admin_order_product_lookup_entry(600)
#1 /srv/www/example/public_html/wp-includes/class-wp-hook.php(310): WP_Hook->apply_filters('', Array)
#2 /srv/www/example/public_html/wp-includes/plugin.php(453): WP_Hook->do_action(Array)
#3 /srv/www/example/public_html/wp-includes/post.php(5970): do_action('clean_post_cach...', 600, Object(WP_Post))
#4 /srv/www/example/public_html/wp-content/plugins/woocommerce/includes/data-stores/abstract-wc-order-data-store-cpt.php(311): clean_post_cache(Object(WP_Post))
#5 /srv/www/example/public_html/wp-content/plugins/woocommerce/includes/data-stores/abstract-wc-order-data-store-cpt.php(171): Abstract_WC_Order_Data_Store_CPT->clear_caches(Object(WC_Order_Refund))
#6 /srv/www/example/public_html/wp-content/plugins/woocommerce in /srv/www/example/public_html/wp-content/plugins/wc-admin/includes/wc-admin-order-functions.php on line 31
```

So I attempted to rebase off of `master` ( which I think might not have worked perfectly ) and when I went to test to the orders report, I'm seeing the following in the summary:


image

joshuatf:
@timmyc You're right- that issue should be solved in master.  Trying not to rebase from master since https://github.com/woocommerce/wc-admin/pull/1144 is still open and this PR is branched from that.

Despite the error, my `customer_id` columns still get populated on rebuild.  Can you check and see if you have IDs filled in that column?  If not, does manually adding IDs show the correct stats on the order reports?

includes/data-stores/class-wc-admin-reports-orders-data-store.php
```diff
@@ -497,23 +500,23 @@ protected static function get_num_items_sold( $order ) {
...
-				'meta_key'    => '_customer_user', // WPCS: slow query ok.
-				'meta_value'  => $customer_id, // WPCS: slow query ok.
-				'post_type'   => 'shop_order',
-				'post_status' => array( 'wc-on-hold', 'wc-processing', 'wc-completed' ),
-				'numberposts' => 2,
+		$customer_orders = $wpdb->get_var(
+			$wpdb->prepare(
+				"SELECT COUNT(*) FROM ${orders_stats_table} WHERE customer_id = %d AND date_created < %s",
+				$customer_id,
+				$order->get_date_created()->getTimestamp()
```
psealock: >If you request a period where a customer placed their first order and subsequent orders, this person is tracked as new since they became a new customer during this period.

If I placed an order in November

screen shot 2018-12-31 at 5 58 56 pm

And place an order in December, then Orders Report for this month (December 1 -31st), should show me as a returning customer:

screen shot 2018-12-31 at 5 53 05 pm

Maybe something wrong with this query?

psealock: Maybe I'm looking at it wrong, but `AND date_created < %s` should use the `after` query parameter, not the order timestamp so new/returning would be relative to the time window in question.

psealock (commented):
>Uses the new customer ID to associate both users and guests with an order

If I checkout as a guest, what should I see? It doesn't seem to be counted, is that right?

screen shot 2018-12-31 at 6 12 40 pm

If so, should we tally up new/returning/guest ?

includes/data-stores/class-wc-admin-reports-orders-data-store.php
joshuatf: @psealock `is_returning_customer()` is actually a static method run during order updates and report building and will not have access to `before` or `after` query params.

The `is_returning_customer` column is stored as a bool in the order lookup table. 
 This function will check if a user with the same ID made an order before the date of the order in question and if so, marks this one as a returning customer.

Do you think there needs to be any changes to the logic here?

joshuatf: If you have an order with a customer ID filled in from a previous period (as in your example where a customer ID is filled on an order in November and you're checking Dec 1-31st where an order is present with the same customer ID) then this person should be marked as returning.

If you have orders where the customer ID is filled in for those conditions, then the logic may be buggy here.

timmyc:
Just checking in to see if this PR is ready for some more testing - looks like maybe a rebase first?

joshuatf:
@timmyc I think I'll wait for #1144 to get merged first since there are pending changes there and that will make testing easier.

timmyc:
Thanks @joshuatf! Now that 1144 has merged in this just needs a rebase then another review pass?

joshuatf:
This has been rebased but tests will fail until https://github.com/woocommerce/wc-admin/issues/1263 is resolved.

jeffstieler:
I pushed https://github.com/woocommerce/wc-admin/pull/1180/commits/1e7b673512e0e9ab636068d4241ec5e61245a96d to update the initial hook used to setup the customer update hooks so that they're attached for tests and API requests. The same assertion is failing, but I think it might be a bad expectation, not a bug. If both test customers have 2 orders associated with them, why would they be considered "new" customers?

```
There was 1 failure:

1) WC_Tests_Reports_Orders::test_populate_and_query_multiple_intervals
Product includes, negative filter for 2 product: SELECT
							MAX(date_created) AS datetime_anchor,
							DATE_FORMAT(date_created, '%Y-%m-%d %H') AS time_interval
							, COUNT(*) as orders_count, SUM(num_items_sold) as num_items_sold, SUM(gross_total) AS gross_revenue, SUM(coupon_total) AS coupons, SUM(refund_total) AS refunds, SUM(tax_total) AS taxes, SUM(shipping_total) AS shipping, ( SUM(net_total) - SUM(refund_total) ) AS net_revenue, AVG(num_items_sold) AS avg_items_per_order, ( SUM(net_total) - SUM(refund_total) ) / COUNT(*) AS avg_order_value, ( COUNT( DISTINCT( customer_id ) ) -  COUNT( DISTINCT( CASE WHEN returning_customer = 0 THEN customer_id END ) ) ) AS num_returning_customers, COUNT( DISTINCT( CASE WHEN returning_customer = 0 THEN customer_id END ) ) AS num_new_customers
						FROM
							wptests_wc_order_stats

						WHERE
							1=1
							 AND date_created <= '2019-01-10T16:59:39' AND date_created >= '2019-01-10T16:00:20'
							 AND (  wptests_wc_order_stats.returning_customer = 1 )
						GROUP BY
							time_interval
						ORDER BY
							time_interval DESC
						LIMIT 0,10
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
-        'num_new_customers' => 2
+        'num_new_customers' => 0
@@ @@
-                'num_new_customers' => 2
+                'num_new_customers' => 0
```

joshuatf:
>I pushed 1e7b673 to update the initial hook used to setup the customer update hooks so that they're attached for tests and API requests. 

Nice!  Thanks for this @jeffstieler - that resolved the previous failing tests.

>If both test customers have 2 orders associated with them, why would they be considered "new" customers?

You're right; this is a bad expectation as it should be 0 since no orders are found from this query (I must have fudged this in the rebase).   They are considered `new` customers when you pull a period with their initial order.  If we were to query any period after their initial order, they would be considered `returning`.

tests/reports/class-wc-tests-reports-orders.php
```diff
@@ -371,7 +371,7 @@ public function test_populate_and_query_multiple_intervals() {
 				'net_revenue'             => $net_revenue,
 				'avg_items_per_order'     => $num_items_sold / $orders_count,
 				'avg_order_value'         => $net_revenue / $orders_count,
-				'num_returning_customers' => $orders_count - $new_customers,
+				'num_returning_customers' => 0,
```
jeffstieler: Curious - was `$orders_count - $new_customers` no longer correct? If so, why?

jeffstieler (commented):
This looks good, but I'm struck by the lack of a test where `num_returning_customers` is non-zero. It would be nice to very that case as well.

joshuatf:
@jeffstieler Added a test case for returning customers in https://github.com/woocommerce/wc-admin/pull/1180/commits/8223e84322c4af9e4fa473db5cf317f68dee86e8

tests/reports/class-wc-tests-reports-orders-stats.php
```diff
@@ -806,8 +806,8 @@ public function test_populate_and_query_multiple_intervals() {
 				'net_revenue'             => $net_revenue,
 				'avg_items_per_order'     => $num_items_sold / $orders_count,
 				'avg_order_value'         => $net_revenue / $orders_count,
-				'num_returning_customers' => $orders_count,
-				'num_new_customers'       => 0,
+				'num_returning_customers' => 0,
+				'num_new_customers'       => 2,
```
Aljullu: No need to change that, but I am curious, I noticed in some places we are using the `$new_customers` variable and in some others we are hardcoding `2`. Is there any reason for that?
```suggestion
				'num_new_customers'       => $new_customers,
```

Aljullu (commented):
Tests seem to be failing in Travis but locally work fine for me. Any idea what might be happening here?

tests/reports/class-wc-tests-reports-orders-stats.php
joshuatf: I directly added `2` to most of the areas I changed since some were listed and some weren't.  It was probably used before because the number of calculated new customers was much higher and dependent on number of orders.

Updated all these references to use `$new_customers` so we can quickly change this in the future if needed.

joshuatf:
@Aljullu I'm not entirely sure why the tests were failing.  I suspect there's a time dependent bug happening around the hour change, but can't seem to replicate at the moment.

Updated the `$new_customers` references you mentioned.

joshuatf:
Failing tests issued noted here - https://github.com/woocommerce/wc-admin/issues/1316

Aljullu:
> @Aljullu I'm not entirely sure why the tests were failing. I suspect there's a time dependent bug happening around the hour change, but can't seem to replicate at the moment.

Thank you for taking care of filling a bug for that @joshuatf.

I would say that's ready to merge, not sure if @jeffstieler wants to take a last look before approving it?

tests/reports/class-wc-tests-reports-orders-stats.php
Aljullu: :clap:

jeffstieler (approved):
Looks good! 🛳 it.

peterfabian:
As I'm writing some unit test now, it seems strange how this behaves.
E.g.: Customer_1 created one order at 10:00 and another at 11:30. Now splitting data into time intervals, this customer is considered a new customer in interval for hour 10 and also in the interval for hour 11. 

Just wanted to confirm that this is how we want it to work @joshuatf @LevinMedia

Issue #1156, opened 2018-12-22
Chart: Update X-Axis Logic for Single Date


timmyc:
1. Single days and hourly intervals: When the selected date range is a single day, (ie: a single day picked from the custom date range picker, "today", or "yesterday" in the presets) the default interval should be "Hours"
2. When viewing an entire day with hourly intervals, the x-axis ticks should start at 12am and end at 11pm. 
3. When the time frame is today's date, and set to hourly intervals, the graph only plots up to and including the current hour of the day. 
4. When viewing single day time frames the secondary X axis should always include the month, day and year. 

<img width="975" alt="screen shot 2018-12-14 at 10 05 45 pm" src="https://user-images.githubusercontent.com/4500952/50039772-ac20de00-ffec-11e8-88c6-03971b28fe09.png">

Aljullu:
I created #1195 which fixes 1, 2 and 4. But regarding this one:

> When the time frame is today's date, and set to hourly intervals, the graph only plots up to and including the current hour of the day.

My opinion is that this should be implemented in the API. I think the client component must be 'dumb' and print all data it receives, without checking if it's in the past or future.

Adding a limitation to only display past data would not allow us or other developers to create charts with future data like an _Expected revenue_ chart, for example, which would be a legit chart to implement for an extension.

Thoughts @timmyc?

timmyc:
Excellent point @Aljullu - I think that is the same bug reported in #957 - just a different interval, but should be the same regardless - the API should zero fill in all intervals on both ends of a date range when there is no data.

Aljullu:
@timmyc I wouldn't say #957 is exactly the same issue as this one.

Actually, taking a second look at this issue, I think the fix is simpler than what I originally thought and should be implemented in the client, before we make the request. Currently, when we request the data for today we set the params to:
```
{
  after: 2017-12-27T00:00:00+00:00,
  before: 2017-12-27T23:59:59+00:00, // No matter what time of the day it is, we always request the data until 11:59PM
}
```
It would probably make sense to set the `before` param to the current time, that will make the API to return only the hours up to the current one.

That means no API changes are required. If this solution sounds good to you, I will create a PR with this solution tomorrow.

timmyc:
Ah gotcha - yes that seems like a good solution too 👍

PR #1195, opened 2018-12-27
 Improve Charts X-axis when displaying a single date 

Aljullu:
Fixes part 1, 2 and 4 from #1156.

Implements several improvements to charts when displaying single dates.

### Screenshots
image

- Go to any report and select _Today_ or _Yesterday_ in the date picker.
- Verify the interval changes to _By hour_.
- Verify all hours between 12AM and 23PM appear in the X-axis.
- Make the screen smaller and verify some ticks disappear, so they don't overlap.
- Hover the chart dots and verify the green line shows the hours correctly (before, it was always showing 12PM, as you can see in the screenshot above taken before fixing this).
- Verify the date shown in the first tick is in the format _Dec 12, 2018_.

### DIFF
--- a/packages/components/src/chart/d3chart/test/utils.js
+++ b/packages/components/src/chart/d3chart/test/utils.js
@@ -401,7 +401,7 @@ describe( 'getYTickOffset', () => {
 	} );
 } );
 
-describe( 'getdateSpaces', () => {
+describe( 'getDateSpaces', () => {
 	it( 'return an array used to space out the mouseover rectangles, used for tooltips', () => {
 		const testDateSpaces = getDateSpaces( dummyOrders, testUniqueDates, 100, testXLineScale );
 		expect( testDateSpaces[ 0 ].date ).toEqual( '2018-05-30T00:00:00' );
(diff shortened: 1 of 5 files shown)


packages/components/src/chart/d3chart/test/utils.js
```diff
@@ -157,6 +158,228 @@ describe( 'getXLineScale', () => {
 	} );
 } );
 
+describe( 'getXTicks', () => {
```
timmyc: These tests. 💯 

Thanks so much for taking the time to add some coverage around this logic.

packages/components/src/chart/d3chart/utils.js
```diff
@@ -299,6 +299,21 @@ export const calculateXTicksIncrementFactor = ( uniqueDates, maxTicks ) => {
 	return factors.find( f => uniqueDates.length / f < maxTicks );
 };
 
+/**
+ * Given an array of dates, returns true if the first and last one belong to the same day.
+ * @param {array} dates - an array of dates
+ * @returns {boolean} whether the first and last date are different hours from the same date.
+ */
+const areDatesInTheSameDay = dates => {
```
timmyc: Since this method is working off of UTC - how will this work properly with the site's timezone?

timmyc (approved):
This is testing out great for me, and love the fact that most of the lines in this PR are for tests! I had one question about how timezones will play with the logic - but we could handle that in a follow-up PR too if need be.

packages/components/src/chart/d3chart/utils.js
Aljullu: Good catch. That was a bug, I was manually modifying the date to convert it to UTC and then getting its UTC values. Of course, that shouldn't be needed.

I fixed it here: 000151b488de6e6e64c71b20d1b1c8931bb562aa. It should work exactly the same but code is simpler.

PR #1196, opened 2018-12-27
Allow Search box to break into several lines

### LINKED ISSUE #1091: Search box overflows window width when its contents are too wide
In the _Customers_ report filters, selecting several items makes the search box to overflow the window size:

screenshot from 2018-12-14 14-00-31

Aljullu:
Fixes #1091.

Allows the Search box to break into several lines if there are many tags.

### Accessibility

### Screenshots
input

- Select _Advanced filters_ and add a _Name_ filter.
- Add three or four names.
- Resize the window to make sure the search box breaks into several lines instead of overflowing the screen.

### DIFF
--- a/packages/components/src/filters/advanced/style.scss
+++ b/packages/components/src/filters/advanced/style.scss
@@ -94,7 +94,6 @@
 
 	display: flex;
 	align-items: center;
-	white-space: nowrap;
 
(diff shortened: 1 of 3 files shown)


timmyc (approved):
This is testing out great for me, thanks for fixing this one up Albert

Issue #1192, opened 2018-12-27
Dashboard: Create Top Categories Block


timmyc:
This task involves creating a new "Top Categories" Block for the dashboard leaderboard section:

![image](https://user-images.githubusercontent.com/22080/50460231-1446bc80-0929-11e9-8585-f9a354689e73.png)

The table should show the categories with the most items sold:

Category|Items Sold|G. Revenue
---------|-----------|-----------
Category Name linked Product Listing|# of Products sold from category in period|Revenue of Products sold

@LevinMedia should we link the category name to the product listing for that category `/wp-admin/edit.php?s&post_status=all&post_type=product&action=-1&product_cat=CATEGORY-NAME`

LevinMedia:
@timmyc once filtering lands in the tabular data, it should link to the category report, filtered by the category name. 

<img width="830" alt="screen shot 2018-12-27 at 8 36 46 am" src="https://user-images.githubusercontent.com/4500952/50487323-afd03f80-09b2-11e9-8ad6-26972aa2bfe9.png">

PR #1198, opened 2018-12-28
Dashboard: Add Leaderboard Component

timmyc:
For #1192

This branch introduces a new `<Leaderboard />` component that is heavily inspired by the `<ReportTable />` component. Initially I started to add a prop to re-use all of the table.js logic for a new Categories Leaderboard block, but that seemed to get a bit too complicated - as the tables on the report pages are much more full featured ( filtering, download button, compare checkbox etc ).

So I opted to make a similar component just for Dashboard leaderboards... and here we are!

In this branch I have implemented the `Leaderboard` component, and updated the "Top Selling Products" leaderboard to use the new component. While doing so the following little buggers were fixed:

- Migrated Top Selling Products to use the `wc-api` store / fresh data.
- Fixed the Product name not showing up within the Top Selling Products leaderboard
- Top Selling Products now respects/uses the Date Picker range selected on the Dashboard

### Accessibility

### Screenshots

top-selling-branch

- Open the Dashboard and verify the Top Selling Products leaderboard displays properly ( loading indicator / rendering etc )
- Change the date range using the Date Picker - verify the change triggers a refresh of the Top Selling Products leaderboard


client/dashboard/index.js
```diff
@@ -25,7 +25,7 @@ export default class Dashboard extends Component {
 				<StorePerformance />
 				<div className="woocommerce-dashboard__columns">
 					<div>
-						<TopSellingProducts />
+						<TopSellingProducts query={ query } path={ path } />
```
joshuatf: I'm not seeing `path` as a prop type in top selling products or the leader board component.  Is this used somewhere else?

client/dashboard/leaderboard/index.js
```diff
@@ -0,0 +1,96 @@
...
+import ReportError from 'analytics/components/report-error';
+import { getReportTableData } from 'store/reports/utils';
+import withSelect from 'wc-api/with-select';
+import './style.scss';
+
+export class Leaderboard extends Component {
+	render() {
+		const { getHeadersContent, getRowsContent, isRequesting, isError, items, title } = this.props;
+		const data = get( items, [ 'data' ], [] );
+		const rows = isRequesting || isError ? [] : getRowsContent( data );
```
joshuatf: Is the `isRequesting || isError ? [] : ` portion redundant since we check for errors and requesting below?  Or is there a chance that malformed data could be returned for the rows?

client/dashboard/top-selling-products/index.js
```diff
@@ -86,53 +88,24 @@ export class TopSellingProducts extends Component {
...
-				rows={ rows }
-				rowsPerPage={ 5 }
-				title={ title }
-				totalRows={ 5 }
+			<Leaderboard
+				endpoint="products"
+				getHeadersContent={ this.getHeadersContent }
+				getRowsContent={ this.getRowsContent }
+				query={ this.props.query }
+				tableQuery={ tableQuery }
```
joshuatf: Nice!  Love how clean this feels.

joshuatf (approved):
Looking good, Timmy!  Couple questions about the `path` prop and rows assignment, but otherwise tests well and pre-approving 👍

joshuatf (dismissed):
Approved a bit too early as jest tests are failing.  Added a commit to migrate tests from top selling products into the leaderboard component.  https://github.com/woocommerce/wc-admin/pull/1198/commits/f831083dfed7f352003fb939bb946e03cd769bb0

There's still a failing test with the API that needs to be fixed.

timmyc:
Thanks for the review and contributions @joshuatf - I've skipped the API test for now as I think that might be a bit beyond the scope of this PR right now with moving the data layer to the wc-api flavor of `select`

joshuatf (approved):
Changes look good; also agree that the API tests aren't a necessity right now.  :shipit:

timmyc:
Thanks for merging this in @joshuatf

PR #1199, opened 2018-12-28
 Extract IE11 specific styles to IE stylesheet

### LINKED ISSUE #357: Create IE11 Specific Stylesheet
From discussion in #353 - it appears the amount of css generated from polyfills and fixes that are IE11 specific are generating a substantial amount of css code. As such, this task involves creation of an IE11 stylesheet(s) that should only be loaded for IE11 users.

joshuatf:
Fixes #357 

Adds an IE specific stylesheet for components and extracts existing styles specific to IE11 to individual stylesheets.

Note that this does not extract one-liner fixes that can be applied across different browsers, but instead focuses on larger chunks of code where we should maintain separate styles for IE.

This PR also does not attempt to fix other existing issues in IE.

@Aljullu Are there any other pieces of IE specific styles you think should be pulled into this stylesheet?

1.  Boot up IE11.
2.  Check that the ie.css stylesheet is loading and no styling regressions have occurred between this branch and master.
3.  Check that the ie.css stylesheet is not loading under non-IE browsers.


Aljullu (approved):
> @Aljullu Are there any other pieces of IE specific styles you think should be pulled into this stylesheet?

I agree with you we should only extract big chunks of CSS and keep the small bits in the common CSS files as long as they don't interfere with other browser styles. Having as much as CSS code shared between IE11 and other browsers will make our job easier to detect and fix bugs in the future.

I took a quick look at all PRs with `IE11` in the name and I think you picked all the relevant chunks of code. So it's ready to merge for me.

:shipit:

PR #1201, opened 2018-12-28
Sort default category response by category ID

### LINKED ISSUE #1137: Undefined index for date sorting in category data store
**Describe the bug**
Getting a notice when making calls to the categories API:

```
Notice: Undefined index: date_created in /wp-content/plugins/wc-admin/includes/data-stores/class-wc-admin-reports-categories-data-store.php on line 116
```

**To Reproduce**

1.  Make a call to `/wp-json/wc/v3/reports/categories`
2.  Append `date` as the orderby param.
3.  Note the error.

**Expected behavior**
Not sure if we should use the `date_created` in order by or remove entirely from here since date sorting doesn't seem to need to be an option for the categories table.

I'm leaning towards the latter since this sorting doesn't seem to benefit users and causes performance issues.

joshuatf:
Fixes #1137 

Fixes the warning for no `date_created` field and sets the default sorting to use the category ID if orderby is set to the default value of date.

### Before
screen shot 2018-12-28 at 5 54 52 pm

### After
screen shot 2018-12-28 at 5 51 57 pm

1.  Make a request to the categories endpoint without any orderby param `wp-json/wc/v3/reports/categories`.
2.  Check that no notices/errors are given.
3.  Make a request to the categories endpoint with the orderby param set to `date_created`.
4.  Note the error that this is not a valid field.

### DIFF
--- a/includes/api/class-wc-admin-rest-reports-categories-controller.php
+++ b/includes/api/class-wc-admin-rest-reports-categories-controller.php
@@ -252,9 +252,9 @@ class WC_Admin_REST_Reports_Categories_Controller extends WC_REST_Reports_Contro
 		$params['orderby']       = array(
 			'description'       => __( 'Sort collection by object attribute.', 'wc-admin' ),
 			'type'              => 'string',
-			'default'           => 'date',
+			'default'           => 'category_id',
 			'enum'              => array(
-				'date',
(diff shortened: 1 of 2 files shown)


timmyc (approved):
LGTM and testing well 🙌

PR #1202, opened 2018-12-28
Update chart button active state for dashboard charts

### LINKED ISSUE #1185: Dashboard: Add Active/In-Active Styling to Chart Type
Currently the Charts section in the dashboard does not correctly style the Chart-Type buttons based on the type of charts that are actively being displayed:

image

__Design__

image

joshuatf:
Fixes #1185 

Applies chart button styling to the dashboard charts.

### Before
screen shot 2018-12-28 at 6 10 24 pm

### After
screen shot 2018-12-28 at 6 10 09 pm

1. Visit the dashboard charts `/wp-admin/admin.php?page=wc-admin#/`.
2. Click on a chart type.
3. Check that the active chart type is highlighted correctly.

### DIFF
--- a/client/dashboard/dashboard-charts/index.js
+++ b/client/dashboard/dashboard-charts/index.js
@@ -80,7 +80,7 @@ class DashboardCharts extends Component {
 						>
 							<IconButton
 								className={ classNames( 'woocommerce-chart__type-button', {
-									'woocommerce-chart__type-button-selected': query.type === 'line',
+									'woocommerce-chart__type-button-selected': ! query.type || query.type === 'line',
 								} ) }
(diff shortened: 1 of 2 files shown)


Aljullu (approved):
Looks beautiful. :heart: 

I'm a bit worried about adding a couple of `!important` declarations in the CSS. That might create a specificity war for any third-party developer using this package.

One solution would be increasing our selector specificity with something like `.woocommerce-chart__type-button:not(:disabled):not([aria-disabled="true"]):not(.is-default)`, but I'm not sure that's much better than using `!important`. :slightly_frowning_face: 

So I think we can merge this as is and take a look in the future if we see this is causing us problems.

PR #1203, opened 2018-12-28
Show today charts up to the current hour

### LINKED ISSUE #1156: Chart: Update X-Axis Logic for Single Date

Aljullu:
Fixes the third point and closes #1156.

When displaying a chart with the date range set to _Today_, it will display the dates until the current hour instead of the entire day. To accomplish that, this PR changes the `before` param sent to retrieve chart data to the current hour +1 instead of `23:59:59` until now.

### Screenshots
image

- Go to any report, for example _Orders_.
- In the _Date Range_ dropdown, select _Today_.
- Verify the chart shows data until your current hour +1.
- Verify there are no regressions in other charts.

### DIFF
--- a/client/store/reports/test/utils.js
+++ b/client/store/reports/test/utils.js
@@ -327,7 +327,7 @@ describe( 'getSummaryNumbers()', () => {
 		} );
 
 		setGetReportStats( ( endpoint, _query ) => {
-			if ( '2018-10-10T00:00:00+00:00' === _query.after ) {
+			if ( '2018-10-10T00:00:00' === _query.after ) {
 				return {
 					data: {
 						totals: totals.primary,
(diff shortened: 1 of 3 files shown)


LevinMedia:
Thanks so much for working through this one! 

Outside the scope of this PR but making note: we need to display the date, rather than the hour in the tabular data when viewing by a single day:

screen shot 2018-12-28 at 8 39 31 am

Created a new issue #1205 to track

timmyc (approved):
This is testing well for me 👍

Issue #1187, opened 2018-12-26
Chart: Update Interval Logic


timmyc:
A bit of some shared functionality of #1186 - so could be done here instead of there, but wanted to make sure this was tracked specifically for the charts outside of the dashboard. @LevinMedia would like the following logic used to power the interval selection drop down shown in charts:

**Date range:** 1 day **Intervals:** hours, days
**Date range:** 2 - 6 days **Intervals**: Days
**Date range:** >= 7 days **Intervals**: Days, Weeks
**Date range:** >= 28 days **Intervals**: Days, Weeks, Months
**Date range:** >= 90 days **Intervals**: Days, Weeks, Months, Quarters
**Date range:** >= 365 days **Intervals**: Days, Weeks, Months, Quarters, Years

PR #1208, opened 2018-12-31
Use a date instead of hour for single dates in report tables

### LINKED ISSUE #1205: Use date rather than hour in tabular data when viewing a single day
When viewing a single day (Today, yesterday, or a single custom date) the tabular data shows "12h" in the date column. It should show the date of the selected day:

screen shot 2018-12-28 at 8 39 31 am

joshuatf:
Fixes #1205 

Adds `day` to the allowed list of intervals for periods less than a day so that the date format will parse as a date instead of an hour for single days.

### Before
screen shot 2018-12-31 at 1 25 40 pm

### After
screen shot 2018-12-31 at 1 25 27 pm

1.  Open any reports page with a page (e.g. Revenue).
2.  Select different date ranges with single day periods (Yesterday, Today, and Custom: Same day to same day).
3.  Check that the date in the table shows the date selected and not 12am.

### DIFF
--- a/client/analytics/report/orders/table.js
+++ b/client/analytics/report/orders/table.js
@@ -103,7 +103,7 @@ class OrdersReportTable extends Component {
 
 	getRowsContent( tableData ) {
 		const { query } = this.props;
-		const currentInterval = getIntervalForQuery( query );
+		const currentInterval = getIntervalForQuery( { interval: 'day', ...query } );
 		const { tableFormat } = getDateFormatsForInterval( currentInterval );
 		return map( tableData, row => {
 			const {
--- a/client/analytics/report/orders/table.js
+++ b/client/analytics/report/orders/table.js
@@ -11,7 +11,7 @@ import { map } from 'lodash';
 /**
  * WooCommerce dependencies
  */
-import { appendTimestamp, getCurrentDates } from '@woocommerce/date';
+import { appendTimestamp, defaultTableDateFormat, getCurrentDates } from '@woocommerce/date';
 import { Link, OrderStatus, ViewMoreList } from '@woocommerce/components';
 import { formatCurrency } from '@woocommerce/currency';
 
@@ -126,7 +126,7 @@ class OrdersReportTable extends Component {
 
 			return [
 				{
-					display: formatDate( 'm/d/Y', date ),
+					display: formatDate( defaultTableDateFormat, date ),
 					value: date,
 				},
 				{
--- a/client/analytics/report/orders/table.js
+++ b/client/analytics/report/orders/table.js
@@ -11,12 +11,7 @@ import { map } from 'lodash';
 /**
  * WooCommerce dependencies
  */
-import {
-	appendTimestamp,
-	getCurrentDates,
-	getIntervalForQuery,
-	getDateFormatsForInterval,
-} from '@woocommerce/date';
+import { appendTimestamp, getCurrentDates } from '@woocommerce/date';
 import { Link, OrderStatus, ViewMoreList } from '@woocommerce/components';
 import { formatCurrency } from '@woocommerce/currency';
 
@@ -102,9 +97,6 @@ class OrdersReportTable extends Component {
 	}
 
 	getRowsContent( tableData ) {
-		const { query } = this.props;
-		const currentInterval = getIntervalForQuery( { interval: 'day', ...query } );
-		const { tableFormat } = getDateFormatsForInterval( currentInterval );
 		return map( tableData, row => {
 			const {
 				date,
@@ -134,7 +126,7 @@ class OrdersReportTable extends Component {
 
 			return [
 				{
-					display: formatDate( tableFormat, date ),
+					display: formatDate( 'm/d/Y', date ),
 					value: date,
 				},
 				{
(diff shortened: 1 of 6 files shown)


Aljullu (commented):
This seems to break the chart showing hourly data for single days (#1195 and #1203). :worried: 

I would say the problem here is just about using the [proper format](https://github.com/woocommerce/wc-admin/blob/master/client/analytics/report/revenue/table.js#L105) when displaying date in tables.

Aljullu:
> Maybe I'm misunderstanding the naming of these functions, but I would imagine that overriding the format for hours to show dates might create issues down the road if the table is ever used to show intraday times.

Right, my point was not to use `getDateFormatsForInterval()` there. If the table date format doesn't depend on the interval, it should not be using that function.

However, I see there is an [old comment](https://github.com/woocommerce/wc-admin/blob/master/client/analytics/report/revenue/table.js#L230) which brought me to [this conversation](https://github.com/woocommerce/wc-admin/issues/385#issuecomment-424403038) saying the table should show hourly data when the interval is set to hours. That would invalidate #1205.

I would wait for @LevinMedia to clarify this before doing any further work here.

LevinMedia:
@Aljullu - I caught up with @joshuatf this afternoon - My apologies for the contradictory messages. after further consideration, we'd like to forego the previous discussions you noted, and move forward with the new expectation set by this ticket. 

If one day is selected, the table should show the date as a single row in the tabular data,  and the hours should be left for the chart. It maintains consistency with the other intervals.

joshuatf:
@Aljullu I see what you mean by skipping `getDateFormatsForInterval()`.  That's a better solution if we we never need to use hours in these tables.

My idea was to pass in a default interval of `day` to get the right format for tables unless an interval is specified.  Pushed this is another commit so you can take a look.

`const currentInterval = getIntervalForQuery( { interval: 'day', ...query } );`

@LevinMedia One thing we didn't cover was whether or not hours might ever be used in the tables.  Might there ever be a case where we'd want to show hours in the table (e.g., 12am, 1am, 2am, etc.)?  Or will we always want a date for this column?

client/analytics/report/orders/table.js
```diff
@@ -103,7 +103,7 @@ class OrdersReportTable extends Component {
 
 	getRowsContent( tableData ) {
 		const { query } = this.props;
-		const currentInterval = getIntervalForQuery( query );
+		const currentInterval = getIntervalForQuery( { interval: 'day', ...query } );
```
Aljullu: This doesn't seem to work for me. Maybe this way?
```suggestion
		const currentInterval = getIntervalForQuery( { ...query, interval: 'day' } );
```

Also, this change will have to be made to all other tables that display dates: coupons, customers, downloads and revenue.

Aljullu (commented):
> My idea was to pass in a default interval of day to get the right format for tables unless an interval is specified. Pushed this is another commit so you can take a look.

Oh, I see. That will work too and seems a good solution for me. :+1:

LevinMedia:
> Might there ever be a case where we'd want to show hours in the table (e.g., 12am, 1am, 2am, etc.)? Or will we always want a date for this column?

Anything is possible 🤣 Though for now, I think for now it's safe to assume no.

client/analytics/report/orders/table.js
joshuatf: @Aljullu This seemed to work on all reports for me.  Could you let me know which reports and date range you're seeing it not working?

I don't think we want to swap them since this would always set the interval to `day` and not allow it to be overridden by `query.interval` if set.  If we want it to always show in `mm/dd/yyyy` format and remove the possibility to show hours, I prefer your earlier suggestion of removing `getDateFormatsForInterval()` altogether.

joshuatf: I see now.  I was thinking we could still allow hours in this table if we add intraday date ranges, but manually setting to hourly still produces this error.

We could check if the query period is exactly one day and set the interval to `day` and only use `hour` if it's less than 24 hours, but since @LevinMedia noted that there are no current plans to use hours in these tables, I've added the previous solution you mentioned bypassing `getDateFormatsForInterval()`.

I think you were correct in your original review and results in the cleanest approach instead of trying to make `getDateFormatsForInterval()` fit a situation where it isn't needed.

packages/date/src/index.js
```diff
@@ -367,16 +367,16 @@ export function getAllowedIntervalsForQuery( query ) {
 			allowed = [ 'day', 'week', 'month' ];
 		} else if ( differenceInDays >= 7 ) {
 			allowed = [ 'day', 'week' ];
-		} else if ( differenceInDays > 1 && differenceInDays < 7 ) {
-			allowed = [ 'day' ];
```
Aljullu: According to #1187:

> **Date range:** 2 - 6 days **Intervals**: Days

these two lines shouldn't change.

Aljullu (approved):
It works like a charm! Added a small comment and, after it's addressed, it's ready to be merged.

joshuatf:
@Aljullu Updated to restore that allowed interval period.  Thanks for your review and all your help 👍

PR #1209, opened 2018-12-31
Update dashboard charts on query change

### LINKED ISSUE #1188: Dashboard: Charts Don't Update on Date Picker Change
**Describe the bug**
When viewing the dashboard, if I make a change to the dates via the Date Range Picker, the charts on the page are not updated with the newly selected dates.

**To Reproduce**

1. go to the dashboard, look at the charts shown, note the dates
2. select a different date range using the date picker, i.e. Quarter to Date
3. Note the charts do not update with data for the newly selected period

**Expected behavior**
I expected the charts to update based upon the date picker selection.

joshuatf:
Fixes #1188 

Assigns the query from props to the state query when updated.

1.  Open the dashboard reports `/wp-admin/admin.php?page=wc-admin`.
2.  Change the query params using the date range picker.
3.  Check that the charts are updated when the date range is changed.

### DIFF
--- a/client/dashboard/dashboard-charts/index.js
+++ b/client/dashboard/dashboard-charts/index.js
@@ -5,7 +5,6 @@
 import { __ } from '@wordpress/i18n';
 import classNames from 'classnames';
 import Gridicon from 'gridicons';
-import { isEqual } from 'lodash';
 import { ToggleControl, IconButton, NavigableMenu } from '@wordpress/components';
 import { Component, Fragment } from '@wordpress/element';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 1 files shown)


psealock (commented):
Using the datepicker is working as expected. While we are here, however, I think `query` should be removed from `state` for a few reasons.

All other pages use the url as source of truth, and duplicating `query` in state can cause problems. Already, you can see that the bar/line buttons don't work because they update state, but not the url. The charts update if you manually change the url with `&type=bar`.

I agree with this article and think removing `query` from `state` is a good way to avoid `getDerivedStateFromProps`.

https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html

joshuatf:
@psealock You're right.  Removed this and added temporary `chartType` in state until a PR for https://github.com/woocommerce/wc-admin/issues/1184 lands.

Aljullu (approved):
Much better without `getDerivedStateFromProps`. LGTM. :rocket:

psealock (approved):

Issue #950, opened 2018-11-28
REST API: Create new /v4 of Orders Endpoint


psealock:
Thus far, populating the orders table from the established `/orders` endpoint has been working great. Filtering by Advanced Filters in #943 will introduce query parameters which have no meaning in the context of the endpoint. 

We can keep using `/orders` by filtering client side. For example, an advanced filter of `?status_is_not=pending` can be turned into `?status=...` using `wcSettings.orderStatuses` to explicitly ask for all but pending orders.

`product_includes=65` can also be translated to the correct query argument (`product`), but only accepts a single value. We'll need to filter by multiple ids.

The limitations are clear with query argument `&customers=new`. We can get customer data via `/customer` and compare results with `customer_id` from the order, but that won't give an indication of new/returning.

Likewise, we can't filter by `coupon` in the `/orders` endpoint.

Seems to me like we'll need to write `/reports/orders`. Do you agree @peterfabian ?

peterfabian:
I think you're right. Maybe we could add required filters to `/orders` endpoint? 

Because otherwise, `/reports/orders` would return the same data (or a subset of data from `orders`), and would just add support for extra filters, if I understand the problem correctly?

@claudiosanches what is your view on this? Do you have any preference regarding new endpoint/updating `/orders`?

claudiosanches:
@claudiosanches I happy to introduce new endpoints for reports.
When possible, it's always great to introduce new filters into the current endpoints, still most of filters here probably more related to reports, so I don't see any problem having a new endpoint, also probably better that we don't need to list so many data at once too, since our reports displays less data and having the IDs it's possible to query the regular endpoint for more details.

It's nice that we are going back to our initial idea, where we introduce all fields required by our new reporting system, and doing less requests :smiley:

timmyc:
Seems like if we want to fully support the advanced filter query args, we will likely need to version the `orders` endpoint to `/v4` - or as suggested above we can try to use [this filter](https://github.com/woocommerce/woocommerce/blob/master/includes/api/v2/class-wc-rest-orders-v2-controller.php#L418) to manipulate the requisite query args:

Advanced Filter Arg | /wc/v3 Argument |
--------------------|-------------------|
`status_is (string)`|`status (string)`
`status_is_not (string)`| NA
`product_includes (int|array)`|`product (int)`
`product_excludes (int|array)`| NA
`coupon_includes (int|array)`| NA
`coupon_excludes (int|array)`| NA
[...]

PR #1211, opened 2018-12-31
Add the order status column to lookup table

### LINKED ISSUE #475: Add order_status column to lookup tables.
To prevent unnecessary joining with `wp_posts` table each time, I think it would help to have `order_status` column in the lookup tables.

Also, think about how to make the UX working for selecting order statuses... should it be a setting somewhere or should it be an always visible JS UI element with pre-selected defaults?

joshuatf:
Fixes #475 

Adds the `status` column to the `wc_order_stats` lookup table.  And changes filtering to reflect this in the order stats endpoint.

1.  Rebuild order reports so that the `status` column is populated with data `/wp-admin/admin.php?page=wc-status&tab=tools`.
2.  Visit the Order report page `/wp-admin/admin.php?page=wc-admin#/analytics/orders`.
3.  Under Advanced Filters, test the Order Status option using both `Is` and `Is not` options.
4.  Check that the stats (**not the orders table**), update with the correct order stats.  

Note that the existing order data store will need to be updated in order to work with status filters ( #950  ) and that only a select few statuses are currently recorded in the orders table ( `completed`, `processing`, `on-hold` ).

### DIFF
--- a/includes/class-wc-admin-api-init.php
+++ b/includes/class-wc-admin-api-init.php
@@ -401,6 +401,7 @@ class WC_Admin_Api_Init {
 			shipping_total double DEFAULT 0 NOT NULL,
 			net_total double DEFAULT 0 NOT NULL,
 			returning_customer boolean DEFAULT 0 NOT NULL,
+			status varchar(200) NOT NULL,
 			PRIMARY KEY (order_id),
 			KEY date_created (date_created)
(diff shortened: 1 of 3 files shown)


joshuatf:
@psealock There was an error being thrown in the background process.  Should be resolved now- can you give it another look and let me know if it works for you?

PR #1213, opened 2019-01-01
Fix chart legend checkboxes not reacting to clicks

Aljullu:
This PR fixes a regression that made it impossible to enable/disable chart legend items in report charts.

### Accessibility

### Screenshots
legend

- Go to the _Revenue_ report (or any other).
- Try deactivating and activating again legend items clicking on their name or checkbox.
- Verify the chart updates to hide/show them accordingly.
- Go to the _Dashboard_ and look for a chart.
- Try deactivating legend items and verify they can't be disabled.

### DIFF
--- a/packages/components/src/chart/index.js
+++ b/packages/components/src/chart/index.js
@@ -114,7 +114,7 @@ class Chart extends Component {
 
 	handleLegendToggle( event ) {
 		const { data, mode } = this.props;
-		if ( mode ) {
+		if ( mode === 'block' ) {
 			return;
 		}
 		const orderedKeys = this.state.orderedKeys.map( d => ( {


psealock (approved):
LGTM. Now that `block` has been introduced as a value for the `mode` prop, one sentence about it in the PropType definition comment would be helpful.

joshuatf (approved):
Testing well and LGTM.  Reran failing tests and everything passed.  I think they might have been due to the issue @psealock outlined in https://github.com/woocommerce/wc-admin/issues/1218.

Aljullu:
> Now that `block` has been introduced as a value for the `mode` prop, one sentence about it in the PropType definition comment would be helpful.

Right, actually I'm not sure `mode` is the best prop to use here. `mode` is used to know whether we are comparing items or time periods, which is needed for screen readers to properly read the data.

I think it would make sense to remove `mode=block` and instead create two new single-purpose props:
* `interactive` (default=true): which will be used to know if legend items must be clickable.
* `showHeaderControls` (default=true): will be used to show or hide the header controls.

How does it sound @psealock? I can take care of refactoring this but don't want to do it before knowing a second opinion. :slightly_smiling_face: 

> Testing well and LGTM. Reran failing tests and everything passed. I think they might have been due to the issue @psealock outlined in #1218.

Thanks!

psealock:
>I think it would make sense to remove mode=block and instead create two new single-purpose props:

Oh yeah, absolutely. It makes sense that a chart can be `block` and be either mode.

Is there a need to split the `block` functionality or can `interactive` handle both legend items and header controls?

Aljullu:
> Is there a need to split the `block` functionality or can `interactive` handle both legend items and header controls?

My point for splitting it into two different props is that third-party developers might want to add interactive charts without header controls. For example, a report showing two charts side by side in which they must be interactive, but they are controlled by common controls.

Or the other way around, I can also imagine a use case with charts displaying the header controls but the legend not being interactive.

Actually, probably a better name for that prop would be `interactiveLegend` instead of `interactive`.

---

So I will proceed merging this PR and work on the refactor in a follow-up.

PR #1216, opened 2019-01-01
DateRange: use DayOfWeek from moment locale

psealock:
Remove explicit `firstDayOfWeek` prop from `react-dates` in favour of relying on `moment.localeData`. 

Gutenberg does this and it makes sense to align how we derive these types of settings.

1. Go to _Settings_ and change "Week Starts On" to something new.
2. In the console confirm moment's `localeData` reflects your chosen day of week, `moment.localeData()._week.dow`.
3. Confirm the Custom DatePicker also reflects your chosen day of week by starting the week on that day.

### DIFF
--- a/lib/client-assets.php
+++ b/lib/client-assets.php
@@ -180,9 +180,6 @@ function wc_admin_print_script_settings() {
 		'embedBreadcrumbs' => wc_admin_get_embed_breadcrumbs(),
 		'siteLocale'       => esc_attr( get_bloginfo( 'language' ) ),
 		'currency'         => wc_admin_currency_settings(),
-		'date'             => array(
-			'dow' => get_option( 'start_of_week', 0 ),
-		),
 		'orderStatuses'    => format_order_statuses( wc_get_order_statuses() ),
 		'stockStatuses'    => wc_get_product_stock_status_options(),
 		'siteTitle'        => get_bloginfo( 'name' ),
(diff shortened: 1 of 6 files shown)


Aljullu (commented):
This is testing well for me, however I noticed [we use](https://github.com/woocommerce/wc-admin/blob/master/packages/date/src/index.js#L497) `wcSettings.date` in `packages/date/src/index.js`, which now will be `undefined`. What consequences will have that? Maybe that bunch of code could be removed?

psealock:
Nice find @Aljullu, thank you! 

I went ahead and changed the `loadLocaleData` function to respect what Gutenberg uses in every way for all locales except for 2 properties. 

`longDateFormat` is added by Gutenberg to use PHP style formats, ie `LLL: "F j, Y g:i a`, which moment can't understand. So I override these in favour of translated Moment strings.

`weekdaysMin` is provided by settings, so I use those.

### Test

1. Run unit tests
2. Confirm `moment.localeData()._longDateFormat` is js moment strings, not PHP
3. See the day labels in the Custom calendar correctly rendered as 3 letter abbreviations instead of 2 (for english, at least).

Would you mind taking this for another spin? Thanks

Aljullu (approved):
LGTM. Just a small comment, but it's ready to be merged. :rocket:

PR #1223, opened 2019-01-02
Fix/search css tags

psealock:
Search tags that were not `inlineTags` had incorrect styles applied to them.

**Designs:** p6riRB-3lz-p2

### Before

screen shot 2019-01-02 at 2 41 44 pm

### After

screen shot 2019-01-02 at 2 55 28 pm

1. Go to Products Report > Show > Product Comparison.
2. Add a few products to Search and see that styles are correctly applied.
3. Go to Orders > Show > Advanced Filters > Add a Filter > Products.
4. Add a few products to Search and see that styles are correctly applied for inline rendered tags.

### DIFF
--- a/packages/components/src/search/index.js
+++ b/packages/components/src/search/index.js
@@ -158,7 +158,9 @@ class Search extends Component {
 		const shouldRenderTags = this.shouldRenderTags();
 
 		return (
-			<div className={ classnames( 'woocommerce-search', className ) }>
+			<div className={ classnames( 'woocommerce-search', className, {
+				'has-inline-tags': inlineTags,
+			} ) }>
 				<Gridicon className="woocommerce-search__icon" icon="search" size={ 18 } />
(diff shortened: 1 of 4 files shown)


packages/components/src/search/index.js
```diff
@@ -158,7 +158,9 @@ class Search extends Component {
 		const shouldRenderTags = this.shouldRenderTags();
 
 		return (
-			<div className={ classnames( 'woocommerce-search', className ) }>
+			<div className={ classnames( 'woocommerce-search', className, {
+				'is-inline-tags': inlineTags,
```
joshuatf: Very nitpicky and semantic, but should we name this `has-inline-tags`?

joshuatf (approved):
Nice job, Paul!  Looks so much better.  One small question about class naming, but otherwise LGTM.

PR #1225, opened 2019-01-02
Add dashboard chart user preferences

### LINKED ISSUE #1184: Dashboard: Persist Chart Selections
Currently the Dashboard Chart section allows users to toggle on/off which charts are shown - but these selections are only stored in component state, and are reset upon reload.

image

This issue involves the following items:
- [ ] Persist the chart selections using the `updateCurrentUserData` [mutation](https://github.com/woocommerce/wc-admin/blob/master/client/wc-api/user/mutations.js#L3)
- [ ] Load the persisted chart selections, and display the correct charts on load
- [ ] Persist the chart type selection ( bar vs line ) default to line
- [ ] Load the persisted chart type on load

joshuatf:
Fixes #1184 

Persists shown dashboard charts and chart type for users.  We store hidden charts much like we store hidden columns in report tables as opposed to the shown charts to avoid always hiding newly introduced charts.

### Screenshots
screen shot 2019-01-02 at 5 48 48 pm

1.  Visit the dashboard `wp-admin/admin.php?page=wc-admin`.
2.  Turn on/off some charts.
3.  Refresh the page and make sure that previously hidden charts remain hidden.
4.  Change to a bar chart by clicking the bar chart icon to the right of the section header.
5.  Refresh the page and make sure that the bar chart is still shown.

### DIFF
--- a/client/dashboard/dashboard-charts/index.js
+++ b/client/dashboard/dashboard-charts/index.js
@@ -42,14 +42,14 @@ class DashboardCharts extends Component {
 		userPrefChartType: prevUserPrefChartType,
 	} ) {
 		const { userPrefCharts, userPrefChartType } = this.props;
-		if ( ! isEqual( userPrefCharts, prevUserPrefCharts ) ) {
+		if ( userPrefCharts && ! isEqual( userPrefCharts, prevUserPrefCharts ) ) {
 			/* eslint-disable react/no-did-update-set-state */
 			this.setState( {
 				hiddenChartKeys: userPrefCharts,
 			} );
 			/* eslint-enable react/no-did-update-set-state */
 		}
-		if ( userPrefChartType !== prevUserPrefChartType ) {
+		if ( userPrefChartType && userPrefChartType !== prevUserPrefChartType ) {
 			/* eslint-disable react/no-did-update-set-state */
 			this.setState( {
 				chartType: userPrefChartType,
--- a/client/dashboard/dashboard-charts/index.js
+++ b/client/dashboard/dashboard-charts/index.js
@@ -4,10 +4,13 @@
  */
 import { __ } from '@wordpress/i18n';
 import classNames from 'classnames';
-import Gridicon from 'gridicons';
-import { ToggleControl, IconButton, NavigableMenu } from '@wordpress/components';
 import { Component, Fragment } from '@wordpress/element';
+import { compose } from '@wordpress/compose';
+import Gridicon from 'gridicons';
+import { isEqual } from 'lodash';
 import PropTypes from 'prop-types';
+import { ToggleControl, IconButton, NavigableMenu } from '@wordpress/components';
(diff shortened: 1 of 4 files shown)


psealock (commented):
Getting an error here. Since I have no hidden charts, it may be because there is no default `[]` for `hiddenChartKeys`?

screen shot 2019-01-03 at 1 19 07 pm

joshuatf:
@psealock Thanks for catching that.  There is a default, but I made a mistake by assigning user preferences that were an empty string after mount once they're retrieved from the API.  Added a check to make sure they're set before assigning to state.

This one has been rebased and should be ready for another test.

client/dashboard/dashboard-charts/index.js
```diff
@@ -18,47 +21,85 @@ import { EllipsisMenu, MenuItem, SectionHeader } from '@woocommerce/components';
...
+			this.setState(
+				prevState => {
+					const hidden = prevState.hiddenChartKeys.includes( key );
+					const hiddenChartKeys = prevState.hiddenChartKeys.filter( chartKey => chartKey !== key );
+
+					if ( ! hidden ) {
+						hiddenChartKeys.push( key );
+					}
+					return { hiddenChartKeys };
+				},
```
Aljullu: There is a [lodash function](https://lodash.com/docs/4.17.11#xor) that implements exactly this.
```ES6
return { hiddenChartKeys: xor( prevState.hiddenChartKeys, [ key ] ) };
```

client/dashboard/dashboard-charts/index.js
```diff
@@ -18,47 +21,85 @@ import { EllipsisMenu, MenuItem, SectionHeader } from '@woocommerce/components';
...
 
 	handleTypeToggle( type ) {
 		return () => {
-			this.setState( {
-				chartType: type,
+			this.setState( { chartType: type }, () => {
+				const userDataFields = {
+					[ 'dashboard_chart_type' ]: this.state.chartType,
+				};
+				this.props.updateCurrentUserData( userDataFields );
```
Aljullu: Is there any reason to do this after the state has been updated? Given that `updateCurrentUserData` and `setState` are async, couldn't we do both of them in parallel?
 
```ES6
this.setState( { chartType: type };
const userDataFields = {
	[ 'dashboard_chart_type' ]: type,
};
```

Aljullu (approved):
Works as expected and code looks good. I added a couple of questions regarding the code, but it's ready to be merged. :rocket:

client/dashboard/dashboard-charts/index.js
joshuatf: This is awesome @Aljullu!  Thanks for the tip; had no idea.

client/dashboard/dashboard-charts/index.js
joshuatf: Since these are both toggle dependent, I don't see any harm in running them in parallel.

PR #1226, opened 2019-01-02
Fix bar chart in narrow viewports

### LINKED ISSUE #1190: Chart: Can Not Plot >88 Days in Narrow Viewport
via #1100 

Can not plot more than 88 days in bar chart mode on mobile viewports (<=375px wide)

<=88 days:

screen shot 2018-12-14 at 11 10 52 pm

>=89 days:

screen shot 2018-12-14 at 11 10 38 pm

This problem exists at other viewport / date range combinations and it likely has to do with the number of bars we're trying to render in such a small space. We'll likely have to find some kind of broad solution to address

Aljullu:
Fixes #1190.

The bar chart was empty when there were many bars to be displayed and the viewport was too narrow.

By default, each date took at least 3 pixels: 1 for each bar and 1 as a padding between dates. When displaying an entire year, that meant that if the available space was less than 1095px (365*3), each bar would have less than a pixel, so they were not visible because they were 0 pixels wide.

This PR fixes that with two approaches:
* Eliminates the padding between dates when the available space is too narrow. When displaying an entire year, that will be 730px-1095px.
* When the available space is so narrow that removing the padding is not enough, allows the chart to take the minimum needed space and makes it scrollable.

### Screenshots
image

- Go to the _Revenue_ report or any other.
- Select a wide date range (for example, _Last year_).
- Select bars as the chart type.
- Make the viewport wider and narrower and verify bars are always visible.
- Make sure the tooltip is shown in the correct position when hovering the bars.


LevinMedia:
This is off to a great start, thank you! I'm not sure if it's within the scope of this PR, but the bug seems to have gotten more significant in this branch: 

The bar chart tends to be narrower than the full width of the chartable area. 

screen shot 2019-01-02 at 12 55 13 pm

I created a note to track it in the sweep issue for charts, however there hasn't been an issue assigned to address it just yet. https://github.com/woocommerce/wc-admin/issues/1100#issuecomment-449151827

packages/components/src/chart/d3chart/chart.js
```diff
@@ -89,6 +89,24 @@ class D3Chart extends Component {
...
+		return type === 'bar' && widthWithoutMargins < minimumWideWidth;
+	}
+
+	getWidth() {
+		const {	data, margin, type, width } = this.props;
+		const widthWithoutMargins = width - margin.left - margin.right;
+		const columnsPerDate = data && data.length ? Object.keys( data[ 0 ] ).length - 1 : 0;
+		const minimumWidth = this.shouldBeCompact() ? data.length * columnsPerDate : data.length * ( columnsPerDate + 1 );
+
+		return type === 'bar' ? Math.max( widthWithoutMargins, minimumWidth ) : widthWithoutMargins;
```
psealock: Since these functions get called so much, could it be worth it to exit early with `widthWithoutMargins` for bar charts. Same above with `shouldBeCompact`.

psealock (approved):
Nice work, this is a complicated one.

I can't reproduce what David has above, but it looks good to me.

One improvement I can think of is when scrolling horizontally, the tooltip doesn't update. That would be nice, but perhaps its not needed because tooltips are activated on tap for mobile.

packages/components/src/chart/d3chart/chart.js
Aljullu: Good idea! Implemented here: e782db0147d9b14cd29bc1977164c37b6060b7c6.

Aljullu:
> One improvement I can think of is when scrolling horizontally, the tooltip doesn't update. That would be nice, but perhaps its not needed because tooltips are activated on tap for mobile.

That's a bit tricky. [I made the tooltip to hide](a1a6923528f4dc4d452ee1a19b766fa1969b1a90) when scrolling, I think that's good enough for now. If we want the tooltip to scroll with the chart, we can open an issue for that in the future.

LevinMedia:
>  I would definitely leave that for another PR if that sounds ok.

No worries! Thank you!

PR #1227, opened 2019-01-02
Add margin to inline search tags

Aljullu:
This PR fixes a regression that removed the margins between inline search tags.

### Screenshots
_Before:_
image

_After:_
image

- Go to the _Customers_ report, select _Advanced filters_ and add a _Name_ filter.
- Add enough names to make them break in several lines.
- Verify there is a margin between tags.
- Go to the _Products_ report, select _Product Comparison_ and add a couple of products.
- Verify there is a margin between tags (= there is no regression).

### DIFF
--- a/packages/components/src/search/style.scss
+++ b/packages/components/src/search/style.scss
@@ -10,18 +10,16 @@
 		fill: $core-grey-light-900;
 	}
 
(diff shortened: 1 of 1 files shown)


psealock (approved):
Cool, much better. Thanks

PR #1229, opened 2019-01-02
Fix date issues introduced in #1203

### LINKED ISSUE #1218: Date: Unit Tests fail 11pm UTC
**Describe the bug**
Unit tests fail at 11pm UTC on Travis for one hour. 

**To Reproduce**
I haven't tried, but I believe you can replicate locally by running unit tests at 11pm local time.

**Cause**

`getCurrentPeriod` adds one hour to `primaryEnd`, which is the next day at 11pm.

https://github.com/woocommerce/wc-admin/blob/8494e21bfa7f5463106c4c47c95e6b79396d6b86/packages/date/src/index.js#L175

**Additional context**

We can easily fix the unit tests to accomodate this, however, some investigation might reveal a problem with the UI when viewing reports late at night.

Aljullu:
Fixes #1218 and some other issues introduced in #1203.

This PR undo's the changes introduced in #1203 and tries a simpler approach to display _Today_ charts until the current time.

### Screenshots
image

- Go to any report, for example _Revenue_.
- In the _Date Range_ dropdown, select _Today_.
- Verify the chart shows data until your current hour.
- Verify there are no regressions in other charts.

---

- Go to any report, for example _Revenue_.
- In the _Date Range_ dropdown, select a custom date, for example `2018/01/02`.
- Verify data is loaded properly.

### DIFF
--- a/client/store/reports/utils.js
+++ b/client/store/reports/utils.js
@@ -111,6 +111,24 @@ export function isReportDataEmpty( report ) {
 }
 
 /**
+ * Returns true if `date` is today.
+ *
+ * @param  {String|Object} date Date string or object in a format that can be
+ * understood by the Date JS object. For example `yyyy-mm-dd`.
+ * @returns {Boolean} Whether `date` is today.
+ */
+function isToday( date ) {
+	const d = new Date( date );
+	const today = new Date();
+
+	return (
+		d.getFullYear() === today.getFullYear() &&
+		d.getMonth() === today.getMonth() &&
+		d.getDate() === today.getDate()
+	);
+}
+
+/**
  * Constructs and returns a query associated with a Report data request.
  *
  * @param  {String} endpoint Report API Endpoint
@@ -122,12 +140,13 @@ function getRequestQuery( endpoint, dataType, query ) {
 	const datesFromQuery = getCurrentDates( query );
 	const interval = getIntervalForQuery( query );
 	const filterQuery = getFilterQuery( endpoint, query );
+	const endingTimeOfDay = isToday( datesFromQuery[ dataType ].before ) ? 'now' : 'end';
 	return {
 		order: 'asc',
 		interval,
 		per_page: MAX_PER_PAGE,
 		after: appendTimestamp( datesFromQuery[ dataType ].after, 'start' ),
-		before: appendTimestamp( datesFromQuery[ dataType ].before, 'end' ),
+		before: appendTimestamp( datesFromQuery[ dataType ].before, endingTimeOfDay ),
 		...filterQuery,
 	};
 }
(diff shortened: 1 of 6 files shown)


client/store/reports/utils.js
```diff
@@ -110,6 +110,24 @@ export function isReportDataEmpty( report ) {
...
+	const d = new Date( date );
+	const today = new Date();
+
+	return (
+		d.getFullYear() === today.getFullYear() &&
+		d.getMonth() === today.getMonth() &&
+		d.getDate() === today.getDate()
+	);
+}
+
```
psealock: Should this go in `/packages/date`? If so, it would need some testing.

Alternatively, below in line 143 you could keep it all in moment.js land. `getCurrentDates` always returns a Moment object for before and after:

https://github.com/woocommerce/wc-admin/blob/8b6143b79f8715b7aa19446b05d0b11d988503db/packages/date/src/index.js#L23-L24

```js
const end = datesFromQuery[ dataType ].before;
const endingTimeOfDay = end.isSame( moment(), 'day' ) ? 'now' : 'end';
```

psealock (approved):
Looking good @Aljullu ! Thanks for the fix.

Testing well. Just one comment and a possible alternative for `isToday`.

client/store/reports/utils.js
Aljullu: > Alternatively, below in line 143 you could keep it all in moment.js land.

I like this solution! However, I noticed `getCurrentDates` was not returning Moment objects but string dates. I [fixed](https://github.com/woocommerce/wc-admin/pull/1229/commits/53a1640ef1ae2f936089e5cbaeaf19ff67f14c15) that and updated `appendTimestamp` to accept Moment objects so everything keeps working as usual.

I know you pre-aproved this, but since the changes were bigger than expected, you might want to take another look at this PR. :slightly_smiling_face:

psealock (approved):
@Aljullu Nice... I missed that. Looks like I need to update the jsDoc typedef.

This all looks good and you've handled all the use cases of `getCurrentDates`. 

Small thing, but since the usage of `appendTimestamp` has changed, what do you think about renaming it to `getTimestamp` and only accepting Moment objects?

Aljullu:
Thanks for reviewing it again @psealock!

> what do you think about renaming it to `getTimestamp`

What's the reasoning behind that? To me, a function named `getTimestamp` would return the timestamp but would not take `date` as a param. In the other hand, `appendTimestamp` does that: takes a `date` and returns it with an appended timestamp.

I'm ok with renaming the function and changing its implementation. But I'm not sure `getTimestamp` really describes what it is currently doing without any changes. What do you think?

> and only accepting Moment objects

Sounds good! Changed in this commit: d8d1ca4ad0d091eb3ebb7181320e2d3b3eddcbaf.

PR #1230, opened 2019-01-03
Add dashboard chart interval preferences

### LINKED ISSUE #1186: Dashboard: Add Interval Drop-Down to Chart Section
The Chart section in the dashboard calls for an interval drop down to be shown which controls the interval value for all the charts:

wc-admin dashboard analytics preview mode - invision 2018-12-26 11-47-55

This task involves creating that interval drop-down, and persisting the selected value using the `updateCurrentUserData` mutation.

Also the following logic should be used for controlling which interval options are displayed here based upon the dates selected in the date picker in the dashboard:

**Date range:** 1 day **Intervals:** hours, days
**Date range:** 2 - 6 days **Intervals**: Days
**Date range:** >= 7 days **Intervals**: Days, Weeks
**Date range:** >= 28 days **Intervals**: Days, Weeks, Months
**Date range:** >= 90 days **Intervals**: Days, Weeks, Months, Quarters
**Date range:** >= 365 days **Intervals**: Days, Weeks, Months, Quarters, Years

We might want to make that logic a util method which can be used both in the chart, and in the dashboard here.

joshuatf:
Fixes #1186

Adds an interval selector to dashboard charts and saves them to user preferences to be persisted when accessing the dashboard in the future.

### Screenshots
screen shot 2019-01-03 at 3 52 10 pm

1.  Visit the Dashboard.
2.  Make sure you've select a date range or preset long enough for multiple intervals.
3.  Check that the interval dropdown adjusts based on the date range selected (try yesterday or today to limit to only hourly).
4.  Change the interval to another type and check that the chart updates accordingly.
5.  Check that on refresh the chart interval is persisted.

### Questions
* The persisted interval isn't immediately loaded resulting in a secondary load and flash as the charts refresh.  Do we want to delay charts from showing while user preferences are being requested or is it okay as is?
* The interval selector positioning on mobile is a bit hacky with position absolute.  This should not cause issues with overlapping, but would creating another prop for `SectionHeader` be a better idea so we can restructure those elements?  I'm concerned that adds extra surface area to a component for a single use case.


client/dashboard/dashboard-charts/index.js
```diff
@@ -109,14 +119,56 @@ class DashboardCharts extends Component {
...
+			week: __( 'By week', 'wc-admin' ),
+			month: __( 'By month', 'wc-admin' ),
+			quarter: __( 'By quarter', 'wc-admin' ),
+			year: __( 'By year', 'wc-admin' ),
+		};
+
+		return (
+			<SelectControl
+				className="woocommerce-chart__interval-select"
+				value={ this.state.interval }
```
jeffstieler: Should we check if the interval in state is in the allowed intervals list? Right now this relies on `ReportChart`'s use of `getIntervalForQuery()` to pass the proper interval to the `Chart` component.

client/dashboard/dashboard-charts/index.js
```diff
@@ -109,14 +119,56 @@ class DashboardCharts extends Component {
 		);
 	}
 
+	renderIntervalSelector() {
+		const allowedIntervals = getAllowedIntervalsForQuery( this.props.query );
+		if ( ! allowedIntervals || allowedIntervals.length < 1 ) {
```
jeffstieler: Should we render the interval selector if there's only one option?

jeffstieler (commented):
Just some questions about handling the `<select>` value. This tests well and code looks good otherwise.

client/dashboard/dashboard-charts/index.js
joshuatf: I like the current method because the report chart does filter out intervals that aren't allowed, but we can keep a user's preferences even if the current list does not allow it.

For example, if they select 'Quarter' for the interval and switch to a shorter time period that doesn't allow quarterly charting, they still see the chart without issue and when they switch back to a period that allows quarters, their interval will be switched back to quarters.

client/dashboard/dashboard-charts/index.js
joshuatf: Good question. 🤔   I know we still continue to show it in other charts, so maybe we should remove it across all charts if there is only 1 interval option.  However, it may just act as an information cue to users, but not entirely sure.  @LevinMedia can you weigh in here?

LevinMedia: @jeffstieler @joshuatf For consistency's sake I think we should. Thanks!

LevinMedia: I am noticing a strange behavior when I refresh the page - all of the available chart are appearing for a spit second. Some finish loading, others don't then they disappear. 

screen shot 2019-01-07 at 9 10 49 pm

jeffstieler (approved):

PR #1231, opened 2019-01-03
Split `d3chart/utils.js` into several files

Aljullu:
The file `d3chart/utils.js` was 848 lines long and was the largest one in the entire project, that made working with it quite hard.

This PR splits that file (and the test file associated with it) into several other files. There shouldn't be any functionality change, if there is, that's a bug.

Number of lines for each file after this PR:
image

- Navigate around, play with charts and verify there are no regressions.


psealock (approved):
Builds without errors, tests pass, renders without error and much easier to work with. Thanks!

timmyc:
🎉 thanks for taking the time to do a bit of cleanup here @Aljullu

PR #1233, opened 2019-01-04
Fix/dev docs file path

psealock:
Individual devDocs pages aren't working (`/wp-admin/admin.php?page=wc-admin#/devdocs/*`) due to an incorrect path. This PR updates the path to the correct `docs/componenents/[packages|analytics]` folder.

I added an accepted property `isAnalyticsComponent` in the `examples.json` config, but am open to a more elegant solution.

### DIFF
--- a/client/devdocs/docs.js
+++ b/client/devdocs/docs.js
@@ -18,7 +18,8 @@ class Docs extends Component {
 	}
 
 	getReadme() {
-		const readme = require( `../../docs/components/${ this.props.filePath }.md` );
+		const { filePath, docPath } = this.props;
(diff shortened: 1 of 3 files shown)


joshuatf (commented):
Looking good so far.  Out of curiosity, will there ever be components outside of `analytics` and `packages`?  If so, one idea might be to pass a `docPath` property with `analytics` in examples.json that defaults to `packages` if none is set.

psealock:
>If so, one idea might be to pass a docPath property with analytics in examples.json that defaults to packages if none is set.

Good idea, thanks. I've done just that.

timmyc (approved):
Looking good and testing well for me.

PR #1234, opened 2019-01-04
Add leaderboard section

### LINKED ISSUE #1183: Dashboard: Create Leaderboards Section
The dashboard designs call for a Leaderboards section to be displayed below the charts:

image

This issue involves creation of the Leaderboard `<SectionHeader />`, along with the pictured `< EllipsisMenu />` for the three different leaderboard options, and the input for "Rows Per Table". Since we are not going to support customizing sections in v1, we don't need to add in the "Remove Block" to the menu at this time.

All selections in the Ellipsis Menu though should persist between page loads - using the same logic that is used to store the toggling of table columns in reports.

joshuatf:
Fixes #1183 

This PR adds:
* The section header for Leaderboards
* The ellipsis dropdown menu with leaderboard toggles and rows per table option
* User preference persistence for selected leaderboards and rows per table
* Moves the `Leaderboard` to components and creates a separate `Leaderboards` section.

### Accessibility

### Screenshots
screen shot 2019-01-04 at 3 05 59 pm

1.  Open the dashboard and note the addition of the leaderboard section.
2.  Try toggling the visible leaderboards in the ellipsis menu.  Currently the only one that has been added is "Top Products."
3.  Check that the Top Products leaderboard remains on/off based on your selected preferences after refreshing.
4.  Check that changing rows per table changes the number of visible rows in the leaderboard.
5.  Check that refreshing the page persists rows per table setting.

### Questions
* Same issue as #1230 with user preferences.  User preferences are retrieved async which results in a small flash as the chart updates after loading.  Should we wait until user preferences are retrieved or is this okay?
* Do we want to change the per page options?  Currently set to 1-20. /cc @LevinMedia


jeffstieler:
In testing the functionality, I noticed that there seems to be a flash of default values for the Top Selling Products leaderboard:



(After a refresh, you can see the initial load of 5 rows before a new query is issued that shows 4 - the preference I set)

Will check out the code now.

client/dashboard/leaderboards/index.js
```diff
@@ -0,0 +1,184 @@
...
+		];
+		return (
+			<EllipsisMenu label={ __( 'Choose which leaderboards to display', 'wc-admin' ) }>
+				<Fragment>
+					<MenuTitle>{ __( 'Leaderboards', 'wc-admin' ) }</MenuTitle>
+					{ allLeaderboards.map( leaderboard => {
+						return (
+							<MenuItem onInvoke={ this.toggle( leaderboard.key ) } key={ leaderboard.key }>
+								<ToggleControl
+									label={ __( `${ leaderboard.label }`, 'wc-admin' ) }
```
jeffstieler: The leaderboard labels are already passed through i18n.

client/dashboard/leaderboards/index.js
```diff
@@ -0,0 +1,184 @@
...
+			userPrefLeaderboardRows &&
+			parseInt( userPrefLeaderboardRows ) !== parseInt( prevUserPrefLeaderboardRows )
+		) {
+			/* eslint-disable react/no-did-update-set-state */
+			this.setState( {
+				rowsPerTable: parseInt( userPrefLeaderboardRows ),
+			} );
+			/* eslint-enable react/no-did-update-set-state */
+		}
+	}
```
jeffstieler: This seems really complex to effectively copy props into state (and the `eslint-disable`s smell)..

Is there a different way to manage the components state and still perform the side effects of updating user preferences without mapping to props? Perhaps trigger a `getCurrentUserData()` inside `componentDidMount()` and write to state there?

client/dashboard/leaderboards/index.js
joshuatf: You're right, good catch!  Updated.

client/dashboard/leaderboards/index.js
joshuatf: While I agree that it seems overly complex, I think this is the [accepted method of handling this](https://reactjs.org/docs/react-component.html#componentdidupdate) and how we do so in other components. Unfortunately, we can’t set this in `componentDidMount()` since user data is fetched asynchronously and would not get set in time. 

If you have other suggestions, I’m open to trying them.

joshuatf:
Thanks for testing, Jeff!

>In testing the functionality, I noticed that there seems to be a flash of default values for the Top Selling Products leaderboard:

This is my first question from above under the “questions” section and also an issue we have in other components where we receive user preferences. If we want to change how this works, I see 2 viable options (let me know if you have other ideas):
* We don’t render the component (or set a loader) until user data is fetched.
* We hydrate the initial state from a server side call to get user preferences. 

I’m partial to the 2nd since this will result in better user experience and less time waiting, especially if we’re using `get_user_meta()` which already uses caching. I think if we go that route, I’d prefer to handle in a separate PR if that’s okay with you to keep the scope down on this PR.

jeffstieler:
> This is my first question from above under the “questions” section and also an issue we have in other components where we receive user preferences.

Whoops, I must have glossed over that. Sorry.

> I’m partial to the 2nd since this will result in better user experience and less time waiting, especially if we’re using `get_user_meta()` which already uses caching.

I also prefer that approach over the other, but both seem fine.

> I’d prefer to handle in a separate PR if that’s okay with you to keep the scope down on this PR.

Sure. I'll give this another review and we can work on changing the user preferences in another PR.

client/dashboard/leaderboards/index.js
jeffstieler: Seems like this will be a non-issue when the initial state fetch is refactored.

client/dashboard/leaderboards/index.js
```diff
@@ -0,0 +1,184 @@
...
+					);
+
+					if ( ! hidden ) {
+						hiddenLeaderboardKeys.push( key );
+					}
+					return { hiddenLeaderboardKeys };
+				},
+				() => {
+					const userDataFields = {
+						[ 'dashboard_leaderboards' ]: this.state.hiddenLeaderboardKeys,
```
jeffstieler: Optional refactor: What do you think about using an object instead of an array for the leaderboard preferences? I think that would clean up a lot of the toggling logic.

jeffstieler (approved):
Code looks good and tests well - optional refactor of the leaderboard preferences in comments.

timmyc:
> Should we wait until user preferences are retrieved or is this okay?

Maybe we can pre-load those prefs and stash them in `wcSettings` to help resolve the flash - or we just show a few placeholders in these sections. I'm fine either way seems like the placeholder would be an easier implementation.

client/dashboard/leaderboards/index.js
joshuatf: I think that's a great idea.  We can use the lodash `xor` function that @Aljullu recommended [here](https://github.com/woocommerce/wc-admin/pull/1225#discussion_r245731281) to simplify the array logic down to one line.

Added in https://github.com/woocommerce/wc-admin/pull/1234/commits/917f75226c6630967cb9ddd404a5063fd78afd6a.

joshuatf:
Thanks for reviewing @jeffstieler and @timmyc!

User preferences initial state issue outlined here - https://github.com/woocommerce/wc-admin/issues/1250

PR #1235, opened 2019-01-04
Remove unused show_product_categories_in_rest function

### LINKED ISSUE #1232: Unused `show_product_categories_in_rest()` should be removed
**Describe the bug**
As @jeffstieler noted (thanks!) this function is no longer being used and should be removed.

**Additional context**
Was a part of the category API in https://github.com/woocommerce/wc-admin/pull/1131 until this was later removed in favor of using the WC API.

joshuatf:
Fixes #1232

Removes a function that was previously used when querying categories from the WP api which are now queried instead from the WC api.

1.  Run phpunit tests and make sure all are passing.
2. Make sure categories still show correctly on category reports.


Aljullu (approved):
Everything seems to work fine without this function.

:shipit:

PR #1236, opened 2019-01-04
Add dashboard chart links to reports

### LINKED ISSUE #1191: Dashboard: Add Click Action for Charts
The charts displayed on the dashboard are meant to provide a quick view into key stats for a store, but also a launching point to dive deeper into a set of data.

image

The goal with this enhancement is for the entire chart block to be a click target to take the user to the associated report/summary chart. So in the above screenshot clicking anywhere within the Items Sold block would take the user to the Products Report ( `/wp-admin/admin.php?page=wc-admin#/analytics/products` with default summary number / chart shown ) for the current dashboard date range - while clicking on Average Order Value would take the user to `/wp-admin/admin.php?page=wc-admin#/analytics/orders?chart=avg_order_value` with the date range added too.

__Questions__
- I'm not sure how to best handled this from an a11y standpoint ( @ryelle any ideas or input here? )?
- @LevinMedia I think we also discussed the need for a little bit of design input here. Perhaps a hover state on the chart blocks? Also does this really work well with the existing hover states for the tooltips when hovering on a data point in the chart? One alternate idea is to just make the chart block titles linked to the corresponding report which is the pattern used on the similar linking of dashboard tiles to deeper reports in WPCOM site stats:

stats trout bummin wordpress com 2018-12-26 12-24-24

joshuatf:
Fixes #1191 

Makes the charts link directly to the reports on click and adds screen reader links for accessibility.

### Accessibility

### Screenshots
screen shot 2019-01-04 at 6 16 23 pm

1.  Open the dashboards page.
2.  Enable each chart under the "Charts" section and click on each checking to make sure it goes to the correct report when clicked.
3.  Using your keyboard, tab through the page to make sure the screen reader link for each report is accessible.
4.  Using a screen reader, verify that you are still able to read through the charts in addition to being able to access the link to the report.

### Questions
* @ryelle I was able to tab through and access the links via screen reader, but this is somewhat unfamiliar territory for me so I appreciate any guidance you can give.
>Sighted keyboard users will tab through the page, and when they land on the link, it will appear above the graph, then they can move to the next block (because none of the graph will be interactive/tab-able content - unless we want to give focus to the blue boxes too, but if they don't have an action, is that useful?)

Does this mean we want to skip over the chart entirely when tabbing through so these users cannot access the information in (hovered) chart points?
* @LevinMedia I've made the screen reader label visible on tabbing through the page, as seen in the above screenshot "Items Sold Report", on tabbing and styled the same as the WordPress dashboard.  Does this look okay or do we have other styles in place for wc-admin?

### DIFF
--- a/client/dashboard/dashboard-charts/block.js
+++ b/client/dashboard/dashboard-charts/block.js
@@ -42,13 +42,9 @@ class ChartBlock extends Component {
 				className="woocommerce-dashboard__chart-block-wrapper"
 				onClick={ this.handleChartClick }
 			>
-				<Card
-					className="woocommerce-dashboard__chart-block"
-					title={ charts[ 0 ].label }
-					onClick={ this.handleChartClick }
-				>
+				<Card className="woocommerce-dashboard__chart-block" title={ charts[ 0 ].label }>
 					<a
-						className="woocommerce-dashboard__chart-block screen-reader-text"
+						className="screen-reader-text"
 						href={ getAdminLink(
 							'admin.php?page=wc-admin#/analytics/' +
 								charts[ 0 ].endpoint +
--- a/client/dashboard/dashboard-charts/block.js
+++ b/client/dashboard/dashboard-charts/block.js
@@ -16,7 +16,7 @@ import { getAdminLink, history } from '@woocommerce/navigation';
  */
 import ReportChart from 'analytics/components/report-chart';
 import './block.scss';
-import { __ } from '@wordpress/i18n';
+import { __, sprintf } from '@wordpress/i18n';
 
 class ChartBlock extends Component {
 	handleChartClick = () => {
@@ -52,8 +52,8 @@ class ChartBlock extends Component {
 								charts[ 0 ].key
 						) }
 					>
-						{ /* translators: appended to end of chart type for report link */
-						charts[ 0 ].label + __( ' Report', 'wc-admin' ) }
+						{ /* translators: %s is the chart type */
+						sprintf( __( '%s Report', 'wc-admin' ), charts[ 0 ].label ) }
 					</a>
 					<ReportChart
 						charts={ charts }
--- a/client/dashboard/dashboard-charts/block.js
+++ b/client/dashboard/dashboard-charts/block.js
@@ -4,6 +4,7 @@
  */
 import { Component } from '@wordpress/element';
 import PropTypes from 'prop-types';
+import { __, sprintf } from '@wordpress/i18n';
 
 /**
  * WooCommerce dependencies
@@ -16,7 +17,6 @@ import { getAdminLink, history } from '@woocommerce/navigation';
  */
 import ReportChart from 'analytics/components/report-chart';
 import './block.scss';
-import { __, sprintf } from '@wordpress/i18n';
 
 class ChartBlock extends Component {
 	handleChartClick = () => {
--- a/client/dashboard/dashboard-charts/block.js
+++ b/client/dashboard/dashboard-charts/block.js
@@ -2,21 +2,33 @@
 /**
  * External dependencies
  */
-import { Component, Fragment } from '@wordpress/element';
+import { Component } from '@wordpress/element';
 import PropTypes from 'prop-types';
 
 /**
  * WooCommerce dependencies
  */
 import { Card } from '@woocommerce/components';
+import { getAdminLink, history } from '@woocommerce/navigation';
 
 /**
  * Internal dependencies
  */
 import ReportChart from 'analytics/components/report-chart';
 import './block.scss';
+import { __ } from '@wordpress/i18n';
 
 class ChartBlock extends Component {
+	handleChartClick = () => {
+		const { charts } = this.props;
+
+		if ( ! charts || ! charts.length ) {
+			return null;
+		}
+
+		history.push( 'analytics/' + charts[ 0 ].endpoint + '?chart=' + charts[ 0 ].key );
+	};
+
 	render() {
 		const { charts, endpoint, path, query } = this.props;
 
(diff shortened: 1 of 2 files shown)


LevinMedia:
Hey @joshuatf looking fantastic!

After seeing this in action, and discussing with @jameskoster I'd like to remove the hover interaction from the legend portion of the chart that highlights a single series. With that removed, the only hover effect would be the tool tips  in the chart. 

Two bugs: 

First when you click the chart, it takes you to the desired report just fine, however if you click the back button to return to the dashboard, the tool tips no longer appear in the chart when you hover. 

Second, hovering over the chart header only activates the hover state in the header. However, if you hover anywhere else in in the card, the hover appears over the entire card including the header. 

Hovering over the header should also show the hover state in the rest of the card too.

client/dashboard/dashboard-charts/block.js
```diff
@@ -25,8 +37,28 @@ class ChartBlock extends Component {
...
+				className="woocommerce-dashboard__chart-block-wrapper"
+				onClick={ this.handleChartClick }
+			>
+				<Card
+					className="woocommerce-dashboard__chart-block"
+					title={ charts[ 0 ].label }
+					onClick={ this.handleChartClick }
+				>
+					<a
+						className="woocommerce-dashboard__chart-block screen-reader-text"
```
ryelle: Did you mean to use the class `woocommerce-dashboard__chart-block` again?

client/dashboard/dashboard-charts/block.js
```diff
@@ -25,8 +37,28 @@ class ChartBlock extends Component {
...
+						href={ getAdminLink(
+							'admin.php?page=wc-admin#/analytics/' +
+								charts[ 0 ].endpoint +
+								'?chart=' +
+								charts[ 0 ].key
+						) }
+					>
+						{ /* translators: appended to end of chart type for report link */
+						charts[ 0 ].label + __( ' Report', 'wc-admin' ) }
```
ryelle: You can use sprintf & placeholders to translate this correctly, or just use the chart label directly.

```js
import { __, sprintf } from '@wordpress/i18n';
…
/* translators: %s is the chart type */
sprintf( __( '%s Report', 'wc-admin' ), charts[ 0 ].label );
```

client/dashboard/dashboard-charts/block.js
```diff
@@ -25,8 +37,28 @@ class ChartBlock extends Component {
...
-				<Card className="woocommerce-dashboard__chart-block" title={ charts[ 0 ].label }>
+			<div
+				role="presentation"
+				className="woocommerce-dashboard__chart-block-wrapper"
+				onClick={ this.handleChartClick }
+			>
+				<Card
+					className="woocommerce-dashboard__chart-block"
+					title={ charts[ 0 ].label }
+					onClick={ this.handleChartClick }
```
ryelle: Do you need the second `onClick` here?

ryelle (changes requested):
I was seeing some bugs with the legend hover "sticking" when i'm not hovered on the legend item anymore, but it looks like that'll be dropped so I think it's fine.

I did see some intermittent issues where tabbing through the chart data points didn't show the tooltips, but if I changed browser tabs away & back again it seems fine? Not sure what's going on there.

> Does this mean we want to skip over the chart entirely when tabbing through so these users cannot access the information in (hovered) chart points?

Nope– that was my assumption that we would need to get rid of the tooltips because I assumed they worked like `<Popover />`, and were outside the HTML structure of the chart (and would therefore not trigger the click event, since it wouldn't bubble to the container). Since we do have the tooltips, what you have here works well.

client/dashboard/dashboard-charts/block.js
joshuatf: Much better way of handling this- thanks!

joshuatf:
@ryelle Thanks for the review!  Made the changes you requested; let me know if things look okay now.

@LevinMedia I did intermittently see the tooltip issue.  I'm also seeing this behavior on all charts from other reports as well when navigating to another page and back, so if you think it's okay, we can address this in another PR (issue noted here - https://github.com/woocommerce/wc-admin/issues/1243).

For the legend hover, I've added a pseudo element that should prevent the hover handling there and also allows us to set the cursor to a pointer in this area.

client/dashboard/dashboard-charts/block.js
```diff
@@ -2,21 +2,33 @@
...
  */
 import { Card } from '@woocommerce/components';
+import { getAdminLink, history } from '@woocommerce/navigation';
 
 /**
  * Internal dependencies
  */
 import ReportChart from 'analytics/components/report-chart';
 import './block.scss';
+import { __, sprintf } from '@wordpress/i18n';
```
ryelle: This should be under external dependencies

ryelle (approved):
accessibility-wise, this is working as expected 👍 One note about code organization, but once that's moved this is good to merge.

LevinMedia:
Heya @joshuatf just tested again one small bug remains. When you hover in the chart header, the color doesn't update to the hover state directly behind each series in the legend, so you get a bit of a stripe effect.

screen shot 2019-01-07 at 8 54 08 pm

As we noted in our slack convo, there seems to be an issue with the responsive behavior of the charts on the dashboard - but we can tackle that in another issue. This screen capture shows a comparison between the behavior of the revenue report and the dashboard charts

https://cl.ly/446168d1e8d4

joshuatf:
Thank you both for reviewing!

Legend color issue resolved in https://github.com/woocommerce/wc-admin/pull/1236/commits/d7e214bf4b8593e22a56cbc5a651693ff457e958 and i18n moved to external dependencies.

I've created an issue to track the grid responsiveness problem that @LevinMedia noted [here](https://github.com/woocommerce/wc-admin/issues/1249) so we can resolve in a separate PR.

PR #1244, opened 2019-01-07
Defer chart drawing until chart component has rendered

### LINKED ISSUE #1243: Chart tooltips disappear on history back
**Describe the bug**
When navigating back in the browser after visiting another page, the chart tooltips are *sometimes* no longer visible.

**To Reproduce**

1. Go to any page with a chart.
2. Click on a link to go to another page within wc-admin.
3. Navigate back.
4. Note that the tooltips don't appear on chart hover or repeat a couple times if the first time doesn't produce the issue.

**Expected behavior**
Expected tooltips to work.

**Screenshots**
screen shot 2019-01-07 at 12 33 20 pm

**Additional context**
Seems to happen more on the dashboard charts page as @LevinMedia noted in https://github.com/woocommerce/wc-admin/pull/1236.  It may have something to do with API requests returning data in time since that page has more charts.

joshuatf:
Fixes #1243 

The chart render includes a ref to the tooltips.  This is passed to the child `D3Base` component that is not available on initial render.

This PR defers the drawing until the `D3Chart` component has rendered so the ref is available.

This bug is not obvious since the chart will render multiple times on intial load, but when navigating using the router, we'll see the tooltips not work when the render function is only hit once.

### Screenshots
jan-07-2019 13-43-56

1.  Open any page with a chart.
2.  Navigate to another page within wc-admin so that the router is used.
3.  Navigate back using your browser's back button.
4.  Check that the chart is loaded correctly each time and that tooltips work.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1266,7 +1266,7 @@
         },
         "globby": {
           "version": "8.0.1",
-          "resolved": "http://registry.npmjs.org/globby/-/globby-8.0.1.tgz",
+          "resolved": "https://registry.npmjs.org/globby/-/globby-8.0.1.tgz",
           "integrity": "sha512-oMrYrJERnKBLXNLVTqhm3vPEdJ/b2ZE28xN4YARiix1NOIOBPEpOUnm844K1iu/BkphCaf2WNFwMszv8Soi1pw==",
           "dev": true,
           "requires": {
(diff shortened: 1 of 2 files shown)


Aljullu (approved):
Thanks for working on this. It's testing well. :rocket:

PR #1246, opened 2019-01-07
Add v4 orders endpoint

### LINKED ISSUE #950: REST API: Create new /v4 of Orders Endpoint

joshuatf:
Fixes #950 

Adds the new endpoint so we can use new filters and return data from lookup tables.

### Screenshots
screen shot 2019-01-09 at 2 32 29 pm

1.  Make sure you have coupons and products associated with orders in your lookup tables.
2.  Rebuild your reports to ensure statuses are in the lookup tables.
3.  Send a request to `wp-json/wc/v4/reports/orders` and test the following params **(note the v4 namespace)**:
`before (date)`
`after (date)`
`page (int)`
`per_page (int)`
`orderby (date, num_items_sold, or net_total)`
`order (asc or desc)`
`status_is (string - status type without 'wc-' prefix)`
`status_is_not (string - status type without 'wc-' prefix)`
`product_includes (int or array)`
`product_excludes (int or array)`
`coupon_includes (int or array)`
`coupon_excludes (int or array)`
`customer_type (string 'returning' or 'new')`
`extended_info (bool)`
4.  Run phpunit tests to make sure everything is ✅

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-orders-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-orders-data-store.php
@@ -50,18 +50,15 @@ class WC_Admin_Reports_Orders_Data_Store extends WC_Admin_Reports_Data_Store imp
 	);
 
 	/**
-	 * Returns comma separated ids of included coupons, based on query arguments from the user.
-	 *
-	 * @param array $query_args Parameters supplied by the user.
-	 * @return string
+	 * Constructor
 	 */
-	protected function get_included_coupons( $query_args ) {
-		$included_coupons_str = '';
-
-		if ( isset( $query_args['coupons'] ) && is_array( $query_args['coupons'] ) && count( $query_args['coupons'] ) > 0 ) {
-			$included_coupons_str = implode( ',', $query_args['coupons'] );
-		}
-		return $included_coupons_str;
+	public function __construct() {
+		global $wpdb;
+		$table_name = $wpdb->prefix . self::TABLE_NAME;
+		// Avoid ambigious columns in SQL query.
+		$this->report_columns['order_id']     = $table_name . '.' . $this->report_columns['order_id'];
+		$this->report_columns['date_created'] = $table_name . '.' . $this->report_columns['date_created'];
+		$this->report_columns['customer_id']  = $table_name . '.' . $this->report_columns['customer_id'];
 	}
 
 	/**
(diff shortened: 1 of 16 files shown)


includes/data-stores/class-wc-admin-reports-orders-data-store.php
```diff
@@ -233,320 +149,136 @@ public function get_data( $query_args ) {
...
+			if ( $allowed_statuses ) {
+				$subqueries[] = "status IN ( '" . implode( "','", $allowed_statuses ) . "' )";
+			}
 		}
 
-		$order = wc_get_order( $post_id );
-		if ( ! $order ) {
-			return;
+		if ( isset( $query_args['status_is_not'] ) && is_array( $query_args['status_is_not'] ) && count( $query_args['status_is_not'] ) > 0 ) {
+			$forbidden_statuses = array_map( array( $this, 'normalize_order_status' ), $query_args['status_is_not'] );
```
timmyc: ☠️ we do not speak of the forbidden_statuses

includes/data-stores/class-wc-admin-reports-orders-data-store.php
```diff
@@ -233,320 +149,136 @@ public function get_data( $query_args ) {
...
-				if ( $customer_id ) {
-					$data['customer_id'] = $customer_id;
-					$format[] = '%d';
+	protected function include_extended_info( &$orders_data, $query_args ) {
+		foreach ( $orders_data as $key => $order_data ) {
+			$order                       = wc_get_order( $order_data['order_id'] );
+			$extended_info               = new ArrayObject();
+			$extended_info['products']   = array();
+			$extended_info['categories'] = array();
+			foreach ( $order->get_items() as $item ) {
```
timmyc: I think we had this discussion before on another PR ( with @psealock me thinks ), and I'm fine with doing this in a follow-up, but couldn't we be potentially duplicating some queries here for getting taxonomies?

timmyc (commented):
Quite the PR here Josh 😮 ! I've left a few comments that I figured I'd get posted while I do some manual testing against the endpoint. Also would love to get some more 👀 on this PR from perhaps @peterfabian and @claudiulodro

includes/data-stores/class-wc-admin-reports-orders-data-store.php
timmyc: The only idea I had for the categories again was to maybe pre-fetch all the categories prior to the foreach loops once, and create a map of them $terms[ id ] => $term->name sort of thing. Or perhaps even build up that mapping and one for the products as we loop through the orders.

Quite possibly I'm being overly paranoid here, and mysql or WP query caching will already help us out here.

includes/data-stores/class-wc-admin-reports-orders-data-store.php
```diff
@@ -26,175 +20,101 @@ class WC_Admin_Reports_Orders_Data_Store extends WC_Admin_Reports_Data_Store imp
...
 		if ( $excluded_coupons ) {
-			$where_filters[] = " {$orders_stats_table}.order_id NOT IN (
-				SELECT
-					DISTINCT {$wpdb->prefix}wc_order_coupon_lookup.order_id
-				FROM
-					{$wpdb->prefix}wc_order_coupon_lookup
-				WHERE
-					{$wpdb->prefix}wc_order_coupon_lookup.coupon_id IN ({$excluded_coupons})
-				)";
+			$sql_query_params['from_clause']  .= " JOIN {$order_coupon_lookup_table} ON {$order_stats_lookup_table}.order_id = {$order_coupon_lookup_table}.order_id";
```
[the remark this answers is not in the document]
joshuatf: Smart thinking!  Updated this for coupons and products.

includes/data-stores/class-wc-admin-reports-orders-data-store.php
joshuatf: Took a stab at optimizing this a bit in https://github.com/woocommerce/wc-admin/pull/1246/commits/4c6119fd4f2d204acfb4fc6b3cc1b1223c5dd640 but it may need a bit more work.

Feedback is welcome 😄

claudiulodro (approved):
Works well in my testing and the data stores look good to me. 👍

joshuatf:
Rebased and added https://github.com/woocommerce/wc-admin/pull/1246/commits/f7e08ba4f429072b050a963b6c23cfa0acf5ff0d which cleans up product/category mapping and fixes some mapping issues.

Could someone give this another look to make sure `extended_info` looks okay before we merge?

psealock:
@joshuatf I'm about halfway through review this, but need to run out to car mechanic. I'll pick this up again in a few hours

psealock (approved):
>Could someone give this another look to make sure extended_info is returning the right products and categories before we merge?

By looking at the code, the endpoint is returning whats I'm expecting. Nice work.

PR #1247, opened 2019-01-07
Add new props to chart component

Aljullu:
Coming from the discussion here: https://github.com/woocommerce/wc-admin/pull/1213#issuecomment-450838602

This PR removes the option `block` from the `mode` prop in charts. `mode` should be used to specify whether the chart is comparing time periods or different items, so screen readers can read the values properly. Using it for other purposes might be problematic.

Instead, this PR adds three new props to the chart component:
- `interactiveLegend`
- `legendPosition`
- `showHeaderControls`

### Accessibility

### Screenshots
_(this PR shouldn't produce any visual change)_

- Go to the Dashboard.
- With the devtools, search for `.bargroup` and verify it has an `aria-label` value (that means it's a `mode=time-comparison` chart).
_Before:_
image
_After:_
image
- Verify there are no regressions in charts.

### DIFF
--- a/client/analytics/components/report-chart/index.js
+++ b/client/analytics/components/report-chart/index.js
@@ -68,7 +68,18 @@ export class ReportChart extends Component {
 	}
 
 	render() {
-		const { query, itemsLabel, mode, path, primaryData, secondaryData, selectedChart } = this.props;
+		const {
+			interactiveLegend,
+			itemsLabel,
+			legendPosition,
+			mode,
+			path,
+			primaryData,
+			query,
+			secondaryData,
+			selectedChart,
+			showHeaderControls,
+		} = this.props;
 
 		if ( primaryData.isError || secondaryData.isError ) {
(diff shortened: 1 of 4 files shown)


timmyc (approved):
Looking good and testing well for me. Thanks for cleaning up the props documentation too. If you could add a new item to the 1.4.0 section [here](https://github.com/woocommerce/wc-admin/blob/8873164747ad07c21268aca79f0c596dcc2e243d/packages/components/CHANGELOG.md) detailing the new props that would be awesome.

All good to merge after that.

Aljullu:
> If you could add a new item to the 1.4.0 section [here](https://github.com/woocommerce/wc-admin/blob/8873164747ad07c21268aca79f0c596dcc2e243d/packages/components/CHANGELOG.md) detailing the new props that would be awesome.

Done: c11e2a716f8febe0a40dc73d757b57d9d517d34c. :+1:

PR #1248, opened 2019-01-07
Download Report API: Add path to response, fix ordering, and fix schema typo

### LINKED ISSUE #1245: Downloads report: file name should link to the file
Follow-up from: https://github.com/woocommerce/wc-admin/pull/1238#discussion_r245564538.

According to the designs (p6riRB-3dU-p2), the _Downloads_ table should link to the file via the filename cell.
image

That will require modifying the API endpoint or making another request to get the download links.

emdashcodes:
Fixes #1245.

This PR fixes a few issues with the download report endpoints.

* It adds a `download_path` to the response which contains a URL to the download of the file. 
* It fixes a typo between `download_count` and `downloads_count` in the schema.
* It fixes the ?orderby option so it actually sorts. Tests included.

Note on `download_path`:
This links to the file path that is stored in the product data. This can link externally or to the media library. It is different from a WC download link. Download links are tied to orders/users, so to use that we would need to generate permissions for each admin and link here otherwise.

- Running `phpunit` should be enough to test the changes here, though you can make manual requests or test the report page as well.

### DIFF
--- a/includes/api/class-wc-admin-rest-reports-downloads-controller.php
+++ b/includes/api/class-wc-admin-rest-reports-downloads-controller.php
@@ -107,8 +107,10 @@ class WC_Admin_REST_Reports_Downloads_Controller extends WC_REST_Reports_Control
 		$product_id                  = intval( $data['product_id'] );
 		$_product                    = wc_get_product( $product_id );
 		$file_path                   = $_product->get_file_download_path( $data['download_id'] );
+
 		$filename                    = basename( $file_path );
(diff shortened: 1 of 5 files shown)


Aljullu (approved):
This is looking good to me. :shipit: 

Probably that's outside of the scope of this PR, but I couldn't see `match: any` behaving differently than `match: all`.

Using `match: any` with a `downloadIp_includes` that matches one download but a `user_includes` that doesn't match any, returns 0 results while I would expect it to return 1. Am I missing something or is there a bug here?

justinshreve:
Sounds like a bug - I'll take a look in a follow-up PR!

PR #1255, opened 2019-01-08
Add store performance indicator endpoint

### LINKED ISSUE #912: REST API: Store Performance Endpoint
Prior Discussion: p90Yrv-Rv-p2

screen shot 2018-10-25 at 7 56 55 pm

The post linked above, and the design snapshot give a good glimpse into the why/need behind this issue. Essentially the end-goal is to have one REST endpoint that we, and other application developers can use to retrieve all related store performance stats for a given time period.

Thinking the following arguments should be supported:

`before`: Retrieve performance stats for the dates before this
`after`: Retrieve stats for dates after this
`stats[]`: comma-seperated/array of stats to retrieve

The final argument I think might need to include a "namespace" of sorts to tie into the related stat endpoint. I.E. if I wanted to retrieve the stat totals for average order value ( order report ), and total shipping ( revenue report ), I would request something like `orders/avg_order_value,revenue/shipping` as the `stats` in the request.

When discussing this at the meetup, we felt that using `rest_do_request()` to get the stat totals requested from the various endpoints seemed like a reasonable, and extendible way to build this out - but I'm absolutely open to other suggestions and approaches.

The other idea was that an `OPTIONS` request to this endpoint would return all the associated "totals" / performance indicators that are supplied in the various `/reports/$object/stats` endpoints.

/cc @claudiosanches @peterfabian

emdashcodes:
Fixes #912.

This PR adds a new `/wc/v3/reports/performance-indicators` endpoint, that allows for receiving multiple totals numbers from stat report endpoints.

You pass a `before` and `after` date, as well as a list of stats from a whitelisted list. You can see the list by making an OPTIONS request to `/wc/v3/reports/performance-indicators`. The list of allowed stats is automatically generated by any endpoint that is named `/reports/$endpoint/stats` and contains `totals` in the response.

* Use Postman or another tool to make requests. Example request: `http://binarydev.org/wp-json/wc/v3/reports/performance-indicators?after=2018-01-01T00%3A00%3A00%2B00%3A00&before=2019-01-01T14%3A51%3A00%2B00%3A00&stats=orders/avg_order_value`. Spot check values.
* Use Postman or another tool to make a request to `OPTIONS /wc/v3/reports/performance-indicators`. Make sure that a list of allowed values is displayed.

### DIFF
--- a/includes/api/class-wc-admin-rest-reports-performance-indicators-controller.php
+++ b/includes/api/class-wc-admin-rest-reports-performance-indicators-controller.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * REST API Store Performance controller
+ * REST API Performance indiciators controller
  *
  * Handles requests to the /reports/store-performance endpoint.
  *
@@ -10,12 +10,12 @@
 defined( 'ABSPATH' ) || exit;
 
 /**
- * REST API Reports Store Performance controller class.
+ * REST API Reports Performance indicators controller class.
  *
  * @package WooCommerce/API
  * @extends WC_REST_Reports_Controller
  */
-class WC_Admin_REST_Reports_Store_Performance_Controller extends WC_REST_Reports_Controller {
+class WC_Admin_REST_Reports_Performance_Indicators_Controller extends WC_REST_Reports_Controller {
 
 	/**
 	 * Endpoint namespace.
@@ -29,7 +29,7 @@ class WC_Admin_REST_Reports_Store_Performance_Controller extends WC_REST_Reports
 	 *
 	 * @var string
 	 */
-	protected $rest_base = 'reports/store-performance';
+	protected $rest_base = 'reports/performance-indicators';
 
 	/**
 	 * Maps query arguments from the REST request.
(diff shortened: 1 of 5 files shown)


includes/api/class-wc-admin-rest-reports-performance-indicators-controller.php
```diff
@@ -0,0 +1,268 @@
...
+				if ( empty( $data['schema']['properties']['totals']['properties'] ) ) {
+					continue;
+				}
+
+				foreach ( $data['schema']['properties']['totals']['properties'] as $property_key => $schema_info ) {
+					$allowed_stats[] = $prefix . '/' . $property_key;
+				}
+			}
+		}
+		return $allowed_stats;
```
timmyc: Should this have a filter on it to allow for extensions to register allowed_stats from other non `/wc/v3/reports` endpoints - or possibly to exclude stats from some users?

includes/api/class-wc-admin-rest-reports-performance-indicators-controller.php
```diff
@@ -0,0 +1,268 @@
...
+		/**
+		 * Filter a report returned from the API.
+		 *
+		 * Allows modification of the report data right before it is returned.
+		 *
+		 * @param WP_REST_Response $response The response object.
+		 * @param object           $report   The original report object.
+		 * @param WP_REST_Request  $request  Request used to generate the response.
+		 */
+		return apply_filters( 'woocommerce_rest_prepare_report_performance_indicators', $response, $stat_data, $request );
```
timmyc: Sometimes I should read all the code prior to commenting. This seems like it will be sufficient to allow for extending this endpoint with stats support from other sources in extensions. 🌟

tests/api/reports-performance-indicators.php
```diff
@@ -0,0 +1,155 @@
...
+	public function test_register_routes() {
+		$routes = $this->server->get_routes();
+
+		$this->assertArrayHasKey( $this->endpoint, $routes );
+	}
+
+	/**
+	 * Test getting indicators.
+	 */
+	public function test_get_indicators() {
```
timmyc: Great tests, thank you.

includes/api/class-wc-admin-rest-reports-performance-indicators-controller.php
```diff
@@ -0,0 +1,268 @@
...
+			$allowed_stats = implode( ', ', $allowed_stats );
+		}
+
+		$params            = array();
+		$params['context'] = $this->get_context_param( array( 'default' => 'view' ) );
+		$params['stats']   = array(
+			'description'       => sprintf(
+				/* translators: Allowed values is a list of stat endpoints. */
+				__( 'Limit response to specific report stats. Allowed values: %s.', 'wc-admin' ),
+				$allowed_stats
```
timmyc: Ahh this is where it might be nice to make the result of `get_allowed_stats()` filterable - so if an `OPTIONS` request is sent to the endpoint, any extension-supported stat would be listed here.

timmyc (approved):
This is awesomeness. Really love how you implemented this PR Justin - nice work. I had a couple of questions around extensibility - it seems like as-is it _could_ work just fine, but had the one question about filtering of allowed stat types so extension developers could have their supported stats appear in the `OPTIONS` request.

This is testing out quite well for me, so pre-approving as-is.

timmyc:
❤️ the filter additions. This is going to make for some great extensibility options!

PR #1260, opened 2019-01-09
Handle *_between params in Reports controllers

### LINKED ISSUE #1239: Create common handler for *_between parameters from advanced number filter usage

Spurred from this discussion: https://github.com/woocommerce/wc-admin/pull/1144#discussion_r243793408

All reports that use the advanced number filter have the potential to have a `between` rule value, which effectively gets normalized to both `min` and `max` values.

The normalization of the `between` value should occur in a base class or other reusable location so that report data stores only need to handle the `min` and `max` parameters. (Just keeping it DRY)

* Performing the normalization in JS at or near the [`getQueryFromActiveFilters()`](https://github.com/woocommerce/wc-admin/blob/7ee1a36a0cf8adfc775309ff5ed3b0c7eb6ada98/packages/navigation/src/filters.js#L105) level
* Copy-pasta the same normalization code across Reports REST controllers that need it

Yes. The `between` rule of the advanced number filter won't work without this.

jeffstieler:
Fixes #1239

Adds a utility method to `WC_Admin_Reports_Interval` to normalize numeric range`_between` parameters into `_min` and `_max`. It's only used by the Customers Report endpoint now, but can be used for all reports that make use of the advanced numeric filter.

### Accessibility

N/A - this is purely in the REST API.

- Verify that the `phpunit` tests pass
- Use Postman or another tool to make some test requests to `/wc/v3/reports/customers`
- Test these params: `orders_count_between`, `total_spend_between`, `avg_order_value_between`

### DIFF
--- a/includes/api/class-wc-admin-rest-reports-customers-controller.php
+++ b/includes/api/class-wc-admin-rest-reports-customers-controller.php
@@ -389,8 +389,8 @@ class WC_Admin_REST_Reports_Customers_Controller extends WC_REST_Reports_Control
 		);
 		$params['orders_count_between']     = array(
 			'description'       => __( 'Limit response to objects with an order count between two given integers.', 'wc-admin' ),
-			'type'              => 'string',
-			'validate_callback' => 'rest_validate_request_arg',
+			'type'              => 'array',
+			'validate_callback' => array( 'WC_Admin_Reports_Interval', 'rest_validate_between_arg' ),
 		);
 		$params['total_spend_min']         = array(
 			'description'       => __( 'Limit response to objects with a total order spend greater than or equal to given number.', 'wc-admin' ),
@@ -404,8 +404,8 @@ class WC_Admin_REST_Reports_Customers_Controller extends WC_REST_Reports_Control
 		);
 		$params['total_spend_between']     = array(
 			'description'       => __( 'Limit response to objects with a total order spend between two given numbers.', 'wc-admin' ),
-			'type'              => 'string',
-			'validate_callback' => 'rest_validate_request_arg',
+			'type'              => 'array',
+			'validate_callback' => array( 'WC_Admin_Reports_Interval', 'rest_validate_between_arg' ),
 		);
 		$params['avg_order_value_min']     = array(
 			'description'       => __( 'Limit response to objects with an average order spend greater than or equal to given number.', 'wc-admin' ),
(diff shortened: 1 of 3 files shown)


includes/api/class-wc-admin-rest-reports-customers-controller.php
```diff
@@ -402,7 +404,7 @@ public function get_collection_params() {
 		);
 		$params['total_spend_between']     = array(
 			'description'       => __( 'Limit response to objects with a total order spend between two given numbers.', 'wc-admin' ),
-			'type'              => 'array',
+			'type'              => 'string',
```
joshuatf: Curious if there's a reason we're opting to use strings here and parse inside `normalize_between_params` instead of typing it as an array.  Is this because we don't know if they've filled in both between fields?

```php
		$params['total_spend_between']     = array(
			'description'       => __( 'Limit response to objects with a total order spend between two given numbers.', 'wc-admin' ),
			'type'              => 'array',
			'items'             => array(
				'type' => 'float',
			),
		);
```

includes/class-wc-admin-reports-interval.php
```diff
@@ -474,4 +474,38 @@ public static function intervals_missing( $expected_interval_count, $db_records,
...
+		foreach ( $param_names as $param_name ) {
+			if ( empty( $request[ $param_name . '_between' ] ) ) {
+				continue;
+			}
+
+			$min_param = $param_name . '_min';
+			$max_param = $param_name . '_max';
+			$range     = explode( ',', $request[ $param_name . '_between' ] );
+
+			if ( isset( $range[0] ) && is_numeric( $range[0] ) ) {
```
joshuatf: Same question here.  I would expect a count of 2 to be given or assume this would be malformed and expected the client side to validate having both fields.  But maybe leaving off the front-end validation (as it is now) is easier for users.

joshuatf (commented):
Looking really good and testing well!  This is a really great approach to normalizing between params 🎉 

One question about whether or not we should enforce the lower and upper values for this param to be valid, but otherwise this looks ready to merge.

includes/api/class-wc-admin-rest-reports-customers-controller.php
jeffstieler: I went with string parsing here because the in its current form, the advanced number filter produces a comma delimited value, not an array. I'll take a look and see how difficult it might be to change that (or parse on the JS side) since I agree that its cleaner to accept an array at the endpoint.

jeffstieler: Switched back to an array in https://github.com/woocommerce/wc-admin/pull/1260/commits/ad9cbf4f6e3264e2d06fbea2f80778fc35eb4e35.

includes/class-wc-admin-reports-interval.php
jeffstieler: In https://github.com/woocommerce/wc-admin/pull/1260/commits/ad9cbf4f6e3264e2d06fbea2f80778fc35eb4e35, I added validation and more normalization logic - the lower numeric value is always set to `_min`.

includes/api/class-wc-admin-rest-reports-customers-controller.php
jeffstieler: I'll open a separate PR to switch the numeric filter to generate an array.

tests/api/reports-interval.php
```diff
@@ -796,4 +796,45 @@ public function test_next_year_start() {
 		}
 	}
 
+	/**
```
timmyc: Thanks for the tests Jeff 🌟

timmyc (approved):
Once Jeff informed me I was testing the wrong endpoint, all is working well for me with this PR. Thanks for the tests too.

PR #1267, opened 2019-01-09
Numeric filter: generate a true array value for “between” rule

jeffstieler:
Numeric filter: generate a true array value for “between” rule instead of a comma-delimited string.

Addresses a point made by @joshuatf here: https://github.com/woocommerce/wc-admin/pull/1260#discussion_r246256654

Remove arbitrary/unnecessary string parsing from the advanced number filter value.

### Accessibility

- Navigate to a Customers Report
- Add a numeric filter, set to "between", enter two values
- Click "filter"
- Verify that the values are in a URL encoded array (`_between%5B0%5D=`)

### DIFF
--- a/packages/components/src/filters/advanced/number-filter.js
+++ b/packages/components/src/filters/advanced/number-filter.js
@@ -4,7 +4,7 @@
  */
 import { Component, Fragment } from '@wordpress/element';
 import { SelectControl, TextControl } from '@wordpress/components';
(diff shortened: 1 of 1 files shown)


timmyc (approved):
Code looks good and is testing well for me.

joshuatf (approved):
Nice!  LGTM :shipit:

PR #1271, opened 2019-01-10
Hook up customers report table to real data.

### LINKED ISSUE #918: Analytics: Customers Report Table

jeffstieler:
Fixes #918 _I think?_

Hook up the Customers report table to the real endpoint.

### Screenshots

screen shot 2019-01-09 at 5 46 51 pm

- Navigate to Customers report
- Verify that data loads from your DB
- Verify that only registered customer names link to user profiles
- Verify that a dash (-) shows when sign up date is null

### DIFF
--- a/client/analytics/report/customers/config.js
+++ b/client/analytics/report/customers/config.js
@@ -18,7 +18,7 @@ export const filters = [
 		param: 'filter',
 		showFilters: () => true,
 		filters: [
-			{ label: __( 'All Registered Customers', 'wc-admin' ), value: 'all' },
+			{ label: __( 'All Customers', 'wc-admin' ), value: 'all' },
 			{ label: __( 'Advanced Filters', 'wc-admin' ), value: 'advanced' },
 		],
 	},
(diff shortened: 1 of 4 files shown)


Aljullu (approved):
Code looks good and works as expected. :rocket: 

> Fixes #918 I think?

Not yet, we are still missing the table summary.

image

I also noted I get some unexpected results with filters. Is that still a WIP or should I fill bugs?

jeffstieler:
> I also noted I get some unexpected results with filters. Is that still a WIP or should I fill bugs?

Please file bugs for the filters that aren't working as expected. Thanks!

Issue #1272, opened 2019-01-10
Customers report: table displays country code but filter uses country names


Aljullu:
Not a big deal, but a usability improvement we could consider.

Imagine I have customers from Spain (_ES_) and Andorra (_AD_):
![image](https://user-images.githubusercontent.com/3616980/50962462-9824ab00-14ca-11e9-963b-e5d3bb8011ca.png)

I want to filter Spanish customers. After seeing _Spain_ is represented as _ES_ in the table, I expect the country filter will work the same, so I write _ES_...

![image](https://user-images.githubusercontent.com/3616980/50962625-f18cda00-14ca-11e9-9713-8854f491cc1e.png)

... and none of the suggestions are Spain.

Same happens for many other countries. If I didn't know _ES_ represents _Spain_, I would have no way to filter those users without searching for a list of country codes in Google.

***Suggested solutions***
- Customers country filter should suggest options not only based on the name but also the country code.
- A nice addition would be adding a `title` to the country code with the country name, so hovering with the mouse would reveal which country belongs to that code:
![image](https://user-images.githubusercontent.com/3616980/50962987-c6ef5100-14cb-11e9-9a01-cd8af75da96c.png)
That country name should be used also as the screen reader text.

PR #1274, opened 2019-01-10
Makes table dates accessible

Aljullu:
While filling #1272 I noticed dates in tables are not accessible and were read like numbers by screen readers.

This PR creates a `<Date>` component that makes it easy to add accessible dates in tables.

### Accessibility

- Go to any report and navigate the table with a screen reader.
- Verify dates are read in an understandable way.

### DIFF
--- a/client/analytics/report/coupons/table.js
+++ b/client/analytics/report/coupons/table.js
@@ -4,13 +4,12 @@
  */
 import { __, _n } from '@wordpress/i18n';
 import { Component } from '@wordpress/element';
-import { format as formatDate } from '@wordpress/date';
 import { map } from 'lodash';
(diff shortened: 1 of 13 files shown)


joshuatf (approved):
Nice work, @Aljullu!  This is testing well for me.

Do the d3chart tooltips also get read by screen readers? I wasn't able to access them via my screen reader, but if they do we might want to also add screen reader text for those.

Aljullu:
Thanks for the review @joshuatf!

> Do the d3chart tooltips also get read by screen readers?

Not really, but you should be able to focus on line points/individual bars and the screen reader should read the title and value of that point/bar.

PR #1279, opened 2019-01-11
Date: Remove hour from default interval in getAllowedIntervalsForQuery

### LINKED ISSUE #1278: Incorrect interval displayed in default report view (hours, should be days)
**Describe the bug**
When viewing a report, the default interval is set to hour. 

**To Reproduce**
1. On master, navigate to a report (ie: revenue)
2. note the interval is set to "hours"

**Expected behavior**
The interval should be set to day, and hours should only be available as an interval when viewing a single day. 

**Screenshots**
screen shot 2019-01-10 at 3 42 29 pm

timmyc:
Fixes #1278

Currently when loading the dashboard, or a chart without any date picker parameters set `getAllowedIntervalsForQuery()` is returning `[ 'hour', 'day' ]` as valid intervals for the time selection, resulting in `hour` being selected by default. The desired behavior is to default to `day`, and this also saves a ton of extra queries to render the charts in the dashbaord.

### Accessibility
No interaction changes

- Open the dashboard, verify the interval selection in the charts area only shows 'days' as an option
- Open any report and verify the same
- Change the date range in the date picker to verify other date range options present the expected interval options.

### DIFF
--- a/packages/date/src/index.js
+++ b/packages/date/src/index.js
@@ -395,7 +395,7 @@ export function getAllowedIntervalsForQuery( query ) {
 				allowed = [ 'day', 'week', 'month', 'quarter' ];
 				break;
 			default:
-				allowed = [ 'hour', 'day' ];
+				allowed = [ 'day' ];
 				break;
 		}
 	}
--- a/packages/date/test/index.js
+++ b/packages/date/test/index.js
@@ -21,6 +21,7 @@ import {
 	getDateDifferenceInDays,
 	getPreviousDate,
 	getChartTypeForQuery,
+	getAllowedIntervalsForQuery,
 } from '../src';
 
 describe( 'appendTimestamp', () => {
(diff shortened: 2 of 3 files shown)


packages/date/test/index.js
```diff
@@ -82,6 +83,53 @@ describe( 'toMoment', () => {
 	} );
 } );
 
+describe( 'getAllowedIntervalsForQuery', () => {
```
joshuatf: Nice!  Thanks for these tests 😄

packages/date/src/index.js
```diff
@@ -395,7 +395,7 @@ export function getAllowedIntervalsForQuery( query ) {
 				allowed = [ 'day', 'week', 'month', 'quarter' ];
 				break;
 			default:
-				allowed = [ 'hour', 'day' ];
+				allowed = [ 'day' ];
```
joshuatf: This is working well.  Just thinking out loud: I'm wondering what periods hit here (just the "XX to date" presets?) and if we'd ever want to offer the hour as an option for the default case.

`allowed = [ 'day', 'hour' ];`

This would still default to `day` and allow hours.  Although my gut is telling me we might not want hours for those cases anyway.

joshuatf (approved):
Nice work!  One question about the default cases, but it might be a non-issue and otherwise LGTM!

packages/date/src/index.js
timmyc: I think the only time this default clause should be hit is when the `query` object passed in as an argument has a `period` attribute that is either empty ( which is the default behavior when hitting any report or the dashboard it seems), or for some reason a period is passed in that is not in [this list](https://github.com/woocommerce/wc-admin/blob/master/packages/date/src/index.js#L37-L49). So I think the logic is pretty safe, ultimately as @LevinMedia stated in the issue, `hour` should only ever be an option when looking at `today` or `yesterday` in the predefined periods, or if looking at a day or less, which is handled in the `custom` period logic higher up.

I do think it would be worth adding in some more tests around the `custom` period handling above to make sure all is well there too.

joshuatf: Okay that sounds similar to what I was thinking and seems logical.  LGTM! 🚀

PR #1285, opened 2019-01-11
Store all order statuses

### LINKED ISSUE #1251: REST API: Store all order statuses
Relevant discussion - p90Yrv-Vb-p2

Blocked by #1246

Currently we only store 3 order statuses in the `wc_order_stats` table (`processing`, `on-hold`, `completed`).  Since we want to be able to view stats from other order types and also custom order statuses, we want to instead store all order statuses.

This will involve:
* Creating an array of *excluded* order statuses with the default excluded being `refunded`, `pending`, `failed`, and `cancelled`.
* Replacing the `get_report_order_statuses()` that act as a list of included status types with the new array of excluded order statuses.  These will eventually be pulled from the Settings page.
* Making sure that all non-excluded statuses are used across all reports that involve querying or joining on orders.
* Allowing excluded types to still be queried by the advanced filter in the Orders report.

joshuatf:
Fixes #1251 

This PR:
* Stores all order statuses
* Adds an array of excluded statuses
* Filters all products, coupons, categories, taxes, variations, and orders by the default (non-excluded) statuses if no status query is provided

1.  Make sure that you have regenerated lookup tables and that the `status` column is filled in `wc_order_stats`.
2.  Create requests to each of the endpoints.
3. Check that stats returned included stats from all orders excluding the default excluded statuses (`refunded`, `pending`, `failed`, `cancelled`).
4.  Make a request to the `/wp-json/wc/v3/reports/orders/stats` endpoint and check that `status_is` and `status_is_not` query params override default excluded statuses (e.g., you can still query by `failed` order statuses).
5.  Check that all phpunit tests are passing.

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-categories-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-categories-data-store.php
@@ -60,6 +60,16 @@ class WC_Admin_Reports_Categories_Data_Store extends WC_Admin_Reports_Data_Store
 	);
 
 	/**
+	 * Constructor
+	 */
+	public function __construct() {
+		global $wpdb;
+		$table_name = $wpdb->prefix . self::TABLE_NAME;
+		// Avoid ambigious column order_id in SQL query.
+		$this->report_columns['orders_count'] = str_replace( 'order_id', $table_name . '.order_id', $this->report_columns['orders_count'] );
+	}
+
+	/**
 	 * Return the database query with parameters used for Categories report: time span and order status.
 	 *
 	 * @param array $query_args Query arguments supplied by the user.
@@ -93,7 +103,7 @@ class WC_Admin_Reports_Categories_Data_Store extends WC_Admin_Reports_Data_Store
 
 		$order_status_filter = $this->get_status_subquery( $query_args );
 		if ( $order_status_filter ) {
-			$sql_query_params['from_clause']  .= " JOIN {$wpdb->prefix}posts ON {$order_product_lookup_table}.order_id = {$wpdb->prefix}posts.ID";
+			$sql_query_params['from_clause']  .= " JOIN {$wpdb->prefix}wc_order_stats ON {$order_product_lookup_table}.order_id = {$wpdb->prefix}wc_order_stats.order_id";
 			$sql_query_params['where_clause'] .= " AND ( {$order_status_filter} )";
 		}
 
@@ -216,9 +226,6 @@ class WC_Admin_Reports_Categories_Data_Store extends WC_Admin_Reports_Data_Store
 			'fields'        => '*',
 			'categories'    => array(),
 			'extended_info' => false,
-			// This is not a parameter for products reports per se, but maybe we should restricts order statuses here, too?
-			'order_status'  => parent::get_report_order_statuses(),
-
 		);
 		$query_args = wp_parse_args( $query_args, $defaults );
 
--- a/includes/data-stores/class-wc-admin-reports-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-data-store.php
@@ -784,6 +784,13 @@ class WC_Admin_Reports_Data_Store {
 			}
 		}
 
+		if ( ( ! isset( $query_args['status_is'] ) || empty( $query_args['status_is'] ) )
+			&& ( ! isset( $query_args['status_is_not'] ) || empty( $query_args['status_is_not'] ) )
+		) {
+			$allowed_statuses = array_map( array( $this, 'normalize_order_status' ), $this->get_report_order_statuses() );
+			$subqueries[]     = "{$wpdb->prefix}wc_order_stats.status IN ( '" . implode( "','", $allowed_statuses ) . "' )";
+		}
+
 		return implode( " $operator ", $subqueries );
 	}
 
--- a/includes/data-stores/class-wc-admin-reports-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-data-store.php
@@ -251,12 +251,13 @@ class WC_Admin_Reports_Data_Store {
 	 * If there are less records in the database than time intervals, then we need to remap offset in SQL query
 	 * to fetch correct records.
 	 *
-	 * @param array $intervals_query Array with clauses for the Intervals SQL query.
-	 * @param array $query_args Query arguements.
-	 * @param int   $db_interval_count Database interval count.
-	 * @param int   $expected_interval_count Expected interval count on the output.
+	 * @param array  $intervals_query Array with clauses for the Intervals SQL query.
+	 * @param array  $query_args Query arguements.
+	 * @param int    $db_interval_count Database interval count.
+	 * @param int    $expected_interval_count Expected interval count on the output.
+	 * @param string $table_name Name of the db table relevant for the date constraint.
 	 */
-	protected function update_intervals_sql_params( &$intervals_query, &$query_args, $db_interval_count, $expected_interval_count ) {
+	protected function update_intervals_sql_params( &$intervals_query, &$query_args, $db_interval_count, $expected_interval_count, $table_name ) {
 		if ( $db_interval_count === $expected_interval_count ) {
 			return;
 		}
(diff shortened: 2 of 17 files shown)


includes/data-stores/class-wc-admin-reports-products-data-store.php
```diff
@@ -72,6 +72,16 @@ class WC_Admin_Reports_Products_Data_Store extends WC_Admin_Reports_Data_Store i
...
 	);
 
+	/**
+	 * Constructor
+	 */
+	public function __construct() {
+		global $wpdb;
+		$table_name = $wpdb->prefix . self::TABLE_NAME;
+		// Avoid ambigious column order_id in SQL query.
+		$this->report_columns['orders_count'] = str_replace( 'order_id', $table_name . '.order_id', $this->report_columns['orders_count'] );
```
jeffstieler: Same question about `str_replace()`.

includes/data-stores/class-wc-admin-reports-products-stats-data-store.php
```diff
@@ -40,6 +40,16 @@ class WC_Admin_Reports_Products_Stats_Data_Store extends WC_Admin_Reports_Produc
...
 	);
 
+	/**
+	 * Constructor
+	 */
+	public function __construct() {
+		global $wpdb;
+		$table_name = $wpdb->prefix . self::TABLE_NAME;
+		// Avoid ambigious column order_id in SQL query.
+		$this->report_columns['orders_count'] = str_replace( 'order_id', $table_name . '.order_id', $this->report_columns['orders_count'] );
```
jeffstieler: Same question about `str_replace()`.

includes/data-stores/class-wc-admin-reports-taxes-data-store.php
```diff
@@ -61,8 +61,9 @@ class WC_Admin_Reports_Taxes_Data_Store extends WC_Admin_Reports_Data_Store impl
 	public function __construct() {
 		global $wpdb;
 		$table_name                          = $wpdb->prefix . self::TABLE_NAME;
-		// Avoid ambigious column tax_rate_id in SQL query.
-		$this->report_columns['tax_rate_id'] = $table_name . '.' . $this->report_columns['tax_rate_id'];
+		// Avoid ambigious columns in SQL query.
+		$this->report_columns['tax_rate_id']  = $table_name . '.' . $this->report_columns['tax_rate_id'];
+		$this->report_columns['orders_count'] = str_replace( 'order_id', $table_name . '.order_id', $this->report_columns['orders_count'] );
```
jeffstieler: Same question about `str_replace()`.

includes/data-stores/class-wc-admin-reports-variations-data-store.php
```diff
@@ -66,6 +66,15 @@ class WC_Admin_Reports_Variations_Data_Store extends WC_Admin_Reports_Data_Store
...
 	);
 
+	/**
+	 * Constructor
+	 */
+	public function __construct() {
+		global $wpdb;
+		$table_name = $wpdb->prefix . self::TABLE_NAME;
+		// Avoid ambigious column order_id in SQL query.
+		$this->report_columns['orders_count'] = str_replace( 'order_id', $table_name . '.order_id', $this->report_columns['orders_count'] );
```
jeffstieler: Same question about `str_replace()`.

includes/data-stores/class-wc-admin-reports-variations-data-store.php
```diff
@@ -91,11 +100,10 @@ protected function get_sql_query_params( $query_args ) {
...
 
-		if ( is_array( $query_args['order_status'] ) && count( $query_args['order_status'] ) > 0 ) {
-			$statuses = array_map( array( $this, 'normalize_order_status' ), $query_args['order_status'] );
-
-			$sql_query_params['from_clause']  .= " JOIN {$wpdb->prefix}posts ON {$order_product_lookup_table}.order_id = {$wpdb->prefix}posts.ID";
-			$sql_query_params['where_clause'] .= " AND {$wpdb->prefix}posts.post_status IN ( '" . implode( "','", $statuses ) . "' ) ";
+		$order_status_filter = $this->get_status_subquery( $query_args );
+		if ( $order_status_filter ) {
+			$sql_query_params['from_clause']  .= " JOIN {$wpdb->prefix}wc_order_stats ON {$order_product_lookup_table}.order_id = {$wpdb->prefix}wc_order_stats.order_id";
+			$sql_query_params['where_clause'] .= " AND ( {$order_status_filter} )";
```
jeffstieler: Is there an opportunity to DRY this up? Perhaps a base "joins on wc_order_stats" class that handles the `from_clause`, `where_clause`, `normalize_order_by()`, and the `orders_count` column?

jeffstieler (commented):
In testing the `/reports/categories` endpoint - I'm seeing 165 results with `status_is = 'failed'`. There are only 33 rows in my `wc_order_stats` table with a status of `wc-failed`.. is this correct?

includes/data-stores/class-wc-admin-reports-categories-data-store.php
```diff
@@ -59,6 +59,16 @@ class WC_Admin_Reports_Categories_Data_Store extends WC_Admin_Reports_Data_Store
...
 	);
 
+	/**
+	 * Constructor
+	 */
+	public function __construct() {
+		global $wpdb;
+		$table_name = $wpdb->prefix . self::TABLE_NAME;
+		// Avoid ambigious column order_id in SQL query.
+		$this->report_columns['orders_count'] = str_replace( 'order_id', $table_name . '.order_id', $this->report_columns['orders_count'] );
```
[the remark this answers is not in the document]
joshuatf: It would be simpler, but I think it would cause more confusion when debugging and means we are defining this value twice instead of filtering it.  

For example, if we redefine this variable entirely and someone changes the value in `$report_columns` instead of the construct method, no errors occur, but the code does not work as intended.

On the other hand, if we filter the value and the user changes the `$report_columns` value from:
`'orders_count'   => 'COUNT(DISTINCT order_id) as orders_count'` to
`'orders_count'   => 'COUNT(order_id) as orders_count'`
then the function continues to work as expected.  And if we filter a wrong value like `'orders_count'   => 'COUNT(DISTINCT order_id) as order_id_alias'`, we at least get an error message.

Neither one of these are very ideal solutions though, so I've opened another issue to address how we deal with ambiguous columns [here](https://github.com/woocommerce/wc-admin/issues/1308).  If it's okay with you, we can deal with that in another PR.

includes/data-stores/class-wc-admin-reports-variations-data-store.php
joshuatf: Hmm, I'm not sure another class would be the best solution here.  Report data stores already extend from `WC_Admin_Reports_Data_Store`, but quite a few also override the `normalize_order_by()` and other methods.

The bigger issue here is that there are about 3-4 different patterns being used in these data stores  to append query parameters which makes it difficult to create a universal class that does not result in overriding in most data stores.  For example, here are a few different variations just for the "where" clause:
```php
$sql_query_params['where_clause'] .= " AND ( {$order_status_filter} )";

$where_subclause = implode( " $operator ", array_filter( array( $where_subclause, $order_status_filter ) ) );

$products_where_clause .= " AND ( {$order_status_filter} )";

$where_filters[] = $order_status_filter;

$sql_query_params['where_clause'] .= " AND ( {$order_status_filter} )";
$sql_query_params['where_clause'] .= ' AND variation_id > 0';
```

It seems like we could make a call outside of and adjacent to the `get_sql_query_params()` that might work for most data stores, but that would also be semantically incorrect and we'd be introducing more patterns that may come back to haunt us later.

I do agree that it would be nice if this were dryer, but it seems like a fairly large refactor that might be outside the scope of this PR.  If we decided to do something like that, it might be worth breaking into bite-size chunks (like just refactoring `get_sql_query_params()` to use the same patterns) to create more manageable PRs.

joshuatf:
@jeffstieler Thanks for taking the time to review.

>I'm seeing 165 results with status_is = 'failed'. There are only 33 rows in my wc_order_stats table with a status of wc-failed.. is this correct?

That's very possible if each of those failed 33 orders had a couple categories with a few products, but it could also be a bug.  It's hard to say without seeing the data, but feel free to drop me a DB dump file and I can help investigate if you'd like.

includes/data-stores/class-wc-admin-reports-data-store.php
```diff
@@ -784,6 +785,13 @@ protected function get_status_subquery( $query_args, $operator = 'AND' ) {
 			}
 		}
 
+		if ( ( ! isset( $query_args['status_is'] ) || empty( $query_args['status_is'] ) )
+			&& ( ! isset( $query_args['status_is_not'] ) || empty( $query_args['status_is_not'] ) )
+		) {
+			$excluded_statuses = array_map( array( $this, 'normalize_order_status' ), $this->get_excluded_report_order_statuses() );
+			$subqueries[]      = "{$wpdb->prefix}wc_order_stats.status NOT IN ( '" . implode( "','", $excluded_statuses ) . "' )";
+		}
+
```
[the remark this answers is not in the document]
joshuatf: Good idea.  Updated in https://github.com/woocommerce/wc-admin/pull/1285/commits/930ab80c968ba6c8119c2608fae9c0f47a9d28d5

jeffstieler:
> That's very possible if each of those failed 33 orders had a couple categories with a few products, but it could also be a bug. It's hard to say without seeing the data, but feel free to drop me a DB dump file and I can help investigate if you'd like.

I double checked the report data and it all looks good. 👍

jeffstieler (approved):

PR #1287, opened 2019-01-11
 Update activity button focus styles 

### LINKED ISSUE #1130: Design issue : For header icon menu tray
**Describe the bug**
In the Header right-hand side Menu icon tray, there is a minor design issue which I found while testing it.
Hover and click time bug.

**To Reproduce**

1. Go to 'WooCommerce Dashboard'
2. Click on 'Stock' from right-hand side menu tray
3. Again click on  'stock' the panel will be hidden but the stock background color will not be removed it will stay there.

**Expected behavior**
When we click on stock and the panel will display at that time stock button background color should stay it is correct but when we click again on the stock button and the panel will be hidden at that time stock button background color should be removed.

**Screenshots**
ezgif com-video-to-gif

* OS: Ubuntu 18.04
* Browser Chrome
* Version latest

Aljullu:
Fixes #1130

- Update activity button focus styles.
-  Set overflow to 'auto' instead of 'scroll' in activity panels 

### Accessibility

#### Focus styles
activity-panel

#### Scroll bar
_Before:_
Notice the scroll bar even though there are no contents:
image

_After:_
Now there is no scrollbar:
image

- Navigate through the activity panel with the keyboard and verify focus styles match summary number styles (a border around the button instead of a change in the background color).
- Open activity panels and verify there is no scroll bar when the contents don't overflow.

### DIFF
--- a/client/header/activity-panel/style.scss
+++ b/client/header/activity-panel/style.scss
@@ -77,22 +77,26 @@
 		max-width: min-content;
 		min-width: 80px;
 		width: 100%;
+		font-size: 11px;
 		height: 60px;
-		border-bottom: 3px solid $white;
-
 		@include breakpoint( '>960px' ) {
+			font-size: 13px;
 			height: 80px;
 		}
 
 		&.is-active {
 			color: $core-grey-dark-700;
-			border-bottom: 3px solid $woocommerce;
 			box-shadow: none;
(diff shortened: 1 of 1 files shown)


emdashcodes (approved):
🚢 it. Looks good to me.

PR #1289, opened 2019-01-11
Avoid duplicated Y-axis values

Aljullu:
When the max Y value of a chart was 0, the axis was displaying several 0's one on top of the other, making it look blurry. This PR fixes that.

### Screenshots
_Before:_
image

_After:_
image

- Go to a chart whose max value is 0, for example, the _Orders_ report with the date range set to a day you know there wasn't any order.
- Verify 0 is rendered smoothly.


emdashcodes (approved):
Change looks good to me and is testing well.

The only thing this PR needs is an entry in the CHANGELOG for the components package. Pre-approving pending that.

Aljullu:
@justinshreve 

> The only thing this PR needs is an entry in the CHANGELOG for the components package.

Done!

54f6feba3e99981ff1b0b951b2077d498fe05b3f

PR #1290, opened 2019-01-11
Fix chart date formatting in safari by switching to moment

### LINKED ISSUE #1240: Charts: Tooltips content is NaN in Safari
When testing #1236, I noticed the tooltip labels in Safari are not displaying correctly, I see `NaN, 0NaN   [total]`. This is happening on the dashboard and on the reports pages.

screen shot 2019-01-04 at 4 56 36 pm

screen shot 2019-01-04 at 4 27 04 pm

emdashcodes:
Fixes #1240 (chart tooltip content returning NaN in Safari). It also fixes an issue in Safari where the x-axis would be off by one day, and a few other issues like `aria-label`s also being NaN.

This PR updates date handling in the chart component to use `moment` instead of the date object. In Safari, the format `YYYY-MM-DD HH:mm:ss` can fail/return NaN ([looks like there are numerous people running into a similar issue](https://www.google.com/search?q=Safari+%2B+Date+%2B+invalid+date&oq=Safari+%2B+Date+%2B+invalid+date)). Moment can correctly handle the format returned by the API and return a valid date object.

### Screenshots

**Chrome Before**

chrome_before

**Safari Before**

safari_before

**Chrome After**

chrome_after

**Safari After**

safari_after

* Run `npm test` to run all the tests, including date tests.
* Open any line chart with data.
* Verify the x-axis contains the correct range.
* Hover over the tooltip and verify a correct date is displayed.
* Test bar chart.

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -11,6 +11,7 @@
 - `getColor()` method in chart utils now requires `keys` and `colorScheme` to be passed as separate params.
 - Fix to avoid duplicated Y-axis ticks when the Y max value was 0.
 - Remove decimals from Y-axis when displaying currencies.
+- Fix date formatting on charts in Safari.
 
 # 1.3.0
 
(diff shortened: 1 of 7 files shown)


timmyc (approved):
Spot checked multiple charts in Chrome and Safari, and all is working well for me. Thanks for fixing this up Justin.

PR #1291, opened 2019-01-11
Regenerate report tables using Action Scheduler

### LINKED ISSUE #860: Analytics: Enhancements to Rebuild Reports Tool

jeffstieler:
Fixes #860.

This PR supplants work done in https://github.com/woocommerce/wc-admin/pull/1050.

This PR introduces (re)generating report tables in batches. The batch size can be customized based on the type (customer, order, etc). In order to avoid timeouts and memory issues when creating many batches (for large databases), the batch creation itself is batched (and that size can be customized).

This PR also introduces a mechanism that allows for job to be delayed until another one is finished. This is used to ensure that the orders and products lookup tables are (re)generated after the customers lookup is finished.

_I tested this against a store with ~6500 customers and ~19000 orders._

**To test:**
- Verify that on install, reports tables are generated (can check Tools > Scheduled Actions if it's taking a while)
- Verify that WooCommerce > Status > Tools > Regenerate Reports works as it did before
- Test against a large dataset (lots of customers, products, orders) - verify that actions get scheduled


timmyc (commented):
This is rad. Couple of quick questions to share while i get a wc.com vagrant instance up and running and take it for a spin.

includes/class-wc-admin-api-init.php
```diff
@@ -30,9 +62,38 @@ public function __construct() {
...
 		add_action( 'woocommerce_after_register_post_type', array( 'WC_Admin_Api_Init', 'orders_data_store_init' ), 20 );
 		// Initialize Customers Report data store sync hooks.
-		// Note: we need to hook into 'wp' before `wc_current_user_is_active`.
+		// Note: we need to hook in before `wc_current_user_is_active`.
 		// See: https://github.com/woocommerce/woocommerce/blob/942615101ba00c939c107c3a4820c3d466864872/includes/wc-user-functions.php#L749.
-		add_action( 'wp', array( 'WC_Admin_Api_Init', 'customers_report_data_store_init' ), 9 );
+		add_action( 'wp_loaded', array( 'WC_Admin_Api_Init', 'customers_report_data_store_init' ) );
+
+		// Initialize scheduled action handlers.
+		add_action( self::QUEUE_BATCH_ACTION, array( __CLASS__, 'queue_batches' ), 10, 3 );
```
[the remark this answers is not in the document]
jeffstieler: I'm not sure what you mean - 10 is the default priority - this is really to make sure the 3 args are passed to the callback.

includes/class-wc-admin-api-init.php
```diff
@@ -355,49 +424,141 @@ public static function customers_report_data_store_init() {
...
-	 * @return bool
+	 * @param string $action Single batch action name.
+	 * @return int Batch size.
 	 */
-	public static function customer_lookup_store_init( $updater = null ) {
-		// TODO: this needs to be updated a bit, as it no longer runs as a part of WC_Install, there is no bg updater.
-		global $wpdb;
+	public static function get_batch_size( $action ) {
+		$batch_sizes = array(
+			self::QUEUE_BATCH_ACTION     => 100,
```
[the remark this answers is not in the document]
jeffstieler: These were completely arbitrary choices for batch sizes - I didn't realize this was the same as REST API result sizes 😛. This number relates to the number of Scheduled Actions that will be queued in any given action run. Limiting this should avoid problems for sites that can't handle more than 100 posts being created at a time.

includes/class-wc-admin-api-init.php
```diff
@@ -355,49 +424,141 @@ public static function customers_report_data_store_init() {
...
+			self::CUSTOMERS_BATCH_ACTION => 25,
+			self::ORDERS_BATCH_ACTION    => 10,
+		);
+		$batch_size  = isset( $batch_sizes[ $action ] ) ? $batch_sizes[ $action ] : 25;
```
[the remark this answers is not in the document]
jeffstieler: If we try to `get_batch_size()` for a new action hook that hasn't been added to the above array. Probably a bit defensive, but seems harmless enough. 😄

includes/class-wc-admin-api-init.php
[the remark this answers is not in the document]
timmyc: gah, mondays.

includes/class-wc-admin-api-init.php
```diff
@@ -312,39 +373,47 @@ public static function orders_data_store_init() {
...
+	public static function orders_lookup_batch_init() {
+		$batch_size  = self::get_batch_size( self::ORDERS_BATCH_ACTION );
+		$order_query = new WC_Order_Query(
+			array(
+				'return'   => 'ids',
+				'limit'    => 1,
+				'paginate' => true,
+			)
+		);
+		$result      = $order_query->get_orders();
```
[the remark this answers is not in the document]
jeffstieler: Done in 10a686b.

timmyc (approved):
Code is looking good to me, and this tested out well against my local woocommerce.com snapshot. I believe it took around 40 minutes to run on the default action scheduler settings, and no changes to vvv from a memory standpoint.

includes/class-wc-admin-api-init.php
```diff
@@ -312,39 +373,52 @@ public static function orders_data_store_init() {
...
+		$batch_size  = self::get_batch_size( self::ORDERS_BATCH_ACTION );
+		$order_query = new WC_Order_Query(
+			array(
+				'return'  => 'ids',
+				'limit'   => $batch_size,
+				'page'    => $batch_number,
+				'orderby' => 'ID',
+				'order'   => 'ASC',
+			)
+		);
```
peterfabian: Will this include refunds correctly?

includes/class-wc-admin-api-init.php
```diff
@@ -355,49 +429,146 @@ public static function customers_report_data_store_init() {
...
+		$blocking_jobs = self::queue()->search(
+			array(
+				'status'   => 'pending',
+				'orderby'  => 'date',
+				'order'    => 'DESC',
+				'per_page' => 1,
+				'claimed'  => false,
+				'search'   => $prerequisite_action, // search is used instead of hook to find queued batch creation.
+			)
+		);
```
peterfabian: I'm not sure, but it might be worth checking that action does not stay in `pending` status if one of the prerequisites fail in some unexpected way. Otherwise the dependents would never run.

Unfortunately, I'm not sure if there is a better way... Maybe there could be a cleanup task scheduled, i.e. if the number of pending prerequisite jobs does no decrease in a day, unschedule all of them and reschedule again?

includes/class-wc-admin-api-init.php
```diff
@@ -312,39 +373,52 @@ public static function orders_data_store_init() {
...
+				'limit'   => $batch_size,
+				'page'    => $batch_number,
+				'orderby' => 'ID',
+				'order'   => 'ASC',
+			)
+		);
+		$order_ids = $order_query->get_orders();
+
+		foreach ( $order_ids as $order_id ) {
+			// TODO: schedule single order update if this fails?
```
peterfabian: This  is a valid concern, a bunch of orders might be missing if one fails. Catching the error and scheduling events for individual orders is a clever workaround, if that would be possible.

includes/class-wc-admin-api-init.php
```diff
@@ -312,39 +373,52 @@ public static function orders_data_store_init() {
...
+			WC_Admin_Reports_Orders_Data_Store::sync_order( $order_id );
 			WC_Admin_Reports_Products_Data_Store::sync_order_products( $order_id );
```
peterfabian: Should this maybe include other lookups as well? Coupons, taxes, etc.

peterfabian (commented):
Great work, Jeff. I have a couple of concerns around dependencies and error handling as commented, but otherwise a solid one. 
Really like i) that you added tests, ii) your clear style of coding. 

I'm generating a bunch of orders to test it out now.

includes/class-wc-admin-api-init.php
jeffstieler: Ah, yes - I only handled what was currently being processed in the "renegerate" logic. I'll add the other lookup tables here as well.

includes/class-wc-admin-api-init.php
jeffstieler: Can you elaborate on this or perhaps say it in a different way? I'm not sure I'm following.

If a job throws an uncaught Exception, Action Scheduler will mark it as `failed` ([#](https://github.com/Prospress/action-scheduler/blob/7995d4232b3e4655669d1191c76e47fbce173e05/classes/ActionScheduler_Abstract_QueueRunner.php#L63)) - if the job just doesn't succeed, but doesn't throw any uncaught Exceptions, Action Scheduler will mark it as `complete`.

In both cases, that job would not be in the search results for `pending` actions.

includes/class-wc-admin-api-init.php
jeffstieler: I think you're pointing out the difference in arguments vs. the previous `wc_get_orders()` call ([#](https://github.com/woocommerce/wc-admin/pull/1291/files#diff-86dddbc0241abd7346998f87a6b24228L385))- right?

If so - I think this **does** include refunds, which is a departure from the original logic, and probably a bug.

includes/class-wc-admin-api-init.php
jeffstieler: Added in f788e00.

includes/class-wc-admin-api-init.php
```diff
@@ -355,49 +429,146 @@ public static function customers_report_data_store_init() {
...
+	 *
+	 * @param int    $range_start Starting batch number.
+	 * @param int    $range_end Ending batch number.
+	 * @param string $single_batch_action Action to schedule for a single batch.
+	 * @return void
+	 */
+	public static function queue_batches( $range_start, $range_end, $single_batch_action ) {
+		$batch_size = self::get_batch_size( self::QUEUE_BATCH_ACTION );
+		$range_size = 1 + ( $range_end - $range_start );
+		$schedule   = time() + 5;
```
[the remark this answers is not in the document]
jeffstieler: Renamed to `$action_timestamp` in 19cc37d.

includes/class-wc-admin-api-init.php
peterfabian: Looking at this, and seeing that downloads data store has no update function, it might be worth noting somewhere that donwloads are not rebuilt using this mechanism at the moment.

peterfabian (commented):
I've tested regenerating the data and it seems to work fine with the following exceptions:
- admin with user_id 1 was not present in `wp_wc_customer_lookup` for some reason after regeneration
- `customer_id` in `wp_wc_order_stats` was 0 for all orders made by admin
- `returning_customer` seems to not work correctly

Have you encountered similar problems by any chance?

Please see below a couple of records from the beginning of the `wp_wc_order_stats` after regeneration:
screenshot 2019-01-16 at 11 39 20

jeffstieler:
> I've tested regenerating the data and it seems to work fine with the following exceptions:
> 
> * admin with user_id 1 was not present in wp_wc_customer_lookup for some reason after regeneration
> * customer_id in wp_wc_order_stats was 0 for all orders made by admin

That's due to `'role' => 'customer'` in the `WP_User_Query`, which I thought was reasonable. Do you think we should create "customers" for all users of a store, or just `customer`s?

> * returning_customer seems to not work correctly

The logic that generates this value wasn't changed - are you OK with opening a new issue for this separate from this PR?

peterfabian:
> That's due to 'role' => 'customer' in the WP_User_Query, which I thought was reasonable. Do you think we should create "customers" for all users of a store, or just customers?

Ah, you're right, of course. But then, when I update the order later, it created the entry for admin user in `wp_wc_customer_lookup`, so, it's inconsistent.

> The logic that generates this value wasn't changed - are you OK with opening a new issue for this separate from this PR?

Sure, I can open a new issue. I just thought it seemed to work ok before and the logic seemed sound, so I wasn't sure if it wasn't related to the way how the data is created here or something else.

jeffstieler:
> when I update the order later, it created the entry for admin user in `wp_wc_customer_lookup`, so, it's inconsistent.

Ah.. hmm that's a problem. Was the admin user the actual customer placing that order, or did you just update it? If the admin was the customer - what behavior would you expect? How likely of a scenario do you think it is that non-customer role users will place **real** orders on a store?

peterfabian:
The order was originally created by the admin, then updated by adding a private note after the report tables got regenerated, at which point a new entry in the `wp_wc_customer_lookup` appeared. 
Probably not a big problem here, but I'm not sure, I would assume there won't be that many non-customer users on a typical e-shop, so why not include all users in the job? Then we just avoid these headaches. 

I'm really not sure how often this would happen.

On a related note, we also need to keep in mind that some admins/store managers can create orders on behalf of customers. Not 100% sure if there is a recommended workflow for that scenario and if those store admins would create new users to submit order on their users' behalf? Again, something that should perhaps be solved in a separate issue.

jeffstieler:
> Probably not a big problem here, but I'm not sure, I would assume there won't be that many non-customer users on a typical e-shop, so why not include all users in the job? Then we just avoid these headaches.

Fair point.. I'll remove the `role` restriction from the query. 👍 

> On a related note, we also need to keep in mind that some admins/store managers can create orders on behalf of customers. Not 100% sure if there is a recommended workflow for that scenario and if those store admins would create new users to submit order on their users' behalf? Again, something that should perhaps be solved in a separate issue.

I'll test this flow, but a guest customer should be created so long as the store admin specifies a billing email for the customer when they manually create the order - from the backend of course. If the store admin checks out through the front end the order will be associated with their user.

jeffstieler:
> I'll test this flow

Confirmed - manual order created in the backend with a guest results in a new customer lookup row.

jeffstieler:
@peterfabian - removed the `role` restriction in 476c632.

peterfabian (approved):
LGTM.

larssn:
This never made it to release?

rrennick:
> This never made it to release?

The PR was merged & issue closed.

> However the delete button didn't show

Could you make a new issue and post a screenshot?

PR #1297, opened 2019-01-14
Add initial state hydration for user data

### LINKED ISSUE #1250: Try to hydrate state with user data
Currently the user data is loaded asynchronously after page render.  This causes an extra flash in content after user preferences are loaded for any component that uses user preferences:
* Dashboard charts (hidden charts and intervals)
* Leaderboards (rows and hidden leaderboards)
* Report tables (hidden columns)

We should try and load these into the initial state so they're available at run time.  We can probably do this by [localizing the script](https://codex.wordpress.org/Function_Reference/wp_localize_script) so we can get the initial data from the server..  Alternatively, we could wait to load these charts until the user fetch data is ready, but this will increase perceived load times and worsen user experience.

**To Reproduce**

1. Go to the dashboard.
2. Set user preferences for hidden charts, leaderboards, or number of rows.
3. Refresh the page.
4. Note the initial load and subsequent load after user preferences are fetched.

**Screenshots**
doubleload

joshuatf:
Fixes #1250 

Adds in the user data from the server prior to fetching in the state.  This allows us to load initial user preferences without the secondary flash after user data is fetched.

1.  Visit the following places that contain user preferences:
* Dashboard: Rows per leaderboard
* Dashboard: Shown leaderboards
* Dashboard: Shown charts
* Reports: Shown user columns
2. Note that the user preferences are reflected on initial load and a subsequent flash and API request is not made after user preferences are loaded asynchronously  🚤 
3. Toggle user preferences and make sure they load correctly on refresh.

### DIFF
--- a/client/wc-api/user/selectors.js
+++ b/client/wc-api/user/selectors.js
@@ -8,7 +8,8 @@ import { DEFAULT_REQUIREMENT } from '../constants';
 const getCurrentUserData = ( getResource, requireResource ) => (
 	requirement = DEFAULT_REQUIREMENT
 ) => {
-	return requireResource( requirement, 'current-user-data' ).data || {};
+	const initalState = wcAdminServer.currentUserData; // eslint-disable-line
+	return requireResource( requirement, 'current-user-data' ).data || initalState;
 };
 
 export default {
--- a/client/wc-api/user/selectors.js
+++ b/client/wc-api/user/selectors.js
@@ -8,7 +8,7 @@ import { DEFAULT_REQUIREMENT } from '../constants';
 const getCurrentUserData = ( getResource, requireResource ) => (
 	requirement = DEFAULT_REQUIREMENT
 ) => {
-	const initalState = wcAdminServer.currentUserData; // eslint-disable-line no-undef
+	const initalState = wcSettings.currentUserData;
 	return requireResource( requirement, 'current-user-data' ).data || initalState;
 };
 
--- a/client/wc-api/user/selectors.js
+++ b/client/wc-api/user/selectors.js
@@ -8,7 +8,7 @@ import { DEFAULT_REQUIREMENT } from '../constants';
 const getCurrentUserData = ( getResource, requireResource ) => (
 	requirement = DEFAULT_REQUIREMENT
 ) => {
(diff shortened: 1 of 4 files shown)


client/wc-api/user/selectors.js
```diff
@@ -8,7 +8,8 @@ import { DEFAULT_REQUIREMENT } from '../constants';
 const getCurrentUserData = ( getResource, requireResource ) => (
 	requirement = DEFAULT_REQUIREMENT
 ) => {
-	return requireResource( requirement, 'current-user-data' ).data || {};
+	const initalState = wcAdminServer.currentUserData; // eslint-disable-line
```
Aljullu: Not a big deal, but what do you think about only disabling `no-undef` rule, so other eslint rules will still be applied to that line:
```suggestion
	const initalState = wcAdminServer.currentUserData; // eslint-disable-line no-undef
```

Aljullu (approved):
Love this PR, that improves the user experience a lot, good job! :clap: 

I added a small suggestion, but looks good to be merged. :rocket:

lib/admin.php
```diff
@@ -227,6 +227,23 @@ function wc_admin_enqueue_script() {
...
+		$current_user_data[ $user_field ] = get_user_meta( get_current_user_id(), 'wc_admin_' . $user_field, true );
+	}
+
+	wp_localize_script(
+		WC_ADMIN_APP,
+		'wcAdminServer',
+		array( 'currentUserData' => $current_user_data )
+	);
+}
+add_action( 'admin_enqueue_scripts', 'wc_admin_hydrate_state' );
```
jeffstieler: What do you think about adding `currentUserData` to the `wcSettings` object we're already creating? https://github.com/woocommerce/wc-admin/blob/eaffb0d607f48c3999f2a0dc834e0ea98d82f984/lib/client-assets.php#L139

joshuatf: I like the idea of consolidating where this code gets placed, but I'm also a little wary of adding this to `wcSettings` since I believe that's traditionally used for global settings and user specific settings are kept separate.

It looks like we're already some user specific settings to the `wcSettings` global, so I'll append it there for now and we can re-address later if we need to separate.

PR #1299, opened 2019-01-14
Replace 'extract-text-webpack-plugin' with 'mini-css-extract-plugin'

Aljullu:
Webpack [recomends](https://webpack.js.org/plugins/extract-text-webpack-plugin/#usage) using `mini-css-extract-plugin` instead of `extract-text-webpack-plugin` to load (S)CSS files since Webpack 4.

- Run `npm start`.
- Verify there isn't any message in the console that reads as follows:
`Entrypoint undefined = extract-text-webpack-plugin-output-filename`
- Verify there are no regressions anywhere else.


timmyc (approved):
Looks good and testing well for me. This aligns also with what is in `woocommerce-gutenberg-product-blocks` too 👍

PR #1301, opened 2019-01-14
Fix watch package files in Linux

Aljullu:
`fs.watch()` [doesn't work recursively in Linux](https://nodejs.org/api/fs.html#fs_caveats), that made modifying `packages` files a nightmare, because you had to call `npm start` or `npm build:packages` everytime you modified a file.

There is a NPM package that seems to work the same as `fs.watch` and fixes that and other issues:

https://github.com/yuanchuan/node-watch

It would be great to verify this PR doesn't cause any regression in Mac OS/Windows:
- Run `npm start`.
- Make a change in a file inside `packages` folder.
- Refresh the browser and verify the change has been applied.

### DIFF
--- a/bin/packages/watch.js
+++ b/bin/packages/watch.js
@@ -5,6 +5,7 @@ const fs = require( 'fs' );
 const { execSync } = require( 'child_process' );
 const path = require( 'path' );
 const chalk = require( 'chalk' );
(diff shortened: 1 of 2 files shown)


timmyc (approved):
Nice improvement here to make linux devs have an easier time, thanks Albert.

PR #1302, opened 2019-01-15
Fix javascript conflict with Gutenberg Products block plugin.

jeffstieler:
Rename `wcSettings` object to avoid collisions with the [WooCommerce Gutenberg Products block plugin](https://github.com/woocommerce/woocommerce-gutenberg-products-block).

- On `master`
- Install + Activate the products block plugin
- Verify that wc-admin screens do not render
- Checkout this branch
- Verify that wc-admin screens do render

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -13,7 +13,7 @@
 	"globals": {
 		"wp": true,
 		"wpApiSettings": true,
-		"wcSettings": true,
+		"wcAdminSettings": true,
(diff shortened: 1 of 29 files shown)


joshuatf (dismissed):
Seems like a smart move and fixes the blank screen.  :shipit: 

Also, not sure if you meant to include or not, but the `package-lock.json` was committed.

PR #1305, opened 2019-01-15
Advanced Filters: append time stamp to dates for API query

psealock:
API expects timestamped values for advanced filter parameters such as `registered_before` or `last_active_between`. This PR adds timestamps for applicable filters.

https://github.com/woocommerce/wc-admin/pull/1267 set a pattern of using a true array for the `between` rule, so changes reflect this direction.

Lastly, `updateFilter` is introduced to `AdvancedFilters` methods due to oddities of using `onFilterChange` multiple times to amend different properties of the same filter. 

### Testing

1. Customers Report > Show > Advanced Filters.
2. Add either _Registered_ or _Last active_ filters.
3. Using values for before, after, and between ensure the resulting API query parameters reflect timestamps.
4. Check that tabular results reflect the filters you've set.

### DIFF
--- a/client/store/reports/test/utils.js
+++ b/client/store/reports/test/utils.js
@@ -11,6 +11,7 @@ import {
 	getSummaryNumbers,
 	getFilterQuery,
 	getReportTableData,
+	timeStampFilterDates,
 } from '../utils';
 import * as ordersConfig from 'analytics/report/orders/config';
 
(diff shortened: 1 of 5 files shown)


client/store/reports/test/utils.js
```diff
@@ -527,3 +528,71 @@ describe( 'getReportTableData()', () => {
...
+
+	it( 'should append timestamps to activeFilters using the Date component', () => {
+		const activeFilter = {
+			key: 'my_date',
+			rule: 'before',
+			value: '2018-04-04',
+		};
+		const timeStampedActiveFilter = timeStampFilterDates( advancedFilters, activeFilter );
+
+		expect( timeStampedActiveFilter.value ).toHaveLength( 25 );
```
timmyc: Any reason for checking the length vs the value here? Could we at least have a comment or another test with what the value should be?

client/store/reports/test/utils.js
```diff
@@ -527,3 +528,71 @@ describe( 'getReportTableData()', () => {
...
+	} );
+
+	it( 'should append start of day for "after" rule', () => {
+		const activeFilter = {
+			key: 'my_date',
+			rule: 'after',
+			value: '2018-04-04',
+		};
+		const timeStampedActiveFilter = timeStampFilterDates( advancedFilters, activeFilter );
+		expect( timeStampedActiveFilter.value ).toContain( 'T00:00:00+00:00' );
```
timmyc: Also think this would read a bit better if it explicitly had an equality test for the value here and in the before test below too. But just my .02 :)

client/store/reports/utils.js
```diff
@@ -46,6 +46,41 @@ export function getFilterQuery( endpoint, query ) {
...
+ * Add timestamp to advanced filter parameters involving date. The api
+ * expects a timestamp for these values similar to `before` and `after`.
+ *
+ * @param {object} config - advancedFilters config object.
+ * @param {object} activeFilter - an active filter.
+ * @returns {object} - an active filter with timestamp added to date values.
+ */
+export function timeStampFilterDates( config, activeFilter ) {
+	const advancedFilterConfig = config.filters[ activeFilter.key ];
+	if ( 'Date' === get( advancedFilterConfig, [ 'input', 'component' ] ) ) {
```
timmyc: Instead of having all this logic in the if clause, what do you think about testing for `Date` first, and if falsey, `return activeFilter` - essentially [return early](https://github.com/Automattic/wp-calypso/blob/master/docs/coding-guidelines/javascript.md#indentation-and-line-breaks) when possible.

timmyc (commented):
Looking good, just a couple of minor comments.

client/store/reports/utils.js
psealock: Thanks @timmyc. Are you suggesting this?

```js
if ( 'Date' !== get( advancedFilterConfig, [ 'input', 'component' ] ) ) {
    return activeFilter;
}
if ( Array.isArray( value ) ) {
    // do some stuff
   return something;
}

// do some other stuff
return anotherThing;
```

psealock:
Thanks for the suggestions @timmyc! I changed the tests for explicit equality checks and refactored `timeStampFilterDates` for better readability.

timmyc (approved):
Thanks for the updates Paul, looking good and testing great.

PR #1311, opened 2019-01-15
Remove `barfocus` elements from charts

Aljullu:
Bar charts were rendering a `.barfocus` element for each day, it was used to show the background color when the day was hovered/focused.

There was another element, named `.barmouse`, that was used to detect `hover` events.

This PR merges both of them.

### Accessibility

### Screenshots
_(This PR shouldn't produce any visible change)_
image

- Go to any report page and change the chart type to _Bar chart_.
- Hover it and navigate it with the keyboard.
- Verify there are no regressions: bars are still hoverable/focusable and the background color changes when they are hovered/focused.
- In your browser devtools search for `.barmouse` and verify there are 0.


timmyc (approved):
Nice trash pickup here! Testing out well for me with mouse and keyboard interactions both.

PR #1312, opened 2019-01-15
Add missing thousands separator in charts Y-axis

### LINKED ISSUE #1303: Chart: y-axis missing number formatting
When viewing a chart with y-axis ticks > 1000, no thousands separator is shown:

image

**To Reproduce**

1. I imported sales data from wc.com to get the above chart, or you could smooth generate a large number of data, or create a test order with a thousand items or something.
2. Open a report with a chart

**Expected behavior**
I expected the axis ticks to format the numbers properly.

Aljullu:
Fixes #1303.

Adds the missing thousands' separator in Y-axis percentage and number labels.

### Screenshots
_Before:_
image

_After:_
image

Quoting #1303:
> 1. I imported sales data from wc.com to get the above chart, or you could smooth generate a large number of data, or create a test order with a thousand items or something.
> 2. Open a report with a chart
- Verify the thousands' separator is displayed.

### DIFF
--- a/packages/components/src/chart/index.js
+++ b/packages/components/src/chart/index.js
@@ -272,13 +272,13 @@ class Chart extends Component {
 
 		switch ( valueType ) {
 			case 'average':
-				yFormat = '.0f';
+				yFormat = ',.0f';
 				break;
 			case 'currency':
 				yFormat = '$.3~s';
 				break;
 			case 'number':
-				yFormat = '.0f';
+				yFormat = ',.0f';
 				break;
 		}
 		return (


packages/components/src/chart/index.js
```diff
@@ -272,13 +272,13 @@ class Chart extends Component {
 
 		switch ( valueType ) {
 			case 'average':
-				yFormat = '.0f';
+				yFormat = ',.0f';
```
timmyc: I'm not really savvy with d3 formatting, but does this somehow pull in the user's locale and format properly?

Aljullu: Yes, it does. This comma here will be replaced by a dot or any other symbol depending on the user locale format.

On top of the file we specify the default format:

https://github.com/woocommerce/wc-admin/blob/master/packages/components/src/chart/index.js#L29-L34

However, I tried experimenting with other locales and it looks like we don't honor them, and we are always using the default format. I filled a bug for that:  #1322.

timmyc (approved):
Thanks for the details about locale, and 👍 on the follow up issue. I dropped it in this sprint in case we have time for it.

PR #1325, opened 2019-01-16
Customers Report: remove extended data call

psealock:
Fixes https://github.com/woocommerce/wc-admin/issues/1306

Remove `extendItemsMethodNames` from Customers Report Table. Recent API work removes the requirement for the second call for more information.

This is the only instance of `extendTableData` but I think we should keep it the repo because it can be useful for extension developers. cc @Aljullu 

1. Customers Report
2. Using the Network tab, see only one request for `customers` keyword instead of two.

### DIFF
--- a/client/analytics/report/customers/table.js
+++ b/client/analytics/report/customers/table.js
@@ -115,9 +115,11 @@ export default class CustomersReportTable extends Component {
 				name
 			);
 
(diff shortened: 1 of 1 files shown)


timmyc (approved):
Confirmed the odd ,,,,,,,,, request is not sent, LGTM

PR #1328, opened 2019-01-16
Advanced Filters: Ranges: Allow update only on complete selection

psealock:
Fixes #

Advanced Filters enables the "Filter" button when a filter has all values defined AND it will produce a new url that is different than the current one.

This PR updates the logic for "between" values (Numerical or Date. Although testing with Date ranges won't work until #1305 gets merged). You can only update the filters if both max and min (or before/after) values are set.

### Screenshots

screen shot 2019-01-17 at 12 50 26 pm

1. Customers Report > Show > Advanced Filters > Add a filter > No. of Orders.
2. See the "Filter" button disabled.
3. Add a minimum value.
4. See the "Filter" button disabled.
5. Add a maximum value.
6. See the "Filter" button enabled.

### DIFF
--- a/packages/navigation/src/test/filters.js
+++ b/packages/navigation/src/test/filters.js
@@ -126,6 +126,20 @@ describe( 'getQueryFromActiveFilters', () => {
 		expect( nextQuery.with_select_is ).toBeUndefined();
 		expect( nextQuery.with_search_includes ).toBeUndefined();
 	} );
+
+	it( 'should only reflect complete filters with multiple values', () => {
+		const activeFilters = [
+			{ key: 'valid_date', rule: 'between', value: [ '2018-04-04', '2018-04-10' ] },
+			{ key: 'invalid_date_1', rule: 'between', value: [ '2018-04-04', undefined ] },
+			{ key: 'invalid_date_2', rule: 'between', value: '2018-04-04' },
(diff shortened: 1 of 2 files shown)


packages/navigation/src/test/filters.js
```diff
@@ -126,6 +126,18 @@ describe( 'getQueryFromActiveFilters', () => {
 		expect( nextQuery.with_select_is ).toBeUndefined();
 		expect( nextQuery.with_search_includes ).toBeUndefined();
 	} );
+
+	it( 'should only reflect complete filters with multiple values', () => {
```
joshuatf: Thanks for this test 👍

joshuatf (commented):
Looks good and is testing well.

One thing I noticed was that if you use another match type (i.e., less than or more than) and enter a value to enable a filter, the filter is still enabled if you switch to `between` and allow you to submit a bad value.

Not sure if that's something we want to address here or another PR.

psealock:
Nice find @joshuatf ! I hadn't factored in that scenario. I added an explicit check for the `between` rule instead of just the presence of an array for the value. I also updated the test to reflect this case.

jeffstieler (approved):
Code looks good and tests well! :shipit:

Issue #1295, opened 2019-01-11
Reduce Margins on Small Viewports


timmyc:
[Original Report](https://github.com/woocommerce/wc-admin/issues/1098#event-2028206501)

See more of what matters on mobile phones: Reduce the top and bottom margin around the filters on mobile (16px)

before:
<img width="447" alt="screen shot 2018-12-14 at 5 03 52 pm" src="https://user-images.githubusercontent.com/4500952/50037026-68b17a00-ffc2-11e8-9e5f-24ca6d4d6f3b.png">

After:
<img width="449" alt="screen shot 2018-12-14 at 5 03 41 pm" src="https://user-images.githubusercontent.com/4500952/50037030-71a24b80-ffc2-11e8-9786-8ec6cf06f6ec.png">

PR #1329, opened 2019-01-17
Reports: Reduce Margins on Small Viewports

psealock:
Fixes https://github.com/woocommerce/wc-admin/issues/1295

Reduce margin bottom on `.woocommerce-filters__basic-filters` for small viewports

### Screenshots

screen shot 2019-01-17 at 1 10 20 pm

1. Go any report
2. See the margin between the last filter and the rest of the report get smaller for `<782px`

### DIFF
--- a/client/layout/style.scss
+++ b/client/layout/style.scss
@@ -11,6 +11,10 @@
 	@include breakpoint( '>960px' ) {
 		margin-top: 100px;
 	}
+
+	@include breakpoint( '<782px' ) {
+		margin-top: 60px;
+	}
 }
 
 .woocommerce-layout .woocommerce-layout__main {
--- a/packages/components/src/filters/advanced/style.scss
+++ b/packages/components/src/filters/advanced/style.scss
@@ -27,7 +27,7 @@
 	}
 
 	@include breakpoint( '<400px' ) {
-		margin: $gap-smaller 0;
+		margin: $gap-small 0;
 	}
 }
 
@@ -210,6 +210,6 @@
 	}
 
 	.separator {
-		padding: 0 8px;
+		text-align: center;
 	}
 }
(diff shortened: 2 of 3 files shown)


joshuatf (commented):
Nice!  A lot more usable on mobile.  Two quick notes:

* I think we may need to bring down the margin on top a bit to match styling in #1295.  Looks to be about 60px from the top.
```scss
.woocommerce-layout__primary {
	margin: 60px 0 0 $fallback-gutter-large;
	margin: 60px 0 0 $gutter-large;

	@include breakpoint( '>960px' ) {
		margin-top: 100px;
	}
}
```

* When an advanced filter is added, we get those larger gaps again.  Maybe we should apply `$gap-small` on those too? /cc @LevinMedia 
screen shot 2019-01-17 at 1 55 03 pm

psealock:
Good suggestions @joshuatf 

* Brought down the margin top on `.woocommerce-layout__primary`
* Reduced the margin top/bottom when Advanced Filters are added
* Reapplied left/right border on Advanced Filters card (it disappears for full width cards)

LevinMedia:
Looking great! Rather than wrapping to a new line when we have a "Between" selected, can we keep it on a single line? 

screen shot 2019-01-17 at 12 42 54 pm

screen shot 2019-01-17 at 12 43 51 pm

psealock:
@LevinMedia I pushed a change to make "between" selection inputs on the same line, which looks much better.

For viewports `<400px` there isn't really a way to make it fit:

screen shot 2019-01-18 at 11 09 48 am

One solution is to move the "X":

screen shot 2019-01-18 at 11 09 18 am

What do you think?

jeffstieler:
Ping @LevinMedia for a response to https://github.com/woocommerce/wc-admin/pull/1329#issuecomment-455358293

LevinMedia:
thanks @jeffstieler 

@psealock I think that last option with the X above, opposite the filter title is a good solution. Lets give that a shot. 

Thanks!

LevinMedia:
@psealock looks good to me! Thank you!

timmyc (approved):
LGTM

PR #1334, opened 2019-01-17
Add autocompleter input type and fix load options

### LINKED ISSUE #1331: Order number undefined in filter
**Describe the bug**
When adding an order number filter in the downloads report, the console shows a map error for any non-integer value.

**To Reproduce**

1. Go to the downloads report and select the advanced filter `/wp-admin/admin.php?page=wc-admin#/analytics/downloads?filter=advanced`.
2. Select the `Order Number` filter.
3.  Type any string or remove any value after typing.
4.  Note the console error.

**Expected behavior**
Expected no error and this field should probably only allow numeric entry.

**Screenshots**
screen shot 2019-01-17 at 11 09 56 am
screen shot 2019-01-16 at 5 33 55 pm

joshuatf:
Fixes #1331 

Adds an `inputType` option to autocompleters to allow `number` as a type for the input and restrict input.  Also fixes a console mapping error when data returned from the API is undefined.

### Screenshots
screen shot 2019-01-17 at 1 03 08 pm

1.  Visit the downloads report with advanced filters. `wp-admin/admin.php?page=wc-admin#/analytics/downloads?filter=advanced`
2.  Add the Order Number filter.
3.  Make sure the input type is number and only allows numeric entry.
4.  Check that no error is thrown when deleting text entirely from the input box.
5.  Make sure other autocompleter fields are still working as expected.

### DIFF
--- a/packages/components/src/search/autocomplete.js
+++ b/packages/components/src/search/autocomplete.js
@@ -146,6 +146,9 @@ export class Autocomplete extends Component {
 		const promise = ( this.activePromise = Promise.resolve(
 			typeof options === 'function' ? options( query ) : options
 		).then( optionsData => {
+			if ( ! optionsData ) {
+				return;
+			}
 			const { selected } = this.props;
 			if ( promise !== this.activePromise ) {
 				// Another promise has become active since this one was asked to resolve, so do nothing,
--- a/packages/components/src/search/autocompleters/orders.js
+++ b/packages/components/src/search/autocompleters/orders.js
@@ -24,6 +24,7 @@ import { computeSuggestionMatch } from './utils';
 export default {
 	name: 'orders',
 	className: 'woocommerce-search__order-result',
+	inputType: 'number',
 	options( search ) {
 		let payload = '';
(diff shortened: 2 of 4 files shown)


Aljullu (approved):
This is working as expected and looks good to me.

I see Chrome displays arrow up/down buttons in inputs of `type="number"`:
image

I wonder if they make any sense or just pollute the UI. What's your opinion about adding this CSS to hide them [here](https://github.com/woocommerce/wc-admin/blob/master/packages/components/src/search/style.scss#L55)? :thinking: 
```CSS
&[type="number"]::-webkit-outer-spin-button,
&[type="number"]::-webkit-inner-spin-button {
	-webkit-appearance: none;
	margin: 0;
}
```

joshuatf:
@Aljullu I had a similar thought on these since typically you would not search for an order number by scrolling.

I was hesitant to add this for all number inputs, but on second thought I can't think of a use case for autocompleters where we'd want to scroll through numbers.  Added this in for now and we can revisit later if it becomes an issue.

PR #1335, opened 2019-01-17
REST API: Allow sorting by stock_status in stocks endpoint

### LINKED ISSUE #966: /wc/v3/reports/stock should allow sorting products by `stock_status`
For the Stock report (#174) we will sort the products and variations by `stock_status` by default.

The current API endpoint only allows sorting by `stock_quantity`, `date`, `id`, `include`, `title` and `slug`.

joshuatf:
Fixes #966 

Allows sorting by `stock_status` on stocks endpoint.

1.  Make a request to `/wp-json/wc/v3/reports/stock`
2.  Pass `stock_status` as the `orderby` param.
3.  Check that sorting returns by status.
4.  Change the `order` param to `desc` and check that sorting by status is reversed.

### DIFF
--- a/client/analytics/report/stock/table.js
+++ b/client/analytics/report/stock/table.js
@@ -43,12 +43,13 @@ export default class StockReportTable extends Component {
 			{
 				label: __( 'Status', 'wc-admin' ),
 				key: 'stock_status',
+				isSortable: true,
+				defaultSort: true,
 			},
 			{
 				label: __( 'Stock', 'wc-admin' ),
 				key: 'stock_quantity',
 				isSortable: true,
-				defaultSort: true,
 			},
 		];
 	}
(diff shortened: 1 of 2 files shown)


Aljullu (commented):
This is testing well for me. But would it be possible to use `stock_quantity` as the second sort value?

Sorting by `stock_status` with `order=desc`, `Out of stock` products appears on top, so I would expect the products with a lower `stock_quantity` to appear above:

image

Same, when sorting in the other direction (when `In stock` products appear on top), I would expect products with the higher `stock_quantity` to appear above).

I don't know if that's possible, but that's what I would expect as a user.

I created #1340 as a follow-up of this PR, it updates the UI to use `stock_status` by default. Feel free to merge in into your branch.

joshuatf:
@Aljullu That seems reasonable.  I added a commit to do this and also reversed the order of status quantity since it seems like you'd want to see the larger (absolutely bigger values) stock values first when sorting with this method.  Let me know what you think.

Aljullu (commented):
Thanks for making these changes @joshuatf, it looks much better now. The only last issue I noticed is that you need to click two times on the _Status_ header to reverse the order, now:

status

claudiosanches (approved):
PHP part looks good.

Aljullu (approved):
> I noticed that issue across a number of reports as it seems the order is disconnected from `tableQuery` and `query`. Opened a separate issue here to handle that. Let me know if that sounds okay to you.

Thanks for filling an issue for that @joshuatf!

PR #1337, opened 2019-01-17
Add country code as keyword for countries search

### LINKED ISSUE #1272: Customers report: table displays country code but filter uses country names

joshuatf:
Fixes #1272 

Adds the country code to the list of searched keywords and adds a tooltip with the country name to the table.

### Accessibility

### Screenshots
screen shot 2019-01-17 at 4 18 04 pm
screen shot 2019-01-17 at 4 17 42 pm

1. Visit the customers report page and add the advanced filter `wp-admin/admin.php?page=wc-admin#/analytics/customers?filter=advanced`.
2.  Select "Country" as a filter option.
3.  Type a country code that contains letters not in its actual name (e.g., GQ for Equatorial Guinea).
4.  Note the country is shown in the dropdown list.
5.  Hover over the country code in the customers table and note the tooltip with the country name.
6.  Using a screen reader, check that the country name is read when navigating through the country column.

### Questions
* When typing something like `ES` (the country code for Spain), there are enough other matches that Spain is still never shown since results are alphabetical and the maximum shown is 10.  Should we:
  * Allow more than 10 results.
  * Sort by best match (probably best in a follow-up PR).
  * Leave as is.
* Should we include the country code in the label of the search dropdown?  (e.g., Spain (ES) ).

### DIFF
--- a/client/analytics/report/customers/table.js
+++ b/client/analytics/report/customers/table.js
@@ -3,7 +3,8 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { Component } from '@wordpress/element';
+import { Component, Fragment } from '@wordpress/element';
+import { Tooltip } from '@wordpress/components';
 
 /**
  * WooCommerce dependencies
@@ -90,6 +91,12 @@ export default class CustomersReportTable extends Component {
 		];
 	}
 
+	getCountryName( code ) {
+		const countries = ( wcSettings.dataEndpoints && wcSettings.dataEndpoints.countries ) || [];
+		const country = countries.find( c => c.code === code );
+		return country ? country.name : null;
+	}
+
 	getRowsContent( customers ) {
 		return customers.map( customer => {
 			const {
@@ -106,6 +113,7 @@ export default class CustomersReportTable extends Component {
 				city,
 				country,
 			} = customer;
+			const countryName = this.getCountryName( country );
 
 			const customerNameLink = user_id ? (
 				<Link href={ 'user-edit.php?user_id=' + user_id } type="wp-admin">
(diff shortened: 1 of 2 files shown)


Aljullu (approved):
Thanks for taking this issue! It's working well for me and it's a big improvement for screen readers. :clap: 

> When typing something like ES (the country code for Spain), there are enough other matches that Spain is still never shown since results are alphabetical and the maximum shown is 10. Should we: 

In my opinion, we should only display countries whose country code matches the string or any of the words that are part of the name start with that string.

So writing `es` should return _Estonia_ and _Spain_ but shouldn't return _Maldiv**es**_ or _French Polyn**es**ia_.

Another example: `po` should return _**Po**land_, _**Po**rtugal_ and _French **Po**lynesia_ but not _Singa**po**re_.

How does it sound? We can open a follow-up for this.

> Should we include the country code in the label of the search dropdown? (e.g., Spain (ES) ).

I don't think that's needed, but maybe @LevinMedia has another opinion? I noticed however _United States (US)_ and _United Kingdom (UK_) already have it.

timmyc:
Nice one! I agree with @Aljullu sentiments above. Not sure how high priority it is though, but worth creating a follow up issue for.

joshuatf:
Thanks for the review @Aljullu!  I like that suggestion; issue noted here - https://github.com/woocommerce/wc-admin/issues/1348

PR #1338, opened 2019-01-17
Remove wc-api customers

Aljullu:
After #1325 was merged, I think we don't need these files anymore.

- Verify there are no regressions and data is loading fine.


timmyc (approved):
NIce! Testing well for me

PR #1341, opened 2019-01-17
Set default order to 'desc' in time tables

Aljullu:
Makes time-ordered tables to display most recent rows first by default.

### Screenshots
_Before:_
image

_After:_
image

- Go to the _Revenue_ report.
- Verify most recent dates are the first ones displayed in the table.
- Repeat the same in the _Orders_, the _Downloads_ nd the _Customers_ report.


timmyc (approved):
Seems like a reasonable fix to me. Looks good.

PR #1343, opened 2019-01-17
Add performance indicators to dashboard

### LINKED ISSUE #1281: REST API: Add Report Link and Name to /preformance-indicators
This is a follow up to #1255 - the Store Performance Indicators endpoint. After having some conversations with @justinshreve about this, it kind of seems like it would be nice for the `/wc/v3/reports/performance-indicators` to return a label/name for each stat, and a link to the full report page for the given stat also because the Performance Indicators in the UI will ultimately link to their associated report/stat combo.

I'm thinking we might be able to leverage the existing schema on each of the stats endpoint to gather this data since we are already doing an `OPTIONS` [request to each endpoint](https://github.com/woocommerce/wc-admin/blob/master/includes/api/class-wc-admin-rest-reports-performance-indicators-controller.php#L66) we could gather the `description` attribute on each stat, and possibly add `report` to the `_links` portion fo the response ( though tbh not sure if this is proper RESTful use ):

options-request

But totally open to other ideas here. Ultimately though, if we can have all needed info returned by the `/performance-indicators` endpoint ( values, link to report, stat name ) then the performance indicators section of the dash can be fully powered by the API and be completely extendable via PHP ✨

### LINKED ISSUE #1282: Dashboard: Add All Store Performance Options
Since #1255 has introduced the new `/performance-indicators` endpoint, we can now hook up _Real_ data to _Store Performance_ section of the dashboard. The following is a list of the indicators that should be available ( via p6riRB-3qG-p2), and the associated report/tab combo they should link to:

image

For this issue we could populate the list of available indicators within the JS component, or consider doing #1281 first and making it entirely API driven first - up to whomever picks this up!

emdashcodes:
Fixes #1282. Closes #1281.

This PR hooks up the performance indicators on the dashboard to the new performance indicator REST API.

It also makes some more adjustments to the REST API to make this possible including report links, description/label, chart sorting, and a method of indicating which total should be a performance indicator and of what format.

A few stats off the list in #1282 are missing, because they do not exist in the reports/REST API yet. Once added, they will automatically show up here.

🌟 This also supports extensions. You can test `Follow Up Emails` via https://github.com/woocommerce/woocommerce-follow-up-emails/pull/698 with this. 🌟

### Accessibility

### Screenshots

screen shot 2019-01-17 at 11 44 32 am

screen shot 2019-01-17 at 11 44 44 am

* View the dashboard, test different date ranges, and verify performance indicators show.
* Try showing/hiding indicators and testing defaults. You can clear out your usermeta (key dashboard_performance_indicator) to test defaults.
* Bonus points: Test with follow up emails.


timmyc:
> This also supports extensions. You can test Follow Up Emails via woocommerce/woocommerce-follow-up-emails#698 with this.

❗️❗️❗️❗️❗️

client/dashboard/store-performance/index.js
```diff
@@ -2,9 +2,15 @@
 /**
  * External dependencies
  */
-import { __ } from '@wordpress/i18n';
+import { __, sprintf } from '@wordpress/i18n';
 import { ToggleControl } from '@wordpress/components';
 import { Component, Fragment } from '@wordpress/element';
+import { compose } from '@wordpress/compose';
+import { withDispatch } from '@wordpress/data';
+import { getCurrentDates, appendTimestamp, getDateParamsFromQuery } from '@woocommerce/date';
```
timmyc: Minor but it seems like in some components we have been adding a `/** WooCommerce Dependencies ` block for internal packages.

client/dashboard/store-performance/index.js
```diff
@@ -16,99 +22,196 @@ import {
...
 
+					if ( ! primaryItem || ! secondaryItem ) {
+						return null;
+					}
+
+					const href =
+						( primaryItem._links &&
+							primaryItem._links.report[ 0 ] &&
+							primaryItem._links.report[ 0 ].href ) ||
+						'';
```
timmyc: Should this and `reportUrl` default to `false` so `SummaryNumber` knows there isn't a link if one was not supplied?

client/lib/number/index.js
```diff
@@ -1,3 +1,11 @@
+/** @format */
```
timmyc: Nice I think this is a good location for this logic.

includes/api/class-wc-admin-rest-reports-performance-indicators-controller.php
```diff
@@ -46,12 +107,15 @@ protected function prepare_reports_query( $request ) {
...
+	 * Get information such as allowed stats, stat labels, and endpoint data from stats reports.
 	 *
-	 * @return array
+	 * @return WP_Error|True
 	 */
-	public function get_allowed_stats() {
-		global $wp_rest_server;
+	private function get_indicator_data() {
+		// Data already retrieved.
+		if ( ! empty( $this->endpoints ) && ! empty( $this->labels ) && ! empty( $this->allowed_stats ) ) {
```
timmyc: Nice!

includes/api/class-wc-admin-rest-reports-revenue-stats-controller.php
```diff
@@ -136,51 +136,61 @@ public function get_item_schema() {
...
 			'orders_count'   => array(
-				'description' => __( 'Amount of orders', 'wc-admin' ),
+				'description' => __( 'Amount of orders.', 'wc-admin' ),
 				'type'        => 'integer',
 				'context'     => array( 'view', 'edit' ),
 				'readonly'    => true,
 			),
 			'num_items_sold' => array(
-				'description' => __( 'Amount of orders', 'wc-admin' ),
+				'description' => __( 'Items sold.', 'wc-admin' ),
```
timmyc: Nice little schema trash pickup here too!

lib/client-assets.php
```diff
@@ -153,7 +153,8 @@ function wc_admin_print_script_settings() {
 	}
 
 	$preload_data_endpoints = array(
-		'countries' => '/wc/v3/data/countries',
+		'countries'              => '/wc/v3/data/countries',
```
timmyc: should these be `/v4` now?

lib/client-assets.php
```diff
@@ -211,6 +214,27 @@ function wc_admin_print_script_settings() {
...
+/**
+ * Sets default values for user preferences.
+ *
+ * @param array $user_data Array of user data.
+ * @return array Filtered array of user data.
+ */
+function wc_admin_get_user_defaults( $user_data ) {
+	// If no settings for performance indicators are stored, these are the defaults to disable.
+	if ( ! is_array( $user_data['dashboard_performance_indicators'] ) ) {
+		$user_data['dashboard_performance_indicators'] = array(
```
timmyc: Nice! /cc @LevinMedia do these look like good defaults for Performance Indicators?

packages/components/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Fix to avoid duplicated Y-axis ticks when the Y max value was 0.
 - Remove decimals from Y-axis when displaying currencies.
 - Fix date formatting on charts in Safari.
+- Make `href`/linking optional in SummaryNumber.
```
timmyc: I just published `1.4.0` yesterday so this will need to be under a new "unreleased" heading in this file for `1.4.1`

packages/components/src/summary/number.js
```diff
@@ -147,7 +153,7 @@ SummaryNumber.propTypes = {
 };
 
 SummaryNumber.defaultProps = {
-	href: '/analytics',
+	href: '',
```
timmyc: Ah disregard my prior `false` comment about the urls above.

timmyc (commented):
Amaze.

Some minor comments in the code - I think the only real change might be API versioning. Tests are passing for me, and thanks for adding/updating them.

I'm going to test out the FUE extension too to get the full experience too, but did want to drop this .gif here as when I first activated this branch, and visited the dashboard, this is what I was seeing - its like there is an initial load, and then a subsequent refresh added in two more indicators. Once I toggled some off, and thus had user preferences set, I could no longer see the issue:

loading-indicators

Other than that, this is testing out wonderfully for me, and really takes the dashboard utility level up immensely. Being able to customize these stats, and quickly dive in deeper for further analysis is a real treat that I'm certain store operators will love.

timmyc:
@LevinMedia would love to have you give this one a review too.

LevinMedia:
I'm so excited to test this, I'm leaving this comment before I test this. ❤️

lib/client-assets.php
justinshreve: This got updated on the rebase, but actually breaks. I opened up https://github.com/woocommerce/wc-admin/pull/1354 to address it.

packages/components/CHANGELOG.md
justinshreve: Fixed. thanks!

lib/client-assets.php
justinshreve: Just a note that these are the ones that are **hidden** by default (per the issue). So

justinshreve: This list is now here:
https://github.com/woocommerce/wc-admin/pull/1343/commits/bcca36065894e5f3fda239bb80759e8d23e96468#diff-71321a1ffc66cd298d29c3719d481cd7R171

justinshreve:
@timmyc It took me a bit to debug/track down what was happening with the preferences but I think the issue is fixed now. I've also handled the other PR feedback and rebased this.

If the code looks OK, since this PR is a bit bigger, I wouldn't mind getting this merged and handle any larger changes / design feedback in a future PR.

lib/client-assets.php
LevinMedia: Lets make default excluded summary numbers:

```
'taxes/order_tax',
'taxes/shipping_tax',
'downloads/download_count',
```

that gives us an even 10 in the grid.

timmyc (approved):
Changes look great - verified no more flickering on first load when no prefs are set. Totally agree that any items outstanding can be done in a follow-up pr(s).

Awesome feature Justin!

lib/client-assets.php
justinshreve: Done

PR #1344, opened 2019-01-17
Remove 'wc-admin' client/store implementation

### LINKED ISSUE #1094: Cleanup client/store
After #1006 was merged, some contents under `client/store` can be removed since they are no longer used anywhere in the app.

coderkevin:
Fixes #1094

This removes the `wc-admin` wp.data store, which is no longer used by any of the components.
It's functionality has been replaced by the `wc-api` store, under client/wc-api

Still to do: There are 3 files left under `client/store` which are in use:
 - reports/utils.js
 - constants.js
 - utils.js

These should probably be moved elsewhere, but I'm not exactly sure where the best place is, and I didn't want to complicate this PR with extra stuff.

- `npm test` should run successfully
- All client behavior should be the same, no differences.


Aljullu (approved):
Nice cleanup. :shipit:

timmyc:
Nice Kevin, thanks!

timmyc:
@coderkevin looks like this needs a rebase now, but could we do that and get this landed?

justinshreve:
I've rebased and retested this. Looks good to me still. Merging 👍.

PR #1350, opened 2019-01-18
Add single category filter to products report

### LINKED ISSUE #1046: Product Report: Add filter by single category
~Determine if the~ Product Report should be filterable by a "Single Category" much like it currently is filterable by "Single Product".

### Why?

It would be nice to link the products count from the Categories table to the Products Report and look at only products of that category.

joshuatf:
Fixes #1046

Allows filtering by a single category on the products report.  Also links the product count on the categories table to the product report for a given category.

### Screenshots
screen shot 2019-01-18 at 5 05 59 pm

1.  Open the products report `/wp-admin/admin.php?page=wc-admin#/analytics/products`
2.  Filter by "Single Product Category" and search and select a category.
3.  Check that results are filtered by that category.
4.  Go to the categories report `/wp-admin/admin.php?page=wc-admin#/analytics/categories`.
5.  Click on the product count in a category row to make sure it takes you to the product report filtered by that category.

### DIFF
--- a/client/analytics/report/categories/table.js
+++ b/client/analytics/report/categories/table.js
@@ -11,6 +11,7 @@ import { map } from 'lodash';
  * WooCommerce dependencies
  */
 import { formatCurrency, getCurrencyFormatDecimal } from '@woocommerce/currency';
+import { getNewPath, getPersistedQuery } from '@woocommerce/navigation';
 import { Link } from '@woocommerce/components';
 
 /**
@@ -69,8 +70,9 @@ class CategoriesReportTable extends Component {
 	getRowsContent( categoryStats ) {
 		return map( categoryStats, categoryStat => {
 			const { category_id, items_sold, net_revenue, products_count, orders_count } = categoryStat;
-			const categories = this.props.categories;
+			const { categories, query } = this.props;
 			const category = categories[ category_id ];
+			const persistedQuery = getPersistedQuery( query );
 
 			return [
 				{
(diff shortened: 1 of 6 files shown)


client/analytics/report/categories/table.js
```diff
@@ -85,7 +86,17 @@ class CategoriesReportTable extends Component {
...
 				},
 				{
-					display: numberFormat( products_count ),
+					display: (
+						<Link
+							href={
+								'admin.php?page=wc-admin#/analytics/products?filter=single_category&categories=' +
+								category_id
+							}
+							type="wp-admin"
```
Aljullu: `<Link>` has a `wc-admin` type to make links shorter:
```ES6
						<Link
							href={
								'products?filter=single_category&categories=' +
								category_id
							}
							type="wc-admin"
						>
```

Aljullu (commented):
This is looking good to me. I just noticed a minor issue: when switching from _Single Product Category_ to _Single Product_ (or the other way around), the search input is not cleared out.

single

client/analytics/report/categories/table.js
psealock: You'll want to use `getPersistedQuery` here to make sure things like date parameters persist. 

```js
				<Link
					href={ getNewPath( persistedQuery, 'categories', {
						filter: 'single_category',
						categories: category.id,
					} ) }
					type="wc-admin"
				>
```

joshuatf: Awesome- thanks for both of these tips!

joshuatf:
>I just noticed a minor issue: when switching from Single Product Category to Single Product (or the other way around), the search input is not cleared out.

@Aljullu Good catch finding that.  Looks like `selectedTag` is never cleared out so this behavior would occur across any filter with searches.  Added a quick fix in https://github.com/woocommerce/wc-admin/pull/1350/commits/5b345c786895b3b54efe8f7abb8c903c210280d3 to check and see if the filter matches the current one, otherwise it sets `selectedTag` to null.

Aljullu (approved):
Thanks for fixing that! It looks good to me. :rocket:

PR #1351, opened 2019-01-18
Move search icon inside input wrapper

### LINKED ISSUE #1349: Search icon moves in autocomplete when results are fetched
**Describe the bug**
When results are rendered, the search icon in the autocomplete box gets moved down to vertically center within the results + input box when it should only be vertically centered within the input box.

**To Reproduce**

1. Go to the products report and select advanced filters.
2. Search by single product with a term that results in multiple products.
3. Note the search icon (magnifying glass) position after the list is rendered.

**Expected behavior**
Expected the icon to remain vertically centered inside the input box.

**Screenshots**
screen shot 2019-01-18 at 4 51 02 pm

joshuatf:
Fixes #1349 

Moves search icon inside input wrapper so we can set it's position relative to input and not the results.

### Before
screen shot 2019-01-18 at 5 34 11 pm

### After
screen shot 2019-01-18 at 5 34 54 pm

1.  Open the products report and filter by "Single Product."
2.  Type in a search query with results and check that the search icon stays in the correct place.
3.  Check other search inputs to make sure the search icon remains in the correct place.

### DIFF
--- a/packages/components/src/search/index.js
+++ b/packages/components/src/search/index.js
@@ -157,12 +157,12 @@ class Search extends Component {
 		};
 		const shouldRenderTags = this.shouldRenderTags();
 		const inputType = autocompleter.inputType ? autocompleter.inputType : 'text';
+		const searchIcon = <Gridicon className="woocommerce-search__icon" icon="search" size={ 18 } />;
 
 		return (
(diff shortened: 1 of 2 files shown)


Aljullu (approved):
Works like a charm. :rocket:

PR #1354, opened 2019-01-18
Fix country filter by adding v4 of data/countries API

emdashcodes:
In https://github.com/woocommerce/wc-admin/pull/1298, endpoints were bumped to use v4.

The code that loads country data into wcSettings was calling v4 of `/data/countries` which doesn't exist, so a notice was showing:

`PHP Notice:  Undefined index: /wc/v4/data/countries in wc-admin/lib/client-assets.php on line 201`, which stopped the autocompleter from showing any results.

To Test:
* Run this branch and make sure the notice is gone.
* Make sure the country autocompleter works under the customer report.

### DIFF
--- a/includes/api/class-wc-admin-rest-data-countries-controller.php
+++ b/includes/api/class-wc-admin-rest-data-countries-controller.php
@@ -1,0 +1,27 @@
+<?php
+/**
+ * REST API Data countries controller.
+ *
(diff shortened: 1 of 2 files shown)


timmyc (approved):
Thanks for the fix Justin, testing well for me.

PR #1360, opened 2019-01-20
Center points in single-date charts

### LINKED ISSUE #1270: Chart: Plot Single Point in Center
Via #1100 

When viewing a single day, with the interval set to day ( or any singular data point ), and the chart set to line chart - center the points in the graph

screen shot 2018-12-14 at 10 32 19 pm

Aljullu:
Fixes #1270.

Centers points in single-date charts

It also avoids creating the `<path>` element (_aka_ the line) when there is only one single date.

### Screenshots
_Before:_
image

_After:_
image

- Go to the _Revenue_ report (or any other).
- Select _Year to Date_ as the date range and _By quarter_ as the interval, for example. So only one point appears in the line chart.
- Verify it's centered in the middle of the chart.


timmyc (approved):
Looking all nice and centered now, thanks Albert

PR #1364, opened 2019-01-21
Add settings page with excluded order statuses

### LINKED ISSUE #1252: Settings page creation and excluded statuses
A "Settings" page is needed to store global settings for reporting.  Initially this will involve:
* Creating the settings page.
* Creating the "Excluded Statuses" settings.
* Storing these settings in `wcSettings`.
* Hooking up saved settings to the excluded statuses array in #1251 

screen shot 2019-01-08 at 3 45 38 pm

Before starting on this task, check with @LevinMedia for the final design on this page.

joshuatf:
Fixes #1252 

Sorry for the length of this PR; in hindsight it may have benefited from being broken down into a few different PRs.  This PR:

* Adds the Settings route and page.
* Adds a reusable `Setting` component.
* Adds settings to the wc-api to read and update settings specific to wc-admin.
* Hooks up the excluded order status settings to those used in the reports.
* Add custom settings to the `wcSettings` global.
* Adds extensibility to the `wcSettings` global and to the settings page.

### Accessibility

### Screenshots
screen shot 2019-01-23 at 1 56 16 pm

1.  Add custom statuses to your WooCommerce order statuses.  You can use [this plugin](https://wordpress.org/plugins/custom-order-statuses-woocommerce/) if you don't already have statuses added.
2.  Open the new Settings page at `wp-admin/admin.php?page=wc-admin#/analytics/settings`.
3.  Check that statuses are shown and that any custom statuses are shown as well.
4.  Change excluded statuses and save.  Refresh and make sure that settings persist.
5.  Check that reports reflect the excluded statuses chosen.
6.  Use the "Reset Defaults" button and make sure the default excluded statuses are restored (cancelled, failed, and pending).

### Design
The design in #1252 differs from some of the default WooCommerce styles and those have in wc-admin. 
 After a slack discussion with @LevinMedia we are holding off for confirmation on changes to the following before making further changes:
* Checkboxes
* Buttons
* Section header
* Success notice after save (this will probably need to be a separate PR)


willgorham:
Just my 2¢, but it seems like it would be more intuitive for the setting to be _Included_ Statuses, so you select the ones you want to view.

jeffstieler:
Testing now - on first load of the settings page, trying to select a status results in this console error: `index.js:74 Uncaught TypeError: settings[name].push is not a function at Settings.handleInputChange (index.js:74)`.

Things seem to work after clicking "reset defaults".

LevinMedia:
>Testing now - on first load of the settings page, trying to select a status results in this console error: index.js:74 Uncaught TypeError: settings[name].push is not a function at Settings.handleInputChange (index.js:74). Things seem to work after clicking "reset defaults".

I saw the same issue - when I first loaded the screen, none of the statuses were checked by default - clicking reset defaults caused the default statues to be checked.

client/analytics/settings/config.js
```diff
@@ -0,0 +1,62 @@
...
+export const analyticsSettings = applyFilters( SETTINGS_FILTER, [
+	{
+		name: 'woocommerce_excluded_report_order_statuses',
+		label: __( 'Excluded Statuses:', 'wc-admin' ),
+		inputType: 'checkboxGroup',
+		options: [
+			{
+				key: 'defaultStatuses',
+				options: orderStatuses.filter(
+					status => defaultOrderStatuses.includes( status.value ) && 'refunded' !== status.value
```
jeffstieler: Nitpick: `refunded` status is already filtered out above.

client/analytics/settings/index.js
```diff
@@ -0,0 +1,135 @@
...
+			hasError: true,
+		} );
+		/* eslint-disable no-console */
+		console.warn( error );
+		/* eslint-enable no-console */
+	}
+
+	resetDefaults = () => {
+		if (
+			window.confirm(
```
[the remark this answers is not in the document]
jeffstieler: I think that it's OK with the confirmation dialog - without, I'd want the user to click "save" as well.

client/analytics/settings/index.js
```diff
@@ -0,0 +1,135 @@
...
+
+	handleInputChange( e ) {
+		const { checked, name, type, value } = e.target;
+		const { settings } = this.state;
+
+		if ( 'checkbox' === type ) {
+			if ( checked ) {
+				settings[ name ].push( value );
+			} else {
+				remove( settings[ name ], v => v === value );
```
jeffstieler: [`_.without()`](https://lodash.com/docs/4.17.11#without) might be a good substitute here.

client/analytics/settings/setting.js
```diff
@@ -0,0 +1,141 @@
...
+import './setting.scss';
+import sanitizeHTML from 'lib/sanitize-html';
+
+class Setting extends Component {
+	renderInput = () => {
+		const { handleChange, name, inputType, options, value } = this.props;
+		const id = uniqueId( name );
+
+		switch ( inputType ) {
+			case 'checkboxGroup':
```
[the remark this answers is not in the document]
jeffstieler: I'm OK with the addition of groups, but if you do want to cut this and keep everything in one "group", perhaps add an indicator to custom statuses?

client/analytics/settings/setting.js
```diff
@@ -0,0 +1,141 @@
...
+
+		return (
+			<div className="woocommerce-setting">
+				<div className="woocommerce-setting__label">{ label }</div>
+				<div className="woocommerce-setting__options">
+					{ this.renderInput() }
+					{ helpText && (
+						<span
+							className="woocommerce-setting__help"
+							dangerouslySetInnerHTML={ sanitizeHTML( helpText ) }
```
jeffstieler: If the help text for the analytics settings used `interpolateComponents()` instead of markup, would we need `dangerouslySetInnerHTML` and `sanitizeHTML`?

client/wc-api/settings/operations.js
```diff
@@ -0,0 +1,71 @@
...
+	return [ ...readSettings( resourceNames, fetch ) ];
+}
+
+function update( resourceNames, data, fetch = apiFetch ) {
+	return [ ...updateSettings( resourceNames, data, fetch ) ];
+}
+
+function readSettings( resourceNames, fetch ) {
+	if ( resourceNames.includes( 'settings' ) ) {
+		const url = '/wc/v3/settings/wc_admin';
```
jeffstieler: Should this use `NAMESPACE`?

client/wc-api/settings/operations.js
```diff
@@ -0,0 +1,71 @@
...
+	const settingsFields = [ 'woocommerce_excluded_report_order_statuses' ];
+
+	if ( resourceNames.includes( resourceName ) ) {
+		const url = NAMESPACE + '/settings/wc_admin/';
+		const settingsData = pick( data[ resourceName ], settingsFields );
+
+		const promises = Object.keys( settingsData ).map( setting => {
+			let value = settingsData[ setting ];
+			if ( Array.isArray( value ) ) {
+				value = value.join( ',' );
```
jeffstieler: Is this necessary? _more below.._

includes/api/class-wc-admin-rest-setting-options-controller.php
```diff
@@ -0,0 +1,50 @@
...
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Setting Options controller.
+ *
+ * @package WooCommerce Admin/API
+ * @extends WC_REST_Setting_Options_Controller
+ */
+class WC_Admin_REST_Setting_Options_Controller extends WC_REST_Setting_Options_Controller {
```
jeffstieler: Could we skip this controller override if we weren't turning an array into a comma delimited string?

includes/class-wc-admin-api-init.php
```diff
@@ -350,6 +352,20 @@ public static function filter_rest_endpoints( $endpoints ) {
...
+			&& isset( $endpoints['/wc/v4/settings/(?P<group_id>[\w-]+)'][3] )
+			&& $endpoints['/wc/v4/settings/(?P<group_id>[\w-]+)'][3]['callback'][0] instanceof WC_Admin_REST_Setting_Options_Controller
+			&& $endpoints['/wc/v4/settings/(?P<group_id>[\w-]+)'][4]['callback'][0] instanceof WC_Admin_REST_Setting_Options_Controller
+			&& $endpoints['/wc/v4/settings/(?P<group_id>[\w-]+)'][5]['callback'][0] instanceof WC_Admin_REST_Setting_Options_Controller
+		) {
+			$endpoints['/wc/v4/settings/(?P<group_id>[\w-]+)'][0] = $endpoints['/wc/v4/settings/(?P<group_id>[\w-]+)'][3];
+			$endpoints['/wc/v4/settings/(?P<group_id>[\w-]+)'][1] = $endpoints['/wc/v4/settings/(?P<group_id>[\w-]+)'][4];
+			$endpoints['/wc/v4/settings/(?P<group_id>[\w-]+)'][2] = $endpoints['/wc/v4/settings/(?P<group_id>[\w-]+)'][5];
+		}
+
```
jeffstieler: Could we skip this controller override if we weren't turning an array into a comma delimited string?

jeffstieler (commented):
In addition to the console error on initial interaction, this seems to break using the "order status" advanced filter when the status is one of the excluded ones. _cc: @LevinMedia for UX on that_

client/analytics/settings/config.js
joshuatf: Good catch!  Will get this change made.

client/analytics/settings/index.js
joshuatf: Agreed- will make that change.

client/wc-api/settings/operations.js
joshuatf: Hmm, yes, this should probably be bumped to v4.

includes/api/class-wc-admin-rest-setting-options-controller.php
joshuatf: This is a tricky one.  The reason for this and the controller override is the way that WC Settings handles `multiselect` settings.  It expects an array of values and the only way to provide this is using the same query params with brackets to pass the value:

`value[]=option1&value[]=option2`

Passing the above string through `apiFetch` doesn't work since the brackets get escaped and unfortunately we also can't pass them via the data object since we can't provide the same value twice:

```js
const data = {
    value: 'option1',
    value: 'option2',
}
```

I'm open to other solutions and maybe I'm missing something obvious that handles this better.

client/wc-api/settings/operations.js
joshuatf: (Answer below)

includes/class-wc-admin-api-init.php
joshuatf: (Answer above)

joshuatf:
@jeffstieler @LevinMedia Thanks for the reviews!  Will get that initial state error fixed up.

@LevinMedia Any other design changes you want on this page for the time being or are you okay with it as is for now?

@willgorham Thanks for the input!  Included statuses does seem a bit more intuitive, but part of the reason that statuses are excluded is so that if a new custom status is added (which is more likely to be a "positive" or "successful" status), that these are already included in the reports.

For example, if I add a new custom status of `shipped`, I would want this in my reports.  Using excluded statuses means this custom status is already included without the need to make changes to settings.

willgorham:
@joshuatf Makes sense, I figured that was probably the reasoning. Thanks for the response!

includes/api/class-wc-admin-rest-setting-options-controller.php
justinshreve: These overrides will eventually become the new core endpoints/core functionality -- so should we instead use this opportunity to change the handling of multiselect settings to look for `value0`, `value1` instead of comma handling?

lib/client-assets.php
```diff
@@ -197,10 +197,13 @@ function wc_admin_print_script_settings() {
 		),
 		'currentUserData'       => $current_user_data,
 	);
+	$settings = wc_admin_add_custom_settings( $settings );
 
 	foreach ( $preload_data_endpoints as $key => $endpoint ) {
 		$settings['dataEndpoints'][ $key ] = $preload_data[ $endpoint ]['body'];
 	}
+
+	$settings = apply_filters( 'wc_admin_wc_settings', $settings );
```
justinshreve: 👍

client/analytics/settings/setting.js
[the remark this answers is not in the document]
justinshreve: I'm fine with the addition of groups here too.

emdashcodes (commented):
A few more minor comments, but good job on this 👍.

includes/api/class-wc-admin-rest-setting-options-controller.php
jeffstieler: For some reason I couldn't "suggest a change" for this PR, so here's a branch showing everything working without juggling the array around as a string: https://github.com/woocommerce/wc-admin/pull/1389

LevinMedia:
@joshuatf I'm good with where this is at for now - just for reference here's the issue that was created back in November to address transient notifications. I think we should go that route for this page. 

https://github.com/woocommerce/wc-admin/issues/806

client/analytics/settings/setting.js
```diff
@@ -0,0 +1,141 @@
...
+	 * Array of options used for when the `inputType` allows multiple selections.
+	 */
+	options: PropTypes.arrayOf(
+		PropTypes.shape( {
+			/**
+			 * Input value for this option.
+			 */
+			value: PropTypes.string,
+			/**
+			 * Label for this option or above a group for a group `inpuType`.
```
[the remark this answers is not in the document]
joshuatf: Thanks for catching!  😄 Fixed.

client/analytics/settings/setting.js
joshuatf: Updated to use `interpolateComponents()`.

client/analytics/settings/index.js
joshuatf: On second thought, I think we should stick with `remove()` here since `without()` requires us to reassign the constant and makes things a bit messier here.

joshuatf:
@LevinMedia Awesome!  Will get an issue made to address the notifications here.

@jeffstieler Thanks for the review!  Made those changes and this is ready for another review.

>In addition to the console error on initial interaction, this seems to break using the "order status" advanced filter when the status is one of the excluded ones. cc: @LevinMedia for UX on that

Did you encounter this breaking somewhere or were you expecting not to see excluded statuses in the filter dropdown?  Per discussion in p90Yrv-Vb-p2 we're keeping all order statuses in the orders filter so that users can continue to see stats on these.  The excluded statuses only affect "default" reporting where these statuses should not alter the stats.

client/analytics/settings/setting.js
```diff
@@ -0,0 +1,135 @@
...
+	renderInput = () => {
+		const { handleChange, name, inputType, options, value } = this.props;
+		const id = uniqueId( name );
+
+		switch ( inputType ) {
+			case 'checkboxGroup':
+				return options.map(
+					optionGroup =>
+						optionGroup.options.length > 0 && (
+							<div className="woocommerce-setting__options-group" key={ optionGroup.key }>
```
timmyc: Not 100% certain but thinking we might need `aria-labelledby` here [#](https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20160317/examples/radio/radio.html)

client/wc-api/settings/operations.js
```diff
@@ -0,0 +1,71 @@
...
+	return [ ...readSettings( resourceNames, fetch ) ];
+}
+
+function update( resourceNames, data, fetch = apiFetch ) {
+	return [ ...updateSettings( resourceNames, data, fetch ) ];
+}
+
+function readSettings( resourceNames, fetch ) {
+	if ( resourceNames.includes( 'settings' ) ) {
+		const url = '/wc/v4/settings/wc_admin';
```
timmyc: should we use `NAMESPACE` here?

timmyc (approved):
Testing out great for me... just a couple of minor comments in the code. Pre-approving pending comments from anybody else ( Jeff is AFK mostly this week )

client/analytics/settings/index.js
```diff
@@ -0,0 +1,134 @@
...
+			}
+		} else {
+			settings[ name ] = value;
```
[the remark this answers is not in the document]
joshuatf: This is a terrific idea.  Hopefully with our next setting we'll have a good use case to test this.

client/analytics/settings/setting.js
joshuatf: Added in https://github.com/woocommerce/wc-admin/pull/1364/commits/9a97ebfeab3474a6292b8c4fde12c359e4bd3b58 and tested with a screen reader.  I tied it to the setting label since the group label ("Custom Statuses" in the case of the 2nd group) may or may not be present here.

joshuatf:
Thanks for the review, @timmyc!  Made those changes and gave this a rebase.

I think all other comments and requested changes have been made (one question remains to @jeffstieler above), so I'm going to merge this in so we can test transient notices.  If any other issues pop up we can handle in follow-up PRs.

PR #1365, opened 2019-01-21
 Fill full plot area in bar charts 

### LINKED ISSUE #1269: Chart: Issues with Bar Series not Filling Full Plot Area
Via #1100

The width of the charted area changes when toggling between line and bar chart. 

screen recording 2018-12-14 at 10 39 pm

Also for both line and bar charts in particular on smaller viewport, the maximum plotable area is not used. 

screen-shot-2018-12-14-at-10 41 16-pm

Aljullu:
Fixes #1269.

- Change bar chart scale to use [`range`](https://github.com/d3/d3-scale#band_range) instead of [`rangeRound`](https://github.com/d3/d3-scale#band_rangeRound), so it better adapts to the available width. It has the drawback that bars might look a bit blurry because they might take decimal widths, but I tested in a non-HiDPI screen and it looks good-enough.
- Refactor scale tests, so they only test the params we pass to d3 functions, instead of values returned by those functions.

### Screenshots
_Before:_
image

_After:_
image

- Go to the _Revenue_ report, for example.
- Select _Last Year_ as the date range.
- Switch to bar chart.
- Verify it fills all the available space (similar to the line chart).

### DIFF
--- a/packages/components/src/chart/d3chart/utils/scales.js
+++ b/packages/components/src/chart/d3chart/utils/scales.js
@@ -33,7 +33,7 @@ export const getYMax = lineData => {
 export const getXScale = ( uniqueDates, width, compact = false ) =>
 	d3ScaleBand()
 		.domain( uniqueDates )
-		.rangeRound( [ 0, width ] )
+		.range( [ 0, width ] )
 		.paddingInner( compact ? 0 : 0.1 );
 
 /**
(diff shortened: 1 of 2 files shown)


timmyc (approved):
Thanks for all the updates to the tests and supporting logic here. The fix is testing well for me 👍

PR #1369, opened 2019-01-22
Increase default `wc-api` timeout.

### LINKED ISSUE #1304: Analytics: Customers Report Endless Loop
When viewing the customers report, and accessing a new page of data - there appears to be an endless loop that continually requests the same API resource every few seconds.

**To Reproduce**

1. Your database must have at least 75 customers to have a paged data set. I'm using a woocommerce.com backup that has over 6k customers
2. Open your network console, and load the customers report. Click on the pagination link to load up another page of data

This is a very boring .gif of the loop in action:

My endless loop file is too large :) - I will reproduce and post it later today.

jeffstieler:
Fixes #1304.

Large datasets may cause responses to take longer than 5 seconds, which causes fresh-data to continually issue repeat requests until the first finally responds.

This PR increases the default timeout value to 30 seconds.

- On `master` branch
- Open a report that will query a sufficiently large dataset and/or throttle network requests using your browser's development tools (the request must take longer than 5 seconds)
- Navigate to another table page
- Verify that duplicate requests are made every 5 seconds while the first hasn't responded
- Check out this branch, refresh report page
- Navigate to another table page
- Verify that only one request is made (provided it returns in less than 30 seconds)

### DIFF
--- a/client/wc-api/constants.js
+++ b/client/wc-api/constants.js
@@ -2,12 +2,12 @@
 /**
  * External dependencies
  */
-import { SECOND, MINUTE } from '@fresh-data/framework';
+import { MINUTE } from '@fresh-data/framework';
 
 export const NAMESPACE = '/wc/v4';
 
 export const DEFAULT_REQUIREMENT = {
-	timeout: 5 * SECOND,
+	timeout: 1 * MINUTE,
 	freshness: 5 * MINUTE,
 };
 


timmyc:
While I think 30 seconds _should_ be more than enough, check out the network requests when loading the customers page in firefox:

image

7 seconds for the request to the `/me` endpoint - which I suppose we could look at eliminating now that we are pre-hydrating the state with the same data, but checkout that request to the customers endpoint. **41+ seconds**

So obviously it appears there are some performance issues in the endpoint itself - I'd like to think that running VVV locally is _somewhat_ like some shoddy shared hosting that is out in the wild... I do also fear that some hosts might close requests that take that long to respond.

Anyhow coming back to the fix here at hand. Should we bump this to a minute for now, just to ensure it isn't hit, and then do a follow-up issue specifically around the customers endpoint to try and see why it is taking so long?

FWIW, on the second load of the same page, it appears some caching has kicked in and the response is much quicker:

image

jeffstieler:
>  Should we bump this to a minute for now, just to ensure it isn't hit, and then do a follow-up issue specifically around the customers endpoint to try and see why it is taking so long?

👍

jeffstieler:
Upped the timeout to 1 minute in 6594518.

timmyc:
Word. So I do think we still need to dig into the `fresh-data` internals here a bit @coderkevin to think about how best to handle this sort of scenario. Slow REST responses I fear will be a reality in the WordPress wild - should I open up an issue specifically around that here or would the fresh-data repo be better?

timmyc (approved):
This change fixes the originally reported issue for me.

PR #1371, opened 2019-01-22
First attempt to fix tests failing around end of hour

### LINKED ISSUE #1316: Tests fail when started at the end of the hour
**Describe the bug**
When starting tests at the end of the hour (e.g. 10:59), they sometimes seem to fail.  I suspect that this has to do with the `before` and `after` query parameters used for the data store in 

**To Reproduce**

1. Start unit tests at the end of the hour (or replicate this by modifying system time).
2. Unit tests should complete after the hour is over.
3. Note the failed tests.

**Expected behavior**
Expected tests to not fail.

**Additional context**
Occurred in this issue - https://github.com/woocommerce/wc-admin/pull/1180

peterfabian:
Fixes #1316 

- Noticed incorrect interval arithmetic for hours and days, so fixed those, including tests. 
- Added test case to cover the problematic example.
- Tried to update test so that start/end times are always predictable and orders end up inside correct time interval

- Run tests
- Extensively test the API/run tests around end of hour.

### DIFF
--- a/includes/class-wc-admin-reports-interval.php
+++ b/includes/class-wc-admin-reports-interval.php
@@ -170,11 +170,29 @@ class WC_Admin_Reports_Interval {
 
 		switch ( $interval ) {
 			case 'hour':
-				$diff_timestamp = (int) $end_datetime->format( 'U' ) - (int) $start_datetime->format( 'U' );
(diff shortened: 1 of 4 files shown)


peterfabian:
Thanks @jeffstieler !

PR #1377, opened 2019-01-23
Allow customer date_last_active to be null

### LINKED ISSUE #1368: Customers Report: date_last_active sometimes malformed
**Describe the bug**
`date_last_active` property for some customers in the `reports/customers` endpoint is malformed:

screen shot 2019-01-22 at 3 57 03 pm

screen shot 2019-01-22 at 3 54 43 pm

**To Reproduce**

1. Customers Report
2. See oddly formed data in the response

*Additional context**

These orders and customers were created with WC-Smooth-Generator.

joshuatf:
Fixes #1368 

Allows `date_last_active` to be null since sometimes customers/users are created without having made orders.

We set `date_last_active` to null since we don't know if the user has actually been active since an admin could have created and updated this user.

### Screenshots
screen shot 2019-01-23 at 3 48 56 pm

1.  Deactivate/reactivate the plugin and make sure the `date_last_active` column in `wc_customer_lookup` has been made nullable.  You may need to delete this table and rebuild data for this to work.
2.  Create a user through the admin panel.
3.  Update that user.
4.  Note that the user is in the customers table and `date_last_active` is null.
5.  Note that no date is shown and "Invalid date" is not shown in the customers report for this user `wp-admin/admin.php?page=wc-admin#/analytics/customers`.

### DIFF
--- a/client/analytics/report/customers/table.js
+++ b/client/analytics/report/customers/table.js
@@ -148,7 +148,9 @@ export default class CustomersReportTable extends Component {
 					value: username,
 				},
 				{
-					display: <Date date={ date_last_active } visibleFormat={ defaultTableDateFormat } />,
+					display: date_last_active && (
+						<Date date={ date_last_active } visibleFormat={ defaultTableDateFormat } />
+					),
 					value: date_last_active,
 				},
 				{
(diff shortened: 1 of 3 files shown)


Aljullu (approved):
Tested and it works as expected. It will need to be rebased, though.

timmyc:
I gave this a quick rebase, and will merge it in so we can make sure it gets in the release next week.

PR #1378, opened 2019-01-23
Use tableQuery sorting when url query is not present

### LINKED ISSUE #1363: Sort order in table is different than initial query
**Describe the bug**
When passing a `tableQuery` prop to a report table, it seems the sort order here is not tied to the `query` prop and sorting (chevron icons and click actions) are initially incorrect.

For example, passing this:
```js
tableQuery={ {
    order: query.order || 'asc',
} }
```

Still results in a "chevron-down" icon and the initial sorting click on this column does nothing.

**To Reproduce**

1. Go to the Downloads report `/wp-admin/admin.php?page=wc-admin#/analytics/downloads` and select a range with multiple rows of data.
2. Note the "Date" column has a chevron-down icon (even though the default already order is "ascending").
3. Click on the "Date" column to sort it.
4.  Note that the column sorting does not change on the first click.

**Expected behavior**
Expected initial values passed to the `tableQuery` to be used for the initial sorting in the table.

**Screenshots**
jan-21-2019 11-24-04

joshuatf:
Fixes #1363

Applies default `tableQuery` to the sorting table when url params aren't set so that initial sorting (arrow) is properly set.

### Before
screen shot 2019-01-23 at 5 22 16 pm

### After
screen shot 2019-01-23 at 5 20 45 pm

1.  Visit the Stock report `wp-admin/admin.php?page=wc-admin#/analytics/stock`.
2.  Check that the arrow is a chevron-up (ascending).
3.  Check that other reports show the correct icon.

### DIFF
--- a/client/analytics/components/report-table/index.js
+++ b/client/analytics/components/report-table/index.js
@@ -193,6 +193,7 @@ export default compose(
 		const selectProps = {
 			primaryData,
 			tableData: extendedTableData,
+			query: { ...tableQuery, ...query },
 		};
 
 		if ( columnPrefsKey ) {


jeffstieler (approved):
Nice and simple change, works for me. 🚢 it.

timmyc:
@LevinMedia wondering if this might fix the sorting issue you were looking at earlier

Issue #1102, opened 2018-12-15
Analytics: Tabular Data Filtering


LevinMedia:
_Note I repurposed this "Sweep" issue since it only had one item on it pertaining to filtering tabular data, I've re-worded David's original report a bit too - Timmy_

Currently when interacting with the tabular data text filter box ( i.e. on the Products report searching for a product name ), selecting a product from the list applies a report-wide compare filter via adding in the following query args to the url: `?filter=compare-products&products=581`.

Note the design spec at p6riRB-3lZ-p2 - specifically the section "Filtering" (The second section in overview") The search field should allow for string searches that match product titles, and potentially  product SKUs (Not included in the original design, but would be a nice to have feature)

![screen recording 2018-07-14 at 09 33 am](https://user-images.githubusercontent.com/4500952/50047431-aae1c680-0069-11e9-88be-de49166171e5.gif)

[Prototype](https://automattic.invisionapp.com/share/ENPN7Y3ZF5Y#/screens/308750685)

The desired outcome of interacting with these search fields in the tabular data is to apply filter to the report for that single item selected - not enter "compare" mode in the reports. As such we might need to introduce a new query arg to track this table filter as it is stand-alone from outside of the context of the advanced filters. Like other filters, it would be useful to deep link via a URL with the proper filter arg to a pre-filtered version of a report too.

The functionality exists on the following reports:
* Products
* Categories
* Coupons
* Taxes
* Customers

Perhaps in this first issue we can focus on the Products report, but be aware a similar solution is needed across all the other reports as well.

LevinMedia:
**Bug**

Note the design spec at p6riRB-3lZ-p2 - specifically the section "Filtering" (The second section in overview") The search field should allow for string searches that match product titles, and potentially  product SKUs (Not included in the original design, but would be a nice to have feature) This is different than the current behavior. 

![screen recording 2018-07-14 at 09 33 am](https://user-images.githubusercontent.com/4500952/50047431-aae1c680-0069-11e9-88be-de49166171e5.gif)

[Prototype](https://automattic.invisionapp.com/share/ENPN7Y3ZF5Y#/screens/308750685)

---

Placeholder for additional items

LevinMedia:
Noting that when filtering in this manner, the graph should redraw and summary numbers should update to reflect the filtered rows.

PR #1379, opened 2019-01-23
Move 'getLabels' functions into lib/async-requests

Aljullu:
This PR moves the logic to load filter labels into `lib/async-requests`, the goal is to DRY and to make them reusable outside of the `config.js` files. I needed that while working on #1102.

### Screenshots
image

This PR shouldn't have produced any visible change for users. A good way to test is testing that all filter autocompleters keep working and there are no regressions.


client/analytics/report/downloads/config.js
```diff
@@ -126,10 +119,13 @@ export const advancedFilters = {
...
-					id: order.id,
-					label: '#' + order.id,
-				} ) ),
+				getLabels: async value => {
+					const orderIds = value.split( ',' );
+					return await orderIds.map( orderId => ( {
+						id: orderId,
+						label: '#' + orderId,
+					} ) );
+				},
```
Aljullu: This change is not related to the others, but I saw we could avoid an API request here.

client/lib/async-requests/index.js
```diff
@@ -32,3 +37,40 @@ export function getRequestByIdString( path, handleData = identity ) {
...
+		label: category.name,
+	} )
+);
+
+export const getCouponLabels = getRequestByIdString( NAMESPACE + 'coupons', coupon => ( {
+	id: coupon.id,
+	label: coupon.code,
+} ) );
+
+export const getCustomerLabel = getRequestByIdString( NAMESPACE + 'customers', customer => ( {
```
jeffstieler: Nitpick: label vs. labels.

jeffstieler (approved):
Code looks good and tests well - one minor naming nitpick. Either way - :shipit:

PR #1380, opened 2019-01-23
Add regex expression option to autocompleters

### LINKED ISSUE #1348: Country autocomplete search should not be fuzzy
**Describe the bug**
Country codes were added as keywords to country searches in https://github.com/woocommerce/wc-admin/pull/1337, however some countries will not be shown by their code as enough other matches (>10) are shown alphabetically first.

As @Aljullu suggested, we could restrict the results to partial matches that start with the query string instead of matching any letter in the results.  Full context here - https://github.com/woocommerce/wc-admin/pull/1337#pullrequestreview-193517460

Another option I mentioned is to sort results by best match instead of alphabetically.  However, this may take more time to implement so I'm a little more partial to @Aljullu's suggestion.

**To Reproduce**

1. Go to the customers report and use the advanced filter.
2. Select "Country" as a filter.
3. Type a country code or name and note that matches are fuzzy and country codes do not necessarily come up if the result list is too long.

**Expected behavior**
Expected to see a country show up by its country code.

**Screenshots**
Spain (country code ES) is a good example of this:

screen shot 2019-01-18 at 11 07 45 am

joshuatf:
Fixes #1348

Adds a regex search option to autocompleters so we can change how options are searched.

### Screenshots
screen shot 2019-01-23 at 6 15 33 pm

1.  Visit the Customers report.
2.  Filter by country.
3.  Make sure the search returns results that match country codes or searches that match the start of the country's name (not fuzzy matching).
4.  Check that other search autocompleters continue to work as expected.

### DIFF
--- a/packages/components/src/search/autocompleters/countries.js
+++ b/packages/components/src/search/autocompleters/countries.js
@@ -23,6 +23,9 @@ export default {
 	options() {
 		return wcSettings.dataEndpoints.countries || [];
 	},
+	getSearchExpression( query ) {
+		return '^' + query;
+	},
 	getOptionKeywords( country ) {
 		return [ country.code, decodeEntities( country.name ) ];
 	},
--- a/packages/components/src/search/autocomplete.js
+++ b/packages/components/src/search/autocomplete.js
@@ -193,7 +193,10 @@ export class Autocomplete extends Component {
 			}
 		}
 		// create a regular expression to filter the options
-		const search = new RegExp( escapeRegExp( query ), 'i' );
+		const expression = 'undefined' !== typeof completer.getSearchExpression
+			? completer.getSearchExpression( escapeRegExp( query ) )
+			: escapeRegExp( query );
+		const search = new RegExp( expression, 'i' );
 		// filter the options we already have
 		const filteredOptions = filterOptions( search, this.state.options, selected );
 		// update the state


packages/components/src/search/autocompleters/countries.js
```diff
@@ -23,6 +23,9 @@ export default {
 	options() {
 		return wcSettings.dataEndpoints.countries || [];
 	},
+	getSearchExpression( query ) {
+		return '^' + query;
```
Aljullu: My initial impression was that we should match not only the first word but any word in a country name.
```suggestion
		return '^.*\\b' + query;
```
So searching for `Salvador` or `Korea` would return results, which will not appear if we only test the first word of the country name.

The drawback is that other queries would return irrelevant results (for example, when searching `th`):
image

So, thinking it twice, I think your solution is better.

Aljullu (approved):
That's a great improvement for the country autocompleter. Code looks good and is testing well.

packages/components/src/search/autocompleters/countries.js
joshuatf: Thanks for the review, @Aljullu!  

>My initial impression was that we should match not only the first word but any word in a country name.

🤔 I like that solution too.  I'm a bit torn here as to which would be a better option.  I'll go ahead and get this merged for now and we can adjust later if we need to.

PR #1381, opened 2019-01-23
Implement tabular data filtering

### LINKED ISSUE #1102: Analytics: Tabular Data Filtering

Aljullu:
Fixes #1102.

This PR implements tabular data filtering for report tables. Notice there are several pending tasks/open issues detailed at the end of this post. I will open issues/create PR for them as follow-ups of this one.

### Screenshots

search

- Enter some product names in the table search bar and verify filtering works as expected.
- Try deleting them and adding new ones.
- Try adding _Single Product_, _Comparison_, etc. filters and verify table searches are reset.
- While having _Single Product_, _Comparison_, etc. filters on, start searching with the table search bar and verify filters are reset.
- Trying searching partial strings, for example `sun` and verify `Sunglasses` appear.
- Try adding search strings directly to the URL after the `search=` parameter, separated with a comma, and verify the table is updated with the correct results.
- Try refreshing the page and verify searches are preserved.
- Repeat the process above with all other reports that have search bars in the table: Categories, Coupons, Taxes and Customers. Notice some are not working yet (see section below).

### What is not working? & Pending tasks
In order to avoid this PR growing too much and to get some early feedback before continuing, I will leave some work for follow-ups:
- The _Clear all_ from designs is implemented in: #1410.
- `variations` autocompleter is not working (if I understood it correctly, it will be fixed with #1391). That makes searching in the variations table not to work either.
- The Rest API call used to populate the autocompleter is not reused later to get the IDs to populate the table, causing two different API requests to be called when we really only need one (#1453).
- The app was crashing when a search was returning 0 results. For now, I made the app to ignore the search in those cases and added a `@TODO` [comment](https://github.com/woocommerce/wc-admin/pull/1381/files#diff-2459417b12e8fe2d482313c6d8f9819aR151) (#1451).
- Current `products` wc endpoint doesn't allow searching by `sku`.
- Current `customer` wc endpoint doesn't allow searching by `name + ' ' + last_name` (#1455).
- Current `customer`  wc-admin endpoint doesn't allow limiting the results by customer ID (#1457).
- Current `taxes`  wc-admin endpoint doesn't allow limiting the results by tax ID (#1456).
- I created a generic `getItems( type )` method and migrated `getCategories()` to use it. It would make sense to migrate `getOrders()` to `getItems( 'orders' )` as well in order to DRY (#1438).
- Handle the case when there is more than one page of results (#1454).


LevinMedia:
Heya @Aljullu thanks for tackling this one! Just wanted to note, the search results should also allow for a string - so in this case if I typed woo, then hit enter, it would filter to all of my products that contained the string "Woo" 

The functionality is noted in #1102 

>The search field should allow for string searches that match product titles, and potentially product SKUs (Not included in the original design, but would be a nice to have feature). 

I do like the ability to match a product directly as it's set up now... perhaps we could integrate the string search for title and SKIU as a result at the top of the drop down so users have the option for both? Something along these lines:

screen-shot-2019-01-23-at-10 29 10-am

Aljullu:
Thanks for pointing this out, @LevinMedia. Today I have been playing with this idea a bit, I updated a WIP here but it's still far from ready.

customers

Currently, the only report that allows filtering by string is the customers report. I updated it to allow partial strings, so searching for `adr` would return `Adrienne O'Conner`, for example (see above). I'm not sure about the performance implications of this, however.

Implementing this in the other reports will require updating several API endpoints to allow this kind of filtering, so I would like to get some input from somebody with more experience on this.

**tl;dr**

We want to allow filtering products, categories, customers, coupons, taxes, etc. by string (see #1102). For example, searching for `Tee` and `Fannypack` should return the products `Isotope Tee`, `Proton Tee`, etc.

@timmyc @peterfabian do you have any thoughts on the best way to approach this? I can see this being a bit challenging in some reports, for example in _Products_ where we don't store the product names/skus in the same table used to populate the report.

SwaggerHub will also need to be updated with information about the parameters to use for filtering.

timmyc:
This is looking fantastic Albert 🌟 

Just wanting to make sure I understand the issue at hand now - so just focusing on the products report use case first. When using the new "All products with a title that includes %string%" - I'm seeing the following requests fired:

`http://woo.test/wp-json/wc/v4/reports/products?after=2019-01-01T00%3A00%3A00%2B00%3A00&before=2019-01-25T23%3A59%3A59%2B00%3A00&extended_info=true&order=desc&orderby=items_sold&page=1&per_page=25&products=secre&_locale=user`

and

`http://woo.test/wp-json/wc/v4/reports/products/stats?after=2018-01-01T00%3A00%3A00%2B00%3A00&before=2018-01-25T23%3A59%3A59%2B00%3A00&interval=day&order=asc&per_page=100&products=secre&_locale=user`

Which both are erroring due to a string being passed in to the `products` param, while it [expects](https://github.com/woocommerce/wc-admin/blob/master/includes/api/class-wc-admin-rest-reports-products-controller.php#L317-L324) an array of product id(s). I guess one solution that we could explore here is to grab all the product ID's returned from the initial search against the products endpoint ( that we use to display the autocompleter results ), and pass those in as an array in the `products` arg.

I guess we could run into issues here if a site has a large product catalog, and the initial string search spans more than one page of matching product results - and we only have the matching product IDs from the first batch - but perhaps we could change the "All products with a title that includes %string%" text if this happens to something like "Over x Products have a title that includes %string% - view first 100 matching results". But still kind of think that is an edge case.

That glitch aside, do you think that solution could work since we already have the needed object IDs?
I _think_ we could probably use the same approach on taxes and coupons since they follow a similar flow - query another endpoint to get autocompleter results - which return the requisite object ids.

Aljullu:
Thanks @timmyc! That's a good idea, fetching the IDs first and then populating the report with those IDs. That would work in all reports where we have an autocompleter (all the ones with a searchable-table, if I'm not wrong).

The only drawback I can see is that in some circumstances, that would require some extra API calls.

Currently, autocompleter API calls limit the number of results to 10, so in the case there were more than 10 matches, we would need an extra call to fetch the other IDs.

Something similar would happen if the search terms are specified in a parameter in the URL before the user interacts with the page, for example, somebody refreshing the page or directly modifying the URL. A query like: `...?search_param=cap,belt,hoodie,sunglasses,shoe` would mean we have to perform 5 API calls to get the IDs for each term + 1 call to fill the report with those IDs.

I assume those are edge cases, so I will start working in this direction.

LevinMedia:
>Currently, autocompleter API calls limit the number of results to 10, so in the case there were more than 10 matches, we would need an extra call to fetch the other IDs.

@Aljullu  I'm not totally confident, but this sounds like a similar problem to what we were experiencing in the Legacy product block. I know @ryelle did some ✨ to make it return the complete list... maybe this is something we could leverage here too?

timmyc:
This is kind of starting to sound a wee bit like the discussion around limiting the number of data points to compare in a chart. I personally feel the top 10 matches is sufficient myself, but I suppose we could bump that up to 20. But my vote is to kind of stick with it as-is, and see if it is even a problem brought up by users down the road.

Just my .o2

Aljullu:
Thanks @LevinMedia and @timmyc for the feedback! I uploaded a working version with @timmyc's approach that should be ready for review.

client/analytics/report/products/index.js
```diff
@@ -24,7 +24,7 @@ import VariationsReportTable from './table-variations';
 export default class ProductsReport extends Component {
 	render() {
 		const { path, query } = this.props;
-		const isProductDetailsView = query.products && 1 === query.products.split( ',' ).length;
+		const isProductDetailsView = query.filter === 'single_product';
```
timmyc: Nice clean up here.

client/wc-api/items/operations.js
```diff
@@ -0,0 +1,65 @@
...
+	return filteredNames.map( async resourceName => {
+		const prefix = getResourcePrefix( resourceName );
+		const endpoint = typeEndpointMap[ prefix ];
+		const query = getResourceIdentifier( resourceName );
+		const url = NAMESPACE + `/${ endpoint }${ stringifyQuery( query ) }`;
+
+		try {
+			const items = await fetch( {
+				path: url,
+			} );
```
timmyc: Kind of the edge case we were talking about before where a query might return more than one page of results, should we somehow expose `X-WP-Total` or `X-WP-TotalPages` so we could display messaging that more results exist?

packages/components/src/search/autocompleters/categories.js
```diff
@@ -33,12 +35,31 @@ export default {
 			};
 			payload = stringifyQuery( query );
 		}
-		return apiFetch( { path: `/wc/v3/products/categories${ payload }` } );
+		return apiFetch( { path: `/wc/v4/products/categories${ payload }` } );
```
timmyc: good catch 👍

packages/components/src/table/index.js
```diff
@@ -236,13 +237,14 @@ class TableCard extends Component {
...
 			rowsPerPage,
 			searchBy,
-			searchParam,
 			showMenu,
 			summary,
 			title,
 			totalRows,
 		} = this.props;
 		const { selectedRows, showCols } = this.state;
+		const searchedValues = query.search ? query.search.split( ',' ) : [];
```
timmyc: do we want to trim away white space or check length of `query.search` here too?

packages/components/src/table/index.js
```diff
@@ -277,13 +279,15 @@ class TableCard extends Component {
...
+					searchBy && (
 						<Search
+							allowFreeTextSearch={ true }
+							inlineTags
 							key="search"
-							placeholder={ labels.placeholder || __( 'Search by item name', 'wc-admin' ) }
-							type={ compareBy || searchBy }
 							onChange={ this.onSearch }
-							selected={ searchParam && getIdsFromQuery( query[ searchParam ] ).map( id => ( { id } ) ) }
+							placeholder={ labels.placeholder || __( 'Search by item name', 'wc-admin' ) }
```
timmyc: Like the default here - nice.

client/wc-api/items/operations.js
timmyc: Totally fine with this being follow-up btw, just thinking out loud.

timmyc (commented):
Wow this is fantastic @Aljullu! Couple of comments in the code, and I did hit a few things while manually testing. I'm okay with however you would like to proceed, fix some in this PR or do in follow-ups as this PR is pretty full featured already.

__Debounce Search Requests__
Thinking we might want to debounce the API calls when typing in the search field, if i'm a slower typer you can see a request being fired on each keyup:

search-requests

__Trim Query / Don't search for empty strings__
If I just enter spaces into a search field, a search request to the API is still sent - thinking we should trim that value and only do a search if there is a length:

search-with-spaces

__Exception on Taxes Report__
When clicking 'show all that include NAME" on the taxes report, an exception happens:

taxes-exception

taxes-report

LevinMedia:
@Aljullu I just tested the branch again, and it blew my mind. In a good way. So cool. Thank you! 🎉

packages/components/src/table/index.js
Aljullu: That's a good idea! I think trimming the input of the `<Search>` box would the best solution here, so it will also affect the filters search boxes. Created an issue for that: #1437.

client/wc-api/items/operations.js
Aljullu: Right, actually I think it shouldn't be that difficult making several consecutive queries when there is more than one page of results, similarly as what we do for the chart.

I think that would be a better solution than adding a message in the UI, what do you think?

For now, I added a note at the top comment of this PR and will create a new issue for it when we merge this one.

Aljullu:
> Debounce Search Requests
> Trim Query / Don't search for empty strings

Both sound good to me! We should implement that in the `Autocomplete`/`Search` components, so it will affect filters autocompleter too. I created a couple of issues: #1436 and #1437.

> Exception on Taxes Report

Good catch! Should be fixed in: a3636005e5fb0cdb969d6a12d1c5a97d1bd59435.

@timmyc feel free to take another look! I will be working on the follow-ups.

timmyc (approved):
Thanks for opening up those follow-up issues @Aljullu - I confirmed the exception in the taxes report is all fixed up. Lets land this because it is fantastic!

client/wc-api/items/operations.js
timmyc: I still think we just have a hard limit of the number of results we fetch and be okay with it, but perhaps add some messaging that "x more results match"

PR #1382, opened 2019-01-23
Customer report: Update lifetime spend label and re-order columns + default sorting

LevinMedia:
This PR updates the "Lifetime Spend" column label, to "Total Spend" so that it matches the label in the advanced filter. 

It also re-orders the table columns, making "Last Active" the second column, and the default sorting column as well. 

**Before:**

screen shot 2019-01-23 at 10 09 58 am

**After:**

screen shot 2019-01-23 at 10 09 22 am


jeffstieler (approved):
Looks good! :shipit:

PR #1384, opened 2019-01-23
Flag: Remove React package in favor of emojis

psealock:
Fixes https://github.com/woocommerce/wc-admin/issues/1275

`react-world-flags` was too large of a dependency to ship with wc-admin. This PR explores using emojis from the user's system to render flags.

I removed the `round` attribute because it doesn't really work well with curved flags, [which many flag emojis have](https://emojipedia.org/flag-for-vanuatu/).

### Accessibility

I enable an `aria-hidden` attribute on flags in the countries autocompleter, otherwise screen readers will read "flag Australia, Australia" for each option, which is redundant.

### Before

**App Size**: 11.41 MB

screen shot 2019-01-24 at 9 29 58 am

screen shot 2019-01-24 at 9 30 24 am

### After

**App Size**: 7.19 MB

screen shot 2019-01-24 at 11 23 44 am

screen shot 2019-01-24 at 11 24 06 am

### Removing wp-emoji.js

If a browser/OS doesn't support emojis, Core replaces them with an image using Twitter's Twemoji. It turns out, this functionality's check to see if emojis are supported always returns false for modern browsers. 

Gutenberg has disabled this feature in https://github.com/WordPress/gutenberg/pull/6151 for separate reasons, but I feel ok doing the same until Core can update to allow modern browsers to keep emojis.

### Test instructions:

1. Customers Report > Show > Advanced Filters > Country
2. Search for country and see flags being rendered correctly.
3. Activity Panel > Orders.
4. See orders being rendered with correct flags.

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,6 +131,7 @@
     "d3-shape": "^1.2.2",
     "d3-time-format": "^2.1.3",
     "dompurify": "1.0.9",
+    "emoji-flags": "^1.2.0",
     "gfm-code-blocks": "1.0.0",
     "gridicons": "3.1.1",
     "history": "4.7.2",
@@ -147,7 +148,6 @@
     "react-router-dom": "4.3.1",
     "react-slot-fill": "^2.0.1",
     "react-transition-group": "^2.4.0",
-    "react-world-flags": "1.2.4",
     "redux": "^4.0.0"
   },
   "husky": {
--- a/packages/components/src/flag/index.js
+++ b/packages/components/src/flag/index.js
@@ -3,40 +3,36 @@
 /**
  * External dependencies
  */
-import ReactFlag from 'react-world-flags';
 import classnames from 'classnames';
 import PropTypes from 'prop-types';
+import emojiFlags from 'emoji-flags';
+import { get } from 'lodash';
 
 /**
- * Use the `Flag` component to display a country's flag.
+ * Use the `Flag` component to display a country's flag using the operating system's emojis.
  *
- * @return { object } -
+ * @return { object } - React component.
  */
-const Flag = ( { code, order, round, height, width, className } ) => {
-	const classes = classnames( 'woocommerce-flag', className, {
-		'is-round': round,
-	} );
+const Flag = ( { code, order, className, size, hideFromScreenReader } ) => {
+	const classes = classnames( 'woocommerce-flag', className );
 
 	let _code = code || 'unknown';
-	if ( order && order.shipping ) {
+	if ( order && order.shipping && order.shipping.country ) {
 		_code = order.shipping.country;
 	} else if ( order && order.billing && order.billing.country ) {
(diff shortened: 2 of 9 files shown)


jeffstieler:
I got the following error trying to open the Orders panel:

```
index.js:9 Uncaught Error: Expected 1 country code as the first argument
    at Object.countryCode (index.js:9)
    at Flag (index.js:30)
```

psealock:
Thanks for taking a look @jeffstieler, that error should be fixed in https://github.com/woocommerce/wc-admin/pull/1384/commits/3e1483ac427493cd6e9a7b5b2fda001a6f850707

package.json
```diff
@@ -131,6 +131,7 @@
     "d3-shape": "^1.2.2",
     "d3-time-format": "^2.1.3",
     "dompurify": "1.0.9",
+    "emoji-flags": "^1.2.0",
```
timmyc: Shouldn't we remove `"react-world-flags": "1.2.4",` too?

packages/components/src/flag/index.js
```diff
@@ -3,40 +3,36 @@
...
-const Flag = ( { code, order, round, height, width, className } ) => {
-	const classes = classnames( 'woocommerce-flag', className, {
-		'is-round': round,
-	} );
+const Flag = ( { code, order, className, size, hideFromScreenReader } ) => {
+	const classes = classnames( 'woocommerce-flag', className );
 
 	let _code = code || 'unknown';
-	if ( order && order.shipping ) {
+	if ( order && order.shipping && order.shipping.country ) {
```
timmyc: Nice catch

wc-admin.php
```diff
@@ -192,3 +192,8 @@ function wc_admin_woocommerce_updated() {
 	WC_Admin_Notes_Settings_Notes::add_notes_for_settings_that_have_moved();
 }
 add_action( 'woocommerce_updated', 'wc_admin_woocommerce_updated' );
+
+/*
+ * Remove the emoji script as it always defaults to replacing emojis with Twemoji images.
```
timmyc: Should we link to the related gutenberg url you mentioned in your PR description here?

timmyc (commented):
Couple of minor comments but this is testing out well for me. I am seeing some errors on the devdocs page, but I think they might be in master too, so going to confirm that first.

/cc @LevinMedia to make sure the little wavy flags are visually okay compared to the circular counterparts.

package.json
psealock: whoops, thanks

wc-admin.php
psealock: definitely, fixed.

jeffstieler (approved):
Code looks good and tests well. 🚢 it!

_EDIT:_ after removing `react-world-flags` from `package-lock.json` 🙂

PR #1385, opened 2019-01-24
Add table summary to Customers Report

### LINKED ISSUE #918: Analytics: Customers Report Table

jeffstieler:
Fixes #918.

This PR seeks to add a table summary to the customers report, powered by a new `/reports/customers/stats` endpoint.

### Screenshots

Design:
screen shot 2019-01-23 at 6 08 59 pm

As implemented:
screen shot 2019-01-23 at 6 11 25 pm

- Open Customers Report
- Verify that the summary displays
- Verify that the totals change (and are correct) when adding filters to the report

**Note:** the aggregate values are calculated only for customers that have placed orders. If your report contains 10 customers, but only 4 have placed orders, the summary totals will be based on those 4 customers, not all 10.

### DIFF
--- a/client/analytics/report/customers/table.js
+++ b/client/analytics/report/customers/table.js
@@ -25,6 +25,7 @@ export default class CustomersReportTable extends Component {
 
 		this.getHeadersContent = this.getHeadersContent.bind( this );
 		this.getRowsContent = this.getRowsContent.bind( this );
+		this.getSummary = this.getSummary.bind( this );
 	}
 
 	getHeadersContent() {
@@ -187,6 +188,30 @@ export default class CustomersReportTable extends Component {
 		} );
 	}
 
+	getSummary( totals ) {
+		if ( ! totals ) {
+			return [];
+		}
+		return [
+			{
+				label: __( 'customers', 'wc-admin' ),
+				value: numberFormat( totals.customers_count ),
+			},
+			{
+				label: __( 'average orders', 'wc-admin' ),
+				value: numberFormat( totals.avg_orders_count ),
+			},
+			{
+				label: __( 'average lifetime spend', 'wc-admin' ),
+				value: formatCurrency( totals.avg_total_spend ),
+			},
+			{
+				label: __( 'average order value', 'wc-admin' ),
+				value: formatCurrency( totals.avg_avg_order_value ),
+			},
+		];
+	}
+
 	render() {
 		const { query } = this.props;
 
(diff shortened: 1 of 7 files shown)


includes/api/class-wc-admin-rest-reports-customers-stats-controller.php
```diff
@@ -0,0 +1,353 @@
...
+		$args['orders_count_min']    = $request['orders_count_min'];
+		$args['orders_count_max']    = $request['orders_count_max'];
+		$args['total_spend_min']     = $request['total_spend_min'];
+		$args['total_spend_max']     = $request['total_spend_max'];
+		$args['avg_order_value_min'] = $request['avg_order_value_min'];
+		$args['avg_order_value_max'] = $request['avg_order_value_max'];
+		$args['last_order_before']   = $request['last_order_before'];
+		$args['last_order_after']    = $request['last_order_after'];
+
+		$between_params = array( 'orders_count', 'total_spend', 'avg_order_value' );
```
psealock: `last_active` and `registered` are missing from this array. They are also missing in the `/reports/customers` endpoint. Perhaps a clue as to why thats not working

client/analytics/report/customers/table.js
```diff
@@ -187,6 +188,30 @@ export default class CustomersReportTable extends Component {
...
+			return [];
+		}
+		return [
+			{
+				label: __( 'customers', 'wc-admin' ),
+				value: numberFormat( totals.customers_count ),
+			},
+			{
+				label: __( 'average orders', 'wc-admin' ),
+				value: numberFormat( totals.avg_orders_count ),
```
psealock: One decimal point is enough I think.

screen shot 2019-01-24 at 3 20 15 pm

includes/api/class-wc-admin-rest-reports-customers-stats-controller.php
psealock: Created an issue to track separately here https://github.com/woocommerce/wc-admin/issues/1388

psealock (commented):
I only got halfway through this before having to stop until later, but looking good at first glance.

psealock (approved):
LGTM and working as expected, besides the issue #1386 #1387 #1388 which are best dealt with separately.

PR #1389, opened 2019-01-24
Remove comma delimited string, use array instead

jeffstieler:
Addresses this conversation: https://github.com/woocommerce/wc-admin/pull/1364/files#r250321564


timmyc:
I restarted the build on circle ci to see if the issue would ✨ magically clear up.

client/wc-api/settings/operations.js
```diff
@@ -39,15 +34,11 @@ function updateSettings( resourceNames, data, fetch ) {
 	const settingsFields = [ 'woocommerce_excluded_report_order_statuses' ];
 
 	if ( resourceNames.includes( resourceName ) ) {
-		const url = NAMESPACE + '/settings/wc_admin/';
+		const url = '/wc/v3/settings/wc_admin/';
```
justinshreve: We should still use `NAMESPACE` here so we don't have to find this and update it upon core merge.

That means we still need a v4 file, though it can be empty (see where we do this for product categories for example: https://github.com/woocommerce/wc-admin/blob/master/includes/api/class-wc-admin-rest-product-categories-controller.php).

These extra files will be needed for core merge anyway - See a v3 example at https://github.com/woocommerce/woocommerce/blob/master/includes/api/class-wc-rest-product-tags-controller.php

emdashcodes (commented):
The settings page is working for me (after resetting like in the parent PR) -- i just have one comment about the endpoint version.

client/wc-api/settings/operations.js
jeffstieler: Addressed in 8051d461.

includes/api/class-wc-admin-rest-setting-options-controller.php
```diff
@@ -16,35 +16,10 @@
...
-			} else {
-				return parent::validate_setting_multiselect_field( $values, $setting );
-			}
-		}
-
-		return $final_values;
-	}
-
-}
+}
```
justinshreve: Per WP's PHP coding standards & linter rules, PHP files must end with a newline character. I think this is causing Travis to fail as well.

justinshreve:
Closing this since the change has been integrated into #1364.

PR #1286, opened 2019-01-11
Added segmenting support to REST API

peterfabian:
Fixes #1035.

First version of segmenting for `/orders/stats` (thus it should potentially also work for `/revenue/stats`, as that one uses the same data store).

Supports segmenting `orders/stats` by:
- product: `http://local.wordpress.test/wp-json/wc/v3/reports/orders/stats/?after=2018-01-11T12:52:01&before=2019-01-08T22:52:00&interval=month&order=desc&orderby=date&segmentby=product`
- variation (when one product is selected by a filter, e.g.: `product_includes=X`): `http://local.wordpress.test/wp-json/wc/v3/reports/orders/stats/?after=2018-01-11T12:52:01&before=2019-01-08T22:52:00&interval=month&order=desc&orderby=date&segmentby=variation&product_includes=9`
- category: `http://local.wordpress.test/wp-json/wc/v3/reports/orders/stats/?after=2018-01-11T12:52:01&before=2019-01-08T22:52:00&interval=month&order=desc&orderby=date&segmentby=category`
- coupon: `http://local.wordpress.test/wp-json/wc/v3/reports/orders/stats/?after=2018-01-11T12:52:01&before=2019-01-08T22:52:00&interval=month&order=desc&orderby=date&segmentby=coupon`
- customer_type: `http://local.wordpress.test/wp-json/wc/v3/reports/orders/stats/?after=2018-01-11T12:52:01&before=2019-01-08T22:52:00&interval=month&order=desc&orderby=date&segmentby=customer_type`

Segmenting of `products/stats` by:
- product
- category
- variation

Segmenting of `coupons/stats` by:
- product
- variation
- coupon
- category

Segmenting of `taxes/stats` by:
- tax_rate_id
- are other segments needed?

### Notes:
- At the moment, the paging and sorting only applies to intervals, not to segments (second dimension), i.e. it's adding all segments to totals and all intervals returned in the response. I'm not sure if that is a viable way to handle it, in case of 1000s of products or variations, it could be slow (?).
- The segments are currently always sorted by `segment_id`--does this need to change somehow?
- There is some creative math in some places, e.g. if shipping or coupons need to be split by product/variation, etc. Currently, 
  - shipping and shipping_tax are split in `wc_order_product_lookup` according to item quantity, e.g. order 1 has 2 line items: 4 x product_1 and 3 x product_2. Shipping for product_1 is thus 4/7 * total shipping, for product_2 it's 3/7 of total shipping for the whole order.
[...]

PR #1391, opened 2019-01-24
Add Product Detail Report & Finish Comparison Chart Mode Functionality

### LINKED ISSUE #564: Product Detail Report

### LINKED ISSUE #695: Component: Comparison Chart

emdashcodes:
Fixes #564, Fixes #695. Dependent on #1286.

This PR wires up the product comparison, single product view, variation view, and variation comparison to the new segmentation API functionality. It allows you to compare many different products and variations over a set time-line. It does contain some API tweaks to make this all possible.

It also completes #695 by fully implementing the 'item-comparison' mode. Some code for this already existed and just needed finished/wired up.

Some notes/known issues for the future:
* Sometimes colors are not always different enough in graph values to tell them apart, sometimes they are. I want to log this so we can investigate further why this happens.
* The product category filters do not work yet. We should be able to pattern off of this PR to make it work though in a separate issue/PR.
*  Smooth Generator doesn’t seem to generate valid variations data, so data for those don't display properly. I added some handling around this, but it works fine with real orders. We should open up an issue in the smooth generator repo, but otherwise I don't think we need to do anything here.
* You can select up to 5 products to compare at a time on the chart. We can adjust this as needed.

### Accessibility

### Screenshots

screen shot 2019-01-24 at 2 40 58 pm

screen shot 2019-01-24 at 2 41 07 pm

screen shot 2019-01-24 at 2 42 01 pm

screen shot 2019-01-24 at 2 44 55 pm

screen shot 2019-01-24 at 2 50 06 pm

* Test the various different product detail modes on the product report. Single product, product comparison, variable product, variable comparison, Top Products filters, and Top variations filters. Check chart data is what you would expect. Check table and summary numbers as well.
* Spot check existing time based graphs.

### DIFF
--- a/client/analytics/components/report-chart/index.js
+++ b/client/analytics/components/report-chart/index.js
@@ -6,12 +6,10 @@ import { Component } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import { format as formatDate } from '@wordpress/date';
 import PropTypes from 'prop-types';
-import { find, get } from 'lodash';
 
 /**
  * WooCommerce dependencies
  */
-import { flattenFilters } from '@woocommerce/navigation';
 import {
 	getAllowedIntervalsForQuery,
 	getCurrentDates,
@@ -28,66 +26,37 @@ import { Chart } from '@woocommerce/components';
 import { getReportChartData, getTooltipValueFormat } from 'wc-api/reports/utils';
 import ReportError from 'analytics/components/report-error';
 import withSelect from 'wc-api/with-select';
-
-export const DEFAULT_FILTER = 'all';
+import { getChartMode } from './utils';
 
 /**
  * Component that renders the chart in reports.
  */
 export class ReportChart extends Component {
-	getSelectedFilter( filters, query ) {
-		if ( filters.length === 0 ) {
-			return null;
-		}
-
-		const filterConfig = filters.pop();
-
-		if ( filterConfig.showFilters( query ) ) {
-			const allFilters = flattenFilters( filterConfig.filters );
-			const value = query[ filterConfig.param ] || DEFAULT_FILTER;
-			const selectedFilter = find( allFilters, { value } );
-			const selectedFilterParam = get( selectedFilter, [ 'settings', 'param' ] );
-
-			if ( ! selectedFilterParam || Object.keys( query ).includes( selectedFilterParam ) ) {
-				return selectedFilter;
-			}
-		}
-
-		return this.getSelectedFilter( filters, query );
+	getItemChartData() {
+		const { primaryData, selectedChart } = this.props;
+		const chartData = primaryData.data.intervals.map( function( interval ) {
+			const intervalData = {};
+			interval.subtotals.segments.forEach( function( segment ) {
+				if ( segment.segment_label ) {
+					const label = intervalData[ segment.segment_label ]
+						? segment.segment_label + ' (#' + segment.segment_id + ')'
+						: segment.segment_label;
+					intervalData[ label ] = {
+						value: segment.subtotals[ selectedChart.key ] || 0,
+					};
+				}
+			} );
+			return {
+				date: formatDate( 'Y-m-d\\TH:i:s', interval.date_start ),
+				...intervalData,
+			};
+		} );
+		return chartData;
 	}
 
-	getChartMode() {
-		const { filters, query } = this.props;
-		if ( ! filters ) {
-			return;
-		}
-		const clonedFilters = filters.slice( 0 );
-		const selectedFilter = this.getSelectedFilter( clonedFilters, query );
-
-		return get( selectedFilter, [ 'chartMode' ] );
-	}
-
-	render() {
-		const {
-			interactiveLegend,
-			itemsLabel,
-			legendPosition,
-			mode,
-			path,
-			primaryData,
-			query,
-			secondaryData,
-			selectedChart,
-			showHeaderControls,
-		} = this.props;
-
-		if ( primaryData.isError || secondaryData.isError ) {
-			return <ReportError isError />;
-		}
-
+	getTimeChartData() {
+		const { query, primaryData, secondaryData, selectedChart } = this.props;
 		const currentInterval = getIntervalForQuery( query );
-		const allowedIntervals = getAllowedIntervalsForQuery( query );
-		const formats = getDateFormatsForInterval( currentInterval, primaryData.data.intervals.length );
 		const { primary, secondary } = getCurrentDates( query );
 		const primaryKey = `${ primary.label } (${ primary.range })`;
 		const secondaryKey = `${ secondary.label } (${ secondary.range })`;
@@ -115,6 +84,23 @@ export class ReportChart extends Component {
 			};
 		} );
 
+		return chartData;
+	}
+
+	renderChart( mode, isRequesting, chartData ) {
+		const {
+			interactiveLegend,
+			itemsLabel,
+			legendPosition,
+			path,
+			query,
+			selectedChart,
+			showHeaderControls,
+			primaryData,
+		} = this.props;
+		const currentInterval = getIntervalForQuery( query );
+		const allowedIntervals = getAllowedIntervalsForQuery( query );
+		const formats = getDateFormatsForInterval( currentInterval, primaryData.data.intervals.length );
 		return (
 			<Chart
 				allowedIntervals={ allowedIntervals }
@@ -122,16 +108,16 @@ export class ReportChart extends Component {
 				dateParser={ '%Y-%m-%dT%H:%M:%S' }
 				interactiveLegend={ interactiveLegend }
 				interval={ currentInterval }
-				isRequesting={ primaryData.isRequesting || secondaryData.isRequesting }
+				isRequesting={ isRequesting }
 				itemsLabel={ itemsLabel }
 				legendPosition={ legendPosition }
-				mode={ mode || this.getChartMode() }
+				mode={ mode }
 				path={ path }
 				query={ query }
 				showHeaderControls={ showHeaderControls }
 				title={ selectedChart.label }
 				tooltipLabelFormat={ formats.tooltipLabelFormat }
-				tooltipTitle={ selectedChart.label }
+				tooltipTitle={ ( 'time-comparison' === mode && selectedChart.label ) || null }
 				tooltipValueFormat={ getTooltipValueFormat( selectedChart.type ) }
 				type={ getChartTypeForQuery( query ) }
 				valueType={ selectedChart.type }
@@ -139,6 +125,40 @@ export class ReportChart extends Component {
 				x2Format={ formats.x2Format }
 			/>
 		);
+	}
+
+	renderItemComparison() {
+		const { primaryData } = this.props;
+
+		if ( primaryData.isError ) {
+			return <ReportError isError />;
+		}
+
+		const isRequesting = primaryData.isRequesting;
+		const chartData = this.getItemChartData();
+
+		return this.renderChart( 'item-comparison', isRequesting, chartData );
+	}
+
+	renderTimeComparison() {
+		const { primaryData, secondaryData } = this.props;
+
+		if ( ! primaryData || primaryData.isError || secondaryData.isError ) {
+			return <ReportError isError />;
+		}
+
+		const isRequesting = primaryData.isRequesting || secondaryData.isRequesting;
+		const chartData = this.getTimeChartData();
+
+		return this.renderChart( 'time-comparison', isRequesting, chartData );
+	}
+
+	render() {
+		const { mode } = this.props;
+		if ( 'item-comparison' === mode ) {
+			return this.renderItemComparison();
+		}
+		return this.renderTimeComparison();
 	}
 }
 
@@ -171,7 +191,7 @@ ReportChart.propTypes = {
 	/**
 	 * Secondary data to display in the chart.
 	 */
-	secondaryData: PropTypes.object.isRequired,
+	secondaryData: PropTypes.object,
 	/**
 	 * Properties of the selected chart.
 	 */
@@ -180,10 +200,21 @@ ReportChart.propTypes = {
 
 export default compose(
 	withSelect( ( select, props ) => {
-		const { query, endpoint } = props;
+		const { query, endpoint, filters } = props;
+		const chartMode = props.mode || getChartMode( filters, query ) || 'time-comparison';
+
+		if ( 'item-comparison' === chartMode ) {
+			const primaryData = getReportChartData( endpoint, 'primary', query, select );
+			return {
+				mode: chartMode,
+				primaryData,
+			};
+		}
+
 		const primaryData = getReportChartData( endpoint, 'primary', query, select );
 		const secondaryData = getReportChartData( endpoint, 'secondary', query, select );
 		return {
+			mode: chartMode,
 			primaryData,
 			secondaryData,
 		};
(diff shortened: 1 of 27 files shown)


includes/data-stores/class-wc-admin-reports-data-store.php
```diff
@@ -666,6 +666,24 @@ protected function get_included_products( $query_args ) {
...
+	 */
+	protected function get_included_variations( $query_args ) {
+		$included_variations = array();
+		$operator          = $this->get_match_operator( $query_args );
+
+		if ( isset( $query_args['variations'] ) && is_array( $query_args['variations'] ) && count( $query_args['variations'] ) > 0 ) {
+			$included_variations = $query_args['variations'];
+		}
+
+		$included_variations_str = implode( ',', $included_variations );
```
jeffstieler: I know the endpoint sanitizes these as integers, but should we be defensive and sanitize here too?

client/wc-api/products/selectors.js
```diff
@@ -0,0 +1,50 @@
...
+
+const getProducts = ( getResource, requireResource ) => (
+	query = {},
+	requirement = DEFAULT_REQUIREMENT
+) => {
+	const resourceName = getResourceName( 'product-query', query );
+	const ids = requireResource( requirement, resourceName ).data || [];
+	const products = ids.reduce(
+		( acc, id ) => ( {
+			...acc,
```
jeffstieler: I _think_ iteratively adding to an array using `.push()` will perform better here. See: https://jsperf.com/es6-add-element-to-create-new-array-concat-vs-spread-op/7

client/analytics/report/products/index.js
```diff
@@ -18,37 +19,100 @@ import { charts, filters } from './config';
...
+				: 'time-comparison';
+		const compareObject =
+			isSingleProductView && isSingleProductVariable ? 'variations' : 'products';
+		const label =
+			isSingleProductView && isSingleProductVariable
+				? __( '%s variations', 'wc-admin' )
+				: __( '%s products', 'wc-admin' );
+
+		return {
+			isProductDetailsView,
```
jeffstieler: Is this used?

client/analytics/components/report-chart/index.js
```diff
@@ -180,7 +199,16 @@ ReportChart.propTypes = {
 
 export default compose(
 	withSelect( ( select, props ) => {
-		const { query, endpoint } = props;
+		const { query, endpoint, filters } = props;
+		const chartMode = props.mode || getChartMode( filters, query ) || 'time-comparison';
```
Aljullu: Now that we are getting the chart mode inside `withSelect`, I think it would make sense to overwrite the `mode` prop at this step.

This way, inside the render method we don't need to compute the `chartMode` again, and we don't have different values for `mode` and `chartMode`, which might be confusing.

Would that work?

client/analytics/components/report-chart/utils.js
```diff
@@ -0,0 +1,44 @@
...
+import { find, get } from 'lodash';
+
+/**
+ * WooCommerce dependencies
+ */
+import { flattenFilters } from '@woocommerce/navigation';
+
+export const DEFAULT_FILTER = 'all';
+
+export function getSelectedFilter( filters, query, selectedFilterArgs = {} ) {
```
Aljullu: Thanks for extracting that into a `utils` file. :clap:

packages/components/src/chart/index.js
```diff
@@ -228,14 +249,14 @@ class Chart extends Component {
...
-		const { legendPosition, mode, isViewportWide } = this.props;
+		const { legendPosition, mode, isViewportWide, isViewportMedium } = this.props;
 		if ( legendPosition ) {
 			return legendPosition;
 		}
 		if ( isViewportWide && mode === 'time-comparison' ) {
 			return 'top';
 		}
-		if ( isViewportWide && mode === 'item-comparison' ) {
+		if ( isViewportMedium && mode === 'item-comparison' ) {
```
Aljullu: This change is making the chart to look very small in narrow viewports:

image

AFAIK, the goal was to switch to a single column layout (the legend appearing below the chart) in viewports smaller than 1280px. Maybe @LevinMedia can confirm.

Aljullu (commented):
It's amazing seeing this taking shape. Good job @justinshreve! :tada:

LevinMedia (commented):
Really, really excited to see this come to life! 🎉 

---

Intermittently when I select five products to compare only four show up. All five appear in the filter, and all five appear in the tabular data... it's just missing one in the chart: 

screen recording 2019-01-25 at 08 49 am

I've also seen cases where even fewer show up too:

screen shot 2019-01-25 at 9 14 14 am

---

When selecting a large number of items to compare, not all are appearing as an option to check in the chart. In the example below, I selected 25 products, but it's cut off at 15. It happens with smaller values too. I tested comparing 10 products, but only 8 appeared in the chart. 

screen recording 2019-01-25 at 08 57 am

---

Regarding the color scale you mentioned as a know issue... I'm pretty sure I know what's going on here... Back in the day, Rob and I specified colors for charts with five stops or less: 

https://github.com/woocommerce/wc-admin/issues/375

When there are five or more series, we let math takeover, and we create as many stops as we need... so what I assume is happening, is that when you compare a huge number of things, the color scale is based on the total number of compared items, rather than the specific number of things we allow people to compare in the chart at once (in this case, five). You can see that happening in the gif above - all the colors are deep purple / blue, on the far end of the scale. 

---

This one is more difficult to reproduce - however, I've seen it happen pretty consistently in testing. When I click "Clear All" in the comparison filter and the chart starts to reset, all of the compared items move into the chart header for a moment, before the two default time based series appear. 

screen shot 2019-01-25 at 9 17 21 am

the last time I reproduced this, I selected the items to compare, made the comparison, then changed the summary number. When I clicked clear all, they flashed in the chart header. 

---

When hovering over the various series in the legend, there's a flash of all series each time you hover between a series. It would be great if this was a smooth flow that would highlight each series without the flash in between. The GIF doesn't pick it up so well because of the low frame rate. 

screen recording 2019-01-25 at 09 26 am

---

I haven't had a chance to test the single product w/ variations view. I'll set it up and make a few sales to report back this afternoon.  Did I mention this is amazing? Thanks again.

packages/components/src/chart/index.js
LevinMedia: Affirmative - lets switch to single column view <1280 thanks for the catch @Aljullu !

LevinMedia:
One other bug I just noticed - when viewing a bar chart, and it's in the the scrollable mode (happens when there are more bars than pixels in the chartable area) it blows out of the container. the overflow should be hidden. 

screen shot 2019-01-25 at 9 56 22 am

client/analytics/components/report-chart/index.js
```diff
@@ -180,7 +199,16 @@ ReportChart.propTypes = {
...
+
+		if ( 'item-comparison' === chartMode ) {
+			const primaryData = getReportChartData( endpoint, 'primary', query, select );
+			return {
+				primaryData,
+			};
+		}
+
 		const primaryData = getReportChartData( endpoint, 'primary', query, select );
 		const secondaryData = getReportChartData( endpoint, 'secondary', query, select );
```
psealock: I see the `secondaryData` still being requested. Is that because a variation isn't detected until a products request returns?

client/analytics/report/products/index.js
```diff
@@ -62,3 +126,34 @@ ProductsReport.propTypes = {
...
 };
+
+export default compose(
+	withSelect( ( select, props ) => {
+		const { query } = props;
+		const { getProducts, isGetProductsRequesting, getProductsError } = select( 'wc-api' );
+		const isSingleProductView = query.products && 1 === query.products.split( ',' ).length;
+		if ( isSingleProductView ) {
+			const includeArgs = { include: query.products };
+			const products = getProducts( includeArgs );
```
psealock: We make this exact same request to populate tags in the Search component. Just noting that there is an opportunity here (in another PR).

psealock (commented):
Wow @justinshreve ! This is testing well for me and I'd like to continue tomorrow. I just left two comments to note things to potentially address in a follow up PR

client/analytics/components/report-chart/index.js
justinshreve: I can't seem to reproduce this for a variable product. Can you try again and/or let me know what your request looks like?

I added a console.log here, and I only see `secondaryData` when a time comparison chart is happening (a single product that is not variable, or the all products view) -- for comparison charts and the variation view, I don't see the console log.

client/analytics/report/products/index.js
justinshreve: I left a `TODO` for this in https://github.com/woocommerce/wc-admin/pull/1391/commits/275356e15585f88dc9f21d0a676dc6e111633c0d.

packages/components/src/chart/index.js
justinshreve: Done in https://github.com/woocommerce/wc-admin/pull/1391/commits/275356e15585f88dc9f21d0a676dc6e111633c0d.

client/analytics/components/report-chart/index.js
justinshreve: Good catch - updated in https://github.com/woocommerce/wc-admin/pull/1391/commits/275356e15585f88dc9f21d0a676dc6e111633c0d.

client/wc-api/products/selectors.js
justinshreve: I cleaned this up in https://github.com/woocommerce/wc-admin/pull/1391/commits/275356e15585f88dc9f21d0a676dc6e111633c0d to match other selectors.

includes/data-stores/class-wc-admin-reports-data-store.php
justinshreve: I added some extra sanitization here to be safe. https://github.com/woocommerce/wc-admin/pull/1391/commits/275356e15585f88dc9f21d0a676dc6e111633c0d

client/analytics/components/report-chart/index.js
```diff
@@ -115,23 +84,41 @@ export class ReportChart extends Component {
...
+	renderChart( mode, isRequesting, chartData ) {
+		const {
+			interactiveLegend,
+			itemsLabel,
+			legendPosition,
+			path,
+			query,
+			selectedChart,
+			showHeaderControls,
+			chartMode,
```
Aljullu: I think this should have changed from `chartMode` to `mode`, otherwise it's always `undefined` for me.

packages/components/src/chart/d3chart/style.scss
```diff
@@ -7,6 +7,7 @@
 .d3-chart__container {
 	position: relative;
 	width: 100%;
+	overflow-x: scroll;
```
Aljullu: This is adding a scrollbar under all charts in Linux (probably Windows too).

image

I took some time trying to figure out what was going on here, given that I [worked on this](https://github.com/woocommerce/wc-admin/pull/1226) some time ago. The issue seems to be that when the legend is on the left, `.d3-chart__container` doesn't adapt its size.

Adding:

```CSS
.woocommerce-chart__body-row .d3-chart__container {
	width: calc( 100% - 320px );
}
```

to `chart/d3chart/style.scss` seems to fix the issue for me.

Let me know if that helps. :+1:

Aljullu (commented):
Thanks for addressing the feedback, @justinshreve! I just noticed a couple of small things.

client/analytics/components/report-chart/index.js
justinshreve: Good catch -- fixed to `mode`.

packages/components/src/chart/d3chart/style.scss
justinshreve: @Aljullu Thanks for digging into that and finding the answer! That fix is working for me as well. I've updated the PR.

LevinMedia:
@justinshreve all of your updates are looking good to me. Thanks for all the work on this!

LevinMedia:
@justinshreve I just discovered two more issues, but I think they look like candidates for new PRs. 

When testing the single product view with variations, looking at data from "Today" the entire x-axis says 12am - but only when you're in bar-chart view. 

screen shot 2019-01-29 at 2 18 52 pm

Second issue: the times are off, My store is set to Los Angeles time, which means I placed this order in the 2pm hour. It looks like it's calculating off of GMT. (+8hrs)

justinshreve:
I'll still have to hold off on merging until the segmentation PR is ready, but does this seem ready to get a ✅ and remove the review label?

timmyc (approved):
This PR has gotten more 👀than most - and rightfully so because it is some amazing work! I'm marking this as ready to land. As @LevinMedia pointed out, lets do open up issues for any remaining items that were found and not fully fixed up here. I definitely think this could use some sweep testing too to suss out other edge cases that have yet to be found.

And yes lets do rally around the segmentation REST PR now and get that one pushed over the finish line too so we can get this in 🎉

PR #1400, opened 2019-01-26
Update usage of lib/number to @woocommerce/number

timmyc:
This builds upon #1399 and the introduction of the new `@woocommerce/number` package.

All usage of `formatNumber` has been updated to use the export from the new package. And also to get around the circular dependency between `currency` and `number`, the use cases of where `formatNumber()` was being called with a type of currency have been updated to use `formatCurrency` directly.

__To Test__
- Click around through the dashboard and reports, and verify all things are still rendering as expected.


timmyc:
Moving this back to in-progress to fix the failing test

timmyc:
I fixed the tests by adding locutus in as a dev dependency. It seems that maybe until they support proper es6 `import` this might be the only option, but all ears to other ideas.

emdashcodes (approved):
This is testing well for me. Changes look good as well.

Just one question I had, I saw this earlier: https://github.com/woocommerce/wc-admin/blob/fe46fe1955b75fa8923af2e80d54b792c38a994b/tests/js/setup-globals.js#L24 -- should we add `number` there as well?

Pre-approving.

PR #1409, opened 2019-01-29
Build: Updates from latest npm publish

timmyc:
Just did a `lerna version` and publish to npm, these are the resulting changes.

Verify wc-admin boots up and loads up properly and tests pass.


emdashcodes (approved):
Versions are matching what I see on npm. Builds/tests correctly, and all tests pass.

PR #1410, opened 2019-01-29
Add 'Clear' button to Search input

Aljullu:
Follow-up of #1381.

Adds a _Clear_ button to table search input.

### Accessibility

### Screenshots
search-2

- Go to any report with a table search input.
- Make a search and verify a _Clear_ button appears.
- Press it and verify it removes the search terms.
- Add an advanced filter and verify the _Clear_ button doesn't appear when you add a term.


timmyc:
See that this is still blocked, but any thoughts about adding `esc` key support to trigger the same clearing functionality?

Aljullu:
@joshuatf thanks for the review!

> See that this is still blocked, but any thoughts about adding `esc` key support to trigger the same clearing functionality?

@timmyc Currently, the `ESCAPE` key hides the autocompleter but doesn't clear the input box contents. I think that's the correct behavior or, at least, the one I would expect. Do you think that's ok?

PR #1412, opened 2019-01-29
Update loopkup tables when an order is deleted

### LINKED ISSUE #1307: Lookup tables should be updated when orders are deleted
**Describe the bug**
The lookup tables are not affected when an order is deleted causing incorrect stats.  Each data store should contain information about how to remove data when deletion occurs.

https://github.com/woocommerce/wc-admin/pull/1180 includes an action to remove order stats data when an order is deleted, but the other tables are not yet complete.

**To Reproduce**

1. Delete an order.
2. Note the record/stats are still in the tables.

**Expected behavior**
Expected deleting orders to update the lookup tables.

claudiosanches:
Fixes #1307

The lookup tables are not affected when an order is deleted causing incorrect stats. Each data store should contain information about how to remove data when deletion occurs.

1. Create an order with products, taxes and coupons.
2. You can check that data is created in `wc_order_stats`, `wc_order_coupon_lookup`, `wc_order_product_lookup`, and `wc_order_tax_lookup`.
3. Delete your order created in step 1.
4. Check your database, note that all order data is only deleted from `wc_order_stats`, but still in the other tables.
5. Apply this patch and repeat steps 1, 2 and 3.
6. Check that all order data is cleaned from all lookup tables.

Note: There's some code duplication to delete order, still I think that is best to control it in each data store, instead of handing all inside `WC_Admin_Reports_Orders_Stats_Data_Store::delete_order`.


includes/data-stores/class-wc-admin-reports-orders-stats-data-store.php
```diff
@@ -455,18 +455,26 @@ public static function update( $order ) {
...
+				$order_id
 			)
 		);
+
+		/**
+		 * Fires when orders stats are deleted.
+		 *
+		 * @param int $order_id Order ID.
+		 */
+		do_action( 'woocommerce_reports_orders_stats_data_deleted', $order_id );
```
joshuatf: Nice!  Love this solution.

includes/data-stores/class-wc-admin-reports-coupons-data-store.php
```diff
@@ -58,6 +58,7 @@ public static function init() {
 		add_action( 'save_post', array( __CLASS__, 'sync_order_coupons' ) );
 		add_action( 'clean_post_cache', array( __CLASS__, 'sync_order_coupons' ) );
 		add_action( 'woocommerce_order_refunded', array( __CLASS__, 'sync_order_coupons' ) );
+		add_action( 'woocommerce_reports_orders_stats_data_deleted', array( __CLASS__, 'sync_on_order_delete' ), 10 );
```
joshuatf: One thing I noticed is that this is only fired when an order is deleted permanently.  I'm wondering if we should also fire this hook when an order is trashed as users might be confused about the resulting data.

Or should we keep the behavior as is and make a note of this in user docs?

includes/data-stores/class-wc-admin-reports-coupons-data-store.php
```diff
@@ -345,4 +346,22 @@ public static function sync_order_coupons( $order_id ) {
 		}
 	}
 
+	/**
+	 * Clean coupons data when an order is deleted.
+	 *
+	 * @param int $order_id Order ID.
+	 */
+	public static function sync_on_order_delete( $order_id ) {
```
joshuatf: Since this function is the same for all data stores, should we move it into `WC_Admin_Reports_Data_Store` to dry things up a bit?

tests/reports/class-wc-tests-reports-orders-stats.php
```diff
@@ -3005,4 +3005,104 @@ public function test_populate_and_query_multiple_intervals() {
...
+	 * @covers WC_Admin_Reports_Coupons_Data_Store::sync_on_order_delete
+	 * @covers WC_Admin_Reports_Taxes_Data_Store::sync_on_order_delete
+	 */
+	public function test_order_deletion() {
+		global $wpdb;
+
+		WC_Helper_Reports::reset_stats_dbs();
+
+		// Tables to check.
+		$tables = array(
```
joshuatf: Thanks for these tests.  Very concise here 👍

peterfabian:
Side note: it would be nice to also handle trashed orders in the same issue, as it's closely related. 
If you think it would be better to be handled in a separate issue/PR, we can create a new one, but either way, I think trashed orders should not be included in the stats.

includes/data-stores/class-wc-admin-reports-coupons-data-store.php
claudiosanches: We can exclude when an order is moved to trash, also we can restore it if the order is restored. No problem.

claudiosanches:
@peterfabian will be easier to include AS just after #1426 is merged. I as about to start working on AS, but stopped since Jeff started another integration with that, so I prefer wait to have the things done by Jeff first.

includes/data-stores/class-wc-admin-reports-coupons-data-store.php
claudiosanches: Since we need new hooks for https://github.com/woocommerce/wc-admin/pull/1459 probably will make no big difference.

claudiosanches:
@joshuatf @peterfabian

> Side note: it would be nice to also handle trashed orders in the same issue, as it's closely related.
If you think it would be better to be handled in a separate issue/PR, we can create a new one, but either way, I think trashed orders should not be included in the stats.

I'm checking that `wc_order_stats` already have a `status` table, I think that is better handle filtering by status, instead of trying to change statuses for all those objects as well, basically is not possible to "trash" any of this objects because they don't handle statuses.

This is definitely something that we could discuss, but sounds like a problem for another PR.

> One thing I noticed is that this is only fired when an order is deleted permanently. I'm wondering if we should also fire this hook when an order is trashed as users might be confused about the resulting data.
>
> Or should we keep the behavior as is and make a note of this in user docs?

If handing "trash" we should use `woocommerce_trash_` instead of `woocommerce_delete_`, like WooCommerce core already does:

https://github.com/woocommerce/woocommerce/blob/99b633d1e562497700729aaf8411837109bb17da/includes/data-stores/class-wc-coupon-data-store-cpt.php#L207-L217

claudiosanches:
@joshuatf I almost forget about it:

>  I think we may also want to sync in the `WC_Admin_Reports_Downloads_Data_Store` store as well.

WooCommerce core already controls `wc_download_log` table, so by default when an order is deleted, all download data is removed.

claudiosanches:
Travis is having trouble with one of the tests in this last build: https://travis-ci.org/woocommerce/wc-admin/jobs/487640022#L507
But seems to be fine by the other tests.

PR #1423, opened 2019-01-30
Add transient notices

### LINKED ISSUE #806: Add Support for Transient Notifications

Iinbox notifications are intended for prompting the user of things like:
– they have a new best sales in a day record
– they have an expiring (or expired) woo extension subscription
– they haven’t set up something that wasn’t ever set up before
– content that could help with the day to day tasks of managing and optimizing a store

And Store Alerts are intended for
– shipping is enabled, but there are no shipping methods in any shipping zones
– things that had been working but are now not working (e.g. API key no longer valid)
– version and/or security updates

But some of our existing admin notices don’t seem to fit well into either category, e.g.:

screen shot 2018-11-08 at 4 30 09 pm

In this particular example, the store manager updated the status on two bookings – @laurendavissmith  pointed this one out to me today.

@LevinMedia wrote:

> sounds like a transient notification – we’ve done something similar with woocommerce services. Can we take the same approach there? Per our IRL convo, let’s try sliding it in on the bottom right corner of the screen

p90Yrv-S3-p2

I also discussed this briefly with @ryelle - we'll want to ensure the solution is an accessible one

Let's prioritize this immediately after Store Alerts, that way extension developers will then have the tools to be able to cover all three kinds of notices.

### LINKED ISSUE #1420: Saved settings should show success notification
#1364 introduces a Settings page.  When settings are saved at `wp-admin/admin.php?page=wc-admin#/analytics/settings`, a success notification should be shown to the user on successful save.  

In order to implement this, we need to first resolve https://github.com/woocommerce/wc-admin/issues/806 which brings in transient notifications.

joshuatf:
Fixes #806 fixes #1420 

Adds in support for transient notices so we can display "toast" style notifications.

### Accessibility

### Screenshots
screen shot 2019-01-31 at 7 19 07 pm

1.  Visit the Settings page `wp-admin/admin.php?page=wc-admin#/analytics/settings`.
2. Click on "Save changes."
3. Note the transient notice in the bottom right corner of the screen.
4. Using a screen reader, test the notice again and make sure that it's read aloud.

### Questions
* Should we keep using `wc-api` for this or is worth creating a separate store?
* Styling is taken directly from the Gutenberg component.  Should we make any changes here  @LevinMedia?


client/wc-api/notices/operations.js
```diff
@@ -0,0 +1,25 @@
...
+	return [ ...createNotice( resourceNames, data ) ];
+}
+
+function createNotice( resourceNames, data ) {
+	const resourceName = 'notices';
+
+	if ( resourceNames.includes( resourceName ) ) {
+		const notice = { ...{ id: uniqueId( 'notice-' ) }, ...data[ resourceName ] };
+		// @TODO: this should append the notice to the existing state.
+		return [ { [ 'notices' ]: { data: [ notice ] } } ];
```
joshuatf: This is currently not a great implementation as it overwrites anything already existing in the store instead of appending a new one.  

Is there any way to grab an existing copy of the store to merge the new notice in or should we be creating an entirely new store instead of trying to fit this into fresh data? /cc @coderkevin

client/wc-api/settings/operations.js
```diff
@@ -62,7 +62,7 @@ function updateSettings( resourceNames, data, fetch ) {
 function settingsToSettingsResource( settings ) {
 	const settingsData = {};
 	settings.forEach( setting => ( settingsData[ setting.id ] = setting.value ) );
-	return { [ 'settings' ]: { data: settingsData } };
+	return { [ 'settings' ]: { data: settingsData, lastReceived: new Date() } };
```
joshuatf: Updates here don't actually seem to get mapped to the store.  We need `lastReceived` to detect whether or not the update is still in progress and I would expect the error to get mapped to the store if the fetch in `updateSettings()` fails.

Any thoughts on how to handle this or did I approach this the wrong way? /cc @coderkevin

client/header/transient-notices/index.js
```diff
@@ -0,0 +1,48 @@
...
+import withSelect from 'wc-api/with-select';
+
+class TransientNotices extends Component {
+	render() {
+		const { className, notices } = this.props;
+		const classes = classnames( 'woocommerce-transient-notices', className );
+
+		return (
+			<section className={ classes }>
+				{ notices && notices.map( notice => <TransientNotice key={ notice.id } { ...notice } /> ) }
```
timmyc: minor but should we just use the option second iterator argument from Array.map to populate the key here since theoretically we don't have any logic in place to prevent duplicate key.ids?

client/header/transient-notices/transient-notice.js
```diff
@@ -0,0 +1,102 @@
...
+		return (
+			<CSSTransition in={ this.state.visible } timeout={ 300 } classNames="slide">
+				<div className={ classes } role="alert">
+					<Notice
+						status={ status }
+						isDismissible={ isDismissible }
+						onRemove={ onRemove }
+						actions={ actions }
+					>
+						{ message }
```
timmyc: I recall us discussing somewhere ( can't quite remember where ) some of the a11y implications of these toast-style notifications. I seem to think we talked about using [speak](https://github.com/WordPress/gutenberg/blob/master/packages/a11y/src/index.js#L33) here but /cc @ryelle in case you might have some thoughts on how best to handle this.

client/wc-api/settings/selectors.js
```diff
@@ -1,14 +1,35 @@
...
  * Internal dependencies
  */
 import { DEFAULT_REQUIREMENT } from '../constants';
 
 const getSettings = ( getResource, requireResource ) => ( requirement = DEFAULT_REQUIREMENT ) => {
-	return requireResource( requirement, 'settings' ).data;
+	return requireResource( requirement, 'settings' ).data || {};
+};
+
+const getSettingsError = getResource => () => {
```
timmyc: Should we add some tests for these new methods here?

client/header/index.js
```diff
@@ -80,6 +81,14 @@ class Header extends Component {
 			'is-scrolled': isScrolled,
 		} );
 
+		const notices = [
```
timmyc: Is this testing data?

timmyc (commented):
Couple of quick comments and questions - but this is testing well for me.

client/header/transient-notices/index.js
joshuatf: I think that’s probably a good idea. My original thought here was that we could pass in IDs to override existing messages. However since the use case for that may be slim, we can probably remove it.

client/header/index.js
joshuatf: It was test data that was removed but not pushed apparently 😓

client/header/transient-notices/transient-notice.js
joshuatf: Interesting.  Any benefit to using `speak()` over `role="alert"` here @ryelle ?

client/wc-api/settings/selectors.js
joshuatf: I don't think we have tests around any of the functions in `wc-api` as of right now.  Certainly not a bad idea though- do we want to implement tests for all of these?

I can add some in pending @coderkevin's response as to whether or not we continue using `wc-api` here or create a separate store.

client/header/transient-notices/transient-notice.js
ryelle: `speak` uses a container that has slightly different semantics – `role="alert"` will interrupt the screen reader as soon as it changes, while with `speak` you can control whether it should "politely" interrupt (waiting for the SR to finish reading the current action, generally preferred), or "assertively" interrupt (same behavior as `role="alert"`).

You might also want to use it to give slightly different content to a screen reader– for example, a lot of "toast-style" notices are hard to navigate focus to, so while sighted users can see and mouse to a button/link, maybe we just want to read a string message to a screen reader user (with the usual caveat that we shouldn't be showing very different content to SR users).

joshuatf:
@coderkevin Thanks for the input!  I figured that would be the case.  I'll create a new store for this data then.

client/wc-api/settings/operations.js
coderkevin: `lastReceived` should automatically be set by `fresh-data` when you return the resources from the promise. I'm not sure what the problem is here. I'll contact you.

client/analytics/settings/index.js
```diff
@@ -59,9 +61,31 @@ class Settings extends Component {
 		}
 	};
 
+	componentDidUpdate() {
+		const { isError } = this.props;
+
+		// @TODO: Add `&& ! isRequesting` to this condition when lastReceived is correctly updated.
+		if ( this.state.saving ) {
+			if ( ! isError ) {
+				this.props.addNotice( {
```
Aljullu: Nitpicking here, but given that we are destructuring `this.props` 5 lines above, could we extract `addNotice` there too?
```ES6
const { addNotice, isError } = this.props;
```

As a personal taste, I would also destructure `this.state` at the beginning of the method, this way it's easy to see which props/state properties a method is using just looking at the first lines of it.

joshuatf: Absolutely.  Good idea 👍

client/header/transient-notices/transient-notice.js
joshuatf: Thanks for that explanation, @ryelle.  Added in `speak` in favor of `role="alert"` here.

joshuatf:
@Aljullu This is ready for another look if you have time.  Registered a separate store to house and append notices.

@josemarques Could you check the design on this if you have a few?  Styling is mostly pulled from Gutenberg, but want to make sure spacing and positioning looks okay to you.

client/header/transient-notices/transient-notice.js
```diff
@@ -0,0 +1,104 @@
...
+				</div>
+			</CSSTransition>
+		);
+	}
+}
+
+TransientNotice.propTypes = {
+	/**
+	 * Array of action objects.
+	 * See https://wordpress.org/gutenberg/handbook/designers-developers/developers/components/notice/
```
timmyc: Nice having the links here to the docs 🙌

timmyc (approved):
Code is looking great, thanks for all the tests, and I tested with voice over and `speak` is working well in my testing.

joshuatf:
@timmyc Thanks for the review!

@josemarques If you get a chance, can you let me know if you think we need to make any design changes before merging this one in?

josemarques:
@joshuatf I think we should apply the Google Material snackbar concept for these transient notices:
https://material.io/design/components/snackbars.html

> Snackbars inform users of a process that an app has performed or will perform. They appear temporarily, towards the bottom of the screen. They shouldn’t interrupt the user experience, and they don’t require user input to disappear.

Specifically:
- We can keep the styling of the notices but need to add drop shadow to emphasize elevation in that element.
- Material places these elements at the bottom left of the page. 


Here's a suggestion for the box shadow values:
box-shadow: 0 3px 6px rgba(0,0,0,0.16), 0 3px 6px rgba(0,0,0,0.23)

joshuatf:
@josemarques Fantastic- thanks for reviewing!  It does look much better with a box shadow.

I'm not sure if you were in the discussion, but apparently there was an IRL discussion (referred to in p90Yrv-S3-p2) around placing notices in the bottom right.  I went ahead and moved to the bottom left for now, but if that needs to be changed again, let me know.

Let me know if everything looks good now.

josemarques:
> @josemarques Fantastic- thanks for reviewing! It does look much better with a box shadow.
> 
> I'm not sure if you were in the discussion, but apparently there was an IRL discussion (referred to in p90Yrv-S3-p2) around placing notices in the bottom right. I went ahead and moved to the bottom left for now, but if that needs to be changed again, let me know.
> 
> Let me know if everything looks good now.

I don't have wc-admin set up atm so I'll trust your best judgement.

joshuatf:
Okay, I'll go ahead and get this merged in for now then.  We can open another issue if we want to make any styling changes going forward.

PR #1438, opened 2019-01-31
Merge getOrders and getItems

Aljullu:
Follow-up of #1381.

This PR merges `getOrders()` with `getItems()`.

### Screenshots
_(this PR shouldn't have produced any visible change)_

- Go to the _Products_ report and verify product categories are displayed correctly.
- Go to the _Categories_ report and verify they are displayed correctly.
- Try making searches in a table and verify table items are filtered properly.
- Open the _Orders_ activity panel and verify they are ordered with most recent first.

### Rationale behind using JS maps

- For `getOrders` we need results to stay ordered as they come from the API, that's why we were using an array.
- For `getItems` we need fast access to an item by id (see `categories[ category_id ]` in the diff below :point_down: ), that's why we were using an object.

The only JavaScript data structure that has both benefits are JS maps: they keep values sorted while granting fast access by key.

I noticed we are not using JS maps anywhere else in `wc-admin` or `gutenberg`, but it still seemed the correct data structure to use here and Babel should take care of transpiling it in case there is any browser incompatibility.


timmyc (approved):
This is testing out well for me - also took it for a spin in IE11 due to its partial support for `map` and confirmed all is well there too!

PR #1439, opened 2019-01-31
Trim search box input

### LINKED ISSUE #1437: Trim Search box text
From: https://github.com/woocommerce/wc-admin/pull/1381#pullrequestreview-198359354

> __Trim Query / Don't search for empty strings__
> If I just enter spaces into a search field, a search request to the API is still sent - thinking we should trim that value and only do a search if there is a length:
> 
> search-with-spaces

It should probably be applied to the `Search` component, so it affects the table search and filters.

Aljullu:
Fixes #1437.

Trims the input of search boxes.

### Screenshots
image
_(notice the spaces before the search query are ignored)_

- Go to the _Products_ report and select _Product Comparison_ as the filter.
- Open the _Network_ tab of your browser devtools.
- Start typing spaces.
- Verify no API calls are sent.
- After the spaces, write `sun` (or any string that matches one of your products), and verify `Sunglasses` appear.
- Remove everything in the input box and verify no extra API call is sent.

### DIFF
--- a/packages/components/src/search/test/autocomplete.js
+++ b/packages/components/src/search/test/autocomplete.js
@@ -3,6 +3,7 @@
  * External dependencies
  */
 import { mount } from 'enzyme';
+import { Button } from '@wordpress/components';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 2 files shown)


packages/components/src/search/test/autocomplete.js
```diff
@@ -35,6 +60,27 @@ describe( 'Autocomplete', () => {
 		autocomplete.instance().search( { target: { value: search } } );
 		autocomplete.update();
 
-		expect( autocomplete.find( '.' + suggestionClassname ).length ).toBe( 1 );
+		expect( autocomplete.find( Button ).filter( '.' + suggestionClassname ).length ).toBe( 1 );
+	} );
+
+	it( 'trims spaces from input', () => {
```
timmyc: 🌟 very nice

timmyc (approved):
Working as expected - thanks for fixing this up and for adding tests.

PR #1440, opened 2019-01-31
Add GPL License

### LINKED ISSUE #1434: License Request
**problem:**
there is no `LICENSE` file in the project

**solution I'd like:**
Adding an `Apache 2.0` copy into `LICENSE` file would be more then enough.

**Alternatives I've considered:**
the `BSD` and `MIT` license's are really popular and they do make things lot easier,
Humans have a short live, who knows how many hours this project did require,
In short any license is better than `GPL`

I would like to start using and maybe extending and developing on this project but can not do that unless there is a LICENSE file in the project, if there are multiple contributors this could also in the end lead to no one having full authority on the project

**Additional context**
It would take lot of time and efforts to latter change the license to `Apache 2.0`,
it is best to do it at the begin see below for a license change example:
SoftEtherVPN/SoftEtherVPN/issues/516

emdashcodes:
In `CONTRIBUTING.md`, we state our license and link to `LICENSE.md`, but that file doesn't exist. 

This PR adds a license file.

Fixes #1434.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,264 @@
+The GNU General Public License, Version 2, June 1991 (GPLv2)
+============================================================
+
+> Copyright (C) 1989, 1991 Free Software Foundation, Inc.
+> 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
+
+Everyone is permitted to copy and distribute verbatim copies of this license
+document, but changing it is not allowed.
+
+
+Preamble
+--------
+
+The licenses for most software are designed to take away your freedom to share
+and change it. By contrast, the GNU General Public License is intended to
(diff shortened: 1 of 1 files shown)


Aljullu (commented):
Good catch! This is looking good to me. I noticed in `gutenberg`'s [license](https://github.com/WordPress/gutenberg/blob/master/LICENSE.md) there is a preamble (`WordPress - Web publishing software`) and one more section (`END OF TERMS AND CONDITIONS`).

I wonder if we would like to have those in our license file too.

justinshreve:
Thanks for the review! Good point -- it looks like the WC core Repo contains some of those sections already too so I've updated this PR to match Gutenberg's format.

timmyc (approved):
LGTM

PR #1442, opened 2019-01-31
Packages: Release @woocommerce/components 1.4.2

ryelle:
[1.4.2 has already been released,](https://www.npmjs.com/package/@woocommerce/components) this PR just updates the changelog & bumps the version.

I think the package release docs need to be updated, but I'm not entirely sure I did this correctly – I manually changed the version number in package.json, so it only released the components package.


timmyc (approved):
Thanks for bumping the version @ryelle

PR #1444, opened 2019-02-01
Update page query to 1 when base query is changed

### LINKED ISSUE #1405: Page query parameter can hide data when switching/filtering reports
**Describe the bug**
When switching from a report with `page` query parameter greater than one to a report or filtered version of the same report with less than one page of data makes it look like there is no data.

**To Reproduce**

1. View a report with many pages of data
2. Click to the second or third page of tabular data
3. Filter the report so that less than one page of data (<25 items) is returned.
4. See no rows in the table and no indication that you are looking at a non-existent second page

**Expected behavior**
Either revert back to page 1 or give an indication that you are looking at page 2 and a way to go back to page 1.

**Screenshots**
screen shot 2019-01-29 at 12 04 57 pm

joshuatf:
Fixes #1405 

Switches back to the first page whenever the query is modified by changing filters or sorting.

### Screenshots
feb-01-2019 17-08-22

1. Go to any report page that has multiple pages.
2. Use the pagination to navigate to any page after page 1 on the report table.
3.  Remain on the same page and change the filter or sorting.
4.  Note that the report has returned to the first page.

### DIFF
--- a/client/layout/controller.js
+++ b/client/layout/controller.js
@@ -9,7 +9,7 @@ import { find, last, isEqual } from 'lodash';
 /**
  * WooCommerce dependencies
  */
-import { getPersistedQuery, stringifyQuery, updateQueryString } from '@woocommerce/navigation';
+import { getNewPath, getPersistedQuery, history, stringifyQuery } from '@woocommerce/navigation';
 
 /**
  * Internal dependencies
@@ -70,7 +70,7 @@ class Controller extends Component {
 		const baseQuery = this.getBaseQuery( this.props.location.search );
 
 		if ( prevQuery.page > 1 && ! isEqual( prevBaseQuery, baseQuery ) ) {
-			updateQueryString( { page: 1 } );
+			history.replace( getNewPath( { page: 1 } ) );
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


client/layout/controller.js
```diff
@@ -64,11 +64,35 @@ const getPages = () => {
 };
 
 class Controller extends Component {
+	componentDidUpdate( prevProps ) {
+		const prevQuery = this.getQuery( prevProps.location.search );
+		const prevBaseQuery = this.getBaseQuery( prevProps.location.search );
+		const baseQuery = this.getBaseQuery( this.props.location.search );
+
+		if ( prevQuery.page > 1 && ! isEqual( prevBaseQuery, baseQuery ) ) {
+			updateQueryString( { page: 1 } );
```
Aljullu: `updateQueryString` relies on `history.push()`, which adds a new element to the user's browser history. I think in this case it might make more sense to use `history.replace()`.

```suggestion
			history.replace( getNewPath( { page: 1 } ) );
```

You can import `history` and `getNewPath` from `@woocommerce/navigation` as well.

Aljullu (commented):
This is testing well and code looks good, I just added a small suggestion.

Aljullu (approved):

client/layout/controller.js
joshuatf: Definitely makes more sense- thanks for that!

PR #1447, opened 2019-02-01
Fix incorrect label async request paths

### LINKED ISSUE #1424: 404 request when comparing two products
**To Reproduce**
- Select _Product Comparison_ as the filter.
- Select two products to compare.
- Click on `Compare`.
- Check the `Network` tab in your browser devtools and verify there is a call to the endpoint `/wc/v4products` (notice the missing slash) that returned 404.
image

### LINKED ISSUE #1425: Product report: reloading the page with an active comparison clears the filters from the UI but doesn't update the chart/table
**To Reproduce**
- Select _Product Comparison_ as the filter.
- Introduce two products and click on `Compare`.
- Verify the chart and table are updated.
- Reload the page.
- Notice the tags disappeared from the _Compare Products_ filter BUT the chart and table still display the comparison being applied. That's confusing and when it happens there is no easy way to clear the comparison.

compare-2

I think the correct solution would be preserving the filters between reloads.

joshuatf:
Fixes #1424 fixes #1425 

Fixes the request paths for getting labels to prevent the 404 and also show persisted comparison labels on page reload.

### Screenshots
screen shot 2019-02-01 at 5 40 38 pm

1. Go to the products report.
2. Select 2 product using the product comparison filter and click "Compare."
3. Make sure no 404s are shown under network requests.
4. Refresh the page.
5.  Make sure that the comparison tags persist after reloading.
6.  Make sure that coupons, customers, and product variation comparison filters continue to work as expected.

### DIFF
--- a/client/lib/async-requests/index.js
+++ b/client/lib/async-requests/index.js
@@ -39,30 +39,30 @@ export function getRequestByIdString( path, handleData = identity ) {
 }
 
 export const getCategoryLabels = getRequestByIdString(
-	NAMESPACE + 'products/categories',
+	NAMESPACE + '/products/categories',
 	category => ( {
 		id: category.id,
 		label: category.name,
 	} )
 );
 
-export const getCouponLabels = getRequestByIdString( NAMESPACE + 'coupons', coupon => ( {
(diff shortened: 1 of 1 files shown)


Aljullu (approved):
Tested and it fixed both issues. Thanks for taking this @joshuatf!

PR #1464, opened 2019-02-04
Exclude trashed orders from reports

### LINKED ISSUE #1463: Trashed orders are reflected in stats
**Describe the bug**
Trashed orders are still counted towards all stats.

**To Reproduce**

1. Trash any order from wp admin.
2.  Note that the stats are still reflected for that order on the Orders report.

**Expected behavior**
Expected these not to be tracked at all.  We can probably append `trash` to the excluded statuses.

joshuatf:
Fixes #1463

Excludes orders with a trashed status from reports.

### Before
screen shot 2019-02-04 at 4 28 03 pm

### After
screen shot 2019-02-04 at 4 27 20 pm

1.  Check the orders report.
2.  Trash an order within that period.
3.  Check that the order is no longer reflected in report stats.

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-data-store.php
@@ -417,8 +417,8 @@ class WC_Admin_Reports_Data_Store {
 	 * @return array
 	 */
 	protected static function get_excluded_report_order_statuses() {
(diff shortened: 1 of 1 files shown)


Aljullu (approved):
Tested and it works as expected.

Tests are failing but I understand they are fixed in #1466.

claudiosanches (approved):
Good catch!

PR #1468, opened 2019-02-04
Fix hourly labels in bar charts X-axis

### LINKED ISSUE #1413: Incorrect X-Axis label when viewing bar chart graph with hourly intervals
When looking at hourly interval data in bar chart mode, all ticks on the x-axis are labeled 12am.

screen shot 2019-01-29 at 2 18 52 pm

Affecting all reports, and both chart types: standard and comparison

Aljullu:
Fixes #1413.

Fixes the X-axis labels in bar charts when the interval is set to _hours_.

### Screenshots
_Before:_
image

_After:_
image

- Go to the _Orders_ report.
- Select _Today_ in the date picker and set _Hours_ as the interval.
- Verify the correct hour is displayed below each chart.


timmyc (approved):
Working well for me, thanks for adding the CHANGELOG updates.

PR #1472, opened 2019-02-04
navigation: Lazy-create history

### LINKED ISSUE #1433: Activity Panel: Embedded panel appends `#/` to the URL
Embedding the activity panel on existing WooCommerce pages makes those URLs redirect to a version with `#/' appended.

Ie. if you go to `/wp-admin/edit.php?post_type=shop_order` you will be redirected to `/wp-admin/edit.php?post_type=shop_order#/`. Can we disable this?

coderkevin:
Fixes #1433

This will also address the blocks issue: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/327

`history.createHashHistory()` is being called whenever this package is
imported, which causes the URL to redirect to a hash. This is
unnecessary in cases where elements of wc-admin are used outside of a
completely controlled/routed page.

This lazy-creates the history object only when it's needed and therefore
avoids the URL redirection unless history features are in use.

**Compatibility note**
I've suggested a major version change for this to 2.0.0 because it breaks backward compatibility by replacing the `history` export from `@woocommerce/navigation` with a `getHistory()` export. I've updated all references to `history` within this repo, but code outside of this repo will need to adjust to update to this version.

### Accessibility

- [n/a] All animations respect [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion)
- [n/a] All text has [at least a 4.5 color contrast with its background](https://webaim.org/resources/contrastchecker/)

- Navigate through analytics pages to ensure routing works correctly.
- Navigate to `/wp-admin/edit.php?post_type=shop_order` and ensure it is not redirected to `/wp-admin/edit.php?post_type=shop_order#/`

### DIFF
--- a/packages/navigation/CHANGELOG.md
+++ b/packages/navigation/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 2.0.0
+
+- Replace `history` export with `getHistory` (allows for lazy-create of history)
+
 # 1.1.0
 
 - Rename `getTimeRelatedQuery` to `getPersistedQuery`
--- a/packages/navigation/src/history.js
+++ b/packages/navigation/src/history.js
@@ -6,4 +6,13 @@ import { createHashHistory } from 'history';
 
 // See https://github.com/ReactTraining/react-router/blob/master/FAQ.md#how-do-i-access-the-history-object-outside-of-components
 
-export default createHashHistory();
+let _history;
+
+function getHistory() {
+	if ( ! _history ) {
+		_history = createHashHistory();
+	}
+	return _history;
+}
+
+export { getHistory };
--- a/client/dashboard/dashboard-charts/block.js
+++ b/client/dashboard/dashboard-charts/block.js
@@ -10,7 +10,7 @@ import { __, sprintf } from '@wordpress/i18n';
  * WooCommerce dependencies
  */
 import { Card } from '@woocommerce/components';
-import { getAdminLink, history } from '@woocommerce/navigation';
+import { getAdminLink, getHistory } from '@woocommerce/navigation';
 
 /**
  * Internal dependencies
@@ -26,7 +26,7 @@ class ChartBlock extends Component {
 			return null;
 		}
 
-		history.push( 'analytics/' + charts[ 0 ].endpoint + '?chart=' + charts[ 0 ].key );
+		getHistory().push( 'analytics/' + charts[ 0 ].endpoint + '?chart=' + charts[ 0 ].key );
 	};
 
 	render() {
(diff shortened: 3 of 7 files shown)


packages/navigation/src/history.js
```diff
@@ -6,4 +6,13 @@ import { createHashHistory } from 'history';
...
+let _history;
+
+function getHistory() {
+	if ( ! _history ) {
+		_history = createHashHistory();
+	}
+	return _history;
+}
+
+export { getHistory };
```
psealock: Just curious here. Why wrap the export in an new object? When this gets imported `import { getHistory } from './history';` I would think that would break (it doesn't).

Does webpack do something here, or am I just looking at it wrong?

psealock (approved):
Working as expected.

packages/navigation/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 2.0.0
```
joshuatf: Seems like a good call.

joshuatf (commented):
Nice work on this @coderkevin!

Everything continues to work as expected and this does clear up the hashtag on `/wp-admin/edit.php?post_type=shop_order` but I noticed if we hit individual orders we still get the hashtag appended (e.g., `/wp-admin/post.php?post=283796&action=edit`).  Is this correct?

If that route doesn't need to be controlled by wc-admin, it would be nice to also remove the hashtag there as it has caused some bugs to pop up previously.

packages/navigation/src/history.js
coderkevin: Since we use `getHistory` as an import and also export it, I used the named export notation here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export#Using_named_exports

coderkevin:
> I noticed if we hit individual orders we still get the hashtag appended (e.g., `/wp-admin/post.php?post=283796&action=edit`). Is this correct?

If that's the case, then something is calling `getHistory` on that page. I'd say if that's not the behavior we want, then we should log a bug, investigate and see if the `getHistory` call can be removed. What are your thoughts on that?

coderkevin:
> I noticed if we hit individual orders we still get the hashtag appended (e.g., `/wp-admin/post.php?post=283796&action=edit`). Is this correct?

I looked into this a bit more, and it's being triggered by the breadcrumbs header here: https://github.com/woocommerce/wc-admin/blob/master/client/header/index.js#L92

This probably merits a different discussion about the breadcrumbs header. I've created #1480 to track it.

joshuatf (approved):
Thanks for creating that issue around the breadcrumbs @coderkevin!  This one looks good to go then.  :shipit:

PR #1479, opened 2019-02-05
Guest customers are no longer treated always as new customers.

### LINKED ISSUE #1476: Returning customer definition for guest customers
**Describe the bug**
It seems to me that guests are currently always counted as new customers even after 1st order.

**To Reproduce**

1. Create an order with a particular email, e.g. `guest1@test.test`.
2. Create another order with the same email as in 1.
3. Check `wp_wc_order_stats` to see that `customer_id` is the same, but `returning_customer` is set to 0 for both orders. 

**Expected behavior**
I would expect the second order to have `returning_customer` set to 1.

This is caused by fetching the `customer_id` via `WC_Admin_Reports_Customers_Data_Store::get_customer_id_by_user_id`. This means that guest with user_id = 0 has no `customer_id` and the subsequent code never checks if there is any previous order for the customer, which I think is not correct?

cc'ing @joshuatf @jeffstieler as you guys maybe have better info on the background of these decisions, also related to #1180.

peterfabian:
Fixes #1476 (if it's actually a bug)

Currently, all guest customers are treated as new customers, even if they already ordered with the same email address. 
This PR changes it so that guests can also be viewed as returning customers.

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-customers-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-customers-data-store.php
@@ -392,6 +392,34 @@ class WC_Admin_Reports_Customers_Data_Store extends WC_Admin_Reports_Data_Store 
 	}
 
 	/**
+	 * Gets the customer ID for given order, either from user_id, if the order was done
+	 * by a registered customer, or from the billing email in the provided WC_Order.
+	 *
+	 * @param WC_Order $order Order to get customer data from.
+	 * @return int|false The ID of the retrieved customer, or false when there is no such customer.
+	 */
+	public static function get_customer_id_from_order( $order ) {
+		$user_id = $order->get_user_id();
+		if ( $user_id ) {
+			return self::get_customer_id_by_user_id( $user_id );
+		}
+
+		$email = $order->get_billing_email( 'edit' );
+
+		if ( empty( $email ) ) {
+			return false;
+		}
+
+		$guest_customer_id = self::get_customer_id_by_email( $email );
+
+		if ( $guest_customer_id ) {
+			return $guest_customer_id;
+		}
+
+		return false;
+	}
+
+	/**
 	 * Gets the guest (no user_id) customer ID or creates a new one for
 	 * the corresponding billing email in the provided WC_Order
 	 *
@@ -502,6 +530,26 @@ class WC_Admin_Reports_Customers_Data_Store extends WC_Admin_Reports_Data_Store 
 			$wpdb->prepare(
 				"SELECT customer_id FROM {$table_name} WHERE user_id = %d LIMIT 1",
 				$user_id
+			)
+		); // WPCS: unprepared SQL ok.
+
+		return $customer_id ? (int) $customer_id : false;
+	}
+
+	/**
+	 * Retrieve a customer id by billing email.
+	 *
+	 * @param string $email Billing email for the user.
+	 * @return false|int Customer ID if found, boolean false if not.
+	 */
+	public static function get_customer_id_by_email( $email ) {
+		global $wpdb;
+
+		$table_name  = $wpdb->prefix . self::TABLE_NAME;
+		$customer_id = $wpdb->get_var(
+			$wpdb->prepare(
+				"SELECT customer_id FROM {$table_name} WHERE email = %s AND user_id IS NULL LIMIT 1",
+				$email
 			)
 		); // WPCS: unprepared SQL ok.
 
--- a/tests/reports/class-wc-tests-reports-orders-stats.php
+++ b/tests/reports/class-wc-tests-reports-orders-stats.php
@@ -3287,7 +3287,7 @@ class WC_Tests_Reports_Orders_Stats extends WC_Unit_Test_Case {
 		$customer_1 = WC_Helper_Customer::create_customer( 'cust_1', 'pwd_1', 'user_1@mail.com' );
 
 		$order_1_time = time();
-		$order_3_time = $order_1_time - 1 * HOUR_IN_SECONDS;
+		$order_3_time = $order_1_time - 1 * HOUR_IN_SECONDS; // @todo: I wonder if this messes up new/returning customer
 
 		// Order 3: 4 x product 1, done one hour earlier.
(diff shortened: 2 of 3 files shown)


includes/data-stores/class-wc-admin-reports-customers-data-store.php
```diff
@@ -391,6 +391,34 @@ public function get_data( $query_args ) {
...
+	 * @param WC_Order $order Order to get customer data from.
+	 * @return int|false The ID of the retrieved customer, or false when there is no such customer.
+	 */
+	public static function get_customer_id_from_order( $order ) {
+		$user_id = $order->get_user_id();
+		if ( $user_id ) {
+			return self::get_customer_id_by_user_id( $user_id );
+		}
+
+		$email = $order->get_billing_email( 'edit' );
```
joshuatf: If we go with the above refactor, I think we can drop the rest of this function and return the following:

`return get_or_create_guest_customer_from_order( $order );`

includes/data-stores/class-wc-admin-reports-customers-data-store.php
```diff
@@ -508,6 +536,26 @@ public static function get_customer_id_by_user_id( $user_id ) {
...
+	 * @param string $email Billing email for the user.
+	 * @return false|int Customer ID if found, boolean false if not.
+	 */
+	public static function get_customer_id_by_email( $email ) {
+		global $wpdb;
+
+		$table_name  = $wpdb->prefix . self::TABLE_NAME;
+		$customer_id = $wpdb->get_var(
+			$wpdb->prepare(
+				"SELECT customer_id FROM {$table_name} WHERE email = %s AND user_id IS NULL LIMIT 1",
```
joshuatf: It looks like this is only searching for customers without a `user_id` (guests only).  Is there a difference between this function and `get_guest_by_email()`?

I think we can remove it if we go with the above refactor, but if I'm wrong we might want to rename this to something like `get_guest_customer_id_by_email()`.

tests/reports/class-wc-tests-reports-orders-stats.php
```diff
@@ -3287,7 +3287,7 @@ public function test_segmenting_by_product_and_variation() {
 		$customer_1 = WC_Helper_Customer::create_customer( 'cust_1', 'pwd_1', 'user_1@mail.com' );
 
 		$order_1_time = time();
-		$order_3_time = $order_1_time - 1 * HOUR_IN_SECONDS;
+		$order_3_time = $order_1_time - 1 * HOUR_IN_SECONDS; // @todo: I wonder if this messes up new/returning customer
```
joshuatf: Very good question.  It's okay right now, but there are 2 potential places where this would create issues:

1.  In [this issue](https://github.com/woocommerce/wc-admin/issues/1490#issuecomment-460917176), I proposed a potential solution that tests against IDs as well as dates.  Since `$order_3` would have a larger ID, this would fail.  

I think the solution I proposed there is a bad one though, so this is probably a non-issue.

2.  In this case, because `$order_1` is created first, but `$order_3` has an earlier date, `$order_1` should be marked as returning.  I touched on this at the end of [this issue](https://github.com/woocommerce/wc-admin/issues/1473):

>Furthermore, we would also need to check all orders by a customer and update if a status is changed on a previous order.

If an order date is updated, we would need to update all their orders again with the correct returning value if the order date is earlier than all previous orders.

includes/data-stores/class-wc-admin-reports-customers-data-store.php
peterfabian: I agree it's not the nicest solution. I did what you say initially, but then I noticed we need a static method, so I created the new fn (`get_guest_by_email` casts numbers based on member var `$column_types` , thus can't be static).

includes/data-stores/class-wc-admin-reports-customers-data-store.php
```diff
@@ -391,6 +391,34 @@ public function get_data( $query_args ) {
...
 	}
 
+	/**
+	 * Gets the customer ID for given order, either from user_id, if the order was done
+	 * by a registered customer, or from the billing email in the provided WC_Order.
+	 *
+	 * @param WC_Order $order Order to get customer data from.
+	 * @return int|false The ID of the retrieved customer, or false when there is no such customer.
+	 */
+	public static function get_customer_id_from_order( $order ) {
```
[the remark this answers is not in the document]
peterfabian: Originally, I created this function to have a possibility of getting the `customer_id` without creating it, if it's missing. 

However, thinking about it now in light of p90Yrv-XN-p2, I think we might be getting rid of `returning_customer` column, which means we probably don't need the `get_customer_id_from_order` function.

peterfabian:
I think I'm putting this on hold until we decide on a way fwd in p90Yrv-XN-p2.

tests/reports/class-wc-tests-reports-orders-stats.php
joshuatf: This should be a non-issue after https://github.com/woocommerce/wc-admin/pull/1517 is merged in.

includes/data-stores/class-wc-admin-reports-customers-data-store.php
joshuatf: Maybe we could remove the casting from `get_guest_by_email()` then since only the customer ID is needed anyway.

includes/data-stores/class-wc-admin-reports-customers-data-store.php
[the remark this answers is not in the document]
joshuatf: @peterfabian Based on what I discovered in opening both PRs related to p90Yrv-XN-p2 we might be keeping `returning_customer` after all since my other method was less performant.

Let me know if you disagree with my assessment, but I think this PR is still a good one to get merged otherwise.

peterfabian:
Thanks for the updated comments. I will work to sort these out tomorrow.

joshuatf:
@peterfabian I made some changes to the customers data store in https://github.com/woocommerce/wc-admin/pull/1619 to work on #1552.

I think that one fixes the guest issue, but I'm thinking we should keep the tests in this PR even if we go with the refactor made in #1619.

peterfabian:
Closing in favour of #1681, as this one is now outdated.

PR #1481, opened 2019-02-05
Fix product comparison + color scale on charts with lots of options

### LINKED ISSUE #1401: Fix color scale on charts with lots of options
When implementing #1391, I noticed the chart comparison can return chart colors are not always different enough in graph values to tell them apart.

screen recording 2019-01-25 at 08 57 am

David wrote (https://github.com/woocommerce/wc-admin/pull/1391#pullrequestreview-196615472):

> Regarding the color scale you mentioned as a know issue... I'm pretty sure I know what's going on here... Back in the day, Rob and I specified colors for charts with five stops or less:

> #375

> When there are five or more series, we let math takeover, and we create as many stops as we need... so what I assume is happening, is that when you compare a huge number of things, the color scale is based on the total number of compared items, rather than the specific number of things we allow people to compare in the chart at once (in this case, five). You can see that happening in the gif above - all the colors are deep purple / blue, on the far end of the scale.

emdashcodes:
Fixes #1401.

This PR fixes an issue with the chart item colors when > 5 items are loaded into a chart. Previously, the color scale would be based on all possible options. Instead, the scale is now based on visible elements only. It also moves the selection limit and color values to a constants file, so these can easily be adjusted later.

I've also fixed an issue with the product comparison mode that I made while merging/rebasing #1391.

### Screenshots

screen shot 2019-02-05 at 3 28 12 pm

* Open a chart with many many options (like a single variation), or a comparison chart with lots of product comparisons selected.
* Verify color for selected elements contain greens, etc and not just purple elements.

### DIFF
--- a/packages/components/src/chart/constants.js
+++ b/packages/components/src/chart/constants.js
@@ -1,0 +1,13 @@
+/** @format */
+
+// This is the max number of items that can be selected/shown on a chart at one time.
+// If this number changes, the color scale also needs to be adjusted.
+export const selectionLimit = 5;
+export const colorScales = [
+	[],
+	[ 0.5 ],
+	[ 0.333, 0.667 ],
+	[ 0.25, 0.50, 0.75 ],
+	[ 0.20, 0.40, 0.60, 0.80 ],
+	[ 0.16, 0.32, 0.48, 0.64, 0.80 ],
+];
--- a/client/analytics/report/products/index.js
+++ b/client/analytics/report/products/index.js
@@ -31,7 +31,7 @@ class ProductsReport extends Component {
 		const isProductDetailsView =
 			'top_items' === query.filter ||
 			'top_sales' === query.filter ||
-			'single_product' === query.filter;
+			'compare-products' === query.filter;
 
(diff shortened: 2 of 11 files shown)


packages/components/src/chart/constants.js
```diff
@@ -0,0 +1,13 @@
+/** @format */
+
+// This is the max number of items that can be selected/shown on a chart at one time.
+// If this number changes, the color scale also needs to be adjusted.
+export const selectionLimit = 5;
```
joshuatf: Nice move separating these!

joshuatf (approved):
This is testing really well for me!

One minor oddity I noticed was the selected items were sometimes adjacent and sometimes not upon refresh.  But I think this can be addressed in a separate issue if it's even a problem.
screen shot 2019-02-06 at 4 52 10 pm
screen shot 2019-02-06 at 4 52 33 pm

I also managed to get all purple one time but can't seem to replicate so I think it's okay.

screen shot 2019-02-06 at 4 58 55 pm

PR #1491, opened 2019-02-06
 Show empty chart and summary numbers when search returns no results 

### LINKED ISSUE #1451: Table search: Don't try to fill the report if we know there are no results
When doing a search with the table input, a query might return no results. In that case, we want to avoid sending API calls that we know will return no results or 404.

This issue consists on three parts:
- [x] Table #1452.
- [ ] Chart #1491.
- [ ] Summary numbers #1491.

Aljullu:
Fixes #1451.

Follow-up of #1452.

Show empty chart and summary numbers when a search returns no results and avoid doing useless API calls.

### Accessibility

### Screenshots
image

* Make a search that will return no results.
* Verify the chart and summary numbers are empty.
* In your browser devtools _Network_ tab, verify there is no call sent to `/wc/v4/reports/products/stats`.


client/analytics/components/report-summary/index.js
```diff
@@ -27,9 +27,27 @@ import withSelect from 'wc-api/with-select';
  * Component to render summary numbers in reports.
  */
 export class ReportSummary extends Component {
+	formatVal( val, type ) {
+		return 'currency' === type ? formatCurrency( val ) : formatValue( type, val );
```
joshuatf: What do you think about adding this as a type directly to `formatValue()` [here](https://github.com/woocommerce/wc-admin/blob/847e9788f4f8581de6016dce39e023efe5d6bd93/packages/number/src/index.js#L49)?

Not sure if there are issues if we do this with interdependencies between packages since we'd need to import `formatCurrency()`.

client/analytics/components/report-summary/index.js
```diff
@@ -39,23 +57,14 @@ export class ReportSummary extends Component {
...
-				const delta = calculateDelta( primaryTotals[ key ], secondaryTotals[ key ] );
 				const href = getNewPath( { chart: key } );
-				const prevValue = isCurrency
-					? formatCurrency( secondaryTotals[ key ] )
-					: formatValue( type, secondaryTotals[ key ] );
 				const isSelected = selectedChart.key === key;
-				const value = isCurrency
-					? formatCurrency( primaryTotals[ key ] )
-					: formatValue( type, primaryTotals[ key ] );
+				const { delta, prevValue, value } = this.getValues( key, type );
```
joshuatf: Nice. Much more readable here.

client/analytics/components/report-summary/index.js
```diff
@@ -105,11 +114,31 @@ ReportSummary.propTypes = {
...
+		isRequesting: false,
+	},
 };
 
 export default compose(
 	withSelect( ( select, props ) => {
 		const { query, endpoint } = props;
+		if ( query.search && ! ( query[ endpoint ] && query[ endpoint ].length ) ) {
+			return {
+				emptySearch: true,
```
joshuatf: Nitpicking here, but could we rename this to `emptyResults` or ``emptySearchResults`?  Someone else glancing at this might get confused and think this prop means no search query was made.

joshuatf (commented):
Awesome work @Aljullu.  Only the search result query is made on this page now.

Left a comment about possibly moving the currency formatting into the numbers package and another super nitpicky comment about naming, but this is testing well!

client/analytics/components/report-summary/index.js
Aljullu: Yes, that was my main concern as well, I didn't want to add `@woocommerce/currency` as a dependency in `@woocommerce/number`.

Looking around, I found that was discussed [here](https://github.com/woocommerce/wc-admin/pull/1399/commits/ce4729f3853a0bccd61085e2f72990b5e82bf2c4#r251197977) and the decision was not to add `type=currency` in `formatValue`. So I would leave it as it is now. How does it sound?

client/analytics/components/report-summary/index.js
Aljullu: Good point, makes a lot of sense!

Fixed in 21c1490d47028d06a15a101c401c89bb21fcdf52.

Aljullu:
@joshuatf thanks for the review! Feel free to take another look when you have a second!

_(Travis build is failing because of #1498)_

client/analytics/components/report-summary/index.js
joshuatf: Sweet- thanks!

client/analytics/components/report-summary/index.js
joshuatf: Good find on that discussion!  Ya, you definitely made the right call by keeping it separate.

joshuatf (approved):
Thanks for making that change and for posting the circular dependency discussion link!  This one looks good to merge after a rebase (#1498 is approved now).

PR #1497, opened 2019-02-07
Fix initial order sync stalling

jeffstieler:
_Warning: 🐇🕳_

This issue was introduced (in part) by #1426.

### The Problem(s)

The logic introduced in #1426 to determine order processing success (and whether or not a retry should be attempted) relies on a success indicator returned from each sync class/method. This proved to be problematic:

* `shop_order_refund`-type posts were skipped by the order stats and product sync methods, but not the coupon and tax sync methods (causing a non +/- 4 result)
* The success logic in `WC_Admin_Reports_Products_Data_Store::sync_order_products()` was faulty - on initial syncs, refunded products were attempted to be deleted from the lookup table but since they hadn't been inserted yet, 0 rows were affected by the query and a non-success code was returned

Additionally:

* Occasionally the "next run" of a job considered to be blocking a "dependent action" was NULL (https://github.com/woocommerce/wc-admin/pull/1426#issuecomment-459696449)
* The blocking job could be a false positive where the scheduled hook was actually "queue dependent action" and not "queue batch action"

The best test data is the WCCOM development environment. Start it fresh and rebuild reports - make sure that all scheduled actions complete. (My environment stalled at 197 pending jobs on two separate attempts before this fix)

### DIFF
--- a/includes/class-wc-admin-api-init.php
+++ b/includes/class-wc-admin-api-init.php
@@ -672,12 +672,24 @@ class WC_Admin_Api_Init {
 			)
 		);
 
-		if ( $blocking_jobs ) {
-			$blocking_job       = current( $blocking_jobs );
-			$after_blocking_job = $blocking_job->get_schedule()->next()->getTimestamp() + 5;
+		$next_job_schedule = null;
+		$blocking_job_hook = null;
 
+		if ( $blocking_jobs ) {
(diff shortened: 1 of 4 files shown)


timmyc (approved):
Couple of comments about comments, but this tested will for me agains the wccom developer data set. My macbook only levitated 3 times while the jobs were running!

PR #1499, opened 2019-02-07
Empty chart message

### LINKED ISSUE #582: Chart with empty data set

Aljullu:
Fixes #582.

Updates the design of empty charts according to https://github.com/woocommerce/wc-admin/issues/582#issuecomment-459209125.

A couple of notes:
- I changed the color from `light-grey-700` to `grey-dark-300` in order to comply with the minimum contrast ratio.
- In narrow viewports (mobile) I reduced the font size of the message, otherwise it was overlapping the chart y-axis lines.

### Accessibility

### Screenshots
image
image

- Go to the _Products_ report and select a day with no orders in the day picker.
- Verify a message `No data for the selected date range` appears in the chart.
- Reset the date selection to include some results and verify the message disappears.
- In the table, search for a product that doesn't exist, for example introduce `adfads` in the table search input.
- Verify the chart displays a message `No data for the current search`.

### DIFF
--- a/client/analytics/components/report-chart/index.js
+++ b/client/analytics/components/report-chart/index.js
@@ -218,10 +218,13 @@ ReportChart.defaultProps = {
 export default compose(
 	withSelect( ( select, props ) => {
 		const { query, endpoint, filters } = props;
-		if ( query.search && ! ( query[ endpoint ] && query[ endpoint ].length ) ) {
-			return {};
-		}
 		const chartMode = props.mode || getChartMode( filters, query ) || 'time-comparison';
+
+		if ( query.search && ! ( query[ endpoint ] && query[ endpoint ].length ) ) {
+			return {
+				mode: chartMode,
+			};
+		}
 
 		if ( query.search && ! ( query[ endpoint ] && query[ endpoint ].length ) ) {
 			return {
--- a/client/analytics/components/report-chart/index.js
+++ b/client/analytics/components/report-chart/index.js
@@ -2,6 +2,7 @@
 /**
  * External dependencies
  */
+import { __ } from '@wordpress/i18n';
 import { Component } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import { format as formatDate } from '@wordpress/date';
@@ -89,6 +90,7 @@ export class ReportChart extends Component {
 
 	renderChart( mode, isRequesting, chartData ) {
 		const {
+			emptySearch,
 			interactiveLegend,
 			itemsLabel,
 			legendPosition,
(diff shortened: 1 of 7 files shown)


client/analytics/components/report-chart/index.js
```diff
@@ -222,6 +228,7 @@ export default compose(
 
 		if ( query.search && ! ( query[ endpoint ] && query[ endpoint ].length ) ) {
```
psealock: Check if the value is an array

```suggestion
		if ( query.search && ! ( Array.isArray( query[ endpoint ] ) && query[ endpoint ].length ) ) {
```

packages/components/src/chart/d3chart/chart.js
```diff
@@ -124,6 +128,11 @@ class D3Chart extends Component {
 		const xLineScale = getXLineScale( uniqueDates, adjWidth );
 		const xScale = getXScale( uniqueDates, adjWidth, compact );
 		const xTicks = getXTicks( uniqueDates, adjWidth, mode, interval );
+
+		this.setState( {
+			isEmpty: yMax === 0,
```
psealock: I know this is likely to catch all instances for our purposes, but since we are releasing `<Chart />` as a package, it could be used for display of entirely negative numbers. Is there another check that may be more explicit?

psealock (commented):
Nice, working as expected. Just a small suggestion and a comment about how `isEmpty` is derived.

client/analytics/components/report-chart/index.js
joshuatf: I think this is always returned as a comma delimited string, but if I'm wrong we'll also need to update this in all other areas.

joshuatf (approved):
Looks good and is testing well.  After addressing @psealock's question about a negative `yMax` I think this is good to merge.

timmyc:
Tested this out and it is looking waaaay better then before

_Before_
zero-bar-chart

_After_
image

Nice work @Aljullu

client/analytics/components/report-chart/index.js
Aljullu: Right, like @joshuatf said, `query[ endpoint ]` is a comma-separated string.

packages/components/src/chart/d3chart/chart.js
Aljullu: Good point. Currently the `<Chart />` component is not prepared to display negative values, but if at some point we want to support it, it makes sense to implement new features with that in mind.

I refactored completely the way we check whether the chart is empty in c24bbba226b2112f859fcea98afb9e31b9040e3f. This way, I could also remove the state from the component.

Aljullu:
Thanks for the review @psealock @joshuatf @timmyc!

@psealock, would you like to take another look?

psealock (approved):
LGTM 🚢

PR #1500, opened 2019-02-07
Add Feature Flags

### LINKED ISSUE #1345: Add Feature Flags
Since we are punting the activity panels feature until after 3.6, we need the concept of feature flags to disable the feature, but keep it around for the future. This also comes in handy so we can continue to use `wc-admin` to develop/build updates for core in the future.

I think we should support multiple environments (production, dev). The flags should also be available via PHP if possible (the activity panel contains PHP code as well).

Some prior art/related discussions:

* https://github.com/WordPress/gutenberg/issues/11016
* https://github.com/WordPress/gutenberg/issues/11016#issuecomment-447210849
* https://github.com/Automattic/wp-calypso/tree/master/config#feature-flags
* https://github.com/WordPress/gutenberg/pull/13324

emdashcodes:
Closes #1345.

This PR adds feature flags to `wc-admin 🎉.

This allows us to enable/disable features depending on the enviornment. Right now, the environments this supports are `development`, `plugin`, and `core`. Think of `development`, `staging`, and `production` in Calypso for example. You can access the flags from both PHP and JS.  The approach I used on the JS side also allows Webpack to do dead code elimination [like Gutenberg](https://github.com/WordPress/gutenberg/blob/bbdd7a70389652f483a64f66b078f992fc62f917/docs/designers-developers/developers/feature-flags.md#dead-code-elimination). For more information on all this, [view the documentation that is a part of this PR](https://github.com/woocommerce/wc-admin/blob/f129156463c5359bf086c628ecca07270565cea3/docs/feature-flags.md).

I have based this approach off of a mix of the proposed Gutenberg PR (https://github.com/WordPress/gutenberg/pull/13324) and Calypso's implementation. I prefer checking against flags in the code, instead of a specific "phase" of the plugin.

Some notes/thoughts for the future:

* How should this effect documentation (if it at all)? I don't think we will be using these flags in packages (packages have versioning) so I don't think we need to do any pruning of documents just yet... but if we start documenting a feature, we might want to prune that out for "plugin documentation" or "core documentation".
* API conditional file loading. I haven't changed any of the API loading in this PR. Some endpoints are used across features. Since these endpoints are also versioned from core, it seems like we can leave loading alone too.
* Should we have two separate `plugin` environments? I imagine at some point, we may be making changes to a released WordPress.org feature plugin, and at the same time might be cutting a GitHub preview release. Should we think about adding a `beta` environment as well?
* I added a `core.json` config file, but our core merge process is TBD so this is more of a reminder.

### Screenshots

Screenshot of devdocs and activity panel disabled in the bundled `plugin`:

screen shot 2019-02-08 at 1 36 33 pm

* Review the documentation. Does it match the usage in this PR?
* Run `npm start` (close any current running build, since the commands have changed)
* Click around various parts of `wc-admin` and WooCommerce pages. Functionally, in the development environment, nothing should have changed.
* Run `npm run-script build:release` and build a `wc-admin.zip` plugin. Upload this to a different test site (wpsandbox for example). Enable wc-admin.
* Verify that the activity panel does not display.
* Verify that the devdocs does not display.
* Verify that you can see the dashboard and analytics pages.


joshuatf (commented):
So much 🔥 🔥  in this PR.  Awesome work, Justin!

Left a few comments in the code, but overall this looks great.

>How should this effect documentation (if it at all)?

I agree probably not at all for the time being.  I could be wrong, but I don't think that flags should affect packages in any way.

>API conditional file loading. I haven't changed any of the API loading in this PR. Some endpoints are used across features. Since these endpoints are also versioned from core, it seems like we can leave loading alone too.

I think it's *probably* okay as is right now, but I left a comment about a possible issue.  We override core endpoints if they use the same namespace we could cause issue if they ever match (but currently we're all `v4` and core is using `v3`).

>Should we have two separate `plugin` environments?

I like the idea of having a `beta` environment as well.  Seems like something we'll use.

client/layout/controller.js
```diff
@@ -23,42 +23,48 @@ import DevDocs from 'devdocs';
 const getPages = () => {
 	const pages = [
 		{
-			container: Dashboard,
-			path: '/',
+			container: DevDocs,
```
timmyc: Might be nice to put DevDocs behind a feature flag that is only present in `development` mode.

docs/feature-flags.md
```diff
@@ -0,0 +1,47 @@
...
+| Environment | Description                                                                                                                                                            |
+|-------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| development | Development - All features should be enabled in development. These flags are also used in both JS and PHP tests. Ran using `npm start`.                                |
+| plugin      | Plugin - A packaged release of the featured plugin, for GitHub WordPress.org. Ran using `npm run-script build:release`. |                                    |
+| core        | Core - assets/files ready and stable enough for core merge. @todo No build process exists yet.
+
+
+## Adding a new flag
+
+Flags can be added to the files located in the `config/` directory. Make sure to add a flag for each environment and explicitly set the flag to false.
```
timmyc: I think the Calypso feature flags allow you to omit a flag from a config, and if it is not present it is considered that the feature is not enabled for that env. I believe initially you had to be explicit in adding a flag in all configs like it is setup here but changed during a re-write.

I'm fine with making this a requirement myself, but just figured I'd mention the path that happened on Calypso for reference.

docs/feature-flags.md
```diff
@@ -0,0 +1,47 @@
...
+Please add new feature flags alphabetically so they are easy to find.
```
timmyc: 🙌

docs/feature-flags.md
```diff
@@ -0,0 +1,47 @@
...
+To check if a feature is enabled, you can simplify check the boolean value of a feature:
+
+```
+{ window.wcAdminFeatures[ 'activity-panels' ] && <ActivityPanel /> }
+```
+
+We also expose CSS classes on the `body` tag, so that you can target specific feature states:
+
+```
+<body class="wp-admin woocommerce-page woocommerce-feature-disabled-analytics woocommerce-feature-enabled-activity-panels  ....">
```
timmyc: Nice!

docs/feature-flags.md
```diff
@@ -0,0 +1,47 @@
...
+## Basic Use - Client
+
+The `window.wcAdminFeatures` constant is a global variable containing the feature flags.
+
+Instances of `window.wcAdminFeatures` are replaced during the webpack build process by using webpack's [define plugin](https://webpack.js.org/plugins/define-plugin/). Using `webpack` for this allows us to eliminate dead code when making minified/production builds (`plugin`, or `core` environments).
+
+To check if a feature is enabled, you can simplify check the boolean value of a feature:
+
+```
+{ window.wcAdminFeatures[ 'activity-panels' ] && <ActivityPanel /> }
```
timmyc: I like the simplicity of just checking the global - but wondering if you considered adding a util method like `isFeatureEnabled` or something along those lines. Probably would result in more code throughout the app, but the only benefit I see there is being able to update the logic in one file vs dozens if/when we need to change it for some reason.

docs/feature-flags.md
```diff
@@ -0,0 +1,47 @@
...
+```
+
+## Basic Use - Server
+
+Feature flags are also available via PHP. To ensure these are consistent with the built client assets, `includes/feature-flags.php` is generated by the plugin build process or `npm start`. Do not edit `includes/feature-flags.php` directly.
+
+To check if a feature is enabled, you can use the `wc_admin_is_feature_enabled()`:
+
+```
+if ( wc_admin_is_feature_enabled( 'activity-panels' ) ) {
```
timmyc: Building upon the comment above, we could even make the method name the same in JS `wcAdminIsFeatureEnabled()`

lib/admin.php
```diff
@@ -236,7 +246,7 @@ function wc_admin_enqueue_script() {
 add_action( 'admin_enqueue_scripts', 'wc_admin_enqueue_script' );
 
 /**
- * Adds an admin body class.
+ * Adds body classes to the main wp-admin wrapper, allowing us to better target elements in specific scenarios.
```
timmyc: Really, really like this. This is something I wish was present in calypsoify too.

package.json
```diff
@@ -17,12 +17,13 @@
     "build:packages": "node ./bin/packages/build.js",
     "build:core": "cross-env NODE_ENV=production webpack",
     "build": "npm run build:packages && npm run build:core",
-    "build:release": "./bin/build-plugin-zip.sh",
+    "build:release": "cross-env WC_ADMIN_PHASE=plugin ./bin/build-plugin-zip.sh",
```
timmyc: Nice!

timmyc (commented):
This is looking great Justin - really think this will be a great addition in core as well to allow us to actively develop JS functionality there ( thinking even around blocks ) and keep things feature flagged. Few comments, but will take this for a proper test tomorrow morning.

client/layout/controller.js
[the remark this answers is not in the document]
timmyc: Jinx! Or are you reading my mind Josh!?!

timmyc:
Tagging @ryelle for any input here - I could see this being pretty useful in Woo core too for blocks development once everything lands in core.

config/plugin.json
```diff
@@ -0,0 +1,8 @@
+{
+	"features": {
+		"activity-panels": false,
+		"analytics": true,
+		"analytics/categories": false,
```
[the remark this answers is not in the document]
justinshreve: No discussion around this. I think the plan is still to have a categories report -- I just wanted to provide an example here for a "sub-feature", since the category report is still using Swagger. I can remove this flag before merge.

docs/feature-flags.md
justinshreve: @timmyc I definitely considered a utility method here but couldn't quite get it to work 100%. I just spent a bit more time confirming this.

The trade off here basically boils down to webpack's code elimination.

Right now, with this branch, if webpack encountered this line and the flag was turned off, this code would not end up in the plugin/production bundle. I tried a few different methods to try to get it to phase a function, but the code ends up being bundled (even if not displayed to the user).

One potential solution (also raised in the Gutenberg thread -- which I think they decided against) is to write a custom babel plugin that could transform these function calls into the global check before webpack hits it.

I'm personally fine just checking a variable. The extra work would basically be for different syntax since these flags should ever only be boolean values (so the function logic should never really change anyway). What do you think?

docs/feature-flags.md
justinshreve: See https://github.com/woocommerce/wc-admin/pull/1500#discussion_r255174276

docs/feature-flags.md
justinshreve: I just checked, and you actually can omit flags and still have it default to the feature being off (undefined).  So nothing will break if we forget to do this.

However, I would say let's keep the documentation line here. I like us having these be explicit in each env so you can quickly open the file and see what's up.

client/layout/controller.js
justinshreve: Done in https://github.com/woocommerce/wc-admin/pull/1500/commits/81607dc415c90a146867a373fdf157b892ff2631.

bin/generate-feature-config.php
```diff
@@ -0,0 +1,27 @@
...
+
+$phase = isset( $_SERVER['WC_ADMIN_PHASE'] ) ? $_SERVER['WC_ADMIN_PHASE'] : ''; // WPCS: sanitization ok.
+if ( empty( $phase ) || ! in_array( $phase, array( 'development', 'plugin', 'core' ) ) ) {
+	$phase = 'core';
+}
+$config_json = file_get_contents( 'config/' . $phase . '.json' );
+$config      = json_decode( $config_json );
+
+$write  = "<?php\n";
+$write .= "// This file is auto-generated as part of the build process. Changing values here without the correct assets may break things.\n";
```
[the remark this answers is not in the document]
justinshreve: Done in https://github.com/woocommerce/wc-admin/pull/1500/commits/81607dc415c90a146867a373fdf157b892ff2631.

config/plugin.json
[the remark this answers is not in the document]
justinshreve: Removed the flag here since we won't end up needing it: https://github.com/woocommerce/wc-admin/pull/1500/commits/81607dc415c90a146867a373fdf157b892ff2631.

client/layout/controller.js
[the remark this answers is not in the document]
justinshreve: 😆.

Done in https://github.com/woocommerce/wc-admin/pull/1500/commits/81607dc415c90a146867a373fdf157b892ff2631.

lib/admin.php
```diff
@@ -66,104 +70,110 @@ function wc_admin_register_page( $options ) {
...
-	add_menu_page(
-		__( 'WooCommerce Analytics', 'wc-admin' ),
-		__( 'Analytics', 'wc-admin' ),
-		'manage_options',
-		'wc-admin#/analytics/revenue',
-		'wc_admin_page',
-		'dashicons-chart-bar',
-		56 // After WooCommerce & Product menu items.
-	);
+	if ( wc_admin_is_feature_enabled( 'analytics' ) ) {
```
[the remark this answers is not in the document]
justinshreve: > I'm thinking it's mostly okay

I think so too, and wrapping the endpoints in different checks just might be confusing/more trouble than it is worth. For example, dashboard relies on the same endpoints as analytics, and I think there is some overlap between the activity panel too (adding additional fields to a response for example).

> but if they're disabling because of a conflict in WC core where our API files override, this might cause an issue.

Since these are versioned from the core endpoints I think users shouldn't notice an issue, unless they are testing `plugin`/`development`, (i.e these are v4 and anything "real" will use v3). As long as we keep bumping the `namespace` in `wc-admin` whenever we do a big core merge we should be OK -- So when core hits v4, we should use v5.

justinshreve:
> I like the idea of having a `beta` environment as well. Seems like something we'll use.

I added a `@todo` for this to webpack here: https://github.com/woocommerce/wc-admin/pull/1500/commits/e106aaae653e4f4759e094e22a284dc6e249bb8f

We don't currently have a separate build process for a beta version, but when we do we can easily add a new phase and pass it to the build script.

justinshreve:
@timmyc @joshuatf Thank you both for the feedback here! I've addressed some comments, and answered others above.

tests/js/setup-globals.js
```diff
@@ -64,4 +64,10 @@ global.wcSettings = {
 	},
 };
 
+const config = require( '../../config/development.json' );
+window.wcAdminFeatures = {};
+config && config.features && Object.keys( config.features ).forEach( ( key ) => {
+  window.wcAdminFeatures[ key ] = config.features[ key ];
+} );
```
Aljullu: Is there any benefit on assigning each property one by one instead of assigning `config.features` to `wcAdminFeatures` directly?
```ES6
window.wcAdminFeatures = config && config.features ? config.features : {};
```

webpack.config.js
```diff
@@ -156,6 +168,7 @@ const webpackConfig = {
 				transform: content => content,
 			} ) )
 		),
+
```
Aljullu: There is no need for this empty line, right? :upside_down_face: 
```suggestion
```

Aljullu (commented):
This is amazing @justinshreve, good job! :tada: 

I added a couple of questions regarding the code, but overall it looks awesome.

package.json
```diff
@@ -17,12 +17,13 @@
...
+    "build:php": "php bin/generate-feature-config.php",
```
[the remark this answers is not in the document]
justinshreve: I originally went with `build:php` thinking we might add something else in the future, but I agree with you -- since this doesn't really build PHP files/parts of the app and just config, I think `build:feature-config` sounds good.

webpack.config.js
justinshreve: Thanks, fixed in https://github.com/woocommerce/wc-admin/pull/1500/commits/45badff93d44a9920192796be5875bab131c0a93

tests/js/setup-globals.js
justinshreve: Nope -- I've updated this line in https://github.com/woocommerce/wc-admin/pull/1500/commits/45badff93d44a9920192796be5875bab131c0a93. Thanks!

package.json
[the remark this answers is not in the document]
justinshreve: Updated in https://github.com/woocommerce/wc-admin/pull/1500/commits/45badff93d44a9920192796be5875bab131c0a93.

bin/generate-feature-config.php
```diff
@@ -0,0 +1,28 @@
+<?php
+/**
+ * Generates an array of feature flags, based on the config used by the client application.
+ *
+ * @package WooCommerce Admin
+ */
+
+$phase = isset( $_SERVER['WC_ADMIN_PHASE'] ) ? $_SERVER['WC_ADMIN_PHASE'] : ''; // WPCS: sanitization ok.
+if ( empty( $phase ) || ! in_array( $phase, array( 'development', 'plugin', 'core' ) ) ) {
```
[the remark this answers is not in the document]
justinshreve: Good catch -- the in_array is enough here. I get in the habit of checking variables in PHP to avoid throwing notices, but in this case I just defined it so it's unnecessary.

Fixed in https://github.com/woocommerce/wc-admin/pull/1500/commits/45badff93d44a9920192796be5875bab131c0a93.

justinshreve:
Thanks for the review @Aljullu! Your comments have been addressed.

docs/feature-flags.md
timmyc: ✨thanks for the in-depth explanation - I hadn't considered / wasn't aware of the webpack implications here. Totally okay with it knowing that now.

docs/feature-flags.md
timmyc: Ah nice that it works that way. I'm fine with it as-is, just commenting about prior work on the same topic.

timmyc (approved):
Thanks for the updates Justin - just went through the full test instructions and all is working wonderfully! I'd love to get this in, and use the flag'ed build for 0.7.0's release 🎉

PR #1506, opened 2019-02-08
Update customer type structure

### LINKED ISSUE #1504: Change returning customers to track by date_first_order
As discussed in p90Yrv-XN-p2 we need to restructure how returning customers are calculated.

This issue involves:

#### Add first order date to customers table

Remove `returning_customer` from the `wc_order_stats` table and add `date_first_order` to the customers table.

#### Sync date of first order on order save

The `date_first_order` table should be updated with the earliest customer order date on order save hook.

#### Change the API response for customer type

This will need a JOIN on the customers table and a change in how `customer_type` is calculated.
https://github.com/woocommerce/wc-admin/blob/8f7c32fedd3379977c9eeb24fc2d30dc82569282/includes/data-stores/class-wc-admin-reports-orders-data-store.php#L49

#### Remove `is_returning_customer()`

We should be able to remove this once `returning_customer` is removed from the order stats table.
https://github.com/woocommerce/wc-admin/blob/8f7c32fedd3379977c9eeb24fc2d30dc82569282/includes/data-stores/class-wc-admin-reports-orders-stats-data-store.php#L517

### LINKED ISSUE #1505: Re-sync customer first order date on order update
Blocked by #1504 and #1473.

**Describe the bug**
As described in p90Yrv-XN-p2 any date or status updates to an order could render the returning customer status incorrect.

**To Reproduce**

1. Create an order.
2. Create another order with the same customer.
3. Note that the `date_first_order` is using the first order.
4. Update the the first order to have a date after the second order.
5. Note the `date_first_order` is still referencing the first.
6. The same thing happens if an order status is changed from an included to an excluded status.

**Expected behavior**
Expected when any order is changed to recalculate the `date_first_order` if necessary.

joshuatf:
Fixes #1504, Fixes #1505

### Accessibility

- Ex: Open page `url`
- Click XYZ…

### DIFF
--- a/includes/class-wc-admin-api-init.php
+++ b/includes/class-wc-admin-api-init.php
@@ -579,12 +579,13 @@ class WC_Admin_Api_Init {
 				WC_Admin_Reports_Products_Data_Store::sync_order_products( $order_id ),
 				WC_Admin_Reports_Coupons_Data_Store::sync_order_coupons( $order_id ),
 				WC_Admin_Reports_Taxes_Data_Store::sync_order_taxes( $order_id ),
+				WC_Admin_Reports_Customers_Data_Store::sync_order_customer( $order_id ),
 			)
 		);
 
 		// If all updates were either skipped or successful, we're done.
 		// The update methods return -1 for skip, or a boolean success indicator.
-		if ( 4 === absint( $result ) ) {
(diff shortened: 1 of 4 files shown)


joshuatf:
Closing this as it looks like we're going in a different direction.

PR #1515, opened 2019-02-10
Added timezone handling

### LINKED ISSUE #489: Add support for timezones to the REST API data layer.
Currently, the API assumes all times are in UTC, which is not the case, the API should stick to local timezone set in WP settings, and create time buckets accordingly.

### LINKED ISSUE #1414: Product report displaying inaccurate hourly data on chart
The times are off in the product report. (pictured below) My store is set to Los Angeles time, which means I placed this order in the 2pm hour. It looks like it's calculating off of GMT. (+8hrs) 

screen shot 2019-01-29 at 2 28 14 pm

However, it's interesting to note that the times ARE correct in the order report and the revenue report: 

screen shot 2019-01-29 at 2 25 57 pm

screen shot 2019-01-29 at 2 41 52 pm

peterfabian:
Fixes #489, fixes #1414.

This PR changes semantics of `date_created` columns so that they are no longer storing UTC time, but local time. 
That makes wc-admin API consistent with WordPress REST API, where `after` and `before` parameters are compared against `post_date`, not `post_date_gmt`.

It also allows us to slice and dice the data into time intervals in the database without requiring any additional database commands. If we stored the UTC timestamp and wanted to group the data by time intervals, we would have to convert the timestamp to correct timezone in the database, which would require having timezone tables filled in the db (by running `mysql_tzinfo_to_sql /usr/share/zoneinfo | mysql -u root mysql`) and that might be difficult on some hostings. I wanted to avoid that.

That also means that datetime without timezone info is assumed to be in local timezone set in WordPress general settings, i.e. `after=2019-02-01T00:00:00` means after Feb 1st in the local timezone, not UTC.

Also, this PR adds conversion of forced timezone information to local timezone, i.e. `2019-02-01T00:00:00+00:00` is a UTC time, which gets converted to `2019-02-01 01:00:00` in Berlin or `2019-01-31 19:00:00` in New York for db queries.

- Database rebuild is required for this change, as previous data are no longer valid, plus data type of `date_created` column changed from timestamp to datetime
- check that correct orders are being included when running with different date/time parameters


justinshreve:
I'll do a deeper review of this tomorrow, but I did some testing on this today when looking at #1414 and it fixes the issue there 👍. I edited your PR description so it will get closed when this merges.

includes/class-wc-admin-reports-interval.php
```diff
@@ -26,6 +26,44 @@ class WC_Admin_Reports_Interval {
...
+		$datetime->setTimezone( new DateTimeZone( 'GMT' ) );
+		return $datetime;
+	}
+
+	/**
+	 * Returns default 'before' parameter for the reports.
+	 *
+	 * @return DateTime
+	 */
+	public static function default_before() {
```
justinshreve: Nice idea/refactor here, allowing the different data stores to be cleaned up when setting defaults 👍.

includes/class-wc-admin-reports-interval.php
```diff
@@ -422,9 +461,6 @@ public static function next_year_start( $datetime, $reversed = false ) {
 	 * @return DateTime
 	 */
 	public static function iterate( $datetime, $time_interval, $reversed = false ) {
-		// $result_datetime           =
-		// $result_timestamp_adjusted = $result_datetime->format( 'U' ) - 1;
-		// $result_datetime->setTimestamp( $result_timestamp_adjusted );
 		return call_user_func( array( __CLASS__, "next_{$time_interval}_start" ), $datetime, $reversed );
```
justinshreve: Since it looks like `$time_interval` comes from the `query` args, should we do any checking to be safe that it is a valid interval function before calling it?

emdashcodes (approved):
I have one comment/question, but otherwise this is looking good to me and is testing great after rebuilding. Hour charts are working well and displaying the correct order time. Other intervals are also looking good. All tests pass via `phpunit`. Pre-approving.

timmyc:
Looks all good to go after a rebase, thanks @peterfabian for getting this one fixed up!

includes/class-wc-admin-reports-interval.php
peterfabian: I think it's validated for REST queries via the controllers, e.g. 

https://github.com/woocommerce/wc-admin/blob/368c02e53086f9f3a21451e14b2b08ab664fbd15/includes/api/class-wc-admin-rest-reports-orders-stats-controller.php#L366-L379

However, if used directly as `WC_Admin_Reports_*_Query` or `WC_Admin_Reports_*_Data_Store`, then invalid value might result in fatals, so it's a valid concern. I think there are more args where this is true.

justinshreve: Ah yeah, good point on REST handing validation.

It might be worth adding some protection to this and other args later, but definitely doesn't need to be a blocker. 👍

peterfabian:
Thanks for the reviews and testing.

PR #1517, opened 2019-02-11
Check and update customer first orders to fix returning_customers

### LINKED ISSUE #1505: Re-sync customer first order date on order update

joshuatf:
Fixes #1505

Checks if the `returning_customer` flags need to be updated for a customer's order on order update.  Prevents multiple "first time orders" for customers when dates match or an order date is set in the past.  Also shaves 10 seconds off test run times 🎉 

Per discussion in p90Yrv-XN-p2 I think this is a more performant option than the one outlined in #1504. 

1.  Make sure phpunit tests all pass.
2.  Make multiple orders with the same customer.
3.  Set the latter order date earlier than the first order made by that customer.
4.  Check that all orders are tagged with `returning_customer` in the database except the order with the earliest creation date.

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-orders-stats-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-orders-stats-data-store.php
@@ -528,7 +528,42 @@ class WC_Admin_Reports_Orders_Stats_Data_Store extends WC_Admin_Reports_Data_Sto
 			return false;
 		}
 
+		// Order is older than previous first order.
+		if ( $order->get_date_created() < new WC_DateTime( $first_order->date_created ) ) {
+			self::set_customer_first_order( $customer_id, $order->get_id() );
+			return false;
+		}
+		// First order date has changed and next oldest is now the first order.
+		if (
+			(int) $order->get_id() === (int) $first_order->order_id &&
+			$order->get_date_created() > new WC_DateTime( $first_order->date_created ) &&
+			$oldest_orders[1] &&
+			new WC_DateTime( $oldest_orders[1]->date_created ) < $order->get_date_created()
+		) {
+			self::set_customer_first_order( $customer_id, $oldest_orders[1]->order_id );
+			return true;
+		}
+
 		return (int) $order->get_id() !== (int) $first_order->order_id;
+	}
+
+	/**
+	 * Set a customer's first order and all others to returning.
+	 *
+	 * @param int $customer_id Customer ID.
+	 * @param int $order_id Order ID.
+	 */
+	protected static function set_customer_first_order( $customer_id, $order_id ) {
+		global $wpdb;
+		$orders_stats_table = $wpdb->prefix . self::TABLE_NAME;
+
+		$wpdb->query(
+			$wpdb->prepare(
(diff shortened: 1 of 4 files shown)


includes/data-stores/class-wc-admin-reports-orders-stats-data-store.php
```diff
@@ -515,24 +515,56 @@ protected static function get_num_items_sold( $order ) {
...
+		}
+
+		$first_order = $oldest_orders[0];
+
+		// Order is older than previous first order.
+		if ( $order->get_date_created() < new WC_DateTime( $first_order->date_created ) ) {
+			self::set_customer_first_order( $customer_id, $order->get_id() );
+			return false;
+		}
+		// First order date has changed and next oldest is now the first order.
```
timmyc: To do this logic here, should we first verify that `$oldest_orders[1]` exists, and if so maybe assign it to a variable to keep things more legible here? `$second_order` maybe? I see the check on 541, but maybe we can skip the entire logic block by checking that earlier.

tests/reports/class-wc-tests-reports-orders-stats.php
```diff
@@ -178,9 +178,6 @@ public function test_populate_and_query_statuses() {
 			$order->set_shipping_total( 0 );
 			$order->set_cart_tax( 0 );
 			$order->save();
-
-			// Wait one second to avoid potentially ambiguous new/returning customer.
-			sleep( 1 ); // @todo Remove this after p90Yrv-XN-p2 is resolved.
```
timmyc: 💤 👋

timmyc (commented):
I just tested with a guest user, three orders. After the first two orders, both showed not returning:

same-day-orders

I then edited the date of the first order to be yesterday, still but had `0` in the returning column.

I placed a third order, and still all had `0` for returning:

three-orders

includes/data-stores/class-wc-admin-reports-orders-stats-data-store.php
joshuatf: Seems like a good idea.  Implemented in https://github.com/woocommerce/wc-admin/pull/1517/commits/0994f3bbf7e67d058b39bb91d2559206f85a45c0

joshuatf:
@timmyc Good catch!  I think @peterfabian has an open PR that resolves guests issues; [here](https://github.com/woocommerce/wc-admin/pull/1479/files#diff-010b483ced6850c6b09b77a5ba03f31aR506) is the line that fixes it.  I'll comment on that PR and let him know that we'll probably still be moving forward with `returning_customer`.

In the meantime, this PR might be easier to test with a newly registered customer until guest issues are resolved.

timmyc (approved):
Testing out well for registered customers LGTM 🚢

PR #1519, opened 2019-02-11
Remove items from search input with backspace key

### LINKED ISSUE #1352: Backspace in a search input should remove the last item
- Go to the _Customers_ report and select _Advanced Filters_ under the _Show_ dropdown.
- Add a filter of type _Name_.
- Write a couple of names.
- Press the backspace.

**Actual results:**
- Nothing happens.

**Expected results:**
- Every time I press the backspace, I would expect the last name to be removed.

names

tiagonoronha:
Fixes #1352 

The title is pretty self explanatory on this one! 😅

### Screenshots

backspace_test

* Go to the _Customers_ report and select _Advanced Filters_ under the _Show_ dropdown.
* Add a filter of type _Name_.
* Write a couple of names.
* Press the backspace.
* It should remove the last item


tiagonoronha:
@justinshreve I can't seem to replicate any of the issues you described. 🤔Any tips?

justinshreve:
Thanks for adding the changelog entry!

> @justinshreve I can't seem to replicate any of the issues you described. 🤔Any tips?

For the reordering issue, make sure that the second or third tag you choose should alphabetically be in front of the others. Right after you hit enter, it will change positions. I just tested on `master` and it's happening there, so we can open a separate issue for it.

For the backspace issue, I'm just removing one right after each other. (Hit backspace twice quickly). You should see one of the items flash back in.

This is on Chrome.

tiagonoronha:
@justinshreve no luck, I'm afraid.

Adding names doesn't reorder:

adding

When deleting items quickly others don't flash back in:

deleting

@Aljullu mind trying this as well?

Aljullu:
@tiagonoronha sure! I could reproduce it in the _Advanced filters_ but not in the Table search input:

image

You can get there going to `analytics/customers?filter=advanced` and adding a _Name_ filter.

Let me know if that doesn't help!

tiagonoronha:
Thanks for the explanation :) I looked into it and @joshuatf appears to be right.

There's an issue open for that already - https://github.com/woocommerce/wc-admin/issues/1520

tiagonoronha:
Tested https://github.com/woocommerce/wc-admin/pull/1539 together with this PR. I can confirm that the reordering issue is gone.

The backspace issue is still around though.

timmyc:
@tiagonoronha could we get this merged in once the conflict is resolved so we can include this change in tomorrow's npm releases?

tiagonoronha:
@timmyc @justinshreve @joshuatf @Aljullu 

The backspace glitch appears to be related to state being updated twice in a very short span of time.

First, the state is updated immediately when Search changes (`onSearchChange`), but then a request is fired to the API via `componentDidUpdate` that calls `updateLabels` that takes a little while to complete, and that's what is causing the issue.

I'm not sure if I understand why that request is needed? Any pointers on how to about fixing the issue? :)

Btw @timmyc I think we can merge this regardless of the issue above, it's a separate thing. Happy to move this conversation to a new issue.

Aljullu:
@tiagonoronha I have to say I couldn't reproduce the backspace issue yet, but I will try to answer your question anyway. :slightly_smiling_face: 

> I'm not sure if I understand why that request is needed? Any pointers on how to go about fixing the issue? :)

That's to allow updating filters directly from the URL. So if the URL changes, the filters react to those changes. Because the URL only has the `id`s, we need to call `getLabels`.

In order to fix your specific issue, I think we could check if `filter.value` has any `id` that is not already in `this.state.selected` before calling `getLabels`. So if there are no new `id`s, there is no need to call `config.input.getLabels( filter.value, query ).then( this.updateLabels );`. I guess that will fix the issue.

tiagonoronha:
@Aljullu thank you, that helps! I've updated the PR with your suggestion, it appears to do the trick. :)

@justinshreve perhaps you can review the changes and test since you can replicate the issue?

PR #1521, opened 2019-02-11
Update 5.0.0 version check and add basic readme.txt

### LINKED ISSUE #1355: Disable analytics and dashboard for users below WP 5.0
WooCommerce 3.6 **will not** have a hard version requirement of WordPress 5.0, however we need to disable analytics and dashboard for stores that can't support it. The rest of their store admin experience should be available. 

@josemarques what kind of notice do we want to display to prompt them to upgrade?

See p6q8Tx-PA-p2 #comment-2818 and p6q6po-9VP-p2 #comment-89107.

> if WP < 5.0 we don’t show the dashboard or reports, and don’t load the blocks JS as well.

> along with a notice/nudge for upgrading.

> Dashboard and analytics are not required for the rest of WooCommerce to function, so we should just disable them and allow the rest of the store to function.

emdashcodes:
Fixes #1355.

This PR adds a WordPress.org `readme.txt` file, containing a "Requires at least" header which the plugin repository uses for showing compatible plugins. I've added a few keywords and such for now, but we can continue filling this out in future PRs before release.

It also updates the notice, requiring 5.0.0.

screen shot 2019-02-11 at 3 20 43 pm

* Test this branch on 4.9.9 or build a zip using `pm run-script build:release` and install it. Verify that you see the version check message.
* Test on 5.0.0 and see everything displays correctly.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -1,0 +1,15 @@
+=== WooCommerce Admin ===
+Contributors: automattic
+Tags: ecommerce, e-commerce, store, sales, reports, analytics, stats, woo commerce, woocommerce
+Requires at least: 5.0.0
+Tested up to: 5.0
+Stable tag: 1.0.0
+License: GPLv2
+License URI: https://github.com/woocommerce/wc-admin/blob/master/LICENSE.md
+
+A feature plugin for a modern, javascript-driven WooCommerce admin experience.
+
+== Description ==
+
+
+== Changelog ==
--- a/readme.txt
+++ b/readme.txt
@@ -1,6 +1,6 @@
 === WooCommerce Admin ===
 Contributors: automattic
-Tags: ecommerce, e-commerce, store, sales, reports, analytics, stats, woo commerce, woocommerce
+Tags: ecommerce, e-commerce, store, sales, reports, analytics, dashboard, activity, notices, insights, stats, woo commerce, woocommerce
 Requires at least: 5.0.0
 Tested up to: 5.0.3
 Stable tag: 1.0.0
--- a/readme.txt
+++ b/readme.txt
@@ -2,7 +2,7 @@
 Contributors: automattic
 Tags: ecommerce, e-commerce, store, sales, reports, analytics, stats, woo commerce, woocommerce
 Requires at least: 5.0.0
-Tested up to: 5.0
+Tested up to: 5.0.3
 Stable tag: 1.0.0
 License: GPLv2
 License URI: https://github.com/woocommerce/wc-admin/blob/master/LICENSE.md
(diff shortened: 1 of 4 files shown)


readme.txt
```diff
@@ -0,0 +1,15 @@
+=== WooCommerce Admin ===
+Contributors: automattic
+Tags: ecommerce, e-commerce, store, sales, reports, analytics, dashboard, activity, notices, insights, stats, woo commerce, woocommerce
+Requires at least: 5.0.0
+Tested up to: 5.0.3
+Stable tag: 1.0.0
+License: GPLv2
+License URI: https://github.com/woocommerce/wc-admin/blob/master/LICENSE.md
+
+A feature plugin for a modern, javascript-driven WooCommerce admin experience.
```
joshuatf: This seems good for now, but should we create a follow-up issue to provide a better description (once we come up with it 😅 )?

Seems like this might be one of the driving factors for installs - https://make.wordpress.org/plugins/2018/03/06/understanding-readme-txt/

wc-admin.php
```diff
@@ -40,17 +40,17 @@ function wc_admin_plugins_notice() {
 		);
 	} else {
 		$message = sprintf(
-			/* translators: 1: URL of Gutenberg plugin, 2: URL of WooCommerce plugin */
-			__( 'The WooCommerce Admin feature plugin requires both <a href="%1$s">Gutenberg</a> and <a href="%2$s">WooCommerce</a> (>3.5) to be installed and active.', 'wc-admin' ),
-			'https://wordpress.org/plugins/gutenberg/',
+			/* translators: 1: URL of WordPress.org, 2: URL of WooCommerce plugin */
+			__( 'The WooCommerce Admin feature plugin requires both <a href="%1$s">WordPress</a> (>=5.0) and <a href="%2$s">WooCommerce</a> (>3.5) to be installed and active.', 'wc-admin' ),
```
joshuatf: Should these be `target="_blank"`?

joshuatf (approved):
Testing well and notices show up correctly depending on WP version.  One comment about links and another about possibly creating a follow-up to update this before release, but otherwise looks good to merge. 👍

wc-admin.php
claudiosanches: No problem having no `target="_blank"`, but we could include buttons to install or update.
We also code change `>=` to `or greater` like WP does already:

```
The WooCommerce Admin feature plugin requires both <a href="%1$s">WordPress</a> 5.0 or greater and <a href="%2$s">WooCommerce</a> 3.5 or greater to be installed and active.
```

readme.txt
justinshreve: Thanks for pointing this out. I definitely consider the readme temporary pending the discussions at p8eNEx-3Dl-p2.

I've created an issue at #1536.

PR #1526, opened 2019-02-12
Add products stats endpoint to categories

psealock:
Fixes https://github.com/woocommerce/wc-admin/issues/1284

### Changes

* Remove SwaggerHub endpoint from Categories Report in favor of using `/reports/products/stats`.
* Enable item comparison mode the chart.
* Add label data and filter according category segmentation, eg `?segmentby=category&categories=19,33`.

1. Categories Report
2. See a request made to `/reports/products/stats` instead of SwaggerHub.
3. Show > Comparison and see the chart change to `item-comparison` mode.
4. Select a few categories to compare and see segments get filtered appropriately.

### DIFF
--- a/client/analytics/report/categories/index.js
+++ b/client/analytics/report/categories/index.js
@@ -4,6 +4,7 @@
  */
 import { Component, Fragment } from '@wordpress/element';
 import PropTypes from 'prop-types';
+import { __ } from '@wordpress/i18n';
 
 /**
  * WooCommerce dependencies
@@ -20,23 +21,53 @@ import ReportChart from 'analytics/components/report-chart';
 import ReportSummary from 'analytics/components/report-summary';
 
 export default class CategoriesReport extends Component {
+	getChartMeta() {
+		const { query } = this.props;
+		const isCategoryDetailsView =
+			'top_items' === query.filter ||
+			'top_revenue' === query.filter ||
+			'compare-categories' === query.filter;
+
+		const isSingleCategoryView = query.categories && 1 === query.categories.split( ',' ).length;
+		const mode =
+			isCategoryDetailsView || isSingleCategoryView ? 'item-comparison' : 'time-comparison';
+		const itemsLabel = __( '%d categories', 'wc-admin' );
+
+		return {
+			itemsLabel,
+			mode,
+		};
+	}
+
 	render() {
 		const { query, path } = this.props;
+		const { mode, itemsLabel } = this.getChartMeta();
+
+		const chartQuery = {
+			...query,
+		};
+
+		if ( 'item-comparison' === mode ) {
+			chartQuery.segmentby = 'category';
+		}
 
 		return (
 			<Fragment>
 				<ReportFilters query={ query } path={ path } filters={ filters } />
 				<ReportSummary
 					charts={ charts }
-					endpoint="categories"
-					query={ query }
+					endpoint="products"
+					query={ chartQuery }
 					selectedChart={ getSelectedChart( query.chart, charts ) }
 				/>
 				<ReportChart
+					filters={ filters }
 					charts={ charts }
-					endpoint="categories"
+					mode={ mode }
+					endpoint="products"
 					path={ path }
-					query={ query }
+					query={ chartQuery }
+					itemsLabel={ itemsLabel }
 					selectedChart={ getSelectedChart( query.chart, charts ) }
 				/>
 				<CategoriesReportTable query={ query } />
@@ -47,4 +78,5 @@ export default class CategoriesReport extends Component {
 
 CategoriesReport.propTypes = {
 	query: PropTypes.object.isRequired,
+	path: PropTypes.string.isRequired,
 };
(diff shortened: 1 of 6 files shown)


includes/class-wc-admin-reports-segmenting.php
```diff
@@ -342,17 +342,31 @@ protected function set_all_segments() {
...
+				{$where_clause}
+				",
+				ARRAY_A
+			); // WPCS: cache ok, DB call ok, unprepared SQL ok.
+
+			foreach ( $categories as $category ) {
+				$id                    = $category['catogory_id'];
+				$segments[]            = $id;
+				$segment_labels[ $id ] = $category['name'];
+			}
```
psealock: `segment_label` was always `NULL` here, preventing the chart from functioning. Also, categories were not being filtered. I could not get `get_categories` to work, so I ended up writing the query.

Is there a better way?

rrennick: Try `wp_list_pluck( $categories, 'term_id' );`.

client/analytics/report/categories/index.js
```diff
@@ -18,25 +18,56 @@ import CategoriesReportTable from './table';
...
+		const { query } = this.props;
+		const isCategoryDetailsView =
+			'top_items' === query.filter ||
+			'top_revenue' === query.filter ||
+			'compare-categories' === query.filter;
+
+		const isSingleCategoryView = query.categories && 1 === query.categories.split( ',' ).length;
+		const mode =
+			isCategoryDetailsView || isSingleCategoryView ? 'item-comparison' : 'time-comparison';
+		const itemsLabel = __( '%s categories', 'wc-admin' );
```
justinshreve: Nit: (and I just realized I did the same thing here that needs changed) https://github.com/woocommerce/wc-admin/blob/64587ad379ddc63b463bedd10460d8c1e34806b9/client/analytics/report/products/index.js#L43-L45) but This should probably be `%d` since it is only ever a count.

client/wc-api/reports/stats/operations.js
```diff
@@ -24,8 +24,6 @@ const statEndpoints = [
 	'taxes',
 	'customers',
 ];
-// @todo Remove once swagger endpoints are phased out.
```
justinshreve: 🎉

includes/class-wc-admin-reports-segmenting.php
```diff
@@ -342,17 +342,31 @@ protected function set_all_segments() {
...
 		} elseif ( 'category' === $this->query_args['segmentby'] ) {
-			$categories = get_categories(
-				array(
-					'taxonomy' => 'product_cat',
-				)
-			);
-			$segments   = wp_list_pluck( $categories, 'cat_ID' );
+			$where_clause = "WHERE taxonomy = 'product_cat'";
+			if ( 0 < count( $this->query_args['categories'] ) ) {
+				$category_ids  = implode( ',', $this->query_args['categories'] );
```
justinshreve: We should make sure these IDs are cast to `int`.

includes/class-wc-admin-reports-segmenting.php
```diff
@@ -342,17 +342,31 @@ protected function set_all_segments() {
...
 
 			foreach ( $segment_objects as $segment ) {
-				$id = $segment->get_id();
-				$segments[] = $id;
+				$id                    = $segment->get_id();
+				$segments[]            = $id;
 				$segment_labels[ $id ] = $segment->get_name();
 			}
 		} elseif ( 'category' === $this->query_args['segmentby'] ) {
-			$categories = get_categories(
```
justinshreve: Instead of the query, does the `include` query argument work instead? It looks like get_categories should support it.

https://github.com/woocommerce/wc-admin/blob/94b296b3460d11d3c14a7eac6349693072315c6b/includes/class-wc-admin-reports-segmenting.php#L314-L318

justinshreve: Or is this the problem you mentioned below?

emdashcodes (commented):
This is testing well for me, nice job! Just a few comments below.

psealock:
Thanks for the suggestions and reviews @justinshreve @rrennick !

I don't know why `get_categories` wasn't returning something usable for me yesterday, but its working now and more readable than a query. `wp_list_pluck` is a great suggestion as well, thanks.

I'm using `%d` now in the `itemsLabel` in categories and products.

client/analytics/report/products/index.js
```diff
@@ -41,8 +41,8 @@ class ProductsReport extends Component {
 			isSingleProductView && isSingleProductVariable ? 'variations' : 'products';
 		const label =
 			isSingleProductView && isSingleProductVariable
-				? __( '%s variations', 'wc-admin' )
-				: __( '%s products', 'wc-admin' );
+				? __( '%d variations', 'wc-admin' )
+				: __( '%d products', 'wc-admin' );
```
justinshreve: Thanks for fixing that here as well!

emdashcodes (approved):
Thanks for the changes, and I'm glad `get_categories` worked. This is testing well for me still. 👍

PR #1533, opened 2019-02-12
Advanced filters alignment in IE11

### LINKED ISSUE #1511: Bug: Advanced Filters Style IE11
When validating the Orders report in IE 11, I noticed the Advanced Filters are not rendering well:

image

__Steps To Reproduce__
- Open any report with advanced filters in IE11
- Add an advanced filter, and note the styling looks like above

tiagonoronha:
Fixes #1511

We are using CSS Grid for the layout of the filters, but unfortunately, support is very poor in IE11 so a few changes are required.

This PR:

* Adds a wrapper (`span`) around the title of the filter. Naked text nodes can't be used as grid columns in IE11. 
* Moves placeholders in the config files to the components 
* Adds IE11 compatibility CSS 

### Screenshots

Before:

52513339-16a26480-2bbf-11e9-9faf-bb5297286ad7

After:

screenshot 2019-02-12 at 18 30 16

* Open any report with advanced filters in IE11
* Add an advanced filter, and see what the styling looks like

### DIFF
--- a/packages/components/src/filters/advanced/date-filter.js
+++ b/packages/components/src/filters/advanced/date-filter.js
@@ -186,7 +186,7 @@ class DateFilter extends Component {
 	}
 
 	render() {
-		const { config, filter, isEnglish } = this.props;
+		const { className, config, filter, isEnglish } = this.props;
 		const { rule } = filter;
 		const { labels, rules } = config;
 		const screenReaderText = this.getScreenReaderText( filter, config );
@@ -195,7 +195,7 @@ class DateFilter extends Component {
 			components: {
 				rule: (
 					<SelectControl
-						className="woocommerce-filters-advanced__rule"
+						className={ classnames( className, 'woocommerce-filters-advanced__rule' ) }
 						options={ rules }
 						value={ rule }
 						onChange={ this.onRuleChange }
@@ -204,7 +204,7 @@ class DateFilter extends Component {
 				),
 				filter: (
 					<div
-						className={ classnames( 'woocommerce-filters-advanced__input-range', {
+						className={ classnames( className, 'woocommerce-filters-advanced__input-range', {
 							'is-between': 'between' === rule,
 						} ) }
 					>
--- a/packages/components/src/filters/advanced/date-filter.js
+++ b/packages/components/src/filters/advanced/date-filter.js
@@ -215,7 +215,7 @@ class DateFilter extends Component {
 		} );
 		/*eslint-disable jsx-a11y/no-noninteractive-tabindex*/
 		return (
-			<fieldset tabIndex="0">
+			<fieldset className="woocommerce-filters-advanced__line-item" tabIndex="0">
 				<legend className="screen-reader-text">{ labels.add || '' }</legend>
 				<div
 					className={ classnames( 'woocommerce-filters-advanced__fieldset', {
--- a/packages/components/src/filters/advanced/date-filter.js
+++ b/packages/components/src/filters/advanced/date-filter.js
@@ -7,7 +7,7 @@ import interpolateComponents from 'interpolate-components';
 import { SelectControl } from '@wordpress/components';
 import { find, partial } from 'lodash';
 import classnames from 'classnames';
-import { sprintf, __, _x } from '@wordpress/i18n';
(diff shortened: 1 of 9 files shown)


timmyc:
Ping @Aljullu - I believe we were able to get grid working elsewhere with a polyfill at one point - though I could be incorrect ( happens often ).

tiagonoronha:
The polyfill does help, but it doesn't fix all the issues like having to manually specify the columns with `-ms-grid-column`. 

It also doesn't fix vertical alignment for text not wrapped by any elements, that's a known IE11 bug. It'd work if the text node was wrapped with a `span` though.

packages/components/src/filters/advanced/style.scss
```diff
@@ -122,19 +124,24 @@
...
 	}
-
-	&.is-english {
-		display: grid;
-		grid-template-columns: 100px 150px auto;
-
-		@include breakpoint( '<782px' ) {
-			display: block;
-		}
-	}
```
psealock: This `is-english` class is more about sentence structure than an `rtl` issue. The first and second columns have a fixed width and the third occupies the rest of the space.

screen shot 2019-02-13 at 4 13 49 pm

When the language is not english, we can't rely on a known order, so we abandon the column width definitions.
See https://github.com/woocommerce/wc-admin/pull/457 for a bit more details.

Aljullu:
We have a mixin named [`set-grid-item-position`](https://github.com/woocommerce/wc-admin/blob/master/client/stylesheets/abstracts/_mixins.scss#L52) that might help you to position grid child elements in IE11.

You can see an example of its usage [here](https://github.com/woocommerce/wc-admin/pull/370/files#diff-7675d4e8d29c184b4dc0c9f255cb14e7R38). The only requirement is that you need a common selector for all grid children, so in this specific case I think there is no other option than wrapping the text node.

tiagonoronha:
Thanks for the insights @psealock and @Aljullu. I'll rework this PR :)

tiagonoronha:
Ok, PR updated.

Quite a few changes since all the placeholders for `interpolateComponents` were on the config files and needed to be moved to the components so that I could add a `span` to the title.

I updated the PR description as well.

packages/components/src/filters/advanced/select-filter.js
```diff
@@ -64,16 +65,20 @@ class SelectFilter extends Component {
...
 	render() {
-		const { config, filter, onFilterChange, isEnglish } = this.props;
+		const { className, config, filter, onFilterChange, isEnglish } = this.props;
 		const { options } = this.state;
 		const { key, rule, value } = filter;
 		const { labels, rules } = config;
 		const children = interpolateComponents( {
-			mixedString: labels.title,
+			mixedString: sprintf(
+				'{{title}}%s{{/title}} {{rule /}} {{filter /}}',
```
jeffstieler: This change causes `SelectControl` type filters with no `rules` to render full width:

screen shot 2019-02-14 at 8 57 42 am

Perhaps we should check for `config.rules` and conditionally include `{{rule /}}` in the `mixedString`?

tiagonoronha: Thanks @jeffstieler, nice find! Updated, should be good now :)

packages/components/src/filters/advanced/date-filter.js
```diff
@@ -186,16 +186,20 @@ class DateFilter extends Component {
...
 	render() {
-		const { config, filter, isEnglish } = this.props;
+		const { className, config, filter, isEnglish } = this.props;
 		const { rule } = filter;
 		const { labels, rules } = config;
 		const screenReaderText = this.getScreenReaderText( filter, config );
 		const children = interpolateComponents( {
-			mixedString: labels.title,
+			mixedString: sprintf(
+				'{{title}}%s{{/title}} {{rule /}} {{filter /}}',
```
jeffstieler: Missed this before - these `mixedStrings` were being passed through the translation function `__()` before.. unless we don't actually need that for RTL support, I think that needs to get restored.

tiagonoronha: You're totally right. Moved the placeholders back to the respective config.js files and included the new `title` placeholder there.

Issue #1457, opened 2019-02-01
'reports/customers' should allow limiting the results per user id


Aljullu:
In order to allow searching customers in the _Customers_ report table, we need the `reports/customers` endpoint to support a new parameter (for example, `customers`) which would accept a list of ids in order to limit the result set to those specific user ids.

Similar to the `products` parameter in the `reports/products` endpoint.

peterfabian:
When you say `customer_id`, is it `customer_id` or `user_id` from the `wc_customer_lookup` table?

Aljullu:
Good question. It should be `user_id`, so it matches the `id` attribute returned by the [Customers WC endpoint](https://woocommerce.github.io/woocommerce-rest-api-docs/#customer-properties).

I updated the issue title/text to make it more clear.

Aljullu:
After #1534 was merged. I think we don't need to fix this issue anymore.

Issue #1454, opened 2019-02-01
Table search: Increase the limit of 10 results


Aljullu:
Coming from https://github.com/woocommerce/wc-admin/pull/1381#discussion_r252481505.

When loading a report with a table search active, we are limiting the number of results to 10. We would like to increase that number and consider adding a UI message if we can't load all of them.

![image](https://user-images.githubusercontent.com/22080/60995573-99168800-a307-11e9-986d-33944266cc0d.png)

timmyc:
Also wanted to capture @jeffstieler's thoughts [from here](https://github.com/woocommerce/wc-admin/pull/1534#issuecomment-463250441)

```
in addition to increasing the `LIMIT` on these queries, I think the way we populate the report table for "all values that include" needs to change. The table is given the `<= 10` IDs from the fuzzy search query rather than the search term itself.
```

dreiris:
This ticket is being closed and the feature request is being moved to Monday.com. The feature will be evaluated with other feature requests and prioritized in Monday.com.

PR #1534, opened 2019-02-12
Hook up customers report table search box

### LINKED ISSUE #1455: 'customers' endpoint should allow searching by first + last_name
Consider we have a user with `name=Lorem` and `last_name=Ipsum`.

Querying `/wc/v4/customers` with parameter `search=Lorem` would return that result but `search=Lorem Ipsum` wouldn't.

We need to be able to search users by full name for the _Name_ filter and table search of the _Customers_ report.

jeffstieler:
Fixes #1455.

This PR hooks the customers report table search header to the customer lookup table data.

Supporting changes introduced:
* Rewire the `v4/customers` endpoint to use the customer lookup table data
* Add support for fuzzy name searching to customers report data store and controllers
* Add support for specifying customer IDs to the customers report data store and controllers
* Rename `customer_id` to `id` in customers endpoint responses for `wc-api` compatibility

### Accessibility

### Screenshots

screen shot 2019-02-12 at 11 38 07 am

- Navigate to Customers Report
- Verify that searching at the top of the table has autocomplete results
- Verify that selecting the first "partial match" option has reasonable/expected results

### Known Issue?

I believe there are some faults in the way these table search components work. When there are more than 10 results, no paging controls are displayed and the stats are only for the results shown in the table. This seems to be true for all report tables using this search box.

### DIFF
--- a/includes/api/class-wc-admin-rest-reports-customers-controller.php
+++ b/includes/api/class-wc-admin-rest-reports-customers-controller.php
@@ -172,7 +172,7 @@ class WC_Admin_REST_Reports_Customers_Controller extends WC_REST_Reports_Control
 			'title'      => 'report_customers',
 			'type'       => 'object',
 			'properties' => array(
-				'customer_id'          => array(
+				'id'                   => array(
 					'description' => __( 'Customer ID.', 'wc-admin' ),
 					'type'        => 'integer',
 					'context'     => array( 'view', 'edit' ),
--- a/includes/api/class-wc-admin-rest-reports-customers-controller.php
+++ b/includes/api/class-wc-admin-rest-reports-customers-controller.php
@@ -60,6 +60,7 @@ class WC_Admin_REST_Reports_Customers_Controller extends WC_REST_Reports_Control
 		$args['avg_order_value_max'] = $request['avg_order_value_max'];
 		$args['last_order_before']   = $request['last_order_before'];
 		$args['last_order_after']    = $request['last_order_after'];
+		$args['customers']           = $request['customers'];
 
 		$between_params_numeric    = array( 'orders_count', 'total_spend', 'avg_order_value' );
 		$normalized_params_numeric = WC_Admin_Reports_Interval::normalize_between_params( $request, $between_params_numeric, false );
@@ -446,6 +447,17 @@ class WC_Admin_REST_Reports_Customers_Controller extends WC_REST_Reports_Control
 			'format'            => 'date-time',
 			'validate_callback' => 'rest_validate_request_arg',
 		);
+		$params['customers']               = array(
+			'description'       => __( 'Limit result to items with specified customer ids.', 'wc-admin' ),
+			'type'              => 'array',
+			'sanitize_callback' => 'wp_parse_id_list',
+			'validate_callback' => 'rest_validate_request_arg',
+			'items'             => array(
+				'type' => 'integer',
+			),
+
+		);
+
 		return $params;
 	}
 }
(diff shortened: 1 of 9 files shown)


timmyc:
> I believe there are some faults in the way these table search components work. When there are more than 10 results, no paging controls are displayed and the stats are only for the results shown in the table. This seems to be true for all report tables using this search box.

➕ 

#1454

packages/components/src/search/autocompleters/customers.js
```diff
@@ -78,7 +76,7 @@ export default {
 	getOptionCompletion( customer ) {
 		return {
 			id: customer.id,
-			label: getName( customer ),
+			label: customer.name,
```
joshuatf: Not sure about UI implications here, but what do you think about adding the email next to customer name?

Related to above comment about customer's without associated names, but would also help us distinguish between same names (when customer creates multiple accounts under different emails or checkout last name field has been removed).

screen shot 2019-02-13 at 2 59 39 pm

joshuatf (commented):
Nice work on this, Jeff.  Couple questions about whether or not we should consider customers without names and email support, but otherwise testing very well.

Aljullu (commented):
This looks really good, thanks for working on this @jeffstieler! This PR seems to fix #1457 as well, right?

I noticed an issue when using the _Name_ filter in _Advanced Filters_:

customer-name


I guess it's because [this line](https://github.com/woocommerce/wc-admin/blob/master/client/analytics/report/customers/config.js#L62) still relies on `customer.first_name` and `last_name` instead of `customer.name`.

jeffstieler:
> ➕
> 
> #1454

@timmyc in addition to increasing the `LIMIT` on these queries, I think the way we populate the report table for "all values that include" needs to change. The table is given the `<= 10` IDs from the fuzzy search query rather than the search term itself.

> This PR seems to fix #1457 as well, right?

@Aljullu if you're OK with taking in the Customer ID (not WP User ID) then yes we can consider that resolved.

> I noticed an issue when using the _Name_ filter in _Advanced Filters_:

Nice catch! Will get that fixed.

packages/components/src/search/autocompleters/customers.js
jeffstieler: I'm good with that so long as design is.. what say you @josemarques @timmyc ?

includes/api/class-wc-admin-rest-reports-customers-controller.php
```diff
@@ -333,8 +334,8 @@ public function get_collection_params() {
 			),
 			'validate_callback' => 'rest_validate_request_arg',
 		);
-		$params['name']                    = array(
-			'description'       => __( 'Limit response to objects with a specfic customer name.', 'wc-admin' ),
+		$params['search']                  = array(
```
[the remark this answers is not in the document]
jeffstieler: I'd like to discuss this behavior in the larger context of the interplay with the advanced filter options. If we can search name and email in the table, should we also have name and email advanced filters? _cc @timmyc_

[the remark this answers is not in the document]
timmyc: Yeah I think the optimal solution here is to make this table search field support searching by name and/or email - that seems the most powerful yet ux-simple way for users to interact with the table. But, that is just my opinion, and I agree this is something we should discuss further with teamx and determine the best path forward.

I also think ( repeating from slack convos ) that we will likely get more feedback on this report and the filters as more users interact with it that could help us in making changes here.

packages/components/src/search/autocompleters/customers.js
timmyc: Building on the comment above, I think we leave it as-is for now, and get more feedback from Ze and David when he is back next week before we make any further additions/changes right there.

timmyc (approved):
@jeffstieler this is testing out well for me, and I think is a great improvement, so thanks for your time in investigating the options and getting this delivered. Marking this as ready to merge, and will make sure we do a bit of discussion around advanced filters in the customers report and beyond during next weeks' wc-admin hangout.

PR #1535, opened 2019-02-12
Add stock stats endpoint and hook up to the stock report.

### LINKED ISSUE #174: Analytics: Stock Report
All base analytics components are being developed for the Revenue Report. Once those components ship, work can commence on this report.

**Design Post**: p6riRB-3dV-p2
**Mock API**: NA uses existing endpoints modified for `v3`
**REST API Post**: p7bje6-16i-p2

image

### Child Tasks

- [x] #919 REST API: Products Endpoint Low Stock Query
- [x] #921 Analytics: Create Stock Report Page
- [ ] #920 Analytics: Stock Report Table

### LINKED ISSUE #920: Analytics: Stock Report Table
**Design Post**: p6riRB-3dV-p2
**Mock API**: NA uses existing endpoints modified for `v3`
**REST API Post**: p7bje6-16i-p2

This issue involves creating the table for the stock report. While this appears to be one of the easier tables we will implement, there are still a few challenges:

- How do we handle retrieving the full parent product name for variations? Additional request to the REST API for certain parent product ID's?
- Table Summary: Currently there isn't a "stats" style endpoint for stock that would return this totals. So that might need to be handled in a follow-up task.

### LINKED ISSUE #1527: REST API: Stock Stats Endpoint
The final piece needed for the Stock Report ( #174 ) is the ability to retrieve stock totals to display in the footer of the report:

image

I initially thought we could possibly add the values needed to the `totals` portion of the `reports/products/stats` endpoints, but that endpoint is date-based so it just didn't feel like a correct place to put this data. And furthermore I feel having a particular `reports/stock/stats` endpoint might be of use in showing the "unread" notification in the Stock Activity panel eventually - and I could see the endpoint being valuable to the mobile apps.

Essentially I'm thinking of a very minimal endpoint that does not accept any arguments, and returns a payload with the following product stock totals returned for each `wc_get_product_stock_status_options()` ( like we have in the stock report [here](https://github.com/woocommerce/wc-admin/blob/master/includes/api/class-wc-admin-rest-reports-stock-controller.php#L105-L112), so would return something like:

```
{
  products: (int),
  lowstock: (int),
  instock: (int),
  outofstock: (int),
  onbackorder: (int),
}
```

I think it would also be prudent to have some aggressive caching on this endpoint due to the meta queries

emdashcodes:
Closes #174, closes #920, closes #1527.

This PR adds a `/reports/stock/stats` endpoint that returns the number of products per stock status. The results are cached per the issue, since these can be slower meta queries. I followed the caching we are using on Store on WP.com (the `/data/counts` endpoint) which seems like it has been working well.

This PR also adds a "on backorder" filter and summary.

I also did a little cleanup on the customers endpoint, since it also does not contain an intervals object.

* Run `phpunit` and make sure all tests are passing.
* Run `npm test` and make sure all tests are passing.
* Go to the stock report, and verify the summary numbers load. Smooth generator products should contain a variety of statuses.
* Go change a product's inventory so it's status changes. Go back to the report and make sure the cache busted.
* Spot check the customers report and revenue report for the summary number, since this PR makes a slight adjustment to the report util method.

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-stock-stats-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-stock-stats-data-store.php
@@ -1,0 +1,137 @@
+<?php
+/**
+ * WC_Admin_Reports_Stock_Stats_Data_Store class file.
+ *
+ * @package WooCommerce Admin/Classes
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * WC_Reports_Stock_Stats_Data_Store.
+ */
+class WC_Admin_Reports_Stock_Stats_Data_Store extends WC_Admin_Reports_Data_Store implements WC_Admin_Reports_Data_Store_Interface {
+
+	/**
+	 * Get stock counts for the whole store.
+	 *
+	 * @param array $query Not used for the stock stats data store, but needed for the interface.
+	 * @return array Array of counts.
+	 */
+	public function get_data( $query ) {
+		$report_data = array();
+		$low_stock_transient_name = 'wc_admin_stock_count_lowstock';
+		$low_stock_count          = get_transient( $low_stock_transient_name );
+		if ( false === $low_stock_count ) {
+			$low_stock_count = $this->get_low_stock_count();
+			set_transient( $low_stock_transient_name, $low_stock_count, DAY_IN_SECONDS * 30 );
+		}
+		$report_data['lowstock'] = $low_stock_count;
+
+		$status_options = wc_get_product_stock_status_options();
+		foreach ( $status_options as $status => $label ) {
+			$transient_name = 'wc_admin_stock_count_' . $status;
+			$count          = get_transient( $transient_name );
+			if ( false === $count ) {
+				$count = $this->get_count( $status );
+				set_transient( $transient_name, $count, DAY_IN_SECONDS * 30 );
+			}
+			$report_data[ $status ] = $count;
+		}
+
+		$product_count_transient_name = 'wc_admin_product_count';
+		$product_count                = get_transient( $product_count_transient_name );
+		if ( false === $product_count ) {
+			$product_count = $this->get_product_count();
+			set_transient( $product_count_transient_name, $product_count, DAY_IN_SECONDS * 30 );
+		}
+		$report_data['products'] = $product_count;
+		return $report_data;
+	}
+
+	/**
+	 * Get low stock count.
+	 *
+	 * @return int Low stock count.
+	 */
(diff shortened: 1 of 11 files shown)


includes/data-stores/class-wc-admin-reports-stock-stats-data-store.php
```diff
@@ -0,0 +1,137 @@
...
+	 *
+	 * @param array $query Not used for the stock stats data store, but needed for the interface.
+	 * @return array Array of counts.
+	 */
+	public function get_data( $query ) {
+		$report_data = array();
+		$low_stock_transient_name = 'wc_admin_stock_count_lowstock';
+		$low_stock_count          = get_transient( $low_stock_transient_name );
+		if ( false === $low_stock_count ) {
+			$low_stock_count = $this->get_low_stock_count();
```
psealock: I think you'll need to check the low inventory thresholds set on individual items as well

psealock (commented):
This is testing well, I just have a comment on checking low threshold amounts set on individual products.

justinshreve:
> I think you'll need to check the low inventory thresholds set on individual items as well

This is a really good point/question and not something I thought about. Thanks for raising it!

After doing some thinking on this and looking around, I think we can solve it, but it will definitely take some more work in a follow-up PR. I think we would need to do some processing when a product stock is updated to set a new piece of meta/status, so that we can easily pull these out in a query. We would also need an upgrade routine to correctly set existing products.

I did some looking around, and right now WooCommerce considers these products "In Stock" on both the products page, and the current stock report. Considering this PR matches current core behavior -- including on the product management screen -- should we leave as is and log an issue for the future (post initial release), or is updating these statuses in all these places something that is important for v1? cc @timmyc @josemarques @LevinMedia 

screen shot 2019-02-13 at 11 48 52 am

screen shot 2019-02-13 at 11 49 00 am

screen shot 2019-02-13 at 11 49 27 am

timmyc (approved):
A trifecta of issue closing in one PR - that has to be some sort of repo record right there 🎉 

This is looking great to me, couple of minor comments on the code, but pre-approving as-is. Lets do follow up with maybe a "Question" issue for how to handle the individual stock levels. Since we are matching what works elsewhere in core it seems good as-is to me, but /ht to @psealock for the eagle eye and catching that.

psealock:
>Considering this PR matches current core behavior -- including on the product management screen -- should we leave as is and log an issue for the future

Sounds good to me, lets create an issue. As a user, it seems this would frustrate me.

PR #1539, opened 2019-02-13
Only update search labels when IDs are mismatched

### LINKED ISSUE #1520: Autocomplete items reordering
When selecting multiple autocomplete items, they seem to reorder themselves alphabetically:

filter

joshuatf:
Fixes #1520

Prevents reordering of labels when adding new selections in advanced filters.

### Before
old

### After
new

1.  Open any report with advanced filters.
2.  Add a filter that allows adding multiple selections via autocomplete search (products, customers, etc).
3.  Add multiple selections.
4.  Verify that selections don't get reordered when adding.

### DIFF
--- a/packages/components/src/filters/advanced/search-filter.js
+++ b/packages/components/src/filters/advanced/search-filter.js
@@ -40,7 +40,12 @@ class SearchFilter extends Component {
 	}
 
 	updateLabels( selected ) {
-		this.setState( { selected } );
+		const prevIds = this.state.selected.map( item => item.id );
+		const ids = selected.map( item => item.id );
+
+		if ( ! isEqual( ids.sort(), prevIds.sort() ) ) {
+			this.setState( { selected } );
+		}
 	}
 
 	onSearchChange( values ) {


packages/components/src/filters/advanced/search-filter.js
```diff
@@ -40,7 +40,12 @@ class SearchFilter extends Component {
 	}
 
 	updateLabels( selected ) {
```
joshuatf: @Aljullu Does this need to be called still from `componentDidUpdate()`?  It looks like maybe searching in the table is supposed to open the advanced filter as well, but I'm not seeing this happening on master.

Commit for reference - https://github.com/woocommerce/wc-admin/pull/1085/commits/6247cc30572980afb04f1686e16300a6e85898ed

Aljullu: Good catch! At that time, the table search input and the _Advanced Filters_ were connected, but that's no longer the case, so I think it can be safely removed.

I created #1544 to address that.

Aljullu (commented):
Thanks for investigating this @joshuatf!

This is working as expected, <s>but #1544 solves this issue too, so I would go for that solution if that sounds good to you.</s>

Aljullu:
Thinking about it twice, #1544 removes the ability to make filters react to changes in the URL. That would make extensibility harder so that's probably not the best option here.

I will approve this PR and close mine, but happy to discuss it further if you want.

Aljullu (approved):

joshuatf:
@Aljullu Thanks for reviewing.  That makes sense; we'll go ahead and merge as is for now then.

PR #1540, opened 2019-02-13
Prevent hash creation from breadcrumbs

### LINKED ISSUE #1480: Breadcrumbs header triggering history
The breadcrumbs header is calling `getPersistedQuery` from `@woocommerce/navigation`, which triggers `createHashHistory()` even when lazy-creating. In certain cases, this is not desired (e.g. when accessing a single order page, `http://my_wordpress_site.local/wp-admin/post.php?post=12345&action=edit`). The hash history created causes the page to reload with a `#/` suffix.

joshuatf:
Fixes #1480 

Removes persisted query from breadcrumbs to avoid appending hashes on non-wc-admin pages.

### Before
screen shot 2019-02-13 at 5 05 03 pm

### After
screen shot 2019-02-13 at 5 05 14 pm

1.  Visit any page that contains the wc-admin breadcrumbs, but is not a wc-admin route (e.g., single orders or products).
2.  After the page has loaded, verify that no has `#/' is appended to the URL.

### DIFF
--- a/client/header/index.js
+++ b/client/header/index.js
@@ -12,7 +12,7 @@ import PropTypes from 'prop-types';
 /**
  * WooCommerce dependencies
  */
-import { getNewPath, getPersistedQuery } from '@woocommerce/navigation';
+import { getNewPath } from '@woocommerce/navigation';
 import { Link } from '@woocommerce/components';
 
 /**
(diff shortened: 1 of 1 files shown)


emdashcodes (approved):
This has definitely been a constant annoyance, so thank you for fixing this one! 😄 

Change looks good, and is testing well for me.

PR #1541, opened 2019-02-13
Update numeric comparisons for customer data

### LINKED ISSUE #1507: Customers Report: Number of Orders Filter Confusing
**Describe the bug**
When using the "Number of Orders" advanced filter on the customers report, and selecting the "More Than" or "Less Than" modifiers - the resulting data displayed shows `>=` or `<=` values. So for example if I want to see customers with "More than 2" orders, customers with 2 orders are displayed in the table.

**To Reproduce**

1. Have a data set with customers whom have more than 2 orders, and 1 order, and exactly 2 orders :)
2. Open the customers report
3. Add "Number of Orders" filter, set it for more than 2, or less than 2
4. Note customers with 2 orders show up in each scenario

**Expected behavior**
Due to the verbiage of the filter, I would expect that when looking for customers with "More Than 2 Orders" that the table would only show customers with a total of > 2 orders, not >=. Likewise for less than.

joshuatf:
Fixes #1507 

Updates the customer comparison filters (avg order value, order count, etc) to be < and > to match phrasing instead of <= and >=.

If `Between` is selected instead of `More than` or `Less than` then the comparison becomes >= and <= so that if you selected "Between 0 and 1 orders" you would see all customers having 0 and 1 orders.  (Does this match your expectations?)

### Screenshots
screen shot 2019-02-13 at 5 22 03 pm

1.  Open the Customers report.
2.  Add advanced filters that allow Less Than, More Than, and Between comparisons.
3.  Verify that Less Than results are actually less than the value and not less than or equal to.
4.  Verify the same with More Than.
5.  Verify that between is inclusive of both values (i.e. >= value1 and <= value2).

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-customers-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-customers-data-store.php
@@ -253,17 +253,18 @@ class WC_Admin_Reports_Customers_Data_Store extends WC_Admin_Reports_Data_Store 
 			$subclauses = array();
 			$min_param  = $numeric_param . '_min';
 			$max_param  = $numeric_param . '_max';
+			$or_equal   = isset( $query_args[ $min_param ] ) && isset( $query_args[ $max_param ] ) ? '=' : '';
 
 			if ( isset( $query_args[ $min_param ] ) ) {
 				$subclauses[] = $wpdb->prepare(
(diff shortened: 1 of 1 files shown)


Aljullu (approved):
Tested and it works as expected. :rocket: 

> (Does this match your expectations?)

It does, at least for me.

PR #1545, opened 2019-02-13
add build:feature-config to readme instructions

rrennick:
#1500 added `npm build:feature-config` to the build process. This PR adds that to the `npm run build` command.

*Testing instructions*
1. Delete `includes/feature-config.php` if it exists.
2. Run `npm run build`
3. Load the dashboard of the site without error.


rrennick:
> Should we instead update the command in `package.json`? so you can still just run `npm run build` and it will run both pieces?

Sounds like a better fix.

rrennick:
@justinshreve Thanks. I updated the PR and testing instructions.

PR #1547, opened 2019-02-13
Segmented Selection: re-apply borders for all browsers

psealock:
Borders were moved to `ie.scss` in https://github.com/woocommerce/wc-admin/pull/1199/files. This PR moves the CSS rules back to the main stylesheets as those rules apply to all browsers.

### Before

screen shot 2019-02-14 at 12 39 55 pm

### After

screen shot 2019-02-14 at 12 40 50 pm

1. Go to any date related report, such as Products.
2. Open the Datepicker and see borders on preset options, eg "Today", "Yesterday".


timmyc (approved):
LGTM, good catch

PR #1548, opened 2019-02-14
Stock Report: fix stock link

psealock:
Links to edit product pages from Stock Report Table had an incorrect url for products whose `parent_id` was zero.

1. Stock Report
2. See links for products with no parent, ie not a variation
3. They should point to the corresponding product edit page in wp-admin


jeffstieler (approved):
Nice simple change that works. :shipit:

PR #1550, opened 2019-02-14
Autocompleter: increase debounce

psealock:
Fixes https://github.com/woocommerce/wc-admin/issues/1436

It turns out debounce was turned on for almost all autocompleters (`isDebounced: true`), but it was only 250ms. This PR increases the pause before a request is made to 400ms.

1. Go to Products Report
2. Show > Single Product
3. Watch the requests as you type. The request will wait 400ms after your last keystroke to fire.


jeffstieler (approved):
Nice! This was particularly painful when testing on large datasets. 🛳 it.

PR #1556, opened 2019-02-14
Add tax refund calculations

### LINKED ISSUE #1449: Refunds are not correctly synced with the tax lookup table
**Describe the bug**
When issuing a line item refund with tax, a new entry is created in the `wc_order_tax_lookup` table.  WC traditionally treats a refund as a type of order, but in order to keep stats accurate, the refund should be calculated against the existing entry.

**To Reproduce**

1. Go to an order in wp-admin.
2. Refund a line item with tax.
3. Note the additional order created with negative tax and an `order_id` of the refund ID.

**Expected behavior**
Expected the change in tax to be calculated on the existing row with the same order ID.

**Screenshots**
`283796` is the order and `283797` is a refund in the following screenshot:
screen shot 2019-02-01 at 6 23 59 pm

joshuatf:
Fixes #1449 

Adjusts the tax amounts when an order is refunded or when a refund is deleted.

### Screenshots
screen shot 2019-02-14 at 1 49 58 pm
screen shot 2019-02-14 at 1 51 09 pm

1.  Make an order that is taxed and has shipping that is taxed.
2.  Refund a partial amount of tax on the product line item and on the shipping line item.
3.  Check that taxes are correctly updated to reflect the refunded amounts.
4.  Delete the refund from the order page.
5.  Check that the original tax amounts are once again shown.

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-taxes-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-taxes-data-store.php
@@ -268,15 +268,24 @@ class WC_Admin_Reports_Taxes_Data_Store extends WC_Admin_Reports_Data_Store impl
 		$num_updated = 0;
 
 		foreach ( $tax_items as $tax_item ) {
+			$item_refunds     = 0;
+			$shipping_refunds = 0;
+			foreach ( $order->get_items() as $item_id => $item ) {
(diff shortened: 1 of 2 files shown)


timmyc (approved):
Looking good and testing well for me 🙌

PR #1557, opened 2019-02-14
Update returning customer values to respect excluded statuses

### LINKED ISSUE #1473: Orders with excluded statuses count towards returning customers
**Describe the bug**
Returning customers checks to see if an order was previously made by a customer in https://github.com/woocommerce/wc-admin/blob/ac4f8a4da7a1e0f9cc54e5f40b7a825bc0e07e28/includes/data-stores/class-wc-admin-reports-orders-stats-data-store.php#L504

Orders that fall into the excluded statuses can still mark a customer as "returning."  Should a customer only be marked as "returning" if they've had a previously successful order?

**To Reproduce**

1. Create a first order by a customer and set to an excluded order status (trash, failed, cancelled, refunded are defaults).
2. Create a second order by the same customer on a different date.
3. Select the date from the second order (not including the date from the original order so that the customer would not qualify as "new").
4.  Note that the customer is marked as "returning" even though the first order failed so they should technically be a new user.

**Expected behavior**
Expected the user to be "new" since the first order never succeeded.

**Additional context**
We can use the excluded statuses array to easily check inside `is_returning_customer()` however this would probably also require us to update ALL orders if the excluded statuses setting is changed to retroactively mark orders as returning in the `wc_order_stats` table.

Furthermore, we would also need to check all orders by a customer and update if a status is changed on a previous order.

joshuatf:
Fixes #1473 

Doesn't count orders that fall within excluded statuses as first-time orders and retroactively updates orders when a status change causes a shift in the original first order.

1.  Create an order with a customer (don't use a guest until https://github.com/woocommerce/wc-admin/pull/1479 is merged).
2.  Make sure that order status is set to an excluded status.
3.  Make another order with that same customer and set to a non-excluded status.
4.  Check that the 2nd order is marked as a new customer.
5.  Update the 1st order to a non-excluded status.
6.  Check that the first order (assuming it has an earlier creation date) is now set to "New" and the 2nd order is set to "Returning."
7.   Make sure all tests pass.


includes/data-stores/class-wc-admin-reports-orders-stats-data-store.php
```diff
@@ -75,6 +75,7 @@ class WC_Admin_Reports_Orders_Stats_Data_Store extends WC_Admin_Reports_Data_Sto
 	public static function init() {
 		add_action( 'woocommerce_refund_deleted', array( __CLASS__, 'sync_on_refund_delete' ), 10, 2 );
 		add_action( 'delete_post', array( __CLASS__, 'delete_order' ) );
+		add_action( 'save_post', array( __CLASS__, 'sync_order' ) );
```
jeffstieler: Is the existing order sync not sufficient here? https://github.com/woocommerce/wc-admin/blob/master/includes/class-wc-admin-reports-sync.php#L172

includes/data-stores/class-wc-admin-reports-orders-stats-data-store.php
```diff
@@ -527,20 +528,24 @@ protected static function is_returning_customer( $order ) {
...
+			! in_array( $order->get_status(), $excluded_statuses, true )
+		) {
 			self::set_customer_first_order( $customer_id, $order->get_id() );
 			return false;
 		}
-		// First order date has changed and next oldest is now the first order.
-		$second_order = isset( $oldest_orders[1] ) ? $oldest_orders[1] : false;
+		// First order date or status has changed and next oldest is now the first order.
+		$date_change   = $second_order && $order->get_date_created() > new WC_DateTime( $first_order->date_created ) &&
+			new WC_DateTime( $second_order->date_created ) < $order->get_date_created();
```
jeffstieler: Nitpick: I think this would be easier to read if each condition was on its own line with a comment.

includes/data-stores/class-wc-admin-reports-orders-stats-data-store.php
jeffstieler: Confirmed that this is unnecessary in my testing.

jeffstieler (approved):
Nice work @joshuatf. 👍

This tested well without the additional `save_post` hook - **please remove before merging.**

Optionally take care of formatting nitpick as well.

includes/data-stores/class-wc-admin-reports-orders-stats-data-store.php
joshuatf: @jeffstieler Thanks for catching that!  This hook is not necessary; just used to speed things up during testing and accidentally committed.

includes/data-stores/class-wc-admin-reports-orders-stats-data-store.php
joshuatf: Agreed.  Refactored in https://github.com/woocommerce/wc-admin/pull/1557/commits/ba3f70edc42ad4e4957b401a2931d99b9f5a2261

PR #1558, opened 2019-02-14
Move IE11 specific CSS to ie.scss file

Aljullu:
This is a follow-up of #1547.

This declaration:

```SCSS
@include set-grid-item-position( 2, 10 );
```

is only required in IE and it generates 90 lines of CSS. I expected it to be more, but still, I think it's not a bad idea to have it in a specific IE file.

Also, I think this declaration is not needed:

```SCSS
display: block;
```

### Screenshots
image

* Go to any date related report, such as Products.
* Open the Datepicker and verify borders are still visible and there are no regressions.

### DIFF
--- a/packages/components/src/ie.scss
+++ b/packages/components/src/ie.scss
@@ -1,5 +1,6 @@
 /**
  * Internal Dependencies
  */
+@import 'segmented-selection/ie.scss';
 @import 'summary/ie.scss';
 @import 'table/ie.scss';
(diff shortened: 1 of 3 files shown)


jeffstieler (approved):
Looks good in Chrome and IE - verified that `components/ie.css` only loads in IE as well. 👍

psealock (approved):
Thanks for the change @Aljullu. I should have dug a bit deeper on this one

PR #1560, opened 2019-02-14
Stop Hello Dolly lyrics from displaying in the notices tab

### LINKED ISSUE #1429: WordPress notice toggle displays Hello Dolly Lyrics
If you have the `Hello Dolly` plugin installed and activated, lyrics will display under the notices tab when expanded. We should hide the lyrics, and make sure that it does not affect the count of notices.

There is a "#dolly" id that we can target.

screen shot 2019-01-30 at 9 46 53 pm

tiagonoronha:
Fixes #1429

Blacklists the Hello Dolly lyrics from showing up in the notices tab by targeting the `#dolly` id.

However, stopping the notice from displaying in the tab creates a few different issues:

screenshot 2019-02-14 at 15 41 13

screenshot 2019-02-14 at 15 42 01

My suggestion would be to hide the lyrics, but that kinda defeats the purpose of moving them to the uncollapsed area.

Any other suggestions? :) @timmyc 

- Install Hello Dolly
- Check if lyrics show up in as a Notice

### DIFF
--- a/client/header/activity-panel/wordpress-notices.js
+++ b/client/header/activity-panel/wordpress-notices.js
@@ -135,6 +135,11 @@ class WordPressNotices extends Component {
 		if ( 'message' === element.id && 'updated' === element.className ) {
 			return false;
 		}
+
+		if ( 'dolly' === element.id ) {
+			return false;
+		}
+
 		return true;
 	}
 
--- a/client/header/activity-panel/wordpress-notices.js
+++ b/client/header/activity-panel/wordpress-notices.js
@@ -137,6 +137,7 @@ class WordPressNotices extends Component {
 		}
 
 		if ( 'dolly' === element.id ) {
+			element.style.display = 'none';
 			return false;
 		}
 
--- a/client/header/activity-panel/wordpress-notices.js
+++ b/client/header/activity-panel/wordpress-notices.js
@@ -116,32 +116,38 @@ class WordPressNotices extends Component {
 
 	// Some messages should not be displayed in the toggle, like Jetpack JITM messages or update/success messages
 	shouldCollapseNotice( element ) {
-		if ( element.classList.contains( 'jetpack-jitm-message' ) ) {
-			return false;
-		}
+		const noticesToHide = [
+			[ null, [ 'jetpack-jitm-message' ] ], // ID and/or array of classes
+			[ 'woocommerce_errors', null ],
+			[ null, [ 'hidden' ] ],
+			[ 'message', [ 'notice', 'updated' ] ],
+			[ 'dolly', null ],
+		];
 
-		if ( 'woocommerce_errors' === element.id ) {
-			return false;
-		}
+		let collapse = true;
+		for ( let i = 0; collapse && i < noticesToHide.length; i++ ) {
+			const [ id, classes ] = noticesToHide[ i ];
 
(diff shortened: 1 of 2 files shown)


justinshreve:
> My suggestion would be to hide the lyrics

I would be fine with this. It would still display on every other WordPress page and just not WooCommerce pages. The only other idea is to move/display it in the footer.

tiagonoronha:
Yeah, I think hiding makes the most sense. Updated the PR!

client/header/activity-panel/wordpress-notices.js
```diff
@@ -135,6 +135,12 @@ class WordPressNotices extends Component {
 		if ( 'message' === element.id && 'updated' === element.className ) {
 			return false;
 		}
+
+		if ( 'dolly' === element.id ) {
+			element.style.display = 'none';
```
joshuatf: I don't think anyone is going to complain if this gets hidden, but we can also fix the layout issues by changing the float.

```suggestion
			element.style.cssFloat = 'none';
```

client/header/activity-panel/wordpress-notices.js
```diff
@@ -135,6 +135,12 @@ class WordPressNotices extends Component {
 		if ( 'message' === element.id && 'updated' === element.className ) {
 			return false;
 		}
+
+		if ( 'dolly' === element.id ) {
```
joshuatf: Optional refactor:  Since this list of notices is growing and may continue to grow, I'm wondering if we should create an array of items with IDs and class names.

```js
const noticesToHide = [
    [ 'message', [ 'notice' ] ], // ID and array of classes
    [ 'dolly', null ],
]
```
Then again, this may not be ideal if we need to apply CSS styles here or target anything outside the realm of IDs and classes.

rrennick:
I have been following this and this morning wondered if it wouldn't be easier to just unhook the notice when in the WC Admin dashboard:

`remove_action( 'admin_notices', 'hello_dolly' );`

client/header/activity-panel/wordpress-notices.js
tiagonoronha: The problem with this is that then the lyrics don't always show up in the same location. In WC appears in one place, WordPress another. So I think it makes the most sense to hide it.

client/header/activity-panel/wordpress-notices.js
tiagonoronha: Decided to give your suggest a try: https://github.com/woocommerce/wc-admin/pull/1560/commits/c979c9d1049e6cc8459e14f8394f142e78bebf50 

Makes code a bit more lengthy but makes it easy to add new ids/classes.

Happy to revert it though!

justinshreve:
@tiagonoronha What do you think about keeping the refactor you did (I think it'll make adding various things easier in the future) but still unhooking hello dolly via PHP? We should be able to use ` wc_admin_is_admin_page` and `wc_admin_is_embed_enabled_wc_page` to do this on `wc-admin` pages only.

tiagonoronha:
@justinshreve sure thing! Will work on that! :)

tiagonoronha:
@justinshreve removed Hello Dolly from JS and added a new function `wc_admin_remove_notices()` that removes the Hello Dolly action.

PR #1561, opened 2019-02-14
Fix downloads stats data store sorting

### LINKED ISSUE #1508: Reports downloads stats data store sorting broken
Methods at the bottom of the class (`sort_intervals` and `interval_cmp`) access private members of its parent. I assume that would cause an error if it runs.

emdashcodes:
Fixes #1508.

The download stats data store was doing some unnecessary over-ridding to make sorting work, when `normalize_order_by` would work by itself. Cleaning this up also stops accessing private members of the parent class.

To Test:
* Make sure you have some downloadable products and downloads logged.
* View the downloads report, and test sorting by date and product name.


timmyc (approved):
Love all that red! Testing well for me.

PR #1566, opened 2019-02-14
Chart legend: show pointer in all clickable area

Aljullu:
Noticed that when reviewing #1563.

Some clickable areas of the chart legend items don't show the pointer icon (see screenshots).

### Screenshots
_Before:_
click-legend-01

_After:_
click-legend-02

- Go to any report with a chart.
- Hover the legend elements.
- Verify the pointer icon is shown when hovering any part of the clickable area.


timmyc (approved):
Good catch, and thanks for the detailed PR with screen grabs 🙇

PR #1567, opened 2019-02-14
update chart type query string key to chartType

### LINKED ISSUE #1450: Chart 'type' parameter conflicts with the stock API endpoint 'type' parameter
- Switch to the bar chart and verify `?type=bar` is appended to the URL.
- Go to the _Stock_ report.

An error appears:

`type is not one of all, lowstock, instock, outofstock, onbackorder.`

rrennick:
Fixes #1450

Replace the chart `type` query string with `chartType` to avoid conflicts with stock report `type` query string.

### Accessibility

- Open the dashboard
- load the product report
- switch to bar graph and check URL has chartType=bar
- switch back to line and check URL has chartType=line

If I should have completed accessibility tests for this, please let me know which ones.


rrennick:
@Aljullu Thanks, merge conflicts resolved. I had some phpcs errors from the merged changes. I will check this in a bit to see if there are any that need to be fixed.

rrennick:
@Aljullu This has passed on Travis so it's ready for review. Thanks.

timmyc:
Looking great @rrennick! Could you add an entry to the [changelog](https://github.com/woocommerce/wc-admin/blob/master/packages/components/CHANGELOG.md) about the prop change here?

packages/components/src/chart/index.js
```diff
@@ -494,7 +494,7 @@ Chart.propTypes = {
 	/**
 	 * Chart type of either `line` or `bar`.
 	 */
-	type: PropTypes.oneOf( [ 'bar', 'line' ] ),
+	chartType: PropTypes.oneOf( [ 'bar', 'line' ] ),
```
Aljullu: Not a big deal, but we usually try to order `propTypes` alphabetically.

packages/components/src/chart/index.js
```diff
@@ -524,7 +524,7 @@ Chart.defaultProps = {
 	showHeaderControls: true,
 	tooltipLabelFormat: '%B %d, %Y',
 	tooltipValueFormat: ',',
-	type: 'line',
+	chartType: 'line',
```
Aljullu: Same for `defaultProps`.

rrennick:
@timmyc @Aljullu Thanks for the review. I have updated the PR with the suggested changes.

Issue #675, opened 2018-10-24
Activity Panel: Inbox: PHP linter warnings


allendav:
There are a few PHP linter warnings being generated for inbox related code.  To see them, run

`npm run lint:php`

and then you'll get (among others):

```
FILE: /Library/WebServer/Documents/wca/wp-content/plugins/wc-admin/includes/class-wc-admin-notes-new-sales-record.php
---------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 2 LINES
---------------------------------------------------------------------------------------------------------------------
 65 | WARNING | strip_tags() is discouraged. Use the more comprehensive wp_strip_all_tags() instead.
 67 | WARNING | strip_tags() is discouraged. Use the more comprehensive wp_strip_all_tags() instead.
---------------------------------------------------------------------------------------------------------------------

FILE: /Library/WebServer/Documents/wca/wp-content/plugins/wc-admin/includes/data-stores/class-wc-admin-notes-data-store.php
------------------------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 20 WARNINGS AFFECTING 11 LINES
------------------------------------------------------------------------------------------------------------------------------------
  40 | WARNING | Usage of a direct database call is discouraged.
  64 | WARNING | Usage of a direct database call is discouraged.
  64 | WARNING | Direct database call without caching detected. Consider using wp_cache_get() / wp_cache_set() or
     |         | wp_cache_delete().
 107 | WARNING | Usage of a direct database call is discouraged.
 107 | WARNING | Direct database call without caching detected. Consider using wp_cache_get() / wp_cache_set() or
     |         | wp_cache_delete().
 142 | WARNING | Usage of a direct database call is discouraged.
 142 | WARNING | Direct database call without caching detected. Consider using wp_cache_get() / wp_cache_set() or
     |         | wp_cache_delete().
 143 | WARNING | Usage of a direct database call is discouraged.
 143 | WARNING | Direct database call without caching detected. Consider using wp_cache_get() / wp_cache_set() or
     |         | wp_cache_delete().
 157 | WARNING | Usage of a direct database call is discouraged.
 157 | WARNING | Direct database call without caching detected. Consider using wp_cache_get() / wp_cache_set() or
     |         | wp_cache_delete().
 185 | WARNING | Usage of a direct database call is discouraged.
[...]

PR #1570, opened 2019-02-14
Phpcs sniff fixes for unit tests

rrennick:
See #675

This has phpcs fixes only. There are several hundred phpcs warnings/errors so I'll approach this with multiple PRs to minimize merge conflicts with other PRs/WIP.

This PR fixes all the sniffs in the `/tests/` folder. The only code change is to replace all the uses of `print_r` with a private function.

- If this passes Travis CI testing then it should be good to merge


rrennick:
@timmyc Thanks for looking it over.

PR #1571, opened 2019-02-14
Taxes Report: fix comparison mode and segmentation labels

psealock:
Fixes https://github.com/woocommerce/wc-admin/issues/1554

When a user selects "Comparison" in the Taxes Report, change the chart mode to `item-comparison` and include `?segmentby=tax_rate_id` in the API request. Also, construct the labels correctly in the segments portion of the response. 

1. Taxes Report.
2. See the chart appear as `time-comparison` for two periods (selected, previous).
2. Show > Comparison.
3. See the chart in `item-comparison` mode.
4. Check that labels in the chart match those found on the table.

### DIFF
--- a/client/analytics/report/taxes/config.js
+++ b/client/analytics/report/taxes/config.js
@@ -3,11 +3,17 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
+import apiFetch from '@wordpress/api-fetch';
+import { find } from 'lodash';
+
+/**
+ * WooCommerce dependencies
+ */
+import { getIdsFromQuery, stringifyQuery } from '@woocommerce/navigation';
 
 /**
  * Internal dependencies
  */
-import { getRequestByIdString } from 'lib/async-requests';
 import { getTaxCode } from './utils';
 import { NAMESPACE } from 'wc-api/constants';
 
@@ -52,15 +58,30 @@ export const filters = [
 			{ label: __( 'All Taxes', 'wc-admin' ), value: 'all' },
 			{
 				label: __( 'Comparison', 'wc-admin' ),
-				value: 'compare-tax-codes',
+				value: 'compare-taxes',
 				chartMode: 'item-comparison',
 				settings: {
 					type: 'taxes',
 					param: 'taxes',
-					getLabels: getRequestByIdString( NAMESPACE + '/taxes', tax => ( {
-						id: tax.id,
-						label: getTaxCode( tax ),
-					} ) ),
+					// @TODO: Core /taxes endpoint should support a collection param 'include'.
+					getLabels: function( queryString = '' ) {
+						const idList = getIdsFromQuery( queryString );
+						if ( idList.length < 1 ) {
+							return Promise.resolve( [] );
+						}
+						const payload = stringifyQuery( {
+							per_page: -1,
+						} );
+						return apiFetch( { path: NAMESPACE + '/taxes' + payload } ).then( taxes => {
+							return idList.map( id => {
+								const tax = find( taxes, { id } );
+								return {
+									id: tax.id,
+									label: getTaxCode( tax ),
+								};
+							} );
+						} );
+					},
 					labels: {
 						helpText: __( 'Select at least two tax codes to compare', 'wc-admin' ),
 						placeholder: __( 'Search for tax codes to compare', 'wc-admin' ),
--- a/client/analytics/report/taxes/index.js
+++ b/client/analytics/report/taxes/index.js
@@ -4,6 +4,7 @@
  */
 import { Component, Fragment } from '@wordpress/element';
 import PropTypes from 'prop-types';
+import { __ } from '@wordpress/i18n';
 
 /**
  * WooCommerce dependencies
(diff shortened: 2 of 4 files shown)


includes/class-wc-admin-reports-segmenting.php
```diff
@@ -363,15 +363,25 @@ protected function set_all_segments() {
 			// 1 -- returning customer
 			$segments = array( 0, 1 );
 		} elseif ( 'tax_rate_id' === $this->query_args['segmentby'] ) {
-			// @todo Do we need to include tax rates that existed in the past, but have never been used? I guess there are other, more pressing problems...
-			// Current tax rates UNION previously used tax rates.
-			$tax_rate_ids = $wpdb->get_results(
-				"SELECT tax_rate_id FROM {$wpdb->prefix}woocommerce_tax_rates
-						UNION 
-						SELECT DISTINCT meta_value FROM {$wpdb->prefix}woocommerce_order_itemmeta where meta_key='rate_id'",
+			$taxes = $wpdb->get_results(
```
justinshreve: I think we should consider refactoring the two queries still present in `set_all_segments`. This and coupons above use direct queries. We should instead use a data store method to make swapping things out easier. (This can be a separate `@todo`/issue/PR).

(I think even `get_categories` should be updated, since CPT access in core is also behind a data store too). We want to be using filterable functions like `wc_get_products` or data store methods and not directly touching the database or post types.

client/analytics/report/taxes/index.js
```diff
@@ -20,22 +21,46 @@ import ReportSummary from 'analytics/components/report-summary';
 import TaxesReportTable from './table';
 
 export default class TaxesReport extends Component {
+	getChartMeta() {
+		const { query } = this.props;
+		const isCompareTaxView = 'compare-tax-codes' === query.filter;
```
justinshreve: I think we should check `query.taxes.split( ',' ).length` here too. Otherwise, the mode immediately switches to compare mode when you change the filter, instead of after clicking the 'compare' button.

emdashcodes (changes requested):
Edit:

It might also be worth you rebasing this branch and then reprocessing your data (since https://github.com/woocommerce/wc-admin/pull/1515 merged).

--

I'm having a couple issues testing this.

1) Compare mode kicks in as soon as you switch to the 'compare' filter. On our other reports, you must select the elements to compare first. See my comment about checking the query.

2) I selected two valid codes in the dropdown, but something weird seems to be happening where one of the labels in the compare is wrong:

All view, showing I have data:

screen shot 2019-02-18 at 3 36 09 pm

Before compare:

screen shot 2019-02-18 at 3 36 25 pm

After hitting compare (notice the one label switched, and the one value is 0):

screen shot 2019-02-18 at 3 36 48 pm

psealock:
Thanks for the review @justinshreve 

RE: Removing direct queries - [WC_Tax](https://docs.woocommerce.com/wc-apidocs/class-WC_Tax.html) doesn't have a direct way of gathering all tax codes. What do you think is the best way to proceed? Add a method to `WC_Tax`? Or create a data store method in `WC_Admin_Reports_Taxes_Stats_Data_Store`?

>Compare mode kicks in as soon as you switch to the 'compare' filter. On our other reports, you must select the elements to compare first.

hmm... switching to the 'compare' filter directly produces a compare mode in Products, Variations, and Categories. Coupons does not work this way, but `getChartMeta` has yet to be implemented in that report. Am I looking at the right thing?

I do now see an issue where selecting two taxes from the chart and clicking "Compare" is not working as expected.

>but something weird seems to be happening where one of the labels in the compare is wrong

Nice find. I can reproduce and will dig into why.

justinshreve:
> hmm... switching to the 'compare' filter directly produces a compare mode... m I looking at the right thing?

Doh! I blame it being Monday. I even recently worked on those compare modes 😆. I just double checked and you're right -- so ignore me about that!

justinshreve:
> Or create a data store method in `WC_Admin_Reports_Taxes_Stats_Data_Store`

This sounds good to me.

psealock:
@justinshreve could you take another look?

I updated `getLabels` for taxes because the API doesn't support an `includes` parameter and simply returns an array of all taxes. Because of this I couldn't use the `getRequestByIdString` helper function.

I also created a data store method in `WC_Admin_Reports_Taxes_Stats_Data_Store` to query the database. I also simplified the query and was able to remove a `JOIN`.

emdashcodes (commented):
I'm still having an issue here where a rate shows up in the chart even though its not in the compare. The table looks fine though.

client/analytics/report/taxes/config.js
```diff
@@ -52,15 +58,29 @@ export const filters = [
...
-						id: tax.id,
-						label: getTaxCode( tax ),
-					} ) ),
+					getLabels: function( queryString = '' ) {
+						const idList = getIdsFromQuery( queryString );
+						if ( idList.length < 1 ) {
+							return Promise.resolve( [] );
+						}
+						const payload = stringifyQuery( {
+							per_page: -1,
```
justinshreve: I don't know if `per_page` works with -1. The API does validation and sets this between 1-100:

https://github.com/woocommerce/woocommerce/blob/d0491072e8a7d0a4535db0ec2bae4a6d43ff64fd/includes/api/v1/class-wc-rest-taxes-controller.php#L665-L673.

Long term, I think we should probably update the API to have an `include` param. That would be consistent with the other filters. That would be a bit more work on top of this PR though. Maybe this is OK for now / initial release of the feature plugin (capped at 100 tax rates via comparison which seams reasonable since you can only chart 5 at once anyway). 

tl;dr I'm OK with maybe just adding a TODO for now about adding an `include` param and the limit here and going with this.

includes/class-wc-admin-reports-segmenting.php
```diff
@@ -370,15 +371,13 @@ protected function set_all_segments() {
...
-			// @todo Do we need to include tax rates that existed in the past, but have never been used? I guess there are other, more pressing problems...
-			// Current tax rates UNION previously used tax rates.
-			$tax_rate_ids = $wpdb->get_results(
-				"SELECT tax_rate_id FROM {$wpdb->prefix}woocommerce_tax_rates
-						UNION 
-						SELECT DISTINCT meta_value FROM {$wpdb->prefix}woocommerce_order_itemmeta where meta_key='rate_id'",
-				ARRAY_A
-			); // WPCS: cache ok, DB call ok, unprepared SQL ok.
-			$segments     = wp_list_pluck( $tax_rate_ids, 'tax_rate_id' );
+			$taxes = WC_Admin_Reports_Taxes_Stats_Data_Store::get_all_taxes();
```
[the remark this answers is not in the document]
justinshreve: I just implemented something similar here too:
https://github.com/woocommerce/wc-admin/pull/1647/

client/analytics/report/taxes/config.js
joshuatf: @justinshreve I think `apiFetch` handles `-1` in batches so this should be okay.  https://github.com/WordPress/gutenberg/blob/1020543c18a7e303e652ba5b27c4e434642ec72c/packages/api-fetch/src/middlewares/fetch-all-middleware.js#L39

But I do agree that long-term it's a good call to have an `include` param instead of fetching all.  

@psealock I think we can handle that in a separate PR if you'd like and I can reword this issue to address the include param - https://github.com/woocommerce/wc-admin/issues/1640

joshuatf (commented):
Nice work on this @psealock!  Good to see the tax labels working.

Left a comment about handling the `include` param in a separate issue, but this seems to be testing well for me.

Probably good to go after addressing @justinshreve's question about filtering returned taxes in segmentation.

psealock:
Thanks for the feedback @joshuatf @justinshreve !

I added an additional argument for filtering by specific tax rates in the data store similar to the one done for Coupons.

I also added a TODO item in the `getLabels` function about the `include` collection param and updated the title in https://github.com/woocommerce/wc-admin/issues/1640 to reflect that.

client/analytics/report/taxes/config.js
joshuatf: I have an open PR that will add the `include` param for use here - https://github.com/woocommerce/wc-admin/pull/1672

emdashcodes (approved):
Thanks for all the updates here! This is looking good and testing well for me now. 🚢 it.

PR #1578, opened 2019-02-15
Update activity panels order endpoint

### LINKED ISSUE #1428: Activity Panel: Orders: Fix customer name
There are two issues with the customer name output on the orders activity card:

---

If an order does not have a first/last name for a guest, the string in the order card of the activity panel looks like this second order:

screen shot 2019-01-30 at 9 33 00 pm

You can make the billing/shipping fields optional with our checkout field editor plugin.

---

The card does not seem to be pulling the stored customers name from the order

screen shot 2019-01-30 at 9 38 26 pm

---

**To Reproduce**
* Remove the first/last name from an order or create an order with no billing first/last name.
* View order in the activity panel.
* Create an order with a registered customer. Notice the name doesn't display.

joshuatf:
Fixes #1428 

* Hooks up the orders activity panel to use the data from `wc_order_stats` for better performance and so that we can later grab user profile information indirectly from the `wc_customer_lookup` table.
* Allows nameless customers in the activity panel
* Adds customer data to the extended info in the orders API response

### Screenshots
screen shot 2019-02-15 at 2 42 44 pm

1.  Check that orders display correctly still in the activity panel.
2.  Install the checkout field editor plugin and set name fields to optional.
3. Create an order with a nameless customer.
4. Note that the `placed by {customer_name}` string does not appear for customers without names.

### DIFF
--- a/client/header/activity-panel/panels/orders.js
+++ b/client/header/activity-panel/panels/orders.js
@@ -32,7 +32,6 @@ import { getAdminLink } from '@woocommerce/navigation';
 import { ActivityCard, ActivityCardPlaceholder } from '../activity-card';
 import ActivityHeader from '../activity-header';
 import ActivityOutboundLink from '../activity-outbound-link';
-import { getOrderRefundTotal } from 'lib/order-values';
 import { QUERY_DEFAULTS } from 'wc-api/constants';
 import withSelect from 'wc-api/with-select';
 
@@ -65,14 +64,13 @@ function OrdersPanel( { orders, isRequesting, isError } ) {
 	);
 
 	const getCustomerString = order => {
-		// We want the billing address, but shipping can be used as a fallback.
-		const address = { ...order.shipping, ...order.billing };
+		const { first_name, last_name } = order.extended_info.customer;
 
-		if ( ! address.first_name && ! address.last_name ) {
+		if ( ! first_name && ! last_name ) {
 			return '';
 		}
 
-		const name = [ address.first_name, address.last_name ].join( ' ' );
+		const name = [ first_name, last_name ].join( ' ' );
 		return sprintf(
 			__(
 				/* translators: describes who placed an order, e.g. Order #123 placed by John Doe */
@@ -95,13 +93,17 @@ function OrdersPanel( { orders, isRequesting, isError } ) {
 							'wc-admin'
 						),
 						{
-							orderNumber: order.number,
+							orderNumber: order.order_id,
 							customerString: getCustomerString( order ),
 						}
 					),
 					components: {
-						orderLink: <Link href={ 'post.php?action=edit&post=' + order.id } type="wp-admin" />,
-						destinationFlag: <Flag order={ order } round={ false } />,
+						orderLink: (
+							<Link href={ 'post.php?action=edit&post=' + order.order_id } type="wp-admin" />
+						),
+						destinationFlag: order.extended_info.customer.country ? (
+							<Flag code={ order.extended_info.customer.country } round={ false } />
+						) : null,
 						// @todo Hook up customer name link
 						customerLink: <Link href={ '#' } type="wp-admin" />,
 					},
@@ -111,16 +113,16 @@ function OrdersPanel( { orders, isRequesting, isError } ) {
 	};
 
 	const cards = [];
-	orders.forEach( ( order, id ) => {
-		const productsCount = order.line_items.reduce( ( total, line ) => total + line.quantity, 0 );
+	orders.forEach( order => {
+		const productsCount = order.extended_info.products.length;
 
-		const total = order.total;
-		const refundValue = getOrderRefundTotal( order );
-		const remainingTotal = getCurrencyFormatDecimal( order.total ) + refundValue;
+		const total = order.gross_total;
+		const refundValue = order.refund_total;
+		const remainingTotal = getCurrencyFormatDecimal( total ) + refundValue;
 
 		cards.push(
 			<ActivityCard
-				key={ id }
+				key={ order.order_id }
 				className="woocommerce-order-activity-card"
 				title={ orderCardTitle( order ) }
 				date={ order.date_created }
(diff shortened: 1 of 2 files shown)


client/header/activity-panel/panels/orders.js
```diff
@@ -162,29 +179,30 @@ function OrdersPanel( { orders, isRequesting, isError } ) {
...
 
 export default compose(
 	withSelect( select => {
-		const { getItems, getItemsError, isGetItemsRequesting } = select( 'wc-api' );
+		const { getReportItems, getReportItemsError, isReportItemsRequesting } = select( 'wc-api' );
 		const ordersQuery = {
 			page: 1,
 			per_page: QUERY_DEFAULTS.pageSize,
-			status: 'processing',
+			status_is: 'processing',
```
joshuatf: Should this only grab orders that are `processing`?  What about `on-hold` items?
/cc @timmyc @jeffstieler

joshuatf: I see there's some discussion around this in https://github.com/woocommerce/wc-admin/issues/935.  An actionable list of orders makes sense to me; I believe `on-hold` may fall into this realm since it often requires manually checking with your bank and confirming payment.

Does that sound accurate?  Any other "actionable" statuses?

timmyc: +1 for having `on-hold` being considered "actionable", but ideally thinking we should either:

- Expose this as a configurable option on the settings page
- Or make it a fitlerable array of statuses that gets passed in via `wcSettings` and consumed in the component(s)

client/header/activity-panel/panels/orders.js
```diff
@@ -64,49 +63,68 @@ function OrdersPanel( { orders, isRequesting, isError } ) {
...
 						}
 					),
 					components: {
-						orderLink: <Link href={ 'post.php?action=edit&post=' + order.id } type="wp-admin" />,
+						orderLink: (
+							<Link href={ 'post.php?action=edit&post=' + order.order_id } type="wp-admin" />
+						),
+						destinationFlag: order.extended_info.customer.country ? (
+							<Flag code={ order.extended_info.customer.country } round={ false } />
+						) : null,
```
timmyc: Thinking while we are here we could hook up this customerLink to the related page in the Customers Report.

client/header/activity-panel/panels/orders.js
```diff
@@ -64,49 +63,68 @@ function OrdersPanel( { orders, isRequesting, isError } ) {
...
-		const refundValue = getOrderRefundTotal( order );
-		const remainingTotal = getCurrencyFormatDecimal( order.total ) + refundValue;
+		const total = order.gross_total;
+		const refundValue = order.refund_total;
+		const remainingTotal = getCurrencyFormatDecimal( total ) + refundValue;
 
 		cards.push(
 			<ActivityCard
-				key={ id }
+				key={ order.order_id }
```
timmyc: nice...

client/header/activity-panel/panels/orders.js
```diff
@@ -64,49 +63,68 @@ function OrdersPanel( { orders, isRequesting, isError } ) {
 		</EllipsisMenu>
 	);
 
-	const orderCardTitle = ( order, address ) => {
-		const name = `${ address.first_name } ${ address.last_name }`;
+	const getCustomerString = order => {
+		const { first_name, last_name } = order.extended_info.customer;
```
timmyc: While I generally trust our API, the output of it is fully filterable... as such do you think we need to verify `order.extended_info` exists first? Perhaps assign it to a const as it is used below too `const orderInfo = order.extended_info || {}`

includes/data-stores/class-wc-admin-reports-orders-data-store.php
```diff
@@ -279,8 +285,12 @@ protected function include_extended_info( &$orders_data, $query_args ) {
 				'products'   => array(),
 				'categories' => array(),
 				'coupons'    => array(),
+				'customer'   => array(),
```
timmyc: I was a bit surprised that there wasn't any test coverage in the [API test suite](https://github.com/woocommerce/wc-admin/blob/master/tests/api/reports-orders.php) around `extended_info` outside of verifying the schema. Might be nice to queue up an issue to add in some more test coverage for the extended_info

timmyc (commented):
Wow this PR is a real treat - love the use of the reporting endpoints + new customer data here. Couple of minor items in the code. As for the "actionable statuses" I'm game with that being handled in a follow-up PR.

Tested this out on an Android Devices 🌟

client/header/activity-panel/panels/orders.js
joshuatf: I was also thinking that maybe this should be a setting option for users, so I'm completely on board with that option.  Issue opened around that [here](https://github.com/woocommerce/wc-admin/issues/1593).

client/header/activity-panel/panels/orders.js
joshuatf: I don't think that option is yet available to us, but I went ahead and opened an issue [here](https://github.com/woocommerce/wc-admin/issues/1594) to handle that.

In the meantime, I hooked up the link but it wont be functional until #1594 is resolved.

client/header/activity-panel/panels/orders.js
joshuatf: That's a good point.  Updated this in https://github.com/woocommerce/wc-admin/pull/1578/commits/46ad004045ba667e1084a900ec0e091b04e79cb5.

Side note: as it stands right now, quite a few of our other views would fail if `extended_info` had been filtered out.  We might want to consider fixing these across the board.

joshuatf: Issue noted for other reports [here](https://github.com/woocommerce/wc-admin/issues/1595).

includes/data-stores/class-wc-admin-reports-orders-data-store.php
joshuatf: Issue created [here](https://github.com/woocommerce/wc-admin/issues/1596).

joshuatf:
@timmyc Thanks for the review!  This is ready for another look.  Created separate issues for testing and customer filtering.

timmyc (approved):
Changes look good and big thanks for opening up the follow-up issues 👍

Issue #1528, opened 2019-02-12
REST API: Create Action Scheduler Group / Queue for Analytics


timmyc:
During devchat #31 - we had a bit of a [discussion around the awesomeness of Action Scheduler](https://woocommercecommunity.slack.com/archives/C4TNYTR28/p1548954948126400) being included in Woo core now, and just how easy it is for developers to implement with the queue'ing system. But the concern was raised that if too many jobs are placed into the default group, it might make it difficult for WP CLI job runners to be configured to prioritize more critical jobs ( i.e. subscription renewals vs populating stats lookup tables ).

As such, this issue is to explore and / implement a group specifically for Woo Analytics on the jobs created here in wc-admin.

@rrennick if you are able to tackle this issue, I think your knowledge in this area would be a huge help. It would also be a nice opportunity for you to check out our queueing logic, and see if there are any issues with how we have things setup.

rrennick:
I created https://github.com/Prospress/action-scheduler/issues/236 to discuss the details/implementation.

rrennick:
The first step in implementing new features in scheduled actions is to test WooCommerce Admin with the latest version of Action Scheduler. Action Scheduler is designed to use the latest version when multiple versions are active. To use the latest version, clone the repository into your plugins folder

```shell
$ git clone git@github.com:Prospress/action-scheduler.git
$ wp plugin activate action-scheduler
```

timmyc:
@rrennick is Woo core not running the latest version? Seems like we should get that in for 3.6 if we are behind.

rrennick:
A new version of AS was released 2 weeks ago - https://github.com/Prospress/action-scheduler/pull/233. The latest version is in WooCommerce Subscriptions 2.5.X so we should find if there were any bugs in the latest that way.

The main advantage to using the repo is that as I work on AS we can test as needed with `git pull`.

justinshreve:
@rrennick @timmyc It looks like we need to open a PR to get Action Scheduler 2.2.0 in `3.6` still, is that correct? I see that we are still running `2.1.0` on `master` there.

rrennick:
@justinshreve Yes, that's on my todo list. We want to give it a bit more time in subscriptions before proceeding with updating in WC.

rrennick:
[...]

PR #1581, opened 2019-02-15
Execute single order pending jobs immediately when wp_debug is enabled

### LINKED ISSUE #1580: Order information is delayed in analytics
**Describe the bug**
The action scheduler takes care of a bunch of heavy lifting in the background, which is awesome.  One downside, is that orders sometimes take a couple minutes to show up.

While I think this is probably okay in the context of a normal site, it does seem to add on a bit of time for manual testing during development.  Adding a temporary `add_action( 'save_post', 'action_being_tested' );` hook resolves the issue, but this is more error prone and needs to be removed.

Could we consider using `WC_Helper_Queue::run_all_pending();` in dev environments?  Or is there anything else that might be slowing down the processing?

/cc @jeffstieler 

**To Reproduce**

1. Create an order.
2. Note that it takes a minute or two sometimes to show up in the `wc_order_stats` table.

**Expected behavior**
Would be nice if information was synced right away.

joshuatf:
Fixes #1580 

Executes any single order pending jobs immediately to avoid the delay in order syncing when developing.

**Questions**
* Does this cause any issues for debugging the action scheduler itself?  My hunch is no, but want to make sure.

1.  Save/update an order.
2.  Note the order is immediately updated.
3.  Run the rebuild tool at WooCommerce->Status->Tools->Rebuild Reports.
4.  Make sure the rebuild is not processed right away and is handled in the background.

### DIFF
--- a/includes/class-wc-admin-reports-sync.php
+++ b/includes/class-wc-admin-reports-sync.php
@@ -79,6 +79,7 @@ class WC_Admin_Reports_Sync {
 
 		// Initialize syncing hooks.
 		add_action( 'wp_loaded', array( __CLASS__, 'orders_lookup_update_init' ) );
+		add_action( 'wp_loaded', array( __CLASS__, 'execute_jobs_in_debug_mode' ) );
 
 		// Initialize scheduled action handlers.
 		add_action( self::QUEUE_BATCH_ACTION, array( __CLASS__, 'queue_batches' ), 10, 3 );
@@ -404,6 +405,25 @@ class WC_Admin_Reports_Sync {
 		foreach ( $customer_ids as $customer_id ) {
 			// @todo Schedule single customer update if this fails?
 			WC_Admin_Reports_Customers_Data_Store::update_registered_customer( $customer_id );
+		}
+	}
+
+	/**
+	 * Execute jobs immediately when debugging.
+	 */
+	public static function execute_jobs_in_debug_mode() {
+		if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {
+			$jobs = self::queue()->search(
+				array(
+					'per_page' => -1,
+					'status'   => 'pending',
+					'claimed'  => false,
+				)
+			);
+
+			foreach ( $jobs as $job ) {
+				$job->execute();
+			}
 		}
 	}
 }
--- a/includes/class-wc-admin-reports-sync.php
+++ b/includes/class-wc-admin-reports-sync.php
@@ -412,7 +412,7 @@ class WC_Admin_Reports_Sync {
 	 * Execute jobs immediately when debugging.
 	 */
 	public static function execute_jobs_in_debug_mode() {
-		if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {
+		if ( apply_filters( 'woocommerce_disable_scheduling', false ) ) {
 			$jobs = self::queue()->search(
 				array(
 					'per_page' => -1,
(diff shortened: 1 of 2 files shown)


timmyc (commented):
Big fan of this change - having to pause while testing new orders was getting a tad bit annoying.

So some thoughts: Instead of adding a new hook, and the new method of `execute_jobs_in_debug_mode` - would it make more sense to add in the `WP_DEBUG` check to the existing hooked method `schedule_single_order_process`, right after the check to ensure the post type is a `shop_order` do something like:

```
if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {
 self::orders_lookup_process_order( $order_id );
 return;
}
```

This would circumvent the queue altogether, and save some queries by not having to see if the job is there/pending. 

At least I think that would work, haven't tested it ;)

One other minor thought on the change is if we merge this in, then it would have an impact on a live site if someone like a developer forgot to turn WP_DEBUG off, but there would probably be bigger problems then the job running right away if that happened.

rrennick:
The downside of doing this is it would interfere with testing work with Action Scheduler (https://github.com/woocommerce/wc-admin/issues/1528#issuecomment-464152453). If this were implemented as a filter instead as a WP_DEBUG check we would be able to turn it off & on as needed.

jeffstieler:
> The downside of doing this is it would interfere with testing work with Action Scheduler ([#1528 (comment)](https://github.com/woocommerce/wc-admin/issues/1528#issuecomment-464152453)). If this were implemented as a filter instead as a WP_DEBUG check we would be able to turn it off & on as needed.

+1

joshuatf:
@rrennick Thanks for the input.  Replaced this with a filter in https://github.com/woocommerce/wc-admin/pull/1581/commits/84d9b9c462c97ba185441ad99bcfc4e58da2cc28.

@timmyc I like the idea of bypassing the scheduler altogether, but in order to do this, we need to run the action on `save_post` instead of `save_post_shop_order`, otherwise it's run too early and information is not updated until the next save.  @jeffstieler are there any issues with changing this back to `save_post`?

jeffstieler:
> in order to do this, we need to run the action on `save_post` instead of `save_post_shop_order`, otherwise it's run too early and information is not updated until the next save. @jeffstieler are there any issues with changing this back to `save_post`?

Are we dependent on other `save_post` hooks for this to work properly? `save_post_{$type}` is fired right before `save_post`: https://github.com/WordPress/WordPress/blob/master/wp-includes/post.php#L3940

rrennick:
@jeffstieler if you use `save_post` then all actions hooked into `save_post_shop_order` will have executed. If you use `save_post_shop_order` then you would need to give it a high priority number to ensure that other save hooks are fired before this one.

jeffstieler:
@rrennick @joshuatf oh duh. Time to make more ☕️ 😄

includes/class-wc-admin-reports-sync.php
```diff
@@ -132,6 +132,11 @@ public static function schedule_single_order_process( $order_id ) {
 			return;
 		}
 
+		if ( apply_filters( 'woocommerce_disable_scheduling', false ) ) {
```
jeffstieler: Optional nitpick: should this filter name include "order" since it only prevents order updates from being queued?

jeffstieler (approved):
On optional nitpick, otherwise this works as intended. Nice addition to the readme as well. 🛳 it!

includes/class-wc-admin-reports-sync.php
joshuatf: Hmm, good point.  My thought process was that we may want to use this same hook to easily allow disabling across wc-admin and maybe eventually core so I opted to keep it generic.

But since we don't have this anywhere else right now, it may be a good idea to make it more descriptive.  I'll update to `woocommerce_disable_order_scheduling` and we can modify in the future if we want to disable scheduling elsewhere.

PR #1582, opened 2019-02-15
Fix double focus on autocomplete search list after options are loaded

### LINKED ISSUE #1576: Table Filter: Results Shown after Fuzzy Filter Selected
**Describe the bug**
When interacting with a table filter, if a fuzzy search term is selected **before** the autocompleter options are returned from the REST API, the result panel is shown again, and the first option to "All Customer Names that start with" is missing the prior-selected search string.

customers-table-filter

**To Reproduce**

1. Navigate to the Customers Report
2. In the table filter, type in a search, and quickly pick the fuzzy search option before the REST API returns with results
3. Note the Results panel gets shown again after the REST request returns, even though you have already selected a fuzzy search.

**Expected behavior**
Since I have selected a fuzzy search - I don't expect the options to be shown like they are since I have already submitted my query.

joshuatf:
Fixes #1576 

Adds `isFocused` to state to prevent re-opening the options list after options are loaded asynchronously.

### Screenshots
feb-15-2019 17-30-24

1.  Type in a search in the Customers table search box.
2.  Quickly select an option for autocomplete options are loaded asynchronously.
3.  Note that the list does not re-open.

### DIFF
--- a/packages/components/src/search/autocomplete.js
+++ b/packages/components/src/search/autocomplete.js
@@ -44,6 +44,7 @@ function filterOptions( search, options = [], exclude = [], maxResults = 10 ) {
 export class Autocomplete extends Component {
 	static getInitialState() {
 		return {
+			isFocused: false,
 			search: /./,
 			selectedIndex: 0,
 			query: undefined,
@@ -156,7 +157,7 @@ export class Autocomplete extends Component {
 		const promise = ( this.activePromise = Promise.resolve(
 			typeof options === 'function' ? options( query ) : options
 		).then( optionsData => {
-			if ( ! optionsData ) {
+			if ( ! optionsData || ! this.state.isFocused ) {
 				return;
 			}
 			const { selected } = this.props;
@@ -209,7 +210,7 @@ export class Autocomplete extends Component {
 		// filter the options we already have
 		const filteredOptions = filterOptions( search, this.state.options, selected );
 		// update the state
-		this.setState( { selectedIndex: 0, filteredOptions, search, query } );
+		this.setState( { isFocused: true, selectedIndex: 0, filteredOptions, search, query } );
 		// announce the count of filtered options but only if they have loaded
 		if ( this.state.options ) {
 			this.announce( filteredOptions );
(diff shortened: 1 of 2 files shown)


packages/components/src/search/autocomplete.js
```diff
@@ -156,7 +157,7 @@ export class Autocomplete extends Component {
 		const promise = ( this.activePromise = Promise.resolve(
 			typeof options === 'function' ? options( query ) : options
 		).then( optionsData => {
-			if ( ! optionsData ) {
+			if ( ! optionsData || ! this.state.isFocused ) {
```
Aljullu: Could we use `state.query`, here? I think it would work as well and this way we wouldn't have to add a new property to the state.
```suggestion
			if ( ! optionsData || ! this.state.query ) {
```

Aljullu (commented):
I confirm this is working as expected. I added a suggestion in the code that might help simplify the code.

packages/components/src/search/autocomplete.js
joshuatf: Great idea- thanks for the suggestion @Aljullu!  Added in https://github.com/woocommerce/wc-admin/pull/1582/commits/b1a772d782241f21bd606e45dfd061003c66bbd4 and it seems to work just as well.

packages/components/src/search/style.scss
```diff
@@ -10,6 +10,13 @@
 		fill: $core-grey-light-900;
 	}
 
+	.woocommerce-tag {
+		max-width: 100%;
+	}
+	.woocommerce-tag .woocommerce-tag__text {
+		max-width: calc(100% - 24px);
+	}
+
```
Aljullu: These changes are duplicated in #1597, right?

Aljullu (approved):
Thanks for applying the feedback, this is looking good now! Just added a comment about two CSS declarations that are duplicated in another PR.

packages/components/src/search/style.scss
joshuatf: Pre ☕️ commit.  Removed so we can address separately in #1597 - thanks!

PR #1583, opened 2019-02-15
Fix variation name format to match core

emdashcodes:
We were replacing `-` with ` / ` in variation names, but not consistently in all places. This PR removes the replacement, which matches how the variations are presented on other pages like the order pages.

Before:

screen shot 2019-02-15 at 8 45 01 am

After:

screen shot 2019-02-15 at 8 58 51 am

screen shot 2019-02-15 at 9 02 27 am

To Test:
* View a variation and check the title.

### DIFF
--- a/client/analytics/report/products/table-variations.js
+++ b/client/analytics/report/products/table-variations.js
@@ -84,7 +84,7 @@ export default class VariationsReportTable extends Component {
 		return map( data, row => {
 			const { items_sold, net_revenue, orders_count, extended_info, product_id } = row;
 			const { stock_status, stock_quantity, low_stock_amount, sku } = extended_info;
-			const name = get( row, [ 'extended_info', 'name' ], '' ).replace( ' - ', ' / ' );
+			const name = get( row, [ 'extended_info', 'name' ], '' );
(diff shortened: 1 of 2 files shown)


timmyc (approved):
+1 on matching the formatting of core.

timmyc:
@justinshreve do you think we need to flag this for design review? Guessing the `/` separator came from designs at one point along the way. Makes sense to match what already exists though.

joshuatf (approved):
Also looks good to me 👍 

If we do end up shifting back to slashes for some reason, we may want to consider filtering these names directly in the API response or in core to avoid re-filtering all over the place.

justinshreve:
> do you think we need to flag this for design review? Guessing the `/` separator came from designs at one point along the way.

Yeah I think they came from the designs but were not implemented 100%. I also agree with @joshuatf that we should probably filter this differently -- especially to get it consistent in core.

I think we can go ahead and merge this so it's consistent in the feature plugin + consistent with core.

Pinging @LevinMedia or @josemarques for feedback. If we want to use slashes, we can reopen a new issue and change it everywhere.

LevinMedia:
Heya @justinshreve , thanks for the catch! Lets go with `-` all the way.

PR #1586, opened 2019-02-15
Screen readers: read date instead of data keys in 'time-comparison' bar charts

### LINKED ISSUE #1543: Screen readers read data keys instead of the date in `time-comparison` bar charts
- Change to a bar chart.
- Use a screen reader to navigate the bars.
- For the first bar of the screenshot below, something like `Quarter to Date (Jan 1 - Feb 13, 2019) 572` is read, while it should be `Week of January 1, 2019 572`.

image

Aljullu:
Fixes #1543.

Improve the text read when navigating `time-comparison` bar charts with a screen reader.

### Accessibility

### Screenshots
image

- Go to the _Revenue_ report.
- Change to a bar chart.
- Use a screen reader + keyboard to navigate the bars.
- When focusing a bar, verify it reads the date + value (something like `February 3, 2018 1,648.00€`).
- Switch the chart interval to `hour` (you might need to change the selected date range).
- Verify it reads the period + hour + value (something like: `Previous Year (Feb 15, 2018) 10PM 723.00€`).


packages/components/src/chart/d3chart/utils/bar-chart.js
```diff
@@ -68,7 +68,16 @@ export const drawBars = ( node, data, params, scales, formats, tooltip ) => {
...
 		.attr( 'aria-label', d => {
-			const label = params.mode === 'time-comparison' && d.label ? d.label : d.key;
+			let label = d.key;
+			if ( params.mode === 'time-comparison' ) {
+				if ( d.label ) {
+					label = d.label;
+				} else {
+					const dayData = data.find( e => e.date === d.date );
+					const date = tooltip.labelFormat( moment( dayData[ d.key ].labelDate ).toDate() );
+					label = params.interval === 'hour' ? `${ d.key } ${ date }` : date;
```
joshuatf: I think this is a nice touch.  One small oddity is that today vs yesterday is read as "Today" and "Previous Period."  I would think this would be either Today/Yesterday or Current Period/Previous Period.

We might also want to add something in for quarterly reports while we're here since the currently quarterly just reads the starting month and I imagine that may be confusing to users.

joshuatf (commented):
Nice additions here, @Aljullu.  This is testing well; I just left a couple notes about specific intervals we might me able to add on to.

On a somewhat unrelated note, do you feel the previous period should be displayed before the current period?  When looking at these charts, it appears revenue is going down at first glance if this year is higher than the previous.  There may be a design reason for this though.  /cc @LevinMedia @josemarques

packages/components/src/chart/d3chart/utils/bar-chart.js
Aljullu: Right, those labels are the same as the ones used in the legend. I updated the logic behind them to make them more specific (_Yesterday_, _Previous week_, etc.), instead of displaying _Previous Period_ for all of them.

image

Let me know what do you think.

Aljullu:
Thanks for the review @joshuatf! Would you like to take another look at the last commit?

> On a somewhat unrelated note, do you feel the previous period should be displayed before the current period?

Indeed, that makes a lot of sense to me.

LevinMedia:
@joshuatf @Aljullu I'll get back to you on the order shortly

> I would think this would be either Today/Yesterday or Current Period/Previous Period.

Initially, @psealock and I I tried using plain language to describe the comparison periods. In the end we opted for Previous Period and Previous Year, which are both standard amongst our competitors and Google Analytics. To make it even more explicit the date ranges were added to the periods as well.

Another reason why the “previous period” language works better: Once you get over the span of a week, the number of days in a period can be quite different. Last month vs Previous Period is a great example. April has 30 days, but March has 31. In this case, you’re comparing all 30 days of April to the last 30 days of March. (2nd – 31st). Google Analytics handles this in the same fashion.

>We might also want to add something in for quarterly reports while we're here since the currently quarterly just reads the starting month and I imagine that may be confusing to users.

I would LOVE to be able to figure something like this out... rather than showing months, I would prefer to show Q1, Q2 etc. - this poses a couple challenges:

Labeling the periods as Q1, Q2 etc could be problematic from an i18n perspective. I'm not sure how we translate that. Any ideas would be most welcome. 

The second issue was [D3 doesn't actually have built in support for labeling quarters](https://stackoverflow.com/questions/20355668/d3-js-how-to-format-tick-values-as-quarters-instead-of-months), we'd have to figure it out on our own. With all the other things we had going on, I left it alone.

packages/date/src/index.js
```diff
@@ -48,6 +48,19 @@ export const presetValues = [
 	{ value: 'custom', label: __( 'Custom', 'wc-admin' ) },
 ];
 
+export const previousPeriodValues = [
```
joshuatf: Nice!  Optional refactor here:  we could use an object here to quickly access all these labels.

```js
export const previousPeriodLabels = [
	today: __( 'Yesterday', 'wc-admin' ),
	yesterday: __( 'Previous Day', 'wc-admin' ),
	week: __( 'Last Week', 'wc-admin' ),
	last_week: __( 'Previous Week', 'wc-admin' ),
	month: __( 'Last Month', 'wc-admin' ),
	last_month: __( 'Previous Month', 'wc-admin' ),
	quarter: __( 'Last Quarter', 'wc-admin' ),
	last_quarter: __( 'Previous Quarter', 'wc-admin' ),
	year: __( 'Previous Quarter', 'wc-admin' ),
	last_year: __( 'Previous Year', 'wc-admin' ),
];
```

joshuatf (approved):
Looks great and feels like a huge improvement to accessibility!  Left an optional refactor for the previous period labels and pre-approving pending @LevinMedia's response to quarters and swapping the bar chart position on current/previous (both of which I think would be okay to handle in a separate PR).

>Trying to use plain language got messy when the selected date range was "to date" such as "Week to date, month to date" etc. Using the label "Previous period", along with the dates in the comparison range is less ambiguous than "Last Week To Date" which could easily be interpreted as last week, up to and including today's date.

Ah, that makes sense.  I think the way @Aljullu handled it in the latest commit is pretty descriptive and matches up okay, but let me know if you think that looks okay.

>Another reason why the “previous period” language works better: Once you get over the span of a week, the number of days in a period can be quite different. Last month vs Previous Period is a great example. April has 30 days, but March has 31. In this case, you’re comparing all 30 days of April to the last 30 days of March. (2nd – 31st). Google Analytics handles this in the same fashion.

I see that now.  That does throw a small wrench into this update.

>Labeling the periods as Q1, Q2 etc could be problematic from an i18n perspective. I'm not sure how we translate that. Any ideas would be most welcome.

I think when it comes to screen readers, we can either read out the date range for this quarter (e.g., October 1st, 2018 to December 31, 2018) or we could read the quarter number and year (e.g., 2018 Fourth Quarter).  Either of those sound like good options?

Aljullu:
> Trying to use plain language got messy when the selected date range was "to date" such as "Week to date, month to date" etc. Using the label "Previous period", along with the dates in the comparison range is less ambiguous than "Last Week To Date" which could easily be interpreted as last week, up to and including today's date.

Thanks for the input @LevinMedia! Could you take a look at my [last changes](https://github.com/woocommerce/wc-admin/pull/1586/commits/34d13deaa982425ca2cba76dacfd2c89c96f06c0#diff-421b158400874b347758f137ec615452R51) in this branch, then? I replaced _Previous Period_ with more descriptive labels (_Yesterday_, _Previous Day_, _Last Week_, _Previous Week_...):

image

But from what I understand in your comment, that would not be a good solution here?

ryelle:
Why does the hourly view change the format to include the period? It ends up adding a lot of repeated content to the screen reader as you tab through the items. `Previous Year (Feb 15, 2018) 10PM 723.00€` could just be `10PM Feb 15, 2018 723.00€`?

I used VoiceOver + Safari to test this with both a bar chart and a line chart, daily and hourly. All read as described, except for the line chart hourly – that just read the hour, ex "11PM $10.00"

I noticed the hover card for hourly (line & bar) could use the corresponding date, too:

screen shot 2019-02-19 at 11 34 01 am

There's also an issue getting VoiceOver into the bar chart, for some reason <kbd>VO</kbd> + <kbd>→</kbd> skips right over the section. It works to get into the line chart regions, but the bar chart doesn't use regions in the same way. Without spending some time with the code I'm not sure I can suggest anything specific though.

Aljullu:
@LevinMedia and I had a chat, today. We will use _Previous Period_ for all periods as we did until now. The rationale behind this is that when comparing months and quarters, some days of the previous month/quarter might be selected as well (for example, comparing December, which has 31 days, with November, which has 30 days). In those cases _Last Month_ and _Last Quarter_ would not be accurate.
image

That wouldn't be a problem for days and weeks, but in order to keep consistency, we will use _Previous Period_ in all cases.

LevinMedia:
>could just be `10PM Feb 15, 2018 723.00€`?

@ryelle @Aljullu I like that suggestion

Aljullu:
> Why does the hourly view change the format to include the period? It ends up adding a lot of repeated content to the screen reader as you tab through the items. `Previous Year (Feb 15, 2018) 10PM 723.00€` could just be `10PM Feb 15, 2018 723.00€`?
> 
> I used VoiceOver + Safari to test this with both a bar chart and a line chart, daily and hourly. All read as described, except for the line chart hourly – that just read the hour, ex "11PM $10.00"
> 
> I noticed the hover card for hourly (line & bar) could use the corresponding date, too:

@ryelle Good points. All of this is fixed in the last commit: 919edd0c4ecbc93df3e0328fe88c055e9507e257.

> There's also an issue getting VoiceOver into the bar chart, for some reason <kbd>VO</kbd> + <kbd>→</kbd> skips right over the section. It works to get into the line chart regions, but the bar chart doesn't use regions in the same way. Without spending some time with the code I'm not sure I can suggest anything specific though.

Right, in the line chart each line is a region, while in the bar chart each date is a region, so navigation is different between them.

---

This PR should be ready for another review.

timmyc (approved):
Looking and working well for me @Aljullu - could we get a note added to the appropriate changelogs - but pre-approving.

PR #1591, opened 2019-02-15
implement WC_ADMIN_ABSPATH in wc-admin.php

### LINKED ISSUE #1590: Implement WC_ADMIN_ABSPATH
**Describe the bug**
`WC_ADMIN_ABSPATH` is defined but never used https://github.com/woocommerce/wc-admin/blob/master/wc-admin.php#L19:L21

rrennick:
Fixes #1590

Implement `WC_ADMIN_ABSPATH`

- Load the dashboard
- Click through to a few different views and reports.

### DIFF
--- a/includes/class-wc-admin-api-init.php
+++ b/includes/class-wc-admin-api-init.php
@@ -33,108 +33,108 @@ class WC_Admin_Api_Init {
 	 */
 	public function init_classes() {
 		// Interfaces.
-		require_once dirname( __FILE__ ) . '/interfaces/class-wc-admin-reports-data-store-interface.php';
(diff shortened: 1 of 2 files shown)


rrennick:
@timmyc Should I also update the ones starting at https://github.com/woocommerce/wc-admin/blob/master/includes/class-wc-admin-api-init.php#L36 (`includes/` would need to be prepended to the file path string) ?

timmyc:
> Should I also update the ones starting at https://github.com/woocommerce/wc-admin/blob/master/includes/class-wc-admin-api-init.php#L36

Sure sounds good to me - do you want to do that here or in a follow-up PR?

rrennick:
Thanks @timmyc

PR #1597, opened 2019-02-18
Fix long search tags

### LINKED ISSUE #1575: Advanced Filters: Long Token-Field Names Hidden in Narrow Viewports
**Describe the bug**
When an advanced filter of the token-field variety ( i.e. Products selected ) has a long name/value, the token field overflows off the screen and can not be fully seen. Furthermore, this prevents the (x) from being clicked on to remove a single selection... the only option is to use the "Clear All Filters" and start over.

image

**To Reproduce**
Steps to reproduce the behavior on an emulator, phone, or resized browser window

1. Have a product with a really long name ( make a few, they are handy! )
2. Open the Products Report, Select Comparison, Pick your long-named product
3. Notice the token/tag overflows the container

**Expected behavior**
The ability to fully see the selected items, and remove them if desired.

joshuatf:
Fixes #1575

In narrow viewports, this cuts off long tags to use only ellipsis.

### Screenshots
screen shot 2019-02-18 at 3 46 24 pm
screen shot 2019-02-18 at 3 49 02 pm

1.  Create a really unnecessarily long, lengthy, elongated, protracted, stretched out, and never-ending product name.
2.  Narrow your viewport 📱 
3.  Check the advanced filters and add the newly created product name to make sure it is cut off with an ellipsis and that the remove button is still visible.
4.  Check an inline search (top of products table) to make sure the tag is also visible here and that the remove all button is not overlapped.

@LevinMedia @josemarques Is the Gutenberg-way of using ellipsis okay here?  Or do we want to not limit the height of tags and show the full name?

### DIFF
--- a/packages/components/src/search/style.scss
+++ b/packages/components/src/search/style.scss
@@ -2,6 +2,11 @@
 
 .woocommerce-search {
 	position: relative;
+	min-width: 0;
+
+	>div {
+		min-width: 0;
+	}
 
 	.woocommerce-search__icon {
 		position: absolute;
--- a/packages/components/src/search/style.scss
+++ b/packages/components/src/search/style.scss
@@ -10,6 +10,14 @@
 		fill: $core-grey-light-900;
 	}
 
+	.woocommerce-tag {
+		max-width: 100%;
+	}
+
+	.woocommerce-tag .woocommerce-tag__text {
+		max-width: calc( 100% - 24px );
+	}
+
 	&:not(.has-inline-tags) {
 		.woocommerce-tag {
 			margin: 8px 6px 0 0;
@@ -25,7 +33,7 @@
 
 	.woocommerce-search__inline-container {
 		width: 100%;
-		padding: 4px 2px 4px 36px;
+		padding: 4px 36px 4px;
 		border: 1px solid $core-grey-light-700;
 		background-color: $white;
 		display: flex;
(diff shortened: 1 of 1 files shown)


packages/components/src/search/style.scss
```diff
@@ -10,6 +10,14 @@
 		fill: $core-grey-light-900;
 	}
 
+	.woocommerce-tag {
+		max-width: 100%;
+	}
+
+	.woocommerce-tag .woocommerce-tag__text {
+		max-width: calc( 100% - 24px );
```
Aljullu: Our linter (`npm run lint:css`) will complain about the spaces inside the parenthesis.
```suggestion
		max-width: calc(100% - 24px);
```

Aljullu (commented):
Great seeing this fixed! :slightly_smiling_face: 

Code looks good and I confirm it works well and as expected, but I couldn't reproduce this screenshot with inline tags:

> screen shot 2019-02-18 at 3 46 24 pm

This is what I get:
image

Adding something like:
```SCSS
.woocommerce-search {
  width: calc(100% - $gap);
}
```
Seems to fix that:
image

joshuatf:
@Aljullu Thanks for the review!

I couldn't quite duplicate the same layout in your screenshot.  I did find a similar one though which I was able to resolve on my end by setting:

```css
.woocommerce-search {
	min-width: 0;
}
```
screen shot 2019-02-19 at 10 48 19 am
screen shot 2019-02-19 at 11 25 23 am


Can you let me know if that works for the overflow you're seeing as well?  If not, could you give me the viewport width you're seeing that at?  Thanks!

Aljullu (approved):
It looks like it has been fixed as well. :+1:

PR #1600, opened 2019-02-18
Reduce net order total by refunded amount less tax and shipping refunds

### LINKED ISSUE #1555: Orders net total doesn't reflect refunds
**Describe the bug**
When a refund is given on an order, the order net total remains unchanged.

**To Reproduce**

1. Refund any amount on an order.
2.  Check the net total in the Orders report.
3.  Note the net total remains unchanged.

**Expected behavior**
Expected the net total to reflect any refunds given.

**Additional context**
Note that refunds may include shipping or taxes as well and the net total is already reduced by those values.  We may want to only apply against those items for line item refunds (not general order refunds). 
 Further discussion in this thread - p90Yrv-Vb-p2

joshuatf:
Fixes #1555 

Reflects refunds in net total without refunds towards shipping or tax.

### Screenshots
screen shot 2019-02-18 at 5 33 44 pm
screen shot 2019-02-18 at 5 33 26 pm

1.  Create an order with shipping and taxes.
2.  Refund the line items on that order with separate amounts for shipping and tax.
3.  Note the refund amount is correct (does not include refunds towards tax or shipping).
4.  Make a manual refund without entering a line item number.
5.  Make sure the refund amount is correct.
6.  Make a manual refund greater than the possible net_total.
7.  Make sure `net_total` is 0.


timmyc (approved):
Code looks good and testing out well for me.

PR #1601, opened 2019-02-18
Don't show variations table when there is an active search

Aljullu:
When searching for a variable product in the table search, we don't want to change the autocompleter neither the table rows from products to variations. This PR fixes that.

### Screenshots
_Before:_
image

_After:_
image

- Search for a product that you know is variable.
- Verify the table is still showing products instead of variations (the easiest way is to check the first column header is `Product Title` instead of `Product / Variation Title`).


psealock (approved):
Nice fix here, working well!

PR #1602, opened 2019-02-18
Change fixed legend button height to min-height

### LINKED ISSUE #1573: Chart: Long Names in Chart Legend Wrap Lines
When viewing a chart that has a legend displayed on the left side ( i.e. Product Comparison mode ), items with long names wrap onto multiple lines and make the legend a bit difficult to read.

legend-overflow

__Steps To Reproduce__
- Edit a product to have an abnormally long name
- Open the Products report, select Product Comparison from the advanced filters
- Pick your long named product along with some others
- Note the wrapping of the product name to multiple lines, and how the value is a bit crowded

__Desired Solution__
Not sure what the best treatment is here. We have yet to introduce a clipping pattern like exists in Calypso where long names/titles receive a fade on the end, and don't overflow. Perhaps a different approach would be to make the legend more grid/table like, or simply adjust the margin of the elements.

/cc @josemarques @LevinMedia

joshuatf:
Fixes #1573 

Sets a minimum height instead of fixed for legend buttons and left-aligns the text.

### Before
screen shot 2019-02-18 at 5 55 32 pm

### After
screen shot 2019-02-18 at 5 54 32 pm

1.  Add a really long product name.
2.  Use the advanced filter to compare multiple products, including the long-name product.
3.  Check that the product name is legible and doesn't overlap other columns.

### DIFF
--- a/packages/components/src/chart/d3chart/legend.scss
+++ b/packages/components/src/chart/d3chart/legend.scss
@@ -142,8 +142,9 @@
 		padding: 0;
 
 		& > button {
-			height: 36px;
+			min-height: 36px;
 			padding: 0 17px;
+			text-align: left;
 		}
 
 		&:first-child {


Aljullu (approved):
Looks good! :shipit: 

Edit: I pushed a couple of commits to this branch by mistake. :sweat_smile:  I [undid](https://github.com/woocommerce/wc-admin/compare/0fea0bc1f0a147b9c19d3169b25ab9e198ab0a32..1661b23a96aebd38ca0f3d07dd2eaba99bcb7e4a) them, so there shouldn't be any problem, but wanted to point it out in case you see something weird.

PR #1606, opened 2019-02-18
Preserve chart colors when changing visible keys in charts with less keys than 'selectionLimit'

### LINKED ISSUE #1604: `time-comparison` charts shouldn't change colors when activating/deactivating legend items
**Steps to reproduce:**
- Deactivate a chart legend item.
- Notice the other item changes its color.

chart-colors

**Expected result:**
- Color should be preserved.

Aljullu:
Fixes #1604.

Avoids lines/bars changing its color when one legend item is deactivated in `time-comparison` charts.

### Screenshots
_Before:_
chart-colors

_After:_
chart-colors-

* Deactivate a chart legend item.
* Verify the active item didn't change its color.

### DIFF
--- a/packages/components/src/chart/d3chart/chart.js
+++ b/packages/components/src/chart/d3chart/chart.js
@@ -28,6 +28,7 @@ import {
 import { drawAxis } from './utils/axis';
 import { drawBars } from './utils/bar-chart';
 import { drawLines } from './utils/line-chart';
+import { getColor } from './utils/color';
 import ChartTooltip from './utils/tooltip';
 
 /**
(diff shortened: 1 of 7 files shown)


jeffstieler:
Looks good for legend items, but colors still change for comparisons.

Aljullu:
> Looks good for legend items, but colors still change for comparisons.

Thanks for taking a look @jeffstieler! I added a [new commit](05a471adcce7da55eef186f0f89f830b5dad4a8d) that extends this PR to comparison charts. It doesn't affect charts that have more data items than the `selectionLimit`, in those case we **do** want to change the colors depending on the visible items to avoid two keys having very similar colors (see #1481).

jeffstieler (approved):
Code looks good and this is testing well on both chart legend items and comparisons. 👍

Issue #1617, opened 2019-02-19
Customer Report: Broken filters


justinshreve:
Using `master`, I can't get the following advanced filters to work on the customers report:

* Username
* Country
* Name
* Email

Other number/date based filters seem to work.

**To Reproduce**
* Go to the customers report.
* Try adding one of the above filters, autocompleting a name, and hitting filter.
* Notice that the results do not change / the same number of customers is returned.

psealock:
Some prior art here:
https://github.com/woocommerce/wc-admin/issues/1387

PR #1618, opened 2019-02-19
Fix customer filters and download user filtering

### LINKED ISSUE #1614: Downloads Report Table missing User Name value
**Describe the bug**

The downloads report table never displays "user name". (The endpoint provides `user_id` but not `username` or customer name)

This missing value also breaks CSV downloading (console error).

**EDIT:** This is caused by a plugin that disabled the `/wp/v2/users` endpoints - solution is to embed the necessary product/user data in the `/v4/reports/customers/` endpoint rather than rely on embeds from other namespaces.

Found testing #1523.

**To Reproduce**

1. Go to Downloads Report (where there is data)
2. Verify that User Name column is empty for all rows

**Expected behavior**

I expect the customer's name or username to display.

**Screenshots**

screen shot 2019-02-18 at 4 30 06 pm

### LINKED ISSUE #1616: Downloads Report: Username filter is totally broken
**Describe the bug**

Username filter on downloads report searches/displays inconsistently. Search is performed is an exact match on the `username` field, which isn't useful. Once a customer **username** is selected, the **email** is displayed in the token, but then the Customer ID is conflated with the WP User ID and the email changes to the wrong user.

It also seems that filtering by this selected user doesn't work. (Perhaps the `user_includes` param doesn't work for this report?)

Found testing #1523.

**To Reproduce**

1. Go to Downloads Report
2. Add a username advanced filter
3. Type the exact username of a customer
4. Verify that the username appears in results list
5. Click username for selection
6. Verify that email is displayed for a moment, then changes to another email.

**Expected behavior**

I expect to be able to fuzzy search on username and have the username be the value consistently displayed to me. I also expect that the username won't change to the incorrect user.

**Screenshots**

**Additional context**

Using WCCOM test data.

emdashcodes:
Fixes #1616, fixes #1614.

This PR does the following:

* A new `searchby` parameter is added, allowing us to fuzzy search name, email, and username. Customer autocompletors are updated to use the new search.
* Renames the `customers` parameter `include`. This is the parameter used in the autocompletor to pull in label IDs and matches other endpoints.
* Fixes an issue where `user_includes`/`user_excludes` was not passed along to one of the downloads controllers.

* Run `phpunit` and make sure all tests run.
* Go to Downloads Report
* Add a username advanced filter
* Type a partial username
* See options appear
* Select one and filter the download results
* Verify results update
* Try switching the option to exclude, see options update
* Go to the customers report
* Try the same thing for name and email (Note that filtering does not actually work here or on master #1617).

### DIFF
--- a/includes/api/class-wc-admin-rest-reports-customers-controller.php
+++ b/includes/api/class-wc-admin-rest-reports-customers-controller.php
@@ -61,7 +61,7 @@ class WC_Admin_REST_Reports_Customers_Controller extends WC_REST_Reports_Control
 		$args['avg_order_value_max'] = $request['avg_order_value_max'];
 		$args['last_order_before']   = $request['last_order_before'];
 		$args['last_order_after']    = $request['last_order_after'];
-		$args['include']             = $request['include'];
+		$args['customers']           = $request['customers'];
 
 		$between_params_numeric    = array( 'orders_count', 'total_spend', 'avg_order_value' );
 		$normalized_params_numeric = WC_Admin_Reports_Interval::normalize_between_params( $request, $between_params_numeric, false );
@@ -458,7 +458,7 @@ class WC_Admin_REST_Reports_Customers_Controller extends WC_REST_Reports_Control
 			'format'            => 'date-time',
 			'validate_callback' => 'rest_validate_request_arg',
 		);
-		$params['include']               = array(
+		$params['customers']               = array(
 			'description'       => __( 'Limit result to items with specified customer ids.', 'wc-admin' ),
 			'type'              => 'array',
 			'sanitize_callback' => 'wp_parse_id_list',
--- a/includes/api/class-wc-admin-rest-reports-customers-controller.php
+++ b/includes/api/class-wc-admin-rest-reports-customers-controller.php
@@ -47,6 +47,7 @@ class WC_Admin_REST_Reports_Customers_Controller extends WC_REST_Reports_Control
 		$args['orderby']             = $request['orderby'];
 		$args['match']               = $request['match'];
 		$args['search']              = $request['search'];
+		$args['searchby']            = $request['searchby'];
 		$args['username']            = $request['username'];
 		$args['email']               = $request['email'];
 		$args['country']             = $request['country'];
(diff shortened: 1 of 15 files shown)


includes/api/class-wc-admin-rest-reports-customers-controller.php
```diff
@@ -60,7 +61,7 @@ protected function prepare_reports_query( $request ) {
 		$args['avg_order_value_max'] = $request['avg_order_value_max'];
 		$args['last_order_before']   = $request['last_order_before'];
 		$args['last_order_after']    = $request['last_order_after'];
-		$args['customers']           = $request['customers'];
+		$args['include']             = $request['include'];
```
jeffstieler: Remove the `customers` parameter breaks the table search on the customers report.

jeffstieler:
It seems that the `user_includes` parameter is being given the Customer ID from the search, not the User ID which it expects - meaning the filter isn't working despite the autocomplete now working as expected.

includes/api/class-wc-admin-rest-reports-customers-controller.php
justinshreve: Thanks for pointing that out, I tried to see where this was used but didn't see anything when I looked.

Instead of updating this to `include` here, I've only updated the parameter on the main `/wc/v4/customers` endpoint, which matches other endpoints (i.e. /wc/v4/coupons?include is used for the main endpoint and `/wc/v4/reports/coupons/?coupons=... is used for filtering)

jeffstieler: I think this is where it's used: https://github.com/woocommerce/wc-admin/blob/463200547c242bf38131b234fe5b9713b6a4268a/client/wc-api/reports/utils.js#L42

justinshreve:
@jeffstieler Thanks for testing this -- I definitely jumbled up the user/customer IDs a bit and it didn't help that one of my users had the same ID for both :). I made some updates in https://github.com/woocommerce/wc-admin/pull/1618/commits/a962f765df61a54511184db3127b2bbae8741ae7.

After considering the options, I ended up updating the download report to just filter based on customer ID. All the filters return the customer ID at the moment, and the customers report (only other place these filters are used) uses customer ID.

I also addressed #1616 here.

Do you mind taking another look?

jeffstieler (approved):
Code looks good and this tests well. Nicely done! 🛳 it.

DilHem:
Hey All,
Its Urgent,

Can anyone provide me C language code for fuzzy logic speed controller. I'm doing this Using ENERGIA interface. 
I'm taking my inputs using a ultrasonic sensor and after fuzzy methods the outputs are taken by DC motors as their inputs and then after they are going to control the speed or the rotation speed of DC motor.

I'm thankful if anyone can provide the C language code for this.

Thank You

PR #1619, opened 2019-02-19
Refactor customer ID creation

### LINKED ISSUE #1263: Registered users do not return a customer ID
**Describe the bug**
#1144 added in a customers data store that tracks stats for both guests and registered customers 🎉

This should create unique customer IDs for all guest and regular customers in the `wc_customers_lookup` table.  Some orders in the `wc_orders_stats` table appear to have a `customer_id` of `0`.

These orders at first glance appear to be associated with already created accounts, so it looks like customer IDs may be bypassed when users are not created directly through WooCommerce (e.g., admins, manual user creation, or other sign-up forms).

This blocks conditions don't pass and a number is not returned:
https://github.com/woocommerce/wc-admin/blob/36a6b53aa262c25b5e2c4c1730b55e3bf0311bcf/includes/data-stores/class-wc-admin-reports-orders-data-store.php#L495-L498

It's possible that the user creation never passed by any of these hooks:
https://github.com/woocommerce/wc-admin/blob/36a6b53aa262c25b5e2c4c1730b55e3bf0311bcf/includes/data-stores/class-wc-admin-reports-customers-data-store.php#L71-L74

**To Reproduce**
1.  Create orders programmatically or rebuild all orders that were associated with an admin account.
2.  Note the `customer_id` is not created for those accounts and associated orders have a `customer_id` of `0`.

**Expected behavior**
Expected a `customer_id` associated with this account and respective orders.

**Screenshots**
screen shot 2019-01-09 at 4 14 24 pm

### LINKED ISSUE #1476: Returning customer definition for guest customers

### LINKED ISSUE #1552: Customers Report Shows non-customer data
**Describe the bug**
When viewing the customers report, admin accounts on the site are shown. I suppose if an admin account has orders associated with it, then they should be shown, but otherwise kind of feels like non Woo user roles shouldn't be shown... i.e. Authors/Contributors should not be displayed on the customers report.

**To Reproduce**

1. Go to the Customers Report
2. Note admin non/customer accounts are shown

joshuatf:
Fixes #1552
Fixes #1476
Fixes #1263 

Refactors the customer ID getting/creation to:
* Only store users with a `customer` role or at least 1 order (admins aren't stored unless they've placed orders).
* Link existing guest customer IDs by email
* Prevent orphaned orders when a guest email isn't present.

#### User types
1.  Add a new admin and update its user profile.
2.  Check to make sure it hasn't been added to the Customers data.
3.  Make an order with that admin.
4.  Check that the admin is now shown in customer data.
5.  Create a customer through WooCommerce or user with the role `customer`.
6.  Check that the customer is already in the data despite not having any orders.

#### Guests
1.  Make an order with a guest.
2.  Use that same email again in another order also without logging in.
3.  Check that the customer ID is the same for both of those orders.

#### Email-less orders
Testing this may require dropping the customers lookup table to allow null values for emails.
1.  Using the checkout fields plugin, disable the email field.
2.  Checkout using a guest account.
3.  Check that a customer and order is made in the lookup tables and that the customer ID matches in the order.

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-customers-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-customers-data-store.php
@@ -414,52 +414,77 @@ class WC_Admin_Reports_Customers_Data_Store extends WC_Admin_Reports_Data_Store 
 	}
 
 	/**
-	 * Gets the guest (no user_id) customer ID or creates a new one for
-	 * the corresponding billing email in the provided WC_Order
+	 * Returns an existing customer ID for an order if one exists.
 	 *
-	 * @param WC_Order $order Order to get/create guest customer data with.
-	 * @return int|false The ID of the retrieved/created customer, or false on error.
+	 * @param object $order WC Order.
+	 * @return int|bool
 	 */
-	public function get_or_create_guest_customer_from_order( $order ) {
+	public static function get_existing_customer_id_from_order( $order ) {
+		$user_id = $order->get_customer_id();
+
+		if ( 0 === $user_id ) {
+			$email = $order->get_billing_email( 'edit' );
+
+			if ( $email ) {
+				return self::get_guest_id_by_email( $email );
+			} else {
+				return false;
+			}
+		} else {
+			return self::get_customer_id_by_user_id( $user_id );
+		}
+	}
+
+	/**
+	 * Get or create a customer from a given order.
+	 *
+	 * @param object $order WC Order.
+	 * @return int|bool
+	 */
+	public static function get_or_create_customer_from_order( $order ) {
 		global $wpdb;
+		$returning_customer_id = self::get_existing_customer_id_from_order( $order );
 
-		$email = $order->get_billing_email( 'edit' );
-
-		if ( empty( $email ) ) {
-			return false;
+		if ( $returning_customer_id ) {
+			return $returning_customer_id;
 		}
 
-		$existing_guest = $this->get_guest_by_email( $email );
-
-		if ( $existing_guest ) {
-			return $existing_guest['customer_id'];
-		}
-
-		$result = $wpdb->insert(
-			$wpdb->prefix . self::TABLE_NAME,
-			array(
-				'first_name'       => $order->get_billing_first_name( 'edit' ),
-				'last_name'        => $order->get_billing_last_name( 'edit' ),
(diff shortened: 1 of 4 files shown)


includes/data-stores/class-wc-admin-reports-customers-data-store.php
```diff
@@ -474,32 +499,7 @@ public function get_guest_by_email( $email ) {
...
-			$wpdb->prepare(
-				"SELECT * FROM {$table_name} WHERE user_id = %d LIMIT 1",
-				$user_id
-			),
-			ARRAY_A
-		); // WPCS: unprepared SQL ok.
-
-		if ( $customer ) {
-			return $this->cast_numbers( $customer );
+			return $guest_row['customer_id'];
```
jeffstieler: We should only query for `customer_id` if that's all we're returning. Also this needs to be casted to an int.

jeffstieler (approved):
Pre-approving (SQL change requested).

This looks good and tests well with the provided instructions. Additionally, I deleted the customer row I had for the `admin` user and it was not recreated when re-running the reports generation. 👍

includes/data-stores/class-wc-admin-reports-customers-data-store.php
joshuatf: Nice!  Thanks for catching that.  Updated in https://github.com/woocommerce/wc-admin/pull/1619/commits/75a315fb947a05ea5d1c360cd35ade1ad9c2a868

PR #1620, opened 2019-02-19
Add 'isCheckbox' prop to 'MenuItem'

### LINKED ISSUE #1609: Report Chart: column filter can't be toggled ON with keyboard
**Describe the bug**

When toggling column display using the keyboard, columns can be toggled OFF but not ON.

Found as part of #1525.

**To Reproduce**

1. Go to any report
2. Tab to columns menu (vertical ellipse) 
3. Toggle a column off with keyboard
4. Verify the column can't be toggled back on with the keyboard

**Expected behavior**

I expect to be able to toggle columns ON with the keyboard.

**Screenshots**

* OS: Mac OS 10.14
* Browser: tested Chrome, FF, IE

Aljullu:
Fixes #1609.

- Adds a new prop `isCheckbox` to `<MenuItem>` that allows creating menu items with a `FormToggle` and [`role=menuitemcheckbox`](https://www.w3.org/TR/wai-aria-1.1/#menuitemcheckbox).
- Updates all menus that were using `<MenuItem>` with [`role=menuitem`](https://www.w3.org/TR/wai-aria-1.1/#menuitem) and had a checkbox inside that, to use the new prop.

### Accessibility

### Screenshots
image

* Go to any report.
* Tab to columns menu (vertical ellipse).
* Verify only menu items are focusable but toggle buttons are not (before, you needed two Tabs to jump from one option to the next).
* Toggle a column off with the keyboard.
* Verify the column can be toggled back on with the keyboard.
* Repeat the same process with the _Dashboard_ menus: _Store Performance_, _Leaderboards_ and _Charts_.


jeffstieler:
I was unable to toggle the columns using the "enter" key.. 😕

packages/components/src/ellipsis-menu/menu-item.js
```diff
@@ -20,26 +20,28 @@ class MenuItem extends Component {
...
 			event.preventDefault();
 			onInvoke();
 		}
 	}
 
 	onKeyDown( event ) {
-		if ( event.keyCode === ENTER || event.keyCode === SPACE ) {
+		const { onInvoke } = this.props;
+
+		if ( onInvoke && ( event.keyCode === ENTER || event.keyCode === SPACE ) ) {
```
psealock: Oddly, this isn't working on `ENTER` for me either.

packages/components/src/table/index.js
```diff
@@ -323,7 +323,7 @@ class TableCard extends Component {
 								return null;
 							}
 							return (
-								<MenuItem key={ key } onInvoke={ this.onColumnToggle( key ) }>
+								<MenuItem key={ key } tabIndex={ null }>
```
psealock: What does the `null` here do? Will this revert to the default prop? Is there a way to be more explicit (or remove altogether)?

psealock (commented):
Looking good and testing well. Just two comments

packages/components/src/ellipsis-menu/menu-item.js
joshuatf: Looks like `onInvoke` is undefined so this condition isn't passing.

joshuatf (commented):
Tabbing is testing well.  Left a comment about `onInvoke` not being defined which seems to be why the enter key doesn't work.

Aljullu:
@jeffstieler @psealock @joshuatf thanks for the reviews! Good catch with the Enter key not working.

I reverted the previous commits and started this PR from scratch with a broader goal to fix the same issue also for the _Dashboard_ menus.

We were rendering checkboxes inside a `role=menuitem` element, which I think is not correct or, at least, not the best option here. Instead, with the new changes I converted those menu items to `role=menuitemcheckbox` and I added `aria-hidden` to the checkbox. That should fix #1609, while still allowing interaction with the Enter key.

I updated the intro post of this PR with a new title and test instructions.

jeffstieler:
I don't know if this needs to be debounced or what, but rapidly toggling twice causes anywhere from 2-4 toggles to occur:



_(I realize it's hard to tell what's happening here - I'm toggling the control twice with either the space or enter key)_

ryelle (commented):
Testing with keyboard-only and with Safari + VoiceOver– works mostly as described, there's a glitch with arrow keys though. You should be able to arrow-down through the items, but there are two focus points per item, the menu item and then the toggle itself - and when focused on the toggle, <kbd>enter</kbd> doesn't trigger the action. I'm assuming this has something to do with how the `NavigableMenu` manages key focus. Not sure this is enough of a bug to prevent merging this, but thought I'd mention in case it's an easy fix :)

Aljullu:
@ryelle thanks for testing it! Arrow keys navigation should be fixed in e240d466a8d26029de58231bef06cde8d6428d0c.

> I don't know if this needs to be debounced or what, but rapidly toggling twice causes anywhere from 2-4 toggles to occur:

@jeffstieler I'm not sure to understand what the problem is here. Do you mean we should disable the toggle buttons while it's updating?

joshuatf:
>I don't know if this needs to be debounced or what, but rapidly toggling twice causes anywhere from 2-4 toggles to occur:

This is caused by updating the state and the subsequent (slightly delayed) update that's received when the props are modified after a successful API request.

Since we're loading all settings into the initial state, we can probably safely skip updating the state within `componentDidUpdate()` and only modify the state if an error is received.  That might be worth opening a separate issue for; I don't mind outlining the problem and/or handling that if that's the direction we want to go.

/cc @jeffstieler @Aljullu

Aljullu:
Good point, I changed that behavior in the last commit (2dd093da7c4d3e55b0101d8d211c379c90596c75), so `<Table>` doesn't update the `showCols` state when the `headers` prop changes.

jeffstieler:
Looks like the rapid toggling issue was fixed with 2dd093d 👍

timmyc:
It seems all the issues found in review have either been handled, or follow-up issues opened. As such I'm marking this one as ready to merge.

timmyc (approved):
:shipit:

PR #1627, opened 2019-02-19
Adds StoreAlerts component

### LINKED ISSUE #115: Component: Store Alerts (aka Priority Notices)
While the Sidebar/Activity panel boasts a new "Inbox" to communicate key items to users - the latest [Dashboard Design](https://sketch.cloud/s/rQMvE/all/prototype/dashboard/play) displays the new Store Alert section where critical/show stopper alerts will be displayed to users:

storenotices

While I believe we still need some more in-depth designs to fully vet the functionality and display of these Store Alerts, I think we can begin on building the basic component that they will call home.

__Features__
- Store Alerts should be able to consume an array alerts, or possibly child components ( up to the implementor )
- An individual alert currently appears to support a severity gridicon, a title, content, a call to action button, and the ability to "snooze" the alert.
- Arrows in the upper right corner allow users to toggle between the alerts in the panel.

__Design Questions__
- Will there be differing severity levels for the alerts? If so what are they and the corresponding icons and colors for the left border.
- Will any alerts have more than one action button?
- When the final alert has been acted upon, will we display a success message, or simply collapse the panel from view?

/cc @jameskoster

tiagonoronha:
Fixes #115 

Basic component, still in development.

screenshot 2019-02-19 at 21 10 40

### DIFF
--- a/client/layout/index.js
+++ b/client/layout/index.js
@@ -67,7 +67,7 @@ class Layout extends Component {
 				<TransientNotices />
 
 				<div className="woocommerce-layout__primary" id="woocommerce-layout__primary">
-					<StoreAlerts />
+					{ window.wcAdminFeatures[ 'store-alerts' ] && <StoreAlerts /> }
 
 					<Notices />
 					{ ! isEmbeded && (
--- a/client/layout/index.js
+++ b/client/layout/index.js
@@ -22,6 +22,7 @@ import Header from 'header';
 import Notices from './notices';
 import { recordPageView } from 'lib/tracks';
 import TransientNotices from './transient-notices';
+import StoreAlerts from './store-alerts';
 
 class Layout extends Component {
 	componentDidMount() {
@@ -66,6 +67,8 @@ class Layout extends Component {
 				<TransientNotices />
 
 				<div className="woocommerce-layout__primary" id="woocommerce-layout__primary">
+					<StoreAlerts />
+
 					<Notices />
 					{ ! isEmbeded && (
 						<div className="woocommerce-layout__main">
--- a/client/layout/store-alerts/index.js
+++ b/client/layout/store-alerts/index.js
@@ -64,8 +64,10 @@ const alertTypes = {
 class StoreAlerts extends Component {
 	constructor( props ) {
 		super( props );
+		const { alerts } = this.props;
+
 		this.state = {
-			currentAlert: 0,
+			currentAlert: alerts ? 0 : null,
 		};
 
 		this.previousAlert = this.previousAlert.bind( this );
@@ -74,7 +76,8 @@ class StoreAlerts extends Component {
 	}
 
 	totalAlerts() {
-		return dummy.length;
+		const { alerts } = this.props;
+		return alerts.length;
 	}
 
 	previousAlert( event ) {
@@ -100,8 +103,9 @@ class StoreAlerts extends Component {
 	}
 
 	render() {
+		const { alerts } = this.props;
 		const { currentAlert } = this.state;
-		const alert = dummy[ currentAlert ] ? dummy[ currentAlert ] : null;
+		const alert = alerts[ currentAlert ] ? alerts[ currentAlert ] : null;
 		const type = alertTypes[ alert.type ] ? alertTypes[ alert.type ] : null;
 		const className = classnames( 'woocommerce-store-alerts', {
 			'is-alert-emergency': type && 'emergency' === alert.type,
@@ -158,3 +162,7 @@ class StoreAlerts extends Component {
 }
 
 export default StoreAlerts;
+
+StoreAlerts.defaultProps = {
+	alerts: dummy || [],
+};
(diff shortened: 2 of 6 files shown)


client/layout/store-alerts/index.js
```diff
@@ -0,0 +1,153 @@
...
+		this.state = {
+			currentAlert: 0,
+		};
+
+		this.previousAlert = this.previousAlert.bind( this );
+		this.nextAlert = this.nextAlert.bind( this );
+		this.totalAlerts = this.totalAlerts.bind( this );
+	}
+
+	totalAlerts() {
```
timmyc: I'm thinking eventually the array of `alerts` will be passed in as a prop, so we could probably update the logic in the component to be ready for that. We could create a `defaultProps` that passes in your `dummy` instance as the `alerts` prop. And you could then use `this.props.alerts` here

client/layout/store-alerts/index.js
```diff
@@ -0,0 +1,153 @@
...
+			} );
+		}
+	}
+
+	render() {
+		const { currentAlert } = this.state;
+		const alert = dummy[ currentAlert ] ? dummy[ currentAlert ] : null;
+		const type = alertTypes[ alert.type ] ? alertTypes[ alert.type ] : null;
+		const className = classnames( 'woocommerce-store-alerts', {
+			'is-alert-emergency': type && 'emergency' === alert.type,
```
timmyc: Instead of checking for `type` here and having `alert.type` in each line, should we just assign `alert.type` to a const somewhere after line 103?

client/layout/store-alerts/index.js
```diff
@@ -0,0 +1,153 @@
...
+								onClick={ this.previousAlert }
+								disabled={ 0 === currentAlert }
+								label={ __( 'Previous Alert', 'wc-admin' ) }
+							/>
+							<span
+								className="woocommerce-store-alerts__pagination-label"
+								role="status"
+								aria-live="polite"
+							>
+								{ currentAlert + 1 } { __( 'of', 'wc-admin' ) } { this.totalAlerts() }
```
timmyc: This might need to use interpolateComponents for a proper translation setup. Here is an [example](https://github.com/woocommerce/wc-admin/blob/master/packages/components/src/filters/advanced/date-filter.js#L69)

client/layout/store-alerts/index.js
```diff
@@ -0,0 +1,153 @@
...
+
+		return (
+			<Card
+				title={ [
+					type && <Dashicon key="icon" icon={ type.icon } />,
+					<Fragment key="title">{ alert.title }</Fragment>,
+				] }
+				className={ className }
+				action={
+					this.totalAlerts() > 1 && (
```
timmyc: If moved to props, we could assign the length of the alerts array to a const and use it throughout here.

`const numberOfAlerts = this.props.alerts.length;`

tiagonoronha:
@timmyc thanks for the feedback! I made a few changes, hopefully it simplified things :)

client/layout/store-alerts/index.js
```diff
@@ -0,0 +1,179 @@
...
+									onClick={ this.previousAlert }
+									disabled={ 0 === currentIndex }
+									label={ __( 'Previous Alert', 'wc-admin' ) }
+								/>
+								<span
+									className="woocommerce-store-alerts__pagination-label"
+									role="status"
+									aria-live="polite"
+								>
+									{ interpolateComponents( {
```
timmyc: Nice!

client/layout/index.js
```diff
@@ -66,6 +67,8 @@ class Layout extends Component {
 				<TransientNotices />
 
 				<div className="woocommerce-layout__primary" id="woocommerce-layout__primary">
+					<StoreAlerts />
```
timmyc: Prior to merging this in, I think we should add a feature flag around this so it only shows up in Dev mode. #1500 has some examples on how that is done.

client/layout/store-alerts/index.js
```diff
@@ -0,0 +1,179 @@
...
+									onClick={ this.nextAlert }
+									disabled={ numberOfAlerts - 1 === currentIndex }
+									label={ __( 'Next Alert', 'wc-admin' ) }
+								/>
+							</div>
+						)
+					}
+				>
+					<div className="woocommerce-store-alerts__message">{ alert.message }</div>
+					<Button isPrimary className="woocommerce-store-alerts__button" href={ alert.action.url }>
```
LevinMedia: Can you change this to `isDefault` ? 

There's some funky override going on in there somewhere, when I change it to `isDefault` the normal state button text is white, when it should be $core-grey-dark-500`

LevinMedia: I'm also seeing the same issues on mobile that are noted above.

I'd like the left margin to match the rest of the content (it should be 16px on the smallest screens) and position the segmented button above the alert title? 

screen shot 2019-02-21 at 2 23 44 pm

tiagonoronha:
@LevinMedia @timmyc 

Made a few changes :)

screenshot 2019-02-21 at 23 38 33

screenshot 2019-02-21 at 23 43 49

Also added the feature flag!

tiagonoronha:
@LevinMedia full context:

screenshot 2019-02-22 at 11 38 37

screenshot 2019-02-22 at 11 39 24

LevinMedia:
Thanks @tiagonoronha looking good!

PR #1628, opened 2019-02-19
Components: Add SearchListControl

### LINKED ISSUE #417: Move SearchListControl to woocommerce/components
Other people should be able to use this base component. Let's move it to the @woocommerce/components npm package. This is a tracking issue to create a pr into [wc-admin.](https://github.com/woocommerce/wc-admin)

It will need…

- Documentation
- A DevDocs example
- Another a11y review
- Check in IE11

ryelle:
Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/417 – port over the SearchListComponent, which is a base component used to create all the searchable + selectable lists in our blocks. This PR adds the components, doc file, and a devdocs entry.

I also fixed the components changelog, since it somehow flipped back to saying 1.5.0 was unreleased.

### Accessibility

### Screenshots

The block on its own doesn't handle any data requests, it expects a list of data, and a subset of selected data.

Example used in a block, populated with category content.
block

And the basic example in devdocs:
devdocs

- Check out the component in devdocs: `/wp-admin/admin.php?page=wc-admin#/devdocs/search-list-control`

To test "in action" in the blocks project:

- On this branch, after building, move to the components directory
- Run `npm link` to set up a symlink for the @woocommerce/components package
- Grab this branch of the woo blocks: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/442
- In the blocks project folder, run `npm link @woocommerce/components` to point this project's dependency to the linked package
- Build the blocks project, `npm run build` (note, **don't** run `npm install` between the link and build commands, it'll blow away your link)
- Add "Products by Category", "Products by Attribute", "Hand-picked Products", or a "Featured Product" block to your post, all use controls based on the `SearchListControl`.
- Expect: these should work as usual, no visual or functional changes.

### DIFF
--- a/packages/components/src/search-list-control/index.js
+++ b/packages/components/src/search-list-control/index.js
@@ -1,0 +1,311 @@
+/**
+ * External dependencies
+ */
+import { __, _n, sprintf } from '@wordpress/i18n';
+import {
+	Button,
+	MenuGroup,
+	Spinner,
+	TextControl,
+	withSpokenMessages,
+} from '@wordpress/components';
+import { Component, Fragment } from '@wordpress/element';
+import { compose, withInstanceId, withState } from '@wordpress/compose';
+import { escapeRegExp, findIndex } from 'lodash';
+import Gridicon from 'gridicons';
+import PropTypes from 'prop-types';
+
+/**
+ * Internal dependencies
+ */
+import { buildTermsTree } from './hierarchy';
+import SearchListItem from './item';
+import Tag from '../tag';
+
+const defaultMessages = {
+	clear: __( 'Clear all selected items', 'woo-gutenberg-products-block' ),
+	list: __( 'Results', 'woo-gutenberg-products-block' ),
+	noItems: __( 'No items found.', 'woo-gutenberg-products-block' ),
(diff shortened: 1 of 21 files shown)


packages/components/src/search-list-control/index.js
```diff
@@ -0,0 +1,311 @@
...
+
+/**
+ * Internal dependencies
+ */
+import { buildTermsTree } from './hierarchy';
+import SearchListItem from './item';
+import Tag from '../tag';
+
+const defaultMessages = {
+	clear: __( 'Clear all selected items', 'woo-gutenberg-products-block' ),
```
timmyc: Should the text domain change to `wc-admin` throughout all the files in the PR?

timmyc (approved):
Code looked quite familiar so this made for a quick review for me - one minor comment about text domain but otherwise looking grand to me. Did the advanced test against the blocks branch so will leave a 👍 over there too. Pre-approving.

packages/components/src/search-list-control/index.js
ryelle: Yeah, my mistake – fixed it 👍

PR #1652, opened 2019-02-21
Fix null values in CSV columns export

### LINKED ISSUE #1463: Trashed orders are reflected in stats

### LINKED ISSUE #1643: Product and customer report fail to download CSV
**Describe the bug**
Visiting the product or customer reports without adding any query parameters results in an error being thrown when clicking "Download."  Adding any filters seems to allow downloading.

**To Reproduce**

1. Go to the products or customers report.
2. Click "Download."
3. Note the console error.

**Expected behavior**
Expected the CSV to download without fail.

**Screenshots**
screen shot 2019-02-20 at 4 51 13 pm

joshuatf:
Fixes #1643 

Uses an empty string in CSV columns when a value is null.  Some columns may be null by default, so this safeguards against failed downloads.

1. Find a report with a null value (e.g., customer sign up date for guests should be null by default).
2. Try to export a report with the above row.
3. Note the successful export and empty column when opening the CSV file.

### DIFF
--- a/packages/csv-export/CHANGELOG.md
+++ b/packages/csv-export/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 1.1.0
+
+- Fix null values in columns
+
 # 1.0.0
 
 - Released package
--- a/packages/csv-export/CHANGELOG.md
+++ b/packages/csv-export/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 1.1.0
 
-- Fix null values in columns
+- Fix error in `getCSVRows` when there is a null or undefined column value
 
 # 1.0.0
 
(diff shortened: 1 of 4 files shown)


packages/csv-export/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 1.1.0
+
+- Fix null values in columns
```
Aljullu: I wonder if this message could be improved a bit? _Fix JS error in `getCSVRows` when there is a null or undefined value_ or something like that?

Aljullu (approved):
I could reproduce the issue and confirm it's fixed now. Pre-approving and added a suggestion to improve the CHANGELOG message.

packages/csv-export/CHANGELOG.md
joshuatf: Ah yes, much better!  Thanks @Aljullu

PR #1653, opened 2019-02-21
Update table header layout to use flexbox

### LINKED ISSUE #1649: Table Header Issues in Mobile Safari
**Describe the bug**
When viewing wc-admin in Mobile Safari ( emulator os 10.1 ), I'm seeing some display issues in the header of tables:

__Orders__
image

__Products__
image

__Customers__
image

**To Reproduce**
Using an iOS device or the simulator that ships with xcode, open up any of the reports noted above and view the table header area.

* Device: iPhone
* OS: 10.1
* Browser mobile safari

joshuatf:
Fixes #1649 

Replaces grid styles with flexbox to support older browser.

### Before
screen shot 2019-02-21 at 3 21 14 pm

### After
screen shot 2019-02-21 at 3 19 34 pm

Check that the styles have not regressed and look correct on wide and narrow viewports for:

* Modern browsers
* iOS <=10.2
* IE11 (Note there is a bug in IE11 that will prevent the page from loading if `WP_SCRIPT_DEBUG` is enabled - https://core.trac.wordpress.org/ticket/46110 )


packages/components/src/card/style.scss
```diff
@@ -18,26 +18,21 @@
 .woocommerce-card__header {
 	padding: ($gap - 3) $gap;
 	border-bottom: 1px solid $core-grey-light-700;
-	display: grid;
+	display: flex;
 	align-items: center;
 
-	.has-action & {
-		grid-template-columns: auto 1fr;
+	>* {
```
Aljullu: I think you could use the class `.woocommerce-card__header-item`, so there is no need to use the universal selector here.
```suggestion
	> .woocommerce-card__header-item {
```

packages/components/src/table/index.js
```diff
@@ -300,39 +300,41 @@ class TableCard extends Component {
...
-							isLink
-						>
-							<DowloadIcon />
-							<span className="woocommerce-table__download-button__label">
-								{ labels.downloadButton || __( 'Download', 'wc-admin' ) }
-							</span>
-						</IconButton>
-					),
 				] }
 				menu={
```
Aljullu: Thinking out loud, I wonder if we should rename this prop if we are going to use it to display other components than the menu itself.

Related to that, I get a JS error because of a PropType mismatch in `<Card>`:
`Warning: Failed prop type: Invalid menu passed to Card. Must be EllipsisMenu`

packages/components/src/table/style.scss
```diff
@@ -6,96 +6,58 @@
...
 	}
 
 	.woocommerce-card__action {
-		justify-self: flex-end;
-		margin: -($gap - 3) 0;
-	}
-
-	.woocommerce-card__menu {
-		justify-self: flex-end;
+		margin-left: auto;
```
Aljullu: This seems to remove the space between the table title and the `Compare` button.

With:
image

Without:
image

Aljullu (commented):
I took a first look and overall it's looking good, I just added some comments in the code.

We are using CSS grids in several other parts of `wc-admin`. Do you think we will need to convert those to flexbox as well?

packages/components/src/card/style.scss
joshuatf: This seems like a safe commit to make since there doesn't seem to be any way of adding an item without `.woocommerce-card__header-item` so I'll go ahead and add this.

But curious if there are any reasons not to use the universal selector here since we would probably want any items directly under the header to share these styles.

packages/components/src/table/index.js
joshuatf: Updated the prop type and description here to fix the error.

>Thinking out loud, I wonder if we should rename this prop if we are going to use it to display other components than the menu itself.

I had a similar thought and then came to the conclusion that `menu` was still a descriptive area to where it was being displayed and actions can be a part of a menu area.  That said, I'm open to other names here if anyone has suggestions.  Maybe we change this to `menuArea` and `actionArea` or `leftAction` and `rightAction` (though that would be inaccurate on mobile).

packages/components/src/table/style.scss
joshuatf: Good eyes!  👀 

Updated in https://github.com/woocommerce/wc-admin/pull/1653/commits/f2f4a8e8d754f8d8035d8f5947c81acff530f006

joshuatf:
Thanks for the review, @Aljullu.

>We are using CSS grids in several other parts of `wc-admin`. Do you think we will need to convert those to flexbox as well?

I think it's probably a smart move for the time being.  Mobile Safari browsers versions from a couple years ago and Android browsers still didn't quite have support for grids yet.  This is primarily an issue in iOS devices since updating the browser version requires updating the entire OS and users with older hardware are probably avoiding the upgrade.

I opened other issues for areas using grids that I noticed had problems on these devices:
https://github.com/woocommerce/wc-admin/issues/1656
https://github.com/woocommerce/wc-admin/issues/1655
https://github.com/woocommerce/wc-admin/issues/1654

packages/components/src/card/style.scss
Aljullu: Well, the universal selector is usually considered not to be very efficient:
https://css-tricks.com/efficiently-rendering-css/#article-header-id-1

In addition, in the future we might want to add a child to `.woocommerce-card__header` that is absolutely positioned or has [`display: contents`](https://caniuse.com/#search=display%3A%20contents). In those cases, if we use a class that must be manually specified instead of the universal selector, we will not need to reset the styles.

packages/components/src/table/index.js
Aljullu: > I had a similar thought and then came to the conclusion that `menu` was still a descriptive area to where it was being displayed and actions can be a part of a menu area. That said, I'm open to other names here if anyone has suggestions. Maybe we change this to `menuArea` and `actionArea` or `leftAction` and `rightAction` (though that would be inaccurate on mobile).

Fair enough.

Aljullu:
> I think it's probably a smart move for the time being. Mobile Safari browsers versions from a couple years ago and Android browsers still didn't quite have support for grids yet. This is primarily an issue in iOS devices since updating the browser version requires updating the entire OS and users with older hardware are probably avoiding the upgrade.
> 
> I opened other issues for areas using grids that I noticed had problems on these devices:
> #1656
> #1655
> #1654

Just saw Justin's comment here: https://github.com/woocommerce/wc-admin/issues/1656#issuecomment-466401801. Do you think we should proceed merging this PR, then?

timmyc:
That is a really good point @Aljullu and @justinshreve re: the iOS support numbers. Since this is functioning in IE11, perhaps we should stick to targeting support for iOS 11/12 - but curious to hear other thoughts on the matter.

joshuatf:
@justinshreve @timmyc Continued the discussion in #1656.

ryelle:
As part of a sweep of `Components` PRs, I landed on this one– is this still worth merging?

joshuatf:
Closing this PR for now.  Looks like we might address issues with grid support later on.

PR #1664, opened 2019-02-21
add action scheduler group

rrennick:
See #1528

Move new scheduled actions to a `wc-admin-data` action group. This is the first piece of implementing scheduled action priority. The group will serve to provide some of the functionality of queues. 

- Click `Rebuild Reports` in WC -> Status -> Tools
- Switch to the `Scheduled Actions` tab
- Filter the `Pending` actions
- Note that the `Action Group` for the WC Admin actions is `wc-admin-data`
- Load the front end of the site
- Repeat every 1-2 minutes until all the report build actions have completed


rrennick:
@justinshreve Thanks, I've updated the unit test.

PR #1670, opened 2019-02-22
Add product name filtering

### LINKED ISSUE #1635: Product search is missing results when content contains keywords
**Describe the bug**
When searching for a product, if another product contains the keywords in the product content, but not the title, these results will not show in the search list.

This seems to be due to the `search` parameter returning results for the keywords in the product name or content and the client-side list filtering out results that don't match.

**To Reproduce**

1. Add a keyword to a product name (e.g. add "Lorem" to a product title).
2.  Add at least 10 newer products with the search term in their content, but not the title (e.g., add "lorem" to the product content of 10 products).
3.  Search for the keyword (go to the Products table and type in "lorem").
4.  Note the lack of results, but results present in API response 😱 

**Expected behavior**
Expected to see results for **product names** matching my search query.

**Screenshots**
screen shot 2019-02-20 at 1 07 46 pm
screen shot 2019-02-20 at 1 07 11 pm
screen shot 2019-02-20 at 1 06 54 pm

joshuatf:
Fixes #1635 

Allows filtering by product name in search results instead of searching entire post content.

1.  Go to the products report and use the product comparison advanced filter.
2.  Search for a product name.
3.  Make sure the results returned are associated with the product name/title and not the content or excerpt.

### DIFF
--- a/includes/class-wc-admin-api-init.php
+++ b/includes/class-wc-admin-api-init.php
@@ -26,9 +26,6 @@ class WC_Admin_Api_Init {
 
 		// Add currency symbol to orders endpoint response.
 		add_filter( 'woocommerce_rest_prepare_shop_order_object', array( __CLASS__, 'add_currency_symbol_to_order_response' ) );
-
-		// Add WP Query where filters.
-		add_filter( 'posts_where', array( __CLASS__, 'add_wp_query_post_title_filter' ), 10, 2 );
 	}
 
 	/**
@@ -409,26 +406,6 @@ class WC_Admin_Api_Init {
 		$response->set_data( $response_data );
 
 		return $response;
-	}
-
-	/**
-	 * Add post title searching to WP Query
-	 *
-	 * @param string $where Where clause used to search posts.
-	 * @param object $wp_query WP_Query object.
-	 * @return string
-	 */
-	public static function add_wp_query_post_title_filter( $where, $wp_query ) {
-		global $wpdb;
-
-		$post_title_search = $wp_query->get( 'post_title__like' );
-		if ( $post_title_search ) {
-			$post_title_search = $wpdb->esc_like( $post_title_search );
-			$post_title_search = ' \'%' . $post_title_search . '%\'';
-			$where            .= ' AND ' . $wpdb->posts . '.post_title LIKE ' . $post_title_search;
-		}
-
-		return $where;
 	}
 }
 
--- a/includes/class-wc-admin-api-init.php
+++ b/includes/class-wc-admin-api-init.php
@@ -418,7 +418,7 @@ class WC_Admin_Api_Init {
 	 * @param object $wp_query WP_Query object.
 	 * @return string
 	 */
-	public function add_wp_query_post_title_filter( $where, &$wp_query ) {
+	public static function add_wp_query_post_title_filter( $where, $wp_query ) {
 		global $wpdb;
 
 		$post_title_search = $wp_query->get( 'post_title__like' );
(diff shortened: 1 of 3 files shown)


includes/class-wc-admin-api-init.php
```diff
@@ -26,6 +26,9 @@ public function __construct() {
 
 		// Add currency symbol to orders endpoint response.
 		add_filter( 'woocommerce_rest_prepare_shop_order_object', array( __CLASS__, 'add_currency_symbol_to_order_response' ) );
+
+		// Add WP Query where filters.
+		add_filter( 'posts_where', array( __CLASS__, 'add_wp_query_post_title_filter' ), 10, 2 );
```
justinshreve: There were some concerns around this in the Bookings REST API: https://github.com/woocommerce/woocommerce-bookings/pull/2204/files#diff-51b120c3d76d89518ebbc32f22a7a0a0R21.

Although the logic will return early, that check still needs to happen for each query. I also feel like this side-steps data stores -- but I know that these REST API endpoints are older and access posts directly and it would be a bigger undertaking to refactor those so it's probably not worth it at this point. We might want to log a `@todo` for that though.

--

In the meantime, could we do something like 

```
	public function get_items( $request ) {
		// Add Filter
		$response = parent::get_items( $request );
		// Remove Filter
		return $response;
	}
```

in the products controller?

Example where we do the opposite in comments: https://github.com/woocommerce/woocommerce/blob/e1255de71709d9e7cd8caa143c56fb4d60ab7101/includes/api/v2/class-wc-rest-order-notes-v2-controller.php#L66-L70

emdashcodes (changes requested):

includes/class-wc-admin-api-init.php
joshuatf: >Although the logic will return early, that check still needs to happen for each query.

This check does happen for each `WP_Query` made on the site, but I think any performance issues are very likely to be negligible even under heavy load given this is just checking if an argument is set.  Is the concern here that a conflict might be made with the `post_title__like` argument?

>I also feel like this side-steps data stores -- but I know that these REST API endpoints are older and access posts directly and it would be a bigger undertaking to refactor those so it's probably not worth it at this point.

Agreed that this would be a pretty big undertaking.  The way `WC_REST_CRUD_Controller` is laid out seems to act as controller, query, and data store all in one.

>We might want to log a `@todo` for that though.

What is the task you were thinking we needed to log here?  If separating into a data store, I think we'd want to handle that directly in WC as this would touch on a large number of rest endpoints.

joshuatf: Added this change in https://github.com/woocommerce/wc-admin/pull/1670/commits/fda44b80bbe5770b4cfca1aa42cdc763642470cb and it does seem to work well, but still curious about the question above if this is really for performance or more of an issue with encapsulation.

justinshreve: > Added this change in [fda44b8](https://github.com/woocommerce/wc-admin/commit/fda44b80bbe5770b4cfca1aa42cdc763642470cb) and it does seem to work well, but still curious about the question above if this is really for performance or more of an issue with encapsulation.

That's a good point. A simple check probably wouldn't have a huge performance impact. I do think the recent update is the way to go since it's more targeted for when we need it.

> What is the task you were thinking we needed to log here? If separating into a data store, I think we'd want to handle that directly in WC as this would touch on a large number of rest endpoints.

We don't need to log anything for `wc-admin`. I was mainly thinking out-loud since the endpoints using the CRUD Controller seem function a bit differently from the other ones, and it seems like maybe it would be hard to swap those out -- but I don't have to worry about it right now.

emdashcodes (approved):
Thanks for the change here. The code is looking good, and this is testing well for me 👍.

joshuatf:
@justinshreve Thanks for the review!

>I do think the recent update is the way to go since it's more targeted for when we need it.

I think you're right; this safer for the time being even if it doesn't affect performance.

PR #1671, opened 2019-02-22
Add tax code filter to tax REST API

### LINKED ISSUE #1639: Search query is ignored in tax REST API
**Describe the bug**
Searching for tax codes to compare on the tax report does not use the search term, but instead returns a list of tax codes (presumably ordered by id or tax class) limited to 10 results.  This will be problematic if more than 10 exist.

There doesn't seem to be a `search` collection param for the tax rest controller.

**To Reproduce**

1. Create more than 10 tax codes.
2. Search for tax codes in the comparison filter on the tax report.
3. Note that you won't be able to find all tax codes.

**Expected behavior**
Expected to be able to find tax codes by search.

**Screenshots**
screen shot 2019-02-20 at 4 06 56 pm

joshuatf:
Fixes #1639 

Adds in a `code` parameter to the tax rest controller so we can search for taxes.

Note that `get_items()` will be removed if https://github.com/woocommerce/woocommerce/pull/22813 gets merged.

1.  Create more than 10 tax codes.
2.  Go to the tax report and search using the tax comparison filter.
3.  Check that the response is accurately searched.
4.  Check the ability to search by the full code (country-state-name-priority).

### DIFF
--- a/includes/api/class-wc-admin-rest-taxes-controller.php
+++ b/includes/api/class-wc-admin-rest-taxes-controller.php
@@ -25,6 +25,14 @@ class WC_Admin_REST_Taxes_Controller extends WC_REST_Taxes_Controller {
 	protected $namespace = 'wc/v4';
 
 	/**
+	 * Constructor
+	 */
+	public function __construct() {
+		add_filter( 'woocommerce_rest_tax_query', array( __CLASS__, 'add_tax_code_query_args' ), 10, 2 );
+		add_filter( 'woocommerce_rest_tax_query_string', array( __CLASS__, 'add_tax_code_filter' ), 10, 2 );
+	}
+
+	/**
 	 * Get all taxes and allow filtering by tax code.
 	 *
 	 * @todo This is mostly copied from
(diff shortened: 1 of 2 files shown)


includes/api/class-wc-admin-rest-taxes-controller.php
```diff
@@ -24,4 +24,162 @@ class WC_Admin_REST_Taxes_Controller extends WC_REST_Taxes_Controller {
...
+		if ( $request['code'] ) {
+			$prepared_args['code'] = $request['code'];
+		}
+		return $prepared_args;
+	}
+
+	/**
+	 * Filter tax codes by code.
+	 *
+	 * @param string $query Sql query string.
```
Aljullu: Nit-picking here, but could we capitalize SQL?
```suggestion
	 * @param string $query SQL query string.
```

Aljullu (approved):
This is looking good to me. I added some small comments, but pre-approving.

joshuatf:
@Aljullu Thanks for reviewing!

Since https://github.com/woocommerce/woocommerce/pull/22813 is closed, it looks like we'll be overriding `get_items()` directly and adding this in.

Removed the hooks and truncated `@todo` comment as well.

PR #1672, opened 2019-02-22
Add tax include param to tax REST API

### LINKED ISSUE #1640: Taxes Core REST API: Add collection param 'include'
**Describe the bug**
When adding tax items through the comparison filter, refreshing the page (and occasionally on adding new items) does not get the correct labels.

This bug is caused by the `getLabels()` call to the tax controller since the tax controller does not inherit from `WC_REST_CRUD_Controller` there is no collection param for `include` to filter results.

**To Reproduce**

1. Add a couple tax labels in the comparison filter (not in order of database IDs).
2. Click "Compare."
3. Refresh the page.
4. Note that the labels are incorrect.

**Expected behavior**
Expected the correct labels to be retrieved.

**Screenshots**
feb-20-2019 16-21-11

joshuatf:
Fixes #1640 

Adds an `include` param to the tax rest endpoint so we can filter by tax rate IDs and return the correct tax codes.

### Screenshots
screen shot 2019-02-22 at 5 04 55 pm
screen shot 2019-02-22 at 5 04 51 pm

1.  Go to the tax report and filter by tax comparison.
2.  Add various tax codes and check that the label is correct even after refreshing.
3.  Test calls to the rest endpoint `wp-json/wc/v4/taxes` with the `include` param and a comma separated list of rate IDs.

### DIFF
--- a/client/analytics/report/taxes/config.js
+++ b/client/analytics/report/taxes/config.js
@@ -3,17 +3,11 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import apiFetch from '@wordpress/api-fetch';
-import { find } from 'lodash';
-
-/**
- * WooCommerce dependencies
- */
-import { getIdsFromQuery, stringifyQuery } from '@woocommerce/navigation';
 
 /**
  * Internal dependencies
  */
+import { getRequestByIdString } from 'lib/async-requests';
 import { getTaxCode } from './utils';
 import { NAMESPACE } from 'wc-api/constants';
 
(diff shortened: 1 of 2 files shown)


Aljullu (commented):
Before merging this one, we should revert the changes added in #1571 for the taxes `getLabels`, right?
https://github.com/woocommerce/wc-admin/pull/1571/files#diff-2243561b44936466721f610b90a69c51R67

I think it would make sense rebasing this PR with the latest master.

joshuatf:
>Before merging this one, we should revert the changes added in #1571 for the taxes `getLabels`, right?

@Aljullu Definitely.  Rebased, reverted that, and also updated the includes filters since it looks like we're going to be directly overriding `get_items()` instead of using hooks based on https://github.com/woocommerce/wc-admin/pull/1671.

Aljullu (approved):
Thanks for rebasing @joshuatf! Code is looking good and it works as expected.

PR #1673, opened 2019-02-22
Fix CSS linter errors

emdashcodes:
Running `npm run lint` caught a couple of CSS errors. This PR fixes those.

Note: I did not update a change log entry. I don't think we need to since it's just a code style change and the CSS gets minified anyway.

To Test:
* Run `npm run lint` and make sure no errors are caught
* Bonus points for visually looking at the Search or SearchList components in devdocs


Aljullu (approved):
This is looking good.

:shipit:

PR #1679, opened 2019-02-22
Table: reset search on compare and reset comparison on search

### LINKED ISSUE #1660: Table comparison should reset search
**To Reproduce**
2. Using the table search, search for three products.
3. Select the checkbox next to two of them and click on '_Compare_'.
4. Notice all of them are compared.

**Expected behavior**
Only the selected ones should be compared.

**Screenshots**
compare-3

Aljullu:
Fixes #1660.

### Screenshots
reset-search

- Go to the _Products_ report, for example.
- Search for three products.
- Select the checkboxes of two of them and click on _Compare_.
- Verify the search is reset and only the compared products appear in the table.
- Now, search for another product.
- Verify the comparison is reset.

### DIFF
--- a/packages/components/src/table/index.js
+++ b/packages/components/src/table/index.js
@@ -160,7 +160,7 @@ class TableCard extends Component {
 	}
 
 	onSearch( values ) {
-		const { compareParam } = this.props;
+		const { compareParam, searchBy } = this.props;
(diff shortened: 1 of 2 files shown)


timmyc (approved):
Verified this is working as expected.

PR #1682, opened 2019-02-23
Display table placeholder instead of empty table when loading search terms

### LINKED ISSUE #1676: Table flashes to an empty table when loading search terms
**Describe the bug**
An empty table (instead of table placeholder) is displayed when loading search terms.

**To Reproduce**
2. Search for a product using the table search input.
3. Notice the table becomes empty, then switches to the table placeholder and then shows the results.

**Expected behavior**
The table shouldn't be empty at any moment.

**Screenshots**
empty-table

Aljullu:
Fixes #1676.

Prevent the report table showing an empty state when loading the search terms, instead, display the table placeholder.

### Screenshots
_Before (notice an empty table is visible for a fraction of a second):_
empty-table

_After:_
search-3

- Go to the _Products_ report, for example.
- Search for a product using the table search input.
- Verify the table is never empty: it goes from displaying all products → placeholder → displaying the searched product.
- Repeat the same for the other reports that have search inputs in the tables.

### DIFF
--- a/client/analytics/report/categories/table.js
+++ b/client/analytics/report/categories/table.js
@@ -134,7 +134,7 @@ class CategoriesReportTable extends Component {
 	}
 
 	render() {
-		const { query } = this.props;
+		const { isRequesting, query } = this.props;
 
 		const labels = {
 			helpText: __( 'Select at least two categories to compare', 'wc-admin' ),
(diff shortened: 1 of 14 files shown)


client/analytics/report/categories/table.js
```diff
@@ -177,9 +178,9 @@ export default compose(
 		};
 
 		const categories = getItems( 'categories', tableQuery );
-		const isError = Boolean( getItemsError( 'categories', tableQuery ) );
-		const isRequesting = isGetItemsRequesting( 'categories', tableQuery );
+		const isCategoriesError = Boolean( getItemsError( 'categories', tableQuery ) );
```
timmyc: 👍 like the added clarity to these variable names.

timmyc (approved):
Great fix Albert - tested out on Customers and Categories - all working as anticipated.

PR #1688, opened 2019-02-25
Don't update state on user preferences props change

### LINKED ISSUE #1687: Prevent flash in user preferences when toggling quickly
**Describe the bug**
When a user preference (e.g., dashboard chart type or visible charts) is toggled on/off quickly, a flash occurs as the props are received after updating.  Originally noted here - https://github.com/woocommerce/wc-admin/pull/1620#issuecomment-465676217

We can probably forego updating from props and refer to the state as the source of truth since the initial state can be set by server hydrated user data.

**To Reproduce**

1. Go to the dashboard reports.
2. Quickly toggle on/off various charts.
3.  Note the flash after the user data is updated.

**Expected behavior**
Expected no flash.

**Screenshots**
feb-25-2019 18-57-24

joshuatf:
Fixes #1687

Prevents toggle flash when quickly toggling user preferences.  Relies on initial state hydration instead of updating the state whenever props are changed after user preferences update.

1.  Toggle the user dashboard charts, chart types, leaderboard, and leaderboard rows quickly on and off.
2.  Check that no flash is produced when toggling quickly.
3.  Make sure preferences are still correct and persist after refreshing.

### DIFF
--- a/client/dashboard/dashboard-charts/index.js
+++ b/client/dashboard/dashboard-charts/index.js
@@ -7,7 +7,7 @@ import classNames from 'classnames';
 import { Component, Fragment } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import Gridicon from 'gridicons';
-import { isEqual, xor } from 'lodash';
+import { xor } from 'lodash';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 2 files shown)


Aljullu (approved):
Thanks for fixing that in the _Dashboard_, @joshuatf! Changes look good to me.

PR #1694, opened 2019-02-26
Display chart and summary numbers placeholder when loading search terms

### LINKED ISSUE #1684: Chart displays 'No data for the current search' for a fraction of a second when loading search terms
Similar to #1676.

**Describe the bug**
The chart displays the message `No data for the current search` when using the table search. That's because it gets no data but doesn't have an `isRequesting` prop.

**To Reproduce**
2. Search for a product using the table search input.
3. Notice the chart displays `No data for the current search` for a fraction of a second before the placeholder is shown.

**Expected behavior**
The message should not be shown before knowing if there are results or not.

**Screenshots**
empty-table

Aljullu:
Fixes #1684.
Follow-up of #1682.

Prevent the report summary numbers and chart showing an empty state when loading the search terms, instead, display the placeholders.

### Accessibility

### Screenshots
loading-chart

* Go to the _Products_ report, for example.
* Search for a product using the table search input.
* Verify the chart never shows the _No data for the current search_ message and summary numbers are never 0. Instead, they go from displaying all products data → placeholder → displaying the searched product data.
* Repeat the same for the other reports that have search inputs in the tables and display summary numbers/charts.

### DIFF
--- a/client/analytics/components/report-chart/index.js
+++ b/client/analytics/components/report-chart/index.js
@@ -135,29 +135,30 @@ export class ReportChart extends Component {
 	}
 
 	renderItemComparison() {
-		const { primaryData } = this.props;
+		const { isRequesting, primaryData } = this.props;
 
 		if ( primaryData.isError ) {
 			return <ReportError isError />;
 		}
 
-		const isRequesting = primaryData.isRequesting;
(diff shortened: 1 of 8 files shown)


timmyc (approved):
Code is looking good - and I could not see the no data placeholder in testing 👍

PR #1706, opened 2019-02-27
Display order number on reports

### LINKED ISSUE #1613: Reports: Order Post ID shown instead of Order Number
**Describe the bug**

Our reports display the Order Post ID, not the Order Number.

**To Reproduce**

Using WCCOM data, or a store where Order Number != Post ID.

1. View any report table that contains "order number"
2. Verify that the Post ID is shown, not the Order Number

**Expected behavior**

The Order Number is displayed instead of the WP Post ID that represents the Order.

**Screenshots**

screen shot 2019-02-18 at 3 47 48 pm

_(Left is Orders list, Right is our report)_

emdashcodes:
Fixes #1613.

This PR updates the orders and downloads report to show the order number instead of order ID. It also adds the number to the API response so we can actually show it on the JS side.

### Screenshots

screen shot 2019-02-27 at 2 03 27 pm

* Use a plugin with custom order numbers, WCCOM Staging, or a snip-it like [this](https://gist.github.com/justinshreve/6ea581827c770da955e5f4e95e342e89) to display an order number separate from the order ID.
* View the orders and downloads report, make sure the new order number displays.

### DIFF
--- a/includes/api/class-wc-admin-rest-reports-controller.php
+++ b/includes/api/class-wc-admin-rest-reports-controller.php
@@ -174,6 +174,20 @@ class WC_Admin_REST_Reports_Controller extends WC_REST_Reports_Controller {
 	}
 
 	/**
+	 * Get the order number for an order. If no filter is present for `woocommerce_order_number`, we can just return the ID.
+	 *
+	 * @param  int $order_id Order ID.
+	 * @return string
+	 */
+	public function get_order_number( $order_id ) {
+		if ( ! has_filter( 'woocommerce_order_number' ) ) {
+			return $order_id;
+		}
+		$order = new WC_Order( $order_id );
+		return $order->get_order_number();
+	}
+
+	/**
 	 * Prepare a report object for serialization.
 	 *
 	 * @param stdClass        $report  Report data.
--- a/includes/data-stores/class-wc-admin-reports-orders-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-orders-data-store.php
@@ -207,6 +207,11 @@ class WC_Admin_Reports_Orders_Data_Store extends WC_Admin_Reports_Data_Store imp
 				$this->include_extended_info( $orders_data, $query_args );
 			}
 
+			foreach ( $orders_data as $key => $order_data ) {
+				$wc_order                            = new WC_Order( $order_data['order_id'] );
+				$orders_data[ $key ]['order_number'] = $wc_order->get_order_number();
+			}
+
 			$orders_data = array_map( array( $this, 'cast_numbers' ), $orders_data );
 			$data        = (object) array(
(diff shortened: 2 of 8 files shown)


includes/data-stores/class-wc-admin-reports-orders-data-store.php
```diff
@@ -207,6 +207,11 @@ public function get_data( $query_args ) {
 				$this->include_extended_info( $orders_data, $query_args );
 			}
 
+			foreach ( $orders_data as $key => $order_data ) {
+				$wc_order                            = new WC_Order( $order_data['order_id'] );
+				$orders_data[ $key ]['order_number'] = $wc_order->get_order_number();
```
timmyc: Any performance concerns here / or is there a way to determine if a site is using something that manipulates order numbers?

timmyc (approved):
Thanks for the test snippet - this is working out well for me. Just the one question about loading the Order object in the data store from a performance standpoint. I attempted to dive into that a bit using query monitor, but i wasn't seeing the total number of queries exposed in the response headers as I anticipated.

Pre-approving.

includes/data-stores/class-wc-admin-reports-orders-data-store.php
justinshreve: Good question. There will definitely be extra queries from this, but I can't think of anyway else to get the order number. The `woocommerce_order_number` filter passes both the order ID and order object to plugins, so even if we reimplemented the filter, we would still need to load the object. Order data does get cached https://github.com/woocommerce/woocommerce/blob/de642211d08a81fb32addefa69cd1d0c1b6f7643/includes/abstracts/abstract-wc-order.php#L64.

However, this was a good suggestion for adding a check. We can use `has_filter` to avoid doing anything and just returning the order ID.

justinshreve:
@timmyc I responded to you here https://github.com/woocommerce/wc-admin/pull/1706#discussion_r261022567. I know you pre-approved this, but do you mind taking another quick 👀 since I changed a few things?

includes/api/class-wc-admin-rest-reports-controller.php
```diff
@@ -173,6 +173,20 @@ public function get_items( $request ) {
...
 	}
 
+	/**
+	 * Get the order number for an order. If no filter is present for `woocommerce_order_number`, we can just return the ID.
+	 *
+	 * @param  int $order_id Order ID.
+	 * @return string
+	 */
+	public function get_order_number( $order_id ) {
+		if ( ! has_filter( 'woocommerce_order_number' ) ) {
```
timmyc: Nice 👍

timmyc:
Nice really like the new method to prevent the extra query - tested with and without your example snippet and all is still working great.

Issue #1691, opened 2019-02-26
Settings: Move Lookup Tables UI to Settings Page


timmyc:
Currently the button that triggers the population of the report lookup tables lives in System | Tools page:

![image](https://user-images.githubusercontent.com/22080/53380058-dcf08e00-3920-11e9-8881-952e9b96b1b7.png)

@LevinMedia is in the process of coming up with some improved designs for this tool, but until those are completed, I think we should do the following items:

- [ ] Remove the button from the System | Tools page
- [ ] Build a new button on the Analytics | Settings page instead which triggers the action to populate the lookup tables
- [ ] When the button is clicked on the Analytics | Settings page - the button should be disabled to prevent double-submission ( and lots and lots of jobs being queued up.

justinshreve:
@LevinMedia @timmyc Can we include a design for clearing pending jobs as well as a part of this issue? 

I have a PR about to go up for the backend of https://github.com/woocommerce/wc-admin/issues/1699 -- but I didn't want to put too much work in the UI until we now what it looks like.

--

My two cents on this moving the buttons: It also seems a little odd to me to move these tools out of system tools. Re-running and/or clearing out queued options seem like uncommon actions (except for the initial setup which can be handled by a note https://github.com/woocommerce/wc-admin/issues/1709) and putting them front and center on a settings page seems like it could be dangerous. These buttons are also at home already with other similar actions (clearing sessions, etc) and make sense for future core merge.

LevinMedia:
@justinshreve 

>Can we include a design for clearing pending jobs as well as a part of this issue?

Affirmative. I want to include a Pause button and a clear button. 

> seems a little odd to me to move these tools out of system tools. Re-running and/or clearing out queued options seem like uncommon actions (except for the initial setup which can be handled by a note #1709) and putting them front and center on a settings page seems like it could be dangerous.

@josemarques and I had a conversation along these lines as well, it's a fair point. @timmyc mentioned it might be behoove us to use the new settings menu as it's powered by JS, rather than PHP, thus affording us more flexibility in design and usage of wc-admin components (such as a date range picker, if we wish to allow a store to process portions of their data at a time.) 

What if we were to show the settings conditionally, only when there is data that needs processing?

justinshreve:
> What if we were to show the settings conditionally, only when there is data that needs processing?

[...]

PR #1715, opened 2019-02-28
Clear scheduled actions on plugin deactivate

### LINKED ISSUE #1699: Create utility to clear all pending Analytics Background Jobs
In the event of a site experiencing resource issues due to a large number of analytics jobs processing, it seems we should have a way to delete all pending Action Scheduler jobs for analytics. 

I'm thinking a method which is very much the inverse of the initial install job that enqueues all the batch jobs - this util would delete all pending analytics jobs. It seems like we should have this method called when WooCommerce Admin is deactivated, but also possibly add a button to _Analytics | Settings_ where users could opt to remove all pending jobs without deactivating the plugin.

emdashcodes:
Fixes #1699.

This PR clears all scheduled `wc-admin` report generation actions when the plugin is deactivated.

screen shot 2019-02-28 at 12 34 26 pm

Note: There is no UI for this yet, pending designs at #1691.

* Queue up some jobs. For example, by running the regen report tool under system status.
* Deactivate the `wc-admin` plugin. You should see some canceled jobs start showing up under the canceled tab.

### DIFF
--- a/includes/class-wc-admin-install.php
+++ b/includes/class-wc-admin-install.php
@@ -14,7 +14,7 @@ class WC_Admin_Install {
 	/**
 	 * Plugin version.
 	 *
-	 * @TODO: get this dynamically?
+	 * @todo get this dynamically?
 	 */
 	const VERSION_NUMBER = '0.6.0';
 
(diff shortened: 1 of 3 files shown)


rrennick (approved):
@justinshreve LGTM 🚀 Great idea.

PR #1716, opened 2019-02-28
Add the number of pending report jobs to wcSettings

### LINKED ISSUE #1692: Setup: Expose Number of Pending Batch Jobs via wcSettings
The lookup table population works by creating "batch" jobs to populate first the [Customers](https://github.com/woocommerce/wc-admin/blob/aaa26118425e555c1e5e406739d76f3f3f0e15ba/includes/class-wc-admin-reports-sync.php#L379) lookup table, then the [Orders](https://github.com/woocommerce/wc-admin/blob/aaa26118425e555c1e5e406739d76f3f3f0e15ba/includes/class-wc-admin-reports-sync.php#L196) ( and supporting ) lookup tables.

While you can view pending jobs within the _Status_ section of WooCommerce, that view is a bit cryptic and doesn't really give a good sense of how much work is left - as batch jobs enqueue more jobs... and really the table of jobs looks a bit confusing.

As such, we are hoping to eventually create an interface for users to see how the lookup table population is progressing. For this issue I'm hoping we can look at finding a way to query the jobs table for the total batch jobs, and total completed for both the Customers and Order lookup population processes. Then we could expose these via wcSettings, or some other fashion, and eventually communicate progress to users.

**EDIT by Jeff:** The status should be based on the number of customers/orders to be processed and the number that have been added to the corresponding lookup tables.

emdashcodes:
Fixes #1692.

This PR adds the number of pending report jobs to `wcSettings`. This is not yet used/displayed anywhere but will be used for showing report generation status.

* Queue up some jobs (run the regen tool).
* Refresh a page and view source. Look for `pendingReportActions`.

### DIFF
--- a/includes/class-wc-admin-reports-sync.php
+++ b/includes/class-wc-admin-reports-sync.php
@@ -125,6 +125,20 @@ class WC_Admin_Reports_Sync {
 		);
 	}
 
+	/**
+	 * Get the number of pending actions for reports.
+	 *
+	 * @return int
+	 */
+	public static function get_pending_actions() {
+		$query   = array(
+			'status' => ActionScheduler_Store::STATUS_PENDING,
+			'group'  => self::QUEUE_GROUP,
+		);
+		$store   = ActionScheduler::store();
+		$pending = $store->query_actions( $query, 'count' );
+		return $pending;
+	}
 
 	/**
 	 * Schedule an action to process a single Order.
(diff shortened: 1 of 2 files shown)


includes/class-wc-admin-reports-sync.php
```diff
@@ -125,6 +125,20 @@ public static function add_regenerate_tool( $tools ) {
 		);
 	}
 
+	/**
+	 * Get the number of pending actions for reports.
+	 *
+	 * @return int
+	 */
+	public static function get_pending_actions() {
```
justinshreve: I'm heading out for the rest of the week, but leaving myself a piece of feedback for later😆.  I'm thinking `get_pending_actions_count` is probably a better name for this.

PR #1717, opened 2019-03-01
Date picker: fix tabs CSS

psealock:
Fix tabpanels in datepicker CSS broken by Gutenberg update. The white outline seen in the screen shot below is a tab's focus state.

### Before

screen shot 2019-03-01 at 4 30 30 pm

### After

screen shot 2019-03-01 at 4 30 14 pm

1. Any report with date range picker (such as Revenue).
2. Open the date range picker and make sure the tabs render and work correctly. Note the focus state when using the keyboard.

### DIFF
--- a/packages/components/src/filters/date/style.scss
+++ b/packages/components/src/filters/date/style.scss
@@ -38,7 +38,7 @@
 	@include set-grid-item-position( 2, 2 );
 }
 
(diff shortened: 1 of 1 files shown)


timmyc (approved):
Good catch - I had to update to the latest g-berg to view the bug in action first, and can confirm this fixes it up

PR #1718, opened 2019-03-01
Update tracking to use wcTracks method

### LINKED ISSUE #1697: Tracks: Use Woo Core's supplied method
Currently our [track utility](https://github.com/woocommerce/wc-admin/blob/master/client/lib/tracks.js#L17) uses `wca_` as a prefix while the work that is being done for Woo Core is using `wcadmin_` https://github.com/woocommerce/woocommerce/pull/22851

For this simple task we should update our usage here to match that in core.

joshuatf:
Fixes #1697 

Uses the core method for tracking and relies on it to add scripts.

1.  Set your node env to anything but `development`.
2.  Checkout the `feature/add-tracks` of WooCommerce.
3. Check that events are recorded on page change.
4. Note that the initial page load will not be tracked until https://github.com/woocommerce/woocommerce/pull/22914 is merged in.
5.  Check that no console errors are thrown when tracking is enabled or disabled.

### DIFF
--- a/lib/client-assets.php
+++ b/lib/client-assets.php
@@ -148,19 +148,6 @@ function format_order_statuses( $statuses ) {
 function wc_admin_print_script_settings() {
 	global $wp_locale;
 
-	// Add Tracks script to the DOM if tracking is opted in, and Jetpack is installed/activated.
-	$tracking_enabled = 'yes' === get_option( 'woocommerce_allow_tracking', 'no' );
-	$tracking_script  = '';
-	if ( $tracking_enabled && defined( 'JETPACK__VERSION' ) ) {
-		$tracking_script  = "var wc_tracking_script = document.createElement( 'script' );\n";
-		$tracking_script .= "wc_tracking_script.src = '//stats.wp.com/w.js';\n"; // @todo Version/cache buster.
-		$tracking_script .= "wc_tracking_script.type = 'text/javascript';\n";
-		$tracking_script .= "wc_tracking_script.async = true;\n";
-		$tracking_script .= "wc_tracking_script.defer = true;\n";
-		$tracking_script .= "window._tkq = window._tkq || [];\n";
-		$tracking_script .= "document.head.appendChild( wc_tracking_script );\n";
-	}
-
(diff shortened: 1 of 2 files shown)


lib/client-assets.php
```diff
@@ -189,18 +176,18 @@ function wc_admin_print_script_settings() {
...
+		'adminUrl'         => admin_url(),
+		'wcAssetUrl'       => plugins_url( 'assets/', WC_PLUGIN_FILE ),
+		'wcAdminAssetUrl'  => plugins_url( 'images/', wc_admin_dir_path( 'wc-admin.php' ) ), // Temporary for plugin. See above.
+		'embedBreadcrumbs' => wc_admin_get_embed_breadcrumbs(),
+		'siteLocale'       => esc_attr( get_bloginfo( 'language' ) ),
+		'currency'         => wc_admin_currency_settings(),
+		'orderStatuses'    => format_order_statuses( wc_get_order_statuses() ),
+		'stockStatuses'    => wc_get_product_stock_status_options(),
+		'siteTitle'        => get_bloginfo( 'name' ),
+		'trackingEnabled'  => 'yes' === get_option( 'woocommerce_allow_tracking', 'no' ),
```
psealock: I think we should remove this in favour of letting Core handle `trackingEnabled` logic. One reason is the `woocommerce_apply_user_tracking` filter allows plugins to turn off tracking despite what the user's `woocommerce_allow_tracking` option value is.

https://github.com/woocommerce/woocommerce/blob/37d3fff3d6c9eeda66e32f1c29bcf3e6fe543a46/includes/tracks/class-wc-site-tracking.php#L25-L27

psealock (approved):
LGTM Josh. Just a comment about `trackingEnabled` and whose responsibility it is to determine that value.

lib/client-assets.php
joshuatf: @psealock I went back and forth on this, but in light of the `woocommerce_apply_user_tracking` filter, you're definitely right.  Removed- thanks for noting this!

Issue #1668, opened 2019-02-22
Zero value items aren't shown when comparing items


joshuatf:
**Describe the bug**
When selecting 2 or more items to compare, an item that is zero value (e.g., a product without sales during a given period) is not shown on the table.

Not urgent, but might be nice to show all the items being compared, even if they are zero.  Thoughts?

**To Reproduce**

1. Go to the products report.
2. Compare a product with sales during a given period and another without sales during that same period using the advanced filter.
3.  Note the table does not include the second product being compared.

**Expected behavior**
Expected all items being compared to show up in the table.

**Screenshots**
<img width="771" alt="screen shot 2019-02-22 at 9 49 47 am" src="https://user-images.githubusercontent.com/10561050/53214709-98e35c00-3688-11e9-90b4-5dd69d0ad6f8.png">
<img width="487" alt="screen shot 2019-02-22 at 9 49 37 am" src="https://user-images.githubusercontent.com/10561050/53214710-98e35c00-3688-11e9-9166-396591080d1e.png">

Aljullu:
I think that's related to #1398.

Aljullu:
Closed by #1765.

Issue #1704, opened 2019-02-27
Legend and tabular data not displaying when comparing series with no data present


LevinMedia:
**Describe the bug**
when comparing series and no data is present, the series in the legend disappear. All of the rows in the tabular data disappear too.  

**To Reproduce**
Create a comparison, select a date range with no data.
Note, the series in the legend disappear
Note the tabular data rows disappear

**Expected behavior**
Even if no data is available, the compared series should remain in the legend with zero values - as well as in the tabular data below (not pictured in screen shot) also with zero values. 

<img width="1380" alt="52029986-91d28f00-24ca-11e9-9a81-538cd0029ecc" src="https://user-images.githubusercontent.com/4500952/53508375-78dbe000-3a6e-11e9-8d17-975abbf0cebe.png">

**Previously noted in :** 
https://github.com/woocommerce/wc-admin/issues/582#issuecomment-459209125

Aljullu:
I think this is the same as #1668. I'm currently working on that one, so I will make sure to verify this one is fixed as well when submitting a PR.

LevinMedia:
Yeah for sure, the tabular data portion is the same as what I'm talking about here - I just want to make sure the absence of the legend is also address, whether that's in #1668 or another pr. 

Thank you!

Aljullu:
Closed by #1765.

PR #1722, opened 2019-03-01
Make products and rates endpoints return zero-value items

Aljullu:
Part of #1668.

Makes `products` and `rates` endpoints return zero-value items.

I wanted to create this PR before working on the other endpoints to get some early feedback.

### Screenshots
_(Notice _Beanie_ appears in the table even though its values are 0)_
image

- Start a _Product Comparison_ making sure one of the products has all values to 0 (it has not been sold during the selected date range).
- Verify the table shows all compared products, even the one that has all values to 0.
- Repeat the same with different filters, with an active table search, etc.
- Now, repeat the same in the _Taxes_ report.

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-data-store.php
@@ -636,14 +636,18 @@ class WC_Admin_Reports_Data_Store {
 	/**
 	 * Returns a comma separated list of the field names prepared to be used for a selection after a join with `default_results`.
 	 *
-	 * @param array  $fields   Array of fields name.
-	 * @param string $id_field Name of the column used as an identifier.
+	 * @param array  $fields           Array of fields name.
+	 * @param string $id_field         Name of the column used as an identifier.
+	 * @param array  $outer_selections Array of fields that are not selected in the inner query.
 	 * @return string
 	 */
-	protected function format_join_selections( $fields, $id_field ) {
+	protected function format_join_selections( $fields, $id_field, $outer_selections = array() ) {
 		foreach ( $fields as $i => $field ) {
 			if ( $field === $id_field ) {
 				$fields[ $i ] = "default_results.id AS {$id_field}";
+			}
+			if ( in_array( $field, $outer_selections, true ) ) {
+				$fields[ $i ] = $this->report_columns[ $field ];
 			}
 		}
 		return implode( ', ', $fields );
--- a/includes/data-stores/class-wc-admin-reports-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-data-store.php
@@ -607,6 +607,49 @@ class WC_Admin_Reports_Data_Store {
 	}
 
 	/**
+	 * Generates a virtual table given a list of IDs.
+	 *
+	 * @param array $ids Array of IDs.
+	 * @return array
+	 */
+	protected function get_ids_table( $ids ) {
+		$selects = array();
+		foreach ( $ids as $id ) {
+			array_push( $selects, "SELECT {$id} AS id" );
+		}
+		return join( ' UNION ', $selects );
+	}
+
+	/**
+	 * Returns a comma separated list of the fields in the `query_args`, if there aren't, returns `report_columns` keys.
+	 *
+	 * @param array $query_args Parameters supplied by the user.
+	 * @return array
+	 */
+	protected function get_fields( $query_args ) {
+		if ( isset( $query_args['fields'] ) && is_array( $query_args['fields'] ) ) {
+			return $query_args['fields'];
+		}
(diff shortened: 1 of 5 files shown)


timmyc:
@joshuatf could you give this one some 👀 if you have a chance today?

includes/data-stores/class-wc-admin-reports-data-store.php
```diff
@@ -606,6 +606,53 @@ protected function get_limit_sql_params( $query_args ) {
 		return $sql_query;
 	}
 
+	/**
+	 * Generates a virtual table given a list of IDs.
```
joshuatf: Awesome 🔥

includes/data-stores/class-wc-admin-reports-taxes-data-store.php
```diff
@@ -178,41 +215,70 @@ public function get_data( $query_args ) {
...
 							GROUP BY
 								{$table_name}.tax_rate_id
-					  		) AS tt"
-			); // WPCS: cache ok, DB call ok, unprepared SQL ok.
-
-			$total_pages = (int) ceil( $db_records_count / $sql_query_params['per_page'] );
-			if ( $query_args['page'] < 1 || $query_args['page'] > $total_pages ) {
-				return $data;
+						) AS {$table_name}
+						RIGHT JOIN ( {$ids_table} ) AS default_results
```
joshuatf: Same comment here about conditionally adding in this JOIN to the default query.

tests/api/reports-taxes.php
```diff
@@ -112,6 +112,106 @@ public function test_get_reports() {
...
+		$product->set_regular_price( 25 );
+		$product->save();
+
+		$wpdb->insert(
+			$wpdb->prefix . 'woocommerce_tax_rates',
+			array(
+				'tax_rate_id'       => 1,
+				'tax_rate'          => '7',
+				'tax_rate_country'  => 'US',
+				'tax_rate_state'    => 'GA',
```
joshuatf: Georgia taxes 🤑

joshuatf (commented):
Awesome work on this @Aljullu!  Definitely a tricky one.

Left a comment about possibly combining the 2 queries and conditionally adding the join, but everything else is testing well for me.

One other item (that we should probably address separately) is the charts not showing zero data.  Is there an issue open or should we make a note of this on the existing issue?

screen shot 2019-03-05 at 3 38 38 pm

includes/data-stores/class-wc-admin-reports-products-data-store.php
```diff
@@ -257,48 +273,91 @@ public function get_data( $query_args ) {
...
+									{$table_name}
+									{$sql_query_params['from_clause']}
+								WHERE
+									1=1
+									{$sql_query_params['where_time_clause']}
+									{$sql_query_params['where_clause']}
+								GROUP BY
+									product_id
+						) AS {$table_name}
+						RIGHT JOIN ( {$ids_table} ) AS default_results
```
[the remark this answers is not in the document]
Aljullu: Done in 719e5908f218c1c4ac01ab0eae81718a541f8103.

Aljullu:
@joshuatf thanks for the review! Feel free to take another look when you have time.

> One other item (that we should probably address separately) is the charts not showing zero data. Is there an issue open or should we make a note of this on the existing issue?

Right, initially I thought it was the same issue, but in fact it has nothing to do, that's a JS issue not related with the endpoints. I'm tracking it as part of #1704 and #1730 was a step towards that fix.

includes/data-stores/class-wc-admin-reports-products-data-store.php
```diff
@@ -128,6 +116,32 @@ protected function get_order_by_sql_params( $query_args ) {
...
 
+	/**
+	 * Fills outer FROM clause of SQL request based on user supplied parameters.
+	 *
+	 * @param array  $query_args Parameters supplied by the user.
+	 * @param string $arg_name   Name of the FROM sql param.
+	 * @param string $id_cell    ID cell identifier, like `table_name.id_column_name`.
+	 * @return array
+	 */
+	protected function get_outer_from_sql_params( $query_args, $arg_name, $id_cell ) {
```
joshuatf: Optional:  Should we change this name to `get_from_sql_params` since it looks like it can fill both from and outer from clauses now?

joshuatf (approved):
Awesome work, @Aljullu.  One comment about renaming `get_outer_from_sql_params()` but otherwise LGTM!

includes/data-stores/class-wc-admin-reports-products-data-store.php
Aljullu: Done!

fa2694deff7e2fb24a00e50d6c0f6b1293763d21

PR #1723, opened 2019-03-01
Hook up StoreAlerts component to notes endpoint

### LINKED ISSUE #1708: Store Alerts: Hook up to notes endpoint
Now that we have the new [Store Alerts](https://github.com/woocommerce/wc-admin/tree/master/client/layout/store-alerts) component, we should hook it up to the rest endpoint to have it populated with real data.

I verified with @allendav that the notes logic is ready to fully support the Store Alerts section, so we should be able to utilize all existing data layer logic to make this happen. The alerts section should display notes with the [types](https://github.com/woocommerce/wc-admin/blob/master/includes/class-wc-admin-note.php#L18-L21) of `error,update`.  Some prior art on consuming notes can be [seen in the inbox code here](https://github.com/woocommerce/wc-admin/blob/b8b6d1801e9472fd820d67709a26e075b2610d55/client/header/activity-panel/panels/inbox.js#L92-L99).

tiagonoronha:
Fixes #1708

alerts

Refactors and hook up the StoreAlerts component to the notes endpoint.

- Create a few notes with the type `update` and error`. Example below.
- Reload a WC Admin page, you should see the alerts at the top

```
$name = 'wc-admin-test-alert-1';
$data_store = WC_Data_Store::load( 'admin-note' );
$note = new WC_Admin_Note();
$note->set_title( __( 'Sed bibendum non augue tincidunt mollis', 'wc-admin' ) );
$note->set_content( __( 'Quisque in efficitur nisi. In hac habitasse platea dictumst. Vivamus ut congue diam.', 'wc-admin' ) );
$note->set_type( WC_Admin_Note::E_WC_ADMIN_NOTE_UPDATE );
$note->set_icon( 'flag' );
$note->set_name( $name );
$note->set_source( 'wc-admin' );
$note->add_action(
  'action-1',
  __( 'Do something', 'wc-admin' ),
  '#'
);
$note->save();
```

### DIFF
--- a/client/layout/store-alerts/index.js
+++ b/client/layout/store-alerts/index.js
@@ -7,60 +7,21 @@ import { Component, Fragment } from '@wordpress/element';
 import { IconButton, Button, Dashicon } from '@wordpress/components';
 import classnames from 'classnames';
 import interpolateComponents from 'interpolate-components';
-import PropTypes from 'prop-types';
+import { compose } from '@wordpress/compose';
 
 /**
  * WooCommerce dependencies
  */
 import { Card } from '@woocommerce/components';
 
+/**
+ * Internal dependencies
+ */
+import withSelect from 'wc-api/with-select';
+import { QUERY_DEFAULTS } from 'wc-api/constants';
+import sanitizeHTML from 'lib/sanitize-html';
+
 import './style.scss';
-
-const dummy = [
-	{
-		title: 'Lorem ipsum dolor sit amet',
-		type: 'alert',
-		message:
-			'Pellentesque accumsan ligula in aliquam tristique. Donec elementum magna ut sapien tincidunt aliquam.',
-		action: {
-			label: 'Button',
-			url: '#',
-		},
-	},
-	{
-		title: 'Sed bibendum non augue tincidunt mollis',
-		type: 'critical',
-		message: 'Quisque in efficitur nisi. In hac habitasse platea dictumst. Vivamus ut congue diam.',
-		action: {
-			label: 'Button',
-			url: '#',
-		},
-	},
-	{
-		title: 'Duis dictum condimentum sem eu blandit',
-		type: 'emergency',
-		message: 'Fusce fermentum magna dolor, vitae faucibus justo ullamcorper eu.',
-		action: {
-			label: 'Button',
-			url: '#',
-		},
-	},
-];
-
-const alertTypes = {
-	emergency: {
-		label: __( 'Emergency', 'wc-admin' ),
-		icon: 'warning',
-	},
-	alert: {
-		label: __( 'Alert', 'wc-admin' ),
-		icon: 'flag',
-	},
-	critical: {
-		label: __( 'Critical', 'wc-admin' ),
-		icon: 'info',
-	},
-};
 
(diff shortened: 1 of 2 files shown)


client/layout/store-alerts/index.js
```diff
@@ -104,18 +65,23 @@ class StoreAlerts extends Component {
...
-			'is-alert-emergency': 'emergency' === type,
-			'is-alert-alert': 'alert' === type,
-			'is-alert-critical': 'critical' === type,
+			'is-alert-error': 'error' === type,
+			'is-alert-update': 'update' === type,
 		} );
+		const actions =
+			alert &&
+			alert.actions.map( action => (
+				<Button key={ action.name } isDefault href={ action.url }>
```
timmyc: Is there any chance an action might not have a name? I don't remember off the top fo my head if it is required or not.

client/layout/store-alerts/index.js
```diff
@@ -104,18 +65,23 @@ class StoreAlerts extends Component {
...
 		const type = alert && alert.type ? alert.type : null;
-		const icon = type && alertTypes[ type ] ? alertTypes[ type ].icon : null;
 		const className = classnames( 'woocommerce-store-alerts', {
-			'is-alert-emergency': 'emergency' === type,
-			'is-alert-alert': 'alert' === type,
-			'is-alert-critical': 'critical' === type,
+			'is-alert-error': 'error' === type,
+			'is-alert-update': 'update' === type,
 		} );
+		const actions =
```
timmyc: Should we do a `return` early, perhaps after `const { alerts } = this.props;` if `! alerts || alerts.length === 0`?

client/layout/store-alerts/index.js
```diff
@@ -151,29 +117,28 @@ class StoreAlerts extends Component {
 						)
 					}
 				>
-					<div className="woocommerce-store-alerts__message">{ alert.message }</div>
-					<Button isDefault className="woocommerce-store-alerts__button" href={ alert.action.url }>
-						{ alert.action.label }
-					</Button>
+					<div
+						className="woocommerce-store-alerts__message"
+						dangerouslySetInnerHTML={ sanitizeHTML( alert.content ) }
```
timmyc: Nice work finding `sanitizeHTML` 👍

client/layout/store-alerts/index.js
timmyc: I tried to add an action without a name and it wouldn't let me, so disregard :)

tiagonoronha:
@timmyc 

>Should we do a return early, perhaps after const { alerts } = this.props; if ! alerts || alerts.length === 0?

Done done!

> Couple of ideas, which we can do in a follow-up PR:
> 
> * Hydrate the state of the alerts on page load via PHP, so if there are alerts, they will be shown immediately.
> * Create a loading state for this alert area... not as excited about that route

I think that on this specific case it makes sense to hydrate the state via PHP; I don't think we want a loading state that then turns into nothing if no alerts exist.

Happy to do it in a follow-up PR.

timmyc:
Let's do handle the initial state hydration in a follow-up PR.

PR #1727, opened 2019-03-04
Add actionable order statuses settings

### LINKED ISSUE #1593: Create actionable order statuses setting
The orders in the activity panel currently show `processing` orders (and `on-hold` pending https://github.com/woocommerce/wc-admin/pull/1578).  These are orders that require further action from the user (verifying payment or fulfilling an order).

Since actionable is subject to the user's setup and they may have custom statuses, we should also create a setting to allow users to pick their own actionable statuses with `processing` and `on-hold` as the defaults.

* Add the actionable statuses setting to the settings page.
* Hook up the actionable statuses to statuses used in the orders activity panel- https://github.com/woocommerce/wc-admin/blob/54426b096c8dffa13f45bdec29423288b3565e39/client/header/activity-panel/panels/orders.js#L182

joshuatf:
Fixes #1593 

Adds a setting for order statuses that required admin action and uses these statuses to pull orders in the orders activity panel.

### Screenshots
screen shot 2019-03-04 at 11 38 16 am
screen shot 2019-03-04 at 11 38 01 am

1. Go to the analytics settings ``.
2. Add/remove actionable order statuses.
3. Refresh the page and check that the new order statuses show up for recent orders in the activiy panel.
4. Uncheck all order statuses and make sure no orders are shown and no additional requests are made.

### DIFF
--- a/client/header/activity-panel/panels/orders.js
+++ b/client/header/activity-panel/panels/orders.js
@@ -205,7 +205,10 @@ export default compose(
 		const ordersQuery = {
 			page: 1,
 			per_page: QUERY_DEFAULTS.pageSize,
-			status_is: [ 'processing', 'on-hold' ],
+			status_is: wcSettings.wcAdminSettings.woocommerce_actionable_order_statuses || [
+				'processing',
+				'on-hold',
+			],
 			extended_info: true,
 		};
 
--- a/client/header/activity-panel/panels/orders.js
+++ b/client/header/activity-panel/panels/orders.js
@@ -202,15 +202,20 @@ OrdersPanel.defaultProps = {
 export default compose(
 	withSelect( select => {
 		const { getReportItems, getReportItemsError, isReportItemsRequesting } = select( 'wc-api' );
+		const orderStatuses = wcSettings.wcAdminSettings.woocommerce_actionable_order_statuses || [
+			'processing',
+			'on-hold',
+		];
 		const ordersQuery = {
 			page: 1,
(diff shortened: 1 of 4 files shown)


client/header/activity-panel/panels/orders.js
```diff
@@ -202,13 +202,21 @@ OrdersPanel.defaultProps = {
...
+		];
 		const ordersQuery = {
 			page: 1,
 			per_page: QUERY_DEFAULTS.pageSize,
-			status_is: [ 'processing', 'on-hold' ],
+			status_is: orderStatuses,
 			extended_info: true,
 		};
 
+		if ( ! orderStatuses.length ) {
```
timmyc: If this setting has no length, wondering if we should show a message in the panel along the lines of "Hey you currently have no order statuses set to display orders here!" Likely could be worded better, but otherwise the panel will always seem broken.

I do like the notion of bailing early before making the API request though 👍

timmyc (approved):
Really neat to see the code required to make this addition to the settings page be so small, props to you for constructing it in such a way to make adding on like this easy to do.

One note/thought about how to handle the case when no statuses are set, I'm fine with that as a follow-up PR too, or just logging an issue and coming back to it later.

client/header/activity-panel/panels/orders.js
joshuatf: Good idea @timmyc.  Noted that issue in https://github.com/woocommerce/wc-admin/issues/1749 so we can follow-up on it.

PR #1728, opened 2019-03-04
Define extended_info in case query is filtered

### LINKED ISSUE #1595: Filtering extended_info breaks reports
**Describe the bug**
As noted [here](https://github.com/woocommerce/wc-admin/pull/1578#discussion_r257342020) since the API response can be filtered by third parties, if `extended_info` was removed, a few of our existing reports would break.

**To Reproduce**

1. Filter the API response for products using `woocommerce_reports_products_select_query` OR in `client/analytics/report/products/table.js` add in `delete row.extended_info;` before `extended_info` is destructured.
2. Visit the products report.
3.  Note the errors.
4.  Note the same behavior in various other reports (coupons, products, leaderboards, orders, maybe others?)

**Expected behavior**
Expected the page to continue loading regardless of missing data.

**Screenshots**
screen shot 2019-02-18 at 12 14 17 pm

joshuatf:
Fixes #1595 

Creates an empty object for `extended_info` in the event that a filter is added and removes extended_info from the report.

1.  Filter various report data.
```php
add_filter( 'woocommerce_reports_coupons_select_query', function ( $results ) {
	foreach ( $results->data as &$data ) {
		unset( $data['extended_info'] );
	}
	return $results;
} );
```
2.  Check that the reports continue to work as expected (less the altered data).

### DIFF
--- a/client/analytics/report/coupons/table.js
+++ b/client/analytics/report/coupons/table.js
@@ -72,7 +72,8 @@ export default class CouponsReportTable extends Component {
 		const persistedQuery = getPersistedQuery( query );
 
 		return map( coupons, coupon => {
-			const { amount, coupon_id, extended_info, orders_count } = coupon;
+			const { amount, coupon_id, orders_count } = coupon;
+			const extended_info = coupon.extended_info || {};
 			const { code, date_created, date_expires, discount_type } = extended_info;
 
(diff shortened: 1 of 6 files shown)


timmyc (approved):
Looks good to me, thanks for the fix Josh.

PR #1729, opened 2019-03-04
Update table columns visibility on finish loading

### LINKED ISSUE #1720: Products Report: columns visibility does not persist
**To Reproduce**
2. Make _SKU_ column visible.
3. Reload the page.
4. Notice _SKU_ column is hidden again.

**Expected behavior**
_SKU_ column should be visible after reload.

The same happens making visible-by-default columns hidden.

Aljullu:
Fixes #1720.

Updates table columns visibility when `onLoading` prop goes from `false` to `true`. In some cases, we were rendering the table placeholder before `userPrefColumns` was available, so the setting was ignored.

### Screenshots
columns-visibility

* Go to any report.
* Toggle some columns' visibility.
* Reload the page.
* Verify columns visibility persists.

### DIFF
--- a/packages/components/src/table/index.js
+++ b/packages/components/src/table/index.js
@@ -58,8 +58,8 @@ class TableCard extends Component {
 		this.selectAllRows = this.selectAllRows.bind( this );
 	}
 
-	componentDidUpdate( { query: prevQuery } ) {
-		const { compareBy, onColumnsChange, query } = this.props;
+	componentDidUpdate( { headers: prevHeaders, isLoading: prevIsLoading, query: prevQuery } ) {
+		const { compareBy, headers, isLoading, onColumnsChange, query } = this.props;
 		const { showCols } = this.state;
 
 		if ( query.filter || prevQuery.filter ) {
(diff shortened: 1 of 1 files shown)


packages/components/src/table/index.js
```diff
@@ -73,6 +73,13 @@ class TableCard extends Component {
 				/* eslint-enable react/no-did-update-set-state */
 			}
 		}
+		if ( ! isLoading && prevIsLoading && ! isEqual( headers, prevHeaders ) ) {
+			/* eslint-disable react/no-did-update-set-state */
+			this.setState( {
+				showCols: headers.map( ( { key, hiddenByDefault } ) => ! hiddenByDefault && key ).filter( Boolean ),
```
coderkevin: Why not use an update function for `setState`? It would guarantee `headers` is up-to-date when the state is set: https://reactjs.org/docs/react-component.html#setstate

Aljullu: Will there be any difference using a function or an object in `setState` considering `headers` is a prop? I think `props` will be up-to-date in both cases, so there is no need to use a function for `setState`. Am I missing something?

coderkevin: Yes, that's true. I must have assumed headers was in state. I don't think `setState` is susceptible to props changing in between, and if it were, another update would soon follow.

coderkevin (approved):
Tests good!

PR #1730, opened 2019-03-04
Use IDs instead of labels to identify chart legend items

### LINKED ISSUE #1659: Line Charts: month abbreviations are used in aria-labels instead of full month names
**To Reproduce**
1. Go to any report.
2. Navigate the line chart with a screen reader.
3. Notice the line regions are named `Previous Period (Feb 20, 2019)` but they should be `Previous Period (February 20, 2019)`.

This affects all intervals.
image

### LINKED ISSUE #1719: Duplicate element IDs in the dashboard
**To Reproduce**

1. Go to the _Dashboard_.
2. With your browser devtools, hover chart legend items and look at their IDs.
3. Hover legend items of the next chart.
4. Notice they have the same ID.

**Expected behavior**
There shouldn't be repeated IDs in the page.

**Screenshots**
image
image

Aljullu:
Fixes #1719. And it will make it easier to fix #1659 and #1704.

This PR:
- Adds `withInstanceId` HOC to chart legend items, so the ids used in several HTML elements are unique.
- Replaces the `label` with the `id` as the identifier for legend items (that will be required for #1659 and #1704).

### Accessibility

### Screenshots
_Notice the ID (`woocommerce-legend-18__item__509`) uses the id `509` instead the label `Belt` and has an instance id `18`:_
image

**Verify there are no regressions:**
- Go to any report (_Products_, for example).
- Interact with the chart legend hovering the elements and activating/deactivating them using the mouse and the keyboard.
- Activate _Top Items by Items Sold_ filter so the `item-comparison` chart is displayed.
- Interact with the chart legend again and verify there are no regressions.

**Verify #1719 is fixed:**
- Go to the _Dashboard_ and, with your browser devtools, verify chart legend items have all unique ids.


packages/components/src/chart/d3chart/legend.js
```diff
@@ -5,6 +5,7 @@
 import { __, sprintf } from '@wordpress/i18n';
 import classNames from 'classnames';
 import { Component, createRef } from '@wordpress/element';
+import { withInstanceId } from '@wordpress/compose';
```
timmyc: TIL about `withInstanceId`

packages/components/src/chart/d3chart/legend.js
```diff
@@ -173,6 +174,8 @@ D3Legend.propTypes = {
 	 * comparison charts when there are many.
 	 */
 	totalLabel: PropTypes.string,
+	// from withInstanceId
```
timmyc: Thanks for the comment on the new proptype

packages/components/src/chart/d3chart/utils/bar-chart.js
```diff
@@ -63,14 +63,10 @@ export const drawBars = ( node, data, params, scales, formats, tooltip ) => {
 		.attr( 'pointer-events', 'none' )
 		.attr( 'tabindex', '0' )
 		.attr( 'aria-label', d => {
-			let label = d.key;
+			let label = d.label || d.key;
 			if ( params.mode === 'time-comparison' ) {
-				if ( d.label ) {
```
timmyc: Love the simplification here.

timmyc (commented):
I'm seeing the following for date-based legends:

ids

Which I thought ids were not supposed to have spaces in them. I realize on `master` it is pretty much the same issue, but was wondering if we need to not have spaces there.

Aljullu:
> Which I thought ids were not supposed to have spaces in them. I realize on master it is pretty much the same issue, but was wondering if we need to not have spaces there.

Good point. I [replaced](a52889ff77f7021698261ea83adc5a111cf262f8) them with `primary` and `secondary`. Given that ids are only used internally, I think it makes sense to use those generic words instead of the labels. I also rebased this branch.

@timmyc, would you mind taking another look?

coderkevin (approved):
Testing checks out and looks good!

Issue #375, opened 2018-09-07
Define specific color stops for charts with five series or less


LevinMedia:
Charts currently pull from the extreme ends of our color scale. For example, on a two series legend we have a dark purple and golden yellow. 

<img width="1448" alt="screen shot 2018-09-07 at 1 30 54 pm" src="https://user-images.githubusercontent.com/4500952/45241842-d69f2480-b2a2-11e8-89fe-be141a5cca36.png">

Using the following color stops will result in a less extreme color scale as the number of series progresses up to 5. After 5 series we can let math take over, creating an even scale with however many stops are required. 

The images below are based  the color scales [Viridis](https://github.com/d3/d3-scale-chromatic#interpolateViridis) and [Plasma](https://github.com/d3/d3-scale-chromatic#interpolatePlasma)

<img width="1995" alt="screen shot 2018-09-07 at 12 14 20 pm" src="https://user-images.githubusercontent.com/4500952/45240455-9c7f5400-b29d-11e8-9eaa-dc57cea86621.png">

greenafrican:
Fixed in #381

PR #1731, opened 2019-03-04
Update chart colors to match spec

### LINKED ISSUE #1705: Incorrect color scale when displaying 5 series
**Describe the bug**
When displaying five series, the color scale should match what is outlined at #375

**To Reproduce**
compare at least five products - note the color scale does not match the link above. 

**Additional context**
Sorry if there was confusion around this - the image in #375 says we should let the math take over at five or more series, when it should actually take over if we're comparing six or more series. 

For now we're limiting comparisons to five things. If in the future we allow comparisons 6 or more things, the first and the last series should always be `t=0` and `t=1` with any remaining series being distributed evenly in between. 

cc @Aljullu

Aljullu:
Fixes #1705.

This PR:
- Updates the constants used to generate the chart colors to match what is defined in #375.
- Updates the chart legend to call `getColors` only on visible items.

### Screenshots
image

- Go to the _Products_ report and active the _Top Products by Items Sold_ filter, so there are at least 5 items displayed in the chart.
- Verify colors match #375.
- Deactivate other items one by one and verify colors still match #375.


LevinMedia:
Looks good, thank you!

timmyc (approved):
Code is looking 🎨 to me

PR #1732, opened 2019-03-04
Skip dispatching fresh data requests when the window/tab is hidden.

### LINKED ISSUE #1565: Data Layer: Disable Requests on inactive tabs
This concept has been brought up in some conversations before, and @psealock mentioned it [here](https://github.com/woocommerce/wc-admin/pull/1549#pullrequestreview-203548718) - so figured we should log an issue around it:

__Feature Request__
When a browser tab is inactive, it would be ideal for the data layer to not request data from the REST API. Currently when this happens, it often will result in a failed request and an the large purple triangle error being shown.

Paul mentioned looking into https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API potentially to see if we could build upon that.

jeffstieler:
Fixes #1565.

Prevents `FRESH_DATA_REQUESTED` from being dispatched when the window/tab isn't visible.

### Accessibility

N/A

- Temporarily decrease the default freshness to something like 15 seconds (in `client/wc-api/constants.js`)
- Visit the WC Admin dashboard
- Open Redux DevTools, filter actions by "req"
- Verify that `FRESH_DATA_REQUESTED` occurs again after your new freshness interval
- Switch tabs
- Wait a minute or so, switch back
- Verify no `FRESH_DATA_REQUESTED` were dispatched while tab was hidden

### DIFF
--- a/client/wc-api/wp-data-store/create-api-client.js
+++ b/client/wc-api/wp-data-store/create-api-client.js
@@ -19,6 +19,9 @@ function createStore( name ) {
 function createDataHandlers( store ) {
 	return {
 		dataRequested: resourceNames => {
+			if ( document.hidden ) {
+				return;
+			}
 			store.dispatch( {
 				type: 'FRESH_DATA_REQUESTED',
 				resourceNames,


coderkevin (approved):
Looks like a reasonable use. I'm hoping to add some functionality like this into fresh-data itself soon.

PR #1733, opened 2019-03-05
Remove un-used file

timmyc:
While cruising around the repo today, I spied this old file, all empty, and alone. I let it know everything would be okay, and `git rm`ed it.

__To Test__
Boot up the branch and make sure nothing goes 💥


Aljullu (approved):
Good catch! You're right it looks like this file is not used anywhere else.

RIP `wc-admin-order-functions.php` :wave:

PR #1737, opened 2019-03-05
Add rebuild tool to settings page

### LINKED ISSUE #1641: Rebuild reports pending notification
After rebuilding reports, a success message is shown that states: "Tool ran."  I know we have [an open issue](https://github.com/woocommerce/wc-admin/issues/1358) to address this and other items, but I think to avoid confusion, it might good idea prior to release to change the success text to read something like:

`Report regeneration has been started.  This may take a while depending on the number of orders.`

screen shot 2019-02-20 at 4 36 00 pm

Thanks for noting this issue again, @warrendholmes.

A text replacement is a good temporary bandaid solution.

Ideally we can implement https://github.com/woocommerce/wc-admin/issues/1358 and close this issue, but that may be an enhancement we plan after v1.

Yes.  The current notice implies that the action is complete, which I imagine will lead to a number of complaints when order data isn't fully populated yet.

### LINKED ISSUE #1691: Settings: Move Lookup Tables UI to Settings Page

joshuatf:
Fixes #1691  fixes #1641 

Moves the rebuild tool to the settings page and adds in 'button' type settings as well as callbacks for custom actions to settings.

### Screenshots
screen shot 2019-03-05 at 2 54 10 pm
screen shot 2019-03-05 at 2 54 04 pm

1.  Go to the settings page `/wp-admin/admin.php?page=wc-admin#/analytics/settings`
2.  Make sure that "Rebuild reports" rebuilds report data.
3.  The button should be disabled after clicking regenerate (currently for 5 minutes, but this has a todo noting that we should check for population completion).

### DIFF
--- a/client/analytics/settings/config.js
+++ b/client/analytics/settings/config.js
@@ -100,21 +100,17 @@ export const analyticsSettings = applyFilters( SETTINGS_FILTER, [
 			'wc-admin'
 		),
 		callback: ( resolve, reject, addNotice ) => {
-			const startedMessage = __(
-				'Report table data is being rebuilt.  Please allow some time for data to fully populate.',
-				'wc-admin'
-			);
 			const errorMessage = __( 'There was a problem rebuilding your report data.', 'wc-admin' );
 
 			apiFetch( { path: '/wc/v3/system_status/tools/rebuild_stats', method: 'PUT' } )
 				.then( response => {
 					if ( response.success ) {
-						addNotice( { status: 'success', message: startedMessage } );
+						addNotice( { status: 'success', message: response.message } );
 						// @todo This should be changed to detect when the lookup table population is complete.
 						setTimeout( () => resolve(), 300000 );
 					} else {
 						addNotice( { status: 'error', message: errorMessage } );
-						resolve();
+						reject();
 					}
 				} )
 				.catch( error => {
--- a/client/analytics/settings/config.js
+++ b/client/analytics/settings/config.js
@@ -3,6 +3,7 @@
  * External dependencies
  */
 import { __, sprintf } from '@wordpress/i18n';
+import apiFetch from '@wordpress/api-fetch';
 import { applyFilters } from '@wordpress/hooks';
 import interpolateComponents from 'interpolate-components';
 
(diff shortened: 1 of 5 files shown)


coderkevin:
I really like the idea of this, but it's strange to me to have a "setting" which isn't really a setting at all and just a trigger button. Also, if I click refresh, the button is clickable again.

Can I suggest that this button be a bona-fide setting of a "rebuilding mode" that can be toggled by the button? If I click the button, the site toggles "rebuilding mode" on, updates the setting as such, and at the completion of the rebuild, toggles "rebuilding mode" off again. This will fix the refresh issue above, and using our existing API data synchronization, it can pop out the button after the setting gets cleared by the rebuilding job.

LevinMedia:
@coderkevin This is mostly a placeholder for a more involved WIP that's running through the design process right now. In the fully fledged version the button would expand a panel to configure the data rebuild, and also provide more specific detail on the status of a rebuild job that's currently running. 

@joshuatf I think this works for the time being.

client/analytics/settings/config.js
```diff
@@ -65,4 +66,36 @@ export const analyticsSettings = applyFilters( SETTINGS_FILTER, [
...
+		inputText: __( 'Rebuild reports', 'wc-admin' ),
+		helpText: __(
+			'This tool will rebuild all of the information used by the reports. ' +
+				'Data will be processed in the background and may take some time depending on the size of your store.',
+			'wc-admin'
+		),
+		callback: ( resolve, reject, addNotice ) => {
+			const errorMessage = __( 'There was a problem rebuilding your report data.', 'wc-admin' );
+
+			apiFetch( { path: '/wc/v3/system_status/tools/rebuild_stats', method: 'PUT' } )
```
justinshreve: This can work for now if it's to involved (I know the needs of this tool might change per the comment above) -- but should we consider implementing a proper action/mutator for this?

joshuatf:
>Can I suggest that this button be a bona-fide setting of a "rebuilding mode" that can be toggled by the button?

@coderkevin Agreed and I believe this is the long-term plan, but this is more of a temporary fix to bring this tool to a more prominent location.  Here are the other issues outlining those improvements:
https://github.com/woocommerce/wc-admin/issues/1358
https://github.com/woocommerce/wc-admin/issues/1572

client/analytics/settings/config.js
joshuatf: I think that's probably a smart move long-term.  It may be something we want to wait for https://github.com/woocommerce/wc-admin/issues/1358 until we decide to add in since it will likely change this.

That also makes me think that while we can import/export the store here, I'm not sure the cleanest way of doing this is from a config file and this may require some refactoring in how settings are handled.

emdashcodes (approved):
Considering some of this is temporary and will be followed up on, this is looking good to me. I do have one comment about `remove_regenerate_tool` which I think would stick around, but pre-approving.

includes/class-wc-admin-reports-sync.php
```diff
@@ -103,6 +104,22 @@ public static function regenerate_report_data() {
...
+	}
+
+	/**
+	 * Remove the regenerate tool from system status tools while
+	 * still allowing use of the tool through the REST API.
+	 *
+	 * @param array $tools Array of tools.
+	 * @return array
+	 */
+	public static function remove_regenerate_tool( $tools ) {
```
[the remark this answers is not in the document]
joshuatf: 🤦‍♂️ Much more concise.  Thanks @justinshreve!  Updated in https://github.com/woocommerce/wc-admin/pull/1737/commits/44c207c8e31ed90c069e63475769f1b150750779

PR #1738, opened 2019-03-05
Update CSV filenames and remove item IDs

### LINKED ISSUE #1642: CSV filenames include a long list of unassociated IDs
**Describe the bug**
When adding 2 or more items to compare and downloading the CSV, the filename includes a list of IDs not relevant to the comparison.

Additionally, while we're here we might consider using underscore to separate params to make this a bit easier to read:
**Current**: param1-value1-param2-value2-param3-value3
**Suggested**: param1-value1_param2-value2_param3-value3

**To Reproduce**

1. Go to the products report.
2. Select 2 or more products to compare.
3.  Download the CSV.
4.  Note the filename.

**Expected behavior**
Expected the product IDs or names that were being compared.

**Screenshots**
Filename: `products-2019-02-20-orderby-items_sold-order-desc-extended_info-true-search-sed aliquid non voluptates non-products-282919,283158,283166,283392,283516`
screen shot 2019-02-20 at 4 52 14 pm

joshuatf:
Fixes #1642 

* Separates params by underscores
* Replaces spaces and underscores in values with hyphens
* Removes the list of IDs from the end of the string caused by found item IDs from search in favor of the search query

### Screenshots
screen shot 2019-03-05 at 6 15 21 pm
Produces:

`products_2019-03-05_orderby-net-revenue_order-desc_period-last-year_compare-previous-year_search-sed-aliquid-non-voluptates-non.csv`

1. Go to the any report that allows searching in the table.
2. Enter a search term that matches multiple items.
3. Download the CSV.
4. Check that the item IDs aren't in the filename, but the search term is.
5. Check that item names don't contain spaces.
6. Check that `extended_info` is not included in the filename.
7. Play around with various parameters and make sure filenames look correct.

### DIFF
--- a/packages/components/src/table/index.js
+++ b/packages/components/src/table/index.js
@@ -129,6 +129,7 @@ class TableCard extends Component {
 
 	onClickDownload() {
 		const { query, onClickDownload, title } = this.props;
+		delete query.extended_info;
 
 		// @todo The current implementation only downloads the contents displayed in the table.
 		// Another solution is required when the data set is larger (see #311).
--- a/packages/components/src/table/index.js
+++ b/packages/components/src/table/index.js
@@ -128,8 +128,13 @@ class TableCard extends Component {
 	}
 
 	onClickDownload() {
-		const { query, onClickDownload, title } = this.props;
+		const { query, onClickDownload, searchBy, title } = this.props;
+
+		// Delete unnecessary items from filename.
 		delete query.extended_info;
+		if ( query.search ) {
+			delete query[ searchBy ];
+		}
 
 		// @todo The current implementation only downloads the contents displayed in the table.
 		// Another solution is required when the data set is larger (see #311).
--- a/packages/components/src/table/index.js
+++ b/packages/components/src/table/index.js
@@ -129,17 +129,18 @@ class TableCard extends Component {
 
 	onClickDownload() {
 		const { query, onClickDownload, searchBy, title } = this.props;
+		const params = Object.assign( {}, query );
 
 		// Delete unnecessary items from filename.
-		delete query.extended_info;
-		if ( query.search ) {
-			delete query[ searchBy ];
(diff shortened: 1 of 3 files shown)


packages/components/src/table/index.js
```diff
@@ -128,7 +128,13 @@ class TableCard extends Component {
 	}
 
 	onClickDownload() {
-		const { query, onClickDownload, title } = this.props;
+		const { query, onClickDownload, searchBy, title } = this.props;
+
+		// Delete unnecessary items from filename.
+		delete query.extended_info;
```
Aljullu: I'm not sure about directly modifying `query`. I think we should clone it before, either using lodash [`omit`](https://lodash.com/docs/4.17.11#omit) or `Object.assign()`.

Aljullu (commented):
This is a big improvement! These functions were written long time ago and, as `wc-admin` progressed and added new parameters, they clearly needed an update. Thanks for working on it, @joshuatf!

I noticed searching for a product with a comma in its name shows `%2c` in the file name:
image
Could we remove this?

I also added a comment in the code, but overall this looks really good.

packages/components/src/table/index.js
coderkevin: I'd second this, we typically make copies and don't modify existing objects/arrays.

coderkevin (approved):
Tests good, and other than the comment about modifying props, looks good. Pre-approving

packages/components/src/table/index.js
joshuatf: Whoops, thanks for catching this.  Updated.

joshuatf:
Thanks for the review @Aljullu and @coderkevin. 

>I noticed searching for a product with a comma in its name shows %2c in the file name

Nice catch.  Decoded params and also switched the regex to replace any non-alphanumeric characters.

PR #1741, opened 2019-03-05
Don't Run Lookup Table Population on Activate

### LINKED ISSUE #1666: Setup: Do Not Run Lookup Table Population on Activate
Revising this issue a bit after talking with @LevinMedia. Instead of auto-running the lookup table population when the plugin is activated - we would like to make this an opt-in action where a user must make an informed choice to run the job(s). Since the population of the lookup tables could be quite resource intensive, we want to make sure users don't accidentally bog down their site by just installing the plugin.

In another issue we will have have some plans to inform users either via Priority Notification and/or an inbox notification to run the job to populate the tables.

emdashcodes:
Fixes #1666.

This PR stops look-up table queuing from happening on plugin activation.

* Delete the `wp_option` `wc_admin_version`, to make sure the `install` or `update` step runs. You can also add an error log between `self::create_tables()` and `self::update_wc_admin_version()` in the `install` method of `class-wc-admin-install.php`.
* Deactivate WooCommerce Admin, and make sure there are no pending `wc-admin` jobs.
* Reactivate WooCommerce Admin, make sure there are still no pending jobs.

### DIFF
--- a/includes/class-wc-admin-install.php
+++ b/includes/class-wc-admin-install.php
@@ -74,7 +74,6 @@ class WC_Admin_Install {
 		wc_maybe_define_constant( 'WC_ADMIN_INSTALLING', true );
 
 		self::create_tables();
-		WC_Admin_Reports_Sync::regenerate_report_data();
 		self::update_wc_admin_version();
 
 		delete_transient( 'wc_admin_installing' );


psealock (commented):
This looks good, just a question so I can fully test this:

>make sure there are no pending wc-admin jobs

I'm a bit out of the loop, how do you do this?

timmyc:
@psealock you can view all pending jobs via Action Scheduler's post list view under WC Status | Scheduled Actions `/wp-admin/admin.php?page=wc-status&tab=action-scheduler&status=pending`

psealock (approved):
Thanks @timmyc !

Testing well

Issue #1680, opened 2019-02-22
Move Single Product Category view from Product report to Category Report


LevinMedia:
The Single Product Category view is available in the Product report. I'd like to move it to the Category Report. 

**Single Product Category view in Product Report**

<img width="1524" alt="screen shot 2019-02-22 at 10 10 31 am" src="https://user-images.githubusercontent.com/4500952/53262007-5d499180-368a-11e9-9b3e-e4bcf5a8fdec.png">

The view above should be accessible when clicking Single Product Category (the label in the category report will need an update too) in the **Category Report** shown below:

<img width="1524" alt="screen shot 2019-02-22 at 10 11 30 am" src="https://user-images.githubusercontent.com/4500952/53262239-eeb90380-368a-11e9-9f52-db1b914d9f0b.png">

With the Single Product Category view moved to the Category Report, we can remove the option from the "Show" menu in the Product Report (relates to #1678)

<img width="1191" alt="screen shot 2019-02-22 at 10 24 51 am" src="https://user-images.githubusercontent.com/4500952/53262777-63407200-368c-11e9-991a-886f50a50f70.png">

Clicking category titles anywhere in the analytics experience should direct a user to the single product category view in the category report. 

cc @psealock @timmyc

PR #1748, opened 2019-03-06
Filters: Simplify options

psealock:
Fixes https://github.com/woocommerce/wc-admin/issues/1678
Fixes https://github.com/woocommerce/wc-admin/issues/1680

* Remove "Top *" from filters list for Coupons, Products, Categories, and Variations Reports.
* Remove "Single Category" and "Category Comparison" filters from Products Report.

The Categories Report uses the `/products/stats` endpoint under the hood. Since the Product Report had category related filters, the resulting filter query used them, but now it must use the Category filters config, so some endpoint mapping was required.

1. Coupons, Products, Categories, and Variations Reports should no longer have "Top *" options in the Show menu. For example, `Top Categories by Items Sold` will no longer be there.
2. Ensure Single Category" and "Category Comparison" filters from Products Report are no longer available but are on the Categories Report.
3. Check that links to categories from other reports do not point to the Products Report.

### DIFF
--- a/client/analytics/report/categories/config.js
+++ b/client/analytics/report/categories/config.js
@@ -79,18 +79,6 @@ export const filters = [
 					},
 				},
 			},
-			{
-				label: __( 'Top Categories by Items Sold', 'wc-admin' ),
-				value: 'top_items',
-				chartMode: 'item-comparison',
-				query: { orderby: 'items_sold', order: 'desc', chart: 'items_sold' },
-			},
-			{
-				label: __( 'Top Categories by Net Revenue', 'wc-admin' ),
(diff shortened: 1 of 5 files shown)


LevinMedia:
heya @psealock I noticed a bug here:

when switching from "All" to "Comparison", it automatically changes the chart to comparison mode: 

screen recording 2019-03-06 at 02 31 pm

that shouldn't happen until a user has selected at least two products and clicks compare.

LevinMedia:
Also noting the first portion of #1680 was missing from this PR. When viewing a single product category in the category report, you should see a list of all it's child products.

PR #1750, opened 2019-03-06
Fix deleted product error

### LINKED ISSUE #1669: Deleting a product after an order was made creates a 500 error
**Describe the bug**
If an order is made with a product and that product is subsequently permanently deleted, an error will be thrown on the products report.

Big thanks to @LevinMedia for finding this bug!

**To Reproduce**

1. Make an order.
2. Delete a product in that order (permanently).
3. Go to the products report.
4. Select a date range with that order in it.
5. Note the error.

**Expected behavior**
Expected to still be able to see the report.  For items that were deleted, we only store quantity and revenue amounts, no names, skus or other information.

@LevinMedia Should we continue to show the product but in place of unknown fields add something like `(Deleted)`?  Or do we exempt these rows from the products report?

**Screenshots**
screen shot 2019-02-22 at 12 56 51 pm

**Additional context**
The 500 error originates from a product not being found [here](https://github.com/woocommerce/wc-admin/blob/2d0b05c9ebd19746048164ddbaa0a639fe97b433/includes/data-stores/class-wc-admin-reports-products-data-store.php#L200).

In addition to checking if a product was found, we may also want to forego some of the JOINS in this data store if the `product_id` is 0.  

Finally, deleting a product should probably trigger updates to any orders it's in within the orders lookup table if we're checking for `0` values in `product_id` anywhere.

joshuatf:
Fixes #1669 

* Adds `(Deleted)` to the name of deleted products
* Prevents the 500 API error that occurs
* Grabs the ID for deleted products from the meta table to keep stats accurate

### Screenshots
screen shot 2019-03-06 at 2 29 05 pm

1. Create an order with a product
2. Delete that product permanently ☠️ 
3. View a timeframe in the products report with that product.
4. Update the order to trigger a sync.
5. Note the lack of error and the product name should display `(Deleted)`.
6. Repeat these steps with a second product.
7. Note that the ID for the products is still retained and that stats are not grouped together by an ID of `0`.

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-products-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-products-data-store.php
@@ -90,6 +90,7 @@ class WC_Admin_Reports_Products_Data_Store extends WC_Admin_Reports_Data_Store i
 	 */
 	public static function init() {
 		add_action( 'woocommerce_reports_delete_order_stats', array( __CLASS__, 'sync_on_order_delete' ), 10 );
+		add_action( 'delete_post', array( __CLASS__, 'delete_product_id' ) );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


Aljullu (approved):
Good catch with this issue! I tested and it works fine, code also looks good to me.

PR #1753, opened 2019-03-06
Fix deleted coupon data sync

### LINKED ISSUE #1751: Deleting a coupon creates empty coupon on re-sync
**Describe the bug**
When permanently deleting a coupon, 

**To Reproduce**

1. Create an order with a coupon.
2. Delete that coupon permanently.
3. Note the coupon name in the coupon report is now the ID of the coupon.
4. Save the order to trigger a resync.
5. Note that an extra row is created with a coupon ID of `0`.

**Expected behavior**
Expected a note that it was deleted similar to how we're handling products - `(Deleted)`.  Also expected accurate stats.

**Screenshots**
screen shot 2019-03-06 at 4 28 09 pm

joshuatf:
Fixes #1751 

Removes coupon data from lookup tables if permanently deleted and prevents coupon ID without an ID from syncing on order sync.

Initially, I tried to grab this under `wp_woocommerce_order_itemmeta` as it appears the ID is still stored under a `coupon_data` meta key.  However, the coupon data seemed to be corrupted (could not unserialize).  Curious if similar behavior has occurred in other installs.

Ideally, we can grab the ID from `coupon_data` since that would allow us to show `(Deleted)` and keep the coupon data information, but if it acts unreliably as it did in my environment then I think we should remove altogether when permanently deleted.

1.  Create an order with a coupon.
2.  Delete the coupon permanently.
3.  Check that the data does not show stats for that coupon under the coupons report.
4.  Update  the order with the deleted coupon to trigger an order sync.
5.  Make sure an emtpy (ID-less) coupon is not created under the coupon report.
6.  Make sure that order data and discounts does not reflect the deletion of the coupon.

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-coupons-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-coupons-data-store.php
@@ -56,6 +56,7 @@ class WC_Admin_Reports_Coupons_Data_Store extends WC_Admin_Reports_Data_Store im
 	 */
 	public static function init() {
 		add_action( 'woocommerce_reports_delete_order_stats', array( __CLASS__, 'sync_on_order_delete' ), 5 );
+		add_action( 'delete_post', array( __CLASS__, 'delete_coupon' ) );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


coderkevin:
I tried this out, and looked at the report to see the coupon I created, then I deleted the coupon, and when I come back to the report I get this:
coupons_ _analytics_ _two_wordpress_test_ _woocommerce

I'm guessing I shouldn't see that row, right?

psealock (approved):
This one is working for me and the code looks good. I can't reproduce what @coderkevin is experiencing. I'll pre-approve unless that issue continues.

>Ideally, we can grab the ID from coupon_data since that would allow us to show (Deleted) and keep the coupon data information, but if it acts unreliably as it did in my environment then I think we should remove altogether when permanently deleted.

I'm for removing altogether. The behaviour makes sense to me, especially since trashing a Coupon without "permanently deleting" it retains all the analytics.

joshuatf:
I'm also unable to reproduce that issue.  @coderkevin can you check if you're still hitting this issue and if so are there any other details you can provide that might help to reproduce?  Thanks!

psealock:
I gave this one another test and its all working well for me. I think we should go ahead an merge this one and raise issues as they arise.

PR #1754, opened 2019-03-06
Show compared keys in chart legends even if their values are 0

Aljullu:
Fixes the chart part of  #1704.

Makes sure compared items are shown in charts even if their values are 0.

#### Product Comparison with selected items
_Before:_
image

_After:_
image

#### Product Comparison without selected items
image

- Go to the _Products_ report or any other with a chart and which allows comparisons.
- Select _Product Comparison_ filter without choosing any product yet.
- Verify only items with _values > 0_ appear in the chart legend (see screenshot above).
- Now select some products using the filter text input, selecting items with _values > 0_ and others with _values = 0_ for that period.
- Verify all selected items, including the ones with _values = 0_ appear in the chart.


coderkevin (approved):
Checks out and looks good!

PR #1755, opened 2019-03-06
Make categories endpoint return zero-value items

Aljullu:
Part of #1668.

Follow-up of #1722.

Makes `categories` endpoint return zero-value items.

### Screenshots
_(Notice _T-Shirts_ appears in the table even though its values are 0)_
image

- Start a _Comparison_ making sure one of the categories has all values to 0 (it has not been sold during the selected date range).
- Verify the table shows all compared categories, even the one that has all values to 0.

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-categories-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-categories-data-store.php
@@ -259,7 +259,7 @@ class WC_Admin_Reports_Categories_Data_Store extends WC_Admin_Reports_Data_Store
 			if ( count( $included_categories ) > 0 ) {
 				$fields          = $this->get_fields( $query_args );
 				$join_selections = $this->format_join_selections( array_merge( array( 'category_id' ), $fields ), 'category_id' );
-				$ids_table       = $this->get_ids_table( $included_categories, 'category_id' );
+				$ids_table       = $this->get_ids_table( $included_categories );
 
 				$prefix     = "SELECT {$join_selections} FROM (";
 				$suffix     = ") AS {$table_name}";
--- a/includes/data-stores/class-wc-admin-reports-categories-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-categories-data-store.php
@@ -119,11 +119,11 @@ class WC_Admin_Reports_Categories_Data_Store extends WC_Admin_Reports_Data_Store
 	 *
 	 * @param array  $query_args Parameters supplied by the user.
 	 * @param array  $sql_query  Current SQL query array.
-	 * @param string $arg_name   Name of the FROM sql param.
+	 * @param string $from_arg   Name of the FROM sql param.
 	 * @param string $id_cell    ID cell identifier, like `table_name.id_column_name`.
 	 * @return array
 	 */
-	protected function get_order_by_params( $query_args, $sql_query, $arg_name, $id_cell ) {
+	protected function get_order_by_params( $query_args, $sql_query, $from_arg, $id_cell ) {
 		global $wpdb;
 		$lookup_table = $wpdb->prefix . self::TABLE_NAME;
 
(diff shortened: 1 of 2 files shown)


includes/data-stores/class-wc-admin-reports-categories-data-store.php
```diff
@@ -165,12 +182,8 @@ protected function normalize_order_by( $order_by ) {
 	 * @return string
 	 */
 	protected function get_included_categories( $query_args ) {
```
psealock: How about renaming to `get_included_categories_string` to differentiate from `get_included_categories_array`?

psealock (commented):
This is testing well. That is some impressive query-fu going on there. I'll defer to someone with a bit more experience for final approval, but its working well for me.

tests/api/reports-categories.php
```diff
@@ -84,6 +84,72 @@ public function test_get_reports() {
...
+		$second_category_id = wp_create_category( 'Second Category' );
+		$product->set_category_ids( array( $second_category_id ) );
+		$product->save();
+
+		WC_Helper_Queue::run_all_pending();
+
+		$uncategorized_term = get_term_by( 'slug', 'uncategorized', 'product_cat' );
+
+		$response = $this->server->dispatch( new WP_REST_Request( 'GET', $this->endpoint ) );
+		$reports  = $response->get_data();
```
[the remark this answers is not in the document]
Aljullu: Right! Fixed in a1374b5a98a1b685c4f988ee2049b424c84bb6ba.

includes/data-stores/class-wc-admin-reports-categories-data-store.php
Aljullu: Thanks for taking a look @psealock!

> How about renaming to `get_included_categories_string` to differentiate from `get_included_categories_array`?

I wondered the same, but I decided not to change it to keep it consistent with the methods defined in [`class-wc-admin-reports-data-store.php`](includes/data-stores/class-wc-admin-reports-data-store.php) (`get_included_products`, `get_included_variations`, `get_excluded_products`, `get_included_coupons`...). Otherwise, to keep them consistent, we should add `_string` at the end of the name of all of them, but I think that would just add noise to many other parts of the codebase.

What do you think?

psealock: Sounds legit to me.

includes/data-stores/class-wc-admin-reports-categories-data-store.php
```diff
@@ -239,11 +252,28 @@ public function get_data( $query_args ) {
...
-			$selections       = $this->selected_columns( $query_args );
-			$sql_query_params = $this->get_sql_query_params( $query_args );
+			$selections          = $this->selected_columns( $query_args );
+			$sql_query_params    = $this->get_sql_query_params( $query_args );
+			$included_categories = $this->get_included_categories_array( $query_args );
+
+			if ( count( $included_categories ) > 0 ) {
+				$fields          = $this->get_fields( $query_args );
+				$join_selections = $this->format_join_selections( array_merge( array( 'category_id' ), $fields ), 'category_id' );
+				$ids_table       = $this->get_ids_table( $included_categories, 'category_id' );
```
psealock: `get_ids_table` accepts only 1 argument. I see other uses also have 2 🤷‍♂️ .

psealock (approved):
Just one comment about an extra argument passed to `get_ids_table`. Otherwise I see the same pattern in https://github.com/woocommerce/wc-admin/pull/1722 and passing unit tests.

includes/data-stores/class-wc-admin-reports-categories-data-store.php
Aljullu: Thanks for taking another look @psealock! This is fixed in 92f6c189995e33384d523788e3524d5c1d5ef069.

PR #1758, opened 2019-03-06
Add StoreAlerts placeholder

### LINKED ISSUE #1736: Store Alerts: Add Alert Count to wcSettings / Loading Indicator
Follow-up issue for #1723 

Currently the Store Alerts are being loaded via the data layer / API after the page loads, which results in the alerts loading after the initial page render. This creates a bit of a "jumpy" experience. As a good in-between, I'm suggesting we consider doing the following:

- Add an element to `wcSettings` - perhaps called `alertCount` - which is just a count of the Store Alerts ( notes ) that are present for the store.
- Then in the app, if that setting number is > 0, we render a placeholder where the Store Alerts are displayed with a loading indicator ( not sure if we have designs for this, but we could likely wing it based off other loading indicator designs ). Then the current logic that requests the alerts from the API will render the content when it arrives.

tiagonoronha:
Fixes #1736 

* Adds alert count via the `wcSettings` global and logic to stop alerts from trying to load when no alerts exist.
* Adds `StoreAlertsPlaceholder` a placeholder component and logic to whether to display the placeholder or `StoreAlerts`.

### Screenshots

alerts-placeholder

- Create a couple of test alerts. See https://github.com/woocommerce/wc-admin/pull/1723 for quick snippet to add a new note.
- Reload a wc-admin page, you should see the placeholder and then the alerts after loading.

### DIFF
--- a/client/layout/store-alerts/index.js
+++ b/client/layout/store-alerts/index.js
@@ -20,6 +20,7 @@ import { Card } from '@woocommerce/components';
 import withSelect from 'wc-api/with-select';
 import { QUERY_DEFAULTS } from 'wc-api/constants';
 import sanitizeHTML from 'lib/sanitize-html';
+import StoreAlertsPlaceholder from './placeholder';
 
 import './style.scss';
 
@@ -60,12 +61,13 @@ class StoreAlerts extends Component {
 	}
 
 	render() {
-		const { alerts } = this.props;
+		const hasAlerts = Boolean( wcSettings.alertCount ) || null;
 
-		if ( ! alerts || alerts.length === 0 ) {
+		if ( ! hasAlerts ) {
 			return null;
 		}
 
+		const { alerts } = this.props;
 		const { currentIndex } = this.state;
 		const numberOfAlerts = alerts.length;
 		const alert = alerts[ currentIndex ] ? alerts[ currentIndex ] : null;
@@ -82,7 +84,7 @@ class StoreAlerts extends Component {
 				</Button>
 			) );
 
-		return (
+		return alert ? (
 			<Card
 				title={ [
 					alert.icon && <Dashicon key="icon" icon={ alert.icon } />,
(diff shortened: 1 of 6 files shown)


lib/client-assets.php
```diff
@@ -190,7 +190,8 @@ function wc_admin_print_script_settings() {
 			'userLocale'    => get_user_locale(),
 			'weekdaysShort' => array_values( $wp_locale->weekday_abbrev ),
 		),
-		'currentUserData'       => $current_user_data,
+		'currentUserData'  => $current_user_data,
+		'alertCount'       => WC_Admin_Notes::get_notes_count(),
```
jeffstieler: This needs to get restricted to just the count of `error` and `update` type notes - otherwise the placeholder can display even though no note will be shown.

client/layout/store-alerts/index.js
```diff
@@ -127,12 +129,20 @@ class StoreAlerts extends Component {
...
+			<StoreAlertsPlaceholder hasMultipleAlerts={ wcSettings.alertCount > 1 } />
 		);
 	}
 }
 
 export default compose(
 	withSelect( select => {
+		const { alertCount } = wcSettings;
+
+		if ( ! alertCount || 0 === alertCount ) {
```
jeffstieler: I think we should eliminate this check - if there are no notes, it will never check for new notes. The current behavior respects the `freshness` interval and periodically checks for new notes.

lib/client-assets.php
tiagonoronha: @jeffstieler I mentioned that in the PR. I updated the method with support for filtering by note type: https://github.com/woocommerce/wc-admin/pull/1758/commits/7e381f4e351a145d02756bb3edede9b434b428c6

client/layout/store-alerts/index.js
tiagonoronha: That makes sense, reverted that commit.

PR #1759, opened 2019-03-07
Ensure that other scheduled actions are claimed before WC Admin actions.

jeffstieler:
This PR seeks to lower the priority of WC Admin scheduled actions so others are claimed first by Action Scheduler. This is an effort to prevent reporting table generation from stalling out important things like webhooks and subscription renewals.

It uses `menu_order` to lower the claim priority of WC Admin scheduled actions. (Exploits this `ORDER BY`: https://github.com/woocommerce/woocommerce/blob/e9f04908293048cf3a154a0771f5db61b01258e2/includes/libraries/action-scheduler/classes/ActionScheduler_wpPostStore.php#L545)

**NOTE:** This doesn't guarantee that non-WC Admin actions will run first, it just ensures that non-WC Admin actions get _claimed_ first. A single claim could still contain WC Admin actions that might have been scheduled before the others.

Testing this is a little tricky.. here's what I did:

- Queue up a WC Admin job: `WC_Admin_Reports_Sync::queue_action( time(), WC_Admin_Reports_Sync::SINGLE_ORDER_ACTION );`
- Queue up a non-WC Admin job, making sure it's definitely **after** the first: `WC()->queue()->schedule_single( time() + 10, 'some_hook' );`
- Use WP CLI to run a batch of 1: `wp action-scheduler --batch-size=1`
- Verify that the reported action ID matches the non-WC Admin action (checking DB is easy)

### DIFF
--- a/wc-admin.php
+++ b/wc-admin.php
@@ -181,3 +181,20 @@ add_action( 'woocommerce_updated', 'wc_admin_woocommerce_updated' );
  * Gutenberg has also disabled emojis. More on that here -> https://github.com/WordPress/gutenberg/pull/6151
  */
 remove_action( 'admin_print_scripts', 'print_emoji_detection_script' );
+
+/**
+ * Filter in our ActionScheduler Store class.
+ *
+ * @return string ActionScheduler Store class name.
+ */
+function wc_admin_set_actionscheduler_store_class() {
+	// Include our store class here instead of wc_admin_plugins_loaded()
+	// because ActionScheduler is hooked into `plugins_loaded` at a
+	// much higher priority.
+	require_once WC_ADMIN_ABSPATH . '/includes/class-wc-admin-actionscheduler-wppoststore.php';
+
+	return 'WC_Admin_ActionScheduler_WPPostStore';
+}
+
+// Hook up our modified ActionScheduler Store.
+add_filter( 'action_scheduler_store_class', 'wc_admin_set_actionscheduler_store_class' );
--- a/wc-admin.php
+++ b/wc-admin.php
@@ -185,9 +185,15 @@ remove_action( 'admin_print_scripts', 'print_emoji_detection_script' );
 /**
  * Filter in our ActionScheduler Store class.
  *
+ * @param string $store_class ActionScheduler Store class name.
  * @return string ActionScheduler Store class name.
  */
-function wc_admin_set_actionscheduler_store_class() {
+function wc_admin_set_actionscheduler_store_class( $store_class ) {
+	// Don't override any other overrides.
+	if ( 'ActionScheduler_wpPostStore' !== $store_class ) {
+		return $store_class;
+	}
+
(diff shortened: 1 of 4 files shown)


rrennick:
This PR works fine. However, the idea I had in mind was extending the post data store & overriding the `create_post_array` to add the `menu_order` field. Extending the class would be a step toward other AS related customizations as well.

jeffstieler:
> This PR works fine. However, the idea I had in mind was extending the post data store & overriding the `create_post_array` to add the `menu_order` field. Extending the class would be a step toward other AS related customizations as well.

I agree that a subclass provides a better home for AS customizations (like for #1699). Are we OK with only being able to support our own AS data store? _(cc: @timmyc)_

I can think of a case where a store is using its own subclass to make minor tweaks (like modifying `create_post_array`) but not changing the WP post storage mechanism. In that case our priority hack would break if it were only implemented using a subclass of `ActionScheduler_Store`.

rrennick:
> I can think of a case where a store is using its own subclass to make minor tweaks

Anyone using their own post store is likely to be extending `ActionScheduler_Store` vs `ActionScheduler_wpPostStore`. They are unlikely to be running on lower end hosting because small stores would not see much benefit from AS optimization. I think, where possible, the first optimization used is WP-CLI as it only requires setting up a cron job.

From my perspective, we should only be trying to add the lower priority to sites using the default data store.

jeffstieler:
Refactored to use a `ActionScheduler_wpPostStore ` subclass in dce9132.

timmyc:
Code is looking and testing well for me ( thanks for the instructions and unit tests ). I guess I'm having a hard time following the concerns around having our own subclass.

> Are we OK with only being able to support our own AS data store? 

Specifically that - do you have an example of where we would need to interact with another AS store outside of our own?

wc-admin.php
```diff
@@ -181,3 +181,20 @@ function wc_admin_woocommerce_updated() {
...
+ *
+ * @return string ActionScheduler Store class name.
+ */
+function wc_admin_set_actionscheduler_store_class() {
+	// Include our store class here instead of wc_admin_plugins_loaded()
+	// because ActionScheduler is hooked into `plugins_loaded` at a
+	// much higher priority.
+	require_once WC_ADMIN_ABSPATH . '/includes/class-wc-admin-actionscheduler-wppoststore.php';
+
+	return 'WC_Admin_ActionScheduler_WPPostStore';
```
rrennick: This should check that the passed value is the AS post data store so it doesn't break a site using a custom store.

rrennick: @jeffstieler Outside of my note above, this looks great and works as expected.

rrennick (approved):

wc-admin.php
jeffstieler: Added a class check in 1ca9bf1.

rrennick:
@jeffstieler LGTM, okay to merge.

Issue #1760, opened 2019-03-07
Show > Comparison switches chart to "comparison" mode too early


psealock:
Comparison mode should only be activated once two items have been chosen.

This needs to be fixed on Products, Coupons, and Categories Reports.

PR #1762, opened 2019-03-07
Fix  chart comparison mode and single Category

psealock:
Fixes https://github.com/woocommerce/wc-admin/issues/1761
Fixes https://github.com/woocommerce/wc-admin/issues/1760

There are two changes in this PR:

Show > Comparison now doesn't change anything until two items are selected and "compare" is clicked. The chart will then switch to "comparison" mode.

Secondly, Categories Report > Show > Single Category now produces segmentation by products in the Chart in comparison mode and uses the Products Table to display tabular results.

1. Categories, Products, or Coupons Report > Show > Comparison.
2. Ensure no change in the chart.
3. Select at least 2 values and click "compare"
4. See the charts update to comparison mode.
5. Categories Report > Show > Single Category and make a selection.
6. See the chart enter comparison mode of products and the table display product results
7. Search in the table for a product, including a fuzzy search. Ensure the chart and summary numbers update accordingly.

### DIFF
--- a/client/analytics/report/categories/index.js
+++ b/client/analytics/report/categories/index.js
@@ -19,18 +19,24 @@ import CategoriesReportTable from './table';
 import getSelectedChart from 'lib/get-selected-chart';
 import ReportChart from 'analytics/components/report-chart';
 import ReportSummary from 'analytics/components/report-summary';
+import ProductsReportTable from '../products/table';
 
 export default class CategoriesReport extends Component {
 	getChartMeta() {
 		const { query } = this.props;
+		const isCompareView =
+			'compare-categories' === query.filter &&
+			query.categories &&
+			query.categories.split( ',' ).length > 1;
+		const isSingleCategoryView = 'single_category' === query.filter && !! query.categories;
 
-		const isCategoryDetailsView = [ 'top_items', 'top_revenue', 'compare-categories' ].includes(
-			query.filter
-		);
-		const mode = isCategoryDetailsView ? 'item-comparison' : 'time-comparison';
-		const itemsLabel = __( '%d categories', 'wc-admin' );
+		const mode = isCompareView || isSingleCategoryView ? 'item-comparison' : 'time-comparison';
+		const itemsLabel = isSingleCategoryView
+			? __( '%d products', 'wc-admin' )
+			: __( '%d categories', 'wc-admin' );
 
 		return {
+			isSingleCategoryView,
 			itemsLabel,
 			mode,
 		};
@@ -38,14 +44,14 @@ export default class CategoriesReport extends Component {
 
 	render() {
 		const { isRequesting, query, path } = this.props;
-		const { mode, itemsLabel } = this.getChartMeta();
+		const { mode, itemsLabel, isSingleCategoryView } = this.getChartMeta();
 
 		const chartQuery = {
 			...query,
 		};
 
 		if ( 'item-comparison' === mode ) {
-			chartQuery.segmentby = 'category';
+			chartQuery.segmentby = isSingleCategoryView ? 'product' : 'category';
 		}
 
 		return (
@@ -55,7 +61,7 @@ export default class CategoriesReport extends Component {
 					charts={ charts }
 					endpoint="categories"
 					isRequesting={ isRequesting }
-					limitProperty="categories"
+					limitProperties={ isSingleCategoryView ? [ 'products', 'categories' ] : [ 'categories' ] }
 					query={ chartQuery }
 					selectedChart={ getSelectedChart( query.chart, charts ) }
 				/>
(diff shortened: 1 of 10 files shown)


client/analytics/report/products/table.js
```diff
@@ -248,6 +248,12 @@ class ProductsReportTable extends Component {
 					orderby: query.orderby || 'items_sold',
 					order: query.order || 'desc',
 					extended_info: true,
+					/**
+					 * @TODO: Add this parameter because the filterQuery will be derived from the wrong config
+					 * because it will always look for products config, not categories. The solution is to pass
+					 * down the configs explicitly.
+					 */
+					categories: query.categories,
```
psealock: Previously in https://github.com/woocommerce/wc-admin/pull/1748 I solved this by mapping endpoints such that filterQuery can use the endpoint to know which config to use according to this map:

https://github.com/woocommerce/wc-admin/blob/31a8e98059f06f1624307f85e12efffba4d04dd0/client/wc-api/reports/utils.js#L20-L37

In a follow up PR, I'd like to pass in the config instead of using this lookup to avoid mapping and any resulting confusion.

includes/class-wc-admin-reports-segmenting.php
```diff
@@ -311,6 +311,7 @@ protected function set_all_segments() {
 				'limit'  => -1,
 			);
 
+			// @todo: filter by categories if $this->query_args['categories'] is set.
 			if ( isset( $this->query_args['product_includes'] ) ) {
```
psealock: When segmenting by products and filtering by `categories`, all products are returned with zero'ed values, which results in oddness in the chart seen below (Sunglasses are not in the Hoodies category)

screen shot 2019-03-07 at 6 21 39 pm

I'd like to address this in a follow up PR.

client/analytics/report/categories/index.js
```diff
@@ -19,33 +19,37 @@ import CategoriesReportTable from './table';
...
+import ProductsReportTable from '../products/table';
 
 export default class CategoriesReport extends Component {
 	getChartMeta() {
 		const { query } = this.props;
+		const isCompareView =
+			'compare-categories' === query.filter &&
+			query.categories &&
+			query.categories.split( ',' ).length > 1;
+		const singleCategoryView = 'single_category' === query.filter && !! query.categories;
```
Aljullu: What do you think about renaming `singleCategoryView` to `isSingleCategoryView`? That way it would be clearer it's a boolean.

psealock:
>Regarding #1760, I thought that behavior was intentional. @justinshreve, do you have any info about it?

Yes, sorry I forgot to mention that I had a conversation with @LevinMedia regarding this. The current behavior was my understanding, but it turns out to not be the desired one.

LevinMedia:
Paul, looking good - one small update - we need to change "categories" to "Products" in the chart, when you're viewing all of the products in a single categoy:

screen shot 2019-03-07 at 2 20 11 pm

LevinMedia:
When I search for a product in single category view, it's exiting the single category view, and only showing me the parent category:

https://cl.ly/92bcdc0d6e4e/Screen%20Recording%202019-03-07%20at%2002.26%20PM.gif

psealock:
@LevinMedia Thanks for taking this through the wringer. I updated such that search in the table now filters the product table while updating the chart and summary numbers accordingly.

If we are going the route of showing products in the Single Category, the behaviour here seems most natural to me and I like depth of information compared to simply showing one single row of data for the single category.

I did make a decision to hide the "Compare" button when looking at products in the table because it doesn't make sense in the context of the Categories Report.

Alternatively, when a user selects two or more products, we can have the "Compare" button jump to the Products Report in Comparison mode. I like this as a feature, but I'm not sure if it makes sense to navigate to a completely different report. Thoughts? Maybe we could change the language to "Compare Products" to signify a redirection to a different report.

client/wc-api/reports/utils.js
```diff
@@ -197,7 +198,7 @@ function getRequestQuery( endpoint, dataType, query, limitBy ) {
  * @param  {String} endpoint  Report API Endpoint
  * @param  {Object} query     Query parameters in the url
  * @param  {Object} select    Instance of @wordpress/select
- * @param  {String} [limitBy] Property used to limit the results. It will be used in the API call to send the IDs.
+ * @param  {Array} [limitBy] Properties used to limit the results. It will be used in the API call to send the IDs.
```
Aljullu: Same.

client/wc-api/reports/utils.js
```diff
@@ -242,7 +243,7 @@ export function getSummaryNumbers( endpoint, query, select, limitBy ) {
  * @param  {String} dataType  'primary' or 'secondary'
  * @param  {Object} query     Query parameters in the url
  * @param  {Object} select    Instance of @wordpress/select
- * @param  {String} [limitBy] Property used to limit the results. It will be used in the API call to send the IDs.
+ * @param  {Array} [limitBy] Properties used to limit the results. It will be used in the API call to send the IDs.
```
Aljullu: And here too. :upside_down_face:

LevinMedia:
Heya, this is working out well for me now - I think your move with the compare button is fine for now. Thanks!

PR #1763, opened 2019-03-07
Add store notice around processing historical data.

### LINKED ISSUE #1709: Notes: Populate Lookup Tables
Since we will no longer be populating the lookup tables automatically on install [ blocked sort of by #1666 ], we should also utilize the new Store Alert component to inform users to do so.

[How to create notes documentation](https://github.com/woocommerce/wc-admin/blob/master/docs/examples/activity-panel-inbox.md)

For the content of the note, how about we go with:

Title: "Reporting tables need to be updated"
Body: "In order to use the new analytics features of WooCommerce Admin, the reporting database tables need to be populated with historical data for your store. All new orders and customers will automatically be viewable within the reports, but in order to view past orders and customers, you will need to populate the lookup tables."
Action Button: "Update Report Tables" - which will run `WC_Admin_Reports_Sync::regenerate_report_data();`

/cc @LevinMedia for copy check

Eventually I think we should also link to documentation on how to configure Action Scheduler for higher volume sites.

emdashcodes:
Fixes #1709.

This PR displays a store notice about processing historical report data. It also updates the note REST API to support filtering by status, so that we can fetch only `unactioned` store alerts (otherwise they would never get dismissed).

One thing to note is that the currently, the alert does not get dismissed unless you refresh the page.

I tried to investigate a couple ways to fix this behavior. 1) Immediately dismissing the alert when "get started" is clicked. The user didn't actually take action yet so this feels incorrect. 2) Adding a REST API update handler and updating the status of the note via JS when the process button is clicked from the settings page 3) hacking in a way to refresh the request by passing a timestamp to the notes query.

None of those options felt good to me. It seems ideally fresh-data should have a callable `refresh` method to force a fresh-data update on a certain query (allowing us to do something like like [refreshPost](https://github.com/WordPress/gutenberg/blob/c29b729a6d107fc5a7ef53efafda126d6fbb68e0/docs/designers-developers/developers/data/data-core-editor.md#refreshpost) on alerts). cc @coderkevin is this something that could be added to `fresh-data`? Any other ideas here? It seems like there may be other instances in the future where we will want to be able to requery something.

### Screenshots

screen shot 2019-03-06 at 10 42 38 pm

To Test:
* Run `phpunit` and make sure all tests pass
* Exit and restart `npm start` so you get the correct dev flags (the store alert flag was recently added)
* Deactivate WooCommerce Admin
* Delete the `wc_admin_version` option from `wp_options`, so that the activation/install hooks kick in
* Reactivate WooCommerce Admin
* Visit the dashboard and you should see the notice

### DIFF
--- a/client/analytics/settings/config.js
+++ b/client/analytics/settings/config.js
@@ -47,14 +47,13 @@ export const analyticsSettings = applyFilters( SETTINGS_FILTER, [
 				'Data will be processed in the background and may take some time depending on the size of your store.',
 			'wc-admin'
 		),
-		callback: ( resolve, reject, addNotice, refreshStoreAlerts ) => {
+		callback: ( resolve, reject, addNotice ) => {
 			const errorMessage = __( 'There was a problem rebuilding your report data.', 'wc-admin' );
 
 			apiFetch( { path: '/wc/v3/system_status/tools/rebuild_stats', method: 'PUT' } )
 				.then( response => {
 					if ( response.success ) {
 						addNotice( { status: 'success', message: response.message } );
-						refreshStoreAlerts();
 						// @todo This should be changed to detect when the lookup table population is complete.
 						setTimeout( () => resolve(), 300000 );
 					} else {
--- a/client/analytics/settings/config.js
+++ b/client/analytics/settings/config.js
@@ -38,6 +38,39 @@ const orderStatuses = Object.keys( wcSettings.orderStatuses )
 
 export const analyticsSettings = applyFilters( SETTINGS_FILTER, [
 	{
+		name: 'woocommerce_rebuild_reports_data',
+		label: __( 'Rebuild reports data:', 'wc-admin' ),
+		inputType: 'button',
+		inputText: __( 'Rebuild reports', 'wc-admin' ),
+		helpText: __(
+			'This tool will rebuild all of the information used by the reports. ' +
+				'Data will be processed in the background and may take some time depending on the size of your store.',
+			'wc-admin'
+		),
+		callback: ( resolve, reject, addNotice, refreshStoreAlerts ) => {
+			const errorMessage = __( 'There was a problem rebuilding your report data.', 'wc-admin' );
+
+			apiFetch( { path: '/wc/v3/system_status/tools/rebuild_stats', method: 'PUT' } )
+				.then( response => {
+					if ( response.success ) {
+						addNotice( { status: 'success', message: response.message } );
+						refreshStoreAlerts();
+						// @todo This should be changed to detect when the lookup table population is complete.
+						setTimeout( () => resolve(), 300000 );
+					} else {
+						addNotice( { status: 'error', message: errorMessage } );
+						reject();
+					}
+				} )
+				.catch( error => {
+					if ( error && error.message ) {
+						addNotice( { status: 'error', message: error.message } );
+					}
+					reject();
+				} );
+		},
+	},
+	{
 		name: 'woocommerce_excluded_report_order_statuses',
 		label: __( 'Excluded Statuses:', 'wc-admin' ),
 		inputType: 'checkboxGroup',
@@ -88,37 +121,5 @@ export const analyticsSettings = applyFilters( SETTINGS_FILTER, [
 		),
 		initialValue: wcSettings.wcAdminSettings.woocommerce_actionable_order_statuses || [],
 		defaultValue: [ 'processing', 'on-hold' ],
-	},
-	{
-		name: 'woocommerce_rebuild_reports_data',
-		label: __( 'Rebuild reports data:', 'wc-admin' ),
-		inputType: 'button',
-		inputText: __( 'Rebuild reports', 'wc-admin' ),
-		helpText: __(
-			'This tool will rebuild all of the information used by the reports. ' +
-				'Data will be processed in the background and may take some time depending on the size of your store.',
-			'wc-admin'
-		),
-		callback: ( resolve, reject, addNotice ) => {
-			const errorMessage = __( 'There was a problem rebuilding your report data.', 'wc-admin' );
-
-			apiFetch( { path: '/wc/v3/system_status/tools/rebuild_stats', method: 'PUT' } )
-				.then( response => {
-					if ( response.success ) {
-						addNotice( { status: 'success', message: response.message } );
-						// @todo This should be changed to detect when the lookup table population is complete.
-						setTimeout( () => resolve(), 300000 );
-					} else {
-						addNotice( { status: 'error', message: errorMessage } );
-						reject();
-					}
-				} )
-				.catch( error => {
-					if ( error && error.message ) {
-						addNotice( { status: 'error', message: error.message } );
-					}
-					reject();
-				} );
-		},
 	},
 ] );
--- a/includes/api/class-wc-admin-rest-admin-notes-controller.php
+++ b/includes/api/class-wc-admin-rest-admin-notes-controller.php
@@ -124,6 +124,12 @@ class WC_Admin_REST_Admin_Notes_Controller extends WC_REST_CRUD_Controller {
 			$args['type'] = $type;
 		}
 
+		$status = isset( $request['status'] ) ? $request['status'] : '';
+		$status = sanitize_text_field( $status );
+		if ( ! empty( $status ) ) {
+			$args['status'] = $status;
+		}
+
 		$notes = WC_Admin_Notes::get_notes( 'edit', $args );
 
 		$data = array();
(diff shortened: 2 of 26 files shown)


includes/class-wc-admin-notes-historical-data.php
```diff
@@ -0,0 +1,72 @@
...
+
+	/**
+	 * Creates a note for regenerating historical data.
+	 */
+	public static function add_note() {
+		$data_store = WC_Data_Store::load( 'admin-note' );
+
+		// First, see if we've already created this kind of note so we don't do it again.
+		$note_ids = $data_store->get_notes_with_name( self::NOTE_NAME );
+		foreach ( (array) $note_ids as $note_id ) {
```
timmyc: Could we just check to see if `$note_ids` is not empty here instead of looping? If there are any notes with the name we could assume it has already been shown right?

includes/class-wc-admin-notes-historical-data.php
```diff
@@ -0,0 +1,72 @@
...
+			$content_data = $note->get_content_data();
+			if ( property_exists( $content_data, 'getting_started' ) ) {
+				return;
+			}
+		}
+
+		$note = new WC_Admin_Note();
+		$note->set_title( __( 'WooCommerce Admin: Historical Analytics Data', 'wc-admin' ) );
+		$note->set_content( __( 'To view your historical analytics data, you must process your existing orders and customers.', 'wc-admin' ) );
+		$note->set_content_data(
```
timmyc: I'm not super savvy with Notes but outside of the check in the loop above, do we have to include `content_data`?

client/analytics/settings/config.js
```diff
@@ -37,6 +37,39 @@ const orderStatuses = Object.keys( wcSettings.orderStatuses )
...
+		inputType: 'button',
+		inputText: __( 'Rebuild reports', 'wc-admin' ),
+		helpText: __(
+			'This tool will rebuild all of the information used by the reports. ' +
+				'Data will be processed in the background and may take some time depending on the size of your store.',
+			'wc-admin'
+		),
+		callback: ( resolve, reject, addNotice, refreshStoreAlerts ) => {
+			const errorMessage = __( 'There was a problem rebuilding your report data.', 'wc-admin' );
+
```
timmyc: I think we should have a TODO here to track the usage of this callback.

includes/class-wc-admin-notes-historical-data.php
```diff
@@ -0,0 +1,72 @@
...
+		);
+
+		$note->save();
+	}
+
+	/**
+	 * Marks a note as actioned so that it does not show up again.
+	 */
+	public static function mark_actioned() {
+		$data_store = WC_Data_Store::load( 'admin-note' );
```
timmyc: Likewise this would be another good spot to drop a track recording in a follow-up PR.

tests/api/admin-notes.php
```diff
@@ -139,6 +139,23 @@ public function test_get_warning_notes() {
 		$this->assertEquals( $notes[0]['title'], 'PHPUNIT_TEST_NOTE_2_TITLE' );
 	}
 
+
+	/**
+	 * Test getting notes of a certain status.
+	 */
+	public function test_get_actioned_notes() {
```
timmyc: Thanks for the test 🌟 - should we add a test case for requesting an invalid action too?

timmyc (commented):
Super slick to see this all come together! Thanks for the detailed test instructions about removing the option, that tripped me up the first time I tried to test without reading the full directions 🤦‍♂️ 

Few comments in the code but overall looking quite dandy.

> One thing to note is that the currently, the alert does not get dismissed unless you refresh the page.

Yeah this is a pickle. Wondering if we should add some logic to `/wc/v3/system_status/tools/rebuild_stats` to see if the note exists, and if so, mark it as actioned. While I don't think this would fully fix the issue, it would cover the use case of someone visiting the settings page and triggering a population without going through the store alert flow.

I guess to make that fully work we would then have to trigger a refresh of the notes data after the `PUT` finishes, which does feel kind of hackish. We could set a shorter freshness level on the alerts query itself. Maybe change it to look every minute or something? That would work, but it would also introduce a bit of a weird UX also... imagine being on a page, the alert being up there, and all of a sudden it just disappears. /cc @LevinMedia for design thoughts on that.

Either way thinking we might need to explore some of those ideas a bit, and do a follow-up PR.

justinshreve:
> Yeah this is a pickle. Wondering if we should add some logic to /wc/v3/system_status/tools/rebuild_stats to see if the note exists, and if so, mark it as actioned. While I don't think this would fully fix the issue, it would cover the use case of someone visiting the settings page and triggering a population without going through the store alert flow.

This is already happening here: https://github.com/woocommerce/wc-admin/pull/1763/files#diff-97f534bbf775546e025dfb61c2cb208eR107 (that function checks if the note exists and marks it)

client/analytics/settings/config.js
justinshreve: Oops -- this was actually left in from a previous commit where I was attempting to refresh them on save -- but it felt a bit too hacky. This should be cleaned up.

justinshreve: Removed in https://github.com/woocommerce/wc-admin/pull/1763/commits/279818019b54f63a5302eba46bb348cbd5f24ce6.

includes/class-wc-admin-notes-historical-data.php
justinshreve: Added a todo in https://github.com/woocommerce/wc-admin/pull/1763/commits/279818019b54f63a5302eba46bb348cbd5f24ce6.

tests/api/admin-notes.php
justinshreve: Done in https://github.com/woocommerce/wc-admin/pull/1763/commits/279818019b54f63a5302eba46bb348cbd5f24ce6.

includes/class-wc-admin-notes-historical-data.php
justinshreve: Yeah good call - this was from an example note, but we really only need to check if it exists. I cleaned this up along with `content_data`.

justinshreve:
Thanks for the review @timmyc. I've addressed your comments in https://github.com/woocommerce/wc-admin/pull/1763/commits/279818019b54f63a5302eba46bb348cbd5f24ce6.

timmyc:
Updates look good, but for some reason now I'm not able to see the notice anymore - even after de-activating and removing the `wc_admin_version` option.

justinshreve:
@timmyc Is the `status` of the note still `actioned` in the database? You might need to update it via PHPMyAdmin to get it to show back up.

timmyc:
I had already deleted the old note record via phpmyadmin:

image

includes/class-wc-admin-notes-historical-data.php
```diff
@@ -0,0 +1,64 @@
...
+
+	/**
+	 * Creates a note for regenerating historical data.
+	 */
+	public static function add_note() {
+		$data_store = WC_Data_Store::load( 'admin-note' );
+
+		// First, see if we've already created this kind of note so we don't do it again.
+		$note_ids = $data_store->get_notes_with_name( self::NOTE_NAME );
+		if ( empty( $note_ids ) ) {
```
timmyc: I think this needs to be `if ( ! empty( $note_ids ) )`

timmyc (commented):
Okay dug in a bit more, I think a couple of things were causing it not to work for me locally, one was the item noted in the code, the other is by not supplying `content_data` when creating the note, the note's content data field in the db is `[]` vs `{}` for other notes. If I manually update the `content_data` field to `{}` all works well. So guess we need to pass in an empty array to `content_data` and maybe add some logic to WC_Admin_Note to sanitize that input a bit and set a proper default if not supplied - but we can do that in a follow-up pr

LevinMedia:
>One thing to note is that the currently, the alert does not get dismissed unless you refresh the page. tried to investigate a couple ways to fix this behavior. 1) Immediately dismissing the alert when "get started" is clicked. The user didn't actually take action yet so this feels incorrect. 

Taking a step back - I'm not sure that's incorrect, I think it's okay to dismiss this notification after a user takes that action. cc: @jameskoster for his thoughts there too.

justinshreve:
Doh @timmyc - thanks for catching that `empty` issue and the `content_data` issue. I hadn't tested with a totally clear note so I missed those -- but those issues should be fixed now. I also added a `@todo` for improving the default in a follow-up PR.

Per David's feedback, I've also pushed some changes here that allow us to update the status of the note when clicked. This now dismisess the note when clicked

I added this in a way that it can be used for different statuses/buttons.

I also added a PUT method for updating the status of a note, and related mutators. These will be needed for https://github.com/woocommerce/wc-admin/issues/1756 as well.

I also rebased and made some changes so this implementation plays nice with the preloading animation.

This needs another review since I've made larger changes in https://github.com/woocommerce/wc-admin/pull/1763/commits/035867cb8e5aa43b5e0f615d55a089a53525c86c, https://github.com/woocommerce/wc-admin/pull/1763/commits/152759539864bf50e341da573cdecabddb60ef85, https://github.com/woocommerce/wc-admin/pull/1763/commits/308e6dc776fc53c764c2fa7a71d4211907436ce0, and https://github.com/woocommerce/wc-admin/pull/1763/commits/5f6c1f8f13b9a7be2cddf86c32d4a44051061e9f.

includes/api/class-wc-admin-rest-admin-notes-controller.php
```diff
@@ -167,6 +178,49 @@ public function get_items_permissions_check( $request ) {
 		return true;
 	}
 
+	/**
+	 * Update a single note.
```
timmyc: Very nice.

includes/class-wc-admin-install.php
```diff
@@ -16,7 +16,7 @@ class WC_Admin_Install {
 	 *
 	 * @todo get this dynamically?
 	 */
-	const VERSION_NUMBER = '0.6.0';
+	const VERSION_NUMBER = '0.8.0';
```
timmyc: Nice catch, I was noticing this while testing out the PR too.

tests/api/admin-notes.php
```diff
@@ -106,6 +106,45 @@ public function test_get_note_without_permission() {
 		$this->assertEquals( 401, $response->get_status() );
 	}
 
+	/**
+	 * Test updating a single note.
+	 */
+	public function test_update_note() {
```
timmyc: Thanks for the tests, and the assertion around invalid permissions too.

timmyc (approved):
All the changes look great to me, and this is testing out well now. Nice work getting the notice to ✨ hide after clicking through.

PR #1764, opened 2019-03-07
Update plugin header to better match the readme

emdashcodes:
This tiny PR updates the plugin header in `wc-admin.php` to better match the readme.txt description, and show WooCommerce as the plugin byline like other WooCommerce plugins.

### Screenshots

screen shot 2019-03-07 at 1 47 59 am

* Visit the plugin page and see the updated description / author.

### DIFF
--- a/wc-admin.php
+++ b/wc-admin.php
@@ -2,8 +2,8 @@
 /**
  * Plugin Name: WooCommerce Admin
  * Plugin URI: https://github.com/woocommerce/wc-admin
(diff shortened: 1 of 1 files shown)


psealock (approved):
🎉 Nice. I suppose it won't always be "new", so maybe omit that? No real objection though

justinshreve:
I'll leave it for now so it matches the README. When we update the plugin and description to include new features (like the activity panel) we can ditch it.

PR #1765, opened 2019-03-08
Make variations and coupons endpoints return zero-value items

### LINKED ISSUE #1398: Coupons that have never been used appear in the autocompleter but not in the report
**To Reproduce**
* Create a coupon, for example: `coupon_with_0_uses`.
* Go to the _Coupon_ report.
* Select _Single Coupon_ as the filter, type `c` and select `coupon_width_0_uses`.
image
* Notice no coupon is displayed in the report, which is confusing:
image

**Expected behavior**
There would be two options here:
* Hide non-used coupons from the autocompleter.
* Make the report endpoint return data for coupons that have not been used.

Aljullu:
Closes #1398, #1668 and #1704.

Follow-up of #1755.

Makes `variations` and `coupons` endpoints return zero-value items.

### Screenshots
_(Notice _Black, US 8.5_ appears in the table even though its values are 0)_
image

- Go to the _Coupons_ report.
- Start a _Comparison_ making sure one of the coupons has all values to 0 (it has not been used during the selected date range).
- Verify the table shows all compared coupons, even the one that has all values to 0.


psealock (commented):
This is testing well for me in Coupons and Variations. I needed to do some url hacking to get the variations tested because search is broken. Issue here, https://github.com/woocommerce/wc-admin/issues/1773.

I see the same pattern previously adopted, but it would be great if @joshuatf could take a look.

psealock (approved):
Josh is AFK and this follows the same pattern as the other and includes unit tests, so we should merge it in. Nice work @Aljullu

PR #1766, opened 2019-03-08
Taxes report: sort tax rates numerically instead of alphabetically

### LINKED ISSUE #1721: Sorting taxes by rate sorts them alphabetically instead of numerically
**To Reproduce**
1. Go to the _Taxes_ report.
2. Sort taxes by _Rate_.
3. If you have two taxes with a different order of magnitude (10% and 2.50%, for example), you will notice they are sorted alphabetically instead of numerically.

**Expected behavior**
10% should be considered bigger than 2.50%.

**Screenshots**
image

Aljullu:
Fixes #1721.

Sort tax rates as numeric values instead of alphabetically.

### Screenshots
_Before (notice `10%` is not correctly sorted):_
image

_After:_
image

* Create two taxes with different order of magnitude (10% and 2.50%, for example) if you don't have them.
* Go to the _Taxes_ report.
* Sort taxes by _Rate_.
* Verify they are sorted numerically, so 10% is considered bigger than 2.50%.

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-taxes-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-taxes-data-store.php
@@ -296,7 +296,7 @@ class WC_Admin_Reports_Taxes_Data_Store extends WC_Admin_Reports_Data_Store impl
 		global $wpdb;
 
 		if ( 'rate' === $order_by ) {
-			return $wpdb->prefix . 'woocommerce_tax_rates.tax_rate';
+			return "CAST({$wpdb->prefix}woocommerce_tax_rates.tax_rate as DECIMAL(18,4))";
 		}
 
 		return $order_by;
--- a/includes/data-stores/class-wc-admin-reports-taxes-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-taxes-data-store.php
@@ -296,7 +296,7 @@ class WC_Admin_Reports_Taxes_Data_Store extends WC_Admin_Reports_Data_Store impl
 		global $wpdb;
 
 		if ( 'rate' === $order_by ) {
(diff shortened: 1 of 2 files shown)


includes/data-stores/class-wc-admin-reports-taxes-data-store.php
```diff
@@ -296,7 +296,7 @@ protected function normalize_order_by( $order_by ) {
 		global $wpdb;
 
 		if ( 'rate' === $order_by ) {
-			return $wpdb->prefix . 'woocommerce_tax_rates.tax_rate';
+			return "CAST({$wpdb->prefix}woocommerce_tax_rates.tax_rate as DECIMAL(18,4))";
```
psealock: I imagine `100.0000` is the upper limit on taxes, should this be `DECIMAL(7,4))`? Or `DECIMAL(5,2))` for two decimal points.

psealock (commented):
Just one comment for now, my testing got interrupted and I'll do some more tonight.

includes/data-stores/class-wc-admin-reports-taxes-data-store.php
Aljullu: Makes sense! I investigated it a bit and the upper limit seems to be `999.9999`. That's because values are stored in a `varchar(8)` and the UI doesn't impose any other limit than that, so percentages over 100% are allowed.

Fixed in 465b9b1fc0efae8547c52050057e483098dc16c0.

> Or DECIMAL(5,2)) for two decimal points.

I don't think there is any big performance difference between both, is it? So probably `DECIMAL(7,4)` is a better option, so sorting is more precise.

psealock (approved):
LGTM! Thanks for the tests too!

PR #1770, opened 2019-03-09
Add a custom “cancel all” method to our scheduled action data store.

### LINKED ISSUE #1699: Create utility to clear all pending Analytics Background Jobs

jeffstieler:
Performs direct database manipulation for performance (with large datasets).

Fixes #1699.

This PR introduces direct SQL queries to clear all pending WC Admin actions and the associated metadata.

In my testing, the new `clear_pending_wcadmin_actions()` method cleared `46430` actions in `1.392324924469s` (tested using `microtime()`).

- Disable default AS queue runner (https://github.com/Prospress/action-scheduler-disable-default-runner/)
- Queue a lot of WC Admin jobs (can call `WC_Admin_Reports_Sync::queue()->schedule_single()` many times in a loop, etc)
- Use WP CLI to kick off AS runner (`wp action-scheduler run`)
- Deactivate the WC Admin plugin
- Verify all pending WC Admin jobs are gone

**NOTE:** If you encounter a PHP Fatal error in the CLI session, that is fixed with: https://github.com/Prospress/action-scheduler/pull/258

### DIFF
--- a/includes/class-wc-admin-actionscheduler-wppoststore.php
+++ b/includes/class-wc-admin-actionscheduler-wppoststore.php
@@ -37,31 +37,34 @@ class WC_Admin_ActionScheduler_WPPostStore extends ActionScheduler_wpPostStore {
 	/**
 	 * Forcefully delete all pending WC Admin scheduled actions.
 	 *
-	 * Directly deletes items from the database for performance.
+	 * Directly trashes items from in database for performance.
 	 */
 	public function clear_pending_wcadmin_actions() {
 		global $wpdb;
 
-		// Remove all scheduled action posts and their metadata.
-		$delete_pending_sql =
-			"DELETE p.*, pm.* FROM {$wpdb->posts} p
-			JOIN {$wpdb->postmeta} pm ON p.ID = pm.post_id
-			WHERE post_type = 'scheduled-action'
-			AND post_status = 'pending'
-			AND post_title LIKE 'wc-admin_%'";
+		// Cancel all pending actions by trashing the posts.
+		// Action Scheduler will handle the cleanup.
+		$action_types = array(
+			WC_Admin_Reports_Sync::QUEUE_BATCH_ACTION,
+			WC_Admin_Reports_Sync::QUEUE_DEPEDENT_ACTION,
+			WC_Admin_Reports_Sync::CUSTOMERS_BATCH_ACTION,
+			WC_Admin_Reports_Sync::ORDERS_BATCH_ACTION,
+			WC_Admin_Reports_Sync::ORDERS_LOOKUP_BATCH_INIT,
+			WC_Admin_Reports_Sync::SINGLE_ORDER_ACTION,
+		);
 
-		// phpcs:ignore WordPress.DB.PreparedSQL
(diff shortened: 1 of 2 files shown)


timmyc (approved):
This is working as described on the tin, I like the check to ensure the store we are operating against is `WC_Admin_ActionScheduler_WPPostStore`

psealock (commented):
>Disable default AS queue runner (https://github.com/Prospress/action-scheduler-disable-default-runner/)

If I don't have AS plugin, I won't need to do this step right?

In other words, this comment is for that scenario. Just seeing if my understanding is correct.

>// If we're using our data store, call our bespoke deletion method.

jeffstieler:
> If I don't have AS plugin, I won't need to do this step right?

AS is included with WC core, you don't need to have it installed as a separate plugin.

You shouldn't have to disable the default runner, and it's actually a good test to keep AS active while trying this functionality out - I'm about to do the same myself and will report back.

jeffstieler:
I've just tested by creating 20k+ jobs, then firing off the CLI runner (`wp action-scheduler run`), waiting for it to start and deactivating the plugin. The in-progress batch ended up encountering a fatal error at this line: https://github.com/Prospress/action-scheduler/blob/master/classes/ActionScheduler_Abstract_QueueRunner.php#L66

I don't think that's a problem _(cc: @rrennick)_ but in any case I'm opening a PR on the AS repo to fix that.

jeffstieler:
Here's the PR: https://github.com/Prospress/action-scheduler/pull/258

rrennick:
> I don't think that's a problem

Someone deactivating/deleting a plugin while a WP CLI command relying on the plugin is an edge case so I don't see that as a problem.

Is there a reason for deleting them instead of just cancelling them? Cancelling would only involve the `wp_posts` table and they will be automatically cleaned up by AS later.

rrennick:
> Are you suggesting issuing an UPDATE query

Yes, that's what I was thinking.

jeffstieler:
Switched to using `wpdb::update()` in ffb24fa. This is ready for another review/test @timmyc @rrennick.

rrennick:
@jeffstieler 🚀  that looks good to me. I won't have a chance to test it until tomorrow morning. if @timmyc gives it a test run without issue then it's fine to merge by me.

timmyc (approved):
Still testing out well for me - even ran without pausing this time, and as expected a job failed gracefully, and all pending were removed. CI is reporting some issues, but maybe just needs a rebase? Pre-approving

jeffstieler:
Recreating to hopefully fix CI issues.

jeffstieler:
@timmyc please approve #1813 😄

PR #1779, opened 2019-03-12
Fix/products stats segment product filter cats

psealock:
Fixes https://github.com/woocommerce/wc-admin/issues/1772

When segmenting `/reports/products/stats` by `product` segments were not filtered properly by caetgory, if supplied as an argument.

1. Using postman, segment by product and filter by a known category id (or ids): `/wp-json/wc/v4/reports/products/stats?categories=<my-category-id>&segmentby=product`
2. See the segments in the response limited to products in the category supplied.

### Question

Performance wise, would it be better to write a query with a join than to continue using `wc_get_products`, which requires the `category` [argument to be an array of slugs](https://github.com/woocommerce/woocommerce/wiki/wc_get_products-and-WC_Product_Query), which means a `get_term_by` call for each category?

### DIFF
--- a/includes/class-wc-admin-reports-segmenting.php
+++ b/includes/class-wc-admin-reports-segmenting.php
@@ -311,9 +311,17 @@ class WC_Admin_Reports_Segmenting {
 				'limit'  => -1,
 			);
 
-			// @todo: filter by categories if $this->query_args['categories'] is set.
 			if ( isset( $this->query_args['product_includes'] ) ) {
 				$args['include'] = $this->query_args['product_includes'];
+			}
+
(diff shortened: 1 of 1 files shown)


joshuatf (approved):
Code looks good and is testing well!

>Performance wise, would it be better to write a query with a join than to continue using wc_get_products

I do think a more performant option would be nice.  I wonder if we should switch to passing slugs into the API?  We'll have access to slugs on the front-end, but it would be nice to keep things consistent across the API.

I'm okay discussing this in another PR if you'd like since we might also want to adjust this in other areas where we currently use product IDs (related if we don't remove this - https://github.com/woocommerce/wc-admin/issues/1789).

PR #1780, opened 2019-03-12
Avoid useless Chart re-renders and clean-up component state

### LINKED ISSUE #1769: Store performance: bad performance when toggling numbers visibility on mobile
Toggling store performance numbers visibility is very slow on mobile.

**To Reproduce**
1. Go to the _Dashboard_.
2. Open the ellipsis menu next to _Store Performance_.
3. Try switching items' visibility.
4. Notice it's very slow, it takes more than a second for the toggle control to change its state.

**Screenshots**
press

**Smartphone:**
* Device: LG G6
* OS: Android 8.0
* Browser: Firefox 67.0a1

Aljullu:
Fixes #1769.

This PR:
- Avoids re-rendering `<ReportChart>` components when interacting with other _Dashboard_ elements.
- Keeps `<Chart>` component state as simple as possible. We were storing complex data objects in the state to avoid re-computing them, but the correct way is not to store that in the component state but to use [memoization](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization).
- I also found out we were overwriting a prop (`yFormat`) in `<Chart>`, that should be fixed in this PR as well.

### Accessibility

### Screenshots
_(Adding a `console.log()` in `<Chart>` [render method](https://github.com/woocommerce/wc-admin/blob/master/packages/components/src/chart/index.js#L272))_
_Before:_
toggle-before
:point_up: That's only with one chart visible in the _Dashboard_, when all available charts were displayed, there were dozens of re-render methods called on every toggle switch.

_After:_
toggle-after

**Verify #1769 is fixed:**
- Go to the _Dashboard_.
- Open the ellipsis menu next to _Store Performance_.
- Try switching items' visibility.
- Verify performance is ok.

**Verify there are no regressions:**
- Go to any report and interact a bit with the chart: hover legend items, select/unselect them, apply a comparison, etc.

### DIFF
--- a/client/analytics/components/report-chart/index.js
+++ b/client/analytics/components/report-chart/index.js
@@ -6,7 +6,7 @@ import { __ } from '@wordpress/i18n';
 import { Component } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import { format as formatDate } from '@wordpress/date';
-import { get } from 'lodash';
+import { get, isEqual } from 'lodash';
 import PropTypes from 'prop-types';
 
 /**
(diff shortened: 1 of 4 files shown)


psealock (approved):
This is working well for me. I am not seeing regressions in other areas such as legend, type toggle, and general use. Nice work!

Issue #1776, opened 2019-03-11
RTL issues


Aljullu:
### Date picker
* [x] Custom date picker is broken (#1810).
![image](https://user-images.githubusercontent.com/3616980/54112813-4bd0eb00-43e7-11e9-9c58-2aae5ac8c673.png)
* [x] Date picker is misaligned (#1803).
![image](https://user-images.githubusercontent.com/3616980/54285225-db21fe00-45a1-11e9-8ccf-5a4feccc8a75.png)
* [x] Dot of active option is rendered on top of text (#1781).
![image](https://user-images.githubusercontent.com/3616980/54112842-61461500-43e7-11e9-84c8-116da85548f2.png)

### Ellipsis menu
* [x] Multi-line elements are left-aligned, but they should be aligned to the right (#1781).
![image](https://user-images.githubusercontent.com/3616980/54112920-8175d400-43e7-11e9-891d-d8060994694f.png)
* [x] Ellipsis menu title should be aligned to the right (#1781).
![image](https://user-images.githubusercontent.com/3616980/54112671-fbf22400-43e6-11e9-9124-5a1fec946ff9.png)

### Chart
* [x] Missing spacing between legend label and value (#1781).
![image](https://user-images.githubusercontent.com/3616980/54112476-88e8ad80-43e6-11e9-8412-cddc9b85bf71.png)
* [x] Missing spacing between elements in chart tooltip (#1781).
![image](https://user-images.githubusercontent.com/3616980/54112579-c816fe80-43e6-11e9-90e9-9c222c13d34c.png)
* [x] Y-axis labels should be on the right side of the chart (#1794).
![image](https://user-images.githubusercontent.com/3616980/54112614-e11faf80-43e6-11e9-8156-67bb4b7e5e07.png)
* [x] In left column of _Dashboard_ charts, tooltip is rendered outside of the screen space (#1794).
![image](https://user-images.githubusercontent.com/3616980/54113026-acf8be80-43e7-11e9-8481-d796015b9a29.png)
* [x] Loading spinner circles are misaligned (https://github.com/WordPress/gutenberg/pull/14418).
![circles](https://user-images.githubusercontent.com/3616980/54285467-4cfa4780-45a2-11e9-8cc8-70157a4358d1.gif)

### Table
* [x] Missing margin between _Download_ button icon and text (#1781).
[...]

PR #1781, opened 2019-03-12
 Load RTL CSS files 

Aljullu:
Fixes most issues (but not all) from #1776.

This PR:
- Installs `webpack-rtl-plugin` so we generate an automated RTL CSS file. It uses a fork of the plugin, as it's done in Gutenberg (https://github.com/WordPress/gutenberg/pull/3844).
- Loads RTL CSS files when using an RTL locale.
- Removes legacy RTL styles from our current CSS.

### Screenshots
image

- Install an RTL language in your WordPress, for example, Arabic.
- Go to _Settings_ and change the default language to that one.
- Press <kbd>Ctrl</kbd>+<kbd>U</kbd> to view the page source and verify wc-admin CSS files contain `rtl` in their name.
- If you want, you can take a look at #1776 and verify most issues are gone! :fire:

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,8 @@
     "stylelint": "9.9.0",
     "stylelint-config-wordpress": "13.1.0",
     "webpack": "4.28.3",
-    "webpack-cli": "3.1.2"
+    "webpack-cli": "3.1.2",
+    "webpack-rtl-plugin": "github:yoavf/webpack-rtl-plugin#develop"
   },
   "dependencies": {
     "@fresh-data/framework": "^0.5.1",
--- a/lib/client-assets.php
+++ b/lib/client-assets.php
@@ -110,6 +110,7 @@ function wc_admin_register_script() {
 		array( 'wp-edit-blocks' ),
 		filemtime( wc_admin_dir_path( 'dist/components/style.css' ) )
 	);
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -106,7 +106,8 @@
     "stylelint": "9.9.0",
     "stylelint-config-wordpress": "13.1.0",
     "webpack": "4.28.3",
-    "webpack-cli": "3.1.2"
+    "webpack-cli": "3.1.2",
+    "webpack-rtl-plugin": "github:yoavf/webpack-rtl-plugin#develop"
```
psealock: Should we target `master` ?

psealock (approved):
Nice work with this one! I just checked against issues in #1776 marked with this PR (1781) and its looking great. Thanks for handling these cases.

Just one comment about targeting the `develop` branch.

package.json
Aljullu: I chose that one to keep consistency with Gutenberg:
https://github.com/WordPress/gutenberg/blob/master/package.json#L123

Ideally, we would be using `webpack-rtl-plugin` from NPM, but we need this PR to be merged upstream before: https://github.com/romainberger/webpack-rtl-plugin/pull/15.

PR #1782, opened 2019-03-12
Update text domain.

### LINKED ISSUE #1735: Update Text Domain
Background: p90Yrv-Ze-p2

Per the feedback from the .org review, we either need to change our slug name to `wc-admin`, or update the textdomain to `woocommerce-admin`.

This task would touch many files 😱 - and would cause conflicts in some open prs too. Any thoughts?

jeffstieler:
Fixes #1735.

Replaces all instances of the `wc-admin` text domain with `woocommerce-admin`. Renames generated `/languages` files to `woocommerce-admin.pot|php` as well.

- Check diff to make sure substitutions were made correctly
- Spot check all WC Admin functionality
- Verify PHP tests pass
- Test plugin build (release, zip)


bin/build-plugin-zip.sh
```diff
@@ -78,8 +78,8 @@ zip -r wc-admin.zip \
 	includes/**/*.php \
 	images/* \
 	$build_files \
-	languages/wc-admin.pot \
-	languages/wc-admin.php \
+	languages/woocommerce-admin.pot \
```
joshuatf: Not sure if necessary or not, but should the build file also be renamed to `woocommerce-admin.zip`?

joshuatf (commented):
Nice work on this behemoth of a PR, @jeffstieler!

One comment about the build file and a quick change is probably needed after https://github.com/woocommerce/wc-admin/pull/1750 that I think you've already noted.

bin/build-plugin-zip.sh
jeffstieler: Yeah I think that makes sense to change as well. 👍

PR #1791, opened 2019-03-13
Update product searching to search by name or sku

### LINKED ISSUE #1739: Product search does not search by product name or sku
**Describe the bug**
The product search table states that it searches by product name or SKU, but it actually uses the default search parameter which uses the post title (product name) or post content.

**To Reproduce**

1. Go to the products report.
2. Search by sku.
3. Note the item does not show.
4. Search by a product's content (not included in sku or name).
5. Note that other items are returned if they contain the keywords in their content.

**Expected behavior**
Expected results only based on sku and name.

**Screenshots**
screen shot 2019-03-05 at 6 32 25 pm

**Additional context**
https://github.com/woocommerce/wc-admin/pull/1670 introduced product name filtering.

joshuatf:
Fixes #1739 

* Adds a `product_search` option to wp_query to allow searching by product name or sku.
* Swaps `search` for `product_search` in wc-api to retrieve searched products.
* Searches by product name or sku in product autocompleters.

### Screenshots
Screen Shot 2019-03-13 at 3 15 14 PM

1.  Create a product with a variation in product name and sku.
2.  Also create a product that has the content of your keywords in the post content, but not the title or sku.
3.  Search the autocompleters and make sure options returned are being searched by name or sku.
4.  Type a query in the top of the products table search and make sure the search also works by name or sku.
5.  Make sure the product created from step 2 does not appear in the above searches.

### Question
Instead of keeping both `product_name` and `product_search`, I opted to drop the former and make all autocompleters for products allow sku searching as well.  Do we want to revert this back?  If not, should we mention the allowed sku searching in the autocompleter placeholder text (`Search for products to compare by product name or sku`)?

### DIFF
--- a/client/wc-api/items/utils.js
+++ b/client/wc-api/items/utils.js
@@ -20,13 +20,9 @@ export function searchItemsByString( select, endpoint, search ) {
 	let isError = false;
 	search.forEach( searchWord => {
 		const query = {
+			search: searchWord,
 			per_page: 10,
 		};
-		if ( 'products' === endpoint ) {
-			query.product_search = searchWord;
-		} else {
-			query.search = searchWord;
-		}
 		const newItems = getItems( endpoint, query );
 		newItems.forEach( ( item, id ) => {
 			items[ id ] = item;
--- a/client/wc-api/items/utils.js
+++ b/client/wc-api/items/utils.js
@@ -20,9 +20,13 @@ export function searchItemsByString( select, endpoint, search ) {
 	let isError = false;
 	search.forEach( searchWord => {
 		const query = {
-			search: searchWord,
 			per_page: 10,
 		};
+		if ( 'products' === endpoint ) {
+			query.product_search = searchWord;
+		} else {
+			query.search = searchWord;
+		}
 		const newItems = getItems( endpoint, query );
 		newItems.forEach( ( item, id ) => {
 			items[ id ] = item;
--- a/includes/api/class-wc-admin-rest-products-controller.php
+++ b/includes/api/class-wc-admin-rest-products-controller.php
@@ -61,11 +61,16 @@ class WC_Admin_REST_Products_Controller extends WC_REST_Products_Controller {
 			'type'              => 'string',
 			'validate_callback' => 'rest_validate_request_arg',
 		);
+		$params['product_search'] = array(
+			'description'       => __( 'Search by similar product name or sku.', 'woocommerce-admin' ),
+			'type'              => 'string',
+			'validate_callback' => 'rest_validate_request_arg',
+		);
 		return $params;
 	}
 
 	/**
-	 * Add product name filtering to the WC API.
+	 * Add product name and sku filtering to the WC API.
 	 *
 	 * @param WP_REST_Request $request Request data.
 	 * @return array
@@ -75,6 +80,10 @@ class WC_Admin_REST_Products_Controller extends WC_REST_Products_Controller {
 
 		if ( ! empty( $request['product_name'] ) ) {
 			$args['post_title__like'] = $request['product_name'];
+		}
+
+		if ( ! empty( $request['product_search'] ) ) {
+			$args['product_search'] = $request['product_search'];
 		}
 
 		return $args;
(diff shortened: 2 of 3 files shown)


joshuatf:
Note that tests are failing as this PR updates the params text domain to `woocommerce-admin` in prep of https://github.com/woocommerce/wc-admin/pull/1782 being merged.

includes/api/class-wc-admin-rest-products-controller.php
```diff
@@ -87,30 +87,51 @@ protected function prepare_objects_query( $request ) {
...
-		if ( $post_title_search ) {
-			$post_title_search = $wpdb->esc_like( $post_title_search );
-			$post_title_search = ' \'%' . $post_title_search . '%\'';
-			$where            .= ' AND ' . $wpdb->posts . '.post_title LIKE ' . $post_title_search;
+		$product_search = $wp_query->get( 'product_search' );
+		if ( $product_search ) {
+			$product_search = $wpdb->esc_like( $product_search );
+			$product_search = ' \'%' . $product_search . '%\'';
+			$where         .= ' AND (' . $wpdb->posts . '.post_title LIKE ' . $product_search . ' OR ';
+			$where         .= 'ps_post_meta.meta_key = "_sku" AND ps_post_meta.meta_value LIKE ' . $product_search . ')';
```
timmyc: Any performance concerns here with the meta query?

includes/api/class-wc-admin-rest-products-controller.php
```diff
@@ -87,30 +87,51 @@ protected function prepare_objects_query( $request ) {
...
+	public static function add_wp_query_product_search_filter( $where, $wp_query ) {
 		global $wpdb;
 
-		$post_title_search = $wp_query->get( 'post_title__like' );
-		if ( $post_title_search ) {
-			$post_title_search = $wpdb->esc_like( $post_title_search );
-			$post_title_search = ' \'%' . $post_title_search . '%\'';
-			$where            .= ' AND ' . $wpdb->posts . '.post_title LIKE ' . $post_title_search;
+		$product_search = $wp_query->get( 'product_search' );
+		if ( $product_search ) {
```
timmyc: Should we check the length here to make sure it isn't empty or full of whitespace, same for `add_wp_query_product_search_join`

timmyc (commented):
>  I opted to drop the former and make all autocompleters for products allow sku searching as well

Seems legit to me, and reading the code it makes sense to. I suppose my only concern is ( and making a wild guess here ) that not all sites leverage `sku`, so we could be joining when not needed. Couple of comments in the code but looking great

client/wc-api/items/utils.js
```diff
@@ -20,9 +20,13 @@ export function searchItemsByString( select, endpoint, search ) {
 	let isError = false;
 	search.forEach( searchWord => {
 		const query = {
-			search: searchWord,
 			per_page: 10,
 		};
+		if ( 'products' === endpoint ) {
+			query.product_search = searchWord;
```
jeffstieler: Do we have to rename `search` to `product_search` here? Since we're already overriding `get_collection_params()`, can't we handle the `search` parameter there?

includes/api/class-wc-admin-rest-products-controller.php
jeffstieler: It's `trim()`ed, so this check is probably adequate.

includes/api/class-wc-admin-rest-products-controller.php
joshuatf: It's certainly not as quick as just searching the posts table, but also not terrible performance and uses the same basic meta query SQL syntax that core does.

if we want to offer SKU searching, we don't have a lot of options here.

client/wc-api/items/utils.js
joshuatf: @jeffstieler Will do!

joshuatf:
>I suppose my only concern is ( and making a wild guess here ) that not all sites leverage sku, so we could be joining when not needed.

@timmyc That's a valid concern.  I added in a check to see if they've been disabled site-wide so at least sites explicitly not using them aren't impacted.

joshuatf:
@jeffstieler Updated to use the `search` param.  This one is ready for another look if you have time.

includes/api/class-wc-admin-rest-products-controller.php
```diff
@@ -55,26 +55,28 @@ public function get_item_schema() {
...
 	 * @return array
 	 */
 	protected function prepare_objects_query( $request ) {
 		$args = parent::prepare_objects_query( $request );
 
-		if ( ! empty( $request['product_name'] ) ) {
-			$args['post_title__like'] = $request['product_name'];
+		if ( ! empty( $request['search'] ) ) {
+			$args['search'] = $request['search'];
+			$args['s']      = false;
```
jeffstieler: Nitpick: use `unset()` instead.

jeffstieler (approved):
Looks good! :ship: it.

PR #1792, opened 2019-03-13
Update/variations search method

### LINKED ISSUE #1773: Variations Search: Results not displayed
**Describe the bug**

When looking at Comparison of Variations in the Product Report, the search box doesn't show results.

**To Reproduce**

1. Show > Single Product > Choose variable product.
2. Enter Comparison mode and search for a known name of a variation, or simply the product name.
3. See no results appear in the dropdown. Open network requests and confirm that `/products/<id>/variations?search=...` returns results.

**Expected behavior**

I'd expect the search dropdown to be populated with those results.

**Screenshots**

Screen Shot 2019-03-11 at 4 28 41 PM

joshuatf:
Fixes #1773 

* Adds `product_search` to variations controller.
* Hooks up `product_search` to variation autocompleters.
* Adds product sku to autocompleter keywords.

### Screenshots
Screen Shot 2019-03-13 at 5 22 57 PM

1. Create a variable product with many products and varying skus.
2. Check that the autocompleter options correctly return products by searched name/sku.

@psealock Can you double check and let me know if this works with your data?

### DIFF
--- a/includes/api/class-wc-admin-rest-product-variations-controller.php
+++ b/includes/api/class-wc-admin-rest-product-variations-controller.php
@@ -63,9 +63,28 @@ class WC_Admin_REST_Product_Variations_Controller extends WC_REST_Product_Variat
 	public function get_items( $request ) {
 		add_filter( 'posts_where', array( 'WC_Admin_REST_Products_Controller', 'add_wp_query_product_search_filter' ), 10, 2 );
 		add_filter( 'posts_join', array( 'WC_Admin_REST_Products_Controller', 'add_wp_query_product_search_join' ), 10, 2 );
+		add_filter( 'posts_groupby', array( __CLASS__, 'add_wp_query_product_search_group_by' ), 10, 2 );
 		$response = parent::get_items( $request );
 		remove_filter( 'posts_where', array( 'WC_Admin_REST_Products_Controller', 'add_wp_query_product_search_filter' ), 10 );
 		remove_filter( 'posts_join', array( 'WC_Admin_REST_Products_Controller', 'add_wp_query_product_search_join' ), 10 );
+		remove_filter( 'posts_groupby', array( __CLASS__, 'add_wp_query_product_search_group_by' ), 10 );
 		return $response;
+	}
+
+	/**
+	 * Group by post ID to prevent duplicates.
+	 *
+	 * @param string $groupby Group by clause used to organize posts.
+	 * @param object $wp_query WP_Query object.
+	 * @return string
+	 */
+	public function add_wp_query_product_search_group_by( $groupby, $wp_query ) {
+		global $wpdb;
+
+		$product_search = $wp_query->get( 'product_search' );
+		if ( $product_search && empty( $groupby ) ) {
+			$groupby .= $wpdb->posts . '.ID';
+		}
+		return $groupby;
 	}
 }
--- a/includes/api/class-wc-admin-rest-product-variations-controller.php
+++ b/includes/api/class-wc-admin-rest-product-variations-controller.php
@@ -22,4 +22,50 @@ class WC_Admin_REST_Product_Variations_Controller extends WC_REST_Product_Variat
 	 * @var string
 	 */
 	protected $namespace = 'wc/v4';
+
+	/**
+	 * Get the query params for collections.
+	 *
+	 * @return array
+	 */
+	public function get_collection_params() {
+		$params                   = parent::get_collection_params();
+		$params['product_search'] = array(
+			'description'       => __( 'Search by similar product name or sku.', 'woocommerce-admin' ),
+			'type'              => 'string',
+			'validate_callback' => 'rest_validate_request_arg',
+		);
+		return $params;
+	}
+
+	/**
+	 * Add product name and sku filtering to the WC API.
+	 *
+	 * @param WP_REST_Request $request Request data.
+	 * @return array
+	 */
+	protected function prepare_objects_query( $request ) {
+		$args = parent::prepare_objects_query( $request );
+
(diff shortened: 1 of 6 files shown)


psealock (commented):
Indeed I am able to search variation names in the table as well as by SKU. Nice work.

1. We can change the input placeholder back to `Search by variation name or SKU`.
2. I ran into one unrelated issue when searching for a Single Product:

Screen Shot 2019-03-14 at 2 07 38 PM

It occurs on master as well, I'm just noting it here in case you are also experiencing the same thing.

3. One related issue is when selecting the search option, the url removes the `?filter=single_product` parameter. I ran into this problem when looking at categories and [solved it](https://github.com/woocommerce/woocommerce-admin/pull/1762) by introducing a table prop called `baseSearchQuery`

https://github.com/woocommerce/woocommerce-admin/blob/80a00415497f4f28340d0331608460a3ddd93fa6/client/analytics/report/categories/index.js#L84

joshuatf:
@psealock Thanks for the review!

1. Updated.
2. Did not run into this issue here or master.  Any filters in place that cause this?
3. I might be misunderstanding the issue, but if we're using a single product filter and subsequently use the search, I would expect the filter to be removed since you wouldn't want to search a single product.

includes/api/class-wc-admin-rest-product-variations-controller.php
```diff
@@ -22,4 +22,69 @@ class WC_Admin_REST_Product_Variations_Controller extends WC_REST_Product_Variat
...
 	protected $namespace = 'wc/v4';
+
+	/**
+	 * Get the query params for collections.
+	 *
+	 * @return array
+	 */
+	public function get_collection_params() {
+		$params                   = parent::get_collection_params();
+		$params['product_search'] = array(
```
jeffstieler: Same question as https://github.com/woocommerce/woocommerce-admin/pull/1791/files#r265597768:

Do we have to use `product_search` instead of `search` here? Since we're already overriding `get_collection_params()`, can't we handle the `search` parameter there?

joshuatf: @jeffstieler I was wondering the same.  My concern was that other plugins taking advantage of the `search` param might run into unexpected behavior, but maybe this is okay given that we've namespaced the API to v4.

jeffstieler: Yeah I think the namespacing saves us here - we're changing things much more drastically in the customers endpoint: https://github.com/woocommerce/woocommerce-admin/blob/master/includes/api/class-wc-admin-rest-customers-controller.php#L18

psealock:
>if we're using a single product filter and subsequently use the search, I would expect the filter to be removed since you wouldn't want to search a single product.

Thats true, except in this case we are searching for variations of a particular product (I removed the chart in all these so it fits nicely):

Notice the title of the table is "Variations" and we are looking at the single product Penguin.

Screen Shot 2019-03-15 at 9 26 35 AM

When I click a search selection, because the top level filter `filter=single_product` is removed from the url, we go back to looking all products, and no longer variations of that single product. I see what you are saying, but the context of the user's search was in regards to the single product, so I think the resulting view should reflect that.

Screen Shot 2019-03-15 at 9 29 09 AM

When I manually add back in the `filter=single_product` param, the filter is correct, but the table is not right. It should be the variations table and not the product.

Screen Shot 2019-03-15 at 9 32 31 AM

psealock:
>I ran into one unrelated issue when searching for a Single Product:

>>Did not run into this issue here or master. Any filters in place that cause this?

Weird, I will dig into why I'm seeing this.

joshuatf:
>Thats true, except in this case we are searching for variations of a particular product.

@psealock You're absolutely right; I was thinking about this the wrong way.  Added `baseSearchQuery` in https://github.com/woocommerce/woocommerce-admin/pull/1792/commits/78fe98772b781b51d9afcd5b5f4a084a431f00ff

@jeffstieler Rebased off `update/wp-query-product-search` and changed params to use `search` instead of `product_search`.

includes/api/class-wc-admin-rest-product-variations-controller.php
```diff
@@ -22,4 +22,70 @@ class WC_Admin_REST_Product_Variations_Controller extends WC_REST_Product_Variat
...
+	 *
+	 * @param WP_REST_Request $request Request data.
+	 * @return array
+	 */
+	protected function prepare_objects_query( $request ) {
+		$args = parent::prepare_objects_query( $request );
+
+		if ( ! empty( $request['search'] ) ) {
+			$args['search'] = $request['search'];
+			$args['s']      = false;
```
jeffstieler: Nitpick: I think it would be better to `unset()` here. _(`false` still passes an `isset()` check that other code might be using)_

includes/api/class-wc-admin-rest-product-variations-controller.php
```diff
@@ -22,4 +22,70 @@ class WC_Admin_REST_Product_Variations_Controller extends WC_REST_Product_Variat
...
+	 * @param string $groupby Group by clause used to organize posts.
+	 * @param object $wp_query WP_Query object.
+	 * @return string
+	 */
+	public function add_wp_query_product_search_group_by( $groupby, $wp_query ) {
+		global $wpdb;
+
+		$search = trim( $wp_query->get( 'search' ) );
+		if ( $search && empty( $groupby ) ) {
+			$groupby .= $wpdb->posts . '.ID';
```
jeffstieler: Nitpick: since we know `$groupby` is empty, there's no need for concatenation.

jeffstieler (approved):
Code looks good and tests well - optionally handle two nitpicks before merging.

PR #1797, opened 2019-03-14
Check if dependencies are met before running deactivation hook

### LINKED ISSUE #1752: Fatal Error on activation
Fatal error: Uncaught Error: Class 'WC_Admin_Reports_Sync' not found.

Error while activating the plugin

WC 3.5.5
WP 5.0
screenshot 2019-03-06 at 3 15 46 pm

joshuatf:
Fixes #1752 

Classes aren't available to remove scheduled actions if dependencies were never met.

1. Deactivate wc-admin.
2. Don't meet the dependencies (you can deactivate WooCommerce to do this or throw `return false;` into `dependencies_satisfied()`).
3. Activate wc-admin.
4. Note the plugin is deactivated automatically with the dependencies notice and no fatal errors are thrown.

### DIFF
--- a/woocommerce-admin.php
+++ b/woocommerce-admin.php
@@ -123,8 +123,10 @@ add_action( 'admin_init', 'wc_admin_possibly_deactivate_wc_admin_plugin' );
  * On deactivating the wc-admin plugin.
  */
 function wc_admin_deactivate_wc_admin_plugin() {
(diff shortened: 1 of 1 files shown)


jeffstieler (approved):
Nice catch. 🚢 it!

PR #1799, opened 2019-03-14
Show error in activity panel if no actionable order statuses exist

### LINKED ISSUE #1749: Empty orders activity panel when no actionable statuses
**Describe the bug**
When no actionable statuses are selected in analytics settings, the activity panel is empty.

**To Reproduce**

1. Go to analytics settings.
2. Uncheck all actionable statuses and save.
3. Refresh and open the Orders activity panel.
4. Note that nothing is displayed.

**Expected behavior**
Expected a message of some type: `You currently don't have any actionable statuses.  To display orders here that need action, select orders that require further review in <Link>settings</Link>.`

**Screenshots**
screen shot 2019-03-06 at 1 13 33 pm

joshuatf:
Fixes #1749 

Adds an error message and link to settings if no actionable statuses are set instead of a blank panel.

### Screenshots
Screen Shot 2019-03-14 at 2 29 53 PM
Screen Shot 2019-03-14 at 2 29 31 PM

1.  Go to Settings and uncheck all actionable statuses.
2.  Save and refresh.
3.  Note the error message and correct link in the Orders tab of the activity panel.
4.  Make sure orders continue to show correctly when actionable statuses are added.

### DIFF
--- a/client/header/activity-panel/panels/orders.js
+++ b/client/header/activity-panel/panels/orders.js
@@ -35,8 +35,21 @@ import ActivityOutboundLink from '../activity-outbound-link';
 import { QUERY_DEFAULTS } from 'wc-api/constants';
 import withSelect from 'wc-api/with-select';
 
-function OrdersPanel( { orders, isRequesting, isError } ) {
+function OrdersPanel( { orders, isRequesting, isError, orderStatuses } ) {
 	if ( isError ) {
+		if ( ! orderStatuses.length ) {
+			return (
+				<EmptyContent
+					title={
+						"You currently don't have any actionable statuses. " +
+						'To display orders here, select orders that require further review in settings.'
+					}
+					actionLabel={ __( 'Settings', 'woocommerce-admin' ) }
+					actionURL={ getAdminLink( 'admin.php?page=wc-admin#/analytics/settings' ) }
+				/>
+			);
+		}
+
 		const title = __(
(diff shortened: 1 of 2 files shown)


client/header/activity-panel/panels/orders.js
```diff
@@ -35,8 +35,21 @@ import ActivityOutboundLink from '../activity-outbound-link';
...
 import withSelect from 'wc-api/with-select';
 
-function OrdersPanel( { orders, isRequesting, isError } ) {
+function OrdersPanel( { orders, isRequesting, isError, orderStatuses } ) {
 	if ( isError ) {
+		if ( ! orderStatuses.length ) {
+			return (
+				<EmptyContent
+					title={
+						"You currently don't have any actionable statuses. " +
```
Aljullu: Should this be wrapped in `__()`?

Aljullu (approved):
Code looks good and this is testing well. Just a couple of comments:

What do you think about adding some margin around the text?
image

client/header/activity-panel/panels/orders.js
joshuatf: Definitely- nice catch! ⚾️

joshuatf:
>What do you think about adding some margin around the text?

Seems like a good idea.  Added some padding around the empty content div so none of these elements can be flush against the panel side.

Thanks for the review @Aljullu!

PR #1800, opened 2019-03-14
Update coupon search to allow searching by code

### LINKED ISSUE #1611: Question: Coupons Endpoint - should search include description?
**Describe the bug**

Found as part of #1525.

The coupons report endpoint `search` parameter searches the coupon description - should it?

This can lead to somewhat confusing results since the description isn't shown in the report table.

**To Reproduce**

1. Go to Coupons Report
2. Search for a term that's in a coupon description, but not it's code (like "woo" on WCCOM data)
3. See results that don't appear to contain the search term

**Expected behavior**

Perhaps the `199...` coupon shouldn't be there? _(See screenshot below)_

**Screenshots**

screen shot 2019-02-18 at 1 03 39 pm

joshuatf:
Fixes #1611 

Searches coupons by code to avoid mismatched results by description and blocking result sets > 10.

### Before
Screen Shot 2019-03-14 at 3 17 26 PM

### After
Screen Shot 2019-03-14 at 3 18 09 PM

1.  Add a coupon with a keyword in the description, but not in the coupon code (title).
2.  Search for that keyword under the coupons table.
3.  Check that the result is not returned.
4.  Check that the results in the network request for the autocompleter also do not return that coupon.

### DIFF
--- a/includes/api/class-wc-admin-rest-coupons-controller.php
+++ b/includes/api/class-wc-admin-rest-coupons-controller.php
@@ -24,4 +24,69 @@ class WC_Admin_REST_Coupons_Controller extends WC_REST_Coupons_Controller {
 	 */
 	protected $namespace = 'wc/v4';
 
+	/**
+	 * Get the query params for collections.
+	 *
+	 * @return array
+	 */
+	public function get_collection_params() {
+		$params                = parent::get_collection_params();
+		$params['search_code'] = array(
+			'description'       => __( 'Limit results to coupons with codes matching a given string.', 'woocommerce-admin' ),
+			'type'              => 'string',
+			'validate_callback' => 'rest_validate_request_arg',
+		);
+		return $params;
+	}
+
+
+	/**
+	 * Add coupon code searching to the WC API.
+	 *
+	 * @param WP_REST_Request $request Request data.
+	 * @return array
+	 */
+	protected function prepare_objects_query( $request ) {
+		$args = parent::prepare_objects_query( $request );
+
+		if ( ! empty( $request['search_code'] ) ) {
+			$args['search_code'] = $request['search_code'];
+		}
+
+		return $args;
+	}
+
+	/**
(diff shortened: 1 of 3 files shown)


emdashcodes (approved):
Code looks good, and this is testing well for me. 👍

includes/api/class-wc-admin-rest-coupons-controller.php
```diff
@@ -24,4 +24,69 @@ class WC_Admin_REST_Coupons_Controller extends WC_REST_Coupons_Controller {
...
+	 * @param object $wp_query WP_Query object.
+	 * @return string
+	 */
+	public static function add_wp_query_search_code_filter( $where, $wp_query ) {
+		global $wpdb;
+
+		$code_search = $wp_query->get( 'search_code' );
+		if ( $code_search ) {
+			$code_search = $wpdb->esc_like( $code_search );
+			$code_search = ' \'%' . $code_search . '%\'';
```
jeffstieler: Nitpick: using `"` here would avoid escape characters.

packages/components/src/search/autocompleters/coupons.js
```diff
@@ -29,7 +29,7 @@ export default {
 		let payload = '';
 		if ( search ) {
 			const query = {
-				search,
+				search_code: search,
```
jeffstieler: Can we avoid renaming this? _I sound like a broken record..._

jeffstieler (commented):
This tests well, just some questions/comments.

packages/components/src/search/autocompleters/coupons.js
joshuatf: No problemo- done!

includes/api/class-wc-admin-rest-coupons-controller.php
joshuatf: Added double quotes here as well as other PRs that override search params.

PR #1803, opened 2019-03-14
Fix date range dropdown for RTL locales

Aljullu:
Fixes second issue from #1776.

* Fixes date range dropdown alignment for RTL locales.
* Fixes calendar error tooltips positioning, until now they were not visible neither in RTL nor LTR locales.

### Screenshots
_Before:_
image

_After:_
image

- Using a RTL language, go to any report.
- Open the date picker dropdown.
- Verify the dropdown is correctly aligned.
- Enter an invalid date.
- Verify the '_Invalid Date_' popover is correctly positioned.
- Bonus points for testing there are no regressions in mobile and LTR languages.

### DIFF
--- a/packages/components/src/calendar/style.scss
+++ b/packages/components/src/calendar/style.scss
@@ -164,10 +164,12 @@
 	display: none;
 
 	.is-error .woocommerce-calendar__input-text:focus + span & {
+		/* rtl:begin:ignore */
 		display: block;
 		left: 50% !important;
 		position: absolute;
 		top: auto !important;
+		/* rtl:end:ignore */
 	}
 
 	.components-popover__content {
--- a/packages/components/src/calendar/style.scss
+++ b/packages/components/src/calendar/style.scss
@@ -151,10 +151,6 @@
 		&::placeholder {
 			color: $core-grey-dark-300;
 		}
-
-		&:focus + span .woocommerce-calendar__input-error {
-			display: block;
-		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


packages/components/src/calendar/style.scss
```diff
@@ -167,6 +163,13 @@
 .woocommerce-calendar__input-error {
 	display: none;
 
+	.is-error .woocommerce-calendar__input-text:focus + span & {
+		display: block;
+		left: 50% !important;
+		position: absolute;
+		top: auto !important;
+	}
```
psealock: Using Chrome and looking at a RTL language, the `left` inline style doesn't get overwritten:

Screen Shot 2019-03-15 at 9 11 04 AM

It looks like the `left: 50% !important;` gets rewritten by the rtl plugin to `right: 50% !important;`. If I change it back to `left: 50% !important;` in the console, it works.

psealock (commented):
I'm not sure if this is a browser issue, but I noted an oddity with the left/right centering of the tooltip error message.

Thanks for tackling this! It works in non RTL as well. My calendar isn't rendering either, could that be related?

packages/components/src/calendar/style.scss
Aljullu: You're right! I tested it in Firefox and wrongly assumed it would work in other browsers too.

I [made](https://github.com/woocommerce/woocommerce-admin/pull/1803/commits/d4847dcc5e3ecc7614cbe08f9ee05d1deeb82226) that CSS ruleset to be ignored by the RTL plugin, so it should work fine now in both browsers.

Aljullu:
Thanks for the review @psealock. Can you take another look now? :slightly_smiling_face: 

> My calendar isn't rendering either, could that be related?

Right, that's the last remaining issue from #1776. I'm trying to figure out what's going on there. I guess some CSS is interfering with `react-dates`, but I couldn't find yet what it is, so I don't know if both issues are related or they are something completely different.

psealock (approved):
LGTM!

PR #1807, opened 2019-03-15
Add onChange event to search focus to restore search results

### LINKED ISSUE #1806: Table search does not allow free text search after losing focus
**Describe the bug**
If you type in the  and lose focus on the input afterwards, the search no longer works when pressing enter and only allows dropdown autocomplete options to be selected.

**To Reproduce**

1. Go to any table with a search.
2. Start typing a query.
3. Click outside the input to remove focus.
4. Return focus to the input and try to hit enter to search.
5. Note that nothing happens.

**Expected behavior**
Expected the search to continue working.

**Screenshots**
Mar-15-2019 12-27-25
(Hitting enter at the end and nothing happens)

joshuatf:
Fixes #1806 

Triggers the `onChange` event when the search input is focused to:
* Restore search results.
* Allow free text searching again.

### Screenshots
Mar-15-2019 13-48-21

1. Type a query into any autocompleter.
2. Click outside the input to lose focus.
3. Click inside the input to refocus.
4. Note the results are restored and pressing 'enter' allows free text search.

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -1,6 +1,7 @@
 # (unreleased)
 - Chart legend component now uses withInstanceId HOC so the ids used in several HTML elements are unique.
 - Chart component: new prop `filterParam` used to detect selected items in the current query. If there are, they will be displayed in the chart even if their values are 0.
+- Expand search results and allow searching when input is refocused in autocompleter.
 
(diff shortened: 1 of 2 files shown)


Aljullu (approved):
Good catch with the issue! Changes look good to me. :shipit:

PR #1809, opened 2019-03-15
Add/clear jobs setting

joshuatf:
Fixes (partially) #1699

Adds the clear jobs tool to the settings page.

### Screenshots
Screen Shot 2019-03-15 at 2 36 52 PM

1. Go to the Settings page.
2. Use the rebuild report data tool to queue some jobs.
3. Use the newly added clear jobs tool.
4. Note the jobs are trashed in the `wp_posts` table and a success notice displays after the tool is ran.

### DIFF
--- a/includes/api/class-wc-admin-rest-system-status-tools-controller.php
+++ b/includes/api/class-wc-admin-rest-system-status-tools-controller.php
@@ -42,29 +42,4 @@ class WC_Admin_REST_System_Status_Tools_Controller extends WC_REST_System_Status
 			)
 		);
 	}
-
-	/**
-	 * Actually executes a tool.
-	 *
-	 * @param  string $tool Tool.
-	 * @return array
-	 */
-	public function execute_tool( $tool ) {
-		$ran     = true;
-		$message = '';
-
-		switch ( $tool ) {
-			case 'rebuild_stats':
-				WC_Admin_Api_Init::regenerate_report_data();
-				$message = __( 'Rebuilding reports data in the background . . .', 'wc-admin' );
-				break;
-			default:
-				return parent::execute_tool( $tool );
-		}
-
-		return array(
-			'success' => $ran,
-			'message' => $message,
-		);
-	}
 }
--- a/includes/class-wc-admin-reports-sync.php
+++ b/includes/class-wc-admin-reports-sync.php
@@ -80,7 +80,7 @@ class WC_Admin_Reports_Sync {
 	 */
 	public static function init() {
 		// Add report regeneration to tools REST API.
-		add_filter( 'woocommerce_debug_tools', array( __CLASS__, 'add_regenerate_tool' ) );
+		add_filter( 'woocommerce_debug_tools', array( __CLASS__, 'add_debug_tools' ) );
 
 		// Initialize syncing hooks.
 		add_action( 'wp_loaded', array( __CLASS__, 'orders_lookup_update_init' ) );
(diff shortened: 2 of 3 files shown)


includes/api/class-wc-admin-rest-system-status-tools-controller.php
```diff
@@ -42,29 +42,4 @@ public function get_tools() {
...
 		);
 	}
-
-	/**
-	 * Actually executes a tool.
-	 *
-	 * @param  string $tool Tool.
-	 * @return array
-	 */
-	public function execute_tool( $tool ) {
```
joshuatf: I don't think this is needed anymore since we run the tool through callbacks and only on the API.  Can anyone confirm this isn't used elsewhere?

includes/class-wc-admin-reports-sync.php
```diff
@@ -119,28 +119,36 @@ public static function clear_queued_actions() {
...
 			array(
-				'rebuild_stats' => array(
+				'rebuild_stats'        => array(
 					'name'     => __( 'Rebuild reports data', 'wc-admin' ),
 					'button'   => __( 'Rebuild reports', 'wc-admin' ),
 					'desc'     => __( 'This tool will rebuild all of the information used by the reports.', 'wc-admin' ),
 					'callback' => array( __CLASS__, 'regenerate_report_data' ),
 				),
+				'clear_analytics_jobs' => array(
+					'name'     => __( 'Analytics jobs', 'wc-admin' ),
```
joshuatf: Open to name changes here- not sure which is the most user-friendly and easiest to understand.

`Pending jobs`
`Queued jobs`
`Background jobs`
`Pending actions`
...

psealock: Pending analytics actions ?

psealock (commented):
I'm confused here. When testing, I use this button:

Screen Shot 2019-03-20 at 12 56 52 PM

which is not the one added here in this PR (the descriptions are different). Are the testing instructions not referring to Analytics > Settings ?

joshuatf:

>Pending analytics actions ?

Updated to this.

>which is not the one added here in this PR (the descriptions are different). Are the testing instructions not referring to Analytics > Settings ?

I'm not quite following here.  This PR adds the button you screenshotted on the Analytics->Settings page.  Did this one not work for you?

psealock:
>I'm not quite following here. This PR adds the button you screenshotted on the Analytics->Settings page. Did this one not work for you?

Ah ok, I see now (and blame coffee withdrawal). There are two buttons added. Where can I see the one added in `includes/class-wc-admin-reports-sync.php`? I'm looking in WooCommerce > Status > Tools and am not seeing it. Am I looking in the wrong place?

Also, now that you've updated the language, the other button should also reflect those changes.

joshuatf:
>Ah ok, I see now (and blame coffee withdrawal)

@psealock I feel your pain 😫☕️

> I'm looking in WooCommerce > Status > Tools and am not seeing it. Am I looking in the wrong place?

We're using a similar pattern to the build tools, where we add it to system tools so we can use it through the API, but don't display the button inside the WC Tools page since we already have it in Analytics->Settings so there should only be a one button.

includes/class-wc-admin-reports-sync.php
```diff
@@ -119,28 +119,36 @@ public static function clear_queued_actions() {
...
+					'button'   => __( 'Clear analytics jobs', 'wc-admin' ),
```
psealock: This button and name should reflect the new updated verbiage.

psealock (approved):
Thanks for the clarification Josh. LGTM, just a note about matching the language with the second button.

PR #1810, opened 2019-03-15
Fix missing date picker calendar in RTL locales

Aljullu:
Fixes the last issue in #1776.

- Fix the date picker calendar not visible in RTL locales.
- Create a `isRTL()` function instead of relying on `window.isRtl` to detect if the page is RTL or not. This way the components can be used in other environments.

### Screenshots
_Before:_
image

_After:_
image

- With a RTL locale, go to any report and open the date picker.
- Verify the calendar appears and you can go to the next and previous month.
- Make sure charts still render the Y-axis on the right side.

### DIFF
--- a/packages/components/src/chart/d3chart/chart.js
+++ b/packages/components/src/chart/d3chart/chart.js
@@ -32,6 +32,8 @@ import { getColor } from './utils/color';
 import ChartTooltip from './utils/tooltip';
 import { selectionLimit } from '../constants';
 
+const isRTL = () => document.documentElement.dir === 'rtl';
+
 /**
  * A simple D3 line and bar chart component for timeseries data in React.
  */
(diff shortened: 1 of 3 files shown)


packages/components/src/chart/d3chart/chart.js
```diff
@@ -32,6 +32,8 @@ import { getColor } from './utils/color';
 import ChartTooltip from './utils/tooltip';
 import { selectionLimit } from '../constants';
 
+const isRTL = () => document.documentElement.dir === 'rtl';
```
jeffstieler: Should this go in `./utils` to keep things DRY?

jeffstieler (commented):
This works as expected - just one comment about making the `isRTL()` function reusable/DRY.

packages/components/src/chart/d3chart/chart.js
Aljullu: Thanks for taking a look @jeffstieler!

You mean moving that function to `d3chart/utils/index.js` or creating a utils file so it's usable in other components?

In Gutenberg, they seem to repeat that function in each component that uses it ([#](https://github.com/WordPress/gutenberg/blob/master/packages/components/src/popover/utils.js#L7), [#](https://github.com/WordPress/gutenberg/blob/master/packages/components/src/date-time/date.js#L16)).

jeffstieler: > You mean moving that function to `d3chart/utils/index.js` or creating a utils file so it's usable in other components?

I was thinking a new utils file.

> In Gutenberg, they seem to repeat that function in each component that uses it

Ah, interesting pattern to establish. Let's not consider this a blocker then.

jeffstieler (approved):

PR #1816, opened 2019-03-15
Show chart hover circles in the correct order

### LINKED ISSUE #1815: Chart: lines and circles aren't rendered in the same order
Notice purple line is rendered on top of all the other ones but, when hovering, the green circle is the one rendered on top.
image

Aljullu:
Fixes #1815.

When hovering line charts, show circles in the same stacking order as the lines (following the legend order).

### Screenshots
_Before:_
image

_After (notice `Long Sleeve Tee` circle appears on top of `V-Neck T-Shirt`):_
image

- Go to any report.
- In `time-comparison` charts, make sure the `Current Period` circles appear always on top of `Previous Period` when hovering the lines.
- In `item-comparison` charts, make sure circles of items with higher total values always appear on top of items with lower values.

### DIFF
--- a/packages/components/src/chart/d3chart/utils/line-chart.js
+++ b/packages/components/src/chart/d3chart/utils/line-chart.js
@@ -22,8 +22,10 @@ import { smallBreak, wideBreak } from './breakpoints';
  * @param {function} xScale - from `getXLineScale`
  * @returns {array} that includes the date, start (x position) and width to mode the mouseover rectangles
  */
-export const getDateSpaces = ( data, uniqueDates, visibleKeys, width, xScale ) =>
-	uniqueDates.map( ( d, i ) => {
+export const getDateSpaces = ( data, uniqueDates, visibleKeys, width, xScale ) => {
+	const reversedKeys = visibleKeys.slice().reverse();
+
+	return uniqueDates.map( ( d, i ) => {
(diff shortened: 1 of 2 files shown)


packages/components/src/chart/d3chart/utils/line-chart.js
```diff
@@ -98,7 +101,8 @@ export const drawLines = ( node, data, params, scales, formats, tooltip ) => {
 		.attr( 'class', 'line-g' )
 		.attr( 'role', 'region' )
 		.attr( 'aria-label', d => d.label || d.key );
-	const dateSpaces = getDateSpaces( data, params.uniqueDates, width, scales.xScale );
+	const reversedKeys = params.visibleKeys.slice().reverse();
```
joshuatf: Will keys always need to be reversed when using this function?  Should we move this inside `getDateSpaces()` or is there a case where keys would already be in the order we need?

joshuatf (approved):
Testing well!  One question about possibly moving the reversing function, but otherwise LGTM.

packages/components/src/chart/d3chart/utils/line-chart.js
Aljullu: Absolutely! It makes more sense to move it inside `getDateSpaces()`.

Done in 95f8b2e38d6efffcbc91984ef753b31f23d57927.

PR #1821, opened 2019-03-18
Check if resources are still loading before skipping dispatch

### LINKED ISSUE #1814: When loading a report and switching to another tab immediately, the report never finishes loading
**To Reproduce**
1. Go to any report.
2. Refresh the page and as fast as you can, switch to another tab.
3. Once the report tab has finished loading, switch back to it.
4. Notice it never finishes loading.

**Screenshots**
_(First tab is `wc-admin`, second tab is `wordpress.org`)_
loading

joshuatf:
Fixes #1814 

Checks if new resources are being requested before skipping dispatch when the browser tab is not active as opposed to updating freshness when a tab is inactive.

1. Load/refresh any report and switch to another browser tab while the page is still loading.
2. Come back to the page after resources should have loaded and check that loading is not stalled.
3. Also check that freshness requests are not made when the tab is not active.  You can speed up testing by reducing default freshness in https://github.com/woocommerce/woocommerce-admin/blob/6f9e87ae02d3389a126f2056478372a8a36f3062/client/wc-api/constants.js#L11 

### Question
Are there performance issues with checking on all requests?  

I don't think there are any real performance implications since the array of resource names is generally small and we're checking if a property exists in the existing state, but we could move this inside the `document.hidden` block if this is a concern.

### DIFF
--- a/client/wc-api/wp-data-store/create-api-client.js
+++ b/client/wc-api/wp-data-store/create-api-client.js
@@ -20,8 +20,8 @@ function createDataHandlers( store ) {
 	return {
 		dataRequested: resourceNames => {
 			const { resources } = store.getState();
-			const newResources = resourceNames.filter( resourceName => ! resources[ resourceName ] );
-			if ( ! newResources.length && document.hidden ) {
+			const newResources = resourceNames.some( resourceName => ! resources[ resourceName ] );
+			if ( ! newResources && document.hidden ) {
 				return;
 			}
 			store.dispatch( {
--- a/client/wc-api/wp-data-store/create-api-client.js
+++ b/client/wc-api/wp-data-store/create-api-client.js
@@ -19,7 +19,9 @@ function createStore( name ) {
 function createDataHandlers( store ) {
 	return {
 		dataRequested: resourceNames => {
-			if ( document.hidden ) {
+			const { resources } = store.getState();
+			const newResources = resourceNames.filter( resourceName => ! resources[ resourceName ] );
+			if ( ! newResources.length && document.hidden ) {
 				return;
 			}
 			store.dispatch( {


client/wc-api/wp-data-store/create-api-client.js
```diff
@@ -19,7 +19,9 @@ function createStore( name ) {
 function createDataHandlers( store ) {
 	return {
 		dataRequested: resourceNames => {
-			if ( document.hidden ) {
+			const { resources } = store.getState();
+			const newResources = resourceNames.filter( resourceName => ! resources[ resourceName ] );
```
Aljullu: We can use [`some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) here, a little bit more performant and it returns a bool:
```suggestion
			const newResources = resourceNames.some( resourceName => ! resources[ resourceName ] );
```

Aljullu (commented):
I confirm this fixes #1814, but I see something weird. :confused: 

I might be doing something wrong, but I still see network requests when `document` is hidden and there are no `newResources`:
image

client/wc-api/wp-data-store/create-api-client.js
joshuatf: Sweet- thanks!

joshuatf:
>I might be doing something wrong, but I still see network requests when document is hidden and there are no newResources:

Good catch.  This returns early in the data handler, but that doesn't prevent the network request from being made.

@coderkevin Can [this line](https://github.com/Automattic/fresh-data/blob/4182afe0ac6e20d78354827cd1277a7265b5ecab/packages/framework/src/client/index.js#L198) be changed to something like:

```js
const resourceOperations = this.dataRequested( resourceNames );
```

So that we can filter operations and prevent them from running?  Or is there a better way of handling this?

Aljullu (approved):
I think we can merge this PR as-is and deal with the network requests in a follow-up or open an issue.

PR #566, opened 2018-10-18
Activity Panel: Inbox: New Sales Record Note

allendav:
See #176 

Adds/updates an activity panel: inbox: note if the previous day's sales are a new record.

Creates a note that will look like this in the inbox (can't be seen in React until #513 is merged):

screen shot 2018-10-18 at 3 57 37 pm

To test:
- update_option `woocommerce_sales_record_date` with a date in the past (e.g. `2018-09-22`)
- update_option `woocommerce_sales_record_amount` with a low value (e.g. `10.0`)
- Create a few orders totalling more than that value, complete them, and change their dates to yesterday.
- Switch to this branch if you haven't already
- Deactivate and re-activate the plugin to get the daily scheduled task added (will repeat daily).
- Install and activate the Admin Notes Test Bench https://github.com/automattic/wc-admin-notes-test-bench
- Navigate to wp-admin > Tools > WC Admin Notes Test Bench
- Look for the sales record note.

screen shot 2018-10-18 at 4 47 55 pm

- If you don't see the note there, install https://wordpress.org/plugins/wp-crontrol/  and navigate to wp-admin > Tools > Cron Events. Look for the wc_admin_daily event and click Run Now.

PR #1823, opened 2019-03-18
Activity Panel: Persist read status for inbox notes

### LINKED ISSUE #1817: Activity Panel: Persist Inbox Last View Time
Child Task of #176

The designs of he Activity Panel Inbox call for an "unread" indicator to be shown in two places, the first being in the Inbox Panel Launcher itself:

unread-notices

And the second spot, which you can not see in the current implementation is on each inbox notif card, from the [invision design](https://automattic.invisionapp.com/d/main#/console/14671248/305076999/preview):

image

It has been quite sometime, but we have had some discussion in the past around how the "read state" should be tracked. I believe what we finally landed on ( need to find a citation of this but /cc @justinshreve @jameskoster @allendav for any other memories ) was:

- We should persist a timestamp, likely on user meta, that records the last time the Inbox Panel is opened.
- This timestamp, could then be utilized to properly set the `unread` prop on the [Inbox Panel tab](https://github.com/woocommerce/woocommerce-admin/blob/master/client/header/activity-panel/index.js#L99) by comparing that time against the `date_created` of admin notes.
- Likewise, when the inbox is opened, ActivityCard's could then have their `unread` [prop](https://github.com/woocommerce/woocommerce-admin/blob/master/client/header/activity-panel/activity-card/index.js#L55) set based upon the "last viewed inbox" timestamp.

__Questions / Thoughts__
- How do we handle the event where the number of inbox notifs shown in the inbox can not all be viewed without scrolling? Do we need to listen to scroll events and set the last viewed timestamp based upon the oldest card that gets scrolled into view?
- If I recall, we also talked about the notion of - even if a card is shown in the view, does that really mean it has been read? Since the notes schema is not setup in such a way that we can track reads-per-user, I believe we felt the time-based unread approach was okay. But just wanted to verify that.
- Since all Activity Panels need to have an `unread` prop set, should we consider creating a REST endpoint specific to this bit of user data? I imagine that would be helpful to support a similar interface within the core mobile apps eventually... and would prevent the need of doing multiple API requests ( Inbox, Orders, Reviews, Stock ) to properly display the unread status.

joshuatf:
Fixes #1817 

Adds a user meta field for the last read time to compare against note creation date and add an "unread" icon.

### Screenshots
Screen Shot 2019-03-18 at 4 50 37 PM

1. Open the Inbox activity panel (user meta `wc_admin_activity_panel_inbox_last_read` should not yet be set) and make sure notes appear unread.
2. Close the panel and re-open.
3. Make sure notes are now all read and icon next to inbox disappears.
4. Add a new note and make sure it appears unread on first open of the activity panel.

### Questions
Do we want to limit updating the read time unless all notes are scrolled into the viewport?  My hunch is no, but further discussion [here](https://github.com/woocommerce/woocommerce-admin/issues/1817).

### DIFF
--- a/client/header/activity-panel/index.js
+++ b/client/header/activity-panel/index.js
@@ -284,7 +284,8 @@ export default withSelect( select => {
 		! Boolean( getNotesError( notesQuery ) ) &&
 		! isGetNotesRequesting( notesQuery ) &&
 		latestNote[ 0 ] &&
-		new Date( latestNote[ 0 ].date_created ).getTime() > userData.activity_panel_inbox_last_read;
+		new Date( latestNote[ 0 ].date_created_gmt ).getTime() >
+			userData.activity_panel_inbox_last_read;
 
 	if ( ! orderStatuses.length ) {
 		return { unreadNotes, unreadOrders: false };
--- a/client/header/activity-panel/index.js
+++ b/client/header/activity-panel/index.js
@@ -260,13 +260,14 @@ class ActivityPanel extends Component {
 }
 
 export default withSelect( select => {
-	const { getCurrentUserData,
+	const {
+		getCurrentUserData,
 		getNotes,
 		getNotesError,
 		getReportItems,
 		getReportItemsError,
 		isGetNotesRequesting,
-		isReportItemsRequesting
+		isReportItemsRequesting,
 	} = select( 'wc-api' );
 	const orderStatuses = wcSettings.wcAdminSettings.woocommerce_actionable_order_statuses || [
 		'processing',
--- a/client/header/activity-panel/index.js
+++ b/client/header/activity-panel/index.js
@@ -276,13 +276,13 @@ export default withSelect( select => {
 
 	const notesQuery = {
 		page: 1,
-		per_page: 3,
+		per_page: 1,
 	};
 
 	const latestNote = getNotes( notesQuery );
 	const unreadNotes =
-		! getNotesError &&
-		! isGetNotesRequesting &&
+		! Boolean( getNotesError( notesQuery ) ) &&
+		! isGetNotesRequesting( notesQuery ) &&
 		latestNote[ 0 ] &&
 		new Date( latestNote[ 0 ].date_created ).getTime() > userData.activity_panel_inbox_last_read;
 
--- a/client/header/activity-panel/index.js
+++ b/client/header/activity-panel/index.js
@@ -92,13 +92,13 @@ class ActivityPanel extends Component {
 
 	// @todo Pull in dynamic unread status/count
 	getTabs() {
-		const { unreadOrders } = this.props;
+		const { unreadNotes, unreadOrders } = this.props;
 		return [
 			{
 				name: 'inbox',
 				title: __( 'Inbox', 'woocommerce-admin' ),
 				icon: <Gridicon icon="mail" />,
-				unread: true,
+				unread: unreadNotes,
 			},
 			{
 				name: 'orders',
@@ -260,14 +260,34 @@ class ActivityPanel extends Component {
 }
 
 export default withSelect( select => {
-	const { getReportItems, getReportItemsError, isReportItemsRequesting } = select( 'wc-api' );
+	const { getCurrentUserData,
+		getNotes,
+		getNotesError,
+		getReportItems,
+		getReportItemsError,
+		isGetNotesRequesting,
+		isReportItemsRequesting
+	} = select( 'wc-api' );
 	const orderStatuses = wcSettings.wcAdminSettings.woocommerce_actionable_order_statuses || [
 		'processing',
 		'on-hold',
 	];
+	const userData = getCurrentUserData();
+
+	const notesQuery = {
+		page: 1,
+		per_page: 3,
+	};
+
+	const latestNote = getNotes( notesQuery );
+	const unreadNotes =
+		! getNotesError &&
+		! isGetNotesRequesting &&
+		latestNote[ 0 ] &&
+		new Date( latestNote[ 0 ].date_created ).getTime() > userData.activity_panel_inbox_last_read;
 
 	if ( ! orderStatuses.length ) {
-		return { unreadOrders: false };
+		return { unreadNotes, unreadOrders: false };
 	}
 
 	const ordersQuery = {
@@ -290,5 +310,5 @@ export default withSelect( select => {
 		}
 	}
 
-	return { unreadOrders };
+	return { unreadNotes, unreadOrders };
 } )( clickOutside( ActivityPanel ) );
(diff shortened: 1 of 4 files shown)


client/header/activity-panel/panels/inbox.js
```diff
@@ -19,8 +20,23 @@ import sanitizeHTML from 'lib/sanitize-html';
 import { QUERY_DEFAULTS } from 'wc-api/constants';
 
 class InboxPanel extends Component {
+	constructor( props ) {
+		super( props );
+		this.state = {
+			lastRead: props.lastRead,
```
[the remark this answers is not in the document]
joshuatf: Added this in https://github.com/woocommerce/woocommerce-admin/pull/1823/commits/7786711e2434687749635976dca82bf674d636ec

joshuatf:
Thanks for the review @Aljullu.  Added a similar approach in https://github.com/woocommerce/woocommerce-admin/pull/1823/commits/9b1f679f804351f62d98fec69f835b9f05240670 to the orders unread status you're working on by grabbing the latest note date.  After some thought, I believe this is a better approach than adding a secondary user meta field.  I'll go into more detail in #1824 

>I guess that would be very unintuitive for users. I prefer the approach you implemented in this PR.

Awesome.  Agreed, wanted to make sure I wasn't alone in thinking this way.

>Edit: just noticed your PR is a Work in Progress, so I don't know if you wanted a review yet.

No worries- good to get early feedback!  😄

client/header/activity-panel/panels/inbox.js
```diff
@@ -17,10 +17,18 @@ import ActivityHeader from '../activity-header';
 import { EmptyContent, Section } from '@woocommerce/components';
 import sanitizeHTML from 'lib/sanitize-html';
 import { QUERY_DEFAULTS } from 'wc-api/constants';
+import withSelect from 'wc-api/with-select';
 
 class InboxPanel extends Component {
+	componentWillUnmount() {
```
Aljullu: Just curious, why `componentWillUnmount` instead of `componentDidMount`? If a note is created while the user has the panel open, it will not be displayed to the user but the `activity_panel_inbox_last_read` date will be assigned to a later date than the note creation date, right?

client/header/activity-panel/index.js
```diff
@@ -254,4 +257,19 @@ class ActivityPanel extends Component {
...
+export default withSelect( select => {
+	const { getCurrentUserData, getNotes, getNotesError, isGetNotesRequesting } = select( 'wc-api' );
+	const userData = getCurrentUserData();
+	const notesQuery = {
+		page: 1,
+		per_page: 3,
+	};
+	const latestNote = getNotes( notesQuery );
+	const unreadNotes =
+		! getNotesError &&
```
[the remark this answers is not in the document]
joshuatf: Nice catch!  Fixed this.

client/header/activity-panel/index.js
```diff
@@ -254,4 +257,19 @@ class ActivityPanel extends Component {
 	}
 }
 
-export default clickOutside( ActivityPanel );
+export default withSelect( select => {
+	const { getCurrentUserData, getNotes, getNotesError, isGetNotesRequesting } = select( 'wc-api' );
+	const userData = getCurrentUserData();
+	const notesQuery = {
+		page: 1,
+		per_page: 3,
```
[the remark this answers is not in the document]
joshuatf: >And do you think we could use per_page: 1 instead of per_page: 3

`per_page` should be 1- accidentally committed `3` from testing 😓 

>What do you think about adding orderby: 'date' just to be sure it doesn't break if we change the default sorting parameter in the future?

The API currently doesn't offer any ordering options for notes.  It is sorted by ID by default.  I had a similar discussion with @allendav previously [here](https://github.com/woocommerce/woocommerce-admin/pull/566#discussion_r226788433).  I'm up for adding in date sorting if you think that's best, but might be easier to handle in another PR if that's okay with you.

client/header/activity-panel/panels/inbox.js
joshuatf: This tradeoff was needed to make it stateless (unless you have another idea here) in https://github.com/woocommerce/woocommerce-admin/pull/1823/commits/7786711e2434687749635976dca82bf674d636ec.

If we update in `componentDidMount` then the read time will always be greater than notes since we are comparing directly against the prop we just updated.

client/header/activity-panel/index.js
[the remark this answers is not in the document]
Aljullu: > might be easier to handle in another PR if that's okay with you

Of course, thanks for explaining that!

client/header/activity-panel/panels/inbox.js
Aljullu: Oh, I understand it now. That's a tricky issue.

I guess we could store the time the component is created in the `constructor` or `componentDidMount` and wait to call `updateCurrentUserData` in `componentWillUnmount`. Something like this:

```ES6
class InboxPanel extends Component {
	constructor() {
		this.mountTime = Date.now();
	}
	componentWillUnmount() {
		const userDataFields = {
			[ 'activity_panel_inbox_last_read' ]: this.mountTime,
		};
	}
	...
```

client/header/activity-panel/index.js
```diff
@@ -254,4 +257,19 @@ class ActivityPanel extends Component {
...
+	const notesQuery = {
+		page: 1,
+		per_page: 1,
+	};
+	const latestNote = getNotes( notesQuery );
+	const unreadNotes =
+		! Boolean( getNotesError( notesQuery ) ) &&
+		! isGetNotesRequesting( notesQuery ) &&
+		latestNote[ 0 ] &&
+		new Date( latestNote[ 0 ].date_created ).getTime() > userData.activity_panel_inbox_last_read;
```
Aljullu: After working on #1860, I think we will need to compare GMT dates here. Otherwise, if the store and the user's browser have a different time zone, this comparison might return unexpected results.

client/header/activity-panel/index.js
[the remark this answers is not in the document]
joshuatf: Issue noted here - https://github.com/woocommerce/woocommerce-admin/issues/1865

client/header/activity-panel/index.js
joshuatf: Updated in inbox and the panel unread indicator to use the notes GMT time instead.

client/header/activity-panel/index.js
```diff
@@ -260,14 +260,36 @@ class ActivityPanel extends Component {
...
 	];
+	const userData = getCurrentUserData();
+
+	const notesQuery = {
+		page: 1,
+		per_page: 1,
+	};
+
+	const latestNote = getNotes( notesQuery );
+	const unreadNotes =
```
timmyc: @joshuatf do you think this logic might also have a home in PHP - thinking down the road if/when mobile apps want to consume the inbox notifs - they too will need the ability to set an unread indicator.

timmyc:
So great to see this, nice work Josh - I'm catching up on all the goings-on after being sick - so a wee bit late to the 🎉 on this one.

PR #1833, opened 2019-03-19
Fix table links to internal reports

psealock:
Table links were pointed to incorrect URL in all tables except the Coupons Report 🤷‍♂️ .

I'm not sure what changed, but they were missing the `/analytics/` portion of the path.

Go to all tables and hover over links to ensure the base url paths are correctly constructed.

### DIFF
--- a/client/analytics/report/categories/breadcrumbs.js
+++ b/client/analytics/report/categories/breadcrumbs.js
@@ -56,7 +56,7 @@ export default class CategoryBreadcrumbs extends Component {
 			<div className="woocommerce-table__breadcrumbs">
 				{ this.getCategoryAncestors( category, categories ) }
 				<Link
-					href={ getNewPath( persistedQuery, 'categories', {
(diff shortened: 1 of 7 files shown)


joshuatf (approved):
Tests well for me, but links were still working okay for me on master as well.  I think the path was relative, so since these are already under the `/analytics` path, they were routed correctly.

I think this is probably a good move to be specific with this links anyways, so pre-approving, but curious if there's a way to replicate the links breaking that you were seeing.

psealock:
> but curious if there's a way to replicate the links breaking that you were seeing.

Maybe it was a dependency. I re `npm start`ed after checking out master.

Aljullu:
> I think this is probably a good move to be specific with this links anyways, so pre-approving, but curious if there's a way to replicate the links breaking that you were seeing.

I could replicate it as well, and seems to be fixed by this PR. :+1:

PR #1808, opened 2019-03-15
Fix#1801 sku filter issue on variation product.

ronakganatra9:
Fix#1801 sku filter issue on variation product.

Fixes #1802 

Added an SKU parameter in class-wc-admin-rest-reports-variation-controller.php
modified query params for SKU sort in datastores/class-wc-admin-reports-variations-data-store.php

### Screenshots

ezgif com-video-to-gif (6)

### Detailed test instructions:

- I have tested it after the changes and it works fine.

PR #1802, opened 2019-03-14
#1801 - Issue fix sku filter on variation products

ronakganatra9:
Fixes #1801 

Added an SKU parameter in class-wc-admin-rest-reports-variation-controller.php
modified query params for SKU sort in datastores/class-wc-admin-reports-variations-data-store.php

### Screenshots

ezgif com-video-to-gif (6)

### Detailed test instructions:

- I have tested it after the changes and it works fine.

Issue #1828, opened 2019-03-18
React Router shows many errors in the console


Aljullu:
It seems to be this issue:

https://github.com/ReactTraining/react-router/issues/6630#issuecomment-473775768

**To Reproduce**
1. Reinstall packages: `rm -rf ./node_modules` and  `npm i`.
2. Open any `wc-admin` page.
3. Notice there are several errors in the console.

**Screenshots**
![image](https://user-images.githubusercontent.com/3616980/54554255-575a7e00-49b4-11e9-8623-4a91df4121ae.png)

PR #1837, opened 2019-03-19
issue fix #1801 SKU filter issue fixed.

### LINKED ISSUE #1801: REST API: Issue when sorting the variation products via SKU in products analytics tabs in single product
**Describe the bug**
When We have multiple orders of the same product with different variation and that both products have different SKUs at that time there is a bug in REST API so when We click on sort by SKU on the product single page on analytics at that time it gives 400 error because there is not SKU parameter in REST API.
filter=single_product page.

File wc-includes/api/class-wc-admin-rest-reports-variations-controller.php

Line 287:
$params['orderby']       = array(
			'description'       => __( 'Sort collection by object attribute.', 'wc-admin' ),
			'type'              => 'string',
			'default'           => 'date',
			'enum'              => array(
				'date',
				'net_revenue',
				'orders_count',
				'items_sold',
			),
		);

It does not take sku option from order by params so it gives 400 error in the rest API response.
wc-admin

Also, the query is not set for the filter by SKU in variation in the data store files.

**To Reproduce**

1. Create a product with 2 or more variation or choose the product which has 2 or more variations.
2. Assign different SKUs to all variants
2. Place different orders of that same product with different variations.
3. Go, WooCommerce admin on analytics menu and click on products
4. Where you will see the product name(which you have ordered with different variation) and Item sold columns 
screenshot-localhost-2019 03 14-12-11-09

5. Click on that product
6. You will be able to see all products with variation
Like
screenshot-localhost-2019 03 14-14-37-46

7. Now you will be able to see both variations have different SKUs.
8. Click on SKU column to sort it.
9. You will see the following error,
ezgif com-video-to-gif (5)

10. The columns will become empty and in REST API response it wil return 400 error.
**Expected behavior**
It should Filter according to SKU in Ascending, descending flow.

**Screenshots**
ezgif com-video-to-gif (5)

* OS: Ubuntu
* Browser Chrome

**Additional context**
I have fixed it and attached a PR on it.

ronakganatra9:
Fixes #1801 #1802 #1808 

Attached a new PR

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-variations-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-variations-data-store.php
@@ -111,7 +111,7 @@ class WC_Admin_Reports_Variations_Data_Store extends WC_Admin_Reports_Data_Store
 	 * Updates the database query with parameters used for Products report: categories and order status.
 	 *
 	 * @param array  $query_args Query arguments supplied by the user.
-	 * @param string $from_arg Name of the FROM sql param.
+	 * @param string $from_arg   Name of the FROM sql param.
 	 *
 	 * @return array             Array of parameters used for SQL query.
 	 */
--- a/includes/data-stores/class-wc-admin-reports-variations-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-variations-data-store.php
@@ -110,7 +110,7 @@ class WC_Admin_Reports_Variations_Data_Store extends WC_Admin_Reports_Data_Store
 	/**
 	 * Updates the database query with parameters used for Products report: categories and order status.
 	 *
-	 * @param array $query_args Query arguments supplied by the user.
+	 * @param array  $query_args Query arguments supplied by the user.
 	 * @param string $from_arg Name of the FROM sql param.
 	 *
 	 * @return array             Array of parameters used for SQL query.
--- a/includes/data-stores/class-wc-admin-reports-variations-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-variations-data-store.php
@@ -94,7 +94,7 @@ class WC_Admin_Reports_Variations_Data_Store extends WC_Admin_Reports_Data_Store
 			$sql_query['order_by_clause'] = $this->normalize_order_by( $query_args['orderby'] );
 		}
 
-		if ( 'postmeta.meta_value' === $sql_query['order_by_clause'] ) {
+		if ( 'meta_value' === $sql_query['order_by_clause'] ) {
 			$sql_query['from_clause'] .= " JOIN {$wpdb->prefix}postmeta AS postmeta ON {$order_product_lookup_table}.variation_id = postmeta.post_id AND postmeta.meta_key = '_sku'";
 		}
 
@@ -159,7 +159,7 @@ class WC_Admin_Reports_Variations_Data_Store extends WC_Admin_Reports_Data_Store
 			return $table_name . '.date_created';
 		}
 		if ( 'sku' === $order_by ) {
-			return 'postmeta.meta_value';
+			return 'meta_value';
 		}
 
 		return $order_by;
--- a/includes/data-stores/class-wc-admin-reports-variations-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-variations-data-store.php
@@ -78,39 +78,41 @@ class WC_Admin_Reports_Variations_Data_Store extends WC_Admin_Reports_Data_Store
 		$this->report_columns['orders_count'] = str_replace( 'order_id', $table_name . '.order_id', $this->report_columns['orders_count'] );
 	}
 
-    /**
-     * Fills ORDER BY clause of SQL request based on user supplied parameters.
-     *
-     * @param array $query_args Parameters supplied by the user.
-     * @return array
-     */
-    protected function get_order_by_sql_params( $query_args ) {
-        global $wpdb;
-        $order_product_lookup_table = $wpdb->prefix . self::TABLE_NAME;
+	/**
+	 * Fills ORDER BY clause of SQL request based on user supplied parameters.
+	 *
+	 * @param array $query_args Parameters supplied by the user.
+	 *
+	 * @return array
+	 */
+	protected function get_order_by_sql_params( $query_args ) {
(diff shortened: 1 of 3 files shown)


psealock (commented):
Thanks for cleaning up the commit history with this one @ronakganatra9.

When I tested this, I needed to add `isSortable: true,` to this block to be able to sort the SKU column.

https://github.com/woocommerce/woocommerce-admin/blob/0f20052de6f9b706baaccfc6f8cb64c713c53d37/client/analytics/report/products/table-variations.js#L39-L43

Tests are also failing. One issue has already been noted https://github.com/woocommerce/woocommerce-admin/issues/1828. The other is a tabs v. spaces issue. 

I don't know why the pre-commit checks didn't pick up on this, but you can run the php codesniffer to automatically fix these:

```
 ./vendor/bin/phpcbf includes/data-stores/class-wc-admin-reports-variations-data-store.php includes/api/class-wc-admin-rest-reports-variations-controller.php
```

ronakganatra9:
Hello @psealock ,

I have implemented the changes and fixed the PHPCS errors and warnings.

includes/data-stores/class-wc-admin-reports-variations-data-store.php
```diff
@@ -127,6 +158,9 @@ protected function normalize_order_by( $order_by ) {
 		if ( 'date' === $order_by ) {
 			return $table_name . '.date_created';
 		}
+		if ( 'sku' === $order_by ) {
+			return 'postmeta.meta_value';
```
psealock: `WC_Admin_Reports_Products_Data_Store` uses just `meta_value` here. I don't think it matters, but any reason not to use the same?

psealock (commented):
Looking good @ronakganatra9 !

Just a small comment. A rebase to combine your commits into one and this is good to go.

Thanks again

includes/data-stores/class-wc-admin-reports-variations-data-store.php
ronakganatra9: Sorry, @psealock I am not understood what you want to explain to me. Can you please elaborate it little more.

ronakganatra9: Hi @psealock , I have tried the same but it was not working if I was using only meta_value. SO, I hav updated as postmeta.metavalue.
As, Comparing to the code both the files have minor structure changes. And it is working fine with postmeta.meta_value also.

Let me know your thoughts on the same.

Thanks & Regards,
Ronak Ganatra

includes/data-stores/class-wc-admin-reports-variations-data-store.php
```diff
@@ -78,11 +78,41 @@ public function __construct() {
...
+		global $wpdb;
+		$order_product_lookup_table = $wpdb->prefix . self::TABLE_NAME;
+
+		$sql_query['order_by_clause'] = '';
+		if ( isset( $query_args['orderby'] ) ) {
+			$sql_query['order_by_clause'] = $this->normalize_order_by( $query_args['orderby'] );
+		}
+
+		if ( 'postmeta.meta_value' === $sql_query['order_by_clause'] ) {
+			$sql_query['from_clause'] .= " JOIN {$wpdb->prefix}postmeta AS postmeta ON {$order_product_lookup_table}.variation_id = postmeta.post_id AND postmeta.meta_key = '_sku'";
```
Aljullu: I have been testing this when a comparison is active and couldn't get it working. You can reproduce it selecting a variable product, then selecting a _Comparison_ filter and selecting some of them:
image

This is a bit complex because the SQL command changes depending if a comparison is active or not. You can use `sku` sorting in the products table as a reference in case it helps:
https://github.com/woocommerce/woocommerce-admin/blob/master/includes/data-stores/class-wc-admin-reports-products-data-store.php#L135
Basically, you will have to make that join to be in the inner query when there is no comparison and to be in the outer query when there is a comparison.

I think the product endpoint is quite similar to this one, so taking a look at the file I linked, should be helpful. :+1:

Aljullu (commented):
Thanks for working on this @ronakganatra9! I have to say this is a complex issue, but it's looking good. I posted some comments that might help you with the work here.

includes/data-stores/class-wc-admin-reports-variations-data-store.php
ronakganatra9: Hi @psealock , Updated the changes.

includes/data-stores/class-wc-admin-reports-variations-data-store.php
ronakganatra9: @Aljullu , Thanks for reviewing these changes and sharing your thoughts. I am looking for this solution till that what I suggest is that let's create a separate issue for this and merge these changes so that it can be less chance of conflicts.

Aljullu: @ronakganatra9 thank you for addressing the other issues!

I would say it's better to wait until the feature is complete before merging it into master. Otherwise, users might try sorting by `sku` with an active comparison and that would make Woocommerce Admin to crash.

As far as I know, there is nobody currently working in the same files you modified in this PR, so I don't expect you to end up with many conflicts. In any case, if you have issues with that, we can help in solving them.

Does that sound good?

ronakganatra9: @Aljullu , Comparision is not available on products analytics page so it is not occurring the issue there. But in variation page of single product, there is a comparison option available so I recommend to push these changes or create a different issue for that because that is a complex issue.

Aljullu: Thanks for your patience here @ronakganatra9, there were so many things going on last week I didn't have time to take another look at this until now.

I created a follow-up PR that will fix the issue with the comparison, so you don't have to worry about that, and we can proceed merging this one. :slightly_smiling_face: 

Thank you again for your help!

Aljullu (approved):

includes/data-stores/class-wc-admin-reports-variations-data-store.php
ronakganatra9: Thank you so much @Aljullu for this.

PR #1838, opened 2019-03-19
Minify in safe mode for RTL styles to prevent stripping z-index

### LINKED ISSUE #1829: Activity Panel: z-index conflict with Editor
**Describe the bug**
When editing a product, if an activity panel is opened, there is a z-index conflict that has the editor/tinymce header floating above the Activity Panel

bug-z-index-product-editor

**To Reproduce**

1. Edit a product
2. Expand any Activity Panel

joshuatf:
Fixes #1829 

The forked `WebpackRTLPlugin` that we're using relies on cssnano v.3.7.1 which minimizes z-index values by default before it was removed by default in v4.

This PR turns that off 🚫 

### Screenshots
Screen Shot 2019-03-19 at 3 04 28 PM

1.  Run `npm start` to rebuild with the new cssnano config.
2.  Check that z-index is not stripped.  ( `.woocommerce-layout__header` should not overlap editor and have a z-index of "99991".)

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -160,6 +160,9 @@ const webpackConfig = {
 		} ),
 		new WebpackRTLPlugin( {
 			suffix: '-rtl',
+			minify: {
+				safe: true,
+			},
 		} ),
 		new MiniCssExtractPlugin( {
 			filename: './dist/[name]/style.css',


Aljullu (approved):
That was an unexpected bug! Thanks for finding the source of it.

LGTM :shipit:

psealock:
wow, nice work finding this one 😮

PR #1839, opened 2019-03-19
Update react router to v5

### LINKED ISSUE #1828: React Router shows many errors in the console

joshuatf:
Fixes #1828 

Updates react router and router dom to v5 to fix existing errors.

### Questions

V5 was just released and while v4.4.x versions would work, they were all released as alpha/beta versions.  While v5 is testing really well, I'm not finding any changelog or summary of differences for v5.  

Do we want to continue with v5 or downgrade back to an earlier version?

1.  Run `rm -rf node_modules` from the wc-admin directory.
2.  Run `npm install`.
3.  Start wc-admin and check that no console errors are thrown and test routing thoroughly to make sure there are no regressions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -148,7 +148,7 @@
     "react-click-outside": "3.0.1",
     "react-dates": "^18.0.4",
     "react-live": "1.12.0",
-    "react-router-dom": "4.3.1",
+    "react-router-dom": "^5.0.0",
     "react-slot-fill": "^2.0.1",
     "react-transition-group": "^2.4.0",
     "redux": "^4.0.0"


Aljullu (approved):
According to this (https://github.com/ReactTraining/react-router/releases/tag/v5.0.0), they say it's a minor release despite the major version bump. So I think it should be safe updating to 5.0.0. I also couldn't find any issue after doing some testing.

justinshreve:
Just as a follow-up to this, it looks like some of our behavior broke when updating to v5, but we didn't quite notice it.

https://github.com/woocommerce/woocommerce-admin/blob/master/packages/components/src/link/index.js#L21-L27

In `@woocommerce/components, `this.context.router` is always coming in null. And If you try to use `<Link>` directly, you get the following error: `Uncaught Error: Invariant failed: You should not use <Link> outside a <Router>` .

I **can** use `<Link>` in `client/` directly, just not anything in `@woocommerce/components`. We didn't notice this, I think, because directly linking to the `#hash` kept us on the page without needing to use the history API. When trying to switch to query string rendering, I noticed the issue -- and have to call `history` directly.  https://github.com/woocommerce/woocommerce-admin/compare/update/route-handling#diff-48e66aa13db262b895d3b598251429e9R18

I'm not sure if it's worth down grading back to v4. We can work around it if we switch our route handling, and it seems to be working fine with the hash URLs. There may be a solution related to how we are bundling the components. I just wanted to note the issue.

PR #1841, opened 2019-03-19
Hide stock panel and menu if stock management is disabled

### LINKED ISSUE #1826: Activity Panel: Hide Stock Panel if Stock Management Disabled
If a store has Inventory Management turned off, the Stock Activity Panel is still shown. Since the goal for this panel is to surface products that are either low stock or out-of-stock, it seems we shouldn't be showing the panel if the site is not being used to manage inventory. But /cc @LevinMedia and @jameskoster to confirm my assumption here.

stock-panel-manage-stock

If everyone is okay with the assumption above, I'm thinking we should not show the Stock panel if the "Enable Stock Management" is turned off in settings.

joshuatf:
Fixes #1826 

If the stock management setting is unchecked in WooCommerce, remove the Stock activity panel and menu item.

### Question

Do we want to also remove the stock column from product/variation reports?

### Screenshots
Screen Shot 2019-03-19 at 4 38 26 PM
Screen Shot 2019-03-19 at 4 38 16 PM

1.  Disable/enable stock management `wp-admin/admin.php?page=wc-settings&tab=products&section=inventory`
2. Check that the activity panel and menu item have been removed when turned off.

### DIFF
--- a/lib/admin.php
+++ b/lib/admin.php
@@ -148,13 +148,15 @@ function wc_admin_register_pages() {
 			)
 		);
 
-		wc_admin_register_page(
-			array(
-				'title'  => __( 'Stock', 'woocommerce-admin' ),
-				'parent' => '/analytics/revenue',
-				'path'   => '/analytics/stock',
-			)
-		);
+		if ( 'yes' === get_option( 'woocommerce_manage_stock' ) ) {
+			wc_admin_register_page(
+				array(
+					'title'  => __( 'Stock', 'woocommerce-admin' ),
+					'parent' => '/analytics/revenue',
+					'path'   => '/analytics/stock',
+				)
+			);
+		}
 
 		wc_admin_register_page(
 			array(
--- a/client/analytics/report/index.js
+++ b/client/analytics/report/index.js
@@ -72,11 +72,13 @@ const getReports = () => {
 			title: __( 'Downloads', 'woocommerce-admin' ),
 			component: DownloadsReport,
 		},
-		{
-			report: 'stock',
-			title: __( 'Stock', 'woocommerce-admin' ),
-			component: StockReport,
-		},
+		'yes' === wcSettings.manageStock
+			? {
+					report: 'stock',
+					title: __( 'Stock', 'woocommerce-admin' ),
+					component: StockReport,
+				}
+			: null,
 		{
 			report: 'customers',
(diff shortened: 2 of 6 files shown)


psealock (commented):
Testing nicely for me.

>Do we want to also remove the stock column from product/variation reports?

A column of `N/A` isn't helpful so I think so.

lib/admin.php
```diff
@@ -148,13 +148,15 @@ function wc_admin_register_pages() {
...
 		);
 
-		wc_admin_register_page(
-			array(
-				'title'  => __( 'Stock', 'woocommerce-admin' ),
-				'parent' => '/analytics/revenue',
-				'path'   => '/analytics/stock',
-			)
-		);
+		if ( 'yes' === get_option( 'woocommerce_manage_stock' ) ) {
```
Aljullu: I don't think it's blocking, but wanted to point it out. When checking/unchecking the stock checkbox, the menu item only appears/disappears on page refresh or when navigating from a non-wc-admin page to a wc-admin page.

So, if they go to _Settings_ to activate stock management and click on _Save_, when hovering _Analytics_, the _Stock_ report will not be visible yet. Not a big deal, but a small inconvenience that might be confusing?
image

I also noticed the report is available through the URL and it displays _Stock_ data even when stock management is disabled. Should we block any kind of navigation to that page completely? Or maybe showing an empty placeholder in case people end up there? Somebody might have the `wc-admin#/analytics/stock` URL in their history/bookmarks and might be confused seeing stock data while it's disabled in the settings.

Aljullu (approved):
Code looks good and its working fine for me too!

Just a small comment/thoughts.

lib/admin.php
joshuatf: I spent a bit of time toying with this.  One option is to check if the setting is actively being changed here, but I think the best solution is to change where options updated in WC instead.

Probably a minor issue, but opened an issue [here](https://github.com/woocommerce/woocommerce/issues/23087) to see if we can change the hook options are saved on.

joshuatf:
Thanks for the review @Aljullu and @psealock!

>A column of N/A isn't helpful so I think so.

I think you're right.  Updated in https://github.com/woocommerce/woocommerce-admin/pull/1841/commits/082f3aa4174744cb9ece50a81a2b2c3527e77116

Aljullu:
The last two commits are testing well too.

PR #1842, opened 2019-03-19
Fix#1768 Added a Margin of 40 pixel for better user interface, on mobile devices.

### LINKED ISSUE #1768: Mobile: Settings page has no bottom padding
**Describe the bug**
There is no spacing between the last line in _Settings_ and the end of the viewport.

**To Reproduce**
1. Go to _Analytics_ > _Settings_.
2. Notice the page has no bottom padding.

**Expected behavior**
There should be some padding at the bottom.

**Screenshots**

ronakganatra9:
Fixes #1768 

### Screenshots
screenshot-localhost-2019 03 19-15-34-11

- Open Settings page in the analytics tab on mobile.
- Now it is visible properly. First, there was no bottom space.

### DIFF
--- a/client/analytics/settings/index.scss
+++ b/client/analytics/settings/index.scss
@@ -11,9 +11,7 @@
 		margin-left: 15%;
 	}
 
-	@include breakpoint( '<782px' ) {
-		margin-bottom: 40px;
-	}
+		margin-bottom: $gap-largest;
 
 	button {
 		margin-right: $gap;
--- a/client/analytics/settings/index.scss
+++ b/client/analytics/settings/index.scss
@@ -11,6 +11,10 @@
 		margin-left: 15%;
 	}
 
+	@include breakpoint( '<782px' ) {
+		margin-bottom: 40px;
+	}
+
 	button {
 		margin-right: $gap;
 	}
--- a/client/analytics/settings/index.scss
+++ b/client/analytics/settings/index.scss
@@ -7,11 +7,11 @@
 }
 
 .woocommerce-settings__actions {
+	margin-bottom: $gap-largest;
+	
 	@include breakpoint( '>1280px' ) {
 		margin-left: 15%;
 	}
-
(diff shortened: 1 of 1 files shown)


client/analytics/settings/index.scss
```diff
@@ -11,6 +11,10 @@
 		margin-left: 15%;
 	}
 
+	@include breakpoint( '<782px' ) {
+		margin-bottom: 40px;
```
Aljullu: We have a SCSS variable that matches `40px` named [`$gap-largest`](https://github.com/woocommerce/woocommerce-admin/blob/master/client/stylesheets/abstracts/_variables.scss#L8), and I think it would be great to use it here. So what do you think about changing this line like this:
```suggestion
		margin-bottom: $gap-largest;
```

ronakganatra9: Hi Ajay, That is great. We can use that variable. I have applied the chnages and commited it.

client/analytics/settings/index.scss
```diff
@@ -11,6 +11,8 @@
 		margin-left: 15%;
 	}
 
+		margin-bottom: $gap-largest;
```
Aljullu: Thank you for applying the changes @ronakganatra9!

Could you just remove the extra tab here?

```suggestion
	margin-bottom: $gap-largest;
```

And one more suggestion, what do you think about moving it above the previous block so it looks like this?
```SCSS
.woocommerce-settings__actions {
	margin-bottom: $gap-largest;

	@include breakpoint( '>1280px' ) {
		margin-left: 15%;
	}
```
This way we have the generic declarations on top and the media queries below.

Once that's done, I will proceed merging.

ronakganatra9: Sounds Good, Applied. Thanks.

ronakganatra9:
Thank you @Aljullu

PR #1844, opened 2019-03-19
Travis: escape hashes (#) from branch names when downloading the ZIP file

### LINKED ISSUE #1843: Travis integration is broken if branch name contains a hash (#)
In `./install-wp-tests.sh` we are fetching a .zip file using the branch name to generate the URL, but if it contains a hash (`#`) it should be escaped to (`%23`).

This could be noted in this PR: #1842, where Travis was failing even though there were no errors.

Aljullu:
Fixes #1843.

Escape hashes from branch names in `install-wp-tests.sh` before attempting to download the ZIP file.

- If Travis passes in this PR, it means it's fixed. :slightly_smiling_face:


ryelle (approved):
test is passing 👍

PR #1846, opened 2019-03-19
Table: update columns visibility on headers change

### LINKED ISSUE #1840: User column preferences no longer respected on page reload
**Describe the bug**
User selected columns in the following reports don't seem to be working:

- Products
- Categories
- Downloads
- Stock

Note that these still show up in the JSON within the page, but the report itself does not seem to use them for the initial state.

**To Reproduce**

1. Go to one of the above reports.
2. Select columns to hide.
3. Refresh.
4. Note how bad those columns would be at a game of hide-and-seek.

**Expected behavior**
Expected user preferences to persist.

**Screenshots**
JSON for hidden columns within HTML:
Screen Shot 2019-03-19 at 4 16 43 PM

Shown columns:
Screen Shot 2019-03-19 at 4 16 57 PM

Aljullu:
Fixes #1840.

This PR:
- Sets `userPrefColumns` early in `<ReportTable>`, so that prop isn't undefined in loading tables.
- Makes it so table column's visibility updates when headers change, without checking whether `isLoading` prop is different.
- Fixes a couple of typos I found. :writing_hand: 

### Screenshots
userprefs

* Hide or unhide some columns.
* Refresh the page.
* Verify columns visibility is persisted.
* Repeat the same for _Categories_, _Downloads_ and _Stock_ reports.

### DIFF
--- a/packages/components/src/table/index.js
+++ b/packages/components/src/table/index.js
@@ -24,7 +24,7 @@ import { getIdsFromQuery, getSearchWords, updateQueryString } from '@woocommerce
  */
 import Card from '../card';
 import CompareButton from '../filters/compare/button';
-import DowloadIcon from './download-icon';
+import DownloadIcon from './download-icon';
 import EllipsisMenu from '../ellipsis-menu';
 import MenuItem from '../ellipsis-menu/menu-item';
 import MenuTitle from '../ellipsis-menu/menu-title';
@@ -317,7 +317,7 @@ class TableCard extends Component {
 							onClick={ this.onClickDownload }
 							isLink
 						>
-							<DowloadIcon />
+							<DownloadIcon />
 							<span className="woocommerce-table__download-button__label">
 								{ labels.downloadButton || __( 'Download', 'woocommerce-admin' ) }
 							</span>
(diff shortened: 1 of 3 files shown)


packages/components/src/table/index.js
```diff
@@ -91,6 +91,10 @@ class TableCard extends Component {
 		}
 	}
 
+	getShowCols( headers ) {
+		return headers.map( ( { key, hiddenByDefault } ) => ! hiddenByDefault && key ).filter( Boolean );
```
joshuatf: When I was checking this out yesterday, I was a bit confused before remembering that we also use `hiddenByDefault` to determine hidden status and not just the default.

Completely optional, but thoughts on changing this to just `hidden`?

joshuatf (approved):
Thanks for tackling this, Albert!

One comment about a variable name while we're here, but optional and everything is testing well!

packages/components/src/table/index.js
Aljullu: I'm not 100% sure about this. In theory, a 3rd party could use the `<Table>` component without an `onColumnsChange()` prop. So columns' visibility is stored completely in the state but the `headers` prop is never updated. I think that's why that prop is named `hiddenByDefault`, because not necessarily it goes in sync with `state.showCols`.

Do you think that makes sense or is it too confusing?

PR #1848, opened 2019-03-19
Add "remind me later" functionality to StoreAlert component.

### LINKED ISSUE #1756: Add "Remind me later" ability to StoreAlert Component
The original designs for StoreAlerts (#115) contain a "remind me later" action. 

screen shot 2019-03-05 at 3 17 26 pm

The WC_Notes class looks like it has a date_reminder field and a way to resurface these later, but no UI or method exists so notes can easily expose the dropdown. There is also no way for specific notice to mark a notice as "snoozable". This might require a new db field. Finally, the REST API also needs updated to allow for filtering snoozed notes out. Right now the date field does nothing. Some of these might be good to handle if we are refactoring the note class a bit.

- [ ] `WC_Admin_Note` should gain a way of marking that a note should have a reminder.
- [ ] The `StoreAlert` component should output a dropdown containing options for reminding later.
- [ ] An action should be dispatched to update the note `date_reminder` field when selected.

@LevinMedia  What should the options be for the "remind me later" dropdown across these notes?

Note: Despite the screenshot, the historical processing note should not have the remind me later dropdown. Notices should have the option of showing it, and [according to Slack](https://a8c.slack.com/archives/C9K5T0XSP/p1552419050100300), we will use it in major update notices.

jeffstieler:
Fixes #1756.

This PR seeks to add the ability to "snooze" `StoreAlert`s by:

* Introducing a new boolean `is_snoozable` column to `wc_admin_notes`
* Hook up `date_reminder` column to endpoint
* Add "remind me later" selection to `<StoreAlert/>` when `is_snoozable` is `true`
* Add recurring scheduled action to "unsnooze" actions when their `date_reminder` is reached

### Accessibility

### Screenshots

Screen Shot 2019-03-19 at 4 10 07 PM

Screen Shot 2019-03-19 at 4 10 25 PM

- Update your `wc_admin_version` option to be something less than `0.8.0`
- Verify that your `wc_admin_notes` table has the `is_snoozable` column
- Create a admin note with the `is_snoozable` set to `true`
- Verify that the note displays with the "remind me later" dropdown button
- Choose an option from that button
- Verify that the note disappears and has status `snoozed` with `date_reminder` populated
- Test "unsnooze" by modifying the `date_reminder` value to be in the past and running the `wc_admin_unsnooze_admin_notes` scheduled action _( Tools > Scheduled Actions > Pending )_
- Verify that the note reappears

### DIFF
--- a/client/layout/store-alerts/index.js
+++ b/client/layout/store-alerts/index.js
@@ -130,6 +130,7 @@ class StoreAlerts extends Component {
 
 		const snooze = alert.is_snoozable && (
 			<Dropdown
+				className="woocommerce-store-alerts__snooze"
 				position="bottom"
 				expandOnMobile
 				renderToggle={ ( { isOpen, onToggle } ) => (
--- a/client/layout/store-alerts/index.js
+++ b/client/layout/store-alerts/index.js
@@ -10,6 +10,7 @@ import interpolateComponents from 'interpolate-components';
 import { compose } from '@wordpress/compose';
 import { noop } from 'lodash';
 import { withDispatch } from '@wordpress/data';
+import moment from 'moment';
 
 /**
  * WooCommerce dependencies
@@ -63,9 +64,10 @@ class StoreAlerts extends Component {
 	}
 
 	renderActions( alert ) {
+		const { updateNote } = this.props;
 		const actions = alert.actions.map( action => {
 			const markStatus = () => {
-				this.props.updateNote( alert.id, { status: action.status } );
+				updateNote( alert.id, { status: action.status } );
 			};
 			return (
 				<Button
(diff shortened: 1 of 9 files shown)


client/layout/store-alerts/index.js
```diff
@@ -62,6 +63,109 @@ class StoreAlerts extends Component {
...
+			);
+		} );
+
+		// TODO: should "next X" be the start, or exactly 1X from the current date?
+		const snoozeOptions = [
+			{
+				newDate: moment()
+					.add( 4, 'hours' )
+					.unix(),
+				label: __( 'Later Today', 'woocommerce-admin' ),
```
timmyc: Yeah maybe this should just say, "In a few more hours" since it could technically be the next day right?

includes/class-wc-admin-install.php
```diff
@@ -168,6 +168,7 @@ private static function get_schema() {
 			source varchar(200) NOT NULL,
 			date_created datetime NOT NULL default '0000-00-00 00:00:00',
 			date_reminder datetime NULL default null,
+			is_snoozable boolean DEFAULT 0 NOT NULL,
```
timmyc: best column name ever.

includes/class-wc-admin-note.php
```diff
@@ -243,6 +246,16 @@ public function get_date_reminder( $context = 'view' ) {
 		return $this->get_prop( 'date_reminder', $context );
 	}
 
+	/**
+	 * Get note snoozability.
```
timmyc: 🤣

includes/class-wc-admin-notes.php
```diff
@@ -12,6 +12,25 @@
...
+
+	/**
+	 * Schedule events and hook appropriate actions.
+	 */
+	public static function init() {
+		$queue = WC()->queue();
+		$next  = $queue->get_next( self::UNSNOOZE_HOOK );
+
+		if ( ! $next ) {
+			$queue->schedule_recurring( time(), HOUR_IN_SECONDS, self::UNSNOOZE_HOOK );
```
timmyc: so this should only really be hit one time correct? or if someone purges out jobs manually....

tests/api/admin-notes.php
```diff
@@ -200,7 +200,32 @@ public function test_get_actioned_notes() {
...
 
 		// get_notes returns all results since 'status' is not one of actioned or unactioned.
-		$this->assertEquals( 2, count( $notes ) );
+		$this->assertEquals( 3, count( $notes ) );
+	}
+
+	/**
+	 * Test note "unsnoozing".
+	 */
+	public function test_note_unsnoozing() {
```
timmyc: Thanks for the test Jeff 👍

timmyc (approved):
Really nice Jeff - this is testing out quite well for me. I just had one question in the codes but going to give this the 👍 pre-approval

LevinMedia (approved):
LGTM. Thanks!

client/layout/store-alerts/index.js
jeffstieler: As implemented yes it might be the next day. We could alternatively make it say, 6pm of the current day and hide the option if it's after 6?

includes/class-wc-admin-notes.php
jeffstieler: Yep. Modeled after the marketplace suggestions: https://github.com/woocommerce/woocommerce/blob/5492369fe935b94aa6a06e1acbd2a58ea0fc09c3/includes/admin/marketplace-suggestions/class-wc-marketplace-updater.php#L25-L36

PR #1860, opened 2019-03-20
Activity Panel: set reviews unread indicator based on real data

### LINKED ISSUE #1819: Activity Panel: Hook up Reviews Panel Unread to Real Data
Related #1817

See the above issue for more context / possible approach to implement a  REST endpoint to help with this. Currently the Reviews Activity Panel `unread` indicator is [hard-coded](https://github.com/woocommerce/woocommerce-admin/blob/master/client/header/activity-panel/index.js#L117) to `true`.  This should be updated to only be true if pending reviews exist.

Aljullu:
Fixes #1819.

Uses real data in order to show or hide the unread indicator next to the _Reviews_ Activity Panel tab.

### Screenshots
reviews3

- Create a product review.
- Go to any `wc-admin` page and verify an unread indicator (red dot) appears next to the _Reviews_ tab of the Activity Panel.
- Click on the tab and verify the new review is also marked with a red dot.
- Close the panel and verify the red dot disappeared from the tab. Click on it again and verify the review doesn't have the red dot anymore either.


client/header/activity-panel/index.js
```diff
@@ -260,35 +261,68 @@ class ActivityPanel extends Component {
...
-	if ( ! isError && ! isRequesting ) {
-		if ( totalOrders > 0 ) {
-			unreadOrders = true;
-		} else {
-			unreadOrders = false;
+		if ( ! isReviewsError && ! isReviewsRequesting ) {
+			numberOfReviews = totalReviews;
+			unreadReviews =
+				reviews.length &&
+				reviews[ 0 ] &&
```
psealock: Maybe this second check should be `reviews[ 0 ]. date_created_gmt` because `new Date( undefined + 'Z')` is "Invalid Date".

psealock (approved):
LGTM @Aljullu,

I see the unread indicator appear on comments that are pre-approved or once they are approved, but not ones that are awaiting approval. It seems to be a good thing to tackle in a follow up PR.

joshuatf (approved):
Nice- testing well and LGTM!

Noted the same issue @psealock did with unapproved reviews, which we'll definitely want to follow up on since those are probably more "actionable."

On a side note, are there plans for adding content here when no reviews exist?  Maybe a small message like `You have no product reviews yet.`?

client/header/activity-panel/index.js
Aljullu: Fixed in: 11bc68aa3e5f18f2d54aa37fbabe56f54f02278c.

Aljullu:
Thanks for the reviews @psealock and @joshuatf:

> I see the unread indicator appear on comments that are pre-approved or once they are approved, but not ones that are awaiting approval. It seems to be a good thing to tackle in a follow up PR.

> Noted the same issue @psealock did with unapproved reviews, which we'll definitely want to follow up on since those are probably more "actionable."

Created #1868 and will work on it today.

> On a side note, are there plans for adding content here when no reviews exist? Maybe a small message like `You have no product reviews` yet.?

Created an issue for that too: #1869.

PR #1861, opened 2019-03-20
Made DELETE responses consistent with WP core (for v4 API)

mikejolley:
This PR implements a different response when forcing DELETE so that we're aligned with WP core.

v3 API mostly returns the old deleted object in the response. However, WP core for posts/pages returns an array in this format:

```php
{
    "deleted": true,
    "previous": {
        ...old object here
    }
}
```

This was raised in https://github.com/woocommerce/woocommerce/issues/22833

My PR copies across some of our v3 endpoints which use `delete_item` in https://github.com/woocommerce/woocommerce-admin/commit/704db97ae891c648217e4d1038ceb2fe3f591c3e I made no changes to the methods themselves in this commit to make review easier. 

https://github.com/woocommerce/woocommerce-admin/commit/719e27c0e1232e32790c194ee699196c6020e5f4 implements to improved delete_item code. I tested each endpoint in my client to ensure valid responses were recieved. To test, delete something, with force=true, and check the response.

Endpoints covered:
- DELETE /orders/x/notes/x
- DELETE /products/x/
- DELETE /products/x/variations/x
- DELETE /products/attributes/x/
- DELETE /shipping/zones/x/methods/x
- DELETE /shipping/zones/x
- DELETE /taxes/classes/x
- DELETE /webhooks/x

This however does need more work and I could use some help doing this, or we just need to wait for a core merge...

1. I'm not sure how to extend `WC_REST_CRUD_Controller` because of the way the `_ADMIN_` classes extend WC core, I have no way of making changes to it's `delete_item` unless I copy `delete_item` to all endpoints. Thats not something I want to do at this point :D The change thats needed in there is similar to the one made in the products controller.

2. I was not able to make changes to delete_item in `WC_Admin_REST_Customers_Controller`. Something about the code in WC-Admin is breaking that endpoint for me. I see no returned records and thus cannot test DELETE


includes/api/class-wc-admin-rest-shipping-zone-methods-controller.php
```diff
@@ -0,0 +1,91 @@
...
+		);
+
+		/**
+		 * Fires after a product review is deleted via the REST API.
+		 *
+		 * @param object           $method
+		 * @param WP_REST_Response $response        The response data.
+		 * @param WP_REST_Request  $request         The request sent to the API.
+		 */
+		do_action( 'rest_delete_product_review', $method, $response, $request );
```
claudiosanches: Wrong hook name.

includes/api/class-wc-admin-rest-webhooks-controller.php
```diff
@@ -0,0 +1,76 @@
...
+		);
+
+		/**
+		 * Fires after a single item is deleted or trashed via the REST API.
+		 *
+		 * @param WC_Webhook       $webhook     The deleted or trashed item.
+		 * @param WP_REST_Response $response The response data.
+		 * @param WP_REST_Request  $request  The request sent to the API.
+		 */
+		do_action( "woocommerce_rest_delete_webhook_object", $webhook, $response, $request );
```
claudiosanches: No need for double quotes.

includes/api/class-wc-admin-rest-product-reviews-controller.php
```diff
@@ -52,4 +52,72 @@ protected function prepare_links( $review ) {
...
+		}
+
+		/**
+		 * Fires after a product review is deleted via the REST API.
+		 *
+		 * @param object           $product_review  The deleted item.
+		 * @param WP_REST_Response $response        The response data.
+		 * @param WP_REST_Request  $request         The request sent to the API.
+		 */
+		do_action( 'rest_delete_product_review', $product_review, $response, $request );
```
claudiosanches: Already have this on core, but this should be called `woocommerce_rest_delete_review`.

includes/api/class-wc-admin-rest-product-reviews-controller.php
```diff
@@ -52,4 +52,72 @@ protected function prepare_links( $review ) {
...
+					'previous' => $previous->get_data(),
+				)
+			);
+		} else {
+			if ( ! $supports_trash ) {
+				return new WP_Error( 'rest_trash_not_supported', __( 'The product review does not support trashing.', 'woocommerce' ), array( 'status' => 501 ) );
+			}
+
+			if ( 'trash' === $product_review->comment_approved ) {
+				return new WP_Error( 'rest_already_trashed', __( 'The comment has already been trashed.', 'woocommerce' ), array( 'status' => 410 ) );
```
claudiosanches: We prefix our errors with new messages, generally prefixed with `woocommerce_`

coderkevin:
This PR has been sitting for a while, what needs to be done to get it ready to review/merge? Is help needed?

mikejolley:
@coderkevin I need to find the time to revise it - it's on my list.

PR #1863, opened 2019-03-20
Try: Method to publish core ready features

### LINKED ISSUE #1859: Remove deprecated locale data loader for i18n
When activating the plugin I see the following notice:

> Notice: gutenberg_get_jed_locale_data is deprecated since version 5.2.0 with no alternative available. 

If we require at least WP 5.0, we can use `wp_set_script_translations` without [the fallbacks.](https://github.com/woocommerce/woocommerce-admin/blob/2032996fa4610ad538c6338a39fa4abe2035393b/lib/client-assets.php#L89-L101)

We also need to attach `wp_set_script_translations` to every script that has translations in it, though, so that the translations generated from wp.org are attached to the correct files. Right now it's only attached to the main app script.

### LINKED ISSUE #1894: Implement process for publishing changes to WooCommerce core
When features are ready to graduate from development or the feature plugin, we need a method of publishing changes to WooCommerce core, while still being able to develop updates to them here in the repository.

This can be done using feature flags and [the approach Blocks and Gutenberg are using to publish features to WooCommerce core](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/456).

This task involves building a npm package containing WooCommerce Admin, creating a PR in WooCommerce core to consume and load it, and documenting the process.

This is mostly implemented in https://github.com/woocommerce/woocommerce-admin/pull/1863 already, but more code needs refactored before it is ready.

emdashcodes:
Fixes #1894. Fixes #1859. This PR adds the ability to build a package containing WooCommerce Admin for use in WooCommerce core. This is the proposed process that we can use to release and merge features to WooCommerce core in the future, [based on the approach Blocks and Gutenberg are using](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/456.). It also allows us to continue developing future updates/features in this codebase thanks to feature flags once features have launched.

I originally started this PR to support onboarding work but having an idea of how we might launch things to core is important to figure out (especially since it involves refactoring to get working). We’ve also talked about this for the activity panel feature (p1552436360111200-slack-wc-admin).

The biggest changes around here is refactoring the PHP code. `WC_Admin_Loader` is now responsible for loading assets, conditionally loading PHP feature code, and adding in the necessary `wcSettings` bits not already in core. This file is imported to core and renamed `WC_Admin_Library` (and processed so text-domains and constants are updated). This prevents the “class X is already registered” errors, and allows us to do some unhooking for more iteration in the plugin.

I think this PR is ready for a more final review and could be merge ready. Please note that certain features would still need more work to be "launchable" -- mainly moving over the REST API endpoints. See question and discussions below.

Related core `try` branch to look at: https://github.com/woocommerce/woocommerce/compare/try/wc-admin-library-package (This contains compiled code, including PHP code -- all the uncollapsed files are worth looking at though).

To Test:
* Deactivate the WooCommerce Admin Plugin.
* `npm pack`
* Inspect the generated `Tarball Contents` output. A `dist` folder and `includes` folder should be included.
* Copy the generated `woocommerce-admin-library-0.10.0.tgz` archive to your WooCommerce core checkout.
* From the core checkout  `git checkout` the `try/wc-admin-library-package` branch. Code: https://github.com/woocommerce/woocommerce/compare/try/wc-admin-library-package
* From the core checkout, `npm install`.
* From the core checkout, run `grunt`.
* Go to `/wp-admin/admin.php?page=wc-admin#/onboarding` and verify that an example component loads.
* Go to the orders page, and see the breadcrumbs show up -- as well as a temporary message. Activity Panel should be disabled.
* Back in the `woocommerce-admin` repo/folder, run  `npm start` to build the dev version.
* Activate the WooCommerce Admin plugin.
* Refresh the onboarding page. You should now see the activity panel.
* Dashboard & Analytics Routes should work.

---

Questions from my original PR/review:

* How should we handle API endpoints and namespacing? It seems like API endpoints should never be merged this way, since they live elsewhere in the core codebase. Should we set a namespace in `wcSettings` and use that in the JS so core will use manually merged endpoints, and the plugin will use plugin endpoints?
* How should we deal with PHP code that is needed for multiple different features? For example, the dashboard feature uses analytics data/endpoints mainly right now.  Should this code live in the analytics folder? We could combine the dashboard and analytics features, but the onboarding feature will also put stuff on the dashboard page. Instead of [automatically loading feature PHP code](https://github.com/woocommerce/woocommerce-admin/compare/try/feature-core-refactor?expand=1#diff-1ae4b2e500c5ac085f2c35ba19fc730aR113),  based on feature-names, we could add a `files` array to the `config/core.json` file, that contains which PHP files should be copied over.
* What should we display or do when a route/page doesn’t exist? This isn’t really related to this PR — but something I noticed as I was testing out different environments.


includes/class-wc-admin-loader.php
```diff
@@ -0,0 +1,533 @@
...
+
+	/**
+	 * Hooks extra neccessary data into the component settings array already set in WooCommerce core.
+	 *
+	 * @param array $settings Array of component settings.
+	 * @return array Array of component settings.
+	 */
+	public static function add_component_settings( $settings ) {
+		// @todo Some of these component settings should be added by feature specific code.
+		$preload_data_endpoints = array(
```
ryelle: Maybe this could use a filter, so the other components you mention can just hook into the filter?

includes/class-wc-admin-loader.php
```diff
@@ -0,0 +1,533 @@
...
+		<?php
+	}
+
+	/**
+	 * Hooks extra neccessary data into the component settings array already set in WooCommerce core.
+	 *
+	 * @param array $settings Array of component settings.
+	 * @return array Array of component settings.
+	 */
+	public static function add_component_settings( $settings ) {
```
ryelle: I wonder if we should pull this wcSettings code out entirely into woocommerce, out of both these packages (this is more the "true source" than what's in block-library). leave it filterable so the feature plugins can alter it, but then we could remove it from the block-library code… 

nothing actionable here, just thinking out loud 🤔

package.json
```diff
@@ -158,5 +158,14 @@
...
       "pre-push": "npm run -s install-if-no-packages && node bin/pre-push-hook.js"
     }
-  }
+  },
+  "files": [
+    "dist/**/*.css",
+    "dist/**/*.js",
+    "dist/feature-config-core.php",
+    "includes/class-wc-admin-loader.php",
+    "includes/onboarding/*.php",
```
ryelle: Each added feature will also have to add a line here too, right? (just checking that I understand the process)

includes/class-wc-admin-loader.php
```diff
@@ -0,0 +1,533 @@
...
+		wp_register_script(
+			WC_ADMIN_APP,
+			self::get_url( "{$entry}/index.js" ),
+			array( 'wc-components', 'wc-navigation', 'wp-date', 'wp-html-entities', 'wp-keycodes', 'wp-i18n' ),
+			self::get_file_version( "{$entry}/index.js" ),
+			true
+		);
+
+		if ( function_exists( 'wp_set_script_translations' ) ) {
+			wp_set_script_translations( WC_ADMIN_APP, 'woocommerce-admin' );
```
ryelle: I mentioned this in #1859 (which happens to be fixed in this PR), but you'll need to add this translations call for each script that has i18n functions.

ryelle (commented):
Woo, this is working! I tested it just as the plugin, then built it and tried the woocommerce build flow, and lastly re-activated the woocommerce-admin plugin. Everything worked as described.

When testing the woocommerce side of things, I got some PHP notices:

> PHP Notice:  Undefined index: id in /woocommerce/includes/wc-admin/class-wc-admin-library.php on line 476
> PHP Notice:  Undefined index: value in /woocommerce/includes/wc-admin/class-wc-admin-library.php on line 476

I also noticed the sidebar is a little disorganized (I expected Dashboard to be first), but I'm not sure if that's related to this PR

Screen Shot 2019-03-21 at 11 39 14 AM

> It seems like API endpoints should never be merged this way, since they live elsewhere in the core codebase. 

What we did in blocks worked well there, I think– get the endpoints to a "stable" state, then merge them into wc core and remove from the feature plugin. For releasing this as a feature plugin you'll need to bundle the API, but for once this is ready to merge to core I think the API should be merged separately, manually. (all IMO, of course)

> How should we deal with PHP code that is needed for multiple different features?

Could a shared parent class help with this kind of thing? I'm not totally sure I understand the issue enough 😅

package.json
justinshreve: Yes, unfortunately. I want to do some more investigation to see if there is a better way than having to add each one here. If we do end up having to add each one, I'll make sure that is documented better.

includes/class-wc-admin-loader.php
justinshreve: I agree with you here 👍. I think we can leave it for 3.6, but I'll open an issue to track this.

package.json
ryelle: I think a different folder + greedy search might be best, since it could be easy to accidentally forget to update this.

includes/class-wc-admin-loader.php
justinshreve: Good idea. 

Added a filter here https://github.com/woocommerce/woocommerce-admin/pull/1863/commits/3d8ab59c461dbe2f9131d68dc40bd82280b9e219#diff-1ae4b2e500c5ac085f2c35ba19fc730aR450, and hooked in from the specific features: https://github.com/woocommerce/woocommerce-admin/pull/1863/commits/3d8ab59c461dbe2f9131d68dc40bd82280b9e219#diff-1c906050672400ebd97de57489183b8cR44

package.json
justinshreve: 👍 I've moved all of these to `includes/features/` so we can just look at `includes/features/**/*.php`.

includes/class-wc-admin-loader.php
justinshreve: Fixed in https://github.com/woocommerce/woocommerce-admin/pull/1863/commits/3d8ab59c461dbe2f9131d68dc40bd82280b9e219.

includes/class-wc-admin-loader.php
```diff
@@ -0,0 +1,533 @@
...
+		// @todo Some of these component settings should be added by feature specific code.
+		$preload_data_endpoints = array(
+			'countries'             => '/wc/v4/data/countries',
+			'performanceIndicators' => '/wc/v4/reports/performance-indicators/allowed',
+		);
+
+		if ( function_exists( 'gutenberg_preload_api_request' ) ) {
+			$preload_function = 'gutenberg_preload_api_request';
+		} else {
+			$preload_function = 'rest_preload_api_request';
```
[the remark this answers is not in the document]
justinshreve: Thanks! These were left over from before, but I dropped these checks since we are indeed requiring WP 5+.

justinshreve:
Thanks you for the review! I've rebased this with the latest master changes, and addressed all the feedback above:

* Fixed dashboard link
* Fixed core id/value notices
* Removed `gutenberg_preload_api_request` check
* Added a filter for preloading data, and moved those endpoints out into their respective feature files
* Moved feature files so we can do a greedier search
* Fixed translation calls

I'm going to set this back to `In Progress`, and address some of the `@todo`s I left in the code -- especially around the embedded features like the activity panel. Those are the major blockers IMO to get this in a mergeable state.

Feedback still welcome!

includes/features/analytics/class-wc-admin-analytics.php
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	/**
+	 * Preload data from the performance indiciators endpoint.
+	 *
+	 * @param array $endpoints Array of preloaded endpoints.
+	 * @return array
+	 */
+	public function add_preload_endpoints( $endpoints ) {
+		$endpoints['countries'] = WC_ADMIN_API_NAMESPACE . '/data/countries';
```
justinshreve: This getting replaced can actually throw a concat linter error in core -- so I might need to do this differently.

```
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
44 | ERROR | String concat is not required here; use a single string
   |       | instead
   |       | (Generic.Strings.UnnecessaryStringConcat.Found)
----------------------------------------------------------------------
```

client/layout/index.js
```diff
@@ -63,12 +63,13 @@ class Layout extends Component {
...
 			<div className="woocommerce-layout">
-				{ window.wcAdminFeatures[ 'activity-panels' ] && <Slot name="header" /> }
+				<Slot name="header" />
 				<TransientNotices />
 
 				<div className="woocommerce-layout__primary" id="woocommerce-layout__primary">
 					{ window.wcAdminFeatures[ 'store-alerts' ] && <StoreAlerts /> }
-
+					{ window.wcAdminFeatures.onboarding &&
+						isEmbeded && <strong>Testing. @todo Remove this before merging.</strong> }
```
justinshreve: Adding this as an example of a piece of code behind a feature flag, that would get exposed in core on an existing WooCommerce page.

config/core.json
```diff
@@ -4,7 +4,7 @@
 		"analytics": false,
 		"dashboard": false,
 		"devdocs": false,
-		"onboarding": false,
+		"onboarding": true,
```
justinshreve: I'll set this to false before merge. Setting to `true` for an example of at least one feature being active.

justinshreve:
I've made more updates here and wrapped up some `@todo`s I had in the core. The breadcrumbs / embedded app entry point is now functional and I think this is ready for another round of reviews.

coderkevin:
Hey @justinshreve I think this PR needs a rebase now, to resolve conflicts.

justinshreve:
Thanks for the heads up, I have just rebased this. Actually I rebased it twice since it literally got a conflict minutes after I did it the first time 🙃.

Considering the refactor, moving of files, etc here hopefully we can get some eyes on this soon, otherwise it will be a PITA to keep up to date. Reviews are welcome even if there are conflicts (just to help move things along).

psealock (approved):
Nice one Justin! 

This is testing exactly as it should according to the directions. I'm trying to work out the mechanics of whats happening here with all the moving parts and I must say, this is really cool how the configs work to apply different features in each environment.

Separately, this simple regex in Woo's grunt build step may be problematic (css classnames and un snake cased filter names come to mind). 

```js
// Replace textdomain.
.replace( /'woocommerce-admin'/g, "'woocommerce'" )
```
In any case, I think this is a working first step that can be iterated upon. As for your questions, I can't comment intelligently until I learn more about how API versioning works in Woo.

justinshreve:
> Separately, this simple regex in Woo's grunt build step may be problematic (css classnames and un snake cased filter names come to mind).

The regex there will only replace 'woocommerce-admin' (note the single quotes -- which is enforced by the linter) -- so I think we are OK in terms of CSS classnames. A similar regex is in use for Product Blocks in core.

includes/features/activity-panel/class-wc-admin-activity-panel.php
```diff
@@ -0,0 +1,73 @@
...
+
+	/**
+	 * Things to do after WooCommerce updates.
+	 */
+	public function woocommerce_updated_note() {
+		WC_Admin_Notes_Settings_Notes::add_notes_for_settings_that_have_moved();
+	}
+}
+
+new WC_Admin_Analytics();
```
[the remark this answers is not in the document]
justinshreve: Thanks for catching that. Fixed.

ryelle (approved):
Looks good – went through the whole test flow and it all worked as expected 👍

PR #1864, opened 2019-03-20
Clear advanced filters when last one is removed

psealock:
When removing the last advanced filter, automatically "clear all". 

It seemed odd to me that you can remove the last filter and then have to press "clear all" in order to fully remove the filter from the report. I went ahead and removed all filters from the report when the user x's out of the last advanced filter.

### Screenshots

remove

1. Go to a report with advanced filters, such as Orders Report.
2. Filter by an advanced filter.
3. Press filter and see the report refresh.
4. Remove the filter by clicking the "X".
5. See the report go back to its unfiltered state and the "Filter" button disabled.

### DIFF
--- a/packages/components/src/filters/advanced/index.js
+++ b/packages/components/src/filters/advanced/index.js
@@ -19,6 +19,7 @@ import {
 	getDefaultOptionValue,
 	getNewPath,
 	getQueryFromActiveFilters,
+	getHistory,
 } from '@woocommerce/navigation';
 
 /**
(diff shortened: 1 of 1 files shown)


joshuatf (approved):
Sweet!

On that same note, I find it a bit odd that when you remove any filter you have to click the "Filter" button again to update.  What do you about doing this whenever any filter is removed?

Pre-approving as this PR is testing well.

psealock:
>What do you about doing this whenever any filter is removed?

Interesting. That does make sense, I'm thinking about how far to go with this because you could remove the "Filter" button altogether and have everything update right away.

PR #1866, opened 2019-03-21
Add low product stock indicator

### LINKED ISSUE #1827: Activity Panel: Hook Up unread indicator on Stock Panel
Another issue in the series started with #1817 - this time the task involves making the `unread` indicator on the Stock Panel work against real data. For the stock panel, products that are either low on stock, or out of stock should be counted as "actionable" and should appropriately set the `unread` indicator properly.

The Stock Panel has yet to be implemented, but we can still get the unread prop in place.

joshuatf:
Fixes #1827 

* Adds a low stock option to the products REST API
* Adds totalCount to items in wc-api
* Adds the unread indicator to the stock panel
* Extracts previous unread indicators to separate file

### Screenshots
Screen Shot 2019-03-21 at 3 14 39 PM

#### Products API
1. Make a request to the products controller with `low_in_stock` param set to a true value `wp-json/wc/v4/products/?low_in_stock=1`
2. Check that products returned are in fact low in stock.
3. Make sure the products returned respect the WC low stock threshold.
4. Set a low stock threshold unique to a product.
5. Make sure that the product only shows up under low stock when its value is under its specific threshold.

#### Stock unread indicator
1. Add at least some products that are considered low in stock.
2. Check that the unread indicator for the stock status is present.
3. Adjust stock settings so that none are considered low.
4.  Check that the unread indicator is no longer present.
5. Check that no regressions have occurred with the other indicators (due to minor refactor).

### Thoughts
* My original plan was to use an option to mark when a product became low stock, but this would become tricky with many out of stock products and would require extra queries on each order placement and product updates.  Plus this groundwork would have been needed anyway for querying these products once we actually implement the stock panel.
* The current solution will probably benefit from caching.  I initially set this to use a transient in https://github.com/woocommerce/woocommerce-admin/commit/19c8228b4e87eceb905abb1b5429b2cd71dd2f11, but we lose the ability to poll this data with freshdata.  We might want to follow this one up with some caching since this is a relatively strenuous query.

### DIFF
--- a/includes/api/class-wc-admin-rest-products-controller.php
+++ b/includes/api/class-wc-admin-rest-products-controller.php
@@ -127,7 +127,7 @@ class WC_Admin_REST_Products_Controller extends WC_REST_Products_Controller {
 		}
 
 		if ( $wp_query->get( 'low_in_stock' ) ) {
-			$low_stock_amount = get_option( 'woocommerce_notify_low_stock_amount', 2 );
+			$low_stock_amount = absint( max( get_option( 'woocommerce_notify_low_stock_amount' ), 1 ) );
 			$where           .= " AND lis_postmeta2.meta_key = '_manage_stock' AND lis_postmeta2.meta_value = 'yes'
 			AND lis_postmeta.meta_key = '_stock' AND lis_postmeta.meta_value IS NOT NULL
 			AND (
--- a/includes/api/class-wc-admin-rest-products-controller.php
+++ b/includes/api/class-wc-admin-rest-products-controller.php
@@ -81,7 +81,7 @@ class WC_Admin_REST_Products_Controller extends WC_REST_Products_Controller {
 		$args = parent::prepare_objects_query( $request );
 
 		if ( ! empty( $request['search'] ) ) {
-			$args['search'] = $request['search'];
+			$args['search'] = trim( $request['search'] );
 			unset( $args['s'] );
 		}
 		if ( ! empty( $request['low_in_stock'] ) ) {
@@ -118,7 +118,7 @@ class WC_Admin_REST_Products_Controller extends WC_REST_Products_Controller {
 	public static function add_wp_query_filter( $where, $wp_query ) {
 		global $wpdb;
 
-		$search = trim( $wp_query->get( 'search' ) );
+		$search = $wp_query->get( 'search' );
 		if ( $search ) {
 			$search = $wpdb->esc_like( $search );
 			$search = "'%" . $search . "%'";
@@ -151,7 +151,7 @@ class WC_Admin_REST_Products_Controller extends WC_REST_Products_Controller {
 	public static function add_wp_query_join( $join, $wp_query ) {
 		global $wpdb;
 
-		$search = trim( $wp_query->get( 'search' ) );
+		$search = $wp_query->get( 'search' );
 		if ( $search && wc_product_sku_enabled() ) {
 			$join .= " INNER JOIN {$wpdb->postmeta} AS ps_post_meta ON ps_post_meta.post_id = {$wpdb->posts}.ID";
 		}
@@ -175,7 +175,7 @@ class WC_Admin_REST_Products_Controller extends WC_REST_Products_Controller {
 	public function add_wp_query_group_by( $groupby, $wp_query ) {
 		global $wpdb;
 
-		$search       = trim( $wp_query->get( 'search' ) );
+		$search       = $wp_query->get( 'search' );
 		$low_in_stock = $wp_query->get( 'low_in_stock' );
 		if ( empty( $groupby ) && ( $search || $low_in_stock ) ) {
 			$groupby = $wpdb->posts . '.ID';
--- a/includes/api/class-wc-admin-rest-products-controller.php
+++ b/includes/api/class-wc-admin-rest-products-controller.php
@@ -55,8 +55,14 @@ class WC_Admin_REST_Products_Controller extends WC_REST_Products_Controller {
 	 * @return array
 	 */
 	public function get_collection_params() {
-		$params           = parent::get_collection_params();
-		$params['search'] = array(
+		$params                 = parent::get_collection_params();
+		$params['low_in_stock'] = array(
+			'description'       => __( 'Limit result set to products that are low or out of stock.', 'woocommerce-admin' ),
+			'type'              => 'boolean',
+			'default'           => false,
+			'sanitize_callback' => 'wc_string_to_bool',
+		);
+		$params['search']       = array(
 			'description'       => __( 'Search by similar product name or sku.', 'woocommerce-admin' ),
 			'type'              => 'string',
 			'validate_callback' => 'rest_validate_request_arg',
@@ -78,6 +84,9 @@ class WC_Admin_REST_Products_Controller extends WC_REST_Products_Controller {
 			$args['search'] = $request['search'];
 			unset( $args['s'] );
 		}
+		if ( ! empty( $request['low_in_stock'] ) ) {
+			$args['low_in_stock'] = $request['low_in_stock'];
+		}
 
 		return $args;
 	}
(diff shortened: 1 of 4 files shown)


includes/api/class-wc-admin-rest-products-controller.php
```diff
@@ -89,50 +98,67 @@ protected function prepare_objects_query( $request ) {
...
 		}
 
+		if ( $wp_query->get( 'low_in_stock' ) ) {
+			$where .= " AND lis_postmeta2.meta_key = '_manage_stock' AND lis_postmeta2.meta_value = 'yes'
+			AND lis_postmeta.meta_key = '_stock' AND lis_postmeta.meta_value IS NOT NULL
+			AND (
+				lis_postmeta3.meta_key = '_low_stock_amount' AND lis_postmeta.meta_value IS NOT NULL
+				AND lis_postmeta3.meta_key = '_low_stock_amount' AND CAST(lis_postmeta.meta_value AS SIGNED) <= CAST(lis_postmeta3.meta_value AS SIGNED)
+				OR lis_postmeta3.meta_key = '_low_stock_amount' AND lis_postmeta.meta_value IS NULL
+				AND lis_postmeta.meta_key = '_stock' AND CAST(lis_postmeta.meta_value AS SIGNED) <= '{$low_stock_amount}'
```
jeffstieler: Looks like `$low_stock_amount` didn't make the jump from https://github.com/woocommerce/woocommerce-admin/pull/1866/commits/19c8228b4e87eceb905abb1b5429b2cd71dd2f11#diff-04ebb0cb4b33c92c166ab779a502d791R530.

Also, could we delete the first two commits?

includes/api/class-wc-admin-rest-products-controller.php
```diff
@@ -145,14 +171,14 @@ public static function add_wp_query_product_search_join( $join, $wp_query ) {
...
 	 * @return string
 	 */
-	public function add_wp_query_product_search_group_by( $groupby, $wp_query ) {
+	public function add_wp_query_group_by( $groupby, $wp_query ) {
 		global $wpdb;
 
-		$search = trim( $wp_query->get( 'search' ) );
-		if ( $search && empty( $groupby ) ) {
+		$search       = trim( $wp_query->get( 'search' ) );
+		$low_in_stock = trim( $wp_query->get( 'low_in_stock' ) );
```
jeffstieler: Why `trim()` here? If `prepare_objects_query()` handled coercion, would this be needed?

includes/api/class-wc-admin-rest-products-controller.php
```diff
@@ -78,6 +84,9 @@ protected function prepare_objects_query( $request ) {
 			$args['search'] = $request['search'];
```
jeffstieler: Should there be a `trim()` here? _(related to a comment below)_

includes/api/class-wc-admin-rest-products-controller.php
```diff
@@ -78,6 +84,9 @@ protected function prepare_objects_query( $request ) {
...
 			unset( $args['s'] );
 		}
+		if ( ! empty( $request['low_in_stock'] ) ) {
+			$args['low_in_stock'] = $request['low_in_stock'];
```
jeffstieler: Should this be coerced to a boolean?

includes/api/class-wc-admin-rest-products-controller.php
joshuatf: You're right, not necessary.

includes/api/class-wc-admin-rest-products-controller.php
joshuatf: Whoops- thanks!

>Also, could we delete the first two commits?

Yep.  Wanted to make sure this was the right direction before dropping those.

includes/api/class-wc-admin-rest-products-controller.php
joshuatf: It's already trimmed elsewhere, but we can trim here instead and remove the other trims to make this more manageable.

joshuatf:
@jeffstieler Thanks for the review!  This one has been rebased and is ready for another look.

includes/api/class-wc-admin-rest-products-controller.php
```diff
@@ -89,50 +98,68 @@ protected function prepare_objects_query( $request ) {
...
 		if ( $search ) {
 			$search = $wpdb->esc_like( $search );
 			$search = "'%" . $search . "%'";
-			$where .= ' AND (' . $wpdb->posts . '.post_title LIKE ' . $search;
+			$where .= " AND ({$wpdb->posts}.post_title LIKE {$search}";
 			$where .= wc_product_sku_enabled() ? ' OR ps_post_meta.meta_key = "_sku" AND ps_post_meta.meta_value LIKE ' . $search . ')' : ')';
 		}
 
+		if ( $wp_query->get( 'low_in_stock' ) ) {
+			$low_stock_amount = get_option( 'woocommerce_notify_low_stock_amount', 2 );
```
jeffstieler: We need to either SQL escape this or parse it as an integer (`get_option` is filterable).

joshuatf: Updated in https://github.com/woocommerce/woocommerce-admin/pull/1866/commits/e466d4029fda1ef2528af988060c2391c3706cd4

includes/api/class-wc-admin-rest-products-controller.php
```diff
@@ -127,7 +127,7 @@ public static function add_wp_query_filter( $where, $wp_query ) {
 		}
 
 		if ( $wp_query->get( 'low_in_stock' ) ) {
-			$low_stock_amount = get_option( 'woocommerce_notify_low_stock_amount', 2 );
+			$low_stock_amount = absint( max( get_option( 'woocommerce_notify_low_stock_amount', 2 ) ) );
```
jeffstieler: `max()` is unnecessary here and will throw a warning if the first argument isn't an array.

```
php > var_dump( max( 'a' ) );
Warning: max(): When only one parameter is given, it must be an array
```

joshuatf: @jeffstieler Changed this back to not use `max()`, but do you know why they do this throughout core?  I assumed there was some plugin that used an array as most instances are written as:

```php
absint( max( get_option( 'woocommerce_notify_low_stock_amount', 2 ) ) );
```

jeffstieler: @joshuatf the core uses of `max()` provide a second argument: https://github.com/woocommerce/woocommerce/blob/f9cdab66eb1f61f6f26001838199aa1a5b8ba01f/includes/admin/reports/class-wc-report-low-in-stock.php#L39

From https://www.php.net/max:
> If the first and only parameter is an array, max() returns the highest value in that array. If at least two parameters are provided, max() returns the biggest of these values.

joshuatf: 🤦‍♂️ I kept reading this as the second parameter that acts as the default value in `get_option()`.  Going to swap this for a max with `1` since this seems to be the most commonly used in core for low stock reporting.

jeffstieler:
Thanks for sticking through this long review @joshuatf.

I have some questions about the query being run for low stock products:

_(I've formatted this by hand, hopefully correctly, and added comments)_

```sql
SELECT SQL_CALC_FOUND_ROWS wp_posts.ID
FROM wp_posts
INNER JOIN wp_postmeta AS lis_postmeta
	ON wp_posts.ID = lis_postmeta.post_id
INNER JOIN wp_postmeta AS lis_postmeta2
	ON wp_posts.ID = lis_postmeta2.post_id
INNER JOIN wp_postmeta AS lis_postmeta3
	ON wp_posts.ID = lis_postmeta3.post_id
WHERE 1=1
AND wp_posts.post_type = 'product'
AND ((wp_posts.post_status = 'publish'))
AND lis_postmeta2.meta_key = '_manage_stock'
AND lis_postmeta2.meta_value = 'yes'
AND lis_postmeta.meta_key = '_stock'
AND lis_postmeta.meta_value IS NOT NULL
AND (
	lis_postmeta3.meta_key = '_low_stock_amount'
	AND lis_postmeta.meta_value IS NOT NULL # redundant condition outside of parens
	AND lis_postmeta3.meta_key = '_low_stock_amount' # redundant with 2 lines above
	AND CAST(lis_postmeta.meta_value AS SIGNED) <= CAST(lis_postmeta3.meta_value AS SIGNED)
	OR
	lis_postmeta3.meta_key = '_low_stock_amount'
	AND lis_postmeta.meta_value IS NULL # contradictory with condition outside of parens
	AND lis_postmeta.meta_key = '_stock' # redundant condition outside of parens
	AND CAST(lis_postmeta.meta_value AS SIGNED) <= '2' # casting null and comparing?
)
GROUP BY wp_posts.ID
ORDER BY wp_posts.post_date DESC, wp_posts.ID DESC
LIMIT 0, 1
```

I'm noticing some redundant/contradictory conditions in here, mostly around `lis_postmeta.meta_value` being null or not. Can you double check this?

Also, `$low_stock_amount` is a string in the query, not sure if that's intentional.

joshuatf:
@jeffstieler Thanks for the review! 
* The post meta null comparisons were meant to be against `lis_postmeta3`.  Fixed those up.
* Also removed/extracted some of those redundant conditions.

```sql
AND lis_postmeta2.meta_key = '_manage_stock'
AND lis_postmeta2.meta_value = 'yes'
AND lis_postmeta.meta_key = '_stock'
AND lis_postmeta.meta_value IS NOT NULL
AND lis_postmeta3.meta_key = '_low_stock_amount'
AND (
	lis_postmeta3.meta_value IS NOT NULL
	AND CAST(lis_postmeta.meta_value AS SIGNED) <= CAST(lis_postmeta3.meta_value AS SIGNED)
	OR lis_postmeta3.meta_value IS NULL
	AND CAST(lis_postmeta.meta_value AS SIGNED) <= {$low_stock_amount}
)
```

jeffstieler (approved):
Everything is looking good! :shipit:

jeffstieler:
I'm going to resolve the conflict and merge this - it will be really helpful for hooking up the stock panel.

joshuatf:
Thanks for reviewing and fixing those conflicts @jeffstieler!

PR #1867, opened 2019-03-21
Create DEFAULT_ACTIONABLE_STATUSES constant

Aljullu:
Given that we were defining actionable statuses in several places, I think it makes sense to store that in a constant.

- Delete `woocommerce_actionable_order_statuses` from `wp_options` table in your site's database.
- Go to _Settings_ and verify _Actionable Statuses_ is set to _Processing_ and _On hold_.
- Verify only _Processing_ and _On hold_ orders appears in the _Orders_ tab of the _Activity Panel_.

### DIFF
--- a/client/analytics/settings/config.js
+++ b/client/analytics/settings/config.js
@@ -12,6 +12,11 @@ import interpolateComponents from 'interpolate-components';
  */
 import { Link } from '@woocommerce/components';
 
+/**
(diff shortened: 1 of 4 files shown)


jeffstieler (approved):
Nice and simple. 🛳 it.

_EDIT: resolve conflicts first 🙃_

PR #1871, opened 2019-03-21
Load on hold reviews in the Activity Panel

### LINKED ISSUE #1868: Activity Panel: Reviews awaiting approval do not appear in the Reviews Activity Panel
Described [here](https://github.com/woocommerce/woocommerce-admin/pull/1860#pullrequestreview-217009429) by @psealock and @joshuatf.

> I see the unread indicator appear on comments that are pre-approved or once they are approved, but not ones that are awaiting approval. It seems to be a good thing to tackle in a follow up PR.

> Noted the same issue @psealock did with unapproved reviews, which we'll definitely want to follow up on since those are probably more "actionable."

Aljullu:
Fixes #1868.

In addition to approved reviews, load also on hold reviews in the Activity Panel.

### Screenshots
image

Detailed test instructions:
- Go to _Settings_ > _Discussion_ and check _Comment must be manually approved_.
- Create a product review.
- Go to any WooCommerce Admin page and verify an unread indicator (red dot) appears next to the _Reviews_ tab of the Activity Panel.
- Click on the tab and verify the new review appears and is marked with a red dot.

### DIFF
--- a/client/header/activity-panel/index.js
+++ b/client/header/activity-panel/index.js
@@ -15,6 +15,7 @@ import { partial, uniqueId, find } from 'lodash';
  */
 import './style.scss';
 import ActivityPanelToggleBubble from './toggle-bubble';
+import { DEFAULT_REVIEW_STATUSES } from 'wc-api/constants';
 import { H, Section } from '@woocommerce/components';
 import InboxPanel from './panels/inbox';
 import OrdersPanel from './panels/orders';
@@ -326,6 +327,7 @@ export default withSelect( select => {
 			orderby: 'date_gmt',
 			page: 1,
 			per_page: 1,
+			status: DEFAULT_REVIEW_STATUSES,
 		};
 		const reviews = getReviews( reviewsQuery );
 		const totalReviews = getReviewsTotalCount( reviewsQuery );
--- a/client/header/activity-panel/index.js
+++ b/client/header/activity-panel/index.js
@@ -340,6 +340,23 @@ export default withSelect( select => {
 				new Date( reviews[ 0 ].date_created_gmt + 'Z' ).getTime() >
 					userData.activity_panel_reviews_last_read;
 		}
+
+		if ( ! unreadReviews && '1' === wcSettings.commentModeration ) {
+			const actionableReviewsQuery = {
+				page: 1,
+				// @todo we are not using this review, so when the endpoint supports it,
+				// it could be replaced with `per_page: 0`
+				per_page: 1,
+				status: 'hold',
+			};
+			const totalActionableReviews = getReviewsTotalCount( actionableReviewsQuery );
+			const isActionableReviewsError = Boolean( getReviewsError( actionableReviewsQuery ) );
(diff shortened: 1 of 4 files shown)


jeffstieler (commented):
Code looks good, and tests well. I just want to make sure that it's expected that the indicator disappears from the "on hold" review here:



If that's expected, consider this approved!

Aljullu:
> I just want to make sure that it's expected that the indicator disappears from the "on hold" review here:

That's a good point. @timmyc and @LevinMedia what are your thoughts here? Should the _Reviews_ panel mark reviews as unread only when they need to be approved, only when they are unread, or they should be marked as unread in both cases?

LevinMedia:
@Aljullu I'm thinking both

Aljullu:
> @Aljullu I'm thinking both

In that case, reviews that are not approved should be sorted on top of approved reviews, even when they are older, right?

Aljullu:
@jeffstieler I updated this PR so 'on hold' reviews are always marked as unread. The only issue is we need to perform two different API requests in order to detect if 1) there are new reviews 2) there are 'on hold' reviews. I think we could consider creating a specific endpoint for that so only one request is needed, but that could be implemented in a follow-up not blocking this one.

Would you like to take a second look?

LevinMedia:
>@LevinMedia in that case, reviews that are not approved should be sorted on top of approved reviews, even when they are older? Or should they be sorted by date anyway?

for now lets keep them sorted by date. @jameskoster and I were discussing giving users the option to choose between showing all reviews, and only reviews that require specific action, but we'd want to handle that in a different issue / PR if we go that route.

jeffstieler:
> The only issue is we need to perform two different API requests in order to detect if 1) there are new reviews 2) there are 'on hold' reviews. I think we could consider creating a specific endpoint for that so only one request is needed, but that could be implemented in a follow-up not blocking this one.

Rather than introduce another query, what do you think about just issuing the same query as the Reviews panel to get the unread status? Sure, it's more data than needed if the user doesn't open the panel, but it actually saves a request if they do.

See it in action: https://gist.github.com/jeffstieler/30c8b87016492e6f76f26a206ed51fcb

Aljullu:
> Rather than introduce another query, what do you think about just issuing the same query as the Reviews panel to get the unread status? Sure, it's more data than needed if the user doesn't open the panel, but it actually saves a request if they do.

I like your approach but if I understand it correctly, it would not display the unread circle if the 'on hold' review is not in the last 25, right? Do you think that's an edge case we can ignore?

jeffstieler:
> it would not display the unread circle if the 'on hold' review is not in the last 25, right?

Ah - good catch. You're right about that.

> Do you think that's an edge case we can ignore?

Probably not - disregard my suggestion. I'll re-review.

Aljullu:
Hey @jeffstieler, did you have time to take a second look at this? :point_up:

jeffstieler (approved):
Looks good - 🛳it!

PR #1872, opened 2019-03-21
Add order milestone notifications.

### LINKED ISSUE #1831: Notes: Order Milestone Notifications
Somewhat related to, but not totally blocked by #1830 

To build upon the new inbox in WooCommerce Admin, we have set up a list of [new inbox notifs to build out](https://docs.google.com/spreadsheets/d/1iBcjWbQtN1DNZmvCmB4zmgGxiTL99_cJgeC2Ew2FvJE/edit#gid=0) ( Private Spreadsheet ). For this enhancement, we will be implementing a number of "Order Milestone" notifications, which fall into three different notifications as defined in the spreadsheet linked above.

Since they are all based upon the number of orders processed, it seems like they should be able to live within the same logic - so I've grouped them together into this one issue.

- First Order
- 10 orders
- 100, 250, 500, 1k, 5k, 10k, 50k, 100k orders

The message titles, copy, and associated actions are all listed in the [spreadsheet](https://docs.google.com/spreadsheets/d/1iBcjWbQtN1DNZmvCmB4zmgGxiTL99_cJgeC2Ew2FvJE/edit#gid=0).

jeffstieler:
Fixes #1831, #2006.

**Functionality Introduced:**

Introduces a new class to manage order milestone notifications.

Milestones:
* First order
* Ten orders
* 100, 250, 500, 1000, 5000, 10000, 500000, 1000000

The first and ten order milestones are triggered immediately as orders are created.

Subsequent milestones are processed hourly. Only one notification is visible at any one time.

Upon plugin activation, the orders count is cached to avoid celebrating milestones passed before activation.

**Changes made to existing code:**

* `WC_Admin_Note`'s default value for `content_data` broke note retrieval - fixed in db97d51
* Inbox item dates weren't being handled properly - notes created were showing future dates - fixed in 1a3aa53

### Screenshots

Screen Shot 2019-03-20 at 3 53 27 PM

Screen Shot 2019-03-21 at 12 25 26 PM

- Starting on a fresh store with no orders, place an order
- Verify the "first order" note is in the inbox
- Verify CTA goes to https://docs.woocommerce.com/document/managing-orders/
- Place 10 orders
- Verify the "ten orders" note is in the inbox
- Verify CTA goes to https://woocommerce.com/success-stories/
- Place 100 orders (smooth generator can really help here)
- Trigger the `wc_admin_process_orders_milestone` cron event to run (use Crontrol to delete job)
- Verify the "100 orders" note is in the inbox
- Verify CTA goes to Orders Report
- Delete all milestone notes from `wc_admin_notes` table
- Delete `wc_admin_version` and `wc_admin_last_orders_milestone` options
- Load an admin page
- Trigger the `wc_admin_process_orders_milestone` cron event to run (use Crontrol to delete job)
- Verify no new order milestone notes in inbox

### DIFF
--- a/client/header/activity-panel/activity-card/index.js
+++ b/client/header/activity-panel/activity-card/index.js
@@ -30,7 +30,9 @@ class ActivityCard extends Component {
 					<H className="woocommerce-activity-card__title">{ title }</H>
 					{ subtitle && <div className="woocommerce-activity-card__subtitle">{ subtitle }</div> }
 					{ date && (
-						<span className="woocommerce-activity-card__date">{ moment( date ).fromNow() }</span>
+						<span className="woocommerce-activity-card__date">
+							{ moment.utc( date ).fromNow() }
+						</span>
 					) }
 				</header>
 				<Section className="woocommerce-activity-card__body">{ children }</Section>
--- a/includes/notes/class-wc-admin-notes-order-milestones.php
+++ b/includes/notes/class-wc-admin-notes-order-milestones.php
@@ -19,6 +19,11 @@ class WC_Admin_Notes_Order_Milestones {
 	const FIRST_ORDER_NOTE_NAME = 'wc-admin-first-order';
 
 	/**
+	 * Name of the "ten orders" note.
+	 */
+	const TEN_ORDERS_NOTE_NAME = 'wc-admin-ten-orders';
+
+	/**
 	 * Allowed order statuses for calculating milestones.
 	 *
 	 * @var array
@@ -30,10 +35,19 @@ class WC_Admin_Notes_Order_Milestones {
 	);
 
 	/**
+	 * Orders count cache.
+	 *
+	 * @var int
+	 */
+	protected $orders_count = null;
+
+	/**
 	 * Hook everything up.
 	 */
 	public function __construct() {
-		add_action( 'woocommerce_new_order', array( $this, 'add_first_order_note' ) );
+		if ( 10 <= $this->get_orders_count() ) {
+			add_action( 'woocommerce_new_order', array( $this, 'first_two_milestones' ) );
+		}
 	}
 
 	/**
@@ -42,18 +56,20 @@ class WC_Admin_Notes_Order_Milestones {
 	 * @return int Total orders count.
 	 */
 	public function get_orders_count() {
-		$status_counts = array_map( 'wc_orders_count', $this->allowed_statuses );
-		$orders_count  = array_sum( $status_counts );
+		if ( is_null( $this->orders_count ) ) {
+			$status_counts       = array_map( 'wc_orders_count', $this->allowed_statuses );
+			$this->orders_count  = array_sum( $status_counts );
+		}
 
-		return $orders_count;
+		return $this->orders_count;
 	}
 
 	/**
-	 * Add a milestone note for the store's first order.
+	 * Add a milestone notes for the store's first order and first 10 orders.
 	 *
 	 * @param int $order_id WC_Order ID.
 	 */
-	public function add_first_order_note( $order_id ) {
+	public function first_two_milestones( $order_id ) {
 		$order = wc_get_order( $order_id );
 
 		// Make sure this is the first pending/processing/completed order.
@@ -75,6 +91,21 @@ class WC_Admin_Notes_Order_Milestones {
 			$note->set_name( self::FIRST_ORDER_NOTE_NAME );
 			$note->set_source( 'woocommerce-admin' );
 			$note->add_action( 'learn-more', __( 'Learn more', 'woocommerce-admin' ), 'https://docs.woocommerce.com/document/managing-orders/' );
+			$note->save();
+		}
+
+		if ( 10 === $orders_count ) {
+			// Add the first ten orders note.
+			$note = new WC_Admin_Note();
+			$note->set_title( __( 'Congratulations on processing 10 orders!', 'woocommerce-admin' ) );
+			$note->set_content(
+				__( "You've hit the 10 orders milestone! Look at you go. Browse some WooCommerce success stories for inspiration.", 'woocommerce-admin' )
+			);
+			$note->set_type( WC_Admin_Note::E_WC_ADMIN_NOTE_INFORMATIONAL );
+			$note->set_icon( 'trophy' );
+			$note->set_name( self::TEN_ORDERS_NOTE_NAME );
+			$note->set_source( 'woocommerce-admin' );
+			$note->add_action( 'browse', __( 'Browse', 'woocommerce-admin' ), 'https://woocommerce.com/success-stories/' );
 			$note->save();
 		}
 	}
--- a/includes/notes/class-wc-admin-notes-order-milestones.php
+++ b/includes/notes/class-wc-admin-notes-order-milestones.php
@@ -34,6 +34,11 @@ class WC_Admin_Notes_Order_Milestones {
 	const LAST_ORDER_MILESTONE_OPTION_KEY = 'wc_admin_last_orders_milestone';
 
 	/**
+	 * Hook to process order milestones.
+	 */
+	const PROCESS_ORDERS_MILESTONE_HOOK = 'wc_admin_process_orders_milestone';
+
+	/**
 	 * Allowed order statuses for calculating milestones.
 	 *
 	 * @var array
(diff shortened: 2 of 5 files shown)


includes/class-wc-admin-notes-order-milestones.php
```diff
@@ -0,0 +1,251 @@
...
+	 * Hook to process order milestones.
+	 */
+	const PROCESS_ORDERS_MILESTONE_HOOK = 'wc_admin_process_orders_milestone';
+
+	/**
+	 * Allowed order statuses for calculating milestones.
+	 *
+	 * @var array
+	 */
+	protected $allowed_statuses = array(
```
timmyc: I feel like we might have chatted about this in slack, but did we want to perhaps make this list filter-able? Perhaps I get too filter happy, but it seems like there are numerous extensions that allow manipulation of order statuses.

includes/class-wc-admin-notes-order-milestones.php
```diff
@@ -0,0 +1,251 @@
...
+		$order = wc_get_order( $order_id );
+
+		// Make sure this is the first pending/processing/completed order.
+		if ( ! in_array( $order->get_status(), $this->allowed_statuses ) ) {
+			return;
+		}
+
+		$orders_count = $this->get_orders_count();
+
+		if ( 1 === $orders_count ) {
```
timmyc: I'm not seeing where `set_last_milestone()` is called for these two instances of first and tenth order?

includes/class-wc-admin-notes-order-milestones.php
```diff
@@ -0,0 +1,251 @@
...
+			$note->set_type( WC_Admin_Note::E_WC_ADMIN_NOTE_INFORMATIONAL );
+			$note->set_icon( 'trophy' );
+			$note->set_name( self::FIRST_ORDER_NOTE_NAME );
+			$note->set_source( 'woocommerce-admin' );
+			$note->add_action( 'learn-more', __( 'Learn more', 'woocommerce-admin' ), 'https://docs.woocommerce.com/document/managing-orders/' );
+			$note->save();
+		}
+
+		if ( 10 === $orders_count ) {
+			// Add the first ten orders note.
```
timmyc: Should the first order milestone be deleted here too, akin to what is done on the other milestones?

`WC_Admin_Notes::delete_notes_with_name( self::ORDERS_MILESTONE_NOTE_NAME );`

includes/class-wc-admin-notes-order-milestones.php
```diff
@@ -0,0 +1,251 @@
...
+			return;
+		}
+
+		$this->set_last_milestone( $current_milestone );
+
+		// We only want one milestone note at any time.
+		WC_Admin_Notes::delete_notes_with_name( self::ORDERS_MILESTONE_NOTE_NAME );
+
+		// Add the milestone note.
+		$note = new WC_Admin_Note();
```
timmyc: All the new note creation in the file here is a bit repetitive - did you consider extracting it into a method on the class here? Or any particular reason not to?

timmyc (commented):
Couple of quick questions...

includes/class-wc-admin-notes-order-milestones.php
jeffstieler: That's a good point about extensions - I'll make this filterable.

includes/class-wc-admin-notes-order-milestones.php
jeffstieler: I'm not sure. Let's ask @MarinaWoo - if a store hits the 10 order milestone (or even the 100) quickly, should only one milestone alert display, or should the first order and 10 order milestones be in the inbox alongside each other?

includes/class-wc-admin-notes-order-milestones.php
jeffstieler: The only things repeated are the type, icon, and source so I didn't see the repetition as that big of a problem.

includes/class-wc-admin-notes-order-milestones.php
timmyc: So the only reason I kind of let my brain mentally wander down that path was a scenario where I get my first order... and what if that is me testing out my site ( which we went around in circles on so don't want to go there too much again ) and then I get the notif, then delete that order, and a new order comes in. Edge case I suppose, but wouldn't that trigger the notif again for the first order? Same for a 10th order getting placed and one being deleted.

includes/class-wc-admin-notes-order-milestones.php
MarinaWoo: It is all so very relative, it is hard to know! I think if someone is tinkering and testing they would be aware that they are 'forcing' the milestones, and therefore if they showed up fast it would be okay. I do think 'deleted' orders should (if possible) be included in the count towards milestones so as to avoid the scenario @timmyc is talking about.

includes/class-wc-admin-notes-order-milestones.php
jeffstieler: Yes in your example the notification would get triggered again if you deleted your test order. But.. wouldn't that be a good thing?

includes/class-wc-admin-notes-order-milestones.php
jeffstieler: As it currently sits, deleted orders would only reset the notifications for the first order and the tenth order. It wouldn't rollback any of the 100+ milestones.

includes/class-wc-admin-notes-order-milestones.php
jeffstieler: Added a filter in 90a1c346.

timmyc:
@jeffstieler can we get a rebase here and I'll take it for a final test?

jeffstieler:
@timmyc rebased and ready for another test.

timmyc:
@jeffstieler I can't seem to get this to work 🤦‍♂️ - so I installed on a site with existing orders, but less than 100. No notifs shown, which is to be expected. I then installed on a new Woo install, placed a test order, and I don't see a notif. Tried manually running the cronjob via crontrol... no luck.

ronakganatra9:
I agree with @timmyc , I have also tested it with existing store as well as placed a new order but it does not send notification properly.

ronakganatra9:
Hello Team,

I have created an addon plugin Customer milestone plugin same like an order milestone.
Here is a GitHub repository link you can provide your thoughts and suggestions on it.

Link: https://github.com/ronakganatra9/customer-milestone-notifier-woocommerce-admin

Thanks & Regards,
Ronak Ganatra

jeffstieler:
@timmyc @ronakganatra9 there were a few bugs - this is ready to be tested again.

includes/notes/class-wc-admin-notes-order-milestones.php
```diff
@@ -0,0 +1,262 @@
...
+
+		add_action( 'woocommerce_after_register_post_type', array( $this, 'init' ) );
+	}
+
+	/**
+	 * Hook everything up.
+	 */
+	public function init() {
+		if ( ! wp_next_scheduled( self::PROCESS_ORDERS_MILESTONE_HOOK ) ) {
+			wp_schedule_event( time(), 'hourly', self::PROCESS_ORDERS_MILESTONE_HOOK );
```
Aljullu: Dummy question here, but shouldn't we add a [`wp_clear_scheduled_hook()`](https://codex.wordpress.org/Function_Reference/wp_clear_scheduled_hook) somewhere to clear this event when the plugin is deactivated?

client/header/activity-panel/activity-card/index.js
```diff
@@ -30,7 +30,9 @@ class ActivityCard extends Component {
 					<H className="woocommerce-activity-card__title">{ title }</H>
 					{ subtitle && <div className="woocommerce-activity-card__subtitle">{ subtitle }</div> }
 					{ date && (
-						<span className="woocommerce-activity-card__date">{ moment( date ).fromNow() }</span>
+						<span className="woocommerce-activity-card__date">
+							{ moment.utc( date ).fromNow() }
```
Aljullu: Thanks for this fix! :clap:

Aljullu (approved):
This is looking really great @jeffstieler! Thanks for working on this. I confirm the issue with less than 10 orders is fixed, so I will approve this PR. If we find any other issue I think we can work on it in a follow-up, so we don't block this one.

Not for this PR, but from a UX point of view, I think these notifications should be dismissable in some way so they don't stay forever in the _Inbox_, what do you think?
image

includes/notes/class-wc-admin-notes-order-milestones.php
jeffstieler: Good catch - added in 37db77b.

timmyc:
Super great to see this merged in @jeffstieler - thanks for the review follow-up @Aljullu

Issue #1847, opened 2019-03-19
Category Report Data Discrepancies


LevinMedia:
**Describe the bug**
The summary numbers below the table aren't affected by the query when comparing categories, resulting in a discrepancy with the other totals on the page. 

**To Reproduce**

1. Go to the category report
2. Compare a number of categories
3. Note the summary numbers at the bottom of the table do not match with other totals on the page. 

**Expected behavior**
The numbers should match.  

**Screenshots**

![](https://slack-imgs.com/?c=1&url=https%3A%2F%2Fd1c0hjomoutdrw.cloudfront.net%2Fitems%2F0Y162Y3e2p0Y01033X3U%2F%255B3a021c3b8bb7e5856380d5b5c7932140%255D_Screen%2BShot%2B2019-03-19%2Bat%2B2.48.50%2BPM.png)

psealock:
Closed via #1873

PR #1873, opened 2019-03-21
Refactor getFilterQuery by explicitly passing down configs

psealock:
### Problem

Configs have been mapped to appropriate files based on `endpoint` values to derive appropriate query objects. This worked well until Variations and Categories used the reports endpoint (`/reports/products`) while still having their own configs. 

I previously used mapping in various util functions to overcome this issue, which caused confusion and made changing things difficult, leading to the regression found in https://github.com/woocommerce/woocommerce-admin/issues/1847.

### Solution

Explicitly pass down `filters` and `advancedFilters` from report index files. This allows independent control over `endpoint` and relevant filter configs by decoupling the association.

This should make for easier debugging because inputs aren't being manipulated halfway through a process.

1. Go to the category report
2. Compare a number of categories
3. Note the summary numbers at the bottom of the table match with other totals on the page.
4. Poke around other reports to ensure no regression has been introduced.

### DIFF
--- a/client/analytics/components/leaderboard/index.js
+++ b/client/analytics/components/leaderboard/index.js
@@ -99,7 +99,12 @@ Leaderboard.propTypes = {
 export default compose(
 	withSelect( ( select, props ) => {
 		const { endpoint, tableQuery, query } = props;
-		const tableData = getReportTableData( endpoint, query, select, tableQuery );
+		const tableData = getReportTableData( {
+			endpoint,
+			query,
+			select,
+			tableQuery,
+		} );
 
 		return { ...tableData };
 	} )
--- a/client/analytics/components/report-chart/index.js
+++ b/client/analytics/components/report-chart/index.js
@@ -252,7 +252,7 @@ ReportChart.defaultProps = {
 
 export default compose(
 	withSelect( ( select, props ) => {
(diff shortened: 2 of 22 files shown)


Aljullu (approved):
That's an awesome refactor @psealock, it improves code readability a lot, good job! :clap: 

I have been testing it and everything looks fine but I would say it doesn't fix #1847. At least, not for me:
image

I have been investigating what's going on here and I think it's a Rest API issue. In my case, it looks like the table summary is adding the numbers of `Accessories`, `Clothing`, `Thsirts` and `Hoodies`. Given that `Tshirts` and `Hoodies` are subcategories of `Clothing`, those items are counted twice. I think that's the discrepancies we are seeing.

---

Query:
image

Returned data:
image

---

I would say we should proceed merging this PR but keeping #1847 open.

psealock:
Thanks for the review @Aljullu and pointing this one out. I can see the `segment_label: null` segments in the response, but their numbers aren't included in the totals for me. I'll see if I can reproduce and follow up in the next PR.

psealock:
>Given that Tshirts and Hoodies are subcategories of Clothing, those items are counted twice.

I can reproduce now. Nice catch and great observation. I think this is a separate issue, so I will close the original and open a new one.

PR #1875, opened 2019-03-21
Swap order of charts and leaderboards on the dashboard

LevinMedia:
Per the original designs - this swaps the order of charts and leaderboards section in the dashboard - Charts come first now.


psealock (approved):
LGTM

PR #1877, opened 2019-03-22
Automatically update versions on build

### LINKED ISSUE #1876: Version compare should use plugin version number
To avoid the need to find/replace version numbers across many files, we should grab the plugin version from the plugin file on install. 

If all version numbers aren't updated before a build, certain updates to the plugin may not run (e.g., database changes).  Not really a bug, but will make build less prone to error.

**To Reproduce**

1. Bump the version number in `woocommerce-admin.php`
2. Make a change to a table in `WC_Admin_Install`.
3. Visit the dashboard.
4. Note the database changes aren't made.

joshuatf:
Fixes #1876

* Updates versions automatically so we don't need to update in multiple places.
* Uses the `package.json` version as the source of truth and updates the version in the plugin meta data to match this version.
* Checks the plugin version in the install script to compare against the existing installed version.

1. Change the version number in `package.json`.
2. Make a temporary database change in `WC_Admin_Install` (to later note that database updates are made on version change).
3. Run `npm start` or `npm build`.
4. Note the version is updated inside `woocommerce-admin.php`.
5. Note the database updates were ran on visiting the dashboard.

### Question

I don't think the `get_plugin_data()` is a performance intensive task, however, we could also just write the version as a constant from the build script since we already have the file open to update the plugin version.  Thoughts?


psealock (approved):
LGTM and is working.

>I don't think the get_plugin_data() is a performance intensive task, however, we could also just write the version as a constant from the build script since we already have the file open to update the plugin version. Thoughts?

I'm not familiar with `get_plugin_data` offhand, but it does make sense to me to have one mechanism for writing and gathering the version number. So, yes I think it would be slightly better to write the version from the build script in both places.

joshuatf:
Thanks for the review @psealock.  Updated this to use a constant instead.

timmyc:
Awesome.

PR #1878, opened 2019-03-22
Refactor notes controller and add ordering params

### LINKED ISSUE #1865: Notes should allow ordering and default to date
**Describe the bug**
Currently the notes are sorted by ID.  This is mostly okay if notes are created in sequential order, but may cause issues if dates are retroactively changed or imported.

We'll also want to check the `latestNote` used to determine unread status to explicitly use date ordering.

**To Reproduce**

1. Open inbox notes.
2. Note the order.
3. Change the date created on a note.
4. Note the order in inbox remains unchanged.

**Expected behavior**
Expected notes chronologically.

joshuatf:
Fixes #1865 

* Refactors collection params in notes controller
* Adds `order` and `orderby` params to the controller and data store
* Sets the default order for inbox and unread notes

1.  Make requests to the notes REST endpoint ``
2. Test with the `order` and `orderby` params.  Order by:
- note_id
- date
- type
- title
- status
3. Check that inbox is still working and ordered by date in descending order.
4. Read notes to mark the last read time.
5. Add a note that is not of type `info` or `warning`.
6. Make sure the unread indicator does not activate for that note.
7. Try making a request to the notes endpoint with an invalid `orderby` param to verify that the invalid error message shows.

### DIFF
--- a/includes/api/class-wc-admin-rest-admin-notes-controller.php
+++ b/includes/api/class-wc-admin-rest-admin-notes-controller.php
@@ -121,7 +121,7 @@ class WC_Admin_REST_Admin_Notes_Controller extends WC_REST_CRUD_Controller {
 		}
 
 		$response = rest_ensure_response( $data );
-		$response->header( 'X-WP-Total', WC_Admin_Notes::get_notes_count( $type, $status ) );
+		$response->header( 'X-WP-Total', WC_Admin_Notes::get_notes_count( $query_args['type'], $query_args['status'] ) );
 
 		return $response;
 	}
@@ -136,6 +136,8 @@ class WC_Admin_REST_Admin_Notes_Controller extends WC_REST_CRUD_Controller {
 		$args            = array();
 		$args['order']   = $request['order'];
 		$args['orderby'] = $request['orderby'];
+		$args['type']    = isset( $request['type'] ) ? $request['type'] : array();
+		$args['status']  = isset( $request['status'] ) ? $request['status'] : array();
 
 		if ( 'date' === $args['orderby'] ) {
 			$args['orderby'] = 'date_created';
@@ -149,18 +151,6 @@ class WC_Admin_REST_Admin_Notes_Controller extends WC_REST_CRUD_Controller {
 		$args['page'] = isset( $request['page'] ) ? intval( $request['page'] ) : 1;
 		if ( $args['page'] <= 0 ) {
 			$args['page'] = 1;
-		}
-
-		$type = isset( $request['type'] ) ? $request['type'] : '';
-		$type = sanitize_text_field( $type );
-		if ( ! empty( $type ) ) {
-			$args['type'] = $type;
-		}
-
-		$status = isset( $request['status'] ) ? $request['status'] : '';
-		$status = sanitize_text_field( $status );
-		if ( ! empty( $status ) ) {
-			$args['status'] = $status;
 		}
 
 		/**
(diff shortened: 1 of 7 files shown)


includes/api/class-wc-admin-rest-admin-notes-controller.php
```diff
@@ -300,19 +317,64 @@ public function prepare_item_for_response( $data, $request ) {
...
-		$params['type']    = array(
-			'description'       => __( 'Type of note.', 'woocommerce-admin' ),
+		$params             = array();
+		$params['context']  = $this->get_context_param( array( 'default' => 'view' ) );
+		$params['order']    = array(
+			'description'       => __( 'Order sort attribute ascending or descending.', 'woocommerce-admin' ),
 			'type'              => 'string',
-			'enum'              => WC_Admin_Note::get_allowed_types(),
+			'default'           => 'desc',
+			'enum'              => array( 'asc', 'desc' ),
```
timmyc: for these string based args, with pre-defined options in an array - in a recent PR in Woo core @claudiosanches had me add in `'sanitize_callback' => 'sanitize_key'` i.e. https://github.com/woocommerce/woocommerce/blob/master/includes/api/wc-blocks/class-wc-rest-blocks-products-controller.php#L326

Which if I followed the trail properly would allow someone passing in `ASC` to be downcased ( but do correct me if I'm wrong Claudio! )

includes/data-stores/class-wc-admin-notes-data-store.php
```diff
@@ -250,6 +250,9 @@ private function save_actions( &$note ) {
 	public function get_notes( $args = array() ) {
 		global $wpdb;
 
+		$order   = isset( $args['order'] ) ? $args['order'] : 'DESC';
```
timmyc: For usage outside the context of a REST request, is there a need to sanitize/validate the args here?

timmyc (commented):
Excellent refactor here Josh. Couple of questions.

includes/api/class-wc-admin-rest-admin-notes-controller.php
joshuatf: Hmm, I tried this with `sanitize_key` and it didn't work with any uppercase characters.  It didn't prevent anything from working here either so I'm not opposed to replacing this if there are added benefits. /cc @claudiosanches

includes/data-stores/class-wc-admin-notes-data-store.php
joshuatf: Good catch!  I don't think so.  I've also removed the `$page` and `$per_page` alterations that occurred here.

I can't find a place where they might have been used, but it looks like they were dealing with negative numbers, so probably passing `-1` to get all records.  Since the `apiFetch` method already handles this under the hood, I think we can safely remove, but I'll keep an eye on it and open a follow-up issue if this creates any problems.

joshuatf: On second look, it does look like the data store is accessed directly for some functions like `WC_Admin_Notes::unsnooze_notes()`.  I've replaced this section with a default arguments pattern like we use in other data stores.

Aljullu (approved):
Activity Panel _Inbox_ is working well, I also tried several queries to the endpoint with different parameters and couldn't find any issue.

includes/api/class-wc-admin-rest-admin-notes-controller.php
claudiosanches: In fact there's no need, since should only expect results like listed in `enum`.

PR #1880, opened 2019-03-22
Activity Panel: add overlay to product placeholder images

### LINKED ISSUE #1067: Product review placeholder image
**Describe the bug**
The placeholder image looks odd with the white background when an image is not set.  Almost looks like 2 misaligned images.

**To Reproduce**
1.  Add a product with no image.
2.  Open the "Reviews" tab from the wc-admin header.
3.  Note the placeholder image with the adjacent user avatar.

**Expected behavior**
Expected a border, slight gray background, or different placeholder image.  @LevinMedia what do you think?  Does this look okay to you or are changes warranted?

**Screenshots**
screen shot 2018-12-11 at 3 07 14 pm

Aljullu:
Fixes #1067.

Adds an overlay to placeholder product images in the _Reviews_ panel of the Activity Panel.

### Screenshots
image

- Create a review in a product without an image and in a product with an image.
- Go to the _Reviews_ panel of the Activity Panel.
- Verify there is an overlay in the product placeholder image but it's not in the product with an image.

### DIFF
--- a/client/header/activity-panel/activity-card/style.scss
+++ b/client/header/activity-panel/activity-card/style.scss
@@ -243,6 +243,26 @@
 			position: relative;
 			top: -42px;
 			border: 2px solid $white;
+			z-index: 2;
+		}
+
+		.woocommerce-review-activity-card__image-overlay__product {
+			display: inline-block;
+			height: 60px;
+			position: relative;
+			width: 60px;
+
+			&.is-placeholder::before {
+				background: $core-grey-dark-500;
+				border-radius: 50%;
+				content: '';
+				position: absolute;
+				left: 0;
+				right: 0;
+				bottom: 0;
+				top: 0;
+				opacity: 0.1;
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


LevinMedia (approved):
The place holder image is looking good!

Not in the scope of this PR, but noting a couple other issues I encountered though:

Prior to leaving my first review, I had a "0" showing up next to the label in the activity panel:

Screen Shot 2019-03-22 at 8 16 13 AM


I'm also getting an "Invalid date" message for the review timestamp

Screen Shot 2019-03-22 at 8 23 12 AM

Also you'll see in the screen shot above, "Verified customer" only showed up on the second review.

Aljullu:
Thanks for the review @LevinMedia!

I created issues for the bugs you reported: #1889, #1890 and #1891 and assigned them to me. Some are a one-line fix, but since they will cause conflicts with other open PRs, I will wait to fix them until next week.

psealock (approved):
LGTM!

One small thing: I like to use `background-color` instead of just `background` in CSS because `background` is much more, https://developer.mozilla.org/en-US/docs/Web/CSS/background.

Aljullu:
> One small thing: I like to use `background-color` instead of just `background` in CSS because `background` is much more, https://developer.mozilla.org/en-US/docs/Web/CSS/background.

Agree! Fixed in 2e68709a763577fd3e3ba80937b72d92beb492fa. Thanks for taking a look, @psealock.

PR #1907, opened 2019-03-22
Support Sequential Order Numbers in Downloads Report order number filter

### LINKED ISSUE #1615: Downloads Report: Order number search mixes id/number fields when matching
**Describe the bug**

**This only occurs when using an extension that modifies order number like Sequential Order Numbers Pro.**

The "order number" advanced filter seems to match on the Order ID, not number, but highlights the match in the order number ... displaying the number in the results list, but then showing the ID when selected. It's not working well... 😅 

Found testing #1523.

**To Reproduce**

1. Go to Downloads Report
2. Add advanced filter "order numer"
3. Type a single digit that will match an Order ID and also be found in the Order Number (I used `3`)
4. Verify that the Order Number is displayed in the results
5. Click an Order Number
6. Verify that the Order ID is displayed in the token

**Expected behavior**

I expect this to compare/display values consistently. I think it should be on the Number.

**Screenshots**

**Additional context**

Using WCCOM test data.

jeffstieler:
Fixes #1615.

This PR seeks to add support for Sequential Order Numbers when filtering the Downloads Report by order number.

It also introduces `order` and `limit` to the number filter query to avoid problems when there are many orders.

### Screenshots

Screen Shot 2019-03-22 at 2 32 41 PM

_Note: It's easiest to test on the WCCOM data as it already has data populated by Sequential Order Numbers Pro._

- Without Sequential Order Numbers (or Pro) enabled
- Go to Downloads Report
- Add advanced filter "order number"
- Type a number that will match the beginning of an order ID (I used `3`)
- Verify that the matches are order IDs
- Verify the filter works
- Enable Sequential Order Numbers (or Pro)
- Go to Downloads Report
- Add advanced filter "order number"
- Type a number that will match the beginning of a sequential order number (I used `5`)
- Verify that the matches are order numbers
- Verify the filter works

Optionally test with the non-pro/free version of Sequential Order Numbers.

### DIFF
--- a/includes/api/class-wc-admin-rest-orders-controller.php
+++ b/includes/api/class-wc-admin-rest-orders-controller.php
@@ -30,6 +30,7 @@ class WC_Admin_REST_Orders_Controller extends WC_REST_Orders_Controller {
 	 */
 	public function get_collection_params() {
 		$params           = parent::get_collection_params();
+		// This needs to remain a string to support extensions that filter Order Number.
 		$params['number'] = array(
 			'description'       => __( 'Limit result set to orders matching part of an order number.', 'woocommerce-admin' ),
 			'type'              => 'string',
--- a/includes/api/class-wc-admin-rest-orders-controller.php
+++ b/includes/api/class-wc-admin-rest-orders-controller.php
@@ -32,8 +32,7 @@ class WC_Admin_REST_Orders_Controller extends WC_REST_Orders_Controller {
 		$params           = parent::get_collection_params();
 		$params['number'] = array(
 			'description'       => __( 'Limit result set to orders matching part of an order number.', 'woocommerce-admin' ),
-			'type'              => 'integer',
-			'sanitize_callback' => 'absint',
+			'type'              => 'string',
 			'validate_callback' => 'rest_validate_request_arg',
 		);
 		return $params;
@@ -51,15 +50,22 @@ class WC_Admin_REST_Orders_Controller extends WC_REST_Orders_Controller {
 
 		// Search by partial order number.
 		if ( ! empty( $request['number'] ) ) {
-			$order_ids = $wpdb->get_col(
+			$partial_number = trim( $request['number'] );
+			$limit          = intval( $args['posts_per_page'] );
+			$order_ids      = $wpdb->get_col(
 				$wpdb->prepare(
-					"SELECT ID FROM {$wpdb->prefix}posts WHERE post_type = 'shop_order' AND ID LIKE %s",
-					intval( $request['number'] ) . '%'
+					"SELECT ID
+					FROM {$wpdb->prefix}posts
+					WHERE post_type = 'shop_order'
+					AND ID LIKE %s
+					LIMIT %d",
+					$wpdb->esc_like( absint( $partial_number ) ) . '%',
+					$limit
 				)
 			);
 
 			// Force WP_Query return empty if don't found any order.
-			$order_ids        = ! empty( $order_ids ) ? $order_ids : array( 0 );
+			$order_ids        = empty( $order_ids ) ? array( 0 ) : $order_ids;
 			$args['post__in'] = $order_ids;
 		}
 
--- a/packages/components/src/search/autocompleters/orders.js
+++ b/packages/components/src/search/autocompleters/orders.js
@@ -3,7 +3,6 @@
  * External dependencies
  */
 import apiFetch from '@wordpress/api-fetch';
-import { isNaN } from 'lodash';
 
 /**
  * WooCommerce dependencies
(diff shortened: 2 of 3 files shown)


includes/api/class-wc-admin-rest-orders-controller.php
```diff
@@ -51,15 +50,59 @@ protected function prepare_objects_query( $request ) {
...
-					"SELECT ID FROM {$wpdb->prefix}posts WHERE post_type = 'shop_order' AND ID LIKE %s",
-					intval( $request['number'] ) . '%'
-				)
-			);
+			$partial_number = trim( $request['number'] );
+			$order_sql      = esc_sql( $args['order'] );
+			$limit          = intval( $args['posts_per_page'] );
+
+			// Handle Sequential Order Numbers.
+			if ( class_exists( 'WC_Seq_Order_Number_Pro', false ) ) {
```
timmyc: In-lieu of having conditional logic for particular extensions - don't you think this could be fixed on a per-extension basis in their own code base by hooking into REST response filters accordingly?

jeffstieler: I think we'd need to add some hooks to allow for extensions to easily modify the handling of the `number` parameter. _(I was going to suggest `woocommerce_rest_orders_prepare_object_query`, but that gets applied before `WC_Admin_REST_Orders_Controller` modifies `post__in`)_

Will code doing so be allowed into the Sequential Order Number extensions and are we OK with the wonky behavior for stores that aren't on the latest version of those extensions?

timmyc: I _think_ it is already filterable [here](https://github.com/woocommerce/woocommerce/blob/89740f9583893581a591651c757b6b9085cf6eb1/includes/api/v2/class-wc-rest-orders-v2-controller.php#L313) but at times I do get lost tracing the class chains in the core rest endpoints. But since all of the WC REST endpoints are extensions of core WP REST functionality, you can always use core WP filters if needed. [Here is an example](https://github.com/Automattic/wc-calypso-bridge/blob/master/store-on-wpcom/inc/wc-calypso-bridge-add-bacs-accounts.php) of me doing some _interesting_ things to manipulate some core REST responses.

While that approach I think isn't totally optimal, as all of our REST logic would execute first, it could work. But I'm fine with other approaches if we need more hooks/filters.

> Will code doing so be allowed into the Sequential Order Number extensions and are we OK with the wonky behavior for stores that aren't on the latest version of those extensions?

I think this is kind of a chicken/egg sort of scenario, but I feel like our best long-term solution for scenarios like this is to make sure there is a way for extension developers to ensure their logic can be supported within the new analytics. I just fear the moment we start adding in specific logic for extension x/y/z we will be engaging in a game of whack-a-mole.

jeffstieler: > I think it is already filterable [here](https://github.com/woocommerce/woocommerce/blob/89740f9583893581a591651c757b6b9085cf6eb1/includes/api/v2/class-wc-rest-orders-v2-controller.php#L313)

That's in the `prepare_object_for_response`, not `prepare_objects_query` - we're specifying a list of Order IDs to query for based on order number. The handling of the `number` parameter needs to happen before the query, not at the time of preparing the query results for the API response.

Looking at `woocommerce_rest_orders_prepare_object_query` again, I think that will work:

```php
add_filter( 'woocommerce_rest_orders_prepare_object_query', function ( $args, $request ) {
	if (
		'/wc/v4/orders' === $request['route'] &&
		isset( $request['number'] )
	) {
		// Extension handles 'number' value here and modifies $args.

		// Remove the 'number' parameter to avoid ID-based query.
		unset( $request['number'] );
	}

	return $args;
}, 10, 2 );
```

> I think this is kind of a chicken/egg sort of scenario, but I feel like our best long-term solution for scenarios like this is to make sure there is a way for extension developers to ensure their logic can be supported within the new analytics. I just fear the moment we start adding in specific logic for extension x/y/z we will be engaging in a game of whack-a-mole.

Cool. Should I open PRs for the two Sequential Order Number extensions then?

timmyc: > Cool. Should I open PRs for the two Sequential Order Number extensions then?

Sure if we can take some of the logic in the PR here and port it over I think that would be great. And would be something we can link to as an example of how to extend wc-admin via plugin logic 👍

jeffstieler: https://github.com/skyverge/woocommerce-sequential-order-numbers/pull/16

joshuatf (approved):
Code here looks good and won't break our existing filter while we wait for https://github.com/skyverge/woocommerce-sequential-order-numbers/pull/16.  

Made a note about a comment to remind us not to change the param from a string to integer, but pre-approving.  ✅

PR #1910, opened 2019-03-25
Hide chart name on smaller viewports and move interval select

### LINKED ISSUE #1574: Chart: Oddities in Narrow Viewports
**Describe the bug**
When viewing a chart in a narrow/mobile viewport, there are a few things that look a tad bit off:

chart-small-viewport-oddities

- In the screenshot above, having "Items Sold" displayed two times, in such close proximity to each other seems redundant. Is the text in the summary number good enough to convey what is being shown in the chart in a small window like this?
- The interval selection drop down gets really lost when it is wrapped onto a new line. It is visually hard to tell that it can be adjusted in this view also.

**To Reproduce**

1. Use a device or an emulator, and open any report with a chart

/cc @LevinMedia @josemarques

joshuatf:
Fixes #1574 

Hides the chart name on smaller viewports since it's already displayed in the summary numbers.

### Before
Screen Shot 2019-03-25 at 3 25 42 PM

### After
Screen Shot 2019-03-25 at 3 25 22 PM

1. Narrow your viewport to <960px.
2. Visit any report page.
3. Note that the title is hidden on the chart, but visible in summary numbers, and interval box is correctly positioned to the left.

### Question

Should we extract some of these styles to a separate stylesheet under `ReportTable`?

If another source were to use the table components from our packages, summary items may not be included adjacent to a chart.  Hiding the chart title in this case would not make sense and may be more specific to our use case.

### DIFF
--- a/packages/components/src/chart/style.scss
+++ b/packages/components/src/chart/style.scss
@@ -32,6 +32,9 @@
 			line-height: 18px;
 			margin-left: $gap;
 			margin-right: $gap;
+			@include breakpoint( '<960px' ) {
+				display: none;
+			}
 		}
 	}
 
@@ -74,16 +77,11 @@
 	min-height: 50px;
 	padding: 8px $gap 0 $gap;
 
-	@include breakpoint( '<782px' ) {
-		padding: 0 $gap;
-		margin-top: -8px;
-	}
-
-	@include breakpoint( '<782px' ) {
+	@include breakpoint( '<960px' ) {
 		border-right: 0;
 		min-height: 0;
-		order: 1;
-		width: 100%;
+		margin-left: 0;
+		padding-left: $gap / 2;
 	}
 
 	#wpbody & .components-select-control__input {
--- a/packages/components/src/chart/style.scss
+++ b/packages/components/src/chart/style.scss
@@ -75,6 +75,7 @@
 	padding: 8px $gap 0 $gap;
 
 	@include breakpoint( '<960px' ) {
+		width: 100%;
 		order: 1;
 		margin-top: -8px;
 		margin-left: 0;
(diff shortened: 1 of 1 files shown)


Aljullu (commented):
This is looking much better now, nice improvement!

> Should we extract some of these styles to a separate stylesheet under `ReportTable`?
> 
> If another source were to use the table components from our packages, summary items may not be included adjacent to a chart. Hiding the chart title in this case would not make sense and may be more specific to our use case.

That's a good point, but I'm not sure to understand why those styles should be moved under `ReportTable`. :thinking:  Why not `ReportChart` instead?

Another solution would be writing the CSS so it's only applied to charts that follow summary numbers:
```CSS
@include breakpoint( '<960px' ) {
	.woocommerce-summary + .woocommerce-chart {
		/* All the styles that hide the chart title */
	}
}
```

timmyc:
Really like this solution - removes the redundant chart labeling in the narrow viewport, and makes everything look much tidier.

joshuatf:
>Why not ReportChart instead?

Mondays are hard 😞 - yes I meant to say `ReportChart`.

>only applied to charts that follow summary numbers

I like your solution here much better.  Added this.

Aljullu (approved):
This is looking good now, thank you! Just one small oddity: I tried removing the Summary Numbers and noticed the line chart/bar chart buttons are now centered. Is that expected?
image
I think that's caused because of the removal of `width: 100%;` from `.woocommerce-chart__interval-select {}`.

joshuatf:
Thanks for reviewing @Aljullu.

>I tried removing the Summary Numbers and noticed the line chart/bar chart buttons are now centered. Is that expected?

I was able to replicate this using Firefox- nice catch.  Added width parameters and seems to be resolved 👍

PR #1911, opened 2019-03-25
Don't query products all products for empty categories

### LINKED ISSUE #1908: Parent Categories: Stat totals include child categories
**Describe the bug**
Discovered in https://github.com/woocommerce/woocommerce-admin/pull/1873#pullrequestreview-217812675, when comparing categories which are parent categories, Summary numbers are off. They most likely include child category numbers as well.

**To Reproduce**

1. In the Categories Report, compare parent categories such as "Music" and "Clothing" if you have woo dummy products.
2. See the totals in the table and chart be less than Summary number totals.

**Expected behavior**

I'd expect them to be the same. See "Items sold" below.

**Screenshots**

Screen Shot 2019-03-25 at 2 01 14 PM

joshuatf:
Fixes #1908 

Empty product categories were not filtering in product stats, affecting the products report when filtered by a category and category summary numbers.

This PR checks if no category slugs are found and adds `-1` to the products SQL query if a category param is set but no products are found.

### Before
Screen Shot 2019-03-25 at 5 48 00 PM

### After
Screen Shot 2019-03-25 at 5 47 10 PM

1. Create at least 2 categories with no products directly underneath (this can be a parent category as long as there are no products directly under it).
2. Make an order not in those categories.
3. Visit the categories report.
4. Check that the stats/summary numbers are correct on the Categories report and don't show revenue for the period with the 2 empty categories.

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-data-store.php
@@ -730,6 +730,11 @@ class WC_Admin_Reports_Data_Store {
 				$cat_slugs[] = $product_cat->slug;
 			}
 		}
+
+		if ( empty( $cat_slugs ) ) {
+			return array();
+		}
+
 		$args = array(
 			'category' => $cat_slugs,
 			'limit'    => -1,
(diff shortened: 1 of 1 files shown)


joshuatf:
@psealock @Aljullu Let me know if this is the same issue you were experiencing.  We might have multiple problems here to address.

Aljullu (approved):
Confirming this fixes the issue you described, but I think it doesn't fix #1908 (at least, not what I described in https://github.com/woocommerce/woocommerce-admin/pull/1873#pullrequestreview-217812675, which I can still replicate).

Let's see what @psealock says. I can create a new issue for the bug I described in #1873, in case @psealock thinks his issue is already fixed.

Anyway, this PR is good to go. :rocket:

psealock:
Changes look good here, but @Aljullu is correct this isn't fixing the problem at hand.

I'll copy and paste this into a new issue so we can merge this PR and not lose the conversation.

joshuatf:
Thanks for the review @Aljullu and @psealock!

>I'll copy and paste this into a new issue so we can merge this PR and not lose the conversation.

Awesome- I'm looking into this one as well.  I have a feeling we have a couple different issues going on here.  Might need a set of your data to recreate this.

PR #1912, opened 2019-03-25
Add coupons count to data store and hook up to summary numbers in orders report

### LINKED ISSUE #1695: Incorrect coupon display
**Describe the bug**
The orders report needs to be more clear about what "coupons" counts in the bottom information table.

**To Reproduce**

1. Place an order using a coupon - you'll need to add a coupon if you don't have a test coupon.
2. Go to the new Orders report
3. Take a look at the bottom row https://cloudup.com/cj6Pjc0paCk - That should be : R100 in coupons

**Expected behavior**
I would have expected a count of coupons used or the total coupon amount used.

**Screenshots**

https://cloudup.com/cj6Pjc0paCk

joshuatf:
Fixes #1695 

~~Adds currency formatting to the coupons summary.~~

Changes the coupon numbers in the table to use distinct coupon code count instead of the total coupon amount.

~~1. Go to the Orders report.
2. Check that the coupons summary number is formatted as a currency.~~

1.  Add some orders with coupons.  Try adding multiple coupons to an order and using the same coupon across different orders.
2.  Check that the Orders report coupon number reflects the unique number of coupons during a selected period.
3.  Check that all unit tests are passing.


Aljullu (commented):
This is testing good, but looking at the design (p6riRB-3dO-p2), I'm not sure if we want to display the monetary value or the absolute number of coupons used. @LevinMedia thoughts?

LevinMedia:
@Aljullu @joshuatf - Since we're just listing the coupons by name in this report, I think it should be the deduplicated number of coupons used, ie if the same coupon is used twice, it gets counted once - similar to the product report where we do the same thing with products sold.

LevinMedia:
@joshuatf @Aljullu since we're not showing coupon monetary value, lets use absolute number here.

joshuatf:
@LevinMedia Okay, so the total coupons used in a given timeframe (not unique/distinct coupons), right?

LevinMedia:
@joshuatf Yep!

LevinMedia:
@joshuatf I just double checked - it looks like we're using the de-duplicated / unique count for products in that report, so we should probably follow suit with coupons. Sorry for the switch.

joshuatf:
@warrendholmes Is this clear if the number refers to unique coupon codes or would you expect the string to be more specific?  E.g., `3 unique coupons`.  /cc @LevinMedia

Aljullu:
@joshuatf I think this PR got some changes from another branches. Would it be easy to rebase it with master?

joshuatf:
@Aljullu Thanks for the heads up- just rebased from master.

Aljullu:
Thanks for the rebase @joshuatf.

This is testing well in the _Orders_ report, but I noticed it breaks the _Coupons_ column of the _Revenue_ report. It's doesn't show any info and when sorting by coupons it crashes (`Unknown column 'coupons' in 'order clause'`):
coupons-column

joshuatf:
Thanks for catching that @Aljullu.  I did not have the same error, but obviously missing information.  Added coupon and coupons_count to the data store so we can access both.

Since we're joining the coupons in the order stats data store now, it might make sense to drop the `coupon_total` from the `wc_order_stats` table and use the coupon table as the source of truth.  I'll create a separate issue for that to keep the scope down on this PR.

Aljullu (approved):
Thanks for fixing that @joshuatf! This is testing well now and code looks good too.

PR #1919, opened 2019-03-26
Readme updates.

timmyc:
A quick round of updates to the readme prior to pushing files to the svn repo.


readme.txt
```diff
@@ -2,7 +2,8 @@
 Contributors: automattic
 Tags: ecommerce, e-commerce, store, sales, reports, analytics, dashboard, activity, notices, insights, stats, woo commerce, woocommerce
 Requires at least: 5.0.0
-Tested up to: 5.0.3
+Tested up to: 5.1.1
+Requires PHP: 5.2
```
psealock: You say `5.2.4` down below, although I don't know if thats a big deal or not.

timmyc: excellent catch - i grabbed all of the PHP items from blocks, so that is likely inconsistent there too 😆

Issue #1889, opened 2019-03-22
Activity Panel: '0' appearing after Reviews tab title when there are no reviews


Aljullu:
When `unreadReviews` is assigned, it must be wrapped with `Boolean()`, otherwise its value is `0` and that's displayed to the user in the tab title.

Report by @LevinMedia [here](https://github.com/woocommerce/woocommerce-admin/pull/1880#pullrequestreview-217823014).

PR #1921, opened 2019-03-27
Release/0.9.0

timmyc:
Changes from the build process, ✂️ a hot new release to bring in all the text-domain changes, bug fixes, and amazing work since v0.8.0.

Here is the zip if you would like to take it for a spin:

[woocommerce-admin.zip](https://github.com/woocommerce/woocommerce-admin/files/3010958/woocommerce-admin.zip)

### DIFF
--- a/docs/components/analytics/report-chart.md
+++ b/docs/components/analytics/report-chart.md
@@ -27,12 +27,12 @@ Whether there is an API call running.
 
 Label describing the legend items.
 
-### `limitProperty`
+### `limitProperties`
 
-- Type: String
+- Type: Array
 - Default: null
 
-Allows specifying a property different from the `endpoint` that will be used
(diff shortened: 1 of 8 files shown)


psealock (approved):
LGTM!

ryelle:
Took a quick spin with the zip on an empty site I use for testing the blocks release builds (has products but no orders) – even in the empty state this worked well 👍🏻

(I did notice #1889, but that's already reported)

Aljullu:
> (I did notice #1889, but that's already reported)

Actually #1889 is a quick fix. :smile: 

If you think it would be great to have it fixed before releasing, I created a PR: #1930.

timmyc:
> If you think it would be great to have it fixed before releasing, I created a PR:

I think we will likely do another release soon,

PR #1926, opened 2019-03-27
Check if customer creation date is not null before formatting

### LINKED ISSUE #1925: WC Customer date created is null when customer doesn't exist
**Describe the bug**
When a customer doesn't exist, but the order has a user ID, `WC_Customer` will attempt to call date on null.

https://github.com/woocommerce/woocommerce-admin/blob/3504aa6d363ba29d94467c6b60940a193159b9c4/includes/data-stores/class-wc-admin-reports-customers-data-store.php#L477

**To Reproduce**

1. Import a wordpress xml export that includes orders with registered customers (users will not be imported).
2. Sync orders.
3. Note the error.

**Expected behavior**
Expected to be able to continue processing order even without customer.

**Screenshots**
Screen Shot 2019-03-27 at 11 12 27 AM

joshuatf:
Fixes #1925 

Checks if the customer date exists before calling date formatting function on it.

1. Import orders with associated registered customers.
2. Note that import succeeds despite not having imported users.


psealock (approved):
LGTM!

PR #1928, opened 2019-03-27
Remove unused categories in orders data store

### LINKED ISSUE #1927: SQL syntax error when no products exist in an order
**Describe the bug**
Not a common case for most sites, but when importing orders, products can be unlinked causing an error when fetching categories for products under that category.

Since we're not currently in need of categories here and they are fetched separately, we can probably remove this call to reduce queries in the orders data store.

**To Reproduce**

1. Import an order or manually remove an order's products.
2. Visit the orders report with only that order.
3. Note the error.

**Expected behavior**
Expected no error to occur.

**Screenshots**
Screen Shot 2019-03-27 at 1 28 51 PM

joshuatf:
Fixes #1927 

No longer using categories here and this prevents the SQL error that occurs when no products exist in an order.

1. Check that the orders report still works as expected.
2. Import or create an order manually with no products.
3. Visit the orders report and filter to only show the order without products.
4. Note the lack of error.


Aljullu (approved):
Tested and it works fine. Nice to see removing code fixes an issue. :slightly_smiling_face:

PR #1933, opened 2019-03-27
Add profile wizard base page

### LINKED ISSUE #1881: Onboarding - Profile Wizard: Create base onboarding route
This task involves creating a new route and basic skeleton folder for the profiling/onboarding wizard. When visiting this route, the rest of `wp-admin`/`wc-admin` should be hidden and the background should be purple.

Screen Shot 2019-05-02 at 2 34 25 PM

emdashcodes:
Fixes #1881.

It creates a base page for the profile wizard work, and does some magic so we can hide wp-admin/WooCommerce elements and take over the page visually.

### Screenshots

Screen Shot 2019-05-02 at 12 16 31 PM

* Change `const profileWizardComplete` to `false` in `client/dashboard/index.js`. Since the onboarding routes are still early and have no method for hiding them, these are defaulted to true.
* Run `npm start`.
* Visit the main `wc-admin` route/dashboard and note that the profile wizard page is shown instead of wp-admin.

### DIFF
--- a/client/dashboard/index.js
+++ b/client/dashboard/index.js
@@ -16,9 +16,16 @@ import Leaderboards from './leaderboards';
 import { ReportFilters } from '@woocommerce/components';
 import StorePerformance from './store-performance';
 import TaskList from './task-list';
+import ProfileWizard from './profile-wizard';
 
 export default class Dashboard extends Component {
 	renderDashboardOutput() {
+		// @todo This should be replaced by a check from the REST API response from #1897.
+		const profileWizardComplete = true;
+		if ( window.wcAdminFeatures.onboarding && ! profileWizardComplete ) {
+			return <ProfileWizard />;
+		}
+
 		// @todo This should be replaced by a check of tasks from the REST API response from #1897.
 		const requiredTasksComplete = true;
 		if ( window.wcAdminFeatures.onboarding && ! requiredTasksComplete ) {
--- a/client/dashboard/profile-wizard/index.js
+++ b/client/dashboard/profile-wizard/index.js
@@ -1,0 +1,37 @@
+/** @format */
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { Component } from '@wordpress/element';
+
+/**
+ * Internal depdencies
+ */
+import './style.scss';
+import { H } from '@woocommerce/components';
+
+export default class ProfileWizard extends Component {
+	componentDidMount() {
+		const body = document.querySelector( '.wp-admin' );
+		body.classList.add( 'woocommerce-profile-wizard__body' );
+	}
+
+	componentWillUnmount() {
+		const body = document.querySelector( '.wp-admin' );
+		body.classList.remove( 'woocommerce-profile-wizard__body' );
+	}
+
+	render() {
+		return (
+			<div>
+				<H className="woocommerce-profile-wizard__header-title">
+					{ __(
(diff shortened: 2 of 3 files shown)


client/dashboard/index.js
```diff
@@ -15,25 +15,37 @@ import Leaderboards from './leaderboards';
...
 	render() {
 		const { query, path } = this.props;
+
+		// @todo This should be replaced by a check to see if the wizard was completed or skipped (#1884, #1883).
+		const profileWizardComplete = true;
 		// @todo This should be replaced by a check of tasks from the REST API response from #1897.
-		const requiredTasksComplete = false;
+		const requiredTasksComplete = true;
+
+		let dashboardOutput;
```
joshuatf: Optional refactor: we could put this into a `renderDashboardOutput()` function and return the output block to avoid using `let`.

client/dashboard/profile-wizard/style.scss
```diff
@@ -0,0 +1,27 @@
...
+
+	.woocommerce-profile-wizard__header-title {
+		color: #fff;
+	}
+
+	#woocommerce-layout__primary {
+		text-align: center;
+		margin-left: auto;
+		margin-right: auto;
+		width: 70%;
```
joshuatf: Should we use a pixel based max-width here or min-width media query?  Might be a touch tight on smaller screens once we start adding more content.

```suggestion
		@include breakpoint( '>960px' ) {
			width: 70%;
		}
```
Screen Shot 2019-03-28 at 12 23 28 PM

client/dashboard/profile-wizard/index.js
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Internal depdencies
+ */
+import './style.scss';
+import { H } from '@woocommerce/components';
+
+export default class ProfileWizard extends Component {
+	componentDidMount() {
+		const body = document.querySelector( '.wp-admin' );
```
joshuatf: Nice solution!  Would `document.body` be faster and less error prone here?  Or is `.wp-admin` needed to target other areas?

joshuatf (approved):
Nice work on this, Justin!  Added a few optional nitpicky comments, but this is testing well and pre-approving.

Notices are displayed across this page and the PR I opened, but I think that's okay to handle separately.  Opened an issue [here](https://github.com/woocommerce/woocommerce-admin/issues/1937) where we can address that.

justinshreve:
Going to close this one for now until we make a decision about the profiler. I can always re-open/rebase it later if needed.

justinshreve:
@joshuatf I've updated the PR here based on your feedback, and also updated the design to match the newest Figma iteration. This branch was also pretty stale, so I rebased it.

Do you mind doing a second review?

joshuatf (approved):
Thanks for those changes, @justinshreve!  

One last minor thing is the layout wrapper padding for this screen is sightly off-center (easier to see agains the centered notice).

Screen Shot 2019-05-07 at 8 33 21 PM

```scss
.woocommerce-profile-wizard__body .woocommerce-layout__main {
    padding-right: 0;
}
```

We can toss this in here if you like or add to a follow-up PR since we'll probably also have to address alerts and other bits on this screen, but pre-approving either way. 🚢

justinshreve:
Thanks @joshuatf -- I went ahead and moved that issue to the current sprint, and pasted a link to your comment. I'll leave the fix for that PR so that we/I can test with some proper notices showing up.

PR #1936, opened 2019-03-28
Sync: Avoid infinite loop when REPLACE-ing SQL rows

psealock:
When the sync process uses `$wpdb->replace` the returned value is the number of affected rows. If the database row already exists this number will be `2`. This can occur if you rebuild your tables and you have guest customers. A new id is created and the row is replaced with a new row and an updated `customer_id`.

Checks to make sure the sync process went smoothly operated under the assumption of a successful `$wpdb->replace` always being `1`.

https://github.com/woocommerce/woocommerce-admin/blob/fc678aa68583bc7db4df5184d11aa03ca81b7b72/includes/class-wc-admin-reports-sync.php#L266-L284

This PR includes a check for a value of `2` as a success indicator.

Thanks @jeffstieler for the pre-flight hand holding while we tracked this one down.

1. Make sure you have some guest checkouts.
2. Analytics > Settings > Rebuild reports.
3. Ensure that no infinite loop is kicked off. Previously, this would result in runaway creation of customers.

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-coupons-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-coupons-data-store.php
@@ -399,7 +399,8 @@ class WC_Admin_Reports_Coupons_Data_Store extends WC_Admin_Reports_Data_Store im
 			 */
 			do_action( 'woocommerce_reports_update_coupon', $coupon_id, $order_id );
 
-			$num_updated += intval( $result );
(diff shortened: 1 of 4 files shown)


Aljullu (approved):
This is testing well and code looks good. It took me a while to be able to reproduce the issue, even though I have several guest checkouts, `intval( $result )` was always resolving to `1`.

Issue #1897, opened 2019-03-22
Onboarding - Task List: Levels & Tasks REST API


justinshreve:
This task involves creating a set of REST API endpoints that define levels, tasks, and subtasks for the user to complete during setup. The details below are my current plans on how I think it should work so it can be extendable, but still power the flows described in the designs. This issue may be broken down into sub-issues in the future, or handled in multiple PRs.

Levels, tasks, and subtasks should be definable via WordPress hooks & filters. Levels and tasks are described in the [task list/home screen thread](https://wp.me/p6riRB-44u). An example for subtasks can be seen in the [add products thread](https://wp.me/p6riRB-45C).

Here are the REST API endpoints that I think are necessary:
```
GET /onboarding/levels - Gets a list of valid levels.
GET /onboarding/tasks - Get a list of valid tasks for the provided `level` query parameter. Contains a list of subtasks for each task.
PUT /onboarding/tasks - Mark a task or subtask as completed. Can accept multiple tasks and subtasks IDs.
```

Below are the fields I think will be necessary to power this, and should be registered via PHP:

Level:
```
slug: A slug for this level.
tasks: An array of task slugs that should be contained in the level.
```

Task:
```
slug: A slug for this task. Extensions should namespace.
label: Label for the task.
description: Description for the task.
illustration: An illustration to be used in the task display.
status: The current status of the task (pulled from the table).
is_visible_conditional: A boolean describing whether the current task should be shown. This value should be filterable, so different combinations of settings can set the state. Example: Checking whether the business is already selling elsewhere and conditionally showing a task to use the same payment provider.
in_progress_conditional: A boolean describing whether a task has been started but is incomplete.
is_completed_conditional: A boolean describing whether the current task is completed. This value should be filterable, so different combinations of settings can set the state. Example: A payment method has been configured and setup.
```

Subtask.
```
slug: A slug for this task.
label: Label for the task.
description: Description for the task.
actions: An array of simple actions that can be completed via the subtask. These can contain a label, display_type (primary, secondary, link) action_type (link, or handler) target (URL or handler - see #XXXX for details about handlers). mark_completed This should be another field attached to actions as a boolean describing if this task should be marked as completed immediately after clicking.
status: The current status of the task (pulled from the table).
is_visible_conditional: If the subtask is visible. Example: The profile step was skipped, so we show different product creation steps.
is_completed_conditional: If the subtask is completed.
```

[...]

PR #1938, opened 2019-03-28
Onboarding: Add levels REST API

joshuatf:
Fixes (part of) #1897

Adds the levels API to request all levels and related tasks.

### Screenshots
Screen Shot 2019-03-28 at 4 57 47 PM

1.  Make a GET request to `/wp-json/wc/v4/onboarding/levels`.
2.  Make sure you get a valid response with intial levels and tasks.
3.  Check that tests are passing.

### Questions / Thoughts

@justinshreve I followed your outline in #1897 pretty closely which was fantastic and very thorough- thanks!  I made a couple minor changes in this PR, but mainly wanted to get your thoughts on conditional fields.

I removed all conditional fields in tasks.  I may have misunderstand the purpose of these, but my thinking is two-fold:
   1. We can change the status using the `woocommerce_onboarding_levels` filter if we want to verify that something is complete or should be hidden.  We can also conditionally remove a task entirely if conditions are met.
  2.  Other status changes that we don't want to handle on API request we can handle prior to this request by updating the `wp_wc_task_status` table.

What do you think?

### DIFF
--- a/includes/api/class-wc-admin-rest-onboarding-levels-controller.php
+++ b/includes/api/class-wc-admin-rest-onboarding-levels-controller.php
@@ -52,55 +52,76 @@ class WC_Admin_REST_Onboarding_Levels_Controller extends WC_REST_Data_Controller
 
 	/**
 	 * Get an array of all levels and child tasks.
+	 *
+	 * @todo Status values below should pull from the database task status once implemented.
 	 */
 	public function get_levels() {
 		$levels = array(
 			array(
-				'slug'  => '',
+				'id'    => 'account',
 				'tasks' => array(
 					array(
-						'slug'                     => '',
-						'description'              => '',
-						'illustration'             => '',
-						'status'                   => '',
-						'is_visible_conditional'   => '',
-						'in_progress_conditional'  => '',
-						'is_completed_conditional' => '',
-						'is_required'              => '',
+						'id'           => 'create_account',
+						'label'        => __( 'Create an account', 'woocommerce-admin' ),
+						'description'  => __( 'Speed up & secure your store', 'woocommerce-admin' ),
+						'illustration' => '',
+						'status'       => 'visible',
+						'is_required'  => false,
 					),
 				),
 			),
 			array(
-				'slug'  => '',
+				'id'    => 'storefront',
 				'tasks' => array(
 					array(
-						'slug'                     => '',
-						'description'              => '',
-						'illustration'             => '',
-						'status'                   => '',
-						'is_visible_conditional'   => '',
-						'in_progress_conditional'  => '',
-						'is_completed_conditional' => '',
-						'is_required'              => '',
+						'id'           => 'add_products',
+						'label'        => __( 'Add your products', 'woocommerce-admin' ),
+						'description'  => __( 'Bring your store to life', 'woocommerce-admin' ),
+						'illustration' => '',
+						'status'       => 'visible',
+						'is_required'  => true,
+					),
+					array(
+						'id'           => 'customize_appearance',
+						'label'        => __( 'Customize Appearance', 'woocommerce-admin' ),
+						'description'  => __( 'Ensure your store is on-brand', 'woocommerce-admin' ),
+						'illustration' => '',
+						'status'       => 'visible',
+						'is_required'  => false,
 					),
 				),
 			),
 			array(
-				'slug'  => '',
+				'id'    => 'checkout',
 				'tasks' => array(
 					array(
-						'slug'                     => '',
-						'description'              => '',
-						'illustration'             => '',
-						'status'                   => '',
-						'is_visible_conditional'   => '',
-						'in_progress_conditional'  => '',
-						'is_completed_conditional' => '',
-						'is_required'              => '',
+						'id'           => 'configure_shipping',
+						'label'        => __( 'Configure shipping', 'woocommerce-admin' ),
+						'description'  => __( 'Set up prices and destinations', 'woocommerce-admin' ),
+						'illustration' => '',
+						'status'       => 'visible',
+						'is_required'  => true,
+					),
+					array(
+						'id'           => 'configure_taxes',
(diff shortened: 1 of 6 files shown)


includes/api/class-wc-admin-rest-onboarding-levels-controller.php
```diff
@@ -0,0 +1,250 @@
...
+	 */
+	public function prepare_item_for_response( $item, $request ) {
+		$data     = $this->add_additional_fields_to_object( $item, $request );
+		$data     = $this->filter_response_by_context( $data, 'view' );
+		$response = rest_ensure_response( $data );
+
+		$response->add_links( $this->prepare_links( $item ) );
+
+		/**
+		 * Filter the ist returned from the API.
```
Aljullu: Typo here?

Aljullu (commented):
I tested and it looks good to me, but I will let @justinshreve take a final look. Tests are green too.

I noticed you didn't add the `is_required` boolean you suggested [here](https://github.com/woocommerce/woocommerce-admin/issues/1897#issuecomment-477002330), is that expected for a follow-up?

includes/api/class-wc-admin-rest-onboarding-levels-controller.php
joshuatf: Yep- thanks! 😄

joshuatf:
Thanks for the review, @Aljullu!  We might be putting this one on hold until onboarding plans are finalized.

>I noticed you didn't add the is_required boolean you suggested here, is that expected for a follow-up?

Must have gotten lopped off after removing conditional values- rebased and added back in.  Thanks for catching that!

includes/api/class-wc-admin-rest-onboarding-levels-controller.php
```diff
@@ -0,0 +1,256 @@
...
+	 * Route base.
+	 *
+	 * @var string
+	 */
+	protected $rest_base = 'onboarding/levels';
+
+	/**
+	 * Register routes.
+	 *
+	 * @since 3.5.0
```
justinshreve: Let's drop all the `@since` lines here. We can add them in when code eventually gets merged to core. (We should also clean up other lines in the repo since we don't know what version code will be in yet).

includes/api/class-wc-admin-rest-onboarding-levels-controller.php
```diff
@@ -0,0 +1,256 @@
...
+						'description'  => __( 'Ensure your store is on-brand', 'woocommerce-admin' ),
+						'illustration' => '',
+						'status'       => 'visible',
+						'is_required'  => false,
+					),
+				),
+			),
+			array(
+				'id'    => 'checkout',
+				'tasks' => array(
```
justinshreve: What do you think about adding an additional filter for each of these subtask arrays?

`woocommerce_onboarding_checkout_tasks` or something? That way an extension can hook into a specific section for example.

emdashcodes (commented):
>  I removed all conditional fields in tasks. I may have misunderstand the purpose of these, but my thinking is two-fold:
> We can change the status using the woocommerce_onboarding_levels filter if we want to verify that something is complete or should be hidden. We can also conditionally remove a task entirely if conditions are met.
> What do you think?

Yeah this sounds good to me. I was just trying to work through how things might be hidden/marked done but I think that can be exposed over the status response only.

I have a couple of small comments.

This is testing well. 👍

includes/api/class-wc-admin-rest-onboarding-levels-controller.php
joshuatf: You got it! 👌 Dropped all other `@since` tags from around the repo as well.

includes/api/class-wc-admin-rest-onboarding-levels-controller.php
joshuatf: I was thinking it would be trivial to find and modify array items from `woocommerce_onboarding_levels`, but maybe this would be easier if we make this an associate array?

```php
$levels = array(
    ...
    'checkout' => array(...)
);
```

Then we can modify in the filter accordingly:

```php
add_filter( 'woocommerce_onboarding_levels', function( $levels ) {
    $levels[ 'checkout' ][] = array(...);
    $levels[ 'boss-level' ] = array(...);
    return $levels
} );
```

Might also need to convert this back to a non-associative array when preparing to retain order in JSON, but I can test if you like this method.

justinshreve: An associate array works for me. I think it saves a little time / is a bit clearer which level you are modifying, so devs don't need to pull out the correct index key first.

joshuatf: Added this in https://github.com/woocommerce/woocommerce-admin/pull/1938/commits/97c6969c0ea976d10d71e130b4dedd623b08a04b.

justinshreve: @joshuatf The response seems not quite right (and the PHP tests are failing:

See account vs checkout

Screen Shot 2019-04-03 at 8 12 22 AM

Screen Shot 2019-04-03 at 8 13 16 AM

joshuatf: I think everything is in the response - was the ID out of order the issue?  We'll lose this ordering once we're in JSON, but went ahead and forced ID to the beginning for readability prior to that.

Also fixed tests for adding in levels to the associative array.

justinshreve: Wow - My eyes totally skipped over `"id": "account"` there so I thought it was missing. Yeah that's what I meant.

emdashcodes (approved):
Thanks for the follow-ups here. This is looking good, and testing well for me 👍.

joshuatf:
Thanks for taking the time to review @justinshreve!

PR #1945, opened 2019-03-29
Cast order totals as floats when calculating net total

### LINKED ISSUE #1944: Non-numeric value error in order syncing
**Describe the bug**
Occasionally, some orders throw a non-numeric value error:

`PHP Warning:  A non-numeric value encountered in /wp-content/plugins/wc-admin/includes/data-stores/class-wc-admin-reports-orders-stats-data-store.php on line 496`

Seems to occur when orders are missing shipping calculations.  We can probably cast those values as float to be cautious.

**To Reproduce**

1. Create an order or import an order on a site without shipping options configured.
2. Note the error.

**Expected behavior**
Expected no error.

joshuatf:
Fixes #1944

Cast to float values just in case nothing is returned when shipping total is not set.

1.  Create a site with no shipping methods (this seems to work for me, but not sure if other factors are at play here).
2.  Create an order.
3.  Make sure the non-numerical warning is not logged.

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-orders-stats-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-orders-stats-data-store.php
@@ -493,11 +493,11 @@ class WC_Admin_Reports_Orders_Stats_Data_Store extends WC_Admin_Reports_Data_Sto
 	 * @return float
 	 */
 	protected static function get_net_total( $order ) {
-		$net_total = $order->get_total() - $order->get_total_tax() - $order->get_shipping_total();
+		$net_total = floatval( $order->get_total() ) - floatval( $order->get_total_tax() ) - floatval( $order->get_shipping_total() );
 
 		$refunds = $order->get_refunds();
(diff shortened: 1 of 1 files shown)


Aljullu (commented):
This is testing well and code looks good, but I was unable to reproduce #1944, so I can't confirm it's fixed. I created a new site without any shipping method and created several orders from `wp-admin` (with and without products), but there was no way I could get that warning. I also tried modifying values directly in the database in order to force the warning to appear with no luck. :confused: 

Let's see if somebody else from the team can reproduce it. Otherwise, I think we can still merge it given that everything works fine and code looks good too.

joshuatf:
@Aljullu Thanks for checking.  FWIW, I'm also having a hard time recreating this issue, but I've seen it pop up a few times now.

Each time it has been around `get_shipping_total()` on newer or imported data so it seems like something is happening in core to not cast that as a float.

timmyc (approved):
LGTM

PR #1949, opened 2019-03-29
Update Variations controller filter names

Aljullu:
Fixes #1948.

In #1866 we updated `WC_Admin_REST_Products_Controller` method names, but we also need to update those names when they are called from `WC_Admin_REST_Product_Variations_Controller`.

### Detailed test instructions:
1. Go to the _Products_ report.
2. Select _Single Product_ filter and choose a variable product.
3. Select _Comparison_ as the second filter.
4. Write something in the _Compare Variations_ input.
5. Verify the server doesn't return with an error.

PR #1952, opened 2019-03-31
Fix SKU sorting in the Variations report when a comparison is active

Aljullu:
Follow-up of #1837.

Fix SKU sorting in the Variations report when a comparison is active.

### Screenshots
image

- First, you will need the fix from #1949, otherwise the Variations autocompleter doesn't work:
`git merge fix/1948-update-variations-controller-filter-names`
- Go to the _Products_ report and filter by a single product, select one that is a variable product.
- Once the variations appear on the table, enable the `sku` column and try sorting by it. Verify it works.
- Now, enable the variations comparison and select some of the comparisons in the input text box.
- Try sorting again by `sku` and verify it works.


psealock (approved):
Testing well. I didn't spot any regressions in the Variations table, well done.

Issue #1953, opened 2019-03-31
Categories endpoint: per_page must be between 1 and 100


Aljullu:
_Products_ and _Categories_ report tables are not loading.

**To Reproduce**

1. Go to the _Products_ or _Categories_ report.
2. Notice the table is not loading.

The error returned by `/wc/v4/products/categories`: `per_page must be between 1 (inclusive) and 100 (inclusive)`.

**Screenshots**
![image](https://user-images.githubusercontent.com/3616980/55293865-66e9b600-53fb-11e9-876d-962092283c2d.png)

PR #1954, opened 2019-04-01
getItems: modify parse flag by isUnboundedRequest

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/1953

Products and Categories reports tables weren't loading due to an error `/wc/v4/products/categories: per_page must be between 1 (inclusive) and 100 (inclusive)`.

https://github.com/woocommerce/woocommerce-admin/pull/1866 introduced a `parse: false` flag to `apiFetch` usage in `getItems`. This allows us to get the response headers but also invalidates the usage of `per_page: -1`. Below is a link to that logic.

https://github.com/WordPress/gutenberg/blob/dee3dcf49028717b4af3164e3096bfe747c41ed2/packages/api-fetch/src/middlewares/fetch-all-middleware.js#L39-L45

This PR changes the logic to reflect the need for each case, ie when we need all the categories vs. when a paginated response along with totals is needed.

1. Go to the Products Report and Categories Report and make sure the tables load.
2. Add at least some products that are considered low in stock.
3. Check that the unread indicator for the stock status is present.

### DIFF
--- a/client/wc-api/items/operations.js
+++ b/client/wc-api/items/operations.js
@@ -36,16 +36,36 @@ function read( resourceNames, fetch = apiFetch ) {
 		const endpoint = typeEndpointMap[ prefix ];
 		const query = getResourceIdentifier( resourceName );
 		const url = NAMESPACE + `/${ endpoint }${ stringifyQuery( query ) }`;
+		const isUnboundedRequest = -1 === query.per_page;
 
 		try {
 			const response = await fetch( {
-				parse: false,
+				/* eslint-disable max-len */
+				/**
+				 * A false parse flag allows a full response including headers which are useful
+				 * to determine totalCount. However, this invalidates an unbounded request, ie
+				 * `per_page: -1` by skipping middleware in apiFetch.
+				 *
+				 * See the Gutenberg code for more:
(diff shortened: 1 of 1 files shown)


client/wc-api/items/operations.js
```diff
@@ -36,16 +36,36 @@ function read( resourceNames, fetch = apiFetch ) {
...
 		const query = getResourceIdentifier( resourceName );
 		const url = NAMESPACE + `/${ endpoint }${ stringifyQuery( query ) }`;
+		const isUnboundedRequest = -1 === query.per_page;
 
 		try {
 			const response = await fetch( {
-				parse: false,
+				/* eslint-disable max-len */
+				/**
+				 * A false parse flag allows a full response including headers which are useful
```
joshuatf: 👍 Thanks for adding this comment

joshuatf (approved):
Thanks for fixing this @psealock!  Changes look good and are testing well.

On a side note, do you think it's worth opening an issue to expand the [readme for the apiFetch](https://github.com/WordPress/gutenberg/blob/dee3dcf49028717b4af3164e3096bfe747c41ed2/packages/api-fetch/README.md)?  Not sure about you, but I remember running into this issue with `parse` and `per_page` the first time and taking a few minutes to realize what was happening.

psealock:
Thanks for the review @joshuatf. Yeah, that was a bit of a mystery of how `per_page = -1` worked and why `parse` would effect that. I am curious why there is no mention of `per_page = -1` already. Maybe they don't want it publicised.

PR #1957, opened 2019-04-01
Include child category stats in parent

### LINKED ISSUE #1915: Category Segments: Child categories included in totals, but not segments
This is a copy/paste from https://github.com/woocommerce/woocommerce-admin/pull/1911#issuecomment-476373476.

joshuatf:
Fixes #1915

Adds child stats to the parents so that parent stats include child stats.

### Questions / Thoughts

I'm pretty sure I'll close this PR in favor of a better solution, but wanted to get some feedback first.

Aside from performance issues that exist here, the biggest issue is that any `_count` fields that grab unique values may be inaccurate due to the getting a distinct count and then merging stats together. (E.g., A parent category having Child Category A and Child Category B that contain products sharing at least one category will add the "unique" order count together which will be off by the number of shared orders.)

I'm thinking the most performant and structurally sound method here would be to create a category lookup table using something similar to a closure pattern built for hierarchical data.

Alternatively, we could prefetch all categories and category relationships (much like what this PR does but instead done before the db queries).  We could then join based on a number of conditions to simulate a hierarchical join.  This resolves the count issues, but retains many of the performance issues in this PR.

/cc @peterfabian 

### Before
Screen Shot 2019-04-02 at 2 57 50 PM

### After
Screen Shot 2019-04-02 at 2 54 59 PM

1.  Create a parent category and any number of child categories.
2.  Create orders that fall under the child categories.
3.  Check reports for the parent category and make sure they include stats from the child.
4.  Note the unique orders count is not accurate for any overlapping categories between products. 😞

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-categories-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-categories-data-store.php
@@ -228,15 +228,16 @@ class WC_Admin_Reports_Categories_Data_Store extends WC_Admin_Reports_Data_Store
 
 		// These defaults are only partially applied when used via REST API, as that has its own defaults.
 		$defaults   = array(
-			'per_page'      => get_option( 'posts_per_page' ),
-			'page'          => 1,
-			'order'         => 'DESC',
-			'orderby'       => 'date',
-			'before'        => WC_Admin_Reports_Interval::default_before(),
-			'after'         => WC_Admin_Reports_Interval::default_after(),
-			'fields'        => '*',
-			'categories'    => array(),
-			'extended_info' => false,
+			'per_page'        => get_option( 'posts_per_page' ),
+			'page'            => 1,
+			'order'           => 'DESC',
+			'orderby'         => 'date',
+			'before'          => WC_Admin_Reports_Interval::default_before(),
+			'after'           => WC_Admin_Reports_Interval::default_after(),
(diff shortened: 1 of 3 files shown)


psealock:
This is working for me, although the `orderby` portion of the query isn't taking this into account

Screen Shot 2019-04-04 at 3 25 08 PM

>I'm thinking the most performant and structurally sound method here would be to create a category lookup table using something similar to a closure pattern built for hierarchical data.

If we go this route, there are some other areas where this could help, performance wise. For instance, when segmenting by product category:

https://github.com/woocommerce/woocommerce-admin/blob/a6290f9f6779ef381436f6e0262a7c5195c52b64/includes/class-wc-admin-reports-segmenting.php#L367-L368

joshuatf:
@psealock Thanks for reviewing this one!

>This is working for me, although the orderby portion of the query isn't taking this into account

Good point.  I attempted to fix the unique counts and the orderby after the query, but ended up dropping it for performance reasons.  I think this is yet another reason to move to a closure pattern table.

jeffstieler:
> I'm pretty sure I'll close this PR in favor of a better solution

Is this ready for review for merge, or is another PR coming?

joshuatf:
@jeffstieler Going to close PR for now and open a discussion around this before we start tossing in new tables.

PR #1960, opened 2019-04-01
Only check for unsnooze note scheduled action in admin_init

### LINKED ISSUE #1946: Action scheduler for notices causes slowdown
**Describe the bug**
Reported in devchat channel: https://woocommercecommunity.slack.com/archives/C4TNYTR28/p1553815081111200

I work for @douglsmith at Simply Charlotte Mason... We tried out the WooCommerce Admin plugin, and it caused a huge slowdown, eventually exhausting the available DB connections. It was running many instances of this same query: `SELECT p.ID FROM wpscm_posts p WHERE p.post_title='wc_admin_unsnooze_admin_notes' AND p.post_type='scheduled-action' AND p.post_status='pending' ORDER BY post_date_gmt ASC LIMIT 1`

I had to take it off the live site as it was killing it. When I re-activated the plugin, it wasn't running that same query, but every so often it would fire off a set (I never caught them, weren't being logged) and they all stayed in sleep mode, so eventually the server ran out of connections. I'll get back to you [Brent from Prospress] when a test site is availalble.

rrennick:
Fixes #1946

The current code checks for the scheduled action on every load. This PR moves that check to `admin_init`.

- Use the query monitor plugin to check that the query identified in the issue is only being run in the dashboard.

### DIFF
--- a/includes/notes/class-wc-admin-notes.php
+++ b/includes/notes/class-wc-admin-notes.php
@@ -18,16 +18,10 @@ class WC_Admin_Notes {
 	const UNSNOOZE_HOOK = 'wc_admin_unsnooze_admin_notes';
 
 	/**
-	 * Schedule events and hook appropriate actions.
+	 * Hook appropriate actions.
 	 */
 	public static function init() {
-		$queue = WC()->queue();
-		$next  = $queue->get_next( self::UNSNOOZE_HOOK );
-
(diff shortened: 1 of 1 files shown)


timmyc:
@jeffstieler could you give this one a review? Thinking we should get this in and do a release since it appears to be causing issues on a few different sites

PR #1975, opened 2019-04-02
Remove coupon_total from order stats table

### LINKED ISSUE #1942: Coupon total no longer necessary in order stats table
Assuming https://github.com/woocommerce/woocommerce-admin/pull/1912 gets merged in, we'll have joins on the coupons table inside the order stats data store.

We should use the coupons table as a single source of truth instead of adding an extra field `coupon_total` to `wc_order_stats`.

This would involve:

* Removing the `coupon_total` field from `wc_order_stats`.
* Using the sum of the `discount_amount` for the joined orders.

joshuatf:
Fixes #1942 

* Removes the `coupon_total` from the order stats table.
* Replaces it with summed data from the join already in place with the coupons table.
* Adds coupon data to tests.

1.  Create an order with coupons.
2.  Check that the coupon amount shown in the Revenue reports is accurate.
3.  Check that all tests pass. ✅


emdashcodes (approved):
Confirmed that the correct coupon amount is displayed, and all tests pass.

Code looks good. 👍

tests/reports/class-wc-tests-reports-orders-stats.php
```diff
@@ -24,11 +24,14 @@ public function test_populate_and_query() {
 		$product->set_regular_price( 25 );
 		$product->save();
 
+		$coupon = WC_Helper_Coupon::create_coupon( 'test-coupon' );
```
timmyc: Thanks for adding in some more tests here.

PR #1976, opened 2019-04-02
Make customer name extendable

### LINKED ISSUE #1579: User profile information not used for customers
User first names and last names that are stored in `wc_customer_lookup` pull information from the billing information on creation.  While this mostly okay, there are cases where the billing info may not be present and shipping or user profile fields could be used as a backup.

This is a follow-up/enhancement to the issue outline here - https://github.com/woocommerce/wc-admin/issues/1428

**Steps to reproduce**

1.  Install the checkout field editor plugin and set billing name fields to optional.
2. Create a new user with first and last names filled out.
3.  Don't fill out the name during checkout.
4.  Note that order information is nameless in reporting.

Would be great if we can pull in this order, cycling to the next if one is not found:
* Billing
* Shipping
* User profile

joshuatf:
Fixes #1579

* Cycles through each name option if one isn't available - billing, shipping, or user profile.
* Adds a filter to hook in and change a customer name.

### Screenshots
Screen Shot 2019-04-02 at 6 56 46 PM
Screen Shot 2019-04-02 at 6 56 34 PM

1. Add the checkout field editor plugin and make shipping/billing name fields optional.
2. Make a new order with each of the following and make sure that the customer name shows up.
  * Registered user name fields
  * Billing name fields
  * Shipping name fields

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-customers-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-customers-data-store.php
@@ -461,16 +461,17 @@ class WC_Admin_Reports_Customers_Data_Store extends WC_Admin_Reports_Data_Store 
 			return $returning_customer_id;
 		}
 
-		$data   = array(
-			'first_name'       => $order->get_billing_first_name( 'edit' ),
-			'last_name'        => $order->get_billing_last_name( 'edit' ),
+		$customer_name = self::get_customer_name_from_order( $order );
+		$data          = array(
+			'first_name'       => $customer_name[0],
+			'last_name'        => $customer_name[1],
 			'email'            => $order->get_billing_email( 'edit' ),
 			'city'             => $order->get_billing_city( 'edit' ),
 			'postcode'         => $order->get_billing_postcode( 'edit' ),
 			'country'          => $order->get_billing_country( 'edit' ),
 			'date_last_active' => date( 'Y-m-d H:i:s', $order->get_date_created( 'edit' )->getTimestamp() ),
 		);
-		$format = array(
+		$format        = array(
 			'%s',
 			'%s',
 			'%s',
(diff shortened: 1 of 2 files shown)


includes/data-stores/class-wc-admin-reports-customers-data-store.php
```diff
@@ -505,6 +506,36 @@ public static function get_or_create_customer_from_order( $order ) {
 		return $result ? $customer_id : false;
 	}
 
+	/**
+	 * Get a customer name from billing, shipping, or user profile.
+	 *
+	 * @param WC_Order $order Order made by customer.
+	 * @return array
+	 */
+	public static function get_customer_name_from_order( $order ) {
```
timmyc: Should we add some unit test coverage here?

joshuatf: Added some tests around the report to cover this.

Also updated the order to begin with profile first since that seems to be what's favored in core.

includes/data-stores/class-wc-admin-reports-customers-data-store.php
```diff
@@ -505,6 +506,43 @@ public static function get_or_create_customer_from_order( $order ) {
...
+		} elseif (
+			$order &&
+			$order->get_shipping_first_name( 'edit' ) ||
+			$order->get_shipping_last_name( 'edit' )
+		) {
+			$first_name = $order->get_shipping_first_name( 'edit' );
+			$last_name  = $order->get_shipping_last_name( 'edit' );
+		}
+
+		return apply_filters( 'woocommerce_customer_name', array( $first_name, $last_name ), $order );
```
jeffstieler: Should we use "report" somewhere in this filter name?

includes/data-stores/class-wc-admin-reports-customers-data-store.php
```diff
@@ -505,6 +506,43 @@ public static function get_or_create_customer_from_order( $order ) {
...
 	}
 
+	/**
+	 * Try to get a customer name from user profile or order information.
+	 *
+	 * @param int      $user_id User ID.
+	 * @param WC_Order $order Order made by customer.
+	 * @return array
+	 */
+	public static function get_customer_name( $user_id = 0, $order = null ) {
```
jeffstieler: Can we eliminate the `$user_id` parameter and just use `$order->get_user_id()` internally?

jeffstieler (approved):
This tests well and the code looks good - just made some optional suggestions. Ready to 🛳.

includes/data-stores/class-wc-admin-reports-customers-data-store.php
joshuatf: We can't quite get rid of `$user_id` here since the it's possible there is no last order made (customer created without order or customer created and order was deleted).  And inversely, `$user_id` could be `0` when there's an order made by a guest.

includes/data-stores/class-wc-admin-reports-customers-data-store.php
joshuatf: Sure!  Prefixed with `woocommerce_reports_` to keep consistent with other filters.

PR #1977, opened 2019-04-02
lookup product title in order meta if product has been deleted

### LINKED ISSUE #1783: Products: Add Product Name to Lookup Table
Follow-Up to #1669 

Per our discussion in last week's hangout, while #1750 does fix the exception when a Product is deleted, we are hoping to implement another solution to ensure a better user experience with a product is deleted.

After more discussion on this, I think we should move ahead with the solution @rrennick posted:

> An alternative I suggested in the hangout was looking up the product name in the order meta when the product had been deleted. There is a bit of overhead for this but as long as it's written so that the lookup only occurs when the product can't be found then it will only affect a small number of queries.

_leaving old idea here just for reference_

The plan discussed is to cache product names in `wc_order_product_lookup` so if a product is ever deleted, we can still display the name of the product as it was titled at the time of the order.

rrennick:
Fixes #1783 

This PR adds a DB query to lookup the last known product name used for a product that has been deleted from the database. To minimize queries, a local cache is used to ensure that only one query is run per deleted product. The product name is shown as `Sample Product (Deleted)`.

### Screenshots

Screen Shot 2019-04-02 at 9 35 40 AM

- Permanently delete a product that has been purchased.
- Enable the report rebuild data tool
- Rebuild report data
- Visit Analytics -> Products
- Select a date range including orders where the deleted product was purchased
- See product name in data table

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-products-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-products-data-store.php
@@ -190,7 +190,7 @@ class WC_Admin_Reports_Products_Data_Store extends WC_Admin_Reports_Data_Store i
 			$extended_info = new ArrayObject();
 			if ( $query_args['extended_info'] ) {
 				$product_id = $product_data['product_id'];
-				$product = wc_get_product( $product_id );
+				$product    = wc_get_product( $product_id );
 				// Product was deleted.
 				if ( ! $product ) {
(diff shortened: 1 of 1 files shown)


emdashcodes (approved):
This is testing well for me -- and the query looks good to me since this should (hopefully) be an edge-case and you added some caching.

I just have one comment about alignment, but otherwise this looks good. Pre-approving.

PR #1983, opened 2019-04-02
add filter length condition to enable filter check

### LINKED ISSUE #1932: Filter button is enabled when changing match select even if there is no filter selected
**To Reproduce**

1. Go to the _Customers_ report.
2. Select _Advanced Filters_ under _Show:_.
3. Change the _Customer Match_ select value from _All_ to _Any_.
4. Notice the _Filter_ button is enabled.

**Expected behavior**
The _Filter_ button should stay disabled until a filter is added.

**Screenshots**
match

rrennick:
Fixes #1932

This PR adds a filter check to disable the filter button if there are no current filters.

- Go to the Customer Report
- Click Advanced Filters
- Change from `All` to `Any`
- The `Filter` button should be disabled until a filter value is added


rrennick:
@joshuatf Thanks :)

PR #1984, opened 2019-04-02
phpcs fixes class-wc-tests-reports-orders-stats.php

rrennick:
See #675

PHPCS fixes for the last file in the /tests/ folder.


rrennick:
@joshuatf Thanks :)

PR #1958, opened 2019-04-01
Scroll to top when navigating pages

Aljullu:
Fixes #1956.

Scrolls the page to the top when navigating between reports.

### Detailed test instructions:

1. Go to the _Dashboard_.
2. Scroll to the bottom.
3. Go to _Analytics_ > _Revenue_, for example.
4. Verify the _Revenue_ report is scrolled to the top, so filters, summary numbers and chart are visible by default.

PR #1989, opened 2019-04-03
Scroll to top only when URL changes

### LINKED ISSUE #1974: Switching between chart types scrolls the page to the top
Probably introduced in #1958.

1. Go to any report.
2. Scroll down.
3. Switch to a bar chart.
4. Notice the page scrolls to the top.

Aljullu:
Fixes #1974.

Related to #1958.

Only scroll to top when URL changes, but not when the query does.

1. Go to the _Dashboard_.
2. Scroll to the bottom.
3. Go to _Analytics_ > _Revenue_, for example.
4. Verify the _Revenue_ report is scrolled to the top, so filters, summary numbers and chart are visible by default.
5. Scroll down a bit and switch to a bar chart.
6. Verify the page is not scrolled to top.

### DIFF
--- a/client/layout/controller.js
+++ b/client/layout/controller.js
@@ -72,6 +72,10 @@ const getPages = () => {
 };
 
 class Controller extends Component {
+	componentDidMount() {
+		window.document.documentElement.scrollTop = 0;
+	}
+
 	componentDidUpdate( prevProps ) {
 		const prevQuery = this.getQuery( prevProps.location.search );
 		const prevBaseQuery = this.getBaseQuery( prevProps.location.search );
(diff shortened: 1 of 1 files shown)


emdashcodes (approved):
👍

coderkevin (approved):
Looks good, tests well.

Only question is, should we scroll to top when navigating to the next page in a list? (e.g. in customers report?)

Aljullu:
> Only question is, should we scroll to top when navigating to the next page in a list? (e.g. in customers report?)

That's a good question. I'm not sure we want to do that because depending on the window size, the summary numbers + chart can take all vertical space, so the table would disappear from the visible area. But we could scroll to the beginning of the table, instead. Tagging @LevinMedia in case he has an opinion about this.

LevinMedia:
@Aljullu @coderkevin if we can make it go to the top of the table, that'd be great!

PR #1992, opened 2019-04-03
SearchListControl: Add an onSearch callback to allow dynamic updating of the list

ryelle:
See https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/505 – We want to 

- There should be no functionality change, since nothing is using this yet.
- You can add an `onSearch` callback to the example in devdocs, and see that it triggers when the search changes.

(once this is merged, I'd also like to do a packages update, so it can be used in the blocks)


emdashcodes (approved):
This looks good to me.

I tested by adding `onSearch={ search => console.log( search ) }` to the devdocs and it worked 👍.

Issue #1584, opened 2019-02-15
Filters should auto-focus the text field


Aljullu:
**Steps to reproduce:**
- Go to the _Products_ report.
- In the filters drop down, select _Single Product_.
- Notice the button to go back is focused.

![filters-focus](https://user-images.githubusercontent.com/3616980/52861549-21b34280-3133-11e9-8c8c-b575613ee5f5.gif)

**Expected results:**
- The text field should be focused, so users can start typing right away.

psealock:
### Prior Art

I have attempted this one and am facing some kind of odd race condition when the search component has a tag in it. Upon mounting the popover, the Tag's "remove" button will successfully receive focus 90% of the time in Chrome and none of the time in Firefox despite all the logic being wrapped in a `setTimeout` called in Popover's `componentDidMount`.

Branch here, https://github.com/woocommerce/woocommerce-admin/tree/fix/search-filter-focus

Accompanying Gutenberg changes here, https://github.com/WordPress/gutenberg/pull/14502

I'm going to pause on this in favour of working on features that move the product forward more for our users.

timmyc:
@psealock I'm going to move this to the sprint 16 backlog - not sure if you are still actively working on this or not.

psealock:
I'm not, but we can give it another shot in 16

PR #1994, opened 2019-04-03
Fix/search filter focus

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/1584

On report pages, the "Show:" dropdown should bring focus to the input if there is one.

### Accessibility

### Screenshots

focus

1. Products Report > Show > Single Product
2. See the focus brought to the input box
3. Click on the "back" button
4. See the focus brought to first element
5. Try again, but this time with a selection. Is focus working?

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -1,8 +1,9 @@
-# (unreleased)
+# 2.0.0 (unreleased)
 - Chart legend component now uses withInstanceId HOC so the ids used in several HTML elements are unique.
 - Chart component now accepts data with negative values.
 - Chart component: new prop `filterParam` used to detect selected items in the current query. If there are, they will be displayed in the chart even if their values are 0.
 - Expand search results and allow searching when input is refocused in autocompleter.
+- Animation Slider: Remove `focusOnChange` in favor of `onExited` so consumers can pass a function to be exectuted after a transition has finished.
 
 # 1.6.0
 - Chart component: new props `emptyMessage` and `baseValue`. When an empty message is provided, it will be displayed on top of the chart if there are no values different than `baseValue`.
--- a/packages/components/src/animation-slider/index.js
+++ b/packages/components/src/animation-slider/index.js
@@ -22,17 +22,9 @@ class AnimationSlider extends Component {
 	}
 
 	onExited() {
-		const { onExited, focusOnChange } = this.props;
+		const { onExited } = this.props;
 		if ( onExited ) {
-			onExited();
-		}
-		if ( focusOnChange ) {
-			const focusable = this.container.current.querySelector(
-				'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])'
-			);
-			if ( focusable ) {
-				focusable.focus();
-			}
+			onExited( this.container.current );
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


psealock:
@Aljullu I was wondering if you could take a look at this one? There is something else stealing focus and I've hit a wall trying to debug. It works fine until there is a product tag already selected. It works sometimes in Chrome (below) and almost never in Firefox.

focus-error

The focus remains either on the `body` or the dropdown button. I've tried multiple approaches with no success. I'm hoping you might have some insight?

Aljullu:
Thanks for working on this tricky issue @psealock. I was testing it and could notice the same issue you reported, but couldn't find what's causing it yet, I will give it another try tomorrow. In any case, what you got so far already handles the most important issue IMO that was happening when navigating inside the dropdown.

psealock:
Thanks for the help @Aljullu! I really appreciate it.

>what you got so far already handles the most important issue IMO that was happening when navigating inside the dropdown

Thats true, we could deploy that portion for now.

Aljullu:
I couldn't find the source of the issue yet, but noticed that replacing [this line of CSS](https://github.com/woocommerce/woocommerce-admin/blob/master/packages/components/src/tag/style.scss#L14) with this:

```CSS
transition: background 0.2s cubic-bezier(0.4, 1, 0.4, 1),
	color 0.2s cubic-bezier(0.4, 1, 0.4, 1);
```

seems to fix the issue:

focus

I don't know exactly what was going on there. Maybe the CSS transition applied to all properties was making the element not to be visible for some ms and that was making the focus not to work?

Let me know if that works for you @psealock!

psealock:
Wow 😲 @Aljullu, nice find here. I had imagined the tag may play a part because it has a popover of its own, but didn't deduce a CSS transition could have played a factor.

I did notice the contents of the dropdown aren't rendered correctly. You can also see the scroll bars in your gif as well.

wierd

I think this may have to do with how the content sizing of the popover are calculated. Its interesting to me how this could be affected by focus.

In any case, a valid point was brought up in the complementary Gutenberg PR:

>It may not be expected / accessible to forcibly "skip" a tabbable element in this way.

In other words, are we running into a11y issues by stealing focus from the first element? Between this comment and the issue seen in the gif, I think its best to limit this PR to directly focusing the input only after the Animation Slider has finished. 

I updated the test instructions to reflect this. How do you feel about this compromise?

Also, many thanks for taking the time to dig into this one!

Aljullu (approved):
> I had imagined the tag may play a part because it has a popover of its own, but didn't deduce a CSS transition could have played a factor.

It was one of the weirdest issues I have seen in my life, honestly! :smile: 

> I think its best to limit this PR to directly focusing the input only after the Animation Slider has finished.

That sounds good. This PR fixes the main concern I had with the filter and I agree with the point made in the Gutenberg PR.

We can fix it in a follow-up, but I still see the scrollbars issue when there is an active tag:

dropdown

packages/components/CHANGELOG.md
```diff
@@ -1,8 +1,9 @@
-# (unreleased)
+# 2.0.0 (unreleased)
```
timmyc: Nice move on the version bump here with the breaking prop change 👍

timmyc:
🔥 nice PR @psealock

Issue #1972, opened 2019-04-02
Dashboard: Add Ability to Register Custom Leaderboards


timmyc:
Currently the list of Leaderboards presented in the Dashboard are [hard coded](https://github.com/woocommerce/woocommerce-admin/blob/master/client/dashboard/leaderboards/index.js#L58). Given a number of the extension's we want to add wc-admin support to are seeking to add custom leaderboards ( p75Izm-1Tp-p2 ) - it feels like a good time to start exploring this!

So the long-range goal for this issue is to allow for an entirely new leaderboard to be registered/supported from an extension. I anticipate this taking longer than one sprint to accomplish, but figured we should begin to work towards the end goal.

_Random Thoughts_
- Perhaps we can provide a "core" leaderboard endpoint that accepts date range picker arguments and outputs an object of leaderboard data, keyed by leaderboard "slug". For example if I'm making a leaderboard for "Bookings" I could hook into the API requests to the `/wc/v4/stats/leaderboard` endpoint, and add in my data for the given date range.
- Alternatively, maybe it would be easier to allow leaderboards to register a data callback method which the REST endpoint would use to assemble the data
- Do we want to limit the number of columns in leaderboard tables?
- Should we refactor current leaderboards that are built stand-alone in JS to use the new scheme?

joshuatf:
Thanks for putting those thoughts together, @timmyc!

>Alternatively, maybe it would be easier to allow leaderboards to register a data callback method which the REST endpoint would use to assemble the data

I'm thinking this one to keep things open-ended and allow them to structure the data as they see fit.

>Do we want to limit the number of columns in leaderboard tables?

I think we should not impose limits, provided horizontal scrolling works efficiently on those tables.

>Should we refactor current leaderboards that are built stand-alone in JS to use the new scheme?

I think refactoring the existing ones is a good idea; this cuts down on the number of components we'll need to maintain.

PR #1995, opened 2019-04-04
Add leaderboard controller and endpoint

joshuatf:
Part of #1972 

Adds a leaderboard endpoint so third party devs can hook in and add custom data.  This will be hooked up to the current leaderboards in a follow-up PR.

### Notes

The response time is very long here, sometimes over 8 seconds.  I've isolated this down to the customer data store request and will investigate this separately, but curious if anyone else has similar response times.

### Screenshots
Screen Shot 2019-04-04 at 5 44 25 PM

1. Make a request to `/wp-json/wc/v4/leaderboards`
2. Try using the available params to modify the response - `before`, `after`, `per_page`, `persisted_query` (e.g. `your=test&query=string`
3. Check that the response and display values match the formatting needed for the leaderboards.

### DIFF
--- a/includes/api/class-wc-admin-rest-leaderboards-controller.php
+++ b/includes/api/class-wc-admin-rest-leaderboards-controller.php
@@ -1,0 +1,209 @@
+<?php
+/**
+ * REST API Leaderboards Controller
+ *
+ * Handles requests to /leaderboards
+ *
+ * @package WooCommerce Admin/API
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Leaderboards controller.
+ *
+ * @package WooCommerce Admin/API
+ * @extends WC_REST_Data_Controller
+ */
+class WC_Admin_REST_Leaderboards_Controller extends WC_REST_Data_Controller {
+	/**
+	 * Endpoint namespace.
+	 *
+	 * @var string
+	 */
+	protected $namespace = 'wc/v4';
+
+	/**
+	 * Route base.
+	 *
+	 * @var string
+	 */
+	protected $rest_base = 'leaderboards';
+
+	/**
+	 * Register routes.
+	 */
+	public function register_routes() {
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base,
+			array(
+				array(
+					'methods'             => WP_REST_Server::READABLE,
+					'callback'            => array( $this, 'get_items' ),
+					'permission_callback' => array( $this, 'get_items_permissions_check' ),
+					'args'                => $this->get_collection_params(),
+				),
+				'schema' => array( $this, 'get_public_item_schema' ),
+			)
+		);
+	}
+
+	/**
+	 * Get an array of all leaderboards.
+	 *
+	 * @param int    $rows Number of rows.
+	 * @param string $after Items after date.
+	 * @param string $before Items before date.
+	 * @return array
+	 */
+	public function get_leaderboards( $rows, $after, $before ) {
+		$leaderboards = array();
+
+		return apply_filters( 'woocommerce_leaderboards', $leaderboards, $rows );
+	}
+
+	/**
+	 * Return all leaderboards.
+	 *
+	 * @param  WP_REST_Request $request Request data.
+	 * @return WP_Error|WP_REST_Response
+	 */
+	public function get_items( $request ) {
+		$leaderboards = $this->get_leaderboards( $request['per_page'], $request['after'], $request['before'] );
+		$data         = array();
+
+		if ( ! empty( $leaderboards ) ) {
+			foreach ( $leaderboards as $leaderboard ) {
+				$response = $this->prepare_item_for_response( $leaderboard, $request );
+				$data[]   = $this->prepare_response_for_collection( $response );
+			}
+		}
+
+		return rest_ensure_response( $data );
+	}
+
+	/**
+	 * Prepare the data object for response.
+	 *
+	 * @param object          $item Data object.
+	 * @param WP_REST_Request $request Request object.
+	 * @return WP_REST_Response $response Response data.
+	 */
+	public function prepare_item_for_response( $item, $request ) {
+		$data     = $this->add_additional_fields_to_object( $item, $request );
+		$data     = $this->filter_response_by_context( $data, 'view' );
+		$response = rest_ensure_response( $data );
+
+		/**
+		 * Filter the list returned from the API.
+		 *
+		 * @param WP_REST_Response $response The response object.
+		 * @param array            $item     The original item.
+		 * @param WP_REST_Request  $request  Request used to generate the response.
+		 */
+		return apply_filters( 'woocommerce_rest_prepare_leaderboard', $response, $item, $request );
+	}
+
+	/**
+	 * Get the query params for collections.
+	 *
+	 * @return array
+	 */
+	public function get_collection_params() {
+		$params             = array();
+		$params['page']     = array(
+			'description'       => __( 'Current page of the collection.', 'woocommerce-admin' ),
+			'type'              => 'integer',
+			'default'           => 1,
+			'sanitize_callback' => 'absint',
+			'validate_callback' => 'rest_validate_request_arg',
+			'minimum'           => 1,
+		);
+		$params['per_page'] = array(
+			'description'       => __( 'Maximum number of items to be returned in result set.', 'woocommerce-admin' ),
+			'type'              => 'integer',
+			'default'           => 5,
+			'minimum'           => 1,
+			'maximum'           => 20,
+			'sanitize_callback' => 'absint',
+			'validate_callback' => 'rest_validate_request_arg',
+		);
+		$params['after']    = array(
+			'description'       => __( 'Limit response to resources published after a given ISO8601 compliant date.', 'woocommerce-admin' ),
+			'type'              => 'string',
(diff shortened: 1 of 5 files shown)


jeffstieler:
Testing on the WCCOM data, the endpoint response isn't matching what I'm seeing on the dashboard:

Request: `/wp-json/wc/v4/leaderboards?after=2018-07-01T00:00:00&before=2018-08-01T00:00:00`

Response: https://gist.github.com/jeffstieler/6e402cb6209377c30202d5c0553fc4bb

Dashboard:

Screen Shot 2019-04-04 at 5 35 57 PM

Also, it doesn't seem that the `per_page` parameter is being used - it's passed as `$rows` to each leaderboard method and then overwritten as the return variable. When I supply `after`, `before`, and `per_page` I get an empty array response.

includes/api/class-wc-admin-rest-leaderboards-controller.php
```diff
@@ -0,0 +1,481 @@
...
+	public function get_coupons_leaderboard( $rows, $after, $before, $persisted_query ) {
+		$coupons_data_store = new WC_Admin_Reports_Coupons_Data_Store();
+		$coupons_data       = $coupons_data_store->get_data(
+			array(
+				'orderby'       => 'orders_count',
+				'order'         => 'desc',
+				'after'         => $after,
+				'before'        => $before,
+				'extended_info' => true,
+			)
```
jeffstieler: Missing `per_page`?

includes/api/class-wc-admin-rest-leaderboards-controller.php
```diff
@@ -0,0 +1,481 @@
...
+	public function get_categories_leaderboard( $rows, $after, $before, $persisted_query ) {
+		$categories_data_store = new WC_Admin_Reports_Categories_Data_Store();
+		$categories_data       = $categories_data_store->get_data(
+			array(
+				'orderby'       => 'items_sold',
+				'order'         => 'desc',
+				'after'         => $after,
+				'before'        => $before,
+				'extended_info' => true,
+			)
```
jeffstieler: Missing `per_page`?

includes/api/class-wc-admin-rest-leaderboards-controller.php
```diff
@@ -0,0 +1,481 @@
...
+	 */
+	public function get_customers_leaderboard( $rows, $after, $before, $persisted_query ) {
+		$customers_data_store = new WC_Admin_Reports_Customers_Data_Store();
+		$customers_data       = $customers_data_store->get_data(
+			array(
+				'orderby' => 'total_spend',
+				'order'   => 'desc',
+				'after'   => $after,
+				'before'  => $before,
+			)
```
jeffstieler: Can this use `per_page`?

Also I'm not even sure the date range is being honored by `WC_Admin_Reports_Customers_Data_Store` - changing the date range doesn't seem to change the response.

includes/api/class-wc-admin-rest-leaderboards-controller.php
```diff
@@ -0,0 +1,481 @@
...
+	 */
+	public function get_products_leaderboard( $rows, $after, $before, $persisted_query ) {
+		$products_data_store = new WC_Admin_Reports_Products_Data_Store();
+		$products_data       = $products_data_store->get_data(
+			array(
+				'orderby'       => 'items_sold',
+				'order'         => 'desc',
+				'after'         => $after,
+				'before'        => $before,
+				'extended_info' => true,
```
jeffstieler: Missing `per_page`?

includes/api/class-wc-admin-rest-leaderboards-controller.php
```diff
@@ -0,0 +1,481 @@
...
+	}
+
+	/**
+	 * Return all leaderboards.
+	 *
+	 * @param  WP_REST_Request $request Request data.
+	 * @return WP_Error|WP_REST_Response
+	 */
+	public function get_items( $request ) {
+		global $wpdb;
```
jeffstieler: This looks to be unused.

joshuatf: Good eyes- this is no longer being used.

includes/api/class-wc-admin-rest-leaderboards-controller.php
joshuatf: `per_page` works, but you're right, the date range is not respected by the customers data store.

Do you think this is worth opening a separate issue to add in date querying?  The store joins the orders table, so it shouldn't add too much overhead.

joshuatf:
@jeffstieler Thanks for the review!  Fixed up those missing `per_page` arguments and gave this a rebase.  

Left a comment above about whether or not we should open an issue to address customer querying by date (seems like we should).

>Testing on the WCCOM data, the endpoint response isn't matching what I'm seeing on the dashboard:

Can you check your `before` and `after` params and let me know if you're still seeing this?  I could not recreate the issue, but I think the `before` date usually runs to end of day (e.g., `2018-08-01T23:59:59` if that makes any difference in your data.

includes/api/class-wc-admin-rest-leaderboards-controller.php
jeffstieler: > Do you think this is worth opening a separate issue to add in date querying? The store joins the orders table, so it shouldn't add too much overhead.

Definitely!

includes/api/class-wc-admin-rest-leaderboards-controller.php
```diff
@@ -49,18 +49,251 @@ public function register_routes() {
...
+				'after'         => $after,
+				'before'        => $before,
+				'per_page'      => $per_page,
+				'extended_info' => true,
+			)
+		);
+
+		$rows = array();
+		foreach ( $coupons_data->data as $coupon ) {
+			$coupon_url  = wc_admin_url( 'analytics/coupons?filter=advanced&coupon_includes=' . $coupon['coupon_id'] );
```
jeffstieler: Should these links include the date portions of the query as well like the dashboard components? See: https://github.com/woocommerce/wc-admin/blob/fc678aa68583bc7db4df5184d11aa03ca81b7b72/client/dashboard/leaderboards/top-coupons.js#L63

includes/api/class-wc-admin-rest-leaderboards-controller.php
```diff
@@ -49,18 +49,251 @@ public function register_routes() {
...
+			$coupon_code = $coupon['extended_info'] && $coupon['extended_info']['code'] ? $coupon['extended_info']['code'] : '';
```
jeffstieler: This looks a little too JavaScripty - won't this throw warnings if `extended_info` or `code` aren't present in the array? I think we'll need to use `isset()` here.

lib/common.php
```diff
@@ -295,3 +295,20 @@ function wc_admin_is_feature_enabled( $feature ) {
 	$features = wc_admin_get_feature_config();
 	return isset( $features[ $feature ] ) && true === $features[ $feature ];
 }
+
+/**
+ * Format a number using the decimal and thousands separator settings in WooCommerce.
+ *
+ * @param int $number Number to be formatted.
```
jeffstieler: Does this support floats as well?

jeffstieler:
> Fixed up those missing per_page arguments and gave this a rebase.

The `per_page` fix looks good. 👍 

> Can you check your before and after params and let me know if you're still seeing this? I could not recreate the issue, but I think the before date usually runs to end of day (e.g., `2018-08-01T23:59:59` if that makes any difference in your data.

That was the issue - I verified that the query issued by the dashboard ran to the end of the day, changed my API request and all is correct. Thanks!

includes/api/class-wc-admin-rest-leaderboards-controller.php
joshuatf: I think the lines note here are from an outdated version.  `$persisted_query` is passed to the URL here which will include date params.

includes/api/class-wc-admin-rest-leaderboards-controller.php
joshuatf: Good call 👍

lib/common.php
joshuatf: You probably wouldn't want to pass a float here since it removes all decimals, but it does support them.  Updated this comment to `mixed`.

includes/api/class-wc-admin-rest-leaderboards-controller.php
jeffstieler: Ah, I misunderstood that query params had to be duplicated in `persisted_query` to get passed to the URLs. 👍

jeffstieler (approved):
:shipit: ❗️

includes/api/class-wc-admin-rest-leaderboards-controller.php
timmyc: > This looks a little too JavaScripty 

never. you can never be too JavaScripty.

timmyc:
This. Is. Awesome.

QQ - And please don't let this block merge, but based upon the perf issues noted - do you think we should add some caching on these endpoints? Even if we do shave off the time in a follow-up PR

joshuatf:
@jeffstieler Thanks for reviewing this and getting it up to snuff!

@timmyc We discussed over zoom, but each of those individual data stores should be cached.  I suspect that something else may be happening to not hit my cache; will look into this and see if something is not working as expected here.

PR #1996, opened 2019-04-04
Fix date filter layout on Android

### LINKED ISSUE #1767: Android: bad UI when focusing inputs in date range picker
**To Reproduce**
1. Go to any report.
2. Open the date picker.
3. Switch to the _Custom tab_.
4. Focus on an input field.
5. Notice the UI looks bad.

**Screenshots**

**Smartphone:**
* Device: LG G6
* OS: Android 8.0
* Browser: Firefox 67.0a1 & Chrome

Aljullu:
Fixes #1767.

Refactors the date filter CSS so it's not broken on Android when the keyboard is open.

### Screenshots
_Before:_

_After:_

You will need an Android device to test it.
- Go to any report.
- Open the date picker and choose the _Custom_ tab.
- Focus on an input so the keyboard appears.
- Verify the layout is not broken.
- Verify there are no regressions when the keyboard is not open.

### DIFF
--- a/packages/components/src/filters/date/style.scss
+++ b/packages/components/src/filters/date/style.scss
@@ -13,6 +13,7 @@
 
 		.components-tab-panel__tab-content {
 			height: calc(100% - 46px);
+			overflow: auto;
 		}
 	}
 }
--- a/packages/components/src/filters/date/style.scss
+++ b/packages/components/src/filters/date/style.scss
@@ -12,7 +12,7 @@
 		}
 
 		.components-tab-panel__tab-content {
-			height: calc(100% - 36px);
+			height: calc(100% - 46px);
 		}
 	}
 }
@@ -90,11 +90,6 @@ button.woocommerce-filters-date__tab {
 
 	&.is-custom {
 		border-top: 1px solid $core-grey-light-700;
-	}
-
-	&.is-sticky-bottom {
-		position: absolute;
-		bottom: 0;
 	}
 }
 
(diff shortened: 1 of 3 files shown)


packages/components/src/filters/date/style.scss
```diff
@@ -91,11 +91,6 @@ button.woocommerce-filters-date__tab {
 	&.is-custom {
 		border-top: 1px solid $core-grey-light-700;
 	}
-
-	&.is-sticky-bottom {
-		position: absolute;
-		bottom: 0;
-	}
```
psealock: Looking on Chrome emulation mode, I see the bottom cut off:

Screen Shot 2019-04-05 at 3 44 27 PM

This was the issue I was trying to prevent with `is-sticky-bottom`. I do think your way is better though, its just that the contents aren't scrollable (at least on Chrome)

Aljullu: Good catch with that @psealock!

> Looking on Chrome emulation mode, I see the bottom cut off:

I imagine we have to make a trade-off for small screens: or we show the bottom cut off or overlapping the calendar.

I made the datepicker scrollable in 6cdd74df3ccfc0ff55fc5e80704bfaa7d5ed4179, I think that's a good enough solution, but let me know what you think.

psealock (approved):
LGTM, thanks for tackling this one @Aljullu

timmyc:
Nice fix Albert!

PR #1999, opened 2019-04-04
Fix withSelect to release resources on unmount

coderkevin:
Previously, withSelect wasn't releasing resources from fresh-data when
it was unmounting. This fixes that behavior and now clears a component's
requirements correctly when unmounting.

1. Use `npm start` for dev build.
2. Go to any analytics page
3. Open JS console and inspect `freshData.wcApi`
4. Click the "eye" icon to add freshData.wcApi.summary as a "live expression", which stays in place even after refreshes. You can then watch the summary change as resources are fetched and received.
5. Click through all the different reports, the number of components should not climb above 6 or 9 at a time, and the number of resources no longer required should be much higher

DevTools_-_two_wordpress_test_wp-admin_admin_php_page_wc-admin

### DIFF
--- a/client/wc-api/wc-api-spec.js
+++ b/client/wc-api/wc-api-spec.js
@@ -13,6 +13,7 @@ import user from './user';
 
 function createWcApiSpec() {
 	return {
+		name: 'wcApi',
 		mutations: {
 			...settings.mutations,
 			...user.mutations,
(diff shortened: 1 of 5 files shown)


Aljullu:
@coderkevin I tried following the test instructions but I get `freshData is not defined`. Do I need to do something to get this value exposed?
image

Aljullu (approved):
This is testing well and I can confirm the number of components is much lower than before.

PR #2000, opened 2019-04-04
Split D3Chart utils/axis file

Aljullu:
Splits `packages/components/src/chart/d3chart/utils/axis.js` in two files: `axis-x.js` and `axis-y.js`.

`axis.js` was already over 300 lines of code, and considering each axis is independent of the other and they don't share any function, I think it will make work easier if they are in two different files.

This PR doesn't add or remove any line of code other than some imports, so running `npm test` and checking that charts still render ok should be enough.


psealock (approved):
Tests are passing and the chart is rendering with no issues. Nice improvement here.

timmyc:
Nice refactor Albert 👍

Issue #1677, opened 2019-02-22
Analytics Copy Cleanup


LevinMedia:
As it stands, some of the data is labeled one way in one report, and another way in another report. The reality is not as bad as that last sentence sounds, it’s mostly situations where we have a label like “Orders” in one place and “Orders Count” in another. These labels should be as short and concise as possible, so in that case, “Orders” would be the preferred label.

This list should serve as a source of truth, if anyone has suggestions for relabeling, please provide feedback here. This list should be applied to all reports, and the dashboard. 

- **Gross Revenue (G. Revenue, G. Rev):**  All the money
- **Net Revenue (N. Revenue, N. Rev):**  All the money – Refunds – Coupons – Total Tax – Shipping
- **Orders**: A count of orders
- **Products Sold:** A count of products sold
- **Items Sold:** A count of individual items sold
- **Average Order Value (Avg. Order Value, AOV):** Net Revenue / Orders
- **Average Items Per Order (Avg. Items Per Order):** Items Sold / Orders
- **Refunds:** A monetary value for refunds issued
- **Refunded Orders:** A count of refunded orders
- **Coupons:** A monetary value for coupons used
- **Discounted Orders:** A count of orders that included at least one coupon
- **Total Tax:** A monetary value for Order Tax + Shipping Tax
- **Order Tax:** A monetary value for taxes collected on an order
- **Shipping Tax:** A monetary value for taxes collected on shipping charges
- **Shipping:** A monetary value for shipping charges
- **Downloads:** A count of downloaded products
- **Customers:** a count of people who have made a purchase and or registered an account with a store. Tracked by Billing email. (The new reports count all users of a store as a “Customer”, not just people who have register an account)
- **Total Spend:** The total amount a customer has spent at a store

In a convo with @timmyc - he had suggested making the labels driven by the API, similar to what we did in the the performance endpoint. That way we can update labels in one spot, and it _**magically**_ changes everywhere.

dreiris:
- Just do a find/replace. 
- Assign David as PR reviewer

PR #2007, opened 2019-04-05
Keep border color and spacing consistent between summary items

dechov:
Slight tweaks to summary item grid, around inner border color and item height:

https://github.com/woocommerce/woocommerce-admin/commit/9cfda69870614088841c7b3d9670159bb0628549:
- switches items' bottom border color to `$inner-border`
- overrides this to `$outer-border` for the last row of the grid (with the help of a tip [here](https://keithclark.co.uk/articles/targeting-first-and-last-rows-in-css-grid-layouts/))
- removes forced `$outer-border` for last item's right border. In an unbalanced grid this counts as an inner border, as I understand – cc @LevinMedia
- …replacing it with the `$outer-border` bottom border, for proper styling on the 1-cell wide grid
- moves styles applicable only to wide viewport to a new media query, rather than configuring for wide viewport and overriding / resetting – FYI @ryelle in case this refactor is not valid

https://github.com/woocommerce/woocommerce-admin/commit/e3980bca045473a8d3b9407c301b1dcd112f937f sets a fixed height for the item label so that even when it wraps, the grid stays intact. The vertical spacing / leading is also adjusted to improve the display when wrapping to two lines, though unfortunately more than two lines would result in overlapping text – not sure if this is an issue.

### Screenshots

Before | After
-- | --
 | 

On Dashboard and Analytics pages, summary items should look just as before, at all viewport sizes, except that all inner borders between summary items should be lighter than the outer border.

Also, at the certain viewport width when the summary item label wraps, the item height should remain the same height as neighboring items – tested with "AMOUNT OF DISCOUNTED ORDERS".


LevinMedia:
This is great stuff @dechov thank you. 

FWIW - the text wrapping should be less of an issue when we we get around to optimizing the labels across the entire experience - which has been noted in https://github.com/woocommerce/woocommerce-admin/issues/1677 but we may yet encounter long labels, so I appreciate that addition.

LevinMedia:
Heya @joshuatf while it's entirely possible that an extension might write a label that is three lines, I wouldn't worry about it too much. These labels should be as brief as possible. I think we should be encouraging brevity in our extension integration docs rather than flexing our UI to accommodate unreasonably long labels.

joshuatf:
@LevinMedia I think encouragement towards shorter summary labels is a good idea, but in reality it may be pretty hard to enforce any of those standards.

Is there a good reason not to use flexbox option pasted above?  It continues to work for the existing layout and if a summary label does happen to hit 3 lines it will allow it to continue to work without overlapping text and breaking the summary number layout.  Or is your thinking that devs will see the broken layout and attempt to reduce their labels character count?

joshuatf:
Hey @dechov 👋 - just wanted to check in and see if you had time to add the above flexbox css.

dechov:
@joshuatf Thanks for the ping. I think the flexbox solution is good, and I can revert https://github.com/woocommerce/woocommerce-admin/commit/e3980bca045473a8d3b9407c301b1dcd112f937f in favor of your change – @LevinMedia let us know if you do in fact prefer it to not look right when it's too long.

Not as sure about the `margin-top: auto;` on the element below.

With it | Without it
-- | --
Screen Shot 2019-04-19 at 3 53 50 PM | Screen Shot 2019-04-19 at 3 56 54 PM

Thoughts? Maybe we can still adjust the spacing?

timmyc:
@dechov will you be able to spend some more time on this so we can get it merged in?

mikejolley:
Added those flexbox rules and tested everything at various widths, also with silly long titles. All works fine and has my 👍 

This is synced with master now also. @timmyc Happy to merge?

packages/components/src/summary/style.scss
```diff
@@ -279,7 +279,9 @@ $inner-border: $core-grey-light-500;
 }
 
 .woocommerce-summary__item {
-	display: block;
+	display: flex;
+    flex-direction: column;
```
dechov: These lines need tab instead of spaces – I see I have the same in my own working copy, due to the copy/paste.

dechov:
@timmyc Sorry, was AFK and missed the ping. At the time I was actually uncertain and awaiting feedback on whether we wanted the whole row of numbers to shift downwards and stay aligned, but I have no strong opinion either way – it seems fine that way.

mikejolley:
@dechov @timmyc I added the rule to align the numbers for now. 

Linting/tests are passing now so I'm marking as approved so @timmyc can merge if happy :)

Issue #2006, opened 2019-04-05
Activity Panel: Notes timezone issues


justinshreve:
The notes feature does not seem to correctly account for timezones.

The timezone in the database is stored as GMT/UTC.

<img width="531" alt="Screen Shot 2019-04-05 at 10 38 37 AM" src="https://user-images.githubusercontent.com/689165/55636570-1b376380-5791-11e9-84db-02a914495106.png">

To reproduce:
* Pick any note and update the `date_created` value a note in your `wp_wc_admin_notes` table to the current time in UTC. Example, It is currently 14:50 UTC (10:50 AM): `2019-04-05 14:40:10`. You can also test the mobile app note here: #2008.
* Set your browser time to anything other than UTC.
* Refresh the inbox panel, and see the time is wrong. For example, EST shows "4 HOURS FROM NOW".

Aljullu:
@jeffstieler did some changes regarding this in #1872, so this issue might be fixed when that PR is merged.

justinshreve:
🤦‍♂️ Thanks for linking that @Aljullu! Assigning this to you @jeffstieler if you don't mind, since you already have a fix :).

Aljullu:
Now that #1872 was merged, this should be fixed.

timmyc:
While working on #2677 I experienced this bug again, so not sure if it was fixed fully the first time, or perhaps we had a regression sneak in somewhere between here and there.

opr:
Howdy folks, I am seeing this on my local development site and JN. The timezone on my site is set to "London", and when I add a note I see "an hour from now". When the site is UTC+0 it shows correctly.

I also noticed this also happens when creating a new JN site, after setting it up change the site's timezone to something other than UTC+0 and then install WooCommerce.

![image](https://user-images.githubusercontent.com/5656702/132385006-e35ed91b-73f8-4b58-8c3a-2782e2a952f7.png)

If it's helpful, here's the code I'm using to add the note.

```php
$note = new Note();
		$note->set_title(
			__(
				'Introducing the Cart and Checkout blocks!',
				'woo-gutenberg-products-block'
			)
		);
		$note->set_content(
			__(
				"Increase your store's revenue with the conversion optimized Cart & Checkout WooCommerce blocks available in the WooCommerce Blocks extension.",
				'woo-gutenberg-products-block'
			)
		);
		$note->set_content_data(
			(object) array(
				'getting_started' => true,
			)
		);
[...]

PR #2008, opened 2019-04-05
Activity Panel: Add a mobile app note

### LINKED ISSUE #1980: Inbox: Woo Mobile App Nudge
From the Notes backlog, this new Note is to encourage users to checkout the new Woo Mobile Apps.

**Note Details**

Item | Details
-----|-------
Type | Info
Title | Install Woo mobile app
Message | Install the WooCommerce mobile app to manage orders, receive sales notifications, and view key metrics — wherever you are.
Store Type | New & Existing
Trigger | Day 2
Cadence | once
CTA Label | Learn more
CTA Target | https://woocommerce.com/mobile/

emdashcodes:
Fixes #1980.

This PR adds a nudge for installing the mobile app. It displays two days after setting up WooCommerce Admin.

There is a small bug present with the notes where the timestamp may display wrong. See https://github.com/woocommerce/woocommerce-admin/issues/2006.

### Screenshots

Screen Shot 2019-04-05 at 10 38 37 AM

* Install the `WP Crontrol` plugin.
* You either need to deactivate/uninstall WooCommerce and reinstall (to get the correct option set), or repeat the following step twice.
* Go under `Tools > Cron Events` and run `wc_admin_daily`. (You may need to do this twice since the note is triggered based on number of days since install, and we are depending on an option to be set).
* Refresh the Inbox and see the mobile app note.

### DIFF
--- a/woocommerce-admin.php
+++ b/woocommerce-admin.php
@@ -98,6 +98,7 @@ function wc_admin_build_file_exists() {
  */
 function wc_admin_do_wc_admin_daily() {
 	WC_Admin_Notes_New_Sales_Record::possibly_add_sales_record_note();
+	WC_Admin_Notes_Mobile_App::possibly_add_mobile_app_note();
 }
 add_action( 'wc_admin_daily', 'wc_admin_do_wc_admin_daily' );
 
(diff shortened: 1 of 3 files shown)


woocommerce-admin.php
```diff
@@ -98,6 +98,7 @@ function wc_admin_build_file_exists() {
  */
 function wc_admin_do_wc_admin_daily() {
 	WC_Admin_Notes_New_Sales_Record::possibly_add_sales_record_note();
+	WC_Admin_Notes_Mobile_App::possibly_add_mobile_app_note();
```
joshuatf: Should we make this a filterable list?

joshuatf (approved):
Looks good and testing well, Justin!

One comment about extensibility, but otherwise LGTM.

Also noted an issue where the unread indicator did not come on, but I believe that's a separate bug that I may have introduced.  Opened an issue for that [here](https://github.com/woocommerce/woocommerce-admin/issues/2012).

woocommerce-admin.php
justinshreve: I think the `wc_admin_daily' action (see a couple lines down) covers the extendability usecase for us. These two methods can be unhooked, and new ones can be hooked in.

PR #2013, opened 2019-04-08
Fix wrong dates in Reviews and Orders panels of the Activity Panel

### LINKED ISSUE #1890: Activity Panel: 'Invalid Date' appearing in the reviews card
Reported by @LevinMedia [here](https://github.com/woocommerce/woocommerce-admin/pull/1880#pullrequestreview-217823014):

> I'm also getting an "Invalid date" message for the review timestamp
> Screen Shot 2019-03-22 at 8 23 12 AM

Aljullu:
Fixes #1890 and some other issues with the date handling of the Activity Panel.

### Screenshots
_Before:_
Screenshot from 2019-04-08 15-17-07

_After:_
image

First, you will need the fix from #1872, so in order to test this, run `git merge add/1831-order-milestone-notifications`.
- Change your store timezone to UTC+14 or something similar.
- Create a new note/order/review.
- Verify the _Unread_ notification appears on the corresponding tab.
- Open the activity panel and verify the corresponding element is highlighted.

### DIFF
--- a/client/header/activity-panel/panels/inbox.js
+++ b/client/header/activity-panel/panels/inbox.js
@@ -90,7 +90,11 @@ class InboxPanel extends Component {
 								title={ note.title }
 								date={ note.date_created }
 								icon={ <Gridicon icon={ note.icon } size={ 48 } /> }
-								unread={ ! lastRead || new Date( note.date_created_gmt ).getTime() > lastRead }
+								unread={
+									! lastRead ||
+									! note.date_created_gmt ||
+									new Date( note.date_created_gmt + 'Z' ).getTime() > lastRead
+								}
 								actions={ getButtonsFromActions( note.actions ) }
 							>
 								<span dangerouslySetInnerHTML={ sanitizeHTML( note.content ) } />
--- a/client/header/activity-panel/unread-indicators.js
+++ b/client/header/activity-panel/unread-indicators.js
@@ -21,7 +21,8 @@ export function getUnreadNotes( select ) {
 		! Boolean( getNotesError( notesQuery ) ) &&
 		! isGetNotesRequesting( notesQuery ) &&
 		latestNote[ 0 ] &&
-		new Date( latestNote[ 0 ].date_created_gmt ).getTime() > userData.activity_panel_inbox_last_read
+		new Date( latestNote[ 0 ].date_created_gmt + 'Z' ).getTime() >
+			userData.activity_panel_inbox_last_read
 	);
 }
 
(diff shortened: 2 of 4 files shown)


client/header/activity-panel/panels/inbox.js
```diff
@@ -90,7 +90,11 @@ class InboxPanel extends Component {
 								title={ note.title }
 								date={ note.date_created }
 								icon={ <Gridicon icon={ note.icon } size={ 48 } /> }
-								unread={ ! lastRead || new Date( note.date_created_gmt ).getTime() > lastRead }
+								unread={
+									! lastRead ||
+									! note.date_created_gmt ||
+									new Date( note.date_created_gmt + 'Z' ).getTime() > lastRead
```
coderkevin: Optional: Is there a better way to check this than creating a date object every time?

client/header/activity-panel/unread-indicators.js
```diff
@@ -21,7 +21,8 @@ export function getUnreadNotes( select ) {
 		! Boolean( getNotesError( notesQuery ) ) &&
 		! isGetNotesRequesting( notesQuery ) &&
 		latestNote[ 0 ] &&
-		new Date( latestNote[ 0 ].date_created_gmt ).getTime() > userData.activity_panel_inbox_last_read
```
coderkevin: I'm personally not a big fan of long, complex boolean checks. Would it be possible to simplify some of these into more descriptive intermediate variables at least? I think this return statement would be pretty confusing for a developer new to this code.

coderkevin (commented):
Looks good, just a couple of comments related to the code around this.

Also, I'd like to suggest setting this PR to be based on `add/1831-order-milestone-notifications` instead of `master` which would both eliminate the need to pre-merge that branch before testing this one, and it would ensure this PR doesn't get merged into master before that one.

coderkevin (approved):
Much clearer, thanks!

client/header/activity-panel/panels/inbox.js
Aljullu: Not that I can think of. We get the date as a string from the endpoint in the form of `2019-03-30T15:28:12`, so we need to parse it in some way to be able to compare it.

timmyc:
Thanks for fixing the time traveling notifications Albert.

Aljullu:
Thanks for the reviews @coderkevin @timmyc!

> Also, I'd like to suggest setting this PR to be based on `add/1831-order-milestone-notifications` instead of `master` which would both eliminate the need to pre-merge that branch before testing this one, and it would ensure this PR doesn't get merged into master before that one.

Now that #1872 has been merged into `master`, this shouldn't be necessary anymore. :+1:

PR #2019, opened 2019-04-09
Fix-#1982

### LINKED ISSUE #1982: Inbox: Giving Feedback Notice
This recurring note will encourage early adopters of wc-admin to give some feedback

**Note Details**

Item | Details
-----|-------
Type | Info
Title | Giving feedback
Message | Are you enjoying the new WooCommerce experience? We'd love to get your feedback.
Store Type | New & Existing
Trigger | Day 3 (after plugin install)
Cadence | every other week
CTA Label | Share feedback
CTA Target | https://github.com/woocommerce/woocommerce-admin/issues/new/choose

Mr-AjayM:
Fixes #1982 

_Added inbox note for admin giving feedback notice after three day of install wc admin._

### Accessibility

- [X] I've tested using only a keyboard (no mouse)

### Screenshots
Screenshot_1


includes/class-wc-admin-install.php
```diff
@@ -72,6 +72,7 @@ public static function install() {
 
 		delete_transient( 'wc_admin_installing' );
 
+        update_option( 'wc_admin_install_timestamp', time() );
```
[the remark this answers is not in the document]
Mr-AjayM: @Aljullu I updated the changes as you have mentioned. Please do let me know if we need to do update anything else. Thanks 🙂

timmyc:
> Cadence of every other week, but I'm not sure how this should be implemented considering notices are currently not dismissable. @timmyc do you have any input here?

Indeed that is a new feature of notes we have yet to explore fully. I imagine the way to go about this is to update the date of the note to bump it back to the top of the inbox.

justinshreve:
A cadence of every other week seems like it would get annoying *without* a dismiss option (I'm thinking of iOS "rate us please" popups. Should we have the button mark the note as actioned to at least prevent this from repeating for users who clicked to provide feedback?

Aljullu:
Thanks for the fixes @Mr-AjayM!

I created an [issue](#2062) to implement the cadence in a follow-up, so we can already merge this PR. We just need to resolve the conflicts, will you be able to take care of this? Thank you!

Mr-AjayM:
Hello @Aljullu,

Thanks for replay! I fixed the conflicts, Can you please do verify it and let me know if we need to do anything else before merging this.

Thanks a lot,
Aj

woocommerce-admin.php
```diff
@@ -153,7 +153,9 @@ function wc_admin_devdocs() {
  */
 function wc_admin_do_wc_admin_daily() {
 	WC_Admin_Notes_New_Sales_Record::possibly_add_sales_record_note();
+	WC_Admin_Notes_Giving_Feedback_Notes::add_notes_for_admin_giving_feedback();
 	WC_Admin_Notes_Mobile_App::possibly_add_mobile_app_note();
+
```
Aljullu: Nitpicking here. I think we can remove this empty line.
```suggestion
```

woocommerce-admin.php
```diff
@@ -207,9 +209,11 @@ function wc_admin_plugins_loaded() {
 	require_once WC_ADMIN_ABSPATH . 'includes/class-wc-admin-install.php';
 	require_once WC_ADMIN_ABSPATH . 'includes/class-wc-admin-api-init.php';
 
+	// Admin note providers.
 	// @todo These should be bundled in the features/ folder, but loading them from there currently has a load order issue.
 	require_once WC_ADMIN_ABSPATH . 'includes/notes/class-wc-admin-notes-new-sales-record.php';
 	require_once WC_ADMIN_ABSPATH . 'includes/notes/class-wc-admin-notes-settings-notes.php';
+	require_once WC_ADMIN_ABSPATH . '/includes/notes/class-wc-admin-notes-giving-feedback-notes.php';
```
Aljullu: Can you remove the leading `/`?
```suggestion
	require_once WC_ADMIN_ABSPATH . 'includes/notes/class-wc-admin-notes-giving-feedback-notes.php';
```

Mr-AjayM:
Hi, @Aljullu Thanks for the suggestions! I have fixed that.

Thank you 🙂

PR #2022, opened 2019-04-09
Add PHP autofix to pre-commit hook

### LINKED ISSUE #2021: PHP Linter autofix should be run as a pre-commit hook
1. Checkout master: `git checkout master`.
2. Run PHP Lint autofix: `npm run lint:php-fix`.
3. See what changed: `git status`.
4. Notice we have several files with issues.
5. We should run PHP Lint autofix as a pre-commit hook to prevent this happening in the future.

Aljullu:
Fixes #2021.

Add PHP Lint autofix to pre-commit hook.

- Edit any PHP file and add something like this:
```PHP
if ( 1 == 1 ) {
   return; // Spaces instead of tabs
}
```
- With git, add that file and do a commit.
- Notice the spaces instead of tabs issue if automatically fixed.
- Notice not adding a period in the comment prevents you from committing and you manually have to fix that issue.

### DIFF
--- a/bin/pre-commit-hook.js
+++ b/bin/pre-commit-hook.js
@@ -102,7 +102,7 @@ files.forEach( file => {
 
 	try {
 		// Apply auto-fix
-		execSync( `./vendor/bin/phpcbf --standard=phpcs.xml.dist ${ file } > /dev/null` );
+		execSync( `./vendor/bin/phpcbf --standard=phpcs.xml.dist ${ file }` );
 	} catch( e ) {
 		try {
 			// Check if there are still errors
--- a/bin/pre-commit-hook.js
+++ b/bin/pre-commit-hook.js
@@ -91,30 +91,43 @@ if ( lintResult.status ) {
 }
 
 // PHP Lint.
-let hasPhpLintErrors = false;
+let phpLintErrors = '';
 let phpFiles = '';
 
 files.forEach( file => {
+	let fileHasPhpLintErrors = false;
 	if ( ! file.endsWith( '.php' ) ) {
 		return;
 	}
 
 	try {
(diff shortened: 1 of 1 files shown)


psealock (commented):
I get a different message when committing your code snippet

Screen Shot 2019-04-10 at 9 53 08 AM

When I remove the comment, the commit is allowed to happen. `npm run lint:php` does catch the errors.

Screen Shot 2019-04-10 at 9 54 59 AM

Is this not what you're experiencing?

Aljullu:
Thanks for the review @psealock! You're right, my example was wrong. Strict comparisons is a warning instead of an error, so it doesn't block commiting. I think that's to be expected and we are currently not catching warnings either.

I updated the steps to reproduce and added some code to check if PHP_CodeSniffer is installed. Feel free to take another look. :+1:

bin/pre-commit-hook.js
```diff
@@ -91,30 +91,55 @@ if ( lintResult.status ) {
...
 	try {
-		execSync( `php -l -d display_errors=0 ${ file }` );
-	} catch( err ) {
-		hasPhpLintErrors = true;
+		// Check if PHP_CodeSniffer is installed.
+		execSync( `./vendor/bin/phpcbf -h` );
+		execSync( `./vendor/bin/phpcs -h` );
+	} catch( e ) {
+		console.log(
+			'PHP_CodeSniffer is not installed. ' +
```
joshuatf: ❤️

joshuatf (approved):
Sweet.  Testing well for fixable phpcbf items.  LGTM if @psealock doesn't have any other concerns.

psealock (approved):
Oh right, I didn't see those were warnings. LGTM

PR #2024, opened 2019-04-09
Reference correct roadmap blogpost

seagyn:
Fixes #

Links to the correct blog post showing release notes for the latest version. Would be nice to be able to link to a category or tag dedicated to WooCommerce Admin instead of having to update this every time.


seagyn:
@Aljullu Awesome!

PR #2026, opened 2019-04-09
Update Link in readme to Proper Location

timmyc:
Quick change of some copy/paste that was pointing the release notes at an old blocks post.


Aljullu (approved):
Good catch. Looks good to me!

PR #2018, opened 2019-04-09
Try: Extension Examples

psealock:
Create an environment to develop example extension plugins that can be activated, tinkered with, and viewed locally. This makes for an ideal setup to test and develop hooks and filters.

These changes introduce a simple "Hello World" report extension with a build step that places the plugin files including webpacked js code in your plugins folder.

### Screenshots

Screen Shot 2019-04-09 at 3 13 06 PM

### Detailed test instructions:

1. `npm run example -- --ext=add-report`.
2. Activate the WooCommerce Admin Add Report Example plugin
3. Visit `/wp-admin/admin.php?page=wc-admin#/analytics/example` to see the report.
4. Change the javascript in `docs/examples/extensions/add-report/js/index.js` and make sure the report reflects those changes.
5. Make any change in a PHP file, docs/examples/extensions/add-report/woocommerce-admin-add-report-example.php`, and see the change reflected in the plugin.

### Alternative

We could also house extension examples in a separate repo such as https://github.com/psealock/woocommerce-admin-extension-examples. While this alleviates the need to copy files outside of the `wc-admin` repo, it requires its own build step and package.json while requiring integrations with our team's current communications, such as Slack and notifications. Plus, the examples may get outdated and unhelpful if they lack attention from being separate.

~One drawback to the examples-in-repo approach here is the `npm run example -- add-report` command needs to be re-run everytime PHP changes are made in order to see them. A solution to this is to establish a symlink between folders in this repo and the ones in the plugins folder. This may introduce a whole new kettle of fish, so I'm open to suggestions here.~

EDIT: You can now make changes to PHP files and see them reflected in the plugins folder. Webpack now copies the whole folder on any changes.

PR #2033, opened 2019-04-10
Extend report submenu items

psealock:
Allow extensions to add a submenu item to the sidebar 

Screen Shot 2019-04-10 at 4 13 31 PM

This PR is also an example of how https://github.com/woocommerce/woocommerce-admin/pull/2018 could be useful in developing and testing extensibility because the code and example reside in the same pull request.

1. `npm run example -- --ext=add-report`.
1. Activate the WooCommerce Admin Add Report Example plugin.
2. Open wc-admin and see the "Example" report menu item appear in the sidebar.
3. Make sure the link works.

### DIFF
--- a/client/analytics/report/index.js
+++ b/client/analytics/report/index.js
@@ -33,7 +33,7 @@ import ReportError from 'analytics/components/report-error';
 import { searchItemsByString } from 'wc-api/items/utils';
 import withSelect from 'wc-api/with-select';
 
-const REPORTS_FILTER = 'woocommerce-reports-list';
+const REPORTS_FILTER = 'woocommerce_admin_reports_list';
 
 const getReports = () => {
 	const reports = [
--- a/docs/examples/extensions/README.md
+++ b/docs/examples/extensions/README.md
@@ -1,0 +1,25 @@
+# WooCommerce Admin Extension Examples
+
+Examples for extending WooCommerce Admin
+
+## Directions
+
+Install dependencies, if you haven't already.
+
+```bash
+npm install
+```
+
+Build the example extension by running the npm script and passing the example name.
+
+```bash
+npm run example -- --ext=<example>
+```
+
+Go to your Wordpress installation's plugins page and activate the plugin. WooCommerce Analytics reports will now reflect the changes made by the example extension.
+
+You can make changes to Javascript and PHP files in the example and see changes reflected upon refresh.
+
+## Example Extensions
+
+`add-report` - Create a "Hello World" report page.
--- a/docs/examples/extensions/add-report/js/index.js
+++ b/docs/examples/extensions/add-report/js/index.js
@@ -1,0 +1,38 @@
+/**
+ * External dependencies
+ */
+import { addFilter } from '@wordpress/hooks';
+import { Fragment } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+
+/**
+ * WooCommerce dependencies
+ */
+import { ReportFilters } from '@woocommerce/components';
+
+const Report = ( { path, query } ) => {
+	return (
+		<Fragment>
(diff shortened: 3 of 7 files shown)


lib/admin.php
```diff
@@ -92,87 +92,70 @@ function wc_admin_register_pages() {
...
+			),
 			array(
 				'title'  => __( 'Settings', 'woocommerce-admin' ),
 				'parent' => '/analytics/revenue',
 				'path'   => '/analytics/settings',
-			)
+			),
 		);
+
+		if ( 'yes' === get_option( 'woocommerce_manage_stock' ) ) {
```
joshuatf: Optional nitpick: maybe we could add this to the above array and check if it not null before registering the page?  If someone were to register there page at the beginning of the array or remove pages, we might lose ordering here.

```php
			'yes' === get_option( 'woocommerce_manage_stock' ) ? array(
				'title'  => __( 'Stock', 'woocommerce-admin' ),
				'parent' => '/analytics/revenue',
				'path'   => '/analytics/stock',
			) : null,
```

lib/admin.php
```diff
@@ -92,87 +92,70 @@ function wc_admin_register_pages() {
...
+				'title'  => __( 'Stock', 'woocommerce-admin' ),
+				'parent' => '/analytics/revenue',
+				'path'   => '/analytics/stock',
+			);
+
+			array_splice( $report_pages, 7, 0, array( $stock_page ) );
+
+		}
+
+		$report_pages = apply_filters( 'woocommerce-admin-report-menu-items', $report_pages );
```
joshuatf: Optional nitpick: We could use snake case here to keep filter names consistent.

```suggestion
		$report_pages = apply_filters( 'woocommerce_admin_report_menu_items', $report_pages );
```

joshuatf (approved):
This is pretty awesome, Paul!  Added a couple optional nitpicks, but this is testing well and LGTM.

>npm run example -- --ext=add-report

I really enjoy this method of handling things.  Playing devil's advocate though, I can see how this might act as a barrier to entry to a developer that just wants to add a "quick" php page and isn't quite used to the npm workflow.

I don't think that's a deal breaker though as they can copy/paste the plugin if they need to.  Just wanted to give a little feedback on that since I know this still isn't finalized in #2018.

psealock:
Thanks for the feedback @joshuatf 

>I can see how this might act as a barrier to entry to a developer that just wants to add a "quick" php page and isn't quite used to the npm workflow.

I did think about checking in compiled `dist` code so that you can copy and paste the repo if you'd like for reasons like you mention. I don't think its worth it though, and this is a gentle introduction 😄 to npm.

joshuatf:
>I don't think its worth it though, and this is a gentle introduction 😄 to npm.

I agree and really like the simple npm command to get things going.  Awesome work on that!

lib/admin.php
```diff
@@ -92,87 +92,66 @@ function wc_admin_register_pages() {
...
+			),
 			array(
 				'title'  => __( 'Settings', 'woocommerce-admin' ),
 				'parent' => '/analytics/revenue',
 				'path'   => '/analytics/settings',
-			)
+			),
 		);
+
+		$report_pages = apply_filters( 'woocommerce_admin_report_menu_items', $report_pages );
```
ryelle: This will also let folks remove all the report pages if they want, is that OK/expected? Should any pages be excluded from this list so they can't be unhooked?

joshuatf: I think this is okay.  If a dev really wanted to remove one of these pages, they could do this without the filter.

psealock: From internal conversations, it seems trying to prevent this kind of misuse, which can be achieved via simple jQuery anyways, is futile.

Issue #1850, opened 2019-03-19
Add historical data import endpoint


jeffstieler:
![Screen Shot 2019-03-19 at 4 57 23 PM](https://user-images.githubusercontent.com/63922/54647441-17ed6980-4a68-11e9-9029-645a92e43f9c.png)

The historical data import tool (above) requires endpoints for the following functionality:

_(It's probably best to split these items into separate PRs)_

- [ ] Updating the Customer and Order counts when a backfill date is chosen or "skip previously imported" is toggled
- [x] Initiating a historical data import #2034 
  * Params: # of days to backfill, whether to skip existing customers/orders
  * Persist the parameters for use in `WC_Admin_Reports_Sync` queries
- [ ] Querying the status of a historical data import
  * Status of current import if in progress - importing customers, importing orders
  * Date of oldest Order imported (used to show how much data has been imported)
- [x] Cancelling an in-process historical data import #2048
- [x] Deleting all previously imported data #2057 

See: p6riRB-3ZT-p2

joshuatf:
>Status of current import if in progress - importing customers, importing orders

@jeffstieler It's technically possible to run 2 imports simultaneously, right?  Do you think we should provide stats on all running imports or try and limit to one import at a time?

/cc @LevinMedia

jeffstieler:
> It's technically possible to run 2 imports simultaneously, right? Do you think we should provide stats on all running imports or try and limit to one import at a time?

I suppose that would be technically possible, but I don't think we want to allow/support that. My vote is to limit to one import at a time.

joshuatf:
>I suppose that would be technically possible, but I don't think we want to allow/support that. My vote is to limit to one import at a time.

I think that's probably the best solution as well.

>Updating the Customer and Order counts when a backfill date is chosen or "skip previously imported" is toggled

These are in reference to the import status, right?  Should we be adding these to the status response or do you think there's benefit added by separating these into a different response?

>Date of oldest Order imported (used to show how much data has been imported)

What were your thoughts on the latest import date?  I don't think we can safely determine number of imports from this since they could skip over a period of imports, but maybe you have something else in mind for this date.

/cc @jeffstieler

jeffstieler:
> > Updating the Customer and Order counts when a backfill date is chosen or "skip previously imported" is toggled
> 
> These are in reference to the import status, right? Should we be adding these to the status response or do you think there's benefit added by separating these into a different response?

[...]

PR #2034, opened 2019-04-10
Add import endpoint and controller

joshuatf:
Fixes (part of) #1850

* Adds the new import endpoint
* Allows regenerating data by number of days and skipping existing records
* Hooks up the settings page regenerate call to the new endpoint

### Thoughts / Questions

* ~This PR removes `customer_lookup_batch_init()` and relevant dependent actions.  It seems like this is safe since customers are now processed during orders, but want to double check.  This also needs to be removed to see benefit of the speed improvements by only processing days or unprocesse orders.  /cc @jeffstieler~
* Should the regenerate query with days/skip options be placed in options for retrieval instead of passed down through multiple functions?  I'm thinking passing as query_args is safe enough, but open to suggestions here.

### Screenshots
Screen Shot 2019-04-11 at 11 17 13 AM

1. Make a POST request to `wp-json/wc/v4/reports/import`.
2. Check that your data regeneration has scheduled actions.
3. Pass the following params to the API and test that it limits the number of orders processed.
  - `days` (int) - Number of days to process prior to today.
  - `skip_existing` (bool) - Skips existing records if already imported.
4. Make sure all tests pass.
5.  Make sure customer data is still updated correctly on order import.
6. Make sure the import option at `wp-admin/admin.php?page=wc-admin#/analytics/settings` still works as expected.


joshuatf:
I think tests are failing due to the change in `WC_Helper_Queue::run_all_pending();`.  Will take a look at that tomorrow and revert or change expectations if necessary.

jeffstieler:
@joshuatf 

> This PR removes `customer_lookup_batch_init()` and relevant dependent actions. It seems like this is safe since customers are now processed during orders, but want to double check. This also needs to be removed to see benefit of the speed improvements by only processing days or unprocesse orders

I'll test it now, but at first glance I think this will break the "is returning" logic and excludes customers who have registered but not placed an order (not sure how prevalent that is, but I think that's handled by the current login in `master`).

jeffstieler:
@joshuatf 

> This PR removes `customer_lookup_batch_init()` and relevant dependent actions. It seems like this is safe since customers are now processed during orders, but want to double check. This also needs to be removed to see benefit of the speed improvements by only processing days or unprocesse orders

I'll test it now, but at first glance I think this will break the "is returning" logic and excludes customers who have registered but not placed an order (not sure how prevalent that is, but I think that's handled by the current login in `master`).

jeffstieler:
@joshuatf 

> This PR removes `customer_lookup_batch_init()` and relevant dependent actions. It seems like this is safe since customers are now processed during orders, but want to double check. This also needs to be removed to see benefit of the speed improvements by only processing days or unprocesse orders

I'll test it now, but at first glance I think this will break the "is returning" logic and excludes customers who have registered but not placed an order (not sure how prevalent that is, but I think that's handled by the current login in `master`).

jeffstieler:
@joshuatf 

> This PR removes `customer_lookup_batch_init()` and relevant dependent actions. It seems like this is safe since customers are now processed during orders, but want to double check. This also needs to be removed to see benefit of the speed improvements by only processing days or unprocesse orders

I'll test it now, but at first glance I think this will break the "is returning" logic and excludes customers who have registered but not placed an order (not sure how prevalent that is, but I think that's handled by the current login in `master`).

jeffstieler:
Ran into this error testing without providing `days` or `skip_existing`:

`Uncaught ArgumentCountError: Too few arguments to function WC_Admin_Reports_Sync::orders_lookup_process_batch(), 1 passed`

jeffstieler:
It also appears that the order batch job arguments aren't getting set properly:

Screen Shot 2019-04-11 at 10 56 35 AM

joshuatf:
Thanks for taking the time to look at this @jeffstieler 😄 

Gave this a rebase to fix conflicts and dropped the changes in queue processing.  All tests are passing locally and hopefully will here as well 🤞.  I have a hunch that expected stats might be off since we processing the action to completion should work, but I'll dig a little deeper and open a separate issue if that's the case.

>I'll test it now, but at first glance I think this will break the "is returning" logic and excludes customers who have registered but not placed an order

I think this should be okay (and all returning customer tests appear to be passing).  Anytime an order is made, the returning customer logic checks against previous orders made by that customer so not having the customer already in place shouldn't affect this.

>It also appears that the order batch job arguments aren't getting set properly

Fixed the arguments issue here, which should solve both problems you mentioned.  Thanks for noting those!

jeffstieler:
> > I'll test it now, but at first glance I think this will break the "is returning" logic and excludes customers who have registered but not placed an order
> 
> I think this should be okay (and all returning customer tests appear to be passing). Anytime an order is made, the returning customer logic checks against previous orders made by that customer so not having the customer already in place shouldn't affect this.

How about the case where a registered customer hasn't placed an order?

jeffstieler:
I just ran the rebuild after truncating my customer and orders tables.

Before, I had 16215 customer rows, after I have 15886.

Before, I had 3133 "returning customer" orders, after I have 3125.

My thought is that the removal of `customer_lookup_batch_init()` is causing these discrepancies.

joshuatf:
>How about the case where a registered customer hasn't placed an order?

@jeffstieler Good question.  I think we've had some discussion back and forth on this as to whether or not non-purchasing customers should show up here or not.  Did we come to the conclusion that they should be shown?

Maybe we could skip the "import all customers" portion of it if the `days` param is specified since that would imply they wanted to limit to a specific set of orders.

>I just ran the rebuild after truncating my customer and orders tables.

I imagine those are mostly customers without orders.

includes/class-wc-admin-reports-sync.php
```diff
@@ -211,47 +183,86 @@ public static function orders_lookup_update_init() {
...
-				'page'    => $batch_number,
-				'orderby' => 'ID',
-				'order'   => 'ASC',
+	public static function get_orders( $limit = 10, $page = 1, $days = false, $skip_existing = false ) {
+		global $wpdb;
+		$where_clause = '';
+		$offset       = $page > 1 ? $page * $limit : 0;
+
+		if ( $days ) {
+			$days_ago      = date( 'Y-m-d H:i:s', time() - ( DAY_IN_SECONDS * $days ) );
```
Aljullu: I find this a bit confusing. If I activate the _Last 7 days_ import on April 16 11am, I would expect all orders from April 9 to be included, even if they happened at 10am. With this implementation, it would only select the orders occurred in the last 168 hours (7 * 24), which would not include an order from April 9 10am, right?

I don't mind leaving it as-is if I'm the only one with those expectations, but just wanted to write it down in case any of you agree. :slightly_smiling_face:

includes/class-wc-admin-reports-sync.php
```diff
@@ -211,47 +183,86 @@ public static function orders_lookup_update_init() {
...
+				{$where_clause}
+				ORDER BY post_date ASC
+				LIMIT %d
+				OFFSET %d",
+				$limit,
+				$offset
 			)
+		); // WPCS: unprepared SQL ok.
+
+		$count = $wpdb->get_var(
```
Aljullu: Could we run the `$count` query earlier so if it's 0 we don't need to run the `$order_ids` query?

Aljullu (commented):
This is testing well for me, but would like @jeffstieler or somebody with more expertise in this area to give the final approval. I just added a couple of thoughts/questions.

includes/class-wc-admin-reports-sync.php
joshuatf: I would expect the number of hours personally, but I recognize that may be too much of a developer mindset.

I think your expectations are probably more in line with what most people would think.  Will add this in.

includes/class-wc-admin-reports-sync.php
joshuatf: Good idea- added in https://github.com/woocommerce/woocommerce-admin/pull/2034/commits/36a12000f84e6dbdfc16080309fccf7abb23a7a0

includes/class-wc-admin-reports-sync.php
joshuatf: Updated in https://github.com/woocommerce/woocommerce-admin/pull/2034/commits/36a12000f84e6dbdfc16080309fccf7abb23a7a0

joshuatf:
@jeffstieler Can you give this another look if you have time this week?

tests/api/reports-import.php
```diff
@@ -0,0 +1,154 @@
...
+		// Use the days param to only process orders in the last day.
+		$request = new WP_REST_Request( 'POST', $this->endpoint );
+		$request->set_query_params( array( 'days' => '1' ) );
+		$response = $this->server->dispatch( $request );
+		$report   = $response->get_data();
+
+		$this->assertEquals( 200, $response->get_status() );
+		$this->assertEquals( 'success', $report['status'] );
+
+		// Run pending twice to process batch and order.
```
timmyc: Really like the test case here - given that in Jeff's prior review there seemed to be some issues around customer creation logic, do you think we should add some test coverage around that too? Or do we have unit test coverage elsewhere that should cover that?

timmyc:
Exciting to see this taking shape! And +1 @jeffstieler if you could give this one final pass that would be great.

jeffstieler:
I discussed the issue of "registered customers that haven't placed an order" with @LevinMedia and @timmyc - we decided that these folks should indeed show in the customers report. This means we need to restore the `customer_lookup_batch_init ()` routine and incorporate the date range functionality.

I'll take a pass at doing so since @joshuatf is AFK.

jeffstieler:
👋 @timmyc - this is ready for another review/test.

timmyc (approved):
New changes are looking good to me, and in testing this is also working well for me. Thanks for picking this one up Jeff, and 🙌 thanks @joshuatf

PR #2038, opened 2019-04-10
Handle “all guests” when getting customers from orders.

jeffstieler:
`WC_Admin_Reports_Orders_Data_Store::get_customers_by_orders()` issues a bad query when none of the orders have a non-guest (non-0) customer ID.

This PR introduces a `empty()` check before issuing an unnecessary (and broken) query for customers.

- Create an order that will show in the Orders panel (has an actionable status) that has no customer information
- Go to the Dashboard and open the Orders panel
- Verify that the API request made returns without error

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-orders-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-orders-data-store.php
@@ -337,24 +337,29 @@ class WC_Admin_Reports_Orders_Data_Store extends WC_Admin_Reports_Data_Store imp
 	}
 
 	/**
-	 * Get customer data from order IDs.
+	 * Get customer data from Order data.
 	 *
-	 * @param array $orders Array of orders.
+	 * @param array $orders Array of orders data.
 	 * @return array
 	 */
 	protected function get_customers_by_orders( $orders ) {
 		global $wpdb;
-		$customer_lookup_table = $wpdb->prefix . 'wc_customer_lookup';
 
-		$customer_ids = array();
+		$customer_lookup_table = $wpdb->prefix . 'wc_customer_lookup';
+		$customer_ids          = array();
+
 		foreach ( $orders as $order ) {
 			if ( $order['customer_id'] ) {
 				$customer_ids[] = $order['customer_id'];
 			}
 		}
-		$customer_ids = implode( ',', $customer_ids );
 
-		$customers = $wpdb->get_results(
+		if ( empty( $customer_ids ) ) {
+			return array();
+		}
+
+		$customer_ids = implode( ',', $customer_ids );
+		$customers    = $wpdb->get_results(
 			"SELECT * FROM {$customer_lookup_table} WHERE customer_id IN ({$customer_ids})",
 			ARRAY_A
 		); // WPCS: cache ok, DB call ok, unprepared SQL ok.
--- a/includes/data-stores/class-wc-admin-reports-orders-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-orders-data-store.php
@@ -350,7 +350,7 @@ class WC_Admin_Reports_Orders_Data_Store extends WC_Admin_Reports_Data_Store imp
 
 		foreach ( $orders as $order ) {
 			if ( $order['customer_id'] ) {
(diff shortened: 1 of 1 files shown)


psealock (commented):
Code looks good. I can't test this because `getUnreadOrders` is always returning false for me. I place an order through the store UI, which results in an order whose status is `processing`. But this request always results in an empty array: 
```
http://tangaroa.test/wp-json/wc/v4/reports/orders?page=1&per_page=0&status_is%5B0%5D=processing&status_is%5B1%5D=on-hold
```
Am I doing something wrong?

psealock:
Ok, I've run into a time/date/timezone issue. After logging the query and placing it, I see a discrepancy

Screen Shot 2019-04-11 at 3 18 40 PM

I'll dive deeper when I get a chance and log an issue

jeffstieler:
@psealock I've only had the order not show up due to caching. I'm curious to see what you find out re: datetime issue.

psealock:
Looks like a known bug https://github.com/woocommerce/woocommerce-admin/pull/2042#event-2268795619 with fix in progress 😮

joshuatf (commented):
This fix works for me, but I don't think that any customer IDs should be 0 (including guests) to throw this error in the first place.

@jeffstieler Can you dump `$result` in `get_or_create_customer_from_order()` and see if you're getting any errors when inserting customers on those orders where no customer ID exists?

I have a couple orders that relate to first name and last name being null that will be fixed by https://github.com/woocommerce/woocommerce-admin/pull/2034/files, but curious if you have any other items preventing customer insertion.

jeffstieler:
@joshuatf here are the results of `get_or_create_customer_from_order()` and the query that `get_customers_by_orders()` issues on `master` when creating an order described by the testing scenario:

```
[29-Apr-2019 17:13:08 UTC] WordPress database error Column 'first_name' cannot be null for query INSERT INTO `wp_wc_customer_lookup` (`first_name`, `last_name`, `email`, `city`, `postcode`, `country`, `date_last_active`) VALUES (NULL, NULL, '', '', '', '', '2019-04-29 17:12:39') made by do_action_ref_array('action_scheduler_run_queue'), WP_Hook->do_action, WP_Hook->apply_filters, ActionScheduler_QueueRunner->run, ActionScheduler_QueueRunner->do_batch, ActionScheduler_Abstract_QueueRunner->process_action, ActionScheduler_Action->execute, do_action_ref_array('wc-admin_process_order'), WP_Hook->do_action, WP_Hook->apply_filters, WC_Admin_Reports_Sync::orders_lookup_process_order, WC_Admin_Reports_Orders_Stats_Data_Store::sync_order, WC_Admin_Reports_Orders_Stats_Data_Store::update, WC_Admin_Reports_Customers_Data_Store::get_or_create_customer_from_order
[29-Apr-2019 17:13:08 UTC] get_or_create_customer_from_order: false
[29-Apr-2019 17:13:30 UTC] SELECT * FROM wp_wc_customer_lookup WHERE customer_id IN ()
```

includes/data-stores/class-wc-admin-reports-orders-data-store.php
```diff
@@ -336,24 +336,29 @@ protected function get_products_by_order_ids( $order_ids ) {
...
 			}
 		}
-		$customer_ids = implode( ',', $customer_ids );
 
-		$customers = $wpdb->get_results(
+		if ( empty( $customer_ids ) ) {
+			return array();
+		}
+
+		$customer_ids = implode( ',', $customer_ids );
```
mikejolley: You need to use array_map here and apply sanitization to the customer_ids - absint would suffice, because it's being appended to the SQL query below without escaping.

mikejolley (changes requested):
This change makes sense since orders can indeed have customer_id 0, but you do need to add escaping/sanitization to the customer IDs before adding them to the SQL query. Once done, this can be merged.

includes/data-stores/class-wc-admin-reports-orders-data-store.php
jeffstieler: Added in 4e8eccd.

mikejolley (approved):

PR #2039, opened 2019-04-10
Packages: Version bumps & changelog maintenance 

ryelle:
It looks like there's been some change to all the packages (whether package bumps or the textdomain change), so I'm going to bump all of them. This PR handles the versioning & changelog incrementing (some of which have been missed). Once merged, I'll run the publish command to push to npm.

**To test**

- Run `npm install` and build/start the project
- Check that there is no functionality change
- Check that the versions & changelogs look good


packages/components/package.json
```diff
@@ -46,13 +46,14 @@
...
+    "lodash": "4.17.11",
+    "memoize-one": "5.0.2",
+    "moment": "2.22.1",
     "prop-types": "15.7.2",
-    "qs": "^6.5.2",
-    "react-dates": "^18.0.4",
-    "react-router-dom": "4.3.1",
-    "react-transition-group": "^2.4.0"
+    "qs": "6.7.0",
+    "react-dates": "17.2.0",
```
psealock: Why the downgrade? Is this to match what Gutenberg is on?

psealock (approved):
Everything looks good to me, working well and no errors in the build step or the console.

packages/components/package.json
ryelle: Exactly - I don't know if there's still a plan to converge into the Gutenberg datepicker, but keeping on the same version made sense.

PR #2042, opened 2019-04-11
Respect manual offsets in default before/after params

### LINKED ISSUE #2035: Default before/after doesn't respect manual offsets
1. Set your store timezone to _UTC+14_, for example.
2. Create an _on hold_ order and rebuild tables.
3. Notice the order doesn't appear in the _Orders_ tab of the Activity Panel.

That's because the order is stored in the lookup table using store's timezone but when we fetch them the manual offset is not used.

Aljullu:
Fixes #2035.

Default values for before/after params did not respect manual offsets which are not recognized by `wc_timezone_string()`, so setting your store timezone to _UTC+14_, for example, might give unexpected results. This PR applies a solution used in WooCommerce that handles the case when the timezone is not one of the recognized ones:
https://docs.woocommerce.com/wc-apidocs/source-function-wc_timezone_string.html#654-659

1. Set your store timezone to _UTC+14_, for example.
2. Create an _on hold_ order and rebuild tables.
3. Verify the order appears in the _Orders_ tab of the Activity Panel.
4. Verify there are no regressions in other queries.

### DIFF
--- a/includes/class-wc-admin-reports-interval.php
+++ b/includes/class-wc-admin-reports-interval.php
@@ -44,8 +44,13 @@ class WC_Admin_Reports_Interval {
 	 * @return DateTime
 	 */
 	public static function default_before() {
-		$datetime = new DateTime();
-		$datetime->setTimezone( new DateTimeZone( wc_timezone_string() ) );
+		$datetime = new WC_DateTime();
+		// Set local timezone or offset.
+		if ( get_option( 'timezone_string' ) ) {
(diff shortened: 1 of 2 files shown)


psealock (approved):
Wow, nice fix here and great timing as I had just run into this bug. This works great for me. Offhand, I'm not overly familiar with the implications of this change, but it seems to make sense and I've tested the charts hourly on the reports and all seems to be in order.

PR #2045, opened 2019-04-11
Packages: Update docs to remove outdated step in process

ryelle:
Since we’re using `from-package`, lerna does not ask for a version in the release process. This step in the docs threw me off, so let's just remove it.

- Make sure the release process sounds correct (which you probably can't do without running a release 🙃)


timmyc (approved):
I recall this portion of releasing components being a tad confusing for me too.

PR #2046, opened 2019-04-11
Build: Clean up package.json & add a `dev` script

ryelle:
As we all noticed yesterday in slack, `build` only builds the "core" files, not the development features. Personally I don't always need to run start, and I find it easier to see when non-`watch` tasks are finished. So I've added `dev` as a new script, letting you run `npm run dev` to just build the plugin once, with the development features turned on.

I've also run [`sort-package-json`](https://www.npmjs.com/package/sort-package-json) on the main `package.json` file to keep things organized – this uses [a set sort order](https://github.com/keithamus/sort-package-json/blob/HEAD/index.js) on all the keys in the package object, and also sorts the `scripts` for us (which was the main reason I ran it). We don't need to keep up with this, I just wanted to see the scripts groups better.

The only actual changes are the addition of `predev` and `dev` in scripts.

- Run `npm run dev`
- It should complete, and now you can see the Dashboard, Analytics, etc
- Make an arbitrary change
- Run `npm run dev`
- Your change should be there when you view the admin


joshuatf (approved):
Nice work on this @ryelle!  Works as expected and updates are visible after running `npm run dev`.  LGTM after a rebase.

>The only actual changes are the addition of predev and dev in scripts.

Thanks for noting this 😄

PR #2047, opened 2019-04-12
Remove lodash from bundle

psealock:
While looking at https://github.com/Prospress/action-scheduler-admin/pull/2 for @rrennick, it occurred to me that since lodash is already on window, we can exclude it from our build.

### Before

5.67 MB

Screen Shot 2019-04-12 at 2 20 11 PM

### After

5.16 MB

Screen Shot 2019-04-12 at 2 22 36 PM

1. `npm start`
2. Check that nothing is broken

### DIFF
--- a/package.json
+++ b/package.json
@@ -140,7 +140,6 @@
     "history": "4.9.0",
     "html-to-react": "1.3.4",
     "interpolate-components": "1.1.1",
-    "lodash": "^4.17.11",
     "marked": "0.6.2",
     "memoize-one": "5.0.2",
     "prismjs": "^1.15.0",
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -38,6 +38,7 @@ const externals = {
 	tinymce: 'tinymce',
 	moment: 'moment',
 	react: 'React',
+	lodash: 'lodash',
 	'react-dom': 'ReactDOM',
 };
 


Aljullu (approved):
Lovely to see the bundle size decreasing! :clap: 

Tested and everything seems to be working fine.

Aljullu:
@psealock do you think we should remove `lodash` from `package.json` now? I noticed `react`, `moment`, etc. are not in `package.json` either.

rrennick:
Did anyone check this with WP 5.0?

Aljullu:
> Did anyone check this with WP 5.0?

@rrennick I was testing it right now with WordPress 5.0.0 and everything seems to work fine. Is there anything specific we should look into?

rrennick:
> Is there anything specific we should look into?

No, both this extension & https://github.com/timmyc/wcs-example-report worked for me. Both stopped working sometime recently. On of the possible causes might have been an auto update to WP.

Aljullu:
> No, both this extension & https://github.com/timmyc/wcs-example-report worked for me. Both stopped working sometime recently. On of the possible causes might have been an auto update to WP.

rrennick:
> I think it's worth creating a separate issue if you still have this issue, so we can better track this.

 My install is 5.1.1 so removing the `lodash` worked for both.

PR #2051, opened 2019-04-12
Scroll to top of the table when navigating table pages

### LINKED ISSUE #2014: Scroll to top of the table when navigating between report table pages
Discussed in #1989 comments. Every time the user navigates to a different page in a table report, we want the viewport to scroll to the top of the table.

image

Aljullu:
Fixes #2014.

### Accessibility

### Screenshots
table-scroll

- Go to any report where there is table pagination.
- Try going to the next/previous table page.
- Verify you are scrolled to the top of the table.
- Try with different window sizes.


client/analytics/components/report-table/index.js
```diff
@@ -101,19 +109,27 @@ class ReportTable extends Component {
...
-				rows={ rows }
-				rowsPerPage={ parseInt( query.per_page ) || QUERY_DEFAULTS.pageSize }
-				summary={ summary }
-				totalRows={ totalResults }
-				{ ...tableProps }
-			/>
+			<Fragment>
+				<div
+					className="woocommerce-report-table__scroll-point"
+					ref={ this.scrollPointRef }
```
joshuatf: Are there issues with adding this ref to the `TableCard` component directly?

client/stylesheets/abstracts/_variables.scss
```diff
@@ -13,6 +13,11 @@ $gap-small: 12px;
 $gap-smaller: 8px;
 $gap-smallest: 4px;
 
+// Header
+$small-header-height: 50px;
```
joshuatf: Awesome- thanks for adding these 😄 

While we're here, can we also use these for the activity panel and breadcrumb heights that are set manually?

joshuatf (commented):
This is testing well for me, Albert- nice changes!

Left a comment about the scroll reference point.  Also when scrolling back up on page change, I would expect a small gap between the header and top of the table, but I may be alone in this thinking so feel free to ignore this if so.

Screen Shot 2019-04-22 at 11 29 49 AM

client/analytics/components/report-table/index.js
Aljullu: The main reason is that the scroll-point needs to be rendered on top of the table, so the browser scrolls some pixels above where the table starts. We would not be able to do that if the ref was assigned directly to the table.
image

client/stylesheets/abstracts/_variables.scss
Aljullu: > While we're here, can we also use these for the activity panel and breadcrumb heights that are set manually?

Sure thing! I will do that in a follow-up if that sounds ok, so we don't block this PR.

Aljullu:
Thanks for the review @joshuatf!

> Also when scrolling back up on page change, I would expect a small gap between the header and top of the table, but I may be alone in this thinking so feel free to ignore this if so.

I like this idea, fixed in fd5d3e7980573ee99a4c1ebbcd2f6a83c7431c3f. This should be ready for another review.

client/analytics/components/report-table/index.js
```diff
@@ -49,6 +53,10 @@ class ReportTable extends Component {
 		}
 	}
 
+	onPageChange() {
+		this.scrollPointRef.current.scrollIntoView();
```
ryelle: I think you need to explicitly pull focus up to the table. when I try to use this with VoiceOver, the screen reader's focus stays on the pagination (it might be different on other screen readers, since I do see that keyboard focus moves to the top)

Aljullu: Good catch with that, it should be fixed in the last commits.

ryelle: Hm, now it seems to move me into the search (or download button), which might be confusing. Do you think you can make it focus onto the table? That's already got a tab-stop to it, and it will bring you directly to the content that changed (vs tabbing through the search/actions).

Aljullu: That's a good point. I made it so it focuses on `.woocommerce-table__table` in c736abff6e0717b771925556ba4715494c3f3364.

We could also focus directly the first table header or table rows if we want to. I'm not sure what would be more intuitive for users.

psealock (approved):
This is testing well for me in Chrome, FF, and Safari. I also tried with VoiceOver and focus is brought to the first button in the table, which makes sense as the desired behavior.

PR #2057, opened 2019-04-15
Add import delete endpoint and batch actions

joshuatf:
Fixes (part of) #1850

* Renames existing batch functions with an `IMPORT_` prefix.
* Adds the delete endpoint to the import controller.
* Adds delete batch actions for orders and customers.
* Adds a customer data store deletion method.

### Screenshots
Screen Shot 2019-04-15 at 4 08 07 PM

1. Make a POST request to `/wp-json/wc/v4/reports/import/delete`
2. Check that batch actions are queued up and that all stats are deleted afterwards.

### DIFF
--- a/includes/class-wc-admin-reports-sync.php
+++ b/includes/class-wc-admin-reports-sync.php
@@ -22,24 +22,44 @@ class WC_Admin_Reports_Sync {
 	const QUEUE_DEPEDENT_ACTION = 'wc-admin_queue_dependent_action';
 
 	/**
-	 * Action hook for processing a batch of customers.
+	 * Action hook for importing a batch of customers.
 	 */
-	const CUSTOMERS_BATCH_ACTION = 'wc-admin_process_customers_batch';
-
-	/**
-	 * Action hook for processing a batch of orders.
-	 */
-	const ORDERS_BATCH_ACTION = 'wc-admin_process_orders_batch';
+	const CUSTOMERS_IMPORT_BATCH_ACTION = 'wc-admin_import_customers_batch';
 
 	/**
 	 * Action hook for initializing the orders lookup batch creation.
 	 */
-	const ORDERS_LOOKUP_BATCH_INIT = 'wc-admin_orders_lookup_batch_init';
+	const CUSTOMERS_DELETE_BATCH_INIT = 'wc-admin_delete_customers_batch_init';
 
 	/**
-	 * Action hook for processing a batch of orders.
+	 * Action hook for importing a batch of customers.
 	 */
-	const SINGLE_ORDER_ACTION = 'wc-admin_process_order';
+	const CUSTOMERS_DELETE_BATCH_ACTION = 'wc-admin_delete_customers_batch';
+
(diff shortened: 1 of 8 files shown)


includes/class-wc-admin-reports-sync.php
```diff
@@ -22,24 +22,44 @@ class WC_Admin_Reports_Sync {
...
-	 * Action hook for processing a batch of orders.
+	 * Action hook for initializing the orders lookup batch deletion.
 	 */
-	const SINGLE_ORDER_ACTION = 'wc-admin_process_order';
+	const ORDERS_DELETE_BATCH_INIT = 'wc-admin_orders_lookup_delete_batch_init';
+
+	/**
+	 * Action hook for deleting a batch of orders.
+	 */
+	const ORDERS_DELETE_BATCH_ACTION = 'wc-admin_delete_orders_batch';
```
Aljullu: Nitpicking here: the two previous action hooks use the wording `orders_lookup` while here it's only `orders`. Is there any reason for this or could it be unified to avoid confusion?

includes/class-wc-admin-reports-sync.php
```diff
@@ -452,6 +496,82 @@ public static function customer_lookup_process_batch( $batch_number ) {
...
+		self::queue_batches( 1, $num_batches, self::CUSTOMERS_DELETE_BATCH_ACTION );
+	}
+
+	/**
+	 * Delete a batch of customers.
+	 *
+	 * @param int $batch_number Batch number to import (essentially a query page number).
+	 * @return void
+	 */
+	public static function customer_lookup_delete_batch( $batch_number ) {
```
Aljullu: It looks like `$batch_number` is not used inside this function, so there would be no need to declare it.

Aljullu (approved):
Fantastic job with this @joshuatf! I tested it and everything seems to be working fine for me. Tests pass as well. Pre-approving + I added a few comments/suggestions below.

includes/class-wc-admin-reports-sync.php
```diff
@@ -22,24 +22,44 @@ class WC_Admin_Reports_Sync {
...
 
 	/**
-	 * Action hook for processing a batch of orders.
+	 * Action hook for initializing the orders lookup batch creation.
 	 */
-	const ORDERS_BATCH_ACTION = 'wc-admin_process_orders_batch';
+	const CUSTOMERS_DELETE_BATCH_INIT = 'wc-admin_delete_customers_batch_init';
+
+	/**
+	 * Action hook for importing a batch of customers.
```
timmyc: Isn't this comment incorrect?

timmyc (commented):
One minor item :)

timmyc:
@jeffstieler were you going to make any other changes here or is this ready for more testing meow?

jeffstieler:
@timmyc this is ready for testing. Thanks for checking.

timmyc (approved):
Updates look good to me, thanks Jeff and Josh!

PR #2060, opened 2019-04-15
update path to feature-config-core.php

### LINKED ISSUE #2053: Fatal error: require_once() ... wp-content/plugins/woocommerce-admin//includes/feature-config.php
**Describe the bug**
Fatal error on activating the plugin after a fresh cloning and `npm run build`.

**To Reproduce**

1. Clone the repo
2. `npm run build`
3. See this during the build process

```
@woocommerce/admin-library@0.10.0 build:feature-config /Applications/MAMP/htdocs/zzz/wp-content/plugins/woocommerce-admin
> > php bin/generate-feature-config.php
> Warning: fopen(dist/feature-config-core.php): failed to open stream: No such file or directory in /Applications/MAMP/htdocs/zzz/wp-content/plugins/woocommerce-admin/bin/generate-feature-config.php on line 38
> Warning: fwrite() expects parameter 1 to be resource, boolean given in /Applications/MAMP/htdocs/zzz/wp-content/plugins/woocommerce-admin/bin/generate-feature-config.php on line 41
> Warning: fclose() expects parameter 1 to be resource, boolean given in /Applications/MAMP/htdocs/zzz/wp-content/plugins/woocommerce-admin/bin/generate-feature-config.php on line 42
```

and then see this on attempting to activate the plugin:

```
Warning: require_once(/Applications/MAMP/htdocs/zzz/wp-content/plugins/woocommerce-admin//includes/feature-config.php): failed to open stream: No such file or directory in /Applications/MAMP/htdocs/zzz/wp-content/plugins/woocommerce-admin/woocommerce-admin.php on line 233

Fatal error: require_once(): Failed opening required '/Applications/MAMP/htdocs/zzz/wp-content/plugins/woocommerce-admin//includes/feature-config.php' (include_path='.:/Applications/MAMP/bin/php/php7.2.10/lib/php') in /Applications/MAMP/htdocs/zzz/wp-content/plugins/woocommerce-admin/woocommerce-admin.php on line 233
```

**Expected behavior**

Error free activation. :)

* Mojave, Chrome, MAMP

rrennick:
See #2053

This PR fixes ensures the `/dist/` path exists for feature-config-core.php in build-feature-config.php. 

- Pull down a new clone of the repo
- In the plugin root `rm -rf dist`
- `npm install`
- `npm run build`
- The run should load cleanly.


rrennick:
@justinshreve the `build-feature-config.php` is in the `/dist/` folder so the current code is trying to write to `/dist/dist/'.

justinshreve:
This doesn't seem to be the case for me. The file is in `bin/`, but I added `getcwd()` and it's running from the root directory. The file gets created in `woocommerce-admin/dist` and `woocommerce-admin/includes`.

FWIW, I deleted my `dist/` folder in the root directory and the config inside `includes` -- and following these steps still shows me a fatal when loading the dashboard:

`PHP Warning:  require_once(/Users/work/dev/woocommerce-admin//includes/feature-config.php): failed to open stream: No such file or directory in /Users/work/dev/woocommerce-admin/woocommerce-admin.php on line 234`

I guess we should update the build script to always generate `feature-config.php`. Right now it doesn't build if you are targeting the core flow .

justinshreve:
We might also want to consider updating the default target to be `development` instead of core:

Currently: 
```
if ( ! in_array( $phase, array( 'development', 'plugin', 'core' ), true ) ) {
	$phase = 'core';
}
```

psealock:
>We might also want to consider updating the default target to be development instead of core:

I agree. Or even `plugin` makes sense.

rrennick:
@justinshreve @psealock Thanks for looking at this.

I deleted the `/dist/` folder for testing and the script doesn't give any indication that the folder doesn't exist. A check to create the folder if it doesn't exist at the beginning of the process would be good.

```shell

> @woocommerce/admin-library@0.10.0 build:feature-config /Users/ronrennick/Sites/ron/wp-content/plugins/wc-admin
> php bin/generate-feature-config.php

```

rrennick:
@justinshreve I've updated the PR description. Even though this doesn't fix #2053, it does fix another instance of that type of error. I changed the target to `development`. Would it be better to 

> Or even `plugin` makes sense since npm run dev runs the process on development phase.

bin/generate-feature-config.php
```diff
@@ -35,6 +35,10 @@
 	}
 	$write .= "\t\t\t)";
 
+	if ( is_file( './dist' ) || ! wp_mkdir_p( './dist' ) ) {
```
ryelle: This file isn't run in a WP context, so this function doesn't exist. What about doing something like this:

```suggestion
	if ( ! is_dir( './dist' ) && ! @mkdir( './dist' ) ) {
```

It'll error if `dist` is a file for some reason, but that _is_ an error case so I think that's an OK option. Maybe we can add a message here to suggest running `npm run clean` to wipe artifacts?

rrennick: @ryelle I have updated the PR with your suggestions. If you 

`rm -rf dist && touch dist && npm run build`

you'll see that it errors out but the message I added is not in the output or in the log. Would it be better to start the build with `npm run clean`?

ryelle: I do see the error, but it's missing a linebreak:

Screen Shot 2019-04-24 at 2 39 37 PM

rrennick: @ryelle Thanks, I have updated the PR & added two line breaks so there is a blank line after the message.

bin/generate-feature-config.php
```diff
@@ -7,7 +7,7 @@
 
 $phase = isset( $_SERVER['WC_ADMIN_PHASE'] ) ? $_SERVER['WC_ADMIN_PHASE'] : ''; // WPCS: sanitization ok.
 if ( ! in_array( $phase, array( 'development', 'plugin', 'core' ), true ) ) {
-	$phase = 'core';
+	$phase = 'development';
```
ryelle: It looks `WC_ADMIN_PHASE` isn't actually passed down to this file when we run `npm run build` (which might also be why you didn't see the message, @rrennick) – we could either change this back to `core`, or change the `build:feature-config` to be `WC_ADMIN_PHASE=core php bin/generate-feature-config.php` … I don't really have a preference either way.

rrennick: @ryelle Thanks again. I've updated the PR & fixed the merge conflicts.

PR #2065, opened 2019-04-16
remove unneeded build:feature-config from zip build

rrennick:
The `npm run build:feature-config` is not needed in the plugin zip build process because the `npm run build` in the next line calls `npm run build:feature-config` first.

- at the command line run `npm run build:release`
- Assuming you working tree is clean this should produce a `woocommerce-admin.zip` in the root folder of the repo.


rrennick:
Thanks @justinshreve

Issue #1851, opened 2019-03-19
Add Historical Data Import Screen


jeffstieler:
Add an "Import Historical Data" section to the Analytics > Settings page.

UI if historical import tool hasn't been run:

![Screen Shot 2019-03-19 at 4 57 23 PM](https://user-images.githubusercontent.com/63922/54647441-17ed6980-4a68-11e9-9029-645a92e43f9c.png)

Makes use of endpoints created in https://github.com/woocommerce/woocommerce-admin/issues/1850.

Values for dropdown:
* All
* Last 365 days
* Last 90 days
* Last 30 days
* Last 7 days
* Last 24 hours
* Custom

![Screen Shot 2019-03-19 at 5 15 07 PM](https://user-images.githubusercontent.com/63922/54648093-93501a80-4a6a-11e9-9cd5-ef6716e8e3c8.png)

"Custom" invokes a date picker. The value passed to API requests should be the number of days from the chosen date to the current date.

UI when an import is in process:

![Screen Shot 2019-03-19 at 5 16 57 PM](https://user-images.githubusercontent.com/63922/54648256-2721e680-4a6b-11e9-95dc-249cb193b7dd.png)

UI after historical tool has been run:

![Screen Shot 2019-03-19 at 5 21 08 PM with delete](https://user-images.githubusercontent.com/63922/54648913-a7e1e200-4a6d-11e9-9fa5-982b0b84adf3.png)

# Tasks
* [x] Create UI (#2073).
* [x] Hook up actions to endpoints (#2177).
* [ ] Read totals from endpoint (#2208).
* [ ] Read import status endpoint.

LevinMedia:
Prototype for reference: 

https://www.figma.com/proto/G1xF07y9TuuhfirzYtqmIvTO/Analytics-Settings?node-id=140%3A0&viewport=-2633%2C1126%2C0.399093&scaling=min-zoom

Aljullu:
@LevinMedia I noticed there are two different designs for stopping an ongoing import (slides 4 and 8):

![image](https://user-images.githubusercontent.com/3616980/56233973-93046880-6084-11e9-89b3-cd7f3055e2ee.png)
![image](https://user-images.githubusercontent.com/3616980/56233986-9d266700-6084-11e9-845f-ecb800c7924e.png)

One has a button with label _Stop_ and a description text next to it and the other only has a button with the label _Stop Import_. Is that in purpose? In that case, it's not clear to me when we should display one or the other.

[...]

PR #2073, opened 2019-04-17
Add Historical Data Import Screen

Aljullu:
Part of #1851.

This PR creates the components for the _Historical Data Import_ section in the _Settings_ screen even though it's not yet connected to the endpoints.

### Screenshots
image

- Go to the _Settings_ page.
- Verify there is an _Import Historical Data_ section.
- Try modifying `<HistoricalData>` [props](https://github.com/woocommerce/woocommerce-admin/pull/2073/files#diff-26f12905770396a5868a1a7987751074R165) to simulate different stages of the page.


client/analytics/settings/historical-data/index.js
```diff
@@ -0,0 +1,185 @@
...
+						<Fragment>
+							<Button isPrimary onClick={ this.changeStep( 'customers' ) }>
+								{ __( 'Start', 'woocommerce-admin' ) }
+							</Button>
+							<Button isDefault onClick={ this.changeStep( 'finished' ) }>
+								{ __( 'Delete Previously Imported Data', 'woocommerce-admin' ) }
+							</Button>
+						</Fragment>
+					) }
+					{ step === 'finished' && (
```
joshuatf: I think instead of using steps here we can let the data from the API drive the available options.  Steps might also create issues where data was imported, but it's not possible to delete it until running through all the steps here again.
 
Would `inProgress` and `hasImportedData` be enough to remove all the step logic?

client/analytics/settings/historical-data/status.js
```diff
@@ -0,0 +1,26 @@
+/** @format */
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { Spinner } from '@wordpress/components';
+
+function HistoricalDataStatus( { status } ) {
+	const statusLabels = {
```
joshuatf: Nice!  Like the organization here.  I wonder if we should add a filter here to allow third party devs to add in their own import status labels.

joshuatf (commented):
Awesome work @Aljullu!  Left a few comments about moving some of the state into props in preparation for the API, but things are looking really good here.

One other thing I noticed was that the select dropdown and date inputs are of varying sizes and look a bit off.  This is probably something we need to tack in a separate PR as it might involve using a stylized select component across wc-admin. /cc @LevinMedia 

Screen Shot 2019-04-18 at 12 45 12 PM

client/analytics/settings/historical-data/period-selector.js
```diff
@@ -0,0 +1,84 @@
...
+		const momentDate = moment( value.date, dateFormat );
+		return (
+			<div className="woocommerce-settings-historical-data__column">
+				<div className="woocommerce-settings-historical-data__column-label">
+					{ __( 'Begining on', 'woocommerce-admin' ) }
+				</div>
+				<DatePicker
+					date={ momentDate.isValid() ? momentDate.toDate() : null }
+					dateFormat={ dateFormat }
+					disabled={ disabled }
```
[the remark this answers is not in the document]
Aljullu: Yes, when you press _Start_ it should disable all form elements:
image

client/analytics/settings/historical-data/index.js
Aljullu: I re-implemented it in its own component trying to make the logic simpler and getting rid of `step`. Let me know how it looks!

Aljullu:
Thanks for the review @joshuatf! I updated this PR with your suggestions. Feel free to take another look when you have a second.

> One other thing I noticed was that the select dropdown and date inputs are of varying sizes and look a bit off. This is probably something we need to tack in a separate PR as it might involve using a stylized select component across wc-admin. /cc @LevinMedia

Good catch! I agree it makes sense creating a component or at least a CSS class that we can use everywhere, so we avoid having this issue in the future. Seeing how big this PR already is, I will work on that in a follow-up as you said. :+1:

client/analytics/settings/historical-data/actions.js
```diff
@@ -0,0 +1,82 @@
...
+					{ __( 'Start', 'woocommerce-admin' ) }
+				</Button>
+			);
+		}
+
+		// Has imported all possible data
+		if ( customersProgress === customersTotal && ordersProgress === ordersTotal ) {
+			return (
+				<Button isDefault onClick={ () => null }>
+					{ __( 'Import Additional Historical Data', 'woocommerce-admin' ) }
```
joshuatf: This text sounds a bit odd to me, like there's more that can be imported.  I kind of expected something like `Re-import Data` since everything has already been imported here. /cc @LevinMedia 

Also, do we want to show the delete option here?

client/analytics/settings/historical-data/status.js
```diff
@@ -0,0 +1,37 @@
...
+import { applyFilters } from '@wordpress/hooks';
+import moment from 'moment';
+import { Spinner } from '@wordpress/components';
+
+/**
+ * WooCommerce dependencies
+ */
+import { useFilters } from '@woocommerce/components';
+
+const HISTORICAL_DATA_STATUS_FILTER = 'woocommerce-import-status';
```
joshuatf: Awesome! 🎉 Small nitpick:  Can we use the `woocommerce_admin_` prefix to keep consistent with the other filters?

```suggestion
const HISTORICAL_DATA_STATUS_FILTER = 'woocommerce_admin_import_status';
```

client/analytics/settings/historical-data/index.js
```diff
@@ -0,0 +1,174 @@
...
+			customersProgress,
+			customersTotal,
+			inProgress,
+			ordersProgress,
+			ordersTotal,
+		} = this.props;
+
+		if ( inProgress ) {
+			if ( customersProgress < customersTotal ) {
+				return 'customers';
```
joshuatf: Side note here- orders can create customers on the fly without first importing customers.  https://github.com/woocommerce/woocommerce-admin/pull/2034 may remove the initial customer import (at least when days are specified).

It probably doesn't affect this and I don't think any changes are needed now, but just wanted to point this out for reference in case we run into any issues later on.

client/analytics/settings/historical-data/period-selector.js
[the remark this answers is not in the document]
joshuatf: It wasn't, but seems to be working for me now!  Maybe I did something wrong 😄

joshuatf (approved):
This is a really fantastic addition to the settings, Albert!  🎉 

Left a couple nitpicks and a question about the button text when imported is completed, but pre-approving pending your judgement and/or @LevinMedia's response.

client/analytics/settings/historical-data/actions.js
Aljullu: > This text sounds a bit odd to me, like there's more that can be imported.

Agree with that, it was also confusing for me. I changed it for now. We can undo it or change it again in case @LevinMedia prefers another copy.

> Also, do we want to show the delete option here?

Actually in #1851 there is a _Delete_ button that was not in Figma designs. I think it makes sense adding it.

Both changes are included in b2f47e436bd446934248c8807bdd4730c4945f5b.

client/analytics/settings/historical-data/status.js
Aljullu: Done in b3058b120a82e16862f51f69adfab27a68bf6e4b. :+1:

client/analytics/settings/historical-data/index.js
Aljullu: Thanks for pointing this out @joshuatf! I'm pretty sure we will have to do some more tweaks once we start connecting this with the real endpoints.

timmyc:
Looking awesome @Aljullu - /cc @LevinMedia @josemarques for a quick design check

timmyc:
@Aljullu I say we merge this in so we don't have to keep rebasing this. We can get a design review after it is merged still and revisit and update as needed. Team X is on meetup this week so it would be nice to land this.

Aljullu:
Sounds good @timmyc!

PR #2086, opened 2019-04-19
Add date_created_gmt property to orders endpoint

### LINKED ISSUE #2085: Missing dates in Orders Activity Panel
**To Reproduce**

1. Create an on hold order.
2. Open the _Orders_ tab of the Activity Panel.
3. Notice there is no date next to it.

image

Aljullu:
Fixes #2085.

In the Activity Panel we need the GMT dates intead of the local dates in order to correctly display the unread notifications. Currently, the `reports/orders` endpoint was only returning the local date. This PR adds `date_created_gmt` as a column in the `wc_order_stats` table and makes the endpoint return that value.

### Screenshots
_Before:_
image

_After:_
image

You will need to force the `wc_orders_stats` table to be recreated. You can do that deactivating the plugin, removing that table, changing the option `wc_admin_version` to something like `0.9.0` and activating the plugin again.
1. Create an on hold order.
2. Open the _Orders_ tab of the Activity Panel.
3. Verify the date appears in every order card.
4. In _Settings_, set your store timezone to a different timezone.
5. Create an order again and verify the time displayed next to the order is correct.


includes/data-stores/class-wc-admin-reports-orders-data-store.php
```diff
@@ -42,15 +43,16 @@ class WC_Admin_Reports_Orders_Data_Store extends WC_Admin_Reports_Data_Store imp
...
-		'status'         => 'REPLACE(status, "wc-", "") as status',
-		'customer_id'    => 'customer_id',
-		'net_total'      => 'net_total',
-		'gross_total'    => 'gross_total',
-		'refund_total'   => 'refund_total',
-		'num_items_sold' => 'num_items_sold',
-		'customer_type'  => '(CASE WHEN returning_customer <> 0 THEN "returning" ELSE "new" END) as customer_type',
+		'order_id'         => 'order_id',
+		'date_created'     => 'date_created',
+		'date_created_gmt' => 'date_created_gmt',
```
joshuatf: I know this is the way this is handled in core right now, but I thought there was a case recently where we considered using timestamps.  I might be misremembering though.  /cc @peterfabian

Are there problems we know of associated with storing in a timestamp and converting to local times after fetch? Or is it  less complicated just to match what's done in core?

joshuatf (commented):
Thanks for taking care of this @Aljullu!  Storing `date_created_gmt` is working well for me.  

If I make an order and subsequently change my store's timezone, saving the order again will update the GMT time to change the original order date.  This would make sense if the old timezone was an incorrect one, but not necessarily if a store was to move to a different timezone.

I'm not sure if this is expected behavior or not, but I guess if this is the way it's handled elsewhere we can keep this pattern.

>You will need to force the wc_orders_stats table to be recreated.

Just as a heads up if you have to do this again in the future (or explain it in testing instructions), you can delete the `wc_admin_version` option to do this a little faster.

includes/data-stores/class-wc-admin-reports-orders-data-store.php
peterfabian: The problem was that the data is sliced into intervals the database (e.g. to days/weeks) and the boundaries of those intervals would be incorrect if time in the database was GMT/UTC. Having both local time and GMT time is possible, but then the local time needs to be updated when time zone changes and one needs to be careful how that works (e.g. I guess you can't just detect current GMT offset and update based on that, as you have DST etc).

peterfabian: discussion with a bit of background: https://a8c.slack.com/archives/C0E1AV8T0/p1549463094980600

timmyc: Is there any reason against storing the GMT also here? Seems like it remedies the issue - I suppose it can create some edge cases if the stores tz changes as @joshuatf pointed out.

Aljullu:
> If I make an order and subsequently change my store's timezone, saving the order again will update the GMT time to change the original order date. This would make sense if the old timezone was an incorrect one, but not necessarily if a store was to move to a different timezone.
> 
> I'm not sure if this is expected behavior or not, but I guess if this is the way it's handled elsewhere we can keep this pattern.

That seems to be WC's behavior in order to respect new timezone's offset between the local date and the GMT date. `wp_posts` is updated this way, so I guess `wc_order_stats` should just respect that.

When `wc-admin_process_order` hook is run, `wc_order_stats` is updated with the dates from the orders. Do you think there is something to fix in this PR?

> Just as a heads up if you have to do this again in the future (or explain it in testing instructions), you can delete the wc_admin_version option to do this a little faster.

Thank you @joshuatf! :smile:

includes/api/class-wc-admin-rest-reports-orders-controller.php
```diff
@@ -156,55 +156,61 @@ public function get_item_schema() {
...
 				),
-				'date_created'   => array(
+				'date_created'     => array(
 					'description' => __( 'Date the order was created.', 'woocommerce-admin' ),
 					'type'        => 'string',
 					'context'     => array( 'view', 'edit' ),
 					'readonly'    => true,
 				),
-				'status'         => array(
+				'date_created_gmt' => array(
```
mikejolley: This new schema and `date_created` above are strings, but core has a `'type' => 'date-time',` for it's dates. Does this need adding?

includes/api/class-wc-admin-rest-reports-orders-controller.php
```diff
@@ -156,55 +156,61 @@ public function get_item_schema() {
...
-				'order_number'   => array(
+				'order_number'     => array(
 					'description' => __( 'Order Number.', 'woocommerce-admin' ),
 					'type'        => 'string',
 					'context'     => array( 'view', 'edit' ),
 					'readonly'    => true,
 				),
-				'date_created'   => array(
+				'date_created'     => array(
 					'description' => __( 'Date the order was created.', 'woocommerce-admin' ),
```
mikejolley: We should add clarity here and match the descriptions in the Woo Core REST API.

`, in the site's timezone.`

and 

`, as GMT.`

includes/data-stores/class-wc-admin-reports-orders-data-store.php
mikejolley: This confused me too originally. TBH, changing store timezone could be a rarity we don't need to account for (just document). We could implement tracks to prove it's uncommon..

I'd suggest sticking to local timestamps for the stats tables, then flushing them on change, rather than trying to handle conversions of existing data.

mikejolley (changes requested):
Looks good, just added some feedback for the schema, and there are some comments around GMT.

includes/data-stores/class-wc-admin-reports-orders-data-store.php
Aljullu: In case we don't want to store the GMT datetime in the stats table, I guess we could use the local datetime which is already stored and calculate the GMT time in the frontend given the store timezone offset. That will require some changes in the frontend but would avoid having to add a new column in the table. Do you think that would be a better solution @mikejolley?

mikejolley: I think it's fine to add this to the table for efficiency as long as we understand the limitations and flush rather than try to convert should the timezone change :)

Aljullu: > and flush rather than try to convert should the timezone change

The thing here is that when the user changes the store's timezone, we don't update any order in the database. It's only when the user edits one of the orders, that the datetimes of that order are updated according to the new timezone, that happens in `posts` and it's immediately replicated in `wc_order_stats`.

Considering that, what would be the point in flushing the `wc_order_stats` table when changing the timezone? We want datetimes from `posts` and `wc_order_stats` to be in sync and that's already happening now.

Am I missing something?

includes/api/class-wc-admin-rest-reports-orders-controller.php
Aljullu: That was fixed in 40c867d531710b5e82f4174b19524be185e4ac18.

includes/api/class-wc-admin-rest-reports-orders-controller.php
Aljullu: Fixed in 40c867d531710b5e82f4174b19524be185e4ac18 too.

Aljullu:
Thanks for the review @mikejolley! Could you take another look?

includes/data-stores/class-wc-admin-reports-orders-data-store.php
mikejolley: I see what you mean. In that case I guess we do nothing; if posts and orders don't update either we just assume the dates in the DB are correct after a timezone change, even if in reality they are not..?

Aljullu: @mikejolley that's how WooCommerce currently works, so I guess it makes sense wc-admin to do the same.

jeffstieler (approved):
Looks good. 🛳 it.

mikejolley (approved):

PR #2091, opened 2019-04-19
Allow free input in the table pagination input field

### LINKED ISSUE #2050: It's not possible to use the backspace in the table page input
**To Reproduce**
- Go to any report where there is table pagination.
- Focus the page input at the bottom of the table.
- Press <kbd>backspace</kbd> or <kbd>Del</kbd>.
- Notice it's not possible to remove the current number.

**Expected behavior**
It should be allowed to remove the number in order to write a new one.

**Screenshots**
image

### LINKED ISSUE #2090: Writing 0 in the table pagination breaks the report
table-pagination-error

**To Reproduce**
1. Go to any report.
2. Write 0 in the input field below the table (Pagination).
3. Notice that produces an error.

Aljullu:
Fixes #2050.
Fixes #2090.

Allows free input in the table pagination input field, which allows the user to remove the current number to write a new one. It also prevents `<Pagination>` considering `0` a valid page, which was causing a 400 error.

### Screenshots
pagination

- Go to any report.
- Try removing the number in the table pagination input field.
- Write 0 and make sure it doesn't try to load that page.
- Write any valid number and verify it loads correctly.

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -1,5 +1,5 @@
 # unreleased
-- Pagination no longer considers `0` a valid input and triggers `onPageValueChange` on the input blur event.
+- Pagination no longer considers `0` a valid input and triggers `onPageChange` on the input blur event.
 - Tweaks to SummaryListPlaceholder height in order to better match SummaryNumber.
 
 # 2.0.0
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -1,3 +1,6 @@
+# unreleased
+- Pagination no longer considers `0` a valid input.
+
 # 2.0.0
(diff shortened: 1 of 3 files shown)


joshuatf (commented):
Adding `0` does not break the report now (does nothing- that's the expected response, right?) but came across another issue here.  It doesn't seem possible to get to any pages with double digits by manual entry (e.g., page 21).  

I imagine higher numbers are generally where manual entry is more useful.  Wondering if we should allow text entry and change page on field blur?

Aljullu:
> I imagine higher numbers are generally where manual entry is more useful. Wondering if we should allow text entry and change page on field blur?

I like this solution. It should be fixed in the last commits, thanks for the review @joshuatf!

joshuatf (approved):
Testing great- thanks for the changes @Aljullu!  LGTM :shipit:

PR #2094, opened 2019-04-22
Add order date filter to customers data store

### LINKED ISSUE #2017: Customers data store does not allow filtering by date
**Describe the bug**
`before` and `after` params are not options for the customers controller or endpoint.  The customers report does not currently have a date filter (but may benefit from one? @LevinMedia @josemarques ).

However, in leaderboard reports, this means that customer leaderboard stats will not change based on date.  Even if we're not adding a date range option to the customers report, we should provide these params as options for customers.

The customers data store joins on the order stats, so we can use the order dates to filter.

**To Reproduce**

1. Go to the dashboard.
2. Change the date.
3. Note that the customers leaderboard does not change.

**Expected behavior**
Expected to be able to filter customers by date.

joshuatf:
Fixes #2017

Allows customers to be filtered by order date `before` and `after` params.  Also hooks up the leaderboard to provide date based stats on customers.

### Screenshots
Screen Shot 2019-04-22 at 5 29 18 PM

1.  Make a request to `/wp-json/wc/v4/reports/customers` with `before` and `after` params.
2. Verify that returned stats are filtered by the order dates.
3.  Check that the dashboard customer's leaderboard now changes based on queried date.

### DIFF
--- a/includes/api/class-wc-admin-rest-leaderboards-controller.php
+++ b/includes/api/class-wc-admin-rest-leaderboards-controller.php
@@ -206,9 +206,11 @@ class WC_Admin_REST_Leaderboards_Controller extends WC_REST_Data_Controller {
 		$customers_data_store = new WC_Admin_Reports_Customers_Data_Store();
 		$customers_data       = $per_page > 0 ? $customers_data_store->get_data(
 			array(
-				'orderby'  => 'total_spend',
-				'order'    => 'desc',
-				'per_page' => $per_page,
+				'orderby'      => 'total_spend',
+				'order'        => 'desc',
+				'order_after'  => $after,
+				'order_before' => $before,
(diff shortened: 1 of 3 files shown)


Aljullu (approved):
Looks good to me and confirming the _Top Customers_ leaderboard now reacts to date changes. :shipit: 

On a side note: while reviewing this I noticed the customers endpoint returns a property named `date_last_order` which is not used anywhere else. Is there any reason to keep it or could we safely remove it in a follow-up?

joshuatf:
Thanks for reviewing @Aljullu!  

>On a side note: while reviewing this I noticed the customers endpoint returns a property named date_last_order which is not used anywhere else.

I think it was a column that an ambiguous name override that was eventually removed, but will check with @jeffstieler just to be sure.

PR #2102, opened 2019-04-23
Update how refunds are displayed in tables

### LINKED ISSUE #2002: Refunds should be distinguished from orders in the orders report.
Once #2001 is resolved, we should distinguish the refunds from orders in the orders report.  Currently, a refund displayed in the report will look exactly like an order but with a negative value.

It might be a good idea to:
* Note that it's a refund somewhere.
* Link to the parent order that the refund relates to.
* Note the negative amount in red.

Should we show the refund ID and the order ID?  Or just the order id?

Here's a tentative example to describe this, but check with @LevinMedia and @josemarques before implementing.
Screen Shot 2019-03-25 at 10 26 18 AM

joshuatf:
Fixes #2002 

Adds styling to negative net totals and links refunds to parent order IDs.

**Dev**
* We could add the `is-negative` class conditionally to a span inside `formatCurrency()` but I'm not sure if this would be ideal.  Alternatively, we could also create a new function, something like `formatCurrencySpan()` that wraps the currency with classes given the value.

**Design**
* We decided not to add the tag `[Refund]` next to the order ID, right?
* Order statuses for refunds are always `Completed`- okay to keep this in the table?

/cc @josemarques @LevinMedia 

### Screenshots
Screen Shot 2019-04-23 at 7 36 29 PM

1.  Drop the `wc_admin_version` option from your `wp_options` table to run the deltas on database tables.
2.  Create a refund.
3. Make sure the styling matches above for net totals.
4. Make sure order links go to the refund's parent order.

### DIFF
--- a/packages/currency/src/index.js
+++ b/packages/currency/src/index.js
@@ -69,3 +69,13 @@ export function getCurrencyFormatString( number ) {
 	}
 	return number.toFixed( precision );
 }
+
+export function renderCurrency( number, currencySymbol ) {
+	if ( 'number' !== typeof number ) {
+		number = parseFloat( number );
+	}
+	if ( number < 0 ) {
+		return <span className="is-negative">{ formatCurrency( number, currencySymbol ) }</span>;
+	}
+	return formatCurrency( number, currencySymbol );
+}
--- a/client/analytics/report/categories/table.js
+++ b/client/analytics/report/categories/table.js
@@ -86,7 +86,12 @@ class CategoriesReportTable extends Component {
 					value: items_sold,
 				},
 				{
-					display: formatCurrency( net_revenue ),
+					display:
+						net_revenue < 0 ? (
+							<span className="is-negative">{ formatCurrency( net_revenue ) }</span>
+						) : (
+							formatCurrency( net_revenue )
+						),
 					value: getCurrencyFormatDecimal( net_revenue ),
 				},
 				{
--- a/client/analytics/report/categories/table.js
+++ b/client/analytics/report/categories/table.js
@@ -10,7 +10,7 @@ import { map } from 'lodash';
 /**
  * WooCommerce dependencies
  */
-import { formatCurrency, getCurrencyFormatDecimal } from '@woocommerce/currency';
+import { formatCurrency, getCurrencyFormatDecimal, renderCurrency } from '@woocommerce/currency';
 import { getNewPath, getPersistedQuery } from '@woocommerce/navigation';
 import { Link } from '@woocommerce/components';
 import { numberFormat } from '@woocommerce/number';
@@ -86,12 +86,7 @@ class CategoriesReportTable extends Component {
 					value: items_sold,
 				},
 				{
-					display:
-						net_revenue < 0 ? (
-							<span className="is-negative">{ formatCurrency( net_revenue ) }</span>
-						) : (
-							formatCurrency( net_revenue )
-						),
+					display: renderCurrency( net_revenue ),
 					value: getCurrencyFormatDecimal( net_revenue ),
 				},
 				{
--- a/client/analytics/report/orders/table.js
+++ b/client/analytics/report/orders/table.js
@@ -178,7 +178,12 @@ export default class OrdersReportTable extends Component {
 					value: formattedCoupons.map( item => item.code ).join( ' ' ),
 				},
 				{
-					display: formatCurrency( net_total, currency ),
+					display:
+						net_total < 0 ? (
+							<span className="is-negative">{ formatCurrency( net_total, currency ) }</span>
+						) : (
+							formatCurrency( net_total, currency )
+						),
 					value: net_total,
 				},
 			];
--- a/client/analytics/report/orders/table.js
+++ b/client/analytics/report/orders/table.js
@@ -102,6 +102,7 @@ export default class OrdersReportTable extends Component {
 				num_items_sold,
 				order_id,
 				order_number,
+				parent_id,
 				status,
 			} = row;
 			const extended_info = row.extended_info || {};
@@ -133,7 +134,15 @@ export default class OrdersReportTable extends Component {
 				},
 				{
 					display: (
-						<Link href={ 'post.php?post=' + order_id + '&action=edit' } type="wp-admin">
+						<Link
+							href={
+								'post.php?post=' +
+								( parent_id ? parent_id : order_id ) +
+								'&action=edit' +
+								( parent_id ? '#order_refunds' : '' )
+							}
+							type="wp-admin"
+						>
 							{ order_number }
 						</Link>
 					),
--- a/client/analytics/report/orders/table.js
+++ b/client/analytics/report/orders/table.js
@@ -90,6 +90,17 @@ export default class OrdersReportTable extends Component {
 		];
 	}
 
+	getCustomerType( customerType ) {
+		switch ( customerType ) {
+			case 'new':
+				return _x( 'New', 'customer type', 'woocommerce-admin' );
+			case 'returning':
+				return _x( 'Returning', 'customer type', 'woocommerce-admin' );
+			default:
+				return _x( 'N/A', 'customer type', 'woocommerce-admin' );
+		}
+	}
+
 	getRowsContent( tableData ) {
 		const { query } = this.props;
 		const persistedQuery = getPersistedQuery( query );
@@ -155,10 +166,7 @@ export default class OrdersReportTable extends Component {
 					value: status,
 				},
 				{
-					display:
-						customer_type === 'new'
-							? _x( 'New', 'customer type', 'woocommerce-admin' )
-							: _x( 'Returning', 'customer type', 'woocommerce-admin' ),
+					display: this.getCustomerType( customer_type ),
 					value: customer_type,
 				},
 				{
--- a/client/analytics/report/orders/table.js
+++ b/client/analytics/report/orders/table.js
@@ -11,7 +11,7 @@ import { map } from 'lodash';
  */
 import { Date, Link, OrderStatus, ViewMoreList } from '@woocommerce/components';
 import { defaultTableDateFormat } from '@woocommerce/date';
-import { formatCurrency } from '@woocommerce/currency';
+import { formatCurrency, renderCurrency } from '@woocommerce/currency';
 import { numberFormat } from '@woocommerce/number';
 
 /**
@@ -187,12 +187,7 @@ export default class OrdersReportTable extends Component {
 					value: formattedCoupons.map( item => item.code ).join( ' ' ),
 				},
 				{
-					display:
-						net_total < 0 ? (
-							<span className="is-negative">{ formatCurrency( net_total, currency ) }</span>
-						) : (
-							formatCurrency( net_total, currency )
-						),
+					display: renderCurrency( net_total, currency ),
 					value: net_total,
 				},
 			];
--- a/client/analytics/report/products/table.js
+++ b/client/analytics/report/products/table.js
@@ -153,7 +153,12 @@ class ProductsReportTable extends Component {
 					value: items_sold,
 				},
 				{
-					display: formatCurrency( net_revenue ),
+					display:
+						net_revenue < 0 ? (
+							<span className="is-negative">{ formatCurrency( net_revenue ) }</span>
+						) : (
+							formatCurrency( net_revenue )
+						),
 					value: getCurrencyFormatDecimal( net_revenue ),
 				},
 				{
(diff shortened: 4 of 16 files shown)


ryelle:
I'm getting the following error when running this branch– is there something I should run to update the table?

WordPress database error: Unknown column `parent_id in `field list`

```
SELECT
	SUM( CASE WHEN parent_id = 0 THEN 1 ELSE 0 END ) as orders_count, SUM(num_items_sold) as num_items_sold, SUM(gross_total) AS gross_revenue, SUM(discount_amount) AS coupons, COUNT(DISTINCT coupon_id) as coupons_count, ABS( SUM( CASE WHEN gross_total &lt; 0 THEN gross_total END ) ) AS refunds, SUM(tax_total) AS taxes, SUM(shipping_total) AS shipping, SUM(net_total) AS net_revenue
FROM
	wp_wc_order_stats
	
LEFT JOIN
	wp_wc_order_coupon_lookup
	ON wp_wc_order_coupon_lookup.order_id = wp_wc_order_stats.order_id
WHERE
	1=1
	 AND wp_wc_order_stats.date_created &lt;= `2018-04-24 23:59:59` AND wp_wc_order_stats.date_created &gt;= `2018-04-01 00:00:00`
	 AND ( wp_wc_order_stats.status NOT IN ( `wc-trash`,`wc-pending`,`wc-failed`,`wc-cancelled` ) )
```

ryelle:
Re creating a wrapper function… what about `renderCurrency`? that way it's clear that it returns a component, while `format…` returns a string.

LevinMedia:
heya @joshuatf I'm unable to load the reports when I'm on this branch.

joshuatf:
@LevinMedia @ryelle I should have added more instructions for testing this.  This branch has a new database column so you'll need to re-run the database deltas to get this to work.  You can do that by deleting `wc_admin_version` from your `wp_options` table.

>Re creating a wrapper function… what about renderCurrency? 

I like this idea- thanks for this @ryelle!  Will get it added soon.

timmyc:
> Re creating a wrapper function… what about renderCurrency? that way it's clear that it returns a component, while format… returns a string.

Love this idea, and this component would also be likely of use in other projects throughout Woo, and could be made aware of any preferences of how to display negative numbers ( i.e. prefixed with a -, inside parens etc ).

joshuatf:
Added `renderCurrency()` to handle how these numbers are displayed.  This is ready for another look when you have time @LevinMedia and @ryelle.  Let me know if you have any more trouble testing.

packages/currency/src/index.js
```diff
@@ -69,3 +69,13 @@ export function getCurrencyFormatString( number ) {
 	}
 	return number.toFixed( precision );
 }
+
+export function renderCurrency( number, currencySymbol ) {
```
psealock: Seems like a good follow up would be to make this a component that recognizes negative formats, as @timmy mentuioned,

psealock (commented):
The code is looking good @joshuatf 

We should apply `renderCurrency` to other places in the report too:

Screen Shot 2019-05-08 at 3 10 13 PM
Screen Shot 2019-05-08 at 3 10 27 PM
Screen Shot 2019-05-08 at 3 10 32 PM

packages/currency/src/index.js
psealock: >So you think this should be its own component?

I do. A good use case is extensions creating a report page with tables displaying currency values. I can see this being really useful as we build out other screens in the future.

LevinMedia:
Heya @joshuatf - looking good! This is quite complex, so bear with me... 

**To begin, I made a partial refund of an existing order.** 

I refunded three items total. Two from one product, and one from another. This is the result: 

Screen Shot 2019-05-09 at 2 17 47 PM

- The negative net revenue is accurate, and it shows at the correct hour 👍  
- There are zero orders for the day - which is great, because a refund doesn't' count as a new order 🎉

taking a closer look at the tabular data:

Screen Shot 2019-05-09 at 2 27 45 PM

- Date looks good 
- Order number looks good (don't think we need to insert [refund] next to the order number), 
- The status looks good. In the case of a partial refund, the order status should remain whatever it was before the refund occurred. If a full refund is issued, the status should switch to refunded. This is a legacy WooCommerce issue that's not ideal, but I think it's the best we can do for now. 👍 
- we should show if the refund was issued by a new or returning customer. If the refund is on the customer's first order, they're still considered new, otherwise they're returning. 
- I like how we're showing which products were refunded, and the number of items that were returned in the refund 🎉 
- Items sold should show -3 ⚠️ 
- if a refund is issued without changing the number of items sold, we shouldn't show the products that were refunded, or a negative value for items sold (those columns should be blank) we should only show a negative net revenue.  
- The summary numbers at the bottom of the table look good other than items sold, which should show -3.

**I took a look through some of the other reports to see what was happening elsewhere:**

Revenue report:

Screen Shot 2019-05-09 at 2 49 49 PM

- Absent any new orders, gross revenue should be zero
- Correct order count, refund amount, and correct net revenue 🎉 

Product Report:

Screen Shot 2019-05-09 at 2 51 39 PM

- in a separate PR, we should remove the word "sold" from "Products Sold" in the footer summary number. 
- similar to the revenue report, the order count should be zero everywhere in this report as well. 

Category Report:

Screen Shot 2019-05-09 at 2 55 20 PM

- similar to the revenue report, the order count should be zero everywhere in this report as well. 

I wasn't able to test the coupons or taxes report, as the order I refunded didn't include either of those things, however, I'd expect them to behave in a similar fashion as the other reports, showing a negative value equal to the portion of the refund that was either tax or a coupon. 

In the process of testing this I noticed a bug in the customer report, which I'll log as a separate issue. As it's currently designed, it should always show ALL customers, from ALL time, however it seems to be inheriting whatever the last date range setting that was applied before the user navigates to the customer report. In any case, the bug exposes that the customer report is counting refunds as orders - the order count should be zero here as well: 

Screen Shot 2019-05-09 at 3 00 04 PM

joshuatf:
Thanks for this awesome and super thorough review, @LevinMedia!

>we should show if the refund was issued by a new or returning customer. If the refund is on the customer's first order, they're still considered new, otherwise they're returning.

I think we should keep this separate for a couple reasons.  I think returning/new customers should only apply to orders (since these are user actions) and not refunds (since those are admin issued).  The second and larger reason is that if we match the refund's returning/new status to the original order, we will invalidate existing returning/new customer logic.

E.g., customer makes an order on 3/1, makes another order on 4/1, and a refund is issued for the first order on 4/15.  If you query April, the customer would be considered "new" since the refund's customer type is going to match the original order.

>Items sold should show -3 ⚠️

Fixed- note that this requires database changes to work.  Dropping `wc_admin_version` from your `wp_options` table will make these changes for you.

>The summary numbers at the bottom of the table look good other than items sold, which should show -3.

Fixed by above solution as well.

>if a refund is issued without changing the number of items sold, we shouldn't show the products that were refunded, or a negative value for items sold (those columns should be blank) we should only show a negative net revenue.

Fixed!

>Absent any new orders, gross revenue should be zero

I believe this should be a negative number and not zero since refunds affect the gross revenue, but let me know if you have reason to believe this is incorrect.

>in a separate PR, we should remove the word "sold" from "Products Sold" in the footer summary number.

Will open a separate issue for this.  I agree that this should be changed, but it looks like we also use the term `products` in the orders report summary stats to represent the unique number of products sold.  Will this be confusing if the orders report is unique products and the products report represents total products but phrasing is the same?

>similar to the revenue report, the order count should be zero everywhere in this report as well.

Fixed!

>I wasn't able to test the coupons or taxes report, as the order I refunded didn't include either of those things, however, I'd expect them to behave in a similar fashion as the other reports, showing a negative value equal to the portion of the refund that was either tax or a coupon.

Taxes should be working as is and reflect negative amounts.  Coupons, however, I believe are unaffected as they can't be refunded.

Screen Shot 2019-05-10 at 10 55 35 AM

>however it seems to be inheriting whatever the last date range setting that was applied before the user navigates to the customer report.

We can address this in the issue you opened, but I imagine this is carried over from allowing date ranges to be used for customer leaderboards.

>In any case, the bug exposes that the customer report is counting refunds as orders - the order count should be zero here as well

Fixed!

packages/currency/src/index.js
joshuatf: Okay, sounds good @psealock!  Opened an issue to handle this component [here](https://github.com/woocommerce/woocommerce-admin/issues/2202).  We can also handle wrapping other currency values in that PR to keep scope down on this one.

psealock:
>As it's currently designed, it should always show ALL customers, from ALL time, however it seems to be inheriting whatever the last date range setting that was applied before the user navigates to the customer report.

I noticed this as well and went ahead and logged the issue, https://github.com/woocommerce/woocommerce-admin/issues/2214

LevinMedia:
>E.g., customer makes an order on 3/1, makes another order on 4/1, and a refund is issued for the first order on 4/15. If you query April, the customer would be considered "new" since the refund's customer type is going to match the original order.

In hindsight, I realize what I said doesn't really make sense. Are you proposing we don't show new or returning on a refund row? How do you feel about showing their new / returning status at the time the refund was issued? 

>I believe this should be a negative number and not zero since refunds affect the gross revenue, but let me know if you have reason to believe this is incorrect.

Refunds affect your net, not your gross. Gross would be the total of any new sales. From that, we subtract out things like shipping, taxes, and refunds to get net. So absent any new orders, with a single refund, gross would be zero, and net would be negative. 

>it looks like we also use the term products in the orders report summary stats to represent the unique number of products sold. Will this be confusing if the orders report is unique products and the products report represents total products but phrasing is the same?

Had to do a double take on this one. In both cases the count of "Products" in the summary numbers at the bottom of these reports are the same thing - the number of unique products that are included in the report. 

>Coupons, however, I believe are unaffected as they can't be refunded.

You can't refunded them directly, however, if you refund an item that originally had a coupon applied, that portion of the sale is returned to the customer, and noted in the order details screen... so potentially a coupon could have a negative value on a particular day - BUT after reviewing the IA of coupon report, I don't think we should worry about it - it's only showing the number of coupons used, and total discounts.

joshuatf:
Thanks for the feedback, @LevinMedia.  My thoughts below:

>In hindsight, I realize what I said doesn't really make sense. Are you proposing we don't show new or returning on a refund row? How do you feel about showing their new / returning status at the time the refund was issued?

I don't think we should display either in this row.  It's not an action that is created by a customer and I'm not sure having it provides useful information.  If we had to put something here, I can see using the original order's new/returning status, but this does come at a cost to changing out most of our existing logic for customer types.

>Refunds affect your net, not your gross. Gross would be the total of any new sales. From that, we subtract out things like shipping, taxes, and refunds to get net. So absent any new orders, with a single refund, gross would be zero, and net would be negative.

Refunds always affect gross revenue.  [Here's an article](https://www.accountingtools.com/articles/what-is-gross-revenue.html) that demonstrates how it's calculated.  They can potentially not affect net total if only tax and shipping fees are refunded.

>Had to do a double take on this one. In both cases the count of "Products" in the summary numbers at the bottom of these reports are the same thing - the number of unique products that are included in the report.

You're right on this one.  These are both unique so "sold" should be removed from the summary numbers.

>You can't refunded them directly, however, if you refund an item that originally had a coupon applied, that portion of the sale is returned to the customer, and noted in the order details screen

I see what you're saying here, but since we don't offer an explicit field to reduce the coupon value in a refund (like we do for individual products) and since coupons are artificial and don't involve any exchange of money, I don't think it would be correct to refund them in any form other than to remove them if an order was fully refunded.  I agree that we should not modify data there unless we decide to remove them on fully refunded orders.

timmyc:
Some great discussion happening here, but I wanted to see if we are feeling good about this as-is and get it merged in for the release.... we could always tackle some of the customer items noted above in follow-up PRs.

FWIW my .02 is not showing new/returning customer on refunds makes the most sense. It is an admin-based action that resulted in that row being there.... but also this:

>  but this does come at a cost to changing out most of our existing logic for customer types.

Kind of sounds like the level of effort to make a change here might be on the larger side of things, so I'd rather we discuss further and/or get some feedback from users on this particular item before deciding what to do.

LevinMedia:
@timmyc - @joshuatf and I caught up this morning - we resolved our differences on the gross / net discussion - Josh is going to update the gross to read zero, per my initial comment.  We also agreed we'd leave new vs returning out of the refund row for the time being.

joshuatf:
@LevinMedia was correct on gross revenue numbers here, so I updated those to reflect this.  This was a bit confusing since our back-end refers to these as gross total (which will remain negative), but updated this to not reflect when these numbers are summed up for gross revenue.

Also gave this branch a rebase and it's ready for another code review as well.  /cc @psealock if you have time this week.

psealock (approved):
Nice work every one, LGTM 🚢

PR #2112, opened 2019-04-25
solved #2111 - both checkbox is working.

### LINKED ISSUE #2111: Month to Due Checkbox is not working properly
**Describe the bug**
Analytic in all the "Month to due" checkbox does not properly work.

**To Reproduce**

1. Go to 'install new WoCoommerce and WooCommerce admin plugin'
2. Click on 'Analytic tab'
3. Click Month to due checkbox then the second one tries to uncheck the checkbox.

**Expected behavior**
Checkbox unchecked and checked when select and unselect.

**Video**
[https://www.screencast.com/t/nBDkCgkK52Pf](https://www.screencast.com/t/nBDkCgkK52Pf)

parmarhardip:
Fixes #2111 

### Accessibility

The checkbox is working please review it and let me know.

### DIFF
--- a/packages/components/src/chart/d3chart/legend.js
+++ b/packages/components/src/chart/d3chart/legend.js
@@ -101,7 +101,7 @@ class D3Legend extends Component {
 								onClick={ handleLegendToggle }
 								id={ `woocommerce-legend-${ instanceId }__item-button__${ row.key }` }
 								disabled={
-									( row.visible && numberOfRowsVisible <= 1 ) ||
+									( row.visible && numberOfRowsVisible <= 1 ) &&
 									( ! row.visible && numberOfRowsVisible >= selectionLimit ) ||
 									! interactive
 								}


timmyc:
Hello @parmarhardip and thanks for the PR and bug report! I think the current behavior is by design - meaning we don't want the chart to display empty, so we don't allow all data series in the charts to be turned off via the checkboxes.

Here is a screen gif of the item for those who might read this:

chart-data-series-toggle

The one thing I think we could potentially change here is applying a disabled state to the checkboxes of the single data series, to communicate better that you can not disable the series when it is the only data set active in the chart.

/cc @LevinMedia and @josemarques just to confirm that the current behavior is indeed by design and thoughts on disabling the checkbox if it can not be un-checked.

josemarques:
> just to confirm that the current behavior is indeed by design and thoughts on disabling the checkbox if it can not be un-checked.

Yeah, doesn't make sense to disable everything. thumbs up to that idea of disabling the checkbox when it's the only option toggled on.

timmyc:
Thanks @josemarques for confirming. As such I'm going to close this and the bug report issue out.

PR #2117, opened 2019-04-25
Add check for wp_set_script_translations.

### LINKED ISSUE #2116: Bug: Fatal when activating on WP Versions < 5.0
**Describe the bug**
While we have an explicit version check for WP Version 5.0 or greater, you can still install and attempt to activate wc-admin in earlier versions of WordPress, which results in the following fatal:

`Fatal error: Uncaught Error: Call to undefined function wp_set_script_translations() in /srv/www/4point9/public_html/wp-content/plugins/woocommerce-admin/includes/class-wc-admin-loader.php:233`

**To Reproduce**

1. Install wc-admin on 4.9.x of WordPress
2. Activate the plugin

**Expected behavior**
I anticipated the warning to be shown about WP < 5.0 not being supported and the plugin not activated.

### LINKED ISSUE #2118: WC version notice doesn't match WC version check
**Describe the bug**
- [The version check checks for WC 3.6](https://github.com/woocommerce/woocommerce-admin/blob/49008986a3cfb77b6bf8f132fc5947d736fe1d6e/woocommerce-admin.php#L115)
- [The notice says you need WC 3.5+](https://github.com/woocommerce/woocommerce-admin/blob/49008986a3cfb77b6bf8f132fc5947d736fe1d6e/woocommerce-admin.php#L82)

**To Reproduce**

1. Use WC 3.5.x
2. Activate wc-admin
3. Error notice displays that you need WC 3.5+, but WC 3.5.x is active.

**Expected behavior**
I'd like to know which WC version I actually need. :P

**Screenshots**
Screen Shot 2019-04-25 at 10 21 20 AM

timmyc:
Fixes #2116
Fixes #2118 

This branch fixes the fatal that is thrown when you attempt to activate the plugin on WordPress 4.9.x or earlier. There is also a couple of PHPCS fixes that were magically done on my behalf.

I also updated the admin alert that is shown to message the required version of WooCommerce being 3.6 now

- Install this branch on WordPress 4.9.x
- Attempt to activate the plugin
- Verify no fatal is thrown and you see the admin alert about WP 5.0 being required:

image

### DIFF
--- a/includes/class-wc-admin-loader.php
+++ b/includes/class-wc-admin-loader.php
@@ -150,8 +150,8 @@ class WC_Admin_Loader {
 			$feature = strtolower( $feature );
 			$file    = WC_ADMIN_FEATURES_PATH . $feature . '/class-wc-admin-' . $feature . '.php';
 			if ( file_exists( $file ) ) {
-				require_once( $file );
-				$feature = ucfirst( $feature );
+				require_once $file;
+				$feature         = ucfirst( $feature );
 				self::$classes[] = 'WC_Admin_' . $feature;
 			}
 		}
@@ -214,6 +214,10 @@ class WC_Admin_Loader {
 	 * Registers all the neccessary scripts and styles to show the admin experience.
 	 */
 	public static function register_scripts() {
+		if ( ! function_exists( 'wp_set_script_translations' ) ) {
+			return;
+		}
+
 		wp_register_script(
 			'wc-csv',
 			self::get_url( 'csv-export/index.js' ),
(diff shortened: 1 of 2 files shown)


rrennick (approved):
These versions also need to be updated:

https://github.com/woocommerce/woocommerce-admin/blob/fix/2116/woocommerce-admin.php#L12:13
https://github.com/woocommerce/woocommerce-admin/blob/fix/2116/woocommerce-admin.php#L82

Otherwise, this looks good. Pre-approving.

rrennick:
@timmyc Looks good 🚢

PR #2125, opened 2019-04-28
Components: fix README url

psealock:
Link in components README updated to `woocommerce-admin` from `wc-admin`. 

I'm pretty sure this is the only instance.


rrennick (approved):
There are also 3 `wc-admin` in https://github.com/woocommerce/woocommerce-admin/blob/master/docs/feature-flags.md

While I was searching I also found https://github.com/woocommerce/woocommerce-admin/blob/master/bin/install-wp-tests.sh#L188 . That is just a comment but the rest of the install script uses the name vs the slug (eg. WooCommerce) so it would be good to fix that one for consistency.

Pre-approving.

Issue #2121, opened 2019-04-26
Dashboard: Add Ability to Move Section Up/Down


timmyc:
Child Issue of #2009 / Dependent upon #2120 

See #2120 for a bit more background - but again for this enhancement I'm thinking we leverage the existing pattern of the Ellipsis menu to expose the ability to move a Dashboard section up/down in display. And again for this feature, I'm thinking we lean a bit on established patterns in the block editor:

![block-order](https://user-images.githubusercontent.com/22080/56834859-7657f580-6827-11e9-82f4-11baec29ddb8.gif)

My reasoning for using the Ellipsis menu again is we don't have an "edit state" for the dashboard. In some early designs, @jameskoster had a notion of a Dashboard Settings button - but I think exposing the option to change the order of the sections works well in the ellipsis menu too.

- Under the new "Customize" Menu Title section in the ellipsis menu, add options to move the section up or down in the display of the dashboard.
- Following the gif above - if a section is at the top already, "Move Up" should not be an option, and likewise if the section is already at the bottom "Move Down" should not be shown.
- Until we have a persistence layer that we are writing/reading from for dashboard preferences, lets just have the sections change order in the dashboard state

josemarques:
@timmyc here's the design for moving section up and down:

![Dashboard - Move block](https://user-images.githubusercontent.com/57050/56974129-29955700-6b66-11e9-9bd0-822be7c1b4eb.jpg)

This adds "Move up" and "Move down" controls to the section settings menu. If the section is at the top position the "Move up" option is not shown, and the same if the section is at the bottom position for the "Move down" option.

We opted for this approach instead of using the ellipsis Gutenberg pattern since that would require having a "edit mode" for this screen that would allow revealing the controls when the user hovered the block.

Issue #2122, opened 2019-04-26
Dashboard: Add Ability to Remove Section


timmyc:
Child Issue of #2009 / Dependent upon #2120 

See #2120 for a bit more background - but again for this enhancement I'm thinking we leverage the existing pattern of the Ellipsis menu to expose the ability to delete/remove a Dashboard section.

- Under the new "Customize" Menu Title section in the ellipsis menu, add an option to Delete the section.
- Thinking we might want to use a confirm dialog to verify removal though not totally certain on that ( /cc @josemarques )
- Until we have a persistence layer that we are writing/reading from for dashboard preferences, lets just have the section removed from display via component state in the dashboard when the delete action happens.

josemarques:
> leverage the existing pattern of the Ellipsis menu to expose the ability to delete/remove a Dashboard section.

This was already present on the block settings menu design, maybe it slipped through:

![Screenshot 2019-04-30 at 22 16 23](https://user-images.githubusercontent.com/57050/56993914-9c69f680-6b95-11e9-83e8-812957237ddf.png)

Here's the updated version with the controls to move the block: 

![Dashboard - Move block](https://user-images.githubusercontent.com/57050/56993950-ad1a6c80-6b95-11e9-8f22-24f6f6cc1ddc.jpg)

josemarques:
> Thinking we might want to use a confirm dialog to verify removal though not totally certain on that

@timmyc 
Since this action is close to the options to move the block up and down there can be a few hits by mistake when trying to perform another action, so that sounds like a good idea.

timmyc:
Thanks @josemarques! Are we calling these areas blocks or sections for the users?

josemarques:
That's good question. We have been using blocks in all the pieces :

![Screenshot 2019-04-30 at 22 36 26](https://user-images.githubusercontent.com/57050/56994980-72fe9a00-6b98-11e9-8490-f39c6da65568.png)

But section actually makes more sense as a combination of a title and a few blocks.

@LevinMedia @jameskoster any thoughts on this?

PR #2128, opened 2019-04-29
Add/moveable dashboard sections

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/2121

* Add the ability to move Dashboard Sections using an Editor-Block like up/down mover.
* Introduce a filter to add Dasboard Sections, `woocommerce_dashboard_sections`.

Screen Shot 2019-04-29 at 3 09 15 PM

### Accessibility

- [ ] All text has [at least a 4.5 color contrast with its background](https://webaim.org/resources/contrastchecker/)=

### DIFF
--- a/client/dashboard/components/section-mover/index.js
+++ b/client/dashboard/components/section-mover/index.js
@@ -10,16 +10,20 @@ import { __ } from '@wordpress/i18n';
 import './style.scss';
 import { IconButton } from '@wordpress/components';
 
-const SectionMover = () => {
+const SectionMover = ( { isFirst, isLast, onMoveUp, onMoveDown } ) => {
 	return (
 		<div className="woocommerce-section-mover">
 			<IconButton
 				className="editor-block-mover__control block-editor-block-mover__control"
+				onClick={ isFirst ? null : onMoveUp }
+				aria-disabled={ isFirst }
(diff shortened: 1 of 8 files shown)


psealock:
### Design Feedback Request

cc @josemarques @LevinMedia 

When looking at a method to move Dashboard sections, there seems to be three options (so far). 

1. Gutenberg style move element. I like the familiarity of this one, but usability suffers because drag and drop isn't ideal with large elements and an edit functionality makes the interface busy.

Screen Shot 2019-04-29 at 3 09 15 PM

2. Move element inside the Ellipsis. This is cleaner and hides a not-often used feature.

Screen Shot 2019-04-30 at 9 07 16 AM

3. Dashboard wide settings Ellipsis. #2122 calls for a method to remove a Dashboard Section. Since a dropdown ellipsis menu makes sense to add/remove sections, we can place the arrows as seen (crudely) below.

Screen Shot 2019-04-30 at 9 39 26 AM

I'd like to proceed with option 3, but want to get some feedback before proceeding.

psealock:
Now that we have feedback in https://github.com/woocommerce/woocommerce-admin/issues/2121 i'm closing this one to start fresh

diouldebalde:
une vivant sur la lune

PR #2131, opened 2019-04-29
Allow renaming Dashboard sections

### LINKED ISSUE #2120: Dashboard: Add ability to edit section name
Child issue of #2009

While I'm not fully certain the ultimate design direction of this feature, I'm hoping we can still get started on implementing the functionality. To enable the dashboard to be fully customizable, we need to offer the ability to change the name of each section:

section-titles

My proposal for this task, until further design direction is given ( and /cc @josemarques for input/thoughts/killer design ideas ), is to create a section ( `<MenutTitle/>` ) inside the Ellipsis menu which will house options to customize section options like this.

ellipsis-menu

Ultimately this could also be where there are options to move a section up/down in the display order too. This approach is based off of the pattern seen in the gutenberg sidebar for editing block text-based attributes:

image

For the scope of this issue, since we don't have persistence of dashboard options yet, is to just have the name change reflected in the state of the section header... i.e. it will return to the normal / default value on page re-load.

So thinking:

- Create new `<MenuTitle />` for "Customize" which only appears when the `dashboard/customizable` config is [true](https://github.com/woocommerce/woocommerce-admin/blob/master/config/development.json#L6)
- Add a text field / label for "Section Title" that defaults to the current section title name
- onChange of this field, update the Section Title that is displayed in the section via component state

Aljullu:
Fixes #2120.

### Accessibility

### Screenshots
dashboard-titles

- Go to the _Dashboard_.
- Open any section ellipsis menu and verify there is an input text to rename the section.
- Rename the section and verify the title is updated when you unfocus the input.
- Rename the section again leaving the input empty.
- Verify instead of an empty title, you get the default title.
- Verify when you close/open the ellipsis menu, the input text was also reset to the default title.

### DIFF
--- a/client/dashboard/customizable.js
+++ b/client/dashboard/customizable.js
@@ -25,19 +25,16 @@ export default class CustomizableDashboard extends Component {
 				{
 					key: 'store-performance',
 					component: StorePerformance,
-					defaultTitle: __( 'Store Performance', 'woocommerce-admin' ),
 					title: __( 'Store Performance', 'woocommerce-admin' ),
 				},
 				{
 					key: 'charts',
 					component: DashboardCharts,
-					defaultTitle: __( 'Charts', 'woocommerce-admin' ),
 					title: __( 'Charts', 'woocommerce-admin' ),
 				},
 				{
 					key: 'leaderboards',
 					component: Leaderboards,
-					defaultTitle: __( 'Leaderboards', 'woocommerce-admin' ),
 					title: __( 'Leaderboards', 'woocommerce-admin' ),
 				},
 			] ),
@@ -72,7 +69,6 @@ export default class CustomizableDashboard extends Component {
 					return (
 						<Section
 							component={ section.component }
-							defaultTitle={ section.defaultTitle }
 							key={ section.key }
 							onTitleUpdate={ this.onSectionTitleUpdate( section.key ) }
 							path={ path }
--- a/client/dashboard/customizable.js
+++ b/client/dashboard/customizable.js
@@ -25,16 +25,19 @@ export default class CustomizableDashboard extends Component {
 				{
 					key: 'store-performance',
 					component: StorePerformance,
+					defaultTitle: __( 'Store Performance', 'woocommerce-admin' ),
 					title: __( 'Store Performance', 'woocommerce-admin' ),
 				},
 				{
 					key: 'charts',
 					component: DashboardCharts,
+					defaultTitle: __( 'Charts', 'woocommerce-admin' ),
 					title: __( 'Charts', 'woocommerce-admin' ),
 				},
 				{
 					key: 'leaderboards',
 					component: Leaderboards,
+					defaultTitle: __( 'Leaderboards', 'woocommerce-admin' ),
 					title: __( 'Leaderboards', 'woocommerce-admin' ),
 				},
 			] ),
@@ -69,6 +72,7 @@ export default class CustomizableDashboard extends Component {
 					return (
 						<Section
 							component={ section.component }
+							defaultTitle={ section.defaultTitle }
 							key={ section.key }
 							onTitleUpdate={ this.onSectionTitleUpdate( section.key ) }
 							path={ path }
--- a/client/dashboard/customizable.js
+++ b/client/dashboard/customizable.js
@@ -12,6 +12,7 @@ import { applyFilters } from '@wordpress/hooks';
 import './style.scss';
 import DashboardCharts from './dashboard-charts';
 import Leaderboards from './leaderboards';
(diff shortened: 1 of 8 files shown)


client/dashboard/customizable.js
```diff
@@ -16,15 +17,65 @@ import StorePerformance from './store-performance';
...
 				<H>{ __( 'Customizable Dashboard', 'woocommerce-admin' ) }</H>
 				<ReportFilters query={ query } path={ path } />
-				<StorePerformance query={ query } />
-				<DashboardCharts query={ query } path={ path } />
-				<Leaderboards query={ query } />
+				{ sections.map( section => {
+					return (
+						<div className="woocommerce-dashboard-section" key={ section.key }>
+							<section.component
+								onTitleUpdate={ this.onSectionTitleUpdate( section.key ) }
```
psealock: Nice solution here with `<section.component>` for inclusion of props.

client/dashboard/store-performance/index.js
```diff
@@ -35,9 +36,16 @@ import './style.scss';
...
+		const { title } = props;
+
 		this.state = {
 			hiddenIndicators: props.hiddenIndicators || [],
+			titleInput: title || __( 'Store Performance', 'woocommerce-admin' ),
 		};
+
+		this.onMenuToggle = this.onMenuToggle.bind( this );
+		this.onTitleChange = this.onTitleChange.bind( this );
+		this.onTitleBlur = this.onTitleBlur.bind( this );
```
psealock: What do you think about moving these three components to `<CustomizableDashboard />`?

This way its less code to write when adding Sections (for extensions as well as us) and `onMenuToggle` can read the title from state when `titleInput === ''`. Plus 2 new methods (remove and moveUp/moveDown) will also follow the same pattern so it would be nice to have all in the same place.

psealock (commented):
Looking good @Aljullu, a comment for you on code organization.

client/dashboard/store-performance/index.js
Aljullu: Agree with that, there is a lot of duplicated code there. Instead of moving this into `<CustomizableDashboard />` I created a new component named `<Section />` and moved those events and the `titleInput` state there (e2c339a3a6e27d69c379353eec9a9256c01371c0). The reason to create a new component instead of moving everything to `<CustomizableDashboard />` is that I think the input state doesn't need to be sent up to the dashboard on every key stroke and it's better to have it contained in the section. So the Dashboard only gets updated when the input is unfocused.

I imagine it's going to be different for the moveUp/moveDown buttons, since they don't need an internal state.

Let me know how this solution looks!

timmyc (commented):
This is super slick! I did run into a weird scenario where it seems there might be some state getting mixed up between <Section/> instances - this gif shows it best:

https://cloudup.com/cIt55OHs1jt

client/dashboard/section.js
```diff
@@ -0,0 +1,58 @@
+/** @format */
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { Component } from '@wordpress/element';
+
+export default class StorePerformance extends Component {
```
psealock: Should be called `Section`

client/dashboard/section.js
```diff
@@ -0,0 +1,58 @@
...
+	constructor( props ) {
+		super( props );
+		const { title } = props;
+
+		this.state = {
+			titleInput: title || __( 'Store Performance', 'woocommerce-admin' ),
```
psealock: You can remove the `||` part of this statement. `onMenuToggle` should simply not `setState` if string no string is entered. This is what Timmy noticed in his comment.

client/dashboard/section.js
```diff
@@ -0,0 +1,58 @@
...
+		};
+
+		this.onMenuToggle = this.onMenuToggle.bind( this );
+		this.onTitleChange = this.onTitleChange.bind( this );
+		this.onTitleBlur = this.onTitleBlur.bind( this );
+	}
+
+	onMenuToggle( isOpen ) {
```
psealock: Can we remove this function in favour of placing the logic in `onTitleBlur`? Just a thought here, I haven't tested the assumption that `onTitleBlur` will definitely get called when closing the dropdown.

psealock (commented):
>I created a new component named `<Section />`

Cool, looking good. I left a few comments, one related to what Timmy noticed.

Aljullu:
Thanks for the reviews and for finding that bug @timmyc @psealock! It should be fixed now and all other comments should be addressed as well.

psealock (approved):
LGTM, thanks for the changes.

PR #2134, opened 2019-04-30
Dashboard: Section add, remove, move up, and move down

### LINKED ISSUE #2121: Dashboard: Add Ability to Move Section Up/Down

### LINKED ISSUE #2122: Dashboard: Add Ability to Remove Section

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/2122
Fixes https://github.com/woocommerce/woocommerce-admin/issues/2121

Add the ability to add/remove Dashboard Sections as well as move them up/down.

### Testing

1. Go to Dashboard.
2. In a Section's Ellipsis Menu, remove a section or move it up/down
3. Add a section back by going to the "plus" button at the bottom
Screen Shot 2019-05-02 at 8 33 45 PM
4. Play around and see if anything breaks.

### DIFF
--- a/client/dashboard/components/section-controls/style.scss
+++ b/client/dashboard/components/section-controls/style.scss
@@ -1,0 +1,14 @@
+/** @format */
+
+.woocommerce-section-controls {
+	border-top: $border-width solid $core-grey-light-500;
+
+	.dashicon {
+		margin: 0 $gap-smaller 0 0;
+		vertical-align: bottom;
+	}
+
+	.woocommerce-ellipsis-menu__item {
+		padding-bottom: 10px;
+	}
+}
--- a/client/dashboard/dashboard-charts/index.js
+++ b/client/dashboard/dashboard-charts/index.js
@@ -25,6 +25,7 @@ import ChartBlock from './block';
 import { getChartFromKey, uniqCharts } from './config';
 import withSelect from 'wc-api/with-select';
 import './style.scss';
+import SectionControls from 'dashboard/components/section-controls';
 
 class DashboardCharts extends Component {
 	constructor( props ) {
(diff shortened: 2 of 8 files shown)


packages/components/src/ellipsis-menu/index.js
```diff
@@ -33,8 +33,12 @@ class EllipsisMenu extends Component {
 			);
 		};
 
-		const renderContent = () => (
-			<NavigableMenu className="woocommerce-ellipsis-menu__content">{ children }</NavigableMenu>
+		// @todo Make all children rendered by render props so Dropdown args can be passed?
```
psealock: `EllipsisMenu` is essentially a wrapper for `Dropdown`, but by rendering children instead of using render props [like `Dropdown` does](https://github.com/WordPress/gutenberg/blob/master/packages/components/src/dropdown/index.js#L94), we're removing built in functionality. In this case, use of argument `onToggle` so that dropdown contents can have access to this argument.

In a follow up PR, I'd like to change `EllipsisMenu` such that it accepts a `renderContent` prop instead of children.

client/dashboard/components/section-controls/style.scss
```diff
@@ -0,0 +1,14 @@
...
+
+.woocommerce-section-controls {
+	border-top: $border-width solid $core-grey-light-500;
+
+	.dashicon {
+		margin: 0 $gap-smaller 0 0;
+		vertical-align: bottom;
+	}
+
+	.woocommerce-ellipsis-menu__item div {
```
Aljullu: What would you think about rewriting this CSS ruleset, so we don't need the wrapping `<div>`?
```CSS
.woocommerce-ellipsis-menu__item {
	padding-bottom: 12px;
}
```

client/dashboard/dashboard-charts/index.js
```diff
@@ -76,21 +77,27 @@ class DashboardCharts extends Component {
...
+									isCheckbox
+									isClickable
+									key={ chart.key }
+									onInvoke={ this.toggle( chart.key ) }
+								>
+									{ __( `${ chart.label }`, 'woocommerce-admin' ) }
+								</MenuItem>
+							);
+						} ) }
+						<SectionControls onToggle={ onToggle } { ...this.props } />
```
Aljullu: I'm a bit afraid passing down all props (`{ ...this.props }`) to the `SectionControls` will become messy in the future as we keep adding new props to each section. How about declaring them one-by-one? It makes code longer and more verbose but I think it will help to understand which props are required by `SectionControls` and which don't.

Aljullu (commented):
Wow, this looks amazing and works like a charm! :heart_eyes: 

One suggestion: could we move sections to the bottom when they are enabled? If I have sections in this order:
1. _Performance_
2. _Charts_
3. _Leaderboards_

Imagine I remove _Performance_ and add it again. I would expect _Performance_ to be added below the _Leaderboards_, but currently it keeps the position it had before being removed. Not sure if it's only me but that felt a bit counter-intuitive to me.

In case a gif makes it clearer:
order

psealock:
@Aljullu Thanks for the review.

I changed the CSS rule to remove an extra `div`, removed `...this.props` in favor of being verbose, and changed the behavior such that added Sections get appended to the bottom of the dashboard.

I also rebased with master.

Aljullu (approved):
Thanks for the changes @psealock! Looking good. :shipit:

PR #2139, opened 2019-04-30
add defaultOrder field to tablecard header

### LINKED ISSUE #2136: TableCard shows descending chevron when query is asc
**Describe the bug**
On initial load the TableCard component shows the descending chevron on the default sort column unless the query passed has an `order` of uppercase `ASC`. Lowercase `asc` is treated as `DESC`.

**To Reproduce**

1. Clone https://github.com/Prospress/action-scheduler-admin/tree/fix/5 into your plugins folder.
2. Activate `Action Scheduler Admin`
3. Run `npm install && npm run start` in that plugin folder
4. In the dashboard go to Analytics -> Scheduled Actions
5. See that the `Scheduled` column chevron show descending while the column is sorted ascending.

**Expected behavior**
Ascending chevron should show.

**Screenshots**

Screen shot to show the default query is populated with `order: 'asc'`

Screen Shot 2019-04-30 at 9 33 45 AM

rrennick:
Fixes #2136

This PR adds a `defaultOrder` field to the headers prototype to allow specifying whether the default sort column is initially sorted in ascending or descending order.

### Screenshots

Screenshot showing `query` is an empty object, header `defaultOrder` set, chevron showing ascending:

Screen Shot 2019-04-30 at 3 15 26 PM

- Tomorrow I'll update my example component branch to use `defaultOrder`
- Alternative would be to test by modifying the example component in this repo

### DIFF
--- a/packages/components/src/table/table.js
+++ b/packages/components/src/table/table.js
@@ -116,7 +116,7 @@ class Table extends Component {
 			'is-scrollable': isScrollable,
 		} );
 		const sortedBy = query.orderby || get( find( headers, { defaultSort: true } ), 'key', false );
-		const sortDir = query.order || DESC;
+		const sortDir = query.order || get( find( headers, { key: sortedBy } ), 'defaultOrder', DESC );
 
 		return (
 			<div
@@ -248,6 +248,10 @@ Table.propTypes = {
 			 * Boolean, true if this column is the default for sorting. Only one column should have this set.
 			 */
 			defaultSort: PropTypes.bool,
+			/**
+			 * String, asc|desc if this column is the default for sorting. Only one column should have this set.
+			 */
+			defaultOrder: PropTypes.string,
 			/**
 			 * Boolean, true if this column should be aligned to the left.
 			 */


rrennick:
I noticed today that clicking the chevron the first time updates the query object from empty to "asc" even when it's already set to default "asc".

rrennick:
> clicking the chevron the first time updates the query object from empty to "asc" even when it's already set to default "asc".

This was fixed in https://github.com/woocommerce/woocommerce-admin/pull/2139/commits/2ef82bc99785706fe49cc8c3c5d365d7540db889

rrennick:
> you'll just want to add a note to the changelog to go along with this.

Added. I'm getting a linter error after merging in master:

```shell
.../wp-content/plugins/woocommerce-admin/client/dashboard/store-performance/index.js
  202:5  error  'hiddenIndicators' is not defined  no-undef
```

psealock:
Weird how that slipped through. This line

https://github.com/woocommerce/woocommerce-admin/blob/99087706ff3e3c1da276aed549bbefc71664525a/client/dashboard/store-performance/index.js#L202

Should be `hiddenIndicators: [],`. cc @Aljullu just to confirm. I tried it out and it works as expected with no indicators.

PR #2148, opened 2019-05-02
Save section/block preferences in user data

### LINKED ISSUE #2040: Dashboard: Create Section/Block Preference Structure with Defaults
Child Issue of #2009

Builds upon #1973

In #1973 a new set of default performance indicators, charts, and leaderboards was put in place. Now that we are embarking on making the dashboard fully customizable, we need to take those defaults, and implement a new persistence object to house the dashboard preferences.

Thinking something like:

```
[
    {
        type: `summary|chart|leaderboard`
        label: `User's label for section`
        blocks: [ `array`, `of`, `blocks` ]
    },
    ....
]
```

I believe we can just use the order of the config objects within the array to persist the display order. And the block display order we agreed doesn't need to be user-configurable/sortable.

The default config should map to what exists on the current dashboard iteration, and the blocks should be those that were chosen #1973

Aljullu:
Fixes #2040.

### Screenshots
image

- Modify section names and enable/disable some blocks.
- Reload the page.
- Verify changes are persisted.


client/dashboard/section.js
```diff
@@ -3,16 +3,31 @@
...
 export default class Section extends Component {
 	constructor( props ) {
 		super( props );
 		const { title } = props;
 
+		this.components = {
+			'store-performance': StorePerformance,
+			charts: DashboardCharts,
+			leaderboards: Leaderboards,
+		};
```
psealock: This is going to pose a problem for extensions adding Sections.

Could you add the component somehow in this part of `applyFilters` in customizable.js?

```js
sections: applyFilters(
    'woocommerce_dashboard_sections',
    props.userPrefSections || defaultSections
),
```
A slightly less clean approach could be a `registerDashboardComponent` filter. What do you think?

Aljullu: Good point. I refactored it in 1caa9d19f37cd96422be8d937c8ec9f8970f6763. Let me know how it looks.

psealock (approved):
Clever solution here @Aljullu. I like the ability to supply a section and have it modified by user preferences in a merge.

Nice refactor too to remove duplicate logic.

I'm approving this but I can also give it another look if you'd like after dealing with merge conflicts.

client/dashboard/customizable.js
```diff
@@ -4,56 +4,79 @@
...
+
+	mergeSections( defaultSections, prefSections ) {
+		const defaultKeys = defaultSections.map( section => section.key );
+		const prefKeys = prefSections.map( section => section.key );
+		const keys = new Set( [ ...prefKeys, ...defaultKeys ] );
+		const sections = [];
+
+		keys.forEach( key => {
+			const prefSection = prefSections.find( section => section.key === key );
+			const defaultSection = defaultSections.find( section => section.key === key );
```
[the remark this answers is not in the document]
Aljullu: Yes, that's to be expected. When you hadn't done any modification to the _Dashboard_, `prefSections` will be empty but that doesn't cause an error when merging the section.

I [added a check](b5fe9cdf1db6ffde39188bb80c20bd59ac1cdf59) to make sure `defaultSection` is defined, though. Since that one is required because it contains the information of which component to use.

client/dashboard/customizable.js
```diff
@@ -4,56 +4,79 @@
...
-					title: __( 'Charts', 'woocommerce-admin' ),
-				},
-				{
-					key: 'leaderboards',
-					component: Leaderboards,
-					title: __( 'Leaderboards', 'woocommerce-admin' ),
-				},
-			] ),
+			sections: this.mergeSections(
+				applyFilters( 'woocommerce_dashboard_sections', sectionsDefinition ),
```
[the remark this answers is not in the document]
Aljullu: Everything sounds good. I made the changes in b02b0d594b8411e2f00ba0b842bd8da40f8dd3b9 and added the README in 6c8d586eb129cefe1d5a35fe2e7c9c81c55071d0.

Aljullu:
Thanks for the review @psealock! I implemented your suggestions but I think it would be great if you could take just one more look anyway, since the merge was quite big and I would like your opinion on the README.

client/dashboard/README.md
```diff
@@ -0,0 +1,48 @@
...
+			title: 'My Example Dashboard Section',
+			isVisible: true,
+			icon: 'arrow-right-alt',
+			hiddenBlocks: [],
+		},
+	];
+} );
+```
+
+Each report is defined by an object containing `key`, `title`, `component` and `hiddenBlocks`.
```
timmyc: Should this say `Each section is ...`

psealock:
```suggestion
Each report is defined by an object containing the following properties.
```

psealock (approved):
LGTM! Working really well, thanks for bringing this together. Just a small suggestion.

Aljullu:
Thanks for the reviews @timmyc @psealock!

PR #2150, opened 2019-05-02
add a defaultValue parameter to the ReportFilters component

### LINKED ISSUE #2149: Add defaultValue parameter to ReportFilters
**Describe the bug**
To display correctly on initial page load the ReportFilters component requires

- An `all` filter be included in the filters
- The `all` value be the value used for the initial data retrieval in the parent component

Both of these restrictions limit the WC developer community taking full advantage of this component.

rrennick:
Fixes #2149

This PR adds an optional `defaultValue` parameter to the ReportFilters component to be used before the default `all` filter. This eliminates both the restrictions identified in the issue. 

- Install my example component https://github.com/Prospress/action-scheduler-admin/pull/6
- Go to Analytics -> Scheduled Actions
- `Pending` status is selected in the Report Filters dropdown component
- Use sorting and filtering to verify the component continues to work as expected.

### DIFF
--- a/client/analytics/components/report-chart/utils.js
+++ b/client/analytics/components/report-chart/utils.js
@@ -22,7 +22,7 @@ export function getSelectedFilter( filters, query, selectedFilterArgs = {} ) {
 
 	if ( filterConfig.showFilters( query, selectedFilterArgs ) ) {
 		const allFilters = flattenFilters( filterConfig.filters );
-		const value = query[ filterConfig.param ] || DEFAULT_FILTER;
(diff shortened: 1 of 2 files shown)


psealock:
Hi @rrennick, in looking into your comment https://github.com/Prospress/action-scheduler-admin/pull/6#issuecomment-490056487 I dug further into this one and saw instances of `'all'` being applied without first checking for a `defaultValue` property in the config.

I pushed changes to fix that, I hope you don't mind.

PR #2154, opened 2019-05-03
EllipsisMenu: Refactor children to renderContent

psealock:
Follow up of #2134 

Since `EllipsisMenu` is essentially a wrapper for Gutenberg's `Dropdown`, we can remove the `children` prop in favour of a renderProp `renderContent` so that function arguments `isOpen`, `onToggle`, and `onClose` can be passed down. 

There should be no visible changes. Load up the app and make sure nothing breaks.


Aljullu (approved):
Everything is looking good to me and it's testing well. Thanks for improving the ellipsis menu @psealock!

PR #2157, opened 2019-05-03
Don't make an API call when there are no performance indicators

### LINKED ISSUE #2153: Removing all store performance indicators produces error 500
1. Open the _Network_ tab of your browser devtools.
2. Remove all _Store Performance_ blocks.
3. Notice the call `/wc/v4/reports/performance-indicators` returns error 500.
image

Aljullu:
Fixes #2153.

This will prevent error 500 from happening when querying the Rest API endpoint with no `stats` parameter.

1. Open the _Network_ tab of your browser devtools.
2. Remove all _Store Performance_ blocks in the _Dashboard_.
3. Reload the page and verify there are no calls to the endpoint `/wc/v4/reports/performance-indicators`.

### DIFF
--- a/client/dashboard/store-performance/index.js
+++ b/client/dashboard/store-performance/index.js
@@ -177,14 +177,16 @@ class StorePerformance extends Component {
 	}
 
 	render() {
-		const { title } = this.props;
+		const { userIndicators, title } = this.props;
 		return (
 			<Fragment>
 				<SectionHeader
 					title={ title || __( 'Store Performance', 'woocommerce-admin' ) }
 					menu={ this.renderMenu() }
 				/>
-				<div className="woocommerce-dashboard__store-performance">{ this.renderList() }</div>
+				{ userIndicators.length > 0 && (
+					<div className="woocommerce-dashboard__store-performance">{ this.renderList() }</div>
+				) }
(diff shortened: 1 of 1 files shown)


joshuatf (commented):
Code looks good here, though I wasn't able to recreate the 500 error by hiding all charts.  Anything in particular we need to do to get the 500 error?

Also, probably out of scope for this PR, but the lack of charts/message when no indicators is present is a bit weird. /cc @LevinMedia 

Screen Shot 2019-05-06 at 4 21 36 PM

Aljullu:
Thanks for testing @joshuatf!

> Code looks good here, though I wasn't able to recreate the 500 error by hiding all charts. Anything in particular we need to do to get the 500 error?

Not really, removing all elements from the _Performance_ section should be enough.


Can you check if the query to `performance-indicators` has these params (notice missing `stats`)?
image

> Also, probably out of scope for this PR, but the lack of charts/message when no indicators is present is a bit weird. /cc @LevinMedia

Agree with that. For now, I hid the gray line (727aacd0c7911824640fdc8c1f10cce44c830aae), which was a bit out of place:
image

psealock (approved):
LGTM!

PR #2177, opened 2019-05-07
Hook up API Historical Data screen to API endpoints

Aljullu:
Part of #1851.

This PR hooks up the _Historical Data Import_ buttons to the REST API endpoints. Notice the endpoints to read the current state of the import are not available yet, so progress bars and the transition between pages don't work yet.

### Accessibility

### Screenshots
image

- Go to the _Import Historical Data_ section of the _Settings_ page.
- Press _Start_ and verify a call to `/wc/v4/reports/import` is made.
- Try modifying the select box value (try with _Custom_, which will enable the date picker) and the _Skip_ checkbox and start the import again, verify the call to the endpoint has the correct parameters.
- Modify `<HistoricalData>` [props](https://github.com/woocommerce/woocommerce-admin/blob/master/client/analytics/settings/historical-data/index.js#L165) to:
```ES6
hasImportedData: true,
```
- Reload the page and press on _Delete Previously Imported Data_.
- Verify a call to `/wc/v4/reports/import/delete` is made.
- Modify `<HistoricalData>` props again to:
```ES6
inProgress: true,
```
- Press _Stop Import_ and verify a call to `/wc/v4/reports/import/cancel` is made.

### DIFF
--- a/client/analytics/settings/historical-data/actions.js
+++ b/client/analytics/settings/historical-data/actions.js
@@ -11,6 +11,9 @@ function HistoricalDataActions( {
 	customersTotal,
 	hasImportedData,
 	inProgress,
+	onDeletePreviousData,
+	onStartImport,
+	onStopImport,
 	ordersProgress,
 	ordersTotal,
 } ) {
@@ -22,7 +25,7 @@ function HistoricalDataActions( {
 					<Button
 						className="woocommerce-settings-historical-data__action-button"
 						isPrimary
-						onClick={ () => null }
+						onClick={ onStopImport }
 					>
 						{ __( 'Stop Import', 'woocommerce-admin' ) }
 					</Button>
@@ -44,7 +47,7 @@ function HistoricalDataActions( {
 		// Has no imported data
 		if ( ! hasImportedData ) {
 			return (
-				<Button isPrimary onClick={ () => null }>
+				<Button isPrimary onClick={ onStartImport }>
 					{ __( 'Start', 'woocommerce-admin' ) }
 				</Button>
 			);
@@ -57,7 +60,7 @@ function HistoricalDataActions( {
 					<Button isDefault onClick={ () => null }>
 						{ __( 'Re-import Data', 'woocommerce-admin' ) }
 					</Button>
-					<Button isDefault onClick={ () => null }>
+					<Button isDefault onClick={ onDeletePreviousData }>
 						{ __( 'Delete Previously Imported Data', 'woocommerce-admin' ) }
 					</Button>
 				</Fragment>
@@ -67,10 +70,10 @@ function HistoricalDataActions( {
 		// It's not in progress and has some imported data
 		return (
 			<Fragment>
-				<Button isPrimary onClick={ () => null }>
+				<Button isPrimary onClick={ onStartImport }>
 					{ __( 'Start', 'woocommerce-admin' ) }
 				</Button>
-				<Button isDefault onClick={ () => null }>
+				<Button isDefault onClick={ onDeletePreviousData }>
 					{ __( 'Delete Previously Imported Data', 'woocommerce-admin' ) }
 				</Button>
 			</Fragment>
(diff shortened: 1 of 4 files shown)


rrennick:
> Go to the Import Historical Data section of the Settings page.

I don't have this section on my settings screen. Is there something I need to do in my config and/or DB to get this section.

Screen Shot 2019-05-07 at 10 36 20 AM

Aljullu:
> I don't have this section on my settings screen. Is there something I need to do in my config and/or DB to get this section.

@rrennick you might need to pull the latest changes from `master`, since this was added in #2073, which I merged today.

rrennick:
I have my AS queue runner disabled & I noticed that if I click start multiple times with or without changing the settings I get two new scheduled tasks each time. Should we cancel existing `pending` actions for this before scheduling the new ones? 

A use case I'm thinking of is if a user initially chose 365 days then changed their mind to 90 days. The existing 365 day task will run before the 90 day one because it has the older scheduled date.

Screen Shot 2019-05-07 at 11 25 12 AM

rrennick (approved):
@Aljullu This looks great except for those two items I noted. 

The issue with cancelling the existing pending tasks is probably separate from this PR as it uses `WC_Admin_Reports_Sync::regenerate_report_data()` to schedule the tasks and should be addressed in that function.

Pre-approving this with the flag change which will hide the `Status: Finalizing` spinner. 🚀

Aljullu:
Thanks for the review @rrennick!

> I have my AS queue runner disabled & I noticed that if I click start multiple times with or without changing the settings I get two new scheduled tasks each time. Should we cancel existing pending actions for this before scheduling the new ones?
> 
> A use case I'm thinking of is if a user initially chose 365 days then changed their mind to 90 days. The existing 365 day task will run before the 90 day one because it has the older scheduled date.

When there is an import in progress, only the _Stop Import_ button is displayed. If the user doesn't stop the current tasks, they aren't allowed to start a new import. You can see it setting `inProgress: true` in the props.

image

Do you think that's enough? Otherwise, I agree it would make sense that the import endpoint clears ongoing imports before starting a new one.

> This function should set `inProgress` to `false` after a successful api call.

I was thinking about this too... but I'm still unsure if `inProgress` should be a prop defined by the API response or it should be part of the internal component state --or both. I delayed its implementation for a follow-up until we know better how the endpoints describing the import status will work.

In #1851 I divided the issue into three tasks, and I was thinking to delay the `inProgress` implementation for the last one. Does that make sense?

rrennick:
> Do you think that's enough?

> I was thinking to delay the inProgress implementation for the last one. Does that make sense?

Yes, both sounds good.

timmyc:
Looking quite awesome @Aljullu  - since we don't have the API endpoint ready yet to read the status of the import - do you think we should remove the progress bars from the view until we have support to properly display that data?

Aljullu:
> since we don't have the API endpoint ready yet to read the status of the import - do you think we should remove the progress bars from the view until we have support to properly display that data?

Currently the screen is not usable until we have those endpoints because transitions between pages rely on that. It looks like @joshuatf started working on the endpoints in #2189, so I think it makes sense to wait.

In case we have to do a release before this is finished, I think it's better to hide the entire block than leaving it half-baked. How does that sound? @timmyc

timmyc:
> In case we have to do a release before this is finished, I think it's better to hide the entire block than leaving it half-baked. How does that sound?

Let's see how far we can get on the status endpoints and hopefully we don't have to hide anything. I'm feeling like if we don't have the status endpoints done for the next release, I'd still prefer to hide the progress bars, and show the partial import tools vs the import all button.

I still feel it is a better experience for the users than the "import all". /cc @LevinMedia

Aljullu:
> Let's see how far we can get on the status endpoints and hopefully we don't have to hide anything. I'm feeling like if we don't have the status endpoints done for the next release, I'd still prefer to hide the progress bars, and show the partial import tools vs the import all button.

@timmyc Makes sense, I created a follow-up here: #2198. It implements these changes and it should be easy to revert when the import status endpoint is ready.

PR #2179, opened 2019-05-07
Add refund filtering to order reports

### LINKED ISSUE #2003: Refund filtering outside of order statuses
Once #2001 is merged, we will be able to filter the orders report by various "types" of refunds.  

Orders are given a status of `refunded` in core if they are full refunded, but this does not apply to partially refunded orders.  With the new `parent_id` that connects refunds and orders in the `wc_order_stats` table, we will be able to filter by:

* Order statuses with the `refunded` status (already in place)
* Refunds themselves (only the negative amounts and not the original orders)
* Orders with refunds (the parent order, but not necessarily associated refunds)

Do we want to implement either of the latter 2 options?  I'm not sure where this would fit filter-wise since these are not statuses, but a quick and easy way would be to add this under advanced filters:

screen-shot-2019-03-26-at-2 41 54-pm

/cc @josemarques @LevinMedia

joshuatf:
Fixes #2003

Allows filtering to view different types of refunds or filter out refunds.

One odd side effect of doing this is that filtering by "None" for refunds still shows orders with a status of "Refunded" but of course won't show their respective refunds.  I believe this is expected behavior but seems odd at first glance.
Screen Shot 2019-05-07 at 7 41 25 PM
Screen Shot 2019-05-07 at 7 41 20 PM

### Screenshots
Screen Shot 2019-05-07 at 7 37 46 PM

1. Make an order without refunds, one with a partial refund, and one with a full refund (set status to `refunded`).
2. Open the orders report.
3. Add the Refunds filter under the Advanced Filters.
4. Filter by various types of refunds and make sure stats are correct.

### DIFF
--- a/client/analytics/report/orders/config.js
+++ b/client/analytics/report/orders/config.js
@@ -162,6 +162,25 @@ export const advancedFilters = {
 				defaultOption: 'new',
 			},
 		},
+		refunds: {
+			labels: {
+				add: __( 'Refunds', 'woocommerce-admin' ),
+				remove: __( 'Remove refunds filter', 'woocommerce-admin' ),
+				rule: __( 'Select a refund filter match', 'woocommerce-admin' ),
+				title: __( '{{title}}Refunds{{/title}} {{filter /}}', 'woocommerce-admin' ),
+				filter: __( 'Select a refund type', 'woocommerce-admin' ),
+			},
+			input: {
+				component: 'SelectControl',
+				options: [
+					{ value: 'all', label: __( 'All', 'woocommerce-admin' ) },
+					{ value: 'partial', label: __( 'Partially refunded', 'woocommerce-admin' ) },
+					{ value: 'full', label: __( 'Fully refunded', 'woocommerce-admin' ) },
+					{ value: 'none', label: __( 'None', 'woocommerce-admin' ) },
+				],
+				defaultOption: 'new',
+			},
+		},
 	},
 };
 /*eslint-enable max-len*/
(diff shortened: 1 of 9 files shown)


rrennick:
@joshuatf When I have all orders visible I see several fully refunded orders. But with the filter set to full refunds, there are no orders listed.

Screen Shot 2019-05-07 at 2 59 28 PM

Screen Shot 2019-05-07 at 2 59 57 PM

Secondly, I should have #2150 updated tomorrow. Once that is merged in the `All` option could be removed from the Refund drop down if it's not going to be filterable. (Currently, it's just removing the filter for me.)

client/analytics/report/orders/config.js
```diff
@@ -162,6 +162,25 @@ export const advancedFilters = {
...
+			},
+			input: {
+				component: 'SelectControl',
+				options: [
+					{ value: 'all', label: __( 'All', 'woocommerce-admin' ) },
+					{ value: 'partial', label: __( 'Partially refunded', 'woocommerce-admin' ) },
+					{ value: 'full', label: __( 'Fully refunded', 'woocommerce-admin' ) },
+					{ value: 'none', label: __( 'None', 'woocommerce-admin' ) },
+				],
+				defaultOption: 'new',
```
psealock: the `defaultOption` should be the value of one of the above options.

psealock (commented):
I'm having the same issue as Ron. 

When I set a previous order's status to "Refunded", I thought a new order with a negative value would be created, ~which is not happening. Is that not what should happen?~

EDIT: I needed to delete the `wc_admin_version` from my options table.

joshuatf:
Thanks for the reviews @rrennick and @psealock!

@rrennick I think you're experiencing the "odd side effect" I mentioned above, but this is currently expected behavior.  Filtering by "all" should not included the refunded order itself, but the negative refund values.  I do agree that this seems counterintuitive at first glance, but makes sense when you're expecting to see the negative values. @LevinMedia not sure if there's any other way to clear up the language further to make this clear.

@psealock Updated the `defaultOption`.  Negative (refund) rows are now showing for you, right?

psealock (approved):
LGTM! I'm getting the behaviour I expect and unit tests look comprehensive. Nice work @joshuatf

PR #2184, opened 2019-05-08
Travis: Exit scripts with error when one occurs

psealock:
If Travis has failing scripts like `npm run lint`, the builds still succeeds because the bash script doesn't persist the error.

This PR persists errors so that builds fail when they should.

### Testing Intructions

1. See the build failed on this PR's first commit. The build should fail because there is a linting error.
2. See the final build on this PR succeed because the linting error has been addressed.

### DIFF
--- a/bin/js_lint_test.sh
+++ b/bin/js_lint_test.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -o errexit
+
 if [[ ${RUN_JS} == 1 ]]; then
 	npm run lint
 	npm run build
--- a/client/dashboard/store-performance/index.js
+++ b/client/dashboard/store-performance/index.js
@@ -199,7 +199,7 @@ export default compose(
 
 		if ( statKeys.length === 0 ) {
 			return {
-				hiddenIndicators,
+				hiddenIndicators: [],
 				userIndicators,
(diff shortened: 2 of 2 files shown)


bin/js_lint_test.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -o errexit
```
Aljullu: :clap:

client/dashboard/store-performance/index.js
```diff
@@ -199,7 +199,7 @@ export default compose(
 
 		if ( statKeys.length === 0 ) {
 			return {
-				hiddenIndicators,
+				hiddenIndicators: [],
```
Aljullu: I think this should be renamed to `hiddenBlocks`, I guess I messed it up when merging master into #2148. In that PR I unified `hiddenCharts`, `hiddenIndicators`, etc. to use the same prop name (`hiddenBlocks`) so props would be consistent between sections.
```suggestion
				hiddenBlocks,
```

Aljullu (approved):
Thanks for fixing this @psealock and good catch @rrennick! I just added a small comment, but pre-approving so we can get this merged soon.

PR #2189, opened 2019-05-08
Add import status endpoint and progress options

### LINKED ISSUE #1850: Add historical data import endpoint

joshuatf:
Fixes #1850

Adds an endpoint to query the current import progress and oldest date of import.  The date will be set to `-1` if `days` are not specific initially.

Also adds an endpoint to query potential totals prior to import.

### Thoughts/notes
* Opted to use options here instead of attempting to query pending/completed batches.  The latter seems like it would get messy quickly and is prone to inaccurate results with completed results.
* Orders total is not calculated until after the customers batch is finished since it's a dependent action.  If we want to show orders total while customers are being processed, we will probably need to create both customer and order batches initially so that we can calculate totals and set the orders batch to pending until the customers batch is complete.
* Imports should still be limited to one import at a time.  Will outline a separate issue to tackle this.

### Screenshots
Screen Shot 2019-05-08 at 3 56 39 PM

1.  Start an import by posting to the `/wp-json/wc/v4/reports/import` endpoint.
2.  Send a get request to `/wp-json/wc/v4/reports/import/status`.
3.  Check that totals and progress count are reflected in the endpoint.
4.  Send a get request to `/wp-json/wc/v4/reports/import/totals` with `day` and `skip_existing` params.
5.  Make sure that total numbers returned are accurate.

### DIFF
--- a/includes/api/class-wc-admin-rest-reports-import-controller.php
+++ b/includes/api/class-wc-admin-rest-reports-import-controller.php
@@ -270,10 +270,11 @@ class WC_Admin_REST_Reports_Import_Controller extends WC_Admin_REST_Reports_Cont
 	 */
 	public function get_import_status( $request ) {
 		$result = array(
+			'is_importing'    => WC_Admin_Reports_Sync::is_importing(),
 			'customers_total' => get_option( 'wc_admin_import_customers_total', 0 ),
 			'customers_count' => get_option( 'wc_admin_import_customers_count', 0 ),
 			'orders_total'    => get_option( 'wc_admin_import_orders_total', 0 ),
-			'orders_count'    => get_option( 'wc_admin_import_orders_total', 0 ),
+			'orders_count'    => get_option( 'wc_admin_import_orders_count', 0 ),
 			'imported_from'   => get_option( 'wc_admin_imported_from_date', false ),
 		);
 
--- a/includes/api/class-wc-admin-rest-reports-import-controller.php
+++ b/includes/api/class-wc-admin-rest-reports-import-controller.php
@@ -71,6 +71,18 @@ class WC_Admin_REST_Reports_Import_Controller extends WC_Admin_REST_Reports_Cont
 				'schema' => array( $this, 'get_import_public_schema' ),
 			)
 		);
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base . '/status',
+			array(
+				array(
+					'methods'             => WP_REST_Server::READABLE,
+					'callback'            => array( $this, 'get_import_status' ),
+					'permission_callback' => array( $this, 'import_permissions_check' ),
+				),
+				'schema' => array( $this, 'get_import_public_schema' ),
+			)
+		);
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


includes/api/class-wc-admin-rest-reports-import-controller.php
```diff
@@ -249,4 +261,27 @@ public function delete_imported_items( $request ) {
...
 	}
+
+	/**
+	 * Get the status of the current import.
+	 *
+	 * @param  WP_REST_Request $request Request data.
+	 * @return WP_Error|WP_REST_Response
+	 */
+	public function get_import_status( $request ) {
+		$delete = WC_Admin_Reports_Sync::delete_report_data();
```
Aljullu: I don't understand this line. Is it a leftover from another function?

Aljullu (commented):
Thanks for working on this @joshuatf! Just a couple of comments/questions:

> Orders total is not calculated until after the customers batch is finished since it's a dependent action. If we want to show orders total while customers are being processed, we will probably need to create both customer and order batches initially so that we can calculate totals and set the orders batch to pending until the customers batch is complete.

According to the designs it looks like we should display the totals even before starting the import:

image

Not sure if that's possible or we should just hide the progress bars when the import hasn't started yet.

joshuatf:
Made some changes to this PR:

* Added `is_importing` to the status route.
* Added a separate totals endpoint (testing instructions above) to query totals prior to importing.
* Set totals for current import prior to beginning import.
* Added tests for totals and status endpoints.

@Aljullu this is ready for another look if you have time.

includes/notes/class-wc-admin-notes.php
```diff
@@ -131,7 +136,21 @@ public static function schedule_unsnooze_notes() {
...
+
+	/**
+	 * Clears all queued actions.
+	 */
+	public static function clear_queued_actions() {
+		$store = ActionScheduler::store();
+
+		if ( is_a( $store, 'WC_Admin_ActionScheduler_WPPostStore' ) ) {
+			// If we're using our data store, call our bespoke deletion method.
+			$store->clear_pending_wcadmin_actions();
```
Aljullu: This method [seems to clear import actions](https://github.com/woocommerce/woocommerce-admin/blob/master/includes/class-wc-admin-actionscheduler-wppoststore.php#L42), right? Shouldn't we create a different method which only cleared note actions?

Aljullu (commented):
Thanks for the last changes @joshuatf!

This is testing great and code looks good to me, I just added a small question in the code.

includes/notes/class-wc-admin-notes.php
joshuatf: Ya, I think you're right and we should actually extract the actions out of that method to make this reusable.  Updated in https://github.com/woocommerce/woocommerce-admin/pull/2189/commits/13753b7e47869c3b55ead67b34dad87757d99945

joshuatf:
Thanks again for the review @Aljullu!  Added an update to separate out action clearing for notes and imports.

Aljullu (approved):
Thanks for the changes @joshuatf! LGTM

includes/class-wc-admin-actionscheduler-wppoststore.php
```diff
@@ -36,23 +36,15 @@ protected function create_post_array( ActionScheduler_Action $action, DateTime $
...
-		$action_types = array(
-			WC_Admin_Reports_Sync::QUEUE_BATCH_ACTION,
-			WC_Admin_Reports_Sync::QUEUE_DEPEDENT_ACTION,
-			WC_Admin_Reports_Sync::CUSTOMERS_IMPORT_BATCH_ACTION,
-			WC_Admin_Reports_Sync::ORDERS_IMPORT_BATCH_ACTION,
-			WC_Admin_Reports_Sync::ORDERS_IMPORT_BATCH_INIT,
-			WC_Admin_Reports_Sync::SINGLE_ORDER_IMPORT_ACTION,
-		);
-
 		foreach ( $action_types as $action_type ) {
```
rrennick: Since this is a public function `$action_types` should probably be cast to an array.

PR #2190, opened 2019-05-08
fix customer name logic on populating customer lookup data

### LINKED ISSUE #2180: Error processing historical data actions
**Describe the bug**

The order of operation in the logic for populating the customer name allows the `get_*_last_name()` functions to be called when `$order` is false in `WC_Admin_Reports_Customers_Data_Store::get_customer_name()`. Here is the stack trace:

```shell
Fatal error: Uncaught Error: Call to a member function get_billing_last_name() on boolean in /Users/ronrennick/Sites/ron/wp-content/plugins/woocommerce-admin/includes/data-stores/class-wc-admin-reports-customers-data-store.php:536
Stack trace:
#0 /Users/ronrennick/Sites/ron/wp-content/plugins/woocommerce-admin/includes/data-stores/class-wc-admin-reports-customers-data-store.php(631): WC_Admin_Reports_Customers_Data_Store::get_customer_name('3', false)
#1 /Users/ronrennick/Sites/ron/wp-content/plugins/woocommerce-admin/includes/class-wc-admin-reports-sync.php(557): WC_Admin_Reports_Customers_Data_Store::update_registered_customer('3')
```

**To Reproduce**

*I'm working with a restored DB from March so I'm not 100% this is how to reproduce*
1. Create an order
2. Process the historical data
3. Manually delete the order from the database
4. Use #2177 to re-import the data
5. The stack trace is added to the log of the scheduled action

I'll submit a PR for this later today.

rrennick:
Fixes #2180

This PR fixes the logic for populating the customer first/last name in the customer lookup table. In PHP conditional evaluation `if ( $a && $b || $c ) {` when `$a` is false then `$c` is evaluated. The new logic only calls the last name lookup when there is a valid user ID/order.

1. Create an order
2. Process the historical data
3. Manually delete the order from the database
4. Use #2177 to re-import the data
5. The scheduled action should complete

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-customers-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-customers-data-store.php
@@ -525,25 +525,24 @@ class WC_Admin_Reports_Customers_Data_Store extends WC_Admin_Reports_Data_Store 
 
 		if (
 			$user_id &&
-			get_user_meta( $user_id, 'first_name', true ) ||
-			get_user_meta( $user_id, 'last_name', true )
+			(
+				get_user_meta( $user_id, 'first_name', true ) ||
+				get_user_meta( $user_id, 'last_name', true )
+			)
 		) {
 			$first_name = get_user_meta( $user_id, 'first_name', true );
 			$last_name  = get_user_meta( $user_id, 'last_name', true );
-		} elseif (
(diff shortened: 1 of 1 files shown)


rrennick:
Thanks @psealock

PR #2193, opened 2019-05-08
Add tracks to import jobs

### LINKED ISSUE #2173: Tracks: Track Start of Import Jobs
In order to gauge the number of import jobs that are succeeding/completing/failing I'm proposing we implement tracks for the following events:

- When an import job ( singular job, not batch job ) begins. Event name: `import_job_start`
- When an import job ( again not batch jobs ) completes Event name: `import_job_complete`

For event props, we should pass along the `type` of job - whether it be a customer import or an order import.

We will lean on the presence of `WC_Tracks` which was introduced in 3.6 to record the events via PHP, details can be seen in the [record_event](https://github.com/woocommerce/woocommerce/blob/master/includes/tracks/class-wc-tracks.php#L75) method.

Aljullu:
Fixes #2173.

Add tracks to import and delete imported data jobs.

- Go to _Settings_ and press _Rebuild reports_.
- Make a call to `/wc/v4/reports/import/delete` (or merge #2198 and press _Delete Previously Imported Data_).
- Wait some minutes and go to Tracks Live View. There, verify the events appear, and they have the `type` in the properties.
  - https://mc.a8c.com/tracks/live/?eventname=wcadmin_import%25&user=&useragent=
  - https://mc.a8c.com/tracks/live/?eventname=wcadmin_delete%25&user=&useragent=


mikejolley:
@Aljullu I enabled tracking and waited a while but I'm not seeing these events in tracks :( Could I be missing a step?

jeffstieler:
I'm getting a fatal trying to test: `PHP Fatal error:  Uncaught Error: Class 'WC_Tracks' not found in /srv/www/wordpress-default/public_html/wp-content/plugins/wc-admin/includes/class-wc-admin-reports-sync.php:542`

timmyc:
I did a rebase of this branch to see if it magically might work, but I'm getting the same error myself:

```
2019-05-14 18:19:05 +0000
unexpected shutdown: PHP Fatal error Uncaught Error: Class 'WC_Tracks' not found in /srv/www/woo-one/public_html/wp-content/plugins/woocommerce-admin/includes/class-wc-admin-reports-sync.php:712 
```

mikejolley:
@jeffstieler @timmyc This can be re-reviewed. The issue was core doesn't load tracks for REST API requests.

jeffstieler:
I'm seeing the events fire now (using a `http_request_args` filter to log outbound requests), but it seems really noisy. Every batch gets a `_start` and `_complete` job, with no indicator of batch number or anything identifying for that particular start/complete pairing.

From #2173:

> * When an import job ( singular job, not batch job ) begins. Event name: `import_job_start`
> * When an import job ( again not batch jobs ) completes Event name: `import_job_complete`

Can you elaborate on the "not batch job" here @timmyc - I'm thinking that probably addresses my "noise" concerns.

timmyc:
> Can you elaborate on the "not batch job" here @timmyc - I'm thinking that probably addresses my "noise" concerns.

Indeed I figured we shouldn't track the batch jobs because, noise, but also because the batch size can be adjusted per-install so it doesn't really yield us much info ( unless we record a batch size as a prop ) but still the end goal here being better visibility into the number of jobs ( not batch jobs ) started, and completed to get a sense of how many failures we are hitting along the way.

Aljullu:
@mikejolley thanks for fixing the fatal. :slightly_smiling_face: 

@jeffstieler @timmyc I'm not sure to understand what the final decision was. When importing orders, for example, these jobs are called:

```
wc-admin_orders_lookup_import_batch_init
wc-admin_import_orders_batch
wc-admin_import_orders_batch
wc-admin_import_orders_batch
...
```

In the way this PR currently works, the events are called at the beginning and end of each `wc-admin_import_orders_batch` job. From what you are saying, I understand the start event should be called when the `wc-admin_orders_lookup_import_batch_init` starts and the complete event when the last `wc-admin_import_orders_batch` finishes, is that right?

jeffstieler:
@Aljullu sorry for the delay in response here.

>  From what you are saying, I understand the start event should be called when the `wc-admin_orders_lookup_import_batch_init` starts and the complete event when the last `wc-admin_import_orders_batch` finishes, is that right?

That would definitely cut down on the "noise" - but I think a decent middle ground solution would be to add batch number and batch size as event props to the `wc-admin_import_orders_batch` events so the events aren't all identical.

Aljullu:
> I think a decent middle ground solution would be to add batch number and batch size as event props to the wc-admin_import_orders_batch events so the events aren't all identical.

@jeffstieler I added a `batch_number` and `batch_size` properties to start & complete events in c83ff6aba0a5cf00b010fc6816d144abb5e1e774. I also added an `imported_count` property to complete events since I thought it might be useful too. Can you take another look?

timmyc:
@jeffstieler could you give this another set of 👀 today?

PR #2194, opened 2019-05-09
Fix/dashboard issues

psealock:
Fixes the following problems with the Dashboard:

1. Moving a section didn't always work.
2. Extensions couldn't access `<SectionsControls>` component.

Fix 1: An index was passed to each section based on its position in the `visibleSections` array, which didn't translate to the `sections` array. This caused a bad index to be passed to `onMove`. So this PR used the index of `sections` instead.

Fix 2: Pass down the `SectionControls` component as a prop in `section.js`. This allows extensions to have access. The other alternative would have been to publish section-controls.js as a woocommerce/component but it doesn't seem useful enough. Consequently, I removed the `/components` folder because it doesn't make sense.

Fix 1: Remove any section besides the last one. Try moving the last one, it should move as directed.

Fix 2: Make sure the page loads and section controls work.


client/dashboard/README.md
```diff
@@ -46,3 +46,4 @@ The component will get the following props:
 - `path` (string): The exact path for this view.
 - `query` (object): The query string for the current view, can be used to read current preferences for time periods or chart interval/type.
 - `title` (string): Title of the section according to the default settings or the user preferences if they had made any modification.
+- `controls` (react component): Controls to move a section up/down or remove it from view to be rendered inside the EllipsisMenu.
```
joshuatf: 👍 Seems convenient and can't think of any reason we'd need it as an exported package component for the time being unless anyone else can think of a good reason.

client/dashboard/section-controls.js
```diff
@@ -36,10 +35,6 @@ class SectionControls extends Component {
 	render() {
 		const { onRemove, isFirst, isLast } = this.props;
 
-		if ( ! window.wcAdminFeatures[ 'analytics-dashboard/customizable' ] ) {
```
joshuatf: If we remove this here, could this possibly let a third party add this component to the dashboard even if the `analytics-dashboard/customizable` flag is disabled?

client/dashboard/customizable.js
```diff
@@ -163,29 +163,34 @@ class CustomizableDashboard extends Component {
...
+								hiddenBlocks={ section.hiddenBlocks }
+								key={ section.key }
+								onChangeHiddenBlocks={ this.onChangeHiddenBlocks( section.key ) }
+								onTitleUpdate={ this.onSectionTitleUpdate( section.key ) }
+								path={ path }
+								query={ query }
+								title={ section.title }
+								onMove={ partial( this.onMove, index ) }
+								onRemove={ this.toggleVisibility( section.key ) }
+								isFirst={ section.key === visibleSectionKeys[ 0 ] }
```
joshuatf: Somewhat unrelated, but might be a good follow-up here: I wonder if passing `visibleSectionKeys` down instead of `isFirst` and `isLast` props would prove useful to plugins that plan to use this section more dynamically.  This would also have the added benefit of reducing the number of props we need to pass down.

client/dashboard/section-controls.js
psealock: Indeed it would, but the filter to add dashboard items is in `customizable.js` which is itself behind the flag.

client/dashboard/customizable.js
psealock: Yeah good idea, that would give more control to extensions. But then we'd require them to apply the logic and render Controls. I'll give this a bit more thought.

```js
<Controls
	onToggle={ onToggle }
	onMove={ onMove }
	onRemove={ onRemove }
	isFirst={ 'leaderboards' === visibleSectionKeys[ 0 ] }
        isLast={ 'leaderboards' === visibleSectionKeys[ visibleSectionKeys.length - 1 ] }
/>
```

psealock:
>This doesn't seem to be fixed for me.

😝 Whoops! Nice find @Aljullu! Right, if you skip over a hidden section, it doesn't change anything on the screen.

https://github.com/woocommerce/woocommerce-admin/pull/2194/commits/2aff58fddad7d8f066b40b8804a457286d77786b Adds a check to see if the skipped section is visible before inserting. If not, it tries the next.

client/dashboard/customizable.js
joshuatf: Ya, that's true.  Could we pass `visibleSectionKeys` and the key down to the control and handle logic there?  That way they're only passing props and no logic changes are necessary inside extensions.

Aljullu:
> stuck_out_tongue_closed_eyes Whoops! Nice find @Aljullu! Right, if you skip over a hidden section, it doesn't change anything on the screen.

Confirming it's solved now, thank you @psealock!

client/dashboard/customizable.js
psealock: Yeah, this is the ideal scenario. I'll merge this in and give some thought on how to make this happen.

psealock:
>Confirming it's solved now, thank you @psealock!

Thanks @Aljullu! Is this approved and ready to go?

Issue #2119, opened 2019-04-26
WordPress Upgrade Nag Results in Oddities


timmyc:
**Describe the bug**
When WordPress core injects the `div.update-nag` at the top of wp-admin pages, the visual layout of pages which include the wc-admin navigation bar go a bit awry:

__Wide viewport with update-nag__
Note the extra padding above store alerts / under the nav bar
![image](https://user-images.githubusercontent.com/22080/56832762-b4eab180-6821-11e9-9b36-f452d638062f.png)

__Wide viewport without nag__
![image](https://user-images.githubusercontent.com/22080/56832797-c7fd8180-6821-11e9-8de6-f06e7c793822.png)
 
__Narrow viewport with update-nag__
![narrow-nag](https://user-images.githubusercontent.com/22080/56833022-3b9f8e80-6822-11e9-8b10-86465ed41f23.png)

__Narrow without nag__
![narrow-no-nag](https://user-images.githubusercontent.com/22080/56833041-4a864100-6822-11e9-894f-df69aacd5606.png)

And here is what the div.update-nag looks like since it is obscured in the screens above:

![update-nag](https://user-images.githubusercontent.com/22080/56833142-7e616680-6822-11e9-8601-6c5adca920b9.png)

**To Reproduce**

1. Setup a site that is not on the latest version of WordPress
2. install wc-admin

Mr-AjayM:
Hi @timmyc 

I can regenerate this issue. Do we need to remove this from the pages or just set display none?

Let me know your inputs for expected behaviour for this.

Thanks

timmyc:
Hello @Mr-AjayM 👋 

I think a valid fix would be to use `display: none` on the pages where the new WooCommerce Admin navigation bar is in use.

PR #2201, opened 2019-05-10
Fix WordPress upgrade nag results in oddities#2119

Mr-AjayM:
Fix WordPress upgrade nag results in oddities #2119

Fix design issue that occurs due to the WordPress upgrade nag when using the older version of WordPress.

### Accessibility

- [X] I've tested using only a keyboard (no mouse)
- [X] All text has [at least a 4.5 color contrast with its background](https://webaim.org/resources/contrastchecker/)

### Screenshots
Screenshot_1
Screenshot_2

WordPress upgrade nag will not display in the WooCommerce Pages if even using the older version of WordPress.

Thanks,


Mr-AjayM:
Hello @joshuatf 

Thanks for sharing your inputs for this!

I have updated both the changes in the separate PR(https://github.com/woocommerce/woocommerce-admin/pull/2213). Kindly review it and let me know if we need to change anything for this.

Thanks,
Aj

PR #2209, opened 2019-05-10
Refactor page handler API

### LINKED ISSUE #1431: Activity Panel: Implementation of useful Inbox notices
Right now, only a handful of notices are implemented in core (store notice, best sales). To properly implement the Inbox feature, we need more useful and timely information.

This issue can serve as a master issue for implementing a MVLP level amount of useful notices for core. This issue can be closed when that criteria is met.

A document for these was started. See p6q8Tx-PA-p2 #comment-2827. This is not a final list.

> We need to come back to it and make it definitive to move forward.

### LINKED ISSUE #1432: Activity Panel: Refactor page handler and wp-admin embed code
See p90Yrv-U3-p2 for more details.

A summary of that thread is that we've learned a lot since the original embed code was written. Specifically things like the page controller,  screen id methods `wc_admin_get_current_screen_id`, `wc_admin_get_embed_enabled_screen_ids` and related code can be refactored.

These will be the main API extensions will use to embed the activity panel and breadcrumb navigation, and in the future might be used for other things like the App Experience, so it's important we spend some time thinking about the interface/documentation.

- [x] Audit core and rewrite wc_admin_get_current_screen_id
- [x] Add a page controller
- [x] Replace breadcrumb handling based on page controller meta 
- [ ] Write documentation showing how this can be used by 3rd party devs

jeffstieler:
Fixes #1432.

This PR introduces a new API for registering new React-powered pages and connecting PHP-powered pages to WooCommerce Admin.

It also introduces developer documentation for using the new API. See: `docs/page-controller.md`.

- Open WooCommerce Admin
- Verify all WP admin menu items are present and link to the proper pages
- For each page: verify the breadcrumb titles and links
- Read the new documentation page


includes/class-wc-admin-loader.php
```diff
@@ -145,6 +121,9 @@ private static function get_path( $file ) {
 	 * Class loader for enabled WooCommerce Admin features/sections.
 	 */
 	public static function load_features() {
+		require_once WC_ADMIN_ABSPATH . 'includes/page-controller/class-wc-admin-page-controller.php';
```
justinshreve: Good call loading these here. I was worried about possible function name collision between the plugin here, and this shipping in core -- but loading here alongside the other feature classes should be late enough.

includes/class-wc-admin-loader.php
```diff
@@ -6,38 +6,14 @@
...
-}
-
 /**
  * WC_Admin_Loader Class.
  */
 class WC_Admin_Loader {
+	/**
+	 * App entry point.
+	 */
+	const APP_ENTRY_POINT = 'wc-admin';
```
justinshreve: 👍

includes/page-controller/class-wc-admin-page-controller.php
```diff
@@ -0,0 +1,398 @@
...
+	 *   @type string       icon         Icon. Dashicons helper class, base64-encoded SVG, or 'none'.
+	 *   @type int          position     Menu item position.
+	 * }
+	 */
+	public function connect_page( $options ) {
+		if ( ! is_array( $options['title'] ) ) {
+			$options['title'] = array( $options['title'] );
+		}
+
+		// TODO: check for null ID, or collision.
```
justinshreve: Minor: We've been using `// @todo` to mark these, so that we can find them all in one format.

includes/page-controller/class-wc-admin-page-controller.php
```diff
@@ -0,0 +1,398 @@
...
+		$this->pages[ $options['id'] ] = $options;
```
justinshreve: I'm wondering if it might be worth running `$options` through a filter before adding it here?

The use case I'm thinking of here is the eCommerce Plan, where menu/submenu information is passed in a similar manner: https://github.com/Automattic/wc-calypso-bridge/blob/master/includes/connect/woocommerce-bookings.php In the future, those connect functions could be re-written to append extra info to these existing arrays.

includes/class-wc-admin-loader.php
```diff
@@ -163,28 +142,586 @@ public static function load_features() {
...
+		wc_admin_register_page(
+			array(
+				'id'     => 'woocommerce-dashboard', // Expected to be overridden if dashboard is enabled.
+				'parent' => 'woocommerce',
+				'title'  => null,
+				'path'   => self::APP_ENTRY_POINT,
+			)
+		);
+
+		// TODO: move this somewhere else?
```
justinshreve: Yeah, I'm thinking loading these in from another file probably makes sense, just because of the sheer number of lines dedicated to connecting.

includes/class-wc-admin-loader.php
```diff
@@ -163,28 +142,586 @@ public static function load_features() {
...
+						'page' => 'wc-settings',
+						'tab'  => 'email',
+					),
+					$admin_page_base
+				),
+			)
+		);
+
+		// WooCommerce > Settings > Emails > Edit email (all email editing).
+		$emails = array(
```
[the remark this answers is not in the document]
jeffstieler: Great idea! I'll use this as inspiration: https://github.com/woocommerce/woocommerce/blob/178c7a7a9cc9c9e760010a8f01e7353e5c6fe611/includes/admin/settings/class-wc-settings-emails.php#L236

mikejolley:
@justinshreve Might be out of scope for this PR but we should try to avoid adding hooks to `__construct` and instead use an `init` method so if classes are constructed multiple times, we don't hook in multiple times. Maybe good to look at once this is merged.

jeffstieler:
@mikejolley did you mean to tag me for that comment? I'll take a look at the constructor hooks.

jeffstieler:
@mikejolley I assume you're referring to the hooks in the constructors of the feature classes. These are meant to be singletons, so are you OK with just changing the constructors to `private` access?

mikejolley:
@jeffstieler I think it can be handled in a separate PR so let's not further complicate this one IMO

rrennick:
> so are you OK with just changing the constructors to private access?

Constructors can only be public as they are always called from outside the class.

jeffstieler:
> Constructors can only be public as they are always called from outside the class.

We're getting into the weeds here, but they _can_ be private to enforce use of singleton pattern "get instance" methods (and are called internally):

```php
<?php

class Test {
    private static $instance;

    public $test;

    private function __construct() {
        $this->test = 'Hello';
    }

    public static function get_instance() {
        if ( ! is_a( self::$instance, __CLASS__ ) ) {
            self::$instance = new self;
        }

        return self::$instance;
    }
}

$test = Test::get_instance();

var_dump( $test->test );

$test2 = new Test();
```

```sh
string(5) "Hello"

Fatal error: Uncaught Error: Call to private Test::__construct() from invalid context in test.php:25
Stack trace:
#0 {main}
  thrown in /test.php on line 25
```

rrennick:
@jeffstieler Thanks, I hadn't seen that before.

docs/page-controller.md
```diff
@@ -0,0 +1,134 @@
...
+
+Add our new report using the appropriate filter:
+
+```javascript
+import { addFilter } from '@wordpress/hooks';
+
+addFilter( 'woocommerce_admin_reports_list', 'my-namespace', ( reports ) => {
+	reports.push( {
+		report: 'example',
+		title: 'Example',
```
timmyc: Minor but shouldn't we have this title translated in the example?

includes/features/analytics-dashboard/class-wc-admin-analytics-dashboard.php
```diff
@@ -64,6 +72,50 @@ public function add_user_data_fields( $user_data_fields ) {
...
 	}
+
+	/**
+	 * Registers dashboard page.
+	 */
+	public function register_page() {
+		wc_admin_register_page(
+			array(
+				'id'     => 'woocommerce-dashboard',
+				'title'  => 'Dashboard',
```
timmyc: Needs i18n correct?

jeffstieler: Fixed in a1229e5.

docs/page-controller.md
jeffstieler: Fixed in a1229e5.

PR #2211, opened 2019-05-10
update default capability for analytics report to view_woocommerce_reports

### LINKED ISSUE #1968: Use `view_woocommerce_reports` check for Menu Item Registraion
By report, please allow us to select the category of user who can see this report. For example, only admins can see P&L reports but an inventory report can be seen by admins and store managers. The permission should be granted per report.

--- Added 2019-04-29 by @timmyc ---

Currently, as @nikonratm [points out below](https://github.com/woocommerce/woocommerce-admin/issues/1968#issuecomment-484265081) the display of the menu items for Analytics/Dashboard is currently scoped to users whom have `manage_options` capabilities.

While on the REST API side of things, we use the [following logic](https://github.com/woocommerce/woocommerce-admin/blob/master/includes/api/class-wc-admin-rest-reports-controller.php#L60) to determine if users can access the reporting endpoints:

`wc_rest_check_manager_permissions( 'reports', 'read' )`

And the legacy reports [also use this capability to register menu items](https://github.com/woocommerce/woocommerce/blob/b95bf94092a2c6fd1de863e4aae92debfcb5fd42/includes/admin/class-wc-admin-menus.php#L71).

Therefore it seems like we should use the same capability for the new Analytics section as well.

rrennick:
Fixes #1968

This PR updates the Analytics menu and submenu items capability permission to `view_woocommerce_reports`. Submenu item capability permission can be changed using the `woocommerce_admin_report_menu_items` filter.

- Log in as a shop manager
- View various analytics screens
- Add a filter to `woocommerce_admin_report_menu_items` and add a `capability` => `manage_options` element to one of the menu items
- refresh the screen with to see that screen not included in the menu

### DIFF
--- a/includes/class-wc-admin-loader.php
+++ b/includes/class-wc-admin-loader.php
@@ -142,12 +142,14 @@ class WC_Admin_Loader {
 	 * @todo The entry point for the embed needs moved to this class as well.
 	 */
 	public static function register_page_handler() {
+		$analytics_cap = apply_filters( 'woocommerce_admin_analytics_menu_capability', 'view_woocommerce_reports' );
 		wc_admin_register_page(
 			array(
-				'id'     => 'woocommerce-dashboard', // Expected to be overridden if dashboard is enabled.
-				'parent' => 'woocommerce',
-				'title'  => null,
-				'path'   => self::APP_ENTRY_POINT,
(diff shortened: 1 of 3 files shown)


timmyc:
Thanks for grabbing this one @rrennick

rrennick:
@mikejolley @timmyc I updated this for the new page controller. Could give it a quick second check?

PR #2217, opened 2019-05-13
Add onboarding profile REST endpoints

### LINKED ISSUE #1883: Onboarding - Profile Wizard: REST API for profile data
This task involves creating REST API methods for dealing with profile data from the various steps, including saving it into the database. I think the storage for this could probably be an option like `wc_onboarding_data` with serialized/stored data. We'll be collecting industry, product types, and various questions on the store details page.

When implementing this issue, extensibility (#1891) should be kept in mind.

joshuatf:
Fixes #1883

Adds endpoints for getting profile data and setting it.

**Profile params**
`account_type`
`store_address_1`
`store_address_2`
`store_zip`
`store_city`
`store_state`
`store_country`
`industry`
`product_types`
`product_count`
`selling_venues`
`other_platform`
`theme`
`purchase`

### Screenshots
Screen Shot 2019-05-13 at 2 33 20 PM

1.  Make a GET request to `/wp-json/wc/v4/onboarding/profile`.
2.  All initial values should be present but set to null.
3.  Post to `/wp-json/wc/v4/onboarding/profile` with any of the profile params.
4.  Make another GET request to `/wp-json/wc/v4/onboarding/profile`.
5.  Note the updated fields.

### DIFF
--- a/includes/api/class-wc-admin-rest-onboarding-levels-controller.php
+++ b/includes/api/class-wc-admin-rest-onboarding-levels-controller.php
@@ -21,7 +21,7 @@ class WC_Admin_REST_Onboarding_Levels_Controller extends WC_REST_Data_Controller
 	 *
 	 * @var string
 	 */
-	protected $namespace = 'wc-internal/v1';
+	protected $namespace = 'wc-admin/v1';
 
 	/**
 	 * Route base.
--- a/includes/api/class-wc-admin-rest-onboarding-levels-controller.php
+++ b/includes/api/class-wc-admin-rest-onboarding-levels-controller.php
@@ -21,7 +21,7 @@ class WC_Admin_REST_Onboarding_Levels_Controller extends WC_REST_Data_Controller
 	 *
 	 * @var string
 	 */
-	protected $namespace = 'wc/v4';
+	protected $namespace = 'wc-internal/v1';
 
 	/**
 	 * Route base.
--- a/includes/api/class-wc-admin-rest-onboarding-profile-controller.php
+++ b/includes/api/class-wc-admin-rest-onboarding-profile-controller.php
@@ -21,7 +21,7 @@ class WC_Admin_REST_Onboarding_Profile_Controller extends WC_REST_Data_Controlle
 	 *
 	 * @var string
 	 */
-	protected $namespace = 'wc-internal/v1';
+	protected $namespace = 'wc-admin/v1';
 
 	/**
 	 * Route base.
--- a/includes/api/class-wc-admin-rest-onboarding-profile-controller.php
+++ b/includes/api/class-wc-admin-rest-onboarding-profile-controller.php
@@ -199,7 +199,14 @@ class WC_Admin_REST_Onboarding_Profile_Controller extends WC_REST_Data_Controlle
 	 */
 	public static function get_profile_properties() {
 		$properties = array(
-			'account_type'   => array(
+			'skipped'         => array(
+				'type'              => 'bool',
+				'description'       => __( 'Whether or not the profile was skipped.', 'woocommerce-admin' ),
+				'context'           => array( 'view' ),
+				'readonly'          => true,
+				'validate_callback' => 'rest_validate_request_arg',
+			),
+			'account_type'    => array(
 				'type'              => 'string',
 				'description'       => __( 'Account type used for Jetpack.', 'woocommerce-admin' ),
 				'context'           => array( 'view' ),
@@ -211,7 +218,7 @@ class WC_Admin_REST_Onboarding_Profile_Controller extends WC_REST_Data_Controlle
 					'google',
 				),
 			),
-			'industry'       => array(
+			'industry'        => array(
 				'type'              => 'array',
 				'description'       => __( 'Industry.', 'woocommerce-admin' ),
 				'context'           => array( 'view' ),
@@ -230,7 +237,7 @@ class WC_Admin_REST_Onboarding_Profile_Controller extends WC_REST_Data_Controlle
 					'type' => 'string',
 				),
 			),
-			'product_types'  => array(
+			'product_types'   => array(
 				'type'              => 'array',
 				'description'       => __( 'Types of products sold.', 'woocommerce-admin' ),
 				'context'           => array( 'view' ),
@@ -250,7 +257,7 @@ class WC_Admin_REST_Onboarding_Profile_Controller extends WC_REST_Data_Controlle
 					'type' => 'string',
 				),
 			),
-			'product_count'  => array(
+			'product_count'   => array(
 				'type'              => 'string',
 				'description'       => __( 'Number of products to be added.', 'woocommerce-admin' ),
 				'context'           => array( 'view' ),
@@ -263,7 +270,7 @@ class WC_Admin_REST_Onboarding_Profile_Controller extends WC_REST_Data_Controlle
 					'1000+',
 				),
 			),
-			'selling_venues' => array(
+			'selling_venues'  => array(
 				'type'              => 'string',
 				'description'       => __( 'Other places the store is selling products.', 'woocommerce-admin' ),
 				'context'           => array( 'view' ),
@@ -276,7 +283,7 @@ class WC_Admin_REST_Onboarding_Profile_Controller extends WC_REST_Data_Controlle
 					'brick-mortar-other',
 				),
 			),
-			'other_platform' => array(
+			'other_platform'  => array(
 				'type'              => 'string',
 				'description'       => __( 'Name of other platform used to sell.', 'woocommerce-admin' ),
 				'context'           => array( 'view' ),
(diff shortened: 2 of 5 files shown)


includes/api/class-wc-admin-rest-onboarding-profile-controller.php
```diff
@@ -0,0 +1,330 @@
...
+ * @package WooCommerce Admin/API
+ * @extends WC_REST_Data_Controller
+ */
+class WC_Admin_REST_Onboarding_Profile_Controller extends WC_REST_Data_Controller {
+	/**
+	 * Endpoint namespace.
+	 *
+	 * @var string
+	 */
+	protected $namespace = 'wc/v4';
```
justinshreve: There was some discussion in slack. Could we update this to `wc-internal/v1`?  p1557756996005400-slack-wc-onboarding.

includes/api/class-wc-admin-rest-onboarding-profile-controller.php
```diff
@@ -0,0 +1,330 @@
...
+	/**
+	 * Get onboarding profile properties.
+	 *
+	 * @return array
+	 */
+	public static function get_profile_properties() {
+		$properties = array(
+			'account_type'    => array(
+				'type'        => 'string',
+				'description' => __( 'Account type used for JetPack.', 'woocommerce-admin' ),
```
justinshreve: Should be a lowercase P in Jetpack  😆

includes/api/class-wc-admin-rest-onboarding-profile-controller.php
```diff
@@ -0,0 +1,330 @@
...
+				'context'     => array( 'view' ),
+				'readonly'    => true,
+			),
+			'store_address_1' => array(
```
justinshreve: Part of the discussion in Slack involved the store address. I think we can probably just save this using the normal WC settings API -- and just use this profile endpoint for all the other fields.

includes/api/class-wc-admin-rest-onboarding-profile-controller.php
joshuatf: Namespaced this and went ahead and namespaced the onboarding levels controller as well.

includes/api/class-wc-admin-rest-onboarding-profile-controller.php
joshuatf: Must be some weird muscle memory happening there.  `lowercase_P_dangit`

includes/api/class-wc-admin-rest-onboarding-profile-controller.php
joshuatf: I had mixed feelings on this since what we could collect here might be different from something later set in WC settings, but I think in retrospect it's probably not necessary to maintain two copies of the store address information.

I dropped these properties, but do you think we should merge in these store setting fields in the GET request here or make a separate API call to settings if we need to grab them?

joshuatf:
Thanks for the feedback, @justinshreve.  Made a few updates and also left a question about whether or not we want to include store settings in this API request.

includes/api/class-wc-admin-rest-onboarding-levels-controller.php
```diff
@@ -21,7 +21,7 @@ class WC_Admin_REST_Onboarding_Levels_Controller extends WC_REST_Data_Controller
 	 *
 	 * @var string
 	 */
-	protected $namespace = 'wc/v4';
+	protected $namespace = 'wc-internal/v1';
```
mikejolley: @justinshreve I know we discussed this on slack but perhaps `wc-admin` would be more appropriate which follows `wc-blocks` lead - both are internal APIs (we just need to work out how to lock them all down).

includes/api/class-wc-admin-rest-onboarding-profile-controller.php
```diff
@@ -0,0 +1,300 @@
...
+		return apply_filters( 'woocommerce_rest_prepare_onboarding_profile', $response, $item, $request );
+	}
+
+	/**
+	 * Get onboarding profile properties.
+	 *
+	 * @return array
+	 */
+	public static function get_profile_properties() {
+		$properties = array(
```
mikejolley: I noticed none of these had any validation in place (enum) - should we limit some of the values that can be set or is it not important?

joshuatf: I was mostly avoiding setting validation since I didn't know if all these values had been finalized yet.  Maybe @justinshreve can chime in here as I think he may know whether they are or not.

justinshreve: To the best of my knowledge, the screens there are finalized so this should be our final list of questions, so some validation might not hurt. These questions have also persisted through design updates, it's just the ordering that has changed.

I'll post some notes below:

**Industry**
_Multiselect_
Options:
* Fashion, apparel & accessories
* Health & beauty
* Art, music, & photography
* Food & Drink
* Home, furniture & garden
* Other

**Product Types**
_Multiselect_
Options:
* Physical products
* Downloads
* Subscriptions
* Memberships
* Composite Products
* Spaces
* Rentals

**Product Count**
_Single Selection_
Options:
* 1-10
* 11-100
* 101-1000
* 1000+

**Selling Venues/Already Selling**
_Single Selection_
Options:
* No
* Yes, on another platform
* Yes, at a brick and motor store
* Yes, on another platform and at a brick and mortar store

**Other Platform**
_Single Selection_
Options:
* Shopify
* BigCommerce
* Magento
* Wix
* Other

**Theme**
_Single Selection_
This should probably just be a theme slug, so maybe some simple validation against the slug pattern?

includes/api/class-wc-admin-rest-onboarding-profile-controller.php
```diff
@@ -0,0 +1,300 @@
...
+				'context'     => array( 'view' ),
+				'readonly'    => true,
+			),
+			'theme'          => array(
+				'type'        => 'string',
+				'description' => __( 'Selected store theme.', 'woocommerce-admin' ),
+				'context'     => array( 'view' ),
+				'readonly'    => true,
+			),
+			'purchase'       => array(
```
justinshreve: In addition to the `purchase` bool, we should probably have another boolean indicating if they skipped the entire profiler.

includes/api/class-wc-admin-rest-onboarding-profile-controller.php
justinshreve: Good question. Looking at the designs again, I don't actually think we need the address information in the other steps, so we can leave them out of the GET request for now. I was thinking maybe `country` since we will be showing prices of extensions, but we would just be using the user's WP language for that.

If we do end of needing data in the profiler though, I have no issues adding it here later since this is an internal endpoint.

includes/api/class-wc-admin-rest-onboarding-profile-controller.php
```diff
@@ -0,0 +1,300 @@
...
+	/**
+	 * Get onboarding profile properties.
+	 *
+	 * @return array
+	 */
+	public static function get_profile_properties() {
+		$properties = array(
+			'account_type'   => array(
+				'type'        => 'string',
+				'description' => __( 'Account type used for Jetpack.', 'woocommerce-admin' ),
```
justinshreve: What's the account type setting for? Would this ever be anything other than just `jetpack` seen in your screenshot?

includes/api/class-wc-admin-rest-onboarding-profile-controller.php
joshuatf: Added in https://github.com/woocommerce/woocommerce-admin/pull/2217/commits/c2252a3738966745db716e202276ff173603eda6

includes/api/class-wc-admin-rest-onboarding-profile-controller.php
joshuatf: The account type can be `new`, `existing`, or `google` if they choose to connect their Google account.  If you think this is property is unnecessary, let me know and I'll remove.

includes/api/class-wc-admin-rest-onboarding-levels-controller.php
joshuatf: Updated to `wc-admin/v1`.

joshuatf:
Thanks for compiling that list @justinshreve!  Added in validation based on those parameters and updated the namespace to `wc-admin/v1` as @mikejolley noted.

includes/api/class-wc-admin-rest-onboarding-profile-controller.php
justinshreve: Gotcha! I think that could be valuable to track/know so we should keep it 👍.

tests/api/onboarding-profile.php
```diff
@@ -0,0 +1,146 @@
...
+		$this->assertArrayHasKey( 'selling_venues', $properties );
+		$this->assertArrayHasKey( 'other_platform', $properties );
+		$this->assertArrayHasKey( 'theme', $properties );
+		$this->assertArrayHasKey( 'items_purchased', $properties );
+	}
+
+	/**
+	 * Test that profiles response changes based on applied filters.
+	 */
+	public function test_profile_extensibility() {
```
justinshreve: Thanks for keeping extendability in mind, and for this test!

PR #2219, opened 2019-05-13
Add filters for customer roles

### LINKED ISSUE #2195: Customer batch import attempts to import non-customers
**Describe the bug**
The customers batch action queries all users regardless of whether or not they are a customer in https://github.com/woocommerce/woocommerce-admin/blob/ab8020531a31d17b2f1c84a1c43778c11b4e0e9b/includes/class-wc-admin-reports-sync.php#L537.

It then attempts to process them using `WC_Admin_Reports_Customers_Data_Store::update_registered_customer()` which will filter out users without orders or users that don't have a role of `customer`.

This doesn't cause any huge issues currently, but won't allow skipping of those users during the import process and will create inaccurate numbers for import stats.

**To Reproduce**

1. Create admin and customer users.
2. Add an error log or dump the customer IDs in https://github.com/woocommerce/woocommerce-admin/blob/ab8020531a31d17b2f1c84a1c43778c11b4e0e9b/includes/class-wc-admin-reports-sync.php#L551
3. Note that non-customer users are queried during the import.

**Expected behavior**
Expected to not attempt to import these non-customer users or go ahead and import all users.

rrennick:
Fixes #2195

This PR limits the customer import to users with the `customer` role. It includes a `woocommerce_admin_import_customer_roles` filter to allow including other user roles.

- Go to Analytics -> Settings
- Delete previously imported data
- Start an import including `All`
- Verify Import scheduled actions complete
- Check the customer lookup table and verify non-customer users without orders are not imported.

### Test data store

- Remove a customer record from `wp_wc_customer_lookup`
- Edit that customer in the dashboard and add/change the billing address
- Check to see that the customer is added to `wp_wc_customer_lookup`
- Add a filter to `woocommerce_admin_customer_roles` and add `subscriber` to the array
- Edit a subscriber user that has no orders
- Check to see that the subscriber is added to `wp_wc_customer_lookup`

### DIFF
--- a/includes/class-wc-admin-reports-sync.php
+++ b/includes/class-wc-admin-reports-sync.php
@@ -274,7 +274,7 @@ class WC_Admin_Reports_Sync {
 				( self::SINGLE_ORDER_IMPORT_ACTION === $existing_job->get_hook() ) ||
 				(
 					self::QUEUE_DEPEDENT_ACTION === $existing_job->get_hook() &&
-					in_array( self::SINGLE_ORDER_IMPORT_ACTION, $existing_job->get_args() )
+					in_array( self::SINGLE_ORDER_IMPORT_ACTION, $existing_job->get_args(), true )
 				)
 			) {
 				return;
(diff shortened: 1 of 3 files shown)


rrennick:
@joshuatf Thanks. I've updated the PR and test instructions for the second filter. Let me know if you think they should both use the same filter.

PR #2220, opened 2019-05-13
Call the installer in `register_activation_hook`

mikejolley:
When testing out PRs I found my reports were broken due to a missing column in my DB (introduced [here](https://github.com/woocommerce/woocommerce-admin/blame/af6a19c7016901cc647a77e83d8a8276a1e52590/includes/class-wc-admin-install.php#L95)).

It seems that the installer only runs if the plugin version changes, so simply de/re-activating the plugin had no effect.

To prevent this, we need only call the installer on activation. This mirrors how WooCommerce call installs - once on activation, and then again if the version changes (upgrades). This will ensure the DB is always current.

- Deactivate
- Reactivate
- Ensure there are no errors on activation.

### DIFF
--- a/woocommerce-admin.php
+++ b/woocommerce-admin.php
@@ -172,7 +172,7 @@ function wc_admin_activate_wc_admin_plugin() {
 
 	// Run the installer on activation.
 	require_once WC_ADMIN_ABSPATH . 'includes/class-wc-admin-install.php';
-	WC_Admin_Install::install();
+	WC_Admin_Install::create_tables();
 }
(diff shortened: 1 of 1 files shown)


mikejolley:
Strange this is erroring on travis but not on my machine. investigating.

mikejolley:
Ok was missing classes. We can get away with just creating tables.

Should really add an autoloader at some point so we don't need to manage includes :)

timmyc:
@mikejolley thanks for the fix, i'll test it out shortly

> Should really add an autoloader at some point so we don't need to manage includes :)

Do you want to create an issue for that?

Issue #2214, opened 2019-05-13
Persisted Dates: Avoid adding date params to Customers, Stock, Settings pages


psealock:
**Describe the bug**

Some screens don't have a datepicker. Navigating to those pages persists the date params from the url, affecting the API queries.

**To Reproduce**

1. Revenue Report.
2. Add a date range, ie "last year".
3. Go to Customers, Stock, or settings page.
4. See the date range url params persist.

**Expected behavior**

Those date range related parameters should not persist.

**Question**

The date parameters also persist to extension-created screens. There may be a way to smartly persist the params only if the date range picker is on the screen.

psealock:
Related issue https://github.com/woocommerce/woocommerce-admin/issues/2176

Issue #2176, opened 2019-05-07
Customers Report: Time params applied to one endpoint, not the other


psealock:
**Describe the bug**

![Screen Shot 2019-05-07 at 2 46 41 PM](https://user-images.githubusercontent.com/1922453/57268232-260a4e80-70d7-11e9-815a-a006d39fa7ad.png)

Summary Numbers and the Table results disagree. 

It looks like time parameters (before and after) are being applied to `/reports/customers` but not `/reports/customers/stats`. 

**To Reproduce**

1. Go to `/wp-admin/admin.php?page=wc-admin#/analytics/customers`.
2. See the discrepancy between Summary Numbers and the Table results.

**Expected behavior**

Numbers should match.

PR #2225, opened 2019-05-14
Persisted Queries: Avoid adding to non time related screens

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/2214
Fixes https://github.com/woocommerce/woocommerce-admin/issues/2176

Some screens don't have a datepicker. Navigating to those pages now don't persist time-related url params.

This PR also introduces Node at LTS for our build process. This is what [Gutenberg](https://github.com/WordPress/gutenberg/blob/master/.nvmrc) uses.

1. Go to Dashboard.
2. Choose a date range.
3. Inspect wc-admin urls. With the exception of devdocs, stock, settings, and customers pages, all urls should have a time related query, eg `?period=last_week&compare=previous_year`.
4. Inspect other urls to make sure no regressions occurred.

### DIFF
--- a/client/layout/controller.js
+++ b/client/layout/controller.js
@@ -5,6 +5,7 @@
 import { Component, createElement } from '@wordpress/element';
 import { parse } from 'qs';
 import { find, last, isEqual } from 'lodash';
+import { applyFilters } from '@wordpress/hooks';
 
 /**
  * WooCommerce dependencies
@@ -19,6 +20,8 @@ import AnalyticsReport from 'analytics/report';
 import AnalyticsSettings from 'analytics/settings';
 import Dashboard from 'dashboard';
 import DevDocs from 'devdocs';
+
+const TIME_EXCLUDED_SCREENS_FILTER = 'woocommerce_admin_time_excluded_screens';
 
 const getPages = () => {
 	const pages = [];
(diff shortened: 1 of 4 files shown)


client/layout/controller.js
```diff
@@ -124,6 +127,12 @@ class Controller extends Component {
 // Update links in wp-admin menu to persist time related queries
 window.wpNavMenuUrlUpdate = function( page, query ) {
 	const search = stringifyQuery( getPersistedQuery( query ) );
+	const timeExcludedScreens = applyFilters( TIME_EXCLUDED_SCREENS_FILTER, [
+		'devdocs',
+		'stock',
+		'settings',
+		'customers',
+	] );
```
psealock: @rrennick, you'll want to add `scheduled-actions` to this array in your plugin so that time-related queries don't get added to the url for the `Scheduled Actions` link in the sidebar. 

For example, scheduled actions sidebar link may look like this:

```
http://tangaroa.test/wp-admin/admin.php?page=wc-admin#/analytics/scheduled-actions?period=last_week&compare=previous_year
```

jeffstieler: Woah, @rrennick do you have a Scheduled Actions report built on this codebase?

jeffstieler: ✨ https://github.com/Prospress/action-scheduler-admin ✨

rrennick: I also need to update it for the new page registering code.

psealock:
Thanks @timmyc 

TIL regex negative lookbehinds are not supported by all browsers. I simplified the regex and removed the lookbehind. Its ready for another review.

client/layout/controller.js
```diff
@@ -121,28 +124,55 @@ class Controller extends Component {
...
+ * Update an anchor's link in sidebar to include persisted queries. Leave excluded screens
+ * as is.
+ *
+ * @param {HTMLElement} item - Sidebar anchor link.
+ * @param {string} nextQuery - A query string to be added to updated hrefs.
+ * @param {Array} excludedScreens - wc-admin screens to avoid updating.
+ */
+export function updateLinkHref( item, nextQuery, excludedScreens ) {
+	/**
+	 * Regular expression with a lookbehind for "page=wc-admin", matching any WooCommerce Admin screen.
```
Aljullu: @psealock should this comment be updated now that you removed the lookbehind?

psealock:
>But I found a behavior that I'm not sure if it's correct

The behaviour you're seeing is expected for the scope of this PR, but brings up an interesting point.

#2250 Is to set a default range, which IMO means that when no parameters are in the query, a default is chosen. For now, the default is hard coded to "Month to date" and #2250 calls for allowing the use to choose instead.

What you're proposing here is to persist the time related queries (whether default or otherwise) to not just the next screen but any subsequent screens until changed.

@LevinMedia, what do you think about this^^^. My opinion is to follow through with #2250 and see if we get any feedback, but we can easily store the query and retrieve it too.

client/layout/controller.js
psealock: Fixed in https://github.com/woocommerce/woocommerce-admin/pull/2225/commits/a25746bb59480104abdd3360e0f5ef359820fd1d

PR #2226, opened 2019-05-14
Refactor main file into singleton class structure

mikejolley:
There were some issues with the main feature plugin file I found whilst doing reviews:

- Mix of functions used by the entirety of the plugin, and functions specific to the feature plugin
- Errors if activating without WooCommerce installed
- Mix of notices vs deactivation if dependencies were not met (it deactivated but showed no notice)
- Installer was not ran when activating (https://github.com/woocommerce/woocommerce-admin/pull/2220)

This PR refactors the main feature plugin file into a class which handles:

- defining constants
- including files
- adding/removing hooks
- activation/deactivation notices

Other non-main file functions have been moved. I create an events class and a functions class for these to live.

To test you'll want to check out the PR and ensure the feature plugin continues to work without err.
You can also test activating/deactivating with and without dependencies to make sure things work as expected.

cc @justinshreve Please let me know if I've moved anything incorrectly, or created new class files in the wrong places. I don't know if my changes will affect builds.


includes/class-wc-admin-install.php
```diff
@@ -250,10 +250,27 @@ public static function wpmu_drop_tables( $tables ) {
 	/**
 	 * Update WC Admin version to current.
 	 */
-	private static function update_wc_admin_version() {
+	protected static function update_wc_admin_version() {
```
rrennick: For consistency should this be `update_db_version` ?

woocommerce-admin.php
```diff
@@ -15,305 +15,302 @@
...
+	 * Define Constants.
+	 */
+	protected function define_constants() {
+		$this->define( 'WC_ADMIN_APP', 'wc-admin-app' );
+		$this->define( 'WC_ADMIN_ABSPATH', dirname( __FILE__ ) . '/' );
+		$this->define( 'WC_ADMIN_DIST_JS_FOLDER', 'dist/' );
+		$this->define( 'WC_ADMIN_DIST_CSS_FOLDER', 'dist/' );
+		$this->define( 'WC_ADMIN_FEATURES_PATH', WC_ADMIN_ABSPATH . 'includes/features/' );
+		$this->define( 'WC_ADMIN_PLUGIN_FILE', __FILE__ );
+		$this->define( 'WC_ADMIN_VERSION_NUMBER', '0.11.0' );
```
justinshreve: I believe this needs bumped to `0.12.0` now.

woocommerce-admin.php
```diff
@@ -15,305 +15,302 @@
...
+		remove_filter( 'woocommerce_settings_groups', array( 'WC_Admin_Library', 'add_settings_group' ) );
+		remove_filter( 'woocommerce_settings-wc_admin', array( 'WC_Admin_Library', 'add_settings' ) );
+
+		remove_action( 'admin_head', array( 'WC_Admin_Library', 'update_link_structure' ), 20 );
+
+		/*
+		* Remove the emoji script as it always defaults to replacing emojis with Twemoji images.
+		* Gutenberg has also disabled emojis. More on that here -> https://github.com/WordPress/gutenberg/pull/6151
+		*/
+		remove_action( 'admin_print_scripts', 'print_emoji_detection_script' );
```
justinshreve: I'm thinking this hook removal actually belongs in `WC_Admin_Loader` so we run it in core too. I must have missed it previously.

mikejolley:
Thanks @justinshreve. I've made those changes and updated to master again. Is anyone else available for second look before it goes out of sync again?

Issue #2222, opened 2019-05-13
Inbox: Repeated SQL Queries on woocommerce.com


timmyc:
**Describe the bug**
With v0.11 running on the dev environment of WooCommerce.com, the database server was getting exhausted with repeated slow queries for:

`SELECT COUNT( * ) FROM wp_posts WHERE post_type = ‘shop_order’ AND post_status = ‘wc-completed’`

The code was https://github.com/woocommerce/woocommerce-admin/blob/master/includes/notes/class-wc-admin-notes-order-milestones.php#L104 to `get_orders_count()` in the init method of the `milestone` class.

In order to stabilize the environment, I had to set an early `return 0` in `get_orders_count()`.

PR #2227, opened 2019-05-14
Notes: Add filters to disable milestone and sales record notes.

timmyc:
The idea for this filter was brought up after we experienced some performance issues on woocommerce.com. While we are exploring a better fix for that in #2222 - it also feels like adding in some filters to disable the order milestone and sales record notes seems like a good thing to have. So here we are!

I originally put the filters for the order milestone in the `init()` method, but figured having the checks at the areas that actually create the notifs would capture any direct calls to those public methods too. But keen for feedback on that placement @jeffstieler 

- Verify that all boots up still, you can add some filters to your theme or elsewhere to test out if you please.

### DIFF
--- a/includes/notes/class-wc-admin-notes-order-milestones.php
+++ b/includes/notes/class-wc-admin-notes-order-milestones.php
@@ -124,6 +124,11 @@ class WC_Admin_Notes_Order_Milestones {
 	 * Used to avoid celebrating milestones that were reached before plugin activation.
 	 */
 	public function backfill_last_milestone() {
+		// If the milestone notes have been disabled via filter, bail.
+		if ( ! $this->are_milestones_enabled() ) {
+			return;
+		}
+
 		$this->set_last_milestone( $this->get_current_milestone() );
 	}
 
@@ -263,9 +268,30 @@ class WC_Admin_Notes_Order_Milestones {
 	}
 
 	/**
+	 * Convenience method to see if the milestone notes are enabled.
+	 */
+	public function are_milestones_enabled() {
+		/**
+		 * Filter to allow for disabling order milestones.
+		 *
+		 * @since 3.7.0
+		 *
+		 * @param boolean default true
+		 */
+		$milestone_notes_enabled = apply_filters( 'woocommerce_admin_order_milestone_enabled', true );
+
+		return $milestone_notes_enabled;
+	}
+
+	/**
 	 * Add milestone notes for other significant thresholds.
 	 */
 	public function other_milestones() {
+		// If the milestone notes have been disabled via filter, bail.
+		if ( ! $this->are_milestones_enabled() ) {
+			return;
+		}
+
 		$last_milestone    = $this->get_last_milestone();
 		$current_milestone = $this->get_current_milestone();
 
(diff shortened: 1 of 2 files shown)


includes/notes/class-wc-admin-notes-order-milestones.php
```diff
@@ -262,10 +267,31 @@ public function get_note_action_for_milestone( $milestone ) {
...
+	 */
+	public function are_milestones_enabled() {
+		/**
+		 * Filter to allow for disabling order milestones.
+		 *
+		 * @since 3.7.0
+		 *
+		 * @param boolean default true
+		 */
+		$milestone_notes_enabled = apply_filters( 'woocommerce_admin_order_milestone_enabled', true );
```
jeffstieler: Nitpick: milestone vs. milestones.

includes/notes/class-wc-admin-notes-order-milestones.php
```diff
@@ -124,6 +124,11 @@ public function get_orders_count( $no_cache = false ) {
 	 * Used to avoid celebrating milestones that were reached before plugin activation.
 	 */
 	public function backfill_last_milestone() {
+		// If the milestone notes have been disabled via filter, bail.
+		if ( ! $this->are_milestones_enabled() ) {
```
jeffstieler: What do you think about putting this in the constructor instead so we avoid scheduling events and adding hooks entirely?

mikejolley: Potentially too early? A lot of the constructors load as the plugins load, so unless your plugin/code runs before then the filters will be added.

It's probably ok to be here so that if it's added via other actions, or called manually, the new filter still applies.

includes/notes/class-wc-admin-notes-order-milestones.php
```diff
@@ -262,10 +267,31 @@ public function get_note_action_for_milestone( $milestone ) {
 		}
 	}
 
+	/**
+	 * Convenience method to see if the milestone notes are enabled.
```
mikejolley: Missing `@return` docblock

includes/notes/class-wc-admin-notes-order-milestones.php
timmyc: Indeed per @mikejolley's comment - I hit issues in testing with the filter in the constructor, and also since the method could be called manually it seemed like having the check here would be safest.

timmyc:
> It looks as though milestone code is spread out throughout the codebase. Do we need a global function (since there is no single class to handle milestone events)?

Not sure I follow you here @mikejolley - aren't all the milestone notes within the one class in `class-wc-admin-notes-order-milestones.php`?

> We could also consider a standard filter name e.g. woocommerce_admin_allow_milestone which passes the milestone name and a bool. That would allow you to target specific milestones by name, or disable all at once.

That is an interesting idea - I just don't know if someone would want certain order milestones to be shown and not others. My main goal was just to provide a way to disable them entirely - I kind of felt like people might want them or not at all... but perhaps someone might want to only show certain levels. Feels edge case though.

timmyc:
Thanks for the reviews, updates per feedback in e46c405

PR #2228, opened 2019-05-14
add match=any support for advanced orders table filter

### LINKED ISSUE #2188: Orders match filter doesn't affect the orders table
**Describe the bug**
The orders match filter can be change to "Any" to allow any of the rules to apply.  `WC_Admin_Reports_Orders_Data_Store` doesn't take this param into account so all filters must always be matched.

**To Reproduce**

1. Go to the orders report.
2. Change the match filter to "Any" and add several filters.
3. Note the summary numbers are correct, but the table results are not.

**Expected behavior**
Expected the match filter to work for both table and summary numbers.

**Screenshots**
Screen Shot 2019-05-08 at 1 55 11 PM

rrennick:
Fixes #2188

This PR add the match query parameter for the order stats table and implements the operator processing used for calculating the table summary.

- Go to Analytics -> Orders
- Choose Advanced filters
- Change Order Match to `Any`
- Add 2 or more filters
- Check that table contains row matching any of the filters
- Check that table summary matches the detail data in the table rows

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-orders-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-orders-data-store.php
@@ -79,7 +79,7 @@ class WC_Admin_Reports_Orders_Data_Store extends WC_Admin_Reports_Data_Store imp
 
 		$status_subquery = $this->get_status_subquery( $query_args );
 		if ( $status_subquery ) {
-			$where_subquery[] = $status_subquery;
+			$sql_query_params['where_clause'] .= " AND {$status_subquery}";
 		}
 
 		if ( $query_args['customer_type'] ) {
@@ -120,7 +120,7 @@ class WC_Admin_Reports_Orders_Data_Store extends WC_Admin_Reports_Data_Store imp
 		}
 
 		if ( 0 < count( $where_subquery ) ) {
-			$sql_query_params['where_clause'] = ' AND (' . implode( " {$operator} ", $where_subquery ) . ')';
+			$sql_query_params['where_clause'] .= ' AND (' . implode( " {$operator} ", $where_subquery ) . ')';
 		}
 
 		return $sql_query_params;
--- a/includes/data-stores/class-wc-admin-reports-orders-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-orders-data-store.php
@@ -79,7 +79,11 @@ class WC_Admin_Reports_Orders_Data_Store extends WC_Admin_Reports_Data_Store imp
 
 		$status_subquery = $this->get_status_subquery( $query_args );
 		if ( $status_subquery ) {
-			$sql_query_params['where_clause'] .= " AND {$status_subquery}";
+			if ( empty( $query_args['status_is'] ) && empty( $query_args['status_is_not'] ) ) {
+				$sql_query_params['where_clause'] .= " AND {$status_subquery}";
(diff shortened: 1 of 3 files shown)


includes/data-stores/class-wc-admin-reports-orders-data-store.php
```diff
@@ -70,24 +70,28 @@ public function __construct() {
...
+		$where_subquery           = array();
 
 		$sql_query_params = $this->get_time_period_sql_params( $query_args, $order_stats_lookup_table );
 		$sql_query_params = array_merge( $sql_query_params, $this->get_limit_sql_params( $query_args ) );
 		$sql_query_params = array_merge( $sql_query_params, $this->get_order_by_sql_params( $query_args ) );
 
 		$status_subquery = $this->get_status_subquery( $query_args );
 		if ( $status_subquery ) {
-			$sql_query_params['where_clause'] .= " AND {$status_subquery}";
+			$where_subquery[] = $status_subquery;
```
joshuatf: I think we'll need to apply this status subquery without an `OR` match operator unless the status query is explicitly set.  Since `get_status_subquery()` by default returns the 'normal' statuses, an `OR` operator would also return excluded statuses even if the user didn't add a status filter.

The code in the order stats data store probably does a better job of explaining this than I did - https://github.com/woocommerce/woocommerce-admin/blob/bc5dc57703b856e95c512883ab85ffa0e4639f41/includes/data-stores/class-wc-admin-reports-orders-stats-data-store.php#L173

rrennick: @joshuatf Thanks. Restored the AND for status in e192317.

rrennick: @joshuatf I've updated the PR with that logic. However, what I noticed is that the default query excludes trash, pending, failed and cancelled orders. When the OR is added for an Any + any status then the default excluded statuses are not excluded. 

As a user I would expect the original exclusion to persist unless I had explicitly added them. To provide that behaviour we would need to mask the default excludes to remove explicitly requested statuses then include both subqueries. Is that the expected/desired behaviour?

joshuatf: I see what you mean here.  I think that could be expected behavior, but I can definitely see the argument for wanting to continue to exclude statuses.  For anyone else following here, if you applied the following filter:

Screen Shot 2019-06-17 at 12 56 14 PM

This could potentially allow `failed` (or any of your other excluded statuses) to now show if they had a returning customer order with a `failed` status.

If we did opt to change this, I think the best method would be to:
1. Continue appending the current and/or status subquery.
2. Check if the applied status filter is in the array of excluded statuses (i.e., they're explicitly querying excluded statuses).
3. If not- append an `AND` query to negate any excluded statuses.
4. If yes- do not append further queries.

rrennick: @joshuatf Thanks for the feedback. That logic makes sense. If we implement it we end up with a situation where a user adds the following with `Any`

- Status: completed
- Product: Product X

The rows of completed Product X are already included in the completed status selection and the product filter has no effect. Either way we are going to end up with unexpected results for some users.

I'll merge this and if user feedback warrants we can change how the filters interact at a later date.

PR #2241, opened 2019-05-15
Add step wrapper/controller to onboarding profile wizard

### LINKED ISSUE #1882: Onboarding - Profile Wizard: Wrapper/Controller
Each screen of the profile wizard contains a header area (footer on mobile). On the initial connection pages, this should show a WooCommerce/Jetpack logo. On the profiler pages, this should show a stepper to indicate progress (see #2164).

This issue involves implementing a wrapper/page controller for the wizard that allows us to add new pages to the wizard easily, and retain the header and other common elements. It should also pass into the stepper what the current page is.

Screen Shot 2019-05-14 at 8 22 01 PM

Screen Shot 2019-05-02 at 2 43 57 PM

Screen Shot 2019-05-03 at 9 05 57 AM

joshuatf:
Fixes #1882 

* Adds a step controller to allow switching between steps based on query params.
* Adds a profile wizard header.
* Adds base for plugin step.

A couple notes on this: /cc @justinshreve 
* Initially separated out the controller, but opted to add the controller logic directly to `profile-wizard/index` since it seemed mostly redundant.  Let me know if you see a good reason to separate.
* The header logo still needs to be updated to the new version.  If you can point to where I can find the png for the new one, I'll update.

### Screenshots
Screen Shot 2019-05-15 at 3 08 53 PM

1. Go to the dashboard.
2. Set `const profileWizardComplete = false;`.
3. Check that the header exists and styling is correct.
4. Add `?step=plugin` to the URL to navigate to the next step.
5. Make sure the next step is shown.


joshuatf:
Thanks again for the logo, @justinshreve!  Added that (had to add a small transform adjustment as the SVG seems like slightly off-center).

>When testing, I noticed that the divider line did not stretch the whole length of my window (see the right corner)

I _think_ this is the layout padding that was removed in #2218 so I gave this a rebase.  Let me know if you still see this spacing.

PR #2246, opened 2019-05-16
Add Stepper component

### LINKED ISSUE #2164: Onboarding - Profile Wizard: Stepper Component
Once a Jetpack connection has been made, the profile wizard contains a stepper to indicate progress. See #1882.

This issue involves implementing a reusable and exportable stepper component following Muriel guidelines.

The component should accept a list of steps along with a boolean if that step has been completed.

Screen Shot 2019-05-02 at 2 43 57 PM

Screen Shot 2019-05-03 at 9 05 57 AM

joshuatf:
Fixes #2164 

Adds in a Muriel style stepper component to be used with the onboarding profiler.

### Thoughts
* Made steps that are prior to the `currentStep` complete by default unless a boolean is passed specifying otherwise.  Still allows the `isComplete` boolean for a step to override this and/or mark the final step complete.
* I did my best to match this design around other Muriel components, but this is not an existing component with guidelines so it's a bit mix and match.  Obviously it's contrasted quite heavily with existing wc-admin components, but I believe that's the direction we'll be headed in for most components going forward.

### Screenshots
Screen Shot 2019-05-16 at 5 09 23 PM
Screen Shot 2019-05-16 at 5 09 12 PM

1.  Open the devdocs `wp-admin/admin.php?page=wc-admin#/devdocs`.
2.  Go through the steps in the stepper component.
3.  Make sure it works as expected.
4.  Make sure styling is correct.

### DIFF
--- a/packages/components/src/stepper/example.md
+++ b/packages/components/src/stepper/example.md
@@ -29,7 +29,7 @@ const MyStepper = withState( {
         steps.forEach( s => s.isComplete = true );
     }
 
-	return (
+    return (
         <div>
             { isComplete ? (
                 <button onClick={ () => setState( { currentStep: 'first', isComplete: false } ) } >
--- a/packages/components/src/stepper/example.md
+++ b/packages/components/src/stepper/example.md
@@ -2,8 +2,9 @@
 import { Stepper } from '@woocommerce/components';
 
 const MyStepper = withState( {
-	currentStep: 'first',
-} )( ( { currentStep, setState } ) => {
+    currentStep: 'first',
+    isComplete: false,
+} )( ( { currentStep, isComplete, setState } ) => {
     const steps = [
         {
             label: 'First',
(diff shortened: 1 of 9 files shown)


justinshreve:
@joshuatf One more note -- I think we need to update the changelog for the components package as well.

packages/components/src/stepper/example.md
```diff
@@ -0,0 +1,68 @@
...
+            key: 'fourth',
+        },
+    ];
+    const currentIndex = steps.findIndex( s => currentStep === s.key );
+
+    if ( isComplete ) {
+        steps.forEach( s => s.isComplete = true );
+    }
+
+	return (
```
timmyc: minor tabbing a bit out of sorts here with the above.

timmyc:
This is looking great - thanks for... stepping up... to build this new component Josh.

</dadjokes>

packages/components/src/stepper/style.scss
```diff
@@ -0,0 +1,71 @@
...
+		padding: $gap-small;
+		font-weight: 600;
+
+		svg {
+			display: none;
+		}
+
+		&.is-active,
+		&.is-complete {
+			color: #6802f3;
```
psealock: The designs look like they introduce a new color scheme. Now (or in follow up) seems like a good time to introduce some variables.

packages/components/src/stepper/example.md
joshuatf: Thanks, Timmy!  Updated 😄

packages/components/src/stepper/style.scss
joshuatf: Great idea @psealock- added in https://github.com/woocommerce/woocommerce-admin/pull/2246/commits/1ea32e1013836d0a342c390400710b26ef7332f4

joshuatf:
Thanks for the ✅ @justinshreve!  Added font sizing and updated colors to the new Figma designs.  I imagine we might need change the naming of the color vars in here (`$muriel-active`) once we start adding in more components, but I think this will do for now.

PR #2247, opened 2019-05-16
[Experiment] Category Lookup Table

mikejolley:
This is an experiment.

Creates a category lookup table which stores term relationships **for the full category hierarchy.**

This allows the reports to group by `category_id` without having sales from child categories ignored..

### DIFF
--- a/includes/class-wc-admin-install.php
+++ b/includes/class-wc-admin-install.php
@@ -189,6 +189,11 @@ class WC_Admin_Install {
 			UNIQUE KEY user_id (user_id),
 			KEY email (email)
 		) $collate;
+		CREATE TABLE {$wpdb->prefix}wc_product_category_lookup (
+			product_id BIGINT UNSIGNED NOT NULL,
+			category_id BIGINT UNSIGNED NOT NULL,
+			PRIMARY KEY (product_id,category_id)
+		) $collate;
 		";
 
 		return $tables;
@@ -220,6 +225,7 @@ class WC_Admin_Install {
 			"{$wpdb->prefix}wc_admin_notes",
 			"{$wpdb->prefix}wc_admin_note_actions",
 			"{$wpdb->prefix}wc_customer_lookup",
+			"{$wpdb->prefix}wc_product_category_lookup",
 		);
 	}
 
--- a/includes/class-wc-admin-product-category-lookup.php
+++ b/includes/class-wc-admin-product-category-lookup.php
@@ -120,6 +120,8 @@ class WC_Admin_Product_Category_Lookup {
 	 * @param int $product_id Product to update.
 	 */
 	protected function update_product_relationships( $product_id ) {
+		global $wpdb;
+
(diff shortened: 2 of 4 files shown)


includes/class-wc-admin-product-category-lookup.php
```diff
@@ -0,0 +1,224 @@
...
+		unset( $this->edited_product_cats[ $category_id ] );
+	}
+
+	/**
+	 * Sync product on save.
+	 *
+	 * @param  int    $post_id Post ID being saved.
+	 * @param  object $post Post object.
+	 */
+	public function sync_products( $post_id, $post ) {
```
timmyc: Any performance concerns for running this on the save events vs using an action scheduler job?

mikejolley: @timmyc Since I just wanted to get this up and running I didn't follow existing patterns so I'm open to modifying approach. The only thing I can think of is, as a lookup table it's probably better to be current, and by not deferring to AS you can run it on a per-product basis only when changes occur.

mikejolley: (if this moved to core it would be in the product data store and run if categories are changed)

timmyc: > as a lookup table it's probably better to be current

Great point, I think queuing reporting-only lookup table operations is a bit "safer" but any lookup table that can impact other admin/front-end views it might be wise to just take the extra queries right away.

rrennick: Most stores won't be changing categories that often. With this, the category hierarchy will already be in memory (for those without persistent caching) so performance wise doing the direct save might be better than queueing to AS.

timmyc:
also /cc @joshuatf for your thoughts since you spent some time on this problem too.

includes/class-wc-admin-install.php
```diff
@@ -189,6 +189,11 @@ private static function get_schema() {
 			UNIQUE KEY user_id (user_id),
 			KEY email (email)
 		) $collate;
+		CREATE TABLE {$wpdb->prefix}wc_product_category_lookup (
+			product_id BIGINT UNSIGNED NOT NULL,
```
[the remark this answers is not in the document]
mikejolley: Getting depth might be tricky as the hierarchy WP stores is flat. Doing this would require a query for each and every term in the store which won't scale.

We might be able to do what we need without depth. It won't be as useful for all use cases, but it will allow quick lookup of all decendants.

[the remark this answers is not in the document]
mikejolley: @joshuatf As I feared, depth is too much of a performance bottleneck -.- It requires looping over every term, and I also found that if you update a term you need to update ALL parents and ALL descendants because depth may change.

I'll scrapping depth and just storing hierarchy instead. That is working and it's more efficient.

rrennick:
@mikejolley Okay to close this one?

PR #2251, opened 2019-05-16
Add a prompt for Changelog entry in PR template.

timmyc:
Adds a prompt to the PR template to include a changelog entry.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -20,3 +20,7 @@ _Replace this with a good description of your changes & reasoning._
 
 <!--- Note: When displaying information based on sample data, such as SwaggerHub, 
 be sure to detail parts affected in Release Notes --->
+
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -20,3 +20,7 @@ _Replace this with a good description of your changes & reasoning._
 
 <!--- Note: When displaying information based on sample data, such as SwaggerHub, 
 be sure to detail parts affected in Release Notes --->
+
+### Changelog Note:
+
+<!--- Please follow the WooCommerce core format using prefixes of Enhancement, Tweak, Dev, Fix  --->
```
joshuatf: Should we also include `Performance`?

timmyc: Yeah I thought about that, but then wasn't sure if we would be doing much Performance work here... but no harm in adding it I suppose

PR #2252, opened 2019-05-17
Update Card component to use Muriel design

### LINKED ISSUE #2165: Onboarding - Profile Wizard: Card Component
See p1556119186006300-slack-wc-onboarding for Figma designs on the card component. 

This issue involves implementing a reusable and exportable card component following Muriel design guidelines. Note that we already have a `wc-admin` card component, so we may want to update the styles of that.

These cards are used throughout the UI for showing information, and other elements like form inputs.

Screen Shot 2019-05-03 at 1 30 18 PM

joshuatf:
Fixes #2165

Updates the existing Card component to use the Muriel design and adds a class to override styles for the Cards used in analytic reports.

### Screenshots
Screen Shot 2019-05-17 at 6 37 31 PM

1.  Find the card in the devdocs `/wp-admin/admin.php?page=wc-admin#/devdocs`.
2.  Make sure the styling matches the Muriel component guidelines.
3.  Make sure that existing cards used around reports are still styled in the wc-admin fashion.

### DIFF
--- a/client/stylesheets/abstracts/_variables.scss
+++ b/client/stylesheets/abstracts/_variables.scss
@@ -33,3 +33,7 @@ $alert-red: $error-red;
 // WordPress defaults
 $adminbar-height: 32px;
 $adminbar-height-mobile: 46px;
+
+// Muriel
+$muriel-card-box-shadow: 0 1px 3px rgba(0, 0, 0, 0.2), 0 2px 2px rgba(0, 0, 0, 0.12),
+	0 0 2px rgba(0, 0, 0, 0.14);
--- a/client/stylesheets/abstracts/_variables.scss
+++ b/client/stylesheets/abstracts/_variables.scss
@@ -35,5 +35,5 @@ $adminbar-height: 32px;
 $adminbar-height-mobile: 46px;
 
 // Muriel
-$muriel-card-box-shadow: 0 1px 3px rgba(0, 0, 0, 0.2), 0 2px 2px rgba(0, 0, 0, 0.12),
-	0 0 2px rgba(0, 0, 0, 0.14);
+$muriel-card-box-shadow: 0 2px 1px -1px rgba(0, 0, 0, 0.2), 0 1px 1px 0 rgba(0, 0, 0, 0.14),
+	0 1px 3px 0 rgba(0, 0, 0, 0.12);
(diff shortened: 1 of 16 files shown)


client/stylesheets/abstracts/_variables.scss
```diff
@@ -33,3 +33,7 @@ $alert-red: $error-red;
 // WordPress defaults
 $adminbar-height: 32px;
 $adminbar-height-mobile: 46px;
+
+// Muriel
+$muriel-card-box-shadow: 0 1px 3px rgba(0, 0, 0, 0.2), 0 2px 2px rgba(0, 0, 0, 0.12),
```
jameskoster: This shadow doesn't match the 1dp elevation styles.

It would be cool to establish vars for all different elevation heights.

LevinMedia: @joshuatf @jameskoster - should I be concerned with the type styles and padding for card headers at this point? Or are we just focused on the actual card styles?

My concern with the size of the card header title in this example is we'll run out of space for things like charts and tabular data where we have other inputs and controls included in the card header. Here's an example: 

Screen Shot 2019-05-17 at 3 34 54 PM

if we use the new Muriel type scale, I think we'd want to use either "Subtitle"  or "Subtitle Small"

Screen Shot 2019-05-17 at 3 39 00 PM

This is an exploration I was working for advanced filters that uses "Subtitle" in the card header. The overall card header height is 72px.

Screen Shot 2019-05-17 at 3 40 56 PM

joshuatf: @LevinMedia and I spoke about discrepancies in Card styles between Muriel and wc-admin.  Opened an issue [here](https://github.com/woocommerce/woocommerce-admin/issues/2259) to discuss further.

joshuatf: @jameskoster Messaged you in Slack about this as well, but the Figma designs for the Card component and the onboarding cards look to have this box-shadow and not the 1dp shadow.

Are those styles wrong or outdated?  Happy to tackle this in a follow-up as well if we don't have the answer now so we can unblock other issues that need this component.

jameskoster: I replied on Slack, but for anyone else following along here - I've updated the shadows in Figma to match the Material elevation styles as closely as possible. However, Figma doesn't account for the `spread` property of CSS drop-shadows, so I suggest using https://material-components.github.io/material-components-web-catalog/#/component/elevation for the actual code.

joshuatf: @jameskoster Thanks for the update.  Updated that box shadow here.  Could you review when you get a chance?

PR #1474, opened 2019-02-05
Format todo comments to match WC Core format

joshuatf:
Fixes #1427 

Updates all the `@todo` comments to be consistent with WC core formatting and make them more searchable.

### Detailed test instructions:

Nothing to test, but todo comments should all be formatted correctly.

PR #2253, opened 2019-05-17
Category Lookup Table - Fix Category Segments

### LINKED ISSUE #1915: Category Segments: Child categories included in totals, but not segments

mikejolley:
Closes #1915 

Second attempt at https://github.com/woocommerce/woocommerce-admin/pull/2247 with a different lookup table structure based on @joshuatf suggestion. I'll leave https://github.com/woocommerce/woocommerce-admin/pull/2247 open until this is reviewed.

Creates a category lookup table which stores term relationships for the full category hierarchy.

The table stores data about category hierarchy; the tree and all IDs within. Storing `depth` was unfortunately not straightforward and requires lots of PHP side processing to work out depth per term. I removed that after bashing my head against the desk a few hours.

Data is synced when a category is edited (and changes), and also on install.

Stats work by joining by category_id and then grouping by the tree_id. This join appears to solve our issues in #1915.

To test:

- Update, deactivate and reactivate. Cron job will build the lookup table in short time.
- View Analytics > Categories. Check the totals look good. 
- Try example tests cases which were broken in https://github.com/woocommerce/woocommerce-admin/issues/1915. Totals should include totals from children.


includes/class-wc-admin-loader.php
```diff
@@ -145,7 +163,7 @@ public static function load_features() {
 	/**
 	 * Registers a basic page handler for the app entry point.
 	 *
-	 * @todo The entry point for the embed needs moved to this class as well.
+	 * TODO: The entry point for the embed needs moved to this class as well.
```
Aljullu: In #1474 we converted all `TODO:` and similar comments to `@todo`. Any reason to change it back? :smile:

mikejolley: phpcs moans about them :)

joshuatf:
@mikejolley Does this need another review or on hold for the time being?

includes/class-wc-admin-install.php
```diff
@@ -265,6 +271,7 @@ public static function create_events() {
 		if ( ! wp_next_scheduled( 'wc_admin_daily' ) ) {
 			wp_schedule_event( time(), 'daily', 'wc_admin_daily' );
 		}
+		wp_schedule_single_event( time() + 10, 'generate_category_lookup_table' );
```
joshuatf: I think this should probably also be run during import.  We also may want to remove any scheduled events in the `on_deactivation()` hook.

mikejolley:
@joshuatf Feedback resolved. Is it ok to leave the sync on import in your hands?

rrennick:
This PR is ready for re-review.

psealock:
>Sounds good! Created a follow-up issue to tackle this here.

>I think this is good to merge once we fix failing tests. When I dump my category table during unit tests, it looks like the wc_category_lookup table is empty. Until we get the action scheduler sync in place, I think we might need to run the lookup hook on create_{$taxonomy}.

@joshuatf it seems wise to not release this until the issue you created #2765 is also included in the release. Am I reading this correctly in that we could potentially run into a scenario where the table doesn't exist but could an attempt to query could happen?

If so, we should add blocked status and tackle #2765 as soon as possible.

src/Loader.php
```diff
@@ -473,7 +491,7 @@ public static function inject_after_notices() {
 	 * Edits Admin title based on section of wc-admin.
 	 *
 	 * @param string $admin_title Modifies admin title.
-	 * @todo Can we do some URL rewriting so we can figure out which page they are on server side?
+	 * TODO: Can we do some URL rewriting so we can figure out which page they are on server side.
```
joshuatf: Nitpick: We should change this to our standardize format of `@todo`:
```suggestion
	@todo Can we do some URL rewriting so we can figure out which page they are on server side?
```

src/Loader.php
```diff
@@ -343,7 +361,7 @@ public static function is_admin_page() {
 	/**
 	 *  Returns true if we are on a "classic" (non JS app) powered admin page.
 	 *
-	 * @todo See usage in `admin.php`. This needs refactored and implemented properly in core.
+	 * TODO: See usage in `admin.php`. This needs refactored and implemented properly in core.
```
joshuatf: See `@todo` nitpick below.

tests/framework/helpers/class-wc-helper-reports.php
```diff
@@ -21,5 +21,6 @@ public static function reset_stats_dbs() {
 		$wpdb->query( "DELETE FROM $wpdb->prefix" . \Automattic\WooCommerce\Admin\API\Reports\Products\DataStore::TABLE_NAME ); // @codingStandardsIgnoreLine.
 		$wpdb->query( "DELETE FROM $wpdb->prefix" . \Automattic\WooCommerce\Admin\API\Reports\Coupons\DataStore::TABLE_NAME ); // @codingStandardsIgnoreLine.
 		$wpdb->query( "DELETE FROM $wpdb->prefix" . \Automattic\WooCommerce\Admin\API\Reports\Customers\DataStore::TABLE_NAME ); // @codingStandardsIgnoreLine.
+		\Automattic\WooCommerce\Admin\CategoryLookup::instance()->regenerate();
```
joshuatf: This works okay as a workaround for the tests, but I think we should still be adding entries to the lookup table when a new category is created.

When I create a new product category currently, no entry is added until I force a regeneration.

rrennick: The reason I needed to add the regenerate call was that the `uncategorized` category wasn't in the lookup table. When you create a new category it is added to the table. This unit test verifies that https://github.com/woocommerce/woocommerce-admin/blob/update/category-lookup-hierarchy/tests/api/reports-categories.php#L110:L129

joshuatf: Ah, that makes sense.

Since `regenerate()` looks like it's already being scheduled, should we just run all scheduled actions here instead of manually calling `regenerate()`?

rrennick: > Since regenerate() looks like it's already being scheduled, should we just run all scheduled actions here instead of manually calling regenerate()?

`regenerate` is scheduled in CRON vs AS. Running the pending actions doesn't generate the table.

rrennick:
> I think we might still want to get make sure new categories are synced with this table before merging.

That's done in https://github.com/woocommerce/woocommerce-admin/blob/update/category-lookup-hierarchy/src/CategoryLookup.php#L56

psealock:
@rrennick @joshuatf Is https://github.com/woocommerce/woocommerce-admin/issues/2765 still needed to be resolved before merging this one? Or can that issue be folded into the sync refactor we spoke about at the GM?

rrennick:
@psealock I think #2765 can be done separately. We discussed breaking the sync up into multiple classes. It would fit better with that.

psealock:
Thanks @rrennick, I'll just give this a last test and look through tomorrow before approving/merging. In the meantime, I reworked https://github.com/woocommerce/woocommerce-admin/issues/2765 to reflect a broader refactor of the `ReportsSync` class.

Issue #2254, opened 2019-05-17
Implement a Muriel Spinner Component


justinshreve:
See https://material.io/design/components/progress-indicators.html#circular-progress-indicators and the plugin installation screen in the Figma designs for more information.

`@wordpress/components` contains a spinner component, but it does not match the circular progress indicator guidelines. Newspack also does not have a component for this. We will need to implement our own spinner styles.

PR #2255, opened 2019-05-17
Add vertical and spinner options to the Stepper Component

justinshreve:
This PR cleans up a few stepper styles (the active step is meant to be bolder than the rest), and adds some additional properties to the `Stepper` component.

Specifically it adds a `direction` prop, allowing us to show the stepper as a vertical component. It also adds a prop that allows us to show a spinner for the current step. This is useful for steps that may be executing automatically, like installing plugins.

Please note, that this spinner should use Muriel/Material patterns, but currently uses core WP styles. See https://material.io/design/components/progress-indicators.html#circular-progress-indicators. This will be handled in a follow-up (See https://github.com/woocommerce/woocommerce-admin/issues/2254).

Screen Shot 2019-05-17 at 12 57 12 PM

To Test:
* Set `profileWizardComplete` to `false` in `client/dashboard/index.js`
* Visit `/wp-admin/admin.php?page=wc-admin#/?step=plugins`
* Note that you see a vertical stepper component, with a spinner step.
* Visit the devdocs and test the horizontal stepper

### DIFF
--- a/packages/components/src/stepper/index.js
+++ b/packages/components/src/stepper/index.js
@@ -5,6 +5,7 @@
 import classnames from 'classnames';
 import { Component, Fragment } from '@wordpress/element';
 import PropTypes from 'prop-types';
+import { Spinner } from '@wordpress/components';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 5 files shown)


packages/components/src/stepper/index.js
```diff
@@ -79,6 +88,21 @@ Stepper.propTypes = {
...
 		} )
 	).isRequired,
+
+	/**
+	 * Direction of the stepper.
+	 */
+	direction: PropTypes.oneOf( [ 'horizontal', 'vertical' ] ),
+
+	/**
+	 * Optional mark the current step as in pending to show a spinner treatment.
```
joshuatf: Minor typo:
```suggestion
	 * Optionally mark the current step as pending to show a spinner.
```

packages/components/src/stepper/index.js
```diff
@@ -16,28 +17,36 @@ import CheckIcon from './check-icon';
...
-                    } );
+					const stepClassName = classnames( 'woocommerce-stepper__step', {
+						'is-active': key === currentStep,
+						'is-complete': 'undefined' !== typeof isComplete ? isComplete : currentIndex > i,
+					} );
 
-                    return (
+					// @todo Update Spinner Styles
+					// https://material.io/design/components/progress-indicators.html
+					const icon = currentStep === key && isPending ? <Spinner /> : (
```
joshuatf: Nice 🚨

packages/components/src/stepper/style.scss
```diff
@@ -15,21 +15,24 @@
...
-			color: $muriel-active;
-
-			.woocommerce-stepper__step-icon {
+			.woocommerce-stepper__step-icon, .components-spinner {
 				background: $muriel-active;
 			}
 		}
 
+		&.is-active {
+			font-weight: 600;
```
joshuatf: Not a blocker, but the horizontal stepper does "bounce" a bit when the font-weight shifts from step to step:

Screen Shot 2019-05-20 at 6 18 06 PM

packages/components/src/stepper/index.js
justinshreve: Wow that original sentence. I blame Friday.

Issue #2244, opened 2019-05-15
Dashboard: REST-API driven chart selections that can be filtered


psealock:
Parent issue https://github.com/woocommerce/woocommerce-admin/issues/2009

>So currently the chart options shown in the Ellipsis menu in the charts section is powered by [this config](https://github.com/woocommerce/woocommerce-admin/blob/master/client/dashboard/dashboard-charts/config.js) which in-turn pulls from individual report configs like [this one for orders](https://github.com/woocommerce/woocommerce-admin/blob/master/client/analytics/report/orders/config.js#L14-L39).

>Looking back the charts config we started making after the playa meetup with the longer-term goal of this config being powered by the REST API, to allow for extension developers to expose/register their own chart types - akin to what has been built for leaderboards, and performance indicators.

>So really I think that issue could go quite deep depending on how far we want to take it - like pulling in the chart settings on each report page via the REST API ( long-term this sounds like a grand idea for extensibility ), but for the dashboard thinking some similar PRs to:

https://github.com/woocommerce/woocommerce-admin/pull/1995

https://github.com/woocommerce/woocommerce-admin/pull/2004

Issue #690, opened 2018-10-25
Component: Dashboard Charts


LevinMedia:
This description is lifted from the dashboard overview post at: p6riRB-3qG-p2

Primary interactions for the graphs (aggregation, graph type, and settings menu) have been moved into the block header, and their selections apply to all of the graphs that are active in the block. The block features a settings menu that allows users to show or hide individual graphs, change the color palette, or remove the block. 

<img width="1200" alt="screen shot 2018-08-09 at 4 25 09 pm" src="https://user-images.githubusercontent.com/4500952/47528134-3a80ab00-d859-11e8-8804-7285721a7f90.png">

Interactive features on individual graphs such as hover and tap states are disabled. Users are also unable to turn a series off or on in the legend. [note the legend shows a colored box without a check] Instead, the entire graph component is clickable, and directs users to the associated report.

<img width="1267" alt="screen shot 2018-08-09 at 6 34 26 pm" src="https://user-images.githubusercontent.com/4500952/47528217-6f8cfd80-d859-11e8-8b4f-49c64ff00755.png">

Missing from the image above would be a focus state, which would wrap an entire graph component. 

**Data available to graph: (as the reports become available)** 

*bold denotes default selection

Gross Revenue > Revenue report – Gross revenue tab selected
**Net Revenue > Revenue Report – Net Revenue tab selected
Orders > Order Report – Orders tab selected**
Average Order Value > Order Report – AOV tab selected
Items Sold > Product Report – Items Sold tab selected
Refunds > Orders Report – Refunds Tab selected Filtered by refunded orders
Discounted Orders > Coupons Report – Discounted Orders tab selected
Gross Discounted > Coupons Report – Gross Discounted tab selected
Total Tax > Tax Report > Total Tax tab selected
Order Tax > Tax Report > Order Tax tab selected
Shipping Tax > Tax Report > Shipping Tax tab selected
Shipping > Revenue Report – Shipping tab active
Downloads > Download Report
Low Stock > Stock Report – Filtered by low stock products
Out of Stock > Stock Report – Filtered by out of stock products
New customers > Orders report – Filtered by new customers
Returning customers > Orders Report  – filtered by returning customers

PR #2258, opened 2019-05-20
Dashboard: Extend charts data

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/2244

This PR adds filters for adding charts/summaryNumbers to existing reports as well as dashboard charts from reports added by extensions.

#2244 calls for a REST API driven approach similar to leaderboards and store performance indicators. I think a js approach is more suitable in this case because existing infrastructure is already responsible for making api calls and the dashboard simply designates which charts to display. In other words, a chart config is supplied and `<ReportChart />` handles the rest.

This logic makes the assumption that dashboard charts are only made up of charts currently found on existing reports or reports introduced via extensions. In either case, the current dashboard settings, which are persisted, can determine which are hidden and which are visible.

In a follow up PR, I can go through the process of adding a chart via extension and possibly expose any holes in this logic.

I also made a few changes while here:

1. Improved the `selectedChart` propType declaration
2. Remove the unused `charts` prop from `<ReportChart />`
3. Added the `downloads` chart to the dashboard and hid it by default. You'll need to delete the following user meta if you want to confirm its hidden by default:

```
DELETE FROM `wp_usermeta` WHERE meta_key = 'wc_admin_dashboard_sections'
```

Make sure the Dashboard and Reports load and function correctly. There should be no visible changes other than the added downloads chart on the dashboard.

### DIFF
--- a/client/dashboard/dashboard-charts/config.js
+++ b/client/dashboard/dashboard-charts/config.js
@@ -1,5 +1,10 @@
 /** @format */
 /**
+ * External dependencies
+ */
+import { applyFilters } from '@wordpress/hooks';
+
+/**
  * Internal dependencies
  */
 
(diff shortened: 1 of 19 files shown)


client/dashboard/dashboard-charts/config.js
```diff
@@ -9,16 +14,23 @@ import { charts as revenueCharts } from 'analytics/report/revenue/config';
...
-		couponsCharts.map( d => ( { ...d, endpoint: 'orders' } ) ),
-		taxesCharts.map( d => ( { ...d, endpoint: 'taxes' } ) )
-	);
+const allCharts = applyFilters(
+	DASHBOARD_CHARTS_FILTER,
+	ordersCharts
+		.map( d => ( { ...d, endpoint: 'orders' } ) )
+		.concat(
+			productsCharts.map( d => ( { ...d, endpoint: 'products' } ) ),
+			revenueCharts.map( d => ( { ...d, endpoint: 'revenue' } ) ),
```
Aljullu: Not related to this PR, but just saw it now. The _Revenue_ report is the first report in the menu but here we list its charts as it was the third one, which makes Dashboard charts order incoherent with reports order.

In #690 there seems to be the order we should use for Dashboard charts, but we are not following it. Should we open an issue to tackle that and get the input from @LevinMedia?

LevinMedia: @Aljullu 

>Should we open an issue to tackle that

Yes please 👍 And thank you!

psealock: Yeah, that makes sense. I'll just address that here before merging.

psealock: I did a basic re-order for this one because I can easily move groups, but opened a new issue to look at titles and order in depth, https://github.com/woocommerce/woocommerce-admin/issues/2266

PR #2263, opened 2019-05-20
include license.txt and update component package.json files

### LINKED ISSUE #2101: License mismatch between package.json and composer.json
In https://github.com/woocommerce/woocommerce-admin/blob/master/composer.json#L6 the license is `GPL-3.0-or-later`.

In https://github.com/woocommerce/woocommerce-admin/blob/master/package.json#L9 (and child packages' `package.json`) it is `GPL-2.0-or-later`.

rrennick:
Fixes #2101 

This is the second half of the license fixes. A followup to #2248

### DIFF
--- a/license.txt
+++ b/license.txt
@@ -1,0 +1,710 @@
+WooCommerce - eCommerce for WordPress
+
+Copyright 2015 by the contributors
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation; either version 3 of the License, or
+(at your option) any later version.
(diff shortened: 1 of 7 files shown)


timmyc:
Is this ready for :eyes: Ron?

rrennick:
> Is this ready for 👀 Ron?

Yes, label added.

rrennick:
> Should we remove LICENSE.md now that we added the .txt version?

I left it because existing installs of the plugin & the wordpress.org plugin page link to the file. We can drop it when we merge into core.

Issue #2268, opened 2019-05-20
Migrate core WC notices to WC Admin


jeffstieler:
High priority:
* `update` (but not `updating` or `updated`)
* `no_shipping_methods`
* `no_secure_connection`
* `template_files`

All other notices should end up in the Inbox panel.

jeffstieler:
This is blocked until we're going to merge to core.

psealock:
Is this still a priority @jeffstieler ?

jeffstieler:
Nope, closing.

PR #2272, opened 2019-05-21
Allow links in admin notice content.

jeffstieler:
In support of https://github.com/woocommerce/woocommerce-admin/issues/2268

Allow `<a>` tags in notice content with the following attributes: `href`, `rel`, `name`, `target`.

```php
$note = new WC_Admin_Note();
$note->set_title( 'Test Link' );
$note->set_content( 'Has a <a href="https://example.com">link</a>.' );
$note->save();
```

- Create a new test note with a link in the content
- Verify note is displayed with the link rendered properly

### DIFF
--- a/includes/notes/class-wc-admin-note.php
+++ b/includes/notes/class-wc-admin-note.php
@@ -347,6 +347,15 @@ class WC_Admin_Note extends WC_Data {
 			'br'     => array(),
 			'em'     => array(),
 			'strong' => array(),
+			'a'      => array(
+				'href'     => true,
+				'rel'      => true,
+				'name'     => true,
+				'target'   => true,
+				'download' => array(
+					'valueless' => 'y',
+				),
+			),
 		);
 
 		$content = wp_kses( $content, $allowed_html );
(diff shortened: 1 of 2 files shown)


includes/notes/class-wc-admin-note.php
```diff
@@ -347,6 +347,12 @@ public function set_content( $content ) {
 			'br'     => array(),
 			'em'     => array(),
 			'strong' => array(),
+			'a'      => array(
+				'href'   => true,
+				'rel'    => true,
+				'name'   => true,
+				'target' => true,
```
Aljullu: Do you think we could add the `download` attribute here? I can imagine some legit uses, and WordPress also allows it here: https://core.trac.wordpress.org/browser/trunk/src/wp-includes/kses.php#L63.

jeffstieler:
@Aljullu whoops! Nice catch. Fixed that and added the `download` attribute in e27e31b.

includes/notes/class-wc-admin-note.php
jeffstieler: Added in e27e31b.

PR #2280, opened 2019-05-22
Dashboard Extentions: Add a section

psealock:
Add a Dashboard Section via an extension example. The new section can modify its order, contents, or title and toggle visibility.

Screen Shot 2019-05-22 at 3 51 14 PM

### Testing instructions

1. `npm run example -- --ext=dashboard-section`
2. Activate the WooCommerce Admin Dashboard Section Example plugin
3. Go to the dashboard and make changes to the Apples section. Rename, move, toggle sections, or hide.
4. Make sure changes persist.

### DIFF
--- a/client/dashboard/store-performance/index.js
+++ b/client/dashboard/store-performance/index.js
@@ -8,7 +8,6 @@ import { compose } from '@wordpress/compose';
 import { withDispatch } from '@wordpress/data';
 import moment from 'moment';
 import { find } from 'lodash';
-import { TextControl } from '@wordpress/components';
 
 /**
  * WooCommerce dependencies
@@ -73,24 +72,16 @@ class StorePerformance extends Component {
 							);
 						} ) }
 						{ window.wcAdminFeatures[ 'analytics-dashboard/customizable' ] && (
-							<Fragment>
-								<div className="woocommerce-ellipsis-menu__item">
-									<TextControl
-										label={ __( 'Section Title', 'woocommerce-admin' ) }
-										onBlur={ onTitleBlur }
-										onChange={ onTitleChange }
-										required
-										value={ titleInput }
-									/>
-								</div>
-								<Controls
-									onToggle={ onToggle }
-									onMove={ onMove }
-									onRemove={ onRemove }
-									isFirst={ isFirst }
-									isLast={ isLast }
-								/>
-							</Fragment>
+							<Controls
+								onToggle={ onToggle }
+								onMove={ onMove }
+								onRemove={ onRemove }
+								isFirst={ isFirst }
+								isLast={ isLast }
+								onTitleBlur={ onTitleBlur }
+								onTitleChange={ onTitleChange }
+								titleInput={ titleInput }
+							/>
 						) }
 					</Fragment>
 				) }
--- a/docs/examples/extensions/dashboard-section/js/global-prices.js
+++ b/docs/examples/extensions/dashboard-section/js/global-prices.js
@@ -1,0 +1,43 @@
+/** @format */
+/**
+ * External dependencies
+ */
+import moment from 'moment';
+
+/**
+ * WooCommerce dependencies
+ */
+import { Card, Chart } from '@woocommerce/components';
+import { formatCurrency } from '@woocommerce/currency';
+
+const data = [];
+
+for ( let i = 1; i <= 20; i++ ) {
+	const date = moment().subtract( i, 'days' );
+	data.push( {
+		date: date.format( 'YYYY-MM-DDT00:00:00' ),
+		primary: {
+			label: 'Global Apple Prices, last 20 days',
+			labelDate: date.format( 'YYYY-MM-DD 00:00:00' ),
+			value: Math.floor( Math.random() * 100 ),
+		},
+	} );
+}
+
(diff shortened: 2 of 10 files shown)


client/dashboard/store-performance/index.js
```diff
@@ -201,7 +201,7 @@ export default compose(
 
 		if ( statKeys.length === 0 ) {
 			return {
-				hiddenBlocks,
+				hiddenIndicators: [],
```
Aljullu: We are using `hiddenBlocks` everywhere else in the file to keep it consistent with other sections, is there any reason to change it here?

client/wc-api/wc-api-spec.js
```diff
@@ -46,7 +46,6 @@ function createWcApiSpec() {
 			},
 			update( resourceNames, data ) {
 				return [
-					...imports.operations.update( resourceNames, data ),
```
Aljullu: I'm not sure to understand why this was removed.

docs/examples/extensions/dashboard-section/js/global-prices.js
```diff
@@ -0,0 +1,44 @@
...
+ * WooCommerce dependencies
+ */
+import { Card, Chart } from '@woocommerce/components';
+import { formatCurrency } from '@woocommerce/currency';
+
+const data = [];
+
+for ( let i = 1; i <= 20; i++ ) {
+	const today = moment().startOf( 'day' );
+	const date = today.subtract( i, 'days' );
```
Aljullu: Not a big deal, but considering we are setting the hour to 00:00:00 below, there is no need to use `startOf( 'day' )` here. I think you can directly initialize the date like this:
```suggestion
	const date = moment().subtract( i, 'days' );
```

docs/examples/extensions/dashboard-section/js/index.js
```diff
@@ -0,0 +1,133 @@
...
+						<div className="woocommerce-ellipsis-menu__item">
+							<TextControl
+								label={ __( 'Section Title', 'woocommerce-admin' ) }
+								onBlur={ onTitleBlur }
+								onChange={ onTitleChange }
+								required
+								value={ titleInput }
+							/>
+						</div>
+						<Controls
```
Aljullu: Looking at this again... I wonder if it would make sense to move the title text input inside `<Controls>` so extension developers don't need to implement it. What do you think @psealock?

client/wc-api/wc-api-spec.js
Aljullu: Oh, just saw #2271.

client/dashboard/store-performance/index.js
psealock: Oops, that was a left over change from debugging. Fixed.

docs/examples/extensions/dashboard-section/js/global-prices.js
psealock: Indeed, good call.

docs/examples/extensions/dashboard-section/js/index.js
psealock: I like that idea!

psealock: Done in https://github.com/woocommerce/woocommerce-admin/pull/2280/commits/d8dd86f501fa7c652c0758faed84318c1adecfa2

docs/examples/extensions/dashboard-section/js/global-prices.js
```diff
@@ -0,0 +1,43 @@
...
+			label: 'Global Apple Prices, last 20 days',
+			labelDate: date.format( 'YYYY-MM-DD 00:00:00' ),
+			value: Math.floor( Math.random() * 100 ),
+		},
+	} );
+}
+
+const GlobalPrices = () => {
+	return (
+		<Card className="woocommerce-dashboard__chart-block" title="Global Apple Prices">
```
Aljullu: This chart looks different from the other ones:

image

It seems to be because it doesn't have the `woocommerce-analytics__card` class, was that in purpose?
```suggestion
		<Card className="woocommerce-dashboard__chart-block woocommerce-analytics__card" title="Global Apple Prices">
```

psealock: Good one, looks way better now.

docs/examples/extensions/dashboard-section/js/global-prices.js
```diff
@@ -0,0 +1,43 @@
...
+import { formatCurrency } from '@woocommerce/currency';
+
+const data = [];
+
+for ( let i = 1; i <= 20; i++ ) {
+	const date = moment().subtract( i, 'days' );
+	data.push( {
+		date: date.format( 'YYYY-MM-DDT00:00:00' ),
+		primary: {
+			label: 'Global Apple Prices, last 20 days',
```
timmyc: 😆 you love your Apples Paul!

timmyc:
@psealock this is rad. I did hit one little glitch that when I have the 'Add Report Example' activated alongside the Dashboard section example, and neither work, i.e. I don't see the new apple section

image

Once I deactivated the "WooCommerce Admin Add Report Example", I could then see the Apple section

Issue #2281, opened 2019-05-22
WC API should allow updating any WC Settings


joshuatf:
The current implementation of `wc-api/settings` limits updates to fields within the `wc_admin` group and specific fields:

```js
	const settingsFields = [
		'woocommerce_excluded_report_order_statuses',
		'woocommerce_actionable_order_statuses',
	];
```

Allowing a value to be passed for the group would allow this API to accept and modify properties across all WC setting groups.

This would be useful for simplifying the API calls needed in the onboarding profiler and would be a useful interface if merged into core.

**Alternative**

Alternatively, we can manually POST to various API routes using `apiFetch` but this implementation does not feel as clean or provide any level of abstraction across components.

PR #2283, opened 2019-05-22
Onboarding - Add shared step actions and step 1 actions

### LINKED ISSUE #2282: Onboarding - Step 1 shared actions and API calls
This step contains the following actions.

* Updating the query path to continue to the next step.
* Updating the profile params with a `skipped` boolean.
* Updating the usage tracking boolean in the options table or WC settings. (Will probably be covered by #2281).

The first 2 items above probably share some common logic between steps and can be passed down as props from a parent.

joshuatf:
Fixes #2282

Passes reusable shared actions as props to step components. `goToNextStep()` and `updateProfile()` are made available as props inside steps.

Note that updating the WC settings (not the profiler settings - https://github.com/woocommerce/woocommerce-admin/issues/2281) and closing the wizard will be handled in follow-up PRs.

### Screenshots
Screen Shot 2019-05-22 at 4 32 39 PM

1.  Change to `const profileWizardComplete = false;` in `dashboard/index.js`
2.  Go to the dashboard `/wp-admin/admin.php?page=wc-admin#`
3.  Click on "Get Started" and make sure this takes you to the next step.
4.  Go back and click on "Proceed without Jetpack or WooCommerce Services" and make sure this updates the profiler setting `skipped` to `true` by sending a GET request to `wp-json/wc-admin/v1/onboarding/profile`.

### DIFF
--- a/client/dashboard/profile-wizard/index.js
+++ b/client/dashboard/profile-wizard/index.js
@@ -11,7 +11,7 @@ import { withDispatch } from '@wordpress/data';
 /**
  * WooCommerce dependencies
  */
-import { stringifyQuery, updateQueryString } from '@woocommerce/navigation';
+import { updateQueryString } from '@woocommerce/navigation';
 
 /**
  * Internal depdencies
@@ -102,11 +102,11 @@ class ProfileWizard extends Component {
 
 	updateProfile( params ) {
 		const { addNotice } = this.props;
-		const payload = stringifyQuery( params );
 
 		return apiFetch( {
-			path: `/wc-admin/v1/onboarding/profile${ payload }`,
+			path: '/wc-admin/v1/onboarding/profile',
 			method: 'POST',
+			data: params,
 		} ).catch( error => {
 			if ( error && error.message ) {
 				addNotice( { status: 'error', message: error.message } );
--- a/client/dashboard/profile-wizard/index.js
+++ b/client/dashboard/profile-wizard/index.js
@@ -2,13 +2,16 @@
 /**
  * External dependencies
  */
+import apiFetch from '@wordpress/api-fetch';
 import { Component, createElement, Fragment } from '@wordpress/element';
+import { compose } from '@wordpress/compose';
 import { pick } from 'lodash';
+import { withDispatch } from '@wordpress/data';
 
 /**
  * WooCommerce dependencies
  */
-import { updateQueryString } from '@woocommerce/navigation';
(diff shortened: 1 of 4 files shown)


client/dashboard/profile-wizard/index.js
```diff
@@ -48,10 +62,51 @@ export default class ProfileWizard extends Component {
...
+
+		return updateQueryString( { step: nextStep.key } );
+	}
+
+	updateProfile( params ) {
+		const { addNotice } = this.props;
+		const payload = stringifyQuery( params );
+
+		return apiFetch( {
+			path: `/wc-admin/v1/onboarding/profile${ payload }`,
```
justinshreve: I doubt we would hit too long of a URL with the number of fields per page, but I'm still thinking we should update the controller to accept POST fields for the data instead of query strings.

Instead of `prepare_objects_query` in the REST controller, I think we could use `get_json_params`, which would allow us to pass the payload in the body.

client/dashboard/profile-wizard/index.js
```diff
@@ -48,10 +62,51 @@ export default class ProfileWizard extends Component {
...
+		}
+
+		return updateQueryString( { step: nextStep.key } );
+	}
+
+	updateProfile( params ) {
+		const { addNotice } = this.props;
+		const payload = stringifyQuery( params );
+
+		return apiFetch( {
```
justinshreve: Should we also convert this to a proper `wc-api` set of selectors/mutator actions? We'll want to be able to rely on flags like 'completed' and 'skipped' to hide the profiler from a GET call, so having state reflected by updates would be good. See https://github.com/woocommerce/woocommerce-admin/blob/master/client/dashboard/index.js#L25.

client/dashboard/profile-wizard/index.js
joshuatf: Sure!  Updated this to use `get_json_params`.  Still passed it through the existing `prepare_objects_query` so we can filter params.

I think this will also resolve #2299.

client/dashboard/profile-wizard/index.js
joshuatf: Absolutely- good idea!  Created an issue around that [here](https://github.com/woocommerce/woocommerce-admin/issues/2303).

client/dashboard/profile-wizard/index.js
justinshreve: Thanks for updating this! Confirmed this resolves #2299 by testing with this branch + Postman directly.

PR #2289, opened 2019-05-22
Add Muriel Styled Spinner

### LINKED ISSUE #2254: Implement a Muriel Spinner Component

justinshreve:
Fixes #2254.

This PR implements a Material styled progress indicator, to match the spinner seen in the Figma designs. See https://material.io/design/components/progress-indicators.html#circular-progress-indicators. It implements an indeterminate circular progress indicator.

### Accessibility

@ryelle Any suggestions on what to do here if `prefers-reduced-motion` is enabled? Just not do an animation? (I can address this in a follow-up PR if needed).

### Screenshots

Screen Shot 2019-05-22 at 4 28 17 PM

* Visit the devdocs, and look at the `<Spinner />` component. I've also added a button to toggle the progress indicator in the `<Stepper />` components.
* You can also test the plugins install step to see it there.


client/devdocs/examples.json
```diff
@@ -23,6 +23,7 @@
   { "component": "SegmentedSelection" },
   { "component": "SplitButton" },
   { "component": "Stepper" },
+  { "component": "Spinner" },
```
joshuatf: Minor nitpick: can we move this up 2 lines to alphabetize?

packages/components/src/stepper/example.md
```diff
@@ -3,8 +3,9 @@ import { Stepper } from '@woocommerce/components';
 
 const MyStepper = withState( {
     currentStep: 'first',
```
joshuatf: Seems like there's a difference in spacing/tabs here that I may have introduced.  Do you mind reindenting this while you're here?

packages/components/src/stepper/style.scss
```diff
@@ -24,11 +25,16 @@
...
+			margin-right: $gap-smallest;
+		}
+
 		&.is-active,
 		&.is-complete {
-			.woocommerce-stepper__step-icon,
-			.components-spinner {
-				background: $muriel-hot-orange-700;
+			.woocommerce-stepper__step-icon {
+				background: $muriel-gray-900;
```
joshuatf: Should we override this color in `profile-wizard/style.scss` so we can keep `$muriel-hot-orange-700` or should the Muriel component use this color by default?

justinshreve: Just double checked the Components Figma, and it actually looks like this should be `$muriel-hot-blue-500` by default -- so I'll update that, and override with the gray in `profile-wizard/style.scss`.

timmyc:
Really cool Justin, noticed this spinner in your video demo you posted earlier ✨

ryelle:
> curious to get @ryelle's thoughts on `prefers-reduced-motion` for the spinner.

We don't actually have to remove all motion with `prefers-reduced-motion`, just "problematic" motion, which is either

1. The animation takes up the majority of the screen, or changes how scroll behavior works (parallax), this could trigger vertigo-like issues, or
2. The animation is a constant loop that distracts from other elements on the page

(basically. there are some other considerations around flashing + seizures, but that doesn't apply here)

Since this is not a full-page spinner (😱)  and it's (hopefully) not continuing to spin while you need to do other things, it should be fine as-is 👍🏻

justinshreve:
Thank you!

Issue #2250, opened 2019-05-16
Settings: Add ability to set default date range selection


timmyc:
**Is your feature request related to a problem? Please describe.**

Requested via [the forums](https://wordpress.org/support/topic/request-save-date-range/):

> I would like to save the date range.
For me it’s most interesting to see everything YTD compared to year before.
Everytime I open the dashboard, I need to change it….. I would like to save this setting.

**Describe the solution you'd like**

Seems like either introducing a setting to set the default date range, or persist the one that is selected on the dashboard as the default for there in user meta would work ( I swear we had this at one point! ).

rrennick:
>  I swear we had this at one point!

#1969

timmyc:
Yeah kind of related, but more what I'm thinking of is we used to persist the last date range selected in user meta. Like it was one of the first things we were saving session to session. I could very well be imagining this though.

PR #2292, opened 2019-05-23
Settings: Add default date settings

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/2250

Persist a default date setting of the user's choosing. 

Screen Shot 2019-05-24 at 11 04 24 AM

1. Analytics > Settings.
2. Select a new date range > Update > Save Changes.
3. Navigate to a report and see the date range applied by default.

### DIFF
--- a/client/analytics/settings/config.js
+++ b/client/analytics/settings/config.js
@@ -15,8 +15,10 @@ import { Link } from '@woocommerce/components';
  * Internal dependencies
  */
 import { DEFAULT_ACTIONABLE_STATUSES } from 'wc-api/constants';
+import DefaultDate from './default-date';
 
 const SETTINGS_FILTER = 'woocommerce_admin_analytics_settings';
+const DEFAUTL_DATE_RANGE = 'period=month&compare=previous_year';
 
 const defaultOrderStatuses = [
 	'completed',
@@ -67,7 +69,8 @@ export const analyticsSettings = applyFilters( SETTINGS_FILTER, [
 				moreLink: <Link href="#" type="external" />, // @todo This needs to be replaced with a real link.
 			},
 		} ),
-		initialValue: wcSettings.wcAdminSettings.woocommerce_excluded_report_order_statuses || [],
+		initialValue:
+			[ ...wcSettings.wcAdminSettings.woocommerce_excluded_report_order_statuses ] || [],
 		defaultValue: [ 'pending', 'cancelled', 'failed' ],
 	},
 	{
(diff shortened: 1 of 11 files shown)


client/analytics/settings/index.js
```diff
@@ -87,7 +87,23 @@ class Settings extends Component {
 		}
 	}
 
+	/**
+	 * Ensure changes are reflected to parameters on the window as well as
+	 * the config for construction of this component when re-navigating to
+	 * the settings page.
```
psealock: Previously, these values updated automatically for objects and arrays due to object equality. I think thats the name, but this is what I mean:

```js
var a = { value: 1 };
var b = a;
b.value = 99;
console.log( a.value ); // 99
```
This won't work for strings, so values need to be updated so that navigating away from the Settings page after saving reflects changes.

Aljullu: Yes, this is kinda confusing. :confused:  I wonder if we should clone all `wcSettings.wcAdminSettings` properties we use in `settings/config.js` so we keep them immutable.

So from:
```ES6
initialValue: wcSettings.wcAdminSettings.woocommerce_excluded_report_order_statuses || [],
```
to:
```ES6
initialValue: [ ...wcSettings.wcAdminSettings.woocommerce_excluded_report_order_statuses ] || [],
```
@psealock what would you think about this? :point_up_2: It doesn't fix any bug but I think knowing `wcSettings.wcAdminSettings` is ONLY updated in `persistChanges` will make our lives easier in the future.

client/analytics/settings/config.js
```diff
@@ -93,4 +95,17 @@ export const analyticsSettings = applyFilters( SETTINGS_FILTER, [
 		initialValue: wcSettings.wcAdminSettings.woocommerce_actionable_order_statuses || [],
 		defaultValue: DEFAULT_ACTIONABLE_STATUSES,
 	},
+	{
+		name: 'woocommerce_default_date_range',
+		label: __( 'Default Date Range:', 'woocommerce-admin' ),
+		inputType: 'component',
+		component: DefaultDate,
+		helpText: __(
+			'Select a default date range. When no range is selected, reports will be viewed by' +
```
timmyc: missing a space between `bythe` here, you can see it in the screenshot above.

client/analytics/settings/index.js
psealock: Good idea, this sets a pattern and lets you know the setting on the window is immutable.

packages/date/src/index.js
```diff
@@ -268,11 +265,26 @@ function getDateValue( period, compare, after, before ) {
...
+	const queryDefaults = parse( defaultDateRange.replace( /&amp;/g, '&' ) );
+
+	if ( period && compare ) {
+		return {
+			period,
+			compare,
+			after: after ? moment( after ) : null,
+			before: before ? moment( before ) : null,
+		};
+	}
```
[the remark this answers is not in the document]
psealock: good point, done.

PR #2293, opened 2019-05-23
Update settings wc-api to accept multiple setting groups

### LINKED ISSUE #2281: WC API should allow updating any WC Settings

joshuatf:
Fixes #2281 

Adds the ability to fetch and update all WC settings other than those in the `wc_admin` group.

1.  Go to the Analytics settings page.
2.  Make sure the correct settings are shown.
3.  Save the data.
4.  Make sure data is persisted on refresh.

### DIFF
--- a/client/wc-api/settings/operations.js
+++ b/client/wc-api/settings/operations.js
@@ -4,7 +4,6 @@
  * External dependencies
  */
 import apiFetch from '@wordpress/api-fetch';
-import { pick } from 'lodash';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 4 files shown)


client/wc-api/settings/operations.js
```diff
@@ -20,57 +19,61 @@ function update( resourceNames, data, fetch = apiFetch ) {
...
-	return { [ 'settings' ]: { data: settings } };
+function settingToSettingsResource( resourceName, data ) {
+	const settings = {};
+	if ( 'undefined' === typeof data.update ) {
+		return '';
+	}
+
+	// @todo This will only return updated fields so fields
+	// not updated may be temporarily overwritten in the store.
+	data.update.forEach( setting => ( settings[ setting.id ] = setting.value ) );
```
joshuatf: While this doesn't affect wc-admin currently, it doesn't feel right to overwrite the existing store with partial data.

I'm not aware of any way to grab existing data from here and merge, but that would be ideal.  Open to other thoughts here or following up in another PR.

justinshreve: Hmm, I don't know of a way to do that here either. Maybe @coderkevin has some ideas?

emdashcodes (approved):
This is testing well for me with the existing analytics settings.

I think we can merge/move forward with this, but maybe we could connect with @coderkevin and his fresh-data knowledge to see if we can fix up that `@todo` in another PR.

PR #2296, opened 2019-05-23
Update Profile Wizard Header to Display Stepper

justinshreve:
This PR does a bit of refactoring to the main profile wizard controller and header component. It allows us to easily define labels for the steps, and display the correct step in the wizard header. To also updates the background header color to white to match the Figma designs.

To Test:
* Go to the `start` or `plugins` step and see that the header area is white, and contains the Jetpack & Plugins logos.
* Go to a page like `/?step=business-details` and note the stepper.
* Shrink your browser to mobile size. The stepper / header should display at the bottom per the designs.
* Spot check analytics on mobile, since I made a CSS adjustment. Things should still look the same.

Desktop:

Screen Shot 2019-05-23 at 10 13 24 AM

Mobile:

Screen Shot 2019-05-23 at 10 13 45 AM

### DIFF
--- a/client/dashboard/profile-wizard/index.js
+++ b/client/dashboard/profile-wizard/index.js
@@ -2,18 +2,20 @@
 /**
  * External dependencies
  */
-import { Component, createElement } from '@wordpress/element';
+import { Component, createElement, Fragment } from '@wordpress/element';
+import { pick } from 'lodash';
 
 /**
  * Internal depdencies
  */
+import ProfileWizardHeader from './header';
 import Plugins from './steps/plugins';
-import Start from './steps/start/';
+import Start from './steps/start';
 import './style.scss';
+import { __ } from '@wordpress/i18n';
 
 const getSteps = () => {
 	const steps = [];
-
 	steps.push( {
 		key: 'start',
 		container: Start,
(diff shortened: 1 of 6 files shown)


client/dashboard/profile-wizard/style.scss
```diff
@@ -140,3 +141,26 @@
 		}
 	}
 }
+
+.woocommerce-profile-wizard__header.is-stepper {
+	display: block;
```
joshuatf: This removes the vertical centering for the stepper.  I think the intention behind this was to get the stepper to full width to show the divider lines in which case we could use this as a substitute:

```scss
.woocommerce-stepper {
    width: 100%;
}
```

client/dashboard/profile-wizard/index.js
```diff
@@ -22,7 +24,31 @@ const getSteps = () => {
 		key: 'plugins',
 		container: Plugins,
 	} );
-
+	steps.push( {
+		key: 'store-details',
+		container: Fragment,
+		label: __( 'Store Details', 'woocommerce-admin' ),
```
joshuatf: Like the idea of using the labels to determine the stepper 👍

client/stylesheets/shared/_reset.scss
```diff
@@ -17,10 +17,13 @@
...
 			right: -14.5em;
 		}
-		#wpcontent,
-		#wpbody-content {
+		#wpcontent {
 			min-height: calc(100vh - #{$adminbar-height-mobile});
 		}
+
+		#wpbody-content {
+			min-height: 100vh;
```
joshuatf: We probably only want to apply this on onboarding pages.  Most pages still having a padding on `#wpbody` that will make this longer than the viewport with no content.

client/dashboard/profile-wizard/style.scss
```diff
@@ -140,3 +141,26 @@
...
+
+	svg > g {
+		transform: initial;
+	}
+
+	@include breakpoint( '<782px' ) {
+		position: absolute;
```
joshuatf: Should this be fixed to the bottom of the screen or sitting at the bottom of the page in mobile?  Most page content will push this below the fold, but not sure if a fixed item on the screen will affect usability. /cc @jameskoster

justinshreve:
I handled the CSS feedback in https://github.com/woocommerce/woocommerce-admin/pull/2296/commits/8802e786398c30daed24491cf2b5d18e7ec9e43d.

I've left the menu fixed to the bottom of the screen for now, pending @jameskoster's feedback. I can adjust this in a follow-up PR if necessary.

Issue #2267, opened 2019-05-20
get_current_page() doesn't work for registerted JS pages


DanReyLop:
`WC_Admin_Page_Controller::get_current_page()` can't be used for fragment-based pages.

The problem is that the fragment part of the URL (everything that comes after the `#`) is **never** sent to the server. So, the PHP logic [here](https://github.com/woocommerce/woocommerce-admin/blob/22b2d7f846fd6cd1b3adcbbbebd08d5136c2b768/includes/page-controller/class-wc-admin-page-controller.php#L107) will never be able to correctly identify a "registered" page. It still works for traditional "connected" pages that don't rely on fragment URLs.

Looks like this is related with [this `@todo`](https://github.com/woocommerce/woocommerce-admin/blob/master/includes/class-wc-admin-loader.php#L468) too.

At the very least, there should be a caveat in that function's documentation, making it clear that it won't work for registered fragment-based pages, and remove the `PHP_URL_FRAGMENT` check since it will never work.

Additionally, what is the reason to use fragment-based URLs? As far as I know, using [HTML5's History API](https://caniuse.com/#feat=history) should allow us to use URLs `wc-admin?something=something` without refreshing the browser.

cc/ @jeffstieler since you added this on https://github.com/woocommerce/woocommerce-admin/pull/2209

jeffstieler:
Ah, yeah I completely missed this. It appears to work because the Dashboard page is registered without anything in the fragment. All child pages are then errantly identified as the parent..

justinshreve:
I did a little exploration on this today (related to https://github.com/woocommerce/woocommerce-admin/pull/2300) and I think we could switch to HTML5's History API, and update the URL structure so that we can actually detect the pages server side, instead of using hash.

https://github.com/woocommerce/woocommerce-admin/compare/try/page-router-update contains a start to this. You should be able to test by going directly to `/wp-admin/admin.php/woocommerce/analytics/products`. There is still a number of things broken like the dashboard, and how the history is updated, but with a little more work this could be cleaned up.

DanReyLop:
> Please excuse the rough/messy try branch.

Messiness in a try/ branch is **always** excused :)

[...]

PR #2300, opened 2019-05-23
Onboarding: Add profiler body class on initial load + respect skipped/completed flags

justinshreve:
This PR adds the `woocommerce-profile-wizard__body` body class on initial load of the PHP app if we know that the profiler will display and if the user is on the main JS page.

This allows us to prevent the initial flash of the sidebar / masterbar (which makes for a smoother transition between Calypso and `wc-admin`).

It also adds a (temporary) PHP flag so that we can enable/disable the profiler, without editing `dashboard/index.js`. (I don't know about you Josh, but it was getting annoying to switch branches, stash the change, etc 😆). In addition to that, I've added some logic so the profiler will hide/show based on the skipped/complete flags.

To Test:
* Set `WOOCOMMERCE_ADMIN_DEV_SHOW_PROFILER` to `true` wherever you keep your development snippets / constants.
* Make sure `skipped` and `completed` are both false via `/wc-admin/v1/onboarding/profile`
* Visit `/wp-admin/admin.php?page=wc-admin#/`. Notice that the admin menus do not display, and the profiler loads in (Another good example of this is running through the plugins flow and seeing the transition between Calypso).
* Test setting `skipped` or `completed` to true, and refresh the page. Note that the profiler does not show.
* Test visiting an analytics route, and make sure the admin bar loads in correctly.

### DIFF
--- a/client/layout/index.js
+++ b/client/layout/index.js
@@ -27,6 +27,12 @@ import StoreAlerts from './store-alerts';
 class Layout extends Component {
 	componentDidMount() {
 		this.recordPageViewTrack();
+
+		const path = this.getPath();
+
+		if ( 'dashboard' !== path ) {
+			document.body.classList.remove( 'woocommerce-profile-wizard__body' );
+		}
 	}
 
 	componentDidUpdate( prevProps ) {
@@ -42,7 +48,7 @@ class Layout extends Component {
 		}
 	}
 
-	recordPageViewTrack() {
+	getPath() {
 		const pathname = get( this.props, 'location.pathname' );
 		if ( ! pathname ) {
 			return;
@@ -56,7 +62,11 @@ class Layout extends Component {
 			path = 'dashboard';
 		}
 
-		recordPageView( path );
+		return path;
+	}
+
+	recordPageViewTrack() {
+		recordPageView( this.getPath() );
 	}
 
 	render() {
--- a/client/layout/index.js
+++ b/client/layout/index.js
@@ -27,12 +27,7 @@ import StoreAlerts from './store-alerts';
 class Layout extends Component {
 	componentDidMount() {
 		this.recordPageViewTrack();
-
-		const path = this.getPath();
-
-		if ( 'dashboard' !== path ) {
-			document.body.classList.remove( 'woocommerce-profile-wizard__body' );
-		}
+		document.body.classList.remove( 'woocommerce-admin-is-loading' );
 	}
 
 	componentDidUpdate( prevProps ) {
@@ -48,7 +43,7 @@ class Layout extends Component {
 		}
 	}
 
-	getPath() {
+	recordPageViewTrack() {
 		const pathname = get( this.props, 'location.pathname' );
 		if ( ! pathname ) {
 			return;
@@ -62,11 +57,7 @@ class Layout extends Component {
 			path = 'dashboard';
 		}
 
-		return path;
-	}
-
-	recordPageViewTrack() {
-		recordPageView( this.getPath() );
+		recordPageView( path );
 	}
 
 	render() {
--- a/includes/api/class-wc-admin-rest-onboarding-plugins-controller.php
+++ b/includes/api/class-wc-admin-rest-onboarding-plugins-controller.php
@@ -206,13 +206,14 @@ class WC_Admin_REST_Onboarding_Plugins_Controller extends WC_REST_Data_Controlle
 			return new WP_Error( 'woocommerce_rest_jetpack_not_active', __( 'Jetpack is not installed or active.', 'woocommerce-admin' ), 404 );
 		}
 
-		$redirect_url = esc_url_raw(
+		$next_step_slug = 'store-details';
+		$redirect_url   = esc_url_raw(
 			add_query_arg(
 				array(
 					'page' => 'wc-admin',
 				),
 				admin_url( 'admin.php' )
-			) . '#/?step=details'
+			) . '#/?step=' . $next_step_slug
 		);
 
 		$connect_url = Jetpack::init()->build_connect_url( true, $redirect_url, 'woocommerce-setup-wizard' );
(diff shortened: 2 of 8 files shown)


includes/features/onboarding/class-wc-admin-onboarding.php
```diff
@@ -26,6 +26,65 @@ public static function get_instance() {
...
+		return $settings;
+	}
+
+	/**
+	 * Adds body classes allowing us hide wp-admin ahead of time, if we know the profiler wizard is going to show.
+	 *
+	 * @param string $admin_body_class Body class to add.
+	 */
+	public function add_admin_body_classes( $admin_body_class = '' ) {
+		global $hook_suffix;
```
timmyc: why is this global needed?

justinshreve: Oops, good catch. It was a copy and paste from the body class function in `WC_Admin_Loader`. I'll remove it.

includes/features/onboarding/class-wc-admin-onboarding.php
```diff
@@ -26,6 +26,65 @@ public static function get_instance() {
...
+	}
+
+	/**
+	 * Returns true if the profiler should be displayed (not completed and not skipped).
+	 *
+	 * @return bool
+	 */
+	public function should_show_profiler() {
+		// @todo Remove this once we have a proper way to dismiss the profiler.
+		if ( ! defined( 'WOOCOMMERCE_ADMIN_DEV_SHOW_PROFILER' ) || false === WOOCOMMERCE_ADMIN_DEV_SHOW_PROFILER ) {
```
joshuatf: Yesss!  💯

includes/features/onboarding/class-wc-admin-onboarding.php
```diff
@@ -26,6 +26,65 @@ public static function get_instance() {
...
+			return false;
+		}
+		$onboarding_data = get_option( 'wc_onboarding_profile', array() );
+
+		// @todo Update this to compare to proper bools (https://github.com/woocommerce/woocommerce-admin/issues/2299).
+		$is_completed = isset( $onboarding_data['completed'] ) && 'true' === $onboarding_data['completed'];
+		$is_skipped   = isset( $onboarding_data['skipped'] ) && 'true' === $onboarding_data['skipped'];
+		return $is_completed || $is_skipped ? false : true;
```
joshuatf: What happens if a user has already gone through the default WC setup wizard?  Do they get hit with this again or do we want to add a note to change this before merging into core?

includes/api/class-wc-admin-rest-onboarding-plugins-controller.php
```diff
@@ -206,13 +206,14 @@ public function connect_jetpack() {
 			return new WP_Error( 'woocommerce_rest_jetpack_not_active', __( 'Jetpack is not installed or active.', 'woocommerce-admin' ), 404 );
 		}
 
-		$redirect_url = esc_url_raw(
+		$next_step_slug = 'store-details';
```
joshuatf: Side note: I know we talked a bit about extensibility before.  Would we ever want to allow hooking into change this step?

client/layout/index.js
```diff
@@ -27,6 +27,12 @@ import StoreAlerts from './store-alerts';
 class Layout extends Component {
 	componentDidMount() {
 		this.recordPageViewTrack();
+
+		const path = this.getPath();
+
+		if ( 'dashboard' !== path ) {
+			document.body.classList.remove( 'woocommerce-profile-wizard__body' );
```
joshuatf: Would it be possible to add a check in `add_admin_body_classes()` to see if we're in the dashboard page before adding this class?  That way we can remove these 3 lines and keep this components scope down.

justinshreve: I have a check in there that checks if we are on a JS page, but the problem is we can't tell which wc-admin route they on because the route is stored in a JS hash that is not passed down to the server.

I did a little exploration to see what we can do, but switching to another routing method / URL structure would be non-trivial. I think we might have to stick with this for now. Does that seem OK?

justinshreve: https://github.com/woocommerce/woocommerce-admin/issues/2267

includes/features/onboarding/class-wc-admin-onboarding.php
justinshreve: Good question. We should probably respect the current flag.

client/layout/index.js
joshuatf: Ya, that seems very reasonable and thanks for your research around the issue!  If we make those routing changes you suggested at a later date, we can revisit this.

includes/features/onboarding/class-wc-admin-onboarding.php
justinshreve: There actually isn't a flag for this exactly. The redirect and notice both just rely on the presence of the `install` notice showing. 

https://github.com/woocommerce/woocommerce/blob/af574f6484a4300d1ec92c569593013f2e2fced2/includes/admin/class-wc-admin.php#L161-L164

https://github.com/woocommerce/woocommerce/blob/d0491072e8a7d0a4535db0ec2bae4a6d43ff64fd/includes/admin/views/html-notice-install.php

I think instead of relying on the old notice, we should probably add a note and/or an issue here on GitHub with the label 'On Core Merge'.

I'm thinking when we merge onboarding to core, we could set the completed flag to true during the upgrade process.

client/layout/index.js
justinshreve: I ended up generalizing this a bit more in the meantime. I now have it adding a `woocommerce-admin-is-loading` class.

I'm still only appending it when necessary, but this allows us to do something similar on other routes in the future (like future Woo App pages). It also allows me to drop the `dashboard` check here. We'll just always remove the class once `<Layout />` has loaded, and the  profiler will append its own profiler class to continue hiding elements.

justinshreve:
Thanks for the review! I've addressed some of the feedback and added some todos/notes for the future. https://github.com/woocommerce/woocommerce-admin/pull/2300/commits/9f636401902a144b723b5074508feb0ab5ddb9d5

includes/features/onboarding/class-wc-admin-onboarding.php
joshuatf: Sounds like a great solution to me!

PR #2304, opened 2019-05-24
Add onboarding step 4

### LINKED ISSUE #1888: Onboarding - Profile Wizard: Step 4 – Store Address
In this step of the wizard, we will collect the store address and email address. Form elements should rely on Gutenberg form components with Muriel styling.

Data should be saved via the REST API added in #1883. These values should also be saved to their relevant WC settings via the WC settings API. We should be able to populate the country/state dropdown with the `/data/countries` endpoint.

store-details-2

store-details-4

joshuatf:
Fixes #1888 

Adds the onboarding step 4.  Design review of this is blocked until rebased from #2296.

### Questions
* There doesn't appear to be a default admin email.  Should the email here overwrite the WordPress `admin_email` (seems risky), set to `woocommerce_stock_email_recipient` which is used for notifications, or save separately under profiler data?
* I've added "Address Line 2" but I believe a city will need to be added as well to have a complete address.  Should we add that as well or remove the 2nd line address?
* Took a bit of liberty on adding an error message to the email field for invalid emails.  This probably needs design review.

### Screenshots
Screen Shot 2019-05-29 at 11 50 58 AM

1. Make sure the profile wizard is visible (with flag or `profileWizardComplete` var).
2. Go to `wp-admin/admin.php?page=wc-admin#/?step=store-details`
3. Fill in form data.
4. Submit the form.
5.  Note that fields were in WC settings.
6.  Note the "Continue" button doesn't show until required fields are filled in.

### DIFF
--- a/client/dashboard/profile-wizard/index.js
+++ b/client/dashboard/profile-wizard/index.js
@@ -2,6 +2,7 @@
 /**
  * External dependencies
  */
+import { __ } from '@wordpress/i18n';
 import apiFetch from '@wordpress/api-fetch';
 import { Component, createElement, Fragment } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
@@ -21,8 +22,8 @@ import ProfileWizardHeader from './header';
 import Plugins from './steps/plugins';
 import Start from './steps/start';
 import Industry from './steps/industry';
+import StoreDetails from './steps/store-details';
 import './style.scss';
-import { __ } from '@wordpress/i18n';
 
 const getSteps = () => {
 	const steps = [];
@@ -36,7 +37,7 @@ const getSteps = () => {
 	} );
 	steps.push( {
 		key: 'store-details',
-		container: Fragment,
+		container: StoreDetails,
 		label: __( 'Store Details', 'woocommerce-admin' ),
 	} );
 	steps.push( {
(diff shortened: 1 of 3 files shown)


justinshreve:
> There doesn't appear to be a default admin email. Should the email here overwrite the WordPress admin_email (seems risky), set to woocommerce_stock_email_recipient which is used for notifications, or save separately under profiler data?

`woocommerce_stock_email_recipient` is only used for stock notifications, so all other store emails would not be sent to this address.

It unfortunately looks like WooCommere uses `admin_email` in a lot of places to set default values for settings after setup (PayPal settings, different email notifications, etc).

I agree just overwriting WordPress' `admin_email` is risky without explaining what's going on though.

Maybe we could just drop the email field from the wizard? They should always have an admin email setup from installing WordPress in the first place, and they will be able to change/configure their email in later steps or settings screens. cc @jameskoster.

justinshreve:
> I've added "Address Line 2" but I believe a city will need to be added as well to have a complete address. Should we add that as well or remove the 2nd line address?

I think we want both "Address Line 2" and "City".

joshuatf:
@justinshreve Thanks for the review!  Rebased this and updated with Muriel components.

I agree with your sentiment on the "Email" field; removed that and added a "City" field to provide a complete address.

Also added updated a few styles here to match the updated designs.

PR #2310, opened 2019-05-27
Add profile onboarding mutators and selectors to wc-api

### LINKED ISSUE #2303: Onboarding - Create WC API store for profile params
Currently #2283 manually creates a call to the onboarding profile controller to update params.  

We should create a store in wc-api not only for better reusability, but also to reflect the correct variables from the store after update to show/hide the profiler:

https://github.com/woocommerce/woocommerce-admin/blob/4120b79de9d7c9f7839be9016bb8acdafdd2f138/client/dashboard/index.js#L27

This issue will involve:
* Creating the wc-api selectors, mutations, and operations.
* Hydrating the initial data for these settings from server-side variables.
* Connecting the values to the dashboard profiler logic in `dashboard/index.js`

joshuatf:
Fixes #2303

Creates an easy way to store and retrieve the profiler information in the state via the API.

1.  Set the `skipped` and `completed` properties to (boolean) false.
2. Note that the profiler is shown.
3. In an HOC add `const { updateProfileItems } = dispatch( 'wc-api' );` in any component.
4. Update the skipped property `this.props.updateProfileItems( { skipped: true } );`.
5. Note that the property is updated and the profiler is hidden.

### DIFF
--- a/client/wc-api/onboarding/operations.js
+++ b/client/wc-api/onboarding/operations.js
@@ -1,0 +1,97 @@
+/** @format */
+
+/**
+ * External dependencies
+ */
+import apiFetch from '@wordpress/api-fetch';
+
+/**
+ * Internal dependencies
+ */
+import { getResourceName } from '../utils';
+
+const NAMESPACE = '/wc-admin/v1';
+
+function read( resourceNames, fetch = apiFetch ) {
+	return [ ...readProfileItems( resourceNames, fetch ) ];
+}
+
+function update( resourceNames, data, fetch = apiFetch ) {
+	return [ ...updateProfileItems( resourceNames, data, fetch ) ];
+}
+
+function readProfileItems( resourceNames, fetch ) {
+	const resourceName = 'onboarding-profile';
+
+	if ( resourceNames.includes( resourceName ) ) {
+		const url = NAMESPACE + '/onboarding/profile';
+
+		return [
+			fetch( { path: url } )
+				.then( profileItemsToResources )
+				.catch( error => {
+					return { [ resourceName ]: { error: String( error.message ) } };
+				} ),
+		];
+	}
+
+	return [];
+}
+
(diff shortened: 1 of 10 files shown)


includes/features/onboarding/class-wc-admin-onboarding.php
```diff
@@ -41,28 +41,23 @@ public function __construct() {
 	 * @return bool
 	 */
 	public function should_show_profiler() {
-		// @todo Remove this once we have a proper way to dismiss the profiler.
-		if ( ! defined( 'WOOCOMMERCE_ADMIN_DEV_SHOW_PROFILER' ) || false === WOOCOMMERCE_ADMIN_DEV_SHOW_PROFILER ) {
```
justinshreve: This was a short lived change 😆.

joshuatf: Haha, I didn't realize it would be gone so quick either.

client/wc-api/onboarding/operations.js
```diff
@@ -0,0 +1,97 @@
...
+ * External dependencies
+ */
+import apiFetch from '@wordpress/api-fetch';
+
+/**
+ * Internal dependencies
+ */
+import { getResourceName } from '../utils';
+
+const NAMESPACE = '/wc-admin/v1';
```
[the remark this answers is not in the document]
joshuatf: Good idea.  Added that and replaced references to `/wc-admin/v1` in https://github.com/woocommerce/woocommerce-admin/pull/2310/commits/6c2bb2f3b0f34158600769490d283e853129aad1

joshuatf:
Thanks for the review @justinshreve!  Implemented your feedback and replaced constants where applicable.

>Edit: Also needs the base branch updated against master.

Ha, thank you- feel like I needed this reminder today 😅

PR #2313, opened 2019-05-27
Add Muriel/Newspack components for use in onboarding

justinshreve:
This PR pulls in the recently published Newspack components, which allows us to reuse some patterns like the form inputs. I did run into a couple things that may need changed, which I've logged here: https://github.com/Automattic/newspack-plugin/pull/72#issuecomment-496292548 -- but this is working well for us to start implementing. I also make a few slight style changes to the main start page, per recent Figma updates.

To Test:
* Visit the main onboarding page, view updated button / checkbox styles.
* Visit `/wp-admin/admin.php?page=wc-admin#/?step=industry` and test the select, input, and button styles.

Screen Shot 2019-05-27 at 3 30 54 PM

Screen Shot 2019-05-27 at 1 16 35 PM

### DIFF
--- a/client/dashboard/profile-wizard/steps/start/index.js
+++ b/client/dashboard/profile-wizard/steps/start/index.js
@@ -3,7 +3,8 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { Button, CheckboxControl, FormToggle } from '@wordpress/components';
+import { FormToggle } from '@wordpress/components';
+import { Button, CheckboxControl } from 'newspack-components';
 import { Component, Fragment } from '@wordpress/element';
 import interpolateComponents from 'interpolate-components';
 
@@ -96,7 +97,7 @@ export default class Start extends Component {
 
 		const trackingLabel = interpolateComponents( {
 			mixedString: __(
-				'Help improve WooCommerce with usage tracking. {{link}}Learn More.{{/link}}',
+				'Help improve WooCommerce with {{link}}usage tracking{{/link}}',
 				'woocommerce-admin'
 			),
 			components: {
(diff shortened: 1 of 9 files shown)


client/dashboard/profile-wizard/style.scss
```diff
@@ -108,6 +108,41 @@
...
+		}
+
+		input[type='text']:focus {
+			box-shadow: none;
+			border: 0;
+		}
+	}
+
+	.muriel-input-text label,
+	.muriel-select label {
```
joshuatf: The empty state label looks slightly off-center vertically.  Not sure if that's on the Muriel side or one of our styles affecting this:

```scss
.muriel-select.empty label {
    top: 15px;
}
```

Similarly the input fields with a value look a pixel or two off.

Screen Shot 2019-05-28 at 1 59 08 PM

postcss.config.js
```diff
@@ -10,6 +10,13 @@ module.exports = {
 				outlines: '#007cba',
 			},
 			themes: {
+				'woocommerce-profile-wizard__body': {
```
joshuatf: Will these color values need to be updated?  If so, I can log some issues around them.  Also curious why the `woocommerce-page` values are all set to `$woocommerce-500`.

I expected those values to match up with [these](https://github.com/woocommerce/woocommerce/blob/master/assets/css/_variables.scss#L11) but I might be misunderstanding the purpose of these variables.

client/dashboard/profile-wizard/steps/start/style.scss
```diff
@@ -20,13 +20,42 @@
...
+		.muriel-checkbox label.components-checkbox-control__label {
+			margin-left: $gap-large;
+		}
+
+		.muriel-checkbox input {
+			background-color: $muriel-hot-purple-600;
+			border-color: $muriel-hot-purple-600;
+		}
+
+		.components-form-toggle__track {
```
joshuatf: One more small change to the toggle, it appears that it's only `18px` in the design:

Screen Shot 2019-05-28 at 2 09 59 PM

And the current toggle is 24px:

Screen Shot 2019-05-28 at 2 10 37 PM

client/dashboard/profile-wizard/steps/start/index.js
```diff
@@ -125,29 +126,33 @@ export default class Start extends Component {
...
-					/>
-				</div>
-
 				<Card>
 					<div className="woocommerce-profile-wizard__benefits">
 						{ benefits.map( benefit => this.renderBenefit( benefit ) ) }
 					</div>
 
-					<Button isPrimary onClick={ this.startWizard }>
+					<div className="woocommerce-profile-wizard__tracking">
```
joshuatf: One small change to this classes CSS after the update:

```scss
.woocommerce-profile-wizard__tracking {
    margin-bottom: $gap;
}
```

client/dashboard/profile-wizard/style.scss
```diff
@@ -108,6 +108,41 @@
...
+	.muriel-input-text.active,
+	.muriel-select.active {
+		border: 2px solid $muriel-hot-purple-600;
+
+		input,
+		select {
+			top: 25px;
+		}
+	}
+
```
joshuatf: Should we set the default button height here as `48px` (currently `40px`)?

Screen Shot 2019-05-28 at 2 06 00 PM

postcss.config.js
justinshreve: I believe these will need to be updated, but they don't seem to be applying correctly for me in all circumstances. The primary/button colors are working (which is what we need in this PR), but changing `toggle` here doesn't seem to work. I'll log an issue for follow-up.

The link you provided are existing core colors, but are not part of the rebrand colors (with the pinks and purples).

Here are the toggle, outlines, primary, and button classes being shown:

Screen Shot 2019-05-28 at 9 04 00 AM

justinshreve: https://github.com/woocommerce/woocommerce-admin/issues/2322

PR #2320, opened 2019-05-28
allow additional parameters in the page query string

### LINKED ISSUE #2314: Plugins that add to URL break woocommerce Admin
When a plugin - in my case security plugin - adds to the URL on the backend Woocommerce Admin links result into a blank page.

Example:
Geo-ip-Block plugin adds &ip-geo-block-auth-nonce=9e0b011227 to the url

https://www.domain.pt/wp-admin/admin.php?page=wc-admin&ip-geo-block-auth-nonce=9e0b011227#/analytics/revenue

The result is a black screen on any of the new Woocommerce admin pages.

Removing the added URL allows page to load correctly.
https://www.domain.pt/wp-admin/admin.php?page=wc-admin#/analytics/revenue
Unsure if there is a potential work arround, however i believe many security plugins are known to add session handles to the URL

Any advise?
Thank you

rrennick:
Fixes #2314 

This PR allows additional query parameters in the URL query string (eg. `?page=wc-admin&var=value#/analytics/revenue`. 

- Repeat without and with PR
- Go to Analytics -> Orders (URL is `/wp-admin/admin.php?page=wc-admin#/analytics/orders`)
- Add a query parameter in the address bar `/wp-admin/admin.php?page=wc-admin&abc=123#/analytics/orders`
- Without PR is an empty page
- With PR the Orders page loads

### Changelog Note:

> Allow additional query string parameters in page URLs.

### DIFF
--- a/includes/page-controller/class-wc-admin-page-controller.php
+++ b/includes/page-controller/class-wc-admin-page-controller.php
@@ -116,7 +116,7 @@ class WC_Admin_Page_Controller {
 
 				if (
 					$page_path === $current_path &&
-					$page_query === $current_query &&
+					0 === strpos( $current_query, $page_query ) &&
 					$page_fragment === $current_fragment
 				) {
 					$this->current_page = $page;


rrennick:
@Aljullu Thanks :)

PR #2325, opened 2019-05-28
Trigger an action server side when admin note actions are clicked

jeffstieler:
In support of #2268.

This PR introduces a new endpoint `/notes/[id]/action/[id]` to trigger actions on the server side when an action is "taken".

The endpoint is hit when admin note actions are clicked - this supplants the current behavior where the `/notes/[id]` endpoint is used to update the `status` of the note to the one found on the taken action. The `status` is still updated, but through a new endpoint.

- Add an admin note with actions if you don't already have some
- Hook into `'woocommerce_admin_note_action'` to verify that a WP hook is triggered
- Verify that an XHR request is made when clicking an action button
- Verify that the WP hook was fired
- Verify that the `status` of the parent note has changed to that of the action

### DIFF
--- a/includes/api/class-wc-admin-rest-admin-note-action-controller.php
+++ b/includes/api/class-wc-admin-rest-admin-note-action-controller.php
@@ -58,7 +58,7 @@ class WC_Admin_REST_Admin_Note_Action_Controller extends WC_Admin_REST_Admin_Not
 		if ( ! $note ) {
 			return new WP_Error(
 				'woocommerce_admin_notes_invalid_id',
-				__( 'Sorry, there is no resouce with that ID.', 'woocommerce-admin' ),
+				__( 'Sorry, there is no resource with that ID.', 'woocommerce-admin' ),
 				array( 'status' => 404 )
 			);
 		}
@@ -77,7 +77,7 @@ class WC_Admin_REST_Admin_Note_Action_Controller extends WC_Admin_REST_Admin_Not
 		if ( ! $triggered_action ) {
 			return new WP_Error(
 				'woocommerce_admin_note_action_invalid_id',
-				__( 'Sorry, there is no resouce with that ID.', 'woocommerce-admin' ),
+				__( 'Sorry, there is no resource with that ID.', 'woocommerce-admin' ),
 				array( 'status' => 404 )
 			);
 		}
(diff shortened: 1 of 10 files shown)


includes/api/class-wc-admin-rest-admin-note-action-controller.php
```diff
@@ -0,0 +1,111 @@
...
+					),
+					'action_id' => array(
+						'description' => __( 'Unique ID for the Note Action.', 'woocommerce-admin' ),
+						'type'        => 'integer',
+					),
+				),
+				array(
+					'methods'             => WP_REST_Server::EDITABLE,
+					'callback'            => array( $this, 'trigger_note_action' ),
+					// @todo - double check these permissions for taking note actions.
```
timmyc: what additional check is needed here beyond using the same `get_item` check?

jeffstieler: `get_item` is essentially a "read" operation, where this is a "write". I'm not sure if those perms should differ, so I made a note.

rrennick:
@jeffstieler I'm getting a fatal on a note that is about a month old. New notes may not produce this error

```
PHP Fatal error:  Uncaught WC_Data_Exception: The admin note status prop cannot be empty. in /wp-content/plugins/woocommerce/includes/abstracts/abstract-wc-data.php:794
```

Somewhat related to this PR, offsite actions/queries take the user away from the dashboard. Could we add a condition to the action query logic where if the query field starts with `http` then the action is opened in a new tab? That way, any action within WCA would use the current window, everything else would be a new tab. This might be better as a separate PR but wanted to mention it for discussion.

jeffstieler:
@rrennick 

> I'm getting a fatal on a note that is about a month old. New notes may not produce this error

Is the `status` of the note action you took empty? That might be causing it (and would be possible today).

> Somewhat related to this PR, offsite actions/queries take the user away from the dashboard. Could we add a condition to the action query logic where if the query field starts with http then the action is opened in a new tab? That way, any action within WCA would use the current window, everything else would be a new tab.

This is a great idea.

> This might be better as a separate PR but wanted to mention it for discussion.

+1

rrennick:
I tried POSTing directly to `/wp-json/wc/v4/admin/notes/6/action/6` with a REST client and the same fatal appears in the log. The action record is

```shell
|         6 |       6 | browse         | Browse         | https://woocommerce.com/success-stories/                           |          |          0 |
```

jeffstieler:
Yep, looks like the action status is empty, which causes a the error to be thrown when updating the note status. I'll add some resilience to this.

jeffstieler:
@rrennick fixed in 99b48e9.

Issue #1985, opened 2019-04-02
Fatal error if customer doesn't exist


n9yty:
**Describe the bug**
In processing the historical data, it aborts while importing batches. In our case, it seemed to be several orders that, for whatever reason, no longer have a customer in the system.

**To Reproduce**
Re-process historical orders with missing customers.

**Expected behavior**
The error should be handled so it does not cause a fatal error, which also seems to abort processing the rest of the items in that particular scheduled group.

One example:

Running 100 actions  36 % [=======================================>                                                                     ] 0:13 / 0:38
Started processing action 1335011
[02-Apr-2019 19:35:52 UTC] PHP Fatal error:  Uncaught Error: Call to a member function date() on null in /var/www/html/wp-content/plugins/woocommerce-admin/includes/data-stores/class-wc-admin-reports-customers-data-store.php:477
Stack trace:
#0 /var/www/html/wp-content/plugins/woocommerce-admin/includes/data-stores/class-wc-admin-reports-orders-stats-data-store.php(407): WC_Admin_Reports_Customers_Data_Store::get_or_create_customer_from_order(Object(WC_Admin_Order))
#1 /var/www/html/wp-content/plugins/woocommerce-admin/includes/data-stores/class-wc-admin-reports-orders-stats-data-store.php(368): WC_Admin_Reports_Orders_Stats_Data_Store::update(Object(WC_Admin_Order))
#2 /var/www/html/wp-content/plugins/woocommerce-admin/includes/class-wc-admin-reports-sync.php(269): WC_Admin_Reports_Orders_Stats_Data_Store::sync_order(16357)
#3 /var/www/html/wp-content/plugins/woocommerce-admin/includes/class-wc-admin-reports-sync.php(255): WC_Admin_Reports_Sync::orders_lookup_process_order(16357)
#4 /var/www/html/wp-includes/class-wp-hook.php(286): WC_Admin_Reports_Syn in /var/www/html/wp-content/plugins/woocommerce-admin/includes/data-stores/class-wc-admin-reports-customers-data-store.php on line 477
Fatal error: Uncaught Error: Call to a member function date() on null in /var/www/html/wp-content/plugins/woocommerce-admin/includes/data-stores/class-wc-admin-reports-customers-data-store.php:477
Stack trace:
[...]

PR #2326, opened 2019-05-28
Prevent potential get_customer_id() & other functions running into a fatal error when $order is a bool (false)

KZeni:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/1985#issue-428407008, https://github.com/woocommerce/woocommerce-admin/issues/1985#issuecomment-496621183, and https://github.com/woocommerce/woocommerce-admin/issues/1985#issuecomment-496651780

As I reported in my comment at https://github.com/woocommerce/woocommerce-admin/issues/1985#issuecomment-496621183, I had a site running the order import on the current version of WooCommerce Admin (0.12.0) and the import would fail due to `Call to a member function get_customer_id() on boolean in /wp-content/plugins/woocommerce-admin/includes/data-stores/class-wc-admin-reports-customers-data-store.php:441`. I addressed this by making sure the `$order` variable isn't false before trying to run `get_customer_id()` against it, and it returns false for that function if the order being checked was false to begin with (which appears to be seamless in terms of normal operation).

This could potentially be resolved earlier on in the process, but this seems like a harmless fix for if/when this does come up. I'm not familiar enough with the code at this point to recommend something other than this, currently.

1. Freshly install WooCommerce Admin 0.12.0 on a longtime running WooCommerce setup with multiple extensions, REST API implementations, etc.
2. Tell it to import All of the past data into WooCommerce Admin.
3. Have it import things for the most part, but then find that WooCommerce => Status => Scheduled Actions has a fatal error preventing the next steps of the import from finishing (the `get_existing_customer_id_from_order()` function was given `$order` that's set to false so the `get_customer_id()` function can't then run on it & has a fatal error.) Retrying the import has it stop immediately at that same step with the same error.
4. Make the edit suggested by this PR to find that its now able to continue forward with the rest of the import.

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-customers-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-customers-data-store.php
@@ -438,22 +438,18 @@ class WC_Admin_Reports_Customers_Data_Store extends WC_Admin_Reports_Data_Store 
 	 * @return int|bool
 	 */
 	public static function get_existing_customer_id_from_order( $order ) {
-		if($order){ // Only try if the order details exist
-			$user_id = $order->get_customer_id();
+		$user_id = $order->get_customer_id();
 
-			if ( 0 === $user_id ) {
-				$email = $order->get_billing_email( 'edit' );
+		if ( 0 === $user_id ) {
+			$email = $order->get_billing_email( 'edit' );
 
-				if ( $email ) {
-					return self::get_guest_id_by_email( $email );
-				} else {
-					return false;
-				}
+			if ( $email ) {
+				return self::get_guest_id_by_email( $email );
 			} else {
-				return self::get_customer_id_by_user_id( $user_id );
+				return false;
 			}
-		}else{
-			return false;
+		} else {
+			return self::get_customer_id_by_user_id( $user_id );
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


KZeni:
I've updated my PR so the `get_or_create_customer_from_order()` function itself (which is calling these functions without the necessary `$order` data) is what's being addressed so it addresses it at its core a bit better (https://github.com/woocommerce/woocommerce-admin/issues/1985#issuecomment-496651780 showed it wasn't just the `get_customer_id()` function I was having issue with and I went back to the fact that https://github.com/woocommerce/woocommerce-admin/issues/1985#issue-428407008 had the issue on `date()` within `get_or_create_customer_from_order()`m and the one thing they had in common was `get_or_create_customer_from_order()` was calling the functions with the fatal errors.)

includes/data-stores/class-wc-admin-reports-customers-data-store.php
```diff
@@ -460,56 +460,60 @@ public static function get_existing_customer_id_from_order( $order ) {
 	 * @return int|bool
 	 */
 	public static function get_or_create_customer_from_order( $order ) {
-		global $wpdb;
-		$returning_customer_id = self::get_existing_customer_id_from_order( $order );
+		if( false !== $order){ // Only try if the order details exist
```
jeffstieler: Rather than wrap this entire block in an `if()` statement, let's just check for the positive case `if ( false === $order )` and `return false;` earlier.

Also, this line doesn't adhere to the WP coding standards. If you `composer install` before committing, PHP Code Sniffer will catch things like this.

timmyc:
@KZeni thanks for the PR! Would you have time to update per @jeffstieler 's feedback above or would you like one of us to jump in and help?

jeffstieler:
Closing in favor of https://github.com/woocommerce/woocommerce-admin/pull/2464 - thanks for your contribution @KZeni!

PR #2327, opened 2019-05-28
Only show unactioned notes in the Inbox panel.

jeffstieler:
And hide notes once an action has been taken.

In support of #2268.

Hide notes from the Inbox that are in a `status` other than `unactioned`. Hide notes once an action has been taken.

### Accessibility

I don't believe the `prefers-reduced-motion` applies for this transition, but please correct me if I'm wrong about that!

- Open a WooCommerce Admin page
- Open the Inbox
- Take an action on a note
- Verify the note transitions and is removed from the inbox
- Refresh page
- Verify the note isn't visible

### DIFF
--- a/client/header/activity-panel/activity-card/style.scss
+++ b/client/header/activity-panel/activity-card/style.scss
@@ -223,7 +223,10 @@
 	opacity: 1;
 	padding: $fallback-gutter;
 	padding: $gutter;
-	transition: opacity 0.6s, height 0s, padding 0s;
+
+	@media screen and (prefers-reduced-motion: no-preference) {
+		transition: opacity 0.3s, height 0s, padding 0s;
+	}
 
 	@include breakpoint( '<782px' ) {
 		grid-template-columns: 64px 1fr;
@@ -237,7 +240,9 @@
 		height: 0;
 		opacity: 0;
 		padding: 0;
-		transition: opacity 0.6s, height 0s 0.6s, padding 0s 0.6s;
+		@media screen and (prefers-reduced-motion: no-preference) {
+			transition: opacity 0.3s, height 0s 0.3s, padding 0s 0.3s;
+		}
 	}
 }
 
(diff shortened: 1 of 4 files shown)


client/header/activity-panel/activity-card/style.scss
```diff
@@ -227,6 +232,13 @@
 	.woocommerce-activity-card__header {
 		margin-bottom: $gap-small;
 	}
+
+	&.actioned {
+		height: 0;
+		opacity: 0;
+		padding: 0;
+		transition: opacity 1s, height 0s 1s, padding 0s 1s;
```
Aljullu: That's my personal opinion, but I felt the animation to be a bit slow. :turtle: What would you think about making it a bit faster to `0.6s` or something like this?

jeffstieler:
> I don't have a strong opinion here, but my point of view is that all `transition` and `animation` declarations should be inside a `prefers-reduced-motion` media query just to play it safe.

I think it might be more jarring for someone who has `prefers-reduced-motion` enabled to have the notice disappear with no transition than to have the opacity animated. What would you do in the case of `prefers-reduced-motion` here?

client/header/activity-panel/activity-card/style.scss
jeffstieler: Done in cf8ea88.

Aljullu:
> I think it might be more jarring for someone who has prefers-reduced-motion enabled to have the notice disappear with no transition than to have the opacity animated. What would you do in the case of prefers-reduced-motion here?

@jeffstieler I was thinking on removing the transition completely and only enabling it when `prefers-reduced-motion` is not defined:

```CSS
@media (prefers-reduced-motion: no-preference) {
	transition: opacity 1s, height 0s, padding 0s;
}
```

Peek 2019-05-31 10-41

I don't think it's _that_ jarring, many applications work this way: removing a file from the file manager, removing a bookmark from the browser, etc. there is no animation at all, the element just disappears and the next element occupies its place.

But as I said, I don't have a strong opinion here, so I'm ok merging this PR as-is, and we can discuss if we want to change this behavior in the future and (maybe) write some guidelines about when `prefers-reduced-motion` should and should not be applied.

jeffstieler:
@Aljullu I agree with your reasoning. In 48a97ba I added the `prefers-reduced-motion` check before using animations, and I also sped up the animation a bit. Thanks for your feedback and review!

PR #2334, opened 2019-05-29
Hook up onboarding step 1 actions

### LINKED ISSUE #2294: Onboarding - Hook up step 1 to settings data
Once both #2283, #2293, and #2310 are merged in, this issue will involve:

* Hooking up the state data reflecting the tracking opt-in to the settings data.
* Updating the skipped profile param via the new wc-api mutators.
* Removing the `updateProfile()` prop passed down from `profile-wizard/index.js`.

Using an HOC, we'll be able to update this through the wc settings API:

```js
updateSettings( { advanced: settingData } );
updateProfileItems( { skipped: true } );
```

This should occur when progressing to the next step inside the existing `startWizard()` method.

joshuatf:
Fixes #2294 

Saves the profile data (skipped) and settings data (wc tracking) for the first step of the onboarding profiler.

### Screenshots
Screen Shot 2019-05-29 at 12 30 39 PM

1.  Make sure `skipped` and `completed` are set to false in your onboarding profile data.
2. Visit the first step - `wp-admin/admin.php?page=wc-admin#/#`
3. Toggle the tracking checkbox on/off and click "Get Started."
4. Check that the WC tracking is enabled/disabled under WC advanced settings based on the toggle.
5. Click "Proceed without Jetpack or WooCommerce Services" and make sure `skipped` is set to true and the wizard is closed.

### DIFF
--- a/client/dashboard/profile-wizard/index.js
+++ b/client/dashboard/profile-wizard/index.js
@@ -96,9 +96,10 @@ class ProfileWizard extends Component {
 		const currentStepIndex = getSteps().findIndex( s => s.key === currentStep.key );
 		const nextStep = getSteps()[ currentStepIndex + 1 ];
 
-		if ( 'undefined' === nextStep ) {
-			await updateProfileItems( { complete: true } );
-			if ( ! isError ) {
+		if ( 'undefined' === typeof nextStep ) {
+			await updateProfileItems( { completed: true } );
+
+			if ( isError ) {
 				addNotice( {
 					status: 'error',
 					message: __( 'There was a problem completing the profiler.', 'woocommerce-admin' ),
@@ -139,7 +140,7 @@ export default compose(
 		return { isError };
 	} ),
 	withDispatch( dispatch => {
-		const { addNotice, updateProfileItems } = dispatch( 'wc-admin' );
+		const { addNotice, updateProfileItems } = dispatch( 'wc-api' );
 
 		return {
 			addNotice,
--- a/client/dashboard/profile-wizard/index.js
+++ b/client/dashboard/profile-wizard/index.js
@@ -3,7 +3,6 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import apiFetch from '@wordpress/api-fetch';
 import { Component, createElement, Fragment } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import { pick } from 'lodash';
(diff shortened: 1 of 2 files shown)


client/dashboard/profile-wizard/index.js
```diff
@@ -90,32 +88,26 @@ class ProfileWizard extends Component {
...
 	}
 
-	goToNextStep() {
+	async goToNextStep() {
+		const { addNotice, isError, updateProfileItems } = this.props;
 		const currentStep = this.getCurrentStep();
 		const currentStepIndex = getSteps().findIndex( s => s.key === currentStep.key );
 		const nextStep = getSteps()[ currentStepIndex + 1 ];
 
 		if ( 'undefined' === nextStep ) {
```
justinshreve: I know it's not changed here, but this should be `'undefined' === typeof nextStep` to correctly work.

client/dashboard/profile-wizard/index.js
```diff
@@ -90,32 +88,26 @@ class ProfileWizard extends Component {
...
-			this.updateProfile( { complete: true } );
+			await updateProfileItems( { complete: true } );
```
justinshreve: I tried commenting out the other steps here, so that clicking the get started button mark the profiler as completed, but I'm getting the following:

`Uncaught (in promise) TypeError: Cannot read property 'apply' of undefined`.

client/dashboard/profile-wizard/index.js
```diff
@@ -90,32 +88,26 @@ class ProfileWizard extends Component {
...
+			if ( ! isError ) {
```
justinshreve: This should be `if ( isError ) {`

client/dashboard/profile-wizard/steps/start/index.js
```diff
@@ -53,31 +56,51 @@ const benefits = [
...
 	}
 
-	startWizard() {
-		// @todo This should update the settings with the tracking selection. See #2281.
-		this.props.goToNextStep();
+	async startWizard() {
+		const { addNotice, isSettingsError, updateSettings } = this.props;
+
+		const allowTracking = this.state.allowTracking ? 'yes' : 'no';
+		await updateSettings( { advanced: { woocommerce_allow_tracking: allowTracking } } );
```
justinshreve: Should we also save the tracking preference if they skip the profiler? As is, we would only ever have people opt-in if they also opt in to Jetpack/WooCommerce Services.

client/dashboard/profile-wizard/index.js
joshuatf: Thanks for noting that!  Turns out I had written `dispatch( 'wc-admin' )` below so this wasn't really connected to `wc-api`.

client/dashboard/profile-wizard/steps/start/index.js
joshuatf: I would think that skipping the profiler means any options on that page won't be applied (I may not have even read the page contents if I'm looking for an obvious way to skip).

/cc @jameskoster Do you have an opinion on this?

PR #2344, opened 2019-05-29
Allow paragraph tags in admin notes.

jeffstieler:
In support of #2268.

Allow `<p>` tags in admin note content.

- Add an admin note that contains paragraph tags in the content
- Verify the paragraph tags are rendered properly


jeffstieler:
> Would make it sense to unify them

Yep! Good catch. I'll open a new PR for that.

PR #2352, opened 2019-05-31
Fix wrong average numbers in chart legends

### LINKED ISSUE #2279: Chart: Average chart types compute totals in the legend
**Describe the bug**

When looking at charts that display averages, the legend should also display an average, not a sum.

**To Reproduce**

1. Go to Orders Report and select the Average Order Value chart.
2. See the SummaryNumbers mismatch the legend.

**Expected behavior**

SummaryNumbers and legend should match. The legend is currently summing the data values, not averaging them.

**Screenshots**

Screen Shot 2019-05-22 at 3 15 37 PM

Aljullu:
Fixes #2279.

Adds a `totals` prop to `<Chart>` component that allows overwriting the numbers shown in the legend. This way, we can directly pass down the averages already returned by the API call.

### Screenshots
_Before:_
image

_After:_
image

- Go to a report that shows an average number, for example _Analytics_ > _Orders_ and _Average Order Value_.
- Verify the number shown in the Summary Numbers is the same as in the Chart legend.

### DIFF
--- a/client/analytics/components/report-chart/index.js
+++ b/client/analytics/components/report-chart/index.js
@@ -103,7 +103,16 @@ export class ReportChart extends Component {
 		return chartData;
 	}
 
-	renderChart( mode, isRequesting, chartData ) {
+	getTimeChartTotals() {
+		const { primaryData, secondaryData, selectedChart } = this.props;
+
+		return {
+			primary: get( primaryData, [ 'data', 'totals', selectedChart.key ], null ),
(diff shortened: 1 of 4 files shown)


Aljullu:
> I think changing `totals` to `legendTotals` might be better. What do you think?

Agree. Fixed in f25d9149f68b7ba131b61bbbee3a65c4fcf7d106.

> I also pushed a commit to close the loop on this one. The Dashboard Section example has an average value use of chart, so I updated that code in e57febd.

Good catch! Thanks for fixing it.

PR #2368, opened 2019-06-06
Misspelling at update config.jp

shoheitanaka:
This sentence is order, but older.

Fixes #

### Accessibility

- Ex: Open page `url`
- Click XYZ…


psealock:
Thank you @shoheitanaka !

Issue #2372, opened 2019-06-07
Fatal error: Uncaught Error: Class 'WC_Admin_Loader' not found


anshu-mwb:
**Describe the bug**
After Activating WooCommerce Plugin as well as WooCommerce Admin when you deactivate WooCommerce Plugin the Fatal Error is displaying.
Here is the Fatal Error-
Fatal error: Uncaught Error: Class 'WC_Admin_Loader' not found ...wp-content/plugins/woocommerce-admin/woocommerce-admin.php on line 137

**To Reproduce**
1. Just need to activate WooCommerce Plugin
2. Then Activate WooCommerce Admin Plugin
3. Now, Deactivate WooCommerce Plugin. 

On Deactivating WooCommerce plugin. You will get the Fatal Error.

**Expected behavior**
On deactivating the WooCommerce Plugin. It should deactivate without any Error message.

**Desktop:**

* OS: Linux mint 18
* Browser Mozilla Firefox
* Version: 67

timmyc:
Thanks for the bug report @anshu-mwb - looks like this bug was fixed in #2226 and is no longer present in `master`. We should hopefully have a new release out with this fix included later this week.

PR #2374, opened 2019-06-07
fix-issue-2372

anshu-mwb:
Fixes #

### Accessibility

- Ex: Open page `url`
- Click XYZ…

### DIFF
--- a/_woocommerce-admin.php
+++ b/_woocommerce-admin.php
@@ -1,0 +1,310 @@
+<?php
+/**
+ * Plugin Name: WooCommerce Admin
+ * Plugin URI: https://github.com/woocommerce/woocommerce-admin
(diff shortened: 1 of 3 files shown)


psealock:
Thanks for the contribution @anshu-mwb,

I actually can't reproduce the issue on master. I'm getting the correct message:

Screen Shot 2019-06-11 at 1 09 42 PM

Are there any steps to reproduce that are missing?

jeffstieler:
It looks like this was taken care of - see: https://github.com/woocommerce/woocommerce-admin/issues/2372#issuecomment-500911568

@anshu-mwb can you confirm that the fatal error no longer occurs from you on `master`? If there is still an issue, can you describe some steps to reproduce?

joshuatf:
I think this issue is fixed by #2226, so I'm going to close this PR for now.  If you still encounter this issue @anshu-mwb feel free to [open a new issue](https://github.com/woocommerce/woocommerce-admin/issues/new/choose).  Thanks for the help!

Issue #2365, opened 2019-06-05
Endless requests breaking sites


psealock:
This is a copy of an issue logged in WooCommerce, https://github.com/woocommerce/woocommerce/issues/23868.

>While adding products in after the update, I started having issues with my server maxing out the entry processes. I went to the logs to see what was happening. I discovered that there were some endpoints that were being repeatedly hit thousands of times until the server resources were expended. Here are the urls that were called over and over and over.

>GET /wp-json/wc/v4/admin/notes?page=1&per_page=25&status=unactioned&type=error%2Cupdate&_locale=user

>GET /wp-json/wp/v2/users/me?context=edit&_locale=user

>GET /wp-json/wc/v4/admin/notes?order=desc&orderby=date&page=1&per_page=1&type=info%2Cwarning&_locale=user

>GET /wp-json/wc/v4/products?low_in_stock=true&page=1&per_page=1&status=publish&_locale=user

>GET /wp-json/wc/v4/reports/orders?page=1&per_page=0&status_is%5B0%5D=processing&status_is%5B1%5D=on-hold&_locale=user

>GET /wp-json/wc/v4/admin/notes?page=1&per_page=25&status=unactioned&type=error%2Cupdate&_locale=user HTTP/1.1" 200 1064

cc @BodyStatements

rrennick:
Note the additional comment suggesting this was related to having PHP 5.6 & updating to 7.X fixed it.

BodyStatements:
I did note that and I added that information to Wordpress.org too. However,
that did not fix the issue for me. I downgraded to the prior version of
Woocommerce and that "fixed it". That wasn't really a "fix", so let me
change that and say that I got my account up and running again by
downgrading.

timmyc:
I have been attempting to recreate this issue locally - so specifically seeing if I can recreate a problem where on legacy Woo pages ( i.e. Orders, Products, Product Edit ). The pattern of the API requests shown in the OP here align with the requests generated from the Activity Panels. Here is a screen grab of my network tab when loading a product listing page:

<img width="1169" alt="product-list-page-first-load" src="https://user-images.githubusercontent.com/22080/59044434-fa43cb00-8832-11e9-8d86-c3b8fcbed3a2.png">

[...]

PR #2377, opened 2019-06-09
Prevent inactive tabs from making requests

Aljullu:
Part of #2365.

This PR avoids making network requests on inactive tabs.

- Temporarily decrease the default freshness to something like 15 seconds (in `client/wc-api/constants.js`).
- Visit any WooCommerce Admin page.
- Verify there are no regressions and data loads correctly.
- Open the _Network_ devtools of your browser and detach them from the main browser window.
- Switch to another tab in your browser.
- Verify WooCommerce Admin doesn't make requests while the tab is inactive.
- Switch to WooCommerce Admin tab again.
- Verify requests are being made again.
- Try other things that you think could break, like opening WooCommerce Admin in an inactive tab, reloading the tab while it's inactive, etc. and verify everything works as expected.

### DIFF
--- a/client/wc-api/imports/operations.js
+++ b/client/wc-api/imports/operations.js
@@ -26,6 +26,10 @@ function read( resourceNames, fetch = apiFetch ) {
 		};
 
 		try {
+			if ( document.hidden ) {
+				return {};
+			}
+
 			const response = await fetch( fetchArgs );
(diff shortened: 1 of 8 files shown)


Aljullu:
Closing this in favor of #2387.

PR #2378, opened 2019-06-09
Don't create Notices tab in Activity Panel if notices HTML element was removed by a plugin

### LINKED ISSUE #2366: Notices: Cannot call 'classList' of undefined
**Describe the bug**

Some plugins remove the `wp__notice-list` element to hide notifications. We should check to make sure the element exists before manipulating its class list.

https://github.com/woocommerce/woocommerce-admin/blob/9526e0ba84d8e1f592f48ff1dc3870c7e52b1b0f/client/header/activity-panel/wordpress-notices.js#L71

**To Reproduce**

Without a plugin such as this one, https://wordpress.org/plugins/hide-plugin-updates-notifications/#description, I'm not sure how to reproduce.

Aljullu:
Fixes #2366.

This PR prevents the creation of the component `<WordPressNotices>` when there isn't a `#wp__notice-list` element in the document.

- Install and enable a plugin like `hide-plugin-updates-notifications`. It was unlisted from WP.org repo, but it can still be found in GitHub: https://github.com/wp-plugins/hide-plugin-updates-notifications.
- Browse around WooCommerce Admin.
- Verify you don't get JS errors.

### DIFF
--- a/client/header/activity-panel/index.js
+++ b/client/header/activity-panel/index.js
@@ -254,11 +254,13 @@ class ActivityPanel extends Component {
 							className="woocommerce-layout__activity-panel-tabs"
 						>
 							{ tabs && tabs.map( this.renderTab ) }
-							<WordPressNotices
-								showNotices={ 'wpnotices' === currentTab }
-								togglePanel={ this.togglePanel }
-								onCountUpdate={ this.updateNoticeFlag }
-							/>
+							{ Boolean( document.getElementById( 'wp__notice-list' ) ) && (
+								<WordPressNotices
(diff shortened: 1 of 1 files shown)


Aljullu:
> This fixes the problem. I just saw the plugin in question was removed for guideline violations. I bet it was because you shouldn't remove these types of elements.
> 
> Do you think this sets a bad precedent to require a check for elements that should always be there?

That's a good point. I don't think we should fix issues with plugins that have been unlisted from .org, but at the same time I feel it's a good practice to make sure DOM elements exist before using them, even if we expect they should always be there.

PR #2384, opened 2019-06-10
Fix search bar in Orders page not being tapable on mobile

### LINKED ISSUE #2383: Orders page search bar is shown below pagination on narrow viewports
**To Reproduce**
1. Go to `/wp-admin/edit.php?post_type=shop_order` with a narrow viewport.
2. Notice the search bar is behind the pagination buttons.

**Screenshots**
image

Aljullu:
Fixes #2383.

### Screenshots
_Before:_
image

_After:_
image

- Go to _WooCommerce_ > _Orders_ with a narrow viewport.
- Verify the search bar is displayed below the navigation buttons.

### DIFF
--- a/client/stylesheets/shared/_reset.scss
+++ b/client/stylesheets/shared/_reset.scss
@@ -7,20 +7,16 @@
 
 	#wpcontent,
 	#wpbody-content {
-		padding-left: 0;
+		padding: 0;
 		overflow-x: hidden !important;
 		min-height: calc(100vh - #{$adminbar-height});
 	}
 
-	&.woocommerce_page_wc-admin {
-		#wpbody-content {
-			padding-bottom: 0;
-		}
-	}
-
-	@include breakpoint( '>782px' ) {
-		#wpbody-content {
-			padding-bottom: 0;
+	@include breakpoint( '<782px' ) {
+		&:not(.woocommerce_page_wc-admin) {
+			#wpbody-content {
+				padding-bottom: 80px;
+			}
 		}
 	}
 
--- a/client/stylesheets/shared/_reset.scss
+++ b/client/stylesheets/shared/_reset.scss
@@ -7,9 +7,21 @@
 
 	#wpcontent,
 	#wpbody-content {
-		padding: 0;
+		padding-left: 0;
 		overflow-x: hidden !important;
 		min-height: calc(100vh - #{$adminbar-height});
+	}
+
+	&.woocommerce_page_wc-admin {
+		#wpbody-content {
+			padding-bottom: 0;
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


client/stylesheets/shared/_reset.scss
```diff
@@ -12,6 +12,14 @@
 		min-height: calc(100vh - #{$adminbar-height});
 	}
 
+	@include breakpoint( '<782px' ) {
+		&:not(.woocommerce_page_wc-admin) {
```
joshuatf: This is testing well, but wondering if we could make the selector a few lines above to be more specific so we can keep down the number of overridden rules in here:

```scss
	.woocommerce_page_wc-admin #wpbody-content {
		padding: 0;
		overflow-x: hidden !important;
		min-height: calc(100vh - #{$adminbar-height});
	}
```

https://github.com/woocommerce/woocommerce-admin/blob/2b59ad63eb218d2b7ab51a8f70283807384ca4c4/client/stylesheets/shared/_reset.scss#L9

Aljullu: I'm not sure it's possible, did you have anything in mind? We want `padding-bottom` to be 0 except when it's not in a wc-admin page and it's a narrow viewport. The only way I saw to do this is how it's currently done in this PR (overwriting the _padding-bottom_) or having to create more rulesets, which I don't think makes the situation any better. But maybe I'm missing something... :confused:

joshuatf: The default padding-bottom for this page from WP core is:

```css
#wpbody-content {
    padding-bottom: 100px;
}
```

So I was thinking, that we could only set `padding-bottom: 0;` to pages with the class `.woocommerce_page_wc-admin` in this block:

https://github.com/woocommerce/woocommerce-admin/blob/2b59ad63eb218d2b7ab51a8f70283807384ca4c4/client/stylesheets/shared/_reset.scss#L9

This way we avoid multiple overrides.  Does that work?  This looks like it works for me, but if you there are other problems that you're seeing from using this method, we can merge as is:

```scss
	#wpcontent,
	&.woocommerce_page_wc-admin #wpbody-content {
		padding: 0;
		overflow-x: hidden !important;
		min-height: calc(100vh - #{$adminbar-height});
	}
```

Aljullu: Sounds good. The only thing missing is that for >782px viewports, we want to set `padding: 0` for all WooCommerce pages (not only WooCommerce Admin pages). I refactored it 7cfc61275b94b16240700e4857dd66009a38c584, hopefully it makes it a bit better.

joshuatf: Nice- LGTM!

PR #2312, opened 2019-05-27
Add progress bars to Historical Data Import screen

Aljullu:
Fixes #1851.

This PR hooks up `import/status` endpoint to _Historical Data Import_ screen.

### Accessibility

- [x] I've tested using only a keyboard (no mouse)
- [x] I've tested using a screen reader

### Screenshots
image

### Detailed test instructions:
- Go to _Settings_ and _Import Historical Data_.
- Start an import.
- Verify progress bars update and reflect the progress status.
- Try pausing an import, deleting previous data, changing the period, reloading the page while an import is running, etc. Verify no errors are shown and page behavior matches the designs here:
https://www.figma.com/proto/G1xF07y9TuuhfirzYtqmIvTO/Analytics-Settings

PR #2385, opened 2019-06-10
Fix some data not to being imported when 'skip_existing' option is enabled

### LINKED ISSUE #2343: If 'skip_existing' is set to true, not all customers are imported
I have a store with 160 customers. When importing all historical data (after having deleted it before) with the option `skip_existing` to true, it only imports 85 of them.

Aljullu:
Fixes #2343.

When `skip_existing` was set to true, in every batch the table of customers/orders to import was getting smaller (because recently imported customers/orders were not included) but we were still increasing the `offset` in each query. That made last batches to import nothing.

Making sure `offset` is always `0` when `skip_existing` is true, fixes the issue.

I found it easier to work on this merging the branch from #2312. That way, you get progress bars in the import tool and you can see how many customers/orders are imported every time.

- Go to _Analytics_ > _Settings_.
- Click on _Delete Previously Imported Data_ and make sure all tasks run.
- Then, start an import making sure you have _Skip previously imported customers and orders_ selected.
- Verify it imports all customers and all orders.


includes/class-wc-admin-reports-sync.php
```diff
@@ -571,6 +572,7 @@ public static function get_user_ids_for_batch( $days, $skip_existing, $query_arg
 		}
 
 		if ( $skip_existing ) {
+			$query_args['paged'] = 1;
```
jeffstieler: What do you think about moving this to https://github.com/woocommerce/woocommerce-admin/blob/fc20a2f2b18b44e9d87e139e5127874b9ac0da43/includes/class-wc-admin-reports-sync.php#L664 ?

Aljullu: Sounds good, changed in 93ebea7f92d615b9d025f019850207ab4abc9c39. I also added an explanatory comment.

psealock:
I'll merge this one to get it into `0.13.0` since its been tested and added comments looking good.

PR #2387, opened 2019-06-10
Improve hidden tab suppression/fix rapid requests

### LINKED ISSUE #2365: Endless requests breaking sites

coderkevin:
Fixes #2365 

This replaces the code from #1732 and later additions to it. I am also proposing this fix instead of #2377.

This fixes the rapid request problem that was happening when switching
tabs while having timed out requests. It changes the approach from
suppressing the actions sent out to preventing the read function from
being called at all.

I think the original problem has occurred because the original approach
was relying on an internal implementation of `apiFetch` from `wp.data`. This
new approach does not have such a dependency. It prevents apiFetch from
even being called. However, in the process, it will also prevent any
newly required resources from being fetched if the user manages to
switch tabs before it is requested. (e.g. refresh, then switch tabs
quickly).

- Follow instructions on #2365 to test.
- Monitor requests in console. Switch tabs and observe requests rapid firing
- Check out this branch
- Follow instructions on #2365 and above to switch tabs and observe that no requests are sent when tab is not active.

### Changelog Note:

Fix rapid request issue when tab is not active.

### DIFF
--- a/client/wc-api/wp-data-store/create-api-client.js
+++ b/client/wc-api/wp-data-store/create-api-client.js
@@ -19,6 +19,12 @@ function createStore( name ) {
 function createDataHandlers( store ) {
 	return {
 		dataRequested: resourceNames => {
+			// This is a temporary fix until it can be resolved upstream in fresh-data.
+			// See: https://github.com/woocommerce/woocommerce-admin/pull/2387/files#r292355276
+			if ( document.hidden ) {
+				return;
+			}
+
 			store.dispatch( {
 				type: 'FRESH_DATA_REQUESTED',
 				resourceNames,
--- a/client/wc-api/wp-data-store/create-api-client.js
+++ b/client/wc-api/wp-data-store/create-api-client.js
@@ -19,11 +19,6 @@ function createStore( name ) {
 function createDataHandlers( store ) {
 	return {
 		dataRequested: resourceNames => {
-			const { resources } = store.getState();
-			const newResources = resourceNames.some( resourceName => ! resources[ resourceName ] );
-			if ( ! newResources && document.hidden ) {
-				return;
-			}
 			store.dispatch( {
 				type: 'FRESH_DATA_REQUESTED',
 				resourceNames,
--- a/client/wc-api/wc-api-spec.js
+++ b/client/wc-api/wc-api-spec.js
@@ -36,6 +36,11 @@ function createWcApiSpec() {
 		},
 		operations: {
 			read( resourceNames ) {
+				if ( document.hidden ) {
+					// Don't do any read updates while the tab isn't active.
+					return [];
+				}
+
 				return [
 					...imports.operations.read( resourceNames ),
 					...items.operations.read( resourceNames ),


client/wc-api/wp-data-store/create-api-client.js
```diff
@@ -19,11 +19,6 @@ function createStore( name ) {
 function createDataHandlers( store ) {
 	return {
 		dataRequested: resourceNames => {
-			const { resources } = store.getState();
-			const newResources = resourceNames.some( resourceName => ! resources[ resourceName ] );
-			if ( ! newResources && document.hidden ) {
```
Aljullu: After removing this, I see `FRESH_DATA_REQUESTED` is being dispatched even when the tab is inactive. That means we update the [`lastRequested`](https://github.com/woocommerce/woocommerce-admin/blob/master/client/wc-api/wp-data-store/reducer.js#L26) value even though no request is made.

That's an issue in the following scenario. Imagine I have two tabs open and WooCommerce Admin is an inactive tab. If I reload it without activating, `FRESH_DATA_REQUESTED` is dispatched (so `lastRequested` updated). Then, switching to WooCommerce Admin will show no data and it will not update it until it timeouts:
Peek 2019-06-11 10-58

Can you reproduce that too?

jeffstieler: I'm able to reproduce this @Aljullu.

jeffstieler: Adding this back in resolved for me:

```javascript
if ( document.hidden ) {
    return;
}
```

coderkevin: That makes sense. I'll see what I can do to adjust this.

Ultimately, I think this will be a temporary fix. I'm working on an update to fresh-data now and I can incorporate a better fix within that.

psealock: Can we add this temporarily with a comment? It would be nice to get this out asap in `0.13.0` today or tomorrow.

jeffstieler: Done in 26cf428.

psealock: I confirmed this solves the issue. As soon as you switch to activate the wc-admin page, a request is made.

PR #2391, opened 2019-06-11
Fix Core Build Process

justinshreve:
This PR makes sure the page handler functions are included in the core build so that https://github.com/woocommerce/woocommerce/pull/23658 works again. It also adds some defense checks against the classList DOM logic. When I was enabling just some select features in core, I was getting JS errors.

### To Test:

* Test the development build normally. Make sure things load.
* Test this branch with https://github.com/woocommerce/woocommerce/pull/23658, following the directions there.


justinshreve:
p1560438385054500-slack-team-isotope

PR #2399, opened 2019-06-12
Remove table data on plugin uninstall

### LINKED ISSUE #2361: Deleting plugin should remove data
**Describe the bug**

When WooCommerce Admin is deleted, data tables created by the plugin remain, including associated data.

**To Reproduce**

1. Make sure the site has WooCommerce Admin installed and at least one order.
2. See data is populated, eg `SELECT COUNT(*) FROM wp_wc_order_stats` is non-zero.
3. Deactivate and delete the plugin.
4. See data is still there.

**Expected behavior**

Data tables should remain after deletion, but have not rows of data. Eg, `SELECT COUNT(*) FROM wp_wc_order_stats` should be zero.

joshuatf:
Fixes #2361 

Removes all wc-admin table data when the plugin is uninstalled.

1. Activate wc-admin if not already active.
2. Make sure you have some table data.
3. Uninstall (delete) the plugin.
4. Note the tables are empty.

### Changelog Note:

Fix: Remove all WooCommerce Admin table data on plugin deactivation


Aljullu (approved):
Testing and looking good. :shipit:

PR #2408, opened 2019-06-12
[WIP] Rest api refactor + package usage

mikejolley:
Second attempt at this which converts all REPORT endpoints to a private `wc-admin` namespace instead of `v4` API. This is due to it using data-stores only found in wc-admin plugin.

There was quite a large chunk of work to do so looking at commits may be easier.

- https://github.com/woocommerce/woocommerce-admin/commit/8ff544f4b95c7167d4b9d2c4ee9a42e646faf514 This brings in the endpoints from the rest-api-package that are used for reports, using a new `wc-admin` REST namespace. Classes themselves are namespaced and named using psr4 for easier autoloading.
- https://github.com/woocommerce/woocommerce-admin/commit/ce8c65684724970f9d079d8540e3573f59286d48 Removes the /api/ directory and all old API classes.
- https://github.com/woocommerce/woocommerce-admin/commit/edcf8919e03540bfcfaa1f20be9b38a13658f3ee Imports the REST API Package so wc-admin has access to v4 endpoints. This does require the Core WooCommerce https://github.com/woocommerce/woocommerce/pull/23797 branch to work.
- https://github.com/woocommerce/woocommerce-admin/commit/02840646341869ef448159c97195c626c86ce846 Implements a class autoloader so we don't need to require everything manually. (also https://github.com/woocommerce/woocommerce-admin/commit/356cc7d36ac93d77b118fce0550f24133fb64dd2 and https://github.com/woocommerce/woocommerce-admin/commit/b000e5ef16b211d8a7259a85dba5a9aea92f91e4) It uses a classmap for older non-psr-4 class names.
- https://github.com/woocommerce/woocommerce-admin/commit/1631c1e39163e5f4ee7d9ffce426f4c6b0f2dffb, https://github.com/woocommerce/woocommerce-admin/commit/ae1c1e07c0d785adaab781a3235df2120ef6effd, and https://github.com/woocommerce/woocommerce-admin/commit/7d51862ac5679303edc1523da2145a6b9adff795 replace `v4/` endpoints with the new `wc-admin/` endpoint across JS files.


psealock:
@mikejolley The changes attempted here are a non issue now, is that right? 

>This is due to it using data-stores only found in wc-admin plugin.

If I understand correctly, when the `src/API` folder is moved to the REST API repo, the data stores will go along with it.

Is the private `wc-admin` namespace still a good idea?

cc @rrennick

mikejolley:
@psealock This is likely stale now, but I still feel using private namespaces is a good idea unless we're confident the new endpoints are stable enough for general developer usage. It gives us more control and the opportunity to iterate quickly without worrying about versioning and 3rd party usage.

psealock:
>It gives us more control and the opportunity to iterate quickly without worrying about versioning and 3rd party usage.

I like the sound of that.

rrennick:
Closing in favour of #3204

PR #2412, opened 2019-06-13
Add changelog script

psealock:
Add a script to automate creation of the changelog based on the project board.

### Features

* A type derived from the `[Type]` label (defaults to `dev`)
* Title
* Pull request number
* Additional labels
* Shoutout to PR author if they are not part of the WooCommerce org on github

### Test
1. `npm install` to get added dependencies.
1. Get a Github api token https://github.com/settings/tokens
2. export variable `GH_API_TOKEN` from bash profile
2. `npm run changelog`

### DIFF
--- a/bin/changelog.js
+++ b/bin/changelog.js
@@ -1,0 +1,141 @@
+#!/usr/bin/env node
+'use strict';
+
+const requestPromise = require('request-promise');
+const chalk = require( 'chalk' );
+const octokit = require( '@octokit/rest' )();
+const promptly = require('promptly');
+
+const headers = {
+	'Content-Type': 'application/json;charset=UTF-8',
+	'Authorization': `token ${process.env.GH_API_TOKEN}`,
+	'Accept': 'application/vnd.github.inertia-preview+json',
+	'User-Agent': 'request'
+};
+
+const columnIds = [];
+
+const getPullRequestType = labels => {
+	const typeLabel = labels.find( label => label.name.includes( '[Type]' ) );
+	if ( ! typeLabel ) {
+		return 'Dev';
+	}
+	return typeLabel.name.replace( '[Type] ', '' );
+};
+
+const getLabels = labels => {
+	return labels
+		.filter( label => ! /\[.*\]/.test( label.name ) )
+		.map( label => label.name )
+		.join( ', ' );
+
+};
+const isCollaborator = async ( username ) => {
+	return requestPromise( {
+		url: `https://api.github.com/orgs/woocommerce/members/${ username }`,
+		headers,
+		resolveWithFullResponse: true
+	} ).then( response => {
+		return response.statusCode === 204;
+	} )
+		.catch( err => {
+			if ( err.statusCode !== 404 ) {
+				console.log( '🤯' );
+				console.log( err.message );
+			}
+		});
+}
+
+const writeEntry = async ( content_url ) => {
+	const options = {
+		url: content_url,
+		headers,
+		json: true
+	}
+	return requestPromise( options )
+		.then( async data => {
+			if ( data.pull_request ) {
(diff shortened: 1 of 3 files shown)


bin/changelog.js
```diff
@@ -0,0 +1,137 @@
...
+				columnIds.push( column.id.toString() );
+				console.log( '| ' + chalk.green( column.id ) +  ' | ' + column.name );
+			} );
+			console.log( '+---------+-----------------------------' );
+			console.log( ' ' );
+		} );
+}
+
+( async () => {
+	console.log( chalk.yellow( 'This program requires an api token. You can create one here: ' ) + 'https://github.com/settings/tokens' );
```
joshuatf: Should we make any recommendations on the scope for the token?

Screen Shot 2019-06-13 at 6 22 18 PM

bin/changelog.js
```diff
@@ -0,0 +1,137 @@
...
+
+const makeChangelog = async column_id => {
+	octokit.paginate(
+		'GET /projects/columns/:column_id/cards',
+		{ headers, column_id },
+		response => response.data.forEach( async card => {
+			await writeEntry( card.content_url );
+		} )
+	).catch( err => {
+		console.log( '🤯' );
```
joshuatf: 🤯 💯

bin/changelog.js
```diff
@@ -0,0 +1,137 @@
...
+		json: true
+	}
+	return requestPromise( options )
+		.then( async data => {
+			if ( data.pull_request ) {
+				const collaborator = await isCollaborator( data.user.login );
+				const type = getPullRequestType( data.labels );
+				const labels = getLabels( data.labels );
+				const labelTag = labels.length ? `(${ labels })` : '';
+				const authorTag = !! collaborator ? '' : `👏 @${ data.user.login }`;
```
joshuatf: Awesome 👏

bin/changelog.js
```diff
@@ -0,0 +1,137 @@
...
+				const entry = `- ${ type }: ${ data.title } #${ data.number } ${ labelTag } ${ authorTag }`;
```
joshuatf: Would it be possible to parse the PR for the suggested changelog entry in https://github.com/woocommerce/woocommerce-admin/blob/master/.github/PULL_REQUEST_TEMPLATE.md?

I was thinking logic along the lines of: look for text following `### Changelog Note:` that contains a prefix of `Tweak:|Dev:|Enhancement:...` and use that text if provided.

bin/changelog.js
Aljullu: If I'm reading this correctly, `collaborator` will always be a boolean or `undefined`, right? Do you think we could get rid of the double negation or am I missing something?
```suggestion
				const authorTag = collaborator ? '' : `👏 @${ data.user.login }`;
```

psealock: 👍

bin/changelog.js
psealock: It would be possible. I like using the title because (besides avoiding regex, which I ❤️) it encourages a short description. 

I'm open to it, however.

bin/changelog.js
psealock: Great idea, done.

bin/changelog.js
psealock: Hmm, yeah here is a good example of a changelog note having a much different tone than the title: https://github.com/woocommerce/woocommerce-admin/pull/2423

How about this (in a folllow up): We change the template to prompt for an optional changelog entry and its required format and let the PR author know that otherwise the title will be used. Then we can scrape the description for a changelog entry in the changelog script.

joshuatf: Sounds like a great way of handling it!

Issue #2421, opened 2019-06-13
Customers Report has default date range applied


jeffstieler:
**Describe the bug**

The Customers report has the default date range applied to it, which also interferes with the advanced filters. We were considering it a "non time related" report as of #2225.

This is likely to affect the Stock report as well.

**To Reproduce**

1. Go to Analytics > Settings
2. Set your default report date range to "today" and save
3. Go to Analytics > Customers
4. Verify that only customers registered today are displayed
5. Add an advanced filter for registration date
6. Select before or after any date
7. Use Network tab in Developer Tools to verify that the outbound request still has a default date value for the opposite range (`before` or `after`)

**Expected behavior**

I expect only the advanced filter to modify the date range for the Customers report.

**Additional context**

I believe the regression occurred here: https://github.com/woocommerce/woocommerce-admin/pull/2292/commits/d8809418930c1ad279b5fd5aaa873e654256dc09#diff-421b158400874b347758f137ec615452R277

PR #2422, opened 2019-06-13
Table queries: remove befor/after from no interval endpoints

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/2421

Default query arguments `before` and `after` were added to table queries in the Customer and Stock Reports. This PR removes those arguments for those tables by using the same logic that was already being applied to the `/stats` endpoints.

1. Go to Analytics > Customers
4. Verify that all customers are displayed
5. In the Network tab verify that endpoints `reports/customers` and `reports/customers/stats` do not have `before` or `after` query arguments.
5. Add an advanced filter for registration date
6. Select before or after any date
7. Use Network tab to verify that endpoints `reports/customers` and `reports/customers/stats` do not have `before` or `after` query arguments, only `registered_before` or `registered_after`.

### DIFF
--- a/client/wc-api/reports/utils.js
+++ b/client/wc-api/reports/utils.js
@@ -342,11 +342,13 @@ export function getReportTableQuery( options ) {
 	const filterQuery = getFilterQuery( options );
 	const datesFromQuery = getCurrentDates( query );
 
+	const noIntervals = includes( noIntervalEndpoints, options.endpoint );
+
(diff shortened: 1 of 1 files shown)


psealock:
I don't think so, but if we go that route we'll just need to remember to remove `customers` from the `noIntervals` array. Thanks for the review.

PR #2424, opened 2019-06-13
Add test to catch missing version updates.

jeffstieler:
This PR seeks to add a test to catch missing version number bumps. The intent is to catch missing version number changes when the `release/x` PR is opened.

- Verify tests pass
- Change either the version in `package.json`, `woocommerce-admin.php`, or the `WC_ADMIN_VERSION_NUMBER` constant.
- Verify test fails

### DIFF
--- a/bin/update-version.php
+++ b/bin/update-version.php
@@ -14,8 +14,8 @@ foreach ( $plugin_file as $line ) {
 	if ( stripos( $line, ' * Version: ' ) !== false ) {
 		$line = " * Version: {$package->version}\n";
 	}
-	if ( stripos( $line, "\tdefine( 'WC_ADMIN_VERSION_NUMBER'," ) !== false ) {
-		$line = "\tdefine( 'WC_ADMIN_VERSION_NUMBER', '{$package->version}' );\n";
+	if ( stripos( $line, ">define( 'WC_ADMIN_VERSION_NUMBER'," ) !== false ) {
+		$line = "\t\t\$this->define( 'WC_ADMIN_VERSION_NUMBER', '{$package->version}' );\n";
 	}
 	$lines[] = $line;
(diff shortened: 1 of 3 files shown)


jeffstieler:
> While you're here, can you update the logic responsible for propagating the version number?

Done in f3cc1fd.

bin/update-version.php
```diff
@@ -14,8 +14,8 @@
 	if ( stripos( $line, ' * Version: ' ) !== false ) {
 		$line = " * Version: {$package->version}\n";
 	}
-	if ( stripos( $line, "\tdefine( 'WC_ADMIN_VERSION_NUMBER'," ) !== false ) {
-		$line = "\tdefine( 'WC_ADMIN_VERSION_NUMBER', '{$package->version}' );\n";
+	if ( stripos( $line, ">define( 'WC_ADMIN_VERSION_NUMBER'," ) !== false ) {
+		$line = "\t\t\$this->define( 'WC_ADMIN_VERSION_NUMBER', '{$package->version}' );\n";
```
psealock: Can we future proof this a bit more by manipulating a constant on the class? Or even just adding a comment above this line alerting anyone who refactors this code about the automation could be sufficient. This was the source of the issue because an extra tab was added during a refactor.

jeffstieler: I went the comment route in 449c274 since the version constant is used in a few places outside of the class: [#](https://github.com/woocommerce/wc-admin/blob/b493cc1c2ef51965084685703755ec7521f0efa5/includes/class-wc-admin-install.php#L45), [#](https://github.com/woocommerce/wc-admin/blob/d8809418930c1ad279b5fd5aaa873e654256dc09/includes/class-wc-admin-loader.php#L113).

Issue #2367, opened 2019-06-05
Action Scheduler: Jobs overloading server


psealock:
**Describe the bug**

An endless loop of two AS jobs have crashed a Bluehost hosted site.

>In regards to your issues with your site facim.org. We found that there seems to have been 2 processes competing with each other for resources that overloaded the server bringing your sites down. We can see that the processes were caused by your cron jobs and the Woocommerce report generator. We recommend optimization of your sites or a limit placed on the 2 processes so that it no longer affects your server.

p1559771077006800-slack-team-isotope

Potentially related to https://github.com/woocommerce/woocommerce-admin/issues/2363, which involves failed jobs re-inserting themselves into the queue.

**To Reproduce**

I'm not sure at the moment, but am hoping to get a look at `debug.log`. Here is a provided snippet:

```
Tue Jun 04 07:59:30.337025 2019] [core:error] [pid 17456:tid 47566796625664] [client 172.69.33.8:0] End of script output before headers: index.php, referer: https://members.facim.org/wp-cron.php?doing_wp_cron=1559660352.0508420467376708984375
[Tue Jun 04 09:05:55.848451 2019] [core:error] [pid 17557:tid 47566781916928] [client 162.158.246.91:0] Script timed out before returning headers: index.php, referer: https://members.facim.org/wp-admin/admin.php?page=wc-reports&tab=customers&range=year
```

geeklore:
Customer added more to the ticket to pass along.

```
I contacted BlueHost and this was the only additional information that they could provide me about the crash. I hope this helps to identify the issue.

---------------- 
script output before headers: index.php, referer: https://members.facim.org/wp-cron.php?doing_wp_cron=1559660352.0508420467376708984375

[Tue Jun 04 09:05:55.848451 2019] [core:error] [pid 17557:tid 47566781916928] [client 162.158.246.91:0] Script timed out before returning headers: index.php, referer: https://members.facim.org/wp-admin/admin.php?page=wc-reports&tab=customers&range=year

previous ticket was escalated per advisement from Cameron Daniel 14485008. Checked processes and they are at 0 checked Top and load is about 1 plenty of free ram. customer says this is intermittent. sar shows nothing. logs show errors from the 4th

[04-Jun-2019 16:05:20 UTC] PHP Warning: mysqli_query(): MySQL server has gone away in /home/facimorg/public_html/wp-includes/wp-db.php 
on line 2007 
[...]

Issue #2363, opened 2019-06-05
Fatal Error: Unhandled wc_get_order returning false


psealock:
**Describe the bug**

There are several instances of unexpected returned values of `wc_get_order` function calls.

This one can create the following fatal error. It should return `-1` for no order found.

https://github.com/woocommerce/woocommerce-admin/blob/af6a19c7016901cc647a77e83d8a8276a1e52590/includes/data-stores/class-wc-admin-reports-coupons-data-store.php#L359-L366

```
PHP Fatal error: Uncaught Error: Call to a member function get_type() on boolean in \plugins\woocommerce-admin\includes\data-stores\class-wc-admin-reports-coupons-data-store.php:362
```
Another potential instance here,

https://github.com/woocommerce/woocommerce-admin/blob/e79d8e058bd920f7f02d721d7f279e08ef1c0c47/includes/api/class-wc-admin-rest-reports-controller.php#L184

And there are two additional potential problems reported in this support comment, https://wordpress.org/support/topic/php-fatal-error-call-to-a-member-function-get_type-on-boolean/

**To Reproduce**

I'm not sure the best way to force these errors besides manipulating the code such that `wc_get_order` returns false.

**Expected behavior**

Errors and false returns from `wc_get_order` function calls should be handled gracefully.

NicolasDorier:
We have this problem as well, it goes away as soon as we disable the plugin.

pavlenex:
Happening [here](https://github.com/btcpayserver/woocommerce-plugin/issues/28) as well.

PR #2425, opened 2019-06-14
Action Scheduler: fix potential endless sync

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/2367
Fixes https://github.com/woocommerce/woocommerce-admin/issues/2363

When syncing orders, attempts to write to the database result in 1 or 0 or -1 depending on success. If the sum of the results isn't right, we retry the sync by creating another action scheduler job. 

When a failure is due to something wrong with an order, such a non-existent order, a new job is created which will inevitably also fail, resulting in an endless loop of job creations.

https://github.com/woocommerce/woocommerce-admin/blob/41135fa19f94fd555f811070da10f124dbdfc7f0/includes/class-wc-admin-reports-sync.php#L454-L470

This PR moves checks in the syncing functions on the `$order` which don't involve database writes. Any failure of these checks now returns early instead of failing at the write step, causing another AS job which will also fail.

See occurrences of this issue in the wild:
https://wordpress.org/support/topic/php-fatal-error-call-to-a-member-function-get_type-on-boolean/
https://wordpress.org/support/topic/over-170-000-scheduled-action-created/

1. Create an order with taxes and a coupon.
2. See the order get synced when the order is reflected on the Revenue, Taxes, and Coupons reports.
3. Create a refund.
4. See the order synced by viewing the Revenue report.

### DIFF
--- a/includes/class-wc-admin-reports-sync.php
+++ b/includes/class-wc-admin-reports-sync.php
@@ -432,6 +432,26 @@ class WC_Admin_Reports_Sync {
 	 * @return void
 	 */
 	public static function orders_lookup_import_order( $order_id ) {
+
+		$order = wc_get_order( $order_id );
+
+		// If the order isn't found for some reason, skip the sync.
+		if ( ! $order ) {
+			return;
+		}
+
+		$type = $order->get_type();
+
+		// If the order isn't the right type, skip sync.
+		if ( 'shop_order' !== $type && 'shop_order_refund' !== $type ) {
+			return;
+		}
+
+		// If the order has no id or date created, skip sync.
+		if ( ! $order->get_id() || ! $order->get_date_created() ) {
+			return;
+		}
+
 		$result = array_sum(
 			array(
 				WC_Admin_Reports_Orders_Stats_Data_Store::sync_order( $order_id ),
--- a/includes/data-stores/class-wc-admin-reports-coupons-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-coupons-data-store.php
@@ -358,6 +358,10 @@ class WC_Admin_Reports_Coupons_Data_Store extends WC_Admin_Reports_Data_Store im
 
 		$order = wc_get_order( $order_id );
 
+		if ( ! $order ) {
+			return -1;
+		}
+
 		// Refunds don't affect coupon stats so return successfully if one is called here.
 		if ( 'shop_order_refund' === $order->get_type() ) {
 			return true;
--- a/includes/data-stores/class-wc-admin-reports-taxes-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-taxes-data-store.php
@@ -311,7 +311,12 @@ class WC_Admin_Reports_Taxes_Data_Store extends WC_Admin_Reports_Data_Store impl
 	 */
 	public static function sync_order_taxes( $order_id ) {
 		global $wpdb;
-		$order       = wc_get_order( $order_id );
+
+		$order = wc_get_order( $order_id );
+		if ( ! $order ) {
+			return -1;
+		}
+
 		$tax_items   = $order->get_items( 'tax' );
 		$num_updated = 0;
 


rrennick:
@psealock I edited the PR description to change the related issue from #2363 to #2367.

includes/class-wc-admin-reports-sync.php
```diff
@@ -432,12 +432,31 @@ public static function orders_lookup_import_batch( $batch_number, $days, $skip_e
 	 * @return void
 	 */
 	public static function orders_lookup_import_order( $order_id ) {
+
+		$order = wc_get_order( $order_id );
+		$type  = $order->get_type();
```
rrennick: This shouldn't be called before the check that it's a valid order.

includes/data-stores/class-wc-admin-reports-coupons-data-store.php
```diff
@@ -350,14 +350,12 @@ protected function get_cache_key( $params ) {
 	 * Create or update an an entry in the wc_order_coupon_lookup table for an order.
 	 *
 	 * @since 3.5.0
-	 * @param int $order_id Order ID.
+	 * @param object $order Order.
 	 * @return int|bool Returns -1 if order won't be processed, or a boolean indicating processing success.
 	 */
-	public static function sync_order_coupons( $order_id ) {
+	public static function sync_order_coupons( $order ) {
```
rrennick: Since this is a public function it should have a check to ensure that the passed parameter is an order object.

includes/data-stores/class-wc-admin-reports-products-data-store.php
```diff
@@ -381,17 +381,12 @@ protected function get_cache_key( $params ) {
 	 * Create or update an entry in the wc_admin_order_product_lookup table for an order.
 	 *
 	 * @since 3.5.0
-	 * @param int $order_id Order ID.
+	 * @param object $order Order.
 	 * @return int|bool Returns -1 if order won't be processed, or a boolean indicating processing success.
 	 */
-	public static function sync_order_products( $order_id ) {
+	public static function sync_order_products( $order ) {
```
rrennick: Same comment on public function parameter validation.

includes/data-stores/class-wc-admin-reports-taxes-data-store.php
```diff
@@ -306,12 +306,11 @@ protected function normalize_order_by( $order_by ) {
 	/**
 	 * Create or update an entry in the wc_order_tax_lookup table for an order.
 	 *
-	 * @param int $order_id Order ID.
+	 * @param object $order Order.
 	 * @return int|bool Returns -1 if order won't be processed, or a boolean indicating processing success.
 	 */
-	public static function sync_order_taxes( $order_id ) {
+	public static function sync_order_taxes( $order ) {
```
rrennick: Same comment on public function parameter validation.

includes/class-wc-admin-reports-sync.php
psealock: Good point. Done in https://github.com/woocommerce/woocommerce-admin/pull/2425/commits/810088e2632744f369435aed0621a1db848b77df

psealock:
Thanks for the review @rrennick.

Because the individual sync methods are public, I went back to their original logic and added a couple checks for `$order` where they were missing.

```php
if ( ! $order ) {
    return -1;
}
```

So the paradigm is still the same as it was originally but we now have checks and early exits on the sync function that aggregates the results and fires off another retry AS job.

includes/class-wc-admin-reports-sync.php
```diff
@@ -432,6 +432,26 @@ public static function orders_lookup_import_batch( $batch_number, $days, $skip_e
 	 * @return void
 	 */
 	public static function orders_lookup_import_order( $order_id ) {
+
+		$order = wc_get_order( $order_id );
```
joshuatf: I think this is probably a good way of handling things.

But I'm also curious if you ran across any specific cases/context where we were able to get an order through `get_orders()` that resulted in `wc_get_order()` returning `false`.

psealock: yes:

https://wordpress.org/support/topic/php-fatal-error-call-to-a-member-function-get_type-on-boolean/

joshuatf: @psealock Maybe I'm missing something in that post, but I'm not seeing the reason why `wc_get_order()` returned false for the order ID.

It doesn't invalidate this PR, just curious if there's something else happening that we should be aware of.

psealock: Good point. This will be something to keep an eye on. Looking at the `get_orders` function in the same file doesn't look like it would produce anything odd.

PR #2428, opened 2019-06-14
Prevent duplicate customer imports

### LINKED ISSUE #2426: Historical import returns many empty customer rows
**Describe the bug**
Importing regular customers correctly imports 40/40 customers for my data set.  After the orders are imported (where guests and non-customers may also be added), I'm showing 397 rows of customers.

The store only contains 284 orders and most of the newly created customers don't seem to be linked to any orders.

**To Reproduce**

1. Delete all data.
2. Re-import all historical data.
3. Note extra customers.

**Expected behavior**
Expected only to see customers with orders or users with a role of "customer."

**Screenshots**
Screen Shot 2019-06-14 at 4 44 42 PM
Screen Shot 2019-06-14 at 4 53 49 PM

joshuatf:
Fixes #2426

Gets the existing customer ID from order stats row in the event an order has already been saved or if an email is blank.

1.  Make sure you have some guest orders (preferably without emails).
2. Delete all data and re-run the imports.
3. Make sure the import runs smoothly and customer accounts are connected with order stats.
4.  Save an existing guest order and make sure a duplicate customer is not created.

### Changelog Note:

Fix: Prevent duplicate guest customer entries when email is missing.

### DIFF
--- a/includes/data-stores/class-wc-admin-reports-customers-data-store.php
+++ b/includes/data-stores/class-wc-admin-reports-customers-data-store.php
@@ -442,9 +442,21 @@ class WC_Admin_Reports_Customers_Data_Store extends WC_Admin_Reports_Data_Store 
 			return false;
 		}
 
+		global $wpdb;
 		$user_id = $order->get_customer_id();
 
 		if ( 0 === $user_id ) {
+			$customer_id = $wpdb->get_var(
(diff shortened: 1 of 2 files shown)


timmyc (approved):
Confirmed import still works as expected, and no duplicates are created when updating guest orders.

PR #2431, opened 2019-06-14
Fix conflicts with WordPress.com toolbar

### LINKED ISSUE #1991: WordPress.com toolbar conflicts with activity header on small screens
On small screens, with the WordPress.com toolbar enabled, I see the following:

It looks like there may been some regressions/changes, because we tried to fix this once before: https://github.com/woocommerce/woocommerce-admin/pull/74, https://github.com/woocommerce/woocommerce-admin/issues/72.

Steps to reproduce:
* Enable the WordPress.com toolbar under `Jetpack > Settings`
* View any WooCommerce page with the activity header displaying

Aljullu:
Fixes #1991.

### Screenshots
WordPress.com toolbar<br>Before | WordPress.com toolbar<br>After
:--------:|:--------:
Screen Shot 2019-06-14 at 15 59 53 | Screen Shot 2019-06-14 at 16 00 00

- Go to _Jetpack_ > _Settings_ > _Writting_ and click on _Enable the WordPress.com toolbar_.
- Navigate your site with a narrow viewport: WooCommerce pages, WooCommerce Admin pages and non-WooCommerce pages.
- Verify the WordPress.com toolbar looks good in all of them and there are no overflowing elements in the page.

### DIFF
--- a/client/stylesheets/shared/_reset.scss
+++ b/client/stylesheets/shared/_reset.scss
@@ -20,7 +20,9 @@
 
 	@include breakpoint( '<782px' ) {
 		// WP breakpoint for mobile menu
+		.wp-responsive-open #woocommerce-embedded-root,
 		.wp-responsive-open #wpbody {
+			position: relative;
 			right: -14.5em;
 		}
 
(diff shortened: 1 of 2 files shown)


client/stylesheets/shared/_reset.scss
```diff
@@ -64,95 +66,42 @@
...
 		position: fixed;
 	}
 }
 
-#adminmenuwrap {
-	z-index: 99992;
+.auto-fold #adminmenu,
+.auto-fold #adminmenuback,
+.auto-fold #adminmenuwrap {
+	z-index: 99992; /* higher than woocommerce-layout__header (99991) */
```
jeffstieler: Is it possible to fix this by lowering the `z-index` of `.woocommerce-layout__header` instead? There are reports of the sticky nav interfering with other UI elements as well: https://wordpress.org/support/topic/terrible-plugin-doesnt-play-with-woocommerce-partial-orders-or-publishpress/#post-11634003

Aljullu: Sounds like a good solution! I decreased it here: 1cd15f56e701d3169dfed9c11038e6e0a806ea32.

I had to overwrite the z-index of `components-popover` to avoid this:
Peek 2019-06-14 16-47

client/stylesheets/shared/_reset.scss
```diff
@@ -64,95 +66,40 @@
...
+		}
 	}
 
-	@include breakpoint( '<782px' ) {
-		.woocommerce-layout__activity-panel-wrapper.is-open,
-		.woocommerce-layout__activity-panel-tabs {
-			width: calc(100vw - 190px);
-		}
+	.components-popover:not(.is-mobile) {
+		z-index: 1 !important; /* below of woocommerce-layout__header */
```
jeffstieler: Is the `!important` necessary here? I think this selector will already be more specific that the Gutenberg one.

Aljullu: Yes, good catch! Fixed in 825f76e13c154a0afc08a2529c72ca6e736beda2.

Issue #2430, opened 2019-06-14
Add product ID Column to Products and Stock Report


apkruger:
This is **NOT** a Bug or Problem but a **feature request**.

**Problem:**
We have a store with lots of products and variations and have lots of different providers we purchase our products from, some of the providers offer a similar "Name/SKU" but the product is different so we use the Product "ID" to calculate sales or do product stock reports so that we can re-order or workout which providers product sells better.

**Requested Option:**
Will it be possible to add the "ID" Column as an Option to display/download in the below reports.

- Products Report:
Under the Products report, an admin must also be able to select the "ID" (Column) as an additional field to display when viewing or when downloading the report.

- Stock Report:
Under the Stock report, an admin must also be able to select the "ID" (Column) as an additional field to display when viewing or downloading the report.

**Current Method we are using**
We currently have to extract the products and variations in different methods and using vlookup to generate our reports. It would be a great feature for us if this could be added.

**Could this be a bigger Feature for more people?**
Maybe this feature could be extended and offer the Woocommerce admin an option to add a few extra Default Database Columns to any report during the Analytics installation "Setup" or "Settings". 

**Appreciation:**
I know this is a low down feature in your priorities but appreciate the effort in the already great Woocommerce Admin Interface, this is really making our lives easier. Great Work!!
![add-id-as-option-to-columns](https://user-images.githubusercontent.com/44058773/59506015-72e7fe80-8ea7-11e9-893b-d09e5aa030e9.png)
![id-for-products-report](https://user-images.githubusercontent.com/44058773/59506047-898e5580-8ea7-11e9-8e60-f86838f335b8.png)
![id-for-stock-report](https://user-images.githubusercontent.com/44058773/59506055-90b56380-8ea7-11e9-90c8-7fbca2f79189.png)

psealock:
Thanks for writing this feature request @apkruger. Including the ID is not currently on our design roadmap. However, the tables are missing the ability to be extended, which will allow you to add columns of your choosing.

I'm moving this issue to the backlog so that we can address the extensibility issue and provide an example of how you might add an ID column.

apkruger:
@psealock thank you very much for your feedback, appreciate your time. Looking forward to using the table extended feature and sample extension for the future :)

apkruger:
[...]

PR #2440, opened 2019-06-17
Report Tables: Add example extension

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/2430

Allow the report tables to be extended by adding a column and designating column contents.

This PR creates an extension example, WooCommerce Admin Table Column Example, to illustrate how to add ID and rating columns.

### Screenshots

Screen Shot 2019-06-17 at 4 39 55 PM

Screen Shot 2019-06-17 at 4 40 15 PM

1. Make sure some of your products have a rating.
2. `npm run example -- --ext=table-column`.
3. Activate the WooCommerce Admin Table Column Example plugin
4. Go to Products Report.
5. See the table has columns ID and rating, which can be toggled on/off.
6. Make sure the other reports load correctly.

### DIFF
--- a/packages/components/src/rating/style.scss
+++ b/packages/components/src/rating/style.scss
@@ -5,6 +5,7 @@
 	vertical-align: middle;
 	display: inline-block;
 	overflow: hidden;
+	white-space: nowrap;
 
 	.gridicon {
 		fill: $core-grey-light-600;
--- a/client/analytics/components/report-table/index.js
+++ b/client/analytics/components/report-table/index.js
@@ -106,13 +106,27 @@ class ReportTable extends Component {
 		const totals = get( primaryData, [ 'data', 'totals' ], {} );
 		const totalResults = items.totalResults;
 		const downloadable = 0 < totalResults;
+
+		/**
+		 * Filter report table.
+		 *
+		 * Enables manipulation of data used to create a report table.
+		 *
+		 * @param {object} reportTableData - data used to create the table.
+		 * @param {string} reportTableData.endpoint - table api endpoint.
+		 * @param {array} reportTableData.headers - table headers data.
+		 * @param {array} reportTableData.rows - table rows data.
+		 * @param {object} reportTableData.totals - total aggregates for request.
+		 * @param {array} reportTableData.summary - summary numbers data.
+		 * @param {array} reportTableData.items - response from api requerst.
+		 */
 		const { headers, ids, rows, summary } = applyFilters( TABLE_FILTER, {
-			endpoint: endpoint,
+			endpoint,
 			headers: getHeadersContent(),
-			ids: itemIdField ? items.data.map( item => item[ itemIdField ] ) : [],
 			rows: getRowsContent( items.data ),
-			totals: totals,
+			totals,
 			summary: getSummary ? getSummary( totals, totalResults ) : null,
+			items,
 		} );
 
 		// Hide any headers based on user prefs, if loaded.
--- a/docs/examples/extensions/table-column/js/index.js
+++ b/docs/examples/extensions/table-column/js/index.js
@@ -1,0 +1,47 @@
+/** @format */
+/**
+ * External dependencies
+ */
+
+import { addFilter } from '@wordpress/hooks';
+import { Rating } from '@woocommerce/components';
+
+addFilter( 'woocommerce_admin_report_table', 'plugin-domain', reportTableData => {
+	if ( 'products' !== reportTableData.endpoint || ! reportTableData.items.data.length ) {
+		return reportTableData;
+	}
+
+	const newHeaders = [
+		...reportTableData.headers,
+		{
+			label: 'ID',
+			key: 'product_id',
+		},
+		{
+			label: 'Rating',
+			key: 'product_rating',
+		},
+	];
+	const newRows = reportTableData.rows.map( ( row, index ) => {
+		const product = reportTableData.items.data[ index ];
+		const newRow = [
+			...row,
+			// product_id is already returned in the response for productData.
+			{
+				display: product.product_id,
+				value: product.product_id,
+			},
+			// average_rating can be found on extended_info on productData.
+			{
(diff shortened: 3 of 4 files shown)


psealock:
@joshuatf absolutely! I thought this existed, but didn't think to look in the `<ReportTable />` component. Seems like its time to start thinking about generated documentation around filters like @justinshreve mentions.

psealock:
>Does this PR add something that can't be done with that?

Upon closer inspection, yes indeed it does. The filter you mentioned allows manipulation of the table properties after they have been processed by `getHeadersContent()` and `getRowsContent()`. Which leaves the extension developer a lump of data to deal with. 

https://github.com/woocommerce/woocommerce-admin/blob/8a4eab761aab38d0e9e12fb828924c47e6a291b5/client/analytics/components/report-table/index.js#L109-L116

The filters proposed here allow the developer access to each row of returned API data at the time of creating a table row. In other words, access to `data[ 0 ]` and `data[ 0 ].extended_info`. This prevents the need for an additional api request.

Screen Shot 2019-06-19 at 12 07 56 PM

In fact, it could be argued that the previous table filter should be removed because it doesn't provide much utility.

psealock:
@joshuatf Does that make sense why the filters added here offer added utility?

@justinshreve Thanks for the review, I added doc blocks and fixed the typo.

joshuatf:
@psealock That makes sense in terms of being able to access the `extended_info`.  Could we add the raw item data to the `TABLE_FILTER` so we have access to `items.data` through the table hook?

https://github.com/woocommerce/woocommerce-admin/blob/d46c853cc208acbb4da55d2516e0340548362b91/client/analytics/components/report-table/index.js#L109

```js
		const { headers, ids, rows, summary } = applyFilters( TABLE_FILTER, {
			endpoint: endpoint,
			headers: getHeadersContent(),
			ids: itemIdField ? items.data.map( item => item[ itemIdField ] ) : [],
			rows: getRowsContent( items.data ),
			totals: totals,
			summary: getSummary ? getSummary( totals, totalResults ) : null,
			items: items,
		} );
```

packages/components/src/rating/style.scss
```diff
@@ -5,6 +5,7 @@
 	vertical-align: middle;
```
joshuatf: One other small thing we could take care of while we're here to vertically align the ratings in tables:

```suggestion
	display: flex;
	align-items: middle;
	justify-content: flex-end;
```

Screen Shot 2019-06-19 at 9 19 37 AM


Not sure if this should only be applied to ratings within tables.

psealock:
>Could we add the raw item data to the TABLE_FILTER so we have access to items.data through the table hook?

Yeah, good point we can. An extension adding an ID column would use code like this, which works btw:

```js
addFilter( 'woocommerce_admin_report_table', 'plugin-domain', tableData => {
	if ( 'products' === tableData.endpoint ) {
		if ( tableData.items.data.length ) {
			const newHeaders = [
				...tableData.headers,
				{
					label: 'ID',
					key: 'product_id',
				},
			];
			const newRows = tableData.rows.map( ( row, index ) => {
				const product = tableData.items.data[ index ];
				const newRow = [
					...row,
					{
						display: product.product_id,
						value: product.product_id,
					},
				];
				return newRow;
			} );
			tableData.headers = newHeaders;
			tableData.rows = newRows;
			return tableData;
		}
		
		return tableData;
	}

	return tableData;
} );
```

I suppose its a matter of preference. The implementation for extensions is much cleaner the other way, but the existing filter is a single point of entry.

One possible hickup with the above code is the assumption that the index of the each dataset corresponds. Knowing the code, I think its a safe assumption, but we'd need to convey that information.

```js
const product = tableData.items.data[ index ];
```

I'm open to feedback on this one. What do you think?

packages/components/src/rating/style.scss
psealock: This breaks the review panel

Screen Shot 2019-06-19 at 2 07 00 PM

It seems vertically aligned as is in Chrome. Are you seeing something different?

Screen Shot 2019-06-19 at 2 07 58 PM

joshuatf:
>An extension adding an ID column would use code like this, which works btw:

Early feedback, but we can probably clean this up just a bit by returning early:

```js
if ( 'products' !== tableData.endpoint || ! tableData.items.data.length ) {
	return tableData;
}
```

>One possible hickup with the above code is the assumption that the index of the each dataset corresponds. Knowing the code, I think its a safe assumption, but we'd need to convey that information.

I think it's probably also safe.  Plus it allows removal/reorganization of rows.

>I'm open to feedback on this one. What do you think?

I think the higher level table filter leads by a few benefits:
* Single filter versus 2n filters for each table added.  Less maintenance and error prone on our part if various hooks should data passed in need to change.
* Access to totals, rows, summary, etc.
* Able to compare entire item dataset and re-order or compare data.

I do agree that the single filter looks a bit cleaner, but feel it comes at a cost of extra data, functionality, and maintenance.  If anyone else wants to weigh in though, I'm also open to other opinions here.

packages/components/src/rating/style.scss
joshuatf: I think in your attached screenshot it's slightly off-center as well (just by a few pixels) causing the row height to shift.

We could apply the flexbox fix to only ratings in tables or skip entirely if you think it's not a big deal.

```scss
.woocommerce-table__item .woocommerce-rating {
	display: flex;
	align-items: middle;
	justify-content: flex-end;
}
```

psealock: Still getting oddness on this. Since this example code isn't getting adopted, I think we should skip adding new css rules just in case.

Screen Shot 2019-06-20 at 2 16 19 PM

psealock:
@joshuatf I refactored to use the existing filter. This point makes sense, `Single filter versus 2n filters for each table added`.

Can you give it another look?

timmyc:
@psealock this would be a great thing to do a demo screen cast of - or even a dev blog post too so we can start building up the library of developer documentation for wc-admin. Nice work!

PR #2441, opened 2019-06-17
Remove the "wpClosedMenu" page setting

DanReyLop:
Conceptually, it doesn't make a lot of sense for a page to specify the menu section it's *not* on. That won't scale past 2 sections, and we're already at that point (`WooCommerce`, `Analytics`, and now a new one).

I'm trying to understand the WC-Admin client-side controller, but this seems redundant: https://github.com/woocommerce/woocommerce-admin/blob/172bb3e2074eee4012dcc0c1672fcd61b865aeaa/client/layout/controller.js#L237-L243
All the opened submenus get closed already a few lines above: https://github.com/woocommerce/woocommerce-admin/blob/172bb3e2074eee4012dcc0c1672fcd61b865aeaa/client/layout/controller.js#L209-L216 

I don't see any adverse effects when I removed those lines, so I did. I had to change the logic that updates menu links to iterate on *all* the menu items, not only the open and closed one.

To test: Navigate through the different WC-Admin pages. Navigate thgough those pages with a date filter set, see that it's preserved. I'm not sure what else to test for, I didn't find anything unusual but I could have missed stuff.

Note: This PR branches off `update/route-handling` so I can work locally wothout going crazy. I can rebase it onto `master` if this PR is accepted before `update/route-handling` lands.

### DIFF
--- a/client/layout/controller.js
+++ b/client/layout/controller.js
@@ -31,13 +31,11 @@ const getPages = () => {
 			container: DevDocs,
 			path: '/devdocs',
 			wpOpenMenu: 'toplevel_page_woocommerce',
-			wpClosedMenu: 'toplevel_page_wc-admin--analytics-revenue',
 		} );
 		pages.push( {
 			container: DevDocs,
 			path: '/devdocs/:component',
 			wpOpenMenu: 'toplevel_page_woocommerce',
-			wpClosedMenu: 'toplevel_page_wc-admin--analytics-revenue',
 		} );
 	}
 
@@ -46,7 +44,6 @@ const getPages = () => {
 			container: Dashboard,
 			path: '/',
 			wpOpenMenu: 'toplevel_page_woocommerce',
-			wpClosedMenu: 'toplevel_page_wc-admin--analytics-revenue',
 		} );
 	}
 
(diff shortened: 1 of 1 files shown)


client/layout/controller.js
```diff
@@ -32,13 +32,11 @@ const getPages = () => {
 			container: DevDocs,
 			path: '/devdocs',
 			wpOpenMenu: 'toplevel_page_woocommerce',
-			wpClosedMenu: 'toplevel_page_wc-admin-path--analytics-revenue',
```
joshuatf: Nice!  Agreed that it would be much better without these.

client/layout/controller.js
```diff
@@ -32,13 +32,11 @@ const getPages = () => {
 			container: DevDocs,
 			path: '/devdocs',
 			wpOpenMenu: 'toplevel_page_woocommerce',
```
joshuatf: I'm wondering if we should rethink the logic behind `wpOpenMenu` as well.

client/layout/controller.js
```diff
@@ -196,7 +190,7 @@ window.wpNavMenuUrlUpdate = function( page, query ) {
 	const nextQuery = stringifyQuery( getPersistedQuery( query ) );
 
 	Array.from(
-		document.querySelectorAll( `#${ page.wpOpenMenu } a, #${ page.wpClosedMenu } a` )
+		document.querySelectorAll( '#adminmenu a' )
```
joshuatf: 👍

client/layout/controller.js
DanReyLop: `wpOpenMenu` doesn't seem as bad. It makes sense for a page to indicate which section it is on. Maybe not ideal to have to specify an ID that's auto-generated by WP-Admin, but not terrible.

joshuatf: Ya, certainly not ideal.  I was thinking we could just find and add the open classes to the parent menu item when an item is clicked instead of specifying `wpOpenMenu`.

I'm wondering if there's a use case for using a different selector in `wpOpenMenu` other than the item's parent.  If so, we can leave it as is.

client/layout/controller.js
psealock: 👍

psealock:
Tests are probably failing due to the base branch. I think you can change the base to master and be all good to merge

DanReyLop:
> I noticed that switching between submenus (e.g, from Analytics->Orders to WooCommerce->Dashboard) causes a full page reload. Not sure if that's from the update/route-handling branch, but I'll let @psealock weigh in and give a 👍 if that looks okay.

That works fine when I applied the commit on `master`, and all the tests pass :)

PR #2442, opened 2019-06-17
Remove updated stock products from Activity Panel

### LINKED ISSUE #2175: After Updating Stock - Product Still Remains in Stock Panel
**Describe the bug**
When updating a product that is either out of stock, or below the low stock inventory threshold to a level that is above the low-stock threshold - the product still appears in the Stock Panel until you refresh the page.

**To Reproduce**

1. Set your low stock threshold to 2
2. Edit a product so the stock level is 1
3. Visit the dashboard, note the indicator is shown for the Stock Panel
4. Expand the panel, then modify the stock of the product you updated in step 2 to a number greater than the low stock threshold
5. Note the product is still shown after updating the stock to a value greater than the low stock threshold

**Expected behavior**
After updating inventory to a level above the low stock threshold, I anticipated the product to no longer be shown. The product is no longer shown after you refresh the page.

Aljullu:
Fixes #2175.

### Accessibility

### Screenshots
Screenshot_2019-06-17 Stock ‹ Analytics ‹ WooCommerce — WooCommerce

- Make sure you have some products in 'low stock'.
- Open the _Stock_ tab of the Activity Panel.
- Modify the stock of one of the products so it's no longer in 'low stock'.
- Verify the opacity of the product card is reduced but the card is still visible.
- Close the panel and open it again.
- Verify the card disappeared.

### DIFF
--- a/client/header/activity-panel/panels/stock/card.js
+++ b/client/header/activity-panel/panels/stock/card.js
@@ -37,12 +37,10 @@ class ProductStockCard extends Component {
 		this.updateStock = this.updateStock.bind( this );
 	}
 
-	componentDidUpdate() {
-		this.quantityInput && this.quantityInput.focus();
-	}
-
 	beginEdit() {
-		this.setState( { editing: true } );
+		this.setState( { editing: true }, () => {
+			this.quantityInput && this.quantityInput.focus();
+		} );
 	}
 
 	cancelEdit() {
--- a/client/header/activity-panel/panels/stock/card.js
+++ b/client/header/activity-panel/panels/stock/card.js
@@ -34,19 +34,29 @@ class ProductStockCard extends Component {
 		this.cancelEdit = this.cancelEdit.bind( this );
 		this.onQuantityChange = this.onQuantityChange.bind( this );
 		this.handleKeyDown = this.handleKeyDown.bind( this );
-		this.updateStock = this.updateStock.bind( this );
+		this.onSubmit = this.onSubmit.bind( this );
 	}
 
 	beginEdit() {
-		this.setState( { editing: true }, () => {
-			this.quantityInput && this.quantityInput.focus();
-		} );
+		const { product } = this.props;
+
+		this.setState(
+			{
+				editing: true,
+				quantity: product.stock_quantity,
+			},
+			() => {
+				this.quantityInput && this.quantityInput.focus();
+			}
+		);
 	}
 
 	cancelEdit() {
+		const { product } = this.props;
+
 		this.setState( {
 			editing: false,
-			quantity: this.props.product.stock_quantity,
+			quantity: product.stock_quantity,
 		} );
 	}
 
@@ -60,35 +70,13 @@ class ProductStockCard extends Component {
 		this.setState( { quantity: event.target.value } );
 	}
 
-	async updateStock() {
-		const { addNotice, product, updateItem } = this.props;
+	onSubmit() {
+		const { product, updateProductStock } = this.props;
+		const { quantity } = this.state;
 
 		this.setState( { editing: false, edited: true } );
 
-		const data = {
-			stock_quantity: this.state.quantity,
-			type: product.type,
-			parent_id: product.parent_id,
-		};
-
-		await updateItem( 'products', product.id, data, response => {
-			if ( response && response.data ) {
-				addNotice( {
-					status: 'success',
-					message: sprintf( __( '%s stock updated.', 'woocommerce-admin' ), product.name ),
-				} );
-			}
-			if ( response && response.error ) {
-				addNotice( {
-					status: 'error',
-					message: sprintf(
-						__( '%s stock could not be updated.', 'woocommerce-admin' ),
-						product.name
-					),
-				} );
-				this.setState( { quantity: product.stock_quantity } );
-			}
-		} );
+		updateProductStock( product, quantity );
 	}
 
 	getActions() {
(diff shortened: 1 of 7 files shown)


client/header/activity-panel/panels/stock/card.js
```diff
@@ -117,6 +135,17 @@ class ProductStockCard extends Component {
 
 	render() {
 		const { product } = this.props;
+		const { edited, editing } = this.state;
+		const { notifyLowStockAmount } = wcSettings;
+		const lowStockAmount = Number.isFinite( product.low_stock_amount )
+			? product.low_stock_amount
+			: notifyLowStockAmount;
+		const isLowStock = product.stock_quantity < lowStockAmount;
```
jeffstieler: Should this be `<=` to match https://github.com/woocommerce/woocommerce/blob/a247e08ad0b0fa93b8ece6ecd48b7b4d7ec432db/includes/admin/class-wc-admin-dashboard.php#L221 ?

Aljullu: Good catch! Fixed in 682fe11c269403652397d51779c9dcc993503422.

client/header/activity-panel/panels/stock/card.js
```diff
@@ -117,6 +135,17 @@ class ProductStockCard extends Component {
...
+
+		if ( ! isLowStock && ! edited && ! editing ) {
```
jeffstieler: It looks like this is the logic to hide the card once the stock has been updated and the panel closed/reopened. Can you add a comment to explain that?

client/wc-api/items/mutations.js
```diff
@@ -5,9 +5,14 @@
  */
 import { getResourceName } from '../utils';
 
-const updateItem = operations => ( type, id, itemData ) => {
+const updateItem = operations => async ( type, id, itemData, callback ) => {
```
jeffstieler: What do you think about using Promises here instead of a callback argument?

timmyc:
Is this ready for some more 👀 @Aljullu ?

client/header/activity-panel/panels/stock/card.js
Aljullu: Added in 4536f027a1d26120827834e877e482ab1c02b5ee. :+1:

client/wc-api/items/mutations.js
Aljullu: I took a look at how Gutenberg deals with success/error messages for async actions and it looks like they add the notices directly from the action creator instead of the component (see [savePost](https://github.com/WordPress/gutenberg/blob/master/packages/editor/src/store/actions.js#L258) for example). So I opted for moving all the logic from the component to the mutation in b19643ed72715a045d9114123789d222933a0497, I would like to hear your thoughts on how it looks.

Aljullu:
> Is this ready for some more eyes @Aljullu ?

Yes, I guess so. I was experimenting with integrating `@wordpress/data-controls` into our dependencies, but I guess there is no need to do it now and in this PR, so I think this can be reviewed right away. I updated the labels accordingly.

client/wc-api/items/mutations.js
```diff
@@ -1,15 +1,47 @@
...
+	const oldStockQuantity = product.stock_quantity;
+	const resourceName = getResourceName( 'items-query-products-item', product.id );
+
+	// Optimistically update product stock
+	operations.updateLocally( [ resourceName ], {
+		[ resourceName ]: { ...product, stock_quantity: newStock },
+	} );
+
+	const result = await operations.update( [ resourceName ], {
+		[ resourceName ]: { id: product.id, stock_quantity: newStock },
```
psealock: `type` and ` parent_id` need to be passed in here as well. There is a switch statement in `update` that requires it.

https://github.com/woocommerce/woocommerce-admin/blob/b19643ed72715a045d9114123789d222933a0497/client/wc-api/items/operations.js#L98-L106

client/header/activity-panel/panels/stock/card.js
```diff
@@ -119,14 +124,27 @@ class ProductStockCard extends Component {
...
 	}
 
 	render() {
 		const { product } = this.props;
-		const { editing } = this.state;
+		const { edited, editing } = this.state;
+		const { notifyLowStockAmount } = wcSettings;
+		const lowStockAmount = Number.isFinite( product.low_stock_amount )
+			? product.low_stock_amount
+			: notifyLowStockAmount;
```
psealock: :+1:

client/wc-api/items/mutations.js
Aljullu: Oh, right I forgot that when refactoring :man_facepalming: , thanks for noticing!

It's fixed in 855587a72f9baca96d4f2c4210495b3d3ebb7884.

Aljullu:
Thanks for taking a look @psealock! This is ready for another review.

Aljullu:
Thanks for reviewing @psealock!

> Items that are low_in_stock aren't being returned by the `/products?low_in_stock=true` endpoint. Below, the Stock Report accurately labels Sunglasses and V-neck in Blue as `low`, but those products aren't being returned by the above endpoint, so I don't see them in the Stock panel.
> 
> Screen Shot 2019-06-27 at 12 20 05 PM
> 
> Can you reproduce this?

That's weird. I can't reproduce it but I think it's worth opening an issue for that, so we can investigate it further. There might be an inconsistency between `products` and `stock` endpoints and the way they calculate which products are in low stock.

I will merge this PR as-is since I think this issue is not related to the changes introduced here, and we can deal with it in a follow-up.

PR #2444, opened 2019-06-18
Route handling: Remove hash in favor of path parameter

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/2267

Remove use of the hash in routes so that backend can know full details of the route. Specifically, so that `WC_Admin_Page_Controller::get_current_page()` can receive the full route parameters because hashes aren't sent server-side.

Further details as explained in #2267:

>The problem is that the fragment part of the URL (everything that comes after the `#`) is **never** sent to the server. So, the PHP logic [here](https://github.com/woocommerce/woocommerce-admin/blob/22b2d7f846fd6cd1b3adcbbbebd08d5136c2b768/includes/page-controller/class-wc-admin-page-controller.php#L107) will never be able to correctly identify a "registered" page. It still works for traditional "connected" pages that don't rely on fragment URLs.

### Proposed Solution

Use `browserHistory` instead of `hashHistory` and place the route in the query as a parameter called `path`:

```
admin.php?page=wc-admin&path={ route }&period=week&compare=previous_year
```

### Test instructions

1. Navigate around the app
2. Use filters and advanced filters
3. Sort table columns
4. And follow links such as those found in table cells
5. Everything work as expected?

### DIFF
--- a/client/layout/controller.js
+++ b/client/layout/controller.js
@@ -53,17 +53,17 @@ export const getPages = () => {
 		pages.push( {
 			container: Analytics,
 			path: '/analytics',
-			wpOpenMenu: 'toplevel_page_wc-admin--analytics-revenue',
+			wpOpenMenu: 'toplevel_page_wc-admin-path--analytics-revenue',
 		} );
 		pages.push( {
 			container: AnalyticsSettings,
 			path: '/analytics/settings',
-			wpOpenMenu: 'toplevel_page_wc-admin--analytics-revenue',
+			wpOpenMenu: 'toplevel_page_wc-admin-path--analytics-revenue',
 		} );
 		pages.push( {
 			container: AnalyticsReport,
 			path: '/analytics/:report',
-			wpOpenMenu: 'toplevel_page_wc-admin--analytics-revenue',
+			wpOpenMenu: 'toplevel_page_wc-admin-path--analytics-revenue',
 		} );
 	}
 
--- a/client/layout/controller.js
+++ b/client/layout/controller.js
@@ -104,19 +104,10 @@ export class Controller extends Component {
 		return query;
 	}
 
-	// @todo What should we display or do when a route/page doesn't exist?
-	render404() {
-		return null;
-	}
-
 	render() {
 		const { page, match, location } = this.props;
 		const { url, params } = match;
 		const query = this.getBaseQuery( location.search );
-
-		if ( ! page ) {
-			return this.render404();
-		}
 
 		window.wpNavMenuUrlUpdate( page, query );
 		window.wpNavMenuClassChange( page, url );
--- a/client/layout/controller.js
+++ b/client/layout/controller.js
@@ -3,14 +3,14 @@
  * External dependencies
  */
 import { Component, createElement } from '@wordpress/element';
-import { parse } from 'qs';
-import { find, last, isEqual } from 'lodash';
+import { parse, stringify } from 'qs';
+import { isEqual, last } from 'lodash';
 import { applyFilters } from '@wordpress/hooks';
 
 /**
  * WooCommerce dependencies
  */
-import { getNewPath, getPersistedQuery, getHistory, stringifyQuery } from '@woocommerce/navigation';
+import { getNewPath, getPersistedQuery, getHistory } from '@woocommerce/navigation';
 
 /**
  * Internal dependencies
@@ -104,19 +104,23 @@ export class Controller extends Component {
 		return query;
 	}
 
+	// @todo What should we display or do when a route/page doesn't exist?
+	render404() {
+		return null;
+	}
+
 	render() {
-		// Pass URL parameters (example :report -> params.report) and query string parameters
-		const { path, url, params } = this.props.match;
-		const query = this.getQuery( this.props.location.search );
-		const page = find( getPages(), { path } );
+		const { page, match, location } = this.props;
+		const { url, params } = match;
+		const query = this.getBaseQuery( location.search );
 
 		if ( ! page ) {
-			return null; // @todo What should we display or do when a route/page doesn't exist?
+			return this.render404();
 		}
 
 		window.wpNavMenuUrlUpdate( page, query );
-		window.wpNavMenuClassChange( page );
-		return createElement( page.container, { params, path: url, pathMatch: path, query } );
+		window.wpNavMenuClassChange( page, url );
+		return createElement( page.container, { params, path: url, pathMatch: page.path, query } );
 	}
 }
 
@@ -125,38 +129,34 @@ export class Controller extends Component {
  * as is.
  *
  * @param {HTMLElement} item - Sidebar anchor link.
- * @param {string} nextQuery - A query string to be added to updated hrefs.
+ * @param {object} nextQuery - A query object to be added to updated hrefs.
  * @param {Array} excludedScreens - wc-admin screens to avoid updating.
  */
 export function updateLinkHref( item, nextQuery, excludedScreens ) {
-	/**
-	 * Regular expression for finding any WooCommerce Admin screen.
-	 * The groupings are as follows:
-	 *
-	 * 0 - Full match
-	 * 1 - "#/" (optional)
-	 * 2 - "analytics/" (optional)
-	 * 3 - Any string, eg "orders"
-	 * 4 - "?" or end of line
-	 */
-	const _exp = /page=wc-admin(#\/)?(analytics\/)?(.*?)(\?|$)/;
-	const wcAdminMatches = item.href.match( _exp );
+	const isWCAdmin = /admin.php\?page=wc-admin/.test( item.href );
 
-	if ( wcAdminMatches ) {
-		// Get fourth grouping
-		const screen = wcAdminMatches[ 3 ];
+	if ( isWCAdmin ) {
+		const search = last( item.href.split( '?' ) );
+		const query = parse( search );
+		const path = query.path || 'dashboard';
+		const screen = path.replace( '/analytics', '' ).replace( '/', '' );
 
-		if ( ! excludedScreens.includes( screen ) ) {
-			const url = item.href.split( 'wc-admin' );
-			const hashUrl = last( url );
-			const base = hashUrl.split( '?' )[ 0 ];
-			const href = `${ url[ 0 ] }wc-admin${ '#' === base[ 0 ] ? '' : '#/' }${ base }${ nextQuery }`;
-			item.href = href;
-		}
+		const isExcludedScreen = excludedScreens.includes( screen );
+
+		const href =
+			'admin.php?' + stringify( Object.assign( query, isExcludedScreen ? {} : nextQuery ) );
+
+		// Replace the href so you can see the url on hover.
+		item.href = href;
+
+		item.onclick = e => {
+			e.preventDefault();
+			getHistory().push( href );
+		};
 	}
 }
 
-// Update links in wp-admin menu to persist time related queries
+// Update's wc-admin links in wp-admin menu
 window.wpNavMenuUrlUpdate = function( page, query ) {
 	const excludedScreens = applyFilters( TIME_EXCLUDED_SCREENS_FILTER, [
 		'devdocs',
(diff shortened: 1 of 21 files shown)


packages/components/src/link/index.js
```diff
@@ -3,40 +3,47 @@
...
+import { getAdminLink, getHistory } from '@woocommerce/navigation';
 
 /**
  * Use `Link` to create a link to another resource. It accepts a type to automatically
  * create wp-admin links, wc-admin links, and external links.
  */
 class Link extends Component {
+	// @todo Investigate further if we can use <Link /> directly.
+	// With React Router 5+, <RouterLink /> cannot be used outside of the main <Router /> elements,
+	// which seems to include components imported from @woocommerce/components. For now, we can use the history object directly.
```
psealock: I agree, there is some potential for simplification. We could remove `getNewPath` which has potential consequences (accepts two query objects, one optional that defaults to the current query), in favor of a more direct approach that simply accepts an object. I think best to follow up on this afterwards.

DanReyLop:
> Specifically, so that WC_Admin_Page_Controller::get_current_page() can receive the full route parameters because hashes aren't sent server-side.

I'm questioning now if this effort is worth it. My initial (naive) approach was to call `get_current_page()` to see if the current page belonged to my extension, so I could enqueue the corresponding JS/CSS bundles. But, since all the navidation on registered pages is client-side, that check wouldn't work. The JS/CSS bundle will need to be enqueued if the current page is "registered", whether it's the Dashboard, an Analytics page, or a page for another WC-Admin-based extension.

So, as long as `WC_Admin_Page_Controller::get_current_page()` can correctly detect "connected" pages, and it detects any page starting with `page=wc-admin` is detected as being a "registered" page, everything should work fine.

I don't see any need for any extension to check in the backend which connected page is it going to be displayed, since all that navigation and routing happens client-side.

Do you see any other problem that the `browserHistory` approach solves?

justinshreve:
> Do you see any other problem that the browserHistory approach solves?

I'm not sure how important they are right now, but the things that come to mind are: 1) the ability to detect the page on initial load for setting things like bread crumbs or the `title` tag early. 2) The potential to pre-load API data on specific pages, to prime state.

DanReyLop:
Good points. So, at least this work isn't for nothing :)

psealock:
>1) the ability to detect the page on initial load for setting things like bread crumbs or the title tag early. 2) The potential to pre-load API data on specific pages, to prime state.

Although this PR isn't a requirement anymore, these are some good points. Especially number 2. @joshuatf would you mind weighing in on this? Do you think the changes proposed here provide enough value to push this one to the finish line?

joshuatf:
The points @justinshreve made are (IMO) the main reason for doing this and very useful.  Between hydrating initial state data and running server-side methods only on specific pages, this PR would be extremely useful.

I say go for it @psealock!

client/layout/controller.js
```diff
@@ -108,19 +109,46 @@ class Controller extends Component {
...
+	render404() {
+		return null;
+	}
+
 	render() {
+		const query = this.getBaseQuery( this.props.location.search );
 		// Pass URL parameters (example :report -> params.report) and query string parameters
-		const { path, url, params } = this.props.match;
-		const query = this.getQuery( this.props.location.search );
+		const match = this.getRouteMatch( query );
```
DanReyLop: This is just duplicating logic that should be in the `<Router>`. In fact, this logic works accidentally because the `<Router>` has a "default" `<Route>` ([here](https://github.com/woocommerce/woocommerce-admin/blob/master/client/layout/index.js#L106)). Since the routes in `getPages` don't exactly match the new generated URLs, [this logic](https://github.com/woocommerce/woocommerce-admin/blob/master/client/layout/index.js#L103-L105) will never be triggered so the page views will never be recorded in tracks.

I think a better solution is to fix the `<Router>` part, and pass down the `match` prop to the `<Controller>` from the Layout component.

psealock: I see what you're saying here. 

We can eliminate one or the other:

```js
// This

{ getPages().map( page => {
	return <Route key={ page.path } path={ page.path } exact component={ Layout } />;
} ) }

// or this
<Route component={ Layout } />
```
And if we eliminate the top section and pass everything to the same route, we can probably remove `<Router />` too. I know you mentioned this somewhere @DanReyLop.

Otherwise, if we remove the default `<Route />` (second block), we'll need to get the matched route give a query param `path=/analytics/stock`. This is my preferred option, but I want to explore each.

Looking into this one and hopefully I'll have something soon.

psealock: >This is just duplicating logic that should be in the `<Router>`

I agree, but I'm unable to get `<Route />` to recognize paths that look like `&path=/analytics/:report`. Alternatively, I can get a correct `match` manually using `getRouteMatch`, but can't pass it down to each `<Route />`.

I can simplify the `<PageLayout />` component:

```js
export class PageLayout extends Component {
	render() {
		return (
			<Router history={ getHistory() }>
				<Route component={ Layout } />
			</Router>
		);
	}
}
```

which works. Or, I can manually get the match, history, and location and pass them directly to `<Layout />`.

```js
export class PageLayout extends Component {
	render() {
		const history = getHistory();
		return (
			<Layout 
			    history={ history } 
			    location={ history.location } 
			    match={ getRouteMatch( parse( history.location.search ) ) }
			/>
		);
	}
}
```
which also works, but doesn't update pages on navigation.

@DanReyLop Do you have an idea that I've missed?

DanReyLop: > Otherwise, if we remove the default <Route /> (second block). We'll need to get the routes to match given a query param of path=/analytics/stock. This is my preferred option, but I want to explore each.

That's my preferred option too :)

> I agree, but I'm unable to get <Route /> to recognize paths that look like &path=/analytics/:report.

I'll take a look, see if I can find the problem.

DanReyLop: That's a bummer, it looks like React Router doesn't parse anything that's after the `?` in the URL. IMHO, we have 2 options:
- Remove the `<Router>` altogether. There's no point on just having just a catch-all `<Route>` that delegates into the `<Controller>`, that's a pointless level of indirection. Let the controller handle everything, at this point react-router isn't doing anything for us, except maybe re-rendering the Controller when the URL changes.
- Create a custom `history` object that fools the `<Router>` into thinking `/analytics/blah` **is** the actual URL path, and not just the querystring. I implemented a proof-of-concept in https://github.com/woocommerce/woocommerce-admin/commit/216bf8dd2ed680309ebd3fe07abfacdf8fd7ee83 (`try/custom-history` branch). It's not pretty. I removed the default catch-all `<Route>` to prove that the Router is doing the heavy lifting. With that approach, we could basically simplify a lot of the logic in the Controller (the Controller would receive a `page` object instead of having to parse the URL again). But it would complicate the `@navigation/history` package.

psealock: >Remove the <Router> altogether.

I have a local branch with this approach. It works well, except re-rendering the Controller, as you mentioned, on route changes. I'll need to sort that out.

>Create a custom `history`

I had the same thought! Thanks for the branch, I'll need to think on the implications.

A third option is to clone or create our own `<Route>` component that accepts a `match` prop instead of deriving that value on its own. This way, we can compute `match` via `getRouteMatch`.

I'll continue tinkering with this one. Thanks for the input @DanReyLop

DanReyLop: > It works well, except re-rendering the Controller, as you mentioned, on route changes. I'll need to sort that out.

Should be relatively easy, you could have code like this in the `Controller`:
https://github.com/ReactTraining/react-router/blob/6e9d819dd579715a6cb7502f31bdb25e206c82d6/packages/react-router/modules/Router.js#L31-L37

> A third option is to clone or create our own <Route> component that accepts a match prop instead of deriving that value on its own. This way, we can compute match via getRouteMatch.

That's interesting, but I think it would be too complicated, and at that point we'll be re-implementing most of that the `<Router>` does for us.

> I'll continue tinkering with this one. Thanks for the input @DanReyLop

No problem! Let me know if there's anything else I can do.

DanReyLop: > A third option is to clone or create our own Route component that accepts a match prop instead of deriving that value on its own. This way, we can compute match via getRouteMatch.

Looks like the component we'd want to modify is `<Switch>`, not `<Route>`:
https://github.com/ReactTraining/react-router/blob/6e9d819dd579715a6cb7502f31bdb25e206c82d6/packages/react-router/modules/Switch.js#L34

psealock: We have 3 possible solutions:

1. Remove the `<Router>` altogether. https://github.com/woocommerce/woocommerce-admin/compare/update/route-handling...test-no-router?expand=1. Manually compute the `page` and `match` props while listening for changes in browser history are re-rendering appropriately.

2.  Modify `history` such that `<Router>` is fooled into thinking `/analytics/blah` is the actual URL path. https://github.com/woocommerce/woocommerce-admin/compare/update/route-handling...try/custom-history?expand=1. Rewrite the `location` property of `history` so that `<Router>` matches the right path.

3. Rewrite `<Switch>` such that it chooses `match` according to our desired criteria. Logic [here](https://github.com/ReactTraining/react-router/blob/6e9d819dd579715a6cb7502f31bdb25e206c82d6/packages/react-router/modules/Switch.js#L34).

Lets rule out 3. It involves modifying functionality in the middle of a process owned by react-router (between `<Router />` and `<Route />`). This seems error prone and potentially brittle with updates of react-router if anything changes in any part of the process.

1 is a viable option, as is 2. I like 2 better, even though its hacky. The reason is because it keeps our meddling to a single method in a single file. A sufficient comment in `history.js` can clue in future readers. Option 1 of removing the router touches on several aspects and broadens the scope of what we are taking responsibility for, as opposed to what we let react-router handle.

Unless @DanReyLop can think of a reason for why removing the router (option 1) will be better, I'd like to merge option 2 into this branch.

DanReyLop: > Lets rule out 3.

I agree with that reasoning.

> I like 2 better, even though its hacky. The reason is because it keeps our meddling to a single method in a single file.

I agree that keeping all the "weird logic" in the custom History object is a good solution. But I'm a bit worried that the custom History object has a really inconsistent behaviour. `history.location` returns the `?path={...}` value, but `history.push(url)` expects the whole URL as a parameter. I believe that the WC-Admin app logic should only deal with paths, and those should be rewritten transparently by the History object to `?page=wc-admin&path={...}` URLs. That's an improvement that can go to the backlog though.

I can see the positives of removing the Router too, since it isn't really doing much for us anyway (removing it is a PR with just 69 lines added). Making the Router work with querystring variables seems like trying to fit a square peg into a triangular hole, maybe we would be better off removing it.

I don't have a strongly preferred solution, if you're looking for a "tie breaker" second opinion, mine ain't it :)

timmyc: Option 2 sounds good to me, as. long as the hacks are documented :)

psealock: >but history.push(url) expects the whole URL as a parameter.

Interesting point, and one to keep in mind. If we change `history.push()` to accept parameters instead of a url, that changes the expectation developers will have on `history`. Not to say this isn't something to think about, but we do have helper methods like `getNewPath` for url creation.

I feel good about moving forward with the `history` modification. I'll add documentation and merge it into this branch today. Thanks for the valuable input!

DanReyLop:
I did a search & replace or URLs that were still in the old format here: https://github.com/woocommerce/woocommerce-admin/pull/2444/commits/8384d988ceed5868ee66b448edf91c0567ed7f34

I haven't tested them though, most of those I don't even know how to get to. But the changes seem trivial enough.

Just an additional note for the PR, this will require a major version release for the package `@woocommerce/navigation`, since the changes in `history` are not backwards-compatible.

client/layout/controller.js
```diff
@@ -104,19 +104,23 @@ export class Controller extends Component {
...
-		const { path, url, params } = this.props.match;
-		const query = this.getQuery( this.props.location.search );
-		const page = find( getPages(), { path } );
+		const { page, match, location } = this.props;
+		const { url, params } = match;
+		const query = this.getBaseQuery( location.search );
 
 		if ( ! page ) {
-			return null; // @todo What should we display or do when a route/page doesn't exist?
+			return this.render404();
```
DanReyLop: This can never happen, since we removed the default `<Route>`

psealock:
Thanks for the reviews.

* render404 logic is removed, since it will never be reached in the controller.https://github.com/woocommerce/woocommerce-admin/pull/2444/commits/df34613c8439c0ca147bae066aa2add5bc4a646c
* Missing leading slash added. https://github.com/woocommerce/woocommerce-admin/pull/2444/commits/6a9075979518d3cec058f01809e7e0a4992eafc7

timmyc:
Coupons leaderboard all happy now :shipit:

PR #2457, opened 2019-06-19
fix type in readme changelog for 0.13.0

rrennick:
I noticed this typo reading the announcement post. It's appropriate that my commit message also has a type in it.

No changelog message.


rrennick:
> should we update the title before merge

Thanks, I updated it.

PR #2462, opened 2019-06-19
open external note action links in a new tab

### LINKED ISSUE #2403: Open offsite note actions in a new tab
From https://github.com/woocommerce/woocommerce-admin/pull/2325#issuecomment-500953530. 

In Inbox Notes, offsite actions/queries take the user away from the dashboard. 

Could we add a condition to the action query logic where if the query field starts with http then the action is opened in a new tab? 

That way, any action within WCA would use the current window, everything else would be a new tab.

rrennick:
Fixes #2403

This PR adds a handler to the note action button click. The handler opens the action link in a new tab if the URL starts with `http`.

- Add a note with an internal to WC Admin link
- Add a note with an external fully qualified URL
- Add a hook to verify that the REST API is being called
```
add_action( 'woocommerce_admin_note_action', function( $note ) {
	error_log( $note );
} );
```
- Load an Analytics screen
- Open the Inbox and click both action buttons
- The external should be opened in a new tab
- Both notes should be logged in debug.log

### Changelog Note:

Tweak: Open external note action links in a new tab

### DIFF
--- a/client/header/activity-panel/panels/inbox.js
+++ b/client/header/activity-panel/panels/inbox.js
@@ -37,7 +37,7 @@ class InboxPanel extends Component {
 		const { triggerNoteAction } = this.props;
 		const href = event.target.href || '';
 
-		if ( 'http' === href.substr( 0, 4 ).toLowerCase() ) {
+		if ( href.length && href.indexOf( 'wc-admin#/' ) <= 0 ) {
 			event.preventDefault();
 			window.open( href, '_blank' );
 		}
--- a/client/header/activity-panel/panels/inbox.js
+++ b/client/header/activity-panel/panels/inbox.js
@@ -33,6 +33,18 @@ class InboxPanel extends Component {
 		this.props.updateCurrentUserData( userDataFields );
 	}
 
+	handleActionClick( event, note_id, action_id ) {
+		const { triggerNoteAction } = this.props;
+		const href = event.target.href || '';
+
+		if ( 'http' === href.substr( 0, 4 ).toLowerCase() ) {
+			event.preventDefault();
+			window.open( href, '_blank' );
+		}
+
+		triggerNoteAction( note_id, action_id );
+	}
+
 	renderEmptyCard() {
 		return (
 			<ActivityCard
@@ -50,7 +62,7 @@ class InboxPanel extends Component {
 	}
 
 	renderNotes() {
-		const { lastRead, notes, triggerNoteAction } = this.props;
+		const { lastRead, notes } = this.props;
 
 		if ( 0 === Object.keys( notes ).length ) {
 			return this.renderEmptyCard();
@@ -65,7 +77,7 @@ class InboxPanel extends Component {
 					isDefault
 					isPrimary={ action.primary }
 					href={ action.url || undefined }
-					onClick={ () => triggerNoteAction( note.id, action.id ) }
+					onClick={ e => this.handleActionClick( e, note.id, action.id ) }
 				>
 					{ action.label }
 				</Button>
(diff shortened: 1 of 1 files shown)


psealock:
Whats the best way to add a note?

I'm trying to paste code like this in various spots with no luck:

```php
$note = new WC_Admin_Note();
$note->set_title( __( 'Test', 'woocommerce-admin' ) );
$note->set_content( __( 'Test content, 'woocommerce-admin' ) );
$note->set_type( WC_Admin_Note::E_WC_ADMIN_NOTE_INFORMATIONAL );
$note->set_icon( 'info' );
$note->set_name( 'hello' );
$note->add_action(
	'view-test',
	__( 'View Test', 'woocommerce-admin' ),
	'http://google.com'
);
$note->save();
```

client/header/activity-panel/panels/inbox.js
```diff
@@ -33,6 +33,18 @@ class InboxPanel extends Component {
 		this.props.updateCurrentUserData( userDataFields );
 	}
 
+	handleActionClick( event, note_id, action_id ) {
+		const { triggerNoteAction } = this.props;
+		const href = event.target.href || '';
+
+		if ( 'http' === href.substr( 0, 4 ).toLowerCase() ) {
```
joshuatf: It seems like all my links, including internal, include the full protocol and domain (e.g., `http://mylocalwp.test/wp-admin/...`.

Maybe we should compare this to see if it doesn't start with `http` and also doesn't start with `wcSettings.adminUrl`?

joshuatf:
>I'm trying to paste code like this in various spots with no luck:

@psealock That looks like it should work.  Is `WC_Admin_Note` loaded already?  No errors being thrown?

I think the [test bench plugin](https://github.com/Automattic/wc-admin-notes-test-bench/) also gives a GUI for adding notes and testing.

client/header/activity-panel/panels/inbox.js
Aljullu: @rrennick what do you think about using [`startsWith`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) here? It looks like a good case to use that method:
```suggestion
		if ( href.toLowerCase().startsWith( 'http' ) ) {
```

rrennick: @joshuatf Do we want to open an order edit or other WC is the current tab or a new one? 

I was using the internal/external division of internal to WooCommerce Admin. I've updated the PR so it works with that division but can easily change to anywhere in the WP dashboard being internal.

rrennick:
@joshuatf or @Aljullu If one of you had a minute to respond to my question above I could get this one finished off.

client/header/activity-panel/panels/inbox.js
joshuatf: Not opposed to either, but I would lean towards anything in the dashboard opening in the same tab, regardless of whether its a wc-admin link or not.

Aljullu: I agree with @joshuatf, from a user perspective I think there should be no division between WooCommerce Admin pages and other WooCommerce pages, so we should treat all of them the same.

But I don't have a strong opinion in either way so the current solution would also work for me.

client/header/activity-panel/panels/inbox.js
```diff
@@ -33,6 +33,18 @@ class InboxPanel extends Component {
 		this.props.updateCurrentUserData( userDataFields );
 	}
 
+	handleActionClick( event, note_id, action_id ) {
+		const { triggerNoteAction } = this.props;
+		const href = event.target.href || '';
+
+		if ( ! href.startsWith( wcSettings.adminUrl ) ) {
```
Aljullu: We still need to verify `href` is not an empty string, right?

```suggestion
		if ( href.length && ! href.startsWith( wcSettings.adminUrl ) ) {
```

rrennick:
@Aljullu I added that suggestion & this is ready for another review.

PR #2463, opened 2019-06-19
Add State alongside City (and other customer info).

KZeni:
Adds an oddly missing State data point when it currently has Country, City, and Postal Code. I have a client with State-based campaigns, currently, so them not having that one data point is something that could/should be addressed for them and others.

Why would State not be included as a possible data point in the reporting if it's not shown by default (similar to City)?

### Accessibility

I simply added a new data point based off of the existing City field so its accessibility qualities should carry over appropriately.

### Changelog Note:

Enhancement: add state/region to Customers Report. 👏🏻 @KZeni


KZeni:
@joshuatf Done!

I can see that, but WooCommerce itself has State as a data point with accompanying functions for processing it, etc. accordingly. So I really think it should be able to be surfaced if it's wanted/needed. For those that sell mostly/exclusively in the US; having State included is a must-have for location information.

Just because it isn't always present doesn't mean it shouldn't be available. For example, "Why have a Coupon column for the order report if some orders don't have coupons?" would apply to the reasoning for not including state/province.

As an aside, I could certainly see the columns name changed to "State/Province" to be more widely applicable if that's a hangup.

jameskoster:
@joshuatf forgive me, where is this being added? :>

KZeni:
@jameskoster The goal is to get State (or rather "State/Province") shown alongside City & other info which is currently being shown as an optional column on Analytics => Customers for WooCommerce stats.

joshuatf:
@jameskoster As @KZeni noted (thanks!), this would be a column in the Customers report.  Stores that don't need this can toggle the column off as needed.  A label of `State / Province` might be more broadly matching across different addresses though- but you can weigh in there as well if you have an opinion.

Screen Shot 2019-06-27 at 11 33 28 PM

LevinMedia:
@KZeni @joshuatf Sorry I'm late to the party. This is fantastic, and something I've been hoping we could add for a while now. Excited to see this added to the mix. 

I did a little bit of digging, I think using "Region" would be a little more inclusive than even State / Province". This is the pattern Google Analytics follows as well:  Country > Region > City

KZeni:
@LevinMedia Oh yeah, that sounds like the best label to go with so far. More inclusive & concise. 👍

jeffstieler:
Just a heads up - I'm removing the [ready to merge] label since we're making a change to the field name. @KZeni please ping after you've updated `state` to `region`, or let us know if you'd like us to make the change instead. Thank you!

KZeni:
@jeffstieler Done.

I renamed the label in the table and API to be "Region" instead of "State". I kept it as "state" behind the scenes since WooCommerce itself considers this as state in its functions, keys, filters, etc. behind the scenes. That way, it's consistent with WooCommerce behind the scenes while having the label be more appropriate for when it's shown.

client/analytics/report/customers/table.js
```diff
@@ -78,16 +78,25 @@ export default class CustomersReportTable extends Component {
...
+			},
+			{
+				label: __( 'Region', 'woocommerce-admin' ),
+				key: 'state',
+				hiddenByDefault: true,
+				isSortable: true,
 			},
 			{
 				label: __( 'Postal Code', 'woocommerce-admin' ),
 				key: 'postal_code',
```
jeffstieler: Adding sorting highlights a bug we have - this should be `postcode`. Can you please fix @KZeni?

jeffstieler:
One more small tweak and this is ready to go @KZeni! Thank you for your contribution.

KZeni:
@jeffstieler Okay! That should be taken care of if I understood things correctly.

PR #2467, opened 2019-06-20
Onboarding: Pull dynamic prices for product types

### LINKED ISSUE #2341: Onboarding - Profile Wizard: Pull WC.com product prices in dynamically
The extensions screen of WooCommerce pulls in product and price information dynamically from WooCommerce.com. We should be able to follow a similar pattern to load/cache prices, without having them hard-coded into the app.

It looks like we can use the following request: `https://woocommerce.com/wp-json/wccom-extensions/1.0/search?category=product-type`

joshuatf:
Fixes #2341

Adds prices for each product dynamically.

### Questions
* Do we need any caching here?  I think for such a small, one-time use request it’s probably okay to leave uncached.
* We had some discussion about removing the 2nd bookings item.  Renamed this “Bookings” to cover both “spaces” and “rentals.”  Also opted to use the WCCOM excerpts to keep things consistent.  Do we need a design review on this or should we revert descriptions?

### Screenshots
Screen Shot 2019-06-20 at 5 24 24 PM

1.  Visit `wp-admin/admin.php?page=wc-admin#/?step=product-types`
2. Make sure product prices, descriptions, and “Learn More” URLs are all pulled in dynamically based on a. Product types “product” key.

### Changelog Note:

Enhancement: Pull dynamic prices for product types during onboarding

### DIFF
--- a/includes/features/onboarding/class-wc-admin-onboarding.php
+++ b/includes/features/onboarding/class-wc-admin-onboarding.php
@@ -145,11 +145,18 @@ class WC_Admin_Onboarding {
 	 * @param array $settings Component settings.
 	 */
 	public function component_settings( $settings ) {
+		$profile = get_option( 'wc_onboarding_profile', array() );
+
 		$settings['onboarding'] = array(
-			'industries'   => self::get_allowed_industries(),
-			'productTypes' => self::get_allowed_product_types(),
-			'profile'      => get_option( 'wc_onboarding_profile', array() ),
+			'industries' => self::get_allowed_industries(),
+			'profile'    => $profile,
 		);
+
+		// Only fetch if the onboarding wizard is incomplete.
+		if ( $this->should_show_profiler() ) {
+			$settings['onboarding']['productTypes'] = self::get_allowed_product_types();
+		}
+
 		return $settings;
 	}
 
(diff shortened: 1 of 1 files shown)


justinshreve:
> Do we need any caching here? I think for such a small, one-time use request it’s probably okay to leave uncached.

I think we probably do need caching with how this is currently built. The `productType` data is appended to `wcSettings`, which is generated on every page load that embeds the activity panel.

You can `console.log( wcSettings.onboarding.productTypes );` in your console on the orders page for example, and see them loaded.

joshuatf:
>I think we probably do need caching with how this is currently built. The productType data is appended to wcSettings, which is generated on every page load that embeds the activity panel.

Good point, @justinshreve .  I added a check to see if the profiler is finished before loading this data since I don't think we'll need it beyond the profiler (at least for now) - https://github.com/woocommerce/woocommerce-admin/pull/2467/commits/d9b1548eadb5a3e52927062f5f2b7566ceefb765

Also cached the remote request for a day in https://github.com/woocommerce/woocommerce-admin/pull/2467/commits/4e23aa2b05fd6084c20dcad9d8c0e08cde6ddb20

Let me know if you think that works okay.

emdashcodes (approved):
Thanks for the updates. I think the caching and extra check work! This is testing great as well.

PR #2468, opened 2019-06-20
Update onboarding track setting

justinshreve:
This PR updates the onboarding tracks setting to save both the continue and skip actions in the profile wizard. See p1560435371012800-slack-wc-onboarding

* Open your network inspector, and click the 'skip' link. Notice the setting successfully save.
* Test the continue option and see the same. Try unchecking the checkbox.


justinshreve:
> Quick sanity check while we're here- they won't have ever set the tracking option before this if they hit this screen right? No need to pull in the existing setting for the tracking option?

I don't think so. The only way would be to reset the the `skipped` or `completed` flags, and if someone does that, I think it's OK to offer the choice again on the screen.

PR #2491, opened 2019-06-24
Allow each page to specify their breadcrumbs without having to render the Header component

DanReyLop:
I've linked the relevant issue on the other extension from here, see it for more context.

There's a problem in how the `<Header>` component is rendered: the component itself is bundled in the main app JS. That means that it's not possible for another extension to render the `<Header>` component on a WC-Admin registered page.

Since the only parameter that's sent to the `<Header>` component is `sections` (basically, an array of breadcrumbs), instead of each section's component rendering the `<Header>` component, each page definition will have a `breadcrumbs` property that will contain that same data.

This has the advantage that we can safely remove the `react-slot-fill` dependency. I'm not aware of any trade-offs of this approach, everything seems to work as expected.


client/layout/controller.js
```diff
@@ -43,24 +45,31 @@ const getPages = () => {
...
 			path: '/',
+			breadcrumbs: [ __( 'Dashboard', 'woocommerce-admin' ) ],
 			wpOpenMenu: 'toplevel_page_woocommerce',
 		} );
 	}
 
 	if ( window.wcAdminFeatures.analytics ) {
-		pages.push( {
-			container: Analytics,
-			path: '/analytics',
```
DanReyLop: This just renders and empty page. I've removed it since there's no link that points to it. Is it important?

client/layout/index.js
```diff
@@ -101,9 +101,24 @@ export class PageLayout extends Component {
...
+												? page.breadcrumbs( props )
+												: page.breadcrumbs ) || []
+										}
+										{ ...props }
+									/>
+								) }
+							/>
+						);
 					} ) }
-					<Route component={ Layout } />
```
DanReyLop: This "default" route shouldn't exist. Is there any reason for it?

psealock: This should be the 404 page

psealock:
Shall I review this, or wait to see what happens with https://github.com/woocommerce/woocommerce-admin/pull/2444?

DanReyLop:
> Shall I review this, or wait to see what happens with #2444?

Let's wait, the `<Router>` changes here will be affected by whatever decision is made on #2444.

DanReyLop:
@psealock This is ready for review now, I rebased with `master` to include the changes on #2444 .

I found out that some breadcrumb links are broken in `master` because of the changes on #2444, I fixed everything I could find.

To test:
- Clicking WP-Admin breadcrumbs on embedded pages should work, doing a full-page reload. For example, clicking `Settings` when you're on `WooCommerce / Settings / Payments`
- Clicking the `WooCommerce` link on an embedded page should work, doing a full-page reload.
- Clicking the `WooCommerce` link on a registered page (like `WooCommerce / Analytics / Products`) should navigate to the Dashboard without reloading the page.
- Clicking a WC-Admin link on a registered page (like `Analytics` in `WooCommerce / Analytics / Products`) should navigate to the page without reloading it.

client/layout/index.js
```diff
@@ -76,9 +76,13 @@ class Layout extends Component {
 
 	render() {
 		const { isEmbedded, ...restProps } = this.props;
+		const { breadcrumbs } = this.props.page;
 		return (
 			<div className="woocommerce-layout">
-				<Slot name="header" />
+				<Header
+					sections={ ( isFunction( breadcrumbs ) ? breadcrumbs( this.props ) : breadcrumbs ) || [] }
```
psealock: Considering pages can be added by 3PD, it would be useful to elaborate `Layout.propTypes` for `page` and make breadcrumbs required. Then, this ternary + `||` expression can be simplified and we get better checks against misuse.

client/layout/controller.js
psealock: Clicking "Analytics" simply goes to the revenue page. Maybe this was a leftover from earlier times? Safe to remove.

Screen Shot 2019-07-04 at 3 21 15 PM

client/layout/index.js
DanReyLop: Sure, done in https://github.com/woocommerce/woocommerce-admin/pull/2491/commits/b4e3bc0323936522a7eb83d85f95237f99e09b59 👍

PR #2492, opened 2019-06-24
Add a filter to allow other extensions to register new WC-Admin-powered pages

DanReyLop:
I've linked the relevant issue on the other extension from here, see it for more context.

This PR adds a filter so that other extensions can "plug into" the WC-Admin client-side router and add new pages.

Since it's a filter that receives the existing list of pages, an extension could remove existing WC-Admin pages. Is that something we want to allow? Or should we only make it possible to *add* new pages?

### DIFF
--- a/client/layout/index.js
+++ b/client/layout/index.js
@@ -3,6 +3,7 @@
  * External dependencies
  */
 import { Component, Fragment } from '@wordpress/element';
+import { useFilters } from '@woocommerce/components';
 import { Router, Route, Switch } from 'react-router-dom';
 import { Slot } from 'react-slot-fill';
 import PropTypes from 'prop-types';
@@ -17,7 +18,7 @@ import { getHistory } from '@woocommerce/navigation';
  * Internal dependencies
  */
 import './style.scss';
-import { Controller, getPages } from './controller';
+import { Controller, getPages, PAGES_FILTER } from './controller';
 import Header from 'header';
 import Notices from './notices';
 import { recordPageView } from 'lib/tracks';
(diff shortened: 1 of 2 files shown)


client/layout/index.js
```diff
@@ -124,6 +125,7 @@ export class PageLayout extends Component {
 		);
 	}
 }
+export const PageLayout = useFilters( PAGES_FILTER )( _PageLayout );
```
psealock: Just curious here what the advantage of `useFilters` is. It appears to require `PAGES_FILTER` to be passed around with little benefit.

psealock:
>Since it's a filter that receives the existing list of pages, an extension could remove existing WC-Admin pages. Is that something we want to allow? 

We've faced this question previously and have decided that safeguarding against this is futile since a jQuery script can also do the same thing. So, in short, this is fine.

client/layout/index.js
DanReyLop: That's so the `<PageLayout>` component (and so, the `<Router>`) is re-rendered when `useFilter(PAGES_FILTER, () => { ... })` is called anywhere. `useFilters` indicated that this component uses the result of `PAGES_FILTER`, so it needs to be re-rendered when that result can potentially change.

psealock: Ah ok I see. Cool, could you add a comment to the code with a word or two about that?

DanReyLop: Sure, done in https://github.com/woocommerce/woocommerce-admin/pull/2492/commits/59f158f5738f61ebfadd9bc0d9a035794137868c

PR #2493, opened 2019-06-24
Onboarding: Add theme step and theme browser

### LINKED ISSUE #2159: Onboarding - Profile Wizard: Step 8 - Theme Browser
On this step of the profile wizard, WooCommerce compatible themes will be displayed to the end user. We can use the new theme info endpoint to pull the information we need: https://github.com/Automattic/woocommerce.com/pull/6035.

Since we have a small number of themes, we can ommit the filters/pagination for v1. We should still have tabs for free and paid themes.

Columns showing related themes should be displayed with image, price, description, and buttons for selection or live demo. The demo functionality will be handled in #2469.

joshuatf:
Fixes #2159

Adds a theme browser to choose the current theme.

Note that this only adds wccom themes and updates profile information; installed themes theme and theme activation will come in future PRs.

### Questions

* The current list of themes is 15 in length from wccom.  We will eventually show all installed themes plus those from wccom.  Do we need any sort of pagination?
* Do we want to change sorting here?  Order alphabetically?  Wccom themes first, then installed themes?  By theme support?

### Screenshots
Screen Shot 2019-06-24 at 6 26 35 PM

1.  Get a local woocommerce.com test site running and use [this branch](https://github.com/Automattic/woocommerce.com/pull/6035).
2.  Visit `wp-admin/admin.php?page=wc-admin#/?step=theme` in your local WP install.
3.  Make sure themes show up correctly, match updated styles in figma, and clicking "Choose" updates the onboarding profiler with the theme information.
4.  Selecting a tab (all/paid/free) should filter respective themes.

### Changelog Note:

Enhancement: Add theme browser in onboarding theme step.

### DIFF
--- a/client/dashboard/profile-wizard/index.js
+++ b/client/dashboard/profile-wizard/index.js
@@ -23,6 +23,7 @@ import ProductTypes from './steps/product-types';
 import ProfileWizardHeader from './header';
 import Start from './steps/start';
 import StoreDetails from './steps/store-details';
+import Theme from './steps/theme';
 import withSelect from 'wc-api/with-select';
 import './style.scss';
 
@@ -58,7 +59,7 @@ const getSteps = () => {
 	} );
 	steps.push( {
 		key: 'theme',
-		container: Fragment,
+		container: Theme,
 		label: __( 'Theme', 'woocommerce-admin' ),
 	} );
 	return steps;
--- a/client/dashboard/profile-wizard/steps/theme/index.js
+++ b/client/dashboard/profile-wizard/steps/theme/index.js
@@ -7,6 +7,7 @@ import { Button } from 'newspack-components';
 import { Component, Fragment } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import { decodeEntities } from '@wordpress/html-entities';
+import { TabPanel } from '@wordpress/components';
 import { withDispatch } from '@wordpress/data';
 
 /**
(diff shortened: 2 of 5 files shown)


justinshreve:
cc @jameskoster on the above questions.

>The current list of themes is 15 in length from wccom. We will eventually show all installed themes plus those from wccom. Do we need any sort of pagination?

I was originally going to say no based on the number of themes, but the page does feel a little long actually looking at it.

The current endpoint just returns 100 results, but we could add some pagination options without breaking backwards compatibility.

> Do we want to change sorting here? Order alphabetically? Wccom themes first, then installed themes? By theme support?

I think Storefront and the upload option (coming soon) should probably be the first two items. I think the remaining options could be by theme support, installed, and then wccom. WCCOM themes are already sorted by `popularity`.

joshuatf:
Thanks for the review, @justinshreve!

>I did have to copy and manually load the JSON, because I can't run MAMP and Docker at the same time, but hopefully we will have that endpoint deployed soon.

I ran into the the same issues.  Ended up keeping Docker on port 80 and changing my local apache config to use a different port.

I'll ping @jameskoster in slack and see if we can get some feedback on pagination/sorting so we can handle that in a follow-up.

jameskoster:
I couldn't get this working locally :( I didn't set up a local woo.com though as it looks like that PR was merged... maybe I still need to? It's looking good from the screenshots though, assuming the labels on those tabs will be centred! :)

Anyway... If it's possible I don't think we should display _all_ installed themes here. We're on-boarding for WooCommerce so there's not much point presenting incompatible themes, with the exception of the current active theme. That should keep the list a bit shorter?

As for the order. I'm thinking:

1. Current active theme
2. Storefront
3. Marketplace suggestions
4. Installed themes that are WC compatible
5. Upload

If the list is still too long, I think an infinite scroll implementation would probably work better than pagination. Pagination within a stepper might feel a bit disorienting.

PR #2496, opened 2019-06-24
Test that REST controllers work with their default parameters.

jeffstieler:
In testing my Reports CSV Exporter class, I noticed that some REST controllers caused SQL errors with their default values.

This PR seeks to fix the few errant/missing defaults and add a test to ensure we don't regress.

- Make some REST API requests to report endpoints with no parameters
- Verify they work
- Check the new test class
- Verify all tests pass

### Changelog Note:

Tweak: fix some report endpoint default params.

### DIFF
--- a/tests/api/report-controllers.php
+++ b/tests/api/report-controllers.php
@@ -1,0 +1,72 @@
+<?php
+/**
+ * REST Controller Tests
+ *
+ * @package WooCommerce\Tests\API
+ * @since 3.6.4
+ */
+
+/**
+ * REST Controller Tests Class
+ *
+ * @package WooCommerce\Tests\API
+ * @since 3.6.4
+ */
+class WC_Tests_API_Report_Controllers extends WP_UnitTestCase {
+	/**
+	 * Ensure that every REST controller works with their defaults.
+	 */
+	public function test_default_params() {
+		$controller_classes = array(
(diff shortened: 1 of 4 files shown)


tests/api/report-controllers.php
```diff
@@ -0,0 +1,107 @@
...
+			$defaults   = array();
+
+			foreach ( $params as $arg => $options ) {
+				if ( isset( $options['default'] ) ) {
+					$defaults[ $arg ] = $options['default'];
+				}
+			}
+
+			$request = new WP_REST_Request();
+			$request->set_default_params( $defaults );
```
psealock: I'm not entirely sure what `set_default_params` does and the [docs](https://developer.wordpress.org/reference/classes/wp_rest_request/set_default_params/) aren't too descriptive, but isn't the purpose of the test to how the endpoint responds without any parameters?

jeffstieler: It's part of what the `WP_REST_Server` class does when [handling routes](https://github.com/WordPress/WordPress/blob/master/wp-includes/rest-api/class-wp-rest-server.php#L882) - since I wasn't using the server in test, I had to manually apply the defaults.

I've switched the test up to make use of the server instead. Hopefully it's easier to grok: da7754bc

Issue #2474, opened 2019-06-21
Tracks: Add priority 1 events


psealock:
Add the following Tracks events to the UI - when planning to work on an event, please update the table status column to say "`in progress - @urgithubhandlehere`

Event | Description | Props | Priority | Status
-- | -- | -- | -- | --
wcadmin_activity_panel_open | When a user clicks on an activity panel to open it | type ( Inbox, Orders, Stock ) | 1 | #2498
wcadmin_inbox_action_click | When a user clicks an action button inside an inbox message | message name | 1 | #2559
wcadmin_store_alert_snooze | When a store alert is snoozed | duration ( tomorrow, next week, next month ), alert name | 1 | #2560
wcadmin_store_alert_action | When an action button is clicked in a store alert | alert name, action | 1 | #2559
wcadmin_dash_section_remove | When a user removes a dashboard section | section type | 1 | #2550 
wcadmin_dash_charts_chart_toggle | When a user toggles a chart on or off in the chart section | chart_name, status ( on \| off ) | 1 | #2550
wcadmin_dash_indicators_toggle | When a user toggles a performance indicator off/on | indicator_name, status ( on \| off ) | 1 | #2550
wcadmin_dash_indicators_click | When a user clicks on a performance indicator | indicator_name | 1 | #2550
wcadmin_dash_leaderboards_toggle | When a user toggles a leaderboard off \| on | leaderboard_name, status | 1 | #2550
wcadmin_analytics_chart_tab_click | When a user changes the chart on a report page | report name, tab clicked | 1 | #2557 
wcadmin_analytics_table_download | When a user clicks download csv | report name, row quantity | 1 | #2557 
wcadmin_analytics_filters_add | When a user adds an advanced filter | report name, filter name | 1 | #2525
wcadmin_analytics_filters_remove | When a user removes an advanced filter | report name, filter name, filter options | 1 | #2525
wcadmin_analytics_filters_clear_all | When a user clicks "clear all filters" link | report name | 1 | #2525
wcadmin_analytics_filters_filter | When the "Filter" button is clicked after changing filters | report name, number of filters active, filter optiions | 1 | #2525
wcadmin_analytics_filter | When the report filter is changed to a preset | report name, type ( single, all, comparison ) | 1 | #2525
wcadmin_datepicker_update | When a user modifies the date range | type: preset vs custom, compare_to, preset ( if applicable ), length of date range if custom, page ( report name or dashboard ) | 3 | #2525

[...]

PR #2498, opened 2019-06-24
Activity Panels: Add track event when panels are opened.

timmyc:
For #2474

This branch adds a tracks event when an Activity Panel is opened. For the ease of debug-ability, this PR also adds in `debug` as a dependency, and creates a new debug log for `wc-admin:tracks`. `freshdata` also uses debug, so figured it might be nice to add it in here too so it makes testing PRs like these tracks ones much easier.

- In your console `localStorage.setItem( 'debug', 'wc-admin:*' );`
- Open up any Woo page that has activity panels
- Toggle each panel on/off and verify the proper call to `recordEvent()` is happening. Note when running in development mode, you shouldn't see an actual call to `t.gif` happening in your network tab.

### DIFF
--- a/client/header/activity-panel/index.js
+++ b/client/header/activity-panel/index.js
@@ -25,6 +25,7 @@ import {
 import InboxPanel from './panels/inbox';
 import OrdersPanel from './panels/orders';
 import StockPanel from './panels/stock';
+import { recordEvent } from 'lib/tracks';
 import ReviewsPanel from './panels/reviews';
 import withSelect from 'wc-api/with-select';
 import WordPressNotices from './wordpress-notices';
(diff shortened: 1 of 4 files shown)


client/lib/tracks.js
```diff
@@ -8,6 +17,8 @@
  */
 
 export function recordEvent( eventName, eventProperties ) {
+	tracksDebug( 'recordevent %s %o', eventName, eventProperties );
```
psealock: 👍

client/header/activity-panel/index.js
```diff
@@ -45,6 +46,13 @@ class ActivityPanel extends Component {
 	}
 
 	togglePanel( tabName ) {
+		const { isPanelOpen, currentTab } = this.state;
+
+		// If no panel is open, or if a new panel is opening, record a track.
```
psealock: I'm having trouble understanding this comment, can you elaborate?

psealock:
```suggestion
// If a panel is being opened, or if an existing panel is already open and a different one is being opened, record a track.
```

Issue #311, opened 2018-08-21
Analytics: CSV Download


timmyc:
![68747470733a2f2f6175746f6d61747469632e696e766973696f6e6170702e636f6d2f7374617469632d7369676e65642f6c6976652d656d6265642f3236363e280a6-2018-08-07-11-28-25](https://user-images.githubusercontent.com/22080/44428576-a426c500-a549-11e8-9447-cfe97f9b8b34.png)

Background: p90Yrv-LB-p2

The background thread above contains some excellent discussion on this feature from both a business side, and a technical approach. But I did want to highlight this comment from @warrendholmes which sums up the feature nicely:

> we will move forward with functionality to implement exporting all the data.

This issue will act as a parent issue for child tasks related to the Download CSV functionality.

**Phase 1: Browser-Side Export of data in table**
Since the ability to fully export _all_ data that could be returned by a particular set of filters will require some additional server side work, we would like to implement an interim solution so we can conduct usability tests against the Revenue Report ( and others ) prior to the completion of the back-end logic.

As such, phase one will implement the same approach that exists in both Calypso Stats and legacy WooCommerce reports, where data that is already present in the client will be exported. [Calypso CSV Download](https://github.com/Automattic/wp-calypso/tree/a70587157b33783ba4d5e0e4a1e136680d459d51/client/my-sites/stats/stats-download-csv) is one implementation that can be used as prior art.

- [x] Tracking phase one in #312.

**Phase 2: Server Side Job / Endpoints**
As outlined in the background post, the long-term desired setup for the CSV Download involves the creation of a new API endpoint that will accept arguments of report type ( i.e. Revenue Report ), and any associated filters ( i.e. date ranges, unit/day/week/month ) and enqueue a job to query and persist the data as a csv file. An additional endpoint will allow the client-side logic to query and determine the status of a pending export job, and get the download URL for the file when it is complete.

Prior discussion hinted at possibly being able to leverage Woo's built in digital download support to secure the data file export process.

__Server Tasks__
- [x] Implement Job logic to query large data sets, and save results to the file system
- [x] Create API endpoints to allow queueing a new job, and requesting status of prior jobs
- [x] Implement logic to ensure the serving of the exported data files is secure, and files may only be retrieved by store users with proper permissions
 
__Client Side Tasks__
[...]

PR #2526, opened 2019-06-27
Add Report CSV Export Endpoint

jeffstieler:
Addresses #311 partially.

This PR adds:

- 2 new endpoints: `/reports/[type]/export` and `/reports/[type]/export[id]/status`
- Export file download capability (on `admin_init`, like core's product export)
- Queued batch creation of report export files
- Example filter for modifying schema when adding fields to report API responses

- `POST /wc/v4/reports/products/export`
- `GET /wc/v4/reports/products/export/<id here>/status` (get URL from `POST` response)
- When 100%, `download_url` should be in status response
- Open download URL in browser
- Verify CSV contents

### DIFF
--- a/docs/examples/extensions/table-column/woocommerce-admin-table-column.php
+++ b/docs/examples/extensions/table-column/woocommerce-admin-table-column.php
@@ -32,6 +32,27 @@ function table_column_register_script() {
 add_action( 'admin_enqueue_scripts', 'table_column_register_script' );
 
 /**
+ * When adding fields to a REST API response, we need to update the schema
+ * to reflect these changes. Not only does this let API consumers know of
+ * the new fields, it also adds them to Report Exports.
+ *
+ * @param array $properties The endpoint item schema properties.
+ * @return array Filtered schema.
+ */
+function add_product_extended_attributes_schema( $properties ) {
+	$properties['extended_info']['average_rating'] = array(
+		'type'        => 'number',
+		'readonly'    => true,
+		'context'     => array( 'view', 'edit' ),
+		'description' => 'Average product rating.',
+	);
+
+	return $properties;
+}
+
+add_filter( 'woocommerce_rest_report_products_schema', 'add_product_extended_attributes_schema' );
+
+/**
  * Extended attributes can be used to obtain any attribute from a WC_Product instance that is
  * available by a `get_*` class method. In other words, we can add `average_rating` because
  * `get_average_rating` is an available method on a WC_Product instance.
--- a/includes/api/class-wc-admin-rest-reports-export-controller.php
+++ b/includes/api/class-wc-admin-rest-reports-export-controller.php
@@ -160,10 +160,9 @@ class WC_Admin_REST_Reports_Export_Controller extends WC_Admin_REST_Reports_Cont
 	public function export_items( $request ) {
 		$report_type = $request['type'];
 		$report_args = empty( $request['report_args'] ) ? array() : $request['report_args'];
-		$user_id     = get_current_user_id();
 		$export_id   = str_replace( '.', '', microtime( true ) );
 
-		$total_rows = WC_Admin_Report_Exporter::queue_report_export( $user_id, $export_id, $report_type, $report_args );
+		$total_rows = WC_Admin_Report_Exporter::queue_report_export( $export_id, $report_type, $report_args );
 
 		if ( 0 === $total_rows ) {
 			$response = rest_ensure_response(
@@ -190,7 +189,7 @@ class WC_Admin_REST_Reports_Export_Controller extends WC_Admin_REST_Reports_Cont
 				)
 			);
 
-			WC_Admin_Report_Exporter::update_export_percentage_complete( $user_id, $report_type, $export_id, 0 );
+			WC_Admin_Report_Exporter::update_export_percentage_complete( $report_type, $export_id, 0 );
 		}
 
 		$data = $this->prepare_response_for_collection( $response );
@@ -207,8 +206,7 @@ class WC_Admin_REST_Reports_Export_Controller extends WC_Admin_REST_Reports_Cont
 	public function export_status( $request ) {
 		$report_type = $request['type'];
 		$export_id   = $request['export_id'];
-		$user_id     = get_current_user_id();
-		$percentage  = WC_Admin_Report_Exporter::get_export_percentage_complete( $user_id, $report_type, $export_id );
+		$percentage  = WC_Admin_Report_Exporter::get_export_percentage_complete( $report_type, $export_id );
 
 		if ( false === $percentage ) {
 			return new WP_Error(
--- a/includes/api/class-wc-admin-rest-reports-export-controller.php
+++ b/includes/api/class-wc-admin-rest-reports-export-controller.php
@@ -43,7 +43,7 @@ class WC_Admin_REST_Reports_Export_Controller extends WC_Admin_REST_Reports_Cont
 				array(
 					'methods'             => WP_REST_Server::EDITABLE,
 					'callback'            => array( $this, 'export_items' ),
-					'permission_callback' => array( $this, 'export_permissions_check' ),
+					'permission_callback' => array( $this, 'get_items_permissions_check' ),
 					'args'                => $this->get_export_collection_params(),
 				),
 				'schema' => array( $this, 'get_export_public_schema' ),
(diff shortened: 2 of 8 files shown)


rrennick:
@jeffstieler I'm getting a `There is no data to export for the given request.` on orders, products, and coupons with no payload on the POST.

Also, this PR has a merge conflict.

jeffstieler:
> I'm getting a `There is no data to export for the given request.` on orders, products, and coupons with no payload on the POST.

That's likely because the default date range for most report queries is one week - which is masked when using the UI. https://github.com/woocommerce/woocommerce-admin/blob/master/includes/data-stores/class-wc-admin-reports-orders-data-store.php#L154-L155

> Also, this PR has a merge conflict.

I'll fix that now.

psealock:
I'm using before and after query args and getting the same `There is no data to export for the given request.` message. Here is my POST request url.

```
http://tangaroa.test/wp-json/wc/v4/reports/products/export?after=2019-06-01T00:00:00&before=2019-06-30T00:00:00
```

The Products Report for the same periods has results.

docs/examples/extensions/table-column/woocommerce-admin-table-column.php
```diff
@@ -31,6 +31,27 @@ function table_column_register_script() {
...
+		'type'        => 'number',
+		'readonly'    => true,
+		'context'     => array( 'view', 'edit' ),
+		'description' => 'Average product rating.',
+	);
+
+	return $properties;
+}
+
+add_filter( 'woocommerce_rest_report_products_schema', 'add_product_extended_attributes_schema' );
```
psealock: Can you add a comment for the filter?

jeffstieler:
> I'm using before and after query args and getting the same `There is no data to export for the given request.` message. Here is my POST request url.

@psealock 

The export endpoint expects the report arguments to be nested under the `report_args` parameter. Try sending this as your JSON body:

```json
{
    "report_args": {
        "after": "2019-06-01T00:00:00",
        "before": "2019-06-30T00:00:00"
    }
}
```

docs/examples/extensions/table-column/woocommerce-admin-table-column.php
jeffstieler: Do you mean for where this filter is applied? https://github.com/woocommerce/woocommerce/blob/d67381148ac192502a9387ac00f9fb181df8dcc7/includes/abstracts/abstract-wc-rest-controller.php#L75

psealock: Oops, yes exactly.

includes/api/class-wc-admin-rest-reports-export-controller.php
```diff
@@ -0,0 +1,248 @@
...
+				__( 'Sorry, there is no export with that ID.', 'woocommerce-admin' ),
+				array( 'status' => 404 )
+			);
+		}
+
+		$result = array(
+			'percent_complete' => $percentage,
+		);
+
+		// @todo - add thing in the links below instead?
```
timmyc: add thing?

includes/export/class-wc-admin-report-csv-exporter.php
```diff
@@ -110,7 +114,11 @@ protected function map_report_controller() {
 		);
 
 		if ( isset( $controller_map[ $this->report_type ] ) ) {
-			// @todo: load the controllers if accessing outside a context where the REST API is loaded?
+			// Load the controllers if accessing outside the REST API.
+			if ( ! did_action( 'rest_api_init' ) ) {
```
timmyc: Do you know what would be calling this outside of the REST API context?

includes/export/class-wc-admin-report-exporter.php
```diff
@@ -0,0 +1,185 @@
...
+		// Initialize scheduled action handlers.
+		add_action( self::REPORT_EXPORT_ACTION, array( __CLASS__, 'report_export_action' ), 10, 4 );
+		// Report download handler.
+		add_action( 'admin_init', array( __CLASS__, 'download_export_file' ) );
+	}
+
+	/**
+	 * Queue up actions for a full report export.
+	 *
+	 * @param string $export_id Unique ID for report (timestamp expected).
```
timmyc: Is there any possibility of two users, both submitting requests for CSV exports at the same time? Edgey I suppose, but should we also append a user ID just in case?

includes/export/class-wc-admin-report-exporter.php
```diff
@@ -0,0 +1,185 @@
...
+	 * @param int    $percentage Completion percentage.
+	 * @return void
+	 */
+	public static function update_export_percentage_complete( $report_type, $export_id, $percentage ) {
+		$exports_status = get_option( self::EXPORT_STATUS_OPTION, array() );
+		$status_key     = self::get_status_key( $report_type, $export_id );
+
+		$exports_status[ $status_key ] = $percentage;
+
+		update_option( self::EXPORT_STATUS_OPTION, $exports_status );
```
timmyc: Heh, cool solution here.

includes/export/class-wc-admin-report-exporter.php
```diff
@@ -0,0 +1,185 @@
...
+		}
+
+		return false;
+	}
+
+	/**
+	 * Serve the export file.
+	 */
+	public static function download_export_file() {
+		// @todo - add nonce? (nonces are good for 24 hours)
```
timmyc: 🤔 if we add in a nonce control, then we will also have to send that via an email link. I'm pretty sure I've seen that pattern other places on wpcom. So nonce + session auth seems reasonable.

tests/api/reports-export.php
```diff
@@ -0,0 +1,148 @@
...
+
+		// Create a GA order.
+		$order = WC_Helper_Order::create_order( 1, $product );
+		$order->set_billing_city( 'Savannah' );
+		$order->set_billing_state( 'GA' );
+		$order->set_billing_postcode( '31401' );
+		$order->set_status( 'completed' );
+		$order->calculate_totals();
+		$order->save();
+
```
timmyc: Would this line be referred to as the _Florida Georgia Line_ in the code? :trollface: 🇺🇸

tests/api/reports-export.php
```diff
@@ -0,0 +1,148 @@
...
+		WC_Helper_Queue::run_all_pending();
+
+		// Check that the status shows 100% and includes a download url.
+		$response = $this->server->dispatch( new WP_REST_Request( 'GET', $status_route ) );
+		$status   = $response->get_data();
+
+		$this->assertEquals( 200, $response->get_status() );
+		$this->assertEquals( 100, $status['percent_complete'] );
+		$this->assertStringMatchesFormat( '%s/wp-admin/?action=woocommerce_admin_download_report_csv&filename=wc-taxes-report-export-%d', $status['download_url'] );
+		$this->assertStringMatchesFormat( '%s/wc/v4/reports/taxes/export/%d/status', $status['_links']['self'][0]['href'] );
```
timmyc: Great test flow here, thanks for the detailed test case. Should we add in a test for attempts to hit the endpoints un-authed too?

jeffstieler:
> Is it possible to test the contents of the CSV?

That was added in the previous PR: https://github.com/woocommerce/woocommerce-admin/blob/master/tests/reports/class-wc-tests-reports-coupons.php#L270

includes/api/class-wc-admin-rest-reports-export-controller.php
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	/**
+	 * Makes sure the current user has access to WRITE the settings APIs.
+	 *
+	 * @param WP_REST_Request $request Full data about the request.
+	 * @return WP_Error|bool
+	 */
+	public function export_permissions_check( $request ) {
+		if ( ! wc_rest_check_manager_permissions( 'settings', 'edit' ) ) {
```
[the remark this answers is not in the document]
jeffstieler: Ah good - catch. Some copy 🍝 from another controller.

I'll change this to `'reports', 'read'` instead, like the base reports controller class.

includes/api/class-wc-admin-rest-reports-export-controller.php
jeffstieler: Ha! Not my best comment..

I meant here: "should we add the download URL in the `_links` instead?

includes/export/class-wc-admin-report-csv-exporter.php
jeffstieler: Anything using the `WC_Admin_Report_CSV_Exporter` class directly. For example, we could build a WP-CLI command for exporting.

includes/export/class-wc-admin-report-exporter.php
jeffstieler: I initially included the user ID in the filename, but later removed it - see the commit history.

I chose to remove it because it doesn't make sense to restrict the files by user, and because we're already using microseconds in the export ID: https://github.com/woocommerce/woocommerce-admin/pull/2526/files/88366dd07c25cbbcacd3bc3a7267d55a9c63a705#diff-e885446237d58d740a1cbf34d483b5b5R163

includes/export/class-wc-admin-report-csv-exporter.php
timmyc: Ahhh, great idea, WP CLI would be a fantastic use case.

includes/export/class-wc-admin-report-exporter.php
jeffstieler: 😎

includes/export/class-wc-admin-report-exporter.php
jeffstieler: If we can punt for now, let's. I believe adding nonces could cause more problems if folks try to use stale email links.

tests/api/reports-export.php
jeffstieler: Done in f3a384d.

includes/api/class-wc-admin-rest-reports-export-controller.php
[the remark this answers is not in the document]
jeffstieler: Done in f3a384d.

PR #2531, opened 2019-06-27
Onboarding - Profile Wizard: Add tracks events

### LINKED ISSUE #2163: Onboarding - Profile Wizard: Implement Tracks
We will want to add tracks for users who opted in on the first screen. 

The list of events can be found via p1560449057014200-slack-wc-onboarding.

justinshreve:
Partially fixes #2163.

This PR implements all the tracks that we can for the profiler on the `wc-admin` side. Another PR for Calypso will be coming soon.

There are a couple more events that need implemented in the theme demo PR, cc @joshuatf,  if you don't mind handling the three `wcadmin_storeprofiler_store_theme_demo_` events from the spreadsheet.

* In your console, enter `localStorage.setItem( 'debug', 'wc-admin:tracks' );`
* Walk through the various steps and see the debug messages in your console.

### DIFF
--- a/client/dashboard/profile-wizard/steps/business-details.js
+++ b/client/dashboard/profile-wizard/steps/business-details.js
@@ -37,13 +37,9 @@ class BusinessDetails extends Component {
 		const { addNotice, goToNextStep, isError, updateProfileItems } = this.props;
 		const { other_platform, product_count, selling_venues } = this.state;
 
-		const alreadySelling = [ 'other', 'brick-mortar-other', 'brick-mortar' ].includes(
-			selling_venues
-		);
-
-		recordEvent( 'storeprofiler_store_business_details', {
+		recordEvent( 'storeprofiler_store_business_details_continue', {
 			product_number: product_count,
-			already_selling: alreadySelling,
+			already_selling: 'no' !== selling_venues,
 			used_platform: other_platform,
 		} );
 
--- a/client/dashboard/profile-wizard/steps/business-details.js
+++ b/client/dashboard/profile-wizard/steps/business-details.js
@@ -18,6 +18,7 @@ import { numberFormat } from '@woocommerce/number';
  */
 import { H, Card } from '@woocommerce/components';
 import withSelect from 'wc-api/with-select';
+import { recordEvent } from 'lib/tracks';
 
 class BusinessDetails extends Component {
 	constructor() {
@@ -35,6 +36,16 @@ class BusinessDetails extends Component {
 	async onContinue() {
 		const { addNotice, goToNextStep, isError, updateProfileItems } = this.props;
 		const { other_platform, product_count, selling_venues } = this.state;
+
+		const alreadySelling = [ 'other', 'brick-mortar-other', 'brick-mortar' ].includes(
+			selling_venues
+		);
+
+		recordEvent( 'storeprofiler_store_business_details', {
+			product_number: product_count,
+			already_selling: alreadySelling,
+			used_platform: other_platform,
+		} );
 
 		await updateProfileItems( { other_platform, product_count, selling_venues } );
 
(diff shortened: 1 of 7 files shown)


client/dashboard/profile-wizard/steps/business-details.js
```diff
@@ -36,6 +37,16 @@ class BusinessDetails extends Component {
 		const { addNotice, goToNextStep, isError, updateProfileItems } = this.props;
 		const { other_platform, product_count, selling_venues } = this.state;
 
+		const alreadySelling = [ 'other', 'brick-mortar-other', 'brick-mortar' ].includes(
```
joshuatf: Could we check if `selling_venues` is just blank or equal to `no`?  Probably set this to `no` as I found a bug here that needs fixing - https://github.com/woocommerce/woocommerce-admin/issues/2533

client/dashboard/profile-wizard/steps/business-details.js
```diff
@@ -36,6 +37,16 @@ class BusinessDetails extends Component {
...
+			selling_venues
+		);
+
+		recordEvent( 'storeprofiler_store_business_details', {
```
joshuatf: Nitpicking, but should these all be prefixed with `wc_store_profiler_` or `wc_admin_store_profiler_`?

client/dashboard/profile-wizard/steps/store-details.js
```diff
@@ -54,6 +55,8 @@ class StoreDetails extends Component {
 		const { addNotice, goToNextStep, isError, updateSettings } = this.props;
 		const { addressLine1, addressLine2, city, countryState, postCode } = this.state;
 
+		recordEvent( 'storeprofiler_store_details', { store_country: countryState.split( ':' )[ 0 ] } );
```
joshuatf: Out of curiosity, do we want to pass an object with a `store_country` param like this (as opposed to just recording an event `storeprofiler_store_country` with a string value of `US`) for the purpose of extending this param in the future?

justinshreve: The second parameter here (event properties) is supposed to be an object, and tracks itself expects a property name and not just a string.

justinshreve: Other properties also get appended to all event before storage in tracks (device, blog ID, etc).

client/dashboard/profile-wizard/steps/business-details.js
justinshreve: `wcadmin_` gets appended by the library, so these will actually become `wcadmin_storeprofiler_store_business_details` in tracks.

I think `storeprofiler_` works (and matches the spreadsheet) -- but let me know if you think we should update these.

client/dashboard/profile-wizard/steps/business-details.js
justinshreve: Haha yeah that's better. Fixed in https://github.com/woocommerce/woocommerce-admin/pull/2531/commits/957e23e92a6131ff847b2e2c800be7a1b7880ebd, thanks!

justinshreve:
I just pushed some changes here that fix up the naming as well -- I forgot to look at the updated tab of the spreadsheet.

client/dashboard/profile-wizard/steps/business-details.js
joshuatf: Looks like you're right.  I was confused since this line prefixes the event - https://github.com/woocommerce/woocommerce-admin/blob/81e76bfeb7be39d5a4fa6ac9200db27b53a1c1dd/client/header/activity-panel/index.js#L53

I'll log an issue to handle that.

>I think storeprofiler_ works (and matches the spreadsheet) -- but let me know if you think we should update these.

PR #2532, opened 2019-06-27
Tracks: Fix double wcadmin_ prefix.

timmyc:
This is a follow-up to #2498 - in that PR I accidentally added the `wc_admin_` prefix to the track name, and I forgot that the base tracks logic in Woo core already adds this in for us. This resulted in the track being currently recorded as `wc_admin_wc_admin_activity_panel_open`.

I also updated the debug statement here to prefix the event name shown in the debugger so hopefully others won't make the same mistake :)

**Before:**
prefix_debug

**After:**
tracks-afer

**To Test**
Follow the same testing instructions in #2498 to verify the event is being logged, and that it only has one `wc_admin` prefix.

### Changelog Note:

Fix: Track name for activity panel open event

### DIFF
--- a/client/lib/tracks.js
+++ b/client/lib/tracks.js
@@ -17,7 +17,7 @@ const tracksDebug = debug( 'wc-admin:tracks' );
  */
 
 export function recordEvent( eventName, eventProperties ) {
-	tracksDebug( 'recordevent %s %o', eventName, eventProperties );
+	tracksDebug( 'recordevent %s %o', 'wc_admin_' + eventName, eventProperties );
 
 	if (
 		! window.wcTracks ||
--- a/client/lib/tracks.js
+++ b/client/lib/tracks.js
@@ -12,7 +12,7 @@ const tracksDebug = debug( 'wc-admin:tracks' );
 /**
  * Record an event to Tracks
  *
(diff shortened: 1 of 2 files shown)


client/lib/tracks.js
```diff
@@ -17,7 +17,7 @@ const tracksDebug = debug( 'wc-admin:tracks' );
  */
 
 export function recordEvent( eventName, eventProperties ) {
-	tracksDebug( 'recordevent %s %o', eventName, eventProperties );
+	tracksDebug( 'recordevent %s %o', 'wc_admin_' + eventName, eventProperties );
```
psealock: Can you fix the comment in the doc block above?

timmyc: good find, updated.

client/lib/tracks.js
```diff
@@ -12,12 +12,12 @@ const tracksDebug = debug( 'wc-admin:tracks' );
 /**
  * Record an event to Tracks
  *
- * @param {String} eventName The name of the event to record, always prefixed with wca_
+ * @param {String} eventName The name of the event to record, always prefixed with wcadmin_
```
psealock: >always prefixed with wcadmin_

you'll need a negation in there 😄

Issue #2527, opened 2019-06-27
Implement Autocomplete Component


justinshreve:
For long `Select` components like country/state selection, we would like to implement an autocomplete like the one that WooCommerce core uses.

`@wordpress/components` has a component named "Autocomplete", that is used for things like the block inserter or `@mentions` -- however, it requires the use of a trigger character/word. This isn't quite what we need.

One option is to use one of the packages listed here:
https://material-ui.com/components/autocomplete/

See the discussion in p1561568187019900-slack-wc-onboarding

oliviertassinari:
> One option is to use one of the packages listed here:
https://material-ui.com/components/autocomplete/

@justinshreve Notice that Material-UI is moving away from third-party integration to provide its own solution: https://deploy-preview-17037--material-ui.netlify.com/components/autocomplete/
This is part of the "Material-UI" as material to build UI (and not only Material Design) effort.

PR #2534, opened 2019-06-28
Set default values for onboarding select dropdowns

### LINKED ISSUE #2533: Onboarding: Business details step doesn't allow you to select "No"
**Describe the bug**
On the business details step, "No" is the default option.  Clicking this won't trigger the change needed to show the continue button.

**To Reproduce**

1. Go to `wp-admin/admin.php?page=wc-admin#/?step=business-details`
2. Select number of products sold and then select "No" immediately from the selling elsewhere dropdown.
3. Note the continue button does not appear as form change is not triggered and considered invalid.

**Expected behavior**
Expected the continue button to appear.

**Screenshots**
Screen Shot 2019-06-28 at 2 00 17 PM

joshuatf:
Fixes #2533

Selecting the first select dropdown doesn't trigger an `onChange` event which doesn't select the option.  This is a temporary workaround to set the first option as selected if no option is currently set.

A better long-term option would be to look at being able to trigger events from the select dropdown or adding disabled options in the component to create default unselectable values in https://github.com/WordPress/gutenberg/blob/master/packages/components/src/select-control/index.js but if we're migrating to Select2 dropdowns, this may no longer be an issue.

### Screenshots
Screen Shot 2019-06-28 at 2 49 01 PM

1. Go to `wp-admin/admin.php?page=wc-admin#/?step=business-details`
2. Select the first option from each select dropdown (or just open the select menu and close).
3. Make sure they get selected and the "Continue" button appears.

### Changelog Note:

Fix: Select first option in onboarding select dropdowns when clicked.

### DIFF
--- a/client/dashboard/profile-wizard/steps/business-details.js
+++ b/client/dashboard/profile-wizard/steps/business-details.js
@@ -76,6 +76,12 @@ class BusinessDetails extends Component {
 		);
 	}
 
+	setDefaultValue( key, options ) {
+		if ( ! this.state[ key ].length ) {
+			this.setState( { [ key ]: options[ 0 ].value } );
+		}
+	}
+
 	render() {
 		const { other_platform, product_count, selling_venues } = this.state;
 
@@ -154,6 +160,7 @@ class BusinessDetails extends Component {
 					<SelectControl
 						label={ __( 'How many products will you add?', 'woocommerce-admin' ) }
 						onChange={ value => this.setState( { product_count: value } ) }
+						onClick={ this.setDefaultValue.bind( this, 'product_count', productCountOptions ) }
 						options={ productCountOptions }
 						value={ product_count }
 						required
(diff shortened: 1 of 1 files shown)


emdashcodes (commented):
This works well for clicking, but not keyboard navigation. Maybe `onFocus` is better?

> but if we're migrating to Select2 dropdowns, this may no longer be an issue.

I don't think we'll switch to Select2 necessarily as it's jQuery based and I'm not seeing any well maintained React select2 wrappers/packages.

I think `@wordpress/components ` offers what we need to build our own - a mix of `Dropdown` and `NavigableMenu` -- see https://github.com/WordPress/gutenberg/blob/master/packages/components/src/font-size-picker/index.js#L51-L89 as an example.

joshuatf:
> This works well for clicking, but not keyboard navigation. Maybe onFocus is better?

Good call.  Updated these and seems to work better, can you give it a check, @justinshreve?

>I don't think we'll switch to Select2 necessarily as it's jQuery based and I'm not seeing any well maintained React select2 wrappers/packages.

Gutenberg dropdown works just as well for me, provided it resolves those performance issues we're seeing with long lists like the country select dropdown.  I imagine it does, but haven't tested yet.

emdashcodes (approved):

justinshreve:
> Gutenberg dropdown works just as well for me, provided it resolves those performance issues we're seeing with long lists like the country select dropdown. I imagine it does, but haven't tested yet.

I haven't tested either yet.

Just a note, but the country input is switching to an autocomplete component: https://github.com/woocommerce/woocommerce-admin/issues/2527

PR #2537, opened 2019-06-28
Fixes issue with alerts on mobile

jobthomas:
Fixes a tiny issue with the notices on mobile - they were pushed below the header because the margin was too small

Before: alerts below bar
Link to image here: https://cld.wthms.co/cHXHtn.

After: alerts not below bar
Link to image here: https://cld.wthms.co/32Qf1F.


timmyc:
@jobthomas thanks for the PR!!

Issue #2541, opened 2019-06-28
Settings: Confirmation alert not shown upon save


timmyc:
**Describe the bug**
While working on some fixes to #2449 I was noticing that the `addNotice()` confirmations were not being shown in the UI after the settings were persisted. At first I thought the problem was due to my branch, but I verified the bug is present in `master` too.

In my debugging it seems like [isGetSettingsRequested()](https://github.com/woocommerce/woocommerce-admin/blob/f6b70fafec8bd689854d713f21b86415cb260fe2/client/wc-api/settings/selectors.js#L33) is always returning `true` because `lastReceived` doesn't seem to be updating properly ( maybe ):

<img width="312" alt="image" src="https://user-images.githubusercontent.com/22080/60365311-36cf9600-999d-11e9-96fa-8858fcbff3d3.png">

**To Reproduce**

1. Analytics | Settings
2. Make a change to an actionable order status
3. Click Save
4. note no message is shown to confirm the settings are changed.

**Expected behavior**
Should see a notice that my settings have been saved.

timmyc:
Just noting this bug is still present in `master` today.

PR #2544, opened 2019-06-28
Settings: Better structure to settings page, and protect form changes

### LINKED ISSUE #2449: Settings: Form is confusing, no warning if settings not saved
With the addition of the Historical Data Import to the settings page, the form for the settings seems a bit confusing to me now:

image

Due to the way the fieldsets on this page are structured - to me it seems like the save/reset buttons are just for the Default Date range. 

Perhaps if we created another `woocommerce-section-header` for the historical data import, it might visually make the action buttons appear tied more to all the settings.

Another item noted in testing is if you modify a setting and do not save - you can navigate away from the page and not be warned that your changes have not been saved.

timmyc:
Fixes #2449

This branch seeks to make the Analytics | Settings page a bit less confusing looking by adding in a proper section header to the Historical Data area, and it adds tracking to the state to set the page as `isDirty` if any changes are made, and subsequently warns users they have unsaved changes prior to navigating away.

**Before**
image

**After**
image

**Notes / Known Issues**
_Attempted to Core_
I initially wanted to use WP Core's `<UnsavedChangesWarning />` for this but it has a [hardcoded state check](https://github.com/WordPress/gutenberg/blob/d95c7d61d8ca284394a5aa77e9f74f9b3aaea24e/packages/editor/src/components/unsaved-changes-warning/index.js#L44) that is tied into the editor's state. In a prior life, this core component supported manually setting the dirty state via a prop, but that was [deprecated last october](https://github.com/WordPress/gutenberg/pull/10952) so I opted to _emulate_ their approach here.

_Warning not shown when navigating via react-router_
The warning of unsaved changes only works on the settings page if you are navigating to a non-wc-admin page, or if doing a hard refresh. I found [this awesome approach](https://stackoverflow.com/questions/32841757/detecting-user-leaving-page-with-react-router) on stackoverflow to use a built in tool from `react-router` to add support for this events, but after implementing I was hitting [this bug in react-router](https://github.com/ReactTraining/react-router/issues/5405) - which even though using the `<Prompt />` component from react-router to stop navigation works, the address bar still updates and causes all sorts of janky-behavior.

_Saving settings prompt not shown_
While working on this branch I also noticed that when I was saving new settings, the notice was not being shown, and the `isDirty` state was not being reset. At first I thought this was a bug in this branch, but indeed it exists in master. So I opened #2541 and added in a temporary fix here to optimistically set the form state back to clean when a save event is triggered. 46968fd

- Open Analytics | Settings - note how amazing the new Historical Data Import section header looks.
- After the awe has subsided, change the settings form above, then attempt to navigate to a non-wc-admin page ( like Products ), and prepare yourself for some JS alert goodness.
- Click cancel, and save your change, then navigate away happily.

### Changelog Note:

Fix: Update layout of Settings Page and notify users when settings are not saved.

### DIFF
--- a/client/analytics/settings/index.js
+++ b/client/analytics/settings/index.js
@@ -36,9 +36,19 @@ class Settings extends Component {
 		this.state = {
 			settings,
 			saving: false,
+			isDirty: false,
 		};
 
 		this.handleInputChange = this.handleInputChange.bind( this );
+		this.warnIfUnsavedChanges = this.warnIfUnsavedChanges.bind( this );
+	}
+
+	componentDidMount() {
+		window.addEventListener( 'beforeunload', this.warnIfUnsavedChanges );
+	}
+
+	componentWillUnmount() {
+		window.removeEventListener( 'beforeunload', this.warnIfUnsavedChanges );
 	}
 
 	componentDidCatch( error ) {
@@ -48,6 +58,18 @@ class Settings extends Component {
 		/* eslint-disable no-console */
 		console.warn( error );
 		/* eslint-enable no-console */
+	}
+
+	warnIfUnsavedChanges( event ) {
+		const { isDirty } = this.state;
+
+		if ( isDirty ) {
+			event.returnValue = __(
+				'You have unsaved changes. If you proceed, they will be lost.',
+				'woocommerce-admin'
+			);
+			return event.returnValue;
+		}
 	}
 
 	resetDefaults = () => {
@@ -64,7 +86,8 @@ class Settings extends Component {
 
 	componentDidUpdate() {
 		const { addNotice, isError, isRequesting } = this.props;
-		const { saving } = this.state;
+		const { saving, isDirty } = this.state;
+		let newIsDirtyState = isDirty;
 
 		if ( saving && ! isRequesting ) {
 			if ( ! isError ) {
@@ -72,6 +95,7 @@ class Settings extends Component {
 					status: 'success',
 					message: __( 'Your settings have been successfully saved.', 'woocommerce-admin' ),
 				} );
+				newIsDirtyState = false;
 			} else {
 				addNotice( {
 					status: 'error',
@@ -82,7 +106,7 @@ class Settings extends Component {
 				} );
 			}
 			/* eslint-disable react/no-did-update-set-state */
-			this.setState( { saving: false } );
+			this.setState( { saving: false, isDirty: newIsDirtyState } );
 			/* eslint-enable react/no-did-update-set-state */
 		}
 	}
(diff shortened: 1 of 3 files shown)


client/analytics/settings/index.js
```diff
@@ -50,6 +60,18 @@ class Settings extends Component {
...
 
+	warnIfUnsavedChanges( event ) {
+		const { isDirty } = this.state;
+
+		if ( isDirty ) {
+			event.returnValue = __(
+				'You have unsaved changes. If you proceed, they will be lost.',
+				'woocommerce-admin'
+			);
+			return event.returnValue;
```
psealock: Where does this string get rendered?

Chrome
Screen Shot 2019-07-01 at 1 05 16 PM

FF
Screen Shot 2019-07-01 at 1 07 56 PM

client/analytics/settings/index.js
```diff
@@ -36,9 +36,19 @@ class Settings extends Component {
...
 			saving: false,
+			isDirty: false,
 		};
 
 		this.handleInputChange = this.handleInputChange.bind( this );
+		this.warnIfUnsavedChanges = this.warnIfUnsavedChanges.bind( this );
+	}
+
+	componentDidMount() {
+		window.addEventListener( 'beforeunload', this.warnIfUnsavedChanges );
```
psealock: how about `history.listen()` for internal navigation?

https://github.com/ReactTraining/history

timmyc:
Some good feedback there @jameskoster - @LevinMedia might have some responses to the history of this being on the settings page. I think users might use the tool more than once, but I agree that it should probably live elsewhere in the long-run

client/analytics/settings/index.js
timmyc: Those are built into the browsers [#](https://stackoverflow.com/questions/51681574/how-to-get-the-leave-site-changes-that-you-made-may-not-be-saved-when-trying/51681800)

client/analytics/settings/index.js
timmyc: So I just tried this out, and it appears to have the same issue that I was experiencing with `<Prompt />` i.e. if I hit cancel via a dialog thrown via `history.block()`, the address bar still updates. I'm guessing `Prompt` might just be wrapping this history module :)

timmyc:
@psealock checked out the option, wasn't working for me - what do you think about getting this in as-is?

psealock:
>I'm also wondering if there's a situation where someone would run this import more than once? 

One possible scenario is someone has chosen to import a subset of data (last 3 months) and wants to come back to the tool to import the rest. Or, maybe a user has imported historical data manually via sql dump and would like to import it to wc-admin after already having done so initially.

>we might consider prompting folks to perform the import upon activation of wc-admin, and avoid sending them to a separate screen at all.

Nonetheless, this is a good idea. How about a notification?

PR #2551, opened 2019-07-01
Onboarding - Add business extension recommendations

### LINKED ISSUE #2518: Onboarding - Profile Wizard: Add Facebook & MailChimp to the Business Tab
See p1561555686016700-slack-wc-onboarding and p1560865236005400-slack-wc-onboarding

The MailChimp and Facebook extensions should be added to the business tabs with toggles for the user to choose if they want them installed or not.

We should save this in the profiler data, because we'll need to pass these slugs along to WooCommerce.com to install.

justinshreve:
Fixes #2518.

This PR adds the Mailchimp and Facebook extensions to the business tab of the profile wizard.

Screen Shot 2019-07-01 at 4 54 59 PM

Note: I noticed certain values for `other_platform` are causing save endpoints to fail with a 400, even though they should be valid. This is not addressed in this PR. I can take a look at this tomorrow and fix in a follow-up PR.

* Go to `/wp-admin/admin.php?page=wc-admin#/?step=business-details`
* Fill out the form and see that the extension recommendations display. Test the toggles, and saving your profile data. You can request `GET /wp-json/wc-admin/v1/onboarding/profile` to make sure the extensions saved.


client/dashboard/profile-wizard/steps/business-details.js
```diff
@@ -83,6 +92,85 @@ class BusinessDetails extends Component {
...
+
+		const extensions = keys( pickBy( this.state.extensions ) );
+		if ( 0 === extensions.length ) {
+			return null;
+		}
+
+		return (
+			<p>
+				{ sprintf(
+					__( 'The following plugins will be installed for free: %s' ),
```
joshuatf: 👍

client/dashboard/profile-wizard/style.scss
```diff
@@ -213,6 +213,112 @@
...
+	}
+
+	.woocommerce-profile-wizard__product-types-card {
+		.woocommerce-profile-wizard__benefit-content {
+			margin-left: 12px;
+		}
+	}
+
+	.woocommerce-profile-wizard__tracking {
+		.components-form-toggle {
```
joshuatf: My toggles look a little off-centered vertically:

Screen Shot 2019-07-02 at 10 20 41 AM

includes/api/class-wc-admin-rest-onboarding-profile-controller.php
```diff
@@ -283,15 +283,27 @@ public static function get_profile_properties() {
 					'other',
 				),
 			),
-			'theme'           => array(
+			'business_extensions' => array(
```
joshuatf: We'll need a couple quick changes in  `tests/api/onboarding-profile.php` to get tests ✅

justinshreve: Done.

justinshreve:
I've fixed the phpunit test.

@joshuatf I can't seem to replicate the off-center toggles. Your toggles look a bit bigger than mine, which could affect the positioning, so I did add a max height/width. Does that help? If not, could you see what CSS is causing that? I'm not having issues in Chrome or Safari.

joshuatf:
@justinshreve Odd, I'm seeing it in all browsers.  I wonder if there's another plugin altering the behavior here.

>Your toggles look a bit bigger than mine

I think I may have been zoomed in.  Mine appear the same size as the ones in your screenshot.

The span here appears a bit off:

Screen Shot 2019-07-03 at 4 13 54 PM

But adding display `block` or `inline-block` to the toggle seems to position that span at the top:

`.components-form-toggle {
    display: inline-block;
}`

Screen Shot 2019-07-03 at 4 14 35 PM

justinshreve:
@joshuatf Thanks -- I agree there must be some plugin difference (or maybe Gutenberg difference) because I already had that class coming in from wp components, but I've added that class in our CSS as well.

I've also adjusted when the extensions display so that the extensions don't disappear/reappear as you select from the selling dropdown (which happened after rebase now that we set a default onFocus).

justinshreve:
> It feels like these styles should be included in @wordpress/components, but I guess that would create duplication between the WP included styles when updating. Is there some discussion around this already that I can read up on?

Good point... I agree. I did a quick search but don't see any discussion around that.

PR #2562, opened 2019-07-03
Header: Adjust sizes at all breakpoints

### LINKED ISSUE #2287: Header Bar Is a Bit Too Tall On Wide But Short Screens
**The header bar is a bit too tall on my MacBook Pro**
When using my external monitor, I have additional vertical space and enjoy having the header bar present in WooCommerce.

But on my MacBook, it occupies about 10% of my available space vertically.  It's not much, but I find myself disabling the plugin to gain that space back.  

Here are some screenshots for comparison.

With header bar
(Link to screenshot:  https://d.pr/rX2NWk)

Without header bar
(Link to screenshot:  https://d.pr/DS0vwQ)

It would be nice to have the header bar be shorter if the browser window is not as tall.  There is a media query for the width already.  One for the height would be nice too.  

A few of my teammates have noticed the same thing.  Based on that sample of daily WooCommerce users this could be something that affects a fair number of users once this is part of the core.

### LINKED ISSUE #2337: Content editor toolbar is hidden by admin header
**Describe the bug**
The admin header is added to every WooCommerce page, including when we are editing products. However WordPress has a feature to show a sticky header when editing the content. This is especially apparant with products that have a lot of text content.
Which means as soon as we need to use the toolbar or the add media button when editing the content we need to scroll all the way up again.

**To Reproduce**

1. Go to edit a product, and add a large amount of text content
2. Make sure the screen option "Enable full-height editor and distraction-free functionality" is turned on.
3. Scroll down to the bottom of the product text content
4. Observe the toolbar is hidden

**Expected behavior**
I would expect the editing toolbar to take priority. Or alternatively have a way to disable the admin header so it doesn't show up on some page, like when editing products.

**Screenshots**
Here is an example of the toolbar being hidden when editing the product content:
image

Here is the same example with the admin plugin deactivated:
image

timmyc:
Fixes #2337 
Fixes #2287 

From discussion in both the issues above, this PR changes the sizing of the Header and Activity Panels to a uniform 56px across all breakpoints. By doing so, this also addresses the bug that was causing the TinyMCE toolbar from being hidden behind the header on products with a large amount of content in the description.

### Screenshots

_Product Edit Page Before with obscured toolbar_
product-toolbar-before

_Product Edit Page with change_
product-toolbar-scroll-after

_Dashboard Before_
dashboard-before

_Dashboard After_
dashboard-after

_Breakpoints in action_
breakpoints

- Browse both wc-admin js pages, and embeded pages at a variety of viewport widths
- Open/close activity panels at the various widths and page types and ensure all is happy.### Changelog Note:

Fix: Change size of Header to 56px - fixes bug in product edit page toolbar being hidden.

/cc @mikkamp

### DIFF
--- a/client/analytics/components/report-table/style.scss
+++ b/client/analytics/components/report-table/style.scss
@@ -11,12 +11,8 @@
 	.woocommerce-feature-enabled-activity-panels & {
 		top: -#{$adminbar-height + $large-header-height + $gap};
 
-		@include breakpoint( '<782px' ) {
+		@include breakpoint( '<1280px' ) {
 			top: -#{$adminbar-height-mobile + $small-header-height + $gap};
-		}
-
-		@include breakpoint( '782px-960px' ) {
-			top: -#{$adminbar-height + $medium-header-height + $gap};
 		}
 	}
 }
--- a/client/analytics/components/report-table/style.scss
+++ b/client/analytics/components/report-table/style.scss
@@ -9,10 +9,10 @@
 	}
 
 	.woocommerce-feature-enabled-activity-panels & {
-		top: -#{$adminbar-height + $large-header-height + $gap};
+		top: -#{$adminbar-height + $header-height + $gap};
 
-		@include breakpoint( '<1280px' ) {
-			top: -#{$adminbar-height-mobile + $small-header-height + $gap};
+		@include breakpoint( '<782px' ) {
+			top: -#{$adminbar-height-mobile + $header-height + $gap};
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


timmyc:
Thanks for the review @jameskoster - I really like the smaller icons/header.

dechov:
As seen in the screenshots above, the unread indicator appears lower on the icon – is this expected?

<=960px | >960px | >960px without `top: 16px;`
-- | -- | --
Screen Shot 2019-07-16 at 6 32 06 PM | Screen Shot 2019-07-16 at 6 32 29 PM | Screen Shot 2019-07-16 at 6 34 49 PM

timmyc:
@dechov good catch, that appears to be a regression from this PR.

Issue #2470, opened 2019-06-20
Onboarding - Profile Wizard: Theme Upload


justinshreve:
Depends on #2159. This issue involves adding a "upload theme" option to the theme browser, allowing users to supply their own .zip file.

PR #2568, opened 2019-07-03
Add a theme REST API to allow uploading themes

joshuatf:
Fixes (part of) #2470

Adds a theme rest api so we can upload themes from a theme uploader component in onboarding.

### Screenshots
Screen Shot 2019-07-03 at 3 49 56 PM

1. Using the body/form-data params, send a POST request to `/wp-json/wc-admin/v1/themes` with the `pluginzip` param set to a theme file.
2. Make sure that you can upload a valid zipped theme.
3. Make sure the theme data needed for the profiler is filtered into the response.
3. Try also uploading a theme that already exists or invalid zip to ensure errors are returned.

### Changelog Note:

Enhancement: Add a themes REST API for uploading themes.

### DIFF
--- a/includes/api/class-wc-admin-rest-themes-controller.php
+++ b/includes/api/class-wc-admin-rest-themes-controller.php
@@ -56,8 +56,8 @@ class WC_Admin_REST_Themes_Controller extends WC_REST_Data_Controller {
 	 * @return WP_Error|boolean
 	 */
 	public function upload_theme_permissions_check( $request ) {
-		if ( ! current_user_can( 'upload_plugins' ) ) {
-			return new WP_Error( 'woocommerce_rest_cannot_view', __( 'Sorry, you are not allowed to install plugins on this site.', 'woocommerce-admin' ), array( 'status' => rest_authorization_required_code() ) );
+		if ( ! current_user_can( 'upload_themes' ) ) {
+			return new WP_Error( 'woocommerce_rest_cannot_view', __( 'Sorry, you are not allowed to install themes on this site.', 'woocommerce-admin' ), array( 'status' => rest_authorization_required_code() ) );
 		}
 
 		return true;
--- a/includes/api/class-wc-admin-rest-themes-controller.php
+++ b/includes/api/class-wc-admin-rest-themes-controller.php
@@ -85,11 +85,10 @@ class WC_Admin_REST_Themes_Controller extends WC_REST_Data_Controller {
 		}
 
 		if ( ! is_wp_error( $install ) && isset( $install['destination_name'] ) ) {
-			$theme  = wp_get_theme( $install['destination_name'] );
 			$result = array(
 				'status'  => 'success',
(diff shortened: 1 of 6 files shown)


includes/api/class-wc-admin-rest-themes-controller.php
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	/**
+	 * Check whether a given request has permission to edit upload plugins/themes.
+	 *
+	 * @param  WP_REST_Request $request Full details about the request.
+	 * @return WP_Error|boolean
+	 */
+	public function upload_theme_permissions_check( $request ) {
+		if ( ! current_user_can( 'upload_plugins' ) ) {
```
justinshreve: Should this be `upload_themes` instead? https://wordpress.org/support/article/roles-and-capabilities/#upload_themes

includes/api/class-wc-admin-rest-themes-controller.php
```diff
@@ -0,0 +1,190 @@
...
+			return new WP_Error( 'woocommerce_rest_cannot_view', __( 'Sorry, you are not allowed to install plugins on this site.', 'woocommerce-admin' ), array( 'status' => rest_authorization_required_code() ) );
```
justinshreve: Nitpick: This should say themes.

emdashcodes (approved):
I was wondering if the endpoint needed some file type validation, but it looks like the upload and package functions take care of that.

Tested with a valid theme and it worked. Tested with duplicate and got the correct error. Tested with a plugin file and a random image/test file -- those failed as well.

I just have two comments around the permissions check, but otherwise pre-approving.

tests/api/themes.php
```diff
@@ -0,0 +1,63 @@
...
+			array(
+				'role' => 'administrator',
+			)
+		);
+	}
+
+	/**
+	 * Test that an error is thrown if the file is missing.
+	 */
+	public function test_missing_upload() {
```
timmyc: Might be nice to add a test case for a user without the proper role attempting to upload a theme.

joshuatf: Added in https://github.com/woocommerce/woocommerce-admin/pull/2568/commits/36f9325228a0041f5c1e00fdc3743659b3992582

includes/api/class-wc-admin-rest-themes-controller.php
joshuatf: I felt I ran across `upload_plugins` when digging through the functions in core, so figure it might be some weird pseudo linked privilege.  But I can't seem to find it right now and `upload_themes` definitely seems better fitting.

joshuatf:
Thanks for the review and feedback @timmyc & @justinshreve!

>I was wondering if the endpoint needed some file type validation, but it looks like the upload and package functions take care of that.

I was wondering the same.  Initially I added in more client-side validation, but removed it since relying on actual response from core may be less error-prone.

PR #2569, opened 2019-07-03
Add countLabel prop to SearchListItem

Aljullu:
In the _Reviews by Product_ block we need to display _X Reviews_ next to each product in the `SearchListControl`:
image

Currently, `SearchListItem` only allows showing a number (which comes from `item.count`). This PR adds a `countLabel` prop to `SearchListItem` that will overwrite what's rendered in the counter. Allowing the consumer to modify it.

### Accessibility

### Screenshots
image

- Modify this line:
https://github.com/woocommerce/woocommerce-admin/blob/master/packages/components/src/search-list-control/index.js#L107
  With this or something similar:
```ES6
		return <SearchListItem countLabel="Lorem Ipsum" showCount { ...args } />;
```
- Go to _WooCommerce_ > _DevDocs_.
- Go to _SearchListControl_ card and press _Toggle loading state_.
- Verify the count label appears next to the list names (see screenshot above).

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Card component: new `description` prop.
 - Card component: new `isInactive` prop.
 - DateRangeFilterPicker (breaking change): Introduced `onRangeSelect` prop and remove `path` prop better control.
+- SearchListItem component: new `countLabel` prop that will overwrite the `item.count` value.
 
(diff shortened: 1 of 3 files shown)


timmyc:
> Would it be possible to get this in the package so we can use it?

Indeed lets merge this and release a new version of `components` so you all can consume it in blocks 👍

PR #2571, opened 2019-07-03
ReportTable: Add back in ids array generation.

### LINKED ISSUE #2570: Analytics: Categories Report Table Crashes when selecting a category to compare
**Describe the bug**
[Original Report from the Forums](https://wordpress.org/support/topic/category-tickbox/)

When selecting a category checkbox in the table to enter comparison mode, an error occurs and the page crashes

**To Reproduce**

1. Go to Categories Report, where data is shown
2. Check the box next to a category in the table
3. :boom:

**Screenshots**

image

timmyc:
Fixes #2570

It appears this regression was introduced as part of #2440 ( /cc @psealock  )- which added in the new filter functionality for the table, but while doing so removed the generation of the `ids` prop that is subsequently passed to `<Table />` and used for the comparison logic.

- First verify the error on `master` by opening any report that offers compare functionality in the table ( Categories, Taxes, Products etc ).
- Click on a checkbox in a table row, note the error
- Apply this branch, repeat the steps and verify you can compare multiple items, and no error occurs.

### Changelog Note:

Fix: Compare checkboxes in report tables

### DIFF
--- a/client/analytics/components/report-table/index.js
+++ b/client/analytics/components/report-table/index.js
@@ -123,6 +123,7 @@ class ReportTable extends Component {
 		const { headers, ids, rows, summary } = applyFilters( TABLE_FILTER, {
 			endpoint,
 			headers: getHeadersContent(),
+			ids: itemIdField ? items.data.map( item => item[ itemIdField ] ) : [],
 			rows: getRowsContent( items.data ),
 			totals,
 			summary: getSummary ? getSummary( totals, totalResults ) : null,


timmyc:
> Since items was added to this array, the ids property is redundant information.

Yes that does sound a bit redundant - and maybe is something we can look at changing later. My vote is just to get this PR merged in and out in the next release to fix the bug. To do the refactor on table to have it in charge of creating the array of `ids` would require a full version bump on components due to it being a breaking change so it seems like just getting this in is the quickest way to fix.

PR #2573, opened 2019-07-03
Notes: Type cast is_snoozeable on insert.

### LINKED ISSUE #2572: Notes: MySql insert log message invalid type for is_snoozable.
**Describe the bug**
[Original report in the forums](https://wordpress.org/support/topic/wordpress-database-error-incorrect-integer-value-for-column-is_snoozable)

From the forum report it looks like the [no connection to woocommerce.com](https://github.com/woocommerce/woocommerce-admin/blob/master/includes/notes/class-wc-admin-notes-woo-subscriptions-notes.php#L180) note is triggering a message in the site's logs around the value being inserted for `is_snoozable` not being casted to an int properly.

I was not able to replicate the warning/log message myself, but I can see how the problem could happen.

timmyc:
Fixes #2572

While I wasn't able to fully re-create the initial bug, this branch seeks to properly type cast the `is_snoozable` value before insertion into the notes database table. I've also modified the unit tests to test the setting/retrieval of the `is_snoozable` value.

- Verify unit tests pass

### Changelog Note:

Fix: Cast is_snoozable as int prior to database save

### DIFF
--- a/includes/data-stores/class-wc-admin-notes-data-store.php
+++ b/includes/data-stores/class-wc-admin-notes-data-store.php
@@ -31,7 +31,7 @@ class WC_Admin_Notes_Data_Store extends WC_Data_Store_WP implements WC_Object_Da
 			'icon'         => $note->get_icon(),
 			'status'       => $note->get_status(),
 			'source'       => $note->get_source(),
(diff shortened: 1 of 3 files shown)


timmyc:
Thanks for the review @rrennick 🙌

PR #2577, opened 2019-07-03
Settings: Fix wsod when no actionable statuses are set.

timmyc:
There have been a number of bug reports in the forums akin to [this one](https://wordpress.org/support/topic/shows-blank-screen/page/2/#post-11697728) where all screens in wc-admin are white-screen-o-deathing with an error message like:

image

I then looked at spread operator usage across the code base, in hopes that something jumped out at me that would be a code path eval'ed on all wc-admin pages, and I focused in on the fix presented here. While I'm not 100% certain this is the cause of all the reports of this bug, this PR fixes a flow where the exact same error is thrown, and crashes all wc-admin pages.

I'm hoping the user [here](https://wordpress.org/support/topic/shows-blank-screen/page/2/#post-11697904) will respond with a screenshot to confirm that is what caused this issue on their site 🤞 

### Steps to create the bug

- On `master` visit Analytics | Settings
- Uncheck all boxes under the actionable order statuses section
- Click save settings, and refresh the page, note the error

- Apply this branch, and repeat the steps above.

### Changelog Note:

Fix: Prevent error when no actionable order statuses are set

### DIFF
--- a/client/analytics/settings/config.js
+++ b/client/analytics/settings/config.js
@@ -24,6 +24,13 @@ const defaultOrderStatuses = [
 	'pending',
 	'on-hold',
 ];
+
+const actionableOrderStatuses = Array.isArray(
+	wcSettings.wcAdminSettings.woocommerce_actionable_order_statuses
+)
+	? wcSettings.wcAdminSettings.woocommerce_actionable_order_statuses
+	: [];
+
 const orderStatuses = Object.keys( wcSettings.orderStatuses )
 	.filter( status => status !== 'refunded' )
 	.map( key => {
@@ -87,7 +94,7 @@ export const analyticsSettings = applyFilters( SETTINGS_FILTER, [
 				'These orders will show up in the Orders tab under the activity panel.',
 			'woocommerce-admin'
 		),
-		initialValue: [ ...wcSettings.wcAdminSettings.woocommerce_actionable_order_statuses ] || [],
+		initialValue: actionableOrderStatuses,
 		defaultValue: DEFAULT_ACTIONABLE_STATUSES,
 	},
 	{
--- a/client/analytics/settings/config.js
+++ b/client/analytics/settings/config.js
@@ -31,6 +31,12 @@ const actionableOrderStatuses = Array.isArray(
 	? wcSettings.wcAdminSettings.woocommerce_actionable_order_statuses
 	: [];
 
+const excludedOrderStatuses = Array.isArray(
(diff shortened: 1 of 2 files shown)


client/analytics/settings/config.js
```diff
@@ -24,6 +24,13 @@ const defaultOrderStatuses = [
 	'pending',
 	'on-hold',
 ];
+
+const actionableOrderStatuses = Array.isArray(
+	wcSettings.wcAdminSettings.woocommerce_actionable_order_statuses
+)
+	? wcSettings.wcAdminSettings.woocommerce_actionable_order_statuses
+	: [];
+
```
psealock: We may as well do the same for `wcSettings.wcAdminSettings.woocommerce_excluded_report_order_statuses`.

client/analytics/settings/config.js
```diff
@@ -24,6 +24,13 @@ const defaultOrderStatuses = [
 	'pending',
 	'on-hold',
 ];
+
+const actionableOrderStatuses = Array.isArray(
+	wcSettings.wcAdminSettings.woocommerce_actionable_order_statuses
```
Aljullu: Good catch with this @timmyc! Per the discussion [here](https://github.com/woocommerce/woocommerce-admin/pull/2292#discussion_r287166239), what do you think about cloning the array before the assignment? I guess now that we are checking that it's an array, using the spread operator should be safe:
```suggestion
	[ ...wcSettings.wcAdminSettings.woocommerce_actionable_order_statuses ]
```

Aljullu:
> I'm thinking to cover all of these, maybe we could nip this at the root and include all settings even if they're empty:

@joshuatf That sounds like a good solution too. Maybe we could apply both of them to play it safe in case somebody or an extension makes weird modifications in the database?

client/analytics/settings/config.js
timmyc: Cloning seems like a good idea - as for the spread operator, i wasn't sure why we would need it if we verify it is an array.

timmyc:
> Maybe we could apply both of them to play it safe in case somebody or an extension makes weird modifications in the database?

Indeed I think we just always never trust anything 😆 even our own API responses we should validate formats before assuming because you just never know what code will exist on a site that might modify things.

client/analytics/settings/config.js
timmyc: I see that the spread was cloning before 🥇

timmyc:
Added in a few more updates after your reviews, thanks everyone.

PR #2594, opened 2019-07-05
Tracks: Fix Link onclick

psealock:
https://github.com/woocommerce/woocommerce-admin/pull/2444 Changed logic involving `onClick` props passed to `<Link />` components such that onclicks weren't being executed, including Tracks events.

This PR makes 3 fixes:

* Make sure onclicks are executed.
* Snake case Tracks events property names for `wcadmin_analytics_filters_filter`.
* Fix logic that enables/disables the "Filter" button because it was based on the old url scheme.

1. In your console `localStorage.setItem( 'debug', 'wc-admin:*' );`
2. Analytics > Orders > Advanced Filters
3. Choose a filter or two, click "Filter"
4. See the event fired.
5. See the "Filter" button disable.

We don't have any configs that are in camel case, so I'm not really sure how to test that part.

### DIFF
--- a/client/analytics/components/report-filters/index.js
+++ b/client/analytics/components/report-filters/index.js
@@ -4,7 +4,7 @@
  */
 import { Component } from '@wordpress/element';
 import PropTypes from 'prop-types';
-import { omitBy, isUndefined } from 'lodash';
+import { omitBy, isUndefined, snakeCase } from 'lodash';
 
 /**
  * WooCommerce dependencies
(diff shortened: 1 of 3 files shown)


rrennick:
@psealock This works. I noticed that if I change the date range after I click the filter it removes the border/shadow indicating that the button has been clicked. Clicking again has no effect except to restore the border/shadow. I'm not sure if that's related to this PR or should be addressed separately.

psealock:
Thanks for the review @rrennick.

I'm not following exactly. After I click "Filter", the UI looks like the screenshot below. If I then change the date range, the "Filter" button should not be re-enabled because the new date range is applied already.

Screen Shot 2019-07-08 at 11 51 40 AM

Is this what you are referring to?

rrennick:
> Is this what you are referring to?

No, but on switching back to this branch I am no longer seeing the behaviour. I may have had cross pollination with another branch.

psealock:
>No, but on switching back to this branch I am no longer seeing the behaviour.

hmmm, I did rebase off master. Maybe it was fixed elsewhere. Thanks for the review.

PR #2596, opened 2019-07-05
Update to new Snackbar component and notice store

### LINKED ISSUE #2445: Replace current notices with snackbars
From https://github.com/woocommerce/woocommerce-admin/issues/2175#issuecomment-493070920

> For the notices we are using the Material snackbar design aligned to the center to page without the wp-admin sidebar. 
> 
> https://material.io/design/components/snackbars.html#
> 
> This should override the notices style we were using before, one example is in the Analytics settings:
> 
> Screenshot 2019-05-15 at 10 52 25

Gutenberg `<Snackbar>` component can be found here:

https://github.com/WordPress/gutenberg/tree/master/packages/components/src/snackbar

joshuatf:
Fixes #2445

Replaces all existing transient notices with the `Snackbar` component.  Also replaces the existing store with the `@wordpress/notices` store.

### Screenshots
Screen Shot 2019-07-05 at 1 51 27 PM

1. Make sure things aren't broken due to dependency changes.
2. Make sure notices still work where they previously did (for example, historical data import notices).
3. Make sure notices disappear after 10 seconds and clicking on them removes them.

### Changelog Note:

Enhancement: Use core Snackbar notices for transient notifications.


webpack.config.js
```diff
@@ -26,8 +26,6 @@ const WC_ADMIN_CONFIG = require( path.join( __dirname, 'config', WC_ADMIN_PHASE
 const externals = {
 	'@wordpress/api-fetch': { this: [ 'wp', 'apiFetch' ] },
 	'@wordpress/blocks': { this: [ 'wp', 'blocks' ] },
-	'@wordpress/components': { this: [ 'wp', 'components' ] },
```
joshuatf: I'm not sure if removing these has other implications, but this prevented new components from working due to missing functions.

justinshreve: I'm not sure about this either. Things seem to be working well for me though with this change.

PR #2598, opened 2019-07-05
Implement "Simple" Select Component

### LINKED ISSUE #2528: Implement "Simple" Dropdown Mode for Select Component
Our current `Select` component (imported from Newspack) implements "native" selection:

Screen Shot 2019-06-27 at 12 58 36 PM

https://material-ui.com/components/selects/#native-select

WooCommerce Core already uses a style similar to "simple", This issue involves implementing a Select component with "simple" styles to use in onboarding and `wc-admin`.

https://material-ui.com/components/selects/#simple-select

justinshreve:
Fixes #2528.

This PR implements a new Muriel styled select component, modeled after core's `SelectControl`. It implements the dropdown using Material's "simple" dropdown style, instead of the native style we were using before.

See https://material-ui.com/components/selects/#simple-select vs https://material-ui.com/components/selects/#native-select.

### Accessibility

- [x] I've tested using voice over

### Screenshots

Screen Shot 2019-07-05 at 10 06 54 AM

Screen Shot 2019-07-05 at 10 28 47 AM

* View the devdocs and test the `SimpleSelectControl` component.
* Optional: View the controls on the business details page `/wp-admin/admin.php?page=wc-admin&step=business-details`.

### Changelog Note:

Added a `SimpleSelectControl` component.

### DIFF
--- a/packages/components/src/simple-select-control/index.js
+++ b/packages/components/src/simple-select-control/index.js
@@ -106,6 +106,7 @@ class SimpleSelectControl extends Component {
 						{ map( options, ( option ) => {
 							const optionValue = option.value;
 							const optionLabel = option.label;
+							const optionDisabled = option.disabled || false;
 							const isSelected = ( currentValue === optionValue );
 							return (
 								<Button
@@ -117,6 +118,7 @@ class SimpleSelectControl extends Component {
 									className={ classNames( {
 										'is-selected': isSelected,
 									} ) }
+									disabled={ optionDisabled }
 									role="menuitemradio"
 									aria-checked={ isSelected }
 								>
@@ -155,12 +157,20 @@ SimpleSelectControl.propTypes = {
 			 * Label for this option.
 			 */
 			label: PropTypes.string,
+			/**
+			 * Disable the option.
+			 */
+			disabled: PropTypes.bool,
 		} )
 	),
 	/**
 	 * A function that receives the value of the new option that is being selected as input.
 	 */
 	onChange: PropTypes.func,
+	/**
+	 * The currently value of the select element.
+	 */
+	value: PropTypes.string,
 };
 
 export default withFocusOutside( SimpleSelectControl );
--- a/packages/components/src/simple-select-control/index.js
+++ b/packages/components/src/simple-select-control/index.js
@@ -1,0 +1,166 @@
+/**
+ * External dependencies
+ */
+import { __, sprintf } from '@wordpress/i18n';
+import { Dropdown, Button, NavigableMenu, withFocusOutside } from '@wordpress/components';
+import { Component } from '@wordpress/element';
+import { map, find } from 'lodash';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+
+/**
+ * A component for displaying a material styled 'simple' select control.
+ */
+class SimpleSelectControl extends Component {
+	constructor( props ) {
+		super( props );
+		this.state = {
+			currentValue: props.value,
+			isFocused: false,
+		};
+	}
+
+	componentDidUpdate( prevProps ) {
+		if (
+			this.props.value !== prevProps.value &&
+			this.state.currentValue !== this.props.value
+		) {
+			/* eslint-disable react/no-did-update-set-state */
+			this.setState( {
+				currentValue: this.props.value,
+			} );
+			/* eslint-enable react/no-did-update-set-state */
+		}
+	}
+
+	handleFocusOutside() {
+		this.setState( { isFocused: false } );
+	}
+
+	handleOnClick( onToggle ) {
+		this.setState( { isFocused: true } );
+		if ( 'function' === typeof onToggle ) {
+			onToggle();
+		}
+		const { onClick } = this.props;
+		if ( 'function' === typeof onClick ) {
+			onClick();
+		}
+	}
+
+	handleOnFocus() {
+		this.setState( { isFocused: true } );
+		const { onFocus } = this.props;
+		if ( 'function' === typeof onFocus ) {
+			onFocus();
+		}
+	}
+
+	onChange( value ) {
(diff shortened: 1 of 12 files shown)


packages/components/src/spinner/example.md
```diff
@@ -1,5 +1,5 @@
 ```jsx
-import { MySpinner } from '@woocommerce/components';
+import { Spinner } from '@woocommerce/components';
```
joshuatf: Good catch!

packages/components/src/simple-select-control/index.js
```diff
@@ -0,0 +1,166 @@
...
+							);
+						} ) }
+					</NavigableMenu>
+				) }
+			/>
+		);
+	}
+}
+
+SimpleSelectControl.propTypes = {
```
joshuatf: Should we also add a `value` propType here?

packages/components/src/simple-select-control/index.js
```diff
@@ -0,0 +1,166 @@
...
+					</Button>
+				) }
+				renderContent={ ( { onClose } ) => (
+					<NavigableMenu>
+						{ map( options, ( option ) => {
+							const optionValue = option.value;
+							const optionLabel = option.label;
+							const isSelected = ( currentValue === optionValue );
+							return (
+								<Button
```
joshuatf: Optional, but might be useful to have a `disabled` property for options.

jameskoster:
> Another small item is the opened options menu also doesn't seem to take up the full width of the dropdown container like it seems to in material-ui. Not sure if that's intentional or not /cc @jameskoster

They should match the Material UI implementation :)

timmyc:
I shall adopt this PR since Justin is out.

timmyc:
Fixed up the missing props and added a new prop for `disabled` in ef277df 

> Left a couple prop comments and one little other oddity (I think also existed on the previous component) is that the only area that's clickable is highlighted in yellow here:

Expanded the click target vertically in 4b4d1eb 

> Another small item is the opened options menu also doesn't seem to take up the full width of the dropdown container like it seems to in material-ui. Not sure if that's intentional or not

Noodled around in here a bit too, at the root of it, this is using WP Core `<Popover />` which is interesting :) - I'm not seeing a quick easy way to make this full width of the "parent" container - I'm honestly thinking if this is really desired we should add this support into the WP Core component instead of layering in a ton of css hacks here. Just my .02 though.

packages/components/src/simple-select-control/style.scss
```diff
@@ -0,0 +1,122 @@
...
+			}
+		}
+	}
+
+	&.is-empty {
+		.woocommerce-simple-select-control__selector {
+			&.components-button {
+				background: #fff;
+				font-size: 16px;
+				line-height: 54px;
```
joshuatf: Nice!  Looks good, but tiny issue when an option is selected:

Screen Shot 2019-07-10 at 10 56 27 AM

Not sure if the line-height fix will work for this, but we can also use flexbox as an alternative:

```scss
    height: 100%;
    display: flex;
    justify-content: center;
    flex-direction: column;
```

packages/components/src/simple-select-control/style.scss
```diff
@@ -0,0 +1,122 @@
...
+
+			&.is-selected {
+				background: $muriel-gray-100;
+			}
+		}
+	}
+
+	&.is-empty {
+		.woocommerce-simple-select-control__selector {
+			&.components-button {
```
joshuatf: One optional nitpick if we want to the left and right sides to also be clickable:

```suggestion
			&.components-button {
				width: 100%;
				left: 0;
				padding-left: $gap;
				padding-right: $gap;
```

But we may have to adjust the pseudo after arrow with `right: $gap;` if we do this.

timmyc:
@joshuatf would you mind giving this one more spin? I fear some of the css changes might have caused regressions elsewhere but I just don't know the designs well enough to see that. Rebased also.

packages/components/src/simple-select-control/style.scss
```diff
@@ -0,0 +1,123 @@
...
+	.woocommerce-simple-select-control__selector {
+		&.components-button {
+			border: 0;
+			background: transparent;
+			box-shadow: none;
+			justify-content: unset;
+			margin-left: 0;
+			position: absolute;
+			top: 0;
+			left: 8px;
```
joshuatf:
```suggestion
```

packages/components/src/simple-select-control/style.scss
```diff
@@ -0,0 +1,123 @@
...
+			width: calc(100% - 24px);
```
joshuatf:
```suggestion
			height: 100%;
			width: 100%;
			left: 0;
			padding-left: $gap;
			padding-right: $gap-largest;
```

packages/components/src/simple-select-control/style.scss
```diff
@@ -0,0 +1,123 @@
...
+				line-height: 56px;
+				font-family: dashicons, sans-serif;
+				font-size: 20px;
+				content: '\f140';
+				z-index: 101;
+				height: 24px;
+				width: 24px;
+				margin-top: 0;
+				top: 0;
+				right: 0;
```
joshuatf:
```suggestion
				right: $gap;
```

packages/components/src/simple-select-control/style.scss
```diff
@@ -0,0 +1,123 @@
...
+			&.is-selected {
+				background: $muriel-gray-100;
+			}
+		}
+	}
+
+	&.is-empty {
+		.woocommerce-simple-select-control__selector {
+			&.components-button {
+				background: #fff;
```
joshuatf: Need to remove this to prevent the border radius corners from being cut into.

```suggestion
```

PR #2612, opened 2019-07-08
exclude old WC auto-draft orders from import

### LINKED ISSUE #2427: Order import count discrepancies
**Describe the bug**
On historical import, I'm seeing 284 orders to be imported.  After importing, I see 275 that were imported.  In my WooCommerce Orders page, I see 215.

At first glance, this may be due to the order status and/or refunds which are also stored in the order stats table.  But the estimated number should probably match up or come close to the actual number imported.

**To Reproduce**

1. Delete all import data.
2. Import orders.
3. Note the amount to be imported.
4. Note the actual number imported.
5. Note the number of orders at `wp-admin/edit.php?post_type=shop_order`

**Expected behavior**
Expected to see an accurate number of orders to be imported.

**Screenshots**
Screen Shot 2019-06-14 at 4 53 49 PM
Screen Shot 2019-06-14 at 5 07 25 PM
Screen Shot 2019-06-14 at 5 07 13 PM

rrennick:
Fixes #2427

WordPress automatically creates a post of post type X when loading the post edit screen. By defualt the auto created post has a post status of `auto-draft`. WP cleans the database daily of posts with the `auto-draft` status. WooCommerce used to (until 3.6.0) change the status of these posts to `wc-auto-draft` (https://github.com/woocommerce/woocommerce/commit/ae17d6f3bdbe4f55b5d9f569dc9624bcca588618). As a result, these posts were not automatically cleaned up by WP.

This PR excludes the old WC auto draft posts from the import process.

- Go to Analytics -> Settings
- Uncheck Skip previous
- Compare order count to (WC Orders screen) All orders + Refunded order counts

### Changelog Note:

Fix: exclude old WC auto-draft orders from import

### DIFF
--- a/client/analytics/settings/historical-data/layout.js
+++ b/client/analytics/settings/historical-data/layout.js
@@ -86,7 +86,7 @@ class HistoricalDataLayout extends Component {
 										total={ customersTotal }
 									/>
 									<HistoricalDataProgress
-										label={ __( 'Orders', 'woocommerce-admin' ) }
+										label={ __( 'Orders and Refunds', 'woocommerce-admin' ) }
 										progress={ ordersProgress }
 										total={ ordersTotal }
 									/>
--- a/includes/class-wc-admin-reports-sync.php
+++ b/includes/class-wc-admin-reports-sync.php
@@ -365,7 +365,7 @@ class WC_Admin_Reports_Sync {
 		$count = $wpdb->get_var(
 			"SELECT COUNT(*) FROM {$wpdb->posts}
 			WHERE post_type IN ( 'shop_order', 'shop_order_refund' )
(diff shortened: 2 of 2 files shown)


rrennick:
@joshuatf Can you check this PR and see if it fixes your discrepancies?

rrennick:
> I'm guessing the majority of my difference (if not all) is coming from refunds.

My total on the import screen was total orders + total refunds. Those are the post types included in the count query. It's possible that you have refund orders in your DB where the original order was deleted or similar.

> Or maybe we should update the import text to let customers know both orders and refunds are being imported?

I think this is the better option. If the number does not add up exactly it's an indicator that more than one type is being imported and minimize the expectation that it should match up.

rrennick:
@joshuatf I updated the label.

PR #2617, opened 2019-07-09
Update onboarding form validation

### LINKED ISSUE #2584: Conditional continue button reveal
Re how we only reveal the continue button when all inputs have a value...

This concept worked better in prototypes, IRL it's kind of confusing because you don't really know which inputs to fill in before it's revealed.

Let's just keep it permanently visible and use input error states to inform users to fill in required fields:

error

joshuatf:
Fixes #2584

Updates the onboarding form validation to show errors and always show "Continue" button.

@jameskoster I used a method that allows us to easily add custom error strings to each field, so if you have any you want to change, feel free to let me know.  I only used generic messages for the business details step since I was a bit unsure on the language to use there.

### Screenshots
Screen Shot 2019-07-09 at 10 31 33 PM

1. Go through the onboarding process.
2. Make sure the "Continue" button is always shown for each step.
3. Make sure validation works for each step and displays errors appropriately.

### DIFF
--- a/client/dashboard/profile-wizard/steps/industry.js
+++ b/client/dashboard/profile-wizard/steps/industry.js
@@ -98,7 +98,7 @@ class Industry extends Component {
 								/>
 							);
 						} ) }
-						{ error && <span class="woocommerce-profile-wizard__error">{ error }</span> }
+						{ error && <span className="woocommerce-profile-wizard__error">{ error }</span> }
 					</div>
 
 					<Button isPrimary onClick={ this.onContinue }>
--- a/client/dashboard/profile-wizard/steps/industry.js
+++ b/client/dashboard/profile-wizard/steps/industry.js
@@ -20,6 +20,7 @@ class Industry extends Component {
 	constructor() {
 		super();
 		this.state = {
+			error: null,
 			selected: [],
 		};
 		this.onContinue = this.onContinue.bind( this );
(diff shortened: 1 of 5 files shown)


client/dashboard/profile-wizard/steps/industry.js
```diff
@@ -80,13 +96,12 @@ class Industry extends Component {
 								/>
 							);
 						} ) }
+						{ error && <span class="woocommerce-profile-wizard__error">{ error }</span> }
```
timmyc: This is throwing a warning in the console for me `s/class/classname`

image

client/dashboard/profile-wizard/steps/product-types.js
```diff
@@ -111,17 +127,16 @@ class ProductTypes extends Component {
 								/>
 							);
 						} ) }
+						{ error && <span class="woocommerce-profile-wizard__error">{ error }</span> }
```
timmyc: Another spot for the s/class/className

client/dashboard/profile-wizard/steps/industry.js
joshuatf: Thanks!  Looks like my linter is on the fritz again.

joshuatf:
>If I skip the required fields, and subsequently select the first ( default ) value, it appears the form field still shows as invalid/error - yet I can proceed to the next step. I would expect the error message to be cleared out once I select a value.

Good catch, Timmy!  We set the default values on focus since the current `SelectControl` doesn't allow a disabled role and would prevent `onChange` from firing when selecting the first item.  Added the validation callback to that function so those errors get removed when the first item is selected.

PR #2625, opened 2019-07-10
Dependencies: use WP installation version of wp.components

psealock:
Use `window.wp.components` instead of bundling `@wordpress/components`.

Our usage of the lastest npm version of components is problematic on two fronts:

1. Increased bundle size
2. Dependencies of `@wordpress/components` may not be available because they aren’t included in WordPress yet.

1. `npm start` and see the app fail.
2. Update and activate Gutenberg plugin.
3. See the app load correctly.

Of course we should wait until we get the `<SnackbarList />` issue resolved first so development can continue.


psealock:
I’m going to close this because I’m not sure we want to give up on using the latest Gutenberg components

PR #2628, opened 2019-07-10
Add WC support check for child themes

### LINKED ISSUE #2590: Theme step - WooCommerce support indicator
This could be something to do with the theme but I wasn't able to determine a cause.

I'm currently seeing a notice that tells me one of my Storefront child themes doesn't support WC:

ups

All the other Storefront child themes that I have installed seem to work fine though.

joshuatf:
Fixes #2590

Checks if the child or parent theme have declared WC support.

### Screenshots
Screen Shot 2019-07-10 at 5 08 26 PM

1.  Install any theme that does not declare WC support but has a parent theme that does.
2. Delete the themes transient - `wc_onboarding_themes`
3. Visit `wp-admin/admin.php?page=wc-admin&step=theme` and make sure the child theme does not have the red "WC not-supported icon."

### DIFF
--- a/includes/features/onboarding/class-wc-admin-onboarding.php
+++ b/includes/features/onboarding/class-wc-admin-onboarding.php
@@ -215,14 +215,22 @@ class WC_Admin_Onboarding {
 	 * @return bool
 	 */
 	public static function has_woocommerce_support( $theme ) {
-		$directory = new RecursiveDirectoryIterator( $theme->theme_root . '/' . $theme->stylesheet );
-		$iterator  = new RecursiveIteratorIterator( $directory );
-		$files     = new RegexIterator( $iterator, '/^.+\.php$/i', RecursiveRegexIterator::GET_MATCH );
+		$themes = array( $theme );
+		if ( $theme->get( 'Template' ) ) {
+			$parent_theme = wp_get_theme( $theme->get( 'Template' ) );
+			$themes[]     = $parent_theme;
+		}
 
-		foreach ( $files as $file ) {
-			$content = file_get_contents( $file[0] );
-			if ( preg_match( '/add_theme_support\(([^(]*)(\'|\")woocommerce(\'|\")([^(]*)/si', $content, $matches ) ) {
-				return true;
+		foreach ( $themes as $theme ) {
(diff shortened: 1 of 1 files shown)


includes/features/onboarding/class-wc-admin-onboarding.php
```diff
@@ -215,14 +212,22 @@ public static function add_uploaded_theme_data( $response ) {
...
 	public static function has_woocommerce_support( $theme ) {
-		$directory = new RecursiveDirectoryIterator( $theme->theme_root . '/' . $theme->stylesheet );
-		$iterator  = new RecursiveIteratorIterator( $directory );
-		$files     = new RegexIterator( $iterator, '/^.+\.php$/i', RecursiveRegexIterator::GET_MATCH );
-
-		foreach ( $files as $file ) {
-			$content = file_get_contents( $file[0] );
-			if ( preg_match( '/add_theme_support\(([^(]*)(\'|\")woocommerce(\'|\")([^(]*)/si', $content, $matches ) ) {
-				return true;
+		$themes = array( $theme );
```
timmyc: Wow I'm really surprised there isn't a WP core util for checking theme support on non-active themes 🤷‍♂ - but in my googlin' and checking https://core.trac.wordpress.org/browser/tags/5.2.1/src/wp-includes/theme.php I don't see anything outside of `current_theme_supports()`

joshuatf: Ya, I thought for sure there would be one too.  I guess since support is declared in PHP instead of template meta tags, it would require loading an entire theme which may cause side effects.

joshuatf:
>Wow this is super slick! My first theme step experience ;) Solid work, and it looks great.

Thanks, @timmyc! 😄 

>I was able to upload bistro, even thought it was already installed. This led to a duplicate key warning in react:

You should get a warning if the theme already exists to prevent this, but I'm wondering if `$theme->stylesheet` could be the same for 2 different themes and doesn't correlate to theme folder name.  Were your folder names different and theme slug identical?

joshuatf:
Here's the error that's normally displayed:
Screen Shot 2019-07-11 at 4 31 41 PM

PR #2636, opened 2019-07-10
update order activity panel to check for all time orders

### LINKED ISSUE #2309: Orders panel shows no orders message when orders exist
**Describe the bug**
A message is shown in the Orders activity panel that the store has not received any orders when it has.

I believe this is due to the before/after params defaulting to a 1 week period so `totalNonActionableOrders` results in `0` if no recent orders have been placed.

https://github.com/woocommerce/woocommerce-admin/blob/4d1e76117ecff27715ea6818008f7919caafe498/client/header/activity-panel/panels/orders.js#L310

Also related - https://github.com/woocommerce/woocommerce-admin/issues/1103

**To Reproduce**

1. Make sure you have orders, but none within the past week.
2. Open the orders activity panel.
3.  Note the message.

**Expected behavior**
Expected to see the message relevant to having previous orders but no actionable statuses:
`Good job, you've fulfilled all of your new orders!`

Or alternatively, a message that mentions there have been no orders within the past week.

**Screenshots**
Screen Shot 2019-05-27 at 11 29 45 AM

rrennick:
Fixes #2309

The order panel did not have a `hasNonActionableOrders` prop so the no orders empty card was always being displayed when there were no current actionable orders. This PR also 

- updates the all orders query to find if there are any orders vs orders in the last week
- updates the icon title to specify `no new orders`

For the reviewer, should the unread indicator also check for any orders?

- Have at least 1 order with processing or on-hold status in the last 7 days
- Check that the order shows in he orders panel
- Cancel that order
- Reload the order panel to check that the no new orders card is shown

### Changelog Note:

Fix: order activity panel check for existing orders.

### DIFF
--- a/client/header/activity-panel/panels/orders.js
+++ b/client/header/activity-panel/panels/orders.js
@@ -28,12 +28,12 @@ import withSelect from 'wc-api/with-select';
 
 class OrdersPanel extends Component {
 	renderEmptyCard() {
-		const { hasNonActionableOrders } = this.props;
-		if ( hasNonActionableOrders ) {
+		const { hasActionableOrders } = this.props;
+		if ( hasActionableOrders ) {
 			return (
 				<ActivityCard
 					className="woocommerce-empty-activity-card"
-					title={ __( 'You have no orders to fulfill', 'woocommerce-admin' ) }
+					title={ __( 'You have no new orders to fulfill', 'woocommerce-admin' ) }
 					icon={ <Gridicon icon="checkmark" size={ 48 } /> }
 				>
 					{ __( "Good job, you've fulfilled all of your new orders!", 'woocommerce-admin' ) }
@@ -283,6 +283,7 @@ export default compose(
 		const allOrdersQuery = {
 			page: 1,
 			per_page: 0,
+			after: '1970-01-01T00:00:01',
 		};
 
 		const totalNonActionableOrders = getReportItems( 'orders', allOrdersQuery ).totalResults;
--- a/client/header/activity-panel/unread-indicators.js
+++ b/client/header/activity-panel/unread-indicators.js
@@ -44,6 +44,7 @@ export function getUnreadOrders( select ) {
 		page: 1,
 		per_page: 0,
 		status_is: orderStatuses,
+		after: '1970-01-01T00:00:01',
 	};
 
 	const totalOrders = getReportItems( 'orders', ordersQuery ).totalResults;


timmyc:
/cc @jameskoster for design thoughts on this, but I concur with Josh that "My vote is to show any actionable orders that were ever placed "

Really to me this feature should mimic how we show the "processing order" number in the wp-admin sidebar, but is more configurable as users select which order statuses should impact the number. Case in point on my test site, there are four recent orders that are processing, but one older one ( from december I think ) that still is bumping the number to five:

orders-processing

So I think we should emulate that same logic with the indicator dot.

jameskoster:
> My vote is to show any actionable orders that were ever placed

I agree. Assuming this is a simple enough change I think it makes sense.

But I would be hesitant about spending too much time on the Orders, Reviews, and Stock sections of the Activity Panel right now, though. As we seek to implement the wc-admin navigation (p6riRB-4Bp-p2) these sections will be much more discoverable via that, than they are currently in the mess that is wp-admin. Consequently I'm not sure we'll _need_ the spotlight on them that the Activity Panel currently affords.

rrennick:
Holding this one off until #2659 is merged.

client/header/activity-panel/panels/orders.js
```diff
@@ -28,12 +28,12 @@ import withSelect from 'wc-api/with-select';
 
 class OrdersPanel extends Component {
 	renderEmptyCard() {
-		const { hasNonActionableOrders } = this.props;
-		if ( hasNonActionableOrders ) {
+		const { hasActionableOrders } = this.props;
+		if ( hasActionableOrders ) {
```
jeffstieler: Aren't actionable orders the ones that need fulfillment?

`hasNonActionableOrders` is essentially "the store has received orders" - which is why its used to display "no **new** orders" vs. "you have no orders".

rrennick: `hasNonActionableOrders` isn't a prop so it was always defaulting to false. The PR now checks to see if there are any actionable orders in history for `hasActionableOrders` where the ones being shown in the panel are the new actionable orders.

jeffstieler: > `hasNonActionableOrders` isn't a prop so it was always defaulting to false.

I'm not sure I follow. `hasNonActionableOrders` is a computed prop here: https://github.com/woocommerce/woocommerce-admin/blob/8384d988ceed5868ee66b448edf91c0567ed7f34/client/header/activity-panel/panels/orders.js#L293. I do see the intended behavior of it in my testing.

rrennick: Possibly I was looking for the prop in the wrong place

Screen Shot 2019-07-17 at 3 50 21 PM

rrennick: Found it.
Screen Shot 2019-07-17 at 3 57 43 PM

rrennick:
Closing as #2659 fixed the issue.

PR #2637, opened 2019-07-11
Fix "page" parameter collision between routing and table pagination.

jeffstieler:
https://github.com/woocommerce/woocommerce-admin/pull/2444 introduced a bug with report table pagination - a collision on the `page` query parameter (since the report query is no longer part of a fragment).

This PR renames the report query parameter from `page` to `paged` (a la WordPress) to avoid the collision.

- Go to a report with multiple pages
- Verify that click the next/prev arrows work
- Verify that typing in a page number works
- Verify that when on a page > 1 that re-sorting table goes back to page 1
- Verify that a page refresh properly loads the report page you were on ( > 1 )

### Changelog Note:

None - fixes an unreleased bug.

### DIFF
--- a/client/layout/controller.js
+++ b/client/layout/controller.js
@@ -86,8 +86,8 @@ export class Controller extends Component {
 
 	componentDidUpdate( prevProps ) {
 		const prevQuery = this.getQuery( prevProps.location.search );
-		const prevBaseQuery = this.getBaseQuery( prevProps.location.search );
-		const baseQuery = this.getBaseQuery( this.props.location.search );
+		const prevBaseQuery = omit( this.getQuery( prevProps.location.search ), 'paged' );
+		const baseQuery = omit( this.getQuery( this.props.location.search ), 'paged' );
 
 		if ( prevQuery.paged > 1 && ! isEqual( prevBaseQuery, baseQuery ) ) {
 			getHistory().replace( getNewPath( { paged: 1 } ) );
@@ -107,16 +107,10 @@ export class Controller extends Component {
 		return parse( search );
 	}
 
-	getBaseQuery( searchString ) {
-		const query = this.getQuery( searchString );
-		delete query.paged;
-		return query;
-	}
-
 	render() {
 		const { page, match, location } = this.props;
 		const { url, params } = match;
-		const query = omit( this.getQuery( location.search ), 'page' );
+		const query = this.getQuery( location.search );
 
 		window.wpNavMenuUrlUpdate( page, query );
 		window.wpNavMenuClassChange( page, url );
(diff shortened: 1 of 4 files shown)


jeffstieler:
This was an unexpected thing to encounter when fixing: https://github.com/woocommerce/woocommerce-admin/commit/f4e2013673dbefce2743f9826eebd4a066a83e0d#r34271330

But I guess it's part of the bug?

client/layout/controller.js
```diff
@@ -109,14 +109,14 @@ export class Controller extends Component {
...
-		delete query.page;
+		delete query.paged;
 		return query;
 	}
 
 	render() {
 		const { page, match, location } = this.props;
 		const { url, params } = match;
-		const query = this.getBaseQuery( location.search );
+		const query = this.getQuery( location.search );
```
psealock: The point of `getBaseQuery` is to remove the unused parameters. In the case of the updated url scheme, we want to pass down everything except for `page: "wc-admin"`, because its irrelevant to the app. `paged` however is important and should be passed down.

So `getBaseQuery` should remove `page` parameter, and this line should go back to using `getBaseQuery`.

jeffstieler: Ah I see. I think there will need to be changes to `componentDidUpdate` as well, since it compares base queries to determine if it needs to reset the page number. If `paged` is part of the base query, I believe it won't allow getting to page 2+.

jeffstieler: Fixed in 4ff4b32.

psealock: >If paged is part of the base query, I believe it won't allow getting to page 2+.

Indeed thats the case. I did some testing by removing `delete query.paged;` and it could not get to page 3. 

`getBaseQuery` seems like a utility function to be used in a variety of places. Can we rename it to `getPagedLessQuery`? Or something like that?

jeffstieler: I only see `getBaseQuery` used within the controller file - where else are you seeing it?

psealock: Nowhere else, I'm just hoping to prevent future confusion

jeffstieler: What do you think about just omitting the `page` parameter in `getQuery()` and removing `getBaseQuery()` entirely? Seems like we never want to use the `page` value since it's going to be the WordPress-registered page anyhow.

psealock: >What do you think about just omitting the page parameter in getQuery()

That sounds ideal.

>and removing getBaseQuery() entirely?

Also idea, just remember to remove `paged` for the comparison made in `componentDidUpdate`

jeffstieler: Done in 697b8a7.

psealock: >What do you think about just omitting the page parameter in getQuery()

Looking great, did this part get omitted?

jeffstieler: Yeah - I didn't initially realize it was unnecessary. It was to catch the now defunct `page` fragment param for pagination.

jeffstieler:
Got a :shipit: in Slack.

Issue #2203, opened 2019-05-10
Bump PHP and WordPress requirements for plugin


joshuatf:
https://github.com/woocommerce/woocommerce-admin/pull/2083 introduced traits and bumped the PHP requirements to PHP 5.4+.

There was some discussion in the above PR about bumping to WP 5.2 which would move PHP requirements to 5.6.20.  So we have 3 options:

* Leave PHP requirements as 5.4+ to support WP 5.0 and 5.1.
* Bump WP version 5.2. and PHP to 5.6.20.
* Remove the traits added in #2083, maintain methods separately (some code duplication) and drop PHP version requirements back down.

peterfabian:
As we're going to merge this code into WC core for 3.7, I don't think we can require WP 5.2 (which is 5 days old at the moment). Also, WP 5.0's minimum requirements is PHP 5.2.4 (unless I'm mistaken). Which leaves us, unfortunately, with supporting PHP 5.2.4 at the moment, I think. Please let me know your thoughts.

rrennick:
An alternative for core would be to have WCA conditional on PHP 5.6 or higher. About 19% of WP installs are < 5.6 per https://wordpress.org/about/stats/. I'm not sure if there are better stats for WooCommerce.

peterfabian:
Our tracker data is quite a bit more optimistic, it shows only 4% of users with PHP v < 5.6, so we could make it load only for PHP 5.6 without losing many users.
And based on discussions with @timmyc it seems like WC admin will most likely depend on WP 5 as well, which is a stronger requirement.

timmyc:
>  it seems like WC admin will most likely depend on WP 5 as well, which is a stronger requirement, so requiring PHP 5.6 should not be a big deal.

Correct if 5.0 is not present, no wc-admin features will be functional in woo core - just like it is currently in the feature plugin.

peterfabian:
We talked about it in the team meeting today. 
Preliminary conclusion: 
- We will announce bumping PHP requirement to 5.6 with a post on dev blog soon
- This requirement will officially become effective at the time of WC 3.7 release, but we prefer to only load the code in case correct PHP version is available (so it would not break existing installations). 

Therefore, according to my understanding, in this PR we need to add code to conditionally load the classes only if WP > 5 && PHP > 5.4.

rrennick:
> As we're going to merge this code into WC core for 3.7

[...]

PR #2643, opened 2019-07-12
Revert "Revert "Update to new Snackbar component and notice store""

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/2203

Reverts woocommerce/woocommerce-admin#2634

This PR reverts the revert to re-implement Snackbar and transient notices changes originally made in https://github.com/woocommerce/woocommerce-admin/pull/2596.

The revert was done to due to dependencies on React version 16.8. This may be a simple issue of bumping WP dependency to `5.2.0`. Alternatively, we may decide to close this PR and stick with React version 16.6.3 found in WP `5.1.1`.

More information in p90Yrv-187


psealock:
Changing this to Blocked until we can resolve the plan around the dependency issue.

psealock:
@joshuatf We're going ahead and bumping our WP dependency to 5.2, which means we'll go ahead with these changes, see p7bje6-1pI. 

Can you take this one over and rebase to update against master?

joshuatf:
This has been rebased and WP version bumped to 5.2.  Ready for a review.

nerrad:
Looks like there's some plugin header fields that could be added here as well in `woocommerce-admin.php`, namely `Requires at least:` and `Requires PHP:` (see [here](https://developer.wordpress.org/plugins/plugin-basics/header-requirements/#header-fields))

psealock:
This will fix https://github.com/woocommerce/woocommerce-admin/issues/2203

joshuatf:
Thanks for catching that @nerrad!  Updated in https://github.com/woocommerce/woocommerce-admin/pull/2643/commits/04caa182660cf594096cd25c39bbebb0cdf936be

@psealock I also noticed that we list the required resource not only in the main plugin class file, but also in `readme.txt`.  Wondering if this header information is read somewhere else or if we need both.

psealock:
>Wondering if this header information is read somewhere else or if we need both.

Looks like we should have both, https://developer.wordpress.org/plugins/wordpress-org/how-your-readme-txt-works/

PR #1979, opened 2019-04-02
Allow negative values in charts

Aljullu:
Fixes #1959.

Adds the ability to display negative numbers to `<Chart>` component.

### Screenshots
_All negative:_
image
_Positive and negative values:_
image
_Bar charts:_
image

### Detailed test instructions:
Currently all our charts only display positive values, so you need to overwrite `chartData` before rendering them.

In https://github.com/woocommerce/woocommerce-admin/blob/master/client/analytics/components/report-chart/index.js#L125, add these lines:
```ES6
for ( let i = 0; i < chartData.length; i++ ) {
	chartData[ i ].primary.value = ( Math.random() - 0.5 ) * 100;
	chartData[ i ].secondary.value = ( Math.random() - 0.5 ) * 100;
}
```
- Go to any report and verify charts look good.
- Switch between line and bar chart.
- Try modifying the code snippet above to force other data combinations:

**Only negative values:**
```ES6
for ( let i = 0; i < chartData.length; i++ ) {
	chartData[ i ].primary.value = Math.random() * -100;
	chartData[ i ].secondary.value = Math.random() * -100;
}
```
  **All 0's:**
```ES6
for ( let i = 0; i < chartData.length; i++ ) {
	chartData[ i ].primary.value = 0;
	chartData[ i ].secondary.value = 0;
}
```
  **Beautiful data going up (used for the screenshots above) :chart_with_upwards_trend: :**
```ES6
for ( let i = 0; i < chartData.length; i++ ) {
	chartData[ i ].primary.value = ( Math.random() - 0.7 + i / 30 ) * 100;
	chartData[ i ].secondary.value = ( Math.random() - 0.7 + i / 30 ) * 100;
}
```

PR #2646, opened 2019-07-12
Chart: Remove z-index from empty message

timmyc:
Sometimes a fix is easier than creating an issue :) - I noticed this while testing recently, and @beaulebens reported it in slack this week too. The empty message shown on charts had a z-index set, which caused it to float above things like the Date Picker, and section toggles in the dashboard. 

__Before__
image

The fix here just removes the z-index value which was added in #1979 ( /cc @Aljullu in case there is a case that still warrants the need for the z-index on the empty message ).

__After__
Dashboard ‹ bend outdoors — WooCommerce 2019-07-12 08-27-15

- Select a date range where no data will be shown in charts on the dashboard
- Shrink your viewport so the popover shown when clicking the kebab menu on the chart section will overlap a chart with an empty message
- Verify the empty message is not shown on top of the kebab menu

### Changelog Note:

Fix: Z-index issue in empty message on chart

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -3,6 +3,7 @@
 - AdvancedFilters component: fire `onAdvancedFilterAction` for match changes.
 - TableCard component: add `onSearch` an `onSort` function props.
 - Add new component `<List />` for displaying interactive list items.
+- Fix z-index issue in `<Chart />` empty message.
 
 # 3.1.0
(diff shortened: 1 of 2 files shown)


timmyc:
Thanks for the context @Aljullu - wonder if a solution there might just be to hide the y-axis horizontal rule markers when the chart is empty too - since there is no data being plotted...

PR #2654, opened 2019-07-15
Onboarding: Add the task list dashboard

### LINKED ISSUE #1896: Onboarding - Task List: Dashboard Page
After completing the profiling step folks will be directed to the WooCommerce Dashboard. This is where the main set up experience will live. On active stores this screen will display analytics, but for new stores there’s no value there. We’ll reveal the analytics overview once the user has made the conscious decision that they are finished with setting up their store.

This task involves adding logic to the `Dashboard` code, and a new feature flag for loading the onboarding section. This may involve some slight refactoring to the dashboard code so we can launch sections separately.

Screen Shot 2019-03-22 at 2 49 04 PM

[Designs](https://wp.me/p6riRB-44u)

joshuatf:
Fixes #1896 

Adds the base for the task list dashboard.

A couple design question /cc @jameskoster 

* I attempted to replace with Gridicons since we already have those imported.  Not all of the icons match up well, so if you want to change any or think we should just individually import the ones in Figma, we can do so.
* Should the header bar be fixed to the top of the screen?  Currently it's fixed on desktop and sits statically on mobile.

### Screenshots
Screen Shot 2019-07-15 at 6 47 21 PM

1. Make sure you have the onboarding feature flag enabled and have completed or skipped the profiler.
2. Change `requiredTasksComplete` to `false` in dashboard/index.js.
3. Visit the dashboard.
4. Check that the dashboard and task list match those in Figma.


client/dashboard/profile-wizard/index.js
```diff
@@ -74,11 +74,13 @@ class ProfileWizard extends Component {
 	componentDidMount() {
 		document.documentElement.classList.remove( 'wp-toolbar' );
 		document.body.classList.add( 'woocommerce-profile-wizard__body' );
+		document.body.classList.add( 'woocommerce-dashboard__body' );
```
timmyc: Minor but I suppose we could combine both classList.add and classList.remove calls.

client/dashboard/task-list/index.js
```diff
@@ -3,31 +3,108 @@
...
+			<Fragment>
+				<div className="woocommerce-task-dashboard__sidebar" />
+				<div className="woocommerce-task-dashboard__header">
+					<div className="woocommerce-task-dashboard__header-site">{ siteUrl }</div>
+					<span className="woocommerce-task-dashboard__header-title">
+						{ __( 'Dashboard', 'woocommerce-admin' ) }
+					</span>
+					<div className="woocommerce-task-dashboard__header-connection">
+						<div className="woocommerce-task-dashboard__header-connection-avatar">
+							{ /* @todo This shoud display the user avatar if the store is connected. */ }
```
timmyc: missing an l in should

joshuatf: Nice spot!  This shoud be fixed now :trollface:

client/dashboard/profile-wizard/index.js
joshuatf: I think we can, but might lose support in IE11 with multiple properties for add/remove - https://caniuse.com/#feat=classlist

joshuatf:
Thanks for the review @timmyc!

>The only other item I continually notice from a ui standpoint is the obvious "flash" of classic wp-admin that is shown when loading up this view ( and the onboarding view ).

Great point.  We'll have this issue for the regular onboarding too and I think the new router will let us detect these pages so we can add classes server-side.  Will log an issue for that

@jameskoster Updated to use Material icons; can you let me know if there are any issues with using SVGs?  Also removed the app and sidebar and restored to the default view.  Did not opt to hide notices or change layout here, but let me know if you think we need to handle any of those.

Screen Shot 2019-07-16 at 12 12 46 PM

jameskoster:
Hm, shouldn't the notices be in the "Notices" section of the Activity Panel?

So... the icons... This is kind of messy but for now it doesn't look like we can include the raw icons in our package because the license isn't bona fide GPL compatible. More discussion on that here: https://a8c.slack.com/archives/C025Q5CPE/p1563188931218300

We'll have to use the [font](https://google.github.io/material-design-icons/#icon-font-for-the-web) instead.

timmyc:
> Hm, shouldn't the notices be in the "Notices" section of the Activity Panel?

That right there is a Store Alert, something we feel is high priority enough to show there :)

joshuatf:
>So... the icons... This is kind of messy but for now it doesn't look like we can include the raw icons in our package because the license isn't bona fide GPL compatible. More discussion on that here: 

Thanks @jameskoster- updated the font to use the Google hosted font, which appears to be the only option that's GPL compatible based on that discussion.  Can you give this another look when you can to confirm?

jameskoster:
Yup, the Google hosted font is fine :)

PR #2655, opened 2019-07-15
Replace 'stringifyQuery' with 'addQueryArgs'

Aljullu:
`@wordpress/url` includes a method named `addQueryArgs` which does exactly the same than our `stringifyQuery`. Given that `@wordpress/url` is already a child dependency, it makes sense using that one instead of duplicating it.

This PR does exactly that, replaces all instances of `stringifyQuery` with `addQueryArgs`.

- Basically we need to test that all API calls keep working. Testing the different reports, the autocompleters, the _Import Historical Data_...

### DIFF
--- a/client/analytics/settings/historical-data/index.js
+++ b/client/analytics/settings/historical-data/index.js
@@ -3,15 +3,11 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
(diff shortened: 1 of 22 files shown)


psealock:
cc @rrennick, this breaks the Scheduled Actions plugin.

PR #2659, opened 2019-07-16
Reduce order status change delay in Orders Panel.

### LINKED ISSUE #2592: Order Panel: Completed Orders Linger in List
**Describe the bug**
After marking an order as completed ( a non-actionable status ) the order is still shown in the activity panel with the old order status.

**To Reproduce**

1. Have an order that has an "actionable" status and shows up in the activity panel.
2. Edit the order to a non-actionable status, i.e. mark it as complete.
3. Note after the POST of the edit order form is completed, the order is still shown in the order activity panel
4. Navigate to the wc-admin dashboard, and the order will still be in the panel.

**Expected behavior**
Completed orders to no longer show in my activity panel

jeffstieler:
Fixes #2592 and #2309.

This PR seeks to reduce/eliminate the delay in order status transitions manifesting in the Orders Panel.

It does so by using the core WooCommerce Orders endpoint to determine if there are actionable orders, as well as trusting the core endpoint data over the reporting table data - both endpoints are hit and the data merged, with the core endpoint overriding the reporting endpoint. This hopefully ensures that we can catch order status (and date, etc) changes in any "direction" - processing to completed, or completed to pending, etc.

Additionally, this PR removes the default date range for the `/reports/orders` endpoint (since it's not present in the core endpoint, and we need parity). This solves #2309.

- Have an order that has an "actionable" status and shows up in the activity panel.
- Edit the order to a non-actionable status, i.e. mark it as complete.
- Note after the POST of the edit order form is completed, the order is no longer in the panel.
- Navigate to the wc-admin dashboard, verify the order is no longer in the panel.

### Changelog Note:

Fix: completed orders lingering in activity panel.


client/header/activity-panel/panels/orders.js
```diff
@@ -265,29 +274,37 @@ export default compose(
...
-		if ( hasActionableOrders ) {
+		// Query the core Orders endpoint for the most up-to-date statuses.
+		const allOrdersQuery = {
+			page: 1,
+			per_page: QUERY_DEFAULTS.pageSize,
+			status: orderStatuses,
+		};
+		const actionableOrders = Array.from( getItems( 'orders', allOrdersQuery ).values() );
+
+		if ( hasActionableOrders && actionableOrders.length ) {
```
joshuatf: I'm unable to see orders here since `hasActionableOrders` is false.  I'm thinking `hasActionableOrders` might not be needed here.  Since we've already queried for `actionableOrders` we can just check its length.

On that same note, we may want to update the unread indicator to use the same test so those aren't out of sync.

includes/api/class-wc-admin-rest-reports-orders-controller.php
```diff
@@ -385,6 +387,24 @@ public function get_collection_params() {
 			'sanitize_callback' => 'wc_string_to_bool',
 			'validate_callback' => 'rest_validate_request_arg',
 		);
+		$params['order_includes']   = array(
```
joshuatf: Nice 👍  Thanks for adding these!

client/header/activity-panel/panels/orders.js
```diff
@@ -265,29 +274,37 @@ export default compose(
 			return { orders: [], isError: true, isRequesting: false, orderStatuses };
 		}
 
-		if ( hasActionableOrders ) {
+		// Query the core Orders endpoint for the most up-to-date statuses.
+		const allOrdersQuery = {
```
joshuatf: Out of scope for this PR, but would be awesome if we could just return the IDs here - https://developer.wordpress.org/reference/classes/wp_query/#return-fields-parameter

includes/data-stores/class-wc-admin-reports-orders-data-store.php
```diff
@@ -151,8 +161,8 @@ public function get_data( $query_args ) {
 			'page'             => 1,
 			'order'            => 'DESC',
 			'orderby'          => 'date_created',
-			'before'           => WC_Admin_Reports_Interval::default_before(),
-			'after'            => WC_Admin_Reports_Interval::default_after(),
+			'before'           => '',
```
joshuatf: I like the idea of removing the default date limitations.  Just want to make sure it matches up with what is done in https://github.com/woocommerce/woocommerce-admin/pull/2636.

client/header/activity-panel/panels/orders.js
jeffstieler: Done in 7a3f4fd.

jeffstieler:
> Also I noticed that orders seem to be in chronological order (with the oldest at top). This may be expected, but quick sanity check while we're here.

Fixed in b5a47e1.

client/header/activity-panel/panels/orders.js
jeffstieler: Hmm.. I'm not seeing `hasActionableOrders` as `false`. It ultimate comes from the unread indicator function: https://github.com/woocommerce/woocommerce-admin/blob/71e7e5cbf46ee3ddac31452463ec527b12cbcf04/client/header/activity-panel/index.js#L148, https://github.com/woocommerce/woocommerce-admin/blob/71e7e5cbf46ee3ddac31452463ec527b12cbcf04/client/header/activity-panel/index.js#L283

includes/data-stores/class-wc-admin-reports-orders-data-store.php
jeffstieler: There might be an errant change in #2636, see: https://github.com/woocommerce/woocommerce-admin/pull/2636/files#r304572283.

jeffstieler:
I accidentally broke the "non actionable" order query - fixed in 9a0f74d. This is ready for another review.

includes/data-stores/class-wc-admin-reports-orders-data-store.php
rrennick: I think this PR is going to do the trick without implementing #2636.

client/header/activity-panel/panels/orders.js
```diff
@@ -266,28 +275,60 @@ export default compose(
 		}
 
 		if ( hasActionableOrders ) {
+			// Query the core Orders endpoint for the most up-to-date statuses.
+			const actionableOrdersQuery = {
+				page: 1,
+				per_page: QUERY_DEFAULTS.pageSize,
+				status: orderStatuses,
+				_fields: [ 'id', 'date_created_gmt', 'status' ],
```
joshuatf: Awesome!  Did not know the `_fields` property was an option here.

client/header/activity-panel/panels/orders.js
```diff
@@ -266,28 +275,60 @@ export default compose(
...
+				orders = reportOrders.map( order =>
+					merge( {}, order, actionableOrdersById[ order.order_id ] || {} )
+				);
+			}
 
 			return { orders, isError, isRequesting, orderStatuses };
 		}
 
+		// Get a count of all orders for messaging purposes.
+		// @todo: add a property to wcSettings for this?
```
joshuatf: Minor nitpick: todos were update to format:

`@todo Thing to do.`

jeffstieler: Amended the last commit to fix this. Thanks.

PR #2660, opened 2019-07-16
Settings: Fix notice shown after save.

### LINKED ISSUE #2541: Settings: Confirmation alert not shown upon save

timmyc:
Fixes #2541 

Currently when you save settings changes on the Analytics | Settings screen, the message informing you that your changes have been saved is not being displayed. As mentioned in the original issue, this was being caused due to the `lastReceived` timestamp on the settings resource not being update on save/mutation.

This appears to be due to the fact that `updateSettings` [calls a batch endpoint](https://github.com/woocommerce/woocommerce-admin/blob/master/client/wc-api/settings/operations.js#L43) and as such, the responses are persisted in a different part of the resource state tree.

The change proposed here is to detect if the group being selected is `wc_admin`, and if so, check to see if we have a `lastReceived` there and use it in the logic to determine if indeed `isGetSettingsRequesting`.

### Screenshots
save-success

Might be nice to test this on master first to see the bug in action...

- Go to Analytics | Settings
- Change a setting, click save
- Note the success toaster message
- Refresh the page and verify your settings indeed have saved.

### Changelog Note:

Fix: Bug that was preventing confirmation dialog from showing after saving settings.

### DIFF
--- a/client/wc-api/settings/selectors.js
+++ b/client/wc-api/settings/selectors.js
@@ -36,7 +36,23 @@ const isGetSettingsRequesting = getResource => group => {
 		return true;
 	}
 
-	return lastRequested > lastReceived;
+	// This selector is used by other "groups", so return early if not looking at wc_admin settings.
+	if ( 'wc_admin' !== group ) {
+		return lastRequested > lastReceived;
+	}
+
+	// Mutation operations for `wc_admin` settings update a different resource (batch endpoint) in fresh-data.
+	// As such we must use lastReceived stamp from that resource to properly compare lastRequested to lastReceived.
+	const { lastReceived: lastMutationReceived } = getResource(
+		getResourceName( 'settings/wc_admin', 'woocommerce_actionable_order_statuses' )
+	);
+
+	// If we don't have a lastReceived on mutations, use the standard resource times.
+	if ( isNil( lastMutationReceived ) ) {
+		return lastRequested > lastReceived;
+	}
+
+	return lastRequested > lastMutationReceived;
 };
 
 export default {
--- a/client/wc-api/settings/selectors.js
+++ b/client/wc-api/settings/selectors.js
@@ -10,6 +10,7 @@ import { isNil } from 'lodash';
  */
 import { DEFAULT_REQUIREMENT } from '../constants';
 import { getResourceName } from '../utils';
+import { analyticsSettings } from 'analytics/settings/config';
 
 const getSettings = ( getResource, requireResource ) => (
 	group,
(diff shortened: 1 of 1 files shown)


client/wc-api/settings/selectors.js
```diff
@@ -36,7 +36,23 @@ const isGetSettingsRequesting = getResource => group => {
 		return true;
 	}
 
-	return lastRequested > lastReceived;
+	// This selector is used by other "groups", so return early if not looking at wc_admin settings.
```
jeffstieler: Nice.

client/wc-api/settings/selectors.js
```diff
@@ -36,7 +36,23 @@ const isGetSettingsRequesting = getResource => group => {
...
+	if ( 'wc_admin' !== group ) {
+		return lastRequested > lastReceived;
+	}
+
+	// Mutation operations for `wc_admin` settings update a different resource (batch endpoint) in fresh-data.
+	// As such we must use lastReceived stamp from that resource to properly compare lastRequested to lastReceived.
+	const { lastReceived: lastMutationReceived } = getResource(
+		getResourceName( 'settings/wc_admin', 'woocommerce_actionable_order_statuses' )
```
jeffstieler: What do you think about using the `analyticsSettings` object to pick a resource name here since the value is filterable? See: https://github.com/woocommerce/woocommerce-admin/blob/b0c2b15a208baebda17c091ab47c874c108c42c9/client/analytics/settings/config.js#L53

timmyc: @jeffstieler so just grab the first key from `analyticsSettings` instead of having the hardcoded one?

jeffstieler: Yeah, that's what I was thinking.

timmyc:
Good idea @jeffstieler  - updated in 8db03c6

client/wc-api/settings/selectors.js
```diff
@@ -36,7 +37,26 @@ const isGetSettingsRequesting = getResource => group => {
 		return true;
```
psealock: I'm not getting the notice to appear. `lastReceived` is undefined when I stop the and the value returned is always `true` so no notice gets generated.

Can you reproduce?

timmyc: settings-save

Just tested again locally, and all is working for me still.

psealock: Ok, I must have goofed something up because its working for me now

PR #2667, opened 2019-07-18
Onboarding: Add products task

### LINKED ISSUE #2479: Onboarding - Task List: Add Products
Blocked by https://github.com/woocommerce/woocommerce-admin/issues/2471

This screen allows users to import, manually create, or migrate products.

Each of the concurrent screens should already exist, so this screen is purely navigational.  However, we'll probably need to redirect back to the task list after completion for each step.

### Add products screen
Screen Shot 2019-06-21 at 4 29 40 PM

Screen Shot 2019-06-21 at 4 30 25 PM

joshuatf:
Fixes #2479 

* Adds the products task.
* Adds task completion checks to `wcSettings`.
* Marks items complete in the task list if complete (currently only on initial page load- follow-up needed to fetch async).
* Fixes the task list item `className` prop.

### Screenshots
Screen Shot 2019-07-18 at 6 50 38 PM

1. Update `const requiredTasksComplete = false;` in dashboard/index.js.
2. Visit the dashboard.
3. Click on the "Products" task item.
4. Make sure each of the sub-task items goes to the correct page.
5. Make sure that after completing the task (i.e., adding a product) you are redirected back to the task list dashboard.
6. Make sure the task list "Products" item is marked completed if you have any products.

**Note:** The redirect back to the task list is only enabled if you had 0 products to begin with.  To test, either use a site with 0 products or comment out the following lines in `set_active_task()`:
```php
			if ( self::check_task_completion( $task ) ) {
				return;
			}
```


client/dashboard/task-list/index.js
```diff
@@ -29,30 +34,40 @@ const getTasks = () => {
...
 			),
-			before: <i class="material-icons-outlined">add_box</i>,
+			before: tasks.products ? (
+				<i class="material-icons-outlined">check_circle</i>
+			) : (
+				<i class="material-icons-outlined">add_box</i>
+			),
 			after: <i class="material-icons-outlined">chevron_right</i>,
-			onClick: noop,
+			onClick: () => updateQueryString( { task: 'products' } ),
```
timmyc: Guessing we will probably want to layer in some tracks on this click event too.

client/dashboard/task-list/tasks/products.js
```diff
@@ -0,0 +1,64 @@
...
+const subTasks = [
+	{
+		title: __( 'Add manually (recommended)', 'woocommerce-admin' ),
+		description: __(
+			'For small stores we recommend adding products manually',
+			'woocommerce-admin'
+		),
+		before: <i class="material-icons-outlined">add_box</i>,
+		after: <i class="material-icons-outlined">chevron_right</i>,
+		onClick: () =>
```
timmyc: Likewise all of the sub tasks should probably have tracks recorded.

client/dashboard/task-list/index.js
joshuatf: Good point.  I'll probably handle this in a follow-up issue if that's okay.  I'm anticipating a small refactor to the list click-handlers to DRY this up a bit, but need to see all of the tasks here to know if that will work.

Issue for tracks outlined [here](https://github.com/woocommerce/woocommerce-admin/issues/2671)

client/dashboard/task-list/index.js
```diff
@@ -29,30 +34,40 @@ const getTasks = () => {
...
 		{
+			key: 'products',
 			title: __( 'Add your first product', 'woocommerce-admin' ),
 			description: __(
 				'Add products manually, import from a sheet or migrate from another platform',
 				'wooocommerce-admin'
 			),
-			before: <i class="material-icons-outlined">add_box</i>,
+			before: tasks.products ? (
+				<i class="material-icons-outlined">check_circle</i>
```
psealock: These and others need to be `className`. I'm getting lots of these:

```
react-dom.js?ver=16.8.4:500 Warning: Invalid DOM property `class`. Did you mean `className`?
```

includes/features/onboarding/class-wc-admin-onboarding-tasks.php
```diff
@@ -0,0 +1,131 @@
...
+	 * Check for task completion of a given task.
+	 *
+	 * @param string $task Name of task.
+	 * @return bool;
+	 */
+	public static function check_task_completion( $task ) {
+		switch ( $task ) {
+			case 'products':
+				$products = wp_count_posts( 'product' );
+				return $products > 0;
```
psealock: I'mn getting this error when clicking "Add manually"

```
Notice: Object of class stdClass could not be converted to int in /srv/www/tangaroa/public_html/wp-content/plugins/wc-admin/includes/features/onboarding/class-wc-admin-onboarding-tasks.php on line 124
```

client/dashboard/task-list/index.js
joshuatf: Updated!

includes/features/onboarding/class-wc-admin-onboarding-tasks.php
joshuatf: Updated this to pass if published or draft products exist.  I think either is in the nature of setting up a product and should probably count as completed.

joshuatf:
Thanks for the reviews @timmyc and @psealock!

Paul, I updated per your feedback, can you double check and make sure things are working error-free?

>Still seeing some of the persisting problems with nav.

I am as well.  Issue logged [here](https://github.com/woocommerce/woocommerce-admin/issues/2668) and will try to address soon.

PR #2674, opened 2019-07-19
Remove path from URL in root/dashboard page

### LINKED ISSUE #2668: Path gets added to the URL in the root (dashboard) page
**Describe the bug**
When visiting the dashboard page, any route/query changes under the same page result in path `&path=%2f` getting added to the URL.  When refreshing, this results in a blank page.

**To Reproduce**

1. Go to the dashboard `wp-admin/admin.php?page=wc-admin`.
2. Navigate within the dashboard to append a query to the URL (see https://github.com/woocommerce/woocommerce-admin/pull/2667 for an example of this).
3. Note the new URL (e.g., `wp-admin/admin.php?page=wc-admin&path=%2F&task=products`.
4. Refresh the page and note the blank page.

**Expected behavior**
Expected the path to not be added under the dashboard.

**Screenshots**
Screen Shot 2019-07-18 at 7 18 44 PM

joshuatf:
Fixes #2668

Remove the path param from the URL on dashboard pages.  Previously `updateQueryString` would add `&path=%2f` to the URL causing errors on route change or refresh.

1.  Enable the onboarding profiler or task list.
2. Try continuing to the next step, navigating back and forth, and refreshing.
3. Make sure `&path=...` does not get appended to any routes in the dashboard.
4. Make sure other routes continue to work as expected.

### DIFF
--- a/packages/navigation/src/index.js
+++ b/packages/navigation/src/index.js
@@ -92,7 +92,11 @@ export function getSearchWords( query = navUtils.getQuery() ) {
  * @return {String}  Updated URL merging query params into existing params.
  */
 export function getNewPath( query, path = getPath(), currentQuery = getQuery() ) {
-	return addQueryArgs( 'admin.php', { page: 'wc-admin', path, ...currentQuery, ...query } );
+	const args = { page: 'wc-admin', ...currentQuery, ...query };
+	if ( '/' !== path ) {
+		args.path = path;
+	}
+	return addQueryArgs( 'admin.php', args );
 }
 
 /**


jeffstieler:
> Try continuing to the next step, navigating back and forth, and refreshing.

I guess I don't know how to test this.. 

I'm not seeing any navigation, and it seems each task list item is a noop click:

Screen Shot 2019-07-19 at 9 21 25 AM

Also, I'm seeing a React error in the console:

`react-dom.js?ver=16.8.4:500 Warning: Invalid DOM property `class`. Did you mean `className`?`

timmyc:
Guessing @joshuatf might have been working on this along-side #2667 - which fixes the classname warning, along with hooking up a child step to Products instead of a `noop`. When testing this stand-alone, I got the same issues as you @jeffstieler but then rebased against master and all is well.

Pushed up the rebased branch just now if you want to take it for another spin.

PR #2676, opened 2019-07-19
Fix missing nav on connected admin pages.

jeffstieler:
The nav bar disappeared from connected admin pages after https://github.com/woocommerce/woocommerce-admin/commit/881f209.

The root cause was calling `WC_Admin_Loader::is_admin_page()` before the `current_screen` hook had fired. This caused the page controller class to return `false` for all calls to `get_current_page()`, breaking the rest of the plugin's detection for enqueuing scripts, etc. 

This PR seeks to change the onboarding reset callback to the `current_screen` hook as well as introduce a "doing it wrong" if `get_current_page()` is called before `current_screen`.

- Go to WooCommerce > Orders
- Verify the nav bar displays
- Go to WooCommerce > Settings
- Verify the nav bar displays


timmyc:
Oh icym the chatter in #wc-admin, a rebase should fix travis on this branch.

PR #2679, opened 2019-07-19
Import all used Gutenberg component styles.

### LINKED ISSUE #2322: postcss.config.js overrides not working
Some of the overrides like `outlines` and `toggle` defined in https://github.com/woocommerce/woocommerce-admin/blob/master/postcss.config.js don't seem to take affect correctly. The blue color is coming through.

Screen Shot 2019-05-28 at 9 09 49 AM

It also seems like at least `secondary` should be a different color, from the other elements. Right now all colors are set to the same shade.

This also affects Muriel/onboarding styles: 
https://github.com/woocommerce/woocommerce-admin/pull/2313#discussion_r287941495

To Test:
* In a file like `dashboard/index.js` add the following:

`import { Button, FormToggle, FormTokenField } from '@wordpress/components';`

```
	constructor() {
		super();
		this.state = {
			textInput: '',
			selectValue1: '1st',
			selectValue2: '',
			checked: false,
			tokens: [],
			suggestions: [ 'Africa', 'America', 'Antarctica', 'Asia', 'Europe', 'Oceania' ],
		};
	}
	renderDashboardOutput() {
		const { query, path } = this.props;

		return (
			<Fragment>
				<Button isPrimary>Button</Button>
				<Button isTertiary>Button 2</Button>

				<FormToggle
					checked={ this.state.checked }
					onChange={ () => this.setState( state => ( { checked: ! state.checked } ) ) }
				/>

				<FormTokenField
					value={ this.state.tokens }
					suggestions={ this.state.suggestions }
					onChange={ tokens => this.setState( { tokens } ) }
					placeholder="Type a continent"
					isBorderless
				/>
			</Fragment>
		);
....
```
* View the dashboard and note some of the blue styles coming through on the toggle or borderless token input.

jeffstieler:
Fixes postcss theme overrides.

Fixes #2322 

This PR seeks to fix the postcss theme overrides by importing in the styles from Gutenberg into WooCommerce Admin.

Postcss-themes will not render out the theme style if it doesn't find corresponding usage in the SCSS it's processing. See: https://github.com/WordPress/gutenberg/blob/c7e5b403509f7bd0d577da48dc5f582b9f9f97a2/packages/postcss-themes/index.js#L12

- Follow the test instructions on the issue
- Verify that the toggle is Woo purple
- Smoke test other areas of WooCommerce Admin to make sure styles are as expected

### DIFF
--- a/client/stylesheets/shared/_gutenberg-components.scss
+++ b/client/stylesheets/shared/_gutenberg-components.scss
@@ -5,4 +5,13 @@ Import Gutenberg component SCSS so webpack's postcss process can handle theme-in
 allows Woo themed components based on the config found in postcss.config.js
  */
 @import 'gutenberg-components/button/style.scss';
+@import 'gutenberg-components/checkbox-control/style.scss';
+@import 'gutenberg-components/dashicon/style.scss';
+@import 'gutenberg-components/form-toggle/style.scss';
+@import 'gutenberg-components/icon-button/style.scss';
+@import 'gutenberg-components/notice/style.scss';
+@import 'gutenberg-components/select-control/style.scss';
 @import 'gutenberg-components/snackbar/style.scss';
+@import 'gutenberg-components/spinner/style.scss';
+@import 'gutenberg-components/text-control/style.scss';
+@import 'gutenberg-components/tooltip/style.scss';
--- a/client/dashboard/style.scss
+++ b/client/dashboard/style.scss
@@ -74,3 +74,29 @@
 		padding-bottom: 10px;
 	}
 }
+
+.woocommerce-dashboard__body {
+	background: $muriel-gray-0;
+	color: $muriel-gray-600;
+	font-family: $default-font;
+
+	#wpbody-content {
+		min-height: 100vh;
+	}
+
+	/* Hide wp-admin and WooCommerce elements when the dashboard body class is present */
+	#adminmenumain,
+	.woocommerce-layout__header,
+	.update-nag,
+	.woocommerce-store-alerts,
+	.woocommerce-message,
+	.notice,
+	.error,
+	.updated {
+		display: none;
+	}
+
+	#wpcontent {
+		margin-left: 0;
+	}
+}
--- a/client/stylesheets/abstracts/_mixins.scss
+++ b/client/stylesheets/abstracts/_mixins.scss
@@ -42,12 +42,84 @@
 }
 
 // Gutenberg mixins. These are temporary until Gutenberg's mixins are exposed.
+/**
+ * Breakpoint mixins
+ */
+
+@mixin break-huge() {
+	@media (min-width: #{ ($break-huge) }) {
+		@content;
+	}
+}
+
+@mixin break-wide() {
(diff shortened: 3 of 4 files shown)


client/stylesheets/shared/_gutenberg-components.scss
```diff
@@ -5,3 +5,12 @@ Import Gutenberg component SCSS so webpack's postcss process can handle theme-in
...
 @import 'gutenberg-components/button/style.scss';
+@import 'gutenberg-components/checkbox-control/style.scss';
+@import 'gutenberg-components/dashicon/style.scss';
+@import 'gutenberg-components/form-toggle/style.scss';
+@import 'gutenberg-components/icon-button/style.scss';
+@import 'gutenberg-components/notice/style.scss';
+@import 'gutenberg-components/select-control/style.scss';
+@import 'gutenberg-components/spinner/style.scss';
+@import 'gutenberg-components/text-control/style.scss';
+@import 'gutenberg-components/tooltip/style.scss';
```
psealock: I tried adding `@import 'gutenberg-components/form-token-field/style.scss';` because I saw form field tokens look off:

Screen Shot 2019-07-22 at 2 40 46 PM

And it led me down a path of adding variable after variable. It reminded me that adding variables here was simply a stop gap until CSS could be exported from Gutenberg independently. Otherwise, we'll be forever configuring these. More info here, https://github.com/WordPress/gutenberg/pull/7716#issuecomment-404468856.

It seems like time to make an upstream contribution. Since we don't use these tokens (at the moment), we can leave as is.

jeffstieler: I'm not grokking what the upstream contribution would be to export SCSS from Gutenberg based on that PR discussion. Is the change to move the styling to the JS files?

psealock: No, but that was a potential solution. `postcss-loader` uses the `postcss.config.js` file to apply Woo branding to styles in Gutenberg repository. For example, here is a style from [form token field](https://github.com/WordPress/gutenberg/blob/master/packages/components/src/form-token-field/style.scss#L90):

```css
.components-form-token-field__token-text {
    background: transparent;
    color: theme(secondary);
}
```

We need to explicitly `@import 'gutenberg-components/form-token-field/style.scss'` so that our build step can process the `theme` function call and apply Woo's `secondary` color. This requires that we also import all scss variables also found in the file, which is what this PR does.

At the very least, Gutenberg will export all its variables, includes, and mixins in a package so that we don't have to go through this process on every update to `@wordpress/components` and instead consume the exported variables at build time. At the very best, our build step builds and themes css at build time without us having to do anything, as would be the case with css in js.

psealock:
@justinshreve Does Onboarding require these themed components? If so, we should merge this. If not, we can focus on an upstream solution in Gutenberg, explained in https://github.com/woocommerce/woocommerce-admin/pull/2679#discussion_r306071187.

jeffstieler:
Is this OK to be merged @psealock?

psealock:
@jeffstieler Yup, lets merge.

PR #2681, opened 2019-07-22
Onboarding: Add WebPreview component for theme previewing

### LINKED ISSUE #2469: Onboarding - Profile Wizard: Theme Demo Browser
Depends on #2159.

This issue involves implementing a demo dialog/modal for viewing the theme in different screen sizes. The mockup shows mobile, tablet, and desktop options.

See https://github.com/Automattic/wp-calypso/tree/master/client/components/web-preview for prior art, or we maybe we can export this to the new `calypso-ui` package.

joshuatf:
Fixes #2469

Adds a `WebPreview` component to load content into an iframe.  Used to preview themes in the theme step in onboarding.

### Screenshots
Screen Shot 2019-07-22 at 4 21 21 PM

1. Visit the theme step in onboarding `wp-admin/admin.php?page=wc-admin&path&step=theme`.
2. Click on "Live Demo" for any of the themes.
3.  Make sure the loader is shown when loading and that the theme preview is loaded into the iframe.
4. Check that content is resized based on the device toggle (desktop/tablet/mobile).
5. Make sure that clicking "Choose" selects the theme and completes the onboarding profiler.

### DIFF
--- a/client/dashboard/profile-wizard/steps/theme/preview.js
+++ b/client/dashboard/profile-wizard/steps/theme/preview.js
@@ -50,6 +50,7 @@ class ThemePreview extends Component {
 					</Button>
 					<div className="woocommerce-theme-preview__theme-name">
 						{ interpolateComponents( {
+							/* translators: Describing who a previewed theme is developed by. E.g., Storefront developed by WooCommerce */
 							mixedString: sprintf(
 								__( '{{strong}}%s{{/strong}} developed by WooCommerce', 'woocommerce-admin' ),
 								title
--- a/client/dashboard/profile-wizard/steps/theme/preview.js
+++ b/client/dashboard/profile-wizard/steps/theme/preview.js
@@ -46,7 +46,7 @@ class ThemePreview extends Component {
 			<div className="woocommerce-theme-preview">
 				<div className="woocommerce-theme-preview__toolbar">
 					<Button className="woocommerce-theme-preview__close" onClick={ onClose }>
-						<i className="material-icons-outlined">arrow_back</i>
+						<i className="material-icons-outlined">close</i>
 					</Button>
 					<div className="woocommerce-theme-preview__theme-name">
 						{ interpolateComponents( {
--- a/client/dashboard/profile-wizard/steps/theme/preview.js
+++ b/client/dashboard/profile-wizard/steps/theme/preview.js
@@ -1,0 +1,82 @@
+/** @format */
+/**
+ * External dependencies
+ */
+import { __, sprintf } from '@wordpress/i18n';
+import { Button } from 'newspack-components';
+import { Component } from '@wordpress/element';
+import interpolateComponents from 'interpolate-components';
+
+/**
+ * WooCommerce dependencies
+ */
+import { WebPreview } from '@woocommerce/components';
+
+const devices = [
+	{
+		key: 'mobile',
+		icon: 'phone_android',
+	},
+	{
+		key: 'tablet',
+		icon: 'tablet_android',
+	},
+	{
+		key: 'desktop',
+		icon: 'desktop_windows',
+	},
+];
+
+class ThemePreview extends Component {
+	constructor() {
+		super( ...arguments );
+
+		this.state = {
+			device: 'desktop',
+		};
+	}
+
+	render() {
+		const { onChoose, onClose, theme } = this.props;
+		const { demo_url, slug, title } = theme;
+		const { device: currentDevice } = this.state;
+
+		return (
+			<div className="woocommerce-theme-preview">
+				<div className="woocommerce-theme-preview__toolbar">
+					<Button className="woocommerce-theme-preview__close" onClick={ onClose }>
+						<i className="material-icons-outlined">arrow_back</i>
+					</Button>
+					<div className="woocommerce-theme-preview__theme-name">
+						{ interpolateComponents( {
+							mixedString: sprintf(
+								__( '{{strong}}%s{{/strong}} developed by WooCommerce', 'woocommerce-admin' ),
+								title
+							),
+							components: {
+								strong: <strong />,
+							},
+						} ) }
+					</div>
+					<div className="woocommerce-theme-preview__devices">
+						{ devices.map( device => (
+							<Button
+								key={ device.key }
+								className="woocommerce-theme-preview__device"
+								onClick={ () => this.setState( { device: device.key } ) }
+							>
+								<i className="material-icons-outlined">{ device.icon }</i>
+							</Button>
+						) ) }
+					</div>
+					<Button isPrimary onClick={ () => onChoose( slug ) }>
+						{ __( 'Choose', 'woocommerce-admin' ) }
+					</Button>
+				</div>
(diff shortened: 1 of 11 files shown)


joshuatf:
Thanks for taking a look @jameskoster!

>I hope there's a way we can hide the bar at the bottom of the preview?

Yes, but this is being handled on the WCCOM side (both the demo URL and bar itself come from WCCOM).  I know we've had some discussion around it, but I'll open an issue there if one does not already exist.

Let me know if you see other areas you'd like to change around the preview component screen.

client/dashboard/profile-wizard/steps/theme/preview.js
```diff
@@ -0,0 +1,82 @@
...
+		return (
+			<div className="woocommerce-theme-preview">
+				<div className="woocommerce-theme-preview__toolbar">
+					<Button className="woocommerce-theme-preview__close" onClick={ onClose }>
+						<i className="material-icons-outlined">arrow_back</i>
+					</Button>
+					<div className="woocommerce-theme-preview__theme-name">
+						{ interpolateComponents( {
+							mixedString: sprintf(
+								__( '{{strong}}%s{{/strong}} developed by WooCommerce', 'woocommerce-admin' ),
```
timmyc: Minor but if we ever intend on having any Marketplace developed themes, or others not developed by us, we might want to have this label be a prop from the theme itself.

packages/components/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Add new component `<List />` for displaying interactive list items.
 - Fix z-index issue in `<Chart />` empty message.
 - Added a new `<SimpleSelectControl />` component.
+- Added a new `<WebPreview />` component.
```
timmyc: 🌟

packages/components/src/web-preview/index.js
```diff
@@ -0,0 +1,90 @@
...
+
+WebPreview.propTypes = {
+	/**
+	 * Additional class name to style the component.
+	 */
+	className: PropTypes.string,
+	/**
+	 * Content shown when iframe is still loading.
+	 */
+	loading: PropTypes.node,
```
timmyc: This prop name made me do a double take - perhaps it is just too close to `isLoading`. What do you think about `loadingContent` or `loadingComponent`?

client/dashboard/profile-wizard/steps/theme/preview.js
joshuatf: That's a good idea.  I'll probably leave this as is for now since we would need more information from the WCCOM side.  But if we do implement something like this, maybe a `source` field in the API response would be enough.

packages/components/src/web-preview/index.js
joshuatf: Agreed.  I originally named it `loadingContent` but that threw an error (reserved name if I remember correctly).  Will update this before merging.

joshuatf:
>The one ux thing that felt mildly off to me was the lack of an ".is-selected" state on the device toggles in the upper right.

Definitely feels like something is needed.  I did not see anything in the designs, but maybe I missed something.  @jameskoster any ideas here?

>( side note, that text looks off in those tabs )

Hmm, the tabs are still showing centered for me @timmyc.  Browser/WP/WC version?

jameskoster:
Curiously Material does not require or even suggest a selected state for action items in an App Bar. I'll follow up with the Muriel team for clarification on this. 

For now, we should ensure the hover / focus styles are applied (ideally with the ripple on click). You can find examples (and code I hope) for this [here](https://material-ui.com/components/app-bar).

jameskoster:
Spoke with the Muriel team, I think for maximum accessibility (based on Gutenberg patterns) the icon / background should be inversely colored when active. I updated the design in Figma to include an example.

timmyc:
> Hmm, the tabs are still showing centered for me @timmyc. Browser/WP/WC version?

I'm running Firefox 68.0.1, WP 5.2.2, WC 3.6.5

timmyc:
@joshuatf setting this to in progress pending changes from the latest round of feedback.

joshuatf:
>I updated the design in Figma to include an example.

Fantastic- thanks @jameskoster!  Added the focus/selected colors.

I held off on adding the ripple effect as I think that's a substantial PR on its own and might be needs to be added directly to newspack-components to avoid a decent amount of code duplication.  I'll open some issues around this.

>I'm running Firefox 68.0.1, WP 5.2.2, WC 3.6.5

I see the same in Firefox.  Thanks for noting that @timmyc - will get an issue opened!

Updated the `loadingContent` prop (seems to work as long as we don't pass to the iframe) and rebased this branch.  This one is ready for another review.

packages/components/src/web-preview/index.js
```diff
@@ -0,0 +1,90 @@
...
+
+WebPreview.propTypes = {
+	/**
+	 * Additional class name to style the component.
+	 */
+	className: PropTypes.string,
+	/**
+	 * Content shown when iframe is still loading.
+	 */
+	loadingContent: PropTypes.node,
```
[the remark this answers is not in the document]
joshuatf: Thanks 🤦‍♂

client/dashboard/profile-wizard/steps/theme/preview.js
```diff
@@ -0,0 +1,85 @@
...
```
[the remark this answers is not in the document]
joshuatf: Good idea- added.

joshuatf:
Thanks for the review @justinshreve- updated per your feedback.

>It also looks like the Deli and Bistro themes do not have a demo button -- but they do have a demo_url in the response.

I see them with demo buttons.  Do you have these themes installed?  If so, previewing for installed themes is not available for the time being.

justinshreve:
> I see them with demo buttons. Do you have these themes installed? If so, previewing for installed themes is not available for the time being.

I do not have either of these themes installed.

I looked at the CSS and it looks like they are actually getting pushed down below another theme. I removed this `height: 100%` and things worked correctly.

Before:

Screen Shot 2019-07-26 at 1 45 49 PM


After:

Screen Shot 2019-07-26 at 1 44 38 PM

joshuatf:
@justinshreve Not quite able to replicate, but can remove that line and replace with `margin-top: auto;` to keep things working.

Updated in https://github.com/woocommerce/woocommerce-admin/pull/2681/commits/52d83e6fe8ed57dcebb428eeb2d74c970ae54993.  Let me know if it looks okay in your browser now.

jameskoster:
There's a couple of (hopefully small) things I think we should tweak:

• When the preview is open, instead of a back arrow in the App Bar let’s use an `x` to close out the preview
• Can we increase the space between the “Choose” button and the device switcher to 16px?
• When previewing “desktop” mode, can the preview occupy the full screen real estate beneath the App Bar?

joshuatf:
Thanks for the feedback @jameskoster.  Updated those 3 items, can you take another look?

Screen Shot 2019-07-31 at 4 55 49 PM

justinshreve:
Great work Josh! 🚢 it.

Issue #1849, opened 2019-03-19
Category report failing to render comparison properly


LevinMedia:
**Describe the bug**
Category report failing to render comparison properly

**To Reproduce**

1. Go to the category report
2. Create a comparison
3. Change the date range to a quarter or longer (ie: last quarter, last year etc)
4. Ensure your period is set to Day
5. Note the complete comparison is not rendered, and as a result, the data in the chart legend is incorrect. 
6. Change the interval to Week, note the comparison renders for the entire period, and the totals are correct in the legend. 

**Expected behavior**
Always render the entire comparison, and display the correct totals in the legend.

**Screenshots**

![](https://cl.ly/6f8a6d7ba2e9/Screen%20Shot%202019-03-19%20at%203.08.21%20PM.png)
missing comparison

![](https://cl.ly/2bba4945d7bb/Screen%20Shot%202019-03-19%20at%203.08.47%20PM.png)
switching to "Weekly" interval renders the entire time period

LevinMedia:
May be related... when looking at a comparison with the date range set to year and the interval to day, nothing loads: 

<img width="2022" alt="Screen Shot 2019-03-21 at 8 18 46 PM" src="https://user-images.githubusercontent.com/4500952/54798654-e961c100-4c16-11e9-86af-f628bb6c1777.png">

PR #2708, opened 2019-07-26
Fix groupby for category segmentation in reports.

jeffstieler:
Found while trying to diagnose #1849.

This PR seeks to fix the database column used for grouping when segmenting reports by category.

Use `term_id` which correlates to the actual taxonomy term rather than `term_taxonomy_id` which is an ID used in the relationship table.

### Screenshots

Before:

Screen Shot 2019-07-25 at 3 14 29 PM

After:

Screen Shot 2019-07-25 at 3 29 30 PM

I used the WCCOM dataset to test this.

- Compare the Themes and Storefront Extension categories (Categories Report) for the 2018 year.
- Note that the chart does not render data, but the table is correct
- Check out this branch
- Note that the chart does render data

### Changelog Note:

Fix: chart display when comparing categories.

### DIFF
--- a/includes/class-wc-admin-reports-coupons-stats-segmenting.php
+++ b/includes/class-wc-admin-reports-coupons-stats-segmenting.php
@@ -295,7 +295,7 @@ class WC_Admin_Reports_Coupons_Stats_Segmenting extends WC_Admin_Reports_Segment
 			RIGHT JOIN {$wpdb->prefix}term_taxonomy ON {$wpdb->prefix}term_relationships.term_taxonomy_id = {$wpdb->prefix}term_taxonomy.term_taxonomy_id
 			";
 			$segmenting_where          = " AND taxonomy = 'product_cat'";
-			$segmenting_groupby        = 'wp_term_taxonomy.term_taxonomy_id';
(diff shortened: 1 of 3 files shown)


mikejolley:
Since term splitting came into effect, both IDs should be identical? (found this PR whilst dealing with merge conflict)

jeffstieler:
> Since term splitting came into effect, both IDs should be identical?

@mikejolley 

I had to refresh myself on term splitting - but it's my understanding that would create new `term_id`s for terms shared between taxonomies.

The problem I ran into that prompted this PR was that the `term_id` was being provided by the comparison UX and being compared to the `taxonomy_term_id` and they were not matching up.

mikejolley:
> I had to refresh myself on term splitting - but it's my understanding that would create new term_ids for terms shared between taxonomies.

Correct. In which case, the taxonomy_term_id and term_id would be identical. Term splitting was introduced to keep those IDs the same. Not sure why the issue would come up though..

rrennick:
> Since term splitting came into effect, both IDs should be identical?

This couldn't be relied on for sites that existed before term splitting came into effect.

PR #2710, opened 2019-07-29
Fix category report comparison chart render.

### LINKED ISSUE #1849: Category report failing to render comparison properly

jeffstieler:
Remove limit from segmentation query and avoid adding dates outside of the initial interval query to response.

Fixes #1849.

This PR seeks to fix the rendering of the chart component when comparing Categories on a report.

It does so by removing the `LIMIT`clause from the query (which seems to cause necessary segments to be omitted, and unnecessary ones to be included) and by omitting time intervals from the response for the given paging parameters.

I'm leaving this as "In Progress" because I'd like some feedback from @peterfabian who originally implemented this.

### Screenshots

Screen Shot 2019-07-26 at 4 12 38 PM

- Go to the category report
- Create a comparison
- Change the date range to a quarter or longer (ie: last quarter, last year etc)
- Ensure your period is set to Day
- Note the complete comparison is rendered

### Changelog Note:

Fix: chart display when comparing categories.

### DIFF
--- a/src/API/Reports/Products/Stats/Segmenter.php
+++ b/src/API/Reports/Products/Stats/Segmenter.php
@@ -99,8 +99,9 @@ class Segmenter extends ReportsSegmenter {
 		$product_segmenting_table = $wpdb->prefix . 'wc_order_product_lookup';
 
 		// LIMIT offset, rowcount needs to be updated to LIMIT offset, rowcount * max number of segments.
-		$limit_parts      = explode( ',', $intervals_query['limit'] );
-		$orig_rowcount    = intval( $limit_parts[1] );
+		$limit_parts   = explode( ',', $intervals_query['limit'] );
+		$orig_rowcount = intval( $limit_parts[1] );
+		// @todo - this is breaking intervals.
 		$segmenting_limit = $limit_parts[0] . ',' . $orig_rowcount * count( $this->get_all_segments() );
 
 		// Can't get all the numbers from one query, so split it into one query for product-level numbers and one for order-level numbers (which first need to have orders uniqued).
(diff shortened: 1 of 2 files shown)


rrennick:
@jeffstieler This worked for me on the comparison but I'm still able to reproduce the partial chart on a single category. I think that is a separate issue in the client and should be done as a separate PR.

I didn't approve but this PR LGTM. @peterfabian do you have an issue with this being merged?

peterfabian:
I don't have an issue, just think restricting a category and keeping the limit is a better approach than removing the limit (if that works in other contexts, I haven't tested it extensively), as it reduces the amount of data needed to be processed and keeps the limit as a kind of fail-safe. I can try to draft an alternative PR if that helps?

rrennick:
> just think restricting a category and keeping the limit is a better approach than removing the limit

It might be due to multiple products on orders and/or products in multiple categories but the data retrieved contains data that filtered out by the segmenting functions. The limit on the query prevents all of the inclusive data for the segment from being retrieved.

jeffstieler:
At a minimum, I'd like to clean the code up before merging - but ultimately I'd really like to keep a `LIMIT` around. I'll try out @peterfabian's suggestion after working the PSR-4 stuff. If you want to open an alternate PR @peterfabian - go for it!

jeffstieler:
I've incorporated @peterfabian's proposed changes and fixed the `LIMIT` clause for category comparisons. This is ready for another review.

PR #2733, opened 2019-08-01
add version string to _doing_it_wrong

### LINKED ISSUE #2732: Doing it wrong function produces fatal error
**Describe the bug**
When triggering a scenario where the _doing_it_wrong function is called we get a fatal error. Here is the error we get:

```
Fatal error: Uncaught ArgumentCountError: Too few arguments to function _doing_it_wrong(), 2 passed in /home/www/wp-content/plugins/woocommerce-admin/includes/page-controller/class-wc-admin-page-controller.php on line 193 and exactly 3 expected in /home/www/wp-includes/functions.php:4735
Stack trace:
#0 /home/www/wp-content/plugins/woocommerce-admin/includes/page-controller/class-wc-admin-page-controller.php(193): _doing_it_wrong('get_current_pag...', 'Current page re...')
#1 /home/www/wp-content/plugins/woocommerce-admin/includes/page-controller/class-wc-admin-page-controller.php(368): WC_Admin_Page_Controller->get_current_page() 
#2 /home/www/wp-content/plugins/woocommerce-admin/includes/page-controller/page-controller-functions.php(49): WC_Admin_Page_Controller->is_registered_page() 
#3 /home/www/wp-content/plugins/woocommerce-admin/includes/class-wc-admin-loader.php(342): wc_admin_is_registered_page() 
#4 /home/www/ in /home/www/wp-includes/functions.php on line 4735
```

**To Reproduce**
I don't have any specific steps to reproduce this, but the user who reported this, mentioned it happened after activating the WooCommerce Social Login plugin: https://woocommerce.com/products/woocommerce-social-login/

**Expected behavior**
According to the documentation the function should be passing three arguments: https://developer.wordpress.org/reference/functions/_doing_it_wrong/

It seems this function was added to version 0.16: https://github.com/woocommerce/woocommerce-admin/commit/57281970a218ff35e7f8ffb312d20a9c84257ca2

**Affected ticket**
2240407-zen

rrennick:
See #2732 

This PR adds the third parameter to the `_doing_it_wrong` call.

- Add the following hook
```
	(new WC_Admin_Page_Controller)->get_current_page();
});
```
- Your debug log should have the `_doing_it_wrong` warning with no other error

### Changelog Note:

Fix: Add version parameter to `_doing_it_wrong` on `current_screen`.

### DIFF
--- a/includes/page-controller/class-wc-admin-page-controller.php
+++ b/includes/page-controller/class-wc-admin-page-controller.php
@@ -190,7 +190,7 @@ class WC_Admin_Page_Controller {
 		// If 'current_screen' hasn't fired yet, the current page calculation
 		// will fail which causes `false` to be returned for all subsquent calls.
 		if ( ! did_action( 'current_screen' ) ) {
-			_doing_it_wrong( __FUNCTION__, esc_html__( 'Current page retreival should be called on or after the `current_screen` hook.', 'woocommerce-admin' ) );
+			_doing_it_wrong( __FUNCTION__, esc_html__( 'Current page retreival should be called on or after the `current_screen` hook.', 'woocommerce-admin' ), '0.12.0' );
 		}
 
 		if ( is_null( $this->current_page ) ) {
--- a/includes/page-controller/class-wc-admin-page-controller.php
+++ b/includes/page-controller/class-wc-admin-page-controller.php
@@ -190,7 +190,7 @@ class WC_Admin_Page_Controller {
 		// If 'current_screen' hasn't fired yet, the current page calculation
 		// will fail which causes `false` to be returned for all subsquent calls.
 		if ( ! did_action( 'current_screen' ) ) {
-			_doing_it_wrong( __FUNCTION__, esc_html__( 'Current page retreival should be called on or after the `current_screen` hook.', 'woocommerce-admin' ), '0.12.0' );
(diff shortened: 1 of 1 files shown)


includes/page-controller/class-wc-admin-page-controller.php
```diff
@@ -190,7 +190,7 @@ public function get_current_page() {
 		// If 'current_screen' hasn't fired yet, the current page calculation
 		// will fail which causes `false` to be returned for all subsquent calls.
 		if ( ! did_action( 'current_screen' ) ) {
-			_doing_it_wrong( __FUNCTION__, esc_html__( 'Current page retreival should be called on or after the `current_screen` hook.', 'woocommerce-admin' ) );
+			_doing_it_wrong( __FUNCTION__, esc_html__( 'Current page retreival should be called on or after the `current_screen` hook.', 'woocommerce-admin' ), '0.12.0' );
```
jeffstieler: This should use `WC_ADMIN_VERSION_NUMBER`, and the current version is `0.16.0`.

rrennick: @jeffstieler Thanks :) Based on https://developer.wordpress.org/reference/functions/_doing_it_wrong/ it should use the '0.16.0' string vs the constant which will change with each release. I have updated the PR with the correct version.

jeffstieler: Nevermind on the version constant - I'm still ☕️ing. 😄

PR #2742, opened 2019-08-02
Add form component for handling form state

### LINKED ISSUE #2741: Form component needed to handle form state
When creating the onboarding profiler, we created a state in each component to handle form values and validation.  While this works, it's not very DRY.

This has become very apparent while working on https://github.com/woocommerce/woocommerce-admin/issues/2482 where there is no easy way to share fields that already exist.

A component to wrap the state that contains field values, error handling, and form submission would greatly reduce the amount of code going forward and mental overhead needed to reproduce the validation logic in each component.

joshuatf:
Fixes #2741

* Adds a Form component and example to handle form state and interactions.
* Modifies the existing store details step to use the new form component.

Note: This still needs to be done for business details, but I may address in a follow-up as it requires a little more refactor.

### Screenshots
Screen Shot 2019-08-02 at 5 01 07 PM

1. Visit the Form component in devdocs and the store details step of onboarding `/wp-admin/admin.php?page=wc-admin&step=store-details`
2.  Test the form validation by add/removing values and attempting to submit.
3.  Make sure no regressions have occurred in the store details step.

### DIFF
--- a/packages/components/src/form/index.js
+++ b/packages/components/src/form/index.js
@@ -9,93 +9,93 @@ import PropTypes from 'prop-types';
  * A form component to handle form state and provide input helper props.
  */
 class Form extends Component {
-    constructor( props ) {
-        super();
+	constructor( props ) {
+		super();
 
 		this.state = {
-            values: props.initialValues,
-            errors: props.errors,
-            touched: props.touched,
-        };
+			values: props.initialValues,
+			errors: props.errors,
+			touched: props.touched,
+		};
 
 		this.getInputProps = this.getInputProps.bind( this );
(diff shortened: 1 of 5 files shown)


packages/components/src/form/index.js
```diff
@@ -0,0 +1,127 @@
...
+import PropTypes from 'prop-types';
+
+/**
+ * A form component to handle form state and provide input helper props.
+ */
+class Form extends Component {
+    constructor( props ) {
+        super();
+
+		this.state = {
```
justinshreve: Theres a bit of weird indentation going on in this file, starting with this line here.

PR #2744, opened 2019-08-02
Short circuit admin title filter when applied by third parties too early.

jeffstieler:
Bail out of our `admin_title` filter if it's invoked before `current_screen`.

Prevents issues like https://github.com/xwp/pwa-wp/issues/206 from triggering the doing it wrong error.

- Invoke the error by using the PWA plugin -or-
```php
        apply_filters( 'admin_title', 'test', 'test' );
});
```


rrennick:
Also closed #2732 .

PR #2751, opened 2019-08-05
Add `help` prop to SimpleSelectControl to properly show errors.

### LINKED ISSUE #2750: SimpleSelectControl errors (help) text is not displayed
**Describe the bug**
In the onboarding profiler, the errors for `SimpleSelectControl` do not show since `SimpleSelectControl` currently does not have a `help` prop.  We can either:

1. Add a `help` prop to the control.
2. Manually add the errors as a span underneath these controls in the form (`errors` can be passed down to children from the `Form` component).

I'm more in favor of option 1, but cc @justinshreve for your opinion.

**Screenshot**
Screen Shot 2019-08-05 at 1 21 11 PM

**To Reproduce**

1. Visit `wp-admin/admin.php?page=wc-admin&step=business-details`
2. Try to hit "Continue" without selecting any options.
3. Note the fields get highlighted in red, but the error text is not shown.

justinshreve:
Fixes #2750.

This PR adds the `help` prop (like other WP/Muriel components have), so that we can properly show validation errors below the input.

### Screenshots

Screen Shot 2019-08-05 at 12 42 53 PM

* Visit wp-admin/admin.php?page=wc-admin&step=business-details
* Try to hit "Continue" without selecting any options.
* Note the fields get highlighted in red, and the error text is shown.

### Changelog Note:
None needed (Component not released yet)

### DIFF
--- a/packages/components/src/simple-select-control/index.js
+++ b/packages/components/src/simple-select-control/index.js
@@ -3,10 +3,11 @@
  */
 import { __, sprintf } from '@wordpress/i18n';
 import { Dropdown, Button, NavigableMenu, withFocusOutside } from '@wordpress/components';
-import { Component } from '@wordpress/element';
+import { Fragment, Component } from '@wordpress/element';
 import { map, find } from 'lodash';
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
+import { withInstanceId } from '@wordpress/compose';
 
 /**
  * A component for displaying a material styled 'simple' select control.
@@ -62,7 +63,7 @@ class SimpleSelectControl extends Component {
 	}
 
 	render() {
-		const { options, label, className } = this.props;
+		const { options, label, className, instanceId, help } = this.props;
 		const { currentValue, isFocused } = this.state;
 		const onChange = ( value ) => {
 			this.onChange( value );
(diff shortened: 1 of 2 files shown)


packages/components/src/simple-select-control/index.js
```diff
@@ -87,19 +91,22 @@ class SimpleSelectControl extends Component {
...
+							aria-expanded={ isOpen }
+							aria-label={ ! isEmpty ? sprintf(
+								/* translators: Label: Current Value for a Select Dropddown */
+								__( '%s: %s' ), label, currentOption && currentOption.label
+							) : label }
+						>
+							<span className="woocommerce-simple-select-control__label">{ label }</span>
+							<span className="woocommerce-simple-select-control__value">{ currentOption && currentOption.label }</span>
+						</Button>
+						{ !! help && <p id={ id + '__help' } className="components-base-control__help">{ help }</p> }
```
jeffstieler: Should this class name be `woocommerce-simple-select-control__help` instead?

justinshreve: I went with `components-base-control__help` here and `components-base-control` in the `Dropdown` classes above, so that these inherit the same base WP component styles as other Muriel or WP core components. This matches the help text returned by other WP components.

PR #2753, opened 2019-08-06
restore wp-admin wpbody-content bottom padding

### LINKED ISSUE #2682: Footer message overlaps buttons in Category Edit page
It has been reported twice in the support forums:

> Category edit page update button is unclickable with the woocommerce footer message “If you like WooCommerce please leave us a ★★★★★ rating. A huge thanks in advance!” on top.

https://wordpress.org/support/topic/category-page-update-button-unclickable-with-woocommerce-footer-on-top/

https://wordpress.org/support/topic/breaks-the-catagory-edit-page/

## Screenshots

Without WooCommerce Admin:

With WooCommerce Admin:

I was unable to reproduce it, but maybe somebody else from the team can? :thinking:

rrennick:
Fixes #2682

This PR restores the bottom padding to the main content area in the dashboard. The bottom padding allowed space for the admin footer. 

Other plugins are styling the footer with more specificity that WC Admin and overriding the CSS that hides the footer. Given that's the case, should the CSS hiding the footer also be removed?

- Edit a product category
- In the browser developer console, disable the `#wpfooter { display: none }`
- The category update button should be clickable

### Changelog Note:

Fix: issue where product category update button was not always clickable

### DIFF
--- a/client/stylesheets/shared/_reset.scss
+++ b/client/stylesheets/shared/_reset.scss
@@ -14,7 +14,7 @@
 
 	@include breakpoint( '>782px' ) {
 		#wpbody-content {
-			padding: 0;
+			padding: 0 0 65px 0;
 		}
 	}
 
--- a/client/stylesheets/shared/_reset.scss
+++ b/client/stylesheets/shared/_reset.scss
@@ -14,7 +14,7 @@
 
 	@include breakpoint( '>782px' ) {
 		#wpbody-content {
-			padding: 0 0 65px 0;
+			padding-left: 0;
 		}
 	}
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -9303,7 +9303,8 @@
         },
         "ansi-regex": {
           "version": "2.1.1",
-          "bundled": true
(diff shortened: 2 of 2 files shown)


client/stylesheets/shared/_reset.scss
```diff
@@ -14,7 +14,7 @@
 
 	@include breakpoint( '>782px' ) {
 		#wpbody-content {
-			padding: 0;
+			padding: 0 0 65px 0;
```
psealock: common.css from WP Core has `padding-bottom: 65px;`, so we can eliminate this block altogether.

rrennick: common.css also has a `padding-left: 20px` that the original CSS was overriding. https://github.com/woocommerce/woocommerce-admin/pull/2753/commits/c45e3bc64af273cc56659de23923702dc0df40db updates the PR to only override the left padding.

psealock: Oh, good call. I didn't see that ✅ 

LGTM then

Issue #2482, opened 2019-06-21
Onboarding - Task List: Configure Shipping


joshuatf:
This step involves first saving a store address, if one is not already set.  If a store address is already step, this part of the step is skipped and continues to the shipping step.

<img width="623" alt="Screen Shot 2019-06-21 at 4 50 07 PM" src="https://user-images.githubusercontent.com/10561050/59910534-bbbe2b00-9444-11e9-96df-a80a15d1d4dc.png">

The next part of the step involves setting shipping zones and prices.  This will 2 shipping options:
* The store's country
* Rest of the world

@jameskoster this interface only provides flat rate shipping options unless they manually go to a separate settings page, right?

<img width="747" alt="Screen Shot 2019-06-21 at 4 54 00 PM" src="https://user-images.githubusercontent.com/10561050/59910786-32f3bf00-9445-11e9-82f8-702c75d3af7d.png">

jameskoster:
> @jameskoster this interface only provides flat rate shipping options unless they manually go to a separate settings page, right?

Correct. This is what we've done on Store on WP.com and our data suggests that it's adequate. Only ~2.5% of customer actually go and create custom shipping rules during setup.

Issue #2581, opened 2019-07-04
Text field hover / focus styles


jameskoster:
On hover let's set the border color to gray 400.

Also, I'm only seeing the focus border styles when I start typing in an input - they should be applied whenever the input is focussed.

justinshreve:
The focus style is fixed in Calypso version of the component that will be exported here https://github.com/Automattic/wp-calypso/pull/34277.

We could also fix it in Newspack (https://github.com/Automattic/newspack-plugin/issues/75) if we think the calypso ui version will be awhile.

justinshreve:
It looks like there is some activity on the Gutenberg issue that is preventing us from using the updated version of the component. 

We could fix this now but creating a local copy of the component or by making a fix in the newspack-components we are both moving away from -- but I think it makes more sense to wait and switch to the fixed version. Is it OK to move this one to the backlog @pmcpinto / @dreiris until that time?

pmcpinto:
@justinshreve

> but I think it makes more sense to wait and switch to the fixed version

Ok, I just moved this issue to the backlog

justinshreve:
Considering the Calypso/Gutenberg issues haven't had any traction, I think we should fix any issues we have in a local component for the purposes of A/B testing. So I've removed the `Blocked` label.

Issue #2574, opened 2019-07-03
Update color-studio to v2 and update color palette


justinshreve:
See p97Xot-12H-p2 for background.

We need to update our usage of `color-studio` and adapt the new colors (see incoming updated Figmas). We are using these colors in both the `WooCommerce Admin` and `Calypso` parts of our onboarding flow for colors of buttons, input elements, etc.

- [ ] Calypso
- [ ] WooCommerce Admin

justinshreve:
We are not quite ready for this yet.

See p1565289310085200-slack-wc-onboarding

PR #2760, opened 2019-08-07
Onboarding: Add shipping rate base task

joshuatf:
Fixes (part of) #2482

This PR adds in the base for the shipping task.  There are still 2 more steps needed for this task, but I'll split them up as this PR is already very lengthy (sorry about that 😓 ).

* Moves the store address fields to a reusable component.
* Moves some reusable styles for onboarding under a body class `woocommerce-onboarding`.
* Fetches shipping zones for "Rest of the world" and for the local country.
* Creates or updates existing shipping methods for said zones.

### Questions/To-do

I feel there is a lot happening in this PR and I probably should have broken it up earlier.  Here are some items that I think need to be addressed (potentially in future PRs).

* Any thoughts on the folder location/naming for setting fields?  I'm thinking this long-term may be a reusable piece for WC if the settings screens gets migrated to React.
* The completion condition check for the zones seems to work here, but long-term I think we're going to keep hitting issues by not fetching updated values to check against client-side.  If we move that way, should we show a loader for the task icon while fetching any task conditions?
* The Muriel grays look to be incorrect here.  Looks like we might need to update color studio.  Does this break any designs by doing this?  It looks like most of the grays have been darkened. /cc @jameskoster 
* When returning to the shipping task after completion, I opted to allow a full walkthrough of the steps again.  After testing both ways, I feel this way feels better, but open to feedback.

### Screenshots
Screen Shot 2019-08-07 at 6 54 14 PM

1.  Make sure styles are not broken in the onboarding profiler as this migrates a number of styles.  (Looks like there might be a few unrelated styling bugs as well).
2. Delete the `woocommerce_store_address` option from `wp_options` to test the store location step.
3. Make sure store address fields and validation work as expected.
4.  Delete existing zones and methods to test the task condition for completion is working.
5. Continue to the next step and make sure that updating rates creates or edits existing flat rate shipping methods (for the store default country, it will only match against zones that have **only** a single location matching the store's country).
6. Make sure that toggling the "Rest of the world" shipping zone on/off disables/enables the method on save.


jeffstieler:
This looks good so far @joshuatf! Two things I noticed while testing functionality:

* I can put in a nonsense postcode - like "banana" which didn't get validated at all
* The `0.00` default should probably be a placeholder in the shipping methods - you have to select all and then enter the value you want

Will move on to code review now.

justinshreve:
> Looks like we might need to update color studio. 

https://github.com/woocommerce/woocommerce-admin/issues/2574 -- Both the actual colors and variable names changed in the v2 update, so we'll need to make sure all the correct things get applied.

client/dashboard/task-list/tasks/shipping/index.js
```diff
@@ -0,0 +1,211 @@
...
+					zone.methods = await apiFetch( { path: `/wc/v3/shipping/zones/${ zone.id }/methods` } );
+					zone.name = __( 'Rest of the world', 'woocommerce-admin' );
+					zone.toggleEnabled = true;
+					shippingZones.push( zone );
+					return;
+				}
+
+				// Return any zone with a single location matching the country zone.
+				zone.locations = await apiFetch( { path: `/wc/v3/shipping/zones/${ zone.id }/locations` } );
+				if ( 1 === zone.locations.length && countryCode === zone.locations[ 0 ].code ) {
```
jeffstieler: Would we consider relaxing this a bit and just checking for the store's country to be in the zone?

In my testing I had a zone for USA+Canada, and a new US-only zone was created.

client/dashboard/components/settings/general/store-address.js
```diff
@@ -0,0 +1,114 @@
...
+		errors.addressLine1 = __( 'Please add an address', 'woocommerce-admin' );
+	}
+	if ( ! values.countryState.length ) {
+		errors.countryState = __( 'Please select a country and state', 'woocommerce-admin' );
+	}
+	if ( ! values.city.length ) {
+		errors.city = __( 'Please add a city', 'woocommerce-admin' );
+	}
+	if ( ! values.postCode.length ) {
+		errors.postCode = __( 'Please add a post code', 'woocommerce-admin' );
```
jeffstieler: Probably outside the scope of this PR, but is it possible to further validate the value? "Banana" isn't a valid postcode anywhere 🍌

client/dashboard/task-list/tasks/shipping/index.js
```diff
@@ -0,0 +1,211 @@
...
+		const { step } = this.state;
+		const steps = this.getSteps();
+		const currentStepIndex = steps.findIndex( s => s.key === step );
+		const nextStep = steps[ currentStepIndex + 1 ];
+
+		if ( nextStep ) {
+			this.setState( { step: nextStep.key } );
+		} else {
+			this.setState( { step: steps[ 0 ].key } );
+			getHistory().push( getNewPath( '/', {}, {} ) );
```
jeffstieler: This seems to be adding `0=%2F` to my query string.

client/dashboard/task-list/tasks/shipping/rates.js
```diff
@@ -0,0 +1,195 @@
...
+					) ) }
+				</div>
+
+				<Button isPrimary onClick={ this.handleSubmit }>
+					{ __( 'Complete task', 'woocommerce-admin' ) }
+				</Button>
+			</Fragment>
+		);
+	}
+}
```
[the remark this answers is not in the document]
joshuatf: Added in https://github.com/woocommerce/woocommerce-admin/pull/2760/commits/515981ce92a524009adf0e1caa0b7025269fd1d4

client/dashboard/task-list/tasks/shipping/rates.js
```diff
@@ -0,0 +1,195 @@
...
+				<div className="woocommerce-shipping-rates">
+					{ shippingZones.map( zone => (
+						<div className="woocommerce-shipping-rate" key={ zone.id }>
+							<div className="woocommerce-shipping-rate__icon">
+								{ zone.locations ? (
+									zone.locations.map( location => (
+										<Flag size={ 24 } code={ location.code } key={ location.code } />
+									) )
+								) : (
+									<i className="material-icons-outlined">public</i>
```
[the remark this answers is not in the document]
joshuatf: Added in https://github.com/woocommerce/woocommerce-admin/pull/2760/commits/515981ce92a524009adf0e1caa0b7025269fd1d4

client/dashboard/task-list/tasks/shipping/index.js
joshuatf: I'm okay to allow multiple countries if you are.  The original designs call for a single country with a single flat rate method, but this opens a can of worms for pre-existing methods if they visit the step after completion.

I think with the current refactor, it's okay to allow zones with multiple locations since we can show the flag for each place.  Also curious if we should note on this step that this applies only to flat rate methods in case a user has another method not shown here. /cc @jameskoster

Screen Shot 2019-08-08 at 1 55 00 PM

client/dashboard/components/settings/general/store-address.js
joshuatf: I was able to use "banana 🍌"  for the actual WC settings.  I don't think any validation is applied here since it can vary so much between countries.

If we want to add in validation here, we might look into a package like [this one](https://github.com/melwynfurtado/postcode-validator).  But I'm also a little hesitant to introduce validation that does not (or may conflict with a store's settings) in WC.

client/dashboard/task-list/tasks/shipping/index.js
joshuatf: Nice catch!  Wrong param order here.

joshuatf:
>Both the actual colors and variable names changed in the v2 update, so we'll need to make sure all the correct things get applied.

Thanks for noting that issue @justinshreve!  Will leave the colors as is for now then until we can update.

Thank you for the review @jeffstieler!

>The 0.00 default should probably be a placeholder in the shipping methods - you have to select all and then enter the value you want

Good point.  I think we should keep the values in place since we need the classes that are added by the control component and the value when updating the method, but I did update this to make it easier to delete/edit.

Alternatively, we could select the whole input value on click, but I think this will require that those events get passed in to the controls through the Gutenberg components.

client/dashboard/task-list/style.scss
```diff
@@ -53,3 +64,74 @@
...
+		margin-right: $gap-large;
+	}
+
+	.woocommerce-shipping-rate__name {
+		align-items: center;
+		display: flex;
+		padding-top: $gap;
+		font-size: 16px;
+		line-height: 22px;
+		color: #1a1a1a;
```
justinshreve: `$muriel-gray-900` is `#1a1a1a`.

client/dashboard/task-list/tasks/shipping/index.js
```diff
@@ -0,0 +1,212 @@
...
+
+	reset() {
+		this.setState( this.initialState );
+	}
+
+	async fetchShippingZones() {
+		this.setState( { isPending: true } );
+		const { countryCode, countryName } = this.props;
+		const shippingZones = [];
+		const zones = await apiFetch( { path: '/wc/v3/shipping/zones' } );
```
justinshreve: I think this is fine for now, but we might want to add a `@todo` here for potentially moving this work over to selectors and mutators -- especially if code starts getting lifted from here for a settings screen overhaul.

client/dashboard/task-list/tasks/shipping/index.js
```diff
@@ -0,0 +1,212 @@
...
+	render() {
+		const { isPending, step } = this.state;
+		const { isSettingsRequesting } = this.props;
+
+		return (
+			<div className="woocommerce-task-shipping">
+				<Card className="is-narrow">
+					<Stepper
+						isPending={ isPending || isSettingsRequesting }
+						isVertical={ true }
```
justinshreve: Minor: You can just pass `isVertical`.

client/dashboard/task-list/tasks/shipping/rates.js
```diff
@@ -0,0 +1,225 @@
...
+					data: {
+						method_id: 'flat_rate',
+						settings: { cost: zoneRates[ zone.id ] },
+					},
+				} );
+			}
+		} );
+
+		// @todo This is a workaround to force the task to mark as complete.
+		// This should probably be updated to use wc-api so we can fetch shipping methods.
```
justinshreve: I see you do include a note about this. Maybe have it at the top of the other file as well?

client/dashboard/task-list/tasks/shipping/index.js
jeffstieler: I defer to design on this one.

> this opens a can of worms for pre-existing methods if they visit the step after completion

Perhaps it's better to have a future issue to handle this case a little more elegantly?

It's probably more trouble than its worth to look for zones containing the store's country and only a flat rate shipping method at this time.

client/dashboard/task-list/tasks/shipping/rates.js
```diff
@@ -0,0 +1,225 @@
...
+										/>
+									) }
+								</div>
+								{ ( ! zone.toggleEnabled || enabledZones.includes( zone.id ) ) && (
+									<div
+										className={ classnames( 'woocommerce-shipping-rate__control-wrapper', {
+											'has-value': zoneRates[ zone.id ],
+										} ) }
+									>
+										<span className="woocommerce-shipping-rate__control-prefix">
```
justinshreve: This can be addressed in a follow-up issue/PR, however not all currencies will have the symbol as a prefix.

Maybe we could lift https://github.com/woocommerce/woocommerce-admin/blob/0f20052de6f9b706baaccfc6f8cb64c713c53d37/packages/components/src/filters/advanced/number-filter.js#L72-L96 into an exportable function or maybe a `PriceInput` component that does the handling for us (prior art: https://github.com/Automattic/wp-calypso/blob/97569a3c821c111dba318498cdf79d45db32a217/client/extensions/woocommerce/components/price-input/index.js).

jeffstieler:
> Good point. I think we should keep the values in place since we need the classes that are added by the control component and the value when updating the method, but I did update this to make it easier to delete/edit.

Very odd that Newspack uses the label as placeholder.. https://github.com/Automattic/newspack-plugin/blob/master/assets/components/src/text-control/index.js#L77

If the "Muriel way" is to hide labels when controls are empty - is there perhaps another way we should be presenting these components? If the styling of the Newspack/Muriel components is getting in the way of good UX I'd rather override that styling or not use those components. cc @jameskoster

justinshreve:
> or not use those components.

We are probably moving away from these in the long-term. Some related background: https://github.com/woocommerce/woocommerce-admin/issues/2581#issuecomment-515145494.

client/dashboard/task-list/style.scss
joshuatf: Thanks!  Updated here and a few other spots as well.

client/dashboard/task-list/tasks/shipping/index.js
joshuatf: Added.

client/dashboard/task-list/tasks/shipping/rates.js
joshuatf: Done!

client/dashboard/task-list/tasks/shipping/rates.js
joshuatf: Added a temporary fix for now.  I agree a `PriceInput` component would be a good way to go about this.  I'm thinking we should hold on until we move away from newspack.

joshuatf:
Thanks for the feedback, @justinshreve!

>It seems like the purple was lost on checkboxes when moving the CSS.

Good catch.  Increase selector specificity to fix those colors.

>The currency symbol and "free shipping" text is really hard for me to see.

I was planning on waiting until the color studio update, but it's probably a better idea to go ahead and at least have them visible for now.  Added some of the new temporary color variables until we do this along with a note to remove them.

rrennick:
@joshuatf This one has merge conflicts.

psealock:
If validation fails or there is no value, should we disable the button, or is this skippable?

Screen Shot 2019-08-20 at 2 37 19 PM

joshuatf:
>If validation fails or there is no value, should we disable the button, or is this skippable?

@psealock  This is a little complicated to answer.  Since a user can enter this task even if they already have non-flat-rate shipping options, this would force them to set up flat-rate shipping zones without any option to skip (other than navigating away via the browser of course).

I'm open to other opinions here or if we need more discussion around this, tackling it in a follow-up since this PR is starting to become a blocker.

PR #2762, opened 2019-08-07
only run uninstall functions if dependencies are available

### LINKED ISSUE #2657: Can't delete plugin if WooCommerce is disabled
We got several reports of users not being able to delete the plugin in the forums:

https://wordpress.org/support/topic/cant-delete-it-3/

https://wordpress.org/support/topic/very-bad-cant-delete/

It looks like the issue only happens when WooCommerce is inactive (thanks @panoslyrakis for finding this!).

**To Reproduce**

1. Install WooCommerce and enable it.
2. Install WooCommerce Admin and enable it.
3. Disable WooCommerce.
4. Disable WooCommerce Admin.
5. Delete WooCommerce Admin.

Notice the process never ends.

**Expected behavior**
The plugin should be removed.

rrennick:
Fixes #2657

This PR adds a dependency check to the uninstall script. If WooCommerce is not active, the plugin is uninstalled without clearing the plugin specific data. It is unlikely that a site owner with significant WC data will be removing this plugin while WC is inactive. For the folks test driving WC & WCA this will not leave a significant amount of data behind.

- Build the production version of the plugin
- Install it in a test install
- Activate both WC & WCA
- Deactivate WC (WCA should also deactivate)
- Delete WCA through to plugin screen

### Changelog Note:

Fix: Allow WooCommerce Admin to be deleted through the plugin screen when WooCommerce is not active.

### DIFF
--- a/uninstall.php
+++ b/uninstall.php
@@ -8,11 +8,5 @@
 defined( 'WP_UNINSTALL_PLUGIN' ) || exit;
 
 require_once dirname( __FILE__ ) . '/woocommerce-admin.php';
-
-if ( WC_Admin_Feature_Plugin::instance()->check_dependencies() ) {
-	WC_Admin_Feature_Plugin::instance()->includes();
-} else {
-	require_once dirname( __FILE__ ) . '/includes/class-wc-admin-install.php';
-}
-
+require_once dirname( __FILE__ ) . '/includes/class-wc-admin-install.php';
 WC_Admin_Install::delete_table_data();
--- a/uninstall.php
+++ b/uninstall.php
@@ -11,5 +11,8 @@ require_once dirname( __FILE__ ) . '/woocommerce-admin.php';
 
 if ( WC_Admin_Feature_Plugin::instance()->check_dependencies() ) {
 	WC_Admin_Feature_Plugin::instance()->includes();
-	WC_Admin_Install::delete_table_data();
+} else {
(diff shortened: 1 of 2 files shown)


uninstall.php
```diff
@@ -9,5 +9,7 @@
 
 require_once dirname( __FILE__ ) . '/woocommerce-admin.php';
 
-WC_Admin_Feature_Plugin::instance()->includes();
-WC_Admin_Install::delete_table_data();
+if ( WC_Admin_Feature_Plugin::instance()->check_dependencies() ) {
```
jeffstieler: It looks like `WC_Admin_Install::delete_table_data()` has no dependencies - perhaps we can just include its source file and call it?

rrennick: It does have dependencies:

```
PHP Fatal error:  Uncaught Error: Class 'WC_Admin_Install' not found in /wp-content/plugins/woocommerce-admin/uninstall.php:15
```

rrennick: > It does have dependencies:

Ignore that. I was relying on the autoloader. Addressed in https://github.com/woocommerce/woocommerce-admin/pull/2762/commits/8676faee9e2aeecdd1c054fff31d67af33a08c17

rrennick:
@jeffstieler I removed the dependency check and restored the visibility of `check_dependencies()`.

PR #2763, opened 2019-08-07
Add revenue question to the business details page

### LINKED ISSUE #2743: Onboarding - Store Profiler: New field on Business Details step
We need to add a new field on the Business Details step:

Field: What's your current revenue?
Options:
0$ (I'm just getting started) 
Up to 2.500$ 
2.500$ - 10.000$ 
10.000$ - 50.000$ 
50.000$ - 250.000$ 
More than 250.000$

This field should only show up if the user selects "Yes,..." in the "Currently selling elsewhere?" question. So the order should be:

- How many products will you add?
- Currently selling elsewhere?
- What's your current revenue?
- Which platform is the store using?

Conversation on Slack about this topic: p1564681475033300-slack-wc-onboarding

justinshreve:
Closes #2743.

This PR adds a new revenue question to the business details form. It should only show if the user answers that they are currently selling elsewhere. 

### Screenshots

Screen Shot 2019-08-07 at 2 38 26 PM

Screen Shot 2019-08-07 at 2 38 32 PM

* Reset the profile wizard (go to the WooCommerce orders page, click help, and then setup wizard).
* Visit the business details page `/wp-admin/admin.php?page=wc-admin&step=business-details`.
* Test the revenue question, and the different behaviors depending on what you answered for the already selling question. Verify an error shows if submitted empty.
* You can inspect the API call or analytics call to make sure the answer is saved.

### DIFF
--- a/client/dashboard/profile-wizard/steps/business-details.js
+++ b/client/dashboard/profile-wizard/steps/business-details.js
@@ -21,6 +21,7 @@ import { numberFormat } from '@woocommerce/number';
 import { H, Card, SimpleSelectControl, Form } from '@woocommerce/components';
 import withSelect from 'wc-api/with-select';
 import { recordEvent } from 'lib/tracks';
+import { formatCurrency } from '@woocommerce/currency';
 
 class BusinessDetails extends Component {
 	constructor() {
@@ -30,6 +31,7 @@ class BusinessDetails extends Component {
 			other_platform: '',
 			product_count: '',
 			selling_venues: '',
+			revenue: '',
 			facebook: true,
 			mailchimp: true,
 		};
@@ -42,12 +44,13 @@ class BusinessDetails extends Component {
 
 	async onContinue( values ) {
 		const { createNotice, goToNextStep, isError, updateProfileItems } = this.props;
-		const { facebook, mailchimp, other_platform, product_count, selling_venues } = values;
+		const { facebook, mailchimp, other_platform, product_count, revenue, selling_venues } = values;
 		const businessExtensions = this.getBusinessExtensions( values );
 
 		recordEvent( 'storeprofiler_store_business_details_continue', {
 			product_number: product_count,
 			already_selling: 'no' !== selling_venues,
+			revenue,
 			used_platform: other_platform,
 			install_facebook: facebook,
 			install_mailchimp: mailchimp,
@@ -56,6 +59,7 @@ class BusinessDetails extends Component {
 		await updateProfileItems( {
 			other_platform,
 			product_count,
+			revenue,
 			selling_venues,
 			business_extensions: businessExtensions,
 		} );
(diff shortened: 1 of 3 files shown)


includes/api/class-wc-admin-rest-onboarding-profile-controller.php
```diff
@@ -274,6 +274,21 @@ public static function get_profile_properties() {
 					'brick-mortar-other',
 				),
 			),
+			'revenue'      => array(
+				'type'              => 'string',
+				'description'       => __( 'Current annual revenue of the store.', 'woocommerce-admin' ),
+				'context'           => array( 'view' ),
+				'readonly'          => true,
+				'validate_callback' => 'rest_validate_request_arg',
+				'enum'              => array(
```
joshuatf: One small issue here, but we might need to tackle in a follow-up- this actual value may vary greatly from currency to currency and in many currencies might be a very small amount.

I wonder if there are any packages that can convert to rough approximations/ranges based on currency.

client/dashboard/profile-wizard/steps/business-details.js
```diff
@@ -42,12 +44,13 @@ class BusinessDetails extends Component {
...
 	async onContinue( values ) {
 		const { createNotice, goToNextStep, isError, updateProfileItems } = this.props;
-		const { facebook, mailchimp, other_platform, product_count, selling_venues } = values;
+		const { facebook, mailchimp, other_platform, product_count, revenue, selling_venues } = values;
 		const businessExtensions = this.getBusinessExtensions( values );
 
 		recordEvent( 'storeprofiler_store_business_details_continue', {
 			product_number: product_count,
 			already_selling: 'no' !== selling_venues,
+			revenue,
```
joshuatf: Should we include the currency here as well?  Either as a separate param or part of the revenue range?

client/dashboard/profile-wizard/steps/business-details.js
```diff
@@ -56,6 +59,7 @@ class BusinessDetails extends Component {
 		await updateProfileItems( {
```
joshuatf: I think unrelated to this branch, but I'm getting an invalid param for `other_platform` being an empty string.  Maybe we can add an empty string as an option or filter out empty values.

includes/api/class-wc-admin-rest-onboarding-profile-controller.php
justinshreve: Really good point.

I've logged a follow-up issue so we can think about this a bit.

https://github.com/woocommerce/woocommerce-admin/issues/2768

client/dashboard/profile-wizard/steps/business-details.js
justinshreve: Fixed in https://github.com/woocommerce/woocommerce-admin/pull/2763/commits/19ae63deb6c634599d4b28df76c19cc67c876543.

client/dashboard/profile-wizard/steps/business-details.js
justinshreve: Fixed in https://github.com/woocommerce/woocommerce-admin/pull/2763/commits/19ae63deb6c634599d4b28df76c19cc67c876543.

PR #2784, opened 2019-08-13
Allow action scheduler custom data store migration

### LINKED ISSUE #2766: Allow custom table migration with Action Scheduler 3.0.0
Action Scheduler 3.0 is introducing custom tables for the default data store. By default, Action Scheduler won't migrate to the new custom table data store if a site is using a store other than the default `ActionScheduler_wpPostStore` class. This prevents overriding custom data stores implemented on the site.

WooCommerce Admin adds the `WC_Admin_ActionScheduler_WPPostStore`. Because of this, Action Scheduler won't migrate to custom tables on any site running WooCommerce Admin.

While this can be overridden returning `true` on the `'action_scheduler_migrate_data_store'` filter. That would also migrate sites using a different data store and also running WC Admin, so that's not the best solution.

Instead, I propose only applying the `WC_Admin_ActionScheduler_WPPostStore` when Action Scheduler is < version 3.0.0.

I had a look at the methods in `WC_Admin_ActionScheduler_WPPostStore`, and they appear to:

1. add a priority to WC Admin actions - this shouldn't really be necessary now that 3.0.0 has an async runner, which will increase default action throughput by at least 6x, likely much more.
2. a method to cancel/trash pending WC Admin actions - I suspect this is being done for better performance than using `self::queue()->cancel_all()` / `as_unschedule_all_actions()` which cancels each action one-by-one. That could be addressed in Action Scheduler core itself by introducing a new API to bulk cancel if needed. Custom tables should be much more performant than the existing store for deleting actions, so this may be less needed also.

/cc @rrennick

rrennick:
Fixes #2766

This PR is a companion to https://github.com/Prospress/action-scheduler/pull/351 which adds new bulk cancel functions to AS 3.0.0. It updates the cancel pending WC Admin actions to use the functions in AS 3.0.0.

- You may want to make a DB backup to restore after testing
- Checkout https://github.com/Prospress/action-scheduler/tree/fix/350
- Migrate to custom tables with `wp action-scheduler migrate`
- install and activate https://github.com/Prospress/action-scheduler-disable-default-runner to prevent the cron runner from processing actions
- Go to Plugins -> Active
- In a second tab go to Tools -> Scheduled Actions
- In a third tab go to Analytics -> Settings
- Uncheck the skip already imported
- Click start
- Use `wp action-scheduler run` until you receive `Completed processing action 12345 with hook: wc-admin_orders_lookup_import_batch_init`
- Switch to the Plugins tab and deactivate WC Admin
- Switch to the Scheduled Actions tab and click the `Canceled` filter
- Go to the last page of canceled actions
- You should see import actions canceled at the (UTC) time WC Admin was deactivated

### Screenshots
Screen Shot 2019-08-16 at 1 18 38 PM

### Changelog Note:

Should not need one.

### DIFF
--- a/src/FeaturePlugin.php
+++ b/src/FeaturePlugin.php
@@ -165,6 +165,11 @@ class FeaturePlugin {
 			return $store_class;
 		}
 
+		// Don't override if action scheduler is 3.0.0 or greater.
+		if ( version_compare( \ActionScheduler_Versions::instance()->latest_version(), '3.0', '>=' ) ) {
+			return $store_class;
+		}
+
 		return 'Automattic\WooCommerce\Admin\Overrides\WPPostStore';
 	}
 
--- a/src/ReportsSync.php
+++ b/src/ReportsSync.php
@@ -230,6 +230,8 @@ class ReportsSync {
 				self::SINGLE_ORDER_IMPORT_ACTION,
 			);
 			$store->clear_pending_wcadmin_actions( $action_types );
+		} elseif ( version_compare( \ActionScheduler_Versions::instance()->latest_version(), '3.0', '>=' ) ) {
+			$store->cancel_actions_by_group( self::QUEUE_GROUP );
 		} else {
 			self::queue()->cancel_all( null, array(), self::QUEUE_GROUP );
 		}


rrennick:
> We'll have to update the WP post store test when AS 3.0.0 gets merged into WC core, right?

The test could be removed when AS 3.0.0 is merged.

PR #2799, opened 2019-08-15
scroll import action to import section of settings page

### LINKED ISSUE #2243: Historical Data get started takes you to Analytics settings
**Describe the bug**
This is a small issue. The original rebuild report data section was at the top of the settings screen. The new import feature is at the bottom. It would be better if the get started historical data CTA button took the user to import data section vs the top of the screen.

**To Reproduce**

1. Re-enable the historical data note
2. Click on 'Get Started'

**Expected behavior**
The import historical data section should be visible.

rrennick:
Fixes #2243

This PR adds an import query arg to the Import Historical Data note. When the query arg is present the settings screen scrolls to the Import section.

It introduces a `<ScrollTo> ... </ScrollTo>` component with an offset attribute that can be optionally used as an adjustment to the top of the component. The top of the component is the default scroll to position.

- Re-create the `get-started` note using the code in this PR.
- Open the Activity panel & Notes.
- Click on the import historical data `Get Started`
- The setting screen should scroll to the import section

### Changelog Note:

Tweak: have Import note action scroll to import section of settings screen.


client/analytics/settings/index.js
```diff
@@ -39,12 +39,15 @@ class Settings extends Component {
...
 
+		this.importRef = createRef();
 		this.handleInputChange = this.handleInputChange.bind( this );
 		this.warnIfUnsavedChanges = this.warnIfUnsavedChanges.bind( this );
+		this.scrollToImport = this.scrollToImport.bind( this );
 	}
 
 	componentDidMount() {
 		window.addEventListener( 'beforeunload', this.warnIfUnsavedChanges );
+		setTimeout( this.scrollToImport, 250 );
```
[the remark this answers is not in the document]
rrennick: Good catch. I added a check to allow for longer page load times.

rrennick:
> I think it would be good to also make this scroll method reusable to other parts of wc-admin. 

Great idea. I added the `<ScrollTo />` component.

rrennick:
This PR needs #2802 merged so the component can be added to the component change log.

packages/components/src/scroll-to/index.js
```diff
@@ -0,0 +1,33 @@
...
+			window.scrollTo( 0, this.ref.current.offsetTop );
+		} else {
+			setTimeout( this.scrollTo, 250 );
+		}
+	}
+
+	render() {
+		this.ref = createRef();
+		return (
+			<span ref={ this.ref }></span>
```
psealock: You'll want to add some buffer to take into account the navbar. If you shrink the window veritcally, you can see the top of the section hidden under the navbar.

Screen Shot 2019-08-23 at 12 49 35 PM

rrennick: @psealock Thanks, good catch. I added an offset parameter in https://github.com/woocommerce/woocommerce-admin/pull/2799/commits/f0e7b2e80c20239b3f49b69b2768c15fbddea73b

rrennick:
In https://github.com/woocommerce/woocommerce-admin/pull/2799/commits/f0e7b2e80c20239b3f49b69b2768c15fbddea73b, I changed the component so it could contain children. 

With this change it could be added to the DevDocs because it can contain visible content. Could I add the DevDocs in a follow-up PR?

PR #2801, opened 2019-08-15
Add empty state to tables

### LINKED ISSUE #2613: Table: Update empty table state
When no table data exists, update the Table view. Below are additional features also included from the original issue https://github.com/Automattic/woocommerce-payments/issues/97

### Current State

Screen Shot 2019-07-09 at 1 02 46 PM

>I added an empty state to the transactions page in figma

>https://www.figma.com/file/cGBqV73qDcMXwRDI5JSy4nH8/Merchant-Dashboard?node-id=298%3A1031

>The row height with the message is the same as a normal table row: 60px

>The text styles for the empty table message should match the empty message style in the chart component:
https://github.com/woocommerce/woocommerce-admin/blob/master/packages/components/src/chart/d3chart/style.scss#L19

>When a table is empty we're not being very consistent with what remains visible in the table header - I'd like to update it so that everything that's visible when table data is present is still visible in the empty state, it's just disabled. The exception to that rule being the kebab menu that allows users to toggle columns on and off. That should still be functional even when there's no data.

>**Header elements that are disabled when the table is empty:** 

>Compare button (when applicable) 
>Search input (when applicable) 
>Download Button

>**Header elements that are always functional:** 

>Kebab that controls which columns are visible

jeffstieler:
Fixes #2613.

This PR seeks to add an empty state to Tables.

In the empty state:
* A message is displayed (same as the chart)
* All table controls are disabled/noop-ed except for column visibility

### Screenshots

Screen Shot 2019-08-15 at 12 59 59 PM

- Open a report
- Select a date range that has no data
- Verify the "no data" message in the table
- Verify the columns can still have their visiblity toggled
- Verify that all report table inputs are disabled (compare, search)

### Changelog Note:

Tweak: add empty dataset treatment for report tables.

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -6,6 +6,8 @@
 - Added a new `<SimpleSelectControl />` component.
 - Added a new `<WebPreview />` component.
 - SearchListItem component: fix long count values being cut-off in IE11.
+- Add `disabled` prop to CompareButton, Search, and TableCard components.
+- Table component: add empty table display.
 
 # 3.1.0
 - Added support for a `countLabel` prop on `SearchListItem` to allow custom counts.
--- a/packages/components/src/filters/compare/button.js
+++ b/packages/components/src/filters/compare/button.js
@@ -12,8 +12,8 @@ import { Button, Tooltip } from '@wordpress/components';
  *
  * @return { object } -
  */
-const CompareButton = ( { className, count, children, helpText, onClick } ) =>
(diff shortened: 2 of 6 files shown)


jeffstieler:
> Unrelated to this PR, but noticed while testing that my woocommerce-ellipsis-menu__toggle is slightly cut off. Do you see this?

Yes I do - I made sure it was still occurring on `master` and then promptly forgot about it 😅.

Issue #2719, opened 2019-07-30
Product sales graph does not match orders count/items sold


stephhjacq:
**Describe the bug**
Product sales graph does not match orders count/items sold.

**To Reproduce**

For this case, the storeholder needs the flexibility to change a simple product to a variable product while tracking historic sales.

1. Look at report for simple product, SKU 987, sales month to date under Analytics > Product
2. Edit product, making it variable. Use SKU 987 for one of the variations
3. Purchase variation 987
4. View report for product by title, sales month to date. Totals are correct, but graph no longer shows the historic purchase

**Expected behavior**
Graph should match data.

**Screenshots**
Simple product: Sticker (SKU 987), before

![https://cld.wthms.co/q5xBiC](https://cld.wthms.co/q5xBiC+)
Full Image: https://cld.wthms.co/q5xBiC+

Variable product: Sticker - Laptop variation (SKU 987), after

![https://cld.wthms.co/gwqfYA](https://cld.wthms.co/gwqfYA+)
Full Image: https://cld.wthms.co/gwqfYA+

**Desktop (please complete the following information):**

* OS: Mac OS 10.14.5
* Browser Chrome Version 75.0.3770.142 (Official Build) (64-bit)

psealock:
Thanks for the bug report @stephhjacq.

I'd to clarify one thing to make sure I understand correctly. In the second screenshot `Variable product: Sticker - Laptop variation (SKU 987), after`, you'd like to see the previous order as well. In other words, "Items Sold" is correct with a number of 2, but the chart and table don't show the original purchase. Is that right?

This is an interesting question on how to handle changed products and historical data. At the very least, there should be cohesion between the summary numbers and chart/table.

nagpai:
>I'd to clarify one thing to make sure I understand correctly. In the second screenshot Variable product: Sticker - Laptop variation (SKU 987), after, you'd like to see the previous order as well. In other words, "Items Sold" is correct with a number of 2, but the chart and table don't show the original purchase. Is that right?

I think what the user needs is the orders placed for the `Simple product - Sticker`  before it got converted to `Variable product - Sticker - Laptop variation`

It seems that the system treats the variable product independent of the legacy simple product. One work around for now is to clone the simple product (change it to draft) and make variants from the cloned product.

JScott-MSI:
I am the original reporter of this issue with support. I have some additional information that might clarify the issue that I am seeing.

[...]

PR #2806, opened 2019-08-16
Fix bug with checking user preferences for report column visibility.

jeffstieler:
_(Found while reproducing https://github.com/woocommerce/woocommerce-admin/issues/2719)_

Fixes a bug introduced in https://github.com/woocommerce/woocommerce-admin/commit/7ecfb1d4639757f849a4fe329314e35e13fac44d#diff-cbe713361af618d887a218e657fcc463R53.

When there are no user preferences for column visibility on a report, the value is an empty string. Since we're storing the hidden columns, if a user chooses to show all columns, we store an empty array.

The above linked commit introduced a `.length` check which caused "don't hide anything" preferences to revert to the column default visibility.

The easiest way to see this in the Products report - SKU column.

This PR fixes the issue by removing the `.length` check and introducing a new `visible` object property rather than overloading the `hiddenByDefault` property.

- On `master`
- Go to Analytics > Products
- Toggle all columns to be visible
- Refresh page
- Verify the SKU column is hidden upon refresh
- Check out this branch
- Refresh
- Verify SKU column is visible

### Changelog Note:

Fix: report column visibility preference bug.

### DIFF
--- a/client/analytics/components/report-table/index.js
+++ b/client/analytics/components/report-table/index.js
@@ -108,14 +108,19 @@ class ReportTable extends Component {
 	}
 
 	filterShownHeaders( headers, hiddenKeys ) {
-		if ( ! hiddenKeys || ! hiddenKeys.length ) {
-			return headers;
+		// If no user preferences, set visibilty based on column default.
+		if ( ! hiddenKeys ) {
+			return headers.map( header => ( {
+				...header,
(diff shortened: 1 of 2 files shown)


dechov:
Just noticed this change – FYI looks like the `propTypes` declaration should be changed to `visible` as well.

PR #2808, opened 2019-08-19
Add Autocomplete component

### LINKED ISSUE #2527: Implement Autocomplete Component

joshuatf:
Fixes #2527 

Adds an autocomplete component with filterable options.

There are a few areas this does not cover and will be handled in a follow-up since this PR is already rather large:
* Swapping out the old `Autocomplete` component inside the `Search` component.
* Swapping out existing dropdowns for items that will benefit from this component (e.g., the country selector).
* Implementing "matching" to bold/highlight matching portions of search queries.

### Questions

* Design-wise this needs a look over.  Should we use transitions for label movements?  Is the menu highlighting contrasted enough? The label/placeholder hiding is a bit different from existing newspack components (I believe they were incorrect and this component more closely mirrors those seen in Material).
* Should this component offer an option to hide the search field, allowing it to become a styled select component? This may consolidate efforts between various components.  If "yes," should this be renamed to `Select` or something similar?

### Accessibility

### Screenshots
Screen Shot 2019-08-19 at 2 12 40 PM

1.  Go to the Autocomplete component in the Devdocs.
2.  Test all 3 examples and make sure they work as expected and are styled as expected.
3.  Test with a keyboard and screen reader.  Make sure the component is a11y friendly.

Here are some references to check against in terms of accessibility standards (note this uses ARIA 1.0 as it seems this still has more support than 1.1 and there aren't plans to deprecate):
* https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.0pattern/combobox-autocomplete-both.html
* https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html
* https://github.com/WordPress/gutenberg/issues/16473#issuecomment-520560175

### DIFF
--- a/packages/components/src/autocomplete/control.js
+++ b/packages/components/src/autocomplete/control.js
@@ -32,14 +32,14 @@ class SearchControl extends Component {
 
 	updateSearch( onSearch ) {
 		return event => {
-			onSearch( event );
+			onSearch( event.target.value );
 		};
 	}
 
 	onFocus( onSearch ) {
 		return event => {
 			this.setState( { isActive: true } );
-			onSearch( event );
+			onSearch( event.target.value );
 		};
 	}
 
@@ -89,11 +89,6 @@ class SearchControl extends Component {
 			aria-activedescendant={ activeId }
 			aria-describedby={
 				hasTags && inlineTags ? `search-inline-input-${ instanceId }` : null
-			}
-			size={ inlineTags
-				? ( ( query.length === 0 && placeholder && placeholder.length ) ||
-					query.length ) + 1
-				: null
 			}
 		/>;
 	}
--- a/packages/components/src/autocomplete/control.js
+++ b/packages/components/src/autocomplete/control.js
@@ -3,6 +3,7 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
+import { BACKSPACE } from '@wordpress/keycodes';
 import { Component, createRef } from '@wordpress/element';
 import classnames from 'classnames';
 import PropTypes from 'prop-types';
@@ -50,7 +51,7 @@ class SearchControl extends Component {
 	onKeyDown( event ) {
 		const { selected, onChange, query } = this.props;
 
-		if ( 8 === event.keyCode && ! query && selected.length ) {
+		if ( BACKSPACE === event.keyCode && ! query && selected.length ) {
 			onChange( [ ...selected.slice( 0, -1 ) ] );
 		}
 	}
--- a/packages/components/src/autocomplete/control.js
+++ b/packages/components/src/autocomplete/control.js
@@ -37,10 +37,10 @@ class SearchControl extends Component {
 		};
 	}
 
-	onFocus( onChange ) {
+	onFocus( onSearch ) {
 		return event => {
 			this.setState( { isActive: true } );
-			onChange( event );
+			onSearch( event );
 		};
 	}
 
--- a/packages/components/src/autocomplete/control.js
+++ b/packages/components/src/autocomplete/control.js
@@ -37,11 +37,6 @@ class SearchControl extends Component {
 		};
 	}
 
-	shouldRenderTags() {
-		const { selected } = this.props;
-		return selected.some( item => Boolean( item.label ) );
-	}
-
 	onFocus( onChange ) {
 		return event => {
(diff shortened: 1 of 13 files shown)


packages/components/src/autocomplete/control.js
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	onBlur() {
+		this.setState( { isActive: false } );
+	}
+
+	onKeyDown( event ) {
+		const { selected, onChange, query } = this.props;
+
+		if ( 8 === event.keyCode && ! query && selected.length ) {
```
justinshreve: <s>Minor: Can you add a comment that 8 is the keyCode for backspace?</s>

Can we use the `BACKSPACE` constant from `@wordpress/keycodes` here?

packages/components/src/autocomplete/example.md
```diff
@@ -0,0 +1,87 @@
...
+	},
+    {
+        key: 'apricot',
+		label: 'Apricot',
+		value: { id: 'apricot' },
+	},
+	{
+        key: 'banana',
+		label: 'Banana',
+		keywords: ['best', 'fruit'],
```
justinshreve: 💯

packages/components/src/autocomplete/test/index.js
```diff
@@ -0,0 +1,161 @@
...
+ */
+import { mount } from 'enzyme';
+import { Button } from '@wordpress/components';
+
+/**
+ * Internal dependencies
+ */
+import { Autocomplete } from '../index';
+
+describe( 'Autocomplete', () => {
```
justinshreve: Great tests here, thanks for including them! 👍

joshuatf:
Thanks for the review @rrennick and @justinshreve 😄 

>The only thing I found is that it needs docs/packages/components/autocomplete.md for the single component doc page to work.

Added these and other missing doc files.

>I think the label should be above the tags and text, like it would be in a text input.

I think you're right.  One small side effect from doing this means tags need to be reduced in height slightly:

Screen Shot 2019-08-20 at 9 56 46 AM

Also applied other feedback and gave this PR a rebase.

rrennick:
@joshuatf Looks great for me.

packages/components/src/autocomplete/index.js
```diff
@@ -0,0 +1,336 @@
...
+		}
+	}
+
+	getFilteredOptions( query ) {
+		const { excludeSelectedOptions, getSearchExpression, maxResults, onFilter, options, selected } = this.props;
+		const selectedKeys = selected.map( option => option.key );
+		const filtered = [];
+
+		// Create a regular expression to filter the options.
+		const expression = getSearchExpression( escapeRegExp( query.trim() ) );
```
psealock: On inline tags with the options always visible, pressing the escape key results in an error `index.js:126 Uncaught TypeError: Cannot read property 'trim' of null`.

Tab should also allow me to move on, but using only the keyboard you can get stuck. See the other comment.

packages/components/src/autocomplete/list.js
```diff
@@ -0,0 +1,211 @@
...
+				nextSelectedIndex = null !== selectedIndex
+					? ( selectedIndex === 0 ? filteredOptions.length : selectedIndex ) - 1
+					: filteredOptions.length - 1;
+				onChange( nextSelectedIndex );
+				this.scrollToOption( nextSelectedIndex );
+				event.preventDefault();
+				event.stopPropagation();
+				break;
+
+			case TAB:
```
psealock: The tab functionality in https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html brings focus to the next focusable element even when the options are visible.

Here, tab brings me to the next focusable element only when the options are hidden. In the case of the first example in dev docs, if you navigate to it using tab, you can't tab out of it unless something is selected.

PR #2827, opened 2019-08-21
use wc-admin urls in devdocs

### LINKED ISSUE #2813: Single DevDoc page breaks address bar URL
**Describe the bug**
When you load any DevDoc detail page for a single component the address bar is updated to a broken link (eg. `https://ron.localhost/wp-admin/https://ron.localhost/wp-admin/?page=wc-admin&path=/devdocs/calendar`)

**To Reproduce**

1. Go to WooCommerce -> DevDocs
2. Click on any component
3. See address bar
4. Do a browser refresh

**Expected behavior**
Browser refresh should work on individual DevDoc pages.

rrennick:
Fixes #2813

This PR changes the URLs used in the DevDocs page to use wc-admin URLs instead of fully qualified dashboard URLs.

- Ensure navigation among regular dashboard pages, WC Admin, DevDocs main & component pages works without issue.

### Changelog Note:

Fix: Bug navigating from DevDoc component pages to WP dashboard pages.

### DIFF
--- a/client/devdocs/index.js
+++ b/client/devdocs/index.js
@@ -12,7 +12,6 @@ import { find, get } from 'lodash';
 import ComponentExample from './example';
 import ComponentDocs from './docs';
 import { Card, Link } from '@woocommerce/components';
-import { getAdminLink } from '@woocommerce/navigation';
 import examples from './examples.json';
 import './style.scss';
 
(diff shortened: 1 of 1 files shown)


client/devdocs/index.js
```diff
@@ -63,15 +62,13 @@ export default class extends Component {
 								component ? (
 									componentName
 								) : (
-									<Link href={ getAdminLink( `?page=wc-admin&path=/devdocs/${ filePath }` ) }>
+									<Link href={ `admin.php?page=wc-admin&path=/devdocs/${ filePath }` }>
```
jeffstieler: What changed in `getAdminLink()` to make it unsuitable for this usage?

rrennick: In the `<Link />` component an `onClick` handler is added if the link type is `wc-admin`. When the link is clicked it appends the link href to the end of the base wc-admin URL. By using the function the onClick is using full URI.

The alternative change that would work is changing the link type in the getAdminLink() call.

rrennick: https://github.com/woocommerce/woocommerce-admin/blob/master/packages/components/src/link/index.js#L51:L53

jeffstieler: Oh, it looks like the `<Link>` component [internally calls `getAdminLink()`](https://github.com/woocommerce/woocommerce-admin/blob/master/packages/components/src/link/index.js#L41). 👍

rrennick:
> Optionally explicitly pass the type="wc-admin" to the <Link> component

@jeffstieler thanks for the suggestion. I added these in https://github.com/woocommerce/woocommerce-admin/pull/2827/commits/006561598975340f00b319497deb4dc899143dc1 .

PR #2832, opened 2019-08-22
Queue customer lookup updates.

### LINKED ISSUE #2702: Customer Data Store: enqueue job instead of adding more sql operations to checkout flow
Currently in the `WC_Admin_Reports_Customers_Data_Store` we are [hooking into a few actions](https://github.com/woocommerce/woocommerce-admin/blob/master/includes/data-stores/class-wc-admin-reports-customers-data-store.php#L72) that are called during the checkout flow:

- `woocommerce_new_customer` -> `WC_Admin_Reports_Customers_Data_Store::update_registered_customer`
- `woocommerce_update_customer` -> `WC_Admin_Reports_Customers_Data_Store::update_registered_customer`
- `updated_user_meta` -> `WC_Admin_Reports_Customers_Data_Store::update_registered_customer_via_last_active`

`update_registered_customer()` subsequently fires off a number of `SELECT`s and potentially an `INSERT` if we are working with a new customer. 

While I haven't done an exact count, nor any benchmarks on these operations, but I'm wondering if it might be better for us to enqueue a job to perform these actions akin to how we do for other new order events. Unless it is critical that we perform these operations real-time during the checkout flow, it seems like we should limit the number of database operations we are injecting into the checkout flow because speed of the checkout process is such a critical metric for stores.

Perhaps we could even chain the customer logic to the [WC_Admin_Reports_Orders_Stats_Data_Store::sync_order](https://github.com/woocommerce/woocommerce-admin/blob/master/includes/data-stores/class-wc-admin-reports-orders-stats-data-store.php#L419) that is already happening on all new orders. This would allow us to leverage the job that is already being enqueued even.

Again speed of the checkout process is critical to conversions. Avoiding adding more sql operations during the checkout flow seems like a good reason to prioritize this.

/cc @joshuatf for any thoughts and @psealock since this is what we talked about today during our 1:1

jeffstieler:
Fixes #2702.

This PR introduces a `SINGLE_CUSTOMER_IMPORT_ACTION` to update the customer lookup table using Action Scheduler.

- Use "Action Scheduler - Disable Default Queue Runner" plugin to see queued jobs before they're executed
- Place an order or create a WP user
- Verify a `wc-admin_import_customer` action is pending
- Verify that when run, the customer lookup table is updated

### Changelog Note:

Tweak: schedule customer lookup table updates instead of running during checkout.

### DIFF
--- a/src/API/Reports/Customers/DataStore.php
+++ b/src/API/Reports/Customers/DataStore.php
@@ -75,8 +75,6 @@ class DataStore extends ReportsDataStore implements DataStoreInterface {
 	 * Set up all the hooks for maintaining and populating table data.
 	 */
 	public static function init() {
-		add_action( 'woocommerce_new_customer', array( __CLASS__, 'update_registered_customer' ) );
-		add_action( 'woocommerce_update_customer', array( __CLASS__, 'update_registered_customer' ) );
 		add_action( 'edit_user_profile_update', array( __CLASS__, 'update_registered_customer' ) );
 		add_action( 'updated_user_meta', array( __CLASS__, 'update_registered_customer_via_last_active' ), 10, 3 );
 	}
--- a/src/ReportsSync.php
+++ b/src/ReportsSync.php
@@ -65,7 +65,12 @@ class ReportsSync {
 	const ORDERS_DELETE_BATCH_ACTION = 'wc-admin_delete_orders_batch';
 
 	/**
-	 * Action hook for importing a batch of orders.
+	 * Action hook for importing a single customer.
+	 */
+	const SINGLE_CUSTOMER_IMPORT_ACTION = 'wc-admin_import_customer';
+
+	/**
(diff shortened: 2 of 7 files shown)


jeffstieler:
> When I create a guest checkout by buying something on incognito mode, I'm not seeing the `wc-admin_import_customer` pending action. However, I see the `wc-admin_import_order` action pending and when I run it, the new customer is created. I'm not sure whats happening, but its the desired result.

That should be the correct behavior (and shouldn't be different from `master`) - WC only creates a new customer if guest checkout is disabled or if the guest checks the "create an account" checkbox. The customer is created for WC-Admin in the order sync because it calls [`Reports\Customers\DataStore::get_or_create_customer_from_order()`](https://github.com/woocommerce/woocommerce-admin/blob/268e968b9ee7f49656e51389f18e854bca2b34f3/src/API/Reports/Customers/DataStore.php#L485).

PR #2837, opened 2019-08-23
Update to Color Studio v2.1.0

### LINKED ISSUE #2574: Update color-studio to v2 and update color palette

justinshreve:
Partially Fixes #2574.

This PR updates the color-studio package to v2.1.0. We were on 1.0.5 before so this contained breaking changes (all the variable names changed). The colors and schemes have also been iterated on as well.

I originally was going to do this just for Onboarding, but I went ahead and applied these to the general components, analytics and dashboard sections as well so we didn't have to keep different color schemes for different parts of the same plugin. So we might want a quick design peak at this (cc @LevinMedia ).

Please note that Calypso changes will be done in a separate PR.

### Screenshots

Screen Shot 2019-08-26 at 1 45 27 PM

Screen Shot 2019-08-26 at 1 44 41 PM

Screen Shot 2019-08-23 at 11 47 59 AM

* Close out of your current npm process, and run `npm install`.
* Make sure the application boots up and you do not receive any undefined variable errors when building the CSS.
* Click around various parts of the WooCommerce Admin experience. See the new colors and make sure everything looks OK.
* In the case of onboarding, you can compare to the Figma files.

### Changelog Note:

Updated to color-studio v2 and refreshed the color scheme


LevinMedia:
Awesome! Seeing the new colors in WC-Admin

I'm thinking we should start moving WC-Admin to match the Color Studio colors being used in the onboarding IE: 

primary buttons (including the split button in the date range picker) and text hyperlinks are using Pink 50 #c9356e 

Any interactive toggles, radio buttons or checkboxes (with the exception of what's in charts) as well as the highlights on top of the summary numbers are using Woo Purple 60 #674399

cc @jameskoster

justinshreve:
> I'm thinking we should start moving WC-Admin to match the Color Studio colors being used in the onboarding IE:

> primary buttons (including the split button in the date range picker) and text hyperlinks are using > Pink 50 #c9356e

> ...

Sounds good to me. It makes sense to do that while we are still in the plugin stage. I can handle doing that Monday in this PR.

PR #2843, opened 2019-08-26
Add filters to all Report filter config objects.

### LINKED ISSUE #2803: Advanced Filters: Allow extensions to add filters
Report configs with Advanced Filters should apply a filter so that extensions can add a config object.

Some documentation on how to configure an object would be required too.

jeffstieler:
Fixes #2803. (Most of it)

This PR adds `applyFilters()` calls to all report filter config objects.

Additionally, in order to improve documentation, the `AdvancedFilters`, `CompareFilter`, `DateRangeFilterPicker`, and `FilterPicker` components have been moved into their own top-level component directories so they have bespoke DevDocs pages.

Inline documentation was not added to the `applyFilters()`, keeping with the precedent set by the Gutenberg team.

I think we should further improve the DevDocs by restructuring to match Calypso. Here's an example: https://github.com/Automattic/wp-calypso/tree/master/client/components/line-chart

- Verify that existing report filters are intact
- Optionally add some code that filters a reports filters

### Changelog Note:

Dev: add ability to modify all report filter controls.
Tweak: change report charts filter name.

### DIFF
--- a/client/analytics/report/categories/config.js
+++ b/client/analytics/report/categories/config.js
@@ -10,12 +10,12 @@ import { applyFilters } from '@wordpress/hooks';
  */
 import { getCategoryLabels } from 'lib/async-requests';
 
-const CATEGORY_REPORT_CHART_FILTER = 'woocommerce_admin_category_report_chart_filter';
+const CATEGORY_REPORT_CHARTS_FILTER = 'woocommerce_admin_category_report_charts';
 const CATEGORY_REPORT_FILTERS_FILTER = 'woocommerce_admin_category_report_filters';
 const CATEGORY_REPORT_ADVANCED_FILTERS_FILTER =
 	'woocommerce_admin_category_report_advanced_filters';
 
-export const charts = applyFilters( CATEGORY_REPORT_CHART_FILTER, [
+export const charts = applyFilters( CATEGORY_REPORT_CHARTS_FILTER, [
 	{
 		key: 'items_sold',
 		label: __( 'Items Sold', 'woocommerce-admin' ),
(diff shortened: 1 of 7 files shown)


packages/components/src/advanced-filters/example.md
```diff
@@ -0,0 +1,140 @@
+```jsx
+import { AdvancedFilters } from '@woocommerce/components';
+const { orderStatuses } = wcSettings;
+
+const path = '';
+const query = {};
```
psealock: If you place `path: '/devdocs'` in the query, the devDocs example will actually update the value when you click "filter", otherwise it goes to Dashboard.

packages/components/src/compare-filter/example.md
```diff
@@ -0,0 +1,23 @@
+```jsx
+import { CompareFilter } from '@woocommerce/components';
+
+const path = '';
+const query = {};
```
psealock: same comment about the path.

packages/components/src/filter-picker/example.md
```diff
@@ -0,0 +1,46 @@
+```jsx
+import { FilterPicker } from '@woocommerce/components';
+
+const path = '';
+const query = {
+    meal: 'breakfast',
```
psealock: I think here too.

psealock:
>Should all reports also have the ability to have <ReportSummary>s and <ReportChart>s filtered in?

That would be customers and stock. I could see the potential for that, but yes lets punt on that because it also seems low priority.

>what do you think about renaming the chart filters to remove the word "filter" from the string? 

yeah, absolutely. Makes sense to do this pre-merge. We should probably make a note of it in the release notes.

PR #2846, opened 2019-08-27
Fix conflict with Blocks 2.4

mikejolley:
See https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/916/files
Fixes https://github.com/woocommerce/woocommerce-admin/issues/2938

WC-Admin calls `get_current_screen` on non-admin screens. These checks protect against that happening.

You can test against Blocks 2.4. Before this patch, viewing the frontend of the store shows the following fatal:
slack-imgs

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -519,6 +519,10 @@ class Loader {
 	 * @return array Array of component settings.
 	 */
 	public static function add_component_settings( $settings ) {
+		if ( ! is_admin() ) {
+			return $settings;
+		}
+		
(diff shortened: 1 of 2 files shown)


psealock:
Since blocks 2.4 is out, we'll probably want to make an 0.18.1 release

mikejolley:
It happens when certain blocks are present only which causes settings to load, and WC admin to hook in.

psealock:
@rrennick I rebased off master and tested. This solves the problem for me, which appears to be caused by the list block. Can you give this another quick look before we merge?

Issue #2684, opened 2019-07-22
Onboarding - Task List: Set up payments


pmcpinto:
**User Story**
As a new merchant
I want some guidance in terms of setting up payments
So I can set them up as quickly as possible

**Acceptance Criteria**
- The payments' methods that the merchant can install depend on the store location ([screen](https://www.figma.com/file/xyKPRZ73kKhDyNjwbFvBJR/Task-List?node-id=1041%3A5976))
- If the store location is unkown the merchant only sees the Stripe and Paypal methods.
- If the merchant selected in the [store profiler](https://www.figma.com/file/b0b2CrxPZz0i1aAPuv985tvK/Store-Profiler?node-id=1150%3A0) that he sells on brick and mortar stores, he can also install [Square](https://www.figma.com/file/xyKPRZ73kKhDyNjwbFvBJR/Task-List?node-id=1041%3A6890)
- If all the payment methods are toggled off, the Proceed button is disabled
- If the merchant has Jetpack and WC Services connected, he can also automatically create an account at the payment provider ([screen](https://www.figma.com/file/xyKPRZ73kKhDyNjwbFvBJR/Task-List?node-id=1001%3A146))
- If the merchant chooses to install the payment methods, he can see a [screen](https://www.figma.com/file/xyKPRZ73kKhDyNjwbFvBJR/Task-List?node-id=994%3A1281) informing him about the installation during the install
- If the merchant opted to automatically create a Stripe account and it already exists an account with that email, he sees this [screen](https://www.figma.com/file/xyKPRZ73kKhDyNjwbFvBJR/Task-List?node-id=1190%3A156) that is going to redirect him to the Stripe sign in flow
- If the merchant opted to automatically create a Stripe account and it doesn't exists an account with that email, it should be possible to authenticate and get the API keys to automatically complete the Stripe connection and forward the user to this [screen](https://www.figma.com/file/xyKPRZ73kKhDyNjwbFvBJR/Task-List?node-id=999%3A125)
- If the merchant (w/ Jetpack and WC Services connected) installed Stripe but opted out to auto create a Stripe account, he should see this [screen](https://www.figma.com/file/xyKPRZ73kKhDyNjwbFvBJR/Task-List?node-id=1001%3A2015) after install. If the clicks Connect he should be redirected to the [Stripe flow](https://www.figma.com/file/xyKPRZ73kKhDyNjwbFvBJR/Task-List?node-id=1011%3A848)
[...]

PR #2853, opened 2019-08-28
Add base payments task & jetpack connection status method

### LINKED ISSUE #2835: Jetpack connection methods
In order to test and get Jetpack connection through various onboarding steps, we need to be able to retrieve:

* Jetpack connection status
* Jetpack connection URL

This could be done via the [Jetpack REST API](https://github.com/Automattic/jetpack/blob/master/docs/rest-api.md), however since the URL is static and the connection status seems unlikely to change without navigating away from wc-admin (sanity check? /cc @justinshreve ), we can probably just grab both of these server side and store them in `wcSettings.onboarding`.

Once both of these variables are accessible, we can apply them and potentially skip the connection step in:
* Task list: Tax
* Task list: Shipping
* Profiler: Plugins

justinshreve:
Part of #2684.
Fixes #2835.

This PR adds part of the functionality for the payment methods task. It displays step 1, which is a list of payment methods that you can configure.

This PR also adds a method for getting the Jetpack Connection Status.

* Stripe & PayPal will always be shown
* Square will be shown if in a supported country, and one of the brick and motor options were selected in the profiler.
* Klarna Checkout will be shown for Sweden, Finland, Norway, and the Netherlands.
* Klarna Payments  will be shown for Denmark, Germany, and Austria.

If WooCommerce Services and Jetpack are installed and connected, an additional checkbox and input are displayed since we can streamline their setup.

### Screenshots

Screen Shot 2019-08-28 at 4 40 48 PM

* Edit `dashboard/index.js` and make sure `requiredTasksComplete` is set to `false`.
* Go to the WooCommerce Admin dashboard.
* Click "Set up Payments"
* Test the various setting combinations mentioned above. Verify an unconnected Jetpack doesn't show the email checkbox and input. Verify Klarna shows up for the listed countries.

### DIFF
--- a/client/dashboard/task-list/style.scss
+++ b/client/dashboard/task-list/style.scss
@@ -41,7 +41,7 @@
 				color: $studio-gray-50;
 			}
 
-			.woocommerce-list__item-description {
+			.woocommerce-list__item-content {
 				display: none;
 			}
 		}
@@ -62,7 +62,7 @@
 		color: $studio-gray-80;
 	}
 
-	.woocommerce-list__item-description {
+	.woocommerce-list__item-content {
 		color: $studio-gray-50;
 	}
 
@@ -192,5 +192,27 @@
 	p {
 		margin-top: 0;
 		font-size: 16px;
+	}
+}
+
+.woocommerce-task-payments {
+	.woocommerce-list__item .woocommerce-list__item-after {
+		align-self: start;
+		margin-left: $gap;
+		margin-top: $gap-large;
+	}
+
+	.woocommerce-task-payments__woocommerce-services-options {
+		border-top: 1px solid $new-muriel-gray-5;
+		margin-top: $gap-small;
+		margin-left: $gap-large;
+		padding-top: $gap-small;
+
+		.components-checkbox-control__label {
+			font-size: 16px;
+			line-height: 22px;
+			padding-left: $gap-small;
+			color: #1a1a1a;
+		}
 	}
 }
--- a/client/dashboard/task-list/style.scss
+++ b/client/dashboard/task-list/style.scss
@@ -202,11 +202,20 @@
 		margin-top: $gap-large;
 	}
 
+	.woocommerce-list__item-title {
+		border-top: 1px solid $studio-gray-5;
+		padding-top: $gap;
+	}
+
 	.woocommerce-task-payments__woocommerce-services-options {
-		border-top: 1px solid $new-muriel-gray-5;
-		margin-top: $gap-small;
-		margin-left: $gap-large;
-		padding-top: $gap-small;
+		border-top: 1px solid $studio-gray-5;
+		margin-top: $gap;
+		margin-left: $gap-larger;
+		padding-top: $gap;
+
+		.components-checkbox-control__input {
+			margin-left: -$gap-larger;
+		}
 
 		.components-checkbox-control__label {
 			font-size: 16px;
--- a/client/dashboard/task-list/tasks/payments/index.js
+++ b/client/dashboard/task-list/tasks/payments/index.js
@@ -259,13 +259,4 @@ export default compose(
 			isJetpackConnected: isJetpackConnected(),
 		};
 	} )
-	/*withDispatch( dispatch => {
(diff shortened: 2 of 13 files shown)


client/dashboard/task-list/tasks/payments/index.js
```diff
@@ -0,0 +1,271 @@
...
+			countryCode,
+			isSettingsError,
+			isSettingsRequesting,
+			settings,
+			profileItems: getProfileItems(),
+			activePlugins: getActivePlugins(),
+			isJetpackConnected: isJetpackConnected(),
+		};
+	} )
+	/*withDispatch( dispatch => {
```
joshuatf: Was this left here for future use?

packages/components/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 4.0.0
+- Changed the <List /> `description` prop to `content` and allowed content nodes to be passed in addition to strings.
```
joshuatf: 👍

src/Features/Onboarding.php
```diff
@@ -81,6 +82,16 @@ public function should_show_profiler() {
 		return $is_completed || $is_skipped ? false : true;
 	}
 
+	/**
+	 * Returns true if the task list should be displayed (not completed or hidden off the dashboard.
+	 *
+	 * @return bool
+	 */
+	public function should_show_tasks() {
```
joshuatf: Awesome!  Just added a `@todo` for this in my PR as well.

client/wc-api/onboarding/selectors.js
```diff
@@ -77,6 +77,24 @@ const getPluginInstallations = getResource => plugins => {
 	return installations;
 };
 
+const isJetpackConnected = ( getResource, requireResource ) => (
```
joshuatf: Sweet- thanks for adding this!

client/wc-api/onboarding/operations.js
```diff
@@ -170,6 +171,28 @@ function activatePluginToResource( response, items ) {
...
+
+		return [
+			fetch( {
+				path: url,
+			} )
+				.then( response => {
+					return { [ resourceName ]: { data: response } };
+				} )
+				.catch( error => {
+					return { [ resourceName ]: { error: String( error.message ) } };
```
joshuatf: Do you think we should return the full error here or just the string?  I've been returning the full error "just in case" we need it, but might be better if I update mine to just the string to keep things consistent if there aren't any use cases for the error codes.

client/dashboard/task-list/style.scss
```diff
@@ -194,3 +194,25 @@
...
+
+.woocommerce-task-payments {
+	.woocommerce-list__item .woocommerce-list__item-after {
+		align-self: start;
+		margin-left: $gap;
+		margin-top: $gap-large;
+	}
+
+	.woocommerce-task-payments__woocommerce-services-options {
+		border-top: 1px solid $new-muriel-gray-5;
```
joshuatf: I'm also seeing the same border top for `woocommerce-list__item-text` in the designs and the payment borders not extending across the checkboxes.

Screen Shot 2019-08-29 at 7 41 04 PM

client/dashboard/task-list/style.scss
```diff
@@ -194,3 +194,25 @@
...
+		margin-top: $gap-small;
+		margin-left: $gap-large;
+		padding-top: $gap-small;
```
joshuatf: Minor nitpick: I think this might be `$gap` (16px).

justinshreve:
> One other question I had (which could be handled in a follow-up if at all) is whether or not we should also be pushing Jetpack/WCS activation here with a message like Do you want to install and activate Jetpack and WooCommerce Services to allow automatic account creation for payment services? Since we do it for tax and shipping already, I was a little surprised not to see it here but maybe it will begin to feel too pushy.

I was wondering this too, but I think it's a mix of not being pushy, and because WCS is not requirement for any of the payment methods like it is for automatic tax calculation, or label printing. WCS in this case just makes set-up a little easier.

cc @jameskoster

PR #2854, opened 2019-08-28
Update no data message in table

LevinMedia:
This small PR updates the default no data error message for tables. The existing message 

"No data for the selected date range"

has a problem on reports that don't feature date range pickers by default. 

The new message is date range agnostic:

"No data to display" 

Eventually we may want to create a method for individual reports to customize this message depending on their content.


psealock (approved):
🚢

PR #2863, opened 2019-08-30
Onboarding - add quick testing links for Calypso flows

justinshreve:
This PR builds upon #2862 by adding quick methods for testing the Calypso parts of the onboarding project (Jetpack and Helper flows). It also combines all these little test methods under one help tab.
 
(Per https://github.com/Automattic/wp-calypso/pull/35793, testing these pieces together can be difficult for outside teams -- this hopefully makes that easier by giving a consistent way for jumping into those flows).

**Note** This PR also replaces `WOOCOMMERCE_CALYPSO_LOCAL` with `WOOCOMMERCE_CALYPSO_ENVIRONMENT`. Right now, `development` is the only value that will correctly load the flows. I think we can probably open the Calypso pieces to `wpcalypso` for wider testing. That would drop the requirement for needing to build Calypso yourself.  

* Go to the orders page
* Verify you still see the setup wizard tab, and use the various reset options and calypso test links.

### DIFF
--- a/src/Features/Onboarding.php
+++ b/src/Features/Onboarding.php
@@ -486,4 +486,21 @@ class Onboarding {
 		wp_safe_redirect( wc_admin_url() );
 		exit;
 	}
+
+	/**
+	 * Reset the onboarding task list and redirect to the dashboard.
+	 */
+	public static function reset_task_list() {
+		if (
+			! Loader::is_admin_page() ||
+			! isset( $_GET['reset_task_list'] ) // WPCS: CSRF ok.
+		) {
+			return;
+		}
+
+		$new_value = 1 === absint( $_GET['reset_task_list'] ) ? 'no' : 'yes'; // WPCS: CSRF ok.
+		update_option( 'woocommerce_task_list_hidden', $new_value );
+		wp_safe_redirect( wc_admin_url() );
+		exit;
+	}
 }
--- a/src/Features/Onboarding.php
+++ b/src/Features/Onboarding.php
@@ -447,16 +447,17 @@ class Onboarding {
 				: '<p><a href="' . wc_admin_url( '&reset_task_list=0' ) . '" class="button button-primary">' . __( 'Disable', 'woocommerce-admin' ) . '</a></p>'
 			);
 
-			$help_tab['content'] .= '<h3>' . __( 'Calypso', 'woocommerce-admin' ) . '</h3>';
 
-			if ( class_exists( 'Jetpack' ) ) {
-				$help_tab['content'] .= '<p>' . __( 'Quickly access the Jetpack connection flow in Calypso.', 'woocommerce-admin' ) . '</p>';
-				$help_tab['content'] .= '<p><a href="' . wc_admin_url( '&test_wc_jetpack_connect=1' ) . '" class="button button-primary">' . __( 'Connect', 'woocommerce-admin' ) . '</a></p>';
+			if ( Loader::is_feature_enabled( 'devdocs' ) && defined( 'WP_DEBUG' ) && WP_DEBUG ) {
+				$help_tab['content'] .= '<h3>' . __( 'Calypso / WordPress.com', 'woocommerce-admin' ) . '</h3>';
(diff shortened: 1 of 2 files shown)


src/Features/Onboarding.php
```diff
@@ -418,24 +419,109 @@ public static function update_help_tab() {
...
+			);
+
+			$help_tab['content'] .= '<h3>' . __( 'Task List', 'woocommerce-admin' ) . '</h3>';
+			$help_tab['content'] .= '<p>' . __( 'If you need to enable or disable the task list, please click on the button below.', 'woocommerce-admin' ) . '</p>' .
+			( 'yes' === $task_list_hidden
+				? '<p><a href="' . wc_admin_url( '&reset_task_list=1' ) . '" class="button button-primary">' . __( 'Enable', 'woocommerce-admin' ) . '</a></p>'
+				: '<p><a href="' . wc_admin_url( '&reset_task_list=0' ) . '" class="button button-primary">' . __( 'Disable', 'woocommerce-admin' ) . '</a></p>'
+			);
+
+			$help_tab['content'] .= '<h3>' . __( 'Calypso', 'woocommerce-admin' ) . '</h3>';
```
joshuatf: Since store owners will also have access to this tab, should we hide this behind a dev flag or do you think it will be useful to them as well?

If it is useful to them, I think changing this to `WordPress.com` may be more recognizable to store owners.

src/Features/Onboarding.php
```diff
@@ -444,19 +530,20 @@ public static function update_help_tab() {
...
+		$new_value = 1 === absint( $_GET['reset_profiler'] ) ? false : true;
+
 		$request = new \WP_REST_Request( 'POST', '/wc-admin/v1/onboarding/profile' );
 		$request->set_headers( array( 'content-type' => 'application/json' ) );
 		$request->set_body(
 			wp_json_encode(
 				array(
-					'completed' => false,
-					'skipped'   => false,
+					'completed' => $new_value,
```
joshuatf: Since this is accessible to store owners, this would update both `skipped` and `completed` to true, right?

I'm wondering if this may create discrepancies in data later via tracks.  Maybe this should always set `completed` to false and `skipped` is the toggleable value?

src/Features/Onboarding.php
```diff
@@ -418,24 +419,109 @@ public static function update_help_tab() {
...
+				array(
+					'body' => array(
+						'home_url'     => home_url(),
+						'redirect_uri' => $redirect_uri,
+					),
+				)
+			);
+
+			$code = wp_remote_retrieve_response_code( $request );
+			if ( 200 !== $code ) {
```
joshuatf: Should we thrown an error when this happens?

src/Features/Onboarding.php
justinshreve: Great point. I think it probably makes sense to hide behind a dev flag.

src/Features/Onboarding.php
justinshreve: https://github.com/woocommerce/woocommerce-admin/pull/2863/commits/52ef43e6054181fd540f56f12d5f12d62a2cc2ba

src/Features/Onboarding.php
justinshreve: Yeah agreed. I've set completed to false, meaning they have to go back through the wizard to have it set to true, or they can 'skip' again.

https://github.com/woocommerce/woocommerce-admin/pull/2863/commits/52ef43e6054181fd540f56f12d5f12d62a2cc2ba

src/Features/Onboarding.php
justinshreve: https://github.com/woocommerce/woocommerce-admin/pull/2863/commits/52ef43e6054181fd540f56f12d5f12d62a2cc2ba

justinshreve:
@joshuatf Thanks for the feedback. Handled in https://github.com/woocommerce/woocommerce-admin/pull/2863/commits/52ef43e6054181fd540f56f12d5f12d62a2cc2ba.

PR #2867, opened 2019-09-02
Refactor client/packages to avoid direct wcSettings usage

mikejolley:
> Note: leaving this pull up for now, but an alternative has been proposed in #2917 (added by @nerrad)

This PR's main points include:

1. Replacing `wcSettings` with a new `wcSharedSettings` global, as well as introducing a new filter that makes its purpose more clear. The rename is to avoid conflicts with core whilst the transition takes place.

2.  A new package `@woocommerce/settings` which reads this global and exports constants. See https://github.com/woocommerce/woocommerce-admin/commit/6a3f733320ebd6fd262e710455fb1133f40aa19e I'm working on Blocks using this also now it's available.

3. wc-admin specific constants are broken out into a separate file which has an alias to make it easy to access. See https://github.com/woocommerce/woocommerce-admin/commit/23f65248aec6e920ca9b66c9f2974d58bfafe0da

4. Cleanup:
    - [This commit updates packages to use the new package](https://github.com/woocommerce/woocommerce-admin/commit/0305d22b56dc7fdf4495022db6db8777ae8c3a41)
    - [Removing unneeded global usage from empty content package](https://github.com/woocommerce/woocommerce-admin/commit/2cab5699f09d579a449f6d409b8882b10a941638)
    - [This commit updates the client to use the new package](https://github.com/woocommerce/woocommerce-admin/commit/1a128673977e37a4abb8294d807dadf16b79bf7c)
    - [Fixing tests because of the new global](https://github.com/woocommerce/woocommerce-admin/commit/7e02768b7d2b67e708568780577a694b2d8a3a01)

All plugins will add variables through the new filter. Shared variables, and variables used by packages, should be added and exported within `@woocommerce/settings package. WC-admin specific stuff should go in `@woocommerce-admin/constants` e.g. onboarding endpoints.

This PR is pretty hefty but much of it is replacing existing `wcSettings` call with the import of the package.

@psealock


nerrad:
While reviewing https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/936 I noticed that `@woocommerce/settings` has been published as a stable version (assuming we're following semver).  Should we publish new _stable_ packages to npm before the pull they are introduced in finishes review?  The approach I would follow is:

- do the package add in it's own pull/branch
- pulls experimenting using the package can import directly from the branch in github temporarily while in review.
- once implementation of the package is accepted _then_ publish.

mikejolley:
@nerrad I did `publish-packages:dev` thinking it wouldn't publish stable, so it wasn't intentional, but I did need the package to be shared in order to implement it on the blocks side. Maybe this needs documentation. I'd be happy to follow documentation on this if it existed 🤓

nerrad:
> I did need the package to be shared in order to implement it on the blocks side. Maybe this needs documentation. I'd be happy to follow documentation on this if it existed 🤓

Ya I grok that the need was there.  I just think we should be wary about publishing packages on npm until they are ready.  No worries.

Re documentation: Ya that would probably be good to have spelled out.  We could base off of something similar to GB ([#](https://github.com/WordPress/gutenberg/blob/master/packages/README.md)). 

>  I did publish-packages:dev thinking it wouldn't publish stable

That sounds like a bug then as it definitely resulted in something unexpected!  cc @psealock or @joshuatf .

src/Features/ActivityPanels.php
```diff
@@ -37,7 +37,7 @@ public static function get_instance() {
 	 */
 	public function __construct() {
 		add_filter( 'wc_admin_get_user_data_fields', array( $this, 'add_user_data_fields' ) );
-		add_action( 'woocommerce_components_settings', array( $this, 'component_settings' ), 20 ); // Run after Automattic\WooCommerce\Admin\Loader.
+		add_action( 'woocommerce_shared_settings', array( $this, 'component_settings' ), 20 ); // Run after Automattic\WooCommerce\Admin\Loader.
```
jeffstieler: I know it wasn't introduced in this commit, but should this be `add_filter()` instead?

packages/settings/index.js
```diff
@@ -0,0 +1,44 @@
+const defaults = {
+	adminUrl: '',
+	countries: [],
+	currency: {
+		code: 'USD',
```
jeffstieler: Again, I know this wasn't introduced in this PR, but isn't the default currency for core WC `GBP`?

packages/settings/README.md
```diff
@@ -0,0 +1,21 @@
+# Shared Settings
+
+Exports global, shared settings from WooCommerce. All settings registered through the `woocommerce_shared_settings` filter are added to the `wcSettings` global and then exported by this package.
+
+Some commonly used settings are handled directly by this package, for example `currency`. Others must be handled by your all.
```
jeffstieler: > Others must be handled by your all.

Typo?

src/Features/ActivityPanels.php
mikejolley: Looks like it...

nerrad: Fixed in the alternative pull I did (https://github.com/woocommerce/woocommerce-admin/pull/2917)

mikejolley:
Putting this on-hold since @nerrad is working on an alternative PR.

nerrad:
Closing this pull but leaving the branch available for now - the branch can be used as source material for some of the refactors Mike did removing necessity for settings.

PR #2870, opened 2019-09-03
Use upstream webpack-rtl-plugin

Aljullu:
We were using a custom branch of `webpack-rtl-plugin` which added the `suffix` option. But the last upstream version adds the `filename` option, so there is no need to use the custom branch anymore.

- Run `npm i && npm start`.
- Set your site language to a RTL locale (Arabic, for example).
- Verify everything still looks ok.

### DIFF
--- a/package.json
+++ b/package.json
@@ -176,7 +176,7 @@
     "url-loader": "1.1.2",
     "webpack": "4.35.3",
     "webpack-cli": "3.3.4",
-    "webpack-rtl-plugin": "github:yoavf/webpack-rtl-plugin#develop"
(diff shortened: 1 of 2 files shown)


Aljullu:
> Unrelated to this change, when I switched to RTL, the chart was still LTR. Is that by design, limitation of the chart components or something we haven't looked at yet?

I'm not an expert about this, but AFAIK charts that show time in the X axis are displayed LTR no matter the writing direction (they follow [quadrant](https://en.wikipedia.org/wiki/Quadrant_(plane_geometry)) directions). Also, that's how Google Analytics shows its charts too, if I'm not wrong.

Issue #2815, opened 2019-08-19
Customer Analytics Last Active date is off


derekakelly:
**Describe the bug**
While viewing the Customer Analytics I noticed that several of the Last Active dates were off by 1 day. It's showing that some customers were last active the day before they signed up. There are also several empty spots.

**To Reproduce**
View customers, and sort by Sign Up

**Expected behavior**
I expected the last active date to match or be after the sign up date.

**Screenshots**
![last active](https://user-images.githubusercontent.com/5824837/63300759-fd355000-c2a6-11e9-99ec-a4dec1b0800d.png)

cojennin:
I was only able to create the off by 1 day issue by doing the following:

1. Set site timezone to UTC+0
2. Create a new customer
3. Set site timezone to UTC-4

When the timezone is set to UTC-4, the customer created when the timezone was set to UTC+0 would now show a `Last Active` date before the `Sign Up` date.

Was this what triggered the issue for you or did you encounter it some other way?

derekakelly:
Hi @cojennin,
I have not adjusted my timezone since my wordpress install was first setup many years ago. Plus, I was only looking at relatively new customers when i discovered the issue.

-Derek

PR #2881, opened 2019-09-04
Fix: persist date_last_active for customer reports

cojennin:
[State](https://github.com/woocommerce/woocommerce-admin/commit/0d5def5fcc3136885bbcd18148ecb2ea0606ecfe) was added as a data point for customer reports but the SQL formatting wasn't updated to account for it. This means `date_last_active` is not always being persisted correctly to customer reports, which seems like it might be the reason for the empty spots in #2815.

- Register an account
- Add a product to the cart
- Check out
- `Last Active` and `Date Registered` should both have dates for the created customer

### Changelog Note:

Fix: persist date_last_active for customer reports 👏 @cojennin

### DIFF
--- a/src/API/Reports/Customers/DataStore.php
+++ b/src/API/Reports/Customers/DataStore.php
@@ -517,6 +517,7 @@ class DataStore extends ReportsDataStore implements DataStoreInterface {
 			'%s',
 			'%s',
 			'%s',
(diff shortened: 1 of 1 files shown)


jeffstieler:
Nice catch @cojennin! Reviewing this now.

Issue #2714, opened 2019-07-29
Data Stores: Expose SQL query objects to extensions


psealock:
Data stores build objects containing clauses of SQL queries before using them to create complete queries for fetching data. By hooking into these objects via filters, we can allow extension developers to modify queries.

For example, here is the Orders Data Store creating two objects `$totals_query` and `$intervals_query` for the summary numbers and chart, respectively. 

https://github.com/woocommerce/woocommerce-admin/blob/9fbee2b13cac2d73e538bcfba967e198b5d9933b/includes/data-stores/class-wc-admin-reports-orders-stats-data-store.php#L237-L238

### Data Stores
- [ ] #2819 /Notes/DataStore.php	
- [ ] #2882 /API/Reports/Categories/DataStore.php
- [ ] #2883 /API/Reports/Coupons/DataStore.php
- [ ] #2884 /API/Reports/Coupons/Stats/DataStore.php
- [ ] #2885 /API/Reports/Customers/DataStore.php
- [ ] #2886 /API/Reports/Customers/Stats/DataStore.php
- [ ] #2891 /API/Reports/DataStore.php
- [ ] #2892 /API/Reports/Downloads/DataStore.php
- [ ] ~~/API/Reports/Downloads/Stats/DataStore.php~~
- [ ] ~~/API/Reports/Orders/DataStore.php~~
- [ ] ~~/API/Reports/Orders/Stats/DataStore.php~~
- [ ] ~~/API/Reports/Products/DataStore.php~~
- [ ] ~~/API/Reports/Products/Stats/DataStore.php~~
- [ ] ~~/API/Reports/Stock/Stats/DataStore.php~~
- [ ] #2894 /API/Reports/Taxes/DataStore.php
- [ ] #2895 /API/Reports/Taxes/Stats/DataStore.php
- [ ] #2896 /API/Reports/Variations/DataStore.php

### Documentation
- [ ] ~~Example extension, depending if https://github.com/woocommerce/woocommerce-admin/issues/1964 is also ready~~

PR #2882, opened 2019-09-04
add WHERE filters to categories data store

rrennick:
See #2714

- Add a WHERE filter to categories data store
- Add a filter to the category ID to include in the results

- Load Analytics Categories without PR
- Clear transients
- Switch to PR
- Load Categories in second tab to compare results

### Changelog Note:

Dev: Add filter to categories datastore where clauses.

### DIFF
--- a/src/API/Reports/Categories/DataStore.php
+++ b/src/API/Reports/Categories/DataStore.php
@@ -114,7 +114,16 @@ class DataStore extends ReportsDataStore implements DataStoreInterface {
 			$sql_query_params['where_clause'] .= " AND ( {$order_status_filter} )";
 		}
 
-		$sql_query_params['where_clause'] .= " AND taxonomy = 'product_cat' ";
+		$where_clause = $sql_query_params['where_clause'] . " AND taxonomy = 'product_cat' ";
+		/**
+		 * Filter the notes WHERE clause before retrieving the category data.
+		 *
+		 * Allows modification of the notes select criteria.
+		 *
+		 * @param string $where_clause The generated WHERE clause.
+		 * @param array  $query_args   The original arguments for the request.
+		 */
+		$sql_query_params['where_clause'] = apply_filters( 'wc_admin_categories_where_clause', $where_clause, $query_args );
 
 		return $sql_query_params;
 	}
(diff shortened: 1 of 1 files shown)


src/API/Reports/Categories/DataStore.php
```diff
@@ -114,7 +114,16 @@ protected function get_sql_query_params( $query_args ) {
 			$sql_query_params['where_clause'] .= " AND ( {$order_status_filter} )";
 		}
 
-		$sql_query_params['where_clause'] .= " AND taxonomy = 'product_cat' ";
+		$where_clause = $sql_query_params['where_clause'] . " AND taxonomy = 'product_cat' ";
```
jeffstieler: What do you think about using the `$where_clause` variable instead of `$sql_query_params` before the `apply_filters()` call?

src/API/Reports/Categories/DataStore.php
```diff
@@ -114,7 +114,16 @@ protected function get_sql_query_params( $query_args ) {
...
+		/**
+		 * Filter the notes WHERE clause before retrieving the category data.
```
jeffstieler: Looks like some copy 🍝from the notes PR 😛.

src/API/Reports/Categories/DataStore.php
```diff
@@ -114,7 +114,16 @@ protected function get_sql_query_params( $query_args ) {
...
+		 *
+		 * Allows modification of the notes select criteria.
+		 *
+		 * @param string $where_clause The generated WHERE clause.
+		 * @param array  $query_args   The original arguments for the request.
+		 */
+		$sql_query_params['where_clause'] = apply_filters( 'wc_admin_categories_where_clause', $where_clause, $query_args );
```
jeffstieler: Should this include `$order_product_lookup_table` like the call to `get_time_period_sql_params()`?

src/API/Reports/Categories/DataStore.php
jeffstieler: Hmm.. on second thought - that would exclude the result of `get_time_period_sql_params()` from the filter. If that's desired perhaps we should switch up the variable, if not please disregard this comment. :)

rrennick:
Superseded by #2961

PR #2887, opened 2019-09-04
Add Stripe Connection Flows

justinshreve:
Part of #2684.

This PR adds the Stripe connection and configuration logic to the payment task.

It took me a bit to work through the logic and flows, but this handles all three possible ways to connect: auto-provisioning, OAuth Connection, and manual setting entry.

I wasn’t sure the best way to persist progress in the task, since the user will possibly get redirected off-site to Stripe (or PayPal in a follow up PR). I thought about just trying to pass back data via query parameters in the redirect URL, but that seemed prone to breakage. I eventually settled on storing progress in a custom option that is preloaded.

Screen Shot 2019-09-04 at 4 33 03 PM

Screen Shot 2019-09-04 at 4 33 25 PM

Screen Shot 2019-09-04 at 4 51 25 PM

#### Testing Directions

* Go under your payment settings and make sure `Live Publishable Key` and `Live Secret Key` are blank for Stripe.

* Delete the `woocommerce_onboarding_payments` option if you have previously tested this branch, or want to run through the steps again.

To Test automatic account creation:

* Run through the payment task, and select Stripe. Select the checkbox and enter an email address that does not already have a Stripe account.
* You should get a success notice that the account created.
* Verify the secret and publishable keys are set under `WooCommerce > Settings > Payments > Stripe`.

* Try again using the same email address. You should get an error message that the account already exists, and are offered a connect button instead.

To Test the OAuth flow (https required):

* You will need a stripe account. I asked at p1567621185174900-slack-woo-partners-hydra if there was a test account to use, but it doesn't sound like it.
*  Run through the payment task, and select Stripe.  Do not select the checkbox.
* Click the Connect button.
* Test both the deny and approve flows from he Stripe OAuth screens.
* Verify the secret and publishable keys are set under `WooCommerce > Settings > Payments > Stripe` after approval.

To test manual configuration:

* Run through the payment task without Jetpack / WCS enabled. If your site doesn’t support SSL, you should also see the manual configuration option.
* Make sure you can fill in the settings. Verify they are set under `WooCommerce > Settings > Payments > Stripe`.

### DIFF
--- a/client/dashboard/task-list/tasks/payments/index.js
+++ b/client/dashboard/task-list/tasks/payments/index.js
@@ -6,16 +6,17 @@
 import { __ } from '@wordpress/i18n';
 import { Fragment, Component } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
-import { filter, noop, keys, pickBy } from 'lodash';
+import { filter, noop, keys, pickBy, difference, last } from 'lodash';
 import { FormToggle, CheckboxControl } from '@wordpress/components';
 import { Button, TextControl } from 'newspack-components';
+import { withDispatch } from '@wordpress/data';
 
 /**
  * WooCommerce dependencies
  */
 import { getCountryCode } from 'dashboard/utils';
 import { Form, Card, Stepper, List } from '@woocommerce/components';
-import { getHistory, getNewPath } from '@woocommerce/navigation';
+import { getAdminLink, getHistory, getNewPath, getQuery } from '@woocommerce/navigation';
 
 /**
  * Internal dependencies
@@ -28,12 +29,60 @@ import PayPal from './paypal';
 class Payments extends Component {
 	constructor() {
 		super( ...arguments );
+
+		const query = getQuery();
+		let step = 'choose';
+		let showIndividualConfigs = false;
+
+		if ( '1' === query.installed && query.methods ) {
+			const methods = query.methods.split( ',' );
+			step = methods[ 0 ];
+			showIndividualConfigs = true;
+		}
+
+		if ( query.configured && query.methods ) {
+			const configured = query.configured.split( ',' );
+			const methods = query.methods.split( ',' );
+			step = difference( methods, configured )[ 0 ] || '';
+			showIndividualConfigs = true;
+		}
+
 		this.state = {
-			step: 'choose',
-			showIndividualConfigs: false,
+			step,
+			showIndividualConfigs,
 		};
+
 		this.completeStep = this.completeStep.bind( this );
 		this.completePluginInstall = this.completePluginInstall.bind( this );
+	}
+
+	componentDidMount() {
+		const query = getQuery();
+		if ( query.configured && query.methods ) {
+			const configured = query.configured.split( ',' );
+
+			if ( 'paypal' === last( configured ) ) {
+				this.props.createNotice(
+					'success',
+					__( 'PayPal connected sucessfully.', 'woocommerce-admin' )
+				);
+			}
+
+			if ( 'stripe' === last( configured ) ) {
+				this.props.createNotice(
+					'success',
+					__( 'Stripe connected successfully.', 'woocommerce-admin' )
+				);
+			}
+
+			const methods = query.methods.split( ',' );
+			const stepsLeft = difference( methods, configured ).length;
+
+			if ( 0 === stepsLeft ) {
+				getHistory().push( getNewPath( {}, '/', {} ) );
+				return;
+			}
+		}
 	}
 
 	getInitialValues() {
@@ -205,6 +254,23 @@ class Payments extends Component {
 		return filter( methods, method => method.visible );
 	}
 
+	getMethodsToConfigure() {
+		const query = getQuery();
+		if ( query.methods ) {
+			return query.methods.split( ',' );
+		}
+
+		const { values } = this.formData;
+		const methods = {
+			stripe: values.stripe,
+			paypal: values.paypal,
+			'klarna-checkout': values.klarna_checkout,
+			'klarna-payments': values.klarna_payments,
+			square: values.square,
+		};
+		return keys( pickBy( methods ) );
+	}
+
 	getPluginsToInstall() {
 		const { values } = this.formData;
 		const pluginSlugs = {
@@ -227,7 +293,12 @@ class Payments extends Component {
 			values.square;
 
 		const { showIndividualConfigs } = this.state;
-		const { activePlugins } = this.props;
+		const { activePlugins, countryCode, isJetpackConnected } = this.props;
+
+		const manualConfig =
+			isJetpackConnected && activePlugins.includes( 'woocommerce-services' ) ? false : true;
+
+		const methods = this.getMethodsToConfigure();
 
 		const steps = [
 			{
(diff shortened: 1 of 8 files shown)


client/dashboard/task-list/tasks/payments/index.js
```diff
@@ -6,30 +6,96 @@
...
+
+		const installed = woocommerce_onboarding_payments && woocommerce_onboarding_payments.installed;
+		const methods =
+			( woocommerce_onboarding_payments && woocommerce_onboarding_payments.methods ) || [];
+		const configured =
+			( woocommerce_onboarding_payments && woocommerce_onboarding_payments.configured ) || [];
+
+		// Handle redirect back from Stripe
+		if (
+			query[ 'stripe-connect' ] &&
```
justinshreve: @joshuatf I ended up refactoring the `constructor` quite a bit in https://github.com/woocommerce/woocommerce-admin/pull/2897 since the first pass here was a bit messy. It may be worth looking at the logic there instead. Hopefully that doesn't make the review too difficult. Let me know if it would be better to pull out just those changes back here.

client/dashboard/task-list/tasks/payments/stripe.js
```diff
@@ -0,0 +1,317 @@
...
+				showManualConfiguration: true,
+			} );
+		}
+	}
+
+	async autoCreateAccount() {
+		const { email, countryCode } = this.props;
+		try {
+			const result = await apiFetch( {
+				path: WCS_NAMESPACE + '/connect/stripe/account',
```
joshuatf: I'm getting a 404 here - `/wp-json/wc/v1/connect/stripe/account?_locale=user`

WCS is installed and Jetpack is connected.

client/dashboard/task-list/tasks/payments/stripe.js
```diff
@@ -0,0 +1,317 @@
...
+	componentDidUpdate( prevProps, prevState ) {
+		if ( false === prevState.showConnectionButtons && this.state.showConnectionButtons ) {
+			this.fetchOAuthConnectURL();
+		}
+	}
+
+	async fetchOAuthConnectURL() {
+		const { returnUrl } = this.props;
+		try {
+			const result = await apiFetch( {
```
joshuatf: Also getting a 404 for this one.

client/dashboard/task-list/tasks/payments/index.js
```diff
@@ -6,30 +6,96 @@
...
 	constructor() {
 		super( ...arguments );
+
+		let step = 'choose';
+		let showIndividualConfigs = false;
+		let forceStripeManual = false;
+
+		const query = getQuery();
+		const { options } = this.props;
+		const { woocommerce_onboarding_payments, woocommerce_stripe_settings } = options;
```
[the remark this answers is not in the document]
justinshreve: Thinking on this, I think you are right it will be better to determine from the different settings groups rather than an additional status option.

> #2897 to avoid a bunch of rebasing.

If you are fine with that, I can handle that overhaul there where I think it will be easier, especially after some of the refactoring I did. I'll still wait to merge both of these until they are ready.

client/dashboard/task-list/tasks/payments/stripe.js
justinshreve: `WooCommerce Stripe Gateway` is correctly installed and activated, correct? The first steps should do this correctly, but I just want to make sure.

https://github.com/Automattic/woocommerce-services/blob/29dfe0ba6fd3075afe08f917a6ff33c321502d9c/woocommerce-services.php#L831-L847

https://github.com/Automattic/woocommerce-services/blob/29dfe0ba6fd3075afe08f917a6ff33c321502d9c/classes/class-wc-connect-stripe.php#L42-L44

Do you see this endpoint under `v2` or `v3` instead?

joshuatf: Thanks for those links- traced it back to this line:

https://github.com/Automattic/woocommerce-services/blob/29dfe0ba6fd3075afe08f917a6ff33c321502d9c/woocommerce-services.php#L446

Seems like `$tos_accepted` is false.  Should we prompt users to accept somewhere or hide/disable this feature if not accepted?

justinshreve: It looks like we can set `update_option( 'woocommerce_setup_jetpack_opted_in', true );` to true and the TOS will get auto-accepted.

I think we might need some additional language for this though. Asked in Slack. p1569351775011500-wc-onboarding-slack.

client/dashboard/task-list/tasks/payments/stripe.js
```diff
@@ -0,0 +1,317 @@
...
+					'success',
+					__( 'Stripe connected successfully.', 'woocommerce-admin' )
+				);
+				return;
+			}
+		} catch ( error ) {
+			let errorTitle, errorMessage;
+			// This seems to be the best way to handle this.
+			// github.com/Automattic/woocommerce-services/blob/cfb6173deb3c72897ee1d35b8fdcf29c5a93dea2/woocommerce-services.php#L563-L570
+			if ( -1 === error.message.indexOf( 'Account already exists for the provided email' ) ) {
```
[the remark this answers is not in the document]
justinshreve: From what I can tell, this error message remains the same. This is also how WooCommerce Services is determining the same state, so it seems to be the best way to handle it right now.

client/dashboard/task-list/tasks/payments/stripe.js
```diff
@@ -0,0 +1,317 @@
...
+		const {
+			completeStep,
+			createNotice,
+			isSettingsError,
+			updateOptions,
+			markConfigured,
+		} = this.props;
+
+		await updateOptions( {
+			woocommerce_stripe_settings: {
```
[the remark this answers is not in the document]
justinshreve: Good call. Done in https://github.com/woocommerce/woocommerce-admin/pull/2897/commits/5be65a516aff4a62e0647bc9e0d79d4b076f7a9e.

client/dashboard/task-list/tasks/payments/stripe.js
```diff
@@ -0,0 +1,317 @@
...
+import { Form, Link } from '@woocommerce/components';
+import withSelect from 'wc-api/with-select';
+import { WCS_NAMESPACE } from 'wc-api/constants';
+
+class Stripe extends Component {
+	constructor( props ) {
+		super( props );
+
+		this.state = {
+			showErrorModal: false,
```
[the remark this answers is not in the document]
justinshreve: Done in https://github.com/woocommerce/woocommerce-admin/pull/2897/commits/5be65a516aff4a62e0647bc9e0d79d4b076f7a9e.

client/dashboard/task-list/tasks/payments/stripe.js
justinshreve: Added a TOS message and acceptance to the profile wizard in https://github.com/woocommerce/woocommerce-admin/pull/2897/commits/5be65a516aff4a62e0647bc9e0d79d4b076f7a9e.

justinshreve:
https://github.com/woocommerce/woocommerce-admin/pull/2897#issuecomment-535050308

Issue #1320, opened 2019-01-16
Data store refactor and pattern changes


joshuatf:
Each data store has a number of reusable methods inherited from `WC_Admin_Reports_Data_Store`.  These work well, but there are a few places we're starting to repeat ourselves and use different patterns that make DRYing it up difficult.

Should we consider making these patterns consistent and extracting repeated areas in each data store to parent methods?  This is a potentially big refactor, so before starting I think we should consider whether or not this is worth the time.

Here's an example of a discussion around an area that would be difficult to DRY up without refactoring first - https://github.com/woocommerce/wc-admin/pull/1285#discussion_r247804092

### Potential updates

Here are a few areas we might want to touch on.  I think that separating these into child issues might make this more manageable if we choose to do these.

- [ ] **Standardize sql query params**
Under `get_sql_query_params()` we use multiple patterns to add where and from filters:

```
$sql_query_params['where_clause'] .= " AND ( {$order_status_filter} )";
```
```
$where_subclause = implode( " $operator ", array_filter( array( $where_subclause, $order_status_filter ) ) );
```
```
$products_where_clause .= " AND ( {$order_status_filter} )";
```
```
$where_filters[] = $order_status_filter;
```
```
$sql_query_params['where_clause'] .= " AND ( {$order_status_filter} )";
$sql_query_params['where_clause'] .= ' AND variation_id > 0';
```
If we standardize these a bit more, we can reduce mental overhead and handle a lot of the common filter queries in a parent method.

Maybe instead of using strings here, we could use an array and join by the operator:
```php
$sql_query_params['where'] = array(
    "( {$order_status_filter} )",
    "variation_id > 0"
);
```

Note that with the above, we may need to consider how this works with default excluded status conditions that should not use the `OR` operator if `match` is set to `ANY` in the query param.

Also we have `get_sql_query_params()` and `update_sql_query_params()` - both do the same thing, but the latter updates in place and also handles totals and intervals.
- [ ] **Ambiguous columns** - https://github.com/woocommerce/wc-admin/issues/1308
- [ ] **DRY up get_data**
The selection query can be extracted out to handle selections if sql params are standardized so we're not repeating the same thing in every data store.
```php
    "SELECT
            {$selections}
        FROM
            {$table_name}
            {$sql_query_params['from_clause']}
        WHERE
            1=1
            {$sql_query_params['where_time_clause']}
            {$sql_query_params['where_clause']}
        ORDER BY
[...]

PR #2891, opened 2019-09-05
add filters to report data store

rrennick:
See #2714

- Add filters to reports data store

- Check that both Dashboard & Analytics screens still function

### Changelog Note:

Dev: Add filters to reports datastore.

### DIFF
--- a/src/API/Reports/DataStore.php
+++ b/src/API/Reports/DataStore.php
@@ -1039,7 +1039,8 @@ class DataStore {
 		 * Allows filtering of the objects included or excluded from reports.
 		 *
 		 * @param array $ids List of object Ids.
-		 * @param array $query_args        The original arguments for the request.
+		 * @param array $query_args The original arguments for the request.
+		 * @param array $field      The field in the query arguments being filtered.
 		 */
 		$ids = apply_filters( 'wc_admin_reports_ ' . $field, $ids, $query_args, $field );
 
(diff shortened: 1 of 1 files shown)


src/API/Reports/DataStore.php
```diff
@@ -440,7 +440,15 @@ protected function cast_numbers( $array ) {
...
-		$selections = $this->report_columns;
+		/**
+		 * Filter the report columns before retrieving data.
+		 *
+		 * Allows filtering of the report columns included in reports.
+		 *
+		 * @param array $report_columns Associative list of columns.
+		 * @param array $query_args     The original arguments for the request.
+		 */
+		$selections = apply_filters( 'wc_admin_reports_selected_columns', $this->report_columns, $query_args );
```
jeffstieler: Nice work on this @rrennick.

I think we're going to need to add more context to this filter (and probably to the DataStore classes as well) in order for developers to reliably determine which report they're modifying.

Though it's expanding the scope quite a bit, I think this PR is the place to do it. What do you think about adding `$report_type` to all of the report `DataStore`s and including it in the `DataStoreInterface` and the filter names?

We also need to add a filter for the `$column_types`.

rrennick: @jeffstieler Thanks for the review 💯 We discussed #2714 and #1320 at our team meetings last week and decided to do the refactor so that these filters provide a more consistent interface across data stores (including the ones you noted in your feedback).

rrennick:
Superseded by #2961

PR #2897, opened 2019-09-05
Add Stripe & PayPal Connection Flows

justinshreve:
Part of #2684. 

This PR adds the Stripe & PayPal connection flows to the payments task list. 

PayPal does not require WCS for the OAuth flow, so two modes are supported here. OAuth connection, or manual key entry if the OAuth connection fails.

Stripe does require WCS. There is an auto-create account mode, a Oauth flow, and manual configuration.

Screen Shot 2019-09-04 at 4 33 03 PM

Screen Shot 2019-09-04 at 4 33 25 PM

Screen Shot 2019-09-04 at 4 51 25 PM

### Screenshots

Screen Shot 2019-09-05 at 3 31 13 PM

Screen Shot 2019-09-05 at 3 31 54 PM

* Delete the `woocommerce_onboarding_payments` option if you have previously tested this branch, the Stripe branch, or want to run through the steps again. We will clear this automatically in the future.

PayPal:

* Go under your payment settings and make sure `API Username` and `API Password` are blank for PayPal Checkout.
* Go to the payments task, and make sure PayPal is selected. Attempt to connect through the OAuth flow. You will need a PayPal Business account.
* Test the manual mode by clearing the above onboarding option, and trying again. Instead of clicking the connect button, manually set your URL to `/wp-admin/admin.php?page=wc-admin&task=payments&paypal-connect=0`.

Stripe:

* Go under your payment settings and make sure `Live Publishable Key` and `Live Secret Key` are blank for Stripe.

See testing directions on https://github.com/woocommerce/woocommerce-admin/pull/2887.

### DIFF
--- a/client/dashboard/task-list/tasks/payments/paypal.js
+++ b/client/dashboard/task-list/tasks/payments/paypal.js
@@ -2,16 +2,255 @@
 /**
  * External dependencies
  */
-import { Component } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+import { Component, Fragment } from '@wordpress/element';
+import apiFetch from '@wordpress/api-fetch';
+import { Button, TextControl } from 'newspack-components';
+import { getQuery } from '@woocommerce/navigation';
+import { withDispatch } from '@wordpress/data';
+import { compose } from '@wordpress/compose';
+import interpolateComponents from 'interpolate-components';
 
 /**
  * WooCommerce dependencies
  */
+import { WC_ADMIN_NAMESPACE } from 'wc-api/constants';
+import { Form, Link } from '@woocommerce/components';
+import withSelect from 'wc-api/with-select';
 
 class PayPal extends Component {
+	constructor( props ) {
+		super( props );
+
+		this.state = {
+			connectURL: '',
+			showManualConfiguration: props.manualConfig,
+		};
+
+		this.updateSettings = this.updateSettings.bind( this );
+	}
+
+	componentDidMount() {
+		const { showManualConfiguration } = this.state;
+
+		const query = getQuery();
+		// Handle redirect back from PayPal
+		if ( query[ 'paypal-connect' ] ) {
+			if ( '1' === query[ 'paypal-connect' ] ) {
+				this.props.markConfigured( 'paypal' );
+				this.props.createNotice(
+					'success',
+					__( 'PayPal connected successfully.', 'woocommerce-admin' )
+				);
+				return;
+			}
+
+			/* eslint-disable react/no-did-mount-set-state */
+			this.setState( {
+				showManualConfiguration: true,
+			} );
+			/* eslint-enable react/no-did-mount-set-state */
+			return;
+		}
+
+		if ( ! showManualConfiguration ) {
+			this.fetchOAuthConnectURL();
+		}
+	}
+
+	componentDidUpdate( prevProps, prevState ) {
+		if (
+			true === prevState.showManualConfiguration &&
+			false === this.state.showManualConfiguration
+		) {
+			this.fetchOAuthConnectURL();
+		}
+
+		if ( false === prevProps.optionsIsRequesting && true === this.props.optionsIsRequesting ) {
+			this.props.setRequestPending( true );
+		}
+
+		if ( true === prevProps.optionsIsRequesting && false === this.props.optionsIsRequesting ) {
+			this.props.setRequestPending( false );
+		}
+	}
+
+	async fetchOAuthConnectURL() {
+		this.props.setRequestPending( true );
+		try {
+			const result = await apiFetch( {
+				path: WC_ADMIN_NAMESPACE + '/onboarding/plugins/connect-paypal',
+				method: 'POST',
+			} );
+			if ( ! result || ! result.connectUrl ) {
+				this.props.setRequestPending( false );
+				this.setState( {
+					showManualConfiguration: true,
+				} );
+				return;
+			}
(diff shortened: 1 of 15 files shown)


client/dashboard/task-list/tasks/payments/paypal.js
```diff
@@ -2,16 +2,255 @@
...
+				{ __( 'Connect', 'woocommerce-admin' ) }
+			</Button>
+		);
+	}
+
+	async updateSettings( values ) {
+		const { createNotice, isSettingsError, updateOptions, markConfigured } = this.props;
+
+		this.props.setRequestPending( true );
+		await updateOptions( {
```
joshuatf: I think this await pattern we've been using will not work with wc-api.  I'm not sure it returns the promise and `isSettingsError` will not yet have updated.

Okay, to handle this in a follow-up as well since we have a few other cases of this that I've also introduced and need correcting.

client/dashboard/task-list/tasks/payments/paypal.js
```diff
@@ -2,16 +2,255 @@
...
-export default PayPal;
+PayPal.defaultProps = {
+	manualConfig: false, // WCS is not required for the PayPal OAuth flow, so we can default to smooth connection.
+};
+
+export default compose(
+	withSelect( select => {
+		const { getOptions, isOptionsRequesting } = select( 'wc-api' );
+		const options = getOptions( [ 'woocommerce_ppec_paypal_settings' ] );
+		const optionsIsRequesting = Boolean(
```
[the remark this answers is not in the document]
justinshreve: Thanks, fixed in https://github.com/woocommerce/woocommerce-admin/pull/2897/commits/5be65a516aff4a62e0647bc9e0d79d4b076f7a9e.

src/API/OnboardingPlugins.php
```diff
@@ -405,6 +418,38 @@ public function finish_wccom_connect( $rest_request ) {
 		);
 	}
 
+	/**
+	 * Returns a URL that can be used to connect to PayPal.
+	 *
+	 * @param  object $rest_request Request details.
```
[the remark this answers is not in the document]
justinshreve: Removed. https://github.com/woocommerce/woocommerce-admin/pull/2897/commits/5be65a516aff4a62e0647bc9e0d79d4b076f7a9e

client/dashboard/task-list/tasks/payments/paypal.js
justinshreve: Logged a follow-up issue for us: https://github.com/woocommerce/woocommerce-admin/issues/2959

justinshreve:
> One little oddity I noticed after setting up PayPal was that the list item was not marked complete and I was also not able to enter the Payments step again when clicking on the item.

I'll fix this in a follow-up. There are still a few payment methods that need added to the payments task still. (Square and Klarna)

justinshreve:
@joshuatf I ended up combining the Stripe & PayPal PRs. I'm sorry this makes it a bit larger, but because of the refactor I did in the PayPal branch, I think this might be easier for you. I handled feedback from the other PR as well (https://github.com/woocommerce/woocommerce-admin/pull/2887).

---

I have been thinking about your comments about `woocommerce_onboarding_payments`. Now that I've jumped back into the code and refreshed myself, I actually feel like `woocommerce_onboarding_payments` might be the best way to handle this right now.

The main purpose is to accurately persist the state of the payments **task** (not the state of the configuration). This is important because the payments task involves leaving the page multiple times, off-site (PayPal & Stripe), and requires being dropped right back into the proper part of the flow.

You are right that we could try to determine state of the task based on the configured settings, but this feels a bit more error prone to me and it seems like the checks could get complex. We'd be checking each payment configuration (which can be a max of 4 if you end up with Square and Klarna options), plus their install status. I see `woocommerce_onboarding_payments ` more of a cache for quickly determining this based on what we already know. We could clear it at the end of the task (and mark overall completion a different way) so it doesn't become stale if the task is re-ran.

> Same question here as #2887 - I see that we need woocommerce_onboarding_payments to determine if they've completed the first step or not,
> but maybe we can not skip the first step so they can revisit it later.

I also think revisiting the task would be easier with this method. We just update one of the flags or clear the option if they want to go back in to a different step. If we are checking each individual payment method configuration, do we need to clear out those settings manually to get back?

What do you think, would you be OK with this path forward?

In the mean time, I'm going to work on the remainder of the payment options that need added still. (If we do want to check the individual settings instead, that might actually be easier in a future PR with all the methods present).

client/dashboard/task-list/tasks/payments/paypal.js
```diff
@@ -0,0 +1,257 @@
...
+		await updateOptions( {
+			woocommerce_ppec_paypal_settings: {
+				...this.props.options.woocommerce_ppec_paypal_settings,
+				api_username: values.api_username,
+				api_password: values.api_password,
+				enabled: 'yes',
+			},
+		} );
+
+		if ( ! isSettingsError ) {
```
joshuatf: Should we add a success message here too when settings are updated successfully?  After saving PayPal settings manually I get bumped back to the task list without any other notice; may be a little confusing to some users.

justinshreve: Good catch, thanks! Added a message on manual save for both PayPal and Stripe: https://github.com/woocommerce/woocommerce-admin/pull/2897/commits/19f561e2b1dcd232e2be1357d785f19ae38b3ed0

PR #2917, opened 2019-09-11
Refactor: Alternative approach to implementing extendable settings from the server

nerrad:
This is a work in progress addressing an alternative to #2867 and a companion to https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/956.

You can read more of the background in the [corresponding wc-blocks pull](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/956.), but essentially this work is being done so:

- `wcSettings` is only output on routes that need it (via enqueueing using the `wc-settings` handle as a dependency).
- introduction of a new interface for explicitly registering data to load with a script (using `AssetDataRegistry`.  This avoids some of the problems with filtering this data.
- introduction of a DIC (Dependency Injection Container) for better php architecture.  Eventually, I'd like to see this be implemented more widely, but to help with gradual migration the container is exposed on `wc_blocks_container()`.
- introduction of a new interface on the javascript side for retrieving and setting data (`getSetting`, `setSetting` - see linked pull and examples in this pull for details)

In this pull:

- back-compat implementation of `wcSettings` utilizing the new `wc-shared-settings` handle dependency for scripts when it is available.  This means that when this branch is merged, the plugin should work regardless of what version of wc-blocks is active in the environment.
- initial refactor pass of existing scripts to utilize the new aliased `@wordpress/wc-admin-settings` package (just an alias to a folder).  I didn't do refactoring of ALL packages/scripts because that would make this pull much harder to review :).  However, ideally, we'd do follow up pulls to utilize the new interface.  Reading from the `wcSettings` global is still possible, however we should move away from that asap.

## Testing

* [ ] Smoke test this branch against https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/956.
* [ ] Smoke test this branch against WC core.
* [ ] Smoke test this branch against master of wc-blocks (or without 956 in it).
* [ ] Verify existing unit tests pass for this pull.

## Next Steps

* [ ] Follow up pulls to finish refactoring existing packages over to use the new settings api (could be opportunity for improving some of the data passed around as well).
* [ ] Eventually remove back-compat code when its no longer needed (i.e. bump minimum WC version requirements).

### Some rationale

For published packages, there's a few problems with the existing usage of the `wcSettings` global.

- it's a (somewhat) hidden dependency.
- easily clobbered.
- any future changes to shape, refactoring is time-consuming
- some properties are mutated - this makes it fairly fragile for code consuming those properties.

Changes here are an improvement iteration:

- An alias allows for switching the backend serving the global without having to refactor a bunch of references.
- Introduction of a canonical source for the initial global and a `getSetting` helper (via code bundled in wc core via the wc-blocks repo).  There is a fallback setup in case wc-admin is run on earlier supported wc versions that load the global via legacy.
- `getSetting` allows for more utility with regards to retrieving values from server delivered global data in a way that protects against unexpected types and missing defaults (h/t to @mikejolley  for the original idea on that).  I introduced a third parameter to the function for including a custom validation/sanitization function to act on the value before returning.  Useful for fairly critical values.
- `setSetting` provides a clean way to set new values for settings shared between packages.  However we may want to consider the ability to set certain setting values as "protected" to prevent unexpected behaviour.

### DIFF
--- a/client/analytics/report/customers/config.js
+++ b/client/analytics/report/customers/config.js
@@ -5,6 +5,7 @@
 import { __, _x } from '@wordpress/i18n';
 import { decodeEntities } from '@wordpress/html-entities';
 import { applyFilters } from '@wordpress/hooks';
+import { COUNTRIES as countries } from '@woocommerce/wc-admin-settings';
 
 /**
  * Internal dependencies
@@ -116,9 +117,6 @@ export const advancedFilters = applyFilters( CUSTOMERS_REPORT_ADVANCED_FILTERS_F
 				component: 'Search',
 				type: 'countries',
 				getLabels: async value => {
-					const countries =
-						( wcSettings.dataEndpoints && wcSettings.dataEndpoints.countries ) || [];
-
 					const allLabels = countries.map( country => ( {
 						id: country.code,
 						label: decodeEntities( country.name ),
--- a/client/analytics/report/customers/table.js
+++ b/client/analytics/report/customers/table.js
@@ -13,6 +13,7 @@ import { defaultTableDateFormat } from '@woocommerce/date';
 import { formatCurrency, getCurrencyFormatDecimal } from '@woocommerce/currency';
 import { Date, Link } from '@woocommerce/components';
 import { numberFormat } from '@woocommerce/number';
+import { COUNTRIES as countries } from '@woocommerce/wc-admin-settings';
 
 /**
  * Internal dependencies
@@ -102,9 +103,8 @@ export default class CustomersReportTable extends Component {
 	}
 
 	getCountryName( code ) {
-		const countries = ( wcSettings.dataEndpoints && wcSettings.dataEndpoints.countries ) || [];
-		const country = countries.find( c => c.code === code );
-		return country ? country.name : null;
+		const country = typeof countries[ code ] !== 'undefined' ? countries[ code ] : null;
+		return country;
 	}
 
 	getRowsContent( customers ) {
--- a/client/analytics/report/customers/table.js
+++ b/client/analytics/report/customers/table.js
@@ -103,8 +103,7 @@ export default class CustomersReportTable extends Component {
 	}
 
 	getCountryName( code ) {
-		const country = typeof countries[ code ] !== 'undefined' ? countries[ code ] : null;
-		return country;
+		return typeof countries[ code ] !== 'undefined' ? countries[ code ] : null;
 	}
 
 	getRowsContent( customers ) {
--- a/client/analytics/report/products/table-variations.js
+++ b/client/analytics/report/products/table-variations.js
@@ -13,12 +13,16 @@ import { Link } from '@woocommerce/components';
 import { formatCurrency, getCurrencyFormatDecimal } from '@woocommerce/currency';
 import { getNewPath, getPersistedQuery } from '@woocommerce/navigation';
 import { numberFormat } from '@woocommerce/number';
+import { getSetting } from '@woocommerce/wc-admin-settings';
 
 /**
  * Internal dependencies
  */
 import ReportTable from 'analytics/components/report-table';
 import { isLowStock } from './utils';
+
+const manageStock = getSetting( 'manageStock', 'no' );
+const stockStatuses = getSetting( 'stockStatuses', {} );
 
 export default class VariationsReportTable extends Component {
 	constructor() {
@@ -64,13 +68,13 @@ export default class VariationsReportTable extends Component {
 				isSortable: true,
 				isNumeric: true,
 			},
-			'yes' === wcSettings.manageStock
+			'yes' === manageStock
 				? {
 						label: __( 'Status', 'woocommerce-admin' ),
 						key: 'stock_status',
 					}
 				: null,
-			'yes' === wcSettings.manageStock
+			'yes' === manageStock
 				? {
 						label: __( 'Stock', 'woocommerce-admin' ),
 						key: 'stock',
@@ -81,7 +85,6 @@ export default class VariationsReportTable extends Component {
 	}
 
 	getRowsContent( data = [] ) {
-		const { stockStatuses } = wcSettings;
 		const { query } = this.props;
 		const persistedQuery = getPersistedQuery( query );
 
@@ -125,7 +128,7 @@ export default class VariationsReportTable extends Component {
 					),
 					value: orders_count,
 				},
-				'yes' === wcSettings.manageStock
+				'yes' === manageStock
 					? {
 							display: isLowStock( stock_status, stock_quantity, low_stock_amount ) ? (
 								<Link href={ editPostLink } type="wp-admin">
@@ -137,7 +140,7 @@ export default class VariationsReportTable extends Component {
 							value: stockStatuses[ stock_status ],
 						}
 					: null,
-				'yes' === wcSettings.manageStock
+				'yes' === manageStock
 					? {
 							display: stock_quantity,
 							value: stock_quantity,
--- a/client/analytics/settings/config.js
+++ b/client/analytics/settings/config.js
@@ -5,6 +5,7 @@
 import { __, sprintf } from '@wordpress/i18n';
 import { applyFilters } from '@wordpress/hooks';
 import interpolateComponents from 'interpolate-components';
+import { getSetting, ORDER_STATUSES } from '@woocommerce/wc-admin-settings';
 
 /**
  * Internal dependencies
@@ -13,7 +14,7 @@ import { DEFAULT_ACTIONABLE_STATUSES } from 'wc-api/constants';
 import DefaultDate from './default-date';
 
 const SETTINGS_FILTER = 'woocommerce_admin_analytics_settings';
-const DEFAUTL_DATE_RANGE = 'period=month&compare=previous_year';
+const DEFAULT_DATE_RANGE = 'period=month&compare=previous_year';
 
 const defaultOrderStatuses = [
 	'completed',
@@ -25,27 +26,33 @@ const defaultOrderStatuses = [
 	'on-hold',
 ];
 
-const actionableOrderStatuses = Array.isArray(
-	wcSettings.wcAdminSettings.woocommerce_actionable_order_statuses
-)
-	? wcSettings.wcAdminSettings.woocommerce_actionable_order_statuses
+const {
+	woocommerce_actionable_order_statuses,
+	woocommerce_excluded_report_order_statuses,
+	woocommerce_default_date_range,
+} = getSetting( 'wcAdminSettings', {
+	woocommerce_actionable_order_statuses: [],
+	woocommerce_excluded_report_order_statuses: [],
+	woocommerce_default_date_range: DEFAULT_DATE_RANGE,
+} );
+
+const actionableOrderStatuses = Array.isArray( woocommerce_actionable_order_statuses )
+	? woocommerce_actionable_order_statuses
 	: [];
 
-const excludedOrderStatuses = Array.isArray(
-	wcSettings.wcAdminSettings.woocommerce_excluded_report_order_statuses
-)
-	? wcSettings.wcAdminSettings.woocommerce_excluded_report_order_statuses
+const excludedOrderStatuses = Array.isArray( woocommerce_excluded_report_order_statuses )
+	? woocommerce_excluded_report_order_statuses
 	: [];
 
-const orderStatuses = Object.keys( wcSettings.orderStatuses )
+const orderStatuses = Object.keys( ORDER_STATUSES )
 	.filter( status => status !== 'refunded' )
 	.map( key => {
 		return {
 			value: key,
-			label: wcSettings.orderStatuses[ key ],
+			label: ORDER_STATUSES[ key ],
 			description: sprintf(
 				__( 'Exclude the %s status from reports', 'woocommerce-admin' ),
-				wcSettings.orderStatuses[ key ]
+				ORDER_STATUSES[ key ]
 			),
 		};
 	} );
@@ -112,7 +119,7 @@ export const analyticsSettings = applyFilters( SETTINGS_FILTER, [
 				'the default date range.',
 			'woocommerce-admin'
 		),
-		initialValue: wcSettings.wcAdminSettings.woocommerce_default_date_range || DEFAUTL_DATE_RANGE,
-		defaultValue: DEFAUTL_DATE_RANGE,
+		initialValue: woocommerce_default_date_range,
+		defaultValue: DEFAULT_DATE_RANGE,
 	},
 ] );
--- a/client/analytics/settings/index.js
+++ b/client/analytics/settings/index.js
@@ -116,7 +116,7 @@ class Settings extends Component {
 	 * @param {object} state - State
 	 */
 	persistChanges( state ) {
-		// TODO: Should remove global state from the file. This creates
+		// @todo Should remove global state from the file. This creates
 		// potential hard to debug side-effects.
 		wcSettings.wcAdminSettings = wcSettings.wcAdminSettings || {};
 		analyticsSettings.forEach( setting => {
--- a/client/analytics/settings/index.js
+++ b/client/analytics/settings/index.js
@@ -116,6 +116,9 @@ class Settings extends Component {
 	 * @param {object} state - State
 	 */
 	persistChanges( state ) {
+		// TODO: Should remove global state from the file. This creates
+		// potential hard to debug side-effects.
+		wcSettings.wcAdminSettings = wcSettings.wcAdminSettings || {};
 		analyticsSettings.forEach( setting => {
 			const updatedValue = state.settings[ setting.name ];
 			wcSettings.wcAdminSettings[ setting.name ] = updatedValue;
--- a/client/dashboard/profile-wizard/steps/business-details.js
+++ b/client/dashboard/profile-wizard/steps/business-details.js
@@ -14,6 +14,7 @@ import { keys, pickBy } from 'lodash';
  * WooCommerce dependencies
  */
 import { numberFormat } from '@woocommerce/number';
+import { getSetting, CURRENCY as currency } from '@woocommerce/wc-admin-settings';
 
 /**
  * Internal dependencies
@@ -22,6 +23,8 @@ import { H, Card, SimpleSelectControl, Form } from '@woocommerce/components';
 import withSelect from 'wc-api/with-select';
 import { recordEvent } from 'lib/tracks';
 import { formatCurrency } from '@woocommerce/currency';
+
+const wcAdminAssetUrl = getSetting( 'wcAdminAssetUrl', '' );
 
 class BusinessDetails extends Component {
 	constructor() {
@@ -50,7 +53,7 @@ class BusinessDetails extends Component {
 		recordEvent( 'storeprofiler_store_business_details_continue', {
 			product_number: product_count,
 			already_selling: 'no' !== selling_venues,
-			currency: wcSettings.currency.code,
+			currency: currency.code,
 			revenue,
 			used_platform: other_platform,
 			install_facebook: facebook,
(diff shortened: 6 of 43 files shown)


client/analytics/report/customers/config.js
```diff
@@ -5,6 +5,7 @@
 import { __, _x } from '@wordpress/i18n';
 import { decodeEntities } from '@wordpress/html-entities';
 import { applyFilters } from '@wordpress/hooks';
+import { COUNTRIES as countries } from '@woocommerce/wc-admin-settings';
```
Aljullu: Should we keep it in uppercase so it's clear that it's a constant when used in the code? That's what we are doing in WooCommerce Blocks.

client/analytics/report/customers/table.js
```diff
@@ -102,9 +103,8 @@ export default class CustomersReportTable extends Component {
 	}
 
 	getCountryName( code ) {
-		const countries = ( wcSettings.dataEndpoints && wcSettings.dataEndpoints.countries ) || [];
-		const country = countries.find( c => c.code === code );
-		return country ? country.name : null;
+		const country = typeof countries[ code ] !== 'undefined' ? countries[ code ] : null;
+		return country;
```
Aljullu: We could skip the `country` assignment and directly return. I think it will not affect code readability.

client/analytics/settings/config.js
```diff
@@ -25,27 +26,33 @@ const defaultOrderStatuses = [
...
-	? wcSettings.wcAdminSettings.woocommerce_actionable_order_statuses
+const {
+	woocommerce_actionable_order_statuses,
+	woocommerce_excluded_report_order_statuses,
+	woocommerce_default_date_range,
+} = getSetting( 'wcAdminSettings', {
+	woocommerce_actionable_order_statuses: [],
+	woocommerce_excluded_report_order_statuses: [],
+	woocommerce_default_date_range: DEFAULT_DATE_RANGE,
+} );
```
Aljullu: Is there any case where `wcAdminSettings` might be defined but be an empty object or miss some of the properties? I imagine if `wcAdminSettings` is `{}`, none of the defaults would be applied here, right?

client/analytics/settings/index.js
```diff
@@ -116,6 +116,9 @@ class Settings extends Component {
 	 * @param {object} state - State
 	 */
 	persistChanges( state ) {
+		// TODO: Should remove global state from the file. This creates
+		// potential hard to debug side-effects.
```
Aljullu: If I understood this correctly, we can't remove this until minimum WooCommerce version is increased, right? Should we explain that in the comment?

client/dashboard/task-list/tasks/shipping/rates.js
```diff
@@ -8,14 +8,16 @@ import { Button, TextControl } from 'newspack-components';
...
 import PropTypes from 'prop-types';
 
 /**
  * WooCommerce dependencies
  */
 import { Flag, Form } from '@woocommerce/components';
 import { getCurrencyFormatString } from '@woocommerce/currency';
+import { CURRENCY } from '@woocommerce/wc-admin-settings';
+
+const { symbol = '$', symbolPosition = 'left' } = CURRENCY;
```
Aljullu: Defaults are already defined in `client/settings/fallbacks.js`. Are they needed here too?

client/header/activity-panel/panels/orders.js
```diff
@@ -266,6 +266,7 @@ export default compose(
 			getReportItemsError,
 			isReportItemsRequesting,
 		} = select( 'wc-api' );
+		wcSettings.wcAdminSettings = wcSettings.wcAdminSettings || {};
```
Aljullu: Is the plan to remove this and use `getSetting` in this file too?

client/header/activity-panel/unread-indicators.js
```diff
@@ -8,6 +8,9 @@ import { DEFAULT_ACTIONABLE_STATUSES } from 'wc-api/constants';
 export function getUnreadNotes( select ) {
 	const { getCurrentUserData, getNotes, getNotesError, isGetNotesRequesting } = select( 'wc-api' );
 	const userData = getCurrentUserData();
+	if ( ! userData ) {
```
Aljullu: Is this needed? `getCurrentUserData` already [has a fallback](https://github.com/woocommerce/woocommerce-admin/blob/master/client/wc-api/user/selectors.js#L12).

client/settings/index.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+import * as SHARED from '@woocommerce/settings';
+
+/**
+ * Internal dependencies
+ */
+import * as FALLBACKS from './fallbacks';
+
+// if `getSetting` is not set, then it was not available so let's do
```
Aljullu: Comment capitalization:
```suggestion
// If `getSetting` is not set, then it was not available so let's do
```

packages/currency/src/index.js
```diff
@@ -18,14 +19,14 @@ import { numberFormat } from '@woocommerce/number';
  * @returns {?String} A formatted string.
  */
 export function formatCurrency( number, currencySymbol ) {
-	// default to wcSettings (and then to $) if currency symbol is not passed in
+	// default to wc-admin-settings (and then to $) if currency symbol is not passed in
```
Aljullu: I think `(and then to $)` should be removed from the comment, now. I would even say this comment could be completely removed, it doesn't seem to add any valuable information that can't be inferred from the code.

client/analytics/report/customers/config.js
nerrad: Ya, we _could_, I mostly was just looking to keep code churn to a minimum here.  Something for a future iteration?

Aljullu: Sure thing! :+1:

client/analytics/settings/config.js
nerrad: Ya this was a tricky one because `wcAdminSettings` could have x number of props.  For the purpose of this library, the fallback defaults are sufficient because those are the only properties used.  I think this is another item that is ripe for further iteration in future refactors though.

So to answer your question, yes if `wcAdminSettings` fallback is `{}` then none of the default fallbacks would apply here.

client/analytics/settings/index.js
nerrad: No, we can remove this in further refactors after this is merged and replace with usage of the new `setSetting` interface.  While `setSetting` and `getSetting` still interact with a global under the hood, it hides it behind an interface so that it's more controlled.  I mostly added this comment just as a reminder to address it.

client/dashboard/task-list/tasks/shipping/rates.js
nerrad: Nope :)

client/header/activity-panel/panels/orders.js
nerrad: Yes.  I think in an initial pass I was looking at leaving some of the "legacy" way of doing things intact so I could do more granular pulls for other refactors in further iterations (this pull was already getting pretty big).  I probably just noticed this global wasn't initialized with a default in this file and added this in.

client/header/activity-panel/unread-indicators.js
nerrad: It doesn't have a fallback in the current iteration as it's using the global directly:

```js
const getCurrentUserData = ( getResource, requireResource ) => (
	requirement = DEFAULT_REQUIREMENT
) => {
	const initialState = wcSettings.currentUserData;
	return requireResource( requirement, 'current-user-data' ).data || initialState;
};
```

After refactor in future iterations it will have a fallback.  The existing logic is fragile and will break things if for some reason currentUser isn't present (which is entirely possible if the global is clobbered).

packages/currency/src/index.js
nerrad: Ya good call, I mostly just went through and replaced old strings and missed the context.

client/analytics/report/products/table-variations.js
```diff
@@ -13,13 +13,17 @@ import { Link } from '@woocommerce/components';
...
 import { numberFormat } from '@woocommerce/number';
+import { getSetting } from '@woocommerce/wc-admin-settings';
 
 /**
  * Internal dependencies
  */
 import ReportTable from 'analytics/components/report-table';
 import { isLowStock } from './utils';
 
+const manageStock = getSetting( 'manageStock', 'no' );
```
mikejolley: Should we change some of these to booleans, if the only possible values are yes/no?

client/dashboard/profile-wizard/steps/business-details.js
```diff
@@ -23,6 +24,8 @@ import withSelect from 'wc-api/with-select';
 import { recordEvent } from 'lib/tracks';
 import { formatCurrency } from '@woocommerce/currency';
 
+const wcAdminAssetUrl = getSetting( 'wcAdminAssetUrl', '' );
```
mikejolley: In my PR I was able to remove this entirely by removing the images in favor of embedded SVGs. May want to do this also - it seemed much cleaner than passing paths around.

client/dashboard/profile-wizard/steps/business-details.js
```diff
@@ -179,7 +182,7 @@ class BusinessDetails extends Component {
 				{ extensionBenefits.map( benefit => (
 					<div className="woocommerce-profile-wizard__benefit" key={ benefit.title }>
 						<div className="woocommerce-profile-wizard__business-extension">
-							<img src={ wcSettings.wcAdminAssetUrl + benefit.icon } alt="" />
+							<img src={ wcAdminAssetUrl + benefit.icon } alt="" />
```
mikejolley: img->svg

client/dashboard/task-list/tasks/shipping/rates.js
```diff
@@ -8,14 +8,16 @@ import { Button, TextControl } from 'newspack-components';
...
 import PropTypes from 'prop-types';
 
 /**
  * WooCommerce dependencies
  */
 import { Flag, Form } from '@woocommerce/components';
 import { getCurrencyFormatString } from '@woocommerce/currency';
+import { CURRENCY } from '@woocommerce/wc-admin-settings';
+
+const { symbol, symbolPosition } = CURRENCY;
```
mikejolley: In the feedback on blocks I mentioned the idea of having a currency prefix/suffix option which contained the symbol in the correct position. This is the type of scanario I was referring to - it makes it easier to render prices if you don't need to conditonally look at left/right/right_space/left_space :)

client/header/activity-panel/index.js
```diff
@@ -30,6 +31,9 @@ import ReviewsPanel from './panels/reviews';
 import withSelect from 'wc-api/with-select';
 import WordPressNotices from './wordpress-notices';
 
+const manageStock = getSetting( 'manageStock', 'no' );
+const reviewsEnabled = getSetting( 'reviewsEnabled', 'no' );
```
mikejolley: Potentially more booleans

client/analytics/report/products/table-variations.js
nerrad: Yup I think they should be.  Punt to another iteration?  I think we'd want to review all the settings and use better types.

client/dashboard/profile-wizard/steps/business-details.js
nerrad: Agreed, but I think it should be done in its own pull rather than slipped in with a larger refactor.  Thoughts?

client/dashboard/task-list/tasks/shipping/rates.js
nerrad: Yup, agreed here also.  Punt to a separate pull?  To ensure we don't forget, do an issue?

client/analytics/report/products/table-variations.js
mikejolley: In the case of these core settings it might be best to do it 'now', mainly because a) if left it might be difficult to change this from string to bool without breakage, and b) we can see in the diff where these values are being used since those lines are being updated as part of this PR anyway.

client/dashboard/profile-wizard/steps/business-details.js
mikejolley: If it makes sense to nuke a setting completely it would save needing to rewrite bits twice? I don't mind. `wcAdminAssetUrl` was the one I think that only had usage in a few places, one of which was the component library (which makes no sense since those components should not have wc admin depdendency).

client/analytics/settings/index.js
```diff
@@ -116,6 +116,9 @@ class Settings extends Component {
 	 * @param {object} state - State
 	 */
 	persistChanges( state ) {
+		// TODO: Should remove global state from the file. This creates
```
Aljullu: We usually use `@todo` for to-do comments (see #1474):
```suggestion
		// @todo Should remove global state from the file. This creates
```

client/header/activity-panel/unread-indicators.js
Aljullu: Good point. In any case, I think it would make sense to handle the missing data case inside `getCurrentUserData()` instead of having to check if `userData` is defined in every consumer.

client/analytics/settings/config.js
Aljullu: I'm ok with improving it in a further iteration. I guess in cases like this one we could just merge both objects, no?

client/analytics/report/products/table-variations.js
nerrad: Hmm, I still would prefer to do in it's own pull because I'd want to review all instances of things that might be better typed.  It's not just that this pull is already larger but it also extends the time before its merged.

I _do_ intend on immediately releasing pulls that address things like this after this is merged.

client/analytics/settings/config.js
nerrad: I'm not sure what you mean by "merge both objects".  The fallback is only used if the property doesn't exist in the settings state.  Another reason for having a fallback is _eventually_ we might load settings via api request vs a global from the server.  In which case having the fallback allows for what to use while the request is resolving.  Of course there'd be other changes needed if we go that route, but the interface for `getSetting` shouldn't need to change.

client/dashboard/profile-wizard/steps/business-details.js
nerrad: Ya I understand the comment about rewriting bits twice, however I think there's value in differentiating between refactors preserving existing behaviour and pulls that change behaviour or enhance by nuking things.  Especially when it comes to having record of that change in the git history (which would be lost with the squash commit for this pull).

So I still think we should do this in it's own pull (which also might address other places where there's unnecessary dependencies).

client/dashboard/profile-wizard/steps/business-details.js
nerrad: As mentioned elsewhere, I think this should be addressed in it's own pull.

client/header/activity-panel/unread-indicators.js
nerrad: > In any case, I think it would make sense to handle the missing data case inside getCurrentUserData() instead of having to check if userData is defined in every consumer.

I agree, and that will be the case once it's refactored (and I'll remove the instance of checking for userData here at that point).

nerrad:
I think I've addressed the comments so far.  I should make it clear that when this pull is approved:

- I'll go through all the comments and ensure there are issues for the points brought up in the review so they aren't forgotten and can be addressed in separate pulls.
- I intend on helping refactor the remaining bits in separate pulls (and addressing the issues above).

The primary point of this pull is to get the direction approved along getting the work supporting it merged in, then future refactor pulls can go more quickly and in smaller chunks.

client/dashboard/profile-wizard/steps/business-details.js
psealock: I agree its best to do it separately for a simpler record of the changes and why.

webpack.config.js
```diff
@@ -34,6 +34,7 @@ const externals = {
 	'@wordpress/html-entities': { this: [ 'wp', 'htmlEntities' ] },
 	'@wordpress/i18n': { this: [ 'wp', 'i18n' ] },
 	'@wordpress/keycodes': { this: [ 'wp', 'keycodes' ] },
+	'@woocommerce/settings': [ 'wc', 'wcSettings' ],
```
psealock: Should we follow the pattern above?

```js
'@woocommerce/settings': { this: [ 'wc', 'wcSettings' ] }
```

client/analytics/settings/config.js
Aljullu: I meant merging the fallback and the value if they are both objects. Something like this:
```diff
export function getSetting(
	name,
	fallback = false,
	validation = ( val ) => val
) {
	const value = allSettings.hasOwnProperty( name )
		? allSettings[ name ]
		: fallback;
+	if ( isPlainObject( fallback ) && isPlainObject( value ) ) {
+		return validation( { ...fallback, ...value } );
+	}
	return validation( value );
}
```
Another idea would be setting the fallback when destructuring:
```ES6
const {
	woocommerce_actionable_order_statuses = [],
	woocommerce_excluded_report_order_statuses = [],
	woocommerce_default_date_range = DEFAULT_DATE_RANGE,
} = getSetting( 'wcAdminSettings' );
```
This way, if the object exists but it has a missing property, it would still be assigned to the default. Do you think that would work?

nerrad: Ahh I see now.  Ya I don't think it's necessary to do merges here.  On the `get`, the calling code should only be concerned with the properties it needs via destructuring, so the fallback only needs to have those properties as well.  The fallback doesn't persist, otherwise we _would_ need something like you propose.

webpack.config.js
nerrad: yup good catch!

nerrad: Fixed in 08aa22b

nerrad:
Ready for more reviews, I've addressed all the feedback.  

Also after merge I plan on going through and creating issues/pulls for all the comments punting things for iteration on followups.

nerrad:
So oddly:

- the travis-ci/pr build passes.
- the travis-ci/push build fails on phpunit tests (because woocommerce doesn't load - or because the wrong version of phpunit is being used).

This appears to be a travis environment thing and not an artifact of this branch (no travis config was changed in this branch) so I'm going to trust the pr build here.

client/analytics/settings/index.js
nerrad: see https://github.com/woocommerce/woocommerce-admin/issues/2943

client/header/activity-panel/panels/orders.js
nerrad: https://github.com/woocommerce/woocommerce-admin/issues/2943

client/header/activity-panel/unread-indicators.js
nerrad: https://github.com/woocommerce/woocommerce-admin/issues/2943

client/analytics/report/products/table-variations.js
nerrad: https://github.com/woocommerce/woocommerce-admin/issues/2943

client/dashboard/profile-wizard/steps/business-details.js
nerrad: https://github.com/woocommerce/woocommerce-admin/issues/2943

client/dashboard/profile-wizard/steps/business-details.js
nerrad: https://github.com/woocommerce/woocommerce-admin/issues/2943

client/dashboard/task-list/tasks/shipping/rates.js
nerrad: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/983

client/header/activity-panel/index.js
nerrad: https://github.com/woocommerce/woocommerce-admin/issues/2943

PR #2919, opened 2019-09-12
Simplified intervals_missing method of TimeInterval class

codemascot:
Simplified `intervals_missing` method of `TimeInterval` class.


psealock:
I will close this one as its a duplicate of https://github.com/woocommerce/woocommerce-admin/pull/2920

codemascot:
@psealock I don't think they are same. :) **Look closely the class name.** This one is for **TimeInterval** class and #2920 is for **DataStore** class. And those classes are in different files. Plus one calls a `static` method by `self` and another calls a native method with `$this`.

psealock:
whoops! Sorry about that @codemascot, reopening...

PR #2921, opened 2019-09-12
Docs: try js apigen

psealock:
Try moving filter constants to a file for apigen. Code to create json object will be in https://github.com/woocommerce/woocommerce-api-docs


timmyc:
Yo @psealock should this remain open, or did you have any plans to pick this back up?

PR #2933, opened 2019-09-19
Reduce size of bundled JS, and overall plugin file size.

jeffstieler:
Addresses this WP.org thread: https://wordpress.org/support/topic/bloat-2/

In researching the issue, I found that `dist/app/index.js` and `dist/embedded/index.js` were essentially the same, save for very little JS and CSS.

Additionally, Webpack was duplicating the source of the WooCommerce Components inside the app build.

This PR seeks to:
* Remove the superfluous `dist/embedded/index.js` file
* Remove the duplicated component code from `dist/app/index.js`

This results in a file size reduction from 3.5MB zipped and 10.5MB unzipped to 1.5MB zipped and 6.6MB unzipped.

Props to @nerrad for help with Webpack resolvers.

- Smoke test every feature, including DevDocs
- Verify that the `wc` global variable has the same structure as on `master`

### Changelog Note:

Performance: reduce JS bundle size.


jeffstieler:
> The `wc` global differs from master in that its missing `wcSettings`, but I'm pretty sure thats unrelated and a rebase will bring that back.

I'm not seeing `wcSettings` under `wc` on `master` either - it's still present as a top-level global.

nerrad:
> I'm not seeing wcSettings under wc on master either - it's still present as a top-level global.

`wcSettings` will always still be present no matter what version is running.
`wc.wcSettings` will only be present (currently) with master of the `woocommerce/woocommerce-blocks` plugin active (and only on routes enqueueing `wc-settings` as a dependency).  Also I have to change the version check in WC/Blocks for loading back-compat I think now that a new version of WC-Admin has been released _without_ the new wcSettings work.

It's kind of a pain for these initial transition points but once we get through these first couple of versions (and I fix the refactors) things should stabilize in terms of what expectations you have.  For now, it's a bit messy to preserve behaviour in all different environments during the transition period.

PR #2940, opened 2019-09-23
Add GitHub pages deploy step to Travis config.

jeffstieler:
This PR seeks to introduce a deploy step on Travis that pushes the built `docs/` folder to the `gh-pages` branch, since it now relies on a symlink to the `packages/components/src` directory.

The deploy should only occur on commits to `master` and on the PHP 7.2 / JS build.

Once this is merged and the first deploy occurs, we can switch the repo settings to build GH pages from the `gh-pages` branch.

Since https://github.com/travis-ci/dpl/pull/989, this _should_ copy the files symlinked. If that doesn't actually occur, this approach will need to be revisited - but we need to commit this to see if it works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,3 +40,15 @@ script:
 branches:
   only:
     - master
+
+deploy:
+  local_dir: docs
+  provider: pages
+  skip_cleanup: true
+  github_token: $GH_PAGES_TOKEN
+  keep_history: true
+  on:
+    condition: $RUN_JS == 1 # only deploy on the PHP 7.2/JS build
+    branch: master
+  target_branch: gh-pages
+  verbose: true


nerrad:
Nifty!  I wasn't aware of the docs for woocommerce admin until just now and I think this is awesome to have up. Gutenberg is just wrapping up implementing [docs using Storybook](https://github.com/WordPress/gutenberg/pull/17475).  One thing I like about that approach is the live component views (which also verify they are working).  I wonder if that's something we should consider implementing at somep oint as well?

jeffstieler:
Our live components live under WooCommerce > DevDocs (when running `npm start` from the repo).

It's probably best long term to just do it the Gutenberg Way ™️.

PR #2942, opened 2019-09-23
Fix pull request branch detection on CI

jeffstieler:
Should fix the build on the `master` branch - problem introduced in https://github.com/woocommerce/woocommerce-admin/pull/2941

### DIFF
--- a/bin/install-wp-tests.sh
+++ b/bin/install-wp-tests.sh
@@ -163,7 +163,7 @@ install_deps() {
 	REPO=$TRAVIS_REPO_SLUG
 	WORKING_DIR="$PWD"
 
-	if [ "$TRAVIS_PULL_REQUEST_BRANCH" != "" ]; then
+	if [ "$TRAVIS_PULL_REQUEST_BRANCH" ]; then
 		BRANCH=$TRAVIS_PULL_REQUEST_BRANCH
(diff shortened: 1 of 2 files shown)


bin/install-wp-tests.sh
```diff
@@ -188,7 +188,7 @@ install_deps() {
 	cd "$WP_CORE_DIR"
 	php wp-cli.phar plugin activate woocommerce
 
-	if [ "$TRAVIS_PULL_REQUEST_BRANCH" != "" ]; then
+	if [ "$TRAVIS_PULL_REQUEST_BRANCH" ]; then
```
psealock: I'm not sure the syntax, but should have an AND statement?

```
if [ "$TRAVIS_PULL_REQUEST_BRANCH" && "$TRAVIS_PULL_REQUEST_BRANCH" != "" ];
```

jeffstieler: It doesn't need an `AND` - it essentially turns into either `if [ "" ]; then` for pushes to `master`, and `if [ "name/of-the-branch" ]; then` for pull requests.

jeffstieler: You can see what the problem is here: https://travis-ci.org/woocommerce/woocommerce-admin/jobs/588609339#L567

PR #2946, opened 2019-09-24
0.19.0 release changes

psealock:
0.19.0 release notes and updates

### DIFF
--- a/packages/components/src/scroll-to/README.md
+++ b/packages/components/src/scroll-to/README.md
@@ -1,0 +1,15 @@
+ScrollTo
+===
+
+
+
+## Usage
+
+```jsx
+```
+
+### Props
+
+Name | Type | Default | Description
+--- | --- | --- | ---
+`offset` | String | ``'0'`` | The offset from the top of the component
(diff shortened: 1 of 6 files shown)


packages/components/src/scroll-to/README.md
```diff
@@ -0,0 +1,15 @@
...
+## Usage
+
+```jsx
+```
+
+### Props
+
+Name | Type | Default | Description
+--- | --- | --- | ---
+`offset` | String | ``'0'`` | The offset from the top of the component
```
nerrad: This file seems incomplete, is it something that gets filled out in later iterations?

rrennick: There was an unmerged PR for that #2873

psealock:
Thanks @nerrad 

>I would suggest for readme.txt that you just point changelog notes to the github repo. It saves a step in the release process (and keeps the size of the readme.txt from growing too large over time).

Great idea!

>are version bumps in php files manually done?

Nope, automated already by pre-release script

PR #2949, opened 2019-09-24
Fix pull request branch detection on CI (try #4)

jeffstieler:
Update the "is PR" check using the example from the Travis docs: https://docs.travis-ci.com/user/job-lifecycle#complex-build-commands


bin/install-wp-tests.sh
```diff
@@ -163,8 +163,7 @@ install_deps() {
 	REPO=$TRAVIS_REPO_SLUG
 	WORKING_DIR="$PWD"
 
-
-	if [ ! -z "$TRAVIS_PULL_REQUEST_BRANCH" ]; then
+	if [ "${TRAVIS_PULL_REQUEST}" != "false" ]; then
```
nerrad: hmm worth a try, but I wonder if this will need to be:

```
if [ "${TRAVIS_PULL_REQUEST}" != false ];
```

would it be easier to just to a truthy check instead?

jeffstieler: I've tried just a truthy check with `TRAVIS_PULL_REQUEST_BRANCH` and it still failed, so this time I'm taking the example directly and just changing the `=` to `!=` in hopes it will actually work this time.

jeffstieler:
Found more to work on, closing.

PR #2899, opened 2019-09-06
Add ability to receive an emailed download link to report export endpoint.

jeffstieler:
Fixes #311.

This PR seeks to introduce an `email` parameter to the Report Export REST endpoint. When a boolean `true` is passed to the endpoint, an action is scheduled to email a link to download the report to the requesting user.

### Screenshots:

Screen Shot 2019-09-06 at 10 28 00 AM

### Detailed test instructions:

- `POST /wc/v4/reports/products/export` with body: `{ "email": true }`
- Verify a `woocommerce_admin_email_report_download_link` scheduled action exists (Tools > Scheduled Actions)
- When 100%, verify the email associated with your user account has received an email with a link to download the export
- Verify the download link in the email works

### Changelog Note:

Enhancement: add option to email a download link when exporting reports.

PR #2951, opened 2019-09-24
Initiate report export from download button

### LINKED ISSUE #311: Analytics: CSV Download

jeffstieler:
Fixes #311 in part.

This PR seeks to initiate background report exports from the report table download button, if the report is more than one page.

While the success notice says you'll receive an email - that won't be true until #2899 is merged.

### Screenshots

Screen Shot 2019-09-24 at 2 07 33 PM

- Open a report with more only 1 page of results
- Click download
- Verify a .csv file is immediately downloaded
- Open a report with more only 1 page of results
- Click download
- Verify a notice is displayed stating you'll receive an email
- Check your `wp-content/uploads` dir for the .csv file

### Changelog Note:

N/A - Changelog can wait until this is fully hooked up.

### DIFF
--- a/client/analytics/components/report-table/index.js
+++ b/client/analytics/components/report-table/index.js
@@ -176,6 +176,7 @@ class ReportTable extends Component {
 		const params = Object.assign( {}, query );
 		const { items, query: reportQuery } = tableData;
 		const { data, totalResults } = items;
+		let downloadType = 'browser';
 
 		// Delete unnecessary items from filename.
 		delete params.extended_info;
@@ -189,10 +190,15 @@ class ReportTable extends Component {
 				generateCSVDataFromTable( getHeadersContent(), getRowsContent( data ) )
 			);
 		} else {
+			downloadType = 'email';
 			initiateReportExport( endpoint, title, reportQuery );
 		}
 
-		recordEvent( 'analytics_table_download', { report: endpoint, rows: totalResults } );
+		recordEvent( 'analytics_table_download', {
+			report: endpoint,
+			rows: totalResults,
+			downloadType,
+		} );
 	}
 
 	onCompare() {
--- a/client/analytics/components/report-table/index.js
+++ b/client/analytics/components/report-table/index.js
@@ -167,14 +167,15 @@ class ReportTable extends Component {
 			endpoint,
 			getHeadersContent,
 			getRowsContent,
+			initiateReportExport,
 			query,
 			searchBy,
 			tableData,
 			title,
-			totalRows,
 		} = this.props;
 		const params = Object.assign( {}, query );
-		const { items } = tableData;
+		const { items, query: reportQuery } = tableData;
+		const { data, totalResults } = items;
 
 		// Delete unnecessary items from filename.
 		delete params.extended_info;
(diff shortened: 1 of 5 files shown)


client/analytics/components/report-table/index.js
```diff
@@ -167,29 +167,32 @@ class ReportTable extends Component {
...
 				generateCSVFileName( title, params ),
-				generateCSVDataFromTable( getHeadersContent(), getRowsContent( items.data ) )
+				generateCSVDataFromTable( getHeadersContent(), getRowsContent( data ) )
 			);
+		} else {
+			initiateReportExport( endpoint, title, reportQuery );
 		}
 
-		recordEvent( 'analytics_table_download', { report: endpoint, rows: totalRows } );
+		recordEvent( 'analytics_table_download', { report: endpoint, rows: totalResults } );
```
psealock: Should we add a parameter to describe what type, ie in-browser or email?

jeffstieler:
> I did find some differences between the browser download and the to-be-emailed version. The columns were different (I only checked Orders Report) and the naming scheme was different too. Apologies if you are aware already as those seem unrelated to the functionality brought in here.

Ah, nice catch - I forgot about these differences! I feel like the ideal export data is somewhere between the two. Let's open an issue to track fixing that.

jeffstieler:
> This is okay as is, but I wonder if there's value in extracting these specialized triggers to a hoc that exposes the callback to use in the ui?
> 
> That would abstract much of that logic away so it can be wrapped in more than one view for triggering context aware exports.

I can see the value in doing this if/when we add more `ReportTable`s to the app. Are you OK with tracking this in an issue for future development, or do you feel strongly about getting this in sooner than later?

nerrad:
> I can see the value in doing this if/when we add more ReportTables to the app. Are you OK with tracking this in an issue for future development, or do you feel strongly about getting this in sooner than later?

In retrospect I see that `ReportTables` is _already_ an abstraction.  So ya I don't think we need to move on this right now.

client/analytics/components/report-table/index.js
jeffstieler: Added in https://github.com/woocommerce/woocommerce-admin/pull/2951/commits/094f116d3fc4e01ee3d9dfa78c55f1a2fa1dc039

PR #2964, opened 2019-09-25
Fix `npm run build` resulting in non functional app.

nerrad:
## TLDR;

There's a bug with the default `WC_ADMIN_PHASE` value that ends up configuring the app to not have any pages registered.  This pull:

- introduces a new `build:plugin` script.
- uses that for the `build` script instead of `build:core`.

Once the `core` build is put in use, we can re-evaluate what a "production" build means, but for now the assumption is that it is equivalent to a plugin release from the standpoint of active features.

So... possible follow-ups:

- is `core` needed as a configuration option?  If not, maybe remove in a subsequent pull.
- if it is needed, then evaluate what are better defaults than everything off.

## How this was discovered (or Alice in Wonderland)

Today I decided it'd be a good idea to execute `npm run build` and verify that the pulls I did worked as expected in a production build.  This led to a good old fashioned rabbit-hole (see p1569440562154300-slack-CBKK5KM41):

- App didn't load after running `npm run build`
- I tried the usual reset trick `rm -rf node_modules` and even `npm run clean` before `npm run build`.  Still no dice.
- Pinged team in slack
- I tried completely wiping my repo and re-installing.
- I tried doing the same on tagged versions (`0.18.0`, `0.17.0`, `0.16.0`, `0.15.0`), no dice.
- I _verified_ at least that the release build (plugin zip) worked. It did 😌 
- At first @jeffstieler said it wasn't working for him, but then he said doing the `node_modules` dump ended up working for him.  But still not for me!!

So, I ended up taking a bit of a break before coming back and doing some debugging and I ended up discovering that there were no pages being registered for the router (`getPages()` was returning an empty array) which in turn led me to the Feature Flag system.  It turned out if the`WC_ADMIN_PHASE` environment variable is undefined, then it defaults to `core`.  Here's what the configuration is for the `core` phase:

```json
{
	"features": {
		"activity-panels": false,
		"analytics": false,
		"analytics-dashboard": false,
		"analytics-dashboard/customizable": false,
		"devdocs": false,
		"onboarding": false,
		"store-alerts": false
	}
}
```

🤔 seems like that means nothing will be active then right?  Right.  That was the problem.

What's puzzling to me is **how was this ever working for others**?  I'm _guessing_ it's possible folks have exported `WC_ADMIN_PHASE` to an explicit value in their environment, but apart from that 🤷‍♂

## Testing

- run `npm run build`
- verify app works!  💰 
- run `npm run dev`
- verify app works with additional features (eg `onboarding`) active. 🏃


psealock:
>What's puzzling to me is how was this ever working for others?

I think everyone is using dev build locally and the release was always setting the variable to `plugin`:

https://github.com/woocommerce/woocommerce-admin/blob/61dd17677267b017a43f043c0b4f38b104df588a/package.json#L27

I think the intention of `build:core` script was a placeholder for building the asset for its inclusion into core, which we'll need to revisit.

Should we introduce `build:plugin` script in this PR and have the `build` script use it? That might prevent confusion because `npm run build` should build a production version, ie minified etc, but not necessarily the build intended to be gathered by core via composer.

nerrad:
Ideally, the default for `npm run build` should be a usable "production" build.  The question then becomes "What is production"? Right now we have essentially two builds:

1. Development - Bleeding edge, all features.
2. Plugin - Plugin release, only features turned on for release.

Essentially `plugin` serves as "production" right now.

When there becomes a build for what gets bundled in core, won't that just be whatever is configured for `plugin` when the merge happens?  I realize that _after_ merge, in further releases of just the feature plugin, it will vary, but woo core would still be pinned to a specific version of the feature plugin right?

I'm wary about having three configurations unless there's demonstrated need for them.  Every extra configuration means there's something else to remember in a release process and increases the potential for human error.

psealock:
>When there becomes a build for what gets bundled in core, won't that just be whatever is configured for plugin when the merge happens?

Seems likely.

>I realize that after merge, in further releases of just the feature plugin, it will vary, but woo core would still be pinned to a specific version of the feature plugin right?

Indeed it will. Core will gather the desired version via composer, however there may be a scenario where a feature will be turned on in development and plugin for those choosing to get the latest, but remain off in Core. I don't know of anything fitting that description at the moment, but that was the idea of having a core feature flag config.

nerrad:
> I don't know of anything fitting that description at the moment, but that was the idea of having a core feature flag config.

So what's the next steps here then?  Since there is no demonstrated _need_ for a core feature flag config at the moment, is that something that can be addressed when the need is present?  In the meantime, is it safe to work with what's been proposed in this pull (since it arguably works with the configurations that ARE being used right now)?  As it is `npm run build` is broken. 

I think this pull gives a way forward for now (since "production" is effectively the "plugin" build) and we can adjust once we have demonstrated need for a third "core" config build?

nerrad:
BTW I think the work done for controlling features in this system rocks :), so nothing here is intended to downplay that work.  I just think we need to make sure `npm run build` works :)

nerrad:
> So what's the next steps here then?

I just re-read your original post and I think I understand what you are suggesting we do better now.  I still think it's kind of early to introduce a configuration that isn't being used but I can re-work this pull.

nerrad:
Okay @psealock I implemented your suggestion and updated the original post in this pull with the changes.  Thoughts?

nerrad:
> Right, that may have been solving a problem that hasn't arisen yet. Is it worth removing in a follow up PR?

I'm on the fence.  With this fix in place it's not harming anything and will save having to add it back in.  I think I understand better the scenario you describe for its intended purpose.

PR #2968, opened 2019-09-26
Change usage tracking prompt to modal

### LINKED ISSUE #2934: Change usage tracking auth message in current onboarding
Accordingly to the wp.org guidelines, the usage tracking authorization message needs to be opted-out by default. So we need to make the following changes in the Store Setup in the current onboarding:

- Remove the usage track auth message: https://cld.wthms.co/49Ypwm
- When the user click's "Let's go" we then display the following dialog with the checkbox opted-out by default: https://www.figma.com/file/b0b2CrxPZz0i1aAPuv985tvK/Store-Profiler?node-id=3154%3A21
- With this change, the event `wcadmin_obw_store_setup` can only be triggered when the user opts-in in the checkbox and clicks "Continue"

@justinshreve, I would be great to have this deployed on 3.8. Do you think you can make the changes before Oct 1? Thanks.

### LINKED ISSUE #2965: Onboarding: change usage tracking auth message in the new onboarding
Accordingly to the wp.org guidelines, the usage tracking authorization message needs to be opted-out by default. So we need to make some changes in the new onboarding:

- In the [Welcome Screen](https://www.figma.com/file/b0b2CrxPZz0i1aAPuv985tvK/Store-Profiler?node-id=1250%3A141) we need to remove the checkbox and copy above the button and add the following copy: By connecting your site you agree to our fascinating [Terms of Service](https://en.wordpress.com/tos/) and to share details with WordPress.com
- When the user clicks on "Get started" or on "Proceed without..." we display a modal with the usage tracking auth message ([mockup](https://www.figma.com/file/b0b2CrxPZz0i1aAPuv985tvK/Store-Profiler?node-id=3152%3A25))
- The event tracking `wcadmin_storeprofiler_welcome_clicked` can only be triggered if the user authorizes the usage tracking in the modal (pls confirm if we're able to get the 2 event properties, considering that with this change we can only track after the user clicks on the checkbox and on continue)
- We also need to consider the use case we're the merchant skips that initial screen because he already has Jetpack and/or WCS activated. In that scenario, we need to display the modal after they click on "Continue" in the first screen they see, which is the [Store Details](https://www.figma.com/file/b0b2CrxPZz0i1aAPuv985tvK/Store-Profiler?node-id=1150%3A732)

justinshreve:
Closes #2965.

This PR changes the usage tracking checkbox to a modal per the designs in #2934. It also makes it opted-out by default (though if you have run through the wizard before, it is defaulted to your previous selection).

It also detects if the Jetpack/WCS were skipped (because the extensions were already installed), and instead shows the modal on the store details page instead. See p1565365006096800-slack-wc-onboarding

I was getting some linter errors unrelated to this PR, so I've fixed a few of those up here to get tests to pass.

### Screenshots

Screen Shot 2019-09-26 at 2 17 34 PM

Screen Shot 2019-09-26 at 2 29 18 PM

* Clear out the `woocommerce_allow_tracking` option from your database.
* Testing with Jetpack or WooCommerce Services disabled so that you see the start screen.
* Click `Get Started` on the start screen. The usage tracking dialog should display. It should default to off.
* Try selecting the box and hitting continue. Verify the option updates in the database.
* Go to store settings and verify the continue button there goes to the industry step.
* Test again with the "Skip WooCommerce Services/Jetpack" link. You should still see the dialog. It should default to your previous selection.
* Enable WooCommerce Services AND Jetpack, enable the profile wizard, and go back to test. You should automatically end up on the store details page. Fill out your details, and click continue. You should see the auth dialog here instead. Verify your address updates, and the option updates in the database.

### DIFF
--- a/client/dashboard/profile-wizard/steps/usage-modal.js
+++ b/client/dashboard/profile-wizard/steps/usage-modal.js
@@ -1,0 +1,143 @@
+/** @format */
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { Button, CheckboxControl } from 'newspack-components';
+import { Component } from '@wordpress/element';
+import { compose } from '@wordpress/compose';
+import { withDispatch } from '@wordpress/data';
+import { get } from 'lodash';
+import interpolateComponents from 'interpolate-components';
+import { FormToggle, Modal } from '@wordpress/components';
+
+/**
+ * Internal depdencies
+ */
+import { Link } from '@woocommerce/components';
+import withSelect from 'wc-api/with-select';
+
+class UsageModal extends Component {
+	constructor( props ) {
+		super( props );
+		this.state = {
+			allowTracking: props.allowTracking,
+		};
+
+		this.onTrackingChange = this.onTrackingChange.bind( this );
+	}
+
+	onTrackingChange() {
+		this.setState( {
+			allowTracking: ! this.state.allowTracking,
+		} );
+	}
+
+	async componentDidUpdate( prevProps ) {
+		const { hasErrors, isRequesting } = this.props;
+		const isRequestSuccessful = ! isRequesting && prevProps.isRequesting && ! hasErrors;
+		const isRequestError = ! isRequesting && prevProps.isRequesting && hasErrors;
+
+		if ( isRequestSuccessful ) {
+			this.props.onClose();
+			this.props.onContinue();
+		}
+
+		if ( isRequestError ) {
+			this.props.createNotice(
+				'error',
+				__( 'There was a problem updating your preferences.', 'woocommerce-admin' )
+			);
+			this.props.onClose();
+		}
+	}
+
+	updateTracking() {
+		const { updateOptions } = this.props;
+		const allowTracking = this.state.allowTracking ? 'yes' : 'no';
+		updateOptions( {
+			[ 'woocommerce_allow_tracking' ]: allowTracking,
+		} );
+	}
+
+	render() {
+		const { allowTracking } = this.state;
+		const { isRequesting } = this.props;
+		const trackingMessage = interpolateComponents( {
+			mixedString: __(
+				"Learn more about how usage tracking works, and how you'll be helping {{link}}here{{/link}}.",
+				'woocommerce-admin'
+			),
+			components: {
+				link: (
+					<Link href="https://woocommerce.com/usage-tracking" target="_blank" type="external" />
+				),
+			},
+		} );
+
+		return (
+			<Modal
+				title={ __( 'Help improve WooCommerce with usage tracking', 'woocommerce-admin' ) }
+				onRequestClose={ () => this.props.onClose() }
+				className="woocommerce-profile-wizard__usage-modal"
+			>
+				<div className="woocommerce-profile-wizard__usage-wrapper">
+					<div className="woocommerce-profile-wizard__usage-modal-message">{ trackingMessage }</div>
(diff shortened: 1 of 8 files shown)


phpcs.xml.dist
```diff
@@ -27,6 +27,12 @@
 		<exclude-pattern>tests/</exclude-pattern>
 	</rule>
 
+	<rule ref="PHPCompatibility.FunctionUse.NewFunctionParameters.ucwords_delimitersFound">
```
justinshreve: This error is

> The function ucwords() does not have a parameter "delimiters" in PHP version 5.4.31 or earlier

However, WordPress requires `5.6.20+` and WooCommerce lists `7.2+` so I think this is something we can ignore.

mikejolley: Woo core still supports 5.6 too fwi. 7.2 is recommended version. You should be able to change the testVersion in the main phpcs file to reflect this. It's 5.4 currently.

justinshreve: 🤦‍♂. Yeah updating the version is much better. Thanks!

justinshreve:
> I couldn't see where this button is?

Sorry - I meant to say the store details page of the onboarding wizard, NOT store settings. If you have WooCommerce Services & Jetpack installed/activated and you visit `/wp-admin/admin.php?page=wc-admin`, you should automatically be taken to it.

> Styling wise, I think the vertical alignment of the modal checkbox is slightly off but I'm using Firefox :)

Can you provide a screenshot of this and your FF version? I just tried in the latest FF on Mac `69.0.1` and the checkbox looked OK to me. There was a slight issue with the continue button which I fixed.

justinshreve:
I've rebased this branch and tested everything again, including w/ Firefox.?

Can I get another review to help get this through? cc @joshuatf

client/dashboard/profile-wizard/style.scss
```diff
@@ -311,3 +315,31 @@
...
+	.woocommerce-profile-wizard__usage-wrapper {
+		flex-grow: 1;
+		display: flex;
+		flex-direction: column;
+
+		a {
+			color: $studio-gray-60;
+		}
+
+		button.is-primary {
```
joshuatf: Could we remove the bottom margin on this button?  Looks like a bit more spacing on the bottom of this modal because of the button margin.

client/dashboard/profile-wizard/steps/usage-modal.js
```diff
@@ -0,0 +1,143 @@
...
+		} );
+	}
+
+	async componentDidUpdate( prevProps ) {
+		const { hasErrors, isRequesting } = this.props;
+		const isRequestSuccessful = ! isRequesting && prevProps.isRequesting && ! hasErrors;
+		const isRequestError = ! isRequesting && prevProps.isRequesting && hasErrors;
+
+		if ( isRequestSuccessful ) {
+			this.props.onClose();
```
joshuatf: Nitpick- we could destructure these props above.

client/dashboard/profile-wizard/steps/usage-modal.js
```diff
@@ -0,0 +1,143 @@
...
+			);
+			this.props.onClose();
+		}
+	}
+
+	updateTracking() {
+		const { updateOptions } = this.props;
+		const allowTracking = this.state.allowTracking ? 'yes' : 'no';
+		updateOptions( {
+			[ 'woocommerce_allow_tracking' ]: allowTracking,
```
joshuatf: Optional nitpick- brackets can be removed here.

PR #2970, opened 2019-09-27
Use version number for cache invalidation

### LINKED ISSUE #2020: Invalidate cache on update via cache version number
The current use of `wp_cache_get` and `wp_cache_set` will only last the duration of the original request.  This means that hitting the API multiple times with the same request won't actually benefit from this caching.

This may be the intended behavior, but since some of these queries are somewhat complex, should we consider setting data in a transient?

This would improve speed quite a bit, but because of how many different queries could be performed, this may result in many transients and possibly better left to third party caching plugins.

jeffstieler:
Fixes #2020.

This PR seeks to do the following:
- Use transients to cache report data store queries
- Invalidate caches using a version number (incremented when data changes)

### Question:

Are there other places we need to invalidate cache?

- Disable any persistent object cache plugins (makes testing easier)
- Open any Report
- Verify a transient containing `wc_report_[report name]` exists with a version number at the end (`:1`)
- Modify a user, order, or coupon
- Verify the `woocommerce_reports_cache_version` option value changes
- Reload the report
- Verify a new transient is created with the new version number appended

### Changelog Note:

Performance: add caching layer to analytics.

### DIFF
--- a/src/API/Reports/Cache.php
+++ b/src/API/Reports/Cache.php
@@ -20,38 +20,22 @@ class Cache {
 	/**
 	 * Cache version. Used to invalidate all cached values.
 	 */
-	const VERSION_OPTION = 'woocommerce_reports_cache_version';
+	const VERSION_OPTION = 'woocommerce_reports';
 
 	/**
-	 * Default version number rollover threshold.
+	 * Invalidate cache.
 	 */
-	const DEFAULT_VERSION_LIMIT = 99999;
-
-	/**
-	 * Increase cache version number.
-	 *
-	 * @return bool
-	 */
-	public static function bump_version() {
-		/**
-		 * Upper limit to roll cache version back to zero.
-		 *
-		 * @param int $limit Limit.
-		 */
-		$limit   = apply_filters( 'woocommerce_reports_cache_version_limit', self::DEFAULT_VERSION_LIMIT );
-		$limit   = is_numeric( $limit ) ? (int) $limit : self::DEFAULT_VERSION_LIMIT;
-		$version = ( self::get_version() + 1 ) % $limit;
-
-		return update_option( self::VERSION_OPTION, $version );
+	public static function invalidate() {
+		\WC_Cache_Helper::get_transient_version( self::VERSION_OPTION, true );
 	}
 
 	/**
 	 * Get cache version number.
 	 *
-	 * @return int
+	 * @return string
 	 */
 	public static function get_version() {
-		$version = (int) get_option( self::VERSION_OPTION, 1 );
+		$version = \WC_Cache_Helper::get_transient_version( self::VERSION_OPTION );
 
 		return $version;
 	}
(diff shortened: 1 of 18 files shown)


src/API/Reports/Cache.php
```diff
@@ -0,0 +1,85 @@
...
+
+	/**
+	 * Get cached value.
+	 *
+	 * @param string $key Cache key.
+	 * @return mixed
+	 */
+	public static function get( $key ) {
+		$version = self::get_version();
+		$value   = get_transient( $key . ':' . $version );
```
mikejolley: This was something I brought up at the GM with @psealock but realised you missed this :)

Using version numbers in transient names causes issues down the line because the transient name is dynamic. It makes cleanup difficult, and leads to lots of stored transient data if those values are not cleaned up regulary.

There is a simple workaround for this. Store the version **within the transient**. 

i.e. the transient data becomes an array like this:

```
[
'version' => X,
'data' => '...original data here...',
]
```

Instead of then reading a transient named `$key . ':' . $version`, you read a transient named `$key` and then check the version that was stored. If it differs, regenerate the data (its invalid). If the version matches, we can return the data (its valid).

src/API/Reports/Cache.php
```diff
@@ -0,0 +1,85 @@
...
+		/**
+		 * Upper limit to roll cache version back to zero.
+		 *
+		 * @param int $limit Limit.
+		 */
+		$limit   = apply_filters( 'woocommerce_reports_cache_version_limit', self::DEFAULT_VERSION_LIMIT );
+		$limit   = is_numeric( $limit ) ? (int) $limit : self::DEFAULT_VERSION_LIMIT;
+		$version = ( self::get_version() + 1 ) % $limit;
+
+		return update_option( self::VERSION_OPTION, $version );
```
mikejolley: This could probably be simplified by using a hash based on timestamp? If the version mismatches the stored data, it's invalidated. It doesn't necessarily need to be a number that increments.

src/API/Reports/Cache.php
```diff
@@ -0,0 +1,85 @@
...
+	 * Default version number rollover threshold.
+	 */
+	const DEFAULT_VERSION_LIMIT = 99999;
+
+	/**
+	 * Increase cache version number.
+	 *
+	 * @return bool
+	 */
+	public static function bump_version() {
```
mikejolley: Would `::invalidate` be clearer?

src/API/Reports/Cache.php
```diff
@@ -0,0 +1,85 @@
...
+	/**
+	 * Update cached value.
+	 *
+	 * @param string $key   Cache key.
+	 * @param mixed  $value New value.
+	 * @return bool
+	 */
+	public static function set( $key, $value ) {
+		$version = self::get_version();
+		$result  = set_transient( $key . ':' . $version, $value, DAY_IN_SECONDS );
```
mikejolley: See notes above on transient naming. We should keep a single $key per dataset. Since things are invalidated inline, the cache duration can be increased too to provide more hits. Month?

jeffstieler:
@mikejolley - addressed your feedback in 89d4ef5.

src/API/Reports/Cache.php
jeffstieler: Went with a single cache version group for the initial implementation, and increased the timeout to 1 week.

PR #2974, opened 2019-09-30
wcSettings: fix countries getSettings

psealock:
`countries` setting was being improperly gathered from the top level of `wcSettings` instead of from the property `dataEndpoints`. This PR fixes that and adds defaults for `dataEndpoints`.

### Before

Screen Shot 2019-09-27 at 11 11 28 AM

### After

The countries dropdown populates with options.

@nerrad Does the blocks repo require a similar change?

### DIFF
--- a/client/analytics/report/customers/config.js
+++ b/client/analytics/report/customers/config.js
@@ -5,7 +5,9 @@
 import { __, _x } from '@wordpress/i18n';
 import { decodeEntities } from '@wordpress/html-entities';
 import { applyFilters } from '@wordpress/hooks';
-import { COUNTRIES as countries } from '@woocommerce/wc-admin-settings';
+import { getSetting } from '@woocommerce/wc-admin-settings';
+
+const { countries } = getSetting( 'dataEndpoints', { countries: {} } );
 
 /**
  * Internal dependencies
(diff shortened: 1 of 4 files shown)


nerrad:
I pushed an update that reverts the original fix and demonstrates how the countries value exposed via `wcSettings.dataEndpoints.countries` should be accessed going forward.   As I mentioned in my initial comment, I made the incorrect assumption that `wcSettings.countries` (which is built by wc-blocks) is the same as `wcSettings.dataEndpoints.countries` (which is added to the wcSettings global by wc-admin) are the same thing.  They aren't.

`wcSettings.countries` is exported on `COUNTRIES` as a part of the `@woocommerce/settings` (`wc-settings` handle) bundle in WC-Blocks.  There it is a map of country-code to country label and that's all wc-blocks needed.

`wcSettings.dataEndpoints.countries` is enhanced on the `wcSettings` global via hooking into the data api in `WC-Blocks`.  It is accessed via using the `getSetting` api (but feasibly you could do a constant export if you wanted in wc-admin, but not necessary).  It differs in that it is a set of countries (grouped in a sets of 100) with each object representing various country details (including endpoint links for the countries endpoint in v4 of the api).

 So there may be a few places in my refactor where I incorrectly replaced usage of `wcSettings.dataEndpoints.countries` with `COUNTRIES` (which is a constant exported aliasing `wcSettings.countries`).

There are a few places where just the `code: country label` map was used so those should be okay.

psealock:
Thanks for the refactor @nerrad 

>So there may be a few places in my refactor where I incorrectly replaced usage of wcSettings.dataEndpoints.countries with COUNTRIES (which is a constant exported aliasing wcSettings.countries).

I quickly did a search and found instances in the Customers Report and went ahead and tidied those up too, https://github.com/woocommerce/woocommerce-admin/pull/2974/commits/86de040ad375679cf11c50614313cf1f99bb7fcc

PR #2975, opened 2019-09-30
wcSettings refactor: fix camel case for currencies in PHP

psealock:
https://github.com/woocommerce/woocommerce-admin/pull/2917 introduced changes to property names in currency settings, causing the following PHP error logs:

Screen Shot 2019-09-30 at 2 37 06 PM

core function `wc_admin_number_format` needed to be updated. I searched the codebase for instances of the snake case and these changes should be the only ones.

### Test

1. Load the dashboard.
2. Ensure no PHP error occur.

### DIFF
--- a/includes/core-functions.php
+++ b/includes/core-functions.php
@@ -20,8 +20,8 @@ function wc_admin_number_format( $number ) {
 	return number_format(
 		$number,
 		0,
-		$currency_settings['decimal_separator'],
-		$currency_settings['thousand_separator']
+		$currency_settings['decimalSeparator'],
+		$currency_settings['thousandSeparator']
 	);
 }
 


nerrad:
Seems to be duplication happening with this kind of data.

- `Automattic\WooCommerce\Blocks\AssetDataRegistry::get_core_data`
- `Automattic\WooCommerce\Admin::get_currency_settings()`

The danger here is that both are filtered separately which makes it relatively easy for extensions to get the values in differents states...

I think it'd be a good idea to pick one place as the authority for currency settings (since Blocks is merged, probably Blocks).  I think things like Currency and Money would be good candidates for Value Objects.  That way the shape of "Currency" is not ambiguous, but defined, which in the case of commerce is pretty important.

Thoughts?

psealock:
>I think it'd be a good idea to pick one place as the authority for currency settings (since Blocks is merged, probably Blocks)

Definitely. Once wc-admin is merged, it would make sense to keep the logic in wc-admin files instead of Blocks. I'll go ahead and merge this for now.

PR #2981, opened 2019-10-01
Activity Panel: Catch notices earlier than priority 10

justinshreve:
The way we are catching notices for the activity panel only catches notices between `priority 11` and `PHP_INT_MAX`. This means notices hooked on the default priority of 10 are not caught, and anything hooked earlier than that isn't caught. This PR hooks into a lower priority. 

Before applying this PR, add the following code to your install:

```
add_action( 'admin_notices', function() {
	echo 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed lectus orci, interdum id risus ut, rhoncus lacinia dolor. Integer condimentum enim dapibus tellus suscipit, sed porta dui eleifend. Suspendisse vel porta risus, at tempor dui. Aliquam pellentesque consectetur tincidunt. Vestibulum non posuere felis. Phasellus sagittis sit amet risus at laoreet. Vestibulum malesuada, lacus eu malesuada porttitor, ante turpis laoreet ex, et dignissim enim est sed purus. Sed bibendum faucibus justo, in sodales ligula cursus sit amet. Fusce sed vestibulum diam, et aliquam mauris.
	Etiam auctor metus eros. Aliquam qui';
}, 10 );
```

* Load a WooCommerce Admin page
* See that the text flashes on first load, and does not show up under the notices panel
* Apply this PR.
* See that the notice is now under the notice panel.

### Changelog Note:

Fix: Catch notices that are hooked earlier than priority 10.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -49,12 +49,12 @@ class Loader {
 	 */
 	public function __construct() {
 		add_action( 'init', array( __CLASS__, 'load_features' ) );
-;		add_action( 'admin_enqueue_scripts', array( __CLASS__, 'register_scripts' ) );
+		add_action( 'admin_enqueue_scripts', array( __CLASS__, 'register_scripts' ) );
 		add_action( 'admin_enqueue_scripts', array( __CLASS__, 'inject_wc_settings_dependencies' ), 14 );
 		add_action( 'admin_enqueue_scripts', array( __CLASS__, 'load_scripts' ), 15 );
-		// old settings injection
+		// Old settings injection.
 		add_filter( 'woocommerce_components_settings', array( __CLASS__, 'add_component_settings' ) );
-		// new settings injection
+		// New settings injection.
 		add_filter( 'woocommerce_shared_settings', array( __CLASS__, 'add_component_settings' ) );
 		add_filter( 'admin_body_class', array( __CLASS__, 'add_admin_body_classes' ) );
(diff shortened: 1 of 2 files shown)


src/Loader.php
```diff
@@ -64,7 +64,7 @@ public function __construct() {
 		add_filter( 'woocommerce_settings_groups', array( __CLASS__, 'add_settings_group' ) );
 		add_filter( 'woocommerce_settings-wc_admin', array( __CLASS__, 'add_settings' ) );
 		add_action( 'admin_head', array( __CLASS__, 'remove_notices' ) );
-		add_action( 'admin_notices', array( __CLASS__, 'inject_before_notices' ) );
+		add_action( 'admin_notices', array( __CLASS__, 'inject_before_notices' ), -9999 );
```
justinshreve: Ideally we would use something like `PHP_INT_MAX`, but this isn't supported in PHP 5.6. I'm not sure how low to go here.

joshuatf: Did you mean `PHP_INT_MIN` here?  I think max is supported.  I'm guessing we could do `PHP_INT_MAX * -1` but that may throw errors on weird edge cases where a server had min/max configured differently.

I think `-9999` will cover most, if not all cases though.

justinshreve: Yeah, I meant `PHP_INT_MIN` here, sorry.

Let's stick with `-9999` for now. If we get reports of something not working we can adjust later.

PR #2987, opened 2019-10-02
Match server-side CSV export format to client-side

### LINKED ISSUE #2980: Server-side CSV download should match client format

The format for in-browser CSV report downloads and those generated on the server-side differ.

Header and row values should conform to the in-browser format. This requires the JS `getRowsContent()` logic to be duplicated in PHP for server-side exports. It may also necessitate a new `getRowsCSVContent()` or similar to be created.

**Additional context**

p90Yrv-1fE-p2

jeffstieler:
Fixes #2980.
Fixes https://github.com/woocommerce/woocommerce-admin/issues/3023

This PR seeks to align the server-side and client-side report export formats by:

* Introducing a `ExportableInterface` which requires `get_export_columns()` and `prepare_item_for_export()` methods to be defined on Controllers
* Allows and escapes commas and double quotes in CSV values
* Uses `WP_REST_Server::response_to_data()` to include embedded objects in responses

One kludge I'm not fond of is the `get_export_items()` check - it's there so we can shape the response of `/revenue/stats` into a single array of values. I'm open to ideas on how to better implement that.

For each report:

- Use date range and filters to get a single-page result set
- Download the report (should be in-browser)
- Use date range and filters to get a multi-page result set
- Download the report (should be server-side, emailed)
- Compare CSV files, they should match

### Changelog Note:

Fix: report export format when generated server-side.

### DIFF
--- a/packages/csv-export/src/index.js
+++ b/packages/csv-export/src/index.js
@@ -5,8 +5,22 @@
 import moment from 'moment';
 import { saveAs } from 'browser-filesaver';
 
+function escapeCSVValue( value ) {
+	let stringValue = value.toString();
+
+	if ( stringValue.match( /[,"\s]/ ) ) {
+		stringValue = '"' + stringValue.replace( /"/g, '""' ) + '"';
+	}
+
+	return stringValue;
+}
+
 function getCSVHeaders( headers ) {
-	return Array.isArray( headers ) ? headers.map( header => header.label ).join( ',' ) : [];
+	return Array.isArray( headers )
+		? headers
+			.map( header => escapeCSVValue( header.label ) )
+			.join( ',' )
+		: [];
 }
 
 function getCSVRows( rows ) {
@@ -18,13 +32,7 @@ function getCSVRows( rows ) {
 							return '';
 						}
 
-						let stringValue = rowItem.value.toString();
-
-						if ( stringValue.includes( ',' ) || stringValue.includes( '"' ) ) {
-							stringValue = '"' + stringValue.replace( /"/g, '""' ) + '"';
-						}
-
-						return stringValue;
+						return escapeCSVValue( rowItem.value );
 					} ).join( ',' )
 				)
 				.join( '\n' )
--- a/packages/csv-export/src/index.js
+++ b/packages/csv-export/src/index.js
@@ -13,9 +13,19 @@ function getCSVRows( rows ) {
 	return Array.isArray( rows )
 		? rows
 				.map( row =>
-					row.map( rowItem =>
-						rowItem.value !== undefined && rowItem.value !== null ? rowItem.value.toString().replace( /,/g, ''
-					) : '' ).join( ',' )
+					row.map( rowItem => {
+						if ( undefined === rowItem.value || null === rowItem.value ) {
+							return '';
+						}
+
+						let stringValue = rowItem.value.toString();
+
+						if ( stringValue.includes( ',' ) || stringValue.includes( '"' ) ) {
+							stringValue = '"' + stringValue.replace( /"/g, '""' ) + '"';
+						}
+
+						return stringValue;
+					} ).join( ',' )
 				)
 				.join( '\n' )
 		: [];
--- a/packages/csv-export/src/index.js
+++ b/packages/csv-export/src/index.js
@@ -8,7 +8,12 @@ import { saveAs } from 'browser-filesaver';
 function escapeCSVValue( value ) {
 	let stringValue = value.toString();
 
-	if ( stringValue.match( /[,"\s]/ ) ) {
+	// Prevent CSV injection.
+	// See: http://www.contextis.com/resources/blog/comma-separated-vulnerabilities/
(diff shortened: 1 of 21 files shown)


psealock:
Here's a review checklist for myself so I can keep track

- [ ] Revenue
- [ ] Orders
- [ ] Products
- [ ] Categories
- [ ] Coupons
- [ ] Taxes
- [ ] Downloads
- [ ] Stock
- [ ] Customers

jeffstieler:
> Some column titles have quotation marks and the timestamp also has quotation marks in the emailed CSV.

There should be no functional difference based on the quotations - that's just a way to encapsulate a column value for CSV if it has any special characters. It looks like the PHP logic see a space and wraps in quotes while the JS doesn't. I can look into changing that behavior, but it shouldn't result in the CSV being interpreted differently by any spreadsheet apps.

> Also, should we review the titles so they match in format?

Yes, but let's save that for another PR - that will require more work/changes for the server side export.. we'll have to store the full filenames of exports in order to accomplish that (we are only storing the type and "id" for now).

packages/csv-export/src/index.js
```diff
@@ -5,17 +5,35 @@
...
 
+function escapeCSVValue( value ) {
+	let stringValue = value.toString();
+
+	if ( stringValue.match( /[,"\s]/ ) ) {
+		stringValue = '"' + stringValue.replace( /"/g, '""' ) + '"';
+	}
+
+	return stringValue;
+}
```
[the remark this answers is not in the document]
psealock: I can reproduce with any negative number.

Screen Shot 2019-10-18 at 1 06 20 PM

src/API/Reports/Products/Controller.php
```diff
@@ -343,4 +345,83 @@ public function get_collection_params() {
...
+	 * Get stock status column export value.
+	 *
+	 * @param array $status Stock status from report row.
+	 * @return string
+	 */
+	protected function _get_stock_status( $status ) {
+		$statuses = wc_get_product_stock_status_options();
+
+		return isset( $statuses[ $status ] ) ? $statuses[ $status ] : '';
+	}
```
psealock: I know this is unrelated to the changes in this PR, but just to note in case there is a quick fix. In-browser download isn't adding status correctly to Products Report. Its working fine for the emailed version.

Screen Shot 2019-10-18 at 1 09 35 PM

psealock: Another note on statuses, this time related to changes here: Products that don't have managed inventory are labelled `N/A` for Status and Stock. The CSV is giving these values of "In Stock" and "0", which isn't correct.

Screen Shot 2019-10-18 at 1 14 04 PM

Screen Shot 2019-10-18 at 1 17 40 PM

psealock: Same thing with Categories Report regarding Statuses and `N/A`.

jeffstieler:
> Did we decide to remove columns that aren't visible in the UI. If so, it would be good for a follow up pull request.

Yeah, we did. I agree it's good for a follow up: https://github.com/woocommerce/woocommerce-admin/issues/3073

packages/csv-export/src/index.js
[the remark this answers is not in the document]
jeffstieler: The apostrophe is to mitigate malicious formula input, see: https://github.com/woocommerce/woocommerce/blob/e007b2317a0acf5810b6ff5b0370a2f309fe6dfa/includes/export/abstract-wc-csv-exporter.php#L354-L378

Supposedly most spreadsheet apps won't render the leading apostrophe, but Numbers on my Mac does.

I think the fix here is to add the apostrophe to the JS CSV export.

src/API/Reports/Products/Controller.php
jeffstieler: > I know this is unrelated to the changes in this PR, but just to note in case there is a quick fix. In-browser download isn't adding status correctly to Products Report.

I'm not able to reproduce that.. even turning off inventory management in settings didn't change the in-browser export.

jeffstieler: > Products that don't have managed inventory are labelled N/A for Status and Stock. The CSV is giving these values of "In Stock" and "0", which isn't correct.

Fixed in f9fc0d61.

> Same thing with Categories Report regarding Statuses and `N/A`.

When viewing a single category, it's actually a product report. Also fixed in f9fc0d61.

psealock: >I'm not able to reproduce that.. even turning off inventory management in settings didn't change the in-browser export. 

I'll reproduce and get more details so I can write a separate issue if need be.

packages/csv-export/src/index.js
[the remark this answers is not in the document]
psealock: Great links in the comments there. I did not know this, but sounds like an appropriate thing to do.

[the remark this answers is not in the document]
jeffstieler: Added CSV injection mitigation to the JS export: 1d8297d.

Issue #2999, opened 2019-10-03
Prioritize wcSettings refactor of @woocommerce/components package


nerrad:
There's a bit of a race condition with the wcSettings refactor that I didn't forsee and mostly has to do with the blocks dependency on the `@woocommerce/components` package which caused this bug: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1017

So... we're held up releasing a new version of the blocks plugin containing the refactor until the refactor of `@woocommerce/components` using the new settings system is complete and a new package is released.

Hence the priority bump and this issue.

PR #457, opened 2018-09-20
i18n: Advanced Filters strings

psealock:
Add translation and context (including link to screenshot) to Advanced Filters such that translators can determine the appropriate translation and the filter is rendered in natural language order.

Relevant discussion in pxLjZ-4T6-p2 and #285 

## Screenshots

English
screen shot 2018-09-25 at 4 29 02 pm

Japanese
screen shot 2018-09-25 at 4 27 28 pm

## Test

1. Go to `/wp-admin/admin.php?page=wc-admin#/analytics/orders?filter=advanced` and poke around the advanced filters to discover any possible regressions.
2. Change the language to something other than English
3. Manually change the translation in the config line 47, https://github.com/woocommerce/wc-admin/blob/f86c16ca7d3af6c2b21c9cf8c3057e52a8fe25f9/client/analytics/report/orders/config.js#L47
to something else (apologies to any Francophiles):
```
{{rule /}} le Status de l'Ordre {{filter /}}
```
4. See the filter rendered accordingly
screen shot 2018-09-25 at 4 21 01 pm
5. Change the values of the filter and see the fieldset's legend update
screen shot 2018-09-25 at 4 38 48 pm
6. Extra Credit: fire up VoiceOver and ensure all legends and labels are available

Fixes #285

PR #3001, opened 2019-10-03
Decouple wcSettings from published packages.

nerrad:
Fixes: #2999

See the issue for background.  This pull originally refactored all packages to use the new `wc-settings-admin` alias but then we realized that packages should _not_ have a coupling to an externally provided global.  This is a gotcha for consumers of packages and ideally any settings/configuration should be provided by consumers of packages.

So with that in mind, I refactored _most_ of the packages to avoid usage of any global (via alias or directly).  The following are some notes around that.

> Note: I did not refactor everything, I made some notes at the end about what remains and some suggested paths to take.  I also did not fix any tests.
> 
I did focused commits where each commit covered a specific component decoupling and I made note of which commit is related to each section below.

## `AdvancedFilters` (6de7e72)

- using `LOCALE.siteLocale`
- refactored to receive `siteLocale` as a prop
- implemented by packages/components/src/filters (ReportsFilters), which has been refactored to receive `siteLocale` as a prop.
- _client_ scripts use `ReportFilters`, so just send `siteLocale` through as a prop from client.
- Docs were referencing `ORDER_STATUSES` from settings so I updated to remove importing from `wc-admin-settings` in the example.  ORDER_STATUSES isn't used anywhere else that I could see in implementation (from the package side).

### Notes (next steps/considerations):

- LOCALE should be something exposed to components via context.  That will reduce the need for passing locale props through.
- The only place siteLocale is used for in `AdvancedFilters` is to derive whether `isEnglish` is true or not.  The only place that is used in children elements is to derive an `is-english` classname for css. Looks like it was added here https://github.com/woocommerce/woocommerce-admin/pull/457/files#r220056210 
- There was a bug with the `isEnglish` function in AdvancedFilters. It didn't account for locales like `en_US` (my site had that).

### Testing

- ensure that `is-english` is still output for `.woocommerce-filters-advanced__fieldset` elements when the siteLocale is English and NOT output when siteLocale is non-english.

## `NumberFilter` (packages/advanced-filters/number-filter) (8e50be1)

- using `CURRENCY` (currencySymbol, symbolPosition).
- now receives `currencySymbol` and `symbolPosition` as props.
- impacts `AdvancedFilters` because that's where this filter is implemented.  So AdvancedFilters needs to receives `currency` as a prop now.
- impacts `ReportFilters` because that is what implements `AdvancedFilters`
- _client_ scripts use `ReportFilters`, so just send `CURRENCY` value through as `currency` prop from client.

### Notes (next steps/considerations)

- CURRENCY should be something exposed to components via context.  That would reduce the need for passing props through various places.
- We should consider implementing a Currency value object for everyting related to currency. This will allow for establishing an interface contract for interacting with currency and type checking it everywhere it's used.

### Testing

- ensure any filters using `NumberFilter` have the currency symbol and position correct.

## `Chart` (components/chart) (6dd7d1d)

- using `CURRENCY` ( symbol, symbolPosition, decimal, thousands )
- now receives `currency` as a prop
- impacts `examples/extensions/dashboard-section/js/global-prices.js`. While the defaults will kick in for Chart, might want to update the example with how to get CURRENCY data from `@woocommerce/settings` (external).

### Notes

- This is another place where passing Currency via context would be useful.
- Same as ^^ regarding implementing Currency as a value object.

### Testing
- ensure anything using the `Chart` component (pretty much all report charts) are correctly implementing currency.

## `EmptyContent` (packages/image-asset) (and `ImageAsset`) (fe8fcab)

- able to completely remove `ImageAsset` component and modify `EmptyContent` to use `` directly while using a embedded svg as the default. This completely decouples settings from being needed.

### Testing

- ensure anything using this component is rendering as expected:
	- generally, if it works as expected for one, it should work as expected for any of them.

## `OrderStatus` (/packages/order-status) (471238e)

- uses `orderStatuses` from `getSetting`.
- change component to receive an `orderStatusMap`. 
- impacts:
	- analytics/report/orders/table.js
	- header/activity-panel/panels/orders.js
	- docs/components/packages/order-status/docs/example.js

## Testing

- anything with order status should be smoke tested.

## `ProductImage` (packages/product-image) (646d439)

- replaced usage of `placeholder.png` from woocommerce assets with an svg value.
- impacts:
	- header/activity-panel/panels/reviews.js
	- header/activity-panel/panels/stock/card.js
	- product-image/docs/example.js
	- packages/search/autocompleters/product.js
	- packages/search/autocompleters/variations.js

### Notes

- this does change the weight from ~14kb to ~15kb to load the image via svg vs what's bundled with woocommerce, but at a mere 1kb gain, I think that's okay :)
- `product-image/placeholder.js` now actually has two svg's in it: `placeholderWhiteBackground` and `placeholderTransparentBackground`.  The `ProductImage` component is just using the white background option for now, but future iterations might include the possibility to select which placeholder to use?

## Testing

- verify all the impacted components/views continue to work as expected. 
- verify dev docs display component as expected.

### `countries` autocompleter (search/autocompleters/countries.js)

This autocompleter is coupled to the `dataEndpoints.countries` setting. I think this should be refactored so that either:

- countries comes from the api for the autocompleter or
- countries is provided to the `Autocomplete` component as a prop and the countries autocompleter can be a function that spits out the configuration with the given countries data hydrating it (eg `getCountriesCompleter( countries)` )

### `packages/currency` functions

Many of these helper functions would serve as really useful utility methods on `Currency` and Money value objects.

Some examples that I created for another project:

- https://github.com/eventespresso/event-espresso-core/blob/master/assets/src/vo/currency.js
- https://github.com/eventespresso/event-espresso-core/blob/master/assets/src/vo/money.js

We could then decouple settings from the package and consumers would just instantiate the value objects as needed for the app.

So I would suggest decoupling by implementing value objects.

### `packages/date`

There are two places where settings data is coupled to the functions:

#### 1. `getDateParamsFromQuery`.

The `defaultDateRange` argument is derived from `woocommerce_default_date_range` setting. This should be passed in as a argument to the function and callers of `getDateParamsFromQuery` can provide the default.

#### 2. `loadLocaleData`.

This uses `userLocale` and `weekdaysShort` from the `locale` object on the settings.  It is also called directly in this file.  imo this should receive `locale` as an argument and it should be instantiated by client, not in the package.

### `packages/navigation`

This package has the `adminUrl` setting coupled to the exported `getAdminLink` function.

I don't think `getAdminLink` should be an export of this package but instead should be something defined/consumed by the client (yes this means extensions would have to define their own function).

Alternatively, what we could do is have a `SiteSettings` value object that is exposed on a package that is instantiated by a configuration object. It could expose an interface for methods/properties that are populated from the configuration object on construct and then clients could take care of instantiating and consuming that instantiated value object where needed.  I'd be wary about making this a "catch-all" for all settings though.

### `packages/number`

The coupling here is for the `CURRENCY` object exposed by wcSettings. My comments here are similar to what I mentioned for `packages/currency` in that the method that currency is used in here (`numberFormat`) can likely be a part of a `Money` value object.


nerrad:
> All the packages are released at the same time if they have changes. Since the Date and Navigation packages also reference wcSettings it would be nice to make those changes in this PR. In a separate PR is cool too, I'm happy to revisit this one if you prefer.

Since there were only a couple changes needed I just did it in this pull.  So should be reviewed/tested again (looked fine to me in smoke tests, but would like confirmed)

psealock:
This is looking good @nerrad 

Respective changelogs will need to be updated.

Also, it occurred to me that other aliases designated by our webpack config refer to packages that actually exist. This makes it possible to use packages outside of our specific context.

For example, usage of `@wordpress/date` in a component gets mapped by our webpack to the global `window.wp.date`, but you can also use the date package and draw from the dependency itself and bundle it.

`@woocommerce/wc-admin-settings` is an alias that just exists in our webpack. What if WCPay project uses a component in their wc-admin extension screen? They will need to update their webpack to also include the alias. Given this, the package updates here are a breaking change.

Alternatively, can `@woocommerce/wc-admin-settings` be a package as well? We would move `client/settings/` into packages. Would that help with blocks too? This logic could all be in one place.

psealock:
### Update

In p1570401723201500-slack-team-isotope, a discussion led to the conclusion that `packages` should be free from any dependency on `@woocommerce/wc-admin-settings` or data coming from the server.

Instead, components, dates, and navigation packages should gather this data via props supplied by any client using these packages. This represents a breaking change and will require updating wc-admin's usage of the packages. Internal documentation to other Woo teams will be helpful.

rrennick:
@nerrad You have a build error in js - https://travis-ci.org/woocommerce/woocommerce-admin/jobs/594871096#L767

nerrad:
> @nerrad You have a build error in js

Thanks Ron, ya that was an export for a file/component that got removed in this refactor (superfluous, unneeded component).  For some reason I had no build issues locally (both with `npm run start` and `npm run build`, which I was doing frequently during the refactor to ensure things worked for each commit I did).  The fix is in b0b1d53

Should note a couple things regarding this pull:

- we need to get it merged or address the coupling issue before releasing an update to @woocommerce packages, so this is a blocker.
- I know the original description is long for this pull but it outlines all the changes
- _very important_ to note this is not ready for merge yet.  I got a good chunk of the work done (ready for review/testing) but there's still some files/components needing decoupled as per my notes near the end of the pull description.  Decoupling could be a bit tricky for these last few things so it _will_ need team input.

packages/components/src/advanced-filters/number-filter.js
```diff
@@ -70,10 +69,16 @@ class NumberFilter extends Component {
...
-	getFormControl( { type, value, label, onChange } ) {
+	getFormControl( {
+		type,
+		value,
+		label,
+		onChange,
+		currencySymbol,
+		symbolPosition,
+	} ) {
+		console.log( symbolPosition, currencySymbol );
```
psealock: leftover `console.log` here

psealock:
It looks like a jest test failing in `ProductImage`

nerrad:
Thanks for reviewing this @psealock :). Good to know we're most of the way done on this pull.  However, before merge there will still need to be the other unfinished items done (see the `Unfinished decoupling` section in the original pull notes.  

- Is this something you guys want me to do (probably need clearing with Pedro).
- Regardless of whether I finish things off or not, there is some more involved decoupling with at least a couple of the remaining things and requires some decision from the team on the approach to take (I gave some suggestions, but there may be other ideas).  For example, the autocompleter work was recently merged I think and might still have some in progress things being done elsewhere, @joshuatf you might have some thoughts on how the decoupling of `countries` from settings should go here.
- decoupling could be done in separate pulls based off of the branch in this pull for easier reviews?

packages/components/src/advanced-filters/number-filter.js
nerrad: Fixed in c799855

psealock:
>before merge there will still need to be the other unfinished items done

Can we merge this and work on the unfinished items separately? If we don't make a packages release in the meantime, it should be ok. Right?

nerrad:
Yup!

psealock:
Thanks @nerrad. I'll take what you've written and create new issues.

psealock:
I'm good to get this merged in this case.

PR #3004, opened 2019-10-07
Add options REST API permission checks

### LINKED ISSUE #2865: Onboarding: Options permissions should be updated
In https://github.com/woocommerce/woocommerce-admin/pull/2857 an options API was added.  This allows any option in `wp_options` to be updated to users with the `manage_options` capability.

This should probably be updated to whitelist specific options or at least filter options and capabilities.  We can whitelist specific options and only allow those to be updated OR we can allow any option to be updated, but provide specific capabilities for an array of options.

```php
$whitelisted_options = array(
     'option_name` => 'edit_posts`
);
```

joshuatf:
Fixes #2865

Adds permission checks to individual options the REST API.

**Question** - should we allow a default with permission `manage_options` for options without defined permissions?

1.  Hit the options API `wp-json/wc-admin/v1/options` with an admin who can `manage_woocommerce` and valid options as params (e.g., `woocommerce_demo_store_notice`).
2. Make sure that options are retrieved/updated for GET/POST requests respectively.  Note that POSTing should be done in JSON via the body.
3. Hit the same API with a user that does not have permissions to `manage_woocommerce`.
4. Make sure a forbidden error is returned.
5. Make sure all tests pass.

### DIFF
--- a/src/API/Options.php
+++ b/src/API/Options.php
@@ -43,7 +43,7 @@ class Options extends \WC_REST_Data_Controller {
 				array(
 					'methods'             => \WP_REST_Server::READABLE,
 					'callback'            => array( $this, 'get_options' ),
-					'permission_callback' => array( $this, 'update_item_permissions_check' ),
+					'permission_callback' => array( $this, 'get_item_permissions_check' ),
 				),
 				'schema' => array( $this, 'get_item_schema' ),
 			)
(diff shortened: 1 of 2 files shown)


src/API/Options.php
```diff
@@ -64,27 +64,80 @@ public function register_routes() {
...
+		$params      = explode( ',', $request['options'] );
+		$permissions = $this->get_option_permissions( $request );
+
+		if ( ! is_array( $params ) ) {
+			return new \WP_Error( 'woocommerce_rest_cannot_view', __( 'You must supply an array of options.', 'woocommerce-admin' ), 500 );
+		}
+
+		foreach ( $params as $option ) {
+			if ( ! isset( $permissions[ $option ] ) || ! $permissions[ $option ] ) {
+				return new \WP_Error( 'woocommerce_rest_cannot_update', __( 'Sorry, you cannot manage these options.', 'woocommerce-admin' ), array( 'status' => rest_authorization_required_code() ) );
```
justinshreve: Minor: Error code mentions `update` but this is a `GET` endpoint.

tests/api/options.php
```diff
@@ -0,0 +1,78 @@
...
+			array(
+				'role' => 'administrator',
+			)
+		);
+	}
+
+	/**
+	 * Test getting options without valid permissions.
+	 *
+	 * @since 3.5.0
```
justinshreve: The `@since` here can be dropped.

tests/api/options.php
```diff
@@ -0,0 +1,78 @@
...
+		$data     = $response->get_data();
+
+		$this->assertEquals( 200, $response->get_status() );
+		$this->assertEquals( get_option( 'woocommerce_demo_store_notice' ), $data['woocommerce_demo_store_notice'] );
+	}
+
+	/**
+	 * Test that options can be updated.
+	 */
+	public function test_update_options() {
```
justinshreve: Should we have a `test_update_options_without_permission` for completeness?

src/API/Options.php
```diff
@@ -64,27 +64,80 @@ public function register_routes() {
...
+	 * Check if a given request has access to get options.
+	 *
+	 * @param  WP_REST_Request $request Full details about the request.
+	 * @return WP_Error|boolean
+	 */
+	public function get_item_permissions_check( $request ) {
+		$params      = explode( ',', $request['options'] );
+		$permissions = $this->get_option_permissions( $request );
+
+		if ( ! is_array( $params ) ) {
```
justinshreve: This check (and the one in the update function) should probably be something like:

`if ( ! isset( $request['options'] ) || ! is_array( $params ) ) {`

Otherwise, passing a blank option shows the error message below `Sorry, you cannot manage these options.`.

src/API/Options.php
joshuatf: Nice catch, updated in https://github.com/woocommerce/woocommerce-admin/pull/3004/commits/ba72b7eb457ba83a6b14966c768ba3d2ea8b8c99

tests/api/options.php
joshuatf: Updated in https://github.com/woocommerce/woocommerce-admin/pull/3004/commits/ba72b7eb457ba83a6b14966c768ba3d2ea8b8c99

tests/api/options.php
joshuatf: Added in Updated in https://github.com/woocommerce/woocommerce-admin/pull/3004/commits/ba72b7eb457ba83a6b14966c768ba3d2ea8b8c99

src/API/Options.php
joshuatf: Updated the permission check and also added a default check for `manage_options` for undefined permissions.

joshuatf:
Thanks for the review @justinshreve - handled feedback and this is ready for another look.

PR #3007, opened 2019-10-08
Onboarding: Skip tax steps if complete and mark task complete

### LINKED ISSUE #2988: Complete Tax Task Todos
While working on implementing tracks to the task list, I noticed there is some remaining `@todo` comments, as well as some tightening up of the flow.

https://github.com/woocommerce/woocommerce-admin/blob/9cdabf51e67123b43103f5e16900422247648679/client/dashboard/task-list/tasks/tax.js#L324-L328

- [ ] `isJetpackConnected` is hard coded and needs updated to use the `isJetpackConnected()` selector. `client/dashboard/task-list/tasks/tax.js`.
- [ ] `pluginsToActivate` is hard coded and should look at `getActivePlugins()` to determine which plugins actually need installed. `client/dashboard/task-list/tasks/tax.js`.
- [ ] The `is-complete` class name needs added to the tax task when it is completed. `client/dashboard/task-list/index.js`.

I started to implement those in my events branch, but noticed that this raised some more issues that need some more work.

Once the above checks are properly implemented, if you try to go through the connect flow via the tax task, after the Jetpack redirect, you will end up on the main task list with query string `?step=store-details` in the URL. Clicking into the tax task shows you the `Good News!` message.

This means you never end up making it to these screens in the tax list of steps after connecting: https://www.figma.com/file/xyKPRZ73kKhDyNjwbFvBJR/Task-List?node-id=976%3A8566.

- [ ] The Jetpack Connect API needs to be extended to accept a step name and optional task, so that steps like tax and shipping can redirect back to the correct place.
- [ ] Make sure the above Figma flow is reachable.

joshuatf:
Fixes #2988

Skips the plugin installation steps if plugins are already installed, skips Jetpack connection if connected, and marks the task complete if a tax rate exists or automatic tax rates are enabled.

>Clicking into the tax task shows you the Good News! message.
This means you never end up making it to these screens in the tax list of steps after connecting: https://www.figma.com/file/xyKPRZ73kKhDyNjwbFvBJR/Task-List?node-id=976%3A8566.

@justinshreve This is referring to this step that allows enabling automatic sales tax pictured below, right?  I'm wondering if it's better just to show the "Good news" screen since these do the same thing and reduce the amount of logic in this code.  Also the "Good news" screen has an option for manually setting taxes.

Screen Shot 2019-10-08 at 12 50 36 PM

### Screenshots
Screen Shot 2019-10-08 at 12 41 15 PM

1. Delete all tax rates and set the option `wc_connect_taxes_enabled` to `no` if it exists.
2. Disconnect jetpack and deactive WCS or Jetpack.
3. Walk through the tax task and leave after the plugin installation step.
4. Return to the task and the plugins step should be skipped (not visible).
5. Do the same for the Jetpack connection step.
6. After re-entering this screen, you should see a "Good news" screen if the store is in a country that supports automatic tax calculation.  Complete it, either via manually adding tax rates or enabling automatic rates.
7. Make sure the task is marked complete in the task list.

### DIFF
--- a/client/dashboard/task-list/tasks/tax.js
+++ b/client/dashboard/task-list/tasks/tax.js
@@ -4,10 +4,9 @@
  */
 import { __ } from '@wordpress/i18n';
 import { Button } from 'newspack-components';
-import { Component, Fragment } from '@wordpress/element';
+import { Component } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import { difference, filter } from 'lodash';
-import { FormToggle } from '@wordpress/components';
 import interpolateComponents from 'interpolate-components';
 import { withDispatch } from '@wordpress/data';
 
@@ -68,11 +67,11 @@ class Tax extends Component {
 
 		// Show the success screen if all requirements are satisfied from the beginning.
 		if (
-			0 === stepIndex &&
+			null !== stepIndex &&
 			( ! pluginsToActivate.length &&
 				isCompleteAddress &&
 				isJetpackConnected &&
-				this.isSupportedCountry() )
+				this.isTaxJarSupported() )
 		) {
 			/* eslint-disable react/no-did-update-set-state */
 			this.setState( { stepIndex: null } );
@@ -100,10 +99,14 @@ class Tax extends Component {
 		}
 	}
 
-	isSupportedCountry() {
-		const { countryCode } = this.props;
+	isTaxJarSupported() {
+		const { countryCode, options } = this.props;
 		const { automatedTaxSupportedCountries = [] } = getSetting( 'onboarding', {} );
-		return automatedTaxSupportedCountries.includes( countryCode );
+
+		return (
+			'1' === options.woocommerce_setup_jetpack_opted_in &&
+			automatedTaxSupportedCountries.includes( countryCode )
+		);
 	}
 
 	completeStep() {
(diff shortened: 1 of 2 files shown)


joshuatf:
Thanks for the feedback 😄 

>Yeah that screenshot is what I was talking about. I agree with you that it makes more sense to just show the good news message. Let's remove the other code 🔥.

Removed 🔥 

>While testing this, the /settings/tax/batch call fails for me. I still get the Your tax settings have been updated. message though. Since the call fails, the task does not get marked as complete.

I think we hit this issue before.  Seems like its related to wcs TOS not being accepted.  I added a check here to make sure we're opted in first and also logged a somewhat related issue about setting this in the start step of the profiler - https://github.com/woocommerce/woocommerce-admin/issues/3018

For testing as is, make sure you have `woocommerce_setup_jetpack_opted_in` set to `1` in your options table or false to force manual configuration.

>I am also still getting redirected to ?step=store-details (which just loads the main task dashboard) when connecting via the tax task. This should redirect back to the Good News! message automatically.

Logged an issue to update this so we can change the redirect behavior on the client-side since the `Connect` component handles this in multiple places. https://github.com/woocommerce/woocommerce-admin/issues/3019

justinshreve:
>  Seems like its related to wcs TOS not being accepted. I added a check here to make sure we're opted in first and also logged a somewhat related issue about setting this in the start step of the profiler - #3018

Doh! I'm guessing the setting got cleared out for me during a deactivate of Jetpack/WCS and I didn't run back through the profiler. Thanks for logging the issue about setting the option when both are detected.

client/dashboard/task-list/tasks/tax.js
```diff
@@ -100,10 +99,14 @@ class Tax extends Component {
...
 
-	isSupportedCountry() {
-		const { countryCode } = this.props;
+	isTaxJarSupported() {
+		const { countryCode, options } = this.props;
 		const { automatedTaxSupportedCountries = [] } = getSetting( 'onboarding', {} );
-		return automatedTaxSupportedCountries.includes( countryCode );
+
+		return (
+			'1' === options.woocommerce_setup_jetpack_opted_in &&
```
justinshreve: We should actually be checking `wc_connect_options` for `tos_accepted`.

The Jetpack option actually gets deleted: https://github.com/Automattic/woocommerce-services/blob/62654d48ff51cf119be5c87a455adbbb6e61c44d/woocommerce-services.php#L416-L431

https://github.com/Automattic/woocommerce-services/blob/62654d48ff51cf119be5c87a455adbbb6e61c44d/woocommerce-services.php#L416-L431

joshuatf: Oh nice catch!  Updated in https://github.com/woocommerce/woocommerce-admin/pull/3007/commits/2740e0e988d2a0fe52510ff36a1f742f9d880b4f

justinshreve:
Sorry @joshuatf, the call is still failing for me and I definitely have the TOS option approved.

It's failing for me on this check here: https://github.com/woocommerce/woocommerce-rest-api/blob/854ac947d84a3093a67af6714f11378990e1a916/src/Controllers/Version4/SettingsOptions.php#L264-L268

When I inspect `$settings`, `wc_connect_taxes_enabled` is not part of the settings here. Maybe we should use the options API to set it instead?

joshuatf:
@justinshreve I can't quite seem to replicate the issue if TOS has been accepted.

Could you do me a favor and see if this line is reached?  https://github.com/Automattic/woocommerce-services/blob/62654d48ff51cf119be5c87a455adbbb6e61c44d/woocommerce-services.php#L450

That hook eventually loads the TaxJar dependency which adds the `wc_connect_taxes_enabled` setting.

https://github.com/Automattic/woocommerce-services/blob/62654d48ff51cf119be5c87a455adbbb6e61c44d/woocommerce-services.php#L631

joshuatf:
Thanks for reviewing again, @justinshreve!

>I wonder if we should do some logic to detect that just to avoid a confusing loop like this? (Which could be done in a follow-up PR).

Definitely.  Created an issue for that here - https://github.com/woocommerce/woocommerce-admin/issues/3051

Thanks so much for taking the time to debug that!

>I had to refresh the page for the completed CSS class to show.

I think this was because settings were only being retrieved outside the component on initial page load.  I moved them into `getTasks()` but realized on rebase that you had already done this when moving that function to `tasks.js` 🥇 .  Seems to work for me now, but can you double check on your end?

PR #3020, opened 2019-10-09
Onboarding: Replace TextControl component

### LINKED ISSUE #2581: Text field hover / focus styles

joshuatf:
Fixes #2581

Replaces the newspack TextControl component with our own so we can add focus styles/events.

I couldn't find a `$muriel-gray-400` - is the "400" here referring to `$dark-gray-400`? /cc @jameskoster 

### Screenshots
Screen Shot 2019-10-09 at 5 43 33 PM

1.  Go through various pages of onboarding profiler and task list that use the `TextControl` component imported from `@woocommerce/components`.
2.  Make sure styling is correct and focus box-shadow colors now work as expected.

### DIFF
--- a/packages/components/src/text-control/style.scss
+++ b/packages/components/src/text-control/style.scss
@@ -1,0 +1,93 @@
+/** @format */
+
+.muriel-input-text {
+	background: $muriel-white;
+	border: 1px solid $muriel-gray-200;
+	border-radius: 0;
+	box-shadow: none;
+	padding: 12px 12px 4px;
+	position: relative;
+
+	&:hover {
+		border-color: $dark-gray-400;
+	}
+
+	label {
+		color: $muriel-gray-500;
+		font-size: 14px;
+		line-height: 21px;
+
+		&.components-base-control__label {
+			margin: 0;
+		}
+	}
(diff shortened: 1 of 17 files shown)


packages/components/src/text-control/style.scss
```diff
@@ -0,0 +1,93 @@
+/** @format */
+
+.muriel-input-text {
+	background: $muriel-white;
+	border: 1px solid $muriel-gray-200;
```
justinshreve: > I couldn't find a $muriel-gray-400 - is the "400" here referring to $dark-gray-400?

The `$muriel-gray-` colors are meant to be temporary and are defined to make the newspack components work. See the note here: https://github.com/woocommerce/woocommerce-admin/blob/master/client/stylesheets/abstracts/_variables.scss#L78

We can/should use the `$studio-` variables directly.

joshuatf: Thanks for linking to those- got those colors updated 😄

PR #3039, opened 2019-10-14
Add Inbox note action indication

### LINKED ISSUE #2804: Client side actions for notes
Currently the action for any given note either allows a URL or sends a server request via `triggerNoteAction` in wc-api.

Some actions may require more interaction with the page and need to be handled with Javascript.  For example, the plugin installer in this issue  https://github.com/woocommerce/woocommerce-admin/pull/2798#discussion_r314179877 is handled server-side but needs an indicator that the action is pending client-side while the plugin is being installed. 

A quick and easy solution may be to set the button/note/action to pending and have a callback that determines completion.  However, it may be worth also investigation a more JS-hook oriented solution to allow finer control.

Screen Shot 2019-08-16 at 2 20 57 PM
(In the above screenshot, "Install now" has been clicked, but there are no visual cues that the install is currently underway).

jeffstieler:
Fixes #2804

This PR introduces the following new functionality:
* Adds a "busy" treatment to Inbox Note action buttons that aren't simply links (or link within wp-admin)
* Allows note actions to be deleted (changes to `WC_Admin_Note` and its `DataStore`)
* Updates the FB note content and actions after installing the extension

### Accessibility

- Ensure you do not have the Facebook for WooCommerce Extension installed
- Ensure your store has a published product (or more)
- Use Crontrol to run the wc_admin_daily job
- Open the WooCommerce > Dashboard
- Verify the facebook note appears in the inbox
- Verify the "learn more" button function
- Verify the "install now" button function (check Plugins > Installed after)
- Verify the note changes its icon, title, body copy, and actions after install
- Verify the "setup" action takes you to the FB settings page

### Changelog Note:

Tweak: add visual feedback for Inbox Note Actions

### DIFF
--- a/client/header/activity-panel/panels/inbox/action.js
+++ b/client/header/activity-panel/panels/inbox/action.js
@@ -1,0 +1,72 @@
+/** @format */
+/**
+ * External dependencies
+ */
+import { Button } from '@wordpress/components';
+import { Component } from '@wordpress/element';
+import { compose } from '@wordpress/compose';
+import { withDispatch } from '@wordpress/data';
+import PropTypes from 'prop-types';
+
+/**
+ * WooCommerce dependencies
+ */
+import { ADMIN_URL as adminUrl } from '@woocommerce/wc-admin-settings';
+
+/**
+ * Internal dependencies
+ */
+
+class InboxNoteAction extends Component {
+	constructor( props ) {
+		super( props );
+
+		this.handleActionClick = this.handleActionClick.bind( this );
+	}
+
+	handleActionClick( event ) {
+		const { action, noteId, triggerNoteAction } = this.props;
+		const href = event.target.href || '';
+
+		if ( href.length && ! href.startsWith( adminUrl ) ) {
+			event.preventDefault();
+			window.open( href, '_blank' );
+		}
+
+		triggerNoteAction( noteId, action.id );
+	}
+
+	render() {
+		const { action } = this.props;
+		return (
+			<Button
+				isDefault
+				isPrimary={ action.primary }
+				href={ action.url || undefined }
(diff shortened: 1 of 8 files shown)


client/header/activity-panel/panels/inbox/action.js
```diff
@@ -0,0 +1,79 @@
...
+		this.setState( { inAction }, () => triggerNoteAction( noteId, action.id ) );
+	}
+
+	render() {
+		const { action } = this.props;
+		return (
+			<Button
+				isDefault
+				isPrimary={ action.primary }
+				isBusy={ this.state.inAction }
```
joshuatf: Not directly related to this PR, but the busy state looks a bit weird:

Screen Shot 2019-10-23 at 12 53 41 PM

src/Notes/WC_Admin_Notes_Facebook_Extension.php
```diff
@@ -97,6 +97,25 @@ public function install_facebook_extension( $note ) {
 
 			$activate_request = array( 'plugins' => 'facebook-for-woocommerce' );
 			$installer->activate_plugins( $activate_request );
+
+			$content = __( 'You\'re almost ready to start driving sales with Facebook. Complete the setup steps to control how WooCommerce integrates with your Facebook store.', 'woocommerce-admin' );
+			$note->set_title( __( 'Market on Facebook — Installed', 'woocommerce-admin' ) );
```
joshuatf: When this note replaces the installation note, I get the following console error:

Screen Shot 2019-10-23 at 12 54 08 PM

jeffstieler: Fixed in acc7b6f.

PR #3040, opened 2019-10-14
Update proceed without Jetpack/WCS to go to store details step

### LINKED ISSUE #2982: Proceeding without Jetpack/WCS should go to the store details step
Currently, if you click `Proceed without WooCommerce Services` or `Proceed without Jetpack & WooCommerce Services`, the profile wizard is skipped and you are taken to the dashboard and task list.

None of the other steps require a Jetpack/WCS connection, so we miss out on getting things like their store address setup. 

I believe this should actually take you into the store details step instead, just like how we automatically take them there if WCS & Jetpack are already installed.

Right now we also use `skip` as a flag for toggling the wizard back on/off. We might want to call this something else since we won't have a skip mechanism?

justinshreve:
Closes #2982.

This PR changes the behavior of the "Proceed without Jetpack/WooCommerce Services" link on the first page of the profile wizard.

Now the link will go to the store details step and the rest of the wizard (just like what happens automatically if Jetpack & WCS are already installed). None of these steps require a Jetpack connection.

This PR removes the skipped flag, since `wcs_jetpack` already tracks if the user skips the step. I updated the Help tab reset to just toggle the completed flag because of this. I think this will be fine, because even if you toggle it back to completed without filling out everything, your existing data still exists. I've added a tracks event so that we can include it in funnels if necessary.

To Test:
* Disable `WooCommerce Services`.
* Toggle the profile wizard back on under the `Help` menu.
* Test the proceed without Jetpack/WCS link. You should end up on the store details step.
* Test toggling the profile wizard back to completed.

### DIFF
--- a/client/dashboard/profile-wizard/steps/start/index.js
+++ b/client/dashboard/profile-wizard/steps/start/index.js
@@ -41,20 +41,26 @@ class Start extends Component {
 	}
 
 	componentDidMount() {
-		const { updateProfileItems } = this.props;
+		const { updateProfileItems, profileItems } = this.props;
 		if (
 			this.props.activePlugins.includes( 'jetpack' ) &&
 			this.props.activePlugins.includes( 'woocommerce-services' )
 		) {
-			updateProfileItems( { wcs_jetpack: 'already-installed' } );
+			// Don't track event again if they revisit the start page.
+			if ( 'already-installed' !== profileItems.plugins ) {
+				recordEvent( 'wcadmin_storeprofiler_already_installed_plugins', {} );
+			}
+
+			updateProfileItems( { plugins: 'already-installed' } );
 			return updateQueryString( { step: 'store-details' } );
 		}
 	}
 
 	async skipWizard() {
-		const { createNotice, isProfileItemsError, updateProfileItems } = this.props;
+		const { createNotice, isProfileItemsError, updateProfileItems, activePlugins } = this.props;
 
-		await updateProfileItems( { wcs_jetpack: 'skipped' } );
+		const plugins = activePlugins.includes( 'jetpack' ) ? 'skipped-wcs' : 'skipped';
+		await updateProfileItems( { plugins } );
 
 		if ( isProfileItemsError ) {
 			createNotice(
@@ -62,7 +68,7 @@ class Start extends Component {
 				__( 'There was a problem updating your preferences.', 'woocommerce-admin' )
 			);
 		} else {
-			recordEvent( 'storeprofiler_welcome_clicked', { get_started: true } );
+			recordEvent( 'storeprofiler_welcome_clicked', { get_started: true, plugins } );
 			return updateQueryString( { step: 'store-details' } );
 		}
 	}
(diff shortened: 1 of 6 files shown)


client/dashboard/profile-wizard/steps/start/index.js
```diff
@@ -54,7 +54,7 @@ class Start extends Component {
 	async skipWizard() {
 		const { createNotice, isProfileItemsError, updateProfileItems } = this.props;
 
-		await updateProfileItems( { skipped: true, wcs_jetpack: 'skipped' } );
+		await updateProfileItems( { wcs_jetpack: 'skipped' } );
```
joshuatf: Wondering if instead of `wcs_jetpack` we could use something like `plugins_skipped`.  In this case I only skipped WCS and not Jetpack.  This would also leave room to grow if we add additional plugins.

justinshreve: This param is also used to detect if the plugins are already installed (to skip the start page), so `plugins_skipped` wouldn't work, but I agree the naming could be better than `wcs_jetpack`.

I renamed this to `plugins` in https://github.com/woocommerce/woocommerce-admin/pull/3040/commits/ef3e3fe5bd7f28295869b1ff60f52c9fd7affef8, and improved our tracking a bit. Do you mind taking another look since I doubled the line count of the PR?

client/dashboard/profile-wizard/steps/start/index.js
```diff
@@ -41,42 +41,58 @@ class Start extends Component {
...
 			return updateQueryString( { step: 'store-details' } );
 		}
 	}
 
 	async skipWizard() {
-		const { createNotice, isProfileItemsError, updateProfileItems } = this.props;
+		const { createNotice, isProfileItemsError, updateProfileItems, activePlugins } = this.props;
 
-		await updateProfileItems( { skipped: true, wcs_jetpack: 'skipped' } );
+		const plugins = activePlugins.includes( 'jetpack' ) ? 'skipped-wcs' : 'skipped';
```
joshuatf: I may be overthinking this, but this won't track if WCS is installed and Jetpack is skipped, right?

Maybe we just keep the `skipped` boolean setting (or update to `skipped_plugins`) and just record the items that are being skipped in tracks.  What do you think?

justinshreve: > Could you check against this.props.activePlugins or would that not work?

This won't work. I need to know that the page was skipped automatically, so that we show the usage tracking modal on the store details step instead: https://github.com/woocommerce/woocommerce-admin/pull/3040/files#diff-5965f2cd86a39747c2fb0ce9d391e444. I can't rely on activePlugins there because it should always contain both by the time you get to the step.

> I may be overthinking this, but this won't track if WCS is installed and Jetpack is skipped, right?

Correct, but we are not handling WCS being installed without Jetpack at all right now. If Jetpack is not installed, we end up showing both sets of benefits -- so `skipped` would apply here. I thought there was a Slack conversation about this, but I can't seem to find it now. But I think the gist is that we should expect this to be an edge case. WooCommerce Services doesn't do anything without Jetpack except show some banners.

> Maybe we just keep the skipped boolean setting (or update to skipped_plugins) and just record the items that are being skipped in tracks. What do you think?

I still need to keep some form of `plugins` == 'already-installed' logic. I think the current enum logic does a good job of telling which page the user ends up seeing based on the different flows we currently have implemented, and allows us to show the modal later if necessary. I could separate this out to two different pieces of meta though, if you think a bool is better here.

joshuatf: >WooCommerce Services doesn't do anything without Jetpack except show some banners. To simplify our logic, we show both plugins if Jetpack isn't present. Reactivating WCS on the activate step doesn't harm anything either.

That makes a lot of sense.  Thanks for the explanation!

>I could separate this out to two different pieces of meta though, if you think a bool is better here.

I think you're right.  Let's keep it as is and we can iterate if necessary.

Issue #3038, opened 2019-10-14
Simplify the new onboarding install for internal testing


pmcpinto:
The new onboarding is going to have so many different paths and events, so the sooner we can start testing what we have the better. In order to start the testing, it would be better to have an easy way to install the new onboarding. Thanks

justinshreve:
Closing this out now that we have a way to create builds, and some directions up at p90Yrv-1iA-p2.

PR #3044, opened 2019-10-15
Add the ability to create custom plugin builds

justinshreve:
Part of https://github.com/woocommerce/woocommerce-admin/issues/3038.

This PR adds the ability to generate custom plugin builds. We would like to start testing onboarding with others inside of Automattic, including non-developers. Being able to generate and provide a zip so testers don't need to build development will increase testing adoption.

I wanted to make this more general purpose, so the `build:release` command now accepts some optional arguments: `--slug` and `--features`. Base feature flags are pulled from `config/plugin.json` and your additional changes are overlaid on top. When the build is complete, a `woocommerce-admin-$slug.zip` file will be generated.

For example, to create a `woocommerce-admin-onboarding.zip` build by enabling onboarding in addition to the feature flags defined in `config/plugin.json`, the command would be:

`npm run build:release -- --slug onboarding --features '{"onboarding":true}'`.

This PR also sets the default calypso environment to `wpcalypso` for now, so that Calypso changes can also easily be tested without needing to build a local copy of Calypso: https://github.com/Automattic/wp-calypso/pull/36790

* Test running the default `npm run build:release`  command and verify things look as they should.
* Test running `npm run build:release -- --slug onboarding --features '{"onboarding":true}'`. Uploading the resulting zip on another test site (such as wpsandbox) should allow you to test onboarding. You may need to reset the wizard under `Orders > Help > Setup Wizard`.

### DIFF
--- a/bin/build-plugin-zip.sh
+++ b/bin/build-plugin-zip.sh
@@ -1,5 +1,17 @@
 #!/bin/bash
 
+ZIP_FILE='woocommerce-admin.zip';
+
+while [ $# -gt 0 ]; do
+	if [[ $1 == '-f' || $1 == '--features' ]]; then
+		export WC_ADMIN_ADDITIONAL_FEATURES="$2"
+	fi
+	if [[ $1 == '-s' || $1 == '--slug' ]]; then
+		ZIP_FILE="woocommerce-admin-$2.zip";
+	fi
+	shift
+done
+
 # Exit if any command fails.
 set -e
 
@@ -74,7 +86,7 @@ build_files=$(ls dist/*/*.{js,css})
 
 # Generate the plugin zip file.
 status "Creating archive... 🎁"
-zip -r woocommerce-admin.zip \
+zip -r ${ZIP_FILE} \
 	woocommerce-admin.php \
 	uninstall.php \
 	includes/ \
--- a/bin/generate-feature-config.php
+++ b/bin/generate-feature-config.php
@@ -11,15 +11,16 @@
  * - plugin: For the standalone feature plugin, for GitHub and WordPress.org.
  * - core: Stable features for WooCommerce core merge.
  */
-$phase = isset( $_SERVER['WC_ADMIN_PHASE'] ) ? $_SERVER['WC_ADMIN_PHASE'] : ''; // WPCS: sanitization ok.
+$phase = isset( $_ENV['WC_ADMIN_PHASE'] ) ? $_ENV['WC_ADMIN_PHASE'] : ''; // WPCS: sanitization ok.
+
 if ( ! in_array( $phase, array( 'development', 'plugin', 'core' ), true ) ) {
 	$phase = 'plugin'; // Default to plugin when running `npm run build`.
 }
 $config_json = file_get_contents( 'config/' . $phase . '.json' );
 $config      = json_decode( $config_json );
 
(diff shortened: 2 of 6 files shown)


psealock:
The plugin build is not rendering the dashboard. Its unrelated to the changes here, but I'll bring it up here and separate it out to another issue if need be.

This is only reproduced using the **plugin build**, not development. Could there be a feature flag issue?

Screen Shot 2019-10-16 at 9 58 54 AM

`profileItems` is not defined

Screen Shot 2019-10-16 at 9 58 39 AM

Which is causing problems here:

https://github.com/woocommerce/woocommerce-admin/blob/dff49b40c06e36e2af16b598de79e1e82a661b78/client/dashboard/task-list/tasks.js#L58

justinshreve:
@psealock Can you upload the resulting .zip file somewhere? Seems like maybe the PHP side of the feature flags isn't loading / prefetching some of the data.

psealock:
Sure. Shouldn't the task list logic not run at all if the onboarding flag is set to false for the plugin?

justinshreve:
Oh sorry, I misunderstood. When you said this was unrelated to the changes, did you mean this is happening on master with the normal build plugin command? I’ll take a look tomorrow.

justinshreve:
@psealock This has been fixed in https://github.com/woocommerce/woocommerce-admin/pull/3053, and I've rebased this PR against it.

psealock:
>did you mean this is happening on master with the normal build plugin command?

Yes, thanks for addressing. I'll have another look at this one.

bin/build-plugin-zip.sh
```diff
@@ -1,5 +1,17 @@
...
+
+while [ $# -gt 0 ]; do
+	if [[ $1 == '-f' || $1 == '--features' ]]; then
+		export WC_ADMIN_ADDITIONAL_FEATURES="$2"
+	fi
+	if [[ $1 == '-s' || $1 == '--slug' ]]; then
+		ZIP_FILE="woocommerce-admin-$2.zip";
+	fi
+	shift
+done
```
psealock: Ah ok, so the slug can be anything, ie `onboarding-test-wpcom`, but isn't required. Thats nice 👍

psealock: Looks like the error reported here, https://github.com/woocommerce/woocommerce-admin/issues/3056

justinshreve: That was indeed the same error @psealock. I've rebased this, and made sure to test the zip with that option cleared out. Should be fixed now.

bin/generate-feature-config.php
```diff
@@ -18,6 +18,14 @@
 $config_json = file_get_contents( 'config/' . $phase . '.json' );
 $config      = json_decode( $config_json );
 
+if ( ! empty( $_SERVER['WC_ADMIN_ADDITIONAL_FEATURES'] ) ) {
+	$additional_features = json_decode( $_SERVER['WC_ADMIN_ADDITIONAL_FEATURES'], true );
```
jeffstieler: This seems to work fine, but would `$_ENV` or `getenv()` be better than `$_SERVER` here? _(`$_SERVER` just makes me think we're in an HTTP request and not a cli script)_

justinshreve: Updated to `$_ENV`. Agreed this makes more sense since they are environment variables. 

https://github.com/woocommerce/woocommerce-admin/pull/3044/commits/4e49066207713f134e7f52cecb966fbea3e757af.

PR #3058, opened 2019-10-17
Onboarding: Add store is ready for launch note

### LINKED ISSUE #2995: Onboarding: WC-Admin Inbox message: Congratulations - your store is ready
Once the task list is complete, we are supposed to show a "Congratulations - your store is ready" inbox notification.

https://www.figma.com/file/xyKPRZ73kKhDyNjwbFvBJR/Task-List?node-id=1243%3A164

cc @pmcpinto Do you know what guide this notification should link to?

joshuatf:
Fixes #2995

Adds a note when all tasks are completed in the task list.

Questions:

* This note is only triggered after page refresh since the hook is not fired over the REST request.  Should we create read and create note queries in wc-api to add this note via REST?
* Is the marketing button URL okay for now?

### Screenshots
Screen Shot 2019-10-17 at 6 02 44 PM

1.  Complete all items in the task list.
2. Refresh or go to another admin page.
3. Note the note in your inbox.
4. Make sure this note is not duplicated when repeating these steps.

### DIFF
--- a/client/dashboard/task-list/index.js
+++ b/client/dashboard/task-list/index.js
@@ -30,6 +30,12 @@ class TaskDashboard extends Component {
 		document.body.classList.add( 'woocommerce-task-dashboard__body' );
 
 		this.recordEvent();
+
+		if ( this.props.inline ) {
+			this.props.updateOptions( {
+				woocommerce_task_list_complete: true,
+			} );
+		}
 	}
 
 	componentWillUnmount() {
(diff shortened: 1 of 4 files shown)


joshuatf:
@justinshreve Thanks for reviewing!

>An alternative to this that might work with little refactoring: Hook into the option update hook update_option_woocommerce_task_list_complete

This is the original hook I attempted (but with `add_option_...` since I think `update_option_...` only runs on existing options), however it never got fired.  I assumed that this was that hook not firing on REST request, but digging a little further realized that the `OnboardingTasks` class never gets initiated during non-admin pages.

Refactored this a bit in https://github.com/woocommerce/woocommerce-admin/pull/3058/commits/c72e072665b63cf92e3389ee138ca4f342790e48 and added another `is_admin()` check beneath the options.  Would you mind taking another look before merge?

PR #3059, opened 2019-10-17
Onboarding: Localize currency ranges inside translated string

### LINKED ISSUE #2768: Business Details Revenue Options i18n

Any other ideas?

joshuatf:
Fixes (part of) #2768

Adds the currency symbol and amount directly to the string to allow translators to update the range based on currency.

### Screenshots
Screen Shot 2019-10-17 at 6 24 13 PM

1. Enable the onboarding profiler if not enabled.
2. Go to the business details step. `wp-admin/admin.php?page=wc-admin&step=business-details`
3. Make sure the revenue and product dropdowns continue working as expected.

### DIFF
--- a/client/dashboard/profile-wizard/steps/business-details.js
+++ b/client/dashboard/profile-wizard/steps/business-details.js
@@ -127,7 +127,8 @@ class BusinessDetails extends Component {
 
 		return _x(
 			format( min ) + ' - ' + format( max ),
-			'store product count range or revenue range',
+			'store product count range or revenue range, range should be modified' +
+				'to better match currency if the conversion differs greatly from USD',
 			'woocommerce-admin'
 		);
 	}
--- a/client/dashboard/profile-wizard/steps/business-details.js
+++ b/client/dashboard/profile-wizard/steps/business-details.js
@@ -119,14 +119,16 @@ class BusinessDetails extends Component {
 	}
 
 	getNumberRangeString( min, max = false, format = numberFormat ) {
+		// The localized strings below do not use sprintf to allow
+		// formatting the range on a per currency basis via translation.
 		if ( ! max ) {
-			return sprintf( _x( '%s+', 'store product count', 'woocommerce-admin' ), format( min ) );
+			return _x( format( min ) + '+', 'store product count or revenue', 'woocommerce-admin' );
 		}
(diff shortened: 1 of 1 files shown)


client/dashboard/profile-wizard/steps/business-details.js
```diff
@@ -119,14 +119,17 @@ class BusinessDetails extends Component {
 	}
 
 	getNumberRangeString( min, max = false, format = numberFormat ) {
+		// The localized strings below do not use sprintf to allow
```
justinshreve: Unfortunately, I don't think this is going to end up working. I ran `npm run i18n` and looked at the resulting strings, and these just drop off the value:

```
#: languages/woocommerce-admin.php:2114
msgctxt "More than a certain revenue amount"
msgid "More than "
msgstr ""
```

I was originally thinking we would hard coded the string as

`More than $250,000` and the string could translate from there.. but you bring up a good point that the store language and currency may differ. Though since we don't currently ask for currency in the wizard I think this would be WC default anyway.

The revenue answer also isn't used inside the store, so I think it's OK that it's a difference currency, as long as it's in a format the user can understand without converting.

client/dashboard/profile-wizard/steps/business-details.js
```diff
@@ -119,14 +119,17 @@ class BusinessDetails extends Component {
...
+		// formatting the range on a per currency basis via translation.
 		if ( ! max ) {
-			return sprintf( _x( '%s+', 'store product count', 'woocommerce-admin' ), format( min ) );
+			return _x( format( min ) + '+', 'store product count or revenue', 'woocommerce-admin' );
 		}
 
-		return sprintf(
-			_x( '%s - %s', 'store product count', 'woocommerce-admin' ),
```
jeffstieler: If this gets restored, it should just positional formatting so RTL languages can swap the values. E.g. `%1$s - %2$s`.

client/dashboard/profile-wizard/steps/business-details.js
joshuatf: Bummer- reverted back to the sprintf so this PR is now just adding a comment and the positional formatting that @jeffstieler mentioned below.

PR #3061, opened 2019-10-17
Add notices to return to task list on product and homepage tasks.

### LINKED ISSUE #2890: Onboarding: Add notice to return to task list outside wc-admin screens
As per the discussion in p1567601494015300-slack-wc-onboarding we want to display a link to return to the task list after publishing the homepage.

We'll need to enqueue a script to run on the edit post page and hook into Gutenberg's save hook to detect when the post is published and then create a notice.

justinshreve:
Closes #2890.

This PR updates the success notices that are displayed when creating a product through the onboarding flow or when creating a homepage through the onboarding flow. This way users can continue setup easily.

Since this requires a separate script to be loaded (separate from the main app entry point or embedded view), I've added a new folder to house scripts like this (I imagine there will be other instances in the future, as we update screens and experiences). I wasn’t sure what exactly to call these, so I’ve named the folder `wp-admin-scripts` (since `wp-scripts` is used by a core NPM package). But I’m open to changing this if anyone has a suggestion.

### Screenshots

Screen Shot 2019-10-16 at 4 06 25 PM

Screen Shot 2019-10-16 at 4 14 20 PM

* Make sure the profile wizard is disabled, but the task list is enabled. (Under `Orders > Help > Setup Wizard`).

To test the products notice:

* Delete all of your products.
* Go to the task list and select “add your first product”, followed by the “Add manually” option.
* Add some product information and hit publish.
* You should see the “you created your first product!” message, with a link to continue setup.

To test the homepage notice:

* Delete your front page.
* Go to the task list and select “personalize your store”. Click “create homepage”.
* Make any desired changes to the homepage template, and then hit publish twice.
* You should see a “Your homepage was published.” Message, with a continue setup link.

### DIFF
--- a/client/wp-admin-scripts/onboarding-homepage-notice/index.js
+++ b/client/wp-admin-scripts/onboarding-homepage-notice/index.js
@@ -21,29 +21,17 @@ const onboardingHomepageNotice = () => {
 	const post = select( 'core/editor' ).getCurrentPost();
 
 	const noticeMeta = {
-		wasSavingPost: select( 'core/editor' ).isSavingPost(),
-		wasAutosavingPost: select( 'core/editor' ).isAutosavingPost(),
 		wasPublishingPost: select( 'core/editor' ).isPublishingPost(),
 		wasStatus: post.status,
 	};
 
 	subscribe( () => {
-		const isSavingPost = select( 'core/editor' ).isSavingPost();
-		const isAutosavingPost = select( 'core/editor' ).isAutosavingPost();
 		const isPublishingPost = select( 'core/editor' ).isPublishingPost();
 		const isStatus = select( 'core/editor' ).getCurrentPost().status;
 
-		// This triggers when the homepage is first published.
 		const shouldTriggerAdminNotice =
-			noticeMeta.wasSavingPost &&
-			! isSavingPost &&
-			noticeMeta.wasPublishingPost &&
-			! isPublishingPost &&
-			! noticeMeta.wasAutosavingPos &&
-			'publish' === isStatus;
+			'publish' === isStatus && noticeMeta.wasPublishingPost && ! isStatus.isPublishingPost;
 
-		noticeMeta.wasSavingPost = isSavingPost;
-		noticeMeta.wasAutosavingPost = isAutosavingPost;
 		noticeMeta.wasPublishingPost = isPublishingPost;
 		noticeMeta.wasStatus = isStatus;
 
@@ -65,11 +53,11 @@ const onboardingHomepageNotice = () => {
 				dispatch( 'core/notices' ).createSuccessNotice(
 					__( 'Your homepage was published.', 'woocommerce-admin' ),
 					{
-						id: 'ONBOARDING_HOME_PAGE_NOTICE',
+						id: 'WOOCOMMERCE_ONBOARDING_HOME_PAGE_NOTICE',
 						actions: [
 							{
 								url: getAdminLink( 'admin.php?page=wc-admin&task=appearance' ),
-								label: 'Continue setup.',
+								label: __( 'Continue setup.', 'woocommerce-admin' ),
 							},
 						],
 					}
--- a/client/wp-admin-scripts/onboarding-homepage-notice/index.js
+++ b/client/wp-admin-scripts/onboarding-homepage-notice/index.js
@@ -1,0 +1,86 @@
+/** @format */
+/**
+ * External dependencies
+ */
+import { select, subscribe, dispatch } from '@wordpress/data';
+import { __ } from '@wordpress/i18n';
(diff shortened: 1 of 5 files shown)


client/wp-admin-scripts/onboarding-homepage-notice/index.js
```diff
@@ -0,0 +1,86 @@
...
+					data: {
+						show_on_front: 'page',
+						page_on_front: post.id,
+					},
+				} );
+
+				dispatch( 'core/notices' ).createSuccessNotice(
+					__( 'Your homepage was published.', 'woocommerce-admin' ),
+					{
+						id: 'ONBOARDING_HOME_PAGE_NOTICE',
```
joshuatf: Do we need to prefix this with `WC_` or `WOOCOMMERCE`?

client/wp-admin-scripts/onboarding-homepage-notice/index.js
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * WooCommerce dependencies
+ */
+import { getAdminLink } from '@woocommerce/navigation';
+
+/**
+ * Displays a notice on page save and updates the hompage options.
+ *
+ * Adapted from WSUWP Help Docs by Washington State University (GPL v2).
```
joshuatf: 🙏

client/wp-admin-scripts/onboarding-homepage-notice/index.js
```diff
@@ -0,0 +1,86 @@
...
+				} );
+
+				dispatch( 'core/notices' ).createSuccessNotice(
+					__( 'Your homepage was published.', 'woocommerce-admin' ),
+					{
+						id: 'ONBOARDING_HOME_PAGE_NOTICE',
+						actions: [
+							{
+								url: getAdminLink( 'admin.php?page=wc-admin&task=appearance' ),
+								label: 'Continue setup.',
```
joshuatf: This string probably needs to be localized.

client/wp-admin-scripts/onboarding-homepage-notice/index.js
```diff
@@ -0,0 +1,86 @@
...
+
+	subscribe( () => {
+		const isSavingPost = select( 'core/editor' ).isSavingPost();
+		const isAutosavingPost = select( 'core/editor' ).isAutosavingPost();
+		const isPublishingPost = select( 'core/editor' ).isPublishingPost();
+		const isStatus = select( 'core/editor' ).getCurrentPost().status;
+
+		// This triggers when the homepage is first published.
+		const shouldTriggerAdminNotice =
+			noticeMeta.wasSavingPost &&
```
joshuatf: I'm not super familiar with all of these states, but it seems like there may be an opportunity to cut down this logic to:
```js
'publish' === isStatus && wasStatus !== isStatus
```

But if this causes issues, feel free to ignore this comment.

src/Features/OnboardingTasks.php
```diff
@@ -134,29 +138,40 @@ public static function check_task_completion( $task ) {
...
+		$messages['product'][6] = sprintf( __( 'You created your first product! <a href="%s">Continue Setup</a>.', 'woocommerce-admin' ), esc_url( wc_admin_url() ) );
+		return $messages;
+	}
+
+	/**
+	 * Hooks into the post page to display a different success notice and sets the active page as the site's home page if visted from onboarding.
+	 */
+	function add_onboarding_homepage_notice_admin_script( $hook ) {
+		global $post;
+		if ( $hook == 'post.php' && 'page' === $post->post_type && isset( $_GET[ 'wc-onboarding' ] ) && 'homepage' === $_GET[ 'wc-onboarding' ] ) { // WPCS: csrf ok.
```
joshuatf: Nitpick refactor: maybe we should create a method `get_active_task()` to check for the `wc-onboarding` get param in case the logic for this changes again.

client/wp-admin-scripts/onboarding-homepage-notice/index.js
justinshreve: Unfortunately, the status changes to `publish` before the post finishes saving, which is too early to clear the other message out.

However, I am able to simplify this down to:

`const shouldTriggerAdminNotice = 'publish' === isStatus && noticeMeta.wasPublishingPost && ! isStatus.isPublishingPost;`

which is much better.

src/Features/OnboardingTasks.php
justinshreve: I've actually backed out my change the the parameter here. There's no reason not to use the same query parameter as `self::ACTIVE_TASK_TRANSIENT`.

justinshreve:
Thanks for the review @joshuatf. Handled your PR feedback above. I was able to simplify the status checks a bit.

PR #3068, opened 2019-10-21
Onboarding: Don't skip start step if TOS not accepted

### LINKED ISSUE #3018: Onboarding: Jetpack is not opted into if Jetpack and WCS are already active
The “start” step of the profiler includes a method to update `woocommerce_setup_jetpack_opted_in` and allow tracking.  This option is also responsible for allowing WCS tools, such as automated taxes, to work properly.

If Jetpack and WCS are already active (but not connected or opted into), this step is skipped and the user never gets opted in, preventing WCS tools from working as `$tos_accepted` is `false`.

https://github.com/Automattic/woocommerce-services/blob/e8cc4ae57ea5c01ee64696cd06a9a4930d2df5e3/woocommerce-services.php#L423

We can probably check if we're opted-in/connected before skipping the start step - https://github.com/woocommerce/woocommerce-admin/blob/109f905ad2ab3bc8755112a6055b88f1a01e8c0c/client/dashboard/profile-wizard/steps/start/index.js#L47

joshuatf:
Fixes #3018 

Doesn't skip the first step of the onboarding profiler if TOS hasn't been accepted yet.

1. Set WCS `tos_accepted` to false. `WC_Connect_Options::update_option( 'tos_accepted', false );`
2. Go to the first step in the Onboarding profiler. `/wp-admin/admin.php?page=wc-admin`
3. Make sure the start step is not skipped to Store Details.
4. Click `Get started`.
5. After proceeding past the start step, visit the start step again.
5. Make sure it's skipped as `tos_accepted` should now be true.

### DIFF
--- a/client/dashboard/profile-wizard/steps/start/index.js
+++ b/client/dashboard/profile-wizard/steps/start/index.js
@@ -162,8 +162,7 @@ class Start extends Component {
 					'Save time at the Post Office by printing USPS shipping labels at home.',
 					'woocommerce-admin'
 				),
-				visible:
-					activePlugins.includes( 'jetpack' ) && ! activePlugins.includes( 'woocommerce-services' ),
+				visible: activePlugins.includes( 'jetpack' ),
 			},
 			{
 				title: __( 'Simple payment setup', 'woocommerce-admin' ),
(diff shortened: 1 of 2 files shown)


client/dashboard/profile-wizard/steps/start/index.js
```diff
@@ -279,16 +280,16 @@ export default compose(
...
+		const options = getOptions( [ 'woocommerce_setup_jetpack_opted_in', 'wc_connect_options' ] );
+		const tosAccepted = get( options, [ 'wc_connect_options' ], {} ).tos_accepted;
 
 		const activePlugins = getActivePlugins();
 		const profileItems = getProfileItems();
 
 		return {
 			isProfileItemsError,
 			activePlugins,
-			allowTracking,
```
justinshreve: Thanks for catching that I left this in there.

joshuatf:
>Because of the getBenefits visible logic, only one benefit ends up getting shown. Should we update this to show all of the WCS benefits?

Good call, @justinshreve - updated to show those benefits.

PR #3086, opened 2019-10-22
Update styles to work with WP 5.3

justinshreve:
There have been CSS changes in 5.3 that have messed up parts of our onboarding layout, and some of the checkboxes on the analytic setting page.

See https://make.wordpress.org/core/2019/10/18/noteworthy-admin-css-changes-in-wordpress-5-3/.

This PR adjusts the styles so everything looks good again.

It also adds a body class, so that things continue to look good in older versions as well.

#### Screenshots / some examples:

Before
Screen Shot 2019-10-22 at 2 06 41 PM

After
Screen Shot 2019-10-22 at 2 07 32 PM

Before
Screen Shot 2019-10-22 at 2 02 40 PM

After
Screen Shot 2019-10-22 at 2 35 08 PM

#### To Test:
* Update your site to WP 5.3 RC 1 or nightly
* Test the various flows and make sure things look OK visually
* Look at the settings analytics page

### DIFF
--- a/client/stylesheets/shared/_global.scss
+++ b/client/stylesheets/shared/_global.scss
@@ -129,3 +129,15 @@
 		display: flex;
 	}
 }
+
+.wp-version-less-than-5-3 {
+	.muriel-input-text.empty input {
+		top: 17px;
+	}
+
+	&.woocommerce-profile-wizard__body .muriel-checkbox input[type='checkbox']:checked::before,
+	.woocommerce-layout input[type='checkbox']:checked::before {
+		margin-left: -2px;
+		margin-top: -3px;
+	}
+}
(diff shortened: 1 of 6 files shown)


justinshreve:
This affects the analytics settings page too, so we probably should add some additional CSS to make this work in both versions.

client/stylesheets/shared/_global.scss
```diff
@@ -110,4 +115,29 @@
...
+		top: 24px;
+	}
+
+	.muriel-checkbox .components-base-control__field {
+		align-items: center;
+		display: flex;
+	}
+}
+
+.wp-version-less-than-5-3 {
```
joshuatf: Instead of introducing a new class, could we update the property that caused the break?

For example, it looks like adding `min-height: auto;` to the muriel input tags will fix the positioning and overlap seen in the images.

```css
.muriel-input-text input {
    min-height: auto;
}
```

justinshreve:
It took me a bit to get the checkbox style working on both versions + (there are some differences between the Newspack checkbox styles and the normal checkbox on the analytics page), but it looks like everything is working in both versions now without the extra CSS class.

Issue #2918, opened 2019-09-12
More Net-Revenue issues -- might be related to issue #2855


JScott-MSI:
**Describe the bug**
Look for a single product sold under the Analytics -> Products report. The report gives Net- Revenue and the total orders and total items sold. 

Run the same report using the Legacy reporting and the Net Revenue is different.

In the Analytics -> Products report, click on the Orders link, which opens into the Analytics -> Orders and the Net-Revenue is different (and matches the Legacy report). If you look at the specific orders listed, in certain orders, the product searched on was fulfilled at no charge and is the difference.

**Expected behavior**
Net Revenue totals should match in the reports... and should match the Legacy reports where appropriate.

**Screenshots**
Screen #1 Shows Legacy Report of Sales by Product
![Products Sold Net legacy](https://user-images.githubusercontent.com/52297130/64744167-9aa32e80-d4b7-11e9-940d-fbadfae1c79d.jpg)

Screen # 2 Shows Analytics-> Product Report - Net Revenue does not match Legacy or Order Drill-down
![Products Sold Net](https://user-images.githubusercontent.com/52297130/64744203-ba3a5700-d4b7-11e9-810d-b7ed23fd4f2d.jpg)

Screen # 3 Shows Drill-down into Orders from the Analytics-> Product report -- Matches Legacy but not Analytics Product Report -- If you add the Retail price of $399.95 for each of the $0.00 orders to the $1,064.90 it totals $2,664.70
![Orders Sold Net_1](https://user-images.githubusercontent.com/52297130/64744243-db02ac80-d4b7-11e9-83e6-12769fe75b8f.jpg)

**Desktop (please complete the following information):**

* OS: IOS
* Browser Firefox
* Version Latest

**Additional context**
This may be related to  issue #2855 where Net Revenue totals did not match.

psealock:
Thanks again @JScott-MSI for the detailed report!

I can reproduce this. It appears Coupons are not calculated correctly in the Product Report.

1. Create an order for a product today and use a coupon to cover 100% of the cost.
2. Revenue > Date Range > Today
3. See the Net revenue correctly display $0.00
4. Products > Single Product > Search for the product purchased
5. See the Net revenue show the cost of the product

JScott-MSI:
[...]

PR #3095, opened 2019-10-23
Address revenue discrepancies

### LINKED ISSUE #2855: Net Revenue numbers don't match
**Describe the bug**
The Net Revenue total on the Revenue page and the Orders page match, however they do not match the Net Revenue on the Product page or the Net Revenue on the Categories page. However the Product and Category page Net revenues match each other.

Woo Admin 1.8.1
Woo 3.7.0

**To Reproduce**
Run the analytics reports

**Expected behavior**
If these are representing Net revenue the total should be the same in all reports, if they are not calculating the same way they need to be labeled in a way that represents what is actually calculated.

**Screenshots**
If applicable, add screenshots to help explain your proble
Revenue Net Discrepencies

* OS: iOS 
* Browser Firefox
* Version Latest

rrennick:
Fixes #2855
See #2918

This PR addresses the revenue discrepancies between Product and Order revenue totals. Categories use Product totals and Revenue uses Order totals.

This PR makes the following changes:
- Re-label the product revenue total on the Product/Categories screens to `Net Product Revenue`

Product revenue does not include coupons, fees, or partial refunds (existing stores may have flat amount refunds not applied to products). If we were to include these amounts in the report total, then the sum of net revenue in the table would not sum to the totals in the header/footer. 

- Exclude `wc-auto-draft` orders from import

The amounts on these orders were being included in the product revenue but excluded from the order revenue.

- Round product line item total amounts on product lookup for tax inclusive pricing. If WC is configured to round tax on the line item, then the tax is also rounded.

This change only affects sites that have tax inclusive pricing. With tax inclusive pricing the line items totals on the order contain unrounded currency fractions. It only takes a few of these to produce a total discrepancy.

- On `master`
- Set WC to tax inclusive pricing
- Set a tax rate of 11% or more
- Use `wc-smooth-generator` to generate 10-20 orders
- Add a partial refund to one of the orders
- Create a manual order and apply a fixed amount coupon
- Create a second order and set the product price to 1/2 of the regular price when adding the line item (for #2918)
- Run the jobs to import the orders
- Load the Product & Order Analytics for today, note discrepancies
- Switch to this PR
- Re-import your data with Skip existing unchecked
- Compare Product & Order Analytics for today

### Changelog Note:

Fix: Address discrepancies in Revenue totals between Analytics screens.

### DIFF
--- a/src/API/Reports/Products/DataStore.php
+++ b/src/API/Reports/Products/DataStore.php
@@ -427,7 +427,7 @@ class DataStore extends ReportsDataStore implements DataStoreInterface {
 				$tax_amount += isset( $tax_data['total'][ $tax_item_id ] ) ? $tax_data['total'][ $tax_item_id ] : 0;
 			}
 
-			$net_revenue = round( $order_item->get_subtotal( 'edit' ), $decimals );
+			$net_revenue = round( $order_item->get_total( 'edit' ), $decimals );
 			if ( $round_tax ) {
 				$tax_amount = round( $tax_amount, $decimals );
 			}
--- a/src/API/Reports/Products/DataStore.php
+++ b/src/API/Reports/Products/DataStore.php
@@ -402,6 +402,8 @@ class DataStore extends ReportsDataStore implements DataStoreInterface {
 
 		$order_items = $order->get_items();
 		$num_updated = 0;
+		$decimals    = wc_get_price_decimals();
+		$round_tax   = 'no' === get_option( 'woocommerce_tax_round_at_subtotal' );
 
 		foreach ( $order_items as $order_item ) {
 			$order_item_id       = $order_item->get_id();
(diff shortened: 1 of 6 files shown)


LevinMedia:
Thanks so much for working through this @rrennick! 

I tested on my end, and I'm now seeing the correct net numbers in the product and category reports! 

My only request would be that you change the label from `net product revenue` to `Net Sales`. 

I've got a [branch](https://github.com/woocommerce/woocommerce-admin/compare/update/change-revenue-to-sales) in the works that will update the labels on all the things, as well as fix the gross, and refund calculations in the revenue report. I'll open a PR for it later today / tonight.

src/API/Reports/Products/DataStore.php
```diff
@@ -402,6 +402,8 @@ public static function sync_order_products( $order_id ) {
 
 		$order_items = $order->get_items();
 		$num_updated = 0;
+		$decimals    = wc_get_price_decimals();
+		$round_tax   = 'no' === get_option( 'woocommerce_tax_round_at_subtotal' );
```
joshuatf: Nice!  I'm wondering if it's worth keeping a list somewhere of all the settings that might affect our data and require re-syncing.

Not something that has to be handled here, but long-term it might be nice to know what requires a re-sync.

rrennick: Good idea. Created #3108 .

PR #3096, opened 2019-10-23
Add tracking to inbox note views.

### LINKED ISSUE #2939: Add new properties to the "wcadmin_activity_panel_open" event
It's necessary to add two additional properties to the `wcadmin_activity_panel_open` event:

- Messages displayed: we need a property that records the messages that are displayed when the user opens the inbox panel. Without that info it's not easy to have an accurate conv. rate of each message, in order to understand if the user is getting value from them.
- Displayed for more than 4 seconds: this property is useful to indicate that someone could have time to read the notification.

jeffstieler:
Fixes #2939.

This PR seeks to introduce a new Tracks event that fires when Inbox Notes are viewed. The ultimate goal is to determine each note's conversion rate.

To accomplish this goal, a new dependency was added: [`react-visibility-sensor`](https://www.npmjs.com/package/react-visibility-sensor).

This PR is based off the work in https://github.com/woocommerce/woocommerce-admin/pull/3039 since it introduces a bespoke `<InboxNoteCard>` component.

- Open your developer console
- Open the Activity Panel, go to Inbox
- Verify that a Tracks event fires for each note visible in the panel
- Scroll down in the panel
- Verify that any newly visible notes have a Tracks event fired for them
- Verify that scrolling back up in the panel does not fire extra events for the previously seen notes

### Changelog Note:

Tweak: track inbox note views.

### DIFF
--- a/client/header/activity-panel/panels/inbox/card.js
+++ b/client/header/activity-panel/panels/inbox/card.js
@@ -1,0 +1,71 @@
+/** @format */
+/**
+ * External dependencies
+ */
+import { Component } from '@wordpress/element';
+import PropTypes from 'prop-types';
+import Gridicon from 'gridicons';
+
+/**
+ * Internal dependencies
+ */
+import { ActivityCard } from '../../activity-card';
+import NoteAction from './action';
+import sanitizeHTML from 'lib/sanitize-html';
+import classnames from 'classnames';
+
+class InboxNoteCard extends Component {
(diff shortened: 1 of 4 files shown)


jeffstieler:
> This will also fire off new events when the menu is closed and reopened. I think that's also expected here, is that right?

I believe so. The original issue prescribed using the panel open event.

PR #3104, opened 2019-10-25
Correcting and clarifying analytics terms and calculations

### LINKED ISSUE #3175: Analytics Terms: Rename Total Sales, Recalculate Gross Sales
Make changes outlined in p90Yrv-1ef-p2 to modify Analytics terms. Previously attempted in https://github.com/woocommerce/woocommerce-admin/pull/3104, the changes will require a database update.

* Rename `gross_total` to `total_sales`
* Create a new column `gross_sales` defined as

```
gross_sales = total_sales + refunds + coupons - tax - shipping
```

* Reintroduce the `coupon_total` column, or figure out how to use the JOIN to calculate the value.

LevinMedia:
Fixes #3175.

### This PR makes changes to three areas of WC-Admin Analytics:

- It updates labels in the UI for clarity and consistency with accounting best practices (see the appendix at the bottom of this PR for updated definitions and calculations.
- It corrects the calculations of "Gross Sales" (Formerly Gross Revenue) and "Returns" (formerly Refunds) in the revenue report and in the dashboard. 
- It adds a new data point: "Total Sales". This number is actually included in the legacy reports, however, it's incorrectly labeled as "Gross Sales", which led to some of the confusion and miss calculation in the new WC-Admin analytics.

#### Technical changes: _-Jeff_

- The `gross_revenue` column has been renamed `total_sales`. A corresponding DB update method has been defined and will need to target the next released version of WC-Admin.
- The "fill in missing segments" functions have been refactored to only backfill fields from the segmentation query, and not the originating report columns (to avoid adding `gross_sales` to segments where it's not needed/expected)

### Additional Notes:

The new `gross_sales` selection is not a stored column value, but rather a calculated value based on the result set for the report query. Because refunds are separate rows in the stats table and not an updated or column value in the original order row, we can't calculate a value pre-insert to store in the database. I'm certainly open to ideas on how to improve this. _-Jeff_

### Detailed Test Instructions:

Testing requires your store has two consecutive days with no orders. If you have an existing test store, trash and delete any test orders from today, and yesterday to get a clean slate. 

**_Note: this will rename a column in your order stats table. You may want to backup your database before following these instructions._**

**Testing a new order:**

- Load up this branch
- Delete or change your `wc_admin_version` option to be less than `0.21.0`
- Load a WooCommerce admin page
- Verify the `gross_revenue` column in your order stats table has been renamed `total_sales`
- From the front end of your store, place a test order **with at least two items**.  It's important that order has all the things too: a coupon, taxes and shipping costs.
- After you place the order, view the order detail screen and change the order's date to yesterday. Be sure to update the order after changing the date. It will take a few seconds for the analytics to record the change in date.
- Next, view the revenue report with the date range set to "Yesterday"
- You should see a new summary number at the top of the report and a new column in the table, both labeled **Total Sales**
- The number in **Total Sales** should match the order total in the order details screen from the order you just placed
- **Gross Sales** should match the combined price of the items purchases before applying the coupon
- **Returns** should be zero
- **Coupons** should display the total amount discounted in the order
- **Net sales** should equate to **Gross Sales - Returns - Coupons**
- **Tax** should display the total tax paid for the order
- **Shipping** should display the total cost of shipping paid for the order
- **Total Sales** should equate to **Net Sales + Tax + Shipping** as previously mentioned, it should match the order total in the order details. 

Example:

Screen Shot 2019-10-24 at 7 52 06 PM

**Testing a refund:**

- Return to the order details screen for the test order. 
- Click refund
- **refund one of the two items** by incrementing the item count up by one
- Note the proportional amount of tax that will be returned to the customer when the item is refunded
- Refund half of the shipping costs
- Process the refund and note the total amount refunded
- View the revenue report set the date range for "Today"
- **Gross Sales** should be zero, as no new orders were placed
- **Returns** should indicate the cost of the refunded item, inclusive of it's share of the original coupon 
- **Coupons** should be zero, as no new orders were placed
- **Net Sales** should match **Returns** per the calculation: **Gross Sales - Returns - Coupons**
- **Taxes** should be a negative number, and match the proportional amount of tax that was refunded
- **Shipping** should be a negative number, and match the amount of shipping that was refunded
- **Total Sales** should be a negative number, and equate to **Net Sales - Taxes - Shipping**
- **Total Sales** should also match the total refund amount noted on the order detail screen.

Example:
Screen Shot 2019-10-24 at 7 52 20 PM
Screen Shot 2019-10-24 at 8 24 06 PM

**Testing aggregate totals:**

Last but not least, we want to look at today and yesterdays data combined.

- In the revenue report, change the date range picker to custom, and select both yesterday and today's date. Hit update. 
- Calculating the summary number should still produce the correct results:  **Gross Sales - returns - coupons = Net Sales** then: **Net Sales + Taxes + Shipping = Total Sales**
- The **Total Sales** amount should match the adjusted total in the order details page

Example:
Screen Shot 2019-10-24 at 7 53 54 PM
Screen Shot 2019-10-24 at 8 24 06 PM

If you want to further test this PR on existing order data, be sure to re-run the historical data importer to get the correct calculations. Also probably get yourself a cold beverage too. 

### Appendix: Updated definitions and calculations

here is some rational behind the updated labels: In the context of  business and analytics, people (myself included) frequently use the terms "Sales" or "Revenue" interchangeably. In in casual conversation they're generally understood to mean the same thing. However, in accounting terms, sales and revenue are two very different things.

> Sales refer to the proceeds received from the selling of goods and services to customers. It represents the contract between the buyer and seller that incorporates transfer of title of goods, in exchange for adequate consideration.
> 
> Revenue indicates the total amount of cash generated by the company from its diverse range of activities. It delineates the income earned from selling goods, delivering services or using capital in any other way, related to the core activities of the business.

Recognizing that income generated in a WooCommerce shop may only be a portion of a business' total revenue, I've update any usage of the word "Revenue" to "Sales".

### Gross Sales (Formerly Gross Revenue)
_Price of product(s) x quantity ordered._

- Gross Sales does not include Returns, Coupons, Taxes, or Shipping.

### Returns (Formerly Refunds)
_Money returned to the customer for a product or service._

- Refunds appear as a negative number on the date that the return occurred.
- Refunds include discounts from coupons.
- Refunded shipping charges and refunded taxes are not included in the returns summary number or column. They appear as negative values in shipping and taxes on the date of the return. (See below)

### Coupons
_Product coupon + the product's proportional share of any cart level coupons._

- The Coupons value combines any line-item coupons with the product's share of a cart-wide coupon.

### Net Sales (Formerly Net Revenue)
_Gross Sales - Refunds - Coupons_

- Net Sales does not include shipping charges or taxes. 
- Assuming there are no other orders on given day: Net Sales will be a positive number on a date that an order is placed, and a negative number on a date that an order is refunded.

### Shipping
_The shipping charges associated with an order or refund._

- Shipping will be a positive number on a date that an order is placed. 
- Shipping will a negative number on a date that an order is refunded.

### Taxes
_The taxes associated with an order or refund._

- Taxes will be a positive number on a date that an order is placed
- Taxes will be a negative number on a date that an order is refunded.

### Total Sales (New in WC-Admin, Previously Gross Sales in legacy reports)
_Net Sales + Taxes + Shipping_

- Assuming there are no other orders on given day: Total Sales will be a positive number on a date that an order is placed, and a negative number on a date that an order is refunded.

### Average Order Value
_Net Sales / Number of orders_

- Eventually may want to add a setting to allow merchants to calculate AOV off of net sales or total sales


jeffstieler:
Tests are failing because the `coupon_total` column was removed from the order stats table - @LevinMedia you and I have old DB schemas 😛.

https://github.com/woocommerce/woocommerce-admin/issues/1942
https://github.com/woocommerce/woocommerce-admin/pull/1975

To fix, we'll either need to reintroduce the column (which was decided to be unnecessary in #1942) or to use the join value in our calculations if possible.

LevinMedia:
@jeffstieler just tested out the branch with the new commits and everything looks like it's in order! Thanks so much! 👍

jeffstieler:
This is ready for review! 🎉

psealock:
>Testing requires your store has two consecutive days with no orders.

Why is this? How will this affect stores in the wild?

Are we not backfilling data?

includes/wc-admin-update-functions.php
```diff
@@ -41,3 +41,20 @@ function wc_admin_update_0201_order_status_index() {
 function wc_admin_update_0201_db_version() {
 	Installer::update_db_version( '0.20.1' );
 }
+
+/**
+ * Rename "gross_total" to "total_sales".
+ * See: https://github.com/woocommerce/woocommerce-admin/issues/3175
+ */
+function wc_admin_update_0210_rename_gross_sales() {
```
psealock: this will probably go out in 0.23.0

includes/wc-admin-update-functions.php
```diff
@@ -41,3 +41,20 @@ function wc_admin_update_0201_order_status_index() {
...
+	global $wpdb;
+
+	$wpdb->query( "ALTER TABLE {$wpdb->prefix}wc_order_stats CHANGE gross_total total_sales DOUBLE" );
```
psealock: This just added `total_sales` with all zeroes instead of renaming `gross_total` which still exists. Is this on purpose?

jeffstieler: No, the intended result is that the column is renamed and the data is the same. What MySQL version are you using?

jeffstieler: Fixed in 551cc31 - though, it's kind of hacky since we shouldn't _really_ be renaming columns in WP (breaks back compat). I could argue that's OK with WC-Admin since we're a feature plugin and haven't made it to core WC yet.

includes/wc-admin-update-functions.php
jeffstieler: Updated version in 551cc31.

jeffstieler:
> > Testing requires your store has two consecutive days with no orders.
> 
> Why is this? How will this affect stores in the wild?

It shouldn't be required, I think @LevinMedia specified that for easier isolation of data and testing.

> Are we not backfilling data? It looks kind of odd

That was related to the bug you pointed out here: https://github.com/woocommerce/woocommerce-admin/pull/3104#discussion_r345490358 - it should be working now.

includes/wc-admin-update-functions.php
rrennick: > I could argue that's OK with WC-Admin since we're a feature plugin and haven't made it to core WC yet.

We probably don't see very many people going back to an earlier version.

LevinMedia:
@psealock @jeffstieler 

>It shouldn't be required, I think @LevinMedia specified that for easier isolation of data and testing.

That's exactly right... not required, just makes it easier to see what's going in for testing.

psealock:
I'm going to make these comments so I can keep track

### Testing a new order:

- [ ] You should see a new summary number at the top of the report and a new column in the table, both labeled Total Sales
- [ ] The number in Total Sales should match the order total in the order details screen from the order you just placed
- [ ] Gross Sales should match the combined price of the items purchases before applying the coupon
- [ ] Returns should be zero
- [ ] Coupons should display the total amount discounted in the order
- [ ] Net sales should equate to Gross Sales - Returns - Coupons
- [ ] Tax should display the total tax paid for the order
- [ ] Shipping should display the total cost of shipping paid for the order
- [ ] Total Sales should equate to Net Sales + Tax + Shipping as previously mentioned, it should match the order total in the order details.

psealock:
On a simple order with tax, coupon (10%), and shipping a few things don't add up.

Screen Shot 2019-11-19 at 3 09 43 PM


> Gross Sales should match the combined price of the items purchases before applying the coupon

This figure should be $339.75, but is $283.13 as seen below. It looks to be taking taxes out, but this isn't exactly adding up either.

>Shipping should display the total cost of shipping paid for the order

This should be $18.12, but is instead $15.10. The coupon was 10%, so Im not sure what happened here.

Screen Shot 2019-11-19 at 3 14 59 PM

Can you reproduce this @LevinMedia ?

jeffstieler:
@psealock I attempted to reproduce, but could not. Can you verify that the necessary DB migrations have taken place? Are there any other specific steps you can list out that we can try?

Here is the only order I've placed today and the report:

Screen Shot 2019-11-19 at 4 14 10 PM

Screen Shot 2019-11-19 at 4 14 25 PM

LevinMedia:
@psealock likewise, I'm also unable to reproduce. I placed this order:

Screen Shot 2019-11-19 at 1 49 23 PM

gross is correct at 175 (the item total + the discount amount in light grey directly below) it all adds up in the report: 

Screen Shot 2019-11-19 at 1 50 59 PM

psealock:
Ok, I see now there is a difference between what the customer sees on the confirmation and what gets reported in the Order page

Screen Shot 2019-11-20 at 4 49 54 PM

This is now checking out, thanks for taking the time to confirm.

psealock:
### Testing a refund:

- [ ] Gross Sales should be zero, as no new orders were placed
- [ ] Returns should indicate the cost of the refunded item, inclusive of it's share of the original coupon
- [ ] Coupons should be zero, as no new orders were placed
- [ ] Net Sales should match Returns per the calculation: Gross Sales - Returns - Coupons
- [ ] Taxes should be a negative number, and match the proportional amount of tax that was refunded
- [ ] Shipping should be a negative number, and match the amount of shipping that was refunded
- [ ] Total Sales should be a negative number, and equate to Net Sales - Taxes - Shipping
- [ ]Total Sales should also match the total refund amount noted on the order detail screen.

psealock:
> Because refunds are separate rows in the stats table and not an updated or column value in the original order row, we can't calculate a value pre-insert to store in the database. I'm certainly open to ideas on how to improve this. -Jeff

So we'd like to eliminate this line, if I understand correctly:

```php
$refunds     = "ABS( SUM( CASE WHEN {$table_name}.net_total < 0 THEN {$table_name}.net_total ELSE 0 END ) )";
```

What if we added columns `shipping_refund`, `tax_refund`, and `net_refund` and a process would run after every refund post addition or alteration that calculates those values based off `parent_id`?  That process would run as part of the import Action Scheduler job. 

This is testing well for me, so maybe a performance improvement like this can be handled separately.

jeffstieler:
> So we'd like to eliminate this line, if I understand correctly

If there was a single row representing an order and all refunds against it, we could eliminate most of the next line as well which is the calculation of `$gross_sales` - only the discount amount would need to be included in an equation to be run on the SQL side.

> maybe a performance improvement like this can be handled separately

To that point, I'm not aware of the exact performance implications of this SQL calculation. I'd assume that only a single `SUM()` and addition/subtraction would outperform this current version, but it might be a micro optimization.

psealock:
>If there was a single row representing an order and all refunds against it, we could eliminate most of the next line as well which is the calculation of $gross_sales

Oh right, of course. 

>To that point, I'm not aware of the exact performance implications of this SQL calculation.

Should we take some measurements? Either way, I'm voting to approve this because its testing well and we can circle back to the calculation.

jeffstieler:
> Should we take some measurements? Either way, I'm voting to approve this because its testing well and we can circle back to the calculation.

I ran an `EXPLAIN` for each query and they're identical. In running the queries a few dozen times each, I saw a roughly ~0.1ms difference (the new query taking about that much longer on average) for my testing dataset.

Using a tool to suggest optimizations (https://www.eversql.com/) picked on the subquery, not any of the newly introduced calculations.

PR #3105, opened 2019-10-25
Add an autofill option to SelectControl

### LINKED ISSUE #3102: Browser autofill/autocomplete issues when filling out the store details page
Reported at p90Yrv-1iA-p2 #comment-2683:

> Browser autocomplete (in Safari 13.0.2 at least) does not populate the Country/State. Let’s make sure this works with all popular OS/browser autocompletes (FWIW I think this is also broken in the current OBW).

joshuatf:
Fixes #3102

Adds an `autofill` prop to `SelectControl` to attempt autofilling a field.  This PR:

* Will add an extra input positioned absolutely behind the base component.  Display and visibility CSS properties prevent autofill from occurring on a field in most browsers.
* Filters existing options in a list, but will not perform any async requests.
* Adds a prop to control the name of the field to encourage the correct autofill information from the browser.
* Selects an option if a single match is found from the list of options.

### Screenshots
Screen Shot 2019-10-25 at 3 46 55 PM

1. Create or find an existing `SelectControl` field with an `autofill` prop.
2.  Create autofill information that matches only one of your options in that `SelectControl`.
3.  Use the browser's autofill feature on another field.
4. Note that the actual `SelectControl` option is set after autofilling.

### DIFF
--- a/client/dashboard/components/settings/general/store-address.js
+++ b/client/dashboard/components/settings/general/store-address.js
@@ -94,6 +94,7 @@ export function StoreAddress( props ) {
 			/>
 
 			<SelectControl
+				autofill="country_state"
 				label={ __( 'Country / State', 'woocommerce-admin' ) }
 				required
 				options={ countryStateOptions }
--- a/client/dashboard/components/settings/general/store-address.js
+++ b/client/dashboard/components/settings/general/store-address.js
@@ -6,7 +6,7 @@ import { __ } from '@wordpress/i18n';
 import { decodeEntities } from '@wordpress/html-entities';
 import { escapeRegExp } from 'lodash';
 import { Fragment } from '@wordpress/element';
-import { useMemo, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { getSetting } from '@woocommerce/wc-admin-settings';
 
 /**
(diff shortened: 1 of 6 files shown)


joshuatf:
@justinshreve Thanks for the review here!

>If I rename the autofill field to state, it prefills just the Ohio text. Makes sense based on the browser doing a guess.

I think this is the downside of WC using a combined country/state field and we may be getting into the weeds trying to solve a highly unique problem we've created.

>Since the component is more general use, maybe it could look for a pattern of something like autofill='country:state', and create a field for each piece of the name?

This is a clever idea.  I've played around with this a bit, but it doesn't significantly increase the logic for this field and some assumptions are made about multiple fields (mostly how the values should be concatenated).

Do you think this type of multi-autofill behavior would be needed anywhere else?  If not, we could consider just added a method to handle this inside the `StoreAddress` component for the country/state field.

justinshreve:
> Do you think this type of multi-autofill behavior would be needed anywhere else? If not, we could consider just added a method to handle this inside the StoreAddress component for the country/state field.

I can’t think of any other specific use cases at the moment except for country/state - so that probably is the best path forward.

joshuatf:
@justinshreve I added a few more commits to this and now pass down a functional autofill component to specific to the country/state field.

PR #3110, opened 2019-10-25
Add currency defaults based on country

### LINKED ISSUE #3101: Onboarding - Set default currency settings based on country
From https://github.com/woocommerce/woocommerce-admin/issues/2768#issuecomment-544628166

> What if instead of asking for the currency upfront we just assume it based on the country and > hard code it?

If the following countries are are selected, we should set smart defaults for the currency, currency position, etc.

* US - dollar
* EU countries - euro
* India - indian rupee
* UK - pound sterling
* Brazil - real
* More 2-3 of the top countries
* Rest of the world - dollar

justinshreve:
Fixes #3101.

This PR sets smart currency defaults based on country for a handful of our top countries and the EU, instead of defaulting to WooCommerce's default of `GBP`. If we don't support the country, USD is defaulted.

* Go to the store details step and select a supported country, or an EU country.
* Continue to the next step.
* Observe the `batch` settings request, or view the currency settings under `WooCommerce > Settings` to make sure the defaults set.
* Select another country, and make sure USD defaults are used.


client/dashboard/utils.js
```diff
@@ -13,3 +13,106 @@ export function getCountryCode( countryState ) {
...
 	return countryState.split( ':' )[ 0 ];
 }
+
+/**
+ * Returns currency data by country/region. Contains code, position, thousands separator, decimal separator, and precision.
+ *
+ * @format
+ * @return {object} Curreny data.
+ */
+export function getCurrencyData() {
```
joshuatf: I wonder if we should move this to the `currency` package.  May be useful there too.

PR #3113, opened 2019-10-25
Always update DB version when plugin is updated.

### LINKED ISSUE #3043: Update Version for releases with no db update
https://github.com/woocommerce/woocommerce-admin/pull/3022 introduced a mechanism to update the version number when a database upgrade is required. This was done for `0.20.1`.

We'll need a way, similar to Core, that bumps the version when there are no database updates.

The logic should also take into account the automation we currently have in place for spreading the version number found in `package.json` during the release process.

https://github.com/woocommerce/woocommerce-admin/blob/32f713a70c9e691294be9d69065ab8f40103d996/bin/update-version.php

jeffstieler:
Fixes #3043.

This PR brings over core WC's "maybe update db" and "needs db update" methods.

It will update the DB version when there are no update callbacks.

- Ensure you are on version `0.20.1`
- Modify `FeaturePlugin.php` to set the version to `0.21.2`
- Load an admin page
- Verify your DB version is now `0.21.2`

### Changelog Note:

Tweak: automatically update DB version during plugin updates.

### DIFF
--- a/src/Install.php
+++ b/src/Install.php
@@ -79,7 +79,7 @@ class Install {
 		self::create_tables();
 		self::create_events();
 		self::create_notes();
-		self::update_db();
+		self::maybe_update_db_version();
 
 		delete_transient( 'wc_admin_installing' );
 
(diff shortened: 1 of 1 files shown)


src/Install.php
```diff
@@ -278,10 +278,35 @@ public static function get_db_update_callbacks() {
 		return self::$db_updates;
 	}
 
+	/**
+	 * Is a DB update needed?
+	 *
+	 * @return boolean
+	 */
+	public static function needs_db_update() {
```
rrennick: This processes the updates but doesn't update the DB version so it ends up being called on every dashboard load.

psealock: I think thats the way its supposed to work as this is the portion that contains the check.

rrennick: Sorry, I should have included the db version I was testing with

```
$ wp option get wc_admin_version
0.19.0
```

I added some debug code to verify that `needs_db_update()` runs on every dashboard load. `wc_admin_version` doesn't get updated to the current version.

jeffstieler:
@psealock have you tested this before and already have completed actions for the upgrade functions? That's where it's possible to get stuck - you have to delete those in order for new ones to get scheduled. This shouldn't impact stores unless they manually change their `wc_admin_version` option.

src/Install.php
rrennick: > and already have completed actions for the upgrade functions? You have to delete those in order for new ones to get scheduled.

That was it.

PR #3116, opened 2019-10-28
Make minor text fixes to onboarding

justinshreve:
This small PR makes some minor text fixes suggested in our internal testing thread.

Optional: Look at the industry and business details step of the wizard. Look at the appearance step.

### DIFF
--- a/client/dashboard/task-list/tasks/appearance.js
+++ b/client/dashboard/task-list/tasks/appearance.js
@@ -190,7 +190,7 @@ class Appearance extends Component {
 				key: 'import',
 				label: __( 'Import demo products', 'woocommerce-admin' ),
 				description: __(
-					'We’ll add some products that it will make it easier to see what your store looks like.',
+					'We’ll add some products that it will make it easier to see what your store looks like',
 					'woocommerce-admin'
 				),
 				content: (
--- a/client/dashboard/task-list/tasks/appearance.js
+++ b/client/dashboard/task-list/tasks/appearance.js
@@ -190,7 +190,7 @@ class Appearance extends Component {
 				key: 'import',
 				label: __( 'Import demo products', 'woocommerce-admin' ),
 				description: __(
-					'We’ll add some products that it will make it easier to see what your store looks like',
+					'We’ll add some products that will make it easier to see what your store looks like',
 					'woocommerce-admin'
 				),
 				content: (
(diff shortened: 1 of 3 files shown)


client/dashboard/task-list/tasks/appearance.js
```diff
@@ -190,7 +190,7 @@ class Appearance extends Component {
 				key: 'import',
 				label: __( 'Import demo products', 'woocommerce-admin' ),
 				description: __(
-					'We’ll add some products that it will make it easier to see what your store looks like.',
+					'We’ll add some products that it will make it easier to see what your store looks like',
```
rrennick: @justinshreve (just checking) The first `it` in the sentence looks like it may have been left behind in an edit as both the following would flow better 

> We’ll add some products that will make it easier to see what your store looks like

> We’ll add some products so that it will make it easier to see what your store looks like

src/Features/Onboarding.php
```diff
@@ -112,7 +112,7 @@ public static function get_allowed_industries() {
 			array(
 				'fashion-apparel-accessories' => __( 'Fashion, apparel & accessories', 'woocommerce-admin' ),
 				'health-beauty'               => __( 'Health & beauty', 'woocommerce-admin' ),
-				'art-music-photography'       => __( 'Art, music, & photography', 'woocommerce-admin' ),
+				'art-music-photography'       => __( 'Art, music & photography', 'woocommerce-admin' ),
```
joshuatf: Nitpicky, but personally, I'm a fan of the Oxford comma to reduce ambiguity.  This sounds a bit like `music & photography` are grouped together.

Is there a thread on this somewhere or precedent for this style?

justinshreve: I personally prefer the oxford comma as well. This request came in through p90Yrv-1iA-p2 so that it would match `Fashion, apparel & accessories` and `Home, furniture & garden`

However, I just searched the Field Guide and both our WP.com and A8C style guides say we should use the oxford comma -- so I'll go ahead and update these before merge.

client/dashboard/task-list/tasks/appearance.js
justinshreve: Thanks for catching that! - went with `We’ll add some products that will make it easier to see what your store looks like`

PR #3119, opened 2019-10-28
Show "add a product" notice to stores without published products

### LINKED ISSUE #2903: WC-Admin Inbox message: Add your first product
**User Story**
As a merchant
I want to receive suggestions of what I should do after activating the plugin
So I can get my shop up and running as quickly as possible

**Hypothesis**
While we don't deploy the task list, we can use the inbox notifications to guide the user towards the most important actions and we can start with "Adding a product".

**Acceptance criteria**
- Display the inbox message about adding a product
- Redirect the user to the Products' area after he clicks on the button
- Display the message just once, for users that don't have any product:
New users: 1h after registration
Existing users: when they login

The details about the inbox message are in the WC-admin Inbox messages file (line 20)

**Additional context**

It's important to have this deployed in the upcoming WC Admin release, as it would allow us to get useful data before releasing the task list and most likely it's going to increase the number of products added during the first days.

jeffstieler:
Fixes #2903 

This PR introduces a new note that prompts stores without products to add one.

Additionally, this PR contains a fix for the primary button styles in Inbox Actions.

#### Questions:

* What is the interplay with this and the OBW Task List? We probably shouldn't show this notice if the task list is being used.
* Should the notice dismiss when the CTA is clicked, or should we try to detect when the product was added?

### Screenshots

Screen Shot 2019-10-28 at 11 33 53 AM

- Ensure your store has no published products
- Use Crontrol to run the wc_admin_daily job
- Verify the "add a product" note is shown in the inbox
- Verify the primary button is styled correctly
- Verify the CTA takes you to the new product edit screen

### Changelog Note:

Enhancement: prompt stores to add their first product(s).

### DIFF
--- a/src/Notes/WC_Admin_Notes_Add_First_Product.php
+++ b/src/Notes/WC_Admin_Notes_Add_First_Product.php
@@ -1,0 +1,56 @@
+<?php
+/**
+ * WooCommerce Admin Add First Product Note Provider.
+ *
+ * Adds a note to the merchant's inbox prompting them to add their first product.
+ *
+ * @package WooCommerce Admin
+ */
+
+namespace Automattic\WooCommerce\Admin\Notes;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * WC_Admin_Notes_Add_First_Product
+ */
+class WC_Admin_Notes_Add_First_Product {
+	/**
+	 * Name of the note for use in the database.
+	 */
+	const NOTE_NAME = 'wc-admin-add-first-product';
+
+	/**
+	 * Possibly add the note.
+	 */
+	public static function possibly_add_first_product_note() {
+		// Only show the note to stores without products.
+		$products = wp_count_posts( 'product' );
+
+		if ( 0 < (int) $products->publish ) {
+			return;
+		}
+
+		$data_store = \WC_Data_Store::load( 'admin-note' );
+
+		// We already have this note? Then exit, we're done.
+		$note_ids = $data_store->get_notes_with_name( self::NOTE_NAME );
+		if ( ! empty( $note_ids ) ) {
+			return;
+		}
+
+		$content = __( 'Grow your revenue by adding products to your store. Add products manually, import from a sheet, or migrate from another platform.', 'woocommerce-admin' );
+
+		$note = new WC_Admin_Note();
+		$note->set_title( __( 'Add your first product', 'woocommerce-admin' ) );
+		$note->set_content( $content );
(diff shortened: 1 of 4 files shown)


rrennick:
> Should the notice dismiss when the CTA is clicked, or should we try to detect when the product was added?

Could we mark the note as actioned when the first product was published?

psealock:
cc @pmcpinto for the questions above

src/Notes/WC_Admin_Notes_Add_First_Product.php
```diff
@@ -0,0 +1,56 @@
...
+
+		$data_store = \WC_Data_Store::load( 'admin-note' );
+
+		// We already have this note? Then exit, we're done.
+		$note_ids = $data_store->get_notes_with_name( self::NOTE_NAME );
+		if ( ! empty( $note_ids ) ) {
+			return;
+		}
+
+		$content = __( 'Grow your revenue by adding products to your store. Add products manually, import from a sheet, or migrate from another platform.', 'woocommerce-admin' );
```
timmyc: `import from a sheet` seems like an odd choice of words here. Should it be import from `csv` or `spreadsheet` at least?

pmcpinto:
> What is the interplay with this and the OBW Task List? We probably shouldn't show this notice if the task list is being used.

The idea is to have this notification as a workaround to help users adding their first product, until the obw task list is available to 100% of new users. Initially (JAN) the task list is only going to be available to 10% of new users.

> Should the notice dismiss when the CTA is clicked, or should we try to detect when the product was added?

It would be better when the product is added, but if overcomplicates too much we can go with when the CTA is clicked.

src/Notes/WC_Admin_Notes_Add_First_Product.php
jeffstieler: This is the same copy used in the OBW task list. Perhaps we should modify both areas in a separate PR?

jeffstieler:
@pmcpinto @timmyc re: removing the note after a product is added - the least _(negatively)_ impactful way to do so is in our daily cron job (that sets up most of these notes). If that is too long of a delay we can leave the current behavior of dismissing on click.

pmcpinto:
> removing the note after a product is added - the least (negatively) impactful way to do so is in our daily cron job (that sets up most of these notes). If that is too long of a delay we can leave the current behavior of dismissing on click.

Ok, so I think it's better to simplify and dismiss it on click. Sounds good? @jeffstieler @timmyc

Issue #2959, opened 2019-09-25
Onboarding: Fix update + await pattern


justinshreve:
via https://github.com/woocommerce/woocommerce-admin/pull/2897#discussion_r321569096

> I think this await pattern we've been using will not work with wc-api. I'm not sure it returns the promise and isSettingsError will not yet have updated.

joshuatf:
It seems like we can use `@wordpress/data-controls` as one solution to this problem (big thanks to @nerrad for pointing out this option!).  https://developer.wordpress.org/block-editor/packages/packages-data-controls/

```js
import { dispatch } from '@wordpress/data-controls';
 
// Action generator using dispatch
export function* myAction() {
    yield dispatch( 'core/edit-post', 'togglePublishSidebar' );
    // do some other things.
}
```

While this solution can work, it seems like the recommended pattern here is to create a file of controls (essentially action creators) and process any other actions after the dispatch from these controls.  This does not seem to allow us to use any kinds of callbacks after the action directly within our components (please correct me if I'm wrong @nerrad).

The main limitation to this seems to be the use fresh data with the generic store.  We'll need to register our store via `registerStore` instead of the current `registerGenericStore` in order to add the controls.

More discussion on that in p1573241460032300-slack-isotope-meetings

I think using Redux thunks would be a more ideal solution, allowing us to directly get the promise inside our component without the need to create a new set of action creators/controls.

```js
updateOptions( ... ).then( () => {
    // do something
} );
```

However, this solution requires us to `applyMiddleware` to the store, which we don't appear to have access to via `wp.data`.

In the meantime, I'm going to un-assign myself from this issue.  We have workarounds that circumvent the current issues with updating data, but long-term it would be great to have a more direct way to know when a specific promise in the store has been resolved.

timmyc:
@joshuatf is this good to close now?

PR #3121, opened 2019-10-29
Onboarding: Redirect to wccom cart if purchases required

### LINKED ISSUE #2160: Onboarding - Profile Wizard: Pass cart items to WooCommerce.com
If a user has decided to purchase extensions or a theme in the previous steps, they will be taken to WooCommerce.com with a pre-loaded cart. WooCommerce.com should also store which site has sent them in a cookie or other value, so that they can be redirected after standard purchase.

This task is for tracking any logic needed for pre-adding the cart items, and for redirecting back to `wp-admin` after purchase. This will require collaborating with Alpha. 

joshuatf:
Fixes #2160 

Redirects to the wccom cart if product types warrant or theme purchase is required.

**Note:** I've experienced some issues with the theme not getting updated in time since it's the last step.  This is related to https://github.com/woocommerce/woocommerce-admin/issues/2959 so will address that in a separate PR.

### Screenshots
Screen Shot 2019-10-29 at 11 23 32 AM

1. Optionally set up a wccom testing site.  https://github.com/Automattic/woocommerce.com
2. Enable the profiler.
3. Walk through the steps, adding some product types that require purchase and a non-installed theme.
4. After choosing a theme, check that you're redirected to the wccom cart/checkout.  
5.  Replace with the wccom testing site URL/TLD if set up.
6.  Note the necessary items in the cart (keep in mind the theme bug mentioned above).
7. Completing the checkout should redirect back to the store dashboard.
8. Re-enable the profiler.
9. Choose product types and a theme that don't require purchase.
10. Note that no redirect occurs.

### DIFF
--- a/client/dashboard/index.js
+++ b/client/dashboard/index.js
@@ -8,6 +8,11 @@ import { compose } from '@wordpress/compose';
 import { get } from 'lodash';
 
 /**
+ * WooCommerce dependencies
+ */
+import { updateQueryString } from '@woocommerce/navigation';
+
+/**
  * Internal dependencies
  */
 import './style.scss';
@@ -29,6 +34,7 @@ class Dashboard extends Component {
 		const prevProfileItems = get( prevProps, 'profileItems', {} );
 
 		if ( profileItems.completed && ! prevProfileItems.completed ) {
+			updateQueryString( {}, '/', {} );
 			this.redirectToCart();
 		}
 	}
--- a/client/dashboard/index.js
+++ b/client/dashboard/index.js
@@ -16,6 +16,14 @@ import ProfileWizard from './profile-wizard';
 import withSelect from 'wc-api/with-select';
 
 class Dashboard extends Component {
+	constructor() {
+		super( ...arguments );
+
+		this.state = {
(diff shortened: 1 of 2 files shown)


joshuatf:
Note that most of this logic gets moved in https://github.com/woocommerce/woocommerce-admin/pull/3124 but the `getProductIds` and `redirectToStore` functions stay the same.

client/dashboard/index.js
```diff
@@ -14,6 +21,59 @@ import ProfileWizard from './profile-wizard';
...
+		}
+	}
+
+	getProductIds() {
+		const productIds = [];
+		const profileItems = get( this.props, 'profileItems', {} );
+
+		profileItems.product_types.forEach( product_type => {
+			if (
+				wcSettings.onboarding.productTypes[ product_type ] &&
```
justinshreve: Instead of using `wcSettings` directly, this should be using `getSetting`. I.e. something like

`import { getSetting } from '@woocommerce/wc-admin-settings';`
`const onboarding = getSetting( 'onboarding', {} );`

PR #3129, opened 2019-10-29
Correctly offset for week start annual date changes

### LINKED ISSUE #2888: Last day of "Week to Date" range has incorrect analytics for the past year period
Description:
When using "Week to Date" range and comparing with 'Previous Year', both Gross Revenue and Net revenue are incorrect for the last date in the range for the past year. For example, if you look at Gross Revenue on Sept 5th 2019, use the 'Week to Date' range and compare it with previous year, analytics for Sept 5th 2018 are incorrect. When using "Month to Date" range the analytics for Sept 5th 2018 are correct.

To reproduce:
1. Make sure you have past data
2. Make some sales
3. Use "Week to Date" range to look at the revenue and compare with last year
4. Write down the revenue for the last date in range of the previous year. For example if testing this on Sept 5th 2019, write down the value for Sept 5th 2018.
5. Use "Month to Date" range
6. The analytics for the last date in range of previous year differs and is the correct one. Using the example from above, Sept 5th 2018 will now display correct analytics.

Expected Behaviour:
In both cases the analytics should match as we are looking at the same date. In our example Sept 5th 2018.

OS: Mojave 10.14.6
Browser: Chrome 76

rrennick:
Fixes #2888

The day of the week for a given date changes year to year. The `Week to Date` calculates the previous year's week uses the same date range. This week started on Monday Oct 28, 2019. The same week last year was Monday Oct 29, 2019.

This PR updates the dashboard to request and label the previous year's week using the correct dates for the previous year.

- Use smooth generator to create 20 orders between the past Sunday and today and 20 orders for the same date range in 2018 **plus** one day on the end
- In `master` load `This Month` and note the chart amount for 1 year ago today
- Switch to `This Week` and compare the amount for 1 day ago yesterday to the previous
- Switch to this branch
- `This Month` should show the same
- `This Week` should have the amounts in the same locations but labelled with the correct dates 

### Changelog Note:

Fix: Apply correct date labels to previous year for This Week comparison.

### DIFF
--- a/packages/date/src/index.js
+++ b/packages/date/src/index.js
@@ -156,9 +156,7 @@ export function getLastPeriod( period, compare ) {
 			'week' === period
 				? primaryStart
 						.clone()
-						.subtract( 1, 'years' )
-						.week( primaryStart.week() )
-						.startOf( 'week' )
+						.subtract( 364, 'days' )
 				: primaryStart.clone().subtract( 1, 'years' );
 		secondaryEnd = secondaryStart.clone().endOf( period );
 	}
@@ -193,9 +191,7 @@ export function getCurrentPeriod( period, compare ) {
 			'week' === period
 				? primaryStart
 						.clone()
-						.subtract( 1, 'years' )
-						.week( primaryStart.week() )
-						.startOf( 'week' )
+						.subtract( 364, 'days' )
 				: primaryStart.clone().subtract( 1, 'years' );
 		secondaryEnd = secondaryStart.clone().add( daysSoFar, 'days' );
 	}
@@ -348,12 +344,16 @@ export const getDateDifferenceInDays = ( date, date2 ) => {
  * @param {String|Moment.moment} date2 - secondary start
  * @param {String} compare - `previous_period`  or `previous_year`
  * @param {String} interval - interval
+ * @param {string} period - period value, ie `week`
  * @return {Moment.moment}  - Calculated date
  */
-export const getPreviousDate = ( date, date1, date2, compare, interval ) => {
+export const getPreviousDate = ( date, date1, date2, compare, interval, period ) => {
 	const dateMoment = moment( date );
 
 	if ( 'previous_year' === compare ) {
+		if ( 'week' === period ) {
+			return dateMoment.clone().subtract( 364, 'days' );
+		}
 		return dateMoment.clone().subtract( 1, 'years' );
 	}
 
--- a/client/analytics/components/report-chart/index.js
+++ b/client/analytics/components/report-chart/index.js
@@ -16,6 +16,7 @@ import {
 	getAllowedIntervalsForQuery,
 	getCurrentDates,
 	getDateFormatsForInterval,
+	getDateParamsFromQuery,
 	getIntervalForQuery,
 	getChartTypeForQuery,
 	getPreviousDate,
@@ -74,6 +75,7 @@ export class ReportChart extends Component {
 	getTimeChartData() {
 		const { query, primaryData, secondaryData, selectedChart } = this.props;
 		const currentInterval = getIntervalForQuery( query );
+		const { period } = getDateParamsFromQuery( query );
 		const { primary, secondary } = getCurrentDates( query );
 
 		const chartData = primaryData.data.intervals.map( function( interval, index ) {
@@ -82,7 +84,8 @@ export class ReportChart extends Component {
 				primary.after,
 				secondary.after,
 				query.compare,
-				currentInterval
+				currentInterval,
+				period
 			);
 
 			const secondaryInterval = secondaryData.data.intervals[ index ];


packages/date/src/index.js
```diff
@@ -156,9 +156,7 @@ export function getLastPeriod( period, compare ) {
 			'week' === period
 				? primaryStart
 						.clone()
-						.subtract( 1, 'years' )
-						.week( primaryStart.week() )
-						.startOf( 'week' )
+						.subtract( 364, 'days' )
```
timmyc: Will this still work okay in leap years?

rrennick: > Will this still work okay in leap years?

Yes because it's an even 52 weeks. On a leap year the dates will be 2 days apart instead of 1 but it will still be the same week number (eg. this week is week 42).

timmyc:
I'm still seeing the bug in my testing... though performing it today, Nov 1, so having to use "Last Month" preset vs "Week to Date" so not sure if that plays into this at all. But "Last year"'s value for October 31 does not match when looking at "Last Month" and "Week to Date"

psealock:
I'm still seeing the error as well. 

I think I have a clue as to whats happening here: The data is correct, but the chart is misrendering date ranges for last year's week.

Looking at the client:

Screen Shot 2019-11-06 at 2 16 22 PM

Last year's week-to-date corresponds to the week number (42, lets say) of 2018, ie November 4 - 7 2018 where as today's date is Nov 6. This was logic @LevinMedia and I went through if I remember correctly.

The API correctly returns the right number for Nov 6, 2018

Screen Shot 2019-11-06 at 2 24 22 PM

But the chart is rendering that data as November 5, 2018

Screen Shot 2019-11-06 at 2 16 36 PM

Specifically, the issue is the chart takes the first series and uses its time/date data for every subsequent series. This fails when the DD MM dates don't match. I suspect we may have something similar happen when comparing dates including Feb 29th.

One fix would be to make the chart render dates for each series according to its own dates data, not the first series' dates. In other words, November 5, 2019 would be compared to November 6th, 2018.

Another fix would be to just calculate last year's week according to date, not week number. Looking at week-to-date compares against a period ending today minus 1 year. I think this is what you are trying to do here @rrennick, but I'm not sure why its not working. 

I think the better approach is the first one, as that shortcut may present problems in the future, ie leap year.

rrennick:
> The data is correct, but the chart is misrendering date ranges for last year's week.

Yes, that's the issue. This PR does not move the data point on the graph. It corrects the labels on last year's data. Here are the screenshots when I switched to this branch this morning:

Screen Shot 2019-11-06 at 8 40 43 AM

Screen Shot 2019-11-06 at 8 41 19 AM

rrennick:
The hover didn't show in the previous
Screen Shot 2019-11-06 at 8 41 04 AM

rrennick:
I updated the test instructions to generate orders for 1 day longer in 2018. Here's my graph with those orders
Screen Shot 2019-11-06 at 9 02 17 AM

rrennick:
> though performing it today, Nov 1, so having to use "Last Month" preset vs "Week to Date" so not sure if that plays into this at all.

The PR is addressing the discrepancy in the `Week to Date` comparison between the current year and last year.

psealock:
I spoke with @LevinMedia about this one. We decided to simply line up the dates with previous year. In other words make the dates absolute from one year against the last. So if "week to date" has three days, the previous year will be 3 days ending with today's date last year. Make sense?

This way, the chart is straightforward and dates line up. There isn't much value in comparing week 42 from this year to week 42 from last year. The value is more from "this week v last week".

rrennick:
> Make sense?

Yes. It's possible doing that will only involve making changes on the API side. If that's the case I'll create a separate PR and close this one.

rrennick:
I've been reconsidering this

> The value is more from "this week v last week".

This PR only changes the alignment of year to year weeks.

> We decided to simply line up the dates with previous year. In other words make the dates absolute from one year against the last.

Here are a couple graphs from different date ranges from a live store for transaction totals by day of week:

Screen Shot 2019-11-11 at 11 47 19 AM
Screen Shot 2019-11-11 at 11 46 35 AM

For a lot of stores comparing a Saturday from one year to a Sunday in the next will make the graph of little value (and confusing for some users).

timmyc:
It seems like we have kind of drifted a bit in this PR - seems like we still have a valid, reproducible bug in #2888 - is that something we can still fix in this branch or perhaps I was just testing incorrectly last week?

rrennick:
> I was just testing incorrectly last week?

If you followed the original testing steps the last day of the previous year had a total of $0. I updated the first step so there would be orders on that day.

psealock:
>For a lot of stores comparing a Saturday from one year to a Sunday in the next will make the graph of little value (and confusing for some users).

Indeed thats true. @LevinMedia and I considered this in our chat, but concluded that having offset lines is more confusing for the majority of users.

psealock:
>It seems like we have kind of drifted a bit in this PR 

The implementation is probably straightforward, its just the dates logic that is kind of complicated because there are several possible scenarios.

rrennick:
Superseded by #3271

PR #3135, opened 2019-10-30
Fix label positioning for SelectControl in IE11

### LINKED ISSUE #3071: SelectControl styling problems in IE11
**Describe the bug**
In IE11 the `SelectControl` has some styling problems with the label and input field.

**To Reproduce**

1. Visit the devdocs or any page using the `SelectControl` component.
2. Note the styling issues with the label and input field.

**Expected behavior**
Expected the field to behave the same as in other browsers.

**Screenshots**
Screen Shot 2019-10-21 at 5 51 52 PM

joshuatf:
Fixes #3071 

Adds specific positioning to the label to prevent an overlap that occurs between the label and input on IE11.

### Screenshots
Screen Shot 2019-10-30 at 7 19 00 PM

1. Open the devdocs or "Store Details" page in the onboarding profiler.
2. Make sure the `SelectControl` label/inputs are styled correctly before and after focus as well as when values are added.

### DIFF
--- a/packages/components/src/select-control/style.scss
+++ b/packages/components/src/select-control/style.scss
@@ -31,6 +31,8 @@
 
 		.components-base-control__label {
 			position: absolute;
+			top: 50%;
+			transform: translateY(-50%);
 			color: $studio-gray-50;
 			font-size: 16px;
 		}


joshuatf:
@justinshreve I wasn't able to replicate using Browserstack using IE11.  Did you notice any console errors?

From what I've seen in the past, the search icon showing that way is usually because the material icons font wasn't loaded.  Maybe we can handle that in a follow-up if the issue is persisting for you.

justinshreve:
> Did you notice any console errors?

> From what I've seen in the past, the search icon showing that way is usually because the material icons font wasn't loaded. Maybe we can handle that in a follow-up if the issue is persisting for you.

Yeah it does look like the font isn't loading correctly. Getting a certificate error which I think means my VM is messed up. I just tested with browser stack and things look good. 👍

joshuatf:
Thanks for the review, @justinshreve!  Can you give this a ✅ if things look good here?

Issue #3067, opened 2019-10-20
Onboarding: Unable to navigate back to wizard, responses not saved


stephhjacq:
**Describe the bug**
I tested the new onboarding starting with the Profile Wizard. 

During the wizard, I selected that I would like to use Subscriptions (as well as physical products and downloads), and was expecting some sort of purchase flow as this was priced at $199, but this selection seemed to do nothing.

![https://cld.wthms.co/qd7ceg](https://cld.wthms.co/qd7ceg+)
Full Image: https://cld.wthms.co/qd7ceg+

I selected Mailchimp, but nothing was installed.

![https://cld.wthms.co/D1Qzo7](https://cld.wthms.co/D1Qzo7+)
Full Image: https://cld.wthms.co/D1Qzo7+

![https://cld.wthms.co/EzXjwE](https://cld.wthms.co/EzXjwE+)
Full Image: https://cld.wthms.co/EzXjwE+

After completing the wizard I was sent to the task list, where I did the first task, `Create home page`. Once that was published, I could not return to the wizard.

Unclear where to go from here:

![https://cld.wthms.co/plxctT](https://cld.wthms.co/plxctT+)
Full Image: https://cld.wthms.co/plxctT+

Going to WooCommerce > Dashboard takes me back to the beginning of the profile wizard, which has not recorded any of my responses:

![https://cld.wthms.co/kT8zrT](https://cld.wthms.co/kT8zrT+)
Full Image: https://cld.wthms.co/kT8zrT+

I tried using the other flow, WooCommerce > Orders > Help > Setup Wizard, however the Orders screen has no help menu.

![https://cld.wthms.co/Q2CfWP](https://cld.wthms.co/Q2CfWP+)
Full Image: https://cld.wthms.co/Q2CfWP+

I found the Help menu under Products, and was able to find the task list by disabling the profile wizard.

I was able to continue with the task list, however I ran across the following error when connecting Stripe, you must be an admin:

![https://cld.wthms.co/T5zNZg](https://cld.wthms.co/T5zNZg+)
Full Image: https://cld.wthms.co/T5zNZg+

I have two profiles on Stripe.com under this account - one for WooCommerce/Bubblestorm and one for my test site. Not sure if that makes me an edge-case, maybe selecting the profile is not required for the average user. 

[...]

PR #3153, opened 2019-10-31
Onboarding: Pass connection parameters to install extensions

### LINKED ISSUE #2477: Onboarding - Task List: Installing Extensions
After a store is successfully connected (i.e., store owner has logged in), they are redirected back to their site to install extensions.

This step will involve receiving extension data, downloading to the site, and activating those extensions.

Alpha has a near working implementation of this so this may just be a matter of hooking things up to that API.

@mattyza When you have a working example of this, would you mind posting a link to that code here?

Screen Shot 2019-06-21 at 4 18 39 PM

justinshreve:
Closes #2477.

This PR wires up the cart URL with the correct connection parameters, so that the in-app purchase flow works after purchasing. 

Note that there are a few issues with the flow (https://github.com/Automattic/woocommerce.com/issues/6617, https://github.com/Automattic/woocommerce.com/issues/6618) but extensions and themes do install and activate correctly.

No matter what I pass for `wccom-back`, this also seems to only redirect to the plugins screens.

@joshuatf Unfortunately, since I have to generate the cart URL server side (nonce), I had to add back in a pending screen. The screen only displays if we actually have products to purchase, and it should be relatively quick. Sorry for the back and forth on that.

### Screenshots

Screen Shot 2019-10-31 at 3 14 30 PM

Screen Shot 2019-10-31 at 3 46 54 PM

Screen Shot 2019-10-31 at 3 48 59 PM

Screen Shot 2019-10-31 at 3 49 36 PM

* Go through the profiler, selecting options that will require purchase (product types and/or a non-installed theme).
* Make sure the theme step redirects to the cart after a brief loading screen.
* Continue purchasing the products on WooCommerce.com, fully completing the purchase flow.
* Click the `Add to site` button.
* Sit through the installation steps.
* You should be redirected back to the plugins screen (temporary for now).
* Visit the dashboard and make sure the profile wizard no longer shows.
* Re-enable the profiler and walk through the steps selecting options that don't require purchase and selecting the currently active theme.
* Note that the wccom redirect is skipped (you shouldn't see the pending screen at all) and the profiler is complete since the theme is already active.

### DIFF
--- a/client/dashboard/profile-wizard/steps/theme/index.js
+++ b/client/dashboard/profile-wizard/steps/theme/index.js
@@ -3,13 +3,13 @@
  * External dependencies
  */
 import { __, sprintf } from '@wordpress/i18n';
-import { addQueryArgs } from '@wordpress/url';
+import apiFetch from '@wordpress/api-fetch';
 import { Button } from 'newspack-components';
 import { Component, Fragment } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import { decodeEntities } from '@wordpress/html-entities';
 import { get } from 'lodash';
-import { getAdminLink, updateQueryString } from '@woocommerce/navigation';
+import { getAdminLink } from '@woocommerce/navigation';
 import Gridicon from 'gridicons';
 import { TabPanel, Tooltip } from '@wordpress/components';
 import { withDispatch } from '@wordpress/data';
@@ -17,13 +17,14 @@ import { withDispatch } from '@wordpress/data';
 /**
  * WooCommerce dependencies
  */
-import { Card, H } from '@woocommerce/components';
+import { Card, H, Spinner } from '@woocommerce/components';
 import { getSetting } from '@woocommerce/wc-admin-settings';
 
 /**
  * Internal depdencies
  */
 import withSelect from 'wc-api/with-select';
+import { WC_ADMIN_NAMESPACE } from 'wc-api/constants';
 import './style.scss';
 import { recordEvent } from 'lib/tracks';
 import ThemeUploader from './uploader';
@@ -38,6 +39,7 @@ class Theme extends Component {
 			activeTab: 'all',
 			chosen: null,
 			demo: null,
+			cart: false,
 			uploadedThemes: [],
 		};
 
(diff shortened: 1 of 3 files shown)


client/dashboard/profile-wizard/steps/theme/index.js
```diff
@@ -38,6 +39,7 @@ class Theme extends Component {
 			activeTab: 'all',
 			chosen: null,
 			demo: null,
+			cart: false,
```
joshuatf: It took me a second to look through and find what this property was used for.  Maybe we could rename to `isRedirecting` or `isRedirectingToCart` for slightly faster grokking.

src/API/OnboardingPlugins.php
```diff
@@ -416,6 +429,34 @@ public function finish_wccom_connect( $rest_request ) {
...
+	 * Generates a cart URL for purchasing products from WooCommerce.com.
+	 * Adds the connection nonce and other connection params so extensions and themes can be automatically installed after purchase.
+	 *
+	 * @param  object $rest_request Request details.
+	 * @return array Cart URL for WooCommerce.com
+	 */
+	public function wccom_cart( $rest_request ) {
+		// Make sure the profile wizard is marked completed when we return.
+		$profile = array_merge( get_option( 'wc_onboarding_profile', array() ), array( 'completed' => true ) );
+		update_option( 'wc_onboarding_profile', $profile );
```
joshuatf: This feels a bit outside the scope of this function and while unlikely, if there is an error below we'll get a completed profiler without redirecting.  Is it possible to keep the `updateProfileItems` that marks completed on the client-side?

src/API/OnboardingPlugins.php
```diff
@@ -416,6 +429,34 @@ public function finish_wccom_connect( $rest_request ) {
...
+
+		$cart_url = add_query_arg(
+			array(
+				'wccom-site'          => site_url(),
+				'wccom-back'          => $rest_request['wccom-back'],
+				'wccom-replace-with'  => $rest_request['wccom-replace-with'],
+				'wccom-woo-version'   => WC_VERSION,
+				'wccom-connect-nonce' => wp_create_nonce( 'connect' ),
```
joshuatf: I'm not opposed to generating this URL server-side if you think it's better, but as an alternative, we could pass the nonce to the component settings so we can grab via `wcSettings.onboarding`.

justinshreve: I originally went this route worried about the nonce expiring before the user got to this part of the wizard, but apparently nonces last a lot longer than I thought:

> By default, a nonce has a lifetime of one day.
>
> WordPress uses a system with two ticks (half of the lifetime) and validates nonces from the current tick and the last tick. In default settings (24h lifetime) this means that the time information in the nonce is related to how many 12h periods of time have passed since Unix epoch. This means that a nonce made between midday and midnight will have a lifetime until midday the next day. The actual lifetime is thus variable between 12 and 24 hours. 

I think we'll have less problems doing this client side. I'll rework this PR today.

justinshreve:
@joshuatf I reworked this in https://github.com/woocommerce/woocommerce-admin/pull/3153/commits/1e8eb4b2670f228ed3bebaba3b1bf42df59f0d0e so the URL is built client side again. I kept around the slight refactor I did that I believe fixes https://github.com/woocommerce/woocommerce-admin/issues/3067#issuecomment-548545478.

justinshreve:
@joshuatf Since you have a local install, you could try bundling a zip of this branch (see the feature flags documentation) and test on a Jurassic Ninja site.

src/Loader.php
```diff
@@ -595,6 +595,8 @@ public static function add_component_settings( $settings ) {
 		// @todo On merge, once plugin images are added to core WooCommerce, `wcAdminAssetUrl` can be retired,
 		// and `wcAssetUrl` can be used in its place throughout the codebase.
 		$settings['wcAdminAssetUrl'] = plugins_url( 'images/', dirname( __DIR__ ) . '/woocommerce-admin.php' );
+		$settings['wcVersion'] = WC_VERSION;
```
joshuatf: Minor, but these = could be indented. (Not sure about you, but my vendor scripts like phpcs are also not working right now).

justinshreve:
Reported the cart url at p1572972516182600-slack-woocommerce-com

PR #3154, opened 2019-11-01
Onboarding: Allow users to navigate backwards in steps

### LINKED ISSUE #3136: Store Profiler: allow users to go back to the previous steps
In the store profiler we should allow users to navigate to previous steps, so they can fine-tune and double-check any information that they provided.

The easiest way seems to be adding a link to each step at the navbar, after they finish it:

Annotation on 2019-10-30 at 11-56-54

joshuatf:
Fixes #3136

Adds an `onClick` function to steps so we can allow users to navigate backwards in steps.

**Question:** This only allows going backwards and while I don't think we generally want to allow going forward and skipping steps, we could instead keep track of the latest step they've finished so they can go back and then back forward to the most recent place they've completed.

### Screenshots
Screen Shot 2019-11-01 at 9 59 00 AM

1. Go to the devdocs and test clicking the various step labels.
2. Make sure horizontal and vertical styling is still correct.
3. Go to the onboarding profiler.
4. Complete a couple steps.
5. Note that you can click on previous step labels to go back, but can't go forward.

### DIFF
--- a/client/dashboard/profile-wizard/index.js
+++ b/client/dashboard/profile-wizard/index.js
@@ -27,44 +27,6 @@ import Theme from './steps/theme';
 import withSelect from 'wc-api/with-select';
 import './style.scss';
 
-const getSteps = () => {
-	const steps = [];
-	steps.push( {
-		key: 'start',
-		container: Start,
-	} );
-	steps.push( {
-		key: 'plugins',
-		container: Plugins,
-	} );
-	steps.push( {
-		key: 'store-details',
-		container: StoreDetails,
-		label: __( 'Store Details', 'woocommerce-admin' ),
-	} );
-	steps.push( {
-		key: 'industry',
-		container: Industry,
-		label: __( 'Industry', 'woocommerce-admin' ),
-	} );
-	steps.push( {
-		key: 'product-types',
-		container: ProductTypes,
-		label: __( 'Product Types', 'woocommerce-admin' ),
-	} );
-	steps.push( {
-		key: 'business-details',
-		container: BusinessDetails,
-		label: __( 'Business Details', 'woocommerce-admin' ),
-	} );
-	steps.push( {
-		key: 'theme',
-		container: Theme,
-		label: __( 'Theme', 'woocommerce-admin' ),
-	} );
-	return steps;
-};
-
 class ProfileWizard extends Component {
 	constructor() {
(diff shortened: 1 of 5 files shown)


rrennick:
> we could instead keep track of the latest step they've finished so they can go back and then back forward to the most recent place they've completed.

That sounds like a great way to handle it. Did you want to do that in this PR or a follow up?

joshuatf:
>As a user I'd expect what I entered to be saved. We could handle this in a follow-up PR though. What do you think?

I agree @justinshreve ; this is something that would not only be helpful for revisiting previous steps, but also in the event the profiler does not get completed as we saw in https://github.com/woocommerce/woocommerce-admin/issues/3067

Created an issue for that [here](https://github.com/woocommerce/woocommerce-admin/issues/3164).

>That sounds like a great way to handle it. Did you want to do that in this PR or a follow up?

Cool - handled in https://github.com/woocommerce/woocommerce-admin/pull/3154/commits/e30cb837db15e271e51545f6c9ee39504bf90bab

This also marks previous steps complete so the `Stepper` in the header will continue showing a checkmark when navigating between items.

client/dashboard/profile-wizard/index.js
```diff
@@ -94,12 +56,61 @@ class ProfileWizard extends Component {
...
+		const steps = [];
+
+		steps.push( {
+			key: 'start',
+			container: Start,
+		} );
+		steps.push( {
+			key: 'plugins',
+			container: Plugins,
+			isComplete: profileItems.hasOwnProperty( 'plugins' ) && null !== profileItems.plugins,
```
justinshreve: Nice. I like having a condition for these now.

joshuatf:
>This happens while the profile data is saving. It feels a bit odd.

Hmm, you're right; it does feel a bit odd @justinshreve.

>Maybe we could use the isPending / purple spinner during save up there?

Currently the `isPending` is only set on the current step so the current would spin and the previous would remain the same.  I think this may also look a little off.

We could refactor the `Stepper` to allow `isPending` on individual steps, which I think would also be good to have, but would prefer to handle that in a follow-up if we decide to.

In https://github.com/woocommerce/woocommerce-admin/pull/3154/commits/ed523daddc27acd078404e9c184462c7861d7d1e I marked all previous steps complete.  Let me know what you think about this option.

PR #3168, opened 2019-11-05
Onboarding: Automatically enable tax calculations for manual configuration

### LINKED ISSUE #3150: Improve the Tax task for non-US stores
As you can see in the video below it's not that easy for a EU merchant to complete the Tax task:
- First, it was necessary to figure it out that it was required to opt-in on the Enable taxes setting
- Then I would need to figure out that it was necessary to go to Tax Settings - Standard rates and then add a tax rate.

I'm wondering if when the merchant clicks `configure` in the Tax task we could automatically opt-in the Enable taxes setting and redirect him to the Standard rates page.

https://www.loom.com/share/e59ee7e6561e4e22a4c8cac09939c125

joshuatf:
Fixes #3150

Enables tax calculations for manual configuration of tax if the user clicks "Configure."  Allows users to directly go to the tax rates page instead of manually enabling and navigating the settings pages.

### Screenshots
Screen Shot 2019-11-05 at 8 32 37 AM

1. Set your store to a TaxJar non-supported country (see - https://developers.taxjar.com/api/reference/#countries )
2. Uncheck "Enable tax rates and calculations" under WooCommerce settings.
3. Go to the tax task.
4. Click "Configure" on the last step.
5. Check that you are redirected to the standard rates page.
6. Go back to the tax task and click on "Configure" again, checking that you are once again redirected to the standard rates page (this time should not update the tax calc option if already enabled and should result in immediate redirect instead of a temporary busy state).

### DIFF
--- a/client/dashboard/task-list/tasks/tax.js
+++ b/client/dashboard/task-list/tasks/tax.js
@@ -256,21 +256,23 @@ class Tax extends Component {
 							{ __( 'Configure', 'woocommerce-admin' ) }
 						</Button>
 						<p>
-							{ interpolateComponents( {
-								mixedString: __(
-									'By clicking "Configure" you\'re enabling tax rates and calculations. More info {{link}}here{{/link}}.',
-									'woocommerce-admin'
-								),
-								components: {
-									link: (
-										<Link
-											href="https://docs.woocommerce.com/document/setting-up-taxes-in-woocommerce/#section-1"
-											target="_blank"
-											type="external"
-										/>
+							{ 'yes' !== generalSettings.woocommerce_calc_taxes &&
+								interpolateComponents( {
+									mixedString: __(
+										'By clicking "Configure" you\'re enabling tax rates and calculations.' +
+											'More info {{link}}here{{/link}}.',
+										'woocommerce-admin'
 									),
-								},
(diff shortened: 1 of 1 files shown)


client/dashboard/task-list/tasks/tax.js
```diff
@@ -227,15 +251,28 @@ class Tax extends Component {
...
-						{ __( 'Configure', 'woocommerce-admin' ) }
-					</Button>
+					<Fragment>
+						<Button isPrimary isBusy={ isPending } onClick={ this.configureTaxRates }>
+							{ __( 'Configure', 'woocommerce-admin' ) }
+						</Button>
+						<p>
+							{ interpolateComponents( {
+								mixedString: __(
+									'By clicking "Configure" you\'re enabling tax rates and calculations. More info {{link}}here{{/link}}.',
```
justinshreve: Should we only show this if `woocommerce_calc_taxes` is no/not set?

joshuatf: Definitely a good idea. Added in https://github.com/woocommerce/woocommerce-admin/pull/3168/commits/12cc0eb58a62cb1552c9bee75460332d24977a00

PR #3170, opened 2019-11-05
Onboarding: Add return to task list after saving taxes

### LINKED ISSUE #3149: Redirect to task list not working after updating Tax settings
**Describe the bug**
After I update the tax settings I'm not automatically redirected to the task

**To Reproduce**

1. Create a store in an european country (Portugal in my case)
2. Install WCS and Jetpack during the Store Profiler
3. Go to the Tax task
4. Click Configure
5. Opt in on enable taxes
6. Go to standard rates, add a row and click save changes

**Expected behavior**
I was expecting to be redirected to the Task list after adding a standard rate

**Screenshots**
https://www.loom.com/share/e59ee7e6561e4e22a4c8cac09939c125

* OS: macOS Mojave 10.14.5
* Browser: Chrome Version 78.0.3904.70

joshuatf:
Fixes #3149

Adds a notice to return to the task list after saving taxes.

Note that this just fires on "Save" click.  There doesn't seem to be a way to hook into the actual save trigger, but open to input here if this doesn't seem like a great solution.

### Screenshots
Screen Shot 2019-11-05 at 6 00 24 PM

1. Go to the task list and use a non-supported TaxJar company to force manual configuration.
2. Click the "Configure" button.
3. Add a tax rate and click "Save."
4. Note that a notice pops up with a link to return to the task list.
5. Note that navigating to the tax settings from the menu (WooCommerce->Settings) does not trigger this notice on save.

### DIFF
--- a/client/dashboard/task-list/tasks/tax.js
+++ b/client/dashboard/task-list/tasks/tax.js
@@ -129,8 +129,6 @@ class Tax extends Component {
 	configureTaxRates() {
 		const { generalSettings, updateSettings } = this.props;
 
-		recordEvent( 'tasklist_tax_config_rates' );
-
 		if ( 'yes' !== generalSettings.woocommerce_calc_taxes ) {
 			this.setState( { isPending: true } );
 			updateSettings( {
@@ -166,7 +164,7 @@ class Tax extends Component {
 			if ( automatedTaxEnabled ) {
 				getHistory().push( getNewPath( {}, '/', {} ) );
 			} else {
-				window.location = getAdminLink( 'admin.php?page=wc-settings&tab=tax&section=standard' );
+				this.configureTaxRates();
 			}
 		} else {
 			createNotice(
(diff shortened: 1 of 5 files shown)


joshuatf:
>We should also add the special active task link to this manual option:

Good catch @justinshreve!  Updated and also enable the tax calc here.

>Is this working for you on 5.2? This is a weird one, since the post notice does work in 5.2 for me, and is also using getAdminLink.

This is odd.  I wonder if there's something funky happening with the build here.  Does this still happen after `npm start`ing again?  I've tried on 5.2, 5.2.1, and 5.2.4 and they all seem to work.  Any other plugins that might be affecting this?  Separate Gutenberg plugin?

justinshreve:
> This is odd. I wonder if there's something funky happening with the build here. Does this still happen after npm starting again? I've tried on 5.2, 5.2.1, and 5.2.4 and they all seem to work. Any other plugins that might be affecting this? Separate Gutenberg plugin?

I'm not really sure what the problem here was.   🙃. I had closed and restarted the build process a couple times yesterday. 

This time I cleaned up my plugins (mostly just installed WooCommerce extensions), ran `npm ci` before `npm start`, and this time things are working on 5.2.4. Gutenberg was never active. Definitely must have been something funky with the package build or something, but I can't seem to replicate it now.

joshuatf:
Thanks @justinshreve!

>Definitely must have been something funky with the package build or something, but I can't seem to replicate it now.

Probably something we'll want to keep an eye on because I've had similar issues crop up from time to time on other PRs as well.

PR #3182, opened 2019-11-06
Allow WP 5.2 to satisfy dependencies

### LINKED ISSUE #3181: Plugin can't be activated with WordPress 5.2
**Describe the bug**
Wc admin states that it can be run with WP 5.2, but does not allow activation with this version of WordPress.

**To Reproduce**

1. Download and install the release for 5.2 - https://wordpress.org/wordpress-5.2.zip
2. Try to activate wc admin with a valid WC version.
3. Note the error.

**Expected behavior**
Expected 5.2 to allow activation.

joshuatf:
Fixes #3181

`version_compare()` in PHP does not recognize `5.2` and `5.2.0` despite some specs saying it should and others saying `5.2` is not a valid version number.  This PR uses `5.2` to allow the WP 5.2 version to work as expected.

It also adds a small refactor to dependencies and respective errors so that we can reduce the number of places the version number is used.

1. Download and install the WP 5.2 release - https://wordpress.org/wordpress-5.2.zip
2. Install WooCommerce >=3.6.
3. Note that activation of wc admin works.
4. Try with later or previous versions of WC/WP and make sure you receive the expected error messages or activation.

### DIFF
--- a/src/FeaturePlugin.php
+++ b/src/FeaturePlugin.php
@@ -221,7 +221,7 @@ class FeaturePlugin {
 
 		if ( ! $woocommerce_minimum_met ) {
 			$errors[] = sprintf(
-				/* translators: URL of WooCommerce plugin */
+				/* translators: 1: URL of WooCommerce plugin, 2: The minimum WooCommerce version number */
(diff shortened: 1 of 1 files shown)


joshuatf:
Thanks for catching those comments, @justinshreve!  Updated in https://github.com/woocommerce/woocommerce-admin/pull/3182/commits/dafd8032d5fa61fda83d21a0492be65083ed443f

Issue #2759, opened 2019-08-07
Split Testing Implementation


pmcpinto:
**New approach:**

Merging the new onboarding to core in a point release presents significant risks, so the new onboarding to the new WC Admin feature plugin and do a point release where new users would receive the suggestion to install/activate the new WC admin right after they activate the WooCommerce plugin and then go through the new onboarding. 

Phase 1: Split test new onboarding experience pre-installed with WC Admin plugin (10% new users)
Phase 2: Split test new onboarding experience pre-installed with WC Admin plugin (50% new users)
Phase 3: Roll out new onboarding experience on WC Core to 100% new users

The mockup for the new flow after the user activates the WooCommerce plugin can be found [here](https://www.figma.com/file/w7YCgNfdJPFFFsFOmiuneZ/Split-testing?node-id=1%3A2).

More info about the new approach on this P2: p6q8Tx-199-p2

-----------------------
Old approach:
Considering that it's not likely that we're able to test the full onboarding experience in one of our hosting partners, it will be necessary to merge the new onboarding into core but only gradually expose it to new users. This progressive approach would allow us to control the sample size, understand if we're moving the main kpis in the right direction and iterate if needed. The idea is to gradually expose the new onboarding in 3 phases:

Phase 1 - randomly assign the new onboarding to 10-20% of new users
Phase 2 - randomly assign the new onboarding to 50% of new users
Phase 3 - assign the new onboarding to 100% of new users

psealock:
There will be some easier ways to split the population. Location or eCommerce plan perhaps.

If you want true A/B test capabilities, as described above, there is some prior art in Calypso: https://github.com/Automattic/wp-calypso/tree/b20b7ea0812694f4041352a02a9a7fe9848a180a/client/lib/abtest

I'm not sure how easily it can be adapted to our needs, but I would estimate this more like a 5.

pmcpinto:
Hey @psealock, I discussed the possibility of splitting the population based on the location, but using a random generator allows to have better control of the sample size. You can check the chat here: p1564136925007200-slack-wc-onboarding

Unfortunately doing the split testing on the eCommerce plan doesn't serve our needs because it won't allow us to test the new Store Profiler, just the Task List.

> I'm not sure how easily it can be adapted to our needs, but I would estimate this more like a 5.

Their estimation is 3 points, but we don't have that value in our point system. I'm going to switch the estimation tag to 5.

pmcpinto:
[...]

PR #3184, opened 2019-11-06
Enable onboarding in the plugin build for those who have opted-in

justinshreve:
Part of woocommerce/woocommerce-admin#2759.

To enable split testing in WooCommerce Core (which can only install .org released versions of the plugin), and to make it easier for internal testers and user testing, this enables the onboarding feature in our plugin build/release. 

There are additional checks so that the relevant code only runs in development, or if a user has opted in via WooCommerce Core (https://github.com/woocommerce/woocommerce/pull/24991) or via a filter.

(The purpose for is filter is so we can release a simple plugin that sets the filter to true for internal testers, and then internal testers just need to use the normal build of WooCommerce Admin).

* Build the development  version of WooCommerce Admin (`npm start` this branch).
* Test a few different onboarding features and make sure everything looks OK.
* Run `npm run build:release` to build the normal plugin release.
* Enable it on a test site and make sure onboarding features do not display.
* Set the `wc_onboarding_opt_in` option to `true` in your database.
* Refresh, and see the onboarding feature active.
* Set the value back to false.
* Refresh, and see no onboarding.
* Set `WOOCOMMERCE_ADMIN_ENABLE_ONBOARDING` to `yes` in your `wp-config.php`, refresh and see the onboarding feature active.

### Changelog Note:

Enhancement: New onboarding experience for opt-in users.

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -89,6 +89,7 @@ function _manually_load_plugin() {
 	define( 'WC_USE_TRANSACTIONS', false );
 	update_option( 'woocommerce_enable_coupons', 'yes' );
 	update_option( 'woocommerce_calc_taxes', 'yes' );
+	define( 'WOOCOMMERCE_ADMIN_SHOW_ONBOARDING', true );
 
 	require_once wc_dir() . '/woocommerce.php';
 	require dirname( __DIR__ ) . '/vendor/autoload.php';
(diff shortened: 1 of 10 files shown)


rrennick:
@justinshreve Just a small nitpick. WC uses `yes` and `no` for settings. For clarity & consistency we should do the same

> Set the `wc_onboarding_opt_in` option to `true` in your database.

tests/bootstrap.php
```diff
@@ -89,6 +89,7 @@ function _manually_load_plugin() {
 	define( 'WC_USE_TRANSACTIONS', false );
 	update_option( 'woocommerce_enable_coupons', 'yes' );
 	update_option( 'woocommerce_calc_taxes', 'yes' );
+	define( 'WOOCOMMERCE_ADMIN_SHOW_ONBOARDING', true );
```
joshuatf: Maybe we could rename this to `WOOCOMMERCE_ADMIN_ONBOARDING_ENABLED` to be slightly more accurate since this won't necessarily force it show at all times.

Issue #2712, opened 2019-07-29
Reorganize class files for PSR-4 autoloading


jeffstieler:
Reorganize class files into the `/src/` directory and add namespaces for PSR-4 autoloading.

Group classes by function/feature/report type.

- [x] Core WC subclasses https://github.com/woocommerce/woocommerce-admin/pull/2728
- [x] Page controller https://github.com/woocommerce/woocommerce-admin/pull/2730
- [x] Admin Notes https://github.com/woocommerce/woocommerce-admin/pull/2731
- [x] Features https://github.com/woocommerce/woocommerce-admin/pull/2736
- [x] Export https://github.com/woocommerce/woocommerce-admin/pull/2734
- [ ] Onboarding
- [x] Sync https://github.com/woocommerce/woocommerce-admin/pull/2739
- [x] Theme Upgrader https://github.com/woocommerce/woocommerce-admin/pull/2740

**API endpoints**

- [ ] Admin Notes
- [ ] Coupons
- [ ] Customers
- [ ] Countries
- [ ] Download IPs
- [ ] Leaderboards
- [ ] Onboarding
- [ ] Orders
- [ ] Categories
- [ ] Reviews
- [ ] Variations
- [ ] Products
- [ ] Settings
- [ ] Taxes
- [ ] Themes

**Reporting classes**

- [ ] Categories
- [ ] Coupons
- [ ] Customers
- [ ] Downloads
- [ ] Exports
- [ ] Imports
- [ ] Orders
- [ ] Performance Indicators
- [ ] Reviews
- [ ] Products
- [ ] Revenue
- [ ] Stock
- [ ] Taxes
- [ ] Variations

**Additional context**

https://github.com/woocommerce/woocommerce-admin/issues/2635

psealock:
@jeffstieler Just double checking here, is this one closed by https://github.com/woocommerce/woocommerce-admin/pull/2717?

jeffstieler:
> is this one closed by #2717?

Yep! Closing.

PR #3186, opened 2019-11-06
move stock stats sync functions to ReportsSync

### LINKED ISSUE #3054: Procedural code missed in PSR-4 reorg
The stock stats data store class file has some procedural code that was missed in the PSR-4 re-org. If the class is not referenced elsewhere, the procedural code isn't loaded.

https://github.com/woocommerce/woocommerce-admin/blob/master/src/API/Reports/Stock/Stats/DataStore.php#L136:L154

This code should be moved to a mainline area where it is always loaded.

rrennick:
Fixes #3054

This PR moves the stock sync functions to the ReportsSync class. These handlers were missed in the PSR-4 code reorg #2712.

- Edit the stock on a product to a value that should set the low stock flag
- Check the stock panel to see that the product is listed
- Edit the stock amount on a product currently shown in the low stock panel that should clear the low stock flag
- Check the stock panel to see that the product is not listed

### Changelog Note:

none needed

### DIFF
--- a/src/API/Reports/Stock/Stats/DataStore.php
+++ b/src/API/Reports/Stock/Stats/DataStore.php
@@ -132,23 +132,3 @@ class DataStore extends ReportsDataStore implements DataStoreInterface {
 		return intval( $query->found_posts );
 	}
 }
-
-/**
- * Clear the count cache when products are added or updated, or when
- * the no/low stock options are changed.
- *
- * @param int $id Post/product ID.
- */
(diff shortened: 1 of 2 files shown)


rrennick:
> Is that expected or should I open a new issue?

Open a new issue as `master` is the same. I also noticed in master that when the update stock button is near the bottom of the window clicking it closes the panel. Maybe we could fix both on the same PR.

Issue #3176, opened 2019-11-06
Merge to Core: Make wc-admin installable to Core


psealock:
wc-admin is to be included in Core releases via composer install in a similar fashion to Blocks and REST API. Example documentation in https://github.com/woocommerce/woocommerce-example-package.

- [x] Publish wc-admin to packagist. [Packagist](https://packagist.org/packages/woocommerce/woocommerce-admin)
- [x] Add a core release process resulting in Github releases that include builds. https://github.com/woocommerce/woocommerce-admin/pull/3190
- [x] Make changes to wc-admin, ie `src/Package.php`. https://github.com/woocommerce/woocommerce-admin/pull/3196
- [x] Make changes to Core. https://github.com/woocommerce/woocommerce/pull/25011

### Notable things
* Ensure bail out mechanism for dependencies is working. ✅ 
* Add tests to Core to ensure package is being installed. ✅ 
* Make sure the correct codebase is working when the plugin is also activated. ✅

psealock:
The only remaining task is the open pull request to add wc-admin as a package in Core. Since we're not targeting the current release, I'm going to add the Blocked label and take this issue out of the sprint.

psealock:
With https://github.com/woocommerce/woocommerce/pull/25011 being merged, the last requirement of this ticket is satisfied. I will close this one out.

PR #3190, opened 2019-11-08
Release process: Add build for Core

psealock:
Partially addresses https://github.com/woocommerce/woocommerce-admin/issues/3176

In order to make wc-admin available for Core inclusion [Packagist](https://packagist.org/packages/woocommerce/woocommerce-admin) exposes the assets created in the source code zip found in releases. See [0.21.0](https://github.com/woocommerce/woocommerce-admin/releases/tag/v0.21.0) for example.

This PR adds an additional Core build and release to `npm run build:release` such that relevant builds with appropriate feature configs can be found in the source code zip.

## Test

Testing this involves creating two releases for plugin and core. One would do this step after `npm run pre-release` and `npm run changelog`, so we can assume version numbers are correct, docs have been updated, and you are on the correct release branch.

1. Make sure you have [hub](https://github.com/github/hub) via `brew install hub`
1. `npm run build:release`
3. `Please enter the version number to tag, for example, 1.0.0:` You can enter something like `0.23.0-test4`
5. Watch the script continue and eventually create the releases on Github with the zips attached
6. Find your [release on Github](https://github.com/woocommerce/woocommerce-admin/releases) and download the zips
8. Compare to make sure all the files and folder are the same in each
9. Check `includes/feature-config.php` in each. The core build has `onboarding` set to false, while the same file in the plugin build has the property set to true.

### Bonus testing

Try creating a custom build:

1. `npm run build:release -- --slug devdocs --features '{"devdocs":true}'`
2. Check the release on Github and inspect the zip's `includes/feature-config.php` file.
3. Devdocs should be turned on.
4. Ensure no Core build has been created.


bin/build-plugin-zip.sh
```diff
@@ -77,10 +77,17 @@ fi
...
-# Run the build.
-status "Generating build... 👷‍♀️"
-npm run build
-npm run docs
+# Build the Core files.
+status "Generating a Core build... 👷‍♀️"
+WC_ADMIN_PHASE=core npm run build
+
+# Make a Github release.
+status "Starting a Github release... 👷‍♀️"
```
justinshreve: So this will create a GitHub release (or at least ask if you want to) every time we build a version of the feature plugin?

I feel like it might be better to have the core build step/GitHub release run as separate commands from the plugin build.

We could add an additional parameter to `npm run build:release` to pass if it should be core or plugin, and only run the GitHub stuff if core is passed.

We also can create additional zips with different features enabled: https://github.com/woocommerce/woocommerce-admin/blob/master/docs/feature-flags.md#building-custom-plugin-builds, which doesn't need to run the core/GitHub commands -- but would with this change.

bin/build-plugin-zip.sh
```diff
@@ -77,10 +77,17 @@ fi
 status "Gathering PHP dependencies... 🐿️"
 composer install --no-dev
 
-# Run the build.
-status "Generating build... 👷‍♀️"
-npm run build
-npm run docs
```
justinshreve: Should docs still be built as a part of this? This doesn't get added back below.

bin/build-plugin-zip.sh
psealock: >We could add an additional parameter to npm run build:release to pass if it should be core or plugin, and only run the GitHub stuff if core is passed.

We could definitely do this. I figured making two releases each time is Ok because an update in core would be done by updating Core's `composer.json`. However, you've raised a great point, which is that we have different zips now (core, plugin, features like OBW) and maybe each release should be more explicit, like the one @joshuatf did https://github.com/woocommerce/woocommerce-admin/releases/tag/v0.21.0-onboarding.

How about if I incorporate the slug/feature commands such that a GH release is made automatically specific to that slug? 

For instance:

* `npm run build:release` creates the plugin configs for GH release `vx.x.x` and `woocomerce-admin.zip`.
* `npm run build:release -- --slug core` creates the core configs for GH release `vx.x.x-core` and `woocomerce-admin-core.zip`.
* `npm run build:release -- --slug onboarding --features '{"onboarding":true}'` creates the modified configs for GH release `vx.x.x-onboarding` and `woocomerce-admin-onboarding.zip`.

bin/build-plugin-zip.sh
psealock: It was redundant because its part of the `npm run pre-release` script done earlier in the release process.

bin/build-plugin-zip.sh
psealock: I decided to do something else:

If a slug or feature set is designated, then I just skip the core release. However, I want to automate the creation of the release a la v0.21.0-onboarding.

Would that work?

justinshreve: > If a slug or feature set is designated, then I just skip the core release. However, I want to automate the creation of the release a la v0.21.0-onboarding.

That sounds good to me. 👍I think it'll be useful to be able to do that without worrying about whats generated for core.

psealock:
@justinshreve could you give this another look? I updated it such that creating a custom build avoids creating a Core build and Core release.

```
npm run build:release -- --slug devdocs --features '{"devdocs":true}'
```

The above command will modify the plugin configs and produce a Github release `v0.23-devdocs` or what ever name you choose and attach the resulting zip file. It will not create a Core release.

config/core.json
```diff
@@ -1,11 +1,11 @@
 {
 	"features": {
-		"activity-panels": false,
-		"analytics": false,
-		"analytics-dashboard": false,
-		"analytics-dashboard/customizable": false,
+		"activity-panels": true,
```
timmyc: all the 🚀

mikejolley:
@timmyc I think this looks ok, you're pushing the built files to git which is correct for the packages. That would be the only thing to confirm is working once pushed to github. If you download the zip for a created release and install it, no build steps should be required. If that works, you're good.

Issue #2237, opened 2019-05-15
Merge: Handling WordPress < 5.2


timmyc:
While WooCommerce Admin already has logic in place to not run if the dependencies are not satisfied ( WC 3.6 or greater and WP 5.0 or greater ) we need to alter this logic a bit to prepare for merging the code to core. This issue will act as a parent issue for all tasks related to this update.

- [ ] Create conditional function, perhaps which will eventually live in `includes/wc-conditional-functions.php` to determine `wc_admin_enabled()`. The only logic would be checking for WP Version >= 5.0, and PHP must be > 5.4 but I think we should also make this filterable. 
- [ ] if `wc_admin_enabled()` is falsey, Nav/Activity Panels and new Analytics should not be loaded.
- [ ] If a site is running less than 5.0, we should also display an admin notice informing them of the new features they are missing out on: blocks, activity panels, analytics.

rrennick:
Related #2203

timmyc:
Updated to include the need for the PHP > 5.4 check - @peterfabian should that be PHP >= 5.4 or greater than 5.4?

peterfabian:
Traits are supported from PHP 5.4.0 & I'm not aware of any other features that rely on newer PHP versions, so PHP >= 5.4.0 as far as I know.

psealock:
Things have changed since this issue was originally written. Specifically, when we merge in 3.9 here are the supported versions:

WooCommerce: WP 5.3, 5.2, 5.1

wc-admin: WP 5.3, 5.2

We'll need a mechanism to disable wc-admin for sites running WP 5.1

rrennick:
> We'll need a mechanism to disable wc-admin for sites running WP 5.1

Product blocks does that with https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/woocommerce-gutenberg-products-block.php#L19:L21

With that implementation the rest of the changes for this issue would be implemented in WC core.

psealock:
In Core, packages aren't loaded when PHP is less than 5.6. So we're good there:

https://github.com/woocommerce/woocommerce/blob/cd2a63c2936af7d36531c6cc7610eb56f36e2873/woocommerce.php#L30-L38

I'll add a check for WP < 5.2 in `src/Package.php` of this repo because that code is run by Core before initialising the package.

PR #3196, opened 2019-11-11
Allow core inclusion via Package.php

psealock:
Partially addresses https://github.com/woocommerce/woocommerce-admin/issues/3176
Depends on https://github.com/woocommerce/woocommerce-admin/pull/3190
Fixes https://github.com/woocommerce/woocommerce-admin/issues/2237

Create a `Package.php` file so that Core knows how to initialise wc-admin. Includes logic to not run initialise in WP less than 5.2.

## Test Inclusion

1. Checkout this branch of Core `https://github.com/woocommerce/woocommerce/tree/try/add-wc-admin`. The branch's `composer.json` points to the build of this PR also found in [0.22.0-dev7](https://github.com/woocommerce/woocommerce-admin/releases/tag/v0.22.0-dev7).
2. Deactivate wc-admin.
3. `composer update`.
4. Voila! See wc-admin alive and well in core.
5. Activate wc-admin plugin and check for any conflicts.

## Test version update script

1. Update the version in `package.json`
2. run `php ./bin/update-version.php`
3. See the updated version in `src/Package.php`

## Test version and initialization

Core has no plugin activation, so test that database is initialized

1. You can install the WooCommerce branch on a fresh install, or...
2. Delete the `wc_admin_version` from the options table
3. Refresh and see the option re-added.

## Bonus Test WP 5.1

1. Downgrade to WP 5.1 and see wc-admin not load at all.

### DIFF
--- a/src/Package.php
+++ b/src/Package.php
@@ -1,0 +1,54 @@
+<?php
+/**
+ * Returns information about the package and handles init.
+ *
+ * @package Automattic/WooCommerce/WCAdmin
+ */
+
+namespace Automattic\WooCommerce\Admin;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Main package class.
+ */
+class Package {
+
+	/**
+	 * Version.
+	 *
+	 * @var string
+	 */
+	const VERSION = '0.22.0';
+
+	/**
+	 * Init the package.
+	 *
+	 * Only initialize for WP 5.2 or greater.
+	 */
+	public static function init() {
+		$wordpress_minimum_met = version_compare( get_bloginfo( 'version' ), '5.2', '>=' );
+		if ( ! $wordpress_minimum_met ) {
+			return;
+		}
+		FeaturePlugin::instance()->init();
+	}
+
+	/**
+	 * Return the version of the package.
+	 *
+	 * @return string
+	 */
+	public static function get_version() {
+		return self::VERSION;
+	}
+
+	/**
+	 * Return the path to the package.
+	 *
+	 * @return string
+	 */
+	public static function get_path() {
+		return dirname( __DIR__ );
+	}
+}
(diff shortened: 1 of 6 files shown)


src/Package.php
```diff
@@ -0,0 +1,59 @@
...
+ * Returns information about the package and handles init.
+ *
+ * @package Automattic/WooCommerce/WCAdmin
+ */
+
+namespace Automattic\WooCommerce\Admin;
+
+defined( 'ABSPATH' ) || exit;
+
+require_once WC_ABSPATH . 'packages/woocommerce-admin/includes/core-functions.php';
```
mikejolley: I see you're using `require_once` here which is good, but I see you had to add `function_exists` to all of the functions included in this files. Why was that?

Another alternative would be to move these requires statements to your `FeaturePlugin` class (which is autoloaded once) to include during `init`. Maybe that will remove the need for function exists?

psealock: Why was that?

There were conflicts, but your suggestion worked, thanks!

psealock:
Thanks for taking a look @mikejolley. I took your suggestion and removed the `files` property from `composer.json` because those files are now explicitly required.

rrennick:
@psealock Getting the following error activating WC Admin with WP CLI

```Fatal error: Cannot redeclare wc_admin_number_format() (previously declared in /wp-content/plugins/woocommerce/packages/woocommerce-admin/includes/core-functions.php:18)```

**Comment by psealock** [comment]
>Getting the following error activating WC Admin with WP CLI

Thanks @rrennick. I just tried the cli and could not reproduce. Could you check the following?

* From your Core WooCommerce folder, `composer show`. You should see the following for wc-admin. I just want to confirm you have the latest. I found `composer clearcache` works... sometimes.
```
woocommerce/woocommerce-admin                  dev-build/0.22.0-dev8 a2e6857

```
* From your wc-admin folder, re run `composer install` from the latest commit on the branch of this PR. This is to confirm the autoloader has been written with this branch.

PR #3202, opened 2019-11-11
POC store navigation

rrennick:
This PR implements a proof of concept for separating site and store navigation. It includes a number of filters to allow easy additions of WC extensions to the store navigation.

To demonstrate one of the filters:
- switch to this branch
- `composer install`
- Install and activate WooCommerce Subscriptions
- Go to WooCommerce -> Subscriptions
- You should see the full site menu where WooCommerce is the only top level WC menu visible
- Add the following snippet and refresh the page
```
add_filter( 'woocommerce_navigation_post_types', function( $post_types ) {
	$post_types[] = 'shop_subscription';
	return $post_types;
} );
```
- You should see the store menu where the Dashboard is the only top level WP menu visible


timmyc:
Really nice approach here @rrennick - I like how we could take this, and ship a POC in a six week cycle potentially. We would like still need to do some hackish things along the way to hide away various wp-admin elements, but it seems manageable.

pmcpinto:
Following the idea of removing Calypsoify (#3880), I think that we should focus this exploration within the scope of the eCommerce plan. 

You can see [here](https://cloudup.com/c5bunpEYef2) a rough draft of how the navigation would look like in the eCommerce plan after removing Calypsoify

joshuatf:
This PR has been rebased and modified a bit from the original.  Per Slack conversation, we're going to manually opt-in items to the new menu.  This PR:
* Creates the API for adding new "category" and "submenu" items.
* Creates a base (unstyled navigation).
* Adds in core menu items and settings.
* Creating a feature flag and option to hide the wc nav until completion.

This PR does not include follow-up issues:
* Navigation styling or interaction.
* Adding menu items from non-core areas (e.g., wc-admin, marketing, and 3rd party plugin menu items to come in a follow-up).
* Post type configuration with the wc nav screen.
* Menu ordering.
* Finalizing the items on the original wp admin menu.

### Screenshots

Screen Shot 2020-04-15 at 3 51 55 PM

### Testing

1. Add the option `woocommerce_navigation_enabled` and set to `1`.
1. Make sure that visiting a core WC page (not post types yet) shows the new wc navigation menu and hides the old.
1. Make sure the links work as expected.
1. Make sure only items your user account is capable of viewing are shown on the new menu.
1. Make sure that menu items that have been migrated are not shown in the wp-admin menu, with the exception of `Plugins` and `User profile`.
1. Optionally attempt to add a category (top-level item) or submenu item in a 3rd party plugin.
`\Automattic\WooCommerce\Admin\Features\Navigation::add_menu_category()`

```
	/**
	 * Adds a top level menu item to the navigation.
	 *
	 * @param string $title Menu title.
	 * @param string $capability WordPress capability.
	 * @param string $slug Menu slug.
	 * @param string $url URL or menu callback.
	 * @param string $icon Menu icon.
	 * @param int    $order Menu order.
	 * @param bool   $migrate Migrate the menu option and hide the old one.
	 */
	public static function add_menu_category( $title, $capability, $slug, $url = null, $icon = null, $order = null, $migrate = true )

	/**
	 * Adds a child menu item to the navigation.
	 *
	 * @param string $parent_slug Parent item slug.
	 * @param string $title Menu title.
	 * @param string $capability WordPress capability.
	 * @param string $slug Menu slug.
	 * @param string $url URL or menu callback.
	 * @param string $icon Menu icon.
	 * @param int    $order Menu order.
	 * @param bool   $migrate Migrate the menu option and hide the old one.
	 */
	public static function add_menu_item( $parent_slug, $title, $capability, $slug, $url = null, $icon = null, $order = null, $migrate = true )
```

rrennick:
@joshuatf Great to see the progress on this. I noticed a couple things

- HTML encoded entities in the menu items
Screen Shot 2020-04-23 at 2 16 26 PM
- The WC submenu is always filtered so the only way to navigate to the new menu is by manually entering an address in the address bar
Screen Shot 2020-04-23 at 2 17 21 PM

Since I wrote some of the PR I'll leave this for someone else to review. The PR is behind a feature flag so I would be fine with merging it.

client/index.js
```diff
@@ -9,12 +9,22 @@ import { render } from '@wordpress/element';
...
+import Navigation from './navigation';
 import 'wc-api/wp-data-store';
 import { withSettingsHydration } from '@woocommerce/data';
 
 const appRoot = document.getElementById( 'root' );
+const navigationRoot = document.getElementById( 'woocommerce-embedded-navigation' );
 const settingsGroup = 'wc_admin';
 
+if ( navigationRoot ) {
+	render( <Navigation />, navigationRoot );
```
psealock, line 21: Lets use the new Settings store so we don't have to use `import { getSetting } from '@woocommerce/wc-admin-settings';`. See below `withSettingsHydration` so that the navigation component can always have the most up to date settings. You can retrieve the settings with `getSettingsForGroup` selector.

If something dynamically alters the `wcNavigation`, we can receive updates without a page refresh.

client/navigation/index.js
```diff
@@ -0,0 +1,54 @@
...
+						const { slug, title, url } = item;
+						return (
+							<li
+								key={ slug }
+								className="woocommerce-navigation__menu-item"
+							>
+								<a href={ url }>{ title }</a>
+								{ item.children.length && (
+									<ul className="woocommerce-navigation__submenu">
+										{ item.children.map( ( subitem ) => {
```
psealock, line 33: nitpick: can we add `const { slug, title, url } = subitem;`?

src/Loader.php
```diff
@@ -341,6 +342,14 @@ public static function register_scripts() {
 
 		wp_set_script_translations( 'wc-date', 'woocommerce-admin' );
 
+		wp_register_script(
```
psealock, line 345: lets wrap this in a feature check using `wc_admin_get_feature_config` so the nav isn't loaded when unreleased.

src/FeaturePlugin.php
```diff
@@ -16,6 +16,7 @@
 use \Automattic\WooCommerce\Admin\Notes\WC_Admin_Notes_Woo_Subscriptions_Notes;
 use \Automattic\WooCommerce\Admin\Notes\WC_Admin_Notes_Tracking_Opt_In;
 use \Automattic\WooCommerce\Admin\Notes\WC_Admin_Notes_WooCommerce_Payments;
+use \Automattic\WooCommerce\Admin\Features\Navigation;
```
psealock, line 19: We could wrap this in a feature flag too so unnecessary calculations aren't made.

src/Features/Navigation.php
```diff
@@ -0,0 +1,459 @@
...
+					return $key;
+				}
+			}
+		}
+
+		return null;
+	}
+
+	/**
+	 * Hides all WP admin menus items and adds screen IDs to check for new items.
```
psealock, line 367: This is a nice solution to let extension devs hook into this optionally 🎉

psealock:
Just noting here that the menu/submenu hierarchy will be 3 levels deep, and perhaps in undetermined depth. See pbIJXs-67-p2#comment-185. Obviously can be done in a follow up, but just wanted to note that here.

client/index.js
joshuatf: Awesome- updated in https://github.com/woocommerce/woocommerce-admin/pull/3202/commits/c896aab78b371cc82459845c23c2892ad58de372

src/Loader.php
joshuatf: This actually appeared to be registered twice and is for the exported nav package, not the new navigation feature.

Removed the extra registration, but can see this might be a point of confusion going forwards with those 2 items so similarly named.

joshuatf:
>Just noting here that the menu/submenu hierarchy will be 3 levels deep, and perhaps undetermined in depth. 

Good point; updated https://github.com/woocommerce/woocommerce-admin/pull/3202/commits/c896aab78b371cc82459845c23c2892ad58de372 to recursively include menu items so we're not limited by depth.

>To add to the follow up tasks, getting the "WooCommerce" link in the regular wp-admin nav bar to fire off the new sidebar like the following would be great.

Follow-up task for that [here](https://github.com/woocommerce/woocommerce-admin/issues/4204) and additional follow-up tasks created under the label "Navigation."

joshuatf:
>HTML encoded entities in the menu items

Good eyes, Ron.  Added that as part of the follow-up styling issue - https://github.com/woocommerce/woocommerce-admin/issues/4202

PR #3203, opened 2019-11-11
Try/slot fill activity panel

jeffstieler:
This PR is meant to be an exploration of the SlotFill components in Gutenberg and how they can be applied to WC-Admin for extensibility.

The ultimate goal of this PR is to:
* Refactor the Activity Panel to use SlotFill
* Create exported Fills components for 3PD consumption
* Create an example extension using the exported Fill

Everything was working well up to f8e0cca. Once I moved the SlotFill components to the `packages/components` dir, they stopped working. Slack conversations lead me to believe there needs to be some sort of manually set up context to glue them back together, but I have had to move on to other work after getting stuck on this.

Slack threads for additional context:
* p1572481117181500-slack-wc-admin
* p1572999268237100-slack-wc-admin


psealock:
Is this still blocked? I'm working on a currency example extension where I'd like to be able to add a UI element to reports so they render an option for users to choose a currency. Currently, this is only possible via Advanced Filters, which make for terrible user experience.

Screen Shot 2019-11-28 at 2 32 03 PM
Screen Shot 2019-11-28 at 2 32 35 PM
Screen Shot 2019-11-28 at 2 32 48 PM

This seems like the perfect use case for the Slot/Fill approach. Are there still issues with Notices? Would it be straightforward to add Slot/Fill in a basic fashion to Reports?

packages/components/src/index.js
```diff
@@ -59,3 +60,4 @@ export { default as TextControlWithAffixes } from './text-control-with-affixes';
 export { default as useFilters } from './higher-order/use-filters';
 export { default as ViewMoreList } from './view-more-list';
 export { default as WebPreview } from './web-preview';
+export { SlotFillProvider } from '@wordpress/components';
```
psealock: Why does this need to be exported? Is it minimum version requirement for WP thing?

psealock: Oh, I see now. So we can guarantee its available for extensions.

jeffstieler:
@psealock this PR doesn't need to land to work SlotFill into the reports - we'd need to build new SlotFill components for that portion of the app anyhow. I could put together a try branch for that, but it is definitely more complicated than the activity panel and could take a while and require a lot of code changes.

PR #3206, opened 2019-11-11
Onboarding - Fix homepage notice on Gutenberg save

### LINKED ISSUE #3155: Onboarding: Gutenberg event not fired on save
**Describe the bug**
For the onboarding task script used for the homepage, I'm getting the following error when loading the editor:

`wp.domReady is not a function`

**To Reproduce**

1. WP 5.2.4
2. Delete your homepage if necessary.
3. Go to the "Customize Appearance" task.
4. Create a homepage.
5.  Note the console errors and return to task list notice not showing.

**Expected behavior**
Expected no errors and the task list return notice to show.

**Screenshots**
Screen Shot 2019-11-01 at 1 01 11 PM

justinshreve:
Fixes #3155.

I wasn't able to reproduce all the 5.2.4 errors myself, but did have trouble getting the notice to fire correctly in 5.3.

This PR rearchitects the homepage notice logic, based on Josh's PR: #3192. I think my original path of subscribing to the changes and looking for the right combinations of events would be more useful if we weren't already loading the JS on a very specific path. Hooking onto the click targets also allows me to more consistently display the notice at the right time across versions.

Note that in 5.3, the notice is a `Snackbar` notification. 5.2 retains the green notice. This is to stay consistent with other post notices in those respective versions.

##### 5.2

Screen Shot 2019-11-11 at 12 31 15 PM

##### 5.3

Screen Shot 2019-11-11 at 12 29 36 PM

* Test using 5.2.4/5.2.5-alpha
* Delete your homepage and enable the task list
* Go to the appearance task and create a homepage
* Hit the publish button twice, and you should see a success notice.
* Update to 5.3.
* Delete your homepage
* Go to the appearance task and create a homepage
* Hit the publish button twice, and you should see a success notice.

### DIFF
--- a/client/wp-admin-scripts/onboarding-homepage-notice/index.js
+++ b/client/wp-admin-scripts/onboarding-homepage-notice/index.js
@@ -2,9 +2,10 @@
 /**
  * External dependencies
  */
-import { select, subscribe, dispatch } from '@wordpress/data';
+import { dispatch } from '@wordpress/data';
 import { __ } from '@wordpress/i18n';
 import apiFetch from '@wordpress/api-fetch';
+import domReady from '@wordpress/dom-ready';
 
 /**
  * WooCommerce dependencies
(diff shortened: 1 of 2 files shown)


client/wp-admin-scripts/onboarding-homepage-notice/index.js
```diff
@@ -2,73 +2,103 @@
 /**
  * External dependencies
  */
-import { select, subscribe, dispatch } from '@wordpress/data';
+import { dispatch } from '@wordpress/data';
 import { __ } from '@wordpress/i18n';
 import apiFetch from '@wordpress/api-fetch';
+import domReady from '@wordpress/dom-ready';
```
joshuatf: Nice 👍 Looks like this solved any issues I was having.

client/wp-admin-scripts/onboarding-homepage-notice/index.js
```diff
@@ -2,73 +2,103 @@
...
-		const isPublishingPost = select( 'core/editor' ).isPublishingPost();
-		const isStatus = select( 'core/editor' ).getCurrentPost().status;
+/**
+ * Returns a promise and resolves when the post has been saved and notices have shown.
+ *
+ * @return {Promise} Promise for overlay existence.
+ */
+const saveCompleted = () => {
+	if (
+		null === document.querySelector( '.components-notice__content' ) &&
```
joshuatf: One concern I have is another plugin creating a notice that could give a false positive here.  Maybe we could update the selector to `.post-publish-panel__postpublish` which I think is only visible after publishing.

PR #3213, opened 2019-11-12
Onboarding: Add email marketing note if profiler is complete

### LINKED ISSUE #3079: Inbox notification suggesting to sign up to email marketing
In the current onboarding, we have a step where the user can sign up for WC email marketing, which is something that we don't have in the new one.

Displaying an inbox notification about signing up for email marketing seems to be the best option.

@justinshreve I want to check one thing with you before I move forward with creating the notification content, etc:
It's important that the user confirms which email is he going to use to receive our newsletters, so I think there are two options:
- Displaying the email input field in the notification, so the user can add/change the email address and then when he clicks in the button he's just redirected to the Mailchimp page with the captcha
- Not displaying the email input field in the notification and when the user clicks in the button he's redirected to the Mailchimp page where he can add/change the email address (AFAIR they have this option)

Which one do you think it's easier?

joshuatf:
Fixes #3079

Adds the email marketing signup as a note after the profiler is complete if wc-admin has been installed for at least 2 days.

**Questions**
* The copy is mostly taken from the old signup form.  Any changes we should make here? /cc @pmcpinto @jameskoster
* The message is not set to show until wc-admin has been installed for at least 2 days and the profiler has been completed.  I think 2 days is a good buffer to avoid information overload, but open to input here.
* This Mailchimp form question is a little weird given they've already installed WC.  Wondering if there's a different form or some param we should be passing here.

Screen Shot 2019-11-12 at 7 42 20 PM

### Screenshots
Screen Shot 2019-11-12 at 7 41 28 PM

1. Have a site that's had wc-admin installed for at least 2 days.
2. Complete the onboarding profiler.
3. Run the `wc_admin_daily` cron event.
4. Note the note and make sure the link "Yes please!" points to the correct signup form.

### DIFF
--- a/src/Events.php
+++ b/src/Events.php
@@ -16,6 +16,7 @@ use \Automattic\WooCommerce\Admin\Notes\WC_Admin_Notes_Giving_Feedback_Notes;
 use \Automattic\WooCommerce\Admin\Notes\WC_Admin_Notes_Mobile_App;
 use \Automattic\WooCommerce\Admin\Notes\WC_Admin_Notes_New_Sales_Record;
 use \Automattic\WooCommerce\Admin\Notes\WC_Admin_Notes_Tracking_Opt_In;
+use \Automattic\WooCommerce\Admin\Notes\WC_Admin_Notes_Onboarding_Email_Marketing;
 
 /**
  * WC_Admin_Events Class.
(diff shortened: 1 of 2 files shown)


timmyc:
Looking good to me 👍

joshuatf:
>Agreed, I wonder if we (or another team/marketing) could make a new form for this context? (Though I think this could be merged and updated later).

Sounds like a good plan.  I'll merge as is.  @justinshreve  do you think we should ping another team about this or allow this to surface and get feedback after another round of testing?

pmcpinto:
Hey @justinshreve @joshuatf let me ping Team Pi to see if they can change the form and to confirm if the 2 days buffer to display the notification affects their email marketing lifecycle.

PR #3215, opened 2019-11-12
Select control: fix misalignment and hide scroll bar

Aljullu:
### Screenshots
_Before:_
Peek 2019-11-12 14-40

_After:_
Peek 2019-11-12 14-38

- Open the select control and verify there is no scroll bar (this issue might not be reproducible with Mac OS because it hides scrollbars by default).
- Open the select control and with the keyboard (<kdb>Alt</kbd>+<kdb>Tab</kbd>) focus the previous element. Verify the text is not misaligned.

### DIFF
--- a/packages/components/src/search/style.scss
+++ b/packages/components/src/search/style.scss
@@ -44,6 +44,7 @@
 		font-size: 13px;
 		color: #72777c;
 		position: static;
+		transform: none;
 		margin: 0;
 		width: 100%;
 	}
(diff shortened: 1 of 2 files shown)


rrennick:
@Aljullu I'm still seeing a difference in font size/position in Chrome between Single Product and the other 2 filters

Screen Shot 2019-11-12 at 4 00 01 PM
Screen Shot 2019-11-12 at 3 58 06 PM

Aljullu:
@rrennick I think that's related to #3214, the _Single Product_ filter should show the product name on top but that was not working in `master` when I created this PR.

I rebased this branch with the latest `master` and it should look good now:
image

PR #3220, opened 2019-11-12
Onboarding - Fix busted tracks events and add missing event

### LINKED ISSUE #3208: Task List - Add 2 missing events
It will be necessary to add two events that are missing in the Tax and Customize Appearance tasks:

- `wcadmin_tasklist_tax_setup_automated_proceed` (line 42)
- `wcadmin_tasklist_appearance_continue_setup` (line 49)

The detailed info is available in the sheet.

justinshreve:
Fixes #3208.

This PR:

* Adds a `wcadmin_tasklist_appearance_continue_setup` event that is fired when you click on the continue setup homepage notice
* Renames `wcadmin_tasklist_tax_setup_automated_simple` to `wcadmin_tasklist_tax_setup_automated_proceed` per p1573575958016300-slack-wc-onboarding
* Fixes a condition where plugins would finish installing on the Tax task, but onComplete and the `wcadmin_tasklist_tax_install_extensions` event wouldn't fire because the task is immediately unmounted when pluginsToActive becomes 0.
* Fixes `wcadmin_storeprofiler_store_theme_upload` and the theme upload onComplete handler by moving the necessary theme data hook to the correct location.

* Enter `localStorage.setItem( 'debug', 'wc-admin:*' );` into your developer console.

* Enable the task list
* Delete your homepage
* Go to the appearance task and create a new homepage
* Publish, and click the “continue setup” button
* See the `tasklist_appearance_continue_setup` event fire

* Make sure that you have WooCommerce Services and Jetpack enabled and your country set to US (so you get the “good news” message)
* Click confirm or skip and verify `wcadmin_tasklist_tax_setup_automated_proceed` is fired.

* Deactivate WooCommerce Services or Jetpack
* Go to the tax task, and install the offered extensions
* Verify that the `wcadmin_tasklist_tax_install_extensions` event fires

* Enable the store profile wizard
* Go to the theme step and upload a theme
* Verify that the `wcadmin_storeprofiler_store_theme_upload` event fires, and you see your uploaded theme image / name / description.

### DIFF
--- a/client/dashboard/task-list/tasks/appearance.js
+++ b/client/dashboard/task-list/tasks/appearance.js
@@ -14,7 +14,7 @@ import { withDispatch } from '@wordpress/data';
  * WooCommerce dependencies
  */
 import { Card, Stepper, TextControl } from '@woocommerce/components';
-import { getHistory, getNewPath } from '@woocommerce/navigation';
+import { getHistory, getNewPath, getQuery } from '@woocommerce/navigation';
 import { getSetting, setSetting } from '@woocommerce/wc-admin-settings';
 
 /**
@@ -46,6 +46,13 @@ class Appearance extends Component {
 		this.importProducts = this.importProducts.bind( this );
 		this.updateLogo = this.updateLogo.bind( this );
 		this.updateNotice = this.updateNotice.bind( this );
+	}
+
+	componentDidMount() {
+		const { from } = getQuery();
+		if ( 'homepage' === from ) {
+			recordEvent( 'tasklist_appearance_continue_setup', {} );
+		}
 	}
 
 	async componentDidUpdate( prevProps ) {
--- a/client/dashboard/task-list/tasks/appearance.js
+++ b/client/dashboard/task-list/tasks/appearance.js
@@ -14,7 +14,7 @@ import { withDispatch } from '@wordpress/data';
  * WooCommerce dependencies
  */
 import { Card, Stepper, TextControl } from '@woocommerce/components';
-import { getHistory, getNewPath, getQuery } from '@woocommerce/navigation';
+import { getHistory, getNewPath } from '@woocommerce/navigation';
 import { getSetting, setSetting } from '@woocommerce/wc-admin-settings';
 
 /**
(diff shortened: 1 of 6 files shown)


rrennick:
@justinshreve I didn't see (WCS & JP enabled)

> (so you get the “good news” message)

But the `tasklist_appearance_continue_setup` event fired

I'm running a build of the plugin on a jurassic.ninja site.

client/wp-admin-scripts/onboarding-homepage-notice/index.js
```diff
@@ -77,8 +82,11 @@ const onboardingHomepageNotice = () => {
 				type: notificationType,
 				actions: [
 					{
-						url: getAdminLink( 'admin.php?page=wc-admin&task=appearance' ),
 						label: __( 'Continue setup.', 'woocommerce-admin' ),
+						onClick: () => {
+							recordEvent( 'tasklist_appearance_continue_setup', {} );
```
joshuatf: Probably not an issue, but just curious if `recordEvent` is async.  Is there any chance this request has not completed before we change pages?

Looks like it might be using `ga` under the hood and I think this is okay to fire events on page change, but just want to verify.

justinshreve: The closest answer I could find for this was in the Calypso docs https://github.com/Automattic/wp-calypso/blob/a5c00596e329b79f185803328c3af8caade233bb/client/lib/analytics/README.md#what-is-the-analytics-queue which I think use the same underlying `stats.wp.com/w.js` script -- and it looks like we could run into issues here.

Since we don't currently have an analytics queue (which might be something to consider for WooCommerce Admin), I ended up moving this check to the next pageload of appearance -- do you think this works?

https://github.com/woocommerce/woocommerce-admin/pull/3220/commits/51bbbf95868ec1004b2137ad443e968da73fa212

client/dashboard/task-list/tasks/appearance.js
```diff
@@ -48,6 +48,13 @@ class Appearance extends Component {
 		this.updateNotice = this.updateNotice.bind( this );
 	}
 
+	componentDidMount() {
+		const { from } = getQuery();
+		if ( 'homepage' === from ) {
```
joshuatf: I think we should also remove this query from the URL after recording the event, just to avoid accidentally re-triggering this event.

justinshreve: I ended up implementing a tracks queue in a separate PR that removes this:
https://github.com/woocommerce/woocommerce-admin/pull/3250

PR #3232, opened 2019-11-13
Onboarding: Remove notice to run the old setup wizard

### LINKED ISSUE #3228: Onboarding: Old setup wizard notice is shown
**Describe the bug**
When visiting the WP dashboard, the old setup wizard notice is shown even if the new onboarding experience is enabled.

**To Reproduce**

1. Create a new site.
2. Enable the new onboarding experience.
3. Go to the dashboard.
4. Note the notice.

**Expected behavior**
Expected to not see this notice or see one that points to the new onboarding profiler.

**Screenshots**
Screen Shot 2019-11-13 at 1 18 11 PM

joshuatf:
Fixes #3228

Removes the old setup wizard notice if the new onboarding is active.

I don’t think we’ll want to modify content in this notice since most of our notifications are moving towards using the inbox, so this PR simply removes the old notice.

### Screenshots (before)
Screen Shot 2019-11-13 at 5 19 54 PM

1. Install WC in a new WP install and don’t go through any setup steps.
2. Make sure onboarding is enabled.
3. Navigate to any core pages (e.g., Dashboard, Plugins, etc).
4. Make sure the old notice is no longer shown.

### DIFF
--- a/src/Features/Onboarding.php
+++ b/src/Features/Onboarding.php
@@ -81,6 +81,7 @@ class Onboarding {
 		add_action( 'current_screen', array( $this, 'calypso_tests' ) );
 		add_filter( 'woocommerce_admin_is_loading', array( $this, 'is_loading' ) );
 		add_filter( 'woocommerce_rest_prepare_themes', array( $this, 'add_uploaded_theme_data' ) );
+		add_filter( 'woocommerce_show_admin_notice', array( $this, 'remove_install_notice' ), 10, 2 );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


justinshreve:
This is looking good to me as well.

> I don’t think we’ll want to modify content in this notice since most of our notifications are moving towards using the inbox, so this PR simply removes the old notice.

Not a blocker for this PR, but I wonder if we should consider showing something using WooCommerce Admin's StoreAlert notice on the other WooCommerce pages (Orders, etc)? Maybe something we could do in phase 2 / before launching to 100% of users? Or do we think that onboarding taking over the dashboard section is enough? cc @pmcpinto

pmcpinto:
> Not a blocker for this PR, but I wonder if we should consider showing something using WooCommerce Admin's StoreAlert notice on the other WooCommerce pages (Orders, etc)? Maybe something we could do in phase 2 / before launching to 100% of users? Or do we think that onboarding taking over the dashboard section is enough? cc @pmcpinto

Good suggestion. I think it will depend on how many users are going to complete the Task List during phase 1. If the % is low inserting that notification could be useful to raise awareness.

joshuatf:
Good point @justinshreve - opened an issue for that [here](https://github.com/woocommerce/woocommerce-admin/issues/3242).

> I think it will depend on how many users are going to complete the Task List during phase 1. If the % is low inserting that notification could be useful to raise awareness.

Agreed.  Let's take a look after phase 1 and make a judgement call based on the data.

PR #3245, opened 2019-11-14
Onboarding: Update check for TOS accepted in tax step

### LINKED ISSUE #3151: Tax task - automated tax message displayed after setting up taxes manually
**Describe the bug**
After manually completing the Tax step if I go back to that step it says that I'm able to automatically calculate taxes, while the first time I started the task it says that I need to calculate them manually

**To Reproduce**

1. Create store with a European address, WCS and Jetpack installed
2. Go to the Tax task and click to configure them manually on settings
3. Configure the taxes on Settings and go back to the task list
4. Enter in the Tax task and see a different message

**Screenshots**
https://www.loom.com/share/ce07ff8682a44f8788d60b5bb17ff3f3

* OS: macOS Mojave 10.14.5
* Browser: Chrome v 78.0.3904.70

joshuatf:
Fixes (possibly) #3151

Checks for `wc_connect_options.tos_accepted` as well as `woocommerce_setup_jetpack_opted_in` in the tax step.

I believe what is happening is that `woocommerce_setup_jetpack_opted_in` gets updated to `true` in the profiler.  This usually will cause `wc_connect_options.tos_accepted` to be set to `true` on `before_woocommerce_init` but because the page is never reloaded, this event does not get triggered.

`wc_connect_options` is preloaded so the old value will still be in place, causing the tax step to initially not allow automatic taxes until the page is refreshed or the fresh data freshness to expire.

@pmcpinto Can you test out this branch to see if it resolves the issue you were having?  [Here's a prebuilt wc admin zip.](https://mc.a8c.com/includes/img-uploader/files/1573719878-woocommerce-admin-onboarding.zip)

### Screenshots
Screen Shot 2019-11-14 at 4 22 51 PM

1. On a fresh site, walk through the profiler being careful to not refresh the page.
2. Complete the profiler with a valid tax jar country and installing WCS and Jetpack.
3. Go to the task list tax step.
4. After data is loaded, the above automatic tax screen should be shown.

### DIFF
--- a/client/dashboard/task-list/tasks/tax.js
+++ b/client/dashboard/task-list/tasks/tax.js
@@ -106,12 +106,12 @@ class Tax extends Component {
 	}
 
 	isTaxJarSupported() {
-		const { countryCode, wc_connect_options } = this.props;
+		const { countryCode, tosAccepted } = this.props;
 		const { automatedTaxSupportedCountries = [], taxJarActivated } = getSetting( 'onboarding', {} );
 
 		return (
 			! taxJarActivated && // WCS integration doesn't work with the official TaxJar plugin.
-			wc_connect_options.tos_accepted &&
+			tosAccepted &&
 			automatedTaxSupportedCountries.includes( countryCode )
 		);
 	}
@@ -380,11 +380,10 @@ export default compose(
 		const countryCode = getCountryCode( generalSettings.woocommerce_default_country );
 		const activePlugins = getActivePlugins();
 		const pluginsToActivate = difference( [ 'jetpack', 'woocommerce-services' ], activePlugins );
-		const wc_connect_options = get(
-			getOptions( [ 'wc_connect_options' ] ),
-			'wc_connect_options',
-			{}
-		);
+		const options = getOptions( [ 'wc_connect_options', 'woocommerce_setup_jetpack_opted_in' ] );
+		const wc_connect_options = get( options, 'wc_connect_options', {} );
+		const tosAccepted =
+			wc_connect_options.tos_accepted || options.woocommerce_setup_jetpack_opted_in;
 
 		return {
 			countryCode,
@@ -396,7 +395,7 @@ export default compose(
 			taxSettings,
 			isJetpackConnected: isJetpackConnected(),
 			pluginsToActivate,
-			wc_connect_options,
+			tosAccepted,
 		};
 	} ),
 	withDispatch( dispatch => {


pmcpinto:
Hey @joshuatf and @justinshreve, I just tested it with a store with Portugal as location (one of the TaxJar countries) and it still asks me to fill the taxes manually:

https://www.loom.com/share/5e8077bb416b46e1a033cf2906d098d7

A couple of mins later I entered again in the Tax task and now it showed me the Automated tax message:

https://www.loom.com/share/5245d668bea1423089bc16dae5a18556

joshuatf:
@pmcpinto We're having a hard time recreating the issue to even see manual configuration the first time.  Could you share a video of your store from the start of the profiler all the way to the tax step?

Maybe that will help us uncover something we're missing about the setup that causes this issue.

pmcpinto:
Hey @joshuatf, here it goes the video: https://www.loom.com/share/95c9d2d0d2c14fb7914fc16e5d929984

Usually it happens this: initially it asks me to fill the taxes manually and a couple of mins later it shows me the Automated taxes message

joshuatf:
@pmcpinto Sorry this is taking so long, but I'm still unable to replicate following the steps in the video and using the same settings.  Maybe there's something specific to your environment.

* Do you see this on Jurassic Ninja sites?  Somewhere else?  What's your setup for your WP site?
* Any other plugins aside from WCS, Jetpack, WC, and WC-Admin?

pmcpinto:
Hey @joshuatf, here it goes my feedback:

> Do you see this on Jurassic Ninja sites? Somewhere else? What's your setup for your WP site?

I'm using Pressable, WP 5.3

> Any other plugins aside from WCS, Jetpack, WC, and WC-Admin?

Just Akismet Anti-Spam which comes installed when I create a WP site on Pressable. Btw, Jetpack it's also installed when I create a new site.

joshuatf:
@pmcpinto I managed to recreate the issue on a Pressable site.  It looks Pressable is using pretty heavy object caching, so the TOS values in the database are not updated for quite some time.

I would think that a call to `update_option()` would invalidate that in their object cache.  It's possible they already have something like this, but that it doesn't check during rest requests.  Should we open an issue with Pressable to address this? /cc @justinshreve @rrennick

rrennick:
> click "Flush Object Cache" to notice that this resolves the issue immediately.

In that case, the updated value is in the DB but not where it's in the cache. It's possible that the REST response itself is being cached.

joshuatf:
>In that case, the updated value is in the DB but not where it's in the cache. It's possible that the REST response itself is being cached.

Confirmed that the updated value is in the DB.  I don't think the REST response is being cached (or if it is, it's an additional layer to the object caching).  When dumping the options directly from a wp-admin screen `wp_die( get_option( 'option_name' ) )` the old values are still shown.

justinshreve:
> When dumping the options directly from a wp-admin screen wp_die( get_option( 'option_name' ) ) the old values are still shown.
> Should we open an issue with Pressable to address this? /cc @justinshreve @rrennick

That sounds like a good path forward, since we should be able to rely on `update_option` updating any option caches.

joshuatf:
Made a note of this on the Pressable channel and will follow-up there.  In the meantime, I'm going to go ahead and merge this PR as I think it still offers a safety net in case the page has not been refreshed.

PR #3247, opened 2019-11-14
Make WooCommerce brand localizable

### LINKED ISSUE #3246: WooCommerce Admin does not respect WooCommerce Branding
**Describe the bug**
WooCommerce Admin does not allow the WooCommerce Branding extension to replace "WooCommerce" with your shop's brand name.

**To Reproduce**
Have both WooCommerce Admin and WooCommerce Branding enabled, and a custom "name" for the shop set in the WooCommerce Branding settings. This name is used by all other extensions, but WooCommerce Admin continues to use "WooCommerce" in the Admin dashboard's breadcrumbs.

**Expected behavior**
WooCommerce Admin should use the name set by WooCommerce Branding in the Admin dashboard breadcrumbs and anywhere else "WooCommerce" might appear in WooCommerce Admin.

**Screenshots**
woocommerce-admin-dashboard-branding-name

**Desktop**

* OS: Windows 10 1909
* Firefox
* Version 70.0.1 64-bit

Aljullu:
Closes #3246

The word _WooCommerce_ is localizable in WooCommerce Core:

https://github.com/woocommerce/woocommerce/blob/master/includes/admin/class-wc-admin-menus.php#L62

So it makes sense making it localizable here too.

### DIFF
--- a/client/dashboard/profile-wizard/header-logo.js
+++ b/client/dashboard/profile-wizard/header-logo.js
@@ -2,6 +2,7 @@
 /**
  * External dependencies
  */
+import { __ } from '@wordpress/i18n';
 import { Component } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import classNames from 'classnames';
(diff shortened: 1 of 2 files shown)


rrennick:
@Aljullu There are 2 other strings here:

https://github.com/woocommerce/woocommerce-admin/blob/master/client/dashboard/profile-wizard/header-logo.js#L17

Aljullu:
Good catch @rrennick! Should be fixed in the last commit (10536f4c323ac38723b87e92a7e4f9cc974c38ab).

jeffstieler:
Will this fix https://github.com/woocommerce/woocommerce-admin/issues/3246?

rrennick:
> Will this fix #3246?

Good catch. Yes.

Aljullu:
@jeffstieler @rrennick I don't think it fixes it. I initially thought WooCommerce Branding would hook up into the translation system to replace the name, but it seems to do it with some other hacks.

So I created this PR after #3246, but it doesn't fix it.

rrennick:
> I initially thought WooCommerce Branding would hook up into the translation system to replace the name

It does, but only the translations handled in PHP.

PR #3250, opened 2019-11-14
Add a tracks queue to delay event recording & log embed page views

justinshreve:
While fixing some issues in #3220, I added an event that would fire during a window.location redirect. Josh raised a potential issue with `recordEvent` not finishing before the redirect happens.

After doing some research, we realized that this could happen and Calypso works around it with an analytics queue. They store the event details in localStorage (which will run before the redirect) and process the events on the next page view. The [Calypso code](https://github.com/Automattic/wp-calypso/blob/7b1cb7ba4f36656487e1ff6261b752bb5bcb7590/client/lib/analytics/index.js#L282) for it is actually pretty straight forward, so this PR adds our own implementation of it. You can use `queueRecordEvent` in these instances.

In addition, I've also added functionality to record embedded page views (the bits of wc-admin on classic pages). This allows the queue to run on classic pages as well, and seems useful for prioritizing future pages for overhaul (cc @timmyc).

* Enter `localStorage.setItem( 'debug', 'wc-admin:*' );` into your console. Leave your console open.
* Enable the task list
* Delete your homepage
* Go to the appearance task and create a new homepage
* Publish, and click the “continue setup” button
* You should see some debug about the event getting added to the queue (if the redirect happens too fast, there will be more debug on the next page)
* When the appearance task loads again, you should see a debug message about the queue processing, and the `tasklist_appearance_continue_setup` event will fire
* Disable WooCommerce Services and delete your tax rates
* Go to the tax task and click 'skip' when you are offered Jetpack & WooCommerce Services
* You should see the `tasklist_tax_install_extensions` event fire on the tax settings page.

Screen Shot 2019-11-14 at 11 01 59 AM

### DIFF
--- a/client/dashboard/task-list/tasks/appearance.js
+++ b/client/dashboard/task-list/tasks/appearance.js
@@ -14,7 +14,7 @@ import { withDispatch } from '@wordpress/data';
  * WooCommerce dependencies
  */
 import { Card, Stepper, TextControl } from '@woocommerce/components';
-import { getHistory, getNewPath, getQuery } from '@woocommerce/navigation';
+import { getHistory, getNewPath } from '@woocommerce/navigation';
 import { getSetting, setSetting } from '@woocommerce/wc-admin-settings';
 
 /**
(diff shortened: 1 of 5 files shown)


timmyc:
Very good idea to get some inspiration on how to fix this problem from Calypso @justinshreve ! Do you think this should get put in woo core at some point in case other teams encounter a similar issue? I suppose we can wait and see if that happens then move it to the core tracks implementation down the line.

justinshreve:
Agreed it would be great to get this usable outside our context too. Logged an issue for that here: https://github.com/woocommerce/woocommerce-admin/issues/3260.

PR #3251, opened 2019-11-14
make store revenue range amounts translatable

### LINKED ISSUE #2768: Business Details Revenue Options i18n

rrennick:
Fixes #2768

This PR adds translation & translation instructions for the revenue amount ranges on the business details screen of onboarding.

- Reset onboarding with `wp option delete wc_onboarding_profile`
- Enable onboarding in wp-config.php with `WOOCOMMERCE_ADMIN_ONBOARDING_ENABLED`
- Navigate through the onboarding flow to the business details
- Select one of the physical location options
- Choose one of the revenue amounts
- Continue
- Check the saved value `wp option get wc_onboarding_profile`

### Changelog Note:

Fix: Enable translation of currency amounts for business details.


rrennick:
> I think we might still want to hard code the currency symbol here as well

Do you mean add the currency symbol to the option value? Currently, it's only displayed in the option content.

justinshreve:
I mean the translated string would be something like

`label: __( '$2,500 - $10,000', 'equivalent of range $2500 - $10000 USD', 'woocommerce-admin' )`

So that the entire dollar value and range is translated for this question (which would avoid the store/client currency and the store language being different and showing a confusing range).

rrennick:
> So that the entire dollar value and range is translated for this question

@justinshreve I've updated the PR for this.

Issue #2986, opened 2019-10-02
Fix/Explain package version discrepency between @woocommerce/components and main package.json


nerrad:
Recently brought up in a slack convo:

> any special reason why https://github.com/woocommerce/woocommerce-admin/blob/master/packages/components/package.json#L29 isnt' using the latest @wordpress/components?

Note, current version of `@wordpress/components` in the main package.json is **8.0**

https://github.com/woocommerce/woocommerce-admin/blob/master/package.json#L72

PR #3253, opened 2019-11-15
Use same wp-components version in bundle

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/2986

This PR aligns version numbers for `wp-components` for the main package and our own components.

I also noticed that we bundle wp-components, yet still require it as a resource. So I removed it as a dependency for loaders.

## Test

1. `'npm start` and check the app works without causing any errors.
2. In the console, confirm `wp.components` is undefined since we don't load it anymore


psealock:
Thanks for the review. 

I added a change to allow short array syntax in php since [this is allowed with 5.6](https://www.php.net/manual/en/language.types.array.php).

PR #3262, opened 2019-11-15
Fetch autocomplete countries from the API rather than wcSettings object.

### LINKED ISSUE #3028: Refactor: Decouple wcSettings (via @woocommerce/wc-admin-settings) from countries autocompleter
## Context:

- #2999 is the master ticket.

### `countries` autocompleter (search/autocompleters/countries.js)
This autocompleter is coupled to the `dataEndpoints.countries` setting. I think this should be refactored so that either:

- countries comes from the api for the autocompleter or
- countries is provided to the `Autocomplete` component as a prop and the countries autocompleter can be a function that spits out the configuration with the given countries data hydrating it (eg `getCountriesCompleter( countries)` )

The latter might be a good interim way to do things but does change the interface a bit.

jeffstieler:
Fixes #3028.

This PR seeks to decouple the countries autocompleter from the `wcSettings` object, fetching them from the API instead. It also fixes a bug in the country advanced filter in the Customers Report.

- Open the Customers Report
- Add a country advanced filter
- Verify it's function (note that there appears to be a bug, in `master` as well, that keeps the selected country text in the input after selection)

### Changelog Note:

Tweak: Search component: remove dependency on settings global from countries autocompleter.

### DIFF
--- a/client/analytics/report/customers/config.js
+++ b/client/analytics/report/customers/config.js
@@ -120,13 +120,13 @@ export const advancedFilters = applyFilters( CUSTOMERS_REPORT_ADVANCED_FILTERS_F
 				type: 'countries',
 				getLabels: async value => {
 					const allLabels = countries.map( country => ( {
-						id: country.code,
+						key: country.code,
 						label: decodeEntities( country.name ),
 					} ) );
 
 					const labels = value.split( ',' );
 					return await allLabels.filter( label => {
-						return labels.includes( label.id );
+						return labels.includes( label.key );
 					} );
 				},
 			},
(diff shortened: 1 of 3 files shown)


rrennick:
@jeffstieler As soon as I click on the text entry field in the country filter I get errors in my console

```
index.js:16 Uncaught TypeError: autocompleter.options(...).then is not a function
    at Search.fetchOptions (index.js:16)
    at SelectControl.search (index.js:420)
```

jeffstieler:
@rrennick I'm unable to reproduce using the country filter on the customers report. Where are you seeing the error?

PR #3267, opened 2019-11-18
Add wc admin card classes to leaderboards

### LINKED ISSUE #3075: Leaderboards: Header styles regression
**Describe the bug**

In the dasboard's leaderboards section, table headers have something off in the CSS. As seen below, the title styles on the right are correct.

Screen Shot 2019-10-22 at 9 44 48 PM

**To Reproduce**

1. Go to the Dashboard and see Leaderboards section

joshuatf:
Fixes #3075

Adds the analytics card class to various states for leaderboards to avoid default card styling.

### Before
Screen Shot 2019-11-18 at 2 12 16 PM

### After
Screen Shot 2019-11-18 at 2 15 27 PM

1. Go to the leaderboards section in the dashboard.
2. Make sure the leaderboard card header styles match other analytic card styles in wc-admin.
3. Change the date to a time where data exists and does not exist for that leaderboard.  Make sure all states show the correct leaderboard header style (no data, loading, and has data).

### DIFF
--- a/client/analytics/components/leaderboard/index.js
+++ b/client/analytics/components/leaderboard/index.js
@@ -3,6 +3,7 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
+import classnames from 'classnames';
 import { Component } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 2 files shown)


client/analytics/components/leaderboard/index.js
```diff
@@ -50,13 +51,15 @@ export class Leaderboard extends Component {
 		const { isRequesting, isError, totalRows, title } = this.props;
 		const rows = this.getFormattedRows();
 
+		const classes = classnames( 'woocommerce-leaderboard', 'woocommerce-analytics__card' );
```
jeffstieler: Nitpick - this doesn't need to use `classnames` - it could just be a string literal.

jeffstieler:
Are we using the default `Card` styling anywhere? If not, we might want to consider making the default style the same as what we're using in analytics.

joshuatf:
>Are we using the default Card styling anywhere? If not, we might want to consider making the default style the same as what we're using in analytics.

@jeffstieler We're using the default Card styling for onboarding, but this may eventually be updated and the current plan is to import and restyle the Gutenberg `Card` component across all of wc-admin.  We'll update accordingly once we have access to that component (relevant discussion - p1574215904392900-slack-wc-admin ).

PR #3277, opened 2019-11-19
Decouple currency and numbers packages from `wcSettings`.

### LINKED ISSUE #3029: Refactor: Decouple wcSettings (via @woocommerce/wc-admin-settings) from packages/currency functions
## Context:

- #2999 is the master ticket.

### `packages/currency` functions

Many of these helper functions would serve as really useful utility methods on `Currency` and Money value objects.

Some examples that I created for another project:

- https://github.com/eventespresso/event-espresso-core/blob/master/assets/src/vo/currency.js
- https://github.com/eventespresso/event-espresso-core/blob/master/assets/src/vo/money.js

We could then decouple settings from the package and consumers would just instantiate the value objects as needed for the app.

So I would suggest decoupling by implementing value objects.

### LINKED ISSUE #3032: Refactor: Decouple wcSettings (via @woocommerce/wc-admin-settings) from packages/number
## Context:

- #2999 is the master ticket.

### `packages/number`

The coupling here is for the `CURRENCY` object exposed by wcSettings. My comments here are similar to what I mentioned for `packages/currency` in that the method that currency is used in here (`numberFormat`) can likely be a part of a `Money` value object.

jeffstieler:
Fixes #3029.
Fixes #3032.

This PR seeks to remove the `wcSettings` dependency from the Currency and Number packages.

### `Currency`

The currency package has been rewritten to export a `Currency` class instead of several utility functions. It takes an object in the constructor for configuration - which matches the structure of the `wcSettings` currency options.

A new `lib/currency-format` module has been introduced that exports the same API as the previous version of `@woocommerce/currency` so that the consuming parts of the WC-Admin app don't need to change (too much).

### `Number`

The exported methods of the number package have been rewritten to accept a configuration object as their first parameter - this allows for composition using lodash's `partial()` - which is done in the new `lib/number-format` module.

A new `lib/number-format` module has been introduced that exports the same API as the previous version of `@woocommerce/number` so that the consuming parts of the WC-Admin app don't need to change (too much).

### `AdvancedFilters`

Make the `currency` prop required, and expect an instance of the new `Currency` object.

- Verify that numbers and monetary values are properly formatted on all WC-Admin screens (check activity panels, reports, filters, etc)
- Verify that changing store currency settings changes the formatting
- Review changes to documentation

### Changelog Note:

Fix: decouple Currency and Number packages from global wcSettings object.

### DIFF
--- a/client/analytics/components/leaderboard/test/index.js
+++ b/client/analytics/components/leaderboard/test/index.js
@@ -8,8 +8,8 @@ import { mount, shallow } from 'enzyme';
 /**
  * WooCommerce dependencies
  */
-import { formatCurrency, getCurrencyFormatDecimal } from '@woocommerce/currency';
-import { numberFormat } from '@woocommerce/number';
+import { formatCurrency, getCurrencyFormatDecimal } from 'lib/currency-format';
+import { numberFormat } from 'lib/number-format';
 
 /**
  * Internal dependencies
--- a/client/analytics/components/report-filters/index.js
+++ b/client/analytics/components/report-filters/index.js
@@ -10,12 +10,13 @@ import { omitBy, isUndefined, snakeCase } from 'lodash';
  * WooCommerce dependencies
  */
 import { ReportFilters as Filters } from '@woocommerce/components';
-import { LOCALE, CURRENCY } from '@woocommerce/wc-admin-settings';
+import { LOCALE } from '@woocommerce/wc-admin-settings';
 
 /**
  * Internal dependencies
  */
 import { recordEvent } from 'lib/tracks';
+import { Currency } from 'lib/currency-format';
 
 export default class ReportFilters extends Component {
 	constructor() {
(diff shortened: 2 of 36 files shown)


rrennick:
@jeffstieler In the Products screen there are integer columns that have decimal places.

Screen Shot 2019-11-19 at 1 35 41 PM

jeffstieler:
@rrennick fixed in 34101e3.

jeffstieler:
@nerrad - do you have the same concerns with passing the `Currency` instance as a prop here, like in https://github.com/woocommerce/woocommerce-admin/pull/3278#pullrequestreview-320202815?

jeffstieler:
@nerrad made the object immutable in 555da66. I'm going to hold off on setters since our current use case is to tie a single instance to the store's settings (which I don't believe get updated in the client app without a page refresh).

PR #3278, opened 2019-11-19
Decouple date package from `wcSettings`.

### LINKED ISSUE #3030: Refactor: Decouple wcSettings (via @woocommerce/wc-admin-settings) from packages/date
## Context:

- #2999 is the master ticket.

### `packages/date`

There are two places where settings data is coupled to the functions:

#### 1. `getDateParamsFromQuery`.

The `defaultDateRange` argument is derived from `woocommerce_default_date_range` setting. This should be passed in as a argument to the function and callers of `getDateParamsFromQuery` can provide the default.

#### 2. `loadLocaleData`.

This uses `userLocale` and `weekdaysShort` from the `locale` object on the settings.  It is also called directly in this file.  imo this should receive `locale` as an argument and it should be instantiated by client, not in the package.

jeffstieler:
Fixes #3030.

This PR seeks to remove the `wcSettings` dependency from the Date package.

### `Date`

The exported methods of the date package that depended on `getSetting()` have been rewritten to accept a configuration object as their second parameter - this allows for composition using lodash's `partialRight()` - which is done in the new `lib/date` module.

`loadLocaleData()` is also no longer called within the date package, but in the `lib/date` module instead.

A new `lib/date` module has been introduced that exports the same API as the previous version of `@woocommerce/date` so that the consuming parts of the WC-Admin app don't need to change (too much).

### `AdvancedFilters` and `ReportFilters`

Add a new required `storeDate` prop as a means to pass down date initialization values from `@woocommerce/date`, but from `lib/date` in the app.

- Verify that date values are properly formatted on all WC-Admin screens (check activity panels, reports, filters, etc)
- Verify that the default date range analytics setting functions
- Review changes to documentation

### Changelog Note:

Fix: decouple Date package from global wcSettings object.

### DIFF
--- a/packages/components/src/date-range-filter-picker/index.js
+++ b/packages/components/src/date-range-filter-picker/index.js
@@ -6,7 +6,6 @@ import { Component, createRef } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import { Dropdown } from '@wordpress/components';
 import PropTypes from 'prop-types';
-import moment from 'moment';
 
 /**
  * Internal dependencies
@@ -32,16 +31,25 @@ class DateRangeFilterPicker extends Component {
 		this.resetCustomValues = this.resetCustomValues.bind( this );
 	}
 
+	formatDate( date, format ) {
+		if ( date.isMoment && date.isMoment() ) {
+			return date.format( format );
+		}
+
+		return '';
+	}
+
 	getResetState() {
 		const { period, compare, before, after } = this.props.dateQuery;
+
 		return {
 			period,
 			compare,
 			before,
 			after,
 			focusedInput: 'startDate',
-			afterText: after ? after.format( shortDateFormat ) : '',
-			beforeText: before ? before.format( shortDateFormat ) : '',
+			afterText: this.formatDate( after, shortDateFormat ),
+			beforeText: this.formatDate( before, shortDateFormat ),
 			afterError: null,
 			beforeError: null,
 		};
@@ -60,8 +68,8 @@ class DateRangeFilterPicker extends Component {
 				compare,
 			};
 			if ( 'custom' === selectedTab ) {
-				data.after = after ? after.format( isoDateFormat ) : '';
-				data.before = before ? before.format( isoDateFormat ) : '';
+				data.after = this.formatDate( after, isoDateFormat );
+				data.before = this.formatDate( before, isoDateFormat );
 			} else {
 				data.after = undefined;
 				data.before = undefined;
@@ -162,8 +170,8 @@ DateRangeFilterPicker.propTypes = {
 	dateQuery: PropTypes.shape( {
 		period: PropTypes.string.isRequired,
 		compare: PropTypes.string.isRequired,
-		before: PropTypes.instanceOf( moment ),
-		after: PropTypes.instanceOf( moment ),
+		before: PropTypes.object,
+		after: PropTypes.object,
 		primaryDate: PropTypes.shape( {
 			label: PropTypes.string.isRequired,
 			range: PropTypes.string.isRequired,
--- a/packages/components/src/date-range-filter-picker/index.js
+++ b/packages/components/src/date-range-filter-picker/index.js
@@ -32,7 +32,7 @@ class DateRangeFilterPicker extends Component {
 	}
 
 	formatDate( date, format ) {
-		if ( date.isMoment && date.isMoment() ) {
+		if ( date && date.isMoment && date.isMoment() ) {
 			return date.format( format );
 		}
 
--- a/packages/components/src/date-range-filter-picker/index.js
+++ b/packages/components/src/date-range-filter-picker/index.js
@@ -32,7 +32,7 @@ class DateRangeFilterPicker extends Component {
 	}
 
 	formatDate( date, format ) {
-		if ( date && date.isMoment && date.isMoment() ) {
+		if ( date && date._isAMomentObject && ( 'function' === typeof date.format ) ) {
 			return date.format( format );
 		}
 
--- a/packages/components/src/date-range-filter-picker/index.js
+++ b/packages/components/src/date-range-filter-picker/index.js
@@ -167,14 +167,10 @@ DateRangeFilterPicker.propTypes = {
 		primaryDate: PropTypes.shape( {
 			label: PropTypes.string.isRequired,
 			range: PropTypes.string.isRequired,
-			after: PropTypes.instanceOf( moment ).isRequired,
-			before: PropTypes.instanceOf( moment ).isRequired,
 		} ).isRequired,
 		secondaryDate: PropTypes.shape( {
 			label: PropTypes.string.isRequired,
 			range: PropTypes.string.isRequired,
-			after: PropTypes.instanceOf( moment ).isRequired,
-			before: PropTypes.instanceOf( moment ).isRequired,
 		} ).isRequired,
 	} ).isRequired,
 };
--- a/packages/components/src/date-range-filter-picker/index.js
+++ b/packages/components/src/date-range-filter-picker/index.js
@@ -8,11 +8,6 @@ import { Dropdown } from '@wordpress/components';
 import PropTypes from 'prop-types';
 
 /**
- * WooCommerce dependencies
- */
-import { getCurrentDates, getDateParamsFromQuery, isoDateFormat } from '@woocommerce/date';
-
-/**
  * Internal dependencies
  */
 import DatePickerContent from './content';
@@ -37,7 +32,9 @@ class DateRangeFilterPicker extends Component {
 	}
 
 	getResetState() {
-		const { period, compare, before, after } = getDateParamsFromQuery( this.props.query );
+		const { storeDate, query } = this.props;
+		const { getDateParamsFromQuery } = storeDate;
+		const { period, compare, before, after } = getDateParamsFromQuery( query );
 		return {
 			period,
 			compare,
@@ -56,7 +53,7 @@ class DateRangeFilterPicker extends Component {
 	}
 
 	onSelect( selectedTab, onClose ) {
-		const { onRangeSelect } = this.props;
+		const { isoDateFormat, onRangeSelect } = this.props;
 		return event => {
 			const { period, compare, after, before } = this.state;
 			const data = {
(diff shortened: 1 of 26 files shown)


jeffstieler:
> Neat approach here but I'm still a bit wary about the strict expectations the date components have not only on the shape of the `storeDate` prop, but also the behaviour of functions it exposes. It'd be nice to avoid that tight coupling and have the date components only receive initialization values.

Totally agree - I forgot about the type of public API this would be creating and only focused on changing as little code as possible and keeping the number of props low.

jeffstieler:
Simplified the expected props for the date filter in 324ae1c.

jeffstieler:
@nerrad are you OK with creating an issue to work the changes requested to `ReportFilters` at a later time? It doesn't seem to be related to the `wcSettings` refactor, and isn't blocking components release - right?

packages/components/src/date-range-filter-picker/index.js
```diff
@@ -6,6 +6,7 @@ import { Component, createRef } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import { Dropdown } from '@wordpress/components';
 import PropTypes from 'prop-types';
+import moment from 'moment';
```
nerrad: Why is this imported here?  `moment` is a fairly heavy package so ideally it wouldn't be a dependency for the published `@woocommerce/components` package.  I see that you are passing along a moment instance and thus checking for it with prop-types.  However by virtue of this import, it means that you are also bundling moment with `@woocommerce/components`.  I don't think the additional prop-type check is worth it for this additional bundle weight.

packages/components/src/date-range-filter-picker/index.js
```diff
@@ -160,21 +157,26 @@ DateRangeFilterPicker.propTypes = {
...
+	dateQuery: PropTypes.shape( {
+		period: PropTypes.string.isRequired,
+		compare: PropTypes.string.isRequired,
+		before: PropTypes.instanceOf( moment ),
+		after: PropTypes.instanceOf( moment ),
+		primaryDate: PropTypes.shape( {
+			label: PropTypes.string.isRequired,
+			range: PropTypes.string.isRequired,
+			after: PropTypes.instanceOf( moment ).isRequired,
+			before: PropTypes.instanceOf( moment ).isRequired,
```
nerrad: To avoid importing moment, can you type check against `object` here instead and then add defensive programming elsewhere in whatever component is actually _implementing_  `before`/`after` (i.e. [you could check if `moment.isMoment` is present and `true`](https://momentjscom.readthedocs.io/en/latest/moment/05-query/10-is-a-moment/)).

packages/components/src/filters/index.js
```diff
@@ -6,6 +6,7 @@ import { __ } from '@wordpress/i18n';
 import { Component, Fragment } from '@wordpress/element';
 import { find } from 'lodash';
 import PropTypes from 'prop-types';
+import moment from 'moment';
```
nerrad: Same thing here, avoid creating a dependency on `moment` for `@woocommerce/components`.

packages/components/src/date-range-filter-picker/index.js
jeffstieler: You're right - an import of this size isn't worth the type check. I'll rework it.

jeffstieler: On second thought - `moment` is already part of `@woocommerce/components` via `DatePicker`, `DateFilter`, `DateRange`, and `D3Chart`.

nerrad: Which in my opinion is unfortunate (see #3289) (moment contributes ~240kb to the package weight).  I still think you should avoid the import here so it's easier to address bundle size later, especially when you can simply use something like the following here:

```js
if ( after.isMoment && after.isMoment() ) {
 // stuff
}
```

jeffstieler:
> Looks like there should be a moment verification in the onSelect method? Also after and before is passed through to another component, does that component check for moment?

I just pushed the WIP commit to move to my laptop 😄

jeffstieler:
Ok, this is ready for another review!

nerrad:
Code wise this seems okay.  However, there's the following issues when I test:

#### Error when selecting custom dates from the date-picker  for a report

The report I tested with was orders.
I got the following error after I selected a custom date range and submitted it:

```
index.js:225 Uncaught TypeError: Cannot read property 'diff' of null
    at getDateValue (index.js:225)
    at getCurrentDates (index.js:289)
    at wrapper (lodash.js?ver=4.17.15:5193)
    at ReportFilters.render (index.js:69)
    at finishClassComponent (react-dom.js?ver=16.9.0:17177)
    at updateClassComponent (react-dom.js?ver=16.9.0:17132)
    at beginWork$1 (react-dom.js?ver=16.9.0:18643)
    at HTMLUnknownElement.callCallback (react-dom.js?ver=16.9.0:341)
    at Object.invokeGuardedCallbackDev (react-dom.js?ver=16.9.0:391)
    at invokeGuardedCallback (react-dom.js?ver=16.9.0:448)
```

#### Deprecated react warning

This can probably be handled in a separate issue, but I get a deprecation warning from the version of React that ships with WP 5.3 for the usage of `componentWillMount`.  That's probably something that should be addressed at some point.

### moment deprecation warning

This is also something that could be handled in a separate issue because I think it happens in master as well.   I get this deprecation notice when viewing the settings page:

```
Deprecation warning: value provided is not in a recognized RFC2822 or ISO format. moment construction falls back to js Date(), which is not reliable across all browsers and versions. Non RFC2822/ISO date formats are discouraged and will be removed in an upcoming major release. Please refer to http://momentjs.com/guides/#/warnings/js-date/ for more info.
```
Stack trace:
```
Arguments: 
[0] _isAMomentObject: true, _isUTC: false, _useUTC: false, _l: undefined, _i: false, _f: undefined, _strict: undefined, _locale: [object Object]
Error
    at Function.createFromInputFallback (https://one.wordpress.test/wp-includes/js/dist/vendor/moment.js?ver=2.22.2:320:98)
    at configFromInput (https://one.wordpress.test/wp-includes/js/dist/vendor/moment.js?ver=2.22.2:2606:19)
    at prepareConfig (https://one.wordpress.test/wp-includes/js/dist/vendor/moment.js?ver=2.22.2:2577:13)
    at createFromConfig (https://one.wordpress.test/wp-includes/js/dist/vendor/moment.js?ver=2.22.2:2544:44)
    at createLocalOrUTC (https://one.wordpress.test/wp-includes/js/dist/vendor/moment.js?ver=2.22.2:2631:16)
    at createLocal (https://one.wordpress.test/wp-includes/js/dist/vendor/moment.js?ver=2.22.2:2635:16)
    at hooks (https://one.wordpress.test/wp-includes/js/dist/vendor/moment.js?ver=2.22.2:12:29)
    at HistoricalDataStatus (https://one.wordpress.test/wp-content/plugins/woocommerce-admin/dist/app/index.js?ver=1573509263:9168:270)
    at renderWithHooks (https://one.wordpress.test/wp-includes/js/dist/vendor/react-dom.js?ver=16.9.0:15246:18)
    at updateFunctionComponent (https://one.wordpress.test/wp-includes/js/dist/vendor/react-dom.js?ver=16.9.0:17063:20) 
    in HistoricalDataStatus (created by FilteredComponent)
    in FilteredComponent (created by HistoricalDataLayout)
    in HistoricalDataLayout (created by ComponentWithSelect)
    in ComponentWithSelect (created by Context.Consumer)
    in Context.Consumer (created by WithSelect(HistoricalDataLayout))
    in WithSelect(HistoricalDataLayout) (created by HistoricalData)
    in HistoricalData (created by WithSpokenMessages(HistoricalData))
    in WithSpokenMessages(HistoricalData) (created by Settings)
    in Settings (created by FilteredComponent)
    in FilteredComponent (created by WithDispatch(FilteredComponent))
    in WithDispatch(FilteredComponent) (created by ComponentWithSelect)
    in ComponentWithSelect (created by Context.Consumer)
    in Context.Consumer (created by WithSelect(WithDispatch(FilteredComponent)))
    in WithSelect(WithDispatch(FilteredComponent)) (created by Controller)
    in Controller (created by Layout)
    in Layout (created by Router.Consumer)
    in Router.Consumer (created by Route)
    in Route (created by _PageLayout)
    in _PageLayout (created by FilteredComponent)
    in FilteredComponent
```

Also, something to consider for another issue is to remove the dependency on moment altogether in these packages.  Although what we have here is sufficient for now.  What if down the road, WordPress switches away from moment, or we switch to a different date package because it's less weight and works better (or even can start using native js i18n date functions when they become more widely available)?  With the current implementation, there'd be some gotchas we'd have to deal with.

So what would be the solution?  Create a Date Adaptor that our components use instead.  Under the hood it could be powered by moment, but we'd be able to swap it out for any other date package down the road without messing with any other code because we control the interface.  Something to think about.  When it comes to things like dates and money values, I _really_ think it's critical to control the interfaces a bit more and keep any libraries used behind an abstraction so they can be swapped out more easily.

jeffstieler:
I'm not getting the moment deprecation notice on the settings page (Analytics > Settings, right?). Did it require any interactions? I tried using a preset date range and updating the settings, reloading the page. Never saw the notice.

nerrad:
> I'm not getting the moment deprecation notice on the settings page (Analytics > Settings, right?).

Hmm.  Maybe it depends on WP version (and thus the version of moment bundled with it)?  I'm running WP 5.3.

I also wonder if it might have to do with blocks being active?  Maybe it's affecting what settings environment is present.

Or maybe it has to do with progress in onboarding.  If so, this would be a tricky one.  I frequently just try to skip through the onboarding steps so maybe I borked something.

Regardless of the source, it _appears_ to be coming from this:

https://github.com/woocommerce/woocommerce-admin/blob/9105a315e22bac682b479726815521c50c1d27a7/client/analytics/settings/historical-data/status.js#L26-L28

Which means that there should probably be some validation that `importDate` is in a `RFC2822` or `ISO format`.  But again, could be handled in a separate issue because I think this is happening with master too.

packages/components/src/date-range-filter-picker/index.js
jeffstieler: It looks like we both misread/understood this interface. `moment.isMoment()` as referenced in the docs is a utility function on the exported `moment` module, not on an instance of `moment()`. It looks like we can look for a `_isAMomentObject` property instead.

jeffstieler:
Thanks for the testing and review @nerrad - looks like I missed the `isoDateFormat` prop in the refactor. I've also fixed the Moment detection mechanism - this is ready for another test/review.

I'm going to hold off on the issues beyond this so we can land the refactor.

PR #3281, opened 2019-11-20
Move card styles to components

psealock:
Some card styles found themselves in `client/analytics/reports/styles.scss`. This meant Dashboard uses of `Card` didn't get these styles as well as extensions.

`woocommerce-analytics__card` was a class applied to override styles of Card in the report, but needed to be applied to every use of `Card`, so I removed the class and added the styles directly to Card.

`woocommerce-analytics__table-placeholder` was an unused class.

I deleted `client/analytics/reports/styles.scss` to prevent further use.

### Before

Screen Shot 2019-11-20 at 2 09 59 PM

Screen Shot 2019-11-20 at 2 10 15 PM

### After

Screen Shot 2019-11-20 at 2 09 41 PM

Screen Shot 2019-11-20 at 2 09 15 PM

1. Visit the dashboard and see the card header styles match the rest of the app.
2. Poke around to see if any regressions occured.


packages/components/src/card/style.scss
```diff
@@ -3,19 +3,35 @@
 .woocommerce-card {
 	margin-bottom: $gap-large;
 	background: $studio-white;
-	border-radius: 3px;
-	box-shadow: $muriel-box-shadow-1dp;
-	transition: box-shadow 0.2s cubic-bezier(0.4, 1, 0.4, 1);
+	border-radius: 0;
```
joshuatf: These style changes will break card styling in onboarding.

psealock: Thanks @joshuatf, I was fearing this. And sorry I didn't see your open PR.

>Are we using the default Card styling anywhere? If not, we might want to consider making the default style the same as what we're using in analytics.

~ from @jeffstieler https://github.com/woocommerce/woocommerce-admin/pull/3267#issuecomment-555016688

The answer I found is no. So how about adding an `onboarding_card` (or something) class to onboarding's use of Card to keep it separate? That will override base styles. One reason to do this the exported `Card` styles are broken and would remain so in your PR.

psealock:
Closing in favor of https://github.com/woocommerce/woocommerce-admin/pull/3267

PR #3285, opened 2019-11-20
Reports sync refactor

### LINKED ISSUE #2765: ReportsSync: Refactor
The logic to sync lookup tables has reached a point where a refactor can greatly improve the codebase. The code will be easier to read, less error prone, and easier to extend. 

### Current State
https://github.com/woocommerce/woocommerce-admin/blob/af6a49bfda543382d9f268de2ec5bb0ec2c6c419/src/ReportsSync.php

### Proposal
The sync and regeneration of this table currently run on install or category edit.  We should consider moving these into something like `ReportsSync` to allow re-syncing of data and potentially moving the sync into a background process.

If we do this, it may be worthwhile to abstract some of the base logic for syncing and add syncs into individual classes to make this easier to grok and prevent the `ReportsSync` class from becoming too large.  Each individual class could define items or hooks like:

* Dependencies
* Batch sizes
* Batch creation/deletion item retrieval

Fixes here should also address the sync mechanisms of the Category Lookup table introduced in https://github.com/woocommerce/woocommerce-admin/pull/2253.

joshuatf:
Fixes #2765

Refactors the action scheduling logic for orders and customers to make adding new data easier and more maintainable.  It does this by separating shared functionality for schedulers into `SchedulerTraits` and import functionality into `ImportScheduler`.

**SchedulerTraits**

This trait facilitates adding batch scheduling, action clearing, and action dependencies to any class.  This allows each scheduler to be declarative about its actions and dependencies so that other controllers don't need to be knowledgeable about all of its actions.  Controllers importing this trait should describe 3 items:

* `$name` - Unique identifying name for the scheduler.
* `get_actions()` - Returns an array of key => value pairs for the action name and the hook for that action respectively.
* `get_dependencies()` - Returns an array of key => values pairs for action names and a dependency hook to look for.

It also handles a bunch of trivial functions, such as suffixing actions with the scheduler name to help prevent collision between actions, rescheduling blocked/dependent actions, and flattening arguments sent to the scheduler.

**ImportScheduler**

This abstract class houses functionality that handles batch importing and deletion specific to importer schedulers.

In addition to basic scheduling traits, classes that inherit functions from `ImportScheduler` just need to describe a few methods to work:
* `get_items()` - This retrieves the original item IDs and total count.
* `get_total_imported()` - Used to determine how many items need to be deleted.
* `import()` - The function that action scheduler will run when a single import is being processed, passed the ID from `get_items()`.
* `delete()` - The function that action scheduler will run when deleting a batch of items, passed the number of items to delete.

1. Go to Analytics->Settings.
2. Test importing and deleting data.
3. Make sure data is imported correctly.
4. Check to make sure no oddities (unnecessary additional scheduled actions) occur with scheduled actions in `wp_posts`.


src/ReportExporter.php
```diff
@@ -97,18 +89,16 @@ public static function queue_report_export( $export_id, $report_type, $report_ar
...
 
 		if ( 0 < $num_batches ) {
-			ReportsSync::queue_batches( 1, $num_batches, self::REPORT_EXPORT_ACTION, $report_batch_args );
+			self::queue_batches( 1, $num_batches, 'export_report', $report_batch_args );
 
 			if ( $send_email ) {
 				$email_action_args = array( get_current_user_id(), $export_id, $report_type );
-
-				ReportsSync::queue_dependent_action( self::REPORT_EXPORT_EMAIL_ACTION, $email_action_args, self::REPORT_EXPORT_ACTION );
+				self::queue_batches( 1, $num_batches, 'email_report_download_link', $email_action_args );
```
jeffstieler: This needs to get reverted to a single dependent job. The batches running `export_report` create the CSV, and then after we should send a single email.

As this currently sits, it creates several unnecessary jobs that all produce no emails (because of the unexpected batch arguments).

src/ReportExporter.php
```diff
@@ -36,46 +36,38 @@ class ReportExporter {
...
-		if ( is_null( self::$queue ) ) {
-			self::$queue = WC()->queue();
-		}
-
-		return self::$queue;
+	 * @return array
+	 */
+	public static function get_actions() {
+		return array_merge(
+			self::get_scheduler_actions(),
```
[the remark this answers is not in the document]
joshuatf: Could you elaborate on your confusion here a bit more?  

The idea for `get_actions()` is to allow a scheduler to define its own actions. These actions are used later to help with hooks and action clearing clean-up. The method `get_scheduler_actions()` returns the default actions for a scheduler but any scheduler can opt to not use these and/or add their own. 

>What do you think about moving the actions of a SchedulerTraits consuming class into a parameter to be passed into SchedulerTraits::init() instead?

We could rename this method to `get_default_scheduler_actions()` which may make a bit more sense. But I feel passing an array to another method actually makes this more difficult to follow and does a poorer job of encapsulating actions  in specific schedulers.

This scheduler is missing a `$name` property, which is why these actions names are not fully fleshed out. These should look more like `wc-admin_queue_batches_report_exporter` if the `$name` is set to `report_exporter`. This is a good case for a separate interface to enforce these rules, but I think handling this in a follow-up might be best if we choose to do this.

src/Schedulers/SchedulerTraits.php
```diff
@@ -0,0 +1,353 @@
...
+				$action_hook,
+				$args,
+				static::$group
+			);
+
+			// We need to remove the action in case multiple jobs are being run at the same time.
+			remove_action( $action_hook, array( static::class, $action_name ), 10 );
+		} else {
+			add_action( $action_hook, array( static::class, $action_name ), 10, count( $args ) );
+		}
```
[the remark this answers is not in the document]
joshuatf: I like this approach better and it seems to work well. Added in https://github.com/woocommerce/woocommerce-admin/pull/3285/commits/4b659e3b8ce33e56647097a5e44382dbf88844a3

src/Schedulers/SchedulerTraits.php
```diff
@@ -0,0 +1,353 @@
...
+			if (
+				( static::get_action( $action_name ) === $existing_job->get_hook() ) ||
+				(
+					static::get_action( 'schedule_action' ) === $existing_job->get_hook() &&
+					in_array( self::get_action( $action_name ), $existing_job->get_args(), true )
+				)
+			) {
+				return true;
+			}
+		}
```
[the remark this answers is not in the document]
joshuatf: Nice catch.  Added in https://github.com/woocommerce/woocommerce-admin/pull/3285/commits/780613b3ccc701a36869770d20a63817f7c5993c

src/Schedulers/SchedulerTraits.php
```diff
@@ -0,0 +1,353 @@
...
+	/**
+	 * Get the next blocking job for an action.
+	 *
+	 * @param string $action_name Action name.
+	 */
+	public static function get_next_blocking_job( $action_name ) {
+		$dependency = self::get_dependency( $action_name );
+
+		if ( ! $dependency ) {
+			return;
```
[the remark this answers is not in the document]
joshuatf: Added in https://github.com/woocommerce/woocommerce-admin/pull/3285/commits/780613b3ccc701a36869770d20a63817f7c5993c

src/ReportExporter.php
joshuatf: You're right.  Updated this and the dependency (which was inverted) in https://github.com/woocommerce/woocommerce-admin/pull/3285/commits/45e2c8f08d817e9e74c75d69b39e2e5a617ed9aa

joshuatf:
Thanks for the review, @jeffstieler and @psealock!

Made some changes based on your feedback.

>What do you think about renaming the "Schedulers"? For some reason that name is throwing me off - perhaps Queue-something makes more sense?

I think Scheduler makes the most sense here unless you have any better alternatives. In this case I think we are scheduling actions to go into a queue.  I also think “scheduling” covers more aspects than “queueing," such as scheduling batches and checking dependencies as opposed to simply adding to a queue.

We could rename to something like `QueueActionTraits` and `QueueImportBase` but I think these are less accurate names for what these classes do.

src/ReportExporter.php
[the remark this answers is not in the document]
joshuatf: Updated those trait methods to `get_default_scheduler_actions()` in https://github.com/woocommerce/woocommerce-admin/pull/3285/commits/697aed6d461fe221a11d3bc22c9d3648d360692a

timmyc:
@psealock or @jeffstieler could either of you give this another 👀

src/ReportExporter.php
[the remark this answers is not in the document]
jeffstieler: > The method get_scheduler_actions() returns the default actions for a scheduler but any scheduler can opt to not use these and/or add their own.

Are the default actions not required? Perhaps I'm just not following correctly. If none of the schedulers using the trait merged the default actions in with theirs, would actions and batches still get queued?

If the "default" scheduler actions are always required (for scheduling actions and batches), the interface is brittle if it depends on classes using the trait to preserve them.

src/Schedulers/SchedulerTraits.php
```diff
@@ -223,13 +226,8 @@ public static function get_next_blocking_job( $action_name ) {
 				$blocking_job_hook = $blocking_job->get_hook();
 				$next_job_schedule = $blocking_job->get_schedule()->next();
 
-				// Eliminate the false positive scenario where the blocking job is
-				// actually another queued dependent action awaiting the same prerequisite.
-				// Also, ensure that the next schedule is a DateTime (it can be null).
-				if (
-					is_a( $next_job_schedule, 'DateTime' ) &&
-					( static::get_action( 'schedule_action' ) !== $blocking_job_hook )
-				) {
```
jeffstieler: Is the false positive scenario no longer possible?

jeffstieler:
Fix for the report download link email works 👍. Just some other comments.

src/ReportExporter.php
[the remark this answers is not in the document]
joshuatf: I see what you mean.

The default actions are not required, but it's likely that a class implementing this trait will use them more often than not.  I think we should allow the option to not use the default actions as other methods within the trait are useful without the batch and scheduler actions this trait provides.

If you think that it would be better to create this as more of an "opt-out" approach for these actions, I can update.

src/Schedulers/SchedulerTraits.php
jeffstieler: Nevermind - I see. It isn't possible because the dependent action is no longer including its dependency in its own arguments. Right?

joshuatf: If I understood the false positive scenario correctly, where a queue search would potentially return the dependency check action because it contained the name of the dependent action, then this isn't possible since we no longer store the name of the dependent action as an argument for the action.

For example, if previously the `dependent_action` action had `import_customers_batch_init` as an argument to define the dependency, this would return a false positive for the `import_customers_batch_init` action existing.  Does that sound right?

src/Schedulers/ImportScheduler.php
```diff
@@ -0,0 +1,194 @@
...
+
+	/**
+	 * Get items based on query and return IDs along with total available.
+	 *
+	 * @param int      $limit Number of records to retrieve.
+	 * @param int      $page  Page number.
+	 * @param int|bool $days Number of days prior to current date to limit search results.
+	 * @param bool     $skip_existing Skip already imported items.
+	 */
+	public static function get_items( $limit, $page, $days, $skip_existing ) {
```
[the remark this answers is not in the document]
joshuatf: Good call 👍

src/Schedulers/SchedulerTraits.php
joshuatf: >Nevermind - I see. It isn't possible because the dependent action is no longer including its dependency in its own arguments. Right?

Didn't refresh before adding my comment- yep, that's correct.

src/Schedulers/OrdersScheduler.php
```diff
@@ -0,0 +1,202 @@
...
+		if ( 'shop_order' !== $type && 'shop_order_refund' !== $type ) {
+			return;
+		}
+
+		// If the order has no id or date created, skip sync.
+		if ( ! $order->get_id() || ! $order->get_date_created() ) {
+			return;
+		}
+
+		$result = array_sum(
```
mikejolley: This block of code is pretty creative :) 

// The update methods return -1 for skip, or a boolean success indicator.

If this is true, the array_sum may be summing booleans. I believe newer versions of PHP throw warnings when types mismatch in math function. You could cast these to `(int)` to prevent that happening.

Or refactor to make clearer, along these lines:

```php
$results = [
     OrdersStatsDataStore::sync_order( $order_id ),
     ProductsDataStore::sync_order_products( $order_id ),
     CouponsDataStore::sync_order_coupons( $order_id ),
     TaxesDataStore::sync_order_taxes( $order_id ),
];

ReportsCache::invalidate();

if ( count( array_intersect( $results, [ -1, false ] ) ) ) {
	self::schedule_action( 'import', array( $order_id ) );
}
```

joshuatf: Nice catch, Mike!  All creativity points go to @jeffstieler for working up this solution.

Updated this in https://github.com/woocommerce/woocommerce-admin/pull/3285/commits/e20f159a770a7a28a391e257f91a07291392aee8

src/ReportExporter.php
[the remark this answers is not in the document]
joshuatf: Updated this in https://github.com/woocommerce/woocommerce-admin/pull/3285/commits/7b07a4430a7c265c456da3f03c2bff27f58fda1a - let me know what you think.

joshuatf:
@jeffstieler Updated per your feedback.  Also handled some feedback left from, @mikejolley.

This PR has been rebased to clean up some conflicts and rename some hooks as well.

src/Schedulers/OrdersScheduler.php
jeffstieler: Nice - checking for `-1` or `false` is much easier to understand and less error prone. 👍

src/ReportsSync.php
```diff
@@ -9,197 +9,40 @@
...
 	/**
 	 * Hook in sync methods.
 	 */
 	public static function init() {
-		// Initialize syncing hooks.
-		add_action( 'wp_loaded', array( __CLASS__, 'customers_lookup_update_init' ) );
-		add_action( 'wp_loaded', array( __CLASS__, 'orders_lookup_update_init' ) );
+		// Initialize scheduler hooks.
+		foreach ( self::get_schedulers() as $scheduler ) {
+			$scheduler::init();
```
jeffstieler: Since there's a filter inside `get_schedulers()`, should we check to make sure `init` is callable here?

src/ReportsSync.php
```diff
@@ -208,669 +51,116 @@ public static function get_import_totals( $days, $skip_existing ) {
...
-			if (
-				( self::SINGLE_ORDER_IMPORT_ACTION === $existing_job->get_hook() ) ||
-				(
-					self::QUEUE_DEPENDENT_ACTION === $existing_job->get_hook() &&
-					in_array( self::SINGLE_ORDER_IMPORT_ACTION, $existing_job->get_args(), true )
-				)
-			) {
-				return;
+		foreach ( self::get_schedulers() as $scheduler ) {
+			if ( $scheduler::is_importing() ) {
```
jeffstieler: Same question here about checking if `is_importing` is callable (re: filter). Optionally checking to see if the trait is being used if you think that's more robust (and possible).

src/ReportsSync.php
```diff
@@ -208,669 +51,116 @@ public static function get_import_totals( $days, $skip_existing ) {
...
-				$next_job_schedule->getTimestamp() + 5,
-				self::QUEUE_DEPENDENT_ACTION,
-				array( $action, $action_args, $prerequisite_action ),
-				self::QUEUE_GROUP
-			);
-		} else {
-			self::queue()->schedule_single( time() + 5, $action, $action_args, self::QUEUE_GROUP );
+		self::reset_import_stats( $days, $skip_existing );
+		foreach ( self::get_schedulers() as $scheduler ) {
+			$scheduler::schedule_action( 'import_batch_init', array( $days, $skip_existing ) );
```
jeffstieler: Same comment about checking for callable methods.

src/ReportsSync.php
```diff
@@ -9,197 +9,40 @@
...
 	 * @return array
 	 */
-	public static function get_import_totals( $days, $skip_existing ) {
-		$orders         = self::get_orders( 1, 1, $days, $skip_existing );
-		$customer_roles = apply_filters( 'woocommerce_analytics_import_customer_roles', array( 'customer' ) );
-		$customer_query = self::get_user_ids_for_batch(
-			$days,
-			$skip_existing,
+	public static function get_schedulers() {
+		return apply_filters(
```
jeffstieler: Rather than immediately returning here, this could be a good place to do interface checking as mentioned in other comments.

src/Schedulers/ImportScheduler.php
```diff
@@ -0,0 +1,183 @@
...
+
+	/**
+	 * Delete a batch by passing the count to be deleted to the child delete method.
+	 *
+	 * @return void
+	 */
+	public static function delete_batch() {
+		wc_admin_record_tracks_event( 'delete_import_data_job_start', array( 'type' => static::$name ) );
+
+		$batch_size = static::get_batch_size( static::get_batch_size( 'delete' ) );
```
jeffstieler: Should there be two calls to `get_batch_size()` here?

jeffstieler:
I'm noticing some oddities when running the historical import.

The imported from date is returning `-1` on a small incremental update:

Screen Shot 2019-12-09 at 9 39 33 AM

The orders and refunds counts are off - the number reported complete is much lower than the total in the UI, but I'm seeing _almost_ that total when I look at the database - but off by 2. I've replicated this on two different test stores.

For example:

Screen Shot 2019-12-09 at 11 01 00 AM

This is "done" - there are no more pending actions to run. `19,932` rows were created in the order stats table.

Again, on this store:

Screen Shot 2019-12-09 at 10 36 57 AM

`338` orders in the table.

src/Schedulers/ImportScheduler.php
joshuatf: Fixed.

src/ReportsSync.php
joshuatf: I've added a check to make sure the parent class is `ImportScheduler` for all schedulers added here.  That will ensure we have all of the methods called that you mentioned in other comments without needing to check for their existence every time.

joshuatf:
>The imported from date is returning -1 on a small incremental update:

That's the correct response, but had to update the client-side to check for the number `-1` instead of a string.

>The orders and refunds counts are off - the number reported complete is much lower than the total in the UI, but I'm seeing almost that total when I look at the database - but off by 2. I've replicated this on two different test stores.

I'm not able to recreate this.  Can you provide steps or a dump of your database for one of these stores so I can attempt to debug?

How does this compare on master?  Are the actual imported numbers also off by 2?

src/ReportsSync.php
joshuatf: Resolved via https://github.com/woocommerce/woocommerce-admin/pull/3285/commits/e7fd4e813f7c3618fd07d33243930e801c008a9e

src/ReportsSync.php
joshuatf: Resolved via https://github.com/woocommerce/woocommerce-admin/pull/3285/commits/e7fd4e813f7c3618fd07d33243930e801c008a9e

src/ReportsSync.php
joshuatf: Resolved via https://github.com/woocommerce/woocommerce-admin/pull/3285/commits/e7fd4e813f7c3618fd07d33243930e801c008a9e

jeffstieler:
> How does this compare on master? Are the actual imported numbers also off by 2?

Looks like I'm also off by `2` on `master`, but the UI doesn't hang. Perhaps that's the `-1` fix. I'll retest.

jeffstieler:
Seems like the `-1` change didn't fix the UI stopping prematurely. The progress bars are not completing when using this branch. I'm on the built in version of Action Scheduler. The best reliable dataset to use (to potentially reproduce) is the WCCOM dataset available on the FG.

jeffstieler:
Here are the responses (unique, in order, duplicates pruned) from the import status endpoint, after "deleting previous data" and re-running the import for all data:

```javascript
{"is_importing":false}

{"customers":{"imported":0,"total":139},"orders":{"imported":0,"total":340},"imported_from":-1,"is_importing":true}

{"customers":{"imported":0,"total":139},"orders":{"imported":0,"total":340},"imported_from":-1,"is_importing":false}

{"customers":{"imported":139,"total":139},"orders":{"imported":125,"total":340},"imported_from":-1,"is_importing":false}

{"customers":{"imported":139,"total":139},"orders":{"imported":300,"total":340},"imported_from":-1,"is_importing":false}
```

This was running on a local store (not WCCOM - for time's sake) running AS 3.0. The UI oddities are consistent with what I was seeing on the WCCOM dataset as well, so I chose the smaller store for this test.

Here's the status responses from `master`:

```javascript
{"is_importing":true,"customers_total":139,"customers_count":0,"orders_total":340,"orders_count":0,"imported_from":"-1"}

{"is_importing":false,"customers_total":139,"customers_count":0,"orders_total":340,"orders_count":0,"imported_from":"-1"}

{"is_importing":true,"customers_total":139,"customers_count":139,"orders_total":340,"orders_count":0,"imported_from":"-1"}

{"is_importing":true,"customers_total":139,"customers_count":139,"orders_total":340,"orders_count":50,"imported_from":"-1"}

{"is_importing":true,"customers_total":139,"customers_count":139,"orders_total":340,"orders_count":100,"imported_from":"-1"}

{"is_importing":true,"customers_total":139,"customers_count":139,"orders_total":340,"orders_count":160,"imported_from":"-1"}

{"is_importing":true,"customers_total":139,"customers_count":139,"orders_total":340,"orders_count":220,"imported_from":"-1"}

{"is_importing":false,"customers_total":139,"customers_count":139,"orders_total":340,"orders_count":280,"imported_from":"-1"}

{"is_importing":false,"customers_total":139,"customers_count":139,"orders_total":340,"orders_count":340,"imported_from":"-1"}
```

joshuatf:
@jeffstieler I attempted to import the wccom data several times and was able to import each time.  The imports completed, showing the completed message and showing the correct stats in the database:

Screen Shot 2019-12-18 at 7 21 07 PM
Screen Shot 2019-12-18 at 7 20 32 PM

I'm wondering if the call to `/status` is hanging on your end.  Are calls still being made to `/status` under your network requests or have they stopped?

jeffstieler:
> I attempted to import the wccom data several times and was able to import each time. The imports completed, showing the completed message and showing the correct stats in the database:

Hmm.. there might be something amiss with my local environment. I'm trying to set up the Docker-based WCCOM to re-test. (If you're using Docker.. please share any tips on how to get test data imported into the container)

> I'm wondering if the call to `/status` is hanging on your end. Are calls still being made to `/status` under your network requests or have they stopped?

The last test run I did continued to make `/status` calls, and ended like this:

Screen Shot 2019-12-18 at 9 35 27 AM

joshuatf:
Tested on a flywheel environment and a docker environment, both using the wccom dump file.  Both environments managed to complete, importing all orders and customers.

Maybe we can get someone else on @woocommerce/wc-admin to test this out as well and see if they're able to reproduce the incorrect import stats.

jeffstieler:
> Maybe we can get someone else on @woocommerce/wc-admin to test this out as well and see if they're able to reproduce the incorrect import stats.

Please! I am totally stuck on this.

I was able to get the Docker-based WCCOM environment up and running, where I imported the WCCOM DB export from the FG. That stalls here:

Screen Shot 2019-12-20 at 9 06 30 AM

I see `13558` customer rows, and `15245 ` order rows in the database.

rrennick:
> That stalls here:

I tried to repeat this. My screen did stall at a different location but all of the orders were imported

Screen Shot 2019-12-30 at 4 02 48 PM

Screen Shot 2019-12-30 at 4 03 05 PM

Then I deleted the imported data, switched to `master` branch and repeated the import. The import was successful and the import screen stalled at a different location. In early December I imported this data into a non-wccom install and there were no issues with the import screen stalling. There is likely something in the custom code on wccom that is interfering with the import status check api calls.

rrennick:
I created #3496 which may account for the results @jeffstieler had above.

I imported the wccom DB into a regular WP install & the import ran to completion except for the progress bar issue noted above. I think this one is good to go.

joshuatf:
Thanks for taking the time to review this rather large PR and outlining that issue, @rrennick!

joshuatf:
@jeffstieler Can you give this a 👍 if everything else looks okay and we can tackle the progress bar issues in https://github.com/woocommerce/woocommerce-admin/issues/3496?

PR #3290, opened 2019-11-21
make customer order count consistent between customer and stats

### LINKED ISSUE #2989: customer order count inconsistent
**Describe the bug**
While working on #2961 I noticed that the user could see a discrepancy in a customer's order stats.

https://github.com/woocommerce/woocommerce-admin/blob/master/src/API/Reports/Customers/DataStore.php#L66
```
		'orders_count'     => 'SUM( CASE WHEN parent_id = 0 THEN 1 ELSE 0 END ) as orders_count',
		'total_spend'      => 'SUM( gross_total ) as total_spend',
		'avg_order_value'  => '( SUM( gross_total ) / COUNT( order_id ) ) as avg_order_value',
```

Average order value is calculated from the total order count, while the order count shown to the user is the count of orders without a parent.

**To Reproduce**

1. Search the order stats for a user with order(s) with parents
2. Go to Analytics  -> Customers
3. Search for that user
4. Numbers will not math up per screenshot

**Expected behavior**
Order count should be consistent per customer.

**Screenshots**
Screen Shot 2019-10-02 at 3 13 21 PM

Note: `Canadas` is my test to see that translations are working :)

rrennick:
Fixes #2989

This PR updates the customer average order total to be calculated from the orders total shown divided by the number of orders shown. It updates the table summary to use the same calculations.

Also contains a number of PHPCS fixes.
### Screenshots

Screen Shot 2019-11-21 at 1 26 03 PM

1. Search the order stats for a user with order(s) with parents
2. Go to Analytics  -> Customers
3. Search for that user
4. Numbers will now total up consistently per screenshot

### Changelog Note:

Fix: Make the order count between customer and table total consistent.


rrennick:
@jeffstieler Could you review this one as it makes changes to the calculations you worked out with @LevinMedia ?

src/API/Reports/Customers/DataStore.php
```diff
@@ -75,7 +75,7 @@ protected function assign_report_columns() {
 			'date_last_order'  => "MAX( {$wpdb->prefix}wc_order_stats.date_created ) as date_last_order",
 			'orders_count'     => 'SUM( CASE WHEN parent_id = 0 THEN 1 ELSE 0 END ) as orders_count',
 			'total_spend'      => 'SUM( total_sales ) as total_spend',
-			'avg_order_value'  => '( SUM( total_sales ) / COUNT( order_id ) ) as avg_order_value',
+			'avg_order_value'  => 'CASE WHEN SUM( CASE WHEN parent_id = 0 THEN 1 ELSE 0 END ) = 0 THEN NULL ELSE SUM( total_sales ) / SUM( CASE WHEN parent_id = 0 THEN 1 ELSE 0 END ) END AS avg_order_value',
```
jeffstieler: What do you think about putting the `CASE` statement for `orders_count` into a variable that can be reused here?

src/API/Reports/Customers/Stats/DataStore.php
```diff
@@ -100,20 +100,21 @@ public function get_data( $query_args ) {
 			$this->subquery->add_sql_clause( 'select', 'SUM( total_sales ) AS total_spend,' );
 			$this->subquery->add_sql_clause(
 				'select',
-				'CASE WHEN COUNT( order_id ) = 0 THEN NULL ELSE COUNT( order_id ) END AS orders_count,'
+				'SUM( CASE WHEN parent_id = 0 THEN 1 END ) as orders_count,'
```
jeffstieler: Same question about putting this in a variable for reuse.

src/API/Reports/Customers/DataStore.php
rrennick: > What do you think about putting the CASE statement for orders_count into a variable that can be reused here?

I would prefer to do that. I wasn't sure if `orders_count` could be used in other summary calculations in all the supported versions of MySQL. eg.

```
CASE WHEN orders_count <= 0 THEN NULL ELSE SUM( total_sales ) / orders_count END AS avg_order_value
```

jeffstieler: I mean a PHP variable, like in the [gross sales calculation](https://github.com/woocommerce/woocommerce-admin/blob/master/src/API/Reports/Orders/Stats/DataStore.php#L80-L86).

rrennick:
> I mean a PHP variable, like in the gross sales calculation.

@jeffstieler Thanks, that's a great suggestion. I did that in https://github.com/woocommerce/woocommerce-admin/pull/3290/commits/412e484289e8a3adbe44dd47a3fef8bfd1835a56 . I did the same with `total_spend` as it was also used twice.

Issue #3120, opened 2019-10-28
Gutenberg SASS Variables: Adopt Base Styles package


psealock:
https://github.com/WordPress/gutenberg/pull/17883 Creates a Base Styles package to expose SASS variables. Until now, we have had to copy/paste these into our repo so that postCSS processing could apply Woo themes to Gutenberg SASS files.

https://github.com/woocommerce/woocommerce-admin/blob/b15dfac78a871370d0bd61f06b2160f9a5d0e14a/client/stylesheets/abstracts/_variables.scss#L25

Now that those variables are exposed, we should use them as an import.

simison:
You could prolly replace those `$muriel-` variables with `$studio-` variables from https://github.com/Automattic/color-studio, too.

PR #3292, opened 2019-11-21
Component styles: Explicitly import from Gutenberg

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/3120

https://github.com/woocommerce/woocommerce-admin/pull/3253 removed enqueued styles for Gutenberg components since we were bundling those components anyways.

This PR adds those styles back explicitly. They are also run through our post-css process so variables needed to be added. Since the [Base Styles](https://github.com/WordPress/gutenberg/tree/master/packages/base-styles) package is out, we can simply use import those variables and remove the ones we copy/pasted.

1. Use the app and ensure no styles look off.

### DIFF
--- a/client/stylesheets/abstracts/_variables.scss
+++ b/client/stylesheets/abstracts/_variables.scss
@@ -1,12 +1,5 @@
 /** @format */
 
-@import 'node_modules/@wordpress/base-styles/colors';
-@import 'node_modules/@wordpress/base-styles/variables';
-@import 'node_modules/@wordpress/base-styles/mixins';
-@import 'node_modules/@wordpress/base-styles/breakpoints';
-@import 'node_modules/@wordpress/base-styles/animations';
-@import 'node_modules/@wordpress/base-styles/z-index';
-
 $fallback-gutter: 24px;
 $fallback-gutter-large: 40px;
 $gutter: var(--main-gap);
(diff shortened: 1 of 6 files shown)


client/stylesheets/abstracts/_variables.scss
```diff
@@ -1,5 +1,12 @@
 /** @format */
 
+@import 'node_modules/@wordpress/base-styles/colors';
+@import 'node_modules/@wordpress/base-styles/variables';
+@import 'node_modules/@wordpress/base-styles/mixins';
+@import 'node_modules/@wordpress/base-styles/breakpoints';
+@import 'node_modules/@wordpress/base-styles/animations';
+@import 'node_modules/@wordpress/base-styles/z-index';
```
jeffstieler: Can we alias this in the webpack config like `gutenberg-components`?

psealock:
>Can we alias this in the webpack config like gutenberg-components?

@jeffstieler: Because these variables need to be loaded before our own `_variables.scss`, which is loaded by sass loader, the alias isn't available like the one for `gutenberg-components`. Instead I added it to the loader itself in https://github.com/woocommerce/woocommerce-admin/pull/3292/commits/464bd61388e4cdd71210c0d8a27d46fc2e6764b4.

Since there are name conflicts, I can't add the path to `includedPaths`. How does this look?

psealock:
I'm going to merge this so the experience isn't completely broken, but happy to address feedback in a follow up PR

PR #3294, opened 2019-11-22
Decouple Navigation from `wcSettings`.

### LINKED ISSUE #3031: Refactor: Decouple wcSettings (via @woocommerce/wc-admin-settings) from packages/navigation
## Context:

- #2999 is the master ticket.

### `packages/navigation`

This package has the `adminUrl` setting coupled to the exported `getAdminLink` function.

I don't think `getAdminLink` should be an export of this package but instead should be something defined/consumed by the client (yes this means extensions would have to define their own function).  However, `getAdminLink` could be defined on the `@woocommerce/settings` (currently exported in the blocks repo) external (so accessible via `wc.wcSettings.getAdminLink` when `wc-settings` is a handle on an enqueued script).  

Alternatively, what we could do is have a `SiteSettings` value object that is exposed on a package that is instantiated by a configuration object. It could expose an interface for methods/properties that are populated from the configuration object on construct and then clients could take care of instantiating and consuming that instantiated value object where needed.  I'd be wary about making this a "catch-all" for all settings though.

jeffstieler:
Fixes #3031 

Removes the `wcSettings` coupling from Navigation package by moving `getAdminLink()` to `wc-admin-settings`.

### `Link`

In turn, this also decouples the `<Link />` component from the Navigation package. All usage of `Link` for "wp-admin" type links had to be refactored to call `getAdminLink()` from the client app.

- Test all links rendered by the WC-Admin app

### Changelog Note:

Tweak: remove global settings dependency from Navigation package.

### DIFF
--- a/client/analytics/report/customers/table.js
+++ b/client/analytics/report/customers/table.js
@@ -13,7 +13,7 @@ import { defaultTableDateFormat } from '@woocommerce/date';
 import { formatCurrency, getCurrencyFormatDecimal } from 'lib/currency-format';
 import { Date, Link } from '@woocommerce/components';
 import { formatValue } from 'lib/number-format';
-import { getSetting } from '@woocommerce/wc-admin-settings';
+import { getAdminLink, getSetting } from '@woocommerce/wc-admin-settings';
 
 const { countries } = getSetting( 'dataEndpoints', { countries: {} } );
 
@@ -128,7 +128,7 @@ export default class CustomersReportTable extends Component {
 			const countryName = this.getCountryName( country );
 
 			const customerNameLink = user_id ? (
-				<Link href={ 'user-edit.php?user_id=' + user_id } type="wp-admin">
+				<Link href={ getAdminLink( 'user-edit.php?user_id=' + user_id ) } type="wp-admin">
 					{ name }
 				</Link>
 			) : (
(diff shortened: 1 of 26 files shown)


client/analytics/report/customers/table.js
```diff
@@ -128,7 +128,7 @@ export default class CustomersReportTable extends Component {
 			const countryName = this.getCountryName( country );
 
 			const customerNameLink = user_id ? (
-				<Link href={ 'user-edit.php?user_id=' + user_id } type="wp-admin">
+				<Link href={ getAdminLink( 'user-edit.php?user_id=' + user_id ) } type="wp-admin">
```
nerrad: could also be written as (similar to how it's being done elsewhere):

```js
getAdminLink( `user-edit.php?user_id=${ user_id }` )
```

(I think it reads better but totally up to you)

jeffstieler:
> How does that sound?

Sounds good - so we'd want to land the changes to `woocommerce-gutenberg-product-blocks` before modifying and landing this PR, correct?

nerrad:
I think you can land this pr, but just modify it to check if `getAdminLink` is available on the `wcSettings` first before falling back to the shim (which makes it work with whatever lands in the blocks repo).  Once wc-admin is merged into WC core, that shim can be removed (so you can create a followup issue for that which can be a post-merge task?).  

Are you okay with taking care of the pull in `woocommerce-blocks`?

jeffstieler:
> modify it to check if `getAdminLink` is available on the `wcSettings` first before falling back to the shim

Done in effd743.

> Are you okay with taking care of the pull in woocommerce-blocks?

Yep - I'll work on that now.

PR #3295, opened 2019-11-22
Update onboarding filters with a consistent prefix

### LINKED ISSUE #3288: Make onboarding filters consistent
While working on documentation in #3286, it was noted how inconsistent naming patterns for filters across wc-admin are. Here are a few styles we've used just for onboarding alone:

`woocommerce_onboarding_`
`woocommerce_admin_onboarding_`
`rest_onboarding_`

Since the onboarding code is still fresh and in development, I think we could rename these without issues around backwards compatibility.

justinshreve:
Fixes #3288.

This PR updates the filter names used in onboarding to be more consistent. Per the short conversation in Slack ( p1574361629414700-slack-wc-admin ), General filters are prefixed with `woocommerce_admin_` and REST endpoints with `woocommerce_rest_`

* Boot up the app and make sure everything looks OK.
* Bonus: Run the [add-task example extension](https://github.com/woocommerce/woocommerce-admin/tree/master/docs/examples/extensions) and make sure it works.

### Changelog Note:

Dev: Update onboarding filters with a consistent prefix

### DIFF
--- a/client/dashboard/customizable.js
+++ b/client/dashboard/customizable.js
@@ -255,7 +255,7 @@ export default compose(
 			const profileItems = getProfileItems();
 			const tasks = getTasks( {
 				profileItems,
-				options: getOptions( [ 'woocommerce_onboarding_payments' ] ),
+				options: getOptions( [ 'woocommerce_task_list_payments' ] ),
 				query: props.query,
 			} );
 			const visibleTasks = filter( tasks, task => task.visible );
--- a/client/dashboard/task-list/index.js
+++ b/client/dashboard/task-list/index.js
@@ -276,7 +276,7 @@ export default compose(
 
 		const tasks = getTasks( {
 			profileItems,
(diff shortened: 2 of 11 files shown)


rrennick:
> Should we create a follow-up to also handle updating other filter prefixes around wc admin before v1?

I think we should.

justinshreve:
> Do we also want to update option names? (e.g., woocommerce_onboarding_payments).

I don't think they necessarily need the `woocommerce_admin_` prefix to match the filters, but it might be good to rename `woocommerce_onboarding_payments` to `woocommerce_task_list_payments` to match the other `task_list` related options.

> Should we create a follow-up to also handle updating other filter prefixes around wc admin before v1?

I'll log an issue.

justinshreve:
https://github.com/woocommerce/woocommerce-admin/issues/3301

PR #3313, opened 2019-11-27
Settings: wp.data feature branch

### LINKED ISSUE #3661: Settings Refactor: Split wcSettings and use wp.data
## The Problem

Individual bits of wcSettings data are unrelated and rely on different endpoints for manipulation, each with their own requirements for storage and persistence.

More info in p90Yrv-1uN-p2

Attempts to add wp.data stores for settings in #3313 and #3581 have made great progress, but until the core problem described above is resolved, those efforts will stall. 

## The Plan

This is a parent issue to track progress.

- [ ] Normalise mutable data sources such that they follow the same shape.

The following tasks will need to be accomplished for `currentUserData`, `wcAdminSettings`, `onboarding`, and `options`.

* Remove setSetting from the API so that its clear these are read-only properties.
* Remove the ability to access each peice from being accessed by getSetting. 
* Swap out fresh-data for the new implementation.

- [x] wcAdminSettings https://github.com/woocommerce/woocommerce-admin/pull/3675
- [ ] currentUserData
- [ ] onboarding
- [ ] options

nerrad:
EDIT: This branch is now a feature branch for the new settings data store using wp.data. The original post has been deleted because its not relevant anymore.

This feature branch includes changes from the following pull requests.

* https://github.com/woocommerce/woocommerce-admin/pull/3707
* https://github.com/woocommerce/woocommerce-admin/pull/3675
* https://github.com/woocommerce/woocommerce-admin/pull/3430

Partially fixes https://github.com/woocommerce/woocommerce-admin/issues/3661

### Summary

* wc-api/settings is replaced by packages/data/src/settings.
* Properties in wcSettings that are mutable will be incrementally added to data stores for consumption and mutation.
* Non mutable settings, such as currency, will be accessed as they are currently using `@woocommerce/wc-admin-settings`.
* New hooks exist for initialisation and usage of settings data.

### Testing

Smoke test the app. Most affected areas have been the settings page and the OBW.


psealock:
Noting here that I pushed changes to setup `packages/data` and moved relevant files to the package.

client/analytics/settings/index.js
```diff
@@ -206,22 +206,62 @@ class Settings extends Component {
...
+					! hasFinishedResolution( 'getSettings', [ 'wc_admin' ] )
+				) {
+					startResolution( 'getSettings', [ 'wc_admin' ] );
+					hydrateSettings( settings.current );
+					finishResolution( 'getSettings', [ 'wc_admin' ] );
+				}
+			}, [] );
+			return <WrappedComponent { ...props } />;
+		},
+		'withHydration'
```
psealock: @nerrad is the idea here that `withHydration` will be an exportable HOC? So any component looking to interact with settings could use this?

nerrad: It'd be something primarily wc-admin would use to hydrate it's stores from the server.  So essentially you'd use the hoc to wrap the App entry point so that store hydration is automatically taken care of (and anything else using the store in that view would be interacting with a already hydrated store assuming they are listing wc-admin stores as a dependency).

It feasibly _could_ be abstracted so that it could be exported as a utility other client code could use to hydrate store values but I'd start simple and keep it for internal wc-admin use initially (and thus just hydrating client).  You can see the blocks project usage:  [HOC here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/hocs/with-rest-api-hydration.js) and implementation [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/blocks/products/all-products/frontend.js)

psealock:
For the issues noted, I created follow up issues so they can be tackled separately and updated the original PR description.

https://github.com/woocommerce/woocommerce-admin/issues/3405
woocommerce/woocommerce#32284

client/analytics/settings/index.js
psealock: Also, noting this would be where we interchange `useSelect` with `useSelectWithRefresh`.

I think we can leave this aspect up to the code that handles https://github.com/woocommerce/woocommerce-admin/issues/3405 because this will be likely to change as a result.

psealock: `preloadSettings` is an onboarding thing, which seems to be gathered from data written to `wcSettings.preloadSettings`. So, we can definitely remove from here and update the OBW to also use this new data package. I'm pretty sure we can remove the `getSettings` resolver because it doesn't appear to be used, on master as well.

In fact `store.getSettings( 'wc_admin' )` can be hard coded to an empty object and the page works normally. This has me thinking 🤔, whats the point of HOC `withSettings` anyways? All settings are gathered from `@woocommerce/wc-admin-settings` instead of the `settings` prop. 

The `persistAllSettings` action is useful as it can make a request to update settings on on the server as well as the object on the window via `setSetting`. 

@nerrad how do you feel about moving all the logic currently in `client/settings` to this store? All areas of the app that need to interact with settings can do so via the `withSettings` HOC. 

I think a good course of action will be to remove `withHydration` from this PR, re-add fresh-data settings api (when I removed it, OBW broke), and ready this for final review while making the above changes as part of https://github.com/woocommerce/woocommerce-admin/issues/3405

nerrad: > I'm pretty sure we can remove the getSettings resolver because it doesn't appear to be used, on master as well.

You can remove this if you want, but it's handy having it for if this store get's used by third party extensions and they don't pre-hydrate settings for their view (or it's an environment where pre-hydrating isn't as easily done).

> This has me thinking 🤔, whats the point of HOC withSettings anyways? All settings are gathered from @woocommerce/wc-admin-settings instead of the settings prop.

This is why my comments about the number of settings states and #3405 exists.  In my opinion, for the settings managed by this store `@woocommerce/wc-admin-settings` should ONLY be used for hydrating the initial store state and then from that point on the store should serve as the authority.  When you use the store to persist setting changes to the server, the store state update will automatically signal to any components subscribed to the store (via `withSelect` or `useSelect` mapped to the store selectors) to update.  `@woocommerce/wc-admin-settings` is a poor mechanism for communicating these changes in state.

> The persistAllSettings action is useful as it can make a request to update settings on on the server as well as the object on the window via setSetting.

If you're talking about updating the window object from `persistAllSettings` I don't think that's a good idea.  You'd be coupling window globals to the data store.   In fact, I don't think this window object for settings _should_ be updated at all.  Related to previous conversations I've had about this, I think the only purpose of for the window global is to hydrate the initial settings state from the server (to avoid additional requests).

> @nerrad how do you feel about moving all the logic currently in client/settings to this store? All areas of the app that need to interact with settings can do so via the withSettings HOC.

This seems to be in line with what I'm advocating for. As long as you're not talking about also moving interactions with the window global to the store too.

> I think a good course of action will be to remove withHydration from this PR, re-add fresh-data settings api (when I removed it, OBW broke), and ready this for final review while making the above changes as part of #3405

- re `withHydration` removal... you'll have to make the call on that.  I'm guessing though that preLoad settings would still have a value during a specific state of onboarding?  I was never able to actually see data in that object coming from the server but maybe that was just an artifact of how I was testing.  
- re-add fresh-data settings api:  yeah, should only remove it once you've confirmed anything using it has been decoupled from it and is using the new store.
- re: making above changes part of #3405.  👍

psealock: Thanks for the feedback @nerrad. 

>In my opinion, for the settings managed by this store @woocommerce/wc-admin-settings should ONLY be used for hydrating the initial store state and then from that point on the store should serve as the authority

Absolutely, that makes sense to me now.

>re withHydration removal... you'll have to make the call on that. I'm guessing though that preLoad settings would still have a value during a specific state of onboarding? I was never able to actually see data in that object coming from the server but maybe that was just an artifact of how I was testing

`preloadSettings` is treated just like any other subset of settings. Its loaded onto the settings object:

https://github.com/woocommerce/woocommerce-admin/blob/0860b3b2b3bef9d91906e02cd6ac6fe01a1f1dd2/src/Loader.php#L638

Then retrieved from `wcSettings` like other properties. I'm going to remove it from the settings page, and if I'm thinking about it incorrectly, circle back to `withHydration`.

psealock:
### Update

After some discussion on how to proceed, this one is ready for review. The best way to test this is to use the settings page and ensure no regressions have occurred.

The follow up pull request to make wp.data settings store the source of truth is https://github.com/woocommerce/woocommerce-admin/pull/3430

Edit: @nerrad alternatively, we can continue the conversation in woocommerce/woocommerce-admin#3430 as it progresses and merge that one once we're happy. This way we can reserve final testing of the settings page for this PR.

psealock:
Now that woocommerce/woocommerce-admin#3430 is merged, I think we should treat this PR as a feature branch for the settings migration. Although we could merge this as is, having portions of the app using different sources for settings data seems likely to introduce bugs.

Next steps can be incrementally moving over to usage of the new data store.

timmyc (approved):
Ran through a full smoke test of the entire onboarding and setup checklist progress. Including many of the more involved flows like purchasing and installing an extension from the setup checklist, and configuring TaxJar via WCS and a Jetpack connection.

Viewed each report page ( did not work in-depth with filters ), and manipulated analytics settings. All is working well for me.

Test conducted on an atomic ephemeral site.

timmyc:
@psealock looking good for me, would be nice to get one more person to test this out then get it merged. ( @jeffstieler or @rrennick maybe? )

PR #3316, opened 2019-11-28
Onboarding: Format product count ranges numbers without decimals

### LINKED ISSUE #3312: Onboarding: Product count dropdown has decimal places
**Describe the bug**
On the Business Details step of the profiler, the product count dropdown has decimals.

**To Reproduce**

1. Go to the Business Details step of the profiler.
2. Note the decimals in the dropdown.

**Expected behavior**
Expected integers and no decimals.

**Screenshots**
Screen Shot 2019-11-27 at 9 23 24 PM

joshuatf:
Fixes #3312

Fixes a change to the product count formatting since `numberFormat()` now includes precision by default.

### Before
Screen Shot 2019-11-27 at 9 23 24 PM

### After
Screen Shot 2019-11-28 at 2 13 45 PM

1. Open the onboarding profiler.
2. Go to the business details step.
3. Make sure dropdown numbers are formatted as expected.

### DIFF
--- a/client/dashboard/profile-wizard/steps/business-details.js
+++ b/client/dashboard/profile-wizard/steps/business-details.js
@@ -139,18 +139,18 @@ class BusinessDetails extends Component {
 		return keys( pickBy( values ) ).filter( name => this.extensions.includes( name ) );
 	}
 
-	getNumberRangeString( min, max = false, format = formatValue ) {
+	getNumberRangeString( min, max = false ) {
 		if ( ! max ) {
 			return sprintf(
 				_x( '%s+', 'store product count or revenue', 'woocommerce-admin' ),
-				format( 'number', min )
+				formatValue( 'number', min )
 			);
 		}
 
 		return sprintf(
 			_x( '%1$s - %2$s', 'store product count or revenue range', 'woocommerce-admin' ),
-			format( 'number', min ),
-			format( 'number', max )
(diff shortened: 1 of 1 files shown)


client/dashboard/profile-wizard/steps/business-details.js
```diff
@@ -139,18 +139,18 @@ class BusinessDetails extends Component {
 		return keys( pickBy( values ) ).filter( name => this.extensions.includes( name ) );
 	}
 
-	getNumberRangeString( min, max = false, format = numberFormat ) {
+	getNumberRangeString( min, max = false, format = formatValue ) {
 		if ( ! max ) {
 			return sprintf(
 				_x( '%s+', 'store product count or revenue', 'woocommerce-admin' ),
-				format( min )
+				format( 'number', min )
```
jeffstieler: If you didn't want to change the method signature here (since that's an API change for this function) you could use `partial` to compose it with the first argument when setting the parameter default. Though.. I don't see any calls to `getNumberRangeString()` that override `format` - perhaps its unnecessary?

joshuatf: Good point.  It doesn't appear to be necessary so I've removed for now (maybe something was removed where it was previously neeeded), but /cc @justinshreve just to double check where this was being used.

PR #3323, opened 2019-11-29
Onboarding: Allow logo to be removed via customize appearance task

### LINKED ISSUE #3322: Onboarding: Custom logo can't be removed via appearance task.
**Describe the bug**
Once set, it's not possible to remove and save a logo in the customize appearance task.  Removing from the customizer then creates a loading issue.  This appears to be a bug I introduced in https://github.com/woocommerce/woocommerce-admin/pull/3304 🤦‍♂ 

**To Reproduce**

1. Add a custom logo via this task.
2. Revisit the task and attempt to remove and "Proceed."
3. Refresh the page and note that removal is not persisted.
4. Remove the logo via the customizer.
5. Go to this task and notice the loading/pending state.

**Expected behavior**
Expected to be able to remove a logo from this task.

**Screenshots**
Screen Shot 2019-11-29 at 10 22 59 AM

joshuatf:
Fixes #3322

Fixes a bug where the logo could not be removed in this task.

### Screenshots
Screen Shot 2019-11-29 at 12 10 50 PM

1. Go to the onboarding task list "Customize Appearance" task.
2. Add a logo and save if it doesn't exist.
3. Revisit the step and remove the logo and click "Proceed."
4. Refresh the page to make sure the logo was updated in theme mods and changes persist.

### DIFF
--- a/client/dashboard/task-list/tasks/appearance.js
+++ b/client/dashboard/task-list/tasks/appearance.js
@@ -172,7 +172,7 @@ class Appearance extends Component {
 	updateLogo() {
 		const { options, themeMods, updateOptions } = this.props;
 		const { logo } = this.state;
-		const updateThemeMods = logo ? { ...themeMods, custom_logo: logo.id } : themeMods;
+		const updateThemeMods = { ...themeMods, custom_logo: logo ? logo.id : null };
 
 		recordEvent( 'tasklist_appearance_upload_logo' );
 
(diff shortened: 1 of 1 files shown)


joshuatf:
>There are a number of API requests on this step and each one took quite a while to resolve. I had to wait > 10 seconds to see the uploaded logo when going to this step. Maybe some room to optimise those requests?

I'm not seeing slow speeds on this personally, but we can definitely prefetch the logo data to make this move a bit faster.

>If you complete the personalisation step (adding a logo) then immediately repeat it without a page refresh first, the selected logo will not be shown.

Nice catch, @mikejolley!  I was going to fix that in here after you noted it, but I think the prefetching above will probably solve this issue anyway instead of the need to invalidate the cached request.

Opened https://github.com/woocommerce/woocommerce-admin/issues/3334 to solve both of those issues.

PR #3324, opened 2019-11-29
Onboarding: Update global currency settings on address change

### LINKED ISSUE #3318: Onboarding: Currency should update to match store address
**Describe the bug**
After selecting a country in the Store Details step of the profiler, the currency in the following steps defaults to WC's default currency of GBP.

**To Reproduce**

1. With a new store, start the profiler.
2. Change the country in the profiler store details step to any country outside the UK.
3. Note that the currency is still in pounds for the Business Details step.

**Expected behavior**
Expected the currency to match the store's default country.

joshuatf:
Fixes #3318

Previously, the store address updated the currency but this only occurred on page refresh.  This PR introduces several changes:
* Add functions to set the wc-admin currency in the current session.
* Updates the currency based on the store address input of the profiler.
* Standardize currency config properties and price formatting.

### Screenshots
Screen Shot 2019-11-29 at 2 45 44 PM

Screen Shot 2019-11-29 at 2 45 55 PM

1. Open the onboarding profiler.
2. Update your store country in the store address step to one of the countries listed in this PR - https://github.com/woocommerce/woocommerce-admin/pull/3110/files
3. Without a hard refresh of the page, navigate to "Business Details."
4. Make sure the revenue dropdown has the correct currency.
5. Make sure the rest of wc-admin currencies continue working as expected.
6.  Update to a currency not listed in `getCurrencyData()`.
7.  Check that the currency works as expected in wc-admin.


client/dashboard/profile-wizard/steps/business-details.js
```diff
@@ -276,38 +277,35 @@ class BusinessDetails extends Component {
 		const revenueOptions = [
 			{
 				key: 'none',
-				label: _x( "$0 (I'm just getting started)", '$0 revenue amount', 'woocommerce-admin' ),
+				label: sprintf(
+					_x( "%s (I'm just getting started)", '$0 revenue amount', 'woocommerce-admin' ),
```
mikejolley: Since this string has placeholders, a `/* Translators...` comment should be added before this line.

client/dashboard/profile-wizard/steps/business-details.js
```diff
@@ -276,38 +277,35 @@ class BusinessDetails extends Component {
...
+					formatCurrency( 0 )
+				),
 			},
 			{
 				key: 'up-to-2500',
-				label: _x( 'Up to $2500 USD', 'equivalent of $2500 USD', 'woocommerce-admin' ),
+				label: sprintf(
+					_x( 'Up to %s', 'Up to a certain revenue amount', 'woocommerce-admin' ),
```
mikejolley: Missing `/* Translators....` I also wonder if `_x` is needed after adding translators comments for this and the other occurrences in this PR. Context is used to distinguish strings with different meaning but the same content - why do these need special context?

client/dashboard/profile-wizard/steps/business-details.js
joshuatf: Added in https://github.com/woocommerce/woocommerce-admin/pull/3324/commits/9a88072b8c8265120be44eab1af7d9b26bb1b064

client/dashboard/profile-wizard/steps/business-details.js
joshuatf: Added in https://github.com/woocommerce/woocommerce-admin/pull/3324/commits/9a88072b8c8265120be44eab1af7d9b26bb1b064.

I think the context may have been a previous iteration when formatting for revenues and product count were handled together.  Removed as it does not appear to be necessary currently.

joshuatf:
>The following error was in the console:
TypeError: countryState.split is not a function index.js:21322:23

I managed to get this error to occur once when moving from a country:state option to a country only option.  But after rebuilding with `npm start/build` I could not recreate the error after 30 attempts.

Can you rebuild and see if you're able to see this still?

>It is slightly confusing that this input shows as a 'dropdown' but doesn't dropdown unless you delete the contents manually.

I tend to agree with this.  If I recall correctly, the previous version of the [material autocomplete](https://material-ui.com/components/autocomplete/) (which this component was loosely based around) had this same behavior, but it seems the new version opens the menu item list every time.

I'll open an issue to address this.

client/dashboard/profile-wizard/steps/business-details.js
```diff
@@ -276,38 +277,37 @@ class BusinessDetails extends Component {
...
+				label: this.getNumberRangeString( 50000, 250000, formatCurrency ),
 			},
 			{
 				key: 'more-than-250000',
-				label: _x(
-					'More than $250,000 USD',
-					'equivalent of more than $250,000 USD',
-					'woocommerce-admin'
+				label: sprintf(
+					_x( 'More than %s', 'More than a certain revenue amount', 'woocommerce-admin' ),
```
mikejolley: This one is still missing translators.

Issue #2345, opened 2019-05-30
Extensibility: Ensure Multi-Currency Plugins can properly extend wc-admin


APlusDesign:
**Describe the bug**

If using WooComm Multi-Currency Plugin, under standard reporting you will see that reports are split. You get 1 report per currency that you have enabled.

In the new WooComm Analytics, you do not get this, instead, it is taking all orders from all currencies and grouping them together under a singular report.

It also does not attempt to convert those other currency amount orders into the main currency, (which would not be ideal), it actually takes the value of an order made in another currency and adds it to the revenue report in the main currency, for example, it will take 1,500 THB and add it into your revenue report as $1,500 USD

**To Reproduce**

Install WooComm multi currency
Create multi currency orders
Import the new data to WooComm analytics
View Analytics -> Revenue report

**Expected behavior**

It should have split reporting, one per currency, in the exact same way that the default WooComm -> Reports has had for a while.

**Desktop (please complete the following information):**

OS Name	Microsoft Windows 10 Home Single Language
Chrome
Version 74.0.3729.169 (Official Build) (64-bit)

**Additional context**

Considering this feature already exists in basic reports, it will be a requirement for the seller to have it in the new Analytics otherwise anyone with WooComm Multi Currency plugin will have incorrect and unusable analytics data.

Aljullu:
Thanks for opening this issue @APlusDesign.

Pointing out another user has reported a similar issue with [WooCommerce Currency Switcher](https://codecanyon.net/item/woocommerce-currency-switcher/8085217) in the support forums:

https://wordpress.org/support/topic/multi-currency-support-16/

timmyc:
Thanks for the feedback @APlusDesign - as you stated in your initial issue:

> Considering this feature already exists in basic reports, it will be a requirement for the seller to have it in the new Analytics otherwise anyone with WooComm Multi Currency plugin will have incorrect and unusable analytics data.

The Multi Currency Plugin is what has extended the legacy reports to make those reports function as expected. Likewise, if that plugin, or other currency plugins want to extend the new WooCommerce Admin reports, they will have to add support to their own code to do so.

While developing WooCommerce Admin, we have added in a variety of ways for extension developers to hook in / modify data in both the REST API / PHP side of things, along with options to manipulate items on the front-end. But ultimately it is up to the extension developers to add support.

We are more than happy to assist extension developers if they encounter issues with building/extending the new reports, but we can't add support for every extension ourselves.

[...]

PR #3328, opened 2019-12-02
Add currency extension

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/2345

Now that the [dataStore refactor](https://github.com/woocommerce/woocommerce-admin/pull/2961) introduced a way to add or modify SQL queries, some complex operation are possible for extensions. This PR aims to illustrate the addition of a UI element to introduce a query argument used to eventually modify SQL queries.

A currency viewer extension is introduced to Reports allowing the user to filter items by currency. Since currency data is part of post meta, an SQL join statement is required.

### Changes to wc-admin

* There were discrepancies between filter names using plural names, ie category v. categories, which were fixed.
* `persistedQueries` allow users to keep certain params in the url when navigating between reports. A filter was added.
* `getQueryFromConfig` was amended to use filter values from added configs.

### Screenshots

Screen Shot 2019-12-02 at 3 09 33 PM

Screen Shot 2019-12-03 at 2 55 38 PM

1. Create two or more orders for today.
1. On the last order, change the postmeta value for key `_order_currency` to `NZD` or `ZAR`. If you are using vvv, you can go to the last page of the `wp_postmeta` table and edit the key value pair directly in phpmyadmin.
1. You may need to clear your cache `wp transient delete woocommerce_reports-transient-version`.
1. `npm run example -- --ext=sql-modification`
2. Activate WooCommerce Admin SQL modification Example
3. Go to any report and change the time to "today"
4. Manipulate the currency dropdown to see the results update accordingly.
5. Check all the reports.

### Follow up steps

* Clean up filter names with respect to pluralization. Also think about the catch-all hook names like `wc_admin_clauses_select` and see if one can be made for each report.
* The extension makes use of `wcSettings` by directly manipulating the global. It should be using `AssetDataRegistry`instead.
* When the currency dropdown is manipulated, currency symbols used in the chart and table should reflect that. I have a few ideas on this, but best to create a solution in the next PR.


docs/examples/extensions/sql-modification/js/index.js
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Add a dropdown to a report.
+ *
+ * @param {array} filters - set of filters in a report.
+ * @return {array} amended set of filters.
+ */
+const addCurrencyFilters = filters => {
+	return [
+		{
+			label: __( 'Currency', 'woocommerce-admin' ),
```
rrennick: Text domain should be `plugin-domain`.

docs/examples/extensions/sql-modification/woocommerce-admin-sql-modification.php
```diff
@@ -0,0 +1,197 @@
...
+	wp_register_script(
+		'sql-modification',
+		plugins_url( '/dist/index.js', __FILE__ ),
+		array(
+			'wp-hooks',
+			'wp-element',
+			'wp-i18n',
+			'wp-plugins',
+			'wc-components',
+			'wc-settings',
```
rrennick: The example script wasn't being enqueued until I commented out this dependency.

PR #3336, opened 2019-12-03
Remove extra debug information for onboarding plugins install

### LINKED ISSUE #3265: Onboarding: Plugins fail on first install attempt
**Describe the bug**
When installing plugins for the first time, a failure usually occurs prompting a retry which then succeeds.

**To Reproduce**

1. On a fresh install, go through the profiler and install Mailchimp/Facebook extensions on the business extensions step.
2. Note the error on first attempt (sometimes?).
3. Note the network error.

**Expected behavior**
Expected plugins to succeed on install and activate the first time.

**Screenshots**
Screen Shot 2019-11-18 at 10 08 01 AM

justinshreve:
Closes #3265.

As we haven't been able to reproduce this failing outside of fresh Jurassic Ninja sites, and the debug info we did gather shows the error occuring in WP core functionality, we are removing the extra debug information. See https://github.com/woocommerce/woocommerce-admin/issues/3265#issuecomment-561009312.

I've adjusted the error messages slightly so we get a little more visibility of what failed, but I don't think we need all the additional details written to logs on all sites.

### Testing Directions

* To fully test that the error message works, go into src/API/OnboardingPlugins.php and make a change to install_plugin that would allow one of the two woocommerce_rest_plugin_install errors to occur. For example, you could change `'slug' => $slug`, to `'slug' => '',`.
* Delete WooCommerce Services, and enable the profile wizard.
* Go to the first step, and see the installation fail.
* Remove your change, and test again, seeing the installation succeed.

### DIFF
--- a/src/API/OnboardingPlugins.php
+++ b/src/API/OnboardingPlugins.php
@@ -171,8 +171,8 @@ class OnboardingPlugins extends \WC_REST_Data_Controller {
 
 		require_once ABSPATH . 'wp-admin/includes/plugin.php';
 
-		$slug              = $plugin;
-		$path              = $allowed_plugins[ $slug ];
+		$path              = $allowed_plugins[ $plugin ];
+		$slug              = sanitize_key( $plugin );
 		$installed_plugins = get_plugins();
 
 		if ( in_array( $path, array_keys( $installed_plugins ), true ) ) {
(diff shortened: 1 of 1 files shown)


justinshreve:
@joshuatf I'll fix that in a follow-up PR, thanks!

PR #3338, opened 2019-12-03
Add autocomplete props to address fields in onboarding and adjust country/state matching

justinshreve:
From the internal testing thread:

> When using auto-fill on Chrome Address 2 was populated with the same content as Address 1.

This PR adds some autocomplete attributes to the store details page, so auto-fill knows how to use the fields.

It also adjusts the logic on the country/state autocomplete code, because I was having trouble consistently filling across Chrome & Safari. Also, if you prerefilled data was abbreviated `OH` for example instead of `Ohio`, results wouldn't always show up correctly.

- Test using your browser's autofill feature. I tested in Safari & Chrome.
- Try editing your autofill data (for Safari this is in the `Contacts` app, for Chrome this is under Settings > Addresses) and use different formats for Country and State (for example US vs United States, OH vs Ohio).

### DIFF
--- a/client/dashboard/components/settings/general/store-address.js
+++ b/client/dashboard/components/settings/general/store-address.js
@@ -86,20 +86,21 @@ export function getCountryStateAutofill( options, countryState, setValue ) {
 	useEffect(
 		() => {
 			let filteredOptions = [];
+			const countrySearch = new RegExp( escapeRegExp( autofillCountry ), 'i' );
 			if ( autofillState.length || autofillCountry.length ) {
-				const countrySearch = new RegExp(
-					escapeRegExp( autofillCountry.replace( /\s/g, '' ) ),
-					'i'
-				);
-				filteredOptions = options.filter( option =>
-					countrySearch.test( option.label.replace( '-', '' ).replace( /\s/g, '' ) )
-				);
+				filteredOptions = options.filter( option => countrySearch.test( option.label ) );
 			}
 			if ( autofillCountry.length && autofillState.length ) {
 				const stateSearch = new RegExp( escapeRegExp( autofillState.replace( /\s/g, '' ) ), 'i' );
 				filteredOptions = filteredOptions.filter( option =>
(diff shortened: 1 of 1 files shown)


client/dashboard/components/settings/general/store-address.js
```diff
@@ -86,20 +86,21 @@ export function getCountryStateAutofill( options, countryState, setValue ) {
...
-				);
+				filteredOptions = options.filter( option => countrySearch.test( option.label ) );
 			}
 			if ( autofillCountry.length && autofillState.length ) {
 				const stateSearch = new RegExp( escapeRegExp( autofillState.replace( /\s/g, '' ) ), 'i' );
 				filteredOptions = filteredOptions.filter( option =>
 					stateSearch.test( option.label.replace( '-', '' ).replace( /\s/g, '' ) )
 				);
+
+				if ( filteredOptions.length > 1 ) {
```
joshuatf: Nice solution.  Could we also use this same logic to grab state keys like “GA” for “Georgia?”

justinshreve: How doeshttps://github.com/woocommerce/woocommerce-admin/pull/3338/commits/9484db968b6b6f63b365d421fca3707090cc562e look?

timmyc: @joshuatf can you 👀 this again?

justinshreve: @timmyc I *think* @joshuatf is AFK now, since it's the 11th for him. I believe this comment addresses his feedback though. If things look OK to you, we could merge and do a follow-up PR later if necessary.

timmyc: sounds good to me!

justinshreve:
@timmyc Can you give me a green ✅as well? Merging is blocked without an approval and the UI blocks self approvals 😄. Thanks!

PR #3339, opened 2019-12-03
Stardardize hooks

### LINKED ISSUE #3301: Consistent Filter Naming
We are inconsistent with our filter/action naming throughout the WooCommerce Admin code base.

For example we have

* woocommerce_admin_
* wc_admin_
* woocommerce_

We want to standardize on using `woocommerce_admin_` per the conversation at p1574361629414700-slack-wc-admin and `woocommerce_rest_` for REST API endpoints.

rrennick:
Fixes #3301

*Note* The unit test issue is the `date()` calls in `ReportSync.php`. I've left those to be addressed in #3285 

Once WC Admin is merged into WC core, users and third party developers will see it as part of WooCommerce. With that in mind, this PR eliminates the use of `woocommerce_admin_` as a hook/filter prefix except plugin functions (eg. update routines) and onboarding (as WC core currently has an onboarding feature). 

At a later date when the core onboarding is eliminated we can look at renaming `woocommerce_admin_onboarding_` hooks/filters to `woocommerce_onboarding`.

This PR updates the hook/filter names as follows:

- renames `woocommerce_reports_` hook prefix with `woocommerce_analytics_`
- adds `woocommerce_navigation_` hook prefix
- updates `wc_admin_` hooks to `woocommerce_admin_` or one of the above two prefixes
- standardizes `woocommerce_note_` prefix 

- Check that the dashboard and Analytics continue to function.

### Changelog Note:

Dev: Standardize hook and filter prefixes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,10 +64,10 @@ There are a number of helper scripts exposed via our `package.json` (below list 
 
 ### Debugging synced lookup information:
 
-To debug synced lookup information in the database, you can bypass the action scheduler and immediately sync order and customer information by using the `woocommerce_disable_order_scheduling` hook.
+To debug synced lookup information in the database, you can bypass the action scheduler and immediately sync order and customer information by using the `woocommerce_analytics_disable_order_scheduling` hook.
 
 ```php
-add_filter( 'woocommerce_disable_order_scheduling', '__return_true' );
+add_filter( 'woocommerce_analytics_disable_order_scheduling', '__return_true' );
 ```
 
 ### Using `debug` package.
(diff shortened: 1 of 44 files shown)


rrennick:
> I found that I would think could easily be woocommerce_analytics_*.

@psealock Should I clarify that this is only PHP hooks/filters. With #3203 undecided, I thought we should postpone the JS ones until we decide on that.

> What do you mean by "plugin function"?

The install/update routines, which features are enabled, etc. The hooks/filters that are not specific to any feature set.

psealock:
I rebased from master and updated extension examples to reflect new hooks.

>Note The unit test issue is the date() calls in ReportSync.php. I've left those to be addressed in #3285

Given this, I think its ok to merge.

PR #3347, opened 2019-12-04
Default the Calypso environment to production

### LINKED ISSUE #3055: Onboarding: Adjust wizard query string and launch Calypso flows
As a part of public user testing and implementing split testing (https://github.com/woocommerce/woocommerce-admin/issues/2759), we need to launch the new onboarding Calypso flows into production.

As a part of this, we will need to adjust the query string we are currently using:

The new feature flagged code uses the same `?from=woocommerce-setup-wizard` query string that the current production version of onboarding users. We need to pick a different query string so both versions of the WooCommerce Calypso flows can be used in production at the same time. https://github.com/Automattic/wp-calypso/blob/809fe3b8ee3b046fcaf569828dee313eefb053ea/client/document/index.jsx#L78-L81

justinshreve:
Closes #3055.

Once https://github.com/Automattic/wp-calypso/pull/38168 is merged, we need to update the default Calypso environment to be `production` instead `wpcalypso` (the only environment onboarding is currently enabled for). This PR handles that.

* Test with D36472-code
* Connect (either through the normal profile wizard flow or via the connection buttons under `Settings > Help > Setup Wizard`).
* Verify that you end up on `wordpress.com` and not `wpcalypso.wordpress.com` for both the Jetpack and WooCommerce.com connection flows.

### DIFF
--- a/src/API/OnboardingPlugins.php
+++ b/src/API/OnboardingPlugins.php
@@ -300,8 +300,7 @@ class OnboardingPlugins extends \WC_REST_Data_Controller {
 		$redirect_url = apply_filters( 'woocommerce_admin_onboarding_jetpack_connect_redirect_url', esc_url_raw( $request['redirect_url'] ) );
 		$connect_url  = \Jetpack::init()->build_connect_url( true, $redirect_url, 'woocommerce-onboarding' );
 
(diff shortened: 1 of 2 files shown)


jeffstieler:
+1 - just tested well for me alongside the Phab.

PR #3350, opened 2019-12-04
add filter to allow WC Admin to be disabled

### LINKED ISSUE #2223: Analytics: Add Setting to Disable New Analytics
Per discussions from recent meetings, we would like to add a new store setting that allows for turning off the new Analytics system. To capture the reasoning here a bit for those who haven't been involved in the conversations, the reasons for allowing analytics to be turned off are:

- Established stores may already have an analytics system in place that they like, and don't want the new reports.
- High volume stores don't want to have the added database writes that come along with the analytics lookup tables.
- Some stores may prefer to not opt-in for the new analytics right away, this allows control to turn the feature on when they would like to use it.

**Tasks**
The following items need to be completed:

- [ ] Create a new store setting, thinking we should create a new 'sub tab' under the Advanced section of Settings, but /cc @josemarques for thoughts
- [ ] Determine the verbiage to display, work with Pi for help
- [ ] If Analytics are turned off, add logic to turn off population of lookup tables
- [ ] If Analytics are turned off, default WooCommerce link back to the Orders Listing page.

rrennick:
Fixes #2223

After much discussion on the topic, this PR adds a filter that allows completely disabling WooCommerce Admin.

- Add the filter `add_filter( 'woocommerce_analytics_disabled', '__return_true' );`

### Changelog Note:

Tweak: Add a filter that allows disabling WooCommerce Admin.

### DIFF
--- a/src/FeaturePlugin.php
+++ b/src/FeaturePlugin.php
@@ -54,6 +54,15 @@ class FeaturePlugin {
 	 * Init the feature plugin, only if we can detect both Gutenberg and WooCommerce.
 	 */
 	public function init() {
+		/**
+		 * Filter allowing WooCommerce Admin to be disabled.
+		 *
+		 * @param bool $disabled False.
+		 */
+		if ( apply_filters( 'woocommerce_analytics_disabled', false ) ) {
+			return;
+		}
+
 		$this->define_constants();
 
 		require_once WC_ADMIN_ABSPATH . '/includes/core-functions.php';


src/FeaturePlugin.php
```diff
@@ -54,6 +54,15 @@ protected function __construct() {}
 	 * Init the feature plugin, only if we can detect both Gutenberg and WooCommerce.
 	 */
 	public function init() {
+		/**
+		 * Filter allowing WooCommerce Admin to be disabled.
+		 *
+		 * @param bool $disabled False.
+		 */
+		if ( apply_filters( 'woocommerce_analytics_disabled', false ) ) {
```
justinshreve: @rrennick If this filter is to disable all of WooCommerce Admin (and not just the analytics feature), shouldn't this filter be something like `woocommerce_admin_disabled`?

rrennick: @justinshreve I named it consistent with #3339 which hasn't been merged yet.

rrennick: @justinshreve Gotcha, thanks for catching that. I created #3361 to rename the filter.

timmyc: @rrennick I missed this PR during the review phase, but will this still work in the context when the feature is pulled in via Packagist in Woo core?

rrennick: > will this still work in the context when the feature is pulled in via Packagist in Woo core?

Yes, it should.

PR #3358, opened 2019-12-05
Fix revenue question translation with currency conversion

### LINKED ISSUE #2768: Business Details Revenue Options i18n

justinshreve:
Closes #2768.

This PR implements @joshuatf's idea for solving the revenue translation issue:

> My current thought is that if we really want to solve this we can go back to getting a rough range based on the currency. Since @justinshreve has added currency data for the most commonly selected countries, we could create an exchange rate value which we can then round to the nearest nth number.

I didn't end up storing the exchange rates with the rest of the currency info. That info is in an exportable public currency package, and I worry about the rates getting used the wrong way if we expose them there. They are meant to be rough and not useful in most situations, so I opted to just include them in the conversion function, inside the dashboard.

### Screenshots

France:

Screen Shot 2019-12-05 at 12 36 35 PM

United States:

Screen Shot 2019-12-05 at 12 36 03 PM

* Enable the profile wizard, and select a non-US country. We currently only set smart currency defaults for the top countries + EU, so select one of the following regions: https://github.com/woocommerce/woocommerce-admin/blob/master/packages/currency/src/index.js#L133
* Continue through to the business details page, and expose the revenue question (Yes, I'm currently selling elsewhere).
* Confirm that the revenue question has been translated and converted to use a rough range in the correct currency.

### DIFF
--- a/client/dashboard/profile-wizard/steps/business-details.js
+++ b/client/dashboard/profile-wizard/steps/business-details.js
@@ -141,6 +141,11 @@ class BusinessDetails extends Component {
 	}
 
 	convertCurrency( value ) {
+		const region = getCurrencyRegion( this.props.settings.woocommerce_default_country );
+		if ( 'US' === region ) {
+			return value;
+		}
+
 		// These are rough exchange rates from USD.  Precision is not paramount.
 		// The keys here should match the keys in `getCurrencyData`.
 		const exchangeRates = {
@@ -159,10 +164,11 @@ class BusinessDetails extends Component {
 			CA: 1.32,
 		};
 
-		const region = getCurrencyRegion( this.props.settings.woocommerce_default_country );
 		const exchangeRate = exchangeRates[ region ] || exchangeRates.US;
+		const digits = exchangeRate.toString().split( '.' )[ 0 ].length;
+		const multiplier = Math.pow( 10, 2 + digits );
 
-		return Math.round( value * exchangeRate / 1000 ) * 1000;
+		return Math.round( value * exchangeRate / multiplier ) * multiplier;
 	}
 
 	numberFormat( value ) {
--- a/client/dashboard/profile-wizard/steps/business-details.js
+++ b/client/dashboard/profile-wizard/steps/business-details.js
@@ -24,6 +24,7 @@ import { recordEvent } from 'lib/tracks';
 import { formatCurrency } from 'lib/currency-format';
 import Plugins from 'dashboard/task-list/tasks/steps/plugins';
 import { pluginNames } from 'wc-api/onboarding/constants';
+import { getCurrencyRegion } from 'dashboard/utils';
 
 const wcAdminAssetUrl = getSetting( 'wcAdminAssetUrl', '' );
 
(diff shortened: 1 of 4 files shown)


client/dashboard/profile-wizard/steps/business-details.js
```diff
@@ -139,18 +140,47 @@ class BusinessDetails extends Component {
...
+		const exchangeRate = exchangeRates[ region ] || exchangeRates.US;
+
+		return Math.round( value * exchangeRate / 1000 ) * 1000;
+	}
+
+	numberFormat( value ) {
+		return formatValue( 'number', value );
+	}
+
+	getNumberRangeString( min, max = false, format = this.numberFormat ) {
```
justinshreve: @joshuatf: I had to add back in the `format` parameter that was removed in https://github.com/woocommerce/woocommerce-admin/commit/1b2fb37ada966aea43593977d9eaa91082f06d48. The product question uses it to format regular numbers, and the revenue question uses it to format currency.

packages/currency/src/index.js
```diff
@@ -127,6 +127,7 @@ export default class Currency {
 /**
  * Returns currency data by country/region. Contains code, symbol, position, thousands separator, decimal separator, and precision.
  *
+ * Dev Note: When adding new currencies below, the exchange rate array should also be updated in WooCommerce Admin's `business-details.js`.
```
joshuatf: Not opposed to the way you've done it here as that may be better, but wanted to get your opinion on adding the exchange rates directly to the items below for easy access:

```
		CA: {
			code: 'CAD',
			symbol: '$',
			symbolPosition: 'left',
			thousandSeparator: ',',
			decimalSeparator: '.',
			precision: 2,
            usdExchangeRate: 1.32
		},
```

This would be easier to maintain and may cut down some of the logic here.  Or are there some benefits to separating this out?

client/dashboard/profile-wizard/steps/business-details.js
joshuatf: That makes sense- thanks for the explanation 😄

client/dashboard/profile-wizard/steps/business-details.js
```diff
@@ -139,18 +140,47 @@ class BusinessDetails extends Component {
...
+			RU: 63.74,
+			TR: 5.75,
+			MX: 19.37,
+			CA: 1.32,
+		};
+
+		const region = getCurrencyRegion( this.props.settings.woocommerce_default_country );
+		const exchangeRate = exchangeRates[ region ] || exchangeRates.US;
+
+		return Math.round( value * exchangeRate / 1000 ) * 1000;
```
joshuatf: Nice!  Optional, but we could add a zero for every additional digit beyond the first digit to round currencies with larger differences in exchange rates.

E.g., MX would be `10000`, VND would become `10000000`.

This might make the numbers a bit more rounded some currencies:

**Before**
Screen Shot 2019-12-09 at 1 41 51 PM

**After**
Screen Shot 2019-12-09 at 1 46 43 PM

packages/currency/src/index.js
justinshreve: My main reasoning was mentioned in the PR description above (which might be better as a comment):

> I didn't end up storing the exchange rates with the rest of the currency info. That info is in an exportable public currency package, and I worry about the rates getting used the wrong way if we expose them there. They are meant to be rough and not useful in most situations, so I opted to just include them in the conversion function, inside the dashboard.

If you're not that worried about it and think documentation is enough (or naming it something like `roughExchangeRate`) I'm happy to include it in the currency package though.

client/dashboard/profile-wizard/steps/business-details.js
justinshreve: Nice suggestion @joshuatf! Handled in https://github.com/woocommerce/woocommerce-admin/pull/3358/commits/df799125fd068ed477d9a4ad789bbe2345444783.

packages/currency/src/index.js
joshuatf: Sorry, I must have skipped right over this above.

That makes a lot of sense to me.  Let's keep it as is.

client/dashboard/profile-wizard/steps/business-details.js
```diff
@@ -159,10 +164,11 @@ class BusinessDetails extends Component {
 			CA: 1.32,
 		};
 
-		const region = getCurrencyRegion( this.props.settings.woocommerce_default_country );
 		const exchangeRate = exchangeRates[ region ] || exchangeRates.US;
+		const digits = exchangeRate.toString().split( '.' )[ 0 ].length;
+		const multiplier = Math.pow( 10, 2 + digits );
```
joshuatf: Nice!  Creative way of doing this.

PR #3362, opened 2019-12-05
Fix JS error on DevDocs

jeffstieler:
### To reproduce:
* Go to DevDocs
* Open the Inbox panel
* 💥

The root of the issue is calling `moment.updateLocale()` with unregistered locales (for example, WP's `fr_FR`). See: https://github.com/woocommerce/woocommerce-admin/blob/master/packages/components/src/date-range-filter-picker/docs/example.js#L25

This PR seeks to remove unnecessary locale update call ultimately letting the `@wordpress/date` module handle it.

#### Question:

Should we remove the `loadLocaleData()` export from `@woocommerce/date`?

* Go to DevDocs
* Open the Inbox panel
* Verify no explosions
* Check that `moment` is still loading the proper locale data:
  * This is best done on non-`en_US`
  * Open console, compare `wcSettings.locale.weekdaysShort` and `moment.localeData()._weekdaysShort`

### Changelog Note:

Fix: errant moment locale data loading.

### DIFF
--- a/client/lib/date.js
+++ b/client/lib/date.js
@@ -32,14 +32,9 @@ import {
 	weekTicksThreshold,
 	defaultTableDateFormat,
 	getDateFormatsForInterval,
-	loadLocaleData,
 	dateValidationMessages,
 	validateDateInputForRange,
 } from '@woocommerce/date';
-
-// Load the store's locale.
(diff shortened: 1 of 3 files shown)


psealock:
I'm merging this so it can be included in today's release.

PR #3386, opened 2019-12-06
don't load analytics dashboard for users without needed capability

### LINKED ISSUE #3276: Header should only show if current_user_can manage_woocommerce
Users who have no capability to access the wc-admin page shouldn't be shown a link to that page.

Ideally I don't think users should even be shown the WooCommerce Admin header if they don't have ability to use any of the plugin's features.

I have added css to hide the header from users who don't need it.

No.

**Additional context**
I have users who can do nothing but view/edit orders. They process the orders for shipment. They can't view reports, they can't edit products, they can't do anything. In the screenshot below you can see on the left how empty their admin menu is. They shouldn't even be shown the wc-admin header. In fact, if they click that purple link, they get an error that they don't have access to the page. Why should they even be shown the link.

I could be wrong, but shouldn't that entire header bar only be shown if `current_user_can( 'manage_woocommerce' ) || current_user_can( 'view_woocommerce_reports' )`

chrome_uix5VgjPVB

rrennick:
Fixes #3276 

This PR adds checks to the dashboard loader to ensure that the current user has the required WP capability to use the analytics functionality.

- Install a role editing plugin
- Create a role that has `read` plus all `shop_order` post type caps
- Create a user with this role
- Log into the dashboard as a shop manager & verify that screens load & activity panel works
- Log in as test user
- Go to orders page
- Check that WCA scripts aren't included
- Check that `root` and `embedded-root` are not injected into the page

### Changelog Note:

Fix: remove the header when user doesn't have required permissions

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -141,6 +141,15 @@ class Loader {
 	}
 
 	/**
+	 * Helper function indicating whether the current user has the required analytics capability.
+	 *
+	 * @return bool
+	 */
+	public static function user_can_analytics() {
+		return current_user_can( static::get_analytics_capability() );
+	}
+
+	/**
 	 * Returns if a specific wc-admin feature is enabled.
 	 *
 	 * @param  string $feature Feature slug.
@@ -401,7 +410,7 @@ class Loader {
 			return;
 		}
 
-		if ( ! current_user_can( static::get_analytics_capability() ) ) {
+		if ( ! static::user_can_analytics() ) {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


src/Loader.php
```diff
@@ -378,6 +401,10 @@ public static function load_scripts() {
 			return;
 		}
 
+		if ( ! current_user_can( static::get_analytics_capability() ) ) {
```
jeffstieler: Nitpick - what do you think about putting this in a helper function?

rrennick: > what do you think about putting this in a helper function?

Added in `82f59db2`

PR #3390, opened 2019-12-06
scroll to notices when made visible while scrolled down

### LINKED ISSUE #3127: Triggering Notices view when scrolled causes confusion
**Describe the bug**
With the sticky WooAdmin bar the Notices link triggers the show/hide of notices at the top of the screen. If you're already past this section and click 'Notices' nothing aside from the purple underline seems to occur. Yes the notices are displayed above the current screen view but the user doesn't realize this. It would be nice if clicking Notices also scrolls back to the notices so they are placed into view.

As well when notices are visible the 'Help' tab vanishes.

**To Reproduce**

1. Go to any WooCommerce screen.
2. Scroll down a screen.
3. Click the 'Notices' button in the WooAdmin bar.
4. Note nothings seems to happen. The notices do get displayed but the user doesn't realize this.

**Expected behavior**
It would be nice for this action to trigger a scrollback to the notices section.

**Screenshots**
e01aa8596921b4a3913b75a61e988f41

* OS: MacOS Mojave 10.14.5
* Browser: Chrome 77.0.3865.90
* WordPress: 5.2.4
* WooCommerce: 3.7.1
*WooCommerce Admin: 0.20.1

rrennick:
Fixes #3127

This PR will cause the browser to scroll up to the notices when the activity button is clicked and the page is scrolled down where they wouldn't be seen.

- Ensure you have a admin notice that is visible on all dashboard pages
- Go to any Analytics page
- While scrolled up show notices
- Scroll down until all notices are hidden
- Hide then show notices
- Repeat on a WooCommerce page (eg. Orders) where the WC Admin header is shown

### Changelog Note:

Tweak: Scroll to notices when displayed while the notice area is scrolled out of view.

### DIFF
--- a/client/header/activity-panel/wordpress-notices.js
+++ b/client/header/activity-panel/wordpress-notices.js
@@ -10,6 +10,9 @@ import { IconButton } from '@wordpress/components';
 import { intersection, noop, partial } from 'lodash';
 import PropTypes from 'prop-types';
 
+// Height of the activity header in px.
+const HEADER_HEIGHT = 56;
+
 class WordPressNotices extends Component {
 	constructor() {
 		super();
@@ -176,6 +179,12 @@ class WordPressNotices extends Component {
 		notices.classList.remove( 'woocommerce-layout__notice-list-hide' );
 		screenMeta && screenMeta.classList.add( 'is-hidden-by-notices' );
 		screenLinks && screenLinks.classList.add( 'is-hidden-by-notices' );
+
+		const noticePosition = notices.getBoundingClientRect();
+		if ( noticePosition.y < HEADER_HEIGHT ) {
+			window.scrollBy( 0, noticePosition.y - HEADER_HEIGHT );
+		}
+		this.setState( { noticesOpen: true } );
 	}
 
 	hideNotices() {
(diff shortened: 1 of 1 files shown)


client/header/activity-panel/wordpress-notices.js
```diff
@@ -176,6 +179,12 @@ class WordPressNotices extends Component {
 		notices.classList.remove( 'woocommerce-layout__notice-list-hide' );
 		screenMeta && screenMeta.classList.add( 'is-hidden-by-notices' );
 		screenLinks && screenLinks.classList.add( 'is-hidden-by-notices' );
+
+		const noticePosition = notices.getBoundingClientRect();
+		if ( noticePosition.y < HEADER_HEIGHT ) {
+			window.scrollBy( 0, noticePosition.y - HEADER_HEIGHT );
```
jeffstieler: Would it suffice to always scroll to the top using `window.scrollBy( 0, -window.scrollY );`

jeffstieler: We should probably change this - the `x` and `y` properties aren't supported in IE or Safari: https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect

rrennick: @jeffstieler Thanks for catching that. I've updated the PR as suggested.

PR #3391, opened 2019-12-06
Add filters to WP notice collapsing behavior.

### LINKED ISSUE #2909: Add the ability for 3rd party plugins to declare notices as primary
Some WooCommerce plugins create notices that are important and WooCommerce Admin is hiding them under the "Notices" inbox.

I know that the whole idea behind this is to keep the backend clean, but we can't just hide all the notices and pretend only the WooCommerce ones are important.

Example:
I got this invoicing plugin that informs the administrator when the invoice was not issued because of some error on the order.

No one sees this if it's hidden under the Notices inbox, and it's important that they do see it:
image

I'm now getting several support tickets from clients stating that there's no feedback on the plugin actions. There is, but they're using WooCommerce Admin and they're not seeing it.

How can a developer create notices that will not be hidden by WooCommerce Admin now, and when it gets merged into core, like the "Order updated" one?

=====

Adding my comment to the OP for visibility - @timmyc 

Based off some internal conversations in: p90Yrv-1o6-p2

I'm proposing we implement the following solutions ( /ht and quoting @DanReyLop )

- The result of [`shouldCollapseNotice(element)`](https://github.com/woocommerce/woocommerce-admin/blob/master/client/header/activity-panel/wordpress-notices.js#L119) is filterable, so extensions can hook into it and return false their notices.
- The [`noticesToHide`](https://github.com/woocommerce/woocommerce-admin/blob/master/client/header/activity-panel/wordpress-notices.js#L121) array is filterable, so extensions can hook into it and add their own css class.

jeffstieler:
Fixes #2909.

This PR seeks to allow 3PDs to filter the WordPress notice collapsing behavior.

It introduces two new filters: `woocommerce_admin_notices_to_show` and `woocommerce_admin_should_hide_notice`. Additionally, it corrects a variable name - `noticesToHide` is used for "notices to not collapse" so I renamed it to `noticesToShow`.

- Verify that existing WP notice collapsing behavior remains unchanged
- Use the `woocommerce_admin_notices_to_show` filter to modify the notices to show, verify its function
- Use the `woocommerce_admin_should_hide_notice` filter to modify the result for an element, verify its function

### Changelog Note:

Enhancement: allow filtering of hidden WP notices.

### DIFF
--- a/client/header/activity-panel/style.scss
+++ b/client/header/activity-panel/style.scss
@@ -279,7 +279,16 @@
 }
 
 #wp__notice-list {
-	padding-right: 20px;
+	padding-left: 5px;
+	padding-right: 5px;
+
+	@include breakpoint( '<782px' ) {
+		padding-left: 20px;
+		padding-right: 20px;
+		& > div {
+			margin: 20px 0 10px 0;
+		}
+	}
 }
 
 .woocommerce-layout__notice-list .jitm-card {
--- a/client/header/activity-panel/wordpress-notices.js
+++ b/client/header/activity-panel/wordpress-notices.js
@@ -81,11 +81,9 @@ class WordPressNotices extends Component {
 			document.getElementById( 'woocommerce-layout__notice-list' );
 
 		let count = 0;
-		for ( let i = 0; i <= notices.children.length; i++ ) {
-			const notice = notices.children[ i ];
-			if ( ! notice ) {
-				continue;
-			} else if ( 0 === notice.innerHTML.length ) {
+
+		for ( const notice of Array.from( notices.children ) ) {
+			if ( 0 === notice.innerHTML.length ) {
 				// Ignore empty elements in this part of the DOM.
 				continue;
 			} else if ( ! this.shouldCollapseNotice( notice ) ) {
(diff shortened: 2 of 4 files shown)


timmyc:
@jeffstieler this is looking great - since we have had requests for how to do this from the community - do you think we should have a code snippet in our examples directory with how to ensure one's admin alert is shown?

psealock:
+1 for an example

timmyc:
I still say we land this with an example. The future of the W notif panel is still foggy, so lets proceed like it is going to ship to core still.

jeffstieler:
Found a bug while writing the example extension where the `NodeList` of WP notices was being modified during iteration, causing notices to be skipped over. Fixed in d48b970.

Example added in bfcdaf4.

timmyc:
I think we should land this still since we don't have a full consensus yet on the future of the W panel.

jeffstieler:
@timmyc added a styling tweak to address the width differences in df82116 - can you take another look before we merge?

timmyc:
Thanks for the margin fix, LGTM

Issue #3372, opened 2019-12-06
Update Minimum WordPress Version to 5.3


timmyc:
Epic parent issue #3384 

To set the stage for using React hooks, lets bump the required WP version of the plugin to 5.3

PR #3393, opened 2019-12-08
Bump minimum WP version to 5.3

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/3372

Bumps version. 

### Testing

Double check that references to WP 5.2 have all been updated.

### DIFF
--- a/src/Package.php
+++ b/src/Package.php
@@ -24,10 +24,10 @@ class Package {
 	/**
 	 * Init the package.
 	 *
-	 * Only initialize for WP 5.2 or greater.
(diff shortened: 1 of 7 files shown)


src/Package.php
```diff
@@ -27,7 +27,7 @@ class Package {
 	 * Only initialize for WP 5.2 or greater.
```
jeffstieler: Missed `5.2`.

psealock:
Thanks for catching that @jeffstieler, its now been addressed.

PR #3400, opened 2019-12-09
Onboarding: Update shipping task button text

### LINKED ISSUE #3399: Onboarding: Update shipping task button text
**Describe the bug**
As noted in https://github.com/woocommerce/woocommerce-admin/pull/3331#issuecomment-561691267:

>This button label should only read "Complete task" if step 3 doesn't exist. IE when the store address doesn't meet the required criteria for us to advertise our label printing services. Otherwise it should be "Proceed".

**To Reproduce**

1. Go to the shipping task.
2. Make sure the store country is set to US, CA, AU, or, GB.
3. Note that the 2nd step says "Complete" when more steps exist.

**Expected behavior**
Expected this button to say "Proceed" when more steps exist.

**Screenshots**
Screen Shot 2019-12-09 at 3 14 38 PM

joshuatf:
Fixes #3399

Updates the button text depending on whether or not remaining steps in the task exist.

### Screenshots
Screen Shot 2019-12-09 at 3 21 37 PM

Screen Shot 2019-12-09 at 3 21 14 PM

1. Set your store default country to US, AU, CA, or GB.
2. Go to the shipping task and make sure that the 2nd step button text reads "Proceed."
3. Set your store to a country not listed above.
4. Make sure the 2nd step of the shipping task button text reads "Complete task."

### DIFF
--- a/client/dashboard/task-list/tasks/shipping/index.js
+++ b/client/dashboard/task-list/tasks/shipping/index.js
@@ -144,9 +144,12 @@ class Shipping extends Component {
 
 	getSteps() {
 		const { countryCode } = this.props;
-		const plugins = [ 'GB', 'CA', 'AU' ].includes( countryCode )
-			? [ 'woocommerce-shipstation-integration' ]
-			: [ 'jetpack', 'woocommerce-services' ];
+		let plugins = [];
+		if ( [ 'GB', 'CA', 'AU' ].includes( countryCode ) ) {
+			plugins = [ 'woocommerce-shipstation-integration' ];
+		} else if ( 'US' === countryCode ) {
+			plugins = [ 'jetpack', 'woocommerce-services' ];
+		}
 
 		const steps = [
 			{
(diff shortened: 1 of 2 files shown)


client/dashboard/task-list/tasks/shipping/index.js
```diff
@@ -174,6 +174,11 @@ class Shipping extends Component {
 				),
 				content: (
 					<ShippingRates
+						buttonText={
+							[ 'US', 'GB', 'CA', 'AU' ].includes( countryCode )
```
justinshreve: Nitpick: We have a few of these country code arrays in here now (for example `visible: [ 'US', 'GB', 'CA', 'AU' ].includes( countryCode ),` on the plugins step).

If WooCommerce Services/Shipping or ShipStation adds support for more countries, we'll need to make sure all the arrays are updated correctly. Should we instead have an array of countries for WCS, an array for ShipStation, and then use those in the correct places?

joshuatf: Agreed.  I updated this a bit in https://github.com/woocommerce/woocommerce-admin/pull/3400/commits/a7aed344ed0a5030a57e6549307b0f1f59afe345 to determine task visibility by the plugins so we can consolidate the country checks down to a single area.  I think these steps are actually dependent on plugins so I'm thinking this makes more sense.

Let me know if you think this makes sense or if you have something else in mind.

justinshreve: Yeah I think that's a great way to handle it 👍. I'll go ahead and re-test this, and then merge it, since I believe you are AFK now.

PR #3402, opened 2019-12-09
Onboarding -  Add button to continue setup after importing products

### LINKED ISSUE #3235: Add Products - Import Products - display a notification to go back to the task
We should display a notification guiding the user after he imports the products. Right now the only call to action he sees is to check the products:

https://cld.wthms.co/Eh4z9a

justinshreve:
Fixes #3235.

This PR adds a button to the product import page to continue setup, if the user entered the import flow through onboarding.

### Screenshots

Screen Shot 2019-12-09 at 2 12 21 PM

* Delete your products
* Enable the task list and go to the products task and select import
* Walk through the importer steps (you can use [this CSV file for testing](https://docs.woocommerce.com/document/product-csv-importer-exporter/dummy-data/))
* Verify you see the continue setup button
* Leave, and import products the normal way (`Tools > Import`) and notice the primary button is just "Add Products"

### DIFF
--- a/client/dashboard/task-list/tasks/products.js
+++ b/client/dashboard/task-list/tasks/products.js
@@ -37,7 +37,7 @@ const subTasks = [
 		after: <i className="material-icons-outlined">chevron_right</i>,
 		onClick: () => recordEvent( 'tasklist_add_product', { method: 'import' } ),
 		href: getAdminLink(
-			'edit.php?post_type=product&page=product_importer&wc_onboarding_active_task=products'
+			'edit.php?post_type=product&page=product_importer&wc_onboarding_active_task=product-import'
 		),
 	},
 	{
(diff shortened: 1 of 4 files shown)


client/dashboard/task-list/tasks/products.js
```diff
@@ -37,7 +37,7 @@ const subTasks = [
 		after: <i className="material-icons-outlined">chevron_right</i>,
 		onClick: () => recordEvent( 'tasklist_add_product', { method: 'import' } ),
 		href: getAdminLink(
-			'edit.php?post_type=product&page=product_importer&wc_onboarding_active_task=products'
+			'edit.php?post_type=product&page=product_importer&wc_onboarding_active_task=product-import'
```
joshuatf: I was originally thinking we could keep the `wc_onboarding_active_task` as `products` for any product related task.  But feel free to leave this as is if you feel it offers benefit or better readability.

justinshreve: The problem is that `products` will be considered complete by the time the `done` page loads, because products will exist. Using a separate key here that gets cleared when the import is done makes it much easier to hook in and update the button.

PR #3403, opened 2019-12-10
Add starter pack script

psealock:
Adds a starter pack script for basic usage of [wp-scripts](https://github.com/WordPress/gutenberg/tree/master/packages/scripts) with modification for WooCommerce Admin. All linting and unit testing defaults to standard WP environment designated by wp-scripts.

1. `npm run create-wc-extension`
2. `cd ..`
3. cd into your new folder (your chosen name lowercased and hyphenated)
4. `npm install && npm start`
4. Activate the plugin
5. Go to any wc-admin page and see "Hello World" appear in the console.

## Next Steps

* Possibly add our overrides such as `.eslintrc` or `phpunit.xml.dist`. We'd need to be sure these are the standards we'd like to promote to the ecosystem.
* If this is helpful, move this to `packages` so that developers can simply `npx create-wc-extension` independently of this repo.


bin/starter-pack/_main.php
```diff
@@ -0,0 +1,33 @@
...
+	wp_register_script(
+		'{{extension_slug}}',
+		plugins_url( '/build/index.js', __FILE__ ),
+		array(
+			'wp-hooks',
+			'wp-element',
+			'wp-i18n',
+			'wc-components',
+		),
+		filemtime( dirname( __FILE__ ) . '/build/index.js' ),
```
nerrad: If you are using the latest `wp-scripts`, these values should be provided automatically by the `*.asset.php` files automatically output by the [Dependency Extraction Webpack Plugin](https://github.com/WordPress/gutenberg/tree/master/packages/dependency-extraction-webpack-plugin) that is included with `@wordpress/scripts`.  You can see this at work in the wc blocks project.

bin/starter-pack/_webpack.config.js
```diff
@@ -0,0 +1,14 @@
...
+	externals: {
+		'@woocommerce/components': [ 'window', 'wc', 'components' ],
+		'@woocommerce/csv-export': [ 'window', 'wc', 'csvExport' ],
+		'@woocommerce/currency': [ 'window', 'wc', 'currency' ],
+		'@woocommerce/date': [ 'window', 'wc', 'date' ],
+		'@woocommerce/navigation': [ 'window', 'wc', 'navigation' ],
+		'@woocommerce/number': [ 'window', 'wc', 'number' ],
+		'@woocommerce/settings': [ 'window', 'wc', 'wcSettings' ],
+	}
+};
```
nerrad: If you go with my suggestion earlier to incorporate the dependency extraction webpack plugin, you can enhance the config of that plugin here so it accounts for `@woocommerce/*` externals too.

bin/starter-pack/starter-pack.js
```diff
@@ -0,0 +1,63 @@
...
+
+	fs.copy( path.join( __dirname, 'src' ), path.join( folder, 'src' ), maybeThrowError );
+
+	fs.copy( folder, path.join( '../', extensionSlug ), error => {
+		maybeThrowError( error );
+
+		fs.remove( folder, maybeThrowError );
+	} );
+
+	console.log( '\n' );
```
nerrad: maybe use `process.stdout.write` instead?

bin/starter-pack/_main.php
psealock: So I can provide an empty array here? What about `wc-components`?

nerrad: no you don't provide an empty array, you provide dependencies that are prepped for you automatically by the webpack plugin:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/a1717cbbe811466fefac04b141be60128e56c356/src/Assets.php#L155-L172

For wc scripts, you'd need to make sure they are registered with the webpack plugin:

Here: https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/a1717cbbe811466fefac04b141be60128e56c356/bin/webpack-helpers.js#L24-L45

And: https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/a1717cbbe811466fefac04b141be60128e56c356/bin/webpack-helpers.js#L272-L277

bin/starter-pack/starter-pack.js
psealock: Switched in https://github.com/woocommerce/woocommerce-admin/pull/3403/commits/c3710022ff03864280e9b64b6fbcd72872a646ee

bin/starter-pack/_main.php
psealock: Thanks for the explanation, I added the Dependency Extraction Webpack Plugin in https://github.com/woocommerce/woocommerce-admin/pull/3403/commits/7262f3d164c291d3bc87aa3179fb806317968ef9

You can test it out by placing something like this in the `index.js` file of the extension and watching webpack create the dependencies automatically.

```js
import { Button } from '@wordpress/components';
import { Chart } from '@woocommerce/components';

console.log( Button );
console.log( Chart );
```

bin/starter-pack/_main.php
```diff
@@ -14,20 +14,30 @@ function add_extension_register_script() {
...
+		$script_url,
+		$script_asset['dependencies'],
+		$script_asset['version'],
 		true
 	);
 
+	wp_register_style(
+		'{{extension_slug}}',
+		plugins_url( '/build/style.css', __FILE__ ),
+		array(),
```
nerrad: Might want to add a comment here about ensuring any dependencies the style might have (`wp-components` etc) are manually added here.

psealock:
Thanks for the reviews and suggestions @nerrad !

The dependency extraction webpack plugin is pretty cool. Thats been added, along basic CSS usage and a comment about style dependencies.

>Should this be made a package (similar to how @wordpress/scripts is a package) so that projects can just install it and run with it?

I think thats the ultimate goal. I hope to get a sense of the usefulness of this script and make this a package.

>We also should standardize around linting/code styles before really promoting this.

I agree here too. We'll need to reconcile those differences first as it would be great to get the projects aligned.

timmyc:
Thanks for the review @nerrad

PR #3418, opened 2019-12-12
AssetDataRegistry: update SQL example

psealock:
The SQL Modification example previously used `wp_add_inline_script` to add data to the `wcSettings` object.

This PR updates that logic to use `AssetDataRegistry` instead.

1. `npm run example -- --ext=sql-modification`
2. Activate the SQL Modification Example plugin
3. Visit a report with currencies, such as Revenue
4. Ensure the currency dropdown is functional

Screen Shot 2019-12-13 at 12 51 57 PM

### DIFF
--- a/docs/examples/extensions/README.md
+++ b/docs/examples/extensions/README.md
@@ -26,3 +26,4 @@ You can make changes to Javascript and PHP files in the example and see changes 
 - `add-task` - Create a custom task for the onboarding task list.
 - `dashboard-section` - Adding a custom "section" to the new dashboard area.
 - `table-column` - An example of how to add column(s) to any report.
+- `sql-modification` - An example of how to modify SQL statements.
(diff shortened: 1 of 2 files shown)


psealock:
>For pure basic functional extensions, I think this example is good.

Right, this is really a most basic implementation purely for simplicity. I added the example you wrote in p90Yrv-1rG-p2 to the comments.

>However, you also want to make sure you include the wc-settings handle as a dependency in the registered script

Done in https://github.com/woocommerce/woocommerce-admin/pull/3418/commits/1a9cc0c7a5bd577d6c586b8cc967ffc521b49ca1

>If an extension was built using the DependencyExtractionWebpackPlugin and had @woocommerce/settings aliased to the wc-settings handle (and external) (like this example in the blocks repo), then that would automatically be taken care of whenever the script imports from @woocommerce/settings.

Indeed it would. It would be good to get some convergence of these extension examples with the starter pack which uses DependencyExtractionWebpackPlugin.

PR #3430, opened 2019-12-17
Data Layer: Settings page to use Settings store

psealock:
Depends on https://github.com/woocommerce/woocommerce-admin/pull/3313

* Apply wp.data store settings as source of truth for the settings page. Settings kept in component state are used to track user edits before clicking "save".
* Expose HOC `withSettings` to apply the following props:
    * getSetting
    * setSetting
    * getSettingError
    * isRequestingSetting

The global `wcSettings` object is only used for initialization of state. Afterwards, the store is  responsible for persisting setting changes store state and server.


packages/data/src/settings/resolvers.js
```diff
@@ -3,20 +3,13 @@
 /**
  * Internal dependencies
  */
-import { NAMESPACE } from '../constants';
 import { updateSettingsForGroup, updateErrorForGroup } from './actions';
+import { getAllSettings } from '@woocommerce/wc-admin-settings';
```
nerrad: I think I mentioned this elsewhere, but this tightly couples the data store to the server provided values and that isn't a good idea.  You should not be hydrating directly in the stores but hydrate via app initialization.  Removing this coupling also removes the need for the resolver.  

So I'd personally restore the HYDRATE_STORE action and use that in app initialization.

packages/data/src/settings/resolvers.js
```diff
@@ -3,20 +3,13 @@
...
 
-/**
- * External dependencies
- */
-import { apiFetch } from '@wordpress/data-controls';
-
-export function* getSettings( group ) {
-	const url = `${ NAMESPACE }/settings/${ group }`;
+export function* getSetting( group ) {
```
nerrad: So with this change, what this means is if the `getSettings` (note plural) selector is called before `getSetting` selector, it will return an empty object because the state will not have been hydrated yet.  Which is another reason why if you're not querying the server for the data, just eliminate the resolver and hydrate the state during app initialization (without coupling to the server provided data).

nerrad: Another problem with this resolver, is that resolvers must have the same arguments as the sibling selector otherwise resolution state will not be updated correctly.

packages/data/src/settings/selectors.js
```diff
@@ -26,6 +26,31 @@ export const getSettings = ( state, group ) => {
...
+ *
+ * @returns {mixed}  The value present in the settings state for the given
+ *                   name.
+ */
+export function getSetting( state, group, name, fallback = false, filter = val => val ) {
+	const resourceName = getResourceName( group, name );
+	const value = ( state[ resourceName ] && state[ resourceName ].data ) || fallback;
+	return filter( value, fallback );
+}
+
```
nerrad: I like having a selector for a specific setting 👍

packages/data/src/settings/withSettings.js
```diff
@@ -0,0 +1,41 @@
...
+import { useDispatch, useSelect } from '@wordpress/data';
+import { createHigherOrderComponent } from '@wordpress/compose';
+import { partial } from 'lodash';
+
+/**
+ * Internal dependencies
+ */
+import { STORE_NAME } from './constants';
+
+export const withSettings = group =>
```
nerrad: The HOC is a good idea.  However I would create a `useSettings` custom hook with the necessary logic so it can be used directly in functional components vs wrapping with the hook (which keep the tree levels more flat).  Then you can still have a HOC for places where you want to avoid refactoring a `React.Component` but the hoc would just use the new hook.

Also, as mentioned earlier, it's an anti-pattern to pass the actual selector through to components because then they aren't subscribed to state changes.  One change you could make here to make it easier for consumers is have the custom hook (and the new hoc suggested earlier) receive an optional set of arguments for what specific setting changes to select (like what you did for the curried function using `group` here.

psealock:
Thank you @nerrad for the excellent feedback. I made some changes and have some questions and thoughts.

### Changes

* I added `withHydration` HOC to tie hydration of the settings store to app initialization. This is a good idea, and better than using resolvers.
* I removed the `withSettings` HOC. Since its an anti-pattern to pass the selectors themselves instead of their results as you mention, it means the HOC would have to pass down the entire settings object, similar to how you had it originally. I agree with your sentiment that you should only get the data you're using, so I don't think the HOC makes sense.
* In place of the HOC, I changed the settings page to use `withSelect` and `withDispatch`. Any settings that are required by the component are passed as props.
* `getAllSettings` returns an amended `allSettings` object instead of creating a new object.

### Question

>Resolution state is only attached to selectors not actions. So this won't do anything. That's why I was dispatching the resolution state for the getSettings selector in the original code I added.

The resolution state that I had attached to the action was working. Thanks for the info. Now that its no longer in an HOC, its stopped working. When I dispatch the resolution state for the `getSettings` or `getSetting` selector instead it also does not work. What is the best way to determine if an action has finished? I'm not sure whats going on under the hood, but it doesn't make sense to me to use the name of a selector thats not being used during the course of the action. 

### Thoughts

I'd like to give a shot at making a custom hook `useSettings` in a follow up PR into your original PR #3313 once this one is merged. There is one hesitation I have in using it with the settings page component:

>You could also move `isDirty` and `persisting` state to the store as well to make that much cleaner and easier to reason with allowing the component to just listen to those states and act accordingly. IMO it would make the logic much easier to maintain and keep the component strictly presentational.

The store is responsible for gathering/updating settings and persisting changes to the server. If I understand correctly, you're suggesting adding a mechanism for also keeping track of a form's state. In other words, keep track of edits made to settings as well as a way to merge those edits at a later point. If edits are present, `isDirty` becomes truthy. This is actually a method used extensively in Store on wp.com and the reasoning is because there is a likelyhood of multiple users editing the same page.

In this case, I disagree with the premise that introducing a mechanism in the store for tracking a component's state is simpler. I think we can use `useState` in the functional component to accomplish the same thing. Am I understanding correctly? It seems fairly basic to reason about a component using local state to track local changes before persisting them, but I'm open to suggestions.

Thanks again for your thoughtful comments!

client/settings/fallbacks.js
```diff
@@ -54,17 +54,16 @@ export const WC_ASSET_URL = allSettings.wcAssetUrl;
...
+	allSettings.ADMIN_URL = ADMIN_URL;
+	allSettings.COUNTRIES = COUNTRIES;
+	allSettings.CURRENCY = CURRENCY;
+	allSettings.LOCALE = LOCALE;
+	allSettings.ORDER_STATUSES = ORDER_STATUSES;
+	allSettings.SITE_TITLE = SITE_TITLE;
+	allSettings.WC_ASSET_URL = WC_ASSET_URL;
+	allSettings.DEFAULT_DATE_RANGE = DEFAULT_DATE_RANGE;
+
+	return allSettings;
```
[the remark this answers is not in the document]
psealock: `import * as allSettings from '@woocommerce/wp-admin-settings';` won't get all settings. It will get constants such as `DEFAULT_DATE_RANGE` as well as methods such as `getSetting`. Properties like `max_columns` aren't exported. In order to remove this function, I'd have to change the index file to export the entire `SOURCE` object as well.

client/analytics/settings/index.js
```diff
@@ -204,12 +197,23 @@ class Settings extends Component {
 }
 
 export default compose(
-	withSettings( 'wc_admin' ),
+	withSelect( select => {
+		const { isResolving, getSetting, getLastSettingsErrorForGroup } = select( SETTINGS_STORE_NAME );
+		return {
+			isRequesting: isResolving( 'setSetting', [ 'wc_admin' ] ),
+			settings: getSetting( 'wc_admin', 'wcAdminSettings' ),
```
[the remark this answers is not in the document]
psealock: >Where is `wcAdminSettings` coming from?

Its an object on the `wcSettings` object. You can see it in the console:

```js
wcSettings.wcAdminSettings.woocommerce_actionable_order_statuses
// ["processing", "on-hold"]
```
`wcAdminSettings` is in fact the settings object that powers the settings page. 

I agree with you that it should be more like this:

```js
const actionableOrderStatuses = getSetting( 'wc_admin', 'woocommerce_actionable_order_statuses' );
```
but that doesn't work as its currently structured. Despite the redundant naming, I'd like to avoid manipulating how this object is stored because its used by onboarding. I could refactor the store to look for keys multiple levels deep, but I'd rather stick as closely to the current usage and revisit how this object is stored when I address how OBW interacts with the result of this work.

rrennick:
@nerrad Would you have time to give this another pass?

client/settings/fallbacks.js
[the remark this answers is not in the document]
nerrad: > import * as allSettings from '@woocommerce/wp-admin-settings'; won't get all settings. It will get constants such as DEFAULT_DATE_RANGE as well as methods such as getSetting.

I think we may have two different definitions of "all settings" here.  For me, the term "all settings" refers to everything the settings api exposes on the `wc.wcSettings`  api which includes a subset of specific properties exported on the `wcSettings` global converted to constants and `getSetting` and `setSetting` for everything else. 

It sounds like your definition of the phrase "all settings" refers to _everything_ exposed on `wcSettings` global from the server which includes any properties added via filter.

Is this correct?  If so, I still stand by what I originally wrote because the main rationale behind `getSetting` was a mechanism for retrieving the setting value for things that aren't exported on the _core_ `@woocommerce/settings` (or `wc.wcSettings`) api but instead are injected via filter. This provides a safe fallback for when/if that filter is removed or not present.  So not only do I not think it necessary to have something like a `getAllSettings` function but I think it is potentially fragile to have it.

> Properties like max_columns aren't exported.

Correct, because they are not intended to be exposed on the _core_ `@woocommerce/settings` api being that they are added via filter.  So to get `max_colums` you would do `getSetting( 'max_columns')`.

## Going Forward

However, with all the above said, I see that you are intentionally wanting to expose _all properties_ exposed on the `wcSettings` global (as opposed to `wc.wcSettings` api) because of the move to make the new `wc/admin/settings` data store the only source for settings after it's initialized.

With that in mind, here's what I suggest be done.

- don't define and export `getAllSettings` from the `fallbacks.js` file.  It is something _specific_ to wc-admin and will not be exposed by core (either via blocks or if/when the `@woocommerce/settings` api is eventually added to woo core).
- define and export `getAllSettings` from `client/settings/index.js` directly.

client/analytics/settings/index.js
```diff
@@ -158,113 +112,56 @@ class Settings extends Component {
...
-		const { settings } = this.state;
+		const nextSettings = { ...wcAdminSettings };
 
 		if ( 'checkbox' === type ) {
 			if ( checked ) {
-				settings[ name ].push( value );
+				nextSettings[ name ].push( value );
 			} else {
-				remove( settings[ name ], v => v === value );
+				remove( nextSettings[ name ], v => v === value );
```
nerrad: This mutates the original array that is found on the existing settings object _even though_ you cast the initial object to a new object (the values of the properties will still be the original values. So for the first block, you should probably do something like this:

```
nextSettings[ name ] = [ ...nextSettings[name], value ];
```

For the second block in the condition `remove` mutates the array, so you should use the native `Array.prototype.filter` function which returns a new array:

```js
nextSettings[name] = nextSettings[name].filter( ( v ) => v !== value )
```

This also has the advantage of completely removing lodash dependency here.

It's _very important_ to be watchful against any direct mutations of state because that can lead to unexpected, and sometimes hard to troubleshoot bugs. It's something that's easy to miss in multi-dimensional arrays/objects because of shallow cloning.

packages/data/src/settings/actions.js
```diff
@@ -68,11 +69,17 @@ export function* persistSettingsForGroup( group ) {
...
 
 	// get data slice for keys
-	const dirtyData = yield select( STORE_NAME, 'getSettingsForGroupSlice', group, dirtyKeys );
+	const dirtyData = yield select( STORE_NAME, 'getSettingsForGroup', group, dirtyKeys );
 	const url = `${ NAMESPACE }/settings/${ group }/batch`;
-	const update = dirtyKeys.map( key => {
-		return { id: key, value: dirtyData[ key ] };
-	} );
+
+	// todo: update should be an array of objects like { id: key, value: dirtyData[ key ] }.
```
nerrad: In the blocks repo we follow the convention of doing `@todo` for todos' I'm guessing that might be different here but just mentioning in case.

For this todo, do you think there's enough info to help with future work on this (if it's done)?  For example, would any future developer be able to derive why this isn't being done now in this reducer?

packages/data/src/settings/actions.js
```diff
@@ -68,11 +69,17 @@ export function* persistSettingsForGroup( group ) {
...
+	const update = dirtyKeys.reduce( ( updates, key ) => {
+		const u = Object.keys( dirtyData[ key ] ).map( k => {
+			return { id: k, value: dirtyData[ key ][ k ] };
+		} );
+		return concat( updates, u );
+		// return { id: key, value: dirtyData[ key ] };
```
nerrad: Is this just to help when the todo is implemented?

packages/data/src/settings/selectors.js
```diff
@@ -50,7 +50,7 @@ export const getSettingsForGroup = ( state, group, keys ) => {
 };
 
 export const getIsPersisting = ( state, group ) => {
-	return state[ group ] && Boolean( state[ group ].isPersisting );
+	return state[ group ] && Boolean( state[ group ].persisting );
```
nerrad: I realize that `persisting` is the property on the state. But since this is a boolean value shouldn't we just rename the property to `isPersisting` in the state (and make the necessary adjustments in the reducer as well). Convention is to have boolean properties represented by the `is` or `has` prefixes.

packages/data/src/settings/reducer.js
```diff
@@ -55,7 +64,8 @@ const receiveSettings = (
 						data: state[ group ] ? [ ...state[ group ].data, ...groupIds ] : groupIds,
```
nerrad: This should probably be:

```
data: state[ group ] && state[ group ].data ? 
```

packages/data/src/settings/reducer.js
```diff
@@ -55,7 +64,8 @@ const receiveSettings = (
...
 						error,
 						lastReceived: time,
-						persisting,
+						persisting: state[ group ] ? state[ group ].persisting : false,
```
nerrad: Good guards here but this will potentially add `undefined` to the `persisting` property instead of a boolean. So instead, what about doing:

```
persisting: state [ group ] ? Boolean( state[ group ].persisting ) : false,
```

packages/data/src/settings/reducer.js
```diff
@@ -55,7 +64,8 @@ const receiveSettings = (
...
+						dirty: state[ group ] ? state[ group ].dirty : [],
```
nerrad: This potentially sets the `dirty` property to undefined instead of to an empty array.  So instead, it should probably be something like:

```
dirty: state[ group ] && state[ group ].dirty ?
```

client/settings/fallbacks.js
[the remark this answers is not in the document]
psealock: >It sounds like your definition of the phrase "all settings" refers to everything exposed on wcSettings global from the server which includes any properties added via filter.

>Is this correct?

Yes, thanks for clarifying and I understand what you mean by:

> the main rationale behind getSetting was a mechanism for retrieving the setting value for things that aren't exported on the core @woocommerce/settings (or wc.wcSettings) api but instead are injected via filter.

The issue is I need a way to get all the data found in `window.wcSettings` into state when the app initializes, as you mentioned.

In `client/settings/index.js` I can define `getAllSettings` like below, but I'm not entirely sure of the consequences with blocks and its manipulation of settings, especially in regards to versions etc. Is this what you had in mind? 

```js
export function getAllSettings() {
    return window.wcSettings;
}
```

packages/data/src/settings/withHydration.js
```diff
@@ -0,0 +1,40 @@
+/** @format */
```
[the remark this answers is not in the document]
psealock: good idea. I did the same with `useSettings.js` to `use-settings.js` in https://github.com/woocommerce/woocommerce-admin/pull/3430/commits/5042b60a7585d4126c8f27e431e905307b412e6e

packages/data/src/settings/useSettings.js
```diff
@@ -24,12 +24,16 @@ export const useSettings = ( group, settingsKeys = [] ) => {
...
+		updateAndPersistSettingsForGroup,
+		updateSettingsForGroup,
+		setIsDirty,
+	} = useDispatch( STORE_NAME );
 	const updateSettings = useCallback(
 		( name, data ) => {
-			// this action would also mark the specific keys from this slice as dirty
-			// if they changed (see new action example)
 			updateSettingsForGroup( group, { [ name ]: data } );
+			setIsDirty( group, [ name ] );
```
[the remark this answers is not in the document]
psealock: I eliminated `setIsDirty` action in favour of just adding the dirty key in actions `UPDATE_SETTINGS_FOR_GROUP` or `UPDATE_ERROR_FOR_GROUP` in https://github.com/woocommerce/woocommerce-admin/pull/3430/commits/900602c94247441f5e3ce3fd3d66fb179aaa30da.

This means an action was needed to clear the dirty ids after persistence as well as initial hydration, so I introduced `clearIsDirty`.

client/analytics/settings/index.js
psealock: Nice catch Darren, that shallow cloning is indeed tricky.

packages/data/src/settings/actions.js
psealock: >do you think there's enough info to help with future work on this

Good question. No, it appears to me to be already addressed. It was leftover from the original copy/paste. I removed it.

packages/data/src/settings/selectors.js
psealock: persiting -> isPersisting in https://github.com/woocommerce/woocommerce-admin/pull/3430/commits/aaea4456c7c7000fca09f1bd0f3215666c57f2e4

psealock:
Thanks for the review @nerrad! I addressed your comments and have a remaining question for you about `getAllSettings`. Cheers 🌮

packages/data/src/settings/with-settings-hydration.js
```diff
@@ -0,0 +1,44 @@
...
+import { useSelect } from '@wordpress/data';
+
+/**
+ * Internal dependencies
+ */
+import { STORE_NAME } from './constants';
+
+export const withSettingsHydration = group => OriginalComponent => {
+	return props => {
+		const settings = useRef( getAllSettings() );
```
nerrad: I'm sorry I didn't catch this earlier, but I just realized that you are coupling the settings source (`getAllSettings()` from an alias) with an exported package.  

I think what would be better here is if the HOC receives both `group` and `settings` as it's arguments. That way any consumers (which in this case will be the wc-admin _client_) would just provide the settings object as the source for hydrating the internal data store. 

Either that, or:

- Add back in a resolver for the settings api endpoint to the data store.
- use `window.wcSettings` directly here as the source (instead of `getAllSettings`) and if it's undefined, you can allow the hydration to fallback to the resolver.

With all the work that's already gone into this pull, it's probably easier to just do the former option (provide `group` and `settings` as arguments to the `withSettingsHydration` hoc.)

client/settings/fallbacks.js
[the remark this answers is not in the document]
nerrad: Actually, right now the only purpose for the `@woocommerce/wc-admin-settings` alias is to provide backward compat for every place the original `wcSettings` global was being used and to connect with the settings api exposed in `@woocommerce/settings` (which may or may not be present depending on what version of Woo core wc-admin is active with).

Since with this pull and the subsequent refactors after this is merged across the wc-admin codebase to switch away from `@woocommerce/wc-admin-settings` to the new data api, you could just pass in `window.wcSettings` as the settings source to `withSettingsHydration`. 

After the refactors over to the new wp.data api are complete across the client application, you can then remove the alias `@woocommerce/wc-admin-settings` configuration and the related logic in `client/settings/`.

psealock:
Thanks for the suggestions @nerrad. I think including `window.wcSettings` as an argument to `withSettingsHydration` makes sense and decouples the new data stores from `@woocommerce/wc-admin-settings`. Changes made in https://github.com/woocommerce/woocommerce-admin/pull/3430/commits/ed3eaf8c802c230ddaddf29c0f47c47caccfa2c4.

Implementation now looks like below and I can remove `getAllSettings`.

```js
const HydratedPageLayout = withSettingsHydration( 'wc_admin', window.wcSettings )(
    PageLayout
);
render( <HydratedPageLayout />, appRoot );
```

PR #3433, opened 2019-12-17
update i18n scripts for making JSON files

### LINKED ISSUE #3298: Translations: wp i18n make-json produces no files
**Describe the bug**

The command `npm run i18n` creates a `.pot` file for translations. First, the js and then the php. The js gets overwritten by the php.

First reported in https://wordpress.org/support/topic/wordpress-i18n-how-to-translate-js-strings/

**To Reproduce**

1. `npm run i18n`
2. Check the resulting `languages/woocommerce-admin.pot` file for strings derived from js files, and see none exist.
3. Run just the js this time, `npm run -s i18n:js` and see the js strings available in the same file.

**Expected behavior**

I'm not an expert here, but it seems the js and php should be concatenated.

rrennick:
Fixes #3298

This PR addresses a number of issues with the existing `i18n` scripts and instructions. A`.pot` file is formatted with 3 sections per original string

- References: location(s) in the source where the string was found
- Original(s): the original untranslated string(s)
- Translation(s): one or more translations depending on the translation function called

```
#: includes/connect-existing-pages.php:33 src/Features/Analytics.php:142
#: client/header/activity-panel/panels/orders.js:198
#: client/layout/controller.js:64
msgid "Settings"
msgstr "
```

## Issues addressed

1. The `pot-to-php` script converted the `.js` pot file to a PHP file. The final `woocommerce-admin.pot` file contained no `.js` references for `wp i18n make-json` resulting in the command reporting success while creating no `.json` files.
1. The `npm run i18n:js` command only scanned the `client` folder so the strings in the `packages` folder weren't included in the translation file.
1. The `js` pot file contains source references (which are useful for translators) while the `json` files are compiled based on the registered script handles of the `dist/*` files. Those references needed to be updated after translation but before making the `json` files.

- Run `npm run i18n`
- Edit `languages/woocommerce-admin.pot` and add translations for a few strings that have both `.php` and `.js` references
- Save the file as `woocommerce-admin-xx_YY.po` where `xx_YY` is a locale other than `en_US`
- Run `npm run i18n:json`
- Copy `.mo` and `.json` files to the `wp-content/languages/plugins` folder
- Change your locale to match the translation
- Check the dashboard for the translated version of those strings using the original file references to determine where the strings should appear
- Review changes to updated readme files

### Changelog Note:

Dev: Update translation scripts and instructions for creating `.json` files.


CONTRIBUTING.md
```diff
@@ -20,10 +20,11 @@ Language packs are automatically generated once 95% of the plugin's strings are
 
 ### Testing translations in development without language packs
 
-1. Export the translations strings as a `.po` file from the [translate.wordpress.org project](https://translate.wordpress.org/projects/wp-plugins/woocommerce-admin).
-1. Rename `.po` file to the following format: `woocommerce-admin-xx_YY.po` (Where xx_YY is your locale, like it_IT)
-1. Use WP-CLI to generate needed JSON files for JavaScript-based strings: `wp i18n make-json [.po file here] --no-purge`
-1. Move `.po` and `.json` files to `/wp-content/languages/plugins`.
+1. Generate a translation file with `npm run i18n`.
+1. Rename `woocommerce-admin.pot` file to the following format: `woocommerce-admin-xx_YY.po` (Where xx_YY is your locale, like it_IT).
```
joshuatf: Optional: if we're just using this to create a new language file, it would be a nice touch to allow an argument to be passed here to generate the file.  E.g., `npm run 18n --lang=it_IT`

package.json
```diff
@@ -20,18 +20,19 @@
...
     "build:feature-config": "php bin/generate-feature-config.php",
     "build:packages": "node ./bin/packages/build.js",
     "build:release": "./bin/build-plugin-zip.sh",
     "clean": "rimraf ./dist ./packages/*/build ./packages/*/build-module ./packages/*/build-style",
     "predev": "npm run -s install-if-deps-outdated && php ./bin/update-version.php",
     "dev": "cross-env WC_ADMIN_PHASE=development npm run build:feature-config && cross-env WC_ADMIN_PHASE=development npm run build:packages && cross-env WC_ADMIN_PHASE=development webpack",
     "docs": "node ./bin/generate-docs",
-    "i18n": "npm run -s i18n:js && npm run -s i18n:php && npm run -s i18n:check && npm run -s i18n:pot",
+    "i18n": "npm run -s i18n:js && npm run -s i18n:check && npm run -s i18n:pot && npm run -s i18n:build",
+    "i18n:build": "php bin/combine-pot-files.php languages/woocommerce-admin.po languages/woocommerce-admin.pot",
```
joshuatf: I'm not quite following on why the po and pot files need to be combined here.  Could you give me a quick rundown of why we do this?

rrennick: > Could you give me a quick rundown of why we do this?

Each tool generates a complete file including a header. If the same output file name is used, both tools overwrite an existing file. Some strings exist in both files. If you switch to master and run `npm run i18n` then edit the .pot file you'll see references like `#: languages/woocommerce-admin.php:123`. Those originate from the `npm run i18n:js`.

We could just remove one of the headers and concatenate the files. But, if we do that then any string that is in both JS & PHP appears in the file for translation twice.

CONTRIBUTING.md
rrennick: > it would be a nice touch to allow an argument to be passed here to generate the file.

Fantastic idea. When I updated the readme for this I did not note that is was optional because explaining both paths might make the instructions more confusing.

languages/README.md
```diff
@@ -8,7 +8,15 @@ If you'd like to contribute a translation, please follow the Localizing section
 The generated POT template file is not included in this repository. To create this file locally, follow instructions from [README.md](https://github.com/woocommerce/woocommerce-admin/blob/master/README.md) to install the project, then run the following command:
 
 ```
-npm run build
+npm run i18n -- lang=xx_YY
```
joshuatf: Small update here: the `--` can be removed.

package.json
joshuatf: That makes sense.  Out of curiosity, how will translations be managed going forward?  It seems like any changes to strings will require a new po file and any previous translations will be overwritten.  But maybe this is best left to translation editors like Poedit.

bin/combine-pot-files.php
```diff
@@ -0,0 +1,104 @@
...
+	if ( ! is_file( $argv[ $index ] ) ) {
+		echo "File not found: {$argv[ $index ]}\n";
+		exit;
+	}
+}
+
+/**
+ * Check whether an output locale has been requested.
+ */
+if ( isset( $argv[3] ) && 0 === stripos( $argv[3], 'lang=' ) ) {
```
joshuatf: Nice 👍

languages/README.md
rrennick: > I'm also having trouble loading the mo files

The mo files are created with the translation tool (poEdit, GlotPress, etc.).

bin/combine-pot-files.php
rrennick: > Out of curiosity, how will translations be managed going forward?

If the translator is using GlotPress then the existing strings & translations are stored in the DB and there is a feature for updating to a new version of the strings. 

For other tools I suspect the translators copy the source file to another location since translation files are always replaced in plugin/theme updates.

languages/README.md
rrennick: @joshuatf I removed the `--` and updated the instructions in `CONTRIBUTING.md` to be consistent.

joshuatf: @rrennick Sorry if my issue wasn't clear.  I've already created the mo files and moved them to `wp-content/languages/plugins` per testing instructions but they don't seem to be having any effect.

When checking the loaded text domains using the plugin mentioned above, I don't see `woocommerce-admin` come up in the list.  Wondering if something is wrong with my environment or if you're able to recreate the issue.

rrennick: @joshuatf It is working for me
```
~/Sites/ron/wp-content/plugins/woocommerce-admin/languages
$ ls
README.md			woocommerce-admin-en_CA.po
~/Sites/ron/wp-content/plugins/woocommerce-admin/languages
$ ls ~/Sites/ron/wp-content/languages/plugins/woocommerce-admin*
/Sites/ron/wp-content/languages/plugins/woocommerce-admin-en_CA-2459417b12e8fe2d482313c6d8f9819a.json
/Sites/ron/wp-content/languages/plugins/woocommerce-admin-en_CA-27b105c531d6e08bd8855676f3a6164e.json
/Sites/ron/wp-content/languages/plugins/woocommerce-admin-en_CA-5c8a66f908738b94586c976c499edfe5.json
/Sites/ron/wp-content/languages/plugins/woocommerce-admin-en_CA-7bc4e52d9439cde288e4fdb2e16c0e61.json
/Sites/ron/wp-content/languages/plugins/woocommerce-admin-en_CA-946ffa241f2f7a67d85b6416fd6b9f50.json
/Sites/ron/wp-content/languages/plugins/woocommerce-admin-en_CA-bb2b82fbdf1b425e54cdec46f60a7f9b.json
/Sites/ron/wp-content/languages/plugins/woocommerce-admin-en_CA-cacb17be7bdb63bd3bb9c23d5a2305d7.json
/Sites/ron/wp-content/languages/plugins/woocommerce-admin-en_CA-d5abdf7f23e9521b725160dbd08295f1.json
/Sites/ron/wp-content/languages/plugins/woocommerce-admin-en_CA-eddaf6c0602832520b4958e0a7d25e59.json
/Sites/ron/wp-content/languages/plugins/woocommerce-admin-en_CA-f5fad241b4f471337325962233df4cd4.json
/Sites/ron/wp-content/languages/plugins/woocommerce-admin-en_CA.mo
```

Screen Shot 2019-12-20 at 9 17 38 AM

While investigating this I found lines like
```
#: src/API/Leaderboards.php:321 client/analytics/report/downloads/config.js:49
```
in the .po file exported from GlotPress so I'll need to modify the replacement script to account for that patter.

rrennick: > I'll need to modify the replacement script to account for that pattern.

Done in `166294f6`

PR #3434, opened 2019-12-17
Allow report caching layer to be filtered off.

### LINKED ISSUE #3072: Unit test last query outputs are now last caching query
**Describe the bug**
As a result of #2970, unit tests that output the last DB query either output the data store caching query

```
query: INSERT INTO `wptests_options` (`option_name`, `option_value`, `autoload`) VALUES ( ...
```

when the DS had to retrieve the data from the database. If the query results were retrieved from cache, then a related query is output.

**Expected behavior**
Alternatives to consider

- remove the query output
- add a method of disabling caching, use that in most unit tests while leaving enough tests to test caching
- add a method of optionally saving the last query for output when a unit test fails

jeffstieler:
Fixes #3072.

This PR seeks to restore the last query output in unit test failures for the Order Stats tests.

It does so by adding a filter to allow report caching to be conditionally disabled. It introduces the `woocommerce_analytics_report_should_use_cache` filter, which short circuits caching (by returning `false`).

### Changelog Note:

Enhancement: allow report cache layer to be turned off.

### DIFF
--- a/src/API/Reports/DataStore.php
+++ b/src/API/Reports/DataStore.php
@@ -120,6 +120,23 @@ class DataStore extends SqlQuery {
 	}
 
 	/**
+	 * Whether or not the report should use the caching layer.
+	 *
+	 * Provides an opportunity for plugins to prevent reports from using cache.
+	 *
+	 * @return boolean Whether or not to utilize caching.
+	 */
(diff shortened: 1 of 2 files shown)


timmyc:
Nice one Jeff!

PR #3444, opened 2019-12-18
Add purchase modal at the end of the profile wizard

### LINKED ISSUE #3441: Onboarding - Implement purchase modal in the profile wizard
This issue involves implementing a modal at the end of the profile wizard (after theme selection) that asks the user if they would like to purchase their products or not.

Screen Shot 2019-12-18 at 8 27 35 AM

Selecting "I'll do it later" should redirect the user to the task list. Selecting purchase now should lead them into the WooCommerce.com in-app purchase / extension flow.

The modal should be built in a way that it can be embedded elsewhere.

justinshreve:
Closes #3441.

This PR adds a modal at the end of the profile wizard that allows the user to confirm if they would like to purchase products now, or later (via the task list). I've tried to build this in a way that the modal can easily be used from the task as well.

It also fixes the intermittent bug we were seeing at the end of the theme step where the app would get stuck in a `updateProfileItems` loop.

One thing that I think needs further follow-up, is selecting Storefront or one of the other free themes still shows the dialog and redirects you to the cart (`master` also currently adds these items to the WC.com cart). I think we'll want to install the free themes from WordPress.org instead. cc @joshuatf.

### Screenshots

Screen Shot 2019-12-18 at 3 15 09 PM

* Enable the profile wizard
* Make a product type selection or theme selection that requires purchase
* Finish the theme step, and see the modal appear
* Walk through the profile wizard again and make selections that don't require purchase
* No modal should be displayed

### DIFF
--- a/client/dashboard/index.js
+++ b/client/dashboard/index.js
@@ -2,16 +2,9 @@
 /**
  * External dependencies
  */
-import { addQueryArgs } from '@wordpress/url';
 import { Component } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import { get } from 'lodash';
-
-/**
- * WooCommerce dependencies
- */
-import { getSetting } from '@woocommerce/wc-admin-settings';
-import { getNewPath, updateQueryString } from '@woocommerce/navigation';
 
 /**
  * Internal dependencies
@@ -19,74 +12,42 @@ import { getNewPath, updateQueryString } from '@woocommerce/navigation';
 import './style.scss';
 import CustomizableDashboard from './customizable';
 import ProfileWizard from './profile-wizard';
+import CartModal from './profile-wizard/cart-modal';
 import withSelect from 'wc-api/with-select';
-import { isOnboardingEnabled } from 'dashboard/utils';
+import { isOnboardingEnabled, getProductIdsForCart } from 'dashboard/utils';
 
 class Dashboard extends Component {
+	constructor( props ) {
+		super( props );
+		this.state = {
+			showCartModal: false,
+		};
+	}
 	componentDidUpdate( prevProps ) {
 		const profileItems = get( this.props, 'profileItems', {} );
 		const prevProfileItems = get( prevProps, 'profileItems', {} );
 
+		// We only want to show the modal during the transition between the profile wizard and dashboard and if products were selected.
 		if ( profileItems.completed && ! prevProfileItems.completed ) {
-			updateQueryString( {}, '/', {} );
-			this.redirectToCart();
-		}
-	}
-
-	getProductIds() {
-		const productIds = [];
-		const profileItems = get( this.props, 'profileItems', {} );
-		const onboarding = getSetting( 'onboarding', {} );
-		const productTypes = profileItems.product_types || [];
-
-		productTypes.forEach( productType => {
-			if (
-				onboarding.productTypes[ productType ] &&
-				onboarding.productTypes[ productType ].product
-			) {
-				productIds.push( onboarding.productTypes[ productType ].product );
+			const productIds = getProductIdsForCart( profileItems );
+			if ( productIds.length ) {
+				/* eslint-disable react/no-did-update-set-state */
+				this.setState( { showCartModal: true } );
+				/* eslint-enable react/nod-ddi-update-set-state */
 			}
-		} );
-
-		const theme = onboarding.themes.find( themeData => themeData.slug === profileItems.theme );
-
-		if ( theme && theme.id && ! theme.is_installed ) {
-			productIds.push( theme.id );
 		}
-
-		return productIds;
-	}
-
-	redirectToCart() {
-		if ( ! isOnboardingEnabled() ) {
-			return;
-		}
-
-		const productIds = this.getProductIds();
-		const backPath = getNewPath( {}, '/', {} );
-		const { connectNonce } = getSetting( 'onboarding', {} );
-
-		if ( ! productIds.length ) {
-			return;
-		}
-
-		document.body.classList.add( 'woocommerce-admin-is-loading' );
-
-		const url = addQueryArgs( 'https://woocommerce.com/cart', {
-			'wccom-site': getSetting( 'siteUrl' ),
-			'wccom-woo-version': getSetting( 'wcVersion' ),
-			'wccom-replace-with': productIds.join( ',' ),
-			'wccom-connect-nonce': connectNonce,
-			'wccom-back': backPath,
-		} );
-		window.location = url;
 	}
 
 	render() {
 		const { path, profileItems, query } = this.props;
+		const { showCartModal } = this.state;
 
 		if ( isOnboardingEnabled() && ! profileItems.completed ) {
 			return <ProfileWizard query={ query } />;
+		}
+
+		if ( isOnboardingEnabled() && showCartModal ) {
+			return <CartModal onClose={ () => this.setState( { showCartModal: false } ) } />;
 		}
 
 		if ( window.wcAdminFeatures[ 'analytics-dashboard/customizable' ] ) {
--- a/client/dashboard/index.js
+++ b/client/dashboard/index.js
@@ -4,7 +4,6 @@
  */
 import { Component } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
-import { get } from 'lodash';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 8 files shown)


client/dashboard/utils.js
```diff
@@ -36,6 +36,35 @@ export function getCurrencyRegion( countryState ) {
 	return region;
 }
 
+/**
+ * Gets the product IDs for items selected for purchase.
```
joshuatf: Nice - thanks for moving this 👍

Small nitpick: might update this description to read:

```suggestion
 * Gets the product IDs for items based on the product types and theme selected in the onboarding profiler.
```

client/dashboard/profile-wizard/cart-modal.js
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	render() {
+		return (
+			<Modal
+				title={ __(
+					'Would you like to purchase and install the following features now?',
+					'woocommerce-admin'
+				) }
+				onRequestClose={ () => this.props.onClose() }
```
joshuatf: Not a big deal and not sure if there's a way around this, but this will also close when clicking outside the modal; which I would expect to return to the theme selection if it was in the background.

client/dashboard/style.scss
```diff
@@ -179,3 +179,44 @@
 		/* stylelint-enable */
 	}
 }
+
+.components-modal__frame.woocommerce-cart-modal {
```
joshuatf: There a few minor differences in heading and p tag sizes between this and the design.

#### Current
Screen Shot 2019-12-19 at 2 15 32 PM


#### Design
Screen Shot 2019-12-19 at 2 21 14 PM

client/dashboard/index.js
```diff
@@ -2,93 +2,54 @@
...
 
 	render() {
 		const { path, profileItems, query } = this.props;
+		const { showCartModal } = this.state;
 
 		if ( isOnboardingEnabled() && ! profileItems.completed ) {
 			return <ProfileWizard query={ query } />;
 		}
 
+		if ( isOnboardingEnabled() && showCartModal ) {
```
joshuatf: The only issue with separating the modal from the profile wizard here is the theme previews can't be shown in the background.  May not be super important though; /cc @jameskoster in case you have a strong opinion one way or the other.

#### Current
Screen Shot 2019-12-19 at 2 17 54 PM


#### Design
Screen Shot 2019-12-19 at 2 23 45 PM

client/dashboard/profile-wizard/cart-modal.js
```diff
@@ -0,0 +1,134 @@
...
+					title: productInfo.label,
+					content: productInfo.description,
+				} );
+			}
+
+			const themeInfo = find( themes, theme => {
+				return theme.id === productId;
+			} );
+
+			if ( themeInfo && themeInfo.slug !== 'Storefront' ) {
```
joshuatf: Not opposed to separating Storefront and installing from the wp.org repo, but if we do that we may also want to check if this is the only product ID.  Otherwise, we should probably Storefront in this list since it will be added via wccom.

On a similar note, would we want to install any free themes from wp.org or explicitly set any that should bypass wccom?  If the former, we could check against price here to allow any themes that are free to be installed without visiting wccom.

client/dashboard/profile-wizard/cart-modal.js
```diff
@@ -0,0 +1,134 @@
...
+						'woocommerce-admin'
+					) }
+				</p>
+
+				<div className="woocommerce-cart-modal__actions">
+					<Button isLink onClick={ () => this.props.onClose() }>
+						{ __( "I'll do it later", 'woocommerce-admin' ) }
+					</Button>
+
+					<Button isPrimary isDefault onClick={ () => this.redirectToCart() }>
```
joshuatf: Optional: might be a nice addition for slower connections if we add a loading state when this is clicked.

client/dashboard/index.js
justinshreve: I'm moving some things around to make this possible.

client/dashboard/profile-wizard/cart-modal.js
justinshreve: I think we should probably install any of the free themes from .org instead of WooCommerce.com. It saves multiple steps for the user if they didn't pick any other products, and it's less confusing than sending them to a cart with a $0.00 product.

It looks like the other free themes (like Deli) are in the .org repo.

What do you think about this being done in a follow-up? I think a couple pieces will need to be done. I can back out the `Storefront` check here and add a todo / log an issue.

client/dashboard/utils.js
justinshreve: Fixed in https://github.com/woocommerce/woocommerce-admin/pull/3444/commits/d7762abe8ad6717ef3db01792f1b1fe9f3584906.

client/dashboard/profile-wizard/cart-modal.js
justinshreve: I moved some logic around, so now this opens in front of the theme screen. You can now close the modal and reopen it.

https://github.com/woocommerce/woocommerce-admin/pull/3444/commits/d7762abe8ad6717ef3db01792f1b1fe9f3584906

client/dashboard/style.scss
justinshreve: Fixed in https://github.com/woocommerce/woocommerce-admin/pull/3444/commits/d7762abe8ad6717ef3db01792f1b1fe9f3584906.

client/dashboard/index.js
justinshreve: I have to move some things around to make this work (namely mark the wizard as completed at a different point) but this now opens in front of the theme step.

https://github.com/woocommerce/woocommerce-admin/pull/3444/commits/d7762abe8ad6717ef3db01792f1b1fe9f3584906

client/dashboard/profile-wizard/cart-modal.js
justinshreve: Added in https://github.com/woocommerce/woocommerce-admin/pull/3444/commits/d7762abe8ad6717ef3db01792f1b1fe9f3584906.

justinshreve:
Hey @joshuatf I addressed your feedback in https://github.com/woocommerce/woocommerce-admin/pull/3444/commits/d7762abe8ad6717ef3db01792f1b1fe9f3584906 -- except for the free themes (which I replied to [here](https://github.com/woocommerce/woocommerce-admin/pull/3444#discussion_r359921124)).

client/dashboard/profile-wizard/cart-modal.js
joshuatf: I think that's a good plan.  Outlined an issue for it [here](https://github.com/woocommerce/woocommerce-admin/issues/3455).

client/dashboard/profile-wizard/cart-modal.js
```diff
@@ -102,19 +107,19 @@ class CartModal extends Component {
...
 				</p>
 
 				<div className="woocommerce-cart-modal__actions">
-					<Button isLink onClick={ () => this.props.onClose() }>
+					<Button isLink onClick={ () => this.props.onClickLater() }>
 						{ __( "I'll do it later", 'woocommerce-admin' ) }
 					</Button>
 
-					<Button isPrimary isDefault onClick={ () => this.redirectToCart() }>
+					<Button isPrimary isDefault isBusy={ isRedirecting } onClick={ () => this.onClickNow() }>
```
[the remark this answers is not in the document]
justinshreve: Renamed to `onClickPurchaseNow` in https://github.com/woocommerce/woocommerce-admin/pull/3444/commits/c4f2866960f415f07f174de0897d379e715435de.

client/dashboard/profile-wizard/cart-modal.js
```diff
@@ -102,19 +107,19 @@ class CartModal extends Component {
...
+				<p className="woocommerce-cart-modal__help-text">
 					{ __(
 						"You won't have access to this functionality until the extensions have been purchased and installed.",
 						'woocommerce-admin'
 					) }
 				</p>
 
 				<div className="woocommerce-cart-modal__actions">
-					<Button isLink onClick={ () => this.props.onClose() }>
+					<Button isLink onClick={ () => this.props.onClickLater() }>
```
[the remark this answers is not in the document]
justinshreve: Renamed to `onClickPurchaseLater` and added a busy state in https://github.com/woocommerce/woocommerce-admin/pull/3444/commits/c4f2866960f415f07f174de0897d379e715435de.

client/dashboard/profile-wizard/index.js
```diff
@@ -25,17 +26,31 @@ import Start from './steps/start';
...
 import withSelect from 'wc-api/with-select';
+import { getProductIdsForCart } from 'dashboard/utils';
 import './style.scss';
 
 class ProfileWizard extends Component {
 	constructor() {
 		super( ...arguments );
+		this.state = {
+			showCartModal: false,
+			cartRedirect: false,
```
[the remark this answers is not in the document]
justinshreve: Renamed in https://github.com/woocommerce/woocommerce-admin/pull/3444/commits/c4f2866960f415f07f174de0897d379e715435de.

PR #3465, opened 2019-12-20
Invalidate cache on settings change

### LINKED ISSUE #3431: DataStore Cache: Should be invalidated after settings change
**Describe the bug**

When a setting is changed, data requests include the same queries, so cached results reflecting the previous settings are returned.

**To Reproduce**

1. Go to Orders Report and note a status that you see.
2. Add that status as an excluded status on the Settings page.
3. Return to the Orders Report and refresh. You will see that status still being included.
4. `wp transient delete woocommerce_reports-transient-version`
5. Refresh and see that status correctly being excluded.

**Expected behavior**

A change in settings should invalidate caches.

jeffstieler:
Fixes #3431 

This PR seeks to invalidate the current cache version number when Analytics Settings are updated.

Without doing so, changes in excluded order statuses won't be reflected on already-cached reports.

1. Go to Orders Report and note a status that you see.
1. Add that status as an excluded status on the Settings page.
1. Return to the Orders Report and refresh. You will see that status correctly being excluded.

### Changelog Note:

Fix: invalidate Reports cache when changing Analytics settings.

### DIFF
--- a/src/API/SettingOptions.php
+++ b/src/API/SettingOptions.php
@@ -11,6 +11,8 @@ namespace Automattic\WooCommerce\Admin\API;
 
 defined( 'ABSPATH' ) || exit;
 
+use \Automattic\WooCommerce\Admin\API\Reports\Cache as ReportsCache;
+
 /**
  * Setting Options controller.
  *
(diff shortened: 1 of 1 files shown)


client/header/activity-panel/panels/reviews.js
```diff
@@ -162,7 +162,7 @@ class ReviewsPanel extends Component {
 					</Fragment>
 				);
 			} else {
-				buttonUrl = getAdminLink( 'edit-comments.php' );
+				buttonUrl = getAdminLink( 'edit-comments.php?comment_type=review' );
```
timmyc: Eh? Any reason why this change is in this PR?

jeffstieler: Totally a mistake. I'll fix.

PR #3471, opened 2019-12-23
Onboarding: Add another option to platforms.

### LINKED ISSUE #3449: Onboarding - store profiler: add an additional option in the Currently selling elsewhere question
We should add the following option in Business Details - Currently selling elsewhere?, in order to try to understand if we have users going through the onboarding that already own a different WC store:

Yes, I own a different store powered by WooCommerce

Screen Shot 2019-12-18 at 18 09 46

timmyc:
Fixes #3449

This change adds another option to the business details step of the OBW which allows users to select that they have another WooCommerce store:

Dashboard ‹ bend outdoors — WooCommerce 2019-12-23 10-40-22

__To Test__
- If you have yet to do so, opt-in to the new Site Profiler / OBW via the "Help" menu in WooCommerce Orders. While there you can also launch the Site Profiler.
- On the business profile page, verify the new option for other platforms is shown as seen above

### DIFF
--- a/client/dashboard/profile-wizard/steps/business-details.js
+++ b/client/dashboard/profile-wizard/steps/business-details.js
@@ -370,6 +370,10 @@ class BusinessDetails extends Component {
 				label: __( 'Yes, on another platform', 'woocommerce-admin' ),
 			},
 			{
+				key: 'other-woocommerce',
+				label: __( 'Yes, I own a different store powered by WooCommerce', 'woocommerce-admin' ),
+			},
+			{
 				key: 'brick-mortar',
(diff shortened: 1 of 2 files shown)


timmyc:
> Nice! Option looks good. We may want to also check annual revenue if this option is selected, but pre-approving.

Oh good point Josh - @pmcpinto any opinion here?

pmcpinto:
> Oh good point Josh - @pmcpinto any opinion here?

I think we can keep the annual revenue question. It can useful to see if they're opening a new store because they already have a striving store and they're expanding, or the existing store doesn't generate enough revenue and they're trying to set up a different business, etc.

timmyc:
@joshuatf update applied in f68da67 do you mind doing one last 👀

Issue #3455, opened 2019-12-20
Onboarding: Install free themes from dotorg


joshuatf:
When the last step of the profiler is reached, the user has the option to select a theme.  When selecting a free theme (currently limited to StoreFront), the theme is added to the cart and installed via wccom.

However, if a free theme is chosen, we should not display it in the cart and install from WordPress.org.

#### Steps
1. Visit `wp-admin/admin.php?page=wc-admin&step=theme`
2. Select a free theme.
3. Note the free theme is in the cart items list and only installed if going through the checkout process.

#### Screenshots
<img width="638" alt="Screen Shot 2019-12-20 at 11 31 19 AM" src="https://user-images.githubusercontent.com/10561050/71228150-5c21d900-231c-11ea-8a84-8778f084febf.png">
<img width="1021" alt="Screen Shot 2019-12-20 at 11 31 09 AM" src="https://user-images.githubusercontent.com/10561050/71228151-5c21d900-231c-11ea-8bd2-01434cabbc2e.png">

joshuatf:
Note that this is a follow-up to https://github.com/woocommerce/woocommerce-admin/pull/3444 and should branch from there or wait until it's merged to master.

justinshreve:
> When selecting a free theme (currently limited to StoreFront)

https://woocommerce.com/products/deli/ and https://woocommerce.com/products/boutique/ are both free as well. They are both luckily present in the .org repository too.

https://wordpress.org/themes/deli/
https://wordpress.org/themes/boutique/

joshuatf:
Thanks for noting those, @justinshreve!  Forgot they were in there.

PR #3482, opened 2019-12-25
Onboarding: Add theme install and activation endpoints

joshuatf:
Fixes (part of) #3455

This PR introduces 2 new endpoints for installing and activating allowed themes.

1.  Post to `wc-admin/onboarding/themes/install?theme={theme_slug}` with one of the following theme slugs:
```
Array
(
    [0] => twentynineteen // this will be your currently active theme slug
    [1] => storefront
    [2] => deli
    [3] => boutique
    [4] => navolio-light
)
```
2. Check that the theme was installed if not already installed and that a success message is returned.
3. Attempt the above again with an invalid theme slug.
4. Post to `wc-admin/onboarding/themes/activate?theme={theme_slug}` endpoint with a valid and already installed theme slug.
5. Check that the theme was activated.
6. Repeat from step 4 with an invalid slug and check that an error is returned.

### DIFF
--- a/src/API/OnboardingThemes.php
+++ b/src/API/OnboardingThemes.php
@@ -1,0 +1,177 @@
+<?php
+/**
+ * REST API Onboarding Themes Controller
+ *
+ * Handles requests to install and activate themes.
+ *
+ * @package WooCommerce Admin/API
+ */
+
+namespace Automattic\WooCommerce\Admin\API;
+
+use Automattic\WooCommerce\Admin\Features\Onboarding;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Onboarding Themes Controller.
+ *
+ * @package WooCommerce Admin/API
+ * @extends WC_REST_Data_Controller
+ */
+class OnboardingThemes extends \WC_REST_Data_Controller {
+	/**
+	 * Endpoint namespace.
+	 *
+	 * @var string
+	 */
+	protected $namespace = 'wc-admin';
+
+	/**
+	 * Route base.
+	 *
+	 * @var string
+	 */
+	protected $rest_base = 'onboarding/themes';
+
+	/**
+	 * Register routes.
+	 */
+	public function register_routes() {
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base . '/install',
+			array(
+				array(
+					'methods'             => \WP_REST_Server::EDITABLE,
+					'callback'            => array( $this, 'install_theme' ),
+					'permission_callback' => array( $this, 'update_item_permissions_check' ),
+				),
+				'schema' => array( $this, 'get_item_schema' ),
+			)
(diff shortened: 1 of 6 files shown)


src/API/OnboardingThemes.php
```diff
@@ -0,0 +1,223 @@
...
+	 *
+	 * @param  WP_REST_Request $request Full details about the request.
+	 * @return WP_Error|array Theme installation status.
+	 */
+	public function install_theme( $request ) {
+		$allowed_themes = Onboarding::get_allowed_themes();
+		$theme          = sanitize_title_with_dashes( $request['theme'] );
+
+		if ( ! in_array( $theme, $allowed_themes, true ) ) {
+			return new \WP_Error( 'woocommerce_rest_invalid_theme', __( 'Invalid theme.', 'woocommerce-admin' ), 404 );
```
timmyc: Not sure if 404 is the best code to use here since this is a valid endpoint/request, perhaps return a WP_Error / 500 instead?

src/API/OnboardingThemes.php
```diff
@@ -0,0 +1,223 @@
...
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Onboarding Themes Controller.
+ *
+ * @package WooCommerce Admin/API
+ * @extends WC_REST_Data_Controller
+ */
+class OnboardingThemes extends \WC_REST_Data_Controller {
```
timmyc: Can we add some test coverage for these new endpoints? Perhaps at least some assertions around the permission callbacks.

joshuatf: Good idea!  Added tests for theme install and activation as well as the previous plugin API while I was there.

src/API/OnboardingThemes.php
joshuatf: I think `404` might still be the best here since `500` might indicate something broken on the server and not an error due to an invalid client query.

A `404` doesn't necessarily mean that the endpoint is not found, but that a [resource isn't found](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_Client_errors).  (In this case, the theme).

joshuatf:
Added some test coverage around themes and plugins per your feedback; thanks for the review @timmyc!

tests/api/onboarding-plugins.php
```diff
@@ -0,0 +1,123 @@
...
+	 */
+	public function test_install_without_permission() {
+		$response = $this->server->dispatch( new WP_REST_Request( 'POST', $this->endpoint . '/install' ) );
+		$this->assertEquals( 401, $response->get_status() );
+	}
+
+	/**
+	 * Test that installing a valid plugin works.
+	 */
+	public function test_install_plugin() {
```
timmyc: 🌟 nice thanks for adding these in too.

PR #3490, opened 2019-12-27
Onboarding: Add Skip Step link on Themes Step

### LINKED ISSUE #3460: Onboarding - store profiler: unable to proceed in the Theme selector step with an active theme
It seems that in some cases the users are not able to proceed in the Theme selector step when they already have an active theme using a Jurassic ninja site. I wasn't able to replicate this error. More info here pb22l9-5T-p2 (comment 236) and here pb22l9-5F-p2 (comment 264)

timmyc:
Fixes #3460 

Sometimes things go sideways during the theme install step of the profiler. Typically this happens when a theme is already installed, and clicking on "Choose" doesn't do anything. I've only been able to reproduce this bug once, so @pmcpinto and I discussed in the issue that a good work-around for this would be to expose a "Skip this step" link if a theme is already chosen, and it supports WooCommerce. So that is the solution presented in this branch.
### Screenshots

skip-step

- On your test site, have a theme selected that **does not** support Woo
- Visit the profiler theme step, and verify the "Skip this step" link pictured above is not shown.
- Also verify toggling between the All | Paid | Free theme tabs up top still function as expected since some of that logic was tweaked a bit in this branch.
- Activate a Woo-supported theme.
- Revisit the theme step and verify the skip link is shown.
- Click the "Skip this step" link and confirm you are then directed to the dashboard.
- Bonus points to watch for the track being recorded on your last action

### Changelog Note:

Tweak: Onboarding - Add Skip Step link on Themes Step.

### DIFF
--- a/client/dashboard/profile-wizard/steps/theme/index.js
+++ b/client/dashboard/profile-wizard/steps/theme/index.js
@@ -137,7 +137,8 @@ class Theme extends Component {
 	}
 
 	skipStep() {
-		recordEvent( 'storeprofiler_store_theme_skip_step' );
+		const { activeTheme = '' } = getSetting( 'onboarding', {} );
+		recordEvent( 'storeprofiler_store_theme_skip_step', { activeTheme } );
 		this.props.goToNextStep();
 	}
 
@@ -207,7 +208,7 @@ class Theme extends Component {
 		return sprintf( __( '%s per year', 'woocommerce-admin' ), decodeEntities( price ) );
 	}
 
-	doesCurrentThemeSupportWooCommerce() {
+	doesActiveThemeSupportWooCommerce() {
 		const { activeTheme = '' } = getSetting( 'onboarding', {} );
 		const allThemes = this.getThemes();
 		const currentTheme = allThemes.find( theme => theme.slug === activeTheme );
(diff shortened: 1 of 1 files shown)


pmcpinto:
cc @jameskoster 

Some users were having problems in the Theme step and we felt that this was the best workaround to prevent them from being stuck in this step.

client/dashboard/profile-wizard/steps/theme/index.js
```diff
@@ -89,6 +90,11 @@ class Theme extends Component {
 		this.setState( { demo: theme } );
 	}
 
+	skipStep() {
+		recordEvent( 'storeprofiler_store_theme_skip_step' );
```
joshuatf: Might be worthwhile to track the active theme when skipped so we can see if there's a pattern associated with skipped themes.

client/dashboard/profile-wizard/steps/theme/index.js
```diff
@@ -155,6 +161,13 @@ class Theme extends Component {
 		return sprintf( __( '%s per year', 'woocommerce-admin' ), decodeEntities( price ) );
 	}
 
+	doesCurrentThemeSupportWooCommerce() {
```
joshuatf: Minor and nitpicky, but semantically I like `activeTheme` over `currentTheme` since this has less ambiguous meaning (e.g., current theme inside a live preview demo vs currently active theme on the site).

timmyc:
Thanks @joshuatf for the great suggestions - updates applied in 9b7e62c 👍 

image

PR #3503, opened 2020-01-02
Onboarding: Add option to track appearance completion

### LINKED ISSUE #3458: Onboarding - customize appearance step: mark task as completed even if the user skips all the steps
The Customize Appearance step should be marked as completed even if the user skips the Create Homepage step, which is the only one required at this moment.

joshuatf:
Fixes #3458 

Allow the "Customize Appearance" step to be completed without creating a homepage.

### Screenshots
Screen Shot 2020-01-02 at 4 18 03 PM

1.  Delete any homepages created from the task list appearance task.
2. Walk through the steps, skipping each, and completing at the end.
3. Make sure the appearance task is complete and persists on refresh.
4. Delete the `woocommerce_task_list_appearance_complete` option if you need to retest this.

### DIFF
--- a/client/dashboard/task-list/tasks.js
+++ b/client/dashboard/task-list/tasks.js
@@ -27,15 +27,15 @@ import Payments from './tasks/payments';
 
 export function getAllTasks( { profileItems, options, query, toggleCartModal } ) {
 	const {
-		hasHomepage,
 		hasPhysicalProducts,
 		hasProducts,
+		isAppearanceComplete,
 		isTaxComplete,
 		shippingZonesCount,
 	} = getSetting( 'onboarding', {
-		hasHomepage: false,
 		hasPhysicalProducts: false,
 		hasProducts: false,
+		isAppearanceComplete: false,
 		isTaxComplete: false,
 		shippingZonesCount: 0,
 	} );
@@ -93,7 +93,7 @@ export function getAllTasks( { profileItems, options, query, toggleCartModal } )
 			content: __( 'Create a custom homepage and upload your logo', 'wooocommerce-admin' ),
 			icon: 'palette',
 			container: <Appearance />,
-			completed: hasHomepage,
+			completed: isAppearanceComplete,
 			visible: true,
 		},
 		{
--- a/client/dashboard/task-list/tasks/appearance.js
+++ b/client/dashboard/task-list/tasks/appearance.js
@@ -35,6 +35,7 @@ class Appearance extends Component {
 		};
 
 		this.state = {
+			isDirty: false,
 			isPending: false,
 			logo: null,
 			stepIndex: 0,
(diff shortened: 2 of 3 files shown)


rrennick:
@joshuatf This works as described. I noticed I could click both `Proceed` and `Skip this step` on the image upload whether or not an image is selected. Is that the intended behaviour?

joshuatf:
@rrennick Good question.  The design includes both.  "Skip" will not acknowledge any changes done, while "Proceed" will add/remove/update to the selected option.

I could see us optionally removing the skip option, but I can also see an argument for its usefulness. /cc @pmcpinto @jameskoster for your opinions.

pmcpinto:
I think it might confuse users if they don't see the "skip" button. The "proceed" button is only active if the user uploaded some image, correct?

rrennick:
> The "proceed" button is only active if the user uploaded some image, correct?

`Proceed` was always active for me.

pmcpinto:
I think it would be better to have the button as inactive until the user uploads an image

joshuatf:
>I think it would be better to have the button as inactive until the user uploads an image

@pmcpinto This would mean that logos cannot be removed using this task anymore.  (If a user removes an image and clicks "Proceed" the logo will be removed in settings).

I'm okay with that if that's the direction we want to go, just wanted to note the consequences to disabling that button.

pmcpinto:
> @pmcpinto This would mean that logos cannot be removed using this task anymore. (If a user removes an image and clicks "Proceed" the logo will be removed in settings).

Ah ok, just to double-check: if the user clicks "remove image" but it doesn't click "proceed", the logo it's still in settings? In order to completely remove it, he needs to click "remove image" and then "proceed"?

rrennick:
> This would mean that logos cannot be removed using this task anymore.

My concern was that Proceed could be clicked as soon as the screen opens. If it's initially disabled and then enabled once they complete an image related activity. If the Skip is the only active option on load then it's clear to the user that they have skipped adding an image.

joshuatf:
>In order to completely remove it, he needs to click "remove image" and then "proceed"?

@pmcpinto That's correct.

>My concern was that Proceed could be clicked as soon as the screen opens. If it's initially disabled and then enabled once they complete an image related activity.

@rrennick I see what you mean.  One option would be to disable the "Proceed" button if no image is selected or one has not been removed (`isDirty` state), assuming we want to continue allowing users to remove images from here.

rrennick:
> One option would be to disable the "Proceed" button if no image is selected or one has not been removed (isDirty state), assuming we want to continue allowing users to remove images from here.

Sounds perfect.

joshuatf:
@rrennick Disabled this button following those rules in https://github.com/woocommerce/woocommerce-admin/pull/3503/commits/d059485907d309f534ddcc6d0140ff52a4f72ae5

So if a logo is set or if the state is dirty (removed a logo after already adding one), then "Proceed" is enabled.

pmcpinto:
> One option would be to disable the "Proceed" button if no image is selected or one has not been removed (isDirty state), assuming we want to continue allowing users to remove images from here.

Sound like the best option. Thanks!

PR #3504, opened 2020-01-02
Sync refactor tweaks

### LINKED ISSUE #3496: Importer progress bars not updating when in background
**Describe the bug**
In testing #3285, I found that if the import completes while the browser is not the foreground window, the progress bars are not updated to show that the import is complete. This occurs on both that PR and master.

**To Reproduce**

1. Deactivate WC Admin
2. Create 500 orders with smooth generator
3. Activate WC Admin
4. Start the import & stay on the settings screen until the progress bars show the import in progress
5. Put the Settings screen in the background & wait for the import to complete
6. Bring Settings to foreground & see that the message is updated but the progress bar(s) are not.

**Screenshots**
Screen Shot 2019-12-31 at 6 57 50 PM

* OS: MacOS
* Browser: Chrome
* Version Latest

rrennick:
Fixes #3496 

This is a followup PR to #3285. With AS 2.2.X, the probability of the import status api call occurring while actions are in progress is about 5% (3/60). With AS 3.0, that probability is much closer to 100% when the last import actions are processing. This PR adds a check for `in-progress` actions if there are no remaining `pending` actions.

- Activate Action Scheduler 3.0
- Delete existing imported analytics data
- Import analytics data
- Remain on screen to verify that 
- The `Finalizing ...` message appears
- The import ends with the message `Historical data from onward imported`

### DIFF
--- a/src/Schedulers/ImportScheduler.php
+++ b/src/Schedulers/ImportScheduler.php
@@ -38,8 +38,20 @@ abstract class ImportScheduler {
 				'group'    => self::$group,
 			)
 		);
+		if ( empty( $pending_jobs ) ) {
+			$in_progress = self::queue()->search(
+				array(
+					'status'   => 'in-progress',
+					'per_page' => 1,
+					'search'   => 'import',
+					'group'    => self::$group,
+				)
+			);
+		} else {
+			$in_progress = true;
+		}
 
-		return ! empty( $pending_jobs );
+		return ! empty( $in_progress );
 	}
 
 	/**
--- a/src/Schedulers/ImportScheduler.php
+++ b/src/Schedulers/ImportScheduler.php
@@ -47,8 +47,6 @@ abstract class ImportScheduler {
 					'group'    => self::$group,
 				)
 			);
-		} else {
-			$in_progress = true;
 		}
 
 		return ! empty( $pending_jobs ) || ! empty( $in_progress );
--- a/src/Schedulers/ImportScheduler.php
+++ b/src/Schedulers/ImportScheduler.php
@@ -51,7 +51,7 @@ abstract class ImportScheduler {
 			$in_progress = true;
 		}
 
-		return ! empty( $in_progress );
+		return ! empty( $pending_jobs ) || ! empty( $in_progress );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


jeffstieler:
@rrennick looks like this is bringing a lot of commits that were squashed in the merge to master. Can you rebase or cherry pick just your new commits?

rrennick:
> Can you rebase or cherry pick just your new commits?

Cherry picked.

src/Schedulers/ImportScheduler.php
```diff
@@ -38,8 +38,20 @@ public static function is_importing() {
...
+					'search'   => 'import',
+					'group'    => self::$group,
+				)
+			);
+		} else {
+			$in_progress = true;
+		}
 
-		return ! empty( $pending_jobs );
+		return ! empty( $in_progress );
```
joshuatf: The logic with the if/else above was a little hard to follow.  Maybe we could make this:

```php
return ! empty( $pending_jobs ) || ! empty( $in_progress_jobs );
```

src/Schedulers/SchedulerTraits.php
```diff
@@ -234,7 +234,11 @@ public static function get_next_blocking_job( $action_name ) {
 		if ( is_array( $blocking_jobs ) ) {
 			foreach ( $blocking_jobs as $blocking_job ) {
 				$blocking_job_hook = $blocking_job->get_hook();
-				$next_job_schedule = $blocking_job->get_schedule()->next();
+				if ( method_exists( $blocking_job->get_schedule(), 'get_date' ) ) {
```
joshuatf: Curious why we use `get_date()` here- is `next()` not available to in progress actions?  And is `get_date()` not available to pending actions?

rrennick: `next()` logs a deprecated warning in 3.0. `get_date()` was added in 3.0.

joshuatf: That makes sense - thanks for the explanation.

src/Schedulers/ImportScheduler.php
joshuatf: Not sure if you've finished updating or not, but I was thinking we remove the latter else portion:

```
else {
	$in_progress = true;
}
```

In favor of checking `$pending_jobs` and `$in_progress_jobs`.

rrennick: @joshuatf Thanks for the followup. Removed in `0d42422`.

PR #3507, opened 2020-01-03
Add SelectControl debouncing and keyboard fixes

### LINKED ISSUE #3424: SelectControl: Bugs/regressions
**Describe the bug**

I noticed a bugs and a regression while using the SelectControl component.

* Debounce no longer working and request sent on focus.
* Tabbing to the next element does not always close the dropdown.
* `esc` causes an error.

**To Reproduce**

Go to a report with the comparison box: `/wp-admin/admin.php?page=wc-admin&filter=compare-products&path=%2Fanalytics%2Fproducts`

1. Open the network tab and see a request fired on focus. The results don't populate the dropdown. Maybe that is the intention? A request is fired immediately after each keystroke instead of being debounced. Also, the previous implementation kept the results in component state, which was convenient for coming back with the same query.

2. After entering a query string, tab to the next element on the page. Notice this functioning properly. Try again, but this time with one result already selected. The dropdown should disappear.

Screen Shot 2019-12-16 at 1 04 06 PM

3. Place focus on the input box and enter a search string. Once results populate the dropdown, press `esc` and see an error in the console.

Screen Shot 2019-12-16 at 1 06 05 PM

joshuatf:
Fixes #3424 

* Adds a debouncing prop to `SelectControl` for typed queries.
* Fixes the escape key.
* Tab key now closes the list

### Screenshots
Screen Shot 2020-01-03 at 12 48 05 PM

1. Open the Product comparison or any other `Search` component.
2. Check that typed queries are debounced (500ms).
3. Hit the escape key and make sure no console error is logged.
4. Search for an item and select it.
5. Search again, but hit tab while still focused in the search control.
6. Make sure the options list is closed and the first selected item's close button is focused.
7. Make sure there are no regressions in other implementations of `SelectControl` (onboarding and devdocs).

### DIFF
--- a/packages/components/src/search/index.js
+++ b/packages/components/src/search/index.js
@@ -153,6 +153,7 @@ class Search extends Component {
 					onFilter={ this.appendFreeTextSearch }
 					onSearch={ this.fetchOptions }
 					options={ options }
+					searchDebounceTime={ 500 }
 					searchInputType={ inputType }
 					selected={ selected }
 					showClearButton={ showClearButton }
--- a/packages/components/src/search/index.js
+++ b/packages/components/src/search/index.js
@@ -87,6 +87,10 @@ class Search extends Component {
 	}
 
 	fetchOptions( previousOptions, query ) {
+		if ( ! query ) {
+			return [];
+		}
+
 		const autocompleter = this.getAutocompleter();
 		return autocompleter.options( query ).then( async response => {
 			const options = this.getFormattedOptions( response, query );
--- a/packages/components/src/select-control/control.js
+++ b/packages/components/src/select-control/control.js
@@ -149,7 +149,7 @@ class Control extends Component {
 		}
 
 		// Show the search query when focused on searchable controls.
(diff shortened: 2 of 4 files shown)


rrennick:
@joshuatf Nice :) This works as described. 

If I use the down key to highlight an item then hit tab, it closes the list. Most suggestion lists I have used treat a tab as a select this. You can see this in the Chrome developer console under Elements -> element.style. Type `di` then tab and it autofills `display`. I just wanted to check whether this should do the same.

joshuatf:
>Most suggestion lists I have used treat a tab as a select this. 

@rrennick I believe your thinking here is correct.  The examples [here](https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html) have the same implementation of selecting the current item on tab in addition to tabbing to the next item.

Added in https://github.com/woocommerce/woocommerce-admin/pull/3507/commits/3add89b1fdb96ca247cce710ba26972f38e56877

PR #3514, opened 2020-01-03
Analytics: Change Refunds to Returns

timmyc:
While working through some docs and chatting with @LevinMedia - he pointed out that he was hoping to rename **Refunds** in the context of the Revenue report to be **Returns**. This PR does that. I opted to not rename the element in the REST API response here as I'm concerned this will cause some issues with the mobile apps ( /cc @astralbodies ) - and since we don't have versions on these endpoints, I'm just going with the labeling change for now.### Screenshots

- Open the dashboard, verify Refunds now says returns in the Performance Indicator
- Also on the dashboard, verify the Refunds chart now says returns

### Changelog Note:

Tweak: Change Refunds to Returns on the dashboard.

### DIFF
--- a/client/dashboard/dashboard-charts/config.js
+++ b/client/dashboard/dashboard-charts/config.js
@@ -54,7 +54,7 @@ const defaultCharts = [
 		key: 'items_sold',
 	},
 	{
-		label: __( 'Refunds', 'woocommerce-admin' ),
+		label: __( 'Returns', 'woocommerce-admin' ),
 		report: 'revenue',
 		key: 'refunds',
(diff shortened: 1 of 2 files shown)


rrennick:
I'm not sure we should be changing this. A merchant can sell digital goods, virtual products, and physical goods. `Refunds` can be used with all three but `Returns` only applies to physical products.

timmyc:
FWIW it has already been changed on the Revenue report.

astralbodies:
Thanks for the cc, @timmyc!

PR #3524, opened 2020-01-07
Onboarding: Add reminder to complete profiler

### LINKED ISSUE #3242: Onboarding: Add notice to complete onboarding profiler / task list
In https://github.com/woocommerce/woocommerce-admin/pull/3232 we removed the old notice that prompts the user to finish the setup wizard.

@justinshreve brought up a great point that we might want to do the same for the new onboarding profiler and/or task list.

This is something we can re-evaluate in phase 2 based on how many completed the profiler or task list.

Screen Shot 2019-11-13 at 5 19 54 PM

joshuatf:
Fixes #3242

Adds a notice to complete or skip the profiler.  @pmcpinto Just to double check, we wanted to give a secondary action to skip the profiler, right? (As opposed to just dismissing the notice).

This also adds a way to mark the note as `actioned` both in client-side and server-side since both are capable of marking the profiler complete.

~Tracking for note actions will be handled in a follow-up.~ Tracking is handled by default for all note actions.

### Screenshots
Screen Shot 2020-01-07 at 8 35 17 PM

1. Enable the profiler.
2. Visit any non-dashboard WC pages and note the reminder to complete the profiler in the store alerts section.
3. Walk through the profiler, selecting a theme, and completing the profiler.
4. Note that the note is removed.
5. Delete the note (name - `wc-admin-onboarding-profiler-reminder`) in the `wc_admin_notes` table.
6. Re-enable the profiler and make sure the note is added again.
7. Click "Continue Store Setup" and make sure this brings you to the onboarding profiler.
8. Go back to the note and click "Skip Setup" and make sure this marks the notice `actioned` (hides it) and takes you to the task list.
9. Delete the note.
10. Enable the profiler via the Help tab and disable it again.
11. Make sure the note has been removed.

### DIFF
--- a/client/dashboard/profile-wizard/index.js
+++ b/client/dashboard/profile-wizard/index.js
@@ -22,6 +22,7 @@ import Industry from './steps/industry';
 import Plugins from './steps/plugins';
 import ProductTypes from './steps/product-types';
 import ProfileWizardHeader from './header';
+import { QUERY_DEFAULTS } from 'wc-api/constants';
 import Start from './steps/start';
 import StoreDetails from './steps/store-details';
 import Theme from './steps/theme';
@@ -144,15 +145,15 @@ class ProfileWizard extends Component {
 		const nextStep = this.getSteps()[ currentStepIndex + 1 ];
 
 		if ( 'undefined' === typeof nextStep ) {
-			this.finishWizard();
+			this.possiblyShowCart();
 			return;
 		}
 
 		return updateQueryString( { step: nextStep.key } );
 	}
 
-	finishWizard() {
-		const { profileItems, updateProfileItems } = this.props;
+	possiblyShowCart() {
+		const { profileItems } = this.props;
 
 		// @todo This should also send profile information to woocommerce.com.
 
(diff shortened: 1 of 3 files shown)


pmcpinto:
> Adds a notice to complete or skip the profiler. @pmcpinto Just to double check, we wanted to give a secondary action to skip the profiler, right? (As opposed to just dismissing the notice).

Yep that's correct. Do you want a separate issue for the event tracking on this notification or it's already covered on this one?

joshuatf:
>Do you want a separate issue for the event tracking on this notification or it's already covered on this one?

@pmcpinto No need to 😄 - opened https://github.com/woocommerce/woocommerce-admin/pull/3533 and shortly after realized all note actions are currently tracked with note details.  Event name will be `wcadmin_store_alert_action` with the note details and action details.

rrennick:
@joshuatf Nice work. This works great. When I disabled the profiler it took me to the store setup (same action as when I enabled it. Is that the intended behaviour?

> 10. Enable the profiler via the Help tab and disable it again.

joshuatf:
>When I disabled the profiler it took me to the store setup (same action as when I enabled it. Is that the intended behaviour?

@rrennick It should take you to the task list (dashboard) and disable the profiler.  I should have worded this a bit better, but this step was mostly a sanity check to make sure disabling this step by other means than the note action still worked as expected.

PR #3541, opened 2020-01-08
Try using version of phpunit that supports random order

rrennick:
This is a test PR to see whether we can update PHPUnit to a version that supports random ordering of tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,15 +12,15 @@ matrix:
   - name: "PHP 7.2 unit tests"
     php: 7.2
     env: WP_VERSION=latest WP_MULTISITE=0 WP_CORE_DIR=/tmp/wordpress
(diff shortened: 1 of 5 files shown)


jeffstieler:
Don't we need to update the lock file for this?

jeffstieler:
@rrennick it looks like `bin/phpunit.sh` might need to be updated to run the `vendor` version of `phpunit` - I'm still seeing version 6.x in CI

rrennick:
I'm going to close this as it's not practical to implement in Travis until the lowest PHP test version is 7.1.

Changing the phpunit version directly in `composer.json` works with a `composer update`. 

`composer require phpunit/phpunit:~7.5` fails because phpunit 6.X is configured with a dependency on `phpunit-mock-objects` which is incompatible with phpunit 7.X. Changing the phpunit version between tests isn't practical for this use case.

PR #3543, opened 2020-01-09
Allow 5.3-alpha builds for unit tests.

jeffstieler:
`version_compare()` considers the alpha "less than" the release.

This fixes an issue I had on my VVV environment running `5.3-alpha-45282-src`.

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -91,6 +91,14 @@ function _manually_load_plugin() {
 	update_option( 'woocommerce_calc_taxes', 'yes' );
 	define( 'WOOCOMMERCE_ADMIN_ONBOARDING_ENABLED', true );
 
+	// Fix alpha builds of WP (like on VVV).
+	global $wp_version;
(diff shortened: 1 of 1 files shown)


tests/bootstrap.php
```diff
@@ -91,6 +91,14 @@ function _manually_load_plugin() {
 	update_option( 'woocommerce_calc_taxes', 'yes' );
 	define( 'WOOCOMMERCE_ADMIN_ONBOARDING_ENABLED', true );
 
+	// Fix alpha builds of WP (like on VVV).
+	global $wp_version;
+	$alpha_tag = strpos( $wp_version, '-alpha' );
+	if ( false !== $alpha_tag ) {
+		// phpcs:ignore
```
rrennick: Can we change this and line #44 to `// phpcs:ignore WordPress.WP.GlobalVariablesOverride` so that it passes the sniffs without warning?

jeffstieler:
This really shouldn't be necessary. I've updated my dev environment.

Issue #3392, opened 2019-12-06
Testing in WooCommerce Core


timmyc:
With @psealock's work that brought us to:

https://github.com/woocommerce/woocommerce/pull/25011

It is time to put some various flows through the paces. Here are some scenarios I would like us to test out... please do add more if you think of them. But testing the branch above in the following setups.

- [x] Clean WooCommerce install ( no prior wc-admin present ). Verify wc-admin functionality.
- [x] wc-admin installed, but not activated
- [x] Same as above, but then activate wc-admin
- [x] wc-admin installed, and activated. Like the blocks repository, if this setup is encountered, verify the plugin takes precedence.
- [x] Same as above, but then deactivate and uninstall the wc-admin plugin. Database tables for wc-admin *should not* be dropped, and wc-admin functionality that is now active via woo core should operate as expected.

Mainly I'm very curious to test out db migrations as well. I have a feeling the last test flow above could result in tables being dropped, re-added, but empty which would lead to some 😢

jeffstieler:
> I have a feeling the last test flow above could result in tables being dropped, re-added, but empty which would lead to some 😢

I can confirm - deleting the plugin results in all tables being truncated 😱

timmyc:
> I can confirm - deleting the plugin results in all tables being truncated

Seems like we might be able to add some logic to uninstall to prevent this from happening based off Woo core version

jeffstieler:
> Seems like we might be able to add some logic to uninstall to prevent this from happening based off Woo core version

Was working on this before I saw your comment - chose to look for the plugin files instead of checking version: https://github.com/woocommerce/woocommerce-admin/pull/3546

jeffstieler:
Question - what should happen if someone deletes core WooCommerce, but the WooCommerce Admin feature plugin is active?

Right now, that would cause all WC-Admin tables to be dropped, and then the active plugin would be dormant because its dependencies aren't met.

jeffstieler:
Looks like the `wc_admin_process_orders_milestone` cron hook is getting doubled up when the feature plugin is enabled.

jeffstieler:
I'm not seeing the `wc_admin_daily` event get created (or at least hooked into as far as WP Crontrol is concerned) when WC Admin loads from the core package.

jeffstieler:
#3580 preserves the scheduled actions and cron events when the feature plugin is disabled.

psealock:
[...]

PR #3546, opened 2020-01-10
Revamp uninstallation when WooCommerce Admin is in core.

jeffstieler:
Found when testing https://github.com/woocommerce/woocommerce-admin/issues/3392.

This PR seeks to add detection of WooCommerce Admin in core WooCommerce, and avoid `DROP`ing table data when the feature plugin in uninstalled.

It also renames all of our options and user meta keys to be prefixed by `woocommerce_` rather than `wc_`. (It automatically migrates the keys as well) Having the unabbreviated prefix allows core WC to clean up options and user meta for us when WC itself is uninstalled.

**NOTE: these instructions will destroy all of your WC data. Make a backup first!**

- Add `define( 'WC_REMOVE_ALL_DATA', true );` to your config
- Run WooCommerce Core version https://github.com/woocommerce/woocommerce/pull/25011
- Activate WooCommerce Admin from this branch (it can be helpful to remove the `.git` and `node_modules` directories prior to deletion)
- Load the dashboard
- Go to Analytics > Settings, start an import
- Go to any report (with more than 1 page of results) and download it
- Verify that `wc_admin_` options have been migrated to `woocommerce_admin_`
- Verify that `wc_admin_` user meta have been migrated to `wc_admin_`
- Disable WooCommerce Admin plugin
- Delete WooCommerce Admin plugin
- Verify that all WooCommerce Admin table data is intact (like `wp_wc_order_stats`)
- Disable WooCommerce plugin
- Delete WooCommerce plugin
- Verify that all WooCommerce and WooCommerce Admin tables are gone
- Verify that all WooCommerce and WooCommerce Admin options and user meta is gone

### Changelog Note:

Fix: handle uninstallation when in feature plugin mode.

### DIFF
--- a/src/Install.php
+++ b/src/Install.php
@@ -64,7 +64,10 @@ class Install {
 		// Add wc-admin report tables to list of WooCommerce tables.
 		add_filter( 'woocommerce_install_get_tables', array( __CLASS__, 'add_tables' ) );
 
-		// Handle option name migrations.
+		// Migrate option names by filtering their default values.
+		// This attaches a targeted filter for each migrated option name that will retreive
+		// the old value and use it as the default for the new option. This default
+		// will be used in the first retreival of the new option.
 		foreach ( self::$migrated_options as $new_option => $old_option ) {
 			add_filter( "default_option_{$new_option}", array( __CLASS__, 'handle_option_migration' ), 10, 2 );
 		}
--- a/src/Install.php
+++ b/src/Install.php
@@ -20,7 +20,7 @@ class Install {
 	/**
 	 * Plugin version option name.
 	 */
-	const VERSION_OPTION = 'wc_admin_version';
+	const VERSION_OPTION = 'woocommerce_admin_version';
 
 	/**
 	 * DB updates and callbacks that need to be run per version.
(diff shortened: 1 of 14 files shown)


uninstall.php
```diff
@@ -7,6 +7,15 @@
 
 defined( 'WP_UNINSTALL_PLUGIN' ) || exit;
 
+// Don't delete table data if core WooCommerce contains WooCommerce Admin.
+// This could alternatively check for a specific WooCommerce version rather than a file.
+if (
+    defined( 'WC_ABSPATH' ) &&
```
joshuatf: Nice tactic here!

I think it may be an edge case (if at all), but does `uninstall.php` run after the `plugins_loaded`?  Is there a case where `WC_ABSPATH ` is not yet defined since wc-admin loaded first?

jeffstieler: `uninstall.php` is invoked in `uninstall_plugin()` through `delete_plugins()` and always occurs after `plugins_loaded`.

See:
https://github.com/WordPress/WordPress/blob/a5e52e04ee5f4943b3883e7857bdc28a7594bf68/wp-admin/plugins.php#L380

https://github.com/WordPress/WordPress/blob/f793398e4f26051a49abd03a084fb43d3c578c3b/wp-admin/includes/ajax-actions.php#L4541

jeffstieler:
> This seems like it will prevent the uninstall from running even when deactivating WC as well. Do we have a plan for handling uninstall in core (when the feature plugin is not present)?

Ooooh.. good catch. I'll work on updating this to handle core WC uninstallation.

timmyc:
@jeffstieler is this ready for another set of 👀 ?

jeffstieler:
@timmyc - yes, let me get the testing instructions updated then give it a spin!

timmyc:
FWIW I'm seeing the same issue I noted in my review above when the wc-admin feature plugin is not installed... [noted here](https://github.com/woocommerce/woocommerce/pull/25011#issuecomment-573258037)

jeffstieler:
@timmyc the table deletion should be fixed in the core PR. In this PR I've added a bug fix for double-initialization of the `FeaturePlugin` class when the plugin and package versions were both present.

src/Install.php
```diff
@@ -47,6 +63,36 @@ public static function init() {
 
 		// Add wc-admin report tables to list of WooCommerce tables.
 		add_filter( 'woocommerce_install_get_tables', array( __CLASS__, 'add_tables' ) );
+
+		// Handle option name migrations.
+		foreach ( self::$migrated_options as $new_option => $old_option ) {
+			add_filter( "default_option_{$new_option}", array( __CLASS__, 'handle_option_migration' ), 10, 2 );
```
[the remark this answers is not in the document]
jeffstieler: Added in cdaeb02.

PR #3552, opened 2020-01-13
add PHP 7.3 to Travis unit tests

rrennick:
This PR adds PHP 7.3 to the unit test suite in Travis CI.


.travis.yml
```diff
@@ -6,9 +6,12 @@ dist: trusty
...
 matrix:
   include:
-  - name: "PHP 7.2 unit tests, PHP Coding standards check and JS tests"
-    php: 7.2
+  - name: "PHP 7.3 unit tests, PHP Coding standards check and JS tests"
+    php: 7.3
     env: WP_VERSION=latest WP_MULTISITE=0 WP_CORE_DIR=/tmp/wordpress RUN_PHPCS=1 RUN_JS=1
+  - name: "PHP 7.2 unit tests"
+    php: 7.2
+    env: WP_VERSION=latest WP_MULTISITE=0 WP_CORE_DIR=/tmp/wordpress
```
joshuatf: Would passing `RUN_PHPCS=1` as an argument here work or do we need the `$TRAVIS_PHP_VERSION == "7.2"` below?

rrennick: > Would passing RUN_PHPCS=1 as an argument here work

Yes, that would work but run the code sniffer twice. The WC sniffs are a couple versions behind core so I'm going to look at a separate PR to see if we can get our config more in line with WC core.

rrennick: @joshuatf Thanks for the feedback. I'm putting this on hold pending #3571 being merged.

joshuatf: ✅ on https://github.com/woocommerce/woocommerce-admin/pull/3571 - if you want to rebase this one and ping me, I'll take another look.

rrennick: @joshuatf Thanks for the suggestion. I added a `COMPOSE_DEV` variable to signal using package PHPUnit.

bin/travis.sh
```diff
@@ -4,7 +4,8 @@
 if [ "$1" == 'before' ]; then
 	cd "$WP_CORE_DIR/wp-content/plugins/woocommerce-admin/"
 	npm run build:feature-config
-	if [[ "$RUN_PHPCS" == "1" || "$RUN_RANDOM" == "1" ]]; then
+	# Load PHPUnit version in dev dependencies for PHP ^7.2 until WP unit tests support PHPUnit 8.X
+	if [[ "$COMPOSE_DEV" == "1" ]]; then
```
jeffstieler: Aside from a potential reduction in execution time, is there a reason to skip the dev dependencies in CI?

Nitpick: why "compose" dev and not "composer" dev?

rrennick: > is there a reason to skip the dev dependencies in CI?

PHPUnit 7 requires PHP 7.1 or higher

```
This version of PHPUnit is supported on PHP 7.1, PHP 7.2, and PHP 7.3.
You are using PHP 7.0.25 (/home/travis/.phpenv/versions/7.0.25/bin/php).
```

Looking at WC, the way they have implemented it is https://github.com/woocommerce/woocommerce/blob/master/.travis.yml#L71 . Would it be better to do it through allowing both versions in composer.json & always install dev?

> why "compose" dev and not "composer" dev?

Updated.

jeffstieler:
@rrennick it looks like this is still failing in CI - do we need to update the `composer.lock` for the install to work ?

rrennick:
> do we need to update the composer.lock for the install to work ?

Yeah, `composer install` doesn't update the lock file. If my last commit doesn't work we can use the same method as WC core.

rrennick:
@jeffstieler The scripts on this one are now working correctly.

PR #3557, opened 2020-01-14
add action scheduler package via composer

rrennick:
This PR loads Action Scheduler 3.1.0 via composer.

It removes the custom AS post store which had the priority setting code.

- Install & activate Action Scheduler `master` - https://github.com/woocommerce/action-scheduler
- Edit the registered version to `3.1.1` https://github.com/woocommerce/action-scheduler/blob/master/action-scheduler.php#L39 (This works because plugins are loaded in alphabetical order.)
- Allow the migration to run
- Switch to this branch
- run `composer install`
- Add the following test snippet
```php
add_action( 'admin_menu', function() {
error_log(ActionScheduler_Versions::instance()->latest_version());	
});
```
- Load a dashboard page
- Deactivate Action Scheduler
- debug.log should have
```
[14-Jan-2020 20:07:49 UTC] 3.0.2
[14-Jan-2020 20:07:52 UTC] 3.0.2
[14-Jan-2020 20:07:52 UTC] 3.0.1
```
- Re-import your analytics data
- verify order & customer imports processed correctly

### Changelog Note:

Dev: Add Action Scheduler 3.1.0


src/Schedulers/SchedulerTraits.php
```diff
@@ -262,8 +262,14 @@ public static function do_action_or_reschedule() {
 		// or schedule to run now if no blocking jobs exist.
 		$blocking_job = static::get_next_blocking_job( $action_name );
 		if ( $blocking_job ) {
+			$schedule = $blocking_job->get_schedule();
+			if ( method_exists( $schedule, 'get_date' ) ) {
```
timmyc: What would happen here so schedule wouldn't have this method? Should we add a comment to clarify?

tests/api/reports-import.php
```diff
@@ -107,7 +107,7 @@ public function test_import_params() {
 		$order_2->save();
 
 		// Delete order stats so we can test import API.
-		$wpdb->query( "DELETE FROM {$wpdb->posts} WHERE post_type = 'scheduled-action'" );
+		ReportsSync::clear_queued_actions();
```
timmyc: Nice!

src/Schedulers/SchedulerTraits.php
rrennick: Nice catch. Added in `8646371`.

src/Schedulers/SchedulerTraits.php
```diff
@@ -234,6 +234,7 @@ public static function get_next_blocking_job( $action_name ) {
 		if ( is_array( $blocking_jobs ) ) {
 			foreach ( $blocking_jobs as $blocking_job ) {
 				$blocking_job_hook = $blocking_job->get_hook();
+				// next() was deprecated in AS 3.0 and superseded by get_date().
```
timmyc: Thanks!

timmyc:
@rrennick just saw the label change, is this good to test further or were you going to do some more work?

rrennick:
> is this good to test further or were you going to do some more work?

I need to address the issue with the unit tests not passing.

rrennick:
Adding a note here on the reasons for the changes to the unit tests:

- In the WP Post store, the pending action list is retrieved with a default ordering of action ID ASC. In normal queue processing, the custom store does order actions by action id on the innermost sort. Our unit tests bypass the normal queue processing and just processes a list of retrieved actions. Tests were intermittently failing because 2 orders created for the same customer might not be processed in the same order on every execution.
- The PR changes some assumptions to explicit requirements in the batch queue tests. 

@timmyc This is ready for another review. Can we add a February release to Zenhub to move this PR to?

bin/composer-package-update.sh
```diff
@@ -0,0 +1,32 @@
...
+	output 1 "./vendor doesn't exist!"
+	output 1 "run \"composer install\" before proceeding."
+fi
+
+
+# Convert textdomains
+output 3 "Updating package textdomains..."
+
+# Replace text domains within packages with woocommerce
+find ./vendor/action-scheduler -iname '*.php' -exec sed -i.bak -e "s/'action-scheduler' /'woocommerce-admin' /g" -e "s/\"action-scheduler\" /'woocommerce-admin' /g" {} \;
```
claudiosanches: We are improving it on WooCommerce core, so we avoid replacing non-translatable strings: https://github.com/woocommerce/woocommerce/pull/25536

rrennick: @claudiosanches Thanks for pinging us with this. In case you find it useful, I added a `--no-warnings` to the node command in our implementation so it wouldn't output a line for every text domain it updated.

rrennick:
@timmyc This is updated to AS 3.1 and ready for another review.

jeffstieler:
Is a specific version of WC required for this?

I'm getting these errors after disabling the AS plugin:

```
PHP Fatal error:  Uncaught Error: Call to undefined function as_get_scheduled_actions() in /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce/includes/queue/class-wc-action-queue.php:158
Stack trace:
#0 /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-admin/src/Schedulers/ImportScheduler.php(43): WC_Action_Queue->search(Array)
#1 /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-admin/src/ReportsSync.php(65): Automattic\WooCommerce\Admin\Schedulers\ImportScheduler::is_importing()
#2 /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-admin/src/ReportsSync.php(133): Automattic\WooCommerce\Admin\ReportsSync::is_importing()
#3 /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-admin/src/API/Reports/Import/Controller.php(289): Automattic\WooCommerce\Admin\ReportsSync::get_import_stats()
#4 /srv/www/wordpress-default/public_html/wp-includes/rest-api/class-wp-rest-server.php(946): Automattic\WooCommerce\Admin\API\Reports\Import\Controlle in /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce/includes/queue/class-wc-action-queue.php on line 158
```

rrennick:
> I'm getting these errors after disabling the AS plugin:

Did you run `composer install`? If you ran that in another branch AS may have been removed.

jeffstieler:
Yes I ran `composer install` (just did again to be sure even) - I'm getting those PHP errors when loading the Analytics > Settings page.

rrennick:
@jeffstieler To proceed with this PR we would need to include `action-scheduler.php` before `vendor/autoload.php`. Do you have any concerns about doing that?

jeffstieler:
> To proceed with this PR we would need to include `action-scheduler.php` before `vendor/autoload.php`. Do you have any concerns about doing that?

Can you elaborate on why the autoloader isn't able to handle this use case?

rrennick:
> Can you elaborate on why the autoloader isn't able to handle this use case?

See https://github.com/woocommerce/action-scheduler/pull/468#issuecomment-587576735

timmyc:
@rrennick guessing we might be closing this one out?

rrennick:
Due to the issue with composer autoloading we didn't have time in the core release schedule to implement this.

PR #3560, opened 2020-01-15
wp.data Settings Refactor: update settings state

psealock:
This PR changes the shape of settings state. Specifically, it removes the `group` property because changes introduced in #3430 make in not necessary.

New state has a shape like this.

```js
const state = {
	error, // error messages
	lastReceived // timestamp
	ids,  // array of data property names
	dirty, // array of property ids that have been changed without persistence
	[ id ]: { data, error, lastReceived }, // a sample data entry
}
```

### Testing Instructions

1. Load the settings page and make sure it renders
2. Change and save a settings
3. Make sure its working


client/analytics/settings/config.js
```diff
@@ -22,8 +22,9 @@ const DEFAULT_ORDER_STATUSES = [
 	'pending',
 	'on-hold',
 ];
+const DEFAULT_DATE_RANGE = 'period=month&compare=previous_year';
```
psealock: This default was added by the client in `client/settings`, but I think it belongs here because its a default that does not come from the server and is only required here.

rrennick:
> My only comment/question is whether or not we may need this code for other setting groups in the future. It seems like an opportunity to make this code reusable for other settings pages in the future.

Also, can we provide some utility to other extensions?

psealock:
Great point here @joshuatf about reusability by extension. I am wondering if I've jumped the gun on this one. The core of the issue is that groups allows for addition/removal of settings properties in its own namespace. However, as it is now, there's no reason an extension can't use this data store. 

```js
// in PHP or wherever
window.extSettings = { my_setting: 123 };

// index.js
const appRoot = document.getElementById( 'root' );
render( withSettingsHydration( window.extSettings )( ExtensionApp ), appRoot );

// app.js
const ExtensionApp = () => {
	const { my_setting } = useSettings( [ 'my_setting' ] );
	
	return (
		<div>{ my_setting }</div>
	);
}
```
The only caveat to that is unique groups avoid collision of duplicate properties and allow for wholesale removal. Are these points (or any others I'm not seeing) enough of a downside to keep things the way they are?

nerrad:
It looks like the reason why there was a group (that I could see) is because the settings endpoint appeared to use that in it's route.  Are there any other routes that might be other groups other than `wc_admin`? 

According to the [controller](https://github.com/woocommerce/woocommerce-rest-api/blob/master/src/Controllers/Version4/Settings.php) it looks like there could be?  If so, and you want extensions to be able to use this data store powered by the REST API (and the `Settings` route) for "groups" other than wc_admin then I think you'll have to abandon this pull. Otherwise, you tightly couple this data store to _only_ the `wc_admin` group using the `Settings` api.

I see two choices here:

- keep this refactor but move the data store into `client/` (because it's utility is specific to the wc-admin app).
- drop this refactor and keep in `packages` to be published as a package because then other woo products can consume the package to use the same interface for interacting with other registered settings groups (along with `wc_admin`).

psealock:
>It looks like the reason why there was a group (that I could see) is because the settings endpoint appeared to use that in it's route. 

Oh yes, of course. I overlooked this aspect.

>Are there any other routes that might be other groups other than wc_admin?

I don't think so. But I think abandoning this pull request is the best way forward. This allow for use of the API settings endpoint by extensions.

PR #3573, opened 2020-01-15
Re-add stockStatuses to wcSettings

psealock:
Stock Statuses were missing from `wcSettings`

This pr adds `stockStatuses` to the settings object.

### Before

Screen Shot 2020-01-16 at 10 27 19 AM

### After

Screen Shot 2020-01-16 at 10 27 37 AM

### Detailed test instructions:

1. Ensure stock statuses are displayed on the Stock Report

PR #3574, opened 2020-01-16
Reports: Use new data store

psealock:
Depends on https://github.com/woocommerce/woocommerce-admin/pull/3560

Migrate the Reports pages away from using `@woocommerce/wc-admin-settings` as a source for settings in favour of the new settings data store found in packages.

When reviewing this PR, I'd suggest adding the `?w=1` parameter to avoid looking at whitespace changes.

Make sure the Reports are functioning as they should.

You may notice stock statuses aren't rendering in the Stock Report. That fix is in https://github.com/woocommerce/woocommerce-admin/pull/3573

### DIFF
--- a/packages/data/src/settings/selectors.js
+++ b/packages/data/src/settings/selectors.js
@@ -30,7 +30,7 @@ export const getIsDirty = ( state, keys = [] ) => {
 export const getSettingsForKeys = ( state, keys ) => {
 	const allSettings = getSettings( state );
 	return keys.reduce( ( accumulator, key ) => {
-		accumulator[ key ] = allSettings[ key ] || null;
+		accumulator[ key ] = allSettings[ key ];
 		return accumulator;
 	}, {} );
 };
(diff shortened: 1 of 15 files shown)


packages/data/src/settings/selectors.js
```diff
@@ -30,7 +30,7 @@ export const getIsDirty = ( state, keys = [] ) => {
 export const getSettingsForKeys = ( state, keys ) => {
 	const allSettings = getSettings( state );
 	return keys.reduce( ( accumulator, key ) => {
-		accumulator[ key ] = allSettings[ key ] || null;
+		accumulator[ key ] = allSettings[ key ];
```
psealock: In order to allow destructuring with defaults, these values need to be `undefined`.

psealock:
closing in favour of https://github.com/woocommerce/woocommerce-admin/pull/3581 which has similar changes, but different base.

PR #3580, opened 2020-01-16
Avoid redundant/unnecessary cleanup.

jeffstieler:
Found while testing #3392.

The feature plugin doesn't need to clean up when the core package exists.

This PR adds a constant `WC_ADMIN_PACKAGE_EXISTS` when WooCommerce Admin exists within core WooCommerce.

When the constant is present, the plugin won't "clean up" on deactivation because that would disrupt the version loaded from core.

**NOTE:** This PR **does not** provide a mechanism to handle divergence between the feature plugin and the core package. Any changes to DB schema, event scheduling, etc, will need to be considered and potentially handled in the future. (Should plugin users decide to deactivate)

- Copy `src/Package.php` to `packages/woocommerce-admin/src/Package.php` in your `woocommerce` directory (running https://github.com/woocommerce/woocommerce/pull/25011)
- Activate `woocommerce-admin` from this PR is active
- Deactivate `woocommerce-admin`
- Verify that the `wc_admin_daily` cron event still exists (use WP Crontrol)

### DIFF
--- a/src/Install.php
+++ b/src/Install.php
@@ -379,6 +379,7 @@ class Install {
 		if ( ! wp_next_scheduled( 'wc_admin_daily' ) ) {
 			wp_schedule_event( time(), 'daily', 'wc_admin_daily' );
 		}
+		// @todo This is potentially redundant when the core package exists.
 		wp_schedule_single_event( time() + 10, 'generate_category_lookup_table' );
 	}
 
--- a/src/FeaturePlugin.php
+++ b/src/FeaturePlugin.php
@@ -96,6 +96,12 @@ class FeaturePlugin {
 	 * @return void
 	 */
 	public function on_deactivation() {
+		// Don't clean up if the WooCommerce Admin package is in core.
+		// NOTE: Any future divergence from the core package will need to be accounted for here.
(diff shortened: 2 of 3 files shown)


src/Install.php
```diff
@@ -379,6 +379,7 @@ public static function create_events() {
 		if ( ! wp_next_scheduled( 'wc_admin_daily' ) ) {
 			wp_schedule_event( time(), 'daily', 'wc_admin_daily' );
 		}
+		// @todo This is potentially redundant when the core package exists.
```
rrennick: Could this be conditional on a `wp_next_scheduled()` check?

jeffstieler: I don't think `wp_next_scheduled()` will work here because it's not a recurring event.

jeffstieler: It could stop queuing two events, but if the first has already run and completed - we'd schedule another one.

jeffstieler:
Going to merge - the redundant category lookup generation isn't a huge deal.

src/Install.php
rrennick: Better for a follow up then.

PR #3582, opened 2020-01-17
Ignore vendor directory for stylelint

timmyc:
Fixes an exception being thrown by `npm run lint` - which I found while testing #3562. The proposed fix here is to not run the stylelint against the `vendor` directory

- On `master`, run `npm run lint`, note the error thrown
- Check out this branch and run again, no error 🥇 

### Changelog Note:

Dev: Skip `vendor` directory when running stylelint

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,7 +1,7 @@
 {
 	"extends": "stylelint-config-wordpress",
 	"ignoreFiles": [
-        "./css/main.scss"
+        "./vendor/**/*.scss"
     ],
 	"rules": {
 		"at-rule-empty-line-before": null,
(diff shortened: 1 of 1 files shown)


.stylelintrc.json
```diff
@@ -1,5 +1,8 @@
 {
 	"extends": "stylelint-config-wordpress",
+	"ignoreFiles": [
+        "./css/main.scss"
```
rrennick: This didn't work for me but `"./vendor/**/*.scss"` did.

timmyc: weird that is what I had, but must have mixed up my commit, thanks Ron I'll fix that up

PR #3588, opened 2020-01-20
Onboarding: Center button to continue without installation

### LINKED ISSUE #3570: Onboarding: Continue without installation button isn't centered
**Describe the bug**
In the Business Details step, if the plugin install fails, the skip button is not centered.

**To Reproduce**

1. You can intentionally cause an install to fail by updating by replacing `$slug` with an empty string `''` in https://github.com/woocommerce/woocommerce-admin/blob/5d5f55e3b19fcceb928ee14fb05a1b7d7554a4a6/src/API/OnboardingPlugins.php#L195
1. Remove Mailchimp and/or Facebook plugins.
1. Visit the Business Details step.
1. Attempt to install these plugins on this step.
1. Note the button placement.

**Expected behavior**
Expected some spacing and centering for this button.

**Screenshots**
Screen Shot 2020-01-15 at 4 12 34 PM

joshuatf:
Fixes #3570 

Centers an uncentered button.

#### Before
Screen Shot 2020-01-15 at 4 12 34 PM

#### After
Screen Shot 2020-01-20 at 1 01 33 PM

1. You can intentionally cause an install to fail by updating by replacing `$slug` with an empty string `''` in https://github.com/woocommerce/woocommerce-admin/blob/5d5f55e3b19fcceb928ee14fb05a1b7d7554a4a6/src/API/OnboardingPlugins.php#L195
1. Remove Mailchimp and/or Facebook plugins.
1. Visit the Business Details step.
1. Attempt to install these plugins on this step.
1. Note the button placement.

### DIFF
--- a/client/dashboard/profile-wizard/steps/business-details.js
+++ b/client/dashboard/profile-wizard/steps/business-details.js
@@ -285,22 +285,24 @@ class BusinessDetails extends Component {
 				</div>
 
 				{ installExtensions && (
-					<Plugins
-						onComplete={ () => {
-							goToNextStep();
-						} }
-						onSkip={ () => {
-							goToNextStep();
-						} }
-						onError={ () => {
-							this.setState( {
-								extensionInstallError: true,
-								isInstallingExtensions: false,
-							} );
-						} }
-						autoInstall
-						pluginSlugs={ extensionsToInstall }
-					/>
+					<div className="woocommerce-profile-wizard__card-actions">
+						<Plugins
+							onComplete={ () => {
+								goToNextStep();
+							} }
+							onSkip={ () => {
+								goToNextStep();
+							} }
+							onError={ () => {
+								this.setState( {
+									extensionInstallError: true,
+									isInstallingExtensions: false,
+								} );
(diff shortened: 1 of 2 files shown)


joshuatf:
>Is that in purpose? Otherwise, do you want me to open an issue?

@Aljullu It does look like a smidge too much.  Does your "Retry" button also look vertically not centered?  Is that Firefox?

Aljullu:
> Does your "Retry" button also look vertically not centered? Is that Firefox?

Yes, it seems to happen in all buttons. I created an issue here: #3597.

Regarding the margin between the last plugin and the button being too big, it happens in Chrome too:

imatge

I think when [we hide the border](https://github.com/woocommerce/woocommerce-admin/blob/master/client/dashboard/profile-wizard/style.scss#L174), we should remove the padding or margin too. That makes it look better for me, at least.

joshuatf:
>I think when we hide the border, we should remove the padding or margin too. That makes it look better for me, at least.

Good call.  I hid the margin on the last child and that seems to match the designs perfectly.

PR #3589, opened 2020-01-20
Onboarding: Allow Jetpack connection to be skipped in tax task

### LINKED ISSUE #3567: Onboarding: Tax step can't be completed without Jetpack connection
**Describe the bug**
If Jetpack and WCS are already installed, the "Install" step shown in the design below is not shown.  This leaves only the "Connect" step currently which only provides one option ("Connect") leaving this task incomplete if the user does not connect.

We should do one of the following:
1. Add back in the "Enable automated tax calculations" step (removed since this step has duplicated efforts between the "Good news" screen).
1. Add a button to the "Connect" step to opt out and instead manually configure.
1. Still show the "Install" step even if both plugins are installed, but Jetpack is not connected (feels weird given the next step is "Connect).

I'm leaning towards option 2 above, but open to opinions.  /cc @jameskoster @pmcpinto 

**To Reproduce**

1. Install and activate Jetpack and WCS.
1. Disconnect Jetpack.
1. Visit the tax task.
1. Note the inability to complete the task without connecting Jetpack.

**Expected behavior**
Expected to be able to complete all tasks without connecting Jetpack.

**Screenshots**

#### Original design
Screen Shot 2020-01-15 at 5 23 49 PM

#### Current when WCS and Jetpack installed but not connected
Screen Shot 2020-01-15 at 4 54 48 PM

joshuatf:
Fixes #3567 

Adds a skip button for the connect step in the tax task and tracks event.

### Screenshots
Screen Shot 2020-01-20 at 1 57 08 PM

1. Make sure your store is in a TaxJar supported country.  (e.g., `US`).
2. Disconnect Jetpack.
3. Walk through the tax task to the "Connect" step.
4. Make sure the skip button is shown and works as expected.
5. Make sure using the skip or connect option record respective track events with the correct `connect` property.

### DIFF
--- a/client/dashboard/task-list/tasks/steps/connect.js
+++ b/client/dashboard/task-list/tasks/steps/connect.js
@@ -38,17 +38,22 @@ class Connect extends Component {
 	}
 
 	render() {
-		const { hasErrors, isRequesting } = this.props;
+		const { hasErrors, isRequesting, onSkip, skipText } = this.props;
 
-		return hasErrors ? (
-			<Button isPrimary onClick={ () => location.reload() }>
-				{ __( 'Retry', 'woocommerce-admin' ) }
-			</Button>
-		) : (
+		return (
 			<Fragment>
-				<Button isBusy={ isRequesting } isPrimary onClick={ this.connectJetpack }>
-					{ __( 'Connect', 'woocommerce-admin' ) }
-				</Button>
+				{ hasErrors ? (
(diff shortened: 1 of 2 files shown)


joshuatf:
Thanks for the review @Aljullu!

>The only minor issue I found is that when the Skip button is focused/active, it looks misaligned with the Connect button:

Good catch.  I'm seeing this Firefox, but not in Chrome so will follow-up in the Firefox button issue you opened.

PR #3590, opened 2020-01-20
Onboarding: Mark profiler complete on WC update

### LINKED ISSUE #3564: Onboarding: Update profiler to complete during WC upgrade
**Describe the bug**
When the onboarding feature flag is set to true for core, upgrading WC core could result in users having to go through the new onboarding experience after having already working through the old onboarding expereince.

See the following for more details:
https://github.com/woocommerce/woocommerce-admin/blob/5d5f55e3b19fcceb928ee14fb05a1b7d7554a4a6/src/Features/Onboarding.php#L108

**To Reproduce**

1. Make sure the `wc_onboarding_opt_in` option is set to `yes`.
1. Go through the old onboarding process.
1. "Upgrade" WC by setting the packaged version of wc-admin `onboarding` flag to `true` (and running necessary `npm run build` in wc-admin directory.
1. Note the new onboarding experience being shown.

**Expected behavior**
Expected to not see the profiler if the old onboarding has already been completed.

joshuatf:
Fixes #3564

Marks the profiler complete after WC is updated from an older version if the `install` notice is still shown (used to determine if the old onboarding should be shown).

Note that we'll need a follow-up in WC core on when onboarding is enabled by default in core to modify the check for non-new installs (todo note in the code).

1. On a new install (or one where the previous onboarding is still enabled), enable onboarding.
2. Force an upgrade by incrementing the version number found in `class-wooocommerce.php`.
3. Make sure the onboarding profiler is still shown.
4. Disable the new onboarding experience and complete the old one.
5. Enable the new onboarding experiencing manually via by setting the `wc_onboarding_opt_in` to `yes`.
6. Force another version update.
7. Make sure the new onboarding experience is not shown.
8. Enable the profiler again via the "Help" tab.
9. Fill out some data in the profiler and complete it.
10. Bump the WC core version again and make sure no data is lost in `woocommerce_onboarding_profile`.

### DIFF
--- a/src/Features/Onboarding.php
+++ b/src/Features/Onboarding.php
@@ -63,6 +63,7 @@ class Onboarding {
 		// Adds the ability to toggle the new onboarding experience on or off.
 		// @todo This option should be removed when merging the onboarding feature to core.
 		add_action( 'current_screen', array( $this, 'enable_onboarding' ) );
+		add_action( 'woocommerce_updated', array( $this, 'maybe_mark_complete' ) );
 
 		if ( ! Loader::is_onboarding_enabled() ) {
 			add_action( 'current_screen', array( $this, 'update_help_tab' ), 60 );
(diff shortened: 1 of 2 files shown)


psealock:
>(or one where the previous onboarding is still enabled)

How can I make this happen? Deactivate wc-admin and go through Core's old onboarding without completing it, then re-activate wc-admin?

joshuatf:
>How can I make this happen? Deactivate wc-admin and go through Core's old onboarding without completing it, then re-activate wc-admin?

The old onbvoarding considers itself active if the `install` notice exists.  Anything that qualifies as a "new install" should enable this on update (and possibly activation) which can occur if any of the following are true:
* No products exist
* No previous WC version exists
* No shop page exists

A new install is obviously easiest but if you want to do this on your current install, you can qualify with one of the above and then bump the version number in `class-woocommerce.php` to trigger an update.

PR #3612, opened 2020-01-22
Onboarding: Add missing space between manual tax info and link

### LINKED ISSUE #3604: Space is missing to separate two sentences in the setup wizard
**Describe the bug**
When configuring a new store, in the step to set up taxes, a space is missing to separate the phrases "By clicking 'Configure' you're enabling tax rates and calculations." and "More info here." See the screenshot below.

Screenshot from 2020-01-21 14-49-53

Those two strings are generated here:

https://github.com/woocommerce/woocommerce-admin/blob/ac0eadcaf8828903e9875305cc00c478e7293e19/client/dashboard/task-list/tasks/tax.js#L271

On a related note, for i18n purposes, I wonder if it would be better to combine the two sentences in a single string? Keep in mind that I'm not familiar with how `__()` works in the JS world, so I might be missing something.

**To Reproduce**

1. Go to /wp-admin/admin.php?page=wc-admin&task=tax
2. In the second step, see the string mentioned above.

joshuatf:
Fixes #3604

Adds a space between the two sentences underneath the "Configure" button.

### Screenshots
Screen Shot 2020-01-22 at 4 29 22 PM

1. Enable the task list and go to the tax task.
2. Force manual configuration by setting to a country outside of TaxJar supported countries.
3. Make sure the option `woocommerce_calc_taxes` is set to `no`.
4. Note the space between the sentences `By clicking "Configure" you're enabling tax rates and calculations. More info here.`

### DIFF
--- a/client/dashboard/task-list/tasks/tax.js
+++ b/client/dashboard/task-list/tasks/tax.js
@@ -276,7 +276,7 @@ class Tax extends Component {
 								interpolateComponents( {
 									mixedString: __(
 										'By clicking "Configure" you\'re enabling tax rates and calculations.' +
-											'More info {{link}}here{{/link}}.',
+											' More info {{link}}here{{/link}}.',
 										'woocommerce-admin'
 									),
 									components: {
--- a/client/dashboard/task-list/tasks/tax.js
+++ b/client/dashboard/task-list/tasks/tax.js
@@ -275,8 +275,9 @@ class Tax extends Component {
 							{ 'yes' !== generalSettings.woocommerce_calc_taxes &&
 								interpolateComponents( {
 									mixedString: __(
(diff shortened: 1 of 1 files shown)


timmyc:
> The translation file build script for JSX strips leading and trailing spaces from strings before concatenating them.

TIL

joshuatf:
>The translation file build script for JSX strips leading and trailing spaces from strings before concatenating them.

Had no idea, thanks for the explanation, @rrennick!  

I'll update this one to use a single line for the time being, but I do like the idea of not trimming white spaces.  Seems like this could lead to some confusion if we let any of these sneak in.

joshuatf:
Updated this temporarily to use a single non-concatenated line.  Opened an issue so we can address trimmed whitespaces [here](https://github.com/woocommerce/woocommerce-admin/pull/3612).

PR #3614, opened 2020-01-22
Add Timeline component

v18:
This adds a Timeline component that will be used in another WooCommerce project. The usage instructions can be found in the README that's been added in the `timeline` source folder.

image

### Accessibility

Make sure unit tests succeed:
- `npm test packages/components/src/timeline`

Do a round of manual testing using Storybook
- `npm run storybook`
- Make sure both the `Empty` and `Filled` stories work as expected.
- Mess around with the knobs and make sure component behaves as expected.

### DIFF
--- a/packages/components/src/timeline/index.js
+++ b/packages/components/src/timeline/index.js
@@ -9,7 +9,7 @@ import { __ } from '@wordpress/i18n';
 const Timeline = ( { className, items } ) => {
 	const timelineClassName = classnames( 'woocommerce-timeline', className );
 
-	if ( items.length === 0 ) {
+	if ( ! items || items.length === 0 ) {
 		return (
 			<div className={ timelineClassName }>
 				<p className="timeline_no_events">
--- a/packages/components/src/timeline/index.js
+++ b/packages/components/src/timeline/index.js
@@ -65,11 +65,16 @@ Timeline.propTypes = {
 			/**
 			 * Headline displayed for the list item.
 			 */
-			headline: PropTypes.element.isRequired,
+			headline: PropTypes.oneOfType( [
+				PropTypes.element,
+				PropTypes.string,
+			] ).isRequired,
 			/**
 			 * Body displayed for the list item.
 			 */
-			body: PropTypes.arrayOf( PropTypes.element ),
+			body: PropTypes.arrayOf(
+				PropTypes.oneOfType( [ PropTypes.element, PropTypes.string ] )
+			),
 			/**
 			 * Allows users to toggle the timestamp on or off.
 			 */
--- a/packages/components/src/timeline/index.js
+++ b/packages/components/src/timeline/index.js
@@ -6,30 +6,10 @@ import moment from 'moment';
 import PropTypes from 'prop-types';
 import { __ } from '@wordpress/i18n';
 
-const daysToTimelineItems = ( day, dayNum ) => {
-	const dayToTimelineItem = ( item, itemIndex ) => {
-		const itemTimeString = moment( item.datetime ).format( 'h:mma' );
-		const itemKey = dayNum + '-' + itemIndex;
-		return (
-			<li key={ itemKey }>
-				{ item.headline } <span>{ itemTimeString }</span>
-				{ item.body.map( function( line, bodyLineIndex ) {
-					const bodyLineKey =
-						dayNum + '-' + itemIndex + '-' + bodyLineIndex;
-					return <p key={ bodyLineKey }>{ line }</p>;
-				} ) }
-			</li>
-		);
-	};
-
-	const dayString = moment( dayNum.toString() ).format( 'MMMM D, YYYY' );
-	return (
-		<li key={ dayNum }>
-			{ dayString }
-			<ul>{ day.map( dayToTimelineItem ) }</ul>
-		</li>
-	);
-};
+/**
+ * Internal dependencies
+ */
+import TimelineGroup from './timeline-group';
 
 const Timeline = ( { className, items } ) => {
 	const timelineClassName = classnames( 'woocommerce-timeline', className );
@@ -45,14 +25,14 @@ const Timeline = ( { className, items } ) => {
 	}
 
 	// Sort all the items reverse chronologically
-	items.sort( function( a, b ) {
+	items.sort( ( a, b ) => {
 		return a.datetime - b.datetime;
 	} );
 
 	// Group the items into local day collections
 	const days = [];
 
-	items.forEach( function( item ) {
+	items.forEach( ( item ) => {
 		const itemLocalDatetime = moment.unix( item.datetime );
 		const dayNum = parseInt( itemLocalDatetime.format( 'YYYYMMDD' ), 10 );
 		if ( ! ( dayNum in days ) ) {
@@ -61,9 +41,13 @@ const Timeline = ( { className, items } ) => {
 		days[ dayNum ].push( item );
 	} );
 
+	const timelineGroups = days.map( ( day, dayIndex ) => (
+		<TimelineGroup key={ dayIndex } items={ day } groupKey={ dayIndex } />
+	) );
+
 	return (
 		<div className={ timelineClassName }>
-			<ul>{ days.map( daysToTimelineItems ) }</ul>
+			<ul>{ timelineGroups }</ul>
 		</div>
 	);
 };
--- a/packages/components/src/timeline/index.js
+++ b/packages/components/src/timeline/index.js
@@ -1,7 +1,105 @@
-const Timeline = () => {
+/** @format */
+/**
+ * External dependencies
+ */
+import classnames from 'classnames';
+import moment from 'moment';
+import PropTypes from 'prop-types';
+import { __ } from '@wordpress/i18n';
+
+const Timeline = ( { className, items } ) => {
+	const timelineClassName = classnames( 'woocommerce-timeline', className );
+
+	// Sort all the items reverse chronologically
+	items.sort( function( a, b ) {
+		return a.datetime - b.datetime;
+	} );
+
+	// Group the items into local day collections
+	const days = [];
+
+	items.forEach( function( item ) {
+		const itemLocalDatetime = moment.unix( item.datetime );
+		const dayNum = parseInt( itemLocalDatetime.format( 'YYYYMMDD' ), 10 );
+		if ( ! ( dayNum in days ) ) {
+			days[ dayNum ] = [];
+		}
+		days[ dayNum ].push( item );
+	} );
+
+	if ( 0 === items.count ) {
+		return (
+			<div className={ timelineClassName }>
+				<p class="timeline_no_events">
+					{ __( 'No events to display' ) }
+				</p>
+			</div>
+		);
+	}
+
 	return (
-		<div>Hello world.</div>
+		<div className={ timelineClassName }>
+			<ul>
+				{ days.map( function( day, dayNum ) {
+					const dayString = moment( dayNum ).format( 'MMMM D, YYYY' );
+					return (
+						<li key={ dayNum }>
+							{ dayString }
+							<ul>
+								{ day.map( function( item ) {
+									const timeString = moment( item.datetime ).format( 'h:mma' );
+									return (
+										<li>
+											{ item.headline } <span>{ timeString }</span>
+											{ item.body.map( function( line ) {
+												return (
+													<p>
+														{ line }
+													</p>
+												);
+											} ) }
+										</li>
+									);
+								} ) }
+							</ul>
+						</li>
+					);
+				} ) }
+			</ul>
+		</div>
 	);
+};
+
+Timeline.propTypes = {
+	/**
+	 * Additional CSS classes.
+	 */
+	className: PropTypes.string,
+	/**
+	 * An array of list items.
+	 */
+	items: PropTypes.arrayOf(
+		PropTypes.shape( {
+			/**
+			 * Timestamp (in seconds) for the timeline item.
+			 */
+			datetime: PropTypes.number.isRequired,
+			/**
+			 * GridIcon for the Timeline item.
+			 */
+			gridicon: PropTypes.string.isRequired,
+			/**
+			 * Headline displayed for the list item.
+			 */
+			headline: PropTypes.string.isRequired,
+			/**
+			 * Body displayed for the list item.
+			 */
+			body: PropTypes.arrayOf(
+				PropTypes.string
+			),
+		} )
+	).isRequired,
 };
 
 export default Timeline;
--- a/packages/components/src/timeline/index.js
+++ b/packages/components/src/timeline/index.js
@@ -1,0 +1,7 @@
+const Timeline = () => {
+	return (
+		<div>Hello world.</div>
+	);
+};
+
+export default Timeline;
--- a/packages/components/src/timeline/index.js
+++ b/packages/components/src/timeline/index.js
@@ -9,7 +9,6 @@ import { __ } from '@wordpress/i18n';
 /**
  * Internal dependencies
  */
-import Card from '../card';
 import TimelineGroup from './timeline-group';
 
 export const groupByOptions = {
@@ -31,11 +30,11 @@ const Timeline = ( props ) => {
 	// Early return in case no data was passed to the component.
 	if ( ! items || items.length === 0 ) {
 		return (
-			<Card title={ 'Timeline' } className={ timelineClassName }>
-				<p className="timeline_no_events">
+			<div className={ timelineClassName }>
+				<p className={ 'timeline_no_events' }>
 					{ __( 'No data to display', 'woocommerce-admin' ) }
 				</p>
-			</Card>
+			</div>
 		);
 	}
 
@@ -77,9 +76,9 @@ const Timeline = ( props ) => {
 		.map( ( g ) => <TimelineGroup key={ g.id.toString() } group={ g } /> );
 
 	return (
-		<Card title={ 'Timeline' } className={ timelineClassName }>
+		<div className={ timelineClassName }>
 			<ul>{ timelineGroups }</ul>
-		</Card>
+		</div>
 	);
 };
 
--- a/packages/components/src/timeline/index.js
+++ b/packages/components/src/timeline/index.js
@@ -29,20 +29,21 @@ const Timeline = ( { className, items } ) => {
 		return a.datetime - b.datetime;
 	} );
 
-	// Group the items into local day collections
-	const days = [];
+	// Group items by days.
+	// TODO: Maybe the items passed into the Timeline component
+	// should already be grouped and sorted?
+	const groups = items.reduce( ( acc, curr ) => {
+		const itemLocalDatetime = moment.unix( curr.datetime );
+		const formattedDate = itemLocalDatetime.format( 'YYYYMMDD' );
 
-	items.forEach( ( item ) => {
-		const itemLocalDatetime = moment.unix( item.datetime );
-		const dayNum = parseInt( itemLocalDatetime.format( 'YYYYMMDD' ), 10 );
-		if ( ! ( dayNum in days ) ) {
-			days[ dayNum ] = [];
-		}
-		days[ dayNum ].push( item );
-	} );
+		return {
+			...acc,
+			[ formattedDate ]: [ ...( acc[ formattedDate ] || [] ), curr ],
+		};
+	}, {} );
 
-	const timelineGroups = days.map( ( day, dayIndex ) => (
-		<TimelineGroup key={ dayIndex } items={ day } groupKey={ dayIndex } />
+	const timelineGroups = Object.keys( groups ).map( ( key ) => (
+		<TimelineGroup key={ key } items={ groups[ key ] } groupKey={ key } />
 	) );
 
 	return (
--- a/packages/components/src/timeline/index.js
+++ b/packages/components/src/timeline/index.js
@@ -2,7 +2,6 @@
  * External dependencies
  */
 import classnames from 'classnames';
-import moment from 'moment';
 import PropTypes from 'prop-types';
 import { __ } from '@wordpress/i18n';
 
@@ -10,21 +9,10 @@ import { __ } from '@wordpress/i18n';
  * Internal dependencies
  */
 import TimelineGroup from './timeline-group';
-
-export const groupByOptions = {
-	DAY: 'day',
-	WEEK: 'week',
-	MONTH: 'month',
-};
-
-export const orderByOptions = {
-	ASC: 'asc',
-	DESC: 'desc',
-};
+import { sortByDateUsing, groupItemsUsing } from './util';
 
 const Timeline = ( props ) => {
 	const { className, items, groupBy, orderBy } = props;
-
 	const timelineClassName = classnames( 'woocommerce-timeline', className );
 
 	// Early return in case no data was passed to the component.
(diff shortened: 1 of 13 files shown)


packages/components/src/timeline/index.js
```diff
@@ -0,0 +1,7 @@
+const Timeline = () => {
+	return (
+		<div>Hello world.</div>
```
timmyc: This string needs to be passed through the translation method. :trollface: I jest - excited to see this contribution getting started and happy to help in any way @v18

v18: LOL. thank you 🙏

allendav:
@v18 - any updates?

joshuatf:
Hi @v18 👋 - just checking to see if this is still being worked on.

allendav:
@joshuatf - it is

allendav:
@reykjalin @v18 @marcinbot @LevinMedia 

I've pushed a work in progress that adds accepting a prop of items, sorts them into local dates, and begins rendering them. 100% untested.

packages/components/src/timeline/index.js
```diff
@@ -0,0 +1,106 @@
...
+	items.forEach( function( item ) {
+		const itemLocalDatetime = moment.unix( item.datetime );
+		const dayNum = parseInt( itemLocalDatetime.format( 'YYYYMMDD' ), 10 );
+		if ( ! ( dayNum in days ) ) {
+			days[ dayNum ] = [];
+		}
+		days[ dayNum ].push( item );
+	} );
+
+	if ( 0 === items.length ) {
```
timmyc: Could we move this up earlier in the class here to return the "No Events" immediately when we know `items` is empty?

packages/components/src/timeline/index.js
```diff
@@ -0,0 +1,106 @@
...
+				</p>
+			</div>
+		);
+	}
+
+	return (
+		<div className={ timelineClassName }>
+			<ul>
+				{ days.map( function( day, dayNum ) {
+					const dayString = moment( dayNum ).format( 'MMMM D, YYYY' );
```
timmyc: Should we consider using localized date formats here? I think the immediate use case might be US only iirc, but would be nice to keep it less locale specific this being a component that will be reused elsewhere.

timmyc:
oh and maybe a rebase on this branch to get CI happy again 🙇

packages/components/src/timeline/index.js
psealock: There is now the date setting from site settings:

```js
import { getSetting } from '@woocommerce/wc-admin-settings';
import { format } from '@wordpress/date';

const dateFormat = getSetting( 'dateFormat' );
const date = format( dateFormat, new Date() );
```
As mentioned above, the component shouldn't interact with settings. Instead the format can be passed down as props.

allendav: @LevinMedia - your design suggested the format given. What say ye?

packages/components/src/timeline/index.js
allendav: Fixed in https://github.com/woocommerce/woocommerce-admin/pull/3614/commits/ce9069e3c8e2835992a279000d78ebc85590ea7d

reykjalin:
I'm currently using Storybook to properly see what I'm doing (loving that so far btw), is that a recommended approach here?

<hr />

How do we feel about splitting the Timeline component up into smaller pieces?

Following the designs posted in paJDYF-vb-p2 by @LevinMedia I'm thinking something along the lines of 
- `<TimelineItem>` as a base component (individual items in designs).
- `<TimelineGroup>` that groups together several `<TimelineItem>`s (the day groupings in designs)
- `<Timeline>`, which contains a list of `<TimelineGroup>`s (the whole component in designs)

The only thing a user would see is really just the `<Timeline>` component, and provides a structured list of items for the Timeline.

We could also just have several functions in the `<Timeline>` component that take care of actually structuring the data, I'm simply wondering what would be the best way to maintain this.

<hr />

Sidenote; I say `TimelineGroup` here above (generic), but would we be better served by making this `TimelineDay` or something similar? This is a timeline after all, so days make sense as groupings, but so could weeks/months/years in a wider context.
Forcing days would be easier to implement for now I think, but it does limit the component _somewhat_ going forward.

cc @timmyc @psealock

timmyc:
@reykjalin 

```

    <TimelineItem> as a base component (individual items in designs).
    <TimelineGroup> that groups together several <TimelineItem>s (the day groupings in designs)
    <Timeline>, which contains a list of <TimelineGroup>s (the whole component in designs)

```

That all sounds and looks good to me. I think that will help keep the code a bit more tidy as well, and allow for some targeted tests around each distinct component. I don't see anyone ever importing and using `<TimelineGroup/>` or `<TimelineItem/>` directly, but making them distinct components consumed and used in `<Timeline />` seems totally reasonable to me.

packages/components/src/timeline/index.js
```diff
@@ -0,0 +1,86 @@
...
+	// should already be grouped and sorted?
+	const groups = items.reduce( ( acc, curr ) => {
+		const itemLocalDatetime = moment.unix( curr.datetime );
+		const formattedDate = itemLocalDatetime.format( 'YYYYMMDD' );
+
+		return {
+			...acc,
+			[ formattedDate ]: [ ...( acc[ formattedDate ] || [] ), curr ],
+		};
+	}, {} );
```
reykjalin: I think it makes more sense that the items are passed to the timeline grouped and sorted, but I've implemented it like this for now.

An alternative would be to pass something like a flag to the component to indicate whether to group+sort by hour/days/weeks/months/years/etc, but that seems unnecessarily complicated. It's nice for the user, but difficult to implement properly in the component.
The user will arguably have more freedom if they get to group+sort the data however they want and the component doesn't mess with the data.

@timmyc @psealock @allendav I'd love to hear your thoughts on what you think makes the most sense here.

timmyc: What about having the group by functionality be triggered by a prop? Something like `groupBy` that defaults to "day" maybe? And likewise have a prop for sorting order ASC|DESC Then the data could be passed in ungrouped/sorted and the component would take care of the rest. This would also allow for future controls to group items by week, month, quarter etc...

reykjalin: Sounds good, I'll do that then 🙂

reykjalin: Done in https://github.com/woocommerce/woocommerce-admin/pull/3614/commits/17a835c62c97cca053e0c99cac3f39577a72a8a8 👍

packages/components/src/timeline/index.js
```diff
@@ -0,0 +1,131 @@
...
+			timeToMoment( g.id ).isSame( timeToMoment( i.datetime ), groupBy );
+
+		const groupIndex = groups.findIndex( ( g ) =>
+			haveSameMoment( g, item )
+		);
+		if ( groupIndex < 0 ) {
+			return [
+				...groups,
+				{
+					id: item.datetime,
```
dwainm: Why are we switching from calling the data `datetime` to `id`? When reading the sorting function below, seeing `.id` being used confused me a little bit, but then I noted that ID is actually `datetime `.

LevinMedia:
@reykjalin  Looking great!  👍  I've got one overarching piece of feedback to start, but I'll probably come back with some more later today. 

Screen Shot 2020-03-25 at 10 06 12 AM

Right now the timeline is rendering in chronological order, it should be in reverse chronological order - the most recent things always on top.

reykjalin:
@LevinMedia you can actually change the ordering by passing in `'desc'` as the value for the `groupBy` prop like so:

```jsx
import { orderByOptions } from './Timeline';

<Timeline items={ items } orderBy={ orderByOptions.DESC } />
```

`orderByOptions.DESC` is just a convenience variable that contains the string `desc`.

LevinMedia:
Excited to see this coming along! 🎉 
I've made a few small css tweaks where I could. A few things that still need to be resolved/addressed:

- There should only ever be an 8px gap between the top or bottom of background of the icon, and the vertical line that connects all the events
- the vertical line should extend from below the date stamp down to the first event in a group. 
- An event should be able to specify a color for the background of the icon, but it's only used if it's the most recent event. 
- ensure we can use material icons
- Ensure we can add links into event titles
- Ensure we can add other components like buttons into the event body.

reykjalin:
@LevinMedia update on what you've outlined above:

- [x] There should only ever be an 8px gap between the top or bottom of background of the icon, and the vertical line that connects all the events
- [x] the vertical line should extend from below the date stamp down to the first event in a group.
- [x] Ensure we can add links into event titles
- [x] Ensure we can add other components like buttons into the event body.

>An event should be able to specify a color for the background of the icon, but it's only used if it's the most recent event.

This shouldn't be too difficult; I'll just need to change the data passed in for the icon a little bit.

>ensure we can use material icons

I have a couple of questions about this:

1. Are you referring to [these icons](https://material-ui.com/components/material-icons/)?
1. Are we supporting both GridIcons **and** material icons?
    - I imagine there might be some overlap in names, so it might be difficult to support both if we're only passing the name of the icon to the timeline. If we have users pass in the actual icon component they want to use, supporting anything as an icon is not an issue.
    - If we want to make using material icons easy we may need to [include a font](https://material-ui.com/components/icons/#icon-font-icons) that has them already, otherwise we'll need to include the whole icon set (1100 icons) and a huuuuuuge switch statement. cc @psealock @timmyc if you have any input here.

packages/components/src/timeline/index.js
reykjalin: Good catch, I can see how that is confusing.

The reason I went with ID is that the datetime is used to identify which timeline group this is and where it fits in the overall timeline, and it didn't feel semantically correct to call it `datetime`. Your point is valid though; this is definitely confusing.

Do you have any suggestions here? I'd rather not call this `datetime` because this isn't really about the date; it's a way to identify this group and sort the list of groups.
Maybe `key` would be better?

dwainm: The only confusing part here is that it is used for sorting as well. I think in this case it would maybe be good to just add comment around the sorting by ID to identify the type of data?  Besides that, I understand your reasoning.

LevinMedia:
>If we have users pass in the actual icon component they want to use, supporting anything as an icon is not an issue.

@reykjalin Was chatting about this with @jameskoster about this a bit more this am - passing in the icon component is probably the best bet - then we can pull from the core icon package for our specific needs:

**Success (Payment authorized, payment charged, evidence submitted):**
https://github.com/WordPress/gutenberg/blob/master/packages/icons/src/library/check.js

**info (New dispute, dispute won / lost, charge failures):**
https://github.com/WordPress/gutenberg/blob/master/packages/icons/src/library/info.js

**Plus (Added to deposit)**
https://github.com/WordPress/gutenberg/blob/master/packages/icons/src/library/plus.js

**Undo (Successful refund)**
https://github.com/WordPress/gutenberg/blob/master/packages/icons/src/library/undo.js

**Minus icon (Removed from deposit)**
https://github.com/WordPress/gutenberg/blob/master/packages/icons/src/library/minus.js

**Undo icon (Refund)**
https://github.com/WordPress/gutenberg/blob/master/packages/icons/src/library/undo.js

packages/components/src/timeline/timeline-item.js
```diff
@@ -0,0 +1,71 @@
+/**
+ * External dependencies
+ */
+import classnames from 'classnames';
+import moment from 'moment';
+import PropTypes from 'prop-types';
+import GridIcon from 'gridicons';
+
+const GRIDICON_SIZE = 16;
```
marcinbot, line 9: I think it might be worth bumping this to at least 18, since the icons are pretty small:
Screenshot 2020-04-14 at 18 00 47

packages/components/src/timeline/index.js
```diff
@@ -0,0 +1,131 @@
...
+		orderByOptions.ASC === orderBy ? sortAscending : sortDescending;
+
+	// Group items together, sort, and map to Timeline groups.
+	const timelineGroups = items
+		.reduce( itemsToGroups, [] )
+		.sort( sortMethod )
+		.map( ( g ) => <TimelineGroup key={ g.id.toString() } group={ g } /> );
+
+	return (
+		<Card title={ 'Timeline' } className={ timelineClassName }>
```
marcinbot, line 80: Would be good to be able to change the title here.

packages/components/src/timeline/index.js
```diff
@@ -0,0 +1,131 @@
...
+			 */
+			datetime: PropTypes.number.isRequired,
+			/**
+			 * GridIcon for the Timeline item.
+			 */
+			gridicon: PropTypes.string.isRequired,
+			/**
+			 * Headline displayed for the list item.
+			 */
+			headline: PropTypes.string.isRequired,
```
marcinbot, line 107: Sometimes we want to add a link to the headline, for example using `createInterpolateElement`. This results in the headline being and object and not a string which leads to a warning in the console. We might want to update here, in other places and the body element to allow embedding React elements.

packages/components/src/timeline/index.js
marcinbot: Looking into this further, it could make the component more flexible if we were to get rid of the `Card` entirely and leave it to the consumer to embed it in one, if required.

allendav:
Thank you for the review @marcinbot !

@reykjalin - we probably should include a unit test to ensure things render in the expected order

packages/components/src/timeline/timeline-item.js
```diff
@@ -0,0 +1,71 @@
...
+
+TimelineItem.propTypes = {
+	/**
+	 * Additional CSS classes.
+	 */
+	className: PropTypes.string,
+	/**
+	 * An array of list items.
+	 */
+	item: PropTypes.shape( {
```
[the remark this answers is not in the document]
reykjalin: Hmmm, I think this might have to be a bit more complicated than providing a property for a class, or rather we need to reduce the scope and simplify our implementation.
If we just include a `class` prop that implies it will be applied to the whole item, i.e. both the icon and the associated text. So I think it'd be better to do one of two things:

1. Expand the `gridicon` prop to an object that has `icon` and `color` props. `icon` will contain the actual `<GridIcon />` element, whereas `color` contains the CSS color value.
2. The `gridicon` prop contains the `<GridIcon />` element, and we introduce a new prop, `iconColor` that will contain the CSS color value.

Personally I'm leaning towards option 2. What do you think?

packages/components/src/timeline/timeline-item.js
reykjalin: >they are also off-center
That's interesting! They're centered in Storybook 🤔. I'd be interested in seeing what's happening differently in wcpay.

I'll reach out in Slack so we can chat about this in more detail.

>18, since the icons are pretty small

No problem! This seems bigger in Storybook, so again I'd ben interested in seeing what's different in WCPay.

reykjalin: Screenshot:

image

packages/components/src/timeline/index.js
reykjalin: Yeah that sounds good.

Changed in https://github.com/woocommerce/woocommerce-admin/pull/3614/commits/45927f5aac4121f955a60e16791dbee954f11b6c 👍

packages/components/src/timeline/style.scss
```diff
@@ -0,0 +1,114 @@
...
+					background: $valid-green;
+				}
+
+				&.is-warning {
+					background: $notice-yellow;
+				}
+
+				&.is-error {
+					background: $error-red;
+				}
```
reykjalin, line 83: See https://github.com/Automattic/woocommerce-payments/pull/344#discussion_r413108027 for discussions in WCPay on whether styling icons like this here is appropriate or not. Feel free to add additional comments below either here or in WCPay.

packages/components/src/timeline/index.js
reykjalin: Just pinging @LevinMedia to bring his attention to this.

packages/components/src/timeline/timeline-item.js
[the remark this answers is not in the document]
reykjalin: I think this has been properly implemented so far, what do you think @marcinbot ?

packages/components/src/timeline/timeline-item.js
reykjalin: This has been resolved, both as a discussion in p1587567205003600-slack-wcpay-events and with some stricter CSS.

packages/components/src/timeline/index.js
reykjalin: The final commit to get this change live was made with https://github.com/woocommerce/woocommerce-admin/pull/3614/commits/9b10bbaab00ca4d555bf30cddb9eac6e348808b0

packages/components/src/timeline/style.scss
```diff
@@ -0,0 +1,89 @@
...
+			font-size: 14px;
+			line-height: 16px;
+		}
+
+		.woocommerce-timeline-item__body {
+			display: flex;
+			flex-direction: column;
+
+			color: $studio-gray-60;
+			font-size: 14px;
```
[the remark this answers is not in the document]
reykjalin: Discussion and resolution in p1587567205003600-slack-wcpay-events.
Abstract: There are some upcoming changes to general layout that will make the font sizes fit better into the current designs. We'll wait for those before changing anything here.

packages/components/src/timeline/timeline-item.js
```diff
@@ -0,0 +1,71 @@
...
+		 */
+		gridicon: PropTypes.string.isRequired,
+		/**
+		 * Headline displayed for the list item.
+		 */
+		headline: PropTypes.string.isRequired,
+		/**
+		 * Body displayed for the list item.
+		 */
+		body: PropTypes.arrayOf( PropTypes.string ),
```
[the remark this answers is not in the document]
reykjalin: Fixed in df4952903a2e9bfdeefa847cd45020d0fe04ceba.

packages/components/src/timeline/index.js
reykjalin: I've refactored the code quite significantly, which includes just calling this `datetime` and then renaming the sorting function to `sortByDate`.

If you don't mind @dwainm could you take another look and let me know whether I just threw the readability out the window or if the code has been improved? 🙂

packages/components/src/timeline/util.js
```diff
@@ -0,0 +1,62 @@
...
+const groupByOptions = {
+	DAY: 'day',
+	WEEK: 'week',
+	MONTH: 'month',
+};
+
+const sortAscending = ( groupA, groupB ) => groupA.datetime - groupB.datetime;
+const sortDescending = ( groupA, groupB ) => groupB.datetime - groupA.datetime;
+
+const sortByDateUsing = ( orderBy ) => {
```
dwainm, line 20: Thank you @reykjalin. I like that you're being more explicit here.  This reads nicely.

reykjalin:
I think the PR is finally at a point where it can be seriously reviewed. I still need to add tests, but it seems to work as intended in manual testing.

What follows should be:

1. Design review, i.e. does the component as it is right now make sense in terms of look and function?
1. Code review; is the coding style ok? Are there any bugs?
1. Testing; does the component _actually_ work as intended?

packages/components/src/timeline/index.js
reykjalin: Resolving this as per https://github.com/woocommerce/woocommerce-admin/pull/3614#discussion_r413855447

allendav:
Thank you @reykjalin ! @joshuatf @timmyc - what's Isotope's process for un-drafting this and kicking off the review?

packages/components/src/timeline/index.js
LevinMedia: @allendav @reykjalin Yeah if we can localize the date format, that would be good. Thanks!

package-lock.json
```diff
@@ -21143,7 +21143,7 @@
       "dev": true
     },
     "prettier": {
-      "version": "1.19.1",
+      "version": "npm:wp-prettier@1.19.1",
```
marcinbot, line 21146: We probably don't need to modify `package.json` as part of this PR. Was this committed by mistake?

packages/components/src/timeline/timeline-item.js
```diff
@@ -0,0 +1,77 @@
...
+ */
+import classnames from 'classnames';
+import moment from 'moment';
+import PropTypes from 'prop-types';
+
+const TimelineItem = ( props ) => {
+	const { item, className } = props;
+
+	const itemClassName = classnames( 'woocommerce-timeline-item', className );
+	const itemTimeString = moment.unix( item.datetime ).format( 'h:mma' );
```
marcinbot, line 12: In addition to localizing the date requested here https://github.com/woocommerce/woocommerce-admin/pull/3614#discussion_r417694303 we should also think about localizing the time.

package-lock.json
reykjalin: Yes, this was accidental! Fixed in b41d04960e07bc4fde2c49746c00416a886d88b3.

packages/components/src/timeline/index.js
reykjalin: @timmyc @psealock is there a reason to use `format` from `@wordpress/date` over the [moment localization](https://momentjs.com/docs/#/displaying/format/) options?

Also, if the component should not interact with the settings, how do we get the default format? Should we hard code the default format to `F j, Y` (what it is now), or is there a way to get the localized date format without interacting with the settings?

psealock: >is there a reason to use format from @wordpress/date over the moment localization options?

We decided not to use moment localizations because the file size in substantial.

>how do we get the default format?

Can the component accept a `dateFormat` prop? We don't want to have components tied to settings because then they can't be reused outside of this context.

reykjalin: Yeah, a `dateFormat` prop is easy to add, but I’m wondering what the default value of that prop should be.
I don’t think it should be a required prop, and thus we should provide a default value.

marcinbot: [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Dates)'s `YYYY-MM-DD`? 

Similarly, for the time format, which we should also think about, use `hh:mm` and the 24-hour clock.

By default the format won't include month names, but if the component consumer wants to override that, they can do so by providing the prop.

reykjalin: Fixed in:

- https://github.com/woocommerce/woocommerce-admin/pull/3614/commits/bdea8da421766d67505a6bde345e93fefa56b5f6
- https://github.com/woocommerce/woocommerce-admin/pull/3614/commits/a4ec62031548bb12491f4b4be92df77840af201b
- https://github.com/woocommerce/woocommerce-admin/pull/3614/commits/c0a0303124113aaa67a241c4a744dac3ad393e0e

Default date is set to what @LevinMedia's design show in Figma: `January 1, 2020`.

@psealock @timmyc I think we're ready to ship this, pending an accepted review.

packages/components/src/timeline/timeline-item.js
reykjalin: This has been fixed as of https://github.com/woocommerce/woocommerce-admin/pull/3614/commits/c0a0303124113aaa67a241c4a744dac3ad393e0e :)

reykjalin:
I think we're ready for a proper review on this. I can't think of anything that's missing functionality-wise right now?

cc @allendav @marcinbot @timmyc @psealock

reykjalin:
>This is really well written. Good job team 🎉

Thank you ❤️

>Likewise, we now using https://github.com/WordPress/gutenberg/tree/master/packages/icons for icons.

Have the icons been merged, or are they waiting on a PR like the base-styles?
If they've been merged and we'd like to use them in the stories, where can I find documentation on what icons are available? I can't find any information about them 😕.

>The pull request to add in the newest base styles hasn't been merged yet, so I think this is good to go as is. Maybe make a note that it'll need attention at some point.

Do you usually add notes like this in code, or do you open issues? Where do I make a note of this?

>Also, can you add a line to the changelog?

Changelog entry added in 7480d26fe91f5d3f0d040255a959aab1443e21aa.

cc @psealock

timmyc:
@reykjalin are we good to merge this in? Typically the author of the PR does the merge after approval.

reykjalin:
Sorry @psealock, this got drowned in my backlog and I forgot to click merge. Merging now!

PR #3616, opened 2020-01-22
Update PHP 5.6 in Travis test matrix to WC 3.8.1

rrennick:
Unit tests are failing the PHP 5.6 test suite because WooCommerce 3.9.0 requires PHP 7.0. This PR pins the PHP 5.6 run from the test matrix to WooCommerce 3.8.1.

Ref: https://github.com/woocommerce/woocommerce/blob/master/readme.txt#L6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ matrix:
     env: WP_VERSION=latest WP_MULTISITE=0 WP_CORE_DIR=/tmp/wordpress
   - name: "PHP 5.6 unit tests"
     php: 5.6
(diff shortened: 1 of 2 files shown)


jeffstieler:
We claim support of WC `3.6`+ - should we not just modify the PHP 5.6 test to use WC `3.8.x`?

https://github.com/woocommerce/woocommerce-admin/blob/master/readme.txt#L38

rrennick:
@jeffstieler Thanks for that suggestion. This is ready for another look.

PR #2900, opened 2019-09-06
Hook up download button to export endpoint.

jeffstieler:
In support of #311. 

Also moves comparison, search, and download actions to ReportTable component.

**_NOTE:_**

This PR now contains the contents of https://github.com/woocommerce/woocommerce-admin/pull/2951 and https://github.com/woocommerce/woocommerce-admin/pull/2899 as well. It will serve as the roll-up PR for all CSV download UI changes.

This PR seeks to move the `<Card>` actions from the published `TableCard` component to the `ReportTable` component in WooCommerce Admin where the actions are used.

The rationale here being:
- There was too much of the WooCommerce Admin specific report UX in the `TableCard` component (it even used the `woocommerce-analytics__card` class, which would probably break for consumers)
- We will likely need to make changes to the download behavior, which was solely in the `TableCard` component

In addition to moving the compare, search, and download actions to `ReportTable`, this PR also moves the IE-specific styles for those items which is a new feature for WooCommerce Admin. A new Webpack entry `ie` has been created so that a `dist/ie/style.css` is created. The `FixStyleOnlyEntriesPlugin` plugin prunes the unneeded `dist/ie/index.js` that Webpack is compelled to create.

### Detailed test instructions:

- Check all Analytics tables, verify they function as before
- Open a report with more only 1 page of results
- Click download
- Verify a .csv file is immediately downloaded
- Open a report with more only 1 page of results
- Click download
- Verify a notice is displayed stating you'll receive an email
- Verify you receive an email, and the download link works

### Changelog Note:

Tweak: Moved WC-Admin specific actions from TableCard to ReportTable.

PR #3618, opened 2020-01-23
Fix Search component token display in ReportTable

jeffstieler:
Fixes a regression from https://github.com/woocommerce/woocommerce-admin/pull/2900.

This PR seeks to fix the display of search terms/tokens in report tables. Additionally, it fixes the styling of the "clear all" button at the right of the input.

### Screenshots

**Before**

Screen Shot 2020-01-22 at 5 47 24 PM

**After**

Screen Shot 2020-01-22 at 5 44 16 PM

- Open any report with a `<Search>` at the top of the `<ReportTable>`, like Customers report
- Search for a term, select a result
- Verify that a removable token is displayed for the selection

### Changelog Note:

Fix: report table search component.

### DIFF
--- a/client/analytics/components/report-table/index.js
+++ b/client/analytics/components/report-table/index.js
@@ -317,8 +317,9 @@ class ReportTable extends Component {
 		const totals = get( primaryData, [ 'data', 'totals' ], {} );
 		const totalResults = items.totalResults;
 		const downloadable = 0 < totalResults;
-		const searchWords = getSearchWords( query );
-		const searchedLabels = searchWords.map( v => ( { id: v, label: v } ) );
+		// Search words are in the query string, not the table query.
+		const searchWords = getSearchWords( this.props.query );
+		const searchedLabels = searchWords.map( v => ( { key: v, label: v } ) );
 
 		/**
 		 * Filter report table.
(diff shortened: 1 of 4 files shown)


client/analytics/components/report-table/index.js
```diff
@@ -317,8 +317,9 @@ class ReportTable extends Component {
 		const totals = get( primaryData, [ 'data', 'totals' ], {} );
 		const totalResults = items.totalResults;
 		const downloadable = 0 < totalResults;
-		const searchWords = getSearchWords( query );
-		const searchedLabels = searchWords.map( v => ( { id: v, label: v } ) );
+		// Search words are in the query string, not the table query.
+		const searchWords = getSearchWords( this.props.query );
```
psealock: Destructuring `query` from `this.props` above would be cleaner.

psealock:
You can rebase after https://github.com/woocommerce/woocommerce-admin/pull/3616 is merged to get tests to pass.

client/analytics/components/report-table/index.js
jeffstieler: I didn't go that route because we're destructuring `query` from `tableData` on line 308.

PR #3623, opened 2020-01-24
Onboarding: Fix shipping rates overlapping currency prefix

### LINKED ISSUE #3606: [WC 3.9 rc4] Setup taxes: currency symbol overlaps with shipping cost value
**Describe the bug**
The currency symbol and the shipping cost value are overlapping on the "Set up shipping" step of the Dashboard provided by WooCommerce Admin. See screenshot below:

Screenshot from 2020-01-21 14-42-21

**To Reproduce**

1. Set the store currency in /wp-admin/admin.php?page=wc-settings to any currency that has a currency symbol with two or more characters. I tested with Brazilian reais.  
2. Go to /wp-admin/admin.php?page=wc-admin&task=shipping
3. See that, on the second step ("Set shipping costs"), the currency symbol and the shipping cost value are overlapping.

* OS: Linux
* Browser: Firefox
* Version: 72.0.1

joshuatf:
Fixes #3606 

* Adds several state classes to `TextControlWithAffixes`
* Replaces the `TextControl` component in rates with `TextControlWithAffixes`
* Always formats shipping rates and always shows labels.

This forces users to always enter an amount for a shipping amount or set to 0 for free shipping.  I feel this is less ambiguous, but open for feedback. /cc @jameskoster @pmcpinto 

#### Before
Screen Shot 2020-01-24 at 4 57 56 PM

#### After
Screen Shot 2020-01-24 at 4 57 04 PM

1. Try this PR with multiple currencies, preferably some with multiple characters like the Brazilian reais.
1. Enable the task list and visit the shipping step.
1. Make sure the currency value and currency symbol don't overlap and are always spaced appropriately.
1. Make sure a properly formatted value is always used for a shipping rate.

### DIFF
--- a/client/dashboard/task-list/style.scss
+++ b/client/dashboard/task-list/style.scss
@@ -193,15 +193,29 @@
 			margin-bottom: 0;
 		}
 
+		.components-base-control__label {
+			display: block;
+			position: relative;
+			top: -8px;
+			width: 100%;
+			font-size: 12px;
+		}
+
 		.text-control-with-affixes__prefix,
 		.text-control-with-affixes__suffix {
-			display: none;
 			font-size: 16px;
 			line-height: 24px;
 			color: $studio-gray-50;
 			border: 0;
 			padding: 0;
 			align-items: center;
+			display: flex;
+			top: -12px;
+		}
+
+		.components-text-control__input {
+			position: relative;
+			top: -12px;
 		}
 
 		.text-control-with-affixes__prefix {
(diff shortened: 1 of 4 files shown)


jameskoster:
@joshuatf does the cost default to 0? If so I think this is good. Otherwise we might need to add a helper text beneath the input to say something like "Set price to 0 for free shipping".

joshuatf:
>does the cost default to 0? If so I think this is good.

👍 @jameskoster It does default to 0 if nothing exists.  So it's never blank/empty anymore.

pmcpinto:
@joshuatf the solution you proposed works perfectly 👍

PR #3632, opened 2020-01-28
Try: Moving Customers to main Woo Menu

timmyc:
Background: p6q8Tx-1qj-p2

This try branch moves the Customers report from the Analytics menu, to a sub menu item under WooCommerce.

### Screenshots

customers

- Verify Customers appears under the WooCommerce menu, and clicking on it loads the report and highlights the sidebar menu item
- Verify other analytic reports ( i.e. Revenue ) still work as expected.

### Changelog Note:

Enhancement: Move Customers report to WooCommerce Menu

### DIFF
--- a/src/Features/Analytics.php
+++ b/src/Features/Analytics.php
@@ -134,8 +134,8 @@ class Analytics {
 			array(
 				'id'     => 'woocommerce-analytics-customers',
 				'title'  => __( 'Customers', 'woocommerce-admin' ),
-				'parent' => 'woocommerce-analytics',
-				'path'   => '/analytics/customers',
+				'parent' => 'woocommerce',
+				'path'   => '/customers',
 			),
 			array(
 				'id'     => 'woocommerce-analytics-settings',
--- a/client/analytics/report/index.js
+++ b/client/analytics/report/index.js
@@ -118,13 +118,15 @@ class Report extends Component {
 			return null;
 		}
 
-		const { params, isError } = this.props;
+		const { params, path, isError } = this.props;
 
 		if ( isError ) {
 			return <ReportError isError />;
 		}
 
-		const report = find( getReports(), { report: params.report } );
+		const reportParam = params.report || path.replace( /^\/+/, '' );
+
+		const report = find( getReports(), { report: reportParam } );
 		if ( ! report ) {
 			return null;
 		}
--- a/client/layout/controller.js
+++ b/client/layout/controller.js
@@ -67,6 +67,12 @@ export const getPages = () => {
 		} );
 		pages.push( {
 			container: AnalyticsReport,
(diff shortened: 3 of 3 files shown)


src/Features/Analytics.php
```diff
@@ -134,8 +134,8 @@ public function register_pages() {
 			array(
 				'id'     => 'woocommerce-analytics-customers',
 				'title'  => __( 'Customers', 'woocommerce-admin' ),
-				'parent' => 'woocommerce-analytics',
-				'path'   => '/analytics/customers',
+				'parent' => 'woocommerce',
+				'path'   => '/customers',
```
jeffstieler: Should this be broken out from the analytics "feature" now?

pmcpinto:
@timmyc can you confirm that the tracking won't be affected? Thanks

src/Features/Analytics.php
timmyc: Perhaps we could break it into its own feature... because it seems like there may be some child features of it too coming up.

timmyc:
@pmcpinto this change shouldn't impact any of the current tracks implementations.

rrennick:
@timmyc Any further changes planned for this one from our discussion last week?

garymurray:
> @timmyc Any further changes planned for this one from our discussion last week?

Is there still a chance this will get into the WC 4.0 release? Not sure what is being referenced above.

rrennick:
> Not sure what is being referenced above.

We have 1M+ active installs where users would be accustomed to finding the menu item under Analytics. We discussed whether it might be better to duplicate the menu item rather than move it.

garymurray:
> We have 1M+ active installs where users would be accustomed to finding the menu item under Analytics. We discussed whether it might be better to duplicate the menu item rather than move it.

Duplication could work - as long as we have a plan to then probably remove it from the old location in the next release perhaps.

I think a notification in the Inbox would also be a good way to inform customer it is moving and will be permanently moved in the next release.

pmcpinto:
Just for reference, the Customers report compromises 2% of the wcadmin pageviews. The main dashboard takes 71% of the pv, followed by the revenue report with 12%.

We have about ~500pv/day in the Customers report, so maybe it's not worth the effort of duplicating it. I suggest simplifying and just assume the new placement. Also, if we duplicate it that would require to also change the tracking, so we can distinguish them. Thoughts?

garymurray:
With 2% of pageviews - I would suggest moving it and just informing our merchants where it moved to.

rrennick:
@garymurray Thanks for the stat.

@timmyc This works great except for the customer search which doesn't return anything when you click on a suggested customer.

rrennick:
> This works great except for the customer search

Looks like that is #3682

PR #3645, opened 2020-01-30
Onboarding: Fix toggle tracking events

### LINKED ISSUE #3643: Tracking onboarding enabled sites
**Describe the bug**
There are a few issues with how we track sites that have new the new onboarding experience enabled:
* The event `wcadmin_onboarding_toggled` is prefixed with `wcadmin` so it becomes `wcadmin_wcadmin_onboarding_toggled`
* Currently the event is only recorded when the `Onboarding::enable_onboarding()` method is hit.  We should change this to watch for updates to the `woocommerce_onboarding_opt_in` option.
* Some sites appear to be tracking various onboarding events, such as `storeprofiler_store_theme_choose` but never hit the `wcadmin_wcadmin_onboarding_toggled` event.  This is probably related to https://github.com/woocommerce/woocommerce/issues/25526 but a quick sanity check to make sure I'm not missing something would be helpful /cc @rrennick for your thoughts

**To Reproduce**

1. Enable the new onboarding experience via the "Help" tab.
2. Note the event name.
3. Check out stats (using looker or hue) for various onboarding events in tracks.
4. Notice that grouped by `userid` there are many users with onboarding events, yet they never hit the `wcadmin_wcadmin_onboarding_toggled` event.

**Expected behavior**
Expected any site that has onboarding enabled to be tracked.

**Screenshots**
Screen Shot 2020-01-30 at 2 44 35 PM

joshuatf:
Fixes #3643

Fixes a double `wcadmin_` prefix on the event and tracks on any update of the option in case this option is updated from some other code directly (e.g., `update_option( 'woocommerce_onboarding_opt_in', 'yes' );` is used directly in core).

1. Try toggling onboarding on/off via the "Help" tab.
1. Make sure the `wcadmin_onboarding_toggle` event is fired with the correct value.
1. Try adding `update_option( 'woocommerce_onboarding_opt_in', 'yes' );` or `update_option( 'woocommerce_onboarding_opt_in', 'no' );` and making sure the event is fired. (Note this will only work when the option is updated and does not stay the same value.

### DIFF
--- a/src/Features/Onboarding.php
+++ b/src/Features/Onboarding.php
@@ -62,11 +62,6 @@ class Onboarding {
 	public function __construct() {
 		$this->add_toggle_actions();
 
-		if ( ! Loader::is_onboarding_enabled() ) {
-			add_action( 'current_screen', array( $this, 'update_help_tab' ), 60 );
-			return;
-		}
-
 		// Include WC Admin Onboarding classes.
 		if ( self::should_show_tasks() ) {
 			OnboardingTasks::get_instance();
@@ -89,6 +84,11 @@ class Onboarding {
 		// Track the onboarding toggle event earlier so they are captured before redirecting.
 		add_action( 'add_option_' . self::OPT_IN_OPTION, array( $this, 'track_onboarding_toggle' ), 1, 2 );
 		add_action( 'update_option_' . self::OPT_IN_OPTION, array( $this, 'track_onboarding_toggle' ), 1, 2 );
+
+		if ( ! Loader::is_onboarding_enabled() ) {
+			add_action( 'current_screen', array( $this, 'update_help_tab' ), 60 );
+			return;
+		}
 	}
 
 	/**
--- a/src/Features/Onboarding.php
+++ b/src/Features/Onboarding.php
@@ -74,8 +74,6 @@ class Onboarding {
 
 		// Rest API hooks need to run before is_admin() checks.
 		add_filter( 'woocommerce_rest_prepare_themes', array( $this, 'add_uploaded_theme_data' ) );
-		add_action( 'woocommerce_theme_installed', array( $this, 'delete_themes_transient' ) );
-		add_action( 'after_switch_theme', array( $this, 'delete_themes_transient' ) );
 
 		// Add onboarding notes.
 		new WC_Admin_Notes_Onboarding_Profiler();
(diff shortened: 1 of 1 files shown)


rrennick:
@joshuatf I tested using the help toggle and the tracks events weren't being queued

Screen Shot 2020-01-30 at 11 39 43 AM

To double check I added an `error_log()` call to `track_onboarding_toggle()` and nothing was logged. The option must be updated before the hooks are added.

rrennick:
@joshuatf To provide an easy path for addressing #3648, can we add an `Onboarding::init()` hooked into a hook before `admin_init` and move the filter hooks to that function?

joshuatf:
@rrennick It looks like #3648 is already taken care of.  Are you thinking that moving to `Onboarding::init()` would be helpful to consolidate initialization hooks into a single place or are you thinking more of something like `Onboarding::enable()`?

rrennick:
> would be helpful to consolidate initialization hooks into a single place

It's good to have the hooks in one function and filters in another rather than the constructor.

This now has conflicts which may be a result of me changing the base branch.

joshuatf:
Hey @rrennick - updated this per your feedback and this is ready for another look!

> tested using the help toggle and the tracks events weren't being queued

I think the screenshot you linked is only for JS based tracks and can't be seen there.  I'm seeing them fired every time onboarding is disabled/enabled.  Could you double check?

>It's good to have the hooks in one function and filters in another rather than the constructor.

I separated these into several methods.  There's obviously a bit of overlap in each given shared logic, but let me know if you think this looks better.

joshuatf:
Thanks for the review, @rrennick!

>Not related to this PR. The Disable still shows after onboarding has been disabled

This doesn't happen for me.  I wonder if you have another flag that might be enabling onboarding?

* `WP_DEBUG` and `WP_SCRIPT_DEBUG` enabled.
* The old onboarding option `wc_onboarding_opt_in` set to `yes`
* The constant `WOOCOMMERCE_ADMIN_ONBOARDING_ENABLED` set to `true`

rrennick:
> WP_DEBUG and WP_SCRIPT_DEBUG enabled.

Yeah, sorry, I forgot about those. Are we getting in the range where we want to remove that logic or maybe change it to a filter? There will be a percentage of sites running with WP_DEBUG enabled.

PR #3655, opened 2020-01-31
Add outline styling to focused textarea element

vbelolapotkov:
Textarea and input elements look different in the form because outline styling is not
applied to text area element.

### Screenshots
**Before** 
wc_admin_outline_before

**After**
wc_admin_outline_after

- Create a form with input and text area elements.
- Focus each element and compare the outline.

### DIFF
--- a/client/stylesheets/shared/_global.scss
+++ b/client/stylesheets/shared/_global.scss
@@ -99,6 +99,7 @@
 	}
 
 	input:focus,
+	textarea:focus,
 	select:focus,
 	button:focus {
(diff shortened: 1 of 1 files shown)


dechov:
Good change 👍. On a separate note, the default box-shadow blur, spread, and transparency got updated in WP 5.3, after the original styling here was done (as shown in the first screenshot there). Should we update to `box-shadow: 0 0 0 1px $studio-woocommerce-purple-60;`? (cc @LevinMedia)

PR #3679, opened 2020-02-07
Add Storybook for documenting and testing components in isolation

haszari:
Adds basic storybook config (based on WooCommerce Blocks => Gutenberg config) & one example story for `Rating`.

### Screenshots

Screen Shot 2020-02-12 at 12 24 02 PM

1. Clone this branch, `npm install` (a few new things).
2. `npm run storybook` to run storybook dev server.
3. Browse to `http://localhost:6007` and browse component stories. Stories are alongside each component in a `stories` folder (similar to Gutenberg project).
3. Fiddle with the knobs for `Rating` component.
4. Write a story for a component :)

### Changelog Note:

> Dev: Added Storybook for documenting internal components.


packages/components/src/rating/stories/index.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Internal dependencies
+ */
+import Rating from '../';
+import '../style.scss';
```
joshuatf: This broke my build, but removing this line got it up and running.

Screen Shot 2020-02-09 at 5 30 45 PM

packages/components/src/rating/stories/index.js
```diff
@@ -0,0 +1,12 @@
...
+
+export default {
+	title: 'WooCommerce Admin/components/Rating',
```
joshuatf: Don't necessarily need a change here, just curious why the top level "WooCommerce Admin" is needed or if this is just a common structure we're using across all repos.

packages/components/src/rating/stories/index.js
```diff
@@ -0,0 +1,12 @@
...
+	component: Rating,
+};
+
+export const Default = () => <Rating rating={ 4.6 } totalStars={ 5 } />;
```
joshuatf: Would be awesome to define some knobs in here to control `rating`, `totalStars`, and `size` to demo some Storybook functionality for upcoming components.

packages/components/src/rating/stories/index.js
haszari: We'll need to resolve this somehow. Without the styles pulled in, the component doesn't render properly:

Screen Shot 2020-02-12 at 10 55 38 AM

Is there a top-level styles file (for all components) that we should pull into the storybook preview? Any other ideas how we can solve this?

packages/components/src/rating/stories/index.js
haszari: I'm following the pattern we're using in blocks - I think it's good to represent the "namespace" where components come from, and any hierarchy. If we combine our storybooks in future this would be even more useful :)

packages/components/src/rating/stories/index.js
haszari: Good point, I'll try that (a good excuse to play with knobs) :)

haszari: Thanks for the nudge - added knobs for all three props 🎛

packages/components/src/rating/stories/index.js
haszari: Aha! There is a mega style file for all components, now fixed here: ef17c27a42feb4f1045de37df7c740833576cc91

haszari:
Note: I've rebased this PR to fix the conflicts (😓 package-lock!!). Ready for re-review, would be good to merge this before it gets stale again cc @joshuatf @nerrad

nerrad:
I'm happy to review if the wc-admin squad wants but since they will largely be assuming maintenance and adding new stories for the storybook framework I think it'd probably be best if wc-admin reviews and makes the call on merging.  FWIW I do think it'd be very valuable to centralize around using storybook for exposing and developing components so I think folks know where I stand on whether it should merge or not :)

haszari:
Rebased again (package-lock conflicts 🌊). Would be great to get this merged - @joshuatf can you help spread the word? @timmyc @psealock do we need more reviews to merge this?

PR #3687, opened 2020-02-09
Add deactivation note for feature plugin

joshuatf:
Fixes #

Adds a note to deactivate older versions of wc admin when the included core version is greater than the feature plugin running.

Notes:
* We can't use a standard URL with a nonce since the notes action is stored in the DB and the nonce would expire.
* We probably need to add some permission checks on notes at some point to prevent notes from being visible to those without permissions.  Currently this will show a permission error after clicking the action if the user does not have permission to disable plugins.

### Screenshots
Screen Shot 2020-02-09 at 1 12 37 PM

1. Symlink or copy this branch to the WC 4.0 beta `woocommerce/packages/woocommerce-admin` directory.
1. Run an older version as a feature plugin (or force this by changing the version constant in `Package.php`).
1. Visit any WC page with store alerts.
1. Note the error.
1. Disable the plugin with the action.
1. Make sure the feature plugin is successfully disabled.

### DIFF
--- a/src/Notes/WC_Admin_Notes_Update_Version.php
+++ b/src/Notes/WC_Admin_Notes_Update_Version.php
@@ -18,6 +18,13 @@ class WC_Admin_Notes_Update_Version {
 	const NOTE_NAME = 'wc-admin-update-version';
 
 	/**
+	 * Attach hooks.
+	 */
+	public function __construct() {
+		add_action( 'init', array( $this, 'deactivate_feature_plugin' ) );
+	}
+
+	/**
 	 * Creates the note to deactivate the older version.
 	 */
 	public static function add_note() {
@@ -38,7 +45,7 @@ class WC_Admin_Notes_Update_Version {
 		$note->add_action(
 			'deactivate-feature-plugin',
 			__( 'Deactivate', 'woocommerce-admin' ),
-			wc_admin_url(),
+			wc_admin_url( '&action=deactivate-feature-plugin' ),
 			'unactioned',
 			true
 		);
@@ -50,5 +57,26 @@ class WC_Admin_Notes_Update_Version {
 	 */
 	public static function delete_note() {
 		WC_Admin_Notes::delete_notes_with_name( self::NOTE_NAME );
+	}
+
+	/**
+	 * Deactivate feature plugin.
+	 */
+	public function deactivate_feature_plugin() {
+		/* phpcs:disable WordPress.Security.NonceVerification */
+		if (
+			! isset( $_GET['page'] ) ||
+			'wc-admin' !== $_GET['page'] ||
+			! isset( $_GET['action'] ) ||
+			'deactivate-feature-plugin' !== $_GET['action'] ||
+			! defined( 'WC_ADMIN_PLUGIN_FILE' )
+		) {
+			return;
+		}
+		/* phpcs:enable */
+
+		$deactivate_url = admin_url( 'plugins.php?action=deactivate&plugin=' . rawurlencode( WC_ADMIN_PLUGIN_FILE ) . '&plugin_status=all&paged=1&_wpnonce=' . wp_create_nonce( 'deactivate-plugin_' . WC_ADMIN_PLUGIN_FILE ) );
+		wp_safe_redirect( $deactivate_url );
+		exit;
 	}
 }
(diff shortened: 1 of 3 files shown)


rrennick:
@joshuatf The logic on this works perfectly except that with the autoloaders the `Package.php` autoloaded by WC is the plugin file `plugins/woocommerce-admin/src/Package.php` vs the WC `/packages/...` file. We'll have to look at excluding `Package.php` from the local autoloading.

jeffstieler:
@rrennick is that change required to approve this PR?

rrennick:
> is that change required to approve this PR?

Yes, the notice will never be displayed on any version since `Package.php` was added because it's comparing the version constants in the standalone copy of the plugin.

rrennick:
I have updated the PR by adding `\Compose` to the namespace on the `Package` class. Adding the folder to the namespace prevents the WC Admin autoloader from finding the class file. The WC core autoloader isn't using PSR-4 for autoloading packages, so its autoloader finds and loads the `Package` class in the `packages` folder.

This requires a PR in WC Core which I'll create in a few minutes.

rrennick:
@jeffstieler This one is ready for review.

jeffstieler:
> I have updated the PR by adding `\Compose` to the namespace on the `Package` class. Adding the folder to the namespace prevents the WC Admin autoloader from finding the class file. The WC core autoloader isn't using PSR-4 for autoloading packages, so its autoloader finds and loads the `Package` class in the `packages` folder.

Just to make sure I understand this (and that's usually where I ask for a lengthy comment block explaining the code): the addition of `\Compose` to the namespace (or any additional path) causes the `Package` class to not get autoloaded when any other class in the `\Automattic\WooCommerce\Admin` namespace is referenced?

If I'm understanding that correctly - why "Compose" as the additional path? Was that supposed to be `Composer`?

rrennick:
> to the namespace (or any additional path) causes the Package class to not get autoloaded when any other class in the \Automattic\WooCommerce\Admin namespace is referenced

More or less. The autoloading function loads a class file when that class is referenced. When WC Admin is active as a standalone plugin, the Package class file isn't loaded because the SA plugin doesn't reference it.

> why "Compose" as the additional path? Was that supposed to be Composer?

I was thinking of differentiating between composing the standalone plugin and the composer package. I'd be okay with `Composer` or `Composed`. If the class name were something different then I would have added `Package` to the namespace. `Composed` might signal that it's a "composed package".

src/Package.php
```diff
@@ -39,6 +41,13 @@ public static function init() {
 
 		// Avoid double initialization when the feature plugin is in use.
 		if ( defined( 'WC_ADMIN_VERSION_NUMBER' ) ) {
+			$update_version = new WC_Admin_Notes_Update_Version();
+			if ( version_compare( WC_ADMIN_VERSION_NUMBER, self::VERSION, '<' ) ) {
+				$update_version::add_note();
+			} else {
+				$update_version::delete_note();
```
psealock: Why not just have the `if` statement and not create the note at all in the `else` case.

rrennick: The user could update WCA to the current version. Without the delete_note, the notice would continue to display.

src/Notes/WC_Admin_Notes_Update_Version.php
```diff
@@ -0,0 +1,82 @@
...
+
+namespace Automattic\WooCommerce\Admin\Notes;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * WC_Admin_Notes_Update_Version.
+ */
+class WC_Admin_Notes_Update_Version {
+	const NOTE_NAME = 'wc-admin-update-version';
```
psealock: Should this be `wc-deactivate-outdated-plugin` to be more reflective of the call to action?

rrennick: Went with `wc-admin-deactivate-plugin`.

psealock:
https://github.com/woocommerce/woocommerce-admin/pull/3687/commits/f29854925217946e4c37fa5713cf78be0d53fc35 makes the renamed note name the class and filename for consistency.

src/Package.php
```diff
@@ -39,6 +46,13 @@ public static function init() {
 
 		// Avoid double initialization when the feature plugin is in use.
 		if ( defined( 'WC_ADMIN_VERSION_NUMBER' ) ) {
+			$update_version = new WC_Admin_Notes_Deactivate_Plugin();
+			if ( version_compare( WC_ADMIN_VERSION_NUMBER, self::VERSION, '<' ) ) {
+				$update_version::add_note();
+			} else {
+				$update_version::delete_note();
```
peterfabian: If the `WC_ADMIN_VERSION_NUMBER` is not defined, that means the plugin is either inactive or not installed, so the notice should not be displayed. 
However, in my testing with WC Admin 0.23.3, the code never hits `$update_version::delete_note();`.

PR #3707, opened 2020-02-12
Settings: get and update mutable settings via store

psealock:
Depends on https://github.com/woocommerce/woocommerce-admin/pull/3675
Related to #3661

Update `tax` and `general` settings usages by the Onboarding Wizard to use `@woocommerce/data` instead of `wc-api/settings`.

This PR also 
* Removes `wc-api/settings` because it is no longer used.
* Introduce resolvers are so that settings can be gathered via API request or with initial hydration via `withSettingsHydration`. 
* Change `isPersisting` to `isRequesting` because GET requests are also being made.

### Test

Go through the entire OBW and Task List and make sure data is gathered, displayed, and persisted correctly.


client/dashboard/task-list/tasks/tax.js
```diff
@@ -148,10 +149,10 @@ class Tax extends Component {
 	}
 
 	async updateAutomatedTax() {
-		const { createNotice, isTaxSettingsError, updateSettings } = this.props;
+		const { createNotice, isTaxSettingsError, updateAndPersistSettingsForGroup } = this.props;
 		const { automatedTaxEnabled } = this.state;
 
-		await updateSettings( {
+		await updateAndPersistSettingsForGroup( 'tax', {
```
psealock: @nerrad, here I have a call to update and persist a settings group, which works fine in other instances. The difference in this one is that the `tax` settings group is instantiated via an API call instead of hydration via `wcSettings`, which is all working as it should.

This persistence call creates an error on WP 5.3.2, but not on the latest version of Gutenberg.

Screen Shot 2020-02-12 at 5 02 54 PM

I traced the fix back to this pull request: https://github.com/WordPress/gutenberg/pull/17558

As someone who reviewed that PR, would you happen to have any insight as to why this error is happening in the previous to changes in that PR? I'm curious because there are backward compat issues, unless this is pointing to something I'm doing wrong.

nerrad: I'm not sure how the linke pull relates to this issue here because it is dealing with a _selector_ not an action which is what is causing the error for you (if indeed line 155 here is the cause). So that might be a red herring.  However, the one thing it _might_ be related to is that `updateAndPersistSettingsForGroup` internally results in the select control being called which _is_ modified by the linked pull you pointed to.  Essentially it _looks_ from the error like a requested selector/action does not exist in the store at the time redux/routine is trying to call it. You could review all selector, action and store names to make sure that they don't have any typos?

Another potential issue here is that ultimately you cannot rely on the response from a dispatched action (which you are doing with await) - across various versions of WordPress core and Gutenberg, wp.data dispatched actions return different final values and I'm pretty sure the GB team is moving towards always returning undefined. So to rely on the return value is fragile.

Instead the action should be dispatched directly and then your component reacts however it needs once the state is updated by the dispatched action. So that may mean what state gets updated by the dispatched action, it may mean you need to refactor the component to account for that, it may mean introducing a new action/selector and using that. I'm not sure exactly what might be needed in this case without doing a deep dive (which I'm willing to do if needed).

Another potential problem here is that I see `setSetting` is being called for the `onboarding` object.  Doesn't that mutate the global? I wonder if that's causing something persisted in the data store to get mutated which results in this intermediate value error (unlikely, but probably still should be handled...I'm guessing this is just something not refactored yet)?

psealock: Thanks for taking a look @nerrad .

>However, the one thing it might be related to is that `updateAndPersistSettingsForGroup` internally results in the select control being called which is modified by the linked pull you pointed to.

Right, so that action calls `persistSettingsForGroup` which has two selects (in pseudo code):

```js
import { select } from '@wordpress/data-controls';

const dirtyKeys = yield select( STORE_NAME, 'getDirtyKeys', group );
const dirtyData = yield select( STORE_NAME, 'getSettingsForGroup', group, dirtyKeys );
```
The first works fine, but the second does not. Here is part of the faulty line:

```js
registry[registry.select(storeKey)[selectorName].hasResolver ? '__experimentalResolveSelect' : 'select']
```
The second selector `getSettingsForGroup` has a resolver, so the ternary expression results in `__experimentalResolveSelect`.

Stepping through the code, `registry.__experimentalResolveSelect` does not exist in the old version of Gutenberg, but does after the merge of https://github.com/WordPress/gutenberg/pull/17558. 

Here is the problematic function (from a commit before the above PR was merged):
https://github.com/WordPress/gutenberg/blob/4857ad58c1241b3d63d21a6880c989b85746c3dc/packages/data-controls/src/index.js#L127-L135

Why does the selector need to use `__experimentalResolveSelect`? That selector has already resolved and its results already in state. I can't gather from the PR why it is defined in newer versions of Gutenberg, but not in the one shipped with 5.3.2.

>Another potential issue here is that ultimately you cannot rely on the response from a dispatched action (which you are doing with await) - across various versions of WordPress core and Gutenberg, wp.data dispatched actions return different final values and I'm pretty sure the GB team is moving towards always returning undefined. So to rely on the return value is fragile.

Good point. I think thats unrelated to these changes, but can be addressed in a follow up.

>Another potential problem here is that I see setSetting is being called for the onboarding object. Doesn't that mutate the global?

Yes, thats a problem. But my approach is to handle the mutable settings of `wcSettings` one at a time. This pull addresses `preloadSettings`. Up next will be `onboarding` and `currentUserData`. This way I can remove `wc-api` one folder at a time and thus have incremental pull requests from master after this series of PRs is merged.

psealock: If I copy the problematic selector function `getSettingsForGroup` and rename it `getAlreaydResolvedSettingsForGroup` such that it does not have a matching resolver by the same name, everything works as expected.

Although this could be a work-around, it really seems dirty and an indication that something is wrong.

nerrad:
> The second selector getSettingsForGroup has a resolver, so the ternary expression results in __experimentalResolveSelect.

> Stepping through the code, registry.__experimentalResolveSelect does not exist in the old version of Gutenberg, but does after the merge of WordPress/gutenberg#17558.

Ahh okay, I think I know what the issue is now. We're importing `@wordpress/data-controls` directly from the latest package instead of using what is bundled and exposed with WordPress but using `@wordpress/data` from what is bundled with WordPress.  So you need to define `@wordpress/data-controls` as an external in the wc-admin webpack config.  I missed this because I'm used to relying on `@wordpress/dependency-extract-webpack-plugin` for managing all WordPress dependencies and externals which wc-admin isn't currently doing (_strongly_ recommend you guys switch over to that at some time).

psealock:
Absolutely! Good catch @nerrad . And yes, having used `@wordpress/dependency-extract-webpack-plugin`, we should save ourselves a few headaches by adding it to the repo.

So `wp.dataControls` does not exist in the latest WP (5.3.2), so we have to import it from node_modules instead of the wp global. But this got me thinking. Perhaps using the `select` found on `wp.data` would work? 

Sure enough, if I make the following changes in my actions file (packages/data/src/settings/actions.js), then values are defined in the function. I removed usage of select from `@wordpress/data-controls` and used `select` from `@wordpress/data` instead.

```diff
-import { apiFetch, select } from '@wordpress/data-controls';
+import { apiFetch } from '@wordpress/data-controls';
+import { select } from '@wordpress/data';
 import { concat } from 'lodash';
 
 /**
@@ -59,7 +60,7 @@ export function* persistSettingsForGroup( group ) {
        // first dispatch the is persisting action
        yield setIsPersisting( group, true );
        // get all dirty keys with select control
-       const dirtyKeys = yield select( STORE_NAME, 'getDirtyKeys', group );
+       const dirtyKeys = yield select( STORE_NAME ).getDirtyKeys( group );
        // if there is nothing dirty, bail
        if ( dirtyKeys.length === 0 ) {
                yield setIsPersisting( group, false );
@@ -67,7 +68,7 @@ export function* persistSettingsForGroup( group ) {
        }
 
        // get data slice for keys
-       const dirtyData = yield select( STORE_NAME, 'getSettingsForGroup', group, dirtyKeys );
+       const dirtyData = yield select( STORE_NAME ).getSettingsForGroup( group, dirtyKeys );
        const url = `${ NAMESPACE }/settings/${ group }/batch`;
 
        const update = dirtyKeys.reduce( ( updates, key ) => {
```
I'll do some more extensive testing, but it seems legit so far. Any reason why this might cause a problem?

nerrad:
> So wp.dataControls does not exist in the latest WP (5.3.2), so we have to import it from node_modules instead of the wp global. But this got me thinking. Perhaps using the select found on wp.data would work?

My apologies if I made it seem it doesn't exist in 5.3.2 but it actually does:

Image 2020-02-14 at 8 53 43 AM

The difference is that what is included with WP 5.3.2 is a version of select that still resolves automatically but not with the new experimental method. So it'll work.  The beauty with relying on the external is your code will then work no matter what version of WP is active. So I _strongly_ recommend you implement what I suggested in my response (by simply modifying your webpack config).

psealock:
>it doesn't exist in 5.3.2 but it actually does:

Screen Shot 2020-02-17 at 7 57 38 AM

Ok, thanks @nerrad. Now I'm curious why my 5.3.2 install doesn't have it

rrennick:
Awesome work on this @psealock. The only thing I found in testing on a new JN site was when I set up automatic taxes, taxes weren't enabled in the WC settings.

PR #3715, opened 2020-02-13
Make analytics tables use the site's date format setting

### LINKED ISSUE #3666: Date format does not appear to be localized
**Describe the bug**
When viewing tables in WC Admin dates seem to always be formatted as `MM/DD/YYYY`. 

**To Reproduce**

1. Go to change store address to an Australian address in WooCommerce > Settings > General
2. Go to Analytics > Orders
3. View order **Date** column
4. Dates will be still in US format.

**Expected behavior**
Date format should be localized for the store's region as in WooCommerce's existing admin pages. 

**Screenshots**
Screenshot
Screenshot: https://d.pr/i/UG8qtX

becdetat:
Fixes #3666

Note that the bug report only talks about the Orders page, however it is also present on the Downloads, Customers and Revenues pages.

I've pulled the site's date format setting in to `window.wcSettings` so that it is available to the analytics tables.

Eg. (using `F, j Y` date format):

image

Note that this is localised:

image

Base functionality:

- Go to `/wp-admin/options-general.php` and set the date format to something not `m/d/Y`
- Go to `/wp-admin/admin.php?page=wc-admin&path=%2Fanalytics%2Forders` and see the date format in use

Localization:

- Go to `/wp-admin/options-general.php`, set the language to something different, and set the date format to something that includes the month name (eg. `j F Y`)
- Go to `/wp-admin/admin.php?page=wc-admin&path=%2Fanalytics%2Forders` and see the localized month name

### Changelog Note:

Fix: Make analytics tables use the site's date format setting


client/analytics/report/customers/table.js
```diff
@@ -109,6 +108,8 @@ export default class CustomersReportTable extends Component {
 	}
 
 	getRowsContent( customers ) {
+		const { dateFormat } = window.wcSettings;
```
psealock: You'll want to use `getSettings` as you mentioned in a previous conversation.
```js
import { getSettings } from '@woocommerce/wc-admin-settings';
```
`wcSettings` has been an abused dumping ground for settings, some of which are mutable. We're in the midst of a refactor to right some wrongs, but introduced `getSettings` as a way to access `wcSettings` in a more uniform manner. `dateFormat` is unlikely to be changed by wc-admin, so while this works, it'll be best to stick to convention.

src/Loader.php
```diff
@@ -690,6 +690,7 @@ public static function add_component_settings( $settings ) {
 		$settings['wcVersion']         = WC_VERSION;
 		$settings['siteUrl']           = site_url();
 		$settings['onboardingEnabled'] = self::is_onboarding_enabled();
+		$settings['dateFormat']        = get_option( 'date_format' );
```
psealock: 👍

psealock:
Originally, I mentioned that we may want to replace the old approach with this one in other places in the app. After looking into it, I think thats not the case because I now remember the purpose. The translated moment format strings are useful because the WP site setting is just one type, but we need a short and long:

https://github.com/woocommerce/woocommerce-admin/blob/cc615a7f1f214febaa944bbea893928f2df1a3fd/packages/date/src/index.js#L105-L106

This is for rendering things like this:

Screen Shot 2020-02-13 at 4 45 13 PM

This just an FYI and I wanted to close the loop on this in case I accidentally sent you working further on this. Cheers

client/analytics/report/customers/table.js
becdetat: Ok, I'm much more comfortable with using `getSetting` instead of using a global directly. Thanks for this. It's changed now.

client/analytics/report/customers/table.js
```diff
@@ -109,6 +108,8 @@ export default class CustomersReportTable extends Component {
 	}
 
 	getRowsContent( customers ) {
+		const dateFormat = getSetting( 'dateFormat' );
```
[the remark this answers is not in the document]
becdetat: Done

client/analytics/report/downloads/table.js
```diff
@@ -10,11 +10,11 @@ import moment from 'moment';
 /**
```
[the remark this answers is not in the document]
becdetat: Done

becdetat:
Hi @joshuatf 

>It looks like there’s still a couple date formats we can update in the coupons table in coupons/table.js.

Good catch, that's done now.

> There’s also some date formatting in places like the advanced filters and the historical data import, [...] if it's not that simple I think a follow-up may be a better option

It looks like the range formatting is done [here](https://github.com/woocommerce/woocommerce-admin/blob/cc615a7f1f214febaa944bbea893928f2df1a3fd/packages/date/src/index.js#L101) which uses i18n to adjust the format. I don't think it needs to be further customised and I don't think it could be without adding more core settings which does seem out of scope of this PR.

PR #3717, opened 2020-02-13
Create reusable E2E testing environment package.

jeffstieler:
This PR seeks to introduce a new package, `@woocommerce/e2e-env` which can be used to bootstrap an End-to-End testing environment for WooCommerce and extensions.

- Review instructions in `packages/e2e-env/README.md`
- Follow usage instructions in the `woocommerce-admin` repo
- Test in the `woocommerce-gutenberg-products-block` repo:
- Check out https://github.com/woocommerce/woocommerce-gutenberg-products-block/compare/try/e2e-package?expand=1
- Copy all files from `packages/e2e-env` to `node_modules/@woocommerce/e2e-env`


jeffstieler:
cc: @nerrad @juliaamosova 😄

packages/e2e-env/config/jest.setup.js
```diff
@@ -0,0 +1,271 @@
...
+ */
+const OBSERVED_CONSOLE_MESSAGE_TYPES = {
+	warning: 'warn',
+	error: 'error',
+};
+
+async function setupBrowser() {
+	await clearLocalStorage();
+	await setBrowserViewport( 'large' );
+}
```
nerrad: This file seems to be something that would be better suited as a part of a `@woocommerce/e2e-utils` package that can then be used for exposing all woocommerce test utilities (including what is currently in woocommerce core). 

I understand the value in plugins just having one package to pull in for setting up an e2e testing environment and all the related utilities, but there's also common use-case where users will want to run e2e tests against their own environment (which they can). Docker is just a convenience. So I wonder if we should follow the pattern established by GB (and elsewhere I've seen) where we have a `@woocommerce/e2e-test-utils` (which extends `@wordpress/e2e-test-utils`) and split out the `woocommerce/env` stuff to its own package (which means it could feasibly evolve for other things including phpunit tests, local dev environment etc).

packages/e2e-env/docker-compose.yaml
```diff
@@ -0,0 +1,58 @@
+version: '3.3'
+
+services:
+
+  db:
+    container_name: "${APP_NAME}_db"
```
nerrad: good call on `${APP_NAME}` here to reduce container conflict on hosts.

packages/e2e-env/docker-compose.yaml
```diff
@@ -0,0 +1,58 @@
...
+        context: ./docker/wp-cli
+    restart: on-failure
+    environment:
+        WORDPRESS_PORT: 8084
+        WORDPRESS_HOST: wordpress-www:80
+        WORDPRESS_TITLE: "WooCommerce Core E2E Test Suite"
+        WORDPRESS_LOGIN: admin
+        WORDPRESS_PASSWORD: password
+        WORDPRESS_EMAIL: "admin@woocommercecoree2etestsuite.com"
+        DOMAIN_NAME:
```
nerrad: I'm assuming these can be overridden by `.env`?

juliaamosova:
@jeffstieler really great work! I am always amazed at how fast you are able to produce something like a complex package such as the one you are working on it. 

I am AFK today so I can't take a look at this PR in detail at the moment. But I will do so on Monday/Tuesday next week. One thing I wanted to note now though is that we have a PR https://github.com/woocommerce/woocommerce/pull/25417 open in core to implement a general [puppeteer-utils](https://github.com/Automattic/puppeteer-utils) package that includes jest and puppeteer setup plus a collection of puppeteer specific helper functions. Let's make sure your package accounts for that. Unless you have any objections?

Also, would it make more sense for the package to live in the `woocommerce` organization as opposed to in `woocommerce-admin` repository? Not a big deal but I think that if it to be used by all woo product teams and perhaps 3rd party developers, it is better to have the package in a central place. Or perhaps in core?

nerrad:
@juliaamosova , related to this comment:

> Also, would it make more sense for the package to live in the woocommerce organization as opposed to in woocommerce-admin repository? Not a big deal but I think that if it to be used by all woo product teams and perhaps 3rd party developers, it is better to have the package in a central place. Or perhaps in core?

Is this post: p90Yrv-1yE-p2

There's argument made in the linked p2 that we should pick a repo to serve as the mono-repo for all `@woocommerce` packages. Typically, folks outside the woo org are going to discover `@woocommerce` packages from npm so it doesn't matter much where we publish from. But having all the packages in a mono-repo does make it easier to maintain and keep on top of those packages (imo). Still, I hold this opinion loosely, so I'm not opposed to what you suggest but I think it's a wider conversation that needs input from multiple folks before committing to it (hence the p2 😄 ).

jeffstieler:
> As mentioned at one point in the code feedback, I think it'd be worthwhile to split out the environment package to be separate from an `@woocommerce/e2e-test-utils` package. I think it's reasonable to expect that consumers might want the option to run e2e tests against a different host than docker (for example, eventually we might have a set of tests that run automatically on WooCommerce.com staging or as a part of an automated review process for the theme marketplace (both themes and the marketplace). In those cases it'd be valuable to just pull in the WooCommerce specific e2e utils if that's all that's needed.

Now that I'm aware of the `@wordpress/env` and `@wordpress/e2e-test-utils` packages, I agree with this approach. Alignment with core is A Good Idea ™.

> `@wordpress/env` is coming along really well and has multiple contributions happening to it (and getting merged). I think it'd be really cool if we contribute to that effort instead of building our own - we could contribute the permissions fix thing from here and also the mechanism you added for an initialization script with wp-cli. It would compliment nicely the current options in that library. I'd be happy to help with that if needed.

TIL. 👍

packages/e2e-env/docker-compose.yaml
jeffstieler: It looks like I've forgotten to hook up `.env` here - I'll hold off on making changes since it looks like we'll abandon this PR.

nerrad:
> and @wordpress/e2e-test-utils packages, I agree with this approach. Alignment with core is A Good Idea ™.

To be clear, I still think we'll need a `@woocommerce/e2e-test-utils` package that extends the `@wordpress/e2e-test-utils` package. Reason is we'll have specific WooCommerce test utils helpers that will never be included in the WordPress package because they are WC specific.

jeffstieler:
> To be clear, I still think we'll need a `@woocommerce/e2e-test-utils` package that extends the `@wordpress/e2e-test-utils` package. Reason is we'll have specific WooCommerce test utils helpers that will never be included in the WordPress package because they are WC specific.

Yep, sorry I wasn't explicit in calling that out. I'm sure we'll want/need a `@woocommerce/env` as well.

packages/e2e-env/config/default.json
```diff
@@ -0,0 +1,62 @@
+{
+  "url": "http://localhost:8084/",
```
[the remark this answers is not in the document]
juliaamosova: In WooCore, this file is being used by e2e tests directly as it contains the details needed for running the tests such as user details, billing and shipping addresses etc. It looks like it is intended for the same use here. IMO, this file should be out of this package.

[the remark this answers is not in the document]
nerrad: > IMO, this file should be out of this package.

Agreed (especially if that's provided through Woo core test helpers).

packages/e2e-env/config/jest.setup.js
juliaamosova: @nerrad I originally thought that it would be a good idea to combine both the Docker setup + e2e utils in one package but I agree with your argument regarding separating them.

juliaamosova:
> Is this post: p90Yrv-1yE-p2
> 
> There's argument made in the linked p2 that we should pick a repo to serve as the mono-repo for all @woocommerce packages. Typically, folks outside the woo org are going to discover @woocommerce packages from npm so it doesn't matter much where we publish from. But having all the packages in a mono-repo does make it easier to maintain and keep on top of those packages (imo). Still, I hold this opinion loosely, so I'm not opposed to what you suggest but I think it's a wider conversation that needs input from multiple folks before committing to it (hence the p2 😄 ).

@nerrad thanks! I commented on the P2 with reasons why I think the package should be a part of WooCore repository.

juliaamosova:
> To be clear, I still think we'll need a @woocommerce/e2e-test-utils package that extends the @wordpress/e2e-test-utils package. Reason is we'll have specific WooCommerce test utils helpers that will never be included in the WordPress package because they are WC specific.

+1 to the above ^. 

Also with https://github.com/woocommerce/woocommerce/pull/25417 now merged to core, we'd need to rethink what puppeteer + jest config is included in the additional Docker env package.

jeffstieler:
Closing in favor of https://github.com/woocommerce/woocommerce/pull/26109.

PR #3719, opened 2020-02-13
Remove sideloaded images, host on woocommerce.wordpress.com

### LINKED ISSUE #3716: [4.0 beta] Review the image assets bundled with WC Admin
Currently, WC Admin includes 5MBs of images in the images/onboarding directory, which contributes to the total package size of WC core considerably (total ±12MBs). We need to optimize and remove the unneeded ones.

timmyc:
Fixes #3716 

Current builds of woocommerce-admin include over 5mb of images that are ultimately sideloaded - downloaded from the plugin directory, and uploaded to the media library. These images drastically increase the size of the build and in-turn, inflate the size of WooCommerce core to over 12mb.

I attempted to run the images through a jpg optimizer, and they still weighed in at 2.7mb total - which to me feels still way too costly for what they are ultimately used for. Furthermore, any existing woo site will never use these images, since they are only leveraged during the Homepage creation step in the new site task list.

As such, the proposed solution here involves deletion of the files from the repository, and instead hosting them on `woocommerce.wordpress.com`. The sideload logic still functions the same, so when a homepage is being created, the needed images are downloaded from WPCOM, and inserted into the media library.

This does run the risk of those images potentially being deleted someday on accident, so not sure if this is the most optimal solution. I did read through the plugin repository guidelines and did not see any rules that this approach would conflict with.

/cc @peterfabian 

- Disable Jetpack, otherwise the new post flow will be directed to WordPress.com and that kind of makes testing this more difficult
- Go to Orders > Help > Setup Wizard and click the Enable button for the task list:

image

- Go back to the Woo dashboard, and click the Personalize Your Store step, and then the create homepage button:

image

- Verify the image(s) used on the example layout include the images related to the industry you selected on your site, and that the images have been sideloaded/hosted from your test site's media library.

### Changelog Note:

Remove sideloaded images and host externally to save on build size


timmyc:
> My thoughts are that we should branch from version/1.0 and periodically merge that branch in with master.

Apologies on that, I have changed the base branch and removed the cherry pick label. Old woo core habits die hard I suppose.

psealock:
>I have changed the base branch and removed the cherry pick label.

Nice 🎉 It looks like you branched off master though, so a few other commits got included. Sorry about this, but maybe your original game plan to cherry-pick is best. Code looks good though.

nerrad:
> I did read through the plugin repository guidelines and did not see any rules that this approach would conflict with.

While it's debatable, sideloading/loading images from CDN has been a point of contention in the past for images. It's covered specifically [this guideline](https://developer.wordpress.org/plugins/wordpress-org/detailed-plugin-guidelines/#7-plugins-may-not-track-users-without-their-consent) where it states:

> Offloading assets (including images and scripts) that are unrelated to a service.

Arguably these images _are_ part of the service (the service being WooCommerce here onboarding?), but I just bring this up as something we might want to verify with the plugins team.

timmyc:
Thanks @nerrad, I also spoke with @beaulebens about this today too. I'm going to email the review team to verify prior to merge. Otherwise I think our only option will be to remove the images and update the templates - which I don't think is the end of the world.

I just don't know if we can justify shipping these stock photos in every download of Woo when it only will be used by new stores ( possibly ).

timmyc:
Email sent:

> Greetings Team-I'm working on trimming down the size of the WooCommerce Admin plugin, that is going to be merged to our WooCommerce plugin soon. Currently we ship a number of images that are used to create homepage templates for new WooCommerce stores. The only problem with this is the images are large and inflate the size of the WooCommerce Admin .zip file, which will in-turn be a large size bump to WooCommerce itself.
As such I was exploring the notion of hosting these images externally on a WordPress.com site and the images are then sideloaded ( only when needed by a homepage template ) into the local WordPress/WooCommerce install.
The code itself can be seen in the following Pull Request:
https://github.com/woocommerce/woocommerce-admin/pull/3719
As @nerrad points out in a comment on the PR, this could be a violation of the repository guidelines so I wanted to vet that carefully before we merged this approach.
Thanks in advance for the help!
Timmy

timmyc:
Heard back from the Plugin Review team and they confirmed @nerrad's concerns over this PR conflicting with the guidelines. Specifically, loading external assets ( anything, css, js, images ) conflicts with GDPR and as such is not allowed. 

Thanks for raising the concern to make sure we do the right thing here everyone.

Closing in favor of #3731 3731

PR #3722, opened 2020-02-14
Only add onboarding settings on wc-admin pages when task list should …

### LINKED ISSUE #3663: [Performance] Use caching to check if someone has physical products
**Describe the bug**
Query used to fetch data in https://github.com/woocommerce/woocommerce-admin/blob/master/src/Features/OnboardingTasks.php#L174 is pretty heavy. For egs, in my setup of 10k products, I am getting these results:

```
# Query_time: 0.314502  Lock_time: 0.000193 Rows_sent: 1  Rows_examined: 123735
SET timestamp=1580822949;
SELECT   wc_posts.ID FROM wc_posts  LEFT JOIN wc_term_relationships ON (wc_posts.ID = wc_term_relationships.object_id) INNER JOIN wc_postmeta ON ( wc_posts.ID = wc_postmeta.post_id ) WHERE 1=1  AND (
  wc_term_relationships.term_taxonomy_id IN (2,3,4,5)
) AND (
  ( wc_postmeta.meta_key = '_virtual' AND wc_postmeta.meta_value = 'no' )
) AND wc_posts.post_type = 'product' AND ((wc_posts.post_status = 'publish' OR wc_posts.post_status = 'draft' OR wc_posts.post_status = 'pending' OR wc_posts.post_status = 'private')) GROUP BY wc_posts.ID ORDER BY wc_posts.post_date DESC LIMIT 0, 1;
```

As you can see, 300ms for every page refresh while on WooCommerce related admin screens is quite impactful. This query will be slower depending upon how many products are there.

The performance here is slow because to identify if a product is not virtual, this query creates a JOIN between posts and postmeta table.

Please cache the results of this query.

**To Reproduce**

1. Go to WooCommerce dashboard after activating wc-admin plugin.
2. See you query log.

* OS: [e.g. iOS] MacBook Pro
* Browser [e.g. chrome, safari] Firefox
* Version [e.g. 22] 72.02

timmyc:
Fixes #3663 

This branch seeks to remove the expensive query for physical products that is happening on any Woo page where the new navigation bar is displayed. 

Just a bit of background on the query first. It appears the check for physical products is used to determine if the Shipping step in the checklist [should be displayed or not](https://github.com/woocommerce/woocommerce-admin/blob/master/client/dashboard/task-list/tasks.js#L110-L124). The boolean of `hasPhysicalProducts` is not used anywhere else besides the task list, as such I based my proposed fix here on that.

Changes Proposed:
In lieu of caching the query, which could be problematic in the context of the onboarding checklist ( granted we could clear the cache after each product save ) I have instead opted to return early from the logic if the request is not a wc-admin page ( i.e. Dashboard or Report ) and more importantly, if the task list should not be shown.

Therefore this query should only happen when a store is in "setup mode" / the setup checklist is being shown, and will only happen on wc-admin pages.

- Open any non wc-admin woo page, like Orders, and verify that the query described in the original issue does not happen.
- Enable the setup checklist via the Help Menu in the orders listing page

enable-task-list

- Verify the checklist still operates as expected ( i.e. Shipping task shows if you have physical products setup )

image

### Changelog Note:

Performance: Remove slow physical products query from non setup checklist pages

### DIFF
--- a/src/Features/OnboardingTasks.php
+++ b/src/Features/OnboardingTasks.php
@@ -165,7 +165,13 @@ class OnboardingTasks {
 	 * @return array
 	 */
 	public function component_settings( $settings ) {
-		$products = wp_count_posts( 'product' );
+		// Bail early if not on a wc-admin powered page, or task list shouldn't be shown.
+		if (
+			! \Automattic\WooCommerce\Admin\Loader::is_admin_page() ||
+			! \Automattic\WooCommerce\Admin\Features\Onboarding::should_show_tasks()
+		) {
+			return $settings;
+		}
 
 		// @todo We may want to consider caching some of these and use to check against
 		// task completion along with cache busting for active tasks.


src/Features/OnboardingTasks.php
```diff
@@ -165,7 +165,13 @@ public function add_media_scripts() {
 	 * @return array
 	 */
 	public function component_settings( $settings ) {
-		$products = wp_count_posts( 'product' );
```
timmyc: I can't see where the `$products` variable was being used in this method so I removed it while in the file. But cc @joshuatf for 👀

joshuatf: Nice- this can definitely be removed.

timmyc:
Thanks for the reviews @joshuatf and @vedanshujain ( and for the bug report too! )

PR #3729, opened 2020-02-18
Stock panel indicator - cache and use lookup tables.

### LINKED ISSUE #3664: [Performance] Use caching to see if there are out of stock products
**Describe the bug**
Similar to #3663, looks like we also query to see if there are any out stock product to determine if we want to show a notification bell in Stock tab button. This query is also quite heavy, for egs:
```
# Query_time: 0.261732  Lock_time: 0.000154 Rows_sent: 1  Rows_examined: 88778
SET timestamp=1580822820;
SELECT SQL_CALC_FOUND_ROWS  wc_posts.*, low_stock_amount_meta.meta_value AS low_stock_amount FROM wc_posts  LEFT JOIN wc_wc_product_meta_lookup wc_product_meta_lookup ON wc_posts.ID = wc_product_meta_lookup.product_id  LEFT JOIN wc_postmeta AS low_stock_amount_meta ON wc_posts.ID = low_stock_amount_meta.post_id AND low_stock_amount_meta.meta_key = '_low_stock_amount'  WHERE 1=1  AND wc_posts.post_type IN ('product', 'product_variation') AND ((wc_posts.post_status = 'publish'))
			AND wc_product_meta_lookup.stock_quantity IS NOT NULL
			AND wc_product_meta_lookup.stock_status IN('instock','outofstock')
			AND (
				(
					low_stock_amount_meta.meta_value > ''
					AND wc_product_meta_lookup.stock_quantity <= CAST(low_stock_amount_meta.meta_value AS SIGNED)
				)
				OR (
					(
						low_stock_amount_meta.meta_value IS NULL OR low_stock_amount_meta.meta_value <= ''
					)
					AND wc_product_meta_lookup.stock_quantity <= 2
				)
			) GROUP BY wc_posts.ID ORDER BY wc_posts.post_date DESC, wc_posts.ID DESC LIMIT 0, 1;
```

This is not as bad as #3663 though because query is async after the page is loaded. That said, this will still hammer DB of our large users. We should cache this query.

**To Reproduce**

1. Go to dashboard with wc-admin activated.
2. Wait for page load, and then for notification icons on Stock button.
3. Check query log.

timmyc:
Fixes #3664 

This branch seeks to fix the performance issue that @vedanshujain outlined in the issue linked above. Some background, the [Activity Panel unread indicators logic](https://github.com/woocommerce/woocommerce-admin/blob/master/client/header/activity-panel/unread-indicators.js#L115-L132) currently fires off a REST API request on each page load to determine if there are low/out of stock products, and if so, the panel gets adorned with the unread indicator dot.

For the solution here, I took inspiration ( read, copied a great deal of the code ) from [the legacy out of stock panel](https://github.com/woocommerce/woocommerce/blob/dd49f89e77acaaa097251fe0a5dd69320ded48c6/includes/admin/class-wc-admin-dashboard.php#L204) that is in Woo Core. If this PR lands, we should really extract this logic into a shared method in Woo core that can be used in both places.

But I did change the logic to shorten the caching of the results to one hour instead of a day. The other added benefit of this approach is the query leverages the product lookup tables that were added in 3.6 for an added performance boost.

Apologies to the large changeset that was created by the autoformatter, the only change to `client/header/activity-panel/index.js` is line 317: `const hasUnreadStock = getUnreadStock();`

in `client/header/activity-panel/unread-indicators.js` the change is in `getUnreadStock()` - which is now set server-side, and pulled in via settings.

- Determine if you have any current low/out of stock products using `master` or another branch. Check the Stock panel to do this.
- Checkout this branch, verify your stock status indicator is the same as it was before.
- Either add stock to the out of stock products, or make a product out of stock to force the change in indicator. Refresh and verify the panel indicator is correct.
- Also verify the expensive query is not happening on every page load.

### Changelog Note:

Performance: Make Stock Panel indicator more performant.

### DIFF
--- a/src/Features/ActivityPanels.php
+++ b/src/Features/ActivityPanels.php
@@ -23,6 +23,11 @@ class ActivityPanels {
 	protected static $instance = null;
 
 	/**
+	 * Low Stock Transient Name.
+	 */
+	const LOW_STOCK_TRANSIENT_NAME = 'woocommerce_admin_low_out_of_stock_count';
+
+	/**
 	 * Get class instance.
 	 */
 	public static function get_instance() {
@@ -42,6 +47,7 @@ class ActivityPanels {
 		// New settings injection.
 		add_filter( 'woocommerce_shared_settings', array( $this, 'component_settings' ), 20 );
 		add_action( 'woocommerce_updated', array( $this, 'woocommerce_updated_note' ) );
+		add_action( 'woocommerce_update_product', array( __CLASS__, 'clear_low_out_of_stock_count_transient' ) );
 	}
 
 	/**
@@ -76,46 +82,29 @@ class ActivityPanels {
 			return false;
 		}
 
-		$stock   = absint( max( get_option( 'woocommerce_notify_low_stock_amount' ), 1 ) );
-		$nostock = absint( max( get_option( 'woocommerce_notify_no_stock_amount' ), 0 ) );
+		$low_stock_out_of_stock_count = get_transient( self::LOW_STOCK_TRANSIENT_NAME );
 
-		$transient_name   = 'woocommerce_admin_low_stock_count';
-		$lowinstock_count = get_transient( $transient_name );
-
-		if ( false === $lowinstock_count ) {
-			$lowinstock_count = (int) $wpdb->get_var(
-				$wpdb->prepare(
-					"SELECT COUNT( product_id )
-					FROM {$wpdb->wc_product_meta_lookup} AS lookup
-					INNER JOIN {$wpdb->posts} as posts ON lookup.product_id = posts.ID
-					WHERE stock_quantity <= %d
-					AND stock_quantity > %d
-					AND posts.post_status = 'publish'",
-					$stock,
-					$nostock
-				)
+		if ( false === $low_stock_out_of_stock_count ) {
+			$low_stock_out_of_stock_count = (int) $wpdb->get_var(
+				"SELECT COUNT( product_id )
+				FROM {$wpdb->wc_product_meta_lookup} AS lookup
+				INNER JOIN {$wpdb->posts} as posts ON lookup.product_id = posts.ID
+				WHERE stock_status IN ( 'onbackorder', 'outofstock' )
+				AND posts.post_status = 'publish'"
 			);
-			set_transient( $transient_name, $lowinstock_count, HOUR_IN_SECONDS );
+			set_transient( $self::LOW_STOCK_TRANSIENT_NAME, $low_stock_out_of_stock_count, HOUR_IN_SECONDS );
 		}
+		return $low_stock_out_of_stock_count > 0;
+	}
 
-		$transient_name   = 'woocommerce_admin_outofstock_count';
-		$outofstock_count = get_transient( $transient_name );
-
-		if ( false === $outofstock_count ) {
-			$outofstock_count = (int) $wpdb->get_var(
-				$wpdb->prepare(
-					"SELECT COUNT( product_id )
-					FROM {$wpdb->wc_product_meta_lookup} AS lookup
-					INNER JOIN {$wpdb->posts} as posts ON lookup.product_id = posts.ID
-					WHERE stock_quantity <= %d
-					AND posts.post_status = 'publish'",
-					$nostock
-				)
-			);
-			set_transient( $transient_name, $outofstock_count, HOUR_IN_SECONDS );
-		}
-
-		return $lowinstock_count > 0 || $outofstock_count > 0;
+	/**
+	 * Clears transient for out of stock indicator
+	 *
+	 * @return boolean
+	 */
+	public function clear_low_out_of_stock_count_transient() {
+		delete_transient( self::LOW_STOCK_TRANSIENT_NAME );
+		return true;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


src/Features/ActivityPanels.php
```diff
@@ -60,13 +60,72 @@ public function add_user_data_fields( $user_data_fields ) {
...
+					INNER JOIN {$wpdb->posts} as posts ON lookup.product_id = posts.ID
+					WHERE stock_quantity <= %d
+					AND posts.post_status = 'publish'",
+					$nostock
+				)
+			);
+			set_transient( $transient_name, $outofstock_count, HOUR_IN_SECONDS );
+		}
+
+		return $lowinstock_count > 0 || $outofstock_count > 0;
```
vedanshujain: [not a blocker]
if we don't care about individual values, then why not just do a single query for `WHERE stock_quantity <= MAX( out_of_stock_count, low_stock_count)`?

src/Features/ActivityPanels.php
```diff
@@ -60,13 +60,72 @@ public function add_user_data_fields( $user_data_fields ) {
...
+				)
+			);
+			set_transient( $transient_name, $lowinstock_count, HOUR_IN_SECONDS );
+		}
+
+		$transient_name   = 'woocommerce_admin_outofstock_count';
+		$outofstock_count = get_transient( $transient_name );
+
+		if ( false === $outofstock_count ) {
+			$outofstock_count = (int) $wpdb->get_var(
```
joshuatf: Since we're just after whether or not low or no stock products exist at all, maybe we can combine this with the above query. (e.g., `<= $nostock || <= $stock`)

src/Features/ActivityPanels.php
```diff
@@ -60,13 +60,72 @@ public function add_user_data_fields( $user_data_fields ) {
...
+	}
+
 	/**
 	 * Add alert count to the component settings.
 	 *
 	 * @param array $settings Component settings.
 	 */
 	public function component_settings( $settings ) {
 		$settings['alertCount'] = WC_Admin_Notes::get_notes_count( array( 'error', 'update' ), array( 'unactioned' ) );
+		$settings['hasLowStockAlerts'] = $this->has_low_stock_products();
```
joshuatf: Nitpicky, but maybe `hasLowStock` or `hasLowStockProducts`?  "Alerts" makes me think there are multiple alerts for this item.

src/Features/ActivityPanels.php
```diff
@@ -60,13 +60,72 @@ public function add_user_data_fields( $user_data_fields ) {
...
+
+		$transient_name   = 'woocommerce_admin_low_stock_count';
+		$lowinstock_count = get_transient( $transient_name );
+
+		if ( false === $lowinstock_count ) {
+			$lowinstock_count = (int) $wpdb->get_var(
+				$wpdb->prepare(
+					"SELECT COUNT( product_id )
+					FROM {$wpdb->wc_product_meta_lookup} AS lookup
+					INNER JOIN {$wpdb->posts} as posts ON lookup.product_id = posts.ID
```
joshuatf: I think we'll need another join on the product meta table since each product can have an individual low product stock threshold.

See https://github.com/woocommerce/woocommerce-admin/blob/5bed94bf9e3775ec58af2a9cc64a35d3fe767bfb/src/API/Reports/Stock/Stats/DataStore.php#L74 for prior art.

timmyc:
@joshuatf revised the query a bit further to leverage the meta lookup table to its full extent. I tested out this query with a product level low stock threshold and the lookup table takes care of that for us.

Regarding the two code paths for the notif indicator and the panel content itself potentially causing bugs, I think going through this performance work has uncovered a bug in the current panel implementation. I have a product that is out of stock ( -34 on hand, so back ordered ) and the new query here for the indicator correctly flags that as a low stock product, and sets the indicator on for "unread" on the panel... but when the panel is opened, the product is not shown.

Since this PR has all these crazy formatting changes, I'm thinking we can just merge in the performance side of the problem and I can follow up with a bug fix on the panel itself.

src/Features/ActivityPanels.php
```diff
@@ -60,13 +66,55 @@ public function add_user_data_fields( $user_data_fields ) {
...
+
+		if ( false === $low_stock_out_of_stock_count ) {
+			$low_stock_out_of_stock_count = (int) $wpdb->get_var(
+				"SELECT COUNT( product_id )
+				FROM {$wpdb->wc_product_meta_lookup} AS lookup
+				INNER JOIN {$wpdb->posts} as posts ON lookup.product_id = posts.ID
+				WHERE stock_status IN ( 'onbackorder', 'outofstock' )
+				AND posts.post_status = 'publish'"
+			);
+			set_transient( $self::LOW_STOCK_TRANSIENT_NAME, $low_stock_out_of_stock_count, HOUR_IN_SECONDS );
```
joshuatf: This throws an error:

```suggestion
			set_transient( self::LOW_STOCK_TRANSIENT_NAME, $low_stock_out_of_stock_count, HOUR_IN_SECONDS );
```

src/Features/ActivityPanels.php
```diff
@@ -42,6 +47,7 @@ public function __construct() {
 		// New settings injection.
 		add_filter( 'woocommerce_shared_settings', array( $this, 'component_settings' ), 20 );
 		add_action( 'woocommerce_updated', array( $this, 'woocommerce_updated_note' ) );
+		add_action( 'woocommerce_update_product', array( __CLASS__, 'clear_low_out_of_stock_count_transient' ) );
```
joshuatf: Noice.

src/Features/ActivityPanels.php
timmyc: yikes, wonder why i didn't get that to throw an error for me. good catch

PR #3730, opened 2020-02-18
Fix/3631

### LINKED ISSUE #3631: Onboarding - industry selection step: add CBD as an industry type
We need to add CBD as an option in the Industry step (2nd step of the Profiler) in order to have a sense of how many CBD focused stores are being created. It can be placed before "Other". In terms of copy it should be: CBD and other hemp-derived products

https://d.pr/i/Wf7g4N

We should also display an optional input for when the user selects "Other", so we can have a clear picture of what other industries are missing in our listing.

octaedro:
Fixes #3631

This PR adds to the onboarding process:

- The option `CBD and other hemp-derived products` to the industries list
- When the user clicks on the `Other` checkbox, an input is displayed.

### Screenshots
Screen Capture on 2020-02-18 at 14-02-20

**Test 1**
1. Start the onboarding process
2. In the `Industry` step check whether the option `CBD and other hemp-derived products` is visible and works
3. Select the option: `CBD and other hemp-derived products`
4. Press the `Continue` button. It should save the information in the database
5. Go back to the `Industry` step, reload the page and verify that the option `CBD and other hemp-derived products` is still checked. That means the data was successfully saved.

**Test 2**
Steps 1 and 2 are the same

3. Select the option: `Other`
4. An empty input should appear under the `Other` checkbox.
5. Fill the input with a description.
6. Press the `Continue` button.
7. Go back to the `Industry` step, reload the page and verify that the option `Other` is still selected and the input has the inserted text.

**Test 3**
First 5 steps are the same

6. After filling the input, press the `Other` checkbox again.
7. The input should disappear. Press again the `Other` checkbox.
8. The inserted text before should now be visible again.

### Changelog Note:
- Now we can add a new description input (for the other industries listed) changing the value `use_description` to `true` in the `get_allowed_industries` method in the file `src/Features/Onboarding.php`.
The current options for each item now are: `label` (_string_), `use_description` (_boolean_), `description_label` (_string_).
- The frontend will show an input for each list option, depending on the list above.
- Some styles were changed in order to better see the added inputs.

### DIFF
--- a/client/dashboard/profile-wizard/steps/industry.js
+++ b/client/dashboard/profile-wizard/steps/industry.js
@@ -6,7 +6,7 @@ import { __ } from '@wordpress/i18n';
 import { Component, Fragment } from '@wordpress/element';
 import { Button, CheckboxControl } from '@wordpress/components';
 import { compose } from '@wordpress/compose';
-import { filter, get, includes } from 'lodash';
+import { filter, get, find, findIndex } from 'lodash';
 import { withDispatch } from '@wordpress/data';
 
 /**
@@ -17,7 +17,7 @@ import { getSetting } from '@woocommerce/wc-admin-settings';
 /**
  * Internal dependencies
  */
-import { H, Card } from '@woocommerce/components';
+import { H, Card, TextControl } from '@woocommerce/components';
 import withSelect from 'wc-api/with-select';
 import { recordEvent } from 'lib/tracks';
 
@@ -31,9 +31,11 @@ class Industry extends Component {
 		this.state = {
 			error: null,
 			selected: profileItems.industry || [],
+			textInputListContent: [],
 		};
 		this.onContinue = this.onContinue.bind( this );
 		this.onChange = this.onChange.bind( this );
+		this.onInputChange = this.onInputChange.bind( this );
 	}
 
 	async onContinue() {
@@ -67,16 +69,20 @@ class Industry extends Component {
 	onChange( slug ) {
 		this.setState(
 			state => {
-				if ( includes( state.selected, slug ) ) {
+				const newSelected = state.selected;
+				const selectedIndustry = find( newSelected, { slug: slug } );
+				if ( selectedIndustry ) {
+					const newTextInputListContent = state.textInputListContent;
+					newTextInputListContent[ slug ] = selectedIndustry.detail;
 					return {
 						selected:
 							filter( state.selected, value => {
-								return value !== slug;
+								return value.slug !== slug;
 							} ) || [],
+						textInputListContent: newTextInputListContent,
 					};
 				}
-				const newSelected = state.selected;
-				newSelected.push( slug );
+				newSelected.push( { slug, detail: state.textInputListContent[ slug ] } );
 				return {
 					selected: newSelected,
 				};
@@ -85,9 +91,20 @@ class Industry extends Component {
 		);
 	}
 
+	onInputChange( value, slug ) {
+		this.setState( state => {
+			const newSelected = state.selected;
+			const industryIndex = findIndex( newSelected, { slug: slug } );
+			newSelected[ industryIndex ].detail = value;
+			return {
+				selected: newSelected,
+			};
+		} );
+	}
+
 	render() {
 		const { industries } = onboarding;
-		const { error, selected } = this.state;
+		const { error, selected, textInputListContent } = this.state;
 
 		return (
 			<Fragment>
@@ -100,14 +117,28 @@ class Industry extends Component {
 				<Card>
 					<div className="woocommerce-profile-wizard__checkbox-group">
 						{ Object.keys( industries ).map( slug => {
+							const selectedIndustry = find( selected, { slug: slug } );
+
 							return (
-								<CheckboxControl
-									key={ slug }
-									label={ industries[ slug ] }
-									onChange={ () => this.onChange( slug ) }
-									checked={ selected.includes( slug ) }
-									className="woocommerce-profile-wizard__checkbox"
-								/>
+								<div key={ `div-${ slug }` }>
+									<CheckboxControl
+										key={ `checkbox-control-${ slug }` }
+										label={ industries[ slug ].label }
+										onChange={ () => this.onChange( slug ) }
+										checked={ selectedIndustry || false }
+										className="woocommerce-profile-wizard__checkbox"
+									/>
+									{ industries[ slug ].use_description &&
+										selectedIndustry && (
+											<TextControl
+												key={ `text-control-${ selectedIndustry.slug }` }
+												label={ industries[ selectedIndustry.slug ].description_label }
+												value={ selectedIndustry.detail || textInputListContent[ slug ] || '' }
+												onChange={ value => this.onInputChange( value, selectedIndustry.slug ) }
+												className="woocommerce-profile-wizard__text"
+											/>
+										) }
+								</div>
 							);
 						} ) }
 						{ error && <span className="woocommerce-profile-wizard__error">{ error }</span> }
--- a/client/dashboard/profile-wizard/steps/industry.js
+++ b/client/dashboard/profile-wizard/steps/industry.js
@@ -52,7 +52,7 @@ class Industry extends Component {
 
 		recordEvent( 'storeprofiler_store_industry_continue', {
 			store_industry: selectedIndustriesList,
-			industriesWithDetail,
+			industries_with_detail: industriesWithDetail,
 		} );
 		await updateProfileItems( { industry: this.state.selected } );
 
--- a/client/dashboard/profile-wizard/steps/industry.js
+++ b/client/dashboard/profile-wizard/steps/industry.js
@@ -45,8 +45,15 @@ class Industry extends Component {
 		}
 
 		const { createNotice, goToNextStep, isError, updateProfileItems } = this.props;
+		const selectedIndustriesList = this.state.selected.map( industry => industry.slug );
+		const industriesWithDetail = filter( this.state.selected, value => {
+			return typeof value.detail !== 'undefined';
+		} );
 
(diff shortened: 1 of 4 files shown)


jeffstieler:
The functionality tests well, but I ended up with an unexpected value for `industry` in my database:

```json
[
  "other",
  {
    "slug": "cbd-other-hemp-derived-products"
  },
  {
    "slug": "other",
    "detail": "Stuff."
  }
]
```

I had previously selected "Other" before switching to this branch to test.

I'll review the code now.

jeffstieler:
Hmm.. it looks like trying to erase the value in the Other description field results in a restoration of the value from `props`:

client/dashboard/profile-wizard/steps/industry.js
```diff
@@ -31,9 +31,11 @@ class Industry extends Component {
 		this.state = {
 			error: null,
 			selected: profileItems.industry || [],
+			textInputListContent: [],
```
jeffstieler: This is an array, but it looks to be used like an object later?

client/dashboard/profile-wizard/steps/industry.js
```diff
@@ -85,9 +91,20 @@ class Industry extends Component {
 		);
 	}
 
+	onInputChange( value, slug ) {
```
jeffstieler: Would `onDetailChange` be a more descriptive/accurate name?

client/dashboard/profile-wizard/steps/industry.js
```diff
@@ -67,16 +69,20 @@ class Industry extends Component {
...
+				if ( selectedIndustry ) {
+					const newTextInputListContent = state.textInputListContent;
+					newTextInputListContent[ slug ] = selectedIndustry.detail;
 					return {
 						selected:
 							filter( state.selected, value => {
-								return value !== slug;
+								return value.slug !== slug;
 							} ) || [],
+						textInputListContent: newTextInputListContent,
```
jeffstieler: I wasn't able to see this state value change in my testing.

client/dashboard/profile-wizard/steps/industry.js
```diff
@@ -67,16 +69,20 @@ class Industry extends Component {
...
 					};
 				}
-				const newSelected = state.selected;
-				newSelected.push( slug );
+				newSelected.push( { slug, detail: state.textInputListContent[ slug ] } );
```
jeffstieler: Perhaps this is the root of the "backspace resets old value" bug?

client/dashboard/profile-wizard/steps/industry.js
```diff
@@ -67,16 +69,20 @@ class Industry extends Component {
 	onChange( slug ) {
```
jeffstieler: Perhaps renaming this `onIndustryChange` or `onCheckboxChange` makes sense now that there are multiple input handlers.

client/dashboard/profile-wizard/steps/industry.js
```diff
@@ -100,14 +117,28 @@ class Industry extends Component {
...
+										onChange={ () => this.onChange( slug ) }
+										checked={ selectedIndustry || false }
+										className="woocommerce-profile-wizard__checkbox"
+									/>
+									{ industries[ slug ].use_description &&
+										selectedIndustry && (
+											<TextControl
+												key={ `text-control-${ selectedIndustry.slug }` }
+												label={ industries[ selectedIndustry.slug ].description_label }
+												value={ selectedIndustry.detail || textInputListContent[ slug ] || '' }
```
jeffstieler: Ah, perhaps this is the culprit for the resetting field value?

packages/components/src/form/docs/example.js
```diff
@@ -77,8 +77,11 @@ export default () => (
 				</Button>
 				<br />
 				<br />
-				Values: { JSON.stringify( values ) }<br />
-				Errors: { JSON.stringify( errors ) }<br />
+				<h3>Return data:</h3>
+				<pre>
+					Values: { JSON.stringify( values ) }<br />
+					Errors: { JSON.stringify( errors ) }
+				</pre>
```
jeffstieler: Was this change/commit intended to be included?

src/API/OnboardingProfile.php
```diff
@@ -240,12 +240,7 @@ public static function get_profile_properties() {
 				'description'       => __( 'Industry.', 'woocommerce-admin' ),
 				'context'           => array( 'view' ),
 				'readonly'          => true,
-				'sanitize_callback' => 'wp_parse_slug_list',
 				'validate_callback' => 'rest_validate_request_arg',
-				'items'             => array(
-					'enum' => array_keys( Onboarding::get_allowed_industries() ),
-					'type' => 'string',
-				),
```
jeffstieler: Is it possible to describe the new structure in JSON Schema here instead of removing validation?

pmcpinto:
> EDIT: It should be noted that this is the value that gets sent to Tracks as well. Previously it was a simple array of strings. Does this affect any funnels? cc: @pmcpinto.

@jeffstieler I'm thinking that the best way to see and filter the data on tracks would be the following:

- Add `BD and other hemp-derived products` to the `store_industry`
- Have a new event prop that records the values added on `Other`

This way we can continue simply comparing the % of users selecting `Other` with the other options, without having the text added to `Other` in the mix. And then we have a separate event prop where we can dig into the details of what was added as `Other`. Makes sense?

src/API/OnboardingProfile.php
octaedro: Good idea! I'll do it. I'll validate the datatype: `json`.

client/dashboard/profile-wizard/steps/industry.js
octaedro: Totally! I should have declared an object instead of an array.

client/dashboard/profile-wizard/steps/industry.js
octaedro: You are right, I'm changing the name now.

client/dashboard/profile-wizard/steps/industry.js
octaedro: I'm saving this value to be able to show the old input content after unchecking the option. As we remove the item from `state.selected` I save the value to not lose it if the user presses the input again.

client/dashboard/profile-wizard/steps/industry.js
octaedro: I will change it too.

client/dashboard/profile-wizard/steps/industry.js
octaedro: Yes! This bug happens when the data is already saved and the `Other` option is unselected and reselected. Now the old value is updating every time the input changes.

packages/components/src/form/docs/example.js
octaedro: Not at all. I never touched this file. Probably I did something wrong with the pull request or the branch... I will remove it.

jeffstieler: It's probably best to remove the commit entirely - you can do so using `git rebase` (I like using it in interactive mode `-i`)

src/API/OnboardingProfile.php
jeffstieler: Interesting.. it seems this isn't used anyhow: https://github.com/woocommerce/woocommerce-admin/blame/b75ee0cbccb0540e69ff7c3b29f6b4f3f8b2039d/src/API/OnboardingProfile.php#L370-L376

Ideally we'd define the full schema, not just "type = json", but since the validation isn't being used we can save that for a future PR. _cc: @joshuatf who wrote the referenced code_

octaedro:
Hi @pmcpinto,

I have some questions about the solution you suggested.

In my approach, I was using an array of `objects` instead of an array of `strings`.
So we will send to the API something like:
`[ { slug: "other", detail: "a detail" } ]` instead of `[ "other" ]`

By sending an array of `objects` we only make one API call instead of two and we would keep all the related information together. Would this solution create an extra complication or problem?

If you think the other approach is better, I can change the way we handle the data, but I wanted to be sure of the details before I start with these changes.

Thanks!

jeffstieler:
@octaedro I don't think @pmcpinto is suggesting a second _event_, just an additional event _property_.

He's suggesting sending the same format as before, where `store_industry` is an array of strings and a new object property `other` could hold the string from the text field.

jeffstieler:
This is testing much better. Let's remove the errantly included commit, and rework the Tracks event data.

octaedro:
Ok, sounds good :-)

Thinking about the new object property.  In the current solution, it's also possible to set other "detail inputs" for other industries (not only for `other`) by changing the variable `use_description` to `true` in the `get_allowed_industries` method in the file `src/Features/Onboarding.php`.

Is it ok if I use an array instead of a string in the new object property? (The idea is to make it easier to add other "detail" text inputs for other industries if we want to in the future)

client/dashboard/profile-wizard/steps/industry.js
```diff
@@ -45,8 +45,15 @@ class Industry extends Component {
...
 		const { createNotice, goToNextStep, isError, updateProfileItems } = this.props;
+		const selectedIndustriesList = this.state.selected.map( industry => industry.slug );
+		const industriesWithDetail = filter( this.state.selected, value => {
+			return typeof value.detail !== 'undefined';
+		} );
 
-		recordEvent( 'storeprofiler_store_industry_continue', { store_industry: this.state.selected } );
+		recordEvent( 'storeprofiler_store_industry_continue', {
+			store_industry: selectedIndustriesList,
+			industriesWithDetail,
```
jeffstieler: Let's stick with `snake_case` for event props.

octaedro: Good idea!

pmcpinto:
> He's suggesting sending the same format as before, where store_industry is an array of strings and a new object property other could hold the string from the text field.

Yes exactly, having the string from the text field in store_industry would add a lot of noise.

PR #3731, opened 2020-02-18
Checklist: Remove sideloaded images to reduce build size, take 2

### LINKED ISSUE #3716: [4.0 beta] Review the image assets bundled with WC Admin

timmyc:
Fixes #3716 
Supersedes #3719 

Current builds of woocommerce-admin include over 5mb of images that are ultimately sideloaded - downloaded from the plugin directory, and uploaded to the media library. These images drastically increase the size of the build and in-turn, inflate the size of WooCommerce core to over 12mb.

I attempted to run the images through a jpg optimizer, and they still weighed in at 2.7mb total - which to me feels still way too costly for what they are ultimately used for. Furthermore, any existing woo site will never use these images, since they are only leveraged during the Homepage creation step in the new site task list.

As such, the _new_ proposed solution was @jameskoster did some big trimming of the `other` industry jpg to `57kb`, and this PR uses that one image across all industries. The other change I made due to the dark image was to change the font color in the image block to be white.

/cc @peterfabian 

- Disable Jetpack, otherwise the new post flow will be directed to WordPress.com and that kind of makes testing this more difficult
- Go to Orders > Help > Setup Wizard and click the Enable button for the task list:

image

- Go back to the Woo dashboard, and click the Personalize Your Store step, and then the create homepage button:

image

- Verify the newly created homepage looks like this regardless of industry chosen:

Edit Page ‹ bend outdoors — WordPress 2020-02-18 11-41-30

### Changelog Note:

Performance: Remove sideloaded images to save on build size


src/API/OnboardingTasks.php
```diff
@@ -187,12 +187,12 @@ public static function sanitize_special_column_name_regex( $value ) {
...
 			return '<!-- wp:cover {"url":"' . esc_url( $image['url'] ) . '","id":' . intval( $image['id'] ) . '} -->
-			<div class="wp-block-cover has-background-dim" style="background-image:url(' . esc_url( $image['url'] ) . ')"><div class="wp-block-cover__inner-container"><!-- wp:paragraph {"align":"center","placeholder":"' . __( 'Write title…', 'woocommerce-admin' ) . '","fontSize":"large"} -->
+			<div class="wp-block-cover has-background-dim" style="background-image:url(' . esc_url( $image['url'] ) . ')"><div class="wp-block-cover__inner-container"><!-- wp:paragraph {"align":"center","placeholder":"' . __( 'Write title…', 'woocommerce-admin' ) . '","textColor":"white","fontSize":"large"} -->
 			<p class="has-text-align-center has-large-font-size">' . __( 'Welcome to the store', 'woocommerce-admin' ) . '</p>
 			<!-- /wp:paragraph -->
 
-			<!-- wp:paragraph {"align":"center"} -->
-			<p class="has-text-align-center">' . __( 'Write a short welcome message here', 'woocommerce-admin' ) . '</p>
+			<!-- wp:paragraph {"align":"center","textColor":"white"} -->
+			<p class="has-text-color has-text-align-center">' . __( 'Write a short welcome message here', 'woocommerce-admin' ) . '</p>
```
becdetat: Why is `has-text-color` needed? The CSS rules are a bit complicated around this class being/not being present.

timmyc: I wondered the exact same thing - this template, from what I can tell, is pretty much a copy/paste from the block editor, so I was following that pattern in this part of the PR. When I applied the white font color from the block editor sidebar, the classname was added.

I'd have to go dig into gutenberg to see the original reason for `has-text-color`

becdetat: That's cool - there's probably a great reason for it

becdetat:
Looks great, testing works well. It would have been _nice_ to have a pre-canned image per industry but I can understand the need to keep the plugin small.

timmyc:
@jameskoster does this look okay:

image

Issue #3709, opened 2020-02-12
Onboarding - store profiler: change JP and WCS screen placement


pmcpinto:
Currently having this screen as the welcome screen is not producing great results. We think that in terms of experience it will be better if we allow users to fill first information about their store and then giving them the option to install & connect Jetpack and WooCommerce Services. This screen and subsequent install & connect flows should be placed at the end of the Store Profiler. So just to summarize:

- The first screen becomes the Store Details screen and the usage tracking authorization message is triggered in this screen (it already happens when the user has JP and WCS installed, because the JP/WCS is skipped). I guess that we need to change the logic to always display the auth message in this screen.
- The JP and WCS screen start being displayed after the user completes the Theme step
- We need to change the name of the event that's triggered when the user clicks in one of the buttons on this screen (`wcadmin_storeprofiler_welcome_clicked`) to `wcadmin_storeprofiler_install_plugins`. The event props stay the same.
- We need to add a new event to track when we just have a pageview on this screen: `wcadmin_storeprofiler_install_plugins`
- If we have time during this cycle we should also consider revamping the design of this screen, to make it more compelling. Design proposal [here](https://www.figma.com/file/nzAjPR4nqdJTDxH32PT4Qg/On-boarding?node-id=15%3A881).*
- The rest of the JP and WCS install & connect flow remains the same.

*- it would be useful to have an estimation of how long it will take to finish this with or without the screen redesign.

joshuatf:
>If we have time during this cycle we should also consider revamping the design of this screen, to make it more compelling. Design proposal here.*

So it looks like in the new design those 3 benefits are always shown and never change- is that right? 
/cc @pmcpinto @jameskoster 

Note the estimate for this issue does not include the design changes which I'll open another issue for so we can tackle.

jameskoster:
@joshuatf I added situational designs [here](https://www.figma.com/file/nzAjPR4nqdJTDxH32PT4Qg/On-boarding?node-id=0%3A1). One for when neither Jetpack or WCS are present. And one for when WCS isn't present, but Jetpack is.

joshuatf:
One small issue we have updating this flow is that currently the cart modal shows products for purchase on the last step.  But since the last step is possibly sending users to connect Jetpack, they won’t see the cart modal.

Maybe this is a good opportunity to also focus the cart modal efforts in the task list, showing the cart modal after the initial task list welcome modal.  What are your thoughts, @jameskoster @pmcpinto ?

[...]

PR #3739, opened 2020-02-19
Onboarding: Update screen order and remove Jetpack connection in profiler

joshuatf:
Fixes (part of) #3709 

* Moves the plugin install screens to the end of the profiler
* Removes the Jetpack connection from the profiler
* Removes the cart modal from the profiler
* Removes usage modal from plugins screen and enforces it on store details step
* Fixes some existing event props

@pmcpinto @jameskoster Some of the wording on the "Benefits" step (previously the start step) is weird since it's at the end.  `Start setting up your WooCommerce store` - not sure if we want to do any quick changes since the new design changes will fix this anyway.

1. Delete the onboarding data `woocommerce_onboarding_profile`
1. Try deactivating/deleting a combination of Jetpack and WCS plugins and also try with Jetpack connected/disconnected.
1. Enable the profiler.
1. Make sure the screen order is correct. (Benefits and Plugins screen moved to the end).
1. Make sure usage modal is enabled during store details step.
1. Walk through all the steps until the Benefits page (previously "start" page).
1. Make sure both "Get Started" and "skip" options work as expected.
1. Make sure plugins screen installs plugins and completes profiler after activation.
1. Make sure no cart modal is shown.


psealock:
>Start setting up your WooCommerce store - not sure if we want to do any quick changes since the new design changes will fix this anyway.

Definitely want to change this. It feels out of place, "Continue" would be better.

psealock:
>Make sure no cart modal is shown.

When does the user pay for premium plugins? You click on it and nothing happens in regard to a plugin like subscriptions. Would it be better to show some sort of instructions at least? I like that the modal gave an option to "do it later".

client/dashboard/components/cart-modal.js
```diff
@@ -37,13 +37,12 @@ class CartModal extends Component {
...
 		this.setState( { purchaseNowButtonBusy: true } );
 		if ( ! productIds.length ) {
-
 			return;
 		}
 
 		recordEvent( 'tasklist_modal_proceed_checkout', {
 			product_ids: productIds,
-			purchase_install: false,
+			purchase_install: true,
```
psealock: Should this just be removed to avoid confusion when looking at Tracks later on?

joshuatf:
>Definitely want to change this. It feels out of place, "Continue" would be better.

Agreed @psealock , but we do have another PR that changes the design and copy for this page hopefully out in a few days so this would be a very temporary change.

>When does the user pay for premium plugins?

The cart modal is still used for the task list, it has just  been removed from the end of the profiler.  More discussion in https://github.com/woocommerce/woocommerce-admin/issues/3709 but curious if you feel we should still have the cart modal in both places (now it would be after plugin install and again at the task list).

client/dashboard/components/cart-modal.js
joshuatf: Not quite sure I follow.  This event prop helps track if they closed the modal without purchasing or opted to purchase now.

pmcpinto:
> @pmcpinto @jameskoster Some of the wording on the "Benefits" step (previously the start step) is weird since it's at the end. Start setting up your WooCommerce store - not sure if we want to do any quick changes since the new design changes will fix this anyway.

Good point. I would change it go "Start enhancing your WooCommerce store"

joshuatf:
>Good point. I would change it go "Start enhancing your WooCommerce store"

Updated.  I think this is enough until we get the new designs for this page shipped.

rrennick:
This needs to be rebased to `version/1.0`.

timmyc:
> This needs to be rebased to version/1.0.

@rrennick this should target `master` since it isn't shipping in 4.0 correct?

joshuatf:
@timmyc @rrennick I believe this should be targeting `master` since we're not planning on including it in 4.0.

rrennick:
@joshuatf I get an error (then whitescreen) on business details when I click `How many products...`.

joshuatf:
@rrennick I'm unable to replicate but it sounds like the issue resolved in this PR - https://github.com/woocommerce/woocommerce-admin/pull/3755

Any other details you can give to recreate the issue?  Does rebasing solve for you?

rrennick:
> Does rebasing solve for you?

@joshuatf It did.

rrennick:
@joshuatf If I choose to install either MailChimp or Facebook I'm not taken to the plugin install screen after selecting the theme. I'd guess this is because theme used to be the last page & by install one of the extensions the plugin install flag is already set.

joshuatf:
>If I choose to install either MailChimp or Facebook I'm not taken to the plugin install screen after selecting the theme.

@rrennick These plugins get installed directly on the Business Details page in the background.  The final plugins screen should only show if Jetpack/WCS need installation/activation.

rrennick:
> The final plugins screen should only show if Jetpack/WCS need installation/activation.

It left me sitting on the theme selection screen after I selected the theme. If the selected plugins are already installed it should still redirect to the task list.

joshuatf:
@rrennick I'm having a hard time reproducing the issue.  You should end up on the benefits screen after theme selection:

Screen Shot 2020-02-27 at 9 39 18 PM

Did you have any console errors or network errors that occurred when you got stuck on the theme step?

client/dashboard/profile-wizard/steps/benefits/index.js
```diff
@@ -153,7 +123,7 @@ class Start extends Component {
 					'Jetpack automatically blocks brute force attacks to protect your store from unauthorized access.',
 					'woocommerce-admin'
 				),
-				visible: ! isJetpackConnected,
+				visible: ! activePlugins.includes( 'jetpack' ),
```
timmyc: Since this logic is repeated a few times in this method, should we just assign the value to a variable and use it instead? `const isJetpackActive = activePlugins.includes( 'jetpack' )`

rrennick: > when did you last do the merge from the 1.0 branch?

Last week's is waiting for @psealock to give it a second look.

joshuatf: Updated this in https://github.com/woocommerce/woocommerce-admin/pull/3739/commits/775cac3c3fa5ce2df3b373a8d91cdab463c477f5

joshuatf:
>Which I think was fixed last week for our last release, so wondering if we need to get a merge from version/1.0 to master and then rebase here?

@timmyc We can wait for that if you prefer, but this PR also doesn't touch on the "Business Details" step or adjacent steps so I'm thinking we could just test around it for the time being since we know this is fixed in `version/1.0`. 

You can skip to the "Theme" step by changing the `step` param in the URL.  `wp-admin/admin.php?page=wc-admin&step=theme`

psealock:
https://github.com/woocommerce/woocommerce-admin/pull/3797 is merged, so the error seen by @timmyc should be resolved in master.

PR #3743, opened 2020-02-20
Fix image sideloading for store industries.

jeffstieler:
Data format for industries changed in https://github.com/woocommerce/woocommerce-admin/pull/3730

Fixes a PHP warning encountered in `sideload_homepage_images()`: `PHP Warning: Illegal offset type in isset or empty in /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-admin/src/API/OnboardingTasks.php on line 344`

- Go through the profiler again, selecting one or more industries
- Ensure your store has less than 4 products (or modify this line: https://github.com/woocommerce/woocommerce-admin/blob/24734f861cd00e2a61d5292b5e3338b43a0da33c/src/API/OnboardingTasks.php#L253)
- Use the task list to create a homepage
- Verify no PHP warnings in error logs, homepage has all expected images
- Ensure your store has more than 4 products (or modify this line: https://github.com/woocommerce/woocommerce-admin/blob/24734f861cd00e2a61d5292b5e3338b43a0da33c/src/API/OnboardingTasks.php#L253)
- Use the task list to create a homepage
- Verify no PHP warnings in error logs, homepage has all expected images

### DIFF
--- a/src/API/OnboardingTasks.php
+++ b/src/API/OnboardingTasks.php
@@ -341,10 +341,15 @@ class OnboardingTasks extends \WC_REST_Data_Controller {
 
 		if ( ! empty( $profile['industry'] ) ) {
 			foreach ( $profile['industry'] as $selected_industry ) {
-				if ( ! is_array( $selected_industry ) || empty( $selected_industry['slug'] ) ) {
+				if ( is_string( $selected_industry ) ) {
+					$industry_slug = $selected_industry;
+				} elseif ( is_array( $selected_industry ) && ! empty( $selected_industry['slug'] ) ) {
+					$industry_slug = $selected_industry['slug'];
+				} else {
 					continue;
 				}
-				$industry_slug        = $selected_industry['slug'];
+				// Capture the first industry for use in our minimum images logic.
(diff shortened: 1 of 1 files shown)


src/API/OnboardingTasks.php
```diff
@@ -341,14 +341,28 @@ private static function sideload_homepage_images( $post_id, $number_of_images )
 
 		if ( ! empty( $profile['industry'] ) ) {
 			foreach ( $profile['industry'] as $selected_industry ) {
-				$images_to_sideload[] = ! empty( $available_images[ $selected_industry ] ) ? $available_images[ $selected_industry ] : $available_images['other'];
+				if ( ! is_array( $selected_industry ) || empty( $selected_industry['slug'] ) ) {
+					continue;
```
joshuatf: Should we offer backwards compatibility here for users that have already completed the profiler and still have the task list enabled?  Or do we think this is more of an edge case?

jeffstieler: While I do think that's an edge case - it's not too hard to handle.

jeffstieler: Done in 7d3fbf0.

PR #3750, opened 2020-02-20
Added event props to identify stores with WCS and Jetpack installed

### LINKED ISSUE #3728: Onboarding: Add event props to identify stores with WCS and Jetpack installed
In order to better identify funnel conversions, we should track if the following are active in a store when a task is being viewed via the `wcadmin_task_view` event:

* WCS is installed
* WCS is active
* Jetpack is installed
* Jetpack is active
* Jetpack is connected

This may be soon replaced by global event props, but this is a good temporary solution until that lands.

See p1581956393013000-slack-wc-onboarding for full discussion.

octaedro:
Fixes #3728 

Added the following event props to `task_view` for tracking:
- `WCS` is installed
- `WCS` is active
- `Jetpack` is installed
- `Jetpack` is active
- `Jetpack` is connected

- Go to the `Onboarding tasks list` and got to any `task`. 
- Check if the event props `wcs_installed`, `wcs_active`, `jetpack_installed`, `jetpack_active` and `jetpack_connected` are being tracked.

### Changelog Note:
- This may be soon replaced by global event props, but this is a good temporary solution until that lands.

### DIFF
--- a/client/dashboard/task-list/index.js
+++ b/client/dashboard/task-list/index.js
@@ -4,7 +4,7 @@
  */
 import { __ } from '@wordpress/i18n';
 import { Component, Fragment } from '@wordpress/element';
-import { get } from 'lodash';
+import { get, intersection } from 'lodash';
 import { compose } from '@wordpress/compose';
 import classNames from 'classnames';
 import { Snackbar, Icon, Button, Modal } from '@wordpress/components';
@@ -15,6 +15,7 @@ import { withDispatch } from '@wordpress/data';
  */
 import { Card, List, MenuItem, EllipsisMenu } from '@woocommerce/components';
 import { updateQueryString } from '@woocommerce/navigation';
+import { getSetting } from '@woocommerce/wc-admin-settings';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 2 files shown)


client/dashboard/task-list/index.js
```diff
@@ -74,15 +75,32 @@ class TaskDashboard extends Component {
...
 	}
 
+	getPluginsInformation() {
+		const { isJetpackConnected } = this.props;
+		const plugins = [ 'jetpack', 'woocommerce-services' ];
+		const { activePlugins, installedPlugins } = getSetting( 'onboarding', {} );
+		const matchedInstalledPlugins = intersection( installedPlugins, plugins );
+		const matchedActivePlugins = intersection( activePlugins, plugins );
+		return {
+			wcs_installed: matchedInstalledPlugins.includes( 'woocommerce-services' ),
```
joshuatf: Instead of using `matchedInstalledPlugins` could we simplify this by just using `installedPlugins.includes( 'woocommerce-services' )`?

Same question for the next 3 lines.

client/dashboard/task-list/index.js
```diff
@@ -74,15 +75,32 @@ class TaskDashboard extends Component {
...
+			wcs_active: matchedActivePlugins.includes( 'woocommerce-services' ),
+			jetpack_installed: matchedInstalledPlugins.includes( 'jetpack' ),
+			jetpack_active: matchedActivePlugins.includes( 'jetpack' ),
+			jetpack_connected: isJetpackConnected,
+		};
+	}
+
 	recordTaskView() {
 		const { task } = this.props.query;
+		const pluginsInformation = this.getPluginsInformation();
```
joshuatf: Nice!  I like how this is neatly extracted to its own method.

client/dashboard/task-list/index.js
octaedro: Thank you Joshua, you are absolutely right! I will change it now.

PR #3700, opened 2020-02-11
'All that include' option removed when input field is empty

octaedro:
Fixes #3682 

Added control to check that when the input field `Search by customer name` is empty, the `All that include` option is not appearing.

### Screenshots
screenshot-one wordpress test-2020 02 11-18_44_36

### Detailed test instructions:
1. Go to `Analytics` > `Customers` report.
2. Click on `Search by customer name` input field.
3. If the input is empty, nothing should be shown.

### Changelog Note:
Added new control in `/packages/components/src/select-control/list.js`.

PR #3755, opened 2020-02-21
Remove free text Search option when no query exists

### LINKED ISSUE #3754: [0.26.0] Select Control throws error on OBW
**Describe the bug**

In the OBW, bringing focus to the Country/State field throws an error.

Screen Shot 2020-02-21 at 4 06 27 PM

**To Reproduce**

1. Enable the OBW in the `version/1.0` branch.
2. Go to Store Details, `/wp-admin/admin.php?page=wc-admin&step=store-details`.
3. Focus on the Country/State dropdown.
4. See the error.

This line causes problems because the option does not have `value` property.

https://github.com/woocommerce/woocommerce-admin/blob/8a23c9ad39ea2aec2478bfdff71552bd308311f8/packages/components/src/select-control/list.js#L154

Simply checking for the `value` property does not fix the problem. The component should not make too many assumptions about the shape of the data. A solution needs to make sure not to regress https://github.com/woocommerce/woocommerce-admin/pull/3700.

joshuatf:
Fixes #3754 

Moves the logic for #3700 into the `Search` component.

### Screenshots
Screen Shot 2020-02-21 at 10 32 41 AM
Screen Shot 2020-02-21 at 10 31 17 AM

1. Open the onboarding profiler store details step.
1. Focus the `SelectControl` (country/state dropdown) and make sure no errors occur.
1. Go to an analytics screen with a `Search` component and free text search option (e.g., Products or Customers).
1. Click on the `Search` component and make sure no options are shown.
1. Type a query and make sure you're given the "free text" search option (e.g. "All products with titles that include...").

### DIFF
--- a/packages/components/src/search/test/index.js
+++ b/packages/components/src/search/test/index.js
@@ -10,6 +10,18 @@ import { shallow } from 'enzyme';
 import { Search } from '../index';
 
 describe( 'Search', () => {
+	it( 'shows the free text search option', () => {
+		const search = shallow(
+			<Search
+				type="products"
+				allowFreeTextSearch
+			/>
+		);
+		const options = search.instance().appendFreeTextSearch( [], 'Product Query' );
+
+		expect( options.length ).toBe( 1 );
+	} );
+
 	it( "doesn't show options with an empty search", () => {
 		const search = shallow(
 			<Search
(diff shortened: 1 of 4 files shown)


packages/components/src/search/test/index.js
```diff
@@ -0,0 +1,37 @@
...
+				type="products"
+				allowFreeTextSearch
+			/>
+		);
+		const options = search.instance().appendFreeTextSearch( [], '' );
+
+		expect( options.length ).toBe( 0 );
+	} );
+} );
+
```
jeffstieler: ❤️

PR #3764, opened 2020-02-24
Onboarding: Redesign plugin benefits screen

### LINKED ISSUE #3709: Onboarding - store profiler: change JP and WCS screen placement

joshuatf:
Fixes #3709 

Redesign the benefits screen to match the new mockups.

@jameskoster Can you give this a quick review- specifically smaller screen sizes?

### Screenshots
Screen Shot 2020-02-24 at 1 50 20 PM

1. Uninstall Jetpack and WooCommerce services.
1. Walk through the profiler.
1. Make sure the benefits (after the theme step) screen matches the design.
1. Make sure install still works.
1. Try deactivating WCS or Jetpack and re-enable the profiler.
1. Make sure the benefits and plugin names change according to what's not installed and activated.


client/dashboard/profile-wizard/steps/benefits/index.js
```diff
@@ -202,87 +187,60 @@ class Benefits extends Component {
 		}
 		const pluginNamesString = pluginsToInstall
 			.map( ( pluginSlug ) => pluginNames[ pluginSlug ] )
-			.join( ' & ' );
+			.join( ' ' + __( 'and', 'woocommerce-admin' ) + ' ' );
```
joshuatf: @rrennick Do you think this will work okay for translation?

rrennick: >Do you think this will work okay for translation?

That works okay. Only the `'and'` is passed through translation. The spaces are concatenated after translation.

joshuatf:
> One small tweak - the button should read "Yes please!" rather than "Install Jetpack". Oh, and the 'disclaimer' should read:
“Jetpack” and “WooCommerce Services” plugins will be installed & activated for free.

@jameskoster This screenshot above is shown if WCS is already installed.  If Jetpack is installed already and Jetpack isn't, then it reads "Install WooCommerce" and “WooCommerce Services plugin will be installed & activated for free."

If neither are active it will read "Yes please!" and “Jetpack and WooCommerce Services plugins will be installed & activated for free."

Do you want to change it to always read "Yes please!" and always show both plugin names?

jameskoster:
Ah, apologies for the confusion.

Yes, let's make the button always read "Yes please!". 

But we only need to display the name of the plugin(s) being installed in the disclaimer.

joshuatf:
Thanks for the feedback, @jameskoster.  Updated this button to always read "Yes please!"

jameskoster:
That's correct, we're going to prompt the connection later as so many people drop out of the flow at this point currently.

joshuatf:
Hey @psealock - that's correct as @jameskoster noted.  We decided to remove the connection portion of the profiler (see [this comment](https://github.com/woocommerce/woocommerce-admin/issues/3709#issuecomment-588147973)).

Issue #3785, opened 2020-02-27
OBW: Jetpack connect requesting state points to bad url


psealock:
**Describe the bug**

On the tax step of the Task List, if you have Jetpack disconnected, the `isBusy` state of the "Connect" button leads to a bad url. Its also misleading because the connection is not happening. The button should be disabled and a spinner should appear in the step number icon.

![Screen Shot 2020-02-27 at 2 48 00 PM](https://user-images.githubusercontent.com/1922453/75405673-33007300-5973-11ea-9821-c39d8db98a9f.png)

**To Reproduce**

1. Disable Jetpack connection
2. Enable the Task List
3. Click on taxes step
4. See the "Connect" button in isBusy state
5. During that time, click on the button and see go to `wp-admin/undefined`

psealock:
Fixed by https://github.com/woocommerce/woocommerce-admin/pull/3786

PR #3786, opened 2020-02-27
OBW Connect: Fix requesting state

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/3785

Fix the Connect step of the stepper such that loading of status is visually displayed by the stepper's number icon and disable the "Connect" button so that it does not lead to bad urls.

### Test

1. Disable Jetpack connection
2. Enable the Task List
3. Click on taxes step
4. See the "Connect" button disabled while connection status is loading
5. See the spinner in the Stepper's number icon

error

### DIFF
--- a/client/dashboard/task-list/tasks/steps/connect.js
+++ b/client/dashboard/task-list/tasks/steps/connect.js
@@ -19,14 +19,14 @@ class Connect extends Component {
 		super( props );
 
 		this.connectJetpack = this.connectJetpack.bind( this );
-		props.setIsStepPending( true );
+		props.setIsPending( true );
 	}
 
 	componentDidUpdate( prevProps ) {
-		const { createNotice, error, isRequesting, setIsStepPending } = this.props;
+		const { createNotice, error, isRequesting, setIsPending } = this.props;
 
 		if ( prevProps.isRequesting && ! isRequesting ) {
-			setIsStepPending( false );
+			setIsPending( false );
 		}
 
 		if ( error && error !== prevProps.error ) {
@@ -100,11 +100,11 @@ Connect.propTypes = {
 	/**
 	 * Control the `isPending` logic of the parent containing the Stepper.
 	 */
-	setIsStepPending: PropTypes.func,
+	setIsPending: PropTypes.func,
 };
 
 Connect.defaultProps = {
-	setIsStepPending: () => {},
+	setIsPending: () => {},
 };
 
 export default compose(
--- a/client/dashboard/task-list/tasks/steps/connect.js
+++ b/client/dashboard/task-list/tasks/steps/connect.js
@@ -100,7 +100,11 @@ Connect.propTypes = {
 	/**
 	 * Control the `isPending` logic of the parent containing the Stepper.
 	 */
-	setIsPending: PropTypes.func.isRequired,
+	setIsPending: PropTypes.func,
+};
+
+Connect.defaultProps = {
+	setIsPending: () => {},
 };
 
 export default compose(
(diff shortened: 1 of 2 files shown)


client/dashboard/task-list/tasks/steps/connect.js
```diff
@@ -47,7 +47,7 @@ class Connect extends Component {
 						{ __( 'Retry', 'woocommerce-admin' ) }
 					</Button>
 				) : (
-					<Button isBusy={ isRequesting } isPrimary onClick={ this.connectJetpack }>
+					<Button disabled={ isRequesting } isPrimary onClick={ this.connectJetpack }>
```
joshuatf: Nice!  This looks and works a lot better.

client/dashboard/task-list/tasks/tax.js
```diff
@@ -369,16 +378,22 @@ class Tax extends Component {
...
 			getActivePlugins,
 			getOptions,
 			getSettings,
 			getSettingsError,
 			isGetSettingsRequesting,
+			isGetJetpackConnectUrlRequesting,
 			isJetpackConnected,
 		} = select( 'wc-api' );
 
+		const jetpackQueryArgs = {
```
joshuatf: This works well but it would be great if we could avoid duplicating logic outside the `Connect` component in case it ever changes.

What do you think about a `setRequesting` prop or `onRequesting` and `onReady` prop for the `Connect` component so that we can set `isPending` to `true|false` inside the tax component?

psealock:
>Let me know if you think that's doable; otherwise this LGTM.

The problem appears to be that`Stepper` renders the icon, but the logic lives in the individual step (at least for `Connect`). I'd say the solution would be to move rendering of the icon into the steps themselves. 

I have run out of time for today, but will think a bit more on this.

joshuatf:
> I'd say the solution would be to move rendering of the icon into the steps themselves.

Rendering of the whole icon or just the pending state?

We've done this previously by passing a `setMethodRequestPending` prop from the component that renders the stepper to the individual steps so that they can control the `isPending` state of the `Stepper`.

Though I could see an argument for moving the state into the actual `Stepper` component and have that pass down a `setPending` method as props to child steps automatically.  Side note- this would be a great opportunity to also pass down `nextStep()` methods where we duplicate much of the logic for every `Stepper` instance.

In either case, I think we'll still need some callback props for the `Connect` component's request state to avoid it becoming too knowledgable of the `Stepper` state.

psealock:
Thanks for the suggestion @joshuatf, this is working well and keeps the pending logic out of `tax.js`. However, it exposes a different problem.

`isPending` is a property used by the `Tax` component to represent any step or request pending. This logic now gets triggered and a redirect happens when it should not.

https://github.com/woocommerce/woocommerce-admin/blob/df27e12d42b10351c06a5c34af2b24e3483a65f2/client/dashboard/task-list/tasks/tax.js#L106-L108

If we go this route, I think a larger refactor is in order. What do you think?

joshuatf:
>isPending is a property used by the Tax component to represent any step or request pending. This logic now gets triggered and a redirect happens when it should not.

That `isPending` check should probably be changed to a `'no' === prevProps.generalSettings.woocommerce_calc_taxes` to see if this setting has just been changed, but it may also need to check to make sure `automatedTaxEnabled` is `false` given https://github.com/woocommerce/woocommerce-admin/pull/3795 was just merged in.

>If we go this route, I think a larger refactor is in order. What do you think?

We can also handle this separately in a larger refactor if you prefer; I know this PR has grown in scope a bit.

psealock:
Thanks for the tips @joshuatf, this is now working as expected and ready for another review.

When `woocommerce_calc_taxes ` option is set to `no` and I click on "No thanks, I'll configure taxes manually" this code gets executed as it should.

https://github.com/woocommerce/woocommerce-admin/blob/acce21e63228422b0f7f074b9e1b006ba9425c00/client/dashboard/task-list/tasks/tax.js#L106-L111

Issue #3780, opened 2020-02-26
[4.0 rc1] Retry button disappears in order to proceed without Mailchimp installation if there is an issue installing Mailchimp


juliaamosova:
**Describe the bug**

Retry button disappears in order to proceed without Mailchimp installation if there is an issue installing Mailchimp. This issue is contingent upon this one: https://github.com/woocommerce/woocommerce-admin/issues/3446. In other words, it only happens if you are not able to proceed with the Mailchimp installation and having errors described in https://github.com/woocommerce/woocommerce-admin/issues/3446. 

**To Reproduce**

Noting here that I could reproduce it with a pre-release of 4.0 rc1 just now on a brand new JN site:

- installed Query Monitor plugin
- installed Debug Bar plugin
- uploaded and activated rc1 zip
- I was selected for the new onboarding by default so didn’t have to change anything in that regard
- Next, I proceeded with new onboarding - selected to install WooCommerce Services and connected the site to WP.com account.
- selected all the paid extensions on the screen about what kind of products I want to sell
- On the FB/Mailchimp step - selected both to be installed and got the following error:

<img width="775" alt="Screen Shot 2020-02-26 at 10 10 05 AM" src="https://user-images.githubusercontent.com/19143190/75361768-ce1b1f00-5885-11ea-8c8b-bfe26e942bff.png">

- `Retry` the button disappeared if I toggled off the Mailchimp installation:

![](http://g.recordit.co/g5RPY905u0.gif)

**Expected behavior**

I expected the button to be there even if I toggled off the Mailchimp installation. Right now it looks like there is no way to proceed. 

**Screenshots**

See above. 

**Desktop (please complete the following information):**

* OS: macOS Catalina 10.15.3
* Browser: Chrome
* Version: 80.0.3987.116 (Official Build) (64-bit)

**Additional context**

N/A

timmyc:
Thanks for logging the issue @juliaamosova - thinking we can force this error state by having the REST API return an error on the mailchimp install test. We definitely want the "Continue without installing" link here to be shown.

psealock:
I can reproduce this locally by forcing an error on the `onboarding/plugins/install` endpoint.

Whats also weird is if I untoggle both facebook and mailchimp, I proceed to the next screen with a success message.

psealock:
Fixed by #3780

PR #3787, opened 2020-02-27
OBW: Fix retry plugin install button disappearing

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/3780

When an API request to install a plugin fails, the button behaviour was off:

* "Retry" button would disappear if you untoggled the failing plugin.
* If you untoggled all plugins, you'd proceed to the next step with a success message.

1. Add `return new \WP_Error( 'plugin Oopsie' );` as the first line in this function to simulate a failed install.
https://github.com/woocommerce/woocommerce-admin/blob/fix/obw-retry-btn/src/API/OnboardingPlugins.php#L165
2. Go to the Business Details step of the OBW.
3. Attempt to install some plugins.
4. When it fails, make sure the button behaviour is as expected.
5. Remove the forced error line and make sure you can install plugins.

### DIFF
--- a/client/dashboard/profile-wizard/steps/business-details.js
+++ b/client/dashboard/profile-wizard/steps/business-details.js
@@ -93,15 +93,15 @@ class BusinessDetails extends Component {
 		await updateProfileItems( updates );
 
 		if ( ! isError ) {
-			if ( 0 === businessExtensions.length ) {
-				goToNextStep();
-				return;
-			}
-
-			this.setState( {
-				installExtensions: true,
-				isInstallingExtensions: true,
-			} );
+			this.setState(
+				{
+					installExtensions: true,
+					isInstallingExtensions: true,
+				},
+				() => {
+					goToNextStep();
+				}
+			);
 		} else {
 			createNotice(
 				'error',
@@ -238,8 +238,7 @@ class BusinessDetails extends Component {
 	}
 
 	renderBusinessExtensions( values, getInputProps ) {
-		const { installExtensions, extensionInstallError } = this.state;
-		const { goToNextStep } = this.props;
+		const { installExtensions } = this.state;
 		const extensionsToInstall = this.getBusinessExtensions( values );
 		const extensionBenefits = [
 			{
@@ -288,10 +287,10 @@ class BusinessDetails extends Component {
 					<div className="woocommerce-profile-wizard__card-actions">
 						<Plugins
 							onComplete={ () => {
-								goToNextStep();
+								this.onContinue( values );
 							} }
 							onSkip={ () => {
-								goToNextStep();
+								this.onContinue( values );
 							} }
 							onError={ () => {
 								this.setState( {
@@ -299,7 +298,6 @@ class BusinessDetails extends Component {
 									isInstallingExtensions: false,
 								} );
 							} }
-							autoInstall={ ! extensionInstallError }
 							pluginSlugs={ extensionsToInstall }
 						/>
 					</div>
--- a/client/dashboard/profile-wizard/steps/business-details.js
+++ b/client/dashboard/profile-wizard/steps/business-details.js
@@ -238,7 +238,7 @@ class BusinessDetails extends Component {
 	}
 
 	renderBusinessExtensions( values, getInputProps ) {
-		const { installExtensions } = this.state;
+		const { installExtensions, extensionInstallError } = this.state;
 		const { goToNextStep } = this.props;
 		const extensionsToInstall = this.getBusinessExtensions( values );
 		const extensionBenefits = [
@@ -299,7 +299,7 @@ class BusinessDetails extends Component {
 									isInstallingExtensions: false,
 								} );
 							} }
-							autoInstall
+							autoInstall={ ! extensionInstallError }
 							pluginSlugs={ extensionsToInstall }
 						/>
 					</div>
--- a/client/dashboard/profile-wizard/steps/business-details.js
+++ b/client/dashboard/profile-wizard/steps/business-details.js
@@ -93,15 +93,15 @@ class BusinessDetails extends Component {
 		await updateProfileItems( updates );
 
 		if ( ! isError ) {
-			this.setState(
-				{
-					installExtensions: true,
-					isInstallingExtensions: true,
-				},
-				() => {
-					goToNextStep();
-				}
-			);
+			if ( 0 === businessExtensions.length ) {
+				goToNextStep();
+				return;
+			}
+
+			this.setState( {
+				installExtensions: true,
+				isInstallingExtensions: true,
+			} );
 		} else {
 			createNotice(
 				'error',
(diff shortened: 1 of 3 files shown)


client/dashboard/task-list/tasks/steps/plugins.js
```diff
@@ -57,7 +57,7 @@ class Plugins extends Component {
 			activatePlugins( pluginSlugs );
 		}
 
-		if ( activatedPlugins.length === pluginSlugs.length ) {
+		if ( pluginSlugs.length > 0 && activatedPlugins.length === pluginSlugs.length ) {
```
joshuatf: Excellent 🎸

client/dashboard/profile-wizard/steps/business-details.js
```diff
@@ -299,7 +299,7 @@ class BusinessDetails extends Component {
 									isInstallingExtensions: false,
 								} );
 							} }
-							autoInstall
+							autoInstall={ ! extensionInstallError }
```
joshuatf: I think this kind of logic could be moved inside the component, but I think if we do this just to avoid returning `null` in the render, we can move up the newly added `Continue` button in the `render()` method of `Plugins` to avoid doing this.

client/dashboard/task-list/tasks/steps/plugins.js
```diff
@@ -108,6 +108,16 @@ class Plugins extends Component {
 			return null;
 		}
 
+		if ( 0 === pluginSlugs.length ) {
```
joshuatf: We can move this above:

```
		if ( autoInstall ) {
			return null;
		}
```

To avoid the `autoInstall={ ! extensionInstallError }` used earlier.

client/dashboard/profile-wizard/steps/business-details.js
psealock: Great suggestion, and it does work. Just to note that it is possible for the button to disappear doing it this way by choosing the another plugin to install:

Screen Shot 2020-02-28 at 4 42 45 PM

In the screen shot above, Mailchimp failed and then I switched the toggles such that FB is now selected. `hasErrors` is false and the length of plugins is greater than one. Granted, this is edge case. What do you think?

joshuatf: Great catch, Paul!  I think it may be an edge case, but still worth addressing.

After diving into this a bit more, I think the real issue here is that this `Plugin` component should not have `autoInstall` since it does not in fact auto-install.  This will require a few changes, but actually simplifies a lot of the logic in this component:

* Remove `autoInstall` prop.
* Add `this.onContinue( values );` to `onComplete` and `onSkip` functions.
* Remove all references to `extensionInstallError` and `isInstallingExtensions` in state.
* Make sure `goToNextStep();` is hit in `onContinue` if no error exists (removing `0 === businessExtensions.length` logic.
* Add continue text and skip text props to `Plugin` component.

This results in a lot more red and tests well on my end.  Feel free to ping me if you have any questions.

psealock: https://github.com/woocommerce/woocommerce-admin/pull/3787/commits/5b9386957a501ac3e729e5f16b0ee71c9d792859 Applies your suggestions. Let me know if I'm missing something, but something is off when I try to install a plugin without forcing a server error.

If I click "Continue", the async `onContinue` function is called, but it gets executed synchronously, which I thought was odd. Sure enough, on `version/1.0` branch this function performs this line synchronously, `await updateProfileItems( updates );`.

https://github.com/woocommerce/woocommerce-admin/blob/5b9386957a501ac3e729e5f16b0ee71c9d792859/client/dashboard/profile-wizard/steps/business-details.js#L93

It seems the old logic, that you suggested I remove is what keeps this function appearing "async" to the user:

>Make sure goToNextStep(); is hit in onContinue if no error exists (removing 0 === businessExtensions.length logic.

`updateProfileItems` is a Fresh Data action that is not async, so `await` does nothing in this case. FWIW, the [settings data store refactor](https://github.com/woocommerce/woocommerce-admin/pull/3313 ) exports actions that are defined as aync.

Am I looking at this and implemented your suggestion correctly here?

joshuatf: >updateProfileItems is a Fresh Data action that is not async, so await does nothing in this case. FWIW, the settings data store refactor exports actions that are defined as aync.

Unfortunately we have a bunch of these that are incorrectly set up (see https://github.com/woocommerce/woocommerce-admin/issues/2959).  Async functions will make this process immensely easier.


>If I click "Continue", the async onContinue function is called, but it gets executed synchronously, which I thought was odd. Sure enough, on version/1.0 branch this function performs this line synchronously:

Hmm, I am seeing an issue, but I'm seeing it performed async with these changes which goes to the next step before the plugins have a chance to be fully installed and activated.  

The `goToNextStep()`should probably be called after we receive both an `onComplete` from the plugins and profile items have finished updating inside `componentDidUpdate()`.  I think this is a worthwhile refactor to this step, but if you want to use the previous `autoInstall` option, I think that's an okay temporary solution if we want to open an issue to update this later.

timmyc:
@psealock do you need more 👀 here?

client/dashboard/profile-wizard/steps/business-details.js
psealock: >Unfortunately we have a bunch of these that are incorrectly set up (see #2959). Async functions will make this process immensely easier.

Oh yeah, I see that now.

>Hmm, I am seeing an issue, but I'm seeing it performed async with these changes which goes to the next step before the plugins have a chance to be fully installed and activated.

Cool, I think we are on the same page on this one. I confuse myself with sync/async.

>The goToNextStep()should probably be called after we receive both an onComplete from the plugins and profile items have finished updating inside componentDidUpdate(). I think this is a worthwhile refactor to this step, but if you want to use the previous autoInstall option, I think that's an okay temporary solution if we want to open an issue to update this later.

I agree that this is a worthwhile refactor. The problem is that `onComplete` is called when the user clicks "Continue", so `BusinessDetails` would need a `componentDidUpdate` if statement to check if profileItems are done requesting (this happens twice), if plugin instillation is done requesting (would need to create a prop on `Plugins` a la `onSuccess`), and possibly if plugin activation is done requesting before calling `goToNextStep`.

With our data store refactors on the horizon, my preference is to keep the `autoInstall` check in this PR because true async functions will clean up the logic so much and we will be refactoring this soon anyways.

if that sounds good to you, I'll revert that last commit.

psealock:
Thanks @timmyc, see https://github.com/woocommerce/woocommerce-admin/pull/3787#discussion_r387359136.

client/dashboard/profile-wizard/steps/business-details.js
timmyc: > if that sounds good to you, I'll revert that last commit.

Ideally I'd like this to get wrapped up and landed in `version/1.0` since it is a reproducible bug that folks are likely hitting in production right now.

psealock: >Ideally I'd like this to get wrapped up and landed in version/1.0 since it is a reproducible bug that folks are likely hitting in production right now.

This is back to its original state that @joshuatf tested. I'll leave it for one last look, but feel free to merge if I'm not online.

PR #3796, opened 2020-02-28
Improve focus on task list

### LINKED ISSUE #3711: Onboarding - task list: improve focus on the task list and allow an option to skip it
Currently there are a lot of options in the WooCommerce menu that can distract the user and derail him from completing the most important steps to have the store ready. In order to improve the focus on the task list we should:

- Hide all the options in the WC menu and just display the setup option, if the user selects `false` as the value of the event prop `setup_client` in `wcadmin_storeprofiler_store_details_continue`
- If the user has `true` as a value or never triggered that even we'll show the menu as always.
- When the user completes the task list we unlock all the options on the menu.
- We include two links at the bottom of the task list: 
  - One allowing the user to skip the task list, unlocking the menu and hiding the task list.
  - Another one allowing the user to complete the task list later, unlocking the menu and keeping the task list on the menu.

- We need to add tracking to the links, probably one event with an even prop to distinguish them: 
Name: `wcadmin_tasklist_skip`
Event props: completed_tasks: number of tasks completed, completed_tasks_names: name of the tasks that were completed, skip_option: "i'll do this later" or "skip store setup" 

Mockup [here](https://www.figma.com/file/nzAjPR4nqdJTDxH32PT4Qg/On-boarding?node-id=15%3A2338)
[P2](p6riRB-5li-p2)

becdetat:
Fixes #3711

This hides WooCommerce menu items while the onboarding task list is being displayed. It also includes the addition of two buttons (styled as links) at the bottom of the task list to have more control over when the task list is displayed, by skipping the store setup (which hides the task list, and hiding WooCommerce menu items) or by doing the store setup later (which still shows the task list along with the rest of the dashboard, and shows the WooCommerce menu items. 

### Accessibility

The action buttons at the bottom of the task list are semantically buttons as they don't have a href and their only interaction is via an `onClick` handler, so I created them as a new `LinkButton` component, which wraps a `button` styled like a link.

- Have yourself a store that has an outstanding task. The easiest way to do this is to go into Settings -> Shipping zone and delete all of the shipping zones. Set this up before changing to the PR branch (because otherwise you might not be able to get to the Settings area via the WooCommerce menu).
- Go to the WooCommerce dashboard. The menu should only display the Setup option:
image
- Clicking 'Do this later' should:
    - continue to show the task list
    - show the rest of the dashboard
    - show all of the menu items under WooCommerce
- Clicking 'Skip store setup' should:
    - hide the task list
    - show the rest of the dashboard
    - show all of the menu items under WooCommerce

To reset state, in phpMyAdmin delete any rows containing 'skip_store_setup' or 'do_this_later'.

### Changelog Note:

Enhancement: Improve focus on task list


client/dashboard/task-list/index.js
```diff
@@ -256,6 +256,58 @@ class TaskDashboard extends Component {
...
+		this.props.updateOptions( {
+			woocommerce_task_list_do_this_later: true,
+		} );
+	};
+
+	renderSkipActions() {
+		const { skipStoreSetup, doThisLater } = this.props;
+
+		return (
+			<div style={ { textAlign: 'center' } }>
```
joshuatf: We can move this into a scss file.  I think there are also some color differences between the design:

**Current**
Screen Shot 2020-02-28 at 1 58 33 PM


**Design**
Screen Shot 2020-02-28 at 2 00 21 PM

client/dashboard/task-list/index.js
```diff
@@ -256,6 +256,58 @@ class TaskDashboard extends Component {
 		);
 	}
 
+	onSkipStoreSetup = () => {
+		const completedTasksNames = this.getTasks()
+			.filter( x => x.completed )
+			.map( x => x.title );
+
+		recordEvent( 'wcadmin_tasklist_skip', {
+			completed_tasks: completedTasksNames.length,
```
joshuatf: I think to keep things a bit more semantic we can rename this to `completed_tasks_count` and `complete_tasks_names` to `complete_tasks` if you don't have any issue with it @pmcpinto.

client/dashboard/task-list/index.js
```diff
@@ -256,6 +256,58 @@ class TaskDashboard extends Component {
 		);
 	}
 
+	onSkipStoreSetup = () => {
+		const completedTasksNames = this.getTasks()
+			.filter( x => x.completed )
+			.map( x => x.title );
```
joshuatf: Let's use the `key` here in case the copy for titles changes in the future.

client/dashboard/task-list/index.js
```diff
@@ -256,6 +256,58 @@ class TaskDashboard extends Component {
...
+
+		recordEvent( 'wcadmin_tasklist_skip', {
+			completed_tasks: completedTasksNames.length,
+			completed_tasks_names: completedTasksNames,
+			skip_option: 'skip store setup'
```
joshuatf: Similarly, can we change this to `reason` and have the options be `skip|later` to make things a bit more easy to grok? /cc @pmcpinto

client/dashboard/customizable.js
```diff
@@ -344,6 +365,32 @@ export default compose(
...
 				) === 'yes';
+
+			const options = getOptions( [
+				'woocommerce_task_list_skip_store_setup',
+				'woocommerce_task_list_do_this_later',
+			] );
+
+			withSelectData.skipStoreSetup = get(
+				options,
+				[ 'woocommerce_task_list_skip_store_setup' ],
```
joshuatf: Do you think we could just use `woocommerce_task_list_hidden` to avoid introducing a new option and new logic to other areas of onboarding?

client/onboarding-experience.js
```diff
@@ -0,0 +1,109 @@
...
+				taskListHidden ||
+				taskListCompleted ||
+				skipStoreSetup ||
+				doThisLater ||
+				optionsRequesting ||
+				profileItemsRequesting,
+		};
+	} )
+)( ( { showMenu } ) => {
+	hideOrShowMenuItemsForTaskList( showMenu );
```
joshuatf: Should we just return early if the `showMenu` is true?  Or are there other things this script should do even if the menu is shown?

client/dashboard/task-list/index.js
```diff
@@ -256,6 +256,58 @@ class TaskDashboard extends Component {
...
+			woocommerce_task_list_do_this_later: true,
+		} );
+	};
+
+	renderSkipActions() {
+		const { skipStoreSetup, doThisLater } = this.props;
+
+		return (
+			<div style={ { textAlign: 'center' } }>
+				{ ! skipStoreSetup && (
```
joshuatf: I think we can simplify this logic as the task list won't be shown at all if they skip, right?

packages/components/src/link-button/index.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * External dependencies
+ */
+import PropTypes from 'prop-types';
+
+const LinkButton = ( { onClick, children } ) => {
```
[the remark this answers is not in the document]
becdetat: Nice, done. I did look for an existing `Button` but I must have missed that one. Thanks.

webpack.config.js
```diff
@@ -88,6 +88,7 @@ const webpackConfig = {
 		app: './client/index.js',
 		ie: './client/ie.scss',
 		...entryPoints,
+		'onboarding-experience': './client/onboarding-experience.js',
```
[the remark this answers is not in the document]
becdetat: Nice, I didn't see this. This is now done.

src/Loader.php
```diff
@@ -404,12 +404,22 @@ public static function register_scripts() {
...
+			self::get_file_version( 'onboarding-experience/index.js' ),
+			true
+		);
 	}
 
 	/**
 	 * Loads the required scripts on the correct pages.
 	 */
 	public static function load_scripts() {
+		wp_enqueue_script( 'wc-onboarding-experience' );
```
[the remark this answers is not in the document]
becdetat: done

client/onboarding-experience.js
```diff
@@ -0,0 +1,109 @@
...
+
+	submenuItemsToHide.forEach( ( x ) => {
+		x.hidden = ! show;
+	} );
+
+	const dashboardLi = allSubmenuItems.filter( ( x ) =>
+		x.classList.contains( 'wp-first-item' )
+	)[ 0 ];
+
+	dashboardLi.children[ 0 ].innerHTML = show ? 'Dashboard' : 'Setup';
```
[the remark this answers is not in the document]
becdetat: Nice catch, thanks :-)

client/onboarding-experience.js
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * Internal dependencies
+ */
+import withSelect from 'wc-api/with-select';
+import { getAllTasks } from './dashboard/task-list/tasks';
+import { isOnboardingEnabled } from 'dashboard/utils';
+
+function hideOrShowMenuItemsForTaskList( show ) {
+	const topLevelPageWooCommerceLi = document.getElementById(
```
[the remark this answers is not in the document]
becdetat: Good idea, this is done

client/dashboard/task-list/index.js
becdetat: Done

client/onboarding-experience.js
becdetat: This isn't just run once, it depends on the option which changes when the skip/later buttons are clicked, so it needs to handle both showing and hiding the menu items, and so it can't just return early otherwise the menu will never get shown without a page refresh

client/dashboard/customizable.js
becdetat: Nice, done

client/dashboard/customizable.js
```diff
@@ -225,14 +225,28 @@ class CustomizableDashboard extends Component {
...
+			taskListHidden,
+			taskListCompleted,
+			skipStoreSetup,
+			doThisLater,
+			requesting,
+		} = this.props;
 		const { sections } = this.state;
 
+		if ( requesting ) {
+			return <Spinner />;
```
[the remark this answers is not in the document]
becdetat: Nice, however I've tried adding `woocommerce_task_list_do_this_later` to the preload and it doesn't seem to make any difference to the load time. That's the only extra option I can see that is required so I'm at a loss. Can you have a look at this for me?

client/dashboard/task-list/index.js
becdetat: Done

client/dashboard/customizable.js
```diff
@@ -252,17 +264,19 @@ class CustomizableDashboard extends Component {
 			primaryDate,
 			secondaryDate,
 		};
+		const showTaskList =
```
joshuatf: We have similar logic above to check if the task list is enabled.  Maybe we can move these into a reusable method of `isTaskListEnabled()`.

Also, I think we can remove the check for `wcSettings.onboarding` here.

client/dashboard/customizable.js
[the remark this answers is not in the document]
joshuatf: In the meantime, I think we should remove `<Spinner />` from here.

client/wp-admin-scripts/onboarding-menu-experience/index.js
```diff
@@ -0,0 +1,115 @@
...
+import withSelect from 'wc-api/with-select';
+import { getAllTasks } from '../../dashboard/task-list/tasks';
+import { isOnboardingEnabled } from 'dashboard/utils';
+
+// Build a setup anchor
+const setupAnchor = document.createElement( 'a' );
+setupAnchor.href = '/wp-admin/admin.php?page=wc-admin';
+setupAnchor.classList.add( 'wp-first-item' );
+setupAnchor.setAttribute( 'aria-current', 'page' );
+setupAnchor.innerHTML = __( 'Setup' );
```
joshuatf: Missing a text-domain here.

client/dashboard/task-list/style.scss
```diff
@@ -36,6 +36,16 @@
 				color: $studio-white;
 			}
 		}
+
+		.skip-actions {
+			text-align: center;
+
+			button.components-button.is-secondary {
+				color: #646970;
```
joshuatf: Instead of using the hex here, we can probably use the `$studio-gray-xx` color here.

I also think the margin between the buttons and task list looks a bit different than the design, but having trouble checking on both of these at the moment.

Asked @jameskoster for the Figma design link and will post back here once I can check the design.

client/dashboard/task-list/index.js
```diff
@@ -256,6 +256,64 @@ class TaskDashboard extends Component {
...
+					{ __( 'Skip store setup', 'woocommerce-admin' ) }
+				</Button>
+				{ ! doThisLater && ' | ' }
+				{ ! doThisLater && (
+					<Button
+						isLink={ true }
+						className="is-secondary"
+						onClick={ this.onDoThisLater }
+					>
+						{ __( "I'll do this later", 'woocommerce-admin' ) }
```
joshuatf: After this button is clicked, the menu opens up but if you're not paying close attention it may seem like nothing happens since the page does not change.  I'm wondering if we want to show any kind of notice on click?  /cc @pmcpinto @jameskoster

client/wp-admin-scripts/onboarding-menu-experience/index.js
```diff
@@ -0,0 +1,115 @@
...
+		} );
+		const visibleTasks = filter( tasks, ( x ) => x.visible );
+		const completedTasks = filter(
+			tasks,
+			( x ) => x.visible && x.completed
+		);
+		const taskListCompleted = visibleTasks.length === completedTasks.length;
+
+		return {
+			showMenu:
```
joshuatf: Maybe we should make the early method `isTaskListEnabled()` a client-side util that we can use here as well.

pmcpinto:
> After this button is clicked, the menu opens up but if you're not paying close attention it may seem like nothing happens since the page does not change. I'm wondering if we want to show any kind of notice on click? /cc @pmcpinto @jameskoster

@joshuatf good point. I think that if the user clicks "I'll do this later" we should redirect him to the wc-admin dashboard. @jameskoster any thoughts?

joshuatf:
@pmcpinto They're technically on the dashboard when this happens (the default content is hidden by the task list).  Do you want to open up the charts beneath the task list when they click "I'll do this later?"

client/dashboard/customizable.js
becdetat: I've removed the `wcSettings.onBoarding` check but I think the logic differs enough that extracting it out into another function is premature, otherwise there's going to be some very funky logic happening in the extracted function.

client/dashboard/task-list/style.scss
becdetat: I've changed to the SASS variable, thanks for that.

I can't see a difference between the site and the design, at least based on https://teamxthings.files.wordpress.com/2020/02/dashboard-local-site.png

Both have a margin of roughly 40px. How precise should we be matching designs? Is there a design or style guide for this kind of thing?

client/dashboard/task-list/index.js
becdetat: When you click that button the rest of the page does load in, but I agree that it isn't very obvious.

client/wp-admin-scripts/onboarding-menu-experience/index.js
becdetat: I still don't think there's enough shared logic to justify extracting anything out and potentially making things harder to grok (having to go through layers of function calls).

becdetat:
@joshuatf 
> @pmcpinto They're technically on the dashboard when this happens (the default content is hidden by the task list). Do you want to open up the charts beneath the task list when they click "I'll do this later?"

It already does open up the charts when "I'll do this later" is clicked, it just takes time for them to appear.

jameskoster:
As long as the charts appear I think it's fine.

pmcpinto:
> As long as the charts appear I think it's fine.
+1

> Similarly, can we change this to reason and have the options be skip|later to make things a bit more easy to grok? /cc @pmcpinto

@joshuatf, sorry I missed this ping. Sounds good to me

client/dashboard/task-list/style.scss
joshuatf: We're usually pretty accurate down to the pixel.  The Figma designs help a lot with that and also provide some sass variable names which saves some time trying to match things up.

I sent you an invite to your a8c email in case you haven't been invited to Figma yet.  Let me know if you don't receive or having trouble logging in.

Since this doesn't really add any specific margins and also eye-balling it looks pretty close, I think we can keep it as is and update later if needed.

client/dashboard/task-list/index.js
```diff
@@ -275,6 +275,64 @@ class TaskDashboard extends Component {
...
+		} );
+	};
+
+	renderSkipActions() {
+		const { doThisLater } = this.props;
+
+		return (
+			<div className="skip-actions">
+				<Button
+					isLink={ true }
```
joshuatf: Optional: this prop doesn't need the `={ true }` bit.  Same for the button below.

client/dashboard/task-list/style.scss
becdetat: Thanks @joshuatf

PR #3798, opened 2020-03-01
Make WooCommerce breadcrumbs use WooCommerce branding if it is installed

### LINKED ISSUE #3246: WooCommerce Admin does not respect WooCommerce Branding

becdetat:
Fixes #3246

This looks for the WooCommerce extension, and if it's installed use the configured store name instead of "WooCommerce".

### Screenshots

image

By "breadcrumbs" I'm talking about the breadcrumb links at the top of the shop dashboard for example.

1. Without the WooCommerce Branding extension installed, check that the breadcrumbs use "WooCommerce" 
2. Install and activate the WooCommerce Branding extension, but don't configure it. Check that the breadcrumbs still use "WooCommerce".
3. Configure the WooCommerce Branding extension, giving the store a name. Check that the breadcrumbs change to the configured store name.
4. Deactivate the WooCommerce Branding extension. Check that the breadcrumbs change back to "WooCommerce".

### Changelog Note:

Fix: Make WooCommerce breadcrumbs use WooCommerce Branding if it is installed

### DIFF
--- a/includes/connect-existing-pages.php
+++ b/includes/connect-existing-pages.php
@@ -54,15 +54,23 @@ function wc_admin_get_core_pages_to_connect() {
  * @return array Filtered breadcrumb pieces.
  */
 function wc_admin_filter_core_page_breadcrumbs( $breadcrumbs ) {
-	$screen_id        = PageController::get_instance()->get_current_screen_id();
-	$pages_to_connect = wc_admin_get_core_pages_to_connect();
+	$screen_id              = PageController::get_instance()->get_current_screen_id();
+	$pages_to_connect       = wc_admin_get_core_pages_to_connect();
+	$woocommerce_breadcrumb = array(
+		add_query_arg( 'page', 'wc-admin', 'admin.php' ),
+		__( 'WooCommerce', 'woocommerce-admin' ),
+	);
 
 	foreach ( $pages_to_connect as $page_id => $page_data ) {
 		if ( preg_match( "/^woocommerce_page_{$page_id}\-/", $screen_id ) ) {
 			if ( empty( $page_data['tabs'] ) ) {
-				$new_breadcrumbs = array( $page_data['title'] );
+				$new_breadcrumbs = array(
+					$woocommerce_breadcrumb,
+					$page_data['title'],
+				);
 			} else {
 				$new_breadcrumbs = array(
+					$woocommerce_breadcrumb,
 					array(
 						add_query_arg( 'page', $page_id, 'admin.php' ),
 						$page_data['title'],
--- a/includes/connect-existing-pages.php
+++ b/includes/connect-existing-pages.php
@@ -57,7 +57,7 @@ function wc_admin_filter_core_page_breadcrumbs( $breadcrumbs ) {
 	$screen_id              = PageController::get_instance()->get_current_screen_id();
 	$pages_to_connect       = wc_admin_get_core_pages_to_connect();
 	$woocommerce_breadcrumb = array(
-		admin_url( 'admin.php?page=wc-admin' ),
+		'admin.php?page=wc-admin',
 		__( 'WooCommerce', 'woocommerce-admin' ),
 	);
 
--- a/includes/connect-existing-pages.php
+++ b/includes/connect-existing-pages.php
@@ -57,7 +57,7 @@ function wc_admin_filter_core_page_breadcrumbs( $breadcrumbs ) {
 	$screen_id              = PageController::get_instance()->get_current_screen_id();
 	$pages_to_connect       = wc_admin_get_core_pages_to_connect();
 	$woocommerce_breadcrumb = array(
-		add_query_arg( 'page', 'wc-admin', 'admin.php' ),
+		admin_url( 'admin.php?page=wc-admin' ),
 		__( 'WooCommerce', 'woocommerce-admin' ),
 	);
 
--- a/src/PageController.php
+++ b/src/PageController.php
@@ -176,6 +176,10 @@ class PageController {
 			}
 		}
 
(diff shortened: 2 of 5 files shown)


timmyc:
@becdetat nice work here! So I think the implementation here might be a bit too focused on sniffing out if the Branding plugin is installed. I pinged @rrennick on this since he added the issue to the sprint cycle, and it sounds like the Branding plugin modifies the translated string name for "WooCommerce" - which is used in registering the [main menu item here in woo core](https://github.com/woocommerce/woocommerce/blob/a7d57a248ed03dac7bad119e71f83dd961f43cb3/includes/admin/class-wc-admin-menus.php#L62).

So it sounds like, and correct me if I'm wrong Ron, but we should be able to pass along that translation perhaps in our `wcSettings` object [#](https://github.com/woocommerce/woocommerce/blob/a7d57a248ed03dac7bad119e71f83dd961f43cb3/includes/admin/class-wc-admin-menus.php#L62) and then use that translated string in the breadcrumbs like you have done here.

Apologies that the issue wasn't explicit about that. We have tried to keep extension-specific logic out of the code base and to focus on just supporting extensibility options instead.

rrennick:
> but we should be able to pass along that translation perhaps in our wcSettings object # and then use that translated string in the breadcrumbs like you have done here.

All of the breadcrumb is filterable through https://github.com/woocommerce/woocommerce-admin/blob/1f261998e7287b77bc13c3d4ee2e84b717da7957/src/PageController.php#L179:L185 except for `WooCommerce`. For consistency, it should be added to the breadcrumb array in PHP and removed from the Header component.

becdetat:
I noticed this:

timmyc:
I think that particular string is [located here](https://github.com/woocommerce/woocommerce/blob/5ea1b72b6a59cf4b48826763ef92496370d1ed8c/includes/admin/helper/views/html-section-nav.php#L18) in woo core. So looks like it is just using the translated menu item... so while it looks odd, seems like it isn't a bug... more an artifact of how the branding extension works.

becdetat:
Ok this is ready for another review - I've moved the WooCommerce breadcrumb to the server except for where the breadcrumbs are generated on the server. The branding plugin's string replacement only works on the server - `__('WooCommerce', 'woocommerce-admin)` is `'WooCommerce'` on the client - so I've also passed the translated string from the server to the client via `wcSettings`.

WooCommerce Branding also seems to have a bug where the breadcrumbs are dropped on some pages such as `WooCommerce | Reports` - in https://github.com/woocommerce/woocommerce-admin/blob/12dfc3bee6ea908ac90bbbb640892618b53c3077/src/Loader.php#L485 `is_embed_page()` returns true for those pages when the branding extension is enabled, so `embed_page_header()` returns early and doesn't render the breadcrumbs. This is present on a fresh install without these changes so I don't think it affects this PR.

becdetat:
@timmyc I can't reproduce this - it was doing this during development but I fixed the issue, so possibly you weren't looking at the tip of the branch?

includes/connect-existing-pages.php
```diff
@@ -54,15 +54,23 @@ function wc_admin_get_core_pages_to_connect() {
  * @return array Filtered breadcrumb pieces.
  */
 function wc_admin_filter_core_page_breadcrumbs( $breadcrumbs ) {
-	$screen_id        = PageController::get_instance()->get_current_screen_id();
-	$pages_to_connect = wc_admin_get_core_pages_to_connect();
+	$screen_id              = PageController::get_instance()->get_current_screen_id();
+	$pages_to_connect       = wc_admin_get_core_pages_to_connect();
+	$woocommerce_breadcrumb = array(
+		add_query_arg( 'page', 'wc-admin', 'admin.php' ),
```
rrennick:
```suggestion
		admin_url( 'admin.php?page=wc-admin' ),
```

src/PageController.php
```diff
@@ -176,6 +176,10 @@ public function get_breadcrumbs() {
 			}
 		}
 
+		$woocommerce_breadcrumb = array( 'admin.php?page=wc-admin', __( 'WooCommerce', 'woocommerce-admin' ) );
```
rrennick:
```suggestion
		$woocommerce_breadcrumb = array( admin_url( 'admin.php?page=wc-admin' ), __( 'WooCommerce', 'woocommerce-admin' ) );
```

becdetat:
Thanks @rrennick, I'll try to keep that in mind :-)

becdetat:
@rrennick I can't reproduce the issue with the analytics pages, can you take a screenshot?

timmyc:
@rrennick could you give this another quick 👀 to update your review?

timmyc:
oh good catch there Paul

rrennick:
> @rrennick could you give this another quick 👀 to update your review?

Looks okay for the issue I saw.

becdetat:
Ok this is looking much better now, I've addressed what @psealock noted:

- Show WooCommerce breadcrumb on analytics pages
- Stop double appending the admin URL to the root WooCommerce breadcrumbs

PR #3800, opened 2020-03-02
Changelog: Add Script

psealock:
This PR replaces the Github changelog script with the same [one used by blocks](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/bin/changelog/README.md) to generate a changelog based on Zenhub.

### Test
You'll need Github and Zenhub API keys in order to generate a changelog.

1. `npm install` to get the added dependencies.
1. `npm run changelog`.
2. Cycle 2's id is `5e417ad8bbef8fcd55922791` for quick reference.


psealock:
>Did I do something wrong?

Looks like you don't have the zenhub or github api keys as environment variables.

>it would be great to get some of these utilities in another repo

Most definitely, you can follow along @senadir's efforts here: https://github.com/WordPress/gutenberg/pull/19866

bin/changelog/common/get-entry.js
```diff
@@ -0,0 +1,89 @@
...
+			.trim();
+		if ( ! title.length ) {
+			title = `${ type }: ${ pullRequest.title }`;
+		} else {
+			title = `${ type }: ${ title }`;
+		}
+	} else {
+		title = `${ type }: ${ pullRequest.title }`;
+	}
+	return `- ${ title } [#${ pullRequest.number }](${ pullRequest.url })${ devNote } ${ authorTag }`;
```
joshuatf: Just curious - it seems like this format is a little different than our current in `readme.txt`.  Will all new entries going forward use this format instead?

PR #3807, opened 2020-03-03
Downloads: add docs

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/3366

Add Downloads Report documentation.

### Test

Check for typos or missing info.


psealock:
Thanks for the review, I'll just add a note about timestamps being in UTC

PR #3812, opened 2020-03-03
Onboarding - business step: add more options in the competitors list and other fixes

### LINKED ISSUE #3658: Onboarding - business step: add more options in the competitors list and other fixes
- We should add the following additional options when the user selects that he's selling on another platform:
Amazon
eBay
Etsy
Squarespace
They should be displayed before the "Other" option

- If the user selects "Other" we should display an option input field, in order to allow him to name the platform that he's using.

- In the first question of this step, we have significant % of users saying that they plan to sell +1000 products, so I'm assuming that they're thinking in terms of the number of units. We could try to make the question more clear, perhaps: How many products do you plan to display?

@jameskoster any idea?

octaedro:
Fixes #3658

Onboarding - Business Details step:
- Added the platforms: `Amazon`, `eBay`, `Etsy`, `Squarespace`, when the user selects that he's selling on another platform.
- Changed the copy: `How many products do you plan to sell?` to `How many products do you plan to display?`
- Added new param to `recorderEvent`
- Added a new input to save the platform's name, when the user selects `other`.

### Screenshots
screenshot-onb-business1
screenshot-onb-business2

- Go to the Onboarding's Business Details step (URL `/wp-admin/admin.php?page=wc-admin&step=business-details`)
- Check that the label `How many products do you plan to display?` is visible.

- Press the `Which platform is the store using?` dropdown menu. Check that the new options are available.
- Select the option: `Other`.
- An input should appear. Add a text inside.
- The text should be saved after `Continue` is pressed.


client/dashboard/profile-wizard/steps/business-details.js
```diff
@@ -89,12 +91,14 @@ class BusinessDetails extends Component {
...
 			revenue,
 			used_platform: otherPlatform,
+			used_platform_name: otherPlatformName,
 			install_facebook: values[ 'facebook-for-woocommerce' ],
 			install_mailchimp: values[ 'mailchimp-for-woocommerce' ],
 		} );
 
 		const _updates = {
 			other_platform: otherPlatform,
+			other_platform_name: otherPlatformName,
```
joshuatf: Maybe this should set the other platform name to an empty string here instead of the added logic below.

```suggestion
			other_platform_name: otherPlatform === 'other' ? otherPlatformName : '',
```

client/dashboard/profile-wizard/steps/business-details.js
```diff
@@ -105,7 +109,13 @@ class BusinessDetails extends Component {
 		const updates = {};
 		Object.keys( _updates ).forEach( ( key ) => {
 			if ( _updates[ key ] !== '' ) {
-				updates[ key ] = _updates[ key ];
+				// "other_platform_name" value is saved only when "other" platform is selected
```
joshuatf: Referring to this section which can be removed with the above addition.

client/dashboard/profile-wizard/steps/business-details.js
```diff
@@ -513,6 +543,7 @@ class BusinessDetails extends Component {
 				{ ( { getInputProps, handleSubmit, values, isValidForm } ) => {
 					// Show extensions when the currently selling elsewhere checkbox has been answered.
 					const showExtensions = values.selling_venues !== '';
+					const otherPlatform = getInputProps( 'other_platform' );
```
joshuatf: Was this created for a specific reason?  Can we move this directly inside the `SelectControl ` props below?

`{ ...getInputProps( 'other_platform' ) }`

client/dashboard/profile-wizard/steps/business-details.js
octaedro: You are totally right! I will change it.
Thank you :slightly_smiling_face:

client/dashboard/profile-wizard/steps/business-details.js
octaedro: Yes. The constant `otherPlatform` is used to check if the user selected the option `other` to show the input text or not. The reason to declare this constant is to not call the function `getInputProps( 'other_platform' )` twice.

joshuatf: Ah! I see.  In that case we may want to check using `values` and then we can remove this constant.  Will add a comment below.

octaedro: I didn't see it. Good idea!

octaedro:
Added suggested improvements from the code review.

PR #3823, opened 2020-03-04
add npm run test:zip command

rrennick:
This PR adds `npm run test:zip` to create a test zip in the root directory for easy uploading to a test install. The zip command filters out all the files in `.distignore` and dev packages in the `vendor` folder.

- `npm run test:zip`
- Upload to a test site & activate
- Check that the screens work correctly


.zipignore
```diff
@@ -0,0 +1,47 @@
...
+vendor/phpunit/*
+vendor/sebastian/*
+vendor/squizlabs/*
+vendor/theseer/*
+vendor/webmozart/*
+vendor/woocommerce/*
+vendor/wp-coding-standards/*
+*.sql
+*.tar.gz
+*.zip
```
psealock: Isn't this just a copy of `.distignore`? The release builds are explicit in what they include.

https://github.com/woocommerce/woocommerce-admin/blob/c8eb1e44c820192dafbf2e0361f9fabfe591affb/bin/github-deploy.sh#L121-L132

Do you think it would be good to reuse that logic?

rrennick: > Isn't this just a copy of .distignore?

I started with .distignore. The zip command needs the `/\*` on the end of folder names to exclude folders.

rrennick: @psealock We could move that zip command to a separate .sh and call it from both. Sound good?

rrennick: @psealock Thanks for the suggestion. The PR has been updated to re-use the deploy zip command. I added the exclusion list to avoid running `composer install --no-dev` to remove the dev packages before `zip`ping.

bin/make-zip.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+#
+# Build a installable plugin zip
+
+output 2 "Creating archive... 🎁"
```
psealock:
```
./bin/make-zip.sh: line 5: output: command not found
```
Make sure to include this function in the script.

```
# Output colorized strings
#
# Color codes:
# 0 - black
# 1 - red
# 2 - green
# 3 - yellow
# 4 - blue
# 5 - magenta
# 6 - cian
# 7 - white
output() {
  echo "$(tput setaf "$1")$2$(tput sgr0)"
}
```

timmyc:
This is super awesome @rrennick - guessing this won't work on non-*nix hosts... though guessing a few of our build tools might not. I don't think that should hold this up but just something for us to consider.

bin/make-zip.sh
```diff
@@ -0,0 +1,34 @@
...
+
+build_files=$(ls dist/*/*.{js,css})
+ZIP_FILE=$1
+
+zip -r ${ZIP_FILE} \
```
joshuatf: It might be nice to zip everything while excluding folders and files that should not be included instead of including all the files needed.

I imagine this may be a source of hiccups in testing if a new root file or vendor is added.

bin/make-zip.sh
rrennick: @psealock Nice catch. Added in `74b5113`

bin/make-zip.sh
rrennick: @joshuatf Added in `886e209`.

PR #3827, opened 2020-03-04
Activity Panels: Remove W Panel

### LINKED ISSUE #3822: Hidden WordPress notices
**Describe the bug**

I'm using WP `admin_notices` notices and now, after an upgrade to WC 4.0, they have stopped showing up anywhere in the WooCommerce admin (including the products list and the product page in admin).
The effect is that every plugin using the admin_notices to communicate with the users is now forcibly silenced.

Example: 
1. Let's say the plugin adds some functionality to the WooCommerce but you have to enter some API credentials to use it properly.
2. Info about the missing credentials normally would be accessible to a user trough the visible notice.
3. A user thinks everything is fine as s/he does not see any problems. Trying to use the functionality causes errors and even some of them are not shown as normally it would be, again, shown as a notice.

**To Reproduce**

1. Add a simple notice using https://developer.wordpress.org/reference/hooks/admin_notices/ code.
2. Go to any WooCommerce linked page.
3.  A notice is not visible and the message is lost forever for the user.

**Expected behavior**

Notices should be visible or at least the message provided by the notice is somehow readable to the user in the admin panel.

* Device: All.

**Additional context**

It breaks communication extension→user in the entire admin panel. 
I see that https://github.com/woocommerce/woocommerce-admin/issues/35 exists but "Handle Admin Notices" should not be "Hide Others Admin Notices".

timmyc:
Fixes #3822

See p1583323270149500-slack-wc-admin for background discussion

From feedback ( internally and externally ) it seems the WordPress Notif menu, and the hiding of admin alerts could create some friction for both developers and end users alike. As such, this PR aims to remove the WordPress notif menu entirely, and have WP admin alerts display normally on all legacy core WooCommerce pages.

Note that this PR still does not display admin alerts on the new Analytics pages. At some point I feel we will need to revisit this, but I based this decision upon how Jetpack does a similar approach on their own internal pages.

__Order Page With Notices__
image

__wc-admin page with notices__
image

### Accessibility
There was some logic removed here around tabbing through the activity panels. So might be good for us to test out that all is still working well there.

It is best if you have some admin alerts present in your test site. If you need some, [our example extension is pretty nice for that](https://github.com/woocommerce/woocommerce-admin/blob/master/docs/examples/extensions/important-admin-notice/woocommerce-admin-important-admin-notice-example.php). To force display of a Store Alert, I updated the db entry for the historical import notice to be "unactioned" to have it reappear:

image

- Test alert display on legacy woo pages ( i.e. Orders, Products etc ). Verify admin alerts and Store Alerts display as expected
- On wc-admin pages ( dash, reports ) verify that only Store Alerts are displayed.

### Changelog Note:

Fix: Remove WordPress alert activity panel.


rrennick:
@timmyc This is working fine for me locally except I don't have the W icon to show notices on analytics pages.

timmyc:
> This is working fine for me locally except I don't have the W icon to show notices on analytics pages.

Indeed currently in this approach we just don't show them at all, only StoreAlerts are shown.

rrennick (approved):
> Indeed currently in this approach we just don't show them at all, only StoreAlerts are shown.

@timmyc In that case this is good to go. Nice work. :shipit:

rrennick:
@pmcpinto Can you confirm this change is fine?

timmyc:
Definitely want @jameskoster to sign off too :)

I might do an alternate PR where the W notif menu is still shown on wc-admin js pages ( dashboard and analytics ) - but I could see that being confusing too

timmyc:
@pmcpinto @jameskoster I'm going to merge this in so we can have a zip file ready for the team to test in the AM, and also hope this will allow an RC2 to get cut soon. If we need to revert this, or look at adding the W menu back on only Analytic pages, we can make that happen and re-build the release.

pmcpinto:
> @pmcpinto Can you confirm this change is fine?

Sounds good to me. Do we need to update any of the new docs?

> I might do an alternate PR where the W notif menu is still shown on wc-admin js pages ( dashboard and analytics ) - but I could see that being confusing too

I think that we can evaluate if in the future it makes sense to keep this feature in the feature plugin.

PR #3828, opened 2020-03-04
Fix tracking on migrated options

joshuatf:
* Adds a constant to `handle_migrated_options` to avoid firing an event when migration occurs.
* Checks to see if post types are registered before attempting to record an event with `WC_Tracks` since it's possibly to fire an event before post types are registered.

I had a difficult time recreating the original error, but this takes 2 preventative measures against it.  /cc @peterfabian or @ObliviousHarmony if you have time to test this.

1. Watch for the event `wcadmin_onboarding_toggled` or add error logging to the events fired in `track_onboarding_toggle()`.
1. Remove the option `woocommerce_onboarding_opt_in` from `wp_options`.
1. Visit any admin page to trigger the option migration.
1. `woocommerce_onboarding_opt_in` should be added to `wp_options` but the event in `track_onboarding_toggle()` should not be recorded.
1. Attempt the steps noted in [this issue]( https://github.com/woocommerce/woocommerce/issues/25766) and make sure it cannot be recreated.

### DIFF
--- a/includes/core-functions.php
+++ b/includes/core-functions.php
@@ -51,10 +51,16 @@ function wc_admin_url( $path = null, $query = array() ) {
  * @param array  $properties Properties to pass along with event.
  */
 function wc_admin_record_tracks_event( $event_name, $properties = array() ) {
+	// WC post types must be registered first for WC_Tracks to work.
+	if ( ! post_type_exists( 'product' ) ) {
+		return;
+	}
+
 	if ( ! class_exists( 'WC_Tracks' ) ) {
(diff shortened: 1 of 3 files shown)


ObliviousHarmony:
@psealock that looks great, exactly what I'd have done :) Ship it!

timmyc:
Props @ObliviousHarmony for all the help here, and thanks @psealock for testing and @joshuatf for the PR.

peterfabian:
Nice work, thanks for the fix y'all!

Issue #3824, opened 2020-03-04
WooCommerce Admin Examples : Add-Report . Filter name change in Woo 4.0


EdithAllison:
For those who are using the WooCommerce Admin Examples "Add-Report" and are working with the beta version of WooCommerce 4.0.

The filter in /woocommerce-admin-add-report-example.php on line 50 
`woocommerce_admin_report_menu_items`
is now called 
`woocommerce_analytics_report_menu_items`

But the extension still fails. It adds a page, but it's empty and throws a JS / React error.  Is there any guidance available how the plugin code needs to be adapated to work with the new Woo 4.0? 

PS: Other examples like "dashboard-section" however work fine in Woo 4.0.

psealock:
>but it's empty and throws a JS / React error.

Hi, Can you add the error message?

EdithAllison:
Sorry, should have added. Error shows as:

index.js:34:729730
`TypeError: e is undefined`

Screenshot with ReactDevTools is here: [https://www.dropbox.com/s/0k5zm7f8th0h9ch/error.png?dl=0](https://www.dropbox.com/s/0k5zm7f8th0h9ch/error.png?dl=0)

psealock:
Hi @EdithAllison, thanks for the bug report!

Here is a fix for the broken extension example, https://github.com/woocommerce/woocommerce-admin/pull/3830.

I'm glad to see these being used and hopefully its helpful to you. If you have any questions, feel free to open a new issue or ask on this one.

EdithAllison:
Hi Paul, many thanks! I've given the fix a go, but still get the same JS error. React is a learning curve for me, so not sure if I'm doing something wrong?

My work environment is WP installed on server (siteground); with my work done on Mac with Terminal. What I've done:

Siteground: Installed WordPress 5.32. 
Siteground: Installed WooCommerce plugin through WP Admin
Siteground: Installed Woo Beta Tester plugin through WP Admin 4.0.0.-rc1
Downloaded WP files onto my Mac 
Mac: cd into plugins folder
Mac: Cloned plugin, applied bugfix 3830, ran node & composer, ran example 
```
git clone https://github.com/woocommerce/woocommerce-admin
cd woocommerce-admin 
git fetch origin pull/3830/head:allison
git checkout allison
npm install 
composer install
npm run example -- --ext=add-report
```
... at this point Terminal hangs but "add-report" folder is created successfully 
[...]

PR #3830, opened 2020-03-04
Add Report Extension Example: Add default props to ReportFilters

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/3824

Add Report extension example was broken due to missing props and a hook name that wasn't updated. Those props shouldn't be required so I added some defaults to account for that.

Table headers weren't styled correctly, so I made sure `TableCard` had the appropriate classes

### Screenshots

Screen Shot 2020-03-05 at 12 22 37 PM

1. `npm run example -- --ext=add-report`.
2. Visit to see the Example working: `/wp-admin/admin.php?page=wc-admin&path=%2Fanalytics%2Fexample`.

### DIFF
--- a/client/analytics/components/leaderboard/index.js
+++ b/client/analytics/components/leaderboard/index.js
@@ -49,7 +49,7 @@ export class Leaderboard extends Component {
 	render() {
 		const { isRequesting, isError, totalRows, title } = this.props;
 		const rows = this.getFormattedRows();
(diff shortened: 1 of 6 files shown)


psealock:
>Should we have one of the columns required to demo required columns for developers?

Good idea, added in https://github.com/woocommerce/woocommerce-admin/pull/3830/commits/324a9448781746842c2a1a8cb7c6a6d4bac82ebf.

PR #3832, opened 2020-03-05
Onboarding: Add offline payment methods

### LINKED ISSUE #3710: Onboarding - payments step: improve flow and include offline payments
The Payments step is performing poorly and this is probably due to some reasons:
- We don't guide users to set up offline payments, while we still have a significant % of stores that activate and use COD and BACS.
- Showing different payment methods together with the subsequent steps can increase cognitive load and cause some confusion.
- We redirect the user to external websites to set up their accounts on Stripe Paypal.

We can't do much about the last point, but we're confident that if we add offline payments set up and tweak the design it will be possible to improve the experience and help users set up their payment methods.

Changes to the current flow:
- Regarding online payments, we're changing the first screen and if the user clicks to set up one of the payments the flow should be identical to the one we have now.
- For COD, the user needs to turn on the option and by doing this it turns the COD Payments setting on core.
- For BACS, if the user clicks to set up we should turn on the respective setting on core, but in this case, we also allow him to fill the bank account details without going to Settings. If he fills the bank details, they should also be recorded in the respective payment settings.
- We're also going to add an option to skip this task. When the user clicks to skip the task we display a modal to get info about the main reason behind skipping the task.
- We also need to make some changes in terms of tracking:

     - Add a new event for when the user clicks "Skip this step": `wcadmin_tasklist_payment_skip_task` (event prop: skip_option: the 3 options that we display in the modal)
     - Add a new event for when the user turns off the radio button:  `wcadmin_tasklist_payment_turnoff_payment` (event prop: payment_method: payment method name)
     - Add a new event for when the user clicks "Done" to complete the task: `wcadmin_tasklist_payment_done` (event prop: payment_methods_installed: list of the payments methods installed)
     - Add a new event for when the user clicks "Set up" in one of the payments or turns on COD: `wcadmin_tasklist_payment_setup` (event props: payment_method_selected: payment method selected, payment_methods_displayed: payment methods displayed). This event ends up replacing `wcadmin_tasklist_payment_choose_method`

[Mockup
](https://www.figma.com/file/nzAjPR4nqdJTDxH32PT4Qg/On-boarding?node-id=29%3A20)

joshuatf:
Fixes #3710 

Adds the offline payment methods (bank transfer and cash on delivery) to the payments task.

### Screenshots
Screen Shot 2020-03-05 at 3 23 25 PM
Screen Shot 2020-03-05 at 3 23 17 PM

1. Enable the task list and visit the payments task (delete the `woocommerce_task_list_payments` option if you need to reset).
1. Toggle the Cash on Delivery method and make sure that it affects the actual setting at `wp-admin/admin.php?page=wc-settings&tab=checkout`.
1. Delete bank transfer accounts if you already have any.
1. Set up the Direct Bank transfer option using any bank details (these are not validated in core either, but added a minimum of account number or IBAN here).
1. Save the configuration and make sure settings appear accurately at `wp-admin/admin.php?page=wc-settings&tab=checkout&section=bacs`.


client/dashboard/task-list/tasks/payments/bacs.js
```diff
@@ -0,0 +1,202 @@
...
+			sort_code: '',
+			iban: '',
+			bic: '',
+		};
+	};
+
+	validate = ( values ) => {
+		const errors = {};
+
+		if ( ! values.account_number && ! values.iban ) {
```
joshuatf: Need a sanity check here.  This requires that either an account number or an IBAN is entered, which I believe is the minimum required across all countries.

There is no validation in core at the moment, so we could optionally also remove this validation.

timmyc:
Thanks for the review here @octaedro

PR #3839, opened 2020-03-06
Add/issue 15 trackings

harriswong:
Add tracking to the following buttons: 
image

### Accessibility

- To verify if tracking is working, you can go to `client/wp-admin-scripts/print-shipping-label-banner/lib/banner-tracks.js` and put a `console.log(eventName)` in the `trackBannerEvent` function.
- Click all the red arrow marker in the screenshot and you should see the event being fired.


client/wp-admin-scripts/print-shipping-label-banner/lib/banner-tracks.js
```diff
@@ -0,0 +1,39 @@
+import { recordEvent } from 'lib/tracks';
+import { getSetting } from '@woocommerce/wc-admin-settings';
+import apiFetch from '@wordpress/api-fetch';
+const JETPACK_NAMESPACE = '/jetpack/v4';
+
+async function getPluginsInformation() {
+	let isJetpackConnected = false;
+	const url = JETPACK_NAMESPACE + '/connection';
+	const result = await apiFetch( {
```
joshuatf: Optional: If we use the wc-api here, we have a selector `isJetpackConnected()` that can check for a Jetpack connection.

client/wp-admin-scripts/print-shipping-label-banner/lib/banner-tracks.js
```diff
@@ -0,0 +1,39 @@
...
+}
+
+/**
+ * Tracks events for banner. By default, include information for jetpack
+ * and woocommerce services.
+ *
+ * @param {string} eventName Name of the event in the form of <source>_<context>_<optional subcontext>_<action>_<optional qualifier>
+ */
+export async function trackBannerEvent( eventName ) {
+	const pluginInformation = await getPluginsInformation();
```
joshuatf: Ideally, I think we should not use an async function here since this will require fetching for each recorded event and also increases the chance an event is not recorded on page switch.

client/wp-admin-scripts/print-shipping-label-banner/lib/banner-tracks.js
```diff
@@ -0,0 +1,39 @@
...
+	recordEvent( eventName, {
+		jetpack_installed: pluginInformation.jetpack_installed,
+		jetpack_connected: pluginInformation.jetpack_connected,
```
joshuatf: If we don't want to use the wc-api method to check for a Jetpack connection, we could alternatively also load it server-side.

src/Features/ShippingLabelBanner.php
```diff
@@ -152,4 +154,20 @@ private function is_supported_country( $country_code ) {
...
+	 * retrieve in component with getSetting('shippingBanner');
+	 *
+	 * @param array $settings Component settings.
+	 * @return array
+	 */
+	public function component_settings( $settings ) {
+		$active_plugins    = Onboarding::get_active_plugins();
+		$installed_plugins = Onboarding::get_installed_plugins();
+
+		$settings['shippingBanner']['isJetPackInstalled']             = in_array( 'jetpack', $installed_plugins, true );
```
joshuatf: Might be a bit more of an edge case, but in the onboarding profiler and task list we have the ability to install and activate plugins client-side.

If you want to guarantee up-to-date data, using the wc-api method `getActivePlugins()` will get the most current array of active plugins.

client/wp-admin-scripts/print-shipping-label-banner/lib/banner-tracks.js
harriswong: Thanks @joshuatf . I will look into how to I can switch to wc-api here.

joshuatf: Feel free to ping me if you need any ideas @harriswong

client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
```diff
@@ -0,0 +1,144 @@
...
+	};
+
+	learnMoreClicked = () => {
+		this.trackBannerEvent( 'shipping_banner_learn_more_click' );
+	};
+
+	trackBannerEvent = ( eventName ) => {
+		recordEvent( eventName, {
+			jetpack_installed: this.activePlugins.includes( 'jetpack' ),
+			jetpack_connected: this.isJetpackConnected,
```
joshuatf: I think you've connected the component successfully, this just needs to be `this.props.isJetpackConnected`.

You can also destructure these above to make things a bit easier to grok:
`const { activePlugins, isJetpackConnected } = this.props`

client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
```diff
@@ -0,0 +1,144 @@
...
+			</div>
+		);
+	}
+}
+
+export default compose(
+	withSelect( ( select ) => {
+		const { getActivePlugins, isJetpackConnected } = select( 'wc-api' );
+		return {
+			activePlugins: getActivePlugins(),
```
joshuatf: This should work okay too without preloading, but if you're worried about a user clicking before this is loaded, we could preload the data outside of onboarding so it's available here as well.

client/wp-admin-scripts/print-shipping-label-banner/dismiss-modal/index.js
```diff
@@ -16,13 +16,18 @@ export default class DismissModal extends Component {
...
+		this.props.trackBannerEvent(
+			'shipping_banner_dismiss_modal_remind_me_later_click'
+		);
 	};
 
 	closeForeverClicked = () => {
 		const { onCloseAll } = this.props;
 		// TODO: Persist something so the banner never appears again.
 		onCloseAll();
+		this.props.trackBannerEvent(
```
c-shultz: I think it would be cleaner to bring in trackBannerEvent on the first line in this function instead of referencing it from this.props directly:
```
 		const { onCloseAll, trackBannerEvent } = this.props; 
```

src/Features/ShippingLabelBanner.php
```diff
@@ -36,6 +36,8 @@ public function __construct() {
 			return;
 		}
 		add_action( 'add_meta_boxes', array( $this, 'add_meta_boxes' ), 6, 2 );
+		add_filter( 'woocommerce_components_settings', array( $this, 'component_settings' ), 20 );
```
superdav42: `woocommerce_components_settings` Is actually deprecated in WC 2.5


There's also a note on `woocommerce_shared_settings`
> Developers, do not use this hook as it is likely to be removed.
Instead, use the data api:
Automattic\WooCommerce\Blocks\Package::container()
    ->get( Automattic\WooCommerce\Blocks\Assets\AssetDataRegistry::class )
    ->add( $key, $value )
So maybe we should do that?

Also probably if we could do it on `meta_box()` or sometime after `should_show_meta_box()` check so we only included it on the pages it is needed.

client/wp-admin-scripts/print-shipping-label-banner/dismiss-modal/index.js
harriswong: I agree!

client/wp-admin-scripts/print-shipping-label-banner/lib/banner-tracks.js
harriswong: Switched to `wc-api`

src/Features/ShippingLabelBanner.php
harriswong: I can replace this to something like
```
$data_registry = Package::container()->get(AssetDataRegistry::class);
if (onboarding not loaded) {
    $data_registry->add( 'onboarding', $settings['onboarding'] );
} else {
    // use onboarding's setting
    $data_registry-><some function to retreieve onboarding data>();
}
```
I couldn't find a way to get the existing settings. Still poking through the `AssetDataRegistry` class, do you know a way to get that please?

client/wp-admin-scripts/print-shipping-label-banner/dismiss-modal/test/index.js
```diff
@@ -0,0 +1,47 @@
...
+
+describe( 'Tracking events in DismissModal', () => {
+	const trackBannerEvent = jest.fn();
+
+	let dismissModalWrapper;
+
+	beforeEach( () => {
+		dismissModalWrapper = shallow(
+			<DismissModal
+				visible={ jest.fn() }
```
c-shultz: `visible` is meant to be treated as a boolean param, so `true` probably makes more sense for the value.

harriswong: Updated

src/Features/ShippingLabelBanner.php
harriswong: Added follow up here: https://github.com/Automattic/woocommerce-shipping-issues/issues/28

PR #3844, opened 2020-03-06
Fix/inconsistent wording downloads report

jobthomas:
Under Analytics > Downloads, there were some inconsistencies in terms of the capitalization and terminology used on the filters vs. table headers below that. 

This PR fixes that:

#### Filters: 

* Product --> Product Title
* Order number --> Order # 

#### Table:

* User Name --> Username

#### Exception

* Didn't add "Address" after "IP" in the table to save space

If the image does not load, please click this link: https://d.pr/i/mowzfp.

### DIFF
--- a/client/analytics/report/downloads/config.js
+++ b/client/analytics/report/downloads/config.js
@@ -52,7 +52,7 @@ export const advancedFilters = applyFilters(
 		filters: {
 			product: {
 				labels: {
-					add: __( 'Product Title', 'woocommerce-admin' ),
+					add: __( 'Product', 'woocommerce-admin' ),
 					placeholder: __( 'Search', 'woocommerce-admin' ),
 					remove: __( 'Remove product filter', 'woocommerce-admin' ),
 					rule: __(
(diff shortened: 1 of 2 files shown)


client/analytics/report/downloads/config.js
```diff
@@ -52,7 +52,7 @@ export const advancedFilters = applyFilters(
 		filters: {
 			product: {
 				labels: {
-					add: __( 'Product', 'woocommerce-admin' ),
+					add: __( 'Product Title', 'woocommerce-admin' ),
```
psealock: I think "Title" would be misleading because the search is for actual products not strings found in the title.

Screen Shot 2020-03-09 at 1 12 53 PM

The Orders Report Advanced Filter for Products is configured the same way.

Screen Shot 2020-03-09 at 1 15 09 PM

client/analytics/report/downloads/table.js
```diff
@@ -55,7 +55,7 @@ export default class DownloadsReportTable extends Component {
 				key: 'order_number',
 			},
 			{
-				label: __( 'User Name', 'woocommerce-admin' ),
+				label: __( 'Username', 'woocommerce-admin' ),
```
psealock: 👍

psealock: I confirmed this is the only instance of this misspelling.

client/analytics/report/downloads/config.js
jobthomas: Cool, I can delete that one again.

jobthomas: Done.

Issue #3841, opened 2020-03-06
Constant not defined when calling `get_active_version()` on WP 5.2


peterfabian:
**Describe the bug**
`\Automattic\WooCommerce\Admin\Composer\Package::get_active_version()` will cause a PHP notice in case when wc-admin plugin is not installed && package is inactive (either when running on WP 5.2 or when disabled via filter on later WP), as `WC_ADMIN_VERSION_NUMBER` is not defined.

**To Reproduce**

1. Use WP 5.2 and install WC 4.0-rc
2. Try to call `\Automattic\WooCommerce\Admin\Composer\Package::get_active_version()`
3. You should see PHP notice about the undefined constant.

Discovered while testing 4.0 on WP 5.2. The use in system status was fixed in https://github.com/woocommerce/woocommerce/pull/25830, but I think it should be fixed in this repo as well.

psealock:
I have a fix https://github.com/woocommerce/woocommerce-admin/pull/3851 to make sure `get_active_version` doesn't reference constants that may not be defined. With the exception of Woo Core, would anyone need to use this function? Should we ask extension developers to use `get_active_version` as THE way to determine if a site has at wc-admin active?

If so, I can add some documentation too.

peterfabian:
Maybe @allendav would be able to say if their plugin is using some parts from wc-admin, if they check for wc admin version? I could see it being useful for extensions built on top of WC Admin, so in my view, having THE way to determine currently loaded version makes sense.

psealock:
>I could see it being useful for extensions built on top of WC Admin, so in my view, having THE way to determine currently loaded version makes sense.

I agree, in the fix's testing instructions I re-create this exact scenario, https://github.com/woocommerce/woocommerce-admin/pull/3851.

psealock:
This was fixed by https://github.com/woocommerce/woocommerce-admin/pull/3851

PR #3851, opened 2020-03-09
fix undefined const WC_ADMIN_VERSION_NUMBER when WP < 5.3

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/3841

>`\Automattic\WooCommerce\Admin\Composer\Package::get_active_version()` will cause a PHP notice in case when wc-admin plugin is not installed && package is inactive (either when running on WP 5.2 or when disabled via filter on later WP), as WC_ADMIN_VERSION_NUMBER is not defined.

Changes in this PR makes sure `\Automattic\WooCommerce\Admin\Composer\Package::get_active_version()` does not access `WC_ADMIN_VERSION_NUMBER` when the plugin is disabled due to minimum requirements not met.

1. Downgrade to WP 5.2 and see wc-admin disabled due to minimum requirements not met.
2. In any plugin you have access to, add this line:
```
error_log( \Automattic\WooCommerce\Admin\Composer\Package::get_active_version() );
```
3. Refresh the page and see and error `Use of undefined constant WC_ADMIN_VERSION_NUMBER` appear in the logs.
3. Copy `src/Package.php` from this PR into Core's packages, ie `woocommerce/packages/woocommerce-admin/src/`
4. Check your logs and see the result be `Inactive` and no error appears.

### DIFF
--- a/src/Package.php
+++ b/src/Package.php
@@ -36,13 +36,21 @@ class Package {
 	private static $package_active = false;
 
 	/**
+	 * Active version
+	 *
+	 * @var bool
+	 */
+	private static $active_version = null;
+
+	/**
 	 * Init the package.
 	 *
 	 * Only initialize for WP 5.3 or greater.
 	 */
 	public static function init() {
-		$wordpress_minimum_met = version_compare( get_bloginfo( 'version' ), '5.3', '>=' );
-		if ( ! $wordpress_minimum_met ) {
+		$feature_plugin_instance = FeaturePlugin::instance();
+		$satisfied_dependencies  = $feature_plugin_instance->has_satisfied_dependencies();
+		if ( ! $satisfied_dependencies ) {
 			return;
 		}
 
@@ -53,7 +61,8 @@ class Package {
 
 		// Avoid double initialization when the feature plugin is in use.
 		if ( defined( 'WC_ADMIN_VERSION_NUMBER' ) ) {
-			$update_version = new WC_Admin_Notes_Deactivate_Plugin();
+			self::$active_version = WC_ADMIN_VERSION_NUMBER;
+			$update_version       = new WC_Admin_Notes_Deactivate_Plugin();
 			if ( version_compare( WC_ADMIN_VERSION_NUMBER, self::VERSION, '<' ) ) {
 				$update_version::add_note();
 			} else {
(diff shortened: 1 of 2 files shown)


src/Package.php
```diff
@@ -43,6 +50,7 @@ class Package {
 	public static function init() {
 		$wordpress_minimum_met = version_compare( get_bloginfo( 'version' ), '5.3', '>=' );
 		if ( ! $wordpress_minimum_met ) {
+			self::$active_version = __( 'Inactive ', 'woocommerce-admin' );
```
joshuatf: Should we leave this as `null` for comparison checks later?  I think the `Inactive` string should be set in core.  We may also need some other changes there as the language is a bit odd currently:

Screen Shot 2020-03-10 at 11 39 00 AM

src/Package.php
```diff
@@ -43,6 +50,7 @@ class Package {
 	public static function init() {
 		$wordpress_minimum_met = version_compare( get_bloginfo( 'version' ), '5.3', '>=' );
```
joshuatf: Not directly from this PR obviously, but I feel the check here should use `get_dependency_errors()` instead of creating required version numbers in multiple places.

src/Package.php
psealock: >Should we leave this as null for comparison checks later?

Probably a good idea. Changed in https://github.com/woocommerce/woocommerce-admin/pull/3851/commits/220d11540329441a6426f56a1efea707002fd44d

src/Package.php
psealock: Good idea! Done in https://github.com/woocommerce/woocommerce-admin/pull/3851/commits/a75cfbb64c6d20db3378a030869a83603fe5fdb7

psealock:
Thanks @joshuatf, this is ready for another look. If you test it again, be sure to transfer the change of `FeaturePlugin/has_satisfied_dependencies` to a public function in `woocommerce/packages/woocommerce-admin/src/` as well.

timmyc:
@psealock thinking this could us a rebase to fix ci, then guessing we want to merge this in for v1.0.1 today?

psealock:
I'm going to merge this because failing tests were fixed in master by https://github.com/woocommerce/woocommerce-admin/pull/3851

PR #3852, opened 2020-03-09
New/shipping banner display logic 13

Ferdev:
Fixes https://github.com/Automattic/woocommerce-shipping-issues/issues/13.

Adds a class to manage the display state of the new shipping label banner defined in the issue above.

### Accessibility

Does not include frontend code.

Execute test command:

`phpunit -c phpunit.xml.dist tests/features/class-wc-tests-shipping-label-banner-display-rules.php`


tests/features/class-wc-tests-shipping-label-banner-display-rules.php
```diff
@@ -0,0 +1,170 @@
...
+	/**
+	 * Setup for tests.
+	 */
+	public function setUp() {
+		parent::setup();
+
+		global $active_plugins;
+		$active_plugins['jetpack'] = 'jetpack/jetpack.php';
+		$active_plugins['wcs']     = 'woocommerce-services/woocommerce-services.php';
+		update_option( 'active_plugins', $active_plugins );
```
harriswong: Does this `update_option` actually perform a DB update during test run?

If we are manipulating the database options in tests, then all subsequent tests run will have a different state for "active plugins". In that case, we may need to restore the original option settings per test case.

src/Features/ShippingLabelBanner.php
```diff
@@ -44,33 +39,11 @@ public function __construct() {
...
-				if ( $product && $product->needs_shipping() ) {
-					return true;
-				}
-			}
+		if ( ! $this->shipping_label_banner_display_rules ) {
+			$this->shipping_label_banner_display_rules = new ShippingLabelBannerDisplayRules();
 		}
 
-		return false;
+		return $this->shipping_label_banner_display_rules->should_display_banner( JETPACK__VERSION );
```
superdav42: There's no guarantee jetpack is installed or active so `JETPACK__VERSION` may be undefined. We should stop passing it as a parameter instead do a `defined` check in `should_display_banner()` and get it if it exists.

src/Features/ShippingLabelBannerDisplayRules.php
```diff
@@ -0,0 +1,192 @@
...
+			return false;
+		}
+
+		if ( $this->jetpack_disconnected() ||
+			$this->jetpack_not_up_to_date( $jetpack_version ) ||
+			$this->wcs_not_installed() ||
+			$this->ups_not_installed() ||
+			$this->fedex_not_installed() ||
+			$this->shipstation_not_installed() ||
+			$this->shippingeasy_not_installed() ) {
```
superdav42: With these rules the banner will basically never show.
I think we [actually](https://github.com/Automattic/woocommerce-shipping-issues/issues/13#issuecomment-595928583) want to check if any of these plugins are installed instead of NOT installed.

src/Features/ShippingLabelBannerDisplayRules.php
```diff
@@ -0,0 +1,192 @@
...
+		return ! is_plugin_active( 'woocommerce-services/woocommerce-services.php' );
+	}
+
+	/**
+	 * Check if jetpack is inactive or disconnected
+	 *
+	 * @return bool
+	 */
+	private function jetpack_disconnected() {
+		return ! is_plugin_active( 'jetpack/jetpack.php' );
```
superdav42: This is only checking if the plugin is installed which does not match the method name or the comments. 
Looks like we only need to check if it is installed according to https://github.com/Automattic/woocommerce-shipping-issues/issues/13#issuecomment-594845206
So I'd just change the method name to `jetpack_installed`

src/Features/ShippingLabelBannerDisplayRules.php
c-shultz: +1
The idea is that users with extensions like Shipstation probably already have an established fulfillment process they use, so there would be less value in us displaying the prompt to them.

src/Features/ShippingLabelBanner.php
Ferdev: I need to pass the jetpack version here as a parameter because it's defined as a constant and it cannot be mocked. So I needed to decouple it from the `ShippingLabelBannerDisplayRules` class. But you're right this could fail... I'll add the check in this method instead.

src/Features/ShippingLabelBannerDisplayRules.php
Ferdev: Ok, thanks guys. I wasn't sure what were the conditions precedence to display the banner. I scanned the related product p2 posts but was difficult to translate that info to the requirements in this issue.

superdav42: As I understand it **ALL** of the following must be true:
* A/B Test Active
* Banner NOT dismissed
* Jetpack >= 4.4 installed and active
* Jetpack Connected
* WCS plugin not installed 
* Shipstation not installed
* UPS not Installed
* Fedex not installed
* Order contains physical products which need shipped
* Store is located in US
* Store currency is set to USD

The only item I'm a little uncertain about is the A/B test. I think the plan was to not use an A/B test at first when released with WooCommerce Admin but when it is eventually integrated into WooCommerce 4.1 we'd use a 50/50 or other A/B test. I don't remember where I saw that but I think for now it means we don't need to worry about an A/B test.

c-shultz: @superdav42,

In regards to the A/B tests, we are planning on targeting 25% of relevant sites when we initially launch: https://github.com/Automattic/woocommerce-shipping-issues/issues/11

We are rushing to get the core functionality of the prompt complete, so I have this marked as a lower priority, but so far this is still the plan

c-shultz: [I have a A/B test branch based on this branch](https://github.com/woocommerce/woocommerce-admin/compare/shipping-banner...add/ab_test?expand=1) I'll make a PR after this merges in case any big changes are made. You can disregard the A/B check in this PR since I'll add it in that branch.

So we actually want to display the banner on sites with WCS installed that have **not** accepted the terms of service. The banner will be a friendlier way for merchant to accept the ToS and launch the shipping modal.

So (excluding the A/B test which I'll handle on the other branch), the conditions should be:

- Banner NOT dismissed
- Jetpack >= 4.4 installed and active
- Jetpack Connected
- No incompatible extensions installed, so
   - Shipstation not installed
   - UPS not Installed
   - Fedex not installed
   - ShippingEasy not installed
- Order contains physical products which need to be shipped ([we should check that the order status is not set to complete](https://a8c.slack.com/archives/CKNC7KACA/p1584395171186200))
- Store is located in US
- Store currency is set to USD
- `WCS plugin not installed` OR` WCS is installed *AND* ToS have NOT been accepted *AND* WCS version is 1.22.5 or greater`
    - The 1.22.5 or greater requirement is so we can launch the shipping modal from the banner

c-shultz: Note: we should be able to check the ToS from wc-admin by calling WCS's `WC_Connect_Options::get_option( 'tos_accepted' )`

Ferdev:
Thanks for the updated info, guys! :heart: I've refactored the display logic, please give it another look whenever you can.

src/Features/ShippingLabelBannerDisplayRules.php
```diff
@@ -0,0 +1,237 @@
...
+		return version_compare( $jetpack_version, $this->min_jetpack_version, '>=' );
+	}
+
+	/**
+	 * Checks if Fedex shipping method is not installed.
+	 *
+	 * @return bool
+	 */
+	private function fedex_not_installed() {
+		return ! is_plugin_active( 'woocommerce-shipping-fedex/woocommerce-shipping-fedex.php' );
```
timmyc: Using the path/slugs to determine if a plugin is active or not doesn't cover the use case of a plugin being installed in a differently named directory. Totally an edge case, but if possible it is nice to check for existence of constant defined in the plugin, or a unique method name.

tests/features/class-wc-tests-shipping-label-banner-display-rules.php
```diff
@@ -0,0 +1,356 @@
+<?php
```
timmyc: Thanks for all the tests here @Ferdev  🙇

src/Features/ShippingLabelBanner.php
```diff
@@ -44,33 +39,36 @@ public function __construct() {
...
-		}
+			if ( class_exists( 'Jetpack_Data' ) ) {
+				$user_token = Jetpack_Data::get_access_token( JETPACK_MASTER_USER );
 
-		$base_location = wc_get_base_location();
-		if ( ! $this->is_supported_country( $base_location['country'] ) ) {
-			return false;
-		}
+				$jetpack_connected = isset( $user_token->external_user_id );
+				$jetpack_version   = JETPACK_VERSION;
```
c-shultz: I'm seeing this defined as `JETPACK__VERSION` (double underscore) not `JETPACK_VERSION`

src/Features/ShippingLabelBanner.php
```diff
@@ -44,33 +39,36 @@ public function __construct() {
...
 
-		// At this point (no packaging data), only show if there's at least one existing and shippable product.
-		foreach ( $order->get_items() as $item ) {
-			if ( $item instanceof \WC_Order_Item_Product ) {
-				$product = $item->get_product();
-				if ( $product && $product->needs_shipping() ) {
-					return true;
-				}
+			if ( class_exists( 'WC_Connect_Loader' ) ) {
+				$wcs_version = WC_Connect_Loader::get_wcs_version();
```
c-shultz:
```suggestion
				$wcs_version = \WC_Connect_Loader::get_wcs_version();
```

src/Features/ShippingLabelBanner.php
```diff
@@ -44,33 +39,36 @@ public function __construct() {
...
+			}
 
-		// At this point (no packaging data), only show if there's at least one existing and shippable product.
-		foreach ( $order->get_items() as $item ) {
-			if ( $item instanceof \WC_Order_Item_Product ) {
-				$product = $item->get_product();
-				if ( $product && $product->needs_shipping() ) {
-					return true;
-				}
+			if ( class_exists( 'WC_Connect_Loader' ) ) {
```
c-shultz:
```suggestion
			if ( class_exists( '\WC_Connect_Loader' ) ) {
```

src/Features/ShippingLabelBanner.php
```diff
@@ -44,33 +39,36 @@ public function __construct() {
...
-		if ( ! $order ) {
-			return false;
-		}
-		// Restrict showing the metabox to supported store currencies.
-		$base_currency = get_woocommerce_currency();
-		if ( ! $this->is_supported_currency( $base_currency ) ) {
-			return false;
-		}
+			if ( class_exists( 'Jetpack_Data' ) ) {
+				$user_token = Jetpack_Data::get_access_token( JETPACK_MASTER_USER );
```
c-shultz:
```suggestion
				$user_token = \Jetpack_Data::get_access_token( JETPACK_MASTER_USER );
```

src/Features/ShippingLabelBanner.php
```diff
@@ -44,33 +39,36 @@ public function __construct() {
...
 
-		if ( ! $order ) {
-			return false;
-		}
-		// Restrict showing the metabox to supported store currencies.
-		$base_currency = get_woocommerce_currency();
-		if ( ! $this->is_supported_currency( $base_currency ) ) {
-			return false;
-		}
+			if ( class_exists( 'Jetpack_Data' ) ) {
```
c-shultz:
```suggestion
			if ( class_exists( '\Jetpack_Data' ) ) {
```

src/Features/ShippingLabelBanner.php
```diff
@@ -44,33 +39,36 @@ public function __construct() {
...
-		foreach ( $order->get_items() as $item ) {
-			if ( $item instanceof \WC_Order_Item_Product ) {
-				$product = $item->get_product();
-				if ( $product && $product->needs_shipping() ) {
-					return true;
-				}
+			if ( class_exists( 'WC_Connect_Loader' ) ) {
+				$wcs_version = WC_Connect_Loader::get_wcs_version();
+			}
+			if ( class_exists( 'WC_Connect_Options' ) ) {
```
c-shultz:
```suggestion
			if ( class_exists( '\WC_Connect_Options' ) ) {
```

src/Features/ShippingLabelBanner.php
```diff
@@ -44,33 +39,36 @@ public function __construct() {
...
+				$wcs_tos_accepted = WC_Connect_Options::get_option( 'tos_accepted' );
```
c-shultz:
```suggestion
				$wcs_tos_accepted = \WC_Connect_Options::get_option( 'tos_accepted' );
```

src/Features/ShippingLabelBannerDisplayRules.php
Ferdev: Thanks for the tip! I replaced all `is_plugin_active` statements with `class_exists`.

Ferdev:
> The logic looks great. I ran into a few PHP errors on my environment.

All PHP errors addressed!

PR #3855, opened 2020-03-09
convert install timestamp to integer

### LINKED ISSUE #3829: PHP Warning: A non-numeric value encountered in NoteTraits.php on line 34
I'm receiving the following warning each time I start WordPress:

Woocommerce Admin PHP Warning: A non-numeric value encountered in [MyPath]\wp-content\plugins\woocommerce-admin\src\Notes\"NoteTraits.php on line 34"

My WP configuration is as follows:
WordPress Version: 5.3.2
Woocommerce Version: Version 3.9.3
OS: Windows: NT 10.0 x64
Web server: Apache/2.4.41 (Win64) OpenSSL/1.1.1d PHP/7.4.2
PHP version: 7.4.2 (Supports 64bit values)
Database: MySQL 8.0.19 x64

rrennick:
Fixes #3829

This PR casts the install timestamp to an `int` to eliminate a warning in PHP 7.4. The `option_value` column is a string field. Earlier versions of PHP would implicitly convert a numeric value to the appropriate type.

- delete the `woocommerce_admin_install_timestamp` option
- refresh the dashboard
- check that `woocommerce_admin_install_timestamp` option was created
- turn off tracking in the WC settings
- delete the `wc-admin-usage-tracking-opt-in` note
- refresh the dashboard
- tracking note should not be added to the DB

### Changelog Note:

Fix: type warning on install timestamp in PHP 7.4

### DIFF
--- a/src/Notes/NoteTraits.php
+++ b/src/Notes/NoteTraits.php
@@ -23,9 +23,9 @@ trait NoteTraits {
 	 */
 	public static function wc_admin_active_for( $seconds ) {
 		// Getting install timestamp reference class-wc-admin-install.php.
-		$wc_admin_installed = get_option( 'woocommerce_admin_install_timestamp', false );
+		$wc_admin_installed = (int) get_option( 'woocommerce_admin_install_timestamp', false );
 
-		if ( false === $wc_admin_installed ) {
+		if ( ! $wc_admin_installed ) {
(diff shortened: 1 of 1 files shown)


psealock:
we'll need to cherry-pick 🍒 this into master

PR #3848, opened 2020-03-06
Merge final `version/1.0` branch with `master`

timmyc:
Bringing in the last bits of the version/1.0 branch to master. Once this is merged any future fixes for the `1.0.x` version should land in master and then be cherry picked back to the release branch.

PR #3857, opened 2020-03-10
Cherry Pick: Changelog: Add Script (#3800)

psealock:
This is the exact same changes as https://github.com/woocommerce/woocommerce-admin/pull/3800, but that was merged after the final merge of `version/1.0` to master in https://github.com/woocommerce/woocommerce-admin/pull/3848.

So this just makes sure #3800 makes its way into master.

EDIT: Alternatively, we can do one more final version/1.0 -> master merge


timmyc:
@psealock should we be able to use [hub](https://github.com/github/hub) to cherry pick this in to master?

psealock:
>@psealock should we be able to use hub to cherry pick this in to master?

Hub is just an easy way to interact with git, right? Are you saying there is a way to cherry-pick without a pull request? If so, I think you can cherry pick into a branch, but because we cant push to master I don't think this will work. Am I understanding correctly?

PR #3874, opened 2020-03-12
Fix failing sync tests

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/3875

Fix failing tests

### Test

```
 vendor/bin/phpunit -c phpunit.xml.dist tests/api/reports-import.php
```


tests/api/reports-import.php
```diff
@@ -134,7 +136,7 @@ public function test_import_params() {
 		$reports  = $response->get_data();
 
 		$this->assertEquals( 200, $response->get_status() );
-		$this->assertCount( 1, $reports );
+		$this->assertCount( 2, $reports );
```
psealock: Two orders were created, so the `$reports` would have two items. I have no idea how this passed before.

PR #3888, opened 2020-03-13
#3793 Tweak: Create admin note if Jetpack plugin doesn't get installed due to an error during OBW

### LINKED ISSUE #3793: Install Jetpack and WCS reminder: inbox notification
Some users are experiencing problems during the Jetpack and WCS install, so we should send them an inbox notification with a reminder to retry to install the extensions:

**Title:** Uh oh...There was a problem during the Jetpack and WooCommerce Services install. Please try again
**Copy:** We noticed that there was a problem during the Jetpack and WooCommerce Services install. Please try again and enjoy all the advantages of having the plugins connected to your store! Sorry for the inconvenience. The "Jetpack" and "WooCommerce Services" plugins will be installed & activated for free.
**Store type:** New users that had trouble installing Jetpack and WCS during the onboarding (`wcadmin_install_plugin_error`)
**Trigger:** Right after the install failed (`wcadmin_install_plugin_error`)
**Cadence:** Once
**CTA leads to:** Install Jetpack and WooCommerce Services and redirects the user to the [wp.com registration/login flow](https://www.figma.com/file/b0b2CrxPZz0i1aAPuv985tvK/Store-Profiler?node-id=1667%3A560)
**CTA copy:** Install plugins

becdetat:
Fixes #3793

This creates an admin note prompting the user to retry install of the Jetpack or WooCommercev Services plugins if they failed on installation during the wizard.

1. Delete the Jetpack plugin if it is installed
2. Simulate a failing plugin install by adding `true ||` to the condition on line ~241~ 253 of `OnboardingPlugins.php`:
image
3. Delete all existing notes from the `wp_wc_admin_notes` table
4. Activate the onboarding wizard via WooCommerce -> Reports -> Help -> Setup wizard -> Profile Setup Wizard -> Enable
5. Proceed through the wizard, make sure you choose to install Jetpack
6. When Jetpack installation fails, navigate to `/wp-admin/admin.php?page=wc-reports` to break out of the onboarding wizard.
7. open the inbox and open the new message
8. Click the "Install plugins" CTA
9. The button will go inactive for a while then the page will reload
10. The message will have been actioned (won't be shown)
11. The Jetpack and WooCommerce Services plugins should be installed and activated.

The notification should appear in the inbox as follows:

image

### Changelog Note:

Tweak: Create admin note if Jetpack or WooCommerce Services plugin doesn't get installed due to an error during OBW


src/API/OnboardingPlugins.php
```diff
@@ -156,6 +157,19 @@ public function update_item_permissions_check( $request ) {
 		return true;
 	}
 
+	/**
+	 * Create an alert notification in response to an error installing a plugin.
+	 *
+	 * @param string $slug The slug of the plugin being installed.
+	 */
+	private function create_install_plugin_error_inbox_notification_for_jetpack_installs( $slug ) {
+		if ( 'jetpack' !== $slug ) {
```
joshuatf: Do we also want to check for `woocommerce-services` here?

Additionally, I think it might be nice to move some of this logic inside the `possibly_add_install_jetpack_note` method and check there instead.

src/Notes/WC_Admin_Notes_Install_Jetpack_Plugin.php
```diff
@@ -0,0 +1,53 @@
...
+		$note->set_content( $content );
+		$note->set_content_data( (object) array() );
+		$note->set_type( WC_ADMIN_Note::E_WC_ADMIN_NOTE_INFORMATIONAL );
+		$note->set_icon( 'mail' );
+		$note->set_name( self::NOTE_NAME );
+		$note->set_source( 'woocommerce-admin' );
+		$note->add_action(
+			'install-jetpack-plugin',
+			__( 'Install plugins', 'woocommerce-admin' ),
+			admin_url( 'admin.php?page=wc-admin&step=plugins' ),
```
joshuatf: The onboarding profiler may be complete or turned off at this point.  We may have to do this through a callback method and install via the API directly.

There are some examples in other notes, but feel free to ping me on Slack if you need any tips.

src/API/OnboardingPlugins.php
```diff
@@ -233,6 +249,8 @@ public function install_plugin( $request ) {
 			);
 			wc_admin_record_tracks_event( 'install_plugin_error', $properties );
 
+			$this->create_install_plugin_error_inbox_notification_for_jetpack_installs( $slug );
```
joshuatf: This works well for adding the note, but we'll probably need a way to remove the note as well if they install on their own.

We could check on every load if these plugins exist in active plugins, but this is a little brute force.  The `activate_PLUGINNAME` would work if we want to remove this note only after activation.

joshuatf:
This could probably also use a rebase from master while you're at it.

src/API/OnboardingPlugins.php
joshuatf: It's also likely that we'll be making this plugins API less onboarding specific and more for general use.

Since we don't know exactly how that's going to flesh out yet, I think a comment in the above docblock that this should be moved to a filter should be enough.

becdetat: > Do we also want to check for `woocommerce-services` here?

Sure, but I couldn't actually find where `woocommerce-services` was being installed here, I assumed that they were part of the same extension.

becdetat: > Additionally, I think it might be nice to move some of this logic inside the `possibly_add_install_jetpack_note` method and check there instead.

I'm not sure I agree with this 100% as it means `possibly_add_install_jetpack_note` has to know about slugs, shifting that logic down another making it harder to find, but I'll do this anyway.

becdetat: > I think a comment in the above docblock that this should be moved to a filter should be enough.
Done

src/API/OnboardingPlugins.php
becdetat: Done 😅 

https://media.giphy.com/media/R32vNOs0qpIqI/giphy.gif

becdetat:
> This could probably also use a rebase from master while you're at it.

Done :-)

src/Notes/WC_Admin_Notes_Filters.php
```diff
@@ -0,0 +1,31 @@
...
+
+ /**
+  * This class adds actions to manage notes.
+  */
+class WC_Admin_Notes_Filters {
+	/**
+	 * Init
+	 */
+	public function init() {
+		add_action( 'activate_jetpack/jetpack.php', array( $this, 'action_jetpack_woocommerce_services_activation_note' ) );
```
joshuatf: Nice! 💯 

Do you think we should move these to a constructor in `WC_Admin_Notes_Install_Jetpack_And_WooCommerce_Services_Plugin` and initialize that somewhere to keep the logic contained?  My concern is that this class could become a dumping ground for all note filters.

src/Notes/WC_Admin_Notes_Install_Jetpack_And_WooCommerce_Services_Plugin.php
```diff
@@ -0,0 +1,75 @@
...
+		$note->set_content( $content );
+		$note->set_content_data( (object) array() );
+		$note->set_type( WC_ADMIN_Note::E_WC_ADMIN_NOTE_INFORMATIONAL );
+		$note->set_icon( 'mail' );
+		$note->set_name( self::NOTE_NAME );
+		$note->set_source( 'woocommerce-admin' );
+		$note->add_action(
+			'install-jetpack-plugin',
+			__( 'Install plugins', 'woocommerce-admin' ),
+			admin_url( 'admin.php?page=wc-admin&step=plugins' ),
```
joshuatf: This step and path will be going away soon (see https://github.com/woocommerce/woocommerce-admin/issues/3894 ) so we may want to try and action this via the plugins API or a callback method here server-side.

Feel free to ping me if you need any more clarification on this 😄

src/Notes/WC_Admin_Notes_Install_Jetpack_And_WooCommerce_Services_Plugin.php
```diff
@@ -0,0 +1,75 @@
...
+	const NOTE_NAME = 'wc-admin-install-jetpack-plugin';
+
+	/**
+	 * Possibly add the Install Jetpack note.
+	 *
+	 * @param string $slug The slug of the plugin being installed.
+	 */
+	public static function possibly_add_install_jetpack_and_woocommerce_services_note( $slug ) {
+		// Exit early if we're not installing the Jetpack or the WooCommerce Services plugins.
+		if ( 'jetpack' !== $slug && 'woocommerce-services' !== $slug ) {
```
joshuatf: I actually agree with your comment:

https://github.com/woocommerce/woocommerce-admin/pull/3888#discussion_r394765799

Long-term we will be making the plugins API more generic and not onboarding specific so the solution is probably to move this check to an action:

```
do_action( 'install_plugin_error', $slug, $error, ... )
```

And hook that action:
```
add_action( 'install_plugin_error', array( $this, 'possibly_add_jetpack_wcs_error_note' );
function possibly_add_jetpack_wcs_error_note( $slug ) {
    if ( $slug === 'jetpack' || $slug === 'woocommerce-services' ) {
        WC_Admin_Notes_Install_Jetpack_And_WooCommerce_Services_Plugin ::add_install_jetpack_and_woocommerce_services_note()
    }
}
```

But in the meantime, we can just add a `@todo` note here since we don't know exactly how that API will change.

src/API/OnboardingPlugins.php
joshuatf: Sorry, one more thing I missed on my last review- I think we'll want to call `create_install_plugin_error_inbox_notification_for_jetpack_installs()` on activation fail as well.

src/Notes/WC_Admin_Notes_Filters.php
becdetat: I think that's a premature refactor, I like how there's separation between the hook and `WC_Admin_Notes_Install_Jetpack_And_WooCommerce_Services_Plugin` so I would rather keep that for now at least. Save refactoring it until this does actually start to become a dumping ground and the way forward for refactoring it becomes clearer.

src/Notes/WC_Admin_Notes_Install_Jetpack_And_WooCommerce_Services_Plugin.php
becdetat: Ok, sounds good.

A pattern I've seen in WP's hooks is a dynamic hook name, like:

```
do_action( 'install_plugin_error_' . $slug, $error, ... );
```

which mean that instead of having to implement an early return in the handler you can just make it declarative in the action registration:

```
add_action( 'install_plugin_error_jetpack', array( $this, 'add_install_jetpack_and_woocommerce_services_note' ) );
add_action( 'install_plugin_error_woocommerce-services, array( $this, 'add_install_jetpack_and_woocommerce_services_note' ) );
```

src/API/OnboardingPlugins.php
becdetat: Done - https://github.com/woocommerce/woocommerce-admin/pull/3888/files#diff-f09c46da6591739c273804943f0ce4b6R311

becdetat: This is done

src/Notes/WC_Admin_Notes_Install_Jetpack_And_WooCommerce_Services_Plugin.php
becdetat: Done

becdetat:
Woohoo, I think I've covered everything requested now. @joshuatf

becdetat:
I've just rebased this. @joshuatf can you have one more look please?

src/Notes/WC_Admin_Notes_Install_Jetpack_And_WooCommerce_Services_Plugin.php
```diff
@@ -0,0 +1,123 @@
...
+	 *
+	 * @return boolean If the plugin was installed and activated.
+	 */
+	private function install_and_activate_plugin( $plugin ) {
+		$install_request = array( 'plugin' => $plugin );
+		$installer       = new \Automattic\WooCommerce\Admin\API\OnboardingPlugins();
+		$result          = $installer->install_plugin( $install_request );
+
+		if ( is_wp_error( $result ) ) {
+			return false;
```
joshuatf, line 111: It doesn't look like these booleans are used right now and currently this note will be actioned even if install fails again.

I think for the time being we can just set to `return;` and add a `@todo` comment to reset the actioned status of the note on failure.

src/Notes/WC_Admin_Notes_Filters.php
becdetat: Ok, sorry, I just don't know the system well enough yet. Will do.

becdetat: Fixed in 0aa9c0386d318abab2432fbf4c6a073be27e3cd3

src/Notes/WC_Admin_Notes_Install_Jetpack_And_WooCommerce_Services_Plugin.php
becdetat: Fixed in d0802a2bda60ed0a71bed7bba52f1979e2dad50b

becdetat:
@jeffstieler 

> This is testing well for the most part. Notably this functionality seems be missing:
> 
> > You should be taken back to the Jetpack/WooCommerce Services install step of the OBW

Whoops, those instructions became out of date - the flow of the OBW changed and installing the plugins from the admin  note no longer redirects back to the OBW. I've updated the instructions to suit.

src/Notes/WC_Admin_Notes_Filters.php
```diff
@@ -0,0 +1,23 @@
...
+ */
+class WC_Admin_Notes_Filters {
+	/**
+	 * Remove (by actioning) activation note for Jetpack and WooCommerce Services
+	 * when either of those plugins are activated.
+	 */
+	public function action_jetpack_woocommerce_services_activation_note() {
+		WC_Admin_Notes_Install_Jetpack_And_WooCommerce_Services_Plugin::action_install_jetpack_and_woocommerce_services_note();
+	}
+}
```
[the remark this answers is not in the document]
becdetat: I'm not sure how I ended up with that :-/ . I've cleaned this up in a3b520d02871f296a1c948f49be19cea85ff1308.

src/Notes/WC_Admin_Notes_Install_Jetpack_And_WooCommerce_Services_Plugin.php
```diff
@@ -0,0 +1,121 @@
...
+		}
+
+		$content = __( 'We noticed that there was a problem during the Jetpack and WooCommerce Services install. Please try again and enjoy all the advantages of having the plugins connected to your store! Sorry for the inconvenience. The "Jetpack" and "WooCommerce Services" plugins will be installed & activated for free.', 'woocommerce-admin' );
+
+		$note = new WC_Admin_Note();
+		$note->set_title( __( 'Uh oh...There was a problem during the Jetpack and WooCommerce Services install. Please try again.', 'woocommerce-admin' ) );
+		$note->set_content( $content );
+		$note->set_content_data( (object) array() );
+		$note->set_type( WC_ADMIN_Note::E_WC_ADMIN_NOTE_INFORMATIONAL );
+		$note->set_icon( 'mail' );
```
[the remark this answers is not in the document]
becdetat: Changed to plugins in b60328abdf8a009199c90abb2ede4481615063f5

src/Notes/WC_Admin_Notes_Install_Jetpack_And_WooCommerce_Services_Plugin.php
```diff
@@ -0,0 +1,121 @@
...
+		// Exit early if there is already a note to install Jetpack.
+		$note_ids = $data_store->get_notes_with_name( self::NOTE_NAME );
+		if ( ! empty( $note_ids ) ) {
+			return;
+		}
+
+		$content = __( 'We noticed that there was a problem during the Jetpack and WooCommerce Services install. Please try again and enjoy all the advantages of having the plugins connected to your store! Sorry for the inconvenience. The "Jetpack" and "WooCommerce Services" plugins will be installed & activated for free.', 'woocommerce-admin' );
+
+		$note = new WC_Admin_Note();
+		$note->set_title( __( 'Uh oh...There was a problem during the Jetpack and WooCommerce Services install. Please try again.', 'woocommerce-admin' ) );
```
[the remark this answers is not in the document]
becdetat: There should - I copied this directly from the copy in the issue. Fixed in f7f5fcb8c09d62e7708acf8cd60ef8acf20b1049

src/Notes/WC_Admin_Notes_Install_Jetpack_And_WooCommerce_Services_Plugin.php
```diff
@@ -0,0 +1,121 @@
...
+ */
+
+namespace Automattic\WooCommerce\Admin\Notes;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * WC_Admin_Notes_Install_Jetpack_And_WooCommerce_Services_Plugin
+ */
+class WC_Admin_Notes_Install_Jetpack_And_WooCommerce_Services_Plugin {
```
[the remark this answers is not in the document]
becdetat: Sure, fixed in eb0593e8a3a50134f49a4c5ae19f2fe44243d302

src/Notes/WC_Admin_Notes_Install_Jetpack_And_WooCommerce_Services_Plugin.php
```diff
@@ -0,0 +1,121 @@
...
+			true
+		);
+
+		$note->save();
+	}
+
+	/**
+	 * Action the Install Jetpack notes, if any exists.
+	 */
+	public static function action_install_jetpack_and_woocommerce_services_note() {
```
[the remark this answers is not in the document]
becdetat: renamed to `action_note` in  eb0593e8a3a50134f49a4c5ae19f2fe44243d302

src/Notes/WC_Admin_Notes_Install_Jetpack_And_WooCommerce_Services_Plugin.php
```diff
@@ -0,0 +1,121 @@
...
+		$data_store = \WC_Data_Store::load( 'admin-note' );
+		$note_ids   = $data_store->get_notes_with_name( self::NOTE_NAME );
+
```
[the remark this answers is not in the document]
becdetat: This was filled with sharks - `PluginsHelper::get_active_plugin_slugs()` didn't return the just-activated plugin in the `activated_{$plugin}` action so I had to do some hacking - but it's done in 5ae961760ef52a4933f576ea4ae9c24082299f0f

becdetat:
@jeffstieler thanks for the feedback, this is ready for another review if you're keen :-)

src/Notes/WC_Admin_Notes_Install_Jetpack_And_WooCommerce_Services_Plugin.php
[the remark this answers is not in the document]
jeffstieler: Is that perhaps because [`activate_{$plugin}`](https://github.com/WordPress/WordPress/blob/master/wp-admin/includes/plugin.php#L690) occurs before the plugin has been activated?

Maybe hooking into [`activated_plugin`](https://github.com/WordPress/WordPress/blob/master/wp-admin/includes/plugin.php#L717) would allow for the removal of the `action_note_wcs()` and `action_note_jp()` intermediate functions.

src/Notes/WC_Admin_Notes_Install_JP_And_WCS_Plugins.php
```diff
@@ -0,0 +1,150 @@
...
+	public static function action_note_jp() {
+		self::action_note( 'jp' );
+	}
+
+	/**
+	 * This gets triggered when the WooCommerce Services plugin is activated.
+	 */
+	public static function action_note_wcs() {
+		self::action_note( 'wcs' );
+	}
```
jeffstieler, line 85: A note about these functions here: https://github.com/woocommerce/woocommerce-admin/pull/3888#discussion_r406283718

src/Notes/WC_Admin_Notes_Install_Jetpack_And_WooCommerce_Services_Plugin.php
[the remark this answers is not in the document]
becdetat: Awesome, I missed that hook somehow. Fixed in 8bc90853f43ab79f8f980fc6a4acfeeb619999e3 which simplifies things quite a bit.

Issue #3889, opened 2020-03-13
Github releases include files that should be ignored


psealock:
Github releases include many files that should not be there.

For .org releases, it complicates the copy/paste, making the process more error prone.

For Core releases, it means users need to download unneeded files, increasing the plugin bundle size.

## Included now

![Screen Shot 2020-03-13 at 5 02 17 PM](https://user-images.githubusercontent.com/1922453/76588802-e610c500-654c-11ea-8217-3a3a9ea02c1d.png)

## What should be included

![Screen Shot 2020-03-13 at 4 57 27 PM](https://user-images.githubusercontent.com/1922453/76588815-f032c380-654c-11ea-96ff-c72c1753acf7.png)

It looks like some ignored files weren't included in https://github.com/woocommerce/woocommerce-admin/pull/3823. Perhaps this is a good case for explicitly including files instead of excluding files?

pmcpinto:
I'm closing this one as it seems that it was fixed by #3890

PR #3890, opened 2020-03-13
Zip: include files instead of exclude

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/3889

This PR goes back to explicitly naming which files to add to a zip instead of which files to exclude. Although it means `npm run test:zip` will remove dev packages from your local install, I think the pain point is small compared to the possibility of erroneously adding a file.

I'm open to debate on that point.

The zip goes from 6.4 MB to 1.3 MB

### Test

1. Run `npm run test:zip` and inspect files. It should be the ones from https://github.com/woocommerce/woocommerce-admin/issues/3889.
2. Add the zip to a test site and make sure the app loads correctly.

### DIFF
--- a/bin/make-zip.sh
+++ b/bin/make-zip.sh
@@ -21,52 +21,15 @@ output 2 "Creating archive... 🎁"
 
 ZIP_FILE=$1
 
-# Folders listed first then individual files
-zip -r ${ZIP_FILE} . \
-	-x \
-	.git/\* \
-	tests/\* \
-	bin/\* \
-	config/\* \
-	node_modules/\* \
-	vendor/bin/\* \
-	vendor/dealerdirect/\* \
-	vendor/doctrine/\* \
-	vendor/phar-io/\* \
-	vendor/phpcompatibility/\* \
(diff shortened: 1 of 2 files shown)


bin/make-zip.sh
```diff
@@ -21,52 +21,15 @@ output 2 "Creating archive... 🎁"
...
+zip -r ${ZIP_FILE} \
+	woocommerce-admin.php \
+	uninstall.php \
+	includes/ \
+	images/ \
+	$build_files \
+	languages/woocommerce-admin.pot \
+	readme.txt \
+	src/ \
+	vendor/
```
joshuatf: I think there are some vendor files that we don't need to include in the release.  See https://github.com/woocommerce/woocommerce-admin/pull/3823/commits/760bcd346fd9c3195cdc9177b2b7320810a33fdf

rrennick: The vendor subfolders that were being excluded are not needed for the built plugin but do add to the size to the zip

PR #3895, opened 2020-03-13
"Personalize your store" reminder: new inbox notification

### LINKED ISSUE #3789: "Personalize your store" reminder: new inbox notification
Following the recent change in the homepage creation redirect (#3708), we want to add an inbox notification that can act as a reminder to personalize the homepage that was automatically created:

**Title:** Personalize your store's homepage
**Copy:** The homepage is one of the most important entry points in your store. When done right it can lead to higher conversions and engagement. Don't forget to personalize the homepage that we created for your store during the onboarding.
**Store type:** New users that created a homepage through the task list (`wcadmin_tasklist_appearance_create_homepage`)(`create_homepage:true`)
**Trigger:** 5 days after registration or when the user completes the task list (`wcadmin_tasklist_completed`)
**Cadence:** Once
**Action CTA leads to:** the page that was created as homepage
**CTA copy**: Personalize homepage

octaedro:
Fixes #3789

Added note to Personalize Store. It will appear after 5 days from registration or when the user completes the task list.

### Screenshots
screenshot-personalize

- After triggering the cron, go to the WooCommerce Dashboard.
- Press the` Inbox` button in the Navbar
- A note with the text: `Personalize your store's homepage` should be visible.
- Press the` Personalize homepage` button.
- It should redirect to the `Front Page` editor

### DIFF
--- a/src/Notes/WC_Admin_Notes_Personalize_Store.php
+++ b/src/Notes/WC_Admin_Notes_Personalize_Store.php
@@ -1,0 +1,70 @@
+<?php
+/**
+ * WooCommerce Admin Personalize Your Store Note Provider.
+ *
+ * Adds a note to the merchant's inbox prompting them to personalize their store.
+ *
+ * @package WooCommerce Admin
+ */
+
+namespace Automattic\WooCommerce\Admin\Notes;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * WC_Admin_Notes_Personalize_Store
+ */
+class WC_Admin_Notes_Personalize_Store {
+	/**
+	 * Note traits.
+	 */
+	use NoteTraits;
+
+	/**
+	 * Name of the note for use in the database.
+	 */
+	const NOTE_NAME = 'wc-admin-personalize-store';
+
+	/**
+	 * Possibly add the note.
+	 */
+	public static function possibly_add_personalize_store_note() {
+		// Only show the note to stores with homepage.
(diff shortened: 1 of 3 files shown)


src/Notes/WC_Admin_Notes_Personalize_Store.php
```diff
@@ -0,0 +1,70 @@
...
+	 */
+	const NOTE_NAME = 'wc-admin-personalize-store';
+
+	/**
+	 * Possibly add the note.
+	 */
+	public static function possibly_add_personalize_store_note() {
+		// Only show the note to stores with homepage.
+		$homepage_id = get_option( 'woocommerce_onboarding_homepage_post_id', false );
+		if ( 0 > $homepage_id ) {
```
joshuatf: Good call here!  Can we move this logic (zero checking) to the `woocommerce_onboarding_homepage_post_id` creation so we can just check `if ( $homepage_id )` here?

https://github.com/woocommerce/woocommerce-admin/blob/8585ed89b80c394b9cec160ded8c591a8fbfa20d/src/API/OnboardingTasks.php#L416

Would be useful in case we need to check this in other areas and ever get a `0` return value from `wp_insert_post()`.

octaedro: Good idea Josh!
I just added it in the commit [623350f](https://github.com/woocommerce/woocommerce-admin/pull/3895/commits/623350ff91cdc1fe2da1834ae2845333b48bf026)

PR #3901, opened 2020-03-13
Install and acitvate WCS from banner

harriswong:
Fixes https://github.com/Automattic/woocommerce-shipping-issues/issues/12

This change is based on the existing work from plugins https://github.com/woocommerce/woocommerce-admin/blob/master/client/dashboard/task-list/tasks/steps/plugins.js. It is to add similar behaviour to shipping banner such that clicking "Create shipping label" allows us to automate WCS installation and activation.

### Accessibility

- Deactivate WooCommerce Services (aka WCS)
- Delete WCS from `wordpress/wp-content/plugins`
- Open up an order page (ie. wp-admin/post.php?post=14&action=edit)
- Click "Create shipping label"
- Wait a bit
- Go to plugins page and confirm WCS is installed and activated (ie. wp-admin/plugins.php?plugin_status=all&paged=1&s)

### DIFF
--- a/src/API/Init.php
+++ b/src/API/Init.php
@@ -74,7 +74,7 @@ class Init {
 			'Automattic\WooCommerce\Admin\API\Themes',
 		);
 
-		if ( Loader::is_onboarding_enabled() ) {
+		if ( Loader::is_onboarding_enabled() || Loader::is_feature_enabled( 'shipping-label-banner' ) ) {
 			$controllers = array_merge(
 				$controllers,
 				array(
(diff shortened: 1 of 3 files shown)


src/API/Init.php
```diff
@@ -74,7 +74,7 @@ public function rest_api_init() {
 			'Automattic\WooCommerce\Admin\API\Themes',
 		);
 
-		if ( Loader::is_onboarding_enabled() ) {
+		if ( Loader::is_onboarding_enabled() || Loader::is_feature_enabled( 'shipping-label-banner' ) ) {
```
harriswong: We will enable the onboarding REST end points for shipping-label-banner as well. By doing this, we probably no longer need to set onboarding settings in `component_settings` (`src/Features/ShippingLabelBanner.php`) anymore.

harriswong:
I agree. Especially we are still in the middle of adding some more functionalities into the component. I think it will be more clear what we can pull out later.

PR #3902, opened 2020-03-13
Add shipping banner error notice

c-shultz:
Fixes https://github.com/Automattic/woocommerce-shipping-issues/issues/33

Introduces error message to banner: 
image

The PR makes the error message always present, so we'll need to set the state accordingly on a separate PR.

Error message is enabled by setting `isSetupError` state variable to `true`.
And the error reason is controlled by setting the `setupErrorReason` to specific values. e.g. setting to  `setupErrorTypes.DOWNLOAD` means that the download failed.


harriswong:
👍 🚢

PR #3908, opened 2020-03-16
Add WooCommerce support if store is using default theme

### LINKED ISSUE #3630: Onboarding - theme selection step: WC support label on Twenty Twenty theme
Currently in the theme selection step, we display an alert message saying that the Twenty Twenty theme doesn't support WC. That alert message should be removed as the theme is compatible with WC, the 3.8.1 release was mostly focused on compatibility improvements: p6q8Tx-1m3-p2

Is it possible to confirm if this is a problem on our side? Maybe it's an issue on the theme. Thanks

Captura de ecrã 2020-01-27, às 11 03 56

joshuatf:
Fixes #3630 

Adds the WooCommerce support tag in the profiler when a default theme is active since WC support is added directly via core for these themes.

### Before
Screen Shot 2020-03-16 at 2 03 19 PM

### After
Screen Shot 2020-03-16 at 2 14 45 PM

1. Activate a default Twenty-* theme.
2. Delete the `wc_onboarding_themes` transient.
3. Visit the “Themes” step of the profiler.
4. Make sure the red icon stating the theme does not support WooCommerce is not shown.
5. Activate a theme that does not have WC support.
6. Make sure the icon is shown.

### DIFF
--- a/src/Features/Onboarding.php
+++ b/src/Features/Onboarding.php
@@ -309,6 +309,11 @@ class Onboarding {
 				$themes[ $slug ]  = $theme_data;
 			}
 
+			// Add the WooCommerce support tag for default themes that don't explicitly declare support.
+			if ( wc_is_wp_default_theme_active() ) {
+				$themes[ $active_theme ]['has_woocommerce_support'] = true;
+			}
+
 			$themes = array( $active_theme => $themes[ $active_theme ] ) + $themes;
 
 			set_transient( self::THEMES_TRANSIENT, $themes, DAY_IN_SECONDS );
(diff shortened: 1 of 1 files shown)


octaedro:
Hi @joshuatf, you did a nice job here.
There is a Travis check that is failing, it said that the function `wc_is_wp_default_theme_active` is not defined.
Could you check this out?

joshuatf:
Thanks for the heads up @octaedro!

It looks like this method was introduced in 4.0.0 ( https://github.com/woocommerce/woocommerce/pull/25216 ) so will not work with our travis checks in 5.6 since it uses WC 3.8.1.]

I added a check to see if this method exists before we call it.  This means that the WC support check will not work for older versions of WC but since this is a minor feature that just displays an icon if support isn't found, I think this is safe to exclude for older versions of WC.  /cc @timmyc if you have any thoughts.

PR #3912, opened 2020-03-17
Reduce asset sizes for shipping banner

superdav42:
Fixes https://github.com/Automattic/woocommerce-shipping-issues/issues/30

Reduce JS size by only using Dashicons and no gridicons.
Also Include the styles used from wordpress/components instead of all of the styles

### DIFF
--- a/src/Features/ShippingLabelBanner.php
+++ b/src/Features/ShippingLabelBanner.php
@@ -126,17 +126,18 @@ class ShippingLabelBanner {
 	 * @param string $hook current page hook.
 	 */
 	public function add_print_shipping_label_script( $hook ) {
+		$rtl = is_rtl() ? '.rtl' : '';
 		wp_enqueue_style(
 			'print-shipping-label-banner-style',
-			Loader::get_url( 'print-shipping-label-banner/style.css' ),
-			array(),
+			Loader::get_url( "print-shipping-label-banner/style{$rtl}.css" ),
+			array( 'wp-components' ),
 			Loader::get_file_version( 'print-shipping-label-banner/style.css' )
 		);
 
 		wp_enqueue_script(
 			'print-shipping-label-banner',
 			Loader::get_url( 'wp-admin-scripts/print-shipping-label-banner.js' ),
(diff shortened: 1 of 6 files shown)


harriswong:
I see the last commit ac7c67c2eab9b0914c0425213ccfec265e73b788 is somehow a duplicate of the existing dc4e725d in `shipping-banner`. I hope git is smart enough to remove on of them.

src/Features/ShippingLabelBanner.php
```diff
@@ -126,17 +126,18 @@ private function count_shippable_items( \WC_Order $order ) {
 	 * @param string $hook current page hook.
 	 */
 	public function add_print_shipping_label_script( $hook ) {
+		$rtl = is_rtl() ? '.rtl' : '';
```
harriswong: @superdav42  Is it `-rtl` or `.rtl`?

src/Features/ShippingLabelBanner.php
```diff
@@ -126,17 +126,18 @@ private function count_shippable_items( \WC_Order $order ) {
...
 		wp_enqueue_style(
 			'print-shipping-label-banner-style',
-			Loader::get_url( 'print-shipping-label-banner/style.css' ),
-			array(),
+			Loader::get_url( "print-shipping-label-banner/style{$rtl}.css" ),
+			array( 'wp-components' ),
```
harriswong: `client/wp-admin-scripts/print-shipping-label-banner/setup-notice/index.js` still uses `wp-components`, shouldn't this be inside `wp_enqueue_script`?

src/Features/ShippingLabelBanner.php
superdav42: You're right it's -rtl.

src/Features/ShippingLabelBanner.php
superdav42: `wp-components` is actually a stylesheet dependency so it needs to be in the styles deps. It's exactly what the [doc says](https://developer.wordpress.org/block-editor/packages/packages-components/#usage).

superdav42:
It's still loading about 100kb just for the dashicons because of the way the svg files are unlined in the JS. It seems really inefficient, maybe I'll open an issue there.

superdav42:
Looks like we could possibly use the `@wordpress/fonts` package to save even more space. I'll investigate if we have time.

PR #3915, opened 2020-03-17
Show loading state when installing or activating WCS

harriswong:
Fixes https://github.com/Automattic/woocommerce-shipping-issues/issues/34

Disable and show a "busy" state for the "Create shipping label" button when installing or activating WCS. 

### Accessibility

- Deactivate WooCommerce Service
- Remove the plugin from your plugin folder `wp-content/plugins/woocommerce-admin/woocommerce-services`
- Go to an order page "wp-admin/post.php?post=14&action=edit"
- Click "Create shipping label"
image
- When it is installing/activating WCS, the button should look like this:
Kapture 2020-03-17 at 9 43 57
- Click "dismiss button", nothing should happen during installation/activation.
image


harriswong:
@c-shultz rebased `shipping-banner` and resolved conflict.

PR #3924, opened 2020-03-18
Made the data store reading of content_data more resilient

### LINKED ISSUE #3923: Uncaught Exception on WPMU
**Describe the bug**
I get an "Uncaught Exception" error on an individually activated site in a WordPress Multisite Network.

Here the full error:

```
Uncaught Exception: The admin note content_data prop must be an instance of stdClass. in /var/web/site/public_html/wp-content/plugins/woocommerce/includes/abstracts/abstract-wc-data.php on line 810
error()
set_content_data()
read()
read()
__construct()
should_show_notice()
show_reminder()
apply_filters()
do_action()
do_action()
set_current_screen()
set_current_screen()
require_once()
```

- [ All individually activated ] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [ x ] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/storefront/).
-  [] I can reproduce this bug consistently using the steps above.

**WordPress Environment**
WPMU 5.3.2

Others have that problem too!
https://wordpress.org/support/topic/uncaught-wc_data_exception-the-admin-note-content_data-prop-must-be-an-instance/

ObliviousHarmony:
Fixes #3923.

Prior to the release of WC-Admin 0.10.0 the default `content_data` for the `WC_Admin_Note` class was an empty array. This was changed because all usages of the content data expect it to be an stdClass instance. The `WC_Admin_Note::set_content_data()` method throws an exception when passed anything other than an stdClass instance, which leads to a failure when loading notes created with the old default. This PR adds some resiliency here so that loading can proceed as expected.

- **Note: I did add some test cases around this too!**
- Find a notice in the database and change its `content_data` field to `[]`.
- Attempt to view this notice, it should throw an exception when loading without this PR.

### Changelog Note:

Fix - Made the admin note loading more resilient to prevent failures when loading notes with invalid content_data.


ObliviousHarmony:
Made this against the wrong branch and that is one nasty rebase!

PR #3927, opened 2020-03-18
Onboarding: Create flat rate and free shipping methods from task list

### LINKED ISSUE #3761: [4.0 beta1] Shipping On boarding UX confusion
In the new on boarding steps for shipping, when I see "free shipping" I was initially confused as I was assuming that the free shipping method was going to get created

Link to screenshot: https://d.pr/i/h4DH8N

Instead it is a flat rate set to $0.

Also, because of the "free shipping" text, I was hoping that I could include a free shipping **and** be able to add a flat rate in this setup process instead of just one shipping option.

joshuatf:
Fixes #3761

Changes the shipping methods to flat rate or shipping depending on whether or not the shipping zone is set to `0` cost or greater.

This PR also disables all other zones when updating a method, which I think is the best way to handle this without too much confusion.  For example, if you already have a free method and you set a shipping cost to `5.00` in the list, it will create a new flat rate method and disable the free one.  If you update again to `6.00` from the list, it will modify the existing method. /cc @pmcpinto @jameskoster if you have a different opinion on how this should work when previous methods exist.

1. Visit the shipping task.
1. Try adding free (0.00) shipping method costs to your store's zone or the Rest of World zone.
1. Check `wp-admin/admin.php?page=wc-settings&tab=shipping` to make sure methods are updated correctly.
1. Revisit the shipping task and try alternating the costs to be free or a flat rate.
1. Make sure that if a free/flat rate does not exist that one is created and if one already exists, the first one matching the method type (free/flat) is edited.
1. Try adding multiple zones of the same type in `wp-admin/admin.php?page=wc-settings&tab=shipping&zone_id=1`.
1. Revisit the shipping task and make sure that all zones except the one created/edited from the task list is updated. 

### Changelog Note:

Fix: Create flat rate or free shipping methods from the onboarding task list

### DIFF
--- a/client/dashboard/task-list/tasks/shipping/index.js
+++ b/client/dashboard/task-list/tasks/shipping/index.js
@@ -70,7 +70,7 @@ class Shipping extends Component {
 						path: `/wc/v3/shipping/zones/${ zone.id }/methods`,
 					} );
 					zone.name = __( 'Rest of the world', 'woocommerce-admin' );
-					zone.toggleEnabled = true;
+					zone.toggleable = true;
 					shippingZones.push( zone );
 					return;
 				}
--- a/client/dashboard/task-list/tasks/shipping/rates.js
+++ b/client/dashboard/task-list/tasks/shipping/rates.js
@@ -32,13 +32,13 @@ class ShippingRates extends Component {
 		this.updateShippingZones = this.updateShippingZones.bind( this );
 	}
 
-	getShippingMethods( zone ) {
+	getShippingMethods( zone, type = null ) {
+		if ( ! type ) {
+			return zone.methods;
+		}
+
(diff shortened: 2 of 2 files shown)


pmcpinto:
> This PR also disables all other zones when updating a method, which I think is the best way to handle this without too much confusion. For example, if you already have a free method and you set a shipping cost to 5.00 in the list, it will create a new flat rate method and disable the free one. If you update again to 6.00 from the list, it will modify the existing method.

👍sounds like a great option.

joshuatf:
Nice catch, @timmyc!  Added another check to see if a cost exists before trying to grab its value.  This is ready for another review.

timmyc:
maybe a quick rebase to get ci all happy

PR #3929, opened 2020-03-18
Onboarding: Modified the recorded data in profile wizard

octaedro:
This is a fix for the PR: [3730](https://github.com/woocommerce/woocommerce-admin/pull/3730)

In the second step of the profile wizard, `industries_with_detail` will be recorded as a `string` instead of an `object`.

1. Start the onboarding process
2. In the `Industry` step check whether the option `CBD and other hemp-derived products` is visible and works
3. Select the option: `Other`
4. Fill the input with a description.
5. Press the `Continue` button.
6. Check that the value is recorded under `industries_with_detail` value in the `storeprofiler_store_industry_continue` event.

### DIFF
--- a/client/dashboard/profile-wizard/steps/industry.js
+++ b/client/dashboard/profile-wizard/steps/industry.js
@@ -5,7 +5,7 @@ import { __ } from '@wordpress/i18n';
 import { Component, Fragment } from '@wordpress/element';
 import { Button, CheckboxControl } from '@wordpress/components';
 import { compose } from '@wordpress/compose';
-import { filter, get, find, findIndex } from 'lodash';
+import { filter, find, findIndex, forEach, get } from 'lodash';
 import { withDispatch } from '@wordpress/data';
 
 /**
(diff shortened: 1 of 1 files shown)


client/dashboard/profile-wizard/steps/industry.js
```diff
@@ -52,13 +52,19 @@ class Industry extends Component {
...
-		const industriesWithDetail = filter( this.state.selected, ( value ) => {
-			return typeof value.detail !== 'undefined';
+
+		// With this 'forEach' we'll record all the 'industries_with_detail' available.
+		// For now it's only being used for the input 'Other'.
+		forEach( this.state.selected, ( value ) => {
+			if ( typeof value.detail !== 'undefined' ) {
+				recordEvent( 'storeprofiler_store_industry_continue', {
+					industries_with_detail: value.detail,
+				} );
```
jeffstieler: I don't think we should record duplicate events just to capture the "other" fields.

Why not turn the `this.state.selected` values into a comma separated list of string `.detail` values?

octaedro: Sounds good! I'll do it.
Thank you :slightly_smiling_face:

octaedro: Solved in commit [3c0a6ae](https://github.com/woocommerce/woocommerce-admin/pull/3929/commits/3c0a6ae09b190f613ac85e718fdc9e4eca30a03d).
The selected industries details are converted to a string (comma-separated list) to send to the `recordEvent` method

PR #3933, opened 2020-03-18
Add a/b test to enable prompt for 25% of eligible sites.

c-shultz:
Fixes https://github.com/Automattic/woocommerce-shipping-issues/issues/11

Adds A/B test so onboarding prompt enabled for 25% of eligible sites.

This has a 25% of setting the option `woocommerce_shipping_prompt_ab` to `b` which will enable the prompt. A value of `a` will disable the prompt.

We will adjust this percentage and likely turn off the check altogether eventually.

- Make sure your site is configured so the `should_allow_banner()` check passes
- Open order page
- You have a 25% chance of seeing the banner
- Delete `woocommerce_shipping_prompt_ab` option and reload order page to try again

### DIFF
--- a/src/Features/ShippingLabelBannerDisplayRules.php
+++ b/src/Features/ShippingLabelBannerDisplayRules.php
@@ -109,7 +109,7 @@ class ShippingLabelBannerDisplayRules {
 		// If it doesn't exist yet, generate it for later use and save it, so we always show the same to this user.
 		if ( ! $ab_test ) {
 			$ab_test = 1 !== wp_rand( 1, 4 ) ? 'a' : 'b'; // 25% of users. b gets the prompt.
-			update_option( 'woocommerce_shipping_prompt_ab', $ab_test );
+			update_option( 'woocommerce_shipping_prompt_ab', $ab_test, false );
 		}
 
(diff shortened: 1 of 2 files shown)


src/Features/ShippingLabelBannerDisplayRules.php
```diff
@@ -101,6 +101,25 @@ public function __construct( $jetpack_version, $jetpack_connected, $wcs_version,
...
 	public function should_display_banner() {
+		if ( ! $this->should_allow_banner() ) {
+			return false;
+		}
+		$ab_test = get_option( 'woocommerce_shipping_prompt_ab' );
+
+		// If it doesn't exist yet, generate it for later use and save it, so we always show the same to this user.
+		if ( ! $ab_test ) {
+			$ab_test = 1 !== wp_rand( 1, 4 ) ? 'a' : 'b'; // 25% of users. b gets the prompt.
+			update_option( 'woocommerce_shipping_prompt_ab', $ab_test );
```
superdav42:
```suggestion
			update_option( 'woocommerce_shipping_prompt_ab', $ab_test, false );
```
Set third param to false to disable autoload since this is an option we will not use on every page request.

PR #3662, opened 2020-02-04
Use cron for unsnoozing admin notes

jeffstieler:
Fixes #3360.

This PR seeks to replace the Action Scheduler based "unsnooze" functionality with a recurring WP-Cron job.

There has been a lot of confusion and backlash around the Action Scheduler implementation:

* https://github.com/woocommerce/woocommerce-admin/issues/1946
* https://github.com/woocommerce/woocommerce-admin/issues/2364
* https://github.com/woocommerce/woocommerce-admin/issues/3360
* https://github.com/woocommerce/woocommerce-admin/issues/3659

### Detailed test instructions:

- Use WP Crontrol to verify a `wc_admin_unsnooze_admin_notes` cron event exists
- Go to Tools > Scheduled Actions, verify no `wc_admin_unsnooze_admin_notes` events are pending

### Changelog Note:

Tweak: use cron instead of Action Scheduler for unsnoozing notes.

PR #3936, opened 2020-03-18
remove replace_actionscheduler_store_class function 

rrennick:
This PR is based to `master` but also needs to be cherry-picked to `version/1.0`. This removes the custom Action Scheduler data store class.


jeffstieler:
We can probably delete the `WPPostStore` class file too right?

rrennick:
> We can probably delete the WPPostStore class file too right?

Thanks for catching that. Removed.

psealock:
@jeffstieler There is a failing test having to do with the removal of `WPPostStore`, would you be able to take a look?

jeffstieler:
We can fix the failing test for sure.. but..

`PHP Fatal error:  Call to undefined method ActionScheduler_wpPostStore::cancel_actions_by_group()`

The base class for `ActionScheduler_wpPostStore` is `ActionScheduler_Store` which has a `cancel_actions_by_group()` method: https://github.com/woocommerce/action-scheduler/blob/90454ed6381400d2ba02d88578033cc3f8a2abc9/classes/abstracts/ActionScheduler_Store.php#L264

@rrennick - the PHP `5.6` test is using WC `3.8.1` which includes AS `2.2.5`. It appears that AS `2.2.5` is the one in use since `ActionScheduler_Store::cancel_actions_by_group()` was added in `3.0.0`. Should we consider bundling AS 3+ with WooCommerce Admin, or perhaps bumping the minimum WC version to 3.9?

jeffstieler:
Nevermind on bundling AS with WCA. This is ready for another review.

Note that I found some dead code left over from #3662 that I removed as well.

peterfabian:
Side note: WC 3.9 still only includes AS 2.x, so that wouldn't change much.

PR #3939, opened 2020-03-19
Animation slider example fix

jconroy:
Fixes Animation slider example.

### Screenshots

Getting to the end of the pages the navigation would break.
Before: https://d.pr/i/osSm1p
After: https://d.pr/i/V6EBV5

- Pull down this branch
- Build
- Go to AnimationSlider devdocs e.g. `wp-admin/admin.php?page=wc-admin&path=/devdocs&component=animation-slider`
- Try the example - notice the forward/back buttons correctly are disabled and the animations go back and forward through all the correct pages (rather than showing empty pages and not being bale to go back)


jconroy:
Thanks @psealock @joshuatf - rebased as requested and looking good (apologies for the delay)

PR #3950, opened 2020-03-19
Shipping banner load in the proper place

superdav42:
Fixes https://github.com/Automattic/woocommerce-shipping-issues/issues/44
and https://github.com/Automattic/woocommerce-shipping-issues/issues/32
Loads WCS into its proper locations. Also hides wc-admin shipping banner when the modal opens.

### DIFF
--- a/client/wp-admin-scripts/print-shipping-label-banner/index.js
+++ b/client/wp-admin-scripts/print-shipping-label-banner/index.js
@@ -6,6 +6,8 @@ import ShippingBanner from './shipping-banner';
 import './store';
 
 const metaBox = document.getElementById( 'wc-admin-shipping-banner-root' );
+const args =
+	( metaBox.dataset.args && JSON.parse( metaBox.dataset.args ) ) || {};
 
 // Render the header.
-render( <ShippingBanner />, metaBox );
+render( <ShippingBanner itemsCount={ args.shippable_items_count } />, metaBox );
--- a/client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
+++ b/client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
@@ -264,6 +264,9 @@ export class ShippingBanner extends Component {
 					),
 				},
 			} );
+			document.getElementById(
+				'woocommerce-admin-print-label'
+			).style.display = 'none';
 		}
 	}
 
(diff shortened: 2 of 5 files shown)


harriswong:
👀

client/wp-admin-scripts/print-shipping-label-banner/index.js
```diff
@@ -6,6 +6,8 @@ import ShippingBanner from './shipping-banner';
 import './store';
 
 const metaBox = document.getElementById( 'wc-admin-shipping-banner-root' );
+const args =
+	( metaBox.dataset.args && JSON.parse( metaBox.dataset.args ) ) || {};
 
 // Render the header.
-render( <ShippingBanner />, metaBox );
+render( <ShippingBanner itemsCount={ args.shippable_items_count } />, metaBox );
```
harriswong: It's back!

client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
```diff
@@ -149,6 +149,26 @@ export class ShippingBanner extends Component {
...
+
+		return `
+<div id="woocommerce-order-label" class="postbox ">
+	<button type="button" class="handlediv" aria-expanded="true">
+		<span class="screen-reader-text">${ togglePanelText } ${ title }</span>
+		<span class="toggle-indicator" aria-hidden="true"></span>
+	</button>
+	<h2 class="hndle"><span>${ title }</span></h2>
+	<div class="inside">
+		<div class="wcc-root woocommerce wc-connect-create-shipping-label" data-args="${ argsJsonString }">
```
harriswong: Since `generateMetaBoxHtml()` is rendering HTML. I was thinking maybe we could change  `generateMetaBoxHtml()` into a React component which it will render the same HTML by passing the parameter as props in.

The tricky part is how we could do `.insertAdjacentHTML( 'beforebegin', shippingLabelContainerHtml );` in React. 

I think having composited components in React-land is better than inserting raw HTML directly into the DOM. But even if we have the component, we still need to do `.insertAdjacentHTML()` kinda thing to insert these meta box directly in the DOM. I am not sure what benefits we will get by converting these HTML to a component.

I am 👍  with this but you may want others' opinions.

superdav42: I think the best option is actually to load the html via an ajax endpoint. That way any changes in the future in the WCS codebase would be reflected here. That would mean we would need another release of WCS and that seems like too much.

client/wp-admin-scripts/print-shipping-label-banner/index.js
superdav42: turns out we needed it after all

PR #3951, opened 2020-03-19
Keep "Create shipping label" busy until asset is done loading

harriswong:
Fixes https://github.com/Automattic/woocommerce-shipping-issues/issues/42

Currently the `isBusy` prop on the "Create shipping label" button relies on the `isRequesting` status from `wc-api`. This status means either installation or activation is in progress.

What we want to do is to have the "Create shipping label" continue to spin even during the assets loading state. 

Change button to rely on a new state `isShippingLabelButtonBusy`. This state is set to `false` only if `loadWcsAssets` finishes loading. 

Note, this state is set outside of `openWcsModal` so there is a brief moment the button goes back to "enable" before the modal is popped up. This should be fixed by https://github.com/woocommerce/woocommerce-admin/pull/3950/files since the panel will be replaced by WCS panel.

If we want the button to look busy until the WCS modal pops up without the slight moment of "enabled" mentioned above, we could just keep the `isBusy` set to true once  "Create shipping label" is clicked, and only set it back to `false` upon error. 

### Accessibility

1. Deactivate WCS
2. Delete WCS
3. Click "Create shipping label"
4. The button should continue to show as "busy" until the modal shows up.

### DIFF
--- a/client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/test/index.js
+++ b/client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/test/index.js
@@ -71,6 +71,7 @@ describe( 'Tracking clicks in shippingBanner', () => {
 			<ShippingBanner
 				isJetpackConnected={ isJetpackConnected }
 				activePlugins={ activePlugins }
+				installPlugins={ jest.fn() }
 				activatedPlugins={ [] }
 				installedPlugins={ [] }
 				wcsPluginSlug={ 'woocommerce-services' }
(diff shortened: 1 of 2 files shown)


client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/test/index.js
```diff
@@ -172,16 +173,6 @@ describe( 'Create shipping label button', () => {
...
-	it( 'should show a busy loading state when installing or activating ', () => {
-		shippingBannerWrapper.setProps( {
-			isRequesting: true,
-		} );
-		const createShippingLabelButton = shippingBannerWrapper.find( Button );
-		expect( createShippingLabelButton.length ).toBe( 1 );
-		expect( createShippingLabelButton.prop( 'disabled' ) ).toBe( true );
-		expect( createShippingLabelButton.prop( 'isBusy' ) ).toBe( true );
-	} );
-
```
harriswong: Testing this in line 264 `describe( 'In the process of installing, activating, loading assets for WooCommerce Service', () => {` test case instead.

c-shultz:
I'm going to go ahead and merge so I can clear up the conflicts with my branch :)

harriswong:
Thank you!

PR #3974, opened 2020-03-23
Onboarding: Remove plugins step from profiler

### LINKED ISSUE #3894: Onboarding: Remove plugin installation step from profiler
From discussion in p6riRB-5li-p2 we should remove the final plugins step from the profiler and install plugins in the background.

Following the benefits screen, the installation should then occur in the background if the user has opted into the plugin installation.

https://github.com/woocommerce/woocommerce-admin/pull/3888 will introduce inbox notices if installation fails, prompting the user to re-attempt installation.

/cc @pmcpinto if you have any additional thoughts to add to this issue.

Screen Shot 2020-03-13 at 2 00 46 PM

joshuatf:
Fixes #3894

Removes the plugins step from the profiler and installs plugins directly from the benefits page.  Note that this does not include error handling as this is being handled in https://github.com/woocommerce/woocommerce-admin/pull/3888

1. Disable WCS and/or Jetpack.
1. Enable and walk through the profiler until the benefits step. `wp-admin/admin.php?page=wc-admin&step=benefits`
1. Try opting to install plugins and not install them.
1. Make sure plugins are installed and activated correctly and you're redirected to the task list afterwards instead of the previous plugins screen.


psealock:
Hi @joshuatf, code is looking good so far. I don't understand this step:

```
Try opting to install plugins and not install them.
```
I clicked "Yes please" and it went ahead and installed correctly. Can you clarify?

joshuatf:
>I clicked "Yes please" and it went ahead and installed correctly. Can you clarify?

Sorry for the confusion, @psealock.  I just meant try both options, "Yes please" and "No thanks" to make sure both actions complete correctly.

While we're at it, it's probably worth throwing an error in https://github.com/woocommerce/woocommerce-admin/blob/a306a39eafe268b57420c59162d16001c5299c6d/src/API/OnboardingPlugins.php#L272

becdetat:
I've just ran this, everything looks good except for the 'Yes please!' button going busy when selecting 'No thanks'.

joshuatf:
@becdetat @psealock Thanks for the feedback!

I took a look and it seems like both buttons get the busy state, but the `default` button styling might need some changes to actually show a busy state.  I'll log a follow-up issue around that.

But regarding this and future PRs, I wonder if we should show both buttons as busy or just the one clicked.  I opened some random apps and found both approaches being used.  Not sure if there's a "correct" standard of doing this.  /cc @jameskoster

jameskoster:
> Almost feels like the solution there might be to disable the button that didn't get clicked.

+1

joshuatf:
@timmyc @becdetat @psealock I went ahead and disabled the buttons when actioned in this PR.  Also simplified the state a bit.

Can I get one more sanity check around button states on this to make sure things work as expected?

PR #3975, opened 2020-03-23
Hide the WC Shipping banner until the create shipping label popup is open

Ferdev:
Fixes https://github.com/Automattic/woocommerce-shipping-issues/issues/51

Hides the WooCommerce Services banner when the "create shipping label" button is pressed. Once WooCommerce Services is installed, activated and the create label popup is open, the WooCommerce Services banner is displayed again.

### DIFF
--- a/client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
+++ b/client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
@@ -173,13 +173,13 @@ export class ShippingBanner extends Component {
 			.catch( () => this.setState( { wcsSetupError: true } ) );
 	}
 
-	generateMetaBoxHtml( title, args ) {
+	generateMetaBoxHtml( nodeId, title, args ) {
 		const argsJsonString = JSON.stringify( args ).replace( /"/g, '&quot;' ); // JS has no native html_entities so we just replace.
 
 		const togglePanelText = __( 'Toggle panel:', 'woocommerce-admin' );
 
 		return `
-<div id="woocommerce-order-label" class="postbox ">
+<div id="${ nodeId }" class="postbox">
 	<button type="button" class="handlediv" aria-expanded="true">
 		<span class="screen-reader-text">${ togglePanelText } ${ title }</span>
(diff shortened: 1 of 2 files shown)


psealock:
>Could we use a shared data store between WCS and shipping-banner? So shipping banner can load the same store and capture a "loaded" state change from WCS.

I don't have too much knowledge on the logic at play here, but this seems like a sensible approach. https://github.com/woocommerce/woocommerce-admin/pull/3313 will introduce importable wp.data data stores that will be accessible to wc-admin and plugins to share.

The use of MutationObserver seems to indicate there's probably a better approach. I'm having WCS banner/modal confusion so I'm not following what exactly is happening, so maybe it is the only approach (at least at the moment).

Ferdev:
I wasn't sure about subscribing to the WooCommerce Services store. But I gave it a try and looks like it works pretty good. Please @harriswong , give it another try and let me know what you think!

client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
```diff
@@ -293,6 +297,26 @@ export class ShippingBanner extends Component {
...
+				if ( siteState ) {
+					const orderState = siteState.shippingLabel[ orderId ];
+					if ( orderState ) {
+						if ( orderState.showPurchaseDialog ) {
+							if ( window.jQuery ) {
+								window
+									.jQuery( '#woocommerce-order-label' )
+									.show();
+							}
+							wcsStoreUnsubscribe();
```
harriswong: Since this is called inside `subscribe()` wouldn't this introduce infinite recursion? 

I took a look at https://redux.js.org/api/store/#subscribelistener point 2 and it says that the unsubscribe won't have any effect until the next dispatch call, so I think this is ok!

harriswong: Yep, breakpoint only got called once. 👍

Ferdev: yeah, the docs mention that's the way to unsubscribe from a redux store :smiley:

PR #3976, opened 2020-03-23
Add flex to banner to center vertical text alignment

harriswong:
Fixes https://github.com/Automattic/woocommerce-shipping-issues/issues/50

Align text to center. From:
https://user-images.githubusercontent.com/572862/77210358-7065a580-6ac6-11ea-812b-0c9d2022b2e3.png

To:
image

### Accessibility

### Screenshots
image
image
image
image

- Open order page
- Resize window to make sure text is vertically align and visible. Should match above screenshots./

### DIFF
--- a/client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
+++ b/client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
@@ -334,6 +334,38 @@ export class ShippingBanner extends Component {
 						src={ wcAdminAssetUrl + 'shippingillustration.svg' }
 						alt={ __( 'Shipping ', 'woocommerce-admin' ) }
 					/>
+					<div className="wc-admin-shipping-banner-blob">
+						<h3>
+							{ __(
+								'Print discounted shipping labels with a click.',
+								'woocommerce-admin'
+							) }
+						</h3>
+						<p>
+							{ interpolateComponents( {
+								mixedString: this.state.installText,
+								components: {
+									tosLink: (
+										<ExternalLink
+											href="https://wordpress.com/tos"
+											target="_blank"
+											type="external"
+										/>
+									),
+									wcsLink: (
+										<ExternalLink
+											href="https://woocommerce.com/products/shipping/"
+											target="_blank"
+											type="external"
+											onClick={
+												this.woocommerceServiceLinkClicked
+											}
+										/>
+									),
+								},
+							} ) }
+						</p>
+					</div>
 					<Button
 						disabled={ isShippingLabelButtonBusy }
 						isPrimary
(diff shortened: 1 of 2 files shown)


client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
```diff
@@ -333,6 +333,39 @@ export class ShippingBanner extends Component {
 						src={ wcAdminAssetUrl + 'shippingillustration.svg' }
 						alt={ __( 'Shipping ', 'woocommerce-admin' ) }
 					/>
+					<div className="wc-admin-shipping-banner-blob">
```
harriswong: Looks like a huge change but I added only the `<div>` to wrap the text.

client/wp-admin-scripts/print-shipping-label-banner/style.scss
```diff
@@ -8,26 +8,30 @@
...
+
+	.wc-admin-shipping-banner-container {
+		display: flex;
+		align-items: center;
+		margin-top: 18px;
+	}
+
 	.wc-admin-shipping-banner-illustration {
-		float: left;
-		margin: -2px 8px 36px -2px;
```
harriswong: Removed `float` and `other margin because I am relying on flex to align them.

client/wp-admin-scripts/print-shipping-label-banner/style.scss
```diff
@@ -8,26 +8,30 @@
...
-		max-width: 75%;
 		margin: 2px 0 11px;
 		font-size: 15px;
 		line-height: 19px;
 	}
 
 	.components-button.is-button {
-		float: right;
-		margin: 18px 61px 10px 25px;
+		margin: 10px 61px 10px auto;
```
harriswong: Using `margin-right: auto` to push the button to the right in flex.

client/wp-admin-scripts/print-shipping-label-banner/style.scss
```diff
@@ -119,19 +123,17 @@
...
+		}
+
 		.components-button.is-button {
-			float: none;
-			left: 50%;
-			-webkit-transform: translateX(-50%);
-			transform: translateX(-50%);
-			position: absolute;
-			bottom: -30px;
-			margin: 0;
```
harriswong: Relying on `flex-wrap: wrap;` instead.

Ferdev:
Looks good! I wonder if the mobile version is looking the way it should. @elizaan36 can you confirm the last screenshot is good? Thanks!

harriswong:
@elizaan36 I updated the vertical alignment of the paragraph in the banner. Please review!

elizaan36:
Thanks @Ferdev, looks like the banner on mobile has extra padding on the bottom for some reason. Don't know if that's a quick CSS tweak or not, but maybe we could somehow make the content centered in the card so there isn't that extra white space below.

elizaan36:
Vertical alignment of the paragraph on the larger screen size looks good @harriswong

harriswong:
@elizaan36 @Ferdev I removed the bottom padding in mobile view. Please take a look again.

image

image

elizaan36:
Nice @harriswong ! A question + clarification:

- Is the X to dismiss positioned absolute in the banner? Looks like it's pushing the rest of the content to the left as if it's it's own div
- I should have specified _how much_ padding, my bad. What I'd expect to see is equal padding on the top and bottom. Is there any way to match the padding on the top? There's just a bit more space there.

Thanks again, I'm sure we'll be gtg after this.

harriswong:
@elizaan36 Good catch! 

> Is the X to dismiss positioned absolute in the banner?
Yes it is still absolute, but I added `margin-right` on the paragraph for some reasons so the paragraph looks distorted in mobile view. Removing that.

>  Is there any way to match the padding on the top? 
Yep! For sure, let me fix that.

harriswong:
I adjusted the bottom to match the top. 

Desktop view: 
image

For mobile:
image

PR #3977, opened 2020-03-23
Reduce queries in Inbox Panel.

jeffstieler:
This PR seeks to improve the performance of the Inbox Panel by:
* Only requesting fields necessary for the component render/behavior
* Avoid `WC_Data`'s meta retrieval since it's errant anyhow (collisions with post table object IDs)

## Detailed test instructions:
- Verify the Inbox displays properly
- Verify that note actions work as expected - including simple links and actions (like installing plugins)

## Changelog Note:

Performance: trim down inbox note API request.

### DIFF
--- a/client/header/activity-panel/panels/inbox/index.js
+++ b/client/header/activity-panel/panels/inbox/index.js
@@ -126,6 +126,18 @@ export default compose(
 			orderby: 'date',
 			order: 'desc',
 			status: 'unactioned',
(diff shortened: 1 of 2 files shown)


rrennick (approved):
@jeffstieler In my testing, on `master` the combined load time was 1898 ms. With this PR 1337 ms.  Nice improvement 🚢

PR #3979, opened 2020-03-23
Don't change the banner text during WCS installation/activation

harriswong:
Fixes https://github.com/Automattic/woocommerce-shipping-issues/issues/52

When the user click "Create shipping label" button on the banner, the banner message should not change. 

### Accessibility

- Deactivate WooCommerce Services
- Delete WooCommerce Services 
- Go to an order page
- Click "Create shipping label", the banner will now install and activate WCS
- This message should not change throughout:
image

###

### DIFF
--- a/client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
+++ b/client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
@@ -39,6 +39,7 @@ export class ShippingBanner extends Component {
 			wcsAssetsLoading: false,
 			wcsSetupError: false,
 			isShippingLabelButtonBusy: false,
+			installText: this.getInstallText(),
 		};
 	}
 
@@ -349,7 +350,7 @@ export class ShippingBanner extends Component {
 					</h3>
 					<p>
 						{ interpolateComponents( {
-							mixedString: this.getInstallText(),
+							mixedString: this.state.installText,
 							components: {
 								tosLink: (
 									<ExternalLink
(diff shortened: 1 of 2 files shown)


client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
```diff
@@ -349,7 +350,7 @@ export class ShippingBanner extends Component {
 					</h3>
 					<p>
 						{ interpolateComponents( {
-							mixedString: this.getInstallText(),
+							mixedString: this.state.installText,
```
harriswong: Since `getInstallText()` function return text based on the `activePlugins` props, the text will get updated when `wc-api` re-render by updating `activePlugins` prop.

I could add logics with this state `isShippingLabelButtonBusy` inside `getInstallText()` , so that it will return different text depending on if the "Create shipping label" has been clicked or not.

Instead, I decided to add a local state `getInstallText` that is initialized once at constructor. This state will not be updated during re-render so our message will not change throughout.

client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/test/index.js
```diff
@@ -383,3 +383,75 @@ describe( 'Setup error message', () => {
...
+
+	it( 'should show install text "By clicking "You\'ve already installed WooCommerce Shipping."..." when WooCommerce Service is already installed.', () => {
+		const shippingBannerWrapper = createShippingBannerWrapper( {
+			activePlugins: [ 'woocommerce-services' ],
+			installedPlugins: [ 'woocommerce-services' ],
+		} );
+		const createShippingLabelText = shippingBannerWrapper.find( 'p' );
+		expect( createShippingLabelText.text() ).toBe( activatedMessage );
+	} );
+} );
```
Ferdev: :heart:

PR #3953, opened 2020-03-20
Marketing Tab MVP

jconroy:
Work in progress PR for the Marketing Tab MVP. 

We envisage splitting this out into separate PRs for easier review.

## Background

In WooCommerce 4.1 we are hoping to make _marketing_ a “first-class citizen” in WooCommerce. 

The PR introduces a new dashboard tab that we hope will grow to become a robust marketing hub within WooCommerce. Our goal is to bring order to the current disjointed marketing experience our merchants are getting, giving them access a dashboard that manages, controls and guides their experience marketing their WooCommerce store.

An initial design for the MVP...

TJxWH-uVD30YNNYe2Wp0W1ElTn8jtiwvfXYiRhn9nClgJieq07bnAHOAXuNvt85CgRrkrFRIvRzzCZgpNSATNXA7nC2ukHkWh2_xh2WZYLr8OULzPAlmiBG8fRACt6xxkfTOUFV3

Our focus has been on the following three main areas to start and from which we we build on:

* **Simplification:** the experience of installing and then actually using most plugins in WooCommerce can be disjointed. As of now, we are creating a central place where merchants can set up and manage their installed marketing extensions. 
* **Curation:** finding an extension that says it can help you market your store is not hard, but finding the ones that actually do a great job is much harder. We will be providing a limited set of recommendations to merchants to help get them started on the right foot.
* **Education:** learning through trial and error is a great way to learn - but by surfacing a range of evergreen articles and guides on marketing our aim is to help accelerate our merchants marketing skills and impact.  

We will be testing the impact of this new marketing dashboard by integrating with a small selection of extensions. Based on usage and alongside merchant feedback, we will no doubt be making iterations and changes as required and then may look to open these areas up for further integration, while maintaining our goals of simplification, curation, and education.

---

## High Level Specs

For this MVP the tab will consist of 4 key components.

* Welcome Card
* Installed Extensions
* Recommended Extensions
* WooCommerce knowledge base

### Welcome Card

Welcome Card

A simple component that displays a welcome image and some text that can be permanently dismissed.

### Installed marketing extensions

Installed Extensions

The most complicated of the components. This component (for the MVP) provides a centralized place to access relevant links for:
* Facebook for WooCommerce Extension
* MailChimp for WooCommerce Extension

Both extensions feature as part of the onboarding wizard.

If neither of the extensions are installed - the card will not be shown.

From the card merchants will be able to quickly access support, settings, and or dashboards.

Installed marketing extensions would have one of three states:

[...]

PR #3980, opened 2020-03-24
Add showPagePicker, showPerPagePicker props to Pagination component

danielbitzer:
I would like to use the Pagination component in the slider I'm working on for the marketing tab (#3953). Rather than duplicate this functionality we can gain a lot by using the existing Pagination component. All we need is to hide the parts we don't need. 

This this RP adds 2 new props `showPagePicker` and `showPerPagePicker` which remove parts of the pagination component that I don't want for the marketing slider.

Screenshot

The PR adds a story for the Pagination component in 2577879 which you can use to try out the new properties. 

60NOwo

### Accessibility

- [x] All text has [at least a 4.5 color contrast with its background](https://webaim.org

- Fire up storybook e.g. `npm run storybook`
- Go to the Pagination component
- Enable/disable pickers using the checkboxes for "Page Picker" and "Per Page Picker"

### DIFF
--- a/packages/components/src/pagination/README.md
+++ b/packages/components/src/pagination/README.md
@@ -26,3 +26,5 @@ Name | Type | Default | Description
 `onPerPageChange` | Function | `noop` | A function to execute when the per page option is changed
 `total` | Number | `null` | (required) The total number of results
 `className` | String | `null` | Additional classNames
+`showPagePicker` | Boolean | `true` | Whether the page picker should be shown.
+`showPerPagePicker` | Boolean | `true` | Whether the per page picker should shown.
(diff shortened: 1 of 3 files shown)


haszari:
> If you create a stories folder, ie https://github.com/woocommerce/woocommerce-admin/tree/master/packages/components/src/rating/stories, then npm run storybook will render the component and allow change of props for testing.

Great idea, sounds right to me @psealock ! 

@danielbitzer Ping me if you get stuck with this, we have good time zone crossover.

Also should be mentioned – in blocks we now have [storybook built & published on every push to master](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2006) (check it out [here](https://woocommerce.github.io/woocommerce-gutenberg-products-block/)), looks like it's fairly simple to set up (props @nerrad). Would be good to add similar to this repo.

jconroy:
👋 @psealock I've:
 
* added a Story with 2577879
* updated the PR description to reflect the additional change
* added test instructions

danielbitzer:
> Should we add another prop to hide/show "Page 2 of 10"? Are you using CSS to hide it in the banner?

I considered this. What would be you preference @psealock? If it existed we would use it.

jconroy:
Went ahead and added a showPageArrowsLabel prop. 

Also rebased on master to pick up the storybook fix in https://github.com/woocommerce/woocommerce-admin/pull/4001

PR #3983, opened 2020-03-24
Fix "Back to..." link disappearing after editing a term.

### LINKED ISSUE #3964: Back to product categories disappears
**Describe the bug**
After working within a product category the "back to categories" button near top disappears.

**To Reproduce**
1. Go to 'product categories'
2. Click on 'customize category'
3. SAVE
4. Try to navigate back with <-- back to categories  ---- disappears

**Screenshots**

**Expected behavior**

- [ ] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [ ] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/storefront/).
- [ ] I can reproduce this bug consistently using the steps above.

**WordPress Environment**
<details>
```
Copy and paste the system status report from **WooCommerce > System Status** in WordPress admin.
```
</details>

Konamiman:
After editing a taxonomy term (e.g. a category or a tag), a success message with a "Back to <taxonomy>es" link appears, but it disappears after a couple of seconds.

The problem is that `Loader.php` defines an `admin_notices` that wraps all notices in a hidden div. This div includes another one with class `wp-header-end`. There's a piece of code in `common.js` in WP core that moves all notices inside that inner div, so as soon as that js executes, the notice is hidden.

Notice divs having the `inline` class are not moved, therefore the fix consists of a one-line JavaScript that adds that class to the notice div before `common.js` executes.

Fixes #3964

1. Open the WooCommerce categories page (Products > Categories)
2. Open the details for a category or create a new one, save changes.
3. Verify that the "Item updated" with "Back to categories" link doesn't disappear.

### Screenshots

image

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -78,6 +78,7 @@ class Loader {
 		add_action( 'admin_head', array( __CLASS__, 'remove_notices' ) );
 		add_action( 'admin_notices', array( __CLASS__, 'inject_before_notices' ), -9999 );
 		add_action( 'admin_notices', array( __CLASS__, 'inject_after_notices' ), PHP_INT_MAX );
+		add_action( 'admin_head', array( __CLASS__, 'make_success_and_error_notices_inline' ) );
 
 		// priority is 20 to run after https://github.com/woocommerce/woocommerce/blob/a55ae325306fc2179149ba9b97e66f32f84fdd9c/includes/admin/class-wc-admin-menus.php#L165.
 		add_action( 'admin_head', array( __CLASS__, 'remove_app_entry_page_menu_item' ), 20 );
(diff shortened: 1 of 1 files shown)


rrennick:
@Konamiman Thanks for submitting the PR. WooCommerce Admin does not currently have any jQuery code in it. This problem should be solvable without introducing jQuery into a React application. I'd suggest looking at adding conditional logic to

> The problem is that Loader.php defines an admin_notices that wraps all notices in a hidden div.

Konamiman:
Closing because https://github.com/woocommerce/woocommerce-admin/pull/4019 fixes the same bug in a better way.

PR #3994, opened 2020-03-25
remove use of undefined QUEUE_GROUP

### LINKED ISSUE #3982: Update routine uses undefined constant
**Describe the bug**
```
CRITICAL Uncaught Error: Undefined class constant ‘QUEUE_GROUP’ in /wp-content/plugins/woocommerce/packages/woocommerce-admin/includes/wc-admin-update-functions.php:70
```

The `QUEUE_GROUP` was removed in a previous PR. This error would only occur when a site with a 0.24 or older WCA deactivated with WC4.0.X installed (or WCA 1.0.X).

**Expected behavior**
Update the function to use either the group string or pull it from the current sync code. Since this action is no longer scheduled and all existing instances of the action use the same group, the string should be fine.

rrennick:
Fixes #3982

This PR replaces the use of the constant with a string literal. It also updates the update action check function to only check for `pending` or `completed` actions. The default queue search retrieves actions with any status. It took a while for me to reproduce the issue for testing because I had a `failed` instance of the remove unsnooze update action from other testing I had been doing with AS.

- delete any completed or pending `woocommerce_run_update_callback` actions with  `wc_admin_unsnooze_admin_notes` as the callback
- install WC 3.9.3
- switch to WC Admin to 0.24.0
- update `woocommerce_admin_version` option to 0.24.0
- ensure `wc_admin_unsnooze_notes` action is created
- update WC to 4.0.1
- deactivate WCA
- ensure WCA update action from the first step runs
- check that `wc_admin_unsnooze_notes` action is deleted
- verify no warnings or errors in the debug log

### Changelog Note:

Fix: Eliminate warning on deleting unsnooze note action

### DIFF
--- a/includes/wc-admin-update-functions.php
+++ b/includes/wc-admin-update-functions.php
@@ -68,6 +68,7 @@ function wc_admin_update_0230_db_version() {
  */
 function wc_admin_update_0251_remove_unsnooze_action() {
 	as_unschedule_action( WC_Admin_Notes::UNSNOOZE_HOOK, null, 'wc-admin-data' );
+	as_unschedule_action( WC_Admin_Notes::UNSNOOZE_HOOK, null, 'wc-admin-notes' );
 }
 
 /**
--- a/includes/wc-admin-update-functions.php
+++ b/includes/wc-admin-update-functions.php
@@ -67,7 +67,7 @@ function wc_admin_update_0230_db_version() {
  * Remove the note unsnoozing scheduled action.
  */
 function wc_admin_update_0251_remove_unsnooze_action() {
-	as_unschedule_action( WC_Admin_Notes::UNSNOOZE_HOOK, null, WC_Admin_Notes::QUEUE_GROUP );
+	as_unschedule_action( WC_Admin_Notes::UNSNOOZE_HOOK, null, 'wc-admin-data' );
 }
 
 /**
--- a/src/Install.php
+++ b/src/Install.php
@@ -396,12 +396,23 @@ class Install {
 						array(
 							'per_page' => 1,
 							'hook'     => 'woocommerce_run_update_callback',
-							'search'   => json_encode( array( $update_callback ) ),
+							'search'   => wp_json_encode( array( $update_callback ) ),
 							'group'    => 'woocommerce-db-updates',
+							'status'   => 'pending',
 						)
 					);
 
-					if ( empty( $pending_jobs ) ) {
+					$complete_jobs = WC()->queue()->search(
+						array(
+							'per_page' => 1,
+							'hook'     => 'woocommerce_run_update_callback',
+							'search'   => wp_json_encode( array( $update_callback ) ),
+							'group'    => 'woocommerce-db-updates',
+							'status'   => 'complete',
+						)
+					);
+
+					if ( empty( $pending_jobs ) && empty( $complete_jobs ) ) {
 						WC()->queue()->schedule_single(
 							time() + $loop,
 							'woocommerce_run_update_callback',


becdetat:
> ensure wc_admin_unsnooze_notes action is created

@rrennick how do I get this action to be created? I've installed and activated WC 3.9.3 and  WCA 0.24.0 and I can't see any new actions.

rrennick:
@becdetat Did you update your `woocommerce_admin_version`? The action should be automatically created within a few dashboard page loads.

src/Install.php
```diff
@@ -396,12 +396,23 @@ private static function update() {
...
+						array(
+							'per_page' => 1,
+							'hook'     => 'woocommerce_run_update_callback',
+							'search'   => wp_json_encode( array( $update_callback ) ),
+							'group'    => 'woocommerce-db-updates',
+							'status'   => 'complete',
+						)
+					);
+
+					if ( empty( $pending_jobs ) && empty( $complete_jobs ) ) {
```
jeffstieler: Why was this change needed? Are we actually cleaning up completed jobs?

includes/wc-admin-update-functions.php
```diff
@@ -67,7 +67,7 @@ function wc_admin_update_0230_db_version() {
  * Remove the note unsnoozing scheduled action.
  */
 function wc_admin_update_0251_remove_unsnooze_action() {
-	as_unschedule_action( WC_Admin_Notes::UNSNOOZE_HOOK, null, WC_Admin_Notes::QUEUE_GROUP );
+	as_unschedule_action( WC_Admin_Notes::UNSNOOZE_HOOK, null, 'wc-admin-data' );
```
joshuatf, line 70: Optional, but we're using `wc-admin-data` in a few places now, so it would be nice to have a constant somewhere for this.

src/Install.php
joshuatf: Same question.  I *think* this works because if a job has recently completed with the same name, then we don't need to run it again.

If this is the case, I think we could remove the `status` property entirely and just search the queue once.  The API makes it seem like there are only the options of `complete` and `pending` so we could just not limit by status.

joshuatf: >The API makes it seem like there are only the options of complete and pending so we could just not limit by status.

To add on this, there are other statuses used, so I'm not 100% sure if the `search()` method only returns `complete` and `pending` as the docs seem to suggest or if it will include all.

rrennick: > the docs seem to suggest or if it will include all.

Yes, including `failed` and `canceled`. We want to schedule the update if the existing action has either of those statuses.

includes/wc-admin-update-functions.php
jeffstieler: I only see it in https://github.com/woocommerce/woocommerce-admin/blob/54b037501de15c732b5134cc11090fade35746f8/src/Schedulers/SchedulerTraits.php#L21.

I think it's OK to use the string literal here as we're just cleaning up - this value shouldn't/won't/can't change.

jeffstieler, line 70: It looks like this should be `wc-admin-notes`:

https://github.com/woocommerce/woocommerce-admin/blob/fe7b6cca7a095d1a39043bd1b38fc055cf0b121d/src/Notes/WC_Admin_Notes.php#L25

rrennick: @jeffstieler Originally it was scheduled under `wc-admin-data`. At some point the group was changed to `wc-admin-notes` which would mean that older installs had two of the unsnooze actions running (totalling 1440 over 30 days). That would have contributed to users complaining about the number. I added a second call to remove from the second group.

PR #3996, opened 2020-03-25
Fix/remove address alerts 55

Ferdev:
Fixes https://github.com/Automattic/woocommerce-shipping-issues/issues/55.

The origin & destination addresses in the WC Services create label modal, were showing some red notifications. The reason was the modal was open before it was fully loaded. This PR takes care of that and makes the banner wait until the extension is fully loaded to open the modal.

### DIFF
--- a/client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
+++ b/client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
@@ -300,6 +300,7 @@ export class ShippingBanner extends Component {
 
 			const wcsStoreUnsubscribe = wcsStore.subscribe( () => {
 				const latestState = wcsStore.getState();
+
 				const siteState =
 					latestState.extensions.woocommerce.woocommerceServices[
 						siteId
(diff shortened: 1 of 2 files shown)


client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
```diff
@@ -300,40 +300,61 @@ export class ShippingBanner extends Component {
...
 						siteId
 					];
 				if ( siteState ) {
 					const orderState = siteState.shippingLabel[ orderId ];
 					if ( orderState ) {
-						if ( orderState.showPurchaseDialog ) {
+						if (
+							! orderState.showPurchaseDialog &&
+							! orderState.isFetching
+						) {
```
Ferdev: We open the modal only if the modal is not already open, and it the modal is down getting data from WC Connect Server.

client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
```diff
@@ -300,40 +300,61 @@ export class ShippingBanner extends Component {
...
+									),
+								},
+							} );
+						} else if ( orderState.showPurchaseDialog ) {
 							if ( window.jQuery ) {
 								window
 									.jQuery( '#woocommerce-order-label' )
 									.show();
 							}
 							wcsStoreUnsubscribe();
```
Ferdev: These actions are the ones triggered when the user press "Use address as entered" in the address step of the create shipping label modal.

client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
```diff
@@ -300,40 +302,93 @@ export class ShippingBanner extends Component {
...
+					labelSettingsState &&
+					labelSettingsState.meta &&
+					packageState &&
+					locationsState
+				) {
+					if (
+						shippingLabelState.loaded &&
+						labelSettingsState.meta.isLoaded &&
+						packageState.isLoaded &&
+						isArray( locationsState ) &&
```
Ferdev: When all these flags are true, the shipping label modal has been correctly instantiated.

client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
```diff
@@ -300,40 +302,93 @@ export class ShippingBanner extends Component {
...
+						! this.state.isWcsModalOpen
+					) {
+						if ( window.jQuery ) {
+							this.setState( { isWcsModalOpen: true } );
```
Ferdev: When the "create shipping label" button is clicked, doesn't appear to be an immediate state that lets us know if the button has been clicked or not, so we store that info in the component's state to avoid opening the modal multiple times.

Ferdev:
@superdav42 this PR can be reviewed again. The component now triggers the click event on the "create shipping label" button. Let me know how it works for you.

PR #4000, opened 2020-03-25
WIP: Lazy load order stats fields.

jeffstieler:
WIP.

* Order Stats: move coupon join into sql filter function.

### Accessibility

- Ex: Open page `url`
- Click XYZ…


jeffstieler:
Supplanted by https://github.com/woocommerce/woocommerce-admin/pull/4009

PR #4008, opened 2020-03-26
Hide scrollbars in dismiss modal

Ferdev:
Fixes https://github.com/Automattic/woocommerce-shipping-issues/issues/58.

Steps to test it in the linked issue.

This PR is just a workaround. It's actually an issue in `@wordpress/components`. I created a test site that reproduces the issue: https://github.com/Ferdev/wordpress-components-modal-issue

I'll create an issue in the Gutenberg repository to follow up.

### DIFF
--- a/client/wp-admin-scripts/print-shipping-label-banner/style.scss
+++ b/client/wp-admin-scripts/print-shipping-label-banner/style.scss
@@ -1,6 +1,4 @@
 // This is needed so ExternalLinks appear correctly. Should be part of 'wp-components' style but for some reason it's not.
-@import '@wordpress/components/src/visually-hidden/style';
-
 #woocommerce-admin-print-label {
 	min-height: 100px;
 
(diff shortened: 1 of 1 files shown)


client/wp-admin-scripts/print-shipping-label-banner/style.scss
```diff
@@ -1,6 +1,4 @@
 // This is needed so ExternalLinks appear correctly. Should be part of 'wp-components' style but for some reason it's not.
-@import '@wordpress/components/src/visually-hidden/style';
```
superdav42: Removing this like causes problems with the ExternalLink component. They now look like:
Screenshot from 2020-03-26 15-35-48

superdav42: `opens in a new tab` should be hidden. Was there a reason to remove it to fix the issue here?

Ferdev: not at all, sorry about that Dave, I removed that line by accident. Thanks for catching this!

PR #4010, opened 2020-03-26
Position error message directly under the paragraph

harriswong:
Closes https://github.com/Automattic/woocommerce-shipping-issues/issues/60

Reposition `SetupNotice` so that it doesn't flow to the back of the flex box. Instead, move the `SetupNotice` into the text bulb div.

### Accessibility

- Open up the order page
- Make the error show up. If there is no error, you can open `woocommerce-admin/client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js`, go to line 394 and set `isSetupError` to true to force the error to show.
- Resize windows, confirm error message sits below the text. You should see the following:

image

image

### DIFF
--- a/client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
+++ b/client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
@@ -382,13 +382,18 @@ export class ShippingBanner extends Component {
 											target="_blank"
 											type="external"
 											onClick={
-												this.woocommerceServiceLinkClicked
+												this
+													.woocommerceServiceLinkClicked
 											}
 										/>
 									),
 								},
 							} ) }
 						</p>
+						<SetupNotice
+							isSetupError={ this.isSetupError() }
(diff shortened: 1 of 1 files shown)


client/wp-admin-scripts/print-shipping-label-banner/shipping-banner/index.js
```diff
@@ -382,13 +382,18 @@ export class ShippingBanner extends Component {
 											target="_blank"
 											type="external"
 											onClick={
-												this.woocommerceServiceLinkClicked
+												this
```
superdav42: Stray newline?

harriswong: This is weird. I think it maybe the auto commit bot doing. Let me check.

harriswong: Yep, seems to be from prettify.

c-shultz:
I'm taking the liberty of merging this :)

PR #4011, opened 2020-03-27
Onboarding: only validate other platform name for 'other' selling venues

dechov:
Minor fix to omit validation error (and resulting disabled "Continue" button) on the other platform name text field (introduced in https://github.com/woocommerce/woocommerce-admin/pull/3812/commits/8b052b1068ae9099eddbaf760da212893c9f1e22), in an edge case where the field is first required but value is missing, and then no longer required (e.g. see test instructions below).

(Looks like the coding standards changes got tacked onto the commit via some hook.)

### Screenshots

`master` | this branch
-- | --
 | 

- Open onboarding wizard "Business Details" screen
- Select "Yes, on another platform" for "selling elsewhere"
- Select "Other" for which platform the store is currently using
- Leave the platform name text field blank, and verify that "Continue" button is disabled
- Select "No" for "selling elsewhere", and verify that **"Continue" button is enabled** 😌

### DIFF
--- a/client/dashboard/profile-wizard/steps/business-details.js
+++ b/client/dashboard/profile-wizard/steps/business-details.js
@@ -168,7 +168,10 @@ class BusinessDetails extends Component {
 			} else if ( name === 'other_platform_name' ) {
 				if (
 					! values.other_platform_name &&
-					values.other_platform === 'other'
+					values.other_platform === 'other' &&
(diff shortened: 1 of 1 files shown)


octaedro (approved):
Hi @dechov, nice catch! It's testing well here and the code LGTM 🚀

PR #4017, opened 2020-03-27
Add caching to performance indicators.

jeffstieler:
This PR seeks to increase thespeed of the performance indicators endpoint by:
* Adding caching to the data retrieval
* Changing the request to use "end of the day" for the `before` param instead of "now" - giving caching a chance to work

- Verify that performance indicators work as expected (change date ranges, etc)

### Changelog Note:

Performance: add caching to dashboard performance indicator requests.

### DIFF
--- a/client/dashboard/store-performance/index.js
+++ b/client/dashboard/store-performance/index.js
@@ -262,19 +262,13 @@ export default compose(
 
 		const primaryQuery = {
 			after: appendTimestamp( datesFromQuery.primary.after, 'start' ),
-			before: appendTimestamp(
-				endPrimary,
-				endPrimary.isSame( moment(), 'day' ) ? 'now' : 'end'
-			),
+			before: appendTimestamp( endPrimary, 'end' ),
 			stats: statKeys,
 		};
 
 		const secondaryQuery = {
 			after: appendTimestamp( datesFromQuery.secondary.after, 'start' ),
-			before: appendTimestamp(
-				endSecondary,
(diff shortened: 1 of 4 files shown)


client/dashboard/store-performance/index.js
```diff
@@ -262,19 +262,13 @@ export default compose(
 
 		const primaryQuery = {
 			after: appendTimestamp( datesFromQuery.primary.after, 'start' ),
-			before: appendTimestamp(
```
becdetat: Why was this changed? How does this change behaviour?

jeffstieler: I changed this because it essentially made the request for performance indicators impossible to cache. Without the change, if your chosen date range ended on "today", the `before` timestamp became the value "now" - down to the second.

Since we're already [busting cache on sync](https://github.com/woocommerce/woocommerce-admin/search?q=invalidate&unscoped_q=invalidate) I decided to force this query to always be to the end of day (`T23:59:59`) so that the request would be the same, which is what the cache key is built off of.

jeffstieler:
> Looks awesome. How long does the caching hold on to the values?

Without being invalidated by sync actions, the cache expiry is 1 week: https://github.com/woocommerce/woocommerce-admin/blob/4ca4e7063ada4f139fbd4a3b6a1b7282e52af571/src/API/Reports/Cache.php#L77

client/dashboard/store-performance/index.js
becdetat: Ah of course, that makes sense. Thanks.

Issue #3742, opened 2020-02-19
Currency Formats: Add a filter to dynamically change formats in client


psealock:
**Is your feature request related to a problem? Please describe.**

When extending wc-admin reports such that specific currencies are reflected, there is no way to change the currency format for items inside the chart, summary numbers, and table.

![Screen Shot 2020-02-19 at 4 02 15 PM](https://user-images.githubusercontent.com/1922453/74882605-40968580-53d4-11ea-8dba-dc7638a6149e.png)

**Describe the solution you'd like**

A client-side filter would be an excellent way to allow extensions to modify which format is used dynamically.
**Describe alternatives you've considered**

**Should this be prioritized? Why?**

Yes, currency extensions cannot fully alter values and show a currency different than the one in Woo Core settings.

tivnet:
@psealock Paul, until there is a proper solution, is it possible to just remove the currency symbol?

The markup is not the same as the regular WC:

    <span class="woocommerce-summary__item-value">C$200.00</span>

There is no separate `<span>` for the symbol, so it's harder to alter it with JS.

And the view is unpleasant:

![image](https://user-images.githubusercontent.com/1696330/77834191-28bfc900-7119-11ea-89d1-e32ee473f02c.png)

psealock:
Thanks for the question @tivnet. This is the last piece of this and is not yet implemented. Here is the issue to track https://github.com/woocommerce/woocommerce-admin/issues/3742.

tivnet:
@psealock 
Hi Paul, do I understand correctly that the filter tag will be `woocommerce_admin_report_currency`?
Thanks

psealock:
@tivnet It will be unless code review in the pull request says it should be something else. I'll update the example and blog post with a filter that works once that PR is merged.

psealock:
Closed by https://github.com/woocommerce/woocommerce-admin/pull/4027

PR #4027, opened 2020-03-30
Dynamic Currency with Context API

psealock:
Related to https://github.com/woocommerce/woocommerce-admin/issues/3742

Our usage of `CURRENCY` as a static constant is no longer valid now that the OBW has the ability to mutate the currency. As a result, changing the currency in the OBW can have odd effects if you continue to Reports without refreshing the page.

I decided to use React's [Context API](https://reactjs.org/docs/context.html) to maintain a current Currency object to avoid having to pass down props at multiple levels. In the future, this context can be changed to include all other "static" properties.

The Context API also lends itself well to adding a filter to the currency config in Reports only. This is useful because it allows extensions to modify the currency object used to view a Report without affecting the currency object used throughout the app. In other words, an extension won't be able to filter the Dashboard's currency object.

### Notable Changes

* `lib./number-format.js` has been deleted. It bound the initial currency object to its function, which is error prone since that object is indeed dynamic. Instead, the up to date currency object is passed into the functions explicitly. Its also noteworthy that `currency` also includes normal number data, such as `thousandsSeparator` used to format regular numbers.
* Currency package now returns a Factory Function instead of a Class instance. Usage of its methods required, in some cases, `.bind( Currency )` in order to get `this` to be right. Here is an example of what was wrong if you are interested, https://jsfiddle.net/t1mjwruh/5/. And a bit more on Factory Functions here: https://www.youtube.com/watch?v=ImwrezYhw4w&vl=en.

### Testing

1. Smoke test the app and look for oddly formatted or broken values.
2. Test out the filter by adding this snippet at the top of `client/index.js`
```js
import { addFilter } from '@wordpress/hooks';
addFilter( 'woocommerce_admin_report_currency', 'woocommerce-admin', ( config, query ) => {
	if ( query.currency ) {
		config.symbol = query.currency;
	}
	return config;
} );
```
3. Add a currency query parameter to the url of any report, ie `&currency=💸` and see the currency update. Note that the added query parameter has no effect on the Dashboard or OBW.


client/analytics/report/categories/table.js
```diff
@@ -95,7 +99,7 @@ class CategoriesReportTable extends Component {
 					value: category && category.name,
 				},
 				{
-					display: formatValue( 'number', itemsSold ),
+					display: formatValue( currency, 'number', itemsSold ),
```
becdetat, line 102: Why do you have to pass `currency` into `formatValue` if the value isn't a currency value?

psealock: Good question. I went through much of this refactor and then realized that `formatValue` takes the currency config as its first argument. Turns out what we've been calling currency also has numeric format data like `decimalSeparator`. I decided to leave the naming as currency, but maybe in a follow up I can change it to `numericLocale` or `locale` and include other information as well.

Screen Shot 2020-04-02 at 7 51 54 AM

becdetat: Ok no worries. Everything else looks great :-)

PR #4030, opened 2020-03-31
Marketing inbox note

### LINKED ISSUE #4029: Add an inbox notification for the new Marketing Tab
Similar to #4012 we'd like to add an inbox notification to help merchants find the new Marketing Tab.

An example of existing notifications is

Screen Shot 2020-03-31 at 8 54 26 am

We'll use text similar to the Marketing tab welcome card 

> Grow your customer base and increase your sales with marketing tools built for WooCommerce

I think the aim is for this to show upon upgrade/installation of the new version with the Marketing Tab included but will confirm if we should set a delay/trigger as part of development.

Yes, we're aiming to have the Marketing Tab released in WC 4.1.

jconroy:
Fixes #4029

Adds a note to introduce merchants to the new Marketing Tab/Dashboard/Hub.

**Update** - ([see comment below](https://github.com/woocommerce/woocommerce-admin/pull/4030#issuecomment-606925760)) after some discussion (pbIy4N-3C-p2#comment-127) we've decided to remove two existing inbox notes and replace with this one note directing merchants to the marketing hub.

---

I have not currently put any restrictions on displaying the new note because my understanding is that we want to cover everyone e.g. both
- users that don’t finish the task list (hide it / click to do it later)
- users that complete the task list

I guess that is everyone? If I'm missing some logic needed let me know.

Text currently displayed in the note is
- **Title:** WooCommerce marketing hub
- **Content:** Grow your customer base and increase your sales with marketing tools built for WooCommerce.
- **Action:** Open marketing hub 

Notes
- I went with "marketing hub" to match other community information but am happy to change any of the wording @garymurray @LevinMedia
- Similarly I went with using "open" as the action seemed better than something like "learn more"
- Avoided using something like "New xyz" for the title because it might not always be new. Happy to tweak.

Here is a screenshot with a few of the variations I tried

Screen Shot 2020-03-31 at 4 14 25 pm

* Checkout this branch
* Trigger the wc_admin_daily hook scheduled in WP Cron
* Go to a WooCommerce page
* Click the Inbox icon in the Navbar
* Observe the new note
* Click the action button to be directed to the new Marketing tab


garymurray:
Let's just change the title to: `Connect with your audience`

so it would look like this:

Screenshot 2020-03-31 at 09 51 34

jconroy:
Thanks @garymurray updated the title in 666de72

jconroy:
After some discussion (pbIy4N-3C-p2#comment-127) we've decided to remove two existing inbox notes and replace with this one note directing merchants to the marketing hub.

Onboarding Completed / Market my store (removed with SHA: be0077a) - has had a low usage/conversion rate primarily because it was only triggered after onboarding is _completed_. It was directing customers to the marketing category on WCCOM. The new note will be triggered more and directs to the marketing hub.

Facebook Extension Note (removed with SHA: be0077a) - similarly has a low usage/conversion rate. The marketing hub promotes the extension so make sense to direct merchants there.

I've also rebased this PR on master to pick up other recent changes to notes for Payments.

src/Notes/WC_Admin_Notes_Marketing.php
```diff
@@ -0,0 +1,67 @@
...
+	/**
+	 * Name of the note for use in the database.
+	 */
+	const NOTE_NAME_INTRO = 'wc-admin-marketing-intro';
+
+	/**
+	 * Attach hooks.
+	 */
+	public function __construct() {
+		add_action( 'woocommerce_note_action_install-now', array( $this, 'install' ) );
```
jeffstieler, line 33: Looks like there isn't an `install()` method in this class.

jconroy: Thanks @jeffstieler good catch, copy and past fail.Fixed with 680c05c

src/FeaturePlugin.php
```diff
@@ -185,9 +185,9 @@ public function includes() {
 		new WC_Admin_Notes_Historical_Data();
 		new WC_Admin_Notes_Order_Milestones();
 		new WC_Admin_Notes_Welcome_Message();
-		new WC_Admin_Notes_Facebook_Extension();
 		new WC_Admin_Notes_Tracking_Opt_In();
 		new WC_Admin_Notes_WooCommerce_Payments();
+		new WC_Admin_Notes_Marketing();
```
jeffstieler, line 190: This can be removed now (no constructor).

jconroy:
Thanks @jeffstieler 

> How about using the speaker gridicon for this note?

I don't mind, kind of a nice touch 👍 

> We should either leave the Facebook note class intact (so the note actions continue to work) or we should create a migration routine to remove the Facebook note from stores that have it.

> I'm happy to help with the latter approach if that's what you'd like to do.

Ahh good catch. I think removing the class and note is probably the cleanest approach rather than having the code sit around.

Definitely interested in learning how you'd tackle it / advice - I've had a look at the [examples](https://github.com/woocommerce/woocommerce-admin/blob/master/docs/examples/activity-panel-inbox.md#updating-inbox-items) so I can see the way to remove but not 100% on best place to hook in

jeffstieler:
> Definitely interested in learning how you'd tackle it / advice - I've had a look at the examples so I can see the way to remove but not 100% on best place to hook in

I think the easiest way to do this is with a version migration callback. See:

https://github.com/woocommerce/woocommerce-admin/blob/master/src/Install.php#L30

https://github.com/woocommerce/woocommerce-admin/blob/master/includes/wc-admin-update-functions.php

jconroy:
Does this look right d7b319e ? Thanks for the pointers, appreciate the guidance.

PR #4037, opened 2020-03-31
Support IE11 in `wc-admin`

### LINKED ISSUE #3972: Dashboard does not work in ie11
**Describe the bug**
I tested the wc-admin dashboard (/wp-admin/admin.php?page=wc-admin) and got a blank view. The wp-admin sidebar was visible but none of the reports appeared.

The ie11 console showed `Syntax error index.js( 7, 347320)`.

**To Reproduce**

1. Go to /wp-admin/admin.php?page=wc-admin in ie11
2. See error

**Expected behavior**
Dashboard works in ie11

**Screenshots**

**Desktop:**

* OS: Windows
* Browser IE
* Version 11

mattsherman:
Fixes #3972

This PR fixes the JS syntax errors that were preventing the `wc-admin` pages from loading in IE11.

It is probable that there are some errors that appear in IE11 now that the UI in general works. Any specific errors should be logged and addressed separately as they are uncovered.

1. Ensure you do not have `define( 'SCRIPT_DEBUG', true );` in your `wp-config.php`, as that will cause debug versions of JS files to be sent to the browser, some of which are not compatible with IE11.
2. Run branch (`npm start`).
3. Go to `/wp-admin/admin.php?page=wc-admin` in IE11.
4. Verify that dashboard appears.

You should also run on a new store, and verify that the "Setup" option appears in the WooCommerce sidebar section (see #3796 for details, as it was introduced then).

### Changelog Note:

Fix: wc-admin pages work in Internet Explorer 11.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -133,6 +133,7 @@ const webpackConfig = {
 						'|acorn-jsx' +
 						'|d3-array' +
 						'|debug' +
+						'|marked' +
 						'|regexpu-core' +
 						'|unicode-match-property-ecmascript' +
 						'|unicode-match-property-value-ecmascript)/'
--- a/client/wp-admin-scripts/onboarding-menu-experience/index.js
+++ b/client/wp-admin-scripts/onboarding-menu-experience/index.js
@@ -32,7 +32,7 @@ setupAnchorLi.hidden = true;
 const topLevelPageWooCommerceLi = document.getElementById(
 	'toplevel_page_woocommerce'
 );
-topLevelPageWooCommerceLi.children[ 1 ].append( setupAnchorLi );
+topLevelPageWooCommerceLi.children[ 1 ].appendChild( setupAnchorLi );
 
 function hideOrShowMenuItemsForTaskList( show ) {
 	const allSubmenuItems = [


becdetat (approved):
Whoops, you're fixing my bug 🥇 so thanks..

I couldn't immediately test this in Browserstack because WP sends FQNs pointing to VVV, but I compared the docs on developer.mozilla.org and it all checks out so I'm happy to approve this.

webpack.config.js
```diff
@@ -133,6 +133,7 @@ const webpackConfig = {
 						'|acorn-jsx' +
 						'|d3-array' +
 						'|debug' +
+						'|marked' +
```
becdetat, line 136: What does this do?

becdetat:
Actually I managed to get things running on BrowserStack but it's dying on another error:

image

Which is here:

image

Did you encounter this? Seems like `@wordpress/data` is dropping the ball somewhere?

webpack.config.js
mattsherman: This tells webpack to run the `marked` package js files through babel so that they get transfromed. This needs to happen because `marked` was using destructuring in its require statements, and IE11 doesn't support destructuring.

mattsherman:
> Actually I managed to get things running on BrowserStack but it's dying on another error:
> 
> image
> 
> Which is here:
> 
> image
> 
> Did you encounter this? Seems like `@wordpress/data` is dropping the ball somewhere?

Hrm... Well, I didn't when I was testing my fix originally, but I am now! Ah, wait... it's this...

Make sure you do not have the following line in your `wp-config.php`

```
define( 'SCRIPT_DEBUG', true );
```

If you have this enabled, then debug versions of the JS files are sent to the browser, some of which (not even `woocommerce-admin` ones) will causes errors in IE11, as they use syntax not supported.

In addition, please make sure that you rebuild `woocommerce-admin` -- when I went to double-check this initially I saw the same error you did when running with `define( 'SCRIPT_DEBUG', true );`. However, after toggling that setting, and rebuiliding many times, I can no longer reproduce the exact error you encountered.

mattsherman:
Putting this back up for review, with an updated description -- I would love another set of 👀 on this to make sure that what I've said about `define( 'SCRIPT_DEBUG', true );` is true and that I'm not missing some other detail!

timmyc (approved):
Created a JN site with 4.0.1, verified the bug in IE11 on browserstack. Did a `npm run test:zip` of this branch and uploaded the  zipfile to the test site and confirmed all wc-admin, and a few legacy pages ( orders, status, settings ) are all loading fine on IE11.

becdetat:
Nice! Thanks @mattsherman

PR #4042, opened 2020-04-01
Cast to float, to avoid performing arithmetic with strings

barryhughes:
Casts a numeric string (ie, `'10.00'`) to a float, to avoid warnings/error log noise in more recent versions of PHP.

Currently:

* `$this->get_shipping_total()` returns a string such as `'10.00'`.
* This is used to perform a calculation (the other values are integers).
* While this works—the calculation is still performed correctly—modern PHPs complain about the use of strings in this situation _("PHP Warning: A non-numeric value encountered")._

This change:

* Casts the string to a float.
* This should avoid unwanted noise in the error logs.

### DIFF
--- a/src/Overrides/OrderTraits.php
+++ b/src/Overrides/OrderTraits.php
@@ -31,7 +31,7 @@ trait OrderTraits {
 			return 0;
 		}
 
-		$total_shipping_amount = $this->get_shipping_total();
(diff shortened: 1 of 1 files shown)


octaedro (approved):
Nice catch @barryhughes. Code LGTM and it's testing well here  🚢

barryhughes:
Thanks for the quick review :-)

PR #4046, opened 2020-04-01
Enable wcpay feature flag in plugin and core

### LINKED ISSUE #4023: Enable WooCommerce Payments on plugin and core
See related #4022

allendav:
Fixes #4023 

### Accessibility

### Screenshots

N/A

- Build the plugin and install on a site. Ensure the WooCommerce Payments task appears in the new OBW and that you can use it to install, activate and initiate the KYC flow
- Remove both plugins and then build WooCommerce core from master. Install on a site and repeat the above test

### DIFF
--- a/config/core.json
+++ b/config/core.json
@@ -9,6 +9,6 @@
 		"onboarding": true,
 	  	"shipping-label-banner": true,
 		"store-alerts": true,
-		"wcpay": false
+		"wcpay": true
 	}
 }
--- a/config/plugin.json
+++ b/config/plugin.json
@@ -9,6 +9,6 @@
 		"onboarding": true,
 		"shipping-label-banner": true,
 		"store-alerts": true,
-		"wcpay": false
+		"wcpay": true
 	}
 }


becdetat:
@allendav I don't understand from your test instructions what is meant to happen here. Probably because I don't know how this feature flag works. My setup is using VVV with WC 4.0.1 installed from the plugins screen, then with wc-admin cloned from master (and testing from your branch) and I couldn't see any difference from wc-admin master and your branch. Can you add some more instructions and what you would expect to see differently between your branch and master?

joshuatf:
@becdetat Usually in the master branch running `npm start` we're looking at the dev environment, but the feature flags let us define what's included in the dev, plugin, or core. 
 Here's how you can test the other phases:

* **plugin** - Run `npm run test:zip` from your `woocommerce-admin` directory.  This will build a `woocommerce-admin.zip` file ready to upload to a test site.  
* **core** - Run `npm run prepack`. This will build a core ready version that you can symlink or copy to the `woocommerce/packages/woocommerce-admin` directory.  Note that you'll probably also want to run `composer dumpautoload` under `woocommerce` to fix the autoloading.  And of course remember to deactivate the initial plugin so we can test the version included in core.

Depending on your environment it can be good to also test these on https://jurassic.ninja/ if you need to quickly spin up a new site.

Let me know if you have any questions :)

Issue #3881, opened 2020-03-12
Move plugin information and Jetpack status out of onboarding settings


joshuatf:
Currently the wc-api includes selectors such as `isJetpackConnected` and `getActivePlugins` that exist under onboarding.

These are selectors that may benefit a whole host of areas outside of onboarding.  While it's possible to use these selectors without onboarding being active, they have potential to throw errors with data not being preloaded.

This issue will involve:
* Moving these selectors out of onboarding since they are not onboarding specific.
* Moving the preloaded settings data out onboarding and (possibly?) always preload this data or create a flag/filter to easily do so.
* Updating all references to existing settings getters (e.g., `const { activePlugins = [] } = getSetting( 'onboarding', {} );` ) to the new namespace or (preferably) use the `getActivePlugins` selector to retrieve via the API.
* Removing preloaded settings introduced in https://github.com/woocommerce/woocommerce-admin/pull/3839

Related to shared settings: https://github.com/Automattic/woocommerce-shipping-issues/issues/28

psealock:
It would be great to get this merged before starting on this project so refactors can happen at the same time. https://github.com/woocommerce/woocommerce-admin/pull/3313

psealock:
@joshuatf, I'm looking into making a new data store at `/packages/data/src/plugins` and moving plugin related information out of `window.wcSettings.oboarding` and into `window.wcSettings.plugins` in case that data needs to be available on load.

This offers lots of reusability for clients, but how about the onboarding API on server? Should the new plugin data store gather data from `/onboarding/plugins` endpoint? I think to make it truly reusable, we'd need to create a plugins endpoint that onboarding can make use of. Thoughts?

psealock:
Also, Onboarding trims the list of `activePlugins` according to `get_allowed_plugins`. I'm trying figure out the best way to handle this and it depends on the goal of moving this info out of Onboarding.

It seems to me that if a plugins data store is to be truely agnostic, then `get_allowed_plugins` will need to be done client side. Otherwise, if plugins data store is Ok with working under the guidelines of Onboarding, then we can just have it operate off of `/onboarding/plugins`.

PR #4048, opened 2020-04-02
Plugins DataStore: decouple from onboarding

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/3881

Make Plugins manipulation its own store to decouple it from onboarding. Selectors and resolvers surrounding installing, activating, and jetpack status or connections are now in a standalone data store in the data package.

### Differences to Fresh Data

The main difference is the inclusion of generator functions as action creators. What does this mean? Previously, analysis of `componentDidUpdate` by comparing `props` to `prevProps` was required to determine how a request was doing and what state it was in.

Now, async/await can be used to explicitly return results or errors from actions:

```js
async function install() {
	const results = await installPlugin( 'jetpack' );
	if ( results.status === 'success' ) {
		// handle success
		onInstall( results );
	} else {
		// handle error
		onError( results );
	}
}
```

### Notable Changes

* All plugin data and manipulation uses `@woocommerce/data`.
* `<Plugins />` is now an `@woocommerce/components` component.
* `<Connect />` now lives in `dashboard/components`

### Test Instructions

OBW and Task List are the only users (besides Shipping-label) to use Plugin data, so a smoke test of those two areas is the best testing.


harriswong:
I pulled the branch and tested the WooCommerce Service banner. 
- Tested tracking with `localStorage.setItem( 'debug', 'wc-admin:*' );`, verified events present.
- Tested no WCS installation to printing label process.
- Tested WCS installed but not activated, to printing label process
- Tested the "You've installed..." banner paragraph
image

The WCS banner code and tests look good to me!

jeffstieler:
Ran into some errors while testing (unclear if all are issues with this PR or present on `master` as well):

* `An error occurred while running 'mapSelect': Cannot read property 'isActive' of undefined @ | with-plugins-hydration.js:16` (with Jetpack inactive - in console, on all page loads that use plugin data)
* `PHP Fatal error:  Uncaught Error: Class 'Automattic\WooCommerce\Admin\API\WP_Error' not found in /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-admin/src/API/OnboardingPlugins.php:578` (when trying to click "verify business details" from WooCommerce Payments set up task

psealock:
Thanks for the review @jeffstieler !

>An error occurred while running 'mapSelect': ...

Weird that my setup swallowed those errors. In any case I traced it back to an assumption that Jetpack was active. Here is the fix for that https://github.com/woocommerce/woocommerce-admin/pull/4048/commits/1f9e6c204a895d2fbca95840ffa8f0ec3bd5718b

>PHP Fatal error: Uncaught Error: Class 'Automattic\WooCommerce\Admin\API\WP_Error'

Thanks for fixing up those incorrect calls to `WP_Error`. Without Jetpack, you shouldn't have gotten to that error. https://github.com/woocommerce/woocommerce-admin/pull/4105 is a recent PR to ensure that. I rebased and changed `isJetpackConnected` to come from the right store in  https://github.com/woocommerce/woocommerce-admin/pull/4048/commits/bfd0eb8f24d4b23f6e20fe6abea226f42e4813cf

PR #4051, opened 2020-04-02
Fix RTL issues with `wc-admin` pages

### LINKED ISSUE #3811: Small RTL issues
**Describe the bug**
When testing 4.0-rc.1, I noticed some RTL issues, mostly of cosmetic character:

- Notice carousel is incorrect and 
- arrows in the Performance section are pointed in the opposite direction
Screenshot 2020-03-03 at 17 11 13

The loading icon is incorrect: small dot is rotating outside the large circle:
Screenshot 2020-03-03 at 17 11 51

There might be other errors I haven't noticed, but I thought I'd mention these just in case.

* OS: macOS 10.14.6
* Browser Chrome 80.0.3987.122

mattsherman:
Fixes #3811

This PR corrects three RTL issues with the `wc-admin` pages:

- Correct store alerts pagination arrows display
- Correct chart legend checkboxs display
- Correct summary item's delta arrow display when flat

The approach used is similar to that used in `wp-calypso` to correct elements that should/shouldn't be flipped when WordPress is in RTL mode ([example](https://github.com/Automattic/wp-calypso/blob/master/client/assets/stylesheets/_main.scss#L109)).

Styles are introduced that are only applied when in RTL mode. These CSS rules are annotated with `/*!rtl:ignore*/` so that when the RTL CSS files are built, `rtlcss` doesn't attempt to automatically 'fix' them.

### Screenshots

Before:

ScreenCapture at Thu Apr 2 13:57:51 EDT 2020

After:

ScreenCapture at Thu Apr 2 14:02:10 EDT 2020

_Note: The `of 2 1` label in the pagination control is okay. This text would be translated in a real RTL locale, and display properly. If this English string was to be shown in a RTL setting, it would be modified with BiDi directional control characters to be ordered properly. But, it is only shown in English in a test environment, so there is no need for that._

1. Run branch.
2. Make sure you have two or more store alerts that always appear at the top of the `wc-admin` pages. You can easily force this by going into your DB and changing the `type` value for two `wp_wc_admin_notes` table rows to `error`.
3. Make sure you have some data in your store, so that you can see a variety of summary item delta arrows. You can use [WooCommerce Smooth Generator](https://github.com/woocommerce/wc-smooth-generator) to quickly populate your store with test data.
4. Go to Analytics (`/wp-admin/admin.php?page=wc-admin&path=%2Fanalytics%2Frevenue&period=last_week&compare=previous_period`) and verify that elements appear properly in LTR.
5. Switch your WordPress to RTL mode. The easiest way to do this for testing is to use the [RTL Tester plugin](https://wordpress.org/plugins/rtl-tester/).
6. Verify that elements appear properly in RTL.

### Changelog Note:

Fix: Proper display of elements in `wc-admin` pages when in a RTL environment.

### DIFF
--- a/client/layout/store-alerts/style.scss
+++ b/client/layout/store-alerts/style.scss
@@ -119,6 +119,12 @@
 
 	.components-button {
 		padding: 4px;
+
+		/*!rtl:ignore*/
+		.rtl & .dashicons-arrow-left-alt2,
+		.rtl & .dashicons-arrow-right-alt2 {
+			transform: scaleX(-1);
+		}
 	}
 
 	.woocommerce-store-alerts__pagination-label {
--- a/packages/components/src/chart/d3chart/legend.scss
+++ b/packages/components/src/chart/d3chart/legend.scss
@@ -112,6 +112,11 @@
 						border: solid $studio-white;
 						border-width: 0 2px 2px 0;
 						transform: rotate(45deg);
+
+						/*!rtl:ignore*/
+						.rtl & {
+							transform: rotate(45deg) scaleX(-1);
+						}
 					}
 				}
 			}
--- a/packages/components/src/summary/style.scss
+++ b/packages/components/src/summary/style.scss
@@ -373,6 +373,11 @@ $inner-border: $core-grey-light-500;
 		margin-right: 3px;
 		fill: currentColor;
 
+		/*!rtl:ignore*/
+		.rtl &.gridicons-arrow-right {
+			transform: scaleX(-1);
+		}
+
 		&.gridicons-arrow-up {
 			transform: rotate(45deg);
 		}


client/layout/store-alerts/style.scss
```diff
@@ -119,6 +119,12 @@
 
 	.components-button {
 		padding: 4px;
+
+		/*!rtl:ignore*/
```
becdetat, line 123: This wouldn't be any good for a RTL user that uses a screen reader. I don't know if that's too much of an edge case to worry about though..

becdetat (approved):
Cool. I think the screen reader issue is really an extreme edge case that could be ignored at the moment.

client/layout/store-alerts/style.scss
mattsherman: The buttons (`IconButton` instances) have labels which render as `aria-label` attributes to provide text for screen readers (you can also see this by hovering over the buttons), so I think that would cover the RTL screen reader scenario. Or, is there something else I'm overlooking?

mattsherman:
Thanks for the review @becdetat ! I think the screen reader concern is addressed by the current implementation, but if not we can open up a followup issue to track that.

PR #4059, opened 2020-04-03
Convert order status setting from blacklist to whitelist.

### LINKED ISSUE #4050: Unexpected Order Statuses Included in Analytics
**Describe the bug**

Unexpected/unregistered order statuses are included in Analytics queries, resulting in incorrect totals in reports.

**To Reproduce**

Either enable a plugin that creates custom order statuses, or modify some existing order statuses to something other than the core statuses, like `wc-partial-payment`, etc.

1. Ensure that you have a few orders with non-standard statuses
1. Make sure the stats tables are up-to-date
1. Use WooCommerce > Reports to look at sales by date for a date range that includes the test orders
1. Use Analytics > Revenue to look at the same date range
1. See the discrepancies

**Expected behavior**

I expect unaccounted for order statuses to be ignored in reporting, in line with what the original core WC reporting does.

**Screenshots**

Original core report:

Screen Shot 2020-04-02 at 9 41 58 AM

New analytics:

Screen Shot 2020-04-02 at 9 41 18 AM

(Caused by 2 orders with `wc-partial-payment` and `wc-scheduled-payment` statuses from the WooCommerce Deposits extension)

**Additional context**

See Slack: p1585775805046000-slack-team-isotope

jeffstieler:
Fixes #4050.

This PR seeks to remove unexpected/defunct/unregistered statuses from Order stats queries. It accomplishes this by changing the "excluded order status" (black list) setting to "included order status" (whitelist). In turn, the query is changed from a `NOT IN` to an `IN` - making the results predictable.

The problem being solved is best illustrated by using an extension that adds custom order statuses. The test store I first saw this error produced had **previously** used WooCommerce Deposits, but no longer had the extension active. This left many orders with custom statuses (like `wc-partial-payment`) that polluted report data because they were included (with the existing black list behavior) in reports. Without the extension active, there was no way to exclude them in settings.

### Screenshots

Screen Shot 2020-04-03 at 12 37 46 PM

- On `master` branch:
- Take note of your "Excluded Statuses" setting on Analytics > Settings
- Check out this branch
- Update https://github.com/woocommerce/woocommerce-admin/blob/b15748708f74270d9ca56d7f79a40af0a36b0eee/src/FeaturePlugin.php#L155 to `1.0.4`
- Load any admin page
- Verify the `woocommerce_excluded_report_order_statuses ` option has been removed
- Verify a new `woocommerce_included_report_order_statuses` option exists
- Go to Analytics > Settings
- Verify that a new "Included Statuses" setting is shown
- Verify the "Included Statuses" value is the inverse of the defunct "Excluded Statuses"
- Enable WooCommerce Deposits extension (set up a store-wide partial payment)
- Place a partial payment order
- Open WooCommerce > Reports
- Select just today for the date range
- In a new tab/window, open Analytics > Revenue
- Ensure the date range includes today
- Verify the numbers for today match between the two reports

### Changelog Note:

Tweak: Change Order Status setting to be inclusive instead of exclusive (fix unexpected orders in reports).

### DIFF
--- a/includes/wc-admin-update-functions.php
+++ b/includes/wc-admin-update-functions.php
@@ -8,6 +8,7 @@
  */
 
 use \Automattic\WooCommerce\Admin\Install as Installer;
+use \Automattic\WooCommerce\Admin\Loader as Loader;
 use \Automattic\WooCommerce\Admin\Notes\WC_Admin_Notes;
 
 /**
@@ -75,4 +76,24 @@ function wc_admin_update_0251_remove_unsnooze_action() {
  */
 function wc_admin_update_0251_db_version() {
 	Installer::update_db_version( '0.25.1' );
+}
+
+/**
+ * Convert the "excluded" order status black list into an "included" white list.
+ * See: https://github.com/woocommerce/woocommerce-admin/issues/4050.
+ */
+function wc_admin_update_104_migration_excluded_order_status() {
+	$order_statuses = array_keys( Loader::get_order_statuses( wc_get_order_statuses() ) );
+	$black_list     = get_option( 'woocommerce_excluded_report_order_statuses', array( 'pending', 'cancelled', 'failed' ) );
+	$white_list     = array_values( array_diff( $order_statuses, $black_list ) );
+
+	update_option( 'woocommerce_included_report_order_statuses', $white_list );
+	delete_option( 'woocommerce_excluded_report_order_statuses' );
+}
+
+/**
+ * Update DB Version.
+ */
+function wc_admin_update_104_db_version() {
+	Installer::update_db_version( '1.0.4' );
 }
--- a/includes/wc-admin-update-functions.php
+++ b/includes/wc-admin-update-functions.php
@@ -85,7 +85,7 @@ function wc_admin_update_0251_db_version() {
 function wc_admin_update_104_migration_excluded_order_status() {
 	$order_statuses = array_keys( Loader::get_order_statuses( wc_get_order_statuses() ) );
 	$black_list     = get_option( 'woocommerce_excluded_report_order_statuses', array( 'pending', 'cancelled', 'failed' ) );
-	$white_list     = array_values( array_diff( $order_statuses, $black_list ) );
+	$white_list     = array_values( array_diff( $order_statuses, $black_list, array( 'refunded' ) ) );
 
 	update_option( 'woocommerce_included_report_order_statuses', $white_list );
 	delete_option( 'woocommerce_excluded_report_order_statuses' );
--- a/client/analytics/settings/config.js
+++ b/client/analytics/settings/config.js
@@ -68,36 +68,6 @@ export const config = applyFilters( SETTINGS_FILTER, {
 		} ),
 		defaultValue: [ 'processing', 'on-hold', 'completed' ],
 	},
-	woocommerce_excluded_report_order_statuses: {
-		label: __( 'Excluded Statuses:', 'woocommerce-admin' ),
-		inputType: 'checkboxGroup',
-		options: [
-			{
-				key: 'defaultStatuses',
-				options: filteredOrderStatuses.filter( status =>
-					DEFAULT_ORDER_STATUSES.includes( status.value )
-				),
-			},
-			{
-				key: 'customStatuses',
-				label: __( 'Custom Statuses', 'woocommerce-admin' ),
-				options: filteredOrderStatuses.filter(
-					status => ! DEFAULT_ORDER_STATUSES.includes( status.value )
-				),
-			},
-		],
-		helpText: interpolateComponents( {
-			mixedString: __(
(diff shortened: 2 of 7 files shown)


includes/wc-admin-update-functions.php
```diff
@@ -76,3 +77,23 @@ function wc_admin_update_0251_remove_unsnooze_action() {
 function wc_admin_update_0251_db_version() {
 	Installer::update_db_version( '0.25.1' );
 }
+
+/**
+ * Convert the "excluded" order status black list into an "included" white list.
+ * See: https://github.com/woocommerce/woocommerce-admin/issues/4050.
+ */
+function wc_admin_update_104_migration_excluded_order_status() {
```
joshuatf, line 85: Should we come up with some way of auto-triggering the data sync when this occurs or do we think this is a rare enough scenario that users should manually import after updating?

If we do want to trigger the import, we can probably only trigger if there are differences between the two currently synced statuses array.

joshuatf (commented):
Code here looks good.

I'm wondering if we're not trading up problems by switching from excluded to included order statuses.  If I recall correctly, the reason for opting to use excluded statuses is that **most** custom order statuses would want to be included (e.g., many stores add a "shipping" or "shipped" status to orders that should be included in reports).  

Do you think that the number of statuses that should be excluded outweighs the number included?  Or are there other benefits to swapping these for included statuses?  Would a filter work here to forcibly exclude these statuses or at least exclude them by default?

includes/wc-admin-update-functions.php
jeffstieler: Order status isn't taken into account when syncing - only in report queries.

jeffstieler:
Thanks for the review @joshuatf.

> Do you think that the number of statuses that should be excluded outweighs the number included? Or are there other benefits to swapping these for included statuses? Would a filter work here to forcibly exclude these statuses or at least exclude them by default?

The main problem included solves versus excluded is when you have custom order statuses that resulted from the use of an extension that is no longer installed/active. With the current (exclusion) behavior - those orphaned statuses **can't** be excluded because the statuses are no longer being filtered in.

Note that this PR introduces the opposite problem - those orphaned statuses can't be **included**, since they are no longer being filtered in.

Perhaps the available choices in the white list should be a combination of all registered statuses and the unique statuses already present in the stats table?

joshuatf:
Thanks for that explanation, Jeff!

>Perhaps the available choices in the white list should be a combination of all registered statuses and the unique statuses already present in the stats table?

This sounds like a good solution to me.  Are you thinking of storing available statuses at the time of sync?

Are you still considering white listing statuses or can we continue to use excluded statuses with this method?

jeffstieler:
> This sounds like a good solution to me. Are you thinking of storing available statuses at the time of sync?

I was planning on just querying for the unique statuses on the fly on the settings screen. (I'll do some quick benchmarking to make sure it's not too expensive a query)

> Are you still considering white listing statuses or can we continue to use excluded statuses with this method?

I still like the idea of the whitelist because it's a direct explanation of what's in the reports. However - without additional logic/consideration, the addition of order statuses (by an extension most likely) would need to be followed up with a trip to the Analytics > Settings page to include them..

jeffstieler:
> Are you still considering white listing statuses or can we continue to use excluded statuses with this method?

Hmmm... I'm reconsidering my previous answer. Perhaps this is the best way forward. I'll look into any potential performance implications of using `IN` vs. `NOT IN` - though I think they'll both use the `status` index and differences will be negligible.

joshuatf:
Barring any performance concerns, my preference is towards exclusion if we're working under the assumption that most custom statuses will want to be excluded.  While we end up with some frustrated/confused users with either method, this seems to be the one that will overall require fewer settings changes to most users.

I think I originally may have spoke with @LevinMedia about this one, so feel free to chime in here if you have any additional thoughts.

>direct explanation of what's in the reports. 

If the issue is about showing inclusion in the report it would be trivial to continue using excluded statuses on the backend and find the difference on the frontend to show included statuses.

Though this means that if a new status is added, it will be checked by default which may or may not cause confusion.

jeffstieler:
I've totally changed my mind on this - and will revert the whitelist conversion in favor of finding "orphaned" statuses to be used in the setting.

> if we're working under the assumption that most custom statuses will want to be excluded

I'm actually working under the opposite assumption - which is why the blacklist makes more sense as new order statuses will automatically be included in reports. (E.g. when a store activates WooCommerce Deposits and starts taking partial payments)

joshuatf:
>if we're working under the assumption that most custom statuses will want to be excluded

I misspoke on this; what I meant to say is exactly how you put it - most custom statuses are _included_.

>finding "orphaned" statuses to be used in the setting.

Cool.  This sounds like a great approach to me.

jeffstieler:
Superseded by https://github.com/woocommerce/woocommerce-admin/pull/4090.

PR #4068, opened 2020-04-07
Update Jetpack Autoloader.

timmyc:
Getting wc-admin on the same version as Woo Core, Blocks, and the REST API.


jeffstieler (approved):
🚀

PR #4071, opened 2020-04-07
Setup Checklist: Fix exception in shipping.

### LINKED ISSUE #4047: Shipping Task: Fatal error
**Describe the bug**

On a fresh install with `1.1.0-plugin` from https://github.com/woocommerce/woocommerce-admin/releases/tag/v1.1.0-plugin, the Shipping Task on the Task List creates a fatal error.

Screen Shot 2020-04-02 at 9 12 35 AM

**To Reproduce**

I haven't been able to reproduce on my local install. On the problematic fresh install, when I refresh the screen, or navigate elsewhere and come back, the error disappears. I'm not sure how to completely reset the shipping values and zones.

1. Create new Ephemeral or fresh install
2. Complete the OBW
3. On the Task List, go to Shipping
4. See the error

timmyc:
Fixes #4047

It appears not all REST API responses to `wc/v3/shipping/zones` return the `methods` attribute always, which results in the exception detailed in #4047. This branch seeks to resolve the bug by sniffing out for the times when `methods` is not present, and returning an empty array shipping methods for zones.

### Screenshots

Example REST response with missing zones attribute that is causing the exception:

Dashboard ‹ WooCommerce ‹ My WordPress Site — WooCommerce - (Private Browsing) 2020-04-07 11-42-09

- On a new test site ( clean site is critical here to ensure the default zone is missing methods ), I used an ephemeral site and first verified the bug by:
- Go through the OBW
- On the Setup Checklist, visit the shipping step, the error should happen after the REST requests are finished.
- **apply this branch**
- Refresh the screen and verify the exception no longer occurs.

### Changelog Note:

Fix: Add null check for shipping zone methods.

### DIFF
--- a/client/dashboard/task-list/tasks/shipping/rates.js
+++ b/client/dashboard/task-list/tasks/shipping/rates.js
@@ -27,6 +27,11 @@ class ShippingRates extends Component {
 	}
 
 	getShippingMethods( zone, type = null ) {
+		// Sometimes the wc/v3/shipping/zones response does not include a methods attribute, return early if so.
+		if ( ! zone || ! zone.methods || ! Array.isArray( zone.methods ) ) {
+			return [];
+		}
+
 		if ( ! type ) {
 			return zone.methods;
 		}


psealock (approved):
I tested this on a fresh Ephemeral site and it fixes the error.

>Sometimes the wc/v3/shipping/zones response does not include a methods attribute

It must not return `type` and `methods` properties because otherwise the return statement would handle it in the ternary. In any case, this is a comprehensive check and fixes the bug. Thanks for fixing  @timmyc.

PR #4079, opened 2020-04-08
Add 'action after' date to admin notes and hourly task

### LINKED ISSUE #3973: No effect clicking on wc-settings wc-hide-notice=update _wc_notice_nonce
Just updated WC 3.9.3 to 4.0.1 and went fine til database updated. I got a notice in WC screens which told me that database update is finished. In notice the button "Thanks" does not work. If I click on it, a page reload is done but notice still comes up.

Any idea how to close/hide notice?

becdetat:
Fixes #3973 

The reason for this is to end up resolving #3973. The first fix was rejected - I tried removing the nonce from the done admin note's action but this was rejected by the wc core team for security reasons. I then tried adding a callback function to the admin note's action, however the nonce seems to be processed before the callback so actioning the note in the callback still  happened too late - the error from the nonce still appeared first. So the idea here is to action the admin note before the nonce expires (24 hours), so it just disappears.

This PR is the first step toward that - it adds the `date_action_after` timestamp to admin notes, and sets up an hourly task to process those timestamps and action the matched admin notes.

The second step is back in the `woocommerce` plugin, setting the `date_action_after` to just less than 24 hours when the 'done' admin note is created.

Check out the `master` branch.

Install the "WP Crontrol" plugin - needed to see cron events.

In phpMyAdmin delete any rows in `wp_wc_admin_notes` with the `name` of `wc-update-db-reminder`.

Check out this PR's branch.

Increment the version in `woocommerce_admin_version` option to trigger a db schema update.

In the woocommerce core repo, edit `includes/admin/notes/class-wc-notes-run-db-update.php` and at the end of the constructor add this:

```
$note_id = $this->update_needed_notice();
$this->update_done_notice($note_id);
$note = new WC_Admin_Note( $note_id );
$note->set_date_action_after( time() + 24*60*60 );
$note->save();
```

Refresh WordPress and the done notice should appear at the top of the page. **Remove the added code from the constructor before continuing.**

Go to Tools -> Cron Events. There should be an entry with a hook name of `wc_admin_action_admin_notes` and a recurrence of one hour.

In phpMyAdmin find the `wp_wc_admin_notes` row with the name of `wc-update-db-reminder` and a `status` of `unactioned`. Check that the `date_action_after` value for that row is roughly 24 hours after the current date/time (this was set when the note was created above). If there is a discrepancy check VVV for clock shift (`vagrant ssh`, `date`). Yes I'm serious.

Modify the row and set the `date_action_after` to some time just less than the current time.

Back in WordPress -> Tools -> Cron Events, find that `wc_admin_action_admin_notes` entry and click 'Run Now'. Wait a couple of seconds then go back to phpMyAdmin. The row `status` should now be `actioned`. Go back to WordPress and go back to a page that was showing the done notice - the notice should now be gone.

### Changelog Note:

Dev: add action after date/time to admin notes, along with an hourly task that automatically actions candidate admin notes

### DIFF
--- a/src/Notes/WC_Admin_Notes.php
+++ b/src/Notes/WC_Admin_Notes.php
@@ -20,10 +20,17 @@ class WC_Admin_Notes {
 	const UNSNOOZE_HOOK = 'wc_admin_unsnooze_admin_notes';
 
 	/**
+	 * Hook used for recurring "action" action.
+	 */
+	const ACTION_HOOK = 'wc_admin_action_admin_notes';
+
+	/**
 	 * Hook appropriate actions.
 	 */
 	public static function init() {
 		add_action( 'admin_init', array( __CLASS__, 'schedule_unsnooze_notes' ) );
+		add_action( 'admin_init', array( __CLASS__, 'schedule_action_notes' ) );
+		add_action( self::ACTION_HOOK, array( __CLASS__, 'action_notes' ) );
 	}
 
 	/**
@@ -38,21 +45,22 @@ class WC_Admin_Notes {
 		$raw_notes  = $data_store->get_notes( $args );
 		$notes      = array();
 		foreach ( (array) $raw_notes as $raw_note ) {
-			$note                               = new WC_Admin_Note( $raw_note );
-			$note_id                            = $note->get_id();
-			$notes[ $note_id ]                  = $note->get_data();
-			$notes[ $note_id ]['name']          = $note->get_name( $context );
-			$notes[ $note_id ]['type']          = $note->get_type( $context );
-			$notes[ $note_id ]['locale']        = $note->get_locale( $context );
-			$notes[ $note_id ]['title']         = $note->get_title( $context );
-			$notes[ $note_id ]['content']       = $note->get_content( $context );
-			$notes[ $note_id ]['icon']          = $note->get_icon( $context );
-			$notes[ $note_id ]['content_data']  = $note->get_content_data( $context );
-			$notes[ $note_id ]['status']        = $note->get_status( $context );
-			$notes[ $note_id ]['source']        = $note->get_source( $context );
-			$notes[ $note_id ]['date_created']  = $note->get_date_created( $context );
-			$notes[ $note_id ]['date_reminder'] = $note->get_date_reminder( $context );
-			$notes[ $note_id ]['actions']       = $note->get_actions( $context );
+			$note                                   = new WC_Admin_Note( $raw_note );
+			$note_id                                = $note->get_id();
+			$notes[ $note_id ]                      = $note->get_data();
+			$notes[ $note_id ]['name']              = $note->get_name( $context );
+			$notes[ $note_id ]['type']              = $note->get_type( $context );
+			$notes[ $note_id ]['locale']            = $note->get_locale( $context );
+			$notes[ $note_id ]['title']             = $note->get_title( $context );
+			$notes[ $note_id ]['content']           = $note->get_content( $context );
+			$notes[ $note_id ]['icon']              = $note->get_icon( $context );
+			$notes[ $note_id ]['content_data']      = $note->get_content_data( $context );
+			$notes[ $note_id ]['status']            = $note->get_status( $context );
+			$notes[ $note_id ]['source']            = $note->get_source( $context );
+			$notes[ $note_id ]['date_created']      = $note->get_date_created( $context );
+			$notes[ $note_id ]['date_reminder']     = $note->get_date_reminder( $context );
+			$notes[ $note_id ]['actions']           = $note->get_actions( $context );
+			$notes[ $note_id ]['date_action_after'] = $note->get_date_action_after( $context );
 		}
 		return $notes;
 	}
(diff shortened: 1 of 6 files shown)


becdetat:
The reason for this is to end up resolving #3973. The first fix was rejected - I tried removing the nonce from the done admin note's action but this was rejected by wc core for security reasons. I then tried adding a callback function to the admin note's action, however the nonce seems to be processed before the callback so actioning the note in the callback still  happened too late - the error from the nonce still appeared first. So the idea here is to action the admin note before the nonce expires (24 hours), so it just disappears.

This PR is the first step toward that - it adds the `date_action_after` timestamp to admin notes, and sets up an hourly task to process those timestamps and action the matched admin notes.

The second step is back in the `woocommerce` plugin, setting the `date_action_after` to just less than 24 hours when the 'done' admin note is created.

src/Notes/WC_Admin_Notes.php
```diff
@@ -19,11 +19,18 @@ class WC_Admin_Notes {
 	 */
 	const UNSNOOZE_HOOK = 'wc_admin_unsnooze_admin_notes';
 
+	/**
+	 * Hook used for recurring "action" action.
```
jeffstieler, line 23: What do you think about using "expiration" or "expire" instead of "action" as a verb? (I know we're already doing that with status, but it's a little awkward in my opinion)

jeffstieler:
I assume we'll need to modify our `woocommerce_admin_version` option value to kick off the DB schema changes?

> Check that the `date_action_after` value for that row is roughly 24 hours after the current date/time (this was set when the note was created above).

Is there a branch of core we should be using to test? How is `date_action_after` getting set for testing?

src/Notes/WC_Admin_Notes.php
timmyc: I see both sides to this one myself. `action` is definitely kind of a loaded word in WP-nomenclature. But also it is the name of how we track this in the DB. My other thought here is should we just add the call of `::action_notes()`  into `[https://github.com/woocommerce/woocommerce-admin/blob/master/src/Events.php#L64](do_wc_admin_daily)` - that way we aren't adding in another recurring job?

becdetat:
> I assume we'll need to modify our woocommerce_admin_version option value to kick off the DB schema changes?

I didn't have to update the version to get the DB change - the new column just appeared after I added it all over the place and refreshed.

> Is there a branch of core we should be using to test? How is date_action_after getting set for testing?

At this point you should be on this PR's feature branch. The `date_action_after` value is set when the note is created earlier in the test instructions.

src/Notes/WC_Admin_Notes.php
becdetat: I would prefer using an existing hook as @timmyc suggests, but is once a day going to be enough for this? It could potentially sit there for 23 hours with an expired nonce.

becdetat:
Thanks @octaedro :-)

I think the problem is that the new column isn't being added to the `wp_wc_admin_notes` table.

@jeffstieler how do i go about changing the version to trigger the table schema update? For some reason I didn't need to do this when developing this, the migration just happened.

jeffstieler:
> @jeffstieler how do i go about changing the version to trigger the table schema update? For some reason I didn't need to do this when developing this, the migration just happened.

If you need to trigger the DB update, just modify the `woocommerce_admin_version` option in the database (`wp_options` table).

becdetat:
> If you need to trigger the DB update, just modify the woocommerce_admin_version option in the database (wp_options table).

Awesome thanks :-)

@octaedro can you try this to get the schema to update?

octaedro:
Hi there, I changed the `woocommerce_admin_version ` version and restarted the server, etc. but it is still the same 😞 
Am I missing something to update the database schema @jeffstieler?

jeffstieler:
I'm encountering an error testing out this branch that points to a potential problem in our Composer setup:

```
[21-Apr-2020 18:18:56 UTC] WordPress database error Unknown column 'date_action_after' in 'field list' for query INSERT INTO `wp_wc_admin_notes` (`name`, `type`, `locale`, `title`, `content`, `icon`, `status`, `source`, `is_snoozable`, `content_data`, `date_created`, `date_reminder`, `date_action_after`) VALUES ('wc-admin-deactivate-plugin', 'update', 'en_US', 'Deactivate old WooCommerce Admin version', 'Your current version of WooCommerce Admin is outdated and a newer version is included with WooCommerce.  We recommend deactivating the plugin and using the stable version included with WooCommerce.', 'warning', 'unactioned', 'woocommerce-admin', '0', '{}', '2020-04-21 18:18:56', NULL, NULL) made by require_once('wp-load.php'), require_once('wp-config.php'), require_once('wp-settings.php'), do_action('plugins_loaded'), WP_Hook->do_action, WP_Hook->apply_filters, Automattic\WooCommerce\Packages::on_init, Automattic\WooCommerce\Packages::load_packages, call_user_func, Automattic\WooCommerce\Admin\Composer\Package::init, Automattic\WooCommerce\Admin\Notes\WC_Admin_Notes_Deactivate_Plugin::add_note, WC_Data->save, WC_Data_Store->create, Automattic\WooCommerce\Admin\Notes\DataStore->create
```

becdetat:
I've updated the instructions slightly to cover updating the `woocommerce_admin_version` option and setting the note's `date_action_after` when the note is created.

@jeffstieler :

> I'm encountering an error testing out this branch that points to a potential problem in our Composer setup

I can't repro this. I created a fresh install using VVV, installed WooCommerce, cloned the woocommerce-admin repo into the plugins folder, ran `npm install`, `composer install`, `npm start`, then switched the branch to this PR's, incremented the `woocommerce_admin_version` option, and the schema got changed in the database and I was able to continue testing.

How is the schema discrepancy related to Composer?

jeffstieler:
> How is the schema discrepancy related to Composer?

I think I'm mistaken here and was confused by seeing `Automattic\WooCommerce\Admin\Composer\Package::init` fired - thinking that would only happen if we weren't using the plugin - but it seems that logic is in the package class.

In my environment, it's causing the schema error because it's trying to create a note due to a version mismatch, and that note creation call is using the new table column. The note creation is happening before the `FeaturePlugin` class has been initialized, and therefore before any `dbDelta()` calls.

becdetat:
@octaedro @jeffstieler Can you try bumping `WC_ADMIN_VERSION_NUMBER` and reloading the page? This may trigger the db update before the table is accessed.

joshuatf:
@becdetat In our Zoom chat we talked about this a bit and I think we're on the same page on the issue now.


### The issue

The table changes in this PR are made in action scheduler so they may not be complete by the time a note is updated.

Unfortunately this means the version bump alone that I mentioned before will not work to resolve this.

### The fix

We need to make sure `dbDelta()` changes have occurred before we make any updates to notes.

We could run a manual check to see if pending actions exist in action scheduler to update the database.  But as an alternative, we could use `SchedulerTraits` to add notes via scheduled actions.  There is a method `get_dependencies()` where dependent actions can be defined so that those actions complete prior to inserting the note.

You can take a look at the class `ReportExporter` to see this in action.

Props to @rrennick for thinking of this and please feel free to add any comments if I missed anything from our discussion.

becdetat:
Thanks @joshuatf @rrennick I'll check out `SchedulerTraits`.

timmyc:
@becdetat based off our last conversation around this - I was under the impression that we were going to close this one out, but wanted to make sure you were okay with that.

becdetat:
True, I was hanging on to it by a shred of hope. I'll close it now.

PR #4085, opened 2020-04-08
Setup Checklist: Remove Setup Menu Item and I'll Do This Later link

### LINKED ISSUE #4067: Setup Checklist: Remove "Setup" Menu Feature
Per feedback from @pmcpinto - we are going to disable the "Setup" menu feature that was added in #3796 before the final release of 1.1. Some of the concerns around the feature are users might be following a tutorial to setup their Woo site, and by hiding the menu items, this could result in some confusion by end users.

We want to keep the other logic in that original PR that includes the new links to Skip / Do It Later, so thinking we could just remove the changes in the above PR to ` src/Features/OnboardingTasks.php` to "turn off" the menu changes. but /cc @becdetat for confirmation.

timmyc:
Fixes #4067

From feedback from @jameskoster and @pmcpinto - we have decided to remove the "Focus on the Setup Checklist" logic that was introduced in #3796 - so this branch removes the WooCommerce menu over-rides that hide all Woo menu items with a "Setup" menu item, and also removes the "I'll Do This Later" link under the setup checklist:

__Before__
before-task-list-focus

__After__
Dashboard ‹ WooCommerce ‹ My WordPress Site — WooCommerce 2020-04-08 10-22-17

- On a **new site** ( or setup your test site so the setup check list displays ), install WooCommerce 4.1 Beta
- Verify the screen looks like the above _before_ shot with Woo menu items hidden and "Setup" shown, and "I'll do this later" link shown.
- Install this branch as a plugin
- Verify the 'Setup' menu item is gone, as is the "I'll do this later" link

### Changelog Note:

Tweak: Remove Setup Menu


psealock (approved):
Testing well according to instructions and code looks good too. Thanks @timmyc

PR #4094, opened 2020-04-09
Use Route based code splitting to optimize bundle size

superdav42:
Implemented [code splitting](https://reactjs.org/docs/code-splitting.html) on each route so only the code needed to display the current route is loaded on the page. Greatly benefits pages which only show the header. Uncompressed JS+CSS size reduced from 926 kb to 627 kb.

Additionally, minification has been added to the production build - resulting in `*.min.js` files.

Further optimizations could be done if we split up `@woocommerce/components` but probably that's for a seperate PR. 

Also creates the command `npm run analyze` to see a breakdown of each bundle and what is inside of it.

### Screenshots
Header Only page on master:
header
Header only page on optimize-bundles:
Screenshot from 2020-04-08 23-54-41
Dashboard page on master:
dashboard
Dashboard page on optimize-bundles:
dashboard optimized

- Ensure `SCRIPT_DEBUG` is `true` in your `wp-config.php`
- Verify non-minified JS is enqueued (check network tab and page source)
- Navigate to all the pages and ensure they still load (smoke test everything)
- Ensure `SCRIPT_DEBUG` is `false` in your `wp-config.php`
- Verify minified JS is enqueued (check network tab and page source)
- Navigate to all the pages and ensure they still load (smoke test everything)
- Verify that `*.js` and `*.min.js` files are generated in both `development` and `production` modes (check that prod `min` files are minified)

### Changelog Note:

Performance: Implemented code splitting and lazy loading.


jeffstieler:
🎉 Just yesterday I started playing with a branch inspired by your work here: https://github.com/Automattic/woocommerce-services/pull/1994 - https://github.com/woocommerce/woocommerce-admin/compare/try/code-splitting

jeffstieler:
@rrennick do you think this PR impacts the i18n JSON scripts?

We're rewriting `.po` file comments beginning with `client/` to `dist/app/index.js` here: https://github.com/woocommerce/woocommerce-admin/blob/d53621206c5d31aecc678427518498154ad35831/bin/make-i18n-json.sh#L14

This PR breaks code out of the `dist/app/index.js` file and into many chunks, like `dist/chunks/activity-panels-inbox.[hash'.min.js`.

Will we need to somehow map/rewrite the comments to the appropriate chunks, or will things continue to work as expected despite this?

jeffstieler:
Nevermind @rrennick - I tested and figured out that this doesn't break any translations. In fact, our rewrite of the `.po` file references seems to be unnecessary and probably harmful.

webpack.config.js
```diff
@@ -220,23 +224,27 @@ const webpackConfig = {
...
 				flatten: true,
 				transform: ( content ) => content,
 			} ) )
 		),
-	],
+		new MomentTimezoneDataPlugin( {
+			startYear: 2000, // This strips out timezone data before the year 2000 to make a smaller file.
+		} ),
+		process.env.ANALYZE && new BundleAnalyzerPlugin(),
+	].filter( Boolean ),
```
senadir, line 247: I love this!

webpack.config.js
```diff
@@ -8,6 +8,9 @@ const CopyWebpackPlugin = require( 'copy-webpack-plugin' );
 const { DefinePlugin } = require( 'webpack' );
 const WebpackRTLPlugin = require( 'webpack-rtl-plugin' );
 const FixStyleOnlyEntriesPlugin = require( 'webpack-fix-style-only-entries' );
+const BundleAnalyzerPlugin = require( 'webpack-bundle-analyzer' )
```
senadir, line 11: I was getting really inconsistent results with webpack-bundle-analyzer (as it, many code end up being unmapped) so I switched to source-map-analyzer, did you get a similar issue?

client/index.js
```diff
@@ -12,6 +12,10 @@ import { PageLayout, EmbedLayout, PrimaryLayout as NoticeArea } from './layout';
 import 'wc-api/wp-data-store';
 import { withSettingsHydration } from '@woocommerce/data';
 
+// Modify webpack pubilcPath at runtime based on location of WordPress Plugin.
+// eslint-disable-next-line no-undef,camelcase
+__webpack_public_path__ = global.wcAdminAssets.path;
```
senadir, line 17: was always wondering how to solve this issue with dynamic loading, awesome!

webpack.config.js
superdav42: I didn't notice any problems like that with webpack-bundle-analyzer.

client/index.js
jeffstieler: Looks like this is causing a console error on non-WC-Admin pages. I'm investigating..

jeffstieler: Red herring! This isn't causing any issue.

rrennick:
> I tested and figured out that this doesn't break any translations. In fact, our rewrite of the .po file references seems to be unnecessary and probably harmful.

@jeffstieler The references need to be there for https://translate.wordpress.org/ to correctly populate the `.json` files. PRs that change the URLs of enqueued scripts need to update the references to update the references from the source script to the enqueued script.

jeffstieler:
@rrennick - I tested that the translation still worked on this branch (created a new `.po`/`.mo` pair that was used) - is there a specific set of steps I should be taking to verify the behavior you're referencing?

rrennick:
> is there a specific set of steps I should be taking to verify the behavior you're referencing?

@jeffstieler After adding some translation strings that have `.js` references run `npm run i18n:json` to produce the json files for the script files.

jeffstieler:
> After adding some translation strings that have .js references run npm run i18n:json to produce the json files for the script files.

Is that to say that the files being created means success, or is a lack of a diff what we're looking for?

rrennick:
> Is that to say that the files being created means success, or is a lack of a diff what we're looking for?

I test the js translation process by generating the files with text changes in some of the translations, generate the files then check the dashboard for the translations. For this PR, I'd translate `Inbox` to `Notifications`, etc. (at least one string for each enqueued script).

rrennick:
@jeffstieler this diff fixes the replacement script but I'm still getting `0 Files`. Possibly due to the string concatenation

```
--- a/bin/make-i18n-json.sh
+++ b/bin/make-i18n-json.sh
@@ -8,12 +8,12 @@ if [ ${WPCLI_VERSION:0:1} -lt "2" -o ${WPCLI_VERSION:0:1} -eq "2" -a ${WPCLI_VER
 fi
 
 # Substitute JS source references with build references
-for T in `find languages -name "*.po"`
+for T in `find languages -name "*.pot"`
        do
                sed \
-                       -e 's/ client\/[^:]*:/ dist\/app\/index.js:/gp' \
-                       -e 's/ packages\/components[^:]*:/ dist\/components\/index.js:/gp' \
-                       -e 's/ packages\/date[^:]*:/ dist\/date\/index.js:/gp' \
+                       -e 's/ client\/[^:]*:/ dist\/app\/index.min.js:/gp' \
+                       -e 's/ packages\/components[^:]*:/ dist\/components\/index.min.js:/gp' \
+                       -e 's/ packages\/date[^:]*:/ dist\/date\/index.min.js:/gp' \
                        $T | uniq > $T-build
                rm $T
                mv $T-build $T
```

jeffstieler:
@rrennick even without that patch it looks like things are working fine for me. Here's what I did:

* `rm -rf dist`
* `npm run build`
* `npm run i18n lang=fr_CA`
* Change language to French (Canada) in POEdit
* Translate "Customers" to "Canadians" in POEdit
* Save .po and .mo file to /languages
* `npm run i18n:json` (from VVV so wp-cli is available)

Doing that on both `master` and this branch resulted in the following files, that only differ in the timestamp values and some line numbers of string locations:

* `woocommerce-admin-fr_CA-1f921cef11f7f414d674d530e6965b98.json`
* `woocommerce-admin-fr_CA-23315bbd9ff4ba4a73d83f93aaea6484.json`
* `woocommerce-admin-fr_CA-946ffa241f2f7a67d85b6416fd6b9f50.json`
* `woocommerce-admin-fr_CA-cacb17be7bdb63bd3bb9c23d5a2305d7.json`
* `woocommerce-admin-fr_CA-d5abdf7f23e9521b725160dbd08295f1.json`
* `woocommerce-admin-fr_CA.mo`
* `woocommerce-admin-fr_CA.po`

rrennick:
> even without that patch it looks like things are working fine for me.

@jeffstieler I used GlotPress instead of poEdit and the translations also work for me. However, `npm run build` only creates the `.min` scripts. When `SCRIPT_DEBUG` is enabled the pages are blank due to lack of un-minified scripts.

jeffstieler:
> However, npm run build only creates the .min scripts. When SCRIPT_DEBUG is enabled the pages are blank due to lack of un-minified scripts.

@rrennick are you suggesting that we only minify on plugin/release builds instead of all `NODE_ENV=production` builds?

rrennick:
> are you suggesting that we only minify on plugin/release builds instead of all NODE_ENV=production builds?

This PR is fine from the i18n perspective :)

client/dashboard/index.js
```diff
@@ -1,17 +1,24 @@
...
  * Internal dependencies
  */
 import './style.scss';
-import CustomizableDashboard from './customizable';
-import ProfileWizard from '../profile-wizard';
 import withSelect from 'wc-api/with-select';
 import { isOnboardingEnabled } from 'dashboard/utils';
+import { Spinner } from '@woocommerce/components';
+
+const CustomizableDashboard = lazy( () =>
```
joshuatf, line 15: This prevents the styles for hiding the dashboard when the profiler is loaded added by `.woocommerce-admin-full-screen`.  Maybe util styles like that could be preloaded beforehand.

Screen Shot 2020-04-28 at 1 07 16 PM

joshuatf: Also, should we center the loading spinner in the above screen?

src/Loader.php
```diff
@@ -294,61 +294,63 @@ public static function register_scripts() {
 			return;
 		}
 
+		$suffix = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '' : '.min';
```
joshuatf, line 297: Turning `SCRIPT_DEBUG` to false resulted in missing files.  Do I need to do something else to get these minified files built?

Screen Shot 2020-04-28 at 2 18 09 PM

src/Loader.php
```diff
@@ -294,61 +294,63 @@ public static function register_scripts() {
...
+
 		wp_register_script(
 			'wc-csv',
-			self::get_url( 'csv-export/index.js' ),
+			self::get_url( 'csv-export/index' . $suffix . '.js' ),
```
joshuatf, line 301: Might be kind of cool if `get_url()` appended the suffix automatically or based on argument, but may be more trouble than its worth.

joshuatf:
Also noting that translations worked well (at least for non-minified versions). 😄

webpack.config.js
```diff
@@ -219,23 +225,37 @@ const webpackConfig = {
...
 				from: `./packages/${ packageName }/build-style/*.css`,
-				to: `./dist/${ packageName }/`,
+				to: `./${ packageName }/`,
 				flatten: true,
 				transform: ( content ) => content,
 			} ) )
 		),
-	],
+		new MomentTimezoneDataPlugin( {
+			startYear: 2000, // This strips out timezone data before the year 2000 to make a smaller file.
```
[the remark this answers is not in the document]
jeffstieler: I'll let @superdav42 answer - it's the same year chosen in his WCS PR: https://github.com/Automattic/woocommerce-services/pull/1994/commits/3738a398fc197236c300938234bba9a6db0ce3d2#diff-11e9f7f953edc64ba14b0cc350ae7b9dR222

My guess is that it's an arbitrary margin of safety before WC was created.

src/Loader.php
jeffstieler: I am unable to reproduce this - are you on commit 5a3ef99f93e700cc5a0e5aa0ab87b9ef44a50ac6?

Can you provide more details as to how you were building the app? `npm start` vs. `npm run build`, etc etc.

src/Loader.php
jeffstieler: Good idea - this also highlighted that I missed adding this suffix to the `wp-admin-scripts` directory!

jeffstieler: Done in 86926aed.

client/dashboard/index.js
jeffstieler: > This prevents the styles for hiding the dashboard when the profiler is loaded added by `.woocommerce-admin-full-screen`. Maybe util styles like that could be preloaded beforehand.

I don't think the styles aren't working - the class is added on component mount which is now delayed because of the lazy load. I had to throttle my browser to 3G to experience a lengthy delay.

I'll experiment with adding the class on the PHP side, but I'm not sure it will solve the problem completely.

> Also, should we center the loading spinner in the above screen?

I think ideally we'd use something other than spinners for these lazy loaded components. I'll see if it's straightforward to use the placeholder treatments from each instead - though that might eat into our bundle size savings a little (if possible).

src/Loader.php
joshuatf: This was under `npm start`.  I thought modifying `SCRIPT_DEBUG` was all that was needed to get use minified scripts.  If `npm run build` is required to get this working, then this is working as expected 👍

jeffstieler: That's very odd. I just ran `rm -rf dist/` and `npm start` and I see both minified and unminified files in the `dist/` folder.

client/dashboard/index.js
joshuatf: >I don't think the styles aren't working - the class is added on component mount which is now delayed because of the lazy load. I had to throttle my browser to 3G to experience a lengthy delay.

This happens to me locally when loading the profiler every time.  It's only for a second but the dashboard is visible for that brief moment.  I forgot that we switched these styles over to `woocommerce-admin-is-loading` which is added to the body class server-side.  If we can make sure that style is loaded beforehand, that should resolve it.

But we can tackle this in a follow-up so we don't block this PR further.

>I'll see if it's straightforward to use the placeholder treatments from each instead - though that might eat into our bundle size savings a little (if possible).

This would be awesome if we can find a reasonably lightweight solution.

src/Loader.php
joshuatf: Restarted a couple times and the JS files were oddly built in the root wc-admin folder.  But I deleted my local branch and pulled the latest changes and so far `npm start` seems to work as well for minified files.

I am getting one console error related to the homepage now:

Screen Shot 2020-04-28 at 8 36 40 PM

jeffstieler:
> This is testing well for me and LGTM provided the npm run build is the expected way to use minified files (which makes sense).

It's not - the work started in 7791882 was to always build both minified and unminified files.

> I'm not entirely sure if we need the changes in 93a6369 if those are about the profiler loading screen discussed in #4094 (comment), but pre-approving pending that commit's removal.

That change doesn't address the comment, it fixes missing styles for the OBW on the new home screen. It's not directly related to this PR, but I found it when investigating your comment.

src/Loader.php
jeffstieler: I'm looking into this and will comment if I figure out the resolution, but... paging @psealock - this error goes away if I comment these lines out: https://github.com/woocommerce/woocommerce-admin/blob/717be1ab5183a7d4d0452a6f78a9917e9a6ae7ae/client/profile-wizard/index.js#L244-L250

jeffstieler: Oh and to fill in some missing context - this problem was introduced by lazy loading the `ProfileWizard` component. If we remove the `lazy` and `Suspense` .. it works fine.

psealock: This is a really odd one because the Warning about the render function not being pure is wrong. If you replace the output of ProfileWizard to be a string, the error occurs still.

If we go back to hydrating these components outside of the `compose` with `withSelect` and `withDispatch` etc, then the error goes away. For example:

```js
const HydradedProfileWizard = withSettingsHydration( 'general', {
	general: window.wcSettings.preloadSettings.general,
} )( ProfileWizard );
```
I'm not sure why that would be.

psealock: I went ahead and pushed a fix in https://github.com/woocommerce/woocommerce-admin/pull/4094/commits/46572ed05bf34df2e2a0ca00a3fb602c10f840b4. Feel free to revert if there is a better way. If we do stick with this, we should also update other instances of this pattern of lumping `with*Hydration` into the `compose` for wp.data functions.

psealock: Also want to note here that this expression results in an error when the ternary is false:

```js
hydrateSettings
? withSettingsHydration( 'general', {
		general: window.wcSettings.preloadSettings.general,
	} )
: () => ( {} ),
```
I think `() => ( {} )` should be something more like `() => Comp => ( <Comp /> )`;

jeffstieler: > I think () => ( {} ) should be something more like () => Comp => ( <Comp /> );

Yep, you're right - I'll change to use lodash's `identity()`.

jeffstieler:
It looks like that initial state error was being caused by an improper order of operations. The `compose()` function is simply a [reexported lodash `flowRight()`](https://github.com/WordPress/gutenberg/blob/master/packages/compose/src/higher-order/compose.js#L4).

I also fixed the fallback to be lodash's `identity()` rather than returning an empty object instead of the passed component 🤦.

A quick set of 👀 would be great before I merge this @joshuatf.

webpack.config.js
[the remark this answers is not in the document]
superdav42: I think I picked it because it was in the example and it's a few years before WordPress was released. It's admittedly very conservative but since I'm not familiar with how moment is used in wc-admin I picked a safe value.

PR #4105, opened 2020-04-09
Added check to show WC Pay option only when Jetpack is connected

### LINKED ISSUE #4092: Setup Checklist: WC Pay Option Shown when Jetpack is not connected
**Describe the bug**
Currently in `master` and the `release\1.1.0` branch, the WC Pay option is shown in the Setup Checklist payment step when Jetpack is installed and activated, but **not connected**. This leads to the "Setup Step" on the install and activate screen failing with a 500 error from the REST API because an active Jetpack connection is required.

**To Reproduce**

1. Using `master` or https://github.com/woocommerce/woocommerce-admin/releases/download/v1.1.0-plugin/woocommerce-admin.zip
2. Venture through the OBW, ensure you use a US-based address.
3. On the Setup Checklist, go to payments

Dashboard ‹ WooCommerce ‹ Extraordinary Winter — WooCommerce 2020-04-08 17-01-38

4. Click 'Setup' next to WooCommerce Payments
5. Note the plugin is installed and activated fine, next click the "Verify" button:

Dashboard ‹ WooCommerce ‹ Extraordinary Winter — WooCommerce 2020-04-08 17-02-08

you should then get an error:

Dashboard ‹ WooCommerce ‹ Extraordinary Winter — WooCommerce 2020-04-08 17-02-40

Dashboard ‹ WooCommerce ‹ Extraordinary Winter — WooCommerce 2020-04-08 17-03-25

**Expected behavior**
WooCommerce Payments option should only be shown when Jetpack is connected.

**Additional context**
Tested on a new JN site.

/cc @allendav

octaedro:
Fixes #4092

In this PR a verification was added to the Setup Checklist payment step, to show WC Pay option only when Jetpack is connected.

### Screenshots
If the site is not connected to Jetpack the WC Pay option shouldn't be visible on the setup checklist payment step:
Dashboard ‹ WooCommerce ‹ Extraordinary Winter — WooCommerce 2020-04-08 17-01-38

- The solution won't work with a site with Jetpack in `Development mode`, so it would be a good idea to use an ephemeral instance to test.
- Install Jetpack but don't connect it.
- Install and activate `WooCommerce` and a `test release` with this code.
- Go to the setup checklist payment step (URL: `/wp-admin/admin.php?page=wc-admin&task=payments`).
- The WC Pay option shouldn't be visible.
- Go to the Jetpack dashboard and connect it (URL: `/wp-admin/admin.php?page=jetpack#/dashboard`).
- Go back to the setup checklist payment step.
- The WC Pay option should be visible now.

### DIFF
--- a/client/dashboard/task-list/tasks/payments/methods.js
+++ b/client/dashboard/task-list/tasks/payments/methods.js
@@ -33,6 +33,7 @@ import PayFast from './payfast';
 export function getPaymentMethods( {
 	activePlugins,
 	countryCode,
+	isJetpackConnected,
 	options,
 	profileItems,
 } ) {
@@ -121,7 +122,7 @@ export function getPaymentMethods( {
 			visible:
 				[ 'US' ].includes( countryCode ) &&
 				! hasCbdIndustry &&
-				profileItems.jetpack_connected,
+				isJetpackConnected,
 			plugins: [ 'woocommerce-payments' ],
 			container: <WCPay />,
 			isConfigured: wcPayIsConfigured,
--- a/client/dashboard/task-list/tasks/payments/methods.js
+++ b/client/dashboard/task-list/tasks/payments/methods.js
@@ -118,7 +118,10 @@ export function getPaymentMethods( {
 				</Fragment>
 			),
 			before: <WCPayIcon />,
-			visible: [ 'US' ].includes( countryCode ) && ! hasCbdIndustry,
+			visible:
+				[ 'US' ].includes( countryCode ) &&
+				! hasCbdIndustry &&
+				profileItems.jetpack_connected,
 			plugins: [ 'woocommerce-payments' ],
 			container: <WCPay />,
 			isConfigured: wcPayIsConfigured,
--- a/client/dashboard/task-list/tasks/payments/index.js
+++ b/client/dashboard/task-list/tasks/payments/index.js
@@ -352,6 +352,7 @@ export default compose(
 			getActivePlugins,
 			getOptions,
 			getUpdateOptionsError,
+			isJetpackConnected,
 			isUpdateOptionsRequesting,
 		} = select( 'wc-api' );
 
@@ -378,6 +379,7 @@ export default compose(
 		const methods = getPaymentMethods( {
 			activePlugins,
 			countryCode,
+			isJetpackConnected: isJetpackConnected(),
 			options,
 			profileItems,
 		} );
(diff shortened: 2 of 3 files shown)


client/dashboard/task-list/tasks/payments/methods.js
```diff
@@ -118,7 +118,10 @@ export function getPaymentMethods( {
 				</Fragment>
 			),
 			before: <WCPayIcon />,
-			visible: [ 'US' ].includes( countryCode ) && ! hasCbdIndustry,
+			visible:
+				[ 'US' ].includes( countryCode ) &&
+				! hasCbdIndustry &&
+				profileItems.jetpack_connected,
```
psealock, line 124: We have a selector that performs this function:

https://github.com/woocommerce/woocommerce-admin/blob/7c0d4a34c6e213c990b16adb8ebc6b68d7a2e109/client/wc-api/onboarding/selectors.js#L95

Under the hood, it uses `wcSettings.dataEndpoints.jetpackStatus`.

timmyc: Ah interesting, thanks for that @psealock - I was the one who suggested the current approach :neckbeard:

octaedro: Hi @psealock, my first approach was using `isJetpackConnected` but later I changed my mind when I saw that it would be a cleaner solution to send the `jetpack connection` state in the `profileItems` param, that already has the `wccom_connected` value.
If you think that using `isJetpackConnected` is better I can go for it easily, I did this solution earlier :laughing:

psealock: I do think it would be better than adding it to `profileItems` because when  https://github.com/woocommerce/woocommerce-admin/pull/4048/ gets merged, all the plugin information will be moved to its own data store.

> a cleaner solution to send the jetpack connection state in the profileItems param, that already has the wccom_connected value.

I hadn't noticed that `wccom_connected` was in profileItems 😄  . I think as part of #4048, `wccom_connected` should also be moved to the new plugins data store (or in a follow up PR).

For now, I think `isJetpackConnected` if you don't mind. That will make the refactor a bit easier. Thanks!

octaedro: Oh, now I see!
Good point @psealock
I'll go for it :slightly_smiling_face:
Thank you!

timmyc: Also that approach might clear up the failing test on CI:

1) WC_Tests_API_Onboarding_Profiles::test_schema

Failed asserting that actual size 16 matches expected size 15.

/tmp/wordpress/wp-content/plugins/woocommerce-admin/tests/api/onboarding-profile.php:102

timmyc:
Thanks @octaedro and @psealock for the teamwork on the fix here, y'all da best!

PR #4111, opened 2020-04-10
Onboarding - business details step: increase max-height on competitors listbox

### LINKED ISSUE #4095: Onboarding - business details step: increase max-height on competitors listbox
The user needs to scroll the listbox in order to see all the competitors available on the list. Increasing the listbox max-height to something like 550px would allow users to check all the competitors without scrolling.

octaedro:
Fixes #4095

In this PR the competitors' Listbox in the Business Details step in the Onboarding process has been enlarged. 

### Screenshots
competitors-listbox

- Go to the `Business Details` step in the Onboarding (URL: `/wp-admin/admin.php?page=wc-admin&step=business-details`).
- In the `Currently selling elsewhere?` listbox, choose the option `Yes, on another platform`.
- Verify that all the competitors are visible at once after clicking the `Which platform is the store using?` listbox.

### Notes
It looks like the [GH diff](https://github.com/woocommerce/woocommerce-admin/pull/4111/files) is acting a little funny since the only change that I made to the file `client/dashboard/profile-wizard/steps/business-details.js` was to add a `div` with `className='business-competitors'` and an extra `tab` to the lines inside the `div`.

### DIFF
--- a/client/dashboard/profile-wizard/style.scss
+++ b/client/dashboard/profile-wizard/style.scss
@@ -29,7 +29,7 @@
 		}
 
 		.business-competitors .woocommerce-select-control__listbox {
-			max-height: 504px;
+			max-height: unset;
 		}
 	}
 
--- a/client/dashboard/profile-wizard/style.scss
+++ b/client/dashboard/profile-wizard/style.scss
@@ -27,6 +27,10 @@
 			max-width: 100%;
 			justify-content: center;
 		}
+
+		.business-competitors .woocommerce-select-control__listbox {
+			max-height: 504px;
+		}
 	}
 
 	.woocommerce-profile-wizard__card-actions {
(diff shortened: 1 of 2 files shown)


client/dashboard/profile-wizard/style.scss
```diff
@@ -27,6 +27,10 @@
 			max-width: 100%;
 			justify-content: center;
 		}
+
+		.business-competitors .woocommerce-select-control__listbox {
+			max-height: 504px;
```
jeffstieler, line 32: This works, but what do you think about using `unset` instead of specifying a height?

octaedro: Good idea! I changed it in the commit [c79716e](https://github.com/woocommerce/woocommerce-admin/pull/4111/commits/c79716e5970f28415cf0d6775de5e1baab27814b).
Thanks!

PR #4115, opened 2020-04-10
Fix usage of WP_Error in non-global namespaces.

jeffstieler:
Fix fatal errors from unescaped usage of `WP_Error` in namespaces.

Found while testing https://github.com/woocommerce/woocommerce-admin/pull/4048.

- Force a non-200 response from any API call in `src/API/OnboardingPlugins.php` that had it's `WP_Error` instantiation modified
- Verify no PHP Fatal Errors occur

### Changelog Note:

Fix: namespaces in Plugins API error handling.


psealock (approved):
Thanks for the cleanup @jeffstieler 🚢

Issue #4106, opened 2020-04-09
[GlobalStep - 4.1 beta 1] On Safari(v13.0.5), "Setup Payment" page only "Done" button is displayed.


gglobalstep:
**Step to Reproduce:**
1. Create a test site using "https://jurassic.ninja/".
2. Complete initial screens and go to dashboard.
3. Go to "Setup" screen.
4. Click on "Setup Payments".
5. Note that blank screen is displayed. 
6. Observe that only "Done" button is displayed.

**Actual Result:**
On Safari(v13.0.5), "Setup Payment" page only "Done" button is displayed.

**Expected Result:**
All payment options should be displayed upon clicking "Setup Payment".

![19](https://user-images.githubusercontent.com/41110392/78920779-8c4edc80-7ab1-11ea-9550-5d1661a032b3.jpg)

timmyc:
@gglobalstep do you have any details around the store address, specifically the country, used in this test? Also which "Industries" did you select. I'm guessing this might be the same bug that @juliaamosova and @vedanshujain hit when testing Beta 2 today which would be the following:

- A non US address was entered in the OBW Store Details ( Step 1 )
- On the Industry page ( step 2 ), "CBD and other hemp-derived products" was selected

The bug is happening because the only payment option that supports CBD sales is Square, and that method [is only visible for US stores](https://github.com/woocommerce/woocommerce-admin/blob/master/client/dashboard/task-list/tasks/payments/methods.js#L263).

The solution I think we should do at this time is to remove "CBD and other hemp-derived products" from the Industries page, unless the store address is in the USA. Does that sound good to you @pmcpinto ?

psealock:
The solution I think we should do at this time is to remove "CBD and other hemp-derived products" from the Industries page, unless the store address is in the USA. 

Agreed, this solves the problem by eliminating this situation.

https://github.com/woocommerce/woocommerce-admin/pull/4117

pmcpinto:
> The solution I think we should do at this time is to remove "CBD and other hemp-derived products" from the Industries page, unless the store address is in the USA. Does that sound good to you @pmcpinto ?

Yep, let's go with this solution. Thanks

PR #4117, opened 2020-04-12
OBW: allow CBD only for US stores

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/4106

Industry type CBD should only be displayed to US based stores. This PR filters the industry list based on store default location to remove the CBD option for non-US based stores.

See p1586527660260000-slack-proton for more info.

1. Fire off the OBW.
2. Set a US store location.
3. Proceed to industry.
4. See the CBD option.
5. Go back to Store Details and enter a non-US location.
6. See the CBD industry option no longer available.
7. When clicking "Continue", see the Network request to `/onboarding/profile` and make sure the CBD option is not also sent.

### DIFF
--- a/client/dashboard/profile-wizard/steps/industry.js
+++ b/client/dashboard/profile-wizard/steps/industry.js
@@ -165,7 +165,7 @@ class Industry extends Component {
 		);
 		const industryKeys = Object.keys( industries );
 
-		const filteredIndustries =
+		const filteredIndustryKeys =
 			region === 'US'
 				? industryKeys
 				: industryKeys.filter(
@@ -185,7 +185,7 @@ class Industry extends Component {
 				</p>
 				<Card>
 					<div className="woocommerce-profile-wizard__checkbox-group">
-						{ filteredIndustries.map( ( slug ) => {
+						{ filteredIndustryKeys.map( ( slug ) => {
 							const selectedIndustry = find( selected, { slug } );
 
 							return (
--- a/client/dashboard/profile-wizard/steps/industry.js
+++ b/client/dashboard/profile-wizard/steps/industry.js
@@ -38,7 +38,7 @@ class Industry extends Component {
 		 * Calls to `updateProfileItems` in the previous screen happen async
 		 * and won't be updated in wc-api's state when this component is initialized.
 		 * As such, we need to make sure cbd is not initialized as selected when a
-		 * user has changed location to non0-US based.
+		 * user has changed location to non-US based.
 		 */
 		const { locationSettings } = props;
 		const region = getCurrencyRegion(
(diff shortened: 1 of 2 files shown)


psealock:
Great call @octaedro! Thats a nice edge case to catch now.

I added a check when saving Store Details to see if CBD needed to be removed from industries. There is a problem, to which I've created a temporary solution. This issue is` updateProfileItems` does not respect the `await` and the next screen is then intialized with outdated with industry items, which will persist even when not visible.

Then core issue is with Fresh Data giving outdated values. This will be resolved when OBW profile items gets its own data store and we can correctly use async/await.

## Additional Test Step
1. Select a US based store.
2. Select CBD industry, proceed to next step.
3. Go back and change to a non-US store location
4. Confirm the CBD option is no longer in the industries step.
5. When clicking "Continue", see the Network request to `/onboarding/profile` and make sure the CBD option is not also sent.

client/dashboard/profile-wizard/steps/industry.js
```diff
@@ -25,11 +27,38 @@ const onboarding = getSetting( 'onboarding', {} );
...
+		/**
+		 * @todo Remove block on `updateProfileItems` refactor to wp.data dataStores.
+		 *
+		 * The following block is a side effect of wc-api not being truly async
+		 * and is a temporary fix until a refactor to wp.data can take place.
+		 *
+		 * Calls to `updateProfileItems` in the previous screen happen async
+		 * and won't be updated in wc-api's state when this component is initialized.
+		 * As such, we need to make sure cbd is not initialized as selected when a
+		 * user has changed location to non0-US based.
```
jeffstieler, line 41: Total nitpick - typo: `non0-US`

client/dashboard/profile-wizard/steps/industry.js
```diff
@@ -130,6 +159,22 @@ class Industry extends Component {
 	render() {
 		const { industries } = onboarding;
 		const { error, selected, textInputListContent } = this.state;
+		const { locationSettings } = this.props;
+		const region = getCurrencyRegion(
+			locationSettings.woocommerce_default_country
+		);
+
+		const filteredIndustries = Object.keys( industries ).filter(
```
jeffstieler, line 167: Nitpick: industries only need to be filtered if the region isn't the US.

client/dashboard/profile-wizard/steps/store-details.js
```diff
@@ -117,7 +118,36 @@ class StoreDetails extends Component {
...
 
-		await updateProfileItems( { setup_client: values.isClient } );
+		const profileItemsToUpdate = { setup_client: values.isClient };
+		const cbdSlug = 'cbd-other-hemp-derived-products';
+		const hasCbdIndustry =
+			profileItems.industry &&
+			profileItems.industry.length &&
+			profileItems.industry.some( ( industry ) => {
+				return cbdSlug === industry || cbdSlug === industry.slug;
+			} );
```
jeffstieler, line 128: Nitpick: since this block already iterates all of the `profileItems.industry` array, it's not offering any performance benefit in the `if()` statement below. We could just always filter the `profileItems.industry` array for non-US `region`s.

psealock: Good call 😄

client/dashboard/profile-wizard/steps/industry.js
psealock: 👍

PR #4133, opened 2020-04-15
Fix confusing UX when knowledge base card fails to retreive posts

danielbitzer:
If the Marketing Tab knowledge base posts can't be retrieved for whatever reason the card is kind of broken. See screenshots below.

This PR:
- Uses the `EmptyContent` component to display feedback to the user (see screenshot)
- Reduces the cache timeout when no posts are found from **24 hours** to **15 minutes**. This means users aren't unnecessarily shown an EmptyContent message for 24 hours. Posts will be retrieved sooner when their internet or WC.com is back online.

### Accessibility

### Screenshots

Before: 
Screenshot
Screenshot: https://d.pr/i/PhuT1s 

After: 
Screenshot
Screenshot: https://d.pr/i/dsX2vV 

- Check out this branch locally
- Go offline, turn off your internet
- Delete the transient `wc_marketing_knowledge_base`
- Load the Marketing Tab
- You should see the EmptyContent state shown above

### DIFF
--- a/client/marketing/overview/knowledge-base/index.js
+++ b/client/marketing/overview/knowledge-base/index.js
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
-import { Component, Fragment } from '@wordpress/element';
+import { Component } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import { __ } from '@wordpress/i18n';
 import { Spinner } from '@wordpress/components';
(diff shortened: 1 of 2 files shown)


danielbitzer:
> icon feels quite large to me 

I actually reduced it heaps from how it's typically used. I checked and this component is actually very widely used in WC Admin however I didn't check exactly how it looks in other places...

danielbitzer:
It is pretty chunky
Screenshot
Screenshot: https://d.pr/i/976cSH

jconroy:
Gotcha, chunky it is!

timmyc:
Confirming that this should be cherry-picked into the release for 4.1?

jconroy:
@timmyc yep that's correct - I've added the "Needs Cherry Picking" tag. Will merge now.

PR #4141, opened 2020-04-16
Add profiler step view tracks

### LINKED ISSUE #4136: Add missing event in JP/WCS splash screen
I just noticed that in the issue that covered all the changes in the JP/WCS splash screen (#3709) I asked to include an event to track pageviews in this screen, but the event name that I provided is the same that we have for clicks: 🤦

> We need to change the name of the event that's triggered when the user clicks in one of the buttons on this screen (wcadmin_storeprofiler_welcome_clicked) to **wcadmin_storeprofiler_install_plugins**. The event props stay the same.
> We need to add a new event to track when we just have a pageview on this screen: **wcadmin_storeprofiler_install_plugins**

Without the pageviews event we're not able to measure the conv. rate of this screen.

We need to add a new event to track the pageviews:

Event name: `wcadmin_storeprofiler_install_plugins_view`
Event prop: `plugins`(to track if the user sees the page with just JP info, just WCS info or both)

joshuatf:
Fixes #4136 

Adds a track for viewing each step in the profiler.

1. Enter localStorage.setItem( 'debug', 'wc-admin:*' ); into your console. Leave your console open.
1. Enable the profiler.
1. Visit a profiler step and make sure the event is recorded `wcadmin_storeprofiler_step_view` with the correct step name in the console.
1. Change steps via the navigation or completing a step.
1. Make sure an event is fired for each step.

### DIFF
--- a/client/dashboard/profile-wizard/index.js
+++ b/client/dashboard/profile-wizard/index.js
@@ -65,14 +65,13 @@ class ProfileWizard extends Component {
 			window.document.documentElement.scrollTop = 0;
 
 			recordEvent( 'storeprofiler_step_view', {
-				step,
+				step: this.getCurrentStep().key,
 			} );
 		}
 	}
 
 	componentDidMount() {
-		const { profileItems, query, updateProfileItems } = this.props;
-		const { step } = query;
+		const { profileItems, updateProfileItems } = this.props;
 
 		document.documentElement.classList.remove( 'wp-toolbar' );
 		document.body.classList.add( 'woocommerce-onboarding' );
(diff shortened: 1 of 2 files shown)


joshuatf:
@pmcpinto I added an event for each step view.  I think between this track and `wcadmin_storeprofiler_already_installed_plugins` we're covered in #4136, but let me know if there's any more information you can't get from this and we can adjust.

pmcpinto:
@joshuatf thanks. This is going to be shipped in 4.1, correct? The content that we display in the JP/WCS is different depending on if the user already has one of the extensions or none of them. So it would be useful to have an event prop identifying that. But if it takes a bit of work (more than 2-3h) I would backlog it and implement this prop in the future.

joshuatf:
@pmcpinto Gotcha.  Added in a `wcadmin_storeprofiler_plugins_to_install` event that has a `plugins` property containing plugins that need to be installed.

@timmyc Can we squeeze this in the 1.1 release?

joshuatf:
Great catch @octaedro!  Updated in https://github.com/woocommerce/woocommerce-admin/pull/4141/commits/e08056dec67b00f2e418831e61f4cbeea81b0c21 - can you take another look?

PR #4143, opened 2020-04-17
Remote inbox notification delivery

### LINKED ISSUE #3878: Remote Inbox Messaging
With the upcoming changes coming to the wc-admin dashboard (pbIJXs-1w-p2), the inbox notification panel is going to occupy a prominent place. So it's more important than never to find a way to control the notifications without waiting for a new release. Ideally, we should be able to control the following:

- Add new notifications
- Delete existing notifications
- Editing the notification content
- Editing the notification criteria (store type, trigger, cadence)

We would like to understand the following:
- Is it possible to have this control over notifications that are sent to users that are connected to woocommerce.com and/or Jetpack?
- What implications does this change is going to have in the existing system?
- Do we need to consider extensibility for the v1 or is it something that can be easily changed in a future version? Marketplace would like to use the inbox notifications as a premium tool for vendors.

At the end of the cooldown it would be great to have the following:
- A rough estimation of how long it can take this implementation
- If we can reduce technical complexity to improve development time
- A list with the largest modules of work

becdetat:
Fixes #3878

p90Yrv-1Hr-p2

The shop polls data sources (TBC) for a JSON file containing specs. Each spec contains a number of rules and information about an admin note that would be created if the rules pass the engine.

Up for conversation:

- how often to poll the JSON data source(s)
- how often to run the engine
- ideas for other rules
- everything else ;-)

There's also a P2 post to discuss what would be needed for a [server-side implementation](https://teamisotope.wordpress.com/2020/05/05/rule-based-notifications-server-side-implementation/).

### JSON description

A test JSON specs file is available from [here](https://gist.githubusercontent.com/becdetat/f17e9617e5ca1211d0579cf079862905/raw/7f0e44287b21cbf53a632887a805a414137000ce/rinds-specs.json).

#### JSON file:

```
[
    <spec>,
    ...
]
```

#### Spec:

```
{
    "slug": "ad-for-automate-woo-2020-04-20",
    "type": "info",
    "icon": "info",
    "status": "unactioned",
    "is_snoozable": 0,
    "source": "woocommerce.com",
    "locales": [
        <Locale>,
        ...
    ],
    "actions": [
        <Action>,
        ...
    ],
    "rules": [
        <Rule>,
        ...
    ]
```

#### Locale

```
{
    "locale": "en_US",
    "title": "Ad for Automate Woo",
    "content": "Hi There! This is an ad for Automate Woo."
},
{
    "locale": "en_AU",
    "title": "Ad for Automate Woo, in en-AU.",
    "content": "en-AU content"
},
{
    "locale": "fr_FR",
    "title": "Annonce pour automatiser woo",
    "content": "Salut! Ceci est une publicité pour Automate Woo."
}
```

#### Action

```
{
    "name": "install-automate-woo",
     "locales": [
        <ActionLocale>,
        ...
    ],
    "url": "?page=automatewoo-dashboard",
    "url_is_admin_query": true,
    "is_primary": true,
    "status": "actioned"
}
```

#### ActionLocale

```
{
    "locale": "en_US",
    "label": "Install"
}
```

#### Rule

Rules in an array are executed as an AND operation. If there are no rules in the array the result is `false`.

Plugins activated:

```
{
    "type": "plugins_activated",
    "plugins": [
        "plugin-slug-1',
        "plugin-slug-2"
    ]
}
```

Publish after time:

```
{
    "type": "publish_after_time",
    "publish_after": "2020-04-22 00:00:00"
}
```

Publish before time:

Note that using both `publish_after_time` and `publish_before_time` allows timeboxing a note which could be useful for promoting a sale.

```
{
    "type": "publish_before_time",
    "publish_before": "2020-04-22 00:00:00"
}
```

Not:

Note that the rules in `operand` get ANDed together into a single boolean.

```
{
    "type": "not",
    "operand": [
        <Rule>,
        ...
    ]
}
```

Or:

Note that the rules in `left_operand` and `right_operand` get ANDed together into a single boolean value for the operand.

```
{
    "type": "or",
    "operands": [
        [
            <Rule>,
            ...
        ],
        [
            <Rule>,
            ...
        ]
    ]
}
```

alternatively:

```
{
    "type": "or",
    "operands": [
        <Rule>,
        ...
    ]
}
```

Fail:

This just returns a `false` value.

```
{
    "type": "fail"
}
```
Plugin version:

This compares the installed version of the plugin to the required version, using the operator. If the plugin isn't activated this returns false.

```
{
    "type": "plugin_version",
    "plugin": "jetpack",
    "version": "8.4.1",
    "operator": "="
}
```

- Switch to this PR's branch.
- Note that this adds a feature flag so `npm start` will have to be re-run
- Install the WP Crontrol plugin to access cron jobs via Tools -> Cron events. Running the `wc_admin_daily` hook triggers a poll, and also triggers a run of the engine at the moment. This will probably be separated out to different events though as the engine should probably be run more frequently than a daily poll. Activating and deactivating plugins also triggers an engine run (but not a poll).
- To check which specs have triggered, use a page that shows the admin notes (such as the WooCommerce dashboard).
- Copy [this gist](https://gist.githubusercontent.com/becdetat/f17e9617e5ca1211d0579cf079862905/raw/99495c525c54663f6a0b90edff81e9ccd5b72afb/rinds-specs.json) somewhere it will be picked up by a local web server. For example, if you're using vvv, `<vvv root>/www/wordpress-two/public_html/rinds-specs.json` will be available at http://two.wordpress.test/rinds-specs.json. (@mattsherman updated link to gist on 2020-05-27)
- Update `DataSourcePoller::DATA_SOURCES` to poll the above file
- In the spec, set the `ad-for-automate-woo-2020-04-22` spec's rule's `publish_after` value to some time in the future.
- If you have WooCommerce Services or Jetpack activated, deactivate them
- Delete all records in the `wp-wc-admin-notes` table
- Trigger a poll

Let's get testing:

- set the `ad-for-automate-woo-2020-04-22` spec's rules's `publish_after_time` time to some time in the past; make sure the `publish_before_time` time is set to some time in the future (@mattsherman updated this on 2020-05-27)
- trigger a poll
- check that the new note appears 

- install and activate both Jetpack and WooCommerce Services
- check that the `more-action-for-jp-wcs` note appears

- deactivate Jetpack
- check that the `NOT-test` note appears and the `ad-for-automate-woo-2020-04-22` note disappears

- test that the `OR-test` note appears if either (or both) of the Jetpack or WooCommerce Services plugins are activated

### Changelog Note:

Enhancement: Remote inbox notification delivery


src/Rinds/RindsSpecsDataSourcePoller.php
```diff
@@ -0,0 +1,75 @@
...
+			self::read_data_source( $url, $specs );
+		}
+
+		// Persist the specs as an option.
+		$existing_option = get_option( RindsEngine::SPECS_OPTION_NAME );
+
+		if ( false === $existing_option ) {
+			add_option( RindsEngine::SPECS_OPTION_NAME, $specs );
+		} else {
+			update_option( RindsEngine::SPECS_OPTION_NAME, $specs );
```
joshuatf, line 40: `update_option()` will add the option if it does not exist, so we can forego this `if/else` logic with a single line to update the option.

src/Rinds/RindsSpecsDataSourcePoller.php
```diff
@@ -0,0 +1,75 @@
...
+	const DATA_SOURCES = array(
+		'http://two.wordpress.test/rinds-specs.json',
+	);
+
+	/**
+	 * Polls the data sources for specs, persists those specs, and delete any
+	 * specs that no longer exist in the data sources.
+	 */
+	public static function poll_data_sources() {
+		$specs = array();
```
joshuatf, line 27: What's this variable used for?  Will this be updated eventually?

src/Rinds/RindsSpecsDataSourcePoller.php
```diff
@@ -0,0 +1,75 @@
...
+	/**
+	 * Read a single data source and insert the read specs into $specs, using
+	 * the spec slug as the key. So if the same or multiple data source has
+	 * a spec with the same slug, the last spec wins.
+	 *
+	 * @param string $url   The URL to read the specs from.
+	 * @param array  $specs The specs to read in to.
+	 */
+	private static function read_data_source( $url, &$specs ) {
+		$response = wp_remote_get( $url );
```
joshuatf, line 57: This may be something for a follow-up, but I imagine we'll need to be able to pass headers or site information with this request.

src/Rinds/RindsSpecsDataSourcePoller.php
```diff
@@ -0,0 +1,75 @@
...
+
+		if ( is_wp_error( $response ) ) {
+			return;
+		}
+
+		$body = $response['body'];
+		$json = json_decode( $body );
+
+		if ( null === $json ) {
```
joshuatf, line 66: Maybe we should check `if ( ! is_array( $json ) )` to prevent a bad response from being looped over.

tests/rinds/plugin-version-rule-processor.php
```diff
@@ -0,0 +1,158 @@
...
+ *
+ * @package WooCommerce\Tests\Rinds
+ */
+
+use Automattic\WooCommerce\Admin\Rinds\PluginVersionRuleProcessor;
+
+/**
+ * class WC_Tests_Rinds_PluginVersionRuleProcessor
+ */
+class WC_Tests_Rinds_PluginVersionRuleProcessor extends WC_Unit_Test_Case {
```
joshuatf, line 13: Thanks for adding all these tests! 🎉

src/Rinds/OrRuleProcessor.php
```diff
@@ -0,0 +1,45 @@
...
+
+	/**
+	 * Performs an OR operation on the rule's left and right operands.
+	 *
+	 * @param object $rule The specific rule being processed by this rule processor.
+	 *
+	 * @return bool The result of the operation.
+	 */
+	public function process( $rule ) {
+		$evaluated_left_operand = $this->rule_evaluator->evaluate( $rule->left_operand );
```
joshuatf, line 35: Does this mean we're limited to two operands with `OR` or is there a way to add more?

src/Rinds/RindsSpecsDataSourcePoller.php
joshuatf: Do we have an idea of how many specs/notes will be stored in a source?

If the list is small, this may not be an issue.  But if data sources store many messages and aren't pruned over time, we'll also probably want to pass an id or a last fetched timestamp so we're not re-checking a huge data source over time.  Pagination could be also something to look at here.

becdetat:
@joshuatf Thanks for the review! I appreciate it.

> > how often to poll the JSON data source(s)
>
> I think the current daily is good, but I can also see an argument for up to twice a day.

I'll leave it as daily for now.

> > how often to run the engine
>
> I think same as above. Do you think there's a reason for them to be different?

I dis at first, because there are time-sensitive notes, but the really reactionary notes get created in response to actions (eg. activating a plugin) anyway.

> > ideas for other rules
>
> Probably quite a few may be needed eventually, but definitely for follow-ups and maybe as needed as this PR is already pretty massive.

I'll leave them for other PRs. I've already got another PR up for "creating your first product".

src/Rinds/FailRuleProcessor.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+namespace Automattic\WooCommerce\Admin\Rinds;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Rule processor that fails.
+ */
+class FailRuleProcessor {
```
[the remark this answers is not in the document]
becdetat: The idea of this was to give the developers of the JSON file (WooCommerce.com) an easy way to void a spec and action the note, since just deleting the spec orphans the note.

src/Rinds/GetRuleProcessor.php
```diff
@@ -0,0 +1,57 @@
...
+class GetRuleProcessor {
+	/**
+	 * Get the processor for the specified rule type.
+	 *
+	 * @param string $rule_type The rule type.
+	 *
+	 * @return object The matching processor for the specified rule type, or a FailRuleProcessor if no matching processor is found.
+	 */
+	public static function get_processor( $rule_type ) {
+		if ( 'plugins_activated' === $rule_type ) {
```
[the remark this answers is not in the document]
becdetat: Nice, I've done this

src/Rinds/RindsSpecsDataSourcePoller.php
```diff
@@ -0,0 +1,75 @@
...
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * RINDS specs data source poller class.
+ * This handles polling RINDS specs from JSON endpoints, and
+ * stores the specs in to the database as an option.
+ */
+class RindsSpecsDataSourcePoller {
+	const DATA_SOURCES = array(
+		'http://two.wordpress.test/rinds-specs.json',
```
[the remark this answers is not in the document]
becdetat: There's a link to my testing JSON file in the testing instructions

src/Rinds/RindsSpecsDataSourcePoller.php
```diff
@@ -0,0 +1,75 @@
...
+
+	/**
+	 * Polls the data sources for specs, persists those specs, and delete any
+	 * specs that no longer exist in the data sources.
+	 */
+	public static function poll_data_sources() {
+		$specs = array();
+
+		// Note that this merges the specs from the data sources based on the slug - last one wins.
+		foreach ( self::DATA_SOURCES as $url ) {
```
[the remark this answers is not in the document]
becdetat: As in open it up to other plugin developers to add their own feed? That sounds dangerous to me, is this something we do?

src/Rinds/RindsSpecsDataSourcePoller.php
becdetat: Nice - I think I copied that from somewhere else

src/Rinds/RindsSpecsDataSourcePoller.php
becdetat: It's used to build up the specs and persist them as an option.

src/Rinds/RindsSpecsDataSourcePoller.php
becdetat: For a first version I wanted to keep this simple - just a static JSON file stored on a server. Making a more complex request and requiring a more complex server-side implementation seems like it should be a future piece of work.

src/Rinds/RindsSpecsDataSourcePoller.php
becdetat: Good point, done.

src/Rinds/OrRuleProcessor.php
becdetat: Limited to two, I didn't really think about more than two operands as you could do that by nesting rules but I can see how that could get messy. The most flexible way would be an array of operands, is that what you're thinking?

becdetat: Ok, instead of `left_operand`/`right_operand` this now takes `operands` - an array of operands, each of which is evaluated by ANDing them together.

becdetat:
@joshuatf I've created a P2 [here](https://teamisotope.wordpress.com/2020/04/15/rule-based-inbox-notification-delivery/) to discuss what data to send to the server on the request

src/Rinds/RindsSpecRunner.php
```diff
@@ -0,0 +1,161 @@
...
+	/**
+	 * Run the spec.
+	 *
+	 * @param object $spec The spec to run.
+	 */
+	public static function run_spec( $spec ) {
+		$data_store = \WC_Data_Store::load( 'admin-note' );
+
+		// Create or update the note.
+		$existing_note_ids = $data_store->get_notes_with_name( $spec->slug );
```
timmyc, line 27: Should this ever return more than one note id? Seems like there should never be duplicates correct?

src/Rinds/RindsSpecsDataSourcePoller.php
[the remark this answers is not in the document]
timmyc: I agree making this filterable could lead to some nefarious activities.

src/Rinds/RindsSpecsDataSourcePoller.php
```diff
@@ -0,0 +1,72 @@
...
+class RindsSpecsDataSourcePoller {
+	const DATA_SOURCES = array(
+		'http://two.wordpress.test/rinds-specs.json',
+	);
+
+	/**
+	 * Polls the data sources for specs, persists those specs, and delete any
+	 * specs that no longer exist in the data sources.
+	 */
+	public static function poll_data_sources() {
```
timmyc, line 26: Since we still don't have a final URL for the source JSON data file, I kind of feel like we might want to make the polling of data sources configurable as a feature - so we can turn it off easily during the build flow.

src/Rinds/RindsSpecRunner.php
becdetat: I don't think there's actually anything that prevents duplicate note names. This is a pattern I've seen wherever we need to update a note.

src/Rinds/RindsSpecsDataSourcePoller.php
becdetat: Makes sense, done

becdetat:
@timmyc Thanks for the review :-) It's a worry that none of the notes became visible - the OR/plugins activated test case definitely should have passed.

> wanted to see if you think there would have been any way to break this up into smaller PRs

I don't think so - the plugin version rule could have been moved into a separate PR but the rest of the rules seem core and it doesn't make a lot of sense to have to review them in isolation.

> Per the WP coding standards we might want to consider either spelling out RINDS - maybe just RemoteInboxNotifications or if we prefer the acronym path like JITM - we should capitalize RINDS

I like the idea of spelling it out - I'll go with `RemoteInboxNotifications`.

> I would like to see a rule for respecting the opt in/out of Marketplace Suggestions

This makes sense to be an explicit rule rather than a special case. I'll add it to my backlog. Do we capture that opt in/out anywhere? I haven't seen it.

src/Rinds/RindsSpecRunner.php
joshuatf: I was thinking about this as well and may provide a small refactor while working on #4283.  I'm not sure if there's ever a use case for allowing multiple notes with the same name.

timmyc:
Thanks for the updates! Using https://gist.githubusercontent.com/becdetat/f17e9617e5ca1211d0579cf079862905/raw/99495c525c54663f6a0b90edff81e9ccd5b72afb/rinds-specs.json and following along with the instructions I have notes appearing now.

I still can't seem to get the Jetpack -or- WCS activated notice to appear. It stays stuck on `preunactioned` after I have activated Jetpack and WCS. I also tried running the cron job again and it stays in the `preunactioned` state.

src/Events.php
```diff
@@ -64,6 +65,7 @@ public function init() {
 	 */
 	public function do_wc_admin_daily() {
 		WC_Admin_Notes_New_Sales_Record::possibly_add_sales_record_note();
+		WC_Admin_Notes_Giving_Feedback_Notes::add_notes_for_admin_giving_feedback();
```
joshuatf, line 68: Side note: if we merge in we'll need to make sure this gets updated to `WC_Admin_Notes_Giving_Feedback_Notes::possibly_add_note();`

src/RemoteInboxNotifications/SpecRunner.php
```diff
@@ -0,0 +1,161 @@
...
+	 *
+	 * @param object $spec The spec to run.
+	 */
+	public static function run_spec( $spec ) {
+		$data_store = \WC_Data_Store::load( 'admin-note' );
+
+		// Create or update the note.
+		$existing_note_ids = $data_store->get_notes_with_name( $spec->slug );
+		if ( 0 === count( $existing_note_ids ) ) {
+			$note = new WC_Admin_Note();
```
joshuatf, line 29: It may be worthwhile to add in a new note class `WC_Admin_Rind_Note()` that makes use of note traits introduced in https://github.com/woocommerce/woocommerce-admin/pull/4340.

This would mean not needing to check against existing notes, notes of marketing type, and the evaluation rules could be added as a condition to the `possibly_add_note()` method.

If you have any other ideas or challenges with how you think this should be implemented, let me know and I can help revise note traits to better fit the notes here.

src/RemoteInboxNotifications/OrRuleProcessor.php
```diff
@@ -0,0 +1,45 @@
...
+
+	/**
+	 * Performs an OR operation on the rule's left and right operands.
+	 *
+	 * @param object $rule The specific rule being processed by this rule processor.
+	 *
+	 * @return bool The result of the operation.
+	 */
+	public function process( $rule ) {
+		foreach ( $rule->operands as $operand ) {
```
joshuatf, line 35: I received this error after running the cron event:

```
Undefined property: stdClass::$operands in /app/public/wp-content/plugins/woocommerce-admin/src/RemoteInboxNotifications/OrRuleProcessor.php on line 35
```

src/RemoteInboxNotifications/SpecRunner.php
```diff
@@ -0,0 +1,161 @@
...
+		$note->set_content_data( (object) array() );
+		$note->set_status( $status );
+		$note->set_type( $spec->type );
+		$note->set_icon( $spec->icon );
+		$note->set_name( $spec->slug );
+		$note->set_source( $spec->source );
+
+		// Clear then create actions.
+		$note->clear_actions();
+		foreach ( $spec->actions as $action ) {
```
joshuatf, line 61: I also received this error after running the cron event:

```
Undefined property: stdClass::$actions in /app/public/wp-content/plugins/woocommerce-admin/src/RemoteInboxNotifications/SpecRunner.php on line 61
```

mattsherman:
When testing I had this appear initially when following the instructions.

ScreenCapture at Wed May 20 14:08:42 EDT 2020

Also, I have been unable to get the `ad-for-automate-woo-2020-04-22` to appear. I thought that maybe it was due to the instruction:

> In the spec, set the ad-for-automate-woo-2020-04-22 spec's rule's publish_after value to some time in the future.

But, I then changed that, deleted all notes from the DB, and retriggered the poll by executing the cron `wc_admin_daily`. I still don't get it to appear. It is in the DB, with a `preunactioned` status. (not sure at all what `preunactioned` means!).

src/RemoteInboxNotifications/DataSourcePoller.php
```diff
@@ -0,0 +1,78 @@
...
+
+	/**
+	 * Read a single data source and insert the read specs into $specs, using
+	 * the spec slug as the key. So if the same or multiple data source has
+	 * a spec with the same slug, the last spec wins.
+	 *
+	 * @param string $url   The URL to read the specs from.
+	 * @param array  $specs The specs to read in to.
+	 */
+	private static function read_data_source( $url, &$specs ) {
```
mattsherman, line 55: The `read_data_source()` function has many concerns. It would be nice to split it up into separate functions focused on a single concern...

- `get_specs_from_data_source()`
- `merge_specs()`

src/RemoteInboxNotifications/DataSourcePoller.php
```diff
@@ -0,0 +1,78 @@
...
+defined( 'ABSPATH' ) || exit;
+
+use Automattic\WooCommerce\Admin\Loader;
+
+/**
+ * Specs data source poller class.
+ * This handles polling specs from JSON endpoints, and
+ * stores the specs in to the database as an option.
+ */
+class DataSourcePoller {
```
mattsherman, line 19: As this class only has static members, I think just namespacing the functions would be better.

src/RemoteInboxNotifications/DataSourcePoller.php
```diff
@@ -0,0 +1,78 @@
...
+	 */
+	private static function read_data_source( $url, &$specs ) {
+		$response = wp_remote_get( $url );
+
+		if ( is_wp_error( $response ) ) {
+			return;
+		}
+
+		$body = $response['body'];
+		$json = json_decode( $body );
```
mattsherman, line 63: The result of `json_decode` isn't JSON. So, a better variable name would be something like `$specs_from_data_source` (or just `$specs` if code is refactored to avoid this function having to concern itself with merging of specs (see above).

src/RemoteInboxNotifications/EvaluateAndGetStatus.php
```diff
@@ -0,0 +1,56 @@
...
+namespace Automattic\WooCommerce\Admin\RemoteInboxNotifications;
+
+defined( 'ABSPATH' ) || exit;
+
+use \Automattic\WooCommerce\Admin\Notes\WC_Admin_Note;
+
+/**
+ * Evaluates the spec and returns a status.
+ */
+class EvaluateAndGetStatus {
```
mattsherman, line 17: I don't think having this as a class really provides a benefit over a single function like `evaluate_spec_and_get_status( $spec, $current_status, $rule_evaluator )`.

src/RemoteInboxNotifications/GetRuleProcessor.php
```diff
@@ -0,0 +1,64 @@
...
+
+defined( 'ABSPATH' ) || exit;
+
+use \Automattic\WooCommerce\Admin\DateTimeProvider\CurrentDateTimeProvider;
+use \Automattic\WooCommerce\Admin\PluginsProvider\PluginsProvider;
+
+/**
+ * Class encapsulating getting the processor for a given rule type.
+ */
+class GetRuleProcessor {
```
mattsherman, line 18: Probably too big of a change at this point, but it could be interesting to consider a filter driven approach... so we could just do something like...

```
function wc_remote_inbox_process_plugins_activated_rule( $result, $rule ) {
    // process
}
add_filter( 'wc_remote_inbox_process_plugins_activated_rule', 'wc_remote_inbox_process_plugins_activated_rule', 10, 3 );

apply_filters( 'wc_remote_inbox_process_plugins_activated_rule', $result, $rule );
```

That would allow for additional extensibility and easy hooking into things for logging/metrics/testing/etc.

src/RemoteInboxNotifications/RemoteInboxNotificationsEngine.php
```diff
@@ -0,0 +1,59 @@
...
+
+use \Automattic\WooCommerce\Admin\Notes\WC_Admin_Note;
+use \Automattic\WooCommerce\Admin\PluginsProvider\PluginsProvider;
+
+/**
+ * RemoteInboxNotifications engine.
+ * This goes through the specs and runs (creates admin notes) for those
+ * specs that are able to be triggered.
+ */
+class RemoteInboxNotificationsEngine {
```
mattsherman, line 20: Similar to other comments... I don't see an advantage of a class with all statics over namespaced functions.

src/Notes/WC_Admin_Note.php
```diff
@@ -23,9 +23,10 @@ class WC_Admin_Note extends \WC_Data {
 	const E_WC_ADMIN_NOTE_INFORMATIONAL = 'info';    // used for presenting informational messages.
 
 	// Note status codes.
-	const E_WC_ADMIN_NOTE_UNACTIONED = 'unactioned'; // the note has not yet been actioned by a user.
-	const E_WC_ADMIN_NOTE_ACTIONED   = 'actioned';   // the note has had its action completed by a user.
-	const E_WC_ADMIN_NOTE_SNOOZED    = 'snoozed';    // the note has been snoozed by a user.
+	const E_WC_ADMIN_NOTE_PREUNACTIONED = 'preunactioned'; // the note is not actioned but shouldn't be displayed.
```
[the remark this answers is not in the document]
becdetat: Done in 6b971411

src/RemoteInboxNotifications/DataSourcePoller.php
```diff
@@ -0,0 +1,78 @@
...
+	const DATA_SOURCES = array(
+		'http://two.wordpress.test/rinds-specs.json',
+	);
+
+	/**
+	 * Polls the data sources for specs, persists those specs, and delete any
+	 * specs that no longer exist in the data sources.
+	 */
+	public static function poll_data_sources() {
+		if ( ! Loader::is_feature_enabled( 'remote-inbox-notifications' ) ) {
```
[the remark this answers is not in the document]
becdetat: Done in 2715ed0e

src/Events.php
becdetat: I'm not actually sure how this got here, it's duplicated from below. I've removed this line.

src/RemoteInboxNotifications/SpecRunner.php
becdetat: This is happening because one of the specs is missing an `actions` property. I don't see how, the current JSON file (https://gist.githubusercontent.com/becdetat/f17e9617e5ca1211d0579cf079862905/raw/99495c525c54663f6a0b90edff81e9ccd5b72afb/rinds-specs.json), which hasn't changed for a while, has actions for all of the specs and I can't reproduce this.

Try deleting the `wc_remote_inbox_notifications_specs` option, making sure you're using the JSON from above, and run the cron job again.

src/RemoteInboxNotifications/OrRuleProcessor.php
becdetat: It sounds like you're using an older version of the JSON file. Download the latest from https://gist.githubusercontent.com/becdetat/f17e9617e5ca1211d0579cf079862905/raw/99495c525c54663f6a0b90edff81e9ccd5b72afb/rinds-specs.json which should have the operands in the OR rule.

becdetat:
@mattsherman 

> Also, I have been unable to get the ad-for-automate-woo-2020-04-22 to appear

Strange, it works for me every time. Can you make sure that the spec's defined window matches your environment time? I had an issue using VVV where the vagrant box had a different system time to my system and reloading VVV fixed it.

becdetat:
@mattsherman 

> For future work, it would be good to try and consider an approach where you could get the skeleton in place, reviewed, and merged and then build on top of it (it that has already been done and I just missed that, due to not being closely involved in the dev of this, please excuse this suggestion!). That way, each review could be quicker and more focused.

Ok, I've taken this on board. I was treating all of the basic rules as a single piece of functionality so I considered this the framework for my other PRs (there are 4 more with a 5th on the way). But I can see that this was a problem with the size of this particular PR.

src/RemoteInboxNotifications/DataSourcePoller.php
```diff
@@ -0,0 +1,78 @@
...
+use Automattic\WooCommerce\Admin\Loader;
+
+/**
+ * Specs data source poller class.
+ * This handles polling specs from JSON endpoints, and
+ * stores the specs in to the database as an option.
+ */
+class DataSourcePoller {
+	const DATA_SOURCES = array(
+		'http://two.wordpress.test/rinds-specs.json',
```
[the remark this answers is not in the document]
becdetat: This was considered and we don't really want anybody to just be able to add a data source using a filter

src/RemoteInboxNotifications/DataSourcePoller.php
```diff
@@ -0,0 +1,78 @@
...
+	);
+
+	/**
+	 * Polls the data sources for specs, persists those specs, and delete any
+	 * specs that no longer exist in the data sources.
+	 */
+	public static function poll_data_sources() {
```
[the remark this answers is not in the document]
becdetat: I've called it `read_specs` because 1. it doesn't return anything so `get` isn't quite right and 2. it reads and persists all specs, not just looking for latest - specs can change.

src/RemoteInboxNotifications/DataSourcePoller.php
```diff
@@ -0,0 +1,78 @@
...
+		// Note that this merges the specs from the data sources based on the slug - last one wins.
+		foreach ( self::DATA_SOURCES as $url ) {
+			self::read_data_source( $url, $specs );
+		}
+
+		// Persist the specs as an option.
+		update_option( RemoteInboxNotificationsEngine::SPECS_OPTION_NAME, $specs );
+
+		// Run the engine.
+		RemoteInboxNotificationsEngine::run();
```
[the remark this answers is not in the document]
becdetat: It came to be in here because I expected to move it out into it's own cron task but that never happened. I've moved it into `Events.php` which is the only place the poller is run. Thanks :-)

src/RemoteInboxNotifications/DataSourcePoller.php
becdetat: I agree, I didn't know you could namespace functions in PHP. I've tried this and couldn't get it to work. The `DataSourcePoller.cs` file looks like:
```
namespace Automattic\WooCommerce\Admin\RemoteInboxNotifications;

function read_specs() {
    ...
}
```
and I'm using it like:
```
use function \Automattic\WooCommerce\Admin\RemoteInboxNotifications\read_specs;
...
read_specs();
```
and I'm getting this error:
```
[19-May-2020 14:57:25 UTC] PHP Fatal error:  Uncaught Error: Call to undefined function Automattic\WooCommerce\Admin\RemoteInboxNotifications\read_specs() in /srv/www/wordpress-two/public_html/wp-content/plugins/woocommerce-admin/src/Events.php:82
```

I can't find any usages of `use function` across wc-admin and wc-core so I can't see if I'm using the namespaces correctly. Do you know why this wouldn't work?

src/RemoteInboxNotifications/DataSourcePoller.php
becdetat: Done in d7ef2b1f

src/RemoteInboxNotifications/DataSourcePoller.php
becdetat: Nice :-) done in d7ef2b1f

src/RemoteInboxNotifications/EvaluateAndGetStatus.php
becdetat: Done in 6e616cc7

src/RemoteInboxNotifications/GetRuleProcessor.php
```diff
@@ -0,0 +1,64 @@
...
+	/**
+	 * Get the processor for the specified rule type.
+	 *
+	 * @param string $rule_type The rule type.
+	 *
+	 * @return object The matching processor for the specified rule type, or a FailRuleProcessor if no matching processor is found.
+	 */
+	public static function get_processor( $rule_type ) {
+		switch ( $rule_type ) {
```
peterfabian, line 27: Do you think it would make sense to have an interface or abstract class for the Rule Processors?

src/RemoteInboxNotifications/SpecRunner.php
```diff
@@ -0,0 +1,161 @@
...
+					return 'en_US' === $l->locale;
+				}
+			)
+		);
+
+		if ( 0 !== count( $en_us_locales ) ) {
+			return $en_us_locales[0];
+		}
+
+		throw new \Exception( __( 'Matching locale or fallback en_US locale not found. Make sure there is at least a en_US locale provided.', 'woocommerce-admin' ) );
```
peterfabian, line 159: Inconsistent use of `\Exception` vs `Exception` in the previous function.

peterfabian:
One thing I'm wondering about is if it would make sense to fetch only the locale relevant for the user? Maybe it's just unnecessary complication, but it seems a bit wasteful to me to send all translations for all messages to every shop?

jeffstieler:
Just noting here that when testing https://github.com/woocommerce/woocommerce-admin/pull/4334 I noticed that the note dates were set to the time the JSON file was fetched/parsed, not when the rules evaluated to true:

Screen Shot 2020-05-26 at 9 57 51 AM

src/RemoteInboxNotifications/GetRuleProcessor.php
```diff
@@ -0,0 +1,64 @@
...
+			case 'plugins_activated':
+				return new PluginsActivatedRuleProcessor(
+					new PluginsProvider()
```
jeffstieler, line 30: Assuming this injection technique is just to ease testing, what do you think about these becoming defaults in the rule processor constructors rather than having to pass them in here?

src/RemoteInboxNotifications/DataSourcePoller.php
```diff
@@ -0,0 +1,79 @@
...
+	 * @return array The specs that have been read from the data source.
+	 */
+	private static function read_data_source( $url ) {
+		$response = wp_remote_get( $url );
+
+		if ( is_wp_error( $response ) ) {
+			return [];
+		}
+
+		$body  = $response['body'];
```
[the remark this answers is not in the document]
becdetat: Wow that's interesting. I've added a check for the body, thanks.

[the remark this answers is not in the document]
jeffstieler: In case you're unaware, WP has some helper methods for this like [`wp_remote_retrieve_body`](https://developer.wordpress.org/reference/functions/wp_remote_retrieve_body/).

src/RemoteInboxNotifications/DataSourcePoller.php
mattsherman: Hmm... odd. That looks like it should work to me, but admittedly I do not consider myself an advanced PHP developer, so perhaps I'm overlooking a crucial detail in regards to namespacing functions.

mattsherman:
@becdetat Thanks for making the updates based on my previous feedback, and for explaining the reasoning for why things were/are implemented as they are!

src/RemoteInboxNotifications/DataSourcePoller.php
```diff
@@ -0,0 +1,79 @@
...
+
+	/**
+	 * Merge the specs.
+	 *
+	 * @param Array $specs_to_merge_in The specs to merge in to $specs.
+	 * @param Array $specs             The master list of specs.
+	 */
+	private static function merge_specs( $specs_to_merge_in, &$specs ) {
+		foreach ( $specs_to_merge_in as $spec ) {
+			$slug           = $spec->slug;
```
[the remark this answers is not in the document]
becdetat: I've added validation before storing invalid specs and some defensive code when processing them for non-required values. See de9b8d70. Thanks :-)

src/RemoteInboxNotifications/RemoteInboxNotificationsEngine.php
```diff
@@ -0,0 +1,59 @@
...
+	const SPECS_OPTION_NAME = 'wc_remote_inbox_notifications_specs';
+
+	/**
+	 * Initialize the engine.
+	 */
+	public static function init() {
+		add_action( 'activated_plugin', array( __CLASS__, 'run' ) );
+		add_action( 'deactivated_plugin', array( __CLASS__, 'run_on_deactivated_plugin' ), 10, 1 );
+	}
```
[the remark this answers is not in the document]
becdetat: Correct.

It wasn't an explicit requirement however the requirements were a bit vague on timing and it made sense for me to hook in to the activation/deactivation events and get the notifications appearing or disappearing reactively. Having the rules run only daily wasn't an original requirement either. I've done something similar in the support for the first product notification (#4260).

src/RemoteInboxNotifications/GetRuleProcessor.php
becdetat: Yes, this is something I would normally do. Not sure why I didn't do this originally. Done in 36488948.

src/RemoteInboxNotifications/SpecRunner.php
becdetat: True, this returns null for use in validation now anyway.

becdetat:
@peterfabian 

> One thing I'm wondering about is if it would make sense to fetch only the locale relevant for the user?

That does makes sense but it would require some logic on the data source's end, whereas now they can just provide a static JSON file. Plus the JSON is only loaded once a day. I think loading everything is ok for MVP.

becdetat:
@jeffstieler 

> Just noting here that when testing #4334 I noticed that the note dates were set to the time the JSON file was fetched/parsed, not when the rules evaluated to true 

Good catch! Fixed in 5d2a1eb4

image

src/RemoteInboxNotifications/GetRuleProcessor.php
becdetat: Done in d1fffcae

jeffstieler:
If this is ready for another review, please change the status to "needs review". 😄

src/RemoteInboxNotifications/SpecRunner.php
```diff
@@ -0,0 +1,191 @@
...
+			return;
+		}
+
+		// Set up the note.
+		$note->set_title( $locale->title );
+		$note->set_content( $locale->content );
+		$note->set_content_data( (object) array() );
+		$note->set_status( $status );
+		$note->set_type( $spec->type );
+		$note->set_icon( $spec->icon );
```
timmyc, line 65: With the new notes designs being merged in, `set_icon` has been deprecated so we should remove this to not get warnings in the woo log files on sites.

becdetat: Done in a68f2a4b

becdetat:
Thanks @timmyc , and thanks to all the reviewers too.

PR #4144, opened 2020-04-17
Options: Add data store

psealock:
Adds an options data store to replace `wc-api/options`.

### Testing Instructions

1. Smoke test the OBW and Task List.
2. Check for no differences in behaviour.
3. The "Customize your store" task list and payment steps included the most significant refactors of logic.


client/dashboard/task-list/index.js
```diff
@@ -409,13 +413,19 @@ class TaskDashboard extends Component {
 }
 
 export default compose(
+	window.wcSettings.preloadOptions
```
becdetat, line 416: Could you do this with an `if` statement? It seems a little complicated for what it does.

packages/data/src/options/actions.js
```diff
@@ -0,0 +1,65 @@
...
+
+import { apiFetch } from '@wordpress/data-controls';
+
+/**
+ * Internal Dependencies
+ */
+import TYPES from './action-types';
+import { WC_ADMIN_NAMESPACE } from '../constants';
+
+export function recieveOptions( options ) {
```
becdetat, line 13: *receive

becdetat:
You've used 'recieved' instead of 'received'. Apart from that and the question about an `if` this looks good :-)

client/dashboard/task-list/index.js
psealock: I originally had it like this:

```js
withOptionsHydration( {
    ...( window.wcSettings.preloadOptions || {} ),
} )
```
It was nicer to read, but I'd rather not call the hydration function if the data isn't there because resolvers in the data store will cache the empty values, preventing future api requests from going through.

client/dashboard/task-list/index.js
```diff
@@ -409,13 +413,19 @@ class TaskDashboard extends Component {
...
+		? withOptionsHydration( {
+				...window.wcSettings.preloadOptions,
+		  } )
+		: () => ( {} ),
```
jeffstieler, line 420: This should be `identity` from lodash: https://github.com/woocommerce/woocommerce-admin/pull/4094/commits/a22d7493210b90117bfbd1230164913ace6dbfe1#diff-713853358fa08698c1bf0f8dfa42ca74R282

packages/data/src/options/actions.js
psealock: Done, thanks!

client/dashboard/task-list/index.js
psealock: Done, thanks!

psealock: I also moved this section to the bottom of the `compose` pile.

packages/data/src/options/selectors.js
```diff
@@ -0,0 +1,77 @@
...
+		return result;
+	}, {} );
+};
+
+/**
+ * Determine if options are being requested.
+ *
+ * @param {Object} state - Reducer state
+ */
+export const isGetOptionsRequesting = ( state ) => {
```
joshuatf, line 48: I think we can replace these additional selectors with those built in to wp data.  Outlined an issue with more detail [here](https://github.com/woocommerce/woocommerce-admin/issues/4384).

packages/data/src/options/action-types.js
```diff
@@ -0,0 +1,9 @@
+const TYPES = {
+	RECIEVE_OPTIONS: 'RECIEVE_OPTIONS',
```
joshuatf, line 2:
```suggestion
	RECEIVE_OPTIONS: 'RECEIVE_OPTIONS',
```

psealock: Fixed in https://github.com/woocommerce/woocommerce-admin/pull/4144/commits/d7061c170d97d703750f0cf53d7cacebc92b9dea

packages/data/src/options/selectors.js
psealock: Thats an excellent idea! The way this is setup is `getOptions` calls resolver `getOptionsWithRequest` if the option hasn't been preloaded, which leaves us with this:

```js
const options = getOptions( args );
const isRequesting = ( 'getOptionsWithRequest', [ args ] ) );
// What we really want is this:
// const isRequesting = ( 'getOptions', [ args ] ) );
```
More on why its that way in p90Yrv-1uN#comment-3842.

One option is to introduce `isOptionsResolving` to map resolver names. This is less than ideal because I'd rather hide implementation details from the consumer.

```js
export const isOptionsResolving = ( state, selectorName, args ) => {
	return select( STORE_NAME ).isResolving(
		selectorName === 'getOptions' ? 'getOptionsWithRequest' : selectorName,
		args
	);
};
```
Any other ideas?

joshuatf: >Another option would be to get rid of getOptions all together and just have any request for data always go through getOption to simplify the code.

I think I like this idea the best.  It's a simple and easy to understand API and fixes issues with caching or needing to call with the exact same arguments used to determine resolution.

I originally wrote `getOptions` with an array for the purpose of avoiding multiple trips to the server when multiple options were needed, but if we wanted to keep this from a performance perspective, maybe we could collect requests for options and then fetch at the end of the stack.

For example, assume option A is cached but B and C need to be fetched and `optionsToRequest` and `fetches` are available across calls.

```js
// in component
const a = getOption( 'a' );
const b = getOption( 'b' );
const c = getOption( 'c' );

// in resolvers, store a persisted array of items that need to be requested
optionsToRequest.push( optionName );

// Use a timeout of 1ms to push to the end of the stack
setTimeout( () => {
  // optionsToRequest should now be ['b', 'c']
  // Avoid fetching multiple times for a group of options.
  fetchPromise = fetches[ optionsToRequest.join(',') ]
  if ( fetchStarted ) {
     yield fetchPromise;
     return;
  }

  // fetch options in array optionsToRequest
  fetches[ optionsToRequest.join(',') ]  = yield apiFetch( {
    path: url,
    method: 'GET',
    data: optionsToRequest
   } );
  
}, 1 );
```

Do you think something like that is feasible?

psealock: Nice one @joshuatf !

I like the request collection technique. The thing that deterred me from this solution was it complicating the hydration part because I'll need to loop through all of the options individually. I'll look for solution to that one though.

timmyc:
@psealock looks like this might need a rebase ( and possibly a tough one at that ) - is there more work that needs to be done here, and can we coordinate it in such a way so you don't get hit with another hefty merge conflict?

psealock:
@timmyc or @joshuatf Can you give this a look?

psealock:
>I'm seeing one oddity on the Setup Checklist.

Yup, thanks. Thats an error which I fixed in https://github.com/woocommerce/woocommerce-admin/pull/4144/commits/0a7ce3d6fd96e84f99e68e399922362e624b2aa0.

I also rebased against master.

psealock:
Thanks for the review @joshuatf. I fixed the status to `success: true | false` in https://github.com/woocommerce/woocommerce-admin/pull/4144/commits/db3eed008fc53e53caeaebb8eb4c96fdc4e57d04

PR #4145, opened 2020-04-17
Add setting to see if WcPay is configured

### LINKED ISSUE #4122: WooCommerce Payments treated as active when navigating back to the site without completing the stripe KYC flow
**Describe the bug**
If you start the Stripe KYC flow, then using the back button to return to the payments task list, WooCommerce Payments is treated as active.

**To Reproduce**

1. Ensure your test site has a US address and Jetpack conected
2. From the payments task list screen, click "Setup" on the WooCommerce Payments card
3. Note you're directed to the KYC flow hosted on stripes server. 
4. Click the back button in the browser to return to the payments task list.
5. Note that WooCommerce payments is now treated as active.

**Screenshot**

Screen Shot 2020-04-14 at 9 10 29 AM

**Expected behavior**
The user should see the WooCommerce Payments card with the recommended styling, including the setup link to Stripe's KYC flow.

joshuatf:
Fixes #4122 

Adds a setting to check if WcPay is configured. /cc @allendav Would be great to get your eyes on this to make sure this is the correct way to check configuration.

### Screenshots

Screen Shot 2020-04-17 at 6 03 42 PM

1. Install, but don't configure WcPay and checkout this branch https://github.com/Automattic/woocommerce-payments/pull/629.
1. Navigate to the task list.
1. Make sure WcPay is still not yet configured and contains the "Set up" button.
1. Set up WcPay.
1. Navigate back to the task list after setup and note that WcPay is configured and now contains a toggle button instead of the previous style.

### DIFF
--- a/client/task-list/tasks/payments/methods.js
+++ b/client/task-list/tasks/payments/methods.js
@@ -40,9 +40,9 @@ export function getPaymentMethods( {
 } ) {
 	const settings = getSetting( 'onboarding', {
 		stripeSupportedCountries: [],
-		wcPayIsConfigured: false,
+		wcPayIsConnected: false,
 	} );
-	const { stripeSupportedCountries, wcPayIsConfigured } = settings;
+	const { stripeSupportedCountries, wcPayIsConnected } = settings;
 
 	const hasCbdIndustry =
 		some( profileItems.industry, {
@@ -112,8 +112,8 @@ export function getPaymentMethods( {
 							'on U.S. issued cards. ',
 						'woocommerce-admin'
 					) }
-					{ wcPayIsConfigured && wcPaySettingsLink }
-					{ ! wcPayIsConfigured && <p>{ tosPrompt }</p> }
+					{ wcPayIsConnected && wcPaySettingsLink }
+					{ ! wcPayIsConnected && <p>{ tosPrompt }</p> }
 					{ profileItems.setup_client && <p>{ wcPayDocPrompt }</p> }
 				</Fragment>
 			),
@@ -124,7 +124,7 @@ export function getPaymentMethods( {
 				isJetpackConnected,
 			plugins: [ 'woocommerce-payments' ],
 			container: <WCPay />,
-			isConfigured: wcPayIsConfigured,
+			isConfigured: wcPayIsConnected,
 			isEnabled:
 				options.woocommerce_woocommerce_payments_settings &&
 				options.woocommerce_woocommerce_payments_settings.enabled ===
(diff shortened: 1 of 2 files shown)


allendav:
Yes, that should work, but I'm not a fan of the multiple calls to WCPay internals - if we change something there we could blow you up. What if WCPay updated an option with the result of the most recent is_stripe_connected call that you could check instead?

joshuatf:
>but I'm not a fan of the multiple calls to WCPay internals

Same- I was looking for an alternative when writing this.

>What if WCPay updated an option with the result of the most recent is_stripe_connected call that you could check instead?

That would be perfect, @allendav!  Do you want me to open a PR in wcpay or is this something you want to add in?

becdetat:
This looks good up to the point where Stripe is set up, I can't proceed past there since I'm not a US citizen and don't have an SSN. The code looks good though.

joshuatf:
Small update to this PR- it's no longer using the filter per https://github.com/Automattic/woocommerce-payments/pull/629 but now uses the `is_connected()` method to check if a connection has been established.

Also renamed the variables and no longer used the word `configured` since the exposed method will not guarantee full configuration.  We'll need a follow-up to address pending or failed connections.

dechov:
@becdetat FYI if you'd like to go through with a test account, you can set a `define` – see [this doc](https://docs.woocommerce.com/document/payments/testing/dev-mode/#section-1).

@joshuatf Will you also be adding the indicator in https://github.com/Automattic/woocommerce-payments/pull/629#issuecomment-629532707 conditional on `needs_setup`, or is that outside the scope of this PR?

joshuatf:
>@becdetat FYI if you'd like to go through with a test account, you can set a define – see this doc.

❤️ 

>@joshuatf Will you also be adding the indicator in Automattic/woocommerce-payments#629 (comment) conditional on needs_setup, or is that outside the scope of this PR?

@dechov Will handle this in a follow-up.  Are we sure that any false values returned from `needs_setup()` will be due to pending verification if the site is already connected? Or do we need to expose another method to check specifically for `$account_status['errors']`?

dechov:
> Are we sure that any false values returned from needs_setup() will be due to pending verification if the site is already connected? Or do we need to expose another method to check specifically for $account_status['errors']?

Good question, as logically it could be due to an error. Taking a closer look, it seems that an error there would be a highly exceptional scenario, since the handling in `is_stripe_connected( false )` is very similar to that in `get_account_status_data`, except that the latter could also give an error if we introduce a bug (or make a backwards-incompatible change) on the server. In short, I don't think we need to worry about this case.

PR #4148, opened 2020-04-20
OBW: "activated" success message upon continuing with existing theme

### LINKED ISSUE #4109: OBW: "activated" success message upon continuing with existing theme
After taking the primary "Continue with my active theme" action on the OBW theme step:

…a message appears that it was "installed and activated on your site":

I would expect either no message or an acknowledgement that the theme was unchanged.

octaedro:
Fixes #4109

This PR solves a problem at the end of the profiler. There was a conditional that always was sending the user to the theme installation flow.

### Screenshots
**Test 1**
In the last step of the new onboarding, when the user presses `Continue with my active theme`, it should finish the process without reinstalling the installed theme.
Screen Capture on 2020-04-19 at 19-05-16

**Test 2**
When the user chooses a new theme, it should be installed and show a message confirming that the process was successful
Screen Capture on 2020-04-19 at 22-46-22

- Go to the last step of the new onboarding (URL: `/wp-admin/admin.php?page=wc-admin&step=theme`).

**Test 1**
- Press `Continue with my active theme`.
- It shouldn't reinstall the selected theme. 
- Verify that the message: `[theme name] was installed and activated on your site` is not shown.

**Test 2**
- Press `Choose` to select one of the available themes.
- It should install the new theme and show a message like: `[theme name] was installed and activated on your site`.

### Changelog Note:
Fixed the conditional that always was sending the user to the theme installation flow.

### DIFF
--- a/client/dashboard/profile-wizard/steps/theme/index.js
+++ b/client/dashboard/profile-wizard/steps/theme/index.js
@@ -92,7 +92,7 @@ class Theme extends Component {
 			location,
 		} );
 
-		if ( theme !== activeTheme && getPriceValue( price ) <= 0 ) {
+		if ( slug !== activeTheme && getPriceValue( price ) <= 0 ) {
 			this.installTheme( slug );
 		} else {
 			updateProfileItems( { theme: slug } );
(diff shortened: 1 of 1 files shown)


client/dashboard/profile-wizard/steps/theme/index.js
```diff
@@ -92,7 +92,7 @@ class Theme extends Component {
 			location,
 		} );
 
-		if ( theme !== activeTheme && getPriceValue( price ) <= 0 ) {
+		if ( slug !== activeTheme && getPriceValue( price ) <= 0 ) {
```
joshuatf, line 95: Nice fix here!  We may want to check the `is_installed` property of the theme as there may be already installed themes in the list besides just the active theme.

octaedro: Thank you @joshuatf, good idea! I added this check and the method call in the commit [ce90633](https://github.com/woocommerce/woocommerce-admin/pull/4148/commits/ce906338f47d1330f767756a50d7fa28a5972d37)
Could you take a look again at this PR?

PR #4091, opened 2020-04-08
Onboarding: Fix "Set store location" in Shipping and Tax tasks on Dashboard

mattsherman:
Fixes #4031

This PR allows the store location to be properly set in the "Set up shipping" and "Set up tax" tasks on the WooCommerce Dashboard.

This issue is encountered when you do not have your store address set in `/wp-admin/admin.php?page=wc-settings`. So, either do not use the “new and improved setup experience" after activating the WooCommerce plugin (pick “Not right now”), or delete your store address after completing the setup wizard.

### Detailed test instructions:

- Run branch. Make sure WooCommerce Admin plugin is activated.

- Go to `/wp-admin/edit.php?post_type=shop_order` to make sure the new onboarding experience is enabled. Skip the setup wizard.

- Add a physical product to your store (“Add your first product”) so that the "Set up shipping" task is available.

- Then, go through both “Set up shipping” and “Set up tax” tasks on the Dashboard.

ScreenCapture at Fri Apr 17 17:54:20 EDT 2020

- Make sure both tasks correctly step through all available steps.
    - Try with both Jetpack not installed, and Jetpack installed and connected to your store, as different steps are presented in those different scenarios.

### Changelog Note:

Fix: "Set store location" step now works in "Set up shipping" and "Set up tax" tasks on the WooCommerce Dashboard.

PR #4149, opened 2020-04-20
OBW: respect Store Location async

### LINKED ISSUE #4031: Clicking continue on set store location task list step triggers console error
**Describe the bug**

Working through the onboarding task list when I click the "Continue" button on the "Set Store Location Step" I get the following console errors on master.

Production Build:

> TypeError: i is not a function

Development Build:

> TypeError: updateSettings is not a function

**Expected behavior**
I continue to the next step.

**Screenshots**

Screen Shot 2020-03-31 at 8 18 02 pm

* OS: macOS
* Browser firefox dev edition
* Version 5.0b10
* WooCommerce 4.0.1

I had a look to see if this was a dupe but couldn't see anything - apologies if it is known.

psealock:
Fixes #4031

1. `<StoreLocation />` had the wrong props for updating locations in the shipping task.
2. Async methods in the shipping and tax tasks called actual async actions after the settings refactor. This caused duplicate logic to be called:

>Functions of the OBW were written using async/await pattern even though data actions weren't actually async functions. This was a known problem #2959. Async functions called by StoreLocation's onComplete called duplicate logic also found in its parent component's componentDidUpdate. But, since the async logic was called synchronously, one or the other wasn't actually firing.

>Settings data stores were refactored and did export async action creators. So that duplicate logic ended up being fired twice.

1. Go through both “Set up shipping” and “Set up tax” tasks on the Dashboard.
2. Delete the `woocommerce_store_address` option and go through each again.
3. Try a variety of jetpack connected and not installed on each.
4. Make sure no steps are skipped and the tasks operate as they should.

### DIFF
--- a/client/dashboard/task-list/tasks/tax.js
+++ b/client/dashboard/task-list/tasks/tax.js
@@ -266,7 +266,8 @@ class Tax extends Component {
 							} );
 							if ( this.shouldShowSuccessScreen() ) {
 								this.setState( { stepIndex: null } );
-							} else {
+								// Only complete step if another update hasn't already shown succes screen.
+							} else if ( this.state.stepIndex !== null ) {
 								this.completeStep();
 							}
 						} }
--- a/client/dashboard/task-list/tasks/tax.js
+++ b/client/dashboard/task-list/tasks/tax.js
@@ -48,6 +48,9 @@ class Tax extends Component {
 		this.configureTaxRates = this.configureTaxRates.bind( this );
 		this.updateAutomatedTax = this.updateAutomatedTax.bind( this );
 		this.setIsPending = this.setIsPending.bind( this );
+		this.shouldShowSuccessScreen = this.shouldShowSuccessScreen.bind(
+			this
+		);
 	}
 
 	componentDidMount() {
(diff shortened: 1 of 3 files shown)


client/dashboard/task-list/tasks/tax.js
```diff
@@ -58,42 +61,53 @@ class Tax extends Component {
 		this.setState( this.initialState );
 	}
 
-	componentDidUpdate( prevProps ) {
+	shouldShowSuccessScreen() {
```
joshuatf, line 64: Nice!  I like the addition of this method.

client/dashboard/task-list/tasks/tax.js
```diff
@@ -58,42 +61,53 @@ class Tax extends Component {
...
+		const { stepIndex } = this.state;
+		const currentStep = this.getSteps()[ stepIndex ];
+		const currentStepKey = currentStep && currentStep.key;
+
+		// If general settings have stopped requesting, check if we should show success screen.
+		if (
+			prevProps.isGeneralSettingsRequesting &&
+			! isGeneralSettingsRequesting
+		) {
+			if ( this.shouldShowSuccessScreen() ) {
```
joshuatf, line 105: Now that you've simplified the logic a bit, could we remove this block and setting the `step` to `null` in favor of just checking `this. shouldShowSuccessScreen()` in the render method?

client/dashboard/task-list/tasks/tax.js
```diff
@@ -58,42 +61,53 @@ class Tax extends Component {
...
+			isGeneralSettingsRequesting,
+		} = this.props;
+		const { woocommerce_calc_taxes: calcTaxes } = generalSettings;
+		const { stepIndex } = this.state;
+		const currentStep = this.getSteps()[ stepIndex ];
+		const currentStepKey = currentStep && currentStep.key;
+
+		// If general settings have stopped requesting, check if we should show success screen.
+		if (
+			prevProps.isGeneralSettingsRequesting &&
```
joshuatf, line 102: Does `updateAndPersistSettingsForGroup` return a promise now?  Is it possible to `await` in the respective update methods?  Optionally we could handle this in a follow-up if it does.

mattsherman:
From initial testing of this (haven't reviewed code yet)...

- The "Set store location" step in each task is not auto-skipped if the store location is already set. It shows the task with the current store location filled out.
- Clicking "Configure" button in "Configure tax rates" step results in brokenness... button stays in "busy" state and nothing happens. No errors in console.

mattsherman:
>     * Clicking "Configure" button in "Configure tax rates" step results in brokenness... button stays in "busy" state and nothing happens. No errors in console.

I think I found the issue... I'll test a fix and push a commit to this PR's branch if it works.

mattsherman:
>     * The "Set store location" step in each task is not auto-skipped if the store location is already set. It shows the task with the current store location filled out.

Working on a fix for this.

mattsherman:
> > ```
> > * The "Set store location" step in each task is not auto-skipped if the store location is already set. It shows the task with the current store location filled out.
> > ```
> 
> Working on a fix for this.

@psealock I can't think of a way to fix this without refactoring the "get next step" handling as I did in #4091. But, that was a relatively big refactor. Perhaps it is best to just live with this as a known issue for 1.1 and cycle back to proper fix of this after the release (ideally, we wouldn't be calling `completeStep()` in `componentDidUpdate()`)? Or, perhaps you can think of a quick fix for this?

client/dashboard/task-list/tasks/tax.js
psealock: It does, would address in follow up.

client/dashboard/task-list/tasks/tax.js
psealock: Great point. I'd rather address in a follow up as I'm hesitant to fiddle with this at this stage.

psealock:
>Perhaps it is best to just live with this as a known issue for 1.1 and cycle back to proper fix of this after the release

I think this is the best course of action. It will allow us time to fix this issue at its core: Moving as much logic as possible to an async nature and/or changing Settings data selectors to differentiate `isRequesting` in the getting context v. updating.

psealock:
I found a bug:

If you complete the setup and have Jetpack connected, when you go to the Tax task you will see this success screen:

Screen Shot 2020-04-21 at 11 57 13 AM

If you run `wp option delete woocommerce_store_address` and redo the Tax task after a refresh, you'll get a prompt to enter address. Once you do and click "continue", the success screen should pop up. Instead it goes back to Dashboard.

Working on a fix now.

psealock:
Fixed in https://github.com/woocommerce/woocommerce-admin/pull/4149/commits/45513cd215bcee220754ac7a3d8067028140e185

psealock:
Tested on Ephemeral site with Jetpack connected, unconnected, with store address, and without store address on shipping and tax tasks. Working well for me.

PR #4157, opened 2020-04-20
Adds the onboarding payments note

### LINKED ISSUE #3791: "Set up payments" reminder: inbox notification
This issue should be completed after #3710 it's done. We want to add a reminder to users that skip the "set up payments" task by selecting the option "I want to use a different payment option"

**Title:** Start accepting payments on your store!
**Copy:** Take payments with the provider that’s right for you - choose from 100+ payment gateways for WooCommerce.
**Store type:** New users that skip the "Set up" payments task list with the option "I want to use a different payment option" (event: `wcadmin_tasklist_payment_skip_task`, event prop: `skip_option` (value: tbd))
**Trigger:** 3 days after skipping the "Set up" payments task list with the option "I want to use a different option" (event: `wcadmin_tasklist_payment_skip_task`, event prop: `skip_option` (value: tbd))
**Cadence:** Once
**Action CTA leads to:** https://woocommerce.com/product-category/woocommerce-extensions/payment-gateways/
**CTA copy:** Learn more

joshuatf:
Fixes #3791 

If payments are skipped in the task list, a note is added 3 days after to reminder store owners to set up payments.

Note that this note is added if the following conditions are met:
* The user skipped the payments task more than 3 days ago.
* The user does not have any enabled payment gateways.

### Screenshots
Screen Shot 2020-04-20 at 6 53 04 PM

1. Delete `woocommerce_task_list_payments` in your options table if it exists.
1. Disable all payment methods.
1. Enable the task list and visit the payments task.
1. Choose the "skip" option.
1. Wait 3 days ... or update the timestamp in `woocommerce_task_list_payments` to speed things up.
1. Check that a note is added.
1.  Click on the "Learn more" button in the note and make sure the note is actioned and you are taken to the wc payment extensions page.

### DIFF
--- a/src/Notes/WC_Admin_Notes_Onboarding_Payments.php
+++ b/src/Notes/WC_Admin_Notes_Onboarding_Payments.php
@@ -1,0 +1,79 @@
+<?php
+/**
+ * WooCommerce Admin: Payments reminder note.
+ *
+ * Adds a notes to complete the payment methods.
+ *
+ * @package WooCommerce Admin
+ */
+
+namespace Automattic\WooCommerce\Admin\Notes;
+
+defined( 'ABSPATH' ) || exit;
+
+use \Automattic\WooCommerce\Admin\Features\Onboarding;
+
+/**
+ * WC_Admin_Notes_Onboarding_Payments.
+ */
+class WC_Admin_Notes_Onboarding_Payments {
+	const NOTE_NAME = 'wc-admin-onboarding-payments-reminder';
+
+	/**
+	 * Creates a note to remind store owners to set up payments.
+	 */
+	public static function possibly_add_onboarding_payments_note() {
+		// This note should only be added if the task list is still shown.
+		if ( ! Onboarding::should_show_tasks() ) {
+			return;
+		}
+
+		// Make sure payments task was skipped at least 3 days ago.
+		$three_days = 3 * DAY_IN_SECONDS;
+		( ( time() - $wc_admin_installed ) >= $seconds );
+		$payments_task = get_option( 'woocommerce_task_list_payments', array() );
+		if (
+			! isset( $payments_task['skipped'] ) ||
+			( time() - $payments_task['timestamp'] ) < $three_days
+		) {
+			return;
+		}
+
+		// Check to see if any gateways have been added.
+		$gateways         = WC()->payment_gateways->get_available_payment_gateways();
+		$enabled_gateways = array_filter(
+			$gateways,
+			function( $gateway ) {
+				return 'yes' === $gateway->enabled;
+			}
+		);
+		if ( ! empty( $enabled_gateways ) ) {
+			return;
+		}
+
+		// Don't add this note if previously added.
+		$data_store = \WC_Data_Store::load( 'admin-note' );
+		$note_ids   = $data_store->get_notes_with_name( self::NOTE_NAME );
+		if ( ! empty( $note_ids ) ) {
+			return;
+		}
+
+		$note = new WC_Admin_Note();
+		$note->set_title( __( 'Start accepting payments on your store!', 'woocommerce-admin' ) );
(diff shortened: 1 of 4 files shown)


src/Notes/WC_Admin_Notes_Onboarding_Payments.php
```diff
@@ -0,0 +1,79 @@
...
+	 */
+	public static function possibly_add_onboarding_payments_note() {
+		// This note should only be added if the task list is still shown.
+		if ( ! Onboarding::should_show_tasks() ) {
+			return;
+		}
+
+		// Make sure payments task was skipped at least 3 days ago.
+		$three_days = 3 * DAY_IN_SECONDS;
+		( ( time() - $wc_admin_installed ) >= $seconds );
```
rrennick, line 33: @joshuatf Should this be `$three_days` instead of `$seconds` ?

joshuatf: Thanks for catching this, @rrennick.  I goofed by leaving this line in.  Removed in https://github.com/woocommerce/woocommerce-admin/pull/4157/commits/4f0f7cdede1bdc1eb6db725f68781880a97247e3 and also renamed to `$three_days_in_seconds` for clarity.

rrennick:
@joshuatf I'm getting a REST API error on this branch

```
Undefined index: items in /Users/ronrennick/Sites/ron/wp-includes/rest-api.php on line 1245
```

I updated `woocommerce_task_list_payments` to `{"skipped":"1","timestamp":"1587700000"}` (should be 4-5 days ago).

src/Notes/WC_Admin_Notes_Onboarding_Payments.php
```diff
@@ -0,0 +1,78 @@
...
+		if ( ! Onboarding::should_show_tasks() ) {
+			return;
+		}
+
+		// Make sure payments task was skipped at least 3 days ago.
+		$three_days_in_seconds = 3 * DAY_IN_SECONDS;
+		$payments_task         = get_option( 'woocommerce_task_list_payments', array() );
+		if (
+			! isset( $payments_task['skipped'] ) ||
+			( time() - $payments_task['timestamp'] ) < $three_days_in_seconds
```
becdetat, line 36: Is the logic correct here? 

age = current time - skipped time

age is _less_ than three days

Should that be _more_ than three days?

becdetat: if I run this without skipping the Payments task (and just ignore it) the `woocommerce_task_list_payments` option doesn't get created and I get this:

```
PHP Notice:  Undefined index: timestamp in /srv/www/wordpress-two/public_html/wp-content/plugins/woocommerce-admin/src/Notes/WC_Admin_Notes_Onboarding_Payments.php on line 40
```

joshuatf: >Should that be more than three days?

I believe the current logic is correct.  If it's less than 3 days, we return and don't add the note.  If it's more than 3 days we want to add it, so we don't return.

>if I run this without skipping the Payments task (and just ignore it) the woocommerce_task_list_payments option doesn't get created and I get this:

Hmm, that's an interesting one.  I think the only way you would hit `Undefined index: timestamp` is if you had already skipped and the  `skipped` index exists; otherwise it should never get evaluated.

It may be that you skipped prior to this PR, so there is `skipped` but no `timestamp`.  I added a check to see if this index is set and return early if not.  Can you let me know if that fixes the error for you @becdetat?

joshuatf:
>I'm getting a REST API error on this branch
Undefined index: items in /Users/ronrennick/Sites/ron/wp-includes/rest-api.php on line 1245

@rrennick I'm also getting that error on my local master so I think it may be unrelated to this PR.

>I updated woocommerce_task_list_payments to {"skipped":"1","timestamp":"1587700000"} (should be 4-5 days ago).

Did this work for you?  I tested with this info serialized (and the timestamp as an `int`) and it worked as expected.

PR #4166, opened 2020-04-21
Reset profiler when visiting old obw URL

### LINKED ISSUE #3695: [4.0 beta1] New OBW: Can't use `wp-admin/admin.php?page=wc-setup` to later go back to the OBW
**Describe the bug**

Testing 4.0 beta1 pre-public release. New OBW: Can't use `wp-admin/admin.php?page=wc-setup` to later go back to the OBW. Not sure if this is intentional so reporting. 

**To Reproduce**

1. Go through new OBW;
2. After completing it, try to append `wp-admin/admin.php?page=wc-setup` to go back to the beginning of OBW;
3. You don’t get to the beginning of OBW where you can select if you want to proceed with older or new OBW. Instead, you get redirected to `wp-admin/admin.php?page=wc-admin ` and task list: 

Screen Shot 2020-02-10 at 10 17 10 AM

**Expected behavior**

I expected to be able to go back to the beginning of OBW where I could select if proceed with new or older OBW. Especially in the case like in this issue where I enabled new OBW via snippet as I was not randomly selected for it. 

I also tried to disable the snippet  (on the same existing site) but was still getting the same result as above. 

**Screenshots**

See above. 

**WordPress (please complete the following information):**

* Core version: 5.3.2
* Gutenberg version (if installed): not installed
* WooCommerce version: 4.0 beta1
* WooCommerce Blocks version: not installed
* Site language: English
* Any other plugins installed: Jetpack by WordPress.com

* OS: macOS Catalina v 10.15.3
* Browser: Chrome
* Version: 80.0.3987.87 (Official Build) (64-bit)

**Additional context**

Beta 4.0 was not released to the public yet - this is found during testing the package we tagged before tagging beta and releasing it to the public.

joshuatf:
Fixes #3695 

Resets the new onboarding profiler when visitng the old OBW path (i.e., `wp-admin/admin.php?page=wc-setup`).

1. Complete the profiler so that the task list is shown.
1. Visit `wp-admin/admin.php?page=wc-setup`.
1. Make sure the profiler is reset.

### DIFF
--- a/src/Features/Onboarding.php
+++ b/src/Features/Onboarding.php
@@ -149,7 +149,7 @@ class Onboarding {
 
 		if ( 'wc-setup' === $current_page ) {
 			delete_transient( '_wc_activation_redirect' );
-			wp_safe_redirect( wc_admin_url() );
+			wp_safe_redirect( wc_admin_url( '&reset_profiler=1' ) );
 		}
 	}
 


joshuatf:
>I wanted to confirm that what I'm seeing is ok.

Yup- that's the correct behavior @octaedro (reset may not have been the best word in retrospect but we've had this parameter in since onboarding began development and was meant to reset).

Currently this just "uncompletes" the profiler so we can view it again.  Thanks for checking!

PR #4179, opened 2020-04-22
Add scrollable styling to left side of Table, and keep updated

dechov:
https://github.com/woocommerce/woocommerce-admin/commit/2b21dba372330c60bca9ae9f2fb5fac15da46ba0 adds to `<Table>` a visible indication of being scrollable to the left, after table has been scrolled to the right – by analogy with the existing right-side indicator.

https://github.com/woocommerce/woocommerce-admin/commit/ac3203954f2408a9e3b75316c9e13b914efc3b0a makes sure that the indicators adapt to changes in column visibility and other updates that could impact the width of the table rows.

Note that latter change would cause an infinite render loop if `setState` is not called conditionally – open to ideas for avoiding this in a more robust way if possible.

### Accessibility

(Copied interaction from existing, so not anticipating any _new_ accessibility concerns.)

### Screenshots

Starting with a table without horizontal scrolling: 

<table>
<tr><th></th><th><code>master</code></th><th>this branch</th></tr>
<tr><td>Toggle columns on</td><td></td><td></td></tr>
<tr><td>Scrolled right</td><td></td><td></td></tr>
</table>

- Navigate to a screen with a table
- Set few enough columns visible to avoid horizontal scrolling
- Add columns to necessitate horizontal scrolling
- **Verify that a shadow appears on the right side**
- Scroll to the right and **verify that a shadow appears on the left side**

### DIFF
--- a/packages/components/src/table/table.js
+++ b/packages/components/src/table/table.js
@@ -54,8 +54,8 @@ class Table extends Component {
 		super( props );
 		this.state = {
 			tabIndex: null,
-			isScrollable: false,
-			isScrollableBack: false,
+			isScrollableRight: false,
+			isScrollableLeft: false,
 		};
 		this.container = createRef();
 		this.sortBy = this.sortBy.bind( this );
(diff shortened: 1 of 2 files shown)


packages/components/src/table/table.js
```diff
@@ -100,11 +105,21 @@ class Table extends Component {
 
 	updateTableShadow() {
 		const table = this.container.current;
+
 		const scrolledToEnd =
 			table.scrollWidth - table.scrollLeft <= table.offsetWidth;
-		this.setState( {
-			isScrollable: ! scrolledToEnd,
-		} );
+		if ( scrolledToEnd && this.state.isScrollable ) {
```
joshuatf, line 111: I think these checks are a fine way to handle things.  👍 Can we also destructure these state properties?

packages/components/src/table/table.js
```diff
@@ -118,9 +133,10 @@ class Table extends Component {
 			rowHeader,
 			rows,
 		} = this.props;
-		const { isScrollable, tabIndex } = this.state;
+		const { isScrollable, isScrollableBack, tabIndex } = this.state;
 		const classes = classnames( 'woocommerce-table__table', classNames, {
 			'is-scrollable': isScrollable,
+			'is-scrollable-back': isScrollableBack,
```
joshuatf, line 139: Nitpicky, but this might be easier to understand if we change these to `isScrollableLeft` and `isScrollableRight`.  What do you think?

dechov: Definitely agreed – done in https://github.com/woocommerce/woocommerce-admin/pull/4179/commits/6b8340e370668309dcbfccf393f28b2096730f37

packages/components/src/table/table.js
dechov: 👍 Done in https://github.com/woocommerce/woocommerce-admin/pull/4179/commits/6b8340e370668309dcbfccf393f28b2096730f37

jameskoster:
I was actually speaking with .org designers about elevation (shadows etc) in the WordPress UI last week. Consensus was actually to avoid it for anything other than popover menus and dialogs. So we should probably try to embrace that in WooCommerce moving forwards.

A persistent scrollbar is probably the simplest approach to indicate scrollability here, although that undoes your hard work. Sorry :(

jameskoster:
ok, as [discussed in Slack](https://a8c.slack.com/archives/C9K5T0XSP/p1589898568493500), I think we can keep this treatment, but tweak the colors so that the "shadows" fade from/to white instead.

fade

dechov:
How does this look?

white-scrollable-indicators

Made a quick attempt to have it show up per row so that the border wasn't faded and the color could match the row background, but couldn't get the height right.

jameskoster:
LGTM :shipit:

PR #4184, opened 2020-04-23
Change "Demo" to "Sample" products

jobthomas:
A few years ago we had a discussion on the name of the files for test products and we choose "sample" data over "dummy" data. I think we should likely stick to that naming (which is also the name of the folder and the files in WC core: https://github.com/woocommerce/woocommerce/tree/master/sample-data)

**First commit changes that wording**

**Second commit:** All titles and questions in the new onboarding wizards only capitalise names and the first letter of the sentence. This seemed a tiny bit off. 

Suggestion after reviewing new onboarding doc by @pmcpinto


pmcpinto:
Looks good to me.

PR #4189, opened 2020-04-23
New home screen pageview tracking

### LINKED ISSUE #4083: New home screen pageview tracking
We'll need to start tracking the number of page views on the new home screen. We also need to track is the user has Jetpack installed and/or connected, so I think that we have 2 options:

**Option 1:** 
- Adding a new value in the `path` event prop of the `wcadmin_page_view`. That value should have the following name: home_screen.
- Add three additional even props for Jetpack, same as we have in `wcadmin_task_view`: jetpack_active, jetpack_connected, jetpack_installed
I imagine that this probably would require us to check the Jetpack props in every wcadmin screen, hurting performance.

**Option 2:**
As an alternative, we can add a new event just for the home screen page views and checking Jetpack in this screen:
Event name: wcadmin_home_page_view
Event props:
jetpack_active (true/false)
jetpack_connected (true/false)
jetpack_installed (true/false)

jeffstieler:
Fixes #4083.

This PR seeks to implement "Option 1" of #4083.

It changes `wcadmin_page_view`'s `path` Tracks event prop to `home_screen` and adds the Jetpack info.

- Use `npm start` to ensure a development build with the new home screen feature enabled
- Enable tracks debugging by entering `localStorage.setItem( 'debug', 'wc-admin:*' )` in console
- Go to WooCommerce > Home
- Verify the tracks event is correct

### DIFF
--- a/client/layout/index.js
+++ b/client/layout/index.js
@@ -3,6 +3,7 @@
  */
 import { Component } from '@wordpress/element';
 import { useFilters } from '@woocommerce/components';
+import { compose } from '@wordpress/compose';
 import { Router, Route, Switch } from 'react-router-dom';
 import PropTypes from 'prop-types';
 import { get, isFunction } from 'lodash';
@@ -12,6 +13,7 @@ import { get, isFunction } from 'lodash';
  */
 import { getHistory } from '@woocommerce/navigation';
 import { getSetting } from '@woocommerce/wc-admin-settings';
+import { PLUGINS_STORE_NAME, withPluginsHydration } from '@woocommerce/data';
 
 /**
  * Internal dependencies
@@ -24,6 +26,7 @@ import { recordPageView } from 'lib/tracks';
 import TransientNotices from './transient-notices';
 import StoreAlerts from './store-alerts';
 import { REPORTS_FILTER } from 'analytics/report';
+import withSelect from 'wc-api/with-select';
 
 export class PrimaryLayout extends Component {
 	render() {
@@ -41,7 +44,7 @@ export class PrimaryLayout extends Component {
 	}
 }
 
-class Layout extends Component {
+class _Layout extends Component {
 	componentDidMount() {
 		this.recordPageViewTrack();
 		document.body.classList.remove( 'woocommerce-admin-is-loading' );
@@ -61,7 +64,13 @@ class Layout extends Component {
 	}
 
 	recordPageViewTrack() {
-		const { isEmbedded } = this.props;
+		const {
+			activePlugins,
+			installedPlugins,
+			isEmbedded,
+			isJetpackConnected,
+		} = this.props;
+
 		if ( isEmbedded ) {
 			const path = document.location.pathname + document.location.search;
 			recordPageView( path, { isEmbedded } );
@@ -83,7 +92,11 @@ class Layout extends Component {
 				: 'dashboard';
 		}
 
-		recordPageView( path );
+		recordPageView( path, {
+			jetpack_installed: installedPlugins.includes( 'jetpack' ),
+			jetpack_active: activePlugins.includes( 'jetpack' ),
+			jetpack_connected: isJetpackConnected,
+		} );
 	}
 
 	render() {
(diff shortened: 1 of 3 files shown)


client/layout/index.js
```diff
@@ -129,6 +144,32 @@ Layout.propTypes = {
...
+			isJetpackConnected: isJetpackConnected(),
+			installedPlugins: getInstalledPlugins(),
+		};
+	} )
+)( _Layout );
+
+const HydratedLayout = withPluginsHydration( {
+	...window.wcSettings.plugins,
+	jetpackStatus: window.wcSettings.dataEndpoints.jetpackStatus,
+} )( Layout );
```
psealock, line 171: You can add these two functions at the same time in `compose`.

```js
const Layout = compose(
    withSelect( /* .... */ ),
    withPluginsHydration(  /* .... */ )
)( _Layout );
```
Also, in the [next wp.data refactor](https://github.com/woocommerce/woocommerce-admin/pull/4144/files#diff-f5d6c918125bcd6501cbf9f5309c544c), I'm adopting a safer version of this in case some properties are not defined.

```js
withPluginsHydration( {
	...( window.wcSettings.plugins || {} ),
	jetpackStatus:
		window.wcSettings.dataEndpoints &&
		window.wcSettings.dataEndpoints.jetpackStatus,
} ),
```

client/layout/index.js
```diff
@@ -24,6 +26,7 @@ import { recordPageView } from 'lib/tracks';
 import TransientNotices from './transient-notices';
 import StoreAlerts from './store-alerts';
 import { REPORTS_FILTER } from 'analytics/report';
+import withSelect from 'wc-api/with-select';
```
psealock, line 29: This `withSelect` can be imported directly from `@wordpress/data` since wc-api is not used. This will work, and its still used in some places when Fresh Data selectors are also used.

client/layout/index.js
```diff
@@ -12,6 +13,7 @@ import { get, isFunction } from 'lodash';
  */
 import { getHistory } from '@woocommerce/navigation';
 import { getSetting } from '@woocommerce/wc-admin-settings';
+import { PLUGINS_STORE_NAME, withPluginsHydration } from '@woocommerce/data';
```
psealock, line 16: If we are going to be using `withPluginsHydration` at the `Layout` level, we can remove the other instances of `withPluginsHydration` which are used in the OBW, Task List, and Shipping Banner. It only needs to be used once.

client/layout/index.js
```diff
@@ -78,10 +87,16 @@ class Layout extends Component {
 
 		// When pathname is `/` we are on the dashboard
 		if ( path.length === 0 ) {
-			path = 'dashboard';
+			path = window.wcAdminFeatures.homepage
+				? 'home_screen'
+				: 'dashboard';
```
psealock, line 92: :+1:

psealock (commented):
Looking good @jeffstieler. I left some comments about hydrating the plugins data.

client/layout/index.js
jeffstieler: How about this, so `jetpackStatus` will be a boolean?

```javascript
withPluginsHydration( {
	...( window.wcSettings.plugins || {} ),
	jetpackStatus:
		window.wcSettings.dataEndpoints &&
		window.wcSettings.dataEndpoints.jetpackStatus || false,
} ),
```

client/layout/index.js
jeffstieler: Are you sure the shipping banner can have the hydration removed? I don't think it's a child of `<Layout />` as it has it's own call to `render()`: https://github.com/woocommerce/woocommerce-admin/blob/345e39b8e84020c1b2bcfba9aa30d8c10e61b732/client/wp-admin-scripts/print-shipping-label-banner/index.js#L27-L30

psealock: Right, good call

psealock (approved):
LGTM, testing well. Thanks @jeffstieler

PR #4193, opened 2020-04-24
POC: how slotfill would work for Homepage

psealock:
This is a POC that is not meant to be merged. I made it to solidify my own understanding of slotFill and it turns out to be a good starting point for some questions around how we'll use it.

This is a distillation of https://github.com/woocommerce/woocommerce-admin/pull/3203 into its most basic elements. 

### Questions

1. Are the patterns proposed in https://github.com/woocommerce/woocommerce-admin/pull/3203 how we envision for the homepage and potentially the rest of the app? What were the drawbacks leading to that PR not getting merged?
2. Instead of exporting content components, ie `<HomepageContent />`, to `@woocommerce/components`, what about a new package such as `@woocommerce/slotfills`?
3. We spoke previously about using the slotFill approach ourselves. I think this is a great idea, but it still requires rendering content anywhere within `<SlotFillProvider />`. Is it worth the added misdirection instead of rendering directly in the `<Homepage />` component's output? 
4. What is the scope of `<SlotFillProvider />`? Should it encompass the entire app? Or limit it to the Homepage?

cc @jeffstieler

### DIFF
--- a/client/homepage/index.js
+++ b/client/homepage/index.js
@@ -26,7 +26,14 @@ const MyPlugin = () => {
 		</HomepageContent>
 	);
 };
-registerPlugin( 'my-plugin', { render: MyPlugin } );
+registerPlugin(
+	'my-plugin-homepage',
+	{ render: MyPlugin },
+	'wc-admin-homepage'
+);
+// Register an unrelated plugin which should not be rendered.
+// Depends on https://github.com/psealock/gutenberg/tree/update/plugins-area-scoping
+registerPlugin( 'my-plugin-unrelated', { render: MyPlugin }, 'something-else' );
 
 // Homepage.js
 const Homepage = () => {
@@ -37,7 +44,7 @@ const Homepage = () => {
 					<span>I am INTERNAL content body</span>
 				</HomepageContent>
 				<HomepageContent.Slot />
-				<PluginArea />
+				<PluginArea scope="wc-admin-homepage" />
 			</SlotFillProvider>
 		</div>
 	);
--- a/client/homepage/index.js
+++ b/client/homepage/index.js
@@ -1,5 +1,46 @@
+import {
+	SlotFillProvider,
+	Panel,
+	PanelBody,
+	createSlotFill,
+} from '@wordpress/components';
+import { registerPlugin, PluginArea } from '@wordpress/plugins';
+
+// exported component available to plugins maybe '@woocommerce/slotfills`
+const { Fill, Slot } = createSlotFill( 'MyPanelSlot' );
+const HomepageContent = ( { children } ) => (
+	<Fill>
+		<Panel header="Content wrapper">
+			<PanelBody>{ children }</PanelBody>
+		</Panel>
+	</Fill>
+);
+HomepageContent.Slot = () => <Slot />;
+
+// my-plugion.js
+// import { HomepageContent } from '@woocommerce/slotfills`;
+const MyPlugin = () => {
+	return (
+		<HomepageContent>
+			<span>I am EXTERNAL content body</span>
+		</HomepageContent>
+	);
+};
+registerPlugin( 'my-plugin', { render: MyPlugin } );
+
+// Homepage.js
 const Homepage = () => {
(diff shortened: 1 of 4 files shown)


jeffstieler:
> 1. Are the patterns proposed in https://github.com/woocommerce/woocommerce-admin/pull/3203 how we envision for the homepage and potentially the rest of the app?

That was certainly _my_ proposal - didn't get a ton of feedback on it (either on the PR or on P2 - p90Yrv-1lH).

> What were the drawbacks leading to that PR not getting merged?

I think the biggest hurdle was that not everything was working 100% - the animation of opening panels broke.

Additionally, there was extra work to do regarding `<PluginArea />` - at the time of writing (and probably now still) it couldn't be namespaced or scoped - so rendering that component would render any Gutenberg plugins whose scripts might be enqueued on the page. They'd likely be inert, but they'd be included.

> 2. Instead of exporting content components, ie `<HomepageContent />`, to `@woocommerce/components`, what about a new package such as `@woocommerce/slotfills`?

We should probably follow Gutenberg's lead here and [use faux packages internally](https://github.com/WordPress/gutenberg/blob/d2e0ecd1c2e14a486b28fc90154a7dff8906238b/docs/designers-developers/developers/slotfills/plugin-post-publish-panel.md), but instruct developers to [use global objects for extending with SlotFills](https://developer.wordpress.org/block-editor/developers/slotfills/).

> 3. We spoke previously about using the slotFill approach ourselves. I think this is a great idea, but it still requires rendering content anywhere within `<SlotFillProvider />`. Is it worth the added misdirection instead of rendering directly in the `<Homepage />` component's output? 

We don't have to use the SlotFills ourselves, but there is opportunity to standardize behavior and reuse code through those components so we will likely want to use them unless we run into difficulty in some edge cases. For example, the SlotFill created in https://github.com/woocommerce/woocommerce-admin/pull/3203 handled all treatment of the "current tab" and rendering the corresponding panel, etc. 

> 4. What is the scope of `<SlotFillProvider />`? Should it encompass the entire app? Or limit it to the Homepage?

I'm not 100% certain. It should be safe to have it encompass the entire app (as the [documentation says](https://github.com/WordPress/gutenberg/tree/e8f4f37d5c4e45ce545ddf55e387100ae996a15e/packages/components/src/slot-fill#usage) to render it "at the root of your application"), but Gutenberg uses it in slightly tighter scopes: https://github.com/WordPress/gutenberg/search?p=2&q=SlotFillProvider&unscoped_q=SlotFillProvider

psealock:
Thanks for the feedback @jeffstieler. The links you've sent put me on a path to reviewing how Gutenberg handles things, and I believe its a pattern we can follow, which this POC and #3203 do.

The only hesitation I have at this point is:

>Additionally, there was extra work to do regarding `<PluginArea />` - at the time of writing (and probably now still) it couldn't be namespaced or scoped - so rendering that component would render any Gutenberg plugins whose scripts might be enqueued on the page. They'd likely be inert, but they'd be included.

I think "they'd likely be inert" is right, but it would be nice to not render unrequired code. It seems to me the PluginArea was designed to house Core Editor slotfill uses by extensions, and so wasn't scoped. But with full site editing coming, this would be functionality would be useful in that context as well. We can offer a PR to allow scoping in Gutenberg and continue our efforts here in the meantime.

psealock:
>We should probably follow Gutenberg's lead here and use faux packages internally, but instruct developers to use global objects for extending with SlotFills.

Yup, totally agree there. When I mentioned `@woocommerce/slotfills`, it was in reference to which global the slotFill resides. Instead of `@woocommerce/components`, maybe there is a better place. In any case, its something that can be worked out as we go.

psealock:
>Additionally, there was extra work to do regarding `<PluginArea />` - at the time of writing (and probably now still) it couldn't be namespaced or scoped

💥 https://github.com/WordPress/gutenberg/pull/21908

psealock:
I think we're on the same page here, especially with a pattern set in Gutenberg. With the PR in Gutenberg to address scoping, I think we can close this PR and move on with slotFill.

PR #4218, opened 2020-04-24
Inbox notification: layout changes

### LINKED ISSUE #4099: Inbox notification: layout changes
image

Designs: https://www.figma.com/file/Pfw5Qa1UOw03RiJere3Wlo/Automattic-Components?node-id=4533%3A0

The inbox notification will remain fairly the same. It will be only necessary to implement the following changes:

- [ ] We're going to stop using icons and start using images. The images are optional and we'll only use them in some specific notifications
- [ ] Every notification will have a "Dismiss" button, allowing the user to remove the notification from their inbox. Clicking "Dismiss" will reveal a menu that prompts the user to choose whether to dismiss just this message or all messages. A dialog appears as a reminder that dismissed notifications cannot be accessed again, and a Snackbar confirms the action. We should keep the notification in the database, even if the user removes it in the frontend.
- [ ] The notification content can include links (I think this is already being covered, but I think it will be useful to have it here as remindrr to check this)
- [ ] The way to access the inbox panel in the home screen be slightly different (see #4097)
- [ ] We need to add additional even props in two events, to cover users that dismiss a notification and that view/click on them in the home screen:

`wcadmin_inbox_note_view`
New event prop: `screen` (value: home screen, analytics_revenue, customers, etc)

`wcadmin_wcadmin_inbox_action_click`*
New event prop: `screen` (value: home screen, analytics_revenue, customers, etc)
New event prop: `note_name_click` (name of the note that received a click in the main cta. eg: wc-admin-wc-helper). Instead of adding this new prop, we can also maintain the `note_name` for this
New event prop: `note_name_dismiss` (name of the note that was dismissed. eg: wc-admin-wc-helper)
New event prop: `note_name_dismiss_all` (`true` for dismiss all messages and `false` for dismiss this messages )
New event prop: `note_name_dismiss_all_confirmation` (`true` for yes I'm sure and `false` for cancel)

*: it would be great to remove the wcadmin duplication from the event name :)

octaedro:
Fixes #4099

This PR contains the following PRs:

- [x] Inbox notification: Added layout changes
https://github.com/woocommerce/woocommerce-admin/pull/4256

- [x] Inbox notification: Remove icons
https://github.com/woocommerce/woocommerce-admin/pull/4258

- [x] Inbox notification: Add dismiss functionality
https://github.com/woocommerce/woocommerce-admin/pull/4262

- [x] Inbox notification: Add undo snackbar button after deletion
https://github.com/woocommerce/woocommerce-admin/pull/4281

- [x] Inbox notification: add event recording
https://github.com/woocommerce/woocommerce-admin/pull/4320

- [x] Inbox notification: added new placeholder and empty card
https://github.com/woocommerce/woocommerce-admin/pull/4379

- [x] Inbox notification: add client unit tests
https://github.com/woocommerce/woocommerce-admin/pull/4386

### Changelog Note:
Enhancement: Inbox notification: layout changes


octaedro:
Hi @timmyc, you're totally right. For now, I removed the remaining `set_icon` calls we had, in the commit 214169c
I will follow up on this issue in a new PR, to follow our deprecation flows of the method.

Also, if you are testing on an ephemeral site and want to add notes type `banner` and `thumbnail` you can do it by executing the following SQL:

````
INSERT INTO `wp_wc_admin_notes` (`name`, `type`, `locale`, `title`, `content`, `icon`, `content_data`, `status`, `source`, `date_created`, `date_reminder`, `is_snoozable`, `layout`, `image`) VALUES
('wc-admin-task-list-complete', 'info', 'en_US', 'Congratulations - your store is ready for launch!', 'You’re ready to take your first order - may the sales roll in!', 'notice', '{}', 'unactioned', 'woocommerce-admin', '2020-03-10 18:38:10', NULL, 0, 'thumbnail', 'https://encrypted-tbn0.gstatic.com/images?q=tbn%3AANd9GcTlm2mX1h6GfMnT9sDD9GsJB7of3MZw3UDzCAQM9EBepiJzyxXH&usqp=CAU'),
('wc-admin-marketing-intro', 'info', 'en_US', 'Connect with your audience', 'Grow your customer base and increase your sales with marketing tools built for WooCommerce.', 'speaker', '{}', 'unactioned', 'woocommerce-admin', '2020-04-09 16:08:41', NULL, 0, 'banner', 'https://encrypted-tbn0.gstatic.com/images?q=tbn%3AANd9GcSAK9Xg4L6FGmDAW5UVtVEv1IXKtGV3-rxYLfAzOBF-fMUdmyWz&usqp=CAU');
````

timmyc:
Perfect, thanks for removing the `set_icon()` calls.

For other extensions ( like Woo core too ) which are still calling `set_icon()` - I'm still thinking we need to add the `set_icon()` method back to the class, and just log a deprecation warning. Otherwise we could get fatals happening from Woo core

octaedro:
It's a good idea. I added the method signature with the log of the deprecation warning in the commit e4c7426

Issue #4231, opened 2020-04-28
Homepage: Layout sections


psealock:
There are two layout sections, left and right.

### Criteria

* The left-hand side of the screen will “infinite scroll” as new content is delivered.
* The right-hand side of the screen will “stick” in place when the left scrolls – providing there is enough real estate to do so.
* When all Inbox notices are dismissed from the left side, the right side becomes full screen

![Screen Shot 2020-04-28 at 11 57 58 AM](https://user-images.githubusercontent.com/1922453/80432194-96912780-8947-11ea-8911-e5428cf036b8.png)

PR #4234, opened 2020-04-28
Homepage: layout sections

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/4231

This PR adds the layout interaction. I placed some nice pastel blocks to see how it behaves as well as a faux dismiss button.

This PR includes a fair amount of temporary code and markup which can be removed as those pieces come into play.

### Screenshots

Screen Shot 2020-04-28 at 3 20 19 PM

1. Scroll around the page. The left side will be an "infinite scroll" and the right will remain like normally displayed blocks.
2. Try smaller screen sizes.
3. Click "Dismiss All" to simulate clearing the inbox. The right panel will occupy the entire space.

@jameskoster Would you be able to take this for a spin to make sure the interaction is as intended?

### DIFF
--- a/client/homepage/layout.js
+++ b/client/homepage/layout.js
@@ -39,12 +39,15 @@ class Layout extends Component {
 
 	componentWillUnmount() {
 		window.removeEventListener( 'scroll', this.handleScroll );
+		window.cancelAnimationFrame( this.handle );
 	}
 
 	handleScroll( e ) {
 		if ( e.target === window.document ) {
-			this.setState( {
-				inboxHeight: this.getInboxHeight(),
+			this.handle = window.requestAnimationFrame( () => {
+				this.setState( {
+					inboxHeight: this.getInboxHeight(),
+				} );
 			} );
 		}
 	}
--- a/client/homepage/layout.js
+++ b/client/homepage/layout.js
@@ -15,7 +15,7 @@ class Layout extends Component {
 		super( props );
 
 		this.state = {
-			//inboxHeight: this.getInboxHeight(),
+			inboxHeight: this.getInboxHeight(),
 			showInbox: true,
 		};
 
@@ -34,25 +34,25 @@ class Layout extends Component {
 	}
 
 	componentDidMount() {
-		//window.addEventListener( 'scroll', this.handleScroll );
+		window.addEventListener( 'scroll', this.handleScroll );
 	}
 
 	componentWillUnmount() {
-		//window.removeEventListener( 'scroll', this.handleScroll );
+		window.removeEventListener( 'scroll', this.handleScroll );
 	}
 
 	handleScroll( e ) {
 		if ( e.target === window.document ) {
 			this.setState( {
-				//inboxHeight: this.getInboxHeight(),
+				inboxHeight: this.getInboxHeight(),
 			} );
 		}
 	}
 
 	render() {
-		const { showInbox } = this.state;
+		const { inboxHeight, showInbox } = this.state;
 		const inboxStyles = {
-			//maxHeight: inboxHeight,
+			maxHeight: inboxHeight,
 		};
 		return (
 			<div
--- a/client/homepage/layout.js
+++ b/client/homepage/layout.js
@@ -2,6 +2,8 @@
  * External dependencies
  */
 import { Component } from '@wordpress/element';
+import { Button } from '@wordpress/components';
+import classnames from 'classnames';
 
 /**
  * Internal dependencies
@@ -14,6 +16,7 @@ class Layout extends Component {
 
 		this.state = {
 			inboxHeight: this.getInboxHeight(),
+			showInbox: true,
 		};
 
 		this.handleScroll = this.handleScroll.bind( this );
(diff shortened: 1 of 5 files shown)


psealock:
Thanks for taking a look @jameskoster 

>I hope you don't mind, I tweaked the scrolling behaviour a bit

Not at all!

I like the alternative approach without js. I am seeing some oddness though.

Scrolling the left hand blue column, which is the entire page, scrolls the sidebar as well:

scroll1

When trying to see more of the green right side content you can't until the left side has completely finished scrolling. Even if I hover over the right side, it won't scroll, which is what I would have expected.

scroll2

>The shadow on the top bar appears when expected

The shadow appears only when the green right side is scrolled, but not the left. Is that the issue? I'm wondering if there isn't some solution in js land.

>No ugly scrollbar in the middle of the screen

Yeah, that is an eye sore. Looks like it may be possible to hide it https://www.geeksforgeeks.org/hide-scroll-bar-but-while-still-being-able-to-scroll-using-css/

psealock:
@jameskoster I reverted your commit so I can keep refining this method and created a branch with your changes here: [add/homepage-layout-sticky-columns](https://github.com/woocommerce/woocommerce-admin/tree/add/homepage-layout-sticky-columns). We certainly can keep exploring that approach, especially since it uses CSS only.

Changes here address the two issues cited: 

1. The scrollbar on the inbox is now hidden (tested in FF, Chrome, and Safari so far). 
2. Scrolling the inbox now activates the drop shadow on the header.

jameskoster:
Almost there! The scrolling still isn't quite right though - the behaviour changes based on where the cursor is - it should be consistent. That's kind of why I was trying to use the css only approach.

scroll

In the gif above, both columns should scroll together because the right-column doesn't fit in my window - it shouldn't stick at all.

In the gif below, the right column should always stick, regardless of where the cursor is.

scroll2

Hope that makes sense!

---

When the Inbox is dismissed, the right column should expand to full-width, which it does. However, the cards inside should be ~50% width to maintain the overall 2 column layout.

jameskoster:
One other Q - can we apply the same grid that we're using on the Analytics screens, if we aren't already? :)

psealock:
>the behaviour changes based on where the cursor is - it should be consistent.

I disagree with this. The reason is that in order to see the bottom of the right column, you have to scroll all the way to the bottom of the left column in order to see it. For example, if you have 30 inbox notifications and you'd like to get to store setup, you'll need to do lots of scrolling to get there.

But perhaps maybe your example, now in branch `add/homepage-layout-sticky-columns` is missing this one piece of logic you mentioned earlier:

>One thing I need help with to complete this change:
>position: sticky should only be applied to `.woocommerce-homepage-column` if its height if less than the height of the window.

So I download the sticky branch and do an experiment: The right green column is larger than the height of the window, so it should not be sticky as mentioned above. I'll add back in the default position of `static` and see what happens.

scroll3

In this case it scrolls as one unified page. Is that the desired effect? If so, we should just have it all static without the endless scroll.

Would you mind if we gathered a third opinion on this? @LevinMedia would you be able to weigh in on this one? The question, as I understand it, is **should the columns respond to scrolling based on the cursor location?** @jameskoster would you say that accurate?

You can compare this branch with `add/homepage-layout-sticky-columns` to do a comparison to see a difference. But be mindful there is a js interaction that is missing from the sticky branch described above that should remove stickyness in the right column when its height is larger than the window.

psealock:
>One other Q - can we apply the same grid that we're using on the Analytics screens, if we aren't already? :)

I'm not following, can you elaborate @jameskoster ?

jameskoster:
The layout on the Analytics screens aligns to a specific grid that we designed, which includes key details like breakpoint values. The home screen layout should align to that same grid, so that the column widths and responsive behaviour is consistent across the whole app.

Tbh I'm not sure if that grid was ever made into a component, or how re-usable it is in general, so perhaps this isn't as trivial as I hope. Unfortunately I can't find a Figma file with the details right now. Hopefully @LevinMedia can point you in the right direction.

client/homepage/layout.js
```diff
@@ -0,0 +1,126 @@
...
+
+	render() {
+		const { inboxHeight, showInbox } = this.state;
+		const inboxStyles = {
+			maxHeight: inboxHeight,
+		};
+		return (
+			<div
+				className={ classnames( 'woocommerce-homepage', {
+					hasInbox: showInbox,
```
jeffstieler, line 89: Should this be in kebab case?

client/homepage/layout.js
```diff
@@ -0,0 +1,126 @@
...
+		const scroll = window.scrollY;
+		const base = 90;
+		return `calc(100vh ${ base - scroll < 0 ? '+' : '-' } ${ Math.abs(
+			base - scroll
+		) }px)`;
+	}
+
+	componentDidMount() {
+		window.addEventListener( 'scroll', this.handleScroll );
+		this.inbox.current.addEventListener( 'scroll', this.handleInboxScroll );
```
joshuatf, line 39: Pretty edge-case scenario, but on resize this could create some height issues.  We could add a resize listener here as well to cover this.

Screen Shot 2020-05-11 at 3 27 12 PM

client/homepage/style.scss
```diff
@@ -0,0 +1,49 @@
...
+		margin-right: -16px;
+		flex-direction: column-reverse;
+	}
+}
+.woocommerce-homepage-column {
+	width: 100%;
+
+	&.is-inbox {
+		overflow-y: scroll;
+		/* Hid the scrollbar in Firefox, Safari, IE 11 */
```
joshuatf, line 26: Minor typo:
```suggestion
		/* Hide the scrollbar in Firefox, Safari, IE 11 */
```

client/homepage/style.scss
```diff
@@ -0,0 +1,49 @@
...
+		scrollbar-width: none;
+		overflow: -moz-scrollbars-none;
+		-ms-overflow-style: none;
+		margin-top: -45px;
+		padding-top: 45px;
+
+		/* Hid the scrollbar in Chrome */
```
joshuatf, line 33:
```suggestion
		/* Hide the scrollbar in Chrome */
```

jameskoster:
Apologies for the late reply @psealock.

> You can compare this branch with `add/homepage-layout-sticky-columns` to do a comparison to see a difference. But be mindful there is a js interaction that is missing from the sticky branch described above that should remove stickyness in the right column when its height is larger than the window.

I checked out that branch and the behaviour there is correct, albeit with the missing interaction added.

timmyc:
Coming in with the late, somewhat wishy-washy vote here - both approaches seem fine to me. If I had to pick between the two`add/hompage-sticky-layout` felt a bit nicer from an interaction standpoint to me. I feel we need to commit and ship at this point so we can begin filling in the homepage with all sorts of useful components for store operator peeps.

Also figured it might be worth checking out Calypso for infinite list inspiration - might be more helpful for Infinite Inbox: https://github.com/Automattic/wp-calypso/tree/master/client/components/infinite-list

psealock:
Closing this one in favor of https://github.com/woocommerce/woocommerce-admin/pull/4347

PR #4244, opened 2020-04-29
Guard against null themes

becdetat:
There was an error where nulls were in the themes array:

```
Uncaught TypeError: Cannot read property 'has_woocommerce_support' of null
```

Reloading the server made the error go away and I couldn't reproduce it, however this PR addresses the array.filter predicate that caused the error.


jeffstieler:
AFAIK we should be targeting `master` for PRs and cherry picking into release branches. Can you change this to `master` @becdetat ?

PR #4256, opened 2020-04-29
Inbox notification: Added layout changes

octaedro:
Fixes #partially-4099

This PR adds:
- 2 new columns to the table `wp_wc_admin_notes `: `layout` and `image`.
- Also, these two props were added to the Note class and in the Notes' API calls.
- Unit tests were modified to adapt them to the changes.

_There are some buttons that look weird because there are some inherited styles that modify them. As soon as they change the buttons will look normal again._

> **_NOTE:_**  As this is a part of a big functionality this branch will be merged with the branch `add/4099`, and it will be that branch that will be merged with `master`.
To make the code reviewing easier, this PR only has the code referring to the functionality described in the title.
If you want to test everything together, all the changes are merged in the branch: `add/4099_test_branch`.

### Screenshots
It should look like:
image

And the flow is something like this:
Screen Capture on 2020-05-19 at 11-54-33

### Detailed test instructions:
- dbDelta should add the `layout` and `image` columns to the table `wp_wc_admin_notes`
- After having the database updated you should add an image to the notes. Below I will leave an SQL to add 2 notes with the 2 types of images we need:
````
INSERT INTO `wp_wc_admin_notes` (`name`, `type`, `locale`, `title`, `content`, `icon`, `content_data`, `status`, `source`, `date_created`, `date_reminder`, `is_snoozable`, `layout`, `image`) VALUES
('wc-admin-task-list-complete', 'info', 'en_US', 'Congratulations - your store is ready for launch!', 'You’re ready to take your first order - may the sales roll in!', 'notice', '{}', 'unactioned', 'woocommerce-admin', '2020-03-10 18:38:10', NULL, 0, 'thumbnail', 'https://encrypted-tbn0.gstatic.com/images?q=tbn%3AANd9GcTlm2mX1h6GfMnT9sDD9GsJB7of3MZw3UDzCAQM9EBepiJzyxXH&usqp=CAU'),
('wc-admin-marketing-intro', 'info', 'en_US', 'Connect with your audience', 'Grow your customer base and increase your sales with marketing tools built for WooCommerce.', 'speaker', '{}', 'unactioned', 'woocommerce-admin', '2020-04-09 16:08:41', NULL, 0, 'banner', 'https://encrypted-tbn0.gstatic.com/images?q=tbn%3AANd9GcSAK9Xg4L6FGmDAW5UVtVEv1IXKtGV3-rxYLfAzOBF-fMUdmyWz&usqp=CAU');
````
[...]

PR #4258, opened 2020-04-29
Inbox notification: Remove icons

octaedro:
Fixes #partially-4099

In this PR the `icon` is removed from all the inbox notifications.

_Please be sure the [PR 4256](https://github.com/woocommerce/woocommerce-admin/pull/4256) has been approved before approving this one_

> **_NOTE:_**  As this is a part of a big functionality this branch will be merged with the branch `add/4099`, and it will be that branch that will be merged with `master`.
To make the code reviewing easier, this PR only has the code referring to the functionality described in the title.
If you want to test everything together, all the changes are merged in the branch: `add/4099_test_branch`.

### Screenshots
screenshot-one wordpress test-2020 04 29-17_52_02

- dbDelta should remove the `icon` column from the table `wp_wc_admin_notes`.
- Verify that the inbox notifications `icons` are not present anymore in the frontend.
- The same with the backend.

### Changelog Note:

Dev: Removed from all the inbox notifications

### DIFF
--- a/client/header/activity-panel/panels/inbox/card.js
+++ b/client/header/activity-panel/panels/inbox/card.js
@@ -3,7 +3,6 @@
  */
 import { Component } from '@wordpress/element';
 import PropTypes from 'prop-types';
-import Gridicon from 'gridicons';
 import VisibilitySensor from 'react-visibility-sensor';
 
 /**
@@ -32,7 +31,6 @@ class InboxNoteCard extends Component {
 				note_name: note.name,
 				note_title: note.title,
 				note_type: note.type,
-				note_icon: note.icon,
 			} );
 
 			this.hasBeenSeen = true;
@@ -82,7 +80,6 @@ InboxNoteCard.propTypes = {
 		id: PropTypes.number,
 		status: PropTypes.string,
 		title: PropTypes.string,
-		icon: PropTypes.string,
 		content: PropTypes.string,
 		date_created: PropTypes.string,
 		date_created_gmt: PropTypes.string,
--- a/client/header/activity-panel/panels/inbox/index.js
+++ b/client/header/activity-panel/panels/inbox/index.js
@@ -132,7 +132,6 @@ export default compose(
 				'title',
 				'content',
 				'type',
-				'icon',
 				'status',
 				'actions',
 				'date_created',
--- a/docs/examples/activity-panel-inbox.md
+++ b/docs/examples/activity-panel-inbox.md
@@ -110,6 +110,8 @@ class WooCommerce_Activity_Panel_Inbox_Example_Plugin_One {
 		$note->set_icon( 'info' );
 		$note->set_name( self::NOTE_NAME );
 		$note->set_source( 'wapi-example-plugin-one' );
+		$note->set_layout('plain');
+		$note->set_image('');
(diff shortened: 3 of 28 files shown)


client/header/activity-panel/panels/inbox/card.js
```diff
@@ -3,7 +3,6 @@
  */
 import { Component } from '@wordpress/element';
 import PropTypes from 'prop-types';
-import Gridicon from 'gridicons';
```
joshuatf, line 6: I see `<GridIcon />` still being used in this file and some icon related settings here which prevents my page from loading.

Following this down, I notice that icon is still being used in `ActivityCard`.  Just making sure that this is only meant to affect inbox notes and not other activity cards, is that right?

joshuatf: Another note on this- the placeholder for `ActivityCard` has an "icon" div.  Not sure if we plan on leaving this or not for other panels using the activity card.

Ideally, we could pass a boolean `icon` prop which would determine if the placeholder shows this or not.

octaedro: Hi @joshuatf,

I think that after merging the [PR 4256](https://github.com/woocommerce/woocommerce-admin/pull/4256) with `add/4099` we will have the missing elements that are blocking the normal behavior of this branch. For this purpose, I created the test_branch (`add/4099_test_branch`) that has all the changes together.

>Following this down, I notice that icon is still being used in ActivityCard. Just making sure that this is only meant to affect inbox notes and not other activity cards, is that right?

Yes, exactly.
I tried to decouple the `InboxNoteCard` from the `ActivityCard`. We still continue using the `ActivityCard` with orders and reviews but not with inbox notifications anymore.
My plan was trying to have an `InboxNoteCard` that we could use as an ActivityCard in the InboxPanel or eventually in the dashboard with no big modifications.

`Gridicon` is not used anymore In the `card.js`.

>Another note on this- the placeholder for ActivityCard has an "icon" div. Not sure if we plan on leaving this or not for other panels using the activity card.

This is a good point. I left the `ActivityCardPlaceholder` -when is loading- and an empty `ActivityCard` -when we don't have results- in the inbox panel, to follow the expected behavior for the inbox panel. In the Home screen it's different. If we don't have results the whole inbox messages column will disappear. For the loading situation, I wasn't sure if we will use something specific or the same `ActivityCardPlaceholder`.

I hope have responded to all the questions you had.

If you think we should change the approach just let me know and I will change it 🙂

jeffstieler:
I'm still seeing the icon in the loading placeholders.

Screen Shot 2020-05-12 at 5 37 27 PM

octaedro:
Hi @jeffstieler,

Yes, in this PR I only removed the icon from the inbox message. 

The inbox panel in this PR shows 3 kinds of cards:
1. The `InboxNoteCard`.
2. A generic `ActivityCard` (when there aren't cards to show -it uses an icon-).
3. And the `ActivityCardPlaceholder` (when it's loading the messages).

After having the new designs for the empty inbox notification and the placeholders, I created the [PR 4379](https://github.com/woocommerce/woocommerce-admin/pull/4379).
That PR removes the icons for every situation in the inbox panel.

Thank you for the alert about this problem Jeff 🙂

PR #4259, opened 2020-04-30
Fix Dashboard flash before OBW chunk loads.

### LINKED ISSUE #4252: Dashboard seen before OBW
**Describe the bug**

The dashboard can be seen before the OBW loads. This is a regression introduced by https://github.com/woocommerce/woocommerce-admin/pull/4094.

**To Reproduce**

1. Enable the profile wizard in the help tab
2. Go to WooCommerce > Home (or Dashboard)
3. Turn on network throttling - Fast 3G should do it
4. Refresh
5. See the wp-admin dash before the OBW chunk loads

**Expected behavior**

The OBW should be the first/only thing visible. Like this:

obw-no-flash-master

**Screenshots**

Here's what it looks like now:

dashboard-flash-optimize-bundles

jeffstieler:
Fixes #4252.

This PR seeks to fix the flash of Dashboard seen before the OBW chunk loads (introduced in the code splitting PR).

It accomplishes this by applying the `woocommerce-admin-full-screen` class to the body at the same time as the `woocommerce-admin-is-loading` class.

1. Enable the profile wizard in the help tab
2. Go to WooCommerce > Home (or Dashboard)
3. Turn on network throttling - Fast 3G or slower
4. Refresh
5. Verify that the dashboard is NOT seen during load

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -678,6 +678,9 @@ class Loader {
 
 		if ( self::is_admin_page() && $is_loading ) {
 			$classes[] = 'woocommerce-admin-is-loading';
+			// Include the full screen class now to prevent the dashboard
+			// from being shown while the <ProfileWizard /> chunk loads.
+			$classes[] = 'woocommerce-admin-full-screen';
 		}
 
 		$features = self::get_features();
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -678,9 +678,6 @@ class Loader {
 
 		if ( self::is_admin_page() && $is_loading ) {
 			$classes[] = 'woocommerce-admin-is-loading';
-			// Include the full screen class now to prevent the dashboard
-			// from being shown while the <ProfileWizard /> chunk loads.
-			$classes[] = 'woocommerce-admin-full-screen';
 		}
 
 		$features = self::get_features();
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -670,7 +670,8 @@ class Loader {
 		/**
 		 * Some routes or features like onboarding hide the wp-admin navigation and masterbar.
 		 * Setting `woocommerce_admin_is_loading` to true allows us to premeptively hide these
-		 * elements while the JS app loads. This class is removed when `<Layout />` is rendered.
+		 * elements while the JS app loads.
+		 * This class needs to be removed by those feature components (like <ProfileWizard />).
 		 *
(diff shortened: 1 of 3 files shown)


src/Loader.php
```diff
@@ -667,10 +667,20 @@ public static function add_admin_body_classes( $admin_body_class = '' ) {
...
+		 *
+		 * @param bool $is_loading If WooCommerce Admin is loading a fullscreen view.
+		 */
+		$is_loading = apply_filters( 'woocommerce_admin_is_loading', false );
+
+		if ( self::is_admin_page() && $is_loading ) {
 			$classes[] = 'woocommerce-admin-is-loading';
+			// Include the full screen class now to prevent the dashboard
+			// from being shown while the <ProfileWizard /> chunk loads.
+			$classes[] = 'woocommerce-admin-full-screen';
```
joshuatf, line 683: I'm a little hesitant to add the `woocommerce-admin-full-screen` class to every loading page.  It contains styles that are specific to full-screen pages and creates one more class that needs to be removed for components making use of the `woocommerce_admin_is_loading` filter.

joshuatf (commented):
Thanks for your work on this, Jeff.  I think as an alternative solution, we could try one of the following:

1. Only remove the `woocommerce-admin-is-loading` class once all chunks and components have loaded.
2. Create a separate loading/full-screen filter that's only removed by components taking advantage of full-screen.

My preference is definitely for option 1 if that's possible since that most closely aligns with the goals of the `woocommerce-admin-is-loading` class.

jeffstieler:
> 1. Only remove the `woocommerce-admin-is-loading` class once all chunks and components have loaded.
> 
> My preference is definitely for option 1 if that's possible since that most closely aligns with the goals of the `woocommerce-admin-is-loading` class.

I think this will work fine. We can remove the handling of the loading class from [`_Layout`](https://github.com/woocommerce/woocommerce-admin/blob/dc79d3622dcab2d349a45b54f60275d2eafc04d1/client/layout/index.js#L56) and instead require all full screen components to remove the class once they're mounted.

Does that work?

src/Loader.php
jeffstieler: Removed in 8160ba3 in favor of delaying the removal of the loading class.

joshuatf (approved):
Nice solution, @jeffstieler!  Testing well for me.  :shipit:

timmyc:
Thanks for following up on this bug Jeff 🙇

PR #4260, opened 2020-04-30
Add stored state between remote inbox notification runs

becdetat:
This adds a stored state option that lets the remote inbox notification system have a persistent state between runs. This is supplemented by a new rule processor that can perform comparison operations on values in the stored state object.

It also pulls the state of the products table at three different stages: first run, on adding a new product, and on performing a bulk import. This allows a rule to be created that will display a note on the first product creation, which is needed to satisfy #4299.

#### Data rule

- Add this URL to the `DATA_SOURCES` array in `DataSourcePoller`: https://gist.githubusercontent.com/becdetat/f17e9617e5ca1211d0579cf079862905/raw/1adef708c9bd16f928cde0e17383446c537f3bd9/rinds-specs.json
- Delete or trash all of your products. They won't be coming back.
- In the database, `DELETE FROM wp_options WHERE option_name = 'wc_remote_inbox_notifications_stored_state'`
- *Reload the page*. This sets the initial state of the stored state option correctly.
- Create a new product
- There should be a new inbox notification talking about your first product and shipping
- Delete the product you created
- Delete that option from the database again
- Delete all of the notes from the database (`DELETE FROM wp_wc_admin_notes`)
- Import a product data file like [this one](https://gist.githubusercontent.com/becdetat/e0dafd8fbe2049f1b2ac34f167d62e4b/raw/7f85132dbcf88c5f3d4e4e96437abab8c464bd42/dummy-data.csv).
- The shipping notification should be back

### Additional spec rules added in this PR

Stored state:

```
{
    "type": "stored_state",
    "index": "there_were_no_products",
    "operation": "=",
    "value": true
}
```


jeffstieler:
I'm getting stuck after this step. No notification appears, and the 'wc_remote_inbox_notifications_data' isn't recreated - even after running the `wc_admin_daily` cron task.

> * Delete or trash all of your products. They won't be coming back.
> * In the database, DELETE FROM wp_options WHERE option_name = 'wc_remote_inbox_notifications_data'
> * Create a new product
> * There should be a new inbox notification talking about shipping

becdetat:
Sorry @jeffstieler the branch seems to have had some issues. I've rebased everything and tested it and it seems to be working fine now.

src/Notes/WC_Admin_Note.php
```diff
@@ -23,7 +23,7 @@ class WC_Admin_Note extends \WC_Data {
 	const E_WC_ADMIN_NOTE_INFORMATIONAL = 'info';    // used for presenting informational messages.
 
 	// Note status codes.
-	const E_WC_ADMIN_NOTE_PENDING    = 'pending';    // the note is not actioned but shouldn't be displayed.
+	const E_WC_ADMIN_NOTE_PENDING    = 'pending';    // the note is pending - hidden but not actioned.
 	const E_WC_ADMIN_NOTE_UNACTIONED = 'unactioned'; // the note has not yet been actioned by a user.
 	const E_WC_ADMIN_NOTE_ACTIONED   = 'actioned';   // the note has had its action completed by a user.
 	const E_WC_ADMIN_NOTE_SNOOZED    = 'snoozed';    // the note has been snoozed by a user.
```
jeffstieler, line 29: Looks like we'll need to add "preunactioned" here - I'm getting this error in testing:

```
PHP Fatal error:  Uncaught WC_Data_Exception: The admin note status prop (preunactioned) is not one of the supported statuses. in /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce/includes/abstracts/abstract-wc-data.php:810
Stack trace:
#0 /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-admin/src/Notes/WC_Admin_Note.php(450): WC_Data->error('admin_note_inva...', 'The admin note ...')
```

jeffstieler: Also:
```
PHP Fatal error:  Uncaught WC_Data_Exception: The admin note icon prop cannot be empty. in /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce/includes/abstracts/abstract-wc-data.php:810
Stack trace:
#0 /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce-admin/src/Notes/WC_Admin_Note.php(412): WC_Data->error('admin_note_inva...', 'The admin note ...')
```

jeffstieler: Whoops - I had old data. Disregard all of this! 😅

src/RemoteInboxNotifications/DataSetupForProducts.php
```diff
@@ -0,0 +1,93 @@
...
+		add_action( 'product_page_product_importer', array( __CLASS__, 'run_on_product_importer' ) );
+		add_action( 'transition_post_status', array( __CLASS__, 'run_on_transition_post_status' ), 10, 3 );
+	}
+
+	/**
+	 * Set initial $data values.
+	 *
+	 * @param $data object The $data object.
+	 */
+	public static function init_data( &$data ) {
```
jeffstieler, line 32: What do you think about having this return a new `$data` object instead of modifying by reference? _(Personally I think it makes it more obvious at the call that data mutation is taking place)_

jeffstieler, line 32: What do you think about having this return a new `$data` object instead of modifying by reference? _(Personally I think it makes it more obvious at the call that data mutation is taking place)_

src/RemoteInboxNotifications/DataSetupForProducts.php
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	/**
+	 * Are there products query.
+	 *
+	 * @return bool
+	 */
+	private static function are_there_products() {
+		$product_query = new \WC_Product_Query( array( 'limit' => -1 ) );
+		$product_count = count( $product_query->get_products() );
```
jeffstieler, line 44: Same `WC_Product_Query` comments as here: https://github.com/woocommerce/woocommerce-admin/pull/4334#discussion_r431280201

src/RemoteInboxNotifications/DataSetupForProducts.php
```diff
@@ -0,0 +1,93 @@
...
+			return;
+		}
+		if ( 'done' !== $_REQUEST['step'] ) {
+			return;
+		}
+		// phpcs:enable
+
+		$data                         = RemoteInboxNotificationsEngine::get_data();
+		$data->there_are_now_products = self::are_there_products();
+		update_option( RemoteInboxNotificationsEngine::DATA_OPTION_NAME, $data );
```
jeffstieler, line 65: This can be handled in a follow up - looks like this is an opportunity for a data update method on the `RemoteInboxNotificationsEngine` class.

src/RemoteInboxNotifications/DataSetupForProducts.php
```diff
@@ -0,0 +1,93 @@
...
+	 *
+	 * @param string $new_status The new status.
+	 * @param string $old_status The old status.
+	 * @param Post   $post       The post.
+	 */
+	public static function run_on_transition_post_status( $new_status, $old_status, $post ) {
+		if (
+			'product' !== $post->post_type ||
+			'publish' !== $new_status ||
+			'publish' === $old_status
```
jeffstieler, line 82: I created a new product quickly (just a name) and published immediately. It did not pass this test, so I had to toggle it to draft and back to trigger the inbox note.

src/RemoteInboxNotifications/DataRuleProcessor.php
```diff
@@ -0,0 +1,39 @@
...
+	 * @param object $data Data.
+	 *
+	 * @return bool The result of the operation.
+	 */
+	public function process( $rule, $data ) {
+		if ( ! isset( $data->{$rule->index} ) ) {
+			return false;
+		}
+
+		$operand = $data->{$rule->index};
```
jeffstieler, line 31: This appears to be unused.

src/RemoteInboxNotifications/GetRuleProcessor.php
```diff
@@ -57,6 +57,8 @@ public static function get_processor( $rule_type ) {
 				return new PluginVersionRuleProcessor(
 					new PluginsProvider()
 				);
+			case 'data':
```
jeffstieler, line 60: I initially didn't understand why this was called `data` - I see now that it allows for rules based on persistent data, not just what can be calculated at engine run time (like if there were no products at one point).

What do you think about naming this `state` or `store_state` instead? If you're good with a change, I think changing all of the `$data` variables makes sense as well.

src/RemoteInboxNotifications/DataRuleProcessor.php
becdetat: It totally is! Thanks, removed.

src/RemoteInboxNotifications/DataSetupForProducts.php
becdetat: I agree, it does make it more obvious. Done in d9de0629.

src/RemoteInboxNotifications/DataSetupForProducts.php
becdetat: I think I've got it - de85c214

src/RemoteInboxNotifications/DataSetupForProducts.php
becdetat: I _think_ this should be fixed in de85c214

src/RemoteInboxNotifications/GetRuleProcessor.php
becdetat: I'm going to go with `stored_state` if you think that's ok

becdetat: Done in dbad715d

jeffstieler:
I'm running into some strange behavior testing this - when testing the post status transition vector, the query seems to miss the product that was just published - the count is always 1 less than the actual number of published products. Additionally, when the products query is issued from the `init_stored_state()` method, it **never** finds products.

Can you (or anyone) reproduce these issues? I've tried turning off object cache, but that didn't help.

becdetat:
> This should be the string ids, not an array

Whoops, fixed

becdetat:
> I'm running into some strange behavior testing this - when testing the post status transition vector, the query seems to miss the product that was just published - the count is always 1 less than the actual number of published products. Additionally, when the products query is issued from the init_stored_state() method, it never finds products.

> Can you (or anyone) reproduce these issues? I've tried turning off object cache, but that didn't help.

I can definitely reproduce both of these issues. I can't tell what would be causing this either, I'll have to do more experimentation tomorrow.

becdetat:
@jeffstieler I've moved back to using a SQL query to get the product count.

The problem with adding a new product was that the product is published with a taxonomy term of 'Uncategorized' which then switches to 'Simple' etc. `WC_Product_Query` filters on taxonomies 'external', 'grouped', 'simple' and 'variable' which excludes the new product. I couldn't figure out how to add 'uncategorized' to the set of included taxonomy terms.

The product with the `init_stored_state` flow is different, for some reason the SQL generated from the query is invalid (`AND (0=1)`):

```
SELECT SQL_CALC_FOUND_ROWS  wp_posts.ID FROM wp_posts  WHERE 1=1  AND ( 
  0 = 1
) AND wp_posts.post_type = 'product' AND ((wp_posts.post_status = 'publish')) GROUP BY wp_posts.ID ORDER BY wp_posts.post_date DESC LIMIT 0, 1
```

src/RemoteInboxNotifications/StoredStateSetupForProducts.php
```diff
@@ -44,18 +44,18 @@ public static function init_stored_state( $stored_state ) {
...
+				FROM {$wpdb->posts}
+				WHERE post_type = 'product'
+				AND post_status = 'publish'
+			"
 		);
-		$products      = $product_query->get_products();
-		$product_count = $products->total;
 
-		return $product_count > 0;
+		return $count > 0;
```
jeffstieler, line 58: Why this change? I don't think we should be querying the posts table directly because the data store classes can be overridden (so the posts table may not be where products are stored).

timmyc: Could we lean on the product lookup table here possibly? https://github.com/woocommerce/woocommerce/blob/bf042a428d32442c6df923c888076935eea59cf5/includes/admin/class-wc-admin-dashboard.php#L231 - maybe a `count` of records there would do the trick

becdetat: The product lookup table includes _all_ products including trashed ones and there's no discriminator, so it would still need a join to the posts table.

becdetat: This same issue is blocking #4515 :-(

timmyc:
@becdetat is this ready for another review? Looks like CI is failing on something in this branch

becdetat:
I need to have another look at the product count logic, and I'll check out Travis then. Should be ready for review after that.

becdetat:
Ok @timmyc this is ready for review again, thanks

timmyc:
Thanks @becdetat - I'm getting some errors on this branch when running it on a fresh JN site. I _think_ it might be due to the db changes that have merged to master around notifications. I did a build from master to double check this and no errors were shown on the same JN site. Could this get one more rebase so I can test it out again?

Also looking at the code, looks like some of the functionality will still run here if the remote inbox feature is disabled:

https://github.com/woocommerce/woocommerce-admin/pull/4260/files#diff-e970958fd6597aca954da1d11807ff78R32

But didn't fully look into that yet either, but wanted to make sure we aren't running any remote inbox items if the feature is turned off.

becdetat:
@timmyc 

> I'm getting some errors on this branch when running it on a fresh JN site. I think it might be due to the db changes that have merged to master around notifications. I did a build from master to double check this and no errors were shown on the same JN site. Could this get one more rebase so I can test it out again?

Done. I was getting errors even on a master build however and I don't think the rebase affected the db issue. I deleted the `woocommerce_admin_version` option from my JN site and notifications started working again.

> Also looking at the code, looks like some of the functionality will still run here if the remote inbox feature is disabled

True, I triggered the init in `FeaturePlugin`. Interestingly the feature configuration isn't available in `FeaturePlugin::includes()` because the filter that loads them hasn't been called yet, so I did the feature check in `RemoteInboxNotificationsEngine::on_admin_init()` instead (which executes later). See 4c25405.

Also 'interestingly' `npm run test:zip` generates a zip using the 'plugin' configuration, which led me on a wild goose chase trying to figure out why the engine wasn't running as it is disabled on that configuration.

timmyc:
@jeffstieler just wanting to make sure all your items are okay here before marking your review as resolved.

jeffstieler:
> @jeffstieler just wanting to make sure all your items are okay here before marking your review as resolved.

Are you OK with a direct SQL query for the product count? If so, I’ll retest this.

src/RemoteInboxNotifications/StoredStateSetupForProducts.php
timmyc: Okay looking at some related code in Woo Core - https://github.com/woocommerce/woocommerce/blob/172ceb51e8e63f23142a71f23a2415bf1c5b938a/includes/class-wc-tracker.php#L333 - kind of _feels_ like this is inline with that logic there. @jeffstieler  definitely a valid question and concern but I think this query is pretty similar to the one we are using there for the tracker product count. And the same logic is powering the other note too so I feel its okay - especially given it is just powering the notes side of things.

becdetat:
@jeffstieler 

> Are you OK with a direct SQL query for the product count? If so, I’ll retest this.

I moved it back to using `WC_Product_Query` after figuring out a workaround for when the query is triggered - https://github.com/woocommerce/woocommerce-admin/pull/4260/files#diff-64f3903390ba4be3ca876d3d5add82acR47

PR #4262, opened 2020-04-30
Inbox notification: Add dismiss functionality

octaedro:
Fixes #partially-4099

This PR adds the dismiss functionality to the inbox notifications.
A soft delete for inbox notifications is implemented here. The client will be able to delete one notification or all of them. A Snackbar will appear after the soft delete bringing the possibility to "Undo" action.

_Please be sure the [PR 4258](https://github.com/woocommerce/woocommerce-admin/pull/4258) has been approved before approving this one_

> **_NOTE:_**  As this is a part of a big functionality this branch will be merged with the branch `add/4099`, and it will be that branch that will be merged with `master`.
To make the code reviewing easier, this PR only has the code referring to the functionality described in the title.
If you want to test everything together, all the changes are merged in the branch: `add/4099_test_branch`.

### Screenshots
Screen Capture on 2020-05-15 at 13-37-13

- First, it's necessary to let dbDelta do its job updating the database (adding the `is_deleted`, `layout` and `image` columns and removing the `icon` column from the table `wp_wc_admin_notes`).
- To be able to see the inbox notifications, press Inbox in the navigation bar.
- The list of inbox notifications should be visible.
- Press `Dismiss` in any message.
- Choose `Dismiss this message`.
- A confirmation modal will be visible. Press `Yes, I'm sure`.
- The message shouldn't be visible anymore.
- Do the same choosing the option `Dismiss all messages`

### Changelog Note:
Dev: Added dismiss functionality to inbox notifications

### DIFF
--- a/client/header/activity-panel/index.js
+++ b/client/header/activity-panel/index.js
@@ -93,10 +93,14 @@ class ActivityPanel extends Component {
 		} ) );
 	}
 
-	handleClickOutside() {
+	handleClickOutside( event ) {
 		const { isPanelOpen, currentTab } = this.state;
+		const isClickOnModalOrSnackbar =
+			event.target.closest(
+				'.woocommerce-inbox-dismiss-confirmation_modal'
+			) || event.target.closest( '.components-snackbar__action' );
 
-		if ( isPanelOpen ) {
+		if ( isPanelOpen && ! isClickOnModalOrSnackbar ) {
 			this.togglePanel( currentTab );
 		}
 	}
--- a/client/header/activity-panel/panels/inbox/card.js
+++ b/client/header/activity-panel/panels/inbox/card.js
@@ -61,7 +61,7 @@ class InboxNoteCard extends Component {
 		} );
 	}
 
-	handleBlur( event, onToggle ) {
+	handleBlur( event, onClose ) {
 		const dropdownClasses = [
 			'woocommerce-admin-dismiss-notification',
 			'components-popover__content',
@@ -74,7 +74,7 @@ class InboxNoteCard extends Component {
 		if ( isClickOutsideDropdown ) {
 			event.preventDefault();
 		} else {
-			onToggle();
+			onClose();
 		}
 	}
 
@@ -82,12 +82,12 @@ class InboxNoteCard extends Component {
 		return (
 			<Dropdown
 				position="bottom right"
-				renderToggle={ ( { onToggle } ) => (
+				renderToggle={ ( { onClose, onToggle } ) => (
 					<Button
 						isTertiary
 						onClick={ onToggle }
 						onBlur={ ( event ) =>
-							this.handleBlur( event, onToggle )
+							this.handleBlur( event, onClose )
 						}
 					>
 						{ __( 'Dismiss', 'woocommerce-admin' ) }
--- a/client/header/activity-panel/panels/inbox/card.js
+++ b/client/header/activity-panel/panels/inbox/card.js
@@ -41,7 +41,6 @@ class InboxNoteCard extends Component {
 				note_name: note.name,
 				note_title: note.title,
 				note_type: note.type,
-				note_icon: note.icon,
 			} );
 
 			this.hasBeenSeen = true;
@@ -262,7 +261,6 @@ InboxNoteCard.propTypes = {
 		id: PropTypes.number,
 		status: PropTypes.string,
 		title: PropTypes.string,
-		icon: PropTypes.string,
 		content: PropTypes.string,
 		date_created: PropTypes.string,
 		date_created_gmt: PropTypes.string,
--- a/client/header/activity-panel/panels/inbox/card.js
+++ b/client/header/activity-panel/panels/inbox/card.js
@@ -62,12 +62,16 @@ class InboxNoteCard extends Component {
 	}
 
 	handleBlur( event, onToggle ) {
-		const targetIsRenderedContent = event.relatedTarget
-			? event.relatedTarget.className.includes(
-					'woocommerce-admin-dismiss-notification'
+		const dropdownClasses = [
+			'woocommerce-admin-dismiss-notification',
+			'components-popover__content',
+		];
+		const isClickOutsideDropdown = event.relatedTarget
+			? dropdownClasses.some( ( className ) =>
+					event.relatedTarget.className.includes( className )
 			  )
 			: false;
-		if ( targetIsRenderedContent ) {
+		if ( isClickOutsideDropdown ) {
 			event.preventDefault();
 		} else {
 			onToggle();
--- a/client/header/activity-panel/panels/inbox/card.js
+++ b/client/header/activity-panel/panels/inbox/card.js
@@ -1,25 +1,34 @@
 /**
  * External dependencies
  */
-import { Component } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+import { cloneElement, Component, Fragment } from '@wordpress/element';
+import { Button, Dropdown, Modal } from '@wordpress/components';
 import PropTypes from 'prop-types';
-import Gridicon from 'gridicons';
 import VisibilitySensor from 'react-visibility-sensor';
+import moment from 'moment';
 
 /**
  * Internal dependencies
  */
-import { ActivityCard } from '../../activity-card';
 import NoteAction from './action';
 import sanitizeHTML from 'lib/sanitize-html';
 import classnames from 'classnames';
 import { recordEvent } from 'lib/tracks';
+import './style.scss';
+import { H, Section } from '@woocommerce/components';
 
 class InboxNoteCard extends Component {
 	constructor( props ) {
 		super( props );
 		this.onVisible = this.onVisible.bind( this );
 		this.hasBeenSeen = false;
+		this.state = {
+			isDismissModalOpen: false,
+			dismissType: null,
+		};
+		this.openDismissModal = this.openDismissModal.bind( this );
+		this.closeDismissModal = this.closeDismissModal.bind( this );
 	}
 
 	// Trigger a view Tracks event when the note is seen.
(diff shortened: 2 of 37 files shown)


client/header/activity-panel/panels/inbox/card.js
```diff
@@ -39,9 +39,116 @@ class InboxNoteCard extends Component {
...
+									'woocommerce-admin'
+								) }
+							</Button>
+						</li>
+						<li>
+							<Button
+								onClick={ () => this.openDismissModal( 'all' ) }
+							>
+								{ __(
+									'Dismiss all message',
```
becdetat, line 85: Should this be 'Dismiss all messages'?

src/Notes/DataStore.php
```diff
@@ -71,7 +74,7 @@ public function read( &$note ) {
 		if ( 0 !== $note_id || '0' !== $note_id ) {
 			$note_row = $wpdb->get_row(
 				$wpdb->prepare(
-					"SELECT name, type, locale, title, content, icon, content_data, status, source, date_created, date_reminder, is_snoozable FROM {$wpdb->prefix}wc_admin_notes WHERE note_id = %d LIMIT 1",
+					"SELECT name, type, locale, title, content, content_data, status, source, date_created, date_reminder, is_snoozable, layout, image FROM {$wpdb->prefix}wc_admin_notes WHERE note_id = %d  AND is_deleted = 0 LIMIT 1",
```
becdetat, line 77: This is causing an error for me - icon is not set in `note_row` but it is needed further below.

client/header/activity-panel/panels/inbox/card.js
octaedro: Thank you @becdetat, you're totally right. I corrected it in the commit [cd71f7a](https://github.com/woocommerce/woocommerce-admin/pull/4262/commits/cd71f7abc8af42899e0e8d407b2af8960fefdc7b) and cherry-picked from the branch [add/4099_add_layout_changes](https://github.com/woocommerce/woocommerce-admin/pull/4256).

src/Notes/DataStore.php
octaedro: Thank you for the warning @becdetat.
Are you using the test branch (`add/4099_test_branch`) for testing?

becdetat: Oh, no I wasn't. That's much better.

becdetat:
Looks good and tests ok. The only thing is it takes 3 clicks to dismiss a single message then another click to open the inbox again. This is a design issue though and won't be as much of an issue once the inbox isn't a tray.

timmyc:
> The only thing is it takes 3 clicks to dismiss a single message then another click to open the inbox again. This is a design issue though and won't be as much of an issue once the inbox isn't a tray.

@jameskoster just making sure you see that little bit of design/interaction feedback.

src/Notes/WC_Admin_Note.php
```diff
@@ -504,6 +537,33 @@ public function clear_actions() {
 		$this->set_prop( 'actions', array() );
 	}
 
+	/**
+	 * Set note layout.
+	 *
+	 * @param string $layout Note layout.
+	 */
+	public function set_layout( $layout ) {
+		$this->set_prop( 'layout', $layout );
```
timmyc, line 546: Wondering if we should check to make sure the layout being set here is a valid one?

src/Notes/WC_Admin_Notes.php
```diff
@@ -100,6 +103,74 @@ public static function delete_notes_with_name( $name ) {
 		}
 	}
 
+	/**
+	 * Update a note.
+	 *
+	 * @param WC_Admin_Note $note The note that will be deleted.
```
timmyc, line 109: This should be "The note that will be updated"

src/Notes/WC_Admin_Notes.php
```diff
@@ -100,6 +103,74 @@ public static function delete_notes_with_name( $name ) {
...
+				'type'       => array( 'info', 'warning' ),
+				'status'     => array( 'unactioned' ),
+				'is_deleted' => 0,
+			)
+		);
+
+		$notes = array();
+		foreach ( (array) $raw_notes as $raw_note ) {
+			$note = new WC_Admin_Note( $raw_note );
+			$note->set_is_deleted( 1 );
```
timmyc, line 167: Should we just use the `delete_note()` function above?

src/API/Notes.php
```diff
@@ -77,6 +77,32 @@ public function register_routes() {
 				'schema' => array( $this, 'get_public_item_schema' ),
 			)
 		);
+
+		register_rest_route(
```
timmyc, line 81: I didn't check the full PR to see if you had done this already, but we need to have unit test coverage on these new endpoints. At a minimum it is nice to have a test to assert that users without proper permissions can not use these new endpoints, and ones that verify the deletes happen as expected.

jameskoster:
> @jameskoster just making sure you see that little bit of design/interaction feedback.

Due to the fact that there's no way to retrieve dismissed notices I think the extra click is fine. The only odd thing to me in the gif above is that the Inbox panel closes after dismissal. Dismissing a single notice doesn't indicate that the user is "finished" with the Inbox, so if possible we should keep the panel open.

jameskoster:
One other thing - and I appreciate it's late in the day to make this request - but would it be possible to use the `[Dropdown](https://wordpress.github.io/gutenberg/?path=/story/components-dropdown--default)` component for the dismiss menu? 

That way we inherit the 100% correct styles, other UX enhancements like animation, and ensure that we're consistent with the rest of the UI if these types of menus are ever updated in core.

src/Notes/WC_Admin_Note.php
octaedro: Good idea! I added a check here in the commit [88f37ed](https://github.com/woocommerce/woocommerce-admin/pull/4262/commits/88f37ed0e31221c64e900116a8acb22d4462cfde)

src/Notes/WC_Admin_Notes.php
octaedro: Totally! Nice catch Timmy. This was modified in the commit [d10a504](https://github.com/woocommerce/woocommerce-admin/pull/4262/commits/d10a5043c34c1278387376bd467f9e1df727dcb4)

src/Notes/WC_Admin_Notes.php
octaedro: Yes, you are right, thank you 🙂
This was changed in the commit [7ca0f10](https://github.com/woocommerce/woocommerce-admin/pull/4262/commits/7ca0f1041653ddf11e1cba08a58b0bbdbe1a0bb2)

src/API/Notes.php
octaedro: I added some unit tests in the [PR 4320 (Inbox notification: add event recording)](https://github.com/woocommerce/woocommerce-admin/pull/4320).
They cover all the processes but I missed the permission check you mentioned.
This was added in the commit [f4b62e2](https://github.com/woocommerce/woocommerce-admin/pull/4320/commits/f4b62e26d9fd84e3ccfe6eb056a543da0c3c1ff3)
Thank you Timmy!

octaedro:
Hi @jameskoster,
You are right about the odd behavior, this was changed in the commit [180a2ec](https://github.com/woocommerce/woocommerce-admin/pull/4256/commits/180a2eceac7a921dca1daba3953d69de8c941b44). I will update the gif now to have the last version in this PR description.

Currently, the dismiss menu (that appears after clicking the `Dismiss` button) is a `Dropdown` if you think that it's necessary I can modify the styles for this component.

jameskoster:
> Currently, the dismiss menu (that appears after clicking the Dismiss button) is a Dropdown if you think that it's necessary I can modify the styles for this component.

Oh, perfect! I thought I didn't see any animation, but maybe that just doesn't come across in the gif?

I appreciate we might not be able to do this right now, but I think it would make the dismissal UX a bit better if we could add some animation. Something like:

animoot

What do you think?

src/Notes/WC_Admin_Note.php
```diff
@@ -504,6 +537,36 @@ public function clear_actions() {
...
 	}
 
+	/**
+	 * Set note layout.
+	 *
+	 * @param string $layout Note layout.
+	 */
+	public function set_layout( $layout ) {
+		$valid_layouts = array( 'banner', 'plain', 'thumbnail' );
+		if ( in_array( $layout, $valid_layouts ) ) {
```
becdetat, line 547: Should this throw an error so that an invalid layout doesn't get swallowed?

src/Notes/WC_Admin_Notes.php
```diff
@@ -100,6 +103,73 @@ public static function delete_notes_with_name( $name ) {
 		}
 	}
 
+	/**
+	 * Update a note.
+	 *
+	 * @param WC_Admin_Note $note The note that will be updated.
+	 * @param array         $requested_updates a list of requested updates.
+	 */
+	public static function update_note( $note, $requested_updates ) {
```
becdetat, line 112: This is potentially confusing, I would expect this to be able to handle any updates I throw at it - ie it should be an exhaustive list of properties that would then need to be maintained. However it can only handle these three properties. I feel like the code that calls this should be handling this logic and just calling `$note->set_status` etc directly.

becdetat:
Testing looks good. One more thing I noticed is that the hit area on the dropdown is limited:

image

If you click in the red area the dropdown is dismissed. This may be an issue with the control rather than with this PR though, in which case this can be ignored. Sorry I can't do a GIF, my droplr is broken for some reason.

src/Notes/WC_Admin_Note.php
octaedro: Sounds good to me 🙂
I added it in the commit [50efe8e](https://github.com/woocommerce/woocommerce-admin/pull/4262/commits/50efe8e3080f604b386172b4354e7b1f72573af9)

src/Notes/WC_Admin_Notes.php
octaedro: I think that what you said makes sense. What I was thinking when I started with this refactor was that probably this method [was created](https://github.com/woocommerce/woocommerce-admin/blob/master/src/API/Notes.php#L201) with the idea of allowing some specific changes and nothing else.
What I did here specifically was:
1. move some parts of the method from `Notes.php` to this file. 
2. add the change that I needed, trying to keep the same structure.

The refactor you mentioned sounds interesting, maybe we can create a specific issue to discuss the change to make it.

octaedro:
> If you click in the red area the dropdown is dismissed. This may be an issue with the control rather than with this PR though, in which case this can be ignored. 

Nice catch @becdetat!
I repaired this in the commit  [9041d72](https://github.com/woocommerce/woocommerce-admin/pull/4262/commits/9041d72988bf8fed2fa26d8b353b4940ac508211)

octaedro:
> I appreciate we might not be able to do this right now, but I think it would make the dismissal UX a bit better if we could add some animation. Something like [this](https://cloudup.com/cDPbtQZjO5H).

Hi @jameskoster,
The solution [you suggested](https://cloudup.com/cDPbtQZjO5H) is great! I like it 😀
Is it ok if we add it in the next iteration?
It would be awesome if I could wrap up the functionality this week and what we currently have even though it isn't the best solution, it reaches the goal.

jameskoster:
I made a small tweak to that design, so that it makes use of the loading state:

reload

But yes, I think it's fine to implement this in the next iteration.

src/Notes/WC_Admin_Notes.php
becdetat: Ok I understand. I didn't realise this was part of a refactor and was adding to existing code. This should be fine then :-)

src/Notes/WC_Admin_Note.php
```diff
@@ -504,6 +513,42 @@ public function clear_actions() {
...
+	public function set_layout( $layout ) {
+		// If we don't receive a layout we will set it by default as "plain".
+		if ( empty( $layout ) ) {
+			$layout = 'plain';
+		}
+		$valid_layouts = array( 'banner', 'plain', 'thumbnail' );
+		if ( in_array( $layout, $valid_layouts, true ) ) {
+			$this->set_prop( 'layout', $layout );
+		} else {
+			$this->error( 'admin_note_invalid_data', __( 'The admin note date prop cannot be empty.', 'woocommerce-admin' ) );
```
becdetat, line 530: I like that you added some error handling here, but is the error message correct?

octaedro: Nice catch! I changed it in the commit [11c71cb](https://github.com/woocommerce/woocommerce-admin/pull/4262/commits/11c71cbb1003fcedc93b24911c97d1b25b1bbc0b).

client/header/activity-panel/index.js
```diff
@@ -93,10 +93,14 @@ class ActivityPanel extends Component {
 		} ) );
 	}
 
-	handleClickOutside() {
+	handleClickOutside( event ) {
 		const { isPanelOpen, currentTab } = this.state;
+		const isClickOnModalOrSnackbar =
```
timmyc, line 98: Is there any chance `event` or `event.target` could be undefined here? I suppose we could add check and clean up the code a bit since event.target.closest is used 2x.

PR #4263, opened 2020-04-30
Don't reschedule imports on failed imports

### LINKED ISSUE #4255: Too many actions being scheduled
**Describe the bug**
Originally reported in https://github.com/woocommerce/action-scheduler/issues/517. I was logged into customer site today that had this issue. The AS actions were all WC (upgrade tasks) or Historical Import. In the space of 5-10 minutes several hundred delete import batch actions were created.

Site orders:
Screen Shot 2020-04-29 at 3 51 40 PM

Delete customer batch action creation should have stopped before this:
Screen Shot 2020-04-29 at 3 51 21 PM

**To Reproduce**

Requires investigation

joshuatf:
Fixes #4255 

This PR removes the reschedule action that occurs after an import fails.

Since most of our responses of `-1` to failed imports occur won't be reconciled between rescheduling, this will result in a permanent loop of orders.

~~Without knowing the exact cause of the failure, we could send record when these errors occur, but not sure if we have better solutions than tracks or if this gives us a real fix.~~

Edit: In this case, the table was not created.  It's possible there are other areas where a sync would fail, but I think we can address them in follow-ups if we see a large amount of items not being synced.  Open to feedback on whether or not we should try and track this.

1. Throw in a `return -1` in `OrdersStatsDataStore::sync_order()` to mimic a failure.
1. Make sure the import completes (but not added to the lookup table) and another import is not queued in AS for the same order.
1. Optionally following testing instructions in https://github.com/woocommerce/woocommerce-admin/issues/4268

### DIFF
--- a/src/Schedulers/OrdersScheduler.php
+++ b/src/Schedulers/OrdersScheduler.php
@@ -166,11 +166,6 @@ class OrdersScheduler extends ImportScheduler {
 		);
 
 		ReportsCache::invalidate();
-
-		// Check if any syncs returned false or -1 and reschedule their import if so.
-		if ( count( array_intersect( $results, array( -1, false ) ) ) ) {
-			self::schedule_action( 'import', array( $order_id ) );
-		}
 	}
 
 	/**


joshuatf:
The database migrations never ran for the site this was tested against.  Migrations both from core and wc-admin had not yet taken place, although the version had already been updated in the database.

My best guess as to why that may have happened is the AS queue becoming overloaded during an update and not allowing the update to occur.  See https://github.com/woocommerce/woocommerce-admin/issues/4268

We may continue to see an issue with rows not being able to be inserted if this happens, but this PR will at least prevent failed rows from being attempted repeatedly.

Issue #4084, opened 2020-04-08
Stats overview


pmcpinto:
**User story:**
As a merchant
I want to take a quick glance at my store indicators
So I can have a sense of how my business is going while I'm in transit to make other tasks

![stats-overview](https://user-images.githubusercontent.com/22080/79792912-9324ed00-8304-11ea-911a-411c93a70597.png)

**Description:** we want to simplify the number of stats that we display on this home screen, as our goal is to turn this new screen as the central hub where we guide the user can quickly access different actions and data points. We're going to use this opportunity to:

- Start introducing traffic data (visitors and page views) for users that have Jetpack connected. Having a quick access to this type of data is important to most merchants, especially the ones that are just getting started and don't want to deal with tools like Google Analytics.
- Simplify the date ranges available in the Stats overview.

**Acceptance criteria:**
- [ ] The user can see by default 4 different data points: Views (Jetpack data), Visitors (Jetpack data), Orders, Total Sales
- [ ] The user can change the data points by clicking on the ellipsis menu. They can select up to six data points: views, visitors, total sales, orders, items sold and net sales
- [ ] The user can select between 3 different date ranges: today, week to date, month to date
- [ ] If the user doesn't have Jetpack installed, he will see a message asking the user to install it. 
- [ ] If the user clicks to install Jetpack, the plugin will be installed automatically and he'll be redirected to the Jetpack signup/login flow.
- [ ] If the user doesn't have Jetpack activated, he will see a message asking the user to activate it. 
- [ ] If the user clicks to activate jetpack, he'll be redirected to the Jetpack signup/login flow.
- [ ] If the user clicks in Visitors or Views, he'll be redirected to the Jetpack dashboard
- [ ] If the user clicks in another data point, such as Orders, he'll be redirected to the respective report in Analytics.
- [ ] If the user clicks in View detailed stats, he'll be redirected to the main Analytics dashboard

**Event tracking:**
Event name: wcadmin_statsoverview_indicators_toggle
Event props: indicator_name, status ( on | off )
Description: When a user toggles a data point off/on

Event name: wcadmin_statsoverview_indicators_click
Event props: indicator_name
Description: When a user clicks on a data point. Including "View detailed stats"

Event name: wcadmin_statsoverview_date_picker_update
Event props: date_range
Description: When a user modifies the date range

Event name: wcadmin_statsoverview_jetpack_click
Event props: link (install, connect, no thanks)
[...]

PR #4264, opened 2020-05-01
Homepage: Add stats overview selection

psealock:
Partially solves https://github.com/woocommerce/woocommerce-admin/issues/4084

This PR starts the Stats Overview component by adding a card and the stat selection logic. More logic, CSS and content will be added in a follow up. Jetpack stuff is omitted for now.

Screen Shot 2020-05-01 at 2 58 47 PM

1. Go to the Homepage.
2. Note the defaults and ones selected already match what is described in https://github.com/woocommerce/woocommerce-admin/issues/4084
2. Toggle a stat (note a very slow refresh, more on that below).
3. Refresh the page and verify your selection persists.

### Note on super slow update of toggle

Something is going on with wc-api and Fresh Data that makes this super slow to update. The Dashboard gets around this by keeping a duplicate copy of user preference data in state and updating that first. I think this is a mistake, but can be added later if need be. The better solution would be to refactor `wc-api/user` to a wp.data dataStore. I'd like to follow up on that.


client/homepage/stats-overview/test/index.js
```diff
@@ -0,0 +1,68 @@
...
+			'statsoverview_indicators_toggle',
+			{
+				indicator_name: 'revenue/total_sales',
+				status: 'off',
+			}
+		);
+	} );
+} );
+
+describe( 'StatsOverview stat selection', () => {
```
timmyc, line 38: I think this test name could be a bit more descriptive - maybe "StatsOverview toggle and persist stat preference" or something like that.

PR #4274, opened 2020-05-01
write RTL css property

rrennick:
I updated my machine to Catalina last weekend and had to install updated versions of Xcode, etc. `node-sass` needed to be rebuilt. Once it was rebuilt `/bin/packages/build.js` generated the following error:

```
internal/fs/utils.js:656
    throw new ERR_INVALID_ARG_TYPE(
    ^

TypeError [ERR_INVALID_ARG_TYPE]: The "data" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received an instance of Result
```

Looking at the code it looks like older versions of `fs.writeFileSync()` would default to the `css` property when passed an object. The latest version requires the parameter to be writeable. 

The PR updates the parameter to specify the `.css` property (consistent with the write two lines above). The PR includes a number of linting fixes. 

- Run `npm run start`

### DIFF
--- a/bin/packages/build.js
+++ b/bin/packages/build.js
@@ -76,16 +76,19 @@ function getBuildPath( file, buildFolder ) {
  */
 function buildFiles( files ) {
 	// Reduce files into a unique sets of javaScript files and scss packages.
-	const buildPaths = files.reduce( ( accumulator, filePath ) => {
-		if ( isJsFile( filePath ) ) {
-			accumulator.jsFiles.add( filePath );
-		} else if ( isScssFile( filePath ) ) {
-			const pkgName = getPackageName( filePath );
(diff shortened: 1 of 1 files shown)


bin/packages/build.js
```diff
@@ -147,9 +156,8 @@ function buildScssFile( styleFile ) {
 
 	const result = deasync( postCSSSync )();
 	fs.writeFileSync( outputFile, result.css );
-
 	const resultRTL = deasync( postCSSRTLSync )( result );
-	fs.writeFileSync( outputFileRTL, resultRTL );
+	fs.writeFileSync( outputFileRTL, resultRTL.css );
```
psealock, line 160: 👍

PR #4288, opened 2020-05-04
Homepage: Add stats numbers

psealock:
Partially solves #4084

Adds the SummaryNumbers used in the dashboard to show data. The SummaryNumber component is added without any styling, including small viewports. I'll handle that in the next PR.

### Screenshots

Screen Shot 2020-05-05 at 1 33 37 PM

### Other Changes

* Update `persistedQueries` such that the navigating to the Home Screen doesn't bring any extra query parameters along.
* Created `store-performance/utils.js` to house some functions for data handling that are re-used.

1. Go to homescreen.
2. Check that the selected overview stats are showing correct data for various periods.

### Changelog Note:

None: this is part of an ongoing feature

### DIFF
--- a/client/homepage/stats-overview/index.js
+++ b/client/homepage/stats-overview/index.js
@@ -4,7 +4,7 @@
 import { __ } from '@wordpress/i18n';
 import { compose } from '@wordpress/compose';
 import { Fragment } from '@wordpress/element';
-import { TabPanel, Button } from '@wordpress/components';
+import { TabPanel } from '@wordpress/components';
 import { xor } from 'lodash';
 import { withDispatch } from '@wordpress/data';
 import PropTypes from 'prop-types';
@@ -18,8 +18,10 @@ import {
 	EllipsisMenu,
 	MenuItem,
 	MenuTitle,
+	Link,
 } from '@woocommerce/components';
 import { getSetting } from '@woocommerce/wc-admin-settings';
+import { getNewPath } from '@woocommerce/navigation';
 
 /**
  * Internal dependencies
@@ -131,18 +133,18 @@ export const StatsOverview = ( { userPrefs, updateCurrentUserData } ) => {
 					</ul>
 				) }
 			</TabPanel>
-			<Button
+			<Link
 				className="woocommerce-stats-overview__more-btn"
-				isLink
+				href={ getNewPath( {}, '/analytics/overview' ) }
+				type="wc-admin"
 				onClick={ () => {
 					recordEvent( 'statsoverview_indicators_click', {
-						key: 'overview',
+						key: 'view_detailed_stats',
 					} );
 				} }
-				href="www.example.com"
 			>
 				{ __( 'View detailed stats' ) }
-			</Button>
+			</Link>
 		</Card>
 	);
 };
--- a/client/homepage/stats-overview/index.js
+++ b/client/homepage/stats-overview/index.js
@@ -4,10 +4,11 @@
 import { __ } from '@wordpress/i18n';
 import { compose } from '@wordpress/compose';
 import { Fragment } from '@wordpress/element';
+import { TabPanel } from '@wordpress/components';
 import { xor } from 'lodash';
 import { withDispatch } from '@wordpress/data';
 import PropTypes from 'prop-types';
-import { recordEvent } from 'lib/tracks';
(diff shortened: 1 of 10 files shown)


client/layout/controller.js
```diff
@@ -245,7 +245,7 @@ export function updateLinkHref( item, nextQuery, excludedScreens ) {
 	if ( isWCAdmin ) {
 		const search = last( item.href.split( '?' ) );
 		const query = parse( search );
-		const path = query.path || 'dashboard';
+		const path = query.path || 'homepage';
```
joshuatf, line 248: Should this check to see if the `homepage` feature is active first?

client/homepage/stats-overview/index.js
```diff
@@ -89,7 +104,47 @@ export const StatsOverview = ( { userPrefs, updateCurrentUserData } ) => {
...
+						title: __( 'Month to date', 'woocommerce-admin' ),
+						name: 'month',
+					},
+				] }
+			>
+				{ ( tab ) => (
+					<ul className={ listClasses }>
+						<StatsList
+							query={ {
+								period: tab.name,
```
joshuatf, line 128: Are we going to introduce a new param `period` for the performance indicators?  While I think this would be nice, I went with the presumption that we would fit performance indicators under the existing `before` and `after` filters that we currently use in https://github.com/woocommerce/woocommerce-admin/pull/4291.

I think it reduces the API complexity a bit if we can keep this params as consistent as possible, but definitely open for discussion.

client/homepage/stats-overview/test/index.js
```diff
@@ -66,3 +79,71 @@ describe( 'StatsOverview toggle and persist stat preference', () => {
 		} );
 	} );
 } );
+
+describe( 'StatsOverview rendering correct elements', () => {
```
joshuatf, line 83: ❤️ Loving all these tests.

client/homepage/stats-overview/style.scss
```diff
@@ -0,0 +1,46 @@
+.woocommerce-stats-overview {
+	.woocommerce-card__body {
+		padding: 0;
+	}
+	.woocommerce-summary {
+		margin: 0;
+	}
+}
+
+.woocommerce-stats-overview__more-btn {
```
joshuatf, line 10: This may be something you were planning on addressing in the follow-up, but one very minor styling issue is this focus on the more button.

Screen Shot 2020-05-07 at 8 42 19 PM

client/homepage/stats-overview/index.js
psealock: Right, so the other component that displays performance indicators is StorePerformance, which uses the `query` object and translates the `period` property into `before` and `after`. Here, I'm keeping consistent with that input even though its not derived from the url query. 

>I think it reduces the API complexity a bit if we can keep this params as consistent as possible, but definitely open for discussion.

As far as the API is concerned, `before` and `after` are still being used as is. StatsList converts the `period` to `before` and `after` using the re-used logic from StorePerformance.

Is that satisfactory?

client/layout/controller.js
psealock: Good call. I added a check in https://github.com/woocommerce/woocommerce-admin/pull/4288/commits/d67115a4ded108f8a574ba592437dacc13c74615

client/homepage/stats-overview/style.scss
psealock: Nice, I made the button `inline-block` so doesn't also take up the entire space in https://github.com/woocommerce/woocommerce-admin/pull/4288/commits/a5c007efd1c40f33e46d9dd15893447ada763b49 and this looks better when it has focus.

Screen Shot 2020-05-08 at 10 42 07 AM

psealock:
Thanks for the review @joshuatf ! I addressed your comments and replied to the question of performance indicators API.

psealock:
>I would expect n/a instead of 0 for this scenario:

I think you are right. This is the same behavior as the Dashboard, but I wonder if its not a regression:

Screen Shot 2020-05-14 at 3 02 42 PM

>This is a little tight on smaller screens. Can we make this 33.33%?

Yup, done in https://github.com/woocommerce/woocommerce-admin/pull/4288/commits/7c0640302a8a5279336407aa930628c83240768a

PR #4289, opened 2020-05-05
Increase selector specificity

pauloiankoski:
Fixes #
`Uncaught DOMException: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.`

The `.wrap` selector is too generic for the section. I work in a project hosted in VIP that has VaultPress as a MU Plugin and it injects a div containing the same _wrap_ class to the same parent div. However, it contains a few other classes on it while the selector we are applying here refers to the `<div class="wrap">` so increasing the specificity (elements with only the class wrap) we can solve this conflict and many others I don't realize.

### Accessibility

It doesn't apply.

- Add another element using the wrap class
- Open Orders page
- Check if you still can see the page title

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -49,7 +49,7 @@ if ( appRoot ) {
 
 	// Render notices just above the WP content div.
 	const wpBody = document.getElementById( 'wpbody-content' );
-	const wrap = wpBody.querySelector( '.wrap' );
+	const wrap = wpBody.querySelector( '[class="wrap"]' );
 	const noticeContainer = document.createElement( 'div' );
 
 	render(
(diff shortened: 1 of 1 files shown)


becdetat:
`wpBody.querySelector( '.wrap.woocommerce' )` is always returning null for me - there isn't a `.woocommerce` class anywhere I tested. Are you sure this is increasing the specificity?

pauloiankoski:
@becdetat I see it at least in these pages:
- wp-admin/admin.php?page=wc-settings
- wp-admin/admin.php?page=wc-status



Also, I'm not sure if you see it but I replaced the original querySelector from `.wrap` with `[class="wrap"]` in order to not find other divs using the same wrap class for other purposes.

becdetat:
>  I see it at least in these pages

Ok, nice, looks good.

PR #4302, opened 2020-05-06
Reduced variables and refactored delete_imported_items method

khanrn:
Reduced variable numbers and refactored the whole `delete_imported_items` method of the `Controller` class of *API/Report/Import/Controller.php* .

### DIFF
--- a/src/API/Reports/Import/Controller.php
+++ b/src/API/Reports/Import/Controller.php
@@ -259,24 +259,21 @@ class Controller extends \Automattic\WooCommerce\Admin\API\Reports\Controller {
 	 * @return WP_Error|WP_REST_Response
 	 */
 	public function delete_imported_items( $request ) {
-		$delete = ReportsSync::delete_report_data();
+		$result = array(
+			'status'  => 'success',
+			'message' => ReportsSync::delete_report_data(),
+		);
 
-		if ( is_wp_error( $delete ) ) {
-			$result = array(
-				'status'  => 'error',
-				'message' => $delete->get_error_message(),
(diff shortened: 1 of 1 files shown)


src/API/Reports/Import/Controller.php
```diff
@@ -259,24 +259,21 @@ public function cancel_import( $request ) {
...
-				'message' => $delete->get_error_message(),
-			);
-		} else {
-			$result = array(
-				'status'  => 'success',
-				'message' => $delete,
-			);
+		if ( is_wp_error( $result['message'] ) ) {
+			$result['status']  = 'error';
+			$result['message'] = $result['message']->get_error_message();
```
becdetat, line 269: Mutating `$result['message']` to avoid a variable seems off to me, I would rather have this explicit

src/API/Reports/Import/Controller.php
```diff
@@ -259,24 +259,21 @@ public function cancel_import( $request ) {
...
 		}
 
-		$response = $this->prepare_item_for_response( $result, $request );
-		$data     = $this->prepare_response_for_collection( $response );
-
-		return rest_ensure_response( $data );
+		return rest_ensure_response(
```
becdetat, line 272: I would definitely prefer having these assigned out to individual variables per method call, this way of nesting function calls is confusing.

becdetat:
Sorry but I don't think this adds any value to the code. While the existing code could probably be cleaned up a little I don't think reducing variables is the way to go about it. Thanks for the contribution though 👍

PR #4312, opened 2020-05-07
update unit test bootstrap for WC 4.2

rrennick:
This PR updates the unit test bootstrap to account for [test folder organization changes in WooCommerce 4.2](https://github.com/woocommerce/woocommerce/pull/26345).

### Testing

- Run `phpunit` against both WC 4.1 and `master`

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -52,6 +52,9 @@ function wc_admin_install() {
  */
 function wc_test_includes() {
 	$wc_tests_framework_base_dir = wc_dir() . '/tests';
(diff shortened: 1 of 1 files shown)


rrennick:
> I'm curious, do we need to handle the legacy case?

@psealock The 5.6 tests run against [WC 3.8.1](https://github.com/woocommerce/woocommerce-admin/blob/master/.travis.yml#L26) so it looks like that's covered.

> I thought tests were run against master of woocommerce.

We use the [latest tagged version](https://github.com/woocommerce/woocommerce-admin/blob/master/bin/install-wp-tests.sh#L172:L177)

> Incidentally, I'm getting test failures. I have master up to date locally.

Which ones?

PR #4182, opened 2020-04-23
Remove obsolete inbox messages

mattsherman:
Fixes #4153, Fixes #4154, Fixes #4155

The following inbox messages are no longer shown to users, as they no longer are needed in the onboarding experience:
* “New feature(s)”
* “Looking for the Store Notice setting”
* “Review your experience”

### Screenshots
These inbox messages will no longer be shown:

#### New installation

* Run this branch
* Perform a new installation of WooCommerce and WooCommerce Admin (make sure you activate woocommerce-admin plugin)
* With WP-CLI, update the woocommerce_admin_install_timestamp option and force the wc_admin_daily cron to run, so that the code that would have added the “Review your experience” message is run (if normally only shows the message after 3 days).

```
wp option update woocommerce_admin_install_timestamp 0
wp cron event run wc_admin_daily
```

* Verify that obsolete inbox messages do not appear

#### Upgrade

* Install WooCommerce and WooCommerce Admin

* With WP-CLI, update the woocommerce_admin_version option so that the WooCommerce Admin plugin thinks it needs to upgrade

```
wp option update woocommerce_admin_version 0.1
```

* Run this branch

* Perform an upgrade of WooCommerce Admin (make sure you activate woocommerce-admin plugin)

* With WP-CLI, update the woocommerce_admin_install_timestamp option and force the wc_admin_daily cron to run, so that the code that would have added the “Review your experience” message is run (if normally only shows the message after 3 days).

```
wp option update woocommerce_admin_install_timestamp 0
wp cron event run wc_admin_daily
```

* Verify that obsolete inbox messages do not appear
### Unit test instructions:

```
./vendor/bin/phpunit tests/notes/class-wc-tests-notes-data-store.php
```

### Changelog Note:

Tweak: Remove obsolete inbox onboarding messages.

Issue #4178, opened 2020-04-22
Change Inbox message: Giving feedback


pmcpinto:
We'll need to change the content, trigger and cadence this message (#1982) as a way to focus on the experience of new users.

**Title:** Give feedback
**Content:** Now that you’ve chosen us as a partner, our goal is to make sure we’re providing the right tools to meet your needs. We're looking forward to having your feedback on the store setup experience so we can improve it in the future.
**CTA leads to:** https://automattic.survey.fm/new-onboarding-survey
**CTA label:** Share feedback
**Timing:** 8 days after WooCommerce activation
**Cadence:** Once

becdetat:
This message was removed in #4182, should this issue be closed @pmcpinto ?

pmcpinto:
Hey @becdetat, this message should be implemented as it's going to replace the one that was removed. Thanks

becdetat:
Ok, no worries

PR #4317, opened 2020-05-07
Add back giving feedback note with some changes

becdetat:
Fixes #

This adds back the giving feedback note (removed in #4182) with new copy, per #4178.

### Screenshots

image

- Edit `src/Notes/WC_Admin_Notes_Giving_Feedback_Notes.php` and change the 8 to 0 on line 44
- Run the `wc_admin_daily` cron event
- Check that the giving feedback event is displayed

### Changelog Note:

Enhancement: Add back giving feedback note with new copy

### DIFF
--- a/src/Notes/WC_Admin_Notes_Giving_Feedback_Notes.php
+++ b/src/Notes/WC_Admin_Notes_Giving_Feedback_Notes.php
@@ -31,7 +31,7 @@ class WC_Admin_Notes_Giving_Feedback_Notes {
 	 * Possibly add a notice setting moved note.
 	 */
 	protected static function possibly_add_admin_giving_feedback_note() {
-		$name       = 'wc-admin-store-notice-giving-feedback';
+		$name       = 'wc-admin-store-notice-giving-feedback-2';
 		$data_store = \WC_Data_Store::load( 'admin-note' );
 
 		// We already have this note? Then exit, we're done.
--- a/src/Notes/WC_Admin_Notes_Giving_Feedback_Notes.php
+++ b/src/Notes/WC_Admin_Notes_Giving_Feedback_Notes.php
@@ -1,0 +1,65 @@
+<?php
+/**
+ * WooCommerce Admin (Dashboard) Giving feedback notes provider
+ *
+ * Adds notes to the merchant's inbox about giving feedback.
+ *
+ * @package WooCommerce Admin
+ */
+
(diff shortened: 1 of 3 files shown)


src/Notes/WC_Admin_Notes_Giving_Feedback_Notes.php
```diff
@@ -0,0 +1,65 @@
...
+	 */
+	public static function add_notes_for_admin_giving_feedback() {
+		self::possibly_add_admin_giving_feedback_note();
+	}
+
+	/**
+	 * Possibly add a notice setting moved note.
+	 */
+	protected static function possibly_add_admin_giving_feedback_note() {
+		$name       = 'wc-admin-store-notice-giving-feedback';
```
joshuatf, line 34: I think we'll want to update this to a new name.  If a user had the previous `wc-admin-store-notice-giving-feedback` note and skipped a version of wc-admin where it was removed to this version, the note would not be updated to the new content.

src/Notes/WC_Admin_Notes_Giving_Feedback_Notes.php
```diff
@@ -0,0 +1,65 @@
...
+			return;
+		}
+
+		// Otherwise, create our new note.
+		$note = new WC_Admin_Note();
+		$note->set_title( __( 'Give feedback', 'woocommerce-admin' ) );
+		$note->set_content( __( 'Now that you’ve chosen us as a partner, our goal is to make sure we\'re providing the right tools to meet your needs. We\'re looking forward to having your feedback on the store setup experience so we can improve it in the future.', 'woocommerce-admin' ) );
+		$note->set_content_data( (object) array() );
+		$note->set_type( WC_Admin_Note::E_WC_ADMIN_NOTE_INFORMATIONAL );
+		$note->set_icon( 'info' );
```
joshuatf, line 55: I know @octaedro is planning on removing note icons in https://github.com/woocommerce/woocommerce-admin/pull/4258.

I'm good with merging this first as long as we create a follow-up issue so it doesn't get overlooked after the icons are removed.  Also cc @octaedro if you want to rebase and handle directly or have any feedback.

src/Notes/WC_Admin_Notes_Giving_Feedback_Notes.php
becdetat: Great point, I've made it unique now.

PR #4326, opened 2020-05-08
Add JS unit tests to pre-push hook

mattsherman:
This PR adds `npm run test` to the pre-push hook.

If we are going to have unit tests, it is important that they run before code is pushed, so that developers do not inadvertently commit code that breaks other parts of the system.

- Create a new branch off of this branch
- Add a test somewhere that will fail, such as...

```
	it( 'should fail pre-push', () => {
		expect( true ).toBe( false );
	} );
```

- Commit this change
- Verify that unit tests are not run pre-commit
- Try to push the commit
- Verify that the pre-push hook fails

```
husky > pre-push hook failed (add --no-verify to bypass)
```

### Changelog Note:

Dev: Add JS unit tests to pre-push hook.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
 	},
 	"husky": {
 		"hooks": {
-			"pre-commit": "lint-staged",
+			"pre-commit": "lint-staged && npm run test",
 			"pre-push": "npm run -s install-if-no-packages && node bin/pre-push-hook.js"
 		}
 	},
--- a/package.json
+++ b/package.json
@@ -74,8 +74,8 @@
 	},
 	"husky": {
 		"hooks": {
-			"pre-commit": "lint-staged && npm run test",
-			"pre-push": "npm run -s install-if-no-packages && node bin/pre-push-hook.js"
+			"pre-commit": "lint-staged",
+			"pre-push": "npm run test && npm run -s install-if-no-packages && node bin/pre-push-hook.js"
 		}
 	},
 	"lint-staged": {


psealock (approved):
I think we avoided this due to the time it would take to run tests on each commit, especially since we have Travis doing it before merge.

But, if it means finding out things are broken sooner then thats a positive change. I'm good to merge this unless anyone from @woocommerce/wc-admin has differing opinion?

mattsherman:
> I think we avoided this due to the time it would take to run tests on each commit, especially since we have Travis doing it before merge.
> 
> But, if it means finding out things are broken sooner then thats a positive change. 

Agreed. 

I think that finding out that unit tests break when merging is too late. The JS unit tests currently take ~30 seconds to run, which seems like a reasonable amount of time for me to "spend" when committing something to gain assurance when developing that my changes are not inadvertently breaking something basic that I failed to consider.

> I'm good to merge this unless anyone from @woocommerce/wc-admin has differing opinion?

TIL... I'm not a member of that team! 😄 Just requested myself. I knew about @woocommerce/isotope

jeffstieler:
If it takes 30 seconds, how about a pre-push instead of pre-commit?

mattsherman:
> If it takes 30 seconds, how about a pre-push instead of pre-commit?

That seems like a reasonable compromise. Especially since we squash and merge.

mattsherman:
@jeffstieler Updated to call JS unit tests pre-push instead of pre-commit. Thanks for the suggestion!

psealock (approved):
I like the compromise as well. 🚢

PR #4327, opened 2020-05-08
Add `jest-dom` eslint plugin

mattsherman:
This PR adds the `jest-dom` eslint plugin to our configuration. This will help us ensure that our DOM-based tests are robust and readable.

`jest-dom/prefer-enabled-disabled` rule was disabled in our `d3chart` tests, which are enzyme-based, as (1) I believe this was a false positive (since my understanding of enzyme tests is that they are testing React components directly, not the DOM), and (2) that test should at some point be re-written to use React Testing Library instead of enzyme.

- Check out branch
- Add an assert that is not allowed by `jest-dom` to a test file, such as...

```
	it( 'should trigger es-lint', () => {
		render( <button /> );

		expect( screen.getByRole( 'button' ).innerHTML ).toBe( '' ); // bad
		expect( screen.getByRole( 'button' ) ).toBeEmpty(); // good
	} );
```

- Verify that your editor shows the eslint error (if your editor is configured to do so, which it definitely should be!)

```
jest-dom/prefer-empty: Use toBeEmpty instead of checking inner html.
```

- Verify that pre-commit hook fails if you try to commit the offending code

### Changelog Note:

Dev: Add `jest-dom` eslint plugin.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,6 +4,7 @@ module.exports = {
 		'prettier',
 		'plugin:testing-library/recommended',
 		'plugin:testing-library/react',
+		'plugin:jest-dom/recommended',
 	],
 	env: {
 		'jest/globals': true,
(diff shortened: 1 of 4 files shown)


psealock (approved):
Fantastic, thanks @mattsherman!

PR #4332, opened 2020-05-10
Task list: Inject Jetpack fallback flow into WCPay task

dechov:
This is a proposed change for conditionally adding Jetpack installation/activation and connection steps into the middle of the WCPay task, adapting components and flow management approach from Shipping and Tax tasks.

Ultimately, WCPay will not require Jetpack (see https://github.com/Automattic/woocommerce-payments/pull/638) but there is uncertainly around when these changes will be released. This change allows the Jetpack dependency to be controlled by a `define` in the WCPay plugin, which is now checked after the task's install step with an AJAX request (if Jetpack is not already connected).

### Screenshots

https://cloudup.com/ciny8Oz1Sam

With WooCommerce Payments not active, go to the task list » "Set up payments" » WooCommerce Payments "Set up". Verify that Jetpack is installed and connected if it needs to be.

WCPay **does** require Jetpack on `master` and in the newest `0.9.1` release. To test the case of WCPay not requiring Jetpack, either use the branch on https://github.com/Automattic/woocommerce-payments/pull/638 or add `define( 'WCPAY_BUILTIN_JETPACK_CONNECTION', true );` anywhere on your site.

### Testing combinations:
- [ ] Jetpack not installed, WCPay requires Jetpack.
- [ ] Jetpack not installed, WCPay doesn't require Jetpack.
- [ ] Jetpack installed but not active, WCPay requires Jetpack.
- [ ] Jetpack installed but not active, WCPay doesn't require Jetpack.
- [ ] Jetpack active but not connected, WCPay requires Jetpack.
- [ ] Jetpack active but not connected, WCPay doesn't require Jetpack.
- [ ] Jetpack connected, WCPay requires Jetpack.
- [ ] Jetpack connected, WCPay doesn't require Jetpack.

In all scenarios, the steps before redirect should match the steps after, and it should proceed to the _next_ step.

### DIFF
--- a/client/dashboard/task-list/tasks/payments/index.js
+++ b/client/dashboard/task-list/tasks/payments/index.js
@@ -356,9 +356,7 @@ export default compose(
 			isUpdateOptionsRequesting,
 		} = select( 'wc-api' );
 
-		const { getActivePlugins, isJetpackConnected } = select(
-			PLUGINS_STORE_NAME
-		);
+		const { getActivePlugins } = select( PLUGINS_STORE_NAME );
 		const activePlugins = getActivePlugins();
 		const profileItems = getProfileItems();
 		const options = getOptions( [
@@ -382,7 +380,6 @@ export default compose(
 		const methods = getPaymentMethods( {
 			activePlugins,
 			countryCode,
-			isJetpackConnected: isJetpackConnected(),
 			options,
 			profileItems,
 		} );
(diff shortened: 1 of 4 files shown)


dechov:
Closing out of optimism that this stop-gap will no longer be necessary or useful by the time of the next code freeze for release, and https://github.com/woocommerce/woocommerce-admin/pull/4332/commits/bf4d4351cbe543113d556d2816a50f46b6665b93 and https://github.com/woocommerce/woocommerce-admin/pull/4332/commits/fa78c318367aba6cdf014f2cd4ef9705de33c2b5 can be extracted into a new PR. Otherwise, this can be reopened.

PR #4334, opened 2020-05-11
Add product count rule

becdetat:
Note that this is branched off https://github.com/woocommerce/woocommerce-admin/pull/4260 and shouldn't be merged until that PR is merged.

This adds a rule that triggers if the given condition matches the number of products (for example, "> 10", or "<= 1000").

The test file used in earlier PRs contains a rule that triggers publishing an admin note when more than 10 products are present. To test, go through the procedure in #4143.

### Additional spec rules added by this PR

Product count:

```
{
    "type": "product_count",
    "operation": ">",
    "value": 10
}
```

### Changelog Note:

Enhancement: Add product count rule


src/RemoteInboxNotifications/ProductsProvider.php
```diff
@@ -0,0 +1,42 @@
...
+	public function are_there_products() {
+		return $this->get_product_count() > 0;
+	}
+
+	/**
+	 * Returns the number of products.
+	 *
+	 * @return integer The number of products.
+	 */
+	public function get_product_count() {
```
timmyc, line 30: Should we use [getProducts()](https://github.com/woocommerce/woocommerce/blob/master/includes/class-wc-product-query.php#L74) here?

becdetat: Thanks 🙏 done. I also added some unit tests since a little more logic ended up in the processor.

jeffstieler:
As a newcomer to reviewing the remote inbox stuff - is there a way to determine which commits are just for the functionality in this PR and not overlap with the parent branches?

becdetat:
That's annoying - it was a tiny PR but it probably needs a rebase now. I'll do this asap.

becdetat:
@jeffstieler Thanks for trying to review this. I've fixed the branch now, you should only see the relevant changes. The other two smaller PRs (#4337 and #4335) have also been fixed.

src/RemoteInboxNotifications/GetRuleProcessor.php
```diff
@@ -59,6 +59,10 @@ public static function get_processor( $rule_type ) {
 				);
 			case 'data':
 				return new DataRuleProcessor();
+			case 'product_count':
+				return new ProductCountRuleProcessor(
+					new \WC_Product_Query( array( 'limit' => -1 ) )
```
jeffstieler, line 64: This is a super expensive query that could potentially cause issues on stores with lots of products. (We also don't need product data, just the count of products..)

Can we rely on the `total` property on the return object instead? https://github.com/woocommerce/woocommerce/blob/1d1413b4ee669cd37ee025c93972ce785a10a0b9/includes/data-stores/class-wc-product-data-store-cpt.php#L2002

Should we be filtering by publish status?

If we can rely on the `total` property, we should be able to reduce the `limit` to `1` instead of ALL products (`-1`).

Aside: I was going to suggest we use `wp_count_posts()` as the [`WC_Install` class does](https://github.com/woocommerce/woocommerce/blob/4b3977868b789b83800f33e06c4ade1da5738c0e/includes/class-wc-install.php#L341), but that won't work if a store has replaced the data store for products.

src/RemoteInboxNotifications/ProductCountRuleProcessor.php
```diff
@@ -0,0 +1,45 @@
...
+	 * Performs a comparison operation against the number of products.
+	 *
+	 * @param object $rule The specific rule being processed by this rule processor.
+	 * @param object $data Remote inbox notifications data.
+	 *
+	 * @return bool The result of the operation.
+	 */
+	public function process( $rule, $data ) {
+		$products = $this->product_query->get_products();
+		$count    = count( $products );
```
jeffstieler, line 37: See above comment about using `->query()` and `total` instead.

src/RemoteInboxNotifications/GetRuleProcessor.php
becdetat: Nice! I've done this in 8db0229f55e7e0cedad824e10d5ebb0a39207f40

src/RemoteInboxNotifications/GetRuleProcessor.php
```diff
@@ -59,6 +59,15 @@ public static function get_processor( $rule_type ) {
 				);
 			case 'data':
 				return new DataRuleProcessor();
+			case 'product_count':
+				return new ProductCountRuleProcessor(
+					new \WC_Product_Query(
+						array(
+							'limit'    => 1,
+							'paginate' => 1,
```
jeffstieler, line 67: Do we not want to limit this query to published products?

Also, specifying `'return' => 'ids'` can further reduce the work done in the query class: https://github.com/woocommerce/woocommerce/blob/ee119e0a7ed24cf8c83983ff013bd188e3081d19/includes/data-stores/class-wc-product-data-store-cpt.php#L2004

Nit: `paginate` should be a boolean.

src/RemoteInboxNotifications/ProductCountRuleProcessor.php
```diff
@@ -0,0 +1,73 @@
...
+	 *
+	 * @param object $product_query The product query.
+	 */
+	public function __construct( $product_query = null ) {
+		$this->product_query = null === $product_query
+			? new \WC_Product_Query(
+				array(
+					'limit'    => 1,
+					'paginate' => true,
+					'return'   => array( 'id' ),
```
jeffstieler, line 29: This should be the string `ids`, not an array.

becdetat: Thanks @jeffstieler this is now fixed

PR #4335, opened 2020-05-11
Add order count rule

becdetat:
Note that this is branched off #4260 and shouldn't be merged until that PR is merged.

This adds a rule that triggers if the given condition matches the number of orders (for example, "> 10", or "<= 1000").

The test file used in earlier PRs contains a rule that triggers publishing an admin note when more than 10 orders are present. To test, go through the procedure in #4143.

### Additional spec rules added by this PR

Order count:

```
{
    "type": "order_count",
    "operation": ">",
    "value": 10
}
```

### Changelog Note:

Enhancement: Add order count rule

### DIFF
--- a/src/RemoteInboxNotifications/GetRuleProcessor.php
+++ b/src/RemoteInboxNotifications/GetRuleProcessor.php
@@ -40,6 +40,10 @@ class GetRuleProcessor {
 				return new PluginVersionRuleProcessor();
 			case 'stored_state':
 				return new StoredStateRuleProcessor();
+			case 'order_count':
+				return new OrderCountRuleProcessor(
+					new OrdersProvider()
+				);
 		}
 
 		return new FailRuleProcessor();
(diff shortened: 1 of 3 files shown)


src/RemoteInboxNotifications/GetRuleProcessor.php
```diff
@@ -59,6 +59,10 @@ public static function get_processor( $rule_type ) {
 				);
 			case 'data':
 				return new DataRuleProcessor();
+			case 'order_count':
+				return new OrderCountRuleProcessor(
+					new OrdersProvider()
```
jeffstieler, line 64: Nit: what do you think about having this occur by default in the `OrderCountRuleProcessor` constructor (so we only pass in an override for testing)?

becdetat: I don't think it's a bad idea, I just have an aversion to default parameters because I've seen code get burned by them in the past - having to debug a bad assumption with a default parameter can be difficult

becdetat: I've ended up doing this anyway since it was suggested in another PR as well...

Issue #4278, opened 2020-05-04
New inbox message: WooCommerce Subscriptions


pmcpinto:
**Title:** Do you need more info about WooCommerce Subscriptions?
**Content:**  WooCommerce Subscriptions allows you to introduce a variety of subscriptions for physical or virtual products and services. Create product-of-the-month clubs, weekly service subscriptions or even yearly software billing packages. Add sign-up fees, offer free trials, or set expiration periods.
**Segmentation:** New users with `subscriptions` as `product_type` in `wcadmin_storeprofiler_store_product_type_continue`
**Timing:** 1 day after plugin install
**Cadence:** Once
**CTA leads to:** https://woocommerce.com/products/woocommerce-subscriptions/?utm_source=inbox
**CTA label:** Learn more
**Type:** Promotional message

Issue #4102, opened 2020-04-09
Inbox notifications: new lifecycle


pmcpinto:
We're reviewing all the notifications that we display at this moment, as we plan to add new messages and remove some of the existing ones. 

I created a new issue for each notification that needs to be created/changed/removed, so this issue is going to be an epic to tie all the changes.

Issue #3794, opened 2020-02-27
Catalog Visibility Options: inbox notification


pmcpinto:
We want to offer a solution for users that skip the "Set up payments" task with the option "I don't need to take payments". Our assumption is that at this stage they only want to show a catalog of products:

**Title:** Transform WooCommerce into an online catalog
**Copy:** Do you want to showcase a catalog of products that aren’t currently for sale online? You can do this by removing eCommerce functionality. The added beauty of this solution is that when you do want to start selling, you can do so with the flick of a switch.
**Store type:** New users that skip the "Set up" payments task list with the option "I don't need to take payments" (event: `wcadmin_tasklist_payment_skip_task`, event prop: `skip_option` (value: tbd)) and don't have any payment option active
**Trigger:** 3 days after skipping the "Set up" payments task list with the option "I don't need to take payments" (event: `wcadmin_tasklist_payment_skip_task`, event prop: `skip_option` (value: tbd))
**Cadence:** once
**Action CTA leads to:** https://woocommerce.com/products/catalog-visibility-options/
**CTA copy:** Learn more
**Type**: Marketing

PR #4340, opened 2020-05-11
Refactor notes and conditions for adding notes

### LINKED ISSUE #4283: Allow users to turn-off extensions suggestions in the inbox panel
**User story:**
As a merchant
I want to have an option to stop receiving promotional content
So I can turn it off if I'm not getting value from it

**Description:** we want to provide a way for users to completely turn off notifications with suggestions of extensions. We're going to use the Marketplace Suggestions setting that's available on WooCommerce -Settings - Advanced - WooCommerce.com

Acceptance criteria:
- [ ] If the user already opted out in the Marketplace Suggestions setting, he shouldn't receive any notification about extensions available in the marketplace.
- [ ] If the user is opted-in in the Marketplace Suggestions setting and then he opts out, he should stop receiving notification about extensions available in the marketplace since that moment.
- [ ] Remove _contextual_ in the setting description: "We show ~contextual~ suggestions for official extensions that may be helpful to your store."

This setting should remove promotional content both from the existing inbox messaging system and the remote one (#4101). What would be the best way to identify the promotional messages? Through the "type" data point?

joshuatf:
Fixes #4283 

This PR:
* Refactors the admin notes and attempts to standardize the methods used between notes.  
* Adds conditions and filters to prevent or remove marketing notes if opted out.
* Adds tests for marketing notes.

It also adds the marketing type to the following notes.  Any others we want to add or remove from this list? /cc @pmcpinto :
* `WC_Admin_Notes_WooCommerce_Payments`
* `WC_Admin_Notes_Onboarding_Email_Marketing`
* `WC_Admin_Notes_Marketing`
* `WC_Admin_Notes_Draw_Attention`

Note that there are a couple of note classes that will probably require a follow-up since they contain multiple notes or break the pattern we traditionally use to add notes.

I've also kept the newly added filters and checks in `NoteTraits` as I think they may be useful as a trait to the RINDS system instead of moving these to an abstract class.

1. Under wccom settings `?page=wc-settings&tab=advanced&section=woocommerce_com` toggle off "Show Suggestions."
1. Note that any of the above notes were removed from your notes.
1. Attempt to add any of the above notes again by matching conditions and triggering the cron event `wc_admin_daily`.  `WC_Admin_Notes_Onboarding_Email_Marketing` does not have any extra conditions that need to be met and is easy to test.
1. Make sure all notes can still be added due to refactoring and that no errors occur.
1. Make sure all tests pass.

### DIFF
--- a/src/API/Plugins.php
+++ b/src/API/Plugins.php
@@ -192,7 +192,12 @@ class Plugins extends \WC_REST_Data_Controller {
 	 * @param string $slug The slug of the plugin being installed.
 	 */
 	private function create_install_plugin_error_inbox_notification_for_jetpack_installs( $slug ) {
-		WC_Admin_Notes_Install_JP_And_WCS_Plugins::possibly_add_install_jp_and_wcs_note( $slug );
+		// Exit early if we're not installing the Jetpack or the WooCommerce Services plugins.
+		if ( 'jetpack' !== $slug && 'woocommerce-services' !== $slug ) {
+			return;
+		}
+
+		WC_Admin_Notes_Install_JP_And_WCS_Plugins::possibly_add_note();
 	}
 
 	/**
--- a/src/Notes/NoteTraits.php
+++ b/src/Notes/NoteTraits.php
@@ -44,23 +44,38 @@ trait NoteTraits {
 	}
 
 	/**
-	 * Add the note if it passes predefined conditions.
+	 * Checks if a note can and should be added.
+	 *
+	 * @return bool
 	 */
-	public static function possibly_add_note() {
-		if ( self::note_exists() ) {
-			return;
-		}
-
+	public static function can_be_added() {
 		$note = self::get_note();
 
 		if ( ! $note instanceof WC_Admin_Note ) {
 			return;
 		}
 
+		if ( self::note_exists() ) {
+			return false;
+		}
+
 		if (
 			'no' === get_option( 'woocommerce_show_marketplace_suggestions', 'yes' ) &&
 			WC_Admin_Note::E_WC_ADMIN_NOTE_MARKETING === $note->get_type()
 		) {
+			return false;
+		}
+
+		return true;
+	}
+
+	/**
+	 * Add the note if it passes predefined conditions.
+	 */
+	public static function possibly_add_note() {
+		$note = self::get_note();
+
+		if ( ! self::can_be_added() ) {
 			return;
 		}
 
--- a/src/Notes/NoteTraits.php
+++ b/src/Notes/NoteTraits.php
@@ -33,4 +33,13 @@ trait NoteTraits {
 
 		return ( ( time() - $wc_admin_installed ) >= $seconds );
 	}
+
+	/**
+	 * Check if the note has been previously added.
+	 */
+	public static function note_exists() {
+		$data_store = \WC_Data_Store::load( 'admin-note' );
+		$note_ids   = $data_store->get_notes_with_name( self::NOTE_NAME );
+		return ! empty( $note_ids );
+	}
 }
--- a/src/Notes/NoteTraits.php
+++ b/src/Notes/NoteTraits.php
@@ -51,6 +51,19 @@ trait NoteTraits {
 			return;
 		}
 
-		self::add_note();
+		$note = self::get_note();
+
+		if ( ! $note instanceof WC_Admin_Note ) {
+			return;
+		}
+
+		if (
+			'no' === get_option( 'woocommerce_show_marketplace_suggestions', 'yes' ) &&
+			WC_Admin_Note::E_WC_ADMIN_NOTE_MARKETING === $note->get_type()
+		) {
+			return;
+		}
+
+		$note->save();
 	}
 }
--- a/src/Notes/NoteTraits.php
+++ b/src/Notes/NoteTraits.php
@@ -42,4 +42,15 @@ trait NoteTraits {
 		$note_ids   = $data_store->get_notes_with_name( self::NOTE_NAME );
 		return ! empty( $note_ids );
 	}
+
+	/**
+	 * Add the note if it passes predefined conditions.
+	 */
+	public static function possibly_add_note() {
+		if ( self::note_exists() ) {
+			return;
+		}
+
+		self::add_note();
+	}
 }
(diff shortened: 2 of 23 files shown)


src/API/Plugins.php
```diff
@@ -192,7 +192,12 @@ public function update_item_permissions_check( $request ) {
 	 * @param string $slug The slug of the plugin being installed.
 	 */
 	private function create_install_plugin_error_inbox_notification_for_jetpack_installs( $slug ) {
-		WC_Admin_Notes_Install_JP_And_WCS_Plugins::possibly_add_install_jp_and_wcs_note( $slug );
+		// Exit early if we're not installing the Jetpack or the WooCommerce Services plugins.
+		if ( 'jetpack' !== $slug && 'woocommerce-services' !== $slug ) {
```
joshuatf, line 196: @becdetat I hope you don't mind; I went back to your original method of checking and returning early here.

I think you were correct about the logic here- we'll just need to follow-up on this to move this check into a filter in onboarding.

src/Notes/NoteTraits.php
```diff
@@ -33,4 +33,37 @@ public static function wc_admin_active_for( $seconds ) {
...
+
+		$note = self::get_note();
+
+		if ( ! $note instanceof WC_Admin_Note ) {
+			return;
+		}
+
+		if (
+			'no' === get_option( 'woocommerce_show_marketplace_suggestions', 'yes' ) &&
+			WC_Admin_Note::E_WC_ADMIN_NOTE_MARKETING === $note->get_type()
```
joshuatf, line 62: I think it would be great to also check this at the `save()` level to prevent bypassing this, but currently there does not appear to be a way of doing this.

Filtering allowed types is one option, but this throws an exception and would require a bigger refactor.

pmcpinto:
@joshuatf here's my feedback:

> WC_Admin_Notes_WooCommerce_Payments
Can you share the link for this one? I'm not finding it

> WC_Admin_Notes_Onboarding_Email_Marketing
This is the one that the ask the user to sign up to receive email updates, correct? I don't see it as a promotion, considering that we're not directly promoting a specific extension. I'd remove the "marketing" type on this one.

> WC_Admin_Notes_Marketing
Is this the message about the marketing tab? I'd also remove the "marketing" type on this, as we're talking about a Core feature.

> WC_Admin_Notes_Draw_Attention
Can you also share the link for this one? I'm not finding it

We have 2 additional messages that should be considered as marketing, in case you want to test the remote system with more messages. However, considering the rules that they have I'm not sure if we'll need to keep it in the existing system instead of moving them to the remote system
#3791 (it was merged a couple of days ago)
#3794 (it's on the development sprint column)

joshuatf:
@pmcpinto So if I understand correctly, we're only updating marketing from non-core items.  This does not affect marketing items in core, correct?

>WC_Admin_Notes_WooCommerce_Payments
Can you share the link for this one? I'm not finding it

https://github.com/woocommerce/woocommerce-admin/blob/master/src/Notes/WC_Admin_Notes_WooCommerce_Payments.php

>This is the one that the ask the user to sign up to receive email updates, correct?

Yes, this is the one titled "Tips, product updates, and inspiration."  Updated to informational.

>WC_Admin_Notes_Marketing
Is this the message about the marketing tab? I'd also remove the "marketing" type on this, as we're talking about a Core feature.

This is the one that says "Grow your customer base and increase your sales with marketing tools built for WooCommerce."  Restored to informational.

>WC_Admin_Notes_Draw_Attention
Can you also share the link for this one? I'm not finding it

https://github.com/woocommerce/woocommerce-admin/blob/master/src/Notes/WC_Admin_Notes_Draw_Attention.php

src/Notes/WC_Admin_Notes_Onboarding_Email_Marketing.php
```diff
@@ -43,6 +40,6 @@ public static function possibly_add_onboarding_email_marketing_note() {
 		$note->set_name( self::NOTE_NAME );
 		$note->set_source( 'woocommerce-admin' );
 		$note->add_action( 'yes-please', __( 'Yes please!', 'woocommerce-admin' ), 'https://woocommerce.us8.list-manage.com/subscribe/post?u=2c1434dc56f9506bf3c3ecd21&amp;id=13860df971&amp;SIGNUPPAGE=plugin' );
-		$note->save();
+		return $note;
```
becdetat, line 43: Is the note type meant to be set to `E_WC_ADMIN_NOTE_MARKETING` here? I can't get any of the generated notes to go away when clearing the marketplace suggestions, the retrieved notes in `possibly_delete_marketing_notes` is empty.

joshuatf: @becdetat Based on @pmcpinto's feedback above, the only (for now) notes remaining that are marketing notes are `WC_Admin_Notes_Draw_Attention` and `WC_Admin_Notes_WooCommerce_Payments`.

But I'm guessing based on the above feedback that those notes may be reclassified as well shortly in which case we won't have any "marketing" notes to test against.

If this is the case, we'll have to manually test by adding `WC_Admin_Note::E_WC_ADMIN_NOTE_MARKETING` to notes and attempting to trigger them.

pmcpinto:
Hey @joshuatf @becdetat, we should keep `WC_Admin_Notes_WooCommerce_Payments` as promotional and change `WC_Admin_Notes_Draw_Attention` to informational/educational.

We have an epic with all the new inbox messages that should be shipped alongside the new home screen: #4102

Some of them should be seen as promotions, namely:
#3794
#4278
#4351

Let me know if this helps. Thanks

joshuatf:
Thanks for the info, @pmcpinto!

@becdetat So currently the note to test is `WC_Admin_Notes_WooCommerce_Payments`.  This note requires the store base country to be in the US and wcadmin must be active for at least a week.

See the added tests in https://github.com/woocommerce/woocommerce-admin/pull/4340/commits/e64e6300d15d9b5520ee23b595c2a847f36dd102 to force these settings.

src/Notes/WC_Admin_Notes_Onboarding_Email_Marketing.php
becdetat: Ok thanks

becdetat:
Thanks @joshuatf @pmcpinto that all makes sense.

This looks great, tests well, and the unit tests all pass 👍

PR #4345, opened 2020-05-11
Fix the WCPay method not appearing as recommended sometimes

### LINKED ISSUE #4249: Incorrect recommended treatment for WooCommerce Payments
**Describe the bug**
WooCommerce payments is missing the "Recommended" treatment in the payments task list

**To Reproduce**
1. Ensure jetpack is connected and the store based in the US.
2. Navigate to the payments task in the onboarding task list

**Expected behavior**
1. WooCommerce Payments should be noted as "recommended" and the setup button should use the primary action styles (pink) 
2. Stripe should no longer be noted as "Recommended" 

**Screenshots**

Observed:
Screen Shot 2020-04-28 at 9 12 05 PM

Expected:
Screen Shot 2020-04-28 at 9 47 42 PM

the colors of the default buttons can be ignored in the second screen shot - it assumes https://github.com/woocommerce/woocommerce-admin/issues/3897 has been merged.

DanReyLop:
Fixes #4249 (discussion and steps to reproduce there)

**The problem:** sometimes, on the Payments task of the onboarding task list, the WCPay card takes a bit to appear, and when it does, it doesn't get the "Recommended" pill. Instead, Stripe keeps being the recommended method, which is visually jarring since the WCPay card is heavily emphasized by its background color.

**The cause:** As discussed in #4249, I don't have the slightlest idea about how to reproduce this issue. Maybe @LevinMedia with his cursed laptop can try and see if this branch fixes the problem.

**The solution:** Instead of assign what payment method is the `recommendedMethod` immutably in the component constructor, recalculate it on `render`. The logic is exactly the same, the only change is that it's recalculated every time the component is rendered, so if WCPay is displayed 2 seconds after the page loads, when it's displayed it will get the `recommended` pill, and Stripe will lose its ribbon.

### DIFF
--- a/client/task-list/tasks/payments/index.js
+++ b/client/task-list/tasks/payments/index.js
@@ -44,13 +44,6 @@ class Payments extends Component {
 			enabledMethods,
 		};
 
-		this.recommendedMethod = 'stripe';
-		methods.forEach( ( method ) => {
-			if ( method.key === 'wcpay' && method.visible ) {
-				this.recommendedMethod = 'wcpay';
-			}
-		} );
-
 		this.completeTask = this.completeTask.bind( this );
 		this.markConfigured = this.markConfigured.bind( this );
 		this.skipTask = this.skipTask.bind( this );
@@ -229,6 +222,13 @@ class Payments extends Component {
 			);
 		}
 
+		let recommendedMethod = 'stripe';
+		methods.forEach( ( method ) => {
+			if ( method.key === 'wcpay' && method.visible ) {
+				recommendedMethod = 'wcpay';
+			}
+		} );
+
 		return (
 			<div className="woocommerce-task-payments">
 				{ methods.map( ( method ) => {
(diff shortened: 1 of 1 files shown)


client/dashboard/task-list/tasks/payments/index.js
```diff
@@ -226,6 +219,13 @@ class Payments extends Component {
 			);
 		}
 
+		let recommendedMethod = 'stripe';
+		methods.forEach( ( method ) => {
```
joshuatf, line 223: Can we move this check into `componentDidUpdate` so this doesn't occur on every render?

We can check if there is a difference in the visible methods between current and previous props.

joshuatf:
Hey @DanReyLop - just wanted to check if you have time to look at this or I'd be happy to help if you agree with the proposed changes.

client/dashboard/task-list/tasks/payments/index.js
DanReyLop: > We can check if there is a difference in the visible methods between current and previous props.

[getPaymentMethods](https://github.com/woocommerce/woocommerce-admin/blob/c7444a716f0d464779ec07b854d42c433777066e/client/dashboard/task-list/tasks/payments/methods.js#L33) returns a new array every time it's called, so the logic to check if the new list of visible methods is the same as the old visible methods would be as complex (if not more) than just recalculating `recommendedMethod` every time.

> Can we move this check into componentDidUpdate so this doesn't occur on every render?

Done in https://github.com/woocommerce/woocommerce-admin/pull/4345/commits/b482b15e80f4733dd1475252f5f8dece938c1862

jeffstieler:
@DanReyLop is this ready for another review?

DanReyLop:
@jeffstieler Yeah, it's ready for review, I forgot to change the label :/

PR #4347, opened 2020-05-12
Homepage: Add layout with sticky columns

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/4231
Replaces https://github.com/woocommerce/woocommerce-admin/pull/4234

This PR adds the layout interaction. I placed some nice pastel blocks to see how it behaves as well as a faux dismiss button.

This PR includes a fair amount of temporary code and markup which can be removed as those pieces come into play.

### Screenshots

Screen Shot 2020-04-28 at 3 20 19 PM

1. Scroll around the page, see if its interacting as expected.
2. Resize the screen and note the right column toggles between position sticky and static depending on window innerHeight.
3. Click "Dismiss All" to simulate clearing the inbox. The right panel will shift to the center.


jeffstieler:
Is there more to be done on this PR now that design has signed off on the interaction? (Curious about the in-progress label vs. needs-review)

Only comment I have on the code is if we'd want to consider making `<Layout />` a functional component instead of a class for later use of custom hooks. But that's certainly not a blocker to get this foundational piece in place.

psealock:
>Is there more to be done on this PR now that design has signed off

Yes, theres some code clean up (like making Layout a functional component as mentioned above) and now that the interaction is good to go, I can remove the dummy blocks in favor of adding in the ones we have already, like the Stats Overview.

client/homepage/layout.js
```diff
@@ -0,0 +1,77 @@
...
+				ref={ content }
+				style={ {
+					position: isContentSticky ? 'sticky' : 'static',
+				} }
+			>
+				<StatsOverview />
+			</div>
+		</div>
+	);
+	// }
```
jeffstieler, line 74: Whoops! 😛

PR #4356, opened 2020-05-12
Advanced Filters: fix currency initialization

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/4346

In https://github.com/woocommerce/woocommerce-admin/pull/4027 changes were made to Currency package such that a `Currency()` was a factory function returning a bunch of methods. This is a distinction from a currency object with properties that some components expect.

In `client/analytics/components/report-filters/index.js` the fix was to pass in the property object directly.

```diff
- currency={ Currency }
+ currency={ Currency.getCurrency() }
```

Other changes included are removal of `new Currency()` because Currency is no longer a class with a constructor, but now a regular function. I'm not sure how this didn't throw an error before.

1. Go to WooCommerce > Customers.
2. Click on the dropdown menu under Show:
3. Select to add an Advanced Filter.
4. Choose `Total Spend` or `AOV`.
5. The filters work as intended.

### Changelog Note:

Fix: Monetary Advanced Filters in Customers Report with correct currency object prop.

### DIFF
--- a/packages/components/src/filters/index.js
+++ b/packages/components/src/filters/index.js
@@ -94,8 +94,13 @@ class ReportFilters extends Component {
 	}
 
 	getDateQuery( query ) {
-		const { period, compare, before, after } = getDateParamsFromQuery( query );
-		const { primary: primaryDate, secondary: secondaryDate } = getCurrentDates( query );
+		const { period, compare, before, after } = getDateParamsFromQuery(
+			query
+		);
+		const {
+			primary: primaryDate,
+			secondary: secondaryDate,
+		} = getCurrentDates( query );
 		return {
 			period,
 			compare,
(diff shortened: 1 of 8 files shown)


psealock:
Thanks for taking a look @mattsherman ! I fixed the error in https://github.com/woocommerce/woocommerce-admin/pull/4356/commits/0821be761d78f3577d72d2d778cc03c0b1f80e59

packages/components/src/filters/index.js
```diff
@@ -224,15 +231,7 @@ ReportFilters.defaultProps = {
...
-	currency: new Currency( {
-		code: 'USD',
-		precision: 2,
-		symbol: '$',
-		symbolPosition: 'left',
-		decimalSeparator: '.',
-		thousandSeparator: ',',
-		priceFormat: '%1$s%2$s',
-	} ),
+	currency: Currency().getCurrency(),
```
mattsherman, line 234: Not introduced by this PR, so I'm not requesting a change in this PR, but I think the naming of the `Currency()` and `getCurrency()` are confusing 🤯, especially since there are also functions such as `formatCurrency()`, which doesn't take a 'currency', but rather an amount (a number)...

Ideally, things would be named something like...

`Currency()` -> `CurrencyFactory()`
`getCurrency() -> `getCurrencyConfig()`
`formatCurrency() -> `formatAmount()`

psealock: Absolutely, thanks for surfacing this! Here is an issue to track and tackle in the next cooldown period: https://github.com/woocommerce/woocommerce-admin/issues/4394

mattsherman: Thanks for creating the new issue!

Issue #4150, opened 2020-04-20
Marketing Plugins: Use plugin package


psealock:
wc-admin now has Plugin manipulation as part of its packages:

https://github.com/woocommerce/woocommerce-admin/tree/master/packages/data/src/plugins

Marketing should consider using the package or contributing to it instead of its own plugin operations, https://github.com/woocommerce/woocommerce-admin/tree/master/client/marketing/data.

cc @jconroy

jconroy:
Nice thanks @psealock

danielbitzer:
Hi @psealock, the new plugins data store returns only an array of plugin slugs for `getInstalledPlugins` and `getActivePlugins` selectors. 

For the marketing tab we need to get a lot more data about the installed plugins. See the current endpoint https://github.com/woocommerce/woocommerce-admin/blob/26f18357ac94f91d459548a830f396c041477118/src/API/MarketingOverview.php#L184-L186

Do you have any recommendations as to the best way forward in this case?

jconroy:
@danielbitzer ahh right

(just musing, and even if had a better API for the plugins themselves to provide that data we'd still need to pull it)

psealock:
@danielbitzer Just noting that there is a flag to add plugins to a whitelist, I know this isn't what you're asking about.

https://github.com/woocommerce/woocommerce-admin/blob/bd6ad79ebccdf1ee3d0036681aa3de548fe456c6/src/API/Plugins.php#L302-L304

>For the marketing tab we need to get a lot more data about the installed plugins.

The class `InstalledExtensions` just gets data about plugins, right? How about adding a method to the plugins API in `src/API/Plugins.php`? Maybe `/get-data`?

@joshuatf What do you think?

danielbitzer:
> The class `InstalledExtensions` just gets data about plugins, right? 

Yes and a lot of the data is specific to marketing.

> How about adding a method to the plugins API in `src/API/Plugins.php`? Maybe `/get-data`?

I think that would work well as long as we can inject marketing specific props to the data via a filter.

danielbitzer:
> How about adding a method to the plugins API in src/API/Plugins.php? Maybe /get-data?

@psealock I'm not quite sure what the best way to proceed with this is. In the marketing tab, when a plugin is activated we need to access certain props like **whether the plugin is configured or not**. This is something that we need to fetch dynamically after activation. 

I think there are few ways forward:
[...]

PR #4362, opened 2020-05-13
OBW: Get utils plugins data from wp.data

psealock:
Follow up to https://github.com/woocommerce/woocommerce-admin/pull/4357 which pulled plugin data from wcSettings. A problem may occur if plugins are added during the OBW, then the wcSettings global will be outdated. By gathering plugin data from the Plugins DataStore, we can be sure the data is fresh.

* Run through the OBW - select an extension for purchase
* Open the Dashboard (ensure the task list is shown)
* Verify no fatal errors are encountered
* Verify the "purchase" step is shown

### DIFF
--- a/client/dashboard/utils.js
+++ b/client/dashboard/utils.js
@@ -46,7 +46,8 @@ export function getCurrencyRegion( countryState ) {
  */
 export function getProductIdsForCart(
 	profileItems,
-	includeInstalledItems = false
+	includeInstalledItems = false,
+	installedPlugins
 ) {
 	const onboarding = getSetting( 'onboarding', {} );
 
@@ -57,7 +58,6 @@ export function getProductIdsForCart(
 	}
 
 	const productIds = [];
-	const plugins = getSetting( 'plugins', {} );
 	const productTypes = profileItems.product_types || [];
 
 	productTypes.forEach( ( productType ) => {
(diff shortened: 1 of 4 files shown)


psealock:
Potentially fixes https://github.com/woocommerce/woocommerce-admin/issues/4376

jconroy:
> Potentially fixes #4376

@psealock I haven't done any digging yet but we've still got #4150 on the table - so we might look at that at the same

psealock:
>still got #4150 on the table

I think thats going to solve this one.

client/dashboard/utils.js
```diff
@@ -57,15 +59,14 @@ export function getProductIdsForCart(
...
-	const plugins = getSetting( 'plugins', {} );
 	const productTypes = profileItems.product_types || [];
 
 	productTypes.forEach( ( productType ) => {
 		if (
 			onboarding.productTypes[ productType ] &&
 			onboarding.productTypes[ productType ].product &&
 			( includeInstalledItems ||
-				! plugins.installedPlugins.includes(
+				! installedPlugins.includes(
```
becdetat, line 69: I'm getting a `Cannot read property 'includes' of undefined` here :-(

psealock: Oops, I forgot to pass in an argument `installedPlugins`. Fix in https://github.com/woocommerce/woocommerce-admin/pull/4362/commits/9e9d7f5e2a5c669a4517668da4f2317fd0cdec5f

psealock:
Thanks for catching that one @becdetat ! This is ready for another review.

PR #4368, opened 2020-05-13
Enable tracking on the fly during onboarding

### LINKED ISSUE #4365: Tracks not available until refresh
**Describe the bug**
In the onboarding we dynamically enable tracking, however this is won't capture events until the next page load since the scripts are currently loaded server-side.

A PR exists in WC to add a method for this.  This issue will require using that newly added method to enable to load the scripts dynamically when opted in before recording events.

It would also be nice to update the debug messaging a bit so that we know if an event is actually recorded or not since it currently always shows the fired event.

**To Reproduce**

1. Set `woocommerce_allow_tracking` to `no`.
1. Visit the profiler first step and opt in to tracking at the modal.
1. Note the console debug messages show events as tracked, but the wp stats script and `recordEvent` method are not yet loaded.

**Expected behavior**
Expected to be able to track events client-side after opting in.

joshuatf:
Fixes #4365 

This PR attempts to immediately load tracking scripts when a user opts in without the need to refresh the page.

### Screenshots
Screen Shot 2020-05-13 at 9 16 46 PM

1. Check out this branch of WC - https://github.com/woocommerce/woocommerce/pull/26493
1. Set `woocommerce_allow_tracking` to `no`.
1. Open your console and enter `localStorage.setItem( 'debug', 'wc-admin:*' )`.
1. Visit the profiler first step and opt in to tracking at the modal.
1. Note the console debug messages shows the events and `_tqk` is defined if opted in.
1. Note that `shouldRecord` is `true` if opted in and not under development mode.

### DIFF
--- a/client/lib/tracks.js
+++ b/client/lib/tracks.js
@@ -18,7 +18,11 @@ const tracksDebug = debug( 'wc-admin:tracks' );
 export function recordEvent( eventName, eventProperties ) {
 	tracksDebug( 'recordevent %s %o', 'wcadmin_' + eventName, eventProperties, {
 		_tqk: window._tkq,
-		shouldRecord: window._tkq && process.env.NODE_ENV !== 'development',
+		shouldRecord:
+			process.env.NODE_ENV !== 'development' &&
+			!! window._tkq &&
+			!! window.wcTracks &&
+			!! window.wcTracks.isEnabled,
 	} );
 
 	if (
--- a/client/lib/tracks.js
+++ b/client/lib/tracks.js
@@ -16,7 +16,10 @@ const tracksDebug = debug( 'wc-admin:tracks' );
  */
 
 export function recordEvent( eventName, eventProperties ) {
-	tracksDebug( 'recordevent %s %o', 'wcadmin_' + eventName, eventProperties );
+	tracksDebug( 'recordevent %s %o', 'wcadmin_' + eventName, eventProperties, {
+		_tqk: window._tkq,
+		shouldRecord: window._tkq && process.env.NODE_ENV !== 'development',
+	} );
 
 	if (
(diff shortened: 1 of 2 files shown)


becdetat:
Everything looks good, the events were showing, however `window._tkq` was undefined until after a page reload after opting in, I don't know if that's expected or not but it doesn't seem right.

joshuatf:
Thanks for taking a look, @becdetat!

>Everything looks good, the events were showing, however window._tkq was undefined until after a page reload after opting in, I don't know if that's expected or not but it doesn't seem right.

If you've checked out the WC branch ( woocommerce/woocommerce#26493 ) it should be defined on any events after this modal if you check the box:

Screen Shot 2020-05-14 at 11 49 06 AM

If that's not the case, can you see if the stats script was appended to the body?  Any errors you're getting in the console?

psealock:
>If that's not the case, can you see if the stats script was appended to the body? Any errors you're getting in the console?

Definitely want to check this out, but it may not matter due to how `window._tkq` is interacted with. If it does not exist (yet) you can still fire an event to populate `window._tkq` and when the script (w.js) is run, its initialization process will handle it.

becdetat:
I'm sorry, I tried this again and `window._tkq` is set. I must have missed changing the woocommerce branch. :-/

joshuatf:
Nice catch, @jeffstieler!

Updated this to set `wcTracks.isEnabled` to `false`.  `_tqk` will still be defined, but the `recordEvent` function will return early.

One side effect of this is that `shouldRecord` will appear false in debugging on older versions of WooCommerce, but this is a minor, transient issue only related to debugging.

timmyc:
Cherry-picked to `release/1.2.0` in  07785f6

Issue #4293, opened 2020-05-05
Task list: broken WCPay "Settings" link


dechov:
<img width="730" alt="Screen Shot 2020-05-05 at 4 13 43 PM" src="https://user-images.githubusercontent.com/1867547/81111706-7afbc300-8eeb-11ea-81d1-0932282502ac.png">

After completing setup of the payment method, the "Settings" link should lead to the WooCommerce Payments gateway settings screen, but it just goes back to the task list.

I believe [this Link](https://github.com/woocommerce/woocommerce-admin/blob/ea368e4e12ef7366e8dc1478f4404f1883e19ec0/client/dashboard/task-list/tasks/payments/methods.js#L91-L97) needs `type="wp-admin"`. Also the `href` could be `getAdminLink( 'admin.php?page=wc-settings&tab=checkout&section=woocommerce_payments' )` (similar to [here](https://github.com/woocommerce/woocommerce-admin/blob/ea368e4e12ef7366e8dc1478f4404f1883e19ec0/client/dashboard/task-list/tasks.js#L154-L156)).

PR #4373, opened 2020-05-14
Task List: Small fixes and adjustments to WCPay task

dechov:
A set of small fixes and tweaks to the WCPay task. In rough order of decreasing priority:
- https://github.com/woocommerce/woocommerce-admin/commit/09ac96f0db759e10c46d685363dbbaa559bd5c40: fixes https://github.com/woocommerce/woocommerce-admin/issues/4293 by changing the "Settings" link type
- https://github.com/woocommerce/woocommerce-admin/commit/ee6df5b7707d98e9408db952a0d8dc5865a54e75: modifies success query param check to match the redirect [[code](https://github.com/Automattic/woocommerce-payments/blob/65e92328f8cea9dc685feed3823a42e8f6baa3eb/includes/class-wc-payments-account.php#L415)]. (The error message won't actually be triggered until [this change](https://github.com/Automattic/woocommerce-payments/compare/update/task-list-return-url) is made.)
- https://github.com/woocommerce/woocommerce-admin/commit/9eecc39023cfda6175680cc6f3dd6e8cf41e4564: removes request to update "enabled" setting, as it can be interrupted by the redirect (leading to an occasional error notice after clicking "Verify details" [[screenshot](https://cloudup.com/cqk8keEsGI0)]), and it's set when the connection is finalized anyway [[code](https://github.com/Automattic/woocommerce-payments/blob/65e92328f8cea9dc685feed3823a42e8f6baa3eb/includes/class-wc-payments-account.php#L404)]
- https://github.com/woocommerce/woocommerce-admin/commit/898739a478866aa45e8f20ae30eed668f2dbecf0: 1px font size tweak to avoid "Terms of Service" widow [[before](https://user-images.githubusercontent.com/1867547/81899521-0aa70e80-9589-11ea-8cd7-1723c43783f8.png), [after](https://user-images.githubusercontent.com/1867547/81899526-0d096880-9589-11ea-8456-1fc0f5627d04.png)] – not sure whether it's the same on all devices/browsers
- https://github.com/woocommerce/woocommerce-admin/commit/aedd06f34802c4994f4c900c4934d7d14a620767: trivial refactor to remove an unnecessary Fragment

(Happy to drop some off the bottom in pursuit of a timely merge for the top ones.)

### Accessibility

- Needs fresh Jetpack connection that hasn't had a WooCommerce Payments account (or "force onboarding" flag set), and dev mode: `define( 'WCPAY_DEV_MODE', true );`
- With WooCommerce Payments plugin deactivated, click "Set up payments" in the task list.
- Verify that the Terms of Service line in the payment method fits on one line. Click "Set up".
- Verify that no error notice shows up right before redirecting to stripe.com.
- To verify the success message on return from stripe.com, apply the change [here](https://github.com/Automattic/woocommerce-payments/compare/update/task-list-return-url).
- Verify that the Settings link takes you to the WooCommerce Payments settings screen, not the task list.

### DIFF
--- a/client/dashboard/task-list/style.scss
+++ b/client/dashboard/task-list/style.scss
@@ -363,6 +363,10 @@
 		.text-style-strong {
 			font-weight: bold;
 		}
+
+		p {
+			font-size: 12px;
+		}
 	}
 
 	.woocommerce-task-payment__after {
--- a/client/dashboard/task-list/tasks/payments/methods.js
+++ b/client/dashboard/task-list/tasks/payments/methods.js
@@ -12,6 +12,7 @@ import interpolateComponents from 'interpolate-components';
  */
 import {
 	getSetting,
+	getAdminLink,
 	WC_ASSET_URL as wcAssetUrl,
 } from '@woocommerce/wc-admin-settings';
 import { Link } from '@woocommerce/components';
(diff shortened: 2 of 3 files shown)


dechov:
Thanks @becdetat for taking a look!

PR #4374, opened 2020-05-14
Add Jetpack connection to plugin benefits step

### LINKED ISSUE #4363: Redirect the user to the JP account creation flow right after the plugin install
Removing the JP account creation flow from the profiler caused a significant drop in terms of the number of JP accounts created coming from Woo (context here: p1HpG7-9c2-p2)

This drop is higher than expected so we'll need to start redirecting users again to the JP account creation flow right after they install the plugin:

We don't need to display the install plugins screen. 

So the flow is going to be: store profiler - JP & WCS screen - JP account creation flow - task list

joshuatf:
Fixes #4363 

* Adds the Jetpack connection to the plugin benefits screen.
* Adds an `autoConnect` and `onError` prop to `Connect`.
* Fixes an issue where existing `activePlugins` were overwritten in state on plugin activation.

### Screenshots
Screen Shot 2020-05-14 at 7 00 35 PM

1. Deactivate and/or uninstall Jetpack and WCS.
1. Walk through the wizard attempting to install and connect via "Yes please!"
1. Make sure plugins are installed and you're redirect to the Jetpack connection flow.
1. Walk through the wizard attempt to skip by clicking "No thanks"
1. Make sure that you land in the task list without installing or connecting to Jetpack.
1. Re-enable the profiler and attempt these all steps again with Jetpack already connected.

### DIFF
--- a/packages/data/src/plugins/actions.js
+++ b/packages/data/src/plugins/actions.js
@@ -12,10 +12,11 @@ import TYPES from './action-types';
 import { WC_ADMIN_NAMESPACE } from '../constants';
 import { pluginNames } from './constants';
 
-export function updateActivePlugins( active ) {
+export function updateActivePlugins( active, added ) {
 	return {
 		type: TYPES.UPDATE_ACTIVE_PLUGINS,
 		active,
+		added,
 	};
 }
 
@@ -123,7 +124,7 @@ export function* activatePlugins( plugins ) {
 		} );
 
 		if ( results && results.status === 'success' ) {
-			yield updateActivePlugins( results.activatedPlugins );
+			yield updateActivePlugins( null, results.activatedPlugins );
 			return results;
 		}
 
--- a/packages/data/src/plugins/actions.js
+++ b/packages/data/src/plugins/actions.js
@@ -12,19 +12,19 @@ import TYPES from './action-types';
 import { WC_ADMIN_NAMESPACE } from '../constants';
 import { pluginNames } from './constants';
 
-export function updateActivePlugins( active, added ) {
+export function updateActivePlugins( active, replace = false ) {
 	return {
 		type: TYPES.UPDATE_ACTIVE_PLUGINS,
 		active,
-		added,
+		replace,
 	};
 }
 
-export function updateInstalledPlugins( installed, added ) {
+export function updateInstalledPlugins( installed, replace = false ) {
 	return {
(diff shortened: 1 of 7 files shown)


packages/data/src/plugins/test/reducer.js
```diff
@@ -35,6 +35,31 @@ describe( 'plugins reducer', () => {
...
+		/* eslint-disable dot-notation */
+
+		expect( state.requesting[ 'getActivePlugins' ] ).toBe( false );
+		expect( state.errors[ 'getActivePlugins' ] ).toBe( false );
+		/* eslint-enable dot-notation */
+
+		expect( state.active ).toHaveLength( 2 );
+		expect( state.active[ 1 ] ).toBe( 'woocommerce-services' );
+	} );
+
```
psealock, line 62: 👍

packages/data/src/plugins/actions.js
```diff
@@ -123,7 +124,7 @@ export function* activatePlugins( plugins ) {
 		} );
 
 		if ( results && results.status === 'success' ) {
-			yield updateActivePlugins( results.activatedPlugins );
+			yield updateActivePlugins( null, results.activatedPlugins );
```
psealock, line 127: `updateActivePlugins` is also used in `with-plugins-hydration.js` and I wonder if active/added isn't clear. How about a `replace` flag that defaults to false? Something like that so the default is using `updateActivePlugins` to add items, but you can also replace it wholesale. Would that make more sense?

joshuatf: I think you're right.  A `replace` flag makes it more clear.

Updated `updateActivePlugins` and `updateInstalledPlugins` to use this pattern.

joshuatf:

I wasn't able to recreate the errors, but might be worth trying a `composer install` and/or `composer dump-autoload` in the `woocommerce/` directory.  Not sure if those paths were updated recently or not.

joshuatf:
Updated testing instructions on this to test a bug that prevented skipping the plugins install.  See - https://github.com/woocommerce/woocommerce-admin/pull/4392

becdetat:
I used a jurassic.ninja site with WC 4.1.0 and didn't have any problems, everything worked as described.

The code looks good.

psealock:
Cherry picked into `release/1.2.0` branch

PR #4377, opened 2020-05-14
Fixed PHP notice about undefined index of items

### LINKED ISSUE #4280: [4.1 RC]: notices are logged using analytics
I’ve upgraded on a staging site wc 4.01 to 4.1 with no problems, but i’m getting notices on my logs(lots of them) when clicking any sub-menu under analytics(orders, categories, coupons, etc.)
The analytics info is displayed fine, but each time i’m using it(only analytics) a notice like this is logged:

[01-May-2020 17:08:46 UTC] PHP Notice: Undefined index: items in /xxx.xxx/wp-includes/rest-api.php on line 1245

I’m using wordpress 5.4.1, woocommerce release candidate 2 and php 7.3.16. This only happens on rc2, wc 4.01 with exactly the same configuration on production site doesn’t make those notices appear when using analytics.
Everything else works great! Thanks for making woocommerce such a great work!

**Expected behavior**
The notices shouldn't appear anymore
- [ x] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [ x] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/storefront/).
- [ x] I can reproduce this bug consistently using the steps above.

**WordPress Environment**
Wp 5.4.1 wc 4.1 rc2 php 7.3.16 after upgrading from wc 4.01

claudiosanches:
It fixes a notice caused by the missing 'items' param required for 'array' types.

> [14-May-2020 18:15:39 UTC] PHP Notice:  Undefined index: items in /var/www/woo/wp-includes/rest-api.php on line 1246

Fixes #4280

- Turn on your PHP logger, you can enable the WP logger by introducing the code bellow to your `wp-config.php`:
```php
define( 'WP_DEBUG', true );
define( 'WP_DEBUG_LOG', true );
define( 'WP_DEBUG_DISPLAY', true );
```
- Go to the WC Admin Dashboard in `/wp-admin/admin.php?page=wc-admin`
- Now check the `wp-content/debug.log`, you should see messages like:
> PHP Notice:  Undefined index: items in /var/www/woo/wp-includes/rest-api.php on line 1246
- Finally apply this patch and reload the dashboard page, check that all notices are gone now.

Note: Tested with WooCommerce 4.1.x and 4.2 beta 1.

### Changelog Note:

Fix: REST API collections schema.

### DIFF
--- a/src/API/Reports/Coupons/Stats/Controller.php
+++ b/src/API/Reports/Coupons/Stats/Controller.php
@@ -358,6 +358,9 @@ class Controller extends \WC_REST_Reports_Controller {
 			'type'              => 'array',
 			'sanitize_callback' => 'wp_parse_slug_list',
 			'validate_callback' => 'rest_validate_request_arg',
+			'items'             => array(
+				'type' => 'string',
+			),
 		);
 
 		return $params;
--- a/src/API/Reports/Customers/Stats/Controller.php
+++ b/src/API/Reports/Customers/Stats/Controller.php
@@ -368,6 +368,9 @@ class Controller extends \WC_REST_Reports_Controller {
 			'type'              => 'array',
 			'sanitize_callback' => 'wp_parse_slug_list',
 			'validate_callback' => 'rest_validate_request_arg',
+			'items'             => array(
+				'type' => 'string',
+			),
 		);
 
 		return $params;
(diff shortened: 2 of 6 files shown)


peterfabian:
Thanks @claudiosanches !

PR #4380, opened 2020-05-15
Connect to Jetpack on the benefits OBW page

becdetat:
This is based on #4374 but ported back to `release/1.1.1` as #4374 uses the new data store stuff that has been added to master since the 1.1.1 branch was cut.

- Deactivate Jetpack
- Go through the OBW
- Make sure you select Yes please on the benefits step
- It should activate Jetpack then redirect to the JP setup page
- Once you connect to JP it should redirect to the WC dashboard (if you're on a public server)

### DIFF
--- a/client/dashboard/components/connect/index.js
+++ b/client/dashboard/components/connect/index.js
@@ -1,0 +1,193 @@
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { Button } from '@wordpress/components';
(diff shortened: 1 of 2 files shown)


becdetat:
@joshuatf I saw that you fixed this in another PR where you cherry-picked my commit. I had the same issue when using the `task-list/tasks/steps/connect.js` component where the buttons would be weird. Since you've fixed it (thank you ;-) ) does this PR still need to be merged, or can I throw it out?

PR #4381, opened 2020-05-15
Separate Jetpack connection logic in shipping task

### LINKED ISSUE #4372: Task List: No Shipping connection step when Jetpack is active
**Describe the bug**

On a US store, when Jetpack is not active, the Shipping task includes a plugin installation/activation step for Jetpack and WooCommerce Services, as well as a Jetpack connection step.

However, if Jetpack _is_ already active, then no step is presented for connecting it [[code](https://github.com/woocommerce/woocommerce-admin/blob/742b532ea379b402f323d767b7356f8cfe99424d/client/dashboard/task-list/tasks/shipping/index.js#L282)] – unlike in the Tax task [[code](https://github.com/woocommerce/woocommerce-admin/blob/742b532ea379b402f323d767b7356f8cfe99424d/client/dashboard/task-list/tasks/tax.js#L347)].

(cc @pmcpinto re: p1HpG7-9c2-p2#comment-39069, though shipping labels isn't given as an example, so I can't be entirely certain whether it applies here.)

**To Reproduce**

1. Go through onboarding flow
2. Agree (at the end) to install and activate Jetpack and WooCommerce Services. (Note: As of 1.2.1, there is no connection flow here, but this is slated to change in https://github.com/woocommerce/woocommerce-admin/issues/4363)
3. "Set up shipping" in the task list
4. See only store location and shipping cost steps

**Expected behavior**

Since Jetpack can be connected in the case when it isn't yet active, I would expect to have the same opportunity to connect even if it's active already.

joshuatf:
Fixes #4372 

Fixes a regression where the Jetpack connection step was not shown if Jetpack was already active.

### Screenshots
Screen Shot 2020-05-15 at 3 34 35 PM

Test the following scenarios:
1. No Jetpack installed/activated - should show plugins and connect step should be shown after activating Jetpack.
1. Jetpack and WCS active but not connected - should show the connect step, but no plugins step.
1. Jetpack and WCS active and connected - should not show the connect or plugins steps.
1. Jetpack active and connected, but WCS not active - should install WCS but not show connect step.
1. Jetpack not connected or active in non-US store - should not install Jetpack or WCS and should not show connect step, but should install `woocommerce-shipstation-integration` if not active.

### DIFF
--- a/client/dashboard/task-list/tasks/shipping/index.js
+++ b/client/dashboard/task-list/tasks/shipping/index.js
@@ -145,23 +145,26 @@ class Shipping extends Component {
 	}
 
 	getPluginsToActivate() {
-		const { countryCode, isJetpackConnected } = this.props;
+		const { countryCode } = this.props;
 
 		const plugins = [];
 		if ( [ 'GB', 'CA', 'AU' ].includes( countryCode ) ) {
 			plugins.push( 'woocommerce-shipstation-integration' );
 		} else if ( countryCode === 'US' ) {
 			plugins.push( 'woocommerce-services' );
-
-			if ( ! isJetpackConnected ) {
-				plugins.push( 'jetpack' );
-			}
+			plugins.push( 'jetpack' );
 		}
 		return difference( plugins, this.activePlugins );
 	}
 
 	getSteps() {
+		const { activePlugins, countryCode, isJetpackConnected } = this.props;
 		const pluginsToActivate = this.getPluginsToActivate();
+		// This requirement may change dynamically, so don't use cached plugins array.
+		const requiresJetpackConnection =
+			! isJetpackConnected &&
+			countryCode === 'US' &&
+			activePlugins.includes( 'jetpack' );
 
 		const steps = [
 			{
@@ -197,7 +200,8 @@ class Shipping extends Component {
 				content: (
 					<ShippingRates
 						buttonText={
-							pluginsToActivate.length
+							pluginsToActivate.length ||
+							requiresJetpackConnection
 								? __( 'Proceed', 'woocommerce-admin' )
 								: __( 'Complete task', 'woocommerce-admin' )
 						}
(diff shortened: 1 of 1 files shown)


psealock:
Nice @joshuatf, here are the results of my testing:

>No Jetpack installed/activated - should show plugins and connect step should be shown after activating Jetpack.

❌ 

I click "Install and enable" to activate Jetpack

Screen Shot 2020-05-18 at 5 31 24 PM

Once I've done that, it exits the task. However, if I try again, I get the Connect step. I think thats a bug because I should see the added Connect task.

Screen Shot 2020-05-18 at 5 32 17 PM

Even here, if I click "Install and enable" I am not taken to the Connect step.

>Jetpack and WCS active but not connected - should show the connect step, but no plugins step.

✅ 

Screen Shot 2020-05-18 at 5 37 23 PM

>Jetpack and WCS active and connected

✅ 

Screen Shot 2020-05-18 at 5 29 39 PM

>Jetpack not connected or active in non-US store - should not install Jetpack or show connect step, but should install WCS ("Enable shipping label printing" step) if not active.

✅ 

WCS already active.

Screen Shot 2020-05-18 at 5 41 12 PM

❌ 

If WCS is not active, I'm not getting "should install WCS ("Enable shipping label printing" step)".

joshuatf:
Thanks for the thorough review, Paul!

>Once I've done that, it exits the task. However, if I try again, I get the Connect step. I think thats a bug because I should see the added Connect task.

I think I goofed on this one by creating a race condition to check for Jetpack activation.  Updated this to always show this step if in the US and if Jetpack is not connected.  If Jetpack plugin installation fails for some reason, they won't reach this step anyway so I believe this should be okay.

>If WCS is not active, I'm not getting "should install WCS ("Enable shipping label printing" step)".

I mis-worded the above and also wasn't clear enough in my testing instructions.  WCS should only show only be attempted in the US, while `woocommerce-shipstation-integration` should be installed in GB, AU, or CA.

I'll update the above testing instructions to reflect that.

PR #4392, opened 2020-05-18
Refactor plugin install flow pending state

joshuatf:
Fixes a bug that would not allow a user to skip the plugin install.

Also attempts to refactor the pending state to make it more manageable, but a long-term fix is needed so we have better visibility of the requests made in our store.

### Screenshots

Screen Shot 2020-05-18 at 7 52 19 PM

1. Deactivate and/or uninstall Jetpack and WCS.
1. Walk through the wizard attempting to install and connect via "Yes please!"
1. Make sure plugins are installed and you're redirect to the Jetpack connection flow.
1. Walk through the wizard attempt to skip by clicking "No thanks"
1. Make sure that you land in the task list without installing or connecting to Jetpack.
1. Re-enable the profiler and attempt these all steps again with Jetpack already connected.

### DIFF
--- a/client/dashboard/profile-wizard/steps/benefits/index.js
+++ b/client/dashboard/profile-wizard/steps/benefits/index.js
@@ -34,7 +34,7 @@ class Benefits extends Component {
 		this.state = {
 			isConnecting: false,
 			isInstalling: false,
-			isPending: false,
+			isActioned: false,
 		};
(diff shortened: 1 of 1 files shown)


jeffstieler:
Please change the base to `release/1.1.2` :smile:

Issue #4399, opened 2020-05-19
Jetpack 404 on connection


joshuatf:
**Describe the bug**
When attempting to walk through the profiler and redirect to the Jetpack connection flow, the connection flow occassionally gets stuck on the Jetpack screen with a `404` or other error.

Upon refresh, the error is resolved and the page is redirected to WordPress.com to finish the connection.

This seems to mostly happen on JN or ephemeral sites and may be a race condition due to installing Jetpack and immediately attempting to connect after installation is complete.  There may be a race condition occurring where Jetpack has not fully initialized and prevents the redirect to wpcom.

This may be a problem that needs to be resolved in Jetpack, but further investigation is required.a

**To Reproduce**

Note that you may need to do this quickly or on a slower server to reproduce.

1. On a new (non-local) site without Jetpack installed, walk through the onboarding profiler.
1. Click "Yes please" on the final step to redirect to the Jetpack flow.
1. Note that you land on the Jetpack screen with an error.

**Expected behavior**
Expected to get redirected directly to connect my site.

**Screenshots**
<img width="670" alt="Screen Shot 2020-05-19 at 8 10 47 PM" src="https://user-images.githubusercontent.com/10561050/82361217-3b68c700-9a13-11ea-9ccc-e34b70db2a4b.png">
<img width="865" alt="Screen Shot 2020-05-19 at 7 44 15 PM" src="https://user-images.githubusercontent.com/10561050/82361232-3d328a80-9a13-11ea-85f6-7bb04a808ab6.png">

becdetat:
In Jetpack the error `Verification secrets not found` happens in either https://github.com/Automattic/jetpack/blob/master/class.jetpack.php#L5402 or possibly https://github.com/Automattic/jetpack-connection/blob/master/src/class-manager.php#L1443

The cause of the error being generated is that the `jetpack_authorize_$user_id` option isn't present, which could be because the plugin isn't fully activated before hitting the connect step, but I can't see how that would be possible.

I also couldn't reproduce this on an ephemeral site.

timmyc:
Since we haven't seen further reports of this, I'm going to close this out. Side-note, wondering if this may be related to Jurassic Ninja reloading itself.

Issue #3470, opened 2019-12-23
Onboarding: Failure to Activate Jetpack


timmyc:
**Describe the bug**
When testing onboarding on a new Jurassic Ninja site against Woo 3.9 beta 2, I'm hitting the following error on the Activate Jetpack step:

![image](https://user-images.githubusercontent.com/22080/71373545-6d612480-256c-11ea-9fac-eefa6eea1b08.png)

Shown in Logs:

```
[23-Dec-2019 18:08:31 UTC] PHP Warning:  require_once(/srv/users/usera2c89776/apps/usera2c89776/public/wp-content/plugins/jetpack/load-jetpack.php): failed to open stream: No such file or directory in /srv/users/usera2c89776/apps/usera2c89776/public/wp-content/plugins/jetpack/jetpack.php on line 146
[23-Dec-2019 18:08:31 UTC] PHP Fatal error:  require_once(): Failed opening required '/srv/users/usera2c89776/apps/usera2c89776/public/wp-content/plugins/jetpack/load-jetpack.php' (include_path='.:/opt/sp/php7.3/lib/php') in /srv/users/usera2c89776/apps/usera2c89776/public/wp-content/plugins/jetpack/jetpack.php on line 146
```

**To Reproduce**

1. Setup a new test site on jurassic.ninja
2. Switch to 3.9.0 beta 2 of Woo
3. Add yourself to the a/b group `update_option( 'woocommerce_setup_ab_wc_admin_onboarding', 'b' );`
4. Go through the onboarding flow and encounter issues in the Jetpack/WCS step

I was able to get things working by deleting Jetpack and WCS from the plugin admin screen, and then all moved along as expected.

joshuatf:
I was not able to recreate the issue using a fresh JN install and [this RC](https://github.com/woocommerce/woocommerce/releases/tag/3.9.0-rc.2).

Do we think this issue is intermittent like https://github.com/woocommerce/woocommerce-admin/issues/3446 or do you hit this error every time, @timmyc?

timmyc:
Just noting that we chatted about this in a zoom and we are going to keep this open and see if any other reports trickle in post 3.9. I have yet to this bug in any other environment outside of jurassic.ninja

timmyc:
I'll be interested to see how many reports of this pop up during 4.0 testing. If not, I say we close this out.

pmcpinto:
[...]

PR #4411, opened 2020-05-20
Handle multiple plugin installs in a single request

### LINKED ISSUE #3446: Onboarding - store profiler: error when trying to install Facebook and Mailchimp extensions
I'm testing the new onboarding with 3.9.0-beta.2 and using WP 5.3.2 on a Pressable website and in the Business Details step I'm getting an error when I try to move forward with the Facebook and Mailchimp install:

Captura de ecrã 2019-12-19, às 10 19 34

I'm seeing the following error in the console: Failed to load resource: the server responded with a status of 500 ()

It's weird because if I go to the plugins page I see that both extensions were installed.

I'm also getting a weird styling in the "continue without installing" link: https://d.pr/i/QmvdwK

joshuatf:
Fixes #3446 
~~Fixes (possibly) #3470~~
~~Fixes (possibly) #4399~~

Updates the plugins endpoint to allow multiple plugins to prevent failure when many plugins needed to be installed.

Previously, many simultaneous requests were made to install plugins which was either causing a timeout or causing the WordPress upgrader cache to be cleared mid-install.  This PR forces them to install synchronously.

Updates:
* Allows multiple plugins to be specified for install.
* Adds more specific error messages for failures shared between server and client.
* Updates the API response to better match WP standards.

### Screenshots
Screen Shot 2020-05-20 at 8 31 53 PM

1. Delete `woocommerce-services`, `jetpack`, `facebook-for-woocommerce`, `mailchimp-for-woocommerce`, and `kliken-marketing-for-google` from your plugins directory.
1. Visit the business details step.
1. Continue, installing all plugins.
1. Make sure all our installed without error.
1. Continue through to Jetpack setup on the benefits step, installing plugins and attempting to connect.
1. Make sure the connection works as expected.
1. Test this locally, on JN sites, ephemeral, and other hosts if possible.
1. Optionally: intentionally force an error to make sure errors work as expected.  E.g.,

`if ( is_wp_error( $api ) ) {` -> `if ( is_wp_error( $api ) || true ) {`

### DIFF
--- a/packages/components/src/plugins/index.js
+++ b/packages/components/src/plugins/index.js
@@ -76,6 +76,7 @@ export class Plugins extends Component {
 		Object.keys( errors ).forEach( ( plugin ) => {
 			createNotice(
 				'error',
+				// Replace the slug with a plugin name if a constant exists.
 				pluginNames[ plugin ]
 					? errors[ plugin ][ 0 ].replace(
 							`\`${ plugin }\``,
@@ -108,12 +109,12 @@ export class Plugins extends Component {
 
 	render() {
 		const {
-			hasErrors,
 			isRequesting,
 			skipText,
 			autoInstall,
 			pluginSlugs,
 		} = this.props;
+		const { hasErrors } = this.state;
 
 		if ( hasErrors ) {
 			return (
--- a/packages/components/src/plugins/index.js
+++ b/packages/components/src/plugins/index.js
@@ -12,6 +12,7 @@ import { withSelect, withDispatch } from '@wordpress/data';
  * WooCommerce dependencies
  */
 import { PLUGINS_STORE_NAME } from '@woocommerce/data';
+import { pluginNames } from 'wc-api/onboarding/constants';
 
 export class Plugins extends Component {
 	constructor() {
@@ -54,26 +55,31 @@ export class Plugins extends Component {
 
 		const installs = await installPlugins( pluginSlugs );
 
-		if ( installs.errors && Object.keys( installs.errors ).length ) {
-			this.handleErrors( installs.errors );
+		if ( installs.errors && Object.keys( installs.errors.errors ).length ) {
+			this.handleErrors( installs.errors.errors );
 			return;
 		}
 
 		const activations = await activatePlugins( pluginSlugs );
 
-		if ( activations.status === 'success' ) {
+		if ( activations.success ) {
 			this.handleSuccess( activations.activePlugins );
 			return;
 		}
 
-		this.handleErrors( activations );
+		this.handleErrors( activations.errors.errors );
 	}
 
 	handleErrors( errors ) {
 		const { onError, createNotice } = this.props;
 
-		Object.keys( errors ).forEach( ( error ) => {
-			createNotice( 'error', errors[ error ] );
+		Object.keys( errors ).forEach( ( plugin ) => {
+			createNotice(
+				'error',
+				pluginNames[ plugin ]
+					? errors[ plugin ][0].replace( `\`${plugin}\``, pluginNames[ plugin ] )
+					: errors[ plugin ][0]
+			);
 		} );
 
 		this.setState( { hasErrors: true } );
--- a/packages/components/src/plugins/index.js
+++ b/packages/components/src/plugins/index.js
@@ -62,8 +62,8 @@ export class Plugins extends Component {
 
 		const activations = await activatePlugins( pluginSlugs );
 
-		if ( activations.success ) {
-			this.handleSuccess( activations.activePlugins );
+		if ( activations.success && activations.data.activated ) {
+			this.handleSuccess( activations.data.activated );
 			return;
 		}
 
@@ -77,8 +77,11 @@ export class Plugins extends Component {
 			createNotice(
 				'error',
 				pluginNames[ plugin ]
-					? errors[ plugin ][0].replace( `\`${plugin}\``, pluginNames[ plugin ] )
-					: errors[ plugin ][0]
+					? errors[ plugin ][ 0 ].replace(
+							`\`${ plugin }\``,
+							pluginNames[ plugin ]
+					  )
+					: errors[ plugin ][ 0 ]
 			);
 		} );
 
--- a/packages/components/src/plugins/index.js
+++ b/packages/components/src/plugins/index.js
@@ -42,7 +42,7 @@ export class Plugins extends Component {
 
 		const {
 			isRequesting,
-			installPlugin,
+			installPlugins,
 			activatePlugins,
 			pluginSlugs,
 		} = this.props;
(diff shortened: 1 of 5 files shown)


joshuatf:
Noting on here that I did get hit with a Mailchimp redirect when attempting hitting the Jetpack redirect, but unrelated to this PR and will tackle in a follow-up if we're able to reproduce.

packages/components/src/plugins/index.js
```diff
@@ -52,36 +53,36 @@ export class Plugins extends Component {
...
+				pluginNames[ plugin ]
+					? errors[ plugin ][ 0 ].replace(
+							`\`${ plugin }\``,
+							pluginNames[ plugin ]
+					  )
+					: errors[ plugin ][ 0 ]
+			);
 		} );
 
 		this.setState( { hasErrors: true } );
```
psealock, line 88: I noticed a bug here, unrelated to these changes: In the render function, `hasErrors` is gathered from props, not state, so the "Retry" or "Continue without installing" options never get displayed when errors occur.

packages/components/src/plugins/index.js
```diff
@@ -52,36 +53,36 @@ export class Plugins extends Component {
...
-			createNotice( 'error', error );
+		Object.keys( errors ).forEach( ( plugin ) => {
+			createNotice(
+				'error',
+				pluginNames[ plugin ]
+					? errors[ plugin ][ 0 ].replace(
+							`\`${ plugin }\``,
+							pluginNames[ plugin ]
+					  )
+					: errors[ plugin ][ 0 ]
```
psealock, line 84: A comment saying you're replacing a slug with a name, if it exists, would be helpful.

packages/components/src/plugins/index.js
joshuatf: Great catch.   Went ahead and updated in https://github.com/woocommerce/woocommerce-admin/pull/4411/commits/9d24050f07eee59955444fd654c82811b21af6a6.

src/API/Plugins.php
```diff
@@ -201,96 +201,115 @@ private function create_install_plugin_error_inbox_notification_for_jetpack_inst
...
 
 		return array(
-			'slug'   => $slug,
-			'name'   => $api->name,
-			'status' => 'success',
+			'data'    => array(
+				'installed' => $installed_plugins,
+				'results'   => $results,
+			),
+			'errors'  => $errors,
```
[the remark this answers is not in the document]
joshuatf: My goal here was to keep things consistent formatting between various API endpoints and as close to the WP format as possible.

While this one does not make use of the error data or other properties, I think it's possible other endpoints may so we should keep this as is.  I'm hoping to convert more endpoints to this format long-term so our API responses are more consistent.

joshuatf:
>I tested locally and on Ephemeral. Unfortunately, I still got the 404 error on Jetpack activation described in #4399.

Bummer.  Was really hoping this would resolve all of those issues, but that one may need to be handled outside of wc-admin.

I'll keep those other 2 issues as not fixed by this PR.

franticpsyx:
Hey folks, a bit late to the party! Really appreciate all the awesome work you are putting in here. Please be a bit more formal with things like this: https://github.com/woocommerce/woocommerce-admin/commit/bd394fd8bb4597813f47830f568324c43fb9f205#diff-21aca8e87b625c206696f3fe25c0162dR209

This project is seen by kittens that get really sad when public functions are renamed.

joshuatf:
Hey @franticpsyx - thanks for the feedback!

These APIs are namespaced to `wc-admin` and aren't meant for external use yet.  Breaking changes will occur and I would not recommend any direct use of these APIs yet (I would also not recommend making any direct calls to a callback function for a route handler).

That said, if you have a feature you'd like to include or need some level of extensibility here, please feel free to open an issue and we can discuss the best way to add it in. 😄

franticpsyx:
Hey @joshuatf thanks so much! 

> I would also not recommend making any direct calls to a callback function for a route handler

Definitely understood, please have a look at https://github.com/woocommerce/woocommerce-admin/issues/4709 as the `/Plugins` API is used this way inside wc admin (WooCommerce Payments install-plugin action).

> aren't meant for external use yet

Is there any document that describes which parts of `wc-admin` are meant for external use? I'm not sure if namespacing is a good signal that a private scope is in place. For example, [this](https://woocommerce.wordpress.com/2020/02/20/extending-wc-admin-reports/) official document about extending WC-Admin reports makes use of namespaced classes. 

A more specific question -- is the **Inbox** supposed to be used by third parties? I hope it is, as we're already using it in Product Bundles. If not, I'd love to see an announcement or document that officially discourages third parties from using all features / code in `wc-admin`, as it's practically all namespaced.

The new Inbox a system that has the potential to clean up the notices area (assuming that those who pollute it want to use a cleaner alternative), so opening it up to everyone would make a lot of sense.

joshuatf:
>Definitely understood, please have a look at #4709 as the /Plugins API is used this way inside wc admin (WooCommerce Payments install-plugin action).

Great point here.  I think ideally the call inside the install plugin action should make an actual dispatch using a `WP_REST_Request` instead of directly calling `install_plugin`.  This could lead to undesired side effects (like the breaking we're seeing now) and circumvents checks on the request, such as permission handling that's handled in the API.

@timmyc opened a PR to address the renamed function so we can take any additional discussion around that to that PR.

I'll make note of this on #4709.  Thanks again for catching that!

>Is there any document that describes which parts of wc-admin are meant for external use? I'm not sure if namespacing is a good signal that a private scope is in place.

To clarify what I mean by "external" use, I'm referring to external plugins or libraries consuming the APIs outside of the context of WC Admin.  As it is right now, it's definitely encouraged to extend those APIs and add any information you'd like to use in reports.

We moved away from versioned APIs and the generic WC namespace previously (`/wp-json/wc-analytics/...` vs `/wp-json/wc/v3`) and this was meant to signify that these are not for use outside of the context of analytics, but I completely agree that this is not 100% clear and could use documentation as to what's expected to change and what can be extended/consumed.

For endpoints like those in the plugins API, I believe the long-term goal here is to version them and put them under the core namespace (/cc @timmyc for your thoughts on this).  But I do think there are still a couple areas that need to be updated before that can happen.

>A more specific question -- is the Inbox supposed to be used by third parties? I hope it is, as we're already using it in Product Bundles.

The inbox is definitely meant to be consumed by third parties 😄

>The new Inbox introduces a system that has the potential to clean up the notices area (assuming that those who pollute it want to use a cleaner alternative), so opening it up to everyone would make a lot of sense.

Really glad to hear you say that.  This is one of the goals of the inbox and we're hoping to see plugin developers use this and share any feedback or additional features they want to add in here.

There's been some discussion on potential limitations to prevent this from being abused or a dumping ground that overwhelms storeowners, but currently there are no limitations on plugins adding notes.

franticpsyx:
Thanks for the detailed response @joshuatf !

> There's been some discussion on potential limitations to prevent this from being abused or a dumping ground that overwhelms storeowners, but currently there are no limitations on plugins adding notes.

💯 

So happy to know!

The new "Dismiss" feature is a great milestone on the way there! Eventually it would be awesome if merchants could silence messages by "source". I've seen notes have a source prop -- so assuming that developers will respect it, WooCommerce is only one feature away from offering a very solid messaging system 👍

PR #4414, opened 2020-05-21
Homepage: Style SummaryNumbers

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/4084
Depends on https://github.com/woocommerce/woocommerce-admin/pull/4427

Styles the new SummaryNumbers shown on the Home screen as well as Overview and reports.

### Accessibility

### Screenshots

Screen Shot 2020-05-26 at 3 20 49 PM

Screen Shot 2020-05-26 at 3 23 58 PM

Screen Shot 2020-05-26 at 3 24 19 PM

Screen Shot 2020-05-26 at 3 23 24 PM

1. Visit Home screen, Reports, and Overview.
2. See the SummaryNumbers against various screen sizes and date ranges.
3. Ensure styles match designs in pbIJXs-4Z-p2


packages/components/src/summary/number.js
```diff
@@ -1,13 +1,16 @@
 /**
  * External dependencies
  */
-import { __, sprintf } from '@wordpress/i18n';
-import { Button } from '@wordpress/components';
+import {
+	Button,
+	Tooltip,
+	__experimentalText as Text,
```
jeffstieler, line 7: TIL! We might be using this tactic for the `TaskList` and `QuickLinks` as well (really the `Card` component I guess)

Does this change to `Text` after a certain version of `@wordpress/components`?

psealock: >Does this change to Text after a certain version of @wordpress/components?

Thats the idea. I like this component because design now includes the `variant` prop: "title:small", which makes creating text really nice.

psealock:
>Are the summary number titles meant to be purple?

Something is strange about our use of `$theme-color: theme( button );` The quick links use the exact same variable, but turn out blue 😕 . I will look into it.

jeffstieler:
> Something is strange about our use of $theme-color: theme( button ); The quick links use the exact same variable, but turn out blue 😕 . I will look into it.

Just remembered - I ran into similar issues starting to refactor our `Card` component to use the new(er) WP components - we have two `postcss` configs! There's one in the project root used for the app, and one in `bin/packages` used for our packages.

psealock:
@jeffstieler, thanks for the tip. In https://github.com/woocommerce/woocommerce-admin/pull/4414/commits/fab18f4f652e854fbda35e720939b4fce8d663a9 I fixed the postCSS issue. In a follow up, I'd love to investigate if there is a better way and hopefully remove our postCSS altogether. Historically, it was added so that webpack could take Gutenberg component styles and add in our purple (now changed to studio-pink) and create new stylesheets with those colors. We could save a lot of bytes by doing it differently, especially since our direction is headed towards using base Gutenberg components as they are.

packages/components/src/summary/placeholder.js
```diff
@@ -20,17 +20,13 @@ export const SummaryNumberPlaceholder = ( { className } ) => (
...
-			<span className="woocommerce-summary__item-data">
-				<span className="woocommerce-summary__item-value" />
-				<div className="woocommerce-summary__item-delta">
-					<span className="woocommerce-summary__item-delta-value" />
-				</div>
-			</span>
-			<span className="woocommerce-summary__item-prev-label" />
-			<span className="woocommerce-summary__item-prev-value" />
-		</span>
+		<div className="woocommerce-summary__item">
```
timmyc, line 23: Nice, thanks for updating the placeholder too.

webpack.config.js
```diff
@@ -172,12 +172,6 @@ const webpackConfig = {
 								],
 							},
 							prependData:
-								'@import "node_modules/@wordpress/base-styles/_colors.scss"; ' +
```
timmyc, line 175: 🎉

jameskoster:
This is looking great!

> thinking that might need to be changed up at some point ( cc @jameskoster )

The styling of the tabs here should match the styling of the tabs used throughout Gutenberg:

tabs

If we can use the same component that's obviously a big win!

For consistency I think we can apply the same styling pattern to the Summary Numbers seeing as they are effectively acting as tabs on the report screens. So move the "active indicator" to the bottom of the "tab", and make it the same color ([`$theme_color`](https://github.com/WordPress/gutenberg/blob/4fe937c2188e0677b03e6eb5d2c54292705f6c45/packages/base-styles/_colors.scss#L34)) as the stock tabs.

psealock:
Thanks for the feedback @jameskoster. 

>For consistency I think we can apply the same styling pattern to the Summary Numbers

Cool, here are the SummaryNumbers on report pages.

Screen Shot 2020-06-08 at 1 18 25 PM

And once again, in the homescreen stats.

Screen Shot 2020-06-08 at 1 18 17 PM

>The styling of the tabs here should match the styling of the tabs used throughout Gutenberg

I'd like to handle this separately. I may do it as part of https://github.com/woocommerce/woocommerce-admin/pull/4427 which brought in those styles, or a new PR depending on how Gutenberg does it.

psealock:
Since the conversation is here, I went ahead an pushed date picker styles to this branch.

Screen Shot 2020-06-08 at 2 49 05 PM

Any tweaks or suggestions?

jameskoster:
Nice, looks good! Only comment is that we shouldn't be seeing the Purple color anywhere. If we use the [`$theme_color`](https://github.com/WordPress/gutenberg/blob/master/packages/base-styles/_colors.scss#L34) var then all the accent details in our UI should blend in with whatever color scheme the user has selected in their profile settings.

psealock:
>How come the purple is appearing in your screenshots @psealock ?

The "View detailed stats" link is `studio-pink" for visited links, is that you you're seeing as purple? The latest are [here](https://github.com/woocommerce/woocommerce-admin/pull/4414#issuecomment-640316242) and [here](https://github.com/woocommerce/woocommerce-admin/pull/4414#issuecomment-640316242). I  didn't update the original description.

jameskoster:
I'm referring to the "active" accent on the tabs, and the "Update" button in the datepicker screenshot you shared. They look like they're using the g2 purple/blue color? That particular color shouldn't actually be implemented anywhere, it basically just serves as a placeholder for the `$theme-color` var, which adapts depending on the user-selected color scheme.

I suppose that if we can't access $theme-color, then we should just hardcode whatever the default value for that is. But if your follow-up will make it accessible, you can ignore this :)

psealock:
I have it hard coded to `$theme-color: #3858e9;` for now. I will address this in the follow up and make it accessible based on the color scheme.

psealock:
>Which for the life of me I could not figure out why that would be failing here.

Thanks TImmy, thats a known bug that is fixed on master.

Issue #4383, opened 2020-05-15
Update `@wordpress/components` and `@wordpress/base-styles` to latest versions


mattsherman:
**Is your feature request related to a problem? Please describe.**

When working on #4350, I wanted to use 'G2' variables defined in `@wordpress/base-styles`. However, the version we are using is older and doesn't include those variables. So, I looked to update the version of `@wordpress/base-styles` that we are dependent on. But, that required a newer version of `@wordpress/components`, and updating that would introduce breaking changes that we would need to address (this was [also noticed by other devs](https://github.com/woocommerce/woocommerce-admin/pull/4131#issuecomment-619957580)). So, I decided to punt and just define the values myself directly in my SCSS. Not ideal, but less risky at the present.

Breaking changes introduced by updating `@wordpress/components` ([see `@wordpress/components` CHANGELOG](https://github.com/WordPress/gutenberg/blob/master/packages/components/CHANGELOG.md) for more details):
- `Button` changes
    - `is-button` classname removed
    - `is-default` classname not applied automatically
    - By default, Buttons come with a fixed height and hover styles
    - `isDefault` prop deprecated
    - `IconButton` deprecated

**Describe the solution you'd like**

We should update the version of `@wordpress/components` and `@wordpress/base-styles` so that we are able to make use of the latest and greatest from core WordPress in wc-admin. This will lead to greater consistency between wc-admin and WordPress, and lessen our tech debt over time.

**Should this be prioritized? Why?**

This should be handled in the next cooldown period, as it is part of general maintenance of our project and will help prevent tech debt from accumulating.

PR #4415, opened 2020-05-21
IconButton: Replace with Button

psealock:
Partial fix to https://github.com/woocommerce/woocommerce-admin/issues/4383

Remove use of `IconButton` in favor of just `Button` so we can update `@wordpress/components`.

See [`@wordpress/components` changelog](https://github.com/WordPress/gutenberg/blob/master/packages/components/CHANGELOG.md).

I also removed `SplitButton` component because it wasn't used anywhere.

### Accessibility

### Screenshots

Here are some quick screenshots of updated buttons

Screen Shot 2020-05-22 at 11 18 16 AM
Screen Shot 2020-05-22 at 11 30 09 AM
Screen Shot 2020-05-22 at 11 46 13 AM
Screen Shot 2020-05-22 at 11 54 39 AM
Screen Shot 2020-05-22 at 11 54 50 AM
Screen Shot 2020-05-22 at 11 54 58 AM
Screen Shot 2020-05-22 at 11 55 43 AM
Screen Shot 2020-05-22 at 11 55 48 AM
Screen Shot 2020-05-22 at 11 56 33 AM
Screen Shot 2020-05-22 at 11 58 44 AM
Screen Shot 2020-05-22 at 11 59 21 AM
Screen Shot 2020-05-22 at 11 59 43 AM

1. Click around the app, checking elements shown in the screenshot.
2. Double check that colors are correct. In particular the svg color should match the text.

### Changelog Note:

Dev: Remove use of `IconButton` in favor fo `Button`


mattsherman:
In Analytics / Orders, the X button for filter criteria is misaligned vertically:

Filter criteria X misalignment

psealock:
Thanks for the review @mattsherman !

I added the 8px of padding to the remove button on Advanced Filters like you suggested.

>Things looked okay to me, but then again, I'm color blind. So, I'll leave the color correctness determination to another set of eyes!

I think we should be good here with the rule `.components-button svg { fill: currentColor } `

mattsherman:
@psealock Oops... noticed you snuck the `jest` commit in while I was approving... I grabbed the latest and ran the following to verify the changes from that commit were good.

```
npm test -- -i packages/components/src/search-list-control/test/index.js
```

psealock:
Thanks for the late night review @mattsherman.

I'm in the middle of a multibranch rebase fiasco here. It looks like test fail. Its a jest snapshot error due to props like `isExpanded= { false }` v. `isExpanded= "false"`

psealock:
looks like reverting that last commit did the trick. Maybe it was a dependency thing happening.

mattsherman:
Hrm... but, after running the unit test manually again after your revert, the tests fail, as I would actually expect!

psealock:
So weird! Something is/was off with snapshots. If you were on a different branch and switched to this one without npm install'ing, that may be the issue.

In any case, I feel good about merging this one because tests pass on travis and oddness surrounds known issues.

PR #4418, opened 2020-05-21
add runtime feature flag filter

### LINKED ISSUE #4417: Allow managed hosts to toggle features
The `woocommerce_admin_features` filter was added to allow enabling/disabling features. However, the client uses the `feature-config` generated during the build process to determine which features are enabled.

Allow feature enabling/disabling on page load.

**Additional context**
A use case for this is a managed host which has its own onboarding process and would want to disable the WC onboarding.

rrennick:
Fixes #4417

This PR introduces the `woocommerce_admin_features_to_enable_disable` filter which allows enabling or disabling a feature through `wcSettings`. If the feature being disabled has a menu item then the `woocommerce_admin_features` filter needs to be used to remove the menu item.

- Build the branch
- Add the filter 
```
add_filter( 'woocommerce_admin_features_to_enable_disable', function( $features ) {
	$features[ 'homepage' ] = true;
	return $features;
} );
```
- The home page should show on the `WooCommerce` menu item
- Change it to `false`
- The dashboard should show on the `WooCommerce` menu item
- Change the filter and add the second filter
```
add_filter( 'woocommerce_admin_features_to_enable_disable', function( $features ) {
	$features[ 'homepage' ] = true;
	$features[ 'marketing' ] = false;
	return $features;
} );
```
- The `Marketing` menu item should be removed
- Navigating to `?page=wc-admin&path=marketing` should render an empty screen

### Changelog Note:

Enhancement: Allow runtime feature toggling.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -121,12 +121,27 @@ class Loader {
 	}
 
 	/**
-	 * Gets an array of enabled WooCommerce Admin features/sections.
+	 * Gets a build configured array of enabled WooCommerce Admin features/sections.
 	 *
-	 * @return bool Enabled Woocommerce Admin features/sections.
+	 * @return array Enabled Woocommerce Admin features/sections.
 	 */
 	public static function get_features() {
 		return apply_filters( 'woocommerce_admin_features', array() );
+	}
+
+	/**
+	 * Gets a runtime array of enabled WooCommerce Admin features/sections.
+	 *
+	 * @return array Woocommerce Admin features/sections.
+	 */
+	protected static function get_enabled_features() {
+		$features_mask    = apply_filters( 'woocommerce_admin_features_to_enable_disable', array() );
+		$add_features     = array_filter( $features_mask );
+		$remove_features  = array_keys( array_diff_key( $features_mask, $add_features ) );
+
+		$enabled_features = self::get_features();
+		$enabled_features = array_diff( $enabled_features, $remove_features );
+		return array_merge( $enabled_features, array_keys( $add_features ) );
 	}
 
 	/**
@@ -162,14 +177,14 @@ class Loader {
 	 * @return bool Returns true if the feature is enabled.
 	 */
 	public static function is_feature_enabled( $feature ) {
-		$features = self::get_features();
+		$features = self::get_enabled_features();
 		return in_array( $feature, $features, true );
 	}
 
 	/**
 	 * Returns if the onboarding feature of WooCommerce Admin should be enabled.
 	 *
-	 * While we preform an a/b test of onboarding, the feature will be enabled within the plugin build, but only if the user recieved the test/opted in.
+	 * While we preform an a/b test of onboarding, the feature will be enabled within the plugin build, but only if the user received the test/opted in.
 	 *
 	 * @return bool Returns true if the onboarding is enabled.
 	 */
@@ -233,7 +248,7 @@ class Loader {
 	 * Class loader for enabled WooCommerce Admin features/sections.
 	 */
 	public static function load_features() {
-		$features = self::get_features();
+		$features = self::get_enabled_features();
 		foreach ( $features as $feature ) {
 			$feature = str_replace( '-', '', ucwords( strtolower( $feature ), '-' ) );
 			$feature = 'Automattic\\WooCommerce\\Admin\\Features\\' . $feature;
@@ -681,7 +696,7 @@ class Loader {
 			$classes[] = 'woocommerce-admin-is-loading';
 		}
 
-		$features = self::get_features();
+		$features = self::get_enabled_features();
 		foreach ( $features as $feature_key ) {
 			$classes[] = sanitize_html_class( 'woocommerce-feature-enabled-' . $feature_key );
 		}
--- a/src/Features/AnalyticsDashboard.php
+++ b/src/Features/AnalyticsDashboard.php
@@ -8,6 +8,8 @@
 
 namespace Automattic\WooCommerce\Admin\Features;
 
+use Automattic\WooCommerce\Admin\Loader;
+
 /**
  * Contains backend logic for the dashboard feature.
  */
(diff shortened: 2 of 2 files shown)


psealock:
Nice feature here @rrennick 

>If the feature being disabled has a menu item then the woocommerce_admin_features filter needs to be used to remove the menu item.

Is it not possible to use the first filter to disable menu items? Having two seems redundant.

packages/navigation/src/index.js
```diff
@@ -168,3 +168,15 @@ export function updateQueryString(
...
+ * Get a feature enabled flag.
+ *
+ * @param {string} feature name.
+ */
+export function getFeatureFlag( feature ) {
+	if ( window.wcSettings.wcAdminFeatures && window.wcSettings.wcAdminFeatures.hasOwnProperty( feature ) ) {
+		return window.wcSettings.wcAdminFeatures[ feature ];
+	}
+	return window.wcAdminFeatures[ feature ];
+}
```
psealock, line 182: A note explaining that this is overwriting `window.wcAdminFeatures` created by webpack for different hosting situations might be helpful in the future.

rrennick:
> Is it not possible to use the first filter to disable menu items? Having two seems redundant.

@psealock What if we applied the same masking logic in the menu rendering that is implemented in the client?

psealock:
> What if we applied the same masking logic in the menu rendering that is implemented in the client?

I think I know what you mean. Here in PHP land we use the enabled features to make a decision when registering a page. We can do the same for registering a menu item. 

https://github.com/woocommerce/woocommerce-admin/blob/714660f67ee03ba94bc0febf30d2acbfe22bca7d/src/Loader.php#L252-L264

Just wondering (I haven't thought too much about it), can just a single filter be applied to the output of `wc_admin_get_feature_config`? That function is what's used to populate `window.wcAdminFeatures`.

timmyc:
@rrennick could we get some updates on this PR so we can move this one forward? We need this for the ecomm plan.

rrennick:
> Just wondering (I haven't thought too much about it), can just a single filter be applied to the output of wc_admin_get_feature_config? 

@psealock Thanks for the fantastic suggestion. The updated PR adds the new filter which is applied throughout `Loader.php`.

@timmyc Note the `woocommerce_admin_features` was removed from the testing code. Only the `woocommerce_admin_features_to_enable_disable` filter is needed.

src/Loader.php
```diff
@@ -885,7 +900,6 @@ public static function add_component_settings( $settings ) {
 		}
 
 		$settings['allowMarketplaceSuggestions'] = WC_Marketplace_Suggestions::allow_suggestions();
-		$settings['connectNonce']                = wp_create_nonce( 'connect' );
```
rrennick, line 888: Is this something I missed in a `master` merge?

timmyc: Appears to have been added here https://github.com/woocommerce/woocommerce-admin/commit/3cdeb916efbd6ace710bb1dcb96fff6c69ee5540, and is still in `master`

rrennick: @timmyc Thanks, I restored the line of code.

psealock:
I can't get this to work.

```php
add_filter( 'woocommerce_admin_features_to_enable_disable', function( $features ) {
    // doesn't work
    return array( 'devdocs' );
    // also doesn't work
    return array(
        'devdocs' => false,
    );
} );
```
`window.wcAdminFeatures.devdocs` is always false.

Is this to disable the homescreen? If so, I created an option to disable it in https://github.com/woocommerce/woocommerce-admin/pull/4450.

timmyc:
> Is this to disable the homescreen? If so, I created an option to disable it in #4450.

@psealock this is so anyone can enable/disable features in wc-admin outside of the config feature flags that are using during build. The most immediate use-case for this is for us to enable features in the ecommerce plan that aren't toggled on in the woo core build.

rrennick:
@psealock The filter works with

```
add_filter( 'woocommerce_admin_features_to_enable_disable', function( $features ) {
    return array(
        'devdocs' => true, // enable
        'homepage' => false, // disable
    );
} );
```

psealock:
Thanks @rrennick 

I can toggle DevDocs with the filter, but the homepage is not responding. Maybe its because I accessed the feature config object directly instead of using `is_feature_enabled`?

https://github.com/woocommerce/woocommerce-admin/blob/3fa2850a5ca3fb9538e83dbb3d31758185276e18/src/Features/AnalyticsDashboard.php#L83

psealock:
@rrennick I rebased this and made some changes due to the introduction of the option to turn off homepage in https://github.com/woocommerce/woocommerce-admin/pull/4523.

The PHP side works after those changes, but the client does not because the configs are passed via webpack:

https://github.com/woocommerce/woocommerce-admin/blob/d88c171678a5cf0e849c34dd81c69bc3954fe9de/webpack.config.js#L211-L213

https://github.com/woocommerce/woocommerce-admin/pull/4523 starts to add it back in via PHP, but I ran out of time. I can continue tomorrow or feel free to take over.

rrennick:
> I wanted to create a PR onto this one, but the rebase made that messy so I created a new one.

@psealock Sounds good. You can take this over.

src/Loader.php
```diff
@@ -121,14 +121,29 @@ public static function is_dev() {
...
 
 	/**
-	 * Gets an array of enabled WooCommerce Admin features/sections.
+	 * Gets a build configured array of enabled WooCommerce Admin features/sections.
 	 *
-	 * @return bool Enabled Woocommerce Admin features/sections.
+	 * @return array Enabled Woocommerce Admin features/sections.
 	 */
 	public static function get_features() {
 		return apply_filters( 'woocommerce_admin_features', array() );
```
psealock, line 129: @rrennick What do you think about just using this existing filter instead of introducing a new one? You'd have to make sure it runs after ours that populates the data. Is that a deal breaker?

timmyc: @psealock in my experience trying to work with that filter, it would only work if done during a build. Though my memory could be foggy - but I believe that is the big difference between that, and @rrennick's approach here on a more run-time filter that can work in a built version of the plugin.

psealock:
Closing in favor of https://github.com/woocommerce/woocommerce-admin/pull/4523

src/Loader.php
rrennick: > in my experience trying to work with that filter, it would only work if done during a build.

After I added that filter I found it could only disable the Devdocs screen dynamically.

PR #4430, opened 2020-05-22
update return customer total to count in range

### LINKED ISSUE #4410: Orders table summary has incorrect totals
**Describe the bug**
I noticed this with WC 4.1.1 and updated the site to WC 4.2 RC see if the issue still existed. Initially what I noticed was that the returning customers total was low. Here's a MTD screenshot with the `Returning Customers` filter applied:

Screen Shot 2020-05-20 at 10 25 45 AM

**Steps to Reproduce**
- Create an order using new billing information
- Check the orders report for the day to verify it shows as a `New` customer
- Create a second order using the same billing information
- Check the orders report 
  - The table shows one `New` and one `Returning` order
  - The summary shows `1 new customer` and `0 returning customers`

The current calculation for `returning customers` deducts the count of new customers from the total customers. While it is technically correct that the report only contains new customers, it's a confusing UX as the table shows returning customers. 

In a YTD report a store might have more than half the orders made by returning customers who placed their first order in that year.

rrennick:
Fixes #4410 

The current calculation of the number of returning customers is the number of customers minus the number of new customers in the date range of the report. This calculates a reasonable total on a short range report (eg. week) but a misleading number on longer range reports (eg. last year).

Screenshots from my dev install using the `Last Year` filter

Current:
Screen Shot 2020-05-22 at 10 10 33 AM

With this PR:
Screen Shot 2020-05-22 at 10 11 06 AM

It will be more useful for this store owner to know that they have approximately 40% returning customers than 5% of the customers in this date range had previous orders.

This PR updates the returning customer count to be the count of unique returning customer IDs in the report date range.

- Create an order using new billing information
- Check the orders report for the day to verify it shows as a `New` customer
- Create a second order using the same billing information
- Check the orders report 
  - The table shows one `New` and one `Returning` order
  - The summary should show `1 new customer` and `1 returning customer`

### Changelog Note:

Fix: Update returning customer total to include customers whose first order was within the report date range.


jeffstieler:
Pinging @jameskoster as well (since David isn't on the project anymore)

joshuatf:
>I spoke about this with @LevinMedia and we concluded that since this is the orders report we should probably just list the customer name in the table row (rather than "new" or "returning"), and the total number of customers in the summary.

Thanks for the response here @jameskoster.  That sounds like a great way of handling it and simplifies the technical side of things a bit.

>When applying a customer type advanced filter on the order report, the only thing that seems to get updated are the table rows.

Thanks for noting this as well.  I can reproduce this as well.  Might be something we need to tackle in a follow-up to keep the scope on this PR down.

@rrennick Did you want me to take over this PR?

rrennick:
> Did you want me to take over this PR?

@joshuatf Please do :)

joshuatf:
Updated this PR so that we are now pulling just total customers for order stats and displaying the customer name in the orders table.  This is ready for review again.

### Screenshot
Screen Shot 2020-06-16 at 7 06 38 PM

### Testing

1. Visit the orders report.
1. Make sure names of customers are being shown in the orders table.
1. Make sure delete customers show as deleted and not an error.
1. Make sure segmenting by products works as expected.
1. Make sure all tests pass.

PR #4434, opened 2020-05-22
Update TaskList design

### LINKED ISSUE #4096: Onboarding Task List changes - home screen
**User story:**
As a new merchant
I want to access the setup tasks alongside other information
So I can explore different areas while keeping in mind that I can easily find the store setup tasks

store-setup

**Description:** our hypothesis is that we can increase the likelihood of task completion if we display it alongside useful information in the inbox panel. At the same time, we also want to make some small adjustments to the current task list: include a progress bar, an estimated time for each task and an easy way to access again the profiler info. 

**Acceptance criteria:**
- [ ] When the user finishes the store profiler he will be redirected to the home screen where he will see the task list, the inbox panel and the stats overview.
- [ ] We're going to have a new task for the Store profiler. If the user finishes the profiler it will be marked as completed. If the user doesn't finish or skips the profiler entirely it will be shown as incomplete.
- [ ] Each task will display estimated time for completion (see the mockup to get the time for each task)
- [ ] If the user wants to skip the task list, he can do it through an option in the ellipsis menu
- [ ] If the user completes the task list we close the task list and display the quick links (store management) card
- [ ] When the user skips the task list we'll display the quick links (store management) card
- [ ] We'll display a progress bar on the top of the task list that should be updated every time a user completes a task (nice to have)
- [ ] Remove the setup checklist from being displayed on the Analytics home screen - if the new Home screen feature is enabled.

**Event tracking:**
We already have all the event tracking covered, as long as all the events in the existing task list are covered in this change.

**Additional links:**
P2 with final design: pbIJXs-4Z-p2
Designs: https://www.figma.com/file/Z2PiNnkhjrUIiR7IrfJRnU/Home-Screen?node-id=970%3A0

jeffstieler:
Fixes #4096.

This PR seeks to update the appearance and functionality of the `<TaskList />` component by:

* Use checkmarks for list icons
* Add a progress bar
* Show time estimates for tasks
* Remove TaskList upon completion
* Don't show TaskList on Analytics Overview (when homepage feature enabled)

Another change made in this PR is to not use the `<TaskListPlaceholder />` when the options are being requested as it causes the placeholder to show in cases where the task list isn't displayed. This is an improvement for those cases, but perhaps makes the scenarios where the task list **is** displayed more jarring, as it "just appears" on fast connections (the placeholder is just for `Suspense` now).

I'm open to reverting that change for sure.

### Accessibility

### Screenshots

Screen Shot 2020-05-22 at 8 59 02 AM

@jameskoster does this look OK? ☝️ 

- Start a dev build (so homepage feature is enabled)
- Verify the design of the TaskList on the home screen
- Verify functions of all steps
- Verify "hide" functionality
- Clear "hidden" option
- Complete all tasks
- Verify task list auto-dismisses
- Reset task list completion options
- Start a prod build (so homepage feature is disabled)
- Verify the design of the TaskList on the home screen
- Verify functions of all steps
- Verify "hide" functionality

### Changelog Note:

Tweak: update design of store setup task list.

### DIFF
--- a/client/dashboard/customizable.js
+++ b/client/dashboard/customizable.js
@@ -311,7 +311,7 @@ class CustomizableDashboard extends Component {
 			<Fragment>
 				{ isTaskListEnabled && (
 					<Suspense fallback={ <Spinner /> }>
-						<TaskList query={ query } inline={ isDashboardShown } />
+						<TaskList query={ query } />
 					</Suspense>
 				) }
 				{ isDashboardShown && this.renderDashboardReports() }
--- a/client/dashboard/customizable.js
+++ b/client/dashboard/customizable.js
@@ -302,7 +302,10 @@ class CustomizableDashboard extends Component {
 	render() {
 		const { query, taskListHidden, taskListComplete } = this.props;
 
-		const isTaskListEnabled = isOnboardingEnabled() && ! taskListHidden;
+		const isTaskListEnabled =
+			isOnboardingEnabled() &&
+			! taskListHidden &&
+			! window.wcAdminFeatures.homepage;
 
 		const isDashboardShown =
 			! isTaskListEnabled || ( ! query.task && taskListComplete );
--- a/client/homepage/layout.js
+++ b/client/homepage/layout.js
@@ -16,11 +16,6 @@ import { get } from 'lodash';
 import PropTypes from 'prop-types';
 
 /**
- * WooCommerce dependencies
- */
-import { Spinner } from '@woocommerce/components';
-
-/**
  * Internal dependencies
  */
 import QuickLinks from '../quick-links';
--- a/client/homepage/layout.js
+++ b/client/homepage/layout.js
@@ -57,8 +57,8 @@ export const Layout = ( props ) => {
 		};
 	}, [] );
 
-	const { query, requestingTaskList, taskListHidden } = props;
-	const isTaskListEnabled = taskListHidden === false;
+	const { query, requestingTaskList, taskListComplete, taskListHidden } = props;
+	const isTaskListEnabled = taskListHidden === false && ! taskListComplete;
 	const isDashboardShown = ! isTaskListEnabled || ! query.task;
 
 	const renderColumns = () => {
@@ -131,6 +131,10 @@ Layout.propTypes = {
 	 */
 	requestingTaskList: PropTypes.bool.isRequired,
 	/**
+	 * If the task list has been completed.
+	 */
+	taskListComplete: PropTypes.bool,
+	/**
 	 * If the task list is hidden.
 	 */
 	taskListHidden: PropTypes.bool,
@@ -149,13 +153,16 @@ export default compose(
 
 		if ( isOnboardingEnabled() ) {
 			const options = getOptions( [
+				'woocommerce_task_list_complete',
 				'woocommerce_task_list_hidden',
 			] );
 			
 			return {
 				requestingTaskList: isGetOptionsRequesting( [
+					'woocommerce_task_list_complete',
 					'woocommerce_task_list_hidden',
 				] ),
+				taskListComplete: Boolean( get( options, [ 'woocommerce_task_list_complete' ] ) ),
 				taskListHidden: get( options, [ 'woocommerce_task_list_hidden' ] ) === 'yes',
 			};
 		}
--- a/client/homepage/layout.js
+++ b/client/homepage/layout.js
@@ -109,6 +109,7 @@ export const Layout = ( props ) => {
 				<TaskList
 					query={ query }
 					inline
+					promptShown
 				/>
 			</Suspense>
 		);
--- a/client/homepage/layout.js
+++ b/client/homepage/layout.js
@@ -106,11 +106,7 @@ export const Layout = ( props ) => {
 
 		return (
 			<Suspense fallback={ <Spinner /> }>
-				<TaskList
-					query={ query }
-					inline
-					promptShown
-				/>
+				<TaskList query={ query } />
 			</Suspense>
 		);
 	};
--- a/client/homepage/layout.js
+++ b/client/homepage/layout.js
@@ -96,7 +96,7 @@ export const Layout = ( props ) => {
 
 	const renderTaskList = () => {
 		if ( requestingTaskList ) {
-			return;
+			return <TaskListPlaceholder />;
 		}
 
 		return (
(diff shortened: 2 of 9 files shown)


client/dashboard/customizable.js
```diff
@@ -302,7 +302,10 @@ class CustomizableDashboard extends Component {
 	render() {
 		const { query, taskListHidden, taskListComplete } = this.props;
 
-		const isTaskListEnabled = isOnboardingEnabled() && ! taskListHidden;
+		const isTaskListEnabled =
+			isOnboardingEnabled() &&
+			! taskListHidden &&
+			! window.wcAdminFeatures.homepage;
```
rrennick, line 308: @jeffstieler if #4418 is merged first this should be updated to `! getFeatureFlag( 'homepage' );`

jameskoster:
Opened a PR on this [here](https://github.com/woocommerce/woocommerce-admin/pull/4436).

A couple of small things I didn't get to:

* I noticed the card header size differs slightly from the quick links. Is it possible to use the [Text](https://wordpress.github.io/gutenberg/?path=/story/components-experimental-text--default) component (title small) for these?
* On that note, are we using the Text component for all instances of text here? The sooner we can adopt that, the sooner our UI will start to feel more cohesive :)
* In the hide menu I'm seeing a [focus state](https://cloudup.com/cXqUPXEkr3s) that I probably shouldn't be upon open. Also, the menu item there should fill the width of the container. 

So cool to see this coming together - it's looking really good! :D

client/task-list/index.js
```diff
@@ -187,56 +198,20 @@ class TaskDashboard extends Component {
...
-						</MenuItem>
-					</div>
-				) }
-			/>
+			<div className="woocommerce-card__menu woocommerce-card__header-item">
+				<EllipsisMenu
+					label={ __( 'Task List Options', 'woocommerce-admin' ) }
+					renderContent={ () => (
+						<div className="woocommerce-task-card__section-controls">
+							<Button onClick={ () => this.hideTaskCard( 'remove_card' ) }>
```
psealock, line 208: Another unrelated comment: It always bugged me that clicking this didn't cause the task list to disappear immediately or at least show a spinner.

client/task-list/style.scss
```diff
@@ -589,20 +582,54 @@
...
+	border: 0;
+	border-top-left-radius: 2px;
+	border-top-right-radius: 2px;
+	height: 4px;
+	width: 100%;
+	vertical-align: top;
+
+	// Firefox
+	& {
+		background-color: #c4c4c4;
```
psealock, line 609: Do we have colors for these?

client/task-list/tasks.js
```diff
@@ -58,95 +58,79 @@ export function getAllTasks( {
...
 			title: __( 'Purchase & install extensions', 'woocommerce-admin' ),
-			content: __(
-				'Purchase, install, and manage your extensions directly from your dashboard',
-				'wooocommerce-admin'
-			),
-			icon: 'extension',
 			container: null,
 			onClick: () =>
 				remainingProductIds.length ? toggleCartModal() : null,
 			visible: productIds.length,
```
psealock, line 80: Clicking "Purchase & install extensions" doesn't do anything for me. Maybe this should be `visible: remainingProductIds.length`?

client/task-list/tasks.js
```diff
@@ -58,95 +58,79 @@ export function getAllTasks( {
...
 	const tasks = [
+		{
+			key: 'store_details',
+			title: __( 'Store details', 'woocommerce-admin' ),
+			container: null,
+			onClick: () => {
+				window.location = getAdminLink(
+					'admin.php?page=wc-admin&reset_profiler=1'
+				);
+			},
```
psealock, line 69: Perhaps good for a follow up: We should use `<Link />` component with these list items. This `onClick` causes a page refresh where one isn't required. Clicking "Store details" should not refresh the page. Secondarily, it removes the "preview" address in the lower right hand side which is useful.

psealock (commented):
>I'm open to reverting that change for sure.

The task list looks weird just appearing. It looks like it depends on two options `woocommerce_task_list_complete` and `woocommerce_task_list_hidden`, but only the later is preloaded. Should we preload them both?

client/task-list/style.scss
jeffstieler: Resolved with Jay's PR.

client/task-list/tasks.js
jeffstieler: I think it's functioning correctly (this is unchanged in the PR) - is it showing complete? If you selected a plugin and have it installed, the step will display as completed.

client/task-list/index.js
jeffstieler: OK to handle this in a follow up?

client/task-list/tasks.js
jeffstieler: I'd definitely like to tackle this in a follow up - at a guess, this is likely to be more involved than changing to a link component since we're potentially going to need to reset the "completed" flag for the task list (and think through the impact to tracking of that completion).

jeffstieler:
> The task list looks weird just appearing. It looks like it depends on two options `woocommerce_task_list_complete` and `woocommerce_task_list_hidden`, but only the later is preloaded. Should we preload them both?

Reworked in 98b8aa9.

client/homepage/layout.js
```diff
@@ -133,6 +125,10 @@ Layout.propTypes = {
 	 * If the task list option is being fetched.
 	 */
 	requestingTaskList: PropTypes.bool.isRequired,
+	/**
+	 * If the task list has been completed.
+	 */
+	taskListComplete: PropTypes.bool,
```
psealock, line 131: Theres a proptype violation here, "Invalid prop `taskListComplete` of type `string` supplied to `Layout`, expected `boolean`." I found it when the task list wasn't visible.

client/task-list/tasks.js
psealock: Ok, maybe it because I'm doing odd things like `const isTaskListEnabled = true;` hard coded to get it to come up.

psealock (approved):
>Reworked in 98b8aa9.

It still makes a request, even though both options are preloaded. In https://a8c.slack.com/archives/C9K5T0XSP/p1590452504044500?thread_ts=1590435699.034400&cid=C9K5T0XSP we discovered that hydration isn't working as expected with lazy loaded components. Maybe thats happening here.

Everything is looking good to me, just a prop type violation and the rest to be addressed in follow ups.

client/homepage/layout.js
jeffstieler: I'm unable to reproduce this - can you provide steps?

PR #4436, opened 2020-05-22
Some minor color tweaks

jameskoster:
I ended up using a dark color for the progress bar in the end @jeffstieler as it will better inherit the theme colors.

I hid the time estimates on completed tasks as well, hopefully I did that correctly!

### DIFF
--- a/client/task-list/index.js
+++ b/client/task-list/index.js
@@ -302,15 +302,16 @@ class TaskDashboard extends Component {
 					{ task.completed && <Icon icon={ check } /> }
 				</div>
 			);
-			task.after = task.time ? (
-				! task.completed && (
+
+			if ( ! task.completed ) {
+				task.after = task.time ? (
 					<span className="woocommerce-task-estimated-time">
 						{ task.time }
 					</span>
-				)
-			) : (
-				<Icon icon={ chevronRight } />
-			);
+				) : (
+					<Icon icon={ chevronRight } />
+				);
+			}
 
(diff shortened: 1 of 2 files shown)


client/task-list/index.js
```diff
@@ -302,9 +302,14 @@ class TaskDashboard extends Component {
...
-			task.after = ( task.time
-				? <span className="woocommerce-task-estimated-time">{ task.time }</span>
-				: <Icon icon={ chevronRight } />
+			task.after = task.time ? (
+				! task.completed && (
+					<span className="woocommerce-task-estimated-time">
+						{ task.time }
+					</span>
+				)
+			) : (
```
joshuatf, line 311: Are the arrows supposed to show on completed tasks or only when the entire task list is completed?  If the former, we should update this to:

```suggestion
			task.after = task.time && ! task.completed ? (
				<span className="woocommerce-task-estimated-time">
					{ task.time }
				</span>
			) : (
```

Current for reference:
Screen Shot 2020-05-25 at 11 18 28 AM


And with the above suggestion:
Screen Shot 2020-05-25 at 11 27 20 AM

jeffstieler: It looks like the chevron is just a fallback for when there isn't a time estimate for the task.

The designs don't show a chevron on completed tasks:

Screen Shot 2020-05-25 at 9 15 29 AM


(Which means this will need an update to not show the chevron on completed tasks that don't have a time estimate)

PR #4437, opened 2020-05-25
Bind onClick event handler for marketing kb article click

jconroy:
Correctly bind the onclick event hander for the Marketing knowledge base article.

Post/post title is currently `undefined` and I don't think the event is triggering properly. I think we lost the bind as part of some refactoring.

- Pull branch
- Modify the `onPostClick()` to add some logging

```
	onPostClick( post, e ) {
		e.preventDefault();
		console.log( post );
		recordEvent( 'marketing_knowledge_article', { title: post.title } );
	}
```

- Build
- Go to Marketing Tab on test site
- Click kb article
- See post object in console e.g. https://d.pr/i/gyNEVo

### DIFF
--- a/client/marketing/overview/knowledge-base/index.js
+++ b/client/marketing/overview/knowledge-base/index.js
@@ -74,7 +74,7 @@ class KnowledgeBase extends Component {
 							className="woocommerce-marketing-knowledgebase-card__post"
 							href={ post.link }
 							key={ index }
-							onClick={ this.onPostClick( this, post ) }
+							onClick={ this.onPostClick.bind( this, post ) }
 							target="_blank"
 							rel="noopener noreferrer"
 						>


danielbitzer:
It would be nice to convert this to a functional component so binding class methods isn't required. Alternatively, rather than make the whole component functional we could declare `onPostClick` inside `getCurrentSlide`.

jconroy:
Sure that sounds like a great follow up. Does this PR fix the issue for the implementation?

danielbitzer:
Is there any way to test that the event is recorded without preventing the default click event? When I add a break point on that line it doesn't stop there. I'm just curious if something needs to be done since it's an external link 🤔

danielbitzer:
We might need to use this function: https://github.com/woocommerce/woocommerce-admin/blob/556e6551423fe24490f8f8f75ebb632336b61527/client/lib/tracks.js#L107-L121

danielbitzer:
Since we're opening the link in a new tab it probably doesn't matter. I tested this change and could see the post title is correctly added to the event data.

PR #4441, opened 2020-05-25
Added new notification to choose a niche

### LINKED ISSUE #4196: New notification: How to choose a niche for your online store
**Title:** How to choose a niche for your online store
**Content:**  Your niche defines the products and services you develop. It directs your marketing. It focuses your attention on specific problems facing your customers. It differentiates you from the competition. Learn more about the five guiding principles to define your niche.
**Segmentation:** 
AND
one of the following values on `wcadmin_storeprofiler_store_business_details_continue`
`product_number`: `0`
`revenue`: `none`
`revenue`: `up to 2500`
**Timing:** On plugin install
**Cadence:** Once
**CTA leads to:** https://woocommerce.com/posts/how-to-choose-a-niche-online-business/?utm_source=inbox
**CTA label:**  Learn more

octaedro:
Fixes #4196

This PR adds a new notification to choose a niche.

This notification should be visible for:
New users with `setup_client`: `false` on`woocommerce_onboarding_profile`
AND
one of the following values on `woocommerce_onboarding_profile`
`product_count`: `0`
`revenue`: `none`
`revenue`: `up to 2500`

To set these values it's necessary to:
- Go to the `Business Details` step (in the new onboarding).
screenshot-one wordpress test-2020 05 25-15_51_04

- Set `How many products do you plan to display?` to `I don't have any products yet.`
- Set `What's your current annual revenue?` to `$0.00 (I'm just getting started)` or `Up to $2,500.00`.

- Verify the note is created and appears.
- Verify the action button works properly.

### Changelog Note:
Enhancement: Added new notification to choose a niche.


src/Notes/WC_Admin_Notes_Choose_Niche.php
```diff
@@ -0,0 +1,72 @@
...
+		$onboarding_profile = get_option( 'woocommerce_onboarding_profile', array() );
+		if (
+			! isset( $onboarding_profile['setup_client'] ) ||
+			! isset( $onboarding_profile['product_count'] ) ||
+			! isset( $onboarding_profile['revenue'] )
+		) {
+			return;
+		}
+
+		// Make sure the client is not setup.
```
jeffstieler, line 41: This value is to denote that the person filling out the OBW is setting up the store for their customer/client (not that there is some "client" in the client/server sense that was set up).

jeffstieler:
I'm not sure this is intentional - but this note won't display if the user says they aren't selling on another platform. So we're only showing this to users that they **don't have any products** but **are selling on another platform**.

@pmcpinto - is the criteria for showing this note correct? At the very least I think we should also include users that say they are NOT selling elsewhere.

pmcpinto:
@jeffstieler @octaedro the notification should be displayed if they don't have any products OR are selling on another platform with revenue below 2.500.

The hypothesis is that if they don't even have any product or are barely generating any revenue, probably they don't have product/market fit and can benefit from having access to useful content about finding a market niche. 

Let me know if it's still not clear or if you think that something doesn't make sense :) Thanks.

src/Notes/WC_Admin_Notes_Choose_Niche.php
```diff
@@ -0,0 +1,72 @@
...
+
+	/**
+	 * Get the note.
+	 */
+	public static function get_note() {
+		$onboarding_profile = get_option( 'woocommerce_onboarding_profile', array() );
+		if (
+			! isset( $onboarding_profile['setup_client'] ) ||
+			! isset( $onboarding_profile['product_count'] ) ||
+			! isset( $onboarding_profile['revenue'] )
```
jeffstieler, line 36: This check prevents the "OR" logic that Pedro describes. We should be able to select "I don't have any products" AND "I'm not selling anywhere else" and see the notice.

octaedro: Nice catch! Thank you @jeffstieler.
This is solved in the commit 7056ecb

octaedro:
Hi @timmyc, you're right! I just did the rebase, could you give this PR another look?

PR #4444, opened 2020-05-27
Try: Collect async resolver calls

psealock:
@joshuatf I'm trying to implement https://github.com/woocommerce/woocommerce-admin/pull/4144#discussion_r427902088 and running into difficulties.

Any suggestions here?

I created a `getOption` selector and resolver to accept a single option name as an argument. I also made changes to `withOptionsHydration` to mark any preloaded options as resolved. Everything is working nicely.

I'm having trouble with the collection of requests you suggested. The function in the setTimeout needs to be a generator function (or maybe just `async`?) to respect keywords like `yield`.

### Directions

* Head to the homepage and see the undefined options. 
* Change the requests to preloaded options such as `woocommerce_task_list_do_this_later` and `woocommerce_task_list_hidden` and see the hydration happening correctly.

### DIFF
--- a/packages/data/src/options/resolvers.js
+++ b/packages/data/src/options/resolvers.js
@@ -23,3 +23,24 @@ export function* getOptionsWithRequest( names ) {
 		return error;
 	}
 }
+
+const optionsToRequest = [];
+const fetches = {};
+
+export function* getOption( name ) {
+	yield setIsRequesting( true );
+	optionsToRequest.push( name );
+
+	setTimeout( function* () {
+		const names = optionsToRequest.join(',');
+		const fetchPromise = fetches[ names ];
+		if ( fetchPromise ) {
+			const results = yield fetchPromise;
+			return results;
+		}
+
+		const url = WC_ADMIN_NAMESPACE + '/options?options=' + names;
+		fetches[ names ] = yield apiFetch( { path: url } );
+		return fetches[ names ];
+	}, 1 );
+}
--- a/packages/data/src/options/resolvers.js
+++ b/packages/data/src/options/resolvers.js
@@ -41,6 +41,7 @@ export function* getOption( name ) {
 
 		const url = WC_ADMIN_NAMESPACE + '/options?options=' + names;
 		fetches[ names ] = yield apiFetch( { path: url } );
+		yield receiveOptions( fetches[ names ] );
 		return fetches[ names ];
 	}, 1 );
 }
--- a/packages/data/src/options/resolvers.js
+++ b/packages/data/src/options/resolvers.js
@@ -32,13 +32,12 @@ export function* getOption( name ) {
 
 	const fetchInProgress = fetches[ names ];
 	if ( fetchInProgress ) {
-		return;
+		return yield fetches[ names ];
 	}
 
 	const url = WC_ADMIN_NAMESPACE + '/options?options=' + names;
-	fetches[ names ] = true;
-	const result = yield apiFetch( { path: url } );
-	yield receiveOptions( result );
+	fetches[ names ] = yield apiFetch( { path: url } );
+	yield receiveOptions( fetches[ names ] );
 
 	// Delete the fetch after to allow wp data to handle cache invalidation.
 	delete fetches[ names ];
--- a/packages/data/src/options/resolvers.js
+++ b/packages/data/src/options/resolvers.js
@@ -8,10 +8,9 @@ import { batchFetch } from './controls';
  * Internal dependencies
  */
 import { WC_ADMIN_NAMESPACE } from '../constants';
-import { receiveOptions, setIsRequesting, setRequestingError } from './actions';
+import { receiveOptions, setRequestingError } from './actions';
 
 export function* getOptionsWithRequest( names ) {
-	yield setIsRequesting( true );
 	const url = WC_ADMIN_NAMESPACE + '/options?options=' + names.join( ',' );
 
 	try {
@@ -29,6 +28,6 @@ export function* getOption( name ) {
 		const result = yield batchFetch( name );
 		yield receiveOptions( { [ name ]: result } );
 	} catch ( error ) {
-		yield setRequestingError( error, [ name ] );
+		yield setRequestingError( error, name );
 	}
 }
--- a/packages/data/src/options/resolvers.js
+++ b/packages/data/src/options/resolvers.js
@@ -26,7 +26,7 @@ export function* getOptionsWithRequest( names ) {
 export function* getOption( name ) {
 	try {
 		const result = yield batchFetch( name );
-		yield receiveOptions( { [ name ]: result } );
+		yield receiveOptions( result );
 	} catch ( error ) {
 		yield setRequestingError( error, name );
 	}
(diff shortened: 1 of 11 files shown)


psealock:
🙇 🧙‍♂️ Thanks for the help here!

Nice one here, that is a clever solution! Its working well and collecting names before making a request.

One issue I can see is the early return in the resolver if statement, `fetchInProgress === true`, will cause `isResolving` to be `false` on some of the options when requests are still actually being made.

In https://github.com/woocommerce/woocommerce-admin/pull/4444/commits/c01969fadd8702e31b6b29880418c0623d1f4930 I added some console.log's to see what was happening and modified the if statement to yield to the fetch in progress as well.

Before, `isResolving` was returning `false` too early, now the opposite is true. But thats arguably better because any logic waiting on `isResolving` won't get a `false` with no data.

What do you think? Is there a better way?

packages/data/src/options/resolvers.js
```diff
@@ -23,3 +23,23 @@ export function* getOptionsWithRequest( names ) {
...
+
+const fetches = {};
+
+export function* getOption( name ) {
+	yield setIsRequesting( name );
+	const names = yield getOptionsToRequest( name );
+
+	const fetchInProgress = fetches[ names ];
+	if ( fetchInProgress ) {
+		return;
```
joshuatf, line 35: Just a heads up that I added this back in since we need to return early here.

Returning `yield apiFetch( { path: url } );` will hit the options endpoint again for each request.

joshuatf:
>One issue I can see is the early return in the resolver if statement, fetchInProgress === true, will cause isResolving to be false on some of the options when requests are still actually being made.

Great eyes @psealock!  I didn't see that before.

I guess the resolution status is marked as complete once the resolver finishes going through the generator (which happens much faster in the option that gets returned early).

I looked at overriding the `isResolving` selector so we could manually call `finishResolution()`, but could not find a way to do this without breaking the baked in resolution state.  I think it may be best here to just handle in our own `isRequesting` selector which I added in https://github.com/woocommerce/woocommerce-admin/pull/4444/commits/6c19881ce424918c16b5e8cf835016d1d908de2f.  This marks the collected options as complete when `receiveOptions` is called.

psealock:
I added error handling in https://github.com/woocommerce/woocommerce-admin/pull/4444/commits/f24f6039e7fe3b78fd5323c4e3989c12b42b2d1f

>I think it may be best here to just handle in our own isRequesting selector

yeah, that works. It is a bit unfortunate to not have a uniform pattern across the dataStores. In fact, this brings us [back to the original problem](https://github.com/woocommerce/woocommerce-admin/pull/4144#discussion_r426356254) that spawned this investigation. A new selector was required to discern if options are resolving or not, and we tried this branch to avoid that problem.

Any opinion why one method is better than the other?

joshuatf:
> yeah, that works. It is a bit unfortunate to not have a uniform pattern across the dataStores.

I agree. This is not a dealbreaker for me in this PR but it would be great if there was a way we could override the built in resolution status. Maybe @nerrad knows a trick for doing this. 


> Any opinion why one method is better than the other?

I definitely think this branch is an improvement. The ability to individually get options, even from separate components, and only have one API request made is enough of an improvement in my book. Plus it simplifies the logic for checking if an option is requesting:


```js
// Previously 
updateOptions( { optionA: ..., optionB: ... } );
isRequesting( [ ‘optionA’, ‘optionB’ ] );  // true
isRequesting( [ ‘optionA’, ]);  // false
isRequesting( [ ‘optionB’ ]);  // false
isRequesting( [ ‘optionB’, ‘optionA’ ]);  // false

// Now
isRequesting( ‘optionA’ ); // true
isRequesting( ‘optionB’ ); // true
```

packages/data/src/options/resolvers.js
```diff
@@ -19,7 +19,32 @@ export function* getOptionsWithRequest( names ) {
...
+	
+	try {
+		const result = yield apiFetch( { path: url } );
+		yield receiveOptions( result );
+	} catch ( error ) {
+		yield setRequestingError( error, names );
+	}
+
+	// Delete the fetch after to allow wp data to handle cache invalidation.
+	delete fetches[ names ];
```
nerrad, line 49: This comment is a bit ambiguous. Are you talking about `wp.data` handling _resolver_ cache invalidation here? If true, then no this won't handle resolver cache invalidation.

packages/data/src/options/resolvers.js
```diff
@@ -19,7 +19,32 @@ export function* getOptionsWithRequest( names ) {
...
 }
+
+const fetches = {};
+
+export function* getOption( name ) {
+	yield setIsRequesting( name );
+	const names = yield getOptionsToRequest( name );
+
+	const fetchInProgress = fetches[ names ];
+	if ( fetchInProgress ) {
```
nerrad, line 34: You could probably just eliminate the declaration of `fetchInProgress` here and use `fetches[names]` directly in the condition.

packages/data/src/options/resolvers.js
nerrad: As Paul pointed out, this will cause the built in cached resolver for `getOption( name )` to be marked as finished - so this effectively means you cannot depend on the wp.data resolution state for any of the options to be accurate. This could be a gotcha for folks not realizing `getOption` has a special resolution state.

packages/data/src/options/resolvers.js
joshuatf: Up for rewording this comment, but I believe it's accurate.  We need to delete the fetch from our local variable to **allow** (keyword) wp data to handle invalidation.  It does not invalidate cache directly, just allows it.

If we don't remove this fetch and we want to invalidate this cache using `invalidateResolution`, it would never actually work since the previous promise has already resolved.

joshuatf:
>By doing things this way, the control will only yield when the batch fetch is complete and all your timeouts are absorbed by the control itself. So you retain the resolution state built in for each option in wp.data while still doing a batched request.

This is a great idea @nerrad.  I'll take a quick whack at it today and see if I can't get something working.

joshuatf:
Okay, got this working by moving the fetch logic into the control as @nerrad suggested (thanks for the suggestion!).  Resolution status seems to be working, though you can see that one option resolves first in this case since its promise is resolved at the end of the stack so I think this is expected behavior.

Screen Shot 2020-06-02 at 6 45 38 PM

I also didn't abstract the batch fetch function yet, though I agree this is probably a good idea.  But we can always tackle this in a follow-up when we need it.

@psealock I removed some of the `isRequesting` logic I added before, but didn't want to hijack your PR too much.  Happy to help clean things up a bit more, just not sure how much of that logic is needed for other areas of this store.

psealock:
Thanks for the awesome help here @joshuatf and @nerrad !

I cleaned this up, added some tests and doc blocks. I'll merge this to the original PR and continue with the refactor and rebase.

packages/data/src/options/resolvers.js
```diff
@@ -1,25 +1,38 @@
...
+ * Request an option value.
+ * 
+ * @param {string} name - Option name
+ */
+export function* getOption( name ) {
+	try {
+		const result = yield batchFetch( name );
+		yield receiveOptions( result );
+	} catch ( error ) {
+		yield setRequestingError( error, name );
```
nerrad, line 36: I'm not sure this will be entirely accurate. With the batch fetch, there's the potential for multiple `names` to be in the request that causes the error right? Would you want that reflected in the error state? That's why in my example I suggested yielding an object from the batchFetch control: `{ result, names }` so you have all the option names that were used in the _actual_ request.

Issue #4303, opened 2020-05-06
Releasing the new home screen on Core


pmcpinto:
Now that Core moved to monthly releases it won’t make sense to ask users to install wcadmin to use the new home screen. So we should merge the feature directly on Core and have the following approach:

July release: new users have automatic access to the new home screen and we display a setting on `Settings - Advanced` that allows existing users to turn-on the new home screen. This setting should be turned on by default to new users and turned-off by default to existing users. The design/copy for this setting is coming soon.

We'll then make a decision of removing the setting and becoming the default experience for everyone in an upcoming release, based on the feedback/data that we'll gather between July and August/September.

EDIT: Make sure `wcadmin_settings_change` and `wcadmin_settings_view` or something similar is working so we can track usage.

pmcpinto:
@timmyc I created this issue for the roll out plan of the new home screen on Core.

timmyc:
This all looks good to me. Few questions/clarity items:

**1: Settings - Advanced**
Are we going to add a new "sub" menu item to this screen ( akin to WooCommerce.com ) for these types of settings? None of the current ones jump out to me as a logical place to add this:

<img width="1388" alt="image" src="https://user-images.githubusercontent.com/22080/81211698-9bc62600-8f88-11ea-9d47-b776348f82a3.png">

**2. on by default to new users and turned-off by default to existing users**
To turn it on by default for new users, should we chain this logic to completion of the profiler?

pmcpinto:
> Are we going to add a new "sub" menu item to this screen ( akin to WooCommerce.com ) for these types of settings? None of the current ones jump out to me as a logical place to add this

Yes, @jameskoster is going to prepare the mockup.

> To turn it on by default for new users, should we chain this logic to completion of the profiler?

I think that we should display it independently if they complete the profiler, so we know for sure that all the new users access the home screen by default. This also makes it easier to compare different cohorts.

jameskoster:
Latest concept here:

![features](https://cldup.com/vslHN_VVne-2000x2000.png)

pmcpinto:
> Latest concept here:

@jameskoster I think that it would be helpful to include in the description a link to a support doc with more info about the new home. Sounds good?


@timmyc @psealock do we need to make any change on `wcadmin_settings_change` and `wcadmin_settings_view` to track the behavior of this new setting?

jameskoster:
[...]

PR #4450, opened 2020-05-27
Homescreen: Add option to disable

psealock:
Fix partially https://github.com/woocommerce/woocommerce-admin/issues/4303

https://github.com/woocommerce/woocommerce-admin/issues/4303 requires the homescreen to be turned off for old users and on for new users, so this PR adds an option to control homescreen display beyond just the feature flag.

The new option is `woocommerce_homescreen_enabled`

Here is the corresponding Core pull request, which is not required to test this PR.

https://github.com/woocommerce/woocommerce/pull/26625

1. Visit the app and see there is no Home screen.
2. The sidebar links should be correct.
3. Set the `woocommerce_homescreen_enabled` option to `yes` and repeat 1, 2 but now the Homescreen should be accessible.

### Changelog Note:

Add a flag to optionally remove Homescreen

### DIFF
--- a/client/dashboard/index.js
+++ b/client/dashboard/index.js
@@ -22,12 +22,11 @@ const ProfileWizard = lazy( () =>
 
 class Dashboard extends Component {
 	render() {
-		const { path, profileItems, query } = this.props;
-
+		const { path, profileItems, query, homepageEnabled } = this.props;
 		if (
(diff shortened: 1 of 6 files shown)


timmyc (approved):
🎉 this is testing out well for me.

PR #4465, opened 2020-05-28
Added new notification to ask about migrating from Shopify.

### LINKED ISSUE #4198: New notification: Do you want to migrate from Shopify to WooCommerce?
**Title:** Do you want to migrate from Shopify to WooCommerce?
**Content:**  Changing eCommerce platforms might seem like a big hurdle to overcome, but it is easier than you might think to move your products, customers, and orders to WooCommerce. This article will help you with going through this process.
**Segmentation:** 
AND
the following values on `wcadmin_storeprofiler_store_business_details_continue`
`already_selling`: `true`
`used_platform`: `shopify`
**Timing:** 2 days after plugin install
**Cadence:** Once
**CTA leads to:** https://woocommerce.com/posts/migrate-from-shopify-to-woocommerce/?utm_source=inbox
**CTA label:** Learn more

octaedro:
Fixes #4198

This PR adds a new notification to ask the client if they want to migrate from Shopify to WooCommerce.

### Accessibility

This notification should be visible for:
New users with `setup_client`: `false` on`woocommerce_onboarding_profile`
AND
the following values on `woocommerce_onboarding_profile`
`already_selling`: `true`
`used_platform`: `shopify`

To set these values it's necessary to:
- Go to the `Business Details` step (in the new onboarding).
screenshot-one wordpress test-2020 05 28-13_30_21

- Set `Currently selling elsewhere?` to `Yes, on another platform`
- Set `Which platform is the store using?` to `Shopify`.

- Also, this note should be visible two days after the plugin was installed.

- Trigger the `cron`.
- Verify the note is created and appears.
- Verify the action button works properly.

### Changelog Note:
Enhancement: Added a new notification to ask the client about migrating from Shopify.

### DIFF
--- a/src/Notes/WC_Admin_Notes_Migrate_From_Shopify.php
+++ b/src/Notes/WC_Admin_Notes_Migrate_From_Shopify.php
@@ -1,0 +1,78 @@
+<?php
+/**
+ * WooCommerce Admin: Migrate from Shopify to WooCommerce.
+ *
+ * Adds a note to ask the client if they want to migrate from Shopify to WooCommerce.
+ *
+ * @package WooCommerce Admin
+ */
+
+namespace Automattic\WooCommerce\Admin\Notes;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * WC_Admin_Notes_Migrate_From_Shopify.
+ */
+class WC_Admin_Notes_Migrate_From_Shopify {
+	/**
+	 * Note traits.
+	 */
+	use NoteTraits;
+
+	/**
+	 * Name of the note for use in the database.
+	 */
+	const NOTE_NAME = 'wc-admin-migrate-from-shopify';
+
+	/**
+	 * Get the note.
+	 */
+	public static function get_note() {
+
+		// We want to show the note after two days.
+		$two_days = 2 * DAY_IN_SECONDS;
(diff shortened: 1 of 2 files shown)


src/Notes/WC_Admin_Notes_Migrate_From_Shopify.php
```diff
@@ -0,0 +1,78 @@
...
+		$note = new WC_Admin_Note();
+		$note->set_title( __( 'Do you want to migrate from Shopify to WooCommerce?', 'woocommerce-admin' ) );
+		$note->set_content( __( 'Changing eCommerce platforms might seem like a big hurdle to overcome, but it is easier than you might think to move your products, customers, and orders to WooCommerce. This article will help you with going through this process.', 'woocommerce-admin' ) );
+		$note->set_type( WC_Admin_Note::E_WC_ADMIN_NOTE_INFORMATIONAL );
+		$note->set_icon( 'info' );
+		$note->set_name( self::NOTE_NAME );
+		$note->set_content_data( (object) array() );
+		$note->set_source( 'woocommerce-admin' );
+		$note->add_action(
+			'choose-niche',
```
jeffstieler, line 70: Copy pasta. 🍝

octaedro: Thanks, this was changed in the commit bf4fd36

PR #4470, opened 2020-05-29
Use plugins store rather than getSettings

becdetat:
Switch to the plugins store rather than `getSettings` in the Jetpack CTA.

- Smoke test the Jetpack CTA, when the plugin is not installed, when it is installed but not activated, and when it is installed and activated (the CTA will disappear).

### DIFF
--- a/client/homepage/stats-overview/install-jetpack-cta.js
+++ b/client/homepage/stats-overview/install-jetpack-cta.js
@@ -84,8 +84,27 @@ function InstallJetpackCta( props ) {
 		);
 	}
 
-	if ( isDismissed || ( isJetpackInstalled && isJetpackActivated ) ) {
+	const doNotShow =
+		isDismissed ||
+		( isJetpackInstalled && isJetpackActivated && isJetpackConnected );
+	if ( doNotShow ) {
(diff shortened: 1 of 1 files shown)


joshuatf (commented):
Nice improvement, Bec!

Since plugins are now detected on activation, this returns `null` for this component as soon as Jetpack is activated.  I think the cta component should also be shown in general if Jetpack is not connected, so we can update this line to reflect that:

https://github.com/woocommerce/woocommerce-admin/blob/b9b466b50353ee7dd141531677306dab14b79ba9/client/homepage/stats-overview/install-jetpack-cta.js#L91

Not directly related to this PR, but we can also remove the `&reset-profiler=0` in the Jetpack connection redirect URL.

https://github.com/woocommerce/woocommerce-admin/blob/b9b466b50353ee7dd141531677306dab14b79ba9/client/homepage/stats-overview/install-jetpack-cta.js#L85

becdetat:
@joshuatf Done in 34d4f3f and 3f5b056

timmyc (approved):
This is testing out well for me 🎉

PR #4473, opened 2020-05-29
Create action for installing and activate plugins

### LINKED ISSUE #3825: Onboarding: Business details step should not use autoInstall
Discovered when reviewing https://github.com/woocommerce/woocommerce-admin/pull/3787 it looks like we should not use `autoInstall` with the `Plugins` component as this is a workaround and creates more issues on plugin failure that need to be handled in the parent component.

In order to do this, we'll need to do a few things
* Add `onComplete/OnStart` or `setRequestPending` methods to the `Plugin` component.
* Remove the `autoInstall` option on this `Plugins` component.
* Add button text props to the `Plugins` component.

Note that we should also wait for the async actions coming out soon in our data store so that we can `await` updates and go to the next step when all have finished.

Related async issue: https://github.com/woocommerce/woocommerce-admin/issues/2959

### Screenshots

Screen Shot 2020-03-04 at 5 09 47 PM

_Navigating between profiler steps after failure._

Screen Shot 2020-03-04 at 5 09 00 PM

_Toggling plugin options after failure._

joshuatf:
Fixes #3825 
Fixes (partially) #2959 

The purpose of this PR is to both rely on the REST API as the source of truth for messages and allow error handling to be overridden in the client as needed.

As we standardize our API responses and data stores, this should make catching errors and showing them much more straightforward.

* Replaces various places where we use the `Plugins` component with `autoInstall` with an action that returns a promise.
* Adds generic error handling tools and notice creation.
* Updates the plugin store to throw catchable errors.

**Follow-up items:**

* Replace areas relying on `isConnecting` with resolvable promises and the URL necessary to connect.
* Improve messages and errors in API responses.

1. Uninstall Jetpack and WCS.
1. Walk through the profiler.
1. Smoke test the Business Details and Benefits screen.
1. Test the Jetpack CTA component.
1. Test areas where the `Plugin` component is used to check for regressions.
1. Throw an error in the plugin API response to make sure errors are handled client-side.
1. Make sure tests are passing.


becdetat:
When I got to the Business Details step I tried advancing with no plugins selected and got this notification:
image
Selecting a plugin and advancing worked.

becdetat:
The benefits screen worked fine, as did using the Stats overview Jetpack CTA.

becdetat:
The code looks great. I would say you just need to not trigger the plugin install if no plugins were selected.

becdetat:
I'm seeing style issues (the "Build a better WooCommerce" continue button and the checkboxes) but they're on master too.

Like Jeff I also noticed that the Jetpack/WC install buttons briefly are busy, then disabled for about a minute, then enabled while the page reloads. Apart from that it looks good.

joshuatf:
Thanks for the review @becdetat and @jeffstieler!  Rebased and ready for another review.

> In my testing it was only busy for a short time and then stayed disabled with no visual indication of activity for several seconds while the plugins were installed and activated.

Great eyes.  The connection state was not busy after plugins were installed.  Updated in https://github.com/woocommerce/woocommerce-admin/pull/4473/commits/b967e972ca885987109b31888297131c0eb0374b in a way that keeps this component stateless.

>EDIT: Also, after choosing "yes please!" I was redirected to a JP screen (with an error because of localhost) which didn't seem to register that I had competed the OBW. Going to WooCommerce > Dashboard brought me back to the first step in the OBW.

This is the current behavior in `master` as well on a localhost site since the redirect contains behavior to turn off the profiler on return from Jetpack.  I do agree this leaves a little room for error though and not the best experience when developing, but let's tackle in a follow-up if possible.

octaedro:
Hi @joshuatf, you did a great job here!
Code is testing well and LGTM. The tests work perfectly! So I will pre-approve this PR.
I did find an edge case in the `Business Details` step.
If I add the `Facebook for WooCommerce` plugin, and I press continue (and the plugin is installed), when I go back to the `Business Details` step, I'm still able to turn it off, but the plugin stays installed anyways.
Maybe for this or a follow-up PR, we can disable the possibility to turn the plugin off (disable the toggle) or uninstall it if the toggle is off, after pressing `Continue`.
I did a gif showing the case: https://d.pr/i/M8QU3b

octaedro (approved):
The code looks great and testing well here! Good job @joshuatf!
I'll approve the PR and I left a comment.

PR #4474, opened 2020-05-30
SelectControl: Select search result options by clicking on item text or icon in Firefox and Safari

### LINKED ISSUE #4209: SelectControl: Can't select options by clicking on underlined option text in Firefox or Safari
Hi.

You can't filter by products when inside /admin.php?page=wc-admin&path=%2Fanalytics%2Fproducts

Also, same problem in categories when you want to filter by single cat... or in coupons when you want to filter with specific coupons. Any filter that needs you to fill an ajax text field will not work on Safari or Firefox.

**To reproduce the behavior:**
1. Use Safari 13 or Firefox
2. Go to the analytics products page
3. Open the "Show" filter
4. Select "Single product"
5. Type something
6. Select the product you want with a mouse click by **clicking directly on the underlined product name text**
7. It will do absolutely nothing and the text field will just reset.

It works perfectly with Chrome.

Woocommerce 4.0.1
WP 5.4

## Working correctly in Chrome

Selecting product from search dropdown in Chrome

## Not working in Firefox

Selecting product from search dropdown in Firefox

[original bug report edited by @mattsherman to provide additional context]

mattsherman:
This 

Fixes #4209

This PR fixes the issue where search results in the `SelectControl` could not be selected by clicking on the option text or icon in Firefox and Safari (it worked in Chrome).

I'm admittedly not 100% sure why setting `tabIndex="-1"` as I do in this PR is necessary for Firefox and Safari to work correctly. However, from testing, it appears that this doesn't cause any other side effects, so it appears to be a reasonable fix.

### Accessibility

#### Working correctly (how it should be now in all browsers)

Selecting product from search dropdown in Chrome

#### Not working correctly (how it was before, in Firefox and Safari)

Selecting product from search dropdown in Firefox

- Run branch
- Verify that you can select items in places where the `SelectControl` is used to show search results, by clicking on the underlined text or icon, in Firefox, Safari, and Chrome. Some places:
    - WooCommerce / Analytics / Orders > Show > Advanced Filters > Add a Filter > Product
    - WooCommerce / Analytics / Products > Show > Single Product
    - WooCommerce / Analytics . Categories > Show > Single Category
    - Onboarding Wizard > Store Details > Country/Region

### Changelog Note:

Fix: Search results selectable by clicking on item text or icon.

### DIFF
--- a/packages/components/src/select-control/list.js
+++ b/packages/components/src/select-control/list.js
@@ -167,6 +167,7 @@ class List extends Component {
 				id={ listboxId }
 				role="listbox"
 				className={ listboxClasses }
+				tabIndex="-1"
 			>
 				{ options.map( ( option, index ) => (
 					<Button


octaedro (approved):
Good job @mattsherman! The code looks good and is testing well here! :shipit:

PR #4480, opened 2020-06-01
Modified layout.js withSelect

### LINKED ISSUE #4479: There is a Warning in Home screen
**Describe the bug**
There is a browser console warning on the Homepage.

**To Reproduce**

1. Go to the Homepage (URL `/wp-admin/admin.php?page=wc-admin`).
2. Open the browser console.
3. A warning message will appear.

**Expected behavior**
No warning message should appear.

**Screenshots**
screenshot

* OS: MacOS
* Browser Chrome

**Additional context**
It looks like the problem is that the component is receiving a `string` ("1") instead of `boolean`.

octaedro:
Fixes #4479 
This PR adds the code necessary to be sure we're always setting a boolean value in `taskListComplete`.

### Screenshots
No warning message should appear.
screenshot

- Go to the Homepage (URL `/wp-admin/admin.php?page=wc-admin`)
- Verify there is no warning message in the browser console.

You can check whether it continues working with a `false` and `true` value. To do this:
- Execute the following SQL in your database:
````
UPDATE `wp_options` SET `option_value` = '0' WHERE `option_name` = 'woocommerce_task_list_complete';
````

### Changelog Note:
Fix: Fixed the problem that creates a browser console warning on the Homepage.

### CODE REVIEW

**Comment by psealock** [comment]
I'm surprised this kind of thing doesn't happen more often. In most cases we add or update options with "yes" or "no" so that we can tell if its never been set when the value is `undefined`.

Here is the offending code:

https://github.com/woocommerce/woocommerce-admin/blob/d315aa96fafafff957e0621e7dd7b9d22fedb7a7/client/task-list/index.js#L53-L55

In addition to the check you added, we should also fix this to "yes" and add an extra comparison for "yes" when the option is a string.

In order to future proof, what do you think about forcing `true` to be "yes" in the new dataStore? cc @joshuatf

**Comment by joshuatf** [comment]
In this case, I think we're generally just checking to see if the value exists at all since it won't be updated later, so wrapping in `Boolean()` solves this issue in the same way as comparing to `yes`.

That said, I think using a value like `yes` forces us to be more explicit and of course aligns more with the conventions we use.

>In order to future proof, what do you think about forcing true to be "yes" in the new dataStore?

I think this is a good idea, but I think we might want to consider extending `migrate_options()` to handle transforming any data.

When we updated the structure of industry slugs, we introduced some breaking changes, but it also forces us to write backwards compatible code with both data structures everywhere.  If we handle this kind of logic in `migrate_options()` we can safely assume the data has been updated to the new type and simplify our logic in stores, controllers, etc.

**Comment by joshuatf** [comment]
>I think this is a good idea, but I think we might want to consider extending migrate_options() to handle transforming any data.

To expand on this a bit, I see a couple ways to handle this.

1.  Allow a callback function in the `$migrated_options` array:

```php
'woocommerce_admin_option_name' => array(
  'name' => 'wc_admin_old_option_name',
  'transform' => function( $old_value ) {
      if ( $old_value === '1' ) {
        return 'yes';
      }
  },
)
```

2. Or just hook the option value on update inside the responsible feature file and transform when updated:

```php
// In feature file
add_filter( "pre_update_option_woocommerce_admin_option_name", array( __CLASS__, 'transform_my_option' );
function transform_my_option( $value ) {
      if ( $value === '1' ) {
        return 'yes';
      }
}

// Make sure our filter fires when wc admin is udpated.
protected static $migrated_options = array(
  'woocommerce_admin_option_name'  => 'woocommerce_admin_option_name',
);
```

psealock:
+1 for handling this PHP side in `migrate_options`. We should probably handle the case of the "0" string too:

```php
if ( $old_value === '0' ) {
    return 'no';
}
```

psealock:
Also noting that whatever front end logic exists should also be updated here too:

https://github.com/woocommerce/woocommerce-admin/blob/b4e9f394ee906beb7c9705f10cf68fcbe18b23f1/client/dashboard/customizable.js#L347-L354

octaedro:
Hi @joshuatf and @psealock, thanks for your review.
I made the changes you mentioned.
Could you give another look at this PR?

PR #4455, opened 2020-05-27
Added new notification to ask about starting a dropshipping business.

octaedro:
Fixes #4197 

This PR adds a new notification to ask the client if they are considering starting a dropshipping business.

### Accessibility

- [x] I've tested using only a keyboard (no mouse)
- [ ] I've tested using a screen reader
- [x] All text has [at least a 4.5 color contrast with its background](https://webaim.org/resources/contrastchecker/)

### Detailed test instructions:
This notification should be visible for:
New users with `setup_client`: `false` on`woocommerce_onboarding_profile`
AND
one of the following values on `woocommerce_onboarding_profile`
`product_count`: `0`
`revenue`: `none`
`revenue`: `up to 2500`

To set these values it's necessary to:
- Go to the `Business Details` step (in the new onboarding).
screenshot-one wordpress test-2020 05 25-15_51_04

- Set `How many products do you plan to display?` to `I don't have any products yet.`
- Set `What's your current annual revenue?` to `$0.00 (I'm just getting started)` or `Up to $2,500.00`.

- Also, this note should be visible one day after the plugin was installed.

- Verify the note is created and appears.
- Verify the action button works properly.

### Changelog Note:
Enhancement: Added new notification to ask the client if they are considering starting a dropshipping business.

PR #4498, opened 2020-06-04
Onboarding profile rule

becdetat:
Adds a rule that provides access to the onboarding profile.

- Make sure you've downloaded the [latest test JSON file](https://gist.githubusercontent.com/becdetat/f17e9617e5ca1211d0579cf079862905/raw/1adef708c9bd16f928cde0e17383446c537f3bd9/rinds-specs.json) or just point the poller at that URL.
- Delete all of the admin notes (`DELETE FROM wp_wc_admin_notes`)
- Set up the onboarding to match PR #4455 
    - WCAdmin active for more than one day
    - setting up client false
    - product count zero
    - revenue none or up-to-2500 ($3000 in the dropdown)
- Check for the "Are you considering starting a dropshipping business?" notification

### Changelog Note:

Enhancement: Add onboarding profile remote inbox notification rule

### DIFF
--- a/src/RemoteInboxNotifications/GetRuleProcessor.php
+++ b/src/RemoteInboxNotifications/GetRuleProcessor.php
@@ -46,6 +46,8 @@ class GetRuleProcessor {
 				return new WCAdminActiveForRuleProcessor();
 			case 'product_count':
 				return new ProductCountRuleProcessor();
+			case 'onboarding_profile':
+				return new OnboardingProfileRuleProcessor();
 		}
 
 		return new FailRuleProcessor();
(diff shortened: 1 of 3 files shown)


timmyc (approved):
This is working as described on the tin - very nice! The one thing I did notice was when clicking on the action to learn more, no tracks event was being recorded for that action. Thinking that is out of scope of this PR though. This notif looks great:

image

becdetat:
> The one thing I did notice was when clicking on the action to learn more, no tracks event was being recorded for that action. 

There should actually be a tracks event in the API whenever a note action is triggered (https://github.com/woocommerce/woocommerce-admin/blob/master/src/API/NoteActions.php#L118) - on the server side rather than in the browser.

PR #4499, opened 2020-06-04
WCPay: Remove Jetpack requirement

### LINKED ISSUE #4416: Setup Checklist: Remove Jetpack requirements for WC Pay
With #3930, WooCommerce Payments began to be shown as a recommended payment method in the new store task list for US stores. But only if Jetpack was active and connected (at least once #4092 was fixed).

Once WooCommerce Payments has a working Jetpack DNA / connection package bundled in to connect to WordPress.com, it will no longer require the Jetpack plugin, so this requirement can be removed.

We can't be sure exactly when the Jetpack DNA work will be finalized. But we know we want to no longer require the Jetpack plugin as soon as that happens, given the boost it will provide for adoption.

### Proposed Solution

The best solution for handling this seems to be to have the WC Pay plugin handle the Jetpack dependency, rather than WC Admin / WC core (which is beholden to monthly release cycles). 

An initial version of this was put together in https://github.com/woocommerce/woocommerce-admin/pull/4332 by @dechov.

If we could include that in WC 4.3, then no other work should be required in WC / WC Admin to drop the requirement. Ideally, we'd ship it in the first possible release in case the Jetpack DNA work is finished in June.

### Alternative Solutions

The only alternative I see is to wait for the WC release _after_ the Jetpack connection work is finished. The risk there is, if the Jetpack work is finished late in the WC release cycle, say June 15, we'll miss out on a month or more of adoption (if the PR freeze for WC 4.3 is before June 15th, we wouldn't be able to ship until 4.4 in August).

The impact of every week missed is significant.

### Additional context

I'll add this to Hydra's Sprint 13 project board for WC Pay so they can review and work on it in the hopes that we can get something finalized well before the 4.3 PR freeze. But I'm opening it here as the work will go into WC Admin.

DanReyLop:
Fixes #4416

cc/ @vbelolapotkov I know you're AFK, so this is just an FYI, you don't need to review it. I think @dechov has enough context to test this :)

- In https://github.com/woocommerce/woocommerce-admin/commit/fbd53439007e893ea413f72ad1c5c8c67a2b5c2d, remove the Jetpack requirement to show the WCPay card in the Payments page of the task list.
- In https://github.com/woocommerce/woocommerce-admin/commit/573f9947da82b097045ac0b37d40bb17b50b0e0f, add all the US territories to the list of "countries" we should show the WCPay payment method in.
- In https://github.com/woocommerce/woocommerce-admin/commit/313a9d637bdd25bcf062483df8f60cfa9df80e7a, remove the "Verify details" call to action after installing & activating WCPay. Instead, after the plugin is installed & activated, the Stripe KYC (or Jetpack connection) process is automatically started.

- Clone the `woocommerce-payments` repo, checkout the `jetpack/trigger-jetpack-connection-flow` branch.
- In the `woocommerce-payments` folder, run `rm -rf node_modules vendor` if you already had the repo cloned.
- Run `npm install && npm run release`, that should generate a `zip` file in `/release/woocommerce-payments.zip`. Copy it to a different folder if you want.
- Start with a site that only has this branch of WooCommerce Admin installed. No Jetpack, no WCPay, nothing else.
- Add this code to your site. It will ensure that, when installing WCPay from "wordpress.org", WC-Admin will "download" and install the `zip` you generated instead.
```php
add_filter( 'plugins_api_result', function ( $res, $action, $args ) {
	if ( ! is_wp_error( $res ) && 'plugin_information' === $action && isset( $args->slug ) && 'woocommerce-payments' === $args->slug ) {
		// Replace this path to where you have the ZIP. If your site is not on your local machine, you'll need to upload the ZIP to a known folder of the server you're using
		$res->download_link = '/Users/daniel/Downloads/woocommerce-payments.zip';
	}
	return $res;
}, 10, 3 );
```
- Make sure the site you're using has US as its store address (or an US territory).
- You should see WCPay in the Payments task list, and you should be able to install it, activate it, and start the Jetpack connection process with a single click.

### Changelog Note:
Enhancement: Jetpack is no longer required to enable WooCommerce Payments.

### DIFF
--- a/client/task-list/tasks/payments/index.js
+++ b/client/task-list/tasks/payments/index.js
@@ -3,7 +3,7 @@
  */
 import { __, sprintf } from '@wordpress/i18n';
 import classnames from 'classnames';
-import { cloneElement, Component } from '@wordpress/element';
+import { cloneElement, Component, Fragment } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import { Button, FormToggle } from '@wordpress/components';
 import { withDispatch } from '@wordpress/data';
@@ -42,10 +42,14 @@ class Payments extends Component {
 		);
 		this.state = {
 			enabledMethods,
+			configuringMethods: {},
 		};
 
 		this.completeTask = this.completeTask.bind( this );
 		this.markConfigured = this.markConfigured.bind( this );
+		this.markConfigurationFinished = this.markConfigurationFinished.bind(
+			this
+		);
 		this.skipTask = this.skipTask.bind( this );
 	}
 
@@ -135,6 +139,7 @@ class Payments extends Component {
 	markConfigured( method ) {
 		const { enabledMethods } = this.state;
 
+		this.markConfigurationFinished( method );
 		this.setState( {
 			enabledMethods: {
 				...enabledMethods,
@@ -149,6 +154,14 @@ class Payments extends Component {
 		} );
 	}
 
+	markConfigurationFinished( method ) {
+		this.setState( {
+			configuringMethods: {
+				[ method ]: false,
+			},
+		} );
+	}
+
 	getCurrentMethod() {
 		const { methods, query } = this.props;
 
@@ -159,9 +172,7 @@ class Payments extends Component {
 		return methods.find( ( method ) => method.key === query.method );
 	}
 
-	getInstallStep() {
-		const currentMethod = this.getCurrentMethod();
-
+	getInstallStep( currentMethod ) {
 		if ( ! currentMethod.plugins || ! currentMethod.plugins.length ) {
 			return;
 		}
@@ -187,6 +198,9 @@ class Payments extends Component {
 							plugins: currentMethod.plugins,
 						} );
 					} }
+					onError={ () =>
+						this.markConfigurationFinished( currentMethod )
+					}
 					autoInstall
 					pluginSlugs={ currentMethod.plugins }
 				/>
@@ -216,28 +230,46 @@ class Payments extends Component {
 		} );
 	}
 
+	getSetupElement( method ) {
+		return cloneElement( method.container, {
+			query: this.props.query,
+			installStep: this.getInstallStep( method ),
+			markConfigured: this.markConfigured,
+			markConfigurationFinished: this.markConfigurationFinished.bind(
+				this,
+				method.key
+			),
+			hasCbdIndustry: method.hasCbdIndustry,
+			key: 'setup-' + method.key,
+		} );
+	}
+
 	render() {
 		const currentMethod = this.getCurrentMethod();
-		const { methods, query } = this.props;
-		const { enabledMethods, recommendedMethod } = this.state;
+		const { methods } = this.props;
+		const { enabledMethods, recommendedMethod, configuringMethods } = this.state;
 		const configuredMethods = methods.filter(
 			( method ) => method.isConfigured
 		).length;
 
-		if ( currentMethod ) {
-			return (
-				<Card className="woocommerce-task-payment-method is-narrow">
-					{ cloneElement( currentMethod.container, {
-						query,
-						installStep: this.getInstallStep(),
-						markConfigured: this.markConfigured,
-						hasCbdIndustry: currentMethod.hasCbdIndustry,
-					} ) }
-				</Card>
-			);
-		}
+		const setupElements = methods.map( ( method ) => {
+			const { key } = method;
+			if ( currentMethod && currentMethod.key === key ) {
+				return (
+					<Card
+						className="woocommerce-task-payment-method is-narrow"
+						key={ 'setup-' + key }
+					>
+						{ this.getSetupElement( method ) }
+					</Card>
+				);
+			} else if ( configuringMethods[ key ] ) {
+				return this.getSetupElement( method );
+			}
+			return null;
+		} );
 
-		return (
+		const methodsList = (
 			<div className="woocommerce-task-payments">
 				{ methods.map( ( method ) => {
 					const {
@@ -248,6 +280,7 @@ class Payments extends Component {
 						key,
 						title,
 						visible,
+						hasInlineSetup,
 					} = method;
 
 					if ( ! visible ) {
@@ -305,6 +338,7 @@ class Payments extends Component {
 									<Button
 										isPrimary={ key === recommendedMethod }
 										isDefault={ key !== recommendedMethod }
+										isBusy={ configuringMethods[ key ] }
 										onClick={ () => {
 											recordEvent(
 												'tasklist_payment_setup',
@@ -315,9 +349,17 @@ class Payments extends Component {
 													selected: key,
 												}
 											);
-											updateQueryString( {
-												method: key,
-											} );
+											if ( hasInlineSetup ) {
+												this.setState( {
+													configuringMethods: {
+														[ key ]: true,
+													},
+												} );
+											} else {
+												updateQueryString( {
+													method: key,
+												} );
+											}
 										} }
 									>
 										{ __( 'Set up', 'woocommerce-admin' ) }
@@ -350,6 +392,13 @@ class Payments extends Component {
 					) }
 				</div>
 			</div>
+		);
+
+		return (
+			<Fragment>
+				{ setupElements }
+				{ methodsList }
+			</Fragment>
 		);
 	}
 }
--- a/client/task-list/tasks/payments/index.js
+++ b/client/task-list/tasks/payments/index.js
@@ -37,12 +37,18 @@ class Payments extends Component {
 		const { methods } = props;
 
 		const enabledMethods = {};
-		methods.forEach(
-			( method ) => ( enabledMethods[ method.key ] = method.isEnabled )
-		);
+		let recommendedMethod = 'stripe';
+		methods.forEach( ( { isEnabled, key, visible } ) => {
+			enabledMethods[ key ] = isEnabled;
+			if ( key === 'wcpay' && visible ) {
+				recommendedMethod = 'wcpay';
+			}
+		} );
+
 		this.state = {
 			enabledMethods,
 			configuringMethods: {},
+			recommendedMethod,
 		};
 
 		this.completeTask = this.completeTask.bind( this );
@@ -59,14 +65,7 @@ class Payments extends Component {
 		}
 		const { createNotice, errors, methods, requesting } = this.props;
 
-		let recommendedMethod = 'stripe';
-		methods.forEach( ( method ) => {
-			const { key, title, visible } = method;
-
-			if ( key === 'wcpay' && visible ) {
-				recommendedMethod = 'wcpay';
-			}
-
+		methods.forEach( ( { key, title } ) => {
 			if (
 				prevProps.requesting[ key ] &&
 				! requesting[ key ] &&
@@ -84,12 +83,6 @@ class Payments extends Component {
 				);
 			}
 		} );
-
-		if ( this.state.recommendedMethod !== recommendedMethod ) {
-			this.setState( {
-				recommendedMethod,
-			} );
-		}
 	}
 
 	completeTask() {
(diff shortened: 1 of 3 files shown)


client/task-list/tasks/payments/methods.js
```diff
@@ -119,9 +118,15 @@ export function getPaymentMethods( {
...
-				! hasCbdIndustry &&
-				isJetpackConnected,
+				[
+					'US',
+					'AS',
+					'PR',
+					'VI',
+					'GU',
+					'MP',
+					'UM', // U.S. and its territories (which WooCommerce classifies as countries)
```
dechov, line 128: Looks like only US and PR are supported, according to [this resource](https://support.stripe.com/questions/stripe-availability-for-outlying-territories-of-supported-countries).

client/task-list/tasks/payments/wcpay.js
```diff
@@ -38,6 +37,17 @@ class WCPay extends Component {
 				)
 			);
 			markConfigured( 'wcpay' );
+		} else if ( this.props.installStep.isComplete ) {
+			this.connect();
```
dechov, line 41: When using the Back button after redirect, it immediately redirects again. Maybe we should only automatically redirect after actual plugin installation, and keep the explicit action button for other cases?

client/task-list/tasks/payments/wcpay.js
```diff
@@ -75,35 +85,17 @@ class WCPay extends Component {
...
-								isDefault
-								isBusy={ isPending }
-								onClick={ this.connect }
-							>
-								{ __( 'Verify details', 'woocommerce-admin' ) }
-							</Button>
-						),
-					},
-				] }
+				currentStep="install"
```
dechov, line 97: Looks like the "Install WooCommerce Payments" spinner keeps spinning even after the notice appears ("Plugins were successfully installed and activated."), without any indication of what's happening. It would be nice – though probably not critical – to give UI feedback here of some sort.

client/task-list/tasks/payments/methods.js
DanReyLop: Great find! Thanks for double-checking. Fixed in https://github.com/woocommerce/woocommerce-admin/pull/4499/commits/fd1cafc9555967aa0f9ef0726d4e46bc41d5fa12

DanReyLop:
> First, this "old version" notice that I suspect won't be a problem in the for-real release of woocommerce-admin:

Correct, won't be a problem.

> And then this bad redirect that will go away with the Calypso update, right?

Yes, the user will be redirected back to the WC-Admin payments list.

client/task-list/tasks/payments/wcpay.js
DanReyLop: > keep the explicit action button for other cases?

The reason I didn't want an explicit action button is that the copy is hard to write. Before this change it said "Verify details with Stripe", but now it will be "Connect to WP.com". I'm working on removing that page altogether, which is David's ultimate goal.

DanReyLop: Can you give this another try? I've changed the UX so everything happens in the methods list: https://github.com/woocommerce/woocommerce-admin/pull/4499/commits/fd1cafc9555967aa0f9ef0726d4e46bc41d5fa12

dechov: Looks like a great change, and works well.

(Haven't looked over the code change in as much detail as I'd like yet, but can do so on Monday. If it would help, we can loop in someone else, like original author(s) of the payments task?)

Noticed something that seemed to me like a new behavior from this change, though I'm not sure – a flash of incomplete styling on page load [[screenshot](https://cloudup.com/cxXl3_WcuYv)] before resolving to the proper appearance. (Maybe somehow related to https://github.com/woocommerce/woocommerce-admin/pull/4345?)

DanReyLop: Argh, that happened because I rebased, the problem (which most times is imperceptible) is that in https://github.com/woocommerce/woocommerce-admin/pull/4345 I only updated `recommendedMethod` on `didComponentUpdate`, and not in the constructor too. I've "reverted" that change in https://github.com/woocommerce/woocommerce-admin/pull/4499/commits/97e153f592e35a57e7b032d7122c20005866d055

client/task-list/tasks/payments/index.js
```diff
@@ -216,28 +230,46 @@ class Payments extends Component {
 		} );
 	}
 
+	getSetupElement( method ) {
+		return cloneElement( method.container, {
+			query: this.props.query,
+			installStep: this.getInstallStep( method ),
+			markConfigured: this.markConfigured,
+			markConfigurationFinished: this.markConfigurationFinished.bind(
```
dechov, line 231: I like the binding of the argument but should it be consistent with `markConfigured`? Can that one be `bind`ed too? (Not a big deal.)

client/task-list/tasks/payments/index.js
```diff
@@ -216,28 +230,46 @@ class Payments extends Component {
...
-					{ cloneElement( currentMethod.container, {
-						query,
-						installStep: this.getInstallStep(),
-						markConfigured: this.markConfigured,
-						hasCbdIndustry: currentMethod.hasCbdIndustry,
-					} ) }
-				</Card>
-			);
-		}
+		const setupElements = methods.map( ( method ) => {
```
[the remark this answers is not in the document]
DanReyLop: Those "setup elements" are supposed to be invisible if they're inline. For example, the WCPay setup element is completely invisible, so we don't need to redirect to a different "page" just to show the installation progress. If the user clicks in "Stripe", then the `setupElement` for Stripe will be displayed (along with any other invisible setup elements that were currently in progress) and nothing else.

It's a bit convoluted, but that's the best I could come up with while keeping the "declarative" list of methods configuration. I'd be happy to improve the naming of things if you can think of something less confusing :)

client/task-list/tasks/payments/index.js
DanReyLop: That's [not as trivial to change](https://github.com/woocommerce/woocommerce-admin/blob/4cf137ff2d1240478395abf787795740b5cd8939/client/task-list/tasks/payments/klarna.js#L24-L27).

client/task-list/tasks/payments/index.js
```diff
@@ -216,28 +230,46 @@ class Payments extends Component {
...
+					</Card>
+				);
+			} else if ( configuringMethods[ key ] ) {
+				return this.getSetupElement( method );
+			}
+			return null;
+		} );
 
-		return (
+		const methodsList = (
```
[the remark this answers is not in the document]
DanReyLop: My bad. If `currentMethod` is present, it means that the querystring is `task=payments&method=_____`, so the only visible UI should be the `currentMethod` setup UI. Fixed in https://github.com/woocommerce/woocommerce-admin/pull/4499/commits/9f27cbba7a2e2c1c14ac456f15a739018072d11f

client/task-list/tasks/payments/index.js
```diff
@@ -305,6 +338,7 @@ class Payments extends Component {
 									<Button
 										isPrimary={ key === recommendedMethod }
 										isDefault={ key !== recommendedMethod }
+										isBusy={ configuringMethods[ key ] }
```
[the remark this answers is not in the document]
DanReyLop: Good point! Fixed in https://github.com/woocommerce/woocommerce-admin/pull/4499/commits/ae4f7d46c5eacf2f619c27f8f392fc03c6550103

client/task-list/tasks/payments/index.js
```diff
@@ -149,6 +154,14 @@ class Payments extends Component {
 		} );
 	}
 
+	markConfigurationFinished( method ) {
```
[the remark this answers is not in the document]
DanReyLop: Done in https://github.com/woocommerce/woocommerce-admin/pull/4499/commits/671303b1ce4bbf241d1f43aa425f0e54a69bec81

client/task-list/tasks/payments/index.js
```diff
@@ -42,10 +42,14 @@ class Payments extends Component {
 		);
 		this.state = {
 			enabledMethods,
+			configuringMethods: {},
```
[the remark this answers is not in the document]
DanReyLop: Done in https://github.com/woocommerce/woocommerce-admin/pull/4499/commits/5fe73fb65ce7760fa1776b2188a6706e64a4228f

client/task-list/tasks/payments/index.js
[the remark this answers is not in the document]
dechov: I realize that the WCPay one is invisible, but figured a) conceptually it seemed more organized to keep the single list of methods and "inline" the setup as a child, and b) hypothetically there can be inline setup elements for other methods that aren't invisible – nothing in the implementation is hiding them and a UI is what's generally rendered by the method components.

If it's kept as it is, then I think there should be a comment somewhere either discouraging `hasInlineSetup` for actual UI or messaging, or else pointing out that it will be at the _top_ of the screen.

dechov:
https://github.com/woocommerce/woocommerce-admin/pull/4499#discussion_r437151229 feels like a worthwhile discussion but this PR looks just about good to go. @joshuatf Would you have a chance to give this a review and/or 👍 to merge?

client/task-list/tasks/payments/index.js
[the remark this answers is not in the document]
DanReyLop: Those are some good points. What do you think about making explicit that those elements will be invisible? https://github.com/woocommerce/woocommerce-admin/pull/4499/commits/0af201a22d44dd801e199264e2f1b44f9db88cdc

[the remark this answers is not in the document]
dechov: That seems fine to me, though I'm starting to think it should be called "hidden setup", "auto setup", or "external setup", rather than "inline setup" – thoughts?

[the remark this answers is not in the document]
DanReyLop: "Auto setup" sounds good, thanks for giving several options 😂 

Changed in https://github.com/woocommerce/woocommerce-admin/pull/4499/commits/3f5f8af47120f76e2faec6c8dacefc8c91f24e8f

Inline setup made sense because you're setting it up without leaving the methods list page, but you're right, "inline" has a different meaning.

client/task-list/tasks/payments/index.js
```diff
@@ -37,15 +37,24 @@ class Payments extends Component {
 		const { methods } = props;
 
 		const enabledMethods = {};
-		methods.forEach(
-			( method ) => ( enabledMethods[ method.key ] = method.isEnabled )
-		);
+		let recommendedMethod = 'stripe';
```
joshuatf, line 40: Was there a reason for moving this back to the constructor?  I hit the bug this time where WC Pay was not in the list of payment methods.

I think the method to check for visible/enabled and updating recommended should be moved to its own function and called in the constructor and in `componentDidUpdate()`.

DanReyLop: > Was there a reason for moving this back to the constructor?

I had only re-calculated `recommendedMethod` on `componentDidUpdate()`, that was a mistake. WCPay doesn't require Jetpack anymore, so recalculating `recommendedMethod` isn't needed at all...

> I hit the bug this time where WC Pay was not in the list of payment methods.

... or so I thought. Anyway, your solution is better :)

client/task-list/tasks/payments/methods.js
```diff
@@ -130,6 +126,7 @@ export function getPaymentMethods( {
 				options.woocommerce_woocommerce_payments_settings.enabled ===
 					'yes',
 			optionName: 'woocommerce_woocommerce_payments_settings',
+			hasAutoSetup: true,
```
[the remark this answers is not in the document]
DanReyLop: > Note that new store methods are being added for easier installation/activation of plugins if you need them in #4473

That's a game-changer! Having to render a React component when you don't care about the UI is quite convoluted. I like this solution much better.

DanReyLop:
https://github.com/woocommerce/woocommerce-admin/pull/4552 is so much better! Thanks for putting that together. The `installAndActivatePlugins` function makes it so much cleaner. We can close this.

> I also encountered this error when trying to connect after following testing instructions with the download link filter, but maybe I did something wrong.

Did you point the path in your snippet to the filesystem path where the `woocommerce-payments.zip` file is? Note that the path needs to be on the WordPress server, so if you're running your local site on Docker or whatever, you'll need to move the `zip` to the Docker volume.

PR #4500, opened 2020-06-04
Inbox: Remove (duplicate) dismiss button on tracking opt-in note

### LINKED ISSUE #4176: Change inbox message: remove dismiss btn in message prompting to opt in to tracking
The new inbox notification layout (#4099) is going to have a dismiss button on each message, so we should remove the button on this specific message, in order to avoid duplicated behavior. 

More info about this inbox message here: #3112

**Note:** this change should only be shipped alongside with the new home screen

mattsherman:
Fixes #4176

With the new inbox notification layout (#4099), every inbox note has a "Dismiss" button, so the "Help WooCommerce improve with usage tracking" note now has two "Dismiss" buttons. This PR removes the "Dismiss" button added explicitly for the tracking opt-in note, so that it only has the common "Dismiss" button as every other note does.

#### Before

ScreenCapture at Wed Jun 3 16:01:54 EDT 2020

#### After

ScreenCapture at Thu Jun 4 11:14:55 EDT 2020

#### Scenarios
- New installation: opt-in note will be created after one week, with a single "Dismiss" button.
- Upgrade less than one week after initial installation: opt-in note will be created after one week, with a single "Dimiss" button.
- Upgrade greater than one week after initial installation: opt-in note, if currently in the inbox, will be modified to ensure only a single "Dismiss" button.

#### Steps
- Ensure your store is at least 1 week old (you can modify the `wc_admin_install_timestamp` option)
- Run branch
- Ensure you are opted *out of* tracking (WooCommerce > Settings > Advanced > WooCommerce.com)
- Use Crontrol plugin to run the `wc_admin_daily` job
- Verify that only the common "Dismiss" button exists on the opt-in note

Note: If you want to re-run the DB upgrade code for this PR, you will need to make sure any pending or completed actions for `wc_admin_update_130_remove_dismiss_action_from_tracking_opt_in_note` are deleted (WooCommerce > Status > Scheduled Actions).

### Changelog Note

No changelog note needed, as it is part of the new inbox layout, which should get its own changelog entry.

### DIFF
--- a/includes/wc-admin-update-functions.php
+++ b/includes/wc-admin-update-functions.php
@@ -91,3 +91,12 @@ function wc_admin_update_110_remove_facebook_note() {
 function wc_admin_update_110_db_version() {
 	Installer::update_db_version( '1.1.0' );
 }
+
+/**
+ * Remove Dismiss action from tracking opt-in admin note.
+ */
+function wc_admin_update_130_remove_dismiss_action_from_tracking_opt_in_note() {
+	global $wpdb;
+
+	$wpdb->query( "DELETE actions FROM {$wpdb->prefix}wc_admin_note_actions actions INNER JOIN {$wpdb->prefix}wc_admin_notes notes USING (note_id) WHERE actions.name = 'tracking-dismiss' AND notes.name = 'wc-admin-usage-tracking-opt-in'" );
+}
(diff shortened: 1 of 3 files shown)


timmyc (approved):
Thanks Matt, this is testing well for me:

image

PR #4501, opened 2020-06-04
Usage of URLSearchParams breaks in IE11

### LINKED ISSUE #4497: Inbox: Usage of URLSearchParams breaks in IE11
**Describe the bug**
While testing out the inbox panel in the home screen in IE11, the page loads initially, but an exception is thrown:

image

Which is due to, TIL, that IE does not have support for this method: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

The usage was introduced in the inbox refactor #4218 cc @octaedro 

**To Reproduce**

Using IE11

1. Go to the home page
2. Once inbox beings to load data, the error should occur

* Browser IE11

octaedro:
Fixes #4497

This PR removes the method `URLSearchParams` (not compatible with IE 11) to use something compatible.
This method was used on the new `Inbox notes` and on the `layout controller`.

- Using IE 11.
- Go to `WooCommerce`| `Documentation` | `search-list-control` (URL `/wp-admin/admin.php?page=wc-admin&path=%2Fdevdocs&component=search-list-control`)
- Verify the page loaded normally.
- Press the `Inbox` button that is in the header.
- Verify the error `'URLSearchParams' is undefined ` is not present in the console.

### Changelog Note:
Fix: Removed URLSearchParams method

### DIFF
--- a/client/layout/controller.js
+++ b/client/layout/controller.js
@@ -21,6 +21,7 @@ import { Spinner } from '@woocommerce/components';
  * Internal dependencies
  */
 import { getSetting } from '@woocommerce/wc-admin-settings';
+import { getUrlParams } from 'utils';
 
 const AnalyticsReport = lazy( () =>
 	import( /* webpackChunkName: "analytics-report" */ 'analytics/report' )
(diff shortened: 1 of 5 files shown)


client/layout/controller.js
```diff
@@ -50,13 +50,24 @@ export const getPages = ( homepageEnabled ) => {
 	const pages = [];
 	const initialBreadcrumbs = [ [ '', wcSettings.woocommerceTranslation ] ];
 
+	const getUrlParams = ( locationSearch ) =>
```
timmyc, line 53: Instead of having this code twice, what do you think about extracting it to a util method and adding some tests around it?

octaedro: It's a great idea @timmyc!
As it will be a `util` used in the header and in the layout components, I created a new folder `client/utils` to keep the utils that are used by different pages of the client.
If you think that it should be different just let me know and I will change it.
Also I added some tests 🙂
The changes are in the commit 9e81c5a

timmyc (approved):
Verified the fix in IE11 on browserstack. Thanks for creating the util and adding tests @octaedro

PR #4505, opened 2020-06-04
Migrate user store to wp.data

### LINKED ISSUE #3375: Data Layer: Migrate user store to wp.data

jeffstieler:
Fixes #3375.

This PR seeks to introduce a new custom hook `useUserPreferences()` and a new HoC `withCurrentUserHydration` to replace the `wc-api` implementation of user data.

It updates several dependent components to be functional components (in order to use the hook).

### Accessibility

- Test customizing the analytics dashboard sections (order, names, and individual section prefs like leaderboard items, etc)
- Verify the "last read" functionality of the Inbox
- Test customizing report columns
- Test toggling stats in the new home screen


packages/data/src/user-preferences/use-user-preferences.js
```diff
@@ -0,0 +1,149 @@
...
+
+				return {
+					updatedUser: updatedUserResponse,
+				};
+			};
+
+			return {
+				isRequesting:
+					hasStartedResolution( 'getCurrentUser' ) &&
+					! hasFinishedResolution( 'getCurrentUser' ),
```
psealock, line 137: Going with the comment below, you should also add empty array here as second arg. I like the abstraction here though.

packages/data/src/user-preferences/use-user-preferences.js
```diff
@@ -0,0 +1,149 @@
...
+				// Propagate the updated User object to the store.
+				receiveCurrentUser( updatedUser );
+
+				// Decode the WooCommerce meta after save.
+				const updatedUserResponse = {
+					...updatedUser,
+					woocommerce_meta: getWooCommerceMeta( updatedUser ),
+				};
+
+				// @todo: Do we need to start/finish resolution here?
```
psealock, line 123: To serve what purpose? In case there is no hydration?

packages/data/src/user-preferences/use-user-preferences.js
```diff
@@ -0,0 +1,149 @@
...
+					);
+
+					return {
+						error,
+						updatedUser,
+					};
+				}
+
+				// Propagate the updated User object to the store.
+				receiveCurrentUser( updatedUser );
```
psealock, line 115: I noticed that toggling the stats on the homescreen has a lag associated with it. How do you feel about optimistically calling `receiveCurrentUser` before the `await saveUser` call above? Then handling errors or even calling `receiveCurrentUser` again with returned values.

The dashboard doesn't have that problem because it also keeps user preferences in local state. We know this is an anti pattern because it creates multiple sources of truth, so I think allowing optimistic saving will prevent that sort of thing.

psealock (commented):
This is really fantastic! I like the cleanup and move to functional components. Just a few comments and questions. One thing also about optimistic updating, which can eventually get user prefs out of local state. Maybe that would be good for a follow up.

packages/data/src/user-preferences/use-user-preferences.js
jeffstieler: I noticed the lag as well, and confirmed that it happens on master too. If you’re OK with handling that in a follow up, that would be my preference as this PR is quite large.

packages/data/src/user-preferences/with-current-user-hydration.js
```diff
@@ -0,0 +1,46 @@
...
+
+			const { isResolving, hasFinishedResolution } = select( STORE_NAME );
+			const {
+				startResolution,
+				finishResolution,
+				receiveCurrentUser,
+			} = registry.dispatch( STORE_NAME );
+
+			if (
+				! isResolving( 'getCurrentUser' ) &&
```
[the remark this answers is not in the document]
jeffstieler: It looks like [`args` defaults to an empty array](https://github.com/WordPress/gutenberg/blob/1e8d78189a9b05628b171016e102e2c50e267c39/packages/data/src/namespace-store/metadata/selectors.js#L65) and [`getCurrentUser()` doesn't take any arguments](https://github.com/WordPress/gutenberg/blob/60744002c56f5fe63e32b40ce03f934d485dcf69/packages/core-data/src/selectors.js#L57).

packages/data/src/user-preferences/use-user-preferences.js
jeffstieler: The empty array argument is unnecessary. See: https://github.com/woocommerce/woocommerce-admin/pull/4505#discussion_r436019544

src/Loader.php
```diff
@@ -401,7 +401,11 @@ public static function register_scripts() {
 		wp_register_script(
 			WC_ADMIN_APP,
 			self::get_url( 'app/index', 'js' ),
-			array( 'wc-components', 'wc-navigation', 'wp-date', 'wp-html-entities', 'wp-keycodes', 'wp-i18n', 'moment' ),
+			array(
+				'wp-core-data',
+				'wc-components',
+				'wp-date',
+			),
```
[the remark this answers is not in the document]
jeffstieler: It's not that we aren't enqueuing those scripts, but they are [dependencies of `wc-components`](https://github.com/woocommerce/woocommerce-admin/blob/aff02875a3975e7a3faffff00fd595389eecc97e/src/Loader.php#L369-L370) and already defined as such. This is just removing duplication.

jeffstieler:
This just underwent a big rebase, an additional review/test is most welcome!

psealock (approved):
Gave this another run through and its working well :shipit:

PR #4516, opened 2020-06-08
Add Insight note

### LINKED ISSUE #4210: New notification: Insight
**Title:** Insight
**Content:**  A WooCommerce powered store needs on average 31 days to get the first sale. You're on the right track! Do you find this type of insight useful?
**Segmentation:** New users
**Timing:** 8 days after registration
**Cadence:** Once
**CTA leads to:** Nowhere*
**CTA label:** Yes

*- if the user clicks yes we replace the button with a message: Thanks for your feedback.

becdetat:
Fixes #4210

There is a requirement in the ticket that the button text change after being clicked. That has been implemented via a new column in the `wp_wc_admin_note_actions` table - `actioned_label`.

No specific tracks events were added because there is [a comprehensive event that is sent whenever a note action is triggered](https://github.com/woocommerce/woocommerce-admin/blob/master/src/API/NoteActions.php#L118).

### Screenshots

image
image

- Your wc-admin installation must be at least 8 days old
- run the wc-admin-daily cron task
- the Insight note should appear in the admin notes
- Click "Yes". The text should change to "Thanks for your feedback" while the page is reloading.


client/header/activity-panel/panels/inbox/action.js
```diff
@@ -69,7 +87,7 @@ class InboxNoteAction extends Component {
 				href={ action ? action.url : undefined }
 				onClick={ this.handleActionClick }
 			>
-				{ dismiss ? label : action.label }
+				{ this.getButtonLabel() }
```
joshuatf, line 90: Should the actioned text be up one level in the `Card`?  https://github.com/woocommerce/woocommerce-admin/blob/ce68a4a48bac98d9da51ddd9b6c24cfb509c6f25/client/header/activity-panel/panels/inbox/card.js#L248

I think it might be best to display the text instead of the buttons, especially if there were multiple options.  There may also be some accessibility concerns around displaying buttons after actioning.

src/Notes/DataStore.php
```diff
@@ -228,12 +228,13 @@ private function read_actions( &$note ) {
...
-					'query'   => $action->query,
-					'status'  => $action->status,
-					'primary' => (bool) $action->is_primary,
+					'id'             => (int) $action->action_id,
+					'name'           => $action->name,
+					'label'          => $action->label,
+					'query'          => $action->query,
+					'status'         => $action->status,
+					'primary'        => (bool) $action->is_primary,
+					'actioned_label' => $action->actioned_label,
```
joshuatf, line 237: Maybe `actioned_text` if we go with my above suggestion?

src/Notes/WC_Admin_Notes_Insight.php
```diff
@@ -0,0 +1,59 @@
...
+		$note->set_type( WC_Admin_Note::E_WC_ADMIN_NOTE_INFORMATIONAL );
+		$note->set_name( self::NOTE_NAME );
+		$note->set_content_data( (object) array() );
+		$note->set_source( 'woocommerce-admin' );
+
+		// Note that there is no corresponding function called in response to
+		// this. Apart from setting the note to actioned a tracks event is
+		// sent in NoteActions.
+		$note->add_action(
+			'affirm-insight',
```
joshuatf, line 49: Side note and definitely outside the scope of this PR, but the number of arguments here is getting pretty long and hard to follow.  We may eventually want to look into using a parsable array of arguments to make this more manageable.

src/Notes/WC_Admin_Notes_Insight.php
```diff
@@ -0,0 +1,59 @@
...
+		$note->set_content( __( 'A WooCommerce powered store needs on average 31 days to get the first sale. You\'re on the right track! Do you find this type of insight useful?', 'woocommerce-admin' ) );
+		$note->set_type( WC_Admin_Note::E_WC_ADMIN_NOTE_INFORMATIONAL );
+		$note->set_name( self::NOTE_NAME );
+		$note->set_content_data( (object) array() );
+		$note->set_source( 'woocommerce-admin' );
+
+		// Note that there is no corresponding function called in response to
+		// this. Apart from setting the note to actioned a tracks event is
+		// sent in NoteActions.
+		$note->add_action(
```
joshuatf, line 48: Not a blocker for this PR, but curious if we should be adding negative options for insights as well.  I.e., "No, this was not helpful." /cc @pmcpinto

src/Notes/WC_Admin_Notes_Insight.php
```diff
@@ -0,0 +1,59 @@
...
+class WC_Admin_Notes_Insight {
+	/**
+	 * Note traits.
+	 */
+	use NoteTraits;
+
+	/**
+	 * Name of the note for use in the database.
+	 */
+	const NOTE_NAME = 'wc-admin-insight';
```
joshuatf, line 26: @pmcpinto Will more insights be added in the future?  If so, let's rename this to be more specific, otherwise you can ignore this comment @becdetat.

joshuatf (commented):
Very cool to see the actioned text on this!  Nice work.  Added a couple minor nitpicky comments, but overall testing well.

client/header/activity-panel/panels/inbox/action.js
becdetat: I agree

becdetat:
@joshuatf I've moved the actioned text up a level as you suggested, I think it's a good improvement. I'll wait for @pmcpinto 's feedback before moving this back to review required.

pmcpinto:
> Not a blocker for this PR, but curious if we should be adding negative options for insights as well. I.e., "No, this was not helpful." /cc @pmcpinto

That can be useful. In an upcoming iteration, we plan to add a pattern that works for any notification when a user clicks "dismiss". But in the meantime, it makes sense to include the "No, this was not helpful." in this specific case.

> @pmcpinto Will more insights be added in the future? 
Yes, the idea is to start adding more insights in the future.

src/Notes/WC_Admin_Notes_Insight.php
becdetat: Done in 0f9379669af5ed0706605dbe95f9a24969739264

becdetat:
I've added the negative option (which was relatively easy) and fixed a little logic error that was stopping the tracks to work (which was surprisingly hard) in 7a77be00. Ready for another review if you wouldn't mind @joshuatf .

client/header/activity-panel/panels/inbox/card.js
```diff
@@ -27,6 +27,7 @@ class InboxNoteCard extends Component {
 		this.state = {
 			isDismissModalOpen: false,
 			dismissType: null,
+			clickedActionText: '',
```
joshuatf, line 30: Nice- I really liked the way you handle this here!

What do you think about setting the default to `null` for this value?  Might make the conditions below easier to manage.

src/Notes/WC_Admin_Notes_Insight_First_Sale.php
```diff
@@ -0,0 +1,67 @@
...
+		// sent in NoteActions.
+		$note->add_action(
+			'affirm-insight-first-sale',
+			__( 'Yes', 'woocommerce-admin' ),
+			false,
+			WC_Admin_Note::E_WC_ADMIN_NOTE_ACTIONED,
+			false,
+			__( 'Thanks for your feedback', 'woocommerce-admin' )
+		);
+		$note->add_action(
```
joshuatf, line 56: Nice!

client/header/activity-panel/panels/inbox/card.js
```diff
@@ -247,22 +254,34 @@ class InboxNoteCard extends Component {
...
+							action={ action }
+							onClick={ () => this.onActionClicked( action ) }
+						/>
+					) ) }
 			</Fragment>
 		);
 	}
 
+	onActionClicked = ( action ) => {
+		this.setState( {
```
joshuatf, line 280: Maybe we can return early here if no action text exists:

```js
if ( ! action.actioned_text ) {
	return;
}
```

joshuatf (commented):
Nice refactor here, @becdetat!  Left a couple quick comments.

I'm noticing now that the page does a full refresh after clicking an action, which means the actioned text flashes on screen for a short moment before the page refreshes.  Do you see this behavior or is it just my local install?  It may be related to the new `onClick` function.

becdetat:
@joshuatf 

> I'm noticing now that the page does a full refresh after clicking an action, which means the actioned text flashes on screen for a short moment before the page refreshes. Do you see this behavior or is it just my local install? It may be related to the new onClick function.

I have the same behaviour but that's just how the href-less actions work - the API call is made which actions the note, and the empty href causes a navigation to the current URL.

client/header/activity-panel/panels/inbox/card.js
becdetat: Sure

becdetat: Done in 3a6052a4

client/header/activity-panel/panels/inbox/card.js
becdetat: Done in 3a6052a4

joshuatf (approved):
Thanks for the all the work on this PR, Bec!

>I have the same behaviour but that's just how the href-less actions work - the API call is made which actions the note, and the empty href causes a navigation to the current URL.

I don't believe that is the intended behavior.  See https://github.com/woocommerce/woocommerce-admin/pull/3039 where async actions were added to notes to avoid page navigation.

But I think this is also broken on master, as seen in the `WC_Admin_Notes_Install_JP_And_WCS_Plugins` note, which no longer works as intended.  So I think this PR is good to go for now and we can tackle those async note action issues in a follow-up.a

jeffstieler:
@becdetat as a heads up - I switched the base branch to `main`. You might need to rebase.

pmcpinto:
@becdetat is Yes and No triggering the same event prop?

becdetat:
@pmcpinto from what I can see (in `src/API/NoteActions.php` within `trigger_note_action` for the record) it is the same event, but with a unique `action_name` prop (as well as an `action_label` prop) which you should be able to use to differentiate between which action was selected.

pmcpinto:
Gotcha. Thanks for the clarification!

PR #4517, opened 2020-06-08
Update task list title

jameskoster:
Updates the title of the task list card on the home screen to read "Finish setup" rather than "Store setup".

This makes it feel like more of a continuation from the OBW.


mattsherman (approved):
LGTM. I agree that this wording change is an improvement.

PR #4518, opened 2020-06-08
Add custom autocompleter support to Search component

luizreis:
This PR updates the Search component to allow for custom autocompleters, so it can be extended further than the default ones. This is needed for some WCPay reports that rely on other APIs.

- Include the Search component with `type="custom"` and provide a custom `autocompleter` ([see default ones](https://github.com/woocommerce/woocommerce-admin/tree/master/packages/components/src/search/autocompleters))
- Results should be suggested when something is typed in the search field, based on the custom autocompleter provided

### Changelog Note:

Tweak - Add custom autocompleter support to Search component

### DIFF
--- a/packages/components/src/search/index.js
+++ b/packages/components/src/search/index.js
@@ -64,6 +64,14 @@ export class Search extends Component {
 			case 'variations':
 				return variations;
 			case 'custom':
+				if (
+					! this.props.autocompleter ||
+					typeof this.props.autocompleter !== 'object'
+				) {
+					throw new Error(
+						"Invalid autocompleter provided to Search component, it requires a completer object when using 'custom' type."
+					);
+				}
 				return this.props.autocompleter;
 			default:
 				return {};
--- a/packages/components/src/search/index.js
+++ b/packages/components/src/search/index.js
@@ -250,7 +250,6 @@ Search.defaultProps = {
 	showClearButton: false,
 	staticResults: false,
 	disabled: false,
-	autocompleter: {},
 };
 
 export default Search;
--- a/packages/components/src/search/index.js
+++ b/packages/components/src/search/index.js
@@ -63,6 +63,8 @@ export class Search extends Component {
 				return usernames;
 			case 'variations':
 				return variations;
+			case 'custom':
+				return this.props.autocompleter;
 			default:
 				return {};
 		}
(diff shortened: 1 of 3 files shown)


packages/components/src/search/index.js
```diff
@@ -63,6 +63,8 @@ export class Search extends Component {
 				return usernames;
 			case 'variations':
 				return variations;
+			case 'custom':
+				return this.props.autocompleter;
```
psealock, line 67: Instead of a default prop for `autocompleter`, maybe this would be a good spot to throw an error with a message explaining that an `autocompleter` object is required. Or a custom propType described in https://reactjs.org/docs/typechecking-with-proptypes.html perhaps?

luizreis: Great idea! I've removed the default type in https://github.com/woocommerce/woocommerce-admin/pull/4518/commits/8df16ff1bb77f1ea9b814bbfc29793b4238b92a0 and added the error in https://github.com/woocommerce/woocommerce-admin/pull/4518/commits/5a3b934b1d9fd4b4ec53bb3cd36ca7f8c259dbb9.

Do you feel like we should check the individual properties of the `autocompleter` and throw the same error if some of them are missing?

Also, should that error message be translated? I've found some examples of dev related errors that aren't, so I didn't wrap it in `__`.

timmyc (commented):
@luizreis could you please also add an item to the [changelog here](https://github.com/woocommerce/woocommerce-admin/blob/master/packages/components/CHANGELOG.md) so we can make sure to include this change when we cut a new release of the components to npm?

luizreis:
> @luizreis could you please also add an item to the [changelog here](https://github.com/woocommerce/woocommerce-admin/blob/master/packages/components/CHANGELOG.md) so we can make sure to include this change when we cut a new release of the components to npm?

@timmyc - sure! Added in https://github.com/woocommerce/woocommerce-admin/pull/4518/commits/a96d4181b83730e4526dc4de02177b1d137cfa27.

packages/components/src/search/index.js
psealock: >Do you feel like we should check the individual properties of the autocompleter and throw the same error if some of them are missing?

I don't think so. Maybe if you went the custom proptype route.

>should that error message be translated?

Good question. I don't think so because a developer will search the codebase for the string.

psealock (approved):
Thanks for the changes @luizreis. Looks good to me 🚢

PR #4522, opened 2020-06-09
Correctly remove reference to debounced updateSliderHeight method

### LINKED ISSUE #4521: TypeError: can't access property "querySelector", this.container.current is null
**Describe the bug**

Visit the marketing tab the go the WooCommerce -> Customers screen (or other WC admin registered screen). You'll get an error like `TypeError: can't access property "querySelector", this.container.current is null`.

If you don't go to the marketing tab first you won't get the error

**To Reproduce**

1. Go to the marketing tab
2. Go to the WooCommerce -> Customers screen
3. Wait a little bit, do some things and view console log
4. See `TypeError: can't access property "querySelector", this.container.current is null`

**Expected behavior**
Shouldn't get the error

**Screenshots**
https://d.pr/i/UEVsdh

jconroy:
Fixes #4521

The slider component used on the marketing tab currently watches the resize event to adjust slide height. 

We're are currently adding the event listener with debounce but not correctly removing it. [Some information about the why we aren't doing it correctly is here for reference](https://chipcullen.com/troubleshooting-adding-and-removing-eventlisteners/).

As a result after viewing the marketing tab then browsing to another wc-admin page results in errors like `TypeError: can't access property "querySelector", this.container.current is null` because the listener is still trying to do its thing.

- Pull branch
- Build
- Go to the marketing tab - resize screen etc.
- Go to WooCommerce -> Customers (or another wc-admin page like Analytics -> Orders
- resize, click around
- Check console to make sure you don't see the event / error firing off.

### DIFF
--- a/client/marketing/components/slider/index.js
+++ b/client/marketing/components/slider/index.js
@@ -21,21 +21,18 @@ class Slider extends Component {
 		this.container = createRef();
 		this.onEnter = this.onEnter.bind( this );
 		this.updateSliderHeight = this.updateSliderHeight.bind( this );
+		this.debouncedUpdateSliderHeight = debounce( this.updateSliderHeight, 50 );
 	}
 
 	/**
 	 * Update the slider height on Resize
 	 */
 	componentDidMount() {
-		// Update the slider height on Resize
-		window.addEventListener(
-			'resize',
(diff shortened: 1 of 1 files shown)


jconroy:
Thanks @psealock

PR #4523, opened 2020-06-09
Runtime feature config override

### LINKED ISSUE #4417: Allow managed hosts to toggle features

psealock:
Fixes #4417

This supercedes https://github.com/woocommerce/woocommerce-admin/pull/4418 with some extra changes regarding dynamic creation of `window.wcAdminFeatures`.

### Background

Features configs were created via webpack for client side code. Now that we require a PHP filter to disable features, it needs to be created in PHP for analysis at runtime. This removes the ability to pass the cli arg `WC_ADMIN_ADDITIONAL_FEATURES` when building a release. I think this was a seldom used feature, so I'm ok with the tradeoff.

### Changes

* Removed usage of `is_homepage_enabled` in favor of `is_feature_enabled`.
* `window.wcAdminFeatures` is now created in PHP instead of webpack and reflects `is_feature_enabled` which takes into account the disable option in the case of homescreen.
* Now that the option value is baked into `window.wcAdminFeatures`, I removed the requirement of the option to be preloaded and evaluated client side, which cleans up the code quite a bit.

### Test

1. Try disabling features with a filter. The following example disables devdocs and homepage.

```php
add_filter( 'woocommerce_admin_features', function( $enabled_features ) {
	$features_to_turn_off = array( 'homepage', 'devdocs' );
	return array_diff( $enabled_features, $features_to_turn_off );
} );
```

2. Toggle the homescreen by changing the value of the option `woocommerce_homescreen_enabled`, which should be "yes" in order to view the homescreen.

3. In each case, make sure the menu items are correct, there are no errors, and the appropriate views render.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -491,6 +491,10 @@ class Loader {
 			return;
 		}
 
+		wp_add_inline_script( WC_ADMIN_APP,
+			'window.wcAdminFeatures = ' . json_encode( self::get_enabled_features() )
+		, 'before' );
+
 		wp_enqueue_script( WC_ADMIN_APP );
 		wp_enqueue_style( WC_ADMIN_APP );
 		wp_enqueue_style( 'wc-material-icons' );
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -476,9 +476,12 @@ class Loader {
 			return;
 		}
 
-		wp_add_inline_script( WC_ADMIN_APP,
-			'window.paul = ' . json_encode( self::get_features() )
-		, 'before' );
+		$features        = self::get_features();
+		$features_object = array();
+		foreach ( $features as $key ) {
+			$features_object[ $key ] = true;
+		}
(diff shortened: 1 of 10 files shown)


src/Loader.php
```diff
@@ -476,6 +454,13 @@ public static function load_scripts() {
 			return;
 		}
 
+		$features        = self::get_features();
+		$enabled_features = array();
+		foreach ( $features as $key ) {
+			$enabled_features[ $key ] = self::is_feature_enabled( $key );
+		}
+		wp_add_inline_script( WC_ADMIN_APP, 'window.wcAdminFeatures = ' . wp_json_encode( $enabled_features ), 'before' );
+
```
psealock, line 463: Is there a better way to do this? I'm open to suggestions.

timmyc (approved):
🦖 most excellent. This is testing well for me. Verified the filter still works when the feature is enabled for the home screen.

PR #4524, opened 2020-06-09
Refactor marketing components to make more reusable.

jconroy:
This PR refactors several existing marketing tab components to make them more reusable in other contexts outside of the marketing tab

Components updated in the PR include
* RecommendedExtensions
* KnowledgeBase
* Slider

RecommendedExtensions
* Converted to a functional component
* Added title & description props so that they can be adjusted depending on context
* Moved to `client/marketing/components/.` - currently located in `client/marketing/overview/.` but will be used outside of the overview page so I've moved them.
* Added category prop so that the category of recommendations displayed can be adjusted depending on the context
* Added tests 🎉

KnowledgeBase
* Converted to a functional component
* Added title & description props so that they can be adjusted depending on context
* Moved to `client/marketing/components/.` - currently located in `client/marketing/overview/.` but will be used outside of the overview page so I've moved them.
* Added category prop so that the category of recommendations displayed can be adjusted depending on the context
* Added tests 🎉 

Marketing API
* Similar to moving the components I've moved the corresponding APIs functionality to a `/src/API/Marketing.php` file are they are no longer specific to the overview page. I'm not wedded to this - just felt better than leaving under marketing overview.
* Recommendations will still be pulled down from a single location and cached in a transient.
* Knowledge base posts will be fetched based on category/context and cached locally per category.

Slider
* Converted to functional component as part of clean up. Functionality should not be changed.

My hope is there should be no visible difference to current marketing tab as a result of these changes. @danielbitzer I'd appreciated your eyes on these changes.


jconroy:
https://github.com/woocommerce/woocommerce-admin/pull/4526 builds of these changes to demonstrate usage.

danielbitzer:
@jconroy this is looking really good. Those tests are cool! 

I haven't dived too deeply but I noticed an issue with the addition of the `category` param to the data resolver functions. Adding that param will need to be fully implemented in the redux selector, action and reducer in addition to the resolver. 

My thinking is in the marketing store the `recommendedPlugins` array could become an object and each category could be a key with an array of plugins as the value. E.g. `state.recommendedPlugins.coupons`. 

The reducer will need to be updated to mutate the state for the `SET_BLOG_POSTS` and `SET_RECOMMENDED_PLUGINS` actions.

jconroy:
Thanks so much for the 👀  @danielbitzer If you have time to walk through that stuff I'd really appreciate it. That's one area that still confuses me.

danielbitzer:
> If you have time to walk through that stuff I'd really appreciate it.

Looks like you've got it mostly sorted already :) Is this ready for review?

jconroy:
@danielbitzer yes please go for it. Please feel free to just push commits for me to learn from.

src/API/Marketing.php
```diff
@@ -0,0 +1,123 @@
...
+
+	/**
+	 * Return installed marketing extensions data.
+	 *
+	 * @param \WP_REST_Request $request Request data.
+	 *
+	 * @return \WP_Error|\WP_REST_Response
+	 */
+	public function get_recommended_plugins( $request ) {
+		// Default to marketing category (if no category set on the kb component).
```
danielbitzer, line 93: Not a blocker but I think it would make more sense to return all recommended plugins if no category is set. It's easier to leave it as is for now though. 

Minor: The comment mentiones the "kb component" incorrectly

jconroy:
Thanks for the review and for helping me with the data stores. Thanks also @Aljullu

PR #4530, opened 2020-06-09
Check for enabled methods before payment task completion

### LINKED ISSUE #4241: The “set up payments” is marked as complete even if I didn’t choose a payments option

Filing this on behalf of @andrewjkerr: pb22l9-8V-p2#comment-397.

Quoting:

>The “set up payments” is marked as complete even if I didn’t choose a payments option

>I clicked on “set up payments” on the setup and then clicked on “continue” without selecting an option but the set up showed it as complete (and took me to the main admin Panel 😅).

I agree with @andrewjkerr here. There is an option to click on `Done` without setting up any payment method:

Screen Shot 2020-04-28 at 4 06 47 PM

Once clicked, the task is being marked as completed and we show the message that the store is now ready to accept payments when it is not:

pay

Instead of `Done` we can say `Skip` and not show the message that the store is ready to accept payments. 

N/A

It is not a deal-breaker but the experience is confusing here. 

**Additional context**

N/A

joshuatf:
Fixes #4241 

Previously having any configured methods (but not enabled) would allow you to click "Done," resulting in a notice that stated your store is now ready to accept payments.

This PR updates this task to require at least one enabled payment before "Done" can be clicked, otherwise only the "skip" option is shown.

Note that at least one payment gateway (wc-pay) may not be fully configured and ready to accept payments so the notice that the store is ready to accept payments may not be entirely accurate.  More discussion on this in #4241 but feedback is welcome here as well.

### Screenshots
Screen Shot 2020-06-09 at 1 24 13 PM

1. Configure at least one payment method.
1. Disable all configured payment options.
1. Make sure the skip option is shown in the payment task.
1. Re-enable at least one of the payment methods.
1. Make sure the "Done" button is shown.

### DIFF
--- a/client/task-list/tasks/payments/index.js
+++ b/client/task-list/tasks/payments/index.js
@@ -239,8 +239,8 @@ class Payments extends Component {
 		const currentMethod = this.getCurrentMethod();
 		const { busyMethod, enabledMethods, recommendedMethod } = this.state;
 		const { methods, query, requesting } = this.props;
-		const configuredMethods = methods.filter(
-			( method ) => method.isConfigured
+		const hasEnabledMethods = Object.keys( enabledMethods ).filter(
+			( method ) => enabledMethods[ method ]
 		).length;
 
 		if ( currentMethod ) {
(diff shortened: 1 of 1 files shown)


octaedro (approved):
Nice job @joshuatf, code looks good and is testing well here. Just take a look at a conflict in the file `client/task-list/tasks/payments/index.js`.

PR #4531, opened 2020-06-09
Add note on first order creation

### LINKED ISSUE #4222: New inbox notification: First order
**Title:** First order
**Content:**  Congratulations on getting your first order! Now it's a great time to learn how to manage your orders.
**Segmentation:** new users
**Timing:** First order received
**Cadence:** Once
**CTA leads to:** https://docs.woocommerce.com/document/managing-orders/?utm_source=inbox
**CTA label:** Learn more

joshuatf:
Fixes #4222 

Adds a note to the inbox on first order creation.

Note that `paginate` and `return` params were removed from the `WC_Order_Query` since these produced unexpected results with drafts and the total (though this could be specific to my install).

/cc @pmcpinto This only creates a note for the user when an order is created via the checkout page, but we could alternatively add this note even if they manually add an order in the dashboard.

### Screenshots
Screen Shot 2020-06-09 at 3 16 41 PM

1. Create a fresh site with no orders (note that I ran into issues trying to manually do this with an older site that had refunds and other order types lying around).
1. Create an order via the front-end.
1. Note the note.
1. On an older site with orders, repeat this process and make sure a note is not added.


pmcpinto:
> /cc @pmcpinto This only creates a note for the user when an order is created via the checkout page, but we could alternatively add this note even if they manually add an order in the dashboard.

That's a good idea, as long as it's something that can be included quickly.

src/Notes/WC_Admin_Notes_First_Order.php
```diff
@@ -0,0 +1,63 @@
...
+	public static function get_note() {
+		// Only add this note if no previous orders exist.
+		$query  = new \WC_Order_Query(
+			array(
+				'limit' => 1,
+			)
+		);
+		$orders = $query->get_orders();
+
+		if ( count( $orders ) > 0 ) {
```
becdetat, line 47: Doesn't the query include the newly created order in the `woocommerce_checkout_create_order` action? I would actually be surprised if that was the case as I saw similar behaviour when querying for products on add.

becdetat:
I won't approve it because it looks like you might add to this anyway :-/ but otherwise it's tip top

src/Notes/WC_Admin_Notes_First_Order.php
joshuatf: I thought the same thing at first, but checking the code shows this fires right before save.  Confirmed on a new site that it works as expected, but this will change anyway as of https://github.com/woocommerce/woocommerce-admin/pull/4531/commits/327f693da60ce0555ad322d56b7f8057f7e77ae5 to check for non checkout orders as well.

joshuatf:
Thanks for taking a look, Bec!  Updated per @pmcpinto's request to fire even on manual order creation.  Ready for re-review.

timmyc (approved):
Looking good, verified note appears and action takes me to an informative article on how to manage my orders 🚀

PR #4532, opened 2020-06-09
Add pre-launch checklist note on task list completion

### LINKED ISSUE #4211: New notification: Ready to launch your store?
**Title:** Ready to launch your store?
**Content:**  To make sure you never get that sinking “what did I forget” feeling, we’ve put together the essential pre-launch checklist.
**Segmentation:** New users
**Timing:** 
When they trigger `wcadmin_tasklist_completed` or completed at least 3 tasks during the first 10 days
**Cadence:** Once
**CTA leads to:** https://woocommerce.com/posts/pre-launch-checklist-the-essentials/?utm_source=inbox
**CTA label:** Learn more

joshuatf:
Fixes #4211 

Adds a note when the task list is completed or if 3 tasks have been completed within 10 days.

### Screenshots
Screen Shot 2020-06-09 at 3 47 37 PM

1. On a newer site ( < 10 days old ) with < 3 tasks completed, run the `wc_admin_daily` cron event.
1. Make sure the note is not added.
1. Complete at least 3 tasks.
1. Trigger `wc_admin_daily` again and make sure the note is added.
1. Delete the note and modify the `woocommerce_admin_install_timestamp` option to be older than 10 days.
1. Trigger `wc_admin_daily` and make sure the note is not added.
1. Complete all tasks.
1. Trigger `wc_admin_daily` and  make sure the note is added.


timmyc (commented):
Nice work here Josh - this is working as expected, with one caveat.

So on my test, I had the following items visually completed in my task list:
- Store Details ( OBW )
- First Product Setup

I then ran the daily cron job, expecting to not see the note, but the note appeared. So decided to take a peek at why:

image

On this particular site, I didn't chose any upgrades during the profiler, but it looks like the purchase step still exists as completed, though I didn't actually complete. So wondering if we want to be precise here maybe we should either omit that step from the count... only issue there is it is a valid step in some cases.

Anyhow I'm leaning towards letting this go as-is, but curious for your thoughts, also @pmcpinto too.

joshuatf:
Great catch on that, @timmyc!  I feel that `purchase` shouldn't be recorded if no products were ever available for purchase, so reworking the logic there should resolve this.

I'll open a separate PR to handle this.

pmcpinto:
> Great catch on that, @timmyc! I feel that purchase shouldn't be recorded if no products were ever available for purchase, so reworking the logic there should resolve this.

I agree, it shouldn't be marked as completed. 

> PR for purchase completion check fix here - #4574

@joshuatf thanks for handling this so quickly. This will only be shipped in 4.4, correct?

jeffstieler (approved):
Code looks good and this is testing well. :shipit: (after fixing conflict of course)

PR #4542, opened 2020-06-10
Fix loading of RTL styles in chunks.

jeffstieler:
This PR seeks to restore RTL styles for chunks generated as a result of code splitting.

It does change the filename format (due to the Webpack plugin used) from `style-rtl.css` to `style.rtl.css`.

- Verify styles load properly in the app (I like using the network tab filtered by CSS)
- Switch to a RTL language like Arabic or Hebrew
- Verify RTL styles load properly in the app (network tab)

### Changelog Note:

Fix: RTL stylesheet loading for split code chunks.


timmyc (approved):
Verified loading works in RTL and non-RTL languages. Thanks Jeff

PR #4548, opened 2020-06-11
Button: Remove deprecated isDefault

### LINKED ISSUE #4538: Button: remove isDefault in favor of isSecondary and tidy up
`isDefault` has been deprecated:

```
Button isDefault prop is deprecated. Please use isSecondary instead.
```
This issue is to track that. 

Also, some other styles are off. We should use the buttons as they are without changing things like color or padding.

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/4538

`@wordpress/components` has deprecated `isDefault` for buttons. This PR replaces it with `isSecondary` unless there already is `isPrimary` prop.

1. Click around the app and see that buttons are rendering correctly.

### Changelog Note:

none: fixes unreleased issues

### DIFF
--- a/client/analytics/settings/historical-data/actions.js
+++ b/client/analytics/settings/historical-data/actions.js
@@ -57,7 +57,7 @@ function HistoricalDataActions( {
 						>
 							{ __( 'Start', 'woocommerce-admin' ) }
 						</Button>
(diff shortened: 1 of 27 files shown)


jeffstieler (approved):
Looks good! :shipit:

PR #4551, opened 2020-06-11
Update hubspot-for-woocommerce extension info

### LINKED ISSUE #4265: Update HubSpot Marketing Hub Integration
**Describe the bug**

HubSpot have released a new plugin

* New: https://wordpress.org/plugins/makewebbetter-hubspot-for-woocommerce/
* Old: https://wordpress.org/plugins/hubwoo-integration/

(preferably) we should either begin work investigating adding a proper API for plugins to use to integrate with the tab so that we don't need maintain integrations OR we will simply need add support for this extension.

Currently we have support for the old plugin.

**Expected behavior**

Support for the new plugin within the Installed Marketing Extensions component.

jconroy:
Fixes #4265

Updates the extension information for hubspot to use the new plugin available on WooCommerce.com at https://woocommerce.com/products/hubspot-for-woocommerce/

https://d.pr/i/GeEjO1

To test the plugin
* Hubspot requires an account - a free one will do
* Needs a HTTPS site to connect / finish setup

@danielbitzer I'm getting the following error when attempting to activate the plugin

```
[11-Jun-2020 04:56:07 UTC] PHP Warning:  session_start(): Cannot start session when headers already sent in /Users/jconroy/Sites/local/wp-content/plugins/hubspot-for-woocommerce/public/class-hubwoo-public.php on line 310
```

Clicking activate twice gets the plugin activated. Other plugins don't appear to have this issue - 

I'm 99% sure will be something that needs to be fixed on the plugin end but I was hoping to get your 👀 to double check my thinking.

(I've just tried activating it normally and still get the session_start() warnings)

### DIFF
--- a/src/Marketing/InstalledExtensions.php
+++ b/src/Marketing/InstalledExtensions.php
@@ -68,7 +68,7 @@ class InstalledExtensions {
 			'mailchimp-for-woocommerce',
 			'facebook-for-woocommerce',
 			'kliken-marketing-for-google',
-			'hubwoo-integration',
+			'hubspot-for-woocommerce',
 			'woocommerce-amazon-ebay-integration',
 		];
 	}
@@ -190,7 +190,7 @@ class InstalledExtensions {
 	 * @return array|bool
 	 */
 	protected static function get_hubspot_extension_data() {
-		$slug = 'hubwoo-integration';
+		$slug = 'hubspot-for-woocommerce';
 
 		if ( ! PluginsHelper::is_plugin_installed( $slug ) ) {
 			return false;
@@ -207,7 +207,7 @@ class InstalledExtensions {
 			}
 
 			$data['settingsUrl'] = admin_url( 'admin.php?page=hubwoo' );
-			$data['docsUrl']     = 'https://docs.makewebbetter.com/hubspot-woocommerce-integration/';
+			$data['docsUrl']     = 'https://docs.makewebbetter.com/hubspot-integration-for-woocommerce/';
 		}
 
 		return $data;


jconroy:
> I'm 99% sure will be something that needs to be fixed on the plugin end but I was hoping to get your 👀 to double check my thinking.

Looks like the issue they do a redirect on plugin activation and this messes up our flow/json responses

danielbitzer:
> Looks like the issue they do a redirect on plugin activation and this messes up our flow/json responses. If you have ideas on how to unhook that I'd be open to it. Failed so far.

I don't think unhooking that in WC Admin is ideal. Can we go back to them and ask them to update their extension?

Perhaps we could set a constant like `WC_ADMIN_ACTIVATING_PLUGIN` when we activate a plugin and then developers could conditionally disable redirects and session starts when it's true?

jconroy:
> Perhaps we could set a constant like WC_ADMIN_ACTIVATING_PLUGIN when we activate a plugin and then developers could conditionally disable redirects and session starts when it's true

Sounds good to me.

danielbitzer:
Looks like `activate_plugin` actually has a `$silent` argument which activates the plugin but doesn't do any activation actions. That could work for us but it could also cause different issues like plugins not being setup properly.

jconroy:
Yeah I saw that earlier and thought the same thing.

jconroy:
We'll get the flags into a separate PR and pass details to devs. This PR should be fine as is for now.

jconroy:
Thanks @danielbitzer

PR #4552, opened 2020-06-11
Try: WC Pay payment method with custom onClick handler

### LINKED ISSUE #4416: Setup Checklist: Remove Jetpack requirements for WC Pay

joshuatf:
Fixes #4416

This branch is an attempt to allow custom click handlers in payment methods.

In a future PR we may want to look at removing the WC Pay container entirely or rethinking how payment containers work, but the existing one is needed for redirect and marking the payment complete.

### Screenshots
Screen Shot 2020-06-11 at 1 49 33 PM

1. Follow testing instructions in https://github.com/woocommerce/woocommerce-admin/pull/4499
1. Make sure no regressions occur in other payment methods.

### DIFF
--- a/client/task-list/tasks/payments/index.js
+++ b/client/task-list/tasks/payments/index.js
@@ -43,6 +43,7 @@ class Payments extends Component {
 		);
 		this.state = {
 			enabledMethods,
+			recommendedMethod: this.getRecommendedMethod(),
 		};
 
 		this.completeTask = this.completeTask.bind( this );
@@ -50,26 +51,22 @@ class Payments extends Component {
 		this.skipTask = this.skipTask.bind( this );
 	}
 
-	componentDidUpdate( prevProps ) {
-		if ( prevProps === this.props ) {
-			return;
-		}
-		const { methods } = this.props;
+	componentDidUpdate() {
+		const { recommendedMethod } = this.state;
 
-		let recommendedMethod = 'stripe';
-		methods.forEach( ( method ) => {
-			const { key, visible } = method;
-
-			if ( key === 'wcpay' && visible ) {
-				recommendedMethod = 'wcpay';
-			}
-		} );
-
-		if ( this.state.recommendedMethod !== recommendedMethod ) {
+		const method = this.getRecommendedMethod();
+		if ( recommendedMethod !== method ) {
 			this.setState( {
-				recommendedMethod,
+				recommendedMethod: method,
 			} );
 		}
+	}
+
+	getRecommendedMethod() {
+		const { methods } = this.props;
+		return (
+			methods.find( ( m ) => m.key === 'wcpay' && m.visible ) || 'stripe'
+		);
 	}
 
 	async completeTask() {
--- a/client/task-list/tasks/payments/index.js
+++ b/client/task-list/tasks/payments/index.js
@@ -212,8 +212,6 @@ class Payments extends Component {
 		const { methods } = this.props;
 		const { key, onClick } = method;
 
-		this.setState( { busyMethod: key } );
-
 		recordEvent( 'tasklist_payment_setup', {
 			options: methods.map( ( option ) => option.key ),
 			selected: key,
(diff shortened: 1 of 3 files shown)


client/task-list/tasks/payments/index.js
```diff
@@ -42,36 +42,34 @@ class Payments extends Component {
...
-				recommendedMethod,
+				recommendedMethod: method,
 			} );
 		}
 	}
 
+	getRecommendedMethod() {
+		const { methods } = this.props;
+		return (
+			methods.find( ( m ) => m.key === 'wcpay' && m.visible ) || 'stripe'
```
DanReyLop, line 69: This will return either `'stripe'` or the WCPay method **object**. Fix:
```suggestion
			methods.find( ( m ) => m.key === 'wcpay' && m.visible ) ? 'wcpay' : 'stripe';
```

client/task-list/tasks/payments/index.js
```diff
@@ -42,36 +42,34 @@ class Payments extends Component {
 			( method ) => ( enabledMethods[ method.key ] = method.isEnabled )
 		);
 		this.state = {
+			busyMethod: null,
```
DanReyLop, line 45: Nice simplification, only allowing a single "busy" method makes the code much simpler, and I don't think many users will be complaining that they can't install several methods at once. I definitely overthought it :)

client/task-list/tasks/payments/index.js
joshuatf: Nice!  Thanks, @DanReyLop!

client/task-list/tasks/payments/index.js
joshuatf: I actually liked the way you implemented it and the idea of being able to quickly install multiple gateways.  The main reason I ended up shifting back towards a single is since multiple install/connection flows may stir up some bugs.  But its definitely something to consider in the long-run.

joshuatf:
@DanReyLop Thanks for the review!  Had to rebase this PR and update the base branch.  I think I managed to get all your changes in https://github.com/woocommerce/woocommerce-admin/pull/4552/commits/fc7de95a7fe9c2678e4ba419dc79b935d171333a but can you double check?

Feel free to push any changes you need as well as merge this if you need it in wc-admin.

DanReyLop:
> I think I managed to get all your changes in fc7de95 but can you double check?

Yes, it's all there.

> Feel free to push any changes you need as well as merge this if you need it in wc-admin.

Great, thanks!

Issue #4544, opened 2020-06-10
Inconsistent color themes and style regressions


jeffstieler:
**Describe the bug**

There are many different (and clashing) color themes at play in `master` now (see screenshots). We need to decide what colors should replace the purples and pinks. (cc: @jameskoster)

Additionally there are style regressions that need to be fixed.

**Screenshots**

![Screen Shot 2020-06-10 at 1 31 32 PM](https://user-images.githubusercontent.com/63922/84311495-02f76d00-ab29-11ea-9838-f8c36edab946.png)

![Screen Shot 2020-06-10 at 1 31 44 PM](https://user-images.githubusercontent.com/63922/84311496-03900380-ab29-11ea-8ad3-b0d34eff43ff.png)

![Screen Shot 2020-06-10 at 1 32 00 PM](https://user-images.githubusercontent.com/63922/84311497-03900380-ab29-11ea-849e-ab5805b5b11d.png)

![Screen Shot 2020-06-10 at 1 44 30 PM](https://user-images.githubusercontent.com/63922/84311498-04289a00-ab29-11ea-9e23-2196c62575ce.png)

![Screen Shot 2020-06-10 at 1 44 59 PM](https://user-images.githubusercontent.com/63922/84311499-04289a00-ab29-11ea-8f05-874d90da30c6.png)

![Screen Shot 2020-06-10 at 2 26 05 PM](https://user-images.githubusercontent.com/63922/84311500-04289a00-ab29-11ea-878c-875e420132f4.png)

![Screen Shot 2020-06-10 at 2 26 33 PM](https://user-images.githubusercontent.com/63922/84311502-04c13080-ab29-11ea-991e-ae173430e30a.png)

![Screen Shot 2020-06-10 at 2 26 58 PM](https://user-images.githubusercontent.com/63922/84311503-04c13080-ab29-11ea-9449-b498f5154b1a.png)

![Screen Shot 2020-06-10 at 2 27 17 PM](https://user-images.githubusercontent.com/63922/84311505-04c13080-ab29-11ea-8006-603e389a6736.png)

![Screen Shot 2020-06-10 at 2 29 03 PM](https://user-images.githubusercontent.com/63922/84311507-0559c700-ab29-11ea-87ec-e3f3e24d15f6.png)

![Screen Shot 2020-06-10 at 2 30 13 PM](https://user-images.githubusercontent.com/63922/84311510-0559c700-ab29-11ea-849e-67f479afa44b.png)

[...]

PR #4558, opened 2020-06-11
Theme color: Use dynamic variable from user preferences

psealock:
Partially addresses https://github.com/woocommerce/woocommerce-admin/issues/4544

Colors for buttons and various other core Gutenberg elements should get wp-admin theme colors.

1. Poke around the app, see button colors rendered correctly.
2. Edit your user profile
Screen Shot 2020-06-12 at 9 31 29 AM
3. Change to a new color scheme.
Screen Shot 2020-06-12 at 9 33 18 AM
4. See the new colors reflected.
Screen Shot 2020-06-12 at 9 48 18 AM

### Note

There are still some instances of colors being overriden where they shouldn't be (breadcrumbs?), some colors need to be tweaked (date range picker), and areas where overrides need to be placed (OBW). I'd like to follow up with those fixes, this PR is just the mechanism for getting G2 colors shown.

Instead of darken, lighten I used `color` function. See https://github.com/postcss/postcss-color-function#notes-for-former-sass-users

### Changelog Note:

- Enhancement: Use Core colors for styling accents.


becdetat:
Roll over on busy state removes the animation, it's like the `:hover` is given more precedence than `.is-busy`.

Not sure if this is due to the changes or if this is on `master` but the 'Dismiss this message' text goes over two lines. Possibly something for @octaedro to look at?

image

The checkboxes on the industry step are still misaligned, not sure if that was meant to be part of this fix:

image

The checkboxes and the step checks are also purple, which doesn't really 'go' with the blue buttons, is that right?

image

becdetat:
Code looks good, everything looks good, I would say that my comment is out of scope...

becdetat (approved):

psealock:
Thanks @becdetat for the review.

Thanks for noting the issues you spotted in the first comment, I'll be sure to address those in the follow up.

PR #4559, opened 2020-06-12
Fix advancedFilter filter order

### LINKED ISSUE #4426: Advanced filters change order after Filter is clicked
**Describe the bug**
Advanced filters, if added without following the dropdown order, change order after the filter button is clicked.

**To Reproduce**

1. Navigate to any report that has an Advanced filters option
2. Add some filters without following the dropdown order
3. Click on Filter
4. The filter list order is changed

**Expected behavior**
The filter order should stay the same after `Filter` is clicked.

**Screenshots**
https://d.pr/i/p5KnMa

h/t @dechov

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/4426

AdvancedFilters were being rendered in the order that the `query` object has. We want to show them in the order they were added. Luckily, the url is a good indicator because we add parameters to the query in order.

This PR introduces a `sort` function to reflect the order in which users add filters.

1. Navigate to any report that has an Advanced filters option
2. Add some filters without following the dropdown order
3. Click on Filter
4. The filter list order should be unchanged.
5. Try playing around with adding and removing filters, clicking in "Filter" and seeing if the order stays as expected.

Note: the diff in the render function is mostly Prettier, I just added the sort call on the `activeFilters` array.

### Changelog Note:

Fix: AdvancedFilters filter order when filtering

### DIFF
--- a/packages/components/src/advanced-filters/index.js
+++ b/packages/components/src/advanced-filters/index.js
@@ -226,6 +226,18 @@ class AdvancedFilters extends Component {
 		onAdvancedFilterAction( 'filter', { ...updatedQuery, match } );
 	}
 
+	orderFilters( a, b ) {
+		const qs = window.location.search;
+		const aPos = qs.indexOf( a.key );
+		const bPos = qs.indexOf( b.key );
+		// If either isn't in the url, it means its just been added, so leave it as is.
+		if ( aPos === -1 || bPos === -1 ) {
+			return 0;
+		}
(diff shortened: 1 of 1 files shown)


joshuatf (approved):
>Luckily, the url is a good indicator because we add parameters to the query in order.

Lucky and super creative solution!  Awesome fix, Paul.  LGTM :shipit:

PR #4562, opened 2020-06-12
Fix Jetpack CTA primary button style

becdetat:
The class names applied to the primary button in the install Jetpack CTA in the stats overview on the home page has changed, so I've had to update the specifier in the CSS.

- Uninstall Jetpack
- Visit the home page
- The primary button should be styled and interact as expected

### DIFF
--- a/client/homescreen/stats-overview/install-jetpack-cta.js
+++ b/client/homescreen/stats-overview/install-jetpack-cta.js
@@ -109,7 +109,7 @@ function InstallJetpackCta( {
 				) }
 			</p>
 			<footer>
-				<Button isPrimary onClick={ install } isBusy={ isInstalling }>
+				<Button isSecondary onClick={ install } isBusy={ isInstalling }>
 					{ getInstallJetpackText() }
 				</Button>
 				<Button onClick={ dismiss } isBusy={ isInstalling }>
--- a/client/homescreen/stats-overview/style.scss
+++ b/client/homescreen/stats-overview/style.scss
@@ -67,40 +67,20 @@ $promo-actions-border: $core-grey-light-500;
 
 article.woocommerce-stats-overview__install-jetpack-promo {
 	h3 {
-		margin: 16px 24px 8px;
+		margin: $gap $gap-large $gap-smaller;
 	}
 	p {
-		margin: 0 24px 16px;
+		margin: 0 $gap-large $gap;
 	}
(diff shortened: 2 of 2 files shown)


client/homepage/stats-overview/style.scss
```diff
@@ -85,7 +85,7 @@ article.woocommerce-stats-overview__install-jetpack-promo {
 			}
 		}
 
-		.woocommerce-layout & button.components-button.is-button.is-primary {
+		.woocommerce-layout & button.components-button.is-primary {
 			margin-left: 24px;
```
joshuatf, line 89: While we're here we can update this to use `$gap-large` and I noticed a few other places in here that could benefit from using some of these vars.

```scss
$gap-largest: 40px;
$gap-larger: 36px;
$gap-large: 24px;
$gap: 16px;
$gap-small: 12px;
$gap-smaller: 8px;
$gap-smallest: 4px;
```

client/homepage/stats-overview/style.scss
```diff
@@ -85,7 +85,7 @@ article.woocommerce-stats-overview__install-jetpack-promo {
...
 			background: transparent;
```
joshuatf, line 90: Is this supposed to be transparent?  If so, I think this button should use the `isSecondary` prop now.  Otherwise we can remove this background property.

Screen Shot 2020-06-12 at 11 47 46 AM

joshuatf (commented):
Noticing some oddities around the background color related to the other component updates.

Also added a quick suggestion around using variables for some of the padding/margin in this component.

psealock:
This should be fixed on master now.

becdetat:
Groovy.

@joshuatf I'm using the named variables now and I stripped out all of the styling apart from margins to take advantage of @psealock 's work.

image

psealock (approved):
This is looking great. One unrelated note is there seems to be a misplaced border:

Screen Shot 2020-06-15 at 4 50 50 PM

Approving these changes, but happy to review again if you want to fix that here.

joshuatf (approved):
Nice!  Love all the red in this PR.

LGTM pending Paul's suggestion- will set this to "Needs feedback" so that doesn't get overlooked.

timmyc:
I'm going to land this PR to fix the major styling issues here, we can grab the border in a follow-up

PR #4567, opened 2020-06-12
Focus date popover container on click

### LINKED ISSUE #4425: Date picker only disappears after an outside click if you focus it first
**Describe the bug**
The date picker only disappears after an outside click if you focus it first.

**To Reproduce**

1. Navigate to the customers report and toggle the advanced filters 
2. Add a date filter, such as `Registered`
3. Click on the text input to open the date picker
4. Click outside of the date picker
4. It doesn't close
5. Click on the date picker and then click outside
6. It closes

**Expected behavior**
The date picker should close on an outside click, even if it was not clicked before.

**Screenshots**
https://d.pr/i/4MddR3

h/t @dechov

joshuatf:
Fixes #4425 

Focuses the popover container on click so clicking outside closes the popover/dropdown.

There is also a `focused` prop in the react dates component that wp components uses under the hood - https://github.com/airbnb/react-dates/blob/480b9902d69e64e62a2ef5c2d3b5fc3adef94444/src/components/DayPickerSingleDateController.jsx#L110

If this prop is meant to actually gain focus and not describe it, it does not appear to be working.

Also worth noting that the solution in this PR only sometimes works in Safari.  I think that this is due to the actual HTML input grabbing focus, but this is probably something we'll need to handle upstream in `@wordpress/components`.

/cc @psealock for any thoughts or additional reasoning I missed on focus around these elements.

### Accessibility

### Screenshots
Screen Shot 2020-06-12 at 4 48 28 PM

1. Open a filter with a dropdown (e.g., Customers - Registered filter).
1. Click on the input box to open the calendar popover.
1. Click anywhere outside of the calendar/input.
1. Note that the input is closed.
1. Retest this using a screen reader.

### DIFF
--- a/packages/components/src/calendar/date-picker.js
+++ b/packages/components/src/calendar/date-picker.js
@@ -79,7 +79,6 @@ class DatePicker extends Component {
 		return (
 			<Dropdown
 				position="bottom center"
-				focusOnMount="container"
 				renderToggle={ ( { isOpen, onToggle } ) => (
 					<DateInput
 						disabled={ disabled }
--- a/packages/components/src/calendar/date-picker.js
+++ b/packages/components/src/calendar/date-picker.js
@@ -79,7 +79,7 @@ class DatePicker extends Component {
 		return (
 			<Dropdown
 				position="bottom center"
-				focusOnMount={ false }
+				focusOnMount="container"
 				renderToggle={ ( { isOpen, onToggle } ) => (
 					<DateInput
 						disabled={ disabled }


packages/components/src/calendar/date-picker.js
```diff
@@ -79,7 +79,7 @@ class DatePicker extends Component {
 		return (
 			<Dropdown
 				position="bottom center"
-				focusOnMount={ false }
+				focusOnMount="container"
```
timmyc, line 82: We should add a changelog entry to the [components changelog](https://github.com/woocommerce/woocommerce-admin/blob/master/packages/components/CHANGELOG.md) with this fix too

timmyc (approved):
This is testing well for me. Just the one item around adding a note to the changelog. But @psealock please do chime in with any other thoughts.

psealock:
>this is probably something we'll need to handle upstream in @wordpress/components

I think this is the solution. On Chrome, when I tab using just the keyboard, I end up in a limbo state where I can't see what has focus. If I tab to an input, I'd expect to be able to use it. At the very least, the controls in the date picker should receive focus.

joshuatf:
>At the very least, the controls in the date picker should receive focus.

@psealock I think you're right.  I updated this to focus on the first element as a workaround until we had an upstream fix in place.

However, on digging a little deeper, it seems that maybe the component the `Dropdown` component is functioning as expected and was broken since overriding the default `focusOnMount` property allows you to implement a custom focus solution.  See https://github.com/WordPress/gutenberg/tree/master/packages/components/src/popover

>Set this prop to false to disable focus changing entirely. This should only be set when an appropriately accessible substitute behavior exists.

I think the element gains focus by default and this may be an issue we've introduced.  Can you take a look at the latest version of this branch and let me know what you think?

jeffstieler:
@joshuatf as a heads up - I switched the base branch to `main`. You might need to rebase.

psealock:
The whole point of leaving focus on the input was so that keyboard or screen reader users could type the date as opposed to using the UI.

This is an improvement to the last iteration because keyboard users can navigate the datepicker right away, but it still feels wrong and the fault lies in the upstream implementation.

psealock:
In fact, I can still reproduce the bug on this branch, which this solution only masks the problem.

1. Tab into the input, see the focus in the first tabbable input of the datepicker
2. Decide you'd like to use the text input instead and shift + tab to go back one tabbable element.
3. See focus on the text input.
4. Click outside the dropdown, see that it does not disappear.

joshuatf:
>The whole point of leaving focus on the input was so that keyboard or screen reader users could type the date as opposed to using the UI.

@psealock This makes sense.  So in your opinion the popover/dropdown should close on any outside click and not just after focus, is that right?

I think that makes sense as well and will open an issue if you agree, though it does not seem like the intended effect based on their naming so I imagine we may uncover other bugs in this process.

psealock:
>So in your opinion the popover/dropdown should close on any outside click and not just after focus, is that right?

I think so

joshuatf:
Cool.  I think I agree with that overall, especially if the dropdown toggle area is an input, then this should be part of the focus area.

Opened an issue in https://github.com/WordPress/gutenberg/issues/23298

Closing this PR.

Issue #3897, opened 2020-03-13
Standardise colors


jameskoster:
TLDR - Remove any style overrides applied to components we're consuming from the WordPress core component package so that we gracefully inherit "[G2](https://www.figma.com/file/fnyj380i05vGzuuH60frLQ/G2-Design?node-id=85%3A6975)" design updates. https://github.com/WordPress/gutenberg/issues/18667

To make WooCommerce feel like a seamless part of the broader WordPress UI, and to better respect user preferences we intend to remove unnecessary applications of color, and be more intentional with our branding. More info in p6riRB-5na

Here's a list of elements I found that will need updating:

# Profiler
* [Step indicator backgrounds](https://cloudup.com/cLHdei5fl0a) should be `$blue-medium-focus` instead of purple.
* Focussed input border should match Gutenberg (`$blue-medium-focus`).
* Elevated elements (state / country dropdown, Digalogs) should use `$shadow-popover` for shadow styling.
* Tab styling on the theme step should inherit default Tab component colors.

# Task list
* Purple icons in the task list should be `$blue-medium-focus`.
* Snackbar styling and positioning should match Gutenberg.
* Image upload button (add a logo task) should match the button in the Image block (https://cloudup.com/cmauCa__tYQ).

# Activity Panel
* Orange notification indicators should use `$alert-red`.
* Tabs should inherit the default colorisation. No purple stroke on active tabs. Icon colors should match text.
* Set side sheet background to `$light-gray-200`.
* Set header background to `$light-gray-500`.
* Error icon in the inbox should be `$alert-yellow` instead of purple.

# Analytics
* "Presets" / "Custom" button group in the date picker should inherit the default button colors.
* The square that indicates an active selection in the date-picker should be `$blue-medium-focus`.
* The date-picker dropdown select colors should match core. Reference: https://cloudup.com/c1u1rQBTec3 
* Cards should ideally be updated to use the [`Card`](https://wordpress.github.io/gutenberg/?path=/story/components-card--default) component. Appreciate that's a big job though. If we can at least ensure the border colors match (use `$light-gray-500`) that'd be great!
* When Summary Numbers are links, the stroke at the top of the active state should be `$blue-medium-focus` instead of purple.
* The bar/link chart switcher buttons should match Gutenberg editor toolbar buttons. E.g. https://cloudup.com/cZFdWiYumvC
* The "by day, by week, by month" dropdown in charts should inherit core styles. Same as the date-picker.
[...]

PR #4573, opened 2020-06-15
Fix Task List styles

psealock:
Partially solves https://github.com/woocommerce/woocommerce-admin/issues/3897

* Purple icons in the task list should be `$blue-medium-focus`.
* Snackbar styling and positioning should match Gutenberg.
* Image upload button (add a logo task) should match the button in the Image block (https://cloudup.com/cmauCa__tYQ).

### Screenshots

Screen Shot 2020-06-16 at 11 19 41 AM
Screen Shot 2020-06-16 at 11 33 48 AM
Screen Shot 2020-06-16 at 11 34 15 AM
Screen Shot 2020-06-16 at 11 57 46 AM

1. Enable Task List.
2. Make sure the styling looks OK, including the steps themselves.
2. Confirm items found in the bullets above.
2. Do an action to fire a snackbar confirmation. See the new position is the same as the editor.
3. See links should be default blue.

### Changelog Note:

no changelog note.


psealock:
Blocked due to needing to know if the Task List gets studio pink buttons like the OBW or default g2 colors.

jameskoster:
None of the buttons should be pink anywhere :D

client/layout/transient-notices/style.scss
```diff
@@ -3,15 +3,14 @@
 .woocommerce-transient-notices {
 	position: fixed;
 	bottom: $gap-small;
-	left: 0;
+	left: 176px;
```
becdetat, line 6: This seems to have affected the OBW as well, is that intentional?

becdetat: And when the menu is collapsed:
image
Fixing these would be a little tricky though, they're a little edge-casey, and aren't a big deal so you could probably leave them for a follow-up.

psealock: I think so, but this was under the Task List section of the issue.

>Snackbar styling and positioning should match Gutenberg.

I think its OK, we can always revert if need be.

becdetat: Cool

PR #4582, opened 2020-06-15
Fix deleted products breaking Downloads Report

### LINKED ISSUE #4421: Fatal error due to product variation removal
**Affected customers**
2968918-zen

**Describe the bug**
Customer is getting this error:

> 2020-05-13T00:35:04+00:00 CRITICAL Uncaught Error: Call to a member function get_file_download_path() on boolean in .../wp-content/plugins/woocommerce-admin/src/API/Reports/Downloads/Controller.php:112
> Stack trace:
> 0 .../wp-content/plugins/woocommerce-admin/src/API/Reports/Downloads/Controller.php(59): Automattic\WooCommerce\Admin\API\Reports\Downloads\Controller->prepare_item_for_response(Array, Object(WP_REST_Request))
> 1 .../wp-includes/rest-api/class-wp-rest-server.php(1015): Automattic\WooCommerce\Admin\API\Reports\Downloads\Controller->get_items(Object(WP_REST_Request))
> 2 .../wp-includes/rest-api/class-wp-rest-server.php(342): WP_REST_Server->dispatch(Object(WP_REST_Request))
> 3 .../wp-includes/rest-api.php(306): WP_REST_Server->serve_request('/wc-analytics/r...')
> 4 .../wp-includes/class-wp-hook.php(287): rest_api_loaded(Object(WP))
> 5 .../wp-includes/class-wp-hook.php( in .../wp-content/plugins/woocommerce-admin/src/API/Reports/Downloads/Controller.php on line 112

This is happening here:
https://github.com/woocommerce/woocommerce-admin/blob/v1.1.3/src/API/Reports/Downloads/Controller.php#L112

This is due to they had a Simple product, switched it to Variable (with downloads), had sales, customers downloaded, and then they switched the product back to Simple again. They are now trying to run download reports and due to the orders have the variation in them they are getting a fatal error. 

**To Reproduce**

1. Create a variable product with 1 variation that has a download attached.
1. Add variation to cart, purchase, and then download file. 
1. Go back to product, switch to simple.
1. This may not be needed, but I did it... Add download to simple product, add to cart, purchase, and download that file.
1. Go to Analytics > Downloads, and the report will not load correctly and fatal error will be logged.

**Expected behavior**
I am not sure if we can obtain proper data for the reports here, especially since this shouldn't happen, but we could check to make sure we have a `$_product` object before trying to use its methods. 

**Screenshots**

Image Link: https://d.pr/i/v5Itcn

jeffstieler:
Fixes #4421.

This PR seeks to fix the fatal errors thrown when viewing the Downloads Report where one of the products containing the download has been deleted. Deleted products will show as `(Deleted)` in the report table.

### Screenshots

Screen Shot 2020-06-15 at 4 11 26 PM

- Create a variable product with 1 variation that has a download attached.
- Add variation to cart, purchase, and then download file.
- Go back to product, switch to simple.
- This may not be needed, but I did it... Add download to simple product, add to cart, purchase, and download that file.
- Go to Analytics > Downloads
- Verify the report does load correctly and the row shows "(Deleted)"

### Changelog Note:

Fix: deleted products breaking Downloads Report.

### DIFF
--- a/client/analytics/report/downloads/table.js
+++ b/client/analytics/report/downloads/table.js
@@ -85,16 +85,31 @@ class CouponsReportTable extends Component {
 				product_id: productId,
 				username,
 			} = download;
-			const { name: productName } = _embedded.product[ 0 ];
 
-			const productLink = getNewPath(
-				persistedQuery,
-				'/analytics/products',
-				{
-					filter: 'single_product',
-					products: productId,
-				}
-			);
+			const { code: errorCode, name: productName } = _embedded.product[ 0 ];
+			let productDisplay, productValue;
+
+			// Handle deleted products.
+			if ( errorCode === 'woocommerce_rest_product_invalid_id' ) {
+				productDisplay = __( '(Deleted)', 'woocommerce-admin' );
+				productValue = __( '(Deleted)', 'woocommerce-admin' );
+			} else {
+				const productURL = getNewPath(
+					persistedQuery,
+					'/analytics/products',
+					{
+						filter: 'single_product',
+						products: productId,
+					}
+				);
+
+				productDisplay = (
+					<Link href={ productURL } type="wc-admin">
+						{ productName }
+					</Link>
+				);
+				productValue = productName;
+			}
 
 			return [
 				{
@@ -104,12 +119,8 @@ class CouponsReportTable extends Component {
 					value: date,
 				},
 				{
-					display: (
-						<Link href={ productLink } type="wc-admin">
-							{ productName }
-						</Link>
-					),
-					value: productName,
+					display: productDisplay,
+					value: productValue,
 				},
 				{
 					display: (
--- a/tests/api/reports-downloads.php
+++ b/tests/api/reports-downloads.php
@@ -166,7 +166,7 @@ class WC_Tests_API_Reports_Downloads extends WC_REST_Unit_Test_Case {
 		$object->set_permission_id( $download->get_id() );
 		$object->set_user_id( 2 );
 		$object->set_user_ip_address( '5.4.3.2.1' );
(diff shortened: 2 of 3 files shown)


tests/api/reports-downloads.php
```diff
@@ -379,6 +379,27 @@ public function test_get_report_with_ip_address_filter() {
 		$this->assertEquals( 'test.png', $download_report['file_name'] );
 	}
 
+	/**
+	 * Test deleted product in report result set.
+	 */
+	public function test_get_report_with_deleted_product() {
```
joshuatf, line 385: Thanks for adding these tests!

client/analytics/report/downloads/table.js
```diff
@@ -85,16 +85,31 @@ class CouponsReportTable extends Component {
...
-				{
-					filter: 'single_product',
-					products: productId,
-				}
-			);
+			const { code: errorCode, name: productName } = _embedded.product[ 0 ];
+			let productDisplay, productValue;
+
+			// Handle deleted products.
+			if ( errorCode === 'woocommerce_rest_product_invalid_id' ) {
```
joshuatf, line 93: Just curious and may not require any change, but wondering if there are cases where we receive an error code for a product but still want to display it.

joshuatf (approved):
Nice fix, Jeff!  Tested and working as described.

Left one question just for clarification, but I think this is good to merge as is regardless.

client/analytics/report/downloads/table.js
jeffstieler: I'm interpreting this as "why even check the code, just check for _any_ error" - is that right?

TBH I don't know the full range of possible error codes here. I think a comprehensive look at deleted objects and their effect on our reports is worth the effort as a follow up.

PR #4587, opened 2020-06-16
Release 1.3.0-beta.1

timmyc:
Getting the 1.3.0 party under way with beta 1.


psealock (approved):
LGTM, thanks @timmyc 

I think its best to merge this right away before we start cherry picking or (lets hope not) making bespoke fixes.

psealock:
I will merge this so that the new version is in master and we don't get the outdated plugin notice. This branch can still exist for us to use in the next 1.3.x releases

PR #4589, opened 2020-06-16
Add a bottom border to the Install Jetpack CTA

becdetat:
A border was forgotten when adding the Install Jetpack CTA to the stats overview. This PR just adds it.

- Uninstall or deactivate Jetpack if you have it installed
- Visit the WC home page. There should be a border between the CTA buttons and the stats.

image


joshuatf (approved):
Nice!  Thanks for these changes, @becdetat!  LGTM 🚢

PR #4599, opened 2020-06-16
Fix theme match on mixed-case directory name

### LINKED ISSUE #4393: New Setup Wizard disallows choosing themes with Uppercase names
**Describe the bug**
After the recent WooCommerce v4 update, it now allows you to choose a Theme at the end of the setup wizard. When you choose Divi, it gives an Error: 

> The requested theme divi could not be installed. Theme API call failed. Not sure if we can do anything about this.

**Note:** We at Elegant Themes, the makers of Divi, are getting a lot of bug reports because of this issue, so we consider this very important to resolve as quick as possible.

**To Reproduce**
1. Install a fresh version of WooCommerce
2. Use the new Setup Wizard
3. On the 5th step you can choose a Theme, and if Divi is installed it will be listed. If you press Choose it will not work. and you will get the following error:
> The requested theme divi could not be installed. Theme API call failed. Not sure if we can do anything about this.

**Expected behavior**
The wizard should allow you to choose installed themes even if they have a capital letter in their name.

**Screenshots**

First case: The default theme is active and not Divi

Second case: Divi is activated before choosing the theme from new WooCommerce setup wizard

**Additional context**
The Divi theme uses a capital first letter, so its directory in wp-content is in fact uppercase, and this is what is not being accounted for when the sanitization is applied, it's lowercasing `Divi` to `divi` which is causing the `in_array` check to fail here:

https://github.com/woocommerce/woocommerce-admin/blob/master/src/API/OnboardingThemes.php#L171-L177

*and here:*

https://github.com/woocommerce/woocommerce-admin/blob/master/src/API/OnboardingThemes.php#L95-L104

The theme is sanitized, and lowercased.

```
$slug = sanitize_key( $theme ); // `Divi` becomes `divi`
```

```
$installed_themes = wp_get_themes(); // returns Divi
```

Then in either case it will fail:

```
if ( in_array( $slug, array_keys( $installed_themes ), true ) ) // $slug is `divi`, so it fails
```
*and*
```
if ( ! in_array( $theme, array_keys( $installed_themes ), true ) ) { // $theme is `divi`, so it fails
```

So, all that needs to change is the use of `sanitize_key` (because it force lowercases the value, and instead you could use:

```
$theme = preg_replace( '/[^a-zA-Z0-9_\-]/', '', $theme );
```

This is based on, and almost the same as, `sanitize_key`, just without the lowercasing.

Note, the implementation on lines ~171-177 seems to have a further bug as the sanitized value `$slug` is not used but `$theme` is instead... 
Screen Shot 2020-05-18 at 12 21 27 PM

travisseitler:
Fixes #4393.

This is an update to the Onboarding Themes Controller so that it can handle cases where a theme directory name includes mixed-case characters.

The change in how directory names are sanitized reflects how a similar bug was addressed in WordPress Core _(Trac #37924, [Cannot delete or update themes in directories containing an uppercase letter from Appearance > Themes](https://core.trac.wordpress.org/ticket/37924))_.

### Changelog Note:

Fix: During theme installation and activation, preserve the character case in theme directory names in our validation checks.

### DIFF
--- a/src/API/OnboardingThemes.php
+++ b/src/API/OnboardingThemes.php
@@ -92,7 +92,7 @@ class OnboardingThemes extends \WC_REST_Data_Controller {
 			return new \WP_Error( 'woocommerce_rest_invalid_theme', __( 'Invalid theme.', 'woocommerce-admin' ), 404 );
 		}
 
-		$slug             = sanitize_key( $theme );
+		$slug             = sanitize_text_field( $theme );
 		$installed_themes = wp_get_themes();
 
 		if ( in_array( $slug, array_keys( $installed_themes ), true ) ) {
@@ -168,10 +168,10 @@ class OnboardingThemes extends \WC_REST_Data_Controller {
 
 		require_once ABSPATH . 'wp-admin/includes/theme.php';
 
-		$slug             = sanitize_key( $theme );
+		$slug             = sanitize_text_field( $theme );
 		$installed_themes = wp_get_themes();
 
-		if ( ! in_array( $theme, array_keys( $installed_themes ), true ) ) {
+		if ( ! in_array( $slug, array_keys( $installed_themes ), true ) ) {
 			/* translators: %s: theme slug (example: woocommerce-services) */
 			return new \WP_Error( 'woocommerce_rest_invalid_theme', sprintf( __( 'Invalid theme %s.', 'woocommerce-admin' ), $slug ), 404 );
 		}


src/API/OnboardingThemes.php
```diff
@@ -168,10 +168,10 @@ public function activate_theme( $request ) {
 
 		require_once ABSPATH . 'wp-admin/includes/theme.php';
 
-		$slug             = sanitize_key( $theme );
+		$slug             = sanitize_text_field( $theme );
```
joshuatf, line 171: If we are going to allow case, I think we should not sanitize twice and instead replace the above instance of:

`$theme          = sanitize_title_with_dashes( $request['theme'] );`

With sanitization that allows case:

`$theme          = sanitize_text_field( $request['theme'] );`

jeffstieler:
Again, thanks for this PR @travisseitler!

We just updated our default branch to be `main`, not `master`. Can you please change the base branch for this PR and rebase your branch? Thank you!

travisseitler:
Closing this in deference to @ayubadiputra's [far superior PR](https://github.com/woocommerce/woocommerce-admin/pull/4706)... :smile:

PR #4618, opened 2020-06-17
Master -> main.

jeffstieler:
> The `tl;dr` version is that “master” in the “master/slave” context is ~noninclusive~ oppressive, and this is the context from which the “master” branch name [appears to be derived from](https://twitter.com/tobie/status/1270290278029631489) and could be changed.

_pb0Spc-Is-p2_

- Switch to this branch in the GitHub browser
- Check documentation links (make sure they work)

Unfortunately we won't be able to check the `main` branch protection in `pre-push-hook` until it's merged (AFAICT).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,8 +48,7 @@ script:
 
 branches:
   only:
-    - master
-    - version/1.0
+    - main
 
 before_deploy:
   # Remove our unneeded symlink.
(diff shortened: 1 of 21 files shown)


timmyc (approved):
Thanks for helping move us forward on this Jeff.

PR #4621, opened 2020-06-17
Inbox panel - Actioned inbox notifications now are visible

### LINKED ISSUE #4602: Inbox panel  - inbox notifications disappear after clicking on the call to action
**Describe the bug**
The inbox notifications are disappearing from the panel (both on the home screen and in the side panel) when clicking in the main call to action of a message

**To Reproduce**
1. Activate wcadmin 1.3.0 beta
2. Enable the home screen
3. Go to WooCommerce - Home and click on the main call to action of a notification
4. See that the notification disappears after clicking on it

**Expected behavior**
I expect to clear a notification only when I click to dismiss it

octaedro:
Fixes #4602

This PR modifies the inbox notifications to make them not disappear after clicking on the call to action.
Additionally, the notifications that received a click on the call to action will be shown as `read`.

### Accessibility

### Screenshots
Screen Capture on 2020-06-17 at 13-10-10

#### Read Inbox notification call to action:
1. Open the Inbox panel, which should be located in the `Home` screen.
2. Choose an already read notification and press its call to action button.
3. Refresh the webpage.
4. The notification should be still there.

#### Unread Inbox notification call to action:

_Add some new inbox notifications with the [helper plugin](https://github.com/woocommerce/woocommerce/wiki/Release-Testing-Instructions-WooCommerce-4.3#helper-plugin)_.

1. Choose an unread notification and press its call to action button.
2. After a few seconds, the notification will appear as `read`.

#### New messages warning
1. If there are any `unread` inbox notification, a red dot should be visible on the `Inbox` button in the header.
9-unread-dot

2. Verify that it's working correctly.
3. Confirm that the unread number next to the `Inbox` title corresponds with the unread messages number.
10-unread-count

#### Dismiss
1. Dismiss an inbox message.
2. Undo the dismiss by pressing the `Undo` button in the `snackbar` (it will appear after dismissing the message).
3. Verify that the process works correctly.
4. Do the same with the `Dismiss all messages` option.

### Changelog Note:
Fix: Inbox notifications now are visible after the call to action button has been pressed.

### DIFF
--- a/client/header/activity-panel/panels/inbox/utils.js
+++ b/client/header/activity-panel/panels/inbox/utils.js
@@ -15,13 +15,14 @@ export function getUnreadNotesCount( notes, lastRead ) {
 		const {
 			is_deleted: isDeleted,
 			date_created_gmt: dateCreatedGmt,
+			status,
 		} = note;
 		if ( ! isDeleted ) {
-			return (
+			const unread =
 				! lastRead ||
 				! dateCreatedGmt ||
-				new Date( dateCreatedGmt + 'Z' ).getTime() > lastRead
-			);
+				new Date( dateCreatedGmt + 'Z' ).getTime() > lastRead;
+			return unread && status === 'unactioned';
 		}
 	} );
 	return unreadNotes.length;
--- a/client/header/activity-panel/panels/inbox/utils.js
+++ b/client/header/activity-panel/panels/inbox/utils.js
@@ -4,7 +4,7 @@
 import { filter } from 'lodash';
 
 /**
- * Get the count of the unread notes.
+ * Get the count of the unread notes from the received list.
  *
  * @param {Array} notes - List of notes, contains read and unread notes.
  * @param {number} lastRead - The timestamp that the user read a note.
--- a/client/header/activity-panel/unread-indicators.js
+++ b/client/header/activity-panel/unread-indicators.js
@@ -28,6 +28,9 @@ export function getUnreadNotes( select ) {
 	if ( ! lastRead ) {
 		return null;
 	}
+
+	// @todo This method would be more performant if we ask only for 1 item per page with status "unactioned".
+	// This change should be applied after having pagination implemented.
 	const notesQuery = {
 		page: 1,
 		per_page: QUERY_DEFAULTS.pageSize,
--- a/client/header/activity-panel/unread-indicators.js
+++ b/client/header/activity-panel/unread-indicators.js
@@ -9,20 +9,19 @@ import { SETTINGS_STORE_NAME, USER_STORE_NAME } from '@woocommerce/data';
 import { DEFAULT_ACTIONABLE_STATUSES } from 'analytics/settings/config';
 import { getSetting } from '@woocommerce/wc-admin-settings';
 import { QUERY_DEFAULTS } from 'wc-api/constants';
+import { getUnreadNotesCount } from './panels/inbox/utils';
 
 export function getUnreadNotes( select ) {
-	const {
-		getNotes,
-		getNotesError,
-		isGetNotesRequesting,
-	} = select( 'wc-api' );
+	const { getNotes, getNotesError, isGetNotesRequesting } = select(
+		'wc-api'
+	);
 
 	const { getCurrentUser } = select( USER_STORE_NAME );
 	const userData = getCurrentUser();
 	const lastRead = parseInt(
 		userData &&
-		userData.woocommerce_meta &&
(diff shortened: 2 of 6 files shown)


client/header/activity-panel/panels/inbox/utils.js
```diff
@@ -15,13 +15,14 @@ export function getUnreadNotesCount( notes, lastRead ) {
 		const {
 			is_deleted: isDeleted,
 			date_created_gmt: dateCreatedGmt,
+			status,
 		} = note;
 		if ( ! isDeleted ) {
-			return (
+			const unread =
```
joshuatf, line 21: One issue with doing it this way is that you could have unread or unactioned notes from 25 notes ago and this would return 0 notes because those notes aren't queried.

This may be more of an edge case, but I just want to surface this potential problem before we merge it if we think this behavior is okay.

client/header/activity-panel/unread-indicators.js
```diff
@@ -31,27 +30,26 @@ export function getUnreadNotes( select ) {
 	}
 	const notesQuery = {
 		page: 1,
-		per_page: 1,
+		per_page: QUERY_DEFAULTS.pageSize,
```
joshuatf, line 33: Related to my above comment- it would be much better to query a single item with the items that we need instead of querying a page of them and filtering client-side.

joshuatf (commented):
Nice work here, Fernando!

Left a couple comments.  Also, I'm curious what the plan is for notes like "Insight" where I don't think we'd want the note to remain or allow continued feedback.  Should this type of note disappear or should the actioned text always appear if they've been actioned? /cc @pmcpinto

client/header/activity-panel/panels/inbox/utils.js
octaedro: Hi @joshuatf thank you for the review 🙂

Since we don't have pagination (or lazyloading) implemented in the notification list, looking at all the notes to verify if any of them are `unread` could be a little bit tricky.
E.g.: If we have 27 notes to show, the only way a user is able to see the other 2 is by dismissing 2 notes. In this case, if the first notification is `unread`, the user will be notified that they have an unread message but they won't be able to see it in a normal navigation flow. They will be able to see it only by dismissing 2 notifications individually.

Having this in mind, I thought it would be better for the user to get an unread message warning for the inbox notifications that they can see/dismiss. But I really would like to know your opinion about this, taking these additional details into consideration.

I think that for a follow-up PR it would be very helpful to add pagination (or lazyloading) to the notification list. And for now, if you think that we should use all the notifications to do the check I can do it with no problem 👍

joshuatf: >Since we don't have pagination (or lazyloading) implemented in the notification list

Good point.  This is not something we have to handle now, but would be nice to in the future for performance and accuracy once we have pagination of some kind.

I think a TODO comment in the indicators will be enough for now.

client/header/activity-panel/unread-indicators.js
joshuatf: A TODO comment above this object would be helpful to remind us we'll need to change this once pagination is introduced.

joshuatf (approved):
I agree with your points, Fernando and this is testing well for me.

Added a comment about leaving a reminder in the code for us to update later, but pre-approving.  Nice job!  🚢

pmcpinto:
> Should this type of note disappear or should the actioned text always appear if they've been actioned? /cc @pmcpinto

That's a great question. I think that it should disappear or at least hide the CTA and say something like "Thank you for your feedback". This is a topic that requires a bit more consideration :) cc @jameskoster for additional feedback

jameskoster:
Yeah I agree that it's probably fine for this one to disappear. It's a one-time action and unlike most other notifications there's very little point in keeping it around after a user has interacted with it. 

There should probably be some confirmation that your selection has been made though - a Snackbar should do the trick. 

I'm also trying to think of a more engaging title than "Insight" 🤔Maybe something like "Did you know?"

joshuatf:
@octaedro If you want we can handle notes disappearing in a follow-up as long as we get an issue created around it before merging this one in.

pmcpinto:
> I'm also trying to think of a more engaging title than "Insight" 🤔Maybe something like "Did you know?"

👍 "Did you know?" seem to be a better option. Thanks for the suggestion

octaedro:
Excellent! Thank you all for your feedback.
I created the issue [4656](https://github.com/woocommerce/woocommerce-admin/issues/4656) to address the notes disappearing and I'll merge this PR.

PR #4624, opened 2020-06-18
Fix logic error that was stopping dismiss link from appearing on the inbox cards

becdetat:

- The dismiss link should be shown on all inbox notes
- Clicking an action on the "Insight - first sale" should replace the actions and the dismiss link with the "Thanks for your feedback" text


psealock (approved):
Awesome thanks! Does this need to be cherry picked as well?

becdetat:
> Does this need to be cherry picked as well?

No, the change that this fixes was merged only a couple of hours ago (#4516)

Issue #4603, opened 2020-06-16
Home screen styling issues


jameskoster:
In the interest of keeping this issue short(ish) I've written brief descriptions of the required changes. Hopefully they all make sense, but if more detail is required please don't hesitate to ask.

### Global

- All cards should have same border color, border radius, and padding values. If we can't use the [`Card`](https://wordpress.github.io/gutenberg/?path=/story/components-card--default) component, we should aim to match its styling consistently.
- There's a lot of inconsistencies in the text styling. Using the [`Text`](https://wordpress.github.io/gutenberg/?path=/story/components-text--default) component wherever possible would eliminate this, but I appreciate we might have to style things "manually" for now. I've included details of where we need to do this in the sections below. 
- Inbox tab in the Activity Panel should be hidden on the home screen

### Inbox

- Spacing in [header](https://cloudup.com/ctBQCm4qDqT) doesn't match [designs](https://cloudup.com/c3GmvprrGry) 
- "Inbox" header should use "Title Small" [`Text`](https://wordpress.github.io/gutenberg/?path=/story/components-text--default) style
- Secondary button color in notifications appears to be gray when it should be blue. Core [`Button`](https://wordpress.github.io/gutenberg/?path=/story/components-button--secondary) for reference.
- There should not be any primary buttons in Inbox notifications, only secondary and tertiary - https://cloudup.com/cvLEZnu52-M

### Task list (PR here: https://github.com/woocommerce/woocommerce-admin/pull/4627)

- I'm seeing [double ellipsis](https://cloudup.com/ceO7atYYQX4) (Brave, latest MacOS). For the "kebab" icon we should use the [core "more" icon](https://github.com/WordPress/gutenberg/blob/master/packages/icons/src/library/more.js)
- The [dismiss dropdown](https://cloudup.com/cPvhlR8-ZOt) styling is looking a bit funky. It should match the one in the [Inbox notifications](https://cloudup.com/cMQPPF1A5NY)
- Card header should use "Title Small" [`Text`](https://wordpress.github.io/gutenberg/?path=/story/components-text--default) style
- Card header padding should match the "medium" size in the core [`Card`](https://wordpress.github.io/gutenberg/?path=/story/components-card-header--default) component. 16px 24px. 
- Padding on task rows should be 16px 24px. That's my bad - the designs were incorrect.
- Task label should use "button" [`Text`](https://wordpress.github.io/gutenberg/?path=/story/components-text--default) style.
[...]

PR #4625, opened 2020-06-18
Stats Overview: Cleanup styles

psealock:
Partial https://github.com/woocommerce/woocommerce-admin/issues/4603

Fix Stats Overview in the Homescreen styles. I reused some styles across all cards on the Homescreen and place `<Text />` component in headers for Quick Links and Task List.

### Points addressed in this PR

- Card header should use "Title Small" [`Text`](https://wordpress.github.io/gutenberg/?path=/story/components-text--default) style
- Card header padding should match the "medium" size in the core [`Card`](https://wordpress.github.io/gutenberg/?path=/story/components-card-header--default) component. 16px 24px.
- Text color in tabs should match the core [`TabPanel`](https://wordpress.github.io/gutenberg/?path=/story/components-tabpanel--default) component
- Text styles and colors of the Jetpack banner should match the Inbox notifications, as should the spacing in the actions buttons.
- "Connect Jetpack" button should be secondary instead of primary
- "View detailed stats" link should not be pink.
- Ellipsis icon should use the [core "more" icon](https://github.com/WordPress/gutenberg/blob/master/packages/icons/src/library/more.js).

### Points not addressed in this PR

- There's a lag when clicking the toggles in the "Display stats" dropdown which makes them feel very odd to use. We should apply a loading state if we can't make this instant.

This is a known issue with the dataStore. I'd like to introduce optimistic results in a follow up PR. Issue here https://github.com/woocommerce/woocommerce-admin/issues/4626

- Spacing in the "Display stats" dropdown should match the [dismiss menu on Inbox notifications](https://cloudup.com/cMQPPF1A5NY), and the toggles should be blue

I'm reusing an often used component for this one, so it will be best implemented in a separate PR.

- Tabs should include comparison label - https://cloudup.com/ctRMMoD_UW7

Core tab-panel component only accepts strings at this time. An upstream PR will need to be made to make this happen.

### Screenshots

Screen Shot 2020-06-18 at 3 10 29 PM

1. Visit the Homepage. 
2. Ensure the points addressed in this PR are satisfied.

### Changelog Note:

none: unreleased feature.

### DIFF
--- a/client/homescreen/stats-overview/style.scss
+++ b/client/homescreen/stats-overview/style.scss
@@ -67,7 +67,7 @@ $promo-actions-border-bottom: $light-gray-tertiary;
 }
 
 article.woocommerce-stats-overview__install-jetpack-promo {
-	h3 {
+	h2 {
 		margin: $gap $gap-large $gap-smaller;
 	}
 	p {
--- a/client/homescreen/stats-overview/style.scss
+++ b/client/homescreen/stats-overview/style.scss
@@ -64,17 +64,6 @@ $promo-actions-border-bottom: $light-gray-tertiary;
 	}
 }
 
-
-.woocommerce-homescreen-card.woocommerce-stats-overview {
-	.woocommerce-summary__item {
-		padding: $gap-large;
-	}
-
-	.components-card__footer.is-size-large {
-		padding-left: $gap-smaller;
-	}
-}
-
 article.woocommerce-stats-overview__install-jetpack-promo {
 	.woocommerce-stats-overview__install-jetpack-promo__content {
 		padding: $gap $gap-large;
--- a/client/homescreen/stats-overview/style.scss
+++ b/client/homescreen/stats-overview/style.scss
@@ -74,15 +74,8 @@ article.woocommerce-stats-overview__install-jetpack-promo {
 		margin: 0 $gap-large $gap;
 	}
 	footer {
+		padding: $gap $gap-large;
 		border-top: 1px solid $promo-actions-border-top;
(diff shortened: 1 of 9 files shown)


client/homescreen/stats-overview/style.scss
```diff
@@ -12,10 +12,6 @@ $promo-actions-border-bottom: $light-gray-tertiary;
 	}
 }
 
-.woocommerce-stats-overview__footer {
-	border-top: 1px solid $outer-border;
-}
-
 .woocommerce-stats-overview__more-btn {
```
joshuatf, line 15: Minor, but while we're here, the more button is very slightly off-center.  Barely noticeable when focused, but I think it should be a quick fix.

joshuatf (approved):
Awesome work.  Really nice to see our efforts becoming more unified around WP components.

>This is a known issue with the dataStore. I'd like to introduce optimistic results in a follow up PR. Issue here #4626

Very excited for this!

jameskoster (approved):
I made some small styling adjustments, mostly to better justify text and align with other elements like the Inbox notifs. 

Before:

before

After:

after

The one thing I didn't touch that I think still needs fixing is the tertiary "no thanks" button. Looking at the [Storybook](https://wordpress.github.io/gutenberg/?path=/story/components-button--tertiary) I think that should be theme-colored. The color seemed to be coming from another component altogether though, so I didn't touch it here.

psealock:
Thanks for the review @joshuatf and @jameskoster for jumping in there.

I made some padding rules more global and added `isTertiary` to the "No thanks" button.

Screen Shot 2020-06-19 at 11 41 31 AM

PR #4627, opened 2020-06-18
Task List: Cleanup styles

psealock:
Depends on https://github.com/woocommerce/woocommerce-admin/pull/4625
Partial https://github.com/woocommerce/woocommerce-admin/issues/4603

- Padding on task rows should be 16px 24px. 
- Task label should use "button" [`Text`](https://wordpress.github.io/gutenberg/?path=/story/components-text--default) style.
- Time estimate color should be [$medium-gray-text](https://github.com/WordPress/gutenberg/blob/master/packages/base-styles/_colors.scss#L30)

Other items were fixed in the previous PR https://github.com/woocommerce/woocommerce-admin/pull/4625

### Screenshots

Screen Shot 2020-06-18 at 3 40 17 PM
Screen Shot 2020-06-18 at 3 40 43 PM

Visit the Homescreen's Task List and Quick Links. Make sure changes appear as described above.

### Changelog Note:

None: fixes unreleased feature.

### DIFF
--- a/packages/components/src/list/index.js
+++ b/packages/components/src/list/index.js
@@ -84,7 +84,7 @@ class List extends Component {
 								) }
 								<div className="woocommerce-list__item-text">
 									<span className="woocommerce-list__item-title">
-										<Text variant="button">{ title }</Text>
+										<Text as="div" variant="button">{ title }</Text>
 									</span>
 									{ content && (
 										<span className="woocommerce-list__item-content">
--- a/packages/components/src/list/index.js
+++ b/packages/components/src/list/index.js
@@ -5,6 +5,7 @@ import classnames from 'classnames';
 import { Component } from '@wordpress/element';
 import { ENTER } from '@wordpress/keycodes';
 import PropTypes from 'prop-types';
+import { __experimentalText as Text } from '@wordpress/components';
 
 /**
  * WooCommerce dependencies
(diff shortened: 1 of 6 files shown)


jameskoster (commented):
One small tweak - my bad for not including this detail in the issue:

When the task is _complete_ the label should be `Body Small` Text style and the color should be `$medium-gray-text`.

jameskoster:
I was able to fix the color. Looks like some sorcery would be required to swap the text style. If that's not a simple change I think we can omit it.

packages/components/src/list/index.js
```diff
@@ -83,7 +84,7 @@ class List extends Component {
 								) }
 								<div className="woocommerce-list__item-text">
 									<span className="woocommerce-list__item-title">
```
joshuatf, line 86: Since `<Text>` will render a `p` tag, we should probably update this to a `div` to be more semantic.

Side note that it would be nice if `Text` allowed `span` in the `as` prop and/or a `className` prop as well, but those are obviously upstream issues.

joshuatf (approved):
Nice clean-up, Paul!  Looks good and pre-approving pending one change.

>I was able to fix the color. Looks like some sorcery would be required to swap the text style. If that's not a simple change I think we can omit it.

Thanks for the touch up, @jameskoster!

psealock:
Thanks again @joshuatf and @jameskoster ! Here is the updated Task List with sorcery of text styles:

Screen Shot 2020-06-19 at 12 10 48 PM

PR #4630, opened 2020-06-18
Add check that the order being queried is a valid order

### LINKED ISSUE #4447: Add defensive type-checking for Reports Controller `get_order_number`
**Describe the bug**

Today while testing how the new draft order status introduced by WooCommerce Blocks 2.6.0 appears in the WC-Admin order list table (as a part of some documentation I'm writing for Woo Happiness), the order network request returned a 500:

Image 2020-05-26 at 7 48 56 PM

The stack trace was:

```
[26-May-2020 23:41:47 UTC] PHP Fatal error:  Uncaught Error: Call to a member function get_type() on boolean in /srv/www/woo/public_html/wp-content/plugins/woocommerce/packages/woocommerce-admin/src/API/Reports/Controller.php:188
Stack trace:
#0 /srv/www/woo/public_html/wp-content/plugins/woocommerce/packages/woocommerce-admin/src/API/Reports/Orders/Controller.php(85): Automattic\WooCommerce\Admin\API\Reports\Controller->get_order_number(1149)
#1 /srv/www/woo/public_html/wp-includes/rest-api/class-wp-rest-server.php(1015): Automattic\WooCommerce\Admin\API\Reports\Orders\Controller->get_items(Object(WP_REST_Request))
#2 /srv/www/woo/public_html/wp-includes/rest-api/class-wp-rest-server.php(342): WP_REST_Server->dispatch(Object(WP_REST_Request))
#3 /srv/www/woo/public_html/wp-includes/rest-api.php(306): WP_REST_Server->serve_request('/wc-analytics/r...')
#4 /srv/www/woo/public_html/wp-includes/class-wp-hook.php(287): rest_api_loaded(Object(WP))
#5 /srv/www/woo/public_html/wp-includes/class-wp-hook.php(311): WP_Hook->apply_filters(NULL, Array)
#6 /srv/www/ in /srv/www/woo/public_html/wp-content/plugins/woocommerce/packages/woocommerce-admin/src/API/Reports/Controller.php on line 188
```

The code producing the error is:

https://github.com/woocommerce/woocommerce-admin/blob/714660f67ee03ba94bc0febf30d2acbfe22bca7d/src/API/Reports/Controller.php#L185-L197

When troubleshooting the state of order number `1149` in my database, I noticed that it does not exist in the my db's `wp_post` table but it _does_ exist in the `wc_order_stats` table. So this is why the fatal is happening. As for why it doesn't exist in the `wp_post` table, a further clue is that the order status as indicated in the `wc_order_stats` table is `wc-checkout-draft`. So, some back story:

In WooCommerce 2.6.0 we introduced the `wc-checkout-draft` status for Woo orders because we now create them whenever the checkout is started via the checkout block. This is so that we can track various changes to the order via the new Store endpoint for orders during the checkout process. Effectively "draft" orders just mean that the checkout has been started by a shopper but has not been submitted yet. At the point of submission the normal statuses are applied to the order for its lifetime.

However, if a shopper _abandons_ checkout without submitting, the order stays in the database until it is cleaned up (deleted) by a scheduled action scheduler task every 24 hours:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/4de7cf163deef5168b053b050ddf35430e64c03e/src/Library.php#L133-L147

What we didn't anticipate, is that there may be plugins (like wc-admin) that have custom tables synced with the main table for orders that will not know that the orders have been deleted.

I'm still not sure long term what should happen here (I've created a corresponding issue in the blocks repo (https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2570), but for the short term, I've confirmed that if some defensive type checking is done in this method, it fixes things:

```php
public function get_order_number( $order_id ) {
	$order = wc_get_order( $order_id );

	if ( ! $order instanceof \WC_Order ) {
		return '';
	}

	if ( 'shop_order_refund' === $order->get_type() ) {
		$order = wc_get_order( $order->get_parent_id() );
	}

	if ( ! has_filter( 'woocommerce_order_number' ) ) {
		return $order->get_id();
	}

	return $order->get_order_number();
}
```

Result:

Image 2020-05-26 at 8 14 31 PM

On our end, we're going to have to implement a fix as well because this issue will surface on older versions of WooCommerce with blocks and WC-Admin (either in core or as a plugin) active but I _do_ think there still should be some defensive type checks in this code to prevent a site from bricking when there is no order for a given id. Probably worth a review to ensure anywhere else an id (or other lookup value) comes from the `wc_order_stats` table is used for retrieving an order (or other data) that there's checks to make sure the expected data exists as well.

**To Reproduce**

1. Activate [WooCommerce blocks 2.6.0](https://github.com/woocommerce/woocommerce-gutenberg-products-block/files/4678919/woocommerce-gutenberg-products-block.zip) (link to zip).
2. Add the checkout block to a page.
3. Add some products to your cart and visit the page with the checkout block.
4. I don't know when new orders are added to the `wc_stats_table` but however that happens, make sure the new draft order is in the table.
5. (to speed things up) manually delete the draft order from your database.
6. Visit the WC-Admin order list table route and you'll see the 500 on the network response.

Alternatively, if you don't mind potentially borking something else, you can just delete any order from the `wp_post` table and it will produce the fatal (if that order also exists in the `wc_order_stats` table.

becdetat:
Fixes #4447

This adds some defensive code when getting the order number within the report API controller.

- Before switching to this PR's branch, fInd a recent order (`SELECT * FROM `wp_posts` where post_type = 'shop_order' order by post_date desc`) and delete it, taking note of the post date
- Go to Analytics/Orders, and see the 500 error when trying to load the orders table
- Switch to this PR's branch
- Reload Analytics/Orders, and see that the table now loads

### Changelog Note:

Fix: Fix an error in the Analytics/Orders table when there is an order deleted directly from the database

### DIFF
--- a/src/API/Reports/Controller.php
+++ b/src/API/Reports/Controller.php
@@ -186,7 +186,7 @@ class Controller extends \WC_REST_Reports_Controller {
 		$order = wc_get_order( $order_id );
 
 		if ( ! $order instanceof \WC_Order ) {
-			return '';
+			return null;
 		}
 
 		if ( 'shop_order_refund' === $order->get_type() ) {
--- a/src/API/Reports/Controller.php
+++ b/src/API/Reports/Controller.php
@@ -185,6 +185,10 @@ class Controller extends \WC_REST_Reports_Controller {
 	public function get_order_number( $order_id ) {
 		$order = wc_get_order( $order_id );
 
+		if ( ! $order instanceof \WC_Order ) {
+			return '';
+		}
+
 		if ( 'shop_order_refund' === $order->get_type() ) {
 			$order = wc_get_order( $order->get_parent_id() );
 		}


src/API/Reports/Controller.php
```diff
@@ -185,6 +185,10 @@ public function get_items( $request ) {
 	public function get_order_number( $order_id ) {
 		$order = wc_get_order( $order_id );
 
+		if ( ! $order instanceof \WC_Order ) {
+			return '';
```
joshuatf, line 189: Minor nitpick: I think `null` might make checks here a little easier if we need them in the future.

joshuatf (approved):
Fixes the error!  One minor nitpick, but pre-approving.

src/API/Reports/Controller.php
becdetat: Nice, done in 2707160

joshuatf (approved):
Sweet- changes look good!  :shipit:

PR #4632, opened 2020-06-18
Fix inbox spacing

jameskoster:
I noticed we added some margin around Inbox notices for when they appear in the side panel. However this was also being applied on the home screen and throwing the layout off:

layoutbad

I fixed that, as well as the spacing and type styles on the Inbox header. Ideally we use the `Text` component for this in future, it took a bit of wrangling to get them appearing consistently, and duplicating the styles feelsbadman.jpg.

layoutgood

### DIFF
--- a/client/header/activity-panel/activity-header/style.scss
+++ b/client/header/activity-panel/activity-header/style.scss
@@ -29,23 +29,11 @@
 }
 
 .woocommerce-layout__inbox-panel-header {
-	padding: $gap * 2 42px $gap;
-
+	h2.woocommerce-layout__activity-panel-header-title,
 	h3.woocommerce-layout__activity-panel-header-title {
 		margin: 0 0 $gap-small / 2;
 		font-size: 20px;
 		line-height: 28px;
 		font-weight: 400;
-	}
-	span {
-		margin-left: 7px;
-		background: #40464d;
-		border-radius: 0.8em;
-		color: $core-grey-light-200;
-		display: inline-block;
-		text-align: center;
-		width: 1.6em;
-		font-size: 14px;
-		vertical-align: top;
 	}
 }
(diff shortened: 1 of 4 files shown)


client/header/activity-panel/activity-header/style.scss
```diff
@@ -29,10 +29,13 @@
 }
 
 .woocommerce-layout__inbox-panel-header {
-	padding: 15px 43px 10px;
-	h3 {
-		margin: 5px 0;
-		font-weight: 500;
+	padding: $gap * 2 42px $gap;
```
joshuatf, line 32: Sorry to dump this on you, but I think we could actually migrate the styles you made in `homescreen/style.scss` here and just add some margin around the inbox panel.

```scss
#activity-panel-inbox {
	margin: 0 20px;
}
```

This way we can keep the base styles the same and don't need to calculate/override padding differences each time we use this component.

client/homescreen/style.scss
```diff
@@ -70,3 +70,23 @@
 	height: 225px;
 	background-color: #6495ed;
 }
+
+
+.is-inbox {
```
joshuatf, line 75: Minor nitpick: not sure if we'll keep this with the above comment, but since this is a state class, it would be good to combine it with the selector its being applied to.  I.e., `.woocommerce-homescreen-column.is-inbox`

joshuatf (commented):
Thanks for the fixes here, @jameskoster!  This looks great and tests well.

I added a comment about moving your updates into the activity panel css and adding another block there.  Let me know if you have any questions.

jameskoster:
Good point. I have tidied the CSS up a bit.

joshuatf (approved):
Awesome work, @jameskoster!  Love all the red in this PR.  ❤️ LGTM!

timmyc:
[cherry picked to release/1.3.0-beta.1](https://github.com/woocommerce/woocommerce-admin/commit/893e6d4b3c6b12f210d9d436384cf169cdfa868b)

PR #4637, opened 2020-06-19
UserPrefs store: Introduce optimistic update

psealock:
Partial fix of https://github.com/woocommerce/woocommerce-admin/issues/4603

>There's a lag when clicking the toggles in the "Display stats" dropdown which makes them feel very odd to use. We should apply a loading state if we can't make this instant.

This was happening because the store wasn't updating until data returned from the server. This PR uses the data already available to update the store immediately so the UI doesn't need to wait for a round trip to the server.

The Dashboard suffered the same affliction but got around that by keeping duplicate data in state. Now that the issue is solved, this PR also removes state from the Dashboard in favor of a single source of truth in the UserPrefs store.

1. Go to Homescreen.
2. Toggle stats in the Stats overview section and see the UI update immediately.
3. Visit the Dashboard.
4. Edit titles, section order, chart type preference, and toggle stats on/off.
5. Be sure no regressions were introduced.

### Changelog Note:

none: unreleased change.

### DIFF
--- a/packages/data/src/user-preferences/use-user-preferences.js
+++ b/packages/data/src/user-preferences/use-user-preferences.js
@@ -91,10 +91,13 @@ export const useUserPreferences = () => {
 					};
 				}
 
-				// Optimistically propagate the woocommerce_meta to the store for instant update.
+				// Optimistically propagate new woocommerce_meta to the store for instant update.
 				receiveCurrentUser( {
-					id: user.id,
-					woocommerce_meta: metaData,
+					...user,
+					woocommerce_meta: {
+						...user.woocommerce_meta,
+						...metaData,
+					},
 				} );
 
 				// Use saveUser() to update WooCommerce meta values.
--- a/packages/data/src/user-preferences/use-user-preferences.js
+++ b/packages/data/src/user-preferences/use-user-preferences.js
@@ -120,9 +120,6 @@ export const useUserPreferences = () => {
 					};
 				}
 
-				// Propagate the updated User object to the store.
-				receiveCurrentUser( updatedUser );
-
 				// Decode the WooCommerce meta after save.
 				const updatedUserResponse = {
 					...updatedUser,
--- a/packages/data/src/user-preferences/use-user-preferences.js
+++ b/packages/data/src/user-preferences/use-user-preferences.js
@@ -91,6 +91,12 @@ export const useUserPreferences = () => {
 					};
 				}
 
+				// Optimistically propagate the woocommerce_meta to the store for instant update.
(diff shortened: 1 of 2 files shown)


packages/data/src/user-preferences/use-user-preferences.js
```diff
@@ -91,6 +91,15 @@ export const useUserPreferences = () => {
...
+				receiveCurrentUser( {
+					...user,
+					woocommerce_meta: {
+						...user.woocommerce_meta,
+						...metaData,
+					},
+				} );
+
 				// Use saveUser() to update WooCommerce meta values.
 				const updatedUser = await saveUser( {
```
joshuatf, line 104: Would it be possible to only call `receiveCurrentUser` once and not again with `receiveCurrentUser( updatedUser )` once updated if no errors are encountered?

It might also be nice (in a follow-up possibly) to do some more error handling and create a notice if preferences aren't being saved.

joshuatf (commented):
Nice work on this, Paul!  I noticed some odd behaviors that occur on updating the store after the request completes, especially if options are toggled "too quickly":

ezgif-5-02cc3bb80201

Left a comment in the code about a possible fix, but let me know what you think or if you think this needs to be handled in a follow-up.

packages/data/src/user-preferences/use-user-preferences.js
psealock: Good catch @joshuatf !

It definitely makes sense not to save twice due to the possibility of fast clicking. I removed that in https://github.com/woocommerce/woocommerce-admin/pull/4637/commits/11fb920dc4f97171f3f3c59572ee0e91a8d05600.

As it stands now, if an error occurs its returned from `updateUserPreferences` so an async function can choose to handle it however it wants.

Alternatively, we can fire off a core notice if there was an error. The UI would update, and a notice would appear once the error is returned.

joshuatf: >As it stands now, if an error occurs its returned from updateUserPreferences so an async function can choose to handle it however it wants.

I do like the idea of a component choosing how it handles error messaging.

Maybe one piece of error handling that belongs in the store is reverting the options with `receiveCurrentUser()` if an error does occur.  And then continuing to pass the error message along so the component can choose how to handle the UI aspect of it.

But this is hopefully an edge case and I'm good if we handle this in a follow-up.

joshuatf (approved):
Really nice addition and improvement to the UI, Paul!

Left one more thought on error handling, but I think it's perfectly fine to handle in a follow-up if you agree with the idea.

PR #4642, opened 2020-06-19
Don't include sourcemaps or unminified JS for "core" builds.

### LINKED ISSUE #4634: Build: Remove Un-Minified Assets From Core Build
Per discussion in slack with @peterfabian - inclusion of the unminified assets in the "core"/production build of WooCommerce Admin is close to doubling the size of the build vs previous versions.

Because WooCommerce core is already a quite large .zip, there is concern that it the overall package size grows too large, some users might encounter memory issues when installing the plugin - so it is important to keep our build size down as much as possible.

The solution we wanted to explore was:

- Remove un-minifed assets from the core build
- And if possible, retain them in the plugin build

For time sake on 1.3/4.3 - hitting the first item above should be the priority.

/cc @jeffstieler

jeffstieler:
Fixes #4634.

This PR avoids creating JS source maps and unminified JS files when building for `core`. Ultimately this is meant to avoid these files ending up in the core WooCommerce plugin zip file.

- Delete your `dist/` folder
- Run `WC_ADMIN_PHASE=core npm run build`
- Verify no `.map` files or JS files without `.min.js` exist in `/dist`
- Run `npm start`
- Verify there are `.map` files and JS files without `.min.js` in `/dist`

cc: @peterfabian

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -19,6 +19,7 @@ const UnminifyWebpackPlugin = require( './unminify' );
 const CustomTemplatedPathPlugin = require( '@wordpress/custom-templated-path-webpack-plugin' );
 
 const NODE_ENV = process.env.NODE_ENV || 'development';
+const WC_ADMIN_PHASE = process.env.WC_ADMIN_PHASE || 'development';
 
 const externals = {
 	'@wordpress/api-fetch': { this: [ 'wp', 'apiFetch' ] },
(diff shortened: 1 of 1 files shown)


mattsherman:
I'll review this.

mattsherman (approved):
Tested according to plan and results as expected.

Confidence-check running `npm start` worked as well.

Approved!

PR #4643, opened 2020-06-19
Added event prop 'screen' to be recorded

### LINKED ISSUE #4633: Missing event prop when clicking in an inbox notification within the Home screen
I was testing the new Home screen and I think that's missing the event prop: `screen` (value: home screen, analytics_revenue, customers, etc) on `wcadmin_wcadmin_inbox_action_click` (see #4099 for more context)

octaedro:
Fixes #4633

The event prop `screen` wasn't being sent with the event `wcadmin_wcadmin_inbox_action_click`.
This PR adds the event prop `screen` to `wcadmin_wcadmin_inbox_action_click`.
This change will apply for all the call to action buttons of the admin notes.

### Accessibility

- Open the Inbox Panel, choose a message, and press its call to action button.
- Verify that the screen name has been sent as a prop in the event `wc_admin_record_tracks_event` when a note's action is triggered. ([Here](https://github.com/woocommerce/woocommerce-admin/blob/16770825fe26015e32f52c75a8331502638af880/src/API/NoteActions.php#L133) is where we send it).
- Go to another page (E.g. the Analytics Overview, URL ->`/wp-admin/admin.php?page=wc-admin&path=%2Fanalytics%2Foverview`).
- Verify the sent screen name corresponds with the page name.
- Also, verify the screen name is sent with other types of admin notes.
E.g.:
screenshot-one wordpress test-2020 06 19-13_17_53

- If you don't have other types of admin notes available to test, you can add one executing these two sentences in the DB:
````
INSERT INTO `wp_wc_admin_notes` (`note_id`, `name`, `type`, `locale`, `title`, `content`, `content_data`, `status`, `source`, `date_created`, `date_reminder`, `is_snoozable`, `layout`, `image`, `is_deleted`, `icon`) VALUES (9999, 'wc-admin-test-note', 'update', 'en_US', 'Lorem ipsum note', 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed eiusmod tempor incidunt ut labore et dolore magna aliqua. Ut enim ad minim veniam.', '{}', 'unactioned', 'woocommerce-admin', '2020-06-17 19:22:48', NULL, 0, 'plain', NULL, 0, 'info');
````
and
````
INSERT INTO `wp_wc_admin_note_actions` (`note_id`, `name`, `label`, `query`, `status`, `is_primary`, `actioned_text`) VALUES (9999, 'learn-more', 'Learn more', '?page=wc-addons&section=helper', 'unactioned', 0, '');
````

### Changelog Note:
Fix: Added event prop `screen` to `wcadmin_wcadmin_inbox_action_click`.

### DIFF
--- a/src/API/NoteActions.php
+++ b/src/API/NoteActions.php
@@ -27,7 +27,7 @@ class NoteActions extends Notes {
 	public function register_routes() {
 		register_rest_route(
 			$this->namespace,
-			'/' . $this->rest_base . '/(?P<note_id>[\d-]+)/action/(?P<action_id>[\d-]+)/screen/(?P<screen>[a-z_-]+)',
+			'/' . $this->rest_base . '/(?P<note_id>[\d-]+)/action/(?P<action_id>[\d-]+)',
 			array(
 				'args'   => array(
 					'note_id'   => array(
@@ -119,7 +119,21 @@ class NoteActions extends Notes {
 			$tracks_event = 'wcadmin_inbox_action_click';
 		}
 
-		$screen_name = $request->get_param( 'screen' );
+		if ( isset( $_SERVER['HTTP_REFERER'] ) ) {
+			parse_str( wp_parse_url( wp_unslash( $_SERVER['HTTP_REFERER'] ), PHP_URL_QUERY ), $queries );
+		}
+		if ( isset( $queries ) ) {
+			$page      = isset( $queries['page'] ) ? $queries['page'] : null;
+			$path      = isset( $queries['path'] ) ? $queries['path'] : null;
+			$post_type = isset( $queries['post_type'] ) ? $queries['post_type'] : null;
+		}
+
+		if ( isset( $page ) ) {
+			$current_page = 'wc-admin' === $page ? 'home_screen' : $page;
+			$screen_name  = isset( $path ) ? substr( str_replace( '/', '_', $path ), 1 ) : $current_page;
+		} elseif ( isset( $post_type ) ) {
+			$screen_name = $post_type;
+		}
 
 		wc_admin_record_tracks_event(
 			$tracks_event,
--- a/src/API/NoteActions.php
+++ b/src/API/NoteActions.php
@@ -27,7 +27,7 @@ class NoteActions extends Notes {
 	public function register_routes() {
 		register_rest_route(
 			$this->namespace,
-			'/' . $this->rest_base . '/(?P<note_id>[\d-]+)/action/(?P<action_id>[\d-]+)',
+			'/' . $this->rest_base . '/(?P<note_id>[\d-]+)/action/(?P<action_id>[\d-]+)/screen/(?P<screen>[a-z_-]+)',
 			array(
 				'args'   => array(
 					'note_id'   => array(
@@ -37,6 +37,10 @@ class NoteActions extends Notes {
 					'action_id' => array(
 						'description' => __( 'Unique ID for the Note Action.', 'woocommerce-admin' ),
 						'type'        => 'integer',
+					),
+					'screen'    => array(
+						'description' => __( 'The name of the screen where the action was triggered.', 'woocommerce-admin' ),
+						'type'        => 'string',
 					),
 				),
 				array(
@@ -115,6 +119,8 @@ class NoteActions extends Notes {
 			$tracks_event = 'wcadmin_inbox_action_click';
 		}
 
+		$screen_name = $request->get_param( 'screen' );
+
 		wc_admin_record_tracks_event(
 			$tracks_event,
 			array(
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/API/NoteActions.php [hunk 1]
```diff
@@ -115,6 +119,8 @@ public function trigger_note_action( $request ) {
 			$tracks_event = 'wcadmin_inbox_action_click';
 		}
 
+		$screen_name = $request->get_param( 'screen' );
```
joshuatf, line 122: Could we not grab the referer from the request in here instead of manually sending it from the client?

src/API/NoteActions.php
```diff
@@ -27,7 +27,7 @@ class NoteActions extends Notes {
 	public function register_routes() {
 		register_rest_route(
 			$this->namespace,
-			'/' . $this->rest_base . '/(?P<note_id>[\d-]+)/action/(?P<action_id>[\d-]+)',
+			'/' . $this->rest_base . '/(?P<note_id>[\d-]+)/action/(?P<action_id>[\d-]+)/screen/(?P<screen>[a-z_-]+)',
```
joshuatf, line 30: If for some reason we can't use the referer, I think it might be better to include these params in the body of the request as opposed to the URL.

joshuatf (commented):
Very creative solution for tackling this, Fernando!  Left a couple comments that may possibly make this easier to manage, but let me know what you think.

src/API/NoteActions.php
octaedro: Hi @joshuatf, thanks for the review.
I chose to send the param to try to have only one method where the screen names come from, but the solution you mentioned (take the referer from the request) is better. I applied it in the commits 8af0e02 and 3ff440f
In this PR I also moved the method `getScreenName` to a more general `utils` file. Now it is not directly affecting this PR but I'll leave this method there because I think that it's a better place for it.

src/API/NoteActions.php
```diff
@@ -115,6 +115,24 @@ public function trigger_note_action( $request ) {
 			$tracks_event = 'wcadmin_inbox_action_click';
 		}
 
+		$screen_name = '';
```
joshuatf, line 118: Optional, but we can extract this block to its own function which might make it easier to follow or reuse going forward.  E.g., `get_screen_name()`.

src/API/NoteActions.php
```diff
@@ -115,6 +115,24 @@ public function trigger_note_action( $request ) {
...
+
+		if ( isset( $_SERVER['HTTP_REFERER'] ) ) {
+			parse_str( wp_parse_url( $_SERVER['HTTP_REFERER'], PHP_URL_QUERY ), $queries ); // phpcs:ignore sanitization ok.
+		}
+		if ( isset( $queries ) ) {
+			$page      = isset( $queries['page'] ) ? $queries['page'] : null;
+			$path      = isset( $queries['path'] ) ? $queries['path'] : null;
+			$post_type = isset( $queries['post_type'] ) ? $queries['post_type'] : null;
```
joshuatf, line 126: This works well for post type list page, but we won't be able to see the type from individual posts this way.  For example, clicking into an individual order or product and actioning a note, the `screen` property will be blank.

I'm not sure how important this is for tracking (feel free to weigh in @pmcpinto), but if we need to solve this, we could grab the post type from the ID.

```php
get_post_type( $queries['post'] );
```

joshuatf (commented):
Awesome- nice refactor, @octaedro!  This is testing pretty well for me; left one comment about individual posts and a possible fix if we need to include these for tracking purposes.

src/API/NoteActions.php
octaedro: Good idea! Sounds good to me, I modified this in the commit e5fdd11

src/API/NoteActions.php
octaedro: Nice catch @joshuatf! Thank you 🙂
I added the method you mentioned in the commit 2636420

pmcpinto:
> I'm not sure how important this is for tracking (feel free to weigh in @pmcpinto), but if we need to solve this, we could grab the post type from the ID.

I'm ok with skipping the identification of specific products/orders. @octaedro @timmyc can this be included on the 4.3 release?

src/API/NoteActions.php
joshuatf: >In this PR I also moved the method getScreenName to a more general utils file. Now it is not directly affecting this PR but I'll leave this method there because I think that it's a better place for it.

Nice clean up!  Agreed, good change here.

joshuatf (approved):
Looks great and is testing well!  I'm sure the screen event will prove very useful.  Nice job, @octaedro 🎉

timmyc:
cherry-picked to `release/1.3.0-rc.2`

PR #4647, opened 2020-06-20
Fix WCPay sometimes not appearing on the task list

### LINKED ISSUE #4628: [4.3] WooCommerce Payments card not displayed in task list
**Describe the bug**

With 4.3-beta-1, the WooCommerce Payments card is not displayed on first load of the payment setup task when Jetpack is not installed during the new store onboarding.

**To Reproduce**

1. Complete all 5 steps of the new store setup
1. Click **No thanks** for WCS & Jetpack
1. Click **Set up payments** task under **Finish setup**
1. Notice no WC Pay card is displayed
1. Refresh the page
1. Notice the WC Pay card is displayed

Video demo: https://d.pr/v/JQYnNn

**Expected behavior**

I expected the WooCommerce Payments card to be displayed as the first option for payment methods to a new store in the US not selling CBD products, even without the Jetpack plugin and connection.

**Screenshots**

image

* OS: MacOS
* Browser: Firefox
* Version: 76.0.1

DanReyLop:
Fixes #4628 
Fixes https://github.com/woocommerce/woocommerce/issues/26812

Fixes a bug in the Payments task list that made WCPay not appear on a clean install.

Many payment methods are region-specific, so the setup checklist needs to know which country the store is setup on to show/hide some of the payment options. For example, WCPay is only available in the US.

The Payments setup screen was using `getOption('woocommerce_default_country')` to get the store country. The problem is that the value of `getOption(...)` doesn't change until the page is refreshed. So, if a user goes through the whole Setup Wizard and lands in the Payments screen without ever refreshing the page (entirely possible since all that flow uses client-side navigation), then whatever address they put in the "Store Details" step won't be picked up by the Payments page logic.

By default, new WooCommerce stores are set in `GB` (Great Britain). So, if a user starts with a clean site and goes through the whole setup wizard, the options they'll see in the Payments page will be as if they were in the UK (Stripe and PayPal, but no WCPay).

Other steps of the wizard that rely on country data (for example, to show or hide the "CBD" option) use the `getSettings('general')` API instead, which is automatically updated without requiring a page reload. So in this PR I do exactly that for the Payments page. I added a fallback to `getOption(...)` so the user doesn't see the wrong payments for a second if they go directly to the Payments page.

Follow the steps to reproduce [here](https://github.com/woocommerce/woocommerce/issues/26812).

Alternatively, this is an easier way I found to trigger the bug:
- Set your store address to Spain, for example (or South Africa if you want to see the Stripe card dissapear too).
- Go to the Setup Wizard.
- On the first step, change your store address to someplace in the US.
- **Without reloading the page**, go through the whole setup wizard (you can pretty much skip all the other steps). You'll be dropped in the setup checklist.
- Click on `Set up payments`. You won't see WCPay (and, if you originally set your address to South Africa, you won't see Stripe either).
- Refresh the page. WCPay (and Stripe) will appear, but at this point the user is confused and sad.

If you repeat the steps but reverse the countries (start with a `US` store, change it to Spain in the wizard), you'll see the same problem in reverse: WCPay appears when it shouldn't.

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -40,6 +40,13 @@ if ( appRoot ) {
 	let HydratedPageLayout = withSettingsHydration( settingsGroup, window.wcSettings )(
 		PageLayout
 	);
+	const hydrateSettings =
+	window.wcSettings.preloadSettings &&
+	window.wcSettings.preloadSettings.general;
+
+	if ( hydrateSettings ) {
+		HydratedPageLayout = withSettingsHydration( 'general', { general: window.wcSettings.preloadSettings.general, } )( HydratedPageLayout );
+	}
 	if ( hydrateUser ) {
 		HydratedPageLayout = withCurrentUserHydration( hydrateUser )( HydratedPageLayout );
 	}
--- a/client/profile-wizard/index.js
+++ b/client/profile-wizard/index.js
@@ -14,7 +14,6 @@ import { updateQueryString } from '@woocommerce/navigation';
 import {
 	ONBOARDING_STORE_NAME,
 	PLUGINS_STORE_NAME,
-	withSettingsHydration,
 	withPluginsHydration,
 } from '@woocommerce/data';
 
@@ -240,10 +239,6 @@ class ProfileWizard extends Component {
 	}
 }
 
-const hydrateSettings =
-	window.wcSettings.preloadSettings &&
-	window.wcSettings.preloadSettings.general;
-
 export default compose(
 	withSelect( ( select ) => {
 		const { getNotes } = select( 'wc-api' );
@@ -279,11 +274,6 @@ export default compose(
 			updateProfileItems,
 		};
 	} ),
-	hydrateSettings
-		? withSettingsHydration( 'general', {
-				general: window.wcSettings.preloadSettings.general,
-		  } )
-		: identity,
 	window.wcSettings.plugins
 		? withPluginsHydration( {
 				...window.wcSettings.plugins,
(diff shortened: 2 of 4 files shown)


DanReyLop:
> I think we can take this a step further and remove options.woocommerce_default_country as the backup because it should be used at all.

I tried that, but then if you navigate to the checklist **without** going through the Setup Wizard first, then there's 1 second or so when `getSettings('general')` hasn't loaded yet so neither Stripe or WCPay are displayed. That's a poor UX. I think having `getOption` as a fallback is a decent compromise this late in the release process.

What do you think the ideal solution would be? Would preloading `settings.general` into `wcSettings` work?

psealock:
>What do you think the ideal solution would be? Would preloading settings.general into wcSettings work?

Yes, exactly. I went ahead and made those changes in https://github.com/woocommerce/woocommerce-admin/pull/4647/commits/b9bc9e141d3c26b8f15f59e9229d0897b1ce09f7.

General settings are now preloaded for the entire app, not just the OBW. I also removed the usage of `options.woocommerce_default_country` to remain consistent. 

### Additional testing instructions

Test the delay mentioned above in https://github.com/woocommerce/woocommerce-admin/pull/4647#issuecomment-647410020 is no longer there.

1. Be sure to have a US based store.
1. Enable the task list.
2. Go to `/wp-admin/admin.php?page=wc-admin&step=theme&task=payments`.
3. See the WC Pay option appear immediately. Previously it appeared after a delay while the setting was being fetched.

DanReyLop:
> Yes, exactly. I went ahead and made those changes in b9bc9e1.

Thanks! This is a better solution for sure :)

PR #4650, opened 2020-06-22
Fix reselecting advanced filters in the customer list

### LINKED ISSUE #4423: Re-selecting Advanced filters on a report clears already applied filters
**Describe the bug**
If the Advanced filters option is already selected but is clicked a second time, already applied advanced filters are reset.

**To Reproduce**

1. Navigate to any report that has an Advanced filters option
2. Select the `Advanced filters` option.
3. Add a filter and apply it.
4. Select the `Advanced filters` again.
5. Previously selected filters are gone

**Expected behavior**
Already applied advanced filters should still be there if the same filter option is selected.

**Screenshots**
https://d.pr/i/EKWh6o

h/t @vbelolapotkov

becdetat:
Fixes #4423

Reselecting advanced filters in the customer list (WooCommerce -> Customers) was clearing all selected filters. This fixes that by adding the advanced args to the config.

It also goes through the rest of the analytics screens that include an "Advanced" or "Comparison" option.

While testing this I noticed console errors regarding missing `key` values for loaded labels, this PR also makes sure there is a unique ID used as the key.

- Open WooCommerce -> Customers
- Set up an advanced filter, make sure to press the "Filter" button to apply it
- Select the filter type dropdown and reselect "Advanced"
- Observe that the advanced filters remain
- Select the filter type dropdown and select a different filter type, then reselect Advanced
- Observe that the advanced filters remain
- Refresh the page
- Observe that the advanced filters remain
- There should be no JS errors in the console relating to missing `key` values

Also test:
- Analytics -> Orders -> Advanced
- Analytics -> Products -> Comparison
- Analytics -> Categories -> Comparison
- Analytics -> Coupons -> Comparison
- Analytics -> Taxes -> Comparison
- Analytics -> Downloads -> Advanced


psealock (commented):
Nice work here, I know that must've been difficult to wade through the Advanced Filter stuff. I think the advanced filters in the Orders screen would need the same treatment.

One alternative thought (I'm not sure the consequences): could clicking on "Advanced Filters" first check to see if advanced filters are already shown? And if so, just close the dropdown without navigation?

becdetat:
Thanks @psealock , not navigating if the button is already selected was a great idea - I implemented it in f5ef0dd.

I still needed the `staticParams` in the report/analytics config though, otherwise when going back to "All Customers" then back to "Advanced Filters" would lose the filters.

While looking at the Orders screen I noticed that many of the other analytics screens had a similar 'Comparison' or 'Advanced' option, so I did the config magic on them too so they all should work properly now.

jeffstieler:

mattsherman (approved):
Tested and works as expected. Nice!

Issue #4579, opened 2020-06-15
Store Profiler - Business details step: install free extensions in 1 click


pmcpinto:
**Description:** we're going to experiment with a new approach to install free extensions in the Store Profiler flow, including WCPay. We should start with a small subset of new users - it seems that the easiest way can be filtering the segment by the store location and industry. Eg: stores with the location in the US and Electronics & Computers as industry. 

Final segmentation: Stores that selected the US as country and Fashion, Apparel, Accessories or Health & Beauty as Industry. It gives ~2.5% of the total number of new users and ~30% of the total number of new users from US

![features-3.png](https://images.zenhubusercontent.com/5d84fe8e1201de0001b6ab61/6183de02-628d-4471-a57b-723079c465c0)

**Acceptance Criteria:**
- [ ] display the changes in the flow just to a specific segment (segmentation tbd)
- [ ] display just one checkbox to install all the free extensions. The checkbox should be toggled on by default
- [ ] show a popover with more info about the extensions when the user hovers the info icon
- [ ] install the following extensions if the user clicked "continue" with the checkbox togged on: Jetpack, WooCommerce Services, Facebook, Mailchimp, Google Ads, WCPay 
- [ ] remove the JP/WCS step for this segment, as they're going to be installed in the Business details step
- [ ] the redirect to the Jetpack account creation flow should only happen after the theme step 

**Event tracking:**
It will be better to add a new event for this step's alternative version:

Event name: `wcadmin_storeprofiler_store_business_details_continue_alternative`
Event prop: `extensions_installed` (value: Yes/No)
Event description: When the user clicks on "Continue" on the alternative version of the Business step
 
Event name: `wcadmin_storeprofiler_store_business_details_popover_alternative`
Event prop: NA
Event description: When the user clicks on the popover

It will be necessary to update `wcadmin_storeprofiler_step_view` to include this alternative step as the value of the `step` prop. It can be named business-details-alternative

**Additional links:**
[Design mockup](https://www.figma.com/file/JH9XMFUCOjfXdr3N09AHRD/On-boarding-iterations-June-'20?node-id=1%3A3)

joshuatf:
Estimate: 5. I see two items that stand out in this issue:

1. Since we're introducing split-testing in WCA, it might be nice to have some generic functions around this so we don't recreate the wheel each time we do this.
[...]

PR #4655, opened 2020-06-22
Refactor Jetpack connection flows

### LINKED ISSUE #4639: Jetpack connection URL should be resolved with promise
In following https://github.com/woocommerce/woocommerce-admin/pull/4473, we should move away from using the `Connect` component where resolving the connection URL is needed without the added UI.

This currently happens on the benefits screen and the install Jetpack CTA.

In order to do this, we may want to look into using `__experimentalResolveSelect` (which seems relatively safe, but needs testing) and push changes upstream in gutenberg to move this away from experimental.  See p1592484873191100-slack-woo-merchandising for discussion on this.

* Remove `Connect` component on Benefits page and install Jetpack CTA.
* Replace with `__experimentalResolveSelect` to get the Jetpack connect URL.
* Remove `autoConnect` logic from `Connect`.
* Open an issue upstream in GB if the experimental selector works well to move this to non-experimental.

joshuatf:
Fixes #4639 

To make the Jetpack connection flow easier to follow, this PR uses the experimental selector in wp data to resolve the connection URL instead of relying on a UI component to handle the connection.

It also removes the `autoConnect` feature to discourage use of this component when an interface is not part of the connection flow.

This will also be helpful for upcoming changes to the onboarding flow noted in https://github.com/woocommerce/woocommerce-admin/issues/4579

**Feedback:** I think that being able to catch errors from the resolver would be a great addition to this, but this would need to be something we handle upstream.

1. Deactivate WCS and/or Jetpack.
1. Walk through the profiler, opting in at the benefits screen.
1. Make sure you are redirected to the Jetpack connection (note that you may need to be a non-local server to fully test the Jetpack connection flow).
1. Repeat this process with the install Jetpack CTA on the homescreen stats section.
1. Optionally throw a `WP_Error` in https://github.com/woocommerce/woocommerce-admin/blob/9213826b60d1255c72a8f5138a9e74cec2b662f0/src/API/Plugins.php#L440 and make sure its message is shown when attempting to connect in either of these places.

### DIFF
--- a/client/dashboard/components/connect/index.js
+++ b/client/dashboard/components/connect/index.js
@@ -24,21 +24,11 @@ class Connect extends Component {
 		props.setIsPending( true );
 	}
 
-	componentDidMount() {
-		const { autoConnect, jetpackConnectUrl } = this.props;
-
-		if ( autoConnect && jetpackConnectUrl ) {
-			this.connectJetpack();
-		}
-	}
-
 	componentDidUpdate( prevProps ) {
 		const {
-			autoConnect,
 			createNotice,
 			error,
 			isRequesting,
-			jetpackConnectUrl,
 			onError,
 			setIsPending,
 		} = this.props;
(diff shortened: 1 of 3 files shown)


mattsherman:
I'll review this

mattsherman (approved):
Tested connecting JP from both profiler and home screen stats overview. Works as expected.

Returned an error in `connect_jetpack()` and verified the error was shown correctly.

Nice work!

PR #4664, opened 2020-06-23
Fix dashboard task list menu placement

### LINKED ISSUE #4663: Task list ellipsis menu icon style issue
The task list ellipsis menu icon is centred, but it should be right-aligned.

**Screenshots**
image

### LINKED ISSUE #4675: Separate the home screen and dashboard styles
Per @joshuatf 's comments in #4664, we should separate the home screen and dashboard styles.

> Confirming that this fixes the issue, however I really think we should be separating out those styles specific to the homescreen and those used for the dashboard.

> This probably means that `woocommerce-homescreen-card` should be renamed to something like `woocommerce-dashboard-card` and separated into a dashboard stylesheet.

becdetat:
so it is styled correctly in the dashboard view.

Fixes #4663
Fixes #4675

The ellipsis menu on the task list in dashboard view is misaligned. This is because the new home screen styles weren't imported to the task list on the dashboard view.

### Screenshots
Before:
image

After:
image

- Configure the home screen off in `/config/development.json` and restart `npm start`, or just build a zip from this branch and upload it to a fresh ephemeral site with the `woocommerce_task_list_hidden` option set to `no`
- Check the dashboard task menu style

### DIFF
--- a/client/task-list/index.js
+++ b/client/task-list/index.js
@@ -32,7 +32,6 @@ import {
  * Internal dependencies
  */
 import './style.scss';
-import '../homescreen/style.scss';
 import CartModal from 'dashboard/components/cart-modal';
 import { getAllTasks } from './tasks';
 import { recordEvent } from 'lib/tracks';
@@ -359,7 +358,7 @@ class TaskDashboard extends Component {
 						<Fragment>
 							<Card
 								size="large"
-								className="woocommerce-task-card woocommerce-homescreen-card"
+								className="woocommerce-task-card woocommerce-dashboard-card"
 							>
 								<progress
 									className={ progressBarClass }
--- a/client/task-list/index.js
+++ b/client/task-list/index.js
@@ -32,6 +32,7 @@ import {
  * Internal dependencies
  */
 import './style.scss';
+import '../homescreen/style.scss';
 import CartModal from 'dashboard/components/cart-modal';
 import { getAllTasks } from './tasks';
 import { recordEvent } from 'lib/tracks';
@@ -314,7 +315,12 @@ class TaskDashboard extends Component {
 			);
 
 			task.title = (
-				<Text as="div" variant={ task.completed ? 'body.small' : 'button' }>{ task.title }</Text>
+				<Text
+					as="div"
+					variant={ task.completed ? 'body.small' : 'button' }
+				>
+					{ task.title }
+				</Text>
 			);
 
 			if ( ! task.completed ) {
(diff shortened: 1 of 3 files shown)


client/task-list/index.js
```diff
@@ -32,6 +32,7 @@ import {
  * Internal dependencies
  */
 import './style.scss';
+import '../homescreen/style.scss';
```
joshuatf, line 35: Confirming that this fixes the issue, however I really think we should be separating out those styles specific to the homescreen and those used for the dashboard.

This probably means that `woocommerce-homescreen-card` should be renamed to something like `woocommerce-dashboard-card` and separated into a dashboard stylesheet.

joshuatf (approved):
This is testing well to fix the issue.  Pre-approving, but I think we should create an issue around separating these styles prior to merging so we can follow-up on this.

becdetat:
@joshuatf I did the followup in this PR as it was easier (and I can't merge yet anyway because of travis issues). Thanks for the suggestion :-)

jeffstieler:
The merge blocking is my fault, sorry! I've removed that requirement from the `main` branch for now while I work with Travis support to restore our builds.

client/task-list/style.scss
```diff
@@ -584,3 +584,27 @@
 		border-top-right-radius: 2px;
 	}
 }
+
+.woocommerce-dashboard-card {
```
joshuatf, line 588: Let's move this to `dashboard/style.scss` since it's more generically used across dashboard items.

There's also one more reference to `woocommerce-homescreen-card` that we can update to `woocommerce-dashboard-card` and then we can remove the style block for `.woocommerce-homescreen-card`.

joshuatf (commented):
>I did the followup in this PR as it was easier (and I can't merge yet anyway because of travis issues). Thanks for the suggestion :-)

Sounds good :)

One more suggestion and I think this is good to go.

timmyc:
Going to close/reopen to get CI happy

timmyc:
hi :)

timmyc:
@becdetat since I commandeered this ship via the force push, I figured the polite thing to do was to also handle the feedback from Josh 😂 - could you verify all looks good here, and then we can merge and pick. Thanks!

becdetat:
Looks good, I'll merge it now

timmyc:
cherry picked to `release/1.3.0-rc.1` in f7a6deb34af4c950029973fa9d3a756b6d763ea5

PR #4666, opened 2020-06-23
Restore unread indicator styling

becdetat:
The unread admin notes count indicator's style seem to have been lost at some point recently. This restores it.

### Screenshots
Before:
image

After:
image

- Look at the home page with some unread messages. The unread count indicator should be styled like a pill.

### DIFF
--- a/client/header/activity-panel/activity-header/style.scss
+++ b/client/header/activity-panel/activity-header/style.scss
@@ -47,4 +47,17 @@
 			vertical-align: top;
 		}
 	}
+
+	span {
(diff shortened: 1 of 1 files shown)


joshuatf (approved):
Looks like it was accidentally removed in https://github.com/woocommerce/woocommerce-admin/pull/4632- thanks for catching this @becdetat!  LGTM :shipit:

timmyc:
Doh! Looks like both @octaedro and I missed this one, and shipped #4690 instead. Thanks for the work here y'all and apologies I didn't see it.

PR #4668, opened 2020-06-23
Add version 1.3.0 option update callback.

jeffstieler:
The callback for updating the DB version to `1.3.0` (necessary when performing other callback, for now) was missed.

- Check out this branch
- Verify that a DB update action is scheduled
- Verify that the DB version is updated

### Changelog Note:

N/A - unreleased bug


jeffstieler:
@becdetat right now we only need to write a version bump callback if there are other callbacks. In the case of no callbacks, the version is bumped automatically: https://github.com/woocommerce/woocommerce-admin/blob/main/src/Install.php#L383

PR #4671, opened 2020-06-23
Fix Historical (deleted) Coupon Reporting

### LINKED ISSUE #4657: Coupon sync doesn't handle historical coupon data correctly
**Describe the bug**

When order coupon data is synced to the `wc_order_coupon_lookup` table, it only includes coupons that are of status `publish` **at the time the sync occurs**. This leads to inaccurate reporting which can be easily seen when comparing to the old WC reports. It also creates a vector to destroy good data if a Historical Data Import is performed after some coupons are deleted or changed from the `publish` status.

We also delete rows from `wc_order_coupon_lookup` when coupons themselves are deleted - even if they've been used in orders.

**To Reproduce**

1. Place an order with a coupon (or find one that already has one)
2. View the new analytics report for it ( Analytics > Revenue, select the date )
3. New tab - view the old analytics report for the date
4. Edit the coupon, set to `draft` status
5. Run Historical Data Import ( Analytics > Settings ) so the order is included
6. Refresh both reports
7. Verify the old report coupon value is still correct
8. Verify the new report coupon value is now 0

**Expected behavior**

I expect historical coupon data to be reported on correctly, regardless of the coupons that currently exist. This includes coupons that have been deleted, edited (status or amount/type).

**Screenshots**

Before:

Screen Shot 2020-06-22 at 10 18 37 AM

After:

Screen Shot 2020-06-22 at 10 21 56 AM

Old (same):

Screen Shot 2020-06-22 at 10 18 52 AM

**Additional context**

p1592827581463600-slack-team-isotope

jeffstieler:
Fixes #4657.

This PR seeks to remedy the handling of deleted/un-published coupons when syncing report data.

Previously, our reporting ignored coupons that had been deleted, regardless of whether or not they were used by customers.

**Changes to coupon syncing / reporting:**

* Preserve `wc_order_coupon_lookup` rows when coupons are deleted
* When syncing orders with deleted coupons applied, grab coupon ID from order item meta if available
  * If meta not available, generate a unique **negative** ID for the coupon so it can still be reported on
* Render deleted coupons in the Coupons Report table

**Potential future improvements**

* Add the deleted coupons to the coupons column in the Orders report table

We could potentially add a `coupon_code` column to the `wc_order_coupon_lookup` that would allow for listing the deleted coupons. Alternatively we could use the order item meta to find the coupon code when generating the `extended_info` on the Orders endpoint, but the former would be easier (but potentially needing data backfilled) and likely more performant.

cc: @jessepearson 

### Screenshots

Screen Shot 2020-06-23 at 2 51 26 PM

1. Place an order with a coupon (or find one that already has one)
1. View the new analytics report for it ( Analytics > Revenue, select the date )
1. New tab - view the old analytics report for the date
1. Edit the coupon, set to draft status
1. Run Historical Data Import ( Analytics > Settings ) so the order is included
1. Refresh both reports
1. Verify the old report coupon value is still correct
1. Verify the new report coupon value is still correct
1. Open Analytics > Coupons Report
1. Verify the deleted coupon is reported on

### Changelog Note:

Fix: Reporting of deleted coupons.

### DIFF
--- a/src/API/Reports/Coupons/DataStore.php
+++ b/src/API/Reports/Coupons/DataStore.php
@@ -69,7 +69,6 @@ class DataStore extends ReportsDataStore implements DataStoreInterface {
 	 */
 	public static function init() {
 		add_action( 'woocommerce_analytics_delete_order_stats', array( __CLASS__, 'sync_on_order_delete' ), 5 );
-		add_action( 'delete_post', array( __CLASS__, 'delete_coupon' ) );
 	}
 
 	/**
@@ -442,27 +441,6 @@ class DataStore extends ReportsDataStore implements DataStoreInterface {
 		 * @param int $order_id  Order ID.
 		 */
 		do_action( 'woocommerce_analytics_delete_coupon', 0, $order_id );
-
-		ReportsCache::invalidate();
-	}
-
-	/**
-	 * Deletes the coupon lookup information when a coupon is deleted.
-	 * This keeps data consistent if it gets resynced at any point.
-	 *
-	 * @param int $post_id Post ID.
-	 */
-	public static function delete_coupon( $post_id ) {
-		global $wpdb;
-
-		if ( 'shop_coupon' !== get_post_type( $post_id ) ) {
-			return;
-		}
-
-		$wpdb->delete(
-			self::get_db_table_name(),
-			array( 'coupon_id' => $post_id )
-		);
 
 		ReportsCache::invalidate();
 	}
--- a/src/API/Reports/Coupons/DataStore.php
+++ b/src/API/Reports/Coupons/DataStore.php
@@ -326,9 +326,9 @@ class DataStore extends ReportsDataStore implements DataStoreInterface {
 		$coupon_data = $coupon_item->get_meta( 'coupon_data', true );
 
 		// Normal checkout orders should have this data.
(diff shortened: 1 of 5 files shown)


src/API/Reports/Coupons/DataStore.php
```diff
@@ -347,14 +380,16 @@ public static function sync_order_coupons( $order_id ) {
 		$coupon_items       = $order->get_items( 'coupon' );
 		$coupon_items_count = count( $coupon_items );
```
becdetat, line 381: It doesn't look like this is used - in fact it doesn't look like it was really used at all

becdetat (approved):
This looks good and works well. Rock on 🎸

src/API/Reports/Coupons/DataStore.php
jeffstieler: It's used here: https://github.com/woocommerce/woocommerce-admin/pull/4671/files/e3e0583296e35f6805c92034014be1b8b23018df#diff-a16bed59f4a4360d1c1fa341022c0adcR437

jeffstieler:
So is this good to merge? I always look for the labels to change from "needs review" to "ready to merge".

jeffstieler:
Closing and reopening in attempt to restart Travis CI.

PR #4674, opened 2020-06-23
Fix inbox note deletion test cases.

jeffstieler:
Counts were incorrect - 4 notes get created in the setup: https://github.com/woocommerce/woocommerce-admin/blob/e959ef474ec8c0a9ea6cea11d9ce944796c158a0/tests/framework/helpers/class-wc-helper-admin-notes.php#L29


becdetat (approved):
Looks good and the tests pass. 🚢

PR #4680, opened 2020-06-24
Check that the possibly_add_note function exists before calling it

### LINKED ISSUE #4658: Fatal Error on Latest beta build
I just updated to the latest beta and got a fatal. I had to revert to 4.2. please advice.

Uncaught Error: Call to undefined method Automattic\WooCommerce\Admin\Notes\WC_Admin_Notes_Deactivate_Plugin::possibly_add_note() in /var/docs/wptheme.tst/public/wp-content/plugins/woocommerce/packages/woocommerce-admin/src/Package.php:56 Stack trace: #0 [internal function]: Automattic\WooCommerce\Admin\Composer\Package::init() woocommerce/woocommerce#1 /var/docs/wptheme.tst/public/wp-content/plugins/woocommerce/src/Packages.php(72): call_user_func() woocommerce/woocommerce#2 /var/docs/wptheme.tst/public/wp-content/plugins/woocommerce/src/Packages.php(48): Automattic\WooCommerce\Packages::load_packages() woocommerce/woocommerce#3 /var/docs/wptheme.tst/public/wp-includes/class-wp-hook.php(287): Automattic\WooCommerce\Packages::on_init() woocommerce/woocommerce#4 /var/docs/wptheme.tst/public/wp-includes/class-wp-hook.php(311): WP_Hook->apply_filters() woocommerce/woocommerce#5 /var/docs/wptheme.tst/public/wp-includes/plugin.php(478): WP_Hook->do_action() woocommerce/woocommerce#6 /var/docs/wptheme.tst/public/wp-settings.php(411): do_action() woocommerce/woocommerce#7 /var/docs/wptheme.tst/public/wp-config.php(90): require_once('/var/docs/wpthe...') woocommerce/woocommerce#8 /var/docs/wptheme in /var/docs/wptheme.tst/public/wp-content/plugins/woocommerce/packages/woocommerce-admin/src/Package.php on line 56

becdetat:
Fixes #4658

So if you:

1. spin up a fresh JN site
2. open an SSH session to the site, because you're going to break it
3. install & activate woocommerce-4.3.0-beta.1
4. install & activate WooCommerce Admin from the plugins screen

the site will crash:

image

What's happening is that `Package.php` that is bundled in WC 4.3.0-beta.1 is trying to call the `possibly_add_note` function on a note however it's using the WCA plugin installed in step 4, which is an older version (1.2.4) that doesn't have the `possibly_add_note` function defined (it's in `NoteTraits.php`).

To fix it, in the SSH session:

```
vim /apps/user________/public/wp-content/plugins/woocommerce-4.3.0-beta.1/packages/woocommerce-admin/src/Package.php
```

find line 56 and change the call to `possibly_add_note` to the following:

```
if ( method_exists ( $update_version, 'possibly_add_note' ) ) {
    $update_version::possibly_add_note();
}
```

Save that and the site should magically start working - the note still won't appear unfortunately and I don't know if that's a dealbreaker or not.

### DIFF
--- a/src/Package.php
+++ b/src/Package.php
@@ -53,7 +53,9 @@ class Package {
 			self::$active_version = WC_ADMIN_VERSION_NUMBER;
 			$update_version       = new WC_Admin_Notes_Deactivate_Plugin();
 			if ( version_compare( WC_ADMIN_VERSION_NUMBER, self::VERSION, '<' ) ) {
-				$update_version::possibly_add_note();
+				if ( method_exists( $update_version, 'possibly_add_note' ) ) {
+					$update_version::possibly_add_note();
+				}
 			} else {
 				$update_version::delete_note();
 			}


src/Package.php
```diff
@@ -53,7 +53,9 @@ public static function init() {
 			self::$active_version = WC_ADMIN_VERSION_NUMBER;
 			$update_version       = new WC_Admin_Notes_Deactivate_Plugin();
 			if ( version_compare( WC_ADMIN_VERSION_NUMBER, self::VERSION, '<' ) ) {
-				$update_version::possibly_add_note();
+				if ( method_exists( $update_version, 'possibly_add_note' ) ) {
```
joshuatf, line 56: This is a solid fix and works well to fix the issue.

This is mostly my fault, but one of the main issues here is that we're mixing two different versions of WCA every time we do this.  I think going forward we have 2 approaches to solve this:

1. Pull the note logic into `Package.php`.  This is messy, but guarantees we always have the note.
1. Add an action here (e.g., `do_action( 'woocommerce_admin_package_override', WC_ADMIN_VERSION_NUMBER,  self::VERSION )`.

This way, no shared knowledge of classes exists between the 2 different versions.

joshuatf (commented):
Really nice work diagnosing and fixing this issue!

I think this is a good fix and if we want to merge this in quickly, this solution works.  I do think a longer-term fix would be nice here since this issue has popped up a couple times.  Left some comments, but open to feedback on those ideas.

jeffstieler:

jeffstieler:
This needs to be rebased from `main` to fix the failing test.

src/Package.php
timmyc: Great suggestion Josh. I think given the timeline for this release cycle we should ship this fix as-is, and refactor during the next cycle.

timmyc:
Rebased.

timmyc (approved):
This is testing out well for me - I made a note to create a follow-up issue per Josh's suggestion but am going to ship this in RC2.

timmyc:
cherry picked to `release/1.3.0-rc.1` in 6c4ee46

PR #4681, opened 2020-06-24
Fix spacing between stats overview and store management cards

jameskoster:
Before:

before

After:

after

### DIFF
--- a/client/homescreen/stats-overview/style.scss
+++ b/client/homescreen/stats-overview/style.scss
@@ -4,8 +4,6 @@ $promo-actions-border-top: $core-grey-light-500;
 $promo-actions-border-bottom: $light-gray-tertiary;
 
 .woocommerce-stats-overview {
-	margin-bottom: $gap-large;
-
 	.woocommerce-card__body {
 		padding: 0;
 	}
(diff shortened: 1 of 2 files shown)


client/homescreen/stats-overview/style.scss
```diff
@@ -4,6 +4,8 @@ $promo-actions-border-top: $core-grey-light-500;
 $promo-actions-border-bottom: $light-gray-tertiary;
 
 .woocommerce-stats-overview {
+	margin-bottom: $gap-large;
```
joshuatf, line 7: Should this spacing be used across all cards on the homescreen?  Maybe this margin should be moved to `.woocommerce-homescreen-card`.

Or better yet, we could apply this to any root item added to a homescreen column:

```scss
.woocommerce-homescreen-column > div {
	margin-bottom: $gap-large;
}
```

Then we can remove component specific styles like the same bottom margin added to `.woocommerce-homescreen .woocommerce-task-dashboard__container`.

joshuatf (commented):
Thanks for this fix, Jay!  Added a quick question about what elements this should be applied to and a potential change so these are covered going forward.

client/homescreen/stats-overview/style.scss
jameskoster: yea that would probably be better :)

timmyc:
Going to close/re-open to get CI to do the dance.

timmyc:
and we're back.

timmyc (approved):
Rebased to get CI happy, tested. all good. thanks Jay!

timmyc:
cherry-picked into `release/1.3.0-rc.1` in 34b8bbc3df7fc6b9369251110031edd781e7e8dd

PR #4682, opened 2020-06-24
Update summary number placeholder styles

jameskoster:
After updating the summary number styles I noticed the placeholder didn't line up perfectly. In particular the heights weren't matching, which was causing issues in the logic around when to "stick" the right-hand column on the Home Screen.

Before:

before

After:

after


timmyc:
cherry picked to `release/1.3.0-rc.1` in a61a90f9e8c543af12eb1ff96ff125aa58ea4ce6

PR #4683, opened 2020-06-24
Refactor task tax to use promise chain

### LINKED ISSUE #4631: Task List - Tax task: redirect to settings after enable automated taxes
**Describe the bug**
On the Tax step, I'm being redirected to the settings after enabling automated taxes: https://d.pr/i/b2sGmf

**To Reproduce**

1. Create a new store and install/connect JP and WCS at the end of the store profiler
2. Go to the Tax task in the Home screen
3. See the message about automated taxes and click to enable them
4. See that you're redirected to Settings
5. If you go back to the task and do the same it not redirects to the home screen

**Expected behavior**
I was expecting to be redirected to the home screen

### LINKED ISSUE #4667: Task List: Setup tax step Success screen bad redirect
**Describe the bug**

When clicking "yes please" on a fresh install's tax task success screen, a redirect to `admin.php?page=wc-settings` when it should not. This only happens the first time. If you try the task again, it behaves as it should.

**To Reproduce**

1. Fresh install
2. Go through the OBW using a US based store, install WC Services, and connect Jetpack.
3. On the Setup tax task of the Task List, you should see the success screen:

Screen Shot 2020-06-23 at 12 33 04 PM

4. Click "Yes please"
5. See a flash of the Homescreen and then a redirect to `admin.php?page=wc-settings`.
6. Go back to Homescreen, refresh (for good measure), and repeat step 3 - 5, this time the redirect will not happen.

joshuatf:
Fixes #4631
Fixes #4667

After the wp data transition, we now have access to async updates.  Since the `await` is now working, this caused some odd behavior on component update.

The logic in this component became tricky to follow and similar to other various updates around onboarding, this PR makes use of returned promises from wp data actions and attempts to minimize the state.

@pmcpinto @jameskoster It looks like with the current behavior, plugins only get installed on the tax step if TOS is already accepted, which currently will only occur if they opt in on the benefits screen from the profiler.  This seems redundant- should we remove the plugins step from here or is it possible to allow plugin installation and opt in to TOS from this step?

1. On a fresh install, walk through the profiler, using a TaxJar supported country (e.g., US) and opting in at the benefits screen in the profiler.
1. Navigate to the tax task.
1. Make sure the "success" screen is shown and that click "Yes please" is successful and returns to the task list without redirecting to the tax configuration screen.
1. Deactivate WCS and/or Jetpack.
1. Go to the task tax again.
1. Check that you can manually configure taxes.
1. Go to General WC Settings and remove required pieces of your address.
1. Navigate back to the task tax.
1. Make sure you're able to fill in your address and complete the steps.
1. Also make sure that "skip" actions continue to work as expected.
1. Make sure busy states work as expected.

### DIFF
--- a/client/profile-wizard/steps/business-details.js
+++ b/client/profile-wizard/steps/business-details.js
@@ -696,11 +696,7 @@ BusinessDetails.contextType = CurrencyContext;
 
 export default compose(
 	withSelect( ( select ) => {
-		const {
-			getSettings,
-			getSettingsError,
-			isGetSettingsRequesting,
-		} = select( SETTINGS_STORE_NAME );
+		const { getSettings, getSettingsError } = select( SETTINGS_STORE_NAME );
 		const { getProfileItems, getOnboardingError } = select(
 			ONBOARDING_STORE_NAME
 		);
@@ -716,7 +712,6 @@ export default compose(
 			isError: Boolean( getOnboardingError( 'updateProfileItems' ) ),
 			profileItems: getProfileItems(),
 			isSettingsError: Boolean( getSettingsError( 'general' ) ),
-			isSettingsRequesting: isGetSettingsRequesting( 'general' ),
 			settings,
 			isInstallingActivating:
 				isPluginsRequesting( 'installPlugins' ) ||
--- a/client/profile-wizard/steps/store-details.js
+++ b/client/profile-wizard/steps/store-details.js
@@ -237,11 +237,7 @@ StoreDetails.contextType = CurrencyContext;
 
 export default compose(
 	withSelect( ( select ) => {
-		const {
-			getSettings,
-			getSettingsError,
-			isGetSettingsRequesting,
-		} = select( SETTINGS_STORE_NAME );
+		const { getSettings, getSettingsError } = select( SETTINGS_STORE_NAME );
 		const { getOnboardingError, getProfileItems } = select(
 			ONBOARDING_STORE_NAME
 		);
@@ -253,12 +249,10 @@ export default compose(
 
 		const { general: settings = {} } = getSettings( 'general' );
 		const isSettingsError = Boolean( getSettingsError( 'general' ) );
-		const isSettingsRequesting = isGetSettingsRequesting( 'general' );
 
 		return {
 			isProfileItemsError,
 			isSettingsError,
-			isSettingsRequesting,
 			profileItems,
 			settings,
 		};
--- a/client/task-list/style.scss
+++ b/client/task-list/style.scss
@@ -584,3 +584,8 @@
 		border-top-right-radius: 2px;
 	}
 }
+
+.woocommerce-task__caption {
+	color: $dark-gray-300;
+	margin-top: $gap;
+}
--- a/client/task-list/tasks/shipping/index.js
+++ b/client/task-list/tasks/shipping/index.js
@@ -302,13 +302,13 @@ class Shipping extends Component {
 
 	render() {
 		const { isPending, step } = this.state;
-		const { isSettingsRequesting } = this.props;
+		const { isUpdateSettingsRequesting } = this.props;
 
 		return (
 			<div className="woocommerce-task-shipping">
 				<Card className="is-narrow">
 					<Stepper
-						isPending={ isPending || isSettingsRequesting }
+						isPending={ isPending || isUpdateSettingsRequesting }
 						isVertical
 						currentStep={ step }
 						steps={ this.getSteps() }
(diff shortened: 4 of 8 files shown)


jameskoster:
If we can simply add a "by installing Jetpack and WooCommerce Services you agree to the TOS" line somewhere, and preserve plugin installation in this step, that would be worth it imo.

jeffstieler:

jeffstieler:
This needs to be rebased from `main` to fix the failing test.

psealock (commented):
This is awesome @joshuatf. What was happening to cause the redirect?

I came across an error.

Screen Shot 2020-06-25 at 4 25 05 PM\

Here is the error in minified code:

Screen Shot 2020-06-25 at 4 26 34 PM

Looks like its coming from 
https://github.com/woocommerce/woocommerce-admin/blob/0b68c1d7c0c5025789071022efa129126da1ded5/client/dashboard/utils.js#L19-L25

To reproduce:

1. Remove Address field from WC settings
2. Tax Task
3. Enter address and continue
4. See the error. If you refresh and try again, it succeeds.

Otherwise the original bug is no longer there.

joshuatf:
Thanks for the review, @psealock!  I wasn't able to reproduce the address bug.  Any other tricks that might help?  What country did you have selected?

>What was happening to cause the redirect?

Your guess (discussed in Slack) as to why this was happening was correct.  The previous method of checking was to monitor changes to options and redirect when a change occurred.

```js
componentDidUpdate() {
	...

	const {	
		woocommerce_calc_taxes: prevCalcTaxes,	
	} = prevProps.generalSettings;	
	if ( prevCalcTaxes === 'no' && calcTaxes === 'yes' ) {	
		window.location = getAdminLink(	
			'admin.php?page=wc-settings&tab=tax&section=standard'	
		);	
	}
}
```

Now that we have wp data actions returning promises, the `await` began to work when updating options and the component would update tax settings and redirect to the core tax screen before unmounting.

The logic in this task was generally a mess from previous development without promises, so I took this as an opportunity to refactor the whole thing.

We have quite a few other places in onboarding that are still in need of this, which I'll make note of in https://github.com/woocommerce/woocommerce-admin/issues/2959.  Besides async patterns that could use some patching from the days of wc-api, we also have some leftover error handling that won't work as expected.

joshuatf:
Also noting I added the change that @jameskoster requested in https://github.com/woocommerce/woocommerce-admin/pull/4683/commits/024cf7671067eb917504a8ef58dd78b59648444f.  This shows the plugin step if plugins are needed or if opt-in to TOS is needed to enable automatic taxes.

Testing steps should be mostly identical, but disable Jetpack or WCS to also note that the plugins step will appear.

Screen Shot 2020-06-25 at 3 29 58 PM

psealock:
>I wasn't able to reproduce the address bug.

I tried again and wasn't able to do so consistently. I'm not sure what changed, but I tried it maybe 10 times and reproduced once. I'm ok with shipping this and seeing if we get reports of it happening in the wild.

>the component would update tax settings and redirect to the core tax screen before unmounting.

Thats what I figured, but was thrown off by `&section=standard` because I didn't see that query param in my testing when redirected. 🤷  Thanks for the explanation and fix.

> disable Jetpack or WCS to also note that the plugins step will appear.

This part is working well for me. I did notice one thing. When click "Install and enable" or "Continue" when having to enter my address, the Success screen showed and automatically moved on. Maybe thats because I already chose that option?

tax

joshuatf:
>This part is working well for me. I did notice one thing. When click "Install and enable" or "Continue" when having to enter my address, the Success screen showed and automatically moved on. Maybe thats because I already chose that option?

Great catch!  I think this is happening because we call `completeStep()` and there's a generic catch to send us back to the task list if there are no more steps.  Since the success screen isn't really part of the steps, it assumes the store address is the last step and redirects back to the task list.

I removed that snippet in https://github.com/woocommerce/woocommerce-admin/pull/4683/commits/ab1f19a8b574283157cef536404f957430598c50 as every step has a way of manually exiting this flow.  Testing this seems to work well, but could use your 👀  to double check.

jameskoster:
img

Nice! Would it be possible to use the `caption` Text style for this? And set the color to `$dark-gray-300`?

joshuatf:
Thanks for taking the time to look at this @jameskoster!  Updated styling for captions in https://github.com/woocommerce/woocommerce-admin/pull/4683/commits/3567fbf2e8a82061c275a48b16f22048397b4df8

psealock (approved):
Changes look good, thanks for addressing :shipit:

PR #4686, opened 2020-06-24
Added new note type

### LINKED ISSUE #4656: Modify inbox note name and behavior
This is a follow up to the [PR 4621](https://github.com/woocommerce/woocommerce-admin/pull/4621). We need to modify the feedback note "Insight". 

- The note should disappear after the action.
- The selection should be made through a confirmation using a Snackbar.
- Change note title from "Insight" to "Did you know?"

octaedro:
Fixes #4656

_This is a follow up to the [PR 4621](https://github.com/woocommerce/woocommerce-admin/pull/4621)._

This PR adds a new note type `survey`. The survey note is deleted after calling its CTA.
Also, the note `WC_Admin_Notes_Insight_First_Sale.php` was modified since it belongs to this type of note. 

### Accessibility

### Screenshots
Screen Capture on 2020-06-24 at 10-29-09

- Delete the notification `wc-admin-insight-first-sale`.
You can use a SQL sentence like:
````
DELETE FROM `wp_wc_admin_notes` WHERE `wp_wc_admin_notes`.`name` = 'wc-admin-insight-first-sale';
````
- Trigger the cron to add the note again. To do it you can use the plugin [WP Crontrol](https://wordpress.org/plugins/wp-crontrol/)
- Verify that the note's title is: `Did you know?`.
- Select one of the note's options.
- The message `Thanks for your feedback` should be visible.
- The note should disappear.

### Changelog Note:
Dev: added new note type and modified the note `wc-admin-insight-first-sale` title.

// cc: @pmcpinto

### DIFF
--- a/client/wc-api/constants.js
+++ b/client/wc-api/constants.js
@@ -22,5 +22,5 @@ export const QUERY_DEFAULTS = {
 	pageSize: 25,
 	period: 'month',
 	compare: 'previous_year',
-	noteTypes: 'info,warning,marketing',
+	noteTypes: 'info,warning,marketing,survey',
 };
--- a/src/API/NoteActions.php
+++ b/src/API/NoteActions.php
@@ -108,11 +108,6 @@ class NoteActions extends Notes {
 			$note->set_status( $triggered_action->status );
 		}
 
-		// Update note state depending on its type.
-		if ( WC_Admin_Note::E_WC_ADMIN_NOTE_SURVEY === $note->get_type() ) {
-			$note->set_is_deleted( 1 );
-		}
-
 		$note->save();
 
 		if ( in_array( $note->get_type(), array( 'error', 'update' ) ) ) {
--- a/src/API/NoteActions.php
+++ b/src/API/NoteActions.php
@@ -11,6 +11,7 @@ namespace Automattic\WooCommerce\Admin\API;
 
 defined( 'ABSPATH' ) || exit;
 
+use Automattic\WooCommerce\Admin\Notes\WC_Admin_Note;
 use \Automattic\WooCommerce\Admin\Notes\WC_Admin_Notes;
 
 /**
@@ -108,7 +109,7 @@ class NoteActions extends Notes {
 		}
 
 		// Update note state depending on its type.
-		if ( 'survey' === $note->get_type() ) {
+		if ( WC_Admin_Note::E_WC_ADMIN_NOTE_SURVEY === $note->get_type() ) {
 			$note->set_is_deleted( 1 );
 		}
 
--- a/src/API/NoteActions.php
+++ b/src/API/NoteActions.php
@@ -107,6 +107,11 @@ class NoteActions extends Notes {
 			$note->set_status( $triggered_action->status );
 		}
 
+		// Update note state depending on its type.
+		if ( 'survey' === $note->get_type() ) {
+			$note->set_is_deleted( 1 );
+		}
+
 		$note->save();
 
 		if ( in_array( $note->get_type(), array( 'error', 'update' ) ) ) {
--- a/src/Notes/WC_Admin_Notes.php
+++ b/src/Notes/WC_Admin_Notes.php
@@ -164,7 +164,7 @@ class WC_Admin_Notes {
 				'orderby'    => 'date_created',
 				'per_page'   => 25,
 				'page'       => 1,
-				'type'       => array( 'info', 'marketing', 'warning' ),
+				'type'       => array( 'info', 'marketing', 'warning', 'survey' ),
 				'is_deleted' => 0,
 			)
 		);
--- a/src/Notes/WC_Admin_Notes.php
+++ b/src/Notes/WC_Admin_Notes.php
@@ -24,6 +24,7 @@ class WC_Admin_Notes {
 	 */
 	public static function init() {
 		add_action( 'admin_init', array( __CLASS__, 'schedule_unsnooze_notes' ) );
+		add_action( 'admin_init', array( __CLASS__, 'possibly_delete_survey_notes' ) );
 		add_action( 'update_option_woocommerce_show_marketplace_suggestions', array( __CLASS__, 'possibly_delete_marketing_notes' ), 10, 2 );
 	}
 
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

**Comment by mattsherman** [comment]
I'll review this

File: client/wc-api/constants.js [hunk 1]
```diff
@@ -22,5 +22,5 @@ export const QUERY_DEFAULTS = {
 	pageSize: 25,
 	period: 'month',
 	compare: 'previous_year',
-	noteTypes: 'info,warning,marketing',
+	noteTypes: 'info,warning,marketing,survey',
```
mattsherman, line 25: Could we use the consts from WC_Admin_Note?

src/API/NoteActions.php
```diff
@@ -107,6 +107,11 @@ public function trigger_note_action( $request ) {
 			$note->set_status( $triggered_action->status );
 		}
 
+		// Update note state depending on its type.
+		if ( 'survey' === $note->get_type() ) {
```
mattsherman, line 111: Could we use the const from WC_Admin_Note?

src/Notes/WC_Admin_Notes.php
```diff
@@ -164,7 +164,7 @@ public static function delete_all_notes() {
 				'orderby'    => 'date_created',
 				'per_page'   => 25,
 				'page'       => 1,
-				'type'       => array( 'info', 'marketing', 'warning' ),
+				'type'       => array( 'info', 'marketing', 'warning', 'survey' ),
```
mattsherman, line 167: Could we use the consts from WC_Admin_Note?

src/API/NoteActions.php
mattsherman, line 111: This checking of type and performing some type-specific logic here is not ideal. Maybe you could hook into this using the `woocommerce_note_action` action instead, so that you could keep this deleting logic in `WC_Admin_Notes_Insight_First_Sale.php`.

mattsherman (changes requested):
Tested. Works as expected, except that the entire page reloads when the note is actioned with 'Yes' or 'No' button. I would have expected the note to just hide as other notes do when you dismiss them.

Left some specific comments in-line.

src/API/NoteActions.php
octaedro: Good idea! It was done in the commit 92839f6

src/Notes/WC_Admin_Notes.php
octaedro: Sounds good to me! I changed it in the commit 3cf623e

client/wc-api/constants.js
octaedro: Sorry, I couldn't follow this suggestion, do you think we should load this client constant from the API?

src/API/NoteActions.php
octaedro: I see, it's a nice catch @mattsherman! This gives me an idea😀
What do you think about creating a hook and a deletion method [in WC_Admin_Notes.php](https://github.com/woocommerce/woocommerce-admin/blob/main/src/Notes/WC_Admin_Notes.php#L25) for this type of note?
Doing that I could follow the same pattern we use to [delete the marketing notes](https://github.com/woocommerce/woocommerce-admin/blob/main/src/Notes/WC_Admin_Notes.php#L227) and we don't have to hook the deletion method in every note of this type.

octaedro:
> Tested. Works as expected, except that the entire page reloads when the note is actioned with 'Yes' or 'No' button. I would have expected the note to just hide as other notes do when you dismiss them.

Hi @mattsherman, thank you for the review 🙂
The change you mentioned makes sense. As it's something that comes from an existing notification, is it ok if we handle that change in a follow-up PR?

client/wc-api/constants.js
mattsherman: Ah, yeah... oops, sorry for making this suggestion. No need to load the constant from the API! Missed that this was client-side JS vs the server-side WC_Admin_Note PHP! 😊

src/API/NoteActions.php
mattsherman: @octaedro I think I follow your idea. I think that would be good.

mattsherman:
> > Tested. Works as expected, except that the entire page reloads when the note is actioned with 'Yes' or 'No' button. I would have expected the note to just hide as other notes do when you dismiss them.
> 
> Hi @mattsherman, thank you for the review 🙂
> The change you mentioned makes sense. As it's something that comes from an existing notification, is it ok if we handle that change in a follow-up PR?

Sure, I think that is reasonable.

mattsherman (approved):
I reviewed the latest commits to use the constants. I didn't re-test the changes, but feel confident approving these changes since they were minor and clear.

Thanks for making those changes. Nice work!

octaedro:
Hi @mattsherman,
I added the changes we mentioned in [this comment](https://github.com/woocommerce/woocommerce-admin/pull/4686#discussion_r445813423) in the commit 7caaec3.
Also, I created [another PR](https://github.com/woocommerce/woocommerce-admin/pull/4711) to handle the webpage refreshing after answering a survey.
If you have time, could you give this PR another look?

src/Notes/WC_Admin_Notes.php
```diff
@@ -24,6 +24,7 @@ class WC_Admin_Notes {
 	 */
 	public static function init() {
 		add_action( 'admin_init', array( __CLASS__, 'schedule_unsnooze_notes' ) );
+		add_action( 'admin_init', array( __CLASS__, 'possibly_delete_survey_notes' ) );
```
mattsherman, line 27: Hooked into this action, the survey note will not get removed from the inbox until a full page reload. Is this what was intended?

If the note should be removed from UI after actioned, perhaps there should be a new option added when actions are registered to specify that a note should be deleted after an action is taken?

mattsherman (changes requested):

src/Notes/WC_Admin_Notes.php
octaedro: Hi @mattsherman, thank you for your review.

>Hooked into this action, the survey note will not get removed from the inbox until a full page reload. Is this what was intended?

Yes, I tried to keep the survey note, after the survey has been answered, visible with the message "Thank you for your feedback" until the next full page reload.
Maybe (just to confirm) we can ask @pmcpinto if the survey note should disappear immediately after it was activated or should stay with the message "Thank you for your feedback" until the page reload (like now).

pmcpinto:
> Maybe (just to confirm) we can ask @pmcpinto if the survey note should disappear immediately after it was activated or should stay with the message "Thank you for your feedback" until the page reload (like now).

It's ok to leave it until the page reload. Thanks

octaedro:
> It's ok to leave it until the page reload. Thanks

Thank you @pmcpinto!

@mattsherman I will add the `Needs Review` tag to this PR. Could you give it another 👀 ?

mattsherman (approved):
Thanks for the updates @octaedro -- nice work on this one!

I re-tested and verified that it works as expected.

PR #4694, opened 2020-06-25
Add education and travel to the onboarding industry types

### LINKED ISSUE #4476: Store Profiler - Industry step: add two additional industries as options
Since we implemented an input field when the user selects Other, there are 2 areas that stand out: 
- Education & learning
- Travel & tourism

So in order to have a clear picture of how many users are really operating in those industries, we should display them as options. They can be placed before the `Other` option.

becdetat:
Fixes #4476

This adds:

- Education and learning
- Travel and tourism

industry options to the onboarding wizard.

Note that the new industries also need to be added to WooCommerce.com (pr TBA).

### Screenshots

image

- Enable tracks debugging in the console: `localStorage.setItem( 'debug', 'wc-admin:tracks' )`
- Open the OBW and ensure that tracking is enabled
- On the industries step, make sure the new industries are displayed
- select the new industries and navigate next
- make sure the new industries are logged in the console:
image
- finish the OBW then start it again
- when you get back to the industries step, the new industries should be selected

### DIFF
--- a/src/Features/Onboarding.php
+++ b/src/Features/Onboarding.php
@@ -310,6 +310,16 @@ class Onboarding {
 					'use_description'   => false,
 					'description_label' => '',
 				),
+				'education-and-learning'          => array(
+					'label'             => __( 'Education and learning', 'woocommerce-admin' ),
+					'use_description'   => false,
+					'description_label' => '',
+				),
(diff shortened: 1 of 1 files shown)


psealock (approved):
Works well, testing as expected.

PR #4695, opened 2020-06-25
Add bundle install UI to Business Details step

joshuatf:
Fixes (partially) #4579 

* Adds the alternative for install UI.
* Installs all business extension plugins (including Jetpack and WCS) on opt in.
* Saves newly added data to tracks and onboarding profile.

Note that this new UI is behind a flag until we have the desired segment determined and a follow-up to address benefits and Jetpack connection flow is in place.

@pmcpinto I just noticed your updates in #4579 around events.  I think this PR covers most of those, but instead of splitting into multiple events, the same `storeprofiler_store_business_details_continue` now includes all plugins as well as a `bundle_install` property to determine if they used the new bundle UI and `install_extensions` to determine if they opted in or out.  It does not yet cover:

* Changes to `wcadmin_storeprofiler_step_view` to add properties around this view.  We'll have to give more thought into how we can hook into that event from this child component.
* Clicking the popover.  I noticed some discussion of events firing when the popover is clicked, but also noticed that it should only show on hovering the icon/button (as it does in this PR).  Could you clarify which events are supposed to happen.

Aside from those two, if you feel any of the other events aren't enough, let me know and we can adjust accordingly.

### Screenshots
Screen Shot 2020-06-25 at 2 14 12 PM

1. Go to the Business Details step of the profiler and make sure the old flow works as expected.
1. In `client/profile-wizard/steps/business-details.js` update line 83 to `this.bundleInstall = true;`.
1. Try toggling the install checkbox on/off.
1. Make sure all 5 plugins are installed and activated when toggled on.
1. Make sure recorded events match expectations.
1. Make sure profile data is updated with all installed extensions.

### DIFF
--- a/client/profile-wizard/steps/business-details.js
+++ b/client/profile-wizard/steps/business-details.js
@@ -233,7 +233,14 @@ class BusinessDetails extends Component {
 
 	getBusinessExtensions( values ) {
 		if ( this.simpleInstall ) {
-			return values.install_extensions ? this.extensions : [];
+			return values.install_extensions
+				? [
+						'jetpack',
+						'woocommerce-services',
+						'woocommerce-payments',
+						...this.extensions,
+				  ]
+				: [];
 		}
 
 		return keys( pickBy( values ) ).filter( ( name ) =>
@@ -422,14 +429,18 @@ class BusinessDetails extends Component {
 						id="woocommerce-business-extensions__checkbox"
 						{ ...getInputProps( 'install_extensions' ) }
 					/>
-					{ __(
-						'Install recommended free business features',
-						'woocommerce-admin'
-					) }
-					<span className="woocommerce-profile-wizard__help-text">
-						{ __( 'Requires an account', 'woocommerce-admin' ) }
+					<span className="woocommerce-business-extensions__label-text">
+						{ __(
+							'Install recommended free business features',
+							'woocommerce-admin'
+						) }
+						<span className="woocommerce-business-extensions__help-text">
+							{ __( 'Requires an account', 'woocommerce-admin' ) }
+						</span>
 					</span>
+				</label>
 
+				<div className="woocommerce-business-extensions__popover-wrapper">
 					<Button
 						isTertiary
 						onMouseEnter={ () =>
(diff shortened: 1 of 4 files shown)


jeffstieler:
There's been an update to where we are disclosing the installation:

FXYntXBRvk-3000x3000

pbIJXs-bo-p2#comment-477

joshuatf:
Thanks for the update, @jeffstieler.  Updated this in https://github.com/woocommerce/woocommerce-admin/pull/4695/commits/68b5faa8acac4ec5725bbf19fcb4ddf2ecf563f1 and ready for review.

Screen Shot 2020-06-25 at 6 59 51 PM

pmcpinto:
> @pmcpinto I just noticed your updates in #4579 around events. I think this PR covers most of those, but instead of splitting into multiple events, the same storeprofiler_store_business_details_continue now includes all plugins as well as a bundle_install property to determine if they used the new bundle UI and install_extensions to determine if they opted in or out. It does not yet cover:

Ok, so the `bundle_install` property can be used to identify if they're in this version of the step or it would be better to have a new prop to identify the version. Eg: version: a|b

> Clicking the popover. I noticed some discussion of events firing when the popover is clicked, but also noticed that it should only show on hovering the icon/button (as it does in this PR). Could you clarify which events are supposed to happen.

I'm good at tracking any of the options. I think on hover is the most comment practice. @jameskoster what do you think?

> Note that this new UI is behind a flag until we have the desired segment determined

Just a heads up that the segment was already added to the description: Final segmentation: Stores that selected the US as country and Fashion, Apparel, Accessories or Health & Beauty as Industry. It gives ~2.5% of the total number of new users and ~30% of the total number of new users from US

octaedro (commented):
Hi @joshuatf, you did a great job here!
The code looks really good but I experienced a few troubles in my testing.

If I don't set `this.bundleInstall` as `true`, the `Continue` button doesn't work in the `Business Details` step.

After setting `this.bundleInstall` as `true`:
Is it ok that it's possible to continue without answering `How many products do you plan to display?` or `Currently selling elsewhere?` in the `Business Details` step? Also, I think that it's not saving my selection in these dropdowns.

It seems like the button `Continue with my active theme`, in the `Theme` step, is not working well. For instance, if I don't select the checkbox `Install recommended free business features`, the button doesn't work.

Screen Capture on 2020-06-25 at 14-25-39

client/profile-wizard/steps/business-details.js
```diff
@@ -392,6 +444,132 @@ class BusinessDetails extends Component {
...
+		return (
+			<div className="woocommerce-business-extensions">
+				<label htmlFor="woocommerce-business-extensions__checkbox">
+					<CheckboxControl
+						id="woocommerce-business-extensions__checkbox"
+						{ ...getInputProps( 'install_extensions' ) }
+					/>
+					<span className="woocommerce-business-extensions__label-text">
+						{ __(
+							'Install recommended free business features',
```
octaedro, line 459: [In the design](https://user-images.githubusercontent.com/63922/85745425-b046b500-b6d3-11ea-8c34-a5a419a735eb.png), it looks like the word `free` is in bold

joshuatf: Good eyes!  Updated in https://github.com/woocommerce/woocommerce-admin/pull/4695/commits/10057953372d071373b129050c6d44a50fb0df1a

joshuatf:
Thank you for reviewing @octaedro!

>If I don't set this.bundleInstall as true, the Continue button doesn't work in the Business Details step.

>Is it ok that it's possible to continue without answering How many products do you plan to display? or Currently selling elsewhere? in the Business Details step? Also, I think that it's not saving my selection in these dropdowns.

Validation was a bit off for both of these.  Fixed this up in https://github.com/woocommerce/woocommerce-admin/pull/4695/commits/177c89aa02a5932c012ec228ae1ef92e039935d8.  Were your selections not being saved because of this?  Could you let me know if they're working okay on your end now?

>It seems like the button Continue with my active theme, in the Theme step, is not working well. For instance, if I don't select the checkbox Install recommended free business features, the button doesn't work.

I did not encounter this issue, but the theme step is definitely up for a refactor in the near future.  I'd say let's tackle this in a follow-up unless it's directly related to this PR.

joshuatf:
>Ok, so the bundle_install property can be used to identify if they're in this version of the step or it would be better to have a new prop to identify the version. Eg: version: a|b

The current `bundle_install` property can be used to identify them.  I'm not opposed to using `a|b` either, especially if we plan on adding more splits or if you see another advantage, but I think the single boolean and naming will make this easy for analyzing data later if we only have these 2 options.

>I'm good at tracking any of the options. I think on hover is the most comment practice. @jameskoster what do you think?

Hover seems to make sense.  The only downside I can see is this event being fired many times, but I suppose that's better than none at all.  We may be able to debounce this a bit as well to keep it from firing constantly.

@jameskoster Do you want to weigh in before we add this event?

>Just a heads up that the segment was already added to the description

Thanks for the heads up @pmcpinto!  Will add this in a follow-up to keep this PR behind the flag since the main issue is still a WIP.

pmcpinto:
> Hover seems to make sense. The only downside I can see is this event being fired many times, but I suppose that's better than none at all. 

Yeah good point. On click has the advantage of making it easier for a mobile user to check the popover

mattsherman:
> > Hover seems to make sense. The only downside I can see is this event being fired many times, but I suppose that's better than none at all.
> 
> Yeah good point. On click has the advantage of making it easier for a mobile user to check the popover

It would seem that we need to handle both hover (for pointer-based interaction) and click (for touch/keyboard-based interaction).

I would only fire the event the first time hover was triggered.

(Side note: I personally would make the info button always require a click to show the popover, as large popovers triggered by hover can be unwieldy. But, that’s @jameskoster’s call!)

jameskoster:
I had a combo of something like the [Button](https://wordpress.github.io/gutenberg/?path=/story/components-button--icon) (icon) and [Popover](https://wordpress.github.io/gutenberg/?path=/story/components-popover--default) components in mind here. I believe that's what core is using for the "[content structure](https://cloudup.com/c2IyP4iHpHk)" feature. So yeah, I think on click is sensible.

joshuatf:
Thanks for the feedback here, everyone!

Updated in https://github.com/woocommerce/woocommerce-admin/pull/4695/commits/83da44ad6a3fa83fce850f65188647def58c54f2 to only show the popover when the button is clicked and also records the event when this occurs.

octaedro (approved):
You did an awesome job here @joshuatf, code looks good and it's testing well here :shipit:

client/profile-wizard/steps/business-details.js
```diff
@@ -100,9 +113,21 @@ class BusinessDetails extends Component {
 			revenue,
 			used_platform: otherPlatform,
 			used_platform_name: otherPlatformName,
-			install_facebook: values[ 'facebook-for-woocommerce' ],
-			install_mailchimp: values[ 'mailchimp-for-woocommerce' ],
-			install_google_ads: values[ 'kliken-marketing-for-google' ],
+			install_woocommerce_services: businessExtensions.includes(
+				'facebook-for-woocommerce'
```
dechov, line 117: Should be `'woocommerce-services'`.

jeffstieler: Thanks for pointing this out! https://github.com/woocommerce/woocommerce-admin/pull/5614

Issue #4147, opened 2020-04-17
Filters to alter the columns used by a DataStore class


daigo75:
Description of the issue
--
I need to be able to extend the reports generated by the WC Admin so that it takes different values. This is needed to be able to prepare correct summary reports in a multi-currency scenario.

Current behaviour
DataStore uses columns like the following (only relevant columns included):
~~~
$this->report_columns = array(
	'gross_sales'     => $gross_sales,
	'total_sales'     => "SUM({$table_name}.total_sales) AS total_sales",
	'coupons'         => 'COALESCE( SUM(discount_amount), 0 ) AS coupons', // SUM() all nulls gives null.
	'refunds'         => "{$refunds} AS refunds",
	'taxes'            => "SUM({$table_name}.tax_total) AS taxes",
	'shipping'        => "SUM({$table_name}.shipping_total) AS shipping",
	'net_revenue'     => "SUM({$table_name}.net_total) AS net_revenue",
	'avg_order_value' => "SUM( {$table_name}.net_total ) / SUM( CASE WHEN {$table_name}.parent_id = 0 THEN 1 ELSE 0 END ) AS avg_order_value",
);
~~~
This produces incorrect results when orders are in different currencies. For example, in a store with EUR as the base currency, running `SUM({$table_name}.total_sales) AS total_sales` against orders for 100 EUR, 100 USD and 100 GBP, would return a total 300 EUR. That's obviously incorrect.

A relatively easy way to fix this issue would be to alter the columns returned by the SELECT, as follows:
~~~
'total_sales' => "SUM({$table_name}.total_sales* FX_RATE_META.meta_value) AS total_sales",
'coupons' => 'COALESCE( SUM(discount_amount * FX_RATE_META.meta_value), 0 ) AS coupons', // SUM() all nulls gives null.
'refunds' => "{$refunds} AS refunds",
'taxes' => "SUM({$table_name}.tax_tota * FX_RATE_META.meta_value) AS taxes",
'shipping' => "SUM({$table_name}.shipping_total * FX_RATE_META.meta_value) AS shipping",
'net_revenue' => "SUM({$table_name}.net_total * FX_RATE_META.meta_value) AS net_revenue",
'avg_order_value' => "SUM( {$table_name}.net_total * FX_RATE_META.meta_value ) / SUM( CASE WHEN {$table_name}.parent_id = 0 THEN 1 ELSE 0 END ) AS avg_order_value",
~~~
[...]

PR #4700, opened 2020-06-25
Added filters for `DataStore` columns

daigo75:
Fixes
* Implements the changes suggested at https://github.com/woocommerce/woocommerce-admin/issues/4147

### Changelog Note:

Tweak - Added filter `wc_analytics_<CONTEXT>_report_columns`, to allow 3rd parties to alter the columns of a report.

### DIFF
--- a/src/API/Reports/Categories/DataStore.php
+++ b/src/API/Reports/Categories/DataStore.php
@@ -72,12 +72,12 @@ class DataStore extends ReportsDataStore implements DataStoreInterface {
 	 */
 	protected function assign_report_columns() {
 		$table_name           = self::get_db_table_name();
-		$this->report_columns = array(
+		$this->report_columns =  apply_filters('wc_analytics_' . $this->context . '_report_columns', array(
 			'items_sold'     => 'SUM(product_qty) as items_sold',
 			'net_revenue'    => 'SUM(product_net_revenue) AS net_revenue',
 			'orders_count'   => "COUNT(DISTINCT {$table_name}.order_id) as orders_count",
 			'products_count' => "COUNT(DISTINCT {$table_name}.product_id) as products_count",
-		);
+		), $table_name, $this);
 	}
 
 	/**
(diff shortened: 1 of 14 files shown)


src/API/Reports/Categories/DataStore.php
```diff
@@ -72,12 +72,12 @@ class DataStore extends ReportsDataStore implements DataStoreInterface {
 	 */
 	protected function assign_report_columns() {
 		$table_name           = self::get_db_table_name();
-		$this->report_columns = array(
+		$this->report_columns =  apply_filters('wc_analytics_' . $this->context . '_report_columns', array(
```
joshuatf, line 75: Just a quick heads up that this filter should be prefixed with `woocommerce_analytics` instead of `wc_analytics`.

src/API/Reports/Coupons/DataStore.php
```diff
@@ -57,11 +57,11 @@ class DataStore extends ReportsDataStore implements DataStoreInterface {
 	 */
 	protected function assign_report_columns() {
 		$table_name           = self::get_db_table_name();
-		$this->report_columns = array(
+		$this->report_columns =  apply_filters('wc_analytics_' . $this->context . '_report_columns', array(
```
joshuatf, line 60: Instead of applying this filter manually in each report, I would much rather see a filter that catches assigning of report columns.  I.e., in https://github.com/woocommerce/woocommerce-admin/blob/cf23784df76838738a60b6bdbda9bc3c63e5ed6c/src/API/Reports/DataStore.php#L101

joshuatf (commented):
Thanks for the PR, @daigo75!

I left a few comments in the code, but overall I'm not opposed to extending the columns here and agree that if external plugins want to break something here, they can do so even without this filter.

I do think that if there is a fundamental issue with how calculations are made, this is something we may want to directly fix in WCA, but this filter could be a short-term fix. /cc @rrennick for your thoughts since you had some discussion in the original issue.

daigo75:
@joshuatf Thanks for your feedback. The changes you propose make sense. My PR was mainly a proof of concept, to illustrate my proposed solution. I also agree that adding the filter in a dozen places is not ideal, but I didn't have sufficient time availability to prepare and test a "cleaner" implementation, which could allow the filters to be called in one place. That's certainly desirable, to simplify implementation and maintenence. 👍

rrennick:
> if external plugins want to break something here, they can do so even without this filter.

Agreed.

becdetat:
> My PR was mainly a proof of concept, to illustrate my proposed solution.

Closing this

Issue #4577, opened 2020-06-15
Store profiler - product types step change


pmcpinto:
**Description:** we aim to iterate on the product types step, in order to increase the number of users that select useful extensions

![Product types.png](https://images.zenhubusercontent.com/5d84fe8e1201de0001b6ab61/675ef778-b51e-44dd-b050-91a78c366148)

**Acceptance Criteria:**
- [ ] Toggle the “Physical products” checkbox on by default
- [ ] Remove the Composite products option
- [ ] Include [Product Bundles](https://woocommerce.com/products/product-bundles/) and [Product Add-ons](https://woocommerce.com/products/product-add-ons/) as options
- [ ] Display monthly pricing instead of annual (see mockup)
- [ ] Add tooltip with more info about the purchase (see mockup)

**Event tracking:**
Probably it won't be necessary additional tracking

**Additional links:**
[Design mockup](https://www.figma.com/file/JH9XMFUCOjfXdr3N09AHRD/On-boarding-iterations-June-'20?node-id=260%3A1491)

becdetat:
Estimate: 2

joshuatf:
Estimate: 2-3

My only concern is changing and formatting those prices from the wccom as monthly prices instead of annual.

jeffstieler:
Agree on @becdetat 's estimate of 2 (assuming we can use our [existing `<Count />` component](https://woocommerce.github.io/woocommerce-admin/#/components/packages/count/README)), and share @joshuatf 's concerns about the potential need to change the WCCOM API. (or can we just divide by 12?)

Question for @jameskoster on the pill/count component color - we just removed a lot of purple from the UI.. are we using purple for this?

joshuatf:
>(or can we just divide by 12?)

After dividing and parsing, I think we can.  It includes an encoded HTML entity for currency that needs to be stripped first and added back on.  Nothing too crazy, just not exactly what you'd expect from the API response:

https://woocommerce.com/wp-json/wccom-extensions/1.0/search?category=product-type

jameskoster:
> Question for @jameskoster on the pill/count component color - we just removed a lot of purple from the UI.. are we using purple for this?

Most likely yes :) The thinking is to be way more intentional when we apply any in-product branding (using purple). The latest direction is essentially to use it around any features that require some interaction or connection with woocommerce.com or wordpress.com. IE purchasing / installing extensions, setting up WCPay, or WCServices.

mattsherman:
[...]

PR #4702, opened 2020-06-26
Toggle the “Physical products” checkbox on by default

becdetat:
Partially addresses #4577

This toggles the "Physical products" product type to be selected by default in the OBW.

1. Reset the OBW by deleting this option: `woocommerce_onboarding_profile`
2. Start the OBW
3. Carry through to the product selection step
4. Note that the "Physical products" product type should be preselected
5. Select some different product types and carry through the rest of the OBW
6. Restart the OBW
7. Carry through to the product selection step
8. Note that the product types you selected in step 5 should be selected

### DIFF
--- a/client/profile-wizard/steps/product-types.js
+++ b/client/profile-wizard/steps/product-types.js
@@ -26,9 +26,13 @@ class ProductTypes extends Component {
 		super();
 		const profileItems = get( props, 'profileItems', {} );
 
+		const { productTypes = {} } = getSetting( 'onboarding', {} );
+		const preselectedProductTypes = Object.keys( productTypes )
+			.filter( key => !! productTypes[ key ].preselect );
+
 		this.state = {
 			error: null,
-			selected: profileItems.product_types || [],
+			selected: profileItems.product_types || preselectedProductTypes,
 		};
 
 		this.onContinue = this.onContinue.bind( this );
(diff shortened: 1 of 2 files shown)


client/profile-wizard/steps/product-types.js
```diff
@@ -26,9 +26,13 @@ class ProductTypes extends Component {
 		super();
 		const profileItems = get( props, 'profileItems', {} );
 
+		const { productTypes = {} } = getSetting( 'onboarding', {} );
+		const preselectedProductTypes = Object.keys( productTypes )
```
joshuatf, line 30: Nice!

src/Features/Onboarding.php
```diff
@@ -340,6 +340,7 @@ public static function get_allowed_product_types() {
 				'physical'      => array(
 					'label'       => __( 'Physical products', 'woocommerce-admin' ),
 					'description' => __( 'Products you ship to customers.', 'woocommerce-admin' ),
+					'preselect'   => true,
```
joshuatf, line 343: What do you think about `default` here?  I'm thinking this would keep us slightly closer to semantics used in core - https://docs.woocommerce.com/document/settings-api/

That said, if you feel strongly that `preselect` is better we can keep it.

joshuatf (approved):
Cool solution here, Bec!  Pre-approving pending one optional change.

src/Features/Onboarding.php
timmyc: Interesting idea Josh. Looking at the docs, seems like the `default` attribute could work nicely here.

becdetat: I'll go with `default` - naming things is hard ;-)

becdetat: Done in 4d802b3aed501d7cb4de4642cfeb1af2c8c53dae

PR #4703, opened 2020-06-26
Remove the Composite products option from the onboarding wizard

becdetat:
Partially addresses #4577

This removes the "Composite products" option from the product selection step in the OBW.

1. start the OBW
2. carry through to the product selection step
3. note that the "Composite products" option is no longer present


joshuatf (approved):
Thanks for this!  Changes look good :shipit:

PR #4704, opened 2020-06-26
Filters: Fix regular filters not working

psealock:
Simple filters weren't working on `main`.

Screen Shot 2020-06-26 at 2 32 50 PM

https://github.com/woocommerce/woocommerce-admin/pull/4650 added properties to the `staticParams` config object such that certain parameters wouldn't change as the user navigates a report. These are mostly used for things like chart type and page. See https://github.com/woocommerce/woocommerce-admin/pull/4650#issuecomment-647336121 for previous discussion where filter parameters were added. This broke the basic filtering because those parameters were considered static.

1. Product Report.
2. Filter by single product: Show > Single Product 
3. See the filter work (this is broken on `main`.

Also go through the tests of #4650 to make sure no regression occured.

- Open WooCommerce -> Customers
- Set up an advanced filter, make sure to press the "Filter" button to apply it
- Select the filter type dropdown and reselect "Advanced"
- Observe that the advanced filters remain
- Select the filter type dropdown and select a different filter type, then reselect Advanced
- Observe that the advanced filters remain
- Refresh the page
- Observe that the advanced filters remain
- There should be no JS errors in the console relating to missing `key` values


samueljseay (commented):
I tested this one locally and the new functionality looks good but it still causes a regression in the customer filters from #4650

reselecting "Advanced Filters" works ok, but changing to "All Customers" then back to "Advanced Filters" causes the filter to vanish. I've got a screencap of this:

regression

psealock:
Thanks for the review @samueljseay 

>reselecting "Advanced Filters" works ok, but changing to "All Customers" then back to "Advanced Filters" causes the filter to vanish. I've got a screencap of this:

Thats a good point. The original requirements of AdvancedFilters weren't that those parameters should persist in this scenario. If that is something we'd like to have, the better solution would be to elevate those parameters to a component state instead of keeping them around in the url. 

My opinion is that this shouldn't be a feature to implement at this time, but paging @becdetat and @timmyc to weigh in.

timmyc:
While not optimal I agree with Paul that the behavior of selecting a preset filter and thus clearing out the active advanced filters is how the feature was originally developed and is okay.

samueljseay (approved):
:+1:

PR #4707, opened 2020-06-26
Pills and tooltips in product step of OBW

becdetat:
Partially addresses #4577

This replaces the existing price text with a new pill component, using per-month amounts. It also adds a tooltip to each pill.

### Accessibility

### Screenshots
image

Smaller viewport:
image

- Start the OBW
- Step through to the product selection task
- Check that the layout matches the [design mockup](https://www.figma.com/file/JH9XMFUCOjfXdr3N09AHRD/On-boarding-iterations-June-'20?node-id=260%3A1491)


client/profile-wizard/style.scss
```diff
@@ -263,7 +263,7 @@
 			}
 
 			label.components-checkbox-control__label {
-				margin-left: $gap-large - $gap-smallest;
+				margin-left: $gap;
```
becdetat, line 266: This is just a small tweak to get the label to line up with the description per the design mockup

client/profile-wizard/steps/product-types.js
```diff
@@ -13,13 +13,35 @@ import { withDispatch, withSelect } from '@wordpress/data';
...
+import './product-types.scss';
+
+function getLabel( description, yearlyPrice ) {
+	if ( ! yearlyPrice ) {
+		return description;
+	}
+
+	const monthlyPrice = ( yearlyPrice / 12.0 ).toFixed( 2 );
+	const priceDescription = sprintf(
+		'$%f per month, billed annually',
```
timmyc, line 32: Looks like we missed a translation wrapper on this text.

packages/components/src/pill/pill.js
```diff
@@ -0,0 +1,7 @@
+export function Pill( { children } ) {
```
timmyc, line 1: A new component is born 🎉  - can we also get a line added to the [changelog](https://github.com/woocommerce/woocommerce-admin/blob/main/packages/components/CHANGELOG.md) about this new component?

timmyc (changes requested):
Looking really great Bec. Couple of minor comments in the code, otherwise all looking good there.

I did have a few questions on the design side of this so cc @jameskoster :

- I verified with the design that the pills should be purple. Given that we have been moving away from purple everywhere else, just wanted to see why we are keeping it in this new component?
- On narrow view ports, we can end up with the pill wrapping onto a second line. How should we handle this?

image

jameskoster:
> I verified with the design that the pills should be purple. Given that we have been moving away from purple everywhere else, just wanted to see why we are keeping it in this new component?

I'm going to confirm this with the design team in our call on Tuesday.

jameskoster:
@becdetat we decided that in this use case we'll make the pills gray :D If you could use `$medium-gray-text` for the stroke and the text, that would be splendid. Sorry for the mix up! :)

client/profile-wizard/steps/product-types.js
becdetat: Whoops, I thought that since `sprintf` came from the i18n library it included translation. Fixed in 6156c3fd24f729209be9f726284ffc98be5ffadf.

packages/components/src/pill/pill.js
becdetat: Done in 0fe3e6d7638c8f8238869080f58591942ee203ab

becdetat:
Thanks @jameskoster, I've updated the screenshot above with the new colours.

Do you have any guidance around how to handle narrow viewports and the pill pushing to the next line?

jameskoster:
Ah, sorry! Yes, I think the wrapping is ok, so long as we can fix the justification. When wrapped, if the title, pill, and description all line up nicely I think we're good :)

becdetat:
@jameskoster can you check the small viewport screenshot in the description above? Everything looks nicely aligned now.

becdetat:
@jameskoster just a follow-up to see if you can check on the layout? Cheers  :-)

packages/components/src/pill/style.scss
```diff
@@ -0,0 +1,8 @@
+.woocommerce-pill {
+	border: 1px solid $medium-gray-text;
```
psealock, line 2: FYI `medium-gray-text` [is getting removed ](https://github.com/WordPress/gutenberg/pull/23648) in the next update of base-styles in https://github.com/woocommerce/woocommerce-admin/pull/4759.

client/profile-wizard/steps/product-types.js
```diff
@@ -13,13 +13,37 @@ import { withDispatch, withSelect } from '@wordpress/data';
...
  */
 import { recordEvent } from 'lib/tracks';
+import './product-types.scss';
+
+function getLabel( description, yearlyPrice ) {
+	if ( ! yearlyPrice ) {
+		return description;
+	}
+
+	const monthlyPrice = ( yearlyPrice / 12.0 ).toFixed( 2 );
```
psealock, line 30: TIL `toFixed` 😄

psealock (commented):
Code is looking great.

>Check that the layout matches the design mockup

The continue button has drifted left somehow.

Screen Shot 2020-07-09 at 2 05 43 PM

If I remember correctly we support all the way to 320px wide displays. The pills could use just a touch of smallerness.

Screen Shot 2020-07-09 at 2 01 20 PM

I also noticed the designs have grey lines and slightly more padding between options:

Screen Shot 2020-07-09 at 2 07 04 PM

packages/components/src/pill/style.scss
jameskoster: Good catch, let's make that `gray-700` if we can :)

jameskoster:
> The pills could use just a touch of smallerness.

Dang, that is annoying! I'm gonna say I'm ok with it though as it feels like an edge case to me. 

If we can just make sure the pills use the [Caption text style](https://wordpress.github.io/gutenberg/?path=/story/components-text--default) (including the line-height) that will at least help a little bit.

becdetat:
Ok, this is much closer to the design now:

image
image

I did adjust the font size of the pills down from 12px to 11px, which fixes the overflow issue at 320px wide viewports.

jameskoster (changes requested):
Could we only adjust the font-size down to 11px on small screens? That way it will maintain consistency with the `caption` text style on all other viewport sizes :)

becdetat:
@jameskoster done in 38ccaa0 with a breakpoint on screens 320px and smaller.

timmyc:
@jameskoster could you give this one final 👀 from the design side of things please?

jameskoster (approved):

timmyc (approved):
Took for one last spin, all is looking great 🚢

PR #4711, opened 2020-06-26
Preventing refresh after answering a survey note

octaedro:
Following up with [PR 4686](https://github.com/woocommerce/woocommerce-admin/pull/4686)

This PR adds the code necessary to prevent a webpage refresh after answering a survey note.

### Accessibility

### Screenshots
Screen Capture on 2020-06-26 at 15-26-20

- Be sure you have the note `wc-admin-insight-first-sale` (WC_Admin_Notes_Insight_First_Sale.php) available and that it is type: `survey`.

If you already have it but with type `info` or with status `actioned` or with is_deleted `1` you should execute this SQL sentence:
````
UPDATE `wp_wc_admin_notes` SET `type` = 'survey', `status` = 'unactioned', `is_deleted` = 0 WHERE `wp_wc_admin_notes`.`name` = 'wc-admin-insight-first-sale'; 
````

- Go to the `Home` screen or open the `Inbox panel` pressing `Inbox` in the header.
- Answer the survey (`yes` or `no`, it's the same).
- The note should change its status to `active`, and the message `Thanks for your feedback` should replace the options `yes` and `no`.
- The webpage shouldn't reload.

### Changelog Note:
Fix: Preventing refresh after answering a survey note

### DIFF
--- a/client/header/activity-panel/panels/inbox/action.js
+++ b/client/header/activity-panel/panels/inbox/action.js
@@ -64,13 +64,20 @@ class InboxNoteAction extends Component {
 		const { action, dismiss, label } = this.props;
 		const isPrimary = dismiss || action.primary;
 
+		let actionHref;
+		if ( action ) {
+			if ( action.url ) {
+				actionHref = action.url;
+			}
+		}
+
 		return (
 			<Button
 				isPrimary={ isPrimary }
 				isSecondary={ ! isPrimary }
 				isBusy={ this.state.inAction }
 				disabled={ this.state.inAction }
-				href={ action ? action.url : undefined }
+				href={ actionHref }
 				onClick={ this.handleActionClick }
 			>
 				{ dismiss ? label : action.label }
--- a/client/header/activity-panel/panels/inbox/action.js
+++ b/client/header/activity-panel/panels/inbox/action.js
@@ -64,20 +64,17 @@ class InboxNoteAction extends Component {
 		const { action, dismiss, label } = this.props;
 		const isPrimary = dismiss || action.primary;
 
-		let actionHref;
-		if ( action ) {
-			if ( action.url ) {
-				actionHref = action.url;
-			}
-		}
-
 		return (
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: client/header/activity-panel/panels/inbox/action.js [hunk 1]
```diff
@@ -64,13 +64,20 @@ class InboxNoteAction extends Component {
...
+		}
+
 		return (
 			<Button
 				isPrimary={ isPrimary }
 				isSecondary={ ! isPrimary }
 				isBusy={ this.state.inAction }
 				disabled={ this.state.inAction }
-				href={ action ? action.url : undefined }
+				href={ actionHref }
```
samueljseay, line 80: Hey @octaedro  I wonder, could we make this code a little more succinct by removing the nested if on line 67 and use a ternary here instead. I think that although the ternary is a little long its a bit easier to understand quickly in this case. 😃 

```suggestion
				href={action && action.url && action.url.length ? action.url : undefined}
```

psealock: Nice, this check for length of string is a good improvement and more thorough .

psealock (approved):
Nice work @octaedro ! This is testing well for me!

I needed to cherry pick this commit onto #4686 to test this. I pushed on `test/4656-with-refresh-changes` to test.

Once the issue that @samueljseay noted is addressed, this is good to merge.

client/header/activity-panel/panels/inbox/action.js
octaedro: Hi @psealock and @samueljseay, thank you for your review. 
Nice catch Sam! The change you proposed sounds very good 😀
I did it in the commit 98aba7c

octaedro:
>Once the issue that @samueljseay noted is addressed, this is good to merge.

Hey @psealock,
as the modification that Sam suggested was addressed in the commit 98aba7c I will merge this PR.

PR #4717, opened 2020-06-29
Update connection flow for Bundle UI

### LINKED ISSUE #4579: Store Profiler - Business details step: install free extensions in 1 click

joshuatf:
Fixes #4579 

* Triggers the Jetpack connection flow from any step on profiler completion if Jetpack is installed and not connected.  Based on slack discussion p1593073481360000-slack-woo-start-manage
* Shows the bundle install UI based on segment defined in #4579 
* Updates cached plugins store in profiler to show/hide benefits step.

One minor annoyance and possible update here or in a future PR is that updating the profile to `complete` throws the user into the dashboard, so if we're still trying to accomplish tasks (like connecting to Jetpack) we need to hide the UI to prevent the "flash" of screens.  One possible fix for this is to create a route or show the profiler when `step` is in the query.

### Screenshots
Screen Shot 2020-06-29 at 3 09 10 PM

**Bundle UI Segmentation**
1. Walk through the profiler, selecting US as the store default country and "Health & Beauty" or "Fashion" for the store industry.
1. Make sure the new bundle UI is shown on the business details step.
1. Go back and update to not include one of those industries or change to a non-US country.
1. Make sure the old business extensions UI is now shown on the Business Details step.

**Benefits plugin caching**
1. Disable Jetpack and/or WCS.
1. Use a non-US store to trigger the old business extensions UI.
1. Walk through to the benefits step and complete.
1. Make sure this step completes and works as expected.
1. Disable Jetpack and/or WCS.
1. With the new bundle UI enabled walk through the profiler, installing extensions.
1. Make sure the benefits page is not shown on completion.

**Jetpack connection flow**
1. Use a non-US store to trigger the old business extensions UI.
1. With Jetpack installed (can be installed during profiler run) but not connected, walk through the profiler.
1. Make sure that you are redirected to the connection flow after completing the benefits step.
1. Disconnected Jetpack.
1. Use a US store and selected industry to trigger the bundle UI.
1. Make sure you're redirected the Jetpack connection flow after the theme step.

### DIFF
--- a/client/profile-wizard/index.js
+++ b/client/profile-wizard/index.js
@@ -5,24 +5,26 @@ import { __ } from '@wordpress/i18n';
 import { Component, createElement, Fragment } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import { identity, pick } from 'lodash';
-import { withDispatch } from '@wordpress/data';
+import { withDispatch, __experimentalResolveSelect } from '@wordpress/data';
 
 /**
  * WooCommerce dependencies
  */
-import { updateQueryString } from '@woocommerce/navigation';
+import { getAdminLink } from '@woocommerce/wc-admin-settings';
 import {
 	ONBOARDING_STORE_NAME,
 	OPTIONS_STORE_NAME,
 	PLUGINS_STORE_NAME,
 	withPluginsHydration,
 } from '@woocommerce/data';
+import { updateQueryString } from '@woocommerce/navigation';
 
 /**
  * Internal dependencies
  */
 import Benefits from './steps/benefits';
 import BusinessDetails from './steps/business-details';
+import { createNoticesFromResponse } from 'lib/notices';
 import Industry from './steps/industry';
 import ProductTypes from './steps/product-types';
 import ProfileWizardHeader from './header';
@@ -36,10 +38,6 @@ import './style.scss';
 class ProfileWizard extends Component {
 	constructor( props ) {
 		super( props );
-		this.state = {
-			cartRedirectUrl: null,
-		};
-
 		this.activePlugins = props.activePlugins;
 		this.goToNextStep = this.goToNextStep.bind( this );
 	}
@@ -103,13 +101,6 @@ class ProfileWizard extends Component {
 	}
 
 	componentWillUnmount() {
-		const { cartRedirectUrl } = this.state;
-
-		if ( cartRedirectUrl ) {
-			document.body.classList.add( 'woocommerce-admin-is-loading' );
-			window.location = cartRedirectUrl;
-		}
-
 		document.documentElement.classList.add( 'wp-toolbar' );
 		document.body.classList.remove( 'woocommerce-onboarding' );
 		document.body.classList.remove( 'woocommerce-profile-wizard__body' );
@@ -211,8 +202,12 @@ class ProfileWizard extends Component {
 	}
 
 	completeProfiler() {
-		const { notes, updateNote, updateProfileItems } = this.props;
-		updateProfileItems( { completed: true } );
+		const {
+			isJetpackConnected,
+			notes,
+			updateNote,
+			updateProfileItems,
+		} = this.props;
 		recordEvent( 'storeprofiler_complete' );
 
 		const profilerNote = notes.find(
@@ -221,6 +216,37 @@ class ProfileWizard extends Component {
 		if ( profilerNote ) {
 			updateNote( profilerNote.id, { status: 'actioned' } );
 		}
+
+		updateProfileItems( { completed: true } ).then( () => {
+			if (
+				this.activePlugins.includes( 'jetpack' ) &&
+				! isJetpackConnected
+			) {
+				this.connectJetpack();
+			}
+		} );
+	}
+
+	connectJetpack() {
+		const { getJetpackConnectUrl, getPluginsError } = this.props;
+
+		document.body.classList.add( 'woocommerce-admin-is-loading' );
+
+		getJetpackConnectUrl( {
+			redirect_url: getAdminLink(
+				'admin.php?page=wc-admin&reset_profiler=0'
+			),
+		} ).then( ( url ) => {
+			const error = getPluginsError( 'getJetpackConnectUrl' );
+			if ( error ) {
+				createNoticesFromResponse( error );
+				document.body.classList.remove(
+					'woocommerce-admin-is-loading'
+				);
+				return;
+			}
+			window.location = url;
+		} );
 	}
 
 	render() {
@@ -254,7 +280,9 @@ export default compose(
 		const { getProfileItems, getOnboardingError } = select(
 			ONBOARDING_STORE_NAME
 		);
-		const { getActivePlugins } = select( PLUGINS_STORE_NAME );
+		const { getActivePlugins, getPluginsError } = select(
+			PLUGINS_STORE_NAME
+		);
 
 		const notesQuery = {
 			page: 1,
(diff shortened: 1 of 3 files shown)


psealock (commented):
This is all looking really nice, and changes all appear pretty solid.

In the Benefits plugin caching testing instructions I can't get step 7 to work.

>Make sure the benefits page is not shown on completion.

I tested on Ephemeral. Is it working for you?

client/profile-wizard/index.js
```diff
@@ -203,16 +201,63 @@ class ProfileWizard extends Component {
...
+				} )
+			);
+		}
+
+		Promise.all( promises ).then( ( responses ) => {
+			if ( ! shouldConnectJetpack ) {
+				return;
+			}
+
+			const jetpackConnectUrl = responses[ 1 ];
```
becdetat, line 247: This feels a little brittle, could you add a comment explaining why it's index 1? That way if a new promise is inserted at least there's an explanation why the jetpack redirect has stopped working.

becdetat:
@joshuatf in "Jetpack connection flow" step 6, instead of directing to the Jetpack connection flow, it directs to the Mailchimp connection flow. Otherwise everything tests well.

client/profile-wizard/index.js
joshuatf: Refactored in https://github.com/woocommerce/woocommerce-admin/pull/4717/commits/fc6a9c4900831c872eacc61c26538b31db1fa174 so we can handle redirection regardless of whether or not a promise exists.

joshuatf:
>@joshuatf in "Jetpack connection flow" step 6, instead of directing to the Jetpack connection flow, it directs to the Mailchimp connection flow. Otherwise everything tests well.

Thanks for the review!  I see the Mailchimp redirect as well.  I'll open an issue to address this in a follow-up since it's not directly related to this PR.

I also updated the logic for redirecting in https://github.com/woocommerce/woocommerce-admin/pull/4717/commits/fc6a9c4900831c872eacc61c26538b31db1fa174 in anticipation of https://github.com/woocommerce/woocommerce-admin/pull/4721 being merged in.

>In the Benefits plugin caching testing instructions I can't get step 7 to work.

@psealock This does work for me.  Did you hit the new Bundle UI and still see the benefits page at the end of the profiler?  I wasn't able to reproduce this.

jeffstieler (changes requested):
The info popover is not navigable using a screen reader so there is no context for the preselected option.

We should also provide more context to screen readers - the info button's label is just "info", not something descriptive like "get information about recommended free features".

joshuatf:
>the info button's label is just "info", not something descriptive like "get information about recommended free features".

Good point.

>The info popover is not navigable using a screen reader so there is no context for the preselected option.

Do you have a suggestion for this @jeffstieler?

jeffstieler:
> > The info popover is not navigable using a screen reader so there is no context for the preselected option.
> 
> Do you have a suggestion for this @jeffstieler?

To elaborate, you can open the popover with the screen reader, but it selects the first checkmark icon and then it's not possible to "move around" within the popover to read the content. It's just broken.

Ideally, we'd hide the checkmarks from the screen reader since they're purely visual and the text items describing the features would be able to be read by the screen reader.

I used VoiceOver on mac to test.

joshuatf:
>but it selects the first checkmark icon and then it's not possible to "move around" within the popover to read the content. It's just broken.

This sounds like the `Popover` selecting the first element inside of itself to gain focus.  I'm not sure if it's possible to highlight focus of specific elements and retain the blur effect needing to close this popover.  `focusOnMount="container"` does not seem to behave as expected.

Somewhat related, I have an open issue for this in Gutenberg ( https://github.com/WordPress/gutenberg/issues/23298 ) where we also hit similar behavior with the date dropdowns not being able to close without focus.

I'll play around with this and see if hiding items from the screen reader solves any of this.

joshuatf:
@jeffstieler Addressed some of the accessibility concerns around the UI in https://github.com/woocommerce/woocommerce-admin/pull/4717/commits/cd55759f997ac24f8f99bdf533ed92f5e7ee6ea1.

This seems to fix the button text and I'm able to drill down and read all group text (features) in the popover, however, I can still receive focus on the icons in the popover even though the text is no longer read from them.  I'm not sure if that's expected behavior or not from VoiceOver.

jeffstieler:
> This seems to fix the button text and I'm able to drill down and read all group text (features) in the popover, however, I can still receive focus on the icons in the popover even though the text is no longer read from them.

Hmm, I'm not able to get this behavior.

When I click the info button with VO, it says:

> Manage your store on the go with the WooCommerce mobile app  Accept credit cards with WooCommerce Payments and 6 more items, group

Then instructs me:

> You are currently on a group. To interact with items in this group, press Control-Option-Shift-Down Arrow.

When I try doing that, nothing gets selected.

joshuatf:
>You are currently on a group. To interact with items in this group, press Control-Option-Shift-Down Arrow.

Hmmm, I think this is for dropping into a group such as a table or menu, but typically VO + left/right would be used to navigate between the items.  (See "read next/previous item" on this page - https://dequeuniversity.com/screenreaders/voiceover-keyboard-shortcuts ).

I don't ever receive the help text to drop into the group and attempting to doesn't allow me.  When interacting with the button, I'm immediately read the first item and can navigate between items using VO + left/right.  Is your interaction here different @jeffstieler?

https://cloudup.com/cu3htyo4pr2

jeffstieler:
Your cloudup link says "upload failed".

> I don't ever receive the help text to drop into the group and attempting to doesn't allow me. When interacting with the button, I'm immediately read the first item and can navigate between items using VO + left/right. Is your interaction here different?

Yes, I have a different experience:

https://cloudup.com/cWiVSN_ISgK

joshuatf:
>Your cloudup link says "upload failed".

Sorry about that.  Interwebs is not my friend today.  Re-uploaded the gif above.

>https://cloudup.com/cWiVSN_ISgK

Which browser is this running in @jeffstieler?  I can't repro the behavior in Safari, but I'm able to in Chrome.  As I understand it, VoiceOver is not fully compatible with Chrome and contains a number of these bugs.

jeffstieler:
> I can't repro the behavior in Safari, but I'm able to in Chrome.

🔔 🔔 🔔  we have a winner!

I was unaware or had forgotten to use Safari with VoiceOver. Everything works great in Safari 🙌 .

jeffstieler (approved):
This tests well and the code looks good. Thanks for fixing the screen reader accessibility in the popover!

Resolve the merge conflict and 🚢  it.

PR #4721, opened 2020-06-30
Added skip profiler functionality

### LINKED ISSUE #4568: Task List - Mark Store details step as incomplete when skipping the profiler
**User story:**
As a new merchant
I want to be able to easily access the store profiler
So I can complete the store information after skipping it

**Description:** with the addition of a skip link in the profiler (#4566), the Store details step should be marked as incomplete if the user skips the store profiler flow.

I'm not sure if this logic was already implemented on #4096  

**Acceptance Criteria:**
- [ ] Display the Store details step as incomplete if the user skipped the store profiler flow.

**Event tracking:**
Currently, we don't have an event that tracks clicks in the task list, so probably is the easiest way to track if users are clicking on the Store Details step:

Event name: `wcadmin_tasklist_click`
Event prop: `task_name` (value: task name)
Event description: When the user clicks to access a specific task

octaedro:
Fixes #4568-partially

This PR adds the functionality to skip the onboarding profiler. Also, it leaves the `Store details` as incomplete in the `Home` screen.

### Accessibility

### Screenshots
skip_profiler1
skip_profiler2

To test this functionality I created the branch `add/4568_test` where I added a link to be able to skip the onboarding profiler.
1. Delete the profiler information with these two SQL sentences:
````
UPDATE `wp_options` SET `option_value` = 'a:0:{}' WHERE `wp_options`.`option_name` = 'woocommerce_onboarding_profile';
UPDATE `wp_options` SET `option_value` = 'no' WHERE `wp_options`.`option_name` = 'woocommerce_task_list_complete';
````
2. Activate the profiler setup wizard (going to `/wp-admin/admin.php?page=wc-settings` -> Setup wizard -> Profile setup wizard -> `Enable`).
3. Fill out a couple of onboarding steps.
4. Press `Skip`.
skip_profiler3
5. Verify the task list on the `Home` screen has the `Store details` marked as incomplete.
6. Press `Store details`. The profiler should open again.
7. After pressing `Skip`, the profile setup wizard (step 2) should appear as disabled. Verify it's disabled when the profiler was skipped and enabled when it wasn't.

/cc @samueljseay This PR is related to the issue [4566](https://github.com/woocommerce/woocommerce-admin/issues/4566). I know Sam is working on it.

### Changelog Note:
Dev: Added skip profiler functionality

### DIFF
--- a/client/dashboard/index.js
+++ b/client/dashboard/index.js
@@ -9,7 +9,7 @@ import { identity } from 'lodash';
 /**
  * WooCommerce dependencies
  */
-import { getSetting } from '@woocommerce/wc-admin-settings';
+import { getAdminLink, getSetting } from '@woocommerce/wc-admin-settings';
 import {
 	ONBOARDING_STORE_NAME,
 	withOnboardingHydration,
@@ -26,22 +26,19 @@ const CustomizableDashboard = lazy( () =>
 	import( /* webpackChunkName: "customizable-dashboard" */ './customizable' )
 );
 
-const ProfileWizard = lazy( () =>
-	import( /* webpackChunkName: "profile-wizard" */ '../profile-wizard' )
-);
-
 class Dashboard extends Component {
 	render() {
 		const { path, profileItems, query } = this.props;
+		const { completed: profileCompleted, skipped: profileSkipped, step: profilerStep } = profileItems;
 		if (
 			isOnboardingEnabled() &&
-			! profileItems.completed &&
+			! profileCompleted &&
+			! profileSkipped &&
 			! window.wcAdminFeatures.homescreen
 		) {
-			return (
-				<Suspense fallback={ <Spinner /> }>
-					<ProfileWizard query={ query } />
-				</Suspense>
+			const lastStep = profilerStep ? `&step=${ profilerStep }` : '';
+			window.location = getAdminLink(
+				`admin.php?page=wc-admin&path=/profiler${ lastStep }`
 			);
 		}
 
--- a/client/dashboard/index.js
+++ b/client/dashboard/index.js
@@ -15,6 +15,7 @@ import {
 	withOnboardingHydration,
 } from '@woocommerce/data';
 import { Spinner } from '@woocommerce/components';
+import { getHistory } from '@woocommerce/navigation';
 
 /**
  * Internal dependencies
@@ -29,7 +30,11 @@ const CustomizableDashboard = lazy( () =>
 class Dashboard extends Component {
 	render() {
 		const { path, profileItems, query } = this.props;
-		const { completed: profileCompleted, skipped: profileSkipped, step: profilerStep } = profileItems;
+		const {
+			completed: profileCompleted,
+			skipped: profileSkipped,
+			step: profilerStep,
+		} = profileItems;
 		if (
 			isOnboardingEnabled() &&
 			! profileCompleted &&
@@ -37,9 +42,10 @@ class Dashboard extends Component {
 			! window.wcAdminFeatures.homescreen
 		) {
 			const lastStep = profilerStep ? `&step=${ profilerStep }` : '';
-			window.location = getAdminLink(
+			const href = getAdminLink(
 				`admin.php?page=wc-admin&path=/profiler${ lastStep }`
 			);
+			getHistory().push( href );
 		}
 
 		if ( window.wcAdminFeatures[ 'analytics-dashboard/customizable' ] ) {
--- a/client/dashboard/index.js
+++ b/client/dashboard/index.js
@@ -30,10 +30,8 @@ const CustomizableDashboard = lazy( () =>
 class Dashboard extends Component {
 	render() {
 		const { path, profileItems, query } = this.props;
-		const {
-			completed: profileCompleted,
-			skipped: profileSkipped,
-		} = profileItems;
+		const { completed: profileCompleted, skipped: profileSkipped } =
+			profileItems || {};
 		if (
 			isOnboardingEnabled() &&
 			! profileCompleted &&
--- a/client/dashboard/index.js
+++ b/client/dashboard/index.js
@@ -9,13 +9,13 @@ import { identity } from 'lodash';
 /**
  * WooCommerce dependencies
  */
-import { getAdminLink, getSetting } from '@woocommerce/wc-admin-settings';
+import { getSetting } from '@woocommerce/wc-admin-settings';
 import {
 	ONBOARDING_STORE_NAME,
 	withOnboardingHydration,
 } from '@woocommerce/data';
 import { Spinner } from '@woocommerce/components';
-import { getHistory } from '@woocommerce/navigation';
+import { getHistory, getNewPath } from '@woocommerce/navigation';
 
 /**
  * Internal dependencies
@@ -41,11 +41,8 @@ class Dashboard extends Component {
 			! profileSkipped &&
 			! window.wcAdminFeatures.homescreen
 		) {
-			const lastStep = profilerStep ? `&step=${ profilerStep }` : '';
-			const href = getAdminLink(
-				`admin.php?page=wc-admin&path=/profiler${ lastStep }`
-			);
-			getHistory().push( getNewPath( {}, '/profiler', {} ) );
+			const lastStep = profilerStep ? { step: profilerStep } : {};
+			getHistory().push( getNewPath( {}, `/profiler`, lastStep ) );
 		}
 
 		if ( window.wcAdminFeatures[ 'analytics-dashboard/customizable' ] ) {
--- a/client/dashboard/index.js
+++ b/client/dashboard/index.js
@@ -33,7 +33,6 @@ class Dashboard extends Component {
 		const {
 			completed: profileCompleted,
 			skipped: profileSkipped,
-			step: profilerStep,
 		} = profileItems;
 		if (
 			isOnboardingEnabled() &&
@@ -41,8 +40,7 @@ class Dashboard extends Component {
 			! profileSkipped &&
 			! window.wcAdminFeatures.homescreen
 		) {
-			const lastStep = profilerStep ? { step: profilerStep } : {};
-			getHistory().push( getNewPath( {}, `/profiler`, lastStep ) );
+			getHistory().push( getNewPath( {}, `/profiler`, {} ) );
 		}
 
 		if ( window.wcAdminFeatures[ 'analytics-dashboard/customizable' ] ) {
--- a/client/dashboard/index.js
+++ b/client/dashboard/index.js
@@ -45,7 +45,7 @@ class Dashboard extends Component {
 			const href = getAdminLink(
 				`admin.php?page=wc-admin&path=/profiler${ lastStep }`
 			);
-			getHistory().push( href );
+			getHistory().push( getNewPath( {}, '/profiler', {} ) );
 		}
 
 		if ( window.wcAdminFeatures[ 'analytics-dashboard/customizable' ] ) {
--- a/client/homescreen/index.js
+++ b/client/homescreen/index.js
@@ -2,14 +2,12 @@
  * External dependencies
  */
 import { compose } from '@wordpress/compose';
-import { Suspense, lazy } from '@wordpress/element';
 import { identity } from 'lodash';
 
 /**
  * WooCommerce dependencies
  */
-import { getSetting } from '@woocommerce/wc-admin-settings';
-import { Spinner } from '@woocommerce/components';
+import { getAdminLink, getSetting } from '@woocommerce/wc-admin-settings';
 import {
 	ONBOARDING_STORE_NAME,
 	withOnboardingHydration,
@@ -21,18 +19,14 @@ import {
 import withSelect from 'wc-api/with-select';
 import { isOnboardingEnabled } from 'dashboard/utils';
 
-const ProfileWizard = lazy( () =>
-	import( /* webpackChunkName: "profile-wizard" */ '../profile-wizard' )
-);
 import Layout from './layout';
 
 const Homescreen = ( { profileItems, query } ) => {
-	const { completed: profilerCompleted, skipped: profilerSkipped } = profileItems;
+	const { completed: profilerCompleted, skipped: profilerSkipped, step: profilerStep } = profileItems;
 	if ( isOnboardingEnabled() && ! profilerCompleted && ! profilerSkipped ) {
-		return (
-			<Suspense fallback={ <Spinner /> }>
-				<ProfileWizard query={ query } />
-			</Suspense>
+		const lastStep = profilerStep ? `&step=${ profilerStep }` : '';
+		window.location = getAdminLink(
+			`admin.php?page=wc-admin&path=/profiler${ lastStep }`
 		);
 	}
 
--- a/client/homescreen/index.js
+++ b/client/homescreen/index.js
@@ -27,7 +27,8 @@ const ProfileWizard = lazy( () =>
 import Layout from './layout';
 
 const Homescreen = ( { profileItems, query } ) => {
-	if ( isOnboardingEnabled() && ! profileItems.completed ) {
+	const { completed: profilerCompleted, skipped: profilerSkipped } = profileItems;
+	if ( isOnboardingEnabled() && ! profilerCompleted && ! profilerSkipped ) {
 		return (
 			<Suspense fallback={ <Spinner /> }>
 				<ProfileWizard query={ query } />
--- a/client/homescreen/index.js
+++ b/client/homescreen/index.js
@@ -12,6 +12,7 @@ import {
 	ONBOARDING_STORE_NAME,
 	withOnboardingHydration,
 } from '@woocommerce/data';
+import { getHistory } from '@woocommerce/navigation';
 
 /**
  * Internal dependencies
@@ -22,12 +23,17 @@ import { isOnboardingEnabled } from 'dashboard/utils';
 import Layout from './layout';
 
 const Homescreen = ( { profileItems, query } ) => {
-	const { completed: profilerCompleted, skipped: profilerSkipped, step: profilerStep } = profileItems;
+	const {
+		completed: profilerCompleted,
+		skipped: profilerSkipped,
+		step: profilerStep,
+	} = profileItems;
 	if ( isOnboardingEnabled() && ! profilerCompleted && ! profilerSkipped ) {
 		const lastStep = profilerStep ? `&step=${ profilerStep }` : '';
-		window.location = getAdminLink(
+		const href = getAdminLink(
 			`admin.php?page=wc-admin&path=/profiler${ lastStep }`
 		);
+		getHistory().push( href );
 	}
 
 	return <Layout query={ query } />;
--- a/client/homescreen/index.js
+++ b/client/homescreen/index.js
@@ -7,12 +7,12 @@ import { identity } from 'lodash';
 /**
  * WooCommerce dependencies
  */
-import { getAdminLink, getSetting } from '@woocommerce/wc-admin-settings';
+import { getSetting } from '@woocommerce/wc-admin-settings';
 import {
 	ONBOARDING_STORE_NAME,
 	withOnboardingHydration,
 } from '@woocommerce/data';
-import { getHistory } from '@woocommerce/navigation';
+import { getHistory, getNewPath } from '@woocommerce/navigation';
 
 /**
  * Internal dependencies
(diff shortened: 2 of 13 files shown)

### CODE REVIEW

File: src/Features/Onboarding.php [hunk 1]
```diff
@@ -921,7 +922,8 @@ public static function add_help_tab() {
 			$task_list_hidden = get_option( 'woocommerce_task_list_hidden', 'no' );
 			$onboarding_data  = get_option( self::PROFILE_DATA_OPTION, array() );
 			$is_completed     = isset( $onboarding_data['completed'] ) && true === $onboarding_data['completed'];
-			$is_enabled       = ! $is_completed;
+			$is_skipped       = isset( $onboarding_data['skipped'] ) && true === $onboarding_data['skipped'];
+			$is_enabled       = ! $is_completed && ! $is_skipped;
```
joshuatf, line 926: It looks like we use this logic several times throughout the code (`! profilerCompleted && ! profilerSkipped`).  It would be great if we could somehow consolidate this logic into a single place so it's less error prone if we need to change it again.

I've also been thinking about how this works with being able to access the profiler after completion/skip and I see a couple potential ways we can improve the existing setup here:

1. Introduce a setting `isProfilerEnabled` from the server that contains this logic and allows us to toggle on/off if we need to override the logic to access the profiler.

2. Always allow the profiler to be accessed through its own route (or if some query param like `step` is defined) and let the UI redirect to the profiler if not complete or skipped.

I'm leaning towards option 2 because this gives us a lot of flexibility and would simplify issues like hiding the UI during connection (exposed in https://github.com/woocommerce/woocommerce-admin/pull/4717), but would love more feedback on this idea.

joshuatf (commented):
Nice job on this one, Fernando!  I left some thoughts about a possible refactor here related to existing code debt we have.

While these aren't caused by this PR, I think it may be a good opportunity for us to do some tidying while we're here.  Feel free to ping me if I can clarify or help with anything here!

src/Features/Onboarding.php
jeffstieler: At first glance option 2 makes a lot of sense and seems to provide opportunity to clean up a lot of conditional logic we have - I'm thinking about the "home" route render and the forced postback currently required to revisit the OBW from the home screen. Perhaps the only gotchas might be changing existing Tracks events?

joshuatf: Thanks for your insights on this, Jeff!

>I'm thinking about the "home" route render and the forced postback currently required to revisit the OBW from the home screen.

This, the soon-to-be task to revisit the profiler, and oddities behind the connection flow make it seem like the right decision to me as well.

>Perhaps the only gotchas might be changing existing Tracks events?

Good question.  I think the main one that might be `storeprofiler_toggled` (`wcadmin_storeprofiler_toggled`) which would no longer be necessary.  Any others you see @jeffstieler?

@pmcpinto are we currently using this or is there any information you think we might miss out on from this event?  I imagine we could still determine if a user revisits after completion with the other existing events.

octaedro: Thank you @joshuatf and @jeffstieler for your suggestions, I made the changes you mentioned in the commit 74f344f.
As soon as Pedro confirms the things related to the event recording I will add the necessary modifications.

@joshuatf could you give another look at this PR?

pmcpinto:
> @pmcpinto are we currently using this or is there any information you think we might miss out on from this event? I imagine we could still determine if a user revisits after completion with the other existing events.

How is this event going to be affected by this change? If the user completes/dismisses the task list and then goes to Help - Setup Wizard - Profile Setup Wizard - Enable, the event will be triggered, correct?

Just a reminder that we aim to include a Dismiss button in the Store Profiler task (#4733), I'm not sure if it will affect some of the logic that's being implemented in this PR. Tks!

octaedro:
> How is this event going to be affected by this change? If the user completes/dismisses the task list and then goes to Help - Setup Wizard - Profile Setup Wizard - Enable, the event will be triggered, correct?

Hi @pmcpinto,
The answer to your question is `yes`, but also this PR  introduces a couple of changes to the way the OBW is accessed. Before it was only possible to access it after executing the method `reset_profiler` (following the steps you mentioned in your comment). This method records the event `storeprofiler_toggled`. Now it's also possible to access it directly using URL: `wp-admin/admin.php?page=wc-admin&path=%2Fprofiler`.
This means that the OBW could be visited without recording the event `storeprofiler_toggled`.
Do you think we should add or modify any of the events? Maybe the event `storeprofiler_toggled` is not necessary anymore. What do you think?

pmcpinto:
@octaedro thanks for the clarification. I think that we should keep `wcadmin_storeprofiler_toggled`, to understand if the number of users enabling the profiler through the help tab is going to decrease.

And in #4568 I asked to implement additional event tracking around clicking in the task list, which allows checking how many users are accessing the profiler through the task list.

octaedro:
Thank you @pmcpinto 🙂
I will leave `wcadmin_storeprofiler_toggled `.

I think that now this PR is ready for another look before merging. I will put this in Slack to see if anyone has time.

client/dashboard/index.js
```diff
@@ -26,22 +26,19 @@ const CustomizableDashboard = lazy( () =>
...
+			! profileSkipped &&
 			! window.wcAdminFeatures.homescreen
 		) {
-			return (
-				<Suspense fallback={ <Spinner /> }>
-					<ProfileWizard query={ query } />
-				</Suspense>
+			const lastStep = profilerStep ? `&step=${ profilerStep }` : '';
+			window.location = getAdminLink(
+				`admin.php?page=wc-admin&path=/profiler${ lastStep }`
```
psealock, line 41: I think using React Router to change route can avoid loading the page twice. Here is an example:

https://github.com/woocommerce/woocommerce-admin/blob/e09675584236dceaa1f91fcb64f11b69ad52f9cc/client/layout/controller.js#L294

client/homescreen/index.js
```diff
@@ -21,17 +19,14 @@ import {
...
-	if ( isOnboardingEnabled() && ! profileItems.completed ) {
-		return (
-			<Suspense fallback={ <Spinner /> }>
-				<ProfileWizard query={ query } />
-			</Suspense>
+	const { completed: profilerCompleted, skipped: profilerSkipped, step: profilerStep } = profileItems;
+	if ( isOnboardingEnabled() && ! profilerCompleted && ! profilerSkipped ) {
+		const lastStep = profilerStep ? `&step=${ profilerStep }` : '';
+		window.location = getAdminLink(
+			`admin.php?page=wc-admin&path=/profiler${ lastStep }`
```
psealock, line 29: same here about React Router.

client/profile-wizard/index.js
```diff
@@ -215,6 +221,16 @@ class ProfileWizard extends Component {
...
 	}
 
+	async skipProfiler() {
+		const { updateProfileItems } = this.props;
+		await updateProfileItems( {
+			skipped: true,
+			step: this.getCurrentStep().key,
+		} );
+		recordEvent( 'storeprofiler_store_details_skip' );
+		window.location = getAdminLink( 'admin.php?page=wc-admin' );
```
psealock, line 231: Same here.

Also, does it make sense to `await` the return of the profile items request here? Can we just navigate to the homescreen right away?

psealock (approved):
Nice @octaedro, this is looking great and testing well. I'm just noting a few places where it makes sense to probably use React Router's `getHistory` to push a new url in case changing `window.location` causes a full page refresh. I haven't looked into it and that may not even be cause a full page refresh, so I'm approving.

Then there is a comment about removing an `await` that may not be needed.

client/dashboard/index.js
octaedro: Hi @psealock, great idea!
I changed it in the commit ebe323c

client/homescreen/index.js
octaedro: I did the same in the commit ebe323c

client/profile-wizard/index.js
octaedro: And I changed it here too 😀

Even though it looks like the `await` is useless, it is necessary to be sure the `updateProfileItems` has finished before the redirection. Otherwise, when the home screen is loading, the check could fail and it would redirect the user back to the OBW.

client/profile-wizard/index.js
```diff
@@ -215,6 +221,17 @@ class ProfileWizard extends Component {
 		}
 	}
 
+	async skipProfiler() {
+		const { updateProfileItems } = this.props;
+		await updateProfileItems( {
```
mattsherman, line 226: If `updateProfileItems` errors, how is it handled?

psealock:
Thanks @octaedro changes look good. @mattsherman [brings up a good point](https://github.com/woocommerce/woocommerce-admin/pull/4721#pullrequestreview-445959666) to address and then I think this one is ready to merge.

octaedro:
Hi @psealock and @mattsherman, thank you for your review 🙂
Nice catch! You're right, we need to handle errors for `updateProfileItems`.
Taking a look at the code I realized we are not doing it in a couple of places ([here](https://github.com/woocommerce/woocommerce-admin/blob/main/client/profile-wizard/steps/theme/index.js#L99) and a few times in [this file](https://github.com/woocommerce/woocommerce-admin/blob/main/client/profile-wizard/index.js)).
If it's ok with you I will merge this PR and I'm going to create a follow-up PR to handle errors every time we use this method.

client/layout/controller.js
```diff
@@ -179,6 +182,20 @@ export const getPages = () => {
 		} );
 	}
 
+	if ( window.wcAdminFeatures.profiler ) {
```
joshuatf, line 185: I think this should just use the `onboarding` feature flag.

```suggestion
	if ( window.wcAdminFeatures.onboarding ) {
```

config/plugin.json
```diff
@@ -13,6 +13,7 @@
 		"shipping-label-banner": true,
 		"store-alerts": true,
 		"wcpay": true,
-		"homescreen": true
+		"homescreen": true,
+		"profiler": true
```
joshuatf, line 17: I don't think we need to introduce a new feature flag, do we?

client/profile-wizard/index.js
joshuatf: I think error handling might be good to handle before merging since this could end in being redirected back here if updating fails ( failure -> navigate to dashboard -> redirect back to profiler since not skipped ).

Maybe at least we could move the below `recordEvent` and `geHistory` calls inside `updateProfileItems().then( // ... here );` so they only fire on success.

joshuatf (commented):
Fantastic work on this, @octaedro 🎉 

Really nice to have this update in place.  Left a couple (hopefully) quick comments that might be worth addressing before merge.

octaedro:
Thank you @joshuatf for your review. I added the error handling in the commit aceb258.

config/plugin.json
octaedro: Good idea! I removed it in the commits 9940b96 and 31c4154.

client/profile-wizard/index.js
```diff
@@ -215,6 +221,32 @@ class ProfileWizard extends Component {
...
+			skipped: true,
+			step: this.getCurrentStep().key,
+		} )
+			.then( ( response ) => {
+				if ( response.status === 'success' ) {
+					recordEvent( 'storeprofiler_store_details_skip' );
+					const href = getAdminLink( 'admin.php?page=wc-admin' );
+					getHistory().push( href );
+				} else {
+					throw new Error();
```
mattsherman, line 236: Are there any details in `response` when it fails that we could log? I'm trying to think about how an HE or dev could troubleshoot things if this situation was encountered. What are the most likely reasons why this would fail?

client/profile-wizard/index.js
```diff
@@ -215,6 +221,32 @@ class ProfileWizard extends Component {
...
+				}
+			} )
+			.catch( () => {
+				createNotice(
+					'error',
+					__(
+						'There was a problem skipping your setup wizard.',
```
mattsherman, line 243:
```suggestion
						'There was a problem skipping the setup wizard.',
```

mattsherman (changes requested):
Thanks for adding error handling. Left a few comment about the error handling.

src/API/OnboardingProfile.php
```diff
@@ -209,6 +209,20 @@ public static function get_profile_properties() {
...
 				'validate_callback' => 'rest_validate_request_arg',
 			),
+			'skipped'             => array(
+				'type'              => 'boolean',
+				'description'       => __( 'Whether or not the profile was skipped.', 'woocommerce-admin' ),
+				'context'           => array( 'view' ),
+				'readonly'          => true,
+				'validate_callback' => 'rest_validate_request_arg',
+			),
+			'step'                => array(
```
joshuatf, line 219: Is the "skip" option only available on the "Store Details" step?  If so, I don't think we need to save the "step" data.

client/dashboard/index.js
```diff
@@ -26,23 +27,25 @@ const CustomizableDashboard = lazy( () =>
...
 		) {
-			return (
-				<Suspense fallback={ <Spinner /> }>
-					<ProfileWizard query={ query } />
-				</Suspense>
+			const lastStep = profilerStep ? `&step=${ profilerStep }` : '';
+			const href = getAdminLink(
+				`admin.php?page=wc-admin&path=/profiler${ lastStep }`
 			);
+			getHistory().push( href );
```
joshuatf, line 48: I think this should actually be:

```suggestion
			getHistory().push( getNewPath( {}, '/profiler', {} ) );
```

Otherwise the URL becomes `https://wp.local/wp-admin/https://wp.local/wp-admin/admin.php?page=wc-admin&path=/profiler&step=store-details` in my local install.

client/profile-wizard/index.js
```diff
@@ -215,6 +221,32 @@ class ProfileWizard extends Component {
...
+		const { createNotice, updateProfileItems } = this.props;
+		updateProfileItems( {
+			skipped: true,
+			step: this.getCurrentStep().key,
+		} )
+			.then( ( response ) => {
+				if ( response.status === 'success' ) {
+					recordEvent( 'storeprofiler_store_details_skip' );
+					const href = getAdminLink( 'admin.php?page=wc-admin' );
+					getHistory().push( href );
```
joshuatf, line 234: This one also needs to be updated:

```suggestion
					getHistory().push( getNewPath( {}, '/', {} ) );
```

client/homescreen/index.js
```diff
@@ -2,37 +2,38 @@
...
+		completed: profilerCompleted,
+		skipped: profilerSkipped,
+		step: profilerStep,
+	} = profileItems;
+	if ( isOnboardingEnabled() && ! profilerCompleted && ! profilerSkipped ) {
+		const lastStep = profilerStep ? `&step=${ profilerStep }` : '';
+		const href = getAdminLink(
+			`admin.php?page=wc-admin&path=/profiler${ lastStep }`
 		);
+		getHistory().push( href );
```
joshuatf, line 36: This one should be updated as well.

joshuatf (commented):
Left a few more clean-up comments I think need to be addressed before we can merge.

Also, want to note here that this breaks the profiler completion since we need to redirect to the dashboard after the profiler finishes, but as long as we get https://github.com/woocommerce/woocommerce-admin/pull/4717 merged in, this will be fixed.

client/profile-wizard/index.js
octaedro: Hi, thank you for your suggestions. I removed the `throw` in the commit bec1b23

client/homescreen/index.js
octaedro: Thank you @joshuatf, I did the updates in the commit 0a560f2

src/API/OnboardingProfile.php
octaedro: I'm not sure if I understood correctly, I'm using the `step` to take the user back to the OBW step where they left off. If you think that this is not necessary just let me know and I'll change it 🙂

joshuatf: I think we won't need it since the "Skip" button is only added to the "Store Details" step in https://github.com/woocommerce/woocommerce-admin/pull/4771/files

We can probably safely remove any reference to this saved step and "`lastStep` since it will always be the first step if they skip.

client/profile-wizard/index.js
```diff
@@ -223,6 +232,29 @@ class ProfileWizard extends Component {
...
+		const { createNotice, updateProfileItems } = this.props;
+		updateProfileItems( {
+			skipped: true,
+			step: this.getCurrentStep().key,
+		} )
+			.then( ( response ) => {
+				if ( response.status === 'success' ) {
+					recordEvent( 'storeprofiler_store_details_skip' );
+					getHistory().push( getNewPath( {}, '/', {} ) );
+				}
```
joshuatf, line 245: We won't need to check `response.status` here since this is already checked in the data actions.  The `.then()` block will only be triggered if the response was successful:

```suggestion
				recordEvent( 'storeprofiler_store_details_skip' );
				getHistory().push( getNewPath( {}, '/', {} ) );
```

client/profile-wizard/index.js
```diff
@@ -87,6 +91,11 @@ class ProfileWizard extends Component {
 			step: this.getCurrentStep().key,
 		} );
 
+		updateProfileItems( {
```
joshuatf, line 94: I think this call can be removed.  Once they update `skipped` to `true` I don't think we need to revert it back to `false`.

It can remain `true` and they either complete the profiler or not.  Leaving `skipped` as is provides us more information.

joshuatf (commented):
Clarified one of my previous suggestions and left a couple minor clean-up comments.  I think the test branch `add/4568_test` may also need a rebase.

client/profile-wizard/index.js
octaedro: Sounds good to me. I removed it in the commit 4b73e58

client/profile-wizard/index.js
octaedro: > We can probably safely remove any reference to this saved step and "lastStep since it will always be the first step if they skip.

Now I see, thank you @joshuatf, I removed it in the commit 37cab6

> We won't need to check response.status here since this is already checked in the data actions. The .then() block will only be triggered if the response was successful:

It seems like there is something tricky going on here, because with some types of errors the response will be `undefined` but in a "success" situation. So the user would be redirected to the home screen without receiving an alert.
The `catch` is triggered because the `response` doesn't have a `status`.
I created the branch `add/4568_test_2` with [the line commented](https://github.com/woocommerce/woocommerce-admin/pull/4721/files#diff-713853358fa08698c1bf0f8dfa42ca74R234) and adding a `die()` in the [update_items method](https://github.com/woocommerce/woocommerce-admin/blob/add/4568_test_2/src/API/OnboardingProfile.php#L129) if you want to test it.

joshuatf: Ah, I see.  Good catch.  So after this line:
https://github.com/woocommerce/woocommerce-admin/blob/bd0d7dce14ebc3809f1d1bacd0155c07dc6345db/packages/data/src/onboarding/actions.js#L56

We're also missing: `throw new Error();` which will allow us to catch errors after `updateProfileItems()`.  Do you mind adding that here?  It seems to get things working for me with your demo branch, but  we should definitely check and make sure it doesn't break other areas of the onboarding.

Screen Shot 2020-07-15 at 6 39 42 PM

src/API/OnboardingProfile.php
joshuatf: Just a heads up (not sure if you're still working on this) but this can be removed still.

client/profile-wizard/index.js
octaedro: Got it! I added the code in the commit 3705935
I did a rebase and updated the test branch (`add/4568_test`). Now I'm testing and it looks like everything is working correctly now.
Could you take another look at this PR?

src/API/OnboardingProfile.php
joshuatf: This property still needs to be removed.

tests/api/onboarding-profile.php
```diff
@@ -99,8 +99,10 @@ public function test_schema() {
 		$data       = $response->get_data();
 		$properties = $data['schema']['properties'];
 
-		$this->assertCount( 15, $properties );
+		$this->assertCount( 17, $properties );
```
joshuatf, line 102: These tests will need to be updated after `skip` is removed.

joshuatf (approved):
Thanks for all the effort on this, Fernando!  I know there was a lot of discussion and revision on this one, but this is a great improvement to the profiler.

There are a couple more issues around error handling and revisiting the profiler that we need to address, but I'll open some issues so we can tackle in a follow up.

Pre-approving pending the removal of the saved `step` property and test updates and then LGTM! 💯

tests/api/onboarding-profile.php
octaedro: Nice catch @joshuatf! I removed the property and corrected the tests in the commit 1511792

PR #4737, opened 2020-07-02
Trigger a remote inbox notifications engine run when the OBW completes

becdetat:
This is required for https://github.com/Automattic/wc-calypso-bridge/issues/559

A requirement of the admin note to be created is that it is created when the store is created. This may as well be when the OBW completes. So this triggers an engine run when the OBW completes.

1. Add a test data source to `DataSourcePoller::DATA_SOURCES`: `https://gist.githubusercontent.com/becdetat/f17e9617e5ca1211d0579cf079862905/raw/1adef708c9bd16f928cde0e17383446c537f3bd9/rinds-specs.json`
2. `DELETE FROM wp_wc_admin_notes` - this sets an initial state with the notes
3. Start the OBW
4. Check that the `wp_wc_admin_notes` table doesn't contain any pending notes
5. Complete the OBW
6. Check that the `wp_wc_admin_notes` table now contains pending and unactioned notes with a source of `'woocommerce.com'`

### DIFF
--- a/src/RemoteInboxNotifications/RemoteInboxNotificationsEngine.php
+++ b/src/RemoteInboxNotifications/RemoteInboxNotificationsEngine.php
@@ -48,7 +48,7 @@ class RemoteInboxNotificationsEngine {
 	public static function update_profile_option( $old_value, $new_value ) {
 		// Return early if we're not completing the profiler.
 		if (
-			( isset( $old_value['complete'] ) && $old_value['completed'] ) ||
+			( isset( $old_value['completed'] ) && $old_value['completed'] ) ||
 			! isset( $new_value['completed'] ) ||
 			! $new_value['completed']
 		) {
--- a/src/RemoteInboxNotifications/RemoteInboxNotificationsEngine.php
+++ b/src/RemoteInboxNotifications/RemoteInboxNotificationsEngine.php
@@ -11,6 +11,7 @@ defined( 'ABSPATH' ) || exit;
 
 use \Automattic\WooCommerce\Admin\PluginsProvider\PluginsProvider;
 use \Automattic\WooCommerce\Admin\Loader;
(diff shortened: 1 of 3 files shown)


src/RemoteInboxNotifications/RemoteInboxNotificationsEngine.php
```diff
@@ -27,6 +28,34 @@ class RemoteInboxNotificationsEngine {
...
+	 * This is triggered when the profile option is updated and if the
+	 * profiler is being completed, triggers a run of the engine.
+	 *
+	 * @param mixed $old_value Old value.
+	 * @param mixed $new_value New value.
+	 */
+	public static function update_profile_option( $old_value, $new_value ) {
+		// Return early if we're not completing the profiler.
+		if (
+			( isset( $old_value['complete'] ) && $old_value['completed'] ) ||
```
psealock, line 51: Is this a typo? `complete` v. `completed` in `isset( $old_value['complete'] )`.

psealock (approved):
This is looking good as far as I can tell. There may have been an extra engine run in there due to a typo, but otherwise looking good.

Here are my DB results after completing the OBW:

Screen Shot 2020-07-07 at 8 17 46 PM

src/RemoteInboxNotifications/RemoteInboxNotificationsEngine.php
becdetat: Oh nice. I copied this from `WC_Admin_Notes_Onboarding_Profiler` so I'll fix that while I'm fixing this.

becdetat: Fixed in 31dbcd3

psealock: LGTM 🚢

PR #4747, opened 2020-07-02
Build: Don't Attempt to Load non-minified Assets in Core Build

### LINKED ISSUE #4738: Broken when SCRIPT_DEBUG is true due to lack of non-minified assets.
**Describe the bug**
When running WooCommerce 4.3.0-rc.2, which includes woocommerce-admin 1.3.0-rc.2 there are multiple errors on my test install due to missing non-minified versions of Javascript assets.

E.g.
With `SCRIPT_DEBUG` set to true, the following JS path is attempted to be loaded:
`/wp-content/plugins/woocommerce/packages/woocommerce-admin/dist/chunks/42.11cc196c5f0cb3a17439.js`

This fails as the non-minified JS was not included in the 1.3.0-rc.2 release (it was in 1.3.0-rc.1). 

Setting `SCRIPT_DEBUG` set to false (or not defined), the following JS path is attempted to be loaded which then works correctly:
`/wp-content/plugins/woocommerce/packages/woocommerce-admin/dist/chunks/42.11cc196c5f0cb3a17439.min.js`

**To Reproduce**

1. Install and set up WooCommerce 4.3.0-rc.2
2. Set `define('SCRIPT_DEBUG', true);` in your `wp-config.php`
3. Navigate to "WooCommerce" in the admin area of the WordPress install - you'll see a bank page instead of the WooCommerce home screen.

timmyc:
Fixes #4738

With the changes shipped in 1.3.0 for not including un-minified files in the production/core build, users could encounter a state where wc-admin attempts to load non-minified JS files when they are not present in the build. The existing logic that determines if `.min` should be included in the file name was only looking to the value of `SCRIPT_DEBUG` was set and truthy. This would result in no wc-admin JS assets being loaded on sites running core builds with `define( 'SCRIPT_DEBUG', 'true' );`

The proposed fix in this PR involves adding a new config feature for `unminified-js`, which is `true` in development and plugin envs, but `false` for core. I had initially considered using the existing [`is_dev`](https://github.com/woocommerce/woocommerce-admin/blob/main/src/Loader.php#L116) method in the `Loader` for this, but that would prevent non-minified files from being loaded on the plugin version via `SCRIPT_DEBUG`.

I also opted to create a new helper method here to determine if `.min` should be used or not called `should_use_minified_js_file` - I primarily did this to allow for writing some unit tests that are independent of the value of `SCRIPT_DEBUG` in the testing environment. I feel it is a pretty sound approach, but am totally open to other ideas/approaches here.

This PR is probably best to test on a Jurassic Ninja site so you can use the beta tester plugin and easily install 4.3.0-rc.2. So to setup a test site, and re-create the bug:

- Launch a new JN site, tick the boxes to install woocommerce and woocommerce-beta-tester.
- Once the site launches, SSH into the instance and add `define( 'SCRIPT_DEBUG', true );` to your `wp-config.php`
- Next head to wp-admin/plugins.php and use the beta tester plugin to upgrade to 4.3.0-rc.2
- Once the latest version is installed and active, launch the Setup Wizard, and with the network tab open, you should see something like this ( note the missing `.min` from the JS chunk files ):

4 3-rc 2-SCRIPT_DEBUGtrue

_Back on your local computer_

- Check out this branch of wc-admin
- Clear out your local `dist` folder: `rm -rf dist`
- Run a **core** build of wc-admin: `WC_ADMIN_PHASE=core npm run build`
- Then the next few steps are just borrowed from the test:zip job:  `composer install --no-dev`
- `rm woocommerce-admin.zip` might not be there, but just to be safe
- and finally `./bin/make-zip.sh woocommerce-admin.zip`

__Back on JN__

- Upload and activate the .zip file you just built
- Reload the onboarding wizard, and it should now work, and you should see `.min` files in the network tab for js even though SCRIPT_DEBUG is on.

4 3-rc 2-with-fix-applied

__Bonus Points__

- You can run the tests locally, or let CI do it for you: `composer test -- --filter="WC_Admin_Tests_Loader"`

### Changelog Note:

Dev: Only load non-minified js files when SCRIPT_DEBUG is on, and env is development or plugin

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -193,6 +193,22 @@ class Loader {
 	}
 
 	/**
+	 * Determines if a minified JS file should be served.
+	 *
+	 * @param  boolean $script_debug Only serve unminified files if script debug is on.
+	 * @return boolean If js asset should use minified version.
+	 */
+	public static function should_use_minified_js_file( $script_debug ) {
+		// un-minified files are only shipped in non-core versions of wc-admin, return true if unminified files are not available.
+		if ( ! self::is_feature_enabled( 'unminified-js' ) ) {
+			return true;
+		}
+
+		// Otherwise we will serve un-minified files if SCRIPT_DEBUG is on, or if anything truthy is passed in-lieu of SCRIPT_DEBUG.
+		return ! $script_debug;
+	}
+
+	/**
 	 * Gets the URL to an asset file.
 	 *
 	 * @param  string $file File name (without extension).
@@ -204,7 +220,8 @@ class Loader {
 
 		// Potentially enqueue minified JavaScript.
 		if ( 'js' === $ext ) {
-			$suffix = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '' : '.min';
+			$script_debug = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG;
+			$suffix = self::should_use_minified_js_file( $script_debug ) ? '.min' : '';
 		}
 
 		return plugins_url( self::get_path( $ext ) . $file . $suffix . '.' . $ext, WC_ADMIN_PLUGIN_FILE );
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -60,7 +60,7 @@ class WC_Admin_Unit_Tests_Bootstrap {
 		tests_add_filter( 'setup_theme', array( $this, 'install_wc' ) );
 
 		// Set up WC-Admin config.
-		tests_add_filter( 'wc_admin_get_feature_config', array( $this, 'add_development_features' ) );
+		tests_add_filter( 'woocommerce_admin_get_feature_config', array( $this, 'add_development_features' ) );
 
 		// load the WP testing environment.
 		require_once $this->wp_tests_dir . '/includes/bootstrap.php';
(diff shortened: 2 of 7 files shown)


src/Loader.php
```diff
@@ -192,6 +192,22 @@ public static function is_onboarding_enabled() {
...
 
+	/**
+	 * Determines if a minified JS file should be served.
+	 *
+	 * @param  boolean $script_debug Only serve unminified files if script debug is on.
+	 * @return boolean If js asset should use minified version.
+	 */
+	public static function should_use_minified_js_file( $script_debug ) {
+		// un-minified files are only shipped in non-core versions of wc-admin, return true if unminified files are not available.
+		if ( ! self::is_feature_enabled( 'unminified-js' ) ) {
```
timmyc, line 203: Theoretically someone could still make this break by filtering `woocommerce_admin_features` to include `unminified-js` on a site running `core` - but I'd say that is super edge case / crazy pills.

becdetat (approved):
As soon as I installed the woocommerce-admin plugin (uploading the zip) I received error messages about `actioned_text` not being in the database. I would say that is just a DB version issue though, as on a page reload that error went away. Other than that everything works as specified ;-)

jeffstieler:
Normally I'd approve this.. but I've added commits. I think it's reasonable that @timmyc could approve now though.

tests/bootstrap.php
```diff
@@ -60,7 +60,7 @@ public function __construct() {
 		tests_add_filter( 'setup_theme', array( $this, 'install_wc' ) );
 
 		// Set up WC-Admin config.
-		tests_add_filter( 'wc_admin_get_feature_config', array( $this, 'add_development_features' ) );
+		tests_add_filter( 'woocommerce_admin_get_feature_config', array( $this, 'add_development_features' ) );
```
timmyc, line 63: Great catch, and interesting to see how this works in the test env ( pulling from the dev config that is ). Answered one of my questions from when I was working on this PR.

jeffstieler:
Approving on behalf of @timmyc, per Slack convo.

jeffstieler (approved):

timmyc:
Thanks for the test/review @becdetat 

> I would say that is just a DB version issue though, as on a page reload that error went away. Other than that everything works as specified ;-)

Yes sorry I forgot to mention that in the testing instructions that you will hit a db version conflict when following the instructions.

timmyc:
Cherry picked to `release/1.3.0-rc.2` in 3832bf35b963b36904b9fc6908559f155f94a7f1

PR #4752, opened 2020-07-06
Only load the performance indicators API endpoints if called from WP Admin.

### LINKED ISSUE #4751: Jetpack requests causing creation of lots of `post_by_email_address` options, can crash site.
**Describe the bug**
Further info about the effects: p6JqRr-2UZ-p2

when registering REST API endpoints in WC Admin, the new Jetpack request is triggered on each pageload where REST API gets initialized, which is always. This creates a new option for each logged in user. 

\Automattic\WooCommerce\Admin\API\Init::rest_api_init
\Automattic\WooCommerce\Admin\API\Reports\PerformanceIndicators\Controller::register_routes
\Automattic\WooCommerce\Admin\API\Reports\PerformanceIndicators\Controller::get_collection_params
\Automattic\WooCommerce\Admin\API\Reports\PerformanceIndicators\Controller::get_indicator_data
\Automattic\WooCommerce\Admin\API\Reports\PerformanceIndicators\Controller::get_jetpack_modules_data does request to `GET` `/jetpack/v4/module/all`, which calls 
Jetpack_Core_API_Module_List_Endpoint->process
Jetpack_Core_API_Module_List_Endpoint->get_modules
Jetpack_Core_Json_Api_Endpoints::prepare_options_for_response
Jetpack_Core_Json_Api_Endpoints::get_remote_value('post-by-email', ...)
update_option('post_by_email_addressUSER_ID', 'NULL')

**Expected behavior**
As a strategic objective, I think we should strive for the REST API endpoints to be loaded only when needed, i.e. only when in admin (unless we want WC Admin to become also framework/API for front end, but I thought that's mostly domain for Blocks package).

When I tried to wrap `new API\Init();` in `is_admin()`, I found out that e.g. some Blocks on the front end need data store for notes, so some more work is needed to decouple things before we can remove things.

Additionally, most of the background requests in WC admin are running in context where `is_admin()` is `false` and `wp_doing_ajax()` is `false` as well, so it's not a good test for when to call the API init.

As a tactical fix for now, we can just drop doing the Jetpack requests when not in admin.

WC Admin v: 1.3-rc.2
WC version: 4.3-rc.2

peterfabian:
First stab at fixing #4751.

The work to register those endpoints should not be necessary on the front end and contexts others than WP Admin.
This is similar to other endpoints, but that work requires more thorough checks and decoupling some items (like data stores) from the API initialisation.

Fixes #4751.

Please see the issue. TL;DR is that I think it's not necessary to load all the API endpoints on screens outside WP Admin area, unless some of the endpoints provide infrastructure to pages on the front end/in other contexts (but it seems to me most of the stats endpoints return 401 for guests and customers on front end, so switching it to 404 should not be a big deal). Also, this should prevent the problem with the performance indicators registration function call to Jetpack and creating too many wp_options.

1. Install WC & Jetpack. 
2. Apply changes
3. Test that performance indicators are still working in the admin area:
 - WooCommerce > Home (if you have the home screen feature enabled), and
 - Analytics > Overview

### Changelog Note:

> Fix: Register performance indicators API endpoints only when called from WordPress admin/backend area.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23625,7 +23625,7 @@
 			"dev": true
 		},
 		"prettier": {
-			"version": "1.19.1",
+			"version": "npm:prettier@1.19.1",
 			"resolved": "https://registry.npmjs.org/wp-prettier/-/wp-prettier-1.19.1.tgz",
 			"integrity": "sha512-mqAC2r1NDmRjG+z3KCJ/i61tycKlmADIjxnDhQab+KBxSAGbF/W7/zwB2guy/ypIeKrrftNsIYkNZZQKf3vJcg==",
 			"dev": true
--- a/package-lock.json
+++ b/package-lock.json
@@ -23625,7 +23625,7 @@
 			"dev": true
 		},
 		"prettier": {
-			"version": "npm:prettier@1.19.1",
+			"version": "1.19.1",
 			"resolved": "https://registry.npmjs.org/wp-prettier/-/wp-prettier-1.19.1.tgz",
 			"integrity": "sha512-mqAC2r1NDmRjG+z3KCJ/i61tycKlmADIjxnDhQab+KBxSAGbF/W7/zwB2guy/ypIeKrrftNsIYkNZZQKf3vJcg==",
 			"dev": true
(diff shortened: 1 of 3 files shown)


peterfabian:
Closing in favor of a different solution, #4755.

PR #4755, opened 2020-07-06
Avoid Jetpack REST API when retrieving active modules 

### LINKED ISSUE #4751: Jetpack requests causing creation of lots of `post_by_email_address` options, can crash site.

jeffstieler:
Fixes #4751. Alternative to https://github.com/woocommerce/woocommerce-admin/pull/4752.

This PR seeks to avoid using the Jetpack REST API to retrieve active modules instead of selectively loading the Performance Indicators endpoint controller (as https://github.com/woocommerce/woocommerce-admin/pull/4752 does).

- Install WC & Jetpack.
- Apply changes
- Test that performance indicators are still working in the admin area:
- WooCommerce > Home (if you have the home screen feature enabled), and
- Analytics > Overview
- Disable Jetpack
- Test that performance indicators are still working in the admin area (but missing "views", "visitors"):
- WooCommerce > Home (if you have the home screen feature enabled), and
- Analytics > Overview

### Changelog Note:

Fix: unnecessary REST API usage leading to potential overflow of Jetpack option sizes.

### DIFF
--- a/src/API/Reports/PerformanceIndicators/Controller.php
+++ b/src/API/Reports/PerformanceIndicators/Controller.php
@@ -43,6 +43,13 @@ class Controller extends \WC_REST_Reports_Controller {
 	protected $endpoints = array();
 
 	/**
+	 * Contains a list active Jetpack module slugs.
+	 *
+	 * @var array
+	 */
+	protected $active_jetpack_modules = null;
+
+	/**
 	 * Contains a list of allowed stats.
 	 *
 	 * @var array
@@ -178,16 +185,42 @@ class Controller extends \WC_REST_Reports_Controller {
 	}
 
 	/**
+	 * Get active Jetpack modules.
+	 *
+	 * @return array List of active Jetpack module slugs.
+	 */
+	public function get_active_jetpack_modules() {
+		if ( is_null( $this->active_jetpack_modules ) ) {
+			if ( class_exists( '\Jetpack' ) ) {
+				$this->active_jetpack_modules = \Jetpack::get_active_modules();
+			} else {
+				$this->active_jetpack_modules = array();
+			}
+		}
+
+		return $this->active_jetpack_modules;
+	}
+
+	/**
+	 * Set active Jetpack modules.
+	 *
+	 * @param array $modules List of active Jetpack module slugs.
+	 */
+	public function set_active_jetpack_modules( $modules ) {
+		$this->active_jetpack_modules = $modules;
+	}
+
+	/**
 	 * Get active Jetpack modules and endpoints.
 	 */
 	public function get_jetpack_modules_data() {
-		if ( ! class_exists( '\Jetpack_Core_Json_Api_Endpoints' ) ) {
+		$active_modules = $this->get_active_jetpack_modules();
+
+		if ( empty( $active_modules ) ) {
 			return;
 		}
 
-		$request  = new \WP_REST_Request( 'GET', '/jetpack/v4/module/all' );
-		$response = rest_do_request( $request );
-		$items    = apply_filters(
+		$items = apply_filters(
 			'woocommerce_rest_performance_indicators_jetpack_items',
 			array(
 				'stats/visitors' => array(
@@ -205,18 +238,8 @@ class Controller extends \WC_REST_Reports_Controller {
 			)
 		);
 
-		if ( is_wp_error( $response ) ) {
-			return $response;
-		}
-
-		if ( 200 !== $response->get_status() || empty( $items ) ) {
-			return;
-		}
-
-		$data = $response->get_data();
-
 		foreach ( $items as $item_key => $item ) {
-			if ( ! $data[ $item['module'] ] || ! $data[ $item['module'] ]['activated'] ) {
+			if ( ! in_array( $item['module'], $active_modules, true ) ) {
 				return;
 			}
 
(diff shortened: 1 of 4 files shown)


src/API/Reports/PerformanceIndicators/Controller.php
```diff
@@ -177,17 +184,43 @@ public function get_analytics_report_data() {
...
 	}
 
+	/**
+	 * Get active Jetpack modules.
+	 *
+	 * @return array List of active Jetpack module slugs.
+	 */
+	public function get_active_jetpack_modules() {
+		if ( is_null( $this->active_jetpack_modules ) ) {
+			if ( class_exists( '\Jetpack' ) ) {
```
timmyc, line 194: When perusing similar approaches via the wonders of opengrok :) I noticed quite a few places doing a `function_exists()` call on `active_jetpack_modules()` too. Not sure if this was due to the method being added at some point, but perhaps it would be safe to check here too just in case.

src/API/Reports/PerformanceIndicators/Controller.php
```diff
@@ -42,6 +42,13 @@ class Controller extends \WC_REST_Reports_Controller {
 	 */
 	protected $endpoints = array();
 
+	/**
+	 * Contains a list active Jetpack module slugs.
+	 *
+	 * @var array
+	 */
+	protected $active_jetpack_modules = null;
```
timmyc, line 50: Really like the caching of the modules, nice.

src/API/Reports/PerformanceIndicators/Controller.php
jeffstieler: Done in 4b1289ef.

src/API/Reports/PerformanceIndicators/Controller.php
```diff
@@ -177,17 +184,43 @@ public function get_analytics_report_data() {
...
 
+	/**
+	 * Get active Jetpack modules.
+	 *
+	 * @return array List of active Jetpack module slugs.
+	 */
+	public function get_active_jetpack_modules() {
+		if ( is_null( $this->active_jetpack_modules ) ) {
+			if ( class_exists( '\Jetpack' ) && method_exists( '\Jetpack', 'get_active_modules' ) ) {
+				$this->active_jetpack_modules = \Jetpack::get_active_modules();
```
peterfabian, line 195: I think this can be set to `null` here if someone filters `jetpack_active_modules` and returns null, so it would get called repeatedly, but that's super edge case, I suppose.

src/API/Reports/PerformanceIndicators/Controller.php
```diff
@@ -42,6 +42,13 @@ class Controller extends \WC_REST_Reports_Controller {
 	 */
 	protected $endpoints = array();
 
+	/**
+	 * Contains a list active Jetpack module slugs.
```
peterfabian, line 46: Small type: a list _of_ slugs?

src/API/Reports/PerformanceIndicators/Controller.php
peterfabian: Agreed, seems like it gets hit ±4 times per request, so caching makes sense 👍

peterfabian (approved):
Besides the nitpicks I found, it seems to work fine in my testing. 👍  
No more options are getting created and home page and analytics work, with and without Jetpack. Also, it seems like a better approach than hitting the REST API code. Nice touch with the added caching.

src/API/Reports/PerformanceIndicators/Controller.php
jeffstieler: Fixed in 26523280.

src/API/Reports/PerformanceIndicators/Controller.php
jeffstieler: Handled in 26523280.

timmyc:
cherry picked to `release/1.3.0-rc.2` in eb599dda11b8614961ca3c5baa16adea2f0a6533

PR #4759, opened 2020-07-07
Update @wordpress/base-styles and replace deprecated variables 

### LINKED ISSUE #4732: Update deprecated color vars
We're using a few [deprecated color variables](https://github.com/WordPress/gutenberg/blob/master/packages/base-styles/_colors.scss#L26-L50) in wc-admin. This issue illustrates which color vars to use in place of those that have been deprecated.

This list is based on what I was able to find actually being used in the code.

* `$light-gray-700` → ~`$light-gray-secondary`~ `$gray-400`
* `$light-gray-500` → ~`$light-gray-secondary`~ `$gray-400`
* `$light-gray-200` → ~`$light-gray-tertiary`~ `$gray-100`
* `$light-gray-100` → ~`$light-gray-tertiary`~ `$gray-100`
* `$dark-gray-900` → ~`$dark-gray-primary`~ `$gray-900`
* `$dark-gray-500` → ~`$medium-gray-text`~ `$gray-700`
* `$dark-gray-300` → ~`$medium-gray-text`~ `$gray-700`
* `$theme-color` → `var(--wp-admin-theme-color)`

samueljseay:
Fixes #4732

This updates `@wordpress/base-styles`, and because `@wordpress/components` depends on variables from it, that must be updated as well.

There are 3 major changes required as part of updating:

1. The $theme-color variable is no longer exposed by base-styles. Instead there are 3 css vars exposed. These are made available by use of a provided mixin. Situations where $theme-color was darkened using scss have been mapped as best as possible to the 2 other darkened shades of the css var that are available such as `--wp-admin-theme-color-darker-20` and `--wp-admin-theme-color-darker-10`. In some cases this means the colors are not exactly the same as before.

2. The post css `theme()` call is no longer available. All uses of this have been consolidated to use of the main theme css var `--wp-admin-theme-color`. This means that calls like `theme(secondary)` or `theme(outline)` etc have all been consolidated to the one color.

3. Many of the variables used for different shades of gray have been deprecated. These have been mapped across to the new gray variables. (Mapped according to the list described in #4732)

### Accessibility

### Screenshots

One notable difference is the hover color of calendar days on the date picker. Due to there not being any use of `secondary` any more the color change is more subtle. Gif below:

cap-style

See the testing instructions from #4558 

Theme color settings should match those in the user preferences, but now just shades of the primary color from the theme.

### Additional note

I've reached out to Jay to request that he take a look at the approach taken with consolidating colors here. But other reviews are welcome!

### DIFF
--- a/client/header/activity-panel/panels/inbox/style.scss
+++ b/client/header/activity-panel/panels/inbox/style.scss
@@ -1,7 +1,8 @@
 .woocommerce-inbox-message {
 	position: relative;
 	color: $gray-text;
-	background: $gray-200;
+	// background: $gray-100;
+	background: $light-gray-placeholder;
 	border-radius: 2px;
 	@include font-size( 13 );
 	margin: 0 0 $gap-large;
@@ -115,7 +116,7 @@
 
 .woocommerce-inbox-message__actions {
 	padding-top: $gap;
-	border-top: 1px solid $gray-100;
+	border-top: 1px solid $gray-200;
 
 	// Ensures any immediate child with a sibling has space between the items
 	& > * + * {
--- a/client/header/activity-panel/panels/inbox/style.scss
+++ b/client/header/activity-panel/panels/inbox/style.scss
@@ -1,8 +1,7 @@
 .woocommerce-inbox-message {
 	position: relative;
 	color: $gray-text;
-	// background: $gray-100;
-	background: $gray-50;
+	background: $gray-100;
 	border-radius: 2px;
 	@include font-size( 13 );
 	margin: 0 0 $gap-large;
@@ -30,7 +29,7 @@
 	}
 
 	&:not(.is-placeholder) {
-		border: 1px solid $gray-400;
+		border: 1px solid $gray-200;
 	}
 
 	&.message-is-unread {
--- a/client/header/activity-panel/panels/inbox/style.scss
+++ b/client/header/activity-panel/panels/inbox/style.scss
@@ -76,7 +76,7 @@
 	}
 
 	.woocommerce-inbox-message__date {
-		color: $medium-gray-text;
+		color: $gray-700;
 		@include font-size( 12 );
 		margin-bottom: $gap;
 		font-style: normal;
@@ -92,7 +92,7 @@
 }
 
 .woocommerce-inbox-message__text {
-	color: $medium-gray-text;
+	color: $gray-700;
 	font-style: normal;
 	font-weight: normal;
 	@include font-size( 14 );
@@ -115,7 +115,7 @@
 
 .woocommerce-inbox-message__actions {
 	padding-top: $gap;
-	border-top: 1px solid $light-gray-tertiary;
+	border-top: 1px solid $gray-100;
 
 	// Ensures any immediate child with a sibling has space between the items
 	& > * + * {
@@ -156,7 +156,7 @@
 .woocommerce-inbox-dismiss-confirmation_wrapper {
 	p {
 		font-size: 16px;
-		color: $medium-gray-text;
+		color: $gray-700;
 	}
 	.woocommerce-inbox-dismiss-confirmation_buttons {
 		text-align: right;
--- a/client/header/activity-panel/panels/inbox/style.scss
+++ b/client/header/activity-panel/panels/inbox/style.scss
@@ -55,7 +55,7 @@
 
 .woocommerce-inbox-message__content {
 	.woocommerce-inbox-message__title {
-		color: $dark-gray-primary;
+		color: $gray-900;
 		@include font-size( 16 );
 		font-style: normal;
 		line-height: 1.5;
--- a/client/header/activity-panel/panels/inbox/style.scss
+++ b/client/header/activity-panel/panels/inbox/style.scss
@@ -1,5 +1,3 @@
-
-
 .woocommerce-inbox-message {
 	position: relative;
 	color: $gray-text;
@@ -31,7 +29,7 @@
 	}
 
 	&:not(.is-placeholder) {
-		border: 1px solid $light-gray-secondary;
+		border: 1px solid $gray-400;
 	}
 
 	&.message-is-unread {
@@ -208,7 +206,7 @@
 }
 
 // Tweak to fix dropdown and placeholder in IE 11
-@media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {
+@media all and ( -ms-high-contrast: none ), ( -ms-high-contrast: active ) {
 	.woocommerce-admin-dismiss-dropdown {
 		margin-top: 0;
 	}
--- a/client/header/activity-panel/panels/inbox/style.scss
+++ b/client/header/activity-panel/panels/inbox/style.scss
@@ -1,7 +1,7 @@
 .woocommerce-inbox-message {
 	position: relative;
 	color: $gray-text;
-	background: $core-grey-light-200;
+	background: $gray-200;
 	border-radius: 2px;
 	@include font-size( 13 );
 	margin: 0 0 $gap-large;
--- a/client/header/activity-panel/panels/inbox/style.scss
+++ b/client/header/activity-panel/panels/inbox/style.scss
@@ -2,7 +2,7 @@
 	position: relative;
 	color: $gray-text;
 	// background: $gray-100;
-	background: $light-gray-placeholder;
+	background: $gray-50;
 	border-radius: 2px;
 	@include font-size( 13 );
 	margin: 0 0 $gap-large;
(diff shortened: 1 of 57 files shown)


client/profile-wizard/style.scss
```diff
@@ -312,7 +310,7 @@
 		box-shadow: $shadow-popover;
 
 		&.is-active {
-			border-color: $blue-medium-focus;
+			border-color: var(--wp-admin-theme-color);
```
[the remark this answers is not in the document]
jameskoster: If I said that, I now retract the statement :D The OBW should match the theme, and by extension the rest of our UI :)

client/stylesheets/abstracts/_variables.scss
```diff
@@ -31,24 +33,24 @@ $spacing: 16px;
...
-$dark-gray-300: $core-grey-dark-300; // This & below have 4.5+ contrast against white
+$gray-700: $core-grey-dark-300; // This & below have 4.5+ contrast against white
 $dark-gray-400: $core-grey-dark-400;
-$dark-gray-500: $core-grey-dark-500;
+$gray-700: $core-grey-dark-500;
 $dark-gray-600: $core-grey-dark-600;
 $dark-gray-700: $core-grey-dark-700;
 $dark-gray-800: $core-grey-dark-800;
-$dark-gray-900: $core-grey-dark-900;
+$gray-900: $core-grey-dark-900;
```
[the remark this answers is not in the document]
jameskoster: +1. The only color var we should create is the WooCommerce purple, for the very few cases in which we need to use it.

jameskoster:
> @jameskoster, this changes the button colors away from Core WP's buttons. Is that the intention for Gutenberg?

The buttons should match core. FWIW they _are_ matching for me on this branch 🤔 

I noticed a few small details that are off though. I'm not sure if they're all 100% down to us, or should be addressed in this PR, but I'm going to list them here anyway :)

* [Hover state of the quick links](https://cloudup.com/cJ6FxGUTtPx)
* [The little purple 'active' square in reports](https://cloudup.com/ca53vrliXgC)
* [Button focus state in filter creation](https://cloudup.com/cl5KCwFVKEk)
* [Focus state on filter row](https://cloudup.com/cTgxYDOmL_5)
* [Focus state on selects in filtering](https://cloudup.com/cwcFvIqM5KD)

The purple / blue details in the screenshots above should be mapped to the theme color.

client/stylesheets/abstracts/_variables.scss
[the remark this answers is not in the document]
samueljseay: Thanks I wondered about that one. Most of this was just a find replace. I'll fix this.

samueljseay:
@jameskoster @psealock 

How can we match other colors from the theme now that base-styles only exposes 3 shades of just one color of the theme?

jameskoster:
For the buttons in particular we should be using the core component out of the box with no overrides, so the theme color should be applied automatically.

For our custom components, maybe we can look at core components for implementation inspiration there? I assume they tackled this issue already :D

samueljseay:
@psealock @jameskoster I have tried to be pragmatic about the refactoring of all the grays here, will need some feedback.

I found that if I tried to map everything directly to a gray that exists in the @wordpress/base-styles that it dramatically changes the look and feel of screens like analytics and the inbox, it makes the entire UI look quite dark and in many places worse if we try to map things as best as possible. I really struggled with this, because I couldn't actually make it look right and with no real baseline to go against I think this approach wasn't going to work.

**I have opted instead** to introduce one more gray variable for WooCommerce. The gray is called $gray-50 and is lighter than any grays in base-styles, but allows us to map most other grays in a better way, it is a blueish light gray shade that existed already in Woo and was used in various places. My thought is that we could look to refactor this gray away later but I think it would require a refactor of the design for those screens like Analytics that heavily rely on grays throughout.

2 relevant commits:

1. I adjusted some direct mappings of one gray to another that felt wrong and made the UI look "wrong": https://github.com/woocommerce/woocommerce-admin/pull/4759/commits/ffe9c16cdd3c0d9931a6bd7906739c15b83b3ee0
2. I introduced a new gray variable "$gray-50". I have used it sparingly to recreate the look and feel of existing screens: https://github.com/woocommerce/woocommerce-admin/pull/4759/commits/cdff94eb33a4edb60cc986667b1c03afcaee63ae

Also I have mapped existing grays:

* Replace dark-gray-primary with gray-900
* Replace medium-gray-text with gray-700
* Replace core-grey-dark-900 with gray-900
* Replace core-grey-light-100 with gray-100
* Replace core-grey-light-200 with gray-200
* Replace core-grey-light with gray-100
* Replace core-grey-light 400 & 500 with grey-100
* Replace core-grey-light-600 with gray-200
* Replace core-grey-light-900 with gray-400
* Replace core-grey-dark 150 with gray-600, remove core-grey-dark 100,200
* Replace core-grey-dark-300 with gray-700
* Remove core-grey-dark-600
* Replace core-grey-dark-400 with gray-700
* Replace core-grey-dark-500 with gray-700.
* Replace core-grey-dark-700 with gray-900.
* Replace core-grey-dark-800 with gray-900.
* Replace core-form-grey with gray-700

samueljseay:
> For the buttons in particular we should be using the core component out of the box with no overrides, so the theme color should be applied automatically.
> 
> For our custom components, maybe we can look at core components for implementation inspiration there? I assume they tackled this issue already :D

I DM'd you a question about this, but I'll also put it here:

Do you mean the Gutenberg button component in @wordpress/components? The reason I ask is, I had a look at the
code there and it indeed only deals with the theme css color in varying shades that is provided by base-styles.
So to me there are a couple of issues here, maybe you can clarify your thoughts on them:
1. Gutenberg is moving towards using just a single color from the theme throughout the base look and feel.
Older WordPress buttons don’t do this, they use multiple colors from the theme
2. I just don’t know how we can use multiple colors from the theme if we’re just picking up the only css vars we now have available, which are three shades of one theme color? In the ectoplasm it is just the purple that is exposed.

jameskoster:
Adding a TLDR of what I said on Slack:

We should align our colors with Gutenberg. In this case that means wc-admin buttons with the Ectoplasm theme active should be purple. This is different to wp-admin where they are green, but matches Gutenberg, so that is fine.

---

I noticed these small issues are still outstanding;

* [Hover state of the quick links](https://cloudup.com/cJ6FxGUTtPx)
* [The little purple 'active' square in reports](https://cloudup.com/ca53vrliXgC)
* [Button focus state in filter creation](https://cloudup.com/cl5KCwFVKEk)
* [Focus state on filter row](https://cloudup.com/cTgxYDOmL_5)
* [Focus state on selects in filtering](https://cloudup.com/cwcFvIqM5KD)

Wherever you see the Woo purple, or blue in the screenshots above, let's use the theme color var.

---

Let's try to avoid adding a new custom gray variable, it will only cause headaches down the road. If you can set everything currently using your gray-50 var to gray-100 I am happy to go through with a fine-tooth comb and make adjustments :)

jameskoster:
Opened a PR on this branch [here](https://github.com/woocommerce/woocommerce-admin/pull/4803), to remove the `gray-50` var and harmonise the home screen (and some analytics) colors.

client/task-list/style.scss
```diff
@@ -40,8 +40,8 @@
 		}
 	}
 
-	.woocommerce-task__estimated-time {
-		color: $core-grey-dark-300;
+	.woocommerce-task-estimated-time {
```
becdetat, line 43: This has broken things a little - replacing the `__` with a single dash:
image

client/header/activity-panel/panels/inbox/style.scss
```diff
@@ -1,9 +1,8 @@
-
-
 .woocommerce-inbox-message {
 	position: relative;
 	color: $gray-text;
-	background: $core-grey-light-200;
+	// background: $gray-100;
```
becdetat, line 4: commented out code

becdetat:
Everything looks good to me apart from a couple of small things after a quick smoke test and scanning the code changes.

client/task-list/style.scss
samueljseay: oof thank you for spotting this! will fix.

client/header/activity-panel/panels/inbox/style.scss
samueljseay: This looks a bit weird. some mistake in merging the other PR to this, I'm not sure, either way thanks for spotting, will fix.

samueljseay:
@jameskoster I've created a new issue: #4818  to track fixing any remaining styling problems, we can create a separate PR for that. I think everything is looking pretty good for a merge here now.

samueljseay:
@becdetat Thanks for your review, I've addressed both issues.

psealock:
Nice job on this one team and especially @samueljseay for seeing this one through

PR #4765, opened 2020-07-07
Limit inbox status to unactioned notes

becdetat:
`pending` status notes were showing in the inbox / on the WooCommerce home screen (`pending` is used by the remote inbox notifications system).

1. `delete from wp_wc_admin_notes` - this resets to an initial state
2. add `https://gist.githubusercontent.com/becdetat/f17e9617e5ca1211d0579cf079862905/raw/1adef708c9bd16f928cde0e17383446c537f3bd9/rinds-specs.json` to the `DATA_SOURCES` array in `src/RemoteInboxNotifications/DataSourcePoller.php`
3. run the `wc_admin_daily` cron task - this runs through the remote inbox notifications poller and engine
4. check that there are `pending` notes in the `wp_wc_admin_notes` table, but that these are not present in the inbox on the dashboard or other pages, and that the unread indicator is correct

### DIFF
--- a/client/header/activity-panel/panels/inbox/index.js
+++ b/client/header/activity-panel/panels/inbox/index.js
@@ -175,6 +175,7 @@ export default compose(
 		const inboxQuery = {
 			page: 1,
 			per_page: QUERY_DEFAULTS.pageSize,
+			status: 'unactioned',
 			type: QUERY_DEFAULTS.noteTypes,
 			orderby: 'date',
 			order: 'desc',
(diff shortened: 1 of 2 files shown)


client/header/activity-panel/panels/inbox/index.js
```diff
@@ -175,6 +175,7 @@ export default compose(
 		const inboxQuery = {
 			page: 1,
 			per_page: QUERY_DEFAULTS.pageSize,
+			status: 'unactioned',
```
mattsherman, line 178: What do you think about extracting these query params to a common function (in a followup PR) that both the inbox and the unread indicator can share? It seems like they always should be in-sync.

mattsherman (approved):
Nice test instructions! Works as expected.

client/header/activity-panel/panels/inbox/index.js
becdetat: Sure

PR #4766, opened 2020-07-07
Onboarding: Don't show tracking modal if site is already opted in

### LINKED ISSUE #574: Remove usage tracking auth message
On the eCommerce plan we're not required to display the authorization message, so the modal should be removed: https://cloudup.com/c1mhDuaiFe2

 #534 was supposed to cover this change, but the modal continues being displayed.

timmyc:
Fixes https://github.com/Automattic/wc-calypso-bridge/issues/574

This branch adds some logic to not show the usage tracking modal if a site has already opted into usage tracking. While this use-case might not happen often ( sites likely don't run through the onboarding wizard more than once ) - it will be immediately useful in the context of the eCommerce Plan where users have already opted into usage tracking.

### Screenshots

__Flow With Usage Not Opted In ( DEFAULT )__
This is pretty much unchanged...

usage-tracking-ff

__Flow With Usage already opted in (NEW)__
usge-opted-in

Two parts of testing fun on this one. For the first, verify the existing flow ( not opted in ) works as expected. To do so:

- Visit `/wp-admin/admin.php?page=wc-settings&tab=advanced&section=woocommerce_com`, and un-check the usage tracking checkbox there, and save settings.
- Open up the OBW via `/wp-admin/admin.php?page=wc-admin&reset_profiler=1` 
- Complete the first store details step, verify the usage modal is shown, and operates as expected.

Next up, revisit the same settings page, and opt-in/ turn on usage tracking.

- Visit `/wp-admin/admin.php?page=wc-admin&reset_profiler=1` again
- Verify after completing the step, the usage modal is not shown.

Worth noting, there is a react warning being shown on the store details step for me. This exists on `main` currently also:

Home ‹ WooCommerce ‹ bend outdoors — WooCommerce 2020-07-07 15-55-35

### Changelog Note:

Tweak: Only show usage tracking modal once in OBW.

### DIFF
--- a/client/profile-wizard/steps/usage-modal.js
+++ b/client/profile-wizard/steps/usage-modal.js
@@ -94,6 +94,14 @@ class UsageModal extends Component {
 	}
 
 	render() {
+		// Bail if site has already opted in to tracking
+		if ( this.props.allowTracking ) {
+			const { onClose, onContinue } = this.props;
+			onClose();
+			onContinue();
+			return null;
+		}
+
 		const { allowTracking } = this.state;
 		const { isRequesting } = this.props;
 		const trackingMessage = interpolateComponents( {


mattsherman:
I'll review this

mattsherman (approved):
@timmyc Thanks for the detailed testing instructions! 🙌 

Tested and received expected results. Works well!

Personal preference... I'd prefer for the logic to be such that the UsageModal isn't even instantiated in the case where it won't be shown at all. That always feels clearer to me -- having a modal that may not show when you "show" it always feels a little off to me. So, I'd favor doing the check in `StoreDetails` instead. (I'd also go so far as to not have any real logic in the modal, just display and collection of user input.)

But, I can see the nicety of keeping the logic contained in the modal itself. So, I think the implementation in this PR is reasonable too.

Approving this!

timmyc:
@mattsherman I had originally attempted to put it in the parent component so it would not be included, but since the usage modal is what dictates proceeding to the next step, it was turning into a much more involved task. Definitely could use a deeper refactor here.

PR #4771, opened 2020-07-08
Rework the store details onboarding screen

### LINKED ISSUE #4566: Store details step - add contextual information and skip link
**User story:**
As a new merchant
I want to have more context on why I should add info about my store and having a way to skip those steps
So I can decide if I should provide that info or skip it entirely 

**Description:** with the removal of the splash screen we need to adjust the header to provide a more friendly welcome, add a skip link beneath the card and informational tooltips to provide more context.

**Acceptance Criteria:**
- [x] Update the header content
- [x] Add a tooltip in the subtitle explaining why it's important to provide this info
- [x] Add a link to skip the Store Profiler
- [x] Add a tooltip in the skip link reminding why merchants should avoid skipping it

**Event tracking:**
Event name: wcadmin_storeprofiler_store_details_skip
Event prop: NA
Description: when the user clicks to skip the profiler

**Additional links:**
[Design mockup](https://www.figma.com/file/JH9XMFUCOjfXdr3N09AHRD/On-boarding-iterations-June-'20?node-id=171%3A2)

samueljseay:
Fixes #4566 **(and is dependent on the changes from #4759 )**

Changes:

* Use the new `Card` and `Flex` from `@wordpress/components`
* Add contextual tooltips
* Adjust existing styles to match [new designs](https://www.figma.com/file/JH9XMFUCOjfXdr3N09AHRD/On-boarding-iterations-June-'20)

### Accessibility

### Screenshots
Screenshot 2020-07-02 at 3 58 35 PM
Screenshot 2020-07-02 at 3 58 48 PM

- Start the setup wizard via the usual means (either after installing WooCommerce the first time or via `WooCommerce > Settings > Help > Setup Wizard`)
- Visually confirm that the page matches [the designs](https://www.figma.com/file/JH9XMFUCOjfXdr3N09AHRD/On-boarding-iterations-June-'20)
- Visually confirm that the tooltips work on hover and that the content is correct.


client/profile-wizard/style.scss
```diff
@@ -91,10 +99,32 @@
...
+			}
+		}
+
 		.woocommerce-card ~ p {
 			font-size: 12px;
 			color: $core-grey-dark-150;
 		}
+
+		.woocommerce-profile-wizard__footer {
+			margin: 34px auto 0 auto;
```
psealock, line 117: I think some bottom margin will be better. It looks very close to the edge of the screen.

Screen Shot 2020-07-09 at 1 21 17 PM

client/profile-wizard/style.scss
```diff
@@ -424,6 +454,18 @@
...
+.woocommerce-profile-wizard__submit {
+	margin-bottom: $gap-small;
+}
+
+.woocommerce-profile-wizard__tooltip-icon {
+	color: $studio-gray-60;
+	display: flex;
+	align-items: center;
+	margin-left: 8px;
+	cursor: pointer;
```
psealock, line 466: I see the pointer on the designs, but it feels wrong to me. Here is Github's:

Screen Shot 2020-07-09 at 1 22 36 PM

Maybe `cursor: help;` ?

client/profile-wizard/steps/store-details.js
```diff
@@ -167,67 +177,110 @@ class StoreDetails extends Component {
...
+						<Tooltip
+							text={ __(
+								'Manual setup is only recommended for\n experienced WooCommerce users or developers.',
+								'woocommerce-admin'
+							) }
+						>
+							<span className="woocommerce-profile-wizard__tooltip-icon">
+								<Icon icon="info-outline" size={ 16 } />
+							</span>
+						</Tooltip>
```
psealock, line 281: I'm wondering if the tooltip should be outside of the link. Especially with the `cursor: pointer;`, it looks like you're supposed to click the info icon to get more info, but that makes you skip the profiler.

psealock (commented):
This is looking great @samueljseay, I just brought up some design related questions.

Does this depend on https://github.com/woocommerce/woocommerce-admin/pull/4721 being merged first?

client/profile-wizard/steps/store-details.js
samueljseay: Yep good point. Will fix.

client/profile-wizard/style.scss
samueljseay: I agree the cursor feels wrong. Will fix.

client/profile-wizard/style.scss
samueljseay: @psealock Could you provide a fuller screenshot? The color of the link looks wrong, and it looks like your screenshot has the link inside the card which isn't right with the design, but maybe I'm just struggling with context because the screenshot is smaller.

samueljseay:
> This is looking great @samueljseay, I just brought up some design related questions.
> 
> Does this depend on #4721 being merged first?

Thats a good question, I didn't make this in the context of #4721 but it probably makes sense for it to get merged first

client/profile-wizard/style.scss
samueljseay: I see what you mean on a short screen, so I've added the same top and bottom margin to the footer to fix this. Not sure why the link looks wrong in your screenshot though 😄

samueljseay:
@psealock thanks for the feedback. I've incorporated your suggestions. I went with the `cursor: help` for the cursor issue.

psealock (approved):
Great! 🚢

psealock:
I went ahead and approved. Once #4721 is merged, there is a `skipProfiler` function you'll need to hook up to.

.eslintrc.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
 	extends: [
 		'plugin:@wordpress/eslint-plugin/recommended',
 		'prettier',
+		'prettier/react',
```
samueljseay, line 5: As part of some updates I had to fix prettier to make linting work properly. This extends is necessary to avoid eslint conflicts with prettier

.prettierrc
```diff
@@ -1,10 +1,11 @@
...
-singleQuote: true
-trailingComma: es5
-bracketSpacing: true
-parenSpacing: true
-jsxBracketSameLine: false
-semi: true
-arrowParens: always
+{
+	"useTabs": true,
+	"printWidth": 80,
```
samueljseay, line 3: In VSCode it no longer seems happy with yaml for the prettierrc. I've changed it to JSON, which fixes that.

samueljseay:
I've incorporated the `skipProfiler` functionality from #4721 to the "skip setup wizard link" in this commit: https://github.com/woocommerce/woocommerce-admin/pull/4771/commits/3a0c6eb348d755d513a98598538c1becf668c92d

This needs a re-review. Maybe you could take a look @psealock ?

client/profile-wizard/steps/store-details.js
```diff
@@ -163,76 +172,117 @@ class StoreDetails extends Component {
...
-				<Card>
-					<Form
-						initialValues={ this.initialValues }
-						onSubmitCallback={ this.onSubmit }
-						validate={ validateStoreAddress }
+					<Tooltip
+						text={ __(
+							'Your store address will help us configure currency\n options and shipping rules automatically.\n This information will not be publicly visible and can\n easily be changed later.',
+							'woocommerce-admin'
+						) }
```
jeffstieler, line 192: We need to supply `aria-label` as well since this component is invisible to screen readers. See the "accessibility note" here: https://github.com/WordPress/gutenberg/tree/master/packages/components/src/tooltip#tooltip

jeffstieler (dismissed):
This looks good visually in wide viewports, but has some glitches for mobile.

On devices < 660px there is a visual stutter with the bottom tooltip popover: 

https://cloudup.com/c59IntY6BNN

I'm unable to trigger the tooltips at all on iOS, or with navigation using the keyboard and/or screen reader.

I know the 700ms delay is part of the `Tooltip` component, but it would be cool if that was shorter. 🤷

client/profile-wizard/steps/store-details.js
samueljseay: Good point, thanks. Will fix.

samueljseay:
> On devices < 660px there is a visual stutter with the bottom tooltip popover:
> 
> https://cloudup.com/c59IntY6BNN
> 
> I'm unable to trigger the tooltips at all on iOS, or with navigation using the keyboard and/or screen reader.
> 
> I know the 700ms delay is part of the Tooltip component, but it would be cool if that was shorter. 🤷

@jeffstieler sounds like most of the issues you're seeing are inside the Tooltip component itself, but I agree they're not ideal anyway. I'll have a think about what I can do here. Maybe some style adjustments can help with the mobile stutter.

timmyc:
Good find on the mobile issues @jeffstieler 

[Some good chatter on this topic in slack too](https://a8c.slack.com/archives/CBKK5KM41/p1595288018230500)

Since we are kind of at the cut-off for beta - I'm proposing we land this as-is for now, and then get some feedback from @jameskoster about how to handle the narrow viewports since Tooltip is a bit buggy there. One option would be to simply hide the tooltips on smaller viewports completely, or perhaps have another element with the same verbiage, that is then displayed instead on the smaller viewport.

psealock (approved):
New commits look good and the functionality is working as expected. 🚢

PR #4773, opened 2020-07-08
Add is_ecommerce remote inbox notifications rule

becdetat:
Partially addresses https://github.com/Automattic/wc-calypso-bridge/issues/559

This adds a rule that passes (or fails, depending on the rule value) if the site is an eCommerce site.

### Screenshots
image

1. Clone [the Calypso bridge](https://github.com/Automattic/wc-calypso-bridge) into your `plugins` directory
2. Activate the plugin
3. Add `https://gist.githubusercontent.com/becdetat/f17e9617e5ca1211d0579cf079862905/raw/fc16ebaf98ac8dae8075b5be6145827892be5502/rinds-specs.json` to the `DATA_SOURCES` array in `src/RemoteInboxNotifications/DataSourcePoller.php`
4. `DELETE FROM wp_wc_admin_notes`
5. Run the `wc_admin_daily` cron task
6. The note about the eCommerce plan should _not_ appear in the inbox, but it should be a pending note in the `wp_wc_admin_notes` table
7. Execute this SQL: `INSERT INTO wp_options(option_name, option_value, autoload) VALUES('at_options', 'a:1:{s:9:"plan_slug";s:9:"ecommerce";}', 'yes')` - this flags the system as being on the eCommerce plan
8. Run the `wc_admin_daily` cron task again
9. The note about the eCommerce plan should now appear in the inbox

### DIFF
--- a/src/RemoteInboxNotifications/DataSourcePoller.php
+++ b/src/RemoteInboxNotifications/DataSourcePoller.php
@@ -16,7 +16,7 @@ defined( 'ABSPATH' ) || exit;
  */
 class DataSourcePoller {
 	const DATA_SOURCES = array(
-		'https://woocommerce.com/wp-json/wccom//inbox-notifications/1.0/notifications.json',
+		'https://woocommerce.com/wp-json/wccom/inbox-notifications/1.0/notifications.json',
 	);
 
 	/**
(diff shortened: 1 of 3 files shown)


psealock (approved):
Thanks @becdetat, this is testing well according to instructions.

This is probably nothing, but I did notice the note was "unactioned" not "pending" in this step:

>The note about the eCommerce plan should not appear in the inbox, but it should be a pending note in the wp_wc_admin_notes table

becdetat:
@psealock I'll retest this to make sure it's working as expected, if it's appearing before the ecommerce option is added then there's something amiss

becdetat:
@psealock it definitely works as I would expect. Possibly you had the `at_options` option set before running the task? I'll merge this and see if the issue comes up again.

PR #4691, opened 2020-06-24
Remove unnecessary rest API init action.

jeffstieler:
Fixes #4604.

This PR seeks to remove a seemingly unnecessary firing of the `rest_api_init` action, which mistakenly omits the expected `WP_REST_Server` instance in the action arguments.

### Detailed test instructions:

- Test several report exports that require an email be sent
- Verify no PHP notices/warnings/errors in your log

### Changelog Note:

Fix: Remove unnecessary `rest_api_init` action that caused incompatibility issues with other plugins.

PR #4659, opened 2020-06-22
Remove duplicate/redundant inbox note after first order received.

jeffstieler:
Fixes #4641.

This PR removes a redundant "first order" note and makes a small change to the wording of the remaining note.

Before:

> Congratulations on getting your first order! Now **it’s** a great time to learn how to manage your orders.

After:

> Congratulations on getting your first order! Now **is** a great time to learn how to manage your orders.

(bolded for visiblity)

### Screenshots

Screen Shot 2020-06-22 at 3 04 42 PM

### Detailed test instructions:

- Start with a new store (or one with no orders)
- Place an order
- Use WP Crontrol to run `wc_admin_process_orders_milestone` and `wc_admin_daily`
- Go to Inbox
- See only one note about the first order

### Changelog Note:

None - unreleased change.

PR #4780, opened 2020-07-08
Post-Release: Sync readme changelog, bump versions to 1.4.0-dev.

timmyc:
Few little cleanup items to get updated after the release of 1.3.0.

The one item to note here is I'm using the same versioning pattern taken by core and bumping `main` to 1.4.0-dev. I'm thinking this should probably happen once a beta release branch is created in the future. This would likely help avoid version number conflicts when doing db delta migrations in the future too.

But other changes beside version number:
- Synced the final changelog that was released to wporg to readme.txt, and updated the final release date.
- Bumped the stable tag and `WC tested up to` to 4.3

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -4,7 +4,7 @@ Tags: ecommerce, e-commerce, store, sales, reports, analytics, dashboard, activi
 Requires at least: 5.3.0
 Tested up to: 5.4.1
 Requires PHP: 5.6.20
-Stable tag: 1.3.0-beta.1
+Stable tag: 1.3.0
 License: GPLv3
 License URI: https://github.com/woocommerce/woocommerce-admin/blob/main/license.txt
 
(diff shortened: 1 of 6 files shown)


readme.txt
```diff
@@ -71,7 +71,7 @@ Release and roadmap notes are available on the [WooCommerce Developers Blog](htt
 
 == Changelog ==
 
-= 1.3.0-beta.1 2020-06-15 =
+= 1.3.0 2020-07-08 =
```
psealock, line 74: Here are some that were included in the RC's not added to this list.

- Dev: Check that the possibly_add_note function exists before calling it #4680
- Dev: Remove unnecessary rest API init action. #4691
- Dev: Don't include sourcemaps or unminified JS for "core" builds. #4642
- Tweak: Remove duplicate/redundant inbox note after first order received. #4659
- Tweak: Fix the embed page CSS so the top content sits better #4622
- Fix: Fix WCPay sometimes not appearing on the task list #4647

psealock (approved):
Thanks for the tidy up @timmyc. Definitely agree about updating `-dev` version as soon as the release branch is cut. Just some items to add to the changelog.

readme.txt
timmyc: Thanks Paul, i'll update those and merge them in. I think they might be missing from the WPORG changelog too since the list in this branch came from `release/1.3.0`

PR #4781, opened 2020-07-08
Update JP Autoloader to 2.0.0

timmyc:
See p7bje6-2ec-p2 for further details. While the release of this update should coincide with 1.4.0/4.4.0 - we should still test this branch alongside 4.3.0 and see how things go.

Might be nice to land #4780 prior to this to get a version number bump for ease and a more realistic test flow.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
 	"require": {
 		"composer/installers": "1.7.0",
 		"php": ">=5.6|>=7.0",
-		"automattic/jetpack-autoloader": "^1.6.0"
+		"automattic/jetpack-autoloader": "^2.0.0"
 	},
 	"require-dev": {
 		"phpunit/phpunit": "7.5.20",
--- a/composer.lock
+++ b/composer.lock
@@ -4,24 +4,24 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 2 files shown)


Aljullu (approved):

Aljullu:
Was testing this PR in parallel with https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2847 and by mistake I approved this one instead of the other. :facepalm:  Feel free to ignore my review above (even though I did some smoke testing with WC Admin and everything looked good).

mattsherman:
@timmyc Can you please provide some testing instructions for this? I'm not familiar enough with the Jetpack Autoloader to know what should be tested in this scenario.

timmyc:
@mattsherman I think just some basic smoke testing is the best we can really do with this one. @Aljullu by chance what steps did you take to test this out in the blocks repo?

Aljullu:
@timmyc I re-installed dependencies and built the project (for Blocks, running this is enough: `composer install && npm i && npm run build`). And then I just did some smoke testing to verify blocks loaded and still worked as expected. In https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2847 there are some specific testing steps but yes, I think the tl;dr is "smoke testing".

jeffstieler (approved):
Looks good to me - tested with core WooCommerce on an older JP autoloader and everything worked as expected (loading from plugin when enabled).

PR #4782, opened 2020-07-08
Added warning message when there isn't a payment method set up

### LINKED ISSUE #4596: Task List - Payments task: add alert message
**Description:** we want to reinforce the importance of setting up a payment method and increase user intent by adding an alert message about this topic

**Acceptance Criteria:**
- [ ] Add an alert message on each page of the Payments task reinforcing the fact that enabling a payment method is required to start accepting orders 

**Event tracking:**
NA 

**Additional links:**
[Design mockup](https://www.figma.com/file/JH9XMFUCOjfXdr3N09AHRD/On-boarding-iterations-June-'20?node-id=111%3A483)

octaedro:
Fixes #4596

This PR adds a warning message for each page of the Payments task if payment methods are incomplete

### Accessibility

### Screenshots
screenshot-one wordpress test-2020 07 08-16_22_11

- Go to the page: `Payment methods` (URL `/wp-admin/admin.php?page=wc-settings&tab=checkout`).
- Be sure, there isn't any payment method selected.
- Go to `home screen` / `dashboard` and press `Set up payments`.
- Verify the warning `Customers cannot place orders at your store until you setup and enable a payment method.` is visible.

Screen Capture on 2020-07-08 at 18-04-23

- Go back to `home screen` / `dashboard` and verify the message is not visible anymore.

- Go to the `Payment methods` again and select a payment method.

- Go to `home screen` / `dashboard` and press `Set up payments`.
- Verify the warning message is not visible.

- Verify the message continues to look as expected in different screen widths.

### Changelog Note:
Dev: Added warning message when there isn't a payment method set up

### DIFF
--- a/client/layout/store-notices/index.js
+++ b/client/layout/store-notices/index.js
@@ -28,10 +28,10 @@ class StoreWarnings extends Component {
 	}
 
 	render() {
-		const showMissingPaimentMethod = this.isVisibleMissingPaimentWarning();
+		const showMissingPaymentMethod = this.isVisibleMissingPaimentWarning();
 
 		return (
-			showMissingPaimentMethod && (
+			showMissingPaymentMethod && (
 				<Notice
 					className="woocommerce-store-warnings"
 					status="warning"
--- a/client/layout/store-notices/index.js
+++ b/client/layout/store-notices/index.js
@@ -16,13 +16,13 @@ import { OPTIONS_STORE_NAME } from '@woocommerce/data';
  */
 import './style.scss';
 import withSelect from 'wc-api/with-select';
-import { getUrlParams } from 'utils';
 
 class StoreWarnings extends Component {
 	isVisibleMissingPaimentWarning() {
 		const { taskListPayments, queryString } = this.props;
-		const urlParams = getUrlParams( queryString );
-		const isPaymentPage = urlParams.task ? urlParams.task === 'payments' : false;
+		const isPaymentPage = queryString.task
+			? queryString.task === 'payments'
+			: false;
 		return ! taskListPayments.completed && isPaymentPage;
 	}
 
@@ -36,19 +36,21 @@ class StoreWarnings extends Component {
 					status="warning"
 					isDismissible={ false }
 				>
-					<p>Customers cannot place orders at your store until you setup and enable a payment method.</p>
+					<p>
+						Customers cannot place orders at your store until you
+						setup and enable a payment method.
+					</p>
 				</Notice>
 			)
-			
 		);
 	}
 }
 
 StoreWarnings.propTypes = {
 	/**
-	 * The current url query string.
+	 * The url query string.
 	 */
-	queryString: PropTypes.string,
+	queryString: PropTypes.object,
 };
 
 export default compose(
@@ -58,5 +60,5 @@ export default compose(
 		return {
 			taskListPayments,
 		};
-	} ),
+	} )
 )( StoreWarnings );
--- a/client/layout/store-notices/index.js
+++ b/client/layout/store-notices/index.js
@@ -1,0 +1,62 @@
+/**
+ * External dependencies
+ */
+import { Component } from '@wordpress/element';
+import { compose } from '@wordpress/compose';
+import PropTypes from 'prop-types';
+import { Notice } from '@wordpress/components';
+
+/**
+ * WooCommerce dependencies
+ */
+import { OPTIONS_STORE_NAME } from '@woocommerce/data';
+
+/**
+ * Internal dependencies
+ */
+import './style.scss';
+import withSelect from 'wc-api/with-select';
+import { getUrlParams } from 'utils';
+
+class StoreWarnings extends Component {
+	isVisibleMissingPaimentWarning() {
+		const { taskListPayments, queryString } = this.props;
+		const urlParams = getUrlParams( queryString );
+		const isPaymentPage = urlParams.task ? urlParams.task === 'payments' : false;
+		return ! taskListPayments.completed && isPaymentPage;
+	}
+
+	render() {
+		const showMissingPaimentMethod = this.isVisibleMissingPaimentWarning();
+
+		return (
+			showMissingPaimentMethod && (
+				<Notice
+					className="woocommerce-store-warnings"
+					status="warning"
+					isDismissible={ false }
+				>
+					<p>Customers cannot place orders at your store until you setup and enable a payment method.</p>
+				</Notice>
+			)
+			
+		);
+	}
+}
+
+StoreWarnings.propTypes = {
+	/**
+	 * The current url query string.
+	 */
+	queryString: PropTypes.string,
+};
+
+export default compose(
+	withSelect( ( select ) => {
+		const { getOption } = select( OPTIONS_STORE_NAME );
+		const taskListPayments = getOption( 'woocommerce_task_list_payments' );
+		return {
+			taskListPayments,
+		};
+	} ),
+)( StoreWarnings );
--- a/client/layout/store-notices/index.js
+++ b/client/layout/store-notices/index.js
@@ -1,6 +1,7 @@
 /**
  * External dependencies
  */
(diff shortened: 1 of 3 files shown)


client/layout/store-notices/index.js
```diff
@@ -0,0 +1,99 @@
...
+
+		return (
+			showMissingPaimentMethod && (
+				<Notice
+					className="woocommerce-store-warnings"
+					status="warning"
+					isDismissible={ false }
+				>
+					<p>
+						Customers cannot place orders at your store until you
```
timmyc, line 40: This string needs to be translated.

client/layout/store-notices/index.js
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * WooCommerce dependencies
+ */
+import { OPTIONS_STORE_NAME } from '@woocommerce/data';
+
+/**
+ * Internal dependencies
+ */
+import './style.scss';
+import withSelect from 'wc-api/with-select';
```
timmyc, line 18: Since the options store is already on `@wordpress/data` seems like we should be able to use `import { withSelect } from '@wordpress/data';` instead as an external dependency?

timmyc (changes requested):
This is testing out nicely for me - responds immediately to the stat change as I interact with toggles on the payments task. Also looking good on narrow viewports - nice!

Couple of quick items need to be addressed, but should be good to go after that.

client/layout/store-notices/index.js
octaedro: Hi @timmyc, thank you for your review.
Nice catch! I added it in the commit d9c0a3d

client/layout/store-notices/index.js
octaedro: Good point! I modified it in the commit 96194bc

client/layout/store-notices/index.js
```diff
@@ -0,0 +1,102 @@
...
+	isVisibleMissingPaimentWarning() {
+		const { paymentMethods, queryString } = this.props;
+		const isPaymentPage = queryString.task
+			? queryString.task === 'payments'
+			: false;
+		return ! Object.keys( paymentMethods ).length && isPaymentPage;
+	}
+
+	render() {
+		const showMissingPaimentMethod = this.isVisibleMissingPaimentWarning();
```
becdetat, line 31: should be "showMissingPaymentMethod"

client/layout/store-notices/index.js
```diff
@@ -0,0 +1,102 @@
...
+	 * The url query string.
+	 */
+	queryString: PropTypes.object,
+};
+
+export default compose(
+	withSelect( ( select ) => {
+		const { getOption } = select( OPTIONS_STORE_NAME );
+
+		const optionNames = [
```
becdetat, line 63: I found that I had four options returned here, even after removing all of the payment provider plugins I had sitting around. I'm guessing that the settings option doesn't get removed when removing the plugin. Deleting the options made this work. You might want to see if there's a different way to check if a payment provider is installed - maybe using `getActivePlugins()` - https://github.com/woocommerce/woocommerce-admin/blob/main/client/homescreen/stats-overview/install-jetpack-cta.js#L165

client/layout/store-notices/index.js
```diff
@@ -0,0 +1,102 @@
...
+import { OPTIONS_STORE_NAME } from '@woocommerce/data';
+
+/**
+ * Internal dependencies
+ */
+import './style.scss';
+
+class StoreWarnings extends Component {
+	isVisibleMissingPaimentWarning() {
+		const { paymentMethods, queryString } = this.props;
```
becdetat, line 23: Could you do this in the `withSelect` below, so that you do all the logic in one place and just get a `showMissingPaymentMethod` prop? That would make it much easier to refactor when another warning gets added.

octaedro: Good point! I moved that logic in the commit 44ed4c7

client/layout/store-notices/index.js
octaedro: Nice catch @becdetat, I corrected it in the commit 31d418f

client/layout/store-notices/index.js
octaedro: Hi @becdetat, thank you for your feedback.
Your idea sounds good to me.

Just to confirm:
if the user has selected one of the following payment methods:  `PayPal, PayFast, Karna Checkout, Stripe, and Square` we should verify if the plugins related to those methods are also activated. If they aren't activated we'll assume the user doesn't want those methods, even though they appear as active in our options. Did I understand correctly?

Thinking about this, I wonder if we should also update the state of the inactive payment gateways to not have inconsistencies with the information we have in the DB.

timmyc: Really good point @becdetat - that hardcoded list of options caused me to pause a bit when I saw it too.

What do you all think about possibly just calling out to the core WC payment gateways endpoint: https://woocommerce.github.io/woocommerce-rest-api-docs/#list-all-payment-gateways

That might require a bit more work though since I don't think we have a data store setup for that yet, but maybe an [apiFetch](https://developer.wordpress.org/block-editor/packages/packages-api-fetch/) could work. 

Alternatively, we could consider this kind of an edge case for now - especially given the context of this warning being shown in the Setup Checklist.

timmyc:
Hey @jameskoster - I was hoping you could have a quick look here. So this is functioning per the spec, there are some interesting edge case scenarios that @becdetat pointed out where we could possibly make this a bit more resilient on the code side of things... but what is causing me to pause on shipping this in 1.4 is just the introduction of the contrasting alerts ( as you can see in the screenshot above ).

This very much echoes some of the feedback we got during the usability research for the new nav - but adding in another notification style with the core notif here, alongside the existing `<StoreAlert />` just to me looks a little _off_ visually.

While I appreciate the messaging we are trying to deliver in this PR, I'm wondering if we should invest some time in refactoring the `<StoreAlert />`s prior to begin shipping this form of messaging in wc-admin.

timmyc:
After a [conversation with Jay in slack](https://a8c.slack.com/archives/C9K5T0XSP/p1594919807125300) we are going to hold off on merging this one for 1.4 - and maybe seek to align the existing StoreAlerts with the core alert styling to make the experience a bit more visually cohesive. I think we should keep this branch around, but just give it a Status Blocked for now.

client/layout/store-notices/index.js
samueljseay: @octaedro also `isVisibleMissingPaimentWarning` should be `isVisibleMissingPaymentWarning`

pmcpinto:
@timmyc can we already implement this warning message?

timmyc:
While I like the spirit of this feature, I'm wondering if we should reconsider the presentation/design side of this with the new homescreen "task" area in mind. To me it seems like surfacing this message as an important "task" on the homescreen might be more valuable as it would cover use cases where someone has possibly dismissed the setup checklist, or is just simply ignoring it completely.

My primary concern with this design is it just creates yet another alert/message design in an already crowded notif world ( admin alerts, store alerts, inbox messages etc ).

pmcpinto:
Good point. @elizaan36 what do you think?

elizaan36:
Is this a problem we've discovered from support tickets or issues from users that dismiss the task list and then don't know how to set up payment methods? I'm trying to understand where the need came from and the extent of the problem to understand the best route to take.

pmcpinto:
> Is this a problem we've discovered from support tickets or issues from users that dismiss the task list and then don't know how to set up payment methods? I'm trying to understand where the need came from and the extent of the problem to understand the best route to take.

AFAIR it was to make sure that users really understand that they need to enable a payment option to get paid. FOMO effect

elizaan36:
ah ok! In that case I'd agree with Timmy 

> While I like the spirit of this feature, I'm wondering if we should reconsider the presentation/design side of this with the new homescreen "task" area in mind.

If we haven't experienced issues or tickets arising from users having trouble with this, it definitely feels like something we can solve for with highlighting the payments setup task in some way rather than presenting a notice.

PR #4790, opened 2020-07-09
Update store management links

jameskoster:
In our new WooCommerce navigation designs we're using the chevron icon to indicate when a menu item will reveal another panel via our "waterfall" pattern.

To promote consistency across our UX we should only use the chevron icon alongside that pattern moving forwards. Seeing as the items in the Store management card do not employ the waterfall pattern, we should remove the chevrons there so that we're ready for when the new navigation lands.

Before:
before

After:
after

Apologies if I butchered the code here but I replaced the chevrons with external icons for both of the `type: 'external'` links, and removed them for all other links.

I did notice however that those external links don't actually open new tabs currently. I'm not sure if that's something we should fix here or in the primitive component. Happy to open a new issue for that if necessary.

### DIFF
--- a/client/quick-links/index.js
+++ b/client/quick-links/index.js
@@ -13,13 +13,13 @@ import {
 	megaphone,
 	box,
 	brush,
+	home,
 	shipping,
 	percent,
 	payment,
 	pencil,
 	lifesaver,
 	external,
-	chevronRight,
 } from '@wordpress/icons';
 import { partial } from 'lodash';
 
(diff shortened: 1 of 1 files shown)


psealock (approved):
Thanks for handling this @jameskoster ! Code looks good :shipit:

jameskoster:
Merging this for now. We should update the icon for the "view my store" link when [this pr](https://github.com/WordPress/gutenberg/pull/23867) is merged.

PR #4792, opened 2020-07-09
Changed logic of "Set up payments" reminder

### LINKED ISSUE #4653: "Set up payments" reminder: change logic
With the removal of a skip button (#4638) we need to update the logic of triggering the following message: #3791

**Store type**: New users
**Trigger**: 5 days after store creation without having a payment method enabled: /wp-admin/admin.php?page=wc-settings&tab=checkout

octaedro:
Fixes #4653

This PR removes the use of `woocommerce_task_list_payments` from the note `Set up payments`.

### Accessibility

### Screenshots
screenshot-one wordpress test-2020 07 09-15_48_27

1. Disable all payment methods. You can do it here: `/wp-admin/admin.php?page=wc-settings&tab=checkout`
1. The note will be triggered after 5 days of the store creation without having a payment method enabled.
1. Check that a note is added.
1.  Click on the "Learn more" button in the note and make sure the note is actioned and you are taken to the wc payment extensions page.

### Changelog Note:
Fix Removed "woocommerce_task_list_payments" from the note "Set up payments".

### DIFF
--- a/src/Notes/WC_Admin_Notes_Onboarding_Payments.php
+++ b/src/Notes/WC_Admin_Notes_Onboarding_Payments.php
@@ -11,8 +11,6 @@ namespace Automattic\WooCommerce\Admin\Notes;
 
 defined( 'ABSPATH' ) || exit;
 
-use \Automattic\WooCommerce\Admin\Features\Onboarding;
-
 /**
  * WC_Admin_Notes_Onboarding_Payments.
  */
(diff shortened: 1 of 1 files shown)


timmyc (approved):
Verified the note does not show up if a payment method is active, but appears when none are active and the store is older than 5 days - thanks Fernando!

Issue #4597, opened 2020-06-16
Task List - highlight WCPay


pmcpinto:
**Description:** we want to explore a different approach to display WCPay in the task list for users that install it before arriving at the home screen (this should happen if we implement this #4579). Let’s assume a user opts in to installing WC Pay. In the task list we could show this user a “Set up WooCommerce Payments” task instead of the current “Set up Payments” task. In parallel, we'd deliver an inbox Notification highlighting other payment options, including existing partners. If the user didn’t install WC Pay during the Profiler, they would see the existing “Set up payments” task.

![screenshot-2020-06-03-at-19.16.39.png](https://images.zenhubusercontent.com/5d84fe8e1201de0001b6ab61/c3977c9e-daf7-4301-b969-89fcca546d94)

**Acceptance Criteria:**
- [ ] If the user has WCPay installed when he arrives at the home screen/task list for the first time, hide the "Set up payments" task and display a "Set up WooCommerce Payments" task. If the user doesn't have WCPay installed, he'll see the "Set up payments" task.
- [ ] When the user clicks on "Set up WooCommerce Payments task" he'll be automatically redirected to the WC Pay set up flow.
- [ ] Trigger the following inbox notification if the "Set up WooCommerce Payments" task is displayed:
**Title:** Set up additional payment providers
**Content:** Set up additional payment providers, using third-party services or other methods.
**Segmentation:** only users that have the "Set up WooCommerce Payments" task
**Timing:** when the user visits the home screen/task list for the first time
**Cadence:** once
**CTA leads to:** /wp-admin/admin.php?page=wc-admin&task=payments
**CTA label:** Set up payments

**Event tracking:**
Perhaps we can take advantage of the existing events on the task list and just add a new value to the task_name prop: wcpay

Events:
wcadmin_task_view
wcadmin_tasklist_task_completed
wcadmin_tasklist_click

**Additional links:**
[Design mockup](https://www.figma.com/file/JH9XMFUCOjfXdr3N09AHRD/On-boarding-iterations-June-'20?node-id=111%3A483)

becdetat:
Estimate: 2

timmyc:
I think Bec's estimate is a good one for this as it is outlined here.

Alternate idea though: Should we refactor the setup checklist steps to be more extensible? Seems like having the steps be filterable would be a great first step ( and allow for ease of customization in the ecommerce plan ). We could even consider implementing a slot fill for the setup checklist too.

[...]

PR #4793, opened 2020-07-10
Highlight WooCommerce Payments task

becdetat:
Partially addresses #4597

If WooCommerce Payments is installed, this replaces the "Set up payments" task with a "Set up WooCommerce Payments" task which redirects straight to the WooCommerce Payments setup flow.

### Screenshots
image

1. Make sure the WooCommerce Payments plugin isn't installed
2. Go to the home page and make sure the "Set up payments" step is present and the "Set up WooCommerce Payments" step is not present
3. Install WooCommerce Payments - you can do this by going in to "Set up payments" and setting up WooCommerce Payments from that screen
4. Go back to the home page and make sure the "Set up WooCommerce Payments" step is present and the "Set up payments" step is not present
5. Select the "Set up WooCommerce Payments" step and verify that it takes you to the WooCommerce Payments setup flow

### DIFF
--- a/client/task-list/index.js
+++ b/client/task-list/index.js
@@ -33,7 +33,7 @@ import {
  */
 import './style.scss';
 import CartModal from 'dashboard/components/cart-modal';
-import { getAllTasks } from './tasks';
+import { getAllTasks, recordTaskViewEvent } from './tasks';
 import { recordEvent } from 'lib/tracks';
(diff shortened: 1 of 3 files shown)


timmyc (approved):
This is testing out well for me locally, and also on a JN site. Confirmed the highlighted item only appears when WCPay is installed. Nice!

PR #4795, opened 2020-07-10
Admin note to install other payment providers

### LINKED ISSUE #4597: Task List - highlight WCPay

becdetat:
Fixes #4597

This creates a note when the WooCommerce Payments plugin is activated. After #4793 the "Set up payments" task is hidden when the WooCommerce Payments plugin is activated, so this gives a method of installing other payment providers.

### Screenshots
image

1. Deactivate the WooCommerce Payments plugin if it is installed
2. Check that the "set up additional payment types" note isn't shown in the inbox
3. Activate the WooCommerce Payments plugin
4. Check that the "set up additional payment types" not is now shown in the inbox
5. Check that clicking the "Set up payments" action navigates to the Payments task screen

### DIFF
--- a/client/task-list/index.js
+++ b/client/task-list/index.js
@@ -195,8 +195,27 @@ class TaskDashboard extends Component {
 	}
 
 	getCurrentTask() {
-		const { task } = this.props.query;
-		const currentTask = this.getTasks().find( ( s ) => s.key === task );
+		const {
+			query,
+			profileItems,
+			payments,
+			activePlugins,
+			installedPlugins,
+			createNotice,
+			isJetpackConnected,
+		} = this.props;
+		const allTasks = getAllTasks( {
+			profileItems,
+			options: payments,
+			query,
+			activePlugins,
+			installedPlugins,
+			createNotice,
+			isJetpackConnected,
+		} );
+		const { task } = query;
+		
+		const currentTask = allTasks.find( ( s ) => s.key === task );
 
 		if ( ! currentTask ) {
 			return null;
@@ -437,6 +456,7 @@ export default compose(
 			completedTaskKeys,
 			activePlugins,
 			installedPlugins,
+			payments,
 		};
 	} ),
 	withDispatch( ( dispatch ) => {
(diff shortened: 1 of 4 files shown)


octaedro (commented):
Hi @becdetat!
You did a really nice job here. The code looks good but I had a few problems testing it.

Do you have a testing `Stripe` user or some code to emulate the `WooCommerce Payments` plugin activation?

My problem is that to set up the `WooCommerce Payments` plugin I need to create a `Stripe` account and it's asking me for a US Social Security number, etc.
screenshot-connect stripe com-2020 07 13-13_46_33

Maybe I did something wrong trying to test the code so please let me know if there is another way to do it.

becdetat:
@octaedro I believe there is a way to set up a test account, however you can activate the WooCommerce Payments plugin without going through the Stripe setup - it's activated at the point where it asks you to set up WooCommerce Payments. You just have to go back to the home page and check the inbox for the new note that should appear when WooCommerce Payments is activated.

octaedro (approved):
Good job @becdetat!
The code looks really nice and following the steps you mentioned it's working well 🚢

PR #4800, opened 2020-07-11
Clean up webpack config and add tsconfig.json

samueljseay:
This doesn't fix a specific issue, but when I was investigating issues I had with the webpack build, I noticed some things we could clean up in the build that are not needed, I think its important to remove unused build config to reduce confusion and to make it easier to refactor or change in future.

I've also added a tsconfig, so that those using IDEs like VSCode can get better IDE support for code navigation and type hints without any extra effort.

Changes:

* Remove separate loader rule for `jsx`. None of our code uses this extension and it was excluding node_modules so even if it was somehow placed there to handle external dependencies it would not have done anything.

* Exclude node_modules from the babel-loader rule. I can't see any reason why we'd need to run babel over external dependencies at this point. This is consistent with other build tools like calypso-build

* Remove an include that was specifically running babel over some node_modules, similar to my reasoning above, we shouldn't need to run babel over these libraries, it is likely some legacy cruft in the build at this point.

* Remove a resolution rule that had a todo to remove after a PR was shipped (it has long since been shipped)

* Add a tsconfig based on calypso's one, but with added aliases for the `@woocommerce` packages.

- Do a basic smoke test of the application to check everything works ok. Look out for console errors or warnings that look unusual.


webpack.config.js
```diff
@@ -103,13 +103,9 @@ const webpackConfig = {
 					amd: false,
 				},
 			},
-			{
-				test: /\.jsx?$/,
-				loader: 'babel-loader',
-				exclude: /node_modules/,
-			},
 			{
 				test: /\.js?$/,
```
jeffstieler, line 107: Should this be `/\.jsx?$/`? I think it was errant before - unless we wanted to match `.j` files...

samueljseay: @jeffstieler Do we need to process `jsx` files, maybe leaving the x out we can discourage naming files jsx? good point about the `?`. How about `/\.js$/` ?

samueljseay: I've fixed this.

samueljseay:
@jeffstieler When you get a chance could you have another quick look at this? I addressed your feedback on that regex. 🙇

jeffstieler:
@samueljseay will do - though I'm hesitant to land this before we cut the `1.4.0` beta since it's cleanup that does have the potential to break the app.

samueljseay:
@jeffstieler That makes sense :) will hold off. no rush on it anyway

webpack.config.js
```diff
@@ -103,40 +103,15 @@ const webpackConfig = {
...
 					loader: 'babel-loader',
 					options: {
-						presets: [
-							[
-								'@babel/preset-env',
-								{ loose: true, modules: 'commonjs' },
-							],
-						],
-						plugins: [ 'transform-es2015-template-literals' ],
+						presets: [ '@wordpress/babel-preset-default' ],
```
samueljseay, line 112: This covers preset env and others. preset env handles template literals as well so no need for the plugin.

psealock (approved):
Changes look good to me and this building well. App works as expected. Thanks @samueljseay

PR #4805, opened 2020-07-14
New notification: Don't forget to test your checkout

### LINKED ISSUE #4213: New notification: Don't forget to test your checkout
**Title:** Don't forget to test your checkout
**Content:**  Make sure that your checkout is working properly before your launch your store. Go through your checkout process in its entirety: from adding a product to your cart, choosing a shipping location, and making a payment.
**Segmentation:**
**Timing:** 
"After triggering:
`wcadmin_product_add_publish` or `wcadmin_product_import_complete`
AND
`task_name: payments` on `wcadmin_tasklist_task_completed`
**Cadence:** Once
**CTA leads to:** Open store frontend in a new tab
**CTA label:** Test checkout

octaedro:
Fixes #4213

This PR adds a note to remind the user to test the checkout process.
This note should be triggered when the user has a payment method AND adds or imports products.

**Segmentation:**
**Timing:** 
"After triggering:
`wcadmin_product_add_publish` or `wcadmin_product_import_complete`
AND
`task_name: payments` on `wcadmin_tasklist_task_completed`
**Cadence:** Once

### Screenshots
screenshot-one wordpress test-2020 07 14-12_54_00

- Execute the cron. To do it you can install [this plugin](https://wordpress.org/plugins/wp-crontrol/) and go to `/wp-admin/tools.php?page=crontrol_admin_manage_page`and press `wc_admin_daily`.
- Add a new product.
- At the end of the process, the note should be visible.
- Remove the note and the variable: `wc_admin_note_test_checkout_product_added` from the database with a sentence like this:
````
DELETE FROM `wp_options` WHERE `wp_options`.`option_name` = 'wc_admin_note_test_checkout_product_added';
DELETE FROM `wp_wc_admin_notes` WHERE `wp_wc_admin_notes`.`name` = 'wc-admin-test-checkout';
````
- Import products.
- At the end of the process, the note should be visible.

- Verify the call to action button in the note opens the store frontend in a new tab

### Changelog Note:
Dev: New notification: Don't forget to test your checkout

### DIFF
--- a/src/Notes/WC_Admin_Notes_Test_Checkout.php
+++ b/src/Notes/WC_Admin_Notes_Test_Checkout.php
@@ -25,72 +25,21 @@ class WC_Admin_Notes_Test_Checkout {
 	 */
 	const NOTE_NAME = 'wc-admin-test-checkout';
 
-	const PRODUCT_ADDED_OPTION_NAME = 'wc_admin_note_test_checkout_product_added';
+	const TASK_LIST_PAYMENTS      = 'woocommerce_task_list_payments';
+	const TASK_LIST_TRACKED_TASKS = 'woocommerce_task_list_tracked_completed_tasks';
 
 	/**
 	 * Constructor.
 	 */
 	public function __construct() {
-		add_action( 'product_page_product_importer', array( $this, 'run_on_product_importer_done' ) );
-		add_action( 'transition_post_status', array( $this, 'run_on_transition_post_status_done' ), 10, 3 );
-	}
-
-	/**
-	 * Runs on product importer steps.
-	 */
-	public function run_on_product_importer_done() {
-		// We're only interested in when the importer completes.
-		// phpcs:disable WordPress.Security.NonceVerification.Recommended
-		if ( ! isset( $_REQUEST['step'] ) ) {
-			return;
-		}
-		if ( 'done' !== $_REQUEST['step'] ) {
-			return;
-		}
-		// phpcs:enable
-
-		$this->update_product_added_and_possibly_add_note();
-	}
-
-	/**
-	 * Runs when a post status transitions, but we're only interested if it is
-	 * a product being published.
-	 *
-	 * @param string $new_status The new status.
-	 * @param string $old_status The old status.
-	 * @param Post   $post       The post.
-	 */
-	public function run_on_transition_post_status_done( $new_status, $old_status, $post ) {
-		if (
-			'product' !== $post->post_type ||
-			'publish' !== $new_status
-		) {
-			return;
-		}
-
-		$this->update_product_added_and_possibly_add_note();
-	}
-
-	/**
-	 * Updates "wc_admin_note_test_checkout_product_added" in the DB and calls "possibly_add_note" method.
-	 */
-	public function update_product_added_and_possibly_add_note() {
-		update_option( self::PRODUCT_ADDED_OPTION_NAME, true );
-		self::possibly_add_note();
+		add_action( 'update_option_' . self::TASK_LIST_PAYMENTS, array( $this, 'possibly_add_note' ) );
+		add_action( 'update_option_' . self::TASK_LIST_TRACKED_TASKS, array( $this, 'possibly_add_note' ) );
 	}
 
 	/**
 	 * Get the note.
 	 */
 	public static function get_note() {
-		// Make sure a product was published or imported.
-		$product_added = get_option(
-			self::PRODUCT_ADDED_OPTION_NAME
-		);
-		if ( ! $product_added ) {
-			return;
-		}
-
 		$onboarding_profile = get_option( 'woocommerce_onboarding_profile', array() );
 
 		// Confirm that $onboarding_profile is set.
@@ -107,9 +56,30 @@ class WC_Admin_Notes_Test_Checkout {
 			return;
 		}
 
-		// Make sure payments task is completed.
-		$payments_task = get_option( 'woocommerce_task_list_payments', array() );
+		// Make sure payments task was completed.
+		$payments_task = get_option( self::TASK_LIST_PAYMENTS, array() );
 		if ( ! isset( $payments_task['completed'] ) ) {
+			return;
+		}
+
+		// Make sure that products were added within the previous 1/2 hour.
+		$query = new \WC_Product_Query(
+			array(
+				'limit'   => 1,
+				'status'  => 'publish',
+				'orderby' => 'date',
+				'order'   => 'ASC',
+			)
+		);
+
+		$products = $query->get_products();
+		if ( 0 === count( $products ) ) {
+			return;
+		}
+
+		$oldest_product_timestamp = $products[0]->get_date_created()->getTimestamp();
+		$half_hour_in_seconds     = 0.5 * HOUR_IN_SECONDS;
+		if ( ( time() - $oldest_product_timestamp ) > $half_hour_in_seconds ) {
 			return;
 		}
 
--- a/src/Notes/WC_Admin_Notes_Test_Checkout.php
+++ b/src/Notes/WC_Admin_Notes_Test_Checkout.php
@@ -1,0 +1,128 @@
+<?php
+/**
+ * WooCommerce Admin Test Checkout.
+ *
+ * Adds a note to remind the user to test their store checkout.
+ *
+ * @package WooCommerce Admin
+ */
+
+namespace Automattic\WooCommerce\Admin\Notes;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * WC_Admin_Notes_Test_Checkout
+ */
+class WC_Admin_Notes_Test_Checkout {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/Notes/WC_Admin_Notes_Test_Checkout.php [hunk 1]
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	/**
+	 * Get the note.
+	 */
+	public static function get_note() {
+		// Make sure a product was published or imported.
+		$product_added = get_option(
+			self::PRODUCT_ADDED_OPTION_NAME
+		);
```
jeffstieler, line 89: Rather than hook into the importer and post status transition hooks, what do you think about checking the status of the products step in the task list instead? Seems like this could be simpler if we were just checking for the products and payments steps to be complete.

src/Notes/WC_Admin_Notes_Test_Checkout.php
```diff
@@ -0,0 +1,128 @@
...
+		$content = __( 'Make sure that your checkout is working properly before your launch your store. Go through your checkout process in its entirety: from adding a product to your cart, choosing a shipping location, and making a payment.', 'woocommerce-admin' );
+
+		$note = new WC_Admin_Note();
+		$note->set_title( __( 'Don\'t forget to test your checkout', 'woocommerce-admin' ) );
+		$note->set_content( $content );
+		$note->set_content_data( (object) array() );
+		$note->set_type( WC_Admin_Note::E_WC_ADMIN_NOTE_INFORMATIONAL );
+		$note->set_name( self::NOTE_NAME );
+		$note->set_source( 'woocommerce-admin' );
+		$note->add_action( 'test-checkout', __( 'Test checkout', 'woocommerce-admin' ), site_url() );
```
jeffstieler, line 125: If we use `wc_get_page_permalink( 'shop' )` here it will direct them to the shop, and [automatically falls back to the site url](https://github.com/woocommerce/woocommerce/blob/c4afd53749c128afb61d0e00fd6e22d46c8fcea5/includes/wc-page-functions.php#L70).

src/Notes/WC_Admin_Notes_Test_Checkout.php
timmyc: Really like this idea Jeff!

octaedro: Hi @jeffstieler, thanks for your review.
Diving a little bit into this part of the code, I realized that the check (that later will be used to [set the products state](https://github.com/woocommerce/woocommerce-admin/blob/main/src/Features/OnboardingTasks.php#L111) in the [task list](https://github.com/woocommerce/woocommerce-admin/blob/main/client/task-list/tasks.js#L137)) is calculated in the backend in the method `check_task_completion ` and it's [asking for products with state published or draft](https://github.com/woocommerce/woocommerce-admin/blob/main/src/Features/OnboardingTasks.php#L181). From what I understand, this could be a problem since we need to show the note after the product publication, but the list item could be set as completed without a published product.
Please correct me if I misunderstood your suggestion or I'm forgetting something.

src/Notes/WC_Admin_Notes_Test_Checkout.php
octaedro: Good idea! I replaced the method in the commit 58c866a

src/Notes/WC_Admin_Notes_Test_Checkout.php
jeffstieler: Hmm.. looks like that wasn't the original intention and was updated later: https://github.com/woocommerce/woocommerce-admin/pull/2667#discussion_r305182796

Perhaps we should look at only marking the step complete if there are published products? cc: @timmyc

timmyc: Yeah I concur there, but cc @pmcpinto to verify for sure. The product step right now will get marked as done even when a product is in the draft state. But I definitely like tying this note to that step being completed because it will eliminate this note possibly being shown on an established site.

pmcpinto:
> The product step right now will get marked as done even when a product is in the draft state. But I definitely like tying this note to that step being completed because it will eliminate this note possibly being shown on an established site.

Can we use the store creation date to avoid show this to existing stores? We have a good amount of new users that add a product without going through the task list step. Is this step being marked as completed even if they add a product without passing by this task?

src/Notes/WC_Admin_Notes_Test_Checkout.php
octaedro: Thank you @jeffstieler and @timmyc, I removed the verification of the products with state `draft` in the commit 14d5df2 and it looks like everything is working well.

I was thinking about when this note should be shown. If the user doesn't set a payment method when the `Add my products` task list is completed, the note won't be shown. Also, we don't save a timestamp when we update the products' task list, which means that we don't know when we completed this task list item. 

The note reminds the user `Don't forget to test your checkout`, so it looks like we should show the note if the user adds their first product and has the other conditions we need. For instance, a user could have everything set but no products for a month. Should we show the note if they suddenly add a product?

If the answer is yes, I think that we should verify the published date of the first product, and if it's within the last 30 (or 15, or 10) minutes and fulfills the other conditions, I would trigger the note.
Does this make sense?

octaedro:
> Can we use the store creation date to avoid show this to existing stores?

Hi @pmcpinto,
Looking at the notification message I was thinking that maybe we could do the verification using the date of the first published product to show this note.
I'm thinking about the case that a user could have everything set in their store but no products for a month. After adding their first product we can incentivize them to take a look at their checkout process.
Maybe we can check if the user added a product in the last half hour to show this note.
Does this make sense?

>We have a good amount of new users that add a product without going through the task list step. Is this step being marked as completed even if they add a product without passing by this task?

Yes, we mark the step as complete even if the user adds a product without passing by this task.

pmcpinto:
> Does this make sense?

Yep, sounds good to me. Thanks for the suggestion!

octaedro:
According to [Pedro's answer](https://github.com/woocommerce/woocommerce-admin/pull/4805#issuecomment-661186494) I added the time validation (30 minutes) to the first published product in the commit 1eb7a0f.
At the same time, now we're checking the status of the products step in the task list (as @jeffstieler suggested) and I added another hook to do the same with the payment methods step.
Also, I did a rebase since there were a few conflicts between branches.
I think that now this PR is ready for another look.

src/Notes/WC_Admin_Notes_Test_Checkout.php
```diff
@@ -0,0 +1,98 @@
...
+			return;
+		}
+
+		// Make sure that products were added within the previous 1/2 hour.
+		$query = new \WC_Product_Query(
+			array(
+				'limit'   => 1,
+				'status'  => 'publish',
+				'orderby' => 'date',
+				'order'   => 'ASC',
```
jeffstieler, line 71: We should specify `'return' => 'ids'` here so we aren't querying data unnecessary to get a count.

src/Notes/WC_Admin_Notes_Test_Checkout.php
```diff
@@ -0,0 +1,98 @@
...
+			)
+		);
+
+		$products = $query->get_products();
+		if ( 0 === count( $products ) ) {
+			return;
+		}
+
+		$oldest_product_timestamp = $products[0]->get_date_created()->getTimestamp();
+		$half_hour_in_seconds     = 0.5 * HOUR_IN_SECONDS;
```
jeffstieler, line 81: Nitpick: `30 * MINUTE_IN_SECONDS` would avoid `float` and `int` comparison.

src/Notes/WC_Admin_Notes_Test_Checkout.php
```diff
@@ -0,0 +1,98 @@
...
+		if ( ( time() - $oldest_product_timestamp ) > $half_hour_in_seconds ) {
+			return;
+		}
+
+		$content = __( 'Make sure that your checkout is working properly before your launch your store. Go through your checkout process in its entirety: from adding a product to your cart, choosing a shipping location, and making a payment.', 'woocommerce-admin' );
```
jeffstieler, line 86: Typo: "before your launch your store" -> "before you launch your store".

jeffstieler (approved):
This tested well - just need to fix a typo and make some minor code tweaks. Pre-approving.

src/Notes/WC_Admin_Notes_Test_Checkout.php
octaedro: Good idea! I modified it in the commit 8d31053

src/Notes/WC_Admin_Notes_Test_Checkout.php
octaedro: Nice catch! I changed it in the commit 85a0054

octaedro:
As all the fixes have been addressed I'm going to merge this PR.

PR #4807, opened 2020-07-14
Setup Checklist: Fix Homepage Template

timmyc:
Fixes a bug reported by @LevinMedia in [slack today](https://a8c.slack.com/archives/C9K5T0XSP/p1594745851055700)

The template code used to create the homepage during the Setup Checklist is currently resulting in some sub-optimal output:

__Before__

Screen Shot 2020-07-14 at 10 07 55 AM

This branch implements a fix that is being used in Storefront to correct the bug: https://github.com/woocommerce/storefront/pull/1422

__After__

image

### Test Instructions

To prep for this test you will need to reset your setup checklist to a clean slate mode - the following commands should get you poised for testing success. Some might be needed, but it could save you some steps.

```
wp option delete woocommerce_task_list_complete
wp option delete woocommerce_task_list_hidden
wp option delete woocommerce_onboarding_homepage_post_id
```

And now you can then do the following test passes, the first time with less than 4 products present on the site, and the second time with 4 or more products as the template changes for those two. You will have to `wp option delete woocommerce_onboarding_homepage_post_id` between the tests.

- Visit `/wp-admin/admin.php?page=wc-admin&task=appearance`
- Click 'Create homepage' button on the step:

image

- Then visit `/wp-admin/edit.php?post_type=page&orderby=date&order=desc` and your freshly created home page should be at the top of the list. Edit it and verify all looks good ( like the after screenie above )

### Changelog Note:

Fix: Homepage template used in setup checklist customization task.

### DIFF
--- a/src/API/OnboardingTasks.php
+++ b/src/API/OnboardingTasks.php
@@ -264,27 +264,19 @@ class OnboardingTasks extends \WC_REST_Data_Controller {
 				<!-- wp:heading {"align":"center"} -->
 				<h2 style="text-align:center">' . __( 'New In', 'woocommerce-admin' ) . '</h2>
 				<!-- /wp:heading -->
-				<!-- wp:woocommerce/product-new {"columns":4} -->
-				<div class="wp-block-woocommerce-product-new">[products limit="4" columns="4" orderby="date" order="DESC"]</div>
-				<!-- /wp:woocommerce/product-new -->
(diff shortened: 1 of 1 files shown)


nerrad (approved):
Changes look good to me!

timmyc:
Thanks for the review @nerrad 🙌

PR #4808, opened 2020-07-14
Fix a React warning in test that the key prop wasn't assigned in List

samueljseay:
An alternative approach I considered was giving all the data used
in tests a `key` prop because that also would have solved it, but
it made me wonder if all these types of data even have key props
on them. Without any type information to validate something like that, I felt
the safer choice was just to use the array index as the key for the
top level node instead.

1. run the tests via `npm test`. You should not see any warnings from React in the output.

### DIFF
--- a/packages/components/src/list/index.js
+++ b/packages/components/src/list/index.js
@@ -49,7 +49,6 @@ class List extends Component {
 						className: itemClasses,
 						content,
 						href,
-						key,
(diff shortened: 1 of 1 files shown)


timmyc (approved):
All green and the warning about keys has faded into the 🌆 - thanks for the fix Sam!

PR #4816, opened 2020-07-15
Fix missing version update callback and add test

jeffstieler:
This PR adds the missing DB update for version `1.4.0` and introduces a test to verify that all DB update callbacks have been defined and the expected version number callback is present.

- Check the code
- Verify the tests pass

### Changelog Note:

N/A


timmyc (approved):
Thanks for catching this Jeff and for adding the test for future me whom might again forget to add a db update callback when bumping the version in `main`

PR #4828, opened 2020-07-16
Update profiler reference links and loading state

### LINKED ISSUE #4821: Revisiting the profiler links and styling
**Describe the bug**
When revisiting the profiler from the "Help" tab under WC screens, the `completed` prop gets reset.  With https://github.com/woocommerce/woocommerce-admin/pull/4721 being merged in, we can simply redirect to this screen.

However, there's also a flash of the wp admin menu that occurs since `should_show_profiler()` (which hides the UI until things are loaded) is no longer `true`.

We may be able to use the path `/profiler` to detect loading of this screen and hide things until loaded.

**To Reproduce**

1. Visit `wp-admin/admin.php?page=wc-admin&path=%2Fprofiler`
1. Note the wp admin menu is shown initially when loading.
1. Go to the Products page.
1. Under the help tab, use the button to revisit the profiler.
1. Note the profiler is "reset" (no longer marked as `completed`).

**Expected behavior**
Expected to be able to revisit the profiler and not be required to walk through to the end to complete it.

joshuatf:
Fixes #4821

* Updates the links to revisit the profiler.
* Hides the wp admin UI when loading profiler routes.

### Screenshots
Screen Shot 2020-07-16 at 4 00 23 PM

1. Go to the WC Products page.
1. Click the "Help" tab and click "Setup wizard."
1. Click the setup wizard and make sure you are able to visit the profiler.
1. Make sure when loading this page or refreshing, the wp admin menu is not shown.

### DIFF
--- a/src/Features/Onboarding.php
+++ b/src/Features/Onboarding.php
@@ -760,10 +760,16 @@ class Onboarding {
 	public function is_loading( $is_loading ) {
 		$show_profiler = self::should_show_profiler();
 		$is_dashboard  = ! isset( $_GET['path'] ); // phpcs:ignore csrf ok.
+		$is_profiler   = isset( $_GET['path'] ) && '/profiler' === $_GET['path']; // phpcs:ignore csrf ok.
+
+		if ( $is_profiler ) {
+			return true;
+		}
 
 		if ( ! $show_profiler || ! $is_dashboard ) {
 			return $is_loading;
 		}
+
 		return true;
 	}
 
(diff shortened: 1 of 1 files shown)


timmyc:
@joshuatf curious for your thoughts here on landing this before the 1.4 code freeze. Since it feels like it is more useful for those who often re-launch the profile wizard ( likely, mainly used internally ) this kind of seems like something that could land after we cut the beta release branch.

becdetat (approved):
Code looks great and it tests perfectly. :shipit:

joshuatf:
@timmyc The profiler still works without this, it just has some odd behavior with resetting and requiring re-completion as well as a flash of the WP Admin dashboard while loading.  I don't think this is necessary to include include before the code freeze.

Are we cherry picking for the code freeze or just holding off on merging PRs in and using `main`?

timmyc:
We haven't created the release branch yet, so anything that goes in main at this point will be in 1.4. If you are feeling confident about this change Josh, it has an approval to merge, guess it makes sense to land it.

PR #4831, opened 2020-07-16
Fix PHP Fatals from Notes DataStore when columns are missing.

### LINKED ISSUE #4823: Invalid data store for admin note after update to version 4.3.0, my site broken
After update to version 4.3.0, my site http://www.toonsmag.com broken. Not working, white screen. In the admin dashboard, it says, “There has been a critical error on your website. Please check your site admin email inbox for instructions”. In frontend “There has been a critical error on your website”.
Is there anyone who has an idea of how to fix the issue?

Site send me an email with the following message:

WordPress version 5.4.2
Current plugin: WooCommerce (version 4.3.0)
PHP version 7.2.31

Error Details
=============
An error of type E_ERROR was caused in line 127 of the file /home/toonsmag/public_html/wp-content/plugins/woocommerce/packages/woocommerce-admin/src/Notes/DataStore.php. Error message: Uncaught Exception: Invalid data store for admin note. in /home/toonsmag/public_html/wp-content/plugins/woocommerce/packages/woocommerce-admin/src/Notes/DataStore.php:127
Stack trace:
#0 /home/toonsmag/public_html/wp-content/plugins/woocommerce/includes/class-wc-data-store.php(159): Automattic\WooCommerce\Admin\Notes\DataStore->read(Object(Automattic\WooCommerce\Admin\Notes\WC_Admin_Note))
woocommerce/woocommerce#1 /home/toonsmag/public_html/wp-content/plugins/woocommerce/packages/woocommerce-admin/src/Notes/WC_Admin_Note.php(87): WC_Data_Store->read(Object(Automattic\WooCommerce\Admin\Notes\WC_Admin_Note))
woocommerce/woocommerce#2 /home/toonsmag/public_html/wp-content/plugins/woocommerce/includes/admin/notes/class-wc-notes-run-db-update.php(273): Automattic\WooCommerce\Admin\Notes\WC_Admin_Note->__construct(‘1’)
woocommerce/woocommerce#3 /home/toonsmag/public_html/wp-includes/class-wp-hook.php(287): WC_Notes_Run_Db_Update::show_reminder(Object(WP_Screen))
woocommerce/woocommerce#4 /home/toonsmag/public_html/wp-includes/class-wp-hook.php(311): WP_Hook->apply_filters(NULL, Array)
woocommerce/woocommerce#5 /home/toonsmag/p

jeffstieler:
Fixes #4823.

This PR seeks to avoid PHP Fatal Errors being thrown when `dbDelta()` fails adding new Notes columns.

While this can't avoid `WordPress database error` messages in the error log, it should stop the stores from completely breaking.

cc: @peterfabian 

- If your test store isn't already experiencing this issue:
- Rename the `layout` and/or `image` and/or `is_deleted` columns in the `wc_admin_notes` table
- Load any admin page
- Verify no PHP Fatals

### Changelog Note:

Fix: PHP Fatal errors when columns are missing from the Notes table.

### DIFF
--- a/src/Notes/DataStore.php
+++ b/src/Notes/DataStore.php
@@ -124,7 +124,7 @@ class DataStore extends \WC_Data_Store_WP implements \WC_Object_Data_Store_Inter
 			 */
 			do_action( 'woocommerce_note_loaded', $note );
 		} else {
-			throw new \Exception( __( 'Invalid data store for admin note.', 'woocommerce-admin' ) );
+			throw new \Exception( __( 'Invalid admin note', 'woocommerce-admin' ) );
 		}
 	}
 
--- a/src/Notes/DataStore.php
+++ b/src/Notes/DataStore.php
@@ -73,7 +73,7 @@ class DataStore extends \WC_Data_Store_WP implements \WC_Object_Data_Store_Inter
 		if ( 0 !== $note_id || '0' !== $note_id ) {
 			$note_row = $wpdb->get_row(
 				$wpdb->prepare(
-					"SELECT name, type, locale, title, content, content_data, status, source, date_created, date_reminder, is_snoozable, layout, image FROM {$wpdb->prefix}wc_admin_notes WHERE note_id = %d LIMIT 1",
+					"SELECT * FROM {$wpdb->prefix}wc_admin_notes WHERE note_id = %d LIMIT 1",
 					$note->get_id()
 				)
 			);
@@ -342,7 +342,7 @@ class DataStore extends \WC_Data_Store_WP implements \WC_Object_Data_Store_Inter
 
 		$query = $wpdb->prepare(
 			// phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared, WordPress.DB.PreparedSQL.InterpolatedNotPrepared
-			"SELECT note_id, title, content, layout, image FROM {$wpdb->prefix}wc_admin_notes WHERE 1=1{$where_clauses} ORDER BY {$args['orderby']} {$args['order']} LIMIT %d, %d",
+			"SELECT * FROM {$wpdb->prefix}wc_admin_notes WHERE 1=1{$where_clauses} ORDER BY {$args['orderby']} {$args['order']} LIMIT %d, %d",
 			$offset,
 			$args['per_page']
 		);
--- a/src/Notes/NoteTraits.php
+++ b/src/Notes/NoteTraits.php
@@ -99,9 +99,11 @@ trait NoteTraits {
 		$note_ids   = $data_store->get_notes_with_name( self::NOTE_NAME );
 
 		foreach ( $note_ids as $note_id ) {
-			$note = new WC_Admin_Note( $note_id );
+			$note = WC_Admin_Notes::get_note( $note_id );
 
-			$data_store->delete( $note );
+			if ( $note ) {
+				$data_store->delete( $note );
+			}
 		}
 	}
 }
--- a/src/Notes/WC_Admin_Note.php
+++ b/src/Notes/WC_Admin_Note.php
@@ -77,6 +77,7 @@ class WC_Admin_Note extends \WC_Data {
 			$this->set_id( $data );
 		} elseif ( is_object( $data ) && ! empty( $data->note_id ) ) {
 			$this->set_id( $data->note_id );
+			unset( $data->icon ); // Icons are deprecated.
 			$this->set_props( (array) $data );
 			$this->set_object_read( true );
 		} else {
--- a/src/Notes/WC_Admin_Note.php
+++ b/src/Notes/WC_Admin_Note.php
@@ -85,7 +85,13 @@ class WC_Admin_Note extends \WC_Data {
 
 		$this->data_store = \WC_Data_Store::load( 'admin-note' );
 		if ( $this->get_id() > 0 ) {
-			$this->data_store->read( $this );
+			try {
+				$this->data_store->read( $this );
+			} catch ( \Exception $e ) {
+				wc_caught_exception( $e, __CLASS__ . '::' . __FUNCTION__, array( $this->get_id() ) );
+				$this->set_id( 0 );
+				$this->set_object_read( true );
+			}
 		}
 	}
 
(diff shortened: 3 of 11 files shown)


peterfabian:
Not at the laptop right now so can't confirm, but wouldn't it cause problems with `update` and `get_notes` methods on the data store as well?

jeffstieler:
> wouldn't it cause problems with `update` and `get_notes` methods on the data store as well?

The PHP Fatal was from the uncaught exception thrown in the `read` method. No other methods throw exceptions.

peterfabian:
Yes, you're technically right (the best kind of right :) ). If this PR should just fix the fatal error, maybe it's better to keep it smaller. 

However, I think those functions I mentioned will still produce database errors, as they refer to columns that (potentially) don't exist. Do you want me to create a new issue for that?

timmyc (commented):
Thank you *very much* for the quick turnaround on this Jeff. I tested the following:

- renamed all three columns added in 1.3 by appending with `_test`
- Verified no fatals happened on the site when viewing the home screen, and the inbox panel
- Truncated my notes + note action tables
- Ran `wc_admin_daily` via Crontrol - again no fatals, but the expected warnings in the log file were shown, and as expected, no new notes were created in the tables due to the missing columns.

Checked around for other instances of `get_note()` being called - and found one other spot that was wondering if we should add a check like you did in other notes:

- [WC_Admin_Notes_WooCommerce_Payments.php](https://github.com/woocommerce/woocommerce-admin/blob/main/src/Notes/WC_Admin_Notes_WooCommerce_Payments.php#L61)

Outside of that one item above, thinking this is looking good but would @peterfabian would be great to have a set of eyes from Proton too to double check.

jeffstieler:
> However, I think those functions I mentioned will still produce database errors, as they refer to columns that (potentially) don't exist.

I looked to see how core handles this (in the downloads data store) and it's issuing a "SELECT *" instead of specifying columns individually. I'll look at using that approach when I'm back at my computer.

jeffstieler:
After some more investigation, I'm not sure we can ever protect against database errors of this kind. Yes, we could potentially avoid specifically selecting newly added columns, but what about queries that use those columns in clauses? Are you thinking we'd always query for the table schemas before issuing any actual data queries @peterfabian?

As far as I can tell there isn't this level of defensive programming in core WC.

For example, the [`wc_tax_rate_classes` table](https://github.com/woocommerce/woocommerce/commit/5d1020300e109379ba733a8c4eefc2d81925b4d6) (one of the most recently added to core) isn't assumed to not exist [when being queried](https://github.com/woocommerce/woocommerce/blob/c38e059ce87addfea8a6ea58e7e5183402a9510f/includes/class-wc-tax.php#L755-L759).

jeffstieler:
Switched to avoiding specifying columns where possible in e3222eb.

peterfabian:
> After some more investigation, I'm not sure we can ever protect against database errors of this kind. Yes, we could potentially avoid specifically selecting newly added columns, but what about queries that use those columns in clauses? Are you thinking we'd always query for the table schemas before issuing any actual data queries @peterfabian?

> As far as I can tell there isn't this level of defensive programming in core WC.

> For example, the wc_tax_rate_classes table (one of the most recently added to core) isn't assumed to not exist when being queried.

I agree and I'm not sure what the best way to handle such cases is. Maybe there isn't a good way in some scenarios. We've only started considering these db changes with a more defensive approach quite recently as we've learned the plugin auto-update functionality will become available to sites with the upcoming WordPress 5.5 update, so we're also figuring it out as we go. 
First time that I know we considered this was the new table introduced by Blocks to reserve stock (`wc_reserved_stock`). If we implemented it as initially intended, it would prevent customers from checkout if the db table wasn't created (so quite bad and not immediately apparent problem). This one (at least the fatal) is maybe even worse because you can't do much if your site goes to WSOD on admin (but at least the front end still worked).

rrennick:
@jeffstieler This works great in the dashboard. When I run a WP CLI commend I get a large number of the following

```
[17-Jul-2020 13:40:39 UTC] PHP Notice:  Trying to access array offset on value of type null in phar:///usr/local/Cellar/wp-cli/2.4.0/bin/wp/vendor/wp-cli/extension-command/src/Plugin_Command.php on line 663
```

becdetat (approved):
Everything looks good to me, I followed the test instructions and then Timmy's instructions and everything works well.

jeffstieler:
@rrennick I was able to run `wp plugin list` without any errors.. is it perhaps some other plugin code you have running?

rrennick:
@jeffstieler Still get it with only WC `master` and the branch active. WP version is 5.4.2.

jeffstieler:
@rrennick which `wp-cli` command are you running? Here's what I'm seeing:

```
vagrant@vvv:/srv/www/wordpress-default/public_html$ wp core version
5.4.2
vagrant@vvv:/srv/www/wordpress-default/public_html$ wp plugin list --status=active
+-------------------+--------+------------------------------+-----------+
| name              | status | update                       | version   |
+-------------------+--------+------------------------------+-----------+
| woocommerce       | active | version higher than expected | 4.4.0-dev |
| woocommerce-admin | active | version higher than expected | 1.4.0-dev |
+-------------------+--------+------------------------------+-----------+
```

rrennick:
@jeffstieler I was using the same

```
$ wp plugin list --status=active
Notice: Trying to access array offset on value of type null in phar:///usr/local/Cellar/wp-cli/2.4.0/bin/wp/vendor/wp-cli/extension-command/src/Plugin_Command.php on line 662
Notice: Trying to access array offset on value of type null in phar:///usr/local/Cellar/wp-cli/2.4.0/bin/wp/vendor/wp-cli/extension-command/src/Plugin_Command.php on line 663
```

What version of WP CLI do you have?

jeffstieler:
> What version of WP CLI do you have?

`WP-CLI 2.5.0-alpha-02486ab`

peterfabian (approved):
Thanks, it solves the problem and I can't replicate the WP-CLI problem @rrennick mentioned. 

(Retracted my previous suggestion, as it only showed with the package included with 4.3, not the WCA plugin)

PR #4834, opened 2020-07-18
Set active theme when OBW is shown via the task list

becdetat:
Currently, using the OBW from the task list with the OBW previously complete doesn't load the `activeTheme` property, which means that the currently active theme isn't shown as active and there isn't a 'skip' link shown at the bottom of the theme step.

This PR moves where the `activeTheme` property is set so that it is loaded in this scenario.

*I'm not sure about this PR as it looks like not having the active theme was the intended behaviour, but it feels wrong and janky.*

1. Make sure the OBW is complete and the task list is shown
2. Start the OBW via the task list
3. Go through to the theme selection step
4. Verify that the currently selected theme is shown as selected
5. "Continue with my active theme" and verify that it works as expected
6. Work through to get back to the theme selection step
7. "Skip this step" and verify that it works as expected
8. Work through to get back to the theme selection step
9. Select another theme and make sure it works as expected

### DIFF
--- a/src/Features/Onboarding.php
+++ b/src/Features/Onboarding.php
@@ -583,13 +583,9 @@ class Onboarding {
 			'profile'    => $profile,
 		);
 
-		// Only fetch if the onboarding wizard is incomplete.
-		if ( self::should_show_profiler() ) {
-			$settings['onboarding']['activeTheme'] = get_option( 'stylesheet' );
-		}
-
 		// Only fetch if the onboarding wizard OR the task list is incomplete.
 		if ( self::should_show_profiler() || self::should_show_tasks() ) {
+			$settings['onboarding']['activeTheme']              = get_option( 'stylesheet' );
 			$settings['onboarding']['stripeSupportedCountries'] = self::get_stripe_supported_countries();
 			$settings['onboarding']['euCountries']              = WC()->countries->get_european_union_countries();
 			$current_user                                       = wp_get_current_user();


src/Features/Onboarding.php
```diff
@@ -583,13 +583,9 @@ public function component_settings( $settings ) {
 			'profile'    => $profile,
 		);
 
-		// Only fetch if the onboarding wizard is incomplete.
-		if ( self::should_show_profiler() ) {
-			$settings['onboarding']['activeTheme'] = get_option( 'stylesheet' );
-		}
-
 		// Only fetch if the onboarding wizard OR the task list is incomplete.
 		if ( self::should_show_profiler() || self::should_show_tasks() ) {
```
joshuatf, line 587: Maybe as a temporary workaround, we also check if the path is the profiler and load this data as well.

`isset( $_GET['path'] ) && '/profiler' === $_GET['path']`

joshuatf (commented):
Nice catch and fix on this, Bec!  This is testing well for me.

The only way I can break this is by completing the task list and profiler and then revisiting the profiler from the help tab.

We could preload this data all the time, but it feels wasteful and I think long-term we'll need to have a discussion about not preloading so much data and relying on pulling this info async.

Left one comment with a possible short-term fix.

becdetat:
Thanks @joshuatf :-)

> The only way I can break this is by completing the task list and profiler and then revisiting the profiler from the help tab.

I can't reproduce this. Opening the profiler from the help tab clears the `completed` value in the profile data, which makes `self::should_show_profiler()` return true and load the data. Is there something I'm missing?

jeffstieler (approved):
I also can't repro what @joshuatf is describing.

I am able to repro the condition where no theme shows as active, and this PR fixes it. Nice catch @becdetat! 🚢  it.

PR #4851, opened 2020-07-22
Remove item "profiler" from the menu

octaedro:
This PR removes the item "profiler" from the menu.

### Accessibility

### Screenshots
Before:
profiler-item-1

Now:
profiler-item-2

- Verify the item `Profiler` is no longer visible in the menu.

### Changelog Note:
Fix: Removed item "profiler" from the menu


timmyc:
Cherry picked to `release\1.4.0-beta.2` in be3c7b0

PR #4853, opened 2020-07-22
Task list - add a shortcut back to store setup

### LINKED ISSUE #4592: Task list - add a shortcut back to store setup
**User story:**
As a new merchant
I want to easily go back to the store setup checklist
So I don't feel lost when I'm exploring or setting up additional details

**Description:** there’s a lot to explore in WooCommerce, so it’s easy for people to become distracted during setup. For example, we noticed that when we were redirecting users to edit their homepage. By displaying a persistent widget we aim to remind people that store setup is incomplete and offer them a pathway back to the setup flow. As the wc-admin top bar has already been implemented across all WooCommerce screens, adding a tab here seems most sensible.

**Acceptance Criteria:**
- [x] Add a new item to the top bar that redirects the user to the home screen, when the setup task list is active
- [x] Hide the Orders, Stock, and Reviews tabs while the setup task list is active
- [x] Display the Orders, Stock, and Reviews tabs and hide the Store Setup item when the user finishes or hides the task list

Nice to have in this first iteration:
- [ ] The store setup icon should be updated depending on how close the user is to completing setup.
- [ ] Reuse the panel pattern employed by the other tabs in the activity panel, to display the task list on the current screen 

**Event tracking:**
We can use the existing `wcadmin_activity_panel_open` event and add the store setup item as a new value of the `tab` prop. As an alternative, we can create a new event just for this case.

**Additional links:**
[Design mockup](https://www.figma.com/file/JH9XMFUCOjfXdr3N09AHRD/On-boarding-iterations-June-'20?node-id=1%3A11)

samueljseay:
Fixes #4592 

This adds the functionality to support a store setup tab displayed in ActivityPanel. The tab currently just redirects to the home page, rather than implementing the nice-to-haves of displaying the task list in the panel. The reason for this, is that I found it would require significant refactoring of the task list to support this currently.

As part of this I have also refactored rendering of tabs inside <ActivityPanel> into 2 new functional components: `<Tabs>` and `<Tab>`. The reason for this is to decompose large untestable components, test them and by virtue of being decomposed they will also be more reusable in future. Ideally <ActivityPanel> would have been refactored to a functional component as well, but it would have been too large a task.

**The logic for showing the setup progress tab is something I'd be keen to have checked. At the moment it only displays if:**

* You're not on the home screen
* You haven't completed the setup task list
* You're not working on one of the setup tasks

### Accessibility

### Screenshots

Screenshot 2020-07-23 at 10 40 43 AM

#### Test the ActivityPanel refactor

Because there has been some refactoring of the ActivityPanel in general it would be really great to smoke test its general functionality such as:

* click outside close the menu
* switching between different panels (note the animations still work)
* view page at mobile size and check the mobile menu toggle still works.

#### Test the Setup Progress tab

Testing the setup progress tab will require having a site with some setup tasks not completed. 

* Visit the home screen and confirm its not shown there.
* Go to a setup task and confirm its not shown there
* Go to other menu items in WooCommerce and confirm its shown there.
* Click the tab. You should be taken back to the home screen


client/header/activity-panel/index.js
```diff
@@ -70,30 +66,35 @@ export class ActivityPanel extends Component {
...
+				isPanelOpen: isTabOpen,
+				mobileOpen: isTabOpen,
+				currentTab: tabName,
+				isPanelSwitching,
+			};
 		} );
 	}
 
+	closePanel() {
+		this.setState( ( state ) => ( {
```
samueljseay, line 87: one thing that was happening a lot before was that ActivityPanel was setting state to be a completely different shaped object in various scenarios. This lead to a lot of confusion about what state was at any given time. I've changed this to always keep state the same shape.

client/header/activity-panel/index.js
```diff
@@ -134,25 +135,45 @@ export class ActivityPanel extends Component {
...
-					taskListHidden === false &&
-					taskListComplete === false
-				)
-			);
+				( taskListHidden === false && taskListComplete === false ) );
 
-		return [
-			! isPerformingSetupTask && showInbox
-				? {
+		if ( ! taskListComplete && showInbox ) {
```
samueljseay, line 148: generating the tab list was just increasing further in complexity and its already quite hard to reason about, so I maintain a second list of tabs for displaying setup. I think this `getTabs` method is ripe for refactoring (its not really easy to test it right now), but it was too much of a diversion to fix it here.

client/header/activity-panel/index.js
```diff
@@ -185,34 +206,29 @@ export class ActivityPanel extends Component {
...
 
 	getPanelContent( tab ) {
+		const { hasUnreadOrders, query, hasUnapprovedReviews } = this.props;
+		const { task } = query;
+
 		switch ( tab ) {
 			case 'inbox':
 				return <InboxPanel />;
 			case 'orders':
-				const { hasUnreadOrders } = this.props;
```
samueljseay, line 202: Not a good idea to declare variables inside case's since they'll clash if you reuse them in separate cases.

client/header/activity-panel/index.js
```diff
@@ -221,14 +237,23 @@ export class ActivityPanel extends Component {
...
 				<div className="woocommerce-layout__activity-panel-wrapper" />
 			);
 		}
 
+		const clearPanel = () => {
+			this.clearPanel();
+		};
+
+		if ( currentTab === 'setup' ) {
+			window.location = getAdminLink( 'admin.php?page=wc-admin' );
```
samueljseay, line 253: Go to the main page if the setup tab was chosen.  This will go away when we implement the nice-to-have of showing the task list in the panel, so I haven't refactored tabs to support going to links for now.

client/header/activity-panel/style.scss
```diff
@@ -44,6 +44,15 @@
 		width: 18px;
 		height: 18px;
 
+		// custom progress icon, requires specific coloring
+		&.setup-progress {
```
samueljseay, line 48: Setup progress is different from all the other icons, its a 2 toned SVG rather than all the other tab icons which are dashicons.

client/header/activity-panel/index.js
timmyc: I agree that this could be a good refactor target too. Given some of the new designs around this part of the app though, I think the current panels are going to go away, and possibly be replaced with more contextual panels... but even on that approach, this would need to be refactored to be more extensible too.

client/header/activity-panel/index.js
```diff
@@ -134,25 +135,45 @@ export class ActivityPanel extends Component {
...
 					name: 'inbox',
 					title: __( 'Inbox', 'woocommerce-admin' ),
 					icon: <i className="material-icons-outlined">inbox</i>,
 					unread: hasUnreadNotes,
-				}
+				},
+				{
+					name: 'setup',
+					title: __( 'Store Setup', 'woocommerce-admin' ),
+					icon: <SetupProgress />,
```
timmyc, line 159: Cool ✨

client/header/activity-panel/tab/index.js
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	return (
+		<Button
+			role="tab"
+			className={ className }
+			tabIndex={ tabIndex }
+			aria-selected={ selected }
+			aria-controls={ `activity-panel-${ name }` }
+			key={ `activity-panel-tab-${ name }` }
```
timmyc, line 37: Minor, but since this same string is used 2x here should we just make a `const` for it?

client/header/activity-panel/tab/test/index.js
```diff
@@ -0,0 +1,247 @@
+import { render, fireEvent } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import PagesIcon from 'gridicons/dist/pages';
+
+import { Tab } from '../';
+
+const renderTab = () =>
```
timmyc, line 7: 😍 thanks for the tests!

client/header/activity-panel/tabs/index.js
```diff
@@ -0,0 +1,67 @@
...
+		} );
+	}, [ tabOpen, selectedTabName ] );
+
+	return (
+		<NavigableMenu
+			role="tablist"
+			orientation="horizontal"
+			className="woocommerce-layout__activity-panel-tabs"
+		>
+			{ tabs &&
```
timmyc, line 32: With this refactor, we could now have a filter right here for the tabs. Given the uncertain future of the activity panels though, my vote is to not add it quite yet - but still really cool that your changes have allowed for a good spot to add it in.

client/header/activity-panel/tabs/index.js
```diff
@@ -0,0 +1,67 @@
...
+				tabs.map( ( tab, i ) => (
+					<Tab
+						key={ i }
+						index={ i }
+						isPanelOpen={ tabState.tabOpen }
```
timmyc, line 37: Any thoughts on destructing `tabState` into the needed attributes used here - or do you like having `tabState.` and `tab.` to kind of keep the logic easier to follow about what attributes are being compared?

client/header/activity-panel/tabs/test/index.js
```diff
@@ -0,0 +1,95 @@
...
+	return [ '0', '1', '2', '3' ].map( ( name ) => ( {
+		name,
+		title: `Tab ${ name }`,
+		icon: <span>icon</span>,
+		unread: false,
+	} ) );
+};
+
+describe( 'Activity Panel Tabs', () => {
+	it( 'correctly tracks the selected tab', () => {
```
timmyc, line 19: Nice test here!

timmyc (commented):
Couple of quick comments, I haven't had a chance to take this for a full test-drive yet. The only other observation I had was is there a need to check in the changes to `package-lock.json`?

psealock:
I keep getting this error about chunks not loading, which we've seen before. Maybe a rebase is needed?

Screen Shot 2020-07-29 at 1 18 47 PM

client/header/activity-panel/index.js
psealock: Can we use `getHistory().push( ... )` to navigate here? If I click this on a wc-admin page, a full page refresh occurs.

client/header/activity-panel/index.js
psealock: There is some logic here that doesn't match. The criteria for showing the setup activity panel icon is `! taskListComplete`, but the task list also is hidden if `woocommerce_task_list_hidden` is set to "yes". In other words, it's possible to be in a state where the setup button is visible, but the task list isn't shown on the Homescreen once you've navigated there.

To reproduce, leave the task list incomplete and dismiss it via:

Screen Shot 2020-07-29 at 1 38 15 PM

We could match the criteria, but I think it might be cool to leave it as is, but set the option `woocommerce_task_list_hidden` to "no" before navigating. This way, a user can dismiss the setup task list, but still access it.

psealock (commented):
This is looking good and testing well. Just a comment about navigation and another about the visibility of the setup tab.

client/header/activity-panel/index.js
samueljseay: @psealock Thanks for the feedback, I've added a fix for this in https://github.com/woocommerce/woocommerce-admin/pull/4853/commits/0f243b2dd50e255879270e9e7039da156c08af84 Can you let me know if I'm on the right track? I tested it and it works as expected for me.

client/header/activity-panel/index.js
samueljseay: Have done this in https://github.com/woocommerce/woocommerce-admin/pull/4853/commits/0f243b2dd50e255879270e9e7039da156c08af84

client/header/activity-panel/tab/index.js
samueljseay: Fixed, thanks!

client/header/activity-panel/tabs/index.js
samueljseay: I think you could go either way with this, but I did refactor it to destructure the state. The attributes are a liittle more confusing but its also more succinct, so I'm happy with that change :+1:

samueljseay:
@timmyc You're right no need for the lock. Have reverted that.

client/header/activity-panel/index.js
```diff
@@ -190,40 +213,52 @@ export class ActivityPanel extends Component {
...
 
+		const clearPanel = () => {
+			this.clearPanel();
+		};
+
+		if ( currentTab === 'setup' ) {
+			// Ensure that if the user is trying to get to the task list they can see it even if
+			// it was dismissed.
+			updateOptions( { woocommerce_task_list_hidden: 'no' } );
+			getHistory().push( getAdminLink( 'admin.php?page=wc-admin' ) );
```
psealock, line 258: The url comes out to this:
```
http://tangaroa.test/wp-admin/http://tangaroa.test/wp-admin/http://tangaroa.test/wp-admin/admin.php?page=wc-admin
```
`getAdminLink` is for external links. Sorry about this not being documented anywhere! This will look more like

https://github.com/woocommerce/woocommerce-admin/blob/d2be77c0176773bad95ca720d195b3cfad9fdea4/client/task-list/tasks.js#L106

samueljseay: Whoops! Thanks, I'll fix that

client/header/activity-panel/index.js
```diff
@@ -191,40 +221,74 @@ export class ActivityPanel extends Component {
...
+				'admin.php?page=wc-admin'
+			);
+
+			// Don't navigate if we're already on the homescreen, this will cause an infinite loop
+			if ( currentLocation !== homescreenLocation ) {
+				// Ensure that if the user is trying to get to the task list they can see it even if
+				// it was dismissed.
+				updateOptions( {
+					woocommerce_task_list_hidden: 'no',
+				} ).then( () => {
```
psealock, line 274: This is a nice addition. I see below we gather `woocommerce_task_list_hidden`. Would it make sense to see if this option is already "no" and skip the update to avoid the waiting for the request?

client/header/activity-panel/index.js
```diff
@@ -71,30 +68,35 @@ export class ActivityPanel extends Component {
...
+				mobileOpen: isTabOpen,
+				currentTab: tabName,
+				isPanelSwitching,
+			};
 		} );
 	}
 
+	closePanel() {
+		this.setState( ( state ) => ( {
+			...state,
```
joshuatf, line 90: Doesn't React automatically merge the state when calling `setState()`?  Is there an advantage to destructuring the state here?

client/header/activity-panel/index.js
```diff
@@ -71,30 +68,35 @@ export class ActivityPanel extends Component {
...
+			currentTab: '',
+		} ) );
+	}
+
 	clearPanel() {
-		this.setState( ( { isPanelOpen } ) =>
-			isPanelOpen ? { isPanelSwitching: false } : { currentTab: '' }
-		);
+		this.setState( ( state ) => {
+			return { ...state, isPanelSwitching: false };
```
joshuatf, line 98: Same question on this one as above.

client/header/activity-panel/index.js
```diff
@@ -191,40 +221,74 @@ export class ActivityPanel extends Component {
...
+			);
+
+			// Don't navigate if we're already on the homescreen, this will cause an infinite loop
+			if ( currentLocation !== homescreenLocation ) {
+				// Ensure that if the user is trying to get to the task list they can see it even if
+				// it was dismissed.
+				updateOptions( {
+					woocommerce_task_list_hidden: 'no',
+				} ).then( () => {
+					const isAdmin = /admin.php\?page=wc-admin/.test(
```
joshuatf, line 275: Nice!  I see we also use a test like this in `isWCAdmin`.  This would be fantastic as a reusable util.

client/header/activity-panel/tab/test/index.js
joshuatf: Love all these tests 💯

joshuatf (commented):
Really awesome to have this feature, @samueljseay!  Great work on this PR; left a couple quick questions, but testing well for me.

client/header/activity-panel/index.js
samueljseay: @joshuatf Thats a good point, it does merge state on the first level (shallow only), I tend to do this out of habit because it doesn't merge state if you're dealing with nested objects. I can remove this though.

client/header/activity-panel/index.js
samueljseay: @joshuatf yeah thats fair, it would be pretty trivial to make this reusable. Will fix.

client/header/activity-panel/index.js
samueljseay: I'll remove this.

client/header/activity-panel/index.js
samueljseay: Yeah good idea. Will fix.

samueljseay:
@joshuatf @psealock 

thanks for the feedback! I believe I've implemented all your suggested changes.

joshuatf (approved):
Testing really well!  Thanks for the updates here, Sam!  LGTM 🚢

timmyc:
@samueljseay pinged you on the original issue that the design changed a bit here, but let's land this and we can take care of those revisions in a follow-up. Solid work!!!

PR #4855, opened 2020-07-23
Bundle @wordpress/data to avoid fatal exceptions

### LINKED ISSUE #4710: Onboarding: Benefits Page Broken on WordPress 5.3.x
**Describe the bug**
When transitioning from the theme selection step in the OBW to the benefits screen, the benefits screen fails to load and an exception is thrown:

Home ‹ WooCommerce ‹ bend outdoors — WooCommerce 2020-06-26 10-32-27

The value of `getActivePlugins()` on the page is returning `undefined` instead of an array which results in the error above. I attempted to reproduce on a JN site, but it was working fine there. I checked out a few different revisions of wc-admin and determined the bug started happening after #4655 landed ( cc @joshuatf  ).

Perplexed, I did the dance of disabling all plugins except for woo/wca, and then realized that I was running an older version of WP on this particular site - 5.3.1 - compared to 5.4.2 which is on JN.

4655 added in the usage of `__experimentalResolveSelect` which appears to have been shipped in [version 4.10.0 of wp.data](https://github.com/WordPress/gutenberg/blob/release/7.1/packages/data/package.json#L3), and subsequently included in WordPress 5.4.x. So thinking that might the usage of `__experimentalResolveSelect` might be the root of the problem here, but I did not verify further.

I did verify that upgrading my site to 5.4.2 fixed the problem.

**To Reproduce**
Steps to reproduce the behavior, use a site running WordPress 5.3.1

1. Go through the onboarding wizard
2. Once you arrive at the benefits step, you should see the error above
**Additional context**

samueljseay:
Fixes #4710 

There was an issue where we depended on `__experimentalResolveSelect` from `@wordpress/data` but it is not available on some versions of `@wordpress/data` that are bundled with Wordpress. To solve this, we are bundling `@wordpress/data` for now.

To test this you'll need an older version of Wordpress. Wordpress 5.3.2 was used in the original issue so I've used that to replicate it. I suggest using the WP Downgrade plugin to install a 5.3.2 in your development environment.

Also you'll need to make sure that your build assets are clean. run `npm run clean`, then `npm start`

Once you have an older version of Wordpress and your build assets are cleaned out and rebuilt, simply go to the WooCommerce home page and make sure it doesn't crash, do a quick smoke test of the home screen and the setup wizard to ensure nothing causes an exception.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -397,7 +397,6 @@ class Loader {
 			array(
 				'moment',
 				'wp-api-fetch',
-				'wp-data',
 				'wp-data-controls',
 				'wp-element',
 				'wp-hooks',
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -24,7 +24,6 @@ const WC_ADMIN_PHASE = process.env.WC_ADMIN_PHASE || 'development';
 const externals = {
 	'@wordpress/api-fetch': { this: [ 'wp', 'apiFetch' ] },
 	'@wordpress/blocks': { this: [ 'wp', 'blocks' ] },
-	'@wordpress/data': { this: [ 'wp', 'data' ] },
 	'@wordpress/editor': { this: [ 'wp', 'editor' ] },
 	'@wordpress/element': { this: [ 'wp', 'element' ] },
 	'@wordpress/hooks': { this: [ 'wp', 'hooks' ] },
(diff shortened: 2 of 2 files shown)


joshuatf (commented):
Thanks for fixing this one, @samueljseay!

It seems like this should work as is, but I'm getting a bunch of errors on the profiler route `wp-admin/admin.php?page=wc-admin&path=%2Fprofiler`.

Screen Shot 2020-07-23 at 5 22 53 PM

Not quite sure if it has something to do with our packaged version of `@wordpress/data` or if I did something wrong during testing.  Ran `npm run clean` a few times to no avail.

jeffstieler:
I'm experiencing the same errors as @joshuatf

samueljseay:
Ah what a pain, thanks @jeffstieler @joshuatf I'll investigate

samueljseay:
Backing out of this one, bundling wp/data has created some unusual problems.

PR #4856, opened 2020-07-23
Fix PHP warning when preloading scripts on WPEngine

jeffstieler:
Fixes report from an HE:

```
Notice: Undefined index: heartbeat in /wp-content/plugins/woocommerce/packages/woocommerce-admin/src/Loader.php on line 548

Notice: Trying to get property 'handle' of non-object in /wp-content/plugins/woocommerce/packages/woocommerce-admin/src/Loader.php on line 516
```

WPEngine unregisters the `heartbeat` script which causes our "preload tag" logic to throw undefined index notices when trying to find `heartbeat` (a dependency of `wp-auth-check`) in the registered scripts array.

This PR seeks to make the preload logic more resilient by using the [`WP_Dependency::query()`](https://developer.wordpress.org/reference/classes/wp_dependencies/query/) method to access registered scripts. It also updates some terminology to not be racist. (whitelist -> allowlist)

- Simulate missing heartbeat by adding `\wp_deregister_script( 'heartbeat' );` to `Loader::load_scripts()` before the call to `self::output_header_preload_tags()`
- Go to a WC Admin page
- Go to a core WC page
- Verify no errors in log

### Changelog Note:

Fix: PHP notices when hosts block certain WP scripts.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -556,12 +556,19 @@ class Loader {
 		}
 
 		foreach ( $dependencies_of_type->queue as $dependency_handle ) {
-			$dependency = $dependencies_of_type->registered[ $dependency_handle ];
+			$dependency = $dependencies_of_type->query( $dependency_handle, 'registered' );
+
+			if ( false === $dependency ) {
+				continue;
+			}
 
 			// Preload the subdependencies first.
 			foreach ( $dependency->deps as $sub_dependency_handle ) {
-				$sub_dependency = $dependencies_of_type->registered[ $sub_dependency_handle ];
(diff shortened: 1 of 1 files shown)


timmyc:
> It also updates some terminology to not be racist. (whitelist -> allowlist)

<3

octaedro (approved):
You did a great job here @jeffstieler! Code LGTM and is testing well on my end 🚢

timmyc:
cherry picked to `release/1.4.0-beta.2` in 24b8b5a

PR #4857, opened 2020-07-23
Automatically add free local shipping zone on completing the profiler

### LINKED ISSUE #4595: Task list - shipping: smart default
**User story:**
As a new merchant
I want to focus on the most important setup tasks
So I can review some details later

**Description:** with Jetpack installed and connected, alongside WCS, we have the opportunity to set things up on the users' behalf, and ask them to review and confirm, rather than start from scratch and think.

**Acceptance Criteria:**
- [ ] If the store has JP installed and connected, WCS active and selected physical as product type in the profiler, automatically turn on print shipping labels (I think this is already happening) and set the following shipping costs:
Local: free shipping
International: disabled
- [ ] Trigger the following notification:
**Title:** Review your shipping settings
**Content:**  Based on the information that you provided we've configured some shipping rates and options for your store:
Domestic orders: free shipping
International orders: disabled
Label printing: enabled
**Segmentation:** stores with shipping automatically marked as completed
**Timing:** when the task is marked as completed 
**Cadence:** once
**CTA leads to:** /wp-admin/admin.php?page=wc-settings&tab=shipping
**CTA label:** Edit shipping settings

**Event tracking:**
We can use the event `wcadmin_tasklist_task_completed` and eventually add a new value to `task_name` when this happens: shipping_automated

**Additional links:**
[Design mockup](https://www.figma.com/file/JH9XMFUCOjfXdr3N09AHRD/On-boarding-iterations-June-'20?node-id=1%3A9)

becdetat:
Fixes #4595

This adds free local shipping as a shipping option at the end of the onboarding wizard if:

- "Physical" was selected as a product type
- No existing shipping zones (except for "Locations not covered by your other zones", which is a system shipping zone that can't be deleted)

1. Delete the profile data option - `wp option delete woocommerce_onboarding_profile` or `delete from wp_options where option_name = 'wp option delete woocommerce_onboarding_profile'`
1. Delete all your shipping zones (WooCommerce -> Settings -> Shipping) except for "Locations not covered by your other zones", which is a system shipping zone that can't be deleted
2. Start the OBW
3. Complete the OBW
4. Go to the WooCommerce home page
5. Check that there is an admin note titled "Review your shipping settings"
6. Check that the "Set up shipping" task is completed
7. Check that the free local shipping zone has been added

### Changelog Note:

Enhancement: Automatically add free local shipping zone on completing the profiler

### DIFF
--- a/src/Features/Onboarding.php
+++ b/src/Features/Onboarding.php
@@ -11,6 +11,7 @@ namespace Automattic\WooCommerce\Admin\Features;
 use \Automattic\WooCommerce\Admin\Loader;
 use \Automattic\WooCommerce\Admin\Notes\WC_Admin_Notes_Onboarding_Profiler;
 use \Automattic\WooCommerce\Admin\PluginsHelper;
+use \Automattic\WooCommerce\Admin\Features\OnboardingSetUpShipping;
 
 /**
  * Contains backend logic for the onboarding profile and checklist feature.
@@ -78,6 +79,9 @@ class Onboarding {
 		// Add actions and filters.
 		$this->add_actions();
 		$this->add_filters();
+
+		// Hook up dependant classes.
+		new OnboardingSetUpShipping();
 	}
 
 	/**
@@ -105,6 +109,15 @@ class Onboarding {
 		// Rest API hooks need to run before is_admin() checks.
 		add_action( 'woocommerce_theme_installed', array( $this, 'delete_themes_transient' ) );
 		add_action( 'after_switch_theme', array( $this, 'delete_themes_transient' ) );
+		add_action(
+			'update_option_' . self::PROFILE_DATA_OPTION,
+			array(
+				$this,
+				'trigger_profile_completed_action',
+			),
+			10,
+			2
+		);
 
 		if ( ! is_admin() ) {
 			return;
@@ -118,6 +131,20 @@ class Onboarding {
 		add_action( 'current_screen', array( $this, 'calypso_tests' ) );
 		add_action( 'current_screen', array( $this, 'redirect_wccom_install' ) );
 		add_action( 'current_screen', array( $this, 'redirect_old_onboarding' ) );
+	}
+
+	/**
+	 * Trigger the woocommerce_onboarding_profile_completed action
+	 *
+	 * @param array $old_value Previous value.
+	 * @param array $value Current value.
+	 */
+	public function trigger_profile_completed_action( $old_value, $value ) {
+		if ( ! isset( $value['completed'] ) || ! $value['completed'] ) {
+			return;
+		}
+
+		do_action( 'woocommerce_onboarding_profile_completed' );
 	}
 
 	/**
--- a/src/Features/Onboarding.php
+++ b/src/Features/Onboarding.php
@@ -144,6 +144,12 @@ class Onboarding {
 			return;
 		}
 
+		/**
+		 * Action hook fired when the onboarding profile (or onboarding wizard,
+		 * or profiler) is completed.
+		 *
+		 * @since 1.5.0
+		 */
 		do_action( 'woocommerce_onboarding_profile_completed' );
 	}
 
--- a/src/Features/Onboarding.php
+++ b/src/Features/Onboarding.php
@@ -140,6 +140,10 @@ class Onboarding {
 	 * @param array $value Current value.
 	 */
 	public function trigger_profile_completed_action( $old_value, $value ) {
+		if ( isset( $old_value['completed'] ) && $old_value['completed'] ) {
+			return;
+		}
+
 		if ( ! isset( $value['completed'] ) || ! $value['completed'] ) {
 			return;
 		}
--- a/src/Features/OnboardingSetUpShipping.php
+++ b/src/Features/OnboardingSetUpShipping.php
@@ -28,5 +28,37 @@ class OnboardingSetUpShipping {
 	 * Set up shipping.
 	 */
 	public static function on_onboarding_profile_completed() {
+		// Only do this if there are no existing shipping methods.
+
+		self::turn_on_printing_shipping_labels();
+		self::set_up_free_local_shipping();
+		self::disable_international_shipping();
+		self::trigger_review_shipping_settings_note();
+		self::track_shipping_automatically_set_up_event();
 	}
+
+	/**
+	 * Turn on printing of shipping labels.
+	 */
+	private static function turn_on_printing_shipping_labels() {}
+
+	/**
+	 * Set up free local shipping.
+	 */
+	private static function set_up_free_local_shipping() {}
+
+	/**
+	 * Disable international shipping.
+	 */
+	private static function disable_international_shipping() {}
+
+	/**
+	 * Trigger the "Review your shipping settings" admin note.
+	 */
+	private static function trigger_review_shipping_settings_note() {}
+
+	/**
+	 * Track the shipping_automatically_set_up event.
+	 */
+	private static function track_shipping_automatically_set_up_event() {}
 }
--- a/src/Features/OnboardingSetUpShipping.php
+++ b/src/Features/OnboardingSetUpShipping.php
@@ -31,19 +31,11 @@ class OnboardingSetUpShipping {
 	 * Set up shipping.
 	 */
 	public static function on_onboarding_profile_completed() {
-		if ( ! self::is_jetpack_installed_and_connected() ) {
-			return;
-		}
-
-		if ( ! PluginsHelper::is_plugin_active( 'woocommerce-services' ) ) {
-			return;
-		}
-
 		if ( ! self::is_physical_selected_as_product_type() ) {
 			return;
 		}
 
-		if ( ! self::are_there_existing_shipping_zones() ) {
+		if ( self::are_there_existing_shipping_zones() ) {
 			return;
 		}
 
@@ -80,21 +72,6 @@ class OnboardingSetUpShipping {
 		}
 
 		return true;
-	}
-
-	/**
-	 * Is Jetpack installed and connected?
-	 *
-	 * @return bool
-	 */
-	private static function is_jetpack_installed_and_connected() {
-		if ( ! class_exists( '\Jetpack_Data' ) ) {
-			return false;
-		}
-
-		$user_token = \Jetpack_Data::get_access_token( JETPACK_MASTER_USER );
-
-		return isset( $user_token->external_user_id );
 	}
 
 	/**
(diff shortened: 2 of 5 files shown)


src/Features/OnboardingSetUpShipping.php
```diff
@@ -0,0 +1,108 @@
...
+
+namespace Automattic\WooCommerce\Admin\Features;
+
+use \Automattic\WooCommerce\Admin\PluginsHelper;
+use \Automattic\WooCommerce\Admin\Notes\WC_Admin_Notes_Review_Shipping_Settings;
+
+/**
+ * This contains logic for setting up shipping when the profiler completes.
+ */
+class OnboardingSetUpShipping {
```
timmyc, line 16: @becdetat what do you think about adding some test coverage for this new class?

becdetat: @timmyc I did consider that, but IMO this class doesn't actually contain a lot of logic and the tests would largely be testing truisms - such as "does `explode( ':', $default_country )[0]` return the first part of `$default_country`".

src/Features/OnboardingSetUpShipping.php
```diff
@@ -0,0 +1,108 @@
...
+		WC_Admin_Notes_Review_Shipping_Settings::possibly_add_note();
+		wc_admin_record_tracks_event( 'shipping_automatically_set_up' );
+	}
+
+	/**
+	 * Are there existing shipping zones?
+	 *
+	 * @return bool
+	 */
+	private static function are_there_existing_shipping_zones() {
```
joshuatf, line 52: Nitpick: can we rename this to `has_existing_shipping_zones`?

src/Features/OnboardingSetUpShipping.php
```diff
@@ -0,0 +1,108 @@
...
+
+		return $zone_count > 0;
+	}
+
+	/**
+	 * Is 'physical' selected as a product type?
+	 *
+	 * @return bool
+	 */
+	private static function is_physical_selected_as_product_type() {
```
joshuatf, line 63: Nitpick: can we rename this to `has_physical_products`?

src/Features/OnboardingSetUpShipping.php
```diff
@@ -0,0 +1,108 @@
...
+		if ( ! $default_country ) {
+			return;
+		}
+
+		$country_code = explode( ':', $default_country )[0];
+		$zone         = new \WC_Shipping_Zone();
+
+		$zone->add_location( $country_code, 'country' );
+
+		$countries = apply_filters(
```
joshuatf, line 95: Could we use `WC_Countries` `get_countries()` method to get this list?

src/Notes/WC_Admin_Notes_Review_Shipping_Settings.php
```diff
@@ -0,0 +1,67 @@
...
+ */
+class WC_Admin_Notes_Review_Shipping_Settings {
+	use NoteTraits;
+
+	const NOTE_NAME = 'wc-admin-review-shipping-settings';
+
+	/**
+	 * Possibly add note.
+	 */
+	public static function possibly_add_note() {
```
joshuatf, line 25: This looks to be the same as the note traits method.  Can we remove this method or does it serve a different purpose?

src/Features/OnboardingSetUpShipping.php
```diff
@@ -0,0 +1,108 @@
...
+			'woocommerce_countries',
+			include WC()->plugin_path() . '/i18n/countries.php'
+		);
+		$zone_name = isset( $countries[ $country_code ] )
+			? $countries[ $country_code ]
+			: null;
```
joshuatf, line 101: Should we set to the zone code here if for some reason this country does not exist in the list?

src/Features/OnboardingSetUpShipping.php
joshuatf: I do agree that a little test coverage around `set_up_free_local_shipping` might be useful just to check if the corect zones were added, but if you think it's not very useful, Bec, then we can skip.

src/Features/Onboarding.php
```diff
@@ -120,6 +133,26 @@ private function add_actions() {
...
+			return;
+		}
+
+		/**
+		 * Action hook fired when the onboarding profile (or onboarding wizard,
+		 * or profiler) is completed.
+		 *
+		 * @since 1.5.0
+		 */
+		do_action( 'woocommerce_onboarding_profile_completed' );
```
joshuatf, line 153: This is a really nice addition!  I can see a couple areas where it would be useful to replace with this action instead of manually checking the profile data.  We might want to create a follow-up issue to update those.

src/Features/Onboarding.php
```diff
@@ -120,6 +133,26 @@ private function add_actions() {
...
 	}
 
+	/**
+	 * Trigger the woocommerce_onboarding_profile_completed action
+	 *
+	 * @param array $old_value Previous value.
+	 * @param array $value Current value.
+	 */
+	public function trigger_profile_completed_action( $old_value, $value ) {
+		if ( ! isset( $value['completed'] ) || ! $value['completed'] ) {
```
joshuatf, line 143: Should we also check the old value here to make sure it was not complete so we don't trigger this action on updating other properties if the profiler was already complete?

joshuatf (commented):
Awesome work on this, Bec!  I really like the addition of the profiler completed action.

Left a few comments/questions in the code.

src/Features/OnboardingSetUpShipping.php
becdetat: Ok I've bowed under the pressure and added a couple of unit tests :-D in 752ed29

src/Features/OnboardingSetUpShipping.php
becdetat: done

src/Features/OnboardingSetUpShipping.php
becdetat: done

src/Features/OnboardingSetUpShipping.php
becdetat: done

src/Notes/WC_Admin_Notes_Review_Shipping_Settings.php
becdetat: Not sure how that happened :-$ fixed

src/Features/OnboardingSetUpShipping.php
becdetat: The location code is used in the zone locations (https://github.com/woocommerce/woocommerce-admin/pull/4857/files/6b20f85297ab6eacad1137746c1e017ff04b8faf#diff-30f4b46e510aacb5afc83382ee23b2aaR93)

src/Features/Onboarding.php
becdetat: done

src/Features/Onboarding.php
becdetat: I've created a follow-up issue: https://github.com/woocommerce/woocommerce-admin/issues/4931

src/Features/Onboarding.php
```diff
@@ -105,6 +109,15 @@ private function add_actions() {
 		// Rest API hooks need to run before is_admin() checks.
 		add_action( 'woocommerce_theme_installed', array( $this, 'delete_themes_transient' ) );
 		add_action( 'after_switch_theme', array( $this, 'delete_themes_transient' ) );
+		add_action(
```
timmyc, line 112: Minor but when specifying a different priority for a hook to operate, it is nice to add a comment explaining the reasoning/background for that decision. Just helps with future us debugging reasoning for certain actions happening before/after others.

timmyc (commented):
Hey @becdetat took this for a spin in its current state and it is testing well. I see you responded to Josh's feedback of things being done but looks like you might not have pushed up the changes yet here to GitHub. I did point out one other little minor item, but would be happy to give this one final test once the changes you described have been pushed to the remote branch here. Looking really great!!

becdetat:
@timmyc 

> looks like you might not have pushed up the changes yet here to GitHub

Whoops! The changes have been pushed now. This should be ok to test again. Sorry about that!

src/Features/Onboarding.php
becdetat: The default priority is 10 which is what is being used here. The only reason the priority is set here is to change the number of arguments to 2, otherwise the hook doesn't work properly.

src/Features/OnboardingSetUpShipping.php
joshuatf: I mean in case the country doesn't exist in the list for some reason, also use the zone code for the name so it's not empty.  But maybe an empty name is more appropriate here if that were to happen.

joshuatf (commented):
I'm still seeing the old changes here, @becdetat.  Could you try to push one more time?  Thanks!

src/Features/Onboarding.php
timmyc: Ah that is right, don't mind me.

becdetat:
@joshuatf 

> I mean in case the country doesn't exist in the list for some reason, also use the zone code for the name so it's not empty. But maybe an empty name is more appropriate here if that were to happen.

That makes sense, I think setting the name to the zone code is a good idea.

I think you might not have pulled in the latest changes - they were definitely there. I've rebased the branch now so you'll have to reset to the origin ;-)

src/Notes/WC_Admin_Notes_Review_Shipping_Settings.php
```diff
@@ -0,0 +1,55 @@
...
+				'woocommerce-admin'
+			)
+		);
+		$note->set_content_data( (object) array() );
+		$note->set_type( WC_Admin_Note::E_WC_ADMIN_NOTE_INFORMATIONAL );
+
+		$note->add_action(
+			'edit-shipping-settings',
+			__( 'Edit shipping settings', 'woocommerce-admin' ),
+			'/wp-admin/admin.php?page=wc-settings&tab=shipping',
```
joshuatf, line 48: This takes me to `wp-admin/wp-admin/admin.php?page=wc-settings&tab=shipping` when I click on it.

I think we should update to the full URL `admin_url( 'admin.php?page=wc-settings&tab=shipping' )`.

joshuatf (approved):
>I think you might not have pulled in the latest changes - they were definitely there. I've rebased the branch now so you'll have to reset to the origin ;-)

Oh this was on Github, but maybe there was some browser caching going on as I saw the old naming `is_physical...` but can see the updated code now.

Testing well and LGTM pending one small change!

src/Notes/WC_Admin_Notes_Review_Shipping_Settings.php
becdetat: done, thanks

PR #4862, opened 2020-07-27
Add the experimental resolver to WCA data package

### LINKED ISSUE #4710: Onboarding: Benefits Page Broken on WordPress 5.3.x

joshuatf:
Fixes #4710 

This PR adds in the experimental resolution selector to WCA so that we can await the resolution of the Jetpack connection URL.

This selector is no longer being labeled "experimental" and appears safe to use, however, going forward we're going to have the same issue with this selector not being readily available in older versions of WP where we pull in the WP hosted version of `@wordpress/data`.  This should be considered a temporary fix and we may want to consider fixing the bundling of `@wordpress/data` in the next cycle.

1. Install an older version of WordPress (e.g., 5.3.2) or use WP Downgrade.
1. Check that the profiler and homescreen feature are able to load.
1. Make sure that Jetpack is disconnected and attempt to walk through the profiler to the benefits screen.
1. Make sure that you are redirected to the Jetpack connection screen (or just Jetpack page if you are testing locally).
1. Check that the Install Jetpack CTA on the homescreen also works and redirects to the Jetpack connection screen.

### DIFF
--- a/client/homescreen/stats-overview/install-jetpack-cta.js
+++ b/client/homescreen/stats-overview/install-jetpack-cta.js
@@ -6,18 +6,18 @@ import { compose } from '@wordpress/compose';
 import { Button } from '@wordpress/components';
 import { useState } from 'react';
 import PropTypes from 'prop-types';
-import {
-	withDispatch,
-	withSelect,
-	__experimentalResolveSelect,
-} from '@wordpress/data';
+import { withDispatch, withSelect } from '@wordpress/data';
 
 /**
  * WooCommerce dependencies
  */
+import {
+	__experimentalResolveSelect,
+	PLUGINS_STORE_NAME,
+	useUserPreferences,
+} from '@woocommerce/data';
 import { H } from '@woocommerce/components';
 import { getAdminLink } from '@woocommerce/wc-admin-settings';
-import { PLUGINS_STORE_NAME, useUserPreferences } from '@woocommerce/data';
 
 /**
  * Internal dependencies
--- a/client/profile-wizard/index.js
+++ b/client/profile-wizard/index.js
@@ -5,7 +5,7 @@ import { __ } from '@wordpress/i18n';
 import { Component, createElement, Fragment } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import { identity, pick } from 'lodash';
-import { withDispatch, __experimentalResolveSelect } from '@wordpress/data';
+import { withDispatch } from '@wordpress/data';
 
 /**
  * WooCommerce dependencies
@@ -17,6 +17,7 @@ import {
 	updateQueryString,
 } from '@woocommerce/navigation';
 import {
+	__experimentalResolveSelect,
 	ONBOARDING_STORE_NAME,
 	OPTIONS_STORE_NAME,
 	PLUGINS_STORE_NAME,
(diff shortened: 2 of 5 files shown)


jeffstieler (commented):
I think this approach is fine to fix our use of an experimental `@wordpress/data` selector. It would be great to note in a comment that this code was directly pulled from: https://github.com/WordPress/gutenberg/blob/909c9274b2440de5f6049ffddfcc8e0e6158df2d/packages/data/src/registry.js#L91-L130

Does this polyfill (for lack of better term) need to become part of our published `@woocommerce/data` package? I'm not sure we want other developers using this.

timmyc:
> Does this polyfill (for lack of better term) need to become part of our published @woocommerce/data package? I'm not sure we want other developers using this.

I think that is a valid concern, but I also can see it being a logical place for this to live for the time being. With Woo's support of WP versions being set to current release - 2 ( currently at 5.2 ), we will be needing this fix until 5.6 ships which appears that is slated for [December 2020](https://wordpress.org/about/roadmap/).

If we do keep it in the package, I think we just need to log an issue to remove it once 5.6 ships, and Woo core requires at least 5.4. We would then do a breaking release to the data package, and remove our "polyfill"

octaedro (commented):
Good job @joshuatf! I tested this code on my end and on an ephemeral site (with WordPress 5.2.1 in both cases) and it's working well.
I think that after handling Jeff's comments this PR will be ready 🙂

joshuatf:
Thanks for the input, everyone.  Added a comment to note this function's origin and that it will be deprecated.

>I think that is a valid concern, but I also can see it being a logical place for this to live for the time being.

After looking to find a better place for this to live, I also feel this may be the best place if we don't mind introducing a breaking version.  I also don't feel strongly on this, so if you're against this @jeffstieler just say the word and we can move it into a utils file.

Issue preemptively noted here - https://github.com/woocommerce/woocommerce-admin/issues/4863

timmyc:
Honestly if we *know* we are going to be removing this, having it live in the npm package is probably the safest place for it to land since it is an external dependency to the project and kind of maps to the flow of how WP Core ships things in packages/and deprecates.

samueljseay (approved):
Looking great @joshuatf thank you for solving this!

psealock (approved):
LGTM, tests well on 5.4 as well as 5.3

PR #4866, opened 2020-07-28
Update wp-prettier to latest beta release to avoid a crash in Prettier

samueljseay:
In the previous release Prettier did not work in VSCode (and I
suspect would fail in other editors too) with an error:

"Could not find module ./parser-babylon". Updating to the latest
version fixes it in my environment.

This bug interrupts a fairly crucial workflow for me, so I'd be keen to
get it fixed/merged.

Testing details:

* Ensure you have done a fresh npm install
* Edit a JS file (ideally in VSCode with a prettier plugin if you can)
* Save it (if your prettier plugin has format on save)

You should not see errors from the plugin (in VSCode the plugin reports
status in the bottom bar) and it should format the file correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -201,7 +201,7 @@
 		"node-watch": "0.6.4",
 		"postcss-color-function": "4.1.0",
 		"postcss-loader": "3.0.0",
-		"prettier": "npm:wp-prettier@2.0.5",
+		"prettier": "npm:wp-prettier@2.0.5-beta-1",
 		"promptly": "3.0.3",
 		"prop-types": "15.7.2",
 		"raw-loader": "4.0.1",
--- a/packages/components/src/search-list-control/index.js
+++ b/packages/components/src/search-list-control/index.js
@@ -109,7 +109,7 @@ export class SearchListControl extends Component {
 		const re = new RegExp( escapeRegExp( search ), 'i' );
 		this.props.debouncedSpeak( messages.updated );
 		const filteredList = list
-			.map( ( item ) => ( re.test( item.name ) ? item : false ) )
+			.map( ( item ) => ( re.test( item.name ) ? item : false)  )
 			.filter( Boolean );
 		return isHierarchical
 			? buildTermsTree( filteredList, list )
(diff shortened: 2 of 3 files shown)


packages/components/src/search-list-control/index.js
```diff
@@ -109,7 +109,7 @@ export class SearchListControl extends Component {
 		const re = new RegExp( escapeRegExp( search ), 'i' );
 		this.props.debouncedSpeak( messages.updated );
 		const filteredList = list
-			.map( ( item ) => ( re.test( item.name ) ? item : false ) )
+			.map( ( item ) => ( re.test( item.name ) ? item : false)  )
```
samueljseay, line 112: Oh hmm that actually looks wrong, huh. I guess there are still some teething issues with the new lint setup

package.json
```diff
@@ -201,7 +201,7 @@
 		"node-watch": "0.6.4",
 		"postcss-color-function": "4.1.0",
 		"postcss-loader": "3.0.0",
-		"prettier": "npm:wp-prettier@2.0.5",
+		"prettier": "npm:wp-prettier@2.0.5-beta-1",
```
psealock, line 204: Is this not a downgrade?

psealock (approved):
This is working well in my VSCode instance

packages/components/src/search-list-control/index.js
psealock: I made a comment in this file, saved it and this issue fixed itself 🤷🏻‍♀️. 

I will go ahead and commit that fix.

package.json
samueljseay: @psealock 🤦 you're right.

Also, I noticed your fix still fails on CI for the error, so maybe our CI env is a little different. 😧

samueljseay: So my confusion around beta-1 being an update was that its the latest version tag by over a month in that repo. But no idea if this is a mistake in the tag or something else.

psealock:
FWIW on `main` the testing instructions work for me using VSCode. Do you have this setting enabled along with the Prettier plugin?

Screen Shot 2020-07-30 at 1 56 34 PM

samueljseay:
@psealock Yeah I'm not seeing this issue any more on `main` 🤷 . Closing this.

PR #4869, opened 2020-07-28
Polyfill core-data saveUser() on WP 5.3.x.

jeffstieler:
Fixes missing `saveUser` action in WP `5.3.x`.

This PR seeks to "polyfill" the `saveUser()` action from `@wordpress/core-data` that is absent in WP `5.3.x`. It does so with minimal code by leveraging the `saveEntityRecord()` action to do the heavy lifting.

In both WP 5.3 and 5.4:
- Verify that showing/hiding stats on the home screen works (and persists after a refresh)
- Verify analytics report column toggling works (and persists after a refresh)
- Verify the OBW can be completed (actually work through the steps) without error
- Verify no errors in console

### Changelog Note:

Fix: bug preventing saving user preferences on WP 5.3.x.

### DIFF
--- a/packages/data/src/user-preferences/test/use-user-preferences.js
+++ b/packages/data/src/user-preferences/test/use-user-preferences.js
@@ -213,4 +213,95 @@ describe( 'useUserPreferences() hook', () => {
 			} );
 		} );
 	} );
+
+	it( 'Polyfills saveUser() on older versions of WordPress', async () => {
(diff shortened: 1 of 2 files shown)


octaedro (approved):
Great job @jeffstieler! Thank you for this fix.
Code looks really good and it's testing well on the mentioned versions :shipit:

PR #4874, opened 2020-07-29
Declare packages dependencies independently of main package.json

psealock:
In preparation to publish packages I saw that we were incorrectly specifying dependencies only found in packages twice. This PR uses a format like `"@woocommerce/csv-export": "file:packages/csv-export",` to tell npm to look at each package's package.json for its own dependencies. Dependencies only used in the packages can be then removed from the main package.json.

This is similar to [how Gutenberg does it](https://github.com/WordPress/gutenberg/blob/a260cda069b1899399795062935f6e53555cf855/package.json#L25-L30).

1. `npm start`
2. Ensure no errors.
3. Poke around the app. If any dependency is missing, you will encounter a fatal error.

### Changelog Note:

- dev: Specify package dependencies in packages themselves.


octaedro (approved):
Good job @psealock, this tests well here and everything looks good. There is a `Travis` check that is failing but the rest looks great :shipit:

PR #4891, opened 2020-07-30
Currency: re-add deprecated formatCurrency

### LINKED ISSUE #4860: formatCurrency is not a function error on WooCommerce Payments pages
**Describe the bug**
Opening a WooCommerce Payments page (let's say the Transactions page) with a version of WC Admin after 227898e712e9e3fc2fc06bd6409e782e7a695f97 (Currency package: make naming clearer) causes the page to not load and `Uncaught TypeError: ra.formatCurrency is not a function` in console.

**To Reproduce**

1. Site with WooCommerce Payments setup
2. Open Payments > Transactions from the nav
3. See error

The error is due to the `formatCurrency` function being renamed to `formatAmount`. I appreciate this is all about how we manage dependences across plugins, and this is more of a request for advise than an issue report.

The immediate problem is that on upgrading to a version of WooCommerce with this change in, a site with the current version of WooCommerce Payments will be unable to use the Payments screens.

That aside, I see a few options:

1. Keep the new version of the Currency package backwards compatible by adding `formatCurrency` back in as an alias for `formatAmount`. The obvious downside here is clutter in your code base.
2. WooCommerce Payments bundles a known version of `@woocommerce/currency`. We've been trying to avoid doing this due to package sizes concerns, but it is an option. It also doesn't help with the immediate problem because current WCPay versions use whatever WC Admin makes available.
3. WooCommerce Payments updates calls to `formatCurrency`, which fixes things for new installs but leaves a problem with supporting mixed versions of WooCommerce.

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/4860

`woocommerce/currency`'s `formatCurrency` was changed to `formatAmount` in version 1.3.0. This causes problems for plugins using the package as part of Core.

This PR adds the deprecated function back in with a warning.

1. Go to a file that uses `formatAmount` and change it to `formatCurrency`. ie: 

https://github.com/woocommerce/woocommerce-admin/blob/330677bc72b46a38121f2b091e6e26c571b5671e/client/analytics/components/report-chart/index.js#L173

2. Go to the page in question. If using the above example, visit any report.
3. See the deprecation warning in the console.

### Changelog Note:

- dev: Re-introduce `@woocommerce/currency` function `formatCurrency` with deprecation warning.

### DIFF
--- a/packages/currency/src/index.js
+++ b/packages/currency/src/index.js
@@ -59,8 +59,9 @@ const CurrencyFactory = ( currencySetting ) => {
 	 */
 	function formatCurrency( number ) {
 		deprecated( 'Currency().formatCurrency', {
-			version: '1.3.0',
+			version: '4.3.0',
 			alternative: 'Currency().formatAmount',
+			plugin: 'WooCommerce',
 			hint: '`formatAmount` accepts the same arguments as formatCurrency',
 		} );
 		return formatAmount( number );
--- a/packages/currency/src/index.js
+++ b/packages/currency/src/index.js
@@ -3,6 +3,7 @@
  */
 import { sprintf } from '@wordpress/i18n';
 import { numberFormat } from '@woocommerce/number';
+import deprecated from '@wordpress/deprecated';
 
 const CurrencyFactory = ( currencySetting ) => {
 	let currency;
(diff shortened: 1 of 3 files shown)


packages/currency/src/index.js
```diff
@@ -48,6 +49,24 @@ const CurrencyFactory = ( currencySetting ) => {
...
+	 * Formats money value.
+	 *
+	 * @deprecated
+	 *
+	 * @param   {number|string} number number to format
+	 * @return {?string} A formatted string.
+	 */
+	function formatCurrency( number ) {
+		deprecated( 'Currency().formatCurrency', {
+			version: '4.3.0',
```
jeffstieler, line 62: When are we planning to fully _remove_ `formatCurrency()`? This produces the message:

> Currency().formatCurrency is deprecated and will be removed from WooCommerce in version 4.3.0

Considering that's a past version, the message is confusing.

jeffstieler (approved):
This looks good and tests well. Just curious if the version should be different for the removal?

packages/currency/src/index.js
timmyc: Going off of the [Woo Core Docs](https://github.com/woocommerce/woocommerce/wiki/Deprecation-in-core) we don't have a real guideline about choosing which version a method will be finally removed in... to play it safe, I feel like we should say it will be removed in the next breaking release ( 5.0.0 ) - and we can standardize on that then easily search the code base for said deprecation notices prior to that release and cleanup the old methods then. Does that sound reasonable to you all?

psealock: Good call team, thanks!

PR #4899, opened 2020-07-30
Only show WCPay task in US based stores

### LINKED ISSUE #4885: [1.4] Make WCPay Feature in Setup Task List More Targeted
A follow-up item to #4793 - currently if a site has WooCommerce Payments installed ( and even not activated ) we now show the Task List option to Setup/Configure WooCommerce Payments in-lieu of the standard payment step. I propose that we make the logic a bit more _strict_ on this to also verify the site is in a supported country ( currently US only ).

Context for this proposal: We are going to eventually begin bundling WCPay with ecommerce plan installs, and as such, users in non-wcpay supported countries would have a bit of a rough UX experience with this. Also I suppose we could have an edge case where a user installs WCPay individually, and would see the task list option, even if their country is not supported.

joshuatf:
Fixes #4885 

Hides the wcpay task if not in a US based store.

### Screenshots
Screen Shot 2020-07-30 at 3 37 58 PM

1. Make sure the `wcpay` feature is enabled in your environment and wc pay is installed.
1. Set your store settings to a US based store.
1. View the task list and make sure the WC pay task exists.
1. Set your store to non-US.
1. Check that the task is not shown in the task list.

### DIFF
--- a/client/task-list/index.js
+++ b/client/task-list/index.js
@@ -26,6 +26,7 @@ import {
 	PLUGINS_STORE_NAME,
 	OPTIONS_STORE_NAME,
 	ONBOARDING_STORE_NAME,
+	SETTINGS_STORE_NAME,
 } from '@woocommerce/data';
 
 /**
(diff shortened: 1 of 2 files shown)


octaedro (approved):
You did a good job here @joshuatf, code LGTM and it's testing well here 🚀

timmyc:
Thanks Josh!

psealock:
@joshuatf I went ahead and cherry picked this one into `release/1.4.0-beta.3`. Let me know if there's any issue with that.

PR #4900, opened 2020-07-30
OBW: fix product types step bugs

### LINKED ISSUE #4876: Store Profiler - Product Types step: missing tooltip and “Physical products” checkbox on by default
**Describe the bug**
According to #4577, the “Physical products” checkbox should be toggled on by default and we should display a tooltip with more info about the purchase when they hover the pricing.

**To Reproduce**

1. Install 4.4 beta version
2. Start the new onboarding
3. Go to the Product types step and see the “Physical products” checkbox untoggled
4. Hover a pricing pill and see that we don't display the tooltip

**Screenshots**
https://www.loom.com/share/a90ab70b8c2145048738bd6a3b653798

* OS: macOS Movaje
* Browser Chrome v 84.0.4147.89

jeffstieler:
Fixes #4876.

This PR fixes some small bugs on the Product Types step of the OBW. It restores "physical products" as the default, and also fixes the display of the tooltips (on the price description).

### Accessibility

### Screenshots

Screen Shot 2020-07-30 at 10 40 18 AM

- Clear profiler data or start a new test store
- Go through the OBW
- Verify on the Product Types step that "Physical products" is checked by default
- Verify that the tooltip works when hovering over the price info

### DIFF
--- a/client/profile-wizard/steps/product-types.js
+++ b/client/profile-wizard/steps/product-types.js
@@ -32,19 +32,22 @@ function getLabel( description, yearlyPrice ) {
 		__( '$%f per month, billed annually', 'woocommerce-admin' ),
 		monthlyPrice
 	);
+	/* eslint-disable @wordpress/i18n-no-collapsible-whitespace */
+	const toolTipText = __(
+		"This product type requires a paid extension.\nWe'll add this to a cart so that\nyou can purchase and install it later.",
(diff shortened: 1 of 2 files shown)


octaedro (approved):
Good job @jeffstieler, this tests well here and code looks great 🚢

PR #4909, opened 2020-07-30
Onboarding: Change account required text on biz step.

### LINKED ISSUE #4881: [WC 4.4 beta1] OBW - Tell us about your business: make more clear what account is required to install recommended features

During OBW on the `Tell us about your business` page, it is not clear what account is required to install recommended features:

account

I assume that the WooCommerce account is required but it is not immediately clear and may not be clear to the new user. It will be great to specify what account is required such as `Requires a WooCommerce account`. 

Another thing that arises here is that even if there is a note regarding the type of account required, there is no indication if the user already has an account. In other words, it is hard to know what to do with this information. Do I need to create an account now? How can I do that? Do I need to exit OBW now to do that? If I exit, how do I go back? If I keep going with OBW, how can I make sure I get all these features later if I don't have an account? 

Would be nice to have a note on the `Tell us about your business` page that answers the questions above. 

N/A.

IMO, it should be prioritized in order for the customers to have a clear understanding of the options they have at that exact moment they are on the `Tell us about your business` page based on whether or not they have an account. 

**Additional context**

N/A.

timmyc:
Fixes #4881

This branch adjusts the text that is shown when users are going to install the recommended bundled extensions.

__Before__

Profiler ‹ WooCommerce ‹ bend outdoors — WooCommerce 2020-07-30 16-15-04

__After__

Profiler ‹ WooCommerce ‹ bend outdoors — WooCommerce 2020-07-30 16-10-26

- Enable the onboarding wizard via Orders > Help
- Enter a US address on the first step, and select fashion/apparel as an industry
- On the biz details step, note the text shown in the footer has changed per the decision made in #4881 

### Changelog Note:

NA unreleased feature

### DIFF
--- a/client/profile-wizard/steps/business-details.js
+++ b/client/profile-wizard/steps/business-details.js
@@ -356,28 +356,26 @@ class BusinessDetails extends Component {
 				</p>
 			);
 		}
-
+		const accountRequiredText = this.bundleInstall
+			? __(
+					'User accounts are required to use these features.',
+					'woocommerce-admin'
+			  )
+			: '';
 		return (
 			<Fragment>
 				<p>
 					{ sprintf(
 						_n(
-							'The following plugin will be installed for free: %s',
-							'The following plugins will be installed for free: %s',
+							'The following plugin will be installed for free: %s. %s',
+							'The following plugins will be installed for free: %s. %s',
 							extensions.length,
 							'woocommerce-admin'
 						),
-						extensionsList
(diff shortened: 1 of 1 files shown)


psealock (approved):
Changes LGTM 🚢

timmyc:
Cherry picked to `release\1.4.0-beta.3` in 70fc4cbe4afe8127a911251e76abad54bea14576

PR #4920, opened 2020-08-03
Fix use of Jetpack autoloader.

### LINKED ISSUE #4910: Jetpack Autloader Updates Needed
Adding a quick issue to the 1.4 milestone to make sure we address the feedback from @ObliviousHarmony's latest comment on p7bje6-2ec-p2

jeffstieler:
Fixes #4910.

This PR seeks to fix our project's usage of the Jetpack autoloader so that the latest version of the package gets included. It also introduces a check to see if the necessary Jetpack constant is defined for local development.

Since the `composer.json` now receives a version for builds, this PR also makes a change to the `prestart` and `prebuild` scripts so they don't bump versions, and introduces a new `bump-version` script that is used by `pre-release`.

Note: We can probably remove the `WC_Admin_Notes_Deactivate_Plugin` class now, but I chose not to do so at this time.

### Testing missing development constant notice

_This also tests the change to use the correct Jetpack autoloader file._

This will require a bit additional setup:

1. Make sure you are running WooCommerce 4.4 beta 1 (or checked out the built release branch for that version - it must be that version, not beta2).
2. Make sure you are running the latest WP 5.5 release candidate.
3. Without doing anything in this branch, you should a notice instructing you to set `JETPACK_AUTOLOAD_DEV`
4. Add `define( 'JETPACK_AUTOLOAD_DEV', true )` to your `wp-config.php` file (you'll want to leave it defined there when developing going forward). **Note:** You _may_ get a fatal error because your checked out version of Woo Core has the constant defined, in that case just comment out the constant definition in Woo Core (Woo core 4.4. will not be released with the defined constant as it's being removed).
5. Verify that the notice disappears in your wp-admin dashboard. 

### Testing version bump changes.

To verify things work as expected, you can run this script locally:

```bash
npm run pre-release
```

Base the test release on the current branch (`fix/4910-autoloader-version`), pick any valid version (e.g. `5.5.5-beta.1`)

After running that you should note the following:

- `composer.json`, `package.json`, `package-lock.json`, `readme.txt`, `src/Package.php`, `src/FeaturePlugin.php`, `woocommerce-admin.php` should be changed.
- The above files should only have a version bump to the specified version
- `composer.json`, should have the `version` property added by the script (otherwise it exists without a `version` property in development).

### Changelog Note:

Fix: defer to WooCommerce Admin version in WooCommerce core if it is newer.

### DIFF
--- a/bin/update-version.php
+++ b/bin/update-version.php
@@ -25,6 +25,9 @@ function replace_version( $filename, $package_json ) {
 		if ( stripos( $line, 'Stable tag: ' ) !== false ) {
 			$line = "Stable tag: {$package_json->version}\n";
 		}
+		if ( stripos( $line, '"name": "woocommerce/woocommerce-admin",' ) !== false ) {
+			$line .= "\t\"version\": \"{$package_json->version}\",\n";
+		}
 		$lines[] = $line;
 	}
 	file_put_contents( $filename, $lines );
@@ -34,3 +37,4 @@ replace_version( 'woocommerce-admin.php', $package );
 replace_version( 'src/FeaturePlugin.php', $package );
 replace_version( 'src/Package.php', $package );
 replace_version( 'readme.txt', $package );
+replace_version( 'composer.json', $package );
--- a/woocommerce-admin.php
+++ b/woocommerce-admin.php
@@ -27,7 +27,7 @@ use \Automattic\WooCommerce\Admin\FeaturePlugin;
  * We want to fail gracefully if `composer install` has not been executed yet, so we are checking for the autoloader.
  * If the autoloader is not present, let's log the failure and display a nice admin notice.
  */
-$autoloader = __DIR__ . '/vendor/autoload.php';
+$autoloader = __DIR__ . '/vendor/autoload_packages.php';
 if ( is_readable( $autoloader ) ) {
 	require $autoloader;
 } else {
--- a/woocommerce-admin.php
+++ b/woocommerce-admin.php
@@ -20,6 +20,7 @@
 defined( 'ABSPATH' ) || exit;
 
 use \Automattic\WooCommerce\Admin\FeaturePlugin;
+use \Automattic\WooCommerce\Admin\Loader;
 
 /**
  * Autoload packages.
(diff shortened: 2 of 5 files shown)


woocommerce-admin.php
```diff
@@ -66,4 +67,77 @@ function() {
...
+/**
+ * Returns whether the current version is a development version
+ * Note this relies on composer.json version, not plugin version.
+ * Development installs of the plugin don't have a version defined in
+ * composer json.
+ *
+ * @return bool True means the current version is a development version.
+ */
+function woocommerce_admin_is_development_version() {
+	$composer_file = __DIR__ . '/composer.json';
```
timmyc, line 79: Nice @jeffstieler really great way of sniffing out the dev version.

timmyc (approved):
Thanks for grabbing this one Jeff. This is testing out well for me per the instructions, and verified the proper version was loading in the Woo SSR. Going to have one other person give this a test before we merge + pick.

bin/update-version.php
```diff
@@ -25,6 +25,9 @@ function replace_version( $filename, $package_json ) {
 		if ( stripos( $line, 'Stable tag: ' ) !== false ) {
 			$line = "Stable tag: {$package_json->version}\n";
 		}
+		if ( stripos( $line, '"name": "woocommerce/woocommerce-admin",' ) !== false ) {
+			$line .= "\t\"version\": \"{$package_json->version}\",\n";
+		}
```
psealock, line 30: You'll need to delete the previous "version" line somhow. Running `npm run bump-version` a few times adds a new version line.

Screen Shot 2020-08-04 at 11 05 24 AM

jeffstieler: This really should only be run for releases. Development builds won't create the version property and we should only commit the changes to the Composer JSON to release branches.

I agree that more robust handling of this version should be built - are you OK with pushing that to a follow up as to not block this PR from landing?

psealock: Yes, totally as this would only present a problem on the second release.

psealock (approved):
LGTM with only issue noted to be addressed as a follow up.

PR #4924, opened 2020-08-04
Task List: dismiss and undo dismiss animation fixed

### LINKED ISSUE #4892: [WC 4.4 beta1] Task List: Dismiss animation
The animation experience here feels confusing. When I dismiss the extension task, it's actually the store personalization task that animates out, which leaves me in a moment of confusion around which task I actually dismissed.

Current implementation:
dismiss

I know this is applied to the Inbox panel, but the animation pattern should be the same as [this](https://cldup.com/hyaHs3ISyP.gif). IE the item that is dismissed is the one that animates in / out :) 

I appreciate animation can be finicky to get right. If we can't fix this for the 4.4 release date then I would be in favor of removing the current animation for now as an interim.

octaedro:
Fixes #4892

This PR replaces a prop we send to the component "CSSTransition" for `item key`.

We had a problem with the component "CSSTransition" since the animation was having an unexpected behavior.

### Screenshots
Before
dismiss

Now
Screen Capture on 2020-08-04 at 18-48-25

1. Go to the OBW and add a product that requires purchase (e.g. Bookings or Bundles).
2. Go to the task list.
3. Dismiss the task `Purchase & install extensions`.
4. Make sure that the animation works as expected and the item that disappears is the selected one.
5. A snack-bar will appear with the button `Undo`, press it.
6. Make sure the undo item animation works as expected. The undeleted item should appear with an animation.
7. Delete `woocommerce_task_list_dismissed_tasks` if you need all the tasks again.

### Changelog Note:

Fix: Task List: dismiss and undo dismiss animation fixed

### DIFF
--- a/packages/components/src/list/index.js
+++ b/packages/components/src/list/index.js
@@ -42,13 +42,14 @@ class List extends Component {
 				className={ listClassName }
 				role="menu"
 			>
-				{ items.map( ( item, i ) => {
+				{ items.map( ( item ) => {
 					const {
 						after,
 						before,
 						className: itemClasses,
 						content,
 						href,
+						key,
 						listItemTag,
 						onClick,
 						target,
@@ -80,7 +81,7 @@ class List extends Component {
 
 					return (
 						<CSSTransition
-							key={ i }
+							key={ key }
 							timeout={ 500 }
 							classNames="woocommerce-list__item"
 						>


timmyc (approved):
Thanks so much for the fix @octaedro - testing well for me.

timmyc:
cherry picked to `release/1.4.0-rc.1` in c1fb7e70d2b40f77f9d29cc4722dd6db25d838ca

PR #4932, opened 2020-08-05
[Experimental] Attempt an update to jest and throttle of workers to improve test speed.

samueljseay:
This is an experiment to improve the performance of jest tests. locally these changes improve test speed by more than `2x` but I'd also like to compare the results on CI and with others before confirming this change.


package.json
```diff
@@ -55,7 +55,7 @@
 		"prestart": "npm run -s install-if-deps-outdated",
 		"start": "cross-env WC_ADMIN_PHASE=development npm run build:packages && cross-env WC_ADMIN_PHASE=development npm run build:feature-config && concurrently \"cross-env WC_ADMIN_PHASE=development webpack --watch\" \"node ./bin/packages/watch.js\"",
 		"pretest": "npm run -s install-if-no-packages",
-		"test": "wp-scripts test-unit-js --config tests/js/jest.config.json",
+		"test": "wp-scripts test-unit-js --config tests/js/jest.config.json --maxWorkers=50%",
```
samueljseay, line 58: The most crucial change is throttling workers (not fixed to a number so that it should work in more environments)

PR #4937, opened 2020-08-05
Load chunked translation data from language packs.

### LINKED ISSUE #2336: Translations are not accepted
Several users reported translations are not honored in the WooCommerce Admin pages, while they appear correctly in the menu.

Thread in the .org repo:

https://wordpress.org/support/topic/translation-is-not-accepted/

jeffstieler:
Fixes #2336. (Again)

The introduction of code splitting broke translations for strings that ended up in JS chunk files. When translate.wordpress.org generates files, it automatically scans all JS in the project. Because the split code chunks aren't registered with WordPress via 
`wp_register_script()` and `wp_set_script_translations()`, the corresponding JSON translation files weren't being loaded.

I think the ideal solution would be to handle this all in JS, but it would require copying some code from Calypso (a Webpack plugin to subscribe to chunk loading, and an i18n context component) and some seemingly major application refactoring.

This PR seeks to provide an all-PHP solution to the problem. When loading the "main" application script (identified by the `WC_ADMIN_APP` handle) the JSON translation files are parsed, the "chunk" specific ones are combined into a single file (and written to the filesystem as a cache for subsequent loads) and served in place of the single JSON file representing just the `dist/app/index.js` file.

The code can handle being loaded as a package within WooCommerce core as well.

- On `main`:
- Set your store language to German (or some other language we have a language pack for (>85%))
- Go to the updates screen, click "update translations" (should be very last button on screen)
- Go to Analytics > Revenue
- Verify that "Gross Sales" is **not** translated
- Check out this branch
- Go to Analytics > Revenue
- Verify that "Gross Sales" **is** translated
- Copy `src/Loader.php` to your local WooCommerce plugin (`/packages/woocommerce-admin/src/`)
- Deactivate WooCommerce Admin plugin
- Go to Analytics > Revenue
- Verify that "Gross Sales" **is** translated

### Changelog Note:

Fix: translation loading from JavaScript files.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -495,7 +495,7 @@ class Loader {
 
 	/**
 	 * Find and combine translation chunk files.
-	 * 
+	 *
 	 * Only targets files that aren't represented by a registered script (e.g. not passed to wp_register_script()).
 	 *
 	 * @param string $lang_dir Path to language files.
@@ -504,16 +504,18 @@ class Loader {
 	 * @return array Combined translation chunk data.
 	 */
 	public static function get_translation_chunk_data( $lang_dir, $domain, $locale ) {
+		global $wp_filesystem;
+
 		// Grab all JSON files in the current language pack.
-		$json_i18n_filenames = glob( $lang_dir . $domain . '-' . $locale . '-*.json' );
+		$json_i18n_filenames       = glob( $lang_dir . $domain . '-' . $locale . '-*.json' );
 		$combined_translation_data = array();
 
 		foreach ( $json_i18n_filenames as $json_filename ) {
-			if ( ! is_readable( $json_filename ) ) {
+			if ( ! $wp_filesystem->is_readable( $json_filename ) ) {
 				continue;
 			}
 
-			$file_contents = \file_get_contents( $json_filename );
+			$file_contents = $wp_filesystem->get_contents( $json_filename );
 			$chunk_data    = \json_decode( $file_contents, true );
 
 			if ( empty( $chunk_data ) ) {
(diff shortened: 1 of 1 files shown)


octaedro (approved):
Great job @jeffstieler! This tests well here and code LGTM :shipit:

src/Loader.php
```diff
@@ -479,6 +479,140 @@ public static function register_scripts() {
...
+	 *
+	 * @return string JSON encoded translations object.
+	 */
+	public static function load_script_translations( $original_translations, $file, $handle, $domain ) {
+		// Make sure the main app script is being loaded.
+		if ( WC_ADMIN_APP !== $handle ) {
+			return $original_translations;
+		}
+
+		// Make sure we're handing the correct domain (could be woocommerce or woocommerce-admin).
```
timmyc, line 578: If we are already verifying that the script is `WC_ADMIN_APP` would there ever be a chance that the text domain would be different? I'm guessing it would only be if another plugin used the same file/handle name?

jeffstieler: Yeah this is definitely overly defensive. Another plugin would need to have a `wc-admin-app` script.

src/Loader.php
```diff
@@ -479,6 +479,140 @@ public static function register_scripts() {
...
+		$cache_filename = self::get_combined_translation_filename( $domain, $locale );
+		$lang_dir       = WP_LANG_DIR . '/plugins/';
+
+		// Allow us to easily interact with the filesystem.
+		require_once ABSPATH . 'wp-admin/includes/file.php';
+		\WP_Filesystem();
+		global $wp_filesystem;
+
+		// First attempt to get a previously generated combined translation file.
+		if (
```
timmyc, line 594: Had one more thought since our conversation on this topic yesterday. Regarding invalidation of the combined file I was thinking of the following scenario:

- This code ships in 4.4, translations are fixed and all is well in the world.
- We work on fab new features for 4.5, which include new strings. Translations happen for new strings
- After updating my site to 4.5, the old cache file in the lang directory still exists, so the code here returns the old file ( at least that is what i'm wondering could possibly happen ).

I didn't test that flow yet but was wondering if it _could_ happen. If so, wondering if we either need to add something to the update routine in each release to clear out the cached/combined translation file... or maybe we can add in the woocommerce_admin_version number to the file name to act as a cache buster in this scenario.

jeffstieler: There is a version string in the filename: https://github.com/woocommerce/woocommerce-admin/pull/4937/files#diff-3864ffe37f3a1de39876ea62cae525f3R491

timmyc: i missed that, thanks!!!

PR #4941, opened 2020-08-06
Reviews: Create wp.data dataStore

### LINKED ISSUE #3376: Data Layer: Migrate reviews store to wp.data

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/3376

Replace wc-api usage of reviews data with wp.data implementation. The reducer and selectors are made to replicate the same API as wc-api.

1. Make sure you have an unread review.
2. Load the homescreen and see an alert dot on the Reviews header tab.
3. Click the Reviews header tab and see the reviews populate.
4. Coerce an error by adding this function to `src/API/ProductReviews.php`

```php
public function get_items() {
    return new WP_Error( 'Take me to flavortown` );
}
```
5. See the Reviews header tab display an error.

### Note

Using the headers to get a total count is an often used pattern, ie `response.headers.get( 'x-wp-total' )`. This PR introduces a control `fetchWithHeaders` which can ultimately be pulled out from the reviews folder and somewhere reusable.

### Changelog Note:

- dev: Replace Review's fresh-data with wp.data data store.


client/wc-api/wc-api-spec.js
```diff
@@ -38,7 +36,6 @@ function createWcApiSpec() {
 					...notes.operations.read( resourceNames ),
 					...reportItems.operations.read( resourceNames ),
 					...reportStats.operations.read( resourceNames ),
-					...reviews.operations.read( resourceNames ),
```
samueljseay, line 41: 👏

packages/data/src/reviews/controls.js
```diff
@@ -0,0 +1,28 @@
...
+const controls = {
+	...dataControls,
+	FETCH_WITH_HEADERS( { options } ) {
+		let headers;
+		return apiFetch( { ...options, parse: false } )
+			.then( ( response ) => {
+				headers = response.headers;
+				return response.json();
+			} )
+			.then( ( data ) => ( { headers, data } ) );
```
samueljseay, line 24: I guess its probably personal preference to not declare a variable outside the scope and manage it this way, but you could do this differently via Promise.all ([which is happy to take promise and non promise values](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all)).

e.g.

```
return apiFetch( { ...options, parse: false } )
			.then( ({headers, json} ) => {
                 return Promise.all([headers,json()])
			} )
			.then( ( [headers,data] ) => ( { headers, data } ) );
```

packages/data/src/reviews/index.js
```diff
@@ -0,0 +1,31 @@
...
+import { STORE_NAME } from './constants';
+import * as selectors from './selectors';
+import * as actions from './actions';
+import * as resolvers from './resolvers';
+import controls from './controls';
+import reducer from './reducer';
+
+const storeSelectors = select( STORE_NAME );
+
+// @todo This is used to prevent double registration of the store due to webpack chunks.
```
samueljseay, line 19: Is there a github issue somewhere for this? It would be great to be able to track when we can remove this conditional by adding a link in the comment if possible.

packages/data/src/reviews/reducer.js
```diff
@@ -0,0 +1,47 @@
...
+				reviews: {
+					...state.reviews,
+					[ JSON.stringify( query ) ]: { data: ids, totalCount },
+				},
+				data: {
+					...state.data,
+					...nextReviews,
+				},
+			};
+			break;
```
samueljseay, line 33: Again personal preference (although I think in this case it would be more succinct):

We probably don't need to mutate the state variable at all in each `case` and then we can remove the `break;` statements and the end return statement. Instead I think it's a bit more idiomatic/succinct to return the state from the case. I also personally like having `{}` around the case but I'm not sure if our lint supports this (it does also let you declare the same var name twice between case statements which can be useful). Anyway, so without the `{}` would look like:

```
case TYPES.UPDATE_REVIEWS:
			const ids = [];
			const nextReviews = reviews.reduce( ( result, review ) => {
				ids.push( review.id );
				result[ review.id ] = review;
				return result;
			}, {} );
			return {
				...state,
				reviews: {
					...state.reviews,
					[ JSON.stringify( query ) ]: { data: ids, totalCount },
				},
				data: {
					...state.data,
					...nextReviews,
				},
			};
```

we lose the return statement at the end and we lose all the breaks. It can also be nice to provide a `default` case that just returns state untouched.

packages/data/src/reviews/test/reducer.js
```diff
@@ -0,0 +1,61 @@
+/**
```
samueljseay, line 1: Just a little thing. we could consider using the jest environment tag here to run these tests in a `node` environment. For every suite we run in `node` we do get a upfront performance bonus because setting up a jsdom env happens per suite and I've seen some metrics saying this can add an additional 1 or more seconds per suite, which will add up over time as we add significantly more tests. Any reducers, generators, anything that doesn't interact with the DOM could be run under the node environment.

The tag is like:

```
/**
 * @jest-environment node
 */
```

samueljseay (approved):
Everything works for me as expected. A few suggestions from me but none are blocking.  Great stuff, its awesome to see this move to `wp.data`!

packages/data/src/reviews/controls.js
psealock: Nice! I was thinking there must be better way to do this. I applied your suggestion in https://github.com/woocommerce/woocommerce-admin/pull/4941/commits/91a8467fc9fa4a6ce0fb5a01505aacc0de933548.

Interesting to note that destructuring of `response` led to `TypeError: Failed to execute 'json' on 'Response': Illegal invocation`. Instead, `response.json()` worked.

packages/data/src/reviews/index.js
psealock: There is. I added it to the comment here and other instances in https://github.com/woocommerce/woocommerce-admin/pull/4941/commits/93cfb484fa5ff7beda8be30d1b124298daf3ac1d

packages/data/src/reviews/test/reducer.js
psealock: Sounds like a good idea. I'm getting `Invalid JSDoc tag name "jest-environment".eslint(jsdoc/check-tag-names)` linter error. May need to address this in a follow up.

packages/data/src/reviews/reducer.js
psealock: Cool, done

timmyc:
` return new WP_Error( 'Take me to flavortown` );`

Ship. It.

PR #4942, opened 2020-08-06
Improve jest performance significantly by downgrading to 24.9.0

samueljseay:
Recently myself and others had noticed significant slowdowns of the test suite in Jest. I frequently had the test suite
taking 130-150 seconds to run.

After some exploration and trial/error I found that https://github.com/facebook/jest/issues/9457 is likely the cause.

This was confirmed after a downgrade to 24.9.0 where I saw test suite runs drop to an average of about 30 seconds per run! Tentatively I'm excited as it seems to have dropped a full minute off the Travis build time too (based on recent builds).

To achieve this I have had to drop `@wordpress/jest-preset-default` for now since it will not work with anything pre Jest 25. I've copied across the bits of the preset we need to our own config.

Also I noticed that running the tests through `wp-scripts` was significantly slowing things down as well. It seemed pretty redundant, so I call jest directly now in the script. Also removing wp-scripts allows test output to work properly in Travis which is great because now you can actually see how long they're taking in the Travis output.

One other way we could get more performance in future is only using `jsdom` environment on tests that actually need it and defaulting to `node` instead. I didn't do that here because the downgrade gave us the biggest performance boost.

## Testing

* To test this do a fresh `npm install` and `npm test`. Compare test run times on this branch to test runs on `main`.


tests/js/setup-globals.js
```diff
@@ -101,7 +101,11 @@ wooCommercePackages.forEach( ( lib ) => {
 } );
 
 const config = require( '../../config/development.json' );
-window.wcAdminFeatures = config && config.features ? config.features : {};
+
+// Check if test is jsdom or node
```
samueljseay, line 105: This check means this setup file can run without crashing on `node` environment or `jsdom` environment, later we can start using node environment on tests that don't need DOM to speed things up further

package.json
```diff
@@ -164,14 +164,14 @@
 		"@wordpress/babel-preset-default": "4.10.0",
 		"@wordpress/browserslist-config": "2.6.0",
 		"@wordpress/custom-templated-path-webpack-plugin": "1.6.0",
-		"@wordpress/jest-preset-default": "5.5.0",
```
psealock, line 167: Potentially naive question: Would downgrading `@wordpress/jest-preset-default` do the task without the other modifications?

jeffstieler:
Just a heads up that we upgraded the WooCommerce E2E env package to Jest >= 25 to fix issues with our project: https://github.com/woocommerce/woocommerce/pull/27090

package.json
samueljseay: @psealock thats not naive, except I'd rather move away from the preset for now because the setup files it runs crash in a `node` environment test (which will be crucial if we want to keep improving performance). It is my plan to raise an issue about this in Gutenberg.

samueljseay:
@jeffstieler Thanks for the heads up, I'll look into that one. I wonder if we could run different versions of jest for the different tests... or if there is another possible solution.

samueljseay:
@jeffstieler my solution to this for now is to use an npm package alias and run jest 24.9.0 for the unit tests only. This should be compatible with anything else that wants to run a different version of Jest.

jeffstieler:
After switching branches, I'm running this command: `rm -rf node_modules && npm install && npm test`

On `main`:

```
Test Suites: 70 passed, 70 total
Tests:       564 passed, 564 total
Snapshots:   52 passed, 52 total
Time:        15.535s
Ran all test suites.
```

On this branch:

```
Test Suites: 70 passed, 70 total
Tests:       564 passed, 564 total
Snapshots:   52 passed, 52 total
Time:        12.381s
Ran all test suites.
```

While there is an improvement, it's not as drastic as I expected given the PR.

My local setup is node `v12.13.0` and npm `6.13.4`.

samueljseay:
@jeffstieler I don't 100% know whats going on but I still think this is worth merging. I will do another test of `main` in my environment though to see if I can reproduce something closer to your result. I have *never* had the tests running even close to 15 seconds in my environment.

samueljseay:
@jeffstieler or anyone else keen to give this approval, I'd really appreciate it.

Detailed findings:

run on `main` output:

```
Test Suites:    70 passed, 70 total
Tests:             564 passed, 564 total
Snapshots:    52 passed, 52 total
Time:              96.786s
```

run on this branch:

```
Test Suites:    70 passed, 70 total
Tests:              564 passed, 564 total
Snapshots:     52 passed, 52 total
Time:              45.956s
```

Other benefits of this work:

1. Test output now works on Travis CI which is helpful
2. Travis CI builds are consistently faster dropping from ~7min to 6min
3. Introducing the global checks in setup files lets us run tests faster in the `node` environment where applicable. (This we could not have without departing from the existing preset)

Although the improvement may not be as marked in some environments I think its good enough to see the significant improvement in some.

becdetat:
Average of five runs on `main` for me is 30.17s and on this branch is 23.12s, which is an improvement of roughly 25%? if my math is right. 💃 

Code looks good 👍

becdetat (approved):

PR #4943, opened 2020-08-06
Migrate notes data store to wp data

### LINKED ISSUE #3380: Data Layer: Migrate notes store to wp.data

joshuatf:
Fixes #3380

Moves the notes data store from wc-api to wp data under the `data` package. 

* Migrates selectors, actions, and resolvers to the new data store.  Caches all notes under `state.notes` and individual queries reference note IDs from `state.noteQueries`.
* Updates references to these methods in components.
* Makes some updates around checking for requesting/resolving status and note updating status (now stored in individual note).  /cc @octaedro for your review on this piece to make sure things work as expected 😄 
* Moves the notices for dismissal into the components.

### Screenshots
Screen Shot 2020-08-06 at 8 26 25 AM

1. Check all places that retrieve notes (activity panel inbox / historical data / unread indicators / store alerts / profile wizard).
1. Test dismissing individual and all notes.
1. Try to undo both the above dismissals using the undo button in the notice.
1. Make sure notes loading status appears correctly when resolving.


src/API/Notes.php
```diff
@@ -327,15 +327,25 @@ public function get_requested_updates( $request ) {
...
 
 	/**
-	 * Undo delete all notes.
+	 * Batch update a set of notes.
 	 *
 	 * @param WP_REST_Request $request Request object.
 	 * @return WP_REST_Request|WP_Error
 	 */
-	public function undoremove_items( $request ) {
+	public function batch_update_items( $request ) {
```
joshuatf, line 335: I realized this endpoint is already set up to batch update, which could be pretty useful.

Not sure if we need to add a deprecated method for `undoremove_items` if we go forward with this update.

octaedro (commented):
Hi @joshuatf, you did an awesome job here! The code looks great, I just left a comment and wanted to ask you if we should remove:
* `getUndoDismissRequesting` from [client/homescreen/layout.js](https://github.com/woocommerce/woocommerce-admin/blob/update/note-data-store/client/homescreen/layout.js#L167) too? It looks like we should use `isBatchUpdating` instead.
* and the `notes` references from [client/wc-api/wc-api-spec.js](https://github.com/woocommerce/woocommerce-admin/blob/update/note-data-store/client/wc-api/wc-api-spec.js#L7).

client/header/activity-panel/panels/inbox/index.js
```diff
@@ -54,7 +48,7 @@ const renderNotes = ( {
 	const notesArray = Object.keys( notes ).map( ( key ) => notes[ key ] );
 
 	return notesArray.map( ( note ) => {
-		if ( isDismissUndoRequesting === note.id ) {
+		if ( note.isUpdating ) {
```
octaedro, line 51: When we are updating a note (when undoing the dismission) `note.isUpdating` is `undefined` but should be `true` to be able to show the placeholder in its place.

src/API/Notes.php
timmyc: > Not sure if we need to add a deprecated method for undoremove_items if we go forward with this update.

I think that is the safe thing to do given prior feedback on matters like this.

joshuatf: Added deprecated method and notice.

joshuatf:
Thanks for that feedback, @octaedro and great catches!  Handled all your feedback and the deprecated method in https://github.com/woocommerce/woocommerce-admin/pull/4943/commits/b56ee5118214bf8a6f212a4a2c69a65c77af2778.

This is ready for re-review.

octaedro:
I'm experiencing some trouble in the `Home` screen / `Dashboard`. It's throwing a couple of console errors when I dismiss all the notes and later `Undo` the dismissal.
Also, probably related, the tests are failing on `WC_Tests_API_Admin_Notes::test_undo_all_notes_delete`.

joshuatf:
Thanks again for reviewing, @octaedro 😄 

Fixed failing tests in https://github.com/woocommerce/woocommerce-admin/pull/4943/commits/4ec45c4678fbb2234779263bb8ce5a04c9020562 and fixed the frontend infinite loop in https://github.com/woocommerce/woocommerce-admin/pull/4943/commits/c59c0371e436f5e12895c2959e72fa47bbcc697e.

Also gave this PR a rebase.  Could you give it another look when you have a moment?

octaedro (approved):
Great job @joshuatf! Everything is working well here and code looks good :shipit:

psealock:
@joshuatf I'm getting a couple warnings on `main` that seem related to this PR. Can you reproduce?

Screen Shot 2020-08-14 at 2 09 54 PM

joshuatf:
I can, thanks for the heads up, @psealock.  PR opened [here](https://github.com/woocommerce/woocommerce-admin/pull/4990) to address that and cleaning up those constants.

PR #4944, opened 2020-08-06
Don't redirect to connect Jetpack if "no thanks!" selected.

### LINKED ISSUE #4939: [4.4 RC1] Benefits Screen Redirects to Jetpack When No Selected
**Describe the bug**
When on the benefits page in `main` or WooCommerce 4.4 RC1, if you select "No Thanks" to setup WCS and Jetpack - when Jetpack is installed + active, one is still redirected into the Jetpack Connection flow.

Worth noting the benefits page works as expected when Jetpack is not active, so when clicking no thanks then, you are sent to the homescreen as expected. As such this feels a bit like an edge case to me, but still would be nice to fix prior to 1.4/4.4 final.

**To Reproduce**

1. Create a new JN site with Jetpack active ( not connected ), and running RC1 of 4.4
2. Proceed through the onboarding profiler to the benefits page
3. Click "No Thanks" when prompted to setup/connect for Jetpack and WCS
4. Note you are redirected into the Jetpack connection flow still. 

**Expected behavior**
To be redirected to the WooCommerce homescreen.

jeffstieler:
Fixes #4939

This PR checks whether or not the user opted in to install Jetpack before redirecting to the connection flow.

- Install and activate (but do not connect) Jetpack
- Go through the OBW
- Select "no thanks!" on the Jetpack step
- Verify that you are redirected to the WooCommerce Home screen, not a Jetpack screen

### Changelog Note:

N/A - unreleased bug.

### DIFF
--- a/client/profile-wizard/index.js
+++ b/client/profile-wizard/index.js
@@ -219,12 +219,17 @@ class ProfileWizard extends Component {
 			getPluginsError,
 			isJetpackConnected,
 			notes,
+			profileItems,
 			updateNote,
 			updateProfileItems,
 		} = this.props;
 		recordEvent( 'storeprofiler_complete' );
+
+		const { plugins } = profileItems;
 		const shouldConnectJetpack =
-			activePlugins.includes( 'jetpack' ) && ! isJetpackConnected;
+			( plugins === 'installed' || plugins === 'installed-wcs' ) &&
+			activePlugins.includes( 'jetpack' ) &&
+			! isJetpackConnected;
 
 		const profilerNote = notes.find(
 			( note ) => note.name === 'wc-admin-onboarding-profiler-reminder'


client/profile-wizard/index.js
```diff
@@ -219,12 +219,17 @@ class ProfileWizard extends Component {
...
+			profileItems,
 			updateNote,
 			updateProfileItems,
 		} = this.props;
 		recordEvent( 'storeprofiler_complete' );
+
+		const { plugins } = profileItems;
 		const shouldConnectJetpack =
-			activePlugins.includes( 'jetpack' ) && ! isJetpackConnected;
+			( plugins === 'installed' || plugins === 'installed-wcs' ) &&
```
joshuatf, line 230: Clever solution!  Really nice we can keep benefits stateless.

joshuatf (commented):
Code here looks good.  Just noting that in our previous discussion via Slack we decided to redirect to Jetpack regardless of whether WCS was additionally installed if Jetpack was already activated.

But I do see how that flow could be confusing, so I'm fairly indifferent on the approach here.  /cc @timmyc since you originally opened the issue.  Feel free to approve this as is if this is the behavior we want to move towards.

timmyc:
> Code here looks good. Just noting that in our previous discussion via Slack we decided to redirect to Jetpack regardless of whether WCS was additionally installed if Jetpack was already activated. But I do see how that flow could be confusing, so I'm fairly indifferent on the approach here. /cc @timmyc since you originally opened the issue. Feel free to approve this as is if this is the behavior we want to move towards.

Ah I must have missed that convo, or the memory of it was released from my buffer. Let's see what @pmcpinto has to say here before we proceed.

jeffstieler:
> Ah I must have missed that convo, or the memory of it was released from my buffer. Let's see what @pmcpinto has to say here before we proceed.

@timmyc I find it very hard to justify dropping a user into the Jetpack connection flow after they selected "No thanks" in response to "Enhance your store with Jetpack". At best it doesn't make sense, at worst it's a blatant disregard for the user's wishes.

pmcpinto:
I vaguely remember that Slack discussion, but I agree that it doesn't make sense to push users to connect to Jetpack if they select "No".

joshuatf (approved):
Looks like this is the behavior we want and code looks good.  LGTM :shipit:

jeffstieler:
Cherry picked to `release/1.4.0-rc.1`.

PR #4945, opened 2020-08-06
Update the colors of the illustrations in the welcome modal

jameskoster:
They now use the theme color var, and are consistent with the rest of the UI. 

I forgot to take a before screenshot, but here's the after:

illustration


timmyc (approved):
Looking fresh Jay - very nice

image

PR #4949, opened 2020-08-07
Remove usage of __experimentalResolveSelect

### LINKED ISSUE #4865: Use wp.data controls as an alternative to `experimentalResolveSelect` in Jetpack CTA
`experimentalResolveSelect` is problematic to use on older versions of Wordpress because its not available on the wp.data global. For now we are bundling a copy of it.

The original work to add `experimentalResolveSelect` is in https://github.com/woocommerce/woocommerce-admin/pull/4655

To remove the bundled copy we could take an approach of implementing a `@wordpress/data` control instead. (see here for a good overview of the purpose of controls: https://unfoldingneurons.com/2020/wordpress-data-store-properties-controls).

samueljseay:
Fixes #4865 

Originally `__experimentalResolveSelect` was introduced to handle resolving the Jetpack connection URL in 2 scenarios. An issue arose with this because it is not bundled in the wp.data global in older versions of Wordpress. The next step after that was to bundle it ourselves, which was okay as a temporary fix. 

Ideally though (in my opinion), we would keep as much logic out of components themselves, relegating the logic to HOCs or in the case of this refactor to a generator action.

This refactor takes a more redux-saga like approach for resolving side effects involved with jetpack connection, by delegating the complex async flow to a generator action instead of having it contained inside components.

This leads to a couple of immediate benefits:

1. Testability. Most of the logic inside the call to action can be tested now easily, and without a lot of mocking. We can test the generator actions separately.

2. Reusability. Although no one is using this functionality yet, the new generator action is exposed by the `@woocommerce/data` package and could easily be used by others to perform a full install and connect to Jetpack with very little additional code.

There were 2 scenarios where `__experimentalResolveSelect` was used to handle getting the jetpack connection URL. This is regression testing of existing functionality.

#### Install Jetpack CTA

1. With jetpack not installed, go to the homescreen and you should see the jetpack call to action displayed
2. If its not installed the button should show "Get jetpack"
3. Click the install button
4. You should see the buttons go to a animated busy state and they should be disabled
5. As the Jetpack plugin is installed and activated the button text should change to "Activate Jetpack", "Connect Jetpack"
6. At the end of this process it should redirect to the Jetpack plugin

#### Dismiss the CTA

1. With jetpack not installed, go to the homescreen and you should see the jetpack call to action displayed
2. If its not installed the button should show "Get jetpack"
3. Click the *dismiss* button
4. The CTA should disappear
5. on a page refresh the CTA should still be gone

#### Profiler 

1. With jetpack not installed on a fresh site, go through the setup wizard completing all the steps up to the theme
2. After the theme step it will ask you whether you would like to install and activate jetpack.
3. Choose yes. After some time the installation should finish and redirect to the Jetpack dashboard.

*Just a note that testing the final step with connection is blocked at the moment for me as landing on Jetpack 8.8 dashboard causes a JS exception that doesn't appear to be related to my changes.*


samueljseay:
*Note* that currently this includes changes from the Jest performance PR, which I branched off so I could have the perf benefits when running tests.

client/homescreen/stats-overview/install-jetpack-cta.js
```diff
@@ -2,109 +2,36 @@
...
-	isJetpackActivated,
-	isJetpackConnected,
-	installAndActivatePlugins,
-	isConnecting,
-	isInstalling,
-} ) {
-	const { updateUserPreferences, ...userPrefs } = useUserPreferences();
-	const [ isDismissed, setIsDismissed ] = useState(
-		( userPrefs.homepage_stats || {} ).installJetpackDismissed
+const getJetpackInstallText = ( jetpackInstallState ) => {
```
joshuatf, line 19: Nice pattern here!

packages/data/src/plugins/actions.js
```diff
@@ -125,6 +126,62 @@ export function* installAndActivatePlugins( plugins ) {
...
+export function* installJetpackAndConnect() {
+	try {
+		yield dispatch( STORE_NAME, 'installPlugins', [ 'jetpack' ] );
+		yield dispatch( STORE_NAME, 'activatePlugins', [ 'jetpack' ] );
+
+		const url = yield select( STORE_NAME, 'getJetpackConnectUrl', {
+			redirect_url: getAdminLink( 'admin.php?page=wc-admin' ),
+		} );
+
+		// getJetpackConnectUrl doesn't throw errors, it sets error state if
```
joshuatf, line 142: My personal preference on this is that we do both (throw errors and set them in state for later handling) to allow consumers to easily access this data.

Without throwing errors, we can create race conditions when error checking in components.

But would love to hear others thoughts on this and it's something we can tackle in a follow-up.

packages/data/src/plugins/actions.js
```diff
@@ -125,6 +126,62 @@ export function* installAndActivatePlugins( plugins ) {
...
+			'getJetpackConnectUrl'
+		);
+
+		if ( error ) {
+			yield createErrorNotice( error );
+		} else {
+			window.location = url;
+		}
+	} catch ( error ) {
+		yield createErrorNotice( error.message );
```
joshuatf, line 156: I feel the data package should remain data centric and avoid interactions with other data stores when possible as well as avoid any opinionated UI changes.

The consumer of these data stores may not want to show notices depending on their use case, but I do see the benefit and DRYness of creating notices directly from the store.  Up for debate, but whichever we choose I think we should be consistent across data stores.

client/profile-wizard/index.js
```diff
@@ -243,28 +239,12 @@ class ProfileWizard extends Component {
...
-					redirectUrl = jetpackConnectUrl;
-				} )
-			);
-		}
-
 		Promise.all( promises ).then( () => {
-			if ( redirectUrl ) {
-				window.location = redirectUrl;
-				return;
+			if ( shouldConnectJetpack ) {
```
joshuatf, line 243: Just want to note that this introduces blocking requests so it will take a performance hit, albeit marginal, by starting the Jetpack install/activation/connection after other promises have resolved.

We may want to consider a loading state in a follow-up if we go this route since on slower hosts this will result in long white screen.

client/profile-wizard/index.js
```diff
@@ -243,28 +239,12 @@ class ProfileWizard extends Component {
...
-					if ( error ) {
-						createNoticesFromResponse( error );
-						return;
-					}
-					redirectUrl = jetpackConnectUrl;
-				} )
-			);
-		}
-
 		Promise.all( promises ).then( () => {
```
joshuatf, line 242: We can probably remove the `Promise.all()` usage here since there's only one promise in this case.

packages/data/src/plugins/actions.js
```diff
@@ -125,6 +126,62 @@ export function* installAndActivatePlugins( plugins ) {
...
+	}
+}
+
+export function* connectToJetpack( failureRedirect ) {
```
joshuatf, line 160: It would be great to make this a control so we can just `yield connectToJetpack` inside `installJetpackAndConnect`.

packages/data/src/plugins/actions.js
```diff
@@ -125,6 +126,62 @@ export function* installAndActivatePlugins( plugins ) {
...
+			STORE_NAME,
+			'getPluginsError',
+			'getJetpackConnectUrl'
+		);
+
+		if ( error ) {
+			yield createErrorNotice( error );
+			window.location = failureRedirect;
+		} else {
+			window.location = url;
```
joshuatf, line 178: I have slightly mixed feelings about setting window location from a data store.  It feels like a decision that a data store should not be making.  That said, I think this action has a very obvious goal and consumers will be aware of this.

packages/data/src/plugins/controls.js
```diff
@@ -0,0 +1,49 @@
...
+ */
+import { controls as dataControls } from '@wordpress/data-controls';
+import apiFetch from '@wordpress/api-fetch';
+
+/**
+ * Internal dependencies
+ */
+import { WC_ADMIN_NAMESPACE } from '../constants';
+
+let optionNames = [];
```
joshuatf, line 12: `pluginNames`?

packages/data/src/plugins/controls.js
```diff
@@ -0,0 +1,49 @@
...
+		return new Promise( ( resolve ) => {
+			setTimeout( function () {
+				const names = optionNames.join( ',' );
+				if ( fetches[ names ] ) {
+					return fetches[ names ].then( ( result ) => {
+						resolve( result[ optionName ] );
+					} );
+				}
+
+				const url = WC_ADMIN_NAMESPACE + '/options?options=' + names;
```
joshuatf, line 36: This looks like an attempt to batch install/activate, is that right?  I'm a little confused by the endpoint here.

client/homescreen/stats-overview/test/install-jetpack-cta.js
```diff
@@ -0,0 +1,132 @@
...
+ * Internal dependencies
+ */
+import { JetpackCTA } from '../install-jetpack-cta';
+import { recordEvent } from 'lib/tracks';
+
+jest.mock( 'lib/tracks', () => ( {
+	recordEvent: jest.fn(),
+} ) );
+
+describe( 'JetpackCTA', () => {
```
joshuatf, line 17: Nice 💯

joshuatf (commented):
Nice work on this refactor, @samueljseay!  Love all the tests and easing the connection flow burden.

I left some comments and questions in the code about how we deal with data store responsibility.  Would love to hear your view on this and any others who have opinions one way or another.

packages/data/src/plugins/actions.js
samueljseay: @joshuatf I def had a double take when writing this code too, I think I could probably adjust this so we can pass a error callback of some kind instead, letting people do whatever they want when the error happens. I'll explore this.

packages/data/src/plugins/actions.js
samueljseay: @joshuatf yes this was something that felt a little out of scope for the refactor since the existing functionality does not throw in many cases and instead puts errors in state. I think we're in agreeance that a pattern of throwing makes much more sense and I would want to follow up in the future and refactor out this pattern.

packages/data/src/plugins/actions.js
samueljseay: Thats a good idea. I'll do that.

austress99:
I'm sorry for this app but I'm gonna learn as fast as possible

packages/data/src/plugins/actions.js
samueljseay: @joshuatf Personally I think its ok for the reason that this is how I'd tackle this in redux saga as well. usually it would be some kind of yield of a client side routing call, but basically redux-saga and our generator actions want you to do side effect stuff here versus inside components. I don't really like the pattern we currently have of trying to render a component and then if some particular state is met mutating window.location there instead. That said some of that distaste of redirecting inside a component may be my own personal bias.

packages/data/src/plugins/controls.js
samueljseay: Sorry about the confusion here. this file was copy pasta left in the PR by accident when I thought I needed a control earlier on. I'll delete this.

packages/data/src/plugins/actions.js
samueljseay: If you're happy to push this out to future refactors I think thats what I'd want to do here. Let me know what you think.

client/profile-wizard/index.js
samueljseay: I'm trying to figure out how this is more blocking than before. The only real difference is that before we ran 2 requests in parallel but the user still had to wait till both had completely resolved before an actual redirect happened? The code that shows the loading state still happens first, and no sooner than before?

packages/data/src/plugins/actions.js
joshuatf: If we throw an error inside this catch block, the consumer can then catch and decide what to do with it.  I like this along with setting the error in the data store so that components have access to the data on both sides.

https://github.com/woocommerce/woocommerce-admin/blob/9be744b63a3beb56b5d94abf8a0725b8822de722/packages/data/src/plugins/actions.js#L83

>One reason I thought this was ok was that it deals with a core store, but then... maybe not every implementer will want to use core/notices? Anyway I'm gonna investigate making this more agnostic.

I think you're right in that most of the time, usage of core/notices will not be a problem, but I do like the separation of this concern and making the data more agnostic.

Did you want to tackle here or a follow-up?

packages/data/src/plugins/actions.js
joshuatf: Absolutely- we can tackle these in a follow-up and make them more consistent across all our data packages.

packages/data/src/plugins/actions.js
```diff
@@ -125,6 +126,62 @@ export function* installAndActivatePlugins( plugins ) {
...
+		// something went wrong.
+		const error = yield select(
+			STORE_NAME,
+			'getPluginsError',
+			'getJetpackConnectUrl'
+		);
+
+		if ( error ) {
+			yield createErrorNotice( error );
+			window.location = failureRedirect;
```
joshuatf, line 176: This will potentially give us a hard redirect as opposed to using the router `getHistory().push()`.  `window.location` is probably more universally useful and makes sense, but just wanted to get your thoughts before merging.

client/profile-wizard/index.js
joshuatf: >The only real difference is that before we ran 2 requests in parallel

This is the main reason.  Running those in parallel seemed to cut down the time pretty significantly on slower hosts where we had complaints around plugin installation and Jetpack connection.

If each request took 4 seconds, running in parallel means completion in 4 seconds, while running synchronously would be 8 seconds.

But the improvements in logic and testability are great here and may be a good tradeoff for the performance loss, I just want to make sure we discuss and acknowledge it before merging.

packages/data/src/plugins/actions.js
joshuatf: Created a follow-up issue to take care of this in all data stores here - https://github.com/woocommerce/woocommerce-admin/issues/4963

samueljseay: Thanks!

packages/data/src/plugins/actions.js
samueljseay: @joshuatf I'm having a go at solving this one here. Thanks, the throw is a good suggestion

client/profile-wizard/index.js
samueljseay: @joshuatf gotcha, that makes sense. Yeah you're right its a trade off. I do wonder if there are other ways we could get performance back in future via further refactoring. I did the minimal work here to get this particular code working without `__experimentalResolveSelect` but I do think there may be ways to optimise.

packages/data/src/plugins/actions.js
samueljseay: @joshuatf yes I think the hard redirect will support *all* failure scenarios for outside consumers. and in rare cases maybe us as well, such as where we host our components in pages other than `admin.php`. We could also make this configurable in future if such a feature is desired?

samueljseay:
@joshuatf Thanks again for the helpful feedback. I think especially your [comments here](https://github.com/woocommerce/woocommerce-admin/pull/4949/files/a9e9efdbcf489f51c6b5d57504b1e21ebda1e329#r467865874) lead to a better solution. I'd be keen for you to have another look.

In summary I've:

1. DRY'd up the actions as you [suggested here](https://github.com/woocommerce/woocommerce-admin/pull/4949#discussion_r467870403) so that the jetpack connection logic is reusable.

2. Refactored the action to take an error handler instead of creating an error notice, and I've made the `installPlugins` and `activatePlugins` more consistent by having them always throw `Error` exceptions. Now in the component we pass `createErrorNotice` as a callback instead of doing it directly in the action. The only thing I don't like is that we import the `createErrorNotice` action from the data plugin when it could probably live somewhere else, but I'm thinking maybe we could refactor that in future.

tsconfig.json
```diff
@@ -21,5 +21,5 @@
 			"@woocommerce/*": [ "../packages/*/src" ]
 		}
 	},
-	"exclude": [ "node_modules", "build" ]
+	"exclude": [ "node_modules", "build", "build-module", "dist", "vendor" ]
```
samueljseay, line 24: Also a little bug fix to tsconfig here, the linter gets overwhelmed by too many included dirs at the moment.

packages/data/src/plugins/actions.js
```diff
@@ -125,6 +126,52 @@ export function* installAndActivatePlugins( plugins ) {
...
+
+export function* installJetpackAndConnect( errorAction ) {
+	try {
+		yield dispatch( STORE_NAME, 'installPlugins', [ 'jetpack' ] );
+		yield dispatch( STORE_NAME, 'activatePlugins', [ 'jetpack' ] );
+
+		const url = yield dispatch( STORE_NAME, 'connectToJetpack' );
+		window.location = url;
+	} catch ( error ) {
+		yield errorAction( error.message );
```
joshuatf, line 158: Nice!  I really like the simplicity of the callback error here.  I think ultimately moving towards throwing errors in all actions to make them consistently and catchable would be ideal, but we have an issue logged for this and can follow-up later.

joshuatf (commented):
Nice improvements here, Sam.  This is testing well for me on the Jetpack CTA on the homescreen.  On the benefits page when opting in with "Yes please" I get stuck at a white screen when attempting to connect.  Do you see this?

Also "Continue with my active theme" on the theme step does not seem to work for me on this branch if both Jetpack and WCA are already installed.

samueljseay:
> On the benefits page when opting in with "Yes please" I get stuck at a white screen when attempting to connect. Do you see this?

@joshuatf so sorry I missed this, in my rush to get it back to review status I missed this. I've fixed this now.

> Also "Continue with my active theme" on the theme step does not seem to work for me on this branch if both Jetpack and WCA are already installed.

This one is a bit of a yikes. The code that handles moving on from this step is problematic and I'm wondering if subtle changes I've introduced have introduced a race condition. The way it works currently clicking that button sets off a chain of `componentDidUpdate` calls (some of which set the state again and cause it to be called again 😱 ). I'm digging in to this one because it sometimes redirects and sometimes doesn't for me locally.

samueljseay:
> On the benefits page when opting in with "Yes please" I get stuck at a white screen when attempting to connect. Do you see this?

> Also "Continue with my active theme" on the theme step does not seem to work for me on this branch if both Jetpack and WCA are already installed.

@joshuatf so sorry I missed these in my hurry to refactor. I've fixed these issues now and would be keen for you to retest. As far as I can tell they now work as expected.

I do just notice I'm getting one lint failure for jsdoc tags on Travis, which is weird because our eslint marks it as a warning right now. I'm looking in to that further.

joshuatf (approved):
>@joshuatf so sorry I missed these in my hurry to refactor. I've fixed these issues now and would be keen for you to retest. As far as I can tell they now work as expected.

Not a problem at all.  Things look great now!  Awesome work on this; really great refactor here.

>I do just notice I'm getting one lint failure for jsdoc tags on Travis, which is weird because our eslint marks it as a warning right now. I'm looking in to that further.

I think these two errors are causing the failure.  We'll just need to fix up those JSDoc tags since https://github.com/woocommerce/woocommerce-admin/pull/4985 was merged in.

```
/home/travis/build/woocommerce/woocommerce-admin/packages/data/src/plugins/test/actions.js
  2:0  error  Invalid JSDoc tag name "jest-environment"  jsdoc/check-tag-names
/home/travis/build/woocommerce/woocommerce-admin/packages/data/src/plugins/test/reducer.js
  2:0  error  Invalid JSDoc tag name "jest-environment"  jsdoc/check-tag-names
```

This  LGTM once we fix those up and give this a rebase.

PR #4951, opened 2020-08-07
eslint-plugin: Fix dependency rule to allow WooCommerce block

psealock:
https://github.com/woocommerce/woocommerce-admin/pull/4714 added an eslint plugin package for distribution for all WooCommerce applications. It modified the `dependency-group.js` such that "External" and "Internal" dependencies are allowed only. There was some discussion on whether or not to include "WooCommerce" as well. For instance:

```js
/**
 * WooCommerce dependencies
 */
import { withPluginsHydration } from '@woocommerce/data';
```
Considering we used Gutenberg as a model for distributing packages and following patterns of documenting internal packages v. internal files, I think we should have the `WooCommerce dependencies` rule.

This PR simply copies Gutenberg's logic for [dependency-group.js](https://github.com/WordPress/gutenberg/blob/master/packages/eslint-plugin/rules/dependency-group.js) and replaces `WordPress` with `WooCommerce`.

1. npm install
2. npm run lint
3. See only 347 warnings, instead of the 891 on `main`

cc @nerrad 

### Changelog Note:

- dev: Fix eslint-plugin dependency rule to allow WooCommerce block

### DIFF
--- a/packages/eslint-plugin/CHANGELOG.md
+++ b/packages/eslint-plugin/CHANGELOG.md
@@ -1,3 +1,7 @@
-# 1.0.0
+# 1.0.0-beta.1
+
+-   Fix eslint-plugin dependency rule to allow WooCommerce block.
+
+# 1.0.0-beta.0
 
 -   Released package
--- a/packages/eslint-plugin/CHANGELOG.md
+++ b/packages/eslint-plugin/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 1.0.0-beta.1
+# 1.0.0-beta.1 (unreleased)
 
 -   Fix eslint-plugin dependency rule to allow WooCommerce block.
 
--- a/packages/eslint-plugin/docs/rules/dependency-group.md
+++ b/packages/eslint-plugin/docs/rules/dependency-group.md
@@ -4,7 +4,7 @@ Ensures that all top-level package imports adhere to dependencies grouping conve
 
 Specifically, this ensures that:
 
-- An import is preceded by "External dependencies" or "Internal dependencies" as appropriate by the import source.
+-   An import is preceded by "External dependencies" or "Internal dependencies" as appropriate by the import source.
 
 ## Rule details
 
@@ -13,6 +13,7 @@ Examples of **incorrect** code for this rule:
 ```js
 import { get } from 'lodash';
 import { Component } from '@wordpress/element';
+import { withPluginsHydration } from '@woocommerce/data';
 import edit from './edit';
 ```
 
@@ -24,6 +25,11 @@ Examples of **correct** code for this rule:
  */
 import { get } from 'lodash';
 import { Component } from '@wordpress/element';
+
+/**
+ * WooCommerce dependencies
+ */
+import { withPluginsHydration } from '@woocommerce/data';
 
 /*
  * Internal dependencies
--- a/packages/eslint-plugin/docs/rules/dependency-group.md
+++ b/packages/eslint-plugin/docs/rules/dependency-group.md
@@ -4,7 +4,7 @@ Ensures that all top-level package imports adhere to dependencies grouping conve
 
 Specifically, this ensures that:
 
--   An import is preceded by "External dependencies" or "Internal dependencies" as appropriate by the import source.
+-   An import is preceded by "External dependencies", "WooCommerce dependencies" or "Internal dependencies" as appropriate by the import source.
 
 ## Rule details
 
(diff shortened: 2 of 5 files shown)


nerrad:
I recall the discussion, but it was ambiguous to me on what the resolution was (with varying opinions). I personally don't think we need the extra rule for a couple reasons:

1. In most cases for Woo extensions, the `@woocommerce` package _is_ an external from the point of view of how it's imported.
2. What real practical benefit does having this extra grouping provide other than a quick way to see what `@woocommerce/*` package dependencies there are. Given the rationale, why don't we also list `@wordpress/*` package dependencies in their own block too? Bottom line, is I'm not sure the extra work for adding imports into three separate groupings is worth the minimal value it provides.

FWIW I've never liked the three groupings in Gutenberg core either and would have preferred it's just Internal/External.

With that said, I realize there's opinions on both sides here. But before we make the change can we make sure it's a decision teams are agreeing on?

For reference, this is the comment I was going by: I recall the discussion, but it was ambiguous to me on what the resolution was (with varying opinions). I personally don't think we need the extra rule for a couple reasons:

1. In most cases for Woo extensions, the `@woocommerce` package _is_ an external from the point of view of how it's imported.
2. What real practical benefit does having this extra grouping provide other than a quick way to see what `@woocommerce/*` package dependencies there are. Given the rationale, why don't we also list `@wordpress/*` package dependencies in their own block too? Bottom line, is I'm not sure the extra work for adding imports into three separate groupings is worth the minimal value it provides.

FWIW I've never liked the three groupings in Gutenberg core either and would have preferred it's just Internal/External.

With that said, I realize there's opinions on both sides here. But before we make the change can we make sure it's a decision teams are agreeing on?

For reference, this is the comment I was going by: I recall the discussion, but it was ambiguous to me on what the resolution was (with varying opinions). I personally don't think we need the extra rule for a couple reasons:

1. In most cases for Woo extensions, the `@woocommerce` package _is_ an external from the point of view of how it's imported.
2. What real practical benefit does having this extra grouping provide other than a quick way to see what `@woocommerce/*` package dependencies there are. Given the rationale, why don't we also list `@wordpress/*` package dependencies in their own block too? Bottom line, is I'm not sure the extra work for adding imports into three separate groupings is worth the minimal value it provides.

FWIW I've never liked the three groupings in Gutenberg core either and would have preferred it's just Internal/External.

With that said, I realize there's opinions on both sides here. But before we make the change can we make sure it's a decision teams are agreeing on?

For reference, this is the comment I was going by: I recall the discussion, but it was ambiguous to me on what the resolution was (with varying opinions). I personally don't think we need the extra rule for a couple reasons:

1. In most cases for Woo extensions, the `@woocommerce` package _is_ an external from the point of view of how it's imported.
2. What real practical benefit does having this extra grouping provide other than a quick way to see what `@woocommerce/*` package dependencies there are. Given the rationale, why don't we also list `@wordpress/*` package dependencies in their own block too? Bottom line, is I'm not sure the extra work for adding imports into three separate groupings is worth the minimal value it provides.

FWIW I've never liked the three groupings in Gutenberg core either and would have preferred it's just Internal/External.

With that said, I realize there's opinions on both sides here. But before we make the change can we make sure it's a decision teams are agreeing on?

For reference, this is the comment I was going by when I first did the package: p6q6po-aLM-p2#comment-91137

samueljseay:
@nerrad Are you saying you'd be happy for us to move to 2 blocks like `internal/external` instead? I am a fan of that change. Right now we have to use some kind of macro or textsnippet or copy paste to manage this rule. I agree with the sentiment that the rule feels verbose and doesn't add value.  (If I had my choice I'd probably do away with groupings altogether personally since I don't see huge value in grouping dependencies in general).

Anyway, my 2 cents is: if we can have a smaller less burdensome rule then I would vote for that.

nerrad:
> Are you saying you'd be happy for us to move to 2 blocks like internal/external instead?

That's what the current rule is in `@woocommerce/eslint-plugin`. So I'm saying I'd be happy to keep what is currently in the plugin and not _add_ an additional block like what is being proposed here. I know WC-Admin has some history of having a `WooCommerce dependencies` block but it's inconsistently applied (from what I remember) and would rather see it not be used.

> Would a p2 post or something similar be a good way to vote for this particular issue?

As I pointed out, there _was_ some discussion in the p2 post I created for the plugin (p6q6po-aLM-p2#comment-91137) and I thought based on the linked comment it was decided to _start_ with the smaller rule (Internal/External) only. I could be wrong though and if you folks think it needs further discussion I'm not opposed to that. 

Sidenote: Holy cow my original comment here was in triplicate in the content! I never noticed when I first posted - maybe something going on with Github at the time...

psealock:
I'll leave this for some time if anyone wants to chime in.

psealock:
Thanks for the input. I will close this and make another PR to remove the WooCommerce doc block.

PR #4956, opened 2020-08-10
Match the requires version to the exact Wordpress version number.

samueljseay:
When we bumped the minimum supported Wordpress version in the plugin version header last time, we didn't use the actual published version number. 

When I try to use Wordpress 5.3 with the current version of the plugin, Wordpress produces an error saying that the version of Wordpress doesn't meet the minimum specified by the plugin (Because we specify 5.3.0 a non existent release as the minimum).

Changing it to `5.3` as it is listed [here](https://wordpress.org/download/releases/)  fixes this in my environment. 

### Testing

1. To test this, install the [WP Downgrade](https://wordpress.org/plugins/wp-downgrade/) plugin and in settings for the plugin enter `5.3`. 

2. Install WooCommerce and WooCommerce Admin plugins

3. activate WooCommerce, activate WooCommerce Admin

4. WooCommerce Admin should activate successfully.


timmyc:
cherry picked to `release/1.4.0` in 3d469ba9440f1436a7661e7365f4f9f7a2fb4c64

becdetat:
Cherry picked to `release/1.5.0-beta.2` in 195ac6d29bd65832be3fc4fd886fab52905b10f6

PR #4958, opened 2020-08-10
Migrate export data store to wp.data

### LINKED ISSUE #3383: Data Layer: Migrate export store to wp.data

jeffstieler:
Fixes #3383 

This PR migrates the `export` store to `wp.data`.

I started to include more functionality in the `@woocommerce/data` package than exists in the `wc-api` version for future usage, but abandoned that as it can always be written later. (The only use of this in our app is the "download" button on report tables)

- View a report with more than a single page of results
- Click "download"
- Verify no HTTP/JS errors occur
- Verify a notice is created that informs you of the emailed report link
- Verify a report is generated an emailed

### Changelog Note:

Dev: migrate export store to wp.data (don't include this in plugin changelog)


packages/data/src/export/actions.js
```diff
@@ -0,0 +1,69 @@
...
+		const result = yield apiFetch( {
+			path: `${ NAMESPACE }/reports/${ type }/export`,
+			method: 'POST',
+			data: {
+				report_args: args,
+				email: true,
+			},
+		} );
+
+		if ( result && result.status === 'success' ) {
```
samueljseay, line 52: I was pretty sure that `apiFetch` returns a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response)? As far as I can tell, the data control calls the `@wordpress/api-fetch` package
directly.

`Response.status` does not contain a string, but a HTTP status code like 200. Could you confirm or deny if thats the case here?

Also I notice that apiFetch will [throw on anything thats not 200 or 300](https://github.com/WordPress/gutenberg/blob/master/packages/api-fetch/src/index.js#L57) so I wonder if perhaps a `catch` is sufficient here?

samueljseay, line 52: I was pretty sure that `apiFetch` returns a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response)? As far as I can tell, the data control calls the `@wordpress/api-fetch` package
directly.

`Response.status` does not contain a string, but a HTTP status code like 200. Could you confirm or deny if thats the case here?

Also I notice that apiFetch will [throw on anything thats not 2xx or 3xx](https://github.com/WordPress/gutenberg/blob/master/packages/api-fetch/src/index.js#L57) so I wonder if perhaps a `catch` is sufficient here?

packages/data/src/export/actions.js
```diff
@@ -0,0 +1,69 @@
...
+			yield setIsRequesting( 'startExport', { type, args }, false );
+
+			return result;
+		}
+
+		if ( result && result.status === 'error' ) {
+			throw new Error( result.message );
+		}
+
+		throw new Error();
```
samueljseay, line 63: Hopefully if you could rely purely on the `catch`, which I feel like you should be able to here, all these additional throws should be able to go away.

samueljseay (commented):
I haven't tested this yet, but keen to get your feedback on the issues I raised with the use of apiFetch.

Looks great though and awesome to see reducer tests.

packages/data/src/export/actions.js
jeffstieler: I see an object in the shape of the endpoint response:

Screen Shot 2020-08-12 at 8 44 31 AM

packages/data/src/export/actions.js
```diff
@@ -0,0 +1,69 @@
...
+		} );
+
+		if ( result && result.status === 'success' ) {
+			yield setExportId( type, args, result.export_id );
+			yield setIsRequesting( 'startExport', { type, args }, false );
+
+			return result;
+		}
+
+		if ( result && result.status === 'error' ) {
```
joshuatf, line 59: We should also probably remove this as well and favor the response status code over a returned `status`.  Sorry for any confusion here with prior art.

client/analytics/components/report-table/index.js
```diff
@@ -197,7 +201,34 @@ const ReportTable = ( props ) => {
 			);
 		} else {
 			downloadType = 'email';
-			initiateReportExport( endpoint, title, reportQuery );
+			startExport( endpoint, reportQuery )
+				.then( () =>
```
joshuatf, line 205: Nice!  Thanks for moving these notices in here.

packages/data/src/export/test/reducer.js
```diff
@@ -0,0 +1,92 @@
+/**
+ * Internal dependencies
+ */
+import reducer from '../reducer';
+import TYPES from '../action-types';
+import { hashExportArgs } from '../utils';
```
joshuatf, line 6: Could we also use `getResourceName` to create the resource name from the selector + arguments?  Or are there differences that require this to be a unique utility?

packages/data/src/export/actions.js
joshuatf: >I was pretty sure that apiFetch returns a Response?

`apiFetch` automatically parses the body unless `parse` is set to `false`.

>Also I notice that apiFetch will throw on anything thats not 2xx or 3xx so I wonder if perhaps a catch is sufficient here?

Agreed.  We'll probably remove the `status` portion of the response going forward and rely on the response code only.

joshuatf (commented):
Code here looks really solid.  Couple quick minor comments.

packages/data/src/export/test/reducer.js
jeffstieler: Oh nice! I was `md5`ing the arguments since report queries can be quite large, but I should be able to use `getResourceName()` instead of directly calling `JSON.stringify()`.

jeffstieler: Updated in dfd7c67.

jeffstieler:
@joshuatf @samueljseay this is ready for another review - I've updated the action creator to rely on HTTP status instead of the response body `status` property.

joshuatf (approved):
Thanks for the changes, Jeff!  Exports are running, notices are created, and I will officially have Sade stuck in my head for the rest of the day.

Once conflicts are fixed this LGTM! 🚢

PR #4962, opened 2020-08-11
Replace useFilters with wp's withFilters

psealock:
Eslint `react-hooks/rules-of-hooks` prevents the way we were using [a custom `useFilters` component](https://github.com/woocommerce/woocommerce-admin/blob/04774f21de707197a89d598bb18c8acee837c373/packages/components/src/higher-order/use-filters/README.md#L18).

It turns out `useFilters` was created before the (dark) ages of hooks and is actually a very outdated fork of `import { withFilters } from '@wordpress/components'`. `withFilters` has changed along the way and its filters now manipulate components instead of arrays or objects passed in filters. See the two README's below to see the differences in usage:

https://github.com/woocommerce/woocommerce-admin/tree/main/packages/components/src/higher-order/use-filters

https://github.com/WordPress/gutenberg/tree/master/packages/components/src/higher-order/with-filters

`withFilters` from Gutenberg now passes React components to the filter function so that extra components can be added to the UI. This is useful, but not the way we were using `useFilters`, which passed an array or object of data.

Our usage of `useFilters` was to trigger refreshes when filters were added. Considering these filters are run on application bootstrap and each page navigation, it seemed that refreshes aren't needed unless extensions add pages dynamically. This seems unlikely, but I've created a dummy extension to test this.

Download the test extension and make sure the filters are working as they should:

[penguins.zip](https://github.com/woocommerce/woocommerce-admin/files/5083949/penguins.zip)

1. See the added page under WooCommerce called "penguins"
2. See the added report under Analytics called "penguins"
3. And see the Analytics setting added under Analytics > Settings. Note settings manipulation causes an error, but that is outside scope of this PR

### Changelog Note:

- dev: deprecate usage of component `useFilters` in favor of using wp.component's `withFilters`.


packages/components/src/higher-order/use-filters/index.js
```diff
@@ -4,6 +4,26 @@
...
+ * Deprecated version of useFilters.
+ *
+ * @deprecated
+ *
+ * @param {string} hookName Hook name exposed to be used by filters.
+ *
+ * @return {Function} Higher-order component factory.
+ */
+export default function useFilters( hookName ) {
+	deprecated( 'useFilters', {
```
joshuatf, line 19: ❤️

client/analytics/settings/historical-data/status.js
```diff
@@ -51,6 +50,4 @@ function HistoricalDataStatus( { importDate, status } ) {
 	);
 }
 
-export default useFilters( HISTORICAL_DATA_STATUS_FILTER )(
-	HistoricalDataStatus
-);
+export default HistoricalDataStatus;
```
joshuatf, line 53: So this can no longer be filtered in any way, is that right?  Do we think it's unlikely to be filtered or is there another way to filter this?

joshuatf (commented):
Thanks for updating this, Paul!  Had no clue about the changes in these hooks.

Left one question in the code just to make sure I understand the changes.  Also, should we add `(deprecated)` to the top of the `useFilters` readme to mitigate use if someone stumbles on it?

packages/components/src/higher-order/use-filters/index.js
timmyc: +1 nice Paul!

client/analytics/settings/historical-data/status.js
psealock: It can, the `statusLabels` are still filtered as they always have been earlier in the file:

https://github.com/woocommerce/woocommerce-admin/blob/b6d467d1b49251a9ee3f569ebccae3834dce32b9/client/analytics/settings/historical-data/status.js#L12

psealock:
Thanks for the review @joshuatf, I added a deprecation line on the README in https://github.com/woocommerce/woocommerce-admin/pull/4962/commits/28eced660c9d00ac936a0f389c4bb821dac9d42f and answered your question about the ability to extend historical data statuses.

joshuatf (approved):
Thanks for the explanation, Paul!  LGTM pending the eslint conflict resolution. 🚢

PR #4965, opened 2020-08-11
Rename "profile" to "store details".

jeffstieler:
Follow up to https://github.com/woocommerce/woocommerce-admin/pull/4721.

This PR changes the labeling of "Profiler" to match the user-facing text of "Store details".

- Verify that all `/profiler` links have been updated to `/store-details`
- Verify that for a new store with onboarding enabled (might have to Help > Setup Wizard > Enable new onboarding) the profiler is loaded when going to the homescreen
- Verify that the "store details" task in the task list works
- Verify that the Help > Setup Wizard > Profile setup wizard button works

### Changelog Note:

N/A

### DIFF
--- a/client/dashboard/index.js
+++ b/client/dashboard/index.js
@@ -32,7 +32,7 @@ class Dashboard extends Component {
 			! profileSkipped &&
 			! window.wcAdminFeatures.homescreen
 		) {
-			getHistory().push( getNewPath( {}, `/profiler`, {} ) );
+			getHistory().push( getNewPath( {}, '/store-details', {} ) );
 		}
 
 		if ( window.wcAdminFeatures[ 'analytics-dashboard/customizable' ] ) {
--- a/client/dashboard/index.js
+++ b/client/dashboard/index.js
@@ -32,7 +32,7 @@ class Dashboard extends Component {
 			! profileSkipped &&
 			! window.wcAdminFeatures.homescreen
 		) {
-			getHistory().push( getNewPath( {}, '/store-details', {} ) );
+			getHistory().push( getNewPath( {}, '/setup-wizard', {} ) );
 		}
 
 		if ( window.wcAdminFeatures[ 'analytics-dashboard/customizable' ] ) {
--- a/client/homescreen/index.js
+++ b/client/homescreen/index.js
@@ -20,7 +20,7 @@ const Homescreen = ( { profileItems, query } ) => {
 	const { completed: profilerCompleted, skipped: profilerSkipped } =
 		profileItems || {};
 	if ( ! profilerCompleted && ! profilerSkipped ) {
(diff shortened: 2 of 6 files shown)


joshuatf (commented):
I think overall this is a good change.  The only confusion for me is that the first step is called "Store Details" as well.  What about "Store Setup" or "Onboarding" even (if that's not too revealing of internals.

joshuatf:
Anyone else on @woocommerce/isotope have opinions on this route?

becdetat:
I don't think it's too confusing although I see your point @joshuatf . What about "Store address"?

becdetat:
@jeffstieler there's a failing test on this that looks related to your changes: `WC_Tests_API_Onboarding_Profiles::test_update_profile_items` - an array that should be empty but isn't.

joshuatf:
>I don't think it's too confusing although I see your point @joshuatf . What about "Store address"?

Do you mean renaming the first step to "Store address" and then the route would change to `store-details`?  I guess that would work as well.  I like "setup" because it broadly covers all things we might do throughout the profiler and mirrors the word in core's original onboarding (`wc-setup`).

>@jeffstieler there's a failing test on this that looks related to your changes: WC_Tests_API_Onboarding_Profiles::test_update_profile_items - an array that should be empty but isn't.

Looks like the same bug we hit before; hopefully a quick rebase should fix that up.

jeffstieler:
@becdetat @joshuatf rebased - looks like that fixed the tests.

becdetat:
@joshuatf 

> Do you mean renaming the first step to "Store address" and then the route would change to store-details? 

Yes, renaming the first step to "Store address" but changing the step from `step=store-details` to `step=store-address`.

I don't think that's required though, it could be done in a follow-up.

joshuatf:
To me this is a bit confusing that we have "Store details > Store details" and I can see this leading to more confusion down the road when referring to the steps or the profiler.

I also feel (being very nitpicky) that semantically, "Store Details > Benefits (plugins)" is not really a store "detail" as much as it is part of the "setup process."

But if we feel strongly that we want to rename the onboarding process to "Store details" then I'd prefer to go ahead and rename the first step in this PR as well. /cc @jameskoster if you have any feelings one way or another on this

jameskoster:
It's a little tricky for me to follow exactly what is happening here. But as far as naming goes, now that we have clearer lines between the profiler and the task list, I think it might make sense to name the profiler "Setup wizard".

jeffstieler:
Cool - thanks @jameskoster. Should we rename the task in the Task List to be "Setup Wizard" instead of "Store Details" as well?

jameskoster:
> Should we rename the task in the Task List to be "Setup Wizard" instead of "Store Details" as well?

I think that makes sense, yes.

psealock (approved):
Nice improvement here! Code looks good and UI working as expected.

PR #4972, opened 2020-08-12
Fix dependency declarations

psealock:
The `dependency-group.js` custom eslint rule requires only Internal and External dependency blocks and requires Internal dependencies to have relative paths.

This PR removes `WooCommerce dependencies` block, changes aliased paths from the `client` folder to relative paths, and removes the webpack alias to the `client` folder.

1. See the linter pass with 254 warnings and no errors.
2. Run `npm start` and ensure no errors are found in the logs.
2. Poke around the app looking for regressions. The likeliest of suspects would be a badly written relative path. I wrote a script to perform the task, so it should be OK.

### Changelog Note:

- Dev: Adhere to dependency-group rule from WooCommerce's eslint plugin.


client/analytics/components/report-summary/index.js
```diff
@@ -6,10 +6,6 @@ import { Component } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import PropTypes from 'prop-types';
 
-/**
```
samueljseay, line 9: such a tiny thing and I don't think it matters a lot but, is it possible to strip the newline before the comment block as well? 😅 No pressure if its a pain.

webpack.config.js
```diff
@@ -151,7 +151,7 @@ const webpackConfig = {
 	},
 	resolve: {
 		extensions: [ '.json', '.js', '.jsx' ],
-		modules: [ path.join( __dirname, 'client' ), 'node_modules' ],
```
samueljseay, line 154: 👏

client/analytics/components/report-summary/index.js
psealock: Yeah I tried getting my replace all to work with a regex, but it was becoming too time consuming. Instead I fixed manually where I could. 

I still have my script, I bet I could re purpose it to find these.

webpack.config.js
samueljseay: Actually I think we can remove this option altogether now since its the default in webpack. Not a biggy though.

samueljseay (approved):
My suggestions are tiny and I don't really feel like they're critical. This tests well for me and is a nice clean up. Thanks for doing the leg work on this 🎉

client/analytics/components/report-summary/index.js
psealock: Woohoo, automation to the rescue. I remove in https://github.com/woocommerce/woocommerce-admin/pull/4972/commits/8dd7971777ff7007612d915e2b60f3ca60541709

PR #4977, opened 2020-08-12
Delete associated customer analytics data when user is deleted from WordPress.

### LINKED ISSUE #3810: User deletion from Wordpress does not remove entry from wp_wc_customer_lookup
When removing a user from Wordpress via the users section, it seems that Woocommerce Admin does not remove entries under wp_wc_customer_lookup

woocommerce_customers
wp_users

jeffstieler:
Fixes #3810.

This PR seeks to delete rows from `wc_customer_lookup` when WordPress users are deleted.

- Go to WooCommerce > Customers
- Choose any customer with a linked name column value and a "sign up" date
- Go to Users, search for the customer's name
- Delete user from results
- Go back to WooCommerce > Customers
- Verify the customer is no longer found

### Changelog Note:

Fix: handle user deletion.

### DIFF
--- a/tests/api/reports-customers.php
+++ b/tests/api/reports-customers.php
@@ -431,4 +431,47 @@ class WC_Tests_API_Reports_Customers extends WC_REST_Unit_Test_Case {
 		$this->assertFalse( CustomersDataStore::get_existing_customer_id_from_order( false ) );
 		$this->assertFalse( CustomersDataStore::get_or_create_customer_from_order( false ) );
 	}
+
+	/**
+	 * Test user deletion.
+	 */
+	public function test_user_deletion() {
+		wp_set_current_user( $this->user );
+
+		// Creating a customer should show up regardless of orders.
+		$customer = WC_Helper_Customer::create_customer( 'customer', 'password', 'customer@example.com' );
+
+		WC_Helper_Queue::run_all_pending();
+
+		$request = new WP_REST_Request( 'GET', $this->endpoint );
+		$request->set_query_params(
+			array(
+				'per_page' => 1,
+			)
+		);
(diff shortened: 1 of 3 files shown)


tests/api/reports-customers.php
```diff
@@ -431,4 +431,43 @@ public function test_customer_retrieval_from_order_bad_order() {
...
+			)
+		);
+		$response  = $this->server->dispatch( $request );
+		$customers = $response->get_data();
+
+		$this->assertEquals( 200, $response->get_status() );
+		$this->assertCount( 1, $customers );
+		$this->assertEquals( $customer->get_id(), $customers[0]['user_id'] );
+
+		// Delete the user associated with the customer.
```
psealock, line 459: Where is the user getting deleted? It obviously is because the test is passing, so wondering what I'm missing.

psealock (approved):
Nice, this is looking good. Tests pass, testing is good as well. I checked out an order from a deleted customer and that looks good too.

Screen Shot 2020-08-13 at 12 04 15 PM

tests/api/reports-customers.php
jeffstieler: Woah this is some voodoo.. I accidentally removed the line calling `wp_delete_user()` in https://github.com/woocommerce/woocommerce-admin/pull/4977/commits/d715efe283fea4c1f183677411f3c5206fdde084#diff-6b0548e1900dcb4006b4abbd3929eb97L460...

jeffstieler: Fixed that up in 43325e4.

becdetat:
Cherry picked to 1.5.0 (to be released in 1.5.0-rc.4) (@jeffstieler)

PR #4978, opened 2020-08-12
Don't directly import from React, use @wordpress/element.

jeffstieler:
I noticed that some of our app files were directly importing from `react`, not `@wordpress/element`. This PR remedies that.

- Verify that the DevDocs still work (full list and single component)
- Verify that the Stats Overview home screen widget Jetpack CTA still functions (should be hidden if you have the user preference)
- Verify that the store address step of the OBW still functions

### Changelog Note:

N/A


psealock (approved):
All looking good, thanks for the cleanup @jeffstieler 🚢

PR #4979, opened 2020-08-13
Adjust renovate to automerge devDependencies

samueljseay:
To lessen the burden of addressing renovate PRs, this automerges all renovate PRs for `devDependencies` (of which a significant amount of renovate PRs are.) Note that the PR checks must still pass before renovate merges them.

Currently we check out renovate PRs for dev dependencies locally and basically just run tests anyway, so we can rely on the PR builds passing for this instead. This is always something we can change again in future if we run into trouble, but this should save some time for developers.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -5,5 +5,11 @@
 	"schedule": [ "before 3am on wednesday" ],
 	"composer": {
 		"enabled": false
-	}
+	},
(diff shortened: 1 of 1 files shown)


psealock (approved):
I like this change, especially since we can easily revert if it causes issues.

PR #4980, opened 2020-08-13
Stock Report: Fix bad text domains

psealock:
Fixes violations of `@wordpress/i18n-text-domain` and turns the rule back on.

Visit the Stock Report and see labels in the status column appear as they should.

### Changelog Note:

Dev: Fix invalid text domains in Stock Report

### DIFF
--- a/client/analytics/report/stock/table.js
+++ b/client/analytics/report/stock/table.js
@@ -155,19 +155,19 @@ class StockReportTable extends Component {
 				value: formatValue( currency, 'number', products ),
 			},
 			{
-				label: __( 'out of stock', outofstock, 'woocommerce-admin' ),
+				label: __( 'out of stock', 'woocommerce-admin' ),
 				value: formatValue( currency, 'number', outofstock ),
 			},
 			{
-				label: __( 'low stock', lowstock, 'woocommerce-admin' ),
+				label: __( 'low stock', 'woocommerce-admin' ),
 				value: formatValue( 'currency, number', lowstock ),
 			},
 			{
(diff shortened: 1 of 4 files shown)


client/analytics/report/stock/table.js
```diff
@@ -155,19 +155,19 @@ class StockReportTable extends Component {
...
 			},
 			{
-				label: __( 'out of stock', outofstock, 'woocommerce-admin' ),
+				label: __( 'out of stock', 'woocommerce-admin' ),
 				value: formatValue( currency, 'number', outofstock ),
 			},
 			{
-				label: __( 'low stock', lowstock, 'woocommerce-admin' ),
+				label: __( 'low stock', 'woocommerce-admin' ),
 				value: formatValue( 'currency, number', lowstock ),
```
samueljseay, line 163: I just noticed, this line.. looks like a bug 😄 this should be `'currency', 'number'` right? I don't know how this works because looking at the function it would likely only return null or undefined based on these inputs: https://github.com/woocommerce/woocommerce-admin/blob/main/packages/number/src/index.js#L49

samueljseay (approved):
Looks good. We should tackle that little bug before its forgotten (maybe in a follow up PR?)

client/analytics/report/stock/table.js
psealock: Yikes! Good catch. Fixed in https://github.com/woocommerce/woocommerce-admin/pull/4980/commits/a8cdc5993ec6f42d14f48b4379a21aa4e89e7779

PR #4982, opened 2020-08-13
Migrated "import" store to "wp.data"

### LINKED ISSUE #3382: Data Layer: Migrate imports store to wp.data

octaedro:
Fixes #3382

This PR migrates the `import` store to `wp.data`.

Also added error handling.

### Screenshots
screenshot-one wordpress test-2020 08 13-11_40_08

- Go to `WooCommerce` | `Analytics` | `Settings` (URL: `/wp-admin/admin.php?page=wc-admin&path=/analytics/settings`)
- If you need to add orders to test you should turn `WC-Admin` off, add a new order and turn it on again.
- The count of `customers` and `orders and refunds` available to import should be visible.
- Verify no HTTP/JS errors occur.
- Press `Start`.
- Everything should work normally.

Re-import data
- After the process finishes, it should show the button `Re-import data`. Press it.
- Start the process again.
- Refresh the browser.
- The button `Stop` should be visible until the importation finishes.

_Be careful when you press `Delete Previously Imported Data`. It could show a false positive since it's a process that runs under the hood_

### Changelog Note:
Dev: migrate import store to wp.data (don't include this in plugin changelog)


psealock (commented):
Looking good @octaedro, this is quite the refactor!

I noted a few things, and found some items in testing but I'm not sure if they how they relate to these changes:

I had more orders than the total, which led to an odd UI
Screen Shot 2020-08-24 at 9 45 42 AM

Here is a shot of the API response showing the same thing.
Screen Shot 2020-08-24 at 9 59 15 AM

It also showed `is_importing: false` and the UI said "Finalizing" with the spinner showing, but it remained that way for a long time and continued to poll the server until I refreshed the page. I would have expected it to stop and show some kind of indication of success.

"Delete Previously Imported Data" didn't really do anything until I refreshed the page some time later. I don't know if thats a different issue.

I also got this warning:

```
Property setImportFinished returned from dispatchMap in useDispatchWithMap must be a function.`
```

Finally, when the button received the `disabled` prop the UI doesn't reflect that, so it can be misleading to the user because the button looks enabled. This obviously is a separate issue to tackle later. Issue here: https://github.com/woocommerce/woocommerce-admin/issues/5046

client/analytics/settings/constants.js
```diff
@@ -0,0 +1,13 @@
...
+import { SECOND } from '@fresh-data/framework';
+
+export const QUERY_DEFAULTS = {
+	pageSize: 25,
+	period: 'month',
+	compare: 'previous_year',
+	noteTypes: 'info,warning,marketing,survey',
+};
+
+export const TIMEOUT = 3 * SECOND;
```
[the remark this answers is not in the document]
octaedro: Good idea! The constants were moved in the commit 2b79e38

client/wc-api/wc-api-spec.js
```diff
@@ -21,10 +19,7 @@ function createWcApiSpec() {
...
 				}
 
-				return [
-					...imports.operations.read( resourceNames ),
-					...items.operations.read( resourceNames ),
-				];
+				return [ ...items.operations.read( resourceNames ) ];
 			},
 			update( resourceNames, data ) {
 				return [ ...items.operations.update( resourceNames, data ) ];
```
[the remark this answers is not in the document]
octaedro: Would removing the file `client/wc-api/wc-api-spec.js` generate some problems [here](https://github.com/woocommerce/woocommerce-admin/blob/main/client/wc-api/wp-data-store/index.js#L17)?

octaedro:
Thank you @psealock for your review. I answered the questions you asked below.

> I had more orders than the total, which led to an odd UI
Screen Shot 2020-08-24 at 9 45 42 AM

I couldn't replicate this issue. I tried with 20 orders and 20 customers and everything worked ok.

> It also showed `is_importing: false` and the UI said "Finalizing" with the spinner showing

This is a preexisting issue, I solved it in the commit d736c7f

> but it remained that way for a long time and continued to poll the server until I refreshed the page.

Weird, this process usually takes a lot of time but I never experienced something like that. Did refreshing the page fix the problem for you?

> I would have expected it to stop and show some kind of indication of success.

The current behavior is inherited from the old version. When the importation process finishes the user will see `Status: All historical data imported` and only the buttons `Re-import data` and `Delete Previously Imported Data` should be visible.

> "Delete Previously Imported Data" didn't really do anything until I refreshed the page some time later. I don't know if thats a different issue.

This is also inherited. It's working but it doesn't tell the user what is going on in the API side.

> I also got this warning:
> Property setImportFinished returned from dispatchMap in useDispatchWithMap must be a function.

Solved in the commit bee9f92

> Finally, when the button received the `disabled` prop the UI doesn't reflect that, so it can be misleading to the user because the button looks enabled. This obviously is a separate issue to tackle later. Issue here: https://github.com/woocommerce/woocommerce-admin/issues/5046

I solved this problem in the commit aada1d3

client/analytics/settings/historical-data/index.js
```diff
@@ -2,90 +2,84 @@
...
-import withSelect from '../../../wc-api/with-select';
 
 class HistoricalData extends Component {
 	constructor() {
 		super( ...arguments );
 
 		this.dateFormat = __( 'MM/DD/YYYY', 'woocommerce-admin' );
+		this.intervalId = -1;
+		this.lastImportStopTimestamp = 0;
+		this.cacheNeedsClearning = true;
```
becdetat, line 29: typo

client/analytics/settings/historical-data/layout.js
```diff
@@ -119,88 +107,86 @@ class HistoricalDataLayout extends Component {
...
-	} = getImportStatus( lastImportStartTimestamp, requirement );
+	} = getImportStatus( lastImportStartTimestamp );
 	const { imported: customersProgress, total: customersTotal } =
 		customersStatus || {};
 	const { imported: ordersProgress, total: ordersTotal } = ordersStatus || {};
-	const isStatusLoading = isGetImportStatusRequesting(
-		lastImportStartTimestamp
+
+	const isError = Boolean(
+		getImportError( lastImportStartTimestamp ) ||
```
becdetat, line 141: is this `a || a`?

client/analytics/settings/historical-data/layout.js
```diff
@@ -119,88 +107,86 @@ class HistoricalDataLayout extends Component {
...
 	);
 	if ( hasImportStarted ) {
 		onImportStarted();
 	}
+
 	const hasImportFinished = Boolean(
-		! isStatusLoading &&
-			inProgress &&
+		inProgress &&
+			! cacheNeedsClearning &&
```
becdetat, line 154: same typo as above

becdetat (approved):
Everything looked fine in the first step until it seemed to get stuck when almost completed importing orders and refunds. When I refreshed the page everything was zero-ed out but the re-import data button wasn't present so I couldn't test the second step.

Assuming that this is just a quirk with my setup (I have thousands of customers and orders) this looks fine.

client/analytics/settings/historical-data/style.scss
```diff
@@ -112,4 +112,7 @@
 .woocommerce-settings-historical-data__actions {
 	align-items: center;
 	display: flex;
+	button:disabled {
+		opacity: 0.5 !important;
+	}
```
psealock, line 117: This works great, but I think should be removed because its a problem across all wc-admin. Issue to track here: https://github.com/woocommerce/woocommerce-admin/issues/5046

psealock (commented):
Changes look good @octaedro ! Nice work. I left a comment about the CSS for disabled button, otherwise looking good.

> Would removing the file client/wc-api/wc-api-spec.js generate some problems here?

yes probably. I deleted the entire `client/wc-api` folder to see what would happen. Theres a lot of conflicts with code still using parts of wc-api. Once this is merged, I think it would good to address those in a follow up.

psealock (approved):
That last comment was meant to be an approval 🎉

client/analytics/settings/historical-data/index.js
octaedro: Good eyes @becdetat! I corrected it in the commit 2637750

client/analytics/settings/historical-data/layout.js
octaedro: Nice catch! This was fixed in the commit e417088

client/analytics/settings/historical-data/style.scss
octaedro: Thank you @psealock, got it!
This was removed in the commit b481d41

octaedro:
As all the suggested changes were addressed I'll proceed to merge this PR.

PR #4984, opened 2020-08-13
Add filter to allow modification of report columns

### LINKED ISSUE #4147: Filters to alter the columns used by a DataStore class

joshuatf:
Fixes #4147

Adds a filter to all report columns to allow adding or modifying existing columns at the database level.

@daigo75 Could you let me know if this adequately covers your use case?

Note that you may need to flush your transients cache if the report columns aren't updated in the response.

1. Add a filter and add or change existing columns:
```php
		add_filter( 'woocommerce_admin_report_columns', function( $columns, $context, $table_name ) {
			if ( 'orders' === $context ) {
				$columns['my_column'] = "RAND(2) as my_column";
				$columns['net_total'] = "({$table_name}.net_total * 2) as net_total";
			}
			return $columns;
		}, 10, 3 );
```
2. Note the API response includes your changes.  E.g., `wp-json/wc-analytics/reports/orders`.
3. If you modified an existing column or have filters in place for the reports table on the JS side, note the changes in your table.

### DIFF
--- a/src/API/Reports/DataStore.php
+++ b/src/API/Reports/DataStore.php
@@ -97,12 +97,15 @@ class DataStore extends SqlQuery {
 	public function __construct() {
 		self::set_db_table_name();
 		$this->assign_report_columns();
-		$this->report_columns = apply_filters(
-			'woocommerce_admin_report_columns',
-			$this->report_columns,
-			$this->context,
-			self::get_db_table_name()
-		);
+
+		if ( property_exists( $this, 'report_columns' ) ) {
+			$this->report_columns = apply_filters(
+				'woocommerce_admin_report_columns',
+				$this->report_columns,
+				$this->context,
+				self::get_db_table_name()
+			);
+		}
 	}
 
 	/**
--- a/src/API/Reports/DataStore.php
+++ b/src/API/Reports/DataStore.php
@@ -97,6 +97,12 @@ class DataStore extends SqlQuery {
 	public function __construct() {
 		self::set_db_table_name();
 		$this->assign_report_columns();
+		$this->report_columns = apply_filters(
+			'woocommerce_admin_report_columns',
+			$this->report_columns,
+			$this->context,
+			self::get_db_table_name()
+		);
 	}
 
 	/**


daigo75:
@joshuatf Thanks for the update. I'm trying to understand how this would work. Would that filter be triggered by every report, and the `$context` variable indicating which report is being generated? That is, it would become `orders`, `products`, etc, depending on the report?

joshuatf:
>That is, it would become orders, products, etc, depending on the report?

That's correct, @daigo75.  This filter would give you access to all of the columns, context (report name), and the database table name, allowing the columns to be modified or add additional columns.  There's an example above that demonstrates this.

Feel free to ping me with any questions you might have about this. 😄

daigo75:
Thanks, this looks suitable. How do I build this new version for testing? Is there a branch I can check out and build?

joshuatf:
Hi @daigo75, you can checkout this branch `add/4147` and then run `composer install` and `npm i && npm run build` to test this branch out.

After we get things approved here, we can also have this ready to use in the next version of WooCommerce Admin.

timmyc:
@daigo75 just wondering if you have had a chance to test this out yet. Want to make sure this provides the proper filters you need in your extension prior to us shipping it.

daigo75:
I didn't have the time to run the tests, yet, I will do that as soon as I can. Thanks for checking.

psealock (approved):
This certainly is a powerful filter, especially when things like `woocommerce_analytics_clauses_select_orders_subquery` don't suffice. Do you think it has potential for misuse? Is there any way to mitigate that?

This is testing well and working as expected.

daigo75:
@psealock all filters and actions have potential for misuse. That's a conversation we already had when I first created the related issue. The idea of a filter is to allow 3rd parties to alter some element. They could replace it with utter garbage and bring the whole site down, but that's an inherent risk with any filter.

I don't think there is any need, nor possibility to mitigate the chances of misuse. If you tried to restrict what the filter can return, that would limit the filter's usefulness and introduce a potential maintenance nightmare as well on your side. For example, in my case the returned columns would be the same original ones, but generated differently (i.e. in the way needed to prepare a correct report). Short of building and running the while SQL to check if the columns are there, there would be no way to know if the output returned by a filter makes sense. 

The bottom line is that, once a developer uses a filter, the risk of such filter breaking something will always be there. That's just how things are. ☺

joshuatf:
>Do you think it has potential for misuse? Is there any way to mitigate that?

I do think that there is some potential for misuse, but hopefully no more than already exists.  We have filters for adding in column data prior to this, so this doesn't expose further risk there.  On the database side, we are granting access to column data, but I believe this to be mostly a matter of convenience as this could be done outside if this hook if someone wanted.

>The bottom line is that, once a developer uses a filter, the risk of such filter breaking something will always be there. That's just how things are. ☺

Overall, I agree with this sentiment and we generally want to err on the side of extensibility.  But there are definitely filters and hooks that have more exposure and are more likely to be abused, so @psealock's question is a good one to double check before merging something like this in.

daigo75:
@josuhatf perhaps it's just a different perception. I rarely object to adding a filter when it can make a solution more extensible, as that's where a lot of the "WordPress power" comes from. Of course, I wouldn't put filters everywhere, because processing them adds some overhead, but, personally, I'm not too concerned about what other developers do with them. If they want to mess things up, they will find a way. I lost count of the sites where the code in WP core or in some other plugin was modified...

daigo75:
@joshuatf I would like to confirm that the filter does exactly what I need. I'm now implementing the necessary code to extend the columns and running some tests.

Will the new version be part of WooCommerce 4.5?

timmyc:
@daigo75 really glad this will allow you to add support to your extension for wc-admin! This will probably ship out in WooCommerce 4.6 - so our release in early October.

PR #4986, opened 2020-08-14
Add updateOptions prop to Layout tests

becdetat:
A `updateOptions` prop was added to `Layout` as a required prop, but it wasn't added to the `Layout` tests, which causes a warning. This just adds a noop property to each time `Layout` is instantiated.

### Testing
Run the tests (`npm test`), there shouldn't be any warnings.


joshuatf (approved):
Thanks for these fixes, Bec!  Testing well and all 🟢 - LGTM :shipit:

PR #4990, opened 2020-08-14
Fix note data actions and consolidate query constants

joshuatf:
Fixes actions that were imported from `wc-api` instead of the data package.  Also consolidates constants into a single file.

1. Make sure no warnings are shown in the console.
1. Make sure you have some store alerts.
1. Trigger the action on the alert and make sure no errors occur.


psealock (approved):
Looks good, nice cleanup @joshuatf 🚢

PR #4992, opened 2020-08-17
Fix typo in reports store action creators.

jeffstieler:
`setReportItemsError()` and `setReportStatsError()` weren't defined in https://github.com/woocommerce/woocommerce-admin/pull/4966

- Verify that the action creators used in the `reports` data store exist

### Changelog Note:

N/A - unreleased change


joshuatf (approved):
Thanks for the fix!  LGTM 🚢

PR #4995, opened 2020-08-18
Remove composite product from 1.4 release notes

timmyc:
Quick revision to the 1.4.0 changelog - Thanks for catching that error @becdetat


becdetat (approved):
looks good

PR #4997, opened 2020-08-18
Orders Panel: get selectors from wc-api

psealock:
https://github.com/woocommerce/woocommerce-admin/pull/4966 mistakingly moved some selectors in the Orders Panel away from wc-api prematurely. This PR changes it back as that task will be handled in https://github.com/woocommerce/woocommerce-admin/issues/3381

Make sure the Orders Notification Panel does not produce a fatal error.

### Changelog Note:

none: unreleased bug


samueljseay (approved):
Works well for me. 🚀

PR #4998, opened 2020-08-18
Refactor task list status to onboarding data store.

### LINKED ISSUE #4662: Onboarding Data Store: Potential stale data with old getSetting/setSetting
I haven't found a bug in the UI, but there are remnants of the old onboarding settings. A refactor or change may expose a bug due to stale data, so we should pro-actively fix this.

https://github.com/woocommerce/woocommerce-admin/blob/4da38966f4bc5a08d390e228b9f30420170dc67e/client/task-list/tasks.js#L38

https://github.com/woocommerce/woocommerce-admin/blob/4da38966f4bc5a08d390e228b9f30420170dc67e/client/task-list/tasks/tax.js#L223

jeffstieler:
Fixes #4662.

This PR seeks to eliminate the multiple sources of truth for the onboarding task list status data. It removes calls to `setSetting()` for onboarding data, instead hydrating data from a new endpoint (`/wc-admin/onboarding/status`) and invalidating the resolution for that as an alternative to `setSetting()`.

- Verify all Task List functionality on a new store and an existing
- Look for oddities in the UX (perhaps the list item takes a while to be marked complete due to slow network?)

### Changelog Note:

Dev: add task list status data to onboarding store.


src/Features/OnboardingTasks.php
```diff
@@ -104,25 +96,55 @@ function( $gateway ) {
...
+	public function component_settings( $settings ) {
+		// Bail early if not on a wc-admin powered page, or task list shouldn't be shown.
+		if (
+			! \Automattic\WooCommerce\Admin\Loader::is_admin_page() ||
+			! \Automattic\WooCommerce\Admin\Features\Onboarding::should_show_tasks()
+		) {
+			return $settings;
+		}
+
+		// If onboarding isn't enabled this will throw warnings.
```
joshuatf, line 137: Quick note in case this gets merged first- onboarding will always be enabled as of https://github.com/woocommerce/woocommerce-admin/pull/4999 but that probably only affects this comment and the check for `onboarding` being set is still a good measure.

client/task-list/tasks/appearance.js
```diff
@@ -160,6 +156,7 @@ class Appearance extends Component {
 			method: 'POST',
 		} )
 			.then( ( response ) => {
+				clearTaskStatusCache();
```
joshuatf, line 159: This is a clever way of making sure the data gets accurately updated, but there does seem to be a small disconnect between these data stores and it feels hacky that we need to clear the resolution cache on a different store to get this working.

@samueljseay I think this may be similar to what you were you discussing in your p2 post; do you have any thoughts here?

src/API/OnboardingTasks.php
```diff
@@ -73,6 +74,19 @@ public function register_routes() {
 				'schema' => array( $this, 'get_public_item_schema' ),
 			)
 		);
+
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base . '/status',
```
joshuatf, line 80: Nice!  A much needed endpoint ❤️

joshuatf (commented):
This is a great refactor and improvement, Jeff.

Things look pretty good to me overall here, I'm just wondering if there's another way to handle the cache invalidation, but there may not be a reasonable one given how these settings are currently handled.

client/task-list/tasks/appearance.js
samueljseay: 🤔 I think regardless of cross store communication, the main issue again for me is complex async flows in components and by concerning the component with cache clearing we do just increase that coupling here. I think a code smell we have is where we put large amounts of store related code into the component as props and thats definitely not a redux-like approach. 

All that said, this is obviously a refactor born out of the most pragmatic path forward ❤️ . The only way to fix problems when you have our existing patterns is to continue importing more information from the store into the component, so I don't have any concerns here, except this component would be another one I'd want to spend some time refactoring later. 

cc @jeffstieler

joshuatf (approved):
>Look for oddities in the UX (perhaps the list item takes a while to be marked complete due to slow network?)

I didn't experience this first-hand, but I can definitely see it being an issue on slower hosts.  I think in a follow-up we should consider optimistically updating these settings.

>So is this good to merge then @joshuatf ?

This is testing well so I think this is good to merge as long as we outline some issues to tackle cross store communication and clearing up some of the resolution invalidation here.

I think long-term this will also hopefully be less of an issue as more stores are developed for these pieces, such as taxes and shipping, so we can mitigate use of the settings as a catch-all for setting storage.

Great work on this, Jeff.

jeffstieler:
Opened https://github.com/woocommerce/woocommerce-admin/issues/5031 to cover the changes we'd like to make around resolution invalidation, etc.

PR #5000, opened 2020-08-18
Add complete and incomplete task counts to tasklist_complete event

### LINKED ISSUE #4973: Add new event prop to wcadmin_tasklist_completed
With the latest changes in the task list / home screen, if a user wants to dismiss the task list he needs to click on "Hide this", which triggers `wcadmin_tasklist_completed`. The interaction and tracking are working fine, however it doesn't tell us much about if they're closing the task because they don't find it useful, if they already completed all the tasks, etc.

Adding an even prop that counts the number of completed tasks is enough to identify the store's state when the user closes the list.

Event prop: tasks_completed: [number of tasks]

joshuatf:
Fixes #4973

Adds incomplete and complete task counts so we can better understand reasons for hiding the task list.

### Screenshots
Screen Shot 2020-08-18 at 3 06 49 PM
Screen Shot 2020-08-18 at 3 06 28 PM

1. Enter `localStorage.setItem( 'debug', 'wc-admin:*' );` into your console. Leave your console open.
1. Enable the task list and visit it.
1. Hide the task list using the dropdown.
1. Note the events logged include counts for complete and incomplete

### DIFF
--- a/client/task-list/index.js
+++ b/client/task-list/index.js
@@ -203,19 +203,11 @@ class TaskDashboard extends Component {
 		} );
 	}
 
-	keepTaskCard() {
-		recordEvent( 'tasklist_completed', {
-			action: 'keep_card',
-		} );
-
-		this.props.updateOptions( {
-			woocommerce_task_list_prompt_shown: true,
-		} );
-	}
-
 	hideTaskCard( action ) {
 		recordEvent( 'tasklist_completed', {
 			action,
+			completed_task_count: this.getCompletedTaskKeys().length,
+			incomplete_task_count: this.getIncompleteTasks().length,
 		} );
 		this.props.updateOptions( {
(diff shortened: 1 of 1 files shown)


client/task-list/index.js
```diff
@@ -203,19 +203,11 @@ class TaskDashboard extends Component {
 		} );
 	}
 
-	keepTaskCard() {
```
timmyc, line 206: Was this method no longer being used?

timmyc (commented):
One minor question above, but this is testing out well for me:

image

client/task-list/index.js
joshuatf: Ya, it seems like it was mistakenly left in there from an older iteration of onboarding.  It's not currently called anywhere.

timmyc (approved):
Thanks for the clarification on that one item. LGTM.

PR #5002, opened 2020-08-18
Add the jest-environment tag to known tags for jsdoc lint

samueljseay:
The JS doc lint recognises the `jest-environment` header as a tag, and it errors because it is not a defined tag. The simplest way to avoid this is to add the tag to known tags in the eslintrc.

I've also added the tag to our reducer tests where possible which will speed these tests up and also prove that this new lint config works.


psealock (approved):
Changes look good and tests/linter are passing.

PR #5018, opened 2020-08-20
Fix components internal deps

### LINKED ISSUE #5006: woocommerce/components 5.0.0 "Plugin" component is tightly coupled to WooCommerce Admin code base
**Describe the bug**
`@woocommerce/components` 5.0.0 doesn't work stand-alone, because the `Plugin` component has some dependencies in wc-admin. Here's the build errors I see:

```
ERROR in ./node_modules/@woocommerce/components/build-module/plugins/index.js
    Module not found: Error: Can't resolve '@woocommerce/data' in '/Users/haszari/_automattic/merch-repos/woocommerce-blocks/node_modules/@woocommerce/components/build-module/plugins'
    
ERROR in ./node_modules/@woocommerce/components/build-module/plugins/index.js
    Module not found: Error: Can't resolve 'lib/notices' in '/Users/haszari/_automattic/merch-repos/woocommerce-blocks/node_modules/@woocommerce/components/build-module/plugins'
```

I took a peek and I see there is a wizzy new `Plugins` component. Looking at the docs/use case (wrangling/installing WP plugins), I'm not surprised this is tightly coupled to Woo. So maybe the easiest fix is to move this component out of the package.

https://github.com/woocommerce/woocommerce-admin/blob/fdeade1887cfe9daa91b6657a8a1ac5aeb084d8e/packages/components/src/plugins/index.js#L9-L15

**To Reproduce**
I haven't reproduced this outside WooCommerce Blocks repo. Here's suggested minimal reproduce steps, but I haven't tried this!

- Bootstrap an [empty react app](https://reactjs.org/docs/create-a-new-react-app.html#create-react-app) - e.g. `npx create-react-app woo-components-standalone-test`.
- Add & install `@woocommerce/components` dependency (`@5.0.0`).
- Import and use a Woo component.
- Build, should see build errors 🤞 

**Expected behavior**
`@woocommerce/components` can be used standalone, independent of WordPress or Woo.

### auto build test?
Long term it would be great if this type of issue could be caught by tooling. E.g. an automated travis build of a demo project (similar to reproduce scenario), or a magic dependency linter, or something even cooler.

- https://www.npmjs.com/package/dependency-check ?

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/5006

The `<Plugins />` component had internal dependencies to the wc-admin client which causes issues for anyone outside wc-admin trying to use the component.

This PR adds a `@woocommerce/data` dependency to the component's `package.json`. An `onResponse` prop was also created so consuming js app can choose to handle the response any way they'd like, ie create a notice.

1. Remove the WooCommerce Services plugin.
1. Enable the Task List.
2. Add the WooCommerce Services plugin via the Task List's Tax task.
3. See the plugin success notice created.

### Changelog Note:

- Fix: Remove internal dependencies from Plugins component.

### DIFF
--- a/packages/components/src/plugins/README.md
+++ b/packages/components/src/plugins/README.md
@@ -1,20 +1,24 @@
-Plugins
-===
+# Plugins
 
 Use `Plugins` to install and activate a list of plugins.
 
 ## Usage
 
 ```jsx
-<Plugins onComplete={ this.complete } pluginSlugs={ [ 'jetpack', 'woocommerce-services' ] } />
+<Plugins
+	onComplete={ this.complete }
+	pluginSlugs={ [ 'jetpack', 'woocommerce-services' ] }
+/>
 ```
 
 ### Props
 
-Name | Type | Default | Description
---- | --- | --- | ---
-`onComplete` | Function |  | Called when the plugin installer is completed
-`onSkip` | Function | `noop` | Called when the plugin installer is skipped
-`skipText` | String |  | Text used for the skip installer button
-`autoInstall` | Boolean | false | If installation should happen automatically, or require user confirmation
-`pluginSlugs` | Array | `[ 'jetpack', 'woocommerce-services' ],` | An array of plugin slugs to install.
+| Name          | Type     | Default                                  | Description                                                               |
+| ------------- | -------- | ---------------------------------------- | ------------------------------------------------------------------------- |
+| `onComplete`  | Function |                                          | Called when the plugin installer is completed                             |
+| `onError`     | Function |                                          | Called when the plugin installer completes with an error                  |
+| `onResponse`  | Function |                                          | Called when the plugin installer response returns                         |
+| `onSkip`      | Function | `noop`                                   | Called when the plugin installer is skipped                               |
+| `skipText`    | String   |                                          | Text used for the skip installer button                                   |
+| `autoInstall` | Boolean  | false                                    | If installation should happen automatically, or require user confirmation |
+| `pluginSlugs` | Array    | `[ 'jetpack', 'woocommerce-services' ],` | An array of plugin slugs to install.                                      |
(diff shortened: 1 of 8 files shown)


packages/components/src/plugins/README.md
```diff
@@ -1,20 +1,24 @@
...
 ### Props
 
-Name | Type | Default | Description
---- | --- | --- | ---
-`onComplete` | Function |  | Called when the plugin installer is completed
-`onSkip` | Function | `noop` | Called when the plugin installer is skipped
-`skipText` | String |  | Text used for the skip installer button
-`autoInstall` | Boolean | false | If installation should happen automatically, or require user confirmation
-`pluginSlugs` | Array | `[ 'jetpack', 'woocommerce-services' ],` | An array of plugin slugs to install.
+| Name          | Type     | Default                                  | Description                                                               |
```
joshuatf, line 16: Thanks for updating these 💯

packages/components/src/plugins/index.js
```diff
@@ -54,11 +50,11 @@ export class Plugins extends Component {
 
 		installAndActivatePlugins( pluginSlugs )
 			.then( ( response ) => {
-				createNoticesFromResponse( response );
+				onResponse( response );
```
joshuatf, line 53: I'm a little wary of introducing a new prop for the response when we have `onComplete` and `onError`.  What do you think about passing the full response to `onComplete` and `onError` so we can handle this without introducing a new prop?

joshuatf (commented):
Nice fix here, Paul!  Left one comment about possibly limiting overall surface area of this component.

psealock:
Thanks for the review @joshuatf !

>I'm a little wary of introducing a new prop for the response when we have onComplete and onError.

After sleeping on it, I came to the same conclusion. Instead https://github.com/woocommerce/woocommerce-admin/pull/5018/commits/e820534c60bea468be7679e3289502394da9dfa2 passes the full response to `onComplete` and `onError` so we don't introduce a new prop.

joshuatf (approved):
Testing well and code looks great!  This LGTM after fixing up conflicts ✅

Issue #4654, opened 2020-06-22
Banner - mobile app


pmcpinto:
**User story:**
As a merchant
I want to be informed about the existence of a mobile app
So I can decide if continue using the mobile browser or install the app

**Description:** the mobile team asked to implement a banner informing users about the existence of a mobile app, which is a common practice among our competitors

![screen-shot-2020-05-14-at-4.19.09-pm.png](https://images.zenhubusercontent.com/5d84fe8e1201de0001b6ab61/d10268a4-0de7-400b-9590-a782c92f4bd9)

Note: ignore the banner on wccom

**Acceptance Criteria:**
- [ ] Display the mobile banner below the 783px breakpoint (the same that exists for the Activity panels) 
- [ ] Dismiss the banner if the user clicks to install the app or dismiss the banner
- [ ] Redirect the user to https://woocommerce.com/mobile/ if they click to install the app

Nice to have
- [ ] Hide the banner from users that already clicked to install the app through the inbox notification
- [ ] Redirect the user to the [App Store](https://itunes.apple.com/us/app/id1389130815?pt=299112&ct=wcadmin_banner&mt=8) or [Google Play](https://play.google.com/store/apps/details?id=com.woocommerce.android&referrer=utm_source%3Dwcadmin_banner%26utm_medium%3Dpage%26utm_campaign%3Dmobile_app_launch), depending on the device OS

**Event tracking:**
Event name: `wcadmin_mobile_banner_click`
Event prop: `button`(values: `install` | `dismiss`)
Description: when the user clicks the mobile app banner

**Additional links:**
Design proposal: p91TBi-2LT-p2
[Design mockup](https://www.figma.com/file/mjrVSgBLeaJ5TOjQwplorK/Smart-app-banner?node-id=55%3A3723)

timmyc:
Since we are seeking to have this code used in two places ( Woo installs and WooCommerce.com ) it kind of _feels_ like we should write this as a standalone package once, and allow it to be used in both spots. I'm going to estimate this at a 5.

pmcpinto:
> Since we are seeking to have this code used in two places ( Woo installs and WooCommerce.com ) it kind of feels like we should write this as a standalone package once, and allow it to be used in both spots. I'm going to estimate this at a 5.

Should we check if the package can be implemented by Gerhard, as he's going to work on the woocommerce.com banner? That would decrease the effort on our side, right?

mattsherman:
[...]

PR #5023, opened 2020-08-21
Refactor the Header component from class to function.

samueljseay:
Working towards #4654 this refactors the `<Header>` component to be functional so that it can use hooks. The plan is to use the `useUserPreferences` hook there to determine if the mobile banner should be rendered or not.

Smoke test that the `<Header>` still works by checking:

* The styles and rendering look OK
* The navigation breadcrumb is rendered correctly and works as before
* The document title is managed properly as before when clicking through navigation links.

### Changelog Note:

Dev: refactor `<Header>` to be a functional component.

### DIFF
--- a/client/header/index.js
+++ b/client/header/index.js
@@ -2,10 +2,9 @@
  * External dependencies
  */
 import { __, sprintf } from '@wordpress/i18n';
-import { Component, createRef } from '@wordpress/element';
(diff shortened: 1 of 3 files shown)


joshuatf (approved):
Thanks for the refactor, Sam!  💯 

Left a couple quick optional additions, but this component is testing well and pre-approving.

PR #5028, opened 2020-08-24
Initial end to end tests

becdetat:
This is based on https://github.com/woocommerce/woocommerce-admin/pull/4343 - big thanks to @rrennick - I've created a new PR to start the review process fresh. If you use the A8C VPN you need to be connected to the VPN for tests to work correctly.

To run the tests:
```
npm i
npm run docker:up
npm run build
npm run test:e2e-dev
```
You can run a single test file like:
```
npm run test:e2e-dev profile-wizard.test.js
```


samueljseay:
@becdetat I just wanted to share some findings on the unit test failures, I'm digging in to them now, but in case you get to them first, I can reproduce the travis failures locally so it may help you find the problem.

If you do an `rm -fr node_modules`, then run `bin/js_lint_test.sh` they fail saying they can't find the i18n rule just like in travis. I'm working on finding out why now, but yeah in case I don't get to fix this, you could investigate locally.

samueljseay:
@becdetat Hope you don't mind, I pushed up a fix for the breaking unit tests.

jeffstieler:
In my testing, the Docker setup needed the `JETPACK_AUTOLOAD_DEV` config value to load the proper plugin - just add this to `tests/e2e/docker/initialize.sh`:

```sh
wp config set JETPACK_AUTOLOAD_DEV true --raw
```

The tests failed for me. Digging a little, it looks like some of the OBW tests fail in their assertions which causes the test code to exit without clicking the "continue" button, which causes all subsequent tests to fail as well.

Perhaps we need to figure out how to exit the test run on the first failure for the OBW?

samueljseay:
@jeffstieler that sounds very tricky. The closest thing I can think of is that we could use jest's [bail](https://github.com/woocommerce/woocommerce-admin/pull/5059) config to stop the E2E tests after the first failure. Maybe this is OK for CI? If we bail after 1 failure we save build time and I guess if you were debugging from there, you could turn bail off and run the e2e locally? 

It might make sense since as you pointed out, a bunch of failures could just be as a result of a test not doing what was expected.

Alternatively we need to write our tests in a way that one test does not rely on the state of the next. Which is obviously challenging. I'm not sure if its possible to do that 100% but maybe we could make improvements? In the case of the OBW I believe that we could do a navigate to the step instead of relying on `clickContinue` from the previous step?

becdetat:
Normally I would like each test to be isolated, however in this case the setup would be the entire suite to that point anyway  which would increase the test time and mean that tests failing at any point would still cascade to the other tests. Originally the OBW test was a single test several hundred lines long, so I was happy with just breaking it up into multiple cascading but smaller tests.

samueljseay:
@becdetat I agree it doesn't seem practical to try an isolate them in this case.

becdetat:
That is awesome, thank you so much @samueljseay.

Can someone else review and approve this please?

timmyc (approved):
This is just fantastic, such a delight to see this ready to merge in. Setup docker and ran the suite locally and all is well.

The only item I'm wondering about is as we continue the discussion around best practices for the e2e suite is that we make sure to get that list of suggestions in the README to keep them close to the code and discoverable for future folks. But lets get this landed because it is wonderful.

PR #5032, opened 2020-08-24
Fix Advanced Filters Screen Reader Text

jeffstieler:
While developing the Product Attribute filter, I noticed that the screen reader text was broken for all advanced filters.

The issue stems from the call to `interpolateComponents()` missing the `title` component. `interpolateComponents()` [will return the `mixedString` if there is a missing component](https://github.com/Automattic/interpolate-components/blob/master/test/test.jsx#L78-L89).

This PR fixes missing title component in advanced filter screen reader interpolation.

- Open a report with advanced filters (like Orders)
- Add an advanced filter and select some values
- Inspect the filter element and verify the `screen-reader-text` spans have intelligible strings without component placeholders like `{{title}}`, etc.

### Changelog Note:

Fix: `<AdvancedFilters />` screen reader text.


psealock (approved):
Thanks for the cleanup @jeffstieler . Code looks good and testing well.

PR #5037, opened 2020-08-25
Add WooCommerce Mobile App Banner Ad for Android and iOS

### LINKED ISSUE #4654: Banner - mobile app

samueljseay:
Fixes #4654 

The feature calls for a mobile app ad banner to be displayed to users on mobile devices.

Based on the discussion in #4654 this implements the following:

1. [an iOS Smart App Banner](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/PromotingAppswithAppBanners/PromotingAppswithAppBanners.html). This banner is a meta tag that is generated in PHP. It will only display on iOS devices. One note about this: **This tag is not directly trackable like the implemented Android banner. If we would like to track its success then I believe [these instructions are relevant](https://stackoverflow.com/questions/12906502/is-it-possible-to-track-click-throughs-from-iphone-smart-banner/20422334#20422334)**. cc @pmcpinto on that one.

2. A mobile app banner ad that **only displays on Android** and only displays at the `738px` breakpoint specified in the issue. To only display this banner ad on Android, we use basic checking of the user agent string. I weighed this up against other approaches and for this kind of niche use case a simple UA string check is (imho) still the best way to do this.

3. The banner ad makes use of user preferences to retain a per user setting that determines if that user has dismissed the Android banner. We don't/can't do anything like this for the iOS Smart App Banner (but in theory we shouldn't need to).

cc @jameskoster you may want to take a look at the design here.

### Accessibility

I haven't tested mobile accessibility of these features (they are only displayed on mobile devices).

### Screenshots

Screenshot 2020-08-25 at 4 09 18 PM
Screenshot 2020-08-25 at 4 16 35 PM

Make sure you've done a fresh `composer install` first.

##### iOS
1. To test the iOS banner, you'll need an iOS device. I used browserstack to test this.
2. Open Wordpress and confirm the banner is not displayed on non WooCommerce pages
3. Open any WooCommerce page and confirm the banner displays

##### Android
1. To test Android it is best to use an Android device. I used browserstack to test this.
2. Open Wordpress and confirm the banner is not displayed on non WooCommerce pages
3. Open any WooCommerce page and confirm the banner displays
4. Test install link which should deep link to the Play Store install page for WooCommerce
5. Confirm that after that link was clicked the banner is no longer displayed (and no longer displayed on subsequent page loads either)
6. Use a different user, or reset user preferences for the existing user (I don't have detailed instructions for this, I just do a full reset of my local environment, but I'm sure clearing it in the DB is possible too).
7. Confirm the banner is displayed again as per 2,3
8. Click the dismiss and confirm it is no longer displayed as per 5

### Changelog Note:

Enhancement: Add mobile banner ads informing users of the WooCommerce native apps.


client/header/index.js
```diff
@@ -33,8 +35,14 @@ export const Header = ( { sections, isEmbedded = false, query } ) => {
 	const rafHandle = useRef( null );
 	const threshold = useRef( null );
 	const siteTitle = getSetting( 'siteTitle', '' );
-	const _sections = Array.isArray( sections ) ? sections : [ sections ];
+	const _sections = useMemo(
```
samueljseay, line 38: Eslint warns about this needing memoization to avoid unnecessary re-renders. so this fixes that.

client/lib/platform/index.js
```diff
@@ -0,0 +1,18 @@
+export const ANDROID_PLATFORM = 'android';
+export const IOS_PLATFORM = 'ios';
+export const UNKNOWN_PLATFORM = 'unknown';
+
+/**
+ * Provide basic detection of platform based on user agent. This is not
+ * a robust check for browser features or the like. You should only use
+ * this for non-critical display logic.
+ */
+export const platform = () => {
```
samueljseay, line 10: Basic UA string checking for platform based on other faux smart app banner approaches I've seen.

client/mobile-banner/index.js
```diff
@@ -0,0 +1,92 @@
...
+
+const SHOW_APP_BANNER_MODIFIER_CLASS = 'woocommerce-layout__show-app-banner';
+
+export const MobileAppBanner = ( { onInstall, onDismiss } ) => {
+	useEffect( () => {
+		const layout = document.getElementsByClassName(
+			'woocommerce-layout'
+		)[ 0 ];
+
+		if ( platform() === ANDROID_PLATFORM ) {
```
samueljseay, line 25: I've tried to make this component as reusable as possible so we can still extract it for others to use such as wc.com. Although this logic is specific to us, it won't do anything if the `woocommerce-layout` element is not present. And as mentioned in the comments it may be able to go away when we refactor activity panel.

client/stylesheets/abstracts/_colors.scss
```diff
@@ -34,3 +34,6 @@ $button-border: darken($button, 20%);
 
 // Blues
 $core-blue-dark-900: #0071a1;
+
+// WooCommerce brand colors
+$woocommerce-brand-purple: #3c2861;
```
samueljseay, line 39: I've had to add a propietary color just for the brand purple here.

src/Features/MobileAppBanner.php
```diff
@@ -0,0 +1,49 @@
...
+			self::$instance = new self();
+		}
+		return self::$instance;
+	}
+
+	/**
+	 * Hook into WooCommerce.
+	 */
+	public function __construct() {
+		add_filter( 'woocommerce_admin_get_user_data_fields', array( $this, 'add_user_data_fields' ) );
```
samueljseay, line 32: A very thin feature class that just adds our user data field. This matches the patterns elsewhere and I couldn't think of a much better place to put it.. (unless it gets jammed in FeaturePlugin somewhere?) Keen for feedback on this.

jameskoster (approved):
I wasn't actually able to get the banner to appear using browserstack. Most likely an issue with my local setup.

Approving from the design side though - LGTM 👍

samueljseay:
@jameskoster I forgot to mention you will need to do a `composer install`

config/development.json
```diff
@@ -13,6 +13,7 @@
 		"store-alerts": true,
 		"unminified-js": true,
 		"wcpay": true,
-		"homescreen": true
+		"homescreen": true,
+		"mobile-app-banner": true
```
psealock, line 17: I didn't see a js check for `window.wcAdminFeatures[ 'mobile-app-banner' ]` or the PHP feature flag check, so these can be removed if we don't plan on using them.

psealock (commented):
Nice one @samueljseay ! I only tested this on my iPhone vi ngrok and the banner worked as expected: only on WC pages and respects the dismiss button.

I have a few somewhat minor comments.

client/mobile-banner/constants.js
```diff
@@ -0,0 +1,7 @@
+// The Play Store link is defined as an exported constant mainly for the sake of testing the Mobile App Banner.
+// It is nearly impossible to fake navigation in JSDOM 16, so exposing this link for mocking allows us to
+// avoid triggering a navigation.
+export const PLAY_STORE_LINK =
+	'https://play.google.com/store/apps/details?id=com.woocommerce.android';
+
+export const TRACKING_EVENT_NAME = 'wcadmin_mobile_banner_click';
```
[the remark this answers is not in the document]
samueljseay: @psealock good point. Will change that.

client/mobile-banner/test/index.js
```diff
@@ -0,0 +1,112 @@
...
+	} );
+
+	it( 'records a tracking event for install', () => {
+		const { queryByRole } = render(
+			<MobileAppBanner onInstall={ () => {} } onDismiss={ () => {} } />
+		);
+
+		fireEvent.click( queryByRole( 'link' ) );
+		expect( recordEvent ).toHaveBeenCalledWith( TRACKING_EVENT_NAME, {
+			button: 'install',
```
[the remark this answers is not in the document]
samueljseay: Yup that makes sense will fix.

config/development.json
samueljseay: @psealock this can't be removed because the Loader.php class loader depends on it to create feature plugins. There are a bunch of these that don't have a corresponding feature class either that get ignored, this could definitely be improved.

samueljseay: The class factory for reference: https://github.com/woocommerce/woocommerce-admin/blob/main/src/Loader.php#L233

If you trace it back it loads the feature set from these json files.

client/header/index.js
samueljseay: @psealock that seems reasonable. Will fix this.

client/mobile-banner/constants.js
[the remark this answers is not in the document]
samueljseay: Fixed in https://github.com/woocommerce/woocommerce-admin/pull/5037/commits/c861ec08312226b8fe6db24ab6361e183bb82137

client/mobile-banner/test/index.js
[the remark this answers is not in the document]
samueljseay: Fixed https://github.com/woocommerce/woocommerce-admin/pull/5037/commits/c861ec08312226b8fe6db24ab6361e183bb82137

client/header/test/index.js
```diff
@@ -10,6 +10,15 @@ jest.mock( '@woocommerce/tracks', () => ( {
 	recordEvent: jest.fn(),
 } ) );
 
+jest.mock( '@woocommerce/data', () => ( {
+	...jest.requireActual( '@woocommerce/data' ),
+	useUserPreferences: () => ( {
+		updateUserPreferences: () => {},
+		// mock to disable the mobile app banner while testing this component
+		android_app_banner_dismissed: true,
```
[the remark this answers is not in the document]
samueljseay: Fixed https://github.com/woocommerce/woocommerce-admin/pull/5037/commits/c861ec08312226b8fe6db24ab6361e183bb82137

client/header/index.js
samueljseay: Fixed in https://github.com/woocommerce/woocommerce-admin/pull/5037/commits/c861ec08312226b8fe6db24ab6361e183bb82137

pmcpinto:
> This tag is not directly trackable like the implemented Android banner. If we would like to track its success then I believe these instructions are relevant. cc @pmcpinto on that one.

Thanks for the heads up. Let me check if we're already part of that Apple's program

config/development.json
psealock: I'm wondering if we should add an array of features that have graduated to always being "on" so they don't have to be declared in the configs. Would you add that here or in a follow up?

psealock (approved):
The code is looking good and now I've tested on Android via Browserstack. The only thing I noticed is it is a bit cramped on narrow screens. We support as low as 320px.

Screen Shot 2020-08-27 at 12 09 56 PM

This issue and the question about having "always on" features remain. I'll pre-approve, but am happy to take another look if you'd like.

config/development.json
samueljseay: Sounds good. I think a follow up PR is ideal. Will ask you for some feedback on what list of features are always on.

Issue #4330, opened 2020-05-09
Filter Product and Order Reports by Variation 


gbruzzo:
**Is your feature request related to a problem? Please describe.**

We would like to be able to **filter reports by product variation**. 

Our store sells products as variations. We are a record label; the parent product is usually a 'release' (an album release); what is actually SOLD are variations (e.g. formats), so VINYL, CD, Download, etc. It is therefore crucial for us to be able to **filter all reports by Variation**

1. **Orders reports** -  One can only filter by Parent Product:  Not only one cannot filter by variation, but the table returned after filtering by a parent product contains parent product names, not variation names. (column : product(s)), as does the CSV. That is of little use. 

2. **Products reports** - One can only filter by Parent Product (in Single Product and Comparison, as well as in the explicit Product Filter): It is true that the table returned contains all individual variations of a given parent product; Alas the 'Order' columns links to the Orders reports, which then only shows the parents again. Again, of very little use, as are the CSV reports.

**Describe the solution you'd like**

In order for Analytics to be of any use, we would need to be able to 

- Explicitly filter by **individual Variations**  in the Product and Order Reports. 

- Explicity filter by **specific Attribute** in the Product and Order Reports (e.g. in our case filter by 'CD' or by 'VINYL' or 'DOWNLOAD') 

- **Alternatively :** create an **ATTRIBUTE REPORT** (to sit side by side with the Category report)

- **When filtering Orders Reports by a parent product:** the table returned should specify all Variations bought in the Product(s) column, not just the Product Parent Name. 

- **Alternatively :** **when filtering Orders Reports by a parent product:** the table returned should contain an extra column with all Variations of the parent product in a given order.

All this information should be reflected and available in .csv

**Describe alternatives you've considered**

The alternative for us is currently to not use Woocommerce admin (it is wholly redundant). 

**Should this be prioritized? Why?**

Because, as it stands, Woocommerce Admin is incomplete and of little use. It burdens installations without offering the depth required (drilldown by Variation and other types of products). 

We currently keep Woocommerce Analytics (and Marketing for that matter) switched off via a filter. 

**Additional context**

Similar points raised in : https://github.com/woocommerce/woocommerce-admin/issues/3665 

Order Report Screenshot with comments 
![Screenshot_2020-05-09_at_16_03_37](https://user-images.githubusercontent.com/16798202/81477495-f1f8ca80-920f-11ea-8ff3-94b1eeca2f99.jpg)

[...]

PR #5038, opened 2020-08-25
Add "Product Attribute" advanced report filter.

jeffstieler:
Partially implements #4330.

This PR introduced a new advanced filter for selecting product attributes. It will be added to the Orders report in a follow up.

### Accessibility

Admittedly, our report filters needs some attention again on the accessibility front. The filter being introduced in this PR is of the same quality as the others.

### Screenshots

Screen Shot 2020-08-25 at 9 46 54 AM

Add an `attribute` filter to the advanced filters config of any report:

```javascript


			attribute: {
				labels: {
					add: __( 'Attribute', 'woocommerce-admin' ),
					placeholder: __( 'Search attributes', 'woocommerce-admin' ),
					remove: __(
						'Remove attribute filter',
						'woocommerce-admin'
					),
					rule: __(
						'Select a product attribute filter match',
						'woocommerce-admin'
					),
					/* translators: A sentence describing a Product filter. See screen shot for context: https://cloudup.com/cSsUY9VeCVJ */
					title: __(
						'{{title}}Attribute{{/title}} {{rule /}} {{filter /}}',
						'woocommerce-admin'
					),
					filter: __( 'Select attributes', 'woocommerce-admin' ),
				},
				rules: [
					{
						value: 'is',
						/* translators: Sentence fragment, logical, "Is" refers to searching for products matching a chosen attribute. Screenshot for context: https://cloudup.com/cSsUY9VeCVJ */
						label: _x(
							'Is',
							'product attribute',
							'woocommerce-admin'
						),
					},
					{
						value: 'is_not',
						/* translators: Sentence fragment, logical, "Is Not" refers to searching for products that don\'t match a chosen attribute. Screenshot for context: https://cloudup.com/cSsUY9VeCVJ */
						label: _x(
							'Is Not',
							'product attribute',
							'woocommerce-admin'
						),
					},
				],
				input: {
					component: 'ProductAttribute',
				},
			},
```

- Go to the report
- Add the "attribute" filter
- Verify the function of selecting an attribute name and value - try different combinations of mouse/keyboard inputs and generally attempt to break things 😄 
- Verify that the URL query is updated when clicking "filter" - `attribute_is[0]=1&attribute_is[1]=2` the value should be a tuple of attribute ID and term ID

### Changelog Note:

Enhancement: add product attribute advanced report filter component.

### DIFF
--- a/client/analytics/report/products/config.js
+++ b/client/analytics/report/products/config.js
@@ -114,6 +114,32 @@ const variationsConfig = {
 			value: 'all',
 		},
 		{
+			label: __( 'Single Variation', 'woocommerce-admin' ),
+			value: 'select_variation',
+			subFilters: [
+				{
+					component: 'Search',
+					value: 'single_variation',
+					path: [ 'select_variation' ],
+					settings: {
+						type: 'variations',
+						param: 'variations',
+						getLabels: getVariationLabels,
+						labels: {
+							placeholder: __(
+								'Type to search for a variation',
+								'woocommerce-admin'
+							),
+							button: __(
+								'Single Variation',
+								'woocommerce-admin'
+							),
+						},
+					},
+				},
+			],
+		},
+		{
 			label: __( 'Comparison', 'woocommerce-admin' ),
 			chartMode: 'item-comparison',
 			value: 'compare-variations',
--- a/client/lib/async-requests/index.js
+++ b/client/lib/async-requests/index.js
@@ -82,14 +82,9 @@ export const getVariationLabels = getRequestByIdString(
 	( variation ) => {
 		return {
 			key: variation.id,
-			label: variation.attributes.reduce(
-				( desc, attribute, index, arr ) =>
-					desc +
-					`${ attribute.option }${
-						arr.length === index + 1 ? '' : ', '
-					}`,
-				''
-			),
+			label: variation.attributes
+				.map( ( { option } ) => option )
+				.join( ', ' ),
 		};
 	}
 );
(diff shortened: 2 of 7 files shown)


psealock (approved):
This is all looking good and testing well according to instructions. The data isn't actually filtered, but I think thats expected. The url does update accordingly.

Those filters are supposed to read like sentences, so my filter reads like so:

```
Attribute is color = blue
```
Would it be odd to rearrange things so it reads more naturally?

```
Attribute color is blue
```

jeffstieler:
@psealock 

> Would it be odd to rearrange things so it reads more naturally?

It wouldn't be odd at all, it would be much better! This weird grammar results from using the `config.labels.title` value for both rendering the input controls and the screen reader text. All of the filters use the same format: `{{title}}Filter name{{/title}} {{rule /}} {{filter /}}`.

To do what you're describing would need something like `{{title}}Attribute{{/title}} {{partial filter value /}} {{rule /}} {{rest of filter 
 value /}}`.

Are you OK with taking care of better screen reader text for all filters in a follow up?

jeffstieler:
Nevermind @psealock - I got it to work! This is ready for another review.

psealock (commented):
Are there missing commits?

I'm seeing the screen reader text display accurately, although there is an extra space after "Attribute".

```
Attribute  Color Is Blue
```
but the UI is not updated. I did a bit of digging and couldn't quickly figure out what the problem was.

Screen Shot 2020-08-27 at 12 57 44 PM

EDIT: the fix described above refers to the screen reader text. We can potentially address the UI in a follow up if need be.

jeffstieler:
@psealock the UI wasn't updated - it can't have the rule ("is") render in the middle of the filter (the 2 search controls) because it's controlled by the filter config "title" string.

psealock (approved):
Cool, it all LGTM 🚢

PR #5049, opened 2020-08-26
When toggling the task list, record a track event.

### LINKED ISSUE #4760: Add Tracks to enable/disable Task List button
**Description:** Add tracking to understand if the user is enabling/disabling the task list through the Help panel 
**Event name:** wcadmin_tasklist_toggled
**Even prop:** enable/disable

Captura de ecrã 2020-07-07, às 11 27 43

samueljseay:
Fixes #4760 

To test this, ensure that the task list toggle is still working correctly.

1. Go to `WooCommerce > Settings`
2. Click `Help` in the top right
3. Click `Setup Wizard` on the left side menu
4. The button will either say `Enable` or `Disable` depending on current state
5. Click the button, it should redirect to `WooCommerce > Home`
6. Go back to the settings as per steps 1-3.
7. The button should say `Enable` if it said `Disable` before and vice versa.

### Changelog Note:

Tweak: When toggling the task list, record a track event.

### DIFF
--- a/src/Features/Onboarding.php
+++ b/src/Features/Onboarding.php
@@ -1053,8 +1053,17 @@ class Onboarding {
 			return;
 		}
 
-		$new_value = 1 === absint( $_GET['reset_task_list'] ) ? 'no' : 'yes'; // phpcs:ignore CSRF ok.
+		$previous_value = 1 === absint( $_GET['reset_task_list'] ) ? 'yes' : 'no'; // phpcs:ignore CSRF ok.
+		$new_value      = 'no' === $previous_value ? 'yes' : 'no'; // phpcs:ignore CSRF ok.
+
 		update_option( 'woocommerce_task_list_hidden', $new_value );
+		wc_admin_record_tracks_event(
+			'wcadmin_tasklist_toggled',
+			array(
+				'previous'  => $previous_value,
+				'new_value' => $new_value,
+			)
+		);
 		wp_safe_redirect( wc_admin_url() );
 		exit;
 	}
--- a/src/Features/Onboarding.php
+++ b/src/Features/Onboarding.php
@@ -1053,15 +1053,13 @@ class Onboarding {
 			return;
 		}
 
-		$previous_value = 1 === absint( $_GET['reset_task_list'] ) ? 'yes' : 'no'; // phpcs:ignore CSRF ok.
-		$new_value      = 'no' === $previous_value ? 'yes' : 'no'; // phpcs:ignore CSRF ok.
+		$task_list_hidden = 1 === absint( $_GET['reset_task_list'] ) ? 'no' : 'yes'; // phpcs:ignore CSRF ok.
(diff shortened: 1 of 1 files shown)


src/Features/Onboarding.php
```diff
@@ -1053,8 +1053,17 @@ public static function reset_task_list() {
 			return;
 		}
 
-		$new_value = 1 === absint( $_GET['reset_task_list'] ) ? 'no' : 'yes'; // phpcs:ignore CSRF ok.
+		$previous_value = 1 === absint( $_GET['reset_task_list'] ) ? 'yes' : 'no'; // phpcs:ignore CSRF ok.
+		$new_value      = 'no' === $previous_value ? 'yes' : 'no'; // phpcs:ignore CSRF ok.
+
 		update_option( 'woocommerce_task_list_hidden', $new_value );
+		wc_admin_record_tracks_event(
```
samueljseay, line 1060: I've deviated from the issue spec here to match the other similar events we track in the Onboarding class. Keen for some feedback here on whether its ok to deviate or whether it should match the spec more exactly.

src/Features/Onboarding.php
```diff
@@ -1053,8 +1053,17 @@ public static function reset_task_list() {
...
+			'wcadmin_tasklist_toggled',
```
psealock, line 1061: You'll get a double prefixed event name, `tasklist_toggled` will suffice.

https://github.com/woocommerce/woocommerce/blob/251e1ea27305234220b9b7952b08cd7e0174adbf/includes/tracks/class-wc-tracks.php#L89

src/Features/Onboarding.php
psealock: I suppose its best to imagine you're digging through events to determine a percentage of people toggling the task list. I think the first SQL is easier to read:
```
WHERE eventname = 'wcadmin_tasklist_toggled' AND eventprop.status = 'enabled'
```
versus
```
WHERE eventname = 'wcadmin_tasklist_toggled' AND eventprop.new_value = 'yes'
```
Its subjective though, what do you think?

psealock (commented):
I haven't taken it for a spin yet, but here is some early feedback for ya.

src/Features/Onboarding.php
samueljseay: Yep I really just focussed on consistency here with existing code, but I think you're right, the former is a much more obvious query. I'll take this approach.

src/Features/Onboarding.php
samueljseay: Awesome thanks for spotting this. will fix.

samueljseay:
@psealock I've implemented changes based on your feedback here: https://github.com/woocommerce/woocommerce-admin/pull/5049/commits/ab6779c81363aeea723eceadba26dbc070f079b2

Keen to hear your thoughts.

psealock (approved):
Looking good and testing well, thanks @samueljseay 🚢

PR #5051, opened 2020-08-26
Fix a bug where an event's data was recorded in camel case.

### LINKED ISSUE #4464: Print Shipping Label step - event tracking missing
The event `wcadmin_tasklist_shipping_label_printing` was working properly a couple of months ago, but I was checking it on Tracks and I'm not able to get any info about it. It seems that it was removed.

I'm not sure if this is related to #4372

samueljseay:
Fixes #4464 

`wcadmin_tasklist_shipping_label_printing` events were being filtered into rejected events because the data being passed was in camel case instead of snake case. This fixes that issue.

1. Go through store onboarding, under store details choose "Australia" for the country
2. After finishing onboarding you should be at the WooCommerce home screen
3. Enable debug of tracking by adding this code in browser console: `localStorage.setItem( 'debug', 'wc-admin:*' )`
4. Choose the "Setup shipping" task from the task list
5. Go through the steps up to install "woocommerce shipstation".
6. Click install
7. You should see the `wcadmin_tasklist_shipping_label_printing` with correct snake case argument `plugins_to_activate` in the console.
8. Repeat steps 1-5, then click dismiss you should also see the results from `7` again with correct snake case argument `plugins_to_activate`

### Changelog Note:
Fix: fix a bug where `wcadmin_tasklist_shipping_label_printing` events were not properly tracked.

### DIFF
--- a/client/task-list/tasks/shipping/index.js
+++ b/client/task-list/tasks/shipping/index.js
@@ -253,14 +253,14 @@ class Shipping extends Component {
 						onComplete={ () => {
 							recordEvent( 'tasklist_shipping_label_printing', {
 								install: true,
-								pluginsToActivate,
+								plugins_to_activate: pluginsToActivate,
 							} );
 							this.completeStep();
 						} }
 						onSkip={ () => {
 							recordEvent( 'tasklist_shipping_label_printing', {
 								install: false,
-								pluginsToActivate,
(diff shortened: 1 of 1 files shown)


psealock (approved):
Very nice! This would make a good addition as a custom linter rule in our package.

PR #5053, opened 2020-08-27
add QuantitySelector component (from WooCommerce blocks)

haszari:
_In progress_

Fixes #

This PR adds the [`QuantitySelector` component from WooCommerce Blocks](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/main/assets/js/base/components/quantity-selector) into `@woocommerce/components` package.

This is a generally useful up/down control which allows the user to enter a number within a given range. In blocks this is used for tweaking the number of each line item in the cart. Could potentially be used in various other places in the front end and in admin too, and will hopefully be useful for clients of the components package.

This PR is a draft for now as the colours used in the styles are hacked in, I need to revise these to align with the other components. 

Also this needs to have a working demo soon - either storybook or the components docs webapp, or (?) use in wc-admin somewhere. I'm shooting for Storybook, but it looks like we need some maintenance on Storybook (saw issues locally) and there's a [new Storybook 6 out which could be good to adopt](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3041). (I'm keen to help with all of that once I've confirmed it's 👌 in blocks repo.) 🚆 

### Accessibility

- Ex: Open page `url`
- Click XYZ…

### DIFF
--- a/packages/components/src/index.js
+++ b/packages/components/src/index.js
@@ -34,6 +34,7 @@ export { default as Pill } from './pill';
 export { default as Plugins } from './plugins';
 export { default as ProductImage } from './product-image';
 export { default as ProductRating } from './rating/product';
+export { default as QuantitySelector } from './quantity-selector';
 export { default as Rating } from './rating';
 export { default as ReportFilters } from './filters';
 export { default as ReviewRating } from './rating/review';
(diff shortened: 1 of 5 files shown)


becdetat:
HI @haszari , I'm portering and just wondering if this is still needed or if we can close the PR for now?

haszari:
@becdetat Happy for this to close - not urgent or a priority for me. We can add/migrate more components over time, this, PR is very stale 👍

becdetat:
Thanks @haszari :-)

PR #5019, opened 2020-08-20
Profile wizard style consistency updates

jameskoster:
Please forgive me @joshuatf. I fixed up the Product Types step but kept finding style issues to fix that spilled into the other steps. It's mostly just a case of utilising the `Text` component where appropriate, replacing color studio vars with core equivalents, and a few spacing tweaks. I hope this isn't too painful to merge 😬

To test: Just work through the Profile Wizard and check for anomalies. I didn't touch anything functional so everything should still work as before.

PR #5059, opened 2020-08-30
Profile wizard style consistency updates

samueljseay:
This is a rebase of #5019 I felt it was safer to create a new branch since they're not my changes.

I've tested the changes and they look good for me locally but @jameskoster could you please pull this down and test it. @joshuatf perhaps you should take a look as well.

I've updated the snapshot test because the rendering has changed (I think maybe we shouldn't have snapshot tests on internal components anyway because they tend to churn a lot).

To test I focussed most of my time around the monthly pricing toggle and all the styling in the profile wizard.

If this looks OK, then we can close #5019


client/profile-wizard/steps/product-types/index.js
```diff
@@ -100,15 +105,17 @@ export class ProductTypes extends Component {
...
-					{ __( 'Choose any that apply' ) }
-				</H>
+				<div className="woocommerce-profile-wizard__step-header">
+					<Text variant="title.small" as="h2">
+						{ __(
+							'What type of products will be listed?',
+							'woocommerce-admin'
+						) }
+					</Text>
+					<Text vairant="body">
```
becdetat, line 115: Typo here

samueljseay: wow thanks for spotting this.

samueljseay: Fixed here: https://github.com/woocommerce/woocommerce-admin/pull/5059/commits/232b98c1e2656973e4654ed562b72592d9f52763

joshuatf (commented):
Changes look good to me.  Will leave final ✅ to @jameskoster

joshuatf:
>I've updated the snapshot test because the rendering has changed (I think maybe we shouldn't have snapshot tests on internal components anyway because they tend to churn a lot).

Ya, I think you're right.

jameskoster (approved):
:shipit:

PR #5068, opened 2020-09-02
Add product attribute filter to Orders Report.

jeffstieler:
Partially implements #4330.

This PR adds the Product Attribute advanced filter to the Orders Report. It handles the query parameters on the PHP side, modifying the orders and order stats queries appropriately.

- Ensure you have orders placed for product variations
- Go to Analytics > Orders
- Add an attribute filter (or several)
- Verify the results - check the stats numbers, summary numbers, and table values

### Changelog Note:

Enhancement: add Product Attribute filter to Orders Report.

### DIFF
--- a/client/analytics/report/orders/config.js
+++ b/client/analytics/report/orders/config.js
@@ -310,6 +310,50 @@ export const advancedFilters = applyFilters(
 					getLabels: getTaxRateLabels,
 				},
 			},
+			attribute: {
+				allowMultiple: true,
+				labels: {
+					add: __( 'Attribute', 'woocommerce-admin' ),
+					placeholder: __( 'Search attributes', 'woocommerce-admin' ),
+					remove: __(
+						'Remove attribute filter',
+						'woocommerce-admin'
+					),
+					rule: __(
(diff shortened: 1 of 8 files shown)


becdetat (approved):
I'm getting some failures when running the tests locally that look like they could be related to this:

```
1) WC_Tests_API_Reports_Categories::test_get_reports
Failed asserting that 8 matches expected 4.

/srv/www/wordpress-two/public_html/wp-content/plugins/woocommerce-admin/tests/api/reports-categories.php:82

2) WC_Tests_API_Reports_Categories::test_get_reports_categories_param
Failed asserting that 8 matches expected 4.

/srv/www/wordpress-two/public_html/wp-content/plugins/woocommerce-admin/tests/api/reports-categories.php:145
```

however those tests are passing in CI so it is probably something to do with my environment.

Apart from that it works beautifully and the code looks great :-)

jeffstieler:
Those tests also fail for me locally on occasion, but not in CI - I suspect they are faulty in the same way as these: https://github.com/woocommerce/woocommerce-admin/pull/4719

PR #5075, opened 2020-09-02
Remove wc api 🎉

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/3387

Remove wc-api, including imports of `withSelect` and various constants.

1. Boot up the app and make sure webpack is happy.
2. Poke around and make sure no fatal errors occur.

### Changelog Note:

- Dev: remove wc-api in favor of `@woocommerce/data`


samueljseay:
@psealock I haven't tested this yet, but after a global search locally I notice that after this removal there is one more mention of the fresh data framework, to import a constant `SECOND`, which is just a variable with value 1000 😁

Could we as part of this remove that one import and also delete fresh data from the package.json? Or in a follow up perhaps. I don't know if fresh data gets tree shaken but if it doesn't, removing it would save us on payload.

psealock:
>Could we as part of this remove that one import and also delete fresh data from the package.json?

Thanks for finding that @samueljseay, fixed in https://github.com/woocommerce/woocommerce-admin/pull/5075/commits/7de71f5a6d963c6a26fb8aae098f044bf35577ae

becdetat (approved):
Looks great 👍

PR #5076, opened 2020-09-02
Automated taxes smart default

### LINKED ISSUE #4594: Task List - tax task: automated taxes smart default
**User story:**
As a new merchant
I want to focus on the most important setup tasks
So I can review some details later

**Description:** with Jetpack installed and connected, alongside WCS, we have the opportunity to set things up on the users' behalf, and ask them to review and confirm, rather than start from scratch and think.

**Acceptance Criteria:**
- [ ] If the store has JP installed and connected, WCS active and is located in a country covered by automated taxes, automatically turn on automated taxes and mark the step as completed
- [ ] Trigger the following notification:
**Title:** Confirm tax settings
**Content:**  Automated tax calculations are enabled on your store through WooCommerce Services. Learn more about automated taxes [here](https://docs.woocommerce.com/document/woocommerce-services/#section-10)
**Segmentation:** stores with automated taxes automatically enabled
**Timing:** when the task is marked as completed 
**Cadence:** once
**CTA leads to:** wp-admin/admin.php?page=wc-settings&tab=tax
**CTA label:** Edit tax settings

**Event tracking:**
We can use the event `wcadmin_tasklist_task_completed` and eventually add a new value to `task_name` when this happens: tax_automated

**Additional links:**
[Design mockup](https://www.figma.com/file/JH9XMFUCOjfXdr3N09AHRD/On-boarding-iterations-June-'20?node-id=1%3A9)

samueljseay:
Fixes #4594

This opts in to automated tax calculation for the user when they complete the profiler if the following conditions are met:

1. Jetpack is installed and connected
2. WCS is installed and TOS accepted
3. You're in a country where automated tax is supported

Originally I looked at solving this in the client, but found that there was an existing pattern for running code after profiler completion which @becdetat implemented in https://github.com/woocommerce/woocommerce-admin/pull/4857

This approach ended up being quite a lot simpler with just one caveat: When completing the profiler in the client we have to force the onboarding task status cache to be invalidated to ensure that when we do a client side navigate to the home page, the status of the tax task will correctly update.

### Accessibility
n/a

This one is a hassle to test as you'll need to:

1. Have Jetpack installed and connected
2. Have WCS installed and accept TOS
3. Have an uncompleted onboarding wizard
4. Once all these are met you can go through the onboarding wizard, choices don't matter too much
5. Once you finish the last step of the Wizard and are redirected to the Woo home page, you should see that the tax setup task is complete.

### Changelog Note:

Enhancement: opt in to automated taxes by default for users who Jetpack and WCS installed.

### DIFF
--- a/client/profile-wizard/index.js
+++ b/client/profile-wizard/index.js
@@ -217,7 +217,7 @@ class ProfileWizard extends Component {
 			updateNote,
 			updateProfileItems,
 			connectToJetpack,
-			clearSettingsCache,
+			clearTaskCache,
 		} = this.props;
 		recordEvent( 'storeprofiler_complete' );
 
@@ -235,14 +235,14 @@ class ProfileWizard extends Component {
 		}
 
 		updateProfileItems( { completed: true } ).then( () => {
+			clearTaskCache();
+
 			if ( shouldConnectJetpack ) {
 				document.body.classList.add( 'woocommerce-admin-is-loading' );
 
 				connectToJetpack(
 					getHistory().push( getNewPath( {}, '/', {} ) )
 				);
-
-				clearSettingsCache();
 			} else {
 				getHistory().push( getNewPath( {}, '/', {} ) );
 			}
@@ -341,8 +341,7 @@ export default compose(
 		} = dispatch( ONBOARDING_STORE_NAME );
 		const { createNotice } = dispatch( 'core/notices' );
 
-		const clearSettingsCache = () => {
-			console.log( 'invalidate occurs' );
+		const clearTaskCache = () => {
 			invalidateResolutionForStoreSelector( 'getTasksStatus' );
 		};
 
@@ -359,7 +358,7 @@ export default compose(
 			updateNote,
 			updateOptions,
 			updateProfileItems,
-			clearSettingsCache,
+			clearTaskCache,
 		};
 	} ),
 	window.wcSettings.plugins
--- a/client/profile-wizard/index.js
+++ b/client/profile-wizard/index.js
@@ -17,7 +17,6 @@ import {
 	OPTIONS_STORE_NAME,
 	PLUGINS_STORE_NAME,
 	withPluginsHydration,
-	SETTINGS_STORE_NAME,
 } from '@woocommerce/data';
 import { recordEvent } from '@woocommerce/tracks';
 
--- a/client/profile-wizard/index.js
+++ b/client/profile-wizard/index.js
@@ -17,6 +17,7 @@ import {
 	OPTIONS_STORE_NAME,
 	PLUGINS_STORE_NAME,
 	withPluginsHydration,
+	SETTINGS_STORE_NAME,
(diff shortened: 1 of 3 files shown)


client/profile-wizard/index.js
```diff
@@ -232,23 +234,19 @@ class ProfileWizard extends Component {
...
-					);
-				} else {
-					getHistory().push( getNewPath( {}, '/', {} ) );
-				}
-			} );
+		updateProfileItems( { completed: true } ).then( () => {
+			clearTaskCache();
+
+			if ( shouldConnectJetpack ) {
+				document.body.classList.add( 'woocommerce-admin-is-loading' );
```
samueljseay, line 241: The extra promise in the chain was redundant here and this is functionally the same as before.

client/profile-wizard/index.js
```diff
@@ -337,9 +335,16 @@ export default compose(
...
 		const { updateOptions } = dispatch( OPTIONS_STORE_NAME );
-		const { updateProfileItems } = dispatch( ONBOARDING_STORE_NAME );
+		const {
+			updateProfileItems,
+			invalidateResolutionForStoreSelector,
+		} = dispatch( ONBOARDING_STORE_NAME );
 		const { createNotice } = dispatch( 'core/notices' );
 
+		const clearTaskCache = () => {
+			invalidateResolutionForStoreSelector( 'getTasksStatus' );
```
samueljseay, line 345: invalidate the task status cache to ensure that we get an updated task status when redirecting.

src/Features/OnboardingAutomateTaxes.php
```diff
@@ -0,0 +1,62 @@
...
+				'on_onboarding_profile_completed',
+			)
+		);
+	}
+
+	/**
+	 * Set up automated taxes.
+	 */
+	public static function on_onboarding_profile_completed() {
+		$jetpack_connected = null;
```
samueljseay, line 33: These checks are based on similar checks we do [here](https://github.com/woocommerce/woocommerce-admin/blob/main/src/Features/ShippingLabelBanner.php#L56). I'd be happy to consider extracting this if I could get some advice on where that logic could belong.

becdetat: It's similar but different. I don't think it's worth extracting just yet. I think if you were to extract it `src/Features/Onboarding.php` would be the place for it.

client/profile-wizard/index.js
```diff
@@ -17,6 +17,7 @@ import {
 	OPTIONS_STORE_NAME,
 	PLUGINS_STORE_NAME,
 	withPluginsHydration,
+	SETTINGS_STORE_NAME,
```
becdetat, line 20: This isn't used, which is causing a CI error

becdetat (approved):
Looks good apart from an unused declaration that is causing CI to fail. I tested it on a JN site and it works well. Nice! Pre-approving.

PR #5094, opened 2020-09-03
Combine translation chunks when language packs are updated.

### LINKED ISSUE #5062: Incorrect use of WP_Filesystem API, resulting in fatal errors
**Describe the bug**

Here - https://github.com/woocommerce/woocommerce-admin/blob/81685688fdfab3e501f44ff0afd6e11beff707a0/src/Loader.php#L576 - `WP_Filesystem()` is called without checking the result, and then the code proceeds to call other WP_Filesystem API methods as if it had succeeded. This is incorrect. If the user has set up WP_Filesystem to work via a remote access method (e.g. FTP, SFTP) and setup fails, then this will result in fatal errors. (See https://codex.wordpress.org/Filesystem_API for an example of proper use).

For example, in the code linked above, immediately afterwards, `WP_Filesystem::is_file()` is called. Since that API call can only return a boolean (true / false, either of which would be a wrong response, since a failure to set up the filesystem link means that the path is neither known to be a file nor not be a file), then the only way for errors to be signalled is to throw an error or exception. Since WooCommerce Admin also doesn't catch those, it crashes.

Examples of users experiencing this problem:
https://wordpress.org/support/topic/ssh-sftp-updater-support-woocommerce/#post-13338117

jeffstieler:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/5062

This PR seeks to avoid using the `WP_Filesystem` instance before it's been initialized, and also moves the translation chunk combination logic to run only when language packs are updated instead of on every script load.

To avoid breaking the load of the combined file when the plugin is updated (but before the language packs are) the plugin version has been removed from the filename.

- Verify you've downloaded a language other than English (Go to Settings > General > Language and change it)
- Add to your `wp-config.php`: `define('FS_METHOD', 'ssh2');`
- Delete all `woocommerce-admin` files in `wp-content/languages/plugins`
- Go to Updates page in admin ( 🔄   icon in toolbar )
- Click "update translations" at the bottom
- Put in FTP (yes FTP) credentials - for me on VVV it was: `vagrant`/`vagrant` @ `local.wordpress.test`
- Verify that `woocommerce-admin` files exist in `wp-content/languages/plugins`
- Verify that `woocommerce-admin-[locale]-wc-admin-app.json` file exists in `wp-content/languages/plugins` for all locales with language packs (like `fr_CA` or `de_DE`)
- Load a WooCommerce Admin screen (like the Orders Report) and verify that the JS-powered strings are translated
- Remove `define('FS_METHOD', 'ssh2');` from your `wp-config.php`
- Repeat all steps

### Changelog Note:

Fix: WP_Filesystem usage when updating translations files.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -91,6 +91,9 @@ class Loader {
 		* Gutenberg has also disabled emojis. More on that here -> https://github.com/WordPress/gutenberg/pull/6151
 		*/
 		remove_action( 'admin_print_scripts', 'print_emoji_detection_script' );
+
+		// Combine JSON translation files (from chunks) when language packs are updated.
+		add_action( 'upgrader_process_complete', array( __CLASS__, 'combine_translation_chunk_files' ), 10, 2 );
 	}
 
 	/**
@@ -473,8 +476,7 @@ class Loader {
 	 * @return string Filename.
 	 */
 	public static function get_combined_translation_filename( $domain, $locale ) {
-		$version  = self::get_file_version( 'js' );
-		$filename = implode( '-', array( $domain, $locale, WC_ADMIN_APP, $version ) ) . '.json';
+		$filename = implode( '-', array( $domain, $locale, WC_ADMIN_APP ) ) . '.json';
 
 		return $filename;
 	}
@@ -490,11 +492,17 @@ class Loader {
 	 * @return array Combined translation chunk data.
 	 */
 	public static function get_translation_chunk_data( $lang_dir, $domain, $locale ) {
+		// So long as this function is called during the 'upgrader_process_complete' action,
+		// the filesystem object should be hooked up.
 		global $wp_filesystem;
 
 		// Grab all JSON files in the current language pack.
 		$json_i18n_filenames       = glob( $lang_dir . $domain . '-' . $locale . '-*.json' );
 		$combined_translation_data = array();
+
+		if ( false === $json_i18n_filenames ) {
+			return $combined_translation_data;
+		}
 
 		foreach ( $json_i18n_filenames as $json_filename ) {
 			if ( ! $wp_filesystem->is_readable( $json_filename ) ) {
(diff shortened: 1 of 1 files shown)


becdetat:
I'm getting stuck on deleting the language files. There is no `plugins` directory in `wp-content/languages`, I'm not sure what I'm doing wrong :-( :

image

jeffstieler:
@becdetat you may not have any if you haven't downloaded language packs ("update translations") before - you can proceed if that was you first run through at step 3!

octaedro (approved):
You did a great job here @jeffstieler. This is testing well on my end and the code looks good too.
Good job!

kimmenbert:
Hi,

I hope that this commit fixes the error: `Warning: ftp_fget() expects parameter 1 to be resource, null given` which is caused by `$file_contents = $wp_filesystem->get_contents( $json_filename );` in the modifed file in this PR. Happened 385k times on a site the last week.

It seems like using $wp_filesystem->get_contents() skips some filesystem checks when access directly like that.

jeffstieler:
Hi @kimmenbert - that's a strange error considering there is a `$wp_filesystem->is_readable()` check before that line: https://github.com/woocommerce/woocommerce-admin/pull/5094/files#diff-3864ffe37f3a1de39876ea62cae525f3R508-R510

If you continue seeing those warnings, please open a new issue: https://github.com/woocommerce/woocommerce-admin/issues/new?template=BUG_REPORT.md

kimmenbert:
For me it looks like the function `$wp_filesystem->is_readable()` straight up just returns true.
Have not looked too much into exactly what class is in use here. But example: https://github.com/WordPress/WordPress/blob/a5edb2a8a15697440dff359fac53c27dd2067db3/wp-admin/includes/class-wp-filesystem-ftpext.php#L469

PR #5096, opened 2020-09-03
Onboarding: Add TOS note to bundle install.

timmyc:
Fixes #5081

This branch adds in a line of text around terms of service when installing the "bundle" of extensions.

### Screenshots

__Before__
Setup Wizard ‹ WooCommerce ‹ bend outdoors — WooCommerce 2020-09-03 13-51-31

__After__
Setup Wizard ‹ WooCommerce ‹ bend outdoors — WooCommerce 2020-09-03 13-48-01

### Detailed test instructions:

- Check out this branch and launch the profile wizard: `wp-admin/admin.php?page=wc-admin&reset_profiler=1`
- Be sure to use a US-based address on the first step and select the fashion industry on step 2.
- Proceed to step 4 and verify your screen looks like the _after_ above.

### Changelog Note:

Tweak: Add terms of service link to bundled extension install.

cc @becdetat

PR #5097, opened 2020-09-04
Add ToS copy to OBW business details

### LINKED ISSUE #5081: Business details step: add copy to 1-click install flow
We need to add the following copy in the 1-click install flow, at the end of the page:

By installing Jetpack and WooCommerce Services plugins for free you agree to our [Terms of Service](https://wordpress.com/tos/).

Here's the mockup: https://d.pr/i/7uvSg2

becdetat:
Fixes #5081 (in 1.5.0)

This adds some text to the business details step if the bundle is able to be installed, including a link to the ToS.

@timmyc [originally did this](https://github.com/woocommerce/woocommerce-admin/pull/5096) however there were some conflicts when cherry-picking it into the 1.5.0-rc.5 branch, so I've recreated it.

### Screenshots
Before:

After:
image

- Check out this branch and launch the profile wizard: `wp-admin/admin.php?page=wc-admin&reset_profiler=1`
- Be sure to use a US-based address on the first step and select the fashion industry on step 2.
- Proceed to step 4 and verify your screen looks like the after above.

### Changelog Note:

Tweak: Add terms of service link to bundled extension install.

### DIFF
--- a/client/profile-wizard/steps/business-details.js
+++ b/client/profile-wizard/steps/business-details.js
@@ -357,14 +357,16 @@ class BusinessDetails extends Component {
 				</p>
 			);
 		}
+
 		const accountRequiredText = this.bundleInstall
 			? __(
 					'User accounts are required to use these features.',
 					'woocommerce-admin'
 			  )
 			: '';
+
 		return (
-			<Fragment>
+			<div className="woocommerce-profile-wizard__footnote">
 				<p>
 					{ sprintf(
 						_n(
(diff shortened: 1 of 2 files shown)


client/profile-wizard/steps/business-details.js
```diff
@@ -377,7 +379,28 @@ class BusinessDetails extends Component {
...
+					<p>
+						{ interpolateComponents( {
+							mixedString: __(
+								'By installing Jetpack and WooCommerce Services plugins for free you agree to our {{link}}Terms of Service{{/link}}.',
+								'woocommerce-admin'
+							),
+							components: {
+								link: (
+									// eslint-disable-next-line jsx-a11y/anchor-has-content
+									<a
```
jeffstieler, line 392: Why does this need to be an anchor instead of a `<Link/>` component like in https://github.com/woocommerce/woocommerce-admin/pull/5096?

jeffstieler (approved):
I'm approving this, as it works fine and will only land for `1.5.x`, but it's curious to me that a `<Link />` component wasn't used. (I was able to swap in a `<Link />` locally without issue FWIW)

client/profile-wizard/steps/business-details.js
becdetat: So I was getting this with `Link`:

```
Uncaught Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined
```

Turns out I was trying to use `Link` from `@wordpress/components` whereas I should have been using the one from `@woocommerce/components`. This is fixed now, thanks for the catch.

PR #4715, opened 2020-06-28
Variation Names: Allow long names to be shown

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/4336
Depends on https://github.com/woocommerce/woocommerce-admin/pull/4704

Variation names not were shown when attributes are more than one word. This PR uses the `woocommerce_product_variation_title_include_attributes` filter to make sure long names are shown.

### Before

Screen Shot 2020-05-11 at 2 41 16 PM

### After

Screen Shot 2020-06-29 at 11 17 22 AM

### Detailed test instructions:

**To Reproduce**

1. Create a variable product or use an existing one.
2. Change one attribute t be more than one word.
3. Create an order.
3. Be sure to clear cache if you've recently visited this report `wp transient delete woocommerce_reports-transient-version`
4. Visit the Product Report > Single Product and search for the variable product.
4. See the Product Report's variation names shown in full.

### Changelog Note:

tweak - Force full variation names to be shown on reports.

PR #5100, opened 2020-09-04
Avoid changing product titles by removing variation title filter usage.

jeffstieler:
I'm considering the use of `woocommerce_product_variation_title_include_attributes` harmful. It was introduced in https://github.com/woocommerce/woocommerce-admin/pull/4715.

When determining the product title, the variations data store will **update the database** if the value differs!

https://github.com/woocommerce/woocommerce/blob/9c6c0d73d817328e3a90b6e3e0b1b0f0b7662270/includes/data-stores/class-wc-product-variation-data-store-cpt.php#L86-94

We're using this filter so attribute values are included in Product Variation titles, but **we should not be modifying products** just to make querying them easier. (I'm sure this was 100% an unintended side effect)

This PR
* Removes the use of the `woocommerce_product_variation_title_include_attributes ` filter
* Generates a title using the attribute values on the client side (JS)
* Queries the attribute values when searching for variations (SQL)

- Visit the Product Report > Single Product and search for a variable product.
- See the Product Report's variation names shown in full.
- Select "single variation" from the dropdown
- Search for variations using attribute values ( like, "orange" color )
- Verify results are shown as expected

### Changelog Note:

Fix: avoid filter usage that potentially modifies Product Variation titles.

### DIFF
--- a/client/analytics/report/products/table-variations.js
+++ b/client/analytics/report/products/table-variations.js
@@ -19,6 +19,13 @@ import { CurrencyContext } from '../../../lib/currency-context';
 const manageStock = getSetting( 'manageStock', 'no' );
 const stockStatuses = getSetting( 'stockStatuses', {} );
 
+const getFullVariationName = ( rowData ) =>
+	get( rowData, [ 'extended_info', 'name' ], '' ) +
+	' - ' +
+	get( rowData, [ 'extended_info', 'attributes' ], [] )
+		.map( ( { option } ) => option )
+		.join( ', ' );
+
 class VariationsReportTable extends Component {
 	constructor() {
 		super();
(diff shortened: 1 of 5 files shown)


octaedro (commented):
Nice job @jeffstieler! Code looks very good but I'm having some trouble with the `unit testing`.
Maybe it's not related but I'm getting the errors:

````
1) WC_Tests_API_Reports_Orders_Stats::test_product_attributes_filter
Undefined index: items

2) WC_Tests_API_Reports_Orders::test_product_attributes_filter
Undefined index: items
````

jeffstieler:
@octaedro do those tests fail for you locally on `main` as well? The CI builds pass and those tests pass for me locally.

octaedro (approved):
You're right @jeffstieler, it was a local issue.
This is testing well here and the code LGTM 🚀

PR #5107, opened 2020-09-08
Add Google Ads & Marketing by Kliken Note Provider

jconroy:
See https://github.com/Automattic/woocommerce.com/issues/8471

Replace the  Facebook Marketing Experts offer with a note for Google Ads & Marketing by Kliken.

Note should be displayed if Google Ads & Marketing by Kliken is installed and the store has at least 20 orders in the past month.

I'm still a little fuzzy on when we should be deleting old notes - I've opted to remove the facebook note on update in this PR similar to what was done previously here https://github.com/woocommerce/woocommerce-admin/pull/4030 - happy for thoughts/feedback/clarification on when to or not to do that. 

In the future this sort of note would be probably be a good candidate for the remote inbox notifications.

### Screenshots

Screen Shot 2020-09-09 at 10 21 32 am

- Create 20 tests orders on your test site in the past month (if you don't already have that many) or adjust `possibly_add_note()` to check for a lower number.
- Install Google Ads & Marketing by Kliken
- Run `wc_admin_daily` cron event (I use WP Crontrol) just to ensure the note event is run.
- Check you have a new inbox note about Google Ads & Marketing by Kliken.
- Check the action links to the WCCOM product page

### Changelog Note:

Enhancement: Replace Facebook Marketing Expert Offer Inbox Note with Google Ads & Marketing Inbox Note


includes/wc-admin-update-functions.php
```diff
@@ -124,3 +124,10 @@ function wc_admin_update_140_change_deactivate_plugin_note_type() {
 function wc_admin_update_140_db_version() {
 	Installer::update_db_version( '1.4.0' );
 }
+
+/**
+ * Remove Facebook Experts note.
+ */
+function wc_admin_update_160_remove_facebook_note() {
+	WC_Admin_Notes::delete_notes_with_name( 'wc-admin-facebook-marketing-expert' );
+}
```
jeffstieler, line 133: This is largely a hidden requirement, so no worries for missing it. When adding DB update callbacks, one needs to bump the version number.

Please create a `wc_admin_update_160_db_version()` that bumps to `1.6.0`. 😄 

Additionally, these functions need to be added to the `$db_updates` array in `src/Install.php`: https://github.com/woocommerce/woocommerce-admin/blob/aecd050aff498d997adea9ec3e00a6bce8bd8a61/src/Install.php#L28

(otherwise they aren't run during updates)

jeffstieler (changes requested):
The code for the note itself looks good, but the update callback needs some adjustment (see comments).

I found this sentence in the note copy to be rather confusing:

> You can also drive free traffic to your store with free listings for only $10 per month!

Free listings are $10/mo? 🙃

jconroy:
Thanks @jeffstieler , I completely missed that db update - good catch 🙏 

> I found this sentence in the note copy to be rather confusing: ....Free listings are $10/mo?

@happystaffyme just double checking you are ok with the provided wording above?

jeffstieler (approved):
Code looks good and tests well. Approving from my standpoint - pending verbiage feedback from @happystaffyme.

happystaffyme:
Approved @jeffstieler !

jconroy:
@timmyc Any chance of getting this one into the 4.6 ?

timmyc:
Indeed, we shall land this for 1.6 @jconroy

PR #5108, opened 2020-09-08
Enable the home screen for everybody

### LINKED ISSUE #4761: Home screen as the default experience for every user
**Description:** the time between 4.3 and 4.4 will be enough for existing users to experiment with the new home screen. This new screen should become the default home for everyone in the Core release that will be shipped in August.

**Acceptance Criteria:**
- [ ] Change the logic so the home screen becomes the default experience for everyone
~- [ ] Display an inbox notification about the new home screen for existing users, if #4590 isn't ready~

becdetat:
Fixes #4761

This removes the ability to opt out of the home screen, making the home screen the default (and only) experience. Note that when this is merged it will unblock https://github.com/woocommerce/woocommerce/pull/27047, so remember to update that ticket when this is merged.

A store that is opted out of the home screen should be ignoring that option and showing the home screen. Smoke test the home screen, other WooCommerce pages and the analytics pages.

### Changelog Note:

Feature: Enable the home screen for everybody

### DIFF
--- a/packages/components/src/summary/number.js
+++ b/packages/components/src/summary/number.js
@@ -32,7 +32,6 @@ import Link from '../link';
  * @param {boolean} props.selected
  * @param {number|string} props.value
  * @param {Function} props.onLinkClickCallback
- * @param {boolean} props.isHomescreen
  * @return {Object} -
  */
 const SummaryNumber = ( {
@@ -47,11 +46,10 @@ const SummaryNumber = ( {
 	reverseTrend,
 	selected,
 	value,
-	onLinkClickCallback,
-	isHomescreen,
+	onLinkClickCallback
 } ) => {
 	const liClasses = classnames( 'woocommerce-summary__item-container', {
-		'is-homescreen': isHomescreen,
+		'is-homescreen': true,
 		'is-dropdown-button': onToggle,
 		'is-dropdown-expanded': isOpen,
 	} );
@@ -208,10 +206,6 @@ SummaryNumber.propTypes = {
 	 * A function to be called after a SummaryNumber, rendered as a link, is clicked.
 	 */
 	onLinkClickCallback: PropTypes.func,
-	/**
-	 * A boolean used to show if the Summary Number is on the Homescreen.
-	 */
-	isHomescreen: PropTypes.bool,
 };
 
 SummaryNumber.defaultProps = {
--- a/packages/components/src/summary/number.js
+++ b/packages/components/src/summary/number.js
@@ -46,7 +46,7 @@ const SummaryNumber = ( {
 	reverseTrend,
 	selected,
 	value,
-	onLinkClickCallback
+	onLinkClickCallback,
 } ) => {
 	const liClasses = classnames( 'woocommerce-summary__item-container', {
 		'is-homescreen': true,
--- a/packages/components/src/summary/number.js
+++ b/packages/components/src/summary/number.js
@@ -49,7 +49,6 @@ const SummaryNumber = ( {
 	onLinkClickCallback,
 } ) => {
 	const liClasses = classnames( 'woocommerce-summary__item-container', {
-		'is-homescreen': true,
 		'is-dropdown-button': onToggle,
 		'is-dropdown-expanded': isOpen,
 	} );
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -243,22 +243,11 @@ class Loader {
 	}
 
 	/**
-	 * Registers a basic page handler for the app entry point.
+	 * Connects existing WooCommerce pages.
 	 *
 	 * @todo The entry point for the embed needs moved to this class as well.
 	 */
 	public static function register_page_handler() {
-		wc_admin_register_page(
-			array(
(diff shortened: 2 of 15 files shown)


jeffstieler (commented):
Combined with https://github.com/woocommerce/woocommerce/pull/27047, this will turn the homescreen on for all stores, and remove the UI that previously allowed users to turn off the homescreen. Is that the intention, or were we wanting to preserve prior "opt outs" of this feature?

In terms of ensuring the option is toggle as a "one off" or "one-shot", what do you think about using the DB version update callback mechanism instead of introducing another option to use as a flag?

becdetat:
> Combined with woocommerce/woocommerce#27047, this will turn the homescreen on for all stores, and remove the UI that previously allowed users to turn off the homescreen. Is that the intention, or were we wanting to preserve prior "opt outs" of this feature?

I didn't realise that the UI to opt out was also being removed. It seems from woocommerce/woocommerce#27047 that the intention is to force all users on to the home screen, as otherwise they will be stuck on the dashboard with no way to opt back in to the home screen. Given that, this doesn't need to be a one-shot and the original `woocommerce_homescreen_enabled` option could be either removed or replaced with a filter to allow plugin authors to opt out of the home screen. I'll ask what the intention is on the woocommerce issue.

> In terms of ensuring the option is toggle as a "one off" or "one-shot", what do you think about using the DB version update callback mechanism instead of introducing another option to use as a flag?

I didn't think  of that 😳  it makes sense to do it that way instead.

becdetat:
The [original issue on woocommerce](https://github.com/woocommerce/woocommerce/issues/26972) reads that we won't be maintaining the two different screens and that the opt-out option should be removed (which is done in https://github.com/woocommerce/woocommerce/pull/27047, which is currently blocked waiting for _this_ PR).

So I'm going to go ahead with removing the option completely and making the home screen the only experience.

@pmcpinto does this sound correct to you?

timmyc:
> So I'm going to go ahead with removing the option completely and making the home screen the only experience.

Correct my understanding is that we will only offer the Home Screen as the default experience for the WooCommerce menu item.

becdetat:
I've removed the special handling for the home screen and made it the only experience. Once this is merged it will unblock https://github.com/woocommerce/woocommerce/pull/27047 (so note to self, add a note to that issue when merging).

@jeffstieler this is ready for another review please :-)

packages/components/src/summary/number.js
```diff
@@ -48,10 +47,9 @@ const SummaryNumber = ( {
 	selected,
 	value,
 	onLinkClickCallback,
-	isHomescreen,
 } ) => {
 	const liClasses = classnames( 'woocommerce-summary__item-container', {
-		'is-homescreen': isHomescreen,
+		'is-homescreen': true,
```
psealock, line 52: This is an artefact from when Summary Numbers differed between on Reports and the Homescreen. I just double checked and this CSS class name doesn't appear to have an impact if you remove `:not(.is-homescreen)` from this line.

https://github.com/woocommerce/woocommerce-admin/blob/6198ccf165e46530caf8040aa2162a5d2e6aa904/packages/components/src/summary/style.scss#L202

So, I think this logic can be removed completely, including the `:not(.is-homescreen)`.

src/Loader.php
```diff
@@ -252,11 +248,9 @@ public static function load_features() {
 	 * @todo The entry point for the embed needs moved to this class as well.
 	 */
 	public static function register_page_handler() {
-		$id = self::is_feature_enabled( 'homescreen' ) ? 'woocommerce-home' : 'woocommerce-dashboard';
-
 		wc_admin_register_page(
 			array(
-				'id'         => $id, // Expected to be overridden if dashboard is enabled.
+				'id'         => 'homescreen',
```
psealock, line 253: Should we keep this as `woocommerce-home` or does it need to be different than the page registration in `src/Features/AnalyticsDashboard.php`? Or maybe we can do away with one?

psealock (approved):
This is all testing fantastic. Just a minor CSS issue and a question about registering the home page potentially twice: Id `homescreen` and id `woocommerce-home`. Otherwise code is looking good.

src/Loader.php
psealock: Oddly enough, in the navigation project this entry is causing issues because the `title` property below is `null`. I'm wondering if its overwriting the other one.

pmcpinto:
> @pmcpinto does this sound correct to you?

@becdetat sorry for the late reply, but I was afk last Thu/Fri. Yes, let's simplify and keep the new home screen as the only home page. Thanks!

packages/components/src/summary/number.js
becdetat: Nice, done in 4ac8ded

src/Loader.php
becdetat: Everything seems to work fine if I remove this registration, done in 3a7e243

psealock (approved):
>Everything seems to work fine if I remove this registration, done in 3a7e243

Great, thats what I was hoping 🚢

PR #5124, opened 2020-09-10
Onboarding - Store Profiler - Removed selected CBD option for eCommerce plan

### LINKED ISSUE #573: Onboarding - Store Profiler: Hide the option "CBD and other hemp-derived products"
WordPress.com doesn't support CBD, so the option `CBD and other hemp-derived products` should be removed from the Industry step in the eCommerce plan.

octaedro:
Fixes [573](https://github.com/automattic/wc-calypso-bridge/issues/573)

WordPress.com doesn't support CBD. This is solved [here](https://github.com/Automattic/wc-calypso-bridge). For users with the `eCommerce` plan who previously selected the `CBD and other hemp-derived products` option before that modification applies, this PR removes the industry `CBD and other hemp-derived products` from the selected industries list.

### Screenshots
screenshot-three wordpress test-2020 09 10-10_49_02

To test this PR it's necessary to use a site with [wc-calypso-bridge](https://github.com/Automattic/wc-calypso-bridge) installed.
- In the `wc-calypso-bridge` project, checkout to `master`. 
- Go to the industry step in the store profiler (URL: `/wp-admin/admin.php?page=wc-admin&path=%2Fsetup-wizard&step=industry&calypsoify=1`)
- Be sure the option `CBD and other hemp-derived products` is selected.
- Go to the `home` screen.
- In `wc-calypso-bridge` checkout to the branch `remove/option-CBD-from-onboarding`
- Be sure you have added `&calypsoify=1` at the end of the URL.
- Go to the payment methods (URL `/wp-admin/admin.php?page=wc-admin&task=payments`)
- Verify the following option is not visible
screenshot-three wordpress test-2020 09 16-09_07_18

### Changelog Note:
Fix: Removed selected CBD option from onboarding for eCommerce plan

### DIFF
--- a/client/task-list/tasks/payments/index.js
+++ b/client/task-list/tasks/payments/index.js
@@ -309,9 +309,11 @@ export default compose(
 		const { createNotice } = dispatch( 'core/notices' );
 		const { installAndActivatePlugins } = dispatch( PLUGINS_STORE_NAME );
 		const { updateOptions } = dispatch( OPTIONS_STORE_NAME );
-		const { invalidateResolutionForStoreSelector } = dispatch(
-			ONBOARDING_STORE_NAME
-		);
+		const {
+			invalidateResolution,
+			invalidateResolutionForStoreSelector,
+		} = dispatch( ONBOARDING_STORE_NAME );
+		invalidateResolution( 'getProfileItems', [] );
 		return {
 			clearTaskStatusCache: () =>
 				invalidateResolutionForStoreSelector( 'getTasksStatus' ),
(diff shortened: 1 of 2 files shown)


timmyc:
@octaedro in-lieu of introducing calypso-centric logic in wc-admin, what do you think about just leveraging the `woocommerce_admin_onboarding_industries` filter in `wc-calypso-bridge` to remove the CBD industry instead?

octaedro:
> @octaedro in-lieu of introducing calypso-centric logic in wc-admin, what do you think about just leveraging the `woocommerce_admin_onboarding_industries` filter in `wc-calypso-bridge` to remove the CBD industry instead?

Good idea @timmyc! I created [this PR](https://github.com/Automattic/wc-calypso-bridge/pull/601) to handle that change.
I'm going to use this wc-admin PR to add a small change in the payment methods since we could have a problem if the user has selected `CBD` before the update. In that case, as we are using a cached response in the payment methods page (`/wp-admin/admin.php?page=wc-admin&task=payments`), the user will continue seeing the `CBD` option after the update. I will force a refresh of that data every time that page is loaded to prevent the error.

becdetat (approved):
LGTM, tests well 👍

timmyc:
Cherry picked to `release/1.6.0-rc.1` in edd9a36

PR #5130, opened 2020-09-11
Don't include "min" suffix in build JavaScript when targeting core.

### LINKED ISSUE #26994: Different languages on the dashboard
Woocommerce 4.3 is set to German, but on my dashboard texts are still displayed in english:

woocommerce_4_3_dashboard

The translation files are correct, apparently there is an error in the output.

Many thanks in advance!

Best regards!

jeffstieler:
Files with "min" are ignored by GlotPress / translate.wp.org.

Fixes https://github.com/woocommerce/woocommerce/issues/26994.

This PR seeks to remove "min" from the minified JS files packaged up for the core build. This is to ensure that the files are parsed by GlotPress and language pack files are built.

Testing this is definitely tricky. There really isn't a way to test this in situ where it will actually solve the problem.

- Run a normal build, e.g. `npm start` or `npm run build`
- Verify that `*.js` and `*.min.js` files are built in `/dist`
- Verify that loading script files works with and without `SCRIPT_DEBUG` defined as `true`
- Delete the `/dist` dir
- Run a core build: `WC_ADMIN_PHASE=core npm run build`
- Verify there are no `*.min.js` files in `/dist`
- Verify the JS contents are minified
- Verify that loading script files works with and without `SCRIPT_DEBUG` defined as `true`

### Changelog Note:

Fix: ensure GlotPress finds JavaScript files when building language packs for core WooCommerce.

### DIFF
--- a/config/core.json
+++ b/config/core.json
@@ -11,7 +11,7 @@
 		"remote-inbox-notifications": false,
 		"shipping-label-banner": true,
 		"store-alerts": true,
-		"unminified-js": false,
+		"minified-js": false,
 		"wcpay": true,
 		"homescreen": true,
 		"mobile-app-banner": true
--- a/config/development.json
+++ b/config/development.json
@@ -11,7 +11,7 @@
 		"remote-inbox-notifications": true,
 		"shipping-label-banner": true,
 		"store-alerts": true,
-		"unminified-js": true,
+		"minified-js": true,
 		"wcpay": true,
 		"homescreen": true,
 		"mobile-app-banner": true
--- a/config/plugin.json
+++ b/config/plugin.json
@@ -11,7 +11,7 @@
 		"remote-inbox-notifications": false,
 		"shipping-label-banner": true,
 		"store-alerts": true,
-		"unminified-js": true,
+		"minified-js": true,
 		"wcpay": true,
 		"homescreen": true,
 		"mobile-app-banner": true
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -180,9 +180,9 @@ class Loader {
 	 * @return boolean If js asset should use minified version.
 	 */
 	public static function should_use_minified_js_file( $script_debug ) {
-		// un-minified files are only shipped in non-core versions of wc-admin, return true if unminified files are not available.
-		if ( ! self::is_feature_enabled( 'unminified-js' ) ) {
-			return true;
+		// minified files are only shipped in non-core versions of wc-admin, return false if minified files are not available.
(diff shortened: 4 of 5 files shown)


samueljseay:
> Verify that loading script files works with and without SCRIPT_DEBUG defined as true

@jeffstieler just want to walk through some of my assumptions here to make sure I understand correctly.

When we package the wc-admin package to be bundled with wc core:

1. we remove the min suffix but the files are still minified for the build?
2. since there are no more `min` suffixed files, the UnminifyWebpackPlugin now has no effect?
3. `SCRIPT_DEBUG` will basically have no effect now, we will always serve the file without a `min` suffix, but its a bit misleading to a user in this environment because the file *will* in fact be minified?

Once wc-core is built will they then minify our files?

config/core.json
```diff
@@ -11,7 +11,7 @@
 		"remote-inbox-notifications": false,
 		"shipping-label-banner": true,
 		"store-alerts": true,
-		"unminified-js": false,
+		"minified-js": false,
```
psealock, line 14: Is is fair to say we're doing this as a temporary measure? If not, I'd find this confusing and expect that our core production release would have `minified-js: true`. But, maybe I'm thinking about this wrong.

psealock (commented):
This tests well according to instructions. No issues found. I'll let Sam have final approval.

jeffstieler:
> When we package the wc-admin package to be bundled with wc core:
> 
> 1. we remove the min suffix but the files are still minified for the build?

Correct. Core wants to keep the build small, so we are still minifying the file contents. GlotPress (translate.wordpress.org) will ignore files with the `.min` suffix, so we omit that in the core build.

> 2. `SCRIPT_DEBUG` will basically have no effect now, we will always serve the file without a `min` suffix, but its a bit misleading to a user in this environment because the file will in fact be minified?

Yes, when loading WooCommerce Admin from the core package, `SCRIPT_DEBUG` has no effect. This is true before introducing this change - just that `.min` was in the filename.

config/core.json
jeffstieler: This feature flag only supports the lack of both minified and unminified JS files in the core build. I chose to rename it because by outward appearance, there are no `*.min.js` files (even though the contents are actually minified).

If we were able to include unminified JS or soucemaps, etc, without inflating the core build size - we might be able to remove this temporary measure.

octaedro (approved):
Nice solution @jeffstieler. This is testing well here and code LGTM :shipit:

PR #5141, opened 2020-09-16
Search all variation attribute values, not just global ones.

jeffstieler:
Fixes an issue with the implementation in #5100.

The prior PR only search "global" product attribute values - those created through the Products > Attributes term management screens.

This PR seeks to reduce the specificity of the SQL `LIKE` clause to match all `attribute_%` meta_keys, not just `attribute_pa_%`.

The included test case illustrates the problem - variations that have "one off" or non-global attributes vs. those that use attributes and values that have been defined in the Product Attribute taxonomy.

To be honest, I'm not sure how to create non-global product attributes through the WooCommerce dashboard. My test store populated with WC Smooth Generator does have products with both `attribute_` and `attribute_pa_` prefixed postmeta keys.

If you query your post meta table and find both types of keys, you can test with the UI:

- Visit the Product Report > Single Product and search for a variable product.
- See the Product Report's variation names shown in full.
- Select "single variation" from the dropdown
- Search for variations using a global attribute value ( like, "orange" color )
- Verify results are shown as expected
- Search for variations using a non-global attribute value ( like, "banana" flavor )
- Verify results are shown as expected

### Changelog Note:

N/A - unreleased bug.

### DIFF
--- a/src/API/ProductVariations.php
+++ b/src/API/ProductVariations.php
@@ -79,7 +79,7 @@ class ProductVariations extends \WC_REST_Product_Variations_Controller {
 		if ( $search ) {
 			$join .= " LEFT JOIN {$wpdb->postmeta} AS attr_search_meta
 						ON {$wpdb->posts}.ID = attr_search_meta.post_id
-						AND attr_search_meta.meta_key LIKE 'attribute_pa_%' ";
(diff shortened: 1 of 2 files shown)


octaedro (approved):
You added a nice improvement @jeffstieler. This is testing well and the code looks good.
Good job 🚀

PR #5144, opened 2020-09-16
Optimize the wc-admin bundle size by unbundling @wordpress/components

### LINKED ISSUE #5034: Unbundling @wordpress/components
Some time ago a decision was made to bundle `@wordpress/components` when React hooks were introduced. Previous versions of Gutenberg bundled with WP versions we supported did not include hooks and we were forced to package all WP components to avoid fatal errors.

Since then, we have become accustomed to using the latest features of `@wordpress/components`, including new components marked with the `__experimental` prefix.

Now that we support WP 5.3, can we unbundle `@wordpress/components`? The React hooks issue is likely resolved, but can we conditionally import `{ Text } from '@wordpress/components'` or `{ __experimentalText } from '@wordpress/components'` depending on availability?

To go one step further, it would be awesome to have some kind of chart/table/diagram to look up what versions of dependencies we support. For example, to quickly know that WP 5.3 includes Gutenberg 7.3, `@wordpress/components` 3.0.0, etc. Bonus if it could be automated. This would allow us to quickly know if using a particular component or feature of React will cause backwards incompatibility issues.

samueljseay:
Fixes #5034

Unbundle `@wordpress/components` to save significant size across all the JS bundles.

To do this, I've forked `__experimentalText` for now as a temporary solution. It is the only piece of `@wordpress/components` that we currently use that cannot be loaded from the browser global. It is exposed from our
components package because it is used in both `client` and `packages` at the moment. Any ideas of a better
place for it would be welcome, I've kept it marked as experimental in the hope that others would not use it without
realizing it is unstable.

This is however a less than ideal solution and I'm happy to hear if anyone has issues with this approach.

As a result of forking `__experimentalText` and unbundling `@wordpress/components` we save a total
of `1.31mb` over all scripts. The `app` entry point alone drops by `353.31kb`.

### Testing

To test this I recommend a smoke test of all the areas of the application, looking for any errors in console. You should also smoke test Wordpress version `5.3` and `5.4` in the same way.

If you would like to see the bundle size differences for yourself then you can do:

1. Check out `main`, run `npm clean && npm i && npm run analyze`. A browser window will open and show you the size of the all the bundles. It's crucial that you run `npm clean` because leftover artifacts from other builds may be counted.

2. Check out this branch and run `npm clean && npm i && npm run analyze`. Compare the sizes to see the differences.

### DIFF
--- a/client/header/activity-panel/activity-header/index.js
+++ b/client/header/activity-panel/activity-header/index.js
@@ -4,8 +4,10 @@
 import classnames from 'classnames';
 import { Component } from '@wordpress/element';
 import PropTypes from 'prop-types';
-import { __experimentalText as Text } from '@wordpress/components';
-import { EllipsisMenu } from '@woocommerce/components';
+import {
+	EllipsisMenu,
+	__experimentalText as Text,
+} from '@woocommerce/components';
 
 /**
  * Internal dependencies
--- a/client/header/activity-panel/panels/help.js
+++ b/client/header/activity-panel/panels/help.js
@@ -2,14 +2,17 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { __experimentalText as Text } from '@wordpress/components';
 import { withSelect } from '@wordpress/data';
 import { Fragment, useEffect } from '@wordpress/element';
 import { applyFilters } from '@wordpress/hooks';
 import { Icon, chevronRight, page } from '@wordpress/icons';
 import { partial } from 'lodash';
 import { getSetting } from '@woocommerce/wc-admin-settings';
-import { List, Section } from '@woocommerce/components';
+import {
+	List,
+	Section,
+	__experimentalText as Text,
+} from '@woocommerce/components';
 import {
 	ONBOARDING_STORE_NAME,
 	PLUGINS_STORE_NAME,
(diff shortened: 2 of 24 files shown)


jeffstieler:
Woo! That's a whopping reduction in build size.

Does the move to using the external support all WP versions that we need to? We support 5.3+ and were specifying version 10.0.0 (I think?) of the WP components.

samueljseay:
@jeffstieler good point. I had forgotten to test 5.3 explicitly. I'll do that.

samueljseay:
@jeffstieler looks like this might be dead in the water. the CHANGELOG for @wordpress/components is inaccurate sadly. Components like `Card` were introduced post 8.3.x and haven't been mentioned in the CHANGELOG at all.

psealock:
Going forward, we could wrap features like Navigation in checks to see if new components exist to future proof things.

```js
if ( window.wcAdminFeatures.navigation && __experimentalNavigation ) {
// ... load Nav 
} else {
// add note saying to upgrade
}
```

psealock:
@samueljseay WP 5.4 includes Gutenberg version 7.5 which includes `__experimentalText`.

https://github.com/WordPress/gutenberg/tree/release/7.5/packages/components/src/text

This means we can do something like this to future proof experimental features.

```js
import { __experimentalText, Text as PossiblyAvailableText } from '@wordpress/components';

const Text = PossiblyAvailableText || __experimentalText;
```

As for `__experimentalNavigation`, we'll need to alter the feature flag to determine availability.

PR #5149, opened 2020-09-16
Taxes Sync: Force float before addition of string numbers

### LINKED ISSUE #5136: Another non-numeric value type issue (Taxes/DataStore.php line 290)
[15-Sep-2020 01:31:36 UTC] PHP Warning:  A non-numeric value encountered in /usr/share/wordpress/wp-content/plugins/woocommerce/packages/woocommerce-admin/src/API/Reports/Taxes/DataStore.php on line 290

I saw a few bugs similar to this one reported and fixed, but not this particular one. 

In this case, if I am on my wordpress dashboard, this one seems to pop up in the logs. 

It is just a wee bit of noise we ignore in production and hopefully an easy fix

### WordPress Environment ###

WC Version: 4.5.2
REST API Version: ✔ 4.5.2
WC Blocks Version: ✔ 3.1.0
Action Scheduler Version: ✔ 3.1.6
WC Admin Version: ✔ 1.5.0
Log Directory Writable: ✔
WP Version: 5.5.1
WP Multisite: –
WP Memory Limit: 640 MB
WP Debug Mode: ✔
WP Cron: ✔
Language: en_US
External object cache: ✔

### Server Environment ###

Server Info: Apache/2.4.38 (Debian)
PHP Version: 7.3.19-1~deb10u1
PHP Post Max Size: 8 MB
PHP Time Limit: 30
PHP Max Input Vars: 1000
cURL Version: 7.64.0
OpenSSL/1.1.1d

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/5136

aka Explorations in PHP numbers as strings.

So apparently PHP adds numbers as strings 🤭 

```php
$number = "33" + "66"; 
error_log( $number); // 99
error_log( gettype( $number ) ); // integer
```
But maybe not with all versions because https://github.com/woocommerce/woocommerce-admin/issues/5136 🤔 

In any case, this PR forces stringed values to be floats before adding them because [WC_Order_Item_Tax](https://woocommerce.github.io/code-reference/classes/WC-Order-Item-Tax.html) returns strings for some values where you'd expect numbers.

Interestingly, PHP is smart enough to caste as double where appropriate.

```php
$number = (float) "33.33" + (float) "66.66"; 
error_log( gettype( $number ) ); // double 👍 
```

### Question

Since PHP handles adding stringed numbers (but not always) are the changes done here best practice? I find it interesting that this doesn't happen ALL the time. Is it possible a simple update to PHP will solve this problem? 

1. Create an order, making sure it has taxes applied.
2. Visit the Taxes Report and make sure it was correctly synced.

### Changelog Note:

- Dev: Caste WC_Order_Item_Tax 's `get_tax_total` and `get_shipping_tax_total` results before adding them

### DIFF
--- a/src/API/Reports/Taxes/DataStore.php
+++ b/src/API/Reports/Taxes/DataStore.php
@@ -287,7 +287,7 @@ class DataStore extends ReportsDataStore implements DataStoreInterface {
 					'tax_rate_id'  => $tax_item->get_rate_id(),
 					'shipping_tax' => $tax_item->get_shipping_tax_total(),
 					'order_tax'    => $tax_item->get_tax_total(),
-					'total_tax'    => $tax_item->get_tax_total() + $tax_item->get_shipping_tax_total(),
+					'total_tax'    => (float) $tax_item->get_tax_total() + (float) $tax_item->get_shipping_tax_total(),
 				),
 				array(
 					'%d',


octaedro (approved):
Good job @psealock. The code LGTM and it's testing well here 🚢 

> Since PHP handles adding stringed numbers (but not always) are the changes done here best practice? I find it interesting that this doesn't happen ALL the time. Is it possible a simple update to PHP will solve this problem?

This is an excellent question. An update to PHP probably would solve the problem, but IMO adding this fix to our solution will make the code more robust in the different environments, so I think that this PR is a good addition.

PR #5152, opened 2020-09-17
Use 'yes' to update modal dismiss option.

samueljseay:
Fixes a bug I noticed in Atomic sites where dismissing the home screen modal would not set the option causing the modal to come back on each reload.

Under the hood our options rest API calls `update_option`. because of the implementation of this, its actually not clear if `update_option` is somehow coeercing the set value to false, or if its not actually setting the option, because `update_option` returns false if it doesn't update it. 

I confirmed though that changing to use `yes`, did set the option correctly with the same user as before, so I think this should fix the issue.

1. Create an atomic ephemeral site
2. Package up this branch as a plugin by running `npm run test:zip`
3. Install the zip file as a plugin via the plugin manager (alongside WooCommerce 4.5.2)
4. Skip the OBW and go to the WooCommerce home screen
5. Click through all the buttons in the modal, clicking "lets go" to dismiss the modal
6. Refresh the page
7. Expected: the modal should not reappear

### Changelog Note:

Fix: persist the option to dismiss the home screen modal.

### DIFF
--- a/client/homescreen/layout.js
+++ b/client/homescreen/layout.js
@@ -123,8 +123,7 @@ export const Layout = ( {
 				<WelcomeModal
 					onClose={ () => {
 						updateOptions( {
-							woocommerce_task_list_welcome_modal_dismissed:
-								'yes',
+							woocommerce_task_list_welcome_modal_dismissed: true,
 						} );
 					} }
 				/>
@@ -167,7 +166,7 @@ export default compose(
 
 		const welcomeModalDismissed =
 			getOption( 'woocommerce_task_list_welcome_modal_dismissed' ) ===
-			'1';
+			'yes';
 
 		const welcomeModalDismissedIsResolving = isResolving( 'getOption', [
 			'woocommerce_task_list_welcome_modal_dismissed',
--- a/client/homescreen/layout.js
+++ b/client/homescreen/layout.js
@@ -123,7 +123,8 @@ export const Layout = ( {
 				<WelcomeModal
 					onClose={ () => {
 						updateOptions( {
-							woocommerce_task_list_welcome_modal_dismissed: true,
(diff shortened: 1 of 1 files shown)


octaedro (commented):
Nice job @samueljseay, it's a good idea using `yes`/`no` instead of a `boolean`.
Shouldn't we add that change also [here](https://github.com/woocommerce/woocommerce-admin/blob/main/client/homescreen/layout.js#L168)?
Otherwise, it should continue showing the modal.

samueljseay:
@octaedro wow it's checking for a 1 there... Yes I'll make that consistent. Thanks for spotting that

samueljseay:
After some more testing on an Atomic site, this isn't working as I thought it was. I'll close this for now.

samueljseay:
@octaedro you were right it was indeed still showing the modal after further testing. (I feel like I'm going crazy because I saw it not come back). Anyway, I've fixed that now and the modal *definitely* does not come back.

octaedro:
> @octaedro you were right it was indeed still showing the modal after further testing. (I feel like I'm going crazy because I saw it not come back). Anyway, I've fixed that now and the modal _definitely_ does not come back.

@samueljseay it looks like the commit 390c7c1 adds the change we needed but also undoes the first modification.

samueljseay:
@octaedro 😰 so sorry for the muck around. must have needed another coffee. Fixed.

octaedro (approved):
Now it's working perfectly! Code looks good and is testing well here 🚀 
Good job @samueljseay!

octaedro:
> @octaedro 😰 so sorry for the muck around. must have needed another coffee. Fixed.

No worries! 🙂

timmyc:
cherry picked to `release/1.6.0` in 6bddca2fea3f605ae8986a015b6c2fe0eb78a169

PR #5153, opened 2020-09-17
Update/components deps

psealock:
In working on https://github.com/woocommerce/navigation/issues/91 I added `@woocommerce/components` dependency which caused the following error:

Screen Shot 2020-09-17 at 9 01 06 PM

This was caused by a mis-matched dependency as `packages/components/package.json` had not been updated. The Currency package had a breaking change, causing the above error. This PR updates to the latest for all wc-admin packages.

If you are really, really keen you can create any application with `@woocommerce/components` as a dependency and see the error. Then, switch to this branch of wc-admin and use npm link ([instructions here](https://github.com/woocommerce/navigation)) to see the error disappear. Note, you'll need to carefully install, build and link in that order.

Or... you can double check the updated versions for data, currency, and navigation are correct 😄 . Since these would be the same versions that wc-admin itself uses on master, its safe to say they are matched correctly.

This PR also preps for a package release point because `@woocommerce/components` is broken.

### Changelog Note:

- Dev: Bump `@woocommerce/components` dependencies.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7973,10 +7973,10 @@
 			"requires": {
 				"@babel/runtime-corejs2": "7.11.2",
 				"@woocommerce/csv-export": "1.2.0",
-				"@woocommerce/currency": "2.0.0",
-				"@woocommerce/data": "1.0.0",
+				"@woocommerce/currency": "3.0.0",
+				"@woocommerce/data": "1.1.0",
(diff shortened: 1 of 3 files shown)


octaedro (approved):
Nice job @psealock. Changes look good and everything is working well :shipit:

PR #5157, opened 2020-09-17
Store Profiler - Industry: reduced padding and removed industry

### LINKED ISSUE #5155: Store Profiler - Industry: reduce padding and remove industry
As a follow up of #5064, we're going to reduce 4px in the top/bottom padding (I think the CSS class is `woocommerce-profile-wizard__checkbox`) and remove the Art, Music & Photography industry to check if we can bring back the Industry step to the same conv. levels that we had before the change.

octaedro:
Fixes #5155

This PR aims to reduce the size of the industries list in the second step of the Store profiler.
- The top/bottom padding of each list item is reduced by 4px
- The industry `Art, Music & Photography` is removed from the list.

### Screenshots
screenshot-one wordpress test-2020 09 17-17_46_55

- Go to the second step of the OBW (URL: `/wp-admin/admin.php?page=wc-admin&path=%2Fsetup-wizard&step=industry`)
- Verify the item `Art, Music & Photography` is not visible.
- Verify that the list looks like the image above.

### Changelog Note:
Dev: Store Profiler - Industry step: reduced padding and removed industry

### DIFF
--- a/client/profile-wizard/style.scss
+++ b/client/profile-wizard/style.scss
@@ -291,8 +291,8 @@
 			margin-top: 0;
 			margin-bottom: 0;
 			position: relative;
-			padding: $gap $gap-large;
-			min-height: 56px;
+			padding: $gap-small $gap-large;
+			min-height: 44px;
 			border-bottom: 1px solid $gray-100;
 
 			.components-base-control {
--- a/src/Features/Onboarding.php
+++ b/src/Features/Onboarding.php
@@ -315,11 +315,6 @@ class Onboarding {
 					'use_description'   => false,
 					'description_label' => '',
 				),
-				'art-music-photography'           => array(
-					'label'             => __( 'Art, music, and photography', 'woocommerce-admin' ),
-					'use_description'   => false,
-					'description_label' => '',
-				),
 				'electronics-computers'           => array(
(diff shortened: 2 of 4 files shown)


samueljseay:
@octaedro Just a tip that your e2e tests are failing because the count of industries has changed. one test failing has a cascading effect to fail the others at the moment since some are dependent on others. It should just be a matter of changing the number of industries:

[here](https://github.com/woocommerce/woocommerce-admin/blob/main/tests/e2e/specs/activate-and-setup/complete-onboarding-wizard.test.js#L47)

and also [here](https://github.com/woocommerce/woocommerce-admin/blob/main/tests/e2e/specs/activate-and-setup/complete-industry-section.js#L12)

octaedro:
> @octaedro Just a tip that your e2e tests are failing because the count of industries has changed. one test failing has a cascading effect to fail the others at the moment since some are dependent on others. It should just be a matter of changing the number of industries:

Thank you @samueljseay, nice catch! I fixed the e2e tests in the commit 887d721

samueljseay (approved):
Looks good and testing well for me. 🚀

timmyc:
cherry picked to `release/1.6.0-rc.1` in 91dde96

PR #5158, opened 2020-09-18
Remove ` @woocommerce/wc-admin-settings` from data package

psealock:
The data package made a reference to `@woocommerce/wc-admin-settings` which is a webpack alias in wc-admin and not an actual package. This means usage outside of the wc-admin context will cause fatal errors.

https://github.com/woocommerce/woocommerce-admin/blob/aeb6b9801d21cf382bfedf879913eea5614c7b89/webpack.config.js#L162-L165

This PR remove that dependency by passing in `getAdminLink` to plugin store actions. Changes are also made in preparation for releasing `@woocommerce/data` and `@woocommerce/components` which uses the data package.

1. `npm run test:zip`
2. Fire up a Jurassic Ninja site
3. Connect to Jetpack from either the Homescreen or OBW

### Changelog Note:

- Dev: Remove `getAdminLink` from data package


becdetat (approved):
Everything looks good :shipit:

PR #5161, opened 2020-09-18
Fix a bug where the activity panel options flicker on non-homescreen pages.

samueljseay:
While investigating another issue I noticed a display bug in ActivityPanel. On non-homescreen pages while resolving task list options from the API it displays one set of tabs and then another once they have resolved, resulting in a janky experience.

This has been resolved here by displaying no tabs while resolving the task list options. Although there is still a delay in the tabs shown it at least doesn't flicker between 2 sets of options.

1. Complete the task list on the homescreen.
2. Go to "Orders" page, note that there should be no flicker between 2 sets of tabs in the `ActivityPanel`

### Changelog Note:

Fix: stop a flicker of 2 sets of options on the ActivityPanel on non-homescreen pages.

### DIFF
--- a/client/header/activity-panel/index.js
+++ b/client/header/activity-panel/index.js
@@ -51,6 +51,7 @@ const ReviewsPanel = lazy( () =>
 
 const manageStock = getSetting( 'manageStock', 'no' );
 const reviewsEnabled = getSetting( 'reviewsEnabled', 'no' );
+
 export class ActivityPanel extends Component {
 	constructor( props ) {
 		super( props );
@@ -136,6 +137,12 @@ export class ActivityPanel extends Component {
 			! query.path &&
 			( requestingTaskListOptions === true ||
 				( taskListHidden === false && taskListComplete === false ) );
+
+		// If still loading the task list options present an empty list to prevent
+		// flickering between 2 lists of options.
+		if ( requestingTaskListOptions && ! isPerformingSetupTask ) {
+			return [];
+		}
 
 		if ( ! taskListComplete && showInbox ) {
 			return [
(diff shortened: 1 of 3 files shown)


psealock (commented):
Nice, thanks @samueljseay. This is working nicely! 

I was wondering if we shouldn't preload those options that make up `requestingTaskListOptions`.

https://github.com/woocommerce/woocommerce-admin/blob/63062d0a227cc39eb923b880f8f44e9c23f88374/client/header/activity-panel/index.js#L394-L396

Sure enough, they are.

https://github.com/woocommerce/woocommerce-admin/blob/63062d0a227cc39eb923b880f8f44e9c23f88374/src/Features/Onboarding.php#L626

But maybe that onboarding file isn't loaded on Core pages, so I checked the console and indeed they exist on the client

Screen Shot 2020-09-23 at 1 11 36 PM

I'm wondering if the Options data store isn't being properly hydrated on `EmbedLayout` renders to avoid an unnecessary resolve?

samueljseay:
@psealock I see a REST request being done to fetch these options, so I'm guessing that does suggest hydration is not working as expected. I can investigate this further.

psealock:
I think changes in this PR are good still. By hydrating the options data store, this is what I meant: https://github.com/woocommerce/woocommerce-admin/pull/5161/commits/fbe9542d13e593caecac6d854efa71d7d45f240d

Its also done in the layout, so this can be refactored such that everything is hydrated only once, which is probably better so we avoid this situation in the future.

https://github.com/woocommerce/woocommerce-admin/blob/fbe9542d13e593caecac6d854efa71d7d45f240d/client/layout/index.js#L219

samueljseay:
@psealock thanks so much for that. I ended up needing to refactor this piece of code for another PR: #5180 So I'm thinking maybe I can circle back on this with your suggested changes after that is merged.

samueljseay:
I'm closing this for now because I want to see if the behaviour still exists with the refactoring that was done in #5180 and then refactor in a new PR if needed.

psealock:
Sounds good. I do think we should change our hydration in such a way they happen in the same place which is high in the node tree so this type of thing doesn't happen.

PR #5162, opened 2020-09-18
Add Opt-Out for Remote Inbox Notifications

### LINKED ISSUE #5140: Remote Inbox: Don't Run Remote Inbox if site is opted out of Marketplace Suggestions
Currently [we have a check in NoteTraits](https://github.com/woocommerce/woocommerce-admin/blob/main/src/Notes/NoteTraits.php#L61) to not create an inbox notification if a site has opted out of receiving Marketplace Suggestions. As we prep to enable Remote Inbox notifs in the core package, I would like for us to use the `woocommerce_show_marketplace_suggestions` option to conditionally run the Remote Inbox poller. For example, if I opt out of Marketplace Suggestions, I feel we shouldn't even make the request to the Remote Inbox feed since those notes will never be created.

This also gives users a way to fully opt out of the remote inbox system via the UI if they prefer to not have a scheduled task fetch an asset from a remote site.

adrianduffell:
Fixes #5140

This adds the ability for the site to opt-out of remote inbox notifications. It uses the existing marketplace suggestions opt-out setting. When opted out, no connections are made to the remote notifications data source.

- Install and activate all of the following extensions. These are currently needed to match the notification rules in the remote feed. 
  - Australia Post Shipping Method
  - Canada Post Shipping Method
  - Royal Mail.
- Manually delete existing notes out of the database table, e.g. `wp_wc_admin_notes`.
- In **Settings > Advanced > wooCommerce.com**, opt-out of Marketplace suggestions.

Screen Shot 2020-09-18 at 12 33 16 pm

- Manually trigger the `wc_admin_daily` scheduled event. For example with WP CLI `wp cron event run wc_admin_daily`.
- Confirm the `Need help setting up your Store?` notification **DOES NOT** appear in the inbox.

- In **Settings > Advanced > wooCommerce.com**, opt-into Marketplace suggestions.

Screen Shot 2020-09-18 at 12 28 35 pm

- Manually trigger the `wc_admin_daily` scheduled event. For example with WP CLI `wp cron event run wc_admin_daily`.
- Confirm the `Need help setting up your Store?` notification  appears in the inbox.

Screen Shot 2020-09-18 at 11 51 31 am

### Changelog Note:

Enhancement: Remote inbox notifications can be opted out of using the Marketplace suggestions setting.

### DIFF
--- a/src/Events.php
+++ b/src/Events.php
@@ -110,9 +110,29 @@ class Events {
 		WC_Admin_Notes_Edit_Products_On_The_Move::possibly_add_note();
 		WC_Admin_Notes_Performance_On_Mobile::possibly_add_note();
 
-		if ( Loader::is_feature_enabled( 'remote-inbox-notifications' ) ) {
+		if ( $this->is_remote_inbox_notifications_enabled() ) {
 			DataSourcePoller::read_specs_from_data_sources();
 			RemoteInboxNotificationsEngine::run();
 		}
+	}
+
+	/**
+	 * Checks if remote inbox notifications are enabled.
+	 *
+	 * @return bool Whether remote inbox notifications are enabled.
+	 */
+	protected function is_remote_inbox_notifications_enabled() {
+		// Check if the feature flag is disabled.
+		if ( ! Loader::is_feature_enabled( 'remote-inbox-notifications' ) ) {
+			return false;
+		}
+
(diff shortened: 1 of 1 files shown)


becdetat (approved):
Looks great! :shipit:

PR #5163, opened 2020-09-18
Avoid PHP8 warning by declaring __wakeup() as public.

dd32:
PHP requires that all magic methods be public. As of PHP8, `__wakeup()` being declared as non-public causes a warning.

```
php -l woocommerce/packages/woocommerce-admin/src/FeaturePlugin.php

PHP Warning:  The magic method Automattic\WooCommerce\Admin\FeaturePlugin::__wakeup() must have public visibility in woocommerce/packages/woocommerce-admin/src/FeaturePlugin.php on line 314
```

Attached **untested** patch simply moves it to public, calls `die()` for safety.

See also: https://github.com/Automattic/jetpack/pull/17200 https://github.com/woocommerce/woocommerce/pull/27707

### Accessibility

- Ex: Open page `url`
- Click XYZ…

### DIFF
--- a/src/FeaturePlugin.php
+++ b/src/FeaturePlugin.php
@@ -313,5 +313,7 @@ class FeaturePlugin {
 	/**
 	 * Prevent unserializing.
 	 */
-	private function __wakeup() {}
+	public function __wakeup() {
+		die();
+	}
 }


becdetat (approved):
Hi @dd32, I could reproduce the issue and the fix is testing well under PHP 8. Thanks for this!

becdetat:
I've recreated this PR in https://github.com/woocommerce/woocommerce-admin/pull/5211 so that CI runs correctly, so I'll close this PR. Thanks again, @dd32 :-)

PR #5164, opened 2020-09-18
Add click handlers to items registered with WC Nav

### LINKED ISSUE #5098: Navigation items in the wc nav should register click handlers
Note: wait on decision for compositional structure in https://github.com/WordPress/gutenberg/pull/25057 before tackling this issue.

WooCommerce Admin should register its own click handlers or components for use in the wc nav project.  Currently, clicking these links navigates to the page using a standard link without using the react router in WCA.

There's a filter for the menu items that can be used to modify the data: `woocommerce_navigation_menu_items`.  It's probably easiest to pass the component used for linking directly to the menu items using the `LinkComponent` prop.

```js
{
  id: 'menu-item',
  title: 'My menu item',
  LinkComponet: NavLink,
}
```

Screen Shot 2020-09-04 at 11 54 09 AM

joshuatf:
Fixes #5098 

Adds click handlers for items registered with the WooCommerce Navigation.

### Screenshots

Screen Shot 2020-09-18 at 3 25 36 PM

1. Check out ~this PR in GB - https://github.com/WordPress/gutenberg/pull/25466~ master, branch has been merged.
1. In GB: `npm run build` then `cd packages/components && yarn link`.
1. Check out https://github.com/woocommerce/navigation/pull/93
1. In wc nav: `yarn link @wordpress/components`
1. Check that WCA items can be navigated to using the click handlers (react router).
1. Optionally, rebase from https://github.com/woocommerce/navigation/pull/89 to see WCA items get updated to the active state on click.

### DIFF
--- a/client/layout/controller.js
+++ b/client/layout/controller.js
@@ -4,7 +4,7 @@
 import { Component, Suspense, lazy } from '@wordpress/element';
 import { parse, stringify } from 'qs';
 import { find, isEqual, last, omit } from 'lodash';
-import { applyFilters } from '@wordpress/hooks';
+import { addFilter, applyFilters } from '@wordpress/hooks';
 import { __ } from '@wordpress/i18n';
 import {
 	getNewPath,
@@ -222,6 +222,27 @@ export class Controller extends Component {
 }
 
 /**
+ * Get a menu item's href with persisted queries.
+ *
+ * @param {string} url - URL of item.
+ * @param {Object} nextQuery - A query object to be added to updated hrefs.
+ * @param {Array} excludedScreens - wc-admin screens to avoid updating.
+ */
+export function getItemHref( url, nextQuery, excludedScreens ) {
+	const search = last( url.split( '?' ) );
+	const query = parse( search );
+	const path = query.path || 'homescreen';
+	const screen = path.replace( '/analytics', '' ).replace( '/', '' );
+
+	const isExcludedScreen = excludedScreens.includes( screen );
+
+	return (
+		'admin.php?' +
+		stringify( Object.assign( query, isExcludedScreen ? {} : nextQuery ) )
+	);
+}
+
+/**
  * Update an anchor's link in sidebar to include persisted queries. Leave excluded screens
  * as is.
  *
@@ -231,18 +252,7 @@ export class Controller extends Component {
  */
 export function updateLinkHref( item, nextQuery, excludedScreens ) {
 	if ( isWCAdmin( item.href ) ) {
-		const search = last( item.href.split( '?' ) );
-		const query = parse( search );
-		const path = query.path || 'homescreen';
-		const screen = path.replace( '/analytics', '' ).replace( '/', '' );
-
-		const isExcludedScreen = excludedScreens.includes( screen );
-
-		const href =
-			'admin.php?' +
-			stringify(
-				Object.assign( query, isExcludedScreen ? {} : nextQuery )
-			);
+		const href = getItemHref( item.href, nextQuery, excludedScreens );
 
 		// Replace the href so you can see the url on hover.
 		item.href = href;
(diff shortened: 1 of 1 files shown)


psealock (commented):
This reminds me of the config v. composibility discussions during the earlier PRs of the Gutenberg Nav project. This PR representing the config approach and [this one](https://github.com/woocommerce/navigation/pull/95) the composibility one. 

> I'm wary of introducing a dependency on WCA into the nav

Either approach will intertwine these two repos. It feels to me that having the logic of url construction in wc-admin creates some magic that may be difficult to debug in the future. Some knowledge about the `woocommerce_navigation_menu_items` hook and its use in wc-admin will be required to effectively debug any problems because the logic resides in two repos.

Its also important to understand that a pattern set forward here is one to be promoted to extension developers looking to include their wc-adminified pages in the new Navigation. Following the pattern set in this PR will require that they render a `NavigationItem` for inclusion via slotFill and in a different part of the application use the `woocommerce_navigation_menu_items` hook to properly add in an `onClick` property that prevents default and manipulates the history. I'm not sure this is something we should promote.

I understand that using Link provides an avenue to interacting with a different `History` if `@woocommerce/components` is bundled instead of used as an external. However, the same problems arise if we ask extension developers to use `onClick` as they will need to import `getHistory` from `@woocommerce/navigation`. At least a set interface can prevent mis-use and provide a uniform experience where we can `console.warn` against things like bundling `Link`

Ideally, some work to `Link` in wc-admin could make it so developers can simply add the Link as child content and internal logic in Link can determine how to handle urls (wc-admin to wc-admin, or legacy to wc-admin). I also think there is a case for moving `Link` to the Navigation package instead of the Component package.

```js
<NavigationItem>
    <Link />
</NavigationItem>
```
What do you think @joshuatf ?

joshuatf:
>This reminds me of the config v. composibility discussions during the earlier PRs of the Gutenberg Nav project.

I'm open to moving this towards composability which should be fairly trivial to do in this PR.  I opted to keep it config based since that's what's currently done with WCA links, but I do prefer the idea of using components here instead.

>Either approach will intertwine these two repos

This does not create a dependency on the wc nav repo, it just hooks into the wc nav hook if present.

I'm against importing WCA components into the wc nav project because this does create a dependency, adds the possibility for breaking changes between WCA versions, and means that the wc nav contains knowledge about wc nav links and routing.

>It feels to me that having the logic of url construction in wc-admin creates some magic that may be difficult to debug in the future. 

This may be true, but I think it is more of a WCA issue.  This PR uses the existing WCA URL constructors that add persistent queries to the URL.  If we decide to change this, then I think it should be a separate PR and should remain in WCA.

>Its also important to understand that a pattern set forward here is one to be promoted to extension developers looking to include their wc-adminified pages in the new Navigation.

I agree that we should set up the correct pattern that we'd like other developers to use.  This is why I strongly feel that all plugins (including WCA) should be responsible for hooking in and adding their own nav items.  If we add these directly to the nav repo, we're setting a precedent that the WC Nav repo is responsible for adding all third party menu items.

I believe using this method, a plugin creating a page through WCA will not require further steps to register click handlers to nav items.  Though I agree that using components sets a better example than `onClick` handlers for other plugins that may be using their own routers and components not in WCA.

>Following the pattern set in this PR will require that they render a NavigationItem for inclusion via slotFill 

Could you expand on where they would also need to add a `NavigationItem` for slot fill?

>and in a different part of the application use the woocommerce_navigation_menu_items hook to properly add in an onClick property that prevents default and manipulates the history. I'm not sure this is something we should promote.

I may be missing a piece here, but currently there is no slot fill necessary (though we could use this in place of the config based approach).  Items are currently registered both client-side and server-side though and I feel there is some redundancy here.  I'm not sure if there's an easy way to overcome this.

>I understand that using Link provides an avenue to interacting with a different History if @woocommerce/components is bundled instead of used as an external. However, the same problems arise if we ask extension developers to use onClick as they will need to import getHistory from @woocommerce/navigation.

As touched on above, if developers are extending screens through WCA, they do not need to do additional work client-side to add click-handlers; this PR should add them to any registered WCA pages.

If they are creating a screen outside of WCA, they will need to use their own history instance since it won't be connected to WCA.

joshuatf:
Closing this out in favor of using SlotFill.

PR #5166, opened 2020-09-18
Show payments task for non-US stores when wcpay installed

joshuatf:
The payment task is hidden when WC Pay is installed, which means that non-US stores see no payments on ecomm plans.  This always shows the normal payments task if the store is non-US based.

### Screenshots
Screen Shot 2020-09-18 at 3 48 40 PM

1. Install wcpay (does not have to be active).
1. Check that in the US the payments task is not shown, but the wc pay task is shown.
1. Change to a non-US store.
1. Check that the payments task is now shown.

### DIFF
--- a/client/task-list/tasks.js
+++ b/client/task-list/tasks.js
@@ -191,7 +191,7 @@ export function getAllTasks( {
 				} );
 				updateQueryString( { task: 'payments' } );
 			},
-			visible: ! woocommercePaymentsInstalled,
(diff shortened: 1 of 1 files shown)


octaedro (approved):
Nice job @joshuatf. This is testing well here and code LGTM 🚀

timmyc:
cherry picked to `release/1.6.0-rc.1` in 8191d38

PR #5167, opened 2020-09-18
Add Variations Report

### LINKED ISSUE #4330: Filter Product and Order Reports by Variation

jeffstieler:
Fixes #4330.

This PR seeks to introduce a Variations Report.

Additionally it:
* Fixes the display of variable products in the Orders report
* Allows for `compareParam` values other than `filter` in the `ReportTable` component

### Accessibility

### Screenshots

Screen Shot 2020-09-18 at 3 54 57 PM

Screen Shot 2020-09-18 at 3 55 15 PM

Screen Shot 2020-09-18 at 3 55 36 PM

Verify the requirements from design:

* Same general layout as product report, but uses product name / variation by default 
* Lists ALL product variations  
* Counts in orders column deep links to orders report filtered by variation 
* Show menu:
  * All variations (default) 
  * Single variation 
  * Comparison 
  * Advanced filters
    * Product (scoped to variable products only)
    * Category 
    * Attribute / terms 

### Changelog Note:

Enhancement: add Variations report.


client/lib/async-requests/index.js
```diff
@@ -77,14 +77,43 @@ export const getTaxRateLabels = getRequestByIdString(
...
+ * attribute option strings.
+ *
+ * @param {Object} variation - variation returned by the api
+ * @param {Array} variation.attributes - attribute objects, with option property.
+ * @param {string} variation.name - name of variation.
+ * @return {string} - formatted variation name
+ */
+export function getVariationName( { attributes, name } ) {
+	const separator =
+		window.wcSettings.variationTitleAttributesSeparator || ' - ';
```
psealock, line 91: May as well use `import { getSetting } from '@woocommerce/wc-admin-settings';` here.

packages/components/src/search/autocompleters/variations.js
```diff
@@ -93,41 +93,69 @@ import ProductImage from '../../product-image';
...
- * @return {string} - variation name
+ * @param {Array} variation.attributes - attribute objects, with option property.
+ * @param {string} variation.name - name of variation.
+ * @return {string} - formatted variation name
  */
-function getVariationName( variation ) {
-	return variation.attributes.map( ( { option } ) => option ).join( ', ' );
+function getVariationName( { attributes, name } ) {
+	const separator =
+		window.wcSettings.variationTitleAttributesSeparator || ' - ';
```
psealock, line 102: same here. Or is this the same function? Can it be placed somewhere reusable?

src/API/Reports/Orders/Stats/Controller.php
```diff
@@ -39,30 +39,32 @@ class Controller extends \Automattic\WooCommerce\Admin\API\Reports\Controller {
...
+		$args['variation_excludes'] = (array) $request['variation_excludes'];
+		$args['coupon_includes']    = (array) $request['coupon_includes'];
+		$args['coupon_excludes']    = (array) $request['coupon_excludes'];
+		$args['tax_rate_includes']  = (array) $request['tax_rate_includes'];
+		$args['tax_rate_excludes']  = (array) $request['tax_rate_excludes'];
+		$args['customer']           = $request['customer'];
+		$args['refunds']            = $request['refunds'];
+		$args['attribute_is']       = (array) $request['attribute_is'];
+		$args['attribute_is_not']   = (array) $request['attribute_is_not'];
+		$args['category_includes']  = (array) $request['categories'];
```
psealock, line 66: If we're making the change from `categories` to `category_includes`, does it make sense to also do that on the client too? I understand this may be a far reaching change and I'm not fully understanding why the change was made, but wondering if uniformity may be helpful in the future.

src/API/Reports/Products/Controller.php
```diff
@@ -38,7 +38,9 @@ class Controller extends \WC_REST_Reports_Controller implements ExportableInterf
 	 * @var array
 	 */
 	protected $param_mapping = array(
-		'products' => 'product_includes',
+		'categories' => 'category_includes',
+		'products'   => 'product_includes',
+		'variations' => 'variation_includes',
```
psealock, line 43: Oh ok, I see. My comment about renaming `categories` is probably handled here.

psealock (commented):
Wow 😮  @jeffstieler this is an impressive changeset! Nice work here, this touches A LOT of logic. I noted below an error I'm getting but its likely unrelated. Otherwise, this is testing well so far. I played around with Variations, Products, and Categories reports with various filters (Single, Advanced, Comparison, All) and no errors so far 🎉 . I'll do some manual checking on numbers tomorrow.

Unrelated to this PR, but I'm getting an error due to an orphaned variation. Maybe I deleted the parent at some point.

```
PHP Fatal error:  Uncaught Error: Call to a member function get_variation_attributes() on boolean in /srv/www/tangaroa/public_html/wp-content/plugins/wc-admin/src/API/Reports/Variations/DataStore.php:232
```
We can just add a check to make sure `$variation_product` actually exists before calling `get_variation_attributes`. If the solution is more involved, I can get some steps to reproduce and create an issue.

https://github.com/woocommerce/woocommerce-admin/blob/852e80ddfd7f5f9331a82aec095cf277115dcb1d/src/API/Reports/Variations/DataStore.php#L227

src/API/Reports/Orders/Stats/Controller.php
jeffstieler: This was definitely confusing to me when I started working on this new report.

These "param mappings" handle differences between what the REST endpoint Controller expects vs. the Query/DataStore that backs it. Keeping the mapping rather than doing a full rename everywhere allows any bookmarks users have created to function as expected. _(A use case I seem to remember us talking about a lot before launch)_

This particular change regarding `categories` needed to happen because the category filter on the Variations Report introduce excluding by category, which doesn't exist for other reports. The mapping takes care of the fact that the Categories Report hits the Products endpoint when specifying categories using `categories` and the Variations Report will either specify `category_includes` or `category_excludes` based on the filter value.

client/lib/async-requests/index.js
jeffstieler: Ah, I must have looked at the wrong code to base this on. We still have some instances of directly accessing `window.wcSettings` - particularly when hydrating values in HoCs.. should we be using `getSetting` everywhere instead?

packages/components/src/search/autocompleters/variations.js
jeffstieler: `lib/async-requests` definitely duplicates logic from the autocompleters. I chose not to tackle that as this PR was already rather large 😅

jeffstieler:
> Unrelated to this PR, but I'm getting an error due to an orphaned variation.

While this wasn't introduced in this PR, it was a reasonably small effort to fix in 9bfc921.

client/lib/async-requests/index.js
jeffstieler: Should `getSetting()` from `@woocommerce/wc-admin-settings` be used in our published packages? It doesn't seem to be in any of them currently..

LevinMedia (approved):
👍

client/lib/async-requests/index.js
psealock: >Should getSetting() from @woocommerce/wc-admin-settings be used in our published packages? It doesn't seem to be in any of them currently..

No, because its not actually a package but a webpack alias instead. So `@woocommerce/wc-admin-settings` outside of wc-admin causes a fatal error.

>should we be using getSetting everywhere instead?

That whole paradigm was a mistake, especially exposing the `setSetting` method. The settings data store is a migration of its usage for items that are mutable. We left the rest because its a large change and they are items not likely to change during usage of wc-admin, ie site title. I think the hydration HOC's should stay using `window` because its clear they are an initial hydration only.

psealock (approved):
Pre-approving pending a `getSetting` change already mentioned.

Very nice to see this happen and excellent effort @jeffstieler !!

jozeflambrecht:
Any idea when this feature will be released? Is there some work left I can help with?

jeffstieler:
Hello @jozeflambrecht,

This will be released in WooCommerce Admin `1.7.0` (which should be included in WooCommerce `4.8.0` unless we hit delays).

> Is there some work left I can help with?

The code is complete and just needs to be part of a tagged release. Thanks for the offer though!

PR #5169, opened 2020-09-21
Navigation package: Declare script dependencies

psealock:
Using `@woocommerce/navigation` as an external in a separate project can cause an issue because it depends on `wp-url` and `wp-hooks` to be loaded first. This PR adds that info in `Loader.php` so those assets are loaded first.

1. Spin up a separate repo and configure webpack to use wc-navigation as an external available at `window.wc.navigation`. Use https://github.com/woocommerce/navigation/compare/add/wc-admin-dep?expand=1 as an example.
2. See it being consumed error free.

....OR

1. See that package depends on those items being available and feel good about it making sense.

https://github.com/woocommerce/woocommerce-admin/blob/cbf2dfb6435cdd68c49815e6c6ff3e743f3a3e4e/packages/navigation/src/index.js#L4-L7

### Changelog Note:

- Dev: Add appropriate script dependencies for `@woocommerce/navigation`

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -323,7 +323,7 @@ class Loader {
 		wp_register_script(
 			'wc-navigation',
 			self::get_url( 'navigation/index', 'js' ),
-			array(),
+			array( 'wp-url', 'wp-hooks' ),
 			$js_file_version,
 			true
 		);


samueljseay:
@psealock Could you help me understand this one?

If we just adjust the dependency array of `wc-navigation` in our plugin PHP code how does that help others consume this package standalone? Wouldn't it make more sense to have those dependencies listed in the package.json for the wc-navigation package, and then your webpack config is gonna point to globals to make it work and its your responsibility to add the dependency array you need to your main package script?

> Using @woocommerce/navigation as an external in a separate project

I think this might be where my misunderstanding lies, do you mean use wc-navigation on a project where wc-admin is already installed? ie not installing it directly as an npm package?

psealock:
>I think this might be where my misunderstanding lies, do you mean use wc-navigation on a project where wc-admin is already installed? ie not installing it directly as an npm package?

Our packages are meant to be used outside of the wc-admin context, or even WP context. 

>Wouldn't it make more sense to have those dependencies listed in the package.json for the wc-navigation package, and then your webpack config is gonna point to globals to make it work and its your responsibility to add the dependency array you need to your main package script?

Yep, you are correct! So wc-admin is a consumer of the script and has the responsibility to add the dependency array. This is what we do with components as well.

https://github.com/woocommerce/woocommerce-admin/blob/9f341c74775ae16d7eef22d264f83d381d6e3825/src/Loader.php#L365-L387

Gutenberg has a really nifty way to handle this where webpack is used to automatically read the code to populate a dependency array in PHP:

https://github.com/WordPress/gutenberg/tree/master/packages/dependency-extraction-webpack-plugin

We should be using it in wc-admin, but are not and its caused a few bugs. This would be a good cooldown issue. One reason is because it only reads WP packages and extra code is required to add WC packages. See our starter pack for how thats done.

https://github.com/woocommerce/woocommerce-admin/blob/main/bin/starter-pack/_webpack.config.js

samueljseay (approved):
🎉

PR #5170, opened 2020-09-22
Remove store registration checks that are no longer needed.

### LINKED ISSUE #4443: Hydration fails when data store registered multiple times
**Describe the bug**
The plugin, settings, and soon-to-come hydration HOCs set the initial state for a component so the state is readily available without needing to send a network request to resolve information.

Lazily loaded Webpack chunks appear to be chunking the stores in multiple files instead of sharing them, causing them to be re-registered when imported.  This then reset the state to the default state set in the store.

Not that there is a temporary fix for this in https://github.com/woocommerce/woocommerce-admin/pull/4433 but this should be handled by building the stores in their own module and that fix eventually removed.

Thanks to @psealock for discovering the cause and for @nerrad helping us debug the issue.

See p1590435699034400-slack-woo-start-manage for full discussion.

**To Reproduce**

1. Log the state of the plugins.
1. Visit a page, such as the Profiler, that is lazily loaded.
1. Note that plugins are initially hydrated and subsequently reset, then filled again after resolution.

**Expected behavior**
Expected the `registerStore` function to only be called once per store.

samueljseay:
Fixes #4443 

Some investigation found that #4443 is no longer a reproducible issue. Why this issue no longer exists is not clear, but it makes sense now to disable the store registration checks as they are only called once. This PR just removes the checks but does not change any other behaviour.

The simplest way to test this is to confirm that store registration does not happen multiple times per store. To do this place a `console.log` statement next to the site of the `registerStore` call. such as here: https://github.com/woocommerce/woocommerce-admin/blob/main/packages/data/src/onboarding/index.js#L23

The data store linked above is the onboarding one, so you could go through the onboarding wizard and look for your `console.log` statement. On any one full page load the `registerStore` call should happen just once. 

### Changelog Note:

Dev: remove checks of store registration that are no longer needed.

### DIFF
--- a/packages/data/src/export/index.js
+++ b/packages/data/src/export/index.js
@@ -2,7 +2,7 @@
  * External dependencies
  */
 
-import { select, registerStore } from '@wordpress/data';
+import { registerStore } from '@wordpress/data';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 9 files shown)


octaedro (approved):
Nice job @samueljseay!
I added some `console.log` to test this PR in the branch `fix/4443-store-rego_test`.
This is testing well here and code LGTM :shipit:

PR #5171, opened 2020-09-22
Remove Customer Analytics Data Upon Order Deletion

### LINKED ISSUE #4946: Analytics: Remove Guest Customer Data from Customers Table Upon Order Delete
While discussing data retention with various folks in #3810 - it was noted that we are not adhering to the Personal Data Retention Settings that WooCommerce offers:

image

Specifically the "Retain Completed Orders" setting.

The proposed solution is:

- When an order is deleted, regardless of reason, either via Retain Completed Orders period lapsing - or by the order object itself being deleted by an admin user, any related data in the Customers Reporting Table should be also deleted unless:
- If *another* order exists for the given `user_id` or guest checkout email address, the the customer reporting table should remain.

adrianduffell:
Fixes #4946 

This removes associated customer analytics data upon order deletion. When analytics data for an order is removed, a WP action is run: `woocommerce_analytics_delete_order_stats`. The Products, Coupon, and Taxes `DataStore` classes hook into this action to further remove their associated data. [Example here](https://github.com/woocommerce/woocommerce-admin/blob/aecd050aff498d997adea9ec3e00a6bce8bd8a61/src/API/Reports/Products/DataStore.php#L503).

This change applies same approach to the Customers `DataStore` to remove associated customer data.

The `woocommerce_analytics_delete_order_stats` action is modified to pass a new `$customer_id` argument. This was added because the Customer ID for an order cannot be reliably obtained from inside the hook. This is because the previous order data is unavailable since the action is run after the order is deleted.

A check is made to ensure data is deleted only if the customer has a single order belonging to it. This is to preserve the customer record in cases where the customer has other orders. A new method is added to the Customers `DataStore` to count the amount of orders for a customer. It is similar to the `get_oldest_orders` method, however does not exclude types of orders.

Tests are added for customer deletion and counting customer orders.

- With a new customer, create two or more orders.
- As an administrator, delete one order by moving it to the trash and permanently deleting it.
- Inspect the `*wc_customer_lookup` table and ensure the customer record exists.
- As an administrator, delete the remaining orders for the new customer.
- Inspect the `*wc_customer_lookup` table and ensure the customer record does not exist

### DIFF
--- a/src/API/Reports/Customers/DataStore.php
+++ b/src/API/Reports/Customers/DataStore.php
@@ -573,6 +573,34 @@ class DataStore extends ReportsDataStore implements DataStoreInterface {
 	}
 
 	/**
+	 * Retrieve the amount of orders made by a customer.
+	 *
+	 * @param int $customer_id Customer ID.
+	 * @return int|null Amount of orders for customer or null on failure.
+	 */
+	public static function get_order_count( $customer_id ) {
+		global $wpdb;
+		$customer_id = absint( $customer_id );
+
+		if ( 0 === $customer_id ) {
+			return null;
+		}
+
+		$result = $wpdb->get_var(
+			$wpdb->prepare(
+				"SELECT COUNT( order_id ) FROM {$wpdb->prefix}wc_order_stats WHERE customer_id = %d",
+				$customer_id
+			)
+		);
+
+		if ( is_null( $result ) ) {
+			return null;
+		}
+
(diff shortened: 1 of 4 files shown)


adrianduffell:
I'm planning to do a bit more testing on this before taking it out of draft.

becdetat (approved):
Looks good, works well, tests pass, there are tests, :shipit:

PR #5173, opened 2020-09-22
Fix header height and positioning for wc nav

### LINKED ISSUE #5172: Navigation height and position should match new WC nav experience
**Describe the bug**
When activating the new nav experience, the header position is too low to account for the previously existent wp admin bar.

Also when folding the new nav, the height of the header is mismatched slightly.  The WCA header height should be updated to 60px to match the new height.

**To Reproduce**

1. Install the new nav plugin.
1. Check out the branch on this PR - https://github.com/woocommerce/navigation/pull/101
1. Note the WCA header position on WCA pages.

**Expected behavior**
Expected the WCA header to be flush with the top of the window.

**Screenshots**
Screen Shot 2020-09-22 at 2 42 24 PM

joshuatf:
Fixes #5172 

Updates the header height to be 60px and puts it flush against the top of the window when the new nav experience is enabled.

#### Before
Screen Shot 2020-09-22 at 2 42 24 PM
Screen Shot 2020-09-04 at 6 13 52 PM

#### After
Screen Shot 2020-09-22 at 3 12 55 PM
Screen Shot 2020-09-22 at 3 12 37 PM

1. Fire up the nav project and check out https://github.com/woocommerce/navigation/pull/101
1. Check that the WCA header sits at the top of the screen.
1. Click the WordPress logo in the top left of the screen.
1. Make sure the folded nav logo/button and WCA header heights line up correctly.

### DIFF
--- a/client/header/activity-panel/style.scss
+++ b/client/header/activity-panel/style.scss
@@ -262,3 +262,7 @@
 .woocommerce-layout__notice-list-hide {
 	display: none;
 }
+
+.has-woocommerce-navigation .woocommerce-layout__activity-panel {
+	top: 0;
+}
--- a/client/header/style.scss
+++ b/client/header/style.scss
@@ -1,7 +1,6 @@
 .woocommerce-layout__header {
 	background: $studio-white;
 	box-sizing: border-box;
-	border-bottom: 1px solid $studio-white;
 	padding: 0;
 	min-height: $header-height;
 	position: fixed;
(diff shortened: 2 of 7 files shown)


psealock (commented):
Yes! This looks great. Changes look good, but the OBW will need some consideration for both the sidebar and header.

Screen Shot 2020-09-24 at 11 34 53 AM

joshuatf:
Good catch, @psealock - opted to update these styles via CSS instead of toggling classes via JS to avoid conflicts and prevent any unstyled flashes of content in https://github.com/woocommerce/woocommerce-admin/pull/5173/commits/d5a29ed50ff2a7a9192a9c101e1e83469b3d9cc9

psealock (approved):
Thanks for the fixes @joshuatf, this is looking good on wc-admin, Core, and OBW pages with and without WC Nav.

Issue #3445, opened 2019-12-18
Add DependencyExtractionWebpackPlugin to handle dependencies 


psealock:
Adding [DependencyExtractionWebpackPlugin](https://github.com/WordPress/gutenberg/tree/master/packages/dependency-extraction-webpack-plugin) can automatically handle dependencies so that enqueued script will always have the correct dependencies required.

[Dependency Extraction Webpack Plugin](https://github.com/WordPress/gutenberg/tree/master/packages/dependency-extraction-webpack-plugin) does a magical thing where arrays of dependencies are automatically created by webpack so that adding something like `import { Navigation } from "@wordpress/components";` automatically creates a dependency array in PHP so that when registering scripts, it is made sure that `@wordpress/components` is loaded first.

This is really handy and the lack of its use has caused bugs in the past. See https://github.com/woocommerce/woocommerce-admin/pull/3429.

To take this issue one step further, a Woo flavored version of this would be fantastic because plugins need to modify the usage in webpack to make it recognize WC packages. See  https://github.com/woocommerce/woocommerce-admin/blob/main/bin/starter-pack/_webpack.config.js

PR #5198, opened 2020-09-25
New Package: DependencyExtractionWebpackPlugin

psealock:
Partially Fixes https://github.com/woocommerce/woocommerce-admin/issues/3445

As `@woocommerce` packages continue to grow in number and usefulness, plugins wishing to consume them will continually need to amend webpack's [DependencyExtractionWebpackPlugin](https://github.com/WordPress/gutenberg/blob/master/packages/dependency-extraction-webpack-plugin/README.md) options `RequestToHandle` and `requestToExternal` so that dependencies can be handled correctly.

This PR introduces a woo-flavored DependencyExtractionWebpackPlugin that extends the WP version such that WC packages are also mapped automatically.

The best way to test this is to create a new extension using `npm run create-wc-extension`. A project is scaffolded out, including use of DependencyExtractionWebpackPlugin. Once this simple case is confirmed, we can start using this package in wc-admin itself.

1. `npm run build`. This builds the packages for consumption later when we mock our new package (See step 3).
1. `npm run create-wc-extension` and give your new plugin a name.
2. Change this line in `<your-new-plugin>/webpack.config.js` to point to the new package since it does not yet exist in npm. There's probably a nice `npm link` solution here, but my tinkering didn't reveal it and this is easy enough.

```diff
- const WooCommerceDependencyExtractionWebpackPlugin = require( '@woocommerce/dependency-extraction-webpack-plugin' );
+ const WooCommerceDependencyExtractionWebpackPlugin = require("../woocommerce-admin/packages/dependency-extraction-webpack-plugin/src/index");
```
3. `npm start`
4. See `<your-new-plugin>/build/index.asset.php` exist and declare an empty array of dependencies because there aren't any just yet.
5. Install some dependencies: `npm install @wordpress/components` and `npm install @woocommerce/navigation`.
6. Use those dependencies by adding this to `<your-new-plugin>/src/index.js`:

```js
import { Button } from "@wordpress/components";
import { getHistory } from "@woocommerce/navigation";

console.log(Button, getHistory);
```
7. restart the proces `npm start`
8. See `<your-new-plugin>/build/index.asset.php` now reflect two dependencies.
9. Check the built js and see the dependencies gathered from `this`, which is the `window`
9. Bonus: Activate the plugin and visit the homescreen. See our consoled components.
10. Confirm that `window.wp.components` exists (wc-admin doesn't load it).

### Changelog Note:

- Dev: Add WooCommerceDependencyExtractionWebpackPlugin package

### DIFF
--- a/bin/starter-pack/_package.json
+++ b/bin/starter-pack/_package.json
@@ -21,6 +21,7 @@
 	},
 	"devDependencies": {
 		"@wordpress/scripts": "^12.2.1",
-		"@woocommerce/eslint-plugin": "1.0.0-beta.0"
+		"@woocommerce/eslint-plugin": "1.0.0-beta.0",
+		"@wordpress/dependency-extraction-webpack-plugin": "^2.8.0"
 	}
 }
--- a/bin/starter-pack/_package.json
+++ b/bin/starter-pack/_package.json
@@ -22,6 +22,6 @@
 	"devDependencies": {
 		"@wordpress/scripts": "^12.2.1",
 		"@woocommerce/eslint-plugin": "1.0.0-beta.0",
-		"@wordpress/dependency-extraction-webpack-plugin": "^2.8.0"
+		"@woocommerce/dependency-extraction-webpack-plugin": "1.0.0-beta.0"
 	}
 }
--- a/packages/dependency-extraction-webpack-plugin/src/index.js
+++ b/packages/dependency-extraction-webpack-plugin/src/index.js
@@ -1,0 +1,83 @@
+const WPDependencyExtractionWebpackPlugin = require( '@wordpress/dependency-extraction-webpack-plugin' );
+
+const WOOCOMMERCE_NAMESPACE = '@woocommerce/';
+
+/**
+ * Given a string, returns a new string with dash separators converted to
+ * camelCase equivalent. This is not as aggressive as `_.camelCase` in
+ * converting to uppercase, where Lodash will also capitalize letters
+ * following numbers.
+ *
+ * @param {string} string Input dash-delimited string.
+ *
+ * @return {string} Camel-cased string.
+ */
+function camelCaseDash( string ) {
+	return string.replace( /-([a-z])/g, ( _, letter ) => letter.toUpperCase() );
+}
+
+const wooRequestToExternal = ( request ) => {
+	if ( request.startsWith( WOOCOMMERCE_NAMESPACE ) ) {
+		return [
+			'wc',
+			camelCaseDash( request.substring( WOOCOMMERCE_NAMESPACE.length ) ),
+		];
+	}
+};
+
+const wooRequestToHandle = ( request ) => {
+	if ( request.startsWith( WOOCOMMERCE_NAMESPACE ) ) {
+		return 'wc-' + request.substring( WOOCOMMERCE_NAMESPACE.length );
+	}
+};
+
+class DependencyExtractionWebpackPlugin extends WPDependencyExtractionWebpackPlugin {
+	externalizeWpDeps( _context, request, callback ) {
+		let externalRequest;
+
+		// Handle via options.requestToExternal first
+		if ( typeof this.options.requestToExternal === 'function' ) {
+			externalRequest = this.options.requestToExternal( request );
+		}
+
+		// Cascade to default if unhandled and enabled
+		if (
+			typeof externalRequest === 'undefined' &&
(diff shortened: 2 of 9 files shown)


bin/starter-pack/_package.json
```diff
@@ -1,25 +1,26 @@
...
+		"check-licenses": "wp-scripts check-licenses",
+		"lint:css": "wp-scripts lint-style",
+		"lint:js": "wp-scripts lint-js",
+		"lint:pkg-json": "wp-scripts lint-pkg-json",
+		"start": "wp-scripts start",
+		"test:e2e": "wp-scripts test-e2e",
+		"test:unit": "wp-scripts test-unit-js"
+	},
+	"devDependencies": {
+		"@wordpress/dependency-extraction-webpack-plugin": "^2.8.0",
```
nerrad, line 19: Aren't you wanting to include `@woocomerce/dependency-extraction-webpack-plugin` here?

packages/dependency-extraction-webpack-plugin/src/index.js
```diff
@@ -0,0 +1,83 @@
...
+			camelCaseDash( request.substring( WOOCOMMERCE_NAMESPACE.length ) ),
+		];
+	}
+};
+
+const wooRequestToHandle = ( request ) => {
+	if ( request.startsWith( WOOCOMMERCE_NAMESPACE ) ) {
+		return 'wc-' + request.substring( WOOCOMMERCE_NAMESPACE.length );
+	}
+};
```
nerrad, line 32: This makes the assumption that every import on the `@woocommerce/*` namespace will have an external (and dependency). Woo Blocks uses the `@woocommerce/*` namespace for a number of aliases only. It'd be good if there would be a way we could signal skipping over aliased namespaces via config.

nerrad (commented):
Love the idea here Paul!  I think it'll be nice for Woo packages to be able to simplify their Webpack configuration using this.

Just a couple comments for review.

bin/starter-pack/_package.json
psealock: Yes! Good eye @nerrad. I left it this way temporarily so that an initial `npm install` for testing this PR will work because that package does not yet exist yet.

psealock:
Thanks for the feedback @nerrad ! This is ready for another review.

>Woo Blocks uses the @woocommerce/* namespace for a number of aliases only

Good point. Instead of putting the onus on consumers to declare aliases, I added a build step to dynamically create a list of packages before packages are built in https://github.com/woocommerce/woocommerce-admin/pull/5198/commits/d0c73a80aeaec1ae70816675b1019ad57b1a7660. The list is used by `wooRequestToExternal` and `wooRequestToHandle` to check if the request should be an external instead of just checking if it begins with `@woocommerce/`.

>Aren't you wanting to include @woocomerce/dependency-extraction-webpack-plugin here?

Yup, this was for easier testing. I'll replace before merge.

packages/dependency-extraction-webpack-plugin/src/index.js
```diff
@@ -1,4 +1,5 @@
 const WPDependencyExtractionWebpackPlugin = require( '@wordpress/dependency-extraction-webpack-plugin' );
+const packages = require( '../assets/packages' );
```
joshuatf, line 2: I may have set things up wrong, but this is causing me some trouble.  Do I need to link something in order for this to work?

Screen Shot 2020-10-05 at 11 27 53 AM

This was after running `npm start` from the newly created extension directory.

joshuatf (commented):
This is a great addition (and clean-up) to the WC ecosystem.  Thanks for taking the time to do this, Paul 🥇 

Had a little trouble testing this with the latest commits, but code looks pretty good to me.  @nerrad or @senadir can you confirm this works for your use case?

packages/dependency-extraction-webpack-plugin/src/index.js
psealock: Thanks for taking this for a spin @joshuatf. The problem lies step 3 of the instructions, which assumed the sibling wc-admin plugin had built packages. I changed the instructions to reflect this if you could give it another shot.

The trouble lies in creating a dependency on a package that isn't published yet, so this will be a lot simpler once the package in on npm.

joshuatf (approved):
Woo!  Got it up and running and dependencies are being loaded correctly.  LGTM after a rebase to fix up any conflicts.  ✅

PR #5199, opened 2020-09-25
Search component: Behavior and styling tweaks

dechov:
A few slight changes to the Search component:

- https://github.com/woocommerce/woocommerce-admin/commit/44e51bb1900663dba0836cfb0ef8c49f1bd1468d fixes vertical misalignment of the "X" button inside the tag:

`master` | this branch
-- | --
misaligned-remove-tag-button | aligned-remove-tag-button

- https://github.com/woocommerce/woocommerce-admin/commit/70edf5074276cc2d5d05cd130a4086bad5d9457d addresses just a slight visual glitch with inconsistent letter spacing between renditions of the placeholder:

`master` | this branch
-- | --
inconsistent-letter-spacing | consistent-letter-spacing

- https://github.com/woocommerce/woocommerce-admin/commit/8cead2f590b352c1e28dcd63bfee8f9757a0b8f7 is a proposed change to the debounced search behavior, which has the filtered options update (prepending free text options in the cases I was testing) _before_ the debounce delay, in addition to _after_ resolution of search results. Unless I'm missing different usages of these props, the main effect is to have the options added (e.g. free text search) in real-time, without waiting for an update, and being able to hit enter without a potentially non-current query being added as a tag.

`master` | this branch
-- | --
debounced-filtering | immediate-filtering

(I think there may be a somewhat broader / adjacent concern here too, around indicating the loading status of search options and/or clearing them when the query changes, but the current behavior – caching of search results until new ones are in – is unaffected here, aside from being mixed with the updated filtered options.)

Side-note: I'd also made a brief attempt at addressing the autocomplete popping back up after reset – seen at the end of those screen captures, or when typing and immediately defocusing – but having the `reset` method set `this.activePromise` to `null` didn't seem to work.

### Accessibility

On a screen with a search input (such as WooCommerce » Customers), verify that the top search suggestion (and new tag, after pressing enter) tracks the current input without a delay, and that the input placeholder and "remove tag" button both appear properly.

### Changelog Note:
Fix: visual issues in the Search component.

### DIFF
--- a/packages/components/src/select-control/index.js
+++ b/packages/components/src/select-control/index.js
@@ -220,7 +220,20 @@ export class SelectControl extends Component {
 	}
 
 	search( query ) {
-		this.setState( { query, isFocused: true } );
+		const searchOptions = this.searchOptions || [];
+		const filteredOptions =
+			query !== null && ! query.length && ! this.props.hideBeforeSearch
+				? searchOptions
+				: this.getFilteredOptions( searchOptions, query );
+
+		this.setState( {
+			query,
+			isFocused: true,
+			selectedIndex: 0,
+			filteredOptions,
+			isExpanded: Boolean( filteredOptions.length ),
+		} );
+
 		this.updateFilteredOptions( query );
 	}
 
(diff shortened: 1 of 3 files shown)


samueljseay (approved):
These changes are awesome and testing well for me. Thanks for making these improvements!

PR #5200, opened 2020-09-25
Update store setup link to redirect to setup wizard

### LINKED ISSUE #4989: Home Screen: Enable store setup task list if user skips the wizard
Currently, the store setup task list is only enabled if the user doesn't skip the wizard. However, it can be valuable to automatically enable the task list when WC is activated:

- A new user can find the task list useful even if he doesn't want to share the store profile info
- He also may have bypassed the wizard without noticing it.

joshuatf:
Fixes #4989

Replaces an old flag `enable_onboarding` that no longer exists with the setup wizard path.

To resolve #4989 the WooCommerce screen that redirects or skips onboarding should also be updated to include `reset_profiler=0`, but it looks like all references to this file may have been removed so this is no longer needed.

https://github.com/woocommerce/woocommerce/blob/51029986b5b8e20933901750478cd0aee00e6c7b/includes/admin/class-wc-admin-setup-wizard.php#L367

1. Use a new site or trigger the onboarding profile note by making sure `skipped` and `completed` are not truthy in the `woocommerce_onboarding_profile` option in your wp_options table.
1. Run the `wc_admin_daily` cron event.
1. Note the onboarding profiler reminder note.
1. Click the `Continue Store Setup` to make sure it directs to the setup wizard.

### DIFF
--- a/src/Notes/OnboardingProfiler.php
+++ b/src/Notes/OnboardingProfiler.php
@@ -53,15 +53,15 @@ class Onboarding_Profiler {
 		$note->add_action(
 			'continue-profiler',
 			__( 'Continue Store Setup', 'woocommerce-admin' ),
-			wc_admin_url( '&enable_onboarding=1' ),
-			NOTE::E_WC_ADMIN_NOTE_UNACTIONED,
+			wc_admin_url( '&path=/setup-wizard' ),
+			Note::E_WC_ADMIN_NOTE_UNACTIONED,
 			true
 		);
 		$note->add_action(
 			'skip-profiler',
 			__( 'Skip Setup', 'woocommerce-admin' ),
 			wc_admin_url( '&reset_profiler=0' ),
-			'actioned',
+			Note::E_WC_ADMIN_NOTE_ACTIONED,
 			false
 		);
 		return $note;


samueljseay:
@joshuatf I know this change will work, but now I'm trying to test your PR and I can't figure a way that you would ever see this note, because you always get redirected to the setup wizard if you haven't completed or skipped it at some point.

Maybe a silly question... is this note even valid any more? I tried:

1. partially completing the OBW, still redirects on visiting homepage
2. completing it then setting the completed value to falsy, still redirects
3. skipping it then setting skipped value to falsy, still redirects

This was however testing with woocommerce core 4.5.2. I can't test woocommerce core in dev at the moment because of some local issues I'm having.

joshuatf:
Thanks for the review, @samueljseay 😄 

If you make sure that the profiler is neither completed nor skipped (can be done quickly by deleting `woocommerce_onboarding_profile`) then this note should be created.

The note can be seen in the inbox panel outside of the homepage (e.g., any Analytics page, any WC embed page, other WCA registered pages, etc).

Screen Shot 2020-09-29 at 9 13 49 AM

samueljseay (approved):
Sorry about the delay on this. Finally got it working. 🎉

timmyc:
Cherry picked to `release/1.6.0` in b12ce2caf6b1dddf6d791d8be7533e3948b05931

PR #5201, opened 2020-09-25
Updating Stripe key field validation to support test keys (#4391)

### LINKED ISSUE #4391: Onboarding: Stripe doesn't accept test keys
I tried using the new onboarding to enable Stripe as a payment method, however, I got an error when I tried to enter the test keys.

**To Reproduce**

1. Run the onboarding and reach the _Payment methods_ step.
2. Add Stripe.
3. Write test keys obtained from stripe.com.
4. Notice it doesn't allow you to save them.

**Expected behavior**
It would be nice that if it detects that the key starts with `pk_test_`, it automatically saves them as test keys.

**Screenshots**
imatge

joelclimbsthings:
Fixes #4391

Currently the form to configure Stripe keys accessed via the task list presents an error when test keys are used.

To resolve I've just updated the regex used to validate the Stripe keys to include the pattern(s) used for test keys as well. These are similar in format but have a `pk_test_abc` instead of `pk_live_abc`. I've also added a `[a-zA-Z0-9_]+` to the end of the regex to enforce that at least one additional alpha-numeric character is provided. 

I'd originally also expanded on the validation messages to include more detail into what was a valid key, but I noticed that the UI only really supports single-line validation strings. If they're too long then they overlap the field/button below on a mobile view. I've included a screenshot of this in the section below.

To resolve that I simply shortened the validation messages and am not currently providing much detail to the user on what exactly is wrong. I could plausibly throw in a CSS tweak and expand the validation messages again if that's a preferred solution, although it would likely be a bit hacky since absolute positioning is used for the validation messages.

### Screenshots
**With long validation messages**:
Screenshot_2020-09-25_15-50-53

**With shortened validation messages (current state)**:
image

- Pick a Stripe supported country.
- Make sure you have not selected "CBD" as an industry previously in the profiler.
- Go to the task list payments task.
- Click on Stripe
- Enter "pk_test_abc" in first field. No error should display.
- enter "sk_test_abc" in second field. No error should display.


client/task-list/tasks/payments/stripe.js
```diff
@@ -167,15 +167,21 @@ class Stripe extends Component {
...
 		const errors = {};
 
-		if ( values.publishable_key.match( /^pk_live_/ ) === null ) {
+		if (
+			values.publishable_key.match( /^pk_(live|test)_[a-zA-Z0-9_]+/ ) ===
+			null
+		) {
 			errors.publishable_key = __(
-				'Please enter a valid publishable key. Valid keys start with "pk_live".',
+				'Please enter a valid publishable key.',
```
octaedro, line 175: Nice catch Joel!
As this problem only happens on mobile, what do you think about adding a CSS code instead of cutting the text to solve it?
It should look like this code:
```
@include breakpoint( '<782px' ) {
	margin-bottom: $gap-small + $gap * 2 !important;
}
```
And probably it should be added [here](https://github.com/woocommerce/woocommerce-admin/blob/main/client/layout/style.scss#L95)

octaedro (commented):
You did a really nice job here @joelclimbsthings! Verifying how the text looks across devices was a great idea!
Code looks good, just left a small comment.
Well done!

client/task-list/tasks/payments/stripe.js
joelclimbsthings: Good call @octaedro ! That CSS snippet works great. As far as I can tell that `<Form>` component doesn't support the `className` property, or allow for custom classes. I could always add a wrapper but that doesn't seem ideal.

That said, I think you could make the argument that we should leave this CSS tweak as is without narrowing it to a particular form. I poked around and saw a couple instances of validation messages on mobile being a bit crowded:

image

Although not entirely overlapping.

samueljseay: @joelclimbsthings fwiw, the <Form> component doesn't actually render it's own component, it clones the children and renders them so you could add the class to a wrapper inside <Form>

samueljseay: > That said, I think you could make the argument that we should leave this CSS tweak as is without narrowing it to a particular form. I poked around and saw a couple instances of validation messages on mobile being a bit crowded:

Yes I think if you're easily uncovering other issues it may be safer to truncate the copy.

samueljseay (approved):
Looks good. And I think the option to keep the copy and adjust the CSS was a good call. 🚀

dechov:
@joelclimbsthings Didn't notice your PR, despite referencing the issue [recently](https://github.com/woocommerce/woocommerce-admin/issues/5225). Thanks for taking this on! A few notes:
- This change sets test keys but in live key fields (resulting admin notice [here](https://cloudup.com/cBJQyI76ACw)) – test settings have a `test_` prefix ([source](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/cac73fce691e1344b56651d856ea8e58d89c9ddc/includes/admin/stripe-settings.php#L44-L57)).
- There's a brand new (to this code) `testmode` setting (see https://github.com/woocommerce/woocommerce-admin/pull/5226) that also needs to be set accordingly.
- As for the validation error message, in my view we could just say: `Valid keys start with "sk_" or "rk_".` That's enough to identify the key type (in context of Stripe), at which point the mode is the only thing that would follow anyway.

PR #5203, opened 2020-09-28
Add rule processors for country and state

becdetat:
This adds rule processors for the selected (in the profiler) country and state.

- Add this test feed to `DataSourcePoller::DATA_SOURCES`: `https://gist.github.com/becdetat/ac0acbaa547d2af7c56d63f8fd8eee82/raw/51a8116a83faddce42e44a0e914a35ca90217386/base-location-test.json`
- Set the store location to something other than US - Texas
- A note with the title "This should appear in Texas stores" should _not_ appear
- Set the store location to US - Texas
- The note with the title "This should appear in Texas stores" _should_ appear

### Changelog Note:

Dev: Add remote inbox notification rule processors for country and state

### DIFF
--- a/src/RemoteInboxNotifications/BaseLocationCountryRuleProcessor.php
+++ b/src/RemoteInboxNotifications/BaseLocationCountryRuleProcessor.php
@@ -1,0 +1,55 @@
+<?php
+/**
+ * Rule processor that performs a comparison operation against the base
+ * location - country.
+ */
+
+namespace Automattic\WooCommerce\Admin\RemoteInboxNotifications;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Rule processor that performs a comparison operation against the base
+ * location - country.
+ */
+class BaseLocationCountryRuleProcessor implements RuleProcessorInterface {
+	/**
+	 * Performs a comparison operation against the base location - country.
+	 *
+	 * @param object $rule         The specific rule being processed by this rule processor.
+	 * @param object $stored_state Stored state.
+	 *
+	 * @return bool The result of the operation.
+	 */
+	public function process( $rule, $stored_state ) {
+		$base_location = wc_get_base_location();
+		if ( ! $base_location ) {
+			return false;
+		}
+
+		return ComparisonOperation::compare(
+			$base_location['country'],
+			$rule->value,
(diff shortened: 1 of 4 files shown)


src/RemoteInboxNotifications/BaseLocationCountryRuleProcessor.php
```diff
@@ -0,0 +1,55 @@
...
+
+	/**
+	 * Validates the rule.
+	 *
+	 * @param object $rule The rule to validate.
+	 *
+	 * @return bool Pass/fail.
+	 */
+	public function validate( $rule ) {
+		if ( ! isset( $rule->value ) ) {
```
samueljseay, line 45: This could be a one liner I think?

`return isset( $rule->value ) && isset( $rule->operation );`

samueljseay, line 45: This could be a one liner I think?

`return isset( $rule->value ) && isset( $rule->operation );`

src/RemoteInboxNotifications/BaseLocationStateRuleProcessor.php
```diff
@@ -0,0 +1,55 @@
...
+
+	/**
+	 * Validates the rule.
+	 *
+	 * @param object $rule The rule to validate.
+	 *
+	 * @return bool Pass/fail.
+	 */
+	public function validate( $rule ) {
+		if ( ! isset( $rule->value ) ) {
```
samueljseay, line 45: Again this could be a one liner.

samueljseay (approved):
Tests well for me, just some suggestions but not a big deal.

src/RemoteInboxNotifications/BaseLocationCountryRuleProcessor.php
becdetat: The pattern I've been sticking to is an if..return for each validation condition. I think it's easier to understand and extend if a new condition gets added later.

timmyc:
Noting I'm unclear if this should go in 1.6 - not a bug fix so leaning towards no. p1602179362104000-slack-C9K5T0XSP?thread_ts=1602173754.094900&cid=C9K5T0XSP.

PR #5205, opened 2020-09-28
Fix a bug where the store task list would remain after completion

### LINKED ISSUE #5138: Store setup link persists after setup tasks are completed
**Describe the bug**
After completing all store setup tasks and dismissing the Setup card on the home screen, the Store setup link in the app bar persists. Also, when clicking on the Store setup link it takes you back to the home screen with the Setup card showing again (after dismissing it.)

**To Reproduce**

1. Create a new test site and update to the latest release
2. Click on 'WooCommerce Home'
3. Complete all Store setup tasks and dismiss card
4. Click on 'Customers', 'Reports', etc
5. See the Store Setup link persist.

**Expected behavior**
Once the store setup task list is complete, the Store setup link in the app bar shouldn't appear at all.

**Screenshots**
Completed tasks 
Screen Shot 2020-09-15 at 8 31 42 AM

Store setup link in Orders
Screen Shot 2020-09-15 at 8 32 13 AM

Store setup link in Reports
Screen Shot 2020-09-15 at 8 32 41 AM

- OS: Mac OS 10.15.6
- Browser: Firefox 80.0.1 and Chrome 85.0.4183.102

samueljseay:
Fixes #5138. #5138's description is not quite accurate, because the task list should disappear itself after completion rather than requiring manual dismissal. This fixes the issue with dismissal as well as the setup link issue.

There is an issue on Atomic sites where for some reason the option `woocommerce_task_list_complete` is set to `no` at some point during the workflow of setting up a site. This is problematic for the task list because of some flawed logic in the component. Because the value gets set to `no`, the component can never update itself to the correct value. The logic [here](https://github.com/woocommerce/woocommerce-admin/blob/main/client/task-list/index.js#L73) always exits because `no` is a truthy value and the flawed logic would pass `no` as a prop value.

Everywhere else we coerce values like `no` and `yes` to a boolean before passing them to the component, so this matches that approach. Setting it to be a proper boolean means that it will correctly get updated when all the tasks a user has are completed (this check is done every time the component mounts or updates.)

1. Create an Atomic site
2. Install latest WooCommerce
3. `npm run test:zip` on this branch and upload/install the zip file to your Atomic site's plugins
4. Activate both WooCommerce and WooCommerce admin
5. Complete the setup wizard
6. Complete all the tasks in the setup task list
7. Expected: The setup task list should vanish after your last task is complete.
8. Visit Customers link, the store setup link should no longer display in the activity panel.
### Changelog Note:

Fix: Bug where sometimes the task list would not vanish after being completed.


samueljseay:
@timmyc I *think* this might need to be cherry picked.

timmyc (approved):
@samueljseay this is testing out well for me on an Atomic Ephemeral site, thank you for the fix.

timmyc:
Cherry picked to `release/1.6.0-rc.1` in 1b73fbb

PR #5206, opened 2020-09-28
Add option remote inbox notifications rule processor

becdetat:
This adds a remote inbox notifications rule processor that enables looking at any option, as required [here](https://teamisotope.wordpress.com/2020/07/02/remote-inbox-notifications/#comment-4342)

### Screenshots
image

- Add this URL to `DataSourcePoller::DATA_SOURCES`: `https://gist.github.com/becdetat/7b0de142b0e708354d546618496a1d26/raw/249a0a653d894443ef9e65d69ac687561a2bd730/option-test.json`
- Set the store currency to something other than USD
- The "This should appear in USD stores" note should _not_ appear
- Set the store currency to USD
- The "This should appear in USD stores" note should now appear

### Changelog Note:

Development: Add remote inbox notifications rule allowing access to any option


timmyc:
Adding the same note as I did on #5207 here:

@becdetat thanks so much for adding in these two new rule processors. Unfortunately though I think we are bit too past the code freeze to introduce new code like this. As such I'm going to clear this from the 1.6 milestone... just not comfortable introducing too much new change at this point in the release cycle.

src/RemoteInboxNotifications/OptionRuleProcessor.php
```diff
@@ -0,0 +1,54 @@
...
+	/**
+	 * Performs a comparison operation against the option value.
+	 *
+	 * @param object $rule         The specific rule being processed by this rule processor.
+	 * @param object $stored_state Stored state.
+	 *
+	 * @return bool The result of the operation.
+	 */
+	public function process( $rule, $stored_state ) {
+		$option_value = get_option( $rule->option_name );
```
joshuatf, line 23: I wonder if we should allow a default option to be passed here as part of the rule.

joshuatf (approved):
Very cool rule and love how easily it integrates with the rule processing system ❤️ 

One optional comment/question, but pre-approving.

src/RemoteInboxNotifications/OptionRuleProcessor.php
becdetat: It took me a while to get back to this, but that's a great idea :-) done in 6c3745395e89155a4ca55f2ee097e9bd916c4755

joshuatf (approved):
Nice, Bec!  Thanks for the addition and the tests 🥇

PR #5207, opened 2020-09-28
Add note status remote inbox notifications rule processor

becdetat:
This adds a remote inbox notifications rule that looks at the status of another note.

For example, a note might only be displayed if another note has not been actioned, to avoid hassling the user when an action has already been taken.

### Screenshots
image

- Start on a store with the "Connect with your audience" admin note _not_ actioned
- Add this url to `DataSourcePoller::DATA_SOURCES`: `https://gist.github.com/becdetat/df80073b4c7283c2be89b26d19723b36/raw/fb794f5916d915f1bde4addc9d8672a7e2789de1/note-status.json`
- The "This should appear when the marketing intro note is actioned" note should _not_ appear
- Action the "Connect with your audience" admin note
- The "This should appear when the marketing intro note is actioned" note should now appear

### Changelog Note:

Dev: Add note status remote inbox notifications rule processor

### DIFF
--- a/src/Notes/Notes.php
+++ b/src/Notes/Notes.php
@@ -282,4 +282,23 @@ class Notes {
 			}
 		}
 	}
+
+	/**
+	 * Get the status of a given note by name.
+	 *
+	 * @param string $note_name Name of the note.
+	 * @return string|bool The note status.
+	 */
+	public static function get_note_status( $note_name ) {
+		$data_store = \WC_Data_Store::load( 'admin-note' );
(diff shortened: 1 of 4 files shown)


timmyc:
@becdetat thanks so much for adding in these two new rule processors. Unfortunately though I think we are bit too past the code freeze to introduce new code like this. As such I'm going to clear this from the 1.6 milestone... just not comfortable introducing too much new change at this point in the release cycle.

joshuatf (approved):
Works as described!  LGTM 🚢

PR #5208, opened 2020-09-28
Add filter by variations in reports note

### LINKED ISSUE #5116: Inbox message: filter by product variations in order and product reports
This notification should be shipped alongside #4330

**Title:** New - filter by product variations in orders and products reports
**Content:** One of the most awaited features has just arrived! You can now have insights into each product variation in the orders and products reports.  
**Segmentation:** Users that have at least one variable product
**Timing:** After the user updates to the WC version that includes this feature)
**Cadence:** Once
**CTA leads to:** the support doc about the new filter
**CTA label:** Learn more

becdetat:
Fixes #5116

~Note that this is still missing the documentation URL for the new feature however the rest is fine for review, I'll just replace the URL once it turns up.~ This is ready for review.

This adds a note about the variations filter in the orders and products reports, which is shown for stores with at least one variable product.

### Screenshots
image

- Don't have any variable products. You can do this by Products -> Filter by product type -> Variable products, then deleting all the products in bulk.
- The "filter by product variations in orders and products reports" note should _not_ appear
- Add a variable product (or restore an existing deleted variable product)
- The "filter by product variations in orders and products reports" note should now appear.

### Changelog Note:

Dev: Add filter by variations in reports note


samueljseay (approved):
🚀 tests well, code looks good.

PR #5212, opened 2020-09-29
Change store details tooltip to a popover

### LINKED ISSUE #5189: [WC 4.6 Beta] Tooltip is getting cut off
**Describe the bug**
In the store profiler in Step 1, the text is no longer wrapping within the tooltips that appear on hover. The first one on step 1 is getting cut off since it's quite long.

**Expected behavior**
In the store profiler tooltips in Step 1, the text should wrap for better readability.

**Steps to reproduce the bug:**
- Create a new site or go to **Settings > Help > Setup Wizard** to access the setup wizard on an existing site
- Go to Step 1
- Hover over first tooltip and see error

**Screenshots**
Current tooltip in step 1:
Screen Shot 2020-09-23 at 12 31 57 PM

What the tooltip looks like with text wrapping:
Screen Shot 2020-09-13 at 8 26 40 PM

* Mac 10.15.6 Firefox

becdetat:
Fixes #5189

The tooltip seemed to be working as designed so rather than messing around with it too much I decided to use the `Popover` component instead (which is already in use for product types). I think the result is better and more semantically correct than a tooltip.

### Screenshots
Before:
image
After:
image

- Start the OBW
- On the store details page, click the information icon next to the "Tell us about your store and we'll get you set up in no time" text

### Changelog Note:

Dev: Change store details tooltip to a popover

### DIFF
--- a/client/profile-wizard/steps/store-details/index.js
+++ b/client/profile-wizard/steps/store-details/index.js
@@ -9,8 +9,6 @@ import {
 	CardFooter,
 	CheckboxControl,
 	FlexItem,
-	Icon,
-	Tooltip,
 	__experimentalText as Text,
 	Popover,
 } from '@wordpress/components';
@@ -42,7 +40,8 @@ class StoreDetails extends Component {
 		this.state = {
 			showUsageModal: false,
 			skipping: false,
-			isPopoverVisible: false,
+			isStoreDetailsPopoverVisible: false,
+			isSkipSetupPopoverVisible: false,
 		};
 
 		// Check if a store address is set so that we don't default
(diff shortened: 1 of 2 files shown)


psealock (commented):
I like the solution here, thanks @becdetat !

There's another one of these on the page, which was also being fickle for me. Nonetheless, I spun up VoiceOver to compare the two from an accessibility perspective.

Screen Shot 2020-09-29 at 9 59 26 PM


 I'm wondering if this one isn't missing an aria label because I couldn't access the content from the keyboard. Needless to say, I found the experience with the Popover to be much better from each perspective. What do you think about giving this one the same treatment?

becdetat:
@psealock 

> What do you think about giving this one the same treatment?

Sure, sounds good. The tooltip has the aria-label set on the inner `span` which may be why it wasn't picked up by voiceover.

psealock (approved):
Lovely, working nicely 🚢

timmyc:
Cherry picked to `release/1.6.0` in https://github.com/woocommerce/woocommerce-admin/commit/bda56752b50030a7c63d1396726f0143ffbe215e

PR #5215, opened 2020-09-29
Declare the note constants on the deprecated class.

jeffstieler:
This avoids triggering PHP errors when third party code references the old class constants. Like core: https://github.com/woocommerce/woocommerce/blob/master/includes/admin/notes/class-wc-notes-run-db-update.php#L154

Fixes this error:

```
PHP Fatal error:  Uncaught Error: Undefined class constant 'E_WC_ADMIN_NOTE_ACTIONED' in /srv/www/wordpress-default/public_html/wp-content/plugins/woocommerce/includes/admin/notes/class-wc-notes-run-db-update.php:270
```

- On `main`
- Trick WooCommerce into thinking it needs to update the DB by deleting the `woocommerce_db_version` option
- Go to WooCommerce > Home
- Note the PHP error
- Check out this branch
- Trick WooCommerce into thinking it needs to update the DB by deleting the `woocommerce_db_version` option
- Go to WooCommerce > Home
- Note no more PHP error

### Changelog Note:

N/A - unreleased bug

### DIFF
--- a/src/Notes/DeprecatedNotes.php
+++ b/src/Notes/DeprecatedNotes.php
@@ -16,6 +16,21 @@ use Automattic\WooCommerce\Admin\DeprecatedClassFacade;
  * WC_Admin_Note (deprecated, use Note instead).
  */
 class WC_Admin_Note extends DeprecatedClassFacade {
+	// These constants must be redeclared as to not break plugins that use them.
+	// Note types.
+	const E_WC_ADMIN_NOTE_ERROR         = 'error';     // used for presenting error conditions.
+	const E_WC_ADMIN_NOTE_WARNING       = 'warning';   // used for presenting warning conditions.
+	const E_WC_ADMIN_NOTE_UPDATE        = 'update';    // i.e. used when a new version is available.
+	const E_WC_ADMIN_NOTE_INFORMATIONAL = 'info';      // used for presenting informational messages.
+	const E_WC_ADMIN_NOTE_MARKETING     = 'marketing'; // used for adding marketing messages.
+	const E_WC_ADMIN_NOTE_SURVEY        = 'survey';    // used for adding survey messages.
+
+	// Note status codes.
+	const E_WC_ADMIN_NOTE_PENDING    = 'pending';    // the note is pending - hidden but not actioned.
+	const E_WC_ADMIN_NOTE_UNACTIONED = 'unactioned'; // the note has not yet been actioned by a user.
+	const E_WC_ADMIN_NOTE_ACTIONED   = 'actioned';   // the note has had its action completed by a user.
+	const E_WC_ADMIN_NOTE_SNOOZED    = 'snoozed';    // the note has been snoozed by a user.
+
 	/**
 	 * The name of the non-deprecated class that this facade covers.
 	 *
--- a/src/Notes/DeprecatedNotes.php
+++ b/src/Notes/DeprecatedNotes.php
@@ -17,19 +17,16 @@ use Automattic\WooCommerce\Admin\DeprecatedClassFacade;
  */
 class WC_Admin_Note extends DeprecatedClassFacade {
 	// These constants must be redeclared as to not break plugins that use them.
-	// Note types.
-	const E_WC_ADMIN_NOTE_ERROR         = 'error';     // used for presenting error conditions.
-	const E_WC_ADMIN_NOTE_WARNING       = 'warning';   // used for presenting warning conditions.
-	const E_WC_ADMIN_NOTE_UPDATE        = 'update';    // i.e. used when a new version is available.
-	const E_WC_ADMIN_NOTE_INFORMATIONAL = 'info';      // used for presenting informational messages.
-	const E_WC_ADMIN_NOTE_MARKETING     = 'marketing'; // used for adding marketing messages.
-	const E_WC_ADMIN_NOTE_SURVEY        = 'survey';    // used for adding survey messages.
-
-	// Note status codes.
-	const E_WC_ADMIN_NOTE_PENDING    = 'pending';    // the note is pending - hidden but not actioned.
-	const E_WC_ADMIN_NOTE_UNACTIONED = 'unactioned'; // the note has not yet been actioned by a user.
-	const E_WC_ADMIN_NOTE_ACTIONED   = 'actioned';   // the note has had its action completed by a user.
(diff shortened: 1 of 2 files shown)


octaedro:
Hi @jeffstieler,
Thank you for this nice and really fast fix!
I created an ephemeral site but something is going wrong there.
After following the steps below I'm getting an error:
- Create a new Jurassic Ninja instance.
- Install WC and WC-Admin plugins (from branch `fix/missing-note-constants`)
- Go through the OBW and finish it.
- Connect the site with Jetpack.
- After returning to the `WooCommerce` home, the notes are not available.
- In the Dashboard (/wp-admin/index.php) the following error will be visible:

````
Fatal error: Uncaught Error: Call to a member function get_content_data() on null in /srv/users/user2ef6738d/apps/user2ef6738d/public/wp-content/plugins/mailchimp-for-woocommerce/admin/class-mailchimp-woocommerce-admin.php:289 Stack trace: #0 /srv/users/user2ef6738d/apps/user2ef6738d/public/wp-includes/class-wp-hook.php(287): MailChimp_WooCommerce_Admin->initial_notice('') #1 /srv/users/user2ef6738d/apps/user2ef6738d/public/wp-includes/class-wp-hook.php(311): WP_Hook->apply_filters(NULL, Array) #2 /srv/users/user2ef6738d/apps/user2ef6738d/public/wp-includes/plugin.php(478): WP_Hook->do_action(Array) #3 /srv/users/user2ef6738d/apps/user2ef6738d/public/wp-admin/admin-header.php(292): do_action('admin_notices') #4 /srv/users/user2ef6738d/apps/user2ef6738d/public/wp-admin/index.php(111): require_once('/srv/users/user...') #5 {main} thrown in /srv/users/user2ef6738d/apps/user2ef6738d/public/wp-content/plugins/mailchimp-for-woocommerce/admin/class-mailchimp-woocommerce-admin.php on line 289
There has been a critical error on your website. Please check your site admin email inbox for instructions.
````

**Comment by psealock** [comment]
Should we add a deprecation notice for these?

**Comment by jeffstieler** [comment]
@octaedro that doesn't appear to be caused by this PR - does it happen on the main branch as well?

**Comment by jeffstieler** [comment]
> Should we add a deprecation notice for these?

@psealock I'm not sure we can. I looked for any PHP magic methods that might be for class constant access, but there didn't seem to be any.

**Comment by octaedro** [comment]
Hi @jeffstieler,

> @octaedro that doesn't appear to be caused by this PR - does it happen on the main branch as well?

I thought that it's related to this issue since I'm seeing different errors in `main` and this branch.

Error in `main`:
screenshot-empirical-armadillo jurassic ninja-2020 09 29-20_46_12


Error in this branch:
screenshot-double-rabbit jurassic ninja-2020 09 29-20_48_39

**Comment by becdetat** [comment]
I've just added a commit that solves the remaining issues. The deprecation notices were being displayed, which is less than ideal because it's going to affect everything that uses admin notes, such as the Mailchimp for WooCommerce plugin. So the first fix in my commit is to remove the deprecation notices.

The second fix was for the `Call to a member function xx() on null` that also originated in Mailchimp for WooCommerce. Turned out that I wasn't returning the result of the `call_user_func_array()` call, so a call to `WC_Admin_Notes::get_note()` in MailChimp for WooCommerce was returning null.

Here's the admin note that Mailchimp for WooCommerce generates with these fixes applied:
image

Here are some testing instructions:

1. Start a JN site
1. While that's happening check out this branch and create a zip: `npm run test:zip`
1. Install WooCommerce
1. Skip the OBW
1. Install the zip created off this branch
1. Go to WooCommerce home and complete the OBW. Make sure Mailchimp for WooCommerce gets installed.
1. On the main WP dashboard, there should be no errors
1. On the WC home, there should be no errors and the Mailchimp for WooCommerce admin note should be displayed.

**Comment by becdetat** [comment]
I'm going to make this Needs Review again :-)

**Comment by jeffstieler** [comment]
I think we should keep the deprecation notices - core uses the same mechanism to deprecate code. Stores in production/live mode should not have `WP_DEBUG` on (common WP knowledge) and therefore won't be displaying these errors.

File: src/Notes/DeprecatedNotes.php [hunk 1]
```diff
@@ -29,6 +41,15 @@ class WC_Admin_Note extends DeprecatedClassFacade {
 	 * @var string
 	 */
 	protected static $deprecated_in_version = '1.6.0';
+
+	/**
+	 * Note constructor. Loads note data.
+	 *
+	 * @param mixed $data Note data, object, or ID.
+	 */
+	public function __construct( $data = '' ) {
```
octaedro, line 50: Just in case, when I was testing this branch in an ephemeral site (following the steps I mentioned [here](https://github.com/woocommerce/woocommerce-admin/pull/5215#issuecomment-700930416)) with the addition of this code I'm seeing [this warning](https://d.pr/i/Qk1Zg7):
```
Warning: call_user_func_array() expects parameter 1 to be a valid callback, first array member is not a valid class name or object in /srv/users/user1f7641a7/apps/user1f7641a7/public/wp-content/plugins/woocommerce-admin/src/DeprecatedClassFacade.php on line 55
```
Is that expected?

octaedro (approved):
Thank you @jeffstieler for the quick fix! I only left a question.
Code looks good and is testing well here :shipit:

src/Notes/DeprecatedNotes.php
octaedro, line 50: Just in case, when I tested this branch in an ephemeral site, following the steps I mentioned [here](https://github.com/woocommerce/woocommerce-admin/pull/5215#issuecomment-700930416), with the addition of this code I'm seeing [the warning](https://d.pr/i/Qk1Zg7):
```
Warning: call_user_func_array() expects parameter 1 to be a valid callback, first array member is not a valid class name or object in /srv/users/user1f7641a7/apps/user1f7641a7/public/wp-content/plugins/woocommerce-admin/src/DeprecatedClassFacade.php on line 55
```
Is that expected?

//UPDATE
What I wanted to say is: without this code, I don't see the warning.

jeffstieler: I was able to reproduce this - thank you!

131b136 should resolve it. Can you confirm?

octaedro (approved):
Now it's working perfectly! Thank you for this fix @jeffstieler, it was a pretty tricky issue.

Issue #5225, opened 2020-09-30
Task list: Turn off Stripe test mode when submitting live keys


dechov:
The Stripe extension **defaults to test mode** [[code](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/f85e41d95f9c9366beaa4942cdb1726f88e796e2/includes/admin/stripe-settings.php#L41)] but the manual config in the Stripe onboarding task (currently shown when WooCommerce Services + Jetpack are not connected) accepts **only live keys**.

When submitting live keys, it should switch to live mode. (It would be nice if it could also accept test keys and remain in test mode, but not in the scope of this issue – see https://github.com/woocommerce/woocommerce-admin/issues/4391.)

(We may also want to provide a link to the payment method settings from this screen, so that this and other settings can be manipulated immediately, if the merchant wishes.)

PR #5226, opened 2020-09-30
Task list: Turn off Stripe test mode when submitting live keys

dechov:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/5225

Disable the (default) **test mode** in the Stripe settings upon configuring **live keys** via the onboarding task.

### Accessibility

### Screenshots & detailed test instructions:

Start with the Stripe extension deactivated (or at least with test mode enabled), and navigate to the Stripe payment method on the Payments onboarding task.

With existing settings (populated on my test site but with test mode enabled, consistent with default):

stripe-settings-before

Go through onboarding task:

entering-keys
success

Now test mode is disabled (after new settings page load):

stripe-settings-after

…and the payment method is ready to use with no further configuration.

### DIFF
--- a/client/task-list/tasks/payments/stripe.js
+++ b/client/task-list/tasks/payments/stripe.js
@@ -140,6 +140,7 @@ class Stripe extends Component {
 				...stripeSettings,
 				publishable_key: values.publishable_key,
 				secret_key: values.secret_key,
+				testmode: 'no',
 				enabled: 'yes',
 			},
 		} );


timmyc (approved):
Thanks for the quick fix here @dechov LGTM.

timmyc:
cherry picked to `release\1.6.0-rc.2` in 3a89e2d67956a38a22c533b0ef82364fea868cdb

PR #5229, opened 2020-09-30
Remove dynamic content hash from chunk names.

jeffstieler:
This PR seeks to introduce stability to the filenames used for async code chunks generated by Webpack.

We were previously using `[chunkhash]` in the `chunkFilename` parameter which changes based on the content of the chunk file. This caused a stressful workflow when deploying the plugin to WP.org where the SVN diffs were large and required removing all of the old chunk files so the new ones could be added.

This PR:
- Removes the dynamic `[chunkhash]` from the chunk filename
- Tells the `SplitChunksPlugin` not to generate chunk names - resulting in either numeric IDs, or named files (specified by the `webpackChunkName` magic comment on dynamic `import()`s)
  - Examples: `1.min.js`, `activity-panels-orders.min.js`
- Introduces a new Webpack plugin `AsyncChunkSrcVersionParameterPlugin` that adds the script version parameter to the chunk URLs for cache busting

- Remove the `/dist` directory
- Build using `npm run build`
- Verify the filenames in `dist/chunks` do not contain hashes
- Smoke test the WC Admin app functionality
- Verify that the `?ver=` param is set on all JS and CSS chunk requests (using dev tools)

### Changelog Note:

Dev: Make code chunk filenames more stable.


chunk-src-version-param.js
```diff
@@ -0,0 +1,77 @@
...
+			{ name: pluginName, stage: 1 },
+			( source ) => {
+				if (
+					source.includes( '// mini-css-extract-plugin CSS loading' )
+				) {
+					return source.replace(
+						'linkTag.href = fullhref;',
+						`linkTag.href = fullhref;
+if ( window.wcAdminAssets && window.wcAdminAssets.version ) {
+	linkTag.href += '?ver=' + window.wcAdminAssets.version;
```
joshuatf, line 60: Magic 🧙

joshuatf (approved):
Nicely done, Jeff!  There's a tiny bit of Webpack sorcery here that I'll admit I don't fully understand, but smoke tested the app and things are working well.  LGTM :shipit:

jeffstieler:
> There's a tiny bit of Webpack sorcery here that I'll admit I don't fully understand

@joshuatf I'm happy to walk through these changes any time. Hit me up on Slack!

PR #5231, opened 2020-09-30
Wrap search control selected items in list

### LINKED ISSUE #3647: SearchListControl: selected tags should be in a list
The list of tags shown in the SearchListControl should be wrapped with `<ul>` and `<li>` tags, to make it easier to navigate them with a screen reader.

imatge

The screenshot is from WC Blocks but the relevant code is in this repo: https://github.com/woocommerce/woocommerce-admin/blob/master/packages/components/src/search-list-control/index.js#L196

joshuatf:
Fixes #3647 

Wrap selected items in `SearchListControl` with `ul` and `li` to improve accessibility.

### Accessibility

### Screenshots
Screen Shot 2020-09-30 at 11 55 24 AM

1. Fire up devdocs (make sure `SCRIPT_DEBUG` is `true` in your wp-config).  Looks like this might be broken on `main` but going to an analytics pages and clicking back to devdocs seems to get it working.
1. Select some items in the `SearchListControl`.
1. Check that styling is the same.
1. Check that tests pass.
1. Fire up a screen reader.
1. Check that items can be selected/read in the list.

### DIFF
--- a/packages/components/src/search-list-control/index.js
+++ b/packages/components/src/search-list-control/index.js
@@ -209,14 +209,19 @@ export class SearchListControl extends Component {
 						</Button>
 					) : null }
 				</div>
-				{ selected.map( ( item, i ) => (
-					<Tag
-						key={ i }
-						label={ item.name }
-						id={ item.id }
-						remove={ this.onRemove }
-					/>
-				) ) }
+				{ selectedCount > 0 ? (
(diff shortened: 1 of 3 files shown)


octaedro (approved):
Good job @joshuatf, this is testing well here and code LGTM 🚢

PR #5232, opened 2020-09-30
Removing breadcrumbs from wc-admin header, replacing with simple titl…

### LINKED ISSUE #3963: Remove breadcrumb in the top bar
In all our recent screen designs, we're simply displaying the screen title in the top bar. The breadcrumb has questionable usability benefits and is causing issues like [this](https://github.com/woocommerce/woocommerce/issues/25918) for screens that exist outside of the WooCommerce hierarchy.

The screen title should utilise the `Text - Subtitle Small` [component](https://wordpress.github.io/gutenberg/?path=/story/components-experimental-text--subtitle-small).

joelclimbsthings:
Fixes #3963 

This is an interim step towards the integration of wc-navigation and the addition of a back button within this header. This issue only encompasses the removal of the breadcrumbs in favor of using a simple title with the name of the current page, with a few small revisions to JS and CSS. 

Some small revisions were required in Layout.php as well to account for embed pages displaying the old-style breadcrumbs temporarily while the React view loaded. 

In the issue it was specified to use the <Text> component with variant `subtitle.small` for the heading. Simple enough, although this component obviously wasn't available on the PHP side, so I added a couple styles to account for any stark visual differences tied to the `is-embed` class. Just let me know if there's a more elegant solution for this available. 

### Screenshots

**Before:**
image

**After:**
image

- Navigate to any page with wc-admin header bar
- Ensure that breadcrumbs are _not_ displayed, and instead a simple title is visible with correct styling/etc
- Navigate to embed page, and ensure that no FOUC-like behavior occurs with subtitle, and it's displayed as expected.

### DIFF
--- a/client/header/index.js
+++ b/client/header/index.js
@@ -5,11 +5,9 @@ import { __, sprintf } from '@wordpress/i18n';
 import { useEffect, useRef, useState } from '@wordpress/element';
 import classnames from 'classnames';
 import { decodeEntities } from '@wordpress/html-entities';
-import { Link } from '@woocommerce/components';
 import { useUserPreferences } from '@woocommerce/data';
-import { getNewPath } from '@woocommerce/navigation';
-import { recordEvent } from '@woocommerce/tracks';
-import { getAdminLink, getSetting } from '@woocommerce/wc-admin-settings';
+import { getSetting } from '@woocommerce/wc-admin-settings';
+import { __experimentalText as Text } from '@wordpress/components';
 
 /**
  * Internal dependencies
@@ -18,23 +16,12 @@ import './style.scss';
 import ActivityPanel from './activity-panel';
 import { MobileAppBanner } from '../mobile-banner';
 
-const trackLinkClick = ( event ) => {
-	const target = event.target.closest( 'a' );
-	const href = target.getAttribute( 'href' );
-
-	if ( href ) {
-		recordEvent( 'navbar_breadcrumb_click', {
-			href,
-			text: target.innerText,
-		} );
-	}
-};
-
 export const Header = ( { sections, isEmbedded = false, query } ) => {
 	const headerElement = useRef( null );
 	const rafHandle = useRef( null );
 	const threshold = useRef( null );
 	const siteTitle = getSetting( 'siteTitle', '' );
+	const pageTitle = sections.slice( -1 )[ 0 ];
 	const [ isScrolled, setIsScrolled ] = useState( false );
 	const { updateUserPreferences, ...userData } = useUserPreferences();
 
@@ -106,29 +93,10 @@ export const Header = ( { sections, isEmbedded = false, query } ) => {
 					onInstall={ dismissHandler }
 				/>
 			) }
-			<h1 className="woocommerce-layout__header-breadcrumbs">
-				{ sections.map( ( section, i ) => {
-					const sectionPiece = Array.isArray( section ) ? (
-						<Link
-							href={
-								isEmbedded
-									? getAdminLink( section[ 0 ] )
-									: getNewPath( {}, section[ 0 ], {} )
-							}
-							type={ isEmbedded ? 'wp-admin' : 'wc-admin' }
-							onClick={ trackLinkClick }
-						>
-							{ section[ 1 ] }
-						</Link>
-					) : (
-						section
-					);
-					return (
-						<span key={ i }>
-							{ decodeEntities( sectionPiece ) }
-						</span>
-					);
-				} ) }
+			<h1 className="woocommerce-layout__header-heading">
+				<Text variant="subtitle.small">
+					{ decodeEntities( pageTitle ) }
+				</Text>
 			</h1>
 			{ window.wcAdminFeatures[ 'activity-panels' ] && (
 				<ActivityPanel isEmbedded={ isEmbedded } query={ query } />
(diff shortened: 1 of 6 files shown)


src/Loader.php
```diff
@@ -812,10 +808,8 @@ public static function embed_page_header() {
 		<div id="woocommerce-embedded-root" class="is-embed-loading">
 			<div class="woocommerce-layout">
 				<div class="woocommerce-layout__header is-embed-loading">
-					<h1 class="woocommerce-layout__header-breadcrumbs">
-						<?php foreach ( $sections as $section ) : ?>
-							<?php self::output_breadcrumbs( $section ); ?>
-						<?php endforeach; ?>
+					<h1 class="woocommerce-layout__header-heading">
+						<?php self::output_breadcrumbs( end( $sections ) ); ?>
```
joshuatf, line 812: Nitpick: we can rename this to something like `output_header_title()` to remove references to breadcrumbs.

client/header/index.js
```diff
@@ -106,29 +93,10 @@ export const Header = ( { sections, isEmbedded = false, query } ) => {
...
-						section
-					);
-					return (
-						<span key={ i }>
-							{ decodeEntities( sectionPiece ) }
-						</span>
-					);
-				} ) }
+			<h1 className="woocommerce-layout__header-heading">
+				<Text variant="subtitle.small">
```
joshuatf, line 97: Nice use of `Text`!

client/header/style.scss
```diff
@@ -33,6 +35,15 @@
 		line-height: $header-height;
 		background: $studio-white;
 
+		> p.is-embed {
```
joshuatf, line 38: This one shows up a bit funky to me in the CSS.  Also curious why one would be `13px` and one would be `14px`.  Do these show up differently on embed vs non-embed pages?

joshuatf (commented):
Great work on this one, Joel!  The header title is testing well for me across all pages.  I see a few references to `breadcrumbs` that we can probably clean up before merging this one (layout.js, controller.js, PageController.php), but otherwise things look great.

>Note: I also cherry-picked a recent merge to main to help resolve my linting issues, so you can safely ignore changes to lint-staged.config.js and package.json as they were already merged through another PR.

We can probably give this one a rebase if those changes are merged to drop any package related changes.

>ensure that no FOUC-like behavior occurs with subtitle

❤️

client/header/style.scss
joelclimbsthings: The `font-size:13;` line is actually just a obsolete style at this point, since the `<Text>` component now used for that header assigns its own styles (which match the embed style at 14px). I've been leaning towards not removing styles to ensure I don't remove anything that's being used in some unexpected way, but I think it's safe to do away with these. 

I'm not sure what you mean regarding the funkiness; I'll ping you on the slacks.

client/header/index.js
joelclimbsthings: Well thanks!

src/Loader.php
joelclimbsthings: Good call; I did this elsewhere but forgot this one!

joelclimbsthings:
Great points Josh! Thanks for the review. I've commited my revisions and rebased away that cherry-picked commit. The only uncertainty remaining is regarding the CSS funkiness you described.

client/header/index.js
joshuatf: One small issue with this is that text will by default produce a `p` tag which is semantically incorrect inside an `h1` tag.

There's an `as` property on `Text` we can use to simplify this a bit:

```js
			<Text
				className="woocommerce-layout__header-heading"
				variant="subtitle.small"
				as="h1"
			>
				{ decodeEntities( pageTitle ) }
			</Text>
```

client/header/style.scss
joshuatf: On second thought, to simplify this a bit, I think we can remove the `is-embed` class altogether and keep the font styles on the `.woocommerce-layout__header-heading` selector.  Then we can remove the wrapping `p` tag below.

```
			font-weight: 600;
			font-size: 14px;
```

src/Loader.php
```diff
@@ -774,18 +774,14 @@ private static function get_embed_breadcrumbs() {
...
+	private static function output_heading( $section ) {
 		if ( ! static::user_can_analytics() ) {
 			return;
 		}
 		?>
-		<span>
-		<?php if ( is_array( $section ) ) : ?>
-			<a href="<?php echo esc_url( admin_url( $section[0] ) ); ?>"><?php echo esc_html( $section[1] ); ?></a>
-		<?php else : ?>
+		<p class="is-embed">
```
joshuatf, line 782: This wrapping `p` tag can be removed with the above change.

joshuatf (commented):
Thanks for the change, @joelclimbsthings!  Things look good; just a coupe quick changes I missed the first time around.

>I see a few references to breadcrumbs that we can probably clean up before merging this one (layout.js, controller.js, PageController.php), but otherwise things look great.

Based on the original issue and our Slack discussion, we'll keep this as `breadcrumbs` since breadcrumbs are still used for the document title.

src/Loader.php
joelclimbsthings: Oh, I like this much better. Committing this and related revisions now :+1:

client/header/index.js
joelclimbsthings: That simplifies all the things, I like it!

joshuatf (approved):
Nice clean-up in the CSS and thanks for the changes!  LGTM :shipit:

PR #5248, opened 2020-10-01
Add/wcpay usage tracking request

### LINKED ISSUE #5294: Store Profiler - simplify usage tracking modal
**Description:** we aim to streamline the usage tracking modal be removing one step if the user wants to opt-in to usage tracking

**Acceptance criteria:**
- [ ] Instead of having the users clicking `Yes, count me in!` and then `Continue` we should display two CTA: 
`No thanks` the user moves forward without opting-in for usage tracking
`Yes, count me in!` the user moves forward agreeing to share the usage tracking data

https://d.pr/i/Mv0VvA

**Event tracking:**
This shouldn't change the logic that we already have in place

vbelolapotkov:
Fixes #5294 and Automattic/woocommerce-payments#810 .

#### Proposed changes

* Update UsageModal UI to have two action buttons instead of 'Count me in' checkbox.
* Make UsageModal configurable with custom title, message and buttons text.
* Add customized modal to request site usage tracking after WC Payments KYC flow is completed.

### Accessibility

### Screenshots

image

image

wcpay-tracking-opt-in

Those changes should work with any version of WooCommerce Payments plugin.

There are several flows to test:
1. Set up from task list on WooCommerce home page.
2. Set up from `task=payments` page.
3. Set up from `Payments > Settings` page.

Before starting each flow make sure that:
* WC Pay dev mode is enabled (add `define( 'WCPAY_DEV_MODE', true )` in wp-config).
* WooCommerce Payments plugin is not activated. [Client-example](https://github.com/Automattic/client-example) plugin might be used to break Jetpack connection and break Payments activation.
* Usage tracking is disabled for the site (`WooCommerce > Settings > Advanced > WooCommerce.com`).
* HTTP tunnel (e.g. ngrok) is up and running.

Common test sequence:
1. Go to initial page according to flow.
2. Click `Set up` button or `WooCommerce Payments` task from task list.
3. Approve Jetpack Connection and complete KYC flow if required.
4. After redirect back the modal should appear only if usage tracking is disabled for the site.
5. Clicking `No thanks` should close modal without changing usage tracking setting. Clicking `I agree` should enable site usage tracking.

It's recommended to go through the full sequence at least once from task list.
Once the plugin is set up, you can go to the redirect page by using following URL paths: 

* For tasks page `/wp-admin/admin.php?page=wc-admin&task=payments&wcpay-connection-success=1&method=wcpay`
* For WC Payments settings page `/wp-admin/admin.php?page=wc-settings&tab=checkout&section=woocommerce_payments&wcpay-connection-success=1`

### Changelog Note:

Enhancement: promote site usage tracking after WooCommerce Payments set up.

### DIFF
--- a/client/profile-wizard/steps/usage-modal.js
+++ b/client/profile-wizard/steps/usage-modal.js
@@ -6,12 +6,7 @@ import { Component } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import { withDispatch, withSelect } from '@wordpress/data';
 import interpolateComponents from 'interpolate-components';
-import {
-	Button,
-	CheckboxControl,
-	FormToggle,
-	Modal,
-} from '@wordpress/components';
+import { Button, Modal } from '@wordpress/components';
 import { Link } from '@woocommerce/components';
 import { OPTIONS_STORE_NAME } from '@woocommerce/data';
 
@@ -19,17 +14,9 @@ class UsageModal extends Component {
 	constructor( props ) {
 		super( props );
 		this.state = {
-			allowTracking: props.allowTracking,
 			isLoadingScripts: false,
+			isRequestStarted: false,
 		};
-
-		this.onTrackingChange = this.onTrackingChange.bind( this );
-	}
-
-	onTrackingChange() {
-		this.setState( {
-			allowTracking: ! this.state.allowTracking,
-		} );
 	}
 
 	async componentDidUpdate( prevProps, prevState ) {
@@ -40,7 +27,13 @@ class UsageModal extends Component {
 			onContinue,
 			createNotice,
 		} = this.props;
-		const { isLoadingScripts } = this.state;
+		const { isLoadingScripts, isRequestStarted } = this.state;
+
+		// We can't rely on isRequesting props only because option update might be triggered by other component.
+		if ( ! isRequestStarted ) {
+			return;
+		}
+
 		const isRequestSuccessful =
 			! isRequesting &&
 			! isLoadingScripts &&
@@ -66,8 +59,7 @@ class UsageModal extends Component {
 		}
 	}
 
-	updateTracking() {
-		const { allowTracking } = this.state;
+	updateTracking( { allowTracking } ) {
 		const { updateOptions } = this.props;
 
 		if ( allowTracking && typeof window.wcTracks.enable === 'function' ) {
@@ -80,6 +72,7 @@ class UsageModal extends Component {
 		}
 
 		const trackingValue = allowTracking ? 'yes' : 'no';
+		this.setState( { isRequestStarted: true } );
 		updateOptions( {
 			woocommerce_allow_tracking: trackingValue,
 		} );
@@ -94,64 +87,70 @@ class UsageModal extends Component {
 			return null;
 		}
 
-		const { allowTracking } = this.state;
 		const { isRequesting } = this.props;
-		const trackingMessage = interpolateComponents( {
-			mixedString: __(
-				'Get improved features and faster fixes by sharing non-sensitive data via {{link}}usage tracking{{/link}} ' +
-					'that shows us how WooCommerce is used. No personal data is tracked or stored.',
-				'woocommerce-admin'
-			),
-			components: {
-				link: (
-					<Link
-						href="https://woocommerce.com/usage-tracking"
-						target="_blank"
-						type="external"
-					/>
+		const { isRequestStarted } = this.state;
+		const isBusy = isRequestStarted && isRequesting;
+
+		const title =
+			this.props.title ||
+			__( 'Build a better WooCommerce', 'woocommerce-admin' );
+		const trackingMessage =
+			this.props.message ||
+			interpolateComponents( {
+				mixedString: __(
+					'Get improved features and faster fixes by sharing non-sensitive data via {{link}}usage tracking{{/link}} ' +
+						'that shows us how WooCommerce is used. No personal data is tracked or stored.',
+					'woocommerce-admin'
 				),
-			},
-		} );
+				components: {
+					link: (
+						<Link
+							href="https://woocommerce.com/usage-tracking"
+							target="_blank"
+							type="external"
+						/>
+					),
+				},
+			} );
+
+		const dismissActionText =
+			this.props.dismissActionText ||
+			__( 'No thanks', 'woocommerce-admin' );
+		const acceptActionText =
+			this.props.acceptActionText ||
+			__( 'Yes, count me in!', 'woocommerce-admin' );
 
 		return (
 			<Modal
-				title={ __(
-					'Build a better WooCommerce',
-					'woocommerce-admin'
-				) }
+				title={ title }
+				isDismissible={ this.props.isDismissible }
 				onRequestClose={ () => this.props.onClose() }
-				className="woocommerce-profile-wizard__usage-modal"
+				className="woocommerce-usage-modal"
 			>
-				<div className="woocommerce-profile-wizard__usage-wrapper">
-					<div className="woocommerce-profile-wizard__usage-modal-message">
+				<div className="woocommerce-usage-modal__wrapper">
+					<div className="woocommerce-usage-modal__message">
 						{ trackingMessage }
 					</div>
-					<div className="woocommerce-profile-wizard__tracking">
-						<CheckboxControl
-							className="woocommerce-profile-wizard__tracking-checkbox"
-							checked={ allowTracking }
-							label={ __(
-								'Yes, count me in!',
-								'woocommerce-admin'
-							) }
-							onChange={ this.onTrackingChange }
-						/>
-
-						<FormToggle
-							aria-hidden="true"
-							checked={ allowTracking }
-							onChange={ this.onTrackingChange }
-							onClick={ ( e ) => e.stopPropagation() }
-							tabIndex="-1"
-						/>
+					<div className="woocommerce-usage-modal__actions">
+						<Button
+							isSecondary
+							isBusy={ isBusy }
+							onClick={ () =>
+								this.updateTracking( { allowTracking: false } )
+							}
+						>
+							{ dismissActionText }
+						</Button>
+						<Button
+							isPrimary
+							isBusy={ isBusy }
+							onClick={ () =>
+								this.updateTracking( { allowTracking: true } )
+							}
+						>
+							{ acceptActionText }
+						</Button>
 					</div>
-					<Button
-						isPrimary
-						isBusy={ isRequesting }
-						onClick={ () => this.updateTracking() }
-					>
-						{ __( 'Continue', 'woocommerce-admin' ) }
-					</Button>
 				</div>
 			</Modal>
 		);
--- a/client/profile-wizard/steps/usage-modal.js
+++ b/client/profile-wizard/steps/usage-modal.js
@@ -87,16 +87,10 @@ class UsageModal extends Component {
 			return null;
 		}
 
-		const { isRequesting } = this.props;
-		const { isRequestStarted } = this.state;
-		const isBusy = isRequestStarted && isRequesting;
-
-		const title =
-			this.props.title ||
-			__( 'Build a better WooCommerce', 'woocommerce-admin' );
-		const trackingMessage =
-			this.props.message ||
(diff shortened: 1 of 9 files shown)


vbelolapotkov:
@LevinMedia your feedback on modal design and flow UX is needed :)

joshuatf (commented):
This is testing well and looks good from a dev standpoint.  My only question is if we should consolidate efforts behind the usage modal since there's also a similar one used in the onboarding experience (though design is a bit different).

Screen Shot 2020-10-06 at 4 17 04 PM


I'll leave final approval to @LevinMedia or @jameskoster on the design front.

vbelolapotkov:
Thanks for reviewing @joshuatf !

>  My only question is if we should consolidate efforts behind the usage modal since there's also a similar one used in the onboarding experience (though design is a bit different).

Fair point! Sorry, I didn't raise it from the very beginning. Current designs are offered by @LevinMedia [here](https://github.com/Automattic/woocommerce-payments/pull/855#issuecomment-698462124). So to make questions a bit more clear: 

1. Should we use the same design (buttons) but different messages in both places (store setup wizard and after WC Pay setup)?
2. If yes, which design should be used?

cc @jameskoster

psealock:
cc @pmcpinto for visibility because we're essentially asking twice for those that have not opted in. Is that the approach we'd like to take?

vbelolapotkov:
> cc @pmcpinto for visibility because we're essentially asking twice for those that have not opted in. Is that the approach we'd like to take?

FYI, more context on this decision in the following places:

* https://github.com/Automattic/woocommerce-payments/pull/855#issuecomment-696486249
* paJDYF-PF-p2

pmcpinto:
Thanks for the ping. In terms of design, we already have an [issue](https://github.com/woocommerce/woocommerce-admin/issues/5294) opened on our side to fine-tune the modal

> cc @pmcpinto for visibility because we're essentially asking twice for those that have not opted in. Is that the approach we'd like to take?

I think this is a necessary evil, as WooCommerce and WCPayments are separate products, so each needs to have its own usage tracking authorization

psealock:
There's enough logic similarities between the component added here and our [usage-modal.js](https://github.com/woocommerce/woocommerce-admin/blob/c809f3f335d8510fc0d723a0a960d510159be5dd/client/profile-wizard/steps/usage-modal.js) that I think its worthwhile to consolidate those into one. If the text differs, can we create a prop for that? Or maybe just a title prop and the text stays the same? @LevinMedia do you have a preference on designs? 

The WC Pay version only requires one click, we could go with that one and keep an eye on the metrics. @pmcpinto does that work from a product standpoint?

pmcpinto:
> The WC Pay version only requires one click v. WC Admin's two clicks, so we could go with the version here and keep an eye on the metrics. @pmcpinto does that work from a product standpoint?

Yep, the issue that I created (#5294) for our modal covers that change in the call to actions. On our side this change will probably be shipped in the wcadmin version that will be included in 4.8, right?

> If the text differs, can we create a prop for that? Or maybe just a title prop and the text stays the same? @LevinMedia do you have a preference on designs?

I think it will be better to keep different titles and descriptions in both modals

vbelolapotkov:
> There's enough logic similarities between the component added here and our usage-modal.js that I think its worthwhile to consolidate those into one.

I'm totally fine updating this PR by passing title, message via the props of the shared component to fix #5294 as well. 

To proceed further I'd need a decision from @LevinMedia, @jameskoster and @pmcpinto on modals styling: should it be the same or different? If same, which styling to apply? Modal in this PR has custom styling applied to look like home screen modal requested in [this comment](https://github.com/Automattic/woocommerce-payments/pull/855#issuecomment-698462124).

vbelolapotkov:
@LevinMedia @jameskoster @pmcpinto any update on this? ☝️

vbelolapotkov:
@joshuatf @psealock I've updated PR to consolidate two usage modal and make the initial one more customizable. Design part was discussed in slack (p1603890802044800-slack-CGGCLBN58). So it's ready for another review round.
Please note, I had to move modal styling to layout file to make it loaded on WC Payments settings page too.

vbelolapotkov:
cc @dechov your feedback would be helpful here too.

LevinMedia:
Per the discussion in p6riRB-5F2-p2

Lets update this modal to use the secondary button style for the opt out button: 

Screen Shot 2020-11-05 at 12 39 40 PM

vbelolapotkov:
> Lets update this modal to use the secondary button style for the opt out button:

👍 Addressed in https://github.com/woocommerce/woocommerce-admin/pull/5248/commits/6bcc2302fcab276abff945aca0af3bd191154aa6

client/profile-wizard/steps/usage-modal.js
```diff
@@ -94,64 +87,70 @@ class UsageModal extends Component {
...
-					<Link
-						href="https://woocommerce.com/usage-tracking"
-						target="_blank"
-						type="external"
-					/>
+		const { isRequestStarted } = this.state;
+		const isBusy = isRequestStarted && isRequesting;
+
+		const title =
+			this.props.title ||
```
joshuatf, line 95: Optional, but we could use default props for these to clean up the logic here a bit.

client/profile-wizard/steps/usage-modal.js
```diff
@@ -66,8 +59,7 @@ class UsageModal extends Component {
 		}
 	}
 
-	updateTracking() {
-		const { allowTracking } = this.state;
+	updateTracking( { allowTracking } ) {
```
joshuatf, line 62: I'm wondering if we should add tracks here to see when the tracking is accepted and record the title of the modal so we know where it was turned on.  Maybe we already have this information through another event though.  /cc @timmyc @pmcpinto

client/profile-wizard/steps/usage-modal.js
```diff
@@ -6,30 +6,17 @@ import { Component } from '@wordpress/element';
...
-	Button,
-	CheckboxControl,
-	FormToggle,
-	Modal,
-} from '@wordpress/components';
+import { Button, Modal } from '@wordpress/components';
 import { Link } from '@woocommerce/components';
 import { OPTIONS_STORE_NAME } from '@woocommerce/data';
 
 class UsageModal extends Component {
```
joshuatf, line 13: I think it might be worthwhile to move this to `packages/components` to make it more reusable.

client/profile-wizard/steps/usage-modal.js
```diff
@@ -94,64 +87,70 @@ class UsageModal extends Component {
...
-						type="external"
-					/>
+		const { isRequestStarted } = this.state;
+		const isBusy = isRequestStarted && isRequesting;
+
+		const title =
+			this.props.title ||
+			__( 'Build a better WooCommerce', 'woocommerce-admin' );
+		const trackingMessage =
+			this.props.message ||
```
joshuatf, line 98: We could also destructure some of these props.

joshuatf (commented):
Thanks for updating this @vbelolapotkov!  It's really great to see this component being changed into something reusable.

I left a couple comments around clean-up.  I'm also not able to see the modal in the onboarding wizard anymore.  Can you confirm whether this is working or if I have a problem in my local environment?

vbelolapotkov:
@joshuatf thanks for review! 

> I left a couple comments around clean-up. I'm also not able to see the modal in the onboarding wizard anymore. Can you confirm whether this is working or if I have a problem in my local environment?

Yes I can see it in setup wizard (gif attached). Have you checked if tracking is disabled before launching setup wizard?

woo_setup_usage_modal

client/profile-wizard/steps/usage-modal.js
vbelolapotkov: Hm, I'm concerned about translations. If I understand correctly default props values will be evaluated at the import time and will be used in place of empty props. While current solution evaluates fallbacks at render time. What do you think?

client/profile-wizard/steps/usage-modal.js
vbelolapotkov: We can do this, but don't think it'll help very much. Technically any plugin can enable whole site tracking by setting option to true. On the other hand we can record tracking enabled event in callback of `wcTracks.enable`.

joshuatf: I was thinking more along the lines of so we could determine which modal (for split testing purposes) provided more opt-ins.

But I suppose we could just put a filter on the option server-side and record the referrer if we're interested in this info since multiple usage modal styles/text don't currently appear on the same page.

joshuatf:
Looks like it's my environment.  Even after updating the tracking option, it still comes through as `yes` so there may be some object caching somewhere preventing me from seeing it.  Spun up a new site and the modal worked well.

Just have the above recommendations and maybe removing the excess padding on the modal actions:
Screen Shot 2020-11-16 at 3 38 16 PM

Also unrelated to this PR (and no need to fix here), but ran into some state/render issues with the existing modal.
Screen Shot 2020-11-16 at 3 19 41 PM

pmcpinto:
When will this change be put into production?

client/profile-wizard/steps/usage-modal.js
vbelolapotkov: As I can see all components in the `packages/components` are stateless and this one is pretty much connected to `@woocommerce/data`. We can split the presentation part and behavior and move only presentation part to the components but to be honest I don't see a lot of value in having this modal without logic.

What do you think?

joshuatf: Agreed on its usefulness without being connected.  We do have another component `Plugins` that is both connected and exported, but we can leave this one as is for now until we need it elsewhere.

If we can destructure props and remove the bottom padding on those actions, I think this LGTM.

joshuatf:
@pmcpinto I think this will go into the 1.8.0 release.

client/profile-wizard/steps/usage-modal.js
vbelolapotkov: Added props destructuring in https://github.com/woocommerce/woocommerce-admin/pull/5248/commits/30251836eecaa12ab08e3b608ad160e42d47314a.
Add updated padding (actually replaced with margin-top) in https://github.com/woocommerce/woocommerce-admin/pull/5248/commits/05b32368e29162ea30582651a79488f2b8f28334

vbelolapotkov:
@joshuatf IMO it's ready for final review in terms of code but webpack errors we discussed in slack concern me a bit. I'd like to test the modal appears after KYC flow from both places: Setup task and from Payments > Connect page.

psealock:
Webpack errors are fixed by https://github.com/woocommerce/woocommerce-admin/pull/5722

vbelolapotkov:
Tested locally on top of #5722 and everything works as expected. Adding block label to make sure #5722 merged first.

joshuatf (approved):
Thanks for bearing with us while we get the WCA dependencies fixed up, @vbelolapotkov!  This is testing great and the new usage modal style is a big improvement.  Tested after rebasing with https://github.com/woocommerce/woocommerce-admin/pull/5722 and changes here LGTM 🚢 

I think this existed prior and we can tackle in a follow-up, but looks like we may need to handle things differently on unmount.  I'll create a follow-up issue so we can tackle this.

Screen Shot 2020-11-24 at 9 58 55 AM

joshuatf:
We may also need to update this section to get tests passing for the new modal:

https://github.com/woocommerce/woocommerce-admin/blob/b6cda52323947eaa925c723d298407e6de9db1c8/tests/e2e/specs/activate-and-setup/complete-store-details-section.js#L70-L86

vbelolapotkov:
> We may also need to update this section to get tests passing for the new modal:

Fixed in https://github.com/woocommerce/woocommerce-admin/pull/5248/commits/d9c56c0319029a061dc78bbe367e7aee39444c8e

Fixed state update warning in https://github.com/woocommerce/woocommerce-admin/pull/5248/commits/e864c3ea8866852c08c54217e0ba204d8f1dcbfa

joshuatf (approved):
Testing well!  Thanks for going ahead and getting those fixes added here.  LGTM 🚢

samueljseay:
I'm gonna go ahead and merge this for you @vbelolapotkov as we want to start cutting 1.8.0 release soon. Thank you for your amazing work here!

PR #5271, opened 2020-10-06
Migrate the devdocs examples to Storybook CSF stories

### LINKED ISSUE #5234: Migrate the devdocs to Storybook
Tasks:

* Migrate usage of legacy devdocs page to the storybook in wc-admin.
* Remove all devdocs related code ( there is a feature flag for it ).
* Stretch Goal: setup the storybook to be published online so it can be linked to from the developer portal ( not sure if it is published anywhere yet )

samueljseay:
Partially fixes  #5234 

This is phase 1 of the work to move to Storybook for documenting `@woocommerce/components`. In this phase our proprietary `devdocs` component format is removed in favour of the Storybook [CSF](https://storybook.js.org/docs/react/api/csf) API.

Because this is already a large changeset, I decided to separate out the removal of `devdocs` into a separate PR. I'm happy to hold off merging this until that PR is merged.

Some notes:

1. Most of our components come with a README currently. We should consider distilling some of the markdown in these READMEs into Storybook's MDX format later, but it would be a huge job. I think it should be updated piece by piece.

2. Not all the styles work perfectly. I have removed most of the issues with styles by importing the Wordpress wp-admin styles but a few anomalies remain. Some unfortunately are situations where we have some very broad styles that impact components when used in WooCommerce, these are tech debt that should be improved at some stage. I am thinking about raising a new issue for this.

3. If this PR is merged as is `devdocs` will stop working, so please keep a note of that, I don't want to merge this till the PR for removal of `devdocs` is done.

4. We're definitely not using Storybook to its fullest, a lot of components cannot currently have their props tweaked which is a lovely thing in Storybook. That would also be added to the tech debt to improve these component stories.

### Screenshot

test

1.  Run `npm i`
2. Run `npm run storybook`. The first run will take a while, it downloads the latest Wordpress wp-admin CSS
3. Check out storybook at the address specified in your terminal and have a look through the components.

### Changelog Note:

Dev: Migrate devdocs examples to Storybook stories


bin/import-wp-css-storybook.sh
```diff
@@ -0,0 +1,26 @@
...
+STORYBOOK_WORDPRESS_DIR="$DIR/../storybook/wordpress";
+STORY_BOOK_CSS_PATH="$DIR/../storybook/wordpress/css";
+TMP_DIR="$DIR/../storybook/wordpress/tmp";
+ARCHIVE_CSS_PATH="wordpress/wp-admin/css";
+ARCHIVE_IMG_PATH="wordpress/wp-admin/images";
+
+mkdir -p $STORY_BOOK_CSS_PATH;
+mkdir -p $TMP_DIR;
+
+function download_and_extract_css {
```
samueljseay, line 12: This basic script runs before storybook and downloads the latest Wordpress to extract the needed styles. Unfortunately the CSS references images so we have to extract those too to avoid the build breaking.

packages/components/src/advanced-filters/stories/index.js
```diff
@@ -149,7 +149,7 @@ const advancedFilters = {
 	},
 };
 
-export default () => (
+export const Basic = () => (
```
samueljseay, line 152: Migration is a lot of copy and paste work, I've kept the story migration very simple for now and it usually involves exporting a `Basic` example that renders the original devdocs example. This would definitely need more improvement at some stage.

packages/components/src/calendar/stories/date-picker.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * External dependencies
+ */
+import moment from 'moment';
+import { withState } from '@wordpress/compose';
+import { DatePicker, H, Section } from '@woocommerce/components';
+
+const dateFormat = 'MM/DD/YYYY';
+
+const DatePickerExample = withState( {
```
samueljseay, line 10: These withState calls unfortunately make Storybook incapable of determining the props for tweaking. Stories like this will need work too.

storybook/preview.js
```diff
@@ -8,7 +8,13 @@ import { addDecorator } from '@storybook/react';
 /**
  * Internal dependencies
  */
+import './wordpress/css/wp-admin.min.css';
```
samueljseay, line 11: import the css extracted from Wordpress zip file.

storybook/preview.js
```diff
@@ -8,7 +8,13 @@ import { addDecorator } from '@storybook/react';
...
 import './style.scss';
 
 addDecorator( withA11y );
 addDecorator( withKnobs );
+addDecorator( ( Story ) => (
+	<div className="woocommerce-layout woocommerce-page">
```
samueljseay, line 17: add a couple of wrapper classes that affect component styles

storybook/style.scss
```diff
@@ -1,14 +1,11 @@
-/**
- * External dependencies
- */
-@import '@wordpress/components/build-style/style';
```
samueljseay, line 4: This doesn't seem to be needed, it was only producing styling anomalies.

storybook/style.scss
```diff
@@ -1,14 +1,11 @@
...
-
 /**
  * Internal Dependencies
  */
+@import '../client/stylesheets/index';
```
samueljseay, line 4: there is some style pollution from our main stylesheet where it is the only reason some components display properly. This is tech debt that we need to fix at some point.

webpack.config.js
```diff
@@ -123,6 +123,7 @@ const webpackConfig = {
 			},
 			{
 				test: /\.s?css$/,
+				exclude: /storybook\/wordpress/,
```
samueljseay, line 126: We don't want this plugin working on the wordpress CSS files imported for storybook.

bin/import-wp-css-storybook.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+STORYBOOK_WORDPRESS_DIR="$DIR/../storybook/wordpress";
+STORY_BOOK_CSS_PATH="$DIR/../storybook/wordpress/css";
+TMP_DIR="$DIR/../storybook/wordpress/tmp";
+ARCHIVE_CSS_PATH="wordpress/wp-admin/css";
```
joshuatf, line 6: I think these paths may need to be relative.  The directory that gets added to my `woocommerce-admin` folder looks like its relative from the root directory.

In my environment:
`woocommerce-admin/Sites/my-store/app/public/wp-content/plugins/woocommerce-admin/storybook/wordpress/css/...`

Screen Shot 2020-10-07 at 10 08 26 AM
Screen Shot 2020-10-07 at 10 01 41 AM

joshuatf (commented):
Nice work here, Sam.  Agreed with most of your points and follow-up notes to clean up some code debt we have.

I wasn't able to load storybook due to the the WP assets location, but otherwise I think things look great here.

bin/import-wp-css-storybook.sh
samueljseay: @joshuatf I think we will need to keep this for a while to make the docs look nice. In future my hope is that we would refactor and improve our components to remove the need for this, but I don't expect this to happen quickly.

bin/import-wp-css-storybook.sh
samueljseay: @joshuatf Could you clarify what your setup is?

It is a relative path to the location of the bash script. I'm trying to work out how that could be a problem if you just have `woocommerce-admin` cloned normally? I can run this script from absolutely any subdirectory in the repo right now and it will correctly place the wordpress assets inside the storybook folder. I can even run it from outside the repo and it places it correctly.

Some ideas I can think of:

1. I haven't tested this on Linux, are you on Linux?
2. Maybe the script's mkdir -p failed with some kind of permission issue due to RW permissions? Could you try `sudo` on these lines: 
```
mkdir -p $STORY_BOOK_CSS_PATH;
mkdir -p $TMP_DIR;
```

joshuatf: 1. No, I'm on OSX.
2. Ran `npm run storybook` from the `woocommerce-admin` dir.
3. I did not notice any, but I will check again and see if I can find any.

The folder is created, it's just created in a deeply nested structure of my entire path to `woocommerce-admin` so my gut tells me this is not permissions related, but I'll test around this today and see if I can figure out what's happening.

samueljseay: @joshuatf If you clone it into a fresh directory does it work?

I'd also be happy to try help debug or at least try get to the bottom of why this script is not working in some situations.

Looking at the rsync output where it can't find the directory it looks like in your environment your woocommerce-admin dir is at:

`/Users/joshua/Local Sites/my-store/app/public/wp-content/plugins/woocommerce-admin` Is that correct? If thats correct then the path that `rsync` is looking for is exactly where I'd expect it to place the tmp dir. it just creates the tmp dir inside the storybook folder.

Sorry I can't be more help!

storybook/style.scss
psealock: This may be because this file included Gutenberg component styles: `@import './shared/gutenberg-components.scss';`. This could go away if we stop bundling wp components.

psealock (approved):
I'm not experiencing the issues @joshuatf pointed out, so you may want to resolve those if Josh can provide more info. Otherwise, I had a poke around and this is working well. As you mentioned, there's lots to follow up on so creating some issues would be good.

Nice work @samueljseay and thanks for improving the dev experience.

bin/import-wp-css-storybook.sh
joelclimbsthings: Just ran into the same issue Josh was here; I wonder if it has to do that we're both using Flywheel for our local setup? It's creating another directory at the plugin root:

image

Poking at it now, but if anyone has any suggestions I'm all 👂🏻 !
cc @samueljseay @joshuatf

joelclimbsthings: Ah, may have figured it out. It fails when there is a space in the path since it's missing quotes (I _think_)

samueljseay: @joelclimbsthings yup I fixed this in a recent PR

samueljseay: https://github.com/woocommerce/woocommerce-admin/pull/6934

joelclimbsthings: 😆 . Just noticed that when I tried to merge my own fix 😭 . Thanks @samueljseay

PR #5274, opened 2020-10-06
Fixed action of "Continue Store Setup" button

### LINKED ISSUE #5251: [WC 4.6 RC1] Home Screen: "Continue Store Setup" button shown when the user exits OBW doesn't work
**Describe the bug**

If you exit onboarding without completing it, you are taken to the Home Screen and there is a prompt where you can select to `Continue Store Setup`. My first impression is that if I click on that button indicating that I want to continue Store Setup, I will be taken back to onboarding. But when I click on it, the page reloads and nothing happens. Does this button suppose to take us to Home Screen? If so, it feels a bit redundant that this notice is on the Home Screen:

setup

The URL contains `enable_onboarding=1` which makes me think I should be going back to the Setup Wizard:

**To Reproduce**

1. Create a brand new test site and upload WC 4.6 RC1. 
2. On the very first screen of onboarding select to `Skip Setup Wizard`.
3. You'd be taken to Home Screen where you'd see the notice in question. 
4. Click on `Continue Store Setup` and observe that the page reloads without taking you anywhere. 

**Expected behavior**

I expected to be taken back to the Setup Wizard upon clicking on that `Continue Store Setup` button in the notice on the Home Screen if I exited onboarding earlier. 

**Screenshots**

See above. 

* OS: macOS Catalina 10.15.6
* Browser: Chrome
* Version: Version 85.0.4183.121 (Official Build) (64-bit)

**Additional context**

N/A.

octaedro:
Fixes #5251

This PR fixes the action of the "Continue Store Setup" button.

### Accessibility

### Screenshots
Screen Capture on 2020-10-05 at 16-17-09

- Create a brand new site or set the `wp_options` variable `woocommerce_onboarding_profile` as `a:0:{}` and delete the note named: `wc-admin-onboarding-profiler-reminder`.

The SQL sentence should look like this:
```
DELETE FROM `wp_wc_admin_notes` WHERE `name` = 'wc-admin-onboarding-profiler-reminder';
UPDATE `wp_options` SET `option_value` = 'a:0:{}' WHERE `option_name` = 'woocommerce_onboarding_profile';
```
- Execute the daily cron (`wc_admin_daily`). To do this you can install and use the plugin [WP Crontrol](https://wordpress.org/plugins/wp-crontrol/).
- Go to the `Home` screen.
- The sign: `Welcome to WooCommerce! Set up your store and start selling` should be visible

screenshot-one wordpress test-2020 10 06-11_36_54

- Press `Continue Store Setup`.
- It should redirect you to the OBW.

- Also, verify the `Skip setup` button sets the note as `actioned`

### Changelog Note:
Fix: Action of "Continue Store Setup" button

### DIFF
--- a/src/Notes/OnboardingProfiler.php
+++ b/src/Notes/OnboardingProfiler.php
@@ -60,8 +60,8 @@ class Onboarding_Profiler {
 		$note->add_action(
 			'skip-profiler',
 			__( 'Skip Setup', 'woocommerce-admin' ),
-			wc_admin_url( '&reset_profiler=0' ),
-			'actioned',
+			false,
+			Note::E_WC_ADMIN_NOTE_ACTIONED,
 			false
 		);
 		return $note;
--- a/src/Notes/OnboardingProfiler.php
+++ b/src/Notes/OnboardingProfiler.php
@@ -53,7 +53,7 @@ class Onboarding_Profiler {
 		$note->add_action(
 			'continue-profiler',
 			__( 'Continue Store Setup', 'woocommerce-admin' ),
-			wc_admin_url( '&enable_onboarding=1' ),
+			wc_admin_url( '&path=/setup-wizard' ),
 			NOTE::E_WC_ADMIN_NOTE_UNACTIONED,
 			true
 		);


joshuatf (commented):
Hey @octaedro 👋  I have a PR I'm about to merge in that does the same.  https://github.com/woocommerce/woocommerce-admin/pull/5200

Do you think we should remove `&reset_profiler=0` from the "skip" option?  This won't actually skip the setup if we do it, it will just action the note, but the setup will still occupy the homescreen.

octaedro:
Hi @joshuatf,
Great, I think that merging your PR will be enough to solve this problem. I'm going to add a comment to the issue #5251 mentioning this.

>Do you think we should remove &reset_profiler=0 from the "skip" option? This won't actually skip the setup if we do it, it will just action the note, but the setup will still occupy the homescreen.

You're right. I thought that it wasn't possible to see this note without skipping the OBW but I realized it is by modifying the URL. So it's a good idea to keep the `&reset_profiler=0`.

I'll close this PR.

PR #5275, opened 2020-10-06
Remove devdocs from wc-admin

samueljseay:
Phase 2 of #5234 this removes all references to `devdocs` from wc-admin. 

As part of the plan to migrate to Storybook we're planning to publish Storybook stories to Github pages so the devdocs section will have no real value to remain in wc-admin. You'll still be able to spin up Storybook locally to view the docs via `npm run storybook`.

**This doesn't remove `docs` which is a docsify site with lots of extensive documentation about components but a lot of other stuff too. I think the plan would be to eventually take some of the textual documentation about components from here and put it into the Storybook stories.**

Note that this is a PR into the Storybook migration PR to separate the changesets for easier reviewing.

1. Enable WP_DEBUG and SCRIPT_DEBUG in your environment
2. Confirm that the devdocs menu item is no longer shown and that there are no errors

### Changelog Note:

Dev: remove devdocs section from wp-admin


psealock (approved):
Testing well and code looks good to me 👍

PR #5276, opened 2020-10-07
Fix a bug in tax automation logic and track the event in analytics

### LINKED ISSUE #5257: Tracks not being triggered when Taxes and Shipping are completed by default
**Describe the bug**
According to #4594 and #4595 the event `wcadmin_tasklist_task_completed` should be triggered when those steps are completed by default, including a new event prop for each. However, I'm not seeing the event be triggered when I reach the home screen with those steps automatically completed.

**To Reproduce**

1. Install WC 4.6 RC1
2. Complete the Store Profiler, including the connection to Jetpack
3. Go to the home screen and see the Shipping and Tax steps automatically completed
4. See the Network tab on Chrome and confirm that `wcadmin_tasklist_task_completed` wasn't triggered

**Expected behavior**
See `wcadmin_tasklist_task_completed` being triggered for the Shipping and Tax tasks when they're automatically completed.

samueljseay:
Fixes #5257 and another undiscovered bug with the tax automation flow.

The original PR to add the tax automation flow missed out the tracking functionality by accident and also left in a bug in the logic where it did not check that the user was in a country that supports tax automation 🤦

This PR fixes both those issues.

1. You'll need a site with WCS installed and TOS accepted, Jetpack installed and connected and you'll need an incomplete onboarding wizard
2. Complete the OBW steps ensuring that you select a tax automation supported country. (The US is fine)
3. You should see the same behaviour as previous where the tax step is marked completed on the task list after finishing the OBW
4. You should also see that an event was tracked.
5. To test the **bug** you could do steps 1-3 but at step 2 choose a non automated tax supported country such as New Zealand. On arriving at the task list the setup tax task should not have been completed for you.

### Changelog Note:

Fix: Fix a bug in tax automation workflow that did not consider the user's country.
Add: Track the tax automation with analytics via the `tasklist_task_completed` event.


becdetat:
Everything looks fine however I couldn't get my JN sites to enable automatic tax calculations. This wasn't an issue with this PR as I couldn't get it to work on my related PR either, which is annoying because it was working this morning. Do you have steps to set up JN so automated taxes work?

samueljseay:
@becdetat I'm now seeing a 500 error during OBW on Atomic sites so maybe similar. Will take a look in the morning.

becdetat:
image
whoops wrong branch :-$

samueljseay:
@becdetat haha I did that to @adrianduffell 's branch earlier today. too much branch swapping

src/Features/OnboardingAutomateTaxes.php
```diff
@@ -67,4 +68,11 @@ public static function on_onboarding_profile_completed() {
 	private static function automated_tax_is_supported() {
 		return in_array( WC()->countries->get_base_country(), \OnboardingTasks::get_automated_tax_supported_countries(), true );
```
joshuatf, line 69: I think we'll need to remove the `\` before `OnboardingTasks` to use the correct namespace.  I see the following error:

Screen Shot 2020-10-07 at 12 05 39 PM

src/Features/OnboardingAutomateTaxes.php
```diff
@@ -55,9 +55,10 @@ public static function on_onboarding_profile_completed() {
 			$wcs_tos_accepted = \WC_Connect_Options::get_option( 'tos_accepted' );
 		}
 
-		if ( $jetpack_connected && $wcs_version && $wcs_tos_accepted ) {
+		if ( $jetpack_connected && $wcs_version && $wcs_tos_accepted && self::automated_tax_is_supported() ) {
```
joshuatf, line 58: Not directly related to the issue in this PR, but I imagine not many sites will be Jetpack connected prior to setup so this will rarely get used right?  Is this mostly for Atomic sites?

joshuatf (approved):
Thanks for the fix-up here, Sam.  Pre-approving pending one change and also left a question unrelated to this PR that's not a blocker to get this merged.

src/Features/OnboardingAutomateTaxes.php
samueljseay: @joshuatf 

> Not directly related to the issue in this PR, but I imagine not many sites will be Jetpack connected prior to setup so this will rarely get used right? Is this mostly for Atomic sites?

I don't know if it was designed for Atomic sites but your intuition sounds right imo that this will rarely get used in any other scenario.

src/Features/OnboardingAutomateTaxes.php
samueljseay: Whoops. Thanks!

timmyc:
Cherry picked to `release/1.6.0` in https://github.com/woocommerce/woocommerce-admin/commit/2c6b94785951127064506820cfd2707feb374cae

PR #5280, opened 2020-10-07
[WIP] Add a github action to monitor bundle size.

samueljseay:
This enables a Github action to monitor JS/CSS bundle size of PRs. It uses [compressed action size](https://github.com/preactjs/compressed-size-action) which Gutenberg uses as well.

The intent of this is not to block PR's if the bundle size ever increases because that would make 
this too inflexible. In future if we can have a way to block in a more lenient way though that might be
preferred.

The idea of this is to inform the reviewer and remind the PR author during the code review process and ensure 
that our bundles don't drastically increase in size due to some change.

This PR should show feedback from the action as a PR check. 

### Changelog Note:

Dev: check JS/CSS bundle size on PRs


samueljseay:
`npm ci` fails on main so its impossible to test this currently, I'll try some test PRs merging this branch into a dummy branch.

PR #5282, opened 2020-10-07
[WIP] Testing bundle size check action - do not merge.

samueljseay:

PR #5284, opened 2020-10-07
Add a bundle size check to PRs

samueljseay:
This bundle size check will not block PRs but it will help inform PR reviews and track the size of our bundles. As part of this I needed to fix up our lerna dependencies a bit because they didn't work with `npm ci`. I used this fix here: https://github.com/lerna/lerna/issues/1865#issuecomment-453277927 to correct the dependencies.

For that reason testers of this PR should do a quick smoke test of the app to make sure everything works as before.

Unfortunately though this means that the bundle check as part of **this** PR will fail because the dependencies are not correctly set up on `main` (which it will use to compare the bundle sizes). Test instructions will allow you to test this as a PR if you wish.

1. General smoke test of the app. Check for errors/issues
2. (optional) checkout a new branch from this branch, add some JS changes (a few lines of valid code should be enough). Create a PR of that branch into this one, to see the bundle check in action.

### Changelog Note:

Dev: Add a Github action to check bundle size on PRs.


joshuatf (commented):
This is awesome!  Will be a great metric to add to our PRs.

I'm having some trouble seeing the bundle size; looks like there are some dependency issues causing npm to fail.

Screen Shot 2020-10-07 at 1 03 26 PM

samueljseay:
@joshuatf I *think* I know why yours is failing.

did you checkout your branch *from* this branch?

ie:

`git checkout dev/bundle-size-check`
`git checkout -b whatever`
add your changes
`git push origin whatever`

The reason you *must* do that for it to work is that you need the package-lock changes from *this* branch to make it work properly because this action uses `npm ci` to install dependencies.

e.g. this PR which was a branch from the branch passes: https://github.com/woocommerce/woocommerce-admin/pull/5282

joshuatf:
@samueljseay Yes, I checked out from this branch `dev/bundle-size-check`.  I also see that this branch is failing- is that expected?

samueljseay:
@joshuatf Yes this branch will always fail against `main` as I mentioned in the description. 

Thats because both the comparison branch and the branch being merged to must have the fixups to package-lock that I introduced in this branch. The reason this is is because the github action does a `npm ci` on both the PR branch and the target branch being merged to, to perform a comparison. If `npm ci` fails on either then it will fail the action.

I might have made a mistake during the process of testing branches though so I'll do some more testing here before passing it back to you :)

samueljseay:
@joshuatf I had another thought here to build confidence. We can separate out the package-lock changes as a separate PR to test and merge and then follow up with this PR and see it pass. I'll do that.

PR #5289, opened 2020-10-07
DO NOT MERGE: Try/test bundle size

joshuatf:
Testing bundle size added in https://github.com/woocommerce/woocommerce-admin/pull/5284

### Accessibility

- Ex: Open page `url`
- Click XYZ…


samueljseay:
@joshuatf thats so weird, it seems that the action run on your PR here checked out the wrong target branch. I'll investigate

samueljseay:
@joshuatf Just pinging you on this in case you miss it, can you confirm if you branched from the original branch? the package-lock must be correct. I commented about it more in the original PR.

samueljseay:
Closing this one. Will do some more testing of my own.

PR #5292, opened 2020-10-08
Navigation: Prep feature flags and option

psealock:
Add a feature flag and option to enable/disable the new Navigation experience. Much of this is based off the enable/disable setup for the Homescreen, which was removed in https://github.com/woocommerce/woocommerce-admin/pull/5108.

* Add feature flag.
* Add the feature `src/Features/Navigation.php`.
* Hydrate options on Embedded pages as well. cc @samueljseay as this is related to your investigations as part of https://github.com/woocommerce/woocommerce-admin/pull/5161.

1. `composer install` and a fresh `npm start` to get the features array built correctly.
1. Visit any page, embedded or not.
2. Check the feature `window.wcAdminFeatures.navigation` reflects your option.
3. Check the option is correctly pre-loaded `wcSettings.preloadOptions.woocommerce_navigation_enabled`.
4. Change the option value to "yes" in your DB and repeat steps 2 and 3.

### Changelog Note:

none: unreleased changes

### DIFF
--- a/client/layout/index.js
+++ b/client/layout/index.js
@@ -222,15 +222,19 @@ export const PageLayout = compose(
 		: identity
 )( _PageLayout );
 
-export class EmbedLayout extends Component {
-	render() {
-		return (
-			<Layout
-				page={ {
-					breadcrumbs: getSetting( 'embedBreadcrumbs', [] ),
-				} }
-				isEmbedded
-			/>
-		);
-	}
-}
+const _EmbedLayout = () => (
+	<Layout
+		page={ {
+			breadcrumbs: getSetting( 'embedBreadcrumbs', [] ),
+		} }
+		isEmbedded
+	/>
+);
+
+export const EmbedLayout = compose(
+	window.wcSettings.preloadOptions
+		? withOptionsHydration( {
+				...window.wcSettings.preloadOptions,
+		  } )
+		: identity
+)( _EmbedLayout );
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -165,6 +165,10 @@ class Loader {
 	 * @return bool Returns true if the feature is enabled.
 	 */
 	public static function is_feature_enabled( $feature ) {
+		if ( 'navigation' === $feature && 'yes' !== get_option( 'woocommerce_navigation_enabled', 'no' ) ) {
+			return false;
+		}
+
 		$features = self::get_features();
 		return in_array( $feature, $features, true );
 	}
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -165,10 +165,6 @@ class Loader {
 	 * @return bool Returns true if the feature is enabled.
 	 */
 	public static function is_feature_enabled( $feature ) {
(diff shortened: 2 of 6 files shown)


src/Loader.php
```diff
@@ -165,6 +165,10 @@ public static function user_can_analytics() {
 	 * @return bool Returns true if the feature is enabled.
 	 */
 	public static function is_feature_enabled( $feature ) {
+		if ( 'navigation' === $feature && 'yes' !== get_option( 'woocommerce_navigation_enabled', 'no' ) ) {
```
joshuatf, line 168: I think this logic can be moved into `Navigation.php` and we can use the `woocommerce_admin_features` filter.  This allows us to continue using `Loader::is_feature_enabled()` to check for features.

See https://github.com/woocommerce/woocommerce-admin/blob/eb6ac6e9708bc4139a4cf11c1f92c703813ba222/src/FeaturePlugin.php#L204 for prior art.

client/layout/index.js
```diff
@@ -222,15 +222,19 @@ export const PageLayout = compose(
...
+const _EmbedLayout = () => (
+	<Layout
+		page={ {
+			breadcrumbs: getSetting( 'embedBreadcrumbs', [] ),
+		} }
+		isEmbedded
+	/>
+);
+
+export const EmbedLayout = compose(
```
joshuatf, line 234: We can probably use `window.wcAdminFeatures` to check if the feature is enabled here.

joshuatf (commented):
Thanks for getting the ball rolling on the migration, Paul!   Left a comment about using a different filter so we can take advantage of some of the existing methods.

It would be great if we could create a folder for our new feature, but I'm pretty sure this would require adjustments to how the features are autoloaded and out of scope for this PR.

client/layout/index.js
psealock: I think it was a mistake to not preload options for embedded pages. See https://github.com/woocommerce/woocommerce-admin/pull/5161 for a ramification of that decision. I think it could happen easily again because its easy to overlook an option being pre loaded on the app, but not in the embedded pages. So, I think pre-loading should happen regardless of Navigation or not.

psealock:
Thanks for the review @joshuatf ! 

Is this what you were suggesting in moving the logic to a filter -> https://github.com/woocommerce/woocommerce-admin/pull/5292/commits/949a964513bff2a488feccfff27353b268cbfc0e ?

From a debugging standpoint, do you think it obfuscates the logic by putting it into a filter? I'm all good with this change, just curious your thoughts.

joshuatf (approved):
Thanks for the changes, Paul.  LGTM!

>From a debugging standpoint, do you think it obfuscates the logic by putting it into a filter? I'm all good with this change, just curious your thoughts.

Ya, it does a bit.  I don't think this is the best pattern, but I'd like to move away from having multiple sources of truth as to whether or not a feature is enabled.  We'll probably need to follow-up to improve this at some point.

PR #5293, opened 2020-10-08
Version bump: WP components and base styles

psealock:
The `Navigation` component is only available in wp.components version `11.0.0`, which requires the latest base styles as well. This bumps each and removes an import to a file that no longer exists (See https://github.com/WordPress/gutenberg/commit/06d17403c68dfdce73a494b6eeff1f9a49c3e244 for why). So this bump is required to accept the Navigation into wc-admin.

1. `npm install && npm start`.
2. Make sure the build succeeds with no errors.
3. Spot check the app for visual regressions.

### Changelog Note:

none: fix for unreleased changes


samueljseay (approved):
tests well. 🚀

PR #5308, opened 2020-10-08
Navigation: Move PHP code to WC Admin

psealock:
Depends on https://github.com/woocommerce/woocommerce-admin/pull/5292

This is pretty much a wholesale copy/paste from WC Navigation of `Menu`, `CoreMenu`, and `Screen`. I've also added a conditional on `Loader::is_feature_enabled( 'navigation' )` in `Navigation.php` to ensure Navigation doesn't work unless the feature flag is on AND the option is set to "yes".

Since the JS isn't available, we can only test classes and root elements.

1. Run `composer install`
1. Set the `woocommerce_navigation_enabled` option to "yes".
2. Visit any WooCommerce page.
3. See the root element present for Navigation to hook onto: `document.getElementById('woocommerce-embedded-navigation')`.
4. See the proper class applied to the body: `document.body.classList.contains('has-woocommerce-navigation');`

### Changelog Note:

none: unreleased changes

### DIFF
--- a/src/Features/Navigation/CoreMenu.php
+++ b/src/Features/Navigation/CoreMenu.php
@@ -5,10 +5,10 @@
  * @package Woocommerce Admin
  */
 
-namespace Automattic\WooCommerce\Navigation;
+namespace Automattic\WooCommerce\Admin\Features\Navigation;
 
-use Automattic\WooCommerce\Navigation\Menu;
-use Automattic\WooCommerce\Navigation\Screen;
+use Automattic\WooCommerce\Admin\Features\Navigation\Menu;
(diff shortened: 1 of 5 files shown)


src/Features/Navigation.php
```diff
@@ -1,32 +0,0 @@
-<?php
```
joshuatf, line 1: Thanks for organizing this under a folder!  💯

joshuatf (approved):
Pre-approving pending potential changes in https://github.com/woocommerce/woocommerce-admin/pull/5292 to feature flag logic.  Nice work 🎉

PR #5309, opened 2020-10-09
Navigation: Add client code

psealock:
Depends on https://github.com/woocommerce/woocommerce-admin/pull/5308

Move client side code over from WC Navigation.

### Screenshots

Screen Shot 2020-10-09 at 1 21 40 PM

1. `composer install` if you haven't already from the previous PR.
2. Use the navigation and see that its at the point of development where WC Navigation repo was.

### Known Issues

https://github.com/woocommerce/navigation/issues

### New Issues

There's something going on with the wp-admin menu: WooCommerce still has "Settings" and Analytics and Marketing should not exist. I'd like to tackle this in a follow up.

Screen Shot 2020-10-09 at 1 25 23 PM

### Changelog Note:

none: changes to unreleased feature


joshuatf (approved):
I did note the issues you mentioned around menu items not being migrated.  Otherwise this is working really well.

I think we should go ahead and get this merged so we can unblock and tackle those issues in a follow-up.

Issue #5285, opened 2020-10-07
Normalize Text File Line Endings


adrianduffell:
**Describe the bug**

The `readme.txt` file has Windows style CRLF line endings, whereas other files like `license.txt` use LF. [Unlike the core repo](https://github.com/woocommerce/woocommerce/blob/aac4db94b2ff060d9b889132e7ff50779f52f819/.gitattributes#L1-L5), rules are not present in the `.gitattributes` file to normalize text files with a consistent line ending style.

This can result in issues seen in #5118 where commits are made with unexpected line endings.

**To Reproduce**

**In macOS**

1. Compare the line endings for the `readme.txt` file and `license.txt` files:
```
$ file readme.txt
readme.txt: UTF-8 Unicode text, with CRLF line terminators
$ file license.txt
license.txt: UTF-8 Unicode text
```

**Expected behavior**
The `readme.txt` file should not have CRLF line endings.

**Additional context**
The `.editorconfig` file [specifies](https://github.com/woocommerce/woocommerce-admin/blob/96e6d3053ab740e792ee1614e43b98190f0be4f1/.editorconfig#L11) file specifies all text files should have LF line endings.

adrianduffell:
The [update-version.php](https://github.com/woocommerce/woocommerce-admin/blob/main/bin/update-version.php) script puts Unix style new lines in the` readme.txt` file when it bumps version numbers.

This is ok if `readme.txt` is changed to use LF line endings. Otherwise, the script should be updated to put CRLF endings there.

adrianduffell:
Re-opening, since `.gitattributes` rules are not yet in place normalize line endings.

adrianduffell:
The `readme.txt` was converted to LF line endings in PR #5517. To prevent the possibility of it reverting back to CRLF, I think it would be worth looking at  [improvements](https://docs.github.com/en/free-pro-team@latest/github/using-git/configuring-git-to-handle-line-endings) to the `.gitattributes` configuration in a follow-up PR.

mattsherman:
That sounds like a good idea.

PR #5310, opened 2020-10-09
Fix Inconsistent Newlines in 1.6.0 Release Readme

adrianduffell:
A Unix style line ending (LF) is present in the `readme.txt` file. This PR changes the line ending to Windows style (CRLF) to make it consistent with the rest of the file.

This is similar to issue #5118. I think a permanent fix should get worked on in issue https://github.com/woocommerce/woocommerce-admin/issues/5285

Note: This does not result in a visible change in the GitHub diff.

This should be merged with the `release/1.6.0` branch.

- Confirm the `readme.txt` file has `CRLF` line endings only.

#### For example, In macOS:
- Open Terminal
- Run the following command:
```
$ cat -e readme.txt | grep Stable
```
- In the command output, ensure the line ending has the `^M$` CRLF marker: 
```
Stable tag: 1.6.0^M$
```
- Run the following command:
```
$ file readme.txt
```
- In the command output, ensure only CRLF line terminators are listed:
```
readme.txt: UTF-8 Unicode text, with CRLF line terminators
```


timmyc:
Thanks for the fix here @adrianduffell - yeah must be something in [update-version.php](https://github.com/woocommerce/woocommerce-admin/blob/main/bin/update-version.php) that is causing this to happen

PR #5312, opened 2020-10-09
Shipping task smart default: add exceptions

### LINKED ISSUE #5298: Task List: shipping task smart default: add exceptions
According to the original issue (#4595), the shipping task should only be marked as completed if the user has JP and WCS installed and connected. Currently the task is being marked as completed even if the user doesn't have JP connected.

In addition, we have a partnership with Shipstation in Australia, Canada and United Kingdom where we display the prompt to install the extension in the Shipping step. It's necessary to add an exception to not mark it as completed in those countries.

becdetat:
Fixes #5298

This adds some exceptions to the completed criteria for the shipping task:

- Jetpack must be connected and WCS must be active
- Country must not be AU, NZ, or UK

- Start with a fresh site
- Go through the OBW. Select a US state.
- When on the home screen, the shipping task should _not_ be completed
- Connect Jetpack
- Return to the home screen, the shipping task should now be completed
- Go through the OBW again, and select a location in AU, NZ, or UK
- Return to the home screen, the shipping task should _not_ be completed again

### Changelog Note:

Fix: Shipping task smart default: add exceptions

### DIFF
--- a/client/task-list/tasks.js
+++ b/client/task-list/tasks.js
@@ -94,6 +94,11 @@ export function getAllTasks( {
 		purchaseAndInstallText = sprintf( purchaseAndInstallFormat, itemName );
 	}
 
+	const shippingCompleted =
+		shippingZonesCount > 0 &&
+		isJetpackConnected &&
+		activePlugins.includes( 'woocommerce-services' );
+
 	const tasks = [
 		{
 			key: 'store_details',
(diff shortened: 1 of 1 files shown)


psealock (approved):
Code and logic looks correct to me according to the issue. Testing works well too, thanks @becdetat

timmyc:
cherry picked to `release/1.6.0` in https://github.com/woocommerce/woocommerce-admin/commit/dab38e21dcfbc54a6a303027ca643515642dac13

PR #5316, opened 2020-10-09
Revise Excluded Countries from Auto Completion of Shipping Task

timmyc:
Follow up to #5312 - the incorrect list of countries was used on the original PR, so this quick change revises the list to include Canada instead of New Zealand.


jeffstieler (approved):
👍

PR #5321, opened 2020-10-11
Fix wrong filter arguments & duplicate schema properties

sultann:
Fixes #

There were a couple of issues in Download Controller

- `woocommerce_report_downloads_prepare_export_item` filter was passing an unknown property instead of `$export_columns` 
- Same class was containing double `product_id` property in schema


becdetat:
Thanks for this @sultann, it works well. I've created another branch with your commits cherry-picked so that it will run through our CI. I also changed the fix to `prepare_item_for_export` to use `$export_item` to match the other report controllers.

This will be merged in my PR #5441. Thanks for contributing!

Issue #5259, opened 2020-10-02
Customer Effort Score


pmcpinto:
**Description:** Evaluating the health of the core store management features will help us understand if each feature is creating the expected value and prioritize which ones should be improved. It will allow us to compare the scores before and after making improvements in a specific feature.

**Acceptance criteria:**
- [ ] Trigger an in-product message to collect feedback from the user after he finishes a specific task (design tbd)
- [ ] Trigger it once per store in the same feature, avoiding overwhelming them with the same survey every time they complete the same task
- [ ] Trigger the survey in the following tasks:
Add a product (`wcadmin_product_add_publish`)
Update product (`wcadmin_product_update`)
Edit order (any of the edit order events)
Change setting (`wcadmin_settings_change`)
Filter analytics (any action on the analytics reports)
* ping me to get the events list for edit order and filter analytics

- [ ] Only target users with usage tracking enabled
- [ ] Collect the feedback using tracks
- [ ] Build the prompt as a separate package so it can be reused by other teams (see discussion here: pbIJXs-m0-p2#comment-890)
- [ ] The comments text field appears and is required only if Very difficult or Somewhat difficult is selected, and the comment will be recorded alongside the effort score.
- [ ] (nice to have for v1) abstract the data collection so it can use different collection targets

**Event tracking:**

Event: `wcadmin_ces_view`
Prop: `action` (action that triggered the survey – eg: wcadmin_product_add_publish)

Event: `wcadmin_ces_feedback`
Props:
`action` (action that triggered the survey – eg: wcadmin_product_add_publish)
`score`: [1-5]
`store_age`: [months]
`product_number`: [number of products]
`order_number`: [number of orders]

Event: `wcadmin_ces_snackbar_view`
Prop: `action` (action that triggered the survey – eg: wcadmin_product_add_publish)

Event: `wcadmin_ces_snackbar_click`
Prop: `action` (provide_feedback or dismiss)

**Additional links:** 
[Project Thread](pbIJXs-m0-p2)
[Final design](pbIJXs-px-p2)

pmcpinto:
@becdetat @adrianduffell I added this to the AC:
> - [ ] The comments text field appears and is required only if Very difficult or Somewhat difficult is selected, and the comment will be recorded alongside the effort score

I'm still evaluating with Elizabeth if it should be required or optional. Feel free to give your feedback :)

elizaan36:
> I'm still evaluating with Elizabeth if it should be required or optional.
[...]

PR #5324, opened 2020-10-11
Basic customer effort score survey functionality

becdetat:
This is a spike of the customer effort score feature described in https://github.com/woocommerce/woocommerce-admin/issues/5259.

At the moment the modal looks like this:
image

It is behind a feature toggle, enabled for development only. This consists of a component in `@woocommerce/customer-effort-score`, which is the modal itself which accepts a callback for submitting the survey. This is wrapped by `client/customer-effort-score-tracks/index.js`, which uses the Tracks library for tracking. This wrapping lets the survey be independent of the data collection method, allowing reuse between teams.

### Steps to test:

1. Add a new product. The modal should be displayed on the edit product page (the page you are taken to after publishing a new product).
2. Click the "Click me" button. Tracks data should be outputted, eg:

```
wcadmin_product_add_publish_effort_score,
{
    score: 0,
    product_count: 146
}
```
3. Test this with editing a product as well

Note that the event name and properties are going to change in subsequent PRs.


client/customer-effort-score-tracks/index.js
```diff
@@ -0,0 +1,66 @@
+/**
+ * External dependencies
+ */
+import { useState } from '@wordpress/element';
+import PropTypes from 'prop-types';
+import { recordEvent } from '@woocommerce/tracks';
+import CustomerEffortScore from '@woocommerce/customer-effort-score';
+
+/**
```
timmyc, line 9: Thanks for all the doc blocks and proptype declarations here. I also like this implementation to layer in tracks.

client/index.js
```diff
@@ -81,3 +82,40 @@ const navigationRoot = document.getElementById(
...
+	if ( ! window.wcAdminFeatures[ 'customer-effort-score' ] ) {
+		return;
+	}
+
+	const root = appRoot || embeddedRoot;
+
+	function getJson() {
+		try {
+			return (
+				window.localStorage.getItem( 'customerEffortScoreTracks' ) ||
```
timmyc, line 97: I wasn't sure if there might have been an established pattern around localStorage use in WP core, but could only find [this persistence plugin](https://github.com/WordPress/gutenberg/tree/4e9b6b51caf315f7c71ee345a94793a35d83ac39/packages/data/src/plugins/persistence) for wp.data.

While I don't think we can just drop that in here and use, they do have an interesting fallback object store. @samueljseay any thoughts on localStorage use in wc-admin?

packages/customer-effort-score/src/index.js
```diff
@@ -0,0 +1,71 @@
...
+	function close() {
+		setScore( 3 ); // TODO let this happen in the UI
+
+		toggleVisible();
+		trackCallback( score );
+	}
+
+	return (
+		<p className="customer-effort-score_modal">
+			{ label } <button onClick={ close }>Click me</button>
```
timmyc, line 44: So is the idea the final design here will have multiple buttons that all call `close()`? Will the design be tightly coupled to what @elizaan36 has created or are we going to make it a more config-based render?

packages/customer-effort-score/src/index.js
```diff
@@ -0,0 +1,71 @@
...
+	);
+}
+
+CustomerEffortScore.propTypes = {
+	/**
+	 * The function to call when the modal is actioned.
+	 */
+	trackCallback: PropTypes.func.isRequired,
+	/**
+	 * Whether or not the tracks is visible. True is used for when
```
timmyc, line 55: Should this be "Whether or not the dialog is visible"?

src/Features/CustomerEffortScoreTracks.php
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	/**
+	 * Hook into the Publish button for products, to trigger the customer
+	 * effort score widget after save.
+	 *
+	 * @param WP_Post $post The post, not used.
+	 */
+	public function add_meta_boxes_product( $post ) {
+		wc_enqueue_js(
```
timmyc, line 48: Interesting - would it be possible to use the product lifecycle hooks here at all? This made me curious to how we do something similar after a product is added as part of the setup checklist, and the approach there is somewhat similar but no localStorage used: https://github.com/woocommerce/woocommerce-admin/blob/main/src/Features/OnboardingTasks.php#L233

timmyc (commented):
Really awesome start here! Had a few questions, but also would love to get some other eyes on this too.

packages/customer-effort-score/src/index.js
becdetat: I think it should be @elizaan36 's design, not abstracting things just for the sake of it otherwise we're trading configurability for the ability to just use the CES package.

packages/customer-effort-score/src/index.js
becdetat: Yes ;-)

src/Features/CustomerEffortScoreTracks.php
```diff
@@ -0,0 +1,66 @@
...
+			"
+			(function() {
+				if ( $( 'h1.wp-heading-inline' ).text().trim() !== '" . __( 'Add new product', 'woocommerce-admin' ) . "' ) {
+					return;
+				}
+
+				$( '#publish' ).click( function() {
+					var json = JSON.stringify( [ {
+						trackName: 'product_add_publish_effort_score',
+						label: '" . __( 'Provide effort for adding a new product', 'woocommerce-admin' ) . "',
```
becdetat, line 58: Note to self: This should be `'How easy was it to add a product?'`

client/index.js
```diff
@@ -81,3 +82,40 @@ const navigationRoot = document.getElementById(
...
+	const json = getJson();
+	const queue = JSON.parse( json );
+
+	queue.forEach( ( item ) => {
+		render(
+			<CustomerEffortScoreTracks
+				initiallyVisible={ true }
+				trackName={ item.trackName }
+				label={ item.label }
+				trackProps={ {} }
```
becdetat, line 114: Note to self: include `item.props` here

src/Features/CustomerEffortScoreTracks.php
becdetat: Also include a `props` value and give it the product count (`product_count`).

src/Features/CustomerEffortScoreTracks.php
becdetat: (Kinda) done in df92a4634e8ee6dc0a8b26fb3f711c6b6fb251b3

src/Features/CustomerEffortScoreTracks.php
becdetat: 1b96f017de4f88854091e7eb1792b7dbad2e7adc

becdetat:
@timmyc I took your suggestion and refactored this to use the product lifecycle. This is ready for another review :-)

jeffstieler:
In my testing, the modal had some z-index conflicts with the tinymce post editor:

Visual mode:

Screen Shot 2020-10-27 at 8 44 39 AM

Text mode:

Screen Shot 2020-10-27 at 8 44 52 AM

jeffstieler (changes requested):
I'm not sure if this is within the scope of this PR, but we'll need a way to target specific screens for these modals. I was able to get the CSE component to show on the Home screen:

Capture 2020-10-27 at 10 03 52

packages/customer-effort-score/src/index.js
```diff
@@ -0,0 +1,70 @@
...
+ * @param {Function} props.toggleVisible Callback to toggle the visible prop.
+ * @param {string}   props.label         The label displayed in the modal.
+ */
+function CustomerEffortScore( {
+	trackCallback,
+	visible,
+	toggleVisible,
+	label,
+} ) {
+	const [ score, setScore ] = useState( 0 );
```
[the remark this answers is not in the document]
becdetat: Interacting with the score doesn't close the modal, there is a submit (and cancel) button on the actual designs (which @adrianduffell is building out in a seperate PR)

src/Loader.php
```diff
@@ -310,6 +315,14 @@ public static function register_scripts() {
 
 		wp_set_script_translations( 'wc-currency', 'woocommerce-admin' );
 
+		wp_register_script(
+			'wc-customer-effort-score',
+			self::get_url( 'customer-effort-score/index', 'js' ),
+			array(),
```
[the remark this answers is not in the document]
becdetat: Probably yes, this seems to happen work without it so far, but I'll put it in.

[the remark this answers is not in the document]
becdetat: Actually no! The `customer-effort-score` component is independent of the measurement method - `CustomerEffortScoreTracks` provides a function to `CustomerEffortScore` that uses wc-tracks to actually report the metric. This is so that other teams can use CustomerEffortScore without needing to depend on Tracks.

src/Loader.php
```diff
@@ -94,6 +95,10 @@ public function __construct() {
 
 		// Combine JSON translation files (from chunks) when language packs are updated.
 		add_action( 'upgrader_process_complete', array( __CLASS__, 'combine_translation_chunk_files' ), 10, 2 );
+
+		if ( self::is_feature_enabled( 'customer-effort-score' ) ) {
```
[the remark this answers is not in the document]
becdetat: Yes it would, and that's awesome.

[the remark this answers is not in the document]
becdetat: 55b4fc9d1d1e6a642e57b51b6b0090243daa8382

becdetat:
@jeffstieler 

> In my testing, the modal had some z-index conflicts with the tinymce post editor

@adrianduffell is doing the visual part of this, I just wanted to get _something_ on screen, so this should be addressed in another PR.

> I'm not sure if this is within the scope of this PR, but we'll need a way to target specific screens for these modals. I was able to get the CSE component to show on the Home screen

Wow, nice testing. I'll address this in another PR. Thanks!

src/Loader.php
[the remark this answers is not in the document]
jeffstieler: Ah, right - whoops!

becdetat:
> > I'm not sure if this is within the scope of this PR, but we'll need a way to target specific screens for these modals. I was able to get the CSE component to show on the Home screen

> Wow, nice testing. I'll address this in another PR. Thanks!

https://github.com/woocommerce/woocommerce-admin/issues/5482 is the issue for only showing the survey in the appropriate screens.

client/customer-effort-score-tracks/customer-effort-score-tracks-container.js
```diff
@@ -0,0 +1,94 @@
...
+	 */
+	queue: PropTypes.arrayOf( PropTypes.object ),
+	/**
+	 * If the queue option is being resolved.
+	 */
+	resolving: PropTypes.bool,
+	/**
+	 * Set up clearing the queue on the next page load.
+	 */
+	clearQueueFromOption: PropTypes.func,
```
jeffstieler, line 65: The `FromOption` portion of this function name spells out some specifics of the implementation. I don't think it lends any clarity to the purpose of the method and could be potentially confusing should that implementation end up changing later on.

src/Features/CustomerEffortScoreTracks.php
```diff
@@ -0,0 +1,146 @@
...
+		$product_count = intval( $products->total );
+
+		return $product_count;
+	}
+
+	/**
+	 * Enqueue the CES survey trigger for a new product.
+	 */
+	private function enqueue_ces_survey_for_new_product() {
+		$queue = get_option( self::CES_TRACKS_QUEUE_OPTION_NAME, array() );
```
jeffstieler, line 93: Once we're past the "spike" version of this, some helper methods to manage the queue would probably be helpful.

jeffstieler (approved):
This is looking good (taking into account the acknowledged follow up work) - let's 🚢  it. (I confirmed that a production build doesn't include the CES)

I left a naming nitpick comment that doesn't need to be addressed.

client/customer-effort-score-tracks/customer-effort-score-tracks-container.js
becdetat: sounds good, I wasn't entirely happy with this naming

becdetat: fixed in 3db8d9a

PR #5352, opened 2020-10-13
Display CES modal once per action per store

### LINKED ISSUE #5338: Customer Effort Modal: Only ever display the modal once per event per store
Part of #5259

To avoid overloading the user with prompts, only ever display the customer effort modal once per event per store. If the modal is dismissed or actioned, don't show it again for the event that triggered it. This should be stored as an option.

becdetat:
Fixes #5338 

The CES modal should only be displayed once per action per store. This is done using an option value (`woocommerce_ces_shown_for_actions`) which contains an array of actions that have had the CES modal shown. The JS that hooks in to the "Publish" button on the "Add new product" page is not added if the option contains the current action. The CES tracks component is also not shown if the option contains the current action.

- Add and publish a new product. The modal should be displayed.
- Refresh the page. The modal should not be displayed.
- Add and publish a new product. The modal should not be displayed.
- Delete the `woocommerce_ces_shown_for_actions` option then add a new product. The modal should be displayed again.

### Changelog Note:

Dev: Display CES modal once per action per store


src/Features/CustomerEffortScoreTracks.php
```diff
@@ -90,15 +112,28 @@ private function get_product_count() {
 	 * Enqueue the CES survey trigger for a new product.
 	 */
 	private function enqueue_ces_survey_for_new_product() {
+		// Only add the JS to trigger the CES modal if
+		// this modal hasn't been dismissed or actioned yet for this action.
+		$shown_for_features = get_option( self::SHOWN_FOR_ACTIONS_OPTION_NAME, array() );
```
jeffstieler, line 117: It looks like we should establish a pattern similar to the inbox notes for checking if a CES prompt has been shown before, to keep this DRYer.

src/Features/CustomerEffortScoreTracks.php
```diff
@@ -110,15 +145,28 @@ private function enqueue_ces_survey_for_new_product() {
 	 * Enqueue the CES survey trigger for an existing product.
 	 */
 	private function enqueue_ces_survey_for_edited_product() {
+		// Only add the JS to trigger the CES modal if
+		// this modal hasn't been dismissed or actioned yet for this action.
+		$shown_for_features = get_option( self::SHOWN_FOR_ACTIONS_OPTION_NAME, array() );
```
jeffstieler, line 150: Using an option for this would prevent showing the same CES prompt to multiple users. Is that desired?

jeffstieler (commented):
I'm thinking we'll want to use WP user meta instead of a (global) option for this.

becdetat:
I wasn't aware there was user metadata. Thanks for this :-D.

src/Features/CustomerEffortScoreTracks.php
becdetat: My understanding is that it should be shown for multiple users but once it is dismissed or completed it shouldn't be shown again, so I think using an option is ok.

becdetat: Yes it is, it is - https://woostartp2.wordpress.com/2020/10/12/customer-effort-score-design-i1/#comment-975

src/Features/CustomerEffortScoreTracks.php
becdetat: I agree, I tend to stick to the rule of three for code reuse, but it makes sense to refactor this now. Are you happy for me to do this in a follow-up PR?

becdetat: I've done this in fc8fffa

mattsherman (approved):
Tested and works as expected. Code looks reasonable. Nice job!

mattsherman:
Oops, missed the file conflict... I'm cool with you merging once that is resolved. If you feel you want another set of 👀 on it before merging though, just ping me and I'll take another look!

PR #5353, opened 2020-10-13
Task List: Match Stripe settings to mode of provided keys

dechov:
Addresses https://github.com/woocommerce/woocommerce-admin/pull/5201#issuecomment-704931521, primarily by configuring _test_ keys & mode when those are provided via the Stripe task.

- Go to Task List » Payments » Stripe
- Verify that mismatched mode (e.g. `pk_live_****` with `sk_test_****`) results in a validation error
- Submit test keys
- Go to WooCommerce » Settings » Payments » Stripe, and verify that test mode and test keys are set

(Can test live mode case as well, and verify that test mode is not set and live keys are set.)

### DIFF
--- a/client/task-list/tasks/payments/methods.js
+++ b/client/task-list/tasks/payments/methods.js
@@ -170,6 +170,15 @@ export function getPaymentMethods( {
 		} );
 	}
 
+	// Whether publishable and secret keys are filled for given mode.
+	const isStripeConfigured =
+		options.woocommerce_stripe_settings &&
+		( options.woocommerce_stripe_settings.testmode === 'no'
+			? options.woocommerce_stripe_settings.publishable_key &&
+			  options.woocommerce_stripe_settings.secret_key
+			: options.woocommerce_stripe_settings.test_publishable_key &&
+			  options.woocommerce_stripe_settings.test_secret_key );
+
 	methods.push(
 		{
 			key: 'stripe',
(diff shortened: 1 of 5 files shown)


dechov:
There are checks in this file and methods.js which should also be updated to support the test key case, especially the latter – will add that today.

psealock (approved):
Code looks good and this is testing well. The correct errors show up when they are supposed to. :shipit:

dechov:
@psealock Thank you! I've just pushed one more small change, to show that the method is configured after submitting test keys (rather than "Set up" again). Would you have a chance to re-review?

client/task-list/tasks/payments/methods.js
```diff
@@ -192,10 +192,16 @@ export function getPaymentMethods( {
 				! hasCbdIndustry,
 			plugins: [ 'woocommerce-gateway-stripe' ],
 			container: <Stripe />,
+			// Whether publishable and secret keys are filled for given mode.
 			isConfigured:
 				options.woocommerce_stripe_settings &&
-				options.woocommerce_stripe_settings.publishable_key &&
-				options.woocommerce_stripe_settings.secret_key,
+				( ( options.woocommerce_stripe_settings.testmode === 'no' &&
```
becdetat, line 198: The logic here is getting very convoluted and looks like it would be difficult to test and maintain. Can you split this out into it's elements above?

becdetat:
Hi @dechov 

>  I've just pushed one more small change, to show that the method is configured after submitting test keys (rather than "Set up" again

This is testing well, however I just have a comment about the change to the logic. Perhaps a unit test around the logic would also be a good idea? It might be overkill though, if you can simplify things by extracting out the elements.

dechov:
@becdetat Thanks – extracted this logic into a separate variable above, simplified it just a bit, and rebased (and also added a commit to fix a small typo which may not warrant its own PR). I think it's more readable now – especially since the two cases have the same indent level – but let me know if you think I should extract out another variable or two for clarity.

octaedro (approved):
Thank you @dechov for this fix. This is testing well and the code looks good 🚢

octaedro:
Cherry picked to release/1.7.0 in afbdf13

PR #5358, opened 2020-10-13
Wait for server-side processing when clicking on a note

jrodger:
A small experiment to explore inbox notifications and waiting for server-side processing to complete before navigating away.

This is 1 of 3 PRs experimenting with approaches (paJDYF-RI-p2#comment-3762 for context).

* Return a promise from triggerNoteAction
* Wait for the promise to resolve before navigating to the action button's URL

* There's a new class in the PR, so run `composer dump-autoload`
* Add this url to DataSourcePoller::DATA_SOURCES: `https://gist.githubusercontent.com/jrodger/d2737468bc55408b4d0ecee35f5c8b44/raw/7d8a4b4dc1112d5ea2b4548fc96cf47d8caf7f90/note-install.json`
* Run the `wc_admin_daily` cron task
* Click the "Sign-up" action button
* The button shows the busy animation, WooCommerce Payments is installed and activated, and finally we successfully load the WooCommerce Payments on-boarding screen

<h3>Pros</h3>

* Nearly everything is handled from the inbox notification
* Successfully redirects the user to the target page after running some server-side setup

<h3>Cons</h3>

* Obviously the implementation is very naive, but we’d need a way to only stall on actions that actually have something to process. If we applied this globally we’d slow down “normal” button presses (for example, links out to WooCommerce.com pages and things like that).
* Having the server-side note code feels a little strange for remote inbox notifications. I don’t know if we might want to encapsulate this logic elsewhere because of this.
* The server-side note class does a lot of the same work as the current WooCommerce Payments note. Would there be some value in combining them?
* We have to define the link we redirect to ahead of time, so no using links that need a nonce or dynamically linking to Jetpack onboarding flows etc.

### DIFF
--- a/client/inbox-panel/action.js
+++ b/client/inbox-panel/action.js
@@ -36,6 +36,10 @@ class InboxNoteAction extends Component {
 		const href = event.target.href || '';
 		let inAction = true;
 
+		// Conditionally suppress navigation for certain notes, obviously this isn't conditional yet so will probably
+		// break things.
+		event.preventDefault();
+
 		if ( href.length && ! href.startsWith( adminUrl ) ) {
 			event.preventDefault();
 			inAction = false; // link buttons shouldn't be "busy".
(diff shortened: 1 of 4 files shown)


jrodger:

PR #5359, opened 2020-10-13
Adding 24px spacing between any sibling nav groups (#5329)

### LINKED ISSUE #5329: Navigation: Spacing should be added between menus
In the designs, there's a gap between the menus.  Previously, the secondary menu was pushed to the bottom of the screen.  However, in current designs, it looks like there is a `24px` gap.

Screen Shot 2020-09-30 at 9 59 34 AM

joelclimbsthings:
Fixes #5329

Adding margin between the two menu subgroups, per the design.

### Screenshots

image

- Click on WooCommerce to view new navigation
- Ensure the margin is present as per the screenshot
- For extra credit, use inspector to confirm 24 pixel spacing

_Note:_ This PR was migrated from the navigation repo [here](https://github.com/woocommerce/navigation/pull/133)

### DIFF
--- a/client/navigation/components/container/style.scss
+++ b/client/navigation/components/container/style.scss
@@ -14,8 +14,12 @@
 		height: 100%;
 	}
 
-	.components-navigation__group:nth-of-type(2) {
-		margin-top: auto;
+	.components-navigation__group {
+		margin-bottom: 0;
+	}
+
+	.components-navigation__group + .components-navigation__group {
+		margin-top: 24px;
 	}
 
 	.woocommerce-navigation__back-to-dashboard {
--- a/client/navigation/components/container/style.scss
+++ b/client/navigation/components/container/style.scss
@@ -14,12 +14,12 @@
 		height: 100%;
 	}
 
-	.components-navigation__group {
-		margin-bottom: 0;
-	}
-
 	.components-navigation__group + .components-navigation__group {
 		margin-top: 24px;
+	}
+
+	.components-navigation__item:last-child {
+		margin-bottom: 0;
 	}
 
 	.woocommerce-navigation__back-to-dashboard {
--- a/client/navigation/components/container/style.scss
+++ b/client/navigation/components/container/style.scss
@@ -18,7 +18,7 @@
 		margin-top: 24px;
 	}
 
-	.components-navigation__item:last-child {
+	.components-navigation__item {
 		margin-bottom: 0;
 	}
 


client/navigation/components/container/style.scss
```diff
@@ -14,8 +14,12 @@
 		height: 100%;
 	}
 
-	.components-navigation__group:nth-of-type(2) {
-		margin-top: auto;
+	.components-navigation__group {
+		margin-bottom: 0;
```
joshuatf, line 18: The margin I'm seeing is on the item `.components-navigation__item` (as opposed to the group) seen here:

Screen Shot 2020-10-13 at 3 53 23 PM

joshuatf (commented):
Thanks for re-opening this PR back in WCA, Joel!

One comment about the item margin, but I'm also thinking long-term that these styles should be added upstream in Gutenberg to keep styles consistent across various consumers of the navigation.

client/navigation/components/container/style.scss
joelclimbsthings: Thanks Josh, just checked and it looks like that margin is also being collapsed, so it doesn't extend the spacing there at all.

client/navigation/components/container/style.scss
```diff
@@ -14,8 +14,12 @@
 		height: 100%;
 	}
 
-	.components-navigation__group:nth-of-type(2) {
-		margin-top: auto;
+	.components-navigation__group + .components-navigation__group {
+		margin-top: 24px;
+	}
+
+	.components-navigation__item:last-child {
```
joshuatf, line 21: I think in the designs this is for all items, so we can update this to:

```suggestion
	.components-navigation__item {
```

joshuatf (approved):
Thanks for the changes, Joel!  Left one quick suggestion, but pre-approving pending that change.

client/navigation/components/container/style.scss
joelclimbsthings: Thanks @joshuatf . Taking a closer look at that design doc you sent me, I noticed that each of those items are supposed to be `32px` in height as well, but they're currently `36px`. Just feeling out in regards to future issues, is that something that it would make sense to address in the scope of this ticket as well?

joshuatf: Good question.  It's really up to you; if it's an easy fix to make, it's okay to toss a fix into these PRs, but if you find it's cumbersome or too far out of scope, a follow-up is almost always okay.

Some of these issues honestly would be better handled upstream though, so I'm hesitant to introduce too many temporary fixes if we're going to remove them after we get fixes pushed to Gutenberg.

PR #5362, opened 2020-10-13
Wait for wcpay installed and update action button text to proceed with setup.

vbelolapotkov:
Similar to #5358. This is 1 of 3 PRs experimenting with approaches (paJDYF-RI-p2#comment-3762 for context).

* Use existing note handler class to install the plugin.
* Update note once the plugin install with next step url/text.
* Direct user to the WC Pay connect page for further set up.

* Add this url to DataSourcePoller::DATA_SOURCES: `https://gist.githubusercontent.com/jrodger/9b757e6c30163ea7c1f62140b8ef0534/raw/df6d7538ca799422627e3da3e6297f8cd105eb8e/note-two-step-install.json`
* Run the wc_admin_daily cron task
* Click the "Install" action button
* The button shows the busy animation, then updates to say "Get Started"
* Clicking "Get Started" navigates to the WCPay setup page

<h3>Pros</h3>

 * Doesn't do anything non-standard with links. Just fires a server-side action and then follows a normal link to get the user to the setup screen.

<h3>Cons</h3>

* As with #5358, we need to tweak the JavaScript so it doesn't stay in the "loading" state. I think this implementation is a little less problematic than #5358, because we're not trying to block a link click and fire it later. However, it still might need a little testing.
* Having the server-side note code feels a little strange for remote inbox notifications. I don’t know if we might want to encapsulate this logic elsewhere because of this. We've taken a slightly different approach here and combined the logic into the existing WooCommerce Payment note class. I'm not sure if that's preferable to creating a new Note class (as in #5361).

### DIFF
--- a/src/Notes/WooCommercePayments.php
+++ b/src/Notes/WooCommercePayments.php
@@ -38,6 +38,7 @@ class WooCommerce_Payments {
 	 */
 	public function __construct() {
 		add_action( 'woocommerce_note_action_install-now', array( $this, 'install' ) );
+		add_action( 'woocommerce_note_action_get-woocommerce-payments', array( $this, 'handle_promo' ) );
 		add_action( 'wc-admin-woocommerce-payments_add_note', array( $this, 'add_note' ) );
 	}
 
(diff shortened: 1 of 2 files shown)


src/Notes/WooCommercePayments.php
```diff
@@ -134,16 +135,52 @@ protected static function is_installed() {
...
+		foreach ( $actions as $action ) {
+			if ( 'get-woocommerce-payments' === $action->name ) {
+				$action->label = 'Set Up';
+				$params        = [
+					'page' => 'wc-admin',
+					'path' => '/payments/connect',
+				];
+				$action->query = \admin_url( \add_query_arg( $params, 'admin.php' ) );
+			}
+			$new_actions[] = clone $action;
```
vbelolapotkov, line 180: This one is weird but important step, because without it actions are not updated in DB on save.

jrodger:
Just noting that I updated a chunk of Vasily's original PR description. I don't want to put words into your mouth Vasily!

jrodger:

PR #5363, opened 2020-10-13
Fix up menu item migrations in WC navigation

### LINKED ISSUE #5350: Navigation: Registration bug in wp-admin sidebar
There's something going on with the wp-admin menu: WooCommerce still has "Settings" and Analytics and Marketing should not exist. This bug was introduced when code was migrated over from WC Navigation.

Screen Shot 2020-10-09 at 1 25 23 PM

joshuatf:
Fixes #5350 

Fixes up the menu item's not being properly migrated after moving WC Nav into WCA.

Note that this does not resolve the submenu shown when hovering WooCommerce.  That one is a bit trickier since we need to hide the menu without loading our stylesheet (or load our styles on all pages).

### Screenshots

Screen Shot 2020-10-13 at 7 43 33 PM

1. Make sure `woocommerce_navigation_enabled` is set to `yes` in your options table.
1. Check a non-WC page.
1. Make sure the only WC related page left is the "WooCommerce" parent menu item.

### DIFF
--- a/src/Features/Analytics.php
+++ b/src/Features/Analytics.php
@@ -113,15 +113,14 @@ class Analytics {
 	 */
 	public function register_pages() {
 		$overview_page = array(
-			'id'       => 'woocommerce-analytics',
-			'title'    => __( 'Analytics', 'woocommerce-admin' ),
-			'path'     => '/analytics/overview',
-			'icon'     => 'dashicons-chart-bar',
-			'position' => 56, // After WooCommerce & Product menu items.
+			'id'           => 'woocommerce-analytics',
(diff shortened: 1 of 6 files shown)


psealock (commented):
This is looking great @joshuatf and the code makes sense. The only thing is the Nav failing to load on WC Settings pages.

Also, a rebase will fix that CI error on e2e test.

psealock (approved):
Oh ok, I just saw that you made an issue about the Settings pages in https://github.com/woocommerce/woocommerce-admin/issues/5382. With that, these changes are good to go 🚢

PR #5281, opened 2020-10-07
Fix inconsistent line endings in readme.txt

adrianduffell:
Fixes #5118

This updates the `Stable tag` line in `readme.txt` to use a `CRLF` style line ending. This makes it consistent with the rest of the file and fixes “mixed line ending” type error messages when opening.

Note: This does not result in a visible change in the GitHub diff.

### Detailed test instructions:

- Confirm the `readme.txt` file has `CRLF` line endings only.

#### For example, In macOS:
- Open Terminal
- Run the following command:
```
$ cat -e readme.txt | grep Stable
```
- In the command output, ensure the line ending has the `^M$` CRLF marker: 
```
Stable tag: 1.6.0-dev^M$
```
- Run the following command:
```
$ file readme.txt
```
- In the command output, ensure only CRLF line terminators are listed:
```
readme.txt: UTF-8 Unicode text, with CRLF line terminators
```

### Changelog Note:

Fix: Inconsistent line endings in readme.txt.

PR #5364, opened 2020-10-14
Change readme.txt Line Endings to Unix Style

adrianduffell:
This PR converts the line ending style for `readme.txt` from Windows (CRLF) to Unix (LF). LF is the same format that WooCommerce core uses.

It fixes the mixed line endings recently created with the `1.6.1` version bump.

The previous CRLF style is problematic because our tooling expects text files to be in LF style.

- `.editorconfig` [specifies](https://github.com/woocommerce/woocommerce-admin/blob/96e6d3053ab740e792ee1614e43b98190f0be4f1/.editorconfig#L11) LF.
- The `update-version.php` script [uses](https://github.com/woocommerce/woocommerce-admin/blob/db5ce423e0feacaa8ad7c165006c0de09dbfcc83/bin/update-version.php#L16-L29) LF (`\n`) when bumping version numbers.

The utility `dos2unix` was used to convert the file.

Previous fixes for mixed line endings have kept the CRLF endings on this file. However I think it would be better to switch the file over to LF to provide compatibility with our tooling.

A separate PR will be created for the `main` branch.

### Detailed testing instructions

- In GitHub, [append `?w=1` to the URL of this PR](https://github.com/woocommerce/woocommerce-admin/pull/5364/files?w=1) to see the diff without whitespace changes. Confirm no actual content changes have occurred. 

Screen Shot 2020-10-14 at 8 09 16 am

- Confirm the `readme.txt` file is in LF format.

#### For example, in macOS:

The following command should NOT display a message about CRLF endings.

```
$ file readme.txt
readme.txt: UTF-8 Unicode text
```


adrianduffell:
Previously https://github.com/woocommerce/woocommerce-admin/pull/5281 https://github.com/woocommerce/woocommerce-admin/pull/5310.

becdetat:
LGTM

PR #5367, opened 2020-10-14
Attempt to check for a theme step specific element when completing e2e test

samueljseay:
The theme selection page e2e test is intermittently failing and it
is not clear why. Its possible that between checking for a button
with class `primary` and then clicking it, that the business step 
is still rendered on screen and then when going to click the button
there is nothing rendered. I can't prove this, but this piece of code
does close up that edge case.

3 runs of this build so far have passed and this is still an improvement
to the test even if it doesn't solve the problem completely.


psealock (approved):
With tests passing and this simply waiting, I think its good to merge. Thanks for looking into it @samueljseay

Issue #5334, opened 2020-10-08
New navigation - getting feedback


pmcpinto:
**User story:**
As a merchant
I want to have a way to provide feedback about the new navigation
So I can easily contact the WooCommerce team when I spot any bugs or potential improvements

**Description:** considering that this is a significant change in the experience we're thinking about an easy way to allow users to provide timely feedback about the new navigation (bugs, improvements), in order to shorten the feedback cycle.

**Acceptance criteria:**
- [ ] TBD

**Event tracking:**
TBD

pmcpinto:
Last week I spoke with @elizaan36 about this and for the initial release we can take advantage of the inbox panel and trigger a message with a link to Crowdsignal to get feedback. We can then evaluate if it makes sense to add something more prominent before releasing it for WC Core users. Elizabeth, are you ok with this approach?

@joshuatf when do you need the inbox message content and survey link?

joshuatf:
@pmcpinto The sooner the better, but as long as we have a few days buffer before we release to implement and test it should be fine.

elizaan36:
> for the initial release we can take advantage of the inbox panel and trigger a message with a link to Crowdsignal to get feedback. We can then evaluate if it makes sense to add something more prominent before releasing it for WC Core users. Elizabeth, are you ok with this approach?

Yep the Inbox message seems like a good approach for ecom plan users, but I have a suspicion we'll need a bigger announcement as we widen the audience.

pmcpinto:
> @pmcpinto The sooner the better, but as long as we have a few days buffer before we release to implement and test it should be fine.

👍 We'll have the inbox content and survey link during next week

elizaan36:
Here's copy for the Inbox message:

Headline: **You have early access to the new WooCommerce navigation - share your feedback**
Body: We're introducing the new navigation to a select group of store owners in this first phase and we'd like to hear your thoughts.
CTA: [[Share feedback](https://automattic.survey.fm/new-navigation)]

Here's the link to the Crowdsignal survey for the CTA: https://automattic.survey.fm/new-navigation

pmcpinto:
@elizaan36 when do you think that the survey should be triggered? 5 days after they start using the new navigation?

elizaan36:
I'm updating this thread based on a convo with @pmcpinto just now - let me know how this copy looks to you.

Two Inbox notifications for feedback on the new nav, linking to the same survey: 
1. Immediately upon rollout of the new navigation
[...]

PR #5381, opened 2020-10-14
Add features section and settings

### LINKED ISSUE #5333: New navigation - disable/enable option
**User story:**
As a merchant
I want an easy way to disable/enable the new navigation
So I can use the navigation that's more convenient for me

**Description:** during the initial phase we aim to allow users to disable and enable the new navigation as they'll need time to get used to the new experience and this can be used as a signal to track the intent of adopting the navigation.

**Acceptance criteria:**
- [ ] TBD (we're evaluating if we should adopt the same setting used for the home screen)

**Event tracking:**
TBD

joshuatf:
Fixes #5333

Adds the features section and setting for the navigation.

Note that a refresh or page navigation is needed to get this working since the option update fires after features are already loaded.  Do we think this is worth fixing (maybe with a forced redirect) or fine as is?

### Screenshots

Screen Shot 2020-10-14 at 2 33 30 PM

1. Visit WooCommerce -> Settings -> Advanced -> Features ( `wp-admin/admin.php?page=wc-settings&tab=advanced&section=features` )
`. 
1. Toggle the feature on/off and save.
1. Check that the feature is turned on or off as necessary. (note the above issue with refresh).

### DIFF
--- a/src/Features/Navigation/Navigation.php
+++ b/src/Features/Navigation/Navigation.php
@@ -22,6 +22,7 @@ class Navigation {
 	public function __construct() {
 		add_filter( 'woocommerce_admin_preload_options', array( $this, 'preload_options' ) );
 		add_filter( 'woocommerce_admin_features', array( $this, 'maybe_remove_nav_feature' ), 0 );
+		add_action( 'update_option_woocommerce_navigation_enabled', array( $this, 'reload_page_on_toggle' ), 10, 2 );
 
 		if ( Loader::is_feature_enabled( 'navigation' ) ) {
 			add_action( 'in_admin_header', array( __CLASS__, 'embed_navigation' ) );
@@ -87,5 +88,21 @@ class Navigation {
 			array(),
 			Loader::get_file_version( 'css' )
 		);
+	}
+
+	/**
+	 * Reloads the page when the option is toggled to make sure all nav features are loaded.
+	 *
+	 * @param string $old_value Old value.
+	 * @param string $value     New value.
+	 */
+	public static function reload_page_on_toggle( $old_value, $value ) {
+		if ( $old_value === $value ) {
+			return;
+		}
+
+		if ( isset( $_SERVER['REQUEST_URI'] ) ) {
+			wp_safe_redirect( wp_unslash( $_SERVER['REQUEST_URI'] ) );
+		}
 	}
 }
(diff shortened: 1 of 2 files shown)


psealock:
Interestingly, I can only get to the screen via url because my UI doesn't get the "Features" sub tab:

Screen Shot 2020-10-15 at 10 07 42 AM

I didn't put this into the sprint just yet because @pmcpinto was going to give a final go-ahead with this mechanism for toggling. I should have labelled it as such on the issue though! 

As for the page refresh, that is less than ideal, but not a deal breaker. Whats the alternative? jQuery 😨 ?

joshuatf:
>I didn't put this into the sprint just yet because @pmcpinto was going to give a final go-ahead with this mechanism for toggling. I should have labelled it as such on the issue though!

We discussed in the hangout yesterday and I think we're good to go on using this for now.  I think this could be better placed somewhere else long-term, but this is the easiest short-term option. /cc @pmcpinto In case you want to add any notes/comments.

>As for the page refresh, that is less than ideal, but not a deal breaker. Whats the alternative? jQuery 😨 ?

Not ideal either, but I'm thinking we may be able to hook into the options update hook and force another redirect when this option is updated.  Haven't attempted this yet, but may solve it.

joshuatf:
>Interestingly, I can only get to the screen via url because my UI doesn't get the "Features" sub tab:

@psealock What version of WC were you running when you saw this?

pmcpinto:
> We discussed in the hangout yesterday and I think we're good to go on using this for now. I think this could be better placed somewhere else long-term, but this is the easiest short-term option. /cc @pmcpinto In case you want to add any notes/comments.

Yeah, I talked a bit about with Elizabeth as well. Taking advantage of what already exists seems to be a good short-term solution and once we ship it in the eComm plan we can evaluate if it makes sense to change the approach. @elizaan36 what do you think?

joshuatf:
@elizaan36 @psealock Do you all have any more feedback on this?

elizaan36:
hey @joshuatf 

> Taking advantage of what already exists seems to be a good short-term solution and once we ship it in the eComm plan we can evaluate if it makes sense to change the approach

This seems right to me too. We already have a similar concept to toggle on and off the new Home screen in this location, and I'm especially thinking that long-time WC users may be familiar with this features area already (if they wanted to turn off the new Home screen.)

I see there's [another issue](https://github.com/woocommerce/woocommerce-admin/issues/5334) open for notifying users about the change and gathering feedback. That's great because there we can use a banner or other more prominent element to introduce the new nav to core users and have the opportunity to make them more comfortable with it. Even if we provide a way to toggle the nav on or off from somewhere more prominent, the toggle (or checkbox in this case) needs to also live somewhere in settings.

becdetat (approved):
This works as described, the code looks great, :shipit:

PR #5387, opened 2020-10-15
Add CES feature toggle

becdetat:
This adds a feature toggle so that the CES feature is only active in development.

This is so we can get the spiked feature into `main` and stop working against the spike branch.

- Check out this branch
- `npm install && npm start` to apply the changes to the feature configuration
- Add a new product. The CES modal should be displayed, because you're in development and the feature is enabled in development.
- Add another product. The CES modal should not be displayed.

### DIFF
--- a/src/Features/CustomerEffortScoreTracks.php
+++ b/src/Features/CustomerEffortScoreTracks.php
@@ -9,8 +9,6 @@ namespace Automattic\WooCommerce\Admin\Features;
 
 defined( 'ABSPATH' ) || exit;
 
-use \Automattic\WooCommerce\Admin\Loader;
-
 /**
  * Triggers customer effort score on several different actions.
  */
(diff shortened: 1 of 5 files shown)


src/Features/CustomerEffortScoreTracks.php
```diff
@@ -35,6 +37,10 @@ public static function get_instance() {
 	 * Constructor. Sets up filters to hook into WooCommerce.
 	 */
 	public function __construct() {
+		if ( ! Loader::is_feature_enabled( 'customer-effort-score' ) ) {
```
adrianduffell, line 40: Suggestion: Put the the feature check in the implementing code. For example:
```php
if ( Loader::is_feature_enabled( 'customer-effort-score' ) ) {
  new CustomerEffortScoreTracks()
}
```

Putting it here means we can't use the class in a standalone way. This can make unit tests for the class brittle since they would be reliant on `Loader` being in a certain state.

adrianduffell:
Tests well. I added a suggestion but it's not a blocker for merge 🚀

psealock (commented):
This is all looking fantastic. I noticed in the other PRs there is some logic to append CES on embedded pages. Since we're introducing a feature flag, it would be best to also add the client logic behind a check of `wcAdminFeatures.customer-effort-score` being `true`.

src/Features/CustomerEffortScoreTracks.php
becdetat: ah of course, thanks

adrianduffell (approved):

adrianduffell:
Looks good!

PR #5389, opened 2020-10-15
Navigation: Declare script dependencies

psealock:
When a 3rd party declares a dependency on `wc-navigation` wc-admin breaks. This is because plain luck has it that loading order works, but an outside influence can change that.

Note that the new Woo flavored Dependency extraction plugin needs to be used here to avoid such problems. Until then, here is a piecemeal fix. 

https://github.com/woocommerce/woocommerce-admin/tree/fix/nav-script-deps/packages/dependency-extraction-webpack-plugin

1. Check that wc-admin depends on its own navigation script and that navigation depends on `wp-components` to review code changes.
2. Bonus: Add a plugin that depends on `@woocommerce/navigation`, for example: https://github.com/psealock/examples. See that including this plugin on `main` results in wc-admin breaking.

### Changelog Note:

none: unreleased change

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -313,7 +313,7 @@ class Loader {
 		wp_register_script(
 			'wc-navigation',
 			self::get_url( 'navigation/index', 'js' ),
-			array( 'wp-url', 'wp-hooks' ),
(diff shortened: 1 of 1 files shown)


psealock:
closing in favour of next PR integrating wc-admin links

PR #5403, opened 2020-10-15
Only show menu group if items exist

### LINKED ISSUE #5402: Large gap is shown when menu group is empty
**Describe the bug**
When a menu group is empty (primary is empty and secondary is shown below), the group is still rendered, causing a larger gap under some categories.

**To Reproduce**

1. Go to "Extensions" or "Tools" under the new nav.
1. Note the larger gap between the category title and first item.
1. Inspect and note the empty group.

**Expected behavior**
Expected the group to not be shown.

**Screenshots**
Screen Shot 2020-10-15 at 4 43 20 PM
Screen Shot 2020-10-15 at 4 43 28 PM

joshuatf:
Fixes #5402 

Fix the gap between the title and menu items.

### Screenshots
Screen Shot 2020-10-15 at 4 53 01 PM
Screen Shot 2020-10-15 at 4 53 07 PM

1. Enable the navigation
1. Navigate to various categories
1. Make sure the gap is identical between all categories.

### DIFF
--- a/client/navigation/components/container/index.js
+++ b/client/navigation/components/container/index.js
@@ -2,7 +2,7 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { useEffect, useState } from '@wordpress/element';
+import { useEffect, useMemo, useState } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import {
 	__experimentalNavigation as Navigation,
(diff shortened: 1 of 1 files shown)


client/navigation/components/container/index.js
```diff
@@ -80,38 +80,42 @@ const Container = ( { menuItems } ) => {
...
+					const primaryItems = menuItems.filter(
+						( item ) =>
+							item.parent === category.id &&
+							item.menuId !== 'secondary'
+					);
+					const secondaryItems = menuItems.filter(
+						( item ) =>
+							item.parent === category.id &&
+							item.menuId === 'secondary'
+					);
```
psealock, line 93: Is it worthwhile to use lodash's [partition](https://lodash.com/docs/4.17.15#partition) here to reduce looping through twice to one? Taking it a step further, should we memoize here as well? `menuItems` rarely changes, but we'll be looping on every render. Would this even have an effect in performance?

psealock (commented):
Nice @joshuatf, this is all looking great and working well. Just a question on performance optimizations while we're in this part of the code.

client/navigation/components/container/index.js
joshuatf: I don't think we can use `partition` here since that will separate into 2 groups and ignore categories, but I came up with a reducer in https://github.com/woocommerce/woocommerce-admin/pull/5403/commits/3fa42b709023c80c60a040f37701b9dfd3ced93e that I think is pretty efficient.

joshuatf:
Thanks for the review @psealock!  Added some performance improvements per your feedback and this is ready for another review.

joelclimbsthings (approved):
Tested, works and looks good @joshuatf . I like the refactor with `useMemo()` :+1:

PR #5404, opened 2020-10-15
Update Notes deprecation versions.

timmyc:
While working on the release last week, I saw we had added deprecation notices for notes under the 1.6.0 release. Since that change didn't go into 1.6.0, this branch updates those version numbers to 1.7.0

I _think_ this is correct but will flag @becdetat for review to verify :)


becdetat (approved):
Looks good :-)

PR #5405, opened 2020-10-15
Homescreen: Indicator labels i18n

### LINKED ISSUE #4854: Diacritics stripped out from UI labels
**Describe the bug**
Various WooCommerce Admin UI elements labels do not handle diacritics properly. A preg_replace that handles escaping of those values strips out any non-latin letters. A simple workaround is to add an "u" PCRE_UTF8 modifier to preg_replace functions in Controllers.

Eg. Performance Indicators in Stats overview panel at the new Home screen. Correct polish translation for "Orders" is "Zamówienia" (notice that it's a single word, the "ó" letter in the middle). Then it is escaped improperly as "Zam wienia" (with a space instead of the diacritic letter, also suggesting that there are 2 words).

When adding "u" modifier to its regex at line 177 of the /wp-content/plugins/woocommerce/packages/woocommerce-admin/src/API/Reports/PerformanceIndicators/Controller.php file, diacritics works fine.

I assume that the bug exists also in other places in the UI, eg. toggles in the three dots menu of the Stats overview panel and it's not limited to this particular screen or panel.

PS. I'm also wondering why the preg_replace escaping function in those Controllers is replacing non-valid characters with space, instead of just empty string? Is it on purpose, or it's an another bug?

**To Reproduce**

1. Switch WP translation to Polish (or other language using an alphabet with diacritics) 
2. Go to WooCommerce->Home.
3. See Orders label is incorrect in Performance Indicators of the Stats overview panel.

**Expected behavior**
Diacritics should not be stripped out from labels.

**Screenshots**
Zrzut ekranu 2020-07-23 o 10 13 07

**Additional context**
WC: 4.3.1
WP: 5.4.2
WC Admin: 1.3.1 (bundled with WC)

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/4854

Descriptions were being overly regex'ed with `/\W+/` to replace non word characters with spaces. This removed special characters like `é`. Since we're fairly certain to be parsing titles and descriptions from our schemas, I think its safe to explicitly remove just the period.

### Screenshots

Screen Shot 2020-10-16 at 11 44 20 AM

You'll need to download and setup a language with diacritics, such as Polish to see the error. Instead of going through that, its much easier to replace a value in our code.

1. Replace "Total Sales." in this line with "Zamówienia."
https://github.com/woocommerce/woocommerce-admin/blob/bc076d5a07b5ecfc0addf38801e19f9ebbc523bf/src/API/Reports/Revenue/Stats/Controller.php#L167
2. Go to the Homescreen and confirm the Polish word is correctly rendered.

### Changelog Note:

- Fix: Respect languages with diacritics in Homescreen labels

### DIFF
--- a/src/API/Reports/PerformanceIndicators/Controller.php
+++ b/src/API/Reports/PerformanceIndicators/Controller.php
@@ -170,7 +170,7 @@ class Controller extends \WC_REST_Reports_Controller {
 					$stat                   = $prefix . '/' . $property_key;
 					$this->allowed_stats[]  = $stat;
 					$stat_label             = empty( $schema_info['title'] ) ? $schema_info['description'] : $schema_info['title'];
-					$this->labels[ $stat ]  = trim( preg_replace( '/\.+/', ' ', $stat_label ) );
+					$this->labels[ $stat ]  = trim( $stat_label, '.' );
 					$this->formats[ $stat ] = isset( $schema_info['format'] ) ? $schema_info['format'] : 'number';
 				}
 
--- a/src/API/Reports/PerformanceIndicators/Controller.php
+++ b/src/API/Reports/PerformanceIndicators/Controller.php
@@ -167,11 +167,10 @@ class Controller extends \WC_REST_Reports_Controller {
 						continue;
 					}
 
-					$stat                  = $prefix . '/' . $property_key;
-					$this->allowed_stats[] = $stat;
-					$stat_label            = empty( $schema_info['title'] ) ? $schema_info['description'] : $schema_info['title'];
-
-					$this->labels[ $stat ]  = trim( preg_replace( '/\W+/', ' ', $stat_label ) );
+					$stat                   = $prefix . '/' . $property_key;
+					$this->allowed_stats[]  = $stat;
+					$stat_label             = empty( $schema_info['title'] ) ? $schema_info['description'] : $schema_info['title'];
+					$this->labels[ $stat ]  = trim( preg_replace( '/\.+/', ' ', $stat_label ) );
 					$this->formats[ $stat ] = isset( $schema_info['format'] ) ? $schema_info['format'] : 'number';
 				}
 


src/API/Reports/PerformanceIndicators/Controller.php
```diff
@@ -167,11 +167,10 @@ public function get_analytics_report_data() {
...
 
-					$stat                  = $prefix . '/' . $property_key;
-					$this->allowed_stats[] = $stat;
-					$stat_label            = empty( $schema_info['title'] ) ? $schema_info['description'] : $schema_info['title'];
-
-					$this->labels[ $stat ]  = trim( preg_replace( '/\W+/', ' ', $stat_label ) );
+					$stat                   = $prefix . '/' . $property_key;
+					$this->allowed_stats[]  = $stat;
+					$stat_label             = empty( $schema_info['title'] ) ? $schema_info['description'] : $schema_info['title'];
+					$this->labels[ $stat ]  = trim( preg_replace( '/\.+/', ' ', $stat_label ) );
```
jeffstieler, line 173: Is the intention to remove leading or trailing periods? If so, I think we can use the second argument to `trim()` instead: `$this->labels[ $stat ]  = trim( $stat_label, '.' );`

timmyc: So I think we had a standardization of labels at one point, thinking @LevinMedia led the charge there. But based on  @psealock's comment [here](https://github.com/woocommerce/woocommerce-admin/issues/4854#issuecomment-709593078) I do think all we are trying to do here is replace the trailing period.

jeffstieler (approved):
👍

octaedro (approved):
This is testing well here and code LGTM :shipit:

timmyc:
cherry picked to `release/1.6.2` in 75658fe

PR #5406, opened 2020-10-16
Use snackbar before CES modal

### LINKED ISSUE #5390: Customer effort score: use a snackbar to trigger the modal
Part of #5259

The modal should be displayed when actioning a snackbar which is displayed after an action.

becdetat:
Fixes #5390

The CES modal/prompt needs to be triggered by a snackbar. This adds the snackbar to the `CustomerEffortScoreTracks` component.

image

- Add a new product
- The snackbar should appear
- Click "Give feedback"
- The CES modal should appear

### Changelog Note:

Dev: Use snackbar before CES modal


mattsherman (changes requested):
I tested this and noticed two missing design elements:

- The snackbar prompt auto-dismisses after a few seconds. In the i2 design, it mentions that the prompt should have an "X" and not auto-dismiss.
- The snackbar is missing a ✏️ emoji preceding the prompt text (i2 design calls for this).

Other than that, the snackbar worked mainly as expected.

However, I have some general usability concerns with the snackbar, which I'll bring up here to get feedback from @elizaan36:

- It is positioned in the lower left of the browser window, which can be far away from where the publish button was that triggers this. I missed the snackbar the first time I tried to trigger it, since it was out of my field of view.
- Clicking on the text prompt question just dismisses the snackbar. I didn't expect this. I think it would be nicer if clicking anywhere in the snackbar except for the "X" triggered the CES modal to be displayed.

becdetat:
> I tested this and noticed two missing design elements

Nice catch! These came in from the second design iteration and so are to be done in different tickets: #5436 #5437 #5439.

mattsherman (approved):
Thanks for pointing out the followup issues where the design details will be addressed.

This PR looks good to merge!

PR #5411, opened 2020-10-16
Display messages for empty content and network request error

### LINKED ISSUE #4581: Consider showing a non-error message for empty posts/extensions in marketing components
**Describe the bug**

As a follow up to https://github.com/woocommerce/woocommerce-admin/pull/4526#issuecomment-644236289 - the knowledge base component currently shows an error message if no posts are available.

We also show a loading screen/icon while posts are fetched.

(re) consider updating the UI to only show the components if posts/extensions are available.

I believe this was discussed during initial development but the current approach was deemed preferable at the time.

ecgan:
- Display empty content message when there is no content from the blogPosts request.
- Display error message when the blogPosts request has error.
- Also remove the use of the toast notification since we are displaying inline text messages above.

### Accessibility

Not applicable - same as existing behavior.

### Screenshots

https://example.test/wp-admin/edit.php?post_type=shop_coupon

When everything works well, you should see blog posts like this (existing behavior): 

image

When the network request is okay but there is zero blogposts returned in the response (new message text here):

Screenshot on 2020-10-16 at 21-51-59

When there is an error in the network request (this is an existing UI):

Screenshot on 2020-10-16 at 16-29-11

- Ex: Open page `https://example.test/wp-admin/edit.php?post_type=shop_coupon`
- Scroll to the bottom to see the UI.
- To simulate network failure, you can change the [line 45 in `client/marketing/data/resolvers.js`](https://github.com/woocommerce/woocommerce-admin/blob/7df2ee2dfe76636a32ff657c2d0962ee4a180391/client/marketing/data/resolvers.js#L45) to be:

  ```js
  path: `${ API_NAMESPACE }/knowledge-base-THIS-WILL-NOT-WORK${ categoryParam }`,  // simulate network failure.
  ```

- To simulate empty blog post content, modify [line 58 in `client/marketing/data/reducer.js`](https://github.com/woocommerce/woocommerce-admin/blob/7df2ee2dfe76636a32ff657c2d0962ee4a180391/client/marketing/data/reducer.js#L58) to be:

  ```js
  [ action.data.category ]: [], // empty array.
  ```

### Changelog Note:

Fix: display the right messages for when there is an error in requesting blog posts and when there is zero blog posts.


layoutd (approved):
LGTM but I'll leave the in-depth React review to someone with more exp.

One change I can see would be to remove the giant error symbol from the empty results message (I guess with `illustration={ '' }` in `EmptyContent`?)

jconroy (approved):
Nice work @ecgan that is testing well for me 👏  - tests are passing and linting looks good too!

Screen Shot 2020-10-19 at 4 33 41 pm

Screen Shot 2020-10-19 at 4 34 20 pm

For the detailed test instructions, it's often useful to elaborate on how the tester can check different situations e.g.

* removing the `_transient_wc_marketing_knowledge_base_coupons` option value completely will simulate an error message
* setting the `_transient_wc_marketing_knowledge_base_coupons` option value to `a:0:{}` will simulate not being able to retrieve any posts.

---

The thing we were considering on the original issue (and apologies it probably wasn't 100% clear due to my wording)

> consider updating the UI to only show the components if posts/extensions are available.

was whether we should hide the entire knowledge base component (for example) if there are not posts. 

So e.g. don't show the following at all

Screen Shot 2020-10-19 at 4 33 41 pm

If there aren't any posts. Or another way of looking at it. Only show the following if there are posts

Screen Shot 2020-10-19 at 4 51 07 pm

We can run with the changes in this PR and get input from @j111q as a follow up on #4581

ecgan:
@jconroy , appreciate you to fill in my knowledge gap below please 😄 

> it's often useful to elaborate on how the tester can check different situations

Is the tester doing blackbox testing, or doing whitebox testing and and is technical and knowledgeable with code...?

> - removing the _transient_wc_marketing_knowledge_base_coupons option value completely will simulate an error message
> - setting the _transient_wc_marketing_knowledge_base_coupons option value to a:0:{} will simulate not being able to retrieve any posts.

How do we actually do that? I searched in code and the closest thing I found is in `src/Features/Marketing.php`: 

```php
const KNOWLEDGE_BASE_TRANSIENT = 'wc_marketing_knowledge_base';
```

FYI, the way I tested the UI was by manually editing the code and the data in the redux store.

jconroy:
> Is the tester doing blackbox testing, or doing whitebox testing

For me I think it should be bit of a mix for a quality review (and depending on complexity).

A reviewer should check that the PR works and fixes the problem but also check the internal workings are ok etc. 

They should also check for the unexpected, what has the author missed?

> and is technical and knowledgeable with code

Yep definitely - but sometimes might be not familiar with the codebase or how the application works - or that area of the codebase if we have multiple teams working on it (like wc-admin).

> How do we actually do that?

Sure..

You might not have touched on [transients](https://developer.wordpress.org/apis/handbook/transients/) yet (tl;dr they are temporary cached data usually stored in the wp_options table - but it is also common for sites to offload them to an object caching system for performance reasons).

If you use a mysql client like sequel pro you'll be able to view/edit any data in the database.

We use the transients API in [places like here](https://github.com/woocommerce/woocommerce-admin/blob/main/src/Features/Marketing.php#L189-L207) to temporarily cache the data to save making constant requests to WCCOM.

> the way I tested the UI was by manually editing the code and the data in the redux store.

Gotcha, that works as well. 

---

Part of the idea of having detailed test instructions (for me) is to help make testing easier so that anyone can pick up the PR and review it - it provides a starting point for testing. It also helps clarify and set a common understanding for what is expected to happen. 

Sometimes the detailed test instructions writen by the author don't even work as expected for the reviewer :-) - so it can be a sometimes be quick way to check a fix.

It is also possible someone in the future might come back to the PR and the instructions provide extra context.

ecgan:
@jconroy , thanks for the detailed explanation! I've updated the description to include more detailed test instructions.

j111q:
Just wanted to +1 on @layoutd's suggestion of removing the giant error icon; it is very giant 😂

Issue #5412, opened 2020-10-16
WCPay inbox note stalls on install action


dechov:
**Describe the bug**

When taking the "Install now" action on the WooCommerce Payments note (which appears for US stores on 7-day-old stores without WCPay), the button **stays "busy" indefinitely**, with no indication of the installation completing (though it does) or of any next step.

![stall-on-install](https://user-images.githubusercontent.com/1867547/96225446-185d2780-0f5f-11eb-8f9c-c082355f5331.gif)

After a moment, the note turns to actioned/read styling (and the timestamp updates due to an apparent unrelated timezone bug) once the plugin is installed & activated and the note action request comes back, but the button stays busy.

See paJDYF-RI-p2#comment-3727 and paqDfi-Pt-p2#comment-2531 – props to @jrodger for noticing this behavior and screen capturing it.

**To Reproduce**

Click on the "Install now" action on the "Try the new way to get paid" Inbox note, and observe that it stays busy forever 😄 

This note is triggered to appear by existing logic if a US store without WooCommerce Payments has existed for 7 days (provided "Display suggestions within WooCommerce" is left checked in advanced settings). Could be a better way, but I was doing this by calling `WooCommerce_Payments::possibly_add_note();` from an init function that runs on each page load, and removing most of the checks from that function. To manually create the note in the DB, run:
```sql
INSERT INTO `wp_wc_admin_notes` (`name`, `type`, `locale`, `title`, `content`, `icon`, `content_data`, `status`, `source`, `date_created`, `date_reminder`, `is_snoozable`, `layout`, `image`, `is_deleted`)
VALUES
	('wc-admin-woocommerce-payments', 'marketing', 'en_US', 'Try the new way to get paid', 'Securely accept credit and debit cards on your site. Manage transactions without leaving your WordPress dashboard. Only with WooCommerce Payments.', 'info', '{}', 'unactioned', 'woocommerce-admin', '2020-10-16 09:45:15', NULL, 0, 'plain', '', 0);
```

…and to create actions, replace `[NOTE_ID]` with the ID of that new note and run:
```sql
INSERT INTO `wp_wc_admin_note_actions` (`note_id`, `name`, `label`, `query`, `status`, `is_primary`, `actioned_text`)
VALUES
	([NOTE_ID], 'learn-more', 'Learn more', 'https://woocommerce.com/payments/', 'unactioned', 0, ''),
[...]

PR #5085, opened 2020-09-03
Add ToS agreement copy to the WCPay setup task

becdetat:
Fixes #5078

This adds copy re accepting the ToS to the WCPay task

### Accessibility

- [x] I've tested using only a keyboard (no mouse)

### Screenshots
image

### Detailed test instructions:

- Create a site in the US, with WooCommerce Payments installed
- Go to the WooCommerce home screen
- The above copy should be added to the WCPay task
- Following the link should open the ToS in a new tab, and not activate the task
- Clicking the task outside of the link should activate the task

### Changelog Note:

Tweak: Add ToS agreement copy to the WCPay setup task

PR #5413, opened 2020-10-16
Redirect instead of stalling on WCPay inbox note install action

dechov:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/5412

Switches the WCPay inbox note **from no URL** – and performing plugin installation on the async action hook, resulting in a busy state that doesn't resolve – **to an `action` param** admin URL, following (plugin _de_-activation) precedent in https://github.com/woocommerce/woocommerce-admin/pull/3687.

This allows redirecting to the next setup step after allowing installation & activation to complete, which is not possible to coordinate from the hook with a note URL alone. Adding a ToS message enables navigating directly into the WCPay onboarding flow.

Builds on earlier commits written and tested by:
- @jrodger in https://github.com/woocommerce/woocommerce-admin/pull/5361 (handling action URL param with redirect)
- @vbelolapotkov in https://github.com/woocommerce/woocommerce-admin/pull/5362 (enhancing existing note class with next step)

Open to hearing reservations (raised in https://github.com/woocommerce/woocommerce-admin/pull/5361 description) about activating a plugin from a (nonce-less) GET URL, and whether it would be better to at least wait for the note status to change (to "actioned", triggered by the parallel AJAX request) before proceeding with the installation.

_Edit:_ Also, is there a better way to be dealing with errors in the handler? Currently it will go to the home screen with the URL param still there.

### Screenshots

wcpay-note-redirect-to-setup

- Add WCPay Inbox note. I used [this change](https://github.com/woocommerce/woocommerce-admin/commit/17d141504f02c43e7721892f12fd5c6dff4f5f48) to force the note to show up, but if the note actions already exist with no URL, the note will need to be deleted first, either directly in the `wp_wc_admin_notes` table or with `Notes::delete_notes_with_name( 'wc-admin-woocommerce-payments' );`
- Take the "Get started" action on the Inbox note
- Verify that the WooCommerce Payments plugin is installed & activated, and that it redirects to the stripe.com account onboarding flow

### Changelog Note:

Fix: Avoid stalling after WCPay installation from Inbox note by facilitating setup process


src/Notes/WooCommercePayments.php
```diff
@@ -128,22 +135,58 @@ protected static function is_installed() {
...
 
-			$activate_request = array( 'plugins' => self::PLUGIN_SLUG );
-			$installer->activate_plugins( $activate_request );
+	/**
+	 * Install & activate WooCommerce Payments plugin, and redirect to setup.
+	 */
+	public function install_on_action() {
+		// TODO: Need to validate this request more strictly since we're taking install actions directly?
+		/* phpcs:disable WordPress.Security.NonceVerification */
+		if (
```
vbelolapotkov, line 165: ToS acceptance is tricky part here. At the moment we kind of can't ensure the action was taken by merchant clicking and accepting the ToS. Maybe before redirecting we need to check ToS acceptance somehow? Related to Automattic/woocommerce-payments#887

src/Notes/WooCommercePayments.php
```diff
@@ -98,15 +98,22 @@ public static function possibly_add_note() {
...
 		$note->set_title( __( 'Try the new way to get paid', 'woocommerce-admin' ) );
-		$note->set_content( __( 'Securely accept credit and debit cards on your site. Manage transactions without leaving your WordPress dashboard. Only with WooCommerce Payments.', 'woocommerce-admin' ) );
+		$note->set_content( __( 'Securely accept credit and debit cards on your site. Manage transactions without leaving your WordPress dashboard. Only with WooCommerce Payments.', 'woocommerce-admin' ) . '<br><br>' . $tos );
 		$note->set_content_data( (object) array() );
 		$note->set_type( Note::E_WC_ADMIN_NOTE_MARKETING );
 		$note->set_name( self::NOTE_NAME );
 		$note->set_source( 'woocommerce-admin' );
 		$note->add_action( 'learn-more', __( 'Learn more', 'woocommerce-admin' ), 'https://woocommerce.com/payments/', Note::E_WC_ADMIN_NOTE_UNACTIONED );
-		$note->add_action( 'install-now', __( 'Install now', 'woocommerce-admin' ), false, Note::E_WC_ADMIN_NOTE_ACTIONED, true );
+		$note->add_action( 'get-started', __( 'Get started', 'woocommerce-admin' ), wc_admin_url( '&action=install-woocommerce-payments' ), Note::E_WC_ADMIN_NOTE_ACTIONED, true );
```
vbelolapotkov, line 116: Action name probably should be something different because it installs, activates and sets up the plugin. So `setup-woocommerce-payments` might be better.

src/Notes/WooCommercePayments.php
```diff
@@ -128,22 +135,58 @@ protected static function is_installed() {
...
 
 	/**
-	 * Install WooCommerce Payments when note is actioned.
+	 * Install and activate WooCommerce Payments.
 	 *
-	 * @param Note $note Note being acted upon.
+	 * @return boolean Whether the plugin was successfully activated.
 	 */
-	public function install( $note ) {
-		if ( self::NOTE_NAME === $note->get_name() && current_user_can( 'install_plugins' ) ) {
```
vbelolapotkov, line 136: I think we should keep at least this check: `current_user_can( 'install_plugins' )`

vbelolapotkov: Haven't noticed the check below, sorry.

src/Notes/WooCommercePayments.php
dechov: 👍 Done in https://github.com/woocommerce/woocommerce-admin/pull/5413/commits/e049694bd19979e151c5fbf36f6c49dd40dc2fe8

src/Notes/WooCommercePayments.php
dechov: I'm not caught up on every nuance there, but as far as I can tell, we may be able to address that issue by noting the acceptance data immediately before redirect (and continue to assume that ToS acceptance was presented in one way or another before then)? cc @marcinbot @allendav @thenbrent

In any case, I don't think this redirect would behave any different in that regard from the WCPay task ones, one of which recently had a ToS acceptance added in https://github.com/woocommerce/woocommerce-admin/pull/5085, prompted by that thread – paJDYF-OD-p2#comment-3315.

marcinbot: I think the consensus was that as long as we present the bit about agreeing to ToS (while also linking to it) we're fine - clicking the button is treated as an agreement.

becdetat (approved):
Tests well, code looks good. Thanks, merge away :-)

timmyc:
Hey y'all ( @dechov @becdetat ) couple of questions on getting this into 1.6.3 and 4.7 

1- Looks like this branch is done against `main` and as such the diff is done against the new file name for this note, in the `release/1.6.x` branches the file is named `WC_Admin_Notes_WooCommerce_Payments.php`. So we will need to port these changes, ideally against the `release/1.6.3` branch as a new PR. I suppose I could do some git fu to create a patch from the commit diff, but I feel like having a PR with testing would be a safer route to go.

2- By removing the old action in this Note class, might we possibly end up with sites who already have the older version of this note ( say from Woo 4.6 ), that could then update, and the old actions in the DB no longer have a corresponding action for `woocommerce_note_action_install-now`?

dechov:
@timmyc Great points. By tomorrow I'll open a PR against `release/1.6.3` and include a change to keep the existing handling behavior as well (which can be added to the main branch as well as a low priority fix). Will just need to make sure that two parallel installation processes wouldn't conflict with each other, and try to prevent that if they could.

timmyc:
Thanks @dechov - sounds perfect.

dechov:
@timmyc I've opened https://github.com/woocommerce/woocommerce-admin/pull/5471 which applies the change and restores the legacy action handler.

> Will just need to make sure that two parallel installation processes wouldn't conflict with each other, and try to prevent that if they could.

Actually this is not a concern – I'd forgotten for a moment that the action name is already different.

PR #5416, opened 2020-10-16
Home Screen: Always show the inbox for temporary gutenberg plugin fix.

### LINKED ISSUE #609: eComm plan - Inbox feature missing in the home screen
The inbox panel disappeared from the home screen in the eCommerce plan with the WooCommerce 4.6 update. This is happening in Calypso and WP-Admin:

Calypso: https://d.pr/i/6LuLM7
WP-Admin: https://d.pr/i/3JLF2Z

timmyc:
Fixes https://github.com/Automattic/wc-calypso-bridge/issues/609

While this is not the ultimate fix for the missing Inbox on the home screen when the Gutenberg plugin is active, it at least gets the inbox back to a functional state there. The change I'm proposing here is to remove the feature we have where when the inbox is empty, it is not shown on the home screen.

So the big change visually would be that when the inbox is empty, we still show the two column layout ( which I prefer TBH ), with an empty inbox:

_Empty Inbox on Home_
image

_How the home screen looks in 4.6 with Gutenberg 9.1.1 Installed_
Home ‹ WooCommerce ‹ bend outdoors — WooCommerce 2020-10-16 08-30-28

__Home Screen Gutenberg 9.1.1, with this branch and notes in inbox__
home-with-inbox

@pmcpinto  wanted to see if you were okay with this work-around until we can come up with a longer-term solution to have the inbox not shown on the home screen when empty.

- Probably best to setup the "error state" for this PR first, so install and activate Gutenberg 9.1.1
- Make sure you have WooCommerce 4.6 installed
- Go to the home screen and note that the inbox is not shown, probably good to verify you had inbox notes before, or you can do this by visiting the orders listing page to see your inbox does have notes
- Checkout this branch and activate. Since this is on 1.7.0-dev, it should be auto used by Jetpack autoloader
- Refresh the home screen a few times and verify the Inbox is now shown.

### Changelog Note:

Tweak: Always show the inbox on the home screen

### DIFF
--- a/client/homescreen/layout.js
+++ b/client/homescreen/layout.js
@@ -64,13 +64,6 @@ export const Layout = ( {
 	const isTaskListEnabled = taskListHidden === false && ! taskListComplete;
 	const isDashboardShown = ! isTaskListEnabled || ! query.task;
 
-	const isInboxPanelEmpty = ( isEmpty ) => {
-		if ( isBatchUpdating ) {
-			return;
-		}
-		setShowInbox( ! isEmpty );
-	};
-
 	if ( isBatchUpdating && ! showInbox ) {
 		setShowInbox( true );
 	}
@@ -80,7 +73,7 @@ export const Layout = ( {
 			<Fragment>
 				{ showInbox && (
 					<div className="woocommerce-homescreen-column is-inbox">
-						<InboxPanel isPanelEmpty={ isInboxPanelEmpty } />
+						<InboxPanel />
 					</div>
 				) }
 				<div


client/homescreen/layout.js
```diff
@@ -80,7 +73,7 @@ export const Layout = ( {
 			<Fragment>
 				{ showInbox && (
 					<div className="woocommerce-homescreen-column is-inbox">
-						<InboxPanel isPanelEmpty={ isInboxPanelEmpty } />
+						<InboxPanel />
```
jeffstieler, line 76: Can we remove the code that uses `isPanelEmpty` in the `InboxPanel` component as well?

timmyc: @jeffstieler given our conversation of how single column layout will likely remove this functionality, I think that is fair to do. Let me push up a change for that.

pmcpinto:
> @pmcpinto wanted to see if you were okay with this work-around until we can come up with a longer-term solution to have the inbox not shown on the home screen when empty.

:shipit:

client/homescreen/layout.js
timmyc: Actually, since I've based this branch of `main` and the Inbox panel logic has since migrated, that will complicate the picking. Is it okay if we remove that logic in a follow-up PR against `main`?

timmyc: While it is dead/un-used code, the fix here should still be valid is I guess what I'm trying to say but having difficulty doing so 😸

jeffstieler (approved):
👍

timmyc:
cherry picked to `release/1.6.2` in https://github.com/woocommerce/woocommerce-admin/commit/02397c25a3ec02d81466197f92abdb9b86152219

PR #5417, opened 2020-10-16
use consistent markdown headers in navigation readme

rrennick:
I noticed this when looking at the package on [npmjs.com](https://www.npmjs.com/package/@woocommerce/navigation) due to the significant increase in the font size.


becdetat (approved):
Looks good, thanks :-).

PR #5420, opened 2020-10-19
Migrate navigation store to data package

### LINKED ISSUE #5351: Navigation: Migrate store to wp.data
Migrating Navigation from WC Navigation repo brought along with it its own store implementation for keeping track of populating navigation items. 

It may make sense to migrate the store to wc-admin's data package. Evaluate if it should be its own store, added onto an existing store, or determine if its needed altogether.

joshuatf:
Fixes #5351 

Migrates nav store to data package.

1. Enable the nav.
1. Navigate to various pages using the nav.
1. Make sure route detection works as expected and active item is correctly marked active.
1. Make sure tests pass.


becdetat (approved):
Works as advertised, tests all pass (with some unrelated warnings) and code looks good. :shipit:

PR #5419, opened 2020-10-19
Enqueue styles on non-WC pages

joshuatf:
Fixes #5380

Add styles to non-WC pages to hide submenu styling.

#### Before
Screen Shot 2020-10-19 at 12 24 59 PM

#### After
Screen Shot 2020-10-19 at 12 24 42 PM

### Detailed test instructions:

1. Enable the nav feature.
1. Hover the "WooCommerce" menu item.
1. Make sure that the submenu and arrow are no longer shown.

PR #5422, opened 2020-10-19
Exclude certain nav items from core migration

### LINKED ISSUE #5401: Reports moved to Settings category in navigation
**Describe the bug**
"Reports" is included under "Settings" in the new navigation.  This is probably because it was left under the original WooCommerce menu item.

Do we want to keep these legacy Reports somewhere or exclude them from the new nav experience?  /cc @jameskoster @pmcpinto 

**To Reproduce**

1. Enable the nav feature.
1. Navigate to the "Settings" category.
1. Note "Reports."

**Screenshots**
Screen Shot 2020-10-15 at 4 38 28 PM

joshuatf:
Fixes #5401

Creates an excluded list of items from being migrated under the WC core menu.  Currently this removes the main WC menu item and the legacy "Reports" menu item.

Also removes the `menuId` from nested items since these are not actually under a "secondary" menu and are primary menu items in their respective categories.

### Screenshots
Screen Shot 2020-10-19 at 3 14 12 PM

1. Enable the nav feature.
1. Navigate to Settings and make sure "Reports" does not exist.
1. Optionally rebase from https://github.com/woocommerce/woocommerce-admin/pull/5419 to see "Reports" hidden under the old menu.

### DIFF
--- a/src/Features/Navigation/CoreMenu.php
+++ b/src/Features/Navigation/CoreMenu.php
@@ -111,7 +111,6 @@ class CoreMenu {
 				'id'         => 'my-extensions',
 				'url'        => 'plugins.php',
 				'migrate'    => false,
-				'menuId'     => 'secondary',
 			)
 		);
 		Menu::add_item(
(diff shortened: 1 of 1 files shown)


psealock (approved):
Works well, tests very well + with rebase. I like the approach to sunsetting old Reports as well.

PR #5423, opened 2020-10-19
Inbox Panel: Remove no longer used isPanelEmpty logic.

timmyc:
This is a follow-up to #5416 - which removed the need for the `isPanelEmpty` logic since we are now always showing the inbox panel to alleviate some of the bugs we had been encountering with the Gutenberg plugin installed alongside wc-admin.

__To Test__
- Verify the Inbox panel is still shown on the home screen, even if no inbox messages are present.
- Verify the Inbox panel still operates on an embedded page like the Orders Listing page.


becdetat (approved):
Looks good, the panel still works 👍

PR #5428, opened 2020-10-20
Convert the marketing tab cards to use the new @wordpress/components Card component.

### LINKED ISSUE #5374: Convert the marketing tab cards to use the new @wordpress/components Card component.
Investigate moving the [Marketing tab cards](https://github.com/woocommerce/woocommerce-admin/tree/main/client/marketing) over to using the @wordpress/components Card component.

Currently, we use the woo card component.

Woo -> https://github.com/woocommerce/woocommerce-admin/tree/main/packages/components/src/card
Wp -> https://github.com/WordPress/gutenberg/tree/master/packages/components/src/card

My understanding is the WP card component is newer and wc-admin is shifting towards using the WP card moving forward.

Some of the recent work in the homescreen uses the WP card component 

https://github.com/woocommerce/woocommerce-admin/blob/main/client/homescreen/stats-overview/index.js

ecgan:
Fixes #5374 

Use new WordPress Card component to replace older WooCommerce Card component.

### Accessibility

Not applicable - pretty much similar to previous UI.

### Screenshots

Screenshot on 2020-10-21 at 23-31-59

Continue:

Screenshot on 2020-10-21 at 23-32-34

- Open the Marketing tab page: https://example.test/wp-admin/admin.php?page=wc-admin&path=%2Fmarketing 
- The new UI should be similar to previous UI, nothing breaks.

### Changelog Note:

Dev: Use new @wordpress/components Card component, replacing older WooCommerce Card component.


jconroy:
@danielbitzer mind taking a review of this please?

client/marketing/overview/style.scss
```diff
@@ -2,6 +2,10 @@
 	max-width: 1032px;
 	margin: 0 auto;
 
+	.components-card {
+		margin-bottom: 24px;
+	}
+
 	.woocommerce-card__header {
 		padding: 16px 24px;
 		border-bottom: 1px solid $gray-100;
```
danielbitzer, line 11: Looks like these `.woocommerce-card__*` styles aren't being used any more and should be removed.

client/marketing/overview/installed-extensions/index.js
```diff
@@ -6,14 +6,14 @@ import { __ } from '@wordpress/i18n';
...
 
 /**
  * Internal dependencies
  */
 import './style.scss';
 import InstalledExtensionRow from './row';
 import { STORE_KEY } from '../../data/constants';
+import { Card } from '../../components';
 
 class InstalledExtensions extends Component {
```
danielbitzer, line 18: It looks like the border colour for each row in this card could be updated to match the border color at the bottom of the card header.

danielbitzer (changes requested):
Nice work here @ecgan!

The new tests look great. I've left a few inline comments.

Just FYI marketing cards are also used on the coupons page where there is now a little spacing issue:

client/marketing/components/card/index.js
```diff
@@ -0,0 +1,49 @@
...
+	const { title, description, children, className } = props;
+
+	return (
+		<WPCard isElevated className={ className }>
+			<CardHeader>
+				<div>
+					<Text variant="title.small" as="h2">
+						{ title }
+					</Text>
+					<Text variant="body">{ description }</Text>
```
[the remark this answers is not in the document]
ecgan: @danielbitzer , yeah actually I was wondering whether I should make it similar to the previous style or use the style that comes with WordPress Text component. Usually I prefer to stick to the design system's style (i.e. WordPress). I'll make it look like the previous style then.

client/marketing/components/card/index.js
```diff
@@ -0,0 +1,49 @@
...
+	CardHeader,
+	__experimentalText as Text,
+} from '@wordpress/components';
+import PropTypes from 'prop-types';
+
+const Card = ( props ) => {
+	const { title, description, children, className } = props;
+
+	return (
+		<WPCard isElevated className={ className }>
```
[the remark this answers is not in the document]
ecgan: @danielbitzer , 

> rather match the style of the cards on the WooCommerce > Home screen.

That's a good idea! 😄👍

ecgan:
@danielbitzer ,

> Just FYI marketing cards are also used on the coupons page where there is now a little spacing issue:

Hey thanks for checking out the UI! I have done some component dependency check just now to make sure I didn't miss out anything or causing issues at other pages, I think it works okay now 😄

client/marketing/components/card/index.js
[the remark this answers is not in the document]
danielbitzer: @ecgan Yep that makes sense. See on the WooCommerce Home screen there is a pretty similar style. Perhaps we can copy how they did that?

danielbitzer (approved):
LGTM 👍  

I just left one thought about the subtitle style.

client/marketing/overview/installed-extensions/style.scss
```diff
@@ -22,7 +18,7 @@
 	}
 
 	&__item:not(:last-child) {
-		border-bottom: 1px solid $gray-100;
+		border-bottom: 1px solid rgb(226, 228, 231);
```
becdetat, line 21: Is there a CSS variable you could use here instead of hardcoding the RGB?

client/marketing/components/card/index.js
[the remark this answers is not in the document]
ecgan: @danielbitzer , that's a good idea! I don't have those "Renew Subscription" cards in my local environment setup. Thank you for pointing these out to me! 😃

client/marketing/overview/installed-extensions/style.scss
ecgan: @becdetat , I end up using `$gray-200` from `@wordpress/base-styles`, which is meant for "light borders" and is also the closest to the desired border color.

client/stylesheets/abstracts/_colors.scss
```diff
@@ -1,11 +1,6 @@
 @import 'node_modules/@automattic/color-studio/dist/color-variables.scss';
 @import 'node_modules/@wordpress/base-styles/colors';
 
-// TODO: Remove this when the next update of @wordpress/base-styles is
-// released (see https://github.com/woocommerce/woocommerce-admin/pull/4707/files#r451921109)
-$gray-700: #757575; // Meets 4.6:1 text contrast against white.
-$gray-100: #f0f0f0;
-
```
ecgan, line 8: @becdetat , FYI I have also removed this block of code based on the TODO comment that you left. 🙂

becdetat:
Thanks @ecgan , everything is looking good and works well :-D

PR #5429, opened 2020-10-20
Add Homescreen layout control

### LINKED ISSUE #5224: Home Screen - layout control
**User story:**
As a merchant
I want to have more control over the layout
So I can tailor the page according to my needs

**Description:** as we iterate on the home screen we think that a single-column layout can be more effective. However, this can be an abrupt change for users, as they're getting used to the two-column layout. The best option seems to provide an intermediate step where we allow users to choose the layout that they want to use. In the future, this can also be used to show/hide specific home screen sections (under evaluation)

**Acceptance criteria:**
- [ ] The user is able to switch between the one-column and two-column layout through a Layout option in the navbar.

**Event tracking:**
Click the Display icon:
`wcadmin_homescreen_display_click`
Event props: none

Select one Display option:
`wcadmin_homescreen_display_option`
Even prop: `display_option` : `single_column` | `two_columns`

**Additional links:**
[P2 with link to Figma ](pbIJXs-ma-p2)

jeffstieler:
Fixes #5224.

This PR seeks to implement the Homescreen layout control described in the attached issue.

The following changes were made:
- Introduction of the `woocommerce_admin_homepage_layout` user meta field
- Replacement of the `hasInbox` class with `two-columns` (controlled by `woocommerce_admin_homepage_layout`)
- Addition of "Display" item to activity panel (note - it's not implemented as a `Tab` since those are being removed)
- Tweaking of styling to match Figma - namely making the single column wider

### Accessibility

### Screenshots

Screen Shot 2020-10-20 at 1 10 09 PM

- Verify that the "display" button/tab is only shown on the Homescreen
- Verify that "two columns" is the default
- Verify that selecting "single column" changes the layout, and persists after a refresh

### Changelog Note:

Enhancement: add ability to toggle homescreen layouts.


client/header/activity-panel/display-options/icons/display.js
```diff
@@ -0,0 +1,44 @@
+/**
+ * External dependencies
+ */
+import { Fragment } from '@wordpress/element';
```
samueljseay, line 4: fwiw in React you can use the empty tag these days instead of importing Fragment. its nice and succinct so I prefer it.

e.g.

```
export default () => (
   <>
     <svg... />
   </>
)
```

client/header/activity-panel/display-options/index.js
```diff
@@ -0,0 +1,84 @@
+/**
+ * External dependencies
+ */
+import {
+	DropdownMenu,
+	MenuGroup,
+	MenuItemsChoice,
+} from '@wordpress/components';
+import { Fragment } from '@wordpress/element';
```
samueljseay, line 9: Again we could use the empty element tag `<></>` here.

client/homescreen/test/index.js
```diff
@@ -142,4 +148,47 @@ describe( 'Homescreen Layout', () => {
...
+				requestingTaskList={ false }
+				taskListHidden={ false }
+				query={ {} }
+				updateOptions={ () => {} }
+			/>
+		);
+
+		// Expect that we're rendering the two-column home screen.
+		expect(
+			container.getElementsByClassName(
```
samueljseay, line 167: Great to see these tests 🎉

samueljseay (commented):
Just a few little things, but looking good overall.

samueljseay:
In testing I noticed the menu doesn't look right at mobile sizes:

Screenshot 2020-10-21 at 2 19 54 PM

jeffstieler:
Add feedback should be addressed. This is ready for another review.

samueljseay (commented):
The changes all look great but the mobile styling still looks broken at most smaller sizes for me. Screenshot:

Screenshot 2020-10-22 at 5 16 01 PM

becdetat:
I've noticed an edge case where if you refresh the page immediately after changing a different layout, the layout selection isn't persisted.

jeffstieler:
> mobile styling still looks broken

@samueljseay I believe those styles broke in the WP components update - it looks the same for me on `main`.

jeffstieler:
> I've noticed an edge case where if you refresh the page immediately after changing a different layout, the layout selection isn't persisted.

How difficult is it for you to trigger that edge case? For me it was quite tricky - I had to almost simultaneously click the layout setting and hit refresh in order to trigger it.

samueljseay (approved):
Approving based on:

1. Addressing the mobile issues separately since they exist on main anyway (can we raise an issue for this?)

2. The edge case of it not saving sounds like its a situation where the XHR request hasn't completed by the time you refresh the page, I don't think we can solve that easily.

PR #5445, opened 2020-10-21
Navigation: extend with WC Admin router links

### LINKED ISSUE #5328: Navigation: Add wc-admin links
Extend Navigation by adding WC Admin links which are router-aware.

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/5328

Replace static links with WC Admin's `<Link />` component so navigation can occur without page refreshes between wc-admin pages.

* Use `registerPlugin` to add Fills to Slots.
* Only add Fills if the page is WC Admin, otherwise there is no router instantiated.
* Export relevant components from wp.components as a work around to the `context` issue.

1. Visit any WC Admin page.
2. Use the Navigation to navigation to any other WC Admin page and note the lack of page refresh.
3. Try going from non-wc admin to wc-admin and vice versa to ensure nothing is broken.

### Follow up issues

* I'd like to get a closer look at the `context` issue and document it in a p2 for discoverability. This way, we can note the issue and continue development.
* I needed to add `id` on the client side to results of `getReports` and `getPages` to match the id's registered in PHP so that Slot and Fill names can be correctly identified. This means that plugins currently using hooks like `woocommerce_admin_pages_list` to add pages will need to update. Since we have that information already from PHP, it feels like there is a better solution to investigate so that we can keep the filter API the same. Here is an example from WC Payments that would need updating.

https://github.com/Automattic/woocommerce-payments/blob/c7f9d58e2d20fc5fb10d3fa5d205f698bcf3b8c6/client/index.js#L24


client/analytics/report/get-reports.js
```diff
@@ -88,11 +88,6 @@ export default () => {
 			title: __( 'Taxes', 'woocommerce-admin' ),
 			component: TaxesReport,
 		},
-		{
-			report: 'downloads',
-			title: __( 'Downloads', 'woocommerce-admin' ),
-			component: DownloadsReport,
-		},
```
psealock, line 95: There were duplicate entries for `downloads`

becdetat (approved):
This looks great and tests really well! Nice work.

psealock:
Thanks @becdetat 🎉 

Issue to track refactoring client side gathering pages and reports so they include an id supplied by PHP https://github.com/woocommerce/woocommerce-admin/issues/5458

PR #5453, opened 2020-10-21
Display a "read blog" message

### LINKED ISSUE #4581: Consider showing a non-error message for empty posts/extensions in marketing components

ecgan:
Fixes #4581 

Display a "read blog" message when there is an error in getting blog posts or there is no blog posts.

### Accessibility

Not applicable.

### Screenshots

Screenshot on 2020-10-22 at 00-53-24

Screenshot on 2020-10-22 at 01-40-32

- Open pafe https://ganwp1.test/wp-admin/admin.php?page=wc-admin&path=%2Fmarketing
- Scroll to the bottom to see the UI.
- To simulate network failure, you can change the [line 45 in `client/marketing/data/resolvers.js`](https://github.com/woocommerce/woocommerce-admin/blob/7df2ee2dfe76636a32ff657c2d0962ee4a180391/client/marketing/data/resolvers.js#L45) to be:

  ```js
  path: `${ API_NAMESPACE }/knowledge-base-THIS-WILL-NOT-WORK${ categoryParam }`,  // simulate network failure.
  ```

- To simulate empty blog post content, modify [line 58 in `client/marketing/data/reducer.js`](https://github.com/woocommerce/woocommerce-admin/blob/7df2ee2dfe76636a32ff657c2d0962ee4a180391/client/marketing/data/reducer.js#L58) to be:

  ```js
  [ action.data.category ]: [], // empty array.
  ```

### Changelog Note:

Enhancement: Display a "read blog" message when there is an error in getting blog posts or there is no blog posts.

### DIFF
--- a/client/marketing/components/knowledge-base/ReadBlogMessage.js
+++ b/client/marketing/components/knowledge-base/ReadBlogMessage.js
@@ -1,36 +1,26 @@
 /**
  * External dependencies
  */
-import { _x } from '@wordpress/i18n';
+import { __ } from '@wordpress/i18n';
+import { Link } from '@woocommerce/components';
+import interpolateComponents from 'interpolate-components';
 
 const ReadBlogMessage = () => {
-	const message1 = _x(
-		'Read ',
-		'Read <a href="https://woocommerce.com/blog/marketing/coupons/">the WooCommerce blog</a> for more tips on marketing your store',
-		'woocommerce-admin'
-	);
-
-	const message2 = _x(
-		'the WooCommerce blog',
-		'Read <a href="https://woocommerce.com/blog/marketing/coupons/">the WooCommerce blog</a> for more tips on marketing your store',
-		'woocommerce-admin'
-	);
-
-	const message3 = _x(
-		' for more tips on marketing your store',
-		'Read <a href="https://woocommerce.com/blog/marketing/coupons/">the WooCommerce blog</a> for more tips on marketing your store',
-		'woocommerce-admin'
-	);
-
-	return (
-		<>
-			{ message1 }
-			<a href="https://woocommerce.com/blog/marketing/coupons/">
-				{ message2 }
-			</a>
-			{ message3 }
-		</>
-	);
+	return interpolateComponents( {
+		mixedString: __(
+			'Read {{link}}the WooCommerce blog{{/link}} for more tips on marketing your store',
+			'woocommerce-admin'
+		),
+		components: {
+			link: (
+				<Link
+					type="external"
+					href="https://woocommerce.com/blog/marketing/coupons/"
+					target="_blank"
+				/>
+			),
+		},
+	} );
 };
 
 export default ReadBlogMessage;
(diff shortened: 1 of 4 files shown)


client/marketing/components/knowledge-base/ReadBlogMessage.js
```diff
@@ -0,0 +1,36 @@
...
+		'Read <a href="https://woocommerce.com/blog/marketing/coupons/">the WooCommerce blog</a> for more tips on marketing your store',
+		'woocommerce-admin'
+	);
+
+	const message3 = _x(
+		' for more tips on marketing your store',
+		'Read <a href="https://woocommerce.com/blog/marketing/coupons/">the WooCommerce blog</a> for more tips on marketing your store',
+		'woocommerce-admin'
+	);
+
```
becdetat, line 24: There's a lot of repetition going on here, you could probably do something like:

```js
return interpolateComponents(
    mixedString: __( 'Read {{link}}the WooCommerce blog{{/link}} for more tips on marketing your store', 'woocommerce-admin' ),
    components: {
        link: <Link href="https://woocommerce.com/blog/marketing/coupons/" />
    }
);
```

See https://github.com/woocommerce/woocommerce-admin/blob/main/client/profile-wizard/steps/benefits/index.js#L259 for more of an example.

becdetat: If you can get the size of this function down you could probably inline it as well to avoid having an extra file.

becdetat:
This works well, I'm just concerned about that `ReadBlogMessage` function. I would also use a `target="_blank"` on the link in the message, otherwise it loads over WordPress.

client/marketing/components/knowledge-base/ReadBlogMessage.js
ecgan: @becdetat , thanks for the heads up! I'll read more about the `interpolateComponents`.

> you could probably inline it as well to avoid having an extra file.

My personal preference is to favor many small files over few large files. Is that the general rule of thumb / preference for the project here? If that's the case, I'll work with that.

ecgan:
@becdetat , I've made some changes based on your comments (which are really helpful! Thank you 🙂 👍 ). Could you review again and approve the PR if they are okay?

client/marketing/components/knowledge-base/ReadBlogMessage.js
becdetat: My opinion is that you can go too far, if something can be written succinctly inline then it is more valuable being co-located where it is used.

If this was something that was going to be reused in multiple places then definitely go with a file per "thing" but in this case it should be inline.

becdetat (approved):
@ecgan 

> Could you review again and approve the PR if they are okay?

Everything looks great, glad that simplification of the content worked out :-D

PR #5455, opened 2020-10-21
Home Screen - migrated orders panel

### LINKED ISSUE #5238: Home Screen - migrate orders panel
**User story:**
As a merchant
I want to easily see how many orders need my immediate attention when I enter WooCommerce
So I can manage them as soon as possible

**Description:** we’re going to migrate the orders panel features to the home screen, as managing orders is one of the most important store management tasks.

**Acceptance criteria:**
- [ ] Display a task in the home screen including up to 5 orders in reverse chronological order

Note: the task title will be different
- [ ] Display a `Manage all orders` CTA that redirects the user to the orders list
- [ ] The order card should include the same elements that we have now: order number, customer name, price, number of products, time, order status and call to action
- [ ] Redirect the user to the respective order if the user clicks in the order number or in the `Open` CTA
- [ ] Redirect the user to the Customer report filtered by the customer name if the user clicks in the customer name 
- [ ] Keep the logic in terms of displaying orders that are "actionable" and the Manage “actionable” order statuses settings at Analytics > Settings
- [ ] Change the copy displayed at Analytics > Settings > Actionable Status (https://d.pr/i/LA5p16) to "Orders with these statuses require action on behalf of the store admin. These orders will show up in the Home Screen - Orders task."

- [ ] The orders activity panel should be removed from the navbar once we have the same feature in the home screen

**Event tracking:**
Can we reuse the same event tracking that we have in the orders panel?

**Additional links:**

octaedro:
Partially fixes [5238](https://github.com/woocommerce/woocommerce-admin/issues/5238)

This PR moves the Orders panel to the `Home` screen. In order to do it, an ActivityPanel component is created in the `Home` screen.
A follow-up PR will be created with the changes and additions necessary to finish up the issue. Those changes are the `Manage all orders` button and the empty activity panel for orders.

### Accessibility

### Screenshots
Screen Capture on 2020-10-22 at 10-43-10

- Go to the `Home` screen.
- Verify the orders panel is visible.
- Confirm that the orders in progress number is right.
- Verify that the order card looks like it should.
- Verify the buttons and links are working well.

### Changelog Note:
Dev: Home Screen - migrate orders panel


client/homescreen/activity-panel/orders/utils.js
```diff
@@ -0,0 +1,45 @@
...
+		per_page: 1, // Core endpoint requires per_page > 0.
+		status: orderStatuses,
+		_fields: [ 'id' ],
+	};
+
+	getItems( 'orders', ordersQuery );
+
+	// Disable eslint rule requiring `latestNote` to be defined below because the next two statements
+	// depend on `getItemsTotalCount` to have been called.
+	// eslint-disable-next-line @wordpress/no-unused-vars-before-return
```
samueljseay, line 35: What do you think about refactoring this a little bit and moving it out of this area? It really should be a selector inside `@woocommerce/data` (in the items store). The only thing that will need to change to accomodate that is to pass in the `orderStatuses` to the selector, that way it's not dependent on the settings store directly to work.

Ideally we would also cover some of this with tests although that will probably require significant mocking right now, so optional imo, but quite important I think to start moving hidden bits of code like this to packages where they belong because it should be reusable.

client/homescreen/activity-panel/panels.js
```diff
@@ -0,0 +1,24 @@
...
+			title: __( 'Orders', 'woocommerce-admin' ),
+			count: countUnreadOrders,
+			initialOpen: true,
+			panel: (
+				<OrdersPanel hasActionableOrders={ countUnreadOrders > 0 } />
+			),
+		},
+		// Add another panel row here
+	];
+	return panels;
```
samueljseay, line 23: Tiny thing but the variable declaration here is not needed. Just return the array directly and we can still add items to it easily.

client/homescreen/activity-panel/style.scss
```diff
@@ -0,0 +1,73 @@
...
+
+	.woocommerce-activity-card {
+		padding: ( $fallback-gutter / 2 ) $gutter;
+		&__header {
+			margin-bottom: $gap-smaller;
+		}
+	}
+}
+
+.components-panel__body-title {
```
samueljseay, line 69: Any concerns with this not being scoped to the panel card? I could see this causing style pollution in other places. I realise that sometimes its necessary but I wanted to check.

client/homescreen/activity-panel/test/index.js
```diff
@@ -0,0 +1,58 @@
...
+				title: 'custom-panel-2',
+				count: 20000,
+				initialOpen: false,
+				panel: <span>Custom panel 2</span>,
+			},
+		] ),
+	};
+} );
+
+// Mock the panels.
```
samueljseay, line 31: This comment should be "mock the orders"?

client/homescreen/activity-panel/test/index.js
```diff
@@ -0,0 +1,58 @@
...
+ */
+import { render, screen } from '@testing-library/react';
+
+/**
+ * Internal dependencies
+ */
+import ActivityPanel from '../';
+
+// Mock the panels.
+jest.mock( '../panels', () => {
```
samueljseay, line 12: This mock reveals a code smell in the implementation of `<ActivityPanel>`. I can see why you're doing it, because if someone adds more rows then the test will fail, but it shows that you're using a private method inside `<ActivityPanel>` to render panels and that they should be a separate component.

Here's what I suggest:

1. Lets move away from the pattern of creating new objects like these panel objects which actually just represents what should be a component. Create a new `<Panel>` component.

A better panel component would be something like:

```
export const Panel = ({title, count, initialOpen, children}) => {

  // ... render the panel here
}
```

Some notes about why:

* every time we make a private method to render children in a component we remove a possible component we can reuse elsewhere. In the implementation of `<ActivityPanel>` as it is now, it cannot be reused and nor can any code used to generate the panels.

* Testing is easier, no mocking needed, test each very simple component separately.

* Note that increasing reuse will over time reduce bundle size because I imagine we're gonna reuse this UI pattern

2. With the new panel component in place you can now have `<ActivityPanel>` render it's children instead. Again making that component more flexible and removing the need for inlining component declarations in objects like we do in `getAllPanels`

3. Once `<ActivityPanel>` renders children we can write ActivityPanel like:

```
<ActivityPanel>
   {theData.map((whatever) => <Panel {...whatever} />)}
</ActivityPanel>
```

Doesn't look like much but in doing so we have decoupled the panel data from `ActivityPanel` (its just UI it doesn't need to know about the *data*.) And we made all the components fully reusable for elsewhere.

client/homescreen/layout.js
```diff
@@ -25,6 +25,7 @@ import '../dashboard/style.scss';
 import TaskListPlaceholder from '../task-list/placeholder';
 import InboxPanel from '../inbox-panel';
 import { WelcomeModal } from './welcome-modal';
+import ActivityPanel from './activity-panel';
```
samueljseay, line 28: Could we move towards always using named exports from now on? Its error prone to use defaults because its very easy to rename something different to what it was exported as and that makes it much harder to find usages in your IDE.

client/homescreen/activity-panel/index.js
```diff
@@ -0,0 +1,59 @@
...
+					</PanelBody>
+				</Card>
+			);
+		} );
+	};
+
+	return <Panel> { renderPanels() } </Panel>;
+};
+
+export default withSelect( ( select ) => {
```
samueljseay, line 55: So i mentioned some code smells with this component in the tests but if you change those things this should also change.

If you want to pass state into `ActivityPanel` you will be able to pass it as props from its parent which is better because we separate the part thats just rendering from the part that should know about state.

samueljseay: One other comment, I think we should consider that any HOC that gets state should use hooks from now on rather than using `compose` and `withSelect`. It makes sense to choose one approach and stick to it for consistency.

samueljseay (changes requested):
It's looking good, but I think we can improve reuse of the components built here. Have left some comments on how we can achieve that but reach out if you have any questions

samueljseay:
@octaedro One more thing thats been mentioned in discussions is that we should promote this new "accordion" component to `@woocommerce/components`. 

I know its a bit of a pain but I'm thinking it might make sense to split off a new PR for that component on its own within the components package and then this PR can follow up by consuming it.

Let me know if you have any issues with that approach. [Relevant discussion here](pbIJXs-ma-p2#comment-861)

octaedro:
Thank you for your review @samueljseay. Could you give this PR another look?

client/header/activity-panel/index.js
```diff
@@ -143,7 +137,7 @@ export class ActivityPanel extends Component {
 			( isEmbedded || location.pathname !== '/' ) &&
 			! isPerformingSetupTask;
 
-		const showOrdersStockAndReviews =
+		const showstockAndReviews =
```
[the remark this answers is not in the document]
octaedro: Nice catch! This was changed in the commit 2dfb29f

client/homescreen/activity-panel/orders/utils.js
octaedro: I like the idea, if it's ok with you I will do this refactor in a follow-up PR.

client/homescreen/activity-panel/panels.js
octaedro: This was changed in the commit 287e3cc

client/homescreen/activity-panel/style.scss
octaedro: Good eyes! I fixed it in the commit 1dbcfec

client/homescreen/activity-panel/test/index.js
octaedro: Fixed in the commit a5dd3fc

client/homescreen/activity-panel/test/index.js
octaedro: This structure was changed in the [PR 5474](https://github.com/woocommerce/woocommerce-admin/pull/5474)

client/homescreen/activity-panel/index.js
octaedro: > One other comment, I think we should consider that any HOC that gets state should use hooks from now on rather than using compose and withSelect. It makes sense to choose one approach and stick to it for consistency.

This was modified in the commit a0c23d2

client/homescreen/activity-panel/index.js
```diff
@@ -0,0 +1,46 @@
...
+				{ panels.map( ( panelData, index ) => {
+					const {
+						className,
+						count,
+						initialOpen,
+						panel,
+						title,
+					} = panelData;
+					return (
+						<AccordionPanel
```
samueljseay, line 32: Yaaas, I don't know about you, but I love this!

client/homescreen/activity-panel/index.js
```diff
@@ -0,0 +1,46 @@
...
+							key={ index }
```
samueljseay, line 33: Not suggesting a change here but I just wanted to note that its *usually* a bad idea to use the index as the key. I think in this case its ok because we don't add/remove/reorder the list of panels, but its worth mentioning that if you were doing that you should have a stable unique ID for each item to use as the key. This video explains it well: https://egghead.io/lessons/react-use-the-key-prop-when-rendering-a-list-with-react

octaedro: Good idea! This was changed in the commit 0d10549

octaedro:
Hi @samueljseay,
I rebased this PR and applied the change you suggested. Could you take a look at this PR again?

jeffstieler (dismissed):
Considering this is targeting `main` and isn't behind a feature flag, I think some visual/design items should be taken care of on this branch.

**Design:**
68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3564383466653865313230316465303030316236616236312f39653961663735622d356661382d343532332d396439652d316163303337346464326539

**Implementd:**
Screen Shot 2020-10-29 at 11 41 51 AM

Based on the design, the title for the card should be "Orders to fulfill" and the count should be to the left of the title. There's also a small visual issue with larger numbers (see 101 in my screenshot).

We should fix the "manage all orders" button before landing this.

jeffstieler:
I'm not sure how you'd like to prioritize this work @samueljseay, but we're going to need to rework the orders fetching logic as well. At a minimum to get rid of the "no orders to fulfill" as the default before even making a request for them:

Capture 2020-10-29 at 12 50 28

The old Orders tab and panel made a few requests - one to power the "unread indicator" and one to retrieve the orders for display. I think that made sense when a user interaction was required to open the panel - while that was 2 requests, it potentially saved work by not retrieving orders before they were needed. The interaction is different now, so we should rework the data retrieval to suit.

client/homescreen/activity-panel/orders/utils.js
octaedro: Hi @samueljseay,
In order to address this fix, I created the commit 558cdd9, could you give it a look?

samueljseay (approved):
Pre approving this, I think you're ready to close this soon? amazing work @octaedro thank you!

🚀

Issue #5335, opened 2020-10-08
New navigation: add support to top extensions


pmcpinto:
**User story:**
As a merchant
I want to access the most important extensions in the new navigation
So I don't need to go back to wp-admin to use a specific extension

**Description:** it will probably take a while until most of the 3pds extensions start supporting the new navigation. It would be good to launch the new navigation already supporting some of the top extensions in our marketplace, so 3PDs can see real examples of how extensions are going to look like in the new navigation and merchants can use those extensions without going to wp-admin.

**Acceptance criteria:**
Add support in the new nav for the following extensions (similar to what we're doing in the Calypso nav):
- [x] WCPay - https://github.com/Automattic/woocommerce-payments/pull/1067
- [x] AutomateWoo - https://github.com/woocommerce/automatewoo/pull/836
- [ ] AutomateWoo Referrals - https://github.com/woocommerce/automatewoo-referrals/pull/100
- [ ] Memberships - https://github.com/woocommerce/woocommerce-memberships/pull/103
- [x] Subscriptions - https://github.com/woocommerce/woocommerce-subscriptions/pull/3868
- [ ] Mailchimp - https://github.com/mailchimp/mc-woocommerce/pull/724
- [ ] Facebook - https://github.com/facebookincubator/facebook-for-woocommerce/pull/1698
- [x] Product addons - https://github.com/woocommerce/woocommerce-product-addons/pull/637
- [x] Taxjar - n/a Shows up under "Settings"
- [ ] Klarna - https://github.com/krokedil/klarna-payments-for-woocommerce/pull/208
- [x] Eway - n/a lives within Settings -> Payments
- [x] Paypal Express - n/a - No menu items
- [x] Stripe - n/a lives within Settings -> Payments
- [x] Payfast - n/a lives within Settings -> Payments
- [x] Australia Post - n/a lives within Settings -> Shipping
- [x] Canada Post - n/a lives within Settings -> Shipping
- [x] Royalmail - n/a lives within Settings -> Shipping
- [x] UPS - n/a lives within Settings -> Shipping
- [x] USPS - n/a lives within Settings -> Shipping
- [x] Square - n/a creates item under Settings
- [ ] WooCommerce Bookings https://github.com/woocommerce/woocommerce-bookings/pull/3001

**Event tracking:**
NA

joshuatf:
Just to double check on this issue @pmcpinto- we're talking about opening PRs in those repos to register support, right?  

[...]

PR #5456, opened 2020-10-21
Set WCA pages as top level in navigation if no parent exists

### LINKED ISSUE #5442: Navigation: WC Admin extended pages no longer included
Pages extending WC Admin should be automatically included in the new Navigation. WC Payments extends WC Admin but can no longer be found in the Nav. This is a regression.

### Steps to reproduce

1. Enable the Nav by setting the `woocommerce_navigation_enabled` option to "yes".
2. Install WooCommerce Payments
3. Visit any WC page.
4. See Payments lacking from the top level menu.
5. Check `wcNavigation.menuItems` and see that Payments is registered:

Screen Shot 2020-10-21 at 2 22 35 PM

Looking at the data, it seems `is_top_level` should be `true` and the parent should be `woocommerce`.\\

### Acceptance Criteria

* WC Admin Extensions, such as WC Payments should be top level items and work correctly

joshuatf:
Fixes #5442 

Sets all WCA-registered pages as top level if no parent was explicitly registered.   Also sets the default order for these items to the end of the list.

Note: this could pretty quickly pollute the top-level of the navigation.  I'm wondering if these shouldn't be put into a category by default unless the category is explicitly stated or if `is_top_level` is explicitly declared.

Also related - https://github.com/woocommerce/woocommerce-admin/issues/5335

### Screenshots

Screen Shot 2020-10-21 at 5 15 51 PM

1. Install WC Pay.
1. Enable the navigation.
1. Check that "Payments" has been added to the end of the primary menu in the navigation.

### DIFF
--- a/src/PageController.php
+++ b/src/PageController.php
@@ -485,8 +485,8 @@ class PageController {
 			'title'        => $options['title'],
 			'capability'   => $options['capability'],
 			'url'          => $options['path'],
-			'order'        => isset( $options['order'] ) ? $options['order'] : 20,
-			'is_top_level' => isset( $options['is_top_level'] ) && $options['is_top_level'],
+			'order'        => isset( $options['order'] ) ? $options['order'] : 100,
+			'is_top_level' => ( isset( $options['is_top_level'] ) && $options['is_top_level'] ) || ! $options['parent'],
 		);
 
 		if ( isset( $options['is_category'] ) && $options['is_category'] ) {


psealock (approved):
>Note: this could pretty quickly pollute the top-level of the navigation. I'm wondering if these shouldn't be put into a category by default unless the category is explicitly stated or if is_top_level is explicitly declared.

I agree with you here. But, not supplying a `parent` and expecting it to be top level is the precedent set already in wc-admin. See WC Pay hooking into wc-admin here:

https://github.com/Automattic/woocommerce-payments/blob/f830f53609acbd742eabd56b03a14ed289d94f7b/includes/admin/class-wc-payments-admin.php#L59-L67

One alternative is that we require a parent of `woocommerce` to be explicitly made to make an item top level. I think we'd rather not change the `wc_admin_register_page` API unless its truly worth it. I'm leaning towards just having the solution proposed here. What do you think?

Testing well and code LGTM 🚢

joshuatf:
>I think we'd rather not change the wc_admin_register_page API unless its truly worth it. I'm leaning towards just having the solution proposed here. What do you think?

I agree with not changing this API, but this could quickly open the doors to many top-level items that should not be in the new navigation.

But since this is mostly a fix of an existing bug and some of the organization around extra items is TBD ( https://github.com/woocommerce/woocommerce-admin/issues/5335 ), I'm good to get this merged and we can follow-up if we need to update the default category.

PR #5457, opened 2020-10-21
Navigation: Include Animation Styles

psealock:
Animation styles from Gutenberg weren't included so the fancy slide effect was missing from Navigation. Since we're bundling wp.components, indirectly used components don't have their CSS included, so we have to add it manually.

1. Turn on navigation by setting the `woocommerce_navigation_enabled` option to "yes".
2. Play around with the Nav going forwards and back to see the animation between menus.


becdetat (approved):
Snappy! Works well. CI is failing but it looks like it just needs another run.

joshuatf (approved):
Nice! ↔️ LGTM here as well.  Restarted those failing tests, though they same to be unrelated to this PR.

PR #5462, opened 2020-10-22
Update dependencies in prep for Composer 2.x.

jeffstieler:
This PR seeks to ready our repository for the upcoming Composer 2.x update.

See: pc2DNy-3j-p2 for more information.

Some packages needed updating to facilitate compatibility with Composer 2, but I experienced errors with the latest Jetpack Autoloader which is why it gets pinned to `2.3.0` in this PR.

- Verify installation works on Composer 1.x:
- `rm -rf vendor && composer install`
- Check that app works (no PHP Fatals missing classes, etc)
- Verify installation works on Composer 2.x:
- `composer self-update --preview`
- `rm -rf vendor && composer install`
- Check that app works (no PHP Fatals missing classes, etc)

### Changelog Note:

N/A

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,9 +40,6 @@
 		}
 	},
 	"autoload": {
-		"classmap": [
-			"includes/"
-		],
 		"psr-4": {
 			"Automattic\\WooCommerce\\Admin\\": "src/"
 		}
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
 	"require": {
 		"composer/installers": "1.9.0",
 		"php": ">=5.6|>=7.0",
-		"automattic/jetpack-autoloader": "2.3.0"
+		"automattic/jetpack-autoloader": "2.2.0"
 	},
 	"require-dev": {
 		"phpunit/phpunit": "7.5.20",
--- a/composer.json
+++ b/composer.json
@@ -8,13 +8,13 @@
 	"prefer-stable": true,
 	"minimum-stability": "dev",
 	"require": {
-		"composer/installers": "1.7.0",
+		"composer/installers": "1.9.0",
 		"php": ">=5.6|>=7.0",
-		"automattic/jetpack-autoloader": "^2.2.0"
+		"automattic/jetpack-autoloader": "2.3.0"
 	},
 	"require-dev": {
 		"phpunit/phpunit": "7.5.20",
-		"woocommerce/woocommerce-sniffs": "0.0.9"
+		"woocommerce/woocommerce-sniffs": "0.1.0"
 	},
 	"config": {
 		"platform": {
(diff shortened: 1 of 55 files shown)


timmyc:
Thanks for picking this up Jeff, I have been seeing the composer 2 update warnings also.

> Jetpack Autoloader which is why it gets pinned to 2.3.0 in this PR

I believe any change to the JP Autoloader we need to coordinate with Proton and Rubik to ensure all Woo Core packages are running on the same version, otherwise 🐉 emerge. So cc @woocommerce/rubik @woocommerce/proton

becdetat (approved):
Aside from @timmyc 's comment everything works well with this 👍

adrianduffell:
This is great @jeffstieler 👍 

What are the next steps to co-ordinate the Jetpack autoloader version bump? I am seeing some Travis jobs fail now because the build is using Composer 2.

```
617 You are using Composer 2, which some of your plugins seem to be incompatible with. Make sure you update your plugins
```
```
620 Your build has been stopped.
```
https://travis-ci.com/github/woocommerce/woocommerce-admin/jobs/407077923

jeffstieler:
@timmyc @adrianduffell it looks like it wasn't necessary to bump the Jetpack Autoloader version, so I've removed that change. This should be good to merge now.

composer.json
```diff
@@ -8,13 +8,13 @@
 	"prefer-stable": true,
 	"minimum-stability": "dev",
 	"require": {
-		"composer/installers": "1.7.0",
+		"composer/installers": "1.9.0",
 		"php": ">=5.6|>=7.0",
-		"automattic/jetpack-autoloader": "^2.2.0"
+		"automattic/jetpack-autoloader": "2.2.0"
```
nerrad, line 13: Instead of pinning these versions, they should be a range. That will help ensure when the package is included in WooCommerce core that there's some flexibility in what actual version gets installed.

nerrad (dismissed):

nerrad:
> What are the next steps to co-ordinate the Jetpack autoloader version bump? I am seeing some Travis jobs fail now because the build is using Composer 2.

The error reported in travis is not because of the jetpack-autoloader bump.

adrianduffell:
> @timmyc @adrianduffell it looks like it wasn't necessary to bump the Jetpack Autoloader version, so I've removed that change. This should be good to merge now.

Sounds good. The recent changes test well for me 🚀 

> The error reported in travis is not because of the Jetpack autoloader version bump.

Agreed, the other changes in this PR should fix the error in Travis.

adrianduffell:
Travis is still blocking the merge because WooCommerce core can't be built. I suggested making a similar change to the core repo here: https://github.com/woocommerce/woocommerce/pull/28096#issuecomment-716928396

octaedro (approved):
Great job @jeffstieler! This is testing well here and code LGTM :shipit:

PR #5466, opened 2020-10-22
Navigation: Adding Tracks by tying into wc-admin page view

### LINKED ISSUE #5332: Navigation: Add Tracks by tying into wc-admin page view
Introduce new parameter `isNavigation` or something to signify the new nav existence on page view Tracks events

joelclimbsthings:
Fixes #5332

Adding the property `has_navigation` to the tracks pageView events to indicate if the new navigation feature is active or not (boolean value).

### Screenshots

image

- Checkout branch
- Run this in your console to view tracks events: `localStorage.setItem( 'debug', 'wc-admin:*' );`
- View an embedded WC page and ensure the `has_navigation` property exists and is `true` within the pageView event, as in the screenshot above.
- Go to a non-embedded page and verify the same flag is present and `true`.
- Disable the new navigation feature, and repeat the last two steps to verify that the value is now `false`.

### DIFF
--- a/client/layout/index.js
+++ b/client/layout/index.js
@@ -78,9 +78,16 @@ class _Layout extends Component {
 			isJetpackConnected,
 		} = this.props;
 
+		const navigationFlag = {
+			has_navigation: !! window.wcAdminFeatures.navigation,
+		};
+
 		if ( isEmbedded ) {
 			const path = document.location.pathname + document.location.search;
-			recordPageView( path, { is_embedded: true } );
+			recordPageView( path, {
+				is_embedded: true,
+				...navigationFlag,
+			} );
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


client/layout/index.js
```diff
@@ -78,9 +78,16 @@ class _Layout extends Component {
 			isJetpackConnected,
 		} = this.props;
 
+		const navigationFlag = {
+			has_navigation: !! window.wcAdminFeatures.navigation,
```
joshuatf, line 82: Definitely an edge case, but it's possible the feature is enabled, but the page we're on has opted out of the navigation experience.  `window.wcNavigation` should show us whether or not the navigation is on for the existing page.

joshuatf (approved):
Great work, Joel!  Love the thorough testing instructions ❤️ 

Left one little nitpicky edge case, but otherwise testing perfectly and pre-approving.

PR #5263, opened 2020-10-02
Add animation to Inbox note deletion

octaedro:
Fixes #5260

This PR animates the Inbox note deletion.

_This PR needs a rebase since it uses changes applied to PR [5252](https://github.com/woocommerce/woocommerce-admin/pull/5252)_

### Screenshots
kZ6UJZKkF7

### Detailed test instructions:

- Verify the note deletion works well on `Home` screen (URL: `/wp-admin/admin.php?page=wc-admin`)
screenshot-one wordpress test-2020 10 01-16_50_13
- Verify the note deletion works well on other screens.
screenshot-one wordpress test-2020 10 01-16_50_51

### Changelog Note:
Dev: Added animation to Inbox note deletion

PR #5471, opened 2020-10-23
Redirect on WCPay inbox note install action on 1.6

### LINKED ISSUE #5412: WCPay inbox note stalls on install action

dechov:
Fixes #5412 **on 1.6.x** by redirecting into WooCommerce Payments setup flow from the relevant Inbox note action.

Addresses https://github.com/woocommerce/woocommerce-admin/pull/5413#issuecomment-714706612:
- https://github.com/woocommerce/woocommerce-admin/commit/bbba670fb221d12024c02e5a8301e23ff919a4a0 **applies the fix in https://github.com/woocommerce/woocommerce-admin/pull/5413** to the `release/1.6.3` base branch
  - Test using same instructions from that PR. ([This code snippet](https://cloudup.com/c7FvOGE0Tve) can be used to trigger the note.)
- https://github.com/woocommerce/woocommerce-admin/commit/3303b8c071922a08e3a30dd39468587758dc8c29 restores the handling for the note action from earlier versions
  - Can test by triggering the note to appear using an earlier version, upgrading to this version, and verifying that the behavior matches https://github.com/woocommerce/woocommerce-admin/issues/5412: WooCommerce Payments is activated but does not proceed to next step.

### Changelog Note:

Fix: Proceed to next step after WCPay installation from Inbox note

### DIFF
--- a/client/header/activity-panel/panels/inbox/index.js
+++ b/client/header/activity-panel/panels/inbox/index.js
@@ -47,14 +47,6 @@ const renderNotes = ( { hasNotes, isBatchUpdating, lastRead, notes } ) => {
 	const notesArray = Object.keys( notes ).map( ( key ) => notes[ key ] );
 
 	return notesArray.map( ( note ) => {
-		if ( note.isUpdating ) {
-			return (
-				<InboxNotePlaceholder
-					className="banner message-is-unread"
-					key={ note.id }
-				/>
-			);
-		}
 		return (
 			<InboxNoteCard
 				key={ note.id }
--- a/src/Notes/WC_Admin_Notes_WooCommerce_Payments.php
+++ b/src/Notes/WC_Admin_Notes_WooCommerce_Payments.php
@@ -37,7 +37,7 @@ class WC_Admin_Notes_WooCommerce_Payments {
 	 * Attach hooks.
 	 */
 	public function __construct() {
-		add_action( 'woocommerce_note_action_install-now', array( $this, 'install' ) );
+		add_action( 'init', array( $this, 'install_on_action' ) );
 		add_action( 'wc-admin-woocommerce-payments_add_note', array( $this, 'add_note' ) );
 	}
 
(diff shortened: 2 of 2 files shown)


jeffstieler:
This tested well, but had a visual glitch for me - mentioning in case there's an issue:

Capture 2020-10-26 at 13 58 45

dechov:
@jeffstieler Seems like the whole Inbox is reloading? Looked a bit but couldn't figure out why it would be reloading at that point. Tested again and it doesn't happen for me – it works just like the screen capture in https://github.com/woocommerce/woocommerce-admin/pull/5413.

timmyc:
I'm seeing the same behavior as @jeffstieler above. My test env is - brand new JN site running 4.6.1 + this branch installed as a plugin.

I've verified that: 

- The note doesn't display immediately ( as expected ) but once I `wp option update woocommerce_admin_install_timestamp 1601510400` to a date over a week in the past, and run the daily job, the note appears.
- The more information action works as expected
- When clicking Get Started, I'm seeing the same behavior, the action fires, then the inbox is re-painted, and then I'm redirected to the KYC flow on Stripe.

So it seems like this is working @dechov with the odd refresh which looses the status indicator. Are you okay with shipping with that known caveat?

dechov:
I'm not sure why it doesn't do that for me on the "Get started" button – maybe the redirect is aborting the AJAX request or something like that – but the above behavior _is_ happening for me on the legacy "Install now" action.

In any case, I see why it does that now, and – my mistake – it is just the _note_ that is updated, not the whole Inbox.  In 1.6, the `InboxNoteCard` is replaced with a placeholder while the note is updating [[code](https://github.com/woocommerce/woocommerce-admin/blob/f172b70fd5b11588655a0dc505c7096583219826/client/header/activity-panel/panels/inbox/index.js#L49-L65)], unlike in `main` [[code](https://github.com/woocommerce/woocommerce-admin/blob/38ceacedfd30d736b91132af56a6974fbe26d21d/client/inbox-panel/index.js#L52-L70)] where it can retain the child element's state.

Not sure what a good fix on short notice would be, except maybe to just remove that placeholder case?

LevinMedia:
>So it seems like this is working @dechov with the odd refresh which looses the status indicator. Are you okay with shipping with that known caveat?

@timmyc @dechov My main concern here is if there's no indication that something is happening that we might loose people because they refresh, or do something else because they think the action they just took has failed to complete.

dechov:
> maybe to just remove that placeholder case?

Done in https://github.com/woocommerce/woocommerce-admin/pull/5471/commits/c8d8b062c78f192d23c1a67daa49527aabd2ca0c, consistent with a change in https://github.com/woocommerce/woocommerce-admin/pull/5263 on the `main` branch.

timmyc:
Thanks for the update @dechov - testing now.

timmyc (approved):
Thanks for the fix on the loading state @dechov this is testing out well for me now using the same steps I noted above.

PR #5473, opened 2020-10-26
Fix PSR-4 Classes

jeffstieler:
Fixes PHP Fatal errors due to missing classes when updating to the `2.4.x` Jetpack autoloader.

Version `2.4.0` strictly enforces PSR-4 conformity, which our classes were failing. This PR seeks to update the class names so that they properly implement PSR-4, specifically that the class name matches the file name.

- `rm -rf vendor && composer install`
- Check that app works (no PHP Fatals missing classes, etc)

### DIFF
--- a/src/DeprecatedClassFacade.php
+++ b/src/DeprecatedClassFacade.php
@@ -47,7 +47,7 @@ class DeprecatedClassFacade {
 	 */
 	public function __call( $name, $arguments ) {
 		_deprecated_function(
-			static::$facade_over_classname . '::' . $name,
+			static::class . '::' . $name,
 			static::$deprecated_in_version
 		);
 		return call_user_func_array(
@@ -67,7 +67,7 @@ class DeprecatedClassFacade {
 	 */
 	public static function __callStatic( $name, $arguments ) {
 		_deprecated_function(
-			static::$facade_over_classname . '::' . $name,
+			static::class . '::' . $name,
 			static::$deprecated_in_version
 		);
 		return call_user_func_array(
--- a/composer.json
+++ b/composer.json
@@ -40,9 +40,6 @@
 		}
 	},
 	"autoload": {
-		"classmap": [
-			"includes/"
-		],
 		"psr-4": {
 			"Automattic\\WooCommerce\\Admin\\": "src/"
 		}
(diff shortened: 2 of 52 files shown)


nerrad:
I didn't review the code, I just did a quick test to verify there were no fatals with the plugin active and it looks good on that front.

However I do get a couple deprecated notices (might be known already, just documenting in case it's not):

> Deprecated: Automattic\WooCommerce\Admin\Notes\WC_Admin_Notes_Deactivate_Plugin::delete_note is deprecated since version 1.6.0 with no alternative available. in /srv/www/woo/public_html/wp-includes/functions.php on line 4846
> Deprecated: Automattic\WooCommerce\Admin\Notes\WC_Admin_Note::get_status is deprecated since version 1.6.0 with no alternative available. in /srv/www/woo/public_html/wp-includes/functions.php on line 4846

mattsherman:
I get the same deprecation note.

src/DeprecatedClassFacade.php
```diff
@@ -67,7 +67,7 @@ public function __call( $name, $arguments ) {
 	 */
 	public static function __callStatic( $name, $arguments ) {
 		_deprecated_function(
-			static::$facade_over_classname . '::' . $name,
+			static::class . '::' . $name,
```
mattsherman, line 70: What is the difference between `get_class( $this )` and `static::class`?

mattsherman (commented):
Poking around on wc-admin screens, things appear to be working correctly. No fatals or errors were encountered.

Left one comment related to getting the class name in `DeprecatedClassFacade`. Once that is addressed, I'll either review changes or approve this PR as-is.

src/DeprecatedClassFacade.php
jeffstieler: There won't be an instance (`$this`) in static calls.

mattsherman: Right, of course. 🤦‍♂️ Can `static::class` be used in line 50 instead of `get_class( $this )` though?

jeffstieler: Yep! Looks like it can, commit incoming.

jeffstieler: Fixed in 0202fe8974c0bcf47cccde3f1116c70883b7d341.

mattsherman (approved):
Code changes look good and running through screens I didn't encounter any issues. Nice work.

Issue #5245, opened 2020-10-01
Home Screen - store management extensibility


pmcpinto:
**User story:**
As a 3PD
I want to easily extend the store management section
So I can add useful shortcuts to my extension

**Description:** our goal is to allow 3PDs to add store management shortcuts to their extensions

**Acceptance criteria:**
~~We have two options:~~

Let's go with the easiest path

- [ ] Allow them to add shortcuts to a new store management category called "Extensions"
~~- [ ] Allow them to add shortcuts to the existing categories (eg Settings). This will also require to implement a predefined list of categories so they're not limited to the existing categories~~

~~There's also a 3rd option, which is allowing them to create categories, but this one can easily get out of control. IE creating categories with different names for similar use cases.~~
**Event tracking:**
NA

**Additional links:**
[P2 with links to the Figma file](pbIJXs-ma-p2)

samueljseay:
@pmcpinto currently our fixed list of store management links does include an external link. What do you think about *not* allowing external links here? Taking people away from their store to somewhere else is sometimes distracting, not always wanted and a way to spam users with unsolicited links, so my feeling is that we should not allow it.

samueljseay:
As for what option we should choose I think we should lock in this one: " Allow them to add shortcuts to a new store management category called "Extensions"

Reasons:

1. Its simpler to implement
2. having a predefined set of categories locks us in future if we ever want to remove some then some store management links of 3dps will stop working, if we have a fixed category we don't need to worry about backwards compatibility as much.
3. I do however think we should design our API for this in a way that would allow to include a custom category feature in future.

pmcpinto:
@samueljseay here's my feedback:

> @pmcpinto currently our fixed list of store management links does include an external link. What do you think about not allowing external links here? Taking people away from their store to somewhere else is sometimes distracting, not always wanted and a way to spam users with unsolicited links, so my feeling is that we should not allow it.

Yep that seems to be a great move to avoid folks from abusing from this area.

> As for what option we should choose I think we should lock in this one: " Allow them to add shortcuts to a new store management category called "Extensions"

Thanks for outlining your reasons 💯 . They all make sense to me, so let's keep them under "Extensions". In the future we can evaluate what you mentioned in the 3rd point, once we ship this and get feedback from the community.

pmcpinto:
@samueljseay I think it could be useful to say something about extending the store management section in the post about the task list extensibility. What do you think?

samueljseay:
[...]

PR #5476, opened 2020-10-27
Reorganize the store management link section and provide extensibility

### LINKED ISSUE #5244: Home Screen - reorganize Store Management section
**Description:** we're going to reorganize the store management areas in different categories, making it easier for users to find the shortcut that they want to access

**Acceptance criteria:**
- [ ] The Marketing & Merchandising category is going to include the following shortcuts: Marketing, Add Products, Personalize my Store, View my Store
- [ ] The Settings category is going to include the following shortcuts: Store Details, Payments, Tax, Shipping
- [ ] The Get Support shortcut is going to be included in the Help panel that's going to be added to the home screen
- [ ] In the new single-column layout, the Store management card should expand to 2 columns.

**Event tracking:**
NA

**Additional links:**

samueljseay:
Fixes #5244 and #5245

This is a refactor of the store management links to reflect the new designs and also provides extensibility, allowing developers to add their own links to this section under an "Extensions" category.

To use the new filter 3PDs will be able to call `addFilter` with the new filter `'woocommerce_admin_homescreen_quicklinks'`. Note that we have explicitly chosen to disable links that link outside of the current host.

Originally the store management links used the `@woocommerce/list` component, but this was no longer suitable because it would have required major redesign of that component and much of that component's functionality didn't feel necessary for this.

Note that the new components I've implemented could be easily promoted to the `@woocommerce/components` package later if we decide they are more broadly useful, but I'd suggest we let this component mature and let the new designs mature before promoting it.

### Accessibility

- [x] I've tested using a screen reader (Voice over)

### Screenshots

There are 2 display modes.

in 2 column mode:

Screenshot 2020-10-27 at 7 39 47 PM

in 1 column mode:

Screenshot 2020-10-27 at 7 40 03 PM

#### Visual/behaviour
1. On a site with the task list completed visit the homescreen
2. Try out all the links provided. 
3. Check that tracking works.
4. Visual inspection against the designs. See pbIJXs-ma-p2
5. Try the different display modes using the new activity panel item "Display"

#### Extensibility

Here's a code snippet for testing extensibility, place it anywhere in our JS code that runs before the store management component is rendered:

```
addFilter(
		'woocommerce_admin_homescreen_quicklinks',
		'my-cool-extension',
		( quickLinks ) => {
			return [
				...quickLinks,
				{
					title: __( 'My Cool link', 'woocommerce-admin' ),
					href: 'foo/bar',
					icon: <div>Hello World</div>,
				},
			];
		}
	);
```

Note that if you want to try a real icon, then you'll need to import it from `@wordpress/icons`. If you'd like to to test the filtering of external links then change `href`'s value to be `https://example.com`.

### Changelog Note:

Dev: Rearrange the store management links under categories and expose extensibility via the `woocommerce_admin_homescreen_quicklinks` filter.


client/stylesheets/shared/_gutenberg-components.scss
```diff
@@ -22,3 +22,7 @@ allows Woo themed components based on the config found in postcss.config.js
 @import 'gutenberg-components/tab-panel/style.scss';
 @import 'gutenberg-components/guide/style.scss';
 @import 'gutenberg-components/animate/style.scss';
+
+:root {
```
samueljseay, line 26: Note that this change is borrowed from another PR that fixes issues with wp components. I can remove this before merge, but its going to be easier to test visuals with it.

client/store-management-links/index.js
```diff
@@ -0,0 +1,224 @@
...
+							title={ category.title }
+						>
+							{ category.items.map(
+								( { icon, listItemTag, title, link } ) => (
+									<QuickLink
+										icon={ icon }
+										key={ listItemTag }
+										title={ title }
+										isExternal={
+											link.linkType === 'external'
```
jeffstieler, line 204: Is there a future use for `linkType`, or is it just to determine the `isExternal` prop? If the latter, what do you think about simplifying this to just be `isExternal` on the link objects instead?

client/store-management-links/quick-link-category/index.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * External dependencies
+ */
+import React from '@wordpress/element';
```
jeffstieler, line 4: Do we need to import `React` here?

client/store-management-links/quick-link-category/test/index.js
```diff
@@ -0,0 +1,22 @@
...
+import { render } from '@testing-library/react';
+import '@testing-library/jest-dom';
+
+/**
+ * Internal dependencies
+ */
+import { QuickLinkCategory } from '..';
+
+describe( 'QuickLinkCategory', () => {
+	it( 'displays the passed title and children', () => {
```
jeffstieler, line 13: I was expecting to see an assertion for the title being displayed. Did that get missed?

client/store-management-links/quick-link-category/test/index.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * External dependencies
+ */
+import { render } from '@testing-library/react';
+import '@testing-library/jest-dom';
```
jeffstieler, line 5: This should be necessary as it's included in [`tests/js/setup-react-testing-library`](https://github.com/woocommerce/woocommerce-admin/blob/1e0061d0df8c177795c58ab9603af95f4aef6610/tests/js/setup-react-testing-library.js#L5).

client/store-management-links/quick-link/index.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * External dependencies
+ */
+import React from '@wordpress/element';
```
jeffstieler, line 4: Same question about necessity.

client/store-management-links/quick-link/index.js
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Internal dependencies
+ */
+import './style.scss';
+
+export const QuickLink = ( { icon, title, href, isExternal, onClick } ) => {
+	return (
+		<div className="woocommerce-quick-links__item">
+			<a
+				onClick={ ( e ) => onClick( e ) }
```
jeffstieler, line 17: Could we just use `onClick` here instead of defining a closure?

client/store-management-links/quick-link/test/index.js
```diff
@@ -0,0 +1,50 @@
+/**
+ * External dependencies
+ */
+import React from '@wordpress/element';
```
jeffstieler, line 4: Needed?

client/store-management-links/quick-link/test/index.js
```diff
@@ -0,0 +1,50 @@
...
+import { render } from '@testing-library/react';
+import { brush } from '@wordpress/icons';
+import '@testing-library/jest-dom';
```
jeffstieler, line 7: Shouldn't be needed.

client/store-management-links/quick-link/test/index.js
```diff
@@ -0,0 +1,50 @@
...
+import userEvent from '@testing-library/user-event';
+
+/**
+ * Internal dependencies
+ */
+import { QuickLink } from '../index';
+
+describe( 'QuickLink', () => {
+	it( 'renders an title, href and icon based on props passed', () => {
```
jeffstieler, line 16: Is this missing the assertion for the icon?

client/store-management-links/quick-link/index.js
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Internal dependencies
+ */
+import './style.scss';
+
+export const QuickLink = ( { icon, title, href, isExternal, onClick } ) => {
+	return (
+		<div className="woocommerce-quick-links__item">
+			<a
```
jeffstieler, line 16: Should this use the `Link` component from `@woocommerce/components`? That would allow for navigating to other wc-admin powered pages without a full page load (if passed the `type` prop). Alternate idea: should we just update the `Link` component to be this `QuickLink` implementation?

jeffstieler (changes requested):
Code review aside, this is testing well but with some console warnings.

client/store-management-links/index.js
```diff
@@ -0,0 +1,224 @@
...
+		// We do not support extension links that take users away from the host.
+		if ( url.origin === window.location.origin ) {
+			acc.push( {
+				icon,
+				link: {
+					href,
+					linkType: 'extension',
+				},
+				title,
+				listItemTag: 'quick-links-extension-link',
```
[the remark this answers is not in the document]
samueljseay: oooh good catch. 🤔 I'll refactor this.

client/store-management-links/index.js
samueljseay: @jeffstieler I went back and forth on this and almost removed it. Your validation makes me think we should remove it. Will do.

client/store-management-links/quick-link-category/index.js
samueljseay: @jeffstieler Yeah, you always have to import React when using jsx.

Soon we should be able to remove this though; https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#how-to-upgrade-to-the-new-jsx-transform

client/store-management-links/quick-link-category/test/index.js
samueljseay: I missed that, thanks!

client/store-management-links/quick-link-category/test/index.js
samueljseay: It sure did! Thanks for spotting that.

client/store-management-links/quick-link/index.js
samueljseay: Yeah you gotta import React to use jsx

client/store-management-links/quick-link/index.js
samueljseay: @jeffstieler thats a good point. My main concern would be styling but we could probably override it with css. 🤔 

I had forgotten that `Link` worked in that way so it *could* make sense to use that here. I'll investigate this.

client/store-management-links/quick-link/index.js
samueljseay: :D yep good spotting

client/store-management-links/quick-link/test/index.js
samueljseay: I'll remove this.

client/store-management-links/quick-link/test/index.js
samueljseay: Oh yes it is, I struggled with doing assertions on the icon svg for some reason. I'll remove that from the assertion

client/homescreen/test/index.js
```diff
@@ -21,7 +21,9 @@ jest.mock( 'task-list', () => jest.fn().mockReturnValue( '[TaskList]' ) );
 jest.mock( 'inbox-panel', () => jest.fn().mockReturnValue( '[InboxPanel]' ) );
 
 // We aren't testing the <QuickLinks /> component here.
-jest.mock( 'quick-links', () => jest.fn().mockReturnValue( '[QuickLinks]' ) );
+jest.mock( 'quick-links', () => ( {
```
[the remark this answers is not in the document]
samueljseay: Good spotting I missed removing this.

client/store-management-links/index.js
```diff
@@ -0,0 +1,224 @@
...
+										title={ title }
+										isExternal={
+											link.linkType === 'external'
+										}
+										href={ link.href }
+										onClick={ () => {
+											recordEvent(
+												'home_quick_links_click',
+												{
+													task_name: listItemTag,
```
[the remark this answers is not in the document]
samueljseay: @jeffstieler yep `task_name` is correct, this track was originally happening in the QuickLinks component heres the OG code:

```
props.recordEvent( 'home_quick_links_click', {
		task_name: listItemTag,
} );
```

client/store-management-links/index.js
samueljseay: Removed in favor of `isExternal` here: https://github.com/woocommerce/woocommerce-admin/pull/5476/commits/b96c0605b75369f792246b9b1755c5c9a7265259

client/homescreen/test/index.js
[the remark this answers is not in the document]
samueljseay: Removed in https://github.com/woocommerce/woocommerce-admin/pull/5476/commits/a25ec4fedf6ec7555a9461b22c798f2fdfea161d

client/store-management-links/index.js
samueljseay: I have opted to go with the `Link` component you suggested in your other comment so have kept the linkType. Changes here: https://github.com/woocommerce/woocommerce-admin/pull/5476/commits/24e2171719d231dcd39305af5ff443a891f509ea

client/store-management-links/quick-link/index.js
samueljseay: This is resolved now when passing to the `Link` component in https://github.com/woocommerce/woocommerce-admin/pull/5476/commits/24e2171719d231dcd39305af5ff443a891f509ea

client/store-management-links/quick-link/test/index.js
samueljseay: Removed the assertion. I still haven't had luck querying svgs by role or any other attribute in jest-dom.

client/store-management-links/quick-link/index.js
samueljseay: I've gone with the `Link` component here and it works well. 🎉 thanks this was a definite improvement. https://github.com/woocommerce/woocommerce-admin/pull/5476/commits/24e2171719d231dcd39305af5ff443a891f509ea

client/store-management-links/index.js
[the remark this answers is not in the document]
samueljseay: @jeffstieler This should be fixed in https://github.com/woocommerce/woocommerce-admin/pull/5476/commits/90de31dab679d216c2f742f38d2dffd3559d88fa I've opted to create a stable key for the item based on a combo of title/tag/href and the only way it'll collide is if devs provide duplicate links which probably isn't a bad thing for them to be warned about.

client/stylesheets/shared/_gutenberg-components.scss
samueljseay: Just a note, I'm going to remove this now and remerge `main` because we have a fix for this.

client/store-management-links/quick-link/index.js
jeffstieler: Curious.. most all of our other JSX files don't import React.

client/store-management-links/quick-link/index.js
```diff
@@ -0,0 +1,39 @@
...
+ * Internal dependencies
+ */
+import './style.scss';
+
+export const QuickLink = ( { icon, title, href, linkType, onClick } ) => {
+	const isExternal = linkType === 'external';
+
+	return (
+		<div className="woocommerce-quick-links__item">
+			<Link
```
jeffstieler, line 19: Why isn't the `linkType` prop being passed to `<Link>`?

jeffstieler (approved):
The code changes look good for the most part, but I do have a question about the `linkType` prop omission from the `<Link>` component (it defaults to `wc-admin` which really only fits for one quick link at the moment). Also - perhaps for a follow up PR - are we going to support the new "extension" `linkType` or perhaps let extensions choose their own `linkType`?

Pre-approving, assuming there's agreement on the `linkType` prop being passed to `<Link>`s.

client/store-management-links/quick-link/index.js
samueljseay: Thanks for spotting this, just an error of omission. Will fix.

jeffstieler: I think for the links we're creating we should explicitly pass a `linkType` of "wp-admin" or "wc-admin". I doubt there's much to break with the regex method for detecting if an external/extension link could actually be a "wc-admin" type, but out of caution how about we force all external/extension links to be of type "wp-admin"?

samueljseay: Sounds good. I'll ship that

samueljseay: @jeffstieler I had another think about this and I want to break down my reasoning about the approach I'll take here based on your feedback.

There are several link types generated by the original quick links code:

`wc-settings`, `wc-admin`, `wp-admin` and `external`.

There are fewer link types respected by the `<Link>` component:

`wp-admin`, `wc-admin`, `external`

I think the right choice here is to ignore the fact that `wp-admin` and `wc-admin` links are not functionally different. Rather we should respect the API in case someone else decides that perhaps external links should behave differently for example.

So steps I'll take are:

1. Remove the `wc-settings` type from existing code. Replace it with `wp-admin` type
2. Normalise all `extension` links to `wp-admin` type
3. For remaining `wc-admin` links I will keep them as `wc-admin` type and pass the link type down to `<Link>`

I think this is similar to what you suggested but your suggestions were a little ambiguous so I've spelled it out here.

samueljseay: @jeffstieler I had another think about this and I want to break down my reasoning about the approach I'll take here based on your feedback.

The link types we normalise to pass down to `<QuickLink>` right now include:

`wc-admin`, `wp-admin`, `external` and `extension`

There are fewer link types respected by the `<Link>` component:

`wp-admin`, `wc-admin`, `external`

I think the right choice here is to ignore the fact that `wp-admin` and `wc-admin` links are not functionally different. Rather we should respect the API in case someone else decides that perhaps external links should behave differently for example.

So steps I'll take are:

1. Remove the `extension` type from existing code. Replace it with `wp-admin` type
2. For remaining `wc-admin` links I will keep them as `wc-admin` type and pass the link type down to `<Link>`

I think this is similar to what you suggested but it felt a little ambiguous so I've spelled it out here.

samueljseay: I've addressed this here: https://github.com/woocommerce/woocommerce-admin/pull/5476/commits/5b5af412fa8b47224f8d804cf1e3289f0b584a40

I'm going to merge this based on your pre-approval, but if you see any problems with the approach I've taken here then please let me know and I can address it in a follow up PR.

PR #5495, opened 2020-10-28
Prevent double fill of Customers nav slot

### LINKED ISSUE #5493: Customers navigation item is double registered
**Describe the bug**
The customers link in WCA is being registered twice in the navigation.

**To Reproduce**

1. Enable the nav feature.
1. Go to the "Analytics" menu.
1. Note 2 "Customers" items are shown.

**Expected behavior**
Expected a single nav item for customers.

**Screenshots**
Screen Shot 2020-10-28 at 3 27 00 PM

joshuatf:
Fixes #5493 

The "Customers" nav item is registered twice in WCA and the slot was being filled twice because of this.  This ignore the `/analytics/customers` one in favor the `/customers` path since it won't reside under the `Analytics` menu in the new navigation.

### Screenshots
Screen Shot 2020-10-28 at 3 49 15 PM

1. Enable the navigation.
1. Make sure that 2 "Customers" items aren't shown.
1. Make sure it continues to work as expected.

### DIFF
--- a/client/analytics/report/get-reports.js
+++ b/client/analytics/report/get-reports.js
@@ -107,7 +107,7 @@ export default () => {
 			report: 'customers',
 			title: __( 'Customers', 'woocommerce-admin' ),
 			component: CustomersReport,
-			id: 'woocommerce-analytics-customers',
+			id: null,
 		},
 		{
 			report: 'downloads',


client/analytics/report/get-reports.js
```diff
@@ -107,7 +107,7 @@ export default () => {
 			report: 'customers',
 			title: __( 'Customers', 'woocommerce-admin' ),
 			component: CustomersReport,
-			id: 'woocommerce-analytics-customers',
+			id: null,
```
psealock, line 110: Weird, so this `id` is purely for slotFill. I'm confused why removing it would prevent a double nav item.

joshuatf: Since WCA registers 2 "Customers" menu items, both with the same ID, they fill the same slot.  By default a slot can be filled any number of times.

It may be worth considering limiting this to a single fill, but in this case I think we should favor the menu item with `/customers` instead of the one with a path of `/analytics/customers` due to its location in the new nav.

psealock (approved):
Thanks @joshuatf , works and LGTM 🚢

PR #5497, opened 2020-10-28
Navigation: Applying styles & behavior for folded menu, including mobile

### LINKED ISSUE #5326: Navigation: Match styles including mobile
Figma files: https://www.figma.com/proto/QNNc834dBBNz1bL2qc9VZyq5/Navigation-experimental?node-id=2096%3A133298&viewport=-13770%2C-6863%2C0.7375516295433044&scaling=scale-down

joelclimbsthings:
Fixes #5326 

Updating styles and behaviour governing new navigation feature, including mobile. This should align with the following acceptance criteria: 

- A closed menu will appear as a square with the site logo of size equal to the wc-admin header.
- With the navigation closed, wc-admin or embedded pages occupy the entire screen width.
- On screens `$break-large: 960px` or below toggling the navigation pushes or overlaps the content, whichever is better.
- At the same breakpoint as above, the Navigation is collapsed by default.

I've also added a listener on the `orientationchange` event to handle the case of the layout for a device switching from landscape to portrait, in which case it will hide the navigation. This change is probably debatable usability wise, so I'm glad to remove it if desired.

### Screenshots
**Folded Before:** 
Screen Shot 2020-10-23 at 3 36 04 PM

**Folded After:**
image

**Open on Desktop:**
image

**Open on Mobile:**
image

- Navigate to WooCommerce pages with navigation feature on.
- Click logo in top-left to fold navigation.
- Ensure your screen width is greater than 960px.
- Observe that it now appears as the "Folded After" screenshot, and does **not** overlap content.
- Shrink viewport to less than 960px.
- With the navigation open, it should appear the same but now overlay the content.
- Refresh the page, and ensure that the  navigation collapses by default (while still with a small viewport).
- You can use Chrome or Firefox dev tools to test a `orientationchange` event to ensure that the navigation collapses when switching to a view smaller than `960px`.


client/navigation/stylesheets/index.scss
```diff
@@ -17,28 +17,53 @@
...
 
+	#woocommerce-embedded-navigation {
+		position: fixed;
+		top: 0;
+		left: 0;
+		width: $navigation-width;
+		height: 100%;
+		box-sizing: border-box;
+		background-color: $gray-900;
+		z-index: 1100;
```
joshuatf, line 28: Out of curiosity, what was the reason behind picking `1100`?  If it's higher than another, a comment explaining why it should be > n may be helpful to someone working on this in the future.

client/navigation/components/header/index.js
```diff
@@ -15,6 +16,22 @@ const Header = () => {
...
+			document.body.classList.add( 'is-folded' );
+		}
+	};
+
+	useEffect( () => {
+		foldOnMobile( document.body.clientWidth );
+
+		window.addEventListener(
+			'orientationchange',
+			( e ) => foldOnMobile( e.target.screen.availWidth ),
```
joshuatf, line 30: Should this use `width` instead of `availWidth`?

client/navigation/components/header/index.js
```diff
@@ -15,6 +16,22 @@ const Header = () => {
...
+		if ( screenWidth <= 960 ) {
+			document.body.classList.add( 'is-folded' );
+		}
+	};
+
+	useEffect( () => {
+		foldOnMobile( document.body.clientWidth );
+
+		window.addEventListener(
+			'orientationchange',
```
joshuatf, line 29: Should this happen on resize as well?  The way the current WP menu changes also occurs on resize, but I can see a case being made for that not happening.

joshuatf (commented):
Nice work, Joel!  This is testing well for me, just left a few questions.

Also curious if this is the expected mobile interaction you described @jameskoster.  You mentioned "sliding"- should this animate at all from folded to unfolded?

client/navigation/components/header/index.js
psealock: `document.body.clientWidth` and `screen.availWidth` give me differing results if I have Chrome console open. Should we stick with one or the other?

psealock (commented):
Thanks @joelclimbsthings, this is testing nicely for me. I added a quick comment on top of Josh's and I did notice one small thing.

When you scroll up, the header bar gets a drop shadow. Can you add it to the Navigation button as well?

Screen Shot 2020-10-29 at 3 09 23 PM

psealock:
One more thing I noticed: It would be great to prevent the animation on mount so it doesn't happen when folding and unfolding the menu. That might be good for a follow up because it could involve a Gutenberg PR.

joelclimbsthings:
> When you scroll up, the header bar gets a drop shadow. Can you add it to the Navigation button as well?

Hey @psealock , I've just made a commit to address this. It's slightly more complex than expected since this shadow is triggered via  class that can't be referenced by our relevant styles. I ended up refactoring the logic that attaches the `is-scrolled` class that creates this shadow and implementing it on our component as well. 

I didn't see any other reusuable hooks in the project, so just let me know if another approach is preferred. It's possible that a more performant solution would be possible (since this one actually attaches two scroll events I believe), but these components don't have a common ancestor or anything to attach a `Context`. No redux, of course, but perhaps `@wordpress/data` can be used for something like this? Haven't dived into that particular rabbit hole just yet.

Screenshot:
image

Also, I actually found a bug that was stemming from the new navigation. The shadow is actually supposed to disappear when you scroll back to the top, but this had been broken. Now fixed as well.

client/navigation/components/header/index.js
psealock: I'm ok with leaving out the resize eventListener. It feels like something only devs do when doing code review 😄

client/hooks/useIsScrolled.js
```diff
@@ -0,0 +1,29 @@
...
+	useEffect( () => {
+		const updateIsScrolled = () => {
+			setIsScrolled( window.pageYOffset > 20 );
+		};
+
+		const scrollListener = () => {
+			rafHandle.current = window.requestAnimationFrame(
+				updateIsScrolled
+			);
+		};
```
psealock, line 18: I'm not sure whats going on, but the contents of the `scrollListener` don't get called on scroll so its not updating on scroll. The `is-scrolled` class is working well if I apply it manually though.

psealock (commented):
Nice catch on the scroll issue @joelclimbsthings. Something isn't testing quite right for me because the `is-scrolled` class never gets applied.

> It's possible that a more performant solution would be possible (since this one actually attaches two scroll events I believe)

The only thing I can think of is some kind of cache or flag to keep track if the listener has already been added to the window. I do see two `scroll` listeners added as a result of `useIsScrolled` in dev tools. It might be good to merge this and see if a better way exists in a follow up if you'd like.

client/hooks/useIsScrolled.js
joelclimbsthings: Strange, it was working reliably for me on Friday. I'll take another look. You're just Firefox or Chrome?

joelclimbsthings: Yeah, can't reproduce this issue @psealock . Which page/etc are you on?

client/navigation/stylesheets/index.scss
```diff
@@ -17,28 +17,59 @@
...
+			margin-left: $header-height;
+		}
+
+		#woocommerce-embedded-navigation {
+			width: $header-height;
+			height: $header-height;
+
+			.woocommerce-navigation-header {
+				&.is-scrolled {
+					box-shadow: 0 8px 8px 0 rgba(85, 93, 102, 0.3);
```
joshuatf, line 58: A variable for this might be good so we don't get out of sync with the header should it ever change.

joshuatf (commented):
Testing well for me.  The `is-scrolled` class did get added, so this looks good on my end unless @psealock is able to reproduce this issue.

client/hooks/useIsScrolled.js
psealock: Ok, it appears to be related to this error being displayed:

Screen Shot 2020-11-03 at 9 41 14 AM

When I comment out `define( 'WP_DEBUG', true );` to turn that off, it works great. I think the header uses some position CSS to go over the error message, but I have no idea why that would interfere with the scroll eventListener not being called.

client/navigation/stylesheets/index.scss
joelclimbsthings: Good call. Done :+1:

client/hooks/useIsScrolled.js
joelclimbsthings: Ah, right. Not sure why it's conflict here either, but I've had my `WP_DEBUG` set to false since my page is overrun with deprecation warnings otherwise.

joelclimbsthings:
I believe I've addressed all of the feedback at this point, with the potential of a couple follow-up issues regarding performance and a possible GB PR.

psealock (approved):
LGTM, although that is weird about the conflict with the warning. Maybe there is a fix to look into after this.

Thanks @joelclimbsthings

PR #5499, opened 2020-10-29
Inbox: Remove Store Setup Alert

### LINKED ISSUE #5460: Inbox: Remove Store Setup Alert
This was a request that came up during the final release testing of 1.6 from @pmcpinto - Now that we have the Activity Panel Store Setup link, this old Store Alert style notification feels quite redundant:

image

This fix might require two different PRs, one using `release/1.6.2` as a base branch, and another against `main` since notes have diverged between those two branches.

Tasks:

- [x] Remove `OnboardingProfiler` [note](https://github.com/woocommerce/woocommerce-admin/blob/6a59bc85703b784aeb0da25e264142fbb278e6f6/src/Notes/OnboardingProfiler.php)
- [x] In `main` we should also remove the related [deprecation facade](https://github.com/woocommerce/woocommerce-admin/blob/main/src/Notes/DeprecatedNotes.php#L545)
- [x] add entry to [delete_obsolete_notes](https://github.com/woocommerce/woocommerce-admin/blob/main/src/Install.php#L469)

mattsherman:
Fixes #5460 

This PR removes the "Welcome to WooCommerce! Set up your store and start selling" alert, as it has be superseded functionality-wise by the Activity Panel Store Setup link.

- Run `main` branch

- Update DB to ensure that the store setup alert is visible

```
UPDATE wp_wc_admin_notes
SET status = 'unactioned'
WHERE name = 'wc-admin-onboarding-profiler-reminder'
```

- Verify store setup alert is visible

- Checkout and run branch

- With WP-CLI, update the `woocommerce_admin_version` option so that the WooCommerce Admin plugin thinks it needs to upgrade and runs the upgrade logic (which will delete the obsolete notes)

```
wp option update woocommerce_admin_version 0.1
```

- Verify store setup alert is no longer visible

### Changelog Note:

Tweak: Remove store setup alert.


becdetat:
This works well and the code looks good. This doesn't remove the note for existing stores with that notice though - you can delete it using https://github.com/woocommerce/woocommerce-admin/blob/main/src/Install.php#L469.

mattsherman:
@becdetat 

> This works well and the code looks good. This doesn't remove the note for existing stores with that notice though - you can delete it using https://github.com/woocommerce/woocommerce-admin/blob/main/src/Install.php#L469.

Hmm... I'm already doing that!

https://github.com/woocommerce/woocommerce-admin/pull/5499/files#diff-e62c61ee511db026c4ed9b6bd2eb5944b07872018f0fcfba17968345e01ada28R475

I wonder if perhaps you missed that, or if for some reason it wasn't functioning properly for you. Can you please re-test this part?

timmyc (approved):
Tested out perfect using the instructions, and I verified that the note is indeed deleted after the update routine is triggered by changing the wc_admin_version option.

mattsherman:
Thanks for testing @becdetat and @timmyc!

PR #5504, opened 2020-10-29
Use the Puppeteer waitForSelector to fix flakyness in e2e tests.

samueljseay:
There was a critical flaw in the helper code to wait on selectors. I am not sure why I wrote that function instead of relying on Puppeteers core `waitForSelector` API.

The incorrect implementation of this helper, was causing a situation where the `page.click` failed sometimes for the theme selection spec in the onboarding wizard.

To clarify what was happening before:

1. my helper did not properly wait for the selector to appear
2. then calling page.click with the same selector failed
3. then it seemed like it was testing the benefits section, but actually it was clicking a secondary button on the theme page, which would pass.

This was confirmed after fixing waitForSelector the benefits page failed because that spec had a selector that was not correct for that page and it started failing. I've also fixed the benefit step to represent what is rendered in that step.

### DIFF
--- a/tests/e2e/utils/lib.js
+++ b/tests/e2e/utils/lib.js
@@ -1,4 +1,4 @@
-export const waitForElementCount = async function ( page, domSelector, count ) {
+export const waitForElementCount = function ( page, domSelector, count ) {
 	return page.waitForFunction(
 		( domSelector, count ) => {
 			return document.querySelectorAll( domSelector ).length === count;
--- a/tests/e2e/utils/lib.js
+++ b/tests/e2e/utils/lib.js
@@ -1,12 +1,5 @@
-export const waitForSelector = async function ( page, selector ) {
-	return await page.evaluate(
-		( selector ) => Boolean( document.querySelector( selector ) ),
-		selector
-	);
-};
-
 export const waitForElementCount = async function ( page, domSelector, count ) {
-	return await page.waitForFunction(
+	return page.waitForFunction(
 		( domSelector, count ) => {
 			return document.querySelectorAll( domSelector ).length === count;
 		},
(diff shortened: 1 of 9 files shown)


tests/e2e/utils/lib.js
```diff
@@ -1,12 +1,5 @@
-export const waitForSelector = async function ( page, selector ) {
-	return await page.evaluate(
-		( selector ) => Boolean( document.querySelector( selector ) ),
-		selector
-	);
-};
-
 export const waitForElementCount = async function ( page, domSelector, count ) {
-	return await page.waitForFunction(
+	return page.waitForFunction(
```
samueljseay, line 2: `return await` is an anti-pattern that eslint would normally catch. I've removed it.

tests/e2e/utils/lib.js
```diff
@@ -1,12 +1,5 @@
-export const waitForSelector = async function ( page, selector ) {
-	return await page.evaluate(
```
samueljseay, line 2: This implementation just waited for the selector to run but it didn't wait on a positive result. I think this was supposed to be a call to `waitForFunction` but it should have just been using `waitForSelector` rather than a helper.

tests/e2e/utils/lib.js
```diff
@@ -1,12 +1,5 @@
...
-		( selector ) => Boolean( document.querySelector( selector ) ),
-		selector
-	);
-};
-
 export const waitForElementCount = async function ( page, domSelector, count ) {
```
jeffstieler, line 1: Looks like we can remove the `async` here.

jeffstieler (approved):
Clean up the unneeded `async` and :shipit:

tests/e2e/utils/lib.js
samueljseay: Ha good spotting. We should probably run lint over this code

PR #5510, opened 2020-10-29
Fixing features contained in subdirectories are not loaded when using Composer v2

### LINKED ISSUE #5509: Features contained in subdirectories are not loaded when using Composer v2 (ie Navigation)
**Describe the bug**
Composer 2 enforces stricter naming conventions that are breaking current logic in loading conditional features. This is currently resulting in the Navigation feature not being loaded when using Composer 2.

**To Reproduce**
- Upgrade to Composer 2.
- Enable Navigation via feature config as well as the `woocommerce_navigation_enabled` option
- Observe that the new navigation still isn't loaded properly

joelclimbsthings:
Fixes #5509

Composer 2 enforces stricter naming conventions that are breaking current logic in loading conditional features. This is currently resulting in the Navigation feature not being loaded when using Composer 2.

To resolve I'm currently fixing the incorrect namespace of the `Navigation.php` file (which composer v1 tolerated), and making a couple changes to the logic to accommodate features that are nested within a subdirectory under `/Features`. Currently opting to use the convention of naming the bootstrapping file within a feature `Init.php`.

### Screenshots

image

- Upgrade to Composer v2
- Run `composer install`
- Enable new Navigation (via feature config as well as `woocommerce_navigation_enabled` option)
- Ensure new navigation is shown when you navigation to WooCommerce

### DIFF
--- a/src/Features/Navigation/Init.php
+++ b/src/Features/Navigation/Init.php
@@ -5,7 +5,7 @@
  * @package Woocommerce Admin
  */
 
-namespace Automattic\WooCommerce\Admin\Features;
+namespace Automattic\WooCommerce\Admin\Features\Navigation;
 
 use Automattic\WooCommerce\Admin\Loader;
 use Automattic\WooCommerce\Admin\Features\Navigation\Screen;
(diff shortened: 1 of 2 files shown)


joshuatf (approved):
Nice!  Works as described and nav is working again on composer v2.  LGTM 🚢

PR #5511, opened 2020-10-30
Only show CES survey on correct page

### LINKED ISSUE #5482: Customer effort score: Only show the survey in the correct screen
As built so far if there is a survey it could potentially be shown on any page per [this](https://github.com/woocommerce/woocommerce-admin/pull/5324#pullrequestreview-517724867).

The survey should only be shown in the correct screen. 

Currently this will only be checking for the product edit screen (the target screen for both adding and editing a product) but the concept will be used for any new trigger actions.

becdetat:
Fixes #5482

This uses `window.pagenow` and `window.adminpage` to restrict which pages certain surveys are shown for.

### Screenshots
Before:

After:
screencast 2020-10-30 16-05-33

- Open two windows and arrange them next to each other
- On the left hand window open the Add Product page
- On the right hand window open the WooCommerce home page
- Back on the left hand window create a new product
- As the left hand window page starts to load the Edit page, refresh the right hand window
- The CES survey should pop up on the LHS screen (the edit page), but not on the RHS screen (the WC home page)

### Changelog Note:

Dev: Only show CES survey on correct page


jeffstieler (approved):
Code looks good and tests well, this can be 🚢  ed - after rebasing to clean up the duplicated commits from the spike PR.

Some implementation notes that might want to get considered for later changes:
- Using options instead of user meta could trigger a CES modal for a different user also on product add/edit screens
- We might want to consider only requesting queue items for the current page rather than all queue items and filtering on the client side. I'm concerned this may not scale well

PR #5512, opened 2020-10-30
Add PSR-4 naming checks to PHP linting

mattsherman:
This PR adds PSR-4 naming checks to the PHP linting, so that we can catch any PSR-4 violations before committing.

It also fixes an issue with `src/Package.php` not being PSR-4 compliant by moving it to `src/Composer/Package.php`.

- Check out branch
- Run `npm run lint:php`
- Verify that a PSR-4 error is found in `src/Features/Navigation/Navigation.php`
- Verify that no PSR-4 error is found in `src/Composer/Package.php`
- Add `error_log( __FILE__ );` to the `init()` method in `src/Composer/Package.php` and in your `../woocommerce/packages/woocommerce-admin/src/Package.php`
- With `woocommerce-admin` active, verify that the error log does not contain the path of the packaged woocommerce-admin in core
- Deactivate `woocommerce-admin` plugin
- Verify that the error log does contain the path of the packaged woocommerce-admin in core

### Changelog Note:

Dev: Add PSR-4 naming checks to PHP linting.


joshuatf (approved):
This is testing great for me.  Thanks so much for adding this @mattsherman and @jeffstieler 💯 🚢

jeffstieler (dismissed):
Sorry for the drive by, from Slack:

> sorry for the lack of detail, i think the changes need to be more like > https://github.com/woocommerce/woocommerce-admin/pull/5510 (edited) 
> 
> my concern is that the Automattic\WooCommerce\Admin\Composer\Package class won’t load from WC core because it isn’t compliant with PSR-4
>
> only the deprecated notes class file should be excluded (in /src) from the PSR-4 check

mattsherman:
Okay, I'll remove the exclusion of `src\Package.php` from this PR, and then a separate PR can be opened to fix `src\Package.php`

jeffstieler:
I added a fix for the `Package.php` PSR-4 breakage and instructions on how to test. This is ready for another review.

samueljseay (approved):
Went through the test instructions and all works as advertised. 🚢

mattsherman:
Thanks for the fix @jeffstieler -- I'll run through it, make sure I understand the fix and verify it works, and then deploy!

PR #5516, opened 2020-11-02
Home Screen - Changed design and onClick behavior

octaedro:
This is a follow-up PR related to [PR 5455](https://github.com/woocommerce/woocommerce-admin/pull/5455)

This PR addresses the changes suggested by @elizaan36 [here](https://github.com/woocommerce/woocommerce-admin/issues/5238#issuecomment-718954046) and via Slack.

The status indicator was modified and also clicking anywhere in the row now opens the order details.

### Screenshots
Screen Capture on 2020-11-02 at 14-32-49

- Verify the order card inside the accordion has its status indicator. E.g.: `Processing`
- Click on the order and verify that the redirection to the order details works as expected.

### Changelog Note:
Dev: Changed design and onClick behavior

### DIFF
--- a/client/header/activity-panel/activity-card/index.js
+++ b/client/header/activity-panel/activity-card/index.js
@@ -84,7 +84,7 @@ class ActivityCard extends Component {
 			return (
 				<Button
 					className="woocommerce-activity-card__button"
-					onClick={ ( element ) => onClick( element ) }
+					onClick={ onClick }
 				>
 					{ this.getCard() }
 				</Button>
--- a/client/header/activity-panel/activity-card/index.js
+++ b/client/header/activity-panel/activity-card/index.js
@@ -7,6 +7,7 @@ import Gridicon from 'gridicons';
 import moment from 'moment';
 import PropTypes from 'prop-types';
 import { H, Section } from '@woocommerce/components';
+import { Button } from '@wordpress/components';
 
 /**
  * Internal dependencies
@@ -14,7 +15,7 @@ import { H, Section } from '@woocommerce/components';
 import './style.scss';
 
 class ActivityCard extends Component {
-	render() {
+	getCard() {
 		const {
 			actions,
 			className,
@@ -76,6 +77,21 @@ class ActivityCard extends Component {
 			</section>
 		);
 	}
+
+	render() {
+		const { onClick } = this.props;
+		if ( onClick ) {
+			return (
+				<Button
+					className="woocommerce-activity-card__button"
+					onClick={ ( element ) => onClick( element ) }
+				>
+					{ this.getCard() }
+				</Button>
+			);
+		}
+		return this.getCard();
+	}
 }
 
 ActivityCard.propTypes = {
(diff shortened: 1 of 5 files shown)


client/header/activity-panel/activity-card/index.js
```diff
@@ -76,13 +77,29 @@ class ActivityCard extends Component {
...
 		);
 	}
+
+	render() {
+		const { onClick } = this.props;
+		if ( onClick ) {
+			return (
+				<Button
+					className="woocommerce-activity-card__button"
+					onClick={ ( element ) => onClick( element ) }
```
samueljseay, line 87: This can just be:
```suggestion
					onClick={ onClick }
```

client/homescreen/activity-panel/orders/index.js
```diff
@@ -155,20 +160,33 @@ class OrdersPanel extends Component {
 
 		const cards = [];
 		orders.forEach( ( order ) => {
```
samueljseay, line 162: Not a big deal but you can destruct the order as its passed into the function:

```suggestion
		orders.forEach( ( { date_created_gmt: dateCreatedGmt } ) => {
```

client/homescreen/activity-panel/orders/index.js
```diff
@@ -155,20 +160,33 @@ class OrdersPanel extends Component {
...
 
 			cards.push(
 				<ActivityCard
-					key={ order.order_id }
+					key={ orderId }
 					className="woocommerce-order-activity-card"
 					title={ orderCardTitle( order ) }
-					date={ order.date_created_gmt }
+					date={ dateCreatedGmt }
+					onClick={ ( element ) => {
```
samueljseay, line 182: Its a tiny thing but this should always be named `event` to avoid confusion because it's not an element. It can also be destructured here if you like.

```suggestion
					onClick={ ( event ) => {
```

or with destructuring:

```suggestion
					onClick={ ( { target } ) => {
```

samueljseay (commented):
Just some little suggestions from me.

client/homescreen/activity-panel/orders/index.js
octaedro: Nice catch! This was changed in the commit 226756e

client/homescreen/activity-panel/orders/index.js
octaedro: I like the idea of deconstructing `order` but I'm not sure if it will be the best solution in this case, since we need to use `order` inside the `forEach` [here](https://github.com/woocommerce/woocommerce-admin/pull/5516/files#diff-29ea5c8bb0a22385b19ffc80538ecd1a1d660f160f7a77debcb03240e62dbfe3R180) and [here](https://github.com/woocommerce/woocommerce-admin/pull/5516/files#diff-29ea5c8bb0a22385b19ffc80538ecd1a1d660f160f7a77debcb03240e62dbfe3R208). If you think it's a good idea to continue with the deconstruction please let me know and I'll change the method `orderCardTitle` and the component `OrderStatus`

octaedro:
Thank you for the review @samueljseay. I addressed the changes you mentioned.
Could you take another look at this PR?

client/homescreen/activity-panel/orders/index.js
samueljseay: Ah sorry I didn't spot those, yep that makes sense to stay as is 😄

samueljseay (approved):
Fantastic stuff, its testing well for me. :shipit:

PR #5519, opened 2020-11-03
Add width: 100% style to the container of the popover menu to expand …

### LINKED ISSUE #5191: Advanced filter menu has broken styles
**Describe the bug**
When selecting an advanced filter to add to an analytics report, the style is broken. Noted while testing https://github.com/woocommerce/woocommerce-admin/pull/5167 however it reproduces in all reports with advanced filters. 

**To Reproduce**

1. Navigate to the orders report
2. In the "show" menu, switch to "Advanced filters"
3. Add a filter
4. Note the highlight only extends as far as the text, rather than across the width of the container

**Screenshots**
Screen Shot 2020-09-23 at 9 30 22 PM

**Additional context**
We may want to consider switching this menu to use the same component / styles that are available in the block editor toolbar menu: 

Screen Shot 2020-09-23 at 9 47 48 PM

cc: @elizaan36

moon0326:
Fixes #5191 

Added ```width: 100%``` style to the direct child of ```div.components-popover__content``` in order to give 100% width to the child ul element.
### Screenshots

**Before**
before

**After**
after

1. Navigate to Analytics -> Orders (or any other pages with Advanced Filters) 
2. Change **Show:** option to "Advanced Filters" from "All Orders"
3 Click "+ Add a Filter"
4. Hover over each menu item.
5. Note that the grey background (hover state) is fully extended to 100% width.

### DIFF
--- a/packages/components/src/advanced-filters/style.scss
+++ b/packages/components/src/advanced-filters/style.scss
@@ -102,6 +102,10 @@
 	.components-popover:not(.is-mobile) .components-popover__content {
 		min-width: 180px;
 	}
+
+	.components-popover__content > div {
+		width: 100%;
+	}
 }
 
 .woocommerce-filters-advanced__fieldset {


becdetat (approved):
Works well, code looks good. Thanks! 🚢

PR #5522, opened 2020-11-03
Wrong class name for querying Categories Report

zzap:
The inline docs example for querying Categories Report is missing `\Categories` in class name.


samueljseay:
@zzap Thanks so much for this! 🚢

samueljseay (approved):

PR #5524, opened 2020-11-03
Navigation: Adding ability for new navigation to scroll when list of links exceed viewport height

### LINKED ISSUE #5508: Navigation: Enable scroll on long list of items
If you shrink the viewport height, navigation doesn't scroll properly. 

Screen Shot 2020-10-30 at 11 07 56 AM

### Deliverables

* Navigation scrolls on short viewports or long menus
* It should scroll independently of main screen area
* It should work on mobile
* Determine the overscroll issue is best suited to being in Gutenberg.
* Ensure the scroll works intuitively with the header bar.
* Make sure the scroll is reset on navigation so user doesn't have to scroll to the top when navigating.
* Add a padding bottom so the last link isn't crowded down there.

joelclimbsthings:
Fixes #5508 

The new side navigation needs to having scrolling capability to accommodate large lists of links or short viewports as per the issue. 

**Requirements:**

- [x] Navigation scrolls on short viewports or long menus
- [x] It should scroll independently of main screen area
- [x] It should work on mobile
- [x] Add a padding bottom so the last link isn't crowded down there.

- [x] Determine the overscroll issue is best suited to being in Gutenberg.
- [x] Ensure the scroll works intuitively with the header bar.
- [x] Make sure the scroll is reset on navigation so user doesn't have to scroll to the top when navigating.

I'm a little uncertain if this change is appropriate in GB, partially due to the fact that it involves the variable `$header-height` which is only available in wc-admin. 

I haven't witnessed any funny business with the header bar, but if @psealock  wants to expand on the weirdness he mentioned then I'd like to be sure I'm not just missing this.

The last point, in terms of the scroll height being reset when navigation to submenus, would likely be a bit of an edge case since most submenus will likely be pretty short. That said, in order to address this properly we might have to issue a PR to the Gutenberg component logic to scroll to the top when navigation happens. I'm still looking into this.

### Screenshots

image

- Check out branch
- Enable new navigation feature
- Go into WooCommerce menu
- If your viewport is larger than the navigation menu, no scroll bar should appear
- Shrink your viewport window to a smaller screensize, and ensure that scrollbar appears
- Simulate or use on mobile, ensure behavior still meets requirements.

### DIFF
--- a/client/navigation/components/container/index.js
+++ b/client/navigation/components/container/index.js
@@ -84,7 +84,7 @@ const Container = ( { menuItems } ) => {
 			<Navigation
 				activeItem={ activeItem ? activeItem.id : null }
 				activeMenu={ activeLevel }
-				onActivateMenu={ ( { menuId } ) => {
+				onActivateMenu={ ( ...args ) => {
 					const navDomRef = document.querySelector(
 						'.components-navigation'
 					);
@@ -93,7 +93,7 @@ const Container = ( { menuItems } ) => {
 						navDomRef.scrollTop = 0;
 					}
 
-					setActiveLevel( menuId );
+					setActiveLevel( ...args );
 				} }
 			>
 				{ activeLevel === 'woocommerce' && dashboardUrl && (
--- a/client/navigation/components/container/index.js
+++ b/client/navigation/components/container/index.js
@@ -84,7 +84,17 @@ const Container = ( { menuItems } ) => {
 			<Navigation
 				activeItem={ activeItem ? activeItem.id : null }
 				activeMenu={ activeLevel }
-				onActivateMenu={ setActiveLevel }
+				onActivateMenu={ ( { menuId } ) => {
+					const navDomRef = document.querySelector(
+						'.components-navigation'
+					);
+
+					if ( navDomRef ) {
+						navDomRef.scrollTop = 0;
+					}
+
+					setActiveLevel( menuId );
+				} }
 			>
 				{ activeLevel === 'woocommerce' && dashboardUrl && (
 					<NavigationBackButton
--- a/client/navigation/components/container/style.scss
+++ b/client/navigation/components/container/style.scss
@@ -3,11 +3,6 @@
 		color: inherit; // This should be handled at component level.
 	}
 
-	.components-navigation {
-		height: 100%;
-		box-sizing: border-box;
-	}
-
 	.components-navigation__menu {
 		display: flex;
 		flex-direction: column;
--- a/client/navigation/style.scss
+++ b/client/navigation/style.scss
@@ -22,6 +22,12 @@
 		z-index: 1100; //Must be greater than z-index on .woocommerce-layout__header
 	}
 
+	.components-navigation {
+		overflow-y: auto;
+		height: calc(100vh - #{$header-height});
+		box-sizing: border-box;
+	}
+
 	&:not(.is-folded) {
 		#wpcontent,
 		#wpfooter {


psealock (commented):
This is great, thanks for the thorough checks of all the points. I think you satisfied the remaining points so I checked them off the list.

The one about scroll is still something to address in my opinion, although it will be an edge case. If you shrink the viewport and scroll down to choose a category at the bottom of the list, the nav changes to show the category's children as expected. But, you are viewing the bottom of the list. I think it would be better to make the scroll zero after every navigation, forward and back.

joshuatf (commented):
Nice work!  ➕ 1️⃣ on Paul's comment, but otherwise looks great.

joelclimbsthings:
Thanks for the feedback guys, this ended up being an easy fix via the `onActivateMenu` prop :+1:

client/navigation/components/container/index.js
```diff
@@ -84,7 +84,17 @@ const Container = ( { menuItems } ) => {
 			<Navigation
 				activeItem={ activeItem ? activeItem.id : null }
 				activeMenu={ activeLevel }
-				onActivateMenu={ setActiveLevel }
+				onActivateMenu={ ( { menuId } ) => {
```
joshuatf, line 87: Nice use of `onActivateMenu` to solve this issue!

client/navigation/components/container/index.js
```diff
@@ -84,7 +84,17 @@ const Container = ( { menuItems } ) => {
...
+					const navDomRef = document.querySelector(
```
joshuatf, line 88: Probably a little bit edge case, but it would be nice to use `useRef()` to get this element instead of a query selector.  We might need to wrap the `Navigation` with a class like `woocommerce-navigation__wrapper` to make that work though.

joshuatf (commented):
Scrolling issue is solved!  Left one small potential clean-up; let me know what you think.

client/navigation/components/container/index.js
joelclimbsthings: Yeah, I started to go for a `useRef()` but realized that it would require a wrapper/etc as you mentioned. Wasn't sure it was worth the extra boilerplate. Is the main concern that the targeted class could change @joshuatf ?

timmyc (commented):
@joelclimbsthings when testing this out and navigating from `/wp-admin` to the Woo homescreen, I'm getting the following error ( disregard the Note warning ;) ):

image

joelclimbsthings:
Can't see to reproduce that error @timmyc , although I noticed in further testing that the back button was actually disappearing on navigation. I've just committed a fix for that.

timmyc (commented):
So I think my prior error report was caused by a problem between the chair and the keyboard over here.

No exceptions seen anymore here, but I am seeing a scrollbar in the nav sidebar when I'm thinking there should be one. Here is a screen grab - tested on MS Edge, seeing something similar in Safari too, though also seeing some other wonky css bugs over there that I'll have to report separately ( not related to this branch I don't think as they exist when nav is off too )

https://cloudup.com/cS3aZ4O7nZ7

octaedro (approved):
Good job @joelclimbsthings! This is testing well here and code LGTM.

joelclimbsthings:
>So I think my prior error report was caused by a problem between the chair and the keyboard over here.
>
>No exceptions seen anymore here, but I am seeing a scrollbar in the nav sidebar when I'm thinking there should be one. Here is a screen grab - tested on MS Edge, seeing something similar in Safari too, though also seeing some other wonky css bugs over there that I'll have to report separately ( not related to this branch I don't think as they exist when nav is off too )

Running Edge v88 on Ubuntu, and I can't seem to reproduce that issue either. o.0

client/navigation/components/container/index.js
joshuatf: There is the possibility of the class changing, but there's also the possibility that the `Navigation` component is used multiple times across a page.  We don't really have a use case for this (yet), but if there are multiple navigation components, the added styling for scroll behavior on the class and this `scrollTop` would affect all of the components and not just this one.

I think adding a wrapping class is little overhead to guard against future conflicts.

PR #5527, opened 2020-11-04
Update/5394 remove label printing mention for non us countries

### LINKED ISSUE #5394: Inbox notification - review shipping settings: remove label printing mention if the store country is not US
The label printing feature that WooCommerce Shipping offers only works in the United States. It will be necessary to create a new notification that targets non-US Stores without the mentioning label printing:
https://d.pr/i/5gIx5l

moon0326:
Fixes #5394 

This PR removes `Label printing: enabled` string from the shipping admin note for non-US countries by checking the store's base country.

Dev note: The code assumes that the WooCommerce Shipping extension and its label printing feature are enabled. I think it would be better if we actually perform a check on the WooCommerce Shipping extension availability and its settings, but I think that's a separate issue and requires a new PR.

### Screenshots

United States (notice we still have `Label printing: enabled`)
beforeth_

non-US countries:

afterth_

It's tricky to test the change locally as we need Jetpack connection. I would recommend using http://jurassic.ninja/ to create a publicly accessible WordPress installation.

1. Create a fresh installation of WordPress.
2. Navigate to Plugins and install WooCommerce
3. Navigate to Plugins and click [ Add New]. Upload the attached zip file.
[woocommerce-admin.zip](https://github.com/woocommerce/woocommerce-admin/files/5484838/woocommerce-admin.zip)

4. Navigate to Plugins -> Installed Plugins and activate both WooCommerce and WooCommerce Admin plugin.
5. Click "WooCommerce" from the left sidebar to kick off the setup wizard.

**Scenario 1** - United States
**Expectation**: We should still see `Label printing: enabled` string from the admin note after finishing up the wizard.

Continuing from the above steps,

1. Fill out the address form and make sure to choose "United States (US) --- :state" from the Country/Region dropdown on the "Store Details" step and click [ Continue ]
2. Check at least one industry from the "Industry" step and click [ Continue ]
3. Choose "Physical products" from the "Product Types" step.
4. Continue and finish the wizard.
5. Make sure to choose "Yes please!" when "Enhance your storewith WooCommece Shipping & Tax" option pops up.
6. Make sure to connect to Jetpack and approve your site.

**Scenario 2** - Non-US countries
**Expectation**: We should not see `Label printing: enabled` string from the admin note after finishing up the wizard.

Same steps as scenario 1, except that you need to choose a country that is not one of United States, Australia, Canada, or United Kingdom. The reason we include Australia, Canada, and United Kingdom is that the shipping onboarding process won't continue for those countries according to https://github.com/woocommerce/woocommerce-admin/blob/main/src/Features/OnboardingSetUpShipping.php#L54

### DIFF
--- a/src/Notes/ReviewShippingSettings.php
+++ b/src/Notes/ReviewShippingSettings.php
@@ -27,7 +27,7 @@ class ReviewShippingSettings {
 	public static function get_note() {
 		$note = new Note();
 
-		$note_lines = array(
+		$content_lines = array(
 			__(
 				"Based on the information that you provided we've configured some shipping rates and options for your store:<br/><br/>",
 				'woocommerce-admin'
@@ -37,13 +37,13 @@ class ReviewShippingSettings {
 		);
 
 		if ( 'US' === WC()->countries->get_base_country() ) {
-			array_push( $note_lines, __( 'Label printing: enabled', 'woocommerce-admin' ) );
+			array_push( $content_lines, __( 'Label printing: enabled', 'woocommerce-admin' ) );
 		}
 
 		$note->set_name( self::NOTE_NAME );
 		$note->set_source( 'woocommerce-admin' );
 		$note->set_title( __( 'Review your shipping settings', 'woocommerce-admin' ) );
-		$note->set_content( implode( '', $note_lines ) );
+		$note->set_content( implode( '', $content_lines ) );
 		$note->set_content_data( (object) array() );
 		$note->set_type( Note::E_WC_ADMIN_NOTE_INFORMATIONAL );
 
--- a/src/Notes/ReviewShippingSettings.php
+++ b/src/Notes/ReviewShippingSettings.php
@@ -27,18 +27,23 @@ class ReviewShippingSettings {
 	public static function get_note() {
 		$note = new Note();
 
+		$note_lines = array(
+			__(
+				"Based on the information that you provided we've configured some shipping rates and options for your store:<br/><br/>",
+				'woocommerce-admin'
+			),
+			__( 'Domestic orders: free shipping<br/>', 'woocommerce-admin' ),
+			__( 'International orders: disabled<br/>', 'woocommerce-admin' ),
+		);
+
+		if ( 'US' === WC()->countries->get_base_country() ) {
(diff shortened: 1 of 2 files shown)


src/Notes/ReviewShippingSettings.php
```diff
@@ -27,18 +27,23 @@ class ReviewShippingSettings {
 	public static function get_note() {
 		$note = new Note();
 
-		$note->set_name( self::NOTE_NAME );
-		$note->set_source( 'woocommerce-admin' );
-		$note->set_title( __( 'Review your shipping settings', 'woocommerce-admin' ) );
-		$note->set_content(
+		$note_lines = array(
```
becdetat, line 30: Would `$content_lines` describe this better?

moon0326: Good point! Yup, I think `$content_lines` makes more sense. I pushed a new commit.

moon0326:
Thank you @becdetat. I made a new commit to reflect your suggestion 👍

becdetat:
Looks good, merge when you're ready :-)

tests/features/class-onboarding-set-up-shipping.php
```diff
@@ -66,4 +67,36 @@ public function test_zone_does_not_get_created_with_no_default_country() {
...
+	public function test_shipping_note_does_not_include_label_printing_for_non_us_countries() {
+		// Given AU:QLD as the default country.
+		update_option( 'woocommerce_default_country', 'AU:QLD' );
+
+		// When note is retrieved.
+		$note = ReviewShippingSettings::get_note()->get_content();
+
+		// Then it should not contain 'Label printing: enabled'.
+		$contains_string = strpos( $note, 'Label printing: enabled' );
+		$this->assertFalse( $contains_string );
```
jeffstieler, line 84: For future reference, there is an [`assertStringContainsString()`](https://phpunit.readthedocs.io/en/7.5/assertions.html#assertstringcontainsstring) assertion. 😄

timmyc:
Great PR @moon0326 - thanks so much for adding test coverage, and being so detailed in your description too <3

moon0326:
@timmyc Thank you :)

PR #5532, opened 2020-11-04
Copy the snackbar component from Gutenberg into wp-admin

becdetat:
_EDIT: I needed to copy the `core/notices` data source (which is where `createNotice` lives), because the entire options object isn't passed down to the snackbar component meaning that each new option needed for the CES feature will have to be implemented in `core/notices` as well._

This copies the [`SnackbarList`](https://github.com/WordPress/gutenberg/blob/master/packages/components/src/snackbar/list.js) and [`Snackbar`](https://github.com/WordPress/gutenberg/blob/master/packages/components/src/snackbar/index.js) components in from Gutenberg, as well as the `core/notices` data source (renamed to `core/notices2`) from the `@wordpress/notices` package. This is so we can iterate over the snackbar designs in the Customer Effort Score feature without being blocked by a Gutenberg release.

NOTE: When the changes to the snackbar are pushed upstream to Gutenberg and released, the components and data source can be removed.

Follow the test instructions in this PR #5406, and make sure the snackbar works as expected.

### Changelog Note:

Dev: Copy the snackbar component and core/notices from Gutenberg into WordPress Admin

### DIFF
--- a/client/layout/transient-notices/index.js
+++ b/client/layout/transient-notices/index.js
@@ -5,12 +5,12 @@ import classnames from 'classnames';
 import { Component } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import PropTypes from 'prop-types';
-import { SnackbarList } from '@wordpress/components';
 import { withDispatch, withSelect } from '@wordpress/data';
 
 /**
  * Internal dependencies
  */
+import SnackbarList from './snackbar/list';
 import './style.scss';
 
 class TransientNotices extends Component {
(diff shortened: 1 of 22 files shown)


mattsherman (approved):
Still works as expected!

becdetat:
@mattsherman Thanks! Turned out I needed to pull in the `core/notices` data source in as well, as all of the modifications to the snackbar also need to be reflected in the data source. Can you retest this please?

mattsherman (approved):
Change works as expected.

We should make sure that we have a specific plan for moving our improvements back upstream to Gutenberg so that our local copy doesn't linger for longer than necessary.

becdetat:
> We should make sure that we have a specific plan for moving our improvements back upstream to Gutenberg so that our local copy doesn't linger for longer than necessary.

Great point. I've got a plan in my mind but I'll get it out into a P2 post tomorrow.

PR #5537, opened 2020-11-05
Add documentation for navigation feature

### LINKED ISSUE #5426: Navigation Documentation: Add wc-admin doc
Continuing with https://github.com/woocommerce/woocommerce-admin/issues/5421 to create a working example for 3PD to register into the Navigation, we'll need some documentation so that plugins can easily integrate.

### Deliverables

* Create a WC Admin README or doc.
* Explain the significance of the Navigation project.
* Provide step by step instructions on registering PHP as well as JS custom links and why that is useful for router-aware links.

cc @pmcpinto

joshuatf:
Fixes #5426 

Adds docs explaining nav features and how to register items.

1. Read the docs.
1. Make sure they make sense and enough info is provided to register plugin items.

### DIFF
--- a/docs/features/navigation.md
+++ b/docs/features/navigation.md
@@ -12,7 +12,7 @@ The fastest way to get started is by creating an example plugin from WooCommerce
 
 `npm run example -- --ext=add-navigation-items`
 
-This will create a new plugin that covers various features of the navigation and helps to register some intial items and categories within the new navigation menu.
+This will create a new plugin that covers various features of the navigation and helps to register some intial items and categories within the new navigation menu.  After running the command above, you can make edits directly to the files at `docs/examples/extensions/add-navigation-items` and they will be built and copied to your `wp-content/add-navigation-items` folder on save.
 
 ### Adding a menu category
 
--- a/docs/features/navigation.md
+++ b/docs/features/navigation.md
@@ -1,0 +1,121 @@
+# WooCommerce Navigation
+
+The WooCommerce Navigation feature is a navigational project designed to create a more intuitive and functional WooCommerce specific navigation.
+
+This API will allow you to add in your own items to the navigation and register pages with the new navigation screens.
+
+### Getting started
+
+This feature is hidden behind a feature flag and can be turned on or off by visiting WooCommerce -> Settngs -> Advanced -> Features and checking the box next to the `Navigation` option.  It can also by controlled programmatically by setting the option `woocommerce_navigation_enable` to `yes` or `no`. 
+
+The fastest way to get started is by creating an example plugin from WooCommerce Admin.  Inside your `woocommerce-admin` directory, enter the following command:
+
+`npm run example -- --ext=add-navigation-items`
+
+This will create a new plugin that covers various features of the navigation and helps to register some intial items and categories within the new navigation menu.
+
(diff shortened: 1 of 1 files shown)


samueljseay (approved):
This looks great and I found it really informative. 🚀

docs/features/navigation.md
```diff
@@ -0,0 +1,121 @@
...
+### Getting started
+
+This feature is hidden behind a feature flag and can be turned on or off by visiting WooCommerce -> Settngs -> Advanced -> Features and checking the box next to the `Navigation` option.  It can also by controlled programmatically by setting the option `woocommerce_navigation_enable` to `yes` or `no`. 
+
+The fastest way to get started is by creating an example plugin from WooCommerce Admin.  Inside your `woocommerce-admin` directory, enter the following command:
+
+`npm run example -- --ext=add-navigation-items`
+
+This will create a new plugin that covers various features of the navigation and helps to register some intial items and categories within the new navigation menu.
+
```
psealock, line 16: I've seen devs get confused by the magic of webpack copying files to the actual plugin. Being able to edit directly in wc-admin may not be intuitive, so a line to the effect of "you can edit directly in wc-amin" pointing to the source files https://github.com/woocommerce/woocommerce-admin/tree/main/docs/examples/extensions/add-navigation-items could be useful.

docs/features/navigation.md
```diff
@@ -0,0 +1,121 @@
...
+        <WooNavigationItem item="example-plugin">
+            <Button onClick={ handleClick }>
+                { __( 'My Link', 'plugin-domain' ) }
+            </Button>
+        </WooNavigationItem>
+    );
+};
+
+registerPlugin('my-plugin', { render: MyPlugin });
+```
```
psealock, line 121: One gotcha I ran into when setting this up from scratch with my example plugin is that when registering scripts, a dependency on `WC_ADMIN_APP` was required when registering js with `wp_register_script` when the slotFill was being used.

https://github.com/woocommerce/woocommerce-admin/blob/d5a0b32126f655bf3703aac21826397bf5a0df40/src/FeaturePlugin.php#L149

Something to that effect could save some headaches for developers. Separately, it would be great to build this into Woo's Dependency Extraction Webpack Plugin: https://github.com/woocommerce/woocommerce-admin/issues/5559

docs/features/navigation.md
joshuatf: Great point.  Added a comment about how this works in https://github.com/woocommerce/woocommerce-admin/pull/5537/commits/7d98b9f3040bfd11d95212cf84fa76b6a7394d60

docs/features/navigation.md
joshuatf: I'm not quite seeing how to replicate this, but I believe I understand the issue.  Let's follow up in https://github.com/woocommerce/woocommerce-admin/issues/5559 and we can append info the docs if necessary.

psealock: Sounds good. Right now, if you try to use slotFill in your app it will cause a fatal error. The reason is because wc-admin needs to load first, so 'wc-admin-app' needs to be added as a dependency. Its fine in the example extensions because thats technically part of wc-admin so load order doesn't matter.

PR #5542, opened 2020-11-05
Customers: Update column heading for date registered.

### LINKED ISSUE #5414: Customers report table column title used in wrong context
**Describe the bug**
The Customers report table column "Registered" title is used in wrong context in Polish translation.

Currently the title translation is shared with front-end Sign Up form submit button "Zarejestruj się", which is a call-to-action and literally means "**Let's sign up**". What we need here is something like "**Registered date**", because this column displays dates of users ("Data rejestracji" or simply "Rejestracja"). Maybe a context added with a _x() function would be helpful?

**To Reproduce**

1. Switch WP language to Polish.
2. Go to WooCommerce->Customers.
3. Look at the Registered column title.

**Screenshots**
Zrzut ekranu 2020-10-16 o 16 36 05

WP 5.5.1
WC 4.6.0
WC Admin 1.6.1 (bundled with WC)

timmyc:
Fixes #5414 

The existing column header for the `date_registered` value of a customer is "Sign Up" which I guess kind of means "Sign Up Date" - but regardless it is conflicting with the "Sign Up" string that is used on the front-end of sites, and resulting in oddities with translations ( as reported in the issue linked above ).

The proposed fix here is to just label the column "Date Registered" so we have a unique key for this context, and to me it reads a bit better overall.

customers-column-change

This can be tested by viewing the Customers page.

### DIFF
--- a/client/analytics/report/customers/table.js
+++ b/client/analytics/report/customers/table.js
@@ -47,7 +47,7 @@ class CustomersReportTable extends Component {
 				isSortable: true,
 			},
 			{
-				label: __( 'Sign Up', 'woocommerce-admin' ),
+				label: __( 'Date Registered', 'woocommerce-admin' ),
 				key: 'date_registered',
 				isSortable: true,
 			},


louwie17 (approved):
Looks great Timmy :)

timmyc:
cherry picked to `release/1.7.0` in https://github.com/woocommerce/woocommerce-admin/commit/7d9b50b9aeeab34dea38fe253ce047f21331800e

PR #5546, opened 2020-11-06
Data: Verify we have an interval array before accessing.

### LINKED ISSUE #5536: Analytics: Exception on Variations Report
**Describe the bug**
While preparing some testing documentation for the release, I encountered an exception on the new Variations Report. 

image

Adding in some checks to the `@woocommerce/data` package locally allowed me to get past the exception, and have the report render:

```
diff --git a/packages/data/src/reports/utils.js b/packages/data/src/reports/utils.js
index 9e7f2e76..2be2c4c3 100644
--- a/packages/data/src/reports/utils.js
+++ b/packages/data/src/reports/utils.js
@@ -369,7 +369,9 @@ export function getReportChartData( options ) {
                }
 
                forEach( pagedData, function ( _data ) {
-                       intervals = intervals.concat( _data.data.intervals );
+                       if ( _data.data && _data.data.intervals && Array.isArray( _data.data.intervals ) ) {
+                               intervals = intervals.concat( _data.data.intervals );
+                       }
                } );
        }
```

I git blamed the above code, and looks like that hasn't changed in 2 years, so don't think this is a new exception, but perhaps something related to the variations intervals logic in the REST API.

This fix seems _okay_ but am curious if there is something with my local data that might be causing this, or also keen to see if others can reproduce this bug with the Variations Report

**To Reproduce**
For me I had to select last year as a date range to pull back some orders with variant data, so wondering perhaps if the bug is specific to longer time periods.

1. Variations Report
2. Open the date range picker and select "Last Year" as the range, or just use this url `wp-admin/admin.php?page=wc-admin&path=%2Fanalytics%2Fvariations&period=last_year&compare=previous_year`
3. There will likely be a few React warnings about required PropTypes to appear in your console ( which we should probably fix these too )
4. Then the exception should happen and the screen should go grey.

cc @jeffstieler

timmyc:
Fixes #5536

While testing the Variations report, I hit a bug today. This branch implements a fix by verifying we have a valid interval data array prior to attempting to use it.

- Go to Variations Report on `main`
- Open the date range picker and select "Last Year" as the range, or just use this url `wp-admin/admin.php?page=wc-admin&path=%2Fanalytics%2Fvariations&period=last_year&compare=previous_year`
- There will likely be a few React warnings about required PropTypes to appear in your console, disregard those for now
- Then the exception should happen and the screen should go grey.
- Check out this branch, refresh the page, and verify all is well

@jeffstieler did mention that this might have been due to me not having sales data 3 years back on my test site. I didn't verify that myself, but figured this check would be good to have regardless.

### Changelog Note:

Dev: Add undefined check in intervals data util

### DIFF
--- a/packages/data/src/reports/utils.js
+++ b/packages/data/src/reports/utils.js
@@ -369,7 +369,13 @@ export function getReportChartData( options ) {
 		}
 
 		forEach( pagedData, function ( _data ) {
-			intervals = intervals.concat( _data.data.intervals );
+			if (
+				_data.data &&
+				_data.data.intervals &&
+				Array.isArray( _data.data.intervals )
+			) {
+				intervals = intervals.concat( _data.data.intervals );
+			}
 		} );
 	}
 


samueljseay:
@timmyc @jeffstieler Could we extend the testing of this to include testing the "year to date" filter on the existing reports and **with Gutenberg installed**. Because I see the exact same exception in that situation which was raised to me by an HE today.

I can reproduce it by the following steps:

* Install Gutenberg plugin
* Go to Analytics > Products
* Open filter dropdown, choose "Year to Date", click "update"

samueljseay:
I've tested my use case and it fixes up my bug perfectly. 🎉

samueljseay (approved):
Tested both use cases and this works well. 🚀

PR #5547, opened 2020-11-06
Do a refresh when redirecting to homescreen after completing OBW

### LINKED ISSUE #5521: Inbox notification missing about additional payments options
**Describe the bug**
According to #4597 we should trigger an inbox notification with more info about additional payment gateways when WCPay is being installed during the profiler. But this is not happening since 4.6 was released.

**To Reproduce**

1. Create a new store
2. Choose the US as a country in the Store Details
3. Choose Fashion, apparel, and accessories as Industry
4. Install the extensions bundle in the Business details step (check that WCPay is included)
5. Visit the home screen and confirm that the inbox notification is missing

**Expected behavior**
To see an inbox notification that allows me to install other payment options.

samueljseay:
Fixes #5521

The inbox notification related to installing wc payments was not showing after completing the OBW. This was due to the fact that we don't do a refresh between installing it and going back to the homescreen.

This changes the `completeProfiler` behaviour to do a full refresh at the end to always ensure that **any relevant inbox notifications are loaded when arriving on this screen.**

To test the original bug is fixed:

1. Go through OBW, choosing US for country and Fashion for industry
2. Keep the "install business features" checkbox checked
3. Finish the OBW
4. It should do a refresh when going to the homescreen and the "Set up additional payment providers" inbox notification should be shown

It would also be a good idea to test the redirect functionality of the "connect to jetpack" flow in the OBW to ensure that it does redirect to the correct URL on completion.

### Changelog Note:

Fix: ensure the "Set up additional payment providers" inbox notification is shown when relevant after completing the OBW.

### DIFF
--- a/client/profile-wizard/index.js
+++ b/client/profile-wizard/index.js
@@ -234,20 +234,26 @@ class ProfileWizard extends Component {
 		updateProfileItems( { completed: true } ).then( () => {
 			clearTaskCache();
 
+			const homescreenUrl = new URL(
+				getNewPath( {}, '/', {} ),
+				window.location.href
+			).href;
+
 			if ( shouldConnectJetpack ) {
 				document.body.classList.add( 'woocommerce-admin-is-loading' );
 
-				connectToJetpack(
-					getHistory().push( getNewPath( {}, '/', {} ) )
-				);
+				connectToJetpack( () => {
+					return homescreenUrl;
+				} );
 			} else {
-				getHistory().push( getNewPath( {}, '/', {} ) );
+				window.location.href = homescreenUrl;
 			}
 		} );
 	}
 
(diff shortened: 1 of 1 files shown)


timmyc:
Wanted to note that @pmcpinto saw a bug happen in this part of the flow today during a design research call:

https://www.loom.com/share/fbd68859c3f74839a6e8bcd78072a33b

He is creating an issue around this right now, but essentially the user was shown the Home Screen after clicking the action button to connect to JP. Definitely some buggyness there that we might need to consider alongside this fix.

pmcpinto:
Issue created here: #5551 5551

octaedro (approved):
Good job @samueljseay! This is working well here and code LGTM 🚢

PR #5550, opened 2020-11-06
Added support for custom actionable statuses

### LINKED ISSUE #5545: Support custom actionable statuses in the new homescreen orders panel
As part of #5455 a bug was found with the existing orders fetching where it would flicker while trying to resolve the list of orders. This was due to a bug where the order statuses would change between renders and force the fetching of the orders twice.

For now, to resolve that issue, we have opted to use only the default order statuses, but this means that actionable statuses defined by third parties won't show in this panel for the first iteration.

Criteria: 

- [ ]  Enhance the new orders panel to support loading all defined actionable statuses when considering which orders to display. 
- [ ] Ensure that no regression of flickering is introduced.

octaedro:
Fixes #5545

This PR adds support for custom actionable statuses. A small refactor was added in order to call the setting that provides the `orderStatuses` only once. Also, a control was added in the `OrdersPanel` to confirm that the `reportOrders` and `actionableOrders` lists were filled before rendering.

### Screenshots
Screen Capture on 2020-11-06 at 11-57-38

- Go to the `Home` screen.
- Verify the orders panel is visible.
- Confirm that the orders in progress (and on hold) number is right.
- Verify that the order card looks like it should.
- Verify the buttons and links are working well.
- Verify there is no flickering when loading.

### Changelog Note:
Fix: Added support for custom actionable statuses

### DIFF
--- a/client/homescreen/activity-panel/index.js
+++ b/client/homescreen/activity-panel/index.js
@@ -9,13 +9,14 @@ import { Accordion, AccordionPanel } from '@woocommerce/components';
  * Internal dependencies
  */
 import './style.scss';
-import { getUnreadOrders } from './orders/utils';
+import { getOrderStatuses, getUnreadOrders } from './orders/utils';
 import { getAllPanels } from './panels';
 
 export const ActivityPanel = () => {
 	const panels = useSelect( ( select ) => {
-		const countUnreadOrders = getUnreadOrders( select );
-		return getAllPanels( { countUnreadOrders } );
+		const orderStatuses = getOrderStatuses( select );
+		const countUnreadOrders = getUnreadOrders( select, orderStatuses );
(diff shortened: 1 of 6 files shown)


jeffstieler (approved):
These changes look good and tested well. I used the WooCommerce Order Status Manager plugin to register a custom status. Before registering that status, I tried using an **un**registered status present on my store - but that caused the `/wc-analytics/orders` request to `400`.

We probably need to either remove the validation of the `status` parameter for that endpoint, or filter in the unregistered statuses (see: https://github.com/woocommerce/woocommerce-admin/blob/87f311856c3cbf04048fcdd002fb474b43d71f86/src/Loader.php#L1104).

octaedro:
Thank you @jeffstieler for your review.
I'll merge this PR and create an issue to tackle the unregistered status problem later.

Issue #5243, opened 2020-10-01
Home Screen - consolidate the task UI


pmcpinto:
**Description:** we aim to consolidate some of the tasks' elements, in order to provide a cohesive experience when the user is completing the tasks displayed in the task list.

![Home (4).png](https://images.zenhubusercontent.com/5d84fe8e1201de0001b6ab61/3bfc31df-61f2-489d-b481-c37ceaffc0c8)

**Acceptance criteria:**
- [ ] Every task area needs to have a title, counter with the number of tasks, and an option to show/hide all the tasks
- [ ] The setup task area is the only one that can be dismissed
- [ ] Instead of showing the progress bar in the setup task area we should display the counter with the number of incomplete tasks
- [ ] The rest of the setup tasks UI can remain the same in this iteration: completed tasks checkmark, dismiss feature, etc
- [ ] The tasks should have the following order: setup, orders, stock, reviews, 3PDs

**Event tracking:**
NA

**Additional links:**
[P2 with links to the Figma file](pbIJXs-ma-p2)

jeffstieler:
Lots of detail into how this should work and be implemented with certain components in this comment: pbIJXs-ma-p2#comment-861

elizaan36:
Yup, I'll add the details from the P2 here and will share in GH only for these types of refinements in the future. 

Sounds like it makes sense to add this functionality to the [core card component](https://href.li/?https://wordpress.github.io/gutenberg/?path=/story/components-card--default) (but within wc-admin to start.) Looks like @octaedro has already made these updates in #5455 🎉

Along those lines, I'll share the updates to the design here to be more in line with the WordPress card component and minimize the required changes.

![Home - Setup tasks](https://user-images.githubusercontent.com/5121465/97031453-ebdc7900-1525-11eb-863d-53043f391f71.png)
Setup tasks only - Setup card can't be collapased (can be dismissed.) 

In the case there are orders that come in while the setup tasks are still incomplete the orders would show below the setup task card.

![Home - Setup tasks with orders](https://user-images.githubusercontent.com/5121465/97031806-71f8bf80-1526-11eb-9fa6-431ddf0d109a.png)
Setup tasks with orders. 

![Home - All agenda items - Orders expanded](https://user-images.githubusercontent.com/5121465/97031467-f5fe7780-1525-11eb-9d4d-8985feaf07e7.png)
Orders expanded with other task types collapsed

[...]

PR #5552, opened 2020-11-06
Don't show the Orders panel on the homescreen with the Task List

jeffstieler:
This PR is a result of a Slack conversation - we decided that the Orders panel should not be shown if the store has yet to receive an order.

- On a store with no orders, verify the Orders panel is not visible
- Either place an order, or switch to a test store with orders
- Verify the Orders panel is visible

### Changelog Note:

Tweak: hide the Orders panel before the store has received orders.

### DIFF
--- a/src/Features/Homescreen.php
+++ b/src/Features/Homescreen.php
@@ -44,6 +44,7 @@ class Homescreen {
 		add_action( 'admin_head', array( $this, 'update_link_structure' ), 20 );
 		add_filter( 'woocommerce_admin_plugins_whitelist', array( $this, 'get_homescreen_allowed_plugins' ) );
 		add_filter( 'woocommerce_admin_preload_options', array( $this, 'preload_options' ) );
+		add_filter( 'woocommerce_shared_settings', array( $this, 'component_settings' ), 20 );
 	}
 
 	/**
@@ -133,5 +134,18 @@ class Homescreen {
 		$options[] = 'woocommerce_default_homepage_layout';
 
 		return $options;
+	}
+
+	/**
+	 * Add data to the shared component settings.
+	 *
+	 * @param array $settings Shared component settings.
+	 */
+	public function component_settings( $settings ) {
+		$allowed_statuses       = array( 'pending', 'processing', 'completed' );
+		$status_counts          = array_map( 'wc_orders_count', $allowed_statuses );
+		$settings['orderCount'] = array_sum( $status_counts );
+
+		return $settings;
 	}
 }
--- a/src/Features/Homescreen.php
+++ b/src/Features/Homescreen.php
@@ -142,8 +142,12 @@ class Homescreen {
 	 * @param array $settings Shared component settings.
 	 */
 	public function component_settings( $settings ) {
-		$allowed_statuses       = array( 'pending', 'processing', 'completed' );
-		$status_counts          = array_map( 'wc_orders_count', $allowed_statuses );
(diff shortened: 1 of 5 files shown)


jeffstieler:
Perhaps a better alternate implementation would be to check the orders count instead?

pmcpinto:
> Perhaps a better alternate implementation would be to check the orders count instead?

I think it would make sense to check if the order count is 0, instead of relying on the task list state. @elizaan36 what do you think?

elizaan36:
hey y'all, it's not a good idea to hide the orders panel depending on when the task list is visible. 
1. We can't assume that people will have completed every task before receiving their first order.
2. In the near future, we're introducing extensibility to the task list so that 3PD's will be able to add tasks. Merchants could have a fully active store will regular orders by the time they realize they need an extension.

In the [design I shared in a comment on #5243](https://github.com/woocommerce/woocommerce-admin/issues/5243#issuecomment-715459610), I had suggested a tweak to the logic - that the orders panel would be hidden until the merchant receives their first order. After that, they'll see the empty state once orders have been fulfilled. This feels like a stronger direction to go in from a user perspective - what do you think @jeffstieler ?

samueljseay:
+1 for waiting for first order, that makes way more sense.

src/Features/Homescreen.php
```diff
@@ -134,4 +135,17 @@ public function preload_options( $options ) {
...
+
+	/**
+	 * Add data to the shared component settings.
+	 *
+	 * @param array $settings Shared component settings.
+	 */
+	public function component_settings( $settings ) {
+		$allowed_statuses       = array( 'pending', 'processing', 'completed' );
+		$status_counts          = array_map( 'wc_orders_count', $allowed_statuses );
+		$settings['orderCount'] = array_sum( $status_counts );
```
louwie17, line 147: Should the allowed statuses also contain cancelled, refunded, and failed? Maybe also on hold?

louwie17 (commented):
Left one comment in relation to the allowed statuses list, otherwise the code looks good and tested well.
Also great job on the unit tests 🤩

src/Features/Homescreen.php
elizaan36: In case this factors in, there's a setting where users can define their own "actionable" order statuses so I think it would be a possibility that they could define Cancelled, Refunded, and Failed statuses as actionable to appear in the Orders panel.

By the way @pmcpinto, Are you seeing those settings when you test the new nav? They're supposed to be in *Analytics > Settings*.

jeffstieler: Hmm, you're right - this is too narrow a list of order statuses to check. How about we just look for any orders that aren't trashed or draft (used by the checkout block)?

jeffstieler: Updated in a96f21d.

elizaan36: sounds good

pmcpinto: > By the way @pmcpinto, Are you seeing those settings when you test the new nav? They're supposed to be in Analytics > Settings.

@elizaan36 yes: https://d.pr/i/lYE8RS

joshuatf (approved):
Testing well and LGTM!  🚢

timmyc:
cherry picked to `release/1.7.3` in https://github.com/woocommerce/woocommerce-admin/commit/095d039272cd775e4e445f1fa9d86499712c58fe

PR #5558, opened 2020-11-09
Post 1.7.0 beta release chores

samueljseay:
This is my first time updating these things, so I'd appreciate someone more experienced taking a look.

Changes:

* Update dev version to `1.8.0-dev` in prep for 1.7.0 final.
* Update changelog to represent the changes released in 1.7.0

### Changelog Note:

n/a


bin/travis.sh
```diff
@@ -4,6 +4,11 @@
 if [ "$1" == 'before' ]; then
 	cd "$WP_CORE_DIR/wp-content/plugins/woocommerce-admin/"
 	npm run build:feature-config
+<<<<<<< HEAD
```
joshuatf, line 7: Conflict here.

joshuatf (commented):
I think we'll also need to run `npm run bump-version` to fix up the version in `FeaturePlugin.php`, but otherwise looks pretty good unless anyone with more experience here wants to chime in.

bin/travis.sh
samueljseay: 😱

samueljseay: Fixed in https://github.com/woocommerce/woocommerce-admin/pull/5558/commits/27d3fcd04c08cafe4237ef65f9143c3534d08861 I also added some new context to that comment for the next person to know when they can remove that code.

samueljseay:
Thanks for the review @joshuatf I have fixed the conflict and I've run `bump-version`

joshuatf (approved):
Nice!  Thanks for that comment as well.  LGTM 🚀

PR #5564, opened 2020-11-09
Navigation: Fix homescreen id on client

psealock:
The homescreen id on page registration changed from `woocommerce-home` to `home`. This made the router aware link in the Nav stop working because a mismatched slot/fill.

1. Turn on the new Nav `woocommerce_navigation_enabled` set to "yes" if you haven't already.
1. Visit any Analytics report or wc-admin page other than Homescreen.
2. Click on "Home" in the Nav.
3. See the page navigate without a page refresh.

### Changelog Note:

none: Change to unreleased feature


joshuatf (approved):
Thanks for the fix up here, Paul!  LGTM.

PR #5540, opened 2020-11-05
Show CES survey after editing an order

mattsherman:
Fixes #5347 

This PR triggers the display of the Customer Effort Score survey modal after editing an order. It only displays if tracking is enabled and the survey has not already been shown.

### Detailed test instructions:

Note: Depending on other outstanding PRs, you may need to make sure that...

- Tracking is enabled: `/wp-admin/admin.php?page=wc-settings&tab=advanced&section=woocommerce_com`
- The options tracking whether the survey has been shown is cleared out: `delete from wp_options where option_name = 'woocommerce_ces_shown_for_actions'` executed

Steps:

- Enable the debugging of Tracks events in your browser console: `localStorage.setItem( 'debug', 'wc-admin:tracks' );`
- Add an order (checkout from cart) on store frontend of site.
- Click on an order in wp-admin.
- Verify that the CES survey snackbar is not shown.
- Go back to the list of orders.
- Bulk modify some orders.
- Click on an order in wp-admin.
- Verify that the CES survey snackbar is not shown.
- Go back to the list of orders.
- Edit an order in wp-admin.
- Verify the CES survey snackbar is displayed.
- Click "Provide Feedback" in the snackbar.
- Verify that the "How easy was it to update an order?" survey modal is shown (how this modal looks will depend on what other PRs related to this have been merged; this PR is not concerned with how the modal looks)
- Verify that if you complete the survey (again, how this is done depends on the state of other open PRs), a Tracks event for `wcadmin_ces_feedback` is logged, with:
    - `action` set to `shop_order_update` 
    - `order_count` set to the number of orders in the store

Other things to test:

- Make sure that the CES modal is shown as expected when adding and editing Products, according to the test instructions in #5352

### Changelog Note:

No changelog note required, as this is part of the larger Customer Effort Score feature being added.

PR #5569, opened 2020-11-09
Update/5348 trigger ces on changing setting

### LINKED ISSUE #5348: Customer effort score - Trigger on changing a setting
Part of #5259

Trigger the customer effort score modal when changing a setting - on the existing `wcadmin_settings_change` tracks event.

Copy should be `How easy was it to update your settings?`

moon0326:
Fixes #5348 

This PR triggers the CES survey when settings get updated.

- Added a base unit test class for the upcoming changes.

### Screenshots
Screen Shot 2020-11-09 at 10 21 19 AMth_

1. Navigate to WooCommerce -> Settings
2. Change a value
3. Click [ Save Changes ]
4. You should see the snackbar at the bottom left corner on the next page.

To repeat the test, delete `woocommerce_ces_shown_for_actions` and `woocommerce_ces_tracks_queue` from the `wp_options` table.


tests/features/class-wc-tests-ces-tracks.php
```diff
@@ -0,0 +1,49 @@
+<?php
+/**
+ * Onboarding Themes Tests.
+ *
+ * @package WooCommerce\Admin\Tests\OnboardingThemes
```
mattsherman, line 5: Comments need to be updated! :-)

moon0326: @mattsherman Thank you for the catch :) Updated it with https://github.com/woocommerce/woocommerce-admin/pull/5569/commits/103ff9682d7074fe0b3e25c33b738794c4d97025

moon0326:
Looking into the failed test.

src/Features/CustomerEffortScoreTracks.php
```diff
@@ -40,17 +40,44 @@ class CustomerEffortScoreTracks {
...
 	public function __construct() {
+		add_action(
+			'admin_init',
+			array(
+				$this,
+				'maybe_clear_ces_tracks_queue',
+			)
+		);
+
+		$this->add_actions();
```
mattsherman, line 68: Is there a reason why we wouldn't want 'admin_init' to also be setup in `add_actions`? And in particular, only is tracking is allowed?

If we do want to keep this separate, maybe `add_actions` should be renamed something like `maybe_enable_survey_enqueing` or `enable_survey_enqueing_if_tracking_is_enabled`.

As it stands, it feels a little confusing/unorganized that we add an action handler and then call another function that says it too is adding actions.

src/Features/CustomerEffortScoreTracks.php
```diff
@@ -40,17 +40,44 @@ class CustomerEffortScoreTracks {
...
+	}
+
+	/**
+	 * Add actions that require woocommerce_allow_tracking.
+	 */
+	private function add_actions() {
+		// Only enqueue a survey if tracking is allowed.
+		$allow_tracking = 'yes' === get_option( 'woocommerce_allow_tracking', 'no' );
```
mattsherman, line 76: I like this optimization! No need to set up the action handlers if we aren't ever going to do anything in them. Nice!

src/Features/CustomerEffortScoreTracks.php
```diff
@@ -242,4 +266,26 @@ function( $item ) use ( $clear_ces_tracks_queue_for_page ) {
...
+
+		$this->enqueue_to_ces_tracks(
+			array(
+				'action'    => self::SETTINGS_CHANGE_ACTION_NAME,
+				'label'     => __(
+					'How easy was it to update your settings?',
+					'woocommerce-admin'
+				),
+				'pagenow'   => 'woocommerce_page_wc-settings',
+				'adminpage' => 'woocommerce_page_wc-settings',
```
mattsherman, line 286: Hmm... do we have anywhere where we document how `pagenow` and `adminpage` should be used? I sorta just copy-pasted my settings in #5540, but seeing how you have set these, it makes me question

(a) what they should be set to, and 
(b) if we really need both `pagenow` and `adminpage`

/cc @becdetat

tests/features/class-wc-tests-ces-tracks.php
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @package Automattic\WooCommerce\Admin\Features
+ */
+
+use \Automattic\WooCommerce\Admin\Features\CustomerEffortScoreTracks;
+
+/**
+ * Class WC_Tests_CES_Tracks
+ */
+class WC_Tests_CES_Tracks extends WC_Unit_Test_Case {
```
mattsherman, line 13: Thanks for including some tests!

mattsherman (changes requested):
This looks like a great start. I left a few minor comments.

Works as expected.

src/Features/CustomerEffortScoreTracks.php
moon0326: I think we can include admin_init in the add_actions as well. I assumed that we need to run maybe_clear_ces_tracks_queue on every call, but it looks like that's not the case after looking at the code again. In that case, I think we can remove add_actions and move the code back to the constructor. I totally agree that add_actions is not the best name for the method.

src/Features/CustomerEffortScoreTracks.php
moon0326: Looking at the https://github.com/woocommerce/woocommerce-admin/blob/update/5348-trigger-ces-on-changing-setting/client/customer-effort-score-tracks/customer-effort-score-tracks-container.js#L34, those two variables are used to filter the queue. The values can be retrieved from `window.pagenow` and `window.adminpage`.

tests/features/class-wc-tests-ces-tracks.php
moon0326: Sure thing! I will add more tests once we merge other PRs together.

src/Features/CustomerEffortScoreTracks.php
mattsherman: Thanks!

mattsherman (approved):
Thanks for making the updates @moon0326! This still works as expected.

Nice job on this PR!

PR #5578, opened 2020-11-10
Refactoring logic handling auto-folding navigation to support resize …

### LINKED ISSUE #5557: Navigation: Collapse menu on resize event
The functionality to collapse the menu on small viewports should function on screen resize.

### Deliverables

* Resizing the screen should collapse/uncollapse the navigation

joelclimbsthings:
Fixes #5557 

Right now the navigation will auto-collapse in response to an `orientationchange` event if the viewport is less than 960px. According to the referenced issue we need to apply this logic to the resize event as well. This change will also add behavior that will auto-extend the nav when the viewport is resized to be greater than the mentioned threshold. 

- Check out branch
- Ensure the navigation feature is enabled
- Navigate to Wp Admin -> WooCommerce to view the new navigation
- With the navigation displayed, resize the viewport to less than 960px
- Ensure that the navigation collapses.
- Resize the viewport again to be greater than 960px
- Ensure that the navigation extends.

### DIFF
--- a/client/navigation/components/header/index.js
+++ b/client/navigation/components/header/index.js
@@ -8,6 +8,7 @@ import { getSetting } from '@woocommerce/wc-admin-settings';
 import { useSelect } from '@wordpress/data';
 import { useEffect } from 'react';
 import classnames from 'classnames';
+import { debounce } from 'lodash';
 
 /**
  * Internal dependencies
@@ -23,7 +24,7 @@ const Header = () => {
 		document.body.classList.toggle( 'is-folded' );
 	};
 
-	const foldOnMobile = ( screenWidth ) => {
+	const foldOnMobile = ( screenWidth = document.body.clientWidth ) => {
 		if ( screenWidth <= 960 ) {
 			document.body.classList.add( 'is-folded' );
 		}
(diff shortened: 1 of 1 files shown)


octaedro (approved):
Nice solution @joelclimbsthings, clean and simple! This is testing well here and code looks great.
I only left a small question, but since it's working well I'll preapprove this PR.

client/navigation/components/header/index.js
```diff
@@ -23,20 +24,30 @@ const Header = () => {
...
-		foldOnMobile( document.body.clientWidth );
+		foldOnMobile();
+		const foldEvents = [
+			{
+				eventName: 'orientationchange',
+				handler: ( e ) => foldOnMobile( e.target.screen.availWidth ),
+			},
+			{
+				eventName: 'resize',
+				handler: debounce( () => foldOnMobile(), 200 ),
```
octaedro, line 44: Is there a reason to use the `debounce ` here?

joelclimbsthings: Just a small performance tweak. While dragging the resize handle on the app, the browser will rapid fire the resize event. Since we're really only interested in executing the callback after we're done resizing we can enforce that with `debounce()`

octaedro:
Cherry picked to release/1.7.0 in f334183

PR #5581, opened 2020-11-10
Fix a bug where welcome modal would briefly display even when dismissed.

### LINKED ISSUE #5555: Home Screen: Welcome Modal Flashes Briefly After Dismissed
**Describe the bug**
While testing out pre-beta of 1.7.0 with Woo 4.6.2 and Gutenberg on a new JN site ( Safari used ) - I was noticing that when doing a hard refresh on the Home Screen, the welcome dialog is briefly shown/flashed during the page load:

https://cloudup.com/c-QiawR21Fg

**To Reproduce**

1. Do OBW, and dismiss the Welcome Modal on the home screen
2. Hard refresh the page and note the welcome modal is briefly shown.

**Expected behavior**
To not see the welcome modal after it has been dismissed.

samueljseay:
Fixes #5555 

An explanation of what happens currently:

1. we call `getOption`
2. straight after that we call `isResolving` to check if that option is loading.

I had assumed that the way `@wordpress/data` works is, that calling a selector would mean that `isResolving` for the backing resolver is immediately and synchronously `true`, but debugging the code this is not the case. Instead `isResolving` is false for a short while.

The simplest way to fix this was to use `hasFinishedResolution` instead. This will only return true once the option is actually loaded. 

**As a side note, we use this pattern a lot in our code base so it might be a good time to look at fixing that.

I have [raised a discussion about this](p1604982087129300-slack-C0926NEBC). If this is indeed an anti-pattern I'll raise an issue to fix it throughout the code base.**

@timmyc had reproduced this in a JN site, and I'm not sure exactly how I triggered it locally, but I had the following setup:

1. Gutenberg installed
2. Woocommerce 4.8.0 dev installed
3. Complete OBW so you can view homescreen, dismiss the welcome modal. (testing that the welcome modal still displays as normal is important here too!)
4. refresh the homescreen page
5. the modal should **not** flicker briefly on screen.

### Changelog Note:

Fix: a bug where the welcome modal briefly displays when loading the home screen.

### DIFF
--- a/client/homescreen/layout.js
+++ b/client/homescreen/layout.js
@@ -144,18 +144,21 @@ Layout.propTypes = {
 export default compose(
 	withSelect( ( select ) => {
 		const { isNotesRequesting } = select( NOTES_STORE_NAME );
-		const { getOption, isResolving } = select( OPTIONS_STORE_NAME );
+		const { getOption, isResolving, hasFinishedResolution } = select(
+			OPTIONS_STORE_NAME
+		);
 
 		const welcomeModalDismissed =
 			getOption( 'woocommerce_task_list_welcome_modal_dismissed' ) ===
 			'yes';
 
-		const welcomeModalDismissedIsResolving = isResolving( 'getOption', [
-			'woocommerce_task_list_welcome_modal_dismissed',
-		] );
+		const welcomeModalDismissedHasResolved = hasFinishedResolution(
+			'getOption',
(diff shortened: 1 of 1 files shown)


jeffstieler (approved):
While I wasn't able to truly reproduce this issue locally, I tested using the Redux DevTools and stepping through the dispatched actions. This looks good!

jeffstieler:
Cherry picked into `release/1.7.0`.

PR #5583, opened 2020-11-10
Fix reference to navigation package and fix transient notices so it l…

### LINKED ISSUE #5553: Setup Checklist: Snackbar Notif & Setup Activity Panel Missing after Product Creation
**Describe the bug**
When performing the Product setup task, I encountered some JS errors on Safari 14:

1 7 0-beta-safari-14 0

**To Reproduce**
I hit this while testing a build from `main` before the release of the beta, but it should be reproducible in the beta release. This was done on a new JN site running Woo 4.6.2 and the test zip.

1. After the OBW, visit the Product Setup step
2. Manually create a product, save it
3. On the screen refresh, note that the SnackBar notification to send me back to the home screen is not shown, and the Activity Panels for store setup is missing
4. See error in the JS console

**Expected behavior**
I expected to see the SnackBar notice to take me back to the Setup checklist, and the Activity Panel for Store Setup too.

**Other Details**
I did have the latest Gutenberg plugin installed, not sure if that impacts this or not.

becdetat:
…ooks at core/notices as well as core/notices2

NOTE this should *not* be cherry-picked, it includes a fix that isn't required (or possible) in the 1.7.0 cut. I'll make another PR for the cherry-pick.

Fixes #5553

There's two parts to this. For some reason including `wc-navigation` as a dependency to the script included when the first product is added causes seemingly non-related errors in Safari only, which killed the activity panel as well as the notification. The second part is that even after fixing the dependencies, the snackbar wasn't appearing due to the fork [added a couple of days ago](https://github.com/woocommerce/woocommerce-admin/pull/5532). This was fixed by referencing both data stores - `core/notices` and the newly added `core/notices2`.

Note that this second issue would have been affecting all existing usages of `core/notices` (only since Monday).

- In Safari,
- delete all of your products. Yes even the ones you like.
- make sure your onboarding task list is visible on the WooCommerce home screen
- "Add my products"
- "Add manually"
- Add a product.
- When the page refreshes after adding the product the success message should appear and the activity panel should still be visible.

### Changelog Note:

Fix: First product script navigation dependency
Fix: Reference both core/notices and core/notices2

### DIFF
--- a/client/layout/transient-notices/index.js
+++ b/client/layout/transient-notices/index.js
@@ -46,10 +46,11 @@ TransientNotices.propTypes = {
 export default compose(
 	withSelect( ( select ) => {
 		// NOTE: This uses core/notices2, if this file is copied back upstream
-		// to Gutenberg this needs to be changed back to core/notices.
-		const notices = select( 'core/notices2' ).getNotices();
+		// to Gutenberg this needs to be changed back to just core/notices.
+		const notices = select( 'core/notices' ).getNotices();
+		const notices2 = select( 'core/notices2' ).getNotices();
 
-		return { notices };
+		return { notices: notices.concat( notices2 ) };
 	} ),
 	withDispatch( ( dispatch ) => ( {
 		// NOTE: This uses core/notices2, if this file is copied back upstream
--- a/src/Features/OnboardingTasks.php
+++ b/src/Features/OnboardingTasks.php
@@ -241,7 +241,7 @@ class OnboardingTasks {
 			return;
 		}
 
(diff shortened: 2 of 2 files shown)


samueljseay (approved):
Question: would you agree that it was erroneous to have wc-navigation as a dep to that script? I couldn't see a good reason for it to be included but just wanted to confirm?

Anyway the changes test well. 🎉

becdetat:
@samueljseay 

> would you agree that it was erroneous to have wc-navigation as a dep to that script?

Yesssss. It seemed incorrect to me, I couldn't see why it was there and I couldn't find a downside to removing it. Finding it was really a process of elimination though, I'm not sure why removing it would fix the errors.

PR #5587, opened 2020-11-10
Change select control padding to match text-control inputs

### LINKED ISSUE #4254: OBW: `Country / Region` select has more left padding than the other input fields
**Describe the bug**

Filing this bug on behalf of @Aljullu. Reference: pb22l9-8V-p2#comment-398. 

In the section `Where is your store based?`, the `Country / Region` select seems to have more left padding than the other input fields. See how the amplifier is not aligned with the first letter of the text inputs.

profile

**To Reproduce**

1. Start new OBW;
2. Observe the issue on the `Where is your store based?` section:

Screen Shot 2020-04-29 at 2 40 13 PM

**Expected behavior**

The amplifier should be aligned with the rest of the first letter inputs. 

**Desktop:**

I tested as per below and reproduced the same:

* OS: macOS Catalina 10.15.4
* Browser: Chrome
* Version: 81.0.4044.122 (Official Build) (64-bit)

**Additional context**

@Aljullu tested on Firefox 77 under Fedora Linux.

louwie17:
Fixes #4254 

Updated the select control padding to use `$gap-small` for left/right padding (right is actually being overwritten already) to match the text control padding (the search control uses the same padding).

Updated the label `left` by `4px` to make up for the changed padding, this way the label also aligns with the input field.

### Screenshots

Screen Shot 2020-11-10 at 12 22 48 PM

- Have WooCommerce and WooCommerceAdmin enabled
- Go to the Woo setup wizard, can access using this url -> `/wp-admin/admin.php?page=wc-admin&path=%2Fsetup-wizard`
- The magnifying glass for Country/Region field should align with the other text fields now.

### DIFF
--- a/packages/components/src/select-control/style.scss
+++ b/packages/components/src/select-control/style.scss
@@ -1,5 +1,3 @@
-
-
 .woocommerce-select-control {
 	position: relative;
 
@@ -10,7 +8,7 @@
 		border: 1px solid $studio-gray-20;
 		border-radius: 3px;
 		background: $studio-white;
-		padding: $gap-small $gap;
+		padding: $gap-small;
 		position: relative;
 
 		.woocommerce-select-control__tags {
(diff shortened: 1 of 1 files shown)


samueljseay (approved):
Looks great, testing well. 🚀

PR #5590, opened 2020-11-10
Home screen - Fixed the orders not displaying bug

### LINKED ISSUE #5567: Home screen - orders task not displaying orders
**Describe the bug**
I have a couple of unfulfilled orders in my test store. However, the orders task doesn't display any order. And the order count doesn't seem to be accurate as well.

**To Reproduce**

1. Install wcadmin 1.7 in an existing store with unfulfilled orders
2. Visit the home screen
3. Check that the orders task is empty and the order count is not correct.
4. Add a new purchase to your store
5. Check that the orders task remains empty

**Expected behavior**
To see a list of orders with actionable status that I defined in the settings: /wp-admin/admin.php?page=wc-admin&path=%2Fanalytics%2Fsettings
To see the order count including only the number of unfulfilled orders

**Screenshots**
What I'm seeing: https://d.pr/i/jgaFnx
A screenshot showing that I have unfulfilled orders: https://d.pr/i/EHQKW0

octaedro:
Fixes #5567

This PR fixes the orders not displaying bug.

### Screenshots
Screen Capture on 2020-11-10 at 15-20-46

- Go to `Analytics` > `Settings` and select all the checkboxes under `Actionable Statuses` (**don't select unregistered statuses**).

screenshot-one wordpress test-2020 11 10-15_08_26

- Go to the `Home` screen. Don't refresh the page in order to not lose the `Analytics > Settings` selection.
- Confirm that the total orders number is right.

screenshot-one wordpress test-2020 11 10-15_14_30

- Go back to the Analytics settings and select only one status (or neither).
- Go to `Home` screen and verify that the orders panel has the selected ones.

### Changelog Note:
Fix: Home screen - Fixed the orders not displaying bug

### DIFF
--- a/client/homescreen/activity-panel/orders/index.js
+++ b/client/homescreen/activity-panel/orders/index.js
@@ -304,12 +304,8 @@ export default withSelect( ( select, props ) => {
 		REPORTS_STORE_NAME
 	);
 
-	if ( countUnreadOrders === null ) {
-		return { isRequesting: true };
-	}
-
-	if ( countUnreadOrders === 0 ) {
-		return { isRequesting: false };
+	if ( ! countUnreadOrders ) {
+		return { isRequesting: countUnreadOrders !== 0 };
 	}
 
 	// Query the core Orders endpoint for the most up-to-date statuses.
@@ -344,6 +340,7 @@ export default withSelect( ( select, props ) => {
 		per_page: 5,
 		extended_info: true,
 		order_includes: map( actionableOrders, 'id' ),
+		status_is: orderStatuses,
 		_fields: [
 			'order_id',
 			'order_number',
--- a/client/homescreen/activity-panel/orders/index.js
+++ b/client/homescreen/activity-panel/orders/index.js
@@ -346,7 +346,6 @@ export default withSelect( ( select, props ) => {
 			'order_id',
 			'order_number',
 			'status',
-			'data_created_gmt',
 			'total_sales',
 			'extended_info.customer',
 			'extended_info.products',
--- a/client/homescreen/activity-panel/orders/index.js
+++ b/client/homescreen/activity-panel/orders/index.js
@@ -339,6 +339,7 @@ export default withSelect( ( select, props ) => {
 		page: 1,
 		per_page: 5,
 		extended_info: true,
+		match: 'any',
 		order_includes: map( actionableOrders, 'id' ),
 		status_is: orderStatuses,
 		_fields: [
--- a/client/homescreen/activity-panel/orders/utils.js
+++ b/client/homescreen/activity-panel/orders/utils.js
@@ -14,7 +14,7 @@ export function getUnreadOrders( select, orderStatuses ) {
 	);
 
 	if ( ! orderStatuses.length ) {
-		return false;
+		return 0;
 	}
 
 	const ordersQuery = {


client/homescreen/activity-panel/orders/index.js
```diff
@@ -344,6 +340,7 @@ export default withSelect( ( select, props ) => {
 		per_page: 5,
 		extended_info: true,
 		order_includes: map( actionableOrders, 'id' ),
+		status_is: orderStatuses,
```
samueljseay, line 343: Should this be `statuses:`?

client/homescreen/activity-panel/orders/index.js
```diff
@@ -304,12 +304,8 @@ export default withSelect( ( select, props ) => {
 		REPORTS_STORE_NAME
 	);
 
-	if ( countUnreadOrders === null ) {
-		return { isRequesting: true };
-	}
-
-	if ( countUnreadOrders === 0 ) {
-		return { isRequesting: false };
+	if ( ! countUnreadOrders ) {
```
samueljseay, line 307: so to clarify, the conditions are:

1. countUnreadOrders === null,  isRequesting: false
2. countUnreadOrders === 0, isRequesting: false
3. countUnreadOrders !== null && countUnreadOrders > 0, isRequesting: true

Is that right? I would have thought that `null` would mean that countUnreadOrders has not resolved yet here?

samueljseay (commented):
Just a couple questions from me to make sure I understand.

client/homescreen/activity-panel/orders/index.js
octaedro: Hi @samueljseay, thank you for your review. Inside the `src/API` the variable name is: [status_is](https://github.com/woocommerce/woocommerce-admin/blob/fix/5567/src/API/Reports/Orders/Controller.php#L58). If I use `statuses` it won't work as expected.

client/homescreen/activity-panel/orders/index.js
octaedro: When `countUnreadOrders` is resolving, it's value will be `null`. So when `! countUnreadOrders` is `true`, we ask if `countUnreadOrders` is not 0 in order to know that it's resolving.

octaedro: So the possible paths are:
```
if ( ! countUnreadOrders ) {
     return { isRequesting: countUnreadOrders !== 0 };
}
```
1. countUnreadOrders === null, isRequesting: `true`
2. countUnreadOrders === 0, isRequesting: `false`
3. countUnreadOrders !== null && countUnreadOrders > 0, it shouldn't get into the `if`

client/homescreen/activity-panel/orders/index.js
samueljseay: Thats all good was just checking if it was a typo or not :+1:

client/homescreen/activity-panel/orders/index.js
samueljseay: Ah you're right, I misread the logic. 🤦 looking good

samueljseay (approved):
This is testing well 🚀  great stuff fixing this

octaedro:
Cherry picked to release/1.7.0 in 7a33e7e

PR #5594, opened 2020-11-11
Trigger CES survey when a user performs an action on analytic pages.

### LINKED ISSUE #5349: Customer effort score: Trigger when filtering in Analytics
Part of #5259

Trigger the customer effort score modal when filtering a report within Analytics.

Copy should be `How easy was it to filter your store analytics?`

moon0326:
Fixes #5349 

This PR triggers Customer Effort Score survey when a user performs an action on one of the pages in Analytics.

- updated [report-filters](https://github.com/woocommerce/woocommerce-admin/blob/68d9b02506b53bf05eed273f86c4a28f720c264a/client/analytics/components/report-filters/index.js) component to emit `woocommerce.admin.analytics.filtered` event via `wp.hooks.doAction` when a user changes filters.
- added a new container component [CustomerEffortScoreTracksContainerAnalytics](https://github.com/woocommerce/woocommerce-admin/blob/68d9b02506b53bf05eed273f86c4a28f720c264a/client/customer-effort-score-tracks/customer-effort-score-tracks-container-analytics.js) to listen for the `woocommerce.admin.analytics.filtered` event and then render [CustomerEffortScoreTracks](https://github.com/woocommerce/woocommerce-admin/blob/68d9b02506b53bf05eed273f86c4a28f720c264a/client/customer-effort-score-tracks/customer-effort-score-tracks.js) conditionally.

With my limited experience in React and the backend codebase, this is the most optimized approach I could find. I would really appreciate any feedback on the implementation 👍 

### Screenshots
Screen Shot 2020-11-10 at 6 08 12 PM

Make sure to delete both `woocommerce_ces_shown_for_actions` and `woocommerce_ces_tracks_queue` from the `wp_options`. Ensure that you have tracking enabled on WooCommerce > Settings > WooCommerce.com.

1. Navigate to one of the pages in Analytics.
2. Change `Date Range:` and click [ Update ] button. You can also use the advanced, single product, and comparison filters.
3. You should see the survey at the bottom left corner. Click the "Give feedback" text and confirm that the modal pops up.
4. After the survey snackbar disappears, keep changing the filters and make sure that the survey does not popup again.
5. Refresh the page and repeat the steps.
6. Make sure the survey does not pop up again after the page refresh.


client/customer-effort-score-tracks/customer-effort-score-tracks-container-analytics.js
```diff
@@ -0,0 +1,95 @@
...
+	};
+
+	// Listen for the action
+	wp.hooks.addAction(
+		ACTION_NAME,
+		'woocommerce.admin',
+		onAnalyticFilterChanged
+	);
+
+	function getComponent() {
```
jeffstieler, line 62: I think we can remove this function definition 😄 .

client/customer-effort-score-tracks/customer-effort-score-tracks-container-analytics.js
```diff
@@ -0,0 +1,95 @@
...
+		return (
+			<CustomerEffortScoreTracks
+				key={ 999 }
```
jeffstieler, line 65: I don't think this key is necessary. Here's some background info on [when to use React keys](https://reactjs.org/docs/reconciliation.html#recursing-on-children).

client/customer-effort-score-tracks/customer-effort-score-tracks-container-analytics.js
```diff
@@ -0,0 +1,95 @@
...
+				action={ ACTION_NAME }
+				label={ __(
+					'How easy was it to filter your store analytics?',
+					'woocommerce-admin'
+				) }
+				onSubmitLabel={ __(
+					'Thank you for your feedback!',
+					'woocommerce-admin'
+				) }
+				trackProps={ {} }
```
jeffstieler, line 75: If this was causing an error in the `<CustomerEffortScoreTracks>` component, perhaps adding a default prop value there would be better?

client/index.js
```diff
@@ -91,4 +94,10 @@ if ( navigationRoot ) {
 		<CustomerEffortScoreTracksContainer />,
 		root.insertBefore( document.createElement( 'div' ), null )
 	);
+
+	// Used to display CES Survey on the Analytic pages
+	render(
+		<CustomerEffortScoreTracksContainerAnalytics />,
+		root.insertBefore( document.createElement( 'div' ), null )
+	);
```
jeffstieler, line 102: I'm curious to know if there's another place to render this component. This is mostly coming from an aesthetic standpoint, but my leaning is to question the addition of more `render()` calls. 🙃

jeffstieler (changes requested):
Overall this PR is testing well and the code looks pretty good! Nice work @moon0326.

I do want to call out that this would be the first introduction of the `doAction()` hook into our client side code. I think we should have more than a single person's (read: mine) buy in for such a change - pinging leads @timmyc @becdetat @samueljseay @psealock for discussion.

client/customer-effort-score-tracks/customer-effort-score-tracks-container-analytics.js
moon0326: Thank you for the catch 👍 Removed with the latest commit.

client/customer-effort-score-tracks/customer-effort-score-tracks-container-analytics.js
moon0326: Removing it did not cause an error.  Updated with the latest commit 👍

client/customer-effort-score-tracks/customer-effort-score-tracks-container-analytics.js
moon0326: Totally agreed 😄. Updated with the latest commit.

client/customer-effort-score-tracks/customer-effort-score-tracks-container-analytics.js
```diff
@@ -0,0 +1,95 @@
...
+	/**
+	 * Function to get WordPress options.
+	 */
+	getOption: PropTypes.func,
+};
+
+export default compose(
+	withSelect( ( select ) => {
+		const { getOption } = select( OPTIONS_STORE_NAME );
+		return { getOption };
```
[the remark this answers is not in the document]
moon0326: Thank you for the link. I updated it with the latest commit.

client/analytics/components/report-filters/index.js
```diff
@@ -63,6 +71,10 @@ class ReportFilters extends Component {
 				} );
 				break;
 			case 'filter':
+				wp.hooks.doAction(
+					'woocommerce.admin.analytics.filtered',
+					'analytics_filters_filter'
```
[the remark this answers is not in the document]
moon0326: We now have the following events. I would appreciate any thoughts on the naming 👍 

- woocommerce_admin_analytics_filter_type_changed (e.g., Changing All Orders -> Advanced Filters) 
- woocommerce_admin_analytics_filter_query_executed (fired when a user clicks 'Filter' after changing some filters)
- woocommerce_admin_analytics_date_range_query_executed (fired when a user changes date range and click the update button)

each event also contains report type and data.

https://github.com/woocommerce/woocommerce-admin/pull/5594/commits/3b316f0eca40d1a3b3c9b3b6776ce68f86dd5e42#diff-f1b1e7c2a3e157c5d4d81e386c29bf0b82056037dcf8e432317988224df3a686R36

client/analytics/components/report-filters/index.js
```diff
@@ -39,6 +43,10 @@ class ReportFilters extends Component {
 	}
 
 	trackFilterSelect( data ) {
```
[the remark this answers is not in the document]
moon0326: Thank you for the suggestion. I changed callback names to onDateSelect, onFilterSelect, and onAdvancedFilterAction.

client/analytics/components/report-filters/index.js
```diff
@@ -31,6 +31,10 @@ class ReportFilters extends Component {
 	}
 
 	trackDateSelect( data ) {
+		wp.hooks.doAction(
```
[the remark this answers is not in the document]
moon0326: Thank you! I was unsure of using a global variable. Importing doAction and addAction worked perfectly.

client/analytics/components/report-filters/index.js
```diff
@@ -39,6 +43,10 @@ class ReportFilters extends Component {
 	}
 
 	trackFilterSelect( data ) {
+		wp.hooks.doAction(
+			'woocommerce.admin.analytics.filtered',
```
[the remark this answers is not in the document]
moon0326: Thank you! Updated with the latest commit.

client/customer-effort-score-tracks/customer-effort-score-tracks-container-analytics.js
```diff
@@ -0,0 +1,95 @@
...
+
+		return false;
+	};
+
+	const onAnalyticFilterChanged = () => {
+		if ( hasShownResultCache === true ) {
+			return;
+		}
+
+		if ( hasShown( ACTION_NAME ) === false ) {
```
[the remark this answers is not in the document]
moon0326: Thank you for the catch. I removed it with the latest commit.

client/index.js
moon0326: I think we need input from @becdetat  😄

client/customer-effort-score-tracks/customer-effort-score-tracks-container-analytics.js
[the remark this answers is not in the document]
moon0326: That's a good idea. Thank you! With the change, I also removed `hasShown()` method as it is unnecessary. The function is a bit cleaner and simpler now :) https://github.com/woocommerce/woocommerce-admin/blob/e7b69e97850ab1230f3882e32c6a59ef4d9dd356/client/customer-effort-score-tracks/customer-effort-score-tracks-container-analytics.js#L36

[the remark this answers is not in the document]
moon0326: @jeffstieler I might be overthinking, but calling `getOption` function within `withSelect` function might be making an unnecessary call (option rest API) to the backend as `CustomerEffortScoreTracksContainerAnalytics` component renders nothing initially. I think it might be better to lazy call `getOption` like the original commit. What do you think?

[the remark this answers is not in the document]
jeffstieler: You are not at all overthinking - that's a nice catch. In other parts of the app we use the [`withOptionsHydration()`](https://github.com/woocommerce/woocommerce-admin/blob/a9acc231e78640a7adf715d486b5780e3c0f39d5/packages/data/src/options/with-options-hydration.js#L12) higher order component to hydrate preloaded options (typically from a global variable) like in [`client/layout/index.js`](https://github.com/woocommerce/woocommerce-admin/blob/d6c80bb709197ea8a4ddd87f517cbde2caca970b/client/layout/index.js#L226-L232).

If you choose to use the hydration approach, you'll need to add the `woocommerce_ces_shown_for_actions` options to the the preloaded options array using the `woocommerce_admin_preload_options` filter. Here's [an example](https://github.com/woocommerce/woocommerce-admin/blob/e89cc8a2ef9d193a376c0e5fb8cf6376e7c86b06/src/Features/Homescreen.php#L46) from the Homescreen feature class.

client/customer-effort-score-tracks/customer-effort-score-tracks-container-analytics.js
```diff
@@ -0,0 +1,95 @@
...
+}
+
+CustomerEffortScoreTracksContainerAnalytics.propTypes = {
+	/**
+	 * Function to get WordPress options.
+	 */
+	getOption: PropTypes.func,
+};
+
+export default compose(
```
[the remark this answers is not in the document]
becdetat: I used a default export for the component file, then exported it named in `index.js`. Mainly because I couldn't figure out how to do a named export with `compose`... if we can do that then all the better.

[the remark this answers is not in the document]
becdetat: I definitely agree we should be using named exports, in `index.js` at the very least

jeffstieler:
To weigh in more on the issue of introducing calls to `doAction()`..

Since we don't currently have any other actions in the client code, this feels like a big change. While I think @moon0326's choice to use hooks for this feature is totally in line with "the WordPress way", I'm not sure this is the spot to introduce extensibility.

Perhaps if we can come up with valuable ways 3PDs could use these hooks for their own purposes, we should keep them. Otherwise I'm in favor of removing the hooks and calling the CES code directly.

becdetat:
@moon0326 

> You can also use the advanced, single product, or comparison filters

I couldn't get this to work. Changing the date filter got the snackbar displaying, but not the advanced filters.

> Make sure the survey does not pop up again after the page refresh

The snackbar should actually keep popping up after a page refresh - it's only when the user opens the CES modal itself that the `woocommerce_ces_shown_for_actions` option would be updated and the snackbar not be displayed for analytics again.

client/index.js
becdetat: I agree, it doesn't seem like something that will scale - it should be possible for third parties to use the CES modal in the same way as we are and adding another container for each use case doesn't seem like it will scale.

I would like to see this done using a store. You can see an example of that in the snackbar implementation itself - there is a `core/notices` store registered in `packages/notices` (actually registered as `core/notices2` since we've forked snackbars from Gutenberg) and consumed in `client/transient-notices`. We could do something similar, registering a `customer-effort-store/surveys` (or similar) store. Then, instead of using an action, getting a survey displayed would be like:

```js
dispatch( 'customer-effort-score/surveys' ).addSurvey(...);
```

becdetat:
@jeffstieler I agree with your opinion around using `doAction`. It is a basically a cross-domain pub/sub, the messy parts of which is what redux is meant to solve. I think using `@wordpress/data` to create a store for this would be more scalable. More here: https://github.com/woocommerce/woocommerce-admin/pull/5594#discussion_r521768068

client/index.js
moon0326: Thank you @becdetat. I will look into those packages and see if I can change how the component gets rendered.

client/customer-effort-score-tracks/customer-effort-score-tracks-container-analytics.js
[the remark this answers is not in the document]
moon0326: @jeffstieler Thank you for the references. I will look into them.

moon0326:
> @moon0326
> 
> > You can also use the advanced, single product, or comparison filters
> 
> I couldn't get this to work. Changing the date filter got the snackbar displaying, but not the advanced filters.
> 
> > Make sure the survey does not pop up again after the page refresh
> 
> The snackbar should actually keep popping up after a page refresh - it's only when the user opens the CES modal itself that the `woocommerce_ces_shown_for_actions` option would be updated and the snackbar not be displayed for analytics again.

That's true. I've updated the test steps.

moon0326:
> To weigh in more on the issue of introducing calls to `doAction()`..
> 
> Since we don't currently have any other actions in the client code, this feels like a big change. While I think @moon0326's choice to use hooks for this feature is totally in line with "the WordPress way", I'm not sure this is the spot to introduce extensibility.
> 
> Perhaps if we can come up with valuable ways 3PDs could use these hooks for their own purposes, we should keep them. Otherwise I'm in favor of removing the hooks and calling the CES code directly.

If that's an option (calling the CES code directly), I think that would make it easier :) 

I tried to decouple by using hooks, but I also think that it is unnecessary at least for now without other solid use cases. I will look into @becdetat 's suggestion mentioned above.

client/index.js
jeffstieler: @becdetat would you be OK with adding a new store for CES notices in a new PR? It seems like we could land this one and then refactor the CES notice creation code.

moon0326: @becdetat @jeffstieler I've pushed a new [commit](https://github.com/woocommerce/woocommerce-admin/pull/5594/commits/c679e4733280d56b59c903f2315533582e2a5988) to call the CES survey directly from the analytics components. It's not the cleanest approach, but it's simple and works for now. I would appreciate some feedback on them please :pray:

client/customer-effort-score-tracks/customer-effort-score-helper.js
```diff
@@ -0,0 +1,37 @@
...
+		return;
+	}
+	cesSurveyForAnalyticsShown = true;
+
+	const appRoot = document.getElementById( 'root' );
+	const embeddedRoot = document.getElementById( 'woocommerce-embedded-root' );
+	const root = appRoot || embeddedRoot;
+
+	render(
+		<CustomerEffortScoreTracks
```
joshuatf, line 24: Pardon my ignorance on this feature as this is my first time diving into this code.

Is there a reason we can't use the existing `core/notices` that are displayed in the `TransientNotices` component?  This takes advantage of the existing stores and doesn't require the addition of new render or mount points.

If we're worried about namespacing the notices, it could be a good time to introduce `woocommerce/notices` as a store.

moon0326: @joshuatf I think it's because we have initialization logic that `core/notices` does not support as shown as [here](https://github.com/woocommerce/woocommerce-admin/blob/update/5349-trigger-ces-on-filtering-analytics/client/customer-effort-score-tracks/customer-effort-score-tracks.js), but I think @becdetat is the best person to answer for the `CustomerEffortScoreTracks` component.

moon0326:
Closing as store implementation is being reviewed in https://github.com/woocommerce/woocommerce-admin/pull/5649

PR #5598, opened 2020-11-11
Package builds: avoid building dependency-extraction-webpack-plugin

psealock:
The dependency-extraction-webpack-plugin wasn't usable because it was being built as an ES6 package, but the code is meant to run in Node and does not need to be transpiled. As a result the starter pack wasn't working.

This PR avoids building dependency-extraction-webpack-plugin and changes the `main` entry point directly to the source. Some eslint and prettier config files are now included in the starter pack. After this, the package will need to be re-released.

1. `npm run build`.
2. `cd packages/dependency-extraction-webpack-plugin`.
3. `npm link`
4. `cd ../..`
5. `npm run create-wc-extension`
6. Follow the instructions and name your new app.
7. `cd ../<new app name>`
8. `npm install`
9. `npm link @woocommerce/dependency-extraction-webpack-plugin`
10. `npm start`
11. The build should start without error.
12. Bonus: If you have prettier and eslint configured in your editor, see that js errors are highlighted and prettier formats correctly.

### Changelog Note:

- Dev: Avoid transpiling dependency-extraction-webpack-plugin package


packages/dependency-extraction-webpack-plugin/package.json
```diff
@@ -16,7 +16,7 @@
 	"bugs": {
 		"url": "https://github.com/woocommerce/woocommerce-admin/issues"
 	},
-	"main": "build/index.js",
+	"main": "src/index.js",
 	"module": "build-module/index.js",
```
samueljseay, line 20: We should remove the module declaration, since that won't function and would confuse someone using this package outside of node.js by accident.

samueljseay, line 20: We should remove the module declaration, since that won't function and would confuse someone using this package outside of node.js by accident.

packages/dependency-extraction-webpack-plugin/package.json
```diff
@@ -16,7 +16,7 @@
...
 	"react-native": "src/index",
```
samueljseay, line 21: This should surely be removed too, I don't think this is intended to be used in a react-native context so it might just add confusion.

samueljseay (approved):
Just a couple little things in the package.json I think we could improve, but this is testing well for me. 

Pre-approving.

PR #5606, opened 2020-11-11
Update non-wca pages registered with nav to embed pages

joshuatf:
Updates pages using the new nav to become embed pages if they are not WCA pages.

Found while registering nav items in external plugins; WCA scripts and styles are not loaded and would need to be manually enqueued otherwise.

Embedding allows us not only to access scripts, but also adds some necessary page styling.  As @psealock noted, the header is also used during mobile/collapsed styling so this seems like the right path forward.

1. Register nav items on a non-WCA and non-embed page. You can use this PR to test - https://github.com/woocommerce/woocommerce-memberships/pull/103
1. Make sure the nav loads.
1. Make sure no regressions occur on other admin pages.


psealock (approved):
This is testing well and changes look good to me. 🚢 

Just noting that these changes now essentially ensure the Nav and header always come as a pair, which is ideal because they are conceptually dependent on each other.

I think a good follow up will be to make that coupling more clear. When reading the code, the association is only gathered if you follow the logic trail.

octaedro:
Cherry picked into `release/1.7.0`.

PR #5610, opened 2020-11-12
Bump WooCommerce minimum requirements in documentation to 4.5

adrianduffell:
Updates the WooCommerce minimum requirements in the `readme.txt` Getting Started documentation to `4.5`.

I removed the patch version number so it becomes just `4.5`. This is consistent with the other software listed in this section, except PHP which looks like a special case.


samueljseay (approved):
Thanks @adrianduffell ! 🚀

Issue #5375, opened 2020-10-14
Implement fancy placeholders style loaders for Marketing Tab


jconroy:
"Fancy" placeholders (well fancier than what we currently use on the Marketing tab) are used within various areas of wc-admin

By way of example, we mimic this placeholder behavior/styles from wc-admin in AutomateWoo on the new presets list table screen (screenshot below capturing the placeholders - you might need to have good contrast to see it clearly)

<img width="1211" alt="Screen Shot 2020-10-14 at 11 05 21 pm" src="https://user-images.githubusercontent.com/355014/95989712-2a14d100-0e72-11eb-9fb6-be174e6a5337.png">

Investigate using a consistent approach to placeholders for the Marketing tab components while content is being loaded.

Currently, the Recommended Extensions and WooCommerce Knowledge Base cards will show a spinner when content loads which is not consistent with other areas of wc-admin.

<img width="828" alt="Screen Shot 2020-10-14 at 11 14 48 pm" src="https://user-images.githubusercontent.com/355014/95990622-4bc28800-0e73-11eb-9847-406ce6e23563.png">

Some reference links:
https://github.com/woocommerce/automatewoo/pull/747 (private AutoamteWoo exension repo)
https://github.com/woocommerce/woocommerce-admin/blob/7c1299c5e2b088d29d8158a3ad806672ef1d14ef/client/stylesheets/abstracts/_mixins.scss#L23-L35

tomalec:
I started by creating more `*Placeholder` components, see #5611

tomalec:
@jconroy If you mean unifying placeholders only in the Marketing tab, the PR is ready for the review, otherwise I'll work on the remaining spinners tomorrow.

tomalec:
Besides `KnowledgebaseCardPost` and `RecommendedExtensionsItem` shown in the OP and covered in PR #5611, there is a number of places that still use the `Spinner` I was not able to reach all of them using UI, but from what it seems some of them are used in relatively small areas, like breadcrumbs, and select fields. I guess spinner may look better there than a grey box of a placeholder.

The only bigger area I found is the `Chart` component, like to one at `/wordpress/wp-admin/admin.php?page=wc-admin&path=%2Fanalytics%2Fcategories`:
![Spinner in the place of a chart](https://user-images.githubusercontent.com/17435/99291723-89cf1500-2840-11eb-8bd2-cd01cadb26dd.png)
[...]

PR #5611, opened 2020-11-12
Add fancy placeholders for Marketing components

tomalec:
Add placeholders for 

- [x] RecommendedExtensionsItem.
- [x] KnowledgeBase.

Addresses https://github.com/woocommerce/woocommerce-admin/issues/5375.

### Accessibility

### Screenshots
GIF animation of placeholders
https://d.pr/i/uTkAtS

1. Ex: Open page `http://wptest.local/wordpress/wp-admin/admin.php?page=wc-admin&path=%2Fmarketing`
2. Scroll down if needed,
3. If the content blinks too fast you can throttle network speed in dev tools.
4. Observe the "Recommended extensions" and "Knowledge base" sections, before the content is fully loaded.


tomalec:
I'm still not sure about file names here. At first, I tried to adapt to the way the existing `placeholder.js` was created. On the other hand, it resulted in more `placeholder.js` files.

Personally, I don't like having different files with different classes called the same way. It makes file searching harder <kbd>Ctrl</kbd> + <kbd>p</kbd>. In some previous projects, I used to work with conventions to call the file like the class it defines if it defines only one, or at least give it a somewhat-specific module name.

jconroy (commented):
Niiiice @tomalec 👍 

Testing well for me with the exception of the icons for the coupon page recommended extensions - the border is still being shown for the placeholders e.g. https://d.pr/v/Nv0QEA

Thanks for adding/updating tests 🙇

jconroy:
> [I'd appreciate some comment on that.](https://github.com/woocommerce/woocommerce-admin/pull/5611#issuecomment-726428767)

Honestly not sure if wc-admin has a convention in place. As you've found there are already a few other uses of `placeholder.js` so I think for now it'd be safe to just take the same approach.

I definitely see your point though - not sure if @timmyc has any guideance for us.

client/marketing/components/knowledge-base/placeholder.js
```diff
@@ -0,0 +1,24 @@
...
+
+	return (
+		<div
+			className={ 'is-loading ' + classNameBase }
+			key={ index }
+			aria-hidden="true"
+		>
+			<div className={ classNameBase + '-img is-placeholder' }></div>
+			<div className={ classNameBase + '-text' }>
+				<h3 className="is-placeholder"></h3>
```
jconroy, line 17: JS tests are flagging a couple of errors for the headings `17:5  error  Headings must have content and the content must be accessible by a screen reader  jsx-a11y/heading-has-content`

Wonder if we can ignore or use another element like a div here? I think the inbox panel might use a div instead of a heading for this

client/marketing/components/recommended-extensions/placeholder.js
```diff
@@ -0,0 +1,28 @@
...
+			<div className="woocommerce-admin-marketing-product-icon">
+				<span className="is-placeholder" />
+			</div>
+
+			<div className={ `${ classNameBase }__text` }>
+				<h4 className="is-placeholder"></h4>
+				<p>
+					<span className="is-placeholder" key={ 1 } />
+					<span className="is-placeholder" key={ 2 } />
+					<span className="is-placeholder" key={ 3 } />
```
ecgan, line 21: We don't need `key` here.

client/marketing/components/recommended-extensions/style.scss
```diff
@@ -3,36 +3,36 @@
...
-	}
 
 	&__items {
 		display: flex;
 		flex-wrap: wrap;
 
 		@include breakpoint( '>600px' ) {
-			> a {
+			> a,
+ > .is-loading.woocommerce-marketing-recommended-extensions-item {
```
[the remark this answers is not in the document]
ecgan: @tomalec , the [VS Code stylelint extension](https://github.com/stylelint/vscode-stylelint) will show error for you here.

client/marketing/components/knowledge-base/test/index.js
```diff
@@ -49,11 +49,11 @@ describe( 'Posts and not loading', () => {
 		);
 	} );
 
-	it( 'should not display the spinner', () => {
+	it( 'should not display the preloader', () => {
 		const { container } = knowledgeBaseWrapper;
 		expect(
-			container.getElementsByClassName( 'components-spinner' )
-		).toHaveLength( 0 );
+			container.querySelector( '.is-loading, .is-preloader' )
```
ecgan, line 55: @tomalec , I don't see any other files here using `is-preloader` class name, I suppose it comes from other components? (Just want to be sure this is not typo / confusion with `is-placeholder`.

client/marketing/components/knowledge-base/placeholder.js
tomalec: Sorry, I forgot to push recent changes 😶 

I was thinking of switching to `div`, but that would require changing CSS, adding more classes, and making it less semantic. AFAIK, for a11y perspective setting `aria-hidden="true"` on any ancestor should do the job - make it hidden for a11y tools. However, for this check, it seems to be not enough. I've added `aria-hidden="true"` explicitly on the `h*` elements to make it happy.

ecgan:
> Personally, I don't like having different files with different classes called the same way. **It makes file searching harder `Ctrl + p`.** In some previous projects, I used to work with conventions to call the file like the class it defines if it defines only one, or at least give it a somewhat-specific module name.

I share the same opinion with you too. An example: when we do `CTRL + P` and search for `index.js`, a lot of files show up, including those from test folders. Also, it makes reading the editor tab title more difficult and takes up more space. Same thing goes with `style.scss` files.

client/marketing/components/knowledge-base/placeholder.js
jconroy: Nice, TIL

tomalec:
I created a separate issue regarding file name conventions at https://github.com/woocommerce/woocommerce-admin/issues/5637.

So as @jconroy suggested, I'd stick to `placeholder.js` for now.

client/marketing/components/knowledge-base/test/index.js
tomalec: Thanks :) It was a typo.

jconroy:
I think the travis build errors are potentially fixed by https://github.com/woocommerce/woocommerce-admin/pull/5635

tomalec:
After merging letest `main`, Travis is all green.
Now, the PR is waiting for another round of reviews :)

joshuatf (approved):
Very nice addition to the marketing page!  Left a couple quick clean-up comments, but this is testing well and pre-approving.

jconroy:
Thanks for the review @joshuatf

timmyc:
> I definitely see your point though - not sure if @timmyc has any guideance for us.

I don't really have a strong opinion here, and think the p2 that got created around this is a better venue to let folks that do have strong opinions chime in 😆 

I will say that the patterns here in wc-admin largely follow those in Gutenberg, and WordPress.com as well for things like having a duplicated file name.

jconroy:
Thanks for chiming in @timmyc 👍

tomalec:
All comments addressed. I hope I didn't miss anything with translations, as I never used it.
I don't need to update any dictionary files at this point, do I?

jconroy:
> I hope I didn't miss anything with translations, as I never used it. I don't need to update any dictionary files at this point, do I?

@tomalec nope you don't need to do anything with these. We make sure the text is translatable and then translations can be created and shared by others https://translate.wordpress.org/projects/wp-plugins/woocommerce/ there are also a range of plugins available that help with translations.

PR #5614, opened 2020-11-12
Fix WCS slug when building Tracks data.

jeffstieler:
Tracks data was checking FB install instead. Props @dechov - https://github.com/woocommerce/woocommerce-admin/pull/4695#discussion_r521565942.

Fixes https://github.com/woocommerce/woocommerce-admin/pull/4695#discussion_r521565942


louwie17 (approved):
Tested locally and Woo Services was still being installed, nice simple change :)

PR #5616, opened 2020-11-12
Adding body class for navigation to indicate when it is expanded, pre…

### LINKED ISSUE #5580: Navigation: Nav placeholder mobile styles
When refreshing the page, a blank background is made using CSS as a placeholder while the Navigation loads. This reduces the visual jarry-ness once the Navigation is loaded.

When viewing screens at mobile widths, the placeholder is present until the Navigation loads in its collapsed state. Since we're collapsing on load, can the placeholder also reflect mobile styles? In other words, it should be a the small black box on the upper left.

### Deliverables

* When refreshing a screen on mobile widths, the placeholder is the small box on the top left, not the uncollapsed version.

joelclimbsthings:
Fixes #5580 

This is addressing the issue of the navigation placeholder (the container that is visible on page load) appearing as if the navigation is expanded before it's folded by default on mobile. It required adding a body class to indicate when the navigation is expanded (and loaded) in addition to the preexisting class that had been added when it was folded. 

### Screenshots

On mobile, how the navigation placeholder appears before change(s):
image

After changes:
image

- Checkout branch
- Ensure the navigation feature is enabled
- Go to WP-Admin -> WooCommerce to view new navigation
- Reduce width of viewport to less than 960 pixels
- Refresh the page
- While loading, you should observe the smaller placeholder as pictured in the second screenshot.

### DIFF
--- a/client/navigation/components/header/index.js
+++ b/client/navigation/components/header/index.js
@@ -19,17 +19,35 @@ const Header = () => {
 	const siteTitle = getSetting( 'siteTitle', '' );
 	const siteUrl = getSetting( 'siteUrl', '' );
 	const isScrolled = useIsScrolled();
+	const navClasses = {
+		folded: 'is-folded',
+		expanded: 'is-expanded',
+	};
+
+	const foldNav = () => {
+		document.body.classList.add( navClasses.folded );
+		document.body.classList.remove( navClasses.expanded );
+	};
+
+	const expandNav = () => {
+		document.body.classList.remove( navClasses.folded );
+		document.body.classList.add( navClasses.expanded );
+	};
 
 	const toggleFolded = () => {
-		document.body.classList.toggle( 'is-folded' );
+		if ( document.body.classList.contains( navClasses.folded ) ) {
+			expandNav();
+		} else {
+			foldNav();
+		}
 	};
 
 	const foldOnMobile = ( screenWidth = document.body.clientWidth ) => {
-		const isSmallScreen = screenWidth <= 960;
-
-		document.body.classList[ isSmallScreen ? 'add' : 'remove' ](
-			'is-folded'
-		);
+		if ( screenWidth <= 960 ) {
+			foldNav();
+		} else {
+			expandNav();
+		}
 	};
 
 	useEffect( () => {
--- a/client/navigation/style.scss
+++ b/client/navigation/style.scss
@@ -11,6 +11,15 @@
 		display: none !important;
 	}
 
+	#wpcontent,
+	#wpfooter {
+		margin-left: $navigation-width;
+
+		@media ( max-width: 960px ) {
+			margin-left: 0;
+		}
+	}
+
 	#woocommerce-embedded-navigation {
 		position: fixed;
 		top: 0;
@@ -20,6 +29,11 @@
 		box-sizing: border-box;
 		background-color: $gray-900;
 		z-index: 1100; //Must be greater than z-index on .woocommerce-layout__header
+
+		@media ( max-width: 960px ) {
+			width: $header-height;
+			height: $header-height;
+		}
 	}
 
 	.components-navigation {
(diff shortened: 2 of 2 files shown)


client/navigation/components/header/index.js
```diff
@@ -19,17 +19,35 @@ const Header = () => {
 	const siteTitle = getSetting( 'siteTitle', '' );
 	const siteUrl = getSetting( 'siteUrl', '' );
 	const isScrolled = useIsScrolled();
+	const navClasses = {
+		folded: 'is-folded',
+		expanded: 'is-expanded',
+	};
+
+	const foldNav = () => {
```
joshuatf, line 27: Nice, I like these explicit functions; will make debugging easier in the future.

client/navigation/style.scss
```diff
@@ -28,14 +42,10 @@
...
 
-	&:not(.is-folded) {
-		#wpcontent,
-		#wpfooter {
-			margin-left: $navigation-width;
-
-			@media ( max-width: 960px ) {
-				margin-left: 0;
-			}
+	&.is-expanded {
```
joshuatf, line 45: This seems like a bit of a code smell.  Do we need the additional class here?  It seems like a binary state selector should only contain a single selector that is toggled on/off unless we plan on introducing more states for the navigation in the future.

joshuatf (commented):
Testing well; just left a comment about whether or not we want to introduce additional classes here.

client/navigation/style.scss
joelclimbsthings: The issue here is that we didn't have any way to apply styles specific to the expanded state that didn't also apply during the loading state (no body classes applied yet), which is what is required here I believe. 

Since this issue is specifically calling for styles that impact only the loading state, the only way to prevent them from influencing the display _after_ loading, and while the nav is expanded, is to have another class to key off of. 

Does that make sense?

joelclimbsthings: Having said that, I'm actually wondering now if there's another body class that already exists that I could utilize. I'll look into it real quick o.0

joshuatf: There is a `woocommerce-admin-is-loading` class, but that will hide other elements on the page.

>Since this issue is specifically calling for styles that impact only the loading state

Makes sense.  Let me tinker with this for a few minutes, but at a glance it does seem like the way you've handled it here may be the most straightforward solution.

joelclimbsthings: I thought I could potentially use the `no-js` body class to key off of, but it seems there's way too much of a delay between the class being removed and the `<Nav />` component being mounted. 

An alternative solution may be to add a body class via `body_class()` server-side that we then take away when the component is mounted (similar to `no-js` but not removed until we're ready). This would simplify the logic overall, but perhaps be a little less intuitive.

joelclimbsthings: Just saw your earlier comment. It doesn't look like the `woocommerce-admin-is-loading` class is attached in a way that will be helpful. Let me know if you come across anything in your tinkerin'!

client/navigation/components/header/index.js
```diff
@@ -19,17 +19,35 @@ const Header = () => {
 	const siteTitle = getSetting( 'siteTitle', '' );
 	const siteUrl = getSetting( 'siteUrl', '' );
 	const isScrolled = useIsScrolled();
+	const navClasses = {
+		folded: 'is-folded',
```
joshuatf, line 23: Existed prior to this PR, but realizing that these classes are generic and do not describe the `body` but instead describe the nav.  Maybe we should update these to `is-wc-nav-folded` and `is-wc-nav-expanded` to more accurately describe what's happening and avoid conflicts.

client/navigation/style.scss
joshuatf: The `no-js` solution is a nice workaround, but I agree that it's not so intuitive and will complicate things further.  I think the solution you've created here is the best for now. 👍 a

joshuatf (approved):
Thanks for bearing with my poking and prodding!  I think we should rename the state classes in here to avoid conflict, but otherwise pre-approving and LGTM :shipit:

PR #5621, opened 2020-11-13
Add Score Comment to Tracks Event

### LINKED ISSUE #5468: Customer effort score: Show comments section in survey
Part of #5259

Add a section for the user to add a comment to their feedback. The section should only be displayed if the user selects "Very difficult" or "Somewhat difficult". The comment is an *optional* field - add an "Optional" label per this screenshot (decided [here](https://woostartp2.wordpress.com/2020/10/16/customer-effort-score-design-i2/#comment-1070)).

If "Neutral", "Somewhat easy", or "Very easy is selected", the comments field will not show and the effort score will be recorded on its own.

ces-update-1

The [WP core component for the text area](https://wordpress.github.io/gutenberg/?path=/story/components-textareacontrol--default) can be used, with the label bumped up to the `Subtitle small` text component to provide better hierarchy in the modal. The CES prompt question is also displayed as `Subtitle small`.

See [this P2 post](https://woostartp2.wordpress.com/2020/10/16/customer-effort-score-design-i2/).

adrianduffell:
Fixes #5468

### Accessibility

- Ex: Open page `url`
- Click XYZ…

### DIFF
--- a/client/customer-effort-score-tracks/customer-effort-score-tracks.js
+++ b/client/customer-effort-score-tracks/customer-effort-score-tracks.js
@@ -74,10 +74,11 @@ function CustomerEffortScoreTracks( {
 		} );
 	};
 
-	const trackCallback = ( score ) => {
+	const trackCallback = ( score, comments ) => {
 		recordEvent( 'ces_feedback', {
 			action,
 			score,
+			comments,
 			store_age: storeAge,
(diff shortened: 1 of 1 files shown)


adrianduffell:
Closing this in favor of PR #5739.

PR #5623, opened 2020-11-13
Add support for an explicit dismiss button to snackbar, as well as an onDismiss callback

### LINKED ISSUE #5437: Customer effort score: Dismiss functionality in snackbar and don't fade away
Part of #5259

There needs to be a dismiss button added to snackbars.

This needs to be added to the Snackbar component as an optional prop, and used by the CES snackbar prompt.

If the dismiss button is present on a snackbar, it shouldn't fade away, and clicking the body of the snackbar shouldn't dismiss it - the only way to dismiss the snackbar is by clicking the "X"

If the snackbar prompt for a given action is dismissed, the prompt should not be shown again for that action.

See https://woostartp2.wordpress.com/2020/10/16/customer-effort-score-design-i2/ for design discussion.

becdetat:
Fixes #5437

This adds support for an explicit dismiss button to the snackbar, which disables the default behaviour of clicking anywhere on the snackbar to dismiss it. It also adds a callback for when the snackbar is dismissed, which is used by the `CustomerEffortScore` component to stop displaying snackbars for the current action.

### Accessibility

### Screenshots
image

- Delete the `woocommerce_ces_shown_for_actions` option
- Add a product
- The snackbar should appear
- Clicking the snackbar body should not dismiss the snackbar
- Clicking the cross should dismiss the snackbar
- Add another product
- The snackbar should _not_ appear

### Changelog Note:

Dev: Add support for an explicit dismiss button to snackbar, as well as an onDismiss callback


mattsherman (changes requested):
Tested and noticed the following issues:

- Snackbar still auto-dismisses if there is no interaction with it.
- Clicking "Provide feedback" link should dismiss the snackbar before showing the CES modal. The snackbar stays visible while the CES modal is displayed and afterwards.
- The cursor still changes to a pointer hand for the entire snackbar, though clicking anywhere except for the "Provide feedback" link and the "X" does nothing. The cursor should stay as the default unless on something actionable.

becdetat:
Thanks @mattsherman ! I'm pretty sure I've sorted these issues now, if you could give it another look.

client/layout/transient-notices/snackbar/index.js
```diff
@@ -40,19 +40,42 @@ function Snackbar(
 		actions = [],
 		onRemove = noop,
 		icon = null,
+		explicitDismiss = false,
+		// onDismiss is a callback executed when the snackbar is dismissed.
+		// It is distinct from onRemove, which _looks_ like a callback but is
+		// actually the function to call to remove the snackbar from the UI.
+		onDismiss = null,
```
mattsherman, line 47: This is confusing... how about we rename `onRemove` to something like `removeSnackbar` or `removeFunction`?

client/layout/transient-notices/snackbar/index.js
```diff
@@ -92,6 +115,9 @@ function Snackbar(
 							isTertiary
 							onClick={ ( event ) => {
 								event.stopPropagation();
+								if ( explicitDismiss ) {
+									onRemove();
+								}
 								if ( onClick ) {
 									onClick( event );
 								}
```
mattsherman, line 123: It would be nice to move this onClick handler to a non-anonymous function.

client/layout/transient-notices/style.scss
```diff
@@ -34,4 +38,9 @@
 		top: 24px;
 		left: 26px;
 	}
+
+	.components-snackbar__explicit-dismiss {
```
mattsherman, line 42: It's confusing having `.components-snackbar-explicit-dismiss` and `components-snackbar__explicit-dismiss`. How about naming this one something like `components-snackbar__dismiss-button`?

mattsherman (changes requested):
Thanks for the changes @becdetat! Things now work as expected.

There is a z-index issue now with the modal, probably because the snackbar is dismissed prior to showing the modal (as it should be).

A quick fix for this in this PR, to allow for things to continue to work with the modal placeholder, would be:

```
.customer-effort-score_modal {
    z-index: 9999;
}
```

When we merge in the actual modal (/cc @adrianduffell), we'll have to make sure that the z-index of it is being calculated properly than being hard-coded.

I think some of the naming of things could be changed to be a little less confusing. Some of these fall outside of the immediate scope of this issue, so I'd be okay with those being part of a follow-up PR if need be.

client/layout/transient-notices/snackbar/index.js
becdetat: I agree this is confusing, I wanted to rename `onRemove` but I'm concerned about backwards compatibility - this is to be merged upstream to Gutenberg

client/layout/transient-notices/snackbar/index.js
becdetat: Done in 52182f66333c9cebaaef1e1f9e1aced9b601ade2

client/layout/transient-notices/style.scss
becdetat: Nice, done in 09de3fd8e9c807290f5305780f71624dca6b99e2

becdetat:
@mattsherman 
> There is a z-index issue now with the modal

Cool, I've made the z index 9999. I couldn't see an issue with the z index in my testing, hopefully it fixes it for you.

client/layout/transient-notices/snackbar/index.js
mattsherman: Ah, right. That makes sense.

client/layout/transient-notices/snackbar/index.js
```diff
@@ -73,11 +108,11 @@ function Snackbar(
...
 			className={ classes }
-			onClick={ onRemove }
+			onClick={ ! explicitDismiss ? dismissMe : noop }
 			tabIndex="0"
-			role="button"
-			onKeyPress={ onRemove }
-			aria-label={ __( 'Dismiss this notice' ) }
+			role={ ! explicitDismiss ? 'button' : '' }
+			onKeyPress={ ! explicitDismiss ? dismissMe : noop }
+			aria-label={ ! explicitDismiss ? __( 'Dismiss this notice' ) : '' }
```
joshuatf, line 115: Missing text domain here.

packages/customer-effort-score/src/style.scss
```diff
@@ -3,6 +3,6 @@
 	top: 300px;
 	left: 300px;
 	background: #faa;
-	z-index: 999;
+	z-index: 9999;
```
joshuatf, line 6: Is this relative to something?  Maybe we could add a comment describing what it needs to be above/below.

joshuatf (commented):
Very cool addition to the notice modal!  Left a couple comments/questions in the code.

packages/customer-effort-score/src/style.scss
mattsherman: We should take care of this in a more robust fashion in #5515, so this is just a temporary fix.

mattsherman (approved):
Tested and works as expected. Once the localization issue from @joshuatf is addressed I think this is good to go.

Thanks for the updates!

packages/customer-effort-score/src/style.scss
becdetat: This is just a temporary modal style, it will be removed soon in #5515

client/layout/transient-notices/snackbar/index.js
becdetat: Interesting, good catch. This is forked from the component in Gutenberg and has been [pushed upstream](https://github.com/WordPress/gutenberg/pull/26952) and the fork will be [removed](https://woostartp2.wordpress.com/2020/11/10/forking-the-snackbar-component-from-gutenberg/) once the changes become available to us via Gutenberg. So, I don't _think_ there is much point in changing this - I imagine that this is correctly translated in Gutenberg.

PR #5624, opened 2020-11-13
Update panel icon order - moving help tab far right

### LINKED ISSUE #5589: Help panel icon switched with Display icon
**Describe the bug**
I was just looking at [this PR](https://github.com/woocommerce/woocommerce-admin/pull/5538) (which is awesome btw) but noticed a small thing. We should display the Help icon on the far right of the app bar per the design. Reason being that the Display option only shows up on the Home screen and this avoids unnecessary movement of app bar icons as users navigate across different screens.

**Screenshot of the design**
Screen Shot 2020-11-10 at 12 14 00 PM

louwie17:
Fixes #5589 

Allowed the Tabs component to render custom components if passed in through the tab object, this way we can handle
the order more easily without making explicit exceptions (like we did for Display Options tab).
The display options component is now passed in as a tab object in the Activity Panel component.

Extra thoughts:
- Could add support for the onTabClick in the Display options component, so other tabs are closed if `Display` is active.
- A tab config could contain an (optional) order integer field, but that didn't seem necessary atm.

### Accessibility

- [✔︎] I've tested using only a keyboard (no mouse)

### Screenshots

Screen Shot 2020-11-13 at 11 25 16 AM

- Have a new WooCommerce store setup
- Go to **WooCommerce > Home**
- Help tab should now be far right

### DIFF
--- a/client/header/activity-panel/index.js
+++ b/client/header/activity-panel/index.js
@@ -212,15 +212,19 @@ export class ActivityPanel extends Component {
 			  }
 			: null;
 
-		const displayOption = showDisplayOptions
+		const displayOptions = showDisplayOptions
 			? {
 					component: DisplayOptions,
 			  }
 			: null;
 
-		return [ inbox, ...stockAndReviews, setup, displayOption, help ].filter(
-			Boolean
-		);
+		return [
+			inbox,
+			...stockAndReviews,
+			setup,
+			displayOptions,
+			help,
+		].filter( Boolean );
 	}
 
 	getPanelContent( tab ) {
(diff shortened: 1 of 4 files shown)


client/header/activity-panel/index.js
```diff
@@ -208,7 +212,15 @@ export class ActivityPanel extends Component {
 			  }
 			: null;
 
-		return [ inbox, ...stockAndReviews, setup, help ].filter( Boolean );
+		const displayOption = showDisplayOptions
```
mattsherman, line 215: Should be named `displayOptions`

mattsherman (approved):
Just one minor request for a const naming change. Once you make that, I'm cool with you merging without another round of review.

Overall, code looks good, and testing went as expected according to test instructions. 

Nice work!

client/header/activity-panel/index.js
louwie17: Updated :) thanks for noticing. Updating status to ready for merge (merging when CI works again)

PR #5628, opened 2020-11-13
Position transient notices based on nav width

### LINKED ISSUE #5622: Navigation: Snackbar misaligned
The new Navigation has a width different than the previous wp-admin sidebar.

`client/layout/transient-notices/style.scss` the left declaration is set to 176px. This should reflect the new width if Nav is enabled.

Screen Shot 2020-11-13 at 4 36 01 PM

### Steps to reproduce

1. Do an action to create a notice, or add one programatically:
* Add a `notices` prop `notices={ [ { id: 1232, content: 'hello, Im a notice' } ] }` to 
https://github.com/woocommerce/woocommerce-admin/blob/a1237ed49636c751fedac1d4574ea49b2d6019ab/client/layout/index.js#L141
* Comment out the `TransientNotice` `withSelect`.
https://github.com/woocommerce/woocommerce-admin/blob/a1237ed49636c751fedac1d4574ea49b2d6019ab/client/layout/transient-notices/index.js#L47-L54
* View any page with the Nav.

### Deliverables

* Maintain current spacing when nav is turned off.
* Apply the same spacing when the nav is turned on.

joshuatf:
Fixes #5622 

Moves the transient notices when using the new wc nav width.

### Screenshots
Screen Shot 2020-11-13 at 1 55 25 PM

1. Add a fake notice to test.  This can be done in `transient-notices/index.js` on line 53 with this line `return { notices: notices.concat( [ { id: 1232, content: 'hello, Im a notice' } ] ) };`
1. Check that without the wc nav, the notice is correctly positioned.
1. Check that with the nav enabled, the notice is correctly positioned and not overlapping the nav.

### DIFF
--- a/client/navigation/stylesheets/_variables.scss
+++ b/client/navigation/stylesheets/_variables.scss
@@ -3,6 +3,7 @@ $navigation-width: 240px;
 $navigation-x-padding: 30px;
 
 // WordPress defaults.
+$admin-menu-width: 160px;
 $admin-bar-height: 32px;
 $admin-bar-height-mobile: 46px;
 
--- a/client/layout/transient-notices/style.scss
+++ b/client/layout/transient-notices/style.scss
@@ -1,7 +1,9 @@
+@import '../../navigation/stylesheets/variables';
+
 .woocommerce-transient-notices {
 	position: fixed;
 	bottom: $gap-small;
-	left: 176px;
+	left: $admin-menu-width + $gap;
 	z-index: 99999;
 
 	@media ( max-width: 960px ) {
@@ -19,6 +21,12 @@
 			margin-left: auto;
 			margin-right: auto;
 		}
+	}
+}
+
+.has-woocommerce-navigation {
+	.woocommerce-transient-notices {
+		left: $navigation-width + $gap;
 	}
 }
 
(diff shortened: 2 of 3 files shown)


client/navigation/stylesheets/_variables.scss
```diff
@@ -3,6 +3,7 @@ $navigation-width: 240px;
 $navigation-x-padding: 30px;
 
 // WordPress defaults.
+$admin-menu-width: 160px;
```
joelclimbsthings, line 6: Top notch variable :+1:

joelclimbsthings:
Changes look good @joshuatf , although in testing I did notice one thing.

Using the old navigation, the notice always stays fixed in the lower left corner, even when collapsed (in mobile). On the new nav the notice can end up in the middle of the page with the navigation folded, especially with a narrow viewport (such as mobile): 

image

joshuatf:
Great catch, @joelclimbsthings.  Moved it left when folded and also left on narrow viewports (even when it's open) to prevent the notice from flowing off-screen.

joelclimbsthings:
Great, LGTM. I did notice this state pictured below that seemed a little weird to me on mobile, but it looks like it's identical to the behavior sans new navigation, so I'm thinking it's expected!

:shipit: 

image

joelclimbsthings (approved):

PR #5633, opened 2020-11-15
Fix the travis builds by using composer 2.0.6 as a temporary measure

samueljseay:
Currently our travis builds are broken due to an incompatibility between composer 2.0.7 and jetpack autoloader. This is being worked on, but for now this temporary downgrade of composer will allow our builds to keep passing.

I've also had to update the package lock again here because it was out of date (there is some issue I keep seeing with woocommerce/e2e breaking with the existing lock. I am not sure why this bug keeps finding its way back in.

Note: this does solve the problem, but I'm experiencing an unrelated new issue from Travis as it can't pull the php:7.3 docker image.


samueljseay:
Closing this in favour of #5635

PR #5635, opened 2020-11-16
Fix the travis build and move the e2e tests to GH actions

samueljseay:
Recently 2 issues arose with builds:

1. There is a change in an interface in Composer 2.0.7 that breaks Jetpack autoloader. This causes all builds to fail. To fix that temporarily we are downgrading composer to 2.0.6 via `self-update`. We can remove that once the fix is made in jetpack autoloader

2. Travis started randomly failing downloading Docker images required to run tests. Since we were already looking to move to GH actions for PR checks, this was an ideal time to migrate the e2e test job to Github actions. Initial results are promising, the Travis version took ~ 9 minutes and the GH action equivalent takes 5.5 mins (and there are things we can do to speed this up even more).


adrianduffell (approved):
This is great, thanks for tackling this. The Composer downgrade in Travis and the new GH action look ok to ship 🚀   

Noting the `package.lock lockfileVersion` downgrades from `2` to `1`, and a lot of lines are removed in the diff. Since all tests pass this seems safe to me.

Issue #5458, opened 2020-10-22
Navigation: Explore removing need for id in client side registration of wc-admin


psealock:
https://github.com/woocommerce/woocommerce-admin/pull/5445 introduced a need for plugins extending wc-admin to supply an `id` for filters `woocommerce_admin_reports_list` and `woocommerce_admin_pages_list` in order to match client side ids for purposes of slotFilling custom links.

>I needed to add id on the client side to results of getReports and getPages to match the id's registered in PHP so that Slot and Fill names can be correctly identified. This means that plugins currently using hooks like woocommerce_admin_pages_list to add pages will need to update. Since we have that information already from PHP, it feels like there is a better solution to investigate so that we can keep the filter API the same. Here is an example from WC Payments that would need updating.

https://github.com/Automattic/woocommerce-payments/blob/c7f9d58e2d20fc5fb10d3fa5d205f698bcf3b8c6/client/index.js#L24

This issue is to track exploring possibilities of refactoring the way pages and reports are calculated client side to automatically grab ids supplied by PHP. This will be useful to avoid changing the API requirements of those filters.

joshuatf:
Another possible option here is to use the `path` property instead of adding an ID.  Since this is unique to pages added in WCA anyway, we could avoid adding the `id` property.

psealock:
>Another possible option here is to use the path property instead of adding an ID

True, I'll need to access to the path property in PHP in that case. Its just a matter of syncing a slug or some other identifier between JS and PHP.

PR #5642, opened 2020-11-16
Navigation: Some navigation menu items left out due to duplicate IDs

### LINKED ISSUE #5626: Some navigation items missing due to duplicate IDs
**Describe the bug**

@elizaan36 pointed out that the "Products" item is missing from the "Settings" category in [another PR](https://github.com/woocommerce/woocommerce-admin/issues/5572#issuecomment-726889677). It looks like this is actually due to conflicting IDs, an issue that could also easily cause issues elsewhere. 

We need to rethink a bit how these IDs are generated to prevent these kinds of conflicts. 

**To Reproduce**

1. Ensure the new navigation is enabled
2. Navigate to WooCommerce -> Settings
3. Observe that the "Products" item is no longer present

**Expected behavior**
All expected items should be present.

joelclimbsthings:
Fixes #5626

This issue was noticed due to the "Products" menu item disappearing from the "Settings" category. On closer inspection, it looks like this stems from that use of fairly generic IDs (ie "products") resulting in conflicts that resulted in some items not being added as they should. For instance, in this case, a menu item had already been registered with the ID of "products," which resulted in the Settings page with the same ID being dismissed. 

To remedy this I've modified the code to identify each item in the context of their `parent` as well, so the resulting ID now follows this format: `parent-name/item-id`. 

This should allow similar items to be added without issue as long as they exist in a different level/category of the navigation hierarchy. 

This PR also includes updating how current slot-fill items are registered to support the new ID structure, as well as updating the documentation to reflect this change. 

Last, we're now printing a warning to the `error_log()` when a a duplicate item attempts registration. I'm not entirely confident of the suggested format for this since I could only find what example of it's use in the project.

### Screenshots

Settings -> Products item now visible:
image

- Checkout branch
- Navigation to WooCommerce -> Settings
- Ensure that the "Products" menu item is now visible in this category, as in the screenshot
- Click back and forth from a few react-based pages (Customer/Home/Analytics reports) and ensure there is still no page refresh.

### DIFF
--- a/client/layout/controller.js
+++ b/client/layout/controller.js
@@ -59,8 +59,7 @@ export const getPages = () => {
 			__( 'Home', 'woocommerce-admin' ),
 		],
 		wpOpenMenu: 'toplevel_page_woocommerce',
-		id: 'home',
-		navParent: 'woocommerce',
+		id: 'woocommerce/home',
 	} );
 
 	if ( window.wcAdminFeatures.analytics ) {
@@ -76,8 +75,7 @@ export const getPages = () => {
 				__( 'Overview', 'woocommerce-admin' ),
 			],
 			wpOpenMenu: 'toplevel_page_wc-admin-path--analytics-overview',
-			id: 'woocommerce-analytics-overview',
-			navParent: 'analytics',
+			id: 'analytics/woocommerce-analytics-overview',
 		} );
 		pages.push( {
 			container: AnalyticsSettings,
@@ -91,8 +89,7 @@ export const getPages = () => {
 				__( 'Settings', 'woocommerce-admin' ),
 			],
 			wpOpenMenu: 'toplevel_page_wc-admin-path--analytics-overview',
-			id: 'woocommerce-analytics-settings',
-			navParent: 'settings',
+			id: 'settings/woocommerce-analytics-settings',
 		} );
 		pages.push( {
 			container: AnalyticsReport,
@@ -102,8 +99,7 @@ export const getPages = () => {
 				__( 'Customers', 'woocommerce-admin' ),
 			],
 			wpOpenMenu: 'toplevel_page_woocommerce',
-			id: 'woocommerce-analytics-customers',
-			navParent: 'woocommerce',
+			id: 'woocommerce/woocommerce-analytics-customers',
 		} );
 		pages.push( {
 			container: AnalyticsReport,
@@ -138,8 +134,7 @@ export const getPages = () => {
 				__( 'Overview', 'woocommerce-admin' ),
 			],
 			wpOpenMenu: 'toplevel_page_woocommerce-marketing',
-			id: 'woocommerce-marketing',
-			navParent: 'marketing',
+			id: 'marketing/woocommerce-marketing',
 		} );
 	}
 
--- a/client/layout/controller.js
+++ b/client/layout/controller.js
@@ -60,6 +60,7 @@ export const getPages = () => {
 		],
 		wpOpenMenu: 'toplevel_page_woocommerce',
 		id: 'home',
+		navParent: 'woocommerce',
 	} );
 
 	if ( window.wcAdminFeatures.analytics ) {
@@ -76,6 +77,7 @@ export const getPages = () => {
 			],
 			wpOpenMenu: 'toplevel_page_wc-admin-path--analytics-overview',
 			id: 'woocommerce-analytics-overview',
+			navParent: 'analytics',
 		} );
 		pages.push( {
 			container: AnalyticsSettings,
@@ -90,6 +92,7 @@ export const getPages = () => {
 			],
 			wpOpenMenu: 'toplevel_page_wc-admin-path--analytics-overview',
 			id: 'woocommerce-analytics-settings',
+			navParent: 'settings',
 		} );
 		pages.push( {
 			container: AnalyticsReport,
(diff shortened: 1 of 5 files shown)


src/Features/Navigation/Menu.php
```diff
@@ -231,6 +233,19 @@ private static function add_item( $args ) {
...
+		if ( isset( self::$menu_items[ $menu_item['id'] ] ) ) {
+			error_log(  // phpcs:ignore
+				sprintf(
+					/* translators: 1: Duplicate menu item path. */
+					esc_html__( 'You have attempted to register a duplicate item with WooCommerce Navigation: %1$s', 'woocommerce-admin' ),
+					'`' . $menu_item['id'] . '`'
+				)
+			);
+			return;
+		}
```
psealock, line 247: Nice addition here Joel

client/layout/controller.js
```diff
@@ -60,6 +60,7 @@ export const getPages = () => {
 		],
 		wpOpenMenu: 'toplevel_page_woocommerce',
 		id: 'home',
+		navParent: 'woocommerce',
```
psealock, line 63: I introduced the `id` parameter just for the purposes of the Nav so I'm confident its not used anywhere else. Given that, do you think it may be worthwhile to just make the id's themselves more unique, ie `id: 'woocommerce/home'`?

psealock (commented):
Changes look great @joelclimbsthings, I'm just wondering if we can simplify the `id`.

psealock:
Speaking of id's, I just realized one thing. I introduced the id for use of identification of screens but didn't really like that now any plugin extending wc-admin now had to also add the same id on the client side which would change the API. 

As an example, [see WC Pay adding a Deposits screen in PHP](https://github.com/Automattic/woocommerce-payments/blob/11c4f56a605fc947ae5dd3b726a7ca9a1965daec/includes/admin/class-wc-payments-admin.php#L73-L80). An `id` is present on the PHP side, but in order to match id's to get SlotFill working I addded the `id` parameter on the client as well. In this PR you're manipulating wc-admin's implementation of this id. On the client side, [here is WC Pay registering the same Deposits page](https://github.com/Automattic/woocommerce-payments/blob/c7f9d58e2d20fc5fb10d3fa5d205f698bcf3b8c6/client/index.js#L35-L40). In order for slotFill to work they will need to supply the same id on the client as they did the PHP side. I didn't like that requirement because I wanted it to work "out of the box", so I created https://github.com/woocommerce/woocommerce-admin/issues/5458 to track a solution.

Sooooo, the short of the long story is: Can we use `path` parameter instead as a unique identifier? That parameter is already available on PHP and JS so no update would be required on wc-admin extensions? If that isn't unique enough, can we use something else? Maybe serialize one of the other parameters?

If so, this PR would solve https://github.com/woocommerce/woocommerce-admin/issues/5458 as well. I hope this makes sense and my rambling isn't too hard to follow.

client/layout/controller.js
joshuatf: +1

src/Features/Navigation/Menu.php
```diff
@@ -231,6 +233,19 @@ private static function add_item( $args ) {
...
 		}
 
+		$menu_item['id'] = $menu_item['parent'] . '/' . $menu_item['id'];
+
+		if ( isset( self::$menu_items[ $menu_item['id'] ] ) ) {
+			error_log(  // phpcs:ignore
+				sprintf(
+					/* translators: 1: Duplicate menu item path. */
+					esc_html__( 'You have attempted to register a duplicate item with WooCommerce Navigation: %1$s', 'woocommerce-admin' ),
+					'`' . $menu_item['id'] . '`'
```
joshuatf, line 243: ❤️

client/layout/controller.js
joelclimbsthings: Definitely :+1:

joelclimbsthings:
Thanks for the thoughts @psealock , I really like the idea of automating the client-side registration. It sounds like the intention would only be to do so for pages already registering through `wc-admin`, right? It would be easy enough to use the `path` as the unique ID, with the only trick being how we expose this to the client-side. 

We _do_ already have `getPages()` in `controller.js`, but these paths are currently set manually for the most part. I know we have tools such as `wp_enqueue_script()` to surface data from the server-side in the client side, but I haven't used them as of yet. Do you have any thoughts around how we tie these together, and if we want to do so in the context of this PR?

joshuatf:
>It sounds like the intention would only be to do so for pages already registering through wc-admin, right?

I'm not sure if @psealock's intention was to replace `id` with `path` just for WCA items or to make this the new ID for all nav items.  The latter is an interesting idea, but we'd need to determine if we're okay with not allowing duplicate paths in the same category.  Some plugins do take advantage of this cross-category, but I think it may be an edge case where a plugin would want to register the exact same path twice in a single category.

>I know we have tools such as wp_enqueue_script() to surface data from the server-side in the client side, but I haven't used them as of yet.

`wp_localize_script` is the WordPress way to pass server vars to client scripts.

>Do you have any thoughts around how we tie these together, and if we want to do so in the context of this PR?

I think @psealock's method avoids passing server to client info and relies on the path being the identifier.

joelclimbsthings:
Closing in favor of https://github.com/woocommerce/woocommerce-admin/pull/5665#issuecomment-732470201

PR #5646, opened 2020-11-17
add dewps request table

psealock:
The WP [Dependency Extraction Webpack Plugin](https://github.com/WordPress/gutenberg/tree/master/packages/dependency-extraction-webpack-plugin) has a table that explains what modules are mapped. I thought it would be a good idea to add what extra modules are mapped with this extended version of the plugin.

1. Read changes to make sure they make sense


samueljseay (approved):
Looks good thanks for adding this extra info. 🚀

PR #5649, opened 2020-11-17
Add/5349 trigger ces on filtering analytics

### LINKED ISSUE #5349: Customer effort score: Trigger when filtering in Analytics

moon0326:
Fixes #5349 

Continuing from PR https://github.com/woocommerce/woocommerce-admin/pull/5594 -- This PR uses a store to trigger CES surveys.

This PR triggers Customer Effort Score survey when a user performs an action on one of the pages in Analytics.

### Screenshots
Screen Shot 2020-11-10 at 6 08 12 PM

Make sure to delete both `woocommerce_ces_shown_for_actions` and `woocommerce_ces_tracks_queue` from the `wp_options`. Ensure that you have tracking enabled on WooCommerce > Settings > WooCommerce.com.

1. Navigate to one of the pages in Analytics.
2. Change `Date Range:` and click [ Update ] button. You can also use the advanced, single product, and comparison filters.
3. You should see the survey at the bottom left corner. Click the "Give feedback" text and confirm that the modal pops up.
4. After the survey snackbar disappears, keep changing the filters and make sure that the survey does not popup again.
5. Refresh the page and repeat the steps.
6. Make sure the survey does not pop up again after the page refresh.


client/analytics/components/report-filters/index.js
```diff
@@ -14,41 +15,55 @@ import {
...
-			this
+		this.onDateSelect = this.onDateSelect.bind( this );
+		this.onFilterSelect = this.onFilterSelect.bind( this );
+		this.onAdvancedFilterAction = this.onAdvancedFilterAction.bind( this );
+	}
+
+	showCesSurveyTrack() {
+		const { addCesSurveyTrack } = this.props;
+		addCesSurveyTrack(
+			'woocommerce_admin_analytics_filtered',
```
becdetat, line 37: The action name doesn't need the `woocommerce_admin` prefix, `analytics_filtered` should be fine

client/analytics/report/categories/index.js
```diff
@@ -55,6 +57,16 @@ export default class CategoriesReport extends Component {
 				: 'category';
 		}
 
+		filters[ 0 ].filters[ 2 ].settings.onClick = () => {
```
becdetat, line 60: This will be a bit brittle, can you use `.filter` to find the right filter instead?

client/analytics/report/products/index.js
```diff
@@ -75,6 +77,16 @@ class ProductsReport extends Component {
 				compareObject === 'products' ? 'product' : 'variation';
 		}
 
+		filters[ 0 ].filters[ 2 ].settings.onClick = () => {
+			addCesSurveyTrack(
```
becdetat, line 81: Can you extract this `onClick` handler into a new file so it's de-duplicated?

client/customer-effort-score-tracks/customer-effort-score-tracks-container.js
```diff
@@ -48,7 +50,7 @@ function CustomerEffortScoreTracksContainer( {
 					key={ index }
 					action={ item.action }
 					label={ item.label }
-					onSubmitLabel={ item.onsubmit_label }
+					onSubmitLabel={ item.onSubmitLabel }
```
becdetat, line 53: This is actually meant to be `onsubmit_label` as it is passed in from the PHP side of things (https://github.com/woocommerce/woocommerce-admin/blob/main/src/Features/CustomerEffortScoreTracks.php#L200)

moon0326: @becdetat I was thinking about moving that logic from PHP to JS since JS also needs to pass the value. It's duplicated. What do you think about moving it to JS so that we don't have to set the value in PHP side?

packages/components/src/compare-filter/index.js
```diff
@@ -27,10 +27,15 @@ export class CompareFilter extends Component {
 		this.state = {
 			selected: [],
 		};
-
 		this.clearQuery = this.clearQuery.bind( this );
 		this.updateQuery = this.updateQuery.bind( this );
 		this.updateLabels = this.updateLabels.bind( this );
+		this.onClickHandlers = [ this.updateQuery ];
```
becdetat, line 33: Rather than using an array here it would be simpler and more understandable to just call `this.updateQuery` and `this.props.onClick` directly in `onButtonClicked()`

client/analytics/report/categories/index.js
moon0326: Thank you for the tip :) I also thought it is a bit brittle. Improved with commit https://github.com/woocommerce/woocommerce-admin/pull/5649/commits/8204aa31cf41f819b56943977a03b9519a345d29

client/analytics/report/products/index.js
moon0326: Extracted to action in https://github.com/woocommerce/woocommerce-admin/pull/5649/commits/8204aa31cf41f819b56943977a03b9519a345d29

client/customer-effort-score-tracks/customer-effort-score-tracks-container.js
becdetat: If you can move it then that would be good :-) . We still need to have a channel for generating the surveys on the server side though.

packages/components/src/compare-filter/index.js
moon0326: Thank you for the suggestion. Updated with commit https://github.com/woocommerce/woocommerce-admin/pull/5649/commits/987f700984ad1bb3406255b395558057b9319605

packages/components/src/compare-filter/index.js
```diff
@@ -79,6 +78,13 @@ export class CompareFilter extends Component {
 		updateQueryString( { [ param ]: idList.join( ',' ) }, path, query );
 	}
 
+	onButtonClicked( e ) {
+		this.updateQuery( e );
```
becdetat, line 82: Nice!

client/customer-effort-score-tracks/customer-effort-score-tracks-container.js
moon0326: Yup, I was thinking about removing [this line](https://github.com/woocommerce/woocommerce-admin/blob/add/5349-trigger-ces-on-filtering-analytics/src/Features/CustomerEffortScoreTracks.php#L200) only and set the default value in JS in case of undefined.

client/analytics/components/report-filters/index.js
```diff
@@ -31,30 +30,19 @@ class ReportFilters extends Component {
...
-				'woocommerce-admin'
-			)
-		);
-	}
-
 	onDateSelect( data ) {
-		const { report } = this.props;
-		this.showCesSurveyTrack();
+		const { report, addCesSurveyTrackForAnalytics } = this.props;
+		addCesSurveyTrackForAnalytics();
```
becdetat, line 35: I think you're going off track a bit here, the data store is meant to be reusable across many different consumers, not just for analytics. Can you make it generic?

client/analytics/components/report-filters/index.js
moon0326: Thank you :) I thought we always need to prefix it for some reason. Updated it with commit [8204aa3](https://github.com/woocommerce/woocommerce-admin/commit/8204aa31cf41f819b56943977a03b9519a345d29#diff-094f070da895dc2eeb07300abbba96a33d495e023ae085eff766a72754ceb281R53Rc)

client/analytics/components/report-filters/index.js
moon0326: @becdetat That's true. For the consumers, there is a generic action [addCesSurveyTrack](https://github.com/woocommerce/woocommerce-admin/blob/8204aa31cf41f819b56943977a03b9519a345d29/client/customer-effort-score-tracks/data/actions.js#L30).

client/customer-effort-score-tracks/customer-effort-score-tracks-container.js
becdetat: It still needs to be settable on the server side - other consumers (both in a8c and potentially 3PDs) will be using both the PHP and the JS (data store) approach to creating CES surveys.

becdetat:
Overall it's looking very good, I think we just need to keep things generic enough that this can be used externally.

client/customer-effort-score-tracks/customer-effort-score-tracks-container.js
moon0326: @becdetat That makes sense. I will update it to be compatible with PHP.

moon0326: Changed it back to underscore naming convention for the key in https://github.com/woocommerce/woocommerce-admin/pull/5649/commits/a2171375260fa1a0c5a6bd4f9d2fdc06fb984505

client/analytics/components/report-filters/index.js
becdetat: Ok cool

client/customer-effort-score-tracks/data/actions.js
```diff
@@ -0,0 +1,63 @@
...
+		pageNow,
+		adminPage,
+		onsubmit_label,
+	};
+}
+
+/**
+ * Add a new CES survey track for the pages in Analytics menu
+ */
+export function addCesSurveyTrackForAnalytics() {
```
becdetat, line 52: Can you change this to call `addCesSurveyTrack`? I think that would make it clearer that `addCesSurveyTrack` is the one that should be used by default.

client/analytics/report/categories/index.js
```diff
@@ -55,6 +62,13 @@ export default class CategoriesReport extends Component {
 				: 'category';
 		}
 
+		const compareProducts = filters[ 0 ].filters.filter(
+			( item ) => item.value === 'compare-categories'
+		);
+		if ( compareProducts.length ) {
```
becdetat, line 68: This is much better. The trouble with looking at the length and conditionally adding the `onClick` handler is that it's potentially going to silently do nothing if the filters change shape - it's gone from too brittle to too flexible. I think if you just take out the `if` statement and assume that `compareProducts[0]` isn't going to fail, this would be robust but fail in a way that makes sense and is easier to track down immediately.

moon0326: That's a really good idea 👍 changed it in https://github.com/woocommerce/woocommerce-admin/pull/5649/commits/a1a9552ced71db7a1ad5708892dc30e8bd789653

client/customer-effort-score-tracks/data/actions.js
moon0326: Good catch! 👍 changed it in https://github.com/woocommerce/woocommerce-admin/pull/5649/commits/a1a9552ced71db7a1ad5708892dc30e8bd789653

client/analytics/report/categories/index.js
```diff
@@ -55,6 +62,10 @@ export default class CategoriesReport extends Component {
 				: 'category';
 		}
 
+		filters[ 0 ].filters.filter(
+			( item ) => item.value === 'compare-categories'
+		)[ 0 ].settings.onClick = addCesSurveyTrackForAnalytics;
```
mattsherman, line 67: Filters have filters? Mildly confusing (and pre-existing). 😐 Anyway...

I think using `find()` instead of `filter()` would be a bit clearer, as you are always using the first item found by the `filter()`.

client/analytics/report/coupons/index.js
```diff
@@ -33,9 +35,18 @@ export default class CouponsReport extends Component {
...
+			isRequesting,
+			query,
+			path,
+			addCesSurveyTrackForAnalytics,
+		} = this.props;
 		const { mode, itemsLabel } = this.getChartMeta();
 
+		filters[ 0 ].filters.filter(
+			( item ) => item.value === 'compare-coupons'
+		)[ 0 ].settings.onClick = addCesSurveyTrackForAnalytics;
```
mattsherman, line 48: Same comment as above.

client/analytics/report/products/index.js
```diff
@@ -75,6 +77,10 @@ class ProductsReport extends Component {
 				compareObject === 'products' ? 'product' : 'variation';
 		}
 
+		filters[ 0 ].filters.filter(
+			( item ) => item.value === 'compare-products'
+		)[ 0 ].settings.onClick = addCesSurveyTrackForAnalytics;
```
mattsherman, line 82: Same comment as above.

client/analytics/report/taxes/index.js
```diff
@@ -29,9 +31,18 @@ export default class TaxesReport extends Component {
...
+			isRequesting,
+			query,
+			path,
+			addCesSurveyTrackForAnalytics,
+		} = this.props;
 		const { mode, itemsLabel } = this.getChartMeta();
 
+		filters[ 0 ].filters.filter(
+			( item ) => item.value === 'compare-taxes'
+		)[ 0 ].settings.onClick = addCesSurveyTrackForAnalytics;
```
mattsherman, line 44: See above. :-)

client/analytics/report/variations/index.js
```diff
@@ -45,6 +53,10 @@ const VariationsReport = ( props ) => {
 		chartQuery.segmentby = 'variation';
 	}
 
+	filters[ 0 ].filters.filter(
+		( item ) => item.value === 'compare-variations'
+	)[ 0 ].settings.onClick = addCesSurveyTrackForAnalytics;
```
mattsherman, line 58: See above.

client/customer-effort-score-tracks/data/actions.js
```diff
@@ -0,0 +1,62 @@
...
+ * @param {string} pageNow value of window.pagenow
+ * @param {string} adminPage value of window.adminpage
+ * @param {string} onsubmit_label label for the snackback onsubmit
+ */
+export function addCesSurveyTrack(
+	action,
+	label,
+	pageNow = window.pagenow,
+	adminPage = window.adminpage,
+	onsubmit_label = undefined
```
mattsherman, line 37: Why explicitly set `onsubmit_label` to `undefined`?

mattsherman (commented):
I did not do a thorough review... I just glanced through the code and left a few minor comments!

client/analytics/report/categories/index.js
moon0326: @mattsherman Thank you 👍 I agree that `find()` is better for this case. Updated in https://github.com/woocommerce/woocommerce-admin/pull/5649/commits/291516cf5f50e30defd6fd6e20c2b40fd00d1203

client/customer-effort-score-tracks/data/actions.js
moon0326: I think it's better for readability since it is explicit and the previous two variables also define a default value, but I don't have a strong preference for it. Do you think we should remove it?

mattsherman: Of course. Since it comes after other params with defaults, it too should have a default. Nevermind!

client/customer-effort-score-tracks/data/reducer.js
```diff
@@ -0,0 +1,42 @@
...
+				queue: action.queue,
+			};
+		case TYPES.ADD_CES_SURVEY_TRACK:
+			// Prevent duplicate
+			const registeredTracks = new Set(
+				state.queue.map( ( track ) => track.action )
+			);
+			if ( registeredTracks.has( action.action ) ) {
+				return state;
+			}
```
mattsherman, line 24: Why map to a `Set` first vs just using the queue directly to find if a duplicate action is in the queue?

mattsherman (changes requested):
Overall, this seems to work as expected. Nice job so far on this!

I left some comments, mainly around naming. I think things would be easier to understand if the naming of the actions was updated a bit to differentiate between queuing up the showing of the CES survey (which is what this code is doing) vs logging a Tracks event for the survey result (which is what the PHP code handles).

client/analytics/report/categories/index.js
```diff
@@ -55,6 +62,10 @@ export default class CategoriesReport extends Component {
 				: 'category';
 		}
 
+		filters[ 0 ].filters.find(
+			( item ) => item.value === 'compare-categories'
+		).settings.onClick = addCesSurveyTrackForAnalytics;
```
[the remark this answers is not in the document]
moon0326: I initially handled it then changed it back as per @becdetat 's suggestion at https://github.com/woocommerce/woocommerce-admin/pull/5649#discussion_r525617526. What do you think about the suggestion?

client/analytics/components/report-filters/index.js
```diff
@@ -135,12 +139,18 @@ class ReportFilters extends Component {
...
-} )( ReportFilters );
+export default compose(
+	withSelect( ( select ) => {
+		const { woocommerce_default_date_range: defaultDateRange } = select(
+			SETTINGS_STORE_NAME
+		).getSetting( 'wc_admin', 'wcAdminSettings' );
+		return { defaultDateRange };
+	} ),
+	withDispatch( ( dispatch ) => {
+		const { addCesSurveyTrackForAnalytics } = dispatch( CES_STORE_KEY );
```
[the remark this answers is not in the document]
moon0326: I think backend and frontend APIs have a slightly different purpose in this case. The backend saves the queue in the database to be used at a later time, but the tracks added from the frontend never get saved. From the API user's (including us and a 3PD dev) point of view, the user wants to add a survey immediately instead of queueing it to be displayed later. 

However, I agree that this can be confusing for those who use both backend and frontend APIs.

cc @becdetat for additional thoughts

client/customer-effort-score-tracks/data/reducer.js
moon0326: That's a good catch 👍 I've updated it in https://github.com/woocommerce/woocommerce-admin/pull/5649/commits/9735e64b91d7762a342f136da051d5bf8acdcf6b

client/analytics/components/report-filters/index.js
```diff
@@ -19,36 +20,38 @@ import { recordEvent } from '@woocommerce/tracks';
...
 			report,
 			...omitBy( data, isUndefined ),
 		} );
 	}
 
-	trackFilterSelect( data ) {
-		const { report } = this.props;
+	onFilterSelect( data ) {
+		const { report, addCesSurveyTrackForAnalytics } = this.props;
+		if ( data.filter === 'single_product' ) {
```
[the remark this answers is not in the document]
moon0326: This is my mistake. I should have left more details in the comment. Updated comments in the latest commit.

TL;DR: The event gets triggered whenever a user selects/switches a filter type -- Single Product, Comparison, and Advanced Filters. The later two filter types have a button to execute a query, so we don't want to trigger a CES query when a user selects the Comparison or Advanced Filter. We want to trigger a CES survey when the user clicks a button for those two filter types.

client/analytics/components/report-filters/index.js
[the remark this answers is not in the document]
mattsherman: That makes sense -- thanks for the clarification. But, this still isn't adding the Tracks log entry, it is just showing the CES survey, right? I think it is best to be as clear as possible in our naming in this regard so that others looking at the code don't misinterpret what is going on.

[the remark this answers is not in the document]
mattsherman: Also, from the API user's standpoint, they are abstracted from where the survey is getting saved, right? If so, we shouldn't use the "Track" terminology.

client/analytics/components/report-filters/index.js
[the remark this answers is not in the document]
mattsherman: Thanks for adding the comment and explaining things -- makes sense!

client/analytics/report/categories/index.js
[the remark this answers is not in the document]
mattsherman: That's a very good point. A fast-fail is often much more useful at tracking down things.

My main concern is that if this errors, the whole CategoriesReport won't be rendered, which seems like a pretty big failure from a user standpoint.

Maybe some automated testing should ultimately be added for the report components so that we can be assured that we will quickly notice if these filters get removed or renamed. But, that could be a future PR.

I'm okay with the fast-fail approach for now.

mattsherman (commented):
Thanks for the explanations and updates @moon0326!

I think at this point, it would be good to see what @becdetat and/or @adrianduffell thinks in regards to the naming issues. I'd personally like to see any reference to "Track"/"Tracks" removed from the client code/API unless it is specifically referring to Tracks. My understanding is that we are designing things such that the backend storage of CES results can be changed by other users of CES.

client/analytics/components/report-filters/index.js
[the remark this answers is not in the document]
moon0326: @mattsherman I think it tracks as the function basically re-uses the [CustomerEffortScoreTracks](https://github.com/woocommerce/woocommerce-admin/blob/add/5349-trigger-ces-on-filtering-analytics/client/customer-effort-score-tracks/customer-effort-score-tracks.js). However, I'm not sure if we want to track 3PD surveys to our system though. I think we need further discussion on that.

client/analytics/report/categories/index.js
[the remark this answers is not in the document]
moon0326: @mattsherman That's a good point. I will try to add a test case if time allows 👍

client/analytics/components/report-filters/index.js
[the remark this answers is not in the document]
adrianduffell: There's some more background about the goals for abstraction and external use in the [umbrella issue](https://github.com/woocommerce/woocommerce-admin/issues/5259). 

I  think we should align closely with the backend naming - `addCesSurveyForAnalytics` or `enqueueCesSurveyForAnalytics` LGTM.

client/customer-effort-score-tracks/data/action-types.js
```diff
@@ -0,0 +1,6 @@
+const TYPES = {
+	SET_CES_SURVEY_TRACKS_QUEUE: 'SET_CES_SURVEY_TRACKS_QUEUE',
+	ADD_CES_SURVEY_TRACK: 'ADD_CES_SURVEY_TRACK',
```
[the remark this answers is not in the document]
adrianduffell: Would this work as just `SET_CES_SURVEY_QUEUE` and `ADD_CES_SURVEY`?

adrianduffell:
This looks great overall @moon0326 👍 

I think we just need to fine-tune the naming here. It seems like "track" is being used refer to different things. I agree about avoiding the use of the word. I think we could aim to use it only when there is usage of the [Tracks package](https://github.com/woocommerce/woocommerce-admin/tree/main/packages/tracks).

I left some suggestions in the code comments.

moon0326:
@adrianduffell Thank you! It looks like everyone agrees on removing *Track(s)* from the method name. I will go ahead and remove them 👍  Thank you @adrianduffell  and @mattsherman  for the suggestions.

client/analytics/components/report-filters/index.js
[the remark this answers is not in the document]
moon0326: @adrianduffell Thank you for the suggestion. Updated in https://github.com/woocommerce/woocommerce-admin/pull/5649/commits/c8ba3dd0c965a554610f4854557ef60ab0a3bd6f

mattsherman (approved):
Thanks for making the naming changes @moon0326 !

I have looked at the latest code changes -- things look good.

I ran through the testing instructions quickly again too -- things still work as expected.

moon0326:
@mattsherman Thank you Matt and everyone! This PR has been fun to work on 😄  I will run a few more tests again and merge 👍

PR #5603, opened 2020-11-11
Update Woo tested and min versions.

timmyc:
Fixes #5561

Per the conversation in the above issue, this branch updates our min supported Woo version to be the current release ( this will be 4.8 when this ships ) -3, so 4.5, and also updated the tested up to to match the latest Woo version.

PR #5650, opened 2020-11-17
Restore Autoloading of WooCommerce Classes in PHP Tests

adrianduffell:
When running PHPUnit locally, the [early exit call](https://github.com/woocommerce/woocommerce/blob/master/includes/class-wc-install.php#L12) in the `class-wc-install.php` is triggered and prevents the tests from running.

This PR restores autoloading of WooCommerce classes in the tests to remove the need to directly require `class-wc-install.php`. 

### Additional information

WooCommerce class autoloading was stopped in PR https://github.com/woocommerce/woocommerce-admin/pull/5462 when `composer install` was removed from the WooCommerce build in `install-wp-tests.sh`.

In PR https://github.com/woocommerce/woocommerce-admin/pull/5603 a change was made to `bootstrap.php ` to require `class-wc-install.php` directly.

### Testing instructions

- Run PHPUnit tests locally and ensure test results are output.

### DIFF
--- a/bin/install-wp-tests.sh
+++ b/bin/install-wp-tests.sh
@@ -180,8 +180,10 @@ install_deps() {
 	git clone --depth 1 --branch $LATEST_WC_TAG https://github.com/woocommerce/woocommerce.git
 
 	# Bring in WooCommerce Core dependencies
- 	cd "woocommerce"
+ 	composer self-update --1
+	cd "woocommerce"
  	composer install --no-dev
+	composer self-update --2
 
 	cd "$WP_CORE_DIR"
 	php wp-cli.phar plugin activate woocommerce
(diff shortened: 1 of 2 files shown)


adrianduffell:
@jeffstieler What are your thoughts on bringing back `composer install` for WooCommerce core in `bin/install-wp-tests.sh`? 

I think it might have been removed in #5462 because of issues with Composer 2. I've worked around that by downgrading to Composer 1 in the script.

bin/install-wp-tests.sh
```diff
@@ -179,6 +179,12 @@ install_deps() {
 	# As zip file does not include tests, we have to get it from git repo.
 	git clone --depth 1 --branch $LATEST_WC_TAG https://github.com/woocommerce/woocommerce.git
 
+	# Bring in WooCommerce Core dependencies
+ 	composer self-update --1
```
adrianduffell, line 183: A downgrade of Composer is required since the latest release of WooCommerce is not compatible with version 2.

jeffstieler (approved):
> @jeffstieler What are your thoughts on bringing back composer install for WooCommerce core in bin/install-wp-tests.sh?
> 
> I think it might have been removed in #5462 because of issues with Composer 2. I've worked around that by downgrading to Composer 1 in the script.

Yep - I removed it because it was causing issues and seemed unnecessary. This PR is testing well for me. :shipit:

PR #5655, opened 2020-11-17
Fixed orders panel flickering

### LINKED ISSUE #5609: Home Screen: Orders Panel flashes empty state during load ( Gutenberg only )
**Describe the bug**
On 1.7.0 I noticed that when the orders data is loading to the activity panel, the empty/no order state is shown briefly a few times:

https://cloudup.com/cDOMVhYp3mg

**To Reproduce**

1. Install 1.7
2. Have some orders with actionable status
3. Hard refresh the home screen and note the behavior seen in the video above

* WooCommerce 4.6.1
* Atomic Site
* Gutenberg 9.1.1 installed and active.

octaedro:
Fixes #5609

This PR fixes the orders panel flickering when Gutenberg is activated.

### Screenshots
Screen Capture on 2020-11-17 at 14-14-38

- Be sure the `Gutenberg` plugin is installed and activated.
- Go to the `Home` screen.
- Verify no flickering happens in the order panel.
- Go to `Analytics` > `Settings` and select different combinations of the checkboxes under `Actionable Statuses` (**don't select unregistered statuses**).
- Go back to the `Home` screen.
- Again, verify there isn't any flickering in the order panel.

### Changelog Note:
Fix: Fixed orders panel flickering

### DIFF
--- a/client/homescreen/activity-panel/orders/index.js
+++ b/client/homescreen/activity-panel/orders/index.js
@@ -304,7 +304,7 @@ export default withSelect( ( select, props ) => {
 		REPORTS_STORE_NAME
 	);
 
-	if ( ! orderStatuses.length && ! countUnreadOrders ) {
+	if ( ! orderStatuses.length && countUnreadOrders === 0 ) {
 		return { isRequesting: false };
 	}
 
--- a/client/homescreen/activity-panel/orders/index.js
+++ b/client/homescreen/activity-panel/orders/index.js
@@ -364,7 +364,6 @@ export default withSelect( ( select, props ) => {
 	if (
 		countUnreadOrders === null ||
 		typeof reportOrders === 'undefined' ||
-		typeof actionableOrders === 'undefined' ||
 		( reportOrders.length && ! actionableOrders.length ) ||
 		isRequesting
 	) {
--- a/client/homescreen/activity-panel/orders/index.js
+++ b/client/homescreen/activity-panel/orders/index.js
@@ -304,8 +304,8 @@ export default withSelect( ( select, props ) => {
 		REPORTS_STORE_NAME
 	);
 
-	if ( ! countUnreadOrders ) {
-		return { isRequesting: countUnreadOrders !== 0 };
+	if ( ! orderStatuses.length && ! countUnreadOrders ) {
+		return { isRequesting: false };
 	}
 
 	// Query the core Orders endpoint for the most up-to-date statuses.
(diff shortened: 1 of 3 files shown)


client/homescreen/activity-panel/orders/index.js
```diff
@@ -361,10 +361,17 @@ export default withSelect( ( select, props ) => {
...
 	let orders = [];
 
-	if ( reportOrders && reportOrders.length ) {
-		if ( actionableOrders && ! actionableOrders.length ) {
-			return { isRequesting: true };
-		}
+	if (
+		countUnreadOrders === null ||
+		typeof reportOrders === 'undefined' ||
+		typeof actionableOrders === 'undefined' ||
```
joshuatf, line 367: Can `actionableOrders` be `undefined`?  Looks like an array is created above.

client/homescreen/activity-panel/orders/index.js
```diff
@@ -304,8 +304,8 @@ export default withSelect( ( select, props ) => {
 		REPORTS_STORE_NAME
 	);
 
-	if ( ! countUnreadOrders ) {
-		return { isRequesting: countUnreadOrders !== 0 };
+	if ( ! orderStatuses.length && countUnreadOrders === 0 ) {
```
joshuatf, line 307: Curious why we separate this logic from the condition below.  It's a little hard to follow, but it seems both are related to whether or not we're still requesting.

joshuatf (commented):
Very nice!  This PR fixes the flickering in the orders panel.  Left a couple questions around the organization of logic, but otherwise testing well.

client/homescreen/activity-panel/orders/index.js
octaedro: Nice catch @joshuatf! We don't need that check. It was removed in the commit dcac5b1

client/homescreen/activity-panel/orders/index.js
octaedro: If we know that at least one `order status` was selected and there aren't orders to show, it's not necessary to continue with the other requests that we do later in the code, [here](https://github.com/woocommerce/woocommerce-admin/blob/b0600d7071ea0f333cff7dc969b8e57297e7b572/client/homescreen/activity-panel/orders/index.js#L319) and [here](https://github.com/woocommerce/woocommerce-admin/blob/b0600d7071ea0f333cff7dc969b8e57297e7b572/client/homescreen/activity-panel/orders/index.js#L355).
So in order to have a more efficient behavior, I added this check before doing those calls to avoid continuing when we already have all the data that we need.
But for sure the logic is a little confusing and a future refactor would be a good idea.

octaedro:
Thank you @joshuatf for your review. I added the change you suggested and another small fix.
Could you give this PR another view 👀 ?

joshuatf (approved):
Testing well and seems to resolve any flickering.  Thanks for the fix here, Fernando!  LGTM 🚢

client/homescreen/activity-panel/orders/index.js
joshuatf: >But for sure the logic is a little confusing and a future refactor would be a good idea.

➕ 1️⃣

octaedro:
Cherry picked into `release/1.7.2`

PR #5657, opened 2020-11-17
Remove importing the class-wc-install.php to allow local tests to run

### LINKED ISSUE #5656: PHP unit tests fail to run locally
**Describe the bug**
When trying to run `composer test` within VVV on my local machine it doesn't do anything, nor shows any errors.

Reason seems to be because of this [line](https://github.com/woocommerce/woocommerce-admin/blob/main/tests/bootstrap.php#L58) which contains `defined( 'ABSPATH' ) || exit;`. ABSPATH is undefined and thus it exists the tests.
This line was added to fix Travis builds in this PR: https://github.com/woocommerce/woocommerce-admin/pull/5603

**To Reproduce**

1. Try running `composer test` in your local environment with the latest of `main`
2. It should immediately be finishing.

**Expected behavior**
It should successfully run all the php unit tests.

louwie17:
Fixes #5656 

Removing the `class-wc-install.php` require as at this stage locally the `ABSPATH` is not defined and it exits the tests instead of running them.

- Checkout this branch
- Try running the php unit tests using `composer test`
- It should successfully run all the tests

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -54,9 +54,6 @@ class WC_Admin_Unit_Tests_Bootstrap {
 		// load test function so tests_add_filter() is available.
 		require_once $this->wp_tests_dir . '/includes/functions.php';
 
-		// load the wc installer.
-		require_once $this->wc_core_dir . '/includes/class-wc-install.php';
-
 		// load WC.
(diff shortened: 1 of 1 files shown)


samueljseay:
I think we can close this now in favor of #5650

PR #5660, opened 2020-11-17
Navigation: Renaming navigation body classes to be less generic

### LINKED ISSUE #5658: Navigation folded/expanded body classes are too generic
The body classes used to govern how the new navigation feature are too generic, being `is-folded` and `is-expanded`. 

We should rename these to something that indicates that they're used on the nav feature.

joelclimbsthings:
Fixes #5658

Quick PR to rename the body classes used for navigation from `is-folded` & `is-expanded` to `is-wc-nav-folded` & `is-wc-nav-expanded` as mentioned (and missed) in [this earlier PR](https://github.com/woocommerce/woocommerce-admin/pull/5616#discussion_r523135657).

- Check out branch
- Navigation to WP-Admin -> WooCommerce
- Click top-left logo to fold and expand navigation
- Ensure behavior is still as expected

### DIFF
--- a/client/navigation/components/header/index.js
+++ b/client/navigation/components/header/index.js
@@ -20,8 +20,8 @@ const Header = () => {
 	const siteUrl = getSetting( 'siteUrl', '' );
 	const isScrolled = useIsScrolled();
 	const navClasses = {
-		folded: 'is-folded',
-		expanded: 'is-expanded',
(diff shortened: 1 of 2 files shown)


psealock (approved):
Thanks Joes, this makes a lot of sense :shipit:

PR #5665, opened 2020-11-18
Remove SlotFill need for clientside id's and make identifiers more unique

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/5626
Fixes https://github.com/woocommerce/woocommerce-admin/issues/5458

This is an alternate to https://github.com/woocommerce/woocommerce-admin/pull/5642

id's were added in wc-admin registered pages so that there could be identical ids matched between `wcNavigation.menuItems` (which is populated by PHP) and client functions `getPages` and `getReports`. Instead, this PR removes those id's and uses the `path` to determine uniqueness. This removes the requirement for wc-admin connected extensions to have to update their filters to add id's.

id's in PHP were also made less generic when being registered by the Nav. This avoids duplication, such as was the case with `Settings > Products`.

1. Ensure all wc-admin pages can navigate without refreshing the page, including Home, Settings > Analytics, and Customers.
2. Ensure Settings > Products is a functioning Nav item.


client/layout/navigation.js
```diff
@@ -53,21 +57,24 @@ const NavigationPlugin = () => {
...
-							{}
-						) }
-						type="wc-admin"
-					>
-						{ item.title }
-					</Link>
-				</WooNavigationItem>
-			) ) }
+			{ reports.map( ( item ) => {
+				const id = `wc_admin/wc-admin&path=/analytics/${ item.report }`;
```
joshuatf, line 61: Nice 💯

src/Features/Navigation/CoreMenu.php
```diff
@@ -57,7 +57,7 @@ public static function get_setting_items() {
 					'parent'     => 'settings',
 					'title'      => $setting,
 					'capability' => 'manage_woocommerce',
-					'id'         => $key,
+					'id'         => 'settings/' . $key,
```
joshuatf, line 60: Curious why we using `settings/` here but `woocommerce/` as the prefix below?

I would expect `wc-settings/` and `wc-status/` or maybe `settings/` and `tools/`.

Does this also mean we will be keeping IDs unique and not concatenating IDs with parent IDs?

src/Features/Navigation/CoreMenu.php
```diff
@@ -185,22 +185,22 @@ public static function get_items() {
 					'parent'     => 'tools',
 					'title'      => __( 'System status', 'woocommerce-admin' ),
 					'capability' => 'manage_woocommerce',
-					'id'         => 'system-status',
+					'id'         => 'woocommerce/system-status',
```
joshuatf, line 188: Heads up that these will most likely be moved based on recommendations in https://github.com/woocommerce/woocommerce-admin/pull/5630#pullrequestreview-533786764

joshuatf (commented):
Very nice clean-up and testing well.  Left a couple questions prior to approving as this affects some other PRs we're working on.

joelclimbsthings:
Great work @psealock , I think this will be a good approach. Tested & works well.

I'm just curious, part of the reasoning behind this change is to make it so other extensions that register pages (via `wc_admin_register_page()`) will have client-side registration automated as well, right? Is this PR meant to achieve that, and if so, how are the `path` properties relayed from the server to the client? 

It looks like we're currently just registering the pages that are manually added within `controller.js`, which for now only includes our own pages as far as I can tell. It's entirely possibly I'm missing something here; I haven't looked _too_ closely, but I was curious!

psealock:
>I'm just curious, part of the reasoning behind this change is to make it so other extensions that register pages (via wc_admin_register_page()) will have client-side registration automated as well, right?

The client side won't be automated, but those who have already registered a client side page won't need to touch that code.

>It looks like we're currently just registering the pages that are manually added within controller.js, which for now only includes our own pages as far as I can tell. 

There is a filter to add pages, `woocommerce_admin_pages_list`. Below is WC Pay using that filter and there may be more out there in the wild.

https://github.com/Automattic/woocommerce-payments/blob/c7f9d58e2d20fc5fb10d3fa5d205f698bcf3b8c6/client/index.js#L35-L40

>how are the path properties relayed from the server to the client?

Any app, including wc-admin, using `wc_admin_register_page` to register pages using PHP will supply a `path`:

https://github.com/woocommerce/woocommerce-admin/blob/cada1266bf9fcf6cf6fad6cbc3b133aabe5e032c/src/Features/Navigation/CoreMenu.php#L135-L152

The item data, including an `id` derived from the `path` parameter (at least in this PR) is relayed to the client via `window.wcNavigation.menuItems`.

Since the `woocommerce_admin_pages_list` filter also supplies that same `path`, we can infer an id. 

Extensions registering with the Nav on their own, ie outside of wc-admin`, are free to match id's in any way they choose in order to get slotFill to work.

### Caveat

If we were starting from scratch, I think it would be prudent to require extensions extending wc-admin to supply a client side id that matched the one supplied in PHP. If we think asking extensions to update their `woocommerce_admin_pages_list` filter is worth it in order to have explicitly defined `id` instead of inferring one from a `path` parameter, then these changes won't be required and we can stick with https://github.com/woocommerce/woocommerce-admin/pull/5642. I'm open to reconsidering the whole premise here because perhaps asking devs to update causes pain at first but results in code clarity on the long term. cc @joshuatf for your input as well.

joshuatf:
>I'm open to reconsidering the whole premise here because perhaps asking devs to update causes pain at first but results in code clarity on the long term.

I agree with this, so let's choose whichever makes the most sense going forward.  We can handle a little clean-up work if we need to.

That said, I'm thinking keeping the `id` is a more sane approach than trying to identify via a `path`.  That does not mean a plugin (read: WooCommerce Admin) cannot opt to use an item's path as the main identifier.

This means that if a plugin wants to register multiple items with the same path, they can do so, but they can still have the benefits of not passing the ID from server to client if they choose to use IDs that match their paths.

psealock:
>I agree with this, so let's choose whichever makes the most sense going forward. We can handle a little clean-up work if we need to.

I'm starting to lean this way. At least if wc-admin extensions don't update, links will still work but will refresh page.

>That does not mean a plugin (read: WooCommerce Admin) cannot opt to use an item's path as the main identifier.

To be clear, thats what this PR does. It only deals with wc-admin and its extensions. Any other extensions can use supply any `id` they'd like so that it matches the one on in PHP.

joelclimbsthings:
Thanks for your earlier comment @psealock ; that definitely clarified a few things for me!

I've been delving a bit further into the various methods we have for approaching navigation registration, so allow me to do a quick brain dump here to document my findings. It looks like two basic approaches:

**Method One: _Wc-admin_ Extension**
1) Register the extension with `wc_admin_register_page()`
This will register with the navigation seemingly _only_ if the `nav_args` property is present. Is this only expected to be used by our own pages, deferring others to instead use the `plugin` centric hooks mentioned below?

With this PR, using this method your item ID is generated based on a derivative of your `path` automatically. 

2) For client-side registration, use the `woocommerce_admin_pages_list` filter and provide the required properties including the same `path` variable as above. This will result in the same ID as the server-side registration above, via the logic included in this PR.
_Note: As far as I can tell this filter isn't currently documented anywhere_ 

**Method Two: Non-Extension**
1) Register directly with the navigation via `Menu::add_plugin_item()` & `Menu::add_plugin_category()`. With this you provide whatever ID you'd like, with no enforcement on our side to try to ensure that it's unique. We should print out some warning if there's a duplicate ID already present.
2) You also need to add a screen if relevant via `Screen:add_screen()`. 
3) For client-side registration, you use Slot/Fill via the `<WooNavigationItem>`, and provide the same ID as you used to register server-side.

**Case Study: WC-Payments**

As a case-study on this, I was also looking into WC-Payments, and specifically [this PR](https://github.com/Automattic/woocommerce-payments/pull/1067) that @joshuatf kindly created to integrate with the navigation. After this PR, this is what's happening with WC-payments specifically:

1) They're registering with `wc_admin_register_page()` as they are extending `wc-admin`. 
Although, since they don't include `nav_args` in this registration, there _are not_ registered with the new navigation at all via this hook.

2) As of the mentioned PR they'll be calling `Menu::add_plugin_category()` & `Menu::add_plugin_item()` to add themselves to the navigation, with hard-coded IDs such as "woocommerce-payments-deposits"

3) They're still using the `woocommerce_admin_pages_list` filter to register client-side, which is causing their IDs to be generated via the `path` via the logic in this PR. 

So, the result being that they _are_ being included via the client & server, but with differing IDs. For example--the `connect` page:

Client: `wc_admin/wc-admin&path=/payments/connect`
Server: `woocommerce-payments-connect`

So, all that said, I'm just trying to get clarity on _what we want to happen_. It looks like we actually don't want outside extensions registering with the nav through the `wc_admin_register_page()` as it is now, since that would also enable them to create nav items outside of the "Plugins" category. 

Given that, perhaps we don't want them to use the `woocommerce_admin_pages_list` filter either? Instead we would just direct them to implement Slot/Fill in order to register client-side. 

That would leave us with these steps for outside extensions:

1) Still call `wc_admin_register_page()`. This doesn't add them to the new nav (due to missing `nav_args`), but registers within `PageController()`. I haven't explored this in depth, but I suspect this is still a necessary step?

2) Register with navigation via `Menu::add_plugin_category()` & `Menu::add_plugin_item()`. They choose their own IDs for each item.

3) To avoid page refreshes for React-based plugins, replace any use of the `woocommerce_admin_pages_list` filter with the new Slot/Fill implementation (or does that filter still serve another purpose?). They'd manage the IDs however they want and ensure that they use the same ones here that they registered in the previous step. 

And internally we'd continue doing this:

1) Register pages via `wc_admin_register_page()`, and _include_ `nav_args` so that items will be registered automatically with navigation. This will use our `path` for the IDs, and allow us to create menu items anywhere in the hierarchy. 
2) Continue to use Slot/Fill as we are within `controller.js` and `navigation.js` using the IDs derived from the `path` properties.

*phew!* Maybe this should have been a P2 post :open_mouth:

joshuatf:
>You also need to add a screen if relevant via Screen:add_screen().

For clarification, this is done for you automatically when adding items, you don't need to do this manually unless you're registering a screen that you didn't register an item for.  https://github.com/woocommerce/woocommerce-admin/blob/main/docs/features/navigation.md

>So, all that said, I'm just trying to get clarity on what we want to happen. It looks like we actually don't want outside extensions registering with the nav through the wc_admin_register_page()

I don't think we do.  I think extensions should be explicit in their item registration.  Furthermore, the rules in WCA pages allow registration in the core menu and not the extensions menu.  The filters and page registrations in WCA is exploitable and opens up the core menu to everyone.

A better solution might be to control this from the navigation feature.  It looks like a possible fix for this right now is to retrieve all pages from the nav and filter those with a path prefix of `/analytics` and register those.  Same rules could apply to customer and marketing pages.

psealock:
Thanks for the reviews and comments. I have addressed the feed back in https://github.com/woocommerce/woocommerce-admin/pull/5665/commits/5da98f95c6e4e871c1fe4195572da3b96a825b1f.

### Changes

* Id's are now unique with the addition of `parent-`. I choose the dash instead of the slash to match ids already in place

Screen Shot 2020-11-24 at 10 25 34 AM

* Since the id's are hard coded, things like `nav_args` needed to be updated
* This change will likely require changes to the PRs made in extensions

### Notes

* There is an extra "Settings" under Settings category. I'm having a hard time figuring out where it comes from.
* I have confirmed that extending wc-admin like below will automatically handle the SlotFill 🎉 

```php
wc_admin_register_page(
		[
			'id'     => 'example-wc-page',
			'title'  => __( 'Examples WC Page', 'woocommerce-payments' ),
			'path'   => '/examples',
			'nav_args' => array(
				'parent' => 'woocommerce',
				'is_top_level' => true,
				'menuId' => 'plugins',
			),
		]
	);
```

joelclimbsthings:
> There is an extra "Settings" under Settings category. I'm having a hard time figuring out where it comes from

I think I ran into the same issue in my earlier PR--here was my fix: 
https://github.com/woocommerce/woocommerce-admin/pull/5642/commits/470575315d7581c9c50eb735b4d65c79e364513f

>I have confirmed that extending wc-admin like below will automatically handle the SlotFill tada
:partying_face:

psealock:
Thanks @joelclimbsthings, works like a charm

joshuatf (commented):
This is testing well for me, but just want to confirm that our solution to the original issue is to request more specific IDs, is that right?

https://github.com/woocommerce/woocommerce-admin/pull/5642 allows multiple identical IDs under different parents, whereas this PR pushes to make existing IDs more specific to circumvent this.

I think either approach is okay, but worth noting that an ID that was autogenerated, like a post type page `shop_order-add-new` could still run into the duplicate ID issue.

psealock:
>a post type `page shop_order-add-new` could still run into the duplicate ID issue.

Since the parent is supplied, we can alter `get_post_type_items` to programatically prepend the parent. Would that work?

```php
$order_items   = Menu::get_post_type_items( 'shop_order', array( 'parent' => 'woocommerce-orders' ) );
```

joelclimbsthings:
Tested and working well for me as well :+1: 

In [my other PR](https://github.com/woocommerce/woocommerce-admin/pull/5642) I was prepending the parent within `Menu::add_item()` which would make IDs a bit more unique no matter how the registration happened (`wp_admin_register_page` or otherwise), whereas here we're making the IDs more unique just within our own extensions, as @joshuatf mentioned. I'm good with this, actually, since modifying the ID as I was can actually be more confusing, but something to keep in mind since third party extensions could be tempted to use really generic IDs such as `payments` or `emails`. 

I think for this reason it's still a good idea to print out a warning in the `error_log()` or similar when a duplicate item registration is attempted. We can do that in this PR, or a subsequent one that is addressing some of the other related action items from [the recent P2 post](https://teamisotope.wordpress.com/2020/11/20/navigation-registration-clarifying-our-approach/).

joshuatf:
>I think for this reason it's still a good idea to print out a warning in the error_log() or similar when a duplicate item registration is attempted.

👍  agreed.  I'm good with relying on unique IDs; let's throw in some error logging here to make it easier to debug.

joshuatf:
>Since the parent is supplied, we can alter get_post_type_items to programatically prepend the parent. Would that work?

I think that works.  With the error logging in place, I don't feel too strongly one way or another, but can't see it hurting to have the parent prepended.

psealock:
Thanks for the continued feed back everyone 😄 

I added the error logging from @joelclimbsthings 's PR in https://github.com/woocommerce/woocommerce-admin/pull/5665/commits/c5e90ccd547f46bfd4a989662cb3102acd652029 and made the post type id's unique in https://github.com/woocommerce/woocommerce-admin/pull/5665/commits/c5290e81fe19144026ae91fc87f1442a04647d9d

### Bonus testing

If you feel compelled, my examples repo https://github.com/psealock/examples registers a Hello World page by extending wc-admin. If you load that plugin, you can test the following.

* Confirm that clicking Example WC Admin Page does not incur a page refresh when already on a wc-admin page
* Duplicate an id in any call of `Menu::add_plugin_item` to see the error logging appear

joelclimbsthings (approved):
I like it! Tested (including bonus steps) and all works well :+1: 

Approved.

joshuatf (approved):
Looks fantastic!  LGTM 🚢

PR #5668, opened 2020-11-18
Only import the Gridicons we need, to reduce package size.

samueljseay:
Currently whenever we use `<Gridicon>` we import the entire library, despite only using a handful of icons across the entire codebase. This does not allow us the opportunity to tree-shake them.

I ran `npm run clean && npm run analyze` before and after adding this change and we end up saving ~120kb total over our entire bundle.

Analyze before change: total bundle size: 3.81MB
Analyze after change: total bundle size: 3.69MB

### Testing Instructions

There are 2 things to test:

1. All the icons work as before. Some of them pass a specific size so confirm that they look ok. Sorry its not easy to summarise this better. Its a wide spanning change over a few different views in the app

2. Bonus: check for yourself that there is indeed a size impact. Ensure that every time you analyze you run this: `npm run clean && npm run analyze`

### DIFF
--- a/client/dashboard/dashboard-charts/index.js
+++ b/client/dashboard/dashboard-charts/index.js
@@ -4,7 +4,8 @@
 import { __ } from '@wordpress/i18n';
 import classNames from 'classnames';
 import { Fragment, useState } from '@wordpress/element';
-import Gridicon from 'gridicons';
+import LineGraphIcon from 'gridicons/dist/line-graph';
+import StatsAltIcon from 'gridicons/dist/stats-alt';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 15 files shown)


joshuatf (approved):
Icons all seem to be working; love the smaller bundle size! 💯 

Could be done in a follow-up, but it would be great to also do the same in the `List` component and in readmes so other devs don't also import the whole Gridicons library.

Also unrelated, but found while perusing icons- seems like the vertical alignment is off here:
Screen Shot 2020-11-18 at 2 35 11 PM

samueljseay:
Thanks for the review @joshuatf I'll see if I can fix that vertical alignment in a follow up PR.

Issue #5559, opened 2020-11-09
Dependency Extraction Webpack Plugin: Include WC_ADMIN_APP if slot/fill being used 


psealock:
When using Woo's Dependency Extraction Webpack Plugin it should handle the use case where slot/fill is being used and `WC_ADMIN_APP` needs to be declared as a dependency when registering scripts via `wp_register_script`

**Describe the solution you'd like**

Detect use of slotFill to insert `WC_ADMIN_APP` into dependencies. Maybe use of `@woocommerce/navigation` can be enough.

**Should this be prioritized? Why?**

Developers using the starter pack via `npm run create-wc-extension` now need to add the dependency explicitly. We should handle this for them.

### Deliverables

* An extension using a slotFill should have `WC_ADMIN_APP` included in its dependencies.
* Check if `WC_ADMIN_APP` is defined or just use the string `wc-admin-app`.

https://github.com/woocommerce/woocommerce-admin/blob/d5a0b32126f655bf3703aac21826397bf5a0df40/src/FeaturePlugin.php#L149

psealock:
Gutenberg PR to include some default dependencies https://github.com/WordPress/gutenberg/pull/27020

psealock:
This turned out to be a problem in wc-admin itself. See a fix here https://github.com/woocommerce/woocommerce-admin/pull/5722

PR #5671, opened 2020-11-18
Dep Ext Webpack Plugin: Map wc-navigation to wc-admin

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/5559

When using SlotFill from `@woocommerce/navigation` a dependency on `wc-admin-app` must be declared. This PR does that which will ultimately also bring in a dependency on `wc-navigation`.

https://github.com/woocommerce/woocommerce-admin/blob/62501ce94073a9646e560bfea7222309a3c5b04f/src/Loader.php#L481-L494

See discussion in https://github.com/WordPress/gutenberg/pull/27020 for previous attempts to solve this one.

1. `npm run build`.
2. `cd packages/dependency-extraction-webpack-plugin`.
3. `npm link`
4. `cd ../..`
5. `npm run create-wc-extension`
6. Follow the instructions and name your new app.
7. `cd ../<new app name>`
8. `npm install`
9. `npm link @woocommerce/dependency-extraction-webpack-plugin`
10. `npm start`
11. Add this code to `src/index.js`
```js
import { Table } from '@woocommerce/components';
import { WooNavigationItem } from '@woocommerce/navigation';

console.log( Button, Table, WooNavigationItem );
```
12. Check that `build/index.assets.php` correctly indicates dependencies on `wc-admin-app` and `wc-components`

### DIFF
--- a/bin/starter-pack/_package.json
+++ b/bin/starter-pack/_package.json
@@ -22,6 +22,6 @@
 	"devDependencies": {
 		"@wordpress/scripts": "^12.2.1",
 		"@woocommerce/eslint-plugin": "1.0.0-beta.0",
-		"@woocommerce/dependency-extraction-webpack-plugin": "1.0.1"
+		"@woocommerce/dependency-extraction-webpack-plugin": "1.1.0"
 	}
 }
--- a/packages/dependency-extraction-webpack-plugin/CHANGELOG.md
+++ b/packages/dependency-extraction-webpack-plugin/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 1.1.0
+
+-   Fix: Map wc-navigation to wc-admin-app
+
 # 1.0.1
 
 -   Fix: Avoid tanspiling packaged code.
--- a/packages/dependency-extraction-webpack-plugin/package.json
+++ b/packages/dependency-extraction-webpack-plugin/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@woocommerce/dependency-extraction-webpack-plugin",
-	"version": "1.0.1",
+	"version": "1.1.0",
 	"description": "WooCommerce Dependency Extraction Webpack Plugin",
 	"author": "Automattic",
 	"license": "GPL-2.0-or-later",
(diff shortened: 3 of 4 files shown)


louwie17 (approved):
Looks great, and tested well :) Just had to switch steps 8 and 9 around in testing instructions.

Approving this, but in relation to [this last comment](https://github.com/WordPress/gutenberg/pull/27020#issuecomment-729627352) will navigation always depend on `wc-admin-app`, or could there be exceptions? Where turning it on/off would be needed. My bad if this is a given 🙈

psealock:
Thanks for the review @louwie17 !

@senadir What do you think about this solution? Only `@woocommerce/navigation` is affected.

>If you're going to implement this, I'd prefer if it can be turned off so we can use that package in our code.

I'd rather make a tool useable internally and externally than cater to this scenario. If this change makes too many assumptions to be useable in the case of Blocks, it may be best to communicate this need to extension developers instead. I can also make the dependency explicit in the starter pack. What do you think?

senadir:
Hey, sorry for the late response here.
Taking from this, there's no dependency on `wc-navigation` anymore? This assumes that navigation is only going to be used in the context of wc-admin and that's fine, a custom rule can be written to revert this PR is navigation was used outside wc-admin.

My previous comment on that PR was if any package has `@woocommerce/*` in its name, it would also add `wc-admin-app` as a dependency.
The above seems like a sane workaround that doesn't effect anyone.

psealock:
>Taking from this, there's no dependency on wc-navigation anymore?

There would be because wc-admin has a dependency on `wc-navigation`

>My previous comment on that PR was if any package has @woocommerce/* in its name, it would also add wc-admin-app as a dependency.

This was my original intention with https://github.com/WordPress/gutenberg/pull/27020. I'd add an the option in the constructor of the class extending the original Dep Ext Webpack Plugin. 

>The above seems like a sane workaround that doesn't effect anyone.

I'll try and figure out a new way to make this happen

psealock:
It turns out this is a wc-admin dependencies issue. Here is the fix https://github.com/woocommerce/woocommerce-admin/pull/5722.

This one can be closed out. Thanks for the reviews everyone!

PR #5672, opened 2020-11-18
Disable Customer Effort Score for Mobile Users

adrianduffell:
This PR disables the Customer Effort Score for mobile users. It does this by preventing items being added to the queue if a mobile device was used.

- Delete the existing WP options state for Customer Effort Score, e.g.  in WP CLI 
`$ wp option delete woocommerce_ces_shown_for_actions`
- On a mobile device, update an existing product.
- Ensure no feedback notification appears.
- On a desktop computer, update an existing product.
- Ensure the feedback notification appears.
Screen Shot 2020-11-19 at 2 10 13 pm

### DIFF
--- a/tests/features/class-wc-tests-ces-tracks.php
+++ b/tests/features/class-wc-tests-ces-tracks.php
@@ -26,16 +26,15 @@ class WC_Tests_CES_Tracks extends WC_Unit_Test_Case {
 	public function setUp() {
 		parent::setUp();
 		update_option( 'woocommerce_allow_tracking', 'yes' );
-		$this->ces = new CustomerEffortScoreTracks();
 	}
 
 	/**
 	 * Verify that it adds correct action to the queue on woocommerce_update_options action.
 	 */
 	public function test_updating_options_triggers_ces() {
-		do_action( 'woocommerce_update_options' );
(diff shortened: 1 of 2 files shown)


tests/features/class-wc-tests-ces-tracks.php
```diff
@@ -26,16 +26,15 @@ class WC_Tests_CES_Tracks extends WC_Unit_Test_Case {
 	public function setUp() {
 		parent::setUp();
 		update_option( 'woocommerce_allow_tracking', 'yes' );
-		$this->ces = new CustomerEffortScoreTracks();
```
adrianduffell, line 29: I moved instantiation of `CustomerEffortScoreTracks` to the individual tests because the constructor has side-effects. 

When it was in `setUp`, a `woocommerce_update_options` action was added prior to every test. This interfered with the mobile test, which needs to check that mobile usage has no observable effects.

mattsherman (approved):
Nice, simple solution!

Tested and works well...
- CES showed up on desktop (Firefox)
- CES showed up on iPad Pro (Safari)
- CES did not show up on iPhone 11 Pro (Safari)

PR #5678, opened 2020-11-18
CES: Update tests to fake that they are in wp-admin

mattsherman:
In #5540, we changed things so that Customer Effort Score (CES) action hook handlers are only setup if running in wp-admin. Unfortunately, this causes our CES unit tests to fail (we didn't catch this initially because the tests were not running properly, see #5650, #5656).

Before:

```
There was 1 failure:

1) WC_Tests_CES_Tracks::test_updating_options_triggers_ces
Failed asserting that a boolean is not empty.

/srv/www/woocommerce-dev/public_html/wp-content/plugins/woocommerce-admin/tests/features/class-wc-tests-ces-tracks.php:38

FAILURES!
Tests: 1, Assertions: 1, Failures: 1.
```

This PR updates the CES unit tests to fake that we are running in wp-admin.

- Run CES unit tests:

```
composer test -- --filter="WC_Tests_CES_Tracks"
```

- Verify that test pass:

```
OK (1 test, 2 assertions)
```

### Changelog Note:

No changelog entry required, as it is part of the larger CES feature being introduced.

### DIFF
--- a/tests/features/class-wc-tests-ces-tracks.php
+++ b/tests/features/class-wc-tests-ces-tracks.php
@@ -7,6 +7,9 @@
 
 use \Automattic\WooCommerce\Admin\Features\CustomerEffortScoreTracks;
 
+// CustomerEffortScoreTracks only works in wp-admin, so let's fake it.
+define( 'WP_ADMIN', true );
+
 /**
  * Class WC_Tests_CES_Tracks
  */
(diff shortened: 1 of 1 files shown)


octaedro (approved):
Good job @mattsherman! The code looks good and is testing well here :shipit:

mattsherman:
Thanks for the review @octaedro !

PR #5682, opened 2020-11-18
Prevent adding duplicate queue item by checking action and label values

### LINKED ISSUE #5670: Doubled up CES survey notifications/snackbars
**Describe the bug**
It's possible to see doubled up notifications/snackbars under some conditions.

**To Reproduce**

1. delete the `woocommerce_ces_shown_for_actions` option (just to be sure)
1. add a new product
1. as soon as you get to the edit screen for the new product and before the notice displays, press "Add New" to navigate back to the add product screen
1. add another product
1. let the notices display on the edit screen

**Expected behavior**
Only one notice should appear for each action type.

The reason is that the survey is enqueued in the `woocommerce_ces_tracks_queue` option and isn't removed until the notices are displayed. I think this could be avoided by checking during enqueuing that the action doesn't already exist in the queue. It needs to be a queue because different surveys could be displayed on the same screen - they're not with the surveys we're creating but this may change as other teams want to use this.

**Screenshots**
image

moon0326:
Fixes #5670 

This PR fixes an edge case where duplicate snackbar notifications displayed as described in the original [issue](https://github.com/woocommerce/woocommerce-admin/issues/5670).

### Screenshots

Screen Shot 2020-11-18 at 2 41 37 PMth_

Make sure to delete the `woocommerce_ces_shown_for_actions` and the tracking option is turned on.

1. Click Products -> Add New
2. Enter the product name and click the [ Publish ] button. When you are redirected to the next page, close the tab immediately (timing is important) to prevent seeing the notification.
3. Navigate back to the admin area and click Products -> Add New to add a new product.
4. Make sure that you see only one snackbar notification.

### DIFF
--- a/src/Features/CustomerEffortScoreTracks.php
+++ b/src/Features/CustomerEffortScoreTracks.php
@@ -220,6 +220,17 @@ class CustomerEffortScoreTracks {
 			array()
 		);
 
+		$has_duplicate = array_filter(
+			$queue,
+			function ( $queue_item ) use ( $item ) {
+				return $queue_item['action'] === $item['action']
+				&& $queue_item['label'] === $item['label'];
+			}
+		);
+		if ( $has_duplicate ) {
+			return;
+		}
+
 		$queue[] = $item;
 
 		update_option(
--- a/src/Features/CustomerEffortScoreTracks.php
+++ b/src/Features/CustomerEffortScoreTracks.php
@@ -223,8 +223,7 @@ class CustomerEffortScoreTracks {
 		$has_duplicate = array_filter(
 			$queue,
 			function ( $queue_item ) use ( $item ) {
-				return $queue_item['action'] === $item['action']
-				&& $queue_item['label'] === $item['label'];
+				return $queue_item['action'] === $item['action'];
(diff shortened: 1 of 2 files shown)


src/Features/CustomerEffortScoreTracks.php
```diff
@@ -220,6 +220,17 @@ private function enqueue_to_ces_tracks( $item ) {
 			array()
 		);
 
+		$has_duplicate = array_filter(
+			$queue,
+			function ( $queue_item ) use ( $item ) {
+				return $queue_item['action'] === $item['action']
+				&& $queue_item['label'] === $item['label'];
```
becdetat, line 227: Just curious, why did you check the label as well as the action?

becdetat (approved):
Very nice!

src/Features/CustomerEffortScoreTracks.php
moon0326: We might not need to check the label. I was thinking of a situation where you want to display two notifications with a different label, but I might be overthinking :) We can remove the label.

Issue #5608, opened 2020-11-11
Navigation: SlotFill not rendered on embedded pages


psealock:
Extensions use the slotFill to write custom components to overwrite conventional links provided to the Nav on registration.

This appears to only happen on wc-admin pages, but not embedded pages. I'm wondering if we're missing that logic on embedded pages because its not loaded?

### Steps to Reproduce

1. Add and activate an extension that uses the slotFill, such as https://github.com/psealock/examples.
2. Go to a wc-admin page and see the slotFill working in the Nav under Extensions > Examples

<img width="221" alt="Screen Shot 2020-11-12 at 10 34 07 AM" src="https://user-images.githubusercontent.com/1922453/98866957-a4b52880-24d2-11eb-9854-40770b2cb237.png">

3. Go to an embedded page and visit the same Nav menu. Note the lack of emoji, which is added by slotFill

<img width="234" alt="Screen Shot 2020-11-12 at 10 35 42 AM" src="https://user-images.githubusercontent.com/1922453/98867097-dfb75c00-24d2-11eb-861d-34ae7ddc4300.png">

Relavant extension code: https://github.com/psealock/examples/blob/master/src/index.js

PR #5686, opened 2020-11-19
Starter Pack: expand scope to all admin pages

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/5608

SlotFill wasn't working with my example because the script was never loaded outside wc-admin pages, but should be loaded on all admin pages. This PR changes that to prevent future headache for extensions using the starter pack to hook into the Nav.

1. Just make sure the change makes sense. Any extension using the starter pack will have their scripts load on all admin pages, not just wc-admin. This is important because the new Nav operates outside of just wc-admin pages.

### Changelog Note:

- tweak: Expand default script loading in Create Extension to all admin pages

### DIFF
--- a/bin/starter-pack/_main.php
+++ b/bin/starter-pack/_main.php
@@ -9,11 +9,6 @@
  * Register the JS.
  */
 function add_extension_register_script() {
-
-	if ( ! class_exists( 'Automattic\WooCommerce\Admin\Loader' ) || ! \Automattic\WooCommerce\Admin\Loader::is_admin_page() ) {
-		return;
-	}
-
 	$script_path       = '/build/index.js';
 	$script_asset_path = dirname( __FILE__ ) . '/build/index.asset.php';
 	$script_asset      = file_exists( $script_asset_path )
--- a/bin/starter-pack/_main.php
+++ b/bin/starter-pack/_main.php
@@ -9,6 +9,10 @@
  * Register the JS.
  */
 function add_extension_register_script() {
+	if ( ! is_admin() ) {
+		return;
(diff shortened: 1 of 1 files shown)


bin/starter-pack/_main.php
```diff
@@ -9,11 +9,10 @@
  * Register the JS.
  */
 function add_extension_register_script() {
-
-	if ( ! class_exists( 'Automattic\WooCommerce\Admin\Loader' ) || ! \Automattic\WooCommerce\Admin\Loader::is_admin_page() ) {
+	if ( ! is_admin() ) {
```
joshuatf, line 12: Not completely opposed to opening this up for all pages, but could we use `is_admin_or_embed_page()` to allow the nav slot fill to work where needed?

joshuatf (commented):
Good call 👍 

Quick question on a possible way to limit loading on every page.

bin/starter-pack/_main.php
psealock: oh yeah good idea. Can we be sure `is_admin_or_embed_page` is available to extensions? It isn't in my initial testing.

joshuatf: If they register a page with the new navigation, they will qualify under `is_admin_or_embed_page()`.  Do you have an example extension that doesn't pass this test?  If so, we can keep `is_admin()`.

psealock: Gotcha, thanks. I updated the code.

adrianduffell (approved):
This code change makes sense 🚀

PR #5696, opened 2020-11-19
Fix snackbar dismissal bug

jeffstieler:
This PR seeks to restore the dismissal functionality of notices that are in the `core/notices` store. When the component was [copied from Gutenberg](https://github.com/woocommerce/woocommerce-admin/commit/8e0a6cf8f1ea850fd4f324e82197d2825dab5687), only the `core/notices2` store was [having dismissal actions dispatched](https://github.com/woocommerce/woocommerce-admin/blob/a1237ed49636c751fedac1d4574ea49b2d6019ab/client/layout/transient-notices/index.js#L58).

- Trigger a snackbar notice - I used the Stock panel (update quantity)
- Verify the snackbar appears and auto dismisses after 10 seconds
- Trigger another
- Click the snackbar before 10 seconds elapses
- Verify it is dismissed

### Changelog Note:

N/A - unreleased bug

### DIFF
--- a/client/layout/transient-notices/index.js
+++ b/client/layout/transient-notices/index.js
@@ -15,7 +15,7 @@ import './style.scss';
 
 class TransientNotices extends Component {
 	render() {
-		const { className, notices, onRemove } = this.props;
+		const { className, notices, onRemove, onRemove2 } = this.props;
 		const classes = classnames(
 			'woocommerce-transient-notices',
 			'components-notices__snackbar',
@@ -27,6 +27,7 @@ class TransientNotices extends Component {
 				notices={ notices }
 				className={ classes }
 				onRemove={ onRemove }
+				onRemove2={ onRemove2 }
 			/>
 		);
 	}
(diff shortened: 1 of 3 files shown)


client/layout/transient-notices/snackbar/list.js
```diff
@@ -40,7 +47,11 @@ function SnackbarList( { notices, className, children, onRemove = noop } ) {
 	} );
 
 	className = classnames( 'components-snackbar-list', className );
-	const removeNotice = ( notice ) => () => onRemove( notice.id );
+	const removeNotice = ( notice ) => () => {
+		onRemove( notice.id );
+		// To be removed when we're no longer using core/notices2.
+		onRemove2( notice.id );
```
jeffstieler, line 53: OMG HAX

client/layout/transient-notices/index.js
```diff
@@ -27,6 +27,7 @@ class TransientNotices extends Component {
 				notices={ notices }
 				className={ classes }
 				onRemove={ onRemove }
+				onRemove2={ onRemove2 }
```
mattsherman, line 30: I pushed a commit to add this -- otherwise, the new notices that we are using in CES do not get removed. :-)

mattsherman (approved):
With the commit I pushed, this works well for me.

I verified that both existing notices and new CES notices dismiss properly.

Thanks for jumping on this!

client/layout/transient-notices/index.js
jeffstieler: Oh whoops! I must have accidentally removed that when cleaning up test code. Thanks!

PR #5697, opened 2020-11-19
Allow navigation feature in plugin and core builds

### LINKED ISSUE #5692: Navigation toggle is visible but feature cannot be enabled
**Describe the bug**
The toggle is available in the features settings, but toggling has no effect since the feature is not included in non-dev builds.

**To Reproduce**

1. Build a non-dev version of WCA.
1. Navigate to `wp-admin/admin.php?page=wc-settings&tab=advanced&section=features`
1. Turn on the feature.
1. Note that the new feature does not work.

**Expected behavior**
Expected either the feature toggle to not exist or to allow toggling in any build.

**Screenshots**
Screen Shot 2020-11-19 at 12 54 49 PM

joshuatf:
Fixes #5692 

Allows the navigation feature to be toggled on/off in all builds.  By default, this feature is still off and behind a feature flag.

### Screenshots
Screen Shot 2020-11-19 at 12 54 49 PM

1. Build a non-dev version of WCA.
1. Navigate to `wp-admin/admin.php?page=wc-settings&tab=advanced&section=features`
1. Turn on the feature.
1. Note that the new nav feature loads (you may need to refresh the page once).

### DIFF
--- a/config/core.json
+++ b/config/core.json
@@ -10,7 +10,7 @@
 		"marketing": true,
 		"minified-js": false,
 		"mobile-app-banner": true,
-		"navigation": false,
+		"navigation": true,
 		"onboarding": true,
 		"remote-inbox-notifications": true,
 		"shipping-label-banner": true,
--- a/config/plugin.json
+++ b/config/plugin.json
@@ -10,7 +10,7 @@
 		"marketing": true,
 		"minified-js": true,
 		"mobile-app-banner": true,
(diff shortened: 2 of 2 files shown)


psealock (approved):
Seems like a good time to make this change. I don't think it should be cherry-picked unless we also cherry pick all the issues we've merged since 1.7.0

PR #5700, opened 2020-11-19
CES: Fix dismissal button (X) style in notice/snackbar

mattsherman:
This PR fixes the dismissal button (X) style in the CES notice/snackbar.

### Screenshots

Before:

Screen Shot 2020-11-19 at 15 41 36

After:

Screen Shot 2020-11-19 at 15 47 01

### Detailed test instructions:

- Make sure tracking is enabled: `/wp-admin/admin.php?page=wc-settings&tab=advanced&section=woocommerce_com`
- Clear out the options tracking whether the survey has been shown:

```
/wp-admin/admin.php?page=wc-settings&tab=advanced&section=woocommerce_com
```

- Add or edit a product in wp-admin.
- When the snackbar/notice is shown, verify that the dismissal button is styled correctly (left margin and cursor).

### Changelog Note:

No changelog entry needed.

PR #5703, opened 2020-11-20
Always call onRemove when action is clicked in Snackbar

mattsherman:
Whenever a Snackbar action is clicked, the snackbar should be removed.

- Trigger snackbars with actions, such as removing an Inbox notice (click the "Undo" action in the snackbar).
- Trigger CES snackbars with action, such as editing a product (see #5700 for testing instructions on how to get a CES snackbar to appear).
- Verify that clicking a Snackbar action removes the Snackbar.

### Changelog Note:

No changelog entry needed.


psealock (approved):
Works well and code looks good.

Issue #5239, opened 2020-10-01
Home Screen - migrate Stock panel


pmcpinto:
**User story:**
As a merchant
I want to easily see how many products are low stock or out of stock when I enter WooCommerce
So I can restock inventory as soon as possible

**Description:** we’re going to migrate the stock panel features to the home screen, as managing inventory is one of the most important tasks for stores with Stock Management enabled

**Acceptance criteria:**
- [ ] Display a Stock task in the home screen in stores with [Stock Management](https://docs.woocommerce.com/document/managing-products/#inventory-tab) enabled
- [ ] Display up to 5 products that are either Low in Stock or Out of Stock
![Home (1).png](https://images.zenhubusercontent.com/5d84fe8e1201de0001b6ab61/0d5fd5eb-8b51-4388-b389-830cfe299a10)
- [ ] The product card should include the same elements that we have now: image, product name, current stock and `Update Stock` CTA. Ideally we should also add the last time that a product was ordered (nice to have)
- [ ] The user is redirected to the product page when clicking in the product name
- [ ] Clicking in the `Update Stock` CTA opens a modal allowing the user to update the product stock without leaving the home screen
![Home (2).png](https://images.zenhubusercontent.com/5d84fe8e1201de0001b6ab61/65163851-7fce-4253-9528-ecbfe4583637) 
- [ ] The stock activity panel should be removed from the navbar once we have the same feature in the home screen

**Event tracking:**
Can we reuse the same event tracking that we have in the stock panel?

**Additional links:**
[P2 with links to the Figma file](pbIJXs-ma-p2)

jeffstieler:
Noting for my own sake that this is the most recent design of the card title:

![Uploading home-all-agenda-items.png…]()

jeffstieler:
> The stock activity panel should be removed from the navbar once we have the same feature in the home screen

@pmcpinto @elizaan36 does that mean the stock panel will only be available on the homescreen? I see that Orders has been removed, but Inbox remains.

pmcpinto:
> @pmcpinto @elizaan36 does that mean the stock panel will only be available on the homescreen? I see that Orders has been removed, but Inbox remains.
Yes, that's correct. @elizaan36 we should consider removing the inbox as well.

> Also - should the Stock panel be hidden when there are no low/out stock products?

I think the idea is to have a similar empty state that we also have for the orders panel, when the user has inventory management enabled. If it's disabled the panel should be hidden.

[...]

PR #5706, opened 2020-11-20
Migrate reviews panel to home screen

### LINKED ISSUE #5240: Home Screen - migrate Reviews panel
**User story:**
As a merchant
I want to easily see how many new reviews my store has when I enter WooCommerce
So I can review them as soon as possible

**Description:** we’re going to migrate the reviews panel features to the home screen, as managing reviews can be an important task for stores with Product Reviews enabled

**Acceptance criteria:**
- [ ] Display a Product Reviews task in the home screen in stores with [Product Reviews](https://docs.woocommerce.com/document/configuring-woocommerce-settings/#section-9) enabled
- [ ] Display up to 5 reviews pending moderation

- [ ] The review card should include the following elements: image, product name, customer name, rating, comment, time and the `approve`, `mark as spam` and `delete` actions
- [ ] The user is redirected to the product page when clicking in the product name and redirected to the Customer report filtered by the customer name if the user clicks in the customer name
- [ ] Include a `Manage all reviews` CTA at the bottom of the task that redirects the user to the reviews list
- [ ] The review activity panel should be removed from the navbar once we have the same feature in the home screen

**Event tracking:**
Can we reuse the same event tracking that we have in the reviews panel?
We also need to add on to cover the moderate actions:
`wcadmin_reviews_action`
Event prop: `action`- `approve` |`mark as spam` | `delete`

**Additional links:**

louwie17:
Fixes #5240 

The bulk of the code is migrating the reviews panel to the home screen. Currently there is a decent amount of overlapping logic between the reviews header panel and the home screen panel.
The home screen panel contains several actions (approve, set as spam, delete), these actions had to be added to the reviews data store - I added the updateReview and deleteReview actions.
Made use of `invalidateResolutionForStoreSelector` to re-trigger updates after update or deletion.

Updated the rating component to allow for custom icons to be passed in (as strings), this way I could show both the normal star, and the outline star as shown in the wireframe.

### Accessibility

### Screenshots

review-panel

- Create new store and add several products, make sure you have the [products reviews](https://docs.woocommerce.com/document/configuring-woocommerce-settings/#section-9) enabled.
- Go to the public product page and add some comments (in incognito so the comments show up as 'pending')
- Go to the Woo home screen
- You should see a `Reviews to moderate` on the left, with the number of pending reviews (Reviews header panel should not be shown - awaiting feedback if just on home screen)
- You should be able to Approve, mark as spam, or delete a review, a popup will appear that allows you to 'undo' this action
- If no reviews are needing review it should not be allowed to collapse.

UPDATE:
There are a couple small styling changes that depend on Jeff's updates: https://github.com/woocommerce/woocommerce-admin/pull/5729/files#diff-beff577e21ba5ae2d55e05bdfcbcc94c7da812eb81f26b5da415b870d79d2c82
I believe the only thing that is noticeable is the margin between the icon and the subtitle.


client/header/activity-panel/activity-card/style.scss
```diff
@@ -262,61 +261,6 @@
 	}
 }
 
-.woocommerce-review-activity-card {
```
louwie17, line 265: Moved some of this to the homescreen reviews style.scss and deleted the rest as the header review panel is removed.

client/header/activity-panel/index.js
```diff
@@ -21,11 +21,7 @@ import { getHistory, getNewPath } from '@woocommerce/navigation';
  */
 import './style.scss';
 import ActivityPanelToggleBubble from './toggle-bubble';
```
louwie17, line 23: Removing any use of the ReviewsPanel

client/homescreen/activity-panel/reviews/index.js
```diff
@@ -0,0 +1,432 @@
...
+	renderReview( review ) {
+		const product =
+			( review &&
+				review._embedded &&
+				review._embedded.up &&
+				review._embedded.up[ 0 ] ) ||
+			null;
+
+		if ( review.isUpdating || this.props.isRequesting ) {
+			return (
```
louwie17, line 160: Showing loading placeholder when an review is updating or when it is updating the entire list.

client/homescreen/activity-panel/reviews/index.js
```diff
@@ -0,0 +1,432 @@
...
+			updateReview,
+			invalidateResolutionForStoreSelector,
+		} = dispatch( REVIEWS_STORE_NAME );
+		const { createNotice } = dispatch( 'core/notices' );
+
+		const clearReviewsCache = ( emptyReviews ) => {
+			invalidateResolutionForStoreSelector( 'getReviews' );
+			if ( emptyReviews ) {
+				invalidateResolutionForStoreSelector( 'getReviewsTotalCount' );
+			}
```
louwie17, line 396: Clearing the `getReviewsTotalCount` if no reviews left, only necessary in the case we revert the last review. Without this it wouldn't update the home screen.

client/homescreen/activity-panel/style.scss
```diff
@@ -1,30 +1,47 @@
...
+		text-align: left;
+		.woocommerce-accordion-title {
+			vertical-align: bottom;
+			margin-right: $gap;
+		}
+	}
+}
 .woocommerce-accordion-card {
+	.components-card__header,
+	.components-panel__body-title .components-button {
```
louwie17, line 13: This is overlapping css between the components button and the header.
css for the empty card-header.
Although maybe this css should be moved to the accordion component?

client/homescreen/activity-panel/index.js
```diff
@@ -43,6 +49,7 @@ export const ActivityPanel = () => {
 							count={ count }
 							initialOpen={ initialOpen }
 							title={ title }
+							collapsible={ count !== 0 }
```
louwie17, line 52: Set collapsible to `false` if the count is 0

packages/components/src/rating/index.js
```diff
@@ -72,6 +73,14 @@ Rating.propTypes = {
...
 	className: PropTypes.string,
+	/**
+	 * Icon used, defaults to StarIcon
+	 */
+	icon: PropTypes.elementType,
+	/**
+	 * Outline icon used, the not selected rating. Defaults to props.icon or StarIcon
+	 */
+	outlineIcon: PropTypes.elementType,
 };
```
louwie17, line 84: This allows me to use the filled star and the outline star:
Screen Shot 2020-11-23 at 4 24 18 PM

client/homescreen/activity-panel/panels.js
```diff
@@ -27,6 +30,18 @@ export function getAllPanels( {
...
+			className: 'woocommerce-homescreen-card',
+			id: 'reviews-panel',
+			count: countUnapprovedReviews,
+			initialOpen: false,
+			panel: (
+				<ReviewsPanel
+					hasUnapprovedReviews={ countUnapprovedReviews > 0 }
+				/>
+			),
+			title: __( 'Reviews to moderate', 'woocommerce-admin' ),
```
samueljseay, line 43: It wouldn't have been obvious because I think it was buried in another issue discussion but the copy for this should now just be "Reviews"

client/homescreen/activity-panel/style.scss
samueljseay: It could make sense to,but I'll leave it up to you, if you wanna do that as a follow up later.

packages/components/src/accordion/test/index.js
```diff
@@ -50,4 +50,22 @@ describe( 'Accordion', () => {
 		expect( screen.queryByText( '10000' ) ).toBeInTheDocument();
 		expect( screen.queryByText( '20000' ) ).toBeInTheDocument();
 	} );
+
```
samueljseay, line 53: Looking over the tests you've added they look great. Good stuff!

samueljseay:
Reviewing the look and feel it looks like some of the spacings and sizings are off from the new designs.

For example:

* In the new design the product image is 33x33px
* In the new design the right side margin of the product image is 16px, looks to be about 20 in yours
* Even with a short product name/description the stars on yours will run on to the next line. I wonder if we should truncate the text and make sure the stars always display on the same line @elizaan36 any input on how we'd handle a long line of product / reviewer name?
* In the new design the time since review was posted is no longer all caps.

There are a few other things, I'm also considering maybe it just wasn't clear that there was some redesign work to do here, so sorry about any ambiguity there. You can check out the designs (figma links at bottom) [here](pbIJXs-ma-p2-p2).

samueljseay (dismissed):
Overall everything looks great, I think the reviews themselves need some redesign applied and there is a piece of copy to change, but otherwise absolutely stellar work. ✨

packages/components/src/accordion/test/index.js
louwie17: > There are a few other things, I'm also considering maybe it just wasn't clear that there was some redesign work to do here, so sorry about any ambiguity there. You can check out the designs (figma links at bottom) [here](pbIJXs-ma-p2-p2).

Thanks I did look at this before and I probably should off asked how to get the specific pixel amounts and such. I just noticed I never saw them as I didn't create an account for Figma.

client/homescreen/activity-panel/style.scss
louwie17: Yeah I suppose the same goes for the activity card, we will probably have to consolidate the css once @jeffstieler stock migration is finished as well, given the Activity Card will largely be used on the home screen instead of the header. It looks like the Inbox would be the only one still using it.

louwie17:
Updated styling changes mentioned by @samueljseay in https://github.com/woocommerce/woocommerce-admin/pull/5706/commits/bbd5e69f38b9132daa080986da8c175b9140483b
These changes did leave some more room between the title, rating and the date, awaiting feedback incase of truncation of the title.

client/homescreen/activity-panel/reviews/index.js
jeffstieler: This should be unnecessary after https://github.com/woocommerce/woocommerce-admin/pull/5729/commits/9d7626d558a5b0b2a1889d73bea039b4a78df6fe lands.

packages/data/src/reviews/actions.js
```diff
@@ -12,6 +19,60 @@ export function updateReviews( query, reviews, totalCount ) {
 	};
 }
 
+export function* updateReview( reviewId, reviewFields, query ) {
```
jeffstieler, line 22: What do you think about creating some status-specific actions like `markReviewAsSpam()` and `approveReview()` so that the component code can be a little simpler?

client/homescreen/activity-panel/reviews/index.js
```diff
@@ -0,0 +1,436 @@
...
+											reviewId,
+											{
+												status: 'untrash',
+											},
+											{
+												_embed: 1,
+											}
+										).then( () =>
+											clearReviewsCache(
+												reviews.length < 2
```
jeffstieler, line 77: Why is this `2`? Doesn't the panel show `5` reviews?

client/homescreen/activity-panel/reviews/index.js
louwie17: @jeffstieler nice, would this require me to use `getItemsTotalCount` instead of `getReviewsTotalCount`?

client/homescreen/activity-panel/reviews/index.js
louwie17: Yeah it shows 5, this was in reference to this change: https://github.com/woocommerce/woocommerce-admin/pull/5706#discussion_r528972306 to only call it if it was the last item being updated.
I could probably check if `reviews.length` would actually be zero (I didn't specifically test that, as I thought it might show as `1` still within the scope).

client/homescreen/activity-panel/style.scss
elizaan36: in case this factors in, just adding a note here that we're going to remove the Inbox from the app header altogether so it will only appear on the Home screen.

elizaan36:
Just chatted with @louwie17 on Slack a bit and we're going to move the star rating to the next line below the customer and product name. In the case of a long product or customer name, the line will wrap. @jeffstieler already has this implemented in his PR so maybe this is another thing to consolidate.

Here's the design with star rating on a new line.
Screen Shot 2020-11-24 at 12 39 14 PM

I also noticed some minor style tweaks that we talked about on Slack, mainly to do with the card header. But I'll wait until one or the other panels is merged to have a look at those styles again.

client/homescreen/activity-panel/reviews/index.js
jeffstieler: I don't think so as their implementations are identical and the total update happens in the resolver.

packages/data/src/reviews/actions.js
louwie17: Yeah not a bad idea, the only caveat was 'reverting' a change, which made having a generic `update` function rather handy. I could do both. In either case would it make sense to move the `createNotice` to the action as well?

jeffstieler: Ah right.. the previous status won't always be the same making reversal unable to take advantage of the "simpler" actions. Let's scrap the idea 🙃 .

> In either case would it make sense to move the `createNotice` to the action as well?

No I don't think we'd want to couple those in the event that UX needed to change or the action could be used in multiple different contexts.

packages/components/src/accordion/test/index.js
samueljseay: @louwie17 did you get sorted out with Figma access? There are a few tools in there for measuring distance between objects in the designs, let me know if you need help with any of that.

louwie17: @samueljseay I did figure that out, although a little overview of the tools and how to's would certainly be helpful. We can do that during our 1:1 to tomorrow?

louwie17:
Ready for review again, fixed merge conflicts with https://github.com/woocommerce/woocommerce-admin/pull/5729 PR, @elizaan36 I believe all the styling changes reflect the wireframe now and the rating is moved down.

@jeffstieler I noticed that non-refresh issue again, after undoing the last item. Turns out it has to do with this change I did - only showing the header: https://github.com/woocommerce/woocommerce-admin/pull/5706/commits/ebd963fc48d07dfa5afccaea2b43ef70a3f7ff02 as it won't render the actual panel in that case and not re-trigger the selector.
I was going to use the same fix I did on yours, but noticed that the `invalidateResolution` calls actually don't do anything (you can see that there are no changes in the screenshot):
Screen Shot 2020-11-25 at 5 26 08 PM

I didn't want to call `invalidateResolutionForStoreSelector( 'getItems' );` as that would clear out `orders` as well. Any thoughts? Should we create specific selectors for the types -> `getOrders`, `getProducts`, etc...?
This can probably be done in a separate PR.

jeffstieler:
@louwie17 it looks like the `collapsible` change is causing some weird behavior with the stock panel as well:

Capture 2020-11-25 at 16 49 01

Here's what it looks like for both the stock and review panels without setting the `collapsible` prop:

Capture 2020-11-25 at 16 51 38

louwie17:
> @louwie17 it looks like the `collapsible` change is causing some weird behavior with the stock panel as well:

@jeffstieler do you mean the drop down arrow disappearing? Once there are zero items it just renders a header as displayed in the first picture here: https://github.com/woocommerce/woocommerce-admin/issues/5239#issuecomment-729897323
Which is what the `collapsible` thing is for, although I can certainly scratch that (or just hide the collapse with css).

client/homescreen/activity-panel/reviews/index.js
```diff
@@ -72,11 +71,7 @@ class ReviewsPanel extends Component {
 											{
 												_embed: 1,
 											}
-										).then( () =>
-											clearReviewsCache(
-												reviews.length < 2
-											)
-										);
+										).then( () => clearReviewsCache() );
```
jeffstieler, line 81: Do these have to be closures? Can we just pass `clearReviewsCache`?

client/homescreen/activity-panel/stock/index.js
```diff
@@ -40,13 +41,24 @@ export class StockPanel extends Component {
...
 		if ( result.success ) {
 			// Request more low stock products.
 			invalidateResolution( 'getItems', [ 'products', productsQuery ] );
+			if ( products.length < 2 ) {
+				invalidateResolution( 'getItemsTotalCount', [
+					'products',
+					getLowStockCountQuery,
+					null,
+				] );
+			}
```
jeffstieler, line 61: Why is the Stock panel being modified in this PR? If there's an issue with the Stock panel, can you open a separate PR?

jeffstieler (approved):
This looks good, just one minor callback nitpick. However, I do think the Stock panel changes should be removed from this PR and moved to a new one.

client/homescreen/activity-panel/stock/index.js
louwie17: Reverted the last commit and took out [the above changes](https://github.com/woocommerce/woocommerce-admin/pull/5706/commits/1a01312b54acbc0db9eab2fb4abb87fcbe469d96), these are no part of a different branch/PR -> https://github.com/woocommerce/woocommerce-admin/commit/16fb47b8b601c938ee58d32eb975f5b0a9265c45

jeffstieler:
I spot checked this after the removal of the Stock panel changes - looks good! :shipit:

PR #5707, opened 2020-11-20
Add navigation item click tracks

### LINKED ISSUE #5695: Navigation: implement event tracking on menu item clicks
It could be useful to track menu item clicks as a complement to the pageviews tracking and to understand how many times the user is visiting the wp-admin pages, etc.

Event name: wcadmin_new_navigation_click
Event prop: menu_item: name

I assume that we're already a bit short of time to include this in the eComm release, so it's ok if we only implement it for the next stage.

joshuatf:
Fixes #5695 

Adds tracking to navigation items, categories, and back buttons.

### Screenshots
Screen Shot 2020-11-20 at 12 29 07 PM

1. Enable the new navigation.
1. Enter `localStorage.setItem( 'debug', 'wc-admin:*' );` into your console. Leave your console open.
1. Click on both slot filled and not slot filled links.
1. Make sure the clicks are tracked.
1. Install and activate the latest dev branch of GB.
1. Check that the back button clicks are tracked.

### DIFF
--- a/client/navigation/components/Item/index.js
+++ b/client/navigation/components/Item/index.js
@@ -2,17 +2,30 @@
  * External dependencies
  */
 import { __experimentalNavigationItem as NavigationItem } from '@wordpress/components';
+import { recordEvent } from '@woocommerce/tracks';
 import { WooNavigationItem, useNavSlot } from '@woocommerce/navigation';
 
 const Item = ( { item } ) => {
 	const slot = useNavSlot( 'woocommerce_navigation_' + item.id );
 	const hasFills = Boolean( slot.fills && slot.fills.length );
 
+	const trackClick = ( id ) => {
+		recordEvent( 'navigation_click', {
+			menu_item: id,
+		} );
+	};
+
+	// Disable reason: The div wrapping the slot item is used for tracking purposes
+	// and should not be a tabbable element.
+	/* eslint-disable jsx-a11y/no-static-element-interactions, jsx-a11y/click-events-have-key-events */
+
 	// Only render a slot if a coresponding Fill exists and the item is not a category
 	if ( hasFills && ! item.isCategory ) {
 		return (
 			<NavigationItem key={ item.id } item={ item.id }>
-				<WooNavigationItem.Slot name={ item.id } />
+				<div onClick={ () => trackClick( item.id ) }>
+					<WooNavigationItem.Slot name={ item.id } />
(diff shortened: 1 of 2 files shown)


adrianduffell (commented):
@joshuatf overall this looks good 👍 

One issue: all of the nav items I tried had events recorded except for the back button. I tried it for both the WordPress Dashboard and WooCommerce Home. I built GB from the repo, used Chrome and also had "Preserve log" enabled in my console settings.

joshuatf:
Thanks for taking a look @adrianduffell!  I'm thinking the built version of GB may not be overriding the WCA bundled version (I think I may have `npm link`'d mine previously).

Can you bump the version of `@wordpress/components` to `11.1.1` in `package.json` and try again?

I don't think we'll handle the other issues around that version bump yet since @psealock is looking at removing the components dependency in cooldown, but it will at least allow testing of the back button tracks for the time being.

adrianduffell (approved):
> Can you bump the version of @wordpress/components to 11.1.1 in package.json and try again?

This worked 🚀

client/navigation/components/Item/index.js
```diff
@@ -2,17 +2,30 @@
...
+	// Disable reason: The div wrapping the slot item is used for tracking purposes
+	// and should not be a tabbable element.
+	/* eslint-disable jsx-a11y/no-static-element-interactions, jsx-a11y/click-events-have-key-events */
+
 	// Only render a slot if a coresponding Fill exists and the item is not a category
 	if ( hasFills && ! item.isCategory ) {
 		return (
 			<NavigationItem key={ item.id } item={ item.id }>
-				<WooNavigationItem.Slot name={ item.id } />
+				<div onClick={ () => trackClick( item.id ) }>
```
psealock, line 26: Does this `onClick` not work directly on `NavigationItem`?

joshuatf: @psealock It currently does not; see https://github.com/WordPress/gutenberg/blob/47feef0a2aaaa38b34fb3525a35cfddaf032c719/packages/components/src/navigation/item/index.js#L73

The children replaces the `ItemUI` which is what houses the onclick props.  We could potentially open a PR in GB to try and pass these props to `children` as well, but I think this would still require the child element to accept and use the `onClick` prop in its component.

psealock (approved):
LGTM 🚢

PR #5709, opened 2020-11-20
Style updates to tax task

elizaan36:
Fixes #

- Tweaked some styles and copy to make the tax task easier to read and understand
- Also updated the hover/focus/active state for the footer link in the Orders panel to be consistent with other button states on the home screen

### Screenshots
Screen Shot 2020-11-20 at 2 48 33 PM

image

- Open Set up tax step from Task list on Home screen
- See visual and copy updates

- Also, hover over "Manage all orders" link in Orders panel to see updates there

Note that I tried to change the buttons to tertiary buttons but not sure if that worked - feel free to comment and let me know anything that's off base.

### DIFF
--- a/client/homescreen/activity-panel/style.scss
+++ b/client/homescreen/activity-panel/style.scss
@@ -64,17 +64,17 @@
 		}
 
 		&:hover {
-			background: $gray-200;
+			background-color: $gray-100;
 		}
 
 		&:active {
-			background: $gray-100;
+			background-color: $gray-100;
 		}
 
 		&:focus {
-			background: $gray-200;
+			background-color: $gray-100;
 			box-shadow: inset 0 0 0 1px $box-shadow-blue,
-				inset 0 0 0 2px $gray-200;
+				inset 0 0 0 2px $gray-100;
 		}
 
 		&:hover,
--- a/client/task-list/style.scss
+++ b/client/task-list/style.scss
@@ -351,7 +351,7 @@
 
 .woocommerce-task-dashboard__container {
 	button.components-button.is-primary {
-		margin: 0;
+		margin: 0 8px 0 0;
 	}
 
 	button.components-button.is-link {
--- a/client/task-list/style.scss
+++ b/client/task-list/style.scss
@@ -496,5 +496,5 @@
 
 .woocommerce-task__caption {
 	color: $gray-700;
-	margin-top: 12px;
+	margin-top: $gap;
 }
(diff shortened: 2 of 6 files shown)


jeffstieler:
My build was erroring out on the undefined `$gap-medium` variable. I changed that to `$gap` for now ([16px, defined here](https://github.com/woocommerce/woocommerce-admin/blob/2c8b17136fdd22e3c45806d7cd8c5a1c15fc2d87/client%2Fstylesheets%2Fabstracts%2F_variables.scss#L22)).

> Note that I tried to change the buttons to tertiary buttons but not sure if that worked - feel free to comment and let me know anything that's off base.

Found the other spot and added that. Please check on the styles now, they might need more tweaking. 😄 

Screen Shot 2020-11-20 at 5 06 08 PM

elizaan36:
Thanks @jeffstieler 
I added a right margin to the primary button and I think this step is looking good now. There could be another margin between the two tertiary buttons but we'd have to create another class for those and I think visually the spacing looks fine.

updated tax step with tertiary buttons

There's only one issue I'm seeing, and I'm not sure why this is happening but sometimes the `woocommerce-task_caption` class doesn't recognize the `$gap` variable. @jameskoster wanted to ask you about this, since there's an extra `.css-1qmnemh-Text` class on there. Is that something to do with the component in Gutenberg, and do you think that might get in the way of the margin on the task_caption class? [You can see the styles in the inspector in this screenshot.](https://share.getcloudapp.com/jkuY7o15)

jameskoster:
> Is that something to do with the component in Gutenberg

I believe that comes from the `Text` component in core which seems to be adding a margin. I'm not sure _why_ that component adds a margin, but it should be safe to override it with a more specific css selector.

elizaan36:
Thanks Jay! I just added one more commit to update another tertiary button in the Customize step.

@jeffstieler if you have a moment, let me know if this still looks good!

jeffstieler:
@elizaan36 looked good to me save for some padding - I tweaked a selector to make it more specific so we win out over Gutenberg.

I'll fix the conflict and then we can 🚢  it!

jeffstieler:
Ok - conflict fixed! Do you mind spot checking this once more @elizaan36? I'll merge afterwards. Ty!

elizaan36:
ah, nice catch. Thanks, and looks good to me!

jeffstieler (approved):

PR #5713, opened 2020-11-22
Install puppeteer just once

samueljseay:
Depending on puppeteer 3 times within our code base was noticeably painful, due to the post install for Puppeteer installing chromium, which takes *ages*! This was slowing down builds a lot and also slowing down anyone doing a fresh install.

This is 2 changes:

1. Remove dependency on puppeteer. We don't need it since its already included by woocommerce/env
2. Run `npm dedupe` to also move the puppeteer dependency higher up the dep tree because woocommerce/env *and* woocommerce/puppeteer-utils both depend on it. This final change reduces puppeteer installs to 1. This change only impacts the lock file.

### Testing instructions

1. ensure your `node_modules` folder is deleted. (`rm -fr node_modules`)
2. run `npm i`
3. Confirm that "downloading Chromium" step happens just once.

### Changelog note

dev: Don't depend on puppeteer directly.


adrianduffell (approved):
Thanks for sorting this out. It tested well for me 🚀

samueljseay:
Just updated this with master changes. I'm not sure how maintainable this approach of dedupe will be but at least this does remove the extra dependency on puppeteer. 

I believe this should be fixed properly after the next release of wc core

jeffstieler:
@samueljseay could we have specified puppeteer as a peer dependency?

samueljseay:
> @samueljseay could we have specified puppeteer as a peer dependency?

No, you declare a peer dependency when you want consumers of your package to install the dependency themselves. 

Woo cores package should have puppeteer as a peer dependency instead so that would force us to have it as a dependency but only once have it installed.

PR #5715, opened 2020-11-23
Remove unused dependencies, convert buildScss to async

samueljseay:
This PR cleans up dependencies where possible to try achieve 2 goals:

1. Speed up fresh npm installs which we all likely do a lot while testing branches
2. Reduce maintenance burden, because every dependency we have is a potential dependency we must maintain.

I've done the following:

1. Remove `deasync` and adjusted the build code to use async where possible to make it work as before. (I recommend testing this by running the build and doing a smoke test of the app screens).

2. I've removed dependencies we aren't using. Some are not clear why they were there (such as deep-freeze). And others I think are leftover from the old devdocs feature which is now gone.

### Changelog Note:

Dev: Reduce npm dependencies, and use async to support building scss files in npm scripts.


bin/packages/build.js
```diff
@@ -214,14 +206,14 @@ function buildPackage( packagePath ) {
...
 
+	// Build package CSS files
+	await buildPackageScss( packagePath );
+
 	process.stdout.write( `${ path.basename( packagePath ) }\n` );
 
 	// Build js files individually.
 	jsFiles.forEach( ( file ) => buildJsFile( file, true ) );
 
-	// Build package CSS files
```
samueljseay, line 222: This is reordered to ensure the CLI logging looks/behaves the same as it did before.

adrianduffell (approved):
Wonderful @samueljseay I'm looking forward to this :thumbsup: 🚀 

I ran a smoke test and did not find any problems. Also confirmed the `build-style` artifacts are produced as expected.

How did you determine the unused dependencies?

samueljseay:
@adrianduffell I used a node package called `depcheck` to get started, but it gives a lot of false warnings (mainly babel stuff), and then from there it was a matter of global search in the code base to confirm if something was not used.

If I was not sure about something I just left it alone so this represents just the ones I'm certain we no longer use.

PR #5717, opened 2020-11-23
Allow WCA registered pages to register plugin items and categories

joshuatf:
This PR allows items that register with WCA to easily add nav args to add to core menus or create new plugin items/categories.

Note that this does not address issues with locking categories or top-level items which will need a follow-up PR.

### Screenshots
Screen Shot 2020-11-23 at 4 36 17 PM

1. Register some items via `wc_admin_register_page` or `wc_admin_connect_page` with `nav_args`.  This WC Pay PR is a good example - https://github.com/Automattic/woocommerce-payments/pull/1067
1. Note that the items are correctly added to either the core menu or plugins menu when the navigation is enabled.
1. Make sure categories can also be added via the `is_category` property.
1. Make sure no regressions have occurred with other WCA registered items.

### DIFF
--- a/src/Features/Navigation/CoreMenu.php
+++ b/src/Features/Navigation/CoreMenu.php
@@ -143,9 +143,9 @@ class CoreMenu {
 			$wca_items[] = array_merge(
 				array(
 					'id'         => 'wc_admin-' . $page['path'],
-					'url'        => $page['path'],
+					'url'        => isset( $page['path'] ) ? $page['path'] : null,
 					'title'      => $page['title'][0],
-					'capability' => $page['capability'],
+					'capability' => isset( $page['capability'] ) ? $page['capability'] : 'manage_woocommerce',
 				),
 				$page['nav_args']
 			);
--- a/src/Features/Navigation/CoreMenu.php
+++ b/src/Features/Navigation/CoreMenu.php
@@ -75,49 +75,43 @@ class CoreMenu {
 	public static function get_categories() {
 		return array(
 			array(
-				'title'        => __( 'Orders', 'woocommerce-admin' ),
-				'capability'   => 'manage_woocommerce',
-				'id'           => 'woocommerce-orders',
-				'order'        => 10,
-				'is_top_level' => true,
+				'title'      => __( 'Orders', 'woocommerce-admin' ),
+				'capability' => 'manage_woocommerce',
+				'id'         => 'woocommerce-orders',
+				'order'      => 10,
 			),
 			array(
-				'title'        => __( 'Products', 'woocommerce-admin' ),
-				'capability'   => 'manage_woocommerce',
-				'id'           => 'woocommerce-products',
-				'order'        => 20,
-				'is_top_level' => true,
+				'title'      => __( 'Products', 'woocommerce-admin' ),
+				'capability' => 'manage_woocommerce',
+				'id'         => 'woocommerce-products',
+				'order'      => 20,
 			),
 			array(
-				'title'        => __( 'Analytics', 'woocommerce-admin' ),
-				'capability'   => 'manage_woocommerce',
-				'id'           => 'woocommerce-analytics',
-				'order'        => 30,
-				'is_top_level' => true,
+				'title'      => __( 'Analytics', 'woocommerce-admin' ),
+				'capability' => 'manage_woocommerce',
+				'id'         => 'woocommerce-analytics',
+				'order'      => 30,
 			),
 			array(
-				'title'        => __( 'Marketing', 'woocommerce-admin' ),
-				'capability'   => 'manage_woocommerce',
-				'id'           => 'woocommerce-marketing',
-				'order'        => 40,
-				'is_top_level' => true,
+				'title'      => __( 'Marketing', 'woocommerce-admin' ),
+				'capability' => 'manage_woocommerce',
+				'id'         => 'woocommerce-marketing',
+				'order'      => 40,
 			),
 			array(
-				'title'        => __( 'Settings', 'woocommerce-admin' ),
-				'capability'   => 'manage_woocommerce',
-				'id'           => 'woocommerce-settings',
-				'menuId'       => 'secondary',
-				'order'        => 10,
-				'url'          => 'admin.php?page=wc-settings',
-				'is_top_level' => true,
+				'title'      => __( 'Settings', 'woocommerce-admin' ),
+				'capability' => 'manage_woocommerce',
+				'id'         => 'woocommerce-settings',
+				'menuId'     => 'secondary',
+				'order'      => 10,
+				'url'        => 'admin.php?page=wc-settings',
 			),
 			array(
-				'title'        => __( 'Tools', 'woocommerce-admin' ),
-				'capability'   => 'manage_woocommerce',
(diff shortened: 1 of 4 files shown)


src/Features/Navigation/Menu.php
```diff
@@ -399,6 +399,13 @@ public function add_core_items() {
...
 		$items = CoreMenu::get_items();
 		foreach ( $items as $item ) {
+			if ( ! isset( $item['is_top_level'] ) && ! isset( $item['parent'] ) ) {
+				if ( isset( $item['is_category'] ) && $item['is_category'] ) {
+					self::add_plugin_category( $item );
+				} else {
+					self::add_plugin_item( $item );
+				}
+			}
 			self::add_item( $item );
```
joelclimbsthings, line 409: Wouldn't this line result in the items being added twice (or attempting to), since `self::add_plugin_item()` calls `self::add_item()` as well?

joelclimbsthings (changes requested):
Hey Josh, tests well & just had a quick question about the code.

src/Features/Navigation/Menu.php
joshuatf: Good call.  You're right and this bug wasn't showing since we don't allow duplicate IDs.  Thanks for catching this.

joshuatf:
Thanks for the feedback @joelclimbsthings.  I ended up refactoring a bit more in https://github.com/woocommerce/woocommerce-admin/pull/5717/commits/25174248b9ba4ecc6bfb3314522ce5a52065f4bd

The latest update does safe guard and explicitly add top-level items in the primary menu.  It does not allow any other top-level items from WCA to be added.

joelclimbsthings:
Thanks @joshuatf . Good work on the refactor; It's definitely a bit clearer what happening to me. 

Just toying around with this and noticed that if I don't have `is_top_level` or `menuId` set during registration that it ends up defaulting to the `primary` menu in the `Settings` category (this is set to the default on `Menu.php`. Wouldn't we want that to default to the `plugins` menu?

joelclimbsthings:
Also noticed that the `is_top_level` property isn't documented anywhere by the looks of it. We'll want to update the documentation for registration to reflect the changes as well; would that happen in this PR?

joshuatf:
>Wouldn't we want that to default to the plugins menu?

This is the question we need to answer.  I think part of the reason we allow WCA items is because they are more likely to be core-related items.  Do we anticipate many third party plugins (more than the items registered in WCA) registering via WCA?  Or will it be primarily for pages we're hoping get merged to core eventually?

>Also noticed that the is_top_level property isn't documented anywhere by the looks of it. We'll want to update the documentation for registration to reflect the changes as well; would that happen in this PR?

We can update in this PR if we decide to go this route.  It was intentionally not documented to avoid misuse initially.  Now that we're opening up the doors a bit in WCA, we also need to decide if not providing a parent will default to this, which I think is maybe the better solution (at least for the plugins menu) and avoids adding one more property into the mix.

joelclimbsthings:
>  Do we anticipate many third party plugins (more than the items registered in WCA) registering via WCA? Or will it be primarily for pages we're hoping get merged to core eventually?

Right, my assumption so far has been that we _do_ want to encourage the use of 3rd party plugins extending `wc-admin` to simplify things, and that that was part of the value-add. It's totally possible that I'm just missing some context here; I'd be interested to get @psealock 's feedback on this point.

> Now that we're opening up the doors a bit in WCA, we also need to decide if not providing a parent will default to this, which I think is maybe the better solution (at least for the plugins menu) and avoids adding one more property into the mix.

This was a point of confusion for me and I'd definitely support consolidating these two if possible.

joshuatf:
Just to be clear, it's not about whether or not we're encouraging 3pd to extend using WCA (we definitely are)- it's about whether or not those pages are ones that are more likely to end up in core menus because they were extended in WCA or if we still expect most of those items under the "Extensions" menu.

Simply put- do we think more items registered in WCA will fall into the core menu or the extensions menu?  Obviously right now this is heavily weighted towards the core menu, but open to changing this default if we think it will be weighted the other way in the future.

>This was a point of confusion for me and I'd definitely support consolidating these two if possible.

I'll create a follow-up to tackle removing this.  I think we can take out `is_top_level` altogether since the core menu is now walled off.

psealock:
>I think we can take out is_top_level altogether since the core menu is now walled off.

This is my understanding as well. Even WCA extensions can't place top level items. Extensions, Analytics, Settings, or any other extension's category is fair game.

joshuatf:
I went ahead and remove `is_top_level` in this PR as well and gave it a rebase.  This simplifies the logic to:
* If no parent is given, the item is assumed top-level.
* If no parent is given in a WCA page, it gets forced to the plugins menu.

https://github.com/woocommerce/woocommerce-admin/pull/5731 addresses issues related to child items and automatically sets menu ID based on parent.

src/Features/Navigation/CoreMenu.php
```diff
@@ -140,45 +134,62 @@ public static function get_items() {
...
 					'title'      => $page['title'][0],
-					'capability' => $page['capability'],
+					'capability' => isset( $page['capability'] ) ? $page['capability'] : 'manage_woocommerce',
 				),
 				$page['nav_args']
 			);
+
+			// Don't allow top-level items to be added to the primary menu.
+			if ( ! isset( $item['parent'] ) || 'woocommerce' === $item['parent'] ) {
+				$item['menuId'] = 'plugins';
```
joelclimbsthings, line 150: If we want to enforce this by switching the `menuId` automatically, maybe we should still print some kind of warning in the error log?

joelclimbsthings (approved):
Looks great! Tests well too. I left one short comment, but I'll go ahead and pre-approve.

src/Features/Navigation/CoreMenu.php
joshuatf: @joelclimbsthings I think this could result in a pretty long error log, even when the expected occurs (wanting to register in the plugins menu).

Or do you think instead we should require `menuId` to be `plugins` if a parent does not exist and skip adding the item if that's the case?

psealock (approved):
This is testing well for me using the WC Pay equivalent PR.

I made these changes on top of this one to get slotFill working https://github.com/woocommerce/woocommerce-admin/pull/5773

src/Features/Navigation/CoreMenu.php
joshuatf: Going to get this merged so we can unblock other PRs.  Let me know what you think on the above @joelclimbsthings and we can follow-up on this.

joelclimbsthings: That's true, I think the condition would have to be a bit different with a warning. My thought was mainly if the `menuId` was actually set to something other than plugins, and we switched it, a warning would be more justified. 
Something like this, within that condition:
```
if ( ! isset( $item['parent'] ) || 'woocommerce' === $item['parent'] ) {
    if( $item['menuId'] && 'plugins' !== $item['menuId'] ) {
        //print warning with error_log()
    }
    $item['menuId'] = 'plugins';
}
```

Or do you think that would still be too noisy @joshuatf ?

Issue #5602, opened 2020-11-11
Inconsistent data on the Payments task


pmcpinto:
I was checking the setup tasks events, completions rate, etc after the 4.6 release and I noticed something strange:

- There's an increase in the `wcadmin_tasklist_task_completed` event, which was expected as we started auto-completing shipping and taxes in some use cases. But then I segmented by `task_name` and I noticed a spike in the Payments value.
- I then took a look at the events that we track within the payments task (`wcadmin_task_view`, `wcadmin_tasklist_payment_setup`, `wcadmin_tasklist_payment_connect_method`) and the numbers are going in the opposite direction (dropping).
- I created a funnel to check the % of task list views that originated a payments task view before/after 4.6 and it was possible to see a drop in the conv. rate after 4.6
- There's a similar drop in the Add product task views %.
- Curiously all the `Customize Appearance` events started increasing after 4.6

I already considered several options:
- There's a problem with tracking
- We introduced a regression and we're not showing some tasks in some countries: this doesn't seem likely as the payments task view conv. dropped in all the countries
- Users are seeing the inbox notifications to review the shipping and taxes smart defaults and get distracted after being redirected to the shipping or tax settings: I'm not seeing significant differences in the shipping and taxes settings views
- Some hosting company removed some tasks from the setup task list: it would be a bit weird for a hosting company to just remove one task.

After pondering around a bit more here's what I think is happening:
- We have a problem with Tracks due to the mismatch between task completion and all the other events in the payments task
- The shipping and tax smart defaults are causing an unexpected distraction due to the following: when we reorganized the task list in August (placing the payments task right after the add products and the customize appearance task at the bottom) we saw immediately an increase in the activity within the payments task and a decrease in the personalize my store. Mainly because personalize my store is kind of out of sight: https://d.pr/i/EzmmU5. But now when the Shipping and Tax tasks are auto-completed the personalize my store task is right below the add products and payments task, diverging users from those two tasks: https://d.pr/i/WZ6xSi 
The only way that I'm seeing to validate this 2nd hypothesis is by revert the smart defaults implementation

pmcpinto:
cc @timmyc

timmyc:
Just adding a few notes here.

[...]

PR #5720, opened 2020-11-23
Revert the smart tax defaults work of #4594.

samueljseay:
#5602 highlights some issues with payment analytics, and it's been determined that the only path forward here is to revert 2 pieces of work: the smart tax defaults and the reordering of the task list to group items completed. 

This PR reverts the smart tax defaults only, the sorting revert will come in a follow up PR.

1. Before completing the OBW ensure you fulfill the requirements for tax smart defaults:

* Have Jetpack connected
* Have WooCommerce Services installed/TOS accepted

2. Complete the OBW choosing an automated tax supported country (such as US)
3. Ensure that automated tax work is **not** completed, by seeing that the tax task is not completed after completing the OBW.

I think its worth mentioning that you don't really need to do this detailed testing as above, clearly the code has been removed, I think it would also be acceptable to just go through the OBW and confirm that nothing breaks and no exceptions are logged.

### Changelog Note:

Dev: revert the smart tax defaults work.


adrianduffell (approved):
This tests well 🚀 

I followed the full testing instructions. The tax task was incomplete as expected after finishing the OBW. I didn't notice any problems with the OBW during the process.

PR #5722, opened 2020-11-23
Packages: Fix dependencies

psealock:
A lack of declared dependencies has been causing problems in extensions, see p90Yrv-1ZJ

The core issue is the lack of Dependency Extraction Webpack Plugin to handle this for us, so we continue to have these issues.

### Fix

The fix is to add dependencies. 

### Wait, hang on

Why are `'wp-data', 'moment'` added to the `wc-navigation` script when those dependencies aren't used in the code? 

I'm not sure, but it points a bigger problem of those dependencies being improperly declared elsewhere. I tried placing them in different package dependencies, but no luck. Adding Dependency Extraction Webpack Plugin will analyze the code and solve this for us. 

1. Create an extension to use `@woocommerce/navigation` or use this one https://github.com/psealock/aaa
2. Activate the plugin and `npm install && npm start`
3. Navigate to a wc-admin screen and see check the console for no errors and this:

Screen Shot 2020-11-24 at 12 30 21 PM

### Changelog Note:

- Dev: Add correct script dependencies to packages

### DIFF
--- a/packages/navigation/src/index.js
+++ b/packages/navigation/src/index.js
@@ -5,12 +5,7 @@ import { addQueryArgs } from '@wordpress/url';
 import { parse } from 'qs';
 import { pick, uniq } from 'lodash';
 import { applyFilters } from '@wordpress/hooks';
-import {
-	Slot,
-	Fill,
-	SlotFillProvider,
-	__experimentalUseSlot as useSlot,
-} from '@wordpress/components';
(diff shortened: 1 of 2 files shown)


joshuatf (approved):
Testing well and LGTM!  This definitely emphasizes our need for the dependency extraction.  Thanks for looking into this and getting this fixed up, Paul. 🎉

PR #5725, opened 2020-11-24
fix: remove Historical Analytics Data notice #5286

### LINKED ISSUE #5286: Remove "Historical Analytics Data" notice
**Describe the bug**

We've been experiencing this issue consistently on a number of sites, usually with the new-style "Database Update Done" notices. 

After a successful database upgrade, clicking "Thanks" seems to have no effect. Usually insisting a few times has the desired effect, but there is at least one local site/db I'm working with where the notice doesn't seem to go away.

I have found the "Historical Analytics Data" notice to behave similarly, probably by design? If the user doesn't want to import Analytics Data, the notice can't be hidden -- clicking "Get Started" has no effect and there is no "Dismiss" action.

**To Reproduce**

1. Update from WooCommerce 3.9
2. Trigger the DB upgrade process from the PHP notice that appears in the Dashboard upon logging in.
3. Wait for the DB upgrade to complete and reload the page.
4. Dismiss the PHP "thanks" notice that appears when the DB upgrades are complete.
5. Go to **WooCommerce > Home**.
6. You'll probably see another JS notice that the DB upgrade process is complete that can't be dismissed, along with the "Historical Analytics Data" notice, which can't be dismissed either.

**Expected behavior**

The new WooCommerce notices implemented in wc-admin should be dismissible, ideally without reloading the page.

On an unrelated note, it would be great if these notices followed all WordPress notice design conventions. Currently, WooCommerce is introducing its own notifications system, which sometimes results in the kind of WP experience it is trying to avoid (consistency suffers when the user sees 2 different notices about the same thing -- updating the DB -- that look different when moving from one admin screen to another).

ilyasfoo:
Fixes #5286

This PR removes the "Historical Analytics Data" notice as now that analytics has been in core since 4.0.

Due to the difficulty of testing this change with an older version of WooCommerce, it might be better to review the source code changes manually instead.

### Changelog Note:

fix: remove Historical Analytics Data notice #5286


louwie17 (dismissed):
Changes look good, but it appears the HistoricalData is also being imported in the FeaturePlugin.php -> woocommerce-admin/src/FeaturePlugin.php:186
See error message in the build: https://travis-ci.com/github/woocommerce/woocommerce-admin/jobs/446151656

Please remove the use of HistoricalData there as well.

ilyasfoo:
My bad, thanks for catching that! I've removed all references to HistoricalData.

Please have another look, thanks!

jeffstieler:
We're going to delte the [`WC_Admin_Notes_Historical_Data`](https://github.com/woocommerce/woocommerce-admin/blob/51fac2fd7c42c169ba9b4046816e7858bdb6d391/src%2FNotes%2FDeprecatedNotes.php#L293) facade class as well. While this isn't 100% in line with backwards compatibility, we've deleted note classes in the past and it's extremely unlikely this class is being used by a 3PD.

ilyasfoo:
Deleted facade in https://github.com/woocommerce/woocommerce-admin/pull/5725/commits/6798709e23a2b46304031ed671ddaf59edc1ca27

jeffstieler (approved):
Looks good. I verified that there were no more references to `HistoricalData` in the PHP. 👍

becdetat:
@ilyasfoo This PR looks like it's ready to merge, can you do that please?

ilyasfoo:
Thank you for mentioning me @becdetat! Sorry I've missed this previously.

PR #5731, opened 2020-11-24
Only show menu items from same parent menu 

### LINKED ISSUE #5718: Secondary menu items show regardless of parent's menu
**Describe the bug**
When an item is registered in the extensions menu, it can still be shown under the primary menu's item parent by specifying that parent ID.

**To Reproduce**

1. Add an item in the extensions menu with a `parent` from the primary menu.
```php
		\Automattic\WooCommerce\Admin\Features\Navigation\Menu::add_plugin_item(
			array(
				'id'     => 'wrong-menu-item',
				'parent' => 'analytics',
				'title'  => 'I should not be shown',
				'url'    => 'https://www.youtube.com/watch?v=dQw4w9WgXcQ',
			)
		);
```
2. Note that menu item shows when clicking the primary menu category specified.

**Expected behavior**
Expected items to respect the menu they're in.

**Screenshots**
Screen Shot 2020-11-23 at 4 51 35 PM

joshuatf:
Fixes #5718

Fixes an issue where menu items could have the same parent, but different menu IDs and still be shown together.

1. Register a menu item in the plugins menu, but using a parent ID from the main menu:
```php

		\Automattic\WooCommerce\Admin\Features\Navigation\Menu::add_plugin_item(
			array(
				'id'     => 'wrong-menu-item',
				'parent' => 'analytics',
				'title'  => 'I should not be shown',
				'url'    => 'https://www.youtube.com/watch?v=dQw4w9WgXcQ',
			)
		);
```
2. Navigate to that category.
3. Make sure the item is not shown.
4. Make sure no other regressions occur with items showing across various menu categories.

### DIFF
--- a/client/navigation/components/container/index.js
+++ b/client/navigation/components/container/index.js
@@ -137,7 +137,7 @@ const Container = ( { menuItems } ) => {
 						const {
 							primary: primaryItems,
 							secondary: secondaryItems,
-							plugin: pluginItems,
+							plugins: pluginItems,
 						} = categorizedItems[ category.id ] || {};
 						return (
 							<NavigationMenu
--- a/client/navigation/components/container/index.js
+++ b/client/navigation/components/container/index.js
@@ -30,8 +30,7 @@ const Container = ( { menuItems } ) => {
 
 	const dashboardUrl = getAdminLink( '' );
 
-	const categories = menuItems.filter( ( item ) => item.isCategory );
-	categories.push( {
+	const parentCategory = {
 		capability: 'manage_woocommerce',
 		id: 'woocommerce',
 		isCategory: true,
@@ -40,7 +39,19 @@ const Container = ( { menuItems } ) => {
 		order: 10,
 		parent: '',
 		title: 'WooCommerce',
-	} );
+	};
+	const categoriesMap = menuItems.reduce(
+		( acc, item ) => {
+			if ( item.isCategory ) {
+				return { ...acc, [ item.id ]: item };
+			}
+			return acc;
+		},
+		{
+			woocommerce: parentCategory,
+		}
+	);
+	const categories = Object.values( categoriesMap );
 
 	const [ activeItem, setActiveItem ] = useState( 'woocommerce-home' );
 	const [ activeLevel, setActiveLevel ] = useState( 'woocommerce' );
@@ -66,9 +77,20 @@ const Container = ( { menuItems } ) => {
 
 	const getMenuItemsByCategory = ( items ) => {
 		return items.reduce( ( acc, item ) => {
+			// Set up the category if it doesn't yet exist.
 			if ( ! acc[ item.parent ] ) {
 				acc[ item.parent ] = [ [], [], [] ];
 			}
+
+			// Check if parent category is in the same menu.
+			if (
+				item.parent !== 'woocommerce' &&
+				categoriesMap[ item.parent ] &&
+				categoriesMap[ item.parent ].menuId !== item.menuId
+			) {
+				return acc;
+			}
+
 			let index = 0;
 			if ( item.menuId === 'secondary' ) {
 				index = 1;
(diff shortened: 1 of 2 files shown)


client/navigation/components/container/index.js
```diff
@@ -87,6 +87,26 @@ const Container = ( { menuItems } ) => {
...
 	const navDomRef = useRef( null );
 
+	const renderNavigationGroup = ( category, items, topLevelTitle = null ) => {
+		if ( ! items.length ) {
+			return null;
+		}
+
+		const isTopLevel = category.id === 'woocommerce';
+
+		if ( ! isTopLevel && items[ 0 ].menuId !== category.menuId ) {
```
joelclimbsthings, line 97: Looking at the code, I think we can expect that this would never be called where `items[ 0 ]` had the same menu ID, but `items[ 1 ]` did not?

client/navigation/components/container/index.js
```diff
@@ -128,44 +148,18 @@ const Container = ( { menuItems } ) => {
...
-										{ pluginItems.map( ( item ) => (
-											<Item
-												key={ item.id }
-												item={ item }
-											/>
-										) ) }
-									</NavigationGroup>
+								{ renderNavigationGroup(
+									category,
+									pluginItems,
```
joelclimbsthings, line 157: So it looks like this will often be called with `0` items for most categories, which returns null as the first line of the function. Works well, although I wonder if we could refactor this to make the code clearer at some point? 

I don't think this has any real performance consequences so I'm not too worried in the context of this PR.

joelclimbsthings (commented):
Hey @joshuatf , looks and works great! Glad to see we were sync'd up with our thinking on this. 

I left a couple comments I was curious about, but likely don't require changes at this point. Just curious what you think.

Also, I tested this using your test instructions and all was well (the item didn't show up), ~~although I actually noticed that the problem didn't show up with the current state of `main` for me either. So I wasn't able to reproduce for :shrug:~~

Edit: Scratch that last comment, it was just because I was listing the parent as `analytics` instead of `woocommerce-analytics`

client/navigation/components/container/index.js
joshuatf: I don't think there's any measurable performance impact, but what would you like to see to make it clearer?

We could remove the `return null` logic and instead wrap each of the groups with an item check.   This isn't very DRY, but if you feel it makes it clearer then I'm not opposed.
```js
{ primaryItems.length && renderNavigationGroup(
	category,
	primaryItems
) }
{ pluginItems.length && renderNavigationGroup(
	category,
	pluginItems,
	__( 'Extensions', 'woocommerce-admin' )
) }
{ secondaryItems.length && renderNavigationGroup(
	category,
	secondaryItems
) }
```

client/navigation/components/container/index.js
joshuatf: Right, since they're already categorized all items should have the same menu ID, though I think there's probably a better way to handle this.

joshuatf:
Thanks for the review @joelclimbsthings - I refactored some of the existing code in https://github.com/woocommerce/woocommerce-admin/pull/5731/commits/51c1b73314ea830f2e6d4fabb9743a18a09a27d9 so that we can filter categorized items ahead of time.

Also found another loophole to move an item to another menu; fixed in https://github.com/woocommerce/woocommerce-admin/pull/5731/commits/4dd5b45e2989eebe1e5e92f7d9207219dc7d9982

client/navigation/components/container/index.js
joelclimbsthings: Looks like this is the approach after the recent changes. Not a big deal at all, but I think this is definitely a bit clearer coming at it with no foreknowledge of what's happening. I like it :+1:

joelclimbsthings (approved):
Good catch on that last loophole :1st_place_medal: 

And I like the refactored approach as well. :shipit:

PR #5739, opened 2020-11-25
Fix/some outstanding CES bugs

### LINKED ISSUE #5468: Customer effort score: Show comments section in survey

becdetat:
This fixes 3 issues that were uncovered by early testing:

- a warning on the type of the `trackProps` prop coming from the settings update enqueue
- the submit label being undefined in the settings update enqueue
- comments not being included when recording the feedback

1. go to the WooCommerce settings page
1. open the console to see any errors
1. change a setting
1. save the change
1. when the page reloads, the CES snackbar prompt should appear
1. click 'Give feedback'
1. note that no console errors appear
1. give a low rating, triggering the comments section to appear
1. enter a comment then send
1. note that the success snackbar content is not 'undefined'
1. there should be a tracks notice in the console, including the comments you provided
image

### DIFF
--- a/client/customer-effort-score-tracks/customer-effort-score-tracks.js
+++ b/client/customer-effort-score-tracks/customer-effort-score-tracks.js
@@ -98,10 +98,11 @@ function CustomerEffortScoreTracks( {
 		} );
 	};
 
-	const recordScore = ( score ) => {
+	const recordScore = ( score, comments ) => {
 		recordEvent( 'ces_feedback', {
 			action,
 			score,
+			comments,
 			store_age: storeAge,
 			...trackProps,
 		} );
@@ -153,7 +154,6 @@ CustomerEffortScoreTracks.propTypes = {
 	 * Whether tracking is allowed or not.
 	 */
 	allowTracking: PropTypes.bool,
-
 	/**
 	 * Whether props are still being resolved.
 	 */
--- a/client/customer-effort-score-tracks/customer-effort-score-tracks.js
+++ b/client/customer-effort-score-tracks/customer-effort-score-tracks.js
@@ -102,7 +102,7 @@ function CustomerEffortScoreTracks( {
 		recordEvent( 'ces_feedback', {
 			action,
 			score,
-			comments,
+			comments: comments || '',
 			store_age: storeAge,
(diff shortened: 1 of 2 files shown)


client/customer-effort-score-tracks/customer-effort-score-tracks.js
```diff
@@ -98,10 +98,11 @@ function CustomerEffortScoreTracks( {
 		} );
 	};
 
-	const recordScore = ( score ) => {
+	const recordScore = ( score, comments ) => {
 		recordEvent( 'ces_feedback', {
 			action,
 			score,
+			comments,
```
mattsherman, line 105: If comments are not entered, they get sent as `undefined`... do we know if this will cause any issues actually recording to Tracks? I don't *think* so.

mattsherman (approved):
Nice job fixing up these outstanding issues.

Tested and things work as expected.

Just left a note about sending an `undefined` to Tracks... I don't think it will be an issue, but you may want to double-check this before merging (I ran out of time to test actually sending to Tracks in my env this morning).

client/customer-effort-score-tracks/customer-effort-score-tracks.js
becdetat: I don't think so either, but I've changed this to use `''` if `comments` is falsey. Thanks :-)

PR #5746, opened 2020-11-25
Cleanup items store actions

### LINKED ISSUE #5742: Items data store action(s) cleanup
**Describe the bug**

In developing https://github.com/woocommerce/woocommerce-admin/pull/5729 I noticed some weird actions and behaviors in the "items" data store (`/packages/data/src/items`):

* When using `updateProductStock()` to update a product, the updated product object gets dispatched as an action (I think this happens errantly in the action creator and actually works because of the product's `type` property)

Screen Shot 2020-11-25 at 11 14 58 AM

* `SET_ITEMS` is used even when updating a single item.. seems like overkill and now has a side effect of errantly updating item totals

**To Reproduce**

1. Go to WooCommerce > Home
2. Ensure you have an out of stock product
3. Update the product stock
4. Use Redux DevTools to verify the errant actions

**Expected behavior**

At the very least, we should eliminate the errant "product as action" behavior.

I think that the addition of a `SET_ITEM` action (singular) and perhaps a `SET_ITEMS_TOTAL_COUNT` action could clean up the state management for this store and make the logic a little easier to follow.

**Screenshots**

See above.

jeffstieler:
Fixes #5742

This PR makes the following changes to the "items" store:

* Eliminates errant product-as-action dispatch from `updateProductStock()`
* Fixes calls to `setError()`
* Add new `SET_ITEM` action for updating single items
* Add new `SET_ITEMS_TOTAL_COUNT` action to separate updating items and the total count of them
* Stop using `getItems()` inside `getItemsTotalCount` - the queries are different

- Go to WooCommerce > Home
- Ensure you have an out of stock product
- Update the product stock
- Use Redux DevTools to verify there is not an action that is really the product object
- Smoke test reviews panel
- Smoke test other items endpoint consumers like category filter, product filter, 

### Changelog Note:

Dev: cleanup some items datastore code.


louwie17 (approved):
Changes look great, and everything still works as intended. Nice catch on the set error 👍  🎉

PR #5754, opened 2020-11-27
Fix small issues with customer effort score package

becdetat:
This fixes some comments that @samueljseay made (p1606345977124800-slack-CBKK5KM41) around the `@woocommerce/customer-effort-score` package:

- update dependencies in `package.json`
- remove lodash dependency
- use `useEffect` rather than just setting state right there in the render. Whoops! 

Smoke-test the CES feature (delete the `woocommerce_ces_shown_for_actions` option to reset the feature if you've tested it before). Add a product, edit a product, change a setting, and/or change an analytics report filter.

To test the package dependencies, you need to create a test plugin that uses the component:

In a terminal, starting from `plugins/woocommerce-admin/`:
```
npm run build
cd packages/customer-effort-score
npm link
cd ../..
npm run create-wc-extension
# give the new extension a name like 'ces-test'
cd ../ces-test
npm install
npm link @woocommerce/customer-effort-score
npm start
```

Go to 'Plugins' in settings and activate the new 'ces-test' plugin

Edit `plugins/ces-test/src/index.js` and replace the contents with:
```js
import CustomerEffortScore from '@woocommerce/customer-effort-score/src';
import './index.scss';

const props = {
    recordScoreCallback: ( params ) => console.log( 'recordScoreCallback', params ),
    label: 'Customer effort score label',
};

console.log( 'the actual thing', CustomerEffortScore );
console.log( 'CustomerEffortScore: ', <CustomerEffortScore { ...props } /> );
```

Go to a wc-admin page and check the console for the two messages and make sure the `CustomerEffortScore` component looks correct.


src/Loader.php
```diff
@@ -367,7 +367,17 @@ public static function register_scripts() {
...
 			self::get_url( 'customer-effort-score/index', 'js' ),
-			array(),
+			array(
+				'wp-components',
+				'wp-compose',
+				'wp-data',
+				'wp-element',
+				'wp-18n',
+				'wp-notices',
+				'classnames',
```
adrianduffell: I don't think we need to include the non-WordPress packages in this array (`classnames`, `prop-types`, `react-transition-group`).  I think those packages are bundled by Webpack instead?

adrianduffell (commented):
Overall this looks good 👍 

I had some permission issues using `npm link` and couldn't complete the testing instructions. Could someone else please try testing it?

src/Loader.php
becdetat: Thanks @adrianduffell , I took these out and fixed a typo (`wp-18n` vs `wp-i18n`) which was stopping this from working :-$

mattsherman (approved):
Smoked tested and things still function as expected.

Ran through testing instructions -- everything worked as expected.

Code looks reasonable. Nice work!

PR #5759, opened 2020-11-27
Prevent unnecessary re-renders when using useSelect

louwie17:
Make sure we don't create either a new function/object/array within a `useSelect` function and return this as part of the data.
`useSelect` only does a [shallowEqual](https://github.com/WordPress/gutenberg/blob/master/packages/data/src/components/use-select/index.js#L148) which would cause new functions/objects/arrays to be different every time (despite the data within being the same).

This in turn would trigger unnecessary [forceRender()](https://github.com/WordPress/gutenberg/blob/master/packages/data/src/components/use-select/index.js#L156) for the components where it used this.
I didn't go through the entire code to check for occurrences of this, I fixed it for the `useUserPreferences` hook, and the home screen `ActivityPanel` component.
I couldn't see a huge visual difference, as we aren't doing anything compute heavy within the components, but you can see the scripting time difference in the screenshots below.

### Screenshots
Before the fix, you can see that scripting time is `~5616ms`.
Screen Shot 2020-11-27 at 10 18 12 AM

After the change: (scripting time: `~3621ms`).
Screen Shot 2020-11-26 at 5 05 59 PM

- Open Woo-commerce home page
- It should all the default panels correctly
- Try updating the display layout, it should correctly update (without error) and be persistent on refresh
- Smoke test places where `useUserPreferences` is being used
  - ReportsTable in every analytics page

### DIFF
--- a/client/homescreen/activity-panel/index.js
+++ b/client/homescreen/activity-panel/index.js
@@ -18,21 +18,24 @@ import {
 import { getAllPanels } from './panels';
 
 export const ActivityPanel = () => {
-	const panels = useSelect( ( select ) => {
+	const panelsData = useSelect( ( select ) => {
 		const totalOrderCount = getSetting( 'orderCount', 0 );
 		const orderStatuses = getOrderStatuses( select );
 		const countUnreadOrders = getUnreadOrders( select, orderStatuses );
 		const manageStock = getSetting( 'manageStock', 'no' );
 		const countLowStockProducts = getLowStockCount( select );
 
-		return getAllPanels( {
+		return {
 			countLowStockProducts,
 			countUnreadOrders,
 			manageStock,
 			orderStatuses,
 			totalOrderCount,
-		} );
+		};
 	} );
+
+	const panels = getAllPanels( panelsData );
+
 	return (
 		<Accordion>
 			<Fragment>
--- a/packages/data/src/user-preferences/use-user-preferences.js
+++ b/packages/data/src/user-preferences/use-user-preferences.js
@@ -28,6 +28,87 @@ const getWooCommerceMeta = ( user ) => {
 	return userData;
 };
 
+// Create wrapper for updating user's `woocommerce_meta`.
+async function updateUserPrefs(
+	receiveCurrentUser,
+	user,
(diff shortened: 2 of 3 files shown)


client/homescreen/activity-panel/index.js
```diff
@@ -18,21 +18,24 @@ import {
...
 			countLowStockProducts,
 			countUnreadOrders,
 			manageStock,
 			orderStatuses,
 			totalOrderCount,
-		} );
+		};
 	} );
+
+	const panels = getAllPanels( panelsData );
```
louwie17, line 37: Generate panel array outside of `useSelect` so that it doesn't force a re-render every-time.

packages/data/src/user-preferences/use-user-preferences.js
```diff
@@ -39,140 +120,76 @@ export const useUserPreferences = () => {
...
-					updatedUser: updatedUserResponse,
-				};
-			};
-
-			return {
-				isRequesting:
-					hasStartedResolution( 'getCurrentUser' ) &&
-					! hasFinishedResolution( 'getCurrentUser' ),
-				userPreferences: userData,
-				updateUserPreferences: updateUserPrefs,
```
louwie17, line 169: `userData` and `updateUserPrefs` where causing the force re-render as both of them where newly created within the `useSelect`. I moved both outside of the `useSelect`, this required some re-arranging.

The functions like `getEntity`, `getEntityRecord`, `getLastEntitySaveError` aren't a problem, as they are coming from the select method and are constant.

packages/data/src/user-preferences/use-user-preferences.js
```diff
@@ -28,6 +28,87 @@ const getWooCommerceMeta = ( user ) => {
 	return userData;
 };
```
louwie17, line 29: Mostly re-arranging, see comment further down in file.

jeffstieler (approved):
Nice work @louwie17! This is testing well for me, the code changes look good, and I see the reduction in scripting time. 🚀

timmyc:
Wow, awesome find and fix @louwie17

PR #5761, opened 2020-11-30
Test install script resets test environment

ilyasfoo:
This PR changes `bin/install-wp-tests.sh` script to additionally do the following prior setting up:

- Drops test database.
- Deletes `wp-config.php`.
- Deletes test suite `/tmp/wordpress-tests-lib`
- Deletes `wp-content/plugins/woocommerce` folder.

The rationale is that I think our setup test script should reset on every run. Use cases:

- Setup ran halfway initially and then subsequently blocked because of existing files & database, thus unable to finish the whole setup.
- Potential intermittent issues due to persisting data that affects tests in between runs.
- Misbehaving tests and you're unsure whether it's your code or the environment.
- To apply updates from tests' dependencies

Note that this doesn't delete everything. Notably, the existing WordPress files will still be reused. 

1. In your PHP test environment, run either the following:
    - `bin/install-wp-tests.sh wc-admin-tests root root`
2. Setup should be successful and finish at `Plugin 'woocommerce' activated.
Success: Activated 1 of 1 plugins.`
3. `./vendor/bin/phpunit` and see if it runs well.

### DIFF
--- a/bin/install-wp-tests.sh
+++ b/bin/install-wp-tests.sh
@@ -20,7 +20,7 @@ while [ $# -gt 0 ]; do
       RESET_FLAG=true
       ;;
   esac
-  shift
+	shift
 done
 
 TMPDIR=${TMPDIR-/tmp}
@@ -164,7 +164,7 @@ install_db() {
 
 	# drop existing database
 	if [ ${RESET_FLAG} = true ]; then
-		mysqladmin drop -f $DB_NAME --user="$DB_USER" --password="$DB_PASS"$EXTRA 2>&1 | true
+		mysqladmin drop -f $DB_NAME --user="$DB_USER" --password="$DB_PASS"$EXTRA || echo "Database $DB_NAME does not exist"
 	fi
 
 	# create database
@@ -185,8 +185,8 @@ install_deps() {
 
 	# delete existing wp-config and woocommerce repository
 	if [ ${RESET_FLAG} = true ]; then
-		rm wp-config.php | true
-		rm -rf wp-content/plugins/woocommerce | true
+		rm wp-config.php || echo "wp-config.php does not exist"
+		rm -rf wp-content/plugins/woocommerce || echo "wp-content/plugins/woocommerce does not exist"
 	fi
 
 	php wp-cli.phar core config --dbname=$DB_NAME --dbuser=$DB_USER --dbpass=$DB_PASS --dbhost=$DB_HOST --dbprefix=wptests_
--- a/bin/install-wp-tests.sh
+++ b/bin/install-wp-tests.sh
@@ -119,6 +119,11 @@ install_test_suite() {
 		local ioption='-i'
 	fi
 
+	# removes testing suite
+	if [ ${RESET_FLAG} = true ]; then
+		rm -rf $WP_TESTS_DIR
+	fi
+
 	# set up testing suite if it doesn't yet exist
 	if [ ! -d $WP_TESTS_DIR ]; then
 		# set up testing suite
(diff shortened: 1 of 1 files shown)


adrianduffell:
> The resulting command will require all positional arguments to be defined as well; I was afraid to move those as it might break things. If you can suggest anything that can help, I would love to hear them!

Maybe the flag is not needed, and the script should always reset any existing dependencies if they exist? I feel like this would be the expected behavior when running the script.

ilyasfoo:
Now that you've mentioned it, I 100% agree that this should've been the default behavior. I'm just going to double confirm that this does not have any impact on anything else (I doubt it does, but just to be safe)

ilyasfoo:
I've made it the default behavior in https://github.com/woocommerce/woocommerce-admin/pull/5761/commits/fb069ddf7508357553e061b68042b8144313759e, checked and it ran well in Travis where it's used.

bin/install-wp-tests.sh
```diff
@@ -166,6 +172,11 @@ install_deps() {
 	cd "$WP_CORE_DIR"
 
 	curl -O https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar
+
+	# delete existing wp-config and woocommerce repository
+	rm wp-config.php || echo "wp-config.php does not exist"
```
adrianduffell, line 177: `rm -f` could be used here to hide the command’s error message when the file does not exist. I see this in the CI logs:

```
+rm wp-config.php
rm: cannot remove ‘wp-config.php’: No such file or directory
+echo 'wp-config.php does not exist'
```

bin/install-wp-tests.sh
```diff
@@ -151,6 +154,9 @@ install_db() {
 		fi
 	fi
 
+	# drop existing database
+	mysqladmin drop -f $DB_NAME --user="$DB_USER" --password="$DB_PASS"$EXTRA || echo "Database $DB_NAME does not exist"
```
adrianduffell, line 158: This command is a little noisy in the CI log because the database doesn't previously exist there.

```
+mysqladmin drop -f wc_admin_test --user=root --password= --host=localhost --protocol=tcp
Warning: Using a password on the command line interface can be insecure.
mysqladmin: DROP DATABASE wc_admin_test failed;
error: 'Can't drop database 'wc_admin_test'; database doesn't exist'
+echo 'Database wc_admin_test does not exist'
Database wc_admin_test does not exist
```

Perhaps it would be better to execute SQL directly so that `IF EXISTS` could be used? Something like:
```
DROP wc_admin_test IF EXISTS
```

adrianduffell (changes requested):
Thanks @ilyasfoo , I think this works well as the default behavior 👍 

 I think we could remove the output when items don't already exist, to prevent false alarms in the CI log. I added some suggestions in the code comments.

ilyasfoo:
You've raised an excellent point @adrianduffell! I ended up silencing removal errors because they do not add useful context to the build process. Plus the file & database creation will end up throwing errors if they exist anyway, making the messages redundant.

Changes made in https://github.com/woocommerce/woocommerce-admin/pull/5761/commits/c499b31725ceebe5d08f185f193eb5488c708d99

bin/install-wp-tests.sh
ilyasfoo: Good point!

bin/install-wp-tests.sh
ilyasfoo: I've tried to use direct query but it did still produce an error message albeit shorter, so ended up with `2>/dev/null` to silence it out.

adrianduffell (approved):
Thanks @ilyasfoo the output LGTM! 🚀

PR #5768, opened 2020-11-30
Generate MD5 hashes without bundling all of Node crypto.

jeffstieler:
While looking at https://github.com/woocommerce/woocommerce-admin/pull/5753, I noticed that `@woocommerce/data` was pretty bloated due to the `crypto-browserify` (really the `node-libs-browser`) package.

Since we're only using the `crypto` library to generate MD5 hashes, this seemed like an easy package to replace.

This change **reduces the build by 1.2MB** ❗ 

- Smoke test WC-Admin, make sure functionality is working
- The two directly impacted areas are the Report Export and the Reviews card (Gravatar)
- `npm ci && npm run analyze` on `main` and this branch - see the size difference

### Changelog Note:

Performance: reduce bundle size by omitting Node crypto library.


joelclimbsthings (approved):
Awesome job @jeffstieler ! And thanks for finally giving me a reason to load some test data and explore some site functionality I'd yet to see. 

Ran through test instructions & recorded size reduction from 3.92 -> 2.69MB, with no issues seen in the aforementioned areas (Gravatar and report export). 

:shipit:

PR #5770, opened 2020-11-30
Dependency Extraction Webpack Plugin: Handle irregular dependency names

psealock:
I noticed in https://github.com/woocommerce/woocommerce-admin/pull/5762 while adding Dependency Extraction Webpack Plugin that `csvExport` and `data` packages don't follow the normal convention. This PR handles those and readies the packgage for another release.

1. See the names of scripts registered 
https://github.com/woocommerce/woocommerce-admin/blob/6fba36c4082a915cb4ef4af49caa6ec6313aa504/src/Loader.php#L350
2. Make sure that csv and data are the only packages that don't translate `@woocommerce/<package>` to `wc-<package>`.
3. Make sure the code makes sense.

### DIFF
--- a/bin/starter-pack/_package.json
+++ b/bin/starter-pack/_package.json
@@ -22,6 +22,6 @@
 	"devDependencies": {
 		"@wordpress/scripts": "^12.2.1",
 		"@woocommerce/eslint-plugin": "1.0.0-beta.0",
-		"@woocommerce/dependency-extraction-webpack-plugin": "1.0.1"
+		"@woocommerce/dependency-extraction-webpack-plugin": "1.1.0"
 	}
 }
--- a/packages/dependency-extraction-webpack-plugin/CHANGELOG.md
+++ b/packages/dependency-extraction-webpack-plugin/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 1.1.0
+
+-   Fix: Handle csv-export and data packages as they don't conform to a pattern.
+
 # 1.0.1
 
 -   Fix: Avoid tanspiling packaged code.
--- a/packages/dependency-extraction-webpack-plugin/CHANGELOG.md
+++ b/packages/dependency-extraction-webpack-plugin/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 1.1.0
 
--   Fix: Handle csv-export and data packages as they don't conform to a pattern.
+-   Fix: Handle irregular package names that don't conform to a pattern.
 
 # 1.0.1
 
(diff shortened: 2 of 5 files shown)


psealock:
@senadir It looks like `@woocommerce/blocks-registry` also doesn't follow a regular pattern in `requestToExternal`.

`@woocommerce/blocks-registry` would normally become `[ 'wc', 'blocksRegistry' ]` but the actual item on the window is `wc.wcBlocksRegistry` so the array should be `[ 'wc', 'wcBlocksRegistry' ]`.

If thats right, shall I add that exception to this PR?

senadir:
Hey Paul, yeah we have a bunch of them
```
const wcDepMap = {
	'@woocommerce/blocks-registry': [ 'wc', 'wcBlocksRegistry' ],
	'@woocommerce/settings': [ 'wc', 'wcSettings' ],
	'@woocommerce/block-data': [ 'wc', 'wcBlocksData' ],
	'@woocommerce/shared-context': [ 'wc', 'wcSharedContext' ],
	'@woocommerce/shared-hocs': [ 'wc', 'wcSharedHocs' ],
};
```
I discussed renaming them with Darren and it seems like it would break backwards comp with no value, so add a custom rule for `wcBlocksRegistry`, other ones can be added by the consumer if he wishes.

psealock:
> other ones can be added by the consumer if he wishes.

I think its just as easy to add all of them. I'm not sure an extension developer will automatically know to add these.

senadir:
I'm not sure if all of them point to the correct thing or if they're exposed publicly via a package or not.
thing like `@woocommerce/settings` and `@woocommerce/block-data`.
I don't think anyone else is using them (apart from us) so that's why I said we can hold on adding them until I'm sure (so we don't deal with backward compatibility later if something change).
Also non of them should be public (not including Woo teams).
If you still think we should add them then go for it.

senadir:
With that in mind, I think we should consider how `@woocommerce/settings` is used in wc-admin repo and blocks repo.
in ours, it's exposed as `[ 'wc', 'wcSettings' ]` while I believe in wc-admin it's on window directly?

psealock:
>while I believe in wc-admin it's on window directly?

It is, but we discourage its use directly. Internally, we have an alias for that "package".

psealock:
@senadir https://github.com/woocommerce/woocommerce-admin/pull/5770/commits/aa59fd5aba4a605721a308ba8f285031b608f8e0 adds a mapping for blocks registry. Since the handle [is regular format](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/c873eec1aa8f2581046cb9691002a4bb6017ffbf/bin/webpack-helpers.js#L26), no need to add an exception there.

psealock:
Ready for review if you have a moment @senadir, thanks!

becdetat:
Pinging @senadir for a review :-)

senadir (approved):
Sorry for blocking this PR and taking so long to review, it looks good to me and it's ready to merge.

PR #5771, opened 2020-12-01
Add onboarding settings only for the home and setup wizard pages

### LINKED ISSUE #5702: [4.8-beta.1] PHP Timeout TwentyTwentyOne/PHP8/WP 5.6 release 1
**Describe the bug**
Logging an issue originally reported by @nerrad on slack today.

Screenshot because transcribing might be time consuming :)

image

While I haven't attempted to repro the error myself, I'm wondering if we should look at adding another "bail early" clause to this onboarding logic so we only spend the cycles/time gathering the onboarding data for `wcSettings` if the active page is wc-admin... otherwise there really shouldn't be any other need to assemble a list of themes that support woo, and all the other bits we are doing to hydrate the state for the OBW.

moon0326:
Fixes #5702 

This PR improves the performance of non-WooCommerce pages by skipping [Onboarding::component_settings](https://github.com/woocommerce/woocommerce-admin/blob/main/src/Features/Onboarding.php#L652) method for the non-WooCommerce pages.

- Removed unnecessary call to **wp_get_themes** at https://github.com/woocommerce/woocommerce-admin/blob/6fba36c4082a915cb4ef4af49caa6ec6313aa504/src/Features/Onboarding.php#L511 Notice that the call is inside a loop.
- Uses screen object to check for the WC Admin pages.

1. Install a fresh WordPress and WooCommerce.
2. Navigate to WooCommerce -> Home or WooCommerce -> Home -> click one of the onboarding items.
3. Open browser console and type`wcSettings.onboarding` and make sure that the object contains onboarding properties such as `themes` and `tasksStatus`
4. Navigate to any pages, except home or setup wizard.
5. Open browser console and type `wcSettings.onboarding` and make sure that the object is missing onboarding properties such as `themes` and `tasksStatus`

### DIFF
--- a/src/Features/Onboarding.php
+++ b/src/Features/Onboarding.php
@@ -507,9 +507,8 @@ class Onboarding {
 			$active_theme     = get_option( 'stylesheet' );
 
 			foreach ( $installed_themes as $slug => $theme ) {
-				$theme_data       = self::get_theme_data( $theme );
-				$installed_themes = wp_get_themes();
-				$themes[ $slug ]  = $theme_data;
+				$theme_data      = self::get_theme_data( $theme );
+				$themes[ $slug ] = $theme_data;
 			}
 
 			// Add the WooCommerce support tag for default themes that don't explicitly declare support.
@@ -645,9 +644,26 @@ class Onboarding {
 	}
 
 	/**
+	 * Check if the current page is one of the WC Admin pages.
+	 *
+	 * @return bool
+	 */
+	public static function is_woocommerce_admin_page() {
+		$is_wc_admin_page = false;
+		$current_screen   = get_current_screen();
+		if ( null !== $current_screen ) {
+			$is_wc_admin_page = 'woocommerce_page_wc-admin' === $current_screen->id;
+		}
+
+		return $is_wc_admin_page;
+	}
+
+	/**
 	 * Add profiler items to component settings.
 	 *
 	 * @param array $settings Component settings.
+	 *
+	 * @return array
 	 */
 	public function component_settings( $settings ) {
 		$profile                = (array) get_option( self::PROFILE_DATA_OPTION, array() );
@@ -655,8 +671,14 @@ class Onboarding {
 			'profile' => $profile,
 		);
 
-		// Only fetch if the onboarding wizard OR the task list is incomplete or currently shown.
-		if ( ! self::should_show_profiler() && ! self::should_show_tasks() ) {
+		// Only fetch if the onboarding wizard OR the task list is incomplete or currently shown
+		// or the current page is one of the WooCommerce Admin pages.
+		if (
+			( ! self::should_show_profiler() && ! self::should_show_tasks()
+			||
+			! self::is_woocommerce_admin_page()
+		)
+		) {
 			return $settings;
 		}
 
--- a/src/Features/Onboarding.php
+++ b/src/Features/Onboarding.php
@@ -645,6 +645,21 @@ class Onboarding {
 	}
 
 	/**
+	 * Determine if the current page is home or setup wizard.
+	 *
+	 * @return bool
+	 */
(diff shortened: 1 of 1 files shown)


src/Features/Onboarding.php
```diff
@@ -648,15 +647,28 @@ public static function delete_themes_transient() {
...
 	 */
 	public function component_settings( $settings ) {
 		$profile                = (array) get_option( self::PROFILE_DATA_OPTION, array() );
 		$settings['onboarding'] = array(
 			'profile' => $profile,
 		);
 
-		// Only fetch if the onboarding wizard OR the task list is incomplete or currently shown.
-		if ( ! self::should_show_profiler() && ! self::should_show_tasks() ) {
+		$is_wc_admin_page = false;
```
becdetat, line 659: Can you extract this out into a method like `should_show_profiler()` and `should_show_tasks()`? That would simplify this method a bit.

moon0326: @becdetat What do you think about extracting it out to a new method like `is_woocommerce_page()`. I checked both `should_show_profiler()` and `should_show_tasks()`, but I think the logic does not fit into the methods.

becdetat: @moon0326 oh yes, extracting it into a new method was what I meant :-) I could have explained that better

moon0326: @becdetat Oh haha now I read it again, it makes sense. Sorry about that 🤦

moon0326: @becdetat Updated in https://github.com/woocommerce/woocommerce-admin/commit/dfb2c2568ffac9078600c865bc88f8bfcb8b28c0

src/Features/Onboarding.php
```diff
@@ -507,9 +507,8 @@ public static function get_themes() {
 			$active_theme     = get_option( 'stylesheet' );
 
 			foreach ( $installed_themes as $slug => $theme ) {
-				$theme_data       = self::get_theme_data( $theme );
-				$installed_themes = wp_get_themes();
```
jeffstieler, line 511: 😱

src/Features/Onboarding.php
```diff
@@ -644,19 +643,42 @@ public static function delete_themes_transient() {
...
+	 */
+	public static function is_woocommerce_admin_page() {
+		$is_wc_admin_page = false;
+		$current_screen   = get_current_screen();
+		if ( null !== $current_screen ) {
+			$is_wc_admin_page = 'woocommerce_page_wc-admin' === $current_screen->id;
+		}
+
+		return $is_wc_admin_page;
+	}
```
jeffstieler, line 659: Could [`PageController::get_current_page()`](https://github.com/woocommerce/woocommerce-admin/blob/cfeb5a163561ff98343e4da0d56e768a454ccac5/src%2FPageController.php#L201) work here instead?

src/Features/Onboarding.php
moon0326: 😄

src/Features/Onboarding.php
moon0326: @jeffstieler Thank you for the suggestion 👍  Yup, I think `PageController::get_current_page()` is more reliable for the purpose. Change is in https://github.com/woocommerce/woocommerce-admin/pull/5771/commits/d6e1ad18f3533ef4b70cd79d30045dbc62d4f253

moon0326: @jeffstieler Updated logic to include the setting only if you are on WooCommerce -> Home or setup wizard page per Timmy's [suggestion](https://github.com/woocommerce/woocommerce-admin/issues/5702#issuecomment-733213233)

octaedro (approved):
Good job @moon0326! The code LGTM and is testing well on my end 🚀

PR #5772, opened 2020-12-01
Update README.md and CONTRIBUTING.md

ilyasfoo:
Improved our docs from my onboarding experience, in the hope to reduce frictions for new contributors.

I've added a `Common Issues` section in contrib docs, which aims to be a comprehensive list of known issues with solution or workaround that is not so obvious to discover. Although I'm not sure if this is better as a different document altogether.

Please let me know what you think!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,13 +54,14 @@ After you've made your updates, you're ready to commit:
 
 ## PHP Unit tests
 
-### Setting up PHP unit tests using VVV
+### Setting up PHP unit tests using [VVV](https://github.com/Varying-Vagrant-Vagrants/VVV)
 
 1. SSH into the Vagrant box:
     1. `cd` down to the Vagrant root (where `www` lives) 
     2. `vagrant ssh`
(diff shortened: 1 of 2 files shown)


adrianduffell:
Thanks @ilyasfoo! I think the addition of the Common Issues documentation is going to be helpful.

> Although I'm not sure if this is better as a different document altogether.

I wonder if the content would better in the GitHub wiki, but with a link to it from the README? I'm thinking the wiki would be simpler to keep up to date.

becdetat:
@ilyasfoo 

> I wonder if the content would better in the GitHub wiki

I agree with @adrianduffell. Our wiki could use some love, and this  seems like something that can and should be added to quite quickly. I've dodged the documentation issue so far by keeping a paper notebook with the tips and tricks I've picked up, which isn't ideal for sharing.

becdetat:
Nice job on this :-D

ilyasfoo:
Thank you for your insights @becdetat @adrianduffell! I totally agree and made a wiki [here](https://github.com/woocommerce/woocommerce-admin/wiki/Common-Issues)

adrianduffell (approved):
Thanks @ilyasfoo this looks good! 👍 🚀

PR #5773, opened 2020-12-01
Navigation: Introduce wc-admin registration client side navArgs

psealock:
While reviewing https://github.com/Automattic/woocommerce-payments/pull/1067 I noticed a problem with using the `path` as an id: When designating a `parent`, you'd have to supply the correctly computed `path` parameter of the parent or wc-admin would need to calculate that on behalf of the extension. Consider the following child item:

```php
			wc_admin_register_page(
				[
					'id'       => 'wc-payments-transactions',
					'title'    => __( 'Transactions', 'woocommerce-payments' ),
					'parent'   => 'wc-payments',
					'path'     => '/payments/transactions',
					'nav_args' => [
						'parent' => 'wc-payments',
						'order'  => 20,
					],
				]
			);
```
The parent would actually need to be `wc_admin-wc-admin&path=/payments/deposits`. So I was thinking that since we need to ask wc-admin extensions to supply `nav_args` in PHP, why not be explicit and ask them to also supply `navArgs` on the client?

This PR makes that change to filters `woocommerce_admin_pages_list` and `woocommerce_admin_reports_list` to accept `navArgs`. Similar to PHP, slotFill does not function if `navArgs` are not supplied.

Same as https://github.com/woocommerce/woocommerce-admin/pull/5665

1. Ensure all wc-admin pages can navigate without refreshing the page, including Home, Settings > Analytics, and Customers.
2. Ensure Settings > Products is a functioning Nav item. 

### Note

The base is `update/simplify-wca-nav-registration` for easy testing of https://github.com/Automattic/woocommerce-payments/pull/1067

### DIFF
--- a/client/analytics/report/README.md
+++ b/client/analytics/report/README.md
@@ -40,17 +40,21 @@ addFilter(
 				report: 'example',
 				title: 'My Example Extension',
 				component: Report,
+				navArgs: {
+					id: 'my-example-report',
+				},
 			},
 		];
 	}
 );
 ```
 
-Each report is defined by an object containing `report`, `title`, `component`.
+Each report is defined by an object containing `report`, `title`, `component`, and optionally `navArgs`.
 
 -   `report` (string): The path used to show the report, ex: `/analytics/example`
 -   `title` (string): The title shown in the breadcrumbs & document title.
 -   `component` (react component): The component containing the report content- everything on the page under the breadcrumbs header.
+-   `navArgs` (object): Supply an id which is the same id as the server side registration.
 
 The component will get the following props:
 
--- a/client/analytics/report/README.md
+++ b/client/analytics/report/README.md
@@ -54,7 +54,7 @@ Each report is defined by an object containing `report`, `title`, `component`, a
 -   `report` (string): The path used to show the report, ex: `/analytics/example`
 -   `title` (string): The title shown in the breadcrumbs & document title.
 -   `component` (react component): The component containing the report content- everything on the page under the breadcrumbs header.
--   `navArgs` (object): Supply an id which is the same id as the server side registration.
+-   `navArgs` (object): Arguments used for the new navigation experience, typically used to supply a matching ID for server-side registered menu items.
 
 The component will get the following props:
 
--- a/client/analytics/report/get-reports.js
+++ b/client/analytics/report/get-reports.js
@@ -57,42 +57,66 @@ export default () => {
 			report: 'revenue',
 			title: __( 'Revenue', 'woocommerce-admin' ),
 			component: RevenueReport,
+			navArgs: {
+				id: 'woocommerce-analytics-revenue',
+			},
 		},
 		{
 			report: 'products',
 			title: __( 'Products', 'woocommerce-admin' ),
 			component: ProductsReport,
+			navArgs: {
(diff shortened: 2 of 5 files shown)


client/analytics/report/get-reports.js
```diff
@@ -57,54 +57,84 @@ export default () => {
...
+					},
 			  }
 			: null,
 		{
 			report: 'customers',
 			title: __( 'Customers', 'woocommerce-admin' ),
 			component: CustomersReport,
 			id: null,
+			navArgs: {
+				id: 'woocommerce-analytics-customers',
```
joelclimbsthings, line 128: Looks like we've got a small regression here resulting in the duplicate `Customers` entry in the nav again. Here's the original commit that fixed this by Josh: https://github.com/woocommerce/woocommerce-admin/commit/96522affd3dcf33290c01a5b69b2966eda4b995b

client/layout/navigation.js
```diff
@@ -57,24 +55,28 @@ const NavigationPlugin = () => {
...
-						<Link
-							className="components-button"
-							href={ getNewPath(
-								persistedQuery,
-								`/analytics/${ item.report }`,
-								{}
-							) }
-							type="wc-admin"
+			{ reports
+				.filter( ( item ) => item.navArgs )
```
joelclimbsthings, line 59: Perhaps we should put this `.filter()` up above just to stay consistent with how it's done with `pages`?

joelclimbsthings (changes requested):
Looks good overall Paul! I feel pretty good about this approach give the troublesome nature of using `path`. 

I left one comment regarding a small regression, and in addition to that I noticed that there is another duplicate `Customers` entry under `Settings > Customers`. The one under Settings is actually marked as the active entry when you go to Customers via the nav:

image

psealock:
Thanks for the review @joelclimbsthings. I added your suggestions on filtering earlier and fixed up the duplicate Customers entries.

The issue you noted of Customers also appearing in Settings is also on `main` branch, so I opened a ticket https://github.com/woocommerce/woocommerce-admin/issues/5782

joelclimbsthings (approved):
Thanks Paul, tests well and LGTM! :ship:

client/analytics/report/README.md
```diff
@@ -40,17 +40,21 @@ addFilter(
...
 );
 ```
 
-Each report is defined by an object containing `report`, `title`, `component`.
+Each report is defined by an object containing `report`, `title`, `component`, and optionally `navArgs`.
 
 -   `report` (string): The path used to show the report, ex: `/analytics/example`
 -   `title` (string): The title shown in the breadcrumbs & document title.
 -   `component` (react component): The component containing the report content- everything on the page under the breadcrumbs header.
+-   `navArgs` (object): Supply an id which is the same id as the server side registration.
```
joshuatf, line 57: Optional nitpick: we could make this comment more specific since it's in the context of registering pages.

```suggestion
-   `navArgs` (object): Arguments used for the new navigation experience, typically used to supply a matching ID for server-side registered menu items.
```

joshuatf (approved):
Great work, Paul.  After taking another look at this code, I do think this feels much more straightforward and less magic.  LGTM pending one optional comment change.

PR #5786, opened 2020-12-02
A/B test creating a business features step for free extensions

### LINKED ISSUE #5287: Store profiler: create a business features step for free extensions
**Description:** the experiment implemented in #4579 showed that we're moving in the right direction. However, there are some concerns over not allowing users to disable some extensions and displaying them in the Business Details step if far from ideal.
We're going to experiment with an intermediate approach while keeping the two existing flows. We aim to validate this approach in the WooCommerce December release and if everything looks good we'll remove the two other flows.

Final segmentation: Stores that selected the `US` as the country and `Other`or `Food & Drinks` as the industry. It gives about 30% of the total number of new users from the US

**Acceptance Criteria:**
- [ ] display the changes in the flow just to a specific segment: Stores that selected the `US` as the country and `Other`or `Food & Drinks` as the industry
- [ ] introduce a new step after the business details called "Business features"
- [ ] display just one checkbox to install all the free extensions. The checkbox should be toggled on by default
- [ ] introduce an accordion where the users can see more details about each extension and disable some of them
- [ ] allow users to install the following extensions: Jetpack, WooCommerce Services, Facebook, Mailchimp, Google Ads, WCPay 
- [ ] install and activate the selected extensions when the user clicks "Continue"
- [ ] remove the JP/WCS step for this segment, as they're going to be installed in the Business details step
- [ ] the redirect to the Jetpack account creation flow should only happen after the theme step 

**Event tracking:**
It will be better to add a new event for this step's alternative version:

Event name: `wcadmin_storeprofiler_store_business_features_continue`
Event description: When the user clicks on "Continue"
Event prop: `all_extensions_installed` (value: True/False) - with all the extensions selected or deselected 
Then we should also have a prop for each extension to cover when they deselect some of them:
`install_woocommerce_services`	
`install_mailchimp`
`install_jetpack`
`install_google_ads`
`install_facebook`
`install_wcpay`
 
Event name: `wcadmin_storeprofiler_store_business_features_accordion_click`
Event prop: NA
Event description: When the user clicks on the accordion

It will be necessary to update `wcadmin_storeprofiler_step_view` to include this alternative step as the value of the `step` prop. It can be named business-features

**Additional links:**
P2 w/ link to the mockups here: pbIJXs-n0-p2

samueljseay:
This fixes #5287 

Major changes:

1. The `<Form>` component (which is a published component) gets a new optional prop (so its backward compatible), called `onChangeCallback`. This is called when any form value changes, allowing the controlling component to act on individual form changes. This was needed to utilise `<Form>` to revalidate and check if the user could access the next tab. This should not impact existing usages of the `<Form>`, but it could be good to test this.

2. Introduces a new flow for choosing business extensions that was specified in #5287 issue description. Please check the issue for reviewing the designs. The simplest way to implement this, ended up being keeping the existing flow intact and where necessary copy pasting code from those components into the new flow. This new flow is only shown if your segmentation matches the following: `Stores that selected the US as the country and Other or Food & Drinks as the industry`

screen1


client/profile-wizard/index.js
```diff
@@ -134,8 +140,10 @@ class ProfileWizard extends Component {
 				profileItems.product_types !== null,
 		} );
 		steps.push( {
-			key: 'business-details',
-			container: BusinessDetails,
+			key: selectiveBundleInstallSegmentation
```
samueljseay, line 143: The reason we have to check the segmentation this high up, is because there is analytics tracking that happens as you progress through the wizard steps, we have to change the step name conditionally here to make sure the tracked event is given the right name. This is a result of the design of the existing code, I don't think theres a better way to do this without a significant refactor.

client/profile-wizard/steps/business-details/data/extension-benefits.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+export const extensionBenefits = [
```
samueljseay, line 6: There was a lot of static data inside the original business details component, which made it hard to read the component (it was close to 1k lines) but also we need some of this data in the new flow as well, so I've moved them into their own folder.

client/profile-wizard/steps/business-details/flows/bundle/index.js
```diff
@@ -963,7 +783,7 @@ class BusinessDetails extends Component {
 
 BusinessDetails.contextType = CurrencyContext;
 
-export default compose(
+export const BundleBusinessDetailsStep = compose(
```
samueljseay, line 786: This is the original step, but I've just moved all the static data out of it.

client/profile-wizard/steps/business-details/flows/selective-bundle/selective-extensions-bundle/index.js
```diff
@@ -0,0 +1,350 @@
...
+					target="_blank"
+					className="woocommerce-admin__business-details__selective-extensions-bundle__link"
+					href={ `https://woocommerce.com/products/${ productName }` }
+				/>
+			),
+		},
+	} );
+};
+
+const installableExtensions = [
```
samueljseay, line 40: I considered moving this to the data folder, but its very specifically tied to the UI rendering of the groups of checkboxes so I've left it here for now.

client/profile-wizard/steps/business-details/flows/selective-bundle/selective-extensions-bundle/index.js
```diff
@@ -0,0 +1,350 @@
...
+						'woocommerce-admin'
+					),
+					'creative-mail-for-woocommerce'
+				),
+			},
+		],
+	},
+];
+
+const initialValues = installableExtensions.reduce(
```
samueljseay, line 123: The initial form values are derived from the available extensions, this is just more flexible and going to save some hassle for the next person to add an extension to the list. They can just add items to installable extensions.

client/profile-wizard/steps/business-details/flows/selective-bundle/selective-extensions-bundle/index.js
```diff
@@ -0,0 +1,350 @@
...
+const FreeBadge = () => {
+	return (
+		<div className="woocommerce-admin__business-details__free-badge">
+			{ __( 'Free', 'woocommerce-admin' ) }
+		</div>
+	);
+};
+
+// Set all props of an object to a value
+const setAllPropsToValue = ( obj, value ) => {
```
samueljseay, line 146: TODO - I'll move this to `lib`. Its generic and useful.

client/profile-wizard/steps/business-details/flows/selective-bundle/selective-extensions-bundle/index.js
```diff
@@ -0,0 +1,350 @@
...
+										<FreeBadge />
+									</div>
+								) ) }
+							</div>
+						) ) }
+				</div>
+				<div className="woocommerce-profile-wizard__business-details__free-features__action">
+					<Button
+						onClick={ () => {
+							onSubmit( values );
```
samueljseay, line 335: the values from the form are passed back up to the parent component by `onSubmit`

client/profile-wizard/steps/business-details/index.js
```diff
@@ -0,0 +1,71 @@
...
+
+	if ( isLoading ) {
+		return (
+			<div className="woocommerce-admin__business-details__spinner">
+				<Spinner />
+			</div>
+		);
+	}
+
+	if ( selectiveBundleInstallSegmentation ) {
```
samueljseay, line 53: segmentation determines if we render the new step I've introduced. But there is existing segmentation for another experiment inside `< BundleBusinessDetailsStep>` that will determine what is rendered there. It was far simpler to leave that segmentation in place than try to bring it here. I hope in future this will all get consolidated to a single step or we will have a proper A/B test framework in place.

packages/components/src/form/index.js
```diff
@@ -33,18 +32,28 @@ class Form extends Component {
...
 		this.setState(
 			( prevState ) => ( {
 				values: { ...prevState.values, [ name ]: value },
 			} ),
-			this.validate
+			() => {
+				this.validate( () => {
+					// onChangeCallback keeps track of validity, so needs to
+					// happen after setting the error state.
+					this.props.onChangeCallback(
```
samueljseay, line 50: This is a handy new callback on form that lets you react to individual value changes and check validity per value change rather than just on submit.

tests/e2e/specs/activate-and-setup/complete-industry-section.js
```diff
@@ -17,28 +22,29 @@ export async function completeIndustrySection( expectedIndustryCount = 8 ) {
 		expectedIndustryCount
 	);
 
-	const industryCheckboxes = await page.$$(
-		'.components-checkbox-control__input'
+	// Select just "fashion" and "health/beauty" to get the single checkbox business section when
```
samueljseay, line 25: Adjusted the e2e tests so that they still work, but I haven't introduced new e2e tests for the new segment yet. I'm going to explore that as I add tests.

tests/e2e/specs/activate-and-setup/complete-industry-section.js
```diff
@@ -17,28 +22,29 @@ export async function completeIndustrySection( expectedIndustryCount = 8 ) {
...
+	// get sibling checkbox
+	const fashionCheckboxContainer = await getPreviousSibling( fashionLabel );
+	const fashionCheckbox = await fashionCheckboxContainer.$( 'input' );
 
-	const industryLabels = await page.$$(
-		'.components-checkbox-control__label'
-	);
-
-	// Select all industries except for CBD to fulfill conditions required by business section tests.
-	for ( let i = 0; i < expectedIndustryCount; i++ ) {
```
samueljseay, line 29: Before we just filled out all the check boxes, I've introduced some helpers to make it easier to fill out a specified checkbox which allows us to control which segmentation happens.

client/lib/collections/index.js
```diff
@@ -26,3 +26,18 @@ export const groupListOfObjectsBy = (
...
 	return {};
 };
+
+/**
+ * Returns a (shallow) copy of an object with all its props set to the specified value
+ *
+ * @param {*} obj the Object to copy.
+ * @param {*} value the value to set all props on the object to.
+ */
+export const setAllPropsToValue = ( obj, value ) => {
```
samueljseay, line 36: renamed this file to collections because lists is a little specific. Added a utility which I needed for this feature.

jeffstieler (changes requested):
From the issue:
> remove the JP/WCS step for this segment, as they're going to be installed in the Business details step

In my testing, I selected WooCommerce Payments and Jetpack to be installed (but not WooCommerce Shipping) and I was shown the Woo+JP Shipping prompt after the theme step.

Screen Shot 2020-12-11 at 9 06 48 AM

I clicked "no thanks" and landed on the Home screen. I was not redirected to the Jetpack connection flow.

samueljseay:
> remove the JP/WCS step for this segment, as they're going to be installed in the Business details step

Damn, I knew I would miss something. Checked over those requirements so many times. Thanks for checking the issue closely.

samueljseay:
@jeffstieler I've changed the logic now, if the user goes through the segmentation for this new experiment they will not be shown that last plugin install step. as per:

> remove the JP/WCS step for this segment, as they're going to be installed in the Business details step

client/profile-wizard/steps/business-details/flows/selective-bundle/selective-extensions-bundle/index.js
```diff
@@ -0,0 +1,341 @@
...
+						<p className="woocommerce-admin__business-details__selective-extensions-bundle__description">
+							{ __(
+								'Add recommended business features to my site'
+							) }
+						</p>
+						<Icon
+							className="woocommerce-admin__business-details__selective-extensions-bundle__expand"
+							icon={ showExtensions ? chevronUp : chevronDown }
+							onClick={ () => {
+								const shouldShowExtensions = ! showExtensions;
```
becdetat, line 259: The naming is making the logic here a bit unclear IMO

client/profile-wizard/steps/business-details/flows/selective-bundle/selective-extensions-bundle/index.js
```diff
@@ -0,0 +1,341 @@
...
+												const allExtensionsDisabled =
+													Object.entries(
+														newState
+													).filter(
+														( [ , val ] ) => val
+													).length === 1 &&
+													newState.install_extensions;
+
+												if ( allExtensionsDisabled ) {
+													// If all the extensions are disabled then disable the "Install Extensions" checkbox too
```
becdetat, line 300: I clearly need a wider monitor...

client/profile-wizard/steps/business-details/index.js
```diff
@@ -0,0 +1,71 @@
...
+
+	const country = settings.general
+		? settings.general.woocommerce_default_country
+		: null;
+
+	const industrySlugs = ( profileItems.industry || [] ).map(
+		( industry ) => industry.slug
+	);
+
+	const selectiveBundleInstallSegmentation =
```
becdetat, line 40: This logic is duplicated in `client/profile-wizard/index.js`, could it be consolidated somewhere?

becdetat:
Everything looks really good and it tests well and works great. I won't approve it until the e2e tests are done (unless you want to do them in a seperate PR?).

client/profile-wizard/steps/business-details/flows/selective-bundle/selective-extensions-bundle/index.js
samueljseay: I blame all the spacing in our code standard :P

client/profile-wizard/steps/business-details/flows/selective-bundle/selective-extensions-bundle/index.js
samueljseay: @becdetat Yeah not great naming I agree. I'm blanking on something better. `newState`? or `toggleShowExtensions`?

client/profile-wizard/steps/business-details/flows/selective-bundle/selective-extensions-bundle/index.js
samueljseay: @becdetat Seriously though, this is something that I always go back and forth on. its easier for the reader for the code to be in place on the control to figure out "whats going on", than say storing the callback somewhere else and forcing them to navigate there to figure out whats going on.

Genuinely our coding standard harms this approach with tabs though... If you have a suggestion of what you would prefer its easy enough for me to make the change.

client/profile-wizard/steps/business-details/index.js
samueljseay: Yep thats a good call, I'll do that.

samueljseay:
@becdetat I will do the tests as a separate PR but I'll do it into this one so an approval can wait, I was just hoping to get some early review to catch things, thanks so much for reviewing this 🙇

client/profile-wizard/steps/business-details/index.js
samueljseay: Done in 75c1f89 thanks for the suggestion!

client/profile-wizard/steps/business-details/flows/selective-bundle/selective-extensions-bundle/index.js
becdetat: Every time I have a `.map` I try to break the contents down into a smaller function. So you would have:

```
installableExtensions.map( { plugins, title } => getInstallableExtension( plugins, title ) )
```

and then in `getInstallableExtension` you've got a `plugins.map` so that would be broken down into another smaller method. Yes you need to navigate to find the code for the mapping but if you can name the map functions clearly that isn't usually a big deal as you've got small functions that don't do as much.

client/profile-wizard/steps/business-details/flows/selective-bundle/selective-extensions-bundle/index.js
becdetat: Honestly you could just use `! showExtensions` twice and it would be clearer, rather than using an intermediate.

samueljseay:
@becdetat I ended up adding a basic e2e test to this in ee200d5

client/profile-wizard/steps/business-details/flows/selective-bundle/selective-extensions-bundle/index.js
samueljseay: ok sure thing.

client/profile-wizard/steps/business-details/flows/selective-bundle/selective-extensions-bundle/index.js
samueljseay: Thats fair, I can move this into a callback or a separate component.

client/profile-wizard/steps/business-details/flows/selective-bundle/selective-extensions-bundle/index.js
samueljseay: I'm not convinced its clearer, the intermediate is to indicate that we've toggled the boolean value. Anyway, I've made the change here: 4fd2d417dfe58c75f7ed2723699fd0b0aafe959b

client/profile-wizard/steps/business-details/flows/selective-bundle/selective-extensions-bundle/index.js
samueljseay: One thing I would say is that this doesn't look crazy in VSCode or require a wide screen. Its just that github renders very wide tabs. However I think it was an improvement to move this rendering into a separate component. I moved the onChange callback out as well, so hopefully thats more readable: bd57378

becdetat (approved):
I think the changes are good, and everything still works great. Thanks :-)

PR #5787, opened 2020-12-02
Use client side filter on the CSV download

### LINKED ISSUE #5202: How to force csv report by email?
Currently, if you click on the download button, you can get the csv report generated by javascript, if the amount of items is small or by email if the report table is bigger than the number of rows per page.

The problem is, when creating custom columns, I can only get them to be displayed on the email csv, not by javascript.
So, how can I force the csv to be sent by e-mail always regardless of the report table size?

becdetat:
Fixes #5202

Kinda.

This takes the same client-side filter that currently allows changing the columns in the report table (`woocommerce_admin_report_table`) and uses it when generating the CSV on the client-side. So the CSV is in sync with the report table.

To test this: follow the instructions in this blog post: https://becdetat.home.blog/2020/12/02/adding-columns-to-analytics-reports-and-csv-downloads/

### Changelog Note:

Dev: Use woocommerce_admin_report_table client-side filter when generating a CSV

### DIFF
--- a/client/analytics/components/report-table/index.js
+++ b/client/analytics/components/report-table/index.js
@@ -152,6 +152,32 @@ const ReportTable = ( props ) => {
 		} ) );
 	};
 
+	const applyTableFilters = ( data, totals, totalResults ) => {
+		const summary = getSummary ? getSummary( totals, totalResults ) : null;
+
+		/**
+		 * Filter report table for the CSV download.
+		 *
+		 * Enables manipulation of data used to create the report CSV.
+		 *
+		 * @param {Object} reportTableData - data used to create the table.
+		 * @param {string} reportTableData.endpoint - table api endpoint.
+		 * @param {Array} reportTableData.headers - table headers data.
+		 * @param {Array} reportTableData.rows - table rows data.
+		 * @param {Object} reportTableData.totals - total aggregates for request.
+		 * @param {Array} reportTableData.summary - summary numbers data.
+		 * @param {Object} reportTableData.items - response from api requerst.
+		 */
+		return applyFilters( TABLE_FILTER, {
+			endpoint,
+			headers: getHeadersContent(),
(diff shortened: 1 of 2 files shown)


client/analytics/components/report-table/index.js
```diff
@@ -165,12 +165,31 @@ const ReportTable = ( props ) => {
...
+			 * @param {Object} reportTableData.items - response from api requerst.
+			 */
+			const { headers, rows } = applyFilters( TABLE_FILTER, {
+				endpoint,
+				headers: getHeadersContent(),
+				rows: getRowsContent( data ),
+				totals,
+				summary: getSummary ? getSummary( totals, totalResults ) : null,
+				items,
+			} );
```
jeffstieler, line 188: Rather than duplicate this code block, what do you think about moving it to a reusable function?

client/analytics/report/downloads/table.js
```diff
@@ -19,7 +19,7 @@ import { getCurrentDates, defaultTableDateFormat } from '@woocommerce/date';
 import ReportTable from '../../components/report-table';
 import { CurrencyContext } from '../../../lib/currency-context';
 
-class CouponsReportTable extends Component {
+class DownloadsReportTable extends Component {
```
jeffstieler, line 22: Nice catch.

client/analytics/components/report-table/index.js
becdetat: I was pondering this while drifting off to sleep last night, wondering what reason I had for duplicating it. Will do :-D

becdetat: @jeffstieler Done in 320b5e80, thanks

louwie17 (approved):
Nice work @becdetat, and thanks for the refactor that Jeff suggested, it works well :)

PR #5796, opened 2020-12-02
CES: Remove CSS transition from CES modal when comments appear

mattsherman:
@elizaan36 has requested (pbIJXs-m0-p2) that the transition when showing/hiding the comments field on the CES modal be removed for now, since we do not have a holistic set of transitions/animations for our design system. This can be revisited in the future. 

- delete the `woocommerce_ces_shown_for_actions` option to reset the feature if you've tested it before
- Add a product. When the snackbar appears, click 'Give Feedback'.
- Select "Somewhat difficult".
- Verify that when the comments field is shown/hidden on the CES modal, that it instantly appears/disappears (no transition/animation) 

Fix: Remove CES modal comments transition.

### DIFF
--- a/packages/customer-effort-score/src/customer-feedback-modal/index.js
+++ b/packages/customer-effort-score/src/customer-feedback-modal/index.js
@@ -11,8 +11,6 @@ import {
 	TextareaControl,
 } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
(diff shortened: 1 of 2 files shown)


joelclimbsthings (approved):
Sweet, sweet code deletion. Well done & tests well :shipit:

elizaan36 (commented):
Thanks @mattsherman. Looks good to me!

Issue #5710, opened 2020-11-20
Nav disappeared after choosing "Set up manually" in tax task


elizaan36:
**Describe the bug**
I was testing the tax task today to update some styles and noticed that when I clicked through after selecting "Set up manually", I was redirected to the correct area **Settings > Tax > Standard rates** but the nav disappeared. 

It reappeared upon refreshing the page, so I'm not sure it's a huge issue but want to share the issue to make sure it's not something that happens when installing extensions if the top level nav item isn't created right away.

**To Reproduce**
Open new WooCommerce site with task list on Home screen.

1. Go to 'Set up tax'
2. Click on 'Set up manually'
3. See nav disappear

**Expected behavior**
The navigation should appear and be located on the **Settings > Tax** item.

**Screenshots**
![set up manually and nav disappears](https://user-images.githubusercontent.com/5121465/99849776-c735ea00-2b41-11eb-8bc9-f84ed733f8a2.gif)

elizaan36:
Also wanted to add that refreshing the page while on the tax task jumps to **Analytics > Stock**. I'd expect to stay on the Home screen.

![jump to analytics](https://user-images.githubusercontent.com/5121465/100001484-d9ee3000-2d88-11eb-9550-321a7dbc7be6.gif)

joshuatf:
@elizaan36 I was unable to reproduce on the latest dev branch.  It's very odd that it appeared on refresh since the tax page is not React-based so the original load should have triggered a full page reload.

Can you open your console and repeat this and take screenshots of any errors that pop up?

elizaan36:
hey @joshuatf I'm not seeing any errors but was able to reproduce the jump to *Analytics > Stock*. https://share.getcloudapp.com/BluQgQov

Looks like the nav doesn't disappear anymore after navigating to "Set up manually", so that's resolved itself on my site at least.

joshuatf:
I do see the active status being displayed wrong in a few places (though I'm seeing it work correctly for taxes).  When we're changing routes, often times there's not an exact match in the menu.

For context, this currently this will mark items active via match score (created by how many params in the URL match a given item's URL) to the final item matched (meaning if there's a match score of `5` for multiple items, it will mark the latter one as active).

[...]

PR #5798, opened 2020-12-02
Add navigation determinant params

joshuatf:
Part 1 of a 2 part exploration to improve route detection.

This PR allows `queryParams` (up for discussion- `requiredParams`?) to be passed along with each individual item to help determine if this item should be marked active or not.  For example passing `'queryParams' => array( 'my_param' )` as a menu item argument means that the current URL location must match the menu item's `my_param` option.

This is useful since some pages are determined active by the `page` param, while others, like those in WCA, are determined by `page` and `path` and those in WC Settings are determined by `tab`.

The follow-up PR will explore a regex alternative.

### Screenshots
Screen Shot 2020-12-02 at 5 39 05 PM

1. Add some items that have possible conflict in params due to a matching score being higher.  For example, you could register a new WCA page:

```php
wc_admin_register_page(
	array(
		'id'     => 'test-determinant-params',
		'title'  => __( 'Test Determinant Params', 'woocommerce-admin' ),
		'path'   => '/test-params&param1=a&param2=b',
		'nav_args' => array()
	)
);
```

2. Navigate to a URL where the match score should be higher for that item, but using a non-matching determinant param (`path` in the case of WCA).  `wp-admin/admin.php?page=wc-admin&param1=a&param2=b`
3. Note that the correct item is matched.

### DIFF
--- a/client/navigation/utils.js
+++ b/client/navigation/utils.js
@@ -46,23 +46,14 @@ export const getFullUrl = ( url ) => {
  *
  * @param {Object} location Window location
  * @param {string} url URL to compare
- * @param {Array} itemDeterminantParams Params required to match this item.
+ * @param {Array} queryParams Params required to match this item.
  * @return {number} Number of matches or 0 if not matched.
  */
-export const getMatchScore = (
-	location,
-	url,
-	itemDeterminantParams = null
-) => {
+export const getMatchScore = ( location, url, queryParams = [] ) => {
 	if ( ! url ) {
 		return;
 	}
 
-	const determinantParams = itemDeterminantParams || [
-		'page',
-		'tab',
-		'path',
-	];
 	const fullUrl = getFullUrl( url );
 	const urlLocation = new URL( fullUrl );
 	const { origin: urlOrigin, pathname: urlPathname } = urlLocation;
@@ -96,7 +87,7 @@ export const getMatchScore = (
 		if ( urlParams[ key ] === locationParams[ key ] ) {
 			matchingParamCount++;
 		} else if (
-			determinantParams.includes( key ) &&
+			queryParams.includes( key ) &&
 			( urlParams[ key ] || locationParams[ key ] )
 		) {
 			// A determinant param was found and not matched.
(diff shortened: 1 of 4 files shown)


client/navigation/utils.js
```diff
@@ -82,11 +83,17 @@ export const getMatchScore = ( location, url ) => {
...
-		if ( urlParams[ key ] === locationParams[ key ] ) {
+	for ( const key in itemParams ) {
+		if ( itemParams[ key ] === locationParams[ key ] ) {
 			matchingParamCount++;
+		} else if (
+			queryParams.includes( key ) &&
+			( itemParams[ key ] || locationParams[ key ] )
+		) {
+			// A determinant param was found and not matched.
+			return 0;
```
psealock, line 94: Nice 👍

psealock (commented):
Good start @joshuatf, this is working well and the case encountered in https://github.com/woocommerce/woocommerce-admin/issues/5710 is resolved.

>up for discussion- requiredParams

I think so. Its more explicit. I think there could be a way for us to calculate these on behalf of extensions. Can we have a list of params that are added automatically, like `page` and the rest are derived from the supplied url? That list of params could be filterable too.

It feels like this would be one less step for extensions

joshuatf:
>Can we have a list of params that are added automatically, like page and the rest are derived from the supplied url? That list of params could be filterable too.

Great idea.  I was trying to avoid this since even though `page` is used to determine pages in core, I didn't want to impose any restrictions.  But if we create a filterable list of default args, then we get the best of both worlds.

psealock:
I took this for a spin to compare the regex approach. I register a page with a url including a query parameter.

```php
function page_content(){
	echo '<div class="wrap"><h2>Testing</h2></div>';
}

add_menu_page( 'My Extension', 'My Extension', 'manage_woocommerce', 'my-extension-slug', 'page_content' );


Menu::add_plugin_item(
	array(
		'id'         => 'my-extension',
		'title'      => __( 'My Extension', 'my-extension' ),
		'capability'    => 'manage_woocommerce',
		'url'        => 'my-extension-slug&tab=new',
		'parent' => 'woocommerce',
		'queryParams' => array( 'page', 'tab' ),
	)
);
```
When the `queryParams` are supplied, this behaves well. Inclusion of the `page`, and `tab` parameters are the only thing required for a correct match, regardless of order. As mentioned, these parameters don't need to be supplied, but can be derived from the `url` parameter plus a filterable list of definites such as `page`. 

This approach is my vote. We get the desired result without relying on extensions to craft difficult regex or maintain our own default regex that supports variable query parameter order.

joshuatf:
Closing this in favor of https://github.com/woocommerce/woocommerce-admin/pull/5818

PR #5799, opened 2020-12-02
Use callback for customers URL in navigation

### LINKED ISSUE #5782: Navigation: Customers regression
Customers is now under Settings as well as the main menu. Only the one under Settings has a working link.

The extra one from Settings seems to be added here
https://github.com/woocommerce/woocommerce-admin/blob/37fc01e0640dbf65684d4990976cdf4d5b9774ab/src/Features/Navigation/Menu.php#L366

Screen Shot 2020-12-02 at 2 46 13 PM

Screen Shot 2020-12-02 at 2 46 22 PM

joshuatf:
Fixes #5782

Uses the callback for the customers URL to avoid double registration.

1. Enable the nav feature.
1. Check that `Customers` works from the main menu.
1. Check that `Customers` does not exist under the `Settings` menu.

### DIFF
--- a/src/Features/Navigation/CoreMenu.php
+++ b/src/Features/Navigation/CoreMenu.php
@@ -168,7 +168,7 @@ class CoreMenu {
 			$customers_item = array(
 				'id'    => 'woocommerce-analytics-customers',
 				'title' => __( 'Customers', 'woocommerce-admin' ),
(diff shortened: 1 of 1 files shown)


psealock (approved):
Nice fix @joshuatf, working well and code looks good.

PR #5800, opened 2020-12-02
Use an empty queue if CES feature is not enabled

### LINKED ISSUE #5797: CES: Analytics filtering CES is always enabled, even if feature flag is not enabled
**Describe the bug**

Even if the `customer-effort-score-tracks` feature is not enabled for the `woocommerce-admin` plugin, the CES functionality in Analytics filtering is enabled.

**To Reproduce**

1. Create a test ZIP of `woocommerce-admin` with the `customer-effort-score-tracks` feature disabled (this will happen by default when you run `npm run test:zip`).
2. Upload the test ZIP and activate the plugin.
3. Verify that CES does not work in Products, Orders, or Settings.
4. Verify that CES works in Analytics.

**Expected behavior**

CES should not work in Analytics if the `customer-effort-score-tracks` feature is not enabled.

moon0326:
Fixes #5797 

This PR respects the feature flag for the CES on the Analytics pages.

I considered two approaches at first.

1.  Check the feature flag in https://github.com/woocommerce/woocommerce-admin/blob/a0187bc0adbd0c8428475292c4466a8ad810229d/client/index.js#L86 and conditionally load `CustomerEffortScoreTracksContainer.`
2. Check the feature flag in [customer-effort-score-tracks-container.js](https://github.com/woocommerce/woocommerce-admin/blob/main/client/customer-effort-score-tracks/customer-effort-score-tracks-container.js) and return an empty array for the `queue` value so that any subsequent `ADD_CES_SURVEY` calls would be ignored.

~~I took approach #2 since it's simple and API consumers don't have to check for the feature flag. If we choose approach #1, which does not include `CustomerEffortScoreTracksContainer` then the API consumers must check for the feature flag before calling the `ADD_CES_SURVEY` action.~~

Updated to use the first approach and left a [comment](https://github.com/woocommerce/woocommerce-admin/pull/5800#issuecomment-737672828) about it.

I would appreciate any suggestions on other approaches.

1. Turn off `customer-effort-score-tracks` in `config/development.json.`
2. Run `npm run dev` to apply the change.
3. Navigate to WooCommerce -> Analytics and change a filter.
4. You should not see a CES survey snackbar.

### DIFF
--- a/client/customer-effort-score-tracks/customer-effort-score-tracks-container.js
+++ b/client/customer-effort-score-tracks/customer-effort-score-tracks-container.js
@@ -75,6 +75,12 @@ CustomerEffortScoreTracksContainer.propTypes = {
 
 export default compose(
 	withSelect( ( select ) => {
+		if (
+			window.wcAdminFeatures &&
+			window.wcAdminFeatures[ 'customer-effort-score-tracks' ] !== true
+		) {
+			return { queue: [], resolving: false };
+		}
 		const { getCesSurveyQueue, isResolving } = select( STORE_KEY );
 		const queue = getCesSurveyQueue();
 		const resolving = isResolving( 'getOption', [ QUEUE_OPTION_NAME ] );
--- a/client/customer-effort-score-tracks/customer-effort-score-tracks-container.js
+++ b/client/customer-effort-score-tracks/customer-effort-score-tracks-container.js
@@ -75,12 +75,6 @@ CustomerEffortScoreTracksContainer.propTypes = {
 
 export default compose(
 	withSelect( ( select ) => {
-		if (
-			window.wcAdminFeatures &&
-			window.wcAdminFeatures[ 'customer-effort-score-tracks' ] !== true
-		) {
-			return { queue: [], resolving: false };
-		}
 		const { getCesSurveyQueue, isResolving } = select( STORE_KEY );
 		const queue = getCesSurveyQueue();
 		const resolving = isResolving( 'getOption', [ QUEUE_OPTION_NAME ] );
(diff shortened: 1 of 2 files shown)


adrianduffell (commented):
Nice work here @moon0326! This tested well for me 👍 

> I would appreciate any suggestions on other approaches.

I’m not opposed to the approach taken but I like the first one better.

I think the first one is safer because the second one still has some code paths executed inside the feature. I feel like that could risk some unknown bug in the feature spilling out despite when the flag is off.

moon0326:
Thank you @adrianduffell for the feedback. I did a few more tests and turned out that [this line of code](https://github.com/woocommerce/woocommerce-admin/blob/cf346f75a20f5704fd72e2d63ec94522b15cdace/client/analytics/components/report-filters/index.js#L164) does not error out without the `CustomerEffortScoreTracksContainer`. I assumed (should not have) that the code will error out without the store initialized. 

I think we are safe to take the first approach. I will make the change 👍

moon0326:
@adrianduffell  The new change is in https://github.com/woocommerce/woocommerce-admin/pull/5800/commits/73865347a0ec6b11d2dfffacf3c82c5b9ea9259d

ilyasfoo:
Nice work 👍 ! I think adding a test would also be great here.

moon0326:
@ilyasfoo Good idea! I will work on it tomorrow (would be my first JS test in this repo 😉 )

mattsherman (approved):
This fix tests well for me. Nice work!

moon0326:
@mattsherman Thank you, Matt 👍 

I'm merging it for now. I think we can address how the store should be turned on/off by the feature flag in a separate ticket.

PR #5801, opened 2020-12-02
Home Screen: Do not show store setup activity panel.

### LINKED ISSUE #5792: Remove Store Setup link from Home screen
**Describe the bug**
The persistent store setup link in the activity panel redirects users back to the Home screen to complete setup. This is super important across other screens, but on the Home Screen it's not necessary. It's confusing because when you click on the link, it doesn't result in any action or change to the screen.

**To Reproduce**

1. Go to 'WooCommerce Home Screen'
2. Click on 'Store Setup'
3. See that it doesn't result in any changes

**Expected behavior**
The store setup link should persist in the Activity panel across all screens except for on the Home screen.

**Screenshots**
https://share.getcloudapp.com/4guOxRWW
Screen Recording 2020-12-02 at 11 19 16 AM

Chatted with @samueljseay and @pmcpinto about this in Slack.

timmyc:
Fixes #5792

@elizaan36 pointed out that it would be good to remove the Store Setup activity panel link from being shown on the home screen ahead of the final 4.8 release. This branch adds the logic, and some tests, to do that.
### Screenshots

__Before__
home-store-setup-before

__After__
home-store-setup-after

- Either reset site options to a state where the Setup Checklist is being shown, or run this branch on a new site.
- Navigate to the home screen and verify the setup link in the activity panel/header area is not shown
- Navigate to another wc-admin powered page like Customers and verify that it **is shown**
- Navigate to an embedded page, like Orders, and verify that it is still shown there also

### DIFF
--- a/client/header/activity-panel/index.js
+++ b/client/header/activity-panel/index.js
@@ -145,7 +145,10 @@ export class ActivityPanel extends Component {
 			! isEmbedded && this.isHomescreen() && ! isPerformingSetupTask;
 
 		const showStoreSetup =
-			! taskListComplete && ! taskListHidden && ! isPerformingSetupTask;
+			! taskListComplete &&
+			! taskListHidden &&
+			! isPerformingSetupTask &&
+			( ! this.isHomescreen() || isEmbedded );
 
 		const inbox = showInbox
 			? {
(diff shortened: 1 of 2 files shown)


joelclimbsthings (approved):
I put it through the gauntlet (as in tested per your instructions) and all was well! 

The added test also fails as it should when removing your change :shipit:

timmyc:
Thank you kindly for the quick review @joelclimbsthings 🙇

timmyc:
cherry picked to `release/1.7.3` in https://github.com/woocommerce/woocommerce-admin/commit/5ca515c0cbb3f093f5d14cc870f6bd0749467e56

PR #5811, opened 2020-12-03
Register core product taxonomies

### LINKED ISSUE #5795: Can't Reach Product Taxonomy Pages in Navigation
**Describe the bug**

The new navigation feature is missing elements for Product Categories, Tags, and Attributes.

**To Reproduce**

1. Ensure the Navigation feature is enabled (WooCommerce > Settings > Advanced > Features)
2. Go to WooCommerce > Products
3. Verify there are no links for the taxonomies

**Expected behavior**

I expect to be able to reach the Product taxonomy management screens.

**Screenshots**

**Current nav:**

Screen Shot 2020-12-02 at 12 03 27 PM

**New nav:**

Screen Shot 2020-12-02 at 1 00 26 PM

joshuatf:
Fixes #5795

* Adds methods to register and detect taxonomy screens.
* Registers core taxonomies and custom product attributes

~~Note that if you create a custom product atribute and visit that page (e.g., `wp-admin/edit-tags.php?taxonomy=pa_test-attr&post_type=product`) that the menu item is not correctly matched.  Going to leave this for a follow-up since decisions made in https://github.com/woocommerce/woocommerce-admin/pull/5798 may affect this.~~ - Fixed by regex approach.

### Screenshots
Screen Shot 2020-12-03 at 5 20 37 PM

1. Enable the nav feature.
1. Navigate to the "Products" category.
1. Make sure the expected items and appear.
1. Make sure navigating to each of the taxonomies and individual term pages works as expected (with the exception of the custom PA mentioned above).


psealock (commented):
I gave this a test run and its all working correctly. Seeing as this touches route matching logic, we can see if a resolution happens in one of the two try PRs before going ahead with this one.

joshuatf:
Agreed.  I'm leaning towards the regex approach because if that gets merged, we don't need to continue adding hacky exception solutions to the matching logic.

joshuatf:
@psealock This has been rebased and is ready for review.  Custom match expressions have been added for tags and product attributes so these now match correctly on all tag pages.

src/Features/Navigation/CoreMenu.php
```diff
@@ -251,6 +277,18 @@ public function register_post_types() {
...
+	/**
+	 * Register all core post types.
+	 */
+	public function register_taxonomies() {
+		Screen::register_taxonomy( 'product_cat' );
+		Screen::register_taxonomy( 'product_tag' );
+		$product_attribute_taxonomies = wc_get_attribute_taxonomies();
+		foreach ( $product_attribute_taxonomies as $taxonomy ) {
+			Screen::register_taxonomy( 'pa_' . $taxonomy->attribute_name );
+		}
```
jeffstieler, line 289: What do you think about matching the product attribute taxonomies with a pattern instead of registering all of them?

If a store has a LOT of attributes, this could be a potential performance hit - especially if it's ultimately unnecessary.

src/Features/Navigation/Screen.php
```diff
@@ -91,6 +114,7 @@ public static function is_woocommerce_page() {
 
 		if (
 			in_array( $post_type, $post_types, true ) ||
+			in_array( $taxonomy, $taxonomies, true ) ||
```
jeffstieler, line 117: I guess a pattern match for the `pa_` taxonomy would go here?

src/Features/Navigation/CoreMenu.php
```diff
@@ -181,7 +196,18 @@ public static function get_items() {
...
+				$product_cat_items['default'],
+				$product_tag_items['default'],
+				array(
+					'id'              => 'woocommerce-product-attributes',
+					'title'           => __( 'Attributes', 'woocommerce-admin' ),
+					'url'             => 'edit.php?post_type=product&page=product_attributes',
+					'capability'      => 'manage_product_terms',
+					'order'           => 40,
+					'parent'          => 'woocommerce-products',
+					'matchExpression' => 'edit.php(?=.*[?|&]page=product_attributes(&|$|#))|edit-tags.php(?=.*[?|&]taxonomy=pa_)(?=.*[?|&]post_type=product(&|$|#))',
```
jeffstieler, line 208: What do you think about documenting these expressions inline? Like: https://github.com/WordPress/WordPress/blob/master/wp-includes/formatting.php#L4552-L4562

jeffstieler (commented):
This tested well for me and the code looks pretty reasonable. I've got one main question about the product attribute taxonomies.

psealock:
Removed the cherry picking label as the 1.8.0 ship has sailed.

timmyc:
> Removed the cherry picking label as the 1.8.0 ship has sailed.

I think we are very likely to have a 1.8.1 since there is a known issue in CES, and we will likely have other bug reports once 4.9 beta lands. So I'm going to add the cherry pick label back @psealock cc @samueljseay

joelclimbsthings:
Just wanted to report that I also took this for a spin, and it tested well. No issues.

src/Features/Navigation/CoreMenu.php
joelclimbsthings: Hadn't seen this before; I like the idea.

joshuatf: Updated to break into commented groups.

src/Features/Navigation/CoreMenu.php
joshuatf: I see what you mean; I'm wondering what are some of the higher numbers we've seen so we can test against scenarios like this.

But for now, I've updated this to test against a simple string test for product taxonomies.  This will not be quite as accurate as getting and registering the taxonomies, but conflicts here should hopefully be edge cases.

src/Features/Navigation/Screen.php
```diff
@@ -99,6 +124,24 @@ public static function is_woocommerce_page() {
...
 	}
 
+	/**
+	 * Check if a given taxonomy is a WooCommerce related taxonomy.
+	 *
+	 * @param string $taxonomy Taxonomy.
+	 * @return bool
+	 */
+	public static function is_woocommerce_taxonomy( $taxonomy ) {
+		if ( in_array( $taxonomy, array( 'product_cat', 'product_tag' ), true ) ) {
```
psealock, line 134: If an extension added a taxonomy, would it need to be included in this array via filter?

psealock (approved):
This is working well and tests without issue. Regex works on arbitrary order of query params too 👍🏽 

I just have a question about extension-created taxonomies and this https://github.com/woocommerce/woocommerce-admin/pull/5811#discussion_r542682376 feels like there should be a follow up issue if I understand correctly.

src/Features/Navigation/Screen.php
joshuatf: No, they would still register via `register_taxonomy()`.

psealock (approved):
Thanks @joshuatf, LGTM 🚢

PR #5812, opened 2020-12-03
Update Loader.php

richardteachout:
Added a tweek to handle a UI error.  Shouldn't affect anything else.. I think it conflicts with a few other plugins?

Fixes #

I was getting a UI error after latest version, this eliminated the error.

### Accessibility

-   Ex: Open page `url`
-   Click XYZ…

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -1091,7 +1091,9 @@ class Loader {
 				$group_settings   = $setting_options->get_group_settings( $group );
 				$preload_settings = [];
 				foreach ( $group_settings as $option ) {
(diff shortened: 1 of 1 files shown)


adrianduffell:
Thanks for the PR @richardteachout ! 

> I think it conflicts with a few other plugins?

To help with testing, do you know which plugins these are?

samueljseay:
@richardteachout any chance you could provide that testing information? If this doesn't see any further activity I might just close it for now, but please re-open if you can provide that information. This will also need a rebase against `main` as its been hanging around a while now. 😃

PR #5814, opened 2020-12-03
Enable CES in core and plugin configs

becdetat:
This enables CES in the core and plugin configs.

It's pretty straightforward but if you wanted to test it you could `npm run test:zip` and upload the generated zip to a JN site and verify that the CES snackbars get displayed.


adrianduffell (approved):
LGTM 🚀

PR #5817, opened 2020-12-03
Send store age in weeks to tracks

becdetat:
This changes the store age sent to tracks as part of the Customer Effort Score so that it is in weeks, not months. Conversation here: https://woostartp2.wordpress.com/2020/09/25/customer-effort-score-2/#comment-1334.

1. It's probably a good idea to see a track being sent before checking out this branch:
    1. Perform an action that will trigger a CES snackbar, such as adding a new product
    1. Note that in the console, a track is sent when the snackbar is displayed. The `store_age` property should be in months, rounded to the nearest month
1. Check out this branch
1. Perform a CES trigger action
1. Note that in the console, the track's `store_age` property is now in weeks, rounded to the nearest week.


moon0326 (approved):
It tested well and LGTM 👍

PR #5818, opened 2020-12-04
Add nav item route regex matching

joshuatf:
Part 2 of route detection exploration.  Part 1 - https://github.com/woocommerce/woocommerce-admin/pull/5798

* This PR adds the ability for nav items to determine active status via regular expressions.
* By default a nav item will be matched if the current location begins with this nav item's URL.
* Matching will still cycle through all registered items and use the last matched item.  This is optional, but may be slightly safer if a catch-all page is registered prior to other items.  For example, the WCA homepage `admin.php?page=wc-admin` would match before any others if an expression had not been added).

While I was pretty opposed to this approach, I'm starting to lean heavily in favor of this approach for a few reasons.
1. The default expression works for 90% of use cases without any fancy logic.
1. It opens up the possibility for negative lookaheads to prevent matching on items even if the path matches.
1. It resolves issues we have with matching both post types and taxonomies.
1. It allows a catch-all page if a plugin wants to mark a different page active.
1. Less logic to test.

1. Register a page in WCA without registering in the nav:

```php
wc_admin_register_page(
    array(
        'id'     => 'unregistered-page',
        'title'  => __( 'Unregistered Page', 'woocommerce-admin' ),
        'path'   => '/unregistered-page',
    )
);
```

2. Navigate to `wp-admin/admin.php?page=wc-admin&path=/unregistered-page`
2. Note the home item is not marked active despite a matching path and start of URL.
3. Navigate to `Home` and its subpages (tasks).
4. Make sure `Home` is now marked active.
5. Navigate to various post type pages (orders and products).
6. Make sure items are correctly marked active.
7. Make sure no other regressions have occurred with active status on items.

### DIFF
--- a/client/navigation/test/utils.js
+++ b/client/navigation/test/utils.js
@@ -81,6 +81,26 @@ describe( 'getMatchingItem', () => {
 		const matchingItem = getMatchingItem( sampleMenuItems );
 		expect( matchingItem.id ).toBe( 'multiple-args' );
 	} );
+
+	it( 'should match an item with irrelevant query parameters', () => {
+		window.location = new URL(
+			getAdminLink(
+				'admin.php?page=wc-admin&path=/test-path&section=section-name&foo=bar'
+			)
+		);
+		const matchingItem = getMatchingItem( sampleMenuItems );
+		expect( matchingItem.id ).toBe( 'multiple-args' );
+	} );
+
+	it( 'should match an item with query parameters in mixed order', () => {
+		window.location = new URL(
+			getAdminLink(
+				'admin.php?foo=bar&page=wc-admin&path=/test-path&section=section-name'
+			)
+		);
+		const matchingItem = getMatchingItem( sampleMenuItems );
+		expect( matchingItem.id ).toBe( 'multiple-args' );
+	} );
 } );
 
 describe( 'getDefaultMatchExpression', () => {
--- a/client/navigation/test/utils.js
+++ b/client/navigation/test/utils.js
@@ -10,8 +10,7 @@ import {
 	addHistoryListener,
 	getFullUrl,
 	getMatchingItem,
-	getMatchScore,
-	getParams,
+	isMatch,
 } from '../utils';
 
 const originalLocation = window.location;
@@ -83,7 +82,7 @@ describe( 'getMatchingItem', () => {
 	} );
 } );
 
-describe( 'getMatchScore', () => {
+describe( 'isMatch', () => {
 	beforeAll( () => {
 		delete window.location;
 		window.location = new URL( getAdminLink( '/' ) );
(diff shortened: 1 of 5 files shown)


psealock:
Awesome @joshuatf, its really nice to get a feel for how these work.

I played around a bit and discovered some exceptions.

```php
// Add a menu page to see what this looks like
function page_content(){
    echo '<div class="wrap"><h2>Testing</h2></div>';
}
add_menu_page( 'My Extension', 'My Extension', 'manage_woocommerce', 'my-extension-slug', 'page_content' );

// Register the item
Menu::add_plugin_item(
		array(
			'id'         => 'my-extension',
			'title'      => __( 'My Extension', 'my-extension' ),
			'capability'    => 'manage_woocommerce',
			'url'        => 'my-extension-slug',
			'parent' => 'woocommerce',
		)
	);
```

Out of the box, this is working great. Visiting `/wp-admin/admin.php?page=my-extension-slug` highlights the matched item. I can even add any number of query parameters and the correct match is found.

However, if I add a query parameter in the wrong order, it does not match. Maybe thats not a huge deal in the simplest cases of `page=*` usually being the first query parameter. But if my item has a query parameter as well, then the requirement of order becomes more difficult to ensure.

```php
'url'        => 'my-extension-slug&tab=one',
```
Visiting `/wp-admin/admin.php?page=my-extension-slug&tab=one` works, but `/wp-admin/admin.php?page=my-extension-slug&view=fullscreen&tab=one` does not.

I suppose in this case, you could add a `matchExpression` to accurately capture the `tab=one` parameter in any given order. I've been dabbling in regular expressions for a while now but by no means an expert. How to craft a regEx with this capability isn't entirely obvious to me. Is this something we want to ask of our extension authors? They will also need to handle query parameter order. 

The more I play with this, the more I see how this is exactly the same solution as https://github.com/woocommerce/woocommerce-admin/pull/5798. What I mean is, the query parameters supplied in the url are considered very important and MUST be included be in the url to indicate a full match. Other query parameters are disregarded as inconsequential. The only difference in the two approaches is this one query parameter order matters, at least in the default sense.

Maybe one improvement to the default expression could be a look forward from `&` to make query order optional. I'm not sure how to do that, but it feels like parsing the url to compare them explicitly is easier.

joshuatf:
>How to craft a regEx with this capability isn't entirely obvious to me. Is this something we want to ask of our extension authors? They will also need to handle query parameter order.

I'd also prefer that 3PDs don't need to touch regex, but I think the default regex matching will work for 99% of these cases and while it's possible for params to be loaded out of order and not match, I think that's pretty edge case and I'm having difficulty finding a way to break this without manually modifying the location URL.

That said, matching params in any order would bullet-proof this and if we can make this the default behavior so that authors don't need to worry about writing regex, then I think we have a win-win.

I'll see what what I can do today.

joshuatf:
@psealock Updated the default expression to match params of menu item in any order in https://github.com/woocommerce/woocommerce-admin/pull/5818/commits/fe1840b2bbc3633d08ce702e6693afd98d7ee470

Put in another way, any arguments that a menu item's URL includes are considered determinant by default, which I actually think is more expected than an arbitrary list of determinant params.  For example, consider the following menu item:

`admin.php?page=custom-page&other_param=a`

Previously `admin.php?page=custom-page` would loosely match the above item, but the inclusion of `other_param=a` in the menu item is probably important given the menu item included this as part of the URL and the expectation is that the URL would include this parameter.  Same logic for hashes as well.

joelclimbsthings:
I'm really liking the looks of this so far, and just learned more about positive look-ahead which is pretty :boom: . 

I haven't tested this thoroughly, but it _looks_ like as it's written now we still have the potential of matching more generic pages at the expense a more exact one? 

For instance, for the following urls:

```
[ 
    "admin.php?page=wc-admin&path=/test-path#anchor", 
    "admin.php?page=wc-admin&path=/test-path" 
] 
```

Assuming the current page includes the `#anchor`, I believe the 2nd item (same `args`, with just the lack of the `hash`) will still be returned as the correct match since it comes after the first (and we're current accepting the final match only). 

Put another way, is it important that these matches be reliable no matter the order of the items in the array? Or can we count on them to always have the more generic matches first?

I also tested this by simply altering the `sampleMenuItems` array in `/client/navigation/test/utils` to include a `.reverse()` on the end, and it results in a couple failing tests.

joshuatf:
Thanks for taking this for a spin @joelclimbsthings!

>Assuming the current page includes the #anchor, I believe the 2nd item (same args, with just the lack of the hash) will still be returned as the correct match since it comes after the first (and we're current accepting the final match only).

Correct.  This approach avoids that issue in 2 ways:
* Negative lookaheads - If a plugin wants to register multiple items that have similar paths, it would be wise to explicitly negate paths that have additional args or a hash, as we do with the wc-admin `home` item.
* Parent pages registered first - This is pretty common anyway, which is why we defer to the last item.  While it's certainly possible that a plugin registers items in reverse order, this is one small extra layer of protection.

We could take this a step further and still provide a match score, so that if all determinant params are met or an exact match is found, we still matched based on highest score.  This would be pretty trivial to add in if we think it's worth it.

joelclimbsthings:
Awesome, just wanted to clarify what we want to happen here @joshuatf . 

>     * Negative lookaheads - If a plugin wants to register multiple items that have similar paths, it would be wise to explicitly negate paths that have additional args or a hash, as we do with the wc-admin `home` item.
> 

I really like that we have this capability, although I think we would want to minimize the need for it since it's not entirely straight-forward.

>     * Parent pages registered first - This is pretty common anyway, which is why we defer to the last item.  While it's certainly possible that a plugin registers items in reverse order, this is one small extra layer of protection.
> 

This was my thought as well. Safe to say we're talking about an edge case to some degree, but potentially a not entirely insignificant one?

> 
> We could take this a step further and still provide a match score, so that if all determinant params are met or an exact match is found, we still matched based on highest score. This would be pretty trivial to add in if we think it's worth it.

I was thinking the same; since it's a trival add, and may prevent 3pd's from having to figure out what a negative look-ahead is, I feel that it might be worth it (despite a bit of increased complexity).

psealock:
@joshuatf, I hope you don't mind I just added a couple tests in https://github.com/woocommerce/woocommerce-admin/pull/5818/commits/2efe78318c476caa407dc455da60a5c46e47c015 for some of the cases I was playing around with.

They pass, but I also tried @joelclimbsthings's trick of adding `reverse()` to the end of the array and those added tests fail too. If adding in match scores is trivial as mentioned earlier, I think should be included. Depending on registration order seems brittle. Alternative thought: would it be better to re-order the items after registration? That might be error prone too.

joshuatf:
Thanks for the reviews @joelclimbsthings and @psealock!  Updated this to again return a match score using the new regex patterns.

So the logic for this now by default in brief is:
* A location must match the path and all params (in any order) in a URL to be considered a match.
* If the location is a match, we get the highest score (1 point for path and a point for each query arg and hash).
* Exact URL matches receive the highest score.

joelclimbsthings (approved):
I spent a little time trying to break this and didn't succeed. I'm sure there are a few edge cases, but they're gracefully handled with the option `matchExpression`. 

LGTM! :shipit:

psealock:
I used tests to cover some pitfalls we were experiencing so I went ahead and pushed them in https://github.com/woocommerce/woocommerce-admin/pull/5818/commits/8bfb2e64841cc4b177205c41c9a527dc05824907. I hope you don't mind @joshuatf

psealock (approved):
This is pretty awesome Josh 😮  and testing well for me.

The only suggestion I have is the addition of comments in `getDefaultMatchExpression`. I imagine someone going through that code in the future would appreciate some commentary.

pre-approving 👍🏽

PR #5819, opened 2020-12-04
Convert before and after dates to UTC in Analytics queries

### LINKED ISSUE #5810: Unable to see some data in  Analytics due to frontend and backend timezone difference
**Describe the bug**

Some stats are not available in Analytics depending on the time of the day you query. I think it's easier to describe this bug with a specific scenario.

Scenario  -  Server timezone is set to UTC (default)

1. A customer places an order on 12/02/20 at 8:00 PM in PST time. `wp_wc_order_stats.date_created` and `wp_wc_order_stats.date_created_gmt` values will be saved in UTC time. In this case, the values for both colulms is 12/03/20 04:00:00 (PST + 8 hours)

2. A store owner who lives in Los Angeles (PST) navigates to Analytics -> Orders and change the date filter to "Today". The page makes an Ajax requests to `wp-json/wc-analytics/reports/orders?orderby=date&order=desc&after=2020-12-02T00%3A00%3A00&before=2020-12-02T23%3A59%3A59&page=1&per_page=25&extended_info=true&_locale=user
`
Note the `before` and `after` values.

3. The store owner sees 0 orders since the `wp_wc_order_stats.date_created` and `wp_wc_order_stats.date_created_gmt` are past 2020-12-02.

4. It's impossible to query the order data since the datepicker does not let me choose a future date.

This happens due to different timezone being used in the backend and frontend.

**To Reproduce**

1. Make sure your WordPress is using the UTC timezone (this is the default setting).
2. Place a test order when the current UTC time is a day ahead of your timezone. Let's say you are in the PST timezone, then you should place a test order after 4 PM since UTC and PST has 8 hours difference. 
3. Make sure the test order is in `wp_wc_order_stats`
4. Navigate to Analytics -> Orders and change the date filter to "Today"
5. Your order count is 0

**Expected behavior**
I should be able to see orders.

**Additional context**
This probably has come up already, but I could not find an issue related to it.
I think this issue can be fixed by sending the local timezone from JS to the backend to re-calculate the time value for the query. We could even make a user-level setting for JS timezone.

moon0326:
Fixes #5810 

This PR fixes issue #5810 by converting `before` and `after` dates into UTC (GMT to be exact). More details are in the issue.

I think the fix is reasonable, but I want to be extra careful since this is how it has been working from the beginning. I would appreciate any inputs/suggestions or alternative solutions.

- Make sure that your server timezone is set to UTC by navigating to WooCommerce -> Status. Check for the `Default timezone is UTC` setting. UTC is the default setting unless you have modified `wp-settings.php` for some reason.
- Set a user timezone either by selecting a city or an offset on Settings -> General page. Let's use "Los Angeles" or "UTC-8" for testing purposes. 
- Enable debug mode in `wp-config.php` by setting `WP_DEBUG` and `WP_DEBUG_LOG` constants to `true`.

1. Make a test order and confirm that the order is in the `wp_wc_order_status` table and both `date_created` and `date_created_gmt` are in UTC time.
2. Open [Orders/DataStore.php](https://github.com/woocommerce/woocommerce-admin/blob/ea83f9ba690329cbfc65f373c4a51ccc6dc88066/src/API/Reports/Orders/DataStore.php#L253) and add `error_log($this->subquery->get_query_statement());`
3. Open a terminal and cd to your WordPress directory. Start watching debug file by executing `tail -f wp-content/debug.log`
4. Navigate to Analytics -> Orders and select "Today" from the datepicker. Selecting "Today" will request YYYY-MM-DD 00:00:00 and YYYY-MM-DD 23:59:59 to the backend.
5. Check the terminal log. You should see a log that contains the following line.
`wp_wc_order_stats.date_created <= '2020-12-04 07:59:59' AND wp_wc_order_stats.date_created >= '2020-12-01 08:00:00'`
6. Confirm that the dates have values of +-(UTC offset of your timezone).

If you don't see anything from the console log, try clearing Analytics cache from WooCommerce -> Status -> Tools -> Clear analytics cache

### DIFF
--- a/src/API/Reports/DataStore.php
+++ b/src/API/Reports/DataStore.php
@@ -692,6 +692,8 @@ class DataStore extends SqlQuery {
 				$datetime_str = $query_args['before']->date( TimeInterval::$sql_datetime_format );
 			} else {
 				$datetime_str = $query_args['before']->format( TimeInterval::$sql_datetime_format );
+				$datetime_str = TimeInterval::convert_local_datetime_to_gmt( $datetime_str )
+											->format( TimeInterval::$sql_datetime_format );
 			}
 			if ( isset( $this->subquery ) ) {
 				$this->subquery->add_sql_clause( 'where_time', "AND {$table_name}.date_created <= '$datetime_str'" );
@@ -705,6 +707,8 @@ class DataStore extends SqlQuery {
 				$datetime_str = $query_args['after']->date( TimeInterval::$sql_datetime_format );
 			} else {
 				$datetime_str = $query_args['after']->format( TimeInterval::$sql_datetime_format );
+				$datetime_str = TimeInterval::convert_local_datetime_to_gmt( $datetime_str )
+											->format( TimeInterval::$sql_datetime_format );
 			}
 			if ( isset( $this->subquery ) ) {
 				$this->subquery->add_sql_clause( 'where_time', "AND {$table_name}.date_created >= '$datetime_str'" );


samueljseay (approved):
Testing well for me. 🚀

jeffstieler:
I think we should potentially revert this change so we can nail down the behavior of date ranges for all possible customer/store timezone scenarios. This fixes a specific case (store tz = GMT) but breaks others (store has local tz).

For example, with this change in place:

* Store timezone offset -5 (ET)
* Place an order at 7:20 PM ET
* Query for orders `after` `T19:00:00`
* There will be no results since the 19 got 5 added to it and compared to the store local time value

`tl;dr` - if the store *isn't using UTC* for it's time offset, the `date_created` and `date_created_gmt` columns in the `wc_order_stats` table will differ.

@moon0326 @samueljseay 

Whatever we decided the functionality should be - this requires substantial test coverage.

moon0326:
@jeffstieler Thank you for the clarification. I assumed that WordPress always uses UTC unless the user somehow decides to manually change the timezone it in `wp-settings.php`. In that case, I totally agree that this is not the best way to fix the issue.

jeffstieler:
> I think it's very rare that a store uses a different timezone in the server since UTC is the WordPress default and UTC timezone is one of the checkpoints on the WC Admin status page.

@moon0326 I'm not sure that assumption is correct. The default check there is looking at the server configuration (using [`date_default_timezone_get()`](https://www.php.net/date_default_timezone_get)), not the WordPress setting value.

moon0326:
@jeffstieler Thank you. It looks like there are no additional checks for the timezone on the status page (it just uses date_default_timezone_get() without validating the value as you mentioned) (https://github.com/woocommerce/woocommerce/blob/aa89afcf957beb6a387ad7d5d36fd7d95b3a353b/includes/class-wc-tracker.php#L249).

I will run a few more tests and report my findings.

moon0326:
As for a possible solution, I think the following `convert_local_datetime_to_server_tiemzone()` function can be used instead of `convert_local_datetime_to_gmt()` function.

```php
/**
 * Converts manual_offset value in seconds to DateTimeZone.
 *
 * @param $manual_offset_in_seconds String manual offset value in seconds.
 *
 * @return \DateTimeZone
 */
public static function convert_manual_offset_to_timezone( $manual_offset_in_seconds ) {
    // WordPress manual offset value can come with +- sign. Normlaize it.
    if ($manual_offset_in_seconds[0] === '+' || $manual_offset_in_seconds[0] === '-') {
        $sign = substr( $manual_offset_in_seconds, 0, 1 );
        $value = substr( $manual_offset_in_seconds, 1 );
        $timezone_value = date( "{$sign}H:i", $value );
    } else {
        // when the value is 0, add + to make it compatible with DateTimeZone
        $timezone_value = '+' . $manual_offset_in_seconds;
    }

    return new \DateTimeZone( $timezone_value );
}

/**
 * Converts local datetime to server timezone
 *
 * @param string $datetime_string String representation of local datetime.
 * @return DateTime
 */
public static function convert_local_datetime_to_server_tiemzone( $datetime_string ) {
    if ( get_option( 'timezone_string' ) ) {
        $timezone = new \DateTimeZone( wc_timezone_string() );
    } else {
        $timezone = self::convert_manual_offset_to_timezone( wc_timezone_offset() );
    }

    $datetime = new \DateTime( $datetime_string, $timezone ) );
    $datetime->setTimezone( new \DateTimeZone( date_default_timezone_get() ) );
    return $datetime;
}
```

I see that there are other time-related functions in [TimeInterval.php](https://github.com/woocommerce/woocommerce-admin/blob/main/src/API/Reports/TimeInterval.php). I will find and test places where they are being used and report back.

moon0326:
I think I overlooked this issue. This issue might be simpler than I initially thought.

I think we don't have to think about the default server timezone as long as we set the correct timezone with `DateTime` instances since `DateTime::format` uses the default timezone to convert the format.

Assuming your system default timezone is UTC,

```php
$tz = new DateTimeZone('America/Los_Angeles');
$date = new DateTime('2020-12-01 00:00:00', $tz);
echo $date->format("Y-m-d H:i:s");

// outputs 2020-12-01 00:00:00
```

```php
$tz = new DateTimeZone('America/Los_Angeles');
$date = new DateTime('2020-12-01 00:00:00');
$date->setTimezone($tz);
echo $date->format("Y-m-d H:i:s");

// outputs 2020-11-30 16:00:00
```

I think removing `$local_tz` from [this line](https://github.com/woocommerce/woocommerce-admin/blob/d8f7ecbe261dd4a12da4c8c642b3d5534a7d9163/src/API/Reports/DataStore.php#L314) fixes the problem and it's relatively safe imo. 

https://github.com/woocommerce/woocommerce-admin/blob/d8f7ecbe261dd4a12da4c8c642b3d5534a7d9163/src/API/Reports/DataStore.php#L314

We also need to update places where we use `new DateTimeZone( wc_timezone_string() )` to check `wc_timezone_offset` as well.

PR #5820, opened 2020-12-04
Add Customer Type column to the Orders report table

### LINKED ISSUE #5181: New or returning customer is missing
Hello,

Im not a pro in programing so bare with me here :) I love this plugin under the page Order i was able to see if the customer was an returning or an new customer. This was important to me since I add different packing (greeting cards, coupons etc) for new customers compared to old customer. Now I cannot find anyware this in an easy way. What happend here? :)

Thanks in advance. 
Best
Johan

becdetat:
Fixes #5181

This adds the customer type to the orders table. The data was being generated and sent on the server, this PR just maps the data to the UI.

### Screenshots
image

1. have some orders in the system
1. go to Analytics/Orders
1. note that the "Customer Type" column is present and populated

### DIFF
--- a/client/analytics/report/orders/table.js
+++ b/client/analytics/report/orders/table.js
@@ -17,6 +17,9 @@ import ReportTable from '../../components/report-table';
 import { CurrencyContext } from '../../../lib/currency-context';
 import './style.scss';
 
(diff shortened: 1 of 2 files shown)


samueljseay (approved):
Looks great, thank you!

PR #5821, opened 2020-12-04
Add page parameter to getNewPath to override default page wc-admin

### LINKED ISSUE #5077: updateQueryString updates the page value to wc-admin, instead of custom value
**Describe the bug**
https://github.com/woocommerce/woocommerce-admin/blob/1c05641d5fd5e2d85f6cbdffb9a403540b52a67c/packages/navigation/src/index.js#L102

If I try to use the updateQueryString(), I am unable to use my own WP-Admin page. It redirects me to 'wc-admin'. As I can see 'wc-admin' is hard set in the getNewPath() method which is used in updateQueryString.

It would be nice if I can set the page option from the query's page parameter.

ilyasfoo:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/5077

This PR adds `page` parameter to `getNewPath` and `updateQueryString` in Navigation API to provide the option for developers to override the default page. Previously it was possible to override the default page through `query` parameter but does so implicitly. This PR aims to improve the developer experience.

Unit tests are added to verify that it should not change previous behavior.

- Run tests `npm run test`
- Run e2e tests `npm run test:e2e`

### Changelog Note:
Enhancement: Add page parameter to override default `wc-admin` page in Navigation API

### DIFF
--- a/packages/navigation/src/index.js
+++ b/packages/navigation/src/index.js
@@ -93,14 +93,16 @@ export function getSearchWords( query = navUtils.getQuery() ) {
  * @param {Object} query object of params to be updated.
  * @param {string} path Relative path (defaults to current path).
  * @param {Object} currentQuery object of current query params (defaults to current querystring).
+ * @param {string} page Page key (defaults to "wc-admin")
  * @return {string}  Updated URL merging query params into existing params.
  */
 export function getNewPath(
 	query,
 	path = getPath(),
-	currentQuery = getQuery()
+	currentQuery = getQuery(),
+	page = 'wc-admin'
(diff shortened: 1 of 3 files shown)


becdetat (approved):
This is great. Thanks for the unit tests. Can you add and commit the changelog note to readme.txt? Then you can go ahead and merge.

ilyasfoo:
Thank you @becdetat! I've added changelog to readme.txt in https://github.com/woocommerce/woocommerce-admin/commit/f6884c02ab4d26d822d93e41de75944f9aab0e97.

Issue #5803, opened 2020-12-03
Advanced filters not working with new/returning customer


ciszere:
**Describe the bug**
I found another issue that is related to my previous error about new/returning customer error. When I try to filter customers - Advanced filters" - "Customer is returning/new" it doesn't work. it just shows all orders regarding if I filter for new or returning

**To Reproduce**

1. Go to Analytics
2. Click on Orders
3. Click on Advanced filters
4. choose customers - Returning/new
5. The Customer list just shows all customers, no difference if i change returning or new

**Expected behavior**
I would see all returning customers when i click returning and so on. 

**Screenshots**

**Desktop (please complete the following information):**

Safari - Latest version
latest version of analytics

<img width="1440" alt="Skärmavbild 2020-12-02 kl  20 38 52" src="https://user-images.githubusercontent.com/60106580/100978089-2b917a00-3542-11eb-9edd-799420e898cc.png">

<img width="1440" alt="Skärmavbild 2020-12-02 kl  20 37 05" src="https://user-images.githubusercontent.com/60106580/100978090-2cc2a700-3542-11eb-8aa2-6f69395017ce.png">

joelclimbsthings:
This definitely seems to be a legitimate bug, and a somewhat nasty one.

To elaborate a bit further, in the example above the table below reflects the correct orders, but the chart (in original screenshots) does not. Here's another screenshot that shows both:

![image](https://user-images.githubusercontent.com/444632/101097402-f43cc080-3575-11eb-8936-9b905d7d24dd.png)

While testing this I also encountered what might be a related issue, but reflects the opposite; in this example the chart reflects the _correct_ total, while the table shows all orders. The filter is for excluding a specific product (including works fine):

![image](https://user-images.githubusercontent.com/444632/101097639-64e3dd00-3576-11eb-98f0-ebda47ea0deb.png)

joelclimbsthings:
I've added this to the `1.8.0` milestone, although we may have to defer to the next release at this point if a solution doesn't present itself pretty quickly.

jeffstieler:
@ciszere just to confirm - by "latest version of analytics" do you mean you're using version `1.7.2` of WooCommerce Admin?

ciszere:
Hello,

I have this version: 1.7.2 :)

Best regards,
Med vänlig hälsning,
Johan Forsberg, Founder

Phone: +46 768 08 82 32
Ciszere AB
Org.nr 559052-7346
www.ciszere.se


[...]

PR #5822, opened 2020-12-04
Fix Orders Report Product/Variation Exclusion Filters

jeffstieler:
Fixes (in part) #5803. It addresses the issues pointed out by @joelclimbsthings in https://github.com/woocommerce/woocommerce-admin/issues/5803#issuecomment-738403405.

This PR fixes two issues with the Orders Report filters:
* Stacking Product, Variation, and Attribute filters would try to apply all filters to the same product in the table query (summary numbers were correct, based on an matching the **order** having these items)
* Excluding a Product or Variation resulted in many duplicate orders in the table

- Go to the Orders Report, using a date range with many orders
- Add a Product filter, exclude a product seen in the results
- Verify the table has no duplicates
- Change the Product filter to "includes"
- Add a Variation filter, include a variation seen in the results (requires variable products that have been ordered)
- Verify that the table count matches the summary numbers

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -81,6 +81,7 @@ Release and roadmap notes are available on the [WooCommerce Developers Blog](htt
 
 = vnext =
 - Fix: Add Customer Type column to the Orders report table. #5820
+- Fix: Product exclusion filter on Orders Report.
 
 = 1.7.0 11/11/2020 =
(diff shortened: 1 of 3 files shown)


samueljseay (approved):
Looking good, testing good. Thank you! ✌️

PR #5823, opened 2020-12-04
Fix Inconsistent Customer Type Parameter in REST endpoints

### LINKED ISSUE #5803: Advanced filters not working with new/returning customer

adrianduffell:
The `wc-analytics/reports/orders` and `/wc-analytics/reports/orders/stats` endpoints have inconsistent parameter names for filtering the customer type: `customer_type` vs `customer`.

This was affecting the Customer Type filter in the orders analytics, reported in issue #5803, where the component expects the parameter names to be the same. This is because it uses a paramater name based on the [key in the filter configuration](https://github.com/woocommerce/woocommerce-admin/blob/831c9ff13a862f22cf53d3ae676daeabbefe90ad/client/analytics/report/orders/config.js#L247) (`customer_type`). 

Screen Shot 2020-12-04 at 1 29 41 pm

To resolve this, this PR updates `/wc-analytics/reports/orders/stats` to accept both `customer_type` and `customer` as parameter names for this filter.  I have however deprecated the `customer` parameter in its description to discourage usage.

### Alternative approach

An alternative approach is to keep the inconsistent names, and modify the filter component(s) to use a different parameter name depending on the endpoint being fetched. I think this is less ideal because the inconsistency can add confusion.

-   For the the `/wc-analytics/reports/orders/stats` REST API endpoint, check that the `customer` and `customer_type` parameters return identical results for values of `new` and `returning`.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -73,6 +73,7 @@ Release and roadmap notes are available on the [WooCommerce Developers Blog](htt
 
 == Unreleased ==
 
+- Tweak: Fix inconsistent REST API paramater name for customer type filtering.
 - Enhancement: Tasks extensibility in Home Screen. #5794
 - Enhancement: Add page parameter to override default wc-admin page in Navigation API. #5821
 - Fix: Invalidate product count if the last product was updated in the list. #5790
--- a/src/API/Reports/DataStore.php
+++ b/src/API/Reports/DataStore.php
@@ -1196,10 +1196,10 @@ class DataStore extends SqlQuery {
 		global $wpdb;
 
 		$customer_filter = '';
-		if ( isset( $query_args['customer'] ) ) {
-			if ( 'new' === strtolower( $query_args['customer'] ) ) {
+		if ( isset( $query_args['customer_type'] ) ) {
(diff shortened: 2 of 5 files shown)


adrianduffell:
This and PR https://github.com/woocommerce/woocommerce-admin/pull/5822 combined should fix the issues reported in issue #5803

jeffstieler (approved):
Nice and simple! I noticed an incorrect parameter description for `customer_type` in the schema - fix that and let's :shipit: ❗

PR #5826, opened 2020-12-04
Highlight help panel with tooltip

### LINKED ISSUE #5082: Help panel: add a tooltip highlighting the panel
**Description:** 
The Help Panel usage has been minimal, as the top-right corner can be a blind spot for users. On the other hand, there's a high % of users clicking in the support links when the open the panel. So we should have some additional investment highlighting the Help panel with a tooltip.

**Acceptance criteria:**
- [ ] Display the tooltip if the user visits a task for the 2nd time (if the task is not complete)
task-list-help-panel
- [ ] The user should only see the tooltip once - if we already displayed the tooltip on the `Add Product` task we shouldn't display it again in other tasks, in order to avoid overwhelming users.

**Event tracking:**
View the tooltip:
`wcadmin_help_tooltip_view`
Prop: task - the task where the tooltip was displayed

Click the tooltip:
`wcadmin_help_tooltip_click`
Prop: task - the task where the tooltip was displayed

louwie17:
Fixes #5082 

Added the `woocommerce_task_list_tracked_started_tasks` option as an object of { task_name: task_visit_count } to keep track of the tasks that where visited, and the `woocommerce_help_panel_highlight_shown` option to manage if the help panel tooltip was shown. The task_visit_count is increased everytime the task is clicked from the task list.

The tooltip will show when it is the second time an uncompleted task has been visited. The tooltip will **only** be hidden if a user has interacted with it. If a user refreshes without interacting it will show the tooltip again.

The tooltip is shown using a HighlightTooltip component, it makes use of an element id to target, and makes use of the wordpress `createPortal` function  and `Popover` component. This way it acts similar to the [Woo admin pointers](https://github.com/woocommerce/woocommerce/blob/master/includes/admin/class-wc-admin-pointers.php#L228).

### Accessibility

### Screenshots

helppanel-popup

- Create a new WooCommerce store, and finish the initial Wizard
- Navigate to one of the unfinished tasks on the home screen and finish it (Ex: `Add my products` and add a product)
- Go to home screen and click on the finished task (no popup should appear)
- Go to home screen and click an unfinished task (not finishing it)
- Go to home screen again and click the same unfinished task
- A popup should appear now, highlighting the `Help` panel, like this:
Screen Shot 2020-12-07 at 3 39 42 PM
- Click `Got it` or the close button.
- Help panel button should still work correctly.
- If navigating back home and to the unfinished task again the popup should not show up (this goes for any other unfinished tasks as well).
- It should of logged the `wcadmin_help_tooltip_view` when the tooltip was viewed, and the `wcadmin_help_tooltip_click` when it was dismissed.

### DIFF
--- a/client/header/activity-panel/highlight-tooltip/index.js
+++ b/client/header/activity-panel/highlight-tooltip/index.js
@@ -2,6 +2,7 @@
  * External dependencies
  */
 import PropTypes from 'prop-types';
+import { __ } from '@wordpress/i18n';
 import {
 	Popover,
 	Card,
@@ -87,16 +88,19 @@ function HighlightTooltip( {
 			{ showHighlight ? (
 				<>
 					<IsolatedEventContainer className="highlight-tooltip__overlay" />
-					<div className="highlight-tooltip__highlighter"></div>
-					<Popover className="highlight-tooltip__popover">
+					<Popover
+						className="highlight-tooltip__popover"
+						noArrow={ false }
+						focusOnMount="container"
+					>
 						<Card
-							size="small"
+							size="medium"
 							className="woocommerce-task-card woocommerce-homescreen-card"
 						>
-							<CardHeader size="small">
+							<CardHeader>
 								{ title }
 								<Button
-									size="small"
+									isSmall
 									onClick={ triggerClose }
 									icon={ close }
 								/>
@@ -108,7 +112,7 @@ function HighlightTooltip( {
 									isPrimary
 									onClick={ triggerClose }
 								>
-									{ closeBtnText }
+									{ closeBtnText || __( 'Close' ) }
 								</Button>
 							</CardFooter>
 						</Card>
--- a/client/header/activity-panel/highlight-tooltip/index.js
+++ b/client/header/activity-panel/highlight-tooltip/index.js
@@ -1,0 +1,103 @@
+/**
+ * External dependencies
+ */
+import {
+	Popover,
+	Card,
+	CardBody,
+	CardHeader,
+	Button,
+	IsolatedEventContainer,
+} from '@wordpress/components';
+import { useState, useEffect, createPortal } from '@wordpress/element';
+import { close } from '@wordpress/icons';
+
+/**
+ * Internal dependencies
+ */
+import './style.scss';
+
+const SHOW_CLASS = 'highlight-tooltip__show';
+export function HighlightTooltipUncontrolled( {
+	title,
+	show,
+	id,
+	onClose,
+	delay,
+} ) {
+	const [ showHighlight, setShowHighlight ] = useState(
+		delay > 0 ? false : show
+	);
+	const [ node, setNode ] = useState( null );
+
+	useEffect( () => {
+		const element = document.getElementById( id );
+		if ( element && ! node ) {
+			const parent = element.parentElement;
+			const container = document.createElement( 'div' );
+			container.classList.add(
+				'highlight-tooltip__container',
+				SHOW_CLASS
+			);
+			parent.appendChild( container );
+			setNode( container );
+		}
+		let timeoutId;
+		if ( delay > 0 ) {
+			timeoutId = setTimeout( () => {
+				timeoutId = null;
+				setShowHighlight( show );
+			}, delay );
+		}
+		return () => {
+			if ( node ) {
+				const parent = node.parentElement;
+				parent.removeChild( node );
+			}
+			if ( timeoutId ) {
+				clearTimeout( timeoutId );
+			}
+		};
+	}, [] );
+
+	useEffect( () => {
+		if ( ! showHighlight && node ) {
+			node.classList.remove( SHOW_CLASS );
+		}
+	}, [ showHighlight ] );
+
+	const triggerClose = () => {
+		setShowHighlight( false );
+		if ( onClose ) {
+			onClose();
+		}
+	};
+
+	if ( ! node || ! showHighlight ) {
+		return null;
+	}
+
+	return createPortal(
+		<div className="highlight-tooltip__portal">
+			<IsolatedEventContainer className="highlight-tooltip__overlay" />
+			<div className="highlight-tooltip__highlighter"></div>
+			<Popover className="highlight-tooltip__popover">
+				<Card
+					size="small"
+					className="woocommerce-task-card woocommerce-homescreen-card"
+				>
+					<CardHeader size="small">
+						{ title }
+						<Button
+							size="small"
+							onClick={ triggerClose }
+							icon={ close }
+						/>
+					</CardHeader>
+					<CardBody></CardBody>
+				</Card>
+			</Popover>
+		</div>,
+		node
+	);
+}
--- a/client/header/activity-panel/highlight-tooltip/index.js
+++ b/client/header/activity-panel/highlight-tooltip/index.js
@@ -1,11 +1,13 @@
 /**
  * External dependencies
  */
+import PropTypes from 'prop-types';
 import {
 	Popover,
 	Card,
 	CardBody,
 	CardHeader,
+	CardFooter,
 	Button,
 	IsolatedEventContainer,
 } from '@wordpress/components';
@@ -18,9 +20,11 @@ import { close } from '@wordpress/icons';
 import './style.scss';
 
 const SHOW_CLASS = 'highlight-tooltip__show';
-export function HighlightTooltipUncontrolled( {
+function HighlightTooltip( {
 	title,
-	show,
+	closeBtnText,
+	content,
+	show = true,
 	id,
 	onClose,
 	delay,
@@ -32,9 +36,10 @@ export function HighlightTooltipUncontrolled( {
 
 	useEffect( () => {
 		const element = document.getElementById( id );
+		let container;
 		if ( element && ! node ) {
 			const parent = element.parentElement;
-			const container = document.createElement( 'div' );
+			container = document.createElement( 'div' );
 			container.classList.add(
 				'highlight-tooltip__container',
 				SHOW_CLASS
@@ -50,9 +55,9 @@ export function HighlightTooltipUncontrolled( {
 			}, delay );
 		}
 		return () => {
-			if ( node ) {
-				const parent = node.parentElement;
-				parent.removeChild( node );
+			if ( container ) {
+				const parent = container.parentElement;
+				parent.removeChild( container );
 			}
 			if ( timeoutId ) {
 				clearTimeout( timeoutId );
@@ -73,31 +78,77 @@ export function HighlightTooltipUncontrolled( {
 		}
 	};
 
-	if ( ! node || ! showHighlight ) {
+	if ( ! node ) {
 		return null;
 	}
 
 	return createPortal(
 		<div className="highlight-tooltip__portal">
-			<IsolatedEventContainer className="highlight-tooltip__overlay" />
-			<div className="highlight-tooltip__highlighter"></div>
-			<Popover className="highlight-tooltip__popover">
-				<Card
-					size="small"
-					className="woocommerce-task-card woocommerce-homescreen-card"
-				>
-					<CardHeader size="small">
-						{ title }
-						<Button
+			{ showHighlight ? (
+				<>
+					<IsolatedEventContainer className="highlight-tooltip__overlay" />
+					<div className="highlight-tooltip__highlighter"></div>
+					<Popover className="highlight-tooltip__popover">
+						<Card
 							size="small"
-							onClick={ triggerClose }
-							icon={ close }
-						/>
-					</CardHeader>
-					<CardBody></CardBody>
-				</Card>
-			</Popover>
+							className="woocommerce-task-card woocommerce-homescreen-card"
+						>
+							<CardHeader size="small">
+								{ title }
+								<Button
+									size="small"
+									onClick={ triggerClose }
+									icon={ close }
+								/>
+							</CardHeader>
+							<CardBody>{ content || null }</CardBody>
+							<CardFooter isBorderless={ true }>
+								<Button
+									size="small"
+									isPrimary
+									onClick={ triggerClose }
+								>
+									{ closeBtnText }
+								</Button>
+							</CardFooter>
+						</Card>
+					</Popover>
+				</>
+			) : null }
 		</div>,
 		node
 	);
 }
+
+HighlightTooltip.propTypes = {
+	/**
+	 * The id of the element it should highlight.
+	 */
+	id: PropTypes.string.isRequired,
+	/**
+	 * Title of the popup
+	 */
+	title: PropTypes.string.isRequired,
+	/**
+	 * Text of the close button.
+	 */
+	closeBtnText: PropTypes.string.isRequired,
+	/**
+	 * Content of the popup, can be either text or react element.
+	 */
+	content: PropTypes.oneOfType( [ PropTypes.string, PropTypes.node ] ),
+	/**
+	 * If to show the popup, defaults to true.
+	 */
+	show: PropTypes.bool,
+	/**
+	 * Callback for when the user closes the popup.
+	 */
+	onClose: PropTypes.func,
+	/**
+	 * This will delay the popup from appearing by the number of ms.
+	 */
+	delay: PropTypes.number,
+};
+
+export { HighlightTooltip };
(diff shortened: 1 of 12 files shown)


client/header/activity-panel/highlight-tooltip/index.js
```diff
@@ -0,0 +1,158 @@
+/**
```
louwie17, line 1: Left it within this folder for now, although it could be moved to it's own package (might rename it in that case).
It's similar to these [WC pointers](https://github.com/woocommerce/woocommerce/blob/master/includes/admin/class-wc-admin-pointers.php#L228)

client/header/activity-panel/highlight-tooltip/index.js
```diff
@@ -0,0 +1,158 @@
...
+		if ( onClose ) {
+			onClose();
+		}
+	};
+
+	if ( ! node ) {
+		return null;
+	}
+
+	return createPortal(
```
louwie17, line 86: Same function the Modal component uses.

src/Features/Onboarding.php
```diff
@@ -583,7 +583,7 @@ public static function has_woocommerce_support( $theme ) {
 			$files     = new \RegexIterator( $iterator, '/^.+\.php$/i', \RecursiveRegexIterator::GET_MATCH );
 
 			foreach ( $files as $file ) {
-				$content = file_get_contents( $file[0] );
+				$content = file_get_contents( $file[0] ); // phpcs:ignore WordPress.WP.AlternativeFunctions
```
louwie17, line 586: This call and the changes on line 978 and 1021 where from the php linter, I couldn't commit otherwise.

client/header/activity-panel/highlight-tooltip/style.scss
```diff
@@ -0,0 +1,42 @@
...
+	.highlight-tooltip__overlay {
+		position: fixed;
+		top: 0;
+		right: 0;
+		bottom: 0;
+		left: 0;
+		background-color: rgba($black, 0.35);
+		z-index: z-index('.components-modal__screen-overlay');
+
+		@include edit-post__fade-in-animation();
```
jeffstieler, line 26: TIL!

I'm not able to see the `0.2s` fade in. 😛

client/header/activity-panel/test/index.js
```diff
@@ -193,4 +197,72 @@ describe( 'Activity Panel', () => {
...
+		} );
+
+		it( 'should not render highlight tooltip when trackedStartedTasks is included twice, not completed, but already shown', () => {
+			const { queryByText } = render(
+				<ActivityPanel
+					requestingTaskListOptions={ false }
+					taskListComplete={ false }
+					taskListHidden={ false }
+					trackedStartedTasks={ [ 'payment', 'payment' ] }
+					trackedCompletedTasks={ [ 'payment' ] }
```
jeffstieler, line 258: The test description says that this is **not** complete.

jeffstieler (changes requested):

src/Features/Onboarding.php
```diff
@@ -973,7 +975,7 @@ public static function calypso_tests() {
 			$connect_url = \Jetpack::init()->build_connect_url( true, $redirect_url, 'woocommerce-onboarding' );
 			$connect_url = add_query_arg( array( 'calypso_env' => $calypso_env ), $connect_url );
 
-			wp_redirect( $connect_url );
+			wp_save_redirect( $connect_url );
```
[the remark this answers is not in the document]
louwie17: That's what the pre-commit hook `phpcs` told me, but given the commit hook is disabled now I can revert this.

client/task-list/index.js
```diff
@@ -33,6 +33,29 @@ export class TaskDashboard extends Component {
 		document.body.classList.add( 'woocommerce-task-dashboard__body' );
 	}
 
+	trackStartedTask = ( taskName ) => {
+		const { trackedStartedTasks, updateOptions } = this.props;
+		const taskStarted = ( trackedStartedTasks || [] ).filter(
+			( task ) => task === taskName
+		);
+		if ( taskStarted.length > 1 ) {
```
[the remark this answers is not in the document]
louwie17: No particular reason, given the `trackedCompletedTasks` was also an array I followed suite with that. Originally I thought I would only ever need one item in the list, but that didn't work given the update didn't happen on the same page that had the show tooltip logic.

Suggestions otherwise? Maybe a key:number object, numbers being the amount a task was visited. The fact that I only needed to know the second time I didn't think it necessary to track after that.

client/header/activity-panel/index.js
```diff
@@ -359,12 +408,24 @@ export default compose(
...
 			isResolving( 'getOption', [ 'woocommerce_task_list_hidden' ] );
+		const trackedCompletedTasks = getOption(
+			'woocommerce_task_list_tracked_completed_tasks'
+		);
+		const trackedStartedTasks = getOption(
+			'woocommerce_task_list_tracked_started_tasks'
+		);
+		const helpPanelHighlightShown = getOption(
+			'woocommerce_help_panel_highlight_shown'
+		);
```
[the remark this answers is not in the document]
louwie17: Thanks, good point and question, I assume it would be on a per user base. I will leave that up to @pmcpinto @elizaan36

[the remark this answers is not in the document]
elizaan36: Yep, this should only be shown once per user.

client/header/activity-panel/highlight-tooltip/index.js
```diff
@@ -0,0 +1,158 @@
...
+		return () => {
+			if ( container ) {
+				const parent = container.parentElement;
+				parent.removeChild( container );
+			}
+			if ( timeoutId ) {
+				clearTimeout( timeoutId );
+			}
+		};
+	}, [] );
```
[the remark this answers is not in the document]
louwie17: Yeah, I should make up my mind here. As the `show` prop is actually kinda useless, it works best with a single id per component, and once hidden it is removed from the DOM again by the parent.

client/header/activity-panel/highlight-tooltip/index.js
```diff
@@ -0,0 +1,158 @@
...
+				<>
+					<IsolatedEventContainer className="highlight-tooltip__overlay" />
+					<Popover
+						className="highlight-tooltip__popover"
+						noArrow={ false }
+						focusOnMount="container"
+					>
+						<Card
+							size="medium"
+							className="woocommerce-task-card woocommerce-homescreen-card"
```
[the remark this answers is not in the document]
louwie17: Most likely not, will double check, and remove if not.

client/header/activity-panel/index.js
```diff
@@ -48,6 +50,8 @@ export class ActivityPanel extends Component {
 		};
 	}
 
+	recordedHelpTooltip = false;
```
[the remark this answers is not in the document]
louwie17: Good point, my head was probably thinking of functional components

client/task-list/index.js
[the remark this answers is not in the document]
jeffstieler: I'm partial to a key:number object, but I don't feel too strongly about it. One potential benefit of the object is that we might be able to extract value from the count .. perhaps send the number of times a step was viewed in the tracks event? That might tell us if particular tasks are confusing or difficult.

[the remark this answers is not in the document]
louwie17: Yeah not a bad idea, in that case I would have to change the count logic to update until it is finished. Instead of not more then 2.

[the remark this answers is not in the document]
jeffstieler: I'm just spitballing here - no need to add extra work unnecessarily.

src/Features/Onboarding.php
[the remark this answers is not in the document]
louwie17: Reverted in: https://github.com/woocommerce/woocommerce-admin/pull/5826/commits/a2ec7e3f870264161420cf241d9c329fa8409a8c

client/header/activity-panel/index.js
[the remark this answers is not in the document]
louwie17: Changed to make use of the user WC meta here: https://github.com/woocommerce/woocommerce-admin/pull/5826/commits/25a87219b47e3498b4d55471f715ac5fb987e098

client/header/activity-panel/highlight-tooltip/index.js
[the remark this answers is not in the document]
louwie17: Added a `useEffect` for the `show` property, to allow a user to re-show by toggling the show property https://github.com/woocommerce/woocommerce-admin/pull/5826/commits/c9f40511b9467a1bb7b73b620ab55968dbd590e8

When the dependency array is empty, it only gets run on initial mount, so we don't want to add any further dependencies here.

client/header/activity-panel/highlight-tooltip/style.scss
louwie17: Yeah I copied that from the [Modal css](https://github.com/WordPress/gutenberg/blob/master/packages/components/src/modal/style.scss#L12).
I can't seem to really see it either 🙈

louwie17:
@jeffstieler when you have time, are you able to re-review this?

client/header/activity-panel/index.js
```diff
@@ -277,6 +280,47 @@ export class ActivityPanel extends Component {
...
+			userPreferencesData.help_panel_highlight_shown;
+		if (
+			task &&
+			highlightShown !== 'yes' &&
+			( startedTasks || {} )[ task ] > 1 &&
+			! trackedCompletedTasks.includes( task )
+		) {
+			if ( ! this.recordedHelpTooltip ) {
+				this.recordedHelpTooltip = true;
+				recordEvent( 'help_tooltip_view' );
```
jeffstieler, line 317: This should probably be in the tooltip component, no?

client/header/activity-panel/index.js
```diff
@@ -338,6 +383,22 @@ export class ActivityPanel extends Component {
 						{ this.renderPanel() }
 					</div>
 				</Section>
+				{ showHelpHighlightTooltip ? (
+					<HighlightTooltip
+						delay={ 1000 }
```
jeffstieler, line 388: I don't think this delay is working. As a test, I changed to `10000` and there was still no delay in the tooltip opening.

client/task-list/index.js
```diff
@@ -33,6 +33,33 @@ export class TaskDashboard extends Component {
...
+	trackStartedTask = ( taskName ) => {
+		const { userPreferences } = this.props;
+		const trackedStartedTasks =
+			userPreferences.task_list_tracked_started_tasks || {};
+		const currentTaskCount =
+			trackedStartedTasks && trackedStartedTasks[ taskName ]
+				? trackedStartedTasks[ taskName ]
+				: 0;
+		// Only record task visits up to 2 times
+		if ( currentTaskCount > 1 ) {
```
jeffstieler, line 45: Why not continue to count the task visits? This could be useful data - especially if added to the [`task_view` event](https://github.com/woocommerce/woocommerce-admin/blob/542cae22a5a4d2abdbfa2fcc2db8906a021cecfb/client%2Ftask-list%2Ftasks.js#L33).

jeffstieler (changes requested):
This is just about there! I noticed the `delay` prop doesn't seem to be working, and I think we can add some value by including the task visit count in a tracks event. _(cc: @pmcpinto)_

client/header/activity-panel/index.js
louwie17: Not necessarily, given the tooltip is more abstract, but it isn't the nicest place. I added a `onShow` callback to the tooltip, and calling it from there.

client/header/activity-panel/index.js
louwie17: Nice catch, I fixed it here: https://github.com/woocommerce/woocommerce-admin/pull/5826/commits/f0a38c741891924e8a21b32db6cb36eb02c8fa06

client/task-list/index.js
louwie17: Added it to the `tasklist_click` event in the end, as that is being called on any of the tasks. Not sure if it's a bug that `task_view` event is only being triggered for the payment task?

louwie17:
@jeffstieler I fixed the delay, and made some small changes in relation to your other suggestions :)

client/task-list/index.js
jeffstieler: > Not sure if it's a bug that `task_view` event is only being triggered for the payment task?

Are you sure? Looking at the `<TaskList>` it seems that the event should fire for all tasks: https://github.com/woocommerce/woocommerce-admin/blob/6dddef0cbcdd05b6e93347bfc87a830c08133928/client%2Ftask-list%2Flist.js#L50

jeffstieler: I see both events fired:
Screen Shot 2020-12-14 at 2 20 40 PM

pmcpinto:
> This is just about there! I noticed the delay prop doesn't seem to be working, and I think we can add some value by including the task visit count in a tracks event. (cc: @pmcpinto)

@jeffstieler are you suggesting add the task count to the new events (wcadmin_help_tooltip_view and wcadmin_help_tooltip_click?

jeffstieler:
@pmcpinto I was suggesting adding the task *view* count to either the `wcadmin_task_view` event or the `wcadmin_tasklist_click` event as @louwie17 currently has it. That way we can see how many times a user visited the task before completing it. What do you think?

client/task-list/index.js
louwie17: Ooh nice, I didn't realize the `recordTaskViewEvent` was exported, and only saw it used once in the file it was defined.
This makes more sense.

pmcpinto:
> @pmcpinto I was suggesting adding the task view count to either the wcadmin_task_view event or the wcadmin_tasklist_click event as @louwie17 currently has it. That way we can see how many times a user visited the task before completing it. What do you think?

Just to confirm, the data would look something like this, correct?

`User count` prop
1: XXXX users
2: XXX users
3: XXXX users
4: XX users
5: X users
.....

jeffstieler:
> Just to confirm, the data would look something like this, correct?

I'm not sure what the 1,2,..5 represent in that structure. You can see the current structure that @louwie17 implemented here:


In text:

```json
{
    "task_name": "appearance",
    "visit_count": 3
}
```

src/Features/Onboarding.php
becdetat: I think https://codex.wordpress.org/Filesystem_API is what you should be using rather than `file_get_contents` directly

louwie17: Meaning something like this?
```
global $wp_filesystem;
$wp_filesystem->get_contents()
```
I just added the phpcs comment to fix a pre-commit error.

louwie17: I decided to revert this change, as it wasn't necessary anymore (no pre-commit hook) and unrelated to any of the changes of this PR.

client/header/activity-panel/test/index.js
```diff
@@ -193,4 +197,72 @@ describe( 'Activity Panel', () => {
...
 	} );
+
+	describe( 'help panel tooltip', () => {
+		it( 'should render highlight tooltip when trackedStartedTasks includes task, task is not completed, and tooltip not shown yet', () => {
+			const { queryByText } = render(
+				<ActivityPanel
+					requestingTaskListOptions={ false }
+					taskListComplete={ false }
+					taskListHidden={ false }
+					trackedStartedTasks={ [ 'payment', 'payment' ] }
```
jeffstieler, line 208: This doesn't seem to match the new format of being an object. Are these tests passing by accident?

louwie17: Good catch, turns out `tobeDefined` checks if it is not `undefined`.
I used `getByText` and `toBeInTheDocument` instead, which correctly fails if it isn't shown.

Also updated the data format to match the new layout.

jeffstieler (approved):
I think this is in a good spot to merge. If @pmcpinto wants to change the tracking data, we can handle that in a follow up. Thanks for sticking with this drawn out review! 🚢  it!

pmcpinto:
> I'm not sure what the 1,2,..5 represent in that structure. Can you provide a full event example without placeholders?

That would be the number of visits: 1 visit, 2 visits and so on

> You can see the current structure that @louwie17 implemented here:

What's the difference between this count and the event count that we get on Tracks for any event?

jeffstieler:
> That would be the number of visits: 1 visit, 2 visits and so on

Oh I see the source of my confusion now - I was describing the data of a single event while you were describing how you'd be using the data once collected. 😌 

> What's the difference between this count and the event count that we get on Tracks for any event?

You know, I momentarily forgot that we'd be able to get an event count per unique user on the Tracks side. It's likely that this `visit_count` prop provides no extra value at all! 😅

pmcpinto:
Ah, now it's clear 😃 In that case, it won't be necessary to have the visit_count prop :)

PR #5794, opened 2020-12-02
Home Screen - modified tasks extensibility

octaedro:
Fixes #5242

This PR adds the modifications required for the task list extensibility.
 
* It should allow 3pds to create setup tasks for their extensions with the following items: task name (with link to an extension page), time to complete the tasks, and a dismiss feature
* If a merchant completes the task list and then installs an extension that has a setup task associated, then the task list will be displayed again with that extension.

### Accessibility

-   [x] I've tested using only a keyboard (no mouse)
-   [ ] I've tested using a screen reader
-   [x] All animations respect [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion)
-   [x] All text has [at least a 4.5 color contrast with its background](https://webaim.org/resources/contrastchecker/)

### Screenshots
screenshot-one wordpress test-2020 12 02-17_01_48

### Detailed test instructions:
1. Run `npm run example -- --ext=add-task`.
2. Activate the plugin `WooCommerce Admin Add Task Example`.
3. Be sure the setup task list is not hidden
```
UPDATE `wp_options` SET `option_value` = 'no' WHERE `wp_options`.`option_name` = 'woocommerce_task_list_hidden';
```
4. Go to the `Home` screen.
5. Verify the task list `Extensions setup` is visible.
6. Try the `Dismiss` button and undo the dismission. Verify everything is working as expected.

7. Hide the setup list (`Finish store setup`) and verify the other list is still there.
8. Hide the `Extensions setup` list and verify that it's not visible anymore. 
screenshot-one wordpress test-2020 12 02-18_49_37
9. Try adding another task * and verify the list now is visible again.
10. Confirm the callback method `onDismiss` is working correctly when a task is dismissed. To do it just take a look at the browser's console after dismissing a task, the message: `The task 'example-[number]' was dismissed
` should be visible.

_* I did that in the branch `add/5242_test` so you can check out that branch and run `npm run example -- --ext=add-task-two`, and activate the plugin `WooCommerce Admin Add Task Example Two`. The task button `Mark as complete` should mark all these tasks as completed after being pressed._

Issue #5242, opened 2020-10-01
Home Screen - tasks extensibility


pmcpinto:
**User story:**
As a 3PD
I want to easily extend the home screen task feature
So I can incorporate recurrent actions in my extension that require immediate attention

**Description:** our goal is to allow 3PDs to create home screen tasks for their extensions, in order to have a cohesive user experience. Ideally we should allow them to implement two different types of tasks:
- Setup tasks: that will be incorporated in the setup task list
- Recurring tasks: following the same approach that we have for orders, reviews and stock

We have a [wc-admin task example](https://github.com/woocommerce/woocommerce-admin/blob/master/docs/examples/extensions/add-task/woocommerce-admin-add-task-example.php), but I'm not sure if it can be reused for this.

**Acceptance criteria:**
- [ ] Allow 3pds to create setup tasks for their extensions with the following items: task name (with link to an extension page), time to complete the tasks and dismiss feature
- [ ] If a merchant completes the task list and then installs and extension that has a setup task associated, then we should display the task list again with that extension
![Home new extension.png](https://images.zenhubusercontent.com/5d84fe8e1201de0001b6ab61/1ea9cd21-5a6d-4f6b-b1ad-8d8d062adca5)
- [ ] Allow 3pds to create recurring tasks and keep the limit that we have for our tasks: max 5 items per task

**Event tracking:**
NA

**Additional links:**
[P2 with links to the Figma file](pbIJXs-ma-p2)

samueljseay:
I think this is possibly several tasks and I'd like to break it up further, but first I want to open a discussion about a possible approach for this.

Right now we have setup management tasks (and checks for if they're complete) defined client side. We don't have a good story for doing client side integration with WooCommerce Admin right now, so it makes sense to move the setup task list to PHP.

I'm proposing a task to move the existing setup management into a server side API and remove that from the client. This would be a good opportunity to simplify the client side including being able to query if the task list is complete from the API, instead of our current rather flaky client side code that checks list completion status.

Once we create an API we can use to query setup tasks, it will be a simpler next step to introduce the creating of new setup tasks via PHP. We won't have to merge our existing client side implementation with the backend one which will likely lead to more subtle issues and bugs imho.

@jeffstieler @octaedro keen to hear what you think of this approach.

One question for @pmcpinto , how relevant is tracking of completing the entire task list after these features are completed? It feels like the task list will be very open ended, is that tracking event still relevant?

pmcpinto:
[...]

PR #5833, opened 2020-12-07
Home Screen - Small refactor to tasks extensibility 

octaedro:
Related issue #5242
This PR adds a few modifications [related to the changes](https://github.com/woocommerce/woocommerce-admin/pull/5794) to the task extensibility. 

The modifications in this PR are:
* a small refactor related to the list handling
* a method to [unregister the extended tasks](https://github.com/woocommerce/woocommerce-admin/pull/5794#pullrequestreview-544454521)
* a default extended task type (`extension`)

### Accessibility

### Screenshots
screenshot-one wordpress test-2020 12 02-17_01_48

0. Checkout `add/5242_update`.
1. Run `npm run example -- --ext=add-task`.
2. Activate the plugin `WooCommerce Admin Add Task Example`.
3. Be sure the setup task list is not hidden
```
UPDATE `wp_options` SET `option_value` = 'no' WHERE `wp_options`.`option_name` = 'woocommerce_task_list_hidden';
```
4. Go to the `Home` screen.
5. Verify the task list `Extensions setup` is visible.
6. Verify the task list was added to `woocommerce_extended_task_list_items`. You can do it with a sentence like this:
```
SELECT * FROM `wp_options` WHERE `option_name` = 'woocommerce_extended_task_list_items'
```
7. Deactivate the plugin: `WooCommerce Admin Add Task Example`
8. Verify the task list was removed from `woocommerce_extended_task_list_items`.

### DIFF
--- a/client/task-list/index.js
+++ b/client/task-list/index.js
@@ -60,14 +60,6 @@ export class TaskDashboard extends Component {
 		} );
 	}
 
-	groupBy( array, key ) {
-		return array.reduce( ( result, currentValue ) => {
-			( result[ currentValue[ key ] ] =
-				result[ currentValue[ key ] ] || [] ).push( currentValue );
-			return result;
-		}, {} );
-	}
-
 	toggleCartModal() {
 		const { isCartModalOpen } = this.state;
 
@@ -90,34 +82,29 @@ export class TaskDashboard extends Component {
 		} = this.props;
 		const { isCartModalOpen } = this.state;
 		const allTasks = this.getAllTasks();
-		const { extension: extensionTasks, setup: setupTasks } = this.groupBy(
-			allTasks,
-			'type'
-		);
+		const { extension: extensionTasks, setup: setupTasks } = allTasks;
 
 		return (
 			<>
 				{ setupTasks && ! isSetupTaskListHidden && (
 					<TaskList
-						allTasks={ allTasks }
 						dismissedTasks={ dismissedTasks }
 						isTaskListComplete={ isTaskListComplete }
 						isExtended={ false }
 						query={ query }
-						specificTasks={ setupTasks }
+						tasks={ allTasks }
 						trackedCompletedTasks={ trackedCompletedTasks }
 					/>
 				) }
 				{ extensionTasks && ! isExtendedTaskListHidden && (
 					<TaskList
-						allTasks={ allTasks }
 						dismissedTasks={ dismissedTasks }
 						isExtendedTaskListComplete={
 							isExtendedTaskListComplete
 						}
 						isExtended={ true }
 						query={ query }
-						specificTasks={ extensionTasks }
+						tasks={ allTasks }
 						trackedCompletedTasks={ trackedCompletedTasks }
 					/>
 				) }
--- a/client/task-list/list.js
+++ b/client/task-list/list.js
@@ -54,6 +54,24 @@ export class TaskList extends Component {
 		this.possiblyTrackCompletedTasks();
 	}
 
+	getUngroupedTasks() {
+		const { tasks: groupedTasks } = this.props;
+		return [].concat(
+			...Object.keys( groupedTasks ).map(
+				( taskType ) => groupedTasks[ taskType ]
+			)
+		);
+	}
+
+	getSpecificTasks() {
+		const { isExtended, tasks: groupedTasks } = this.props;
+		const { extension, setup } = groupedTasks;
+		if ( isExtended ) {
+			return extension;
+		}
+		return setup;
+	}
+
 	possiblyCompleteTaskList() {
 		const {
 			isExtended,
@@ -61,8 +79,8 @@ export class TaskList extends Component {
 			isExtendedTaskListComplete,
 			updateOptions,
 		} = this.props;
-		const isSetupTaskListInComplete = ! isExtended && ! isTaskListComplete;
-		const isExtendedTaskListInComplete =
+		const isSetupTaskListIncomplete = ! isExtended && ! isTaskListComplete;
+		const isExtendedTaskListIncomplete =
 			isExtended && ! isExtendedTaskListComplete;
 		const taskListToComplete = isExtended
 			? { woocommerce_extended_task_list_complete: 'yes' }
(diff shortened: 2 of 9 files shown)


client/task-list/list.js
```diff
@@ -54,15 +54,33 @@ export class TaskList extends Component {
 		this.possiblyTrackCompletedTasks();
 	}
 
+	getUngroupedTasks() {
+		const { tasks: groupedTasks } = this.props;
+		return [].concat(
```
samueljseay, line 59: You can just do:

```
return [
  ...Object.keys( groupedTasks ).map(
				( taskType ) => groupedTasks[ taskType ]
			)
];
```

src/Features/OnboardingTasks.php
```diff
@@ -394,6 +395,22 @@ public static function add_extended_task_list_item( $new_task_name ) {
 		}
 	}
 
+	/**
+	 * Removes item to the extended task list.
```
samueljseay, line 399: Removes item *from* the extended task list.

client/task-list/tasks.js
```diff
@@ -39,6 +39,17 @@ export function recordTaskViewEvent(
 	} );
 }
 
+const groupBy = ( array, key ) => {
```
samueljseay, line 42: Ideally we should have a basic set of tests for this. Also it's a very generic collections method that we should share under lib or something.

samueljseay (changes requested):
Just some minor suggestions from me, otherwise looks good.

src/Features/OnboardingTasks.php
octaedro: Thanks @samueljseay, you're right. This was modified in the commit ddd0fe6

client/task-list/list.js
octaedro: I'm not sure about that. The `groupedTasks` is an object with two lists inside, one contains the setup tasks and the other one the extended tasks. I need the `concat` in order to merge the two lists. Otherwise, the result would be an array with 2 arrays with the split task lists inside.

client/task-list/tasks.js
octaedro: Good idea! I moved the method to `lib/lists` and added some tests in the commit 600651c

octaedro:
Thank you @samueljseay for your review. I addressed the suggested changes, could you give this PR another look?

client/task-list/list.js
samueljseay: @octaedro gotcha I missed the structure there!

then let me suggest another way to simplify :D (not stressed about the change but this can be simpler)

```
return [...Object.values(groupedTasks)].flat()
```

samueljseay (approved):
Just one more little suggestion around the array processing. pre-approving though! Great stuff.

client/task-list/list.js
octaedro: Nice catch @samueljseay !
That was my first option but I realized later that the project should work with [Node >= 8](https://github.com/woocommerce/woocommerce-admin/blob/main/package.json#L223) and `Array.flat` only will work with [Node > 10.23.0](https://node.green/#ES2019-features-Array-prototype--flat--flatMap--Array-prototype-flat), so the tests with lower versions of Node would fail.

samueljseay: @octaedro this code doesn't need to work with Node 8. Any tests for this would be run in a browser like environment. For the browser Array.flat is polyfilled by corejs

client/task-list/list.js
```diff
@@ -54,15 +54,29 @@ export class TaskList extends Component {
 		this.possiblyTrackCompletedTasks();
 	}
 
+	getUngroupedTasks() {
+		const { tasks: groupedTasks } = this.props;
+		return Object.values( groupedTasks ).flat();
+	}
+
+	getSpecificTasks() {
```
jeffstieler, line 62: This seems strange to me.. why are we passing more tasks to a `<TaskList/>` than it's supposed to render?

client/task-list/index.js
```diff
@@ -90,34 +82,29 @@ export class TaskDashboard extends Component {
...
 					/>
 				) }
 				{ extensionTasks && ! isExtendedTaskListHidden && (
 					<TaskList
-						allTasks={ allTasks }
 						dismissedTasks={ dismissedTasks }
 						isExtendedTaskListComplete={
 							isExtendedTaskListComplete
 						}
 						isExtended={ true }
```
jeffstieler, line 105: I know these weren't introduced in this PR, but why does the `<TaskList/>` need to know about the "extended" task list concept? Can't it just receive a `isComplete` and `tasks` props?

docs/examples/extensions/add-task/woocommerce-admin-add-task-example.php
```diff
@@ -41,3 +41,12 @@ function add_task_register_script() {
 	do_action( 'add_woocommerce_extended_task_list_item', 'woocommerce_admin_add_task_example_name' );
 }
 add_action( 'admin_enqueue_scripts', 'add_task_register_script' );
+
+/**
+ * Unregister task list item.
+ */
+function pluginprefix_deactivate() {
+	do_action( 'remove_woocommerce_extended_task_list_item', 'woocommerce_admin_add_task_example_name' );
```
jeffstieler, line 49: This API is in reverse. Our plugin should fire the `do_action()` or `apply_filters()` and other plugins looking to make use of those hook should attach with `add_action()` or `add_filter()`.

src/Features/OnboardingTasks.php
```diff
@@ -47,6 +47,7 @@ public function __construct() {
 		add_action( 'add_option_woocommerce_task_list_tracked_completed_tasks', array( $this, 'track_task_completion' ), 10, 2 );
 		add_action( 'update_option_woocommerce_task_list_tracked_completed_tasks', array( $this, 'track_task_completion' ), 10, 2 );
 		add_action( 'add_woocommerce_extended_task_list_item', array( $this, 'add_extended_task_list_item' ), 10, 2 );
+		add_action( 'remove_woocommerce_extended_task_list_item', array( $this, 'remove_extended_task_list_item' ), 10, 2 );
```
jeffstieler, line 50: Again, this API is in reverse. It will likely need to be replaced by a filtered array instead of an action.

jeffstieler (commented):
@octaedro (cc: @samueljseay)

Something that jumped out at @joshuatf and I is the use of the WordPress Hooks API is in reverse here.

Since our plugin is the originator of the hook, it should be using either `do_action()` or `apply_filters()`. The use of `add_action()` and `add_filter()` are in the consumer side of the hooks API.

https://developer.wordpress.org/plugins/hooks/custom-hooks/

PR #5834, opened 2020-12-07
Revert "Convert before and after dates to UTC in Analytics queries"

moon0326:
Reverts woocommerce/woocommerce-admin#5819


jeffstieler (approved):
:shipit:

PR #5837, opened 2020-12-07
Change the nav feedback note into a learn more note

### LINKED ISSUE #5813: Navigation: Inbox message to inform users of the Nav
Create an inbox message to inform users of the new Navigation.

### Deliverables

* Create a notification
* Apply correct copy (@pmcpinto for wording)
* Add only for eComm plan users
* Decide if this should live in wc-calypso-bridge

joshuatf:
Fixes #5813

Removes the nav feedback note in favor of the learn more note.

2 questions before we can get this mereged (/cc @pmcpinto )
1. What URL would we like to use for the learn more link?
1. Do we want to remove the nav feedback follow-up note or keep it in place?

1. Enable the nav feature before checking out this branch.
1. Run `wc_admin_daily` cron job.
1. Note the note.
1. Delete the note.
1. Trying toggling the nav feature off and back on.
1. Note the note.


src/Notes/NavigationLearnMore.php
```diff
@@ -0,0 +1,56 @@
+<?php
+/**
+ * WooCommerce Admin Navigation Feature Feedback
```
jeffstieler, line 3: ©️ 🍝

jeffstieler (approved):
> * Note the note.
> ...
> * Note the note.

Noted. 😂 

This tested well for me - pre-approving pending any copy/link changes from @pmcpinto.

joshuatf:
@pmcpinto Do we have a link we'd like to use for this note?

pmcpinto:
> @pmcpinto Do we have a link we'd like to use for this note?

@timmyc is working on that document. Perhaps we can just create the link where the new content is going to be posted so we can merge this PR. Timmy, what do you think?

timmyc:
> @timmyc is working on that document. Perhaps we can just create the link where the new content is going to be posted so we can merge this PR. Timmy, what do you think?

So this document will be for all users of the new nav, and not specifically ecomm/wpcom centric correct? If so, then I think we can use the existing structure of doc pages ( i.e. https://docs.woocommerce.com/document/woocommerce-setup-wizard/ ) and say that we can have a nav doc at https://docs.woocommerce.com/document/woocommerce-navigation/

pmcpinto:
> So this document will be for all users of the new nav, and not specifically ecomm/wpcom centric correct?

Nope, this one will be for ecomm users. Similar to what you created for the sunset store initiative.. Make sense?

psealock:
>Nope, this one will be for ecomm users.

In this case, this code belongs in wc-calypso-bridge.

psealock:
Just thinking about this a bit more. If I understand correctly, the note is still OK to be in this repo. The linked document will be eComm user centric, right? If not, this code can still be in this repo because it will only show to users that have it turned on, which is only eComm users in this case.

timmyc:
+1 on either having the note live in wc-calypso-bridge _or_ using the remote inbox system for this instead and leverage the [ecom plan rule](https://github.com/woocommerce/woocommerce-admin/blob/main/src/RemoteInboxNotifications/IsEcommerceRuleProcessor.php)

pmcpinto:
> +1 on either having the note live in wc-calypso-bridge or using the remote inbox system for this instead and leverage the ecom plan rule

Just a heads up that we also aim to have a similar note once we release the nav enabled by default for new core users

joshuatf:
Closing this PR since this will be remote or live in the bridge.  https://github.com/Automattic/wc-calypso-bridge/pull/628

Issue #5560, opened 2020-11-09
Variations Report: Not all attributes available in advanced filters


psealock:
Some attributes are not available when using advanced filters.

I noticed this when using Woo's sample products, but went ahead and created my own variable product from scratch and noticed the same thing.

wc-admin [1.7.0-beta.1 ](https://github.com/woocommerce/woocommerce-admin/releases/tag/v1.7.0-beta.1)
woocommerce Version 4.6.2
Jurassic Ninja site

1. Create an order using Woo's sample product data variable Hoodie `/product/hoodie/?attribute_pa_color=green&attribute_logo=No`.
1. Confirm the Hoodie has variations "Logo" and "Color" in the Edit Product screen.
![Screen Shot 2020-11-09 at 2 41 49 PM](https://user-images.githubusercontent.com/1922453/98492226-0dff2680-229c-11eb-8fe1-4410569db953.png)
2. See the data from the purchase available in the Variations Report.
3. Advanced Filters > Add a filter > Attribute
4. See the "logo" attribute not available in the dropdown.
![Screen Shot 2020-11-09 at 2 41 36 PM](https://user-images.githubusercontent.com/1922453/98492319-5c142a00-229c-11eb-826f-d00feff8b732.png)

jeffstieler:
Thanks for creating this issue @psealock! It had only been previously captured here: https://github.com/woocommerce/woocommerce-admin/issues/5384#issuecomment-709280308

timmyc:
@jeffstieler do you have an ballpark on the level of effort to include non-global attributes? Seems like that could really be a non-performant query.

jeffstieler:
> @jeffstieler do you have an ballpark on the level of effort to include non-global attributes? Seems like that could really be a non-performant query.

I haven't looked into it too deeply, but from what I recall it would require querying/parsing the `meta_value`s for all `_product_attributes` meta keys - so yeah... the potential for poor performance is quite high.

timmyc:
Thanks for confirming my suspicions a bit. Based upon that, I'm going to clear this issue from the milestone. I suppose one thing we should do is make sure this known issue is part of the [Variations Report Doc](https://github.com/woocommerce/woocommerce-admin/blob/main/docs/woocommerce.com/analytics-variations-report.md) and also include this as a note to Happiness and maybe call it out in the 4.8 release post too where the new report will likely be called out as a new feature.

jeffstieler:
[...]

PR #5840, opened 2020-12-07
Add Product Attributes Endpoint

jeffstieler:
Partially implements #5560.

This PR enhances the `/products/attributes` and `/products/attributes/[term]/terms` endpoints to include custom (non-global, not taxonomy-backed) attributes.

A follow up PR will modify the Attribute report filter to use these endpoints for filter value selection.

- Ensure your store has some global Product Attributes defined
- Create a variation that uses some custom attributes
- Search for your custom attribute using `/wc-analytics/products/attributes/?search=`
- Verify the custom attribute is in the results (and any taxonomy backed attributes that match the query)
- Retrieve all terms using `/wc-analytics/products/attributes/[slug]/terms` (substitute the slug of your custom term)
- Verify all terms are returned

Enhancement: Support Custom Product Attributes in REST API

### DIFF
--- a/src/API/Init.php
+++ b/src/API/Init.php
@@ -63,6 +63,7 @@ class Init {
 			'Automattic\WooCommerce\Admin\API\Options',
 			'Automattic\WooCommerce\Admin\API\Orders',
 			'Automattic\WooCommerce\Admin\API\Products',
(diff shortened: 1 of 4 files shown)


becdetat (approved):
This works great (once I figured out how to hack the mainframe and do a REST API call) and the code is all good :-).

:shipit:

PR #5842, opened 2020-12-08
Fix/5831 nav add plugin settings

### LINKED ISSUE #5831: Navigation: Unable to add a Settings item
Discovered in https://github.com/woocommerce/woocommerce-bookings/pull/3001, A Settings item does not show up because its discarded in the `getMenuItemsByCategory` function.

### To reproduce

Add an item in Settings:

```php
Menu::add_plugin_item(
	array(
		'id'            => 'woocommerce-bookings-settings',
		'title'         => __( 'Bookings', 'woocommerce-bookings' ),
		'capability'    => 'read_global_availability',
		'url'           => 'edit.php?post_type=wc_booking&page=wc_bookings_settings',
		'parent'        => 'woocommerce-settings',
	)
);
```
This results in an item in `wcNavigation` with `menuId` of `plugins`.

Screen Shot 2020-12-07 at 3 38 07 PM

`getMenuItemsByCategory` filters out items belonging to a parent that do not also have the same `menuId`. In this case, Settings has a `menuId` of `secondary`. This seems weird to me. A comment here would be useful.

https://github.com/woocommerce/woocommerce-admin/blob/d591271d19e18eceaf760bf8467dc2a03db9d603/client/navigation/components/container/index.js#L87-L93

### Deliverables

* Adding an item with `'parent' => 'woocommerce-settings',` places an item in the Settings nav
* Settings items have no `menuId` separation and should all be in the same section

joelclimbsthings:
Fixes #5831

This is addressing the inability to add a Settings item to the nav via the `add_plugin_item()`. This is disallowed as that hook intentionally only supports adding items to the "Plugins" menu. Since the use case of plugins adding a "settings" page is a common one, this PR adds a similar hook that clearly allows for adding to Settings while still walling off unexpected behavior.

Notes:
- The `menuId` option is discarded for this hook, as it should only add to the `Settings` category in the `secondary` menu.
- The `parent` argument is also discarded, as we're only supporting singular items, not categories, under Settings.
- Right now if a plugin provides either of those two arguments, a warning is printed out via `error_log()`. If this is thought to be too noisy we can remove this bit.
- If we decide to go this route I'll also add notes for this function to the documentation.

### Screenshots

image

- Check out branch
- Attempt to add an item to the Settings like this:
```
\Automattic\WooCommerce\Admin\Features\Navigation\Menu::add_plugin_setting_item(
	array(
		'id'     => 'example-item-settings',
		'title'  => 'Example Item Settings',
		'url'    => '/notreallythere/settings',
	)
);
```
You should see the page appear under the _Settings_ menu as expected.

- Add a dummy `parent` argument to that call, and refresh the page. You should still see the option present under _Settings_, but it should print out a warning to the `error_log`

- Do the same with `menuId`.  You should see the same behavior, while the item still displays as it should under _Settings_.

### DIFF
--- a/src/Features/Navigation/Menu.php
+++ b/src/Features/Navigation/Menu.php
@@ -330,14 +330,14 @@ class Menu {
 	 *      'migrate'    => (bool) Whether or not to hide the item in the wp admin menu.
 	 *    ).
 	 */
-	public static function add_plugin_setting_item( $args ) {
+	public static function add_setting_item( $args ) {
 		unset( $args['order'] );
 
 		if ( isset( $args['parent'] ) || isset( $args['menuId'] ) ) {
 			error_log(  // phpcs:ignore
 				sprintf(
 					/* translators: 1: Duplicate menu item path. */
-					esc_html__( 'The item ID %1$s attempted to register using an invalid option. The arguments `menuId` and `parent` are not allowed for add_plugin_setting_item()', 'woocommerce-admin' ),
+					esc_html__( 'The item ID %1$s attempted to register using an invalid option. The arguments `menuId` and `parent` are not allowed for add_setting_item()', 'woocommerce-admin' ),
 					'`' . $args['id'] . '`'
 				)
 			);
--- a/src/Features/Navigation/Menu.php
+++ b/src/Features/Navigation/Menu.php
@@ -333,13 +333,23 @@ class Menu {
 	public static function add_plugin_setting_item( $args ) {
 		unset( $args['order'] );
 
-			$item_args = array_merge(
-				$args,
-				array(
-					'menuId' => 'secondary',
-					'parent' => 'woocommerce-settings',
+		if ( isset( $args['parent'] ) || isset( $args['menuId'] ) ) {
+			error_log(  // phpcs:ignore
+				sprintf(
+					/* translators: 1: Duplicate menu item path. */
+					esc_html__( 'The item ID %1$s attempted to register using an invalid option. The arguments `menuId` and `parent` are not allowed for add_plugin_setting_item()', 'woocommerce-admin' ),
+					'`' . $args['id'] . '`'
 				)
 			);
+		}
+
+		$item_args = array_merge(
+			$args,
+			array(
+				'menuId' => 'secondary',
+				'parent' => 'woocommerce-settings',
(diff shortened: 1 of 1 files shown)


src/Features/Navigation/Menu.php
```diff
@@ -251,72 +251,111 @@ public static function get_item_menu_id( $item ) {
 	}
 
 	/**
-	 * Adds a plugin item.
+	 * Adds a plugin category.
```
joelclimbsthings, line 254: You can disregard these single line changes; they're just appearing since I reordered the functions; no actual code changes

src/Features/Navigation/Menu.php
```diff
@@ -251,72 +251,111 @@ public static function get_item_menu_id( $item ) {
...
 		if ( 'plugins' !== $menu_id ) {
 			return;
 		}
 
-		self::add_category( $category_args );
+		self::add_item( $item_args );
+	}
+
+	/**
+	 * Adds a plugin setting item.
```
joelclimbsthings, line 322: This function is only actual code change.

src/Features/Navigation/Menu.php
```diff
@@ -251,72 +251,111 @@ public static function get_item_menu_id( $item ) {
...
+	 * @param array $args Array containing the necessary arguments.
+	 *    $args = array(
+	 *      'id'         => (string) The unique ID of the menu item. Required.
+	 *      'title'      => (string) Title of the menu item. Required.
+	 *      'capability' => (string) Capability to view this menu item.
+	 *      'url'        => (string) URL or callback to be used. Required.
+	 *      'migrate'    => (bool) Whether or not to hide the item in the wp admin menu.
+	 *    ).
+	 */
+	public static function add_plugin_setting_item( $args ) {
```
psealock, line 333: `add_setting_item` ?

psealock (commented):
Code looks good and I test this out with Bookings. Everything working as it should.

psealock (approved):
This is working well for me, I changed the name to `add_setting_item`. I'm going ahead with the merge so we can get this into 1.8.

@joelclimbsthings you mentioned a follow up to update the docs. Now that I'm merging, it would be good to do that in the next one.

PR #5845, opened 2020-12-08
Only show card placeholder for item that is updating, not the whole list

louwie17:
Fixes #5706 

Remove the use of the loading placeholder on every item when the list is being updated. Also make sure the `Manage all reviews` disappears when no items are displayed. Both of these changes make the list updates look a lot smoother (see the difference between gif's).
Had to update the update reviews reducer action as well, to not overwrite a review if we are only retrieving an `id`.

### Screenshots

With stutter (old):
update-review-with-stutter

Without stutter (new):
update-review-without-stutter

- Create store with products and reviews enabled
- Add a couple reviews to a product (in incognito window, so it needs to be reviewed)
- Go to WC home screen, and open the review panel
- Update a review (approve, delete, or mark as spam), it should update correctly and not look jittery.

### DIFF
--- a/client/homescreen/activity-panel/reviews/index.js
+++ b/client/homescreen/activity-panel/reviews/index.js
@@ -149,7 +149,7 @@ class ReviewsPanel extends Component {
 				review._embedded.up[ 0 ] ) ||
 			null;
 
-		if ( review.isUpdating || this.props.isRequesting ) {
+		if ( review.isUpdating ) {
 			return (
 				<ActivityCardPlaceholder
 					key={ review.id }
@@ -160,7 +160,7 @@ class ReviewsPanel extends Component {
 				/>
 			);
 		}
-		if ( isNull( product ) ) {
+		if ( isNull( product ) || review.status !== reviewsQuery.status ) {
 			return null;
 		}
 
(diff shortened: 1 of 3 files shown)


packages/data/src/reviews/reducer.js
```diff
@@ -23,9 +23,16 @@ const reducer = (
...
 		case TYPES.UPDATE_REVIEWS:
 			const ids = [];
+			const justFields =
+				query && query._fields && query._fields.length > 0;
 			const nextReviews = reviews.reduce( ( result, review ) => {
 				ids.push( review.id );
-				result[ review.id ] = review;
+				result[ review.id ] = {
+					...( justFields && state.data[ review.id ]
+						? state.data[ review.id ]
```
louwie17, line 32: If we are querying for a couple fields (like `id`), we merge with the old review object if it exists.

client/homescreen/activity-panel/reviews/index.js
```diff
@@ -149,7 +149,7 @@ class ReviewsPanel extends Component {
 				review._embedded.up[ 0 ] ) ||
 			null;
 
-		if ( review.isUpdating || this.props.isRequesting ) {
+		if ( review.isUpdating ) {
 			return (
 				<ActivityCardPlaceholder
```
louwie17, line 154: Only show activity card placeholder when a review is updating.

packages/data/src/reviews/reducer.js
jeffstieler: Would it be safe to always merge?

jeffstieler (approved):
The code looks good and this tested well. I'm just wondering if it's safe to always merge review objects in the reducer? 🤔

PR #5855, opened 2020-12-09
Move collapsible to panels config, fixing orders panel

louwie17:
Instead of setting the collapsible boolean on each activity panels count, I moved it to the panels config, so it can easily be toggled for different panels.
This fixes the 🎉 not showing up for the Orders panel.

### Screenshots

Screen Shot 2020-12-09 at 4 03 18 PM

- Make sure you have a store with at-least one order
- Complete this order
- Navigate to the WC homescreen
- The orders panel should be collapsed with a 🎉  icon and saying - **You’ve fulfilled all your orders**

### DIFF
--- a/client/homescreen/activity-panel/index.js
+++ b/client/homescreen/activity-panel/index.js
@@ -52,6 +52,7 @@ export const ActivityPanel = () => {
 						initialOpen,
 						panel,
 						title,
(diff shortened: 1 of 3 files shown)


jeffstieler (approved):
:shipit:

PR #5856, opened 2020-12-09
Update @wordpress/eslint-plugin dependency to latest version.

nerrad:
This pull request updates the `@wordpress/eslint-plugin` dependency for the `@woocommerce/eslint-plugin` to the latest version (`8.0.0`) from `7.1.0`. 

With these changes:

- A new "eslint-plugin-import" dependency on the main package file due to changes in eslint.
- moving `prettier` and `@wordpress/prettier-config` dependencies from main package.json to the `@woocommerce/eslint-plugin` package.json. This reduces the imports needed by consumers of `@woocommerce/eslint-plugin` hopefully.
- a jsdoc lint plugin update considers `typescript` style jsdocs as errors by default. So I added a setting configuration update to the `@woocommerce/eslint-plugin` config to set jsdoc style to `typescript`. 
- A bunch of new linting rules came with the update so I temporarily converted errors to warnings for followups after merge. This allows getting the package update released sooner and followups can make changes carefully.

## To Test

- There should be no linting errors reported as a result of these changes.
- Verify that eslint fix and prettier format fix still work cli and in your IDE (both worked great for me - I use VSCode).


nerrad:
I'm not sure of your process, does the `@woocommerce/eslint-plugin` package version get bumped in this PR or when you go to release?

samueljseay:
> I'm not sure of your process, does the @woocommerce/eslint-plugin package version get bumped in this PR or when you go to release?

Bumping it in this PR is all good 👍

jeffstieler:
> Verify that eslint fix and prettier format fix still work cli and in your IDE

Code formatting (in IDE) via prettier broke for me using this branch - does it require updating settings in VS Code?

nerrad:
> Code formatting (in IDE) via prettier broke for me using this branch - does it require updating settings in VS Code?

Hmm 🤔 it didn't break for me and I ran `npm run lint:js` as well to make sure cli still worked as expected. One main reason for doing this update was to remove the need for a localized `.prettierrc.json` file. 

What's happening with the formatting in your case?

samueljseay:
@nerrad It breaks for me too. 

When I look at prettier's log its using the .editorconfig for settings. and they're incorrect so prettier still works but doesn't format according to wp standard

I tried removing .editorconfig and that doesn't work either, it falls back to VSCode defaults.

How would this work though? By that I mean, how would the prettier module find the right .prettierrc? Wouldn't we still need something like `extends @wordpress/prettier-config` in a local prettierrc?

nerrad:
I think I've found out why it's working for me - here's my VS Code settings:

```
"editor.codeActionsOnSave": {
    "source.fixAll.eslint": true
  },
  "[javascript]": {
    "editor.formatOnPaste": true,
    // "editor.formatOnSave": true,
    "editor.defaultFormatter": "esbenp.prettier-vscode",
  },
```

So while the default Formatter is set to esbenp.prettier-vscode, the save action is using eslint fix.

> How would this work though? By that I mean, how would the prettier module find the right .prettierrc? Wouldn't we still need something like extends @wordpress/prettier-config in a local prettierrc? (or pointing to a specific prettierrc thats tucked away in node_modules somewhere) or eslint would need to run prettier?

`@wordpress/eslint-plugin/recommended` supposedly [implements prettier config](https://github.com/WordPress/gutenberg/blob/master/packages/eslint-plugin/configs/recommended.js) and should be loading it from the `@wordpress/prettier-config` by default while allowing for local override. However, there _may_ be an artifact of how `@woocommerce/eslint-plugin` is wrapping `@wordpress/eslint-plugin` that is causing that config search to fail? I haven't investigated fully on this but that could be the problem here.

Alternatively, in this case it may not be that the eslint configuration isn't working for linting execution but more an issue with how VSCode is trying to resolve the prettier config and it isn't picking it up from the unique way this package works.

Now that I know why things were working for me and not you folks, I can likely try debugging this on my own at some point (won't be until next week sometime likely).

nerrad:
Hey folks, sorry for the delay, I've revisited because there are number of changes available now with the `@wordpress/eslint-plugin` (plus prettier updates) that would be useful to get out with the `@woocommerce/eslint-plugin` package update. The `.prettierrc.js` file _is_ needed in the package root after all. AFAICT things should work okay now in your IDE with testing.

psealock (approved):
Thanks @nerrad ! Changes look good to me, my IDE is happy, my prettier is also happy, and tests pass.

>I'm not sure of your process, does the @woocommerce/eslint-plugin package version get bumped in this PR or when you go to release?

Feel free to bump the version in the package's `package.json` and `changelog.md` and I can make a release. `1.1.0` makes sense, yeah?

nerrad:
1.1.0-beta ? or just 1.1.0? Would you be able to go ahead and make the necessary changes?

psealock:
https://github.com/woocommerce/woocommerce-admin/pull/5856/commits/558e99eda8cd6e190ced1f8923dc9cc78fb9b514 bumps version in preparation for package to be published.

psealock:
Thanks @nerrad, new package published https://www.npmjs.com/package/@woocommerce/eslint-plugin

PR #5858, opened 2020-12-10
Fix issue where store age is 0 while isResolving for the option is returning false

becdetat:
Upgrading to WordPress 5.6 caused an issue where the CES snackbar was displayed twice for an action.

image

This seems to be because `isResolving` for the `woocommerce_admin_install_timestamp` option was returning false while `getOption` was returning zero, which rendered the `CustomerEffortScore` component, showing the snackbar. `getOption` later returning the correct timestamp again rendered the `CustomerEffortScore` component, this time correctly showing the snackbar again.

This PR adds an extra check that the timestamp is a valid value to the `resolving` value, which stops the first incorrect render.

- on a site with WordPress 5.6 installed
- add a new product
- check that the CES snackbar only appears once

### DIFF
--- a/client/customer-effort-score-tracks/customer-effort-score-tracks.js
+++ b/client/customer-effort-score-tracks/customer-effort-score-tracks.js
@@ -178,6 +178,19 @@ CustomerEffortScoreTracks.propTypes = {
 	createNotice: PropTypes.func,
 };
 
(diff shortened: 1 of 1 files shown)


moon0326 (approved):
Tested well and LGTM 👍

PR #5860, opened 2020-12-10
Using the router to implement persisted queries

### LINKED ISSUE #5684: Navigation: Make sure persisted queries are present on wc-admin links
Persisted queries are query parameters that remain when navigating from report to report. They include things like `period` so users can view reports from the same vantage point.

https://github.com/woocommerce/woocommerce-admin/blob/2a382d72b85ada88df28005f91910294909587c6/packages/navigation/src/index.js#L42-L51

In the old nav, we update the links so that those query parameters are present when they should be.

https://github.com/woocommerce/woocommerce-admin/blob/2a382d72b85ada88df28005f91910294909587c6/client/layout/controller.js#L214

The new Nav needs to also have those parameters present.

### Deliverables

* When looking at a report with any of the persisted queries in the url, any link to another report or overview page should also include those parameters.

joelclimbsthings:
Fixes #5684

Exploring the possibility of relocating the logic governing persisted queries to the router, primarily to address the fact that they don't function as is in the new nav. This does, however, change how this is controlled for the old navigation as well, which is intentional so that we don't have to duplicate logic. 

**Benefits**
- No DOM manipulation required
- Cleaner (somewhat less code required)
- Consolidated all logic governing persisted queries into `navigation` package; works with old and new navigation (and future iterations)

**Known Issues**
- ~~Right now if you navigate to an excluded page (ie Customers), and then back to a non-excluded page, the persisted queries will be lost (since they're only stored in the url for now). There are a number of ways we could address this if it's considered an issue.~~
- Need to fix/add tests (will do if we decide to go this route)

**Other Notes**
- ~~I've only implemented this for basic filters as a PoC, but it should be simple to add to advanced/etc filters if those are supported (need to explore this a bit more)~~ I don't believe this is a problem?

-  Check out branch
- Ensure that the new navigation is enabled.
- Go to an _Analytics_ report page and apply a different date range.
- Then navigate to a different report page (without a refresh), and ensure that the new date range persists.
- Disable the new navigation and do the same, ensuring that the old nav works as always.

### DIFF
--- a/client/layout/controller.js
+++ b/client/layout/controller.js
@@ -2,7 +2,7 @@
  * External dependencies
  */
 import { Component, Suspense, lazy } from '@wordpress/element';
-import { find, isEqual, omit } from 'lodash';
+import { find, isEqual, omit, last } from 'lodash';
 import { applyFilters } from '@wordpress/hooks';
 import { __ } from '@wordpress/i18n';
 import { getNewPath, getHistory } from '@woocommerce/navigation';
@@ -12,6 +12,7 @@ import { Spinner } from '@woocommerce/components';
  * Internal dependencies
  */
 import getReports from '../analytics/report/get-reports';
+import { isWCAdmin } from '../dashboard/utils';
 
 const AnalyticsReport = lazy( () =>
 	import( /* webpackChunkName: "analytics-report" */ '../analytics/report' )
@@ -189,7 +190,7 @@ export class Controller extends Component {
 	render() {
 		const { page, match, query } = this.props;
 		const { url, params } = match;
-
+		patchMenuLinks();
 		window.wpNavMenuClassChange( page, url );
 		return (
 			<Suspense fallback={ <Spinner /> }>
@@ -203,6 +204,24 @@ export class Controller extends Component {
 		);
 	}
 }
+
+// Add click handlers to wc-admin links in wp-admin menu to suppress page refresh when possible
+const patchMenuLinks = function () {
+	Array.from( document.querySelectorAll( '#adminmenu a' ) ).forEach(
+		( item ) => {
+			if ( ! isWCAdmin( item.href ) ) {
+				return;
+			}
+
+			item.onclick = ( e ) => {
+				e.preventDefault();
+				getHistory().push(
+					`admin.php?${ last( item.href.split( '?' ) ) }`
+				);
+			};
+		}
+	);
+};
 
 // When the route changes, we need to update wp-admin's menu with the correct section & current link
 window.wpNavMenuClassChange = function ( page, url ) {
--- a/client/layout/controller.js
+++ b/client/layout/controller.js
@@ -2,22 +2,16 @@
  * External dependencies
  */
 import { Component, Suspense, lazy } from '@wordpress/element';
-import { parse, stringify } from 'qs';
-import { find, isEqual, last, omit } from 'lodash';
+import { find, isEqual, omit } from 'lodash';
 import { applyFilters } from '@wordpress/hooks';
 import { __ } from '@wordpress/i18n';
-import {
-	getNewPath,
-	getPersistedQuery,
-	getHistory,
-} from '@woocommerce/navigation';
+import { getNewPath, getHistory } from '@woocommerce/navigation';
 import { Spinner } from '@woocommerce/components';
 
 /**
  * Internal dependencies
  */
 import getReports from '../analytics/report/get-reports';
-import { isWCAdmin } from '../dashboard/utils';
 
 const AnalyticsReport = lazy( () =>
 	import( /* webpackChunkName: "analytics-report" */ '../analytics/report' )
@@ -42,8 +36,6 @@ const MarketingOverview = lazy( () =>
 const ProfileWizard = lazy( () =>
 	import( /* webpackChunkName: "profile-wizard" */ '../profile-wizard' )
 );
-
-const TIME_EXCLUDED_SCREENS_FILTER = 'woocommerce_admin_time_excluded_screens';
 
 export const PAGES_FILTER = 'woocommerce_admin_pages_list';
 
@@ -198,7 +190,6 @@ export class Controller extends Component {
 		const { page, match, query } = this.props;
 		const { url, params } = match;
 
-		window.wpNavMenuUrlUpdate( query );
 		window.wpNavMenuClassChange( page, url );
 		return (
 			<Suspense fallback={ <Spinner /> }>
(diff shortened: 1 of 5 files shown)


packages/navigation/src/history.js
```diff
@@ -44,8 +49,17 @@ function getHistory() {
 			},
 			createHref: ( ...args ) =>
 				browserHistory.createHref.apply( browserHistory, args ),
-			push: ( ...args ) =>
-				browserHistory.push.apply( browserHistory, args ),
+			push( ...args ) {
+				const [ href ] = args;
```
joelclimbsthings, line 53: There _may_ be an edge case where `history.push` is called with different arguments, but I couldn't produce anything like that in testing; anyone know if this could be an issue?

client/layout/controller.js
```diff
@@ -213,52 +205,22 @@ export class Controller extends Component {
...
-		const href =
-			'admin.php?' +
-			stringify(
-				Object.assign( query, isExcludedScreen ? {} : nextQuery )
-			);
-
-		// Replace the href so you can see the url on hover.
-		item.href = href;
+// Add click handlers to wc-admin links in wp-admin menu to suppress page refresh when possible
+const patchMenuLinks = function () {
```
joelclimbsthings, line 209: I renamed this from `window.wpNavMenuUrlUpdate` but I'm worried this could result in a breaking change somewhere. A quick search on github shows that hook being called, but as far as I can tell it's only in forks of our codebase? (specifically where we were zeroing these out in settings).

client/layout/controller.js
```diff
@@ -213,52 +205,22 @@ export class Controller extends Component {
...
-		'customers',
-		'homescreen',
-	] );
-	const nextQuery = getPersistedQuery( query );
-
-	Array.from(
-		document.querySelectorAll( '#adminmenu a' )
-	).forEach( ( item ) => updateLinkHref( item, nextQuery, excludedScreens ) );
+			item.onclick = ( e ) => {
+				e.preventDefault();
```
psealock, line 217: I wonder if this interferes with the `onClick` on `NavigationItem` we use for Tracks?

psealock (commented):
Between modifying `history.push` and preventing the default on clicking a Nav item, I wonder if this opens ourselves up to future headaches?

I do like the idea of moving that logic so it's accessible to Navigation or other parts of the application. What do you think about using Context API to keep track of persisted queries?

client/layout/controller.js
joelclimbsthings: This part actually isn't new, just relocated from the now deleted `updateLinkHref()` function: https://github.com/woocommerce/woocommerce-admin/blob/67515d8aa6cca68c5ea3171077d4aa38502e52b1/client/layout/controller.js#L242

So I believe it shouldn't behave any differently for tracks/etc?

joelclimbsthings:
Thanks for looking it over @psealock !

> Between modifying `history.push` and preventing the default on clicking a Nav item, I wonder if this opens ourselves up to future headaches?

I mentioned in a comment, the `.preventDefault()` isn't actually new, so I don't think it should cause any new issues (although I may be missing some context somewhere).

> I do like the idea of moving that logic so it's accessible to Navigation or other parts of the application. What do you think about using Context API to keep track of persisted queries?

Yeah, that was my initial thought when diving into this for sure, and it's possible that that is still the best route.

My thought (although maybe this is an assumption) was that we definitely still want the persisted queries to appear in the `history`, since that will allow users to link to it and allow them to survive page refreshes, back btn behavior, etc.

That said, even if we use context to persist the parameters, we'll have to enter these into the history _somehow_. I haven't hacked around with history much before, but I'm inclined to think we would just `replace()` the state to include the parameters persisted through `Context`?

In the end a simple modification of the history seemed a more direct/clearer approach, although I definitely understand the reluctance to modify `history`. Then again, if we're just always replacing the state immediately, is there really a difference? It look liked we were already modifying the history operation to some degree, but I'm not sure how/why.

psealock:
>That said, even if we use context to persist the parameters, we'll have to enter these into the history somehow

Clicking on something that changes a persisted query param (like dates) already sticks around the history because we are using the `Link` component to push a new url to `history`.

https://github.com/woocommerce/woocommerce-admin/blob/19309ff0891b34f63efac07bcf3cfe7d7fede882/packages/components/src/link/index.js#L28

If I understand correctly, the challenge is updating the urls of the Nav when the user is already viewing a report with a persisted parameter that is something other than the default. If those are kept in Context, wouldn't a change in params force a refresh? Otherwise, is there a way to force a refresh on navigation maybe using `forceUpdate`? My understanding is that the persisted queries are included, but they aren't updating because the Navigation component isn't updating.

https://github.com/woocommerce/woocommerce-admin/blob/19309ff0891b34f63efac07bcf3cfe7d7fede882/client/layout/navigation.js#L51

joshuatf (commented):
Really great work on this, Joel!  This is testing pretty well for me.

@psealock brings up some good points.  I would also like to get to the bottom of why the navigation component added via `registerPlugin` is not working.  But this is a very clever workaround.

The only other difference I saw here is that store owners can't open new tabs using the persisted queries since those params are no longer in the navigation item urls.

packages/navigation/src/index.js
```diff
@@ -46,6 +48,39 @@ export const getPersistedQuery = ( query = navUtils.getQuery() ) => {
...
 
+/**
+ * Accepts a query, and returns a query that has updated persistent queries
+ * conditionally.
+ *
+ * @param {Object} query Query containing the parameters.
+ * @return {Object} Modified query object
+ */
+export const patchPersistedQueries = ( query ) => {
+	const excludedScreens = applyFilters( TIME_EXCLUDED_SCREENS_FILTER, [
```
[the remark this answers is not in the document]
joelclimbsthings: Done :+1:

packages/navigation/src/history.js
joelclimbsthings: I added the above argument condition just in case.

joelclimbsthings:
> 
> Clicking on something that changes a persisted query param (like dates) already sticks around the history because we are using the `Link` component to push a new url to `history`.

Ah, understood! Thanks @psealock 

> If I understand correctly, the challenge is updating the urls of the Nav when the user is already viewing a report with a persisted parameter that is something other than the default. If those are kept in Context, wouldn't a change in params force a refresh? Otherwise, is there a way to force a refresh on navigation maybe using `forceUpdate`? My understanding is that the persisted queries are included, but they aren't updating because the Navigation component isn't updating.

I actually just [added a PR yesterday](https://github.com/woocommerce/woocommerce-admin/issues/5684#issuecomment-745641528) as an alternate to this, which is more along these lines (although I didn't find the need for Context).

> I would also like to get to the bottom of why the navigation component added via registerPlugin is not working. But this is a very clever workaround.

Thanks @joshuatf . I added a simple alternate PR [here](https://github.com/woocommerce/woocommerce-admin/pull/5883) that just addresses the issue with the navigation component not re-rendering as expected.

>The only other difference I saw here is that store owners can't open new tabs using the persisted queries since those params are no longer in the navigation item urls.

I added a few thoughts comparing the two approaches on the original issue [here](https://github.com/woocommerce/woocommerce-admin/issues/5684#issuecomment-745641528), and I'd be curious to hear your guys' thoughts!

joelclimbsthings:
Although I like this approach, there's some justified resistance, and many of the benefits will become less prominent as the new navigation takes over. Closing this in favor of https://github.com/woocommerce/woocommerce-admin/pull/5883.

PR #5878, opened 2020-12-14
Add filter for root back URL

### LINKED ISSUE #5866: Nav: Add Filters for WordPress Dashboard link
While testing out the new nav on the eCommerce plan, I started to wonder if we should consider changing the target of the _WordPress Dashboard_ link when at the top level to point to the wordpress.com dashboard instead of `wp-admin`. I chatted briefly with @elizaan36  and @pmcpinto and they thought this might be a good idea too.

As such, I'm thinking it would be nice if we could get a couple of filters added around this, just in case it is something we would like to take advantage of via wc-calypso-bridge.

dashboard-link

Thinking we should consider adding a filter for the [`dashboardUrl`](https://github.com/woocommerce/woocommerce-admin/blob/main/client/navigation/components/container/index.js#L32) along with the [back button label too](https://github.com/woocommerce/woocommerce-admin/blob/main/client/navigation/components/container/index.js#L138)

joshuatf:
Fixes #5866 

Adds a filter to allow changing the root back button URL.

### Screenshots
Screen Shot 2020-12-14 at 4 02 39 PM

1. Filter the URL and label:
```php
add_filter( 'woocommerce_navigation_root_back_url', function() {
  return 'http://wordpress.com';
} );
add_filter( 'woocommerce_navigation_root_back_label', function() {
  return 'Retreat!';
} );
```
2. Enable the navigation.
1. Navigate to the root.
1. Make sure the back button text is updated and goes to the specified URL.

### DIFF
--- a/client/navigation/components/container/index.js
+++ b/client/navigation/components/container/index.js
@@ -10,7 +10,6 @@ import {
 	__experimentalNavigationMenu as NavigationMenu,
 	__experimentalNavigationGroup as NavigationGroup,
 } from '@wordpress/components';
-import { getAdminLink } from '@woocommerce/wc-admin-settings';
 import { NAVIGATION_STORE_NAME } from '@woocommerce/data';
 import { recordEvent } from '@woocommerce/tracks';
 import { withSelect } from '@wordpress/data';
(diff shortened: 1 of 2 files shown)


becdetat (approved):
Looks good, works well :-)

timmyc:
Thanks for the quick fix on this one @joshuatf 🙇

PR #5880, opened 2020-12-15
Allow setting Jetpack connection when JETPACK_MASTER_USER isn't available

millerf:
### A bit of background

With [gold team](https://timgoldp2.wordpress.com/), we are integrating WOO on WPCOM, which comes with its challenges.
One of them is the use of `JETPACK_MASTER_USER`, which we would like not to define on WPCOM, therefore I created this PR that should allow more flexibility without having to add another constant on the global scope.

Let me kown your thoughts or how it can be eventually improved. Also happy to add some tests if needed...

### After merging

I will need to have to wait for this to be merged and deployed to WPCOM before committing my `D54286` patch

### DIFF
--- a/src/Features/OnboardingSetUpShipping.php
+++ b/src/Features/OnboardingSetUpShipping.php
@@ -63,8 +63,13 @@ class OnboardingSetUpShipping {
 			return;
 		}
 
-		$user_token        = \Jetpack_Data::get_access_token( JETPACK_MASTER_USER );
-		$jetpack_connected = isset( $user_token->external_user_id );
+		if ( defined('JETPACK_MASTER_USER') ) {
+			$user_token = \Jetpack_Data::get_access_token( JETPACK_MASTER_USER );
+			$jetpack_connected = isset( $user_token->external_user_id );
+		} else {
+			$jetpack_connected = apply_filters( 'woocommerce_admin_is_jetpack_connected', false);
+		}
+
 		$wcs_version       = \WC_Connect_Loader::get_wcs_version();
 		$wcs_tos_accepted  = \WC_Connect_Options::get_option( 'tos_accepted' );
 
--- a/src/Features/OnboardingSetUpShipping.php
+++ b/src/Features/OnboardingSetUpShipping.php
@@ -63,7 +63,7 @@ class OnboardingSetUpShipping {
 			return;
 		}
 
-		if ( defined('JETPACK_MASTER_USER' ) ) {
+		if ( defined( 'JETPACK_MASTER_USER' ) ) {
 			$user_token = \Jetpack_Data::get_access_token( JETPACK_MASTER_USER );
 			$jetpack_connected = isset( $user_token->external_user_id );
 		} else {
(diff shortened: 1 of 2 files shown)


src/Features/OnboardingSetUpShipping.php
```diff
@@ -63,8 +63,13 @@ public static function on_onboarding_profile_completed() {
 			return;
 		}
 
-		$user_token        = \Jetpack_Data::get_access_token( JETPACK_MASTER_USER );
-		$jetpack_connected = isset( $user_token->external_user_id );
+		if ( defined( 'JETPACK_MASTER_USER' ) ) {
+			$user_token = \Jetpack_Data::get_access_token( JETPACK_MASTER_USER );
+			$jetpack_connected = isset( $user_token->external_user_id );
+		} else {
+			$jetpack_connected = apply_filters( 'woocommerce_admin_is_jetpack_connected', false );
```
becdetat, line 70: Can you add a comment block for this filter?

becdetat:
I couldn't see a way to really test this, the code is straightforward enough. I smoke tested it and everything seems to be working fine.

I just had one comment on adding a comment block to the `apply_filters` call. An example is here: https://github.com/woocommerce/woocommerce-admin/blob/main/src/FeaturePlugin.php#L61

becdetat (approved):

src/Features/OnboardingSetUpShipping.php
millerf: Done on 52a2357

millerf:
@becdetat Thanks a lot for the review!
If it gets merge, how long is your deployment cycle? ie. any idea when will it be deployed on WPCOM?

becdetat:
@millerf It should get released in WooCommerce Admin 1.9.0, which is scheduled mid-January from memory. It will then be included in the next version of WooCommerce Admin, which would be 5.0.

adrianduffell:
I noticed that `JETPACK_MASTER_USER` has been deprecated by Jetpack. Its value is now[ just hardcoded as true](https://github.com/Automattic/jetpack/blob/fff3968a4ac99ef434ed6ea4e45854bd0bea0caf/jetpack.php#L21-L27).

It seems like the connection logic should be fixed to account for this. The hardcoded value probably breaks things since it does not return a user ID anymore? I will create a follow-up issue to explore it further.

@millerf If our use of`JETPACK_MASTER_USER` is removed, would the `woocommerce_admin_is_jetpack_connected` filter still be required for your use case?

millerf:
> @millerf If our use of`JETPACK_MASTER_USER` is removed, would the `woocommerce_admin_is_jetpack_connected` filter still be required for your use case?

Hi there,

Well spotted. No, we won't have any use of this filter. The whole idea was to override the use of JETPACK_MASTER_USER. On our side the filter will only return true...
Should we go the extra step and remove this deprecated constant altogether in this PR?

adrianduffell:
> Should we go the extra step and remove this deprecated constant altogether in this PR?

This sounds good to me 👍  I'm just not sure what to replace the constant with though.

I've created an issue for it [here](https://github.com/woocommerce/woocommerce-admin/issues/5910) and aiming to investigate it further next week.

millerf:
Happy to wait after the investigation, we don't have any particular rush for this to land right now.

PR #5881, opened 2020-12-15
Add link to update WordPress when min version not met

### LINKED ISSUE #5871: Navigation: Add a link to update WP
This is a follow up to https://github.com/woocommerce/woocommerce-admin/pull/5753#issuecomment-735957142.

Navigation will only be available to users with at least WP 5.6.0 installed because of the availability of Gutenberg component `Navigation` after `wp.components` is unbundled. When trying to enable the Navigation feature on an older version of WP users should get this link:

Update WordPress

## Deliverables

* Users on WP < 5.6.0 see a link to update WP versions on the Feature setting.
* Users on WP < 5.6.0 see a disabled checkbox on the Feature setting.
* Users on WP >= 5.6.0 see no link to update and the checkbox is enabled

cc @elizaan36

joshuatf:
Fixes #5871

Adds a link to update the WP version for the new nav feature.

There does not appear to be a way to disable the input directly without JS, so this PR currently just styles the input as disabled.  Open to thoughts here if we think we should disable via JS as well.

### Screenshots
Screen Shot 2020-12-15 at 11 52 39 AM

1. Use a WP version < 5.6.
1. Visit the features settings `wp-admin/admin.php?page=wc-settings&tab=advanced&section=features`
2. Note the link if your user is able to update WP.
3. Update WP.
4. Make sure the link is no longer shown and the nav feature checkbox is full opacity (not disabled).

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -293,6 +293,7 @@ class Loader {
 					'desc'  => $description . $update_text,
 					'id'    => 'woocommerce_navigation_enabled',
 					'type'  => 'checkbox',
+					'class' => $needs_update ? 'disabled' : '',
 				),
 				array(
 					'type' => 'sectionend',
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -263,6 +263,22 @@ class Loader {
 			return $settings;
 		}
 
+		$description  = __(
+			'Adds the new WooCommerce navigation experience to the dashboard',
+			'woocommerce-admin'
+		);
+		$update_text  = '';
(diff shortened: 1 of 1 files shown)


psealock (approved):
Changes look good and this is testing well. I ran it with 5.4 and 5.6.

We should just wait to confirm https://github.com/woocommerce/woocommerce-admin/pull/5753 will definitely get merged before merging this one.

psealock:
Since https://github.com/woocommerce/woocommerce-admin/pull/5753 has been merged, this is ready to merge as well.

PR #5882, opened 2020-12-15
Use WP Card component in favor of WC Card

### LINKED ISSUE #4002: Use the <Card> WordPress component
As noted by @LevinMedia, all uses of the `<Card>` _WooCommerce_ component should be switched to the `<Card>` _WordPress_ component [[src](https://github.com/WordPress/gutenberg/tree/master/packages/components/src/card), [story](https://wordpress.github.io/gutenberg/?path=/story/components-card--default)]. (cc @jameskoster)

One possible approach to consider would be a deprecation path in which the `<Card>` _WooCommerce_ component would be reimplemented as a wrapper for the WordPress component (unless there is some incompatibility I'm unaware of). This way, all downstream usage (inside and outside of this repository) would use the "common" component sooner rather than later, and refactoring could proceed independently.

(For extra context: WooCommerce Payments is currently using `<Card>` from `@woocommerce/components` both [directly](https://github.com/Automattic/woocommerce-payments/blob/ddc2ebc1ce8305e02d93f46b9f687d57d6f6b08d/client/transaction-details/summary/index.js#L9) and [indirectly](https://github.com/Automattic/woocommerce-payments/blob/ddc2ebc1ce8305e02d93f46b9f687d57d6f6b08d/client/transactions/index.js#L10).)

joshuatf:
Fixes #4002 

This PR removes the usage of the WC `Card` component and moves towards using the GB `Card` where possible.

-   Ex: Open page `url`
-   Click XYZ…

### DIFF
--- a/client/profile-wizard/steps/benefits/index.js
+++ b/client/profile-wizard/steps/benefits/index.js
@@ -2,13 +2,13 @@
  * External dependencies
  */
 import { __, _n, sprintf } from '@wordpress/i18n';
-import { Button, __experimentalText as Text } from '@wordpress/components';
+import { Button, Card, CardBody, CardFooter, __experimentalText as Text } from '@wordpress/components';
 import { Component } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import { withDispatch, withSelect } from '@wordpress/data';
 import { filter } from 'lodash';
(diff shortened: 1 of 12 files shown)


joshuatf:
Closing this PR so we can manage this in chunks as opposed to one giant PR.

PR #5892, opened 2020-12-16
Add option to create product by template

### LINKED ISSUE #5296: Task List: add products task - introduce templates
**User story:**
As a new merchant
I want to start with a pre-filled product
So I can easily understand what are the most important fields according to the type of product that I want to sell

**Description:** product setup/management is one of the biggest areas of confusion for merchants. There’s an opportunity in the Home screen task list to make this easier by providing product templates for different product types. The goal of the product template is to help merchants understand how to add new products to make the process faster for themselves in the future. For this reason, it feels like there would be more chance of success if we test first with one industry. Otherwise, there’s the chance that starting with a template could actually cost merchants more time if they need to delete a lot of data and start over. We already have sample content for fashion stores and `Fashion, apparel, and accessories` is our biggest vertical, so in this initial iteration we're only going to target merchants that choose this industry. 

**Acceptance criteria:**
- [ ] Introduce a `Start with a template` option in the `Add Products` task. This option should only be visible for users that selected `Fashion, apparel, and accessories` as the industry.
- [ ] When the user clicks in this option display a modal with 3 different templates:
Physical product
Digital product
Variable product
- [ ] Import the following products from the [sample data](https://github.com/woocommerce/woocommerce/blob/master/sample-data/sample_products.csv) when the user clicks in one of the templates:
Physical product: id - 46
Digital product: id - 73
Variable product: id - 44
- [ ] Open the edit product page with the product that the user selected

**Event tracking:**
- Include `product_template` as value in the `method` prop of `wcadmin_tasklist_add_product`
- Add two new events:

Event name: `wcadmin_tasklist_product_template_selection`
Event prop: `product_type` (physical | digital | variable)
Event description: triggered when the user clicks go

Event name: `wcadmin_tasklist_product_template_dismiss`
Event prop: NA
Event description: triggered when the user clicks in the X to dismiss the modal

louwie17:
Fixes woocommerce/woocommerce-admin#5296 

These changes will allow users to select a product template when creating their first product. The product template is generated from a CSV file very similar to that of the sample-data.csv, and using the same underlying code for importing.
Instead of publishing the product right away it is saved as `auto-draft`, similar to how a blank product is created.

This PR includes a small underlying refactor to the `<List />` package, to allow for custom list item content.
I also was not able to use the `<RadioControl />` component in the end, as it does not support subtitles.

### Accessibility

### Screenshots

https://user-images.githubusercontent.com/2240960/104055597-69c72c80-51c5-11eb-99f9-9b06d0de55c6.mp4

- Load new store and finish the Wizard
- Go to the `Add my products` task
- Click the `Start with a template` option, and select either a physical, digital, variable product
- Once you click `Go`, it should redirect you to an edit page of the new post, with the data from the sample-data.csv (mentioned in the original ticket). Only the title is missing, as it is saved as auto-draft.
- You should be able to save the new product as draft or publish it.
- You should be able to exit without making any changes, and not having the product show up as draft in your product list. 
  - Create new product from template
  - Wait until redirected
  - Without saving go to the **Products > all products** page, the new product should not be displayed.


src/API/Reports/Products/DataStore.php
```diff
@@ -526,4 +527,90 @@ protected function initialize_queries() {
...
 	}
+
+	/**
+	 * Possible add filter to insert product template.
+	 */
+	public static function possibly_add_product_template() {
+		if ( isset( $_GET['post_type'] ) && 'product' === wc_clean( wp_unslash( $_GET['post_type'] ) ) && isset( $_GET['product_template'] ) ) { // phpcs:ignore CSRF ok, sanitation ok.
+			// default_excerpt is last filter called in get_default_post_to_edit.
+			add_filter( 'default_excerpt', array( __CLASS__, 'insert_product_template' ), 10, 2 );
+			// add_action('wp_insert_post', array(__CLASS__, 'insert_product_template_action'), 10, 2);
```
louwie17, line 538: This is not set in stone, but I currently added two options (scratch one of them), these both get triggered when the post-new page gets loaded with a `product_template` param. The third option is scratching this part, creating an REST endpoint for creating a product template, and then redirecting to the product edit page (in all cases the bulk of the ProductCSVTemplateImporter logic would stay the same).

The current add filter option:
- default_excerpt is the last filter called before `get_default_post_to_edit` finishes, used by `post-new.php`.
- Updates the post with the template
- Redirects to the posts edit page and exits the original process.

The second action option using `wp_insert_post`:
- Called when `get_default_post_edit` creates the initial post
- Updates the post
- The title, description, and excerpt are all set after this within `get_default_post_to_edit`, using filters
- So it adds filters for the description and excerpt (see line 566) to use the correct post data

client/task-list/tasks/products/product-template-modal.js
```diff
@@ -0,0 +1,95 @@
...
+	};
+
+	const templateList = templates.map( ( task ) => {
+		task.onClick = () => {
+			setSelectedTemplate( task.key );
+		};
+		task.before = (
+			<div className="woocommerce-task__icon">
+				{ task.key === selectedTemplate ? (
+					<Icon icon={ selectedRadioButtonIcon } />
```
jeffstieler, line 55: Why was this needed? It seems the [`RadioControl`](https://developer.wordpress.org/block-editor/components/radio-control/#do) has a selected indication already?

src/API/Reports/Products/DataStore.php
```diff
@@ -526,4 +527,90 @@ protected function initialize_queries() {
...
+	 * Loads the product template by the 'product_template' param, and updates the new post with the template data.
+	 *
+	 * @param string  $excerpt Default post excerpt.
+	 * @param WP_Post $post    Post object.
+	 * @return string $excerpt
+	 */
+	public static function insert_product_template_action( $excerpt, $post ) {
+		if ( 'product' === $post->post_type && 'auto-draft' === $post->post_status && ! empty( $_REQUEST['product_template'] ) ) {
+			remove_action('wp_insert_post', array(__CLASS__, 'insert_product_template_action'), 10, 2);
+			include_once WC_ADMIN_ABSPATH . 'src/ProductCSVTemplateImporter.php';
```
jeffstieler, line 552: This should get taken care of by the autoloader.

src/API/Reports/Products/DataStore.php
```diff
@@ -101,6 +101,7 @@ protected function assign_report_columns() {
 	 */
 	public static function init() {
 		add_action( 'woocommerce_analytics_delete_order_stats', array( __CLASS__, 'sync_on_order_delete' ), 10 );
+		add_action( 'load-post-new.php', array( __CLASS__, 'possibly_add_product_template' ), 10 );
```
jeffstieler, line 104: (This isn't really the line I'm commenting on)

This DataStore class powers the Products Report API endpoint, so I don't think it's an appropriate home for the "add new product from a template" code. We have an `OnboardingTasks.php` file that [already has sample product importing logic](https://github.com/woocommerce/woocommerce-admin/blob/main/src%2FAPI%2FOnboardingTasks.php#L126) that I think is a more suitable home.

src/API/Reports/Products/DataStore.php
```diff
@@ -526,4 +527,90 @@ protected function initialize_queries() {
...
+			if ( file_exists( $template_path ) && class_exists( '\Automattic\WooCommerce\Admin\ProductCSVTemplateImporter' ) ) {
+				$importer_class = apply_filters( 'woocommerce_product_csv_importer_template_class', '\Automattic\WooCommerce\Admin\ProductCSVTemplateImporter' );
+				$args           = array(
+					'parse' => true,
+				);
+
+				$importer = new $importer_class( $template_path, $args );
+				$importer->update_product( $post->ID );
+
+				add_filter('default_content', function($content, $post) {
```
jeffstieler, line 566: I'd recommend sticking to non-anonymous functions for filter callbacks wherever possible. It's probably best to have these filters ensure that they only run once by removing themselves upon invocation (like the parent function here does). While it's unlikely that `get_default_post_to_edit()` will get called again before the end of this PHP request, it's certainly not impossible.

src/API/Reports/Products/Templates/digital_product.csv
```diff
@@ -0,0 +1,2 @@
...
+"simple, downloadable, virtual",woo-album,Album,0,0,visible,"This is a simple, virtual product.","Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vestibulum sagittis orci ac odio dictum tincidunt. Donec ut metus leo. Class aptent taciti sociosqu ad litora torquent per conubia nostra, per inceptos himenaeos. Sed luctus, dui eu sagittis sodales, nulla nibh sagittis augue, vel porttitor diam enim non metus. Vestibulum aliquam augue neque. Phasellus tincidunt odio eget ullamcorper efficitur. Cras placerat ut turpis pellentesque vulputate. Nam sed consequat tortor. Curabitur finibus sapien dolor. Ut eleifend tellus nec erat pulvinar dignissim. Nam non arcu purus. Vivamus et massa massa.",,,taxable,,1,,0,0,,,,,1,,,15,Music,,,https://woocommercecore.mystagingwebsite.com/wp-content/uploads/2017/12/album-1.jpg,1,1,,,,,,,0,,,,,,,,,1,"Single 1",https://demo.woothemes.com/woocommerce/wp-content/uploads/sites/56/2017/08/single.jpg,"Single 2",https://demo.woothemes.com/woocommerce/wp-content/uploads/sites/56/2017/08/album.jpg
```
jeffstieler, line 2: Is it possible to grab this data from the core file (`/sample-data/sample_products.csv`) instead of having to duplicate these?

src/API/Reports/Products/DataStore.php
jeffstieler: Curious - would it be simpler to import the appropriate sample product (as a new product) and redirect to the edit screen instead?

client/task-list/tasks/products/product-template-modal.js
louwie17: Thanks, didn't expect the detailed look, appreciate it :) This whole map is actually old code, and not using it.

src/API/Reports/Products/DataStore.php
louwie17: Yes, if we kept the individual csv files, as (I believe, not 100% on this) the default `class-wc-product-csv-importer` imports everything within the file.
We might want the individual csv files anyway, as the sample-data has it set to auto publish.

src/API/Reports/Products/DataStore.php
louwie17: Yeah I didn't like making use of these either, and it was kind of a back up if we wanted to go this route.

src/API/Reports/Products/Templates/digital_product.csv
louwie17: Likely, I didn't as I didn't want to specifically match by `id` or `type` from the sample data.
They also have Published set to `true` which would have to be removed (not a deal breaker of-course).

packages/data/src/items/actions.js
```diff
@@ -79,3 +79,22 @@ export function* updateProductStock( product, quantity ) {
 		return false;
 	}
 }
+
+export function* createProductTemplate( itemFields, query ) {
```
jeffstieler, line 83: Are we creating a product template, or a product from a template?

packages/data/src/items/actions.js
```diff
@@ -79,3 +79,22 @@ export function* updateProductStock( product, quantity ) {
...
+		);
+		const newItem = yield apiFetch( {
+			path: url,
+			method: 'POST',
+			data: itemFields,
+		} );
+		yield setItem( 'products', newItem.id, newItem );
+		return newItem;
+	} catch ( error ) {
+		yield setError( 'createItem', error );
```
jeffstieler, line 97: Is this too generic an error name? Should this be `createProductFromTemplate` or similar?

src/API/ProductTemplates.php
```diff
@@ -0,0 +1,117 @@
...
+namespace Automattic\WooCommerce\Admin\API;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * REST API Admin Notes controller class.
+ *
+ * @extends WC_REST_CRUD_Controller
+ */
+class ProductTemplates extends \WC_REST_Products_Controller {
```
jeffstieler, line 17: Naming nitpick again - does this endpoint manage the product templates themselves, or does it facilitate the creation of products based on templates?

src/API/ProductTemplates.php
```diff
@@ -0,0 +1,117 @@
...
+		);
+	}
+
+	/**
+	 * Get a single note.
+	 *
+	 * @param WP_REST_Request $request Request data.
+	 * @return WP_REST_Response|WP_Error
+	 */
+	public function create_item( $request ) {
```
jeffstieler, line 67: Seeing that this function is largely a copy-paste of [`OnboardingTasks::import_sample_product()`](https://github.com/woocommerce/woocommerce-admin/blob/b49c96ee2201445c9cb32a5cdf324c33fd1a3a06/src%2FAPI%2FOnboardingTasks.php#L126), can we refactor that into a reusable method for both endpoints?

And while we're reusing that code.. and this feature is part of the onboarding.. why not have this endpoint code live in `OnboardingTasks.php`?

src/API/Templates/variable_product.csv
```diff
@@ -0,0 +1,2 @@
...
+variable,"V-Neck T-Shirt",0,1,visible,"This is a variable product.","Pellentesque habitant morbi tristique senectus et netus et malesuada fames ac turpis egestas. Vestibulum tortor quam, feugiat vitae, ultricies eget, tempor sit amet, ante. Donec eu libero sit amet quam egestas semper. Aenean ultricies mi vitae est. Mauris placerat eleifend leo.",,,taxable,,1,,0,0,.5,24,1,2,1,,,,"Clothing > Tshirts",,,"https://woocommercecore.mystagingwebsite.com/wp-content/uploads/2017/12/vneck-tee-2.jpg, https://woocommercecore.mystagingwebsite.com/wp-content/uploads/2017/12/vnech-tee-green-1.jpg, https://woocommercecore.mystagingwebsite.com/wp-content/uploads/2017/12/vnech-tee-blue-1.jpg",,,,,,,,,0,Color,"Blue, Green, Red",1,1,Size,"Large, Medium, Small",1,1,1,,,,
```
jeffstieler, line 2: I know this is pulled directly from the core sample data - but do we want to create a "featured" product here?

jeffstieler (commented):
I think we can reduce the amount of code required for this feature by moving it to the `OnboardingTasks.php` API file and refactoring the `import_sample_products()` method to be reusable. I left some feedback as well about naming (hopefully to avoid future confusion).

client/task-list/tasks/products.js
```diff
@@ -57,16 +75,32 @@ const subTasks = [
...
-						<List items={ subTasks } />
-					</CardBody>
-				</Card>
-			</Fragment>
-		);
-	}
+export default function Products() {
+	const [ selectTemplate, setSelectTemplate ] = useState( null );
+
+	const onTaskClick = ( task ) => {
```
louwie17, line 81: Intercept the `onClick` callback, to see if the addProductTemplate is called.

client/task-list/tasks/products/product-template-modal.js
```diff
@@ -0,0 +1,114 @@
...
+	},
+	{
+		key: 'digital',
+		title: __( 'Digital product', 'woocommerce-admin' ),
+	},
+	{
+		key: 'variable',
+		title: __( 'Variable product', 'woocommerce-admin' ),
+	},
+];
```
louwie17, line 35: Currently left as a hardcoded list here.
Depending on customizability, this could be changed to registered configs on the back-end, that we retrieve with the API. I don't think that is quite necessary atm.

packages/data/src/items/reducer.js
```diff
@@ -15,14 +15,15 @@ const reducer = (
...
 			return {
 				...state,
 				data: {
 					...state.data,
 					[ itemType ]: {
-						...state.data[ itemType ],
+						...itemData,
 						[ id ]: {
-							...state.data[ itemType ][ id ],
+							...( itemData[ id ] || {} ),
```
louwie17, line 26: Fix issue, where if we call SET_ITEM with a new item (not retrieved yet).

src/API/OnboardingTasks.php
```diff
@@ -43,7 +43,7 @@ public function register_routes() {
 				array(
 					'methods'             => \WP_REST_Server::CREATABLE,
 					'callback'            => array( $this, 'import_sample_products' ),
-					'permission_callback' => array( $this, 'import_products_permission_check' ),
+					'permission_callback' => array( $this, 'create_products_permission_check' ),
```
louwie17, line 46: Rename to reflect the permission check, and to make it less confusing when used else where.

src/API/OnboardingTasks.php
```diff
@@ -119,32 +142,78 @@ public function get_status_permission_check( $request ) {
...
 
 	/**
-	 * Import sample products from WooCommerce sample CSV.
+	 * Import sample products from given CSV path.
 	 *
+	 * @param  string $csv_file CSV file path.
 	 * @return WP_Error|WP_REST_Response
 	 */
-	public static function import_sample_products() {
+	public static function import_sample_products_from_csv( $csv_file ) {
```
louwie17, line 150: Refactored, so we can use it for `create_product_from_template` as well.

src/API/Templates/digital_product.csv
```diff
@@ -0,0 +1,2 @@
+Type,Name,Published,"Is featured?","Visibility in catalog","Short description",Description,"Date sale price starts","Date sale price ends","Tax status","Tax class","In stock?",Stock,"Backorders allowed?","Sold individually?","Weight (lbs)","Length (in)","Width (in)","Height (in)","Allow customer reviews?","Purchase note","Sale price","Regular price",Categories,Tags,"Shipping class",Images,"Download limit","Download expiry days",Parent,"Grouped products",Upsells,Cross-sells,"External URL","Button text",Position,"Attribute 1 name","Attribute 1 value(s)","Attribute 1 visible","Attribute 1 global","Attribute 2 name","Attribute 2 value(s)","Attribute 2 visible","Attribute 2 global","Meta: _wpcom_is_markdown","Download 1 name","Download 1 URL","Download 2 name","Download 2 URL"
```
louwie17, line 1: Adding these as seperate CSV files, copied from sample_data (with some minor changes) seemed the easiest, as I can just use the `WC_Product_CSV_Importer` class out of the box.

packages/components/src/list/index.js
```diff
@@ -2,119 +2,54 @@
...
-import { Component } from '@wordpress/element';
-import { ENTER } from '@wordpress/keycodes';
 import PropTypes from 'prop-types';
 import { CSSTransition, TransitionGroup } from 'react-transition-group';
 
 /**
  * Internal dependencies
  */
-import Link from '../link';
+import ListItem from './list-item';
```
louwie17, line 11: Moved most of this to the `ListItem` component, allowing us to just use the `List` component as a stand alone with custom children (see readme example).

packages/components/src/list/list-item.js
```diff
@@ -0,0 +1,113 @@
...
+		const { href, linkType } = item;
+
+		if ( linkType ) {
+			return linkType;
+		}
+
+		return href ? 'external' : null;
+	};
+
+	const { item } = props;
```
octaedro, line 34: What do you think about deconstructing the object `item`? Maybe that would make this part of the code easier to read.

octaedro (commented):
Hi @louwie17 , you did awesome work here! I left a small comment and found a couple of things in the log file:

screenshot-professional-hamerkop jurassic ninja-2021 01 18-13_25_47

It looks like something went wrong after pressing `Ok` in the modal.

packages/components/src/list/list-item.js
louwie17: Thanks, done in https://github.com/woocommerce/woocommerce-admin/pull/5892/commits/6e2d0c1310b8ebc17f48b81b22802334608290f2

louwie17:
> Hi @louwie17 , you did awesome work here! I left a small comment and found a couple of things in the log file:
> It looks like something went wrong after pressing `Ok` in the modal.

@octaedro I can't seem to reproduce these logs, I do have `WP_DEBUG` and `WP_DEBUG_LOG` set to True, is there anything else I am missing?
Is there a specific way you can reproduce this?

louwie17:
> Hi @louwie17 , you did awesome work here! I left a small comment and found a couple of things in the log file:
> 
> screenshot-professional-hamerkop jurassic ninja-2021 01 18-13_25_47
> 
> It looks like something went wrong after pressing `Ok` in the modal.

Thanks for pointing this out @octaedro turns out I was calling the `createProductFromTemplate` function in an old way, which caused the json body to just be a string, causing the `foreach` error.
The other warning was already present when creating a product manually, I just added an extra check.
See the https://github.com/woocommerce/woocommerce-admin/pull/5892/commits/30c1aee41d54678a83d4a67dff211e27f05c60aa for changes.

octaedro (approved):
Great job @louwie17! The code looks good and is testing well here! :shipit:

louwie17:
@pmcpinto this will be ready to merge, given it is a decent size feature, would you be able to give it a once over to see if things are as intended?

pmcpinto:
@louwie17 what's the easiest way to test it? Also, feel free to record a quick demo and share it here.

louwie17:
> @louwie17 what's the easiest way to test it? Also, feel free to record a quick demo and share it here.

I can record a quick demo 👍

louwie17:
@pmcpinto Here is the recording demo: https://d.pr/v/hg3p6J
Also I ended at the variation product, it doesn't show any variations at the moment, and to create them you have to set `Used for variations` in the attributes.
This is not as straight forward, should I add one of the variation options to that one, one of 76, 77, or 78 - https://github.com/woocommerce/woocommerce/blob/master/sample-data/sample_products.csv
I just clued into that now in the demo, that, that's the way it works.
It should be a simple fix, and just add one of the above (or all 3) to here: https://github.com/woocommerce/woocommerce-admin/blob/b9f0292ed68fd7ed859c998305459657e6de3956/src/API/Templates/variable_product.csv

pmcpinto:
Thanks @louwie17, looks great! cc @elizaan36 in case you want to take a look

louwie17:
Created woocommerce/woocommerce#32269 as a follow up to add a default variation.

elizaan36:
hey @louwie17 - looks great. This is my bad for not having the latest update in the design, but can we use this icon and copy:

PR #5895, opened 2020-12-16
Update task list to use wp card component

joshuatf:
Fixes (partially) #4002

Replace cards used in task list with WP `Card` component and style accordingly.

Styling here mostly stays true to the previous task list with the exception of card changes, such as border, box shadow, and default padding.

### Screenshots
Screen Shot 2020-12-16 at 6 09 03 PM
Screen Shot 2020-12-16 at 6 08 58 PM
Screen Shot 2020-12-16 at 6 08 53 PM
Screen Shot 2020-12-16 at 6 08 48 PM

1. Go through the task list.
2. Make sure that tasks are styled as expected.

### DIFF
--- a/client/dashboard/style.scss
+++ b/client/dashboard/style.scss
@@ -89,6 +89,11 @@
 .woocommerce-task-dashboard__body {
 	.woocommerce-task-dashboard__container {
 		.woocommerce-task-card {
+			max-width: 680px;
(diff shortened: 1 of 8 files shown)


becdetat (approved):
Everything works well, I smoke tested the tasks list and it was all Tip Top. Code all makes sense. 🚢

PR #5896, opened 2020-12-16
Update onboarding setup wizard to use WP Card component

joshuatf:
Fixes (partialy) #4002

Uses the WP Card instead of WC Card for the setup wizard.

Styles here diverge slightly from the original, most notably:
* The business extensions have borders on top and bottom for both the bundled and non-bundled version.
* Theme card footers place buttons side-by-side as opposed to on top of each other.

My understanding was that the styles used here would more closely follow the default used in GB, so I left these as is, but happy to change to match the previous obw styling if we see fit.

### Screenshots
Screen Shot 2020-12-16 at 6 31 57 PM
Screen Shot 2020-12-16 at 6 31 21 PM
Screen Shot 2020-12-16 at 6 24 13 PM
Screen Shot 2020-12-16 at 6 21 03 PM
Screen Shot 2020-12-16 at 6 20 52 PM
Screen Shot 2020-12-16 at 6 20 36 PM

1. Disconnect JP and/or WCS (to see benefits)
2. Go through the setup wizard.
2. Note the card styling is as expected.
3. Switch industries to show bundled/non-bundled extension styling.

### DIFF
--- a/client/profile-wizard/steps/benefits/index.js
+++ b/client/profile-wizard/steps/benefits/index.js
@@ -2,7 +2,13 @@
  * External dependencies
  */
 import { __, _n, sprintf } from '@wordpress/i18n';
-import { Button, Card, CardBody, CardFooter, __experimentalText as Text } from '@wordpress/components';
+import {
+	Button,
+	Card,
+	CardBody,
+	CardFooter,
+	__experimentalText as Text,
+} from '@wordpress/components';
 import { Component } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import { withDispatch, withSelect } from '@wordpress/data';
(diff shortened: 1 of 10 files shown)


becdetat:
The styling on the theme selection step threw me a bit:

image

I was expecting the centre aligned buttons and snackbar, rather than left and right aligned.

Is this how it's meant to be styled?

joshuatf:
Thanks for the review @becdetat!

>I was expecting the centre aligned buttons

I left the buttons left and right as this is what the GB card wants by default, but we can change this if we choose.  We're updating styles to more closely match the defaults used in the GB component, so just wondering where we draw this line.  Also worth noting, this is actually the styling used for theme cards in the original designs. /cc @LevinMedia @jameskoster 

>My understanding was that the styles used here would more closely follow the default used in GB, so I left these as is, but happy to change to match the previous obw styling if we see fit.

>and snackbar

I think this one is unrelated to these changes, but I'll double check.

psealock (approved):
Great clean up @joshuatf, this is testing well and code changes look good.

PR #5899, opened 2020-12-17
Use wp cards in store alerts

joshuatf:
Fixes (partially) #4002

Updates store alerts to use GB cards and adds tests for the store alerts.

Styles are pretty much identical to pre-existing styles with the exception of some updated placeholder styles.

### Screenshots

**Before**
Screen Shot 2020-12-16 at 7 03 47 PM

**After**
Screen Shot 2020-12-16 at 7 33 09 PM
Screen Shot 2020-12-16 at 7 07 56 PM

1. Trigger at least 2 store alerts (or just add `return true;` to the top of  `CouponsPageMoved::can_be_added()` and refresh a couple times)
2. Check that styles match.
3. Refresh and make sure that the loading (placeholder) style is correct.

### DIFF
--- a/client/layout/store-alerts/index.js
+++ b/client/layout/store-alerts/index.js
@@ -3,14 +3,22 @@
  */
 import { __ } from '@wordpress/i18n';
 import { Component, Fragment } from '@wordpress/element';
-import { Button, Dashicon, SelectControl } from '@wordpress/components';
+import {
+	Button,
+	Card,
+	CardBody,
+	CardFooter,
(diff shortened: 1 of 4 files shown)


psealock (approved):
Great cleanup Josh and thanks for the tests! This is testing well and code looks good 🚢 

>Check that styles match.

Just noting that the intention is to use the GB Card as it comes without styling it according to how we've been doing previously. I don't think that applies here, but just wanted to note that.

PR #5901, opened 2020-12-17
Add SearchListControl isCompact prop

Aljullu:
`SearchListControl` had different styles when it was rendered inside `.components-panel`. That worked well in WC Blocks to render different styles between the main post area and the sidebar:

imatge

However, in the new widgets editor, all blocks will be inside a panel, so the styles conflict.

This PR modifies the CSS of the component so it doesn't rely on being inside a `.components-panel` and instead it uses an `isCompact` prop.

This PR includes some other changes but I commented them inline.

-   Run `npm run storybook`.
-   Check the `SearchListControl` component, toggle the `Compact` checkbox and click on the `Toggle loading state` button.
-   Ensure everything looks good and there isn't anything visually off.

### DIFF
--- a/packages/components/src/search-list-control/style.scss
+++ b/packages/components/src/search-list-control/style.scss
@@ -202,11 +202,7 @@
 	}
 }
 
-.components-panel {
-	.woocommerce-search-list {
-		padding: 0;
-	}
-
+.woocommerce-search-list.is-compact {
 	.woocommerce-search-list__selected {
 		margin: 0 0 $gap;
 		padding: 0;
--- a/packages/components/src/search-list-control/style.scss
+++ b/packages/components/src/search-list-control/style.scss
@@ -103,6 +103,11 @@
 			background: $gray-100;
 		}
 
+		&:active,
+		&:focus {
+			box-shadow: none;
+		}
+
 		&:last-child {
 			border-bottom: none !important;
 		}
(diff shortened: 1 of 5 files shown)


packages/components/src/search-list-control/style.scss
```diff
@@ -104,6 +103,11 @@
 			background: $gray-100;
 		}
 
+		&:active,
+		&:focus {
+			box-shadow: none;
```
Aljullu, line 108: This hides the `:focus`/`:active` box shadow, which was looking a bit off. We already have `:hover` styles, so I think that's enough for a11y:

| Before | After |
| --- | --- |
| imatge | imatge |

packages/components/src/search-list-control/style.scss
```diff
@@ -43,12 +43,11 @@
 }
 
 .woocommerce-search-list__list {
+	border: 1px solid $gray-100;
```
Aljullu, line 46: Before, the list only had top/bottom border but I changed it so it also has left/right border. We have a CSS rule for that in WC Blocks but I think it makes sense upstreaming.

packages/components/src/search-list-control/style.scss
```diff
@@ -202,11 +206,7 @@
 	}
 }
 
-.components-panel {
-	.woocommerce-search-list {
-		padding: 0;
```
Aljullu, line 207: This will fix the visual glitch in the screenshot I posted in the PR description: there was no space between the list and the label below:

imatge

Aljullu:
@jeffstieler want to take a look at this one too? :slightly_smiling_face:

jeffstieler (approved):
LGTM! :shipit:

PR #5903, opened 2020-12-17
Update filters to use wp card component

joshuatf:
Fixes (partially) #4002

Uses the wp card in both the advanced and compare filters.  Also updates some broken styles in these components.

Styling is more or less identical to the previous design with the exception of new card styles (borders and drop shadows).

### Screenshots
Screen Shot 2020-12-17 at 1 43 15 PM
Screen Shot 2020-12-17 at 1 25 15 PM

1. Navigate to reports that contain advanced and compare filters (orders/products).
2. Add various filters.
3. Make sure styling is as expected.

### DIFF
--- a/packages/components/src/advanced-filters/index.js
+++ b/packages/components/src/advanced-filters/index.js
@@ -2,8 +2,17 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
+import {
+	Button,
+	Card,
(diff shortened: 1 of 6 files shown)


psealock (approved):
Another nice clean up, thanks @joshuatf 

The cards look good, filters behaving and looking as they should 🚢

PR #5904, opened 2020-12-17
Update table component to use wp card

joshuatf:
Fixes (partially) #4002 

Updates the table and report table to use the new `Card` component.  Also updates and fixes up some styles in the table.

Styling here is nearly identical to the previous tables with the exception of table borders and shadows.  Note that the inner table border is now darker than the table outline, so I'm thinking we should probably update that to match the outer table border. /cc @elizaan36 @LevinMedia 

### Screenshots
Screen Shot 2020-12-17 at 4 07 25 PM
Screen Shot 2020-12-17 at 4 07 11 PM
Screen Shot 2020-12-17 at 4 07 03 PM

1. Create enough products to be paginated.
1. Navigate through the various reports.
2. Check that the table styles and actions/search/menu are styled correctly.
3. Also check various screen sizes.


LevinMedia:
looking good @joshuatf 

>Note that the inner table border is now darker than the table outline, so I'm thinking we should probably update that to match the outer table border. 

I concur. I noticed the same issue is happening in the chart cards as well. 

Additionally, I'm not sure if it was introduced in this issue, or if this happened previously, but the rows in the table seem to have lost the horizontal rules that separate them from one another... see this screen shot from an older version of the experience. 

Screen Shot 2020-12-17 at 2 14 07 PM

I think we should add those back in. If it's un-related to this issue, I can open a new one. 

One other thing I noticed - Now when the table is the last element on the page, and the page is long enough to require a scroll, there's no more bottom margin, so the table sits flush with the bottom of the screen. I think it would benefit from some additional space below the table, ~40px to match the margins on the left and right.

**Examples:**

Revenue report

Screen Shot 2020-12-17 at 2 06 20 PM

Overview screen

Screen Shot 2020-12-17 at 2 06 03 PM

LevinMedia:
Looking at this in the tables used by WCpay. Generally, it's working 🎉  Both the deposit and transactions screens are picking up the correct styles. 

However, the search bar in the transactions screen is misbehaving (cc @dechov)

Screen Shot 2020-12-17 at 2 24 22 PM


Table header search bars that are in the core reports (such as customers, products, etc) are displaying fine:

Screen Shot 2020-12-17 at 2 27 08 PM

elizaan36:
> I'm thinking we should probably update that to match the outer table border

+1 @joshuatf

Good catch

> I think we should add those back in. If it's un-related to this issue, I can open a new one.

@LevinMedia Weird that the divider lines disappeared, agree that we should add those back in.

+1 on 40px padding at the bottom as well!👏

joshuatf:
Thanks for the feedback @LevinMedia and @elizaan36!  Updated border colors to match the card border, but let me know if we need to tweak.

>Additionally, I'm not sure if it was introduced in this issue, or if this happened previously, but the rows in the table seem to have lost the horizontal rules that separate them from one another... see this screen shot from an older version of the experience.

Looks like it's on `main` as well, but went ahead and fixed up here.

Looking into the WC Pay search box issue now.

joshuatf:
Refactored this a bit more as I think some of this logic can be simplified in https://github.com/woocommerce/woocommerce-admin/pull/5904/commits/69a7d94a559d2c4cb49a584e6d404231454d9f69

This seems to have resolved WC Pay issues and WCA tables still look good on my end.  Can someone give this a sanity check?

psealock (commented):
This is a hefty clean up with lots of CSS removed, thanks Josh. 

I checked the WC Payments tables and they appear to work as expected 🎉 

The only thing I noticed were:

* The "Compare" button touching the search bar, but only at large screens. Mobile sizes were fine.
* Using WP 5.6 includes a gridline. Is that new or was it always the case? Either way, the checkbox at the left is misaligned with the grid.

Screen Shot 2020-12-22 at 2 09 40 PM

joshuatf:
>The "Compare" button touching the search bar, but only at large screens. Mobile sizes were fine.

Nice catch @psealock.  Added margin between actions.

>Using WP 5.6 includes a gridline. Is that new or was it always the case? Either way, the checkbox at the left is misaligned with the grid.

The grid is new (recommended by @LevinMedia above).  Updated the checkbox table cells to respond appropriately to this change.

psealock (approved):
Gridlines look good and this is all working well :shipit:

PR #5906, opened 2020-12-17
Update WooCommerce Dropdown Button Styles

LevinMedia:
Small update to re-introduce the border around the button when it's active / open:

Before:

Screen Shot 2020-12-17 at 3 51 42 PM

After:

Screen Shot 2020-12-17 at 3 50 25 PM

I've also removed the animation that was added to the chevron, since it was pivoting the chevron around a strange point, and generally felt distracting / unnecessary. 

To test, load up an analytics screen, click the date picker or show menu. Note the border now remains around the button when it's active / open.

### DIFF
--- a/packages/components/src/dropdown-button/style.scss
+++ b/packages/components/src/dropdown-button/style.scss
@@ -17,7 +17,6 @@
 		right: 14px;
 		width: 32px;
 		height: 48px;
-		@include animate-transform;
 	}
 
 	&.is-open {
@@ -30,6 +29,7 @@
 	&:active,
 	&.is-open {
 		background-color: $gray-100;
+		border-color: $gray-700;
 	}
 
 	&.is-multi-line .woocommerce-dropdown-button__labels {


elizaan36:
Nice @LevinMedia 
I'm not able to see this change on my local (just had to restart Docker) so would be good for someone else to review, but this looks like a good update to me.

elizaan36:
@LevinMedia was just able to check this out after I fixed the Docker issue. 

I really like the outline you added to the dropdown active state, I've always thought it was weird that it just disappeared.

Removing the animation on the caret is a good call but let's work on an overall system of transitions to make behaviors feel smoother and less abrupt throughout the app. It's something I've been wanting to define as part of the design library work we're doing.

A couple things I noticed that should be future improvements (unrelated to this PR):
- After opening the dropdown, clicking the the outlined button again should close the dropdown
- Also the mobile state for this dropdown needs to fill the width of the viewport. 
image

Otherwise the changes you made look good to me!

psealock (approved):
Wow, thanks @LevinMedia, that was bothering me too. I checked keyboard interaction as well and it works as expected.

>the mobile state for this dropdown needs to fill the width of the viewport.

@elizaan36 I just noticed this too and went ahead and fixed it in https://github.com/woocommerce/woocommerce-admin/pull/5753#issuecomment-747828528

>After opening the dropdown, clicking the the outlined button again should close the dropdown

This actually works for me. Clicking anywhere on the page, including the outlined button closes the dropdown. Were you looking at mobile by chance?

PR #5911, opened 2020-12-18
Add deprecation notice to wc card component

### LINKED ISSUE #4002: Use the <Card> WordPress component

joshuatf:
Fixes #4002 

Adds a notice that the WC Card component will be deprecated.

### Screenshots
Screen Shot 2020-12-18 at 8 58 33 AM

1. Make sure a WC Card exists somewhere (still present on most analytics pages if other PRs have not yet been merged) or add one.
2. Open your console.
3. Check that the deprecation notice exists.

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -8,6 +8,7 @@
 -   Add product attributes support to `<Search />`.
 -   Allow single-selection support to `<Search />`.
 -   Improve handling of `multiple` and `inlineTags` in `<SelectControl />`.
+-   Deprecate use of `<Card>` in favor of the `<Card>` component in `@wordpress/components`.
 
 # 5.1.2
 
--- a/packages/components/src/card/index.js
+++ b/packages/components/src/card/index.js
@@ -3,6 +3,7 @@
  */
 import { Component } from '@wordpress/element';
 import classnames from 'classnames';
+import deprecated from '@wordpress/deprecated';
 import PropTypes from 'prop-types';
 
 /**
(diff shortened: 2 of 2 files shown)


octaedro (approved):
Nice job @joshuatf. This is testing well and code looks good :shipit:

psealock (approved):
Thanks Josh. Could you also add a note to components Changelog?

joshuatf:
>Could you also add a note to components Changelog?

Good idea!  Done in https://github.com/woocommerce/woocommerce-admin/pull/5911/commits/79af1d3f8e42895d6ed0d7054d8571971cec926f

PR #5913, opened 2020-12-18
Fix summary number style regression on analytics reports

LevinMedia:
This is a Draft PR pending design review. cc @elizaan36 

This PR updates the summary number styles for analytics reports. Technically this a fix for style regression, as the summary numbers were previously styled this way, but somewhere along the line the styling was lost.

These styles make it easier to identify the selected metric, and its relationship with the data displayed in the chart directly below. This is accomplished by employing a more visually defined "Tab" style. The active, selected tab is white, matching the chart below. The highlight is moved to the top of the tab. Inactive, resting tabs are a lighter shade of grey that match the grey background in the table header. 

### For comparison

**Older version of Woocommerce Admin**

Screen Shot 2020-12-18 at 8 23 55 AM

**Currently on the main branch**

Screen Shot 2020-12-18 at 11 12 12 AM

**Updates made in this PR**

https://user-images.githubusercontent.com/4500952/102650816-53770500-4120-11eb-9a65-c0fd817a91a5.mov


elizaan36:
Nice. I really like this update because as you mentioned it makes it much more clear which card is selected and tied to the graph below. 

My only concern is that we've sort of educated people by this point (at least on the wc-admin screens like Analytics and Home) that a white background is resting and to expect a grey background on hover or active states.

To align with the meaning we've created I explored the reverse:
Frame 2027

It still maintains the connection between tab and graph but the issue with this is that the card doesn't stand out from the page background.

Side note that I tried a compact version of the SummaryNumber component specifically for use as tabs in Analytics. I'm thinking along these lines for a few reasons:
- I mentioned in Slack that it would make sense to differentiate those tabs a bit more from the other SummaryNumber components that appear as links in the Stats overview (Home screen) and Analytics Overview, since they have different purposes.
- Another reason is in case we wanted to create additional graphs in the future that would require additional tabs, or give merchants more control of which graphs appear.
- Lastly, the SummaryNumber component takes away a lot of the thunder from the graph itself. If these tabs were to take up less space and visual interest, it would be easier to focus on the graph below.

Thoughts?

LevinMedia:
👋 heya @elizaan36 thanks!

>the issue with this is that the card doesn't stand out from the page background and we'd have to bump up the lighter lines/text for accessibility.

I agree, that does make the chart somewhat problematic. I'd prefer to keep the chart white so that it's higher contrast, more legible, and consistent with the other cards in the experience. Also, I think making the active tab grey is a deviation from a fairly standard tab pattern, where the active tab is lighter than the inactive tabs. For example, browser tabs in Chrome and Firefox follow this pattern: 

Screen Shot 2021-01-12 at 9 34 06 AM

Screen Shot 2021-01-12 at 9 35 52 AM

> My only concern is that we've sort of educated people by this point (at least on the wc-admin screens like Analytics and Home) that a white background is resting and to expect a grey background on hover or active states.

I believe the change in context between the home screen and reports, along with the changes in visual design provide enough differentiation to keep people from being confused about the functionality of the summary numbers in both locations. Plus, the inactive tabs on the reports still get the same grey background hover color to indicate that they’re clickable. 

I do like the idea of creating compact versions of the summary numbers though… I think that would actually be really interesting to explore in context of the entire report.

Perhaps in a separate issue we could start by exploring an analytics setting where a merchant can choose between standard or compact components? It would affect everything in the analytics experience: summary numbers, filters, charts, tables etc.  Eventually, when the whole WooCommerce admin is JS powered, we could make it a global setting that affects all the screens across the entire experience.

samueljseay:
@elizaan36 bump on this one :)

elizaan36:
> I believe the change in context between the home screen and reports, along with the changes in visual design provide enough differentiation to keep people from being confused about the functionality of the summary numbers in both locations. Plus, the inactive tabs on the reports still get the same grey background hover color to indicate that they’re clickable.

Yup, I think it works because the resting state is a lighter grey. Looks good @LevinMedia !

> I do like the idea of creating compact versions of the summary numbers though… I think that would actually be really interesting to explore in context of the entire report. 

> Perhaps in a separate issue we could start by exploring an analytics setting where a merchant can choose between standard or compact components? It would affect everything in the analytics experience: summary numbers, filters, charts, tables etc. Eventually, when the whole WooCommerce admin is JS powered, we could make it a global setting that affects all the screens across the entire experience.

Yea, that would be cool to have a global compact mode but I'm not familiar enough with customer feedback on Analytics to know if that's a problem to solve. 

The problem I'm referring to is mainly this one that's slightly awkward, when the highlighted tab is in the top row.
image

Was thinking it could be a good exploration to keep the SummaryNumber tabs along the top of the graph, yeah in some kind of compact version.

LevinMedia:
@samueljseay - @elizaan36 and I followed up on slack, we're good to go with the updates in this PR.  

We discussed the concept of "tab" style breaking down as soon as it goes to two rows, and that is not ideal. However, it's the fix in this branch still makes more obvious which summary number is selected / active vs what's in main at the moment. There's also only one report that breaks to two rows of summary numbers, the revenue report.

adrianduffell:
Hi @LevinMedia, could you please merge this PR with the latest main? It will pick up some recent CI changes that allow the status checks to work properly.

adrianduffell (approved):
This tests well and code looks good. Appreciate this contribution @LevinMedia!

Let's add the changelog entry and testing instructions we discussed on slack, then this is good to merge 🚀

adrianduffell:
@LevinMedia I saw that I could push the changelog/testing instructions, so I have done just that 🙂 

Merging this now, thank you!

Issue #5925, opened 2020-12-21
SummaryNumber's delta does not render floats.


tomalec:
### **Describe the bug**
`<SummaryNumber>` renders only the decimal part of the given delta.

## **To Reproduce**

1. Create an instance of `<SummaryNumber>` with `delta=5.35`, 
    ```js
    <SummaryNumber
		label={ "Net Sales" }
		value={ "$10,802.93" }
		delta={ 5.35 }
	/>
    ```
2. Check how it's rendered.

### **Expected behavior**
I'd like to see the delta with the precision I gave.

### **Screenshots**
![Screenshoot where SummaryNumber renders "5%", even though `delta: 5.35` is provided to the react component.](https://user-images.githubusercontent.com/17435/102824380-02e00000-43dd-11eb-965e-613fd24940a6.png)

### **Desktop (please complete the following information):**

* OS: Win10
* Browser Chrome
* Version 87.0.4280.88 

### **Additional context**
This precision is required for 96-gh-woocommerce/google-listings-and-ads

PR #5926, opened 2020-12-21
Render SummaryNumber delta precisely.

tomalec:
Previously only a given delta was rendered as an integer. Now it renders the number with a given precision.
Fixes https://github.com/woocommerce/woocommerce-admin/issues/5925.

### Accessibility

I didn't add any new interactions

1. Create an instance of `<SummaryNumber>` with `delta=10.12`, 
    ```js
    <SummaryNumber
		label={ "Net Sales" }
		value={ "$24.00" }
		delta={ 10.12 }
	/>
    ```
2. Check how it's rendered.

Or 
1. Build storybook `npm run storybook`
2. Open [`SummaryList` demo](http://localhost:6007/?path=/story/woocommerce-admin-components-summarylist--basic)
3. Check how the number is rendered.

image

## Risks:
There may be places that rely on the previous behavior to format the number. Expecting `<SummaryNumber delta={ 10.0000000001 }/>` to render only `10%`.
After this change may get unexpected results. However, this behavior was not documented before, so I'd consider it was risky in the first place to rely on undocumented side-effects.

## Addtitional comments:
Previously it was truncating the number, not rounding it, which I find misleading from the user perspective.


becdetat (dismissed):
This works well in storybook and the code looks good. I'm a bit concerned about the line endings changes in readme.txt, we've had a few problems with normalising them. Can you unstage the extra lines?

tomalec:
Ah, sorry, that was probably a result of running Github's merge on a Windows machine. I'll fix that.

tomalec:
Ah, sorry. Fixed.

becdetat (approved):
Looks great 👍

tomalec:
Failing Travis seems to be an issue with `main` branch, as it was passing 13hrs ago before 9b28b479a4a832ba9f2b87178540d4abbea9f539.

tomalec:
I can revert the recent `main -> fix/5925-summarynumber-float` merge, and change the `readme.txt` entry in a way it will not be conflicting, to make the checks green.

becdetat:
I can't reproduce the failure locally. It _looks_ like some version dependency didn't install properly.

I'll try pushing an empty commit just to see if it was a transient error.

becdetat:
Not a transient error. I've asked for a second opinion here: p1611109900091100-slack-C9K5T0XSP

becdetat:
@tomalec it looks like this is indeed a problem with the latest `main` - everything has started failing, probably related to something in WooCommerce 4.9.1.

> I can revert the recent main -> fix/5925-summarynumber-float merge, and change the readme.txt entry in a way it will not be conflicting, to make the checks green.

If you can give that a try it would be great :-)

tomalec:
> I'll try pushing an empty commit just to see if it was a transient error.

I think there is no need for that. You can click "Restart build" on the right side at https://travis-ci.com/github/woocommerce/woocommerce-admin/builds/213731255 (once you're logged in with GitHub)

tomalec:
Now, it's not possible to merge `main` to this branch due to https://github.com/woocommerce/woocommerce-admin/issues/6133

tomalec:
I'm merging it with `--no-verify` as the errors have nothing to do with the changes from this PR.

becdetat:
Nice, thanks for this @tomalec

PR #5940, opened 2020-12-22
Fix 1.8.x Navigation requirements

psealock:
Fixes issues relating to 1.8.1 release that enabled Navigation when it should not have. More info in pbIJXs-vY-p2

* Turn off feature flags for Plugin and Core
* Leave feature flag for Development
* Remove Advanced Feature setting section that contained toggle to turn Navigation on/off. Note that this code will be moved to wc-calypso-bridge.

1. In Dev environments, make sure the feature toggle section is no longer visible.
2. Double check that the feature flags were correctly turned off for Plugin and Core.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -241,52 +241,6 @@ class Loader {
 	}
 
 	/**
-	 * Adds the Features section to the advanced tab of WooCommerce Settings
-	 *
-	 * @param array $sections Sections.
-	 * @return array
-	 */
-	public static function add_features_section( $sections ) {
-		$sections['features'] = __( 'Features', 'woocommerce-admin' );
-		return $sections;
-	}
-
-	/**
-	 * Adds the Features settings.
-	 *
-	 * @param array  $settings Settings.
-	 * @param string $current_section Current section slug.
-	 * @return array
-	 */
-	public static function add_features_settings( $settings, $current_section ) {
-		if ( 'features' !== $current_section ) {
-			return $settings;
-		}
-
-		return apply_filters(
-			'woocommerce_settings_features',
-			array(
-				array(
-					'title' => __( 'Features', 'woocommerce-admin' ),
-					'type'  => 'title',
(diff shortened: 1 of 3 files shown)


src/Loader.php
```diff
@@ -240,52 +238,6 @@ public static function load_features() {
...
-	 * @param string $current_section Current section slug.
-	 * @return array
-	 */
-	public static function add_features_settings( $settings, $current_section ) {
-		if ( 'features' !== $current_section ) {
-			return $settings;
-		}
-
-		return apply_filters(
-			'woocommerce_settings_features',
```
joshuatf, line 267: Instead of removing this entirely, could we use a filter to add the feature setting in from Navigation?  That way when we turn it on, we'll have this option available again.

joshuatf (commented):
Looking good!  Left one comment about possibly leaving the feature option in.

I also think we'll need to add a migration change to disable the feature on update so we don't force the feature on without a way to disable it for any sites that already opted in.

https://github.com/woocommerce/woocommerce-admin/blob/c21cf170e013d89e1e7e6f8fa2904cd60ceb963e/src/Install.php#L27

src/Loader.php
psealock: This code is still on `main` branch. I will create an issue to handle 1.9 Nav requirements and think about how to handle this code and its migration to wc-calypso-bridge

psealock:
>I also think we'll need to add a migration change to disable the feature on update so we don't force the feature on without a way to disable it for any sites that already opted in.

Good thinking. If the option is set to "yes" it still won't turn on the Nav because the feature flags are all off. The only thing I can think of that may be problematic is that the Nav will come on automatically when the flags are set to true for Plugin in WC 5.0. What do you think? I feel thats not necessarily a bad thing.

joshuatf (approved):
>If the option is set to "yes" it still won't turn on the Nav because the feature flags are all off.

Ah yes, that makes sense.  Things look good in that case!  LGTM :shipit:

PR #5951, opened 2020-12-23
Fix Failure to Set Up Free Local Shipping

adrianduffell:
This PR fixes a bug preventing free local shipping being set up when onboarding is completed.

At the end of the onboarding process, free local shipping should be set up when certain conditions are met, including that the Jetpack plugin is installed and connected. However the logic determining whether a site is connected to Jetpack was failing due to deprecated usage of Jetpack code. As such, the conditions to create free local shipping are never met and it fails to be set up in onboarding.

The following deprecate usage has been updated:

- The [`JETPACK_MASTER_USER`](https://github.com/Automattic/jetpack/blob/2b2103a4bf3f6767703f92f64e293751729b57bc/jetpack.php#L21-L27) constant.
- The [`Jetpack_Data`](https://github.com/Automattic/jetpack/blob/2b2103a4bf3f6767703f92f64e293751729b57bc/class.jetpack-data.php#L5-L8) class.

This has been replaced with [`Manager::is_active`,](https://github.com/Automattic/jetpack/blob/2b2103a4bf3f6767703f92f64e293751729b57bc/packages/connection/src/class-manager.php#L524) which was recommended by Jetpack developers. This provides[ compatibility with Jetpack back to version `7.5`](https://plugins.svn.wordpress.org/jetpack/tags/7.5/vendor/automattic/jetpack-connection/src/). 

Additionally, the `woocommerce_admin_is_jetpack_connected` filter has been removed because its use-case is no longer needed with usage of `JETPACK_MASTER_USER` removed, per https://github.com/woocommerce/woocommerce-admin/pull/5880#issuecomment-747925876. This filter has not made it into a release yet so I don't think there are backwards-compatibility concerns.

With a clean installation of WC Admin:

-   Complete the onboarding wizard, using the following configuration:
    - US location
    - Store contains physical products
    - Install and connect Jetpack when prompted
    - Install WooCommerce Shipping & Tax when prompted
-  Confirm that a US shipping zone is automatically added in the WooCommerce Shipping settings.

Screen Shot 2020-12-23 at 3 17 15 pm

### DIFF
--- a/src/Features/OnboardingSetUpShipping.php
+++ b/src/Features/OnboardingSetUpShipping.php
@@ -64,6 +64,7 @@ class OnboardingSetUpShipping {
 			return;
 		}
 
+		// Check Jetpack connection.
 		if ( class_exists( Jetpack_Connection_Manager::class ) ) {
  			$jetpack_connected = ( new Jetpack_Connection_Manager() )->is_active();
  		} else {
--- a/src/Features/OnboardingSetUpShipping.php
+++ b/src/Features/OnboardingSetUpShipping.php
@@ -58,8 +58,8 @@ class OnboardingSetUpShipping {
 
 		// Check Jetpack connection.
 		if ( class_exists( Jetpack_Connection_Manager::class ) ) {
- 			$jetpack_connected = ( new Jetpack_Connection_Manager() )->is_active();
- 		} else {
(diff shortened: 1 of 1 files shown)


adrianduffell:
Closing this, since this feature was removed in PR https://github.com/woocommerce/woocommerce-admin/pull/5971.

PR #5961, opened 2020-12-23
Future proof experimental components

### LINKED ISSUE #5873: Make usage of `__experimantal` prefix safe
Use of the `__experimental` prefix will eventually disappear from two components we use from `wp.components`, `Text` and `Navigation`. IE, `__experimentalText`  will just be `Text` in a future release. This something we should future-proof now instead of waiting for problems. I'll note that as of WP 5.6 and Gutenberg 9.4.0, these are both still using the prefix. Doing this will prevent issues in the future:

```js
import { __experimentalText, Text as MaybeText } from '@wordpress/components';
const Text = MaybeText || __experimentalText;
```
Possibly there is a webpack solution to handle this and keep the code clean.

joshuatf:
Fixes #5873

Use non experimental versions of components for future-proofing but fallback to experimental if they don't exist yet.

1. Check out this branch.
2. Make sure things still run as expected, especially around `Text` and navigation components.
3. If you really want to check that things continue to work, optionally spin up a dev version of GB and update named exports in  `@wordpress/components`.

### DIFF
--- a/client/analytics/components/leaderboard/index.js
+++ b/client/analytics/components/leaderboard/index.js
@@ -6,7 +6,8 @@ import {
 	Card,
 	CardBody,
 	CardHeader,
-	__experimentalText as Text,
+	__experimentalText,
+	Text as TextComponent,
 } from '@wordpress/components';
 import { Component } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
@@ -26,6 +27,8 @@ import {
 import ReportError from '../report-error';
 import sanitizeHTML from '../../../lib/sanitize-html';
 import './style.scss';
+
+const Text = TextComponent || __experimentalText;
 
 export class Leaderboard extends Component {
 	getFormattedHeaders() {
--- a/client/analytics/components/leaderboard/index.js
+++ b/client/analytics/components/leaderboard/index.js
@@ -2,13 +2,7 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import {
-	Card,
-	CardBody,
-	CardHeader,
-	__experimentalText,
-	Text as TextComponent,
-} from '@wordpress/components';
+import { Card, CardBody, CardHeader } from '@wordpress/components';
 import { Component } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import { EmptyTable, TableCard } from '@woocommerce/components';
@@ -20,6 +14,7 @@ import {
 	getLeaderboard,
 	SETTINGS_STORE_NAME,
 } from '@woocommerce/data';
+import { Text } from '@woocommerce/experimental';
 
 /**
  * Internal dependencies
@@ -27,8 +22,6 @@ import {
 import ReportError from '../report-error';
 import sanitizeHTML from '../../../lib/sanitize-html';
 import './style.scss';
-
-const Text = TextComponent || __experimentalText;
 
 export class Leaderboard extends Component {
 	getFormattedHeaders() {
--- a/packages/experimental/CHANGELOG.md
+++ b/packages/experimental/CHANGELOG.md
@@ -1,0 +1,3 @@
+# 1.0.0
+
+-   Released package
--- a/packages/experimental/CHANGELOG.md
+++ b/packages/experimental/CHANGELOG.md
@@ -1,3 +1,3 @@
 # 1.0.0
 
--   Released package
+-   Initial package
--- a/packages/experimental/README.md
+++ b/packages/experimental/README.md
@@ -1,0 +1,27 @@
+# Navigation
+
+A collection of component imports and exports that are aliases for components transitioning from experimental to non-experimental.  This package prevents the component from being undefined when the `@wordpress/components` library version is unclear.
+
+## Installation
+
+Install the module
+
+```bash
+npm install @woocommerce/experimental --save
+```
+
+## Usage
+
+Simply import the component name with the `__experimental` prefix.  If found, the non-experimental version will be imported and if not, this will fallback to the experimental version.
+
+```jsx
+import { Text } from '@woocommerce/experimental';
+
+render() {
+	return (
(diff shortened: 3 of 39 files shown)


client/analytics/components/leaderboard/index.js
```diff
@@ -27,6 +28,8 @@ import ReportError from '../report-error';
 import sanitizeHTML from '../../../lib/sanitize-html';
 import './style.scss';
 
+const Text = TextComponent || __experimentalText;
```
jeffstieler, line 31: If it's not possible to do something like this with Webpack, what do you think about creating a new file to house these fallbacks so they don't have to be implemented everywhere?

Something like.. `import { Text } from 'client/experimental-components';`?

joshuatf: That's a much better idea 🤦 

I'll check into Webpack first to see if there's a workaround for something like this.

joshuatf:
@jeffstieler I wasn't able to find a way to alias a specific export from a module.  This PR now adds an `experimental` package that aliases the experimental components.  

I opted to put this into `packages/` as opposed to `client/` since our current `packages/` also make use of some experimental components.

Open to feedback on this route as well.

jeffstieler (commented):
> I opted to put this into `packages/` as opposed to `client/` since our current `packages/` also make use of some experimental components.

That makes sense to me. I was surprised to see that the experimental package looks to be published.. are we sure we want to do that? Is it possible to have this work without publishing the package?

joshuatf:
@jeffstieler Apparently you can just set the package to `"private": true` and lerna will ignore it.  TIL.

Updated in https://github.com/woocommerce/woocommerce-admin/pull/5961/commits/06a41d4746b0bbe7d5a61b737fbdcd7e1e2b9d1b

jeffstieler:
There's a JS lint error: https://travis-ci.com/github/woocommerce/woocommerce-admin/jobs/466364739#L896

Taking another look at this now..

jeffstieler:
@joshuatf I think there's an opportunity to remove the external JS file for this: https://gist.github.com/jeffstieler/4617c96ba76cf81b856d517cf5180af8

joshuatf:
Thanks, @jeffstieler.  That did work.  Still can't quite figure out why tests aren't able to find the `@woocommerce/experimental` package though.  Anyone else on @woocommerce/isotope have experience with this issue?

psealock:
Interesting abstraction to use a package to house the logic. I like that it makes things deliberate. 

If we aren't going to publish the package, we'll be treating it as if it were in-house code that gets bundled in the main js app as well as packages like Components.

I'm bringing up the point because https://github.com/woocommerce/woocommerce-admin/pull/5762 adds Dependency Webpack Extraction Plugin, so it will assume `window.wc.experimental` exists on the window and won't bundle the code.

Since thats not the intention with this package, can you add an exception in this `wooRequestToExternal` function so that `experiments` doesn't get included in the array of `externals`? As a follow up, we'll need to be sure to release a new version of the DEWP package.

https://github.com/woocommerce/woocommerce-admin/blob/21d12e711496aaaef3a85236ee92e2ae344f2891/packages/dependency-extraction-webpack-plugin/src/index.js#L20-L33

I checked that an app like WC Payments that uses a package like `@woocommerce/components` will have the proper variables defined by looking at the compiled code for components:

Screen Shot 2021-01-05 at 11 59 41 AM

Lastly, a note in `packages/experimental/README.md` that explains this is a private package not to be consumed directly would be beneficial.

And of course the alternative would be to publish the package and treat it like other packages that make it available on the window. That would make things consistent, but could introduce some odd version issues of its own. I don't think the benefits outnumber the potential pitfalls here.

packages/experimental/README.md
```diff
@@ -0,0 +1,27 @@
+# Navigation
```
psealock, line 1:
```suggestion
# Experimental
```

packages/experimental/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+# 1.0.0
+
+-   Released package
```
psealock, line 3:
```suggestion
-   Initial package
```

packages/experimental/package.json
```diff
@@ -0,0 +1,31 @@
...
+	},
+	"bugs": {
+		"url": "https://github.com/woocommerce/woocommerce-admin/issues"
+	},
+	"main": "build/index.js",
+	"module": "build-module/index.js",
+	"react-native": "src/index",
+	"dependencies": {
+		"@babel/runtime": "^7.11.2",
+		"@wordpress/components": "10.0.0"
```
psealock, line 25: I thought this might be your tests failing issue, but no luck
```suggestion
		"@wordpress/components": "11.0.0"
```

joshuatf: Thanks for giving it a shot.  Think I've got it.  It was mocking the index file in tests that caused the issue.  Fixed up in https://github.com/woocommerce/woocommerce-admin/pull/5961/commits/7ca27ec689fa313f087d5f54596f38a55386cdcc

joshuatf:
Thanks for the feedback, @psealock.  Committed your suggestions and think I've got a fix for the tests.  Just waiting for the 🟢 🤞

psealock (approved):
Changes look good, just missing the exception in `wooRequestToExternal` so `experimental` package gets bundled.

Since this is ready to go, lets merge and I can make the relevant change directly in https://github.com/woocommerce/woocommerce-admin/pull/5762

joshuatf:
@psealock Handled your feedback in https://github.com/woocommerce/woocommerce-admin/pull/5961/commits/110edc979e21534e0b6d704b21ad4aacd8ebd872 - can I get a quick sanity check to make sure that works to ignore the dependency on the global?

psealock (approved):
Thanks for the changes Josh! Looks great 🚢

PR #5966, opened 2020-12-23
Allow default link behavior when cmd or ctrl are used

### LINKED ISSUE #5916: Link Component: cmd + click doesn't open in a new window
`<Link />` component doesn't respect OSX cmd + click to open links in a new tab. Right clicking and choosing Open Link in a New Tab does work however. I'm not sure what the windows equivalent is.

This is the case with all instances of Link, but its particularly useful on link in Navigation.

### Deliverables

* Enable the shortcut on OSX
*  Make sure the windows equivalent also works
* Determine if there are other often used link shortcuts that we should also address.

joshuatf:
Fixes #5916

Allows opening new tabs via default browser behavior when ctrl or cmd are used.

### Screenshots
Screen Shot 2020-12-23 at 4 58 52 PM

1. Try `cmd` + clicking (in OSX) or `ctrl` + clicking in Windows some `Link` components.
2. Make sure a new tab can be opened if a `href` exists.
3. Try clicking without holding keys.
4. Make sure the link still navigates normally via history update.

### DIFF
--- a/packages/components/src/link/index.js
+++ b/packages/components/src/link/index.js
@@ -15,6 +15,11 @@ class Link extends Component {
 	// With React Router 5+, <RouterLink /> cannot be used outside of the main <Router /> elements,
 	// which seems to include components imported from @woocommerce/components. For now, we can use the history object directly.
 	wcAdminLinkHandler( onClick, event ) {
+		// If cmd or ctrl are used, use default behavior to allow opening in a new tab.
+		if ( event.ctrlKey || event.metaKey ) {
+			return;
+		}
+
 		event.preventDefault();
 
 		// If there is an onclick event, execute it.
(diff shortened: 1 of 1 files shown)


ilyasfoo (approved):
Changes LGTM, works well in my OSX safari & chrome. +1 for converting to functional component 🎉 .

Just throwing ideas here, would it make sense to also add support for `event.altKey` and `event.shiftKey`? Or conversely, are there any reasons to have Link behave differently for both of these keys?

joshuatf:
>would it make sense to also add support for event.altKey and event.shiftKey

Good idea @ilyasfoo!  Added in https://github.com/woocommerce/woocommerce-admin/pull/5966/commits/29aab738e688a9a13323aa7cba95dc790a01e71a

ilyasfoo:
Thanks! Tested again and LGTM 🚢

packages/components/src/link/index.js
```diff
@@ -10,11 +9,22 @@ import { getHistory } from '@woocommerce/navigation';
...
+	const wcAdminLinkHandler = ( onClick, event ) => {
+		// If cmd, ctrl, alt, or shift are used, use default behavior to allow opening in a new tab.
+		if (
+			event.ctrlKey ||
+			event.metaKey ||
+			event.altKey ||
+			event.shiftKey
+		) {
+			return;
+		}
```
psealock, line 26: Nice solution Josh!

timmyc:
Thanks for the fix here Josh - this issue has been open for quite some time 🙇

PR #5969, opened 2020-12-24
Localize regional currency information for use during onboarding setup

### LINKED ISSUE #5760: i18n locale information is not correctly set after Setup Wizard
**Prerequisites (mark completed items with an [x]):**
- [x] I have have carried out troubleshooting steps and I believe I have found a bug.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

**Describe the bug**
I installed woocommerce on a fresh wordpress installation.
I did the setup over the new wizard and choose country liechtenstein and also on a separate installation with country switzerland.

My old PR where i submitted the correct i18n informations: https://github.com/woocommerce/woocommerce/pull/27193

**Expected behavior**
Currency, position of currency, ... should be set based on the i18n predefined setup file.

**Actual behavior**
Basic US informations get set.

**Steps to reproduce the bug (We need to be able to reproduce the bug in order to fix it.)**
Steps to reproduce the bug:
1. Install woocommerce on a fresh wordpress site
2. work with the setup wizard and choose country liechtenstein
3. finish the wizard and change to settings
4. check if the currency is $ (then it doesnt work) or CHF (then it would work)

Thanks
Brian

- [x] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [x] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/storefront/).
- [x] I can reproduce this bug consistently using the steps above.

joshuatf:
Fixes #5760

Uses the WC region info to derive currency information during the obw.

Worth discussing: Is this too much data to preload and localize?  Should we be putting both updates and requests for info into a REST endpoint?

### Screenshots
Screen Shot 2020-12-24 at 1 53 26 PM

1. Go through the setup wizard.
2. Select a different currency than your store's existing currency.
3. Make sure that currency has been updated accordingly on your site. `wp-admin/admin.php?page=wc-settings&tab=general`

### DIFF
--- a/client/profile-wizard/steps/store-details/index.js
+++ b/client/profile-wizard/steps/store-details/index.js
@@ -81,16 +81,16 @@ class StoreDetails extends Component {
 		}
 
 		const Currency = this.context;
-		const region = getCurrencyRegion( countryState );
+		const country = getCountryCode( countryState );
 		const { currencySymbols = {}, localeInfo = {} } = getSetting(
 			'onboarding',
 			{}
 		);
-		const regionInfo = localeInfo[ region ] || localeInfo.US;
+		const countryInfo = localeInfo[ country ] || localeInfo.US;
 		const symbol =
-			currencySymbols[ regionInfo.currency_code ] || currencySymbols.USD;
+			currencySymbols[ countryInfo.currency_code ] || currencySymbols.USD;
 
-		return Currency.formatPhpToJs( { ...regionInfo, symbol } );
+		return Currency.formatPhpToJs( { ...countryInfo, symbol } );
 	}
 
 	onSubmit() {
--- a/client/profile-wizard/steps/store-details/index.js
+++ b/client/profile-wizard/steps/store-details/index.js
@@ -16,7 +16,7 @@ import { Component } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
 import { withDispatch, withSelect } from '@wordpress/data';
 import { Form } from '@woocommerce/components';
-import { getCurrencyData } from '@woocommerce/currency';
+import { getSetting } from '@woocommerce/wc-admin-settings';
 import { ONBOARDING_STORE_NAME, SETTINGS_STORE_NAME } from '@woocommerce/data';
 import { recordEvent } from '@woocommerce/tracks';
 
@@ -81,8 +81,8 @@ class StoreDetails extends Component {
 		}
 
 		const region = getCurrencyRegion( countryState );
-		const currencyData = getCurrencyData();
-		return currencyData[ region ] || currencyData.US;
+		const localeInfo = getSetting( 'localeInfo', {} );
+		return localeInfo[ region ] || localeInfo.US;
 	}
 
 	onSubmit() {
@@ -112,7 +112,7 @@ class StoreDetails extends Component {
 
 		recordEvent( 'storeprofiler_store_details_continue', {
 			store_country: getCountryCode( values.countryState ),
-			derived_currency: currencySettings.code,
+			derived_currency: currencySettings.currency_code,
 			setup_client: values.isClient,
 		} );
 
@@ -124,13 +124,11 @@ class StoreDetails extends Component {
 				woocommerce_default_country: values.countryState,
 				woocommerce_store_city: values.city,
 				woocommerce_store_postcode: values.postCode,
-				woocommerce_currency: currencySettings.code,
-				woocommerce_currency_pos: currencySettings.symbolPosition,
-				woocommerce_price_thousand_sep:
-					currencySettings.thousandSeparator,
-				woocommerce_price_decimal_sep:
-					currencySettings.decimalSeparator,
-				woocommerce_price_num_decimals: currencySettings.precision,
+				woocommerce_currency: currencySettings.currency_code,
+				woocommerce_currency_pos: currencySettings.currency_pos,
+				woocommerce_price_thousand_sep: currencySettings.thousand_sep,
+				woocommerce_price_decimal_sep: currencySettings.decimal_sep,
+				woocommerce_price_num_decimals: currencySettings.num_decimals,
 			},
 		} );
 
--- a/client/profile-wizard/steps/store-details/index.js
+++ b/client/profile-wizard/steps/store-details/index.js
@@ -80,9 +80,17 @@ class StoreDetails extends Component {
 			return null;
 		}
 
+		const Currency = this.context;
 		const region = getCurrencyRegion( countryState );
-		const localeInfo = getSetting( 'localeInfo', {} );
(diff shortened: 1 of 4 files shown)


joshuatf:
@masteradhoc Would you mind testing this PR to see if this resolves the currency update issue for you?

masteradhoc:
@joshuatf will do and update you. Thanks!

masteradhoc:
@joshuatf can u message me on slack? got some questions on this.

joshuatf:
@masteradhoc Which workspace are you in and what's your slack username?

It may also just be easier to post your questions here and I can do my best to answer.

masteradhoc:
@joshuatf woocommercecommunity and masteradhoc 👍

joshuatf:
@masteradhoc I looked for a `masteradhoc` and can't seem to find a Slack account on `woocommercecommunity` with that username.  Would it be under another name?

masteradhoc:
@joshuatf Brian Haas

masteradhoc:
Thanks for the shared work on this and your fast move!
Works as expected. Tested for Liechtenstein, Switzerland and Austria.

closes #5760

joshuatf:
Fixed up a regional issue and decoding the server-side entities from currencies.

Huge shoutout to @masteradhoc for taking the time to help test this 💯

packages/currency/src/index.js
```diff
@@ -159,124 +167,24 @@ const CurrencyFactory = ( currencySetting ) => {
...
 export default CurrencyFactory;
-
-/**
- * Returns currency data by country/region. Contains code, symbol, position, thousands separator, decimal separator, and precision.
- *
- * Dev Note: When adding new currencies below, the exchange rate array should also be updated in WooCommerce Admin's `business-details.js`.
- *
- * @return {Object} Curreny data.
- */
-export function getCurrencyData() {
```
jeffstieler, line 174: Are we not concerned with backward compatibility in our JS packages?

packages/currency/src/index.js
```diff
@@ -159,124 +167,24 @@ const CurrencyFactory = ( currencySetting ) => {
...
 			return formatAmount( number );
 		},
+
+		/**
+		 * Format currency from WooCommerce PHP version to WCA JS version.
+		 *
+		 * @param {Object} phpCurrency Currency to format
+		 * @return {Object} Formatted JS currency.
+		 */
+		formatPhpToJs( phpCurrency ) {
```
jeffstieler, line 177: This feels like an odd thing to have in the public API of the currency package. Could this perhaps be a method to return the same formatted data for a given country code?

E.g.: `Currency.getDataForCountry( countryCode )` ?

client/profile-wizard/steps/store-details/index.js
```diff
@@ -80,9 +80,17 @@ class StoreDetails extends Component {
...
+		const country = getCountryCode( countryState );
+		const { currencySymbols = {}, localeInfo = {} } = getSetting(
+			'onboarding',
+			{}
+		);
+		const countryInfo = localeInfo[ country ] || localeInfo.US;
+		const symbol =
+			currencySymbols[ countryInfo.currency_code ] || currencySymbols.USD;
+
+		return Currency.formatPhpToJs( { ...countryInfo, symbol } );
```
jeffstieler, line 93: Question about this method below..

jeffstieler (changes requested):
This tests well, but I have some questions about the code changes from a public API and back compat standpoint.

packages/currency/src/index.js
joshuatf: I agree it's a bit odd.  Added `Currency.getDataForCountry` but this will require passing in the locale info and symbol info since we can't guarantee access to this localized data.

Not sure if there's a better way to go about this without coupling this data to a package.

packages/currency/src/index.js
joshuatf: Since we just had a couple breaking change versions, I went ahead and added this back in with a deprecation notice.  My guess is that we're the only ones using it since it's such a limited scope of countries, but always up for erring on the side of safety.

joshuatf:
Thanks for the review @jeffstieler.  Agreed on most of your input.  One question on whether or not you're aware of a better way to handle localized setting issues.

jeffstieler (approved):
This is looking much better, and still testing well. One note about the fallback values for `symbol` that I think needs correcting, but then let's :shipit:.

It would be awesome to not need to pass so many complex objects to the `getDataForCountry()` function, but I'm not sure how to accomplish that without either coupling the `wc-admin-settings` or to create a new endpoint and access through the `data` package.

client/profile-wizard/steps/store-details/index.js
```diff
@@ -80,9 +80,17 @@ class StoreDetails extends Component {
...
 
-		const region = getCurrencyRegion( countryState );
-		const currencyData = getCurrencyData();
-		return currencyData[ region ] || currencyData.US;
+		const Currency = this.context;
+		const country = getCountryCode( countryState );
+		const { currencySymbols = {}, localeInfo = {} } = getSetting(
+			'onboarding',
+			{}
+		);
```
psealock, line 88: Sorry for the fly-by comment, but I'm wondering if you should gather this data from the onboarding store rather than `getSetting` which reads data initialised on the window. Can there be a situation where this data gets updated after initial page load? For instance, if a user edits and goes back?

joshuatf: @psealock This data will not get updated after initial load.  It is all the locale info and possible symbols from WC core so this data should not be updated after initial page load.

packages/currency/src/index.js
```diff
@@ -100,10 +100,44 @@ const CurrencyFactory = ( currencySetting ) => {
...
+	 * @param {Object} currencySymbols Currency symbols by symbol code.
+	 * @return {Object} Formatted currency data for country.
+	 */
+	function getDataForCountry(
+		countryCode,
+		localeInfo = {},
+		currencySymbols = {}
+	) {
+		const countryInfo = localeInfo[ countryCode ] || {};
+		const symbol = currencySymbols[ countryInfo.currency_code ] || {};
```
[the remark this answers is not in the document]
joshuatf: Good call.  `countryInfo` can also be simplified here and we can simply check if these are defined or not.

packages/currency/src/index.js
```diff
@@ -100,10 +100,44 @@ const CurrencyFactory = ( currencySetting ) => {
...
+
+		if (
+			! Object.keys( symbol ).length ||
```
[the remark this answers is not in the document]
joshuatf: Updated logic a bit in https://github.com/woocommerce/woocommerce-admin/pull/5969/commits/2427114c012f09d56f4d896a9f5f09c86dd55290

PR #5970, opened 2020-12-25
Try: Replace Accordion with Panel component

joshuatf:
Attempts to replace the `Accordion` with the `@wordpress/components` `Panel` component.

I'm not sure behind the decisions to create the `Accordion` component so we'd need to double check before merging this.

Note that a test is failing related to the panel being open.  I think this may be due to how GB sets the open state and it's not ready by the time the tests render.

### Screenshots

**Before**
Screen Shot 2020-12-25 at 3 28 29 PM

**After**

Screen Shot 2020-12-25 at 3 23 24 PM

1. Open the activity panel.
2. Check that styles match expected. (note some are more closely following GB `Panel` component.
3. Try with/without:
* Orders that need fulfilling
* Reviews
* Low stock products


jeffstieler:
The `Accordion` component is built with the `Panel` components - https://github.com/woocommerce/woocommerce-admin/blob/5e31e91dba792b24e00e71d993cd163a6e3dbb88/packages/components/src/accordion/index.js#L18

Is there just a styling fix to be done here?

joshuatf:
@jeffstieler I guess what I'm wondering is what the advantages to the `Accordion` component are.  The `Panel` component already seems to be an accordion-like component, does the `Accordion` component offer some abstraction or other features we plan to reuse?

client/homescreen/activity-panel/index.js
```diff
@@ -42,32 +48,48 @@ export const ActivityPanel = () => {
...
+					>
+						<PanelRow>{ panel }</PanelRow>
+					</PanelBody>
+				) : (
+					<div className="components-panel__body">
+						<h2 className="components-panel__body-title">
+							<Button
+								className="components-panel__body-toggle"
+								aria-expanded={ false }
+								disabled={ true }
```
louwie17, line 84: The disabled styling is causing it to be faded, could we change it, so it isn't faded, just like below?
Screen Shot 2021-01-04 at 3 04 31 PM

Also when opening a panel below this placeholder isn't showing a gap
Screen Shot 2021-01-04 at 3 09 15 PM

louwie17 (dismissed):
Overall great stuff Josh, just left 2 small styling suggestions.

I am fine with just using the `Panel` component, given there aren't many differences, from what I can see.
Looping @octaedro in, as he originally wrote the Accordion panel, to see if we missed anything obvious that the Accordion panel is doing that the Panel component doesn't.

If we do decide to use the Panel component, we should remove the accordion package.

louwie17:
There is also a failed test: https://travis-ci.com/github/woocommerce/woocommerce-admin/jobs/464853272#L3051

client/homescreen/activity-panel/index.js
joshuatf: >The disabled styling is causing it to be faded, could we change it, so it isn't faded, just like below?

I think this is expected behavior.  Disabled buttons should provide visual cues that they are in fact disabled.  This is the default GB Button component behavior.

>Also when opening a panel below this placeholder isn't showing a gap

I did my best to follow the previous designs, but I may have been mistaken or the component was already broken.  I agree that this gap looks odd, but seemed intentional based on the existing css.  /cc @octaedro @elizaan36 @pmcpinto for any hints or reasoning on this gap.

joshuatf:
Removed the `Accordion` component.  Had planned to deprecate, but it appears to not be released in the wild yet.

>There is also a failed test: https://travis-ci.com/github/woocommerce/woocommerce-admin/jobs/464853272#L3051

From above:

>Note that a test is failing related to the panel being open. I think this may be due to how GB sets the open state and it's not ready by the time the tests render.

I'm somewhat inclined to think these tests aren't necessary as this bleeds into coverage around the `Panel` component, but I'll take another stab at this.  Open to thoughts on this if you have ideas on why this is failing.

louwie17:
> I'm somewhat inclined to think these tests aren't necessary as this bleeds into coverage around the `Panel` component, but I'll take another stab at this. Open to thoughts on this if you have ideas on why this is failing.

Sorry missed the original comment above, and it appears it fails because in the mock data the `collapsible` param is not defined, so it is always false - only rendering the header.
Setting collapsible to `true` in the mock data up above fixes it.

timmyc:
I added @elizaan36 to the list of reviewers so we can get some design input here too if needed.

joshuatf:
Good eyes, @louwie17!  Fixed up in https://github.com/woocommerce/woocommerce-admin/pull/5970/commits/9152aca272ac3020e4b91e42fe97d5b8d77da4bb

client/homescreen/activity-panel/index.js
pmcpinto: I agree that the difference between the gaps looks a bit odd. @elizaan36 what do you think?

elizaan36: Yeah, the gap is intentional to indicate which panel is open so there should be a gap between the open panel and closed panels. Originally we were only going to have one open at a time, but currently since they can all be open then there would be the consistent gap between all panels.

elizaan36:
Hey @joshuatf, just checked out the branch and saw a couple anomalies.

- Looks like we lost the horizontal rules in the panels - one dividing the header from the rest of the content, then ones dividing the rows. Here's a screenshot from the design:
image
- The empty panel number count is showing a `,` character that shouldn't be there, and the background is faded to a lighter grey. This should visually match the other number counts for visual consistency on the screen, even though the panel is inactive. Here's a screenshot there: https://share.getcloudapp.com/qGulbgmn

client/homescreen/activity-panel/index.js
joshuatf: >but currently since they can all be open then there would be the consistent gap between all panels.

Does this mean the gap should exist even when they're closed?  Or still only on open panels?

elizaan36: The gap indicates that a panel is open, and appears between adjacent panels when one accordion is expanded. When the accordion is closed and the adjacent panels are also closed (or empty), there's no gap.

joshuatf:
Thanks for the feedback @elizaan36!

>Looks like we lost the horizontal rules in the panels - one dividing the header from the rest of the content, then ones dividing the rows. Here's a screenshot from the design:

Added this in https://github.com/woocommerce/woocommerce-admin/pull/5970/commits/0365a25b76c2a3c17fe301f75d7e360ab32a5169

>The empty panel number count is showing a , character that shouldn't be there

Fixed. 

>background is faded to a lighter grey.

This is the disabled button style from the GB Button component.  I updated in https://github.com/woocommerce/woocommerce-admin/pull/5970/commits/0365a25b76c2a3c17fe301f75d7e360ab32a5169 as well but similar to above with the header border, I'm a little confused as to when we're supposed to leave the styles as is and when we're supposed to override them.  It seems on one hand we're refactoring components to inherit more styles directly from the GB components and on the other we're creating new ones that override these styles.

I guess this is something that may require a lengthier discussion via p2, but I'm curious what the direction is.  My concern is that similar to other components, like the `Card` we'll do one of the following:
* Eventually drop our own component styling in favor of what's in GB.  In this case, we'll have created a decent amount of technical debt to clean up and components that will need to be deprecated.
* We create components that are layered on top of the GB components (what this PR attempts to undo).  This will be unreliable going forward without bundling `@wordpress/components` and breaking changes could be hard to track across different versions of `@wordpress/components`.
* Rewrite our own version of these components.  No need to worry about breaking changes in this case, but more overhead and maintenance and of course the styling will slowly diverge from what's in GB.

client/homescreen/activity-panel/style.scss
```diff
@@ -1,67 +1,61 @@
...
+		margin-bottom: $gap-large;
+	}
+
+	.components-panel__body-title p {
+		margin-right: $gap;
+	}
+
 	.woocommerce-activity-card {
 		padding: $gap-largest/2 $gutter $gutter;
+		border: 0;
```
elizaan36, line 47: I think we need to replace `border: 0` with `border-bottom: 1px solid $gray-200;` here to add the borders back between the rows.

elizaan36:
This is a really great conversation to have, let's chat about it more so that the next approach will be more thought out.

I just commented on the line where I think we can fix the missing borders between the rows, and we chatted in Slack about the padding fix for the empty state. 🥇

client/homescreen/activity-panel/style.scss
joshuatf: Gotcha.  I had removed this because I was seeing double border thickness in some areas.  Added this in with a small tweak:

```scss
		&:not(:last-of-type) {
			border-bottom: 1px solid $gray-200;
		}
```

Borders between activity cards, but not on the last item to avoid thickened borders.

Screen Shot 2021-01-06 at 2 51 15 PM

elizaan36: Nice one👍

client/homescreen/activity-panel/style.scss
```diff
@@ -1,67 +1,70 @@
...
 	}
 
-	.components-panel__body-title {
-		margin: 0;
-		padding: $gap-small 0;
-		height: 60px;
-		.components-button {
+	.components-panel__row {
+		margin: 0 -#{$gap} -#{$gap} -#{$gap};
+		> div {
```
samueljseay, line 24: Fwiw this is the kind of style that can break based on different versions of `@wordpress/components`. I understand sometimes its impossible to avoid, but is there a way you can avoid relying on immediate child div here?

client/homescreen/activity-panel/index.js
```diff
@@ -42,32 +48,48 @@ export const ActivityPanel = () => {
...
+						key={ id }
+						className={ className }
+						initialOpen={ initialOpen }
+						collapsible={ collapsible }
+						disabled={ ! collapsible }
+					>
+						<PanelRow>{ panel }</PanelRow>
+					</PanelBody>
+				) : (
+					<div className="components-panel__body">
```
samueljseay, line 79: Hmm, so this is also something that might break based on version of `@wordpress/components` because we're relying on trying to imitate the existing markup of the panel component. Have you considered manually styling this component? If you provide the necessary styles in our library then it will increase size of styles a bit, but it should not break if the Panel component changes between versions.

samueljseay (changes requested):
Just a couple of concerns on where you're heavily relying on the Panel component not to change, otherwise it looks good.

client/homescreen/activity-panel/index.js
joshuatf: 100% agreed.  Unfortunately, `@wordpress/components` does not expose a way for us to pass props to the button used here (`PanelBodyTitle` is not exported).  Just overriding via CSS I believe creates an accessibility issue.

This is brittle, but I think a change needs to happen upstream to remove this.

client/homescreen/activity-panel/style.scss
samueljseay: Thats all good 👍

client/homescreen/activity-panel/index.js
samueljseay: Sounds like a fair trade off to me.

samueljseay (approved):
👍 Great stuff

joshuatf:
Thanks for all the reviews!  ❤️ 

Opened a PR in GB to allow passing of button props so we don't need to manually recreate the disabled buttons.
https://github.com/WordPress/gutenberg/pull/28147

PR #5973, opened 2020-12-28
Don't show Stock or Reviews panels before there are Products and/or Orders.

### LINKED ISSUE #5952: Inventory and Reviews empty states showing alongside the store setup task list
**Describe the bug**
The Inventory and Reviews empty states are being displayed alongside the store setup task list, which can distract and confuse the user. We can use the same approach that was implemented for the orders task. 

**To Reproduce**

1. Create a new store with 4.9 beta installed
2. Complete or skip the store profiler
3. Visit the home screen
4. See the inventory and reviews empty states.

**Expected behavior**
To only see the store setup tasks, as I'm a new user and I still don't have any products, reviews or orders

**Screenshots**
https://d.pr/i/L1xBMJ

jeffstieler:
Fixes #5952

This PR seeks to fix the issue observed in #5952 where empty Stock and Reviews panels were displayed on the Homescreen of a brand new store.

It changes the logic of when to include the Stock and Reviews panels to the following:

Stock: inventory management is enabled, total order count is 1 or more, published product count is 1 or more

Reviews: reviews are enabled, published product count is 1 or more

* Create a new store
* Complete or skip the store profiler
* Visit the home screen
* Verify that the Stock and Reviews panels are not visible
* Switch to an existing store with products + orders, with reviews and inventory enabled
* Visit the home screen
* Verify that the Stock and Reviews panels are visible

### DIFF
--- a/client/homescreen/activity-panel/index.js
+++ b/client/homescreen/activity-panel/index.js
@@ -27,6 +27,7 @@ export const ActivityPanel = () => {
 		const manageStock = getSetting( 'manageStock', 'no' );
 		const countLowStockProducts = getLowStockCount( select );
 		const countUnapprovedReviews = getUnapprovedReviews( select );
+		const publishedProductCount = getSetting( 'publishedProductCount', 0 );
 
 		return {
 			countLowStockProducts,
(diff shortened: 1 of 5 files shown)


moon0326 (approved):
Tested well and LGTM :shipit:

Issue #5918, opened 2020-12-21
Help tab inconsistent with site status


rrennick:
**Describe the bug**
I noticed two issues (see screenshot):

- In my local testing container I have the option to connect to wordpress.com onboarding. The site isn't on wordpress.com infrastructure and does not have Jetpack installed.
- I still have the option to `Disable` the task list after I have dismissed/hidden it on the home screen. Clicking `Disable` in the help screen doesn't toggle the setting so I have the option to re-enable the task list. 

**To Reproduce**

1. Go to WooCommerce -> Settings -> Help -> Setup Wizard
2. Click on `Disable` task list

**Expected behavior**
There should be a method a site owner could use to re-enable the task list. The WordPress.com option should only be shown on connected sites.

**Screenshots**
<img width="576" alt="Screen Shot 2020-12-21 at 1 53 58 PM" src="https://user-images.githubusercontent.com/343847/102806925-1ca92500-4394-11eb-9ac7-eb53a426716a.png">

timmyc:
Quick note on some background of the feature via @joshuatf - this is a remnant from the initial development phase of the OBW when there was a good deal of work happening around Jetpack connectivity. As such, seems the most logical thing to do is to remove this.

jeffstieler:
The Calypso section is included when the store has `WP_DEBUG` enabled. Removal should be super low impact.

> There should be a method a site owner could use to re-enable the task list

Perhaps we should move the setup wizard and task list toggles to the "help" menu on the homescreen?

PR #5975, opened 2020-12-29
Fix task list toggle logic in help tab.

jeffstieler:
Partially addresses #5918.

This PR fixes a logic error when checking the status of the Task List - there was JS syntax in place that doesn't work the same way in PHP.

- Hide your Task List from the home screen 
- Go to a non-JS page, like WooCommerce > Orders (if you have orders!)
- Open the Help tab, go to "Setup wizard"
- Note that the Task List toggle button says "enable"
- Click to enable the task list
- Go to the homescreen
- Verify your Task List is visible again
- Go to a non-JS page, like WooCommerce > Orders (if you have orders!)
- Open the Help tab, go to "Setup wizard"
- Note that the Task List toggle button says "disable"

Fix: Task List toggle in help panel.

### DIFF
--- a/src/Features/Onboarding.php
+++ b/src/Features/Onboarding.php
@@ -948,7 +948,10 @@ class Onboarding {
 			'id'    => 'woocommerce_onboard_tab',
 		);
 
-		$task_list_hidden = get_option( 'woocommerce_task_list_hidden', 'no' ) || get_option( 'woocommerce_extended_task_list_hidden', 'no' );
+		$task_list_hidden = (
+			'yes' === get_option( 'woocommerce_task_list_hidden', 'no' ) ||
+			'yes' === get_option( 'woocommerce_extended_task_list_hidden', 'no' )
+		);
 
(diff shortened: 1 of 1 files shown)


joshuatf (commented):
Great catch, Jeff!

Wondering if we can use `should_show_tasks()` here to create a single source of truth for whether or not the task list is shown.

https://github.com/woocommerce/woocommerce-admin/blob/66dd1c9c88433c22b6dd752aa3c0917df097d5f3/src/Features/Onboarding.php#L358

jeffstieler:
> Wondering if we can use `should_show_tasks()`

Nice find - though is the logic faulty there? Should those comparisons be `AND`ed instead of `OR`ed?

joshuatf:
Had a slack convo with @jeffstieler on this (see p1609427338013600-slack-team-isotope).  This logic can probably be simplified in a follow-up, but for now we'll treat `woocommerce_task_list_hidden` and `woocommerce_extended_task_list_hidden` as two separate lists.

jeffstieler:
@joshuatf it looks like the logic already toggle both the "core" list and the "extended" list - so this PR should be good to go, just need that ✅  😄

samueljseay (approved):

PR #5981, opened 2020-12-29
Add label-sync package and config.

timmyc:
Context: p90Yrv-21u-p2

This branch adds the `github-label-sync` and some npm scripts copy/pasted from our friends in Rubik 🌟 to keep the wc-admin labels in sync with the new standard Woo labels.

Nothing to test here, but I suppose you could run the `npm run labels:dry` and note that there shouldn't be any changes that need to be made since I just ran the sync :)


jeffstieler (approved):
Looks good! :shipit: (tested `npm run labels:dry`)

PR #5986, opened 2020-12-30
Reduce Unnecessary Re-renders in Revenue Report

jeffstieler:
This PR seeks to reduce the number of unnecessary re-renders in the Revenue Report and serve as an example for reducing renders in other reports.

It primarily focuses on reducing the number of objects that are equivalent (same data), but fail a strict equals (`===`) comparison because they are new instances/references.

Lodash's `memoize` was used because it was there. It should be easily replaced if we want to remove the dependency on Lodash.

In order to discover/diagnose props that were considered changing (but not changing equivalent values) I used the React DevTools Profile with the "record why" option: 
Screen Shot 2020-12-30 at 9 37 37 AM

I noticed that props like `headers` and `summaryFields` were causing renders, but inspection/logging showed the logical values weren't changing. The exercise then was to stop the references from changing by returning the same object where possible. This required varying strategies of common object references (like `EMPTY_OBJECT`, as seen in Gutenberg) and memoization of selectors / data manipulation functions where possible.

Much, much more can be done here. I think the biggest opportunities are in memoizing our selectors where possible.

### Screenshots

Reduction in `ReportTable` renders, using React DevTools Profiler:

| Before | After |
| --- | --- |
| Screen Shot 2020-12-30 at 9 20 07 AM | Screen Shot 2020-12-30 at 9 23 59 AM |
| | Scripting, Rendering, System, and blocking time are all down so I think this is an improvement |
| Screen Shot 2020-12-30 at 8 57 41 AM | Screen Shot 2020-12-30 at 9 07 56 AM |
| 35 renders, 342.7 ms | 6 renders, 139.7 ms |

- Ensure your test store has `WP_DEBUG` and `SCRIPT_DEBUG` on in `wp-config.php`
- Use `npm start` so you're using a development build of React (needed to support Profiling)
- Thoroughly test the Revenue Report (and another report of your choosing)
- Test changing date ranges, all report modes (single,comparison,advanced)
- Test comparison using checkboxes in the table
- Test table search
- Test table pagination
- Test table sorting

### DIFF
--- a/client/analytics/components/report-table/index.js
+++ b/client/analytics/components/report-table/index.js
@@ -567,6 +567,9 @@ ReportTable.defaultProps = {
 	baseSearchQuery: {},
 };
 
+const EMPTY_ARRAY = [];
+const EMPTY_OBJECT = {};
+
 export default compose(
 	withSelect( ( select, props ) => {
 		const {
@@ -587,7 +590,7 @@ export default compose(
 			( query.search &&
 				! ( query[ endpoint ] && query[ endpoint ].length ) )
 		) {
-			return {};
+			return EMPTY_OBJECT;
 		}
 		const { woocommerce_default_date_range: defaultDateRange } = select(
 			SETTINGS_STORE_NAME
@@ -603,11 +606,10 @@ export default compose(
 					select,
 					filters,
 					advancedFilters,
-					tableQuery,
 					defaultDateRange,
 					fields: summaryFields,
 			  } )
-			: {};
+			: EMPTY_OBJECT;
 		const queriedTableData =
 			tableData ||
 			getReportTableData( {
@@ -632,9 +634,9 @@ export default compose(
 					? extendedTableData.items.data.map(
 							( item ) => item[ itemIdField ]
 					  )
-					: [],
+					: EMPTY_ARRAY,
 			tableData: extendedTableData,
-			query: { ...tableQuery, ...query },
+			query,
 		};
 	} ),
 	withDispatch( ( dispatch ) => {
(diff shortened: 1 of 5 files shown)


client/analytics/components/report-table/index.js
```diff
@@ -587,7 +590,7 @@ export default compose(
 			( query.search &&
 				! ( query[ endpoint ] && query[ endpoint ].length ) )
 		) {
-			return {};
+			return EMPTY_OBJECT;
```
joshuatf, line 593: That's a tricky difference in props.  Nice find!

client/analytics/report/revenue/table.js
```diff
@@ -6,7 +6,7 @@ import { Component } from '@wordpress/element';
 import { format as formatDate } from '@wordpress/date';
 import { withSelect } from '@wordpress/data';
 import { compose } from '@wordpress/compose';
-import { get } from 'lodash';
+import { get, memoize } from 'lodash';
```
joshuatf, line 9: We do have `memoize-one` which we could optionally move to, but there are still many places `lodash` is used, so I'm not opposed to saving this for a follow-up when we're ready to tackle that.

joshuatf (approved):
Awesome performance enhancements!  Much less flashing when viewing component rerenders in dev tools.

Left one optional comment about memoization, but otherwise this LGTM!

client/analytics/report/revenue/table.js
jeffstieler: Unfortunately, `memoize-one` isn't a direct replacement and works fundamentally differently than most memoization libraries - it will only return cached values if subsequent calls receive the same arguments (`===` comparison) and there is no way to customize the cache "key" like I'm doing with Lodash in this PR.

joshuatf:
Since this PR introduces quite a few changes to reports, having a couple more sets of eyes on this to smoke test would be good prior to merge.

joelclimbsthings:
This looks great, and it's something I'd wanted to learn more on so parsing the code is doing me some good. I've added some test data and given the revenue report a smoke test, and everything looks right thus far. 

Nice work.

joelclimbsthings (approved):

Issue #5989, opened 2020-12-31
woocommerce_rest_prepare_product_object called when response is not product object


leewillis77:
**Describe the bug**
The WooCommerce admin page triggers a call to the following URI:
/wp-json/wc-analytics/products?page=1&per_page=1&low_in_stock=true&status=publish&_fields%5B0%5D=id&_locale=user

If that returns matching products, then the `woocommerce_rest_prepare_product_object` filter is called for each of the products in order to enrich the response object. However, in this instance the response isn't a product object at all, it's just a product ID. 

It seems that either:

- A different filter should be invoked when the response is a set of product fields, rather than a full product object. That filter should include the response assembled by WooCommerce, the object, the request, and the requested fields
- There should be an easy way of retrieving the list of desired fields given the inputs available to that filter (as far as I can tell there's no easy way to get the list of requested fields given the current data passed to the filter without re-parsing the request?)

chihsuan:
I'm closing this because the `woocommerce_rest_prepare_product_object` filter doesn't exist in our repo now.

PR #5990, opened 2020-12-31
Fix undefined index error

leewillis77:
The WooCommerce admin page triggers a call to the following URI:
/wp-json/wc-analytics/products?page=1&per_page=1&low_in_stock=true&status=publish&_fields%5B0%5D=id&_locale=user

This currently throws the following error for each product returned as the code assumes that "name" is included in the response (It's a similar issue #5989). 

This PR fixes this issue by only attempting to call `wp_strip_all_tags` on the name if it is present. 

Changelog note:
Fix: Undefined index error thrown on WooCommerce admin page for each low stock item

### DIFF
--- a/src/API/Products.php
+++ b/src/API/Products.php
@@ -157,7 +157,9 @@ class Products extends \WC_REST_Products_Controller {
 				$data->data['last_order_date'] = wc_rest_prepare_date_response( $this->last_order_dates[ $product_id ] );
 			}
 		}
-		$data->data['name'] = wp_strip_all_tags( $data->data['name'] );
+		if ( isset( $data->data['name'] ) ) {
+			$data->data['name'] = wp_strip_all_tags( $data->data['name'] );
+		}
 
 		return $data;
 	}


joshuatf (commented):
Code here looks good and not opposed to merging this in as it seems safe, but even when specifying a field of `id`, I'm still able to access the `name` property prior to server response so I never encounter the error.

@leewillis77 could you share any more details on your testing to receive this error?  Do you have a nameless product that causes this error?

leewillis77:
Create a product in WooCommerce that has "low stock" (stock level is less than the "low stock threshold" as set in WooCommerce » Settings » Products » Inventory), and then access the WooCommerce page in your WordPress admin area. It will trigger a call to the following endpoint which queries for low stock products, but *only* returns their IDs:

`/wp-json/wc-analytics/products?page=1&per_page=1&low_in_stock=true&status=publish&_fields%5B0%5D=id&_locale=user`

That generates the following warning:
`PHP Notice:  Undefined index: name in /Users/lee/sites/wc/wp-content/plugins/woocommerce/packages/woocommerce-admin/src/API/Products.php on line 160`

That's caused by the fact that the REST response only contains the product ID, not the full product object.

joshuatf:
@leewillis77 Right, I understand the endpoint and need for the low stock product.  What I'm having trouble reproducing is the error since the area you've linked to actually has access to the full product object (not just the ID).

>That's caused by the fact that the REST response only contains the product ID, not the full product object.

The actual response only contains the ID, but `prepare_object_for_response()` should have access to all info.  The response fields are limited by wp core's rest controller method:

https://github.com/WordPress/WordPress/blob/2818d14fa82b9cbe9fa65da50bf55671434bb273/wp-includes/rest-api/endpoints/class-wp-rest-controller.php#L422

leewillis77:
The issue is that the line in question (https://github.com/woocommerce/woocommerce-admin/blob/main/src/API/Products.php#L160) is operating on the data (`$data`) prepared for handing through to the response.

Since the request was for IDs only, then the response data in `$data` doesn't include the product name, and therefore doesn't need escaping.

joshuatf:
That is incorrect. $data will include all properties at that point in the
response as they have not been filtered down by requested fields yet.

If your name property is missing, it is because of something else affecting
your product name (possibly a missing product name).

Again, that’s not to say this isn’t possibly a good change but at a minimum
we need a way to reproduce the error. This is also useful in determining if
some other error has occurred upstream that needs to be fixed.

leewillis77:
> That is incorrect. `$data` will include all properties at that point in the response as they have not been filtered down by requested fields yet.

That's not my experience. At that point, the data has been processed by `WC_REST_Products_V2_Controller::prepare_object_for_response() which filters the fields, and it *only* contains the id field. See below dump of `$data` straight after https://github.com/woocommerce/woocommerce-admin/blob/main/src/API/Products.php#L149

```
Products.php: (149) : data is WP_REST_Response Object
(
    [links:protected] => Array
        (
            [self] => Array
                (
                    [0] => Array
                        (
                            [href] => http://wc.test/wp-json/wc-analytics/products/30
                            [attributes] => Array
                                (
                                )

                        )

                )

            [collection] => Array
                (
                    [0] => Array
                        (
                            [href] => http://wc.test/wp-json/wc-analytics/products
                            [attributes] => Array
                                (
                                )

                        )

                )

        )

    [matched_route:protected] => 
    [matched_handler:protected] => 
    [data] => Array
        (
            [id] => 30
        )

    [headers] => Array
        (
        )

    [status] => 200
)
```

> If your name property is missing, it is because of something else affecting your product name (possibly a missing product name).

No, the product with the ID returned by that endpoint is complete.

> Again, that’s not to say this isn’t possibly a good change but at a minimum we need a way to reproduce the error.

The instructions I gave previously reproduce this consistently on my setup (WooCommerce Admin as included with WooCommerce 4.9.0-rc.1 if that makes a difference). I've tried it with three different products being out of stock to rule it out being anything specific to that product.

joshuatf (approved):
>WooCommerce Admin as included with WooCommerce 4.9.0-rc.1 if that makes a difference

This is the issue.  Using the beta version of WC 4.9 I'm able to consistently reproduce the issue.

I'll investigate the changes in WC to make sure they don't affect other endpoints, but this PR looks safe to merge.

Thank you for the PR, @leewillis77.

PR #5992, opened 2021-01-04
Set `$collate` to an empty string in case a database backend doesn't support collation

Krovatkin:
Fixes #

`$collate` variable is undefined while trying to install Woocommerce

### Screenshots

image

- Copy this file https://github.com/aaemnnosttv/wp-sqlite-db/blob/master/src/db.php into `wp-content`
- Copy `wp-config-sample.php` to `wp-config.php`. Use this file as an example:  https://github.com/aaemnnosttv/wp-sqlite-db/blob/master/tests/wp-config.php
- Install Woocommerce

I would really appreciate some feedback on mysterious Travis CI errors and suggestions on how to add a simple test to test this change! Thanks!

PS I also grepped for `$wpdb->has_cap( 'collation' )` and this seems to be the only unhandled occurrence.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -73,6 +73,7 @@ Release and roadmap notes are available on the [WooCommerce Developers Blog](htt
 
 == Unreleased ==
 
+- Fix: Undefined $collate variable when database does not have collation capability. #5992
 - Tweak: Remove visit_count from track, and update task count logic. #5996
 - Fix: Moved certified owner label for review to title. ##5877
 - Fix: Move collapsible config to panels object, to allow for more control. #5855
--- a/src/Install.php
+++ b/src/Install.php
@@ -188,9 +188,7 @@ class Install {
 	protected static function get_schema() {
 		global $wpdb;
 
-		if ( $wpdb->has_cap( 'collation' ) ) {
-			$collate = $wpdb->get_charset_collate();
-		}
+		$collate = $wpdb->has_cap( 'collation' ) ? $wpdb->get_charset_collate() : '';
 
 		// Max DB index length. See wp_get_db_schema().
 		$max_index_length = 191;


adrianduffell:
Hi @Krovatkin , thanks for this PR! Overall the change looks good to me. Nice catch discovering this!

The `wp-sqlite-db` project looks interesting. I am intrigued how you might be using it? 

Could you also please add the following changelog entry in `readme.txt`, under the [Unreleased heading](https://github.com/woocommerce/woocommerce-admin/blob/497c9d3624300f9d59c3b8d92a8315beada820a4/readme.txt#L74)?

```
- Fix: Undefined $collate variable when database does not have collation capability. #5992
```

> I would really appreciate some feedback on mysterious Travis CI errors and suggestions on how to add a simple test to test this change! Thanks!

I'm seeing errors in Travis CI build that were recently fixed when we upgraded to Composer v2. It looks like your fork is a few months out of date and is missing the fix. I think [syncing your fork](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/syncing-a-fork) and rebasing your branch with `main` will sort it out.

As for adding a new test, this doesn't look simple to do 😅

It looks like `Install:create_tables` needs some tests written. For those, I think `$wpdb` needs to be mocked to control the `has_cap( 'collation' )` and `get_charset_collate` return values. Then, the results from the `dbDelta` call would need to be captured and asserted. I am happy to work on this at a later date if you prefer.

adrianduffell (changes requested):
Code looks good 👍  This can be merged once the Travis tests pass and a changelog entry is added.

Krovatkin:
Hi @adrianduffell !

Thanks for getting back to me on this so quickly!
Also, thank you so much for such a great e-commerce platform!!!!

I added an entry to `readme.txt`. The PR is rebased onto the right default and CI is passing now

> The wp-sqlite-db project looks interesting. I am intrigued how you might be using it?

I wanted a very simple sandbox environment to try out WooCommerce platform, to get up and running fast, so I used this plugin and php's built-in webserver 😅 

> As for adding a new test, this doesn't look simple to do 😅 It looks like Install:create_tables needs some tests written. For those, I think $wpdb needs to be mocked to control the has_cap( 'collation' ) and get_charset_collate return values. Then, the results from the dbDelta call would need to be captured and asserted. I am happy to work on this at a later date if you prefer.


Haha, I was **really** scared someone might say exactly that. If you could follow up with test infra + test in another PR, that would be ideal!

adrianduffell (approved):
This tests well and looks good to me 🚀

adrianduffell:
Hi @Krovatkin

Thanks for putting in place those final items! I have merged this now and will will follow up with some tests for `Install:create_tables` in the future.

I like your idea to use wp-sqlite-db with PHP’s built-in web server. I hope to try that out some time too 🙂

PR #6001, opened 2021-01-05
Fix visual glitches on OBW when viewport is small

### LINKED ISSUE #5979: OBW: Product Types Screen Visual Glitches on Mobile
**Describe the bug**
When using the OBW, and on the Product Type screen, the Customizable Products line wraps and creates a bit of visual jank:

obw-product-types-mobile

It looks a little bit better when viewing prices in annual mode:

obw-mobile-types-annual

**Smartphone (please complete the following information):**

* Device: Pixel 5
* OS: Android 11
* Browser stock chrome

moon0326:
Fixes #5979 

This PR addresses visual glitches when the user's viewport is less than 438px and 375px (iPhone 6/7/8)

We still have visual glitches for devices with a viewport less than 354px. I think the ultimate solution would be reducing the overall font-size to adapt to the viewport size, but that requires more tests and discussions. I will open a new PR if we decide to go that route.

The changes in this PR fix the visual glitches for any major mobile phones, except for the iPhone 5 and SE (the first edition).

### Screenshots

Before:

Screen Shot 2021-01-04 at 4 31 52 PM

After:

Screen Shot 2021-01-04 at 4 00 50 PM

1. Install WooCommerce and start OBW
2. Start OBW and continue to the Product Types 
3. Open your browser console and change the view mode to mobile
4. Test with pre-defined mobile devices.
5. Confirm the fix.

### DIFF
--- a/client/profile-wizard/steps/product-types/style.scss
+++ b/client/profile-wizard/steps/product-types/style.scss
@@ -67,3 +67,23 @@
 		}
 	}
 }
+
+@media screen and (max-width: 438px) {
+	.woocommerce-profile-wizard__body {
+		.woocommerce-profile-wizard__product-types {
+			.woocommerce-product-wizard__product-types-label {
+				width: min-content;
+			}
+		}
+	}
+}
+
+@media screen and (max-width: 375px) {
+	.woocommerce-profile-wizard__body {
+		.woocommerce-profile-wizard__product-types {
+			.woocommerce-pill {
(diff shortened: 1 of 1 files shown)


ilyasfoo (approved):
Tests well! 🚢 

I agree with your decision, and I also think supporting a screen smaller than 354px is not worth the investment, this fix should be sufficient for most of our users. Plus the visual bug is not to the extent that it becomes unusable.

PR #6004, opened 2021-01-05
Add welcome modal when coming from Calypso

### LINKED ISSUE #48420: Store|Woo: Provide additional guidance once a Store customer starts using Woo Core
**User story:**
As a Store user
I want to easily access information about the Core interface once I transition from the old UI
So I can easily get adjusted to the new ui without going back to get more information

**Description:** we should make sure that we welcome the user when he arrives at the Core home screen and that he can easily access the support docs without going back to the Store UI

**Acceptance criteria:**
- [ ] Welcome the user when the arrives at the home screen for the first time (perhaps we can take advantage of the existing modal)
- [ ] Allow the user to easily access the support doc that we provide without going back to the Store UI (an inbox notification seems to be the best option)
- [ ] This should only target existing Store users. As the new ones should be automatically redirected to the Core UI once we start the sunset period (I'll open an issue about this)
- [ ] Should we have a way to allow customers to easily go back to the old UI? A CTA in an inbox message?

becdetat:
Fixes https://github.com/Automattic/wp-calypso/issues/48420

The intent of https://github.com/Automattic/wp-calypso/issues/48420 is to show a welcome modal when entering WooCommerce Admin from Calypso. https://github.com/automattic/wp-calypso/issues/48420 adds `&from-calypso` to the URL to WCA, and this PR handles that URL param to display the welcome modal.

### Screenshots
image
*edit* Updated with final copy

- Starting with a fresh store (or by deleting the `woocommerce_task_list_welcome_modal_dismissed` option), visit `/wp-admin/admin.php?page=wc-admin`. You should see the standard welcome modal.
- Add `&from-calypso` to the URL. You should see the Calypso welcome modal.
- Dismiss the modal
- Refresh the page. The modal should not reappear.
- Remove the `&from-calypso` from the URL. The standard welcome modal should not reappear.


becdetat:
@pmcpinto I've added in two tracks events in line with the events created by the existing welcome modal:

- `welcome_from_calypso_modal_open`
- `welcome_from_calypso_modal_close`

pmcpinto:
@becdetat thanks for the heads up. Just a small detail, I think that we should follow the Tracks naming convention and prefix those event names with `wcadmin_` eg: `wcadmin_welcome_from_calypso_modal_open`

becdetat:
@pmcpinto 
> I think that we should follow the Tracks naming convention and prefix those event names with `wcadmin_`
Oh, sorry, the prefix is automatically added by the tracks library so I forgot to include it.

client/homescreen/layout.js
```diff
@@ -185,7 +220,10 @@ export default compose(
 		);
 
 		const shouldShowWelcomeModal =
-			welcomeModalDismissedHasResolved && ! welcomeModalDismissed;
+			welcomeModalDismissedHasResolved &&
+			! welcomeModalDismissed &&
+			welcomeFromCalypsoModalDismissedResolved &&
+			! welcomeFromCalypsoModalDismissed;
```
louwie17, line 226: Should we change the last item here to `! shouldShowWelcomeFromCalypsoModal` instead of `! welcomeFromCalypsoModalDismissed`.
I don't know how common the case is that the user visits Woo admin for the first time coming from calypso, but this would trigger both `shouldShowWelcomeModal` and `shouldShowWelcomeFromCalypsoModal` to be true at the same time.

Actually causing a strange effect, where the welcome modal gets quickly loaded and closed, and then showing the calypso modal, as you can see from the tracks that it generated (on initial window load).
Triggering the open track for both modal's, but then immediately closing the welcome modal. 
Screen Shot 2021-01-06 at 9 24 05 AM

@becdetat would it be worth in this case showing the welcome modal, but replacing the first page with the Calypso page?

louwie17 (changes requested):
Code looks great! I just left one comment for a particular use case where the user visits the WooCommerce store for the first time from Calypso, causing some weird things behind the scenes, see comment below.

client/homescreen/layout.js
louwie17: > Code looks great! I just left one comment for a particular use case where the user visits the WooCommerce store for the first time from Calypso, causing some weird things behind the scenes, see comment below.

cc @pmcpinto to see if this use case might actually happen.

pmcpinto: @becdetat @louwie17 if he's a brand new user he shouldn't see the modal because he never interacted with the Store UI. Makes sense? Or maybe I'm missing something here.

cc @elizaan36 for additional feedback

louwie17: > if he's a brand new user he shouldn't see the modal because he never interacted with the Store UI.

@pmcpinto which modal are you referring to here that shouldn't be shown? The calypso welcome modal or the original welcome modal (the one with 3 pages, mentioning the inbox, and performance stats).

I didn't test this with Calypso, but the way the logic is set up, is that a brand new user would only see the single page modal - Calypso welcome screen (new store management experience).

becdetat:
I've updated this with the final copy

client/homescreen/layout.js
becdetat: @louwie17 
> the way the logic is set up, is that a brand new user would only see the single page modal - Calypso welcome screen (new store management experience)

that was certainly my intent as well. A new user from Calypso should only see the new Calypso welcome modal, and never the original welcome modal. This corresponds with @elizaan36 's design for the new Calypso welcome modal as well.

elizaan36: Just to clarify, the Calypso welcome modal would only be relevant for the existing Calypso users, so anyone that's been a Store user or becomes a Store user during the sunset period. 

If we're talking about brand-new Store users *after* Store sunset, then they should see the regular multi-page welcome modal upon landing on the Home screen. That may be the logic we have, but just wanted to clarify.

elizaan36:
Modal looks great! One super small thing - can we remove the period `.` from the headline in the modal?

becdetat:
@elizaan36 
> can we remove the period . from the headline in the modal?

We can, however the existing welcome modal screens all have periods in the headline so I added it for consistency. I'm happy to remove it if you want though.

elizaan36:
No worries, it's a really small thing. I may go on a period-removing rampage at some point.

client/homescreen/layout.js
pmcpinto: @louwie17 I'm mainly talking about what Elizabeth mentioned here:
 
> If we're talking about brand-new Store users after Store sunset, then they should see the regular multi-page welcome modal upon landing on the Home screen. That may be the logic we have, but just wanted to clarify.

If a user signs up during the sunset period, they should also see the regular welcome modal. Makes sense?

louwie17: Great, thanks for the clarification everyone. I believe everything is working as expected then.
Approving this PR.

readme.txt
```diff
@@ -83,6 +83,7 @@ Release and roadmap notes are available on the [WooCommerce Developers Blog](htt
 - Fix: Updating (non wordpress user) customer with order data
 - Dev: Add documentation for filter `woocommerce_admin_pages_list` and `wc_admin_register_page` #5844
 - Dev: Revert work done in #4857 for automated shipping after OBW is completed #5971
+- Add: Welcome modal when coming from Calypso
```
louwie17, line 86: Just add the PR number here before merging.

louwie17 (approved):
After clarification, everything tested well, good to 🚢  👍 
Just mentioned if you could add the PR number to the readme before merging :)

readme.txt
becdetat: Whoops, thanks

PR #6007, opened 2021-01-05
Add feature toggle and extensibility

joshuatf:
Adds the ability to extend the features list from within features.

Also turns off the nav feature in core. /cc @psealock to verify that this is correct.

### Screenshots

Screen Shot 2021-01-05 at 5 31 24 PM

1. Run the dev version of WCA.
2. Make sure you can see the features section and navigation toggle at `wp-admin/admin.php?page=wc-settings&tab=advanced&section=features`
3. Build a test plugin version from this branch.
4. Make sure that the nav toggle can be seen.
5. Build the core version of this branch.
6. Make sure the feature section does not exist.

### DIFF
--- a/config/core.json
+++ b/config/core.json
@@ -10,7 +10,7 @@
 		"marketing": true,
 		"minified-js": false,
 		"mobile-app-banner": true,
-		"navigation": true,
+		"navigation": false,
 		"onboarding": true,
 		"remote-inbox-notifications": true,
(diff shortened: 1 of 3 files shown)


psealock (approved):
>Also turns off the nav feature in core. /cc @psealock to verify that this is correct.

Yup, thats correct.

Code looks good 🚢

PR #6014, opened 2021-01-06
Remove HomeScreenFeedback note

### LINKED ISSUE #5938: Remove inbox message: HomeScreenFeedback.php
**Description:** we already collected enough qualitative feedback about the home screen, so the [HomeScreenFeedback.php notification ](https://github.com/woocommerce/woocommerce-admin/blob/main/src/Notes/HomeScreenFeedback.php)can be removed.

moon0326:
Fixes #5938 

This PR removes HomeScreenFeedback.php note and its references.

1. Empty `wp_wc_admin_notes` table.
2. Navigate to WordPress -> Plugins and deactivate WooCommerce admin then activate it again
3. Check `wp_wc_admin_notes` table and confirm that there is no record of HomeScreenFeedback


ilyasfoo (commented):
Changes look good! Since we're removing a note, I think it would make sense to add the note to [this array](https://github.com/woocommerce/woocommerce-admin/blob/bbf88f7b0911a2d04bfc4c0e66e152357dbe4c22/src/Install.php#L535-L546) so that existing users get the note deleted from the database when updating.

moon0326:
@ilyasfoo thank you so much for the suggestion. I wasn't aware of that 👍  I will make the change.

samueljseay (approved):
Great!

PR #5922, opened 2020-12-21
Add merchant email notifications

octaedro:
Fixes #5300

This PR adds a new type of admin notification. With this type of notification, it will be possible to send notes via email. The procedure to create them is the same as the other ones.

**A new type**
Now it's possible to create notes type: `email`. That kind of note is sent to a certain role (`administrator` by default). Also, it's possible to extend this functionality by adding new email templates.

**A new status**
The status `sent` was added for the notes type `email`. As soon as the note is sent it's status is set as `sent`. The other statuses work as usual, so if the CTA is activated the note status will be `actioned` (the action triggering will be [addressed in a follow-up PR](https://github.com/woocommerce/woocommerce-admin/pull/5976), here we only send the CTA link but the click is not being tracked).

### Accessibility

-   [x] I've tested using only a keyboard (no mouse)
-   [ ] I've tested using a screen reader
-   [x] All animations respect [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion)
-   [x] All text has [at least a 4.5 color contrast with its background](https://webaim.org/resources/contrastchecker/)

### Screenshots
screenshot-one wordpress test-2020 12 30-14_26_49

### Detailed test instructions:
_In order to test this functionality I created the branches `add/5300_test` and `add/5300_test_plain`, with notes type `email`._

- Checkout the branch "add/5300_test" (`git checkout add/5300_test`).
- Create a zip for testing with `npm run zip:test`.
- Create a `jurassic.ninja` instance (be sure to check the `Include WP Log Viewer plugin` option in the site settings).
- Upload the plugin and activate it.
- Run the cron (this tool can help [WP Crontrol](https://wordpress.org/plugins/wp-crontrol/)).
- You should have received an email, don't open it yet.
- Go to the JN site `debug.log` file. Verify that it's empty.
- Open the email and verify that the JN site `debug.log` file says `Email opened`.

**Plain email**
- In order to force a plain email I created the branch `add/5300_test_plain` so checkout that branch.
- Create the new .zip from this branch and upload it. 
- Run the cron and verify you received the email as expected.
- No opening validation will be triggered here.

PR #6024, opened 2021-01-07
Add your first product: email notification

octaedro:
This is a follow-up of [PR 5922](https://github.com/woocommerce/woocommerce-admin/pull/5922).

This PR adds a note type `email`.
This email note should be sent:
- after three days.
- if the user doesn't have any products.

### Accessibility

### Screenshots
screenshot-one wordpress test-2021 01 08-11_52_25

- Checkout this branch.
- Create a zip for testing with `npm run zip:test`.
- Update the installation date (we need a 3-day old store). You can do it with an SQL sentence like this:

```
UPDATE `wp_options` SET `option_value`=UNIX_TIMESTAMP(DATE_SUB(NOW(), INTERVAL 30 day)) WHERE `option_name` = 'woocommerce_admin_install_timestamp';
```

- Run the cron (this tool can help [WP Crontrol](https://wordpress.org/plugins/wp-crontrol/)).
- Verify the note's status is `sent`. You can use a sentence like this:

```
SELECT `status` FROM `wp_wc_admin_notes` WHERE `name` = 'wc-admin-add-first-product-note'
```

- Go back to the email and click `Add a product`.
- It should take you to the product page.
- Now the note's status should be `actioned`.


moon0326 (commented):
LGTM and tested well 👍 , but the color on the email looks different. This is what I received

Screen Shot 2021-01-08 at 1 13 56 PM

octaedro:
Hi @moon0326

> LGTM and tested well 👍 , but the color on the email looks different. This is what I received
> 
> Screen Shot 2021-01-08 at 1 13 56 PM

That's not a bug. It's inheriting the styles from WooCommerce emails (you can change that in `/wp-admin/admin.php?page=wc-settings&tab=email`).

moon0326 (approved):
Hi @octaedro,

That makes sense :) Thank you for letting me know.

LGTM :shipit:

PR #6032, opened 2021-01-08
Add a new note a day after adding the first product

### LINKED ISSUE #5956: New message: How to customize your product catalog
**Title:** How to customize your product catalog
**Copy:** You want your product catalog and images to look great and align with your brand. This guide will give you all the tips you need to get your products looking great in your store.
**Segmentation:** new users
**Timing:** 1 day after adding the first product
**Cadence:** Once
**CTA leads to:** https://docs.woocommerce.com/document/woocommerce-customizer/?utm_source=inbox
**CTA label:** Learn more

moon0326:
Fixes #5956 

This PR adds a new note with a link to the customizer doc a day after adding the first product.

### Screenshots

Screen Shot 2021-01-08 at 10 15 42 AM

Make sure your store is less than 14 days old. If it has been active for more than 14 days, run the following query.

```update wp_options set option_value = UNIX_TIMESTAMP() where option_name='woocommerce_admin_install_timestamp'```

1. Start with a fresh installation of WooCommerce or delete all published products.
2. Add a new product.
3. Open `wc_posts` and find the product you just added in step #2. Update the `post_date` and `post_date_gmt` to the past date (at least -1 day)
2. Install & activate "WP Crontrol" plugin
3. Navigate to Tools -> Cron Events and run `wc_admin_daily` job
4. Navigate to WooCommerce -> Home and confirm a new note with the title "How to customize your product catalog"

### DIFF
--- a/src/Events.php
+++ b/src/Events.php
@@ -142,11 +142,6 @@ class Events {
 		AddFirstProduct::possibly_add_note();
 		AddingAndManangingProducts::possibly_add_note();
 		CustomizingProductCatalog::possibly_add_note();
-
-		if ( $this->is_remote_inbox_notifications_enabled() ) {
-			DataSourcePoller::read_specs_from_data_sources();
-			RemoteInboxNotificationsEngine::run();
-		}
 	}
 
 	/**
--- a/src/Events.php
+++ b/src/Events.php
@@ -11,6 +11,7 @@ defined( 'ABSPATH' ) || exit;
 use \Automattic\WooCommerce\Admin\Notes\AddingAndManangingProducts;
 use \Automattic\WooCommerce\Admin\Notes\ChooseNiche;
 use \Automattic\WooCommerce\Admin\Notes\ChoosingTheme;
+use \Automattic\WooCommerce\Admin\Notes\CustomizingProductCatalog;
 use \Automattic\WooCommerce\Admin\Notes\GivingFeedbackNotes;
 use \Automattic\WooCommerce\Admin\Notes\InsightFirstProductAndPayment;
 use \Automattic\WooCommerce\Admin\Notes\MobileApp;
@@ -140,6 +141,12 @@ class Events {
 		InsightFirstProductAndPayment::possibly_add_note();
 		AddFirstProduct::possibly_add_note();
 		AddingAndManangingProducts::possibly_add_note();
+		CustomizingProductCatalog::possibly_add_note();
+
+		if ( $this->is_remote_inbox_notifications_enabled() ) {
+			DataSourcePoller::read_specs_from_data_sources();
+			RemoteInboxNotificationsEngine::run();
+		}
 	}
 
 	/**
--- a/tests/notes/class-wc-tests-customizing-product-catalog.php
+++ b/tests/notes/class-wc-tests-customizing-product-catalog.php
@@ -1,0 +1,131 @@
+<?php
+/**
+ * CustomizingProductCatalog note tests
+ *
+ * @package WooCommerce\Admin\Tests\Notes
+ */
+
+use \Automattic\WooCommerce\Admin\Notes\CustomizingProductCatalog;
+use Automattic\WooCommerce\Admin\Notes\Note;
+
+
+/**
+ * Class WC_Tests_Marketing_Notes
+ */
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by adrianduffell — DISMISSED** [review]
Overall the code looks good 👍  Some tests for `get_note()` would be nice however.

I also found the note is triggered for established stores with existing products, but the advice in the note seems unwanted for those stores.

Perhaps there should be another check to determine if the store is new or established? I think you could use the store’s order count to do this. For example, if the store has existing orders, then don't display the note.

**Comment by moon0326** [comment]
> Overall the code looks good 👍 Some tests for `get_note()` would be nice however.
> 
> I also found the note is triggered for established stores with existing products, but the advice in the note seems unwanted for those stores.
> 
> Perhaps there should be another check to determine if the store is new or established? I think you could use the store’s order count to do this. For example, if the store has existing orders, then don't display the note.

Thank you for catching it 👍  I think that should be fixed. I will push a fix with a test.

I'm also going to see if this PR can use the Remote Inbox Notification system instead.

**Comment by moon0326** [comment]
@adrianduffell Hey Adrian, I asked Pedro about the existing store issue. This is his reply: https://github.com/woocommerce/woocommerce-admin/issues/5956#issuecomment-760085082

tl;dr -- the store is active for less than 14 days.

I've updated the logic and added tests.

File: src/Events.php [hunk 1]
```diff
@@ -140,6 +141,12 @@ protected function possibly_add_notes() {
 		InsightFirstProductAndPayment::possibly_add_note();
 		AddFirstProduct::possibly_add_note();
 		AddingAndManangingProducts::possibly_add_note();
+		CustomizingProductCatalog::possibly_add_note();
+
+		if ( $this->is_remote_inbox_notifications_enabled() ) {
+			DataSourcePoller::read_specs_from_data_sources();
+			RemoteInboxNotificationsEngine::run();
+		}
```
louwie17, line 149: I believe this isn't necessary, as it's already being done above where `possible_add_notes` is being triggered.

https://github.com/woocommerce/woocommerce-admin/pull/6032/files#diff-2e1f1cedde54c64bfd2d2d1b61019c13af57b9c5e1dc1315d4c2dcc492ad66e1R96-R99

tests/notes/class-wc-tests-customizing-product-catalog.php
```diff
@@ -0,0 +1,131 @@
+<?php
```
louwie17, line 1: Nice work on these test cases, very thorough 👍 👍

louwie17 (changes requested):
Tested the store being older then 14 day's and the note didn't show 👍 also nice work on the tests.
I left one comment in the `CustomizingProductCatalog.php`, I believe calling the `DataSourcePoller` there isn't necessary (see comment).

src/Events.php
moon0326: Thank you @louwie17 I'm not sure why that those lines are highlighted as new. The last four lines are existing code for the Remote Inbox Engine. I think we can ignore them.

tests/notes/class-wc-tests-customizing-product-catalog.php
moon0326: @louwie17 thank you!

src/Events.php
louwie17: @moon0326 are you sure? they don't show up on the `main` branch: https://github.com/woocommerce/woocommerce-admin/blob/main/src/Events.php#L142

louwie17: Thanks @moon0326, approving this :)

louwie17 (approved):
LGTM 🚀

moon0326:
> LGTM 🚀

Thank you! I need your approval please 😄

louwie17:
> Thank you! I need your approval please 😄

Ooh shoot haha you might need @adrianduffell approval, or you could dismiss his review, if you can (I haven't run into this before):
Screen Shot 2021-01-20 at 10 03 06 PM

moon0326:
> > Thank you! I need your approval please 😄
> 
> Ooh shoot haha you might need @adrianduffell approval, or you could dismiss his review, if you can (I haven't run into this before):
> Screen Shot 2021-01-20 at 10 03 06 PM

Oh nice! Did not know that was a feature :) Thank you!

PR #6040, opened 2021-01-11
Add remote inbox notifications rule that triggers when WCA is updated

### LINKED ISSUE #5957: New message: New! Manage your store activity from the Home screen
_Edit_ (@becdetat): implement a new remote inbox notifications rule to trigger on update (action `woocommerce_admin_updated`) and add the following note to WooCommerce.com's feed:

**Title:** New! Manage your store activity from the Home screen
**Copy:** Start your day knowing the next steps you need to take with your orders, products, and customer feedback. 

Read more about how to use the activity panels on the Home screen.
**Segmentation:** existing users
**Timing:** After updating WooCommerce to 4.9
**Cadence:** Once
**CTA leads to:** https://docs.woocommerce.com/document/home-screen/?utm_source=inbox
**CTA label:** Learn more

becdetat:
Fixes #5957

#5957 wants a message to be displayed after upgrading the WooCommerce Admin version (actually #5957 wants to trigger on the WooCommerce update, but these are practically the same).

This adds a hook to the WCA upgrade process (action hook `woocommerce_admin_updated`) which sets a flag in an option, goes through the Remote Inbox Notifications run, then clears the flag. It also adds a new `wca_updated` rule which just checks on that flag being set.

1. add the test URL to the `DATA_SOURCES` array in `DataSourcePoller.php`: `https://gist.githubusercontent.com/becdetat/4fda7403cd34f49b71bccd147419e34e/raw/7360ec1f0dad9ecc867c62a73dad4006635c2b84/wca-updated.json`
2. Run the `wc_admin_daily` cron  job, which will load in the new spec
3. Open the WooCommerce dashboard. The new note should _not_ appear.
4. Trigger a WCA upgrade by bumping the version option down in the database: `UPDATE wp_options SET option_value = '1.8.0' WHERE option_name = 'woocommerce_admin_version'`
5. Refresh the dashboard. The new note should now be present.


src/RemoteInboxNotifications/WooCommerceAdminUpdatedRuleProcessor.php
```diff
@@ -0,0 +1,36 @@
...
+	/**
+	 * Process the rule.
+	 *
+	 * @param object $rule         The specific rule being processed by this rule processor.
+	 * @param object $stored_state Stored state.
+	 *
+	 * @return bool Whether the rule passes or not.
+	 */
+	public function process( $rule, $stored_state ) {
+		return get_option( RemoteInboxNotificationsEngine::WCA_UPDATED_OPTION_NAME, false );
```
adrianduffell, line 23: Should there be a version number check here?

For example, if there is a note highlighting WooCommerce 5.0 features in the future, then I think it would need to be skipped in cases when the store is only updated to a lesser version like 4.9.1.

adrianduffell (commented):
Overall this looks good 👍 

I’ve just left a question about checking version numbers in the processor.

src/RemoteInboxNotifications/WooCommerceAdminUpdatedRuleProcessor.php
becdetat: Good point, I'll have a think about how to handle that

becdetat: @adrianduffell This isn't needed because there is another rule that checks the version of a plugin. So to target upgrading specifically to WooCommerce 5 you would use the `woocommerce_admin_updated` rule plus the `plugin_version` rule.

ilyasfoo:
@becdetat Since #5957 is fixed in https://github.com/woocommerce/woocommerce-admin/pull/6072, can we close this PR?

becdetat:
@ilyasfoo this is still valid, it adds a new rule type which, even though it isn't used not, it can be used in the future with new remote inbox notification specs.

src/RemoteInboxNotifications/WooCommerceAdminUpdatedRuleProcessor.php
ilyasfoo: @becdetat That makes sense, but would be better if we can add that detail in RemoteInboxNotification readme? It wasn't 100% apparent to me the first time I see it

becdetat: Thanks @ilyasfoo , added in 95797dc

ilyasfoo: Looks good!

ilyasfoo (approved):
Tests well! 🚢

PR #6047, opened 2021-01-12
Add an unread indicator to inbox messages based on existing logic.

### LINKED ISSUE #5627: Unread indication for Inbox messages
In the latest release (since we removed the last Read/Unread functionality), there's no indication of which Inbox notifications have been viewed. This is an update so that merchants will understand what's new in their Inbox. The indication could also be encouragement to open the message and see what it's all about.

A simple blue dot (directly to the left of the time stamp) upon first page load to indicate messages that have arrived since the user was last on the Home screen.

Home - Unread indication

Once the message has been actioned on or the user navigates away, the dot disappears.

All designs are in the Home screen Figma, let me know if this needs any clarification.

samueljseay:
Fixes #5627 by adding a "unread" indicator to unread messages based on existing logic (`unread = date created > activity_panel_inbox_last_read`)

Also note I haven't added tests, there is already test coverage of unread being correctly applied in an existing test.

### Screenshots

Screenshot 2021-01-12 at 7 54 47 PM

-  Trigger an inbox notification (the easiest way is via WP-Crontrol and running `wc_admin_daily` job on a fresh site. This usually shows at least a couple of new user inbox messages.

- Note that the message is initially shown with an unread symbol as in the linked screenshot.

- `activity_panel_inbox_last_read` is updated when the component remounts, so a page refresh should be enough to clear out the unread indicator for your messages. Note that the unread indicator no longer displays.

@elizaan36 Could you take a look at this and review the design? I couldn't find an actual design with measurements etc in Figma so I approximated it from the screenshot provided in the original issue.

### DIFF
--- a/client/inbox-panel/card.js
+++ b/client/inbox-panel/card.js
@@ -2,7 +2,7 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { Component, createRef, Fragment } from '@wordpress/element';
+import { Component, createRef } from '@wordpress/element';
 import { Button, Dropdown, Modal } from '@wordpress/components';
 import PropTypes from 'prop-types';
 import VisibilitySensor from 'react-visibility-sensor';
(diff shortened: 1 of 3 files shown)


client/inbox-panel/card.js
```diff
@@ -2,7 +2,7 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { Component, createRef, Fragment } from '@wordpress/element';
```
louwie17, line 5: Nice removal of `Fragment` :)

louwie17 (approved):
Tested well for me, and changes look good :)
Can be merged after the design approval.

elizaan36:
Looks perfect to me, and like how it's using the `wp-admin-theme-color` 💯

Issue #6023, opened 2021-01-07
[WC 4.9 RC2] `Setup additional payment providers` card doesn't work


juliaamosova:
**Describe the bug**

Happens on both Atomic and JN sites. Clicking on the button `Setup payments` in the `Setup additional payment providers` card reloads the page without redirecting the user to set up payments. 

**To Reproduce**

1. Create a brand new Atomic or JN site with WC 4.9 RC2 build. 
2. Complete onboarding with US address, the fashion industry, and installation of all the free business plugins. 
3. Complete all tasks on the Home screen and hide the task list. 
4. Try clicking on the button `Setup payments` in the `Setup additional payment providers` card.
5. Note that the page reloads without redirecting the user to set up payments. 

![http://g.recordit.co/XCY2Mz2TzW.gif](http://g.recordit.co/XCY2Mz2TzW.gif)

**Expected behavior**

I expected to be taken to the payments set up page (`WooCommerce/Settings/Payments`) when clicking on the button `Setup payments` in the `Setup additional payment providers` card.

**Screenshots**

See gif above. 

**Desktop (please complete the following information):**

* OS: macOS Catalina 10.15.6
* Browser: Chrome
* Version: 87.0.4280.88 (Official Build) (x86_64)

**Additional context**

N/A

juliaamosova:
JN site:

![](http://g.recordit.co/oOemlPovQe.gif)

moon0326:
So the issue is that we don't render the payment task component if `woocommerce_task_list_hidden` is set to `yes`. You can see the code [here](https://github.com/woocommerce/woocommerce-admin/blob/45c0ae9103753eb5c2b7de6ab42254267fd81b24/client/task-list/index.js#L130).

I think we have two options.

1. Display the task components regardless of `woocommerce_task_list_hidden` value. I think users should be able to perform the task even after hiding the task list.

2. Remove the note when the user hides the "Finish setup" panel.

cc @timmyc and @pmcpinto

timmyc:
Great sleuthing @moon0326 - I think we could consider a potential third option here, where if the payments task URL is accessed, but the payments task has been hidden, we then do a page redirect to the Settings > Payments page instead.

Pedro is probably offline until tomorrow at this point, so I think we should ready a PR for option #1 above though, and test it out a bit to ensure no other possible regressions can occur with this logic. The main reason I think we should still direct the to the payments task is it will allow the featured payment extensions for their locale to appear again.

[...]

PR #6026, opened 2021-01-07
Display the task components regardless of woocommerce_task_list_hidde…

moon0326:
Fixes #6023 

This PR disregards `woocommerce_task_list_hidden` value when a task component is requested by appending &task={task name}.

As [noted](https://github.com/woocommerce/woocommerce-admin/issues/6023#issuecomment-756365282) by Timmy in the original issue, this is probably going to be a temporary fix.

### Detailed test instructions:

1. Navigate to WooCommerce -> Home
2. Hide the "Finish setup" panel.
Screen Shot 2021-01-07 at 1 40 39 PMth_

3. Confirm you can still access `http://{Your WP URL}/wp-admin/admin.php?page=wc-admin&task=payments`

PR #6050, opened 2021-01-12
Fixed error hiding core task list

### LINKED ISSUE #6049: Error hiding core task list
**Describe the bug**
If the extended task list is visible, it is not possible to hide the core task list.

**To Reproduce**

1. Add [a task list item](https://github.com/woocommerce/woocommerce-admin/tree/main/docs/examples/extensions/add-task). You can do it by executing something like: `npm run example -- --ext=add-task` in your terminal
2. Go to the `Home` screen.
3. Hide the core task list.
Screen Capture on 2021-01-12 at 15-47-45
4. The task list will still be there.

**Expected behavior**
The list should disappear after pressing `hide`.

octaedro:
Fixes #6049

This PR fixes the error with hiding the core task list. After adding an extended task list, it's not possible to hide the core task list.

_Related also to [issue 6023](https://github.com/woocommerce/woocommerce-admin/issues/6023)_

### Screenshots
Screen Capture on 2021-01-12 at 15-56-42

1. Add [a task list item](https://github.com/woocommerce/woocommerce-admin/tree/main/docs/examples/extensions/add-task). You can do it by executing something like: `npm run example -- --ext=add-task` in your terminal
2. Go to the `Home` screen.
3. Hide the core task list.
4. The task list should disappear.

- Also, go to `/wp-admin/admin.php?page=wc-admin&task=payments` and verify the _setup additional payment providers_ card is still visible.
screenshot-one wordpress test-2021 01 12-16_05_51


client/task-list/test/index.js
```diff
@@ -135,6 +135,24 @@ describe( 'TaskDashboard and TaskList', () => {
 		).toBeNull();
 	} );
 
+	it( 'renders only the extended task list', () => {
```
louwie17, line 138: Could we also add a test for when the query object has `task` defined?

louwie17 (changes requested):
Looks great @octaedro, and thanks for keeping check with the original change that had broken this - https://github.com/woocommerce/woocommerce-admin/pull/6026 
In light of that, it be great if we could have one more test case for when `task` is defined in the query object prop.

joshuatf (commented):
Nice fix, @octaedro!  This PR is testing well; on a somewhat related note to this PR, I wanted to get your thoughts on https://github.com/woocommerce/woocommerce-admin/pull/5975#issuecomment-752989389

It looks like the original designs for the extended task list were in the original task list (which was previously extensible).  Did these designs change or did you run into trouble extending the original task list?

I ask because this introduces quite a few semantic changes around the task list and I can see more issues like this popping up going forward.

client/task-list/test/index.js
octaedro: Good idea! I added the test in the commit b62f9cf

octaedro:
Thank you @joshuatf for your review.

> This PR is testing well; on a somewhat related note to this PR, I wanted to get your thoughts on #5975 (comment)

My plan is to refactor a few things (swap the `do_action` and `add_action` and address some modifications suggested [here](https://github.com/woocommerce/woocommerce-admin/pull/5833#pullrequestreview-560540018)) in a follow-up of [this PR](https://github.com/woocommerce/woocommerce-admin/pull/5833).

> It looks like the original designs for the extended task list were in the original task list (which was previously extensible). Did these designs change or did you run into trouble extending the original task list?

Yes, the design was changed, now we're showing the core list and the extended list apart, for more context here is [the design](https://github.com/woocommerce/woocommerce-admin/issues/5242#issuecomment-733171105) and [the issue](https://github.com/woocommerce/woocommerce-admin/issues/5242).

> I ask because this introduces quite a few semantic changes around the task list and I can see more issues like this popping up going forward.

For sure if you have changes or suggestions please let me know and I'll be happy to address them 🙂

client/task-list/test/index.js
louwie17: Thanks @octaedro, looks great :)

louwie17 (approved):
Thanks for adding the extra test case, LGTM 🚀

joshuatf:
>Yes, the design was changed, now we're showing the core list and the extended list apart, for more context here is the design and the issue.

Gotcha.  Thanks for that context @octaedro.  It seems like the idea is to make sure extensions don't add items under the main task list, is that right?  I'm curious if this is an area that's being walled off intentionally or not (/cc @elizaan36).

If the idea is to intentionally wall off the list, extensions are still able to add tasks via the `woocommerce_admin_onboarding_task_list` filter and we also have an example plugin that shows how to do this.

>My plan is to refactor a few things (swap the do_action and add_action and address some modifications suggested here) in a follow-up of this PR.

Awesome!  A couple things that stood out to me were differences in the logic in a few places.  For example, in some places, this is the `taskListHidden` logic:

```js
const taskListHidden = getOption( 'woocommerce_task_list_hidden' ) === 'yes';
```

While in others it is:

```js
taskListHidden:
	getOption( 'woocommerce_task_list_hidden' ) === 'yes' &&
	getOption( 'woocommerce_extended_task_list_hidden' ) !== 'no',
```

It might be good to create a utils file that houses this logic so we have a single source of truth.

On the server we use the logic:

```php
	public static function should_show_tasks() {
		return 'no' === get_option( 'woocommerce_task_list_hidden', 'no' ) || 'no' === get_option( 'woocommerce_extended_task_list_hidden', 'no' );
	}
```

What's also a bit confusing about all this naming is the semantics of this.  For example, maybe these should be renamed to `taskListsHidden` (note the extra `s`) and rename any variables to differentiate the original task list from the extensions one.

Thanks again for all your work on this!

elizaan36:
>  It seems like the idea is to make sure extensions don't add items under the main task list, is that right? I'm curious if this is an area that's being walled off intentionally or not (/cc @elizaan36).

The main problem to solve is that extensions will add tasks after core setup is complete. So we'd run into the problem of having a completed and dismissed setup task list, then an extension adds a task, and the task list would have to reappear despite the user choice to hide it.

So it's really about making sure extension tasks can operate independently of the setup task list, as they'll be associated with different points in the setup/store management timeline.

> If the idea is to intentionally wall off the list, extensions are still able to add tasks via the woocommerce_admin_onboarding_task_list filter and we also have an example plugin that shows how to do this.

I think it could be problematic if we still allow extensions to add tasks to the store setup task list for the reason above. One solution could be, and this may get super in the weeds, but we could classify the extensions installed in the onboarding wizard differently so they have the ability to show up in the setup task list.

octaedro:
Thank you @joshuatf for your suggestions.
I created the [PR 6081](https://github.com/woocommerce/woocommerce-admin/pull/6081) addressing them. I also added a few modifications that Jeff [proposed here](https://github.com/woocommerce/woocommerce-admin/pull/5833#pullrequestreview-560540018).

PR #6054, opened 2021-01-13
Restrict beta features to only those that have opted into tracking, adding modal for navigation feature

### LINKED ISSUE #5930: Navigation [1.9] Add usage tracking message
**Description:** as requested by Warren here (pbIJXs-vY-p2#comment-1439), users should only be able to enable new features that are being progressively rolled out if they opted into usage sharing.

**Acceptance criteria:**
- [ ] If the user has usage tracking enabled or if he already has the feature enabled, show the option to enable/disable a new feature as we have now.
- [ ] If the user doesn't have usage tracking enabled, display the following content between the features description and the navigation row:
Gathering usage data allows us to make WooCommerce better — your store will be considered as we evaluate new features, judge the quality of an update, or determine if an improvement makes sense. Read about what usage data is tracked `[here](https://woocommerce.com/usage-tracking)`. Enable usage tracking to test new features. Enable usage tracking (CTA - automatically enables tracking when the user clicks on it)
- [ ] The user can only toggle the navigation checkbox after enabling usage tracking
- [ ] This requirement should only happen when a feature is disabled by default for everyone. Eg: Once the nav is enabled by default for new users we drop this requirement.

joelclimbsthings:
Fixes #5930

This PR implements the flow to be triggered when a user attempts to opt into the new navigation feature, detailed in the originating issue. These beta features will be restricted exclusively to those opted into tracking. 

_Possible Concerns_
- The modal is fairly tightly coupled to the navigation feature checkbox at the moment. I'm not sure of the likelihood or timeline if when we may want to use the modal elsewhere, but it will require some minor changes. I'm not sure if that's something we want to tackle in this PR, or leave it until the issue arises.
- I'm using a filter to conditionally load the modal script in `Loader.php` right now. It would be best to factor some of this logic into another file at some point (as mentioned by @joshuatf ) since this file has become a catch-all.
- I'm also using a `filter` instead of an `action` simply because this filter only triggers on the _Settings_ screen, and I wasn't aware of a comparable `action`. Just let me know if one exists.
- There is an async operation to change the option when hitting the _Save_ button, but no loading/etc indicator. I didn't see anything like that used elsewhere, but I'm glad to add it if it's suggested. 

### Screenshots

image

-   Check out branch
-   Ensure tracking is disabled, either removing `woocommerce_allow_tracking` option or settings to "no," and the new navigation is disabled.
-   Navigate to `Settings -> Advanced -> Features` and click the checkbox to enable the new navigation.
- The modal should appear.
- If you check the "Enable usage tracking" checkbox and click "Save," the option `woocommerce_allow_tracking` should change to "yes" and the nav feature checkbox should be checked.
- If you exit in any other way, the nav feature checkbox should remain unchecked.

### DIFF
--- a/client/wp-admin-scripts/beta-features-tracking-modal/container.js
+++ b/client/wp-admin-scripts/beta-features-tracking-modal/container.js
@@ -2,12 +2,29 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { useState } from '@wordpress/element';
+import { useState, useEffect } from '@wordpress/element';
 import { Button, Modal, CheckboxControl } from '@wordpress/components';
 
 export const BetaFeaturesTrackingContainer = () => {
-	const [ isModalOpen, setIsModalOpen ] = useState( true );
+	const [ isModalOpen, setIsModalOpen ] = useState( false );
 	const [ isChecked, setIsChecked ] = useState( false );
+
+	useEffect( () => {
+		const enableNavigationCheckbox = document.querySelector(
+			'#woocommerce_navigation_enabled'
+		);
+
+		enableNavigationCheckbox.addEventListener(
+			'change',
+			( e ) => {
+				if ( e.target.checked ) {
+					e.target.checked = false;
+					setIsModalOpen( true );
+				}
+			},
+			false
+		);
+	}, [] );
 
 	if ( ! isModalOpen ) {
 		return null;
@@ -40,10 +57,7 @@ export const BetaFeaturesTrackingContainer = () => {
 				/>
 			</div>
 			<div className="woocommerce-beta-features-tracking-modal__actions">
-				<Button
-					isPrimary
-					onClick={ () => this.setState( { isModalOpen: false } ) }
-				>
+				<Button isPrimary onClick={ () => setIsModalOpen( false ) }>
 					{ __( 'Save', 'woocommerce-admin' ) }
 				</Button>
 			</div>
--- a/client/wp-admin-scripts/beta-features-tracking-modal/container.js
+++ b/client/wp-admin-scripts/beta-features-tracking-modal/container.js
@@ -1,0 +1,52 @@
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { useState } from '@wordpress/element';
+import { Button, Modal, CheckboxControl } from '@wordpress/components';
+
+export const BetaFeaturesTrackingContainer = () => {
+	const [ isModalOpen, setIsModalOpen ] = useState( true );
+	const [ isChecked, setIsChecked ] = useState( false );
+
+	if ( ! isModalOpen ) {
+		return null;
+	}
+
+	return (
+		<Modal
+			title={ __( 'Build a Better WooCommerce', 'woocommerce-admin' ) }
+			onRequestClose={ () => setIsModalOpen( false ) }
+			className="woocommerce-beta-features-tracking-modal"
+		>
+			<p>
+				{ __(
+					'Testing new features requires sharing non-sensitive data via ',
+					'woocommerce-admin'
+				) }
+				<a href="https://woocommerce.com/usage-tracking">
+					{ __( 'usage tracking', 'woocommerce-admin' ) }
+				</a>
+				{ __(
+					'. Gathering usage data allows us to make WooCommerce better — your store will be considered as we evaluate new features, judge the quality of an update, or determine if an improvement makes sense. No personal data is tracked or stored and you can opt-out at any time.',
+					'woocommerce-admin'
+				) }
+			</p>
+			<div className="woocommerce-beta-features-tracking-modal__checkbox">
+				<CheckboxControl
+					label="Enable usage tracking"
+					onChange={ setIsChecked }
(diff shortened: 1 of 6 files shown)


psealock (approved):
Awesome Joel, this is testing well without errors on JN site.

One thing I noticed is that once you have Navigation on, you can turn off tracking and still have Navigation. Is this a reasonable workaround or something we'd like to limit? cc @pmcpinto 

Otherwise the code looks good.

>The modal is fairly tightly coupled to the navigation feature checkbox at the moment.

This is a shortlived script that will be removed fairly soon. Unless this is a pattern we're going to be using a lot, I don't think spending the energy to make an elegant solution worth it. Perhaps that will change in the future.

>There is an async operation to change the option when hitting the Save button, but no loading/etc indicator. I didn't anything like that used elsewhere, but I'm glad to add it if it's suggested.

A spinner would be a nice touch, but my testing was fairly quick and I didn't notice a significant lag so I think its fine as is.

pmcpinto:
> One thing I noticed is that once you have Navigation on, you can turn off tracking and still have Navigation. Is this a reasonable workaround or something we'd like to limit? cc @pmcpinto

I think that the navigation should also be disabled. Imagine that the user disables usage tracking but continues seeing the navigation enabled, he'll probably wonder if the tracking was actually disabled.  

@joelclimbsthings is it possible to trigger an event if the user decides to enable tracking in this modal? It's going to be useful to measure the reach of this modal.

We're doing something similar in the first step of the store profiler with the following events: wcadmin_storeprofiler_store_details_continue and wcadmin_storeprofiler_store_details_skip

I can open a new issue just for the event tracking if necessary.

client/wp-admin-scripts/beta-features-tracking-modal/container.js
```diff
@@ -0,0 +1,109 @@
...
+import { useState, useEffect, useRef } from '@wordpress/element';
+import { Button, Modal, CheckboxControl } from '@wordpress/components';
+import { withDispatch } from '@wordpress/data';
+import { compose } from '@wordpress/compose';
+import { OPTIONS_STORE_NAME } from '@woocommerce/data';
+
+const BetaFeaturesTrackingModal = ( { updateOptions } ) => {
+	const [ isModalOpen, setIsModalOpen ] = useState( false );
+	const [ isChecked, setIsChecked ] = useState( false );
+	const enableNavigationCheckbox = useRef(
```
joshuatf, line 14: Saw your comment about this in the PR description and agree that it might be useful to target all checkboxes in this section, but this can wait for a future PR. 👍

src/Loader.php
```diff
@@ -260,6 +261,39 @@ public static function add_features_section( $sections ) {
...
+	/**
+	 * Conditionally loads the beta features tracking modal.
+	 *
+	 * @param array $settings Settings.
+	 * @return array
+	 */
+	public static function maybe_load_beta_features_modal( $settings ) {
+		$tracking_enabled = get_option( 'woocommerce_allow_tracking', 'no' );
+
+		if ( 'yes' === $tracking_enabled ) {
```
joshuatf, line 273: 🎉 

Just making a note for the future: we may need to handle this check client-side if Settings pages get reactified since tracking can be toggled without page refresh.

joshuatf (approved):
Awesome work here, Joel 💯.  Left a couple comments for the future, but no changes needed and this is testing really well.

>One thing I noticed is that once you have Navigation on, you can turn off tracking and still have Navigation. Is this a reasonable workaround or something we'd like to limit? cc @pmcpinto

Seems like something worth tackling.  I think we need a better way of getting all feature toggles (not just navigation) and automatically turn them off on update of the tracking option.

I'll open an issue to follow up on this to keep the scope on this PR down.

>@joelclimbsthings is it possible to trigger an event if the user decides to enable tracking in this modal? It's going to be useful to measure the reach of this modal.

Good call @pmcpinto.  @joelclimbsthings up to you if you want to tackle here or a follow-up.  I'm keen on getting this PR merged so we can continue to iterate.

src/Loader.php
joelclimbsthings: Yup, good call.

joelclimbsthings:
> We're doing something similar in the first step of the store profiler with the following events: wcadmin_storeprofiler_store_details_continue and wcadmin_storeprofiler_store_details_skip

I think this would be pretty easy to throw in. What would you like to name the event, and does it need to contain any payload @pmcpinto ?

> I'll open an issue to follow up on this to keep the scope on this PR down.

Thanks @joshuatf , sounds good!

pmcpinto:
> What would you like to name the event, and does it need to contain any payload @pmcpinto ?

It doesn't need to contain any payload. In terms of the name, it can be something like wcadmin_settings_features_tracking_enabled

Thanks!

ChromeOrange:
So if I want to test the new navigation feature I have to share tracking information?

Surely most sites that are testing things like this are staging sites?

In my case I'm using a dev site to test this, any data you get from me will be worthless and may even skew the data you are trying to collect. Every single order on my site is fake, there are no live orders, no real products and no real customers.

pmcpinto:
Hey @ChromeOrange thanks for your feedback We're reassessing this. Stay tuned!

Issue #6051, opened 2021-01-12
Append tags to new nav Crowdsignal survey


elizaan36:
Currently, we’re receiving feedback in [a Crowdsignal survey](http://automattic.survey.fm/new-navigation) from core users that enabled the nav through wcadmin 1.8.2 and in the upcoming weeks we’re also going to get feedback from eCommerce and Business plan users. However, we don’t have a way to segment their feedback.

Specifically, we're wondering if it’s possible to include a parameter indicating if the user is coming from either eCom, Business plan, or other place.

I reached out to the Crowdsignal folks and it would only be possible if we can [append the tags in the URL](https://crowdsignal.com/support/using-custom-tags-to-track-survey-responses/).

This issue is more of a exploration into whether or not we can construct the survey link based on the user account so the results contain the tags we need. If it works we'd lean towards something like that to identify people, rather than having multiple surveys or seeding the survey data with an additional question.

cc @pmcpinto @psealock

pmcpinto:
@joshuatf do you know when this will be available? So we can start identifying if users are coming from the eCommerce plan, Business plan or Core

joshuatf:
@pmcpinto This should be available in 1.9.0, we just need to use the `woocommerce_admin_survey_query` filter in the bridge to add any necessary data.

I'll open an issue there.

joshuatf:
Issue created here - https://github.com/Automattic/wc-calypso-bridge/issues/636

PR #6066, opened 2021-01-13
Add survey URL builder method

joshuatf:
Fixes (partially) #6051

Creates a method to build survey URLs so we can later filter the query and add in `source` to identify which plan users are coming from.

The `Survey` class may be slightly overkill, but didn't feel right tossing this into `core-functions.php`.

### Screenshots

Screen Shot 2021-01-13 at 1 16 55 PM
Screen Shot 2021-01-13 at 1 16 45 PM

1. Optionally add some arguments to the survey url.
```php
add_filter( 'woocommerce_admin_survey_query', function( $query ) {
	$query['source'] = 'ecomm-plan';
	return $query;
} );
```
2. Delete old navigation feedback and store setup notes.
2. Re-trigger those notes.
3. Make sure the URL for those items is correct.
4. Opt in to the navigation if not already opted in.
5. Opt out of the navigation and note the "Share Feedback" button URL in the modal is correct.

### DIFF
--- a/src/Survey.php
+++ b/src/Survey.php
@@ -1,0 +1,45 @@
+<?php
+/**
+ * Survey helper methods.
+ */
+
+namespace Automattic\WooCommerce\Admin;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Survey Class.
+ */
+class Survey {
+	/**
+	 * Get a survey's URL from a path.
+	 *
+	 * @param  string $path Path of the survey.
+	 * @param  array  $query Query arguments as key value pairs.
+	 * @return string Full URL to survey.
+	 */
+	public static function get_url( $path, $query = array() ) {
+		$url = 'https://automattic.survey.fm' . $path;
+
+		$source = self::get_source();
+		if ( $source ) {
+			$query['source'] = $source;
+		}
+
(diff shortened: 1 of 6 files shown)


src/Survey.php
```diff
@@ -0,0 +1,33 @@
...
+class Survey {
+	/**
+	 * Get a survey's URL from a path.
+	 *
+	 * @param  string $path Path of the survey.
+	 * @param  array  $query Query arguments as key value pairs.
+	 * @return string Full URL to survey.
+	 */
+	public static function get_url( $path, $query = array() ) {
+		$url = 'https://automattic.survey.fm' . $path;
```
louwie17, line 22: Would it be worth to making the `automattic.survey.fm` url a class constant:
`const SURVEY_URL = 'https://automattic.survey.fm'`

louwie17 (approved):
This looks good, and tested well.
I left one minor cosmetic suggestion, aside from that you got my approval 🚀

src/Survey.php
joshuatf: Seems like a reasonable idea.  Added in https://github.com/woocommerce/woocommerce-admin/pull/6066/commits/9b97321da6fe649e9c4909d5d7614f1b9ea551e6

PR #6070, opened 2021-01-13
Changelog update

psealock:
Updates Changelog for 1.9. 

I will leave this on hold until we ship the 1.9 final in case of any bugs that creep in.


becdetat (approved):
:shipit:

PR #6072, opened 2021-01-14
New note: manage activity from home screen

### LINKED ISSUE #5957: New message: New! Manage your store activity from the Home screen

becdetat:
Fixes #5957

This adds a new note that only triggers when updating to WooCommerce Admin 1.9.0.

### Screenshots
image

1. Go to the WooCommerce home screen. The notification should not be present.
2. `UPDATE wp_options SET option_value = '1.8.0' WHERE option_name = 'woocommerce_admin_version'`
3. In `ManageStoreActivityFromHomeScreen.php` set the version that is checked (line 41) to `1.9.0-dev` (or whatever the version you are using will update to).
4. Refresh home screen - this will trigger the update. Note should now be present.

### DIFF
--- a/src/Install.php
+++ b/src/Install.php
@@ -137,7 +137,25 @@ class Install {
 		 */
 		if ( ! $version_option || $requires_update ) {
 			self::install();
+			/**
+			 * WooCommerce Admin has been installed or updated.
+			 */
 			do_action( 'woocommerce_admin_updated' );
+
+			if ( ! $version_option ) {
+				/**
+				 * WooCommerce Admin has been installed.
+				 */
+				do_action( 'woocommerce_admin_installed' );
+			}
+
+			if ( $requires_update ) {
+				/**
+				 * An existing installation of WooCommerce Admin has been
+				 * updated.
+				 */
+				do_action( 'woocommerce_admin_updated_existing' );
+			}
 		}
 
 		/*
--- a/src/Install.php
+++ b/src/Install.php
@@ -146,7 +146,7 @@ class Install {
 				/**
 				 * WooCommerce Admin has been installed.
 				 */
-				do_action( 'woocommerce_admin_installed' );
+				do_action( 'woocommerce_admin_newly_installed' );
 			}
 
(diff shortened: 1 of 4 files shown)


src/Install.php
```diff
@@ -137,7 +137,25 @@ public static function check_version() {
 		 */
 		if ( ! $version_option || $requires_update ) {
 			self::install();
+			/**
+			 * WooCommerce Admin has been installed or updated.
+			 */
 			do_action( 'woocommerce_admin_updated' );
```
becdetat, line 143: The `woocommerce_admin_updated` action gets triggered on install or update, so I had to add some more specific actions below - `woocommerce_admin_updated_existing` is what is used in this particular note.

psealock (commented):
This is working well and code looks good.

So these two actions are new, yeah. What do you think about `woocommerce_admin_new_install` instead of `woocommerce_admin_installed`? This would be a fresh install right?

becdetat:
@psealock 

> So these two actions are new, yeah

Yep

> What do you think about `woocommerce_admin_new_install` instead of `woocommerce_admin_installed`? This would be a fresh install right?

It would be a fresh install.

I don't _love_ `woocommerce_admin_new_install`, I prefer a past-tense noun for action names. How about `woocommerce_admin_newly_installed`?

psealock (approved):
`woocommerce_admin_newly_installed` sounds great. And 👍🏽 to keeping the tenses sane

pre-approving

PR #6075, opened 2021-01-14
Disable features on tracking opt out

### LINKED ISSUE #6064: Turn off beta features on tracking opt out
As discussed in https://github.com/woocommerce/woocommerce-admin/pull/6054 we need to consider toggling off beta features when opting out of tracking.

We can use the `update_option_woocommerce_allow_tracking` hook to watch and update feature flags.

This might be a nice opportunity to encapsulate features in a way that allows us to more easily detect and toggle feature flags.  E.g., `Feature::get_enabled_toggleable_features()` and `Feature::disable( $feature_slug )`.  Note that the naming between the feature toggle flag (the wp_option) and an "enabled" feature (dev/core/plugin flag) can be confusing.

joshuatf:
Fixes #6064 

* Refactors features and relevant methods to their own file.
* Disables toggleable features on opt out.

1. Smoke test features and make sure nothing is broken.
2. Opt in to tracking and the navigation feature.
3. Opt out of tracking.
4. Note the navigation feature is turned off.

### DIFF
--- a/src/Features/Features.php
+++ b/src/Features/Features.php
@@ -45,6 +45,7 @@ class Features {
 		add_filter( 'woocommerce_get_settings_advanced', array( __CLASS__, 'maybe_load_beta_features_modal' ), 10, 2 );
 		add_action( 'admin_enqueue_scripts', array( __CLASS__, 'load_scripts' ), 15 );
 		add_filter( 'admin_body_class', array( __CLASS__, 'add_admin_body_classes' ) );
+		add_filter( 'update_option_woocommerce_allow_tracking', array( __CLASS__, 'maybe_disable_features' ), 10, 2 );
 	}
 
 	/**
@@ -68,22 +69,71 @@ class Features {
 	}
 
 	/**
+	 * Get the feature class as a string.
+	 *
+	 * @param string $feature Feature name.
+	 * @return string|null
+	 */
+	public static function get_feature_class( $feature ) {
+		$feature       = str_replace( '-', '', ucwords( strtolower( $feature ), '-' ) );
+		$feature_class = 'Automattic\\WooCommerce\\Admin\\Features\\' . $feature;
+
+		if ( class_exists( $feature_class ) ) {
+			return $feature_class;
+		}
+
+		// Handle features contained in subdirectory.
+		if ( class_exists( $feature_class . '\\Init' ) ) {
+			return $feature_class . '\\Init';
+		}
+
+		return null;
+	}
+
+	/**
 	 * Class loader for enabled WooCommerce Admin features/sections.
 	 */
 	public static function load_features() {
 		$features = self::get_features();
 		foreach ( $features as $feature ) {
-			$feature       = str_replace( '-', '', ucwords( strtolower( $feature ), '-' ) );
-			$feature_class = 'Automattic\\WooCommerce\\Admin\\Features\\' . $feature;
+			$feature_class = self::get_feature_class( $feature );
 
-			// Handle features contained in subdirectory.
-			if ( ! class_exists( $feature_class ) && class_exists( $feature_class . '\\Init' ) ) {
-				$feature_class = $feature_class . '\\Init';
-			}
-
-			if ( class_exists( $feature_class ) ) {
+			if ( $feature_class ) {
 				new $feature_class();
 			}
+		}
+	}
+
+	/**
+	 * Disable a toggleable feature.
+	 *
+	 * @param string $feature Feature name.
+	 * @return bool
+	 */
+	public static function disable( $feature ) {
+		$feature_class = self::get_feature_class( $feature );
+
+		if ( $feature_class && defined( "$feature_class::TOGGLE_OPTION_NAME" ) ) {
+			update_option( constant( "$feature_class::TOGGLE_OPTION_NAME" ), 'no' );
+			return true;
+		}
+
+		return false;
+	}
+
+	/**
+	 * Disable features when opting out of tracking.
+	 *
+	 * @param string $old_value Old value.
+	 * @param string $value New value.
+	 */
+	public static function maybe_disable_features( $old_value, $value ) {
+		if ( 'yes' === $value ) {
+			return;
+		}
+
+		foreach ( self::get_features() as $feature ) {
+			self::disable( $feature );
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


src/Features/Navigation/Init.php
```diff
@@ -16,14 +16,19 @@
  * Contains logic for the Navigation
  */
 class Init {
+	/**
+	 * Option name used to toggle this feature.
+	 */
+	const TOGGLE_OPTION_NAME = 'woocommerce_navigation_enabled';
```
joshuatf, line 22: This is a tiny bit magic.  An alternative would be to store the toggleable features in the `Features` class as an associative array.

```php
$toggleable_features = array(
  'navigation' => 'woocommerce_navigation_enabled'
);
```

Though I do like the encapsulation using the current method.

src/Loader.php
```diff
@@ -7,9 +7,9 @@
 namespace Automattic\WooCommerce\Admin;
 
 use \_WP_Dependency;
-use Automattic\WooCommerce\Admin\Features\Onboarding;
```
joelclimbsthings, line 10: Looks like this was just obsolete already?

src/Features/Features.php
```diff
@@ -0,0 +1,268 @@
...
+	/**
+	 * Disable a toggleable feature.
+	 *
+	 * @param string $feature Feature name.
+	 * @return bool
+	 */
+	public static function disable( $feature ) {
+		$feature_class = self::get_feature_class( $feature );
+
+		if ( $feature_class && defined( "$feature_class::TOGGLE_OPTION_NAME" ) ) {
```
joelclimbsthings, line 110: So, as it is, any feature class with this `TOGGLE_OPTION_NAME` defined will be disabled with tracking?

src/Features/Navigation/Init.php
joelclimbsthings: I'd lean a bit towards the associative array myself, just for clarity. As it is I wouldn't realize that refactoring/renaming `TOGGLE_OPTION_NAME` within a feature class would disrupt any functionality. 

Alternatively, we could rename that constant to something a little more descriptive, or even add a separate boolean flag that actually determines if the feature is disabled with tracking. ie:
```
const DISABLE_WITH_TRACKING = True;
const TOGGLE_OPTION_NAME = 'woocommerce_navigation_enabled';
```
That way the option name variable would just serve to indicate the option name, and the boolean would fairly clearly indicate the relationship with tracking.

joelclimbsthings (commented):
Looks great @joshuatf , definitely love the refactor into `Features.php`! 

Smoke tested the features, and tested the tracking toggle, and no issues found. I added a couple questions on implementation. Interested in your thoughts on those.

src/Features/Navigation/Init.php
joshuatf: Thanks for the feedback here.  I'll update this to use the associative array; I think it does make things more clear.

src/Loader.php
joshuatf: It was, I think removed a while ago.

src/Features/Features.php
joshuatf: That's right.  We'd want to opt out of any toggleable features (presumably all toggleable features are beta features) so this helps with other features going forward.

We may want to rename to `beta_features` as opposed to `toggleable_features` so it's clear that these will be turned off when tracking is opted out.

joshuatf:
@joelclimbsthings Updated this logic in https://github.com/woocommerce/woocommerce-admin/pull/6075/commits/538bbc3ec74349f0958dbe8fd3bfa0ae4e33e4e3

One downside we'll need to be careful of with this method: we need to manually check for the class before adding to the options array as I did in that commit.  It may be easy to slip up and forget that the feature class may or may not be loaded depending on the build.

joshuatf:
Posting here as a reminder to myself:

`is_enabled()` may be better named one of the following to avoid the ambiguity behind being enabled and the feature file being loaded:

* `is_loaded()` - Might not be great since it implies the feature file is already loaded.
* `exists()`
* `has_feature()`
* ...

joelclimbsthings:
Thanks @joshuatf , all looks good and still functions as expected. The only other note I'd add is that it might be a surprise for users that opt out of tracking that the navigation suddenly reverts if they'd forgotten the requirement, or were not the ones to enable navigation in the first place. 

That said, perhaps this is all moot due to the [recent revelation](pbIJXs-wC-p2#comment-1578) that it may be illegal. I'll hold back on approving until we bottom out on that.

joshuatf:
On hold until we determine if we're going to remove the opt-in requirement for beta features.  If we remove that requirement, we'll keep the refactor here less the tracking changes.

joelclimbsthings:
It sounds like [we have a go-ahead](pbIJXs-wC-p2#comment-1647) with this. :ship:

joelclimbsthings (approved):

joshuatf:
Updated the method names to be more semantic in https://github.com/woocommerce/woocommerce-admin/pull/6075/commits/d291a2bc77de1606f0b58a6ad54f48f9dc27c067 - can you give this a glance @joelclimbsthings or @psealock ?

src/Features/Features.php
```diff
@@ -0,0 +1,304 @@
...
+		}
+
+		if ( ! Loader::user_can_analytics() ) {
+			return;
+		}
+
+		$features         = self::get_features();
+		$enabled_features = array();
+		foreach ( $features as $key ) {
+			$enabled_features[ $key ] = self::exists( $key );
```
psealock, line 278: Should this not use `self:: is_enabled( $key )`?

The previous implementation relied on the logic of the option as well

joshuatf: Good call @psealock.  I think this was not the way the client-side feature flags worked as we still had flag checks on the front-end, but this actually simplifies the logic quite a bit and can't see a reason not to do this right now.

Added this in https://github.com/woocommerce/woocommerce-admin/pull/6075/commits/6cddc360298d5722cf369843e0945e94961f056e and ran through the app to make sure things weren't broken.  Looking good on my end, but a quick sanity check would be nice.

joelclimbsthings:
Just updated as per Paul's requested changes and gave it a quick run through again. Still works. I stand by my approval. :standing_man:

PR #6077, opened 2021-01-15
Allow for larger result sets of attribute terms

### LINKED ISSUE #5868: Product Attributes with <3 characters (like XL) not populating in Analytics Filters
Please provide us with the information requested in this bug report. Without these details, we won't be able to fully evaluate this issue. 
Bug reports lacking detail, or for any other reason than to report a bug, may be closed without action.

**Prerequisites (mark completed items with an [x]):**
- [X] I have have carried out troubleshooting steps and I believe I have found a bug.
- [X] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

**Describe the bug**
One of our staff was trying to perform a filtered search for XL t-shirts in Analytics.  She found that the XL size was not populating, while all of the other sizes were populating (Small, Medium, Large, XXL, XXXL).  I was able to confirm this with her.

I was trying to demonstrate other methods of producing those results for her, so she could carry on with her work.  During that process, I said, "you'll have to type a minimum of 3 characters into the search field to yield results."

On that hunch, i tried renamed the XL attribute to Extra Large. VOILA!

**Expected behavior**
When searching for product attributes, the results should most likely start populating after 1 character is entered.  I understand and realize why you wouldn't want this result on the front end.  But, specifically on attributes, we have several of them that are 1 or 2 characters.

**Actual behavior**
XL did not appear in the Filter
Analytics > Orders > Filter > Attributes > Size > XL

**Steps to reproduce the bug (We need to be able to reproduce the bug in order to fix it.)**
Search for a 2 character attribute in Analytics

**Screenshots**
XL
Attributes-XL
Analytics-XL

Extra Large
Analytics-ExtraLarge
Attributes-ExtraLarge

samueljseay:
At least in part this fixes #5868 

The issue is that attributes and attribute terms work quite differently in the search at the moment.

Attributes will not show anything until you start typing, at which point it does a `search` request with what you've typed and limits it to 10 results. This makes sense and works.

Attribute terms however does not do this, first it requests all the terms for the attribute selected, *but* without a per page limit this only returns 10 results, causing the confusion we saw in #5868. However my solution of a per_page limit of `-1` will still limit results to 100. 

A more extensive solution to this (in the case where an attribute has more than 100 terms), would be to ensure that they're all fetched up front and then text search the entire result set, or if possible do a search on the entered text like we do for attributes.

I'm suggesting this as an interim fix that at least makes it more usable for a majority of users.

1. Create a product attribute
2. Give the attribute more than 10 terms
3. Go to Analytics -> Orders
4. Add an attribute filter to the list, choose your attribute.
5. Go to the second input field and click on it, a result set of all your terms should appear

### DIFF
--- a/packages/components/src/advanced-filters/attribute-filter.js
+++ b/packages/components/src/advanced-filters/attribute-filter.js
@@ -91,7 +91,7 @@ const AttributeFilter = ( props ) => {
 			value[ 0 ]
 		) {
 			apiFetch( {
-				path: `/wc-analytics/products/attributes/${ value[ 0 ] }`,
+				path: `/wc-analytics/products/attributes/${ value[ 0 ] }?per_page=-1`,
 			} )
 				.then( ( { id, name } ) => [
 					{
@@ -112,7 +112,7 @@ const AttributeFilter = ( props ) => {
 		}
 		setAttributeTerms( false );
 		apiFetch( {
-			path: `/wc-analytics/products/attributes/${ selectedAttribute[ 0 ].key }/terms`,
+			path: `/wc-analytics/products/attributes/${ selectedAttribute[ 0 ].key }/terms?per_page=-1`,
 		} )
 			.then( ( terms ) =>
 				terms.map( ( { id, name } ) => ( {
--- a/packages/components/src/advanced-filters/attribute-filter.js
+++ b/packages/components/src/advanced-filters/attribute-filter.js
@@ -91,7 +91,7 @@ const AttributeFilter = ( props ) => {
 			value[ 0 ]
 		) {
 			apiFetch( {
-				path: `/wc-analytics/products/attributes/${ value[ 0 ] }?per_page=-1`,
+				path: `/wc-analytics/products/attributes/${ value[ 0 ] }`,
 			} )
 				.then( ( { id, name } ) => [
 					{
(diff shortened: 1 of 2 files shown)


ilyasfoo (approved):
I think this would be a good interim fix since 10 is a pretty low limit and could affect a lot of users IMO. Thanks for the quick work Sam, LGTM 🚢

packages/components/src/advanced-filters/attribute-filter.js
```diff
@@ -112,7 +112,7 @@ const AttributeFilter = ( props ) => {
 		}
 		setAttributeTerms( false );
 		apiFetch( {
-			path: `/wc-analytics/products/attributes/${ selectedAttribute[ 0 ].key }/terms`,
+			path: `/wc-analytics/products/attributes/${ selectedAttribute[ 0 ].key }/terms?per_page=-1`,
```
ilyasfoo, line 115:
```suggestion
			path: `/wc-analytics/products/attributes/${ selectedAttribute[ 0 ].key }/terms?per_page=100`,
```

Though, would it be better if we make it `100` so it would look more deliberate? `-1` does seem a bit misleading about the real limit.

ilyasfoo (approved):

packages/components/src/advanced-filters/attribute-filter.js
samueljseay: Sure thing :)

splaquet:
Thank you so much for revisiting this, @samueljseay !!!

PR #6082, opened 2021-01-17
Add a stale check to PRs and issues.

samueljseay:
For a long time now we have been dealing with a burgeoning and growing list of issues that often go stale. We also have allowed a pattern where PRs can be raised and hang around a long time.

The reason these things are problems is that it introduces extra cognitive burden and time waste (large in my opinion). To anyone trying to review PRs and groom issues in our code base. It is also hard to feel like you can politely close an issue or PR that you believe has become stale.

The stale action is scheduled to do a check at the end of each day over our issues and PRs looking for some that have become stale. It is very conservative in its check. It is just a bot so its a great level of indirection where you don't have to be the bearer of bad news.

Here's a summary of how I've configured it and why:

1. PRs receive a stale notification after 30 days, if no further activity (commits, comments etc) happens it is closed 3 days later. This is to discourage people adding long running stale WIP PRs or just leaving their PR hanging after we've asked for feedback or work on it. Bear in mind that it will not be closed if they do *anything* within the 3 day limit.

2. Issues are given a lot more grace. An issue must be stale for 90 days before a close warning is given. The goal here is to cut back a lot of stale issues that we can't even look at right now because we have too many. There are many issues where we haven't been able to reproduce or haven't received updated feedback from the raiser. **I would like to revisit this limit at some point, because it'd be nice to maybe cut this to 60 days if possible.**

3. I've exempted any **issue** that is tagged with cooldown, priority high, priority critical or any of the feature labels we currently use to categorise work. I'd like to maybe remove these feature labels from the list at some point, and maybe come up with a label we can use when we are creating an issue for something that will go into zenhub (something like "sprint backlog" or something).


louwie17 (commented):
> Bear in mind that it will not be closed if they do anything within the 3 day limit.

I like this!
Although I would waiver to push this to 5 day's. Although someone can always re-open it, if it's important.
It's too bad the `exempt-issue-labels` doesn't support regex, given we did have to update this if we create a new feature label.

Lastly, I assume this would potentially mark a lot of tickets as stale as soon as you merge it?

mattsherman:
> maybe come up with a label we can use when we are creating an issue for something that will go into zenhub

What are examples of issues that *wouldn't* go into ZenHub?

mattsherman:
> > Bear in mind that it will not be closed if they do anything within the 3 day limit.
> I like this!
Although I would waiver to push this to 5 day's. Although someone can always re-open it, if it's important.

I too would lean towards 5 days... three days is pretty common for a long weekend. So, by bumping up to 5, maybe we'd have less re-opening of PRs... though, there really isn't any additional *cost* to reopening a PR vs doing anything else to keep a PR from being closed, so maybe it is best to err on the side of closing PRs quicker.

samueljseay:
> What are examples of issues that wouldn't go into ZenHub?

@mattsherman sorry I don't think I explained this clearly, by Zenhub I actually mean one of the development pipelines in Zenhub. So things that would not go into these pipelines would be issues raised that have not been prioritised or groomed into a pipeline.

@mattsherman @louwie17 5 days sounds good. I'll tweak this.

> Lastly, I assume this would potentially mark a lot of tickets as stale as soon as you merge it?

@louwie17 yes at the "end of day" when the bot runs. My plan is to keep an eye on this myself. I will take responsibility to look through what has been marked as stale (via the new stale label) and make sure everything looks ok.

mattsherman:
> @mattsherman sorry I don't think I explained this clearly, by Zenhub I actually mean one of the development pipelines in Zenhub. So things that would not go into these pipelines would be issues raised that have not been prioritised or groomed into a pipeline.

Ah, that makes sense. Every new issue gets put in a pipeline by default though.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,23 @@
+name: 'Mark or close stale issues and PRs'
+
+on:
+    schedule:
+        - cron: '00 * * * *'
```
mattsherman, line 5: I'm not that familiar with cron syntax... is this once an hour, at the top of the hour?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,23 @@
...
+        steps:
+            - uses: actions/stale@v3.0.14
+              with:
+                  repo-token: ${{ secrets.GITHUB_TOKEN }}
+                  days-before-issue-stale: 90
+                  days-before-pr-stale: 30
+                  days-before-close: 3
+                  stale-issue-message: 'This issue has been automatically marked as stale because it has not had any recent activity. It will be closed if no further activity occurs. Thank you for your contributions.'
+                  stale-pr-message: 'This PR has been automatically marked as stale because it has not had any recent activity. It will be closed if no further activity occurs. Thank you for your contributions.'
+                  close-issue-message: 'This issue was automatically closed due to being stale. Feel free to open a new one if you still experience this problem.'
```
mattsherman, line 19: Is this what we want... to open a new issue? Or, should we suggest re-opening the existing issue?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,23 @@
...
+                  close-pr-message: 'This PR was automatically closed due to being stale.'
+                  stale-pr-label: 'status: stale'
+                  stale-issue-label: 'status: stale'
+                  exempt-issue-labels: 'cooldown period,feature: activity panel,feature: analytics,feature: components,feature: customer effort score,feature: home screen,feature: inbox,feature: marketing,feature: navigation,feature: onboarding,feature: rest api,feature: settings,feature: setup checklist,feature: wcpay,priority: high,priority: critical'
```
mattsherman, line 23: I'm not sure we need/should have the 'feature: ' labels exempt... just because an issue is (perhaps correctly) categorized doesn't mean it still isn't stale.

mattsherman (commented):
Left some thoughts in the yml.

.github/workflows/stale.yml
adrianduffell: I like re-opening the existing issue better, to avoid duplication in search results.

adrianduffell:
> Issues are given a lot more grace. An issue must be stale for 90 days before a close warning is given. The goal here is to cut back a lot of stale issues that we can't even look at right now because we have too many. 

Could we do a dry-run to see which issues would get flagged initially? I think it would be worth identifying which issues are still valuable and prioritizing them to get fixed.

.github/workflows/stale.yml
samueljseay: @mattsherman no this is once a day at "minute 0". I think reading up on cron tab patterns is an exercise for you if you're interested: https://www.netiq.com/documentation/cloud-manager-2-5/ncm-reference/data/bexyssf.html

I'll try summarise though:

cron goes: minute, hour, day(month), month, day(week).

the * means any value so this will run once at minute 0 every day.

You can also use this website to confirm how a value works: https://crontab.guru/#00_*_*_*_*

.github/workflows/stale.yml
samueljseay: Sounds good I'll reword this.

samueljseay:
@adrianduffell 
> Could we do a dry-run to see which issues would get flagged initially? I think it would be worth identifying which issues are still valuable and prioritizing them to get fixed.

I really don't think we need to do this level of overthinking it. the issue must be 90 days old with absolutely *no action on it*. No comments, no "is this being worked on" for over 3 months. This is very low risk. 

My suggestion is what I mentioned earlier: I'll do a real run and take responsibility for manually checking over whats been marked stale. But its important not to stress over things we haven't touched for a long time, or no one has raised because it means they're not high enough priority. If they were we would have fixed them or at least a discussion would be ongoing.

I expect that we may continue tweaking this over time and refining it but I think thats ok.

samueljseay:
> Could we do a dry-run to see which issues would get flagged initially? I think it would be worth identifying which issues are still valuable and prioritizing them to get fixed.

Originally I didn't want to do a dry run for this, but if it helps increase comfort with these changes then thats all good.

.github/workflows/stale.yml
samueljseay: @mattsherman Happy to remove this and see how we go. I'm changing to a dry run initially anyway so the risk is low.

samueljseay:
@adrianduffell could you take another look. I've addressed the feedback. It's on a dry run for now so we can look at the output. If I'm feeling happy with the output I'll submit a final PR to switch it on.

adrianduffell (approved):
LGTM thanks for adding the dry-run 🚀

PR #6083, opened 2021-01-17
Adjust the timing of the Marketing.php note to 5 days.

### LINKED ISSUE #5923: Timing update: Marketing.php message
**Description:** we want to change the timing of the [Marketing.php message](https://github.com/woocommerce/woocommerce-admin/blob/main/src/Notes/Marketing.php), as most users are not focused on marketing tasks during the first 3 days.

Timing: 5 days after plugin install

samueljseay:
Fixes #5923 

This ensures the Marketing.php note does not show until the store has been active 5 days.

### Testing

1. Set the `woocommerce_admin_install_timestamp` to a timestamp 5 days before today's date
2. Install WP Crontrol plugin, in the plugin's list of cron jobs, click "run now" on the `wc_admin_daily` job
3. Confirm that the marketing note is present in the home screen inbox


louwie17 (approved):
Nice work, originally was having some difficulty testing it, as WooAdmin needed to be enabled at the same time Woo did, otherwise it would add the note on activation.

Approving this, although it would be great if you could add a test for this.
Moon recently wrote some tests with similar logic in his PR, incase you wanted to mimic it: https://github.com/woocommerce/woocommerce-admin/pull/6032/files#diff-af3dfdc86549cd57fa13d923011cda7032e1a04a02bc8a2406bdbe7d9872854c

samueljseay:
@louwie17 Thanks for the reminder to add tests. will do

PR #6089, opened 2021-01-18
Try: Add the settings feature and pages

joshuatf:
Fixes #

* Adds the settings feature flags
* Adds the settings pages to WCA and navigation

Coming soon to a repo near you:
* Redirection of old pages
* Actual client-side rendering of pages

1. Check that default Settings behavior is unaffected.
2. Toggle the option for settings at `wp-admin/admin.php?page=wc-settings&tab=advanced`
3. With the new navigation off, check that the Settings menu item path is now `wc-admin&path/settings/general` (note that old URLs are temporarily broken)
4. Turn on the navigation feature (you may have to do this manually via `woocommerce_navigation_enabeld` since the settings pages are broken).
5. Check that the settings pages in the new navigation experience are using the new paths.


src/Features/Navigation/CoreMenu.php
```diff
@@ -42,6 +43,11 @@ public function init() {
 	 * Add registered admin settings as menu items.
 	 */
 	public static function get_setting_items() {
+		// Let the Settings feature add pages to the navigation if enabled.
+		if ( Loader::is_feature_enabled( 'settings' ) ) {
```
louwie17, line 47: Should this also include this condition: `&& 'yes' === get_option( 'woocommerce_settings_enabled', 'no' )` ?
Because if I enable the new nav, it only shows the `Analytics` setting for me, but maybe that's intended.

louwie17 (commented):
Things tested well following your test instructions, I just left one comment with an issue when just the new navigation is enabled, but maybe you already knew this.
But given this is merged into `main` it be nice too fix.

src/Features/Navigation/CoreMenu.php
joshuatf: > Should this also include this condition && 'yes' === get_option( 'woocommerce_settings_enabled', 'no' ) 

Good call.  Used the new `Features:is_enabled( 'settings' )` for this.

>If I enable the new nav, it only shows the Analytics setting for me, but maybe that's intended.

I think that may be related to this bug - https://github.com/woocommerce/woocommerce-admin/issues/6180

joshuatf:
Thanks for the review @louwie17!  Updated the check to see if Settings is enabled via the new feature check.

src/Features/Navigation/CoreMenu.php
louwie17: Thanks for making that change, and indeed this was also fixed with #6180

louwie17 (approved):
LGTM 🚀  👍

louwie17:
Although it should be fine, we might want to wait with merging this until 2.0 is cut, which should be at some point today.

PR #6090, opened 2021-01-18
Change from MenuGroup to label

### LINKED ISSUE #3642: SearchListControl has wrong parent-child roles relationship
**To Reproduce**
1. Install the Axe extension ([Firefox](https://addons.mozilla.org/en-US/firefox/addon/axe-devtools/), [Chrome](https://chrome.google.com/webstore/detail/axe-web-accessibility-tes/lhdoppojpmngadmnindnejefpokejbdd)) so you can do quick accessibility analyses.
2. Go to _WooCommerce_ > _DevDocs_ > scroll down to _SearchListControl_ > press _Toggle loading state_.
3. Go to the _Axe_ tab of your browser devtools and click on _Analyze_. You will see an error appear:
	>  Certain ARIA roles must be contained by particular parents

That's because [`<SearchListItem>` sets `role="menuitemradio"` or `role="menuitemcheckbox"`](https://github.com/woocommerce/woocommerce-admin/blob/master/packages/components/src/search-list-control/item.js#L65). Those two roles require the parent element to have `role="menu"` ([see example](https://www.w3.org/WAI/GL/wiki/Using_ARIA_menus#Success_Criteria_4.1.2_Name.2C_Role.2C_Value)) but the parent component is [`<MenuGroup>`](https://github.com/woocommerce/woocommerce-admin/blob/master/packages/components/src/search-list-control/index.js#L7) which has a hardcoded [`role="group"`](https://github.com/WordPress/gutenberg/blob/master/packages/components/src/menu-group/index.js#L40).

I think the way to proceed would be:
* Create a PR to `gutenberg` to add a `role` prop to `<MenuGroup>`.
* Update our current `<SearchListControl>` component to pass down a `role="menu"` to `MenuGroup`.

hsingyuc:
Fixes #3642 

Change from MenuGroup to using checkbox and radio inputs

This PR removes icons and uses the default input styling.  @Aljullu does this style work for you?

### Accessibility
-   [X] I've tested using only a keyboard (no mouse)
-   [x] I've tested using a screen reader
-   [x] I've used Axe extension

### Screenshots
checkbox
radio

1. Run storybook.
2. Add Single knob and check if renders as expected


packages/components/src/search-list-control/index.js
```diff
@@ -176,12 +176,9 @@ export class SearchListControl extends Component {
 		}
 
 		return (
-			<MenuGroup
-				label={ messages.list }
-				className="woocommerce-search-list__list"
-			>
+			<div className="woocommerce-search-list__list">
```
Aljullu, line 179: Do you think we could make this an `<ul>`? That means inside `renderList()` we will need to wrap `renderItem( ... )` with a `<li>` and it will require some CSS changes.

Aljullu (commented):
This is looking great! Thanks for working on this @hsingyuc. I tested and this seems to be much more accessible and style-wise it looks great too.

I just have a couple of suggestions:

* We should add a changelog notice to the [package CHANGELOG](https://github.com/woocommerce/woocommerce-admin/blob/main/packages/components/CHANGELOG.md).
* While not blocking, I think this is a great opportunity to update the markup so it uses a list instead of being div's. I think that will be much better for accessibility. I left an inline comment below.

Also, small note that this change will automatically fix this issue in WC Blocks: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/1243. :tada:

Aljullu:
> Add Single knob and check if renders as expected

I think a knob like this will be useful in the future. Do you think you can push the relevant code? Either in this PR or in a new one.

packages/components/src/search-list-control/index.js
```diff
@@ -138,7 +138,7 @@ export class SearchListControl extends Component {
 					depth,
 				} ) }
 				{ this.renderList( item.children, depth + 1 ) }
-			</Fragment>
+			</li>
```
Aljullu, line 141: You should keep `this.renderList( item.children, depth + 1 )` outside the `<li>`, otherwise list items are getting nested one inside the other.

I think something like this should work:

```JSX
			<Fragment key={ item.id }>
				<li>
					{ renderItem( {
						item,
						isSelected: this.isSelected( item ),
						onSelect: this.onSelect,
						isSingle,
						search,
						depth,
					} ) }
				</li>
				{ this.renderList( item.children, depth + 1 ) }
			</Fragment>
```

Aljullu (commented):
Thanks for updating the PR @hsingyuc. This is testing great! I left one more suggestion to fix an issue that causes list items to get nested, but besides that this is looking very good.

packages/components/src/search-list-control/index.js
hsingyuc: Thank you Albert!! The change has been pushed.

Aljullu (approved):
Thanks for updating the PR @hsingyuc, this looks great and it's a big step forward in accessibility. :clap: 

Approving from my side. @joshuatf do you want to take a second look before merging into WC Admin?

packages/components/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 -   Allow single-selection support to `<Search />`.
 -   Improve handling of `multiple` and `inlineTags` in `<SelectControl />`.
 -   Deprecate use of `<Card>` in favor of the `<Card>` component in `@wordpress/components`.
+-   Update `<SearchListControl />` to using checkbox and radio inputs.
```
joshuatf, line 12: One small typo here:

```suggestion
-   Update `<SearchListControl />` to use checkbox and radio inputs.
```

joshuatf (approved):
>his looks great and it's a big step forward in accessibility. 👏

Agreed! 👏 We're not currently using this component in WooCommerce Admin, but as long as this works externally using checkboxes and radios seems like a solid idea.

One small typo, that I'll get fixed for you, but otherwise LGTM.  Thanks for the fix, @hsingyuc!

hsingyuc:
@joshuatf  Thank you!!!

octaedro:
Hi @hsingyuc,
Since we updated our test suite checks, could you merge `main` again into your branch to get those checks?
Sorry for that 😕

hsingyuc:
@octaedro  Sure thing. It's done!

octaedro:
Great! I retested this PR and it's ready to merge. I'll do it 🚀 
Thank you @hsingyuc for the contribution!

PR #6108, opened 2021-01-20
WC core is on composer 2 now so no need to downgrade in travis.

samueljseay:
The builds are currently broken, I suspect this is because wc core has gone to composer 2 now so autoloader generation fails for us with our downgrade of composer in place.

If this build passes my suspicion is correct 😄


louwie17:
Re opening to try with composer 2.0.6 instead of 2.0.8

louwie17:
Closing again, didn't get a whole lot further with composer version `2.0.6` and not using `--no-dev`, seems to only install `bamarni/composer-bin-plugin` nothing else.

tomalec:
Fixing the WC version seems to help with unlocking the PR checks https://github.com/woocommerce/woocommerce-admin/pull/6109

PR #6109, opened 2021-01-20
Pin WooCommerce Version in Travis Builds

adrianduffell:
The current Travis builds are broken after the release of WooCommerce `4.9.1`, which is preventing PR checks from passing.

This PR pins the Travis build to WooCommerce `4.5.0`, our minimum supported version. 

To continue testing against the latest version of WooCommerce, a new build for it is added to the matrix and allowed to fail. This will prevent PR checks being blocked by issues with new versions of WooCommerce, whilst still providing visibility about the issue.

-   Ensure that all Travis builds pass, except for "Latest WooCommerce".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,9 @@ services:
     - mysql
     - docker
 
+env:
+    - WC_VERSION=4.5.0
+
 matrix:
     include:
         - name: 'PHP 7.3 unit tests, PHP Coding standards check and JS tests'
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,8 @@ services:
     - docker
 
 env:
-    - WC_VERSION=4.5.0
+    - global:
+          - WC_VERSION=4.5.0
 
 matrix:
     include:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -31,7 +38,7 @@ matrix:
           env: WP_VERSION=latest WP_MULTISITE=0 WP_CORE_DIR=/tmp/wordpress WC_VERSION=3.8.1
     allow_failures:
         php: 7.3
-        env: WP_VERSION=latest WP_MULTISITE=0 WP_CORE_DIR=/tmp/wordpress RUN_RANDOM=1 COMPOSER_DEV=1
```
adrianduffell, line 34: I couldn’t think of a reason why this config should be allowed to fail, so I replaced it with the `Latest WooCommerce` build config.

adrianduffell: Decided to revert replacing this so it can be investigated more.

louwie17 (approved):
Nice work @adrianduffell!
At some point we would probably want to also have a latest (stable version like 4.8.* now) as a test case, that is not allowed to fail.

PR #6115, opened 2021-01-20
Add exclusion on getOptions and method renaming

### LINKED ISSUE #6005: SelectControl excludeSelectedOptions does not work
**Describe the bug**

According to [SelectControl](https://woocommerce.github.io/woocommerce-admin/#/components/packages/select-control/README) readme, the `excludeSelectedOptions` should "exclude already selected options from the options list", but in actual result it does not do that.

**To Reproduce**

```js
/**
 * External dependencies
 */
import { SelectControl } from '@woocommerce/components';
import { useState } from '@wordpress/element';

const TestSelectControl = ( props ) => {
	const { value = [], onChange = () => {} } = props;

	const options = [
		{
			key: 'AUS',
			label: 'Australia',
			value: { id: 'AUS' },
		},
		{
			key: 'USA',
			label: 'United States of America',
			value: { id: 'USA' },
		},
	];

	return (
		<SelectControl
			multiple
			isSearchable
			inlineTags
			excludeSelectedOptions
			options={ options }
			placeholder="Start typing…"
			selected={ value }
			onChange={ onChange }
		/>
	);
};

const TestSelectControlConsumer = () => {
	const [ selected, setSelected ] = useState( [] );

	const handleChange = ( items ) => {
		setSelected( items );
	};

	return <TestSelectControl value={ selected } onChange={ handleChange } />;
};
```

**Expected behavior**

The `excludeSelectedOptions` should exclude already selected options from the options list.

**Screenshots**

The options are already selected but they are still in the options list.

image

hsingyuc:
Fixes #6005 

Add selected exclusion on getOptions and rename `filteredOptions` to `searchOptions`

### Accessibility

-   [X] I've tested using only a keyboard (no mouse)
-   [X] I've tested using a screen reader

### Screenshots
Before
before
After
after

-   npm run storybook
-   Find `selectControl` and check check that already selected options are excluded from the list.

### DIFF
--- a/packages/components/src/select-control/index.js
+++ b/packages/components/src/select-control/index.js
@@ -218,13 +218,19 @@ export class SelectControl extends Component {
 				? cacheSearchOptions
 				: this.getOptionsByQuery( cacheSearchOptions, query );
 
-		this.setState( {
-			query,
-			isFocused: true,
-			selectedIndex: 0,
-			searchOptions,
-			isExpanded: Boolean( searchOptions.length ),
-		} );
+		this.setState(
+			{
+				query,
+				isFocused: true,
+				selectedIndex: 0,
+				searchOptions,
+			},
+			() => {
+				this.setState( {
+					isExpanded: Boolean( this.getOptions().length ),
+				} );
+			}
+		);
 
 		this.updateSearchOptions( query );
 	}
@@ -253,9 +259,13 @@ export class SelectControl extends Component {
 				{
 					selectedIndex: 0,
 					searchOptions,
-					isExpanded: Boolean( searchOptions.length ),
 				},
-				() => this.announce( searchOptions )
+				() => {
+					this.setState( {
+						isExpanded: Boolean( this.getOptions().length ),
+					} );
+					this.announce( searchOptions );
+				}
 			);
 		} ) );
 	}
--- a/packages/components/src/select-control/index.js
+++ b/packages/components/src/select-control/index.js
@@ -27,7 +27,7 @@ export class SelectControl extends Component {
 		super( props );
 		this.state = {
 			...initialState,
-			filteredOptions: [],
+			searchOptions: [],
 			selectedIndex: 0,
 		};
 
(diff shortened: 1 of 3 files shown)


packages/components/src/select-control/index.js
```diff
@@ -27,16 +27,16 @@ export class SelectControl extends Component {
 		super( props );
 		this.state = {
 			...initialState,
-			filteredOptions: [],
+			searchOptions: [],
```
joshuatf, line 30: I think the renaming here is a good call and makes more sense for what these options actually are. 👍 

On a side note, instead of storing in state this would be benefit from some level of memoization, but I think this would need some testing with async functions and is probably outside the scope of this PR.

packages/components/src/select-control/index.js
```diff
@@ -217,63 +212,63 @@ export class SelectControl extends Component {
...
+				: this.getOptionsByQuery( cacheSearchOptions, query );
 
 		this.setState( {
 			query,
 			isFocused: true,
 			selectedIndex: 0,
-			filteredOptions,
-			isExpanded: Boolean( filteredOptions.length ),
+			searchOptions,
+			isExpanded: Boolean( searchOptions.length ),
```
joshuatf, line 226: Because the filtering of excluded options no longer happens before this, this can set the `<List/>` as expanded even though no options will be shown.

To reproduce, search for products on the Products Overview, select them, and then repeat the search when all those options have been selected and note the empty list.

Screen Shot 2021-01-25 at 12 33 42 PM
Screen Shot 2021-01-25 at 12 32 46 PM

We could `return null;` if no options exist in the `List`, but I think since `isExpanded` controls some accessibility features, it might be better to filter the excluded options here.  Maybe we can create a method specifically for filtering out excluded options.  E.g., `getShownOptions()` or `filterExcludedOptions()`.

joshuatf (commented):
Thanks for tackling this issue, @hsingyuc!  This is testing well for me overall.  I made note of one minor issue with an empty list showing.

There are some unrelated issues with selected items showing up in certain places, but unrelated to this PR.  I'll create an issue so we can follow up on those.

Also noting here that this component is probably due for a full refactor, but we may try and push towards a component like this in `@wordpress/components` so this is a good interim fix.

packages/components/src/select-control/index.js
hsingyuc: Do you think this way would work?  I updated it so that if there are options, we setState i`sExpanded` to `true` and vise-versus.

becdetat (approved):
This looks good and works well, thanks @hsingyuc

becdetat:
I think you'll need to pull in the latest changes in `main` to get tests passing before we can merge this.

adrianduffell:
@hsingyuc thanks for rebasing this.

There is a failing e2e test [here](https://github.com/woocommerce/woocommerce-admin/pull/6115/checks?check_run_id=1948918684#step:7:1117). Are you able to take a look?

Please just let us know if you need any support with this 🙂

hsingyuc:
@adrianduffell  Thanks! The changes have been pushed.

PR #6116, opened 2021-01-20
Try: Use settings tabs filter to get setting pages for nav

joshuatf:
Uses the tabs hook instead of the settings page hook to get settings item.

While the official docs and naming implies, I would expect the pages hook to be the right way to register pages.  However, most unofficial WooCommerce how-tos seem to use the tabs hook and this at least works with all the current use cases.

See p2-p7DVsv-ae2#comment-33691 for discussion around this.

### Screenshots
Screen Shot 2021-01-20 at 6 41 27 PM

1. Install and activate Collisimo ( https://wordpress.org/plugins/colissimo-shipping-methods-for-woocommerce/ )
2. Check that the settings tabs includes all default WC setting tabs as well as Collisimo.
3. Note that an additional Collisimo menu item will exist due to migrating un-migrated WooCommerce menu items from the old menu (discussion on whether or not we want to change this in above p2).

### DIFF
--- a/src/Features/Navigation/CoreMenu.php
+++ b/src/Features/Navigation/CoreMenu.php
@@ -42,15 +42,11 @@ class CoreMenu {
 	 * Add registered admin settings as menu items.
 	 */
 	public static function get_setting_items() {
-		$setting_pages = \WC_Admin_Settings::get_settings_pages();
-		$settings      = array();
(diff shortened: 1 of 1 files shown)


adrianduffell (approved):
This tested well and code looks good 🚀

PR #6118, opened 2021-01-21
Set height to push the bottom element down a bit -- workaround

### LINKED ISSUE #6080: Nav: Add Order Button and Screen Options conflict
**Describe the bug**

Original report in p7DVsv-ae2-p2

image

**To Reproduce**
In a narrow viewport with nav enabled, visit the order listing page and note the overlap of the screen options drop down and the Add Order buttona

moon0326:
Fixes #6080 

This PR sets the height of the #screen-meta-links to push down the bottom elements. 

This change works, but this isn't 100% ideal solution IMO. Other WordPress pages simply use `clear: both`, which works just fine. The problem with some WooCommerce Admin pages is that they have an extra header on top of the `#screen-meta-links` and the `#screen-meta-links` has added `position:relative` attribute, which prevents `clear:both` working properly.,

When I remove `position: relative` from the `#screen-meta-links`, you get a new problem when the submenu (?) gets expanded.

Screen Shot 2021-01-18 at 6 54 59 PM

It would be best to refactor the header part (not sure exactly how though), but I think it's not worth doing that to fix this small issue. The change in this PR should be enough in my opinion.

`height: $header-height + 50px;`

Other WordPress pages have 70px space between the search meta links and the text. The `+50px` is calculated. Adding +50px gives 70px between the search meta links and the [ Add Order ] button.

### Screenshots

Screen Shot 2021-01-20 at 6 33 52 PM

1. Navigate to WooCommerce -> Orders
2. If you don't see the search meta links, remove `woocommerce_marketplace_suggestions` from the `wp_options` table and refresh the page.
3. Change the view to mobile and select a small device such as Pixel 2.
4. Confirm that the search meta link does not get covered by the  [ Add Order ] button.

### DIFF
--- a/client/stylesheets/shared/_embed.scss
+++ b/client/stylesheets/shared/_embed.scss
@@ -50,6 +50,7 @@
 
 	#screen-meta-links {
 		position: relative;
+		height: $header-height + 50px;
 	}
 
 	.notice {
--- a/client/stylesheets/shared/_embed.scss
+++ b/client/stylesheets/shared/_embed.scss
@@ -49,8 +49,8 @@
 	}
 
 	#screen-meta-links {
-		position: relative;
-		height: $header-height + 50px;
+        position: relative;
+        margin-bottom: $header-height + 31px;
 	}
 
 	.notice {
--- a/client/stylesheets/shared/_embed.scss
+++ b/client/stylesheets/shared/_embed.scss
@@ -49,8 +49,10 @@
 	}
 
 	#screen-meta-links {
-        position: relative;
-        margin-bottom: $header-height + 31px;
+		position: relative;
+		@include breakpoint( '<782px' ) {
+			margin-bottom: $header-height + 31px;
+		}
 	}
 
 	.notice {


client/stylesheets/shared/_embed.scss
```diff
@@ -50,6 +50,7 @@
 
 	#screen-meta-links {
 		position: relative;
+		height: $header-height + 50px;
```
joshuatf, line 53: Maybe we could use a larger `margin-bottom` instead of an explicit height.  I'm always cautious of setting a height and creating an overflow issue if the contents change.

joshuatf (commented):
Fix works as expected; left an optional alternative to this to avoid setting an explicit height.

Also, I'm seeing some oddities with the new navigation enabled:

Screen Shot 2021-01-21 at 10 49 28 AM
Screen Shot 2021-01-21 at 10 49 19 AM

client/stylesheets/shared/_embed.scss
moon0326: Thank you @joshuatf. Let me play with margin-bottom and see if that works 👍

moon0326: @joshuatf used margin-bottom instead in https://github.com/woocommerce/woocommerce-admin/pull/6118/commits/003454356b6440637cd660689f4785c232e81192.

For the layout issue with the new navigation enabled, I think that's an existing issue. This is happening in the main branch as well. I will create a new ticket to address it.

Screen Shot 2021-01-21 at 3 26 11 PM
Screen Shot 2021-01-21 at 3 26 06 PM

client/stylesheets/shared/_embed.scss
```diff
@@ -49,7 +49,8 @@
 	}
 
 	#screen-meta-links {
-		position: relative;
+        position: relative;
+        margin-bottom: $header-height + 31px;
```
louwie17, line 53: With your other styling fix, this might have to be adjusted (or maybe you kept that in account) I couldn't get my ruler to work haha
But the gap does seem on the bigger side:
Screen Shot 2021-01-22 at 2 34 48 PM

I think it would be good to add a viewport breakpoint here as well (just for the margin-bottom) given it's not necessary on desktop.

louwie17 (approved):
Suggested one minor change, otherwise looks good I guess 😂  You are right in that the css for this stuff is a bit of a mess, we should definitely create an issue for a refactor (if there isn't already).
And the longer I look at it, I almost waiver to do it as part of these changes, although that might take a bit - probably a good cooldown issue.
I also have the ability to scroll on mine, which hides the `screen-meta-links`. 

Oooh and I believe you have some css lint errors with the tabbing of the css.

client/stylesheets/shared/_embed.scss
moon0326: thank you @louwie17 👍  I will check on the gap before merging

moon0326: @louwie17 the change is in https://github.com/woocommerce/woocommerce-admin/pull/6118/commits/6fc1d5aca6ad5a4351b5fa9adc0dcd83c4a69ec3. They both have 70x gap between the meta links and the add order button. I will test both of them together locally when https://github.com/woocommerce/woocommerce-admin/pull/6118 gets approval.

moon0326:
I don't exactly know what fixed this issue, but the issue no longer exists in the latest main branch. Closing it for now.

PR #6119, opened 2021-01-21
Add WooCommerce Version Selection to Docker PHP Test Suite

adrianduffell:
This PR adds the ability to select which WooCommerce version to use in the Docker PHP Test Suite.

```
WC_VERSION=4.9.1 npm run test:php 
```

To support switching versions, the suite needs to be re-installed with the different version of WooCommerce. To achieve this, the currently installed version of WooCommerce is stored in `/tmp/.INSTALLED_WC_VERSION`. If the value in this file is different to the selected version, the install script will automatically run. There might be a better way to store this kind of state in Docker?

WooCommerce 4.5.0 is used as the default version if none is selected. This is the [minimum supported version](https://github.com/woocommerce/woocommerce-admin/blob/34dd48c02e3296491c8aa254ce3b081cc70a526d/readme.txt#L40).

Additionally, to ensure the new `ENTRYPOINT` script changes are updated on user’s machines, an image version number tag was added to the Docker Compose config. This will trigger an image rebuild whenever we bump the version.

The README documentation was also updated.

I will follow this PR up with another to allow the WordPress version to be selected 😄 

-   Run `composer install` to ensure [compatibility](https://github.com/woocommerce/woocommerce-admin/pull/6146) with WooCommerce 4.9.1.
-   Run the test suite normally: `npm run test:php`
-   In the output, confirm the image is rebuilt.

```
Building phpunit
Step 1/12 : FROM alpine
 ```

-   Confirm the installation script is triggered and WooCommerce 4.5.0 is used by default.

```
+ LATEST_WC_TAG=4.5.0
+ cd wp-content/plugins/
+ git clone --depth 1 --branch 4.5.0 https://github.com/woocommerce/woocommerce.git
```

-   Run the test suite against the latest WooCommerce version: `WC_VERSION=4.9.1 npm run test:php`

-   Confirm the installation script is triggered and WooCommerce 4.9.1 is used.

```
+ LATEST_WC_TAG=4.9.1
+ cd wp-content/plugins/
+ git clone --depth 1 --branch 4.9.1 https://github.com/woocommerce/woocommerce.git
```

-   Run the test suite against the latest WooCommerce again: `WC_VERSION=4.9.1 npm run test:php`

-   Confirm the installation script is **NOT** triggered. PHPUnit should begin immediately.

### CODE REVIEW

**Comment by becdetat** [comment]
Possibly I'm holding something wrong but running `npm run test:php` is giving an error when pulling down phpunit:
image
Any ideas on what to do with this?

**Comment by adrianduffell** [comment]
@becdetat Thanks for trying it out!

The error seems like the PHPUnit server was down, which would be unusual. I tried it just now and it worked okay. Wondering if you are still seeing it? 🤞🤔 

This is my output, on a mac with the latest version of Docker Desktop (3.1.0).

```
Step 7/12 : RUN curl -sL https://phar.phpunit.de/phpunit-7.phar > /usr/local/bin/phpunit && chmod +x /usr/local/bin/phpunit
 ---> Running in a008f278ca31
Removing intermediate container a008f278ca31
 ---> ab7f10e87e5d
```

becdetat:
It wasn't that phar.phpunit.de was down, I was able to access it on the host, it was like the network was down in the container. Looks like restarting or upgrading Docker was the solution -  I was on 3.0.4. SMH. Thanks :-)

I'll finish up testing this today.

becdetat (approved):
Works a treat! Thanks @adrianduffell

PR #6122, opened 2021-01-21
Move inbox panel to left column, to increase user interaction

### LINKED ISSUE #5931: Home Screen: layout refinements
**Description:** after the 4.8 release we that a drop in the % of users interacting with the inbox messages (more context here: pbIJXs-vP-p2). This was expected as our main goal is to focus the user's attention on the store setup/management tasks. However, we think that users can completely miss all the notifications and we propose some refinements to the layout.

**Acceptance criteria:**

- [ ] In the two-column layout display the inbox panel below the setup and store management tasks: 
Left column: store setup and/or management tasks + inbox panel
Right column: stats overview + store management shortcuts
- [ ] In the single-column layout display the inbox panel above the stats overview if the user has the store setup tasks enabled. It should be in the following order: store setup tasks, inbox panel, stats overview.

louwie17:
Fixes #5931 

There was a drop in user interaction with the inbox panel, this PR moves the inbox panel to the left column (in two column format), and below setup tasks in the single column format.

### Screenshots

homescreen-rearangement

- Create a new woo store, and finish the onboarding wizard.
- Go to the home screen, and make sure the panels follow this order:
- Two column:
  - Left column: store setup and/or management tasks + inbox panel
  - Right column: stats overview + store management shortcuts (only shows when setup tasks is hidden)
- Single column:
  - store setup tasks, inbox panel, stats overview, store management links (only visible when setup tasks is hidden).
- Hide the setup tasks list, and see if the store management links show up in the right place.

### DIFF
--- a/client/homescreen/test/index.js
+++ b/client/homescreen/test/index.js
@@ -208,7 +208,7 @@ describe( 'Homescreen Layout', () => {
 		const { container } = render(
 			<Layout
 				requestingTaskList={ false }
-				taskListHidden={ false }
+				bothTaskListsHidden={ false }
 				query={ {} }
 				updateOptions={ () => {} }
 			/>
--- a/client/homescreen/test/index.js
+++ b/client/homescreen/test/index.js
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
-import { render, screen } from '@testing-library/react';
+import { render, screen, within } from '@testing-library/react';
 import { useUserPreferences } from '@woocommerce/data';
 
 /**
@@ -11,7 +11,7 @@ import { Layout } from '../layout';
 
 // Rendering <StatsOverview /> breaks tests.
 jest.mock( 'homescreen/stats-overview', () =>
-	jest.fn().mockReturnValue( null )
+	jest.fn().mockReturnValue( '[StatsOverview]' )
 );
 
 // We aren't testing the <TaskList /> component here.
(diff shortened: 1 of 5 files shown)


client/homescreen/test/index.js
```diff
@@ -195,4 +195,45 @@ describe( 'Homescreen Layout', () => {
 			)
 		).toHaveLength( 1 );
 	} );
+
+	it( 'should display the correct blocks in each column', () => {
```
joshuatf, line 199: Thanks for the tests! 🎉

joshuatf (approved):
This looks good and is testing well.

One oddity I noticed with the task list hidden is that a gap exists between the activity panel and inbox messages during loading:

Screen Shot 2021-01-21 at 10 27 41 AM

I think this probably already existed though so we can tackle in a follow-up.

louwie17:
> This looks good and is testing well.
> 
> One oddity I noticed with the task list hidden is that a gap exists between the activity panel and inbox messages during loading:
> 
> I think this probably already existed though so we can tackle in a follow-up.

Nice catch, it might indeed be unrelated, but apparently it's actually a loading screen that fades in and out, look closely at the gif. I wonder if it should actually be an absolute element that overlay's the placeholder, see the second part of the GIF:

inbox-loading

Although it does do it already for each individual gray field within the placeholder, so we could probably remove it. here is the GIF of the original:

old-notes-loading

@elizaan36 are you good with removing that blank fading space on top of the placeholder?

elizaan36:
> @elizaan36 are you good with removing that blank fading space on top of the placeholder? Given it was originally for the sidepanel and might not be relevant on the homescreen.

Definitely! I was going to create an issue for the larger gap between stats and Inbox panel, but this will probably fix it.

PR #6142, opened 2021-01-21
Hide tooltip in welcome modal

### LINKED ISSUE #6125: Home Screen: Welcome Modal has Horizontal Scroll Bar
**Describe the bug**
Reported by @roykho in slack - when one hovers their cursor over the [x] to close the welcome modal, a horizontal scroll bar appears at the bottom of the dialog:

image

This appears to be present in 1.8.x as well per @tammullen's comment in slack, so this doesn't seem to be a regression, but logging it as an issue for us to follow-up on. No details given on browser.

hsingyuc:
Fixes #6125 

Hide tooltip in welcome modal.

This is a temporary fix, it can be removed after https://github.com/WordPress/gutenberg/issues/28311 is fixed.

### Screenshots
before

after

-  Test via Firefox
-   Find welcome-modal file -> layout
-   Replace `shouldShowWelcomeModal` with `true` to keep the modal open
-   Hover over the close button and check that the tooltip does not render and scrollbar does not show

### DIFF
--- a/client/homescreen/welcome-modal/style.scss
+++ b/client/homescreen/welcome-modal/style.scss
@@ -8,11 +8,6 @@
 
 		.components-button {
 			color: $white;
-
-			// @todo Remove this after https://github.com/WordPress/gutenberg/issues/28311 is fixed
-			span {
-				display: none;
-			}
 		}
 	}
 
--- a/client/homescreen/welcome-modal/style.scss
+++ b/client/homescreen/welcome-modal/style.scss
@@ -8,6 +8,10 @@
 
 		.components-button {
 			color: $white;
+
+			span {
+				display: none;
+			}
 		}
 	}
 
--- a/client/homescreen/welcome-modal/style.scss
+++ b/client/homescreen/welcome-modal/style.scss
@@ -9,6 +9,7 @@
 		.components-button {
 			color: $white;
 
+			// @todo Remove this after https://github.com/WordPress/gutenberg/issues/28311 is fixed
 			span {
 				display: none;
 			}
--- a/client/layout/style.scss
+++ b/client/layout/style.scss
@@ -158,6 +158,11 @@
 	}
 }
 
+// @todo Remove this after https://github.com/WordPress/gutenberg/issues/28311 is fixed
+.components-modal__header .components-button svg + span {
+	display: none;
+}
+
 .components-modal__frame.woocommerce-usage-modal {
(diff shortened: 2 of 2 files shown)


client/homescreen/welcome-modal/style.scss
```diff
@@ -8,6 +8,11 @@
 
 		.components-button {
 			color: $white;
+
+			// @todo Remove this after https://github.com/WordPress/gutenberg/issues/28311 is fixed
```
joshuatf, line 12: Thanks for adding this comment!

client/homescreen/welcome-modal/style.scss
```diff
@@ -8,6 +8,11 @@
...
+			span {
```
joshuatf, line 13: I'd love it if this was a more specific selector, but given that this only affects the welcome modal and there are no other classes specific to this, I think this looks okay 👍

joshuatf (approved):
Thanks for the fix, @hsingyuc!  Tested on Safari, Chrome, and FF and was not able to reproduce the bug on this branch.

This LGTM while we wait for a fix upstream in GB to fix the modal tooltip issues.

joshuatf (commented):
On taking another look, I see a few other modals that may use this dismiss button and may be affected by the tooltip bug.  This PR currently fixes the welcome modal, but we could go ahead and fix all modals temporarily.

@hsingyuc would you mind moving this code into `client/layout.scss` and apply to all modals so we can fix all of them until the GB fix is merged?  Maybe we could also target `span`s only after svgs so that we only hide

```scss
// @todo Remove this after https://github.com/WordPress/gutenberg/issues/28311 is fixed
.components-modal__header .components-button svg + span {
	display: none;
}
```

hsingyuc:
@joshuatf  Thank you! The change has been pushed.

adrianduffell:
Hi @hsingyuc, could you please merge this PR with the latest main to fix the CI checks? Sorry this PR has been sitting for a while 😅 

@woocommerce/isotope I think this would be nice to make it into the 2.1 release. Is anyone free to review today?

hsingyuc:
@louwie17  👌  I rebased the PR.

adrianduffell:
Thanks @hsingyuc!

It tested well in the Welcome and Customer Effort Score Modals. LGTM 🚀 

Merging it so it lands in the 2.1 release.

PR #6147, opened 2021-01-21
Go to v3 of stale action to support the arguments we're using

samueljseay:
It's quite hard to test actions in isolation, so I didn't realise that the *latest* version of stale action does not support some of the arguments we're using.

This down-bumps stale to use v3 which still supports these extra flags we're using. I'm not sure why they were removed, and I suspect it may have been accidental. I've raised an issue: https://github.com/actions/stale/issues/287

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -8,7 +8,7 @@ jobs:
     stale:
         runs-on: ubuntu-latest
         steps:
-            - uses: actions/stale@v3.0.14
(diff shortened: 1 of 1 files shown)


samueljseay:
@adrianduffell pinging you to look at this if you can, since you're familiar. Its still going to keep doing dry-runs for now until we're ready to turn it on 😄

adrianduffell (approved):
Looks good 🚀

PR #6149, opened 2021-01-22
Increase makes ops per run for the stale action to audit what will be closed

samueljseay:
This will be removed when we go to switch it off debug mode.


adrianduffell (approved):
Looks good 🚀

PR #6152, opened 2021-01-22
Make stale bot use 60 days for PRs and issues

samueljseay:
Stale bot doesn't yet support differentiating pr and issue close time I don't think. So I've moved it to a blanket value of 60. We'll assess the dry runs and see how it goes.


adrianduffell (approved):
🚀

PR #6154, opened 2021-01-22
Turn stale bot on

samueljseay:
After looking at some dry runs such as: https://github.com/woocommerce/woocommerce-admin/runs/1747054049?check_suite_focus=true I'm confident the stale bot is conservative enough. This PR will switch it on to run 30 ops maximum every hour.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -19,7 +19,5 @@ jobs:
                   close-pr-message: 'This PR was automatically closed due to being stale.'
                   stale-pr-label: 'status: stale'
                   stale-issue-label: 'status: stale'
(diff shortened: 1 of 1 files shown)


louwie17:
> This PR will switch it on to run 30 ops maximum every hour.

I think miss read this in your last PR, but is there a particular reason you are running it every hour, instead of once a day?

samueljseay:
@louwie17 yeah, its because it runs 30 ops only per run to avoid github api limits. So it does a little bit at a time in short bursts. This seems a lot better than having it run over 100s of items once a day.

ilyasfoo (approved):
LGTM! 🚀

PR #6164, opened 2021-01-22
Migrate the Travis unit test suite to Github actions

samueljseay:
The old Travis test runner was quite limited in what versions it ran against and it was also being queued for sometimes up to half an hour while waiting for PRs. I wanted to explore moving this to Github actions.

I'm trying to strike a balance here with not testing everything because I see some problems with testing every combination we support:

1. It ends up being 40+ parallel builds that must pass for a PR to merge. It seems excessive. Its too much information for a dev to wade through and nightmarish to debug when trying to get a PR passing. (At this stage I'm thinking there are still too many testing combos in the workflow)

2. The complexity of supporting every combo is significant because we must adjust composer and phpunit versions for various combos. 

*NOTE* the event to trigger this needs to be changed to `pull_request` before merge, but I'll wait to get some review before doing that.


adrianduffell (approved):
This looks good to me 🚀 

I like how the separate GH install script is simpler than the existing script. At a later date, I think it would be worth unifying the Docker and VVV test environments around using this script too 👍

adrianduffell (changes requested):
I just noticed, I think the JS tests need to be migrated as well?

samueljseay:
@adrianduffell ha I spent so much time trying to get PHP working I forgot entirely. Thanks, I'll add JS tests back in

samueljseay:
@adrianduffell could you take another look? I've separated out the js tests into a separate job so they don't run on every matrix job.

adrianduffell (approved):
Nice work on this @samueljseay ! LGTM 🚀 

I like the separate actions for PHP/JS. Let's change the event to `pull_request` and merge 🎉

adrianduffell:
Screen Shot 2021-01-28 at 10 35 35 am

Just noting we'll need to disable the Travis check in the GitHub repo settings too.

samueljseay:
I noticed a few things I needed to finish to complete this, so I'm gonna keep working on those before getting another review.

1. Cancel previous jobs if a new PR comes in
2. Run the PHP lint and take a commit range like the old one did.
3. break out the lint jobs into separate files so they don't run for every matrix run

PR #6169, opened 2021-01-23
Remove unnecessary if statement

ribaricplusplus:
Since there is a `rm -rf $WP_TESTS_DIR` right before the deleted if statement, `$WP_TEST_DIR` will never exist, so there's no need for that. And since `set -ex` is set, if the rm command fails for any reason, the script will not continue anyway.


samueljseay (approved):
Thanks for this change, this is indeed redundant!

PR #6176, opened 2021-01-25
Disable hooks in Remote Inbox Engine

moon0326:
This PR disables various hooks used in Remote Inbox Engine in an effort to fix the ongoing issue with the WC Admin 1.8.3. [Link](https://a8c.slack.com/archives/C0E1AV8T0/p1611571234078600) to slack message.

1. Install `WP Crontrol` plugin
2. Navigate to Tools -> Cron Events 
3. Run `wc_admin_daily` job
4. Navigate to Products -> Add new and add a product
5. Navigate to Products -> All Products and edit a product

### DIFF
--- a/src/RemoteInboxNotifications/RemoteInboxNotificationsEngine.php
+++ b/src/RemoteInboxNotifications/RemoteInboxNotificationsEngine.php
@@ -26,14 +26,6 @@ class RemoteInboxNotificationsEngine {
 	public static function init() {
 		// Continue init via admin_init.
 		add_action( 'admin_init', array( __CLASS__, 'on_admin_init' ) );
-
-		// Trigger when the profile data option is updated (during onboarding).
-//		add_action(
-//			'update_option_' . Onboarding::PROFILE_DATA_OPTION,
-//			array( __CLASS__, 'update_profile_option' ),
-//			10,
-//			2
-//		);
 	}
 
 	/**
@@ -66,8 +58,10 @@ class RemoteInboxNotificationsEngine {
 			return;
 		}
 
-//		add_action( 'activated_plugin', array( __CLASS__, 'run' ) );
-//		add_action( 'deactivated_plugin', array( __CLASS__, 'run_on_deactivated_plugin' ), 10, 1 );
+		if ( 'yes' !== get_option( 'woocommerce_show_marketplace_suggestions' ) ) {
+			return;
+		}
+
 		StoredStateSetupForProducts::init();
 
 		// Pre-fetch stored state so it has the correct initial values.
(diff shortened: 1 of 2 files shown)


src/RemoteInboxNotifications/RemoteInboxNotificationsEngine.php
```diff
@@ -26,14 +26,6 @@ class RemoteInboxNotificationsEngine {
 	public static function init() {
 		// Continue init via admin_init.
 		add_action( 'admin_init', array( __CLASS__, 'on_admin_init' ) );
-
```
timmyc, line 29: Part of me thinks we leave this action in place here, but instead don't call `self::run` in the `update_profile_option` method. Or we add logic to the `run` class to not have it call out to the remote endpoint.

src/RemoteInboxNotifications/RemoteInboxNotificationsEngine.php
```diff
@@ -66,8 +58,10 @@ public static function on_admin_init() {
 			return;
 		}
 
-		add_action( 'activated_plugin', array( __CLASS__, 'run' ) );
-		add_action( 'deactivated_plugin', array( __CLASS__, 'run_on_deactivated_plugin' ), 10, 1 );
+		if ( 'yes' !== get_option( 'woocommerce_show_marketplace_suggestions' ) ) {
```
timmyc, line 61: I think we are okay to remove these two action hooks, it will definitely result in some notes appearing a day later. cc @pmcpinto

src/RemoteInboxNotifications/StoredStateSetupForProducts.php
```diff
@@ -18,8 +18,6 @@ class StoredStateSetupForProducts {
 	 * Initialize the class
 	 */
 	public static function init() {
-		add_action( 'product_page_product_importer', array( __CLASS__, 'run_on_product_importer' ) );
```
timmyc, line 21: Again wondering if we should modify the methods being called here, allow them to do some of the logic, but stop short of running the call to `RemoteInboxNotificationsEngine::run`

moon0326:
Closing it as this issue has been fixed in woocommerce.com repo.

pmcpinto:
> I think we are okay to remove these two action hooks, it will definitely result in some notes appearing a day later. cc @pmcpinto

@timmyc Do we have any messages based on those hooks?

becdetat:
@pmcpinto 
>> I think we are okay to remove these two action hooks, it will definitely result in some notes appearing a day later. cc @pmcpinto

> @timmyc Do we have any messages based on those hooks?

The "Your first product" note is. It is intended to be triggered on adding the first product.

pmcpinto:
> The "Your first product" note is. It is intended to be triggered on adding the first product.

Ok 👍

PR #6177, opened 2021-01-25
Update product template icon and copy

### LINKED ISSUE #5296: Task List: add products task - introduce templates

louwie17:
Fixes #5296 

Updated the **Start with template** task copy and update the icon.
Updated the @wordpress/icons to the latest - 2.9 so I can use the `templatePartSidebar` icon, although this will soon be renamed to `sidebar` in the new version.

### Screenshots

Screen Shot 2021-01-25 at 2 55 21 PM

- Create a new Woo Store and finish the OBW
- Select **Add my products` on the Woo homescreen.
- **Start with a template** should be present and have a recommended label.
- Icon should reflect https://share.getcloudapp.com/6quQKE6A
- Go back the home screen and make sure the task list icons are displayed correctly (The Blue icon with white checkmark)
- Visit some of the other tasks, and see if they all look good.

### DIFF
--- a/client/task-list/style.scss
+++ b/client/task-list/style.scss
@@ -13,7 +13,7 @@
 				}
 			}
 
-			.woocommerce-list__item:not( .is-complete ) {
+			.woocommerce-list__item:not(.is-complete) {
 				.woocommerce-task__icon {
 					border: 1px solid $gray-100;
 					background: $white;
@@ -57,7 +57,7 @@
 	}
 
 	.components-modal__screen-overlay {
-		background: rgba( 43, 45, 47, 0.4 );
+		background: rgba(43, 45, 47, 0.4);
 	}
 
 	.components-modal__frame {
@@ -216,7 +216,7 @@
 
 	.woocommerce-task-payment__recommended-ribbon {
 		position: absolute;
-		transform: rotate( -45deg ) translate( -50%, -50% );
+		transform: rotate(-45deg) translate(-50%, -50%);
 		background: $studio-gray-80;
 		color: $studio-white;
 		font-size: 11px;
--- a/client/task-list/style.scss
+++ b/client/task-list/style.scss
@@ -13,7 +13,7 @@
 				}
 			}
 
-			.woocommerce-list__item:not(.is-complete) {
+			.woocommerce-list__item:not( .is-complete ) {
 				.woocommerce-task__icon {
 					border: 1px solid $gray-100;
 					background: $white;
@@ -32,7 +32,7 @@
 				height: 32px;
 			}
 
-			.woocommerce-list__item-before svg {
+			.woocommerce-list__item-before .woocommerce-task__icon svg {
 				fill: $white;
 				position: relative;
 				top: 4px;
@@ -57,7 +57,7 @@
 	}
 
 	.components-modal__screen-overlay {
-		background: rgba(43, 45, 47, 0.4);
+		background: rgba( 43, 45, 47, 0.4 );
 	}
 
 	.components-modal__frame {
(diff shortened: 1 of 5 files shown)


client/task-list/style.scss
```diff
@@ -32,7 +32,7 @@
 				height: 32px;
 			}
 
-			.woocommerce-list__item-before svg {
+			.woocommerce-list__item-before .woocommerce-task__icon svg {
```
louwie17, line 35: Make this more specific so that it doesn't overwrite the default styling for icons else where in the task list. As I am using a wordpress/icon in the products task.

package.json
```diff
@@ -101,7 +101,7 @@
 		"@wordpress/hooks": "2.8.0",
 		"@wordpress/html-entities": "2.7.0",
 		"@wordpress/i18n": "3.11.0",
-		"@wordpress/icons": "2.0.0",
+		"@wordpress/icons": "2.9.0",
```
louwie17, line 104: Updated to `2.9` to use the template-part-sidebar icon, this will mostly likely get renamed to sidebar in the next version.

client/task-list/tasks/products/products.js
```diff
@@ -16,12 +17,19 @@ import ProductTemplateModal from './product-template-modal';
 const subTasks = [
 	{
 		key: 'addProductTemplate',
-		title: __( 'Start with a template (recommended)', 'woocommerce-admin' ),
+		title: (
+			<>
+				{ __( 'Start with a template', 'woocommerce-admin' ) }
+				<span className="task-list-recommended-label">
```
becdetat, line 23: There is a `Pill` component in `@woocommerce/components` that you might be able to use, see `client/profile-wizard/steps/product-types.js` for a usage (the product type selection step in the onboarding wizard)

becdetat (commented):
Everything looks good, possibly apart from the pill you've used. What do you think about using the pill from `@woocommerce/components`?

client/task-list/tasks/products/products.js
louwie17: Ooh thanks @becdetat, I didn't know about this. I tried to look for one for a little while, and couldn't find any examples either. I guess `Pill` does make sense for the name, it's tricky sometimes 🙈

louwie17: @becdetat fixed in: https://github.com/woocommerce/woocommerce-admin/pull/6177/commits/94efc8040fb76634f20216c0526f69d15b41f3df

becdetat (approved):
Looks great :-)

PR #6178, opened 2021-01-25
Delete all products before running tests

louwie17:
This should fix a random broken test that fails when products have already been created.
Deleting all the products before hand should fix this test.

**Testing instructions**
- Check if all tests for all php version in travis passed

### DIFF
--- a/tests/api/onboarding-tasks.php
+++ b/tests/api/onboarding-tasks.php
@@ -30,6 +30,16 @@ class WC_Tests_API_Onboarding_Tasks extends WC_REST_Unit_Test_Case {
 				'role' => 'administrator',
 			)
 		);
+
+		$products = wc_get_products(
+			array(
+				'return' => 'ids',
+			)
+		);
+		// delete all products.
+		foreach ( $products as $product ) {
+			$product->delete( true );
+		}
 	}
 
 	/**
--- a/tests/api/onboarding-tasks.php
+++ b/tests/api/onboarding-tasks.php
@@ -47,7 +47,7 @@ class WC_Tests_API_Onboarding_Tasks extends WC_REST_Unit_Test_Case {
 		$taxonomies = get_taxonomies();
 		foreach ( (array) $taxonomies as $taxonomy ) {
 			// pa - product attribute.
-			if ( 'pa_color' === $taxonomy || 'pa_logo' === $taxonomy) {
+			if ( 'pa_color' === $taxonomy || 'pa_logo' === $taxonomy ) {
 				unregister_taxonomy( $taxonomy );
 			}
 		}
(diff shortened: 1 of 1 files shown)


tests/api/onboarding-tasks.php
```diff
@@ -30,6 +30,16 @@ public function setUp() {
 				'role' => 'administrator',
 			)
 		);
+
+		$products = wc_get_products(
+			array(
+				'return' => 'ids',
+			)
+		);
+		// delete all products.
```
joshuatf, line 39: I feel like this kind of thing probably belongs more in a `tearDown` than `setUp`.  Are these products coming from somewhere else or are there defaults added prior to running any tests?

joshuatf (commented):
Nice job finding the cause of this.  Left a comment about where this code ideally belongs.

tests/api/onboarding-tasks.php
louwie17: I believe coming from somewhere else, as I check if at-least more then 10 are created in the product import test. If products with the same SKU already exists it skips them.

louwie17: I moved it to the tearDown, although I do call delete products at the start of the failing test as well as tearDown only gets called after the fact - https://github.com/woocommerce/woocommerce-admin/pull/6178/commits/7991abaf0cab21db177a6da9280c1916cd7e490b

joshuatf: I think I might have been unclear.  What I meant is:

Are there other tests that create products?  If so, product deletion should go into that test's tear down.  I don't think its an individual test's responsibility to reset the environment and creates a bad habit for us to not clean up after we're done testing something.

If products exist by default on all tests, then I think leaving it in the setup is okay.

That said, if this causes a lot of headaches or is difficult to track down, we can leave it in the setup and tackle stabilizing the environment in a follow-up.

louwie17: Gotcha, yeah I could spent sometime during the cooldown to have a look at that. This PR was mostly a quick fix to get the tests up and working again.
I actually removed the `delete all products` for now, and changed the deleting attribute to only color and logo, which both of them rely on. Although `unregister_taxonomy` isn't the right way to do this I have noticed.

I can certainly re-think this, but didn't want to hold up all the other PRs, would skipping them for now be better?

joshuatf (approved):
Sorry about holding you up on this @louwie17.  This addition looks good if it gets tests passing in the meantime.  Thanks for the update!

louwie17:
> Sorry about holding you up on this @louwie17. This addition looks good if it gets tests passing in the meantime. Thanks for the update!

No worries, thanks :)

PR #6179, opened 2021-01-25
Navigation: Change default location and add Product category items

psealock:
Fixes https://github.com/Automattic/wp-calypso/issues/49234

[Early Navigation feedback](https://wp.me/p7DVsv-ae2) points to users being unable to find plugins. This PR changes the default from "Settings" to "Extensions".

Also in this PR is the automatic inclusion of extension items that were under "Products" in the old navigation. They are registered by default in the same way those that fell under "WooCommerce" category.

### Screenshots

Screen Shot 2021-01-26 at 10 40 41 AM

## Problem

Clicking on `Brands` leads to the correct url, but the screen is not registered so the Nav isn't activated for that route. I tried using `Screen::add_screen` without much luck.

1. Install MailChimp
2. Install a plugin that registered in the old nav under "Products". [WooCommerce Brands](https://github.com/woocommerce/woocommerce-brands), for example.
3. See MailChimp (previously under Settings) now located under "Extensions"
4. See Brands appear under "Extensions"

### DIFF
--- a/src/Features/Navigation/Menu.php
+++ b/src/Features/Navigation/Menu.php
@@ -543,9 +543,7 @@ class Menu {
 
 				if ( isset( $query['taxonomy'] ) ) {
 					Screen::register_taxonomy( $query['taxonomy'] );
-				}
-
-				if ( isset( $query['post_type'] ) ) {
+				} elseif ( isset( $query['post_type'] ) ) {
 					Screen::register_post_type( $query['post_type'] );
 				}
 			}
--- a/src/Features/Navigation/Menu.php
+++ b/src/Features/Navigation/Menu.php
@@ -490,11 +490,13 @@ class Menu {
 	public function migrate_core_child_items( $menu ) {
 		global $submenu;
 
-		if ( ! isset( $submenu['woocommerce'] ) ) {
+		if ( ! isset( $submenu['woocommerce'] ) || ! isset( $submenu['edit.php?post_type=product'] ) ) {
 			return;
 		}
 
-		foreach ( $submenu['woocommerce'] as $key => $menu_item ) {
+		$submenu_items = array_merge( $submenu['woocommerce'], $submenu['edit.php?post_type=product'] );
+
+		foreach ( $submenu_items as $key => $menu_item ) {
 			if ( in_array( $menu_item[2], CoreMenu::get_excluded_items(), true ) ) {
 				// phpcs:disable
 				if ( ! isset( $menu_item[ self::CSS_CLASSES ] ) ) {
@@ -508,7 +510,11 @@ class Menu {
 
 			// Don't add already added items.
 			$callbacks = self::get_callbacks();
-			if ( array_key_exists( $menu_item[2], $callbacks ) ) {
+			if ( array_key_exists( htmlspecialchars_decode( $menu_item[2] ), $callbacks ) ) {
+				continue;
+			}
+
+			if ( in_array( $menu_item[2], array( 'product_importer', 'product_exporter', 'product_attributes' ), true ) ) {
 				continue;
 			}
 
--- a/src/Features/Navigation/Menu.php
+++ b/src/Features/Navigation/Menu.php
@@ -523,11 +523,12 @@ class Menu {
 
 			self::add_item(
 				array(
-					'parent'     => 'woocommerce-settings',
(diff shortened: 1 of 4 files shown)


src/Features/Navigation/Menu.php
```diff
@@ -508,17 +513,22 @@ public function migrate_core_child_items( $menu ) {
...
-			if ( array_key_exists( $menu_item[2], $callbacks ) ) {
+			if ( array_key_exists( htmlspecialchars_decode( $menu_item[2] ), $callbacks ) ) {
+				continue;
+			}
+
+			if ( in_array( $menu_item[2], array( 'product_importer', 'product_exporter', 'product_attributes' ), true ) ) {
 				continue;
 			}
 
 			self::add_item(
```
joshuatf, line 524: Can we use `add_plugin_item()` here so we can forego `menuId` and `parent` below?

joshuatf (approved):
Love this fix.  Looks much better and is testing well.  Left one quick optional update, but pre-approving 💯

psealock:
Thanks for the suggestion @joshuatf , I added that for simplicity.

I can't get the Brands plugin page to be recognized as WooCommerce without manually adding it as a taxonomy in https://github.com/woocommerce/woocommerce-admin/pull/6179/commits/afd67cdce4205529b390daf583f411cedea75cc8

Surely there is a better way? You can test by going to `/wp-admin/edit-tags.php?taxonomy=product_brand&post_type=product`

joshuatf:
@psealock That's going to be a tough issue to tackle.  If you want we can try and tackle here, but a follow-up would be okay as well.

We're only going to have access to the URL/callbacks for these unmigrated items.  The best solution may be to be parse the URL and get any taxonomies or post types and register them.

psealock:
>The best solution may be to be parse the URL and get any taxonomies or post types and register them.

Thanks for the suggestion. I applied this in https://github.com/woocommerce/woocommerce-admin/pull/6179/commits/25cf30634c815ed22f96c03084fa1f1beff8eb0a and its working well. I just need a sanity check that there aren't negative consequences with registering something that should not be registered. This may cause duplicate taxonomies or post types, but I think thats ok because the arrays are being used with a simple `in_array` check.

https://github.com/woocommerce/woocommerce-admin/blob/25cf30634c815ed22f96c03084fa1f1beff8eb0a/src/Features/Navigation/Screen.php#L116-L117

src/Features/Navigation/Menu.php
```diff
@@ -506,21 +511,44 @@ public function migrate_core_child_items( $menu ) {
...
+			// Determine if migrated items are a taxonomy or post_type. If they are, register them.
+			$parsed_url   = wp_parse_url( $menu_item[2] );
+			$query_string = isset( $parsed_url['query'] ) ? $parsed_url['query'] : false;
+
+			if ( $query_string ) {
+				$query = array();
+				parse_str( $query_string, $query );
+
+				if ( isset( $query['taxonomy'] ) ) {
+					Screen::register_taxonomy( $query['taxonomy'] );
```
joshuatf, line 545: The only way I can think of to break this is if a taxonomy was part of WooCommerce, but the associated post type was not.  E.g. a `product_label` taxonomy was added to the default `post` post type which would register `post` as a WooCommerce screen since the URL for taxonomies includes the post type `wp-admin/term.php?taxonomy=product_label&tag_ID=209&post_type=post`

This seems pretty edge case, but we could fix by doing an if/else on this logic.  I think we're also fairly safe as is, so I'll leave this up to you.

joshuatf (approved):
>This may cause duplicate taxonomies or post types, but I think thats ok because the arrays are being used with a simple in_array check.

Agree that this is pretty safe as is.  If we want to be double careful we could optionally add `! in_array()` checks to `register_post_type()` and `register_taxonomy()`.

This is testing really for me though.  Left another optional safeguard in the code, but this gets a 👍 from me.

psealock:
I added `in_array` checks in https://github.com/woocommerce/woocommerce-admin/pull/6179/commits/e299896db8d6bb8a0de24eaff2010d03b9c0783d and if/else logic such that an item will register a taxonomy or post_type, but not both in https://github.com/woocommerce/woocommerce-admin/pull/6179/commits/cdaa960ee803e8735418d508ccaa25c71a08e773

PR #6182, opened 2021-01-26
Hide Inventory and Reviews panels if store setup task list is visible

### LINKED ISSUE #6095: Inventory and Reviews empty states are still being displayed alongside the store setup task list
**Describe the bug**
The Inventory and Reviews empty states are being displayed alongside the store setup task list, which can distract and confuse the user. #5952 should have addressed this, so I don't know what happened. Ideally this should be fixed in 5.0 

**To Reproduce**

1. Create a new store with 4.9 installed
2. Complete or skip the store profiler
3. Visit the home screen
4. See the inventory and reviews empty states.

**Expected behavior**
To only see the store setup tasks, as I'm a new user and I still don't have any products, reviews or orders

**Screenshots**
https://d.pr/i/K9RXOP

octaedro:
Fixes #6095

This PR adds the code to hide the `Inventory` and `Reviews` panels if the store setup task list is visible.

### Screenshots
Screen Capture on 2021-01-26 at 12-14-52

- Create a brand new site (e.g.: using Jurassic Ninja)
- Upload a test version of this branch (you can get it by running `npm run test:zip`).
- Finish the OBW.
- Add a new product.
- Hide the task list named "Finish setup".
- Verify the `Reviews` panel is visible now.
- Go to a non-js page (e.g.: `/wp-admin/admin.php?page=wc-settings`).
- Click on `Help` > `Setup wizard` and enable the `Task list`.
- Go to the `Home` screen.
- Verify the `Reviews` panel is not visible anymore.

cc: @pmcpinto

### DIFF
--- a/client/homescreen/activity-panel/index.js
+++ b/client/homescreen/activity-panel/index.js
@@ -26,7 +26,7 @@ import { getAllPanels } from './panels';
 import { getUnapprovedReviews } from './reviews/utils';
 
 export const ActivityPanel = () => {
-	const settings = useSelect( () => {
+	const settingsData = useSelect( () => {
 		const manageStock = getSetting( 'manageStock', 'no' );
 		const publishedProductCount = getSetting( 'publishedProductCount', 0 );
 		const reviewsEnabled = getSetting( 'reviewsEnabled', 'no' );
@@ -65,9 +65,9 @@ export const ActivityPanel = () => {
 	} );
 
 	const panels = getAllPanels( {
-		...settings,
 		...ordersData,
 		...reviewsData,
+		...settingsData,
 		...taskListData,
 	} );
 
--- a/client/homescreen/activity-panel/index.js
+++ b/client/homescreen/activity-panel/index.js
@@ -11,6 +11,7 @@ import {
 	__experimentalText as Text,
 } from '@wordpress/components';
 import { getSetting } from '@woocommerce/wc-admin-settings';
+import { OPTIONS_STORE_NAME } from '@woocommerce/data';
 
 /**
  * Internal dependencies
@@ -46,8 +47,14 @@ export const ActivityPanel = () => {
 			publishedProductCount,
 		};
 	} );
+	const setupTaskListData = useSelect( ( select ) => {
+		const { getOption } = select( OPTIONS_STORE_NAME );
+		return {
+			isTaskListHidden: getOption( 'woocommerce_task_list_hidden' ),
+		};
+	} );
 
-	const panels = getAllPanels( panelsData );
+	const panels = getAllPanels( { ...panelsData, ...setupTaskListData } );
 
 	if ( panels.length === 0 ) {
 		return null;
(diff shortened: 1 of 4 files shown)


client/homescreen/activity-panel/index.js
```diff
@@ -25,29 +26,50 @@ import { getAllPanels } from './panels';
 import { getUnapprovedReviews } from './reviews/utils';
 
 export const ActivityPanel = () => {
-	const panelsData = useSelect( ( select ) => {
-		const totalOrderCount = getSetting( 'orderCount', 0 );
-		const orderStatuses = getOrderStatuses( select );
-		const reviewsEnabled = getSetting( 'reviewsEnabled', 'no' );
-		const countUnreadOrders = getUnreadOrders( select, orderStatuses );
+	const settingsData = useSelect( () => {
```
louwie17, line 29: @octaedro is there a particular reason you split up the `useSelect` into 3 different one's?
Not a huge issue, but I believe keeping a single `useSelect` for all the data is more efficient then multiple `useSelects`.

louwie17 (approved):
Tested well, thanks for re-addressing this @octaedro :) It also stayed hidden even if there was a unapproved review, but appeared once the task list is hidden.
I just left one comment.

client/homescreen/activity-panel/index.js
octaedro: Hi @louwie17, thank you for your review.
As I was consulting 3 different stores I thought it would be a good idea to use different `useSelect`s. So if the reviews store changes we won't call the other two selectors on the other stores (that won't return anything new).
The difference is pretty small and I'm totally ok with the change you suggested since it makes this code easier to read. Please let me know if I'm missing something.

louwie17: > So if the reviews store changes we won't call the other two selectors on the other stores (that won't return anything new).

Unfortunately that's not the case, you did have to get the store object (I think returned by the `register` function, although that's not common practice).
The `useSelect` is triggered when any registered store updates, so in this case all three `useSelect` will get triggered if either the reviews store, or any of the other registered stores are updated.

psealock: >So if the reviews store changes we won't call the other two selectors on the other stores (that won't return anything new).

I think thats ok because selector results are cached based on supplied arguments. Lets fix this up before merge

psealock: It seems like a single useSelect is slightly faster unless they are from different stores, in which case its slightly faster, but will probably change. So your choice @octaedro. I'm ok with either.

p7H4VZ-2Q6-p2

octaedro: Ok, sounds good to me 👍
This change was added in the commit 69f1809

louwie17: @octaedro re-tested it locally and all is good, although one odd thing with the logic, is that the `Orders` panel will be displayed, but the reviews/stock panels won't until the task list is hidden.

psealock (approved):
Thanks for changes @octaedro LGTM 🚢

PR #6185, opened 2021-01-26
Add wait for mysql script to php unit entrypoint

### LINKED ISSUE #5907: MySQL Service Intermittently Unavailable When Docker PHP Test Suite Runs
**Describe the bug**

Occasionally , the MySQL service is not ready when the Docker PHP Test Suite is run and results in an error.

**To Reproduce**

It is difficult to reproduce because it depends on how quickly the MySQL service starts. To try to trigger the error, the app can be repetitively re-started with:

```
npm run test:php; docker-compose -f docker/wc-admin-php-test-suite/docker-compose.yml stop 
```

**Expected behavior**
The app should wait for MySQL to be ready before proceeding. A `wait-for` script could be used to solve this.

**Additional context**

@louwie17  said in the [comment](https://github.com/woocommerce/woocommerce-admin/pull/5297#pullrequestreview-554089030) :

> When I first ran npm run test:php the mysql didn't quite seem to be booted up yet, so it failed when trying to run the setup script (not a stopper for me).
I fixed it by running docker volume rm -f wc-admin-php-test-suite_test-suite after, and then running npm run test:php.
We could add something like this postgres example (the wait-for-postgres.sh script): https://docs.docker.com/compose/startup-order/ to the entrypoint.sh script.

louwie17:
Fixes #5907 

Add script that sleeps for 2seconds if the mysql container isn't up yet. Also added a `docker-compose down`, so the containers are removed after running.

### Screenshots

Screen Shot 2021-01-26 at 2 58 31 PM

- If you have run this script before, run:
  - `docker-compose -f docker/wc-admin-php-test-suite/docker-compose.yml down`
  - `docker image rm wc-admin-php-test-suite_phpunit` (removing the original docker image it created)
- Run `npm run test:php`
- This should pass successfully, or at-least run the tests successfully.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
 		"test": "./node_modules/jest-24.9.0/bin/jest.js --config tests/js/jest.config.json",
 		"test-staged": "./node_modules/jest-24.9.0/bin/jest.js --bail --config tests/js/jest.config.json --findRelatedTests",
 		"test:help": "wp-scripts test-unit-js --help",
-		"test:php": "docker-compose -f docker/wc-admin-php-test-suite/docker-compose.yml run --rm phpunit",
+		"test:php": "docker-compose -f docker/wc-admin-php-test-suite/docker-compose.yml run --rm phpunit ; docker-compose -f docker/wc-admin-php-test-suite/docker-compose.yml down",
 		"test:update-snapshots": "jest --updateSnapshot --config tests/js/jest.config.json",
 		"test:watch": "npm run test -- --watch",
 		"test:zip": "npm run clean && composer i && ./bin/build-test-zip.sh",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -64,7 +64,7 @@
 		"test": "./node_modules/jest-24.9.0/bin/jest.js --config tests/js/jest.config.json",
 		"test-staged": "./node_modules/jest-24.9.0/bin/jest.js --bail --config tests/js/jest.config.json --findRelatedTests",
 		"test:help": "wp-scripts test-unit-js --help",
-		"test:php": "docker-compose -f docker/wc-admin-php-test-suite/docker-compose.yml run --rm phpunit",
+		"test:php": "docker-compose -f docker/wc-admin-php-test-suite/docker-compose.yml run --rm phpunit ; docker-compose -f docker/wc-admin-php-test-suite/docker-compose.yml down",
```
louwie17, line 67: The `docker-compose run` command seems to only remove the `phpunit` container, but leaves the `mysql` container running.
Which is why I add the `docker-composer down`, which kills and removes the container (this is run even if the phpunit tests fail).

adrianduffell (approved):
This works just nice! 🚀

PR #6188, opened 2021-01-26
Return countryInfo as an empty object if not in locale info

### LINKED ISSUE #6187: Country without locale info fails during setup wizard
**Describe the bug**
When going through the setup wizard, a country not included in WC's default locale info results in an error when trying to proceed and breaks the OBW flow.

**To Reproduce**

1. Go to the setup wizard. `wp-admin/admin.php?page=wc-admin&path=%2Fsetup-wizard&step=store-details`
2. Select a country not included in `woocommerce/i18n/locale-info.php` (e.g., `Åland Islands`)
3. Click `Continue`.
4. Note the console error.

**Expected behavior**
Expected to be able to continue, even if currency/locale info is not found.

**Screenshots**
Screen Shot 2021-01-26 at 3 55 48 PM

hsingyuc:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/6155

Return `countryInfo` as an empty object if not in locale info

螢幕快照 2021-01-26 下午5 02 35

### DIFF
--- a/packages/currency/src/index.js
+++ b/packages/currency/src/index.js
@@ -113,10 +113,10 @@ const CurrencyFactory = ( currencySetting ) => {
 		localeInfo = {},
 		currencySymbols = {}
 	) {
-		const countryInfo = localeInfo[ countryCode ];
+		const countryInfo = localeInfo[ countryCode ] || {};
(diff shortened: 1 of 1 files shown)


joshuatf (approved):
Fixes the console error, thank you for the quick fix 🎉

PR #6209, opened 2021-01-27
Skip failing tests in WP 5.3 and 5.4

### LINKED ISSUE #6194: Incompatibility with Wordpress 5.3 and 5.4 (not present in 5.5+)
While building the new unit test matrix in #6164 I noticed a consistent failure in matrix runs of Wordpress 5.3 and 5.4.

The stack trace for the test failure:

```
3) WC_Tests_API_Reports_Orders::test_custom_product_attributes_filter
Undefined index: items
/tmp/wordpress/wp-includes/rest-api.php:1237
/tmp/wordpress/wp-includes/rest-api.php:1237
/tmp/wordpress/wp-includes/rest-api.php:1037
/tmp/wordpress/wp-includes/rest-api/class-wp-rest-request.php:885
/tmp/wordpress/wp-includes/rest-api/class-wp-rest-server.php:884
/tmp/wordpress/wp-content/plugins/woocommerce/tests/legacy/framework/vendor/class-wp-test-spy-rest-server.php:35
/tmp/wordpress/wp-content/plugins/woocommerce-admin/tests/api/reports-orders.php:303
```

It seems this is a real incompatibility that needs investigating and fixing. But first we need to reproduce it in a local dev site. To move forward the unit test migration we've opted to leave 5.3 and 5.4 out of our test runs until we can fix this.

moon0326:
Fixes #6194 

This PR skips the tests that fail only in WP 5.3 and 5.4 due to the difference in how rest-api.php handles schema validation.

More details on here https://github.com/woocommerce/woocommerce-admin/issues/6194#issuecomment-768052452

1. Remove the existing phpunit volume `docker volume rm -f wc-admin-php-test-suite_test-suite`
2. Open docker/wc-admin-php-test-suite/entrypoint.sh
2. Change the following line.

Before: `	bin/install-wp-tests.sh $DB_NAME $DB_USER $DB_PASS $DB_HOST latest true`
After: `	bin/install-wp-tests.sh $DB_NAME $DB_USER $DB_PASS $DB_HOST 5.3 true`

3. Run the test normally and confirm three tests are skipped.

### DIFF
--- a/tests/api/reports-orders-stats.php
+++ b/tests/api/reports-orders-stats.php
@@ -134,7 +134,7 @@ class WC_Tests_API_Reports_Orders_Stats extends WC_REST_Unit_Test_Case {
 	public function test_product_attributes_filter() {
 		global $wp_version;
 		if ( version_compare( $wp_version, '5.5', '<' ) ) {
-			$this->markTestSkipped();
+			$this->markTestSkipped( 'Skipped in older versions of WordPress due to a bug in WP when validating arrays.' );
 		}
 
 		global $wpdb;
--- a/tests/api/reports-orders-stats.php
+++ b/tests/api/reports-orders-stats.php
@@ -132,6 +132,11 @@ class WC_Tests_API_Reports_Orders_Stats extends WC_REST_Unit_Test_Case {
 	 * Test filtering by product attribute(s).
 	 */
 	public function test_product_attributes_filter() {
+		global $wp_version;
+		if ( version_compare( $wp_version, '5.5', '<' ) ) {
+			$this->markTestSkipped();
+		}
+
 		global $wpdb;
 		wp_set_current_user( $this->user );
 		WC_Helper_Reports::reset_stats_dbs();
--- a/tests/api/reports-orders.php
+++ b/tests/api/reports-orders.php
@@ -135,7 +135,7 @@ class WC_Tests_API_Reports_Orders extends WC_REST_Unit_Test_Case {
 	public function test_product_attributes_filter() {
 		global $wp_version;
 		if ( version_compare( $wp_version, '5.5', '<' ) ) {
-			$this->markTestSkipped();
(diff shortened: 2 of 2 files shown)


tests/api/reports-orders.php
```diff
@@ -236,6 +241,11 @@ public function test_product_attributes_filter() {
 	 * Test filtering by custom product attribute(s).
 	 */
```
adrianduffell, line 242:
```suggestion
	 *
	 * Skipped in WP 5.3 & 5.4 due to a bug in WordPress when validating arrays.
	 */
```

tests/api/reports-orders.php
```diff
@@ -133,6 +133,11 @@ public function test_reports_schema() {
 	 * Test filtering by taxonomy-backed product attribute(s).
 	 */
```
adrianduffell, line 134:
```suggestion
	 *
	 * Skipped in WP 5.3 & 5.4 due to a bug in WordPress when validating arrays.
	 */
```

tests/api/reports-orders-stats.php
```diff
@@ -132,6 +132,11 @@ public function test_reports_schema() {
 	 * Test filtering by product attribute(s).
 	 */
```
adrianduffell, line 133:
```suggestion
	 *
	 * Skipped in WP 5.3 & 5.4 due to a bug in WordPress when validating arrays.
	 */
```

adrianduffell (approved):
Nice investigation on this issue @moon0326 👍 

Code looks good and the tests pass for me. Could we add some documentation about why the tests are skipped? I left some suggestions you could commit. Then I think it's ready to merge.

tests/api/reports-orders.php
moon0326: Thank you for the suggestion @adrianduffell 👍 I will commit that.

adrianduffell: That's perfect, thanks!

PR #6200, opened 2021-01-27
Change siteUrl to shopUrl

hsingyuc:
Fixes #6037 

Change siteUrl to shopUrl

### Screenshots
螢幕快照 2021-01-27 下午4 36 20

### Detailed test instructions:
-  Hide the task list
-  Click on `view my store` button and see if redirected to the shop page

PR #6222, opened 2021-01-28
Replaced method `woocommerce_get_page_id` with `wc_get_page_id`

octaedro:
This is a follow-up PR for [6200](https://github.com/woocommerce/woocommerce-admin/pull/6200)

This PR replaces method `woocommerce_get_page_id` with `wc_get_page_id` in `Loader.php`.
The method `woocommerce_get_page_id` has been deprecated since `WooCommerce 3.0`, so we are replacing it with the recommended one (`wc_get_page_id`).

### Screenshots
screenshot-one wordpress test-2021 01 28-15_09_42

- Check out the `main` branch (`git checkout main`).
- Turn on the debug mode adding `define( 'WP_DEBUG', True );` in `wp-config.php`.
-  Confirm you see the deprecation warning.
- Check out this branch (`git checkout fix/6037_update`)
- Verify the warning is not present anymore.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -1099,7 +1099,7 @@ class Loader {
 		$settings['wcAdminAssetUrl'] = plugins_url( 'images/', dirname( __DIR__ ) . '/woocommerce-admin.php' );
 		$settings['wcVersion']       = WC_VERSION;
 		$settings['siteUrl']         = site_url();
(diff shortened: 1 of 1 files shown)


adrianduffell (approved):
Thanks for fixing this @octaedro, it tested well and LGTM 🚀

PR #6224, opened 2021-01-28
Adding text labels to elements that were previously invisible to screen readers

### LINKED ISSUE #6201: Missing text content/aria labels on wp-admin components
**Describe the bug**
As part of an accessibility audit, I've noticed a few items within `wc-admin` that are invisible to screen readers due to missing text content, or other labels.

Items:
- Logo in new navigation that toggles the navigation visibility
- Select above charts in analytics reports that allows you to change the period type

**To Reproduce**

1. Use a screen reader
2. Navigate via the keyboard (tab/shift-tab) to the described components
3. Notice that no audio description is given to describe the state.

**Expected behavior**
The content should be provided that allows the screen reader to describe the state/control accurately to the user

joelclimbsthings:
Fixes #6201 

Adding text-readable labels to components within `wc-admin` that are currently invisible to screen readers.

### Accessibility

Oh yes.

### Screenshots

image

image

-   Enable screen reader
-   Enable new navigation
- Bring WC logo within navigation sidebar into focus. 
- The screenreader should recite "toggle navigation button," or something like that.
- Go to Analytics -> Overview
- Focus select box above charts, as pictured in screenshots section. 
- Screen reader should recite "chart period"

### DIFF
--- a/client/dashboard/dashboard-charts/index.js
+++ b/client/dashboard/dashboard-charts/index.js
@@ -73,6 +73,7 @@ const renderIntervalSelector = ( { chartInterval, setInterval, query } ) => {
 				value: allowedInterval,
 				label: intervalLabels[ allowedInterval ],
 			} ) ) }
+			aria-label="Chart period"
 			onChange={ setInterval }
 		/>
 	);
--- a/client/navigation/components/header/index.js
+++ b/client/navigation/components/header/index.js
@@ -105,6 +105,7 @@ const Header = () => {
 			<Button
 				onClick={ () => toggleFolded() }
 				className="woocommerce-navigation-header__site-icon"
+				aria-label="Toggle navigation"
 			>
 				{ buttonIcon }
 			</Button>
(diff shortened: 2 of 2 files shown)


psealock (commented):
Using Apple's built in Screen reader, I'm getting "Toggle navigation" which is a great improvement. Do you think this should be a toggle or tick box role? The user won't know the state of the dropdown and thus their ability to tab to Nav links.

joelclimbsthings:
> Using Apple's built in Screen reader, I'm getting "Toggle navigation" which is a great improvement. Do you think this should be a toggle or tick box role? The user won't know the state of the dropdown and thus their ability to tab to Nav links.

Good call @psealock . Was just toying with this a bit, and there were two options that seemed workable:
- Change the `aria-label` from _"Fold Navigation"_ to _"Expand Navigation"_ depending on the state of the toggle.
- Add the `switch` role and checked state as mentioned.

I went with the second option for now in the most recent commit, although I _was_ tempted by the first. This option just seemed more in line with what I've seen done, and consistency with other implementations won out.

psealock (approved):
Nice solution! Works great 🚢

PR #6226, opened 2021-01-28
Adding checkbox role to chart legend component to make it compatible with screen readers

### LINKED ISSUE #6206: Custom Checkbox Component used in analytics -> reports not accessible via keyboard, or with screen-reader
**Describe the bug**
The checkboxes used to toggle the comparison periods in the analytics charts are not accessible. They're not semantically recognizable as checkboxes at all, so screen readers cannot tell if they're checked/unchecked. 

**To Reproduce**

- Enable a screen reader
- Use a keyboard to _tab_ to the pictured checkboxes within a report
- Use the `spacebar` to toggle the checkbox, and notice that the screen-reader cannot tell you whether it's checked or not.

**Screenshots**

image

**Expected behavior**
The checkboxes should be structured in a way that allows the screen reader to recognize that it's a checkbox, and what it's current state is.

joelclimbsthings:
Fixes #6206 

The legend component used within the Analytics charts functions as a checkbox, but uses non-semantic markup to implement the functionality. This results in screen readers not giving audio feedback to indicate that it is a checkbox, and whether it's checked or not. 

This PR takes the minimal approach of adding aria labelling to provide this information. Another approach would be to replace the markup with more semantic options, but doesn't seem necessary in this case. 

### Accessibility

:+1: 

### Screenshots

image

-  Enable a screen reader (I'm using NVDA)
-  Navigate to Analytics -> Products
- Navigate to the pictured checkbox above the chart, used for toggling the comparison periods on the chart below. They're also pictured in the screenshots section
- The screen reader should indicate that it is a checkbox. 
- Hit the `spacebar` key to toggle the checkbox. The screen reader should indicate that it is either checked or unchecked.

### DIFF
--- a/packages/components/src/chart/d3chart/legend.js
+++ b/packages/components/src/chart/d3chart/legend.js
@@ -105,6 +105,8 @@ class D3Legend extends Component {
 							onFocus={ handleLegendHover }
 						>
 							<button
+								role="checkbox"
+								aria-checked={ row.visible ? 'true' : 'false' }
 								onClick={ handleLegendToggle }
 								id={ `woocommerce-legend-${ instanceId }__item-button__${ row.key }` }
 								disabled={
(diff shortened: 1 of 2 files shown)


psealock (approved):
Lovely! Working well and a big improvement.

PR #6227, opened 2021-01-29
Adding role to focusable elements in analytics charts for screen reader accessibility

### LINKED ISSUE #6204: Charts in analytics->reports present issues when navigating via keyboard and with a screen reader
**Describe the bug**
While navigating via keyboard, each data point within the charts is tabbable. Given the sheer amount of data points for a chart, this can become a bit of a black hole while navigating via the `tab` key. 

While this could be expected and useful behavior, each data point is unlabed and actually results in the screen reader (NVDA) reciting the world "blank" for each data point. If tabbing behavior is retained for these charts, each point should at least provide useful information.

**To Reproduce**

- Navigate to Analytics -> Products
- Navigate to the chart via the keyboard
- Notice that each data point becomes tabable, and you need to tab through each one to navigate beyond the chart.
- Enable a screen reader, and notice that it recites the world "blank" for each point in the chart.

**Expected behavior**
If tabbing is supported for each point/period within the chart, it should provide useful information to the screen readers.

joelclimbsthings:
Fixes #6204 

Using some screen readers, such as NVDA, tabbing through the analytics charts results in every data point being read as "blank," providing no useful information and possibly endangering the sanity of our users. 

This took a while to track down. It appears that it was only an issue with the particular screen reader I was using (and potentially others), and is something of a known issue:

https://github.com/nvaccess/nvda/issues/7807

Despite that, it seems to be expected and intended behavior for now. I did find that adding a `role` of `graphics-symbol` to the focusable elements actually does trigger the screen reader to read out the aria label (which was already defined). I've added this, although it's a bit debatable on whether this is our issue or NVDA's. 

- Checkout branch
-   Enable NVDA as a screen reader
- Navigate via keyboard (tab key) to the chart, and you'll find that each data point has intentionally been made focusable. When focusing each data point, you should hear the data and number read, and not just the world "blank" as it previously was.

### DIFF
--- a/packages/components/src/chart/d3chart/utils/line-chart.js
+++ b/packages/components/src/chart/d3chart/utils/line-chart.js
@@ -170,6 +170,7 @@ export const drawLines = ( node, data, params, scales, formats, tooltip ) => {
 			.attr( 'cx', ( d ) => scales.xScale( moment( d.date ).toDate() ) )
 			.attr( 'cy', ( d ) => scales.yScale( d.value ) )
 			.attr( 'tabindex', '0' )
+			.attr( 'role', 'graphics-symbol' )
 			.attr( 'aria-label', ( d ) => {
 				const label = formats.screenReaderFormat(
 					d.date instanceof Date ? d.date : moment( d.date ).toDate()


samueljseay:
@joelclimbsthings NVDA is windows only right? are you testing this with one of the windows IE testing VMs?

psealock:
>Navigate via keyboard (tab key) to the chart

Using Chrome and Firefox I can't get the data point to be tabbable. It goes from the summary numbers to the some controls and then segments, but won't stop on data points.

Are you getting data points to be tabbable? I'm wondering what I'm doing wrong because that seems really useful

joelclimbsthings:
> Using Chrome and Firefox I can't get the data point to be tabbable. It goes from the summary numbers to the some controls and then segments, but won't stop on data points.
> 
> Are you getting data points to be tabbable? I'm wondering what I'm doing wrong because that seems really useful

Interesting @psealock , I was able get this to behave the way you described as well, and tracked it down to this line in the same file:
```
	width / params.uniqueDates.length > minDataPointSpacing &&
		series
			.selectAll( 'circle' )
```

It looks like the data points that I'm modifying for accessibility only actually render depending on the viewport width and the number of data points. If you enlarge your viewport and ensure your there aren't too many datapoints (looks like no more than 36), then it should be tabbable as I described.

psealock (approved):
Ok, got this to work and its a nice improvement for sure. I did notice a bug while here though.

The aria-label is the date found on the x-axis, not the the date for the particular segment. In other words, if comparing against last year, the aria label for last year's segment will be created from the current year.

My guess as to why this is done is because the date needs to match the primary segment otherwise the x-scale wouldn't work and the segments wouldn't be on top of each other. The tooltip however has the right data, so it's in there somewhere. 

Feel free to deal with it here, create a follow up, or simply make an issue and move on.

joelclimbsthings:
> Feel free to deal with it here, create a follow up, or simply make an issue and move on.

Great observation @psealock . I decided to go ahead and create a [separate issue here](https://github.com/woocommerce/woocommerce-admin/issues/6255).

PR #6232, opened 2021-01-29
Addressing 'sort by undefined' issue with screen readers in reports t…

### LINKED ISSUE #6207: Navigating the Reports table via the keyboard, with a screen-reader, recites "Sort by undefined" for each cell
**Describe the bug**
When utilizing a screen reader to navigate the reports table, it will recite "Sort by undefined" repeatedly for each cell. 

**To Reproduce**

- Enable a screen reader
- Navigate to Analytics report page with data table
- Navigate over the table via the keyboard, and listen to the screen reader. You'll notice it recite "Sort by undefined" for each cell. 

**Expected behavior**
The sort-by should either be surprised, or provide the actual value that the table is sorted by.

joelclimbsthings:
Fixes #6207 

This is addressing an issue with screen reader text explicitly provided that reads "sort by undefined" when tabbing through the report table, for each cell.

It looks like this stems from the fact that a `screenReaderText` variable is provided and utilized from the `headers` object, but it's meant to fallback to the standard label when not present. This fallback was not in place.

### Screenshots

image

- Checkout branch
-   Enable a screen reader
- Navigate via keyboard to the report table, and tab through the cells.
- Each cell should cause the screen reader to recite "sort by -name of column," instead of "sort by undefined." Among other things.

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -9,6 +9,7 @@
 -   Allow single-selection support to `<Search />`.
 -   Improve handling of `multiple` and `inlineTags` in `<SelectControl />`.
 -   Deprecate use of `<Card>` in favor of the `<Card>` component in `@wordpress/components`.
+-   Fixing screen reader text being undefined for report `<Table>`
 
 # 5.1.2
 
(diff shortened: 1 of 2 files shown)


psealock (approved):
👍🏽 Changes make sense to me. Only thing is a changelog entry for packages

PR #6233, opened 2021-01-29
Refactor panel with withFocusOutside

### LINKED ISSUE #6074: [GlobalStep - WooCommerce 4.9] "Display" option fails to collapse upon invoking "Help" option on header tab for "WooCommerce->Home" Page
## Bug Description:
"Display" option fails to collapse when "Help" option on header tab is invoked on "WooCommerce->Home" Page

## Environment:
Windows 10, macOS Mojave v10.14.6 (18G103)
Browser: Chrome(Version 87.0.4280.88 (64-bit))
                Firefox(Version 83.0)(64-bit)
                Safari: v13.1.1 (14609.2.9.1.3)
                
 
## Steps To Reproduce:
1. Create any test site using Jurassic Ninja.
1. Install and activate all required plugins.
1. Complete the Onboarding setup wizard 
1. Goto WooCommerce->Home
1. Click on "Display" option
1. Click on "Help" option
1. Observe that "Display" option fails to collapse when "Help" option on header tab is invoked

## Actual Result:
"Display" option fails to collapse when "Help" option on header tab is invoked on "WooCommerce->Home" Page

## Expected Result:
"Display" option should collapse when "Help" option on header tab is invoked on "WooCommerce->Home" Page

## Video:

https://user-images.githubusercontent.com/41110392/104608568-0ac35500-56a8-11eb-9495-fdad1fc039fb.mp4

- [ ] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [ ] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/storefront/).
- [x] I can reproduce this bug consistently using the steps above.

hsingyuc:
Fixes #6074 

Refactor panel with `withFocusOutside` and remove `react-click-outside` dependency.

### Screenshots
螢幕快照 2021-01-29 下午3 18 38

螢幕快照 2021-01-29 下午3 18 46

-   Go to WooCommerce home page
-   Click on Display and Help button back and forth, check that the popover and the panel close as expected.
-   Check that the setup store tab continues to work.


samueljseay (dismissed):
@hsingyuc thank you for this, I like the approach here.

There is just one issue which is that older versions of Wordpress by default do not come bundled with a late enough version of `@wordpress/compose` for us to use either of these new hooks (`__experimentalUseFocusOutside` ,`useFocusOnMount`). (I tested this on Wordpress 5.5.3 - make sure Gutenberg plugin is not installed)

This leaves us with a tough decision to make. We can either rely on compose directly or we must find another solution.

I'm not a big fan of adding compose to our bundle because we're trying very hard to move away from that approach, it increases the size of our JS payload.

I think maybe the best solution since both these hooks are small is to copy/paste them from gutenberg into our `lib` folder and use them directly. something like `client/lib/hooks` would be fine. Comments that they can be removed once they're supported in our minimum supported Wordpress version would be useful.

Let me know if you feel comfortable to do this. Thanks again for all your wonderful contributions!

hsingyuc:
@samueljseay  Thanks for the feedback! Could you check if this is what we need here? This required adding `memize` as a dependency. It's small so I think might be ok? https://bundlephobia.com/result?p=memize@1.1.0

There was a `hooks/` folder so I added them under that folder, if we need to move them under `lib/` please let me know.
Thank you!

client/hooks/useCallbackRef.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * External dependencies
+ */
+import memize from 'memize';
+import { useCallback } from '@wordpress/element';
+
+function useCallbackRef( callback, deps ) {
```
joshuatf, line 7: It looks like this was removed recently so we can remove this and copy the latest changes for `useFocusOnMount`

https://github.com/WordPress/gutenberg/blob/master/packages/compose/src/hooks/use-focus-on-mount/index.js

client/hooks/useFocusOnMount.js
```diff
@@ -0,0 +1,60 @@
+/**
```
joshuatf, line 1: Just need to add a quick comment here noting that this can be removed in the future:

```
/**
 * This hook was directly copied from https://github.com/WordPress/gutenberg/blob/master/packages/compose/src/hooks/use-focus-on-mount/index.js
 * to avoid its absence in older versions of WordPress.
 *
 * This can be removed once the minimum supported version of WordPress includes this hook.
 */
```

joshuatf (commented):
Thanks for the changes, @hsingyuc!  This is testing well with the hooks added in.  Just need to add the comment that @samueljseay and give this a rebase.

client/hooks/useFocusOnMount.js
hsingyuc: The PR is ready

joshuatf (approved):
Awesome, thanks for the quick fix!  Testing well and going to get this merged so we can iterate on the activity panel tabs permissions.

becdetat:
Cherry-picked into `release/2.0.2-beta.1`

PR #6234, opened 2021-01-29
Changing headings on report tables to use semantic tags

### LINKED ISSUE #6208: Headings structure on analytics reports pages present issues for keyboard navigation & screen-readers
**Describe the bug**
A common feature with screen-readers is the ability to navigate via the hierarchy of headings on the page. The way these are structured on the Reports pages may present some difficulty in navigation. 

One concrete example: the report tables do not include a heading at all, making them unreachable when navigating via this method. They do include elements that are clearing headings visually, but not semantically. This should be changed to be a semantic heading. 

**To Reproduce**

- Enable a screen reader that supports heading navigation (ie NVDA)
- Navigate to an analytics page with a data table
- Utilize heading navigation via the keyboard to explore the page, and notice that the data-table is unreachable
- View the source, and notice that the headings on the data table are not semantic heading elements

**Screenshots:**

image

**Expected behavior**
The headings on the report tables should be semantic heading elements

joelclimbsthings:
Fixes #6208 

The heading for the report tables currently uses a non-semantic `<p>` tag in the markup, which can cause issues with screen-readers and accessibility tools. This change simply alters that to an `<h2>`. 

### Screenshots

image

- Checkout branch
-   Use the inspector to verify that the pictured heading is using an `<h2>` tag, instead of a `<p>` tag.

### DIFF
--- a/client/analytics/components/leaderboard/test/__snapshots__/index.js.snap
+++ b/client/analytics/components/leaderboard/test/__snapshots__/index.js.snap
@@ -8,7 +8,7 @@ exports[`Leaderboard should render correct data in the table 1`] = `
     <div
       class="components-flex components-card__header is-size-medium e1q7k77g1 css-aafj2m-Flex-HeaderUI eboqfv50"
     >
-      <p
+      <h2
         class="css-1ahfdc3-Text e15wbhsk0"
       />
       <div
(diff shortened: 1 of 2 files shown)


psealock (approved):
Nice cleanup @joelclimbsthings :shipit:

PR #6247, opened 2021-02-02
Try: Refactor head and body heights

### LINKED ISSUE #6235: Navigation: Spike to investigate Header Height Solutions
Per our discussion in the sprint cycle planning call today, @joshuatf noted that we have a high number of issues ( I'll try to get them all linked here ) that involve visual glitches due to a variety of reasons in new navigation: narrow viewports, presence of mobile app dialogs etc.

Instead of trying to work on these independently, and possibly breaking work done by others to fix various bugs in header height, the agreed path forward was to add this issue to the development cycle for someone to investigate possible solutions to manage header height and related css.

We agreed to timebox this spike to 3 "days"

While I have labeled this navigation, some of the bugs impact classic navigation in wc-admin too.

Related Issues

- #6080
- #5980

joshuatf:
Fixes #6235 

Refactors header heights to be more dynamic.

Originally, I liked the idea of using some kind of global data store that contained information about the height/header.  However, this reduces benefits to non-React elements that may need context about the header.  In the end, I went with a more "natural" approach that assumes we only have one header and the body is pushed down according to its height.  In a nutshell:

* Any sticky (header) elements belong in the header.
* The WP body `#wpbody` is always pushed down by the dynamic height of this header.
* Body elements should not "guess" at how much padding is needed above them to compensate for header discrepancies.

Note that we could modify the margin to be added to the actual `body` instead of `#wpbody` if there are benefits to doing so (not seeing any yet).

This approach also favors removing the `wp-toolbar` class, however this can only be done via JS since it's always added when `is_admin()` is `true` so there is some extra CSS to modify this when that isn't possible.

### Screenshots

Screen Shot 2021-02-02 at 12 49 19 PM
Screen Shot 2021-02-02 at 12 49 12 PM
Screen Shot 2021-02-02 at 12 48 11 PM
Screen Shot 2021-02-02 at 12 47 46 PM

1. Test various embed and non-embed WC pages.
2. Test various screen sizes.
3. Test with the new navigation turned both on and off.
4. Test full screen pages (i.e., setup wizard).
5. Test with the mobile banner enabled.
6. Test with various combinations of the above.

### DIFF
--- a/client/header/index.js
+++ b/client/header/index.js
@@ -43,7 +43,11 @@ export const Header = ( { sections, isEmbedded = false, query } ) => {
 	useLayoutEffect( () => {
 		updateBodyMargin();
 		window.addEventListener( 'resize', updateBodyMargin );
-		return () => window.removeEventListener( 'resize', updateBodyMargin );
+		return () => {
+			window.removeEventListener( 'resize', updateBodyMargin );
+			const wpBody = document.querySelector( '#wpbody' );
+			wpBody.style.marginTop = null;
+		};
 	}, [ isModalDismissed ] );
 
 	useEffect( () => {
--- a/client/header/index.js
+++ b/client/header/index.js
@@ -24,21 +24,12 @@ export const Header = ( { sections, isEmbedded = false, query } ) => {
 	const pageTitle = sections.slice( -1 )[ 0 ];
 	const isScrolled = useIsScrolled();
 	const { updateUserPreferences, ...userData } = useUserPreferences();
-
 	const isModalDismissed = userData.android_app_banner_dismissed === 'yes';
+	let debounceTimer = null;
 
 	const className = classnames( 'woocommerce-layout__header', {
 		'is-scrolled': isScrolled,
 	} );
-
-	const updateBodyMargin = () => {
-		const wpBody = document.querySelector( '#wpbody' );
-		if ( ! wpBody ) {
-			return;
-		}
-
-		wpBody.style.marginTop = `${ headerElement.current.offsetHeight }px`;
-	};
 
 	useLayoutEffect( () => {
 		updateBodyMargin();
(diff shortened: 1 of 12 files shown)


psealock (commented):
I like the premise here @joshuatf with the dynamic `marginTop` added to `wpBody`.

There are some interactions noted in https://github.com/woocommerce/woocommerce-admin/pull/6289#pullrequestreview-587121270 that should be addressed either there or in this PR.

Otherwise, this is testing well so far

joshuatf:
@psealock ✅  and merged https://github.com/woocommerce/woocommerce-admin/pull/6289 into this branch so this is ready for review.

joelclimbsthings:
I really like this idea @joshuatf . Good work. 

I'm currently seeing some UI glitches in desktop mode, confirmed with both Firefox & Chrome:

image


Looks like there is extra whitespace due to this little guy being visible:

image

Comparing to main, that icon should only be visible on mobile when the nav is expanded.

client/header/index.js
```diff
@@ -30,6 +31,25 @@ export const Header = ( { sections, isEmbedded = false, query } ) => {
...
+		if ( ! wpBody ) {
+			return;
+		}
+
+		wpBody.style.marginTop = `${ headerElement.current.offsetHeight }px`;
+	};
+
+	useLayoutEffect( () => {
+		updateBodyMargin();
+		window.addEventListener( 'resize', updateBodyMargin );
```
joelclimbsthings, line 45: I wonder if we could benefit from some kind of debounce here on the resize event?

joelclimbsthings (changes requested):

client/header/index.js
joshuatf: Nice, I like this idea.  Will add it in.

joshuatf:
@joelclimbsthings I'm not able to see the errors on desktop you mentioned.  Do you have the latest version of this branch?

This CSS should make that "W" hidden:

Screen Shot 2021-02-11 at 3 41 27 PM

joelclimbsthings:
Yup @joshuatf , seems I do have the latest. Do you have any local changes or unpushed commits?

Looks like in my case that is being overridden:

image

image

client/header/index.js
joshuatf: Added this in https://github.com/woocommerce/woocommerce-admin/pull/6247/commits/52c0a128eb8a868deef21a8fac0cff6ae884cb1f - note I did not use lodash's `debounce` to avoid using `lodash` in case we remove this dependency in the future.

joshuatf:
I'm not able to reproduce the issue, but I'm guessing we're on different versions of WP and/or GB.  Made this selector more specific in https://github.com/woocommerce/woocommerce-admin/pull/6247/commits/741249e34c97bfbd252fc22bff25e127554a9525 - let me know if that resolves the issue for you.

joelclimbsthings:
Great @joshuatf , latest commits work well for me. As mentioned in slack, the rogue styles were related to AutomateWoo, but it was resolved with the greater specificity of you revised selector. 

So far tests well for me after clicking around a bit. The only thing I'd mention is that it _does not_ seem to fix this other issue which you flagged as a potential benefit: https://github.com/woocommerce/woocommerce-admin/issues/6248. We can just handle that in a separate PR.

joelclimbsthings (approved):

PR #6253, opened 2021-02-02
Separate secondary items into separate menu in navigation

### LINKED ISSUE #6150: Navigation: Add Sticky Footer
Settings, Maketplace, and Tools will now be in a "sticky" footer that remains as the user scrolls

Screen Shot 2021-01-22 at 1 23 48 PM

### Deliverables

* The Settings group (with Settings, Marketplace, and Tools links) is fixed to the bottom of the viewport
* A 1px border divides this section from the main navigation
* Check that mobile works too
* Marketplace is the first item (image above is outdated)
* This section remains fixed on scroll

joshuatf:
Fixes #6150

* Converts the navigation to a grid.
* Moves secondary items into a second menu.

My original attempt involved using absolutely positioning and padding to position the menu, however, this created issues with the menu overlapping at various screen sizes and broken in child menus.

This solution allows the secondary menu to grow or shrink in size and still have both menus work cohesively.

### Screenshots

Screen Shot 2021-02-02 at 5 20 42 PM
Screen Shot 2021-02-02 at 5 20 34 PM

1. Navigate through various navigation categories.
2. Make sure everything is visually as expected.
3. Make sure the back to dashboard button works as expected.
4. Shrink the viewport in height.
5. Make sure the secondary menu is visible, but the primary is scrollable.

### DIFF
--- a/client/navigation/components/container/index.js
+++ b/client/navigation/components/container/index.js
@@ -142,105 +142,97 @@ const Container = ( { menuItems } ) => {
 							secondary: secondaryItems,
 							plugins: pluginItems,
 						} = categorizedItems[ category.id ] || {};
-						return (
-							<>
-								{ ( !! primaryItems || !! pluginItems ) && (
-									<NavigationMenu
-										key={ category.id }
-										title={ category.title }
-										menu={ category.id }
-										parentMenu={ category.parent }
-										backButtonLabel={
-											isRootBackVisible
-												? rootBackLabel
-												: category.backButtonLabel ||
-												  null
-										}
-										onBackButtonClick={
-											isRootBackVisible
-												? () => {
-														trackBackClick(
-															'woocommerce'
-														);
-														window.location = rootBackUrl;
-												  }
-												: () =>
-														trackBackClick(
-															category.id
-														)
-										}
-									>
-										{ !! primaryItems && (
-											<NavigationGroup>
-												{ primaryItems.map(
-													( item ) => (
-														<Item
-															key={ item.id }
-															item={ item }
-														/>
+						return [
+							( !! primaryItems || !! pluginItems ) && (
+								<NavigationMenu
+									key={ category.id }
+									title={ category.title }
+									menu={ category.id }
+									parentMenu={ category.parent }
+									backButtonLabel={
+										isRootBackVisible
+											? rootBackLabel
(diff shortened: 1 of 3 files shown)


psealock (commented):
Nice, this is testing well and works without error. I'm seeing unique key warnings when navigating categories though

Screen Shot 2021-02-04 at 7 00 12 PM

joshuatf:
Good catch, @psealock!  Updated the keys for the secondary menu in https://github.com/woocommerce/woocommerce-admin/pull/6253/commits/c6920e48f2ae9e751dd21ab77de4cc32866e94e4

client/navigation/components/container/index.js
```diff
@@ -131,79 +136,110 @@ const Container = ( { menuItems } ) => {
...
-								{ !! pluginItems && (
-									<NavigationGroup
-										title={
-											category.id === 'woocommerce'
-												? __(
-														'Extensions',
-														'woocommerce-admin'
-												  )
-												: null
+							<>
```
louwie17, line 146: @joshuatf I still see the duplicate key error that @psealock saw. I believe it's because the key should technically be on this `<>` fragment element.
Instead of using the fragment you can also return an array, each condition as a separate item (basically just replace `<>` with `[`. Anyway that fixes it on my end.

louwie17 (changes requested):
Tested well, but still seeing the duplicate key error, see comment.

client/navigation/components/container/index.js
joshuatf: Hey @louwie17 - I did not see a duplicate key error, but I did see a separate error for the missing key that you mentioned here.  Updated this to use an array in https://github.com/woocommerce/woocommerce-admin/pull/6253/commits/bdcb03b25e46d3f53789e515f2dd6c073ac2d992.

louwie17: My bad, I guess I was referring to the missing key error. Thanks for updating this :)

louwie17 (approved):

louwie17:
Thanks for the update, no more errors in the console, and still testing well :) 🚀

client/navigation/components/container/index.js
joshuatf: No problem- just making sure we're looking at the same error.  Thanks for your help @louwie17!

PR #6260, opened 2021-02-03
Fixed the Add First Product email note checks

### LINKED ISSUE #6231: [1.9] Inbox: Email Notification Received for Adding First Product, when Products Exist
**Describe the bug**
This was reported in [a8c slack today](https://a8c.slack.com/archives/C3ECCGUVC/p1611946299031800) by @ChaosExAnima 

The site in question was auto-updated to `woocommerce-admin` 1.9.0, and after that, the email notification from introduced in https://github.com/woocommerce/woocommerce-admin/pull/6024 was triggered on the site. It seems there are established products, but they are in draft form, and as such the email notif was triggered.

The site is established, and has been running Woo for over 2 years, so thinking we need to make the logic on this notif a bit more restricted @octaedro / @pmcpinto - maybe have it only send to sites that are _between_ 3 and 30 days old maybe?

Another great idea from Echo: perhaps we should also check and make sure the site has zero sales to cover for the case where products get published and subsequently unpublished.

octaedro:
Fixes #6231

This PR adds two extra checks to the `Add First Product` email note and also modifies the image.
The extra checks are:
* The note won't be sent to stores older than 30 days.
* The note won't be sent to stores with orders.

### Accessibility

### Screenshots
screenshot-one wordpress test-2021 02 02-15_01_45

- Checkout this branch.
- Create a zip for testing with `npm run zip:test`.
- Update the installation date (we need a store between 3 and 30 days old). You can do it with an SQL statement like this:

```
UPDATE `wp_options` SET `option_value`=UNIX_TIMESTAMP(DATE_SUB(NOW(), INTERVAL 5 day)) WHERE `option_name` = 'woocommerce_admin_install_timestamp';
```

- Run the cron (this tool can help [WP Crontrol](https://wordpress.org/plugins/wp-crontrol/)).
- Verify the note's status is `sent`. You can use an SQL statement like this:
```
SELECT `status` FROM `wp_wc_admin_notes` WHERE `name` = 'wc-admin-add-first-product-note'
```
- Now delete the note with an SQL statement like:
```
DELETE FROM `wp_wc_admin_notes` WHERE `name` = 'wc-admin-add-first-product-note';
```
- Add a new order and run the cron.
- No note should have been added.
- Remove the order, add a product and run the cron.
- No note should have been added.
- Delete the product and modify the store creation date to 32 days with an SQL statement like:
```
UPDATE `wp_options` SET `option_value`=UNIX_TIMESTAMP(DATE_SUB(NOW(), INTERVAL 32 day)) WHERE `option_name` = 'woocommerce_admin_install_timestamp';
```
- No note should have been added.


src/Notes/AddFirstProduct.php
```diff
@@ -47,6 +47,16 @@ public static function get_note() {
 			return;
 		}
 
+		// Don't show if there is an orders.
+		$args   = array(
```
timmyc, line 51: I like that you kept the query here pretty "open" so it will capture the edge case of there being prior orders that may have been refunded, or canceled. Any order object presence would imply the site has setup a product at some point 👍

timmyc (approved):
Verified all uses cases work as expected - I used an email logger plugin, and verified that I only receive the email notif when no orders or products have been created, and the site is between 3 and 30 days old. Thanks @octaedro

timmyc:
@psealock @becdetat 

Not sure if we want to do a 1.9.1 for the 5.0 final - it looks like we might need to do one to address https://github.com/woocommerce/woocommerce-admin/issues/6155#issuecomment-772668352 and if so, thinking it might be nice to pick this bug fix in to 1.9.1. While edge case, it might save a few support requests.

Bec I definitely think we should pick this in for 2.0 though.

becdetat:
Cherry-picked into release/2.0.0-rc.2

PR #6228, opened 2021-01-29
Dependency Extraction Webpack Plugin: Add Woo Blocks packages

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/6025

A list of packages was being compiled at build from this repo. The resulting list did not include any packages from https://github.com/woocommerce/woocommerce-gutenberg-products-block so this PR exchanges the automation in favour of a hard coded list.

### Detailed test instructions:

1. `npm run build`
2. `npm run create-wc-extension` and give the plugin a name
3. `cd packages/dependency-extraction-webpack-plugin`
4. `npm link`
5. `cd ../../../<plugin name>`
6. `npm install`
7. `npm link @woocommerce/dependency-extraction-webpack-plugin`
8. `npm start`
9. In your plugins `scr/index.js` file add this code:
```js
import { flag } from '@woocommerce/components';
import { ShippingRatesControl } from '@woocommerce/blocks-checkout';
import { SITE_TITLE } from '@woocommerce/blocks-settings';

console.log( flag, ShippingRatesControl, SITE_TITLE );
```
10. Check the new plugin's `build/index.asset.php` file and check the array of dependencies to make sure its correct.
11. Visit `build/index.js` to make sure the aliases are correct. For example, see `@woocommerce/components` alias correctly to `window.wc.components`:
```js
/***/ "@woocommerce/components":
/*!*********************************************!*\
  !*** external {"this":["wc","components"]} ***!
  \*********************************************/
/*! no static exports found */
/***/ (function(module, exports) {

(function() { module.exports = this["wc"]["components"]; }());
```

PR #6268, opened 2021-02-04
Dependency Extraction Webpack Plugin: bump version to 1.2

psealock:
This PR bumps Dependency Extraction Webpack Plugin to 1.2 in preparation to make a release.

This is the only change since last release https://github.com/woocommerce/woocommerce-admin/pull/6228

### DIFF
--- a/packages/dependency-extraction-webpack-plugin/CHANGELOG.md
+++ b/packages/dependency-extraction-webpack-plugin/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 1.2.0
+
+-   Add WooCommerce Blocks Dependencies.
+
 # 1.1.0
 
 -   Fix: Handle irregular package names that don't conform to a pattern.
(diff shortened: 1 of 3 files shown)


packages/dependency-extraction-webpack-plugin/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 1.2.0
+
+-   Add WooCommerce Blocks Dependencies.
```
louwie17, line 3: Given you referenced the PR change for this in the comments #6228 do we want to add that to the changelog here?

louwie17 (approved):
LGTM, just left one comment about referencing the PR number in the changelog?

packages/dependency-extraction-webpack-plugin/CHANGELOG.md
psealock: Good call! Done in https://github.com/woocommerce/woocommerce-admin/pull/6268/commits/e5c001d6ae7f50bb715ce9b61c81c4d14461047b

Issue #6272, opened 2021-02-04
Make import_sample_products test more robust


louwie17:
The `test_import_sample_products` test randomly fails, which is especially noticeable with the new tests run in Github actions.
The current theory is that it fails because of data left behind from previous tests.
When it fails it imported 1 or 0 products, and it should be importing close to 20 every time, maybe the Importer should be mocked?

Listed as a cooldown issue as it is low priority.

PR #6273, opened 2021-02-04
Skip intermittent failing test

louwie17:
Skips the `test_import_sample_products` which intermittently fails on line 77, where the imported products is less then 2.

Created a follow up cooldown issue for this: https://github.com/woocommerce/woocommerce-admin/issues/6272


samueljseay (approved):
Thanks for doing this

Issue #6136, opened 2021-01-21
Favorite extensions rest endpoint


joshuatf:
Blocked by https://github.com/woocommerce/woocommerce-admin/issues/6131

We need the ability to read, update, and delete favorited extensions via REST API for the new navigation features.

| Endpoint | Params | Method |
| --- | ----------- | ----------- |
| `navigation/favorites` (read all) | `plugin_name` | `GET`
| `navigation/favorites` (create) | `plugin_name` | `POST`
| `navigation/features` (delete) | `plugin_name` | `DELETE`

This issue will involve creating these 3 endpoints.

joshuatf:
Estimate: 3

PR #6275, opened 2021-02-04
Add navigation favorites data store

### LINKED ISSUE #6266: Add favorites to navigation data store
To interact with the REST API in https://github.com/woocommerce/woocommerce-admin/issues/6136 we should add `favorites` to the data store.

* `getFavorites()`
* `addFavorite()`
* `setFavorites()`
* `removeFavorite()`

joshuatf:
Fixes #6266 

Blocked by https://github.com/woocommerce/woocommerce-admin/issues/6136

This PR sets up the data store for favorites and all methods needed by the navigation to interact with favorites.

Note that the data store semantics and conventions are a slight departure from previous conventions used based on p2 discussions around data stores becoming more event focused.  Up for debate on the naming here, but I do find being "event-focused" to make a lot more sense.

1.  Make sure tests pass.
2. Check follow-up PR that connects to the favorites UI. https://github.com/woocommerce/woocommerce-admin/pull/6287
3. Make sure new naming conventions make sense (discussion).


packages/data/src/navigation/reducer.js
```diff
@@ -29,6 +32,100 @@ const reducer = (
...
+				},
+			};
+			break;
+		case TYPES.ADD_FAVORITE_FAILURE:
+			state = {
+				...state,
+				error,
+				requesting: {
+					...state.requesting,
+					addTodo: false,
```
joelclimbsthings, line 69: Thinkin' this should be `addFavorite`?

packages/data/src/navigation/reducer.js
```diff
@@ -29,6 +32,100 @@ const reducer = (
...
+			state = {
+				...state,
+				requesting: {
+					...state.requesting,
+					addTodo: true,
+				},
+			};
+			break;
+		case TYPES.ADD_FAVORITE_SUCCESS:
+			const newFavorites = ! state.favorites.includes( favorite )
```
joelclimbsthings, line 83: This check is also performed in the API logic, so perhaps unnecessary?

packages/data/src/navigation/reducer.js
```diff
@@ -29,6 +32,100 @@ const reducer = (
...
+			state = {
+				...state,
+				requesting: {
+					...state.requesting,
+					removeFavorite: true,
+				},
+			};
+			break;
+		case TYPES.REMOVE_FAVORITE_SUCCESS:
+			const filteredFavorites = state.favorites.filter(
```
joelclimbsthings, line 116: Currently returning the revised (remaining) favorites from this request, so we could just use the response from the API here.

packages/data/src/navigation/reducer.js
joelclimbsthings: Also, same as my other comment; this request currently returns all favorites (including added) on success, so we could just directly plug in the response.

joelclimbsthings (changes requested):
Hey @joshuatf ; looks great! Made a few comments, a couple of which are subject to change depending on feedback on https://github.com/woocommerce/woocommerce-admin/pull/6282

packages/data/src/navigation/reducer.js
joshuatf: Using all favorites is a possibility, though I'm hesitant to remove the logic around whether or not it's included client-side if we add a single item.

There are still potential for side effects in the state that would make it wise to safe guard this.

joshuatf:
Thanks for the early feedback here, @joelclimbsthings!  Let's discuss in https://github.com/woocommerce/woocommerce-admin/pull/6282 before tackling this one.

joshuatf:
Hey @joelclimbsthings this has been rebased and hooked up to the new REST API and is ready for review.

packages/data/src/navigation/actions.js
```diff
@@ -23,3 +29,118 @@ export function addMenuItems( menuItems ) {
...
+	return {
+		type: TYPES.REMOVE_FAVORITE_SUCCESS,
+		favorite,
+		error,
+	};
+}
+
+export function* addFavorite( favorite ) {
+	yield addFavoriteRequest( favorite );
+	yield addFavoriteSuccess( favorite );
```
joelclimbsthings, line 101: I'm not sure why we're calling `addFavoriteSuccess()` before the request here?

joelclimbsthings (changes requested):
Hey @joshuatf , looks great! Tests ran successfully. I had one quick question regarding the `addFavoriteSuccess()` calls. I haven't tested this in the context of the follow-up PR yet (and I have a doc appointment in a few minutes), but can do so a bit later.

We can also just defer that to a review on that PR as well.

joelclimbsthings:
Also, it looks like you're currently having the same issue with the test suite that I was. It's related to the recent release of WC 5 (and potentially a missing docker image?), but was resolved by @samueljseay and @adrianduffell for me.

packages/data/src/navigation/actions.js
joshuatf: Good catch, must have been left over from testing.  Removed in https://github.com/woocommerce/woocommerce-admin/pull/6275/commits/0f15a74afe5a2369c244c53bea6815e13cf1fb7b

joelclimbsthings (approved):
Well done @joshuatf . :ship:

PR #6278, opened 2021-02-04
Skip extensions in navigation flyout menu

### LINKED ISSUE #6276: Navigation: All extensions included in Flyout menu
The flyout menu includes all Extensions. Ultimately, only the ones that are favorited will end up there. Since the flyout goes out with 2.0 before favorites, we should remove all extensions.

Screen Shot 2021-02-05 at 12 01 15 PM

This value could filter out items whose `menuId` is `plugin`

https://github.com/woocommerce/woocommerce-admin/blob/c39aac66b4396fde45fae8fa557ae8c0099daeaa/src/Features/Navigation/CoreMenu.php#L287

joshuatf:
Fixes #6276

Skips extensions in the flyout menu until we have favorites in place.

#### Before
Screen Shot 2021-02-04 at 6 13 15 PM

#### After
Screen Shot 2021-02-04 at 6 12 13 PM

1. Add a plugin that registers a plugin menu item (or register your own via `add_plugin_item()`.
2. Navigate to a non-WC page.
3. Hover the "WooCommerce" menu item.
4. Make sure no extension items are shown.

### DIFF
--- a/src/Features/Navigation/CoreMenu.php
+++ b/src/Features/Navigation/CoreMenu.php
@@ -294,6 +294,11 @@ class CoreMenu {
 		}
 
 		foreach( $top_level_items as $item ) {
+			// Skip extensions.
+			if ( $item['menuId'] === 'plugins' ) {
+				continue;
+			}
+
 			// Skip specific categories.
 			if (
 				in_array(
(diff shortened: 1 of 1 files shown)


psealock (approved):
Works like a charm :shipit:

PR #6281, opened 2021-02-05
Make sure that industry is defined in payment methods

louwie17:
I ran into an instance where `profileItems.industry` was undefined (I skipped the OBW), this caused an error as it tried to run the function `some` on undefined.
This was caused by some recent refactoring, replacing the `some` lodash method with `profileItems.industry.some`.

- Start a new store, and skip the initial onboarding flow, there is a button `Skip store details` at the bottom
- Load the `Set up payments` task, the payment options should load correctly.

### DIFF
--- a/client/task-list/tasks/payments/methods.js
+++ b/client/task-list/tasks/payments/methods.js
@@ -79,9 +79,11 @@ export function getPaymentMethods( {
 		wcPayIsConnected = false,
 	} = onboardingStatus;
 
-	const hasCbdIndustry = profileItems.industry.some( ( { slug } ) => {
-		return slug === 'cbd-other-hemp-derived-products';
-	} );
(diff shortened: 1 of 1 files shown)


samueljseay (approved):
This is all good, but I think at some point we need to fix the redux store to have proper non-null default empty states so we don't have to keep doing this kind of defensive programming.

Thanks for spotting this 🚀

louwie17:
> This is all good, but I think at some point we need to fix the redux store to have proper non-null default empty states so we don't have to keep doing this kind of defensive programming.

@samueljseay I agree with you here, although this instance actually comes from the `OnboardingProfile.php` file, where it is being set to `null` so things would need to be updated a couple more levels down.

becdetat:
Cherry-picked into release/2.0.0-rc.2

PR #6270, opened 2021-02-04
Adding Favorite API Class to Navigation Feature

joelclimbsthings:
Fixes #6131 

Adding the Favorites API to the navigation feature:

```php
Favorites::add_plugin($plugin );
Favorites::remove_plugin($plugin );
Favorites::get_all();
```

This simply manages an array of keys within `update_user_meta` to track the favorites for the user.

### Detailed test instructions:

-   Checkout branch
-   Run `\Automattic\WooCommerce\Admin\Features\Navigation\Favorite::add_plugin('whoa_buddy');`
-  Check `woocommerce_admin_navigation_favorites` under `wp_usermeta`, which should have content `["whoa_buddy"]`
- Run `\Automattic\WooCommerce\Admin\Features\Navigation\Favorite::add_plugin('yo_buddy');`
-  Check `woocommerce_admin_navigation_favorites` under `wp_usermeta`, which should have content `["whoa_buddy", "yo_buddy"]`
- Run `\Automattic\WooCommerce\Admin\Features\Navigation\Favorite::remove_plugin('whoa_buddy');`
-  Check `woocommerce_admin_navigation_favorites` under `wp_usermeta`, which should have content `["yo_buddy"]`

PR #6282, opened 2021-02-05
Adding Navigation Favorites REST API Endpoints

### LINKED ISSUE #6136: Favorite extensions rest endpoint

joelclimbsthings:
Fixes #6136 
Adding simple CRUD API to manage Navigation Favorites:

| Endpoint | Params | Method |
| --- | ----------- | ----------- |
| `navigation/favorites` (read all) | `user_id` | `GET`
| `navigation/favorites` (create) | `item_id`, `user_id` | `POST`
| `navigation/features` (delete) | `item_id`, `user_id` | `DELETE`

All endpoints accept the `user_id` optional parameter to specify user for favorites operations, and it will default to the current user when this is not provided. 

### Questions
- Are permissions handled correctly here? Trying to pattern after a few others I saw in the APIs, but not entirely sure I understand all the conventions, or what is appropriate for this use case.
- Experimenting with a little exception handling, although I haven't done this in PHP before, so let me know if anything looks amiss (or if we'd rather not throw exceptions at all).

### To-do
- Adding tests (doing so now)

_Note:_ This is currently based off of the unmerged `add/6131-api-favorites-registration` branch.

-   Checkout branch
-   Use _Postman_ or other tools to send requests to `/wp-json/wc-admin/navigation/favorites`, using the `GET`, `POST`, and `DELETE` methods. Ensure the request is authenticated with the nonce and cookie headers.
- Trying adding an item, listing them, and deleting. 
- Ensure no unhandled errors or unexpected responses.
- Then add the `user_id` parameter, and run through the specified methods again. 
- Follow along with the `woocommerce_admin_navigation_favorites` in the `wp_usermeta` table to ensure the correct user/etc is being altered.


src/Features/Navigation/Favorites.php
```diff
@@ -0,0 +1,115 @@
...
+		$user = $user_id ? $user_id : get_current_user_id();
+
+		if ( ! $user || ! $item_id ) {
+			throw new \Exception( 'invalid_input' );
+		}
+
+		$all_favorites = self::get_all( $user );
+
+		if ( ! in_array( $item_id, $all_favorites, true ) ) {
+			throw new \Exception( 'does_not_exist' );
```
joshuatf, line 88: This works well, but I have a feeling a thrown exception here may be overkill and may result in a number of accidental exceptions being thrown (by us when the state cannot be fully known or 3PDs not wrapping in a try/catch).

Related to [this comment](https://github.com/woocommerce/woocommerce-admin/pull/6270#discussion_r571450283) most WP functions favor returning a mix of `WP_Error` and `true|false` depending on success.

src/Features/Navigation/Favorites.php
```diff
@@ -0,0 +1,115 @@
...
+		}
+
+		$remaining = array_diff( $all_favorites, [ $item_id ] );
+
+		self::set_meta_value( $user, array_values( $remaining ) );
+
+		return $remaining;
```
joshuatf, line 95: Related to [your comment here](https://github.com/woocommerce/woocommerce-admin/pull/6275/files#r571321029) - one of the potential (edge case) issues here is that due to the nature of async requests, we could fire off 2 remove requests.  If they complete out of order, we'd replace the state with an outdated version of remaining favorites.

src/API/NavigationFavorites.php
```diff
@@ -0,0 +1,206 @@
...
+	 */
+	protected $rest_base = 'navigation/favorites';
+
+	/**
+	 * Register the routes
+	 */
+	public function register_routes() {
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base,
```
joshuatf, line 41: I think these are good endpoints, but defaulting to the current user I think breaks convention a bit here.

I know it's a few extra lines of code now, but I think it's worthwhile to add 3 more endpoints with `/me` suffixes to all of these endpoints that defaults to the current user.

src/API/NavigationFavorites.php
```diff
@@ -0,0 +1,206 @@
...
+
+
+	/**
+	 * Check whether a given request has permission to create favorites.
+	 *
+	 * @param  WP_REST_Request $request Full details about the request.
+	 * @return WP_Error|boolean
+	 */
+	public function add_item_permissions_check( $request ) {
+		if ( ! wc_rest_check_manager_permissions( 'settings', 'edit' ) ) {
```
joshuatf, line 185: Capability check here should probably be `current_user_can( 'edit_users' )` for the default routes and no permission check for the `/me` routes if we add those in.  I believe all users are allowed to edit their own profile settings.

src/API/NavigationFavorites.php
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	/**
+	 * Check whether a given request has permission to delete notes.
+	 *
+	 * @param  WP_REST_Request $request Full details about the request.
+	 * @return WP_Error|boolean
+	 */
+	public function delete_item_permissions_check( $request ) {
+		if ( ! wc_rest_check_manager_permissions( 'settings', 'edit' ) ) {
```
joshuatf, line 199: Same permission comments here.

src/API/NavigationFavorites.php
```diff
@@ -0,0 +1,206 @@
...
+		try {
+			$altered_favorites = Favorites::remove_item( $fav_id, $user_id );
+		} catch ( \Exception $e ) {
+			$error_msg = $e->getMessage();
+
+			if ( 'invalid_input' === $error_msg ) {
+				return new \WP_Error(
+					'woocommerce_favorites_invalid_request',
+					__( 'Sorry, invalid request', 'woocommerce-admin' ),
+					array( 'status' => 400 )
```
joshuatf, line 161: Nice use of error codes! 👏

src/API/NavigationFavorites.php
```diff
@@ -0,0 +1,206 @@
...
+				);
+			}
+
+			if ( 'already_exists' === $error_msg ) {
+				return new \WP_Error(
+					'woocommerce_favorites_already_exists',
+					__( 'Favorite already exists', 'woocommerce-admin' ),
+					array( 'status' => 409 )
```
joshuatf, line 134: 💯

src/API/NavigationFavorites.php
```diff
@@ -0,0 +1,206 @@
...
+			if ( 'invalid_input' === $error_msg ) {
+				return new \WP_Error(
+					'woocommerce_favorites_invalid_request',
+					__( 'Sorry, invalid request', 'woocommerce-admin' ),
+					array( 'status' => 400 )
+				);
+			}
+
+			if ( 'already_exists' === $error_msg ) {
+				return new \WP_Error(
```
joshuatf, line 131: Just thinking out loud here, I wonder if it might make sense to move this returned error to the `add_item()` method and then we can simply return any `WP_Error` instances.

joshuatf (commented):
Really great work here, @joelclimbsthings!  Love the well thought out error code responses.

Left a few comments, mostly around WordPress convention.

src/API/NavigationFavorites.php
joelclimbsthings: Thanks Josh, good call. I've added the `/me` endpoints and changed these to require the `user_id` parameter. 

One curiosity is that as it is now the `/me` endpoints will still technically accept and utilize the `user_id` parameter since they're using the same handler function. They're just not required; is that considered a bad thing? I can break those out, although it seems elegant to be able to utilize the same handler if it's not a concern.

joelclimbsthings: It actually occurred to me that this could be a security flaw as is, addressing now.

src/Features/Navigation/Favorites.php
joelclimbsthings: Ah yes, good call.

src/API/NavigationFavorites.php
joelclimbsthings: Agreed, done.

joelclimbsthings:
Thanks for the thorough review @joshuatf . I've implemented all of the requested feedback. 

I'm not 100% certain of my current approach to disallow `user_id` in the `/me` routes via this commit (https://github.com/woocommerce/woocommerce-admin/pull/6282/commits/2835ecf1d900cb62e3f33e064052f7090bde34f0). Easy enough to break these up if you think that's a better approach.

src/API/NavigationFavorites.php
```diff
@@ -0,0 +1,239 @@
...
+	 * @var string
+	 */
+	protected $rest_base = 'navigation/favorites';
+
+	/**
+	 * Error code to status code mapping.
+	 *
+	 * @var array
+	 */
+	protected $error_to_status_map = array(
```
joshuatf, line 40: Nice ❤️

src/API/NavigationFavorites.php
joshuatf: I would probably break them out to separate functions, but we can reuse the primary function.  For example:

```
get_items( $request ) {
  // original logic to get items by given user id
}

get_items_by_current_user( $request ) {
  $user_id = get_current_user_id();
  // If user not logged in, return error
  $request->set_param( 'user_id', $user_id );
  get_items( $request );
}

**Review by joshuatf — COMMENTED** [review]
Looking great, Joel!  Left a response to your question about the `/me` endpoint methods.

File: src/API/NavigationFavorites.php (hunk 3 above)
Comment (joelclimbsthings) [reply src/API/NavigationFavorites.php:52]: I agree, that's a bit clearer. Done.

**Comment by joelclimbsthings** [comment]
Thanks @joshuatf . Refactored those functions (I think I was having a bit of function fatigue yesterday). Feeling good about it now. Going to get back to trying to get the PHP test suite working so I can add those, but this should be good to review.

File: src/API/NavigationFavorites.php [hunk 10]
```diff
@@ -0,0 +1,299 @@
...
+	}
+
+	/**
+	 * Add a favorite.
+	 *
+	 * @param WP_REST_Request $request Request data.
+	 * @return WP_REST_Response
+	 */
+	public function add_item( $request ) {
+		$user_id = $request->get_param( 'user_id' );
```
joshuatf, line 184: Might be nice to check if this user (ID) exists before adding data.  Currently responds with an empty array for getting from a fake user ID and will allow me to add to random user IDs.

I'm okay tackling this in a follow-up too.

joshuatf (approved):
Awesome work, @joelclimbsthings!  Thanks for all the changes.  Everything is working really well.

Left a couple quick changes, but pre-approving.

joshuatf:
>Going to get back to trying to get the PHP test suite working so I can add those

Nice.  My vote is to get this PR merged so we can unblock and follow-up with some tests.  I just spoke with @psealock about adding more tests.  I'm hoping there's time next week to go back and add more tests around all the new nav work as we've been pretty lax.

joelclimbsthings:
This is ready to merge, although we have a single failing test in CI that isn't failing locally. Trying to figure out, but I'm at EOD and it may wait until tomorrow.

PR #6285, opened 2021-02-08
Add tracks event for WooCommerce Payments install via tasklist and inbox note

chickenn00dle:
This adds a tracks event fired after WooCommerce Payments is installed
via the tasklist.

See p1612278892009000-slack-C0144BMA170 for more details.

1. With a new store, go initiate the payment task in the onboarding task list.
2. Select the Set up button for WooCommerce Payments.
3. After 5-10 minutes check tracks live view for a `wcadmin_woocommerce_payments_install` event with the context prop of `tasklist`.

Note this event is also implemented in https://github.com/woocommerce/woocommerce/pull/29052 with the context of `extensions`

### Changelog

> Tweak: Track the number of installations of WooCommerce Payments via that task list for stores that have opted into tracking.

### DIFF
--- a/src/Notes/WooCommercePayments.php
+++ b/src/Notes/WooCommercePayments.php
@@ -149,6 +149,8 @@ class WooCommercePayments {
 			return false;
 		}
 
+		wc_admin_record_tracks_event( 'woocommerce_payments_install', array( 'context' => 'inbox' ) );
+
 		$activate_request = array( 'plugins' => self::PLUGIN_SLUG );
 		$result           = $installer->activate_plugins( $activate_request );
 		if ( is_wp_error( $result ) ) {
--- a/client/task-list/tasks/payments/methods.js
+++ b/client/task-list/tasks/payments/methods.js
@@ -10,6 +10,7 @@ import {
 } from '@woocommerce/wc-admin-settings';
 import { Link } from '@woocommerce/components';
 import { WC_ADMIN_NAMESPACE } from '@woocommerce/data';
+import { recordEvent } from '@woocommerce/tracks';
 
 /**
  * Internal dependencies
@@ -58,7 +59,13 @@ export function installActivateAndConnectWcpay(
 	};
 
 	installAndActivatePlugins( [ 'woocommerce-payments' ] )
-		.then( () => connect() )
+		.then( () => {
+			recordEvent( 'woocommerce_payments_install', {
+				context: 'tasklist',
+			} );
+
+			connect();
+		} )
 		.catch( ( error ) => {
 			createNoticesFromResponse( error );
 			reject();


chickenn00dle:
I just realized there is an inbox notice that also nudges merchants to install WCPay. I've added a commit for that, this time with a context of `inbox`: https://github.com/woocommerce/woocommerce-admin/pull/6285/commits/e46a8ff83b2d254b3bf28d5fb13a45878a1d4c54

src/Notes/WooCommercePayments.php
```diff
@@ -149,6 +149,8 @@ private function install_and_activate_wcpay() {
 			return false;
 		}
 
+		wc_admin_record_tracks_event( 'woocommerce_payments_install', array( 'context' => 'inbox' ) );
+
```
vbelolapotkov, line 153: This action is used for remote inbox notifications as well [here](9024-gh-Automattic/woocommerce.com/files#diff-cb53b37c568c7a244ecd7e8b35c8a5831d1c1c1e7d7f99f006d4b20fe2036113R407) for promos. Unfortunately I don't see a good way of differentiating between both, except like adding extra query param to remote inbox note and checking it in `install_on_action` method below.

chickenn00dle: Hey @vbelolapotkov 

The link you provided takes me to an empty diff, so I'm not sure what you are referring to here.

Do you mean that `install_and_activate_wcpay()` or `install_on_action()` are used elsewhere in the codebase for some other purpose? I did a quick search and didn't find any other usage for these.

vbelolapotkov: Sorry @chickenn00dle , here is [better link](https://github.com/Automattic/woocommerce.com/blob/trunk/plugins/wccom-plugins/inbox-notifications/includes/notifications.json.php#L168). The idea here is that remote inbox notifications use the same action link leading to execution of `install_on_action`. In that case remote inbox click will be tracked the same way as inbox click. AFAIK there are no active promos at the moment, so it shouldn't cause any issues now, but I think we'd like to differentiate between remote and local inbox clicks in the future.

We might solve this by adding some query param to remote inbox action. E.g. `promo=<promo-name>` or `remote=<remote-inbox-slug>` and check it existence, e.g. like this:
```
private function install_and_activate_wcpay( $context = 'inbox' ) {
		$install_request = array( 'plugins' => self::PLUGIN_SLUG );
		$installer       = new \Automattic\WooCommerce\Admin\API\Plugins();
		$result          = $installer->install_plugins( $install_request );
		if ( is_wp_error( $result ) ) {
			return false;
		}

		wc_admin_record_tracks_event( 'woocommerce_payments_install', array( 'context' => $context ) );
...
      }
...

// TODO: Need to validate this request more strictly since we're taking install actions directly?
		/* phpcs:disable WordPress.Security.NonceVerification */
		if (
			! isset( $_GET['page'] ) ||
			'wc-admin' !== $_GET['page'] ||
			! isset( $_GET['action'] ) ||
			'setup-woocommerce-payments' !== $_GET['action']
		) {
			return;
		}
		/* phpcs:enable */
		if ( ! current_user_can( 'install_plugins' ) ) {
			return;
		}

		$context = 'inbox';
                if ( isset( $_GET['remote'] ) ) {
                         $context = $context . '-' .  $_GET['remote']; // plus need to sanitize query param value.
                }
		$this->install_and_activate_wcpay( $context ); // so if we set remote param in inbox action as `promo-2021-mar` the $context will be `inbox-promo-2021-mar`.
```

@jrodger what do you think here?

adrianduffell: @jrodger bump 🙂

adrianduffell:
@chickenn00dle We're releasing our 2.1 beta on Monday. This PR looks possible to make it into the release if we can merge it before then.

src/Notes/WooCommercePayments.php
jrodger: Thanks for the pings folks, sorry for missing those (especially this close to a release), I'm not sure what happened there.

I agree with your points @vbelolapotkov, but I'm wary of putting work into this when we might run promotions differently in the future. The way we've done things here isn't ideal (For one, @chickenn00dle had no way to know we were using / abusing the functionality in this way).

The remote inbox promotion you've referenced is still running, but I think we can safely end the experiment now. I'll double check this and get it ended if that's how we want to proceed.

So my suggestion would be:

* Ignore the remote inbox promotion, we'll end it as soon as we can.
* Ship the code you have in this PR and ignore the promotion use case.
* Bear in mind that some clicks coming through might have come from the promotion while we get that ended and the notice disappears from merchant's inboxes (hopefully that's OK?).
* When we run another promotion, maybe we'll have a better way of installing the plugin. But if not we consider the tracking at that point.

Does that sound OK to people?

vbelolapotkov (approved):
:shipit:

PR #6286, opened 2021-02-08
Fix the react state update error on homescreen

louwie17:
Fixes the React Warning on dev - `Cannon update component while rendering a different component` (see screenshot)

Screen Shot 2021-02-05 at 5 21 51 PM

This was mostly happening because of creating new object references in the `useSelect` causing for many component updates. Which allowed for a high probability for a state to be updated at the same time a render took place.

This was mostly fixable by a small refactor, I did add the `rememo` library to the `data` package, this allows us to memoize selectors, especially handy for the query selector that maps the id's to the objects.
`wordpress/core-data` does the same thing, see [here](https://github.com/WordPress/gutenberg/blob/master/packages/core-data/src/selectors.js#L99-L106) that's where I got the library from.

- Create a new Woocommerce store and finish the onboarding wizard
- Load the home screen in develop, there should be no red warnings in the console (like the screen shot above)
- Click on `Set up payments` task, no new warnings should pop up (I did have one randomly show up, but couldn't reproduce it after).
- Do a quick smoke test of the home screen in general, make sure the notes render correctly, and the unread indicator shows up (when needed), can dismiss and revert (card correctly disappears), store alerts


client/header/activity-panel/unread-indicators.js
```diff
@@ -13,6 +13,15 @@ import { getSetting } from '@woocommerce/wc-admin-settings';
...
 import { getUnreadNotesCount } from '../../inbox-panel/utils';
 
+const UNREAD_NOTES_QUERY = {
+	page: 1,
+	per_page: QUERY_DEFAULTS.pageSize,
+	status: 'unactioned',
+	type: QUERY_DEFAULTS.noteTypes,
+	orderby: 'date',
+	order: 'desc',
+};
```
louwie17, line 23: Moved this out, so that the variable `query` does not contain a new object reference every time (done in a couple other places as well).

client/layout/store-alerts/index.js
```diff
@@ -275,20 +281,21 @@ export class StoreAlerts extends Component {
...
-		const alertsQuery = {
-			page: 1,
-			per_page: QUERY_DEFAULTS.pageSize,
-			type: 'error,update',
-			status: 'unactioned',
-		};
 
 		// Filter out notes that may have been marked actioned or not delayed after the initial request
-		const filterNotes = ( note ) => note.status === 'unactioned';
-		const alerts = getNotes( alertsQuery ).filter( filterNotes );
```
louwie17, line 290: Moved the filtering to the component instead.

client/layout/store-alerts/test/index.js
```diff
@@ -13,11 +13,13 @@ const alerts = [
 	{
 		title: 'Alert title 1',
 		content: 'Alert content 1',
+		status: 'unactioned',
```
louwie17, line 16: Although we filtered in the `useSelect` by `unactioned` we didn't pass `unactioned` alerts into our tests. This fixes the tests.

client/layout/transient-notices/index.js
```diff
@@ -24,7 +30,7 @@ class TransientNotices extends Component {
 
 		return (
 			<SnackbarList
-				notices={ notices }
+				notices={ notices.concat( notices2 ) }
```
louwie17, line 33: Do the `concat` in the component instead of the `useSelect` as concat creates a new object reference.

packages/data/package.json
```diff
@@ -23,7 +23,8 @@
 	"dependencies": {
 		"@babel/runtime-corejs2": "7.12.5",
 		"@woocommerce/date": "2.1.0",
-		"@woocommerce/navigation": "5.2.0"
+		"@woocommerce/navigation": "5.2.0",
+		"rememo": "^3.0.0"
```
louwie17, line 27: Added `rememo` as mentioned in the PR description (same as what wordpress/core-data uses).

packages/data/src/notes/selectors.js
```diff
@@ -1,7 +1,18 @@
...
+export const getNotes = createSelector(
+	( state, query ) => {
+		const noteIds = state.noteQueries[ JSON.stringify( query ) ] || [];
+		return noteIds.map( ( id ) => state.notes[ id ] );
+	},
+	( state, query ) => [
+		state.noteQueries[ JSON.stringify( query ) ],
+		state.notes,
+	]
+);
```
louwie17, line 15: Make use of the `createSelector` function, this memoizes the return value.
The selector only gets updated when the second parameter function is updated (line 12 & 13). 
In this case the selector is run when:
- the `noteQueries` array is updated for the specific `query`
- When the `notes` state is updated, needed for when a note is dismissed (deleted).

When the above two are not updated it will return the last cached result, which would contain the same object reference if that is what it is returning. (see the [wordpress/core-data](https://github.com/WordPress/gutenberg/blob/master/packages/core-data/src) library for more examples.

samueljseay: I like this a lot. Redux is the perfect layer to do this kind of biz-logic caching and means we can just forget about it at the component layer. Good stuff!

client/layout/transient-notices/index.js
samueljseay: Nice one and good spotting. My only concern is that others are not going to pick up why this is done this way in future (unless they know the nuances of render perf in React), so I'm contemplating whether a comment would be good here maybe? I don't mind if you don't because I don't like those kind of comments a whole lot, but it may help someone.

client/layout/transient-notices/index.js
```diff
@@ -15,7 +15,13 @@ import './style.scss';
 
 class TransientNotices extends Component {
 	render() {
-		const { className, notices, onRemove, onRemove2 } = this.props;
+		const {
+			className,
+			notices,
+			notices2 = [],
+			onRemove,
```
samueljseay, line 22: This seems real silly... we should just wrap these in a single callback and pass it down. I realise this is just the way it was done previously so no biggy if you don't fix. I also realise this will all (hopefully) go away when notes are consolidated to a single store.

client/header/activity-panel/unread-indicators.js
```diff
@@ -31,21 +40,15 @@ export function getUnreadNotes( select ) {
 
 	// @todo This method would be more performant if we ask only for 1 item per page with status "unactioned".
```
samueljseay, line 41: These comments are irrelevant now or they should be moved next to the `const`. I prefer just removing todos that will likely never be addressed but up to you how you tackle this.

client/header/activity-panel/unread-indicators.js
```diff
@@ -31,21 +40,15 @@ export function getUnreadNotes( select ) {
...
-		order: 'desc',
-	};
 
 	// Disable eslint rule requiring `latestNotes` to be defined below because the next two statements
 	// depend on `getNotes` to have been called.
 	// eslint-disable-next-line @wordpress/no-unused-vars-before-return
-	const latestNotes = getNotes( notesQuery );
-	const isError = Boolean( getNotesError( 'getNotes', [ notesQuery ] ) );
-	const isRequesting = isResolving( 'getNotes', [ notesQuery ] );
+	const latestNotes = getNotes( UNREAD_NOTES_QUERY );
```
samueljseay, line 47: This is weird, why don't we just call getNotes without assigning it to a variable and the lint exception can be removed? Again I know not your code, up to you if you fix.

samueljseay (approved):
I've given some suggestions but pre-approving and up to you what you fix. There was just one issue with a comment that needs to be moved or deleted. 

🚀

client/header/activity-panel/unread-indicators.js
louwie17: Did basically as you suggested, calling `getNotes` stand alone first, and then also later when needed for the `getUnreadNotesCount`.

Issue #6213, opened 2021-01-28
Favoriting extensions from the new navigation


pmcpinto:
**User story:**
As a merchant
I want to favorite key extensions
So I can easily access them when I'm using WooCommerce

**Description:** we're going to move forward with just one favoriting method - from the extension title in the new nav.

![home.png](https://images.zenhubusercontent.com/5d84fe8e1201de0001b6ab61/81a403fa-3152-482e-801d-6f12d73cdfc1)

![favoriting-in-the-nav.gif](https://images.zenhubusercontent.com/5d84fe8e1201de0001b6ab61/40854f6e-e239-4fe1-a2a8-13ad3cef549c)

**Acceptance criteria:**
- [x] Upon clicking into the extension menu, a star icon is visible at the top to the right of the extension title, allowing users to favorite that specific extension https://github.com/woocommerce/woocommerce-admin/pull/6287
- [x] Selecting the star icon will change it from outlined to filled, and the extension item will appear in the top level of the nav https://github.com/woocommerce/woocommerce-admin/pull/6290
- [x] The favorited extension appears in the flyout menu from the wp-admin dashboard https://github.com/woocommerce/woocommerce-admin/pull/6290
- [ ] Display a tooltip about the new feature, when the user visits an extension menu for the first time 
- [x] When clicking in a favorited extension through the flyout menu, the user is redirected to the ~~page that 3PDs have identified the top-level page~~ first item under that extension's category https://github.com/woocommerce/woocommerce-admin/pull/6290
- [x] The favorite extensions should be stored as a user preference and not site-wide for all store managers https://github.com/woocommerce/woocommerce-admin/pull/6270
- [ ] Users can favorite a specific page (nice to have - requires more exploration)

**Event tracking:**
Event name: `wcadmin_navigation_favorite`
Event prop: `extension_name`
Event prop: `action` - favorite | unfavorite
Event description: triggered when the a user clicks to favorite or unfavorite an extension

pmcpinto:
@elizaan36 pls check if I'm not missing anything

joshuatf:
@elizaan36 @pmcpinto Does this mean:

1. We're really favoriting categories/menu items?  An extension could have multiple top-level categories if it chooses.
2. If the extension does not have a category, but a single top-level menu item, can it be favorited from the root of the menu?  How would this icon appear?

elizaan36:
> We're really favoriting categories/menu items? An extension could have multiple top-level categories if it chooses.

[...]

PR #6287, opened 2021-02-08
Add favoriting extensions client UI

joshuatf:
Fixes (part of) #6213 

Adds the UI to favorite extensions from the navigation menu.  Note that the colors still need to be updated here (missing Figma design links).

We may also want to consider optimistically updating our data store (up for discussion).

### Screenshots
Screen Shot 2021-02-08 at 11 53 00 AM

1. Install an extension that registers a category.
2. Navigate to that extension's menu.
3. Toggle the favorite on/off and make sure it works as expected.
4. (Assuming #6136 is hooked up) - make sure selections persist on refresh.

### DIFF
--- a/client/navigation/components/category-title/index.js
+++ b/client/navigation/components/category-title/index.js
@@ -1,6 +1,7 @@
 /**
  * External dependencies
  */
+import { __ } from '@wordpress/i18n';
 import { Button } from '@wordpress/components';
 import { NAVIGATION_STORE_NAME } from '@woocommerce/data';
 import { recordEvent } from '@woocommerce/tracks';
@@ -45,7 +46,17 @@ export const CategoryTitle = ( { category } ) => {
 					isTertiary
 					onClick={ toggleFavorite }
 					icon={ isFavorited ? 'star-filled' : 'star-empty' }
-					aria-label={ 'Add this extension to your favorites.' }
+					aria-label={
+						isFavorited
+							? __(
+									'Add this item to your favorites.',
+									'woocommerce-admin'
+							  )
+							: __(
+									'Remove this item from your favorites.',
+									'woocommerce-admin'
+							  )
+					}
 				/>
 			</span>
 		);
--- a/client/navigation/components/category-title/index.js
+++ b/client/navigation/components/category-title/index.js
@@ -8,6 +8,7 @@ import { useDispatch, useSelect } from '@wordpress/data';
 /**
  * Internal dependencies
  */
+import './style.scss';
 
 export const CategoryTitle = ( { category } ) => {
 	const { id, title } = category;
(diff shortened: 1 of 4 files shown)


client/navigation/components/container/style.scss
```diff
@@ -1,6 +1,6 @@
 .woocommerce-navigation {
-	h2 {
-		color: inherit; // This should be handled at component level.
```
joshuatf, line 3: cc @psealock To make sure this doesn't break anything.  I didn't notice any changes after removing this, but wanted to double check.

psealock: I think this was left over from Gutenberg Nav being in a state of flux. We should try it with latest Gutenberg activated and just using the version on WP 5.6 to double check. Otherwise, good to remove

client/navigation/components/category-title/index.js
```diff
@@ -0,0 +1,68 @@
...
+	if ( category.menuId === 'plugins' ) {
+		return (
+			<span className={ className }>
+				<span className={ `${ className }__text` }>{ title }</span>
+				<Button
+					className={ `${ className }__favorite-button` }
+					isTertiary
+					onClick={ toggleFavorite }
+					icon={ isFavorited ? 'star-filled' : 'star-empty' }
+					aria-label={
```
joelclimbsthings, line 49: :heart: 

One entirely optional enhancement here; you can add the `role="checkbox"` and `aria-checked={ isFavorited ? 'true' : 'false' }` to convey to the screen-reader users that it's a togglable state. I'm honestly not certain if this is much of an enhancement, since it seems pretty clear when you add the dynamic `aria-label` as you have here.

joelclimbsthings (approved):
Looks great, left one totally optional comment on the aria stuff. 

One thing I saw that I thought was curious (although not an issue it seems) is that dev tools for both FF & Chrome show the `DELETE` request as another `POST`:

image

But it _is_ indeed a DELETE request and it's hitting the correct endpoint. Just thought that was odd; do you know why that's the case?

client/navigation/components/category-title/index.js
joshuatf: Good thinking on the checkbox.  I'm not sure it matches with the existing Favorite/Unfavorite labeling since the label would then imply the inverse of the checked status (checked, label: Unfavorite).

I'll leave this for a follow-up so we can merge this one and investigate the best way to handle accessibility concerns.

joshuatf:
Thanks for the review, Joel!  Good comments on the accessibility, I'll open an issue so we can follow-up.

>One thing I saw that I thought was curious (although not an issue it seems) is that dev tools for both FF & Chrome show the DELETE request as another POST:

I believe the `apiFetch` handles this by adding `X-HTTP-Method-Override` to the request.  And now thinking about this, I'm wondering if we should update our delete paths to be slightly more RESTful.  `DELETE => favorites/me/{item_id}`.  Will open another issue to address.

PR #6290, opened 2021-02-08
Move favorited menu items to primary menu

joshuatf:
Fixes (part of) #6213 

Moves menu items when favoriting or unfavoriting via the client-side or when favorite exists in user meta.

### Screenshots
Screen Shot 2021-02-08 at 3 01 55 PM
Screen Shot 2021-02-08 at 3 37 24 PM

1. Install and activate a plugin that has registered a plugin menu category.
2. Navigate to that category.
3. Toggle on/off the favorite.
4. Navigate back to the root menu and check that the category is under the correct menu.
5. (Assuming REST API is merged) check that changes persist on page refresh.
6. Add an extension to your favorites.
6. Navigate to a non-WC page.
7. Check that the item appears in the flyout menu.

### DIFF
--- a/client/navigation/components/category-title/index.js
+++ b/client/navigation/components/category-title/index.js
@@ -37,7 +37,7 @@ export const CategoryTitle = ( { category } ) => {
 		} );
 	};
 
-	if ( category.menuId === 'plugins' ) {
+	if ( [ 'plugins', 'favorites' ].includes( category.menuId ) ) {
 		return (
 			<span className={ className }>
 				<span className={ `${ className }__text` }>{ title }</span>
--- a/client/navigation/components/category-title/index.js
+++ b/client/navigation/components/category-title/index.js
@@ -15,9 +15,12 @@ import './style.scss';
 export const CategoryTitle = ( { category } ) => {
 	const { id, title } = category;
 
-	const { favorites } = useSelect( ( select ) => {
+	const { favorites, isResolving } = useSelect( ( select ) => {
 		return {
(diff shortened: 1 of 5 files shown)


psealock (commented):
This is looking good and testing well. I have two small UI nitpicks which could be handled here or in follow ups

1. When un-favoriting an extension, it gets appended to the bottom of the list but comes back to the correct order on page refresh. Any way we can maintain extension order? It may mean passing order to client for sorting.
2. When I visit an extension screen which I have favorited, the star shows un-favorited until the API returns with favorite data. This confused me and I think it would be better to hide the star altogether until favorite data is available.

joshuatf:
Thanks for the review @psealock.  Agreed on both of your points.

>When un-favoriting an extension, it gets appended to the bottom of the list but comes back to the correct order on page refresh. Any way we can maintain extension order? It may mean passing order to client for sorting.

Let's address in a follow-up if that's okay.  I think if we're going this route, we may consider removing all sorting/ordering on the server in favor of client-side only sorting.

>When I visit an extension screen which I have favorited, the star shows un-favorited until the API returns with favorite data. This confused me and I think it would be better to hide the star altogether until favorite data is available.

Hid this until favorites has finished resolving.

psealock (approved):
>Let's address in a follow-up if that's okay. I think if we're going this route, we may consider removing all sorting/ordering on the server in favor of client-side only sorting.

Sounds good to me. Hiding of the star until data arrives looks good now, thanks

PR #6294, opened 2021-02-09
Navigation: Return $menu when no child items exist

psealock:
Roles without capabilities to edit WooCommerce had no child items to migrate. This caused wp-admin sidebar to fatal:

Screen Shot 2021-02-09 at 2 46 58 PM

Reproduce the error on main:
1. Create a new user with "Author" role
2. Login as that user and visit any wp-admin page
3. See the error

See the fix
4. Switch to this branch and see the sidebar load correctly


becdetat (approved):
Looks good. Works as described on the tin.

PR #6295, opened 2021-02-09
Fixes for @woocommerce/dependency-extraction-webpack-plugin

nerrad:
While working on the Stripe cart and checkout blocks integration migration to the Stripe extension, I tried using the `@woocommerce/dependency-extraction-webpack-plugin` (including the latest 1.2.0 release) but it didn't work because it was compiling various `@woocommerce/*` aliases to the wrong scope resulting in something like this in the built file:

```js
/***/ "@woocommerce/blocks-registry":
/*!*********************************************!*\
  !*** external {"this":["wc","wcBlocksRegistry"]} ***!
  \*********************************************/
/*! no static exports found */
/***/ (function(module, exports) {
(function() { module.exports = window[undefined]; }());
```

The above was generated with this webpack config:

```js
const defaultConfig = require( '@wordpress/scripts/config/webpack.config' );
const DependencyExtractionWebpackPlugin = require( '@woocommerce/dependency-extraction-webpack-plugin' );

module.exports = {
	...defaultConfig,
	plugins: [
		...defaultConfig.plugins.filter(
			( plugin ) =>
				plugin.constructor.name !== 'DependencyExtractionWebpackPlugin'
		),
		new DependencyExtractionWebpackPlugin(),
	],
};
```

Besides this issue, `block-settings` shouldn't really be exposed as an external and instead we should expose `@woocommerce/settings` as `@woocommerce/block-settings` is really just something we have specific to the blocks project builds. The "official" interface we should expose for "settings" coming from the server is on the `wc.wcSettings` global which in turn exposes the [`getSetting` function](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/settings/shared/get-setting.js) thus encouraging _read only_ usage of values from the global. So I've made the correct adjustments for that too.

## To Test

* [x] It doesn't look like wc-admin is using this plugin in your webpack config so no impact there.
* [x] I tested this with the Stripe integration work I'm doing (by just directly copying changes into the node_modules install of the Webpack plugin, and the files are built correctly with these changes and the above webpack configuration.

### DIFF
--- a/bin/starter-pack/_package.json
+++ b/bin/starter-pack/_package.json
@@ -22,6 +22,6 @@
 	"devDependencies": {
 		"@wordpress/scripts": "^12.2.1",
 		"@woocommerce/eslint-plugin": "1.1.0",
-		"@woocommerce/dependency-extraction-webpack-plugin": "1.2.0"
+		"@woocommerce/dependency-extraction-webpack-plugin": "1.3.0"
 	}
 }
--- a/packages/dependency-extraction-webpack-plugin/CHANGELOG.md
+++ b/packages/dependency-extraction-webpack-plugin/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 1.3.0
+
+-  Remove `@woocommerce/block-settings` from internal maps and use `@woocommerce/settings` instead. This external exposes the `getSetting` API interface which encourages read only use of the global.
+-  Remove explicitly scoping externals to `this`. The plugin compiler will take care of scoping the external to the correct context and `this` is not correct in some contexts consuming this package.
 # 1.2.0
 
 -   Add WooCommerce Blocks Dependencies. #6228
@@ -8,7 +12,7 @@
 
 # 1.0.1
 
--   Fix: Avoid tanspiling packaged code.
+-   Fix: Avoid transpiling packaged code.
 
 # 1.0.0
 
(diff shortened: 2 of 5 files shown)


psealock (approved):
Changes makes sense, thanks for fixing @nerrad. I updated the version called for in `wc-create-extension` script as well in https://github.com/woocommerce/woocommerce-admin/pull/6295/commits/d49cc66f5a499a282b3334ff05d9af5990c846b8

nerrad:
@psealock I'm not sure what happened here but it doesn't look like the published package has the `settings` property in the built files. While I do see the `settings` property in the source:

Image 2021-02-10 at 6 58 09 AM

If you look in the `build-module` or `build` directories of the package:

Image 2021-02-10 at 6 58 49 AM

So, `@woocommerce/settings` doesn't get properly handled by extractions for consumers of the package.

nerrad:
I also noticed that [this line](https://github.com/woocommerce/woocommerce-admin/blob/ff4d88fbe156b3cfe42586e86e8a8ad35aabab13/packages/dependency-extraction-webpack-plugin/assets/packages.js#L19) didn't get updated 🤦 .

psealock:
Darn! Sorry about that fail getting through my code review. I've been wondering if it should be explicit to any package, ie `@woocommerce/*`.

In any case, here is the follow up ready for review https://github.com/woocommerce/woocommerce-admin/pull/6331

PR #6309, opened 2021-02-10
Update highlight tooltip to make use of anchor logic

louwie17:
Make use of the tooltip `anchorRect` property, allowing us to add the tooltip to the `body` instead of adding it as a child element.

### Screenshots

Screen Shot 2021-02-10 at 10 24 21 AM

- Load a new Woocommerce store and finish the onboarding wizard
- Visit a task from the home screen task list, but don't finish it
- Go back to the homescreen, and visit that same task again
- A popup should show up pointing to the help menu in the top right (this should look and work correctly)

### DIFF
--- a/client/header/activity-panel/highlight-tooltip/index.js
+++ b/client/header/activity-panel/highlight-tooltip/index.js
@@ -12,7 +12,12 @@ import {
 	Button,
 	IsolatedEventContainer,
 } from '@wordpress/components';
-import { useState, useEffect, createPortal } from '@wordpress/element';
+import {
+	useState,
+	useEffect,
+	createPortal,
+	useLayoutEffect,
+} from '@wordpress/element';
 import { close } from '@wordpress/icons';
 import { noop } from 'lodash';
 
@@ -31,22 +36,33 @@ function HighlightTooltip( {
 	onClose,
 	delay,
 	onShow = noop,
+	useAnchor = false,
 } ) {
 	const [ showHighlight, setShowHighlight ] = useState(
 		delay > 0 ? null : show
 	);
 	const [ node, setNode ] = useState( null );
+	const [ anchorRect, setAnchorRect ] = useState( null );
 
 	useEffect( () => {
 		const element = document.getElementById( id );
 		let container;
 		if ( element && ! node ) {
 			// Add tooltip container
-			const parent = element.parentElement;
-			container = document.createElement( 'div' );
-			container.classList.add( 'highlight-tooltip__container' );
-			parent.appendChild( container );
-			setNode( container );
+			if ( ! useAnchor ) {
+				const parent = element.parentElement;
+				container = document.createElement( 'div' );
+				container.classList.add( 'highlight-tooltip__container' );
+				parent.appendChild( container );
+				setNode( container );
+			} else {
+				const parentElement = document.createElement( 'div' );
+				document.body.appendChild( parentElement );
+				container = document.createElement( 'div' );
+				parentElement.appendChild( container );
+				setNode( container );
+				setAnchorRect( element.getBoundingClientRect() );
+			}
 		}
 		const timeoutId = showTooltip( container );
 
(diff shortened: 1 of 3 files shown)


client/header/activity-panel/highlight-tooltip/index.js
```diff
@@ -31,22 +36,33 @@ function HighlightTooltip( {
...
+				container.classList.add( 'highlight-tooltip__container' );
+				parent.appendChild( container );
+				setNode( container );
+			} else {
+				const parentElement = document.createElement( 'div' );
+				document.body.appendChild( parentElement );
+				container = document.createElement( 'div' );
+				parentElement.appendChild( container );
+				setNode( container );
+				setAnchorRect( element.getBoundingClientRect() );
```
louwie17, line 64: When enabling `useAnchor` we append the tooltip to a div tag in the `body` and pass the `anchorRect` of the id referenced to the Popover element.

client/header/activity-panel/highlight-tooltip/index.js
```diff
@@ -78,6 +94,18 @@ function HighlightTooltip( {
 		}
 	}, [ show ] );
 
+	useLayoutEffect( () => {
+		window.addEventListener( 'resize', updateSize );
+		return () => window.removeEventListener( 'resize', updateSize );
```
louwie17, line 99: Make sure the popover updates on window resize.

client/header/activity-panel/highlight-tooltip/index.js
joshuatf: This works perfectly for our use case!

The only thing I notice is that sometimes the position is not correct on the first load (presumably the targeted element is still mounting).  One possible workaround is updating this when `showHighlight` changes.

```js
	useEffect( () => {
		setAnchorRect( element.getBoundingClientRect() );
	}, [ showHighlight ] );
```

joshuatf (commented):
Thanks so much for this quick update, @louwie17!  This is testing well with one minor change.

I was also curious why we opted to use `id` instead of passing a `ref` here.  I'm guessing because the tabs are looped over and this would require creating refs for all the activity panel tabs- is that right?

louwie17:
> I was also curious why we opted to use `id` instead of passing a `ref` here. I'm guessing because the tabs are looped over and this would require creating refs for all the activity panel tabs- is that right?

The reason for `id` was so that it could be triggered from pretty well anywhere within the code, if using `ref` I would have to make sure we trigger this in a similar place as the target element so that we have access to the `ref`.
I actually tried to mimic the [wp-pointer](https://github.com/woocommerce/woocommerce/blob/master/includes/admin/class-wc-admin-pointers.php#L229) that are shown when `&tutorial=true` when creating a new product (try it out by going to `/wp-admin/post-new.php?post_type=product&tutorial=true`). Here they actually allow for someone to step through them (from one to the other).
If we ever wanted to do something like that with our dashboard, using an `id` would be very beneficial, instead of adding the component in each specific section.
Hope that explanation makes sense :) That being said feel free to add support for `ref` if that works better in the Nav.

> The only thing I notice is that sometimes the position is not correct on the first load (presumably the targeted element is still mounting). One possible workaround is updating this when showHighlight changes.

I will check this out, does this happen in our current use case or something that happens when you test it out for the navigation?

joshuatf:
>The reason for id was so that it could be triggered from pretty well anywhere within the code

Gotcha, that makes sense 👍 

>I will check this out, does this happen in our current use case or something that happens when you test it out for the navigation?

Only for the navigation currently.  I'll get a PR out within the hour so we can poke around there.  It could be an issue specific to the nav.

louwie17:
> And resizing naturally fixes the position which leads me to believe the element is still loading on initial position calculation.

Thanks I did a small refactor here: https://github.com/woocommerce/woocommerce-admin/pull/6309/commits/236c18633de34c46cc1485dbb80918bd05e643a3 hopefully that fixes it :)

joshuatf (approved):
Fantastic.  I think the positioning was an error on my end in navigation and was targeting a hidden element with the same ID.

Refactor looks great though ✅  Really appreciate you taking the time to add in this update.

PR #6317, opened 2021-02-10
Javascript error regarding <StoreAlerts> component breaking Homescreen

### LINKED ISSUE #6316: Javascript error regarding <StoreAlerts> component breaking Homescreen
**Describe the bug**
An error is being thrown on the Homescreen (as well as other React-based pages) that are prevent the app from displaying.

```
Uncaught (in promise) TypeError: alert is undefined
```

This is related to the `<StoreAlerts />` component, and is only reproducible when an alert is present on the store.

**Expected behavior**

**Screenshots**

image

joelclimbsthings:
Fixes #6316 

An error was preventing Homescreen from rendering when any admin alerts were present: 
`Uncaught (in promise) TypeError: alert is undefined`

The variable `currentIndex` was being initialized to null when `alerts` was a valid array, but empty. This PR resolves that by initializing `currentIndex` to 0 whenever `alerts` is a valid array.

### Screenshots

image

- Checkout latest main 
- This is only reproducible when you have an unactioned admin note. 
-  Open the `wp_wc_admin_notes` within your store DB, and change any note with a type of `update` to be "_unactioned_" under status. In my case I did this with the `wc-update-db-reminder` note.
- Navigate to the Homescreen
- You should see the error shown in screenshots.
- Now checkout this branch, the error should disappear.

### DIFF
--- a/client/layout/store-alerts/index.js
+++ b/client/layout/store-alerts/index.js
@@ -37,7 +37,7 @@ export class StoreAlerts extends Component {
 		const alerts = this.getAlerts();
 
 		this.state = {
-			currentIndex: alerts.length > 0 ? 0 : null,
+			currentIndex: alerts instanceof Array ? 0 : null,
 		};
 
 		this.previousAlert = this.previousAlert.bind( this );


louwie17:
Thanks for pushing up a fix for this @joelclimbsthings, my bad for not testing this more thoroughly :/ 
I end up merging #6320 in favour of this change (hope you won't hate me for this 😄 ), as setting the index to `0` by default made more sense.

PR #6318, opened 2021-02-11
Stalebot: Update label exclusion list.

timmyc:
After chatting with @pmcpinto, we would like to add all `feature: *` issue labels to the list of labels to exclude from the stalebot's logic.

The reasoning behind this change is that the bot has been closing a number of issues that we want to keep active/open in our backlog, many of which were tagged with feature labels.

Not entirely sure how to test this out :)

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -19,5 +19,5 @@ jobs:
                   close-pr-message: 'This PR was automatically closed due to being stale.'
                   stale-pr-label: 'status: stale'
                   stale-issue-label: 'status: stale'
-                  exempt-issue-labels: 'cooldown period,priority: high,priority: critical'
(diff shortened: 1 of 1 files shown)


samueljseay:
@timmyc this change is all good with me, just checking though, are you sure you don't want to introduce a single new label that we can add to those features that we are driving and just exempt that? Something like "feature work"? I proposed that solution to @pmcpinto and he was in support I think.

One benefit of that is we don't have to do a PR anytime a new feature is introduced.

timmyc:
> Something like "feature work"? I proposed that solution to @pmcpinto and he was in support I think.

We did talk about that too, I think Pedro's concern with that was we would forget to add that label also. I definitely see the concern around the new feature area labels in the future though.

samueljseay (approved):
> We did talk about that too, I think Pedro's concern with that was we would forget to add that label also

we can automate this in future, but for now I'm fine with this interim solution. 🚀

PR #6310, opened 2021-02-10
Move admin menu manipulation from admin_head to admin_menu

mattsherman:
**_Note: This PR should be tested in conjunction with https://github.com/woocommerce/woocommerce/pull/29088_**

### Changes proposed in this Pull Request:

This PR move WooCommerce Admin's admin menu manipulation from `admin_head` hooks to `admin_menu` hooks. This, along with a corresponding fix in `woocommerce/woocommerce`, addresses the issues seen when the WooCommerce admin menu is displayed in Calypso when WordPress.com's Nav Unification is enabled (currently only for Automatticians). See pbIJXs-Ee-p2 for more details.

Specifically, the following issues are fixed when using Calypso with WordPress.com's Nav Unification enabled:

- The WooCommerce submenu item gets replaced by the Home submenu item, and clicking on it correctly shows the WooCommerce Admin home screen
- The Orders count appears

When accessing wp-admin, there should be no change to the WooCommerce admin menus.

#### Before

Before admin_menu

#### After

After admin_menu

Note: The following issues will be addressed in separate PRs:

- Products menu item position: May be fixed by either https://github.com/woocommerce/woocommerce/pull/28763 or https://github.com/Automattic/jetpack/pull/18758
- WooCommerce and Products menu icons: Issue https://github.com/Automattic/wp-calypso/issues/49145; may be fixed by switching to SVG icons (issue with that: https://github.com/Automattic/wp-calypso/issues/49145)

#### General wp-admin testing without Nav Unification

1. Checkout the corresponding `woocommerce/woocommerce` branch (see https://github.com/woocommerce/woocommerce/pull/29088).
2. Checkout this branch.
3. Access wp-admin.
4. Verify that all WooCommerce admin menus appear and function the same as before.

#### Testing Nav Unification

Make your test site act like an "Atomic" site:

5. Connect Jetpack on your test site to WordPress.com.
6. Clone `wpcomsh` repo (see https://github.com/Automattic/wpcomsh for details)
7. Checkout the `add/admin-menu-request-impersonation` branch (577-gh-Automattic/wpcomsh)
8. Add a plugin to `mu-plugins` with the following line, to enable Nav Unification:

```
add_filter( 'jetpack_load_admin_menu_class', '__return_true' );
```

9. Access your connected site from WordPress.com (Calypso) and verify that the menu items appear correctly (excluding issues described above)

### Changelog entry

> Enter a summary of all changes on this Pull Request. This will appear in the changelog if accepted.

[...]

PR #6319, opened 2021-02-11
Navigation: Migrate methods to `admin_menu` hook

psealock:
https://github.com/woocommerce/woocommerce-admin/pull/6310 moves Nav logic to `admin_menu` action so that Calypso can correctly gather correct Sidebar items. As such, this breaks WC Navigation unless the same changes are made in Nav logic. Those changes are this PR.

1. Checkout the branch from this Core pull request https://github.com/woocommerce/woocommerce/pull/29088
2. Turn on Navigation
3. Check the flyout menu that it is correct
4. Enter WooCommerce and check that items are all correct
5. Checkout WC Core release branch [release/5.0](https://github.com/woocommerce/woocommerce/tree/release/5.0) and repeat steps 1 - 4.

### Bonus Testing

Note: I haven't done this yet because my testing env got deleted.

1. Fire up an eComm testing environment with the Nav Unification branch of wpcomsh found in https://github.com/Automattic/wpcomsh/pull/577#issuecomment-769527303
2. Use the branch from the Core pull request https://github.com/woocommerce/woocommerce/pull/29088
3. Visit the Calypso side of things and make sure the Nav Unification and new WC Nav are working as expected.

### Known Issue

The number of orders is not reflected in the flyout menu

Screen Shot 2021-02-11 at 2 14 54 PM

### DIFF
--- a/src/Features/Navigation/Menu.php
+++ b/src/Features/Navigation/Menu.php
@@ -523,13 +523,6 @@ class Menu {
 
 		foreach ( $submenu_items as $key => $menu_item ) {
 			if ( in_array( $menu_item[2], CoreMenu::get_excluded_items(), true ) ) {
-				// phpcs:disable
-				if ( ! isset( $menu_item[ self::CSS_CLASSES ] ) ) {
-					$submenu['woocommerce'][ $key ][] .= ' hide-if-js';
-				} else {
-					$submenu['woocommerce'][ $key ][ self::CSS_CLASSES ] .= ' hide-if-js';
-				}
-				// phpcs:enable
 				continue;
 			}
 
@@ -612,7 +605,11 @@ class Menu {
 
 		foreach ( $menu as $key => $menu_item ) {
 			if ( self::has_callback( $menu_item ) ) {
+				// Disable phpcs since we need to override submenu classes.
+				// Note that `phpcs:ignore WordPress.Variables.GlobalVariables.OverrideProhibited` does not work to disable this check.
+				// phpcs:disable
 				$menu[ $key ][ self::CSS_CLASSES ] .= ' hide-if-js';
+				// phps:enable
 				continue;
 			}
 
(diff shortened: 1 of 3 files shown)


becdetat (approved):
Looks good

becdetat:
I'm merging this, #6310 has been merged and it looks like Matt's [WooCommerce PR](https://github.com/woocommerce/woocommerce/pull/29088#pullrequestreview-588173496) will be merged before code-freeze as well. So this is required for 2.0.0, which I am about to cut.

becdetat:
Cherry-picked to `release/2.0.0`

src/Features/Navigation/Menu.php
```diff
@@ -88,8 +88,9 @@ final public static function instance() {
 	public function init() {
 		add_action( 'admin_menu', array( $this, 'add_core_items' ) );
 		add_filter( 'admin_enqueue_scripts', array( $this, 'enqueue_data' ), 20 );
-		add_filter( 'add_menu_classes', array( $this, 'migrate_menu_items' ), 30 );
-		add_filter( 'add_menu_classes', array( $this, 'migrate_core_child_items' ) );
+
+		add_filter( 'admin_menu', array( $this, 'migrate_core_child_items' ) );
```
joshuatf, line 92: This may fire too early.  Any plugins not using the default priority will be missed.  I think this should probably be `PHP_INT_MAX - 1` and the below `migrate_menu_items` should have a priority of `PHP_INT_MAX - 2`.

This doesn't guarantee that we get all plugins and I think that was part of the issue.  IIRC a plugin did manage to sneak through in testing which was our reason for moving to `admin_menu_classes`.  This may be more edge case though so I can support the move to `admin_menu` provided we tweak the priorities a bit.

joshuatf (commented):
This works pretty well for the nav itself, however the changes here have broken all core settings pages for me.

We'll need to follow up on this since its already been merged.

PR #6325, opened 2021-02-11
Build: Update Jetpack Autoloader dependency.

timmyc:
This branch is a follow-up issue to https://github.com/woocommerce/woocommerce/pull/29102

It seems our JP Autoloader dependency, and how it is declared, is preventing us from shipping v2.0.0 in the 5.1 release branch of Woo Core. [From a conversation in slack](https://a8c.slack.com/archives/C8X6Q7XQU/p1613064060259000) the agreed upon path forward was for packages ( wc-admin and blocks ) to update the JP Autloader version to use the caret notation `^2.9.1`

After this is tested, approved and merged, we will need to cut a `release/2.0.1` branch with this change. 

cc @becdetat

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
 	"require": {
 		"composer/installers": "^1.9.0",
 		"php": ">=7.0",
-		"automattic/jetpack-autoloader": "2.7.1"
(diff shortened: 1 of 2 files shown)


psealock (approved):
Changes look good and in line with what was discussed

Issue #6315, opened 2021-02-10
New navigation is broken when user cannot view reports


joshuatf:
**Describe the bug**
When a user does not have the `view_woocommerce_reports` capability, but can `manage_woocommerce` and `edit_shop_orders`, the navigation is broken on the WooCommerce pages they can view.

This is probably due to `view_woocommerce_reports` being somehow tied to all WooCommerce Admin pages/features.

**To Reproduce**

1. Enable the new navigation.
2. Create a user with the above mentioned capabilities. https://wordpress.org/plugins/leira-roles/
3. Visit the orders page `wp-admin/edit.php?post_type=shop_order`
4. Note the broken nav.

**Expected behavior**
Expected the new navigation to still work on these pages.

**Screenshots**
<img width="760" alt="Screen Shot 2021-02-10 at 6 18 25 PM" src="https://user-images.githubusercontent.com/10561050/107585534-6cee7780-6bcc-11eb-98fa-39ecf570eeba.png">

PR #6329, opened 2021-02-11
Remove categories without menu items

### LINKED ISSUE #6314: Navigation categories shown when no child items exist
**Describe the bug**
When no child items exist, a menu category is still shown.

**To Reproduce**

1. Enable the new navigation.
2. Create a user role that has `manage_woocommerce` capabilities and `view_woocommerce_reports` but not `edit_shop_orders`.  You can use https://wordpress.org/plugins/leira-roles/ to create the roles/caps
3. Note that when visiting a WC page (like `/wp-admin/admin.php?page=wc-admin`) that `Orders >` exists even though no child items exist.

**Expected behavior**
Expected categories with no children to be hidden.

**Screenshots**
Screen Shot 2021-02-10 at 6 09 50 PM

joshuatf:
Fixes #6314

* Removes categories without menu items
* Removes capabilities on categories and leaves responsibility to menu items.  This one is up for discussion, but my thinking here is that this could be confusing to have an item a user could access, but they cannot see due to category restrictions.

### Screenshots

Screen Shot 2021-02-11 at 5 20 44 PM

1. Load up a dev branch of Gutenberg and activate (needed for new `hideIfTargetMenuEmpty` prop).
2. Create a user and role with capabilities that can see some items, but not all.
3. `return true;` if you have trouble seeing the nav in `user_can_analytics()`.  Issue - https://github.com/woocommerce/woocommerce-admin/issues/6315
4. Note that categories without items are hidden.


psealock (approved):
Changes make sense, including removing capabilities from categories. I tested this by adding non existing capabilities to a plugin and testing with Shop Manager role. Its working as expected.

PR #6330, opened 2021-02-12
Bump minimum WP version to 5.4

psealock:
We have an L2 support, which is current + last two releases. Now that WP 5.6 is out, we can bump the minimum version to 5.4.


samueljseay (approved):
🚀

PR #6331, opened 2021-02-12
Dependency Extraction Webpack Plugin: Add @woocommerce/settings to list of packages

psealock:
Discovered in https://github.com/woocommerce/woocommerce-admin/pull/6295#issuecomment-776659446, `@woocommerce/settings` was not in the list of accepted packages.

See https://github.com/woocommerce/woocommerce-admin/pull/6295#issuecomment-776659446 and make sure the changes make sense, as well as the version bump.

### DIFF
--- a/bin/starter-pack/_package.json
+++ b/bin/starter-pack/_package.json
@@ -22,6 +22,6 @@
 	"devDependencies": {
 		"@wordpress/scripts": "^12.2.1",
 		"@woocommerce/eslint-plugin": "1.1.0",
-		"@woocommerce/dependency-extraction-webpack-plugin": "1.3.0"
+		"@woocommerce/dependency-extraction-webpack-plugin": "1.4.0"
 	}
 }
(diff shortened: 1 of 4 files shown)


nerrad (approved):
Change looks great! Thanks for fixing.

I'm not sure if this will also resolve the missing `settings` property in the `build` or `build-modules` directory builds of the package on npm.  Might want to publish this as a _next/beta_ version first to see if that happens again?

psealock:
This package is used in the build process so it does not get a `build` or `build-modules` directory, the source is used instead. I think your directories are left over from a previous version perhaps? 

https://github.com/woocommerce/woocommerce-admin/blob/52b19ab751ed566decad3d22f04fd738964fd09a/bin/packages/build.js#L226-L231

I think `settings` omission from `packages.js` was the sole reason why it wasn't working, but after I publish I will give it a test 😄

psealock:
Tested using `create-wc-extension` and I think we're good here:

Screen Shot 2021-02-15 at 10 58 10 AM

nerrad:
> This package is used in the build process so it does not get a build or build-modules directory, the source is used instead. I think your directories are left over from a previous version perhaps?

Ahh ya, that was it. Confirmed this works in the stripe pr implementing it. Thanks Paul!

PR #6332, opened 2021-02-12
Implement the PayU payment task and fix a bug

### LINKED ISSUE #6172: Payments task - include PayU as an option in India
**Description:** We need to start including PayU as an option in the Payments task for stores located in India. We aim to include it in the WooCommerce Core 5.2 release (code freeze March 18). (context: pbIJXs-Ab-p2)

The flow should be similar to what we have for other local payments, such as RazorPay in India.

**Acceptance Criteria:**
- [ ] If the user selects India as a store location we should display PayU as an option in the Payments task
- [ ] If the user clicks "Set up" we should automatically install PayU from wp.org (link TBD) 
- [ ] After the plugin is installed we should display the 2nd step where the user can add the PayU credentials. 
- [ ] Mark the step as complete after the user adds the Key details and clicks "proceed".

Logo: Available [here](https://corporate.payu.com/brand-portal/)

Title: PayU for WooCommerce
Description: Enable PayU’s exclusive plugin for WooCommerce to start accepting payments in 100+ payment methods available in India including credit cards, debit cards, UPI, & more!

samueljseay:
Fixes #6172 

This adds the PayU plugin as a payment task option. Based on the requirements, we've opted for this just to be a simple task that installs the plugin and lets the user configure the plugin later via it's settings screen.

There was also an undiscovered bug which I reproduced with Klarna and Mollie, which is that when you click continue on the payment task calling `markConfigured` it doesn't persist the enabling of the plugin, so a page reload shows the toggle disabled but also the payment plugin will not be enabled either.

To fix this, I've mimicked the behaviour from the toggle functionality to persist the actual enabled option as well as setting the state.

To test the basic plugin functionality:

1. Go through the OBW and choose India as country, the only other thing that matters is to not have cbd as industry
2. After landing back on home screen, click the setup payment task in the task list
3. You should see PayU listed in the possible options, click the "setup" button for PayU
4. The payu-india plugin should be installed automatically, in the next step instructions matching the copy from the issue should be displayed. After clicking "continue" it should return to the payment plugin list and the toggle should be enabled
5. Do a page refresh, the plugin option should still be enabled.

Test some of the other plugins in the same way to ensure that they are persisted as enabled. Some examples include Klarna and Mollie. For Klarna and Mollie you'll need to change the country in OBW to something like "United Kingdom".


client/task-list/tasks/payments/index.js
```diff
@@ -73,21 +73,35 @@ class Payments extends Component {
 			: 'stripe';
 	}
 
-	markConfigured( method, queryParams = {} ) {
-		const { clearTaskStatusCache } = this.props;
+	async setMethodEnabledOption( method, value ) {
```
samueljseay, line 76: this is the code that was persisting the enabled option extracted so it can be reused.

client/task-list/tasks/payments/index.js
```diff
@@ -73,21 +73,35 @@ class Payments extends Component {
...
 		this.setState( {
 			enabledMethods: {
 				...enabledMethods,
-				[ method ]: true,
+				[ methodName ]: true,
 			},
 		} );
 
-		clearTaskStatusCache();
+		await this.setMethodEnabledOption( method, 'yes' );
```
samueljseay, line 101: markConfigured now persists the option too. I couldn't think of a problem with doing this here.

client/task-list/tasks/payments/index.js
louwie17: Thanks, this is nice, although maybe we should add a check here to see if `enabled` is already set to the same `value`?
Several of the payment methods handle this in the component, so this currently would trigger duplicate update calls:
Stripe: https://github.com/woocommerce/woocommerce-admin/blob/main/client/task-list/tasks/payments/stripe.js#L137
Razorpay: https://github.com/woocommerce/woocommerce-admin/blob/main/client/task-list/tasks/payments/razorpay.js#L43

I believe Paypal as well, so we did just have to consolidate this.

client/task-list/tasks/payments/index.js
```diff
@@ -383,6 +388,7 @@ export default compose(
 			'woocommerce_eway_settings',
 			'woocommerce_razorpay_settings',
 			'woocommerce_mollie_payments_settings',
+			'woocommerce_payubiz_settings',
```
louwie17, line 391: Not your original code, so no worries if you keep it this way (and not sure why I didn't notice till now).
We should have a added config with dependent options (`options_needed: ['woocommerce_payubiz_settings']` some have more then one), so that here we could just loop through all the `methods` and not have to worry about adding one for each new payment method.

client/task-list/tasks/payments/methods.js
```diff
@@ -374,6 +376,27 @@ export function getPaymentMethods( {
...
+				</>
+			),
+			before: <PayUIndiaLogo />,
+			visible: countryCode === 'IN' && ! hasCbdIndustry,
+			plugins: [ 'payu-india' ],
+			container: <PayUIndia />,
+			isConfigured: activePlugins.includes( 'payu-india' ),
+			isEnabled:
+				options.woocommerce_payubiz_settings.enabled &&
+				options.woocommerce_payubiz_settings.enabled === 'yes',
```
louwie17, line 397: With the last paypal changes I added `enabledPaymentGateways`, so you could check for `enabledPaymentGateways.includes('payubiz')` instead.

client/task-list/tasks/payments/index.js
louwie17: I mentioned a slight problem in the comment above, although an extra condition will fix that.

louwie17 (changes requested):
Thanks @samueljseay, things tested pretty well, although I did get a strange list of notices after enabling it for the first time, they look to be `payu-india` plugin issues allowing it to be enabled before configuring (also happened on Jurassic), I assume because it's in debug mode I see this.
Screen Shot 2021-02-12 at 10 16 41 AM

This actually brings up a slightly different thought (concern of mine), is allowing users to `enable` a payment method without finishing the configuration (I know the old payment settings allows for this (for some)), but do we want to keep that flow in the OBW? Or give some kind of visual cue that set up is still needed before enabling the plugin?
I think this is mostly an issue with the plugins where we don't support onboarding or ask for credentials -> PayU, Klarna, Mollie (maybe a couple others)
Not sure if this has been discussed before, cc'ing @pmcpinto @elizaan36

elizaan36:
> This actually brings up a slightly different thought (concern of mine), is allowing users to enable a payment method without finishing the configuration (I know the old payment settings allows for this (for some)), but do we want to keep that flow in the OBW? Or give some kind of visual cue that set up is still needed before enabling the plugin?
I think this is mostly an issue with the plugins where we don't support onboarding or ask for credentials -> PayU, Klarna, Mollie (maybe a couple others)
Not sure if this has been discussed before, cc'ing @pmcpinto @elizaan36

I'm not sure what the requirements were for adding this to the task list, but seems like pretty bad user experience to enable a plugin with no indication that it won't actually work until you configure it in settings. Why aren't we giving users a clear path to configure the setup in order to use the plugin?

louwie17:
> Why aren't we giving users a clear path to configure the setup in order to use the plugin?

@elizaan36 we do say "finish the configuration in the <plugin name> settings" in the set up process, but once the user clicks **Continue** it will enable the plugin. This might of been a miss configuration with the first example, but then adopted by the others.
Example:
Screen Shot 2021-02-12 at 2 19 21 PM

elizaan36:
thanks @louwie17. Just to clarify, is this step included with the PayU payment task?

If it's required to visit settings in order to use the plugin, we could explore including the link to Settings as a step 3. 

For example: 
> **3. Set up your [Gateway] account.** 
[Gateway] can be configured under your store settings. Figure out what you need. 
[Set up]

Thoughts?

louwie17:
> thanks @louwie17. Just to clarify, is this step included with the PayU payment task?

Yeah it is, it say's this: `PayU can be configured under your store settings. Create your PayU account here.`

> If it's required to visit settings in order to use the plugin, we could explore including the link to Settings as a step 3.
> 
> Thoughts?

I am not sure how step 3 differentiates from step 2 here? 

I might of not been totally clear in relation to what I meant with `enable`, I mean enable the specific payment gateway/method for use as payment. Enabling the plugin happens in the first step with installation, this is fine.

Step 2 is most of the time the configuration step, which provides the link to settings if it doesn't provide the onboarding.
Then when the user clicks continue it enables the specific payment gateway/method for use as a payment option.
Screen Shot 2021-02-12 at 2 52 43 PM

I think it's fine to keep it as is, just not enabling the payment method until the settings are added. We can still display the `Set up` button until the settings are added, it would jump to step 2 by default next time the user visits.
Hope that makes a bit more sense.

samueljseay:
@louwie17 you're right this flow is not good. but the bug lies with PayU in this case. The enable option is available on every payment gateway plugin, and their is no form validation to stop a user saving just the "enabled" flag if they go through the form, if the plugin doesn't handle these values being empty I doubt there is much we can do about it? Maybe I'm missing something there though.

> I think it's fine to keep it as is, just not enabling the payment method until the settings are added

how would we realistically do this though? We'd have to interrogate whether some settings were added but which ones? Also there is no requirement to add all the currencies yet their code seems to error for every single currency input field.

louwie17:
@samueljseay true in that the problem is with the PayU plugin, and I don't think this is a blocker for these changes.

The warning messages happened to put more emphasizes on the current flow, which I thought I call out, we can certainly address this or re-think it, in a different PR/issue. As currently a user can still toggle the `enable` button on the (old) payment setting screen as well (so it's not like it's any different there).
I have only seen the new Paypal plugin **actually** redirect to the set up screen when you try to enable it without having set the credentials.

> how would we realistically do this though? We'd have to interrogate whether some settings were added but which ones? Also there is no requirement to add all the currencies yet the PayU plugin code seems to error (actually its just a warning) for every single currency input field.

I guess by looking at how the plugin saves it's credentials and checking if it has been added (not overly great).

samueljseay:
> I guess by looking at how the plugin saves it's credentials and checking if it has been added (not overly great).

Yes this will probably be ok for most plugins. For pay U it supports 10 "credential" sets (key and salt per currency) so I guess we could check if at least one is set, feels a bit like we have to do guesswork to ensure we do the right thing but I don't see a better way.

I agree with most of what you've said though I think we should probably look to refactor this in future. I wonder if this should add to the discussion around fixing payment tasks api.

elizaan36:
Just reading through all the comments again I understand this is probably for a different PR/issue, but it's a great question to bring up and we should definitely make sure that "Set up" appears before "Enable gateway". It's really easy to miss the Settings link otherwise in the current step.

client/task-list/tasks/payments/index.js
samueljseay: Ah yep, I'm gonna leave this one for this PR, mainly because I think a better treatment would be a proper refactor of this code.

client/task-list/tasks/payments/methods.js
samueljseay: good suggestion and thanks for adding that setting. Resolved here: https://github.com/woocommerce/woocommerce-admin/pull/6332/commits/3d81baa0c8db3588d964e907f46fc3a4602e06fa

client/task-list/tasks/payments/index.js
samueljseay: Yep great call, I resolved that here: https://github.com/woocommerce/woocommerce-admin/pull/6332/commits/f2c766263cc8e6a5dea502bb6449fdf99f4e6b65

also note I did a couple more things here:

1. I refactored the function to take a method name instead of relying on the method object, it was just much clearer to read.
2. I extracted the function so it could be tested and added some tests. I think this is simpler than trying to test it via the component itself. The component does way too much at the moment, would be great to break it down in future.

louwie17 (approved):
Thanks for the updates @samueljseay and for taking the time to write a couple extra tests 🥇 
Still testing well for me and good to merge 👍

pmcpinto:
> Just reading through all the comments again I understand this is probably for a different PR/issue, but it's a great question to bring up and we should definitely make sure that "Set up" appears before "Enable gateway". It's really easy to miss the Settings link otherwise in the current step.

I fully agree. We should explore how to make it clear that users still need to fine-tune some settings.

PR #6335, opened 2021-02-12
Use box sizing and padding to fix nav and admin menu styling

joshuatf:
Fixes an issue that caused the admin menu to be pushed down by the header height.

### Screenshots

**Before**
Screen Shot 2021-02-12 at 1 52 41 PM
**After**
Screen Shot 2021-02-12 at 1 47 47 PM

1. Disable the new nav.
2. Check that the admin menu styling is correct.
3. Navigate to various embed and non-embed pages.
4. Enable the new nav.
5. Check both embed and non-embed pages.
6. Check mobile screen sizes to make sure things appear as expected.

### DIFF
--- a/client/layout/style.scss
+++ b/client/layout/style.scss
@@ -22,11 +22,6 @@
 	.update-nag {
 		display: none;
 	}
-
-	#wpbody {
-		padding-top: 0;
-		display: block;
-	}
 }
 
 .woocommerce-admin-is-loading {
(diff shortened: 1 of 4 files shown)


joelclimbsthings (changes requested):
Thanks Josh. 
This works for me on embedded pages, but not on react-based pages:

image

joshuatf:
Thanks, @joelclimbsthings - moved these to the reset file and applied to all pages.

joelclimbsthings (approved):
Thanks Josh, works well :shipit:

PR #6336, opened 2021-02-12
Fix navigation priority settings conflict

joshuatf:
Fixes a conflict between the settings pages PR and nav changes which update the hooks from `admin_menu_classes` to `admin_menu`.

This PR also adds in some fallback measure to prevent the bizarre styling seen when the nav component is loaded, but the server did not correctly detect `is_woocommerce_page()`.

Note this PR may partially break the new settings pages, but since this is behind a flag on not yet actively being worked on, I think we can follow up with a fix here.

### Screenshots

Before
Screen Shot 2021-02-12 at 2 51 44 PM

After
Screen Shot 2021-02-12 at 2 51 27 PM

1. Enable the new nav.
2. Make sure styles aren't broken.
3. Optionally, comment out this line in Screen.php and note that styles get "unbroken" after the nav is loaded.
`$classes .= ' has-woocommerce-navigation';`

### DIFF
--- a/client/header/index.js
+++ b/client/header/index.js
@@ -51,7 +51,7 @@ export const Header = ( { sections, isEmbedded = false, query } ) => {
 		debounceTimer = setTimeout( function () {
 			const wpBody = document.querySelector( '#wpbody' );
 
-			if ( ! wpBody ) {
+			if ( ! wpBody || ! headerElement.current ) {
 				return;
 			}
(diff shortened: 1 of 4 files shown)


joelclimbsthings:
Looks good, and works well. :shipit:

joelclimbsthings (approved):

PR #6337, opened 2021-02-12
Fixing site errors being concealed by new navigation

### LINKED ISSUE #5859: Site errors concealed by new navigation and top header
**Describe the bug**
The content of site errors is currently hidden/overlaid by the  new navigation and the top heading bar. 

**To Reproduce**

1. Enable new navigation
2. Trigger an error (such as my trying to use a non-existing class)
3. Observe the error display

**Expected behavior**

- Errors should be visible with nav expanded
- Errors should be visible with nav folded
- Errors should be visible in mobile view

**Screenshots**

Screenshot from 2020-12-02 12-57-53

joelclimbsthings:
Fixes #5859

Certain site errors are displayed in a way that they are covered up by the new site navigation. This PR amends this to apply appropriate margins to the errors only when the new navigation is enabled and expanded.

### To-do
- Potentially fix by altering changes in https://github.com/woocommerce/woocommerce-admin/pull/6247 to target `body` instead of `#wpbody`
- Respond to negative margin in https://github.com/woocommerce/woocommerce-admin/pull/6371

### Screenshots

image

-   Checkout branch
- Ensure the new navigation is enabled
- Navigate to a WC-Admin page (such as Homescreen)
-   Make a quick code change to cause an error. I did so by changing a variable name in `Screen.php` -> `add_screen()`. 
- Confirm that the resulting error(s) display as they do in the screenshot, and are not covered up by the side navigation.

### DIFF
--- a/client/navigation/style.scss
+++ b/client/navigation/style.scss
@@ -38,7 +38,9 @@ body.is-wc-nav-expanded {
 		height: 100%;
 	}
 
-	font > .xe-notice {
+	font > .xe-notice,
+	font > .xdebug-error,
+	font > .xe-warning {
 		margin-left: calc(#{$navigation-width} + #{$gap});
 	}
 }
@@ -132,9 +134,11 @@ body.is-wc-nav-folded {
 		#wpbody-content {
 			margin-top: $adminbar-height;
 		}
+	}
 
-		font > .xe-notice {
-			margin-top: $header-height;
-		}
+	font > .xe-notice,
+	font > .xdebug-error,
+	font > .xe-warning {
+		margin-top: $header-height;
 	}
 }
--- a/client/navigation/style.scss
+++ b/client/navigation/style.scss
@@ -38,9 +38,7 @@ body.is-wc-nav-expanded {
 		height: 100%;
 	}
 
-	font > .xe-notice,
-	font > .xdebug-error,
-	font > .xe-warning {
+	font > .xdebug-error {
 		margin-left: calc(#{$navigation-width} + #{$gap});
 	}
 }
@@ -136,9 +134,7 @@ body.is-wc-nav-folded {
 		}
 	}
 
-	font > .xe-notice,
-	font > .xdebug-error,
-	font > .xe-warning {
+	font > .xdebug-error {
 		margin-top: $header-height;
 	}
 }
--- a/client/navigation/style.scss
+++ b/client/navigation/style.scss
@@ -37,6 +37,10 @@ body.is-wc-nav-expanded {
 		width: $navigation-width;
 		height: 100%;
 	}
+
+	font > .xe-notice {
+		margin-left: calc(#{$navigation-width} + #{$gap});
+	}
 }
 
 body.is-wc-nav-folded {
@@ -127,6 +131,10 @@ body.is-wc-nav-folded {
 	&:not(.is-wp-toolbar-disabled) {
 		#wpbody-content {
 			margin-top: $adminbar-height;
+		}
+
+		font > .xe-notice {
+			margin-top: $header-height;
 		}
 	}
 }


client/navigation/style.scss
```diff
@@ -33,6 +33,10 @@ body.is-wc-nav-expanded {
 		width: $navigation-width;
 		height: 100%;
 	}
+
+	font > .xe-notice {
```
joshuatf, line 37: These classes don't quite seem to work for me with errors.  I see these classes:

`xdebug-error xe-warning`

client/navigation/style.scss
```diff
@@ -127,4 +131,8 @@ body.is-wc-nav-folded {
 	#woocommerce-embedded-root.is-embed-loading {
 		margin-bottom: -$adminbar-height;
 	}
+
+	font > .xe-notice {
+		margin-top: $header-height;
```
joshuatf, line 136: I'm afraid of going back down this path where we need to define margins on all elements, especially if they can be variable.

Maybe we should move the margin/padding I added for `#wpbody` in https://github.com/woocommerce/woocommerce-admin/pull/6247 to the `body`.  What do you think?

joshuatf:
Just noting here that changes in https://github.com/woocommerce/woocommerce-admin/pull/6354 may need to be take into account if we change to margin/padding on the `body`.

joelclimbsthings:
So I've been exploring the possibility of addressing this issue by moving the margin to the `body` tag as @joshuatf suggested. I've also created a competing PR with that approach here -> https://github.com/woocommerce/woocommerce-admin/pull/6528

I'm feeling a bit less bullish on that approach at the moment.

Adding the dynamic margin to the `body` tag is just a broader change which can impact other site structures. You can see that in the other PR there is an existing issue with the old navigation sidebar:

image

With the margin on the `body`, it's pushing down the content of the older sidebar as you'd expect. Now we _could_ likely address this with additional CSS, but before diving down that rabbit hole I'm questioning if it's worth it.

I definitely see the value in a single solution to control the header margin here, but I think it's worth considering that it's _not_ worth making that a more complex solution based on what's likely a one-off exception. My  thought is that it will be very rare (and likely just this once) that we need to worry about content outside of `#wpbody`. And, if changes in the future cause an issue with site errors not being displayed quite right again, it doesn't seem too critical?

So the options:
- A more complex solution utilizing the `body` for margin, and dealing with exceptions to that margin such as the old sidebar content.
- The single style change in this PR that will operate outside of the existing solution for header bar spacing.

I'm personally leaning towards just allowing the second option for now. If we find that other exceptions are going to exist, then I'd definitely favor revisiting and pursuing the `margin-on-body` solution.

What do you all think?

cc @joshuatf @psealock

psealock (commented):
>The single style change in this PR that will operate outside of the existing solution for header bar spacing.

That feels right to me too. The rules introduced in this PR only affect the `xdebug-error` class and rewriting the logic to accomodate this seems like a poor use of time.

joshuatf (approved):
>That feels right to me too. The rules introduced in this PR only affect the xdebug-error class and rewriting the logic to accomodate this seems like a poor use of time.

@psealock Just a heads up that this is not a huge rewrite and already done here - https://github.com/woocommerce/woocommerce-admin/pull/6528

I'm cautiously approving this PR, but let's keep an eye out for other elements that may be affected by this change.  I know this is only a single element, but that's how the previous header issues began.

To outline why I'm so tentative with this approach, we have a number of combos that create a variability in header height:
* Old nav, wc admin header
* Old nav, full screen
* Old nav, wc admin header, mobile app banner
* New nav, wc admin header
* New nav, wc admin header, mobile app banner
* New nav, wc admin header, mobile app banner, admin bar
* New nav, full screen

Multiply the above x2 for mobile variation and we see how a single innocent change spins up 14 new issues.

TLDR - this looks good to merge, but let's keep an eye out for other issues.

psealock:
>Multiply the above x2 for mobile variation and we see how a single innocent change spins up 14 new issues.

Thanks for the context @joshuatf, ok lets keep this in mind

PR #6046, opened 2021-01-12
Bump Minimum Supported PHP Version to 7.0

adrianduffell:
Fixes #6039

This bumps the minimum supported version of PHP to `7.0` to match [WooCommerce’s version requirements](https://developer.woocommerce.com/2019/12/10/new-minimum-requirements-in-woocommerce-3-9/).

- The minimum PHP version is updated in the readme content and tags.
- The PHP version dependency is updated in the Composer configuration.
- The Travis test matrix has PHP `5.6` removed.

### Detailed test instructions:

Since this changes the Composer config, run `composer install` from a PHP 7 environment and perform a quick smoke test.

PR #6342, opened 2021-02-16
Update Minimum version in docs and test configs

psealock:
In https://github.com/woocommerce/woocommerce-admin/pull/6330 the minimum version of WP was bumped to 5.4, but docs and testing configs weren't updated, so this PR handles that. Minimum WooCommerce version is also bumped to 4.8

### To Test

* See that changes make sense
* See CI tests passing

### DIFF
--- a/docker/wc-admin-php-test-suite/docker-compose.yml
+++ b/docker/wc-admin-php-test-suite/docker-compose.yml
@@ -9,8 +9,8 @@ services:
             - '../../:/app'
         environment:
             - WC_CORE_DIR=/tmp/wordpress/wp-content/plugins/woocommerce
-            - WC_VERSION=${WC_VERSION:-5.0.0}
-            - WP_VERSION=${WP_VERSION:-5.6}
+            - WC_VERSION=${WC_VERSION:-4.8.0}
+            - WP_VERSION=${WP_VERSION:-5.4}
             - DB_USER=root
             - DB_PASS=password
             - DB_NAME=wordpress_test
--- a/docker/wc-admin-php-test-suite/docker-compose.yml
+++ b/docker/wc-admin-php-test-suite/docker-compose.yml
@@ -1,33 +1,43 @@
-version: "3"
+version: '3'
 
 services:
-  phpunit:
-    build: "."
-    image: wc-admin-php-test-suite-phpunit:1.2.1
-    volumes:
-      - "test-suite:/tmp"
-      - "../../:/app"
-    environment:
-      - WC_CORE_DIR=/tmp/wordpress/wp-content/plugins/woocommerce
-      - WC_VERSION=${WC_VERSION:-4.5.0}
-      - WP_VERSION=${WP_VERSION:-5.3}
-      - DB_USER=root
-      - DB_PASS=password
-      - DB_NAME=wordpress_test
-      - DB_HOST=phpunit-db
-    depends_on:
-      - phpunit-db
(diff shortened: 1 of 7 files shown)


adrianduffell (changes requested):
Thanks for this @psealock, overall it looks good.

I just had a question about the default versions for the Docker config. It looks the versions were bumped to the maximum version instead of the minimum.

readme.txt
```diff
@@ -36,8 +36,8 @@ WooCommerce Admin also allows store owners to customize a new dashboard screen w
 
 = Minimum Requirements =
 
-* WordPress 5.3
-* WooCommerce 4.5 or greater
+* WordPress 5.4
+* WooCommerce 4.8 or greater
 * PHP version 7.0 or greater. PHP 7.2 or greater is recommended
```
samueljseay, line 41: its interesting to me that our readme says php 7.0 but our plugin file says php 5.6.20.

I think woo core also says php 5.6.20 on their plugin and I *think* thats still the min supported version by Wordpress. should we change this to reflect that?

adrianduffell: We bumped the PHP version to 7.0 recently [here](https://github.com/woocommerce/woocommerce-admin/pull/6046), but it looks like I missed the version number in the plugin file.

samueljseay: Thanks @adrianduffell I missed that wc had updated that!

docker/wc-admin-php-test-suite/docker-compose.yml
```diff
@@ -1,33 +1,43 @@
...
+    phpunit:
+        build: '.'
+        image: wc-admin-php-test-suite-phpunit:1.2.1
+        volumes:
+            - 'test-suite:/tmp'
+            - '../../:/app'
+        environment:
+            - WC_CORE_DIR=/tmp/wordpress/wp-content/plugins/woocommerce
+            - WC_VERSION=${WC_VERSION:-5.0.0}
+            - WP_VERSION=${WP_VERSION:-5.6}
```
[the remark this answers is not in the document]
psealock: That makes total sense, thanks @adrianduffell. I meant to ask about this specifically. I changed it back to the minimum versions in https://github.com/woocommerce/woocommerce-admin/pull/6342/commits/016dfc9b7ab8b2f1dd0461edb5cb9a9fdffb0fbf

readme.txt
psealock: Good spotting, thank you! Fixed in https://github.com/woocommerce/woocommerce-admin/pull/6342/commits/73a1bf7fc7627327f57c4498e6c6c29b96790d48

adrianduffell (approved):
Looking good! 🚀

docker/wc-admin-php-test-suite/docker-compose.yml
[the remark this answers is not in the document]
adrianduffell: 👍 I think I could document this better. I'll look to do that in the next cooldown.

psealock:
@samueljseay Do you know if GH runs PR checks against the `test-php.yml` on `main` or branch? This test shouldn't be in the matrix.

Screen Shot 2021-02-18 at 12 25 59 PM

samueljseay:
@psealock its a little awkward interaction with the "required checks" feature of github. Even if the matrix run doesn't exist in the yml, if you've marked it as required it'll still come up. The solution is to turn the check off in github settings (its under branch protection)

PR #6347, opened 2021-02-16
Fix Markdown formatting in page-controller.md

tomalec:
Markdown formatting is of for [`page-controller.md`](https://github.com/woocommerce/woocommerce-admin/blob/main/docs/page-controller.md). 
- Some list items are double decorated.
- ' and ` are mismatched, making `code` elements created in the wrong places.

### Accessibility

-   [x] I've tested using a screen reader
-   [n/a] All animations respect [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion)
-   [n/a] All text has [at least a 4.5 color contrast with its background](https://webaim.org/resources/contrastchecker/)

#### Before 
Screenshoot of double-decorated list items.

Screenshoot of mismatched ' and `.

#### After

Screenshoot of fixed list item decorations.

Screenshot of fixed code elements.

-  Preview https://github.com/woocommerce/woocommerce-admin/blob/0b97aa4792c5937dea41744cd3691d6dfdafef64/docs/page-controller.md
- Check if it's formatted correctly.

### Changelog note

I think it's not needed for such a cosmetic change.

### DIFF
--- a/docs/page-controller.md
+++ b/docs/page-controller.md
@@ -70,10 +70,10 @@ WooCommerce Admin implements it's own version of `get_current_screen()` to allow
 
 Some screen ID formats that the function will generate are:
 
--   -   `{$current_screen->action}-{$current_screen->action}-tab-section`
--   -   `{$current_screen->action}-{$current_screen->action}-tab`
--   -   `{$current_screen->action}-{$current_screen->action}` if no tab is present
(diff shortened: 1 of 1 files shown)


samueljseay (approved):
This looks much nicer! Thanks for this contribution!

PR #6348, opened 2021-02-16
New Settings: Turn off in dev mode

psealock:
Since there currently isn't a way to disable the new Settings from the UI. This turns off Settings from development mode because it difficult to get out of for those without easy CLI access.

cc @joshuatf because I know you are still tinkering in this area.

### DIFF
--- a/config/development.json
+++ b/config/development.json
@@ -13,7 +13,7 @@
 		"navigation": true,
 		"onboarding": true,
 		"remote-inbox-notifications": true,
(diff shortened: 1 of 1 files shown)


joshuatf (approved):
I wonder how someone without CLI access could turn this on?

Either way, I'm good with disabling this for now.

psealock:
I think an alternative solution, once we dedicate some time to this will be to simply add the toggle

PR #6354, opened 2021-02-17
override wpbody styles when nav present

### LINKED ISSUE #6350: OBW: Header styles broken with Navigation
With Navigation turned off, the OBW's header progress bar is styled incorrectly

Screen Shot 2021-02-17 at 11 42 17 AM

### To Reproduce

1. Turn on Navigation
2. Visit the OBW
3. See style issues in the header progress bar

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/6350

In the new Header styles refactor, Navigation styles weren't considered so full screen had incorrect `margin-top` and `padding-left`. This PR fixes those styles when `has-woocommerce-navigation` is present on the `body`

Screen Shot 2021-02-17 at 5 26 37 PM

1. Turn on Navigation
2. Visit OBW
3. See things nicely laid out
4. Turn off Navigation
5. See OBW looking good as well

### DIFF
--- a/client/layout/style.scss
+++ b/client/layout/style.scss
@@ -48,11 +48,6 @@
 	font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto,
 		Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif;
 
-	margin-top: -$adminbar-height;
-	@include breakpoint( '<600px' ) {
-		margin-top: -$adminbar-height-mobile;
-	}
-
 	#wpwrap {
 		top: 0;
 	}
@@ -87,12 +82,13 @@
 		#wpbody {
 			padding-left: 0;
 		}
+	}
+}
 
-		#wpcontent {
-			> #wpbody {
-				margin-top: 32px !important;
-			}
-		}
+.wp-toolbar .woocommerce-admin-full-screen {
+	margin-top: -$adminbar-height;
+	@include breakpoint( '<600px' ) {
+		margin-top: -$adminbar-height-mobile;
 	}
 }
 
--- a/client/layout/style.scss
+++ b/client/layout/style.scss
@@ -45,8 +45,8 @@
 .woocommerce-admin-full-screen {
 	background: $studio-gray-0;
 	color: $studio-gray-60;
-	font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell,
-		'Helvetica Neue', sans-serif;
+	font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto,
+		Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif;
 
 	margin-top: -$adminbar-height;
 	@include breakpoint( '<600px' ) {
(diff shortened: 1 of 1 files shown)


client/layout/style.scss
```diff
@@ -82,6 +82,18 @@
 			margin-top: 0 !important;
 		}
 	}
+
+	&.has-woocommerce-navigation {
+		#wpbody {
```
joshuatf, line 87: Just want to note that this may need to be updated depending on the follow up to https://github.com/woocommerce/woocommerce-admin/pull/6337#discussion_r576431976 cc @joelclimbsthings

client/layout/style.scss
```diff
@@ -82,6 +82,18 @@
...
+			padding-left: 0;
+		}
+
+		#wpcontent {
+			> #wpbody {
+				margin-top: 32px !important;
```
joshuatf, line 93: Instead of adding this, I think we could apply the above negative margin fix on `.woocommerce-admin-full-screen` to only pages with toolbars.

```
.wp-toolbar .woocommerce-admin-full-screen {
    margin-top: -$adminbar-height;
}
```

joshuatf (commented):
Thanks for fixing this, Paul.  This works well, but left a quick change we can make here to simplify this.

Ultimately, it would be great if we could slot fill the actual WC header from the profiler, but that's a bigger issue to tackle in the future.

client/layout/style.scss
joelclimbsthings: Right, going to get to this shortly, but my inclination is that we'll be changing everything to target the `body` as @joshuatf suggested.

client/layout/style.scss
psealock: I _think_ this is required because the block above is `!important` and needs to be overwritten. 

When looking at the OBW with Nav turned on, there is no `.wp-toolbar` so it continues to look off:

Screen Shot 2021-02-18 at 12 41 00 PM

client/layout/style.scss
psealock: Doing the refactor described in https://github.com/woocommerce/woocommerce-admin/pull/6337#discussion_r576431976 seems low priority compared to remaining issues slated for 2.1. Would it be better to include this change, then work on the refactor during Cooldown?

joshuatf: I don't mind merging this one in first and @joelclimbsthings can update in his PR, but I'd prefer to not start layering on a bunch of CSS hacks if we can avoid it.

That led us to the untangling a web of bugs that spanned across the app.

client/layout/style.scss
joshuatf: The margin should not be applied to the `#wpbody`.  The margin is added above like so:

```scss
	margin-top: -$adminbar-height;
	@include breakpoint( '<600px' ) {
		margin-top: -$adminbar-height-mobile;
	}
```

But my suggestion is to move that logic out and only apply it when the toolbar is present so that we don't need to apply margin on one element and then conditionally add negative margin on another:

```scss
.wp-toolbar .woocommerce-admin-full-screen {
	margin-top: -$adminbar-height;
	@include breakpoint( '<600px' ) {
		margin-top: -$adminbar-height-mobile;
	}
}
```

Tested and this works with the new nav on or off.

psealock:
Thanks @joshuatf, much better solution. Confirmed working on OBW with and without Nav and same with normal pages.

joshuatf (approved):
Looks good and is testing well!  Thanks for the quick fix, @psealock.

joshuatf:
@psealock Going to merge this in to take advantage of header fixes in an upcoming PR.

PR #6355, opened 2021-02-17
Remove CES actions for adding and editing a product and editing an order

### LINKED ISSUE #6308: Customer Effort Score: remove it from some features in WC 5.2 or 5.3
**Description:** the scores collected between January and March will enough to gather insights on most of the features that we're tracking. We should disable the CES in the following features (ideally in WooCommerce 5.2):

- `product_update`
- `product_add_publish`
- `shop_order_update`

becdetat:
Fixes #6308

This removes the code that triggers a customer effort score survey in response to adding and editing a product and editing an order.

### Testing instructions

1. Add a product. The customer effort score survey should not appear.
2. Edit a product. The customer effort score survey should not appear.
3. Edit an order. The customer effort score survey should not appear.

### DIFF
--- a/TESTING-INSTRUCTIONS.md
+++ b/TESTING-INSTRUCTIONS.md
@@ -3,6 +3,12 @@ Testing instructions
 
 ## Unreleased
 
+### Remove CES actions for adding and editing a product and editing an order #6355
+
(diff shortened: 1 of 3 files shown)


ilyasfoo (approved):
Tests well. LGTM! 🚢

PR #6356, opened 2021-02-17
Refactor test:php npm script to use a post script

### LINKED ISSUE #6279: Unable to run a single test only in Docker PHP test suite
**Describe the bug**

It is not possible to run a single test anymore with `npm run test:php -- --filter=<name of test>`.

This is because the `--filter` option is sent to the last statement in the npm script `docker-compose -f docker/wc-admin-php-test-suite/docker-compose.yml down`.

**To Reproduce**

Run `npm run test:php -- --filter=foo`

This should **NOT** run any tests because none are named `foo`. Observe that all tests run.

**Expected behavior**

Only tests matching the filter should run.

adrianduffell:
Fixes #6279 

This refactors the `test:php` npm script to use a [post script](https://docs.npmjs.com/cli/v6/using-npm/scripts#pre--post-scripts). It fixes a bug where additional parameters could not be passed to the `phpunit` command, because they were sent to the `down` command instead.

Admittedly, this adds some noise to our already large npm scripts collection. I feel like we could replace this technique with a small script at a later date. 

With Docker installed

-   Check that all tests run `npm run test:php`
-   Check no tests are run when this additional parameter is used `npm run test:php -- --filter=foo`

```
PHPUnit 7.5.20 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.3.24
Configuration: /app/phpunit.xml.dist



Time: 4.3 seconds, Memory: 60.50 MB

No tests executed!
``` 
-   Confirm the MySQL container is shutdown after PHPUnit has finished.

```
Stopping wc-admin-php-test-suite_phpunit-db_1 ... done
Removing wc-admin-php-test-suite_phpunit-db_1 ... done
```


louwie17 (approved):
LGTM 🚀  Thanks for fixing this @adrianduffell

PR #6365, opened 2021-02-17
Move capability checks to client

### LINKED ISSUE #6315: New navigation is broken when user cannot view reports

joshuatf:
Fixes #6315

Moves all the privilege checks to client-side and loads WooCommerce Admin for all users.

This adds permission checks to:
* Activity Panel tabs
* WCA Routes
* Menu items

Follow-ups are probably needed to add capability checks for:
* The setup wizard
* Task list
* Inbox notes

### Screenshots

Screen Shot 2021-02-17 at 6 41 44 PM

1. Create various non-admin users with custom capabilities.  Make sure to not include the `view_woocommerce_reports` for at least one role.  https://wordpress.org/plugins/leira-roles/
2. Log in as the non-admin users.
3. Check that the correct menu items are shown.
4. Check that there aren't items shown to the user they should not be able to use or interact with.
5. Enable the new navigation under WooCommerce -> Settings -> Advanced -> Features.
6. Check that the users are able to see the new navigation menu.


client/header/activity-panel/index.js
```diff
@@ -2,17 +2,16 @@
  * External dependencies
```
becdetat, line 2: It's probably a good idea to add some smoke test instructions for this after such a big refactor

joshuatf: @becdetat There are testing instructions in the PR and added to `TESTING-INSTRUCTIONS.md`.

I do think that it's worth a good general smoke testing of the app.  Are you suggesting copying over all the generic smoke testing instructions here from the p2 thread or is there something you'd like more specific testing instructions on?

becdetat: I meant testing the activity panel - it doesn't seem to be covered by the test instructions

becdetat (approved):
Everything tests well, code looks good, I just had the suggestion about adding some general manual test instructions to cover the `activity-panels/index.js` refactor. Pre-approving.

joshuatf:
Thanks for the review, @becdetat. Added a couple lines for smoke testing around the activity panels.

PR #6368, opened 2021-02-18
Add CES track settings tab on updating settings

### LINKED ISSUE #6306: Customer Effort Score: segmenting settings data
**Description:** the settings management experience covers a wide range of use cases and just tracking the overall settings score doesn't give us a full picture of the settings that have the lowest scores and so on. We should find a way to improve the tracking, in order to get the score of specific settings areas (general, products, payments, etc).

I think that the easiest way to track this is through a new event prop just focused on settings (eg: `settings_area` : `general`, `products`..). But I'm open to evaluating other alternatives.

ilyasfoo:
This PR adds `settings_area` prop to CES prompt on settings change. This allows to track which top-level segment in settings the user is giving feedback upon.

Fixes https://github.com/woocommerce/woocommerce-admin/issues/6306.

```
```
```
```
- Go to WooCommerce  > Settings, and select a top-level tab such as Products, Shipping, etc.
- Click on `Save changes`.
- Observe in developer console, `wcadmin_ces_snackbar_view` is logged when CES prompt is displayed.
- In the event props, it should have a new `settings_area` key followed by the value of the settings tab you have selected.

### DIFF
--- a/src/Features/CustomerEffortScoreTracks.php
+++ b/src/Features/CustomerEffortScoreTracks.php
@@ -211,7 +211,7 @@ class CustomerEffortScoreTracks {
 	 * Enqueue the CES survey trigger for setting changes.
 	 */
 	public function run_on_update_options() {
-		// $current_tab is set when WC_Admin_Settings::save is called.
+		// $current_tab is set when WC_Admin_Settings::save_settings is called.
 		global $current_tab;
 
 		if ( $this->has_been_shown( self::SETTINGS_CHANGE_ACTION_NAME ) ) {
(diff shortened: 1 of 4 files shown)


src/Features/CustomerEffortScoreTracks.php
```diff
@@ -211,6 +211,9 @@ function ( $item ) use ( $clear_ces_tracks_queue_for_page ) {
 	 * Enqueue the CES survey trigger for setting changes.
 	 */
 	public function run_on_update_options() {
+		// $current_tab is set when WC_Admin_Settings::save_settings is called.
```
becdetat, line 214: Nice

becdetat (approved):
Perfect! Thanks @ilyasfoo

PR #6371, opened 2021-02-18
Add filter to allow the wp toolbar in navigation

### LINKED ISSUE #6162: Admin toolbar is hidden by the new Navigation
The new admin.php?page=wc-admin completely hides the admin toolbar. Important stuff such as debug bar are hidden. Very inconvenient.

joshuatf:
Fixes #6162 

Adds a filter that allows the toolbar to be shown.  Also refactors hidden toolbar styling to be styled via a single class so we can use it anywhere the toolbar gets hidden.

Note to @joelclimbsthings - since this relies on a negative margin on the body class, we'll probably need to using `padding-top` to adjust the `body` position in https://github.com/woocommerce/woocommerce-admin/pull/6337.

### Screenshots
Screen Shot 2021-02-18 at 10 28 00 AM

1. Check that existing styles work with a combo of the following:
* Navigation turned on/off.
* Embed pages/non-embed pages.
* Onboarding wizard.
2. Disable the wp toolbar via the filter `add_filter( 'woocommerce_navigation_wp_toolbar_disabled', '__return_false' );`
3. Check that the toolbar is shown and the above combo styling is as expected.

### DIFF
--- a/client/header/style.scss
+++ b/client/header/style.scss
@@ -47,8 +47,11 @@
 	}
 }
 
-.has-woocommerce-navigation .woocommerce-layout__header {
+.is-wp-toolbar-disabled .woocommerce-layout__header {
 	top: 0;
+}
+
+.has-woocommerce-navigation .woocommerce-layout__header {
 	left: 0;
 	width: 100%;
 }
--- a/client/layout/style.scss
+++ b/client/layout/style.scss
@@ -85,7 +85,13 @@
 	}
 }
 
-.wp-toolbar .woocommerce-admin-full-screen {
+.is-wp-toolbar-disabled {
+	#wpadminbar {
+		display: none !important;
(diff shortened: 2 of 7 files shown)


joelclimbsthings (approved):
Thanks @joshuatf , did my best to break it with enabling/disabling navigation, mobile, onboarding wizard, and all the things. Did not succeed. 

Do we need a changelog for this type of addition? Also, do we make a habit of documenting these filters in any way when adding them?

joshuatf:
Good call on both fronts, @joelclimbsthings.  Added in https://github.com/woocommerce/woocommerce-admin/pull/6371/commits/776f3df3011d5f654f335794117b4e4e351e243d

joelclimbsthings:
:shipit:

PR #6373, opened 2021-02-18
Order and styles updates to nav footer

elizaan36:
Minor changes to the nav footer:
- Lightened the border color
- Made padding visually consistent on top and bottom
- Reordered items so that `Marketplace` appears above `Settings`

image

### DIFF
--- a/src/Features/Navigation/CoreMenu.php
+++ b/src/Features/Navigation/CoreMenu.php
@@ -104,7 +104,7 @@ class CoreMenu {
 				'title'  => __( 'Settings', 'woocommerce-admin' ),
 				'id'     => 'woocommerce-settings',
 				'menuId' => 'secondary',
-				'order'  => 10,
+				'order'  => 20,
 				'url'    => 'admin.php?page=wc-settings',
 			),
 			array(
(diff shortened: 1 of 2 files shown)


src/Features/Navigation/CoreMenu.php
```diff
@@ -216,7 +216,7 @@ public static function get_items() {
 					'id'         => 'woocommerce-marketplace',
 					'url'        => 'wc-addons',
 					'menuId'     => 'secondary',
-					'order'      => 20,
+					'order'      => 10,
```
joshuatf, line 219: Nice 👍 

Let's also update the `Settings` order to `20` (should be around line 107) in case we need to add things via extensible filters/hooks later.

joshuatf (approved):
Nice job on this, @elizaan36!  Left one quick comment, but pre-approving pending that update.  💯

src/Features/Navigation/CoreMenu.php
elizaan36: updated!

elizaan36:
Thanks Joshua for taking a look! Much appreciated. Just made that change to the Settings order in that last commit.

adrianduffell:
> Just made that change to the Settings order in that last commit.

The `order` change looks good. Merging this so it lands in the 2.1 release 🙂

PR #6374, opened 2021-02-18
Removed @woocommerce/components/card from OBW

### LINKED ISSUE #6166: Remove Usage of @woocommerce\components\card
**Describe the bug**
While testing 5.0, I got the following warning during the OBW:

card-deprecation-obw-biz-details

Looks like it might be [this code](causing this), but thinking it might be good to in general just do a full sweep of the `<Card />`s in the repo to make sure we have tidied up our own usage.

octaedro:
Fixes #6166 

This PR removes the usage of "@woocommerce/components/card" from the step "Business Details" in the OBW.

I also looked for other uses of `@woocommerce/components/card` but I couldn't find any.

### Accessibility

### Screenshots
screenshot-one wordpress test-2021 02 18-16_52_20

- Checkout this branch (`git checkout fix/6166`)
- Go to the `Business Details` step on the OBW (URL `/wp-admin/admin.php?page=wc-admin&path=%2Fsetup-wizard&step=business-features`).
- Verify everything looks good and there aren't any console errors.
- Do some smoke testing of the OBW.


louwie17 (approved):
Thanks for updating this @octaedro, looks great 💯

PR #6382, opened 2021-02-19
Refactor menu item mapping and sorting

### LINKED ISSUE #6346: Unfavoriting an item puts items out of order
**Describe the bug**
Found when testing https://github.com/woocommerce/woocommerce-admin/pull/6290, moving items between menus removes the order.

The solution here is to probably include the `order` property and sort client-side.  We should also consider not handling sorting server-side to avoid sorting multiple times.

**To Reproduce**

1. Go to the nav.
2. Make sure you have multiple items in the extensions menu.
2. Favorite and unfavorite an item (not the last one).
3. Notice the item is not in the correct order.

**Expected behavior**
Expected the item to be in the correct order even after favoriting.

joshuatf:
Fixes #6346 

Refactors the menu utilities to be more performant and include sorting.

1. Activate multiple extensions that register WC menu items.
2. Favorite and unfavorite extensions
3. Check that the menu still works as expected.
4. Check that the order of menu items is correct in the extensions menu after unfavoriting.
5. Check that a user without permission to access a menu item cannot see said menu item.


client/navigation/test/utils.js
```diff
@@ -309,3 +311,247 @@ describe( 'addHistoryListener', () => {
...
+			},
+		];
+		const { categories, items } = getMappedItemsCategories( menuItems );
+
+		expect( items[ 'grand-parent' ] ).toBeDefined();
+		expect( items[ 'grand-parent' ] ).toBeDefined();
+		expect( items[ 'grand-parent' ].plugins.length ).toBe( 1 );
+
+		expect( items.child ).toBeDefined();
+		expect( items.child ).toBeDefined();
```
psealock, line 436: duplicate `expect` here and above

client/navigation/test/utils.js
```diff
@@ -309,3 +311,247 @@ describe( 'addHistoryListener', () => {
...
+		const { items } = getMappedItemsCategories( menuItems );
+
+		expect( items.parent ).toBeDefined();
+		expect( items.parent.plugins ).toBeDefined();
+		expect( items.parent.plugins.length ).toBe( 2 );
+
+		expect( items.primary ).not.toBeDefined();
+	} );
+
+	it( 'should ignore bad menu IDs', () => {
```
psealock, line 522: 👍🏽

psealock (approved):
Nice one @joshuatf 

I have some minor comments, otherwise this is testing well and code looks good. Thanks for the refactor

Issue #6377, opened 2021-02-19
Navigation: Automatically migrated items added to Flyout


psealock:
Automatically migrated items, such as MailChimp, are included in the Flyout menu. Those items should have the `hide-if-js` class appended but do not. 

<img width="353" alt="Screen Shot 2021-02-19 at 1 51 53 PM" src="https://user-images.githubusercontent.com/1922453/108441982-16afc180-72bb-11eb-8c96-47446ff2c07a.png">

### Steps to Reproduce

1. Install and activate MailChimp
2. Visit the WP Dashboard and see the flyout menu when hovering over WooCommerce
3. MailChimp should not be there.

The contents of `$submenu['woocommerce']` show that MailChimp is lacking the `hide-if-js` className

PR #6396, opened 2021-02-21
Navigation: Reset submenu before making Flyout

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/6377

Some automatically migrated menu items are still included in the WooCommerce Submenu, so they are also added to the flyout. This PR makes sure all items have the `hide-if-js` before constructing the flyout.

@joshuatf This works, but I'm not sure its the best way. I'm not fully understanding how MailChimp gets migrated, but it seems that would be a better time to add the `hide-if-js` class than what I've done. Also, just to note this solution results in the class added twice if its already present, which I'll fixup if this is the direction.

1. Activate the MailChimp plugin
2. Turn on Navigation
3. See the flyout on the WP screen and MailChimp should not be included.


joshuatf (commented):
Nice fix!  This is testing well for me.

My only comment is that this should probably be moved into the `migrate_core_child_items` method in `Menu.php`.  

This might need a couple small changes to work as `$submenu_items` merges two arrays and loses the original keys.  Maybe we just loop over `$submenu['woocommerce']` and the product item separately.

psealock:
Thanks for the review @joshuatf 

I moved the function to `migrate_core_child_items` and split out the loop innards into a separate function so product items and woo main items can be iterated over separately.

src/Features/Navigation/Menu.php
```diff
@@ -513,6 +513,54 @@ public function add_core_items() {
 		}
 	}
 
+	/**
+	 * Add an item or taxonomy.
+	 *
+	 * @param array $menu_item Menu item.
+	 */
+	public function add_item_and_taxonomy( $menu_item ) {
+		if ( in_array( $menu_item[2], CoreMenu::get_excluded_items(), true ) ) {
```
joelclimbsthings, line 522: You reference `$menu_item[2]` several times here. Totally optional, but I might assign this to a temporary variable with a descriptive name.

joelclimbsthings (approved):
Testing well, and code looks good. I added one entirely optional note. :ship:

PR #6397, opened 2021-02-22
Add a back button to the Header when the task list is displayed.

### LINKED ISSUE #6293: Add a back button to the top bar
**Overview**
We want to give merchants the ability to easily navigate back to WooCommerce Home from a task list screen so they don't feel lost. Task list screens (accessed via the home screen task list) currently don’t have any indication of how to get back to where you came from and there may be a moment of confusion or feeling stuck before you decide on where to go next. One solution would be to complete the task list flow and route users back to the home screen, but a critical usability improvement in the meantime is to add a back button.

**Background**
[The original issue has since been closed](https://github.com/woocommerce/woocommerce-admin/issues/2811), which took a more holistic approach to reveal a back button on all tertiary screens (i.e. any screen not represented by a menu item in the navigation.) Depending on the effort involved, we could start with the task list screens only and move on to others such as the Edit Product and Edit Order screens, Settings, etc. or try to tackle everything in parallel.

**Screenshots**
Home - Add my products

Mobile
Screen Shot 2021-02-08 at 7 08 56 PM

**Details**
- The page title reflects the name in the task list (i.e. Add products)
- Task list screens include Add products, Add payment methods, Add tax rates, Set up shipping costs, and Personalize my store
- There's a tooltip on hover that indicates the page you'll navigate back to
- When clicked the back button should take the user directly back to the previous screen. Most of the time this will mean going up 1 level in app hierarchy, but in some cases it will go back to another origin screen. If the user lands on a tertiary screen directly (from an email or something similar) it will default to taking the user up 1 level in app hierarchy.
- More details on the component and anatomy are in Figma that I can share separately

cc @pmcpinto @timmyc 

**Event tracking**
Event name: `wcadmin_topbar_back_button`
Event prop: `page_name` - the page where the user clicked in the back button

samueljseay:
Fixes #6293 

This was a difficult one to choose an approach on. I considered many more useful approaches that would support generic backwards nav / hierarchy navigation, but they all would have been a complexity blow out, way beyond the scope of what #6293 is actually trying to solve. (especially because navigation is complex and likely to introduce bugs and edge cases we have not foreseen)

The best alternative solution to this one that I came up with, was to come up with a static map of the parents/children that represent the structure of the home screen, so that if there is no backward URL to navigate to, you could navigate up the hierarchy. This was the simplest and yet still was too complex and raised too many questions.

Instead I have narrowed this down to be a basic check to provide a back button that goes to the home screen when on a task screen (has a query string of `task`). No matter how you arrived at the task list it will always navigate back to the home screen. This also means in the payment task list, that when you go down another navigation layer to the "setup" for each payment task, the back button there will still be the same and take you back to the home screen. In testing this felt fine and wasn't confusing as far as I could tell.

cc  @elizaan36 for testing

### Accessibility

### Screenshots

Screenshot 2021-02-22 at 2 40 33 PM

-   With a fresh install of wc/wc-admin, go to the homescreen, (when redirected to the OBW choose to skip it)
-   On the home screen you'll see the task list (heading "Get ready to start selling")
-  Try clicking on each one of the items (except store details which takes you to the OBW)
- On each screen it should provide a backwards nav chevron on the top left, hovering it should reveal a tooltip with the content "WooCommerce Home"
- Clicking the chevron should return to the home screen in each case.


elizaan36:
Sounds like the best way forward. And works perfectly. It's great that the back button persists with the specific payment methods.💯 

Just a couple copy-related notes:
- We're using sentence case everywhere in the app now, so for example instead of Choose Payment Methods
image
It would be Choose payment methods (only capitalize the first word.)

This applies to all the tasks that are using title case (all words are capitalized.)

- I'm still seeing "Home" for the shipping step
image

samueljseay:
@elizaan36 thanks for that, turns out I missed a couple of menu items that only show in certain scenarios. Could you take another look?

The other option is the setup woocommerce payments option. To make it visible you'll need to have US as chosen location in OBW and have the woocommerce payments extension installed.

As for the title case: not my fault :P the CSS changed it. I've removed the CSS that was causing the capitalization.

elizaan36:
Just tested again and looks good to me!

To confirm, when I click into the WooCommerce Payments task I see this:
image

There's no back button on this screen but that's expected. I've been having trouble connecting to my Stripe account so don't know if that's a problem for testing this flow.

samueljseay:
@elizaan36 you're right, I made a mistake here, we don't need to worry about that task at all. I'll fix this.

TESTING-INSTRUCTIONS.md
```diff
@@ -269,10 +282,19 @@ localStorage.setItem( 'debug', 'wc-admin:tracks' );
         -   The enable/disable button should be correctly reflected in the Woocommerce payment settings screen as well.
 
 Once everything is configured and enabled do a client test
+<<<<<<< HEAD
```
octaedro, line 285: It seems like we should delete this line 🙂

TESTING-INSTRUCTIONS.md
```diff
@@ -269,10 +282,19 @@ localStorage.setItem( 'debug', 'wc-admin:tracks' );
...
+-   Filling in your billing/shipping info then click pay with **Paypal**
+-   # The paypal pay window should pop up correctly without errors.
 
 -   Make sure you have added a product and store homescreen (Finish the **Personalize my store** task)
 -   Navigate to one of the products and add it to the cart
 -   Click **go to checkout**
 -   Paypal should be one of the options to pay
 -   Filling in your billing/shipping info then click pay with **Paypal**
 -   The paypal pay window should pop up correctly without errors.
+    > > > > > > > origin/main
```
octaedro, line 300: Same with this line.

client/header/index.js
```diff
@@ -18,6 +23,59 @@ import { MobileAppBanner } from '../mobile-banner';
...
+		const homeText = __( 'WooCommerce Home', 'woocommerce-admin' );
+
+		const navigateHome = () => {
+			recordEvent( 'topbar_back_button', {
+				page_name: getPageTitle( window.title ),
+			} );
+			updateQueryString( {}, getHistory().location.pathname, {} );
+		};
+
+		// if its a task list page, render a back button to the homescreen
```
octaedro, line 40: Small typo: should be `it's` instead of `its`.

client/header/index.js
```diff
@@ -18,6 +23,59 @@ import { MobileAppBanner } from '../mobile-banner';
...
+	}
+
+	return null;
+};
+
+const getPageTitle = ( defaultTitle ) => {
+	const currentUrl = new URL( window.location.href );
+	const task = currentUrl.searchParams.get( 'task' );
+
+	// If its the task list then render a title based on which task the user is on.
```
octaedro, line 67: Same here.

client/header/index.js
```diff
@@ -18,6 +23,59 @@ import { MobileAppBanner } from '../mobile-banner';
...
+	return (
+		{
+			payments: __( 'Choose payment methods', 'woocommerce-admin' ),
+			tax: __( 'Add tax rates', 'woocommerce-admin' ),
+			appearance: __( 'Personalize your store', 'woocommerce-admin' ),
+			products: __( 'Add products', 'woocommerce-admin' ),
+			shipping: __( 'Set up shipping costs', 'woocommerce-admin' ),
+		}[ task ] || defaultTitle
```
octaedro, line 75: I really like this!

readme.txt
```diff
@@ -130,6 +131,7 @@ Release and roadmap notes are available on the [WooCommerce Developers Blog](htt
 - Dev: Add navigation favorites data store #6275
 - Add: Add navigation intro modal. #6367
 - Fix: Reset Navigation submenu before making Flyout #6396
+>>>>>>> origin/main
```
octaedro, line 134: Another merge conflict line 🙂

octaedro (approved):
Good job @samueljseay! The code looks good and it's testing well here. Just left a couple of small comments, nothing big so I'll pre-approve this.

TESTING-INSTRUCTIONS.md
samueljseay: Whooops. thanks

PR #6399, opened 2021-02-22
Move Lodash to a Peer Dependency in Packages

adrianduffell:
Originally this PR intended to update the version of Lodash, which was flagged by Dependabot in https://github.com/woocommerce/woocommerce-admin/pull/4824 https://github.com/woocommerce/woocommerce-admin/pull/4826 https://github.com/woocommerce/woocommerce-admin/pull/4829.

In the comments, it was agreed to move Lodash to a peer dependency. I chose `^4.17.0` for the version constraint to support the 4.17 usage in WordPress, as [noted](https://github.com/woocommerce/woocommerce-admin/pull/6399#issuecomment-787599959) by @samueljseay. However I used a caret constraint to prevent any future v5+ breaking change being used. You can perform a [semver check](https://jubianchi.github.io/semver-check/#/constraint/^4.17.0) for this.

No changelog is needed since the individual package changelogs are updated seperately. 

Note, this feels like a breaking change since some consumers may need to add Lodash to their `package.json` before this will work. I've flagged this in the changelog entries. Go down the rabbit hole about it [here](https://github.com/semver/semver/issues/502).

-   Check the tests in CI are good.
-   Quick check of WC Admin - however this should have no affect since the it uses the WP-bundled version of Lodash.


samueljseay:
Yeah this won't have any affect due to bundled lodash.

I can see the value of it being a dependency if you're consuming this package but I wonder, @psealock whether this should be a `peer dependency` instead? 

I can't imagine there being much value in stating the dependency otherwise.

samueljseay:
@psealock bump on this ^

psealock:
When https://github.com/woocommerce/woocommerce-admin/pull/5762 is merged, it won't matter what `package.json` specifies because we will be 100% sure that `lodash` will be enqueued and what ever version is on WP Core will be used.

>but I wonder, @psealock whether this should be a peer dependency instead?

Seems appropriate

adrianduffell:
In 498a124 I've moved lodash in the components package to the `peerDependencies`. This passes CI so I think we're good to take this approach across all the packages.

A few questions:

1. I'm wondering which version string to use. If I understand peer dependencies, the consumer will need lodash in their dependencies. I think we would like to provide them some leeway with the version, rather than pinning it to an exact number like we normally do?

2. Would other dependencies besides lodash benefit from being a peer dependency?

samueljseay:
@adrianduffell In my local wordpress the lodash version loaded is 4.17.19 and there hasn't been anything but patch versions to lodash for over 4 years so probably `>= 4.17` should be good.

becdetat (approved):
Looks good to me, and everything works ok after a little run through

PR #6401, opened 2021-02-22
Deprecate Onboarding::has_woocommerce_support

adrianduffell:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/6283

This deprecates `Onboarding::has_woocommerce_support`.

The main reason is [WooCommerce support all themes](https://developer.woocommerce.com/2017/12/09/wc-3-3-will-look-great-on-all-the-themes/) now. As such, it does not seem necessary to detect theme support and restrict the available themes in the onboarding wizard.

A further reason is the method is slow.  It took ~2 seconds on my system for it to scan all of the php files in my themes folder. This contained the themes: `boutique`, `editor-blocks`, `twentynineteen`, `twentyseventeen`, `twentytwenty`, `twentytwentyone`. This could lead to timeouts on slower systems, or when there are larger numbers of themes. It is possibly related to the timeout in issue #5702.

I have deprecated the method and changed it to return `true` for all cases. I think this makes sense according to WooCommerce’s support of all themes.

### Screenshots

Screen Shot 2021-02-26 at 5 40 01 pm

- Clear existing site transients. For example, by using the [Transients Manager](https://wordpress.org/plugins/transients-manager/) plugin, and pressing the "Delete all transients" button.
- Add a new theme to WordPress but **DO NOT** activate it.
- Initialize the Onboarding Wizard.
- See that the Themes step loads fast 😎 
- See that the new theme is listed in the Themes step.


samueljseay (approved):
🚀

PR #6411, opened 2021-02-22
Allow for use of Array.flat in wc-admin client / packages

samueljseay:
My goal here was originally to be able to use polyfills like `Array.flat()` without worrying about breaking IE11, which happened recently.

To achieve this, I've updated the `@wordpress/babel-preset-default` to latest, this has the added benefit of having TS support built in, which is something I'm looking at doing next. I also added the `corejs: 3` option where relevant in our package and client builds (which both declare their own babel config), this allows us to get a much wider range of polyfill support and get new features such as Array.flat which will lessen our dependency on things like lodash. 

Additionally I have removed some npm dependencies we don't need.

To test this I wanted to confirm that the app does not crash in IE11, and that I can safely use Array.flat in both `client` and `packages`.

1. Now add some testing code in packages and client. To do this simply, I added the code in https://github.com/woocommerce/woocommerce-admin/blob/26c631b771a14a88a5ad061ce362ac8f46fef9b5/packages/components/src/badge/index.js#L10 and https://github.com/woocommerce/woocommerce-admin/blob/26c631b771a14a88a5ad061ce362ac8f46fef9b5/client/header/index.js#L29 both of these bits of code should get run on the homescreen as long as you have the setup task list active. I'll leave it as an exercise to the tester if they need to do it without the task list active, any will do.

2. Add this code: 

```
const myThing = [ [ 1 ], [ 2 ], [ 3 ] ];
console.log( 'testing IE 11 support in packages' );
console.log( myThing.flat() );
``` 
3. run a fresh build. `npm run clean && npm i && npm start`
4. At this stage there should be no build errors in console
5. open up browser stack and choose an IE11 VM. If you need help getting local testing setup with Browserstack then let me know.
6. Visit the homescreen on browserstack. Ensure that it doesn't crash or whitescreen, Check the developer console for the correct output which should show which includes console logging an array like `[1,2,3]`
7. smoke test the other screens in IE11


bin/packages/get-babel-config.js
```diff
@@ -25,6 +25,8 @@ const babelConfigs = {
 		presets: map( babelDefaultConfig.presets, ( preset ) =>
 			overrideOptions( preset, '@babel/preset-env', {
 				modules: 'commonjs',
+				corejs: '3',
+				useBuiltIns: 'usage',
```
samueljseay, line 29: This option will add polyfills such as Array.flat based on their usage in the code

psealock (approved):
Looks all legit to me. This is compiling without error locally and I can see the `flat` function call work correctly in Chrome and IE 11 tested via browserstack. 🚢

PR #6414, opened 2021-02-23
Add a changelog lint check and auto-labelling

samueljseay:
Add a lint that ensures users add a changelog to incoming PRs.

This is a pretty basic check that looks for your PR number in the "Unreleased" section of the `readme.txt` and fails if it doesn't find the PR number mentioned in it. We don't strictly look for any other kind of format as this is designed as a reminder more than a strict lint.

Very meta but I'll use this PR to test the action. Here's the action output when I have not yet provided the changelog entry:
Screenshot 2021-02-23 at 1 55 34 PM

and as of https://github.com/woocommerce/woocommerce-admin/pull/6414/commits/1a97e1dda285f3ebb3cb6d21f7fe6c04aed96fd9 it now passes:

Screenshot 2021-02-23 at 1 58 22 PM

As a follow up I've also added auto-labelling of the PR when the changelog entry is missing. Hopefully that makes it clearer to devs what is going on. I did this by adding an in-repo action that does the labelling. When the changelog entry is provided the label will be removed automatically.

The finished product and some additional commits without a changelog:

Screenshot 2021-02-23 at 10 25 31 PM

And then I added the changelog entry back and GH did its job:

Screenshot 2021-02-23 at 10 28 16 PM


adrianduffell (approved):
So good! Thanks @samueljseay 🚀

adrianduffell:
Maybe as a follow-up we could automatically add a label when the changelog is missing 🤔

samueljseay:
@adrianduffell good idea. Let me see if I can do that as part of this.

samueljseay:
@adrianduffell I'd love if you could take another look now that auto labelling is implemented 🎉 🙇

.github/actions/pr-labeler/README.md
```diff
@@ -0,0 +1,18 @@
+# PR Labeler Action
+
+This is a standalone github action hosted within wc-admin that supports
+auto adding or removing a PR label, which could be useful in various
+automation scenarios.
+
+To keep it simple and avoid checking in `node_modules` when you make
+changes to `index.js` you'll need to compile it into the `dist` directory
+via [@vercel/ncc](https://github.com/vercel/ncc).
```
samueljseay, line 9: To host an action in our repo, this was the simplest way, I've included some basic instructions for the next person who comes along. If we need to make this more automated or fancy we can do so later.

.github/actions/pr-labeler/action.yml
```diff
@@ -0,0 +1,17 @@
+name: 'PR labeler'
+description: 'Add or remove a label from a PR'
+inputs:
```
samueljseay, line 3: these are the args the action takes

.github/actions/pr-labeler/index.js
```diff
@@ -0,0 +1,60 @@
...
+		owner: github.context.repo.owner,
+		repo: github.context.repo.repo,
+		issue_number: prNumber,
+		name: label,
+	} );
+};
+
+async function run() {
+	try {
+		const prNumber = getPRNumber();
```
samueljseay, line 29: The main code that adds or removes a label based on the input args.

.github/workflows/lint-changelog.yml
```diff
@@ -0,0 +1,25 @@
...
+    lint-changelog:
+        runs-on: ubuntu-latest
+        steps:
+            - name: Cancel Previous Runs
+              uses: styfle/cancel-workflow-action@0.7.0
+              with:
+                  access_token: ${{ github.token }}
+            - name: Check out repository code
+              uses: actions/checkout@v2
+            - name: Check for changelog entry
```
samueljseay, line 14: The way this works now when running this script, is if there is not a changelog entry it errors so as to report failure on the action but it also exports a label name and action type of remove or add that can be passed along to the github action args. 

the github action always runs regardless of failure thanks to `if: ${{ always() }}`, this means it will always do the labelling but the job will still report failure if the lint check fails.

.gitignore
```diff
@@ -1,7 +1,8 @@
 
 # Directories/files that may be generated by this project
 node_modules/
-dist
+/dist
```
samueljseay, line 4: I adjusted these a bit so that dist still works as before but lets us check in the action's dist folder which is the default directory that `@vercel/ncc` compiles into

adrianduffell:
> @adrianduffell I'd love if you could take another look now that auto labelling is implemented 🎉 🙇

This is great! Thanks so much for adding the auto labelling. I agree with your approach and thanks for documenting it for the next person 👍 

All LGTM 🚀

PR #6418, opened 2021-02-23
Add CES survey for adding product categories, tags, and attributes

### LINKED ISSUE #6307: Customer Effort Score: include additional features
**Description:** we want to start triggering the Customer Effort in certain features that weren't included in the initial launch:

1. Import Products
2. Product Categories
3. Product Attributes
4. Product Tags
5. Search (orders, products, customers)

ilyasfoo:
This PR adds CES survey trigger on adding product categories and tags.

Partially fixes https://github.com/woocommerce/woocommerce-admin/issues/6307.

```
```
```
```

**Testing product categories:**
- Go to Products > Categories.
- Add a new category.
- Observe CES prompt "How easy was it to add a product category?" is displayed.

**Testing product tags:**
- Go to Products > Tags.
- Add a new tag.
- Observe CES prompt "How easy was it to add a product tag?" is displayed.

**Testing product attributes:**
- Go to Products > Attributes.
- Add a new attribute.
- Observe CES prompt "How easy was it to add a product attribute?" is displayed.


src/Features/CustomerEffortScoreTracks.php
```diff
@@ -69,27 +84,70 @@ private function enable_survey_enqueing_if_tracking_is_enabled() {
...
+					return $('.tags tbody > tr').map( function( idx, tr ) { return tr.getAttribute( 'id' ).split( '-' )[ 1 ]; } ).toArray();
+				}
+				// Gets the highest tag ID - assuming it's the latest.
+				function getMaxId() {
+					return Math.max.apply( this, getTagIds() );
+				}
+				// Hook on submit button and sets a 500ms interval function
+				// to determine successful add tag or otherwise.
+				$('#addtag #submit').on( 'click', function() {
+					const lastId = getMaxId();
```
adrianduffell, line 127: Is it possible the most recently added tag won't have the highest ID? For example, if an old tag is added after a  new one is created.

adrianduffell (dismissed):
This looks good @ilyasfoo 👍 I feel like the script is borderline too long/complex to embed inline, but I don’t think it is a blocker.

src/Features/CustomerEffortScoreTracks.php
ilyasfoo: @adrianduffell I'm not sure if I follow, how will an old tag be added after a new one? 🤔

src/Features/CustomerEffortScoreTracks.php
```diff
@@ -69,27 +84,70 @@ private function enable_survey_enqueing_if_tracking_is_enabled() {
...
+	 *
+	 * @param string $action Action name for the survey.
+	 * @param string $label Label for the snackbar.
+	 * @param string $page_now Value of window.pagenow.
+	 * @param string $admin_page Value of window.adminpage.
+	 *
+	 * @return string Generated JavaScript to append to page.
+	 */
+	private function get_script_track_edit_php( $action, $label, $page_now, $admin_page ) {
+		return sprintf(
```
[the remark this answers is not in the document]
ilyasfoo: Hmm, that sounds like a better alternative.. I'll come back to this and test. Thanks!

src/Features/CustomerEffortScoreTracks.php
adrianduffell: Sorry that was confusing 🙂 

I mean when an old tag (existing on their site for a long time) is added to the product. It's going to have a small ID so I'm not sure if `getMaxId()`, if the product already has newer tags added. I'm probably missing something!

src/Features/CustomerEffortScoreTracks.php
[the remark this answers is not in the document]
ilyasfoo: I've made the changes and tested, counting rows works just as well if not better, in terms of stability 
82b8228

src/Features/CustomerEffortScoreTracks.php
ilyasfoo: I don't think adding an old tag to a product will make a difference since the tags table that's currently displayed is not going to change. However, this is no longer relevant since we're no longer using ID, thanks for the suggestion @adrianduffell!

timmyc:
Seems like we are close to landing this one, do we need another review?

ilyasfoo:
Just need to get that ✅

src/Features/CustomerEffortScoreTracks.php
```diff
@@ -69,27 +84,61 @@ private function enable_survey_enqueing_if_tracking_is_enabled() {
...
+						} else {
+							// Form is no longer loading, most likely failed.
+							if ( $( '#addtag .submit .spinner.is-active' ).length < 1 ) {
+								clearInterval( interval );
+							}
+						}
+					}, 500 );
+				});
+			})( jQuery );",
+			addslashes( $action ),
```
adrianduffell, line 134: We could use [esc_js](https://developer.wordpress.org/reference/functions/esc_js/) instead of `addslashes`, since it handles a few more cases that could break strings.

It's not a blocker but was there a reason not to get these strings directly from JS, for example with `window.pagenow`?

adrianduffell (changes requested):
The smaller script is looking good!

I've just noticed an improvement we could make to escape the strings, and then I think this is ready 👍

src/Features/CustomerEffortScoreTracks.php
ilyasfoo: Ahh nice catch, thanks! Changes in bac5746

adrianduffell (approved):
Thanks @ilyasfoo! 🚀

PR #6420, opened 2021-02-23
Add CES survey for search product, order, customer

### LINKED ISSUE #6307: Customer Effort Score: include additional features

ilyasfoo:
This PR adds CES survey trigger on search for products, orders, and customers. Note that the survey action is shared between them, which means that it will only trigger once for either.

We're also adding an event prop `search_area` which can be either `product`, `shop_order`, or `customer`.

Partially fixes https://github.com/woocommerce/woocommerce-admin/issues/6307.

```
```
```
```

**Testing search on products:**
- Go to Products > All Products.
- Type in anything in search bar, click on "Search products".
- Observe CES prompt "How easy was it to use search?" is displayed.

**Testing search on orders:**
- Go to Orders > Orders.
- Type in anything in search bar, click on "Search orders".
- Observe CES prompt "How easy was it to use search?" is displayed.

**Testing search on customers:**
- Go to Customers.
- Type in anything in search bar, and press enter.
- Observe CES prompt "How easy was it to use search?" is displayed.


src/Features/CustomerEffortScoreTracks.php
```diff
@@ -234,4 +269,27 @@ public function run_on_update_options() {
...
+		$allowed_types = array( 'product', 'shop_order' );
+
+		// phpcs:disable WordPress.Security.NonceVerification.Recommended
+		// We're only interested for certain post types.
+		if ( empty( $_GET['post_type'] ) || ! in_array( $_GET['post_type'], $allowed_types, true ) ) {
+			return;
+		}
+
+		// Determine whether request is search by "s" GET parameter.
+		if ( empty( $_GET['s'] ) ) {
```
adrianduffell, line 286: Could [`get_search_query()`](https://developer.wordpress.org/reference/functions/get_search_query/) be used here?

src/Features/CustomerEffortScoreTracks.php
```diff
@@ -234,4 +269,27 @@ public function run_on_update_options() {
...
+
+	/**
+	 * Determine on initiating CES survey on searching for product or orders.
+	 */
+	public function run_on_load_edit_php() {
+		$allowed_types = array( 'product', 'shop_order' );
+
+		// phpcs:disable WordPress.Security.NonceVerification.Recommended
+		// We're only interested for certain post types.
+		if ( empty( $_GET['post_type'] ) || ! in_array( $_GET['post_type'], $allowed_types, true ) ) {
```
adrianduffell, line 281: Could [`get_current_screen()`](https://developer.wordpress.org/reference/functions/get_current_screen/) be used here to find the current post type?

adrianduffell (changes requested):
Thanks @ilyasfoo this tested well and LGTM overall 👍 

I just left some suggestions to use native WP functions so we could avoid disabling PHPCS.

timmyc:
Looks like this will be good to merge after a few tweaks...

src/Features/CustomerEffortScoreTracks.php
ilyasfoo: I've tried this, but unfortunately `get_search_query()` returns empty string during our hook call. Bummer 😞

src/Features/CustomerEffortScoreTracks.php
ilyasfoo: Made the change in d1fb96b. Thanks for the suggestion @adrianduffell!

adrianduffell (approved):
LGTM 🚀 

Thanks for switching over to the native WP functions where it was possible!

It would have been good to have some tests for this. Perhaps that's something we could add later in cooldown.

PR #6440, opened 2021-02-24
Correct the Klarna slug

### LINKED ISSUE #6435: Activating Klarna Payments results in swallowed JS error
**Describe the bug**
When activating Klarna Payments from the "Choose payment methods" task, a JS error occurs and no navigation happens.

**To Reproduce**

1. Set up a new store in the US that will be able to be configured with Klarna
2. Go through onboarding
3. Select the "Choose payment methods" task
4. Set up Klarna
5. Press continue:
image
6. In the console an error will appear:
image
7. and the continue button will have no other effect

**Expected behavior**
I expected no JS errors and for the continue button to navigate back to the payments configuration task page.

**Screenshots**
Further investigation shows that the Klarna Payments method name is being set incorrectly:
image
This should also probably return a useful error and/or be more resilient if the method can't be found: https://github.com/woocommerce/woocommerce-admin/blame/main/client/task-list/tasks/payments/index.js#L100

becdetat:
Fixes #6435

This fixes the Klarna slug in the `Klarna` component (`klarna.js`) to use underscores instead of hyphens. It also makes the method query throw a descriptive error if the method isn't found, rather than swallowing it in a non-descriptive null reference error.

1. Set up a new store with Klarna available as a payment processor
2. Go to the "Choose payment methods" task item
3. Set up Klarna. The plugin will install.
4. Click Continue. It should take you back to the payment methods page - previously it wasn't doing anything but a console error was displayed.


TESTING-INSTRUCTIONS.md
```diff
@@ -4,6 +4,13 @@
 
 ## 2.1.0
 
+### Correct the Klarna slug #6440
+
+1. Set up a new store with Klarna available as a payment processor
```
adrianduffell, line 9:
```suggestion
1. Set up a new store with a UK address so that Klarna available as a payment processor
```

adrianduffell (approved):
Thanks @becdetat! This tested perfectly and code LGTM 🚀 

I left an optional suggestion for the testing instructions.

becdetat:
Thanks @adrianduffell , suggestion taken :-D

PR #6441, opened 2021-02-24
Clean up unused deps and add TS to the build

samueljseay:
This is 2 things:

1. Some clean up of dependencies and scripts we're no longer using. This is timely in my opinion as we add more dependencies and package scripts as part of this, I'm trying to keep the bloat down.

2. This adds TypeScript type checking to our client/package builds via babel. 

For the TypeScript setup I looked to other woo projects such as https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3768 for inspiration but with a few specific changes:

a. Our tsconfig uses `noEmit`. Babel and webpack do the compilation work so we don't actually need TS to compile, we just need it to type check

b. I wanted one more layer of type check reporting to the developer while working, so I added this plugin to the Webpack build: https://github.com/TypeStrong/fork-ts-checker-webpack-plugin ts checks will run in a separate process and the errors will be reported in build output, so you'll see relevant type errors from ts files. You will of course also see type error reporting and TS linting in your VSCode editor thanks to the eslint setup that comes as part of this.

I considered adding some typed code as part of this, but I think I'll do that as a follow up PR. And since we don't have any TS code yet, there is no need for including a CI type check but that would probably be a good idea to include in the follow up PR.

No changelog required

## Testing instructions

A simple way to test that you get TS type checking in the terminal would be to change extension on an existing component (contains JSX) file from `packages` or `client` to `.tsx` and ensure you've run `npm run clean && npm i && npm start`

once the build has happened there is usually a delay till you see the first output from typescript but you should see type errors and warnings for the file you renamed.

Another test is to make sure that you don't see any type warnings or errors on a clean build with no file renames since TS should try to typecheck existing js files.


samueljseay:
@louwie17 maybe you'd be keen to review this one? 😄

webpack.config.js
```diff
@@ -116,7 +117,7 @@ const webpackConfig = {
 				},
 			},
 			{
-				test: /\.js?$/,
+				test: /\.(t|j)sx?$/,
```
louwie17, line 120: That's a funny/clever way to write that regex

louwie17 (approved):
This is awesome @samueljseay thanks for getting going on this :) It makes me eager to start writing some TypeScript.
Also appreciate doing some cleanup.

Screen Shot 2021-02-25 at 10 40 05 AM
Some nice looking type completion below  💯 🤩 
Screen Shot 2021-02-25 at 10 39 52 AM

PR #6443, opened 2021-02-24
Improve changelog linting automation

samueljseay:
This adds some extra logic and steps to the changelog linting workflow that hopefully makes it more useful.

The workflow now is that the lint will check your PR description for the phrase `no changelog` (case insensitive). If it detects this it will skip the linting and the check will automatically pass, also a label will not be added to remind you to add a changelog.

Furthermore the lint now happens every time you edit the PR description, so if you would like to add the `no changelog` tag later on you can do this.

The only thing this won't do is in the case that the label was added during an initial check, then you added the skip phrase later it won't remove the label for you afterwards. I evaluated the complexity of this and for now it didn't feel worth trying to achieve this. It's my thought that it is 1 click to remove the label yourself if you hit this case.

I won't be switching this check to required just yet, but in theory I think it's more robust to do that after these changes are merged.

The skip phrase is contained in this pr description so it won't lint changelog and will pass:

Screenshot 2021-02-25 at 11 37 00 AM

And then editing the PR description and removing the skip:

Screenshot 2021-02-25 at 11 45 26 AM

### DIFF
--- a/.github/workflows/lint-changelog.yml
+++ b/.github/workflows/lint-changelog.yml
@@ -21,7 +21,7 @@ jobs:
                   phrase: 'no-changelog'
                   github-token: ${{ secrets.GITHUB_TOKEN }}
                   pr-message: 'title & body'
-                  search: '["commit_messages", "pull_request"]'
+                  search: '["pull_request"]'
             - name: Check for changelog entry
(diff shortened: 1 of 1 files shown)


adrianduffell:
> when a specific phrase is mentioned in the PR description

😲 very nice!

adrianduffell (approved):
LGTM thanks @samueljseay! 🚀

PR #6453, opened 2021-02-25
Added warning when WC-Admin is active but not being used

octaedro:
This PR adds a warning when WC-Admin is active but not being used.

### Accessibility

### Screenshots
screenshot-one wordpress test-2021 02 25-13_22_04

- We need to simulate that these modifications are present in the `WC-Admin` that `WooCommerce` is using. To do it we will copy [these lines](https://github.com/woocommerce/woocommerce-admin/blob/e60af23541577dff54bb7408eb913045a26252b7/src/Loader.php#L101-L125) into `Loader.php` in the WC-Admin that lives under `packages` in the `WooCommerce` plugin (`packages/woocommerce-admin/src/Loader.php`).
- So we will add into `Loader.php` the `action` in the constructor and the method `is_using_installed_wc_admin_plugin `
- After that go back to the `woocommerce-admin` repo and checkout an older release of the plugin (like 1.8.0) `git checkout release/1.8.0`.
- Run `composer update`.
- Go to the `status` page (`/wp-admin/admin.php?page=wc-status`).

screenshot-one wordpress test-2021 02 25-14_31_06

- Verify that the `WooCommerce Admin package` used is under the `WooCommerce` plugin in the `packages` folder.
- Also the warning: `You have WooCommerce Admin activated but it is not being used.` should be visible.
- Checkout this branch (`git checkout add/poc-plugin-use-warning`) and run `composer update`.
- Now the warning shouldn't be visible anymore and the WC-Admin package used should show a path to this repo.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -94,6 +94,7 @@ Release and roadmap notes are available on the [WooCommerce Developers Blog](htt
 - Fix: Missing i18n in Welcome modal. #6456
 - Fix: Restore visual styles back to Analytics tabs. #5913
 - Add: Add a "rather not say" option to revenue in the profile wizard. #6475
+- Dev: Added warning when WC-Admin is active but not being used #6453
 
 == 2.1.0 ==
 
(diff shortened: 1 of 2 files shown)


samueljseay (approved):
This tests well and is really nice!

Only one suggestion on the copy to make it slightly more explicit:

"You have the WooCommerce Admin plugin activated but it is not being used."

Not a blocker though, thanks for this!

octaedro:
Thanks you @samueljseay for your feedback. I applied the suggested change in the commit c780a6e
I will merge this PR 🙂

PR #6460, opened 2021-02-26
Fix double prefixing of navigation URLs

### LINKED ISSUE #6265: Full URLs get double prefixed with admin url in navigation
**Describe the bug**
In the navigation if you register an item using `admin_url()` the URL ends up getting double prefixed before matching which causes the item to never get matched.

**To Reproduce**

1. Register an item in the new navigation.

```
		Menu::add_plugin_item( array(
			'id'              => 'your-page-id',
			'title'           => __( 'Your page', 'woocommerce-admin' ),
			'url'             => admin_url( 'your-page' ),
		) );
```

2. Use `admin_url( 'your-page' ).
3. Notice the item does not get matched (marked active) since the default match expression includes the prefix twice. `^http:\/\/mysite\.local\/wp\-admin\/http:\/\/mysite\.local\/wp\-admin\/your-page`

**Expected behavior**
Expected the item to be matched when on that page.

joshuatf:
Fixes #6265 

Fixes admin URLs getting double prefixed when a full URL is passed and becoming `http://yourwp.com/wp-admin/http://yourwp.com/wpa-admin/the/path`

1. Register a navigation menu item with a full URL or admin link.
```
	\Automattic\WooCommerce\Admin\Features\Navigation\Menu::add_plugin_item(
		array(
			'id'         => 'my-page',
			'title'      => 'My Page',
			'url'        => admin_url( 'my-page '),
		)
	);
```
2. Enable the navigation.
3. Check that the menu item is marked active when visiting that page.
4. Make sure old menu items are still correctly marked active.


joelclimbsthings (approved):
Tested & looks good! :shipit:

PR #6470, opened 2021-03-01
Dates: Account for leat year in `getLastPeriod`

psealock:
Unit tests weren't passing because `getLastPeriod` wasn't taking into account leap year. This PR makes sure that calculating the last period (ie, last week, last month, last year) always uses `moment().endOf` to calculate the previous period when looking at months.

Take a look at Revenue report for last month against previous year and previous period. Make sure nothing bad happens. Note that the UI looks the exact same because the primary period is used to create the chart's X-axis. Making that accomodate one extra day for the previous period is out of scope of this PR. You can, however, check the requests for data and see that 29th day's data where previously there wasn't.

Screen Shot 2021-03-01 at 3 58 31 PM

Screen Shot 2021-03-01 at 4 28 05 PM

### DIFF
--- a/packages/date/CHANGELOG.md
+++ b/packages/date/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+-   Take into account leap year in calculating `getLastPeriod`.
+
 # 2.1.0
 
 -   Update to @wordpress/eslint coding standards.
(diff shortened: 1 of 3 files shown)


ilyasfoo (approved):
Tests well! 🚢 

>Making that accomodate one extra day for the previous period is out of scope of this PR

I'll open a follow up issue for this later.

PR #6475, opened 2021-03-02
Add a new revenue option, 'rather not say'

### LINKED ISSUE #6459: Add option for "I'd rather not say" for revenue in business details step
In order to prevent false answers, we should add another option to the revenue dropdown for people who may be selling elsewhere but don't want to share their revenue numbers. Otherwise, we may see people selecting 0 when that's incorrect or going back to select "No" in the previous step in order to skip this question.

Frame 2057

My hunch is that there is a significant number of people who find this question uncomfortable and providing this additional option will give them more confidence to complete the onboarding without being forced to share private information.

@pmcpinto

samueljseay:
Fixes #6459 

Reasoning outlined in the ticket, I think this is a great addition to the profile wizard. I did a thorough check around the use of `revenue` and the value is not checked or relied on anywhere that I can see, but I would appreciate any second opinions on this.

### Accessibility

n/a

-   Run through the onboarding wizard up to the business details step
-   For "currently selling" choose "yes, on another platform"
-   The revenue dropdown should appear, open it and confirm the "I'd rather not say" option is present.
-  Choose that option and any others needed for the form to be valid and click continue
-  Confirm no errors are reported. go back from previous section to business details and confirm the option was persisted.

The value stored for this is `rather-not-say`

This feels a bit too minor to worry about testing instructions? Happy to hear other opinions on that.

### DIFF
--- a/TESTING-INSTRUCTIONS.md
+++ b/TESTING-INSTRUCTIONS.md
@@ -11,11 +11,24 @@
 
 ### Deprecate Onboarding::has_woocommerce_support #6401
 
-- Clear existing site transients. For example, by using the [Transients Manager](https://wordpress.org/plugins/transients-manager/) plugin, and pressing the "Delete all transients" button it provides.
-- Add any new theme to WordPress but **DO NOT** activate it.
-- Initialize the Onboarding Wizard.
-- See that the Themes step loads fast 😎 
-- See that the new theme is listed in the Themes step.
+-   Clear existing site transients. For example, by using the [Transients Manager](https://wordpress.org/plugins/transients-manager/) plugin, and pressing the "Delete all transients" button it provides.
+-   Add any new theme to WordPress but **DO NOT** activate it.
+-   Initialize the Onboarding Wizard.
+-   See that the Themes step loads fast 😎
+-   See that the new theme is listed in the Themes step.
+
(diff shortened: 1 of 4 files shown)


TESTING-INSTRUCTIONS.md
```diff
@@ -269,10 +282,19 @@ localStorage.setItem( 'debug', 'wc-admin:tracks' );
         -   The enable/disable button should be correctly reflected in the Woocommerce payment settings screen as well.
 
 Once everything is configured and enabled do a client test
+<<<<<<< HEAD
```
louwie17, line 285: It appears your auto formatter got to this file, and it somehow managed to get some merge conflicts in

louwie17 (approved):
LGTM 🚀 
Just fix the merge conflicts in the testing instructions file.
I would say this would be a great place for an E2E test, or unit test.

I am also indifferent on the testing instructions as it is a pretty minimal change.

TESTING-INSTRUCTIONS.md
samueljseay: Whoops thanks.

samueljseay: Fixed.

samueljseay:
Thanks @louwie17 ! I agree some tests would be good. I'll look to do those in a follow up.

PR #6477, opened 2021-03-02
Add typescript to Nav utils

psealock:
Add TypeScript to Navigation Utils

1. See tests pass
2. See no regressions in Navigation

### DIFF
--- a/client/navigation/utils.ts
+++ b/client/navigation/utils.ts
@@ -15,21 +15,6 @@ interface Item {
 	isCategory: boolean;
 }
 
-interface Category {
-	matchExpression: string;
-	url: string;
-	order: number;
-	title: string;
-	parent: string;
-	menuId: 'primary' | 'favorites' | 'plugins' | 'secondary';
-	capability: string;
-	isCategory: boolean;
-	primary?: Array< Item >;
-	favorites?: Array< Item >;
-	plugins?: Array< Item >;
-	secondary?: Array< Item >;
-}
-
 /**
  * Get the full URL if a relative path is passed.
  */
@@ -167,10 +152,26 @@ export const getMatchingItem = ( items: Array< Item > ): Item | null => {
  */
 export const menuIds = [ 'primary', 'favorites', 'plugins', 'secondary' ];
 
+interface Category {
+	id: string;
+	isCategory: boolean;
+	menuId: 'primary' | 'favorites' | 'plugins' | 'secondary';
+	migrate: boolean;
+	order: number;
+	parent: string;
+	title: string;
+	primary?: Array< Item >;
+	favorites?: Array< Item >;
+	plugins?: Array< Item >;
+	secondary?: Array< Item >;
+}
+
 /**
  * Default categories for the menu.
  */
-export const defaultCategories = {
+export const defaultCategories: {
+	[ key: string ]: Category;
+} = {
 	woocommerce: {
 		id: 'woocommerce',
 		isCategory: true,
--- a/client/navigation/utils.ts
+++ b/client/navigation/utils.ts
@@ -300,13 +300,3 @@ export const getMappedItemsCategories = (
 		categories,
 	};
 };
-
-const assign = (
-	obj: {
-		[ key: string ]: Item;
-	},
-	item: Item
-) => {
-	obj[ item.title ] = item;
-	return obj;
-};
--- a/client/navigation/utils.ts
+++ b/client/navigation/utils.ts
@@ -15,6 +15,21 @@ interface Item {
 	isCategory: boolean;
 }
 
+interface Category {
+	matchExpression: string;
+	url: string;
+	order: number;
+	title: string;
+	parent: string;
+	menuId: 'primary' | 'favorites' | 'plugins' | 'secondary';
+	capability: string;
+	isCategory: boolean;
+	primary?: Array< Item >;
+	favorites?: Array< Item >;
+	plugins?: Array< Item >;
+	secondary?: Array< Item >;
+}
+
 /**
  * Get the full URL if a relative path is passed.
  */
@@ -180,21 +195,6 @@ export const sortMenuItems = ( menuItems: Array< Item > ): Array< Item > => {
 		return a.order - b.order;
 	} );
 };
-
-interface Category {
-	matchExpression: string;
-	url: string;
-	order: number;
-	title: string;
-	parent: string;
-	menuId: 'primary' | 'favorites' | 'plugins' | 'secondary';
-	capability: string;
-	isCategory: boolean;
-	primary?: Array< Item >;
-	favorites?: Array< Item >;
-	plugins?: Array< Item >;
-	secondary?: Array< Item >;
-}
 
 /**
  * Get a flat tree structure of all Categories and thier children grouped by menuId
--- a/client/navigation/utils.ts
+++ b/client/navigation/utils.ts
@@ -217,11 +217,15 @@ export const getMappedItemsCategories = (
 	menuItems: Item[],
 	currentUserCan: ( capability: string ) => boolean | undefined
 ): {
-	items: Item[];
-	categories: Array< Category >;
+	items: {
+		[ key: string ]: Category | { [ key: string ]: Item[] };
+	};
+	categories: {
+		[ key: string ]: Category | Item;
+	};
 } => {
 	const categories: {
-		[ key: string ]: Category;
+		[ key: string ]: Category | Item;
 	} = { ...defaultCategories };
 
 	const items = sortMenuItems( menuItems ).reduce(
--- a/client/navigation/utils.ts
+++ b/client/navigation/utils.ts
@@ -42,6 +42,21 @@ export const getFullUrl = ( url: string ): string => {
 };
 
 /**
+ * Get a default regex expression to match the path and provided params.
+ */
+export const getDefaultMatchExpression = ( url: string ): string => {
+	const escapedUrl = url.replace( /[-\/\\^$*+?.()|[\]{}]/gi, '\\$&' );
+	const [ path, args, hash ] = escapedUrl.split( /\\\?|#/ );
+	const hashExpression = hash ? `(.*#${ hash }$)` : '';
+	const argsExpression = args
+		? args.split( '&' ).reduce( ( acc, param ) => {
+				return `${ acc }(?=.*[?|&]${ param }(&|$|#))`;
+		  }, '' )
+		: '';
+	return '^' + path + argsExpression + hashExpression;
+};
+
+/**
  * Get a match score for a menu item given a location.
  */
 export const getMatchScore = (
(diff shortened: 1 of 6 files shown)


client/navigation/utils.ts
```diff
@@ -67,20 +72,33 @@ export const getDefaultMatchExpression = ( url ) => {
...
+}
+
 /**
  * Adds a listener that runs on history change.
  *
  * @param {Function} listener Listener to add on history change.
  * @return {Function} Function to remove listeners.
  */
-export const addHistoryListener = ( listener ) => {
+export const addHistoryListener = ( listener: Function ) => {
```
samueljseay, line 94: Just giving some suggestions: this could be even more useful if you type the arguments to the callback. Maybe you left this unclear for now because you're not certain about the type which is all good, but you can do something like:

`listener: (someArgument: SomeType ) => void`

client/navigation/utils.ts
```diff
@@ -177,45 +210,65 @@ export const sortMenuItems = ( menuItems ) => {
  * @return {Object} Mapped menu items and categories.
  */
 export const getMappedItemsCategories = (
-	menuItems,
-	currentUserCan = null
+	menuItems: Array< Item >,
+	currentUserCan: Function | null = null
```
samueljseay, line 214: Another situation where a type sig for the function would probably be quite helpful.

if it makes the type declaration too long then you can just do something like:

`type CurrentUserCallback = (someArg: SomeType) => void`

`currentUserCan?: CurrentUserCallback`

client/navigation/utils.ts
```diff
@@ -159,7 +177,7 @@ export const defaultCategories = {
  * @param {Array} menuItems Array of menu items.
  * @return {Array} Sorted menu items.
  */
-export const sortMenuItems = ( menuItems ) => {
+export const sortMenuItems = ( menuItems: Array< Item > ): Array< Item > => {
```
louwie17, line 180: Not sure if the linter forced you to use `Array< Item >` but another way is to use `Item[]` to represent an array of Items. I have a slight preference to the latter, but I guess both would work.

client/navigation/utils.ts
```diff
@@ -67,20 +72,33 @@ export const getDefaultMatchExpression = ( url ) => {
...
-export const addHistoryListener = ( listener ) => {
+export const addHistoryListener = ( listener: Function ) => {
 	// Monkey patch pushState to allow trigger the pushstate event listener.
 	if ( ! window.wcNavigation.historyPatched ) {
 		( ( history ) => {
 			/* global CustomEvent */
 			const pushState = history.pushState;
 			const replaceState = history.replaceState;
-			history.pushState = function ( state ) {
+			history.pushState = function ( state: object ) {
```
louwie17, line 101: I believe `object` types out to be:
```
{ [key: string]: any }
```
Which is in a way a sneaky way to use `any` ;) Would you know at-least some of the properties here that you could create an interface for it? Even if all the key's are optional for now: `?:`

client/navigation/utils.ts
```diff
@@ -177,45 +210,65 @@ export const sortMenuItems = ( menuItems ) => {
...
-	const items = sortMenuItems( menuItems ).reduce( ( acc, item ) => {
-		// Set up the category if it doesn't yet exist.
-		if ( ! acc[ item.parent ] ) {
-			acc[ item.parent ] = {};
-			menuIds.forEach( ( menuId ) => {
-				acc[ item.parent ][ menuId ] = [];
-			} );
-		}
+	const categories: {
+		[ key: string ]: Category | object;
```
louwie17, line 217: Similar as above, what is object here? or is it `{ [ key: string ]: Array< Item > }` as used on line 223?

(sorry not sorry if this is too much information haha)
One thing about `|` when not used as a enum, is that when parsing it you might have to cast it to that type (you might of already did this somewhere). You can do that by using `(<Type> item)` or the `as` keyword -> `item as Category`.
One cool thing, which might be handy for handling two different types is the ` is ` [keyword](https://www.typescriptlang.org/docs/handbook/advanced-types.html), especially for requests where you either get an Error back or the data.
```
function isError(data: MenuItems | Error): data is Error {
  return (data as Error).error !== undefined;
}
```
If we use it in a `if` statement, as `if (isError(data))` the else state will automatically set the type to `MenuItems` not having to do any casting.

psealock:
Thanks for the suggestions @samueljseay @louwie17 ! That was helpful.

I've got the types working besides the `addHistoryListener` function, which I will try to figure out tomorrow.

VSC linter doesn't like this import even though I updated `tsconfig.json` to reflect the webpack alias. Does that change work in your setup @samueljseay? The `npm start` log doesn't complain about it, so it is working.

```js
 import { getAdminLink } from '@woocommerce/wc-admin-settings';
```

client/navigation/utils.ts
```diff
@@ -43,50 +82,69 @@ export const getMatchScore = ( location, itemUrl, itemExpression = null ) => {
...
+}
 
 /**
  * Adds a listener that runs on history change.
  *
  * @param {Function} listener Listener to add on history change.
  * @return {Function} Function to remove listeners.
  */
-export const addHistoryListener = ( listener ) => {
+export const addHistoryListener = ( listener: () => void ): ( () => void ) => {
```
psealock, line 117: @joelclimbsthings I had to mess around with this function and tried to find your PR that introduced this code so I could check against the test instructions. I can't find it because this code was moved over from the old Navigation repo. Do you remember what this code is supposed to do? I think it was so that persisted queries remained intact when navigating around, including Home screen. Does that ring a bell?

joelclimbsthings: Indeed, very large ones! Although that function, `addhistoryListener()` preexisted that, here's the PR for the persisted queries: https://github.com/woocommerce/woocommerce-admin/pull/5883

I also used it for the nav folding behavior as well: https://github.com/woocommerce/woocommerce-admin/pull/5578.

psealock: Thanks @joelclimbsthings

https://github.com/woocommerce/woocommerce-admin/pull/5578 is working, and https://github.com/woocommerce/woocommerce-admin/pull/5883 is testing the same as `main`. Queries persist from report to report, but if you navigate to Homescreen, they don't persist. This is with the new Nav. Using it without Nav going to Homescreen keeps the queries when you go back to a report.

Is this a known thing? I can't remember if we accepted this as fact or if it ever worked.

joelclimbsthings: Ah, yes--that should work as you expected. Looks like something 💔 . I'll create a new issue and try to add a test, but shouldn't hold up this PR.

psealock:
@samueljseay I added typescript support to jest in https://github.com/woocommerce/woocommerce-admin/pull/6477/commits/6a41d74e974cceb6378b6678168549092a1d3ff5 using instructions found in https://jestjs.io/docs/getting-started#using-typescript so that JS tests would pass.

client/navigation/utils.ts
psealock: Thanks @joelclimbsthings !

client/navigation/utils.ts
```diff
@@ -3,31 +3,70 @@
...
+	menuId: MenuId;
+	migrate: boolean;
+	order: number;
+	parent: string;
+	title: string;
+	primary?: Item[];
+	favorites?: Item[];
+	plugins?: Item[];
+	secondary?: Item[];
+}
```
samueljseay, line 32: These definitions look great 🌮

client/navigation/utils.ts
```diff
@@ -3,31 +3,70 @@
...
  */
-export const getMatchScore = ( location, itemUrl, itemExpression = null ) => {
+export const getMatchScore = (
+	location: Location,
+	itemUrl: string,
+	itemExpression: string | null = null
+): number => {
 	if ( ! itemUrl ) {
-		return;
+		return 0;
```
samueljseay, line 69: Whoop, type safety 🎉

client/navigation/utils.ts
```diff
@@ -43,50 +82,69 @@ export const getMatchScore = ( location, itemUrl, itemExpression = null ) => {
...
+}
+
+declare global {
+	interface Window {
+		wcNavigation: wcNavigation;
+	}
+}
+
+interface wcNavigation {
+	menuItems: Array< Item >;
```
samueljseay, line 99: This could use the more succinct `Item[]` not a blocker though.

client/navigation/utils.ts
```diff
@@ -43,50 +82,69 @@ export const getMatchScore = ( location, itemUrl, itemExpression = null ) => {
...
 			/* global CustomEvent */
 			const pushState = history.pushState;
 			const replaceState = history.replaceState;
-			history.pushState = function ( state ) {
-				const pushStateEvent = new CustomEvent( 'pushstate', {
-					state,
-				} );
+			history.pushState = function (
+				state: {
+					[ key: string ]: string;
```
samueljseay, line 126: Not a blocker but there is a built in type that can be useful for this: `Record<string, string>`

client/navigation/utils.ts
```diff
@@ -171,42 +250,64 @@ export const sortMenuItems = ( menuItems ) => {
...
-			acc[ item.parent ] = {};
-			menuIds.forEach( ( menuId ) => {
-				acc[ item.parent ][ menuId ] = [];
-			} );
-		}
+	menuItems: Item[],
+	currentUserCan: ( capability: string ) => boolean
+): {
+	items: {
+		[ key: string ]: Category | { [ key: string ]: Item[] };
```
samueljseay, line 257: This could be a lot less wordy as `Record<string, Category> | Record<string, Item[]>`

client/navigation/utils.ts
```diff
@@ -171,42 +250,64 @@ export const sortMenuItems = ( menuItems ) => {
...
+	} = { ...defaultCategories };
 
-		// Incorrect menu ID.
-		if ( ! acc[ item.parent ][ item.menuId ] ) {
-			return acc;
-		}
+	const items = sortMenuItems( menuItems ).reduce(
+		(
+			acc: {
+				[ key: string ]: Category | { [ key: string ]: Item[] };
```
samueljseay, line 270: Again here, Record could make this easier to read I think.

samueljseay (approved):
Just a few suggestions from me, but pre-approving. Awesome stuff thanks so much for moving the TS adoption along

psealock:
@samueljseay Do you know how to get TS linting to respect our `.eslintrc.js`?

This is failing `npm run lint:js` with `Unable to resolve path to module '@woocommerce/wc-admin-settings'`

```js
/**
 * External dependencies
 */
import { getAdminLink } from '@woocommerce/wc-admin-settings';
```
I have added the webpack alias to the TS config

https://github.com/woocommerce/woocommerce-admin/blob/657e8f90cd43edafafa38f14b367d3377be64b8a/tsconfig.json#L27-L29

but the linter doesn't use that config when checking `.ts` files, nor does it respect `.eslintrc.js` which changes `import/no-unresolved` to a warning

https://github.com/woocommerce/woocommerce-admin/blob/4d16ef4f648ca48c9ecdec94e6e44db461865969/.eslintrc.js#L15

I'm stumped on this

samueljseay:
@psealock yeah I had to do a fair bit of adjusting the eslintrc to fix this stuff in my recent PR. I've also added rule overrides to fix the problems you're mentioning here so you could use these changes in your branch:

https://github.com/woocommerce/woocommerce-admin/blob/637769f7d3d022ebe28ac2b38691a75fd2e4d99c/.eslintrc.js

PR #6484, opened 2021-03-02
Try playwright for e2e tests

### LINKED ISSUE #6313: Make the E2E test suite usable
At the moment the e2e test suite is not fulfilling its purpose, I see several problems:

1. We can't reset DB state between tests meaning its not easy to test multiple similar scenarios as they'll share state of the DB
2. Tests are cumbersome to write and covering a new feature takes a long time
3. We don't test different wordpress version combinations
4. Its not easy to debug failing tests on CI

My proposal is to do several things to improve this:

- [x]  Explore using playwright over puppeteer. The framework has auto waits, video recording and a few other nice features. We could go in on it now before we write too many more tests in puppeteer.

- [x]  Implement a page object pattern for writing new tests. This is a pattern I've used to success on other e2e suites in the past, create objects that represent pages, with methods for specific page interactions. Over time this will (in theory) decrease the burden on writing a new e2e test as you'll have access to many other shared interactions that you can use.

- [ ] Implement resetting of the db between test runs, probably one of the most important features, this is really necessary to write isolated e2e tests.

- [x] If playwright does turn out to be a good choice then implement saving of video artifacts in CI when it fails so we can see whats going on. I haven't done due diligence to see how viable it is to do this on Github actions so some exploration is needed.

louwie17:
Partially fixes #6313 

Update the e2e tests to use [Playwright](https://playwright.dev/) and convert tests to Typescript.
Also added models directory with classes used as page objects.

### Screenshots

Short video with the playwright inspector -> https://d.pr/v/gKZnts

Screen Shot 2021-03-02 at 2 50 14 PM

- Load this PR and run `npm install`
- Run `npm run docker:up`
- Run `npm run test:e2e`, it should run all the tests and pass
- Run `npm run test:e2e-dev`, it should show a chromium browser and run the tests (one of the tests might fail the second time around, given we can't reset the DB yet).
- Run `npm run test:e2e tests/e2e/specs/activate-and-setup/setup-wizard.test.ts`, it should just run that test file.
- Try out the Playwright debugger, by adding `await page.pause()` [here](https://github.com/woocommerce/woocommerce-admin/blob/try/6313_playwright_e2e/tests/e2e/specs/activate-and-setup/complete-store-details-section.ts#L36) and running `npm run test:e2e-dev`, it should pause there and show the playwright inspector and allows you to step through the tests -> https://playwright.dev/docs/inspector


bin/e2e-test-integration.js
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env node
```
louwie17, line 1: Copied this and `bin/wait-for-build.sh` from the `@woocommerce/e2e-environment` with some small changes. As it didn't allow for making use of puppeteer instead.

package.json
```diff
@@ -193,15 +195,17 @@
 		"grunt-checktextdomain": "1.0.1",
 		"grunt-wp-i18n": "1.0.3",
 		"husky": "4.3.8",
-		"jest-24.9.0": "npm:jest@24.9.0",
+		"jest": "^25.5.4",
```
louwie17, line 198: No need for a specific locked version like the above (from what I could see).

tests/e2e/models/DropdownField.ts
```diff
@@ -0,0 +1,30 @@
+import { Page } from 'playwright';
```
louwie17, line 1: Created classes for pages and specific components, all under the models folder.

tsconfig.json
```diff
@@ -15,7 +15,7 @@
 		"esModuleInterop": true,
 		"skipLibCheck": false,
 		"module": "esnext",
-		"typeRoots": [ "**/node_modules/@types" ],
+		"typeRoots": [ "./node_modules/@types", "packages/**/node_modules/@types" ],
```
louwie17, line 18: @samueljseay it couldn't find the `@types/jest` when I installed it, my guess was that `**/node_modules/@types` didn't match the root `node_modules`, not sure why, but the above seemed to fix it.

louwie17:
@samueljseay you might like to check this one out, also all the tests are in TypeScript now :)

package.json
samueljseay: This was done by me in a previous issue to address a problem with jest performance. Running jest locally has really bad performance in 25.x. See https://github.com/facebook/jest/issues/9457

That said, I think things have been getting better in 26... I wonder if we could look at using 26 and see how it goes?

tsconfig.json
samueljseay: Seems fine to me :) if it works it works 👍

tests/e2e/models/onboarding/BenefitsSection.ts
```diff
@@ -0,0 +1,23 @@
...
+
+	async isDisplayed() {
+		await this.page.waitForSelector(
+			'.woocommerce-profile-wizard__container.benefits'
+		);
+	}
+
+	async noThanks() {
+		// Wait for "No thanks" button to become active
+		await this.page.waitForSelector( ':text("No thanks")' );
```
samueljseay, line 18: This is something that playwright solves for us and we should not need to use this very often (except when we just want to assert that something was displayed. like the above `isDisplayed()` method seems fine.

but things like `click` and `fill` wait for the selector to be present and so you don't have to do a wait between operating on elements.

tests/e2e/models/onboarding/BusinessSection.ts
```diff
@@ -0,0 +1,68 @@
...
+			await installFeaturesCheckbox?.check();
+		} else {
+			await installFeaturesCheckbox?.uncheck();
+		}
+	}
+
+	async expandRecommendedBusinessFeatures() {
+		const expandButtonSelector =
+			'.woocommerce-admin__business-details__selective-extensions-bundle__expand';
+		await this.page.waitForSelector( expandButtonSelector );
```
samueljseay, line 49: I won't mark every one of these examples, but here is another one, anywhere we interact with something we were waiting on we should remove `waitForSelector`

tests/e2e/models/onboarding/ThemeSection.ts
```diff
@@ -0,0 +1,18 @@
...
+
+	constructor( page: Page ) {
+		this.page = page;
+	}
+
+	async isDisplayed( ) {
+		await this.page.waitForSelector(
+			':text("Choose a theme")'
+		);
+        await this.page.waitForSelector(
```
samueljseay, line 14: Something went awry with the indentation here. Do you have prettier enabled to auto format your ts files? I know Paul had to change a setting to make that happen.

samueljseay (changes requested):
Looking fantastic, I think it would be worthwhile to remove redundant `waitForSelector` calls but otherwise looking forward to these changes.

package.json
louwie17: Gotcha, I will revert this then, that is also rather unfortunate :( 
I saw people still reproducing it in 26 (https://github.com/facebook/jest/issues/9457#issuecomment-736024084) so I will revert this.

louwie17: I might add a comment though, as an FYI

samueljseay (dismissed):
Couple conflicts need addressing but this looks great. Truly grateful for your work to move the needle on e2e here. Lets hope this lowers the barrier to the team 🚀

samueljseay:
I'm gonna block this for now. There has been some pushback on us using playwright and I'd like to resolve that discussion before we move forward

louwie17:
> I'm gonna block this for now. There has been some pushback on us using playwright and I'd like to resolve that discussion before we move forward

Yeah I see that, no worries at all, what I might do is create a separate PR from this one, with most of the refactors and just having Playwright pruned out, this should be a relatively minimal change. Then we at-least get to still use TypeScript and use Page Objects on our end.

louwie17:
Closing this in favour of https://github.com/woocommerce/woocommerce-admin/pull/6582 and because we are holding of on the Playwright switch.

PR #6486, opened 2021-03-02
Remove active item from navigation store

### LINKED ISSUE #6478: Navigation: Remove unused setActiveItem from store
`setActiveItem` is determined by internal state in `<Container />` so the logic that does that in wp.data's navigation store can be removed.

https://github.com/woocommerce/woocommerce-admin/blob/78f28744267faf3c9d09d5cd80ad22619a0b6ca7/packages/data/src/navigation/actions.js#L12

joshuatf:
Fixes #6478

Removes the unused `activeItem` as well as getters and setters from navigation store.

1. Make sure nothing is broken in the nav.


psealock (approved):
Yup, perfect 👨🏽‍🍳 🚢

PR #6498, opened 2021-03-03
Add TypeScript to CustomerFeedbackModal

adrianduffell:
Adds TypeScript to the `CustomerFeedbackModal` component.

Additionally, the `@types/wordpress__components` type definitions were installed. This was added to the package to resolve lint issues, and to the root to resolve build issues. Let me know if this is not the best approach.

### Detailed Testing Instructions

Smoke test the Customer Effort Score feature, for example:

- Update WooCommerce Settings
- See the snackbar appear `How easy was it to update your settings?`
- Press `Give feedback`
- Interact with the modal and send feedback.
- See the modal close and snackbar appear `Thank you for your feedback!`.

No changelog required. This has been added for the package in the fix for #6671.


adrianduffell:
This is a WIP which currently just adds types to the `CustomerFeedbackModal` component.

It has one error remaining presently:

```
Could not find a declaration file for module '@wordpress/components'. '/Users/xx/Code/woocommerce-admin/node_modules/@wordpress/components/build/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/wordpress__components` if it exists or add a new declaration (.d.ts) file containing `declare module '@wordpress/components';`
  ```

All feedback is welcome ❤️

File: packages/customer-effort-score/src/customer-feedback-modal/index.tsx [hunk 1]
```diff
@@ -27,7 +27,7 @@ import { __ } from '@wordpress/i18n';
  * @param {Function} props.recordScoreCallback Function to call when the results are sent.
  * @param {string} props.label                 Question to ask the customer.
  */
-function CustomerFeedbackModal( { recordScoreCallback, label } ) {
+function CustomerFeedbackModal({ recordScoreCallback, label }: { recordScoreCallback: (score: number, comments: string ) => void, label: string } ) {
 	const options = [
```
samueljseay, line 31: If you're getting lint errors maybe check that your prettier plugin is autoformatting files with `ts` or `tsx` extension. I think @psealock  ran into this

samueljseay:
@adrianduffell for the error thats remaining I believe you can just install the types:

`npm i -D @types/wordpress__components`

packages/customer-effort-score/src/customer-feedback-modal/index.tsx
psealock: It was a VSC settings issue. I had the prettier plugin set for js files, so I added it as the default

adrianduffell: Thanks! Looks like I had the same issue with prettier ignoring these files.

adrianduffell:
This has some lint & e2e errors which I suspect will be resolved once https://github.com/woocommerce/woocommerce-admin/pull/6595 is merged.

I'll pause for now.

becdetat (approved):
Looks great and everything is still working as expected

PR #6505, opened 2021-03-03
Adding client-side filter to navigation root url

joelclimbsthings:
Fixes https://github.com/Automattic/wc-calypso-bridge/pull/659

Adding client-side filter for navigation `rootBackUrl` to support a fix for [this issue in the bridge](https://github.com/Automattic/wc-calypso-bridge/issues/640).

-   Checkout branch
-   Add a snippet of client-side code like this:
```
addFilter( 'woocommerce_navigation_root_back_url', 'plugin-domain', () => 'https://www.gezorad.com');
```
- Navigate to any WooCommerce page with the new navigation, and then click the "WordPress Dashboard" button.
- Confirm that you're taken to the altered URL


psealock (approved):
Nice @joelclimbsthings, this is looking great. 

Do you think the equivalent PHP filter makes sense to keep?

joelclimbsthings:
> 
> Do you think the equivalent PHP filter makes sense to keep?

Good question, I'd image we're the only ones that have used it thus far, so likely safe to remove? I'm not sure if there would be the potential for 3pd to want to change this on the back-end side as opposed to the client. If not, I'd say let's remove it.

joshuatf (commented):
+1 for removing the server-side filter.  I think having multiple filters adds to the confusion and it's pretty likely that a use case needing to modify this is running a more modern JS environment.

Should we also move `rootBackLabel` here as well?  I feel these two are tied together.

joelclimbsthings:
Agreed. Just moved `rootBackLabel` to client-side, along with filter, and removed the server-side filter for `rootBackUrl`.

joshuatf (approved):
Nice!  Changes look good ✅

PR #6513, opened 2021-03-04
add preview site btn in personalize store task

### LINKED ISSUE #6457: Use Gutenberg component to preview site in Personalize store task
We know that store owners struggle to access the front end of their site during store setup. This has been a common theme in qualitative feedback we've received. In the Personalize your store task, an item in the stepper is to "Create a custom homepage" but it's not clear what's customized or what it even looks like. [People have described this step as "misleading".](https://github.com/woocommerce/woocommerce-admin/issues/4239)

To provide an easier way to check the front end after selecting the "Personalize your store" task, we can either provide a link out to their site building tool or directly to the front end. The challenge with providing a direct connection to a customizer is that we don't know exactly what tool people are using to customize the front end of the site. The challenge with providing a direct link to the front end is that we don't know exactly when the user will want to check the front end during this task - after all the steps, or after each step?

One solution given all of these unknowns is to provide a handy link to preview the front end of a site in a new tab. Given that this exact component exists in Gutenberg, we could benefit by reusing it for this purpose.

A new link to `Preview site` would appear in the top bar of only the Personalize your store step.
image

As the user continues through the task list steps, they can open their site in a new tab and keep checking progress.
image

Component in Gutenberg:
Screen Shot 2021-02-20 at 8 15 00 PM

I like this solution because it solves a real known problem _and_ the component exists already.

cc @timmyc @pmcpinto

moon0326:
Fixes #6457 

This PR adds a new button in the personalize store task that opens the frontend for preview. 

### Screenshots
Screen Shot 2021-03-03 at 8 38 11 PM

1. Navigate to Home and click "Personalzie your store" task.
2. Click on the "Preview Site" button on the header.
3. A new tab should open and the URL should be the site URL.
4. Navigate to other tasks such as "Store Details" or "Add products".
5. The "Preview Site" should not be shown on the other tasks.

### DIFF
--- a/TESTING-INSTRUCTIONS.md
+++ b/TESTING-INSTRUCTIONS.md
@@ -9,6 +9,14 @@
 3. Note that all the reports exist and navigating to those reports works as expected.
 4. Check that report menu items are marked active when navigating to that page.
 
+### Add preview site button on the appearance task #6457
+
+1. Navigate to Home and click "Personalzie your store" task.
+2. Click on the "Preview Site" button on the header.
+3. A new tab should open and the URL should be the site URL.
+4. Navigate to other tasks such as "Store Details" or "Add products" .
+5. The "Preview Site" should not be shown on the other tasks.
+
 ### Fix double prefixing of navigation URLs #6460
 
 1. Register a navigation menu item with a full URL or admin link.
--- a/client/header/activity-panel/index.js
+++ b/client/header/activity-panel/index.js
@@ -183,7 +183,7 @@ export const ActivityPanel = ( { isEmbedded, query, userPreferencesData } ) => {
 
 		const previewSite = {
 			name: 'previewSite',
-			title: __( 'Preview Site', 'woocommerce-admin' ),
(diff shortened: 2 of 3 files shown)


elizaan36:
Works great for me @moon0326 ! 🎉
One small thing is that we're using sentence case so it would be "Preview site" where the "s" is lowercase rather than capitalized.

I can see on the main branch that the Display link is removed, so looks like that's been taken care of already!

moon0326:
> Works great for me @moon0326 ! 🎉
> One small thing is that we're using sentence case so it would be "Preview site" where the "s" is lowercase rather than capitalized.
> 
> I can see on the main branch that the Display link is removed, so looks like that's been taken care of already!

Thank you for testing it 👍 I made update in https://github.com/woocommerce/woocommerce-admin/pull/6513/commits/e4d1ace3561c2fae93eaa14538eff20346773ede

becdetat (approved):
Looks good, everything tests well, and the code makes good sense :shipit:

PR #6518, opened 2021-03-04
Add nav intro modal tests

joshuatf:
- Removes unncessary tracking opt-in check (previously used because of remote images)
- Adds tests around the nav intro modal.

1. Make sure tests are ✅


client/navigation/components/intro-modal/test/index.js
```diff
@@ -0,0 +1,89 @@
...
+		useDispatch.mockImplementation( () => ( {
+			updateOptions,
+		} ) );
+
+		const { container } = render( <IntroModal /> );
+
+		expect( container ).toBeEmptyDOMElement();
+		expect( updateOptions ).toHaveBeenCalledWith( {
+			[ INTRO_MODAL_DISMISSED_OPTION_NAME ]: 'yes',
+		} );
```
psealock, line 52: If the modal has already been shown, `dismissModal` is called. Am I understanding correctly that `dismissModal` is called on every screen, on every page refresh? If so, that function calls not only `updateOptions` but `recordEvent` as well. Is it worth revisiting this logic?

psealock (commented):
Thanks for the coverage @joshuatf , tests are looking fantastic.

I left a comment of some functionality that may need tweaking

client/navigation/components/intro-modal/test/index.js
joshuatf: The dismissal is only called when the welcome modal is shown simultaneously and only on the homescreen where the components have been mounted.

I think it's pretty unlikely that it gets called multiple times, but just in case, I added https://github.com/woocommerce/woocommerce-admin/pull/6518/commits/df134316909f500072746ffa33b034af7024b5b7 as a backup measure.

psealock: Ok gotcha, thanks

psealock (approved):
:shipit:

PR #6520, opened 2021-03-05
Remove rest API props from onboarding confirmed as no longer used.

samueljseay:
I've chatted with a few people to confirm this, we're not using these REST API props any more so now they can be removed.

The only risk I was really concerned with was if these props existed already in a user's onboarding profile and we tried to wholesale update a users profile. As far as I can tell this is not an issue, we always update only a select few props at a time and none of our existing code touches these properties.


louwie17 (changes requested):
Nice find @samueljseay, I couldn't find any use cases of it when it was originally added either. I assume it was added as a potential "nice to have".
Both of these params are being checked for in the `onboarding-profile.php` test, if you could fix that :)

louwie17 (approved):
Thanks for fixing the unit tests, LGTM 🚀

PR #6521, opened 2021-03-05
Fix a bug where the JP connection flow would not activate

### LINKED ISSUE #6488: Business features step: fix JP flow
**Description:** the experiment implemented in #5287 showed positive results. Recap here: pbIJXs-GS-p2. We need to fix a bug as we're not redirecting users to the Jetpack log in/account creation flow as expected in this flow

**Acceptance criteria:**
- [ ] If the user installs Jetpack ~~and WooCommerce Shipping & Tax~~, he should be redirected to the Jetpack sign up/log in flow at the end of the profiler. As we have now with the original flow

samueljseay:
Fixes #6488 

Changes.
* See the issue for the discussion, but we're changing the behaviour of triggering the connection flow as part of this. Before it was required that both wc-services and wc-admin were installed before triggering the flow. 

* Also the `plugins` value was not being updated properly in the new flow, but `plugins` is not used in any useful way in the onboarding profile, so I have removed it completely. We now rely on the actual installed plugins to determine what is installed/activated.

### Accessibility

n/a

### Screenshots

n/a

- With a fresh site start the profile wizard
- In store details ensure that US is selected as country
- In industry ensure that "food and drink" is selected
- When you get to the business details step you should see the 2 tab flow of "business details" and "free features". Fill out business details with whatever values you like
- in "free features" choose to disable all the extensions for installation *except* Jetpack.
- once it installs it should go to theme step, continue with current theme from the theme step
- You should arrive on the Jetpack setup connection page.


client/profile-wizard/index.js
```diff
@@ -85,20 +83,6 @@ class ProfileWizard extends Component {
...
-			activePlugins.includes( 'woocommerce-services' ) &&
-			activePlugins.includes( 'jetpack' ) &&
-			profileItems.plugins !== 'already-installed'
-		) {
-			recordEvent(
-				'wcadmin_storeprofiler_already_installed_plugins',
-				{}
-			);
-
-			updateProfileItems( { plugins: 'already-installed' } );
```
samueljseay, line 100: This code was never happening due to profileItems not being loaded. I've confirmed also that we no longer care about `wcadmin_storeprofiler_already_installed_plugins`

client/profile-wizard/index.js
```diff
@@ -221,18 +205,14 @@ class ProfileWizard extends Component {
...
 			connectToJetpack,
 		} = this.props;
 		recordEvent( 'storeprofiler_complete' );
 
-		const { plugins } = profileItems;
 		const shouldConnectJetpack =
-			( plugins === 'installed' || plugins === 'installed-wcs' ) &&
-			activePlugins.includes( 'jetpack' ) &&
-			! isJetpackConnected;
+			activePlugins.includes( 'jetpack' ) && ! isJetpackConnected;
```
samueljseay, line 215: Now we just check if jetpack is installed and not connected. this also means we'll get users we didn't get before if they installed jetpack before doing the OBW. I've checked and this behaviour change is ok.

client/profile-wizard/index.js
```diff
@@ -278,11 +258,29 @@ class ProfileWizard extends Component {
...
+		} = this.props;
+
+		// Track plugins if already installed.
+		if (
+			profileItemsLoaded &&
+			profileItems.plugins !== 'installed' &&
+			activePlugins.includes( 'woocommerce-services' ) &&
+			activePlugins.includes( 'jetpack' )
+		) {
+			updateProfileItems( { plugins: 'installed' } );
```
samueljseay, line 276: The goal here was just to make sure `plugins` gets set to something reasonable. Before it was trying to be set to `already-installed` but as mentioned that didn't work and wasn't very relevant anyway.

client/profile-wizard/index.js
```diff
@@ -278,11 +258,29 @@ class ProfileWizard extends Component {
...
+		const {
+			activePlugins,
+			profileItems,
+			updateProfileItems,
+			profileItemsLoaded,
+		} = this.props;
+
+		// Track plugins if already installed.
+		if (
+			profileItemsLoaded &&
```
louwie17, line 271: Could you add a `isUpdatingProfileItems` check here as well, or have the `updateProfileItems` update optimistically.
I encountered an "almost" infinite loop of triggering a profile items update (see Gif), it did eventually stop.
To reproduce, I went through your initial testing instructions, and then later through the wizard again adding **Woocommerce Shipping** this time, so it would hit the `if`statement below.
You can also reproduce this on a fresh install and selecting both jetpack and woo shipping (woo services)
plugin-installed

louwie17 (dismissed):
@samueljseay following your testing instructions it tested well, although I hit a semi unnoticed bug when also installing Woocommerce services at the same time (for the user it would just make the theme pages very sluggish for a bit) - See comment and gif.

client/profile-wizard/index.js
samueljseay: That makes sense this would happen, thanks for catching this 🤔 I think probably checking that updating is in progress makes the most sense.

samueljseay: Talked to some other team members, I think we can remove `plugins` altogether. I'll update the PR and description accordingly.

.github/workflows/test-php.yml
```diff
@@ -7,7 +7,7 @@ jobs:
             WP_CORE_DIR: '/tmp/wordpress'
             COMPOSER_DEV: 1
 
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-18.04
```
samueljseay, line 10: Recently github have been migrating the default for ubuntu-latest to 20.04, which I think is why our tests have broken. I've pinned to 18.04 till we can migrate across. I think its just our approach with mysql that needs some work.

client/profile-wizard/index.js
samueljseay: Removed this based on discussion.

adrianduffell (approved):
This tested well, thanks for working on the changes today 🚀

PR #6525, opened 2021-03-05
Use wc filter to get status tabs for tools category

### LINKED ISSUE #6524: WooCommerce status tabs not migrated to new navigation
**Describe the bug**
Any tabs registered with the `woocommerce_admin_status_tabs` hook are not migrated.

Currently the menu items are explicitly registered since this hook did not give us access to the built-in tabs.  We should investigate how plugins are using this hook and if it's still possible to collect the tabs for automatic registration in the menu.

**To Reproduce**

1. Register a custom tab with the `woocommerce_admin_status_tabs` filter.
2. Enable the new navigation.
3. Note the tab is not added.

**Expected behavior**
It would be nice if we could automatically add the new tab.

joshuatf:
Fixes #6524 

Uses the status filter to get registered tabs for the tools menu.

Note that this requires copying over the default tabs, but I don't think this is a worse solution than what we had since we were manually adding all these items anyway.

### Screenshots
Screen Shot 2021-03-05 at 9 34 42 AM

1. Register a new tab via the filter.
```
add_filter( 'woocommerce_admin_status_tabs', function ( array $tabs ) {
	$tabs['my-tools-page'] = __( 'My Tools Page', 'your-text-domain' );
	return $tabs;
} );
```

2. Enable the new navigation.
3. Make sure the menu item for the registered tab is shown under `Tools`.


frosso (commented):
FWIW - this is great, thanks!
Maybe in the future the `woocommerce_admin_status_tabs` can be deprecated in favor or a filter name that is more appropriate (since we're not adding tabs anymore)

joelclimbsthings:
Hey @joshuatf ! Good work. I did notice a couple oddities:

I'm showing an excess listing for "Status" under my extensions menu when activating this branch, which points to the System Status page. 

image


I'm _also_ still showing that listing under the Tools menu as expected, but I'm missing the "Import/Export" link (as compared to `main`):

This branch:
image

On `main`: 
image

TESTING-INSTRUCTIONS.md
```diff
@@ -2,6 +2,17 @@
 
 ## Unreleased
 
+### Use wc filter to get status tabs for tools category #6525
+
+1. Register a new tab via the filter.
```
joelclimbsthings, line 7: :heart:

joelclimbsthings (changes requested):

joshuatf:
Great catch on both fronts, @joelclimbsthings!

Pushed fixes in https://github.com/woocommerce/woocommerce-admin/pull/6525/commits/6cdf7577632a6952b2d525cf1f0d36d87bf4908d

joelclimbsthings (approved):
Great work! Tests well. :ship:

PR #6526, opened 2021-03-05
Change "Store Setup" link in top bar to "Finish setup"

elizaan36:
This is a small tweak to the "Store setup" link in the top bar that only appears when the store setup task list isn't complete.

Screen Shot 2021-03-05 at 10 06 35 AM

Using a verb in "Finish setup" makes the link more actionable. Also, this update uses sentence case rather than title case.

Just a note that I also changed the tests but would be great if someone had a look to make sure everything there is still ok.

cc @pmcpinto 

No changelog required

### DIFF
--- a/client/header/activity-panel/index.js
+++ b/client/header/activity-panel/index.js
@@ -131,7 +131,7 @@ export const ActivityPanel = ( { isEmbedded, query, userPreferencesData } ) => {
 
 		const setup = {
 			name: 'setup',
-			title: __( 'Store Setup', 'woocommerce-admin' ),
+			title: __( 'Finish setup', 'woocommerce-admin' ),
 			icon: <SetupProgress />,
 			onClick: () => {
 				const currentLocation = window.location.href;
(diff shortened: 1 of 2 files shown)


samueljseay (approved):
Looks good to me. Thanks for this! Assuming @pmcpinto is happy then 🚀

samueljseay (changes requested):
~~Sorry I didn't check the e2e tests, It's likely this copy is used in those tests and thats why its failing.~~

I think the e2e failure was unrelated.

samueljseay:
I'll take a look at fixing up the tests here.

samueljseay (approved):
🚀

elizaan36:
Thanks Sam for taking a look. I shared with Pedro in Slack before I created the issue so we're gtg. Not sure how to fix the tests but let me know if there's anything I can do.

joshuatf (approved):
Nice tweak- thanks for this!

PR #6536, opened 2021-03-08
Add abtest package and tracks events to validate

chickenn00dle:
This adds an A/B Test package to be used for WCPay experiments. 

In order to validate the randomness of the groups, I've added the following tracks events:

```
wcadmin_abtest_assign_variant

props:
abtest_name
abtest_group
context

wcadmin_abtest_serve_variant

props:
abtest_name
abtest_group
context
```

We've also set up a test in the payment task page when WCPay is a visible method, but with no variation (control = experiment) just so we can validate our randomization approach. I've added the test here as this is where we plan to run our first experiment (see https://wp.me/paJDYF-15q)

~~A note on storing variations in `wp_options`: Currently, in order to persist the assigned group across browsers/different admins, I am storing this as an option. I have not added any code to clean this up as we will only be running one test at a time and this is something we can manually take care of when finishing up one test/starting the next. This is not an ideal approach, so if there is a better one, let me know!~~

Before testing, it is a good idea to replace the ABTest experiment prop in `client/task-list/tasks/payments/index.js ` with something easily identifiable. So something like 

```
<ABTest
	name ="validate_randomization_mechanism"
	control={ card }
	experiment={ <div>Hello Dolly</div> } <<< This line here.
	onComplete={ this.handleABTestComplete }
	key={ key }
/>
```

-  Go to the [payment task page](https://cloudup.com/cewDGNtBBCR) (making sure your test store is eligible for WC Pay)
- This should assign you a group, then save to both localStorage (`validate_randomization_mechanism` key) and wp_options table (`woocommerce_admin_abtests`) Confirm these groups match and the payment task reflects this group correctly.
- Delete the group in localStorage, then refresh the page. Confirm the same group is stored in localStorage once again and the payment task reflects this.
- Delete the option in wp_options, then refresh the page. Confirm the payment task reflects the correct group. Note, we do not re-save the option in this case.
- Delete both the localStorage key and the option, then refresh the page. Do this a few times to confirm you are assigned a different group than the first time.
- Finally, check [Tracks Live View](https://mc.a8c.com/tracks/live/?eventname=%25wcadmin_abtest%) for associated events and ensure they match what is expected. There should be only 1 assign event for each time the payment task is loaded when a group is not stored in either wp_options or localstorage. There should be as many serve events as you viewed the payment task page.

I've also added ABTest Suite unit tests that will run with `npm run test`

### Changelog

> Add: Add ABTest package.


chickenn00dle:
It looks like there are some failing PHP unit tests. Since this PR only touches JS files, I'll go ahead and ignore those.

packages/abtest/src/utils.js
```diff
@@ -0,0 +1,130 @@
...
+		recordABTestEvent( name, CONTROL, 'fetch_failed' );
+		return CONTROL;
+	}
+};
+
+/**
+ * Randomly select between CONTROL and EXPERIMENT groups.
+ */
+export const getRandomGroup = () =>
+	Math.random() < 0.5 ? CONTROL : EXPERIMENT;
```
psealock, line 96: How about extracting the `0.5` value to a prop on `<ABTest />`? There may be a desire for a conservative approach of showing the experiment to only 10% or something

packages/abtest/src/index.js
```diff
@@ -0,0 +1,76 @@
...
+
+		const cachedGroup = getCachedGroup( name );
+		if ( cachedGroup ) {
+			setVariation( cachedGroup );
+			handleComplete();
+			recordABTestEvent( name, cachedGroup, 'from_cache', 'serve' );
+		} else {
+			( async () => {
+				const group = await getAndSetGroup( name );
+				setVariation( group );
```
psealock, line 62: Should this be `setGroup` to keep it all straight? I see this means the value in state must also be `group`, so maybe the const declared two lines above could be `groupFromDB` or something.

psealock (commented):
Nice work here @chickenn00dle, this is well written code and very thorough.

I tested according to instructions and it did exactly as was expected. I left some minor comments with suggestions for improvements.

>A note on storing variations in wp_options: Currently, in order to persist the assigned group across browsers/different admins, I am storing this as an option. I have not added any code to clean this up as we will only be running one test at a time and this is something we can manually take care of when finishing up one test/starting the next. This is not an ideal approach, so if there is a better one, let me know!

Since you are storing the option name as `${ OPTION_NAME }_${ name }`, you could run multiple tests at a time, no?

The only larger issue to discuss is the bespoke option get/set mechanism instead of using WC Admin's options data store. I see two advantages to using the data store: batching of requests and pre-hydrated options. If two ABTests or multiple UI elements make queries for options on the same screen, the data store batches those requests into one for better performance. There is also functionality to easily query for the option on page load and the value is passed to the store for immediate rendering. This avoids the need for the placeholder you added, but isn't required. 

Thanks for the pull request, I think this will be a valuable contribution

packages/abtest/src/utils.js
chickenn00dle: I had considered this initially but based on our needs for this package, decided not to. Now that you point it out though, it does make more sense to expose this as a prop. I've made the change in https://github.com/woocommerce/woocommerce-admin/pull/6536/commits/adf2d654e07132f35f82b787b67a7fdfb995b3b7

packages/abtest/src/index.js
chickenn00dle: Good call. I made this `variation`/`setVariation` to distinguish between component state and util functions. But looking at this again, it makes more sense to make this consistent. I've renamed these and changed the const you mentioned to `newGroup`: https://github.com/woocommerce/woocommerce-admin/pull/6536/commits/fd9021dcf1c2f4a9e1c16078604abf56171cac64

chickenn00dle:

> this is well written code and very thorough.

❤️

> I left some minor comments with suggestions for improvements.

Thank you! I've made changes to address these comments 👍

> Since you are storing the option name as ${ OPTION_NAME }_${ name }, you could run multiple tests at a time, no?

That's correct. I was referring to these options remaining in the merchant's database after an ABTest is complete. As it stands, we'll need to be sure to periodically clean this up for A/B tests that are no longer running.

> There is also functionality to easily query for the option on page load and the value is passed to the store for immediate rendering. This avoids the need for the placeholder you added, but isn't required.

Awesome! I actually hadn't considered this as an option. I think it's worth making the change to avoid the loading state you pointed out. I'm not too familiar with the WC Admin data package, so will spend some time today looking through and seeing if I can work this out.

chickenn00dle:
So I was able to use WC Admin option store to get/set A/B test options! (https://github.com/woocommerce/woocommerce-admin/pull/6536/commits/b9ff995e2ab4fabb017ae8ab2dcbc60fa2dce2b3)

Since I'm not too familiar with these stores, I'm not super confident in my implementation, so I'm sorry in advance if anything seems strange 🙇 

One thing to note is, now that we are using the options store, I was able to consolidate all A/B test names/groups into one array and store this as a single option (`woocommerce_admin_abtests`), so this should deal with the issue I mentioned earlier of polluting the merchant's database.

There were a few points I wasn't sure about as well:

- I believe I've correctly preloaded the abtests option. I wasn't sure where to perform this, so created a new `ABTest` feature class. If there is a better place to do this, let me know.
- Related to this, I'm assuming there is no need to check for an `isResolving` state when dealing with preloaded options via the options store. At least, it seemed that way while testing, so I didn't add a check for this in the `ABTest` component. Is this correct?
- We're no longer using the placeholders, but I didn't remove the code for this as it seems like it might be useful in the future. If you'd prefer I remove this, let me know.

packages/abtest/src/index.js
```diff
@@ -0,0 +1,91 @@
...
+					name,
+					size,
+					getABTestOption,
+					setABTestOption
+				);
+				setGroup( newGroup );
+				recordABTestEvent( name, newGroup, 'from_store', 'serve' );
+			}
+		}
+		handleComplete();
```
psealock, line 74: If a test has expired past an end date, `handleComplete` will be called every time. Is that the intention? The description `A callback fired after A/B Test selects group` maybe suggests otherwise.

psealock (commented):
>I was able to consolidate all A/B test names/groups into one array and store this as a single option (woocommerce_admin_abtests), so this should deal with the issue I mentioned earlier of polluting the merchant's database. I wasn't able to do this before as multiple A/B Tests would create a race, and only one would end up saving to the db. But with the store, this is no longer a problem!

Nice!

>I believe I've correctly preloaded the abtests option. I wasn't sure where to perform this, so created a new ABTest feature class. If there is a better place to do this, let me know.

yup, that should be good

>I'm assuming there is no need to check for an isResolving state when dealing with preloaded options via the options store. At least, it seemed that way while testing, so I didn't add a check for this in the ABTest component. Is this correct?

I think thats generally safe to assume. The only problem I can see is if the ABTest component gets used in this repo, but outside of the wc-admin context. The data stores get hydrated with the pre loaded options data on application mount, but some scripts are loaded outside this context, such as some scripts that enhance Coupon legacy pages.

https://github.com/woocommerce/woocommerce-admin/blob/b17aab156d9fb6e8fbc2ce59ba34d82fd20d9f5f/client/layout/index.js#L255-L261

I don't think there is a need to re-introduce `isResolving`, but I'd just want to make sure that if the option is fetched, it won't cause fatal errors. You could test this by removing the preloading in PHP, which I haven't done. 

>We're no longer using the placeholders, but I didn't remove the code for this as it seems like it might be useful in the future. If you'd prefer I remove this, let me know.

I think its best to remove it

Glad to see this coming together @chickenn00dle !

packages/abtest/src/index.js
chickenn00dle: Ah. Good catch! 

I think it's best not to fire the cb when the test is not active. I've changed this in https://github.com/woocommerce/woocommerce-admin/pull/6536/commits/fb2796bbc7f0a4100a16e3cf9c2bf16c5a6fce9b

chickenn00dle:
> I'd just want to make sure that if the option is fetched, it won't cause fatal errors. You could test this by removing the preloading in PHP, which I haven't done.

I tested this on my end and there were no fatals. When loading an ABTest without preloaded option, the fetching from the store fails and we fallback to control in the `catch` block in `getAndSetGroup()`

> I think its best to remove it

Gotcha. I've removed this in https://github.com/woocommerce/woocommerce-admin/pull/6536/commits/41d7a210acbc8847422b54509d153109d98dfba9

psealock:
>I tested this on my end and there were no fatals. When loading an ABTest without preloaded option, the fetching from the store fails and we fallback to control in the catch block in getAndSetGroup()

I'm just wondering about a future case where someone uses the ABTest component and forgets that pre loading the option is a requirement. Since this is a package, its possible to be consumed outside of the wc-admin context. In other words, WC Pay could import this package for its own tests and not realise there is an assumption that the option is expected to be there already. The only way they'd know something is wrong is by seeing the control rendered 100% of the time.

I think the component would be better if it could handle both scenarios. I realise I'm sending you in circles now (sorry!), but could `getAndSetGroup` be an async function or avoid calling `setCachedGroup` if the option request `isResolving`?

chickenn00dle:
Hey @psealock

> I think the component would be better if it could handle both scenarios. I realise I'm sending you in circles now (sorry!)

I totally agree. And no worries! 👍 

> could getAndSetGroup be an async function

I went with a slightly different approach in https://github.com/woocommerce/woocommerce-admin/pull/6536/commits/fd567cb959ddf91d883664d175db1a3a0909fca3. I actually wasn't sure how to go about making this async while still using the options store selector. So if you are aware of a cleaner approach, please let me know!

> or avoid calling setCachedGroup if the option request isResolving?

This should also be the case with the changes in https://github.com/woocommerce/woocommerce-admin/pull/6536/commits/fd567cb959ddf91d883664d175db1a3a0909fca3

One note about `isResolving`. I noticed that this didn't always return what I was expecting and so I added a check for `hasFinishedResolution`. I'm also not sure if this is the right approach or if I'm mistaken in my use of `isResolving`.

timmyc:
> Since this is a package, its possible to be consumed outside of the wc-admin context. In other words, WC Pay could import this package for its own tests and not realise there is an assumption that the option is expected to be there already.

It is sounding like we might actually want to explore importing this package and using it right away in the wc-pay plugin, so this path definitely needs to be considered.

packages/abtest/src/utils.js
```diff
@@ -0,0 +1,117 @@
...
+	size,
+	isOptionResolving,
+	getOption,
+	updateOptions
+) => {
+	try {
+		const option = getOption( OPTION_NAME );
+
+		if ( isOptionResolving ) {
+			return null;
```
psealock, line 56: With the preloaded option removed, I can see this functioning as it should. Nothing is rendered until the option is returned, and that delay is only on the initial load. When the value is found in local storage, rendering is immediate 🎉 

I do think one improvement could be the addition of an optional prop on `ABTest` for a loading component. If, for whatever reason, I choose not to pre load the option, it would be nice to be able to supply an alternate component while this loads. The README would also be a good place to add some language describing how this works. What do you think?

psealock (commented):
This is working fantastic. I went through the testing instructions again with and without the option pre loaded and its working as it should.

I have one final suggestion and then I think this component is ready to be merged.

packages/abtest/src/utils.js
chickenn00dle: > I do think one improvement could be the addition of an optional prop on ABTest for a loading component. If, for whatever reason, I choose not to pre load the option, it would be nice to be able to supply an alternate component while this loads.

This was actually one of the primary reasons for adding the `onComplete` prop. It allows the parent to deal with any loading state as this may not always mean rendering a temporary component. Using a callback also removes some complexity from the package.

I can also see the benefit of explicitly passing a `loading` prop and moving the logic for this into the ABTest component though, and so I'm fine with either approach.

Let me know if you prefer one over the other and I'll go ahead and make the change, updating the README as well.

chickenn00dle: I've gone ahead and added a section to the README showing how to deal with a loading state using the `onComplete` callback. If you prefer an explicit `loading` prop, let me know!

TESTING-INSTRUCTIONS.md
```diff
@@ -2,6 +2,27 @@
 
 ## Unreleased
 
+### Add abtest package and tracks events to validate #6536
```
becdetat, line 5: This is something we've only just started talking about, but this kind of testing instruction isn't really required in this file, as it will be used to pass to our testers and this isn't testable for anyone but a developer

client/task-list/tasks/payments/index.js
```diff
@@ -284,75 +286,81 @@ class Payments extends Component {
...
+					</CardFooter>
+				</Card>
+			);
+
+			if ( key === 'wcpay' ) {
+				return (
+					<ABTest
+						name="validate_randomization_mechanism"
+						control={ card }
+						experiment={ card } // Eventually, replace with experiment version of card.
```
becdetat, line 354: very cool 🍨

packages/abtest/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+...
+
+const [ isLoading, setIsLoading ] = useState( true );
+
+...
+
+const Loading = () => <div>Loading...</div>;
+
+if isLoading {
```
becdetat, line 46: There should be brackets around `isLoading`

Issue #6482, opened 2021-03-02
Importing and using H from woocommerce/components throws error.


leewillis77:
**Describe the bug**
I'm looking into adding setup tasks as per https://developer.woocommerce.com/2021/02/26/use-setup-tasks-to-provide-a-first-class-on-boarding-experience-for-merchants/

The basics are all working fine, so now I'm trying to work on the content. In order to keep things consistent, I'm attempting to re-use as many of the WordPress/WooCommerce components as possible. For this change, I'm introducing a `CardHeader`, with enclosed `H` component into my task, e.g.

I've required the following packages in my package.json:
```
"dependencies": {
    "@woocommerce/components": "^5.1.2",
    "@woocommerce/navigation": "^5.2.0",
...
```

In my javascript I have the following:
```
import { Button, Card, CardBody, CardHeader } from '@wordpress/components';
import { H } from '@woocommerce/components';
```

My component looks something like this:
```
<Card className="woocommerce-task-card">
    <CardHeader>
        <H>
	    MyHeader
        </H>
    </CardHeader>
...
```

**Expected behavior**
Everything works fine without the `<H></H>` in the `CardHeader`, but as soon as I add the `<H>` sub-component to my card, I expect the component to render in the same style as the blocks on the main WooCommerce admin homepage, however instead my set up tasks are no longer rendered and I receive the following error in the browser console:

```
Uncaught TypeError: Cannot read property 'withViewportMatch' of undefined
    at Module../node_modules/@woocommerce/components/build-module/chart/index.js (index.js:686)
    at __webpack_require__ (bootstrap:19)
    at Module../node_modules/@woocommerce/components/build-module/index.js (index.js:1)
    at __webpack_require__ (bootstrap:19)
    at Module../js/src/setup-tasks.js (bootstrap:83)
    at __webpack_require__ (bootstrap:19)
    at bootstrap:83
    at bootstrap:83
```

ilyasfoo:
Hi @leewillis77,

Thank you so much for filing this issue. I have tried to reproduce with a custom plugin, but I'm unable to get the browser error that you mentioned and my component rendered properly. I'm using WooCommerce `5.1.0-rc.1` and I've deactivated WooCommerce Admin plugin in my test. I'm also using the same imports that you use.

My component:
```jsx
const Task = () => {
  return (
    <Card className="woocommerce-task-card">
      <CardHeader>
        <H>MyHeader</H>
      </CardHeader>
    </Card>
  );
[...]

PR #6542, opened 2021-03-09
Remove hardcoded dependencies from examples.

leewillis77:
Following on from #6482, it was identified that the code examples under [`docs/examples/extensions`](https://github.com/woocommerce/woocommerce-admin/tree/main/docs/examples/extensions) all hardcode the Javascript dependencies, which is suboptimal, and can lead to odd bugs due to missing dependencies. 

This PR:

- Adds the WooCommerce Dependency Extraction webpack plugin
- Uses it during the build for `npm run example`
- Updates examples to use the generated `foo.asset.php` files rather than hardcoding dependencies

I'm not a webpack guru, so please check that everything in that section of the PR has been done sensibly. 

Also of note is that I had to add two sets of PHPCS exclusions relating to lack of nonce checks in `/sql-modification/woocommerce-admin-sql-modification.php` to be able to commit the asset file changes - although this seems to be a pre-existing issue unrelated to this PR. 

No changelog required

### DIFF
--- a/docs/examples/extensions/add-navigation-items/woocommerce-admin-add-navigation-items-example.php
+++ b/docs/examples/extensions/add-navigation-items/woocommerce-admin-add-navigation-items-example.php
@@ -13,17 +13,12 @@ function add_navigation_items_register_script() {
 		return;
 	}
 
+	$asset_file = require __DIR__ . '/dist/index.asset.php';
 	wp_register_script(
 		'add-navigation-items',
 		plugins_url( '/dist/index.js', __FILE__ ),
-		array(
-			'wp-hooks',
-			'wp-element',
-			'wp-i18n',
-			'wc-components',
-			'wp-plugins',
-		),
-		filemtime( dirname( __FILE__ ) . '/dist/index.js' ),
+		$asset_file['dependencies'],
+		$asset_file['version'],
 		true
 	);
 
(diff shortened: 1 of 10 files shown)


docs/examples/extensions/add-navigation-items/woocommerce-admin-add-navigation-items-example.php
```diff
@@ -13,17 +13,12 @@ function add_navigation_items_register_script() {
 		return;
 	}
 
+	$asset_file = require __DIR__ . '/dist/index.asset.php';
```
samueljseay, line 16: 👍

package.json
```diff
@@ -157,6 +157,7 @@
 		"@woocommerce/customer-effort-score": "file:packages/customer-effort-score",
 		"@woocommerce/data": "file:packages/data",
 		"@woocommerce/date": "file:packages/date",
+		"@woocommerce/dependency-extraction-webpack-plugin": "^1.4.0",
```
samueljseay, line 160: 🤔 this package is one of ours, thats in `packages`. I'm not sure we need to list this dependency, but if we did we should perhaps reference the in-repo package like we have with others.

```suggestion
		"@woocommerce/dependency-extraction-webpack-plugin": "file:packages/dependency-extraction-webpack-plugin",
```

samueljseay (changes requested):
Thank you so much for doing this work, it looks excellent, Just one little suggested change from me.

leewillis77:
Updated - thanks!

psealock (approved):
This is great, thank you @leewillis77 !

psealock:
I approved, but I think @samueljseay will need to as well because there are changes requested.

samueljseay (approved):
Thanks again for this! 🚀

PR #6553, opened 2021-03-10
Never skip changelog labelling, default to remove label.

samueljseay:
Rather than skip labelling action this defaults to always remove it, so unless the lint explicitly asked for the changelog label to be added it will be removed.

no changelog


ilyasfoo (approved):
Works like a charm! :shipit:

PR #6562, opened 2021-03-10
Fix hidden menu title on smaller screens

### LINKED ISSUE #6395: Navigation: Menu Title overlapped with latest Gutenberg
Depending on the height of the Navigation, section titles may become overlapped with elements below it.

Screen Shot 2021-02-22 at 10 26 31 AM

## Steps to reproduce

1. Grab the latest Gutenberg (10.0.0 will do) and activate the plugin.
2. With Nav turned on, go to homescreen
2. Resize your browser window to see an overlap occur. You may need to activate some extensions to get the Nav long enough

joshuatf:
Fixes #6395 

Fixes the menu being hidden when the viewport height is too small.

I tested this with the `hasSearch` prop as well and search seemed to be styled correctly.  cc @Copons to make sure I haven't broken anything in https://github.com/WordPress/gutenberg/pull/25315 in case we opt to use this feature in the future.  Is there a reason for the `overflow: hidden;` on the title?

#### Before
Screen Shot 2021-03-10 at 1 15 11 PM

#### After
Screen Shot 2021-03-10 at 1 13 09 PM

1. Enable the new navigation.
2. Shorten your viewport height so that the secondary menu overlaps the main.
3. Make sure the menu title can still be seen.

### DIFF
--- a/TESTING-INSTRUCTIONS.md
+++ b/TESTING-INSTRUCTIONS.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Fix hidden menu title on smaller screens #6562
+
+1. Enable the new navigation.
+2. Shorten your viewport height so that the secondary menu overlaps the main.
+3. Make sure the menu title can still be seen.
 ### Use wc filter to get status tabs for tools category #6525
 
 1. Register a new tab via the filter.
--- a/client/navigation/style.scss
+++ b/client/navigation/style.scss
@@ -17,6 +17,10 @@
 	.components-navigation {
 		box-sizing: border-box;
 	}
+
+	.components-navigation__menu-title {
+		overflow: visible;
+	}
 }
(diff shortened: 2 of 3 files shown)


joelclimbsthings (approved):
Tested this and works well with Gutenberg 10.1.1 installed. I also noticed the that link font size is now smaller; is that what we want?

joshuatf:
Thanks, @joelclimbsthings!

I see the font-size as 13px, do you see something different?  I think that is the expected size ( https://github.com/WordPress/gutenberg/pull/26338 ).

Hopefully this remains the same between GB and the WC nav so the experience between the two is not too jarring (assuming WP core implements the nav at some point).

joelclimbsthings:
Right! Thanks for the reference. Looks good, then!

🚢

PR #6563, opened 2021-03-10
Loader::is_admin_page() throws fatal error if woocommerce_admin_disabled is true.

leewillis77:
Loader::is_admin_page() throws fatal error if woocommerce_admin_disabled is true. 

- Disable woocommerce admin via 
```
add_filter( 'woocommerce_admin_disabled', '__return_true' );
```
- Call `Automattic\WooCommerce\Admin\Loader::is_admin_page()`

A Fatal Error is thrown since is_admin_page() uses `wc_admin_is_registered_page()` which is undefined if admin is disabled.

`PHP Fatal error:  Uncaught Error: Call to undefined function Automattic\WooCommerce\Admin\wc_admin_is_registered_page() in /sites/wc/wp-content/plugins/woocommerce-admin/src/Loader.php:758`

**Note**: Other similar issues may also be present in `Loader`, I haven't checked in detail.


leewillis77:
Test failures unrelated to this PR.

leewillis77:
We recently added support for “setup tasks” to our extensions as per the recent [call-to-arms](https://developer.woocommerce.com/2021/02/26/use-setup-tasks-to-provide-a-first-class-on-boarding-experience-for-merchants/).

However - it’s causing us support issues due to this bug. We're going to patch our extension in the interim, but anyone else following the example code to implement setup tasks will hit the exact same issues with users who have disabled WooCommerce admin. I’m aware that there's a broader discussion about the disabling flag going on in #6568 but it would be great to fix this in the meantime.

erikmolenaarnl:
Please fix this bug.

I always set woocommerce_admin_disabled to true, as I believe a CMS is supposed to only deliver content, and not include extensive reporting tools as this should be done externally to spare resources.

Thanks.

adrianduffell (commented):
Thanks for contributing this @leewillis77 

I wasn't able to reproduce the fatal error using the filter inside an `mu-plugin`. How do you have the filter set up?

The error seems like a deeper issue where parts of WC Admin are running despite being disabled.

leewillis77:
You can't *just* add the filter to trigger this. You also need to then call `Automattic\WooCommerce\Admin\Loader::is_admin_page()`. 

In our case, this is happening as we're registering extension setup actions based on the example code here:
https://github.com/woocommerce/woocommerce-admin/blob/main/docs/examples/extensions/add-task/woocommerce-admin-add-task-example.php

If you want to reproduce the issue, then the following dropped into a mu-plugin demonstrates the issue:
```
add_filter( 'woocommerce_admin_disabled', '__return_true' );
	Automattic\WooCommerce\Admin\Loader::is_admin_page();
} );
```

adrianduffell:
@leewillis77 Sorry, I missed that step in your original instructions 😅 

I was able to reproduce the fatal error.

How did you patch your extension? I think our documentation could be improved with a check for WC Admin being active before the classes are used. For example:

```
	if (  WC()->is_wc_admin_active() ) {
		Automattic\WooCommerce\Admin\Loader::is_admin_page();
	}
} );
```

Let's merge this to help folks using the current documentation. Could you please add a comment why `function_exists` was used? Something like:

```
// Check the function exists in case WC Admin is not loaded. See PR #6563.
```

We also require a changelog entry to readme.txt, then this is good to merge.

leewillis77:
Thanks for the feedback. 

I've added explanatory comment & changelog note as requested. 

In my own extension I'm just doing the same function_exists check myself before calling `Loader::is_admin_page()` so I end up with:

```
if (
    ! function_exists( 'wc_admin_is_registered_page' ) ||
    ! class_exists( 'Automattic\WooCommerce\Admin\Loader' ) ||
    ! Loader::is_admin_page() ||
    ! Onboarding::should_show_tasks()
) {
    return;
}
```

That's obviously a bit messy and overkill (indeed it would be simpler for me just to check for and call `wc_admin_is_registered_page()` directly at this point, but I'm hoping this is a temporary wallpapering....)

adrianduffell (approved):
@leewillis77 Thanks for bringing our attention to this!

LGTM 🚀

adrianduffell:
> I'm hoping this is a temporary wallpapering....

Sounds good. This change will see eventual release in WooCommerce 5.3 so hopefully your patch is only temporary.

PR #6564, opened 2021-03-10
Add filter to profile wizard steps

### LINKED ISSUE #6352: Onboarding: Allow Steps to be filtered
We have encountered challenges when working on customizing the onboarding flow for the eCommerce Plan on WordPress.com. I'm thinking it might be nice if we added in a filter to the `getSteps()` [method](https://github.com/woocommerce/woocommerce-admin/blob/main/client/profile-wizard/index.js#L109-L173) which is what derives the array of steps in the onboarding flow.

At the very least, this would allow for us to remove some steps, like themes for example, or even add in new steps as we see fit.

joshuatf:
Fixes #6352 

Adds a filter to allow steps to be added and/or removed.

### Screenshots
Screen Shot 2021-03-10 at 2 52 28 PM

1. Remove a step via the filter.
```
addFilter( STEPS_FILTER, 'woocommerce-admin', ( steps ) => {
	return steps.filter( ( step ) => step.key !== 'product-types' );
} );
```
2. Navigate to the profile wizard. `wp-admin/admin.php?page=wc-admin&path=%2Fsetup-wizard`.
3. Make sure the filtered step (product types) is not shown.

### DIFF
--- a/TESTING-INSTRUCTIONS.md
+++ b/TESTING-INSTRUCTIONS.md
@@ -65,9 +65,9 @@ Testing `woocommerce_navigation_intro_modal_dismissed`
 3. Make sure the menu title can still be seen.
 ### Add filter to profile wizard steps #6564
 
-1. Remove a step via the filter.
+1. Add the following JS to your admin head.  You can use a plugin like "Add Admin Javascript" to do this:
 ```
-addFilter( STEPS_FILTER, 'woocommerce-admin', ( steps ) => {
+wp.hooks.addFilter( 'woocommerce_admin_profile_wizard_steps', 'woocommerce-admin', ( steps ) => {
 	return steps.filter( ( step ) => step.key !== 'product-types' );
 } );
 ```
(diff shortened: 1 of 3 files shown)


TESTING-INSTRUCTIONS.md
```diff
@@ -2,6 +2,17 @@
 
 ## Unreleased
 
+### Add filter to profile wizard steps #6564
+
+1. Remove a step via the filter.
+```
+addFilter( STEPS_FILTER, 'woocommerce-admin', ( steps ) => {
```
ilyasfoo, line 9: I'm wondering if we should include this test for this at all. I'd imagine testers would have a hard time testing this since it requires some understanding of where to put the code, setting up local env, and running a build.

@woocommerce/isotope I would love to hear if folks have any opinion on this

ilyasfoo (approved):
Tests well! 🚢

I left a comment (more like an opinion) for testing instruction, but not that it matters for this change.

TESTING-INSTRUCTIONS.md
joshuatf: That's a good point @ilyasfoo.  I updated the instructions in https://github.com/woocommerce/woocommerce-admin/pull/6564/commits/827f4b4ed2f84f7162e12ecf80774011ace48af2 to create an easier way to add this snippet on a JN site to make sure it's working correctly.  Also updated the code to use the wp hooks global so no modern build is needed.

ilyasfoo: Ah, I haven't thought about that, makes sense!

PR #6566, opened 2021-03-10
Close activity panel tabs by default and track

### LINKED ISSUE #6449: Orders task - close tab by default
**Description:** As it was mentioned in the recap (pbIJXs-Fn-p2), the usage of the inventory and reviews dropped after we move them to the home screen. This can be influenced by the fact that the Orders task is opened by default when a user visits the home screen and eventually he doesn't notice that there are other tasks below orders: https://d.pr/i/JG3Vnz
We want to see if having the orders task closed when the user visits the home screen can increase the visibility of inventory and reviews, without decreasing the orders task usage.

**Acceptance criteria:**

- [ ] The orders tasks should be closed by default when the user visits the home screen: https://d.pr/i/1kUJxM
- [ ] When the user clicks in one of the tabs, trigger the `wcadmin_activity_panel_open` event. The `tab` event prop should include the name of the task that was opened.
- [ ] Add a new event tracking, so we can understand if the user has the tasks enabled:
Name: `wcadmin_home_screen_view`
Props:
`setup`: enabled | disabled (identify if the setup tasks are visible)
`orders`: enabled | disabled (identify if the orders task is visible)
`inventory`: enabled | disabled (identify if the inventory task is visible)
`reviews`: enabled | disabled (identify if the reviews task is visible)

joshuatf:
Fixes #6449 

Closes the "Orders" activity panel by default and adds in tracking data about visible panels.

@pmcpinto This has some slight differences from the original issue because of where this information can be tracked, so please let me know if they look okay:
*  I renamed the tracks event to `wcadmin_activity_panel_visible_panels`.
* The current data shows what panels are visible, but not necessarily which are hidden. (`{ orders-panel: true }` and not `{ orders-panel: true, inventory-panel: false }`  We'll have to rework the logic in the panels a bit to show all, but not a big deal if this is important.
* This same event includes `taskList` which will be true or false depending on whether or not the task list is shown.

### Screenshots

Screen Shot 2021-03-10 at 4 47 51 PM
Screen Shot 2021-03-10 at 4 48 48 PM

1. Open your browser console and enter `localStorage.setItem( 'debug', 'wc-admin:tracks' );`.
2. With the task list enabled, navigate to the homescreen.
3. Check that the `wcadmin_activity_panel_visible_panels` event is shown with `taskList: true` in its data.
4. Hide the task list.
5. Note that `wcadmin_activity_panel_visible_panels` event is shown with visible activity panels.
6. After refreshing, make sure that the "Orders" activity panel is closed by default.


octaedro (approved):
Nice job @joshuatf, the code looks good and is testing well here. Just in case, I only would add to the testing instructions that it's necessary to have enabled `Verbose` log level in order to see the console output.
Captura de Pantalla 2021-03-12 a la(s) 14 30 29

joshuatf:
Thanks for the review @octaedro!  Added that extra note you recommended.

PR #6568, opened 2021-03-11
Disabling wc-admin: Remove filter

psealock:
Removes the ability to turn off wc-admin.

Now that wc-admin is in Core, more and more functionality is baked into the plugin, and extensions build off its functionality, it makes less and less sense to allow users or plugins to disable it.

This PR removes the ability to disable the plugin and will be a good space to discuss the ramifications.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -110,6 +110,7 @@ Release and roadmap notes are available on the [WooCommerce Developers Blog](htt
 - Dev: Store profiler - Added MailPoet to Business Details step  #6503
 - Dev: Store profiler - Added MailPoet to new Business Details step  #6515
 - Dev: Add tilde (~) to represent client root directory for imports. #6517
+- Dev: Remove ability to disable wc-admin. #6568
 
 == 2.1.0 3/10/2021  ==
 
--- a/src/FeaturePlugin.php
+++ b/src/FeaturePlugin.php
@@ -60,15 +60,6 @@ class FeaturePlugin {
 	 * Init the feature plugin, only if we can detect both Gutenberg and WooCommerce.
 	 */
 	public function init() {
-		/**
-		 * Filter allowing WooCommerce Admin to be disabled.
-		 *
-		 * @param bool $disabled False.
-		 */
-		if ( apply_filters( 'woocommerce_admin_disabled', false ) ) {
-			return;
-		}
-
 		$this->define_constants();
 
 		require_once WC_ADMIN_ABSPATH . '/src/Notes/DeprecatedNotes.php';


psealock:
cc @timmyc @elizaan36 @pmcpinto

joshuatf (approved):
Code-wise this looks good to me.

I would like it if our features were better encapsulated so that they could be toggled on/off, but that would probably be a rather large refactor to some areas of WooCommerce Admin.

leewillis77:
#6563 is related (and could be closed if this gets merged)

timmyc:
Added the blocked label here since this is likely something we will need to do a bit of coordination on.

In 5.2 of core [we will start collecting Tracker data](https://github.com/woocommerce/woocommerce/pull/28535) ( for woocommerce.com connected sites ) to get a sense of how many sites are currently toggling wc-admin off via filter.

I also feel we might want to do some community outreach here and get some further details from store owners / builders who have deactivated wc-admin and see if there are certain features they would like to toggle off, and maybe we need to expand the feature flag system to allow a finer control over active features.

timmyc:
Just saw [this comment](https://github.com/woocommerce/woocommerce-admin/issues/2223#issuecomment-793891620) from @lkraav that really speaks to the desire for finer grained controls of wc-admin features. The primary feature, as mentioned in that comment, that I personally feel users will want to disable is Analytics since many sites do already have an analytics toolkit established that they prefer to use.

Adding in full feature flag support for analytics might be a bit of work considering how integrated it is, but is definitely feasible.

joshuatf:
>Adding in full feature flag support for analytics might be a bit of work considering how integrated it is, but is definitely feasible.

Strongly agree.

roykho:
We should definitely allow ways for plugins to disable specific or all features of WC Admin. A good use case would be Product Vendors extension where the logged in user is a vendor admin. They have their own customized WP menu items such as custom reports and such however they do share the Products menu with the site admin. WC Admin attempts to hijack this and causes issues for vendor admins where the Products menu item disappears.

Since vendor admins don't need to use any of the features, would be good to allow them to disable all of this. Currently the way to do this for Product Vendors is something like this:

```
if ( is_vendor() ) {
     add_filter( 'woocommerce_admin_features', 'remove_features' );
}

function remove_features() {
     return array();
}
```

Of course we could continue to use this method.

timmyc:
Thanks @roykho - another great use case we need to consider.

adrianduffell:
Some more feedback from @erikmolenaarnl in https://github.com/woocommerce/woocommerce-admin/pull/6563#issuecomment-800089591.

> I always set woocommerce_admin_disabled to true, as I believe a CMS is supposed to only deliver content, and not include extensive reporting tools as this should be done externally to spare resources.

adrianduffell:
Just noting our own documentation causes a fatal error when WC Admin is disabled with this filter. Issue https://github.com/woocommerce/woocommerce-admin/issues/6673

adrianduffell:
Another instance where an extension had a fatal error with WC Admin is disabled https://github.com/woocommerce/woocommerce-admin/issues/6703

Since this PR is blocked, I think we need to raise more awareness about this filter among extension developers.

timmyc:
Kicked off a thread to discuss this further among the dev teams: pcShBQ-1i-p2

psealock:
Thanks for the discussion everyone. I'm closing this one in favour of https://github.com/woocommerce/woocommerce-admin/pull/6568 which hopefully accomplishes the goal of keeping wc-admin the platform while allowing parts to be turned off to keep WooCommerce lean.

lkraav:
> Thanks for the discussion everyone. I'm closing this one in favour of https://github.com/woocommerce/woocommerce-admin/pull/6568 which hopefully accomplishes the goal of keeping wc-admin the platform while allowing parts to be turned off to keep WooCommerce lean.

But this is #6568?

adrianduffell:
@lkraav Apologies, it should be #7232.

PR #6576, opened 2021-03-11
Fix CI issues

samueljseay:
PHP unit tests were failing due to a change in the Github actions environment.
E2E tests were failing because of that issue (they rely on port 8084 which is used in ubuntu 20 environment) *and* because of an issue with @woocommerce/e2e-environment

This seeks to fix both issues.

No changelog required.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -14,7 +14,7 @@ jobs:
             - name: Setup PHP
               uses: shivammathur/setup-php@2.9.0
               with:
-                  php-version: '7.4'
+                  php-version: '7.3'
             - name: Install PHP dependencies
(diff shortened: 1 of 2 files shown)


joshuatf (approved):
Tests are ✅  - thanks for the fix!  🚢

.github/workflows/e2e.yml
```diff
@@ -34,5 +34,6 @@ jobs:
                   composer require wp-cli/i18n-command
                   npm run build
                   npm install jest --global
+                  WP_VERSION=5.6.2
```
louwie17, line 37: @samueljseay any particular reason you set the version here instead of in line `28`? under the `env` key.

PR #6582, opened 2021-03-12
Add Typescript and page objects to E2E suite

louwie17:
Partly addresses #6313 

This is basically the original https://github.com/woocommerce/woocommerce-admin/pull/6484 PR just without playwright.
The biggest changes are using TypeScript and Page objects.

### Screenshots

Screen Shot 2021-03-12 at 5 04 09 PM

- Load this PR and run `npm install`
- Run `npm run docker:up`
- Run `npm run test:e2e`, it should run all the tests and pass
- Run `npm run test:e2e-dev`, it should show a chromium browser and run the tests (one of the tests might fail the second time around, given we can't reset the DB yet).
- Run `npm run test:e2e tests/e2e/specs/activate-and-setup/setup-wizard.test.ts`, it should just run that test file.


bin/wait-for-build.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
louwie17, line 1: Had to still copy this one over as there was no way to call it directly in the package. Mostly because of line `11`, it uses a relative path in the original, which wouldn't work for us.

package.json
```diff
@@ -60,8 +60,9 @@
 		"prestart": "npm run -s install-if-deps-outdated",
 		"start": "cross-env WC_ADMIN_PHASE=development npm run build:packages && cross-env WC_ADMIN_PHASE=development npm run build:feature-config && concurrently \"cross-env WC_ADMIN_PHASE=development webpack --watch\" \"node ./bin/packages/watch.js\"",
 		"pretest": "npm run -s install-if-no-packages",
-		"test:e2e": "npm explore @woocommerce/e2e-environment -- npm run test:e2e",
-		"test:e2e-dev": "npm explore @woocommerce/e2e-environment -- npm run test:e2e-dev",
+		"test:e2e": "bash ./bin/wait-for-build.sh && ./node_modules/@woocommerce/e2e-environment/bin/e2e-test-integration.js",
+		"test:e2e-dev": "bash ./bin/wait-for-build.sh && DEBUG=1 ./node_modules/@woocommerce/e2e-environment/bin/e2e-test-integration.js --dev",
+		"test:e2e-debug": "bash ./bin/wait-for-build.sh && DEBUG=1 ./node_modules/@woocommerce/e2e-environment/bin/e2e-test-integration.js --dev --debug",
```
louwie17, line 65: Instead of `npm explore` I am calling the `e2e-test-integration.js` directly, this allows `jest` to use our `babel.config.js`.
When using `npm explore` it won't use our `babel.config.js`.

typings/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module '@woocommerce/e2e-utils';
```
louwie17, line 1: Typescript was complaining when trying to import this library although the `tsconfig.json` has allowJs set to `true`.
I am not sure why that is the case, declaring it as a module here fixed it, I suppose we could create types for the library, as they don't have any for them.

samueljseay (approved):
Great stuff! 🚀

PR #6585, opened 2021-03-14
Docblock fixes in TimeInterval

tivnet:
Fixes #
Docblocks only, to please IDEs.
A couple of places `DateTime -> \WC_DateTime`. Others - just `\`

### DIFF
--- a/src/API/Reports/TimeInterval.php
+++ b/src/API/Reports/TimeInterval.php
@@ -30,14 +30,12 @@ class TimeInterval {
 	 * Converts local datetime to GMT/UTC time.
 	 *
 	 * @param string $datetime_string String representation of local datetime.
-	 *
 	 * @return \DateTimeInterface
 	 * @throws \Exception \DateTime emits Exception in case of an error.
 	 */
 	public static function convert_local_datetime_to_gmt( $datetime_string ) {
 		$datetime = new \DateTime( $datetime_string, new \DateTimeZone( wc_timezone_string() ) );
 		$datetime->setTimezone( new \DateTimeZone( 'GMT' ) );
-
 		return $datetime;
 	}
 
@@ -54,7 +52,6 @@ class TimeInterval {
 		} else {
 			$datetime->set_utc_offset( wc_timezone_offset() );
 		}
-
 		return $datetime;
 	}
 
@@ -75,7 +72,6 @@ class TimeInterval {
 		} else {
 			$datetime->set_utc_offset( wc_timezone_offset() );
 		}
-
 		return $datetime;
 	}
 
@@ -84,7 +80,6 @@ class TimeInterval {
 	 *
 	 * @param string $time_interval Time interval.
 	 * @param string $table_name    Name of the db table relevant for the date constraint.
-	 *
 	 * @return string
 	 */
 	public static function db_datetime_format( $time_interval, $table_name ) {
@@ -109,7 +104,6 @@ class TimeInterval {
 			'year'    => "YEAR({$table_name}.date_created)",
 
 		);
-
 		return $mysql_date_format_mapping[ $time_interval ];
 	}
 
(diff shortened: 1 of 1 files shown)


adrianduffell (changes requested):
@tivnet Thanks for contributing this! Overall it looks good.

 I think ideally we should use [`\DateTimeInterface`](https://www.php.net/manual/en/class.datetimeinterface.php) for all of these return types. It would provide flexibility to change the internal date implementation in the future. Does this work ok in your IDE?

tivnet:
@adrianduffell 

> Does this work ok in your IDE?

In PhpStorm it works fine:

image

Cannot say anything about other IDEs.

adrianduffell:
> In PhpStorm, `@return \DateTimeInterface` works fine:

@tivnet Thanks, let's go ahead with changing it to `\DateTimeInterface` and then I think this will be good to merge 🙂

tivnet:
@adrianduffell Done.

src/API/Reports/TimeInterval.php
```diff
@@ -285,21 +304,23 @@ public static function next_hour_start( $datetime, $reversed = false ) {
 		$hours_offset_timestamp = $timestamp + ( $hour_increment * HOUR_IN_SECONDS - $seconds_into_hour );
 
 		if ( $reversed ) {
-			$hours_offset_timestamp --;
+			$hours_offset_timestamp--;
```
adrianduffell, line 307: Let's keep the space before `--`

src/API/Reports/TimeInterval.php
```diff
@@ -30,18 +30,21 @@ class TimeInterval {
 	 * Converts local datetime to GMT/UTC time.
 	 *
 	 * @param string $datetime_string String representation of local datetime.
-	 * @return DateTime
+	 *
```
adrianduffell, line 33: A new line was added here.

src/API/Reports/TimeInterval.php
```diff
@@ -247,6 +260,7 @@ public static function intervals_between( $start_datetime, $end_datetime, $inter
 					$start_datetime = self::next_week_start( $start_datetime );
 					$week_count++;
 				} while ( $start_datetime <= $end_datetime );
+
```
adrianduffell, line 263: Let's remove the new lines that were added.

src/API/Reports/TimeInterval.php
```diff
@@ -574,10 +608,12 @@ public static function normalize_between_params( $request, $param_names, $is_dat
...
-	 * @param  mixed           $value Parameter value.
-	 * @param  WP_REST_Request $request REST Request.
-	 * @param  string          $param Parameter name.
-	 * @return WP_Error|boolean
+	 * @param mixed            $value   Parameter value.
+	 * @param \WP_REST_Request $request REST Request (Unused).
+	 * @param string           $param   Parameter name.
+	 *
+	 * @return \WP_Error|true
+	 * @noinspection PhpUnusedParameterInspection
```
adrianduffell, line 616: Let's remove the @noinspection tag

adrianduffell (changes requested):
Hi @tivnet,

Thanks for making the `\DateTimeInterface` change. However it looks like some other changes came along with it compared to your original commit. I'm guessing it was caused by auto formatting on save. I've highlighted some of it in the code comments.

Are you able to remove these?

Thank you!

tivnet:
@adrianduffell I changed the IDE settings... Now it removed more blanks :)

adrianduffell:
> @adrianduffell I changed the IDE settings... Now it removed more blanks :)

Thanks @tivnet , the removal of the blank lines in the code is a problem too 😅 . Ideally this PR would just be making docblock changes.

tivnet:
@adrianduffell Never mind :)

PR #6595, opened 2021-03-15
Fix TypeScript build for `packages` and add types for wc/data's onboarding store

samueljseay:
Changes:

1. Made a bunch of changes that allow TS building of packages to work properly with alongside the webpack build
2. Removed some dead code, such as in the Dashboard component it was hydrating some data that was not used at all.
3. Add some basic types to wc data's selectors and also flesh out the types of the onboarding data store
4. Give the onboarding reducer a default state, This makes it easier to reason about whats in the reducer and also more consistent to type. (I think we should do this for all reducers we can, maybe in future some reducers should be reshaped to compartmentalise dynamic data that may change in shape.
5. Fix Jest to work with TS

**Note**: I would have loved to add more usages and further refactoring, but as soon as you add types to a component, there is some follow up stuff you need to type so I stopped here. My next step will be to add types for the `settings` store, which should allow me to convert a couple more `client` components to TS.

### Accessibility

n/a

### Screenshots
n/a 

For testing we should do 2 things:

1. run a fresh build of the JS and a brief smoke test that the app runs without errors, and there are no TS errors in console.
2. A smoke test of dashboard since the redundant props were removed.

No changelog required


.eslintrc.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
 	extends: [ 'plugin:@woocommerce/eslint-plugin/recommended' ],
 	settings: {
-		'import/resolver': 'webpack',
+		'import/resolver': 'typescript',
```
samueljseay, line 4: Imports with mixed js/ts were not working, this fixes that.

.eslintrc.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
 				'plugin:@typescript-eslint/recommended',
 			],
 			rules: {
+				camelcase: 'off',
```
samueljseay, line 30: This rule won't work well with any snake case data we get from the server and add types for. Example in this case was the profile items data. in future we could camelCase convert everything we get from the server, for now switching this off on TS makes sense imo.

bin/packages/build.js
```diff
@@ -63,10 +63,12 @@ const isScssFile = ( filepath ) => {
  * @return {string}             Build path
  */
 function getBuildPath( file, buildFolder ) {
-	const pkgName = getPackageName( file );
+	// if the file has extension of ts, replace with js
+	const fileName = file.replace( '.ts', '.js' );
```
samueljseay, line 67: We still build TS as before but they need to be built as JS files.

bin/packages/build.js
```diff
@@ -198,7 +200,7 @@ function buildJsFileFor( file, silent, environment ) {
  */
 async function buildPackage( packagePath ) {
 	const srcDir = path.resolve( packagePath, SRC_DIR );
-	const jsFiles = glob.sync( `${ srcDir }/**/*.js`, {
+	const jsFiles = glob.sync( `${ srcDir }/**/*.{ts,tsx,js}`, {
```
samueljseay, line 203: We need to process ts and tsx files too.

bin/packages/get-babel-config.js
```diff
@@ -22,27 +22,33 @@ const overrideOptions = ( target, targetName, options ) => {
...
 		plugins,
-		presets: map( babelDefaultConfig.presets, ( preset ) =>
-			overrideOptions( preset, '@babel/preset-env', {
-				modules: 'commonjs',
-				corejs: '3',
-				useBuiltIns: 'usage',
-			} )
-		),
+		presets: [
+			[ '@babel/preset-typescript' ],
```
samueljseay, line 26: add the TS preset to the package build babel config. It is my dream to remove all this seperate build stuff and let webpack compile it instead, I hope to get to this in the next few months but for now, this will have to do.

client/dashboard/index.js
```diff
@@ -37,16 +33,4 @@ class Dashboard extends Component {
...
-		? withOnboardingHydration( { tasksStatus: onboardingData.tasksStatus } )
-		: identity,
-	withSelect( ( select ) => {
-		const { getProfileItems } = select( ONBOARDING_STORE_NAME );
-		const profileItems = getProfileItems();
-
-		return { profileItems };
-	} )
-)( Dashboard );
+export default Dashboard;
```
samueljseay, line 36: The query/path are passed into this component so all this stuff was completely unused.

client/homescreen/index.tsx
```diff
@@ -8,37 +8,46 @@ import { getSetting } from '@woocommerce/wc-admin-settings';
...
 import { getHistory, getNewPath } from '@woocommerce/navigation';
-
+import type { History } from 'history';
 /**
  * Internal dependencies
  */
 import Layout from './layout';
 
-const Homescreen = ( { profileItems, query } ) => {
+type HomescreenProps = ReturnType< typeof withSelectHandler > & {
```
samueljseay, line 20: This seems to work pretty well to type the compose props. for other components, I'd make this a composition of withSelect and withDispatch.

client/homescreen/index.tsx
```diff
@@ -8,37 +8,46 @@ import { getSetting } from '@woocommerce/wc-admin-settings';
...
+			getNewPath( {}, '/setup-wizard', {} )
+		);
 	}
 
 	return <Layout query={ query } />;
 };
 
 const onboardingData = getSetting( 'onboarding', {} );
 
+const withSelectHandler = ( select: WCDataSelector ) => {
```
samueljseay, line 38: Separating out a variable allows us to provide this to the ReturnType generic type.

client/homescreen/index.tsx
```diff
@@ -8,37 +8,46 @@ import { getSetting } from '@woocommerce/wc-admin-settings';
...
+	const { getProfileItems } = select( ONBOARDING_STORE_NAME );
+	const profileItems = getProfileItems();
```
samueljseay, line 40: It doesn't look like much but check this out in VSCode, you'll see that all the typing for profileItems is there. 🎉

packages/data/src/index.ts
```diff
@@ -0,0 +1,119 @@
...
+	| typeof USER_STORE_NAME
+	| typeof OPTIONS_STORE_NAME
+	| typeof NAVIGATION_STORE_NAME
+	| typeof NOTES_STORE_NAME
+	| typeof REPORTS_STORE_NAME
+	| typeof ITEMS_STORE_NAME;
+
+// As we add types to all the package selectors we can fill out these unknown types with real ones. See one
+// of the already typed selectors for an example of how you can do this.
+type WCSelectorType< T > = T extends typeof REVIEWS_STORE_NAME
```
samueljseay, line 94: This is the magic for the wc/data selectors. For now all the stores I haven't typed just have some of the basic stuff thats available on `WPDataSelectors`, but as I fill them out I'll add their selector types in here.

packages/data/src/index.ts
```diff
@@ -0,0 +1,119 @@
...
+	? WPDataSelectors
+	: T extends typeof NOTES_STORE_NAME
+	? WPDataSelectors
+	: T extends typeof REPORTS_STORE_NAME
+	? WPDataSelectors
+	: T extends typeof ITEMS_STORE_NAME
+	? WPDataSelectors
+	: never;
+
+export interface WCDataSelector {
```
samueljseay, line 116: If you type your withSelect's select argument to this, you will get all the available selectors for that store. FWIW this does make the @types/wordpress__data package a bit redundant, but this is much more useful than the very vague high level type that package provides.

packages/data/src/onboarding/reducer.js
```diff
@@ -6,7 +6,21 @@ import TYPES from './action-types';
 const onboarding = (
 	state = {
 		errors: {},
-		profileItems: {},
+		profileItems: {
+			business_extensions: null,
```
samueljseay, line 10: I really think we should be doing this default state stuff everywhere. Its informative to the next person and it also makes the type of the reducer consistent. I just need to fix up the tests to account for this.

packages/data/src/onboarding/selectors.ts
```diff
@@ -0,0 +1,111 @@
...
+	// TODO - fill out this type
+	themeMods: unknown;
+	wcPayIsConnected: false;
+};
+
+export type Industry = {
+	slug: string;
+};
+
+export type ProductCount = '0' | '1-10' | '11-100' | '101 - 1000' | '1000+';
```
samueljseay, line 67: I've used the specific available string values here because I think it's self documenting and helpful to the next person trying to figure out what onboarding does.

packages/data/src/types.ts
```diff
@@ -0,0 +1,8 @@
+// Type for the basic selectors built into @wordpress/data, note these
+// types define the interface for the public selectors, so state is not an
+// argument.
+export type WPDataSelectors = {
+	hasStartedResolution: ( selector: string, args?: string[] ) => boolean;
```
samueljseay, line 5: This doesn't cover everything WPDataSelectors has, I just started with the most commonly used selectors that we use all the time.

tsconfig.json
```diff
@@ -15,7 +15,6 @@
 		"esModuleInterop": true,
 		"skipLibCheck": false,
 		"module": "esnext",
-		"typeRoots": [ "**/node_modules/@types" ],
```
samueljseay, line 18: This should be able to stay removed. tsconfig by default loads types from `@types` and some other places so doing this narrows what it will pick up.

webpack.config.js
```diff
@@ -88,7 +88,11 @@ const webpackConfig = {
 			},
 			{
 				test: /\.(t|j)sx?$/,
-				exclude: /node_modules(\/|\\)(?!(debug))/,
+				exclude: [
+					/node_modules(\/|\\)(?!(debug))/,
+					/build/,
```
samueljseay, line 93: This rule was trying to process built files as TS. These should not be included.

packages/data/src/onboarding/selectors.ts
```diff
@@ -0,0 +1,114 @@
...
+export type RevenueTypeSlug =
+	| 'none'
+	| 'rather-not-say'
+	| 'up-to-2500'
+	| '2500-10000'
+	| '10000-50000'
+	| '50000-250000'
+	| 'more-than-250000';
+
+export type ProfileItemsState = {
```
louwie17, line 100: Not sure what you have done in the past, but I tend to prefer to use `interface` for describing an object, I guess the difference is pretty subtle: https://www.typescriptlang.org/docs/handbook/advanced-types.html#interfaces-vs-type-aliases although we probably do want to decide where we draw the line :)

I guess I like what they said here ;) 

> Because an interface more closely maps how JavaScript objects work by being open to extension, we recommend using an interface over a type alias when possible.
>
> On the other hand, if you can’t express some shape with an interface and you need to use a union or tuple type, type aliases are usually the way to go.

packages/data/src/onboarding/selectors.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export type OtherPlatformSlug =
+	| 'shopify'
+	| 'bigcommerce'
+	| 'wix'
+	| 'amazon'
+	| 'ebay'
+	| 'etsy'
+	| 'squarespace'
+	| 'other';
```
louwie17, line 89: This is handy, although it will be an extra item to keep track off when updating the onboarding flow.

packages/data/src/types.ts
louwie17: Would it be worth to add this to typescript declaration file? I did add a `~/typings` folder for the `d.ts` files.
We could create one for @wordpress/data for now:

`declare module "@wordpress/data" { ... }` or create a folder within `@wordpress-data` with an `index.d.ts`. 
This doesn't have to be part of this PR, but just a thought as we iterate, and wait for @wordpress/data to be written in TS.

tsconfig.json
louwie17: I added a `types` folder in root for type definitions as well, I am curious if it will pick that up automatically.

packages/data/src/onboarding/reducer.js
```diff
@@ -3,13 +3,29 @@
  */
```
louwie17, line 3: Was there a specific reason this stayed as a JS file?

louwie17 (commented):
I haven't fully tested the homescreen and such yet, but when I tried to build I got some import errors, also shown in my vscode in the homescreen typescript file. I encountered something similar for the E2E stuff and had added https://github.com/woocommerce/woocommerce-admin/blob/main/typings/index.d.ts to solve this, not sure why I see this, but you don't?


Screen Shot 2021-03-18 at 10 45 36 AM
Screen Shot 2021-03-18 at 10 42 50 AM

packages/data/src/types.ts
samueljseay: I was hesitant to create one for wordpress__data. It does have its own types but they just aren't very useful to us right now. This is used to compose a type of these methods and those on the wc stores so its not really intended to provide type declarations for wp data.

tsconfig.json
samueljseay: @louwie17

From the docs "This config file will include all packages under ./typings and ./vendor/types, and no packages from ./node_modules/@types. All paths are relative to the tsconfig.json."

So perhaps not. Can I suggest you rename your folder to one of the supported folder names? typings seems good.

bin/packages/build.js
samueljseay: I'm not sure what you mean? Compilation of ts files with straight babel doesn't rename the file that gets generated, so this renames any file thats `ts` to `js` so that webpack doesn't try recompile it.

packages/data/src/onboarding/selectors.ts
samueljseay: @louwie17 yeah I'm aware of the difference. Although its true you can extend a JS object, having a closed type discourages this. It's also shorter and more succinct so I prefer it because the 2 are so close to being the same thing.

You're looking at the old docs though, this is the new docs: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces

They say:

"For the most part, you can choose based on personal preference, and TypeScript will tell you if it needs something to be the other kind of declaration. If you would like a heuristic, use interface until you need to use features from type."

So my feeling is for something like `Window` absolutely it should be an interface, especially because it's probably going to get extended many times, but for a type definition I want to remain closed to extension `type` makes more sense to me.

packages/data/src/onboarding/selectors.ts
samueljseay: @louwie17 there will always be manual updating to do with these kinds of types for sure, but I think its mostly helpful to have it self-documented and keep track of possible values.

packages/data/src/onboarding/reducer.js
samueljseay: I didn't explore how easy it would be to type this. I'll look into it.

bin/packages/build.js
louwie17: I see, sorry I mis-understood, makes total sense then :)

packages/data/src/types.ts
louwie17: No worries, I thought I mention it, all good with me.

tsconfig.json
louwie17: Sure that would be great, the folder is called `typings` I guess this doesn't get automatically picked up either (as from your previous comment)?

samueljseay: @louwie17 sounds good, I'll look at changing this to support that. Thanks for raising that!

packages/data/src/types.ts
samueljseay: @louwie17 I appreciate it for sure. I still feel not great about ignoring `@types/wordpress__data` Maybe we can look at whether we could provide some upstream help to them to make the types better going forward. Its something I'm definitely thinking about.

.eslintrc.js
samueljseay: @louwie17 
> I wasn't a huge fan of graphql

Thats tragic, guess you're not a 10xer unicorn then smh 😛 

> I guess something we would have to do client side

Yeah i imagine a function that transforms the props to camel case / snake case when the data comes in / goes out. Its probably not a huge deal to solve at this point though.

.eslintrc.js
samueljseay: @louwie17 Are you sure about that? I think the ts import resolver is built on the eslint one, but I am not clear exactly on what can be removed I'll look into this.

louwie17: Not sure, @mattsherman added this, but when he added it he also installed the `eslint-import-resolver-webpack` plugin, so I figured it could be removed if you installed a specific typescript resolver.

packages/data/src/onboarding/selectors.ts
samueljseay: @louwie17 I'm usually one for consistency but frankly it doesn't bother me too much what you use, if you prefer interface I say carry on. The docs are clear that it *almost never* matters. I figured for these package types I may as well keep them closed, I really hope no one extends them, but for many other situations `interface` will be just fine. 

I wonder if maybe we should consider using `interface` in client more exclusively, and `type` in packages? I'm not sure. There are some situations where the errors delivered about `interface` are supposedly clearer according to the docs, but I hadn't ever found that an issue.

.eslintrc.js
samueljseay: to be honest, looking deeper at this, the eslint still seems a bit broken as it is. I'll see if I can improve this.

samueljseay: I think this is ok now.

packages/data/src/types.ts
samueljseay: I've declared modules for the 2 @wordpress packages and I've added the type defs for lodash.

tsconfig.json
samueljseay: I've reverted my changes here and opted to manually declare everything we want to include.

packages/data/src/onboarding/selectors.ts
samueljseay: I'm gonna leave this for now. We could later decide on some heuristics/best practices for the team with TS, but I think its ok to use either.

packages/data/src/onboarding/reducer.js
samueljseay: I looked at this, the reducer needs some refactoring. It does this cheeky thing where it destructures all the args regardless of what action is taking place, so there is some stuff there that has not been typed yet. I'll come back to improve this later when I do more typing of this package.

samueljseay:
@louwie17 keen for you to take another look at this. I think i've resolved everything.

louwie17 (approved):
Thanks for addressing the feedback where needed @samueljseay changes look great!
It also tested well, I don't see the errors in my vscode anymore and it builds well. Did a quick smoke test of the homescreen and all looks good.
Ready to ship 🚀

PR #6597, opened 2021-03-15
TRY fixing persisted queries by moving NavigationComponent back out of the router HoC

### LINKED ISSUE #6581: Persisted queries lose context when navigating to excluded page in Woo navigation
**Describe the bug**
When you change the query on an Analytics report, that should persist between views. This works to some degree, but you currently lose the persisted query when navigation to a page that doesn't utilize them (ie homescreen). 

**To Reproduce**

1. Go to Analytics Report
2. Change Time period
3. Navigate to homescreen
4. Navigate back to Analytics Report
5. Notice that the time period is reset again

**Expected behavior**
Time period should remain to the period set.

**Notes**

After looking at this for the bulk of the day, I've figured out that the issue stems from the fact that the `NavigationPlugin`, where the persisted queries logic lies, is now being remounted on every history change (react navigation). This causes the persisted query state to be reset. 

After some digging, I found that this started happening after [this PR](https://github.com/woocommerce/woocommerce-admin/pull/6247) that reconfigured various parts of the app in order to consolidate logic/styles governing the Header component and the rest of the page. 

In this PR the component that the `NavigationPlugin` is rendered in was moved from it's own React root element at the base of the page, to within the Header. I believe this was to bring together the elements (header bar and navigation) that need to stay fixed on the page. Unfortunately the base router causes all content within this React root to be remounted every time the route changes. 

_Two options to address this_:
- Move the root of the NavigationPlugin _back_ out of the header and to it's own React root element. This will avoid remounting and return the original function. The downside is that the styles will be impacted and we won't be able to keep all fixed elements in the same place, as per the aforementioned PR. 
- Use some kind of persistence within the NavigationPlugin to allow the persisted query logic to not be impacted by remounts

joelclimbsthings:
Fixes #6581 

As per the original issue discussion, this PR employs one possible fix for the fact that persisted queries are no longer working as expected in the new navigation feature. 

In short, there was a change made in an [unrelated PR](https://github.com/woocommerce/woocommerce-admin/pull/6247) that resulted in the `NavigationPlugin` component being remounted on every router change, which causes the component to lose state that was being used to implement the persisted queries.

This PR is one potential solution for that issue, which simply involves moving the mentioned component _back_ to where it came from, which is it's own React root on a custom DOM element. 

I don't _believe_ this diminishes the utility of the PR that moved it, but I'm paging @joshuatf to ensure that that's the case. If so, we could possibly try a solution that would involve persisting state via localStorage/etc as mentioned in the originating issue.

1. Go to Analytics Report
2. Change Time period
3. Navigate to homescreen
4. Navigate back to Analytics Report
5. Notice that the time period is _still_ what you set on step 2

### DIFF
--- a/client/header/index.js
+++ b/client/header/index.js
@@ -163,7 +163,6 @@ export const Header = ( { sections, isEmbedded = false, query } ) => {
 			) }
 
 			<div className="woocommerce-layout__header-wrapper">
-				{ /* { window.wcAdminFeatures.navigation && <Navigation /> } */ }
 				{ renderTaskListBackButton() }
 				<Text
 					className={ `woocommerce-layout__header-heading ${ backButtonClass }` }
--- a/client/header/index.js
+++ b/client/header/index.js
@@ -21,7 +21,6 @@ import './style.scss';
 import ActivityPanel from './activity-panel';
 import { MobileAppBanner } from '../mobile-banner';
 import useIsScrolled from '../hooks/useIsScrolled';
-import Navigation from '../navigation';
 
 const renderTaskListBackButton = () => {
 	const currentUrl = new URL( window.location.href );
--- a/client/header/index.js
+++ b/client/header/index.js
@@ -163,7 +163,7 @@ export const Header = ( { sections, isEmbedded = false, query } ) => {
 			) }
 
 			<div className="woocommerce-layout__header-wrapper">
-				{ window.wcAdminFeatures.navigation && <Navigation /> }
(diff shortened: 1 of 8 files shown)


joshuatf (commented):
Nice work finding this issue, @joelclimbsthings.

I do think that this undermines some of the header height work and I believe will create issues when additional header elements are added/removed (such as the mobile app banner).

I think our efforts might best be spent working on fixing the issue with the persisted query.

PR #6603, opened 2021-03-16
Add tracking for mail poet installs

### LINKED ISSUE #6596: Add tracking for Mail Poet installs
https://github.com/woocommerce/woocommerce-admin/pull/6575#issuecomment-799316890 specifies an accidentally missed acceptance criteria that we need to do some tracking for mail poet installs.

Quoting the comment:

`The event prop should be identical to the other extensions props: install_mailpoet: True|False`

samueljseay:
Fixes #6596 

### Accessibility

n/a

### Screenshots

n/a

Note that I won't be adding testing instructions for an added track here. But here's how to test in dev

1. Make sure you have tracks logging enabled in console 
2. Go to the onboarding wizard (doesn't have to be fresh site)
3. Choose US as country, go through steps till you get to "free features" tab
4. Expand the options, disable all the other extensions to install except for mail poet.
5. Click continue, you should see a track logged in console: `wcadmin_storeprofiler_store_business_features_continue` one of the props of the track should be `mailpoet: true`

### DIFF
--- a/client/profile-wizard/steps/business-details/flows/selective-bundle/index.js
+++ b/client/profile-wizard/steps/business-details/flows/selective-bundle/index.js
@@ -91,6 +91,7 @@ class BusinessDetails extends Component {
 				extensionInstallationOptions[ 'woocommerce-services' ],
 			install_mailchimp:
 				extensionInstallationOptions[ 'mailchimp-for-woocommerce' ],
+			install_mailpoet: extensionInstallationOptions.mailpoet,
 			install_jetpack: extensionInstallationOptions.jetpack,
 			install_google_ads:
 				extensionInstallationOptions[ 'kliken-marketing-for-google' ],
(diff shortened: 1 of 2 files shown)


client/profile-wizard/steps/business-details/flows/selective-bundle/index.js
```diff
@@ -91,6 +91,7 @@ class BusinessDetails extends Component {
 				extensionInstallationOptions[ 'woocommerce-services' ],
 			install_mailchimp:
 				extensionInstallationOptions[ 'mailchimp-for-woocommerce' ],
+			install_mailpoet: extensionInstallationOptions.mailpoet,
```
joshuatf, line 94: In retrospect, we probably should have mapped the `extensionInstallationOptions` directly to the event property names, but I suppose this would break historical tracks data if we updated now.

joshuatf (approved):
Testing well and LGTM! 🚢 

Optionally could add this to testing instructions.

client/profile-wizard/steps/business-details/flows/selective-bundle/index.js
samueljseay: Yeah I agree. I did consider we could improve this and keep the prop names but the effort felt out of scope.

PR #6614, opened 2021-03-17
Fix persisted queries being lost when navigating to excluded page (such as Homescreen)

### LINKED ISSUE #6581: Persisted queries lose context when navigating to excluded page in Woo navigation

joelclimbsthings:
Fixes woocommerce/woocommerce-admin#6581 

As per the original issue discussion, this PR employs one possible fix for the fact that persisted queries are no longer working as expected in the new navigation feature. 

In short, there was a change made in an [unrelated PR](https://github.com/woocommerce/woocommerce-admin/pull/6247) that resulted in the `NavigationPlugin` component being remounted on every router change, which causes the component to lose state that was being used to implement the persisted queries.

This PR resolves the issue by moving the state around persisted queries in `wp.data`. This allows us to retain the state over router changes, but not over page refreshes (which could get messy). 

1. Go to Analytics Report
2. Change Time period
3. Navigate to another report
4. Notice that the time period stays the same
5. Navigate to Homescreen
6. Navigate back to Analytics Report
7. Ensure that the time period is _still_ what you set on step 2


packages/data/src/navigation/actions.js
```diff
@@ -16,6 +16,13 @@ export function setMenuItems( menuItems ) {
 	};
 }
 
+export function setPersistedQuery( persistedQuery ) {
```
joshuatf, line 19: I would love to see these actions more "event" driven.  We haven't done a great job of this in the past when creating data stores, but I have some prior art in the nav favorties data store.

I'm thinking the event driving the persisted query might be `PAGE_LOADED` or `HISTORY_CHANGED` or a combo potentially.

client/layout/navigation.js
```diff
@@ -22,15 +24,24 @@ import { isWCAdmin } from '../dashboard/utils';
...
 
 const NavigationPlugin = () => {
-	const [ persistedQuery, setPersistedQuery ] = useState(
-		getPersistedQuery()
-	);
+	const { persistedQuery } = useSelect( ( select ) => {
+		return {
+			persistedQuery: select(
+				NAVIGATION_STORE_NAME
+			).getPersistedQueryFromStore(),
```
joshuatf, line 31: Can we go ahead and use this same persisted queries in other components?  Or did you want to do a follow-up PR for this?

client/layout/navigation.js
```diff
@@ -22,15 +24,24 @@ import { isWCAdmin } from '../dashboard/utils';
...
+
+	const { setPersistedQuery } = useDispatch( NAVIGATION_STORE_NAME );
 
 	const pathIsExcluded = ( path ) =>
 		getQueryExcludedScreens().includes( getScreenFromPath( path ) );
 
 	// Update the persisted queries when history is updated
 	useEffect( () => {
+		if ( Object.keys( getPersistedQuery() ).length ) {
+			setPersistedQuery( getPersistedQuery() );
```
joshuatf, line 43: Event driven setting resolves this, but if we do plan to use this in other components we would want to make sure this is not only set in the navigation as it could be toggled off.

packages/data/src/navigation/selectors.js
```diff
@@ -16,3 +16,7 @@ export const getFavorites = ( state ) => {
 export const isNavigationRequesting = ( state, selector ) => {
 	return state.requesting[ selector ] || false;
 };
+
+export const getPersistedQueryFromStore = ( state ) => {
```
joshuatf, line 20:
```suggestion
export const getPersistedQuery = ( state ) => {
```

`FromStore` seems a little redundant.

joshuatf (commented):
This is an awesome solution, Joel.  💯 

Testing well and fixes the issue on my end.  Left a few comments about how we could structure this for use in other components.

packages/data/src/navigation/selectors.js
joelclimbsthings: This is actually what I originally had, but then I ended up with two separate `getPersistedQuery()` functions within the `NavigationPlugin` component (the other imported from Nav utils), which seemed confusing. So I made this one more specific.

joelclimbsthings:
Hey @joshuatf ,

I believe I've addressed your comments primarily with 925834d9ecaab675e44fb7ef123817d844ab886c. 

Testing instructions remains the same, but everything is now managed through the data store. I had to move `addHistoryListener()` to the navigation package as well. Let me know if you had something else in mind!

client/layout/navigation.js
joelclimbsthings: I've been looking at this a bit, although there are ~22 files that reference persisted queries and I'm a little uncertain about switching everything over before I vet this revised approach. Let me know what you think.

packages/navigation/src/index.js
```diff
@@ -197,6 +207,48 @@ export function updateQueryString(
 	getHistory().push( newPath );
 }
 
+/**
+ * Adds a listener that runs on history change.
+ *
+ * @param {Function} listener Listener to add on history change.
+ * @return {Function} Function to remove listeners.
+ */
+export const addHistoryListener = ( listener ) => {
```
joshuatf, line 216: Nice decision to move this over.  I think this may be something that could potentially benefit others.

packages/navigation/src/index.js
```diff
@@ -61,6 +61,16 @@ export const getQueryExcludedScreens = () =>
...
 
+/**
+ * Given a path, return whether it is an excluded screen
+ *
+ * @param {Object} path Path to check
+ *
+ * @return {boolean} Boolean representing whether path is excluded
+ */
+export const pathIsExcluded = ( path ) =>
+	getQueryExcludedScreens().includes( getScreenFromPath( path ) );
```
joshuatf, line 72: Just thinking out loud here (and definitely not within the scope of this PR)- wondering if this should have been included screens instead of excluded.  May be hard to maintain as more WCA pages are registered.

packages/data/src/navigation/reducer.js
```diff
@@ -25,6 +26,12 @@ const reducer = (
 				menuItems: [ ...state.menuItems, ...menuItems ],
 			};
 			break;
+		case TYPES.SET_PERSISTED_QUERY:
```
joshuatf, line 29: I think we can forego the `setPersistedQuery` entirely and use the events directly here instead of creating a setter.

```suggestion
		case TYPES.ON_HISTORY_CHANGE:
		case TYPES.ON_LOAD:
```

joshuatf (commented):
Fantastic work here, Joel.  Left one small change we can make to move away from setters and towards events, but this looks perfect otherwise.

packages/navigation/src/index.js
joelclimbsthings: 💯 . I was actually thinking this last time I was working on persisted queries, but was a little uncertain since there is a filter associated with it, `TIME_EXCLUDED_SCREENS_FILTER`, that I assume we'd still have to support.

I definitely think a whitelist would be more appropriate here.

packages/navigation/src/index.js
joelclimbsthings: 👍🏻

joelclimbsthings:
Thanks @joshuatf , made that change and another tweak or two in d103563869918ec0613e2cd9f21dcf754e032b32.

packages/data/src/navigation/reducer.js
joelclimbsthings: Good call, done in d103563869918ec0613e2cd9f21dcf754e032b32

packages/data/src/navigation/actions.js
```diff
@@ -89,6 +90,23 @@ export function removeFavoriteSuccess( favorite, error ) {
 	};
 }
 
+export function* onLoad() {
+	yield onHistoryChange();
+}
+
+export function* onHistoryChange() {
+	const persistedQuery = getPersistedQuery();
```
joshuatf, line 98: Just making a note here (doesn't need to be fixed for this PR), in hindsight I guess the original `getPersistedQuery` should have been something more along the lines of `getPersistedQueryFromParams`.

This may be a tiny bit confusing going forward since we have `getPersistedQuery` in the data store as well as a util method to get from params.  But we can tackle this in a follow-up.

joshuatf (approved):
Testing perfectly.  Really elegant solution here, Joel. 

Added a comment for a potential follow-up and it would be great to get an issue to follow-up on using this store elsewhere in the app.  But this branch LGTM 🎉 ✅ 💯

joelclimbsthings:
Great! 🥳  Just created an issue for follow-up -> woocommerce/woocommerce#32262

PR #6616, opened 2021-03-17
Fixed event tracking for merchant email notes

### LINKED ISSUE #6359: Add product email notification - improve Tracks
The event tracking that was implemented is not working properly:

- There's a mismatch between the user id of each event, so when you create a funnel the conv. rate is always 0. More info here: p4qSXL-3LI-p2. The events are:
`wcadmin_wcadmin_email_note_sent`
`wcadmin_wcadmin_email_note_opened`
`wcadmin_wcadmin_inbox_action_click`- `action name` : `wc-admin-add-first-product-note`
- Also the `wcadmin_wcadmin_email_note_opened` seems to have some issues:
We have more `wcadmin_wcadmin_email_note_opened` than `wcadmin_wcadmin_email_note_sent` and the number of event `uniques` and `counts` look the same. Usually we should have more `counts`.

Here's the FG definition: 
Uniques – the number of distinct users associated with an event has been recorded in a timeframe
Counts – the number of times an event has been recorded in a timeframe

octaedro:
Fixes #6359

This PR fixes the event tracking for merchant email notes.

- Create a brand new site.
- Install a plugin to log every sent email (you can use [WP mail logging](https://wordpress.org/plugins/wp-mail-logging/)).
- Install and active [this gist](https://gist.github.com/octaedro/864315edaf9c6a2a6de71d297be1ed88) to create an email note. Just download the file and install it as a plugin.
- After activating the plugin, press `Add Email Notes` to create a note.
- Now go to WooCommerce > Settings > Email (`/wp-admin/admin.php?page=wc-settings&tab=email`) and check the checkbox `Enable email insights` and save changes.
- You will need to run the cron so you can install a plugin like [WP Crontol](https://wordpress.org/plugins/wp-crontrol/)
- Go to Tools > Cron events (`/wp-admin/tools.php?page=crontrol_admin_manage_page`).
- Call the hook `wc_admin_daily` by pressing its `Run Now` link.
Captura de Pantalla 2021-03-17 a la(s) 17 00 11
- Go to Tools > WP Mail Logging Log (`/wp-admin/tools.php?page=wpml_plugin_log`) and verify the testing email note was sent.
- View the message and press `Test action` (a broken image will be visible under the button, but that's expected and only visible in a test environment).

**Now it's time to check the event recording**

- You should check that the recorded `userid` is always the same for the events:  `wcadmin_email_note_sent`, `wcadmin_email_note_opened` and `wcadmin_inbox_action_click`.
- You can use the Tracks live view (reference PCYsg-5hv-p2).

### DIFF
--- a/TESTING-INSTRUCTIONS.md
+++ b/TESTING-INSTRUCTIONS.md
@@ -4,6 +4,19 @@
 
 ## 2.2.0
 
+### Fixed event tracking for merchant email notes #6616
+
+- Create a brand new site.
+- Install a plugin to log every sent email (you can use [WP mail logging](https://wordpress.org/plugins/wp-mail-logging/)).
+- Install and active [this gist](https://gist.github.com/octaedro/864315edaf9c6a2a6de71d297be1ed88) to create an email note. Just download the file and install it as a plugin.
+- After activating the plugin, press `Add Email Notes` to create a note.
+- Now go to WooCommerce > Settings > Email (`/wp-admin/admin.php?page=wc-settings&tab=email`) and check the checkbox `Enable email insights` and save changes.
+- You will need to run the cron so you can install a plugin like [WP Crontol](https://wordpress.org/plugins/wp-crontrol/)
+- Go to Tools > Cron events (`/wp-admin/tools.php?page=crontrol_admin_manage_page`).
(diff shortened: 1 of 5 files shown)


louwie17 (approved):
Nice work @octaedro, looks good and tested well 🚀 
Could you add the testing instructions to the test instruction file as well?

octaedro:
This PR was rebased and the requested testing instructions were added.

PR #6617, opened 2021-03-18
Improve AddFirstProduct email note contents

### LINKED ISSUE #6363: Add product email notification: improve content
We need to make some changes in the email content, as a way to decrease the likelihood of being marked as spam and increase the chances of having a good open rate and CTR:

- Subject: we need to be more specific in the subject. We should mention the goal of the email (add your first product) and ideally personalize the content with the store's name.
- Title: the title is too generic. We should be more specific here as well. Eg: Add your first product
- Content: we should try to personalize the email content with the store's name and inform users that they can disable this type of email notifications in the settings.

ilyasfoo:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/6363

This PR improves the contents of AddFirstProduct email note:

- Add merchant's name to greet (with default as "Hi there")
- Update heading
- Update image

It also changes `NotificationEmail` to utilize WC_Email's capability for placeholder contents in email notes by using the `format_string` class function. This allows for dynamic customization and personalization in email notes.

### Screenshots
image

- Checkout this branch.
- Create a zip for testing with `npm run test:zip`.
- Update the installation date (we need a store between 2 and 5 days old). You can do it with an SQL statement like this:

```
UPDATE `wp_options` SET `option_value`=UNIX_TIMESTAMP(DATE_SUB(NOW(), INTERVAL 4 day)) WHERE `option_name` = 'woocommerce_admin_install_timestamp';
```

- Make sure the `woocommerce_merchant_email_notifications` option is set to `yes`:
```
UPDATE `wp_options` SET `option_value` = 'yes' WHERE `wp_options`.`option_name` = 'woocommerce_merchant_email_notifications';
```

- Run the `wc_admin_daily ` cron job (this tool can help [WP Crontrol](https://wordpress.org/plugins/wp-crontrol/)).

No changelog (landed in 2.2.0)


ilyasfoo:
@pmcpinto Add first product email contents are as per the screenshot, are the changes enough? Please let me know!

src/Notes/AddFirstProduct.php
```diff
@@ -57,7 +57,8 @@ public static function get_note() {
 		}
 
 		$content_lines = array(
-			__( 'Nice one, you’ve created a WooCommerce store! Now it’s time to add your first product.<br/><br/>', 'woocommerce-admin' ),
+			'{greetings}<br/><br/>',
+			__( 'Nice one, you’ve created a WooCommerce store! Now it’s time to add your first product and get ready to start selling.<br/><br/>', 'woocommerce-admin' ),
```
joshuatf, line 61: I know this was here prior to this PR, but I think the comma here is grammatically incorrect.  Either a terminal point or a semicolon may be more appropriate here.

Also, same comment about `sprintf` for `br` tags.

TESTING-INSTRUCTIONS.md
```diff
@@ -409,6 +409,23 @@ Scenario #2
 8. Click on the **Choose payment methods** task, it should not be displaying the **Woocommerce Payments** option.
 9. Go to **Plugins > installed Plugins**, check if the selected plugin features selected in step 4 are installed and activated.
 
+### Improve AddFirstProduct email note contents #6617
+
+- Install the plugin in a fresh site site.
```
joshuatf, line 414: Just to be extra safe, might be worth mentioning that the store should have 0 orders and 0 products.

Also `site site` 😄

joshuatf (commented):
Nice work, @ilyasfoo!  This is testing well, just a few clean-up comments.

ilyasfoo:
Thanks for the review @joshuatf! I've addressed most of the feedback in 624c18a. 

While I agree with the comments to use `sprintf` on the translations to separate HTML, I'd argue that changing it right now would add unnecessary work for our translators, since it has been already translated (I've checked at least for Spanish). Instead, I added a comment in the code to let the next person know if the texts are going to change in the future. Would you agree?

pmcpinto:
> @pmcpinto Add first product email contents are as per the screenshot, are the changes enough? Please let me know!

Yep, I think so. @elizaan36 can you just give a quick look as well? Thanks!

elizaan36:
Looks good to me👍

joshuatf:
Sorry for the late review on this one, @ilyasfoo.

> I'd argue that changing it right now would add unnecessary work for our translators since it has already been translated (I've checked at least for Spanish).

That's probably a fair assessment.  LGTM 🎉

ilyasfoo:
I think you forgot to approve @joshuatf 😄

joshuatf (approved):
Weird, I thought I approved.  LGTM still! 🚢

PR #6624, opened 2021-03-18
Adding typescript resolver to fix linting issues with imports

joelclimbsthings:
No changelog

Just a quick add of a missing dependency for the typescript resolver. This was missing in a recent PR that changed our Eslint config to use the typescript resolver, the result being linting errors for all of our imports:

image

**Edit:**

I'm actually unsure of why so many of the wordpress packages are still unresolved since they _do_ appear in `package.json` dependencies:

image

Is this to be expected, @samueljseay ?

### Testing instructions

-   Checkout branch and run `npm install`
-   Open file in editor such as [this one](https://github.com/woocommerce/woocommerce-admin/blob/cda3b39d082624385a885132b2879db5d3685848/client/wp-admin-scripts/beta-features-tracking-modal/container.js#L0-L1) (I'm using vscode) and ensure that _most_ of the imports do **not** have a linting error. Currently some wordpress packages will still have unresolvable imports.


samueljseay:
@joelclimbsthings I dunno if its expected but the problem was already there as a warning for quite a while which is why in our original eslint these were set as warnings. I will look at trying to fix this at some point though.

samueljseay (changes requested):
This looks great, it just needs to be changed to a dev dependency.

joelclimbsthings:
Ah, good catch @samueljseay . Seems I was typing in `yarn` again, and `--dev` doesn't work with npm 💥 . Just changed.

samueljseay (approved):

PR #6626, opened 2021-03-19
Call existing filters for leaderboards.

dgwatkins:
Fixes category data not being refreshed when selecting a different currency.
The currency filter was added using docs/examples/extensions/sql-modification/

The actual problem is that the selected currency is not part of the arguments, so the cache key is the same whatever currency you selected.

The currency here is just an example taken from your examples folder, it could be any filter.

### Accessibility

This PR doesn't affect accessibility.

-   Create an order with each of these currencies: ZAR, NZD, USD
-   For the sake of testing, you can do this by changing the currency setting each time.
-   Run the example provided with `npm run example -- --ext=sql-modification`
-   Go to the Analytics Dashboard and select different currencies
-   You will notice that the Categories leaderboard is not refreshed unless you refresh the browser

### Suggested solution

There is already a filter to add filter arguments. Calling it in these places, the new argument affects the cache key.


ilyasfoo (approved):
Thank you for making the fix! It tested well for me and I think the changes made sense.

PR #6632, opened 2021-03-22
Try: Use the store timezone to make time data requests

joshuatf:
This is an initial fix for the timezone issues outlined in p90Yrv-2cl-p2.

This PR is a superficial fix for a larger issue and most closely follows Option C from the post mentioned above.  This will make time requests and get periods based on the user store data.

There are still a number of follow-ups needed, even if we choose to only use store timezones, such as using potentially using UTC for API requests or allowing API requests with an appending timezone offset.

Note that tests are failing due to `MomentTimezoneDataPlugin` not being loaded in tests.  If anyone has insight into how to fix this, it would be greatly appreciated 🙇 

1. Create a store with a timezone significantly ahead of or behind the timezone you currently reside in.
2. Create a test order and mark complete.
3. Navigate to various analytics reports and note the time filter is based on the current store time.  E.g., If your store timezone is 12 hours ahead of your current time, you may see `1st - 23rd` instead of `1st - 22nd` for "Month to date" depending on your time of day.
4. Note that the recently added order shows up in analytics reports.

### DIFF
--- a/TESTING-INSTRUCTIONS.md
+++ b/TESTING-INSTRUCTIONS.md
@@ -4,10 +4,12 @@
 
 ### Use the store timezone to make time data requests #6632
 
-1. Create a store with a timezone significantly ahead of or behind the timezone you currently reside in.
-2. Create a test order and mark complete.
-3. Navigate to various analytics reports and note the time filter is based on the current store time.  E.g., If your store timezone is 12 hours ahead of your current time, you may see `1st - 23rd` instead of `1st - 22nd` for "Month to date" depending on your time of day.
-4. Note that the recently added order shows up in analytics reports.
+1. Go to Settings -> General.
+2. Set your store timezone significantly ahead of or behind the timezone you currently reside in.
+3. Create a test order and mark complete.
+4. Navigate to various analytics reports and note the time filter is based on the current store time.  E.g., If your store timezone is 12 hours ahead of your current time, you may see `1st - 23rd` instead of `1st - 22nd` for "Month to date" depending on your time of day.
+5. Note that the recently added order shows up in analytics reports.
+6. Change your timezone and repeat, testing with both locations (e.g., `Amsterdam`) and also UTC offsets (e.g., `UTC-6`).
 
 ### Add plugin installer to allow installation of plugins via URL #6805
 
--- a/TESTING-INSTRUCTIONS.md
+++ b/TESTING-INSTRUCTIONS.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+### Use the store timezone to make time data requests #6632
+
+1. Create a store with a timezone significantly ahead of or behind the timezone you currently reside in.
+2. Create a test order and mark complete.
+3. Navigate to various analytics reports and note the time filter is based on the current store time.  E.g., If your store timezone is 12 hours ahead of your current time, you may see `1st - 23rd` instead of `1st - 22nd` for "Month to date" depending on your time of day.
+4. Note that the recently added order shows up in analytics reports.
+
 ### Add plugin installer to allow installation of plugins via URL #6805
 
 1. Visit any admin page with the params `plugin_action` (`install`, `activate`, or `install-activate`) and `plugins` (list of comma separated plugins). `wp-admin/admin.php?page=wc-admin&plugin_action=install&plugins=jetpack`
--- a/packages/date/src/index.js
+++ b/packages/date/src/index.js
@@ -8,6 +8,8 @@ import { parse } from 'qs';
 
 export const isoDateFormat = 'YYYY-MM-DD';
 
+export const defaultDateTimeFormat = 'YYYY-MM-DDTHH:mm:ss';
+
 /**
  * DateValue Object
  *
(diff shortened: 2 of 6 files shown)


packages/date/src/index.js
```diff
@@ -136,7 +135,13 @@ export function getRangeLabel( after, before ) {
...
  */
 export function getLastPeriod( period, compare ) {
-	const primaryStart = moment().startOf( period ).subtract( 1, period );
+	const primaryStart = moment(
+		moment()
+			.tz( window.wcSettings ? window.wcSettings.timeZone : null )
+			.format( 'YYYY-MM-DD HH:mm:ss' )
+	)
+		.startOf( period )
+		.subtract( 1, period );
```
louwie17, line 144: Is the wrapping of the moment tz call necessary? could we just do:
`moment().tz( window.wcSettings ? window.wcSettings.timeZone : null ).startOf( period ).subtract( 1, period )` ?

Although it appears (when trying locally) that `moment().tz(null)` returns undefined, so maybe we want to make that more explicit, could create a helper function for that (given it is the same logic in the `getCurrentPeriod` method).

louwie17:
@joshuatf maybe you want to add `https://www.npmjs.com/package/moment-timezone-data-webpack-plugin` in the package `package.json`, that might fix the tests. We do have it in our main `package.json`.

Or you could mock the `moments` object and add a `tz` function.

packages/date/src/index.js
joshuatf: >Is the wrapping of the moment tz call necessary?

I think it my initial attempts it was since the new moment would include timezone data, but now that we use that moment when appending timestamps, I think we're good to directly use `.tz()`.

>could create a helper function 

Good idea.  Added a helper `getStoreTimeZoneMoment` and a constant `defaultDateTimeFormat`.

joshuatf:
> maybe you want to add https://www.npmjs.com/package/moment-timezone-data-webpack-plugin in the date package package.json

I think this plugin works its magic via Webpack so unfortunately we can't just add the dependency in.  Adding `moment-timezone` would work, but greatly increase the package bundle size.

Getting the `MomentTimezoneDataPlugin` working during tests would probably solve this issue, but I'm unsure of how to do that.

>Or you could mock the moments object and add a tz function.

I think this may be a bit of a code smell, but even trying to do this I found a lot of challenges in getting the original moment functions working.

I've fixed the tests by avoiding any timezones in testing, though ideally we would be able to test those as well.  Open to ideas if you have any to get this working.

TESTING-INSTRUCTIONS.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Try: Use the store timezone to make time data requests #6632
+
+1. Create a store with a timezone significantly ahead of or behind the timezone you currently reside in.
```
louwie17, line 7: Maybe make these a little more specific, by mentioning you can do this by going to **Settings > General** Timezone

louwie17 (approved):
This tested well, and changes look good!
I think adding a couple more tests would be beneficial, mocking `tz` out should be fine, given we only want to test the logic on our end (making sure it is being called with the timezone).

I was able to mock it successfully by doing this:
```
jest.mock( 'moment', () => {
	const moment = require.requireActual( 'moment' );
	moment.tz = jest.fn();

	return moment;
} );
```

louwie17:
@joshuatf it might also be a good idea to add to the readme that we are making use of the `window.wcSettings.timezone`, given it is an external package.

joshuatf:
>I think adding a couple more tests would be beneficial, mocking tz out should be fine, given we only want to test the logic on our end (making sure it is being called with the timezone).

Agreed @louwie17, but can't seem to get these tests working.  If you define the timezone in the global and then add the mock, I get the following error since presumably the function is chained and needs to be available on the returned value from `moment()`.

Screen Shot 2021-04-01 at 5 08 48 PM
Screen Shot 2021-04-01 at 5 08 42 PM

joshuatf:
>@joshuatf it might also be a good idea to add to the readme that we are making use of the window.wcSettings.timezone, given it is an external package.

Good call.  Added a note to the readme about this.

louwie17:
> > I think adding a couple more tests would be beneficial, mocking tz out should be fine, given we only want to test the logic on our end (making sure it is being called with the timezone).
> 
> Agreed @louwie17, but can't seem to get these tests working. If you define the timezone in the global and then add the mock, I get the following error since presumably the function is chained and needs to be available on the returned value from `moment()`.

Oops my bad I should of tried it a little better, to change the property of the return function we can use the `moment.prototype`, so doing something like this does work:
```
jest.mock( 'moment', () => {
	const moment = require.requireActual( 'moment' );
	moment.tz = true;
	moment.prototype.tz = jest.fn().mockImplementation( () => moment() );

	return moment;
} );
```
Also are `moment.tz()` and `moment().tz()` the same? if so, I would suggest using the first one, given the second instantiates `moment` first. (that's why I just set `moment.tz = true` in the mock, but should really be a function.

louwie17:
> Also are moment.tz() and moment().tz() the same? if so, I would suggest using the first one, given the second instantiates moment first. (that's why I just set moment.tz = true in the mock, but should really be a function.

Replying to my own comment, they are not the same. As mentioned here https://momentjs.com/timezone/docs/#/using-timezones/ and I believe you are using it correctly by using `moment().tz()`.
`moment().tz(timezone)` would convert the users current time into the timezones time
`moment.tz(time, timezone)` just sets the passed in time to the timezone (basically just appending the utc like `-04:00`).

joshuatf:
>Oops my bad I should of tried it a little better, to change the property of the return function we can use the moment.prototype, so doing something like this does work:

Nice 💯  Wasn't sure if we'd be able to mock the prototype of a mocked library more than once, but it looks to be working.

Added some tests around `getStoreTimeZoneMoment`.

If anyone else on @woocommerce/isotope has time, it would be great to get at least another set of 👀  on this.

jeffstieler (approved):
Code looks good and tested well for me using a numeric offset and a named zone. 👍

PR #6635, opened 2021-03-22
Ensure script asset.php files are included in builds

psealock:
Only dependency php files by the name of `index.min.asset.php` were getting included in builds. WC Admin scripts used a different pattern, `*.min.asset.php` for naming those files. This PR makes sure they are included as well.

1. `npm run build` or if you have `npm start` already running that should be fine too.
2. Create a zip of the repo in its current state, `./bin/make-zip.sh test-assets`.
3. Open `tests-assets.zip` and navigate to `dist/wp-admin-scripts/`
4. Make sure js files have an associated `*.min.asset.php` file


joshuatf (approved):
Testing well- thanks for the quick fix, Paul!  💯

PR #6640, opened 2021-03-23
Remove CI tests for WooCommerce 4.7

adrianduffell:
Removes the WooCommerce 4.7 job from the `Run PHP unit tests` Github Action.

This job was previously failing, however I don't think it is worth fixing because this version is lower than the minimum supported WooCommerce version (4.8).

### Detailed test instructions:

-   Confirm CI checks on this PR pass.

No changelog required.

PR #6641, opened 2021-03-23
Remove Composer v1 Downgrade in PHP Tests

### LINKED ISSUE #5784: Remove Composer v1 Downgrade in Tests
In PR https://github.com/woocommerce/woocommerce-admin/pull/5650, Composer is downgraded to version 1 to support installing WooCommerce dependencies. This delays the CI tests.

This should no longer be a requirement once WooCommerce 4.8 is released, which is currently scheduled for December 8th.

Once WooCommerce 4.8 is released, remove the Composer downgrade.

No alternatives.

As a high priority, WooCommerce 4.8 should be checked for compatibility with Composer 1. The CI tests will start using WooCommerce 4.8 as soon as it is released on GitHub.

If it is compatible with Composer 1, I think this can wait until the next cooldown.

adrianduffell:
Fixes #5784

Removes the Composer 1 downgrade when installing WooCommerce for PHP tests. This was added in PR #5650 due to incompatibilities with older WooCommerce versions.

The downgrade is no longer required since the minimum supported WooCommerce version (4.8) and higher supports Composer 2.

-   Confirm CI checks for PHP pass in this PR. 

Note: The PHP check using WooCommerce 4.7 can be ignored since that version is unsupported. I have created a separate [PR](https://github.com/woocommerce/woocommerce-admin/pull/6640) to remove it.

No changelog required.

### DIFF
--- a/bin/install-wp-tests.sh
+++ b/bin/install-wp-tests.sh
@@ -180,10 +180,8 @@ install_deps() {
 	git clone --depth 1 --branch $LATEST_WC_TAG https://github.com/woocommerce/woocommerce.git
 
 	# Bring in WooCommerce Core dependencies
- 	composer self-update --1
 	cd "woocommerce"
  	composer install --no-dev
-	composer self-update --2
 
 	cd "$WP_CORE_DIR"
 	php wp-cli.phar plugin activate woocommerce


louwie17 (approved):
Thanks for updating this @adrianduffell, awesome we don't have to worry about major composer versions anymore 😄

PR #6642, opened 2021-03-23
Update Paths to Exclude from Composer

### LINKED ISSUE #6588: Exclude Unused Paths From Bundle
**Describe the bug**

I was looking at how our package is bundled with WooCommerce Core and noticed some paths which could be excluded.

- [ ] [docker](https://plugins.trac.wordpress.org/browser/woocommerce/trunk/packages/woocommerce-admin/docker) 1.7KB
- [ ] [storybook](https://plugins.trac.wordpress.org/browser/woocommerce/trunk/packages/woocommerce-admin/storybook) 4.8KB
- [ ] [vendor/composer/installers](https://plugins.trac.wordpress.org/browser/woocommerce/trunk/packages/woocommerce-admin/vendor/composer/installers/src/Composer/Installers) 72KB

These paths contain development items so I think would be unused and safe to exclude? This would shave our bundle size a little.

**To Reproduce**

View package contents included in WooCommerce Core.

https://plugins.trac.wordpress.org/browser/woocommerce/trunk/packages/woocommerce-admin
**Expected behavior**

Expecting development files excluded from bundled package.

adrianduffell:
Fixes #6588

This excludes additional paths from the Composer package by adding `export-ignore` git attributes.

- docker
- storybook

These contain development files only so are not needed by consuming packages.

Note I don't think this is possible to test this with Composer until it is in a release. Instead, I'm using `git archive` to provide similar usage.

- Create a git archive 

```
git archive --output=./test.zip --format=zip HEAD
```

-   See that the `docker` and `storybook` paths do not exist in the zip.

No changelog required

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -23,3 +23,4 @@
 /config* export-ignore
 /client* export-ignore
 /docker* export-ignore
+/storybook* export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -22,3 +22,4 @@
 /docs* export-ignore
 /config* export-ignore
 /client* export-ignore
+/docker* export-ignore


octaedro (approved):
Everything looks good to me 🚀

PR #6644, opened 2021-03-23
Remove no-reply from inbox notification emails

### LINKED ISSUE #6362: Add product email notification: change the from and reply-to
We should try to find ways to differentiate the sender and the recipient, as this is a common spam pattern. WordPress core uses wordpress@domain.com, something similar could work for us. AFAIK the WooCommerce Core emails have the same issue: the sender and recipient is the same.

ilyasfoo:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/6362

Overrides [`get_headers`](https://github.com/woocommerce/woocommerce/blob/9398c894c943524928d5b181eaeba9b44eb7a8ad/includes/emails/class-wc-email.php#L410) function from WC_Email class to remove `reply-to` header.

The following tests using AddFirstProduct email note notification:

- Create a zip for testing with `npm run test:zip`.
- Update the installation date (we need a store between 2 and 5 days old). You can do it with an SQL statement like this:

```
UPDATE `wp_options` SET `option_value`=UNIX_TIMESTAMP(DATE_SUB(NOW(), INTERVAL 4 day)) WHERE `option_name` = 'woocommerce_admin_install_timestamp';
```

- Make sure the `woocommerce_merchant_email_notifications` option is set to `yes`:
```
UPDATE `wp_options` SET `option_value` = 'yes' WHERE `wp_options`.`option_name` = 'woocommerce_merchant_email_notifications';
```

- Run the `wc_admin_daily ` cron job (this tool can help [WP Crontrol](https://wordpress.org/plugins/wp-crontrol/)).
- You should receive an email, check the header and make sure `no-reply` is not present.


octaedro (approved):
You did a nice job here @ilyasfoo! The code looks good and is testing well on my end.

I think that maybe we could add testing instructions for our GlobalStep partners. We can avoid the SQL queries by doing the following:
- We can use [this gist](https://gist.github.com/octaedro/864315edaf9c6a2a6de71d297be1ed88) to create merchant email testing notes.
- And we can toggle the option `WooCommerce > Settings > Emails > Enable email insights` to enable the merchant email notes.

I'll pre-approve this PR 🙂

ilyasfoo:
Thank you for reviewing this @octaedro! However, I feel testing from our GlobalStep partner is unnecessary here since this change is too small. Thank you for the gist, will keep this in mind!

PR #6622, opened 2021-03-18
Set default value to array when op is 'contains'

moon0326:
Fixes #6618

This PR sets `array` as a default value when the specified op is `contains`. This essentially prevents an edge case where an option is missing and the default value is not set.

### Detailed test instructions:

1. Clone and start https://github.com/Automattic/woocommerce.com
2. Open `notifications.json.php` from woocommerce.com repository and find a rule that uses the `contains` operator and remove the `default` key. Please make a note of the option name.
3. Open `src/RemoteInboxNotifications/DataSourcepoller.php` from your WooCommerce Admin repository and change the datasource to your local woocommerce.com (woocommerce.test)
4. Make sure your local WooCommerce Admin database does not have the option from step #2
5. Install and activate [WP Crontrol](https://wordpress.org/plugins/wp-crontrol/)
6. Navigate to Tools -> Cron Events and run `wc_admin_daily` job
7. Check your debug log in `wp-content/debug.log`. You should see PHP error.

PR #6645, opened 2021-03-23
Check if value for contains comparison is an array

louwie17:
This is in addition to @moon0326 fix here: #6622 

This PR addresses an issue where the value used for the `contains` comparison in the Remote Inbox notifications could not be an array type, and this throws errors in php 8 when passed into `in_array` (warnings in php 7).
It gracefully defaults to an empty array now, but does show a warning in the WC logs, so it could be fixed.

Not adding this to testing instructions, as it is mostly an extra safety measure.

- Start a new store you can skip the onboarding wizard
- Install and activate the [WP Log viewer](https://wordpress.org/plugins/wp-log-viewer/) plugin
- Make sure your store is older then 3 days, you can do that by updating the `woocommerce_admin_install_timestamp` option to a timestamp older then 3 days, could use [epochconverter for this](https://www.epochconverter.com/).
- Create a product and publish it
- Check the log viewer and the WC logs (**WooCommerce > Status > Logs > log-(todays date)-(random hash)**)
- It should not include warnings or error in relation to the ComparisonOperator
- Now set the `woocommerce_task_list_tracked_completed_tasks` option to something that is not an array (string or boolean), like so with the cli: `wp option set woocommerce_task_list_tracked_completed_tasks false`
- Go back to your product, edit it, and click update (it should update successfully)
- No new debug logs should of appeared, but if you go to the WC logs it should show a warning like this:
`WARNING Operation comparison "contains" option value is not an array, defaulting to empty array.`.

### DIFF
--- a/src/RemoteInboxNotifications/ComparisonOperation.php
+++ b/src/RemoteInboxNotifications/ComparisonOperation.php
@@ -19,18 +19,6 @@ class ComparisonOperation {
 	 * @param string $operation     The operation used to compare the operands.
 	 */
 	public static function compare( $left_operand, $right_operand, $operation ) {
-		if ( $operation && strpos( $operation, 'contains' ) !== false && ! is_array( $left_operand ) ) {
-			$logger = wc_get_logger();
-			$logger->warning(
-				'Operation comparison "contains" option value is not an array, defaulting to empty array.',
-				array(
-					'operation'     => $operation,
-					'left_operand'  => $left_operand,
-					'right_operand' => $right_operand,
-				)
-			);
-			$left_operand = array();
-		}
 		switch ( $operation ) {
 			case '=':
 				return $left_operand === $right_operand;
--- a/src/RemoteInboxNotifications/ComparisonOperation.php
+++ b/src/RemoteInboxNotifications/ComparisonOperation.php
@@ -19,6 +19,18 @@ class ComparisonOperation {
 	 * @param string $operation     The operation used to compare the operands.
 	 */
 	public static function compare( $left_operand, $right_operand, $operation ) {
+		if ( $operation && strpos( $operation, 'contains' ) !== false && ! is_array( $left_operand ) ) {
+			$logger = wc_get_logger();
+			$logger->warning(
+				'Operation comparison "contains" option value is not an array, defaulting to empty array.',
+				array(
+					'operation'     => $operation,
+					'left_operand'  => $left_operand,
+					'right_operand' => $right_operand,
+				)
+			);
+			$left_operand = array();
+		}
 		switch ( $operation ) {
 			case '=':
 				return $left_operand === $right_operand;
--- a/src/RemoteInboxNotifications/OptionRuleProcessor.php
+++ b/src/RemoteInboxNotifications/OptionRuleProcessor.php
@@ -20,9 +20,26 @@ class OptionRuleProcessor implements RuleProcessorInterface {
 	 * @return bool The result of the operation.
 	 */
 	public function process( $rule, $stored_state ) {
-		$default_value = 'contains' === $rule->operation ? array() : false;
+		$is_contains   = $rule->operation && strpos( $rule->operation, 'contains' ) !== false;
+		$default_value = $is_contains ? array() : false;
 		$default       = isset( $rule->default ) ? $rule->default : $default_value;
 		$option_value  = get_option( $rule->option_name, $default );
+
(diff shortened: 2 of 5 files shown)


src/RemoteInboxNotifications/ComparisonOperation.php
```diff
@@ -19,6 +19,18 @@ class ComparisonOperation {
 	 * @param string $operation     The operation used to compare the operands.
 	 */
 	public static function compare( $left_operand, $right_operand, $operation ) {
+		if ( $operation && strpos( $operation, 'contains' ) !== false && ! is_array( $left_operand ) ) {
+			$logger = wc_get_logger();
```
adrianduffell, line 23: **Optional**

I think it's good how this won't fail silently, but I feel like the log message doesn't provide enough info to help someone troubleshoot the issue.

You could add logging earlier when the `woocommerce_task_list_tracked_completed_tasks` option is first retrieved, and check if it is an array then. This would allow a more specific message to be written about the issue.

adrianduffell (changes requested):
Nice work @louwie17 . This tested perfectly for me. Thanks for adding new tests too!

I've just left a few comments.

> Now set the woocommerce_task_list_tracked_completed_tasks option to something that is not an array (string or boolean), like so with the cli: wp option set woocommerce_task_list_tracked_completed_tasks false

Just noting I saw a few more warnings on the site after doing this. 

```
Warning: array_diff(): Expected parameter 1 to be an array, string given in /var/www/html/wp-content/plugins/woocommerce-admin/src/Features/OnboardingTasks.php on line 442
```
```
Warning: Invalid argument supplied for foreach() in /var/www/html/wp-content/plugins/woocommerce-admin/src/Features/OnboardingTasks.php on line 444
```

.github/workflows/test-php.yml
```diff
@@ -32,6 +32,11 @@ jobs:
                       woocommerce: '4.7.0'
                       phpunit: '6.5.9'
                       composer: '1.10.19'
+                    - php: '8.0'
```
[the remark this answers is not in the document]
louwie17: Yeah, it requires a specific `bin` package, I copied this from Woo Core. I agree a separate PR would be better (I thought it would be simple at first).

src/RemoteInboxNotifications/ComparisonOperation.php
louwie17: Yeah that makes sense, will move this back.

src/RemoteInboxNotifications/OptionRuleProcessor.php
```diff
@@ -20,10 +20,27 @@ class OptionRuleProcessor implements RuleProcessorInterface {
 	 * @return bool The result of the operation.
 	 */
 	public function process( $rule, $stored_state ) {
-		$default_value = 'contains' === $rule->operation ? array() : false;
+		$is_contains   = $rule->operation && strpos( $rule->operation, 'contains' ) !== false;
+		$default_value = $is_contains ? array() : false;
```
louwie17, line 24: Make use of $is_contains here as well, as this also matches the `!contains`.

.github/workflows/test-php.yml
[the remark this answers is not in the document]
louwie17: I reverted these changes :)

src/RemoteInboxNotifications/ComparisonOperation.php
louwie17: I moved it to the `OptionsRuleProcessor` instead adding the option name in the warning: https://github.com/woocommerce/woocommerce-admin/pull/6645/commits/a4956639f65384a267be0befeb33fb4225a28fda

louwie17:
@adrianduffell appreciate the suggestions, they are all addressed now.
Also I did add a fix for the `OnboardingTasks.php` warnings, given it was doing an `is_array` check for the old value already and not the new_value. This was mostly caused because of the example, I believe in theory `woocommerce_task_list_tracked_completed_tasks` should only be an array or undefined from our code.

adrianduffell (approved):
@louwie17 Great stuff, thanks for making the changes. The log message looks super helpful now 👍 

LGTM 🚀

.github/workflows/test-php.yml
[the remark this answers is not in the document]
adrianduffell: NP Woo Core’s approach for PHP 8 with the `bin` package looks good.

PR #6653, opened 2021-03-23
Add some guidelines on changelog and testing instructions.

samueljseay:
Global step have not been able to test all the instructions we include in `TESTING-INSTRUCTIONS.md` so I've added some clarifying info on what to document in the `CONTRIBUTING.md` to help contributors.

While there I also added some guidelines on the changelog to make those guidelines feel complete.

No changelog required.


.prettierignore
```diff
@@ -0,0 +1 @@
+*.md
```
samueljseay, line 1: I added this because I seem to be the only one who's environment is auto-formatting markdown. This stops prettier from doing that.

adrianduffell (approved):
Very cool! Thanks @samueljseay ! 🚀

.prettierignore
psealock: I also have formatting on for these files. It is a bit annoying to have formatting changes as part of the PR, but I really like it for difficult things like formatting tables. See this README I recently had to edit and add items to the table, it was pure bliss to have the autoformatting.

https://github.com/woocommerce/woocommerce-admin/blob/main/packages/dependency-extraction-webpack-plugin/README.md

samueljseay: @psealock I don't mind having auto formatting tbh. Lets just switch it on then and encourage everyone else to do so too.

Issue #6660, opened 2021-03-24
Implement experiment package in WC Admin and WC Payments


pmcpinto:
**Description:** one of the top priorities for cycle 11 is implementing an a/b testing package in the WC Admin and WC Payments plugins, as we're going to run experiments on both. Currently, there are two options: the ExPlat Calypso/JS Client released a couple of days ago (pbmo2S-HC-p2) and the work done with Raz (#6536). The goal is to assess which one is the best path forward and implement it on both plugins until the end of the cycle.

Below is a list of features that we'd like to have in the experiments package and during the assessment and implementation it would be good to confirm which ones are going to be available and if there's a possibility to implement them in the future:

**Targeting (Determine if a store is eligible for an experiment)**

- [ ] Can define a set of rules / conditions to determine if a store is eligible for an experiment
- [ ] Available dimensions for those rules:
- Location - store’s setting location (Country)
- Current plugin configuration
Is a plugin installed or not
Is the plugin active or not
What version of the plugin is installed
- Age of Store / GMV - to distinguish between new and established stores
- In an ideal world, we can target stores based on any data we collect about them or their admins via Tracks, WCTracker snapshots, etc.
- [ ] Stores associated with A8C employees / JN are automatically excluded from experiments
- [ ] Advanced: define subgroups to set up for [stratified random sampling](https://en.wikipedia.org/wiki/Stratified_sampling).

**Assignment (Determine what variant an eligible store will receive in an experiment)**

- [ ] Eligible stores (as determined by the targeting rules) are randomly assigned to an experiment variant (will only have a test and control variant to start with).
- Initially, simplified random sampling should work
- In the future, supporting stratified random sampling would be great.
- [ ] Once assigned to a variant, a user continues to receive the same variant through the duration of the experiment.
- If a store is assigned to a variant, all users that log into that store see the same experience
- If a user is assigned to a variant for one store, they see the same variant if they come across the same path in a different store
- [ ] If there is a problem assigning a target store for any reason, it receives the control variant
- [ ] We track how a store was assigned to a specific variant
- Normal assignment - randomly assigned
- Default assignment - fell back on default because of a problem
- User consistency - was assigned to ensure a user continues to receive the same experience
- [ ] Stores can be assigned to multiple experiments concurrently
- Stores are excluded from an experiment if they are already in another experiment affecting that part of the experience
[...]

PR #6669, opened 2021-03-25
Try ExPlat: Add A/A test

psealock:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/6660

Adds ExPlat package for use in WooCommerce environments. Also added is a simple A/A test to run as a trial experiment.

Bonus addition is changes to the build files so that changes to `*.ts` files trigger a rebuild on `npm start`

## Test instructions

Its hard to tell if the process is working because the details are in [Calypso's explat package](https://github.com/Automattic/wp-calypso/tree/trunk/packages/explat-client-react-helpers), but inspecting the Network and Console tabs can give us some clues.

1. Make sure you are opted into data sharing
2. Visit the Homescreen
3. See a request in the Network tab to `https://public-api.wordpress.com/wpcom/v2/experiments/0.1.0/assignments/woocommerce`
4. The payload will include an experiment name and anonymous id
5. Turn off data sharing WooCommerce > Settings > Advanced > woocommerc.com
6. Visit the Homescreen
7. See an error in the console about ExPlat turned off due to data sharing. This error will be caught and handled in production and removes the user from the experiment while rendering the default experience

## A Note about IE 11

`window.fetch` does not exist in IE 11, for which we will soon be dropping support soon. Since errors will be thrown in attempts to call `fetch` which will remove the user from the experiment, I've decided to simply let that happen instead of explicitly handling that scenario with a check for fetch and manually throwing an Error.


jessie-ross (approved):
**Looks nice and simple 😄**

> Getting the anonymous id assumes w.js and its cookie tk_ai are either already present, as would be the case with most of the team's dev environments. Some kind of polling can be added to ensure w.js has initially run.

Yep as this A/A only covers logged in participants we should be fine - but it shouldn't be used for experiments involving anonymous logged-out participants until this is fixed.

> No IE 11 support

This will affect results as we won't be getting an accurate picture of our users but it is probably a good business decision.

> Its hard to tell if its working

Yep! Haha! It is a different kind of development as we need something to work not individually but in aggregate.

But maybe you are talking about just testing the assignment for your logged in user, you can do that by creating an experiment in ExPlat using the WooCommerce platform ([Just need your approval on this PR to add it to the front-end](https://github.com/Automattic/abacus/pull/484)), and then hovering over the variation in your Experiment Overview to get a bookmarklet you can use to test with.

> I wonder what kind of controls can be made from remotely, ie duration, sample size, impression count, and ability to adjust mid-experiment

You can do the following while an experiment is staging:
- Sample size and variation proportion, you can use e.g. 5% control, 5% treatment to test 10% of the userbase.
- Add Exposure Events to further refine who is counted in an experiment (we do both assignment and event exposure analysis separately).
- You can enable and disable an experiment remotely (manually), and it is also automated by the start date and end date fields.

When an experiment is running:
- We explicitly don't allow adjusting any changes to targeting e.g. increasing/decreasing the participants. This is not sound experimentation wise. In the future we want to provide a rollout feature however.
- You can still add Metric Assignments - We backfill them for up to 14 days automatically and can go further manually.
- Stop an experiment by pressing a button.

client/lib/explat/index.ts
```diff
@@ -0,0 +1,67 @@
...
+		.then( ( response ) => response.json() );
+}
+
+const getAnonId = async (): Promise< string | null > => {
+	return new Promise( ( resolve ) => {
+		const cookies = cookie.parse( document.cookie );
+		// setTimeout( () => {
+		resolve( cookies.tk_ai || null );
+		// }, 2000 );
+	} );
```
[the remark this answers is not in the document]
psealock: Indeed, thanks. Fixed in https://github.com/woocommerce/woocommerce-admin/pull/6669/commits/565a8b5c742846d213efbbe607fa803d4b2bac63

client/lib/explat/index.ts
```diff
@@ -0,0 +1,67 @@
...
+};
+
+const logError = (
+	error: Record< string, string > & { message: string }
+): void => {
+	if ( isDevelopmentMode ) {
```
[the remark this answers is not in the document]
psealock: Sounds useful. Is this a Calypso package? I can `console.log` in development mode and use Calypso logging otherwise

client/lib/explat/index.ts
```diff
@@ -0,0 +1,67 @@
...
+		experiment_name: experimentName,
+		anon_id: anonId ?? undefined,
+	} );
+
+	// todo: handle IE 11 case, which does not support window.fetch
+	return window
+		.fetch(
+			`https://public-api.wordpress.com/wpcom/v2/experiments/0.1.0/assignments/woocommerce?${ params }`
+		)
+		.then( ( response ) => response.json() );
```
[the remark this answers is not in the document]
psealock: Nice, done in https://github.com/woocommerce/woocommerce-admin/pull/6669/commits/89c53bfe9990087c756ec3888f7846264abc340a

psealock:
>Yep as this A/A only covers logged in participants we should be fine - but it shouldn't be used for experiments involving anonymous logged-out participants until this is fixed.

Just to clarify here, what do you mean by "logged in". Users will always be logged in to wp-admin, but not necessarily wp.com or Jetpack. 

Doing some tinkering with this code using Jurassic Ninja sites, the anonymous id is successfully gathered whenever Jetpack is included. When it isn't the `tk_ai` cookie is `undefined`. This is odd because Woo also sets this cookie:

https://github.com/woocommerce/woocommerce/blob/23710744c01ded649d6a94a4eaea8745e543159f/includes/tracks/class-wc-tracks-client.php#L76

Taking a closer look, this code seems to be populating the `tk_aip` cookie instead. Its been a while since I've tinkered with that code, but any idea why that might be? I can't find the string `tk_aip` anywhere in the codebase so maybe its a browser thing.

I'm wondering if we can gather the anonymous id from `tk_aip` as well: 

```js
const getAnonId = async (): Promise< string | null > => {
	const cookies = cookie.parse( document.cookie );
	return cookies.tk_ai || cookies.tk_aip || null;
};
```

Then, this brings up a final point, which is privacy. We are prevented from gathering Tracks data from stores that have not opted into a data sharing agreement. Does this fall into the same category, @pmcpinto ?

jessie-ross:
> Just to clarify here, what do you mean by "logged in". Users will always be logged in to wp-admin, but not necessarily wp.com or Jetpack.

The crux is that we need to tie purchases and conversion events to experiment assignments, this is currently either through a `userId` or an `anonId`. If a user with an `anonId` logs in/signs up ExPlat will track that count all the events as from one user. I am not too sure of who is a user/anon for Woo.

> Doing some tinkering with this code using Jurassic Ninja sites, the anonymous id is successfully gathered whenever Jetpack is included. When it isn't the tk_ai cookie is undefined. This is odd because Woo also sets this cookie:

`w2.js` uses an existing `anonId` in `tk_ai` or it generates a new one.

There can be a race condition here as `w2.js` has to load and we have to send an event to guarantee an anonId for a `tk_ai` to be set.

I do recommend polling, you should be able to copy and paste from the Calypso Implementation if it is important, just need to change the references and likely remove the `getTracksLoadPromise` (unless you use the calypso analytics package):
https://github.com/Automattic/wp-calypso/blob/trunk/client/lib/explat/internals/anon-id.ts#L19-L110

> Taking a closer look, this code seems to be populating the tk_aip cookie instead. Its been a while since I've tinkered with that code, but any idea why that might be? I can't find the string tk_aip anywhere in the codebase so maybe its a browser thing.

`tk_aip` is used to store previous anonIds, it can actually be a comma separated list of previous anonIds, but if it is `tk_ai` should be set so it probably isn't an issue.

Search Automattic/analytics for `tk_aip` for more info. (Not sure how to link it properly).

client/lib/explat/index.ts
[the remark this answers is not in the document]
jessie-ross: It is the calypso implementation:
https://github.com/Automattic/wp-calypso/blob/a42ab16eac863365db2c25f9ce29771f19e4076c/client/lib/explat/internals/log-error.ts#L18-L56

You just need to remove the `getLogger()` path and reference - that is for SSR.

psealock:
>w2.js uses an existing anonId in tk_ai or it generates a new one.

Is there a difference, as of now, between w.js and w2.js? WooCommerce is [still enqueuing w.js](https://github.com/woocommerce/woocommerce/blob/9398c894c943524928d5b181eaeba9b44eb7a8ad/includes/tracks/class-wc-site-tracking.php#L51), this should probably be updated.

Something odd is happening because the `tk_ai` cookie generated by WooCommerce is being overwritten by `undefined`. `cookie.parse( document.cookie ).tk_ai` results in `undefined` being passed to `fetchExperimentAssignment`. 

>There can be a race condition here as w2.js has to load and we have to send an event to guarantee an anonId for a tk_ai to be set.

I'm not sure I follow exactly because I can inspect the value of cookies long after all scripts are loaded and executed:

Screen Shot 2021-03-30 at 2 46 05 PM

You can reproduce this by creating a new Jurassic Ninja site without Jetpack, but with WooCommerce. Any ideas to fix this?

See the woo flavored `tk_ai` being generated here, https://github.com/woocommerce/woocommerce/blob/9398c894c943524928d5b181eaeba9b44eb7a8ad/includes/tracks/class-wc-tracks-client.php#L206

psealock:
>If a user with an anonId logs in/signs up ExPlat will track that count all the events as from one user. I am not too sure of who is a user/anon for Woo.

As far as I understand, if an anonymous user with anonymous id connects their site to wp.com via Jetpack, they will get a real id. I forget which process (nosara?) then associates this new Jetpack connected wp.com id with the previous anonymous id so that all preceding events can be tied to the same user.

jessie-ross:
> Is there a difference, as of now, between w.js and w2.js? WooCommerce is still enqueuing w.js, this should probably be updated.

I just created a new WooCommerce site from jurassic.ninja and after accepting stats it looks like you load the same `w.js` and in Calypso which is `w2.js`.

See also p4qSXL-3BD-p2

> Something odd is happening because the tk_ai cookie generated by WooCommerce is being overwritten by undefined. cookie.parse( document.cookie ).tk_ai results in undefined being passed to fetchExperimentAssignment.

I couldn't replicate this by creating a new Jurassic Ninja site without Jetpack, but with WooCommerce. Might help to have precise instructions possibly a screencast.

> As far as I understand, if an anonymous user with anonymous id connects their site to wp.com via Jetpack, they will get a real id. I forget which process (nosara?) then associates this new Jetpack connected wp.com id with the previous anonymous id so that all preceding events can be tied to the same user.

It looks like this is occurring, it happens via `aliasUser` tracks event:

Screen Shot 2021-03-30 at 8 51 40 pm

Interestingly it looks like there is a separate cookie for Jetpack that isn't getting aliased, possibly because of the different paths involved and woo overriding:

Screen Shot 2021-03-30 at 8 53 40 pm

On closer look it seems this cookie only exists for jp+woo configuration, and Jetpack doesn't seem to need it as when you connect to jetpack logging in as a user is the first step.

jessie-ross:
After about 6 jurassic.ninja sites (hope it is ok to create so many) I have found the following:

- Woo loads `w.js` (`w2.js`)  after agreeing to the stats.
- Woo will create an anonId and send it with tracks requests.
- Connecting to Jetpack will trigger the aliasing of the anonId to the userId.
- Once Jetpack has been connected any tracks events will include the userId (including connecting to JP before Woo).

This seems pretty sane to me and gives us what we need to work with.

client/lib/explat/assignment.ts
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	const params = stringify( {
+		experiment_name: experimentName,
+		anon_id: anonId ?? undefined,
+	} );
+
+	try {
+		const response = await window.fetch(
+			`https://public-api.wordpress.com/wpcom/v2/experiments/${ EXPLAT_VERSION }/assignments/woocommerce?${ params }`
```
jessie-ross, line 32: It seems on Calypso Green we may need to send an OAuth token here for authenticated users (Calypso Blue uses cookies), this is something we have run into for Jetpack.

psealock: Jetpack connected users will have an actual id, not an anonId which is what we should be sending, if I understand correctly. 

As far as the OAuth token, this code won't be running on Calypso so it won't need to be included. Or will it?

psealock:
>This seems pretty sane to me and gives us what we need to work with

Fantastic, the `getAnonId` function here needs to replicate how `w.js` identifies the user to replicate the behaviour of Tracks. What do you think about adding a function `getUserIdentity` into the [w.js public API](https://github.com/Automattic/analytics/blob/ca7aa00d8363ce95b5e1627af22ee53580dcb6fe/lib/w/tracks.js#L411)? Otherwise, we'd need to duplicate that logic, which could be error prone.

>we would just need to initialise the ExPlat client after agreeing to the stats which might be necessary legally anyway.

👍🏽

packages/abtest/src/index.ts
```diff
@@ -0,0 +1,48 @@
...
+const exPlatClientReactHelpers = createExPlatClientReactHelpers( exPlatClient );
+const { Experiment: ExperimentOriginal } = exPlatClientReactHelpers;
+
+export const Experiment = ( props: {
+	name: string;
+	defaultExperience: React.ReactNode;
+	treatmentExperience: React.ReactNode;
+	loadingExperience: React.ReactNode;
+} ): JSX.Element | React.ReactNode => {
+	if ( window.wcTracks.isEnabled ) {
```
jessie-ross, line 43: Instead of this I recommend altering the `fetchExperimentAssignment` function, if it throws an error we return the default experience:

```js
export const fetchExperimentAssignment = async ( {
	experimentName,
	anonId,
}: {
	experimentName: string;
	anonId: string | null;
} ): Promise< unknown > => {

if (!window.wcTracks.isEnabled) {
    throw new Error('Tracking is disabled')
}
```

packages/abtest/src/error.ts
```diff
@@ -0,0 +1,46 @@
...
+		};
+
+		if ( isDevelopmentMode ) {
+			console.error( '[ExPlat] ', error.message, error ); // eslint-disable-line no-console
+		} else {
+			// Bail if using IE 11.
+			if ( ! window.fetch ) {
+				return;
+			}
+
```
jessie-ross, line 33: Might want to add a tracks check in here as well as I have a feeling sending error data back to the server without tracking consent might not be ok:

```
if (!window.wcTracks.isEnabled) {
    throw new Error(`Tracking is disabled, can't send error to the server.`)
}
```

psealock:
>>Something odd is happening because the tk_ai cookie generated by WooCommerce is being overwritten by undefined. cookie.parse( document.cookie ).tk_ai results in undefined being passed to fetchExperimentAssignment.

>I couldn't replicate this by creating a new Jurassic Ninja site without Jetpack, but with WooCommerce. Might help to have precise instructions possibly a screencast.

In order to reproduce, you'll need WooCommerce without Jetpack. As you noted, Woo + Jetpack produces two `tk_ai` cookies. When retrieving this value via `cookie.parse( document.cookie ).tk_ai` the Jetpack one is returned. I think this is fine, but when Jetpack isn't there, `tk_ai` is undefined and the Woo flavored anonymous id is stored in `tk_aip`. I assumed that Tracks was working properly, but it appears to be a `w.js` issue because the anonymous id is incorrectly being sent as well.

Here are steps to reproduce:

1. Head to [Jurassic Ninja](https://jurassic.ninja/) and click on "Want more options?"
2. Deselect "Include Jetpack"
3. Select "Include WooCommerce"
4. Click "Launch single site" and head to the site when ready
5. Open the dev console and choose the Application tab (in Chrome)
6. View the site's Cookies under Storage and see `tk_ai` be undefined and `tk_aip` populated with the Woo flavored anonymous id
7. If you're on Woo's Onboarding Wizard click "Skip setup store details" and be sure to click "Yes, count me in!" when asked if you want to participate in data sharing
8. In the console, head to the Network tab and filter requests by `t.gif`
9. Navigate to another wc-admin page like Analytics > Overview
10. See the resulting payload of `t.gif` request. `_ui` will have a value of `undefined`

This appears to be a bug in `w.js`, what do you think @jessie-ross ?

psealock:
Thanks @jessie-ross, Interesting! I saw the domains for `tk_ai` and `tk_aip` different on mine than your's, so I cleared them and started over. This resolved the issue

Screen Shot 2021-04-01 at 3 25 35 PM

packages/abtest/src/index.ts
psealock: I see this working. The majority of users won't have the data sharing turned on, so we'd need to be ok with this error being thrown for most people.

When an error is shown, is the Experiment thrown out for that user or do they fall into the default experience bucket? If so, I'm wondering if that may skew the data?

packages/abtest/package.json
```diff
@@ -0,0 +1,35 @@
+{
+	"name": "@woocommerce/abtest",
```
jessie-ross, line 2: It might be nice to call it `explat` rather than `abtest`, and we have plans to offer more than just a/b tests :)

packages/abtest/src/index.ts
jessie-ross: > When an error is shown, is the Experiment thrown out for that user or do they fall into the default experience bucket?

Throwing this error causes:
- No assignment to occur - they won't get included in the experiment analysis.
- The default experience to show - the user won't notice anything.

> I'm wondering if that may skew the data?

Good point!

There are two factors to consider here:
1. Internal Validity - the correctness of the results without attempting to generalise them to other populations.
2. External Validity - the extent to which results can be generalised to other populations.

Experimenting like this should be internally valid as whether or not a user turns on data sharing is independent of which variation they are assigned to.

However there will be limits to external validity, the population we are testing will be the population of experimenters that enable tracking, the extent to which this would generalise to the full population is hard to say and will likely depend on the nature of the experiment. There is the same issue with using tracks stats to determine Woo's development and I feel it is a matter of "This is the only data can get." as well as keeping in mind that this limitation exists.

I might get @yanirs to confirm that my understanding is correct.

yanirs: :+1: that's a good explanation @jessie-ross.

jessie-ross: Ah, I think it would be great to document this in the readme

psealock: Thanks for the explanation, it makes lots of sense. 

I think you've already answered and its probably a non issue, but just to clarify, I was asking about the mechanism of what happens to a user whose function call `fetchExperimentAssignment` throws an error. Is that user thrown out of the experiment and shown the default experience or is that user assigned to the default group and shown the default experience?

I just want to make sure it's not the latter, as that would cause problems because the user will be unable to send Tracks events which may be used to determine the effectiveness of the experiment.

The next question is: Is there a way to not show the error and/or log it in production? The majority of Woo users have opted out of data sharing, so there will be a lot of noise in LogStash and these users will be getting many errors thrown in the console. Can we avoid this?

jessie-ross: >  Is that user thrown out of the experiment and shown the default experience or is that user assigned to the default group and shown the default experience?

The user will be thrown out and shown the default experience - they won't affect the experiment results.

> The next question is: Is there a way to not show the error and/or log it in production? The majority of Woo users have opted out of data sharing, so there will be a lot of noise in LogStash and these users will be getting many errors thrown in the console. Can we avoid this?

If you include [this change](https://github.com/woocommerce/woocommerce-admin/pull/6669#discussion_r604715238) as well, users that have tracking disabled also won't send any log stash errors. This means users that are thrown out will effectively be invisible.

psealock: Thanks for the response and confirmation. I like the simplicity of this approach, but am hesitant to throw Errors when there isn't an error and the code is working as intended. Would an early return from `fetchExperimentAssignment` have the same effect as throwing an Error?

psealock: >but am hesitant to throw Errors when there isn't an error and the code is working as intended

I just tried a production build and those errors are caught and handled without being surfaced to the console, so I'm on board with this approach

psealock:
I have added throwing errors when users are opted out and rename the package to `explat`. This is now ready for review. Thanks @jessie-ross for the help along the way

package-lock.json
```diff
@@ -10369,6 +10389,16 @@
 				"@xtuc/long": "4.2.2"
 			}
 		},
+		"@woocommerce/abtest": {
```
jessie-ross, line 10392: I think this needs to be updated

packages/explat/src/anon.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * External dependencies
+ */
+import cookie from 'cookie';
+
+export const getAnonId = async (): Promise< string | null > => {
+	return cookie.parse( document.cookie ).tk_ai || null;
```
jessie-ross, line 7: I think we settled on the polling logic here?

packages/explat/src/assignment.ts
```diff
@@ -0,0 +1,39 @@
...
+			`Tracking is disabled, can't fetch experimentAssignment`
+		);
+	}
+
+	const params = stringify( {
+		experiment_name: experimentName,
+		anon_id: anonId ?? undefined,
+	} );
+
+	try {
```
jessie-ross, line 31: No need to wrap this in a try-catch block

packages/explat/src/index.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Internal dependencies
+ */
+import { isDevelopmentMode } from './utils';
+import { logError } from './error';
+import { fetchExperimentAssignment } from './assignment';
+import { getAnonId } from './anon';
+
+declare global {
```
jessie-ross, line 15: Ah I'm surprised this `declare` applies to the other files

jessie-ross (changes requested):
It's looking good so far, I think just the polling logic for `getAnonId` is missing.

packages/explat/src/anon.ts
psealock: Right, we did. I added it in https://github.com/woocommerce/woocommerce-admin/pull/6669/commits/d6d73f9a9330c3083517c7c73ee83b4b57cabdb9

packages/explat/src/assignment.ts
psealock: removed in https://github.com/woocommerce/woocommerce-admin/pull/6669/commits/f59c92ea5b720f0a7c5bf2f747d3af6c242c5396

packages/explat/src/index.ts
```diff
@@ -20,6 +20,8 @@ declare global {
 	}
 }
 
+initializeAnonId().catch( ( e ) => logError( { message: e.message } ) );
```
jessie-ross, line 23: This should be moved to run after tracks is enabled.

packages/explat/src/anon.ts
```diff
@@ -3,6 +3,63 @@
...
+				return;
+			}
+			attempt = attempt + 1;
+		}, anonIdPollingIntervalMilliseconds );
+	} );
+
+	return initializeAnonIdPromise;
+};
+
+export const getAnonId = async (): Promise< string | null > => {
```
jessie-ross, line 63: You should return early here if tracks is disabled, you can return a `null`, otherwise waiting might be indefinite if `initializeAnonId` isn't initialised  - if tracks isn't enabled.

packages/explat/src/index.ts
psealock: Do you mean a check for Tracks enabled?

```js
if ( ! window.wcTracks.isEnabled ) {
	initializeAnonId().catch( ( e ) => logError( { message: e.message } ) );
}
```

packages/explat/src/anon.ts
psealock: Great point, I added the check

packages/explat/src/index.ts
jessie-ross: Like after a user clicks the enable tracking button, I think that is here:

https://github.com/woocommerce/woocommerce-admin/blob/ef6aa388c40abffecb512e1af49c915ebe1449a6/client/profile-wizard/steps/usage-modal.js#L67-L74

and also here:

https://github.com/woocommerce/woocommerce-admin/blob/cda3b39d082624385a885132b2879db5d3685848/client/wp-admin-scripts/beta-features-tracking-modal/container.js#L22

I think within the `.enable()` callback is best.

psealock: Ah I see. Good call. I think a combination of what I suggested plus in the callbacks is best

psealock:
@jessie-ross Nice catch with initialization needing to happen after a user opts in because thats done in the Onboarding Wizard on new sites after which the user proceeds to the Homescreen without a page refresh. I added this in https://github.com/woocommerce/woocommerce-admin/pull/6669/commits/c8ddb9658d9703d2e696d650793e283e967c05c7 and confirmed its working by opting out, then running the Onboarding Wizard, opting back in, and seeing the Experiment assignment call functioning properly.

packages/explat/src/assignment.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const fetchExperimentAssignment = async ( {
+	experimentName,
+	anonId,
+}: {
+	experimentName: string;
+	anonId: string | null;
+} ): Promise< unknown > => {
+	if ( ! window.wcTracks.isEnabled ) {
+		throw new Error(
```
jessie-ross, line 16: Hmm, there is actually an interesting point to be made here about experiment loading behaviour after 

I may have been wrong about this, [your suggestion here](https://github.com/woocommerce/woocommerce-admin/pull/6669#discussion_r608174109) might be along the right tracks:

> returning `{"variations":{},"ttl":60}`

Particularly, I think we want to return `0` ttl here as once experiments are enabled we immediately want to be able to load them. Currently in the client throws an error for a `0` ttl but if we think it is a good idea I can remove that constraint.

packages/explat/src/index.ts
```diff
@@ -20,7 +20,13 @@ declare global {
 	}
 }
 
-initializeAnonId().catch( ( e ) => logError( { message: e.message } ) );
+export const initializeExPlat = (): void => {
+	if ( ! window.wcTracks.isEnabled ) {
```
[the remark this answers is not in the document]
psealock: Good catch, thank you. I added a check for `wcTracks` in https://github.com/woocommerce/woocommerce-admin/pull/6669/commits/ed6e18a042e1eb4c405bc9105d169c9a22b19256

packages/explat/src/assignment.ts
jessie-ross: > Particularly, I think we want to return 0 ttl here as once experiments are enabled we immediately want to be able to load them. Currently in the client throws an error for a 0 ttl but if we think it is a good idea I can remove that constraint.

Just to elaborate on this, the `ttl` is used to determine when to refetch the data, it is `60` in dev environments `3600` in live environments, so _if an experiment attempts to be loaded before `wctracks` is loaded_, it will take 1 hour before an experiment can apply to that user.

psealock: >we immediately want to be able to load them

In my testing, once a user opts in and fires the callback in `wcTracks.enable()` to initialize the anon id, the assignments aren't fetched until the user lands on the page with the Experiment. Are you suggesting calling that function right away and caching the results?

psealock: >if an experiment attempts to be loaded before `wctracks` is loaded, it will take 1 hour before an experiment can apply to that user.

I see, thanks for the context. If an fetchExperimentAssignment is called when the user isn't opted in an error is thrown and no `ttl` value is returned. Does that count as an attempt? If not, I think we are good in this respect.

https://github.com/woocommerce/woocommerce-admin/blob/ed6e18a042e1eb4c405bc9105d169c9a22b19256/packages/explat/src/assignment.ts#L15-L19

jessie-ross: > I see, thanks for the context. If an fetchExperimentAssignment is called when the user isn't opted in an error is thrown and no ttl value is returned. Does that count as an attempt? If not, I think we are good in this respect.

Ah my numbers are wrong actually: In the event of an error it doesn't fetch the `ttl` from the server but rather uses the default `60` seconds ttl.

This isn't horrible and will probably work well enough but might impact some edge cases at some point, this should be fine for the A/A experiment and something we can fix later.

packages/explat/src/index.ts
```diff
@@ -0,0 +1,47 @@
...
+declare global {
+	interface Window {
+		wcTracks: {
+			isEnabled: boolean;
+		};
+	}
+}
+
+export const initializeExPlat = (): void => {
+	if ( window.wcTracks?.isEnabled ) {
```
jessie-ross, line 24: Are we sure `isEnabled` is set within the `.enabled()` callback?

jessie-ross (approved):
Left a comment to be addressed, but other than that this looks ready to go.  🚢

packages/explat/src/index.ts
psealock: Yeah, thats a WC core function that sets `isEnabled` to `true`, loads `w.js` and executes the callback once `w.js` loads

https://github.com/woocommerce/woocommerce/blob/9398c894c943524928d5b181eaeba9b44eb7a8ad/includes/tracks/class-wc-site-tracking.php#L107-L108

moon0326 (approved):
LGTM 👍 x 1000 and tested well. I was able to see the error `[ExPlat]  Tracking is disabled, can't fetch experimentAssignment ` 🚢 🚀

PR #6658, opened 2021-03-24
Add Tests for Database Table Creation

adrianduffell:
Fixes #6038

Adds tests for database table creation.

### Detailed test instructions:

`npm run test:php -- --filter=WC_Admin_Tests_Install`

No changelog required.

PR #6670, opened 2021-03-25
Fix Database Table Creation Test

adrianduffell:
Follow up to PR https://github.com/woocommerce/woocommerce-admin/pull/6658

I noticed some issues with the `DROP TABLE` query in this test, sorry!

1. The syntax was wrong due to some typos 😅 
2. Things were getting muddled because the test suite uses temporary tables in tests. [More context](https://wordpress.stackexchange.com/a/220308).

The test was only passing because the tables already existed.

-   `npm run test:php -- --filter=WC_Admin_Tests_Install`

No changelog required.

### DIFF
--- a/tests/install.php
+++ b/tests/install.php
@@ -33,7 +33,7 @@ class WC_Admin_Tests_Install extends WP_UnitTestCase {
 		);
 
 		// Remove any existing tables in the environment.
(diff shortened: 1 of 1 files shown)


ilyasfoo (approved):
Tests well and LGTM! 🚢

PR #6678, opened 2021-03-25
Add support for php8 unit tests (locally and in GH actions)

louwie17:
This add's a test case in Github actions for running unit tests with PHP 8, I copied this logic from Woo core.
While wanting to test this locally first I also updated the php docker scripts to allow for running it with PHP 8, still some cleaning up to do there.

- `rm -r vendor` and run `composer install`
- Run normal php tests locally with `npm run test:php`, this should run correctly
- Run tests locally with php 8 using `PHP_VERSION=8 npm run test:php`, this should run the tests successfully with PHP 8
  *Note when you run PHP 8 tests it will modify a composer.json within `bin/composer/phpunit`, please revert these changes, the one down side of making use of a custom phpunit package just for version 8.
- Check if the github action unit tests for php 8 runs successfully (build should pass)


src/Notes/MerchantEmailNotifications/NotificationEmail.php
```diff
@@ -135,7 +135,7 @@ public function get_content_html() {
 		return wc_get_template_html(
 			$this->get_template_filename( 'html' ),
 			array(
-				'email_actions'           => $this->format_string( $this->get_actions() ),
+				'email_actions'           => $this->get_actions(),
```
louwie17, line 138: cc'ing @ilyasfoo given you [added the `format_string`](https://github.com/woocommerce/woocommerce-admin/pull/6617) last, in the PHP8 tests it error'ed as `actions` was an object trying to convert it to a string. From looking at the template I assume it should stay an object?

docker/wc-admin-php-test-suite/entrypoint.sh
```diff
@@ -47,9 +47,26 @@ else
...
 fi
 
+
 # Run the install script if the WordPress directory is not found.
 if [ ! -d /tmp/wordpress-tests-lib ]; then
 	install
 fi
 
-exec phpunit "$@"
+if [ "$(php -r "echo version_compare(PHP_VERSION,'8','>=');")" ]; then
```
louwie17, line 56: Same logic as what I have in the `test-php.yml` in the github workflows

docker/wc-admin-php-test-suite/Dockerfile
```diff
@@ -13,33 +15,10 @@ RUN apk --no-cache add \
...
-	php7-tokenizer \
-	php7-xdebug \
-	php7-xml \
-	php7-xmlreader \
-	php7-zip \
-	php7-zlib
+	wget
+	
+COPY install-php-version.sh /tmp/  
+RUN chmod u+x /tmp/install-php-version.sh && /tmp/install-php-version.sh
```
louwie17, line 21: It seems easier to handle the php version logic in a bash script for now, although I am open to suggestions

docker/wc-admin-php-test-suite/Dockerfile
```diff
@@ -1,7 +1,9 @@
-FROM alpine
+FROM alpine:3.13
```
louwie17, line 1: php 8 was not supported in alpine 3.12 and below: https://pkgs.alpinelinux.org/packages?name=php8&branch=v3.13

.github/workflows/test-php.yml
```diff
@@ -59,6 +64,17 @@ jobs:
                   node --version
                   npm --version
                   timedatectl
+            - name: Add PHP8 Compatibility.
```
louwie17, line 67: Same as the way woo core does it -> https://github.com/woocommerce/woocommerce/blob/trunk/.github/workflows/pr-unit-tests.yml#L57

composer.json
```diff
@@ -13,16 +13,19 @@
 		"automattic/jetpack-autoloader": "^2.9.1"
 	},
 	"require-dev": {
-		"phpunit/phpunit": "7.5.20",
 		"woocommerce/woocommerce-sniffs": "0.1.0",
-		"suin/phpcs-psr4-sniff": "^2.2"
+		"suin/phpcs-psr4-sniff": "^2.2",
+		"bamarni/composer-bin-plugin": "^1.4"
```
louwie17, line 18: Removed phpunit, the `composer-bin-plugin` handles this now by using the composer file under `bin/composer/phpunit`, it will use a custom one for running tests in PHP8

docker/wc-admin-php-test-suite/docker-compose.yml
```diff
@@ -2,15 +2,19 @@ version: '3'
 
 services:
     phpunit:
-        build: '.'
-        image: wc-admin-php-test-suite-phpunit:1.2.1
+        build:
+            context: '.'
+            args:
+                - PHP_VERSION=${PHP_VERSION:-7}
+        image: wc-admin-php-${PHP_VERSION:-7}-test-suite-phpunit:1.2.2
```
louwie17, line 9: Added the php version to the image names and also the volume name, so that we have a seperate image for PHP versions. This way it won't need to rebuild every time when switching versions.

src/Notes/MerchantEmailNotifications/NotificationEmail.php
ilyasfoo: You're right, thanks for catching this! I missed to test actions with `format_string`

docker/wc-admin-php-test-suite/Dockerfile
adrianduffell: I'm not sure we should pin the version, since then we can miss out on future security updates.

The way I understand it is the` image` version bump in `docker-compose.yml` should trigger the latest alpine image to be downloaded on the next use.

docker/wc-admin-php-test-suite/Dockerfile
adrianduffell: This sounds good. Since we're at the end of cool down I think it makes sense to try this approach out and we can iterate on it later if we need to.

docker/wc-admin-php-test-suite/docker-compose.yml
```diff
@@ -2,15 +2,19 @@ version: '3'
...
         volumes:
-            - 'test-suite:/tmp'
+            - 'test-suite-php-${PHP_VERSION:-7}:/tmp'
             - '../../:/app'
         environment:
             - WC_CORE_DIR=/tmp/wordpress/wp-content/plugins/woocommerce
-            - WC_VERSION=${WC_VERSION:-4.8.0}
```
adrianduffell, line 12: These were intended to be our minimum supported versions. The idea is that we use our min version as default so that less effort is needed for local development. This is because things that work in the min version should also work in the newer versions due to backwards compatibility.

I think I need to document this better 😄

docker/wc-admin-php-test-suite/Dockerfile
louwie17: Thanks, that indeed was fine, I removed it.

docker/wc-admin-php-test-suite/docker-compose.yml
louwie17: Yeah, I changed it because PHP8 is not support till WP 5.6 and WC 5.1 (I think). I changed this back and in the `entrypoint.sh` I added some logic that sets the version to 5.6 and 5.1 automatically if we are running PHP 8 and it doesn't have higher versions set.

louwie17:
@adrianduffell would you be able to have another look? I updated the items you mentioned (responded in the comment thread).

adrianduffell:
Great thanks @louwie17! This will be awesome to have in place. Sorry I didn't get a chance finish testing it all out today, but will do tomorrow.

adrianduffell:
> *Note when you run PHP 8 tests it will modify a composer.json within bin/composer/phpunit, please revert these changes, the one down side of making use of a custom phpunit package just for version 8.

@louwie17 Could we work around this by adding the paths to `.gitignore`? Sorry if that's a silly question, I haven't got my head around the `bin` package yet.

adrianduffell:
I get an error when switching back to PHP 7 after running the tests with PHP 8.

```
Error: wp-tests-config.php is missing! Please use wp-tests-config-sample.php to create a config file.
```

I think maybe it needs a state check where it automatically re-installs the tests when you switch the version, like how the` WP_VERSION` and `WC_VERSION` has. 

https://github.com/woocommerce/woocommerce-admin/blob/4a019b2a9cb3353c3c98b4b5923b811f621a7ac6/docker/wc-admin-php-test-suite/entrypoint.sh#L43-L47 https://github.com/woocommerce/woocommerce-admin/blob/4a019b2a9cb3353c3c98b4b5923b811f621a7ac6/docker/wc-admin-php-test-suite/entrypoint.sh#L32-L36

I'm not sure this is critical. I'm happy to work on it next cooldown if you prefer.

adrianduffell (changes requested):
@louwie17 This is great stuff, thank you so much for adding PHP 8 support and working through the complexities with PHPUnit. I think this will set us up to be in a great place with our testing. 

I've left a few suggestions and comments, but overall I think this is pretty close.

docker/wc-admin-php-test-suite/entrypoint.sh
```diff
@@ -47,9 +60,26 @@ else
...
+	composer install
```
adrianduffell, line 70: Do we need to run `composer install` here? I think ideally we would run it outside the container so that the tests can start sooner.

It also produced an error for me (see below). I got around it by running `composer install` from the host.

```
A script named bin would override a Composer command and has been skipped
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Your lock file does not contain a compatible set of packages. Please run composer update.

  Problem 1
    - squizlabs/php_codesniffer is locked to version 3.5.8 and an update of this package was not requested.
    - squizlabs/php_codesniffer 3.5.8 requires ext-simplexml * -> it is missing from your system. Install or enable PHP's simplexml extension.
  Problem 2
    - squizlabs/php_codesniffer 3.5.8 requires ext-simplexml * -> it is missing from your system. Install or enable PHP's simplexml extension.
    - wp-coding-standards/wpcs 2.3.0 requires squizlabs/php_codesniffer ^3.3.1 -> satisfiable by squizlabs/php_codesniffer[3.5.8].
    - wp-coding-standards/wpcs is locked to version 2.3.0 and an update of this package was not requested.
```

louwie17: Thanks, we indeed don't need to. I originally had it when the bin package wasn't part of the `composer.json`, but as long as you have run `composer install` locally it's all good.

louwie17:
> @louwie17 Could we work around this by adding the paths to .gitignore? Sorry if that's a silly question, I haven't got my head around the bin package yet.

Unfortunately not, the bin package relies on the composer.json defined for `phpunit` within the `bin/composer` folder, and uses that to install (the `post-install-cmd` script in composer.json will handle the installation automatically).
I don't see a way to ignore any future changes to the files that can shared remotely, I could only find this `git update-index --assume-unchanged bin/composer/phpunit/*`, which would only work locally.

louwie17:
> I get an error when switching back to PHP 7 after running the tests with PHP 8.

I think this should be fixed now, or at-least I can't reproduce it locally. You might want to `rm -r` your vendor folder and run `composer install` again, then run the tests.

Once WP updates there phpunit version to something higher then 7 (phpunit 7 doesn't support php8) we shouldn't need to use the `bin` package anymore.

adrianduffell (approved):
@louwie17 This LGTM! 🚀 

Thanks for the latest round changes. Confirming I didn't have an issue switching back to PHP 7 this time 👍 

Not going to lie the composer.json modifications are painful 😂 I'm happy for it to be tackled later though.

PR #6681, opened 2021-03-26
Build: Fix production script asset names

psealock:
Fixes issue found in https://github.com/woocommerce/woocommerce/pull/29479#issuecomment-806048649

Production builds remove the `.min` suffix in filenames so asset php filenames created by Dependency Extraction Webpack Plugin also reflect this.

https://github.com/woocommerce/woocommerce-admin/blob/4db37f0a8426a4fddd444043c65508803f72e10f/webpack.config.js#L61

This PR checks the `minified-js` feature to confirm a Core build before requiring `asset.php` files.

Confirm this breaks on `main`, then repeat with this branch to see the fix

1. Previous build artefacts will remain, so `rm -rf dist/`
2. Create a Core build `WC_ADMIN_PHASE=core npm run build`
3. Visit any wc-admin page
4. Check the Coupons page
5. Fire up the OBW (the first screen is sufficient)

Bonus: Check a development build with `npm start`

Note: No testing instructions have been added because everything will break if this isn't working, so common workflows will suffice.


samueljseay (approved):
LGTM 🚀

PR #6682, opened 2021-03-26
Improve page object encapsulation, refactor e2e suite, add more e2e tests.

samueljseay:
Thanks to @louwie17 's efforts to include TS and build the page object pattern into our tests. This takes those efforts another step further. The aim is to make basic page objects a lot easier to build by building a lot of the functionality into a base class.

Building on the pattern for encapsulating certain element types as well I've included a base class for these. There are 2 goals here: to minimize having to write constructors, which is often redundant, but also to provide a consistent interface of things you can do with a page object such as `navigate`.

The sections that were written before now function as pages as well, and in future I'd like to make them navigable because many of them are under certain circumstances.

There are also a lot of other changes made here with good reasons:

1. I removed most of the flow functions we had and put their functionality in the relevant page objects. It means a little more repetition in some places currently, but this explicit-ness makes each test scenario easier to reason about and figure out what is happening in my opinion. It was often a struggle to determine what each scenario was doing if it called a flow and overrode an argument in that flow. This makes each scenario mainly readable directly in the test and I do believe tests should be imperative in this way because it makes understanding them and changing them much easier. I could talk more about this, but I think there are plenty of reasons this is better.

2. I added helpers to the BasePage class wherever it helped reduce code repetition. Things I found really helpful like `clickButtonWithText` saved lots of time when writing follow up tests.

3. Instead of having to declare lots of elements as class members I have added helper methods to the base page that let you access things like dropdown, dropdowntypeahead and formtoggle by selector. These are also cached on the instance, so you can freely access them without declaring them as members and having to construct them.

4. Add Slack integration that reports test failures with screenshots, this should help debug failing e2e tests in CI.

5. Fix up the TS building so we can use `npx wc-e2e` directly. I removed all the scripts we had which just proxied to internal scripts in the wc e2e package.

6. Use incognito to fix some weird issues with redirects. Possibly to do with login/logout, but sometimes redirects to the profile wizard would occur randomly in the middle of the payment task tests, changing to incognito fixed this.

I have also added some new tests:

1. A smoke test of analytics pages that they don't crash
2. Payment tasks

No changelog required.


tests/e2e/elements/FormToggle.ts
```diff
@@ -0,0 +1,58 @@
+import { BaseElement } from './BaseElement';
+
+import { hasClass } from '../utils/actions';
+
+export class FormToggle extends BaseElement {
+	// Represents a FormToggle input. Use `selector` to represent the container its found in.
```
louwie17, line 6: The `switchOn` and `switchOff` buttons are quite nicely written, probably a bit safer then the `setCheckbox` & `unsetCheckbox` from the `@woocommerce/e2e-utils`, but it might still be worth to make use of those helper functions.

tests/e2e/pages/BasePage.ts
```diff
@@ -0,0 +1,139 @@
...
+import { BaseElement } from '../elements/BaseElement';
+import { DropdownField } from '../elements/DropdownField';
+import { DropdownTypeaheadField } from '../elements/DropdownTypeaheadField';
+import { FormToggle } from '../elements/FormToggle';
+import { getElementByText } from '../utils/actions';
+
+const config = require( 'config' );
+const baseUrl = config.get( 'url' );
+
+// Represents a page that can be navigated to
```
louwie17, line 11: Nice looking abstract class :)

tests/e2e/specs/activate-and-setup/basic-setup.test.ts
```diff
@@ -34,14 +33,16 @@ describe( 'Store owner can login and make sure WooCommerce is activated', () =>
 } );
 
 describe( 'Store owner can finish initial store setup', () => {
-	it( 'can enable tax rates and calculations', async () => {
-		const wcSettings = new WcSettings( page );
+	const wcSettings = new WcSettings( page );
```
louwie17, line 36: Did creating it here work? I thought I tried this before and `page` was undefined, although this would definitely be preferred, no fun creating it in each individual test.

tests/e2e/specs/activate-and-setup/complete-business-section.ts
```diff
@@ -1,37 +0,0 @@
-/**
```
louwie17, line 1: @samueljseay thanks for deleting these 👍 👍  and moving it to the tests, much nicer that way (it drove me a bit crazy, but clearly not crazy enough to change it :/ )

tests/e2e/utils/flows.ts
```diff
@@ -2,20 +2,19 @@
  * @format
```
louwie17, line 2: I suppose we could eventually do away with this as well given it's mostly in page objects now, aside from the extra login/logout logic.

louwie17:
@samueljseay I left a couple small comments, but looks fantastic 😄 you have structured it quite nicely.
Also thanks for writing those few extra e2e tests.

tests/e2e/utils/flows.ts
samueljseay: @louwie17 Yeah its on my hit list. 🔫 I'll probably remove this as part of this PR

tests/e2e/specs/activate-and-setup/basic-setup.test.ts
samueljseay: @louwie17 Yeah I think this worked fine, I haven't seen any issues.

tests/e2e/utils/flows.ts
samueljseay: I've removed this altogether.

tests/e2e/elements/FormToggle.ts
samueljseay: @louwie17 If it's ok with you, I want to leave this and pick it up next time. I have spent a lot of time on this already and I'm keen to get it shipped for now. Especially because its such a large refactor that will impact other PRs.

package.json
```diff
@@ -60,10 +58,8 @@
 		"prestart": "npm run -s install-if-deps-outdated",
 		"start": "cross-env WC_ADMIN_PHASE=development npm run build:packages && cross-env WC_ADMIN_PHASE=development npm run build:feature-config && concurrently \"cross-env WC_ADMIN_PHASE=development webpack --watch\" \"node ./bin/packages/watch.js\"",
 		"pretest": "npm run -s install-if-no-packages",
-		"test:e2e": "bash ./bin/wait-for-build.sh && ./node_modules/@woocommerce/e2e-environment/bin/e2e-test-integration.js",
-		"test:e2e-dev": "bash ./bin/wait-for-build.sh && DEBUG=1 ./node_modules/@woocommerce/e2e-environment/bin/e2e-test-integration.js --dev",
-		"test:e2e-debug": "bash ./bin/wait-for-build.sh && DEBUG=1 ./node_modules/@woocommerce/e2e-environment/bin/e2e-test-integration.js --dev --debug",
 		"test": "./node_modules/jest-24.9.0/bin/jest.js --config tests/js/jest.config.json",
+		"test:e2e": "WP_VERSION=latest npm run build && npx wc-e2e docker:down && npx wc-e2e docker:up && npx wc-e2e test:e2e",
```
louwie17, line 62: Nice work on using `npx` instead, much better then the absolute path I added.
~~I do think we should remove the `npm run build` part though, that way we can run `npm run dev` on the side and run the tests whenever.
The docker teardown and build up doesn't seem to take too long (although not sure what it will be like when we are just running a specific spec file and debugging).~~

Saw your updated readme with just using the `npx wc-e2e` commands look great to me.

tests/e2e/config/jest.config.js
```diff
@@ -7,7 +7,16 @@ module.exports = {
...
-	// 	'^.+\\.(ts|tsx|js|jsx)$': 'babel-jest',
-	// },
+	transform: {
+		'\\.[jt]sx?$': [
+			'babel-jest',
+			{
+				configFile: path.join(
+					__dirname,
+					'../../../',
+					'babel.config.js'
```
louwie17, line 17: Nice 👍

tests/e2e/config/jest-puppeteer.config.js
```diff
@@ -0,0 +1,10 @@
+const { useE2EJestPuppeteerConfig } = require( '@woocommerce/e2e-environment' );
+
+const puppeteerConfig = useE2EJestPuppeteerConfig( {
+	launch: {
+		browserContext: 'incognito',
+		args: [ '--incognito' ],
```
louwie17, line 6: Nice fix for the redirects 😄

README.md
```diff
@@ -32,29 +32,32 @@ For local development setup using Docker [see here](./docker/wc-admin-wp-env/REA
...
 ```
-npm run test:e2e
+npx wc-e2e test:e2e
 ```
 
 Run a single test by adding the file name:
 
 ```
-npm run test:e2e-dev complete-onboarding-wizard.test.js
+npx wc-e2e test:e2e-dev complete-onboarding-wizard.test.js
```
louwie17, line 60: This didn't work for me, I had to give the direct path `npx wc-e2e test:e2e-dev tests/e2e/specs/activate-and-setup/complete-onboarding-wizard.test.ts`

louwie17 (approved):
Great work on this @samueljseay this tests well, kudos 🚀 
Left one small comment in the readme, otherwise good to merge for me.

README.md
samueljseay: Yep this is a typo, thanks for spotting this, I'll fix it.

samueljseay: Fixed in https://github.com/woocommerce/woocommerce-admin/pull/6682/commits/cb25500d53c490637e966331f0251d59366cbe8c

